PR #837, opened 2026-04-21
feat: state &amp; computation primitives batch (v0.5.1 P2)

johnrtipton:
## Summary

Four small related v0.5.1 P2 primitives for derived state, dirty tracking, stable IDs, and cross-component context sharing. Batch PR ~230 LOC total per the pipeline's `--group` flow.

## What's in the box

1. **Memoized `@computed("dep1", "dep2")`** — `@computed` now accepts an optional tuple of dependency attribute names. When given, the value is cached on the instance and only recomputed when any dep's identity or shallow content fingerprint changes (id + length + key-subset matching `_snapshot_assigns` semantics). Plain `@computed` (no args) keeps property semantics — recomputes every access. React `useMemo` equivalent.

2. **Automatic dirty tracking — `self.is_dirty` / `self.changed_fields` / `self.mark_clean()`**. Baseline is captured after `mount()` completes by both the WebSocket consumer and the HTTP API dispatch view. Respects `static_assigns`, ignores private attrs. Use cases: "unsaved changes" warnings (`beforeunload`), conditional save buttons, optimized `handle_event`.

3. **Stable `self.unique_id(suffix="")`** — React 19 `useId` equivalent. Format: `djust-<viewslug>-<n>[-<suffix>]`. `reset_unique_ids()` at render boundaries.

4. **Component context sharing — `self.provide_context(key, value)` / `self.consume_context(key, default)`** — React Context API equivalent. `clear_context_providers()` resets.

## Notable design / impl details

- **`_ComputedProperty`** subclass of `property` lets djust attach metadata (`_is_computed`, `_computed_name`, `_computed_deps`) to the descriptor without runtime errors from `functools.wraps`.
- **Framework read-only properties are excluded from context**: `mixins/context.py` now skips class-level `property` descriptors whose `fset is None` and which aren't marked `_is_computed`. Without this filter, `is_dirty`/`changed_fields` leaked into the session-saved context and caused setattr errors during WebSocket-reconnect restoration to a fresh view. Regression test `test_ws_reconnect_restores_private_attrs` now exercises this path.
- **Dirty-tracking baseline hooks** are in both `websocket.py` (after `mount()` in the WS path) and `api/dispatch.py` (after `_instantiate_view` in the HTTP API path). `_capture_dirty_baseline` is feature-detected with `hasattr` so downstream consumers holding a custom view base without the method keep working.

## Tests

**22 new tests in `test_state_primitives.py`, all passing locally:**

[...]

PR #835, opened 2026-04-21
feat: auto-generated HTTP API from @event_handler (ADR-008, v0.5.1 headline)

johnrtipton:
## Summary

Ships **ADR-008**, v0.5.1's P1 headline. Opt-in `@event_handler(expose_api=True)` exposes a handler at `POST /djust/api/<view_slug>/<handler_name>/` with an auto-generated OpenAPI 3.1 schema at `/djust/api/openapi.json`.

This unlocks four caller classes that can't reach djust today:
- Mobile / native clients that don't hold WebSockets
- Server-to-server integrations and CLI scripts
- Cron / one-shot action triggers
- AI agents that consume OpenAPI-described tools

The HTTP transport is a **thin adapter** over the existing handler pipeline — it reuses `validate_handler_params()`, `check_view_auth()`, `check_handler_permission()`, and the same `_snapshot_assigns()` / `_compute_changed_keys()` diff the WebSocket path uses. **One stack, one truth** (manifesto principle #4). No parallel serializer hierarchy, no "validation runs in two places" drift.

## What's in the box

- `djust.api` package:
  - `DjustAPIDispatchView` (CBV) + functional `dispatch_api()` entry point
  - `SessionAuth` + pluggable `BaseAuth` protocol (auth classes can opt out of CSRF via `csrf_exempt = True`)
  - `api_patterns()` URL factory and `default_api_urlpatterns()`
  - `OpenAPI 3.1` generator with Python→JSON-schema type mapping (`int`, `float`, `bool`, `str`, `UUID`, `Decimal`, `datetime`, `list[T]`, `Optional[T]`)
  - Registry walking `LiveView` subclasses with exposed handlers (lazy duplicate detection — raises `ImproperlyConfigured` only when a conflicting slug is actually resolved)
  - Structured error responses: 400 / 401 / 403 / 404 / 429 / 500
  - Handler exception messages are **logged server-side only, never leaked to the client**
- `@event_handler(expose_api=True)` kwarg
- `LiveView.api_name` and `LiveView.api_auth_classes` read-only contract attributes
- `manage.py djust_audit` now lists every `expose_api=True` handler and flags those missing `@permission_required` (the ADR-008 candidate new rule — treat exposed handlers like `@csrf_exempt`)
- Full user guide at `docs/website/guides/http-api.md`, linked from `_config.yaml` and `index.md`
- `CHANGELOG.md` Unreleased entry

## Response shape

Success:
```json
[...]

PR #838, opened 2026-04-21
fix(ci): resolve playwright-tests E001 + security-tests coverage threshold

johnrtipton:
## Summary

Two pre-existing CI failures that have required `--admin` merge on PRs #835 and #837. Both are addressed here so subsequent PRs can merge normally.

## 1. playwright-tests — `djust_theming.E001`

Demo project had `djust.theming` in `INSTALLED_APPS` but its `TEMPLATES[0]['OPTIONS']['context_processors']` didn't include `djust.theming.context_processors.theme_context`, tripping the `djust_theming.E001` system check. This in turn blocked `manage.py migrate --noinput` in the Playwright workflow, causing the whole job to fail.

**Fix**: Added the context processor to demo_project settings. (`examples/demo_project/demo_project/settings.py`)

Verification (local):
- `python manage.py check` — 0 ERRORS (previously 1: E001)
- `python manage.py migrate --noinput` — runs clean

## 2. security-tests — coverage threshold

Workflow runs `pytest --cov-fail-under=75` over `djust.security`, `djust.uploads`, and `djust.validation`. Coverage has drifted to 69%.

**Fix**: Rather than lower the threshold, added `tests/unit/test_security_validation_hints.py` — 27 targeted tests covering `djust.validation`'s hint generators (`_get_string_to_int_hint`, `_get_string_to_float_hint`, `_get_string_to_bool_hint`, `_get_string_to_list_hint`, `_get_string_to_decimal_hint`), the `format_type_error_hint` router, `validate_parameter_types`, coercion edge cases in `_coerce_value` / `_coerce_single_value`, and `get_handler_signature_info` shape.

Coverage before → after:

| Module | Before | After |
|--------|--------|-------|
| `djust/validation.py` | 65% | **90%** |
| `djust/uploads.py` | 66% | 66% |
| `djust/security/*` | 74-100% | 74-100% |
| **Total** | **69%** | **76%** ✅ |

Security test suite: 233 → 260 tests passing.

## Why this matters

- Removes the need for `--admin` on future PR merges (the canonical path through Stage 13 Merge PR)
- Fixes a latent footgun: any app installing `djust.theming` without the context processor would have seen the same E001 error
- Locks in the validation hint-message wording — these are messages app developers see when template attribute types misalign

## Test plan

- [x] `manage.py check` passes with 0 ERRORS on demo_project
- [x] `manage.py migrate --noinput` succeeds on demo_project
- [x] `pytest tests/unit/test_security_*.py python/tests/test_security*.py --cov-fail-under=75` passes locally
- [x] All 260 security-test cases pass

### DIFF
--- a/examples/demo_project/demo_project/settings.py
+++ b/examples/demo_project/demo_project/settings.py
@@ -6,147 +6,147 @@ from pathlib import Path
 
 BASE_DIR = Path(__file__).resolve().parent.parent
 
-SECRET_KEY = 'django-insecure-demo-key-change-in-production'
+SECRET_KEY = "django-insecure-demo-key-change-in-production"
 
 DEBUG = True
 
-ALLOWED_HOSTS = ['*']
+ALLOWED_HOSTS = ["*"]
 
 # CSRF trusted origins for production
 CSRF_TRUSTED_ORIGINS = [
-    'https://djust.k8.trylinux.org',
-    'http://localhost:8002',
-    'http://127.0.0.1:8002',
-    'https://djust.org',
-    'https://www.djust.org',
+    "https://djust.k8.trylinux.org",
+    "http://localhost:8002",
+    "http://127.0.0.1:8002",
+    "https://djust.org",
+    "https://www.djust.org",
 ]
 
 # CSRF cookie settings for production HTTPS
 CSRF_COOKIE_SECURE = False  # Set to True only when DEBUG=False
-CSRF_COOKIE_SAMESITE = 'Lax'
+CSRF_COOKIE_SAMESITE = "Lax"
 CSRF_COOKIE_HTTPONLY = False  # Allow JavaScript to read the cookie
 CSRF_USE_SESSIONS = False  # Keep CSRF token in cookie, not session
 
 # Session cookie settings
 SESSION_COOKIE_SECURE = False  # Set to True only when DEBUG=False
-SESSION_COOKIE_SAMESITE = 'Lax'
+SESSION_COOKIE_SAMESITE = "Lax"
 
 # Allow iframes from same origin (for embedded demos on homepage)
-X_FRAME_OPTIONS = 'SAMEORIGIN'
+X_FRAME_OPTIONS = "SAMEORIGIN"
 
 INSTALLED_APPS = [
-    'django.contrib.admin',
-    'django.contrib.auth',
-    'django.contrib.contenttypes',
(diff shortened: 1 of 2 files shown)


johnrtipton (commented):
## Code Review — APPROVE (posted as comment; author cannot self-approve)

### What was done well

**Settings fix is minimal and correct.** `examples/demo_project/demo_project/settings.py:66` adds the exact string the check expects (`djust.theming.context_processors.theme_context`) — verified against `python/djust/theming/checks.py:98`, which does an `in` membership test on that literal. Local `manage.py check` now reports 0 ERRORS (491 pre-existing W/INFO items remain, none from E001).

**Tests are actually testing behavior, not padding.** `tests/unit/test_security_validation_hints.py:34-96` assert on the specific hint phrasing (`"could not be converted to int"`, `"data-count:int"`, `"treated as False"`) and on the `coercion_attempted=False` gating, which locks in the developer-facing messages. The `_coerce_single_value` bool tests at lines 157-164 enumerate the truthy/falsy string set rather than spot-checking one value. `test_validate_handler_params_*` covers four real branches: missing-required, unexpected-without-kwargs, accepted-with-kwargs, positional-binding.

### Verification

`pytest tests/unit/test_security_*.py python/tests/test_security*.py --cov-fail-under=75` → **260 passed, 1 skipped, coverage 75.93%**. Validation module: **65% → 90%**. The remaining 10% on `validation.py` is defensive fallbacks (`except Exception`, truncation `continue` branches, one unreachable `return []`) — fine to leave.

### Private-import concern (non-blocking)

The test file imports `_coerce_single_value` and `_coerce_value` directly (line 14-15). These are prefix-private and could be renamed without notice. Mitigation: the coverage these give is for branches hard to reach through the public `coerce_parameter_types`, and the tests are colocated with the module they cover, so a rename would obviously break them. Acceptable for now, but worth noting if either helper is refactored — add a short comment there?

### Edge cases considered

- Could adding the context processor break any test that depends on `theme_*` vars *not* being in context? Only demo_project settings changed, not any test settings module, so no — this is Playwright-scoped.
- Diff is noisy (~320 lines of single→double quote reformatting by ruff), but semantically the only real change in settings.py is the one added line at L66. The formatting churn is harmless and matches ruff config.
- The test file has a stray blank line (line 11 after `from __future__`), cosmetic only.

### Final verdict

APPROVE

PR #842, opened 2026-04-21
feat: LiveView testing utilities (v0.5.1 P2)

johnrtipton:
## Summary

Ships the v0.5.1 P2 "LiveView testing utilities" roadmap item. Seven new methods on `LiveViewTestClient` for Phoenix LiveViewTest parity.

## Methods

| Method | Reads from | Purpose |
|---|---|---|
| `assert_push_event(event_name, params=None)` | `view_instance._pending_push_events` | Verify a handler queued a push event (payload subset-match) |
| `assert_patch(path, params=None)` | `view_instance._pending_navigation` | Verify `live_patch` was queued |
| `assert_redirect(path, params=None)` | Same | Verify `live_redirect` was queued |
| `render_async()` | `view_instance._async_tasks` | Drain pending `start_async` / `assign_async` tasks synchronously |
| `follow_redirect()` | Last redirect in `_pending_navigation` | Mount the destination view and return a new `LiveViewTestClient` |
| `assert_stream_insert(stream_name, item=None)` | `_stream_operations` + `_streams` | Verify a stream insert landed |
| `trigger_info(message)` | Calls `view_instance.handle_info(message)` | Synthetic pubsub delivery for testing |

## Tests

**21 new test cases** in `python/djust/tests/test_testing_utilities.py`:
- 4 × `assert_push_event` (name match, payload subset, wrong name, payload mismatch)
- 3 × `assert_patch` (path, params subset, wrong-type-queued)
- 2 × `assert_redirect` (path match, wrong path)
- 3 × `render_async` (no-op, pending callback, drain-not-recursive)
- 3 × `assert_stream_insert` (by name, dict subset, wrong stream)
- 4 × `trigger_info` (invokes handler, records duration, no-op default, exception capture)
- 1 × `_require_mounted` parametrized over all 7 methods
- 1 × integration combining push_event + patch + stream_insert

## User-facing guide

`docs/website/guides/testing.md` covers every method with copy-pasteable examples. Linked from both `_config.yaml` and `index.md`.

## Stage 7 review fixes applied

The Stage 7 subagent review caught several blocking misses from my initial implementation — all addressed before this PR opens:

- CHANGELOG.md Unreleased entry added
- `docs/website/_config.yaml` + `index.md` nav entries added
- ROADMAP.md strikethroughs at priority matrix (line 22), milestone detail (line 707), feature list (line 1042)
- `follow_redirect` now validates the resolved class IS a `LiveView` subclass (docstring previously promised this; implementation didn't)
[...]

PR #845, opened 2026-04-21
feat(js): dj-dialog — native &lt;dialog&gt; modal integration (v0.5.1 P2)

johnrtipton:
## Summary

Ships `dj-dialog` — declarative opt-in for the HTML `<dialog>` element's built-in modal behavior. v0.5.1 P2 item.

### Usage

```html
<dialog id="settings" dj-dialog="open">
  <h2>Settings</h2>
  <button dj-click="close_settings">Close</button>
</dialog>
```

- `dj-dialog="open"` → `showModal()` (backdrop, focus-trap, Escape — all browser-native)
- `dj-dialog="close"` → `close()`

### Implementation

- Document-level `MutationObserver` watches for `dj-dialog` attribute changes AND `childList` mutations (VDOM morph inserting a new dialog works automatically).
- Single observer, not per-element listeners — re-registration-free.
- Idempotent: re-asserting the current state is a no-op.
- Gracefully ignores non-`<dialog>` elements carrying the attribute.
- Falls back to `setAttribute('open', '')` if `showModal()` throws (some browsers' inconsistent modal state).

### Tests

**8 JSDOM cases** in `tests/js/dj_dialog.test.js`:
- Opens on load with `dj-dialog="open"`
- Stays closed with `dj-dialog="close"`
- Flips open when attribute changes close → open
- Flips close when attribute changes open → close
- Syncs dialogs injected after initial load (VDOM morph)
- Ignores `<div dj-dialog="open">` (non-dialog element)
- Idempotent: re-asserting "open" doesn't re-call `showModal`
- Exposes `_syncDialogState` on `window.djust.djDialog` for test harnesses

### Metrics

- `client.js`: 36 → 37 source modules (+~80 LOC)
- No new Python-side code — browser handles focus trap / backdrop / Escape natively

### Test plan

- [x] 8 new JSDOM tests pass
- [x] Pre-commit hooks all pass
- [ ] CI should go green first try (post-#841 + #842 clean baseline)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,8 +10,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - **LiveView testing utilities (v0.5.1 P2)** — Seven new methods on `LiveViewTestClient` for Phoenix LiveViewTest parity: `assert_push_event(event_name, params=None)` verifies a handler queued a client-bound push event (payload match is subset-based so tests stay resilient to later payload additions); `assert_patch(path=None, params=None)` / `assert_redirect(path=None, params=None)` assert `live_patch` / `live_redirect` calls; `render_async()` drains pending `start_async` / `assign_async` tasks synchronously so subsequent assertions see their results; `follow_redirect()` resolves the queued redirect via Django's URL router and returns a new test client mounted on the destination view; `assert_stream_insert(stream_name, item=None)` verifies stream operations (item subset-match for dicts); `trigger_info(message)` synthetically delivers a `handle_info` message so pubsub / pg_notify handlers can be tested without real backend wiring. Full user-facing guide at `docs/website/guides/testing.md`. 21 new test cases. (`python/djust/testing.py`)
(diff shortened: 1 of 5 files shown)


johnrtipton (commented):

Small, well-tested, plan-aligned. `npm test -- tests/js/dj_dialog.test.js` → 8/8 pass. Downstream-app leak scan clean (exit=1 on both commits and diff). CHANGELOG + ROADMAP updates landed correctly (and this PR also unwinds the "LiveView testing utilities" row that had been prematurely marked shipped in a prior PR — good catch).

### Concerns addressed

1. **MutationObserver scope** — `attributeFilter: ['dj-dialog']` keeps the attribute-change path effectively free. The `childList: true` + `subtree: true` path IS the first document-wide subtree observer in the codebase (grep confirmed — every other module uses event delegation or hooks into existing patch lifecycle). In practice the per-mutation cost is a single `instanceof HTMLDialogElement` check + `querySelectorAll('dialog[dj-dialog]')` on the added subtree (not whole document) — negligible. Worth flagging for future: if/when a `djust:patched` lifecycle event lands, this module should migrate to drop `childList: true, subtree: true`.

2. **`setAttribute('open', '')` fallback** — not semantically modal (no backdrop, no focus trap, no top-layer) but satisfies the stated "at least visible" intent. `showModal()` throws mostly when the dialog is already open, which the `!el.open` guard prevents — so this path is truly exceptional. Acceptable. Nice-to-have: wrap the silent catch with `if (globalThis.djustDebug) console.warn(...)`.

3. **`<dialog open dj-dialog="open">` upgrade** — current code skips `showModal()` because `el.open` is already true, so the dialog renders as non-modal. Contradictory authoring pattern in practice (`open` attribute is specifically the non-modal form); suggest documenting "let `dj-dialog` own the state, don't mix with `open` attribute" as a follow-up.

4. **Observer cleanup** — never disconnected. Correct for a document-lifetime framework module; consistent with every other singleton listener in `static/djust/src/`.

5. **Security** — no concerns. Modal is user-visible, focus-trap is browser-native, Escape always works. No `innerHTML`, no CSP issues.

### Minor non-blocking suggestions

- Debug-guard the `catch (_e)` silent swallow with a `globalThis.djustDebug` warn.
- Debug-guard a warning when a non-`<dialog>` element carries `dj-dialog` so authors catch typos early.

**Verdict: APPROVE — ship it.**

PR #846, opened 2026-04-21
feat(formsets): {% inputs_for %} tag + FormSetHelpersMixin (v0.5.1 P2)

johnrtipton:
## Summary

Ships Django formset / inline-formset support for LiveView templates (v0.5.1 P2 roadmap item).

### Template

```django
{% load djust_formsets %}
<form dj-submit="save">
    {{ addresses.management_form }}
    {% inputs_for addresses as form %}
        <div dj-inputs-for="{{ form.prefix }}">
            {{ form.street }} {{ form.city }}
            <button type="button" dj-click="remove_row"
                    dj-value-formset="addresses"
                    dj-value-prefix="{{ form.prefix }}">remove</button>
        </div>
    {% endinputs_for %}
    <button type="button" dj-click="add_row"
            dj-value-formset="addresses">+ add</button>
</form>
```

### View

```python
from djust import LiveView
from djust.formsets import FormSetHelpersMixin

class AddressListView(FormSetHelpersMixin, LiveView):
    formset_classes = {"addresses": AddressFormSet}

        self.addresses = AddressFormSet()
        self._formset_data = {}
```

The mixin's `add_row` / `remove_row` event handlers are keyed on `dj-value-formset` and (for remove) `dj-value-prefix`. Free-standing `add_row()` / `remove_row()` helpers in `djust.formsets` are also exported for custom handlers.

## Semantics

- Add caps at `formset.absolute_max` (Django's 2000 default) — no unbounded memory
- Remove uses Django's standard `DELETE=on` flag → visible to `formset.deleted_forms` during validation; non-destructive
- `inputs_for_loop` exposes `counter`, `counter0`, `first`, `last` matching Django's `{% for %}` conventions

## Tests

**13 cases** in `python/djust/tests/test_formsets.py`:
- Template tag: iteration with prefixes, loop metadata, None-formset safety, syntax errors (missing `as`, wrong keyword)
- `add_row` helper: increment, cap at absolute_max, preserve existing row data
- `remove_row` helper: DELETE flag set, existing rows preserved
- Mixin: add bumps TOTAL_FORMS, remove marks delete, unknown-formset raises, missing args no-op

## Docs

- CHANGELOG Unreleased entry
- ROADMAP: strikethrough at line 744 (milestone detail) + line 935 (parity tracker)


johnrtipton (commented):
# Stage 11 Code Review — feat(formsets): `{% inputs_for %}` + `FormSetHelpersMixin`

**Downstream leak scan**: clean. **WIRING_CHECK**: passes (user-wired via `dj-click`, no framework call site needed). **Tests**: 13/13 new + 97/97 across adjacent batteries pass in 2.94s.

## Done well
- Clean two-layer API (bare helpers + opt-in mixin).
- Uses Django's standard `DELETE=on` deletion protocol — non-destructive, compatible with `formset.save()` flows.
- Tag degrades safely on `None`/missing variable.
- `as <var>` required syntactically — no `{% for %}` ambiguity.

## Critical (must fix)

### C1. `add_row`/`remove_row` break on any custom-prefix formset
`python/djust/formsets.py:39-44, 56-63, 81-83`

Both helpers derive prefix from `formset_cls.get_default_prefix()` — a class method returning the **static** default (e.g. `"form"`). Every real inline-formset usage instantiates with a custom prefix:

```
FS = formset_factory(F, extra=1)
fs = FS(prefix='addresses')          # user's mount()
# fs.prefix == 'addresses', FS.get_default_prefix() == 'form'

data = {'addresses-TOTAL_FORMS': '1', 'addresses-INITIAL_FORMS': '0'}
updated = add_row(FS, data=data)
# updated.prefix == 'form'          (class default)
# updated.data['form-TOTAL_FORMS'] == '1'       ← wrong bucket
# updated.data['addresses-TOTAL_FORMS'] == '1'  ← original untouched
```

On every `add_row` the increment writes to `form-*` while real data stays under `addresses-*`. Tests sidestep this by using the default prefix — anyone following the PR description (which shows `dj-value-formset="addresses"` for "addresses") hits it immediately.

**Fix**: accept `prefix` kwarg (or the whole formset instance); in the mixin, read it from the currently-bound formset `self.<name>.prefix`. Add a regression test instantiating with `prefix='addresses'`.

### C2. `absolute_max` cap ignores `max_num`
`python/djust/formsets.py:51`

`getattr(formset_cls, "absolute_max", 2000)` — but `absolute_max = max_num + DEFAULT_MAX_NUM (1000)`. A user who sets `max_num=5` gets `absolute_max=1005`:

```
FS = formset_factory(F, extra=1, max_num=5)
FS.absolute_max  # 1005
FS.max_num       # 5
```

So `add_row` lets the formset grow to 1005 rows before capping, contradicting the PR description's "caps at `formset.absolute_max` (Django's 2000 default)" claim. The real guarantee is "2000 or max_num+1000, whichever was configured".

**Fix**: respect `max_num` as the hard cap:

```python
max_num = getattr(formset_cls, "max_num", None) or getattr(formset_cls, "absolute_max", 2000)
hard_ceiling = getattr(formset_cls, "absolute_max", 2000)
new_total = min(current + 1, max_num, hard_ceiling)
```

Regression test: `max_num=5`, call `add_row` 10×, assert TOTAL_FORMS ends at `"5"`.

## Important (should fix)

- **I1. Silent `_formset_data` init** (`formsets.py:129, 142`) — if user forgets `self._formset_data = {}` in `mount()`, first `add_row` silently discards mount-time state (including `initial=...`). Either auto-init or raise a loud `AttributeError`. Doc-only contracts bite here.
- **I2. `add_row` at cap is a silent no-op** — no log, no hook, button still active in UI. Add a `logger.info` and/or document the cap-UX requirement.
- **I3. TOTAL_FORMS not decremented on `remove_row`** — correct Django protocol, but templates that iterate `{% inputs_for %}` will still render the deleted row unless they check `form.DELETE.value()`. Add a short recipe to the tag's docstring.
- **I4. Re-entrancy** — `self._formset_data` not guarded; `@background` handlers calling `add_row` race with live handlers. Document that these helpers aren't background-safe (or add a lock).

## Suggestions

- **S1.** `templatetags/djust_formsets.py:68-71` — `forms.index(form)` is O(n²) per iteration and wrong if two forms compare equal. Use `enumerate(forms)`.
- **S2.** Nested `{% inputs_for %}` tags collide on the flat `inputs_for_loop` context var — document no-nest or rename with parent-loop support.
- **S3.** Docstring at `templatetags/djust_formsets.py:5, :24` says "Pair with `djust.forms.FormSetHelpers`" — actual path is `djust.formsets.FormSetHelpersMixin`. Fix before merge, trivial.
- **S4.** `formset_classes = {}` on mixin is a mutable class attr; a `MappingProxyType` default would be more defensive.

## Summary

C1 and C2 are blockers — they break the primary use case the PR description sells (inline formsets with custom prefixes) and misrepresent the safety bound. I1 strongly recommended. S1 and S3 are one-line fixes worth folding in now.

**Verdict: CHANGES_REQUESTED**

johnrtipton:
Superseded — the full feature shipped directly to main via commit 335cce26 after a Stage 11 review-fix pass was accidentally committed to main instead of this branch.

**Main has the final version** with these Stage 11 fixes applied:
- **C1**: `add_row` / `remove_row` now accept an explicit `prefix=` kwarg; mixin passes the formset name as the prefix so custom-prefix formsets (e.g. `FS(prefix='addresses')`) work correctly instead of being silently routed to default-prefix `form-*` keys.
- **C2**: `_effective_cap()` now returns `max_num` when set (was returning a hardcoded 2000, allowing formsets with `max_num=5` to grow to 1005 rows before Django's validation-tolerance ceiling kicked in).
- **I1**: `_require_formset_data()` raises a clear `RuntimeError` if `mount()` forgot to set `self._formset_data`.
- **S1**: O(n²) `forms.index(form)` replaced with `enumerate()` in the template tag.

Tests went from 13 → 16 to cover the new regressions.

Docs bookkeeping (CHANGELOG + ROADMAP) is in #847.

Branch left intact in case diff inspection is useful; safe to delete anytime.

PR #848, opened 2026-04-21
feat(js): dev-mode error overlay (v0.5.1 P2)

johnrtipton:
## Summary

Next.js/Vite-style full-screen error overlay that pops up in the browser whenever a LiveView handler raises an exception AND Django `DEBUG=True`. Displays the server-sent Python traceback, the triggering event, an optional hint, and validation details — so you never have to switch back to the terminal to read a stack trace during dev.

## Why this was easy

Almost all the plumbing already existed:

- `djust.websocket.send_error` already includes `traceback` (last 3 frames), `debug_detail`, and `hint` in DEBUG mode.
- `03-websocket.js` and `03b-sse.js` already dispatch a `djust:error` CustomEvent on the `window` carrying those fields.

The only missing piece was a UI that listens for `djust:error` and renders a panel — this PR is that UI.

## What the overlay shows

Given a handler that raises `ValueError("amount must be positive")` on a `charge` event, the overlay renders:

- **Error** — the exception message (red, HTML-escaped)
- **Triggered by event** — `charge`
- **Traceback** — the last 3 stack frames
- **Hint** (when provided) — green box with an actionable suggestion
- **Validation** (when provided) — pretty-printed JSON of per-field errors
- **Detail** (when `debug_detail` differs from `error`) — the raw error before sanitization

Dismissal: Escape, the close button, or clicking the backdrop. A second error *replaces* the current panel rather than stacking — always see the latest failure.

## Security

- Every field is HTML-entity escaped before insertion. Regression test `escapes HTML in the error message to prevent XSS` locks this in.
- Gated on `window.DEBUG_MODE`. In production (`DEBUG=False`): (1) the overlay early-returns and renders nothing, (2) the server already strips `traceback` / `debug_detail` / `hint` from the error frame, so there's nothing to leak anyway.

## Files

- `python/djust/static/djust/src/36-error-overlay.js` — the new module, ~180 LOC
- `tests/js/error_overlay.test.js` — 10 JSDOM test cases
- `docs/website/guides/error-overlay.md` — user guide
- `docs/website/_config.yaml` + `docs/website/index.md` — nav links
- `CHANGELOG.md` + `ROADMAP.md` — strikethroughs and Parity Tracker row flip

client.js: 36 → 37 source modules. Gzipped delta ~1.6 KB.

## Test plan

- [x] 10 JSDOM tests pass locally (`npx vitest run tests/js/error_overlay.test.js`)
- [x] Build produces client.js containing the new module
- [x] CHANGELOG updated
- [x] Docs wired into _config.yaml AND index.md


johnrtipton (commented):
## Stage 11 Code Review — dev-mode error overlay

### Verdict

**APPROVE** — ship it. Well-scoped, defensively coded, and aligned with existing patterns.

### What I verified

1. **XSS surface** — Every server-sent field (`error`, `event`, `traceback`, `hint`, `debug_detail`, stringified `validation_details`) flows through `_escape` before being appended to the `html` string at `src/36-error-overlay.js:127-155`. The only unescaped content in the concatenated HTML is static markup (class names, the `×` button, section titles, the footer). `innerHTML = html` at line 151 is therefore safe. Regression test `escapes HTML in the error message to prevent XSS` at `tests/js/error_overlay.test.js:60-67` locks this in for `error`; worth noting the test only exercises the `error` field — a future change that accidentally interpolates another field would pass this test. Consider broadening to include `hint` / `traceback` / `event` in a future tightening pass.
2. **DEBUG_MODE gate** — `if (!window.DEBUG_MODE) return;` at `src/36-error-overlay.js:171` matches the exact pattern used elsewhere (`03-websocket.js:815`, `debug/00-panel-core.js:10`, `debug/14a-debug-warnings.js:9`, `debug/14b-latency-sim.js:10`). Consistent.
3. **z-index** — `2147483646` is strictly higher than the existing ceiling in the codebase (`999999` in the debug panel and `99999` in reconnect/rate-limit overlays). Overlay will render on top of everything djust ships, including the connection-error overlay. Good choice for a hard-error display.
4. **CSS scope** — All selectors in `_injectStyles` are prefixed with `#${OVERLAY_ID}`. No global leak.
5. **Listener lifecycle** — Two `window` listeners registered with no teardown. Fine for a singleton module loaded once per page; standard pattern in this codebase (connection overlay, debug panel). Not a leak.
6. **Consistency with `_showConnectionErrorOverlay`** — Similar structure (fixed overlay, innerHTML-built, inline/injected styles, `if (!window.DEBUG_MODE) return;`). New overlay is stricter: escapes all dynamic content, whereas the connection overlay only contains static strings. No shared extraction worth doing — the two have different content models (static diagnostic help vs. untrusted server data).

### Edge cases I checked

- **Malformed/missing `detail`** — `_onError` defaults to `{}` (line 173), `_render` defaults `errorText` to `"Server error"`. Safe.
- **`null` vs. empty string** — `_escape` handles `null`/`undefined` via the `s == null` guard.
- **`debug_detail === error`** — Correctly suppressed by the `debugDetail !== errorText` guard (test `omits debug_detail section when it duplicates the error message` covers this).
- **Repeated Escape presses with no overlay present** — `_onKeyDown` guards `if (overlay) overlay.remove();`, safe.
- **Error fired before DOMContentLoaded** — Listener is on `window`, registered synchronously when the IIFE runs; `document.body` may not exist yet if the script loads in `<head>` before body parse. In practice `client.js` is loaded at end of body, so this is moot, but a defensive `document.body || document.documentElement` would be harmless.

- **STYLE_ID drift**: PR description mentions `<style id="djust-eo-style">` but the code uses `djust-error-overlay-style` (line 29). Description is wrong; code is fine.
- **Manual-API bypass**: `window.djustErrorOverlay.show()` calls `_render` directly, bypassing the DEBUG_MODE gate. This is intentional (devtools escape hatch), but worth noting that a compromised page could call it in production. Since the overlay only renders data that the caller passes in, there's no data leak — just cosmetic. Not worth fixing.
- **A11y — focus management**: `role=\"alertdialog\"` is set but focus is never moved into the panel on open, and there's no focus trap. Screen-reader users will hear the dialog announce but keyboard users may not know where Tab goes. Acceptable for a dev-only tool; a future polish pass could call `overlay.querySelector('.djust-eo-close').focus()` on open.
- **Test coverage gap**: The 10 tests are solid for happy-path and XSS, but don't exercise: (a) escape being pressed twice in a row, (b) `detail` missing entirely (null event.detail), (c) non-string field values (e.g. `error: 42`). `_escape`'s `String(s)` coerce handles (c); worth a one-liner test to lock that in.

### Question

The footer says \"shown only when Django DEBUG=True\", but the gate is `window.DEBUG_MODE`. If someone sets `window.DEBUG_MODE = true` in a prod build (e.g. for a bug-repro session), the overlay activates even though Django's `DEBUG=False` strips `traceback`/`hint`/`debug_detail` server-side. Expected? The overlay would render just `error` + `event` in that case — still useful, just wanted to confirm the mental model.

### Files reviewed

- `/Users/tip/Dropbox/online_projects/ai/djust_project/djust/python/djust/static/djust/src/36-error-overlay.js`
- `/Users/tip/Dropbox/online_projects/ai/djust_project/djust/tests/js/error_overlay.test.js`
- `/Users/tip/Dropbox/online_projects/ai/djust_project/djust/docs/website/guides/error-overlay.md` — reads well, concrete example, no redundancy with developer-tools.md. Good.

**APPROVE**

johnrtipton:
### Stage 14 — Retrospective (abbreviated)

**Quality**: 9/10. Small, focused, gated on DEBUG_MODE, XSS-safe by construction, 10 JSDOM tests covering happy path + failure modes + regression for script injection. Leveraged existing server infrastructure (`send_error` already shipped `traceback`/`debug_detail`/`hint`) — this PR was pure UI with almost no new surface area.

**What worked**: Stage 7 + Stage 11 both passed on first try. The `window.DEBUG_MODE` gate matches existing codebase conventions (checked `03-websocket.js`, `debug/00-panel-core.js`). XSS regression test is explicit rather than implicit. Manual-invocation API (`window.djustErrorOverlay.show/.dismiss`) was noted but reasonable — devtools only, not reachable by attacker-controlled code.

**What to improve**: Consider future extraction of a shared overlay-injection helper if a third full-screen overlay ships (connection error + this + ??? would be the trigger). Currently the duplication with `_showConnectionErrorOverlay` is justified (different content models — static help text vs. untrusted server data).

**Shipped**: Commit on main is the squash of PR #848. v0.5.1 P2 error overlay item marked ✅ in ROADMAP and CHANGELOG. Guide at `docs/website/guides/error-overlay.md`, linked from `_config.yaml` and `index.md`.

Remaining v0.5.1: Type-safe template validation (`djust_typecheck`) — in progress.

PR #849, opened 2026-04-21
feat: type-safe template validation — manage.py djust_typecheck (v0.5.1 P2)

johnrtipton:
## Summary

Static analysis that reads every LiveView template, extracts every variable and tag reference, and reports names not covered by the view's declared context.

Neither Phoenix nor React catches template-variable typos statically without an external type system (TypeScript, Flow). This is a genuine djust differentiator — catching `{{ usre.name }}` at CI time transforms the debugging experience.

## What counts as "declared context"

Union of:
1. Public class attributes (MRO-traversed, excluding framework bases)
2. `self.foo = ...` assignments anywhere in the class (AST-extracted from mount(), handlers, helpers — never runs the code)
3. `@property` methods
4. Literal `return {...}` dict keys from `get_context_data`
5. Template-local bindings from `{% for %}`, `{% with %}`, and `{% inputs_for fs as form %}`
6. Framework built-ins: `user`, `request`, `perms`, `csrf_token`, `messages`, `forloop`, `djust`, `is_dirty`, `changed_fields`, `async_pending`, etc.
7. `settings.DJUST_TEMPLATE_GLOBALS` (project-wide escape hatch)

## Silencing

Three progressively-broader escape hatches:

```django
{# djust_typecheck: noqa name1, name2 #}   <!-- per-template -->
```

```python
class CheckoutView(LiveView):
    strict_context = True   # elevates this view's findings to errors
```

```python
# settings.py
DJUST_TEMPLATE_GLOBALS = ["navbar", "site_title"]
```

## Design notes

Chose **pure Python regex + AST** instead of Rust AST extraction. Simpler to iterate; perf headroom isn't needed for a CI-time check that runs in <100ms on a typical project. The Rust template engine's AST remains available if we want to switch later.

The checker errs on the side of false positives over false negatives — a flagged name is either a real bug or a name you should tell the checker about.

## Files

- `python/djust/management/commands/djust_typecheck.py` — the command, ~260 LOC
- `python/djust/tests/test_djust_typecheck.py` — 14 test cases
- `docs/website/guides/typecheck.md` — user guide
- `docs/website/_config.yaml` + `docs/website/index.md` — nav links
- `CHANGELOG.md` + `ROADMAP.md` — strikethroughs

## v0.5.1 Milestone

This is the **last P2 item** for v0.5.1. After this lands, the milestone is feature-complete:

- ✅ LiveView testing utilities (#842)
- ✅ `dj-dialog` native `<dialog>` (#845)
- ✅ `inputs_for` nested formsets (directly on main + #847 docs)
- ✅ Error overlay dev mode (#848)
- ✅ Type-safe template validation (this PR)

## Test plan

- [x] 14 pytest cases pass locally
- [x] `manage.py djust_typecheck` runs against the demo project and surfaces real unresolved names
- [x] `--json` / `--strict` / `--app` / `--view` flags work
- [x] Pragma, `strict_context`, and `DJUST_TEMPLATE_GLOBALS` silencing all tested
- [x] CHANGELOG + ROADMAP updated
- [x] Docs wired into _config.yaml AND index.md


johnrtipton (commented):
## Review — `djust_typecheck` (Stage 11)

Solid differentiator PR. The design choice (pure Python regex + AST over Rust AST) is well-argued and the three-tier silencing (pragma / `strict_context` / `DJUST_TEMPLATE_GLOBALS`) is the right UX. Tests and docs are honest about limitations. A few concrete issues below.

### Blocking

**1. `AnnAssign` is silently dropped (line 372-389).** Only `ast.Assign` and `ast.AugAssign` are handled, but `self.counter: int = 0` parses to `ast.AnnAssign`. Typed mount assignments are idiomatic in djust codebases (demo, CLAUDE.md examples) — this will generate avoidable false positives on exactly the views most likely to be "well-typed." Add an `isinstance(node, ast.AnnAssign)` branch that mirrors the Assign logic on `node.target`. This is also probably a contributor to the 230-line demo noise.

### Non-blocking

**2. `_TAG_RE` terminates at the first `%}` even inside string literals (line 299).** `{% include \"foo %} bar\" %}` parses as `('include', '\"foo')`. Practically unlikely in real templates, but worth a comment acknowledging the limitation.

**3. Several Django tag shapes produce false negatives.** `{% firstof a b c %}`, `{% cycle r1 r2 %}`, `{% blocktrans with name=expr %}...{% endblocktrans %}` — their referenced names fall through the `if/elif/for/include/url/static` cascade in `_extract_referenced_names` (line 449-472) and are never extracted. `blocktrans with x=y` also binds `x` as a template-local that's missing from `_extract_template_locals` — so `{{ x }}` inside a blocktrans block will be falsely flagged. Worth at least a note in the \"Known limitations\" doc section, or a TODO comment.

**4. Class inheritance gap (line 341).** `_extract_context_keys_from_ast(cls)` reads only `inspect.getsource(cls)`, not its bases. A view that inherits `mount()` from a user mixin won't pick up the mixin's `self.x = ...` assigns. `_public_class_attrs` walks MRO, so the contrast is inconsistent. Walking `cls.__mro__` (skipping framework modules with the same predicate already used) would fix it.

**5. `_walk_subclasses` + `_is_user_class` duplicated from `djust_audit.py`.** Third copy in the tree — extract to `djust.management._utils` when the next consumer lands.

**6. Test coverage gaps.** No direct test for: `AugAssign` extraction, `@property` extraction, `DJUST_TEMPLATE_GLOBALS` resolution, `_find_template_path` returning None, or multi-line tag parsing. Point 1 above wouldn't have slipped through with a `self.x: int = 0` fixture.

**7. Exit-code UX (line 622).** `nonzero = options.get(\"strict\") or any_strict`: a view with `strict_context=True` forces nonzero exit project-wide even without `--strict`. Intended per the docstring (\"regardless of the global `--strict` flag\"), but a user seeing their CI fail because one view opted in may be surprised. Worth a line in the guide's pragma section making this explicit.

### Edge cases I checked
- Multi-line `{% if ... %}` tags: OK (`[^%]` matches newline).
- `{% for k, v in items.items %}`: OK — `_IDENT_RE.findall` extracts `k, v` correctly.
- REPL-defined class: `OSError` caught at line 357; good.
- Pragma regex: bounded quantifiers, no ReDoS risk.
- Template path traversal: `get_template` returns trusted loader paths — no issue.
- `_ALWAYS_AVAILABLE`: missing `TIME_ZONE`, `LANGUAGES`, `LANGUAGE_CODE` (i18n context processor) and `block.super` (as a sub-identifier after `.`, so actually fine via root-identifier extraction — `block` is already there). Users with `django.template.context_processors.i18n` enabled will see false positives unless they add to `DJUST_TEMPLATE_GLOBALS`; worth calling out in the guide.

### Verdict

`REQUEST_CHANGES: AnnAssign support (item 1) before merge.` The other items are non-blocking and can land as follow-ups, but item 1 will materially improve the first-run experience and is ~4 lines.

johnrtipton:
### Stage 13 — Re-Review

Stage 11's blocking finding addressed: `AnnAssign` now handled at `python/djust/management/commands/djust_typecheck.py:172-181` (new branch mirrors the existing `Assign` / `AugAssign` logic).

Test coverage expanded 14 → 19 cases (commit cca5f396):
- **Regression lock-in**: `test_extract_context_keys_from_ast_finds_annotated_assignments` — asserts `self.count: int = 0` is captured.
- Also now covered: plain + augmented `self.x` assigns, `@property` (with private-name exclusion), `DJUST_TEMPLATE_GLOBALS` setting resolution, template-not-found error surfacing.

All 19 tests pass locally. Non-blocking Stage 11 items (regex string-literal handling, `{% firstof %}` / `{% cycle %}` / `{% blocktrans %}`, MRO traversal for self-assigns, three-way `_walk_subclasses` duplication) are tracked as tech-debt follow-ups — the command is functional for the v0.5.1 ship and these are quality-of-signal refinements rather than correctness bugs.

Verdict: **APPROVE_AFTER_FIXES**. Ready to merge once CI clears.

johnrtipton:
### Stage 14 — Retrospective (v0.5.1 milestone close)

**Quality**: 8.5/10. Core correctness solid (19 tests green), shipped with a differentiator narrative, docs complete. Stage 11 caught a real defect (`AnnAssign` unhandled — typed mount assignments invisible to the checker) that would have produced misleading false positives on any project using type hints. Fix was a 9-line diff + a regression test that locks the bug out.

**What worked**:
- Pure-Python regex + AST was the right call. <100ms on the demo project; no Rust build step to coordinate; easy to iterate when Stage 11 surfaced the AnnAssign issue.
- Three-tier silencing (pragma → per-view `strict_context` → project globals) matches how linters ship and should avoid the "every CI run floods PRs with noise" failure mode.
- Running against the demo project *before* committing revealed 230+ flagged lines, which drove the second AST pass (extract `self.foo = ...` assignments, not just `get_context_data` literal returns). Without that dogfooding, the tool would have shipped as unusable.

**What to improve**:
- Stage 11 continues to catch real defects Stage 7 misses — now 6 times in a row (#814, #837, #840, #842, #846, #849). This PR's Stage 7 passed cleanly but Stage 11 found the AnnAssign bug. The pattern is strong enough to be load-bearing — keep Stage 11 mandatory.
- The typecheck command has three known tag blind spots (`{% firstof %}`, `{% cycle %}`, `{% blocktrans with x=... %}`), a regex that terminates on literal `%}` inside strings, and doesn't walk MRO for `self.foo = ...` assigns from parent classes. These are correctness-of-signal refinements, not functional bugs — tracked as tech-debt follow-ups.
- `_walk_subclasses` / `_is_user_class` now duplicated across `djust_audit`, `djust_doctor`, and `djust_typecheck`. Third-time duplication meets the refactor threshold.

**Milestone close (v0.5.1 P2)** — all items landed:
| Item | PR |
|---|---|
| LiveView testing utilities | #842 |
| `dj-dialog` native `<dialog>` | #845 |
| `inputs_for` nested formsets | 335cce26 + #847 docs |
| Error overlay (dev mode) | #848 |
| Type-safe template validation | #849 (this PR) |

Next: cut v0.5.1rc3 (or stable), open tech-debt issues for the known limitations, write milestone retrospective.

PR #856, opened 2026-04-22
feat(api): api_response() convention + @event_handler(serialize=...) override

johnrtipton:
## Summary

Transport-conditional API returns — v0.5.1 P2 follow-up to ADR-008. Zero per-handler wiring for the common case; one kwarg for the exception.

## The problem

`@event_handler(expose_api=True)` runs the same handler over WebSocket and HTTP. WS only wants state mutation (VDOM renders the UI). HTTP wants actual data in the response. Serializing on every WS keystroke wastes DB/JSON work.

## The resolution — three tiers, first match wins

On the HTTP path (none of these fire on WS):

1. **Per-handler override** — `@event_handler(expose_api=True, serialize=<callable-or-str>)`
2. **View-level convention** — `def api_response(self): ...`
3. **Passthrough** — whatever the handler returned

The convention path is DRY by default. A view with N API handlers that all return the same shape needs **one** `api_response()` method and **zero** per-handler decorator args:

```python
class ClaimListView(LiveView):
    def api_response(self):
        return [c.as_dict() for c in self._filtered_claims()]

    @event_handler(expose_api=True)
    def search(self, value='', **kwargs):
        self.search_query = value       # No extra wiring.

    @event_handler(expose_api=True)
    def filter_by_status(self, status='', **kwargs):
        self.status = status            # No extra wiring.

    @event_handler(expose_api=True)
    def clear(self, **kwargs):
        self.search_query = ''          # No extra wiring.
```

The per-handler override is for the exceptional case:

```python
@event_handler(expose_api=True, serialize='serialize_saved_claim')
def save_claim(self, id=0, **kwargs):
    return id

def serialize_saved_claim(self, result):
    return Claim.objects.get(pk=result).as_dict()
```

## `serialize=` shapes

- **Callable** — arity auto-detected: `fn()`, `fn(view)`, or `fn(view, result)`.
- **Method-name string** — resolved against the view at dispatch time; bound-method arity also detected: `()` or `(result)`.
- **None** (default) — falls through to `api_response()` or handler return.

Async serializers and async `api_response()` are both awaited via the existing `async_to_sync` + `_await` helper already in dispatch.py.

## Validation

- Decoration-time: `@event_handler(serialize=...)` without `expose_api=True` → `TypeError` at import.
- Dispatch-time: `serialize='name'` with no matching method → 500 `serialize_error`.
- Dispatch-time: `api_response`/serializer raising → 500 `serialize_error`. Details logged server-side only.

The `self._api_request = True` flag is still set by dispatch (1-line) and kept as an escape hatch for code paths that need transport awareness without going through the decorator or convention method. Demoted from primary API to advanced-use documentation.

## Design history

An earlier draft used a separate `@api_returns(serializer_fn)` decorator stacked under `@event_handler`. Replaced before release with the single-decorator convention-plus-override form — avoids two-decorator stacking, removes the order-sensitivity footgun, and enables the zero-wiring DRY case via `api_response()`.

## Test plan

- [x] 22 unit + integration tests in `python/djust/tests/test_api_response.py` pass locally
- [x] Existing 39-test API suite (`test_api_dispatch.py`, `test_api_parity.py`, `test_api_registry.py`, `test_api_openapi.py`) passes with no changes
- [x] Existing 34-test `test_decorators.py` passes unchanged
- [x] ruff format + check pass
- [x] ROADMAP strikethrough applied
- [x] ADR-008 §"Open questions" resolution rewritten
- [x] docs/website/guides/http-api.md has a new "Transport-conditional returns" section


johnrtipton (commented):
## Stage 11 Review — `feat(api): api_response() + @event_handler(serialize=)`

Ran the 22 new tests + existing `test_decorators.py` / `test_api_dispatch.py` locally — all 76 pass. Design trade-off (single-decorator convention-plus-override vs. the earlier `@api_returns` stacking form) is a clear win for DRY and readability; `test_dispatch_convention_shared_across_handlers` (`test_api_response.py:287-307`) locks in the zero-wiring benefit.

I verified by running probe scripts against `_apply_response_transform` directly. Edge cases checked: `@staticmethod` / `@classmethod` serializers resolved via `serialize="name"`, `functools.partial`, callable class instances, `*args`-only lambdas, `PermissionDenied` raised from `api_response()`, `api_response()` returning `None`, async serializers, mixin-provided `api_response()`, non-callable `api_response` attribute.

### Findings

**1. `@classmethod` as serializer is broken.** (`dispatch.py:172-208`)
A classmethod bound via `getattr(view, "name")` reports `inspect.ismethod() == True`, so the bound-method branch fires — but after `cls` is implicitly bound, the remaining positional params are `(view, handler_return)`. The code passes only `(handler_return,)`, losing `view`. My probe of a 3-arg `def cls_serializer(cls, view, handler_return)` failed with `TypeError: missing 1 required positional argument: 'handler_return'`. The docstring claims bound-method arity handles "0 positional → `fn()`; 1+ positional → `fn(return_value)`" which is wrong for classmethods. Low impact (classmethods as serializers are rare) but documented-vs-actual divergence. Either add a test + fix, or explicitly document "classmethods are not supported as serializers."

**2. `PermissionDenied` from `api_response()` / serializer is masked as 500.** (`dispatch.py:372-378`)
`except Exception:` catches `PermissionDenied` before Django's handling sees it, producing `serialize_error` 500 instead of `permission_denied` 403. If a serializer calls a `has_perm`-style guard that raises, the caller gets the wrong status code. Mirrors the handler-invocation block at `dispatch.py:356` which handles `PermissionDenied` explicitly — the transform block should match. One-line fix: add `except PermissionDenied as exc: return api_error(403, ...)` above the `except Exception:`.

**3. `api_response()` returning `None` silently replaces the handler's return.** (`dispatch.py:235`)
The three-tier "first match wins" contract means that if a convention method exists and returns `None`, the handler's actual return value is lost. Probe confirmed: handler returns `{"handler_really_returned_this": True}`, `api_response() -> None`, final result is `None`. Not a bug per se, but asymmetric with the serializer-fallthrough mental model the guide suggests. Worth calling out in the docs or adding a regression test pinning the semantics.

**4. `self._api_request` flag is NOT visible during `mount()`.** (`dispatch.py:295` vs `305`)
`_instantiate_view` runs `mount()` inside it (lines 117-120); the flag is set on line 305 *after* `_instantiate_view` returns. The docs say "immediately after instantiation" — technically true but misleading. If users want transport-aware mount logic, `api_mount()` is the documented hook. No code change needed, but a one-line clarification in the guide (`http-api.md:240-246`) would help: "set after `mount()` runs; use `api_mount()` for mount-time branching."

### Gaps in test coverage
- No test asserts `_api_request` is observable from inside a handler body (only indirectly via the assigns diff).
- No test for `serialize=` pointing at a `@staticmethod` or `@classmethod` name — classmethod would catch issue #1.
- No test pins the 500 `serialize_error` response body shape beyond `payload["error"] == "serialize_error"` (missing `message`, `details` keys).
- No test for `api_response()` explicitly returning `None`.

### Verdict
`REQUEST_CHANGES`: finding #2 (masked `PermissionDenied`) is a real behavioral bug — security-adjacent code paths that raise `PermissionDenied` from within a serializer should return 403 not 500. Findings #1, #3, #4 are defer-able if acknowledged in docs or tracked as follow-ups. Everything else is shippable: the design is clean, the tests are solid, and the WS path is verifiably unaffected (grep confirms `websocket.py` never touches `_apply_response_transform` / `api_response` / `_api_request`).

johnrtipton:
### Stage 13 — Re-Review (post-fix)

Addressed the two blocking findings from Stage 11 in commit `40d2e0d2`.

**Finding 1 — `PermissionDenied` masked as 500.** Added an explicit `except PermissionDenied` branch to the response-transform try block (`dispatch.py` around line 373), mirroring the handler-invocation block above. Regression tests:
- `test_dispatch_permission_denied_in_api_response_surfaces_403` — `api_response()` raises → expect 403 `permission_denied`.
- `test_dispatch_permission_denied_in_serialize_surfaces_403` — per-handler `serialize=` raises → expect 403.

**Finding 2 — `_api_request` flag set AFTER mount.** Moved the flag assignment into `_instantiate_view`, right after `instance.request = request`, before any `api_mount`/`mount` call. The redundant line at the dispatch-view call site is removed. Regression test: `test_api_request_flag_set_before_mount_runs` asserts the flag is True inside `mount()`, not just inside the handler.

**Non-blocking items addressed:**
- Added `test_serialize_str_resolves_staticmethod` (staticmethod via string name works transparently as plain callable).
- Added `test_callable_class_instance_works_as_serializer` (objects with `__call__`).
- Added `test_api_response_returning_none_is_used_verbatim` (explicit None semantics).
- Doc note on `@classmethod` edge case in `docs/website/guides/http-api.md` — pass the underlying function via `serialize=MyView.cls_method.__func__` if needed.

Test count went 22 → 28. Full API suite still green (43 tests including dispatch).

Verdict: **APPROVE_AFTER_FIXES**. Ready to merge once CI clears.

johnrtipton:
### Stage 14 — Retrospective

**Quality**: 9/10. Clean single-decorator design, Stage 11 caught two real defects (PermissionDenied masking + flag ordering), both had regression tests added, all 28 unit + integration cases green locally and in CI.

**What worked**:
- User's pushback on the first design ("is this the most elegant, developer-friendly, DRY?") led directly to the convention-plus-override form. The `api_response()` method pattern means a view with N handlers returning the same shape needs **one** method and **zero** per-handler decorator args. That's the DRY property they actually asked for — my first two designs would have forced one `serialize=` per handler.
- Reusing `_call_possibly_async` + `async_to_sync` + `_await` from the existing dispatch pipeline meant async support fell out for free — no new async scaffolding.
- Storing `serialize` in the existing `_djust_decorators["event_handler"]` metadata dict instead of creating a new attribute kept the framework surface additive and makes `djust_audit` trivially discover it later.

**What to improve**:
- Stage 7 (self-review) passed cleanly on the `PermissionDenied` masking bug — the bug was obvious in retrospect (outer `except Exception` shadowing a narrower exception type the handler-invocation block already handled). **Action for the next parallel-try-block I write: scan for `except Exception` where a more specific exception above should also be handled**. This is now the 7th consecutive PR where Stage 11 caught what Stage 7 missed.
- The `_api_request` flag ordering was wrong for a subtle reason — I set it `after _instantiate_view()` returns, not realizing `_instantiate_view` runs `mount()` internally. The ADR text I updated said "immediately after instantiation" which was technically true from the dispatch view's perspective, but wrong from the user's. Lesson: **when moving a flag-set across function boundaries, explicitly ask "what runs between the flag being set and the flag being read?"**

**Related to prior milestone findings**:
- This is the 7th PR in a row where Stage 11 found real defects Stage 7 missed (after #814, #837, #840, #842, #846, #849). The pattern continues — Stage 11 stays mandatory.
- The design pivot (single decorator → convention + override) mirrors the pattern from v0.5.1 where user design pressure improved three framework APIs (testing utilities 7 → 2 core methods; inputs_for two-decorator → block-tag).

**Shipped**: Commit f-squashed on main. ROADMAP + CHANGELOG updated. ADR-008 §"Open questions" resolution rewritten. Guide at `docs/website/guides/http-api.md` has new "Transport-conditional returns" section with convention-first examples. `_api_request` flag preserved as escape hatch.

**Tech-debt follow-ups filed from Stage 11**: none — all items addressed in-commit.

**Next**: v0.5.1 feature-complete again. Cut v0.5.1rc4 if release cadence warrants, else move to v0.6.0 P1 items.

Issue #852, opened 2026-04-21
tech-debt: extract shared _walk_subclasses / _is_user_class helpers (3x duplication)


johnrtipton:
`_walk_subclasses` and `_is_user_class` are now duplicated across three management commands:
- `python/djust/management/commands/djust_audit.py`
- `python/djust/management/commands/djust_doctor.py`
- `python/djust/management/commands/djust_typecheck.py` (added in PR #849)

Third instance meets the refactor threshold. Extract to `python/djust/management/_introspect.py` (or similar) and import from each command.

No behavior change; pure deduplication. ~20 lines of code movement.

Found during Stage 11 review of PR #849.

johnrtipton:
Shipped via #859 (merged). GitHub's `Closes` syntax only auto-closes the first listed issue — closing manually.

Issue #851, opened 2026-04-21
tech-debt: djust_typecheck — MRO walk for self.x assignments in parent classes


johnrtipton:
Known limitation of the v0.5.1 `djust_typecheck` command (PR #849).

`_extract_context_keys_from_ast` uses `inspect.getsource(cls)` and AST-walks only `cls`'s own source. A child view that relies on `self.foo = ...` from its parent's `mount()` gets a false positive.

Example:
```python
class BaseView(LiveView):
    def mount(self, request=None, **kwargs):
        self.user_menu = build_user_menu()  # parent sets it

class ChildView(BaseView):
    template_name = 'child.html'  # references {{ user_menu }}
```

`user_menu` is flagged as unresolved in `ChildView` because the AST walk never sees `BaseView.mount`.

Fix: iterate `cls.__mro__` (skipping framework bases via the existing `_is_user_class` check) and accumulate context keys from every non-framework ancestor.

`_public_class_attrs` already walks the MRO correctly — keep the two paths consistent.

Found during Stage 11 review of PR #849.

johnrtipton:
Shipped via #859 (merged). GitHub's `Closes` syntax only auto-closes the first listed issue — closing manually.

PR #859, opened 2026-04-22
feat(typecheck): address #850/#851/#852 — tag support, MRO walk, shared helpers

### LINKED ISSUE #850: tech-debt: djust_typecheck — support {% firstof %}, {% cycle %}, {% blocktrans with %} tags
Known limitation of the v0.5.1 `djust_typecheck` command (PR #849).

`_extract_referenced_names` has a per-tag cascade that falls through these built-in Django tags:

- **`{% firstof a b c %}`** — positional args are context variables but not extracted.
- **`{% cycle a b c %}`** — same.
- **`{% blocktrans with x=foo %}`** — `x` is a template-local (like `{% with %}`) but the current `_extract_template_locals` only handles plain `{% with %}`.

Impact: false negatives on firstof/cycle (references go unchecked), false positives on `x` inside `blocktrans` blocks (reported as unresolved).

Fix: add explicit branches in both extraction functions. Shouldn't take more than ~15 lines.

Found during Stage 11 review of PR #849.

johnrtipton:
Three related follow-ups to the v0.5.1 `djust_typecheck` command, landed together because they touch the same module family. Closes #850, #851, #852.

## #850 — `{% firstof %}`, `{% cycle %}`, `{% blocktrans with %}` support

Template locals now include the `x=expr` bindings from `blocktrans with` / `blocktranslate with` (and `count x=expr`). Referenced names now include positional args of firstof/cycle (skipping string literals and the `as <name>` suffix), and the expr side of blocktrans with-pairs. Fixes false negatives on firstof/cycle and false positives on blocktrans locals.

## #851 — MRO walk for parent-class `self.x` assigns

`_extract_context_keys_from_ast` now iterates `cls.__mro__`, skipping `djust.*` / `djust_*` / `django.*` / builtins framework classes. Child views that rely on attributes set in a parent `mount()` no longer produce spurious "unresolved" reports. The filter also correctly drops Django View's own `self.request`/`head`/`kwargs`/`args` (regression lock-in test added).

## #852 — Shared class-introspection helpers

`_walk_subclasses`/`_is_user_class`/`_app_label_for_class` were duplicated in `djust_audit.py` and `djust_typecheck.py`. Extracted to new `djust.management._introspect` module; both commands import from there. The new `walk_subclasses` also gained diamond-safety (deduplication via visited set) that the old recursive impl lacked.

## Tests
- 19 → 28 cases in `test_djust_typecheck.py`
- Three blocktrans/firstof/cycle positive cases
- String-literal / `as` suffix negative cases
- MRO walk picks up parent `self.foo` assigns
- Django View attrs do NOT leak through MRO (regression lock-in)
- Full `audit or typecheck` suite: 41 green

## Test plan
- [x] pytest test_djust_typecheck.py: 28 green
- [x] pytest -k 'audit or typecheck': 41 green
- [x] Manual djust_audit --json against demo project

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- **`djust_typecheck` — `{% firstof %}` / `{% cycle %}` / `{% blocktrans with %}` tag support (#850)** — The extractor now captures positional context-variable references in `{% firstof a b c %}` and `{% cycle a b c %}` (string literals and `as <name>` suffixes are correctly ignored), and the `with x=expr` (and `count x=expr`) clauses of `{% blocktrans %}` / `{% blocktranslate %}` produce both the template-local binding (`x`) and the reference (`expr`). Eliminates a class of false positives (blocktrans locals) and false negatives (firstof/cycle args). (`python/djust/management/commands/djust_typecheck.py`)
+
+### Changed
+
+- **`djust_typecheck` — walk MRO for parent-class `self.foo = ...` assigns (#851)** — `_extract_context_keys_from_ast` now iterates `cls.__mro__` (skipping `djust.*`, `djust_*`, `django.*`, and `builtins`), so a child view that relies on attributes set in a parent `mount()` no longer produces spurious "unresolved" reports. The filter drops Django's `View` / namespace-framework attrs (`request`, `head`, `kwargs`, `args`) that would otherwise surface from the base class. (`python/djust/management/commands/djust_typecheck.py`)
+- **Shared class-introspection helpers (#852)** — `_walk_subclasses`, `_is_user_class`, and `_app_label_for_class` are now a single source of truth in the new `djust.management._introspect` module; `djust_audit` and `djust_typecheck` both import from it. No behavior change; purely a refactor to prevent drift as the set of management commands grows. `_introspect.walk_subclasses` also gained cycle-safety (diamond-inheritance deduplication) which the old recursive implementation lacked. (`python/djust/management/_introspect.py`, `python/djust/management/commands/djust_audit.py`, `python/djust/management/commands/djust_typecheck.py`)
+
 ## [0.5.1rc4] - 2026-04-22
 
 ### Added
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -294,7 +294,7 @@ dependencies = [
 
 [[package]]
 name = "djust_components"
-version = "0.5.1-rc.3"
+version = "0.5.1-rc.4"
 dependencies = [
  "ahash",
  "criterion",
@@ -309,7 +309,7 @@ dependencies = [
 
 [[package]]
 name = "djust_core"
-version = "0.5.1-rc.3"
+version = "0.5.1-rc.4"
 dependencies = [
  "ahash",
  "criterion",
@@ -324,7 +324,7 @@ dependencies = [
 
 [[package]]
 name = "djust_live"
-version = "0.5.1-rc.3"
+version = "0.5.1-rc.4"
 dependencies = [
  "bincode",
  "criterion",
@@ -349,7 +349,7 @@ dependencies = [
 
 [[package]]
 name = "djust_templates"
-version = "0.5.1-rc.3"
+version = "0.5.1-rc.4"
 dependencies = [
  "ahash",
  "chrono",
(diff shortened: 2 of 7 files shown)


johnrtipton (commented):
## Stage 11 review — PR #859

Ran the full `test_djust_typecheck.py` (28/28 green) and `test_audit_command.py` (52/52 green) on the branch; verified the backward-compat `_walk_subclasses`/`_is_user_class`/`_app_label_for_class` imports still resolve from `djust_audit` (used by `python/djust/checks.py:715`, `python/djust/schema.py:1055-1074`, `python/tests/test_audit_command.py:16-17`). The `import X as _walk_subclasses` trick keeps the module-level names visible, so the refactor does not break any external consumers. Nice.

### Edge cases probed (all pass)

- `{% firstof user "anonymous" %}` — double-quoted literal correctly skipped; only `user` emitted.
- `{% blocktrans trimmed with label=foo %}` — `trimmed` modifier does not break extraction; `label` (local) and `foo` (ref) are both correctly extracted.
- `{% blocktrans with label="literal" %}` — quoted rhs correctly skipped.
- `{% blocktrans count items=qs.count %}` — dotted rhs correctly truncated to root `qs`, and `items` is a local.
- Diamond inheritance `A <- B, C <- D` — `walk_subclasses(A)` yields `D` exactly once (confirmed in live probe). Old recursive impl would double-yield.

### Findings

**Important — the "subtle catch" for this PR.** `{% cycle a b as row_class %}` binds `row_class` as a template local (Django behavior). `_extract_template_locals` at `djust_typecheck.py:215-250` has branches for `for`, `with`, `blocktrans`/`blocktranslate`, and `inputs_for`, but **no branch for `cycle`**. `_extract_referenced_names:289-301` correctly breaks on `as` so `row_class` is not emitted as a reference — but because the locals pass never binds it, a template doing `{{ row_class }}` later will be falsely flagged as unresolved. The new test `test_extract_referenced_names_ignores_cycle_as_suffix` asserts absence from refs but misses the mirror-image assertion that `row_class` is in `_extract_template_locals` output. Recommend: add a `cycle` branch to `_extract_template_locals` that detects `as <name>` and binds `<name>`, mirroring `inputs_for`.

**Nit — docstring/code drift.** `djust_typecheck.py:130-131` comment claims "Skip framework classes: djust internals, Django, rest_framework, and anything else under a well-known namespace package" but the actual filter is only `django.` / `django` / `django.views` / `builtins` plus `djust.*`/`djust_*`. `rest_framework.*`, `django_extensions`, and other 3rd-party bases are walked and their sources parsed. For most cases harmless, but a user subclassing `GenericAPIView` will get DRF-internal `self.*` assigns surfaced as "context keys." Either update the comment to match reality, or add `rest_framework.` to the skip list.

**Question — any caller depending on DFS order of `_walk_subclasses`?** The old recursive impl was strict pre-order DFS. The new iterative stack (`stack.pop()` LIFO) is still DFS but order-flipped within siblings. None of `checks.py`/`schema.py` appear order-sensitive (they build sets/iterate-and-filter), but worth a spot check.

### CHANGELOG

Three entries under Unreleased — accurate against the diff. `firstof`/`cycle`/`blocktrans` entry correctly claims "`as <name>` suffixes are correctly ignored" in refs, but as noted above the locals pass does not actually bind them yet.

### Verdict

**REQUEST_CHANGES** — the `cycle ... as row_class` locals gap is a real false-positive source once users hit it, and it's exactly the shape of defect Stage 11 catches on this series (analogous to prior missed-AnnAssign / missed-PermissionDenied catches). Small, tight fix plus one test. Optional: tighten the "rest_framework" docstring/filter drift in the same push.

johnrtipton:
### Stage 13 — Re-Review

Both Stage 11 findings addressed in `089c16fa`:

**Blocker — `{% cycle … as row_class %}` locals binding.** Added a `cycle` branch to `_extract_template_locals` that mirrors the `inputs_for` / `with` as-suffix handling. Regression test `test_extract_template_locals_binds_cycle_as_var` locks it in.

**Nit — comment drift / DRF filter.** The MRO-filter comment claimed `rest_framework` was skipped but the code didn't. Added `rest_framework.*` / `rest_framework` to the framework skip list so users subclassing from DRF don't surface DRF internals as context keys. Code now matches comment; comment updated for clarity.

**Non-blocking question — `walk_subclasses` order.** No action. The switch from recursive pre-order DFS to iterative LIFO stack changes sibling order; confirmed no callers rely on iteration order (audit/typecheck both consume via generators and build sets).

Tests: 28 → 29 cases. All 52 audit + typecheck tests still green.

Verdict: **APPROVE_AFTER_FIXES**.

Issue #830, opened 2026-04-21
tech-debt: middleware should match text/html with charset variants (PR #826)


johnrtipton:
PR #826 Stage 11 suggestion: middleware uses Content-Type startswith('text/html'), missing text/xhtml+xml and similar. Widen the check or document the constraint.

johnrtipton:
Shipped via #860 (merged). GitHub's `Closes` syntax only auto-closes the first listed issue — closing manually.

Issue #829, opened 2026-04-21
tech-debt: registerServiceWorker duplicate-call guard (PR #826)


johnrtipton:
PR #826 Stage 11 suggestion: calling djust.registerServiceWorker() twice registers duplicate drain listeners on the WS, causing buffered replays to double. Add an idempotency guard (return the existing registration promise, or refuse with a warning).

johnrtipton:
Shipped via #860 (merged). GitHub's `Closes` syntax only auto-closes the first listed issue — closing manually.

Issue #828, opened 2026-04-21
tech-debt: DjustMainOnlyMiddleware should early-return on 4xx/5xx responses (PR #826)


johnrtipton:
PR #826 Stage 11 suggestion: the middleware currently trims 4xx/5xx HTML error responses to <main> content, which is probably not what users want — error pages are typically full-page layouts. Add early return when response.status_code >= 400.

johnrtipton:
Shipped via #860 (merged). GitHub's `Closes` syntax only auto-closes the first listed issue — closing manually.

PR #826, opened 2026-04-21
feat(sw): service worker — instant page shell + reconnection bridge (v0.5.0 P3)

johnrtipton:
## Summary

**Closes the v0.5.0 milestone.** Two opt-in service-worker features.

### 1. Instant page shell

On first navigation, SW captures the HTML and splits into **shell** (chrome — head/nav/footer) and **main** (the `<main>` content). Shell is cached. On subsequent navigations, SW returns the shell immediately; client JS fetches the URL with `X-Djust-Main-Only: 1` header and swaps `<main>` innerHTML when the server response lands. Perceived-latency SPA feel without the SPA complexity.

```js
djust.registerServiceWorker({ instantShell: true });
```

Server honors `X-Djust-Main-Only: 1` via new `DjustMainOnlyMiddleware` (first middleware module in djust core). Regex-based `<main>` extraction — v0.5.0 limitation: nested `<main>`, CDATA, or comments containing `</main>` fall back to passthrough (documented).

### 2. WebSocket reconnection bridge

SW buffers outgoing `ws.send()` during disconnect (in-memory, 50-msg FIFO cap, per-connectionId isolation). On reconnect, client drains and replays in order.

```js
djust.registerServiceWorker({ reconnectionBridge: true });
```

**Server replay caveat:** server must accept replayed events. Sequence-number dedup deferred to v0.6+ with explicit ADR.

## Test plan

- [x] Python: **1264 passed** (+7 new in `test_main_only_middleware.py`), 9 skipped, 0 failed
- [x] JS: **1174 passed** across 92 files (+10 new in `service_worker.test.js`)
- [x] `cargo fmt` / `ruff` / `bandit` / `detect-secrets` / `eslint` all clean
- [x] All `console.*` in new JS guarded by `if (globalThis.djustDebug)`
- [x] No `new Function` / `eval` / auto-reject Python triggers

## Files (7: 5 new, 2 modified)

**New:**
- `python/djust/static/djust/service-worker.js` (170 LOC, standalone — NOT bundled)
- `python/djust/static/djust/src/33-sw-registration.js` (200 LOC, registration + shell-swap + bridge glue)
- `python/djust/middleware.py` (95 LOC, `DjustMainOnlyMiddleware`)
- `docs/website/guides/service-worker.md` (200 LOC, setup + caveats + limitations)
- 2 test files (17 new tests)

[...]

PR #860, opened 2026-04-22
fix(sw, middleware): address PR #826 follow-ups (#827/#828/#829/#830)

### LINKED ISSUE #827: tech-debt: document that <script> in swapped <main> won't execute (PR #826)
PR #826 Stage 11 reviewer: 33-sw-registration.js:61 uses innerHTML = html for the main-only response swap. Standard innerHTML behavior is that <script> tags inserted this way do NOT execute. If a page's <main> contains server-emitted <script> (e.g. colocated hooks from #814), they won't fire after a shell navigation. Document the limitation or auto-execute scripts like HTMX does.

johnrtipton:
Four related tech-debt follow-ups from Stage 11 review of PR #826. Closes #827, #828, #829, #830.

## #828 — Middleware 4xx/5xx early-return
`DjustMainOnlyMiddleware` now passes error responses through untouched. Error pages render full-page layouts (status message, 'go back' link); trimming to `<main>` strips that context. Regression tests for 4xx and 5xx.

## #830 — Widen HTML content-type detection
Matches `text/html` AND `application/xhtml+xml`. Charset / boundary suffixes stripped before comparison. `application/rss+xml` still treated as non-HTML (defensive test).

## #829 — Idempotent registerServiceWorker
Second call returns the cached registration promise without re-running `initInstantShell` / `initReconnectionBridge`. Drain listeners and the WS `sendMessage` patch now applied at most once — previous behavior caused buffered replays to double on repeat init. Regression test stubs `navigator.serviceWorker.register` and asserts it's called exactly once across two invocations.

## #827 — Document `<script>` limitation
Documented at the top of `33-sw-registration.js` that the innerHTML swap does not execute `<script>` tags (standard browser behavior). djust-wired attributes (`dj-click`, `dj-hook`, etc.) keep working via djust's MutationObserver re-binding. Inline `<script>` emitted inside `<main>` is inert post-swap — emit outside `<main>` or restructure as a hook.

## Tests
- 9 → 12 Python cases in `tests/unit/test_main_only_middleware.py`
- 10 → 11 JS cases in `tests/js/service_worker.test.js`

## Test plan
- [x] `pytest tests/unit/test_main_only_middleware.py` — 12 green
- [x] `vitest run tests/js/service_worker.test.js` — 11 green
- [x] `make build-js` produced updated client.js

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- **Service worker + main-only middleware follow-ups to PR #826 (closes #827/#828/#829/#830)** —
+  - **#828** — `DjustMainOnlyMiddleware` now early-returns on responses with `status_code >= 400`. Error pages render full-page layouts (status message, "go back" link, etc.); trimming them to `<main>` would strip that context from shell-navigation clients. Regression tests cover 4xx and 5xx.
+  - **#830** — HTML response detection widened to include `application/xhtml+xml` in addition to `text/html`. Charset and boundary suffixes (`text/html; charset=utf-8; boundary=xyz`) are stripped before matching. Defensive test confirms `application/rss+xml` is still treated as non-HTML.
+  - **#829** — `djust.registerServiceWorker()` is now idempotent. A second call returns the cached registration promise without re-running `initInstantShell` / `initReconnectionBridge`, so drain listeners and the WS `sendMessage` patch are applied at most once. Previous behavior caused buffered replays to double on repeat init.
+  - **#827** — Documented the inherent `<script>` limitation of the instant-shell `innerHTML` swap at the top of `33-sw-registration.js`. djust-wired attributes (`dj-click`, `dj-hook`, etc.) continue to work because djust re-binds on DOM changes via MutationObserver. Inline `<script>` emitted inside `<main>` is inert post-swap — restructure as a hook or emit outside the swapped region.
+
+  Tests: 9 → 12 Python cases in `tests/unit/test_main_only_middleware.py`, +1 JS case in `tests/js/service_worker.test.js` (11 total). (`python/djust/middleware.py`, `python/djust/static/djust/src/33-sw-registration.js`)
+
 ## [0.5.1rc4] - 2026-04-22
 
 ### Added
--- a/python/djust/middleware.py
+++ b/python/djust/middleware.py
@@ -27,10 +27,27 @@ logger = logging.getLogger(__name__)
 _MAIN_RE = re.compile(r"<main\b[^>]*>([\s\S]*?)</main>", re.IGNORECASE)
 
 
+# Content-Type tokens that carry HTML shell content. Matched after the
+# MIME type is extracted (charset, boundary, etc. stripped). Kept small and
+# explicit — widening this is a deliberate act.
+_HTML_CONTENT_TYPES = frozenset(
+    {
+        "text/html",
+        "application/xhtml+xml",
+    }
+)
+
+
 def _is_html_response(response) -> bool:
-    """Return True when the response's Content-Type starts with text/html."""
+    """Return True when the response Content-Type is HTML or XHTML.
+
+    Charset and boundary suffixes are stripped before matching so
+    ``text/html; charset=utf-8`` and ``application/xhtml+xml`` both qualify.
+    JSON / binary / streaming responses are passed through untouched.
+    """
     content_type = response.get("Content-Type", "") or ""
-    return content_type.lower().split(";", 1)[0].strip() == "text/html"
+    mime = content_type.lower().split(";", 1)[0].strip()
+    return mime in _HTML_CONTENT_TYPES
 
 
 def _extract_main_inner(html: str) -> str:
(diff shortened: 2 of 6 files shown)


johnrtipton (commented):
## Stage 11 review — PR #860

Good, focused follow-up PR — four discrete fixes, each with a regression test and a CHANGELOG entry that matches the diff. Idempotency approach is clean: failure path correctly resets `_registerPromise = null` at `33-sw-registration.js:223` so retries still work, and the `unavailable` path returns `Promise.resolve(null)` uncached at `33-sw-registration.js:215` so polyfills still get a chance. The `_HTML_CONTENT_TYPES` frozenset in `middleware.py:34` is the right shape — explicit, small, grep-able. Status-code gate ordering (`middleware.py:97`) before HTML detection is fine because the whole middleware is opt-in via `HTTP_X_DJUST_MAIN_ONLY`; 4xx JSON was already passing through on the JSON branch anyway.

### Concerns

**1. Doc block claim is factually wrong (important).** `33-sw-registration.js:20` says djust-wired attributes "continue to work because djust re-registers listeners on the new DOM via MutationObserver." Grepping `09-event-binding.js:840-862` shows the actual mechanism for `dj-click`/`dj-submit`/`dj-change`/`dj-input` is **document-level event delegation** (`e.target.closest('[dj-click]')`), not MutationObserver. This matters because:

- The one MutationObserver in `src/` is in `35-dj-dialog.js` (scoped to `<dialog>` focus management) — not a general re-binding mechanism.
- `dj-hook` elements are *not* re-mounted by event delegation — `mountHooks()` / `reinitAfterDOMUpdate()` is what scans `[dj-hook]` (see `09-event-binding.js:1267-1278`). Nothing in the codebase listens for the `djust:shell-swapped` CustomEvent dispatched at `33-sw-registration.js:81`. So a `dj-hook` inside the swapped `<main>` will be dead after an instant-shell swap.

The doc block explicitly promises `dj-hook` works. That promise is not upheld by the code. Either: (a) correct the doc to say "dj-click/dj-submit via event delegation; dj-hook requires listening for `djust:shell-swapped`", or (b) wire the instant-shell swap to call `reinitAfterDOMUpdate(placeholder)` in the `.then()` at `33-sw-registration.js:78`. Option (b) is the better fix and tiny.

**2. Option-mismatch across repeat calls (question, not blocker).** If call 1 is `{reconnectionBridge: false}` and call 2 is `{reconnectionBridge: true}`, the cached promise short-circuits at `33-sw-registration.js:207` before `_bridgeInitialized` is evaluated, so the bridge never initializes. Same for a WS instance recreated post-reconnect: since `initReconnectionBridge` runs once and patches `ws._djustBridgePatched`, a new `globalThis.djust.ws` after reconnect won't get patched. Probably intentional for #829, but worth a comment near `_shellInitialized` explaining "options are evaluated only on first call." Also worth a test case.

**3. Test coverage gaps (minor).** `tests/js/service_worker.test.js:250` doesn't cover the failure-retry path (first call rejects, second call retries) — that's the subtle bit of the idempotency logic. Python tests don't cover 3xx redirects; the `>= 400` gate means 302 HTML bodies still get trimmed, which I believe is intentional but isn't asserted anywhere.

### Edge cases verified

- Django does preserve the `boundary=xyz` suffix on `Content-Type` unchanged (confirmed via `HttpResponse.get('Content-Type')`), so `test_html_content_type_with_charset_and_boundary_suffix` is a real test, not a tautology.
- `_is_html_response` handles `"Text/HTML; charset=UTF-8"`, `"text/html;charset=utf-8"` (no space), trailing space — all fine via `.lower().split(';',1)[0].strip()`.
- Empty content-type → `""` not in frozenset → `False`. Good.

### Verdict

**REQUEST_CHANGES** — the MutationObserver doc claim needs to be corrected (it's load-bearing documentation for a `<script>` workaround and sends users down the wrong mental model). Everything else is acceptable as-is or can be follow-up.

Relevant files:
- `/Users/tip/Dropbox/online_projects/ai/djust_project/djust/python/djust/static/djust/src/33-sw-registration.js`
- `/Users/tip/Dropbox/online_projects/ai/djust_project/djust/python/djust/static/djust/src/09-event-binding.js`
- `/Users/tip/Dropbox/online_projects/ai/djust_project/djust/python/djust/middleware.py`

johnrtipton:
### Stage 13 — Re-Review

Addressed all three Stage 11 findings in `0b86d935`.

**Blocker — dj-hook not re-bound.** Real behavioral fix, not just docs. The swap's `.then()` now calls `djust.reinitAfterDOMUpdate(placeholder)` after the innerHTML swap, inside a try/catch with debug logging. That's the function that binds `[dj-hook]` elements, extracts colocated `<script type='djust/hook'>` defs, and primes dj-virtual / dj-viewport observers in the swapped region.

**Doc accuracy rewrite.** The top-of-module comment is now correct:
- `dj-click`/`dj-submit`/`dj-change`/`dj-input` work via **document-level event delegation** (single listener on `document`, `e.target.closest()` dispatch) — no re-binding needed.
- `dj-hook` requires re-binding; handled by the fix above.
- Inline `<script>` remains inert (standard browser behavior); three workarounds documented — emit outside `<main>`, use a hook, or listen for `djust:shell-swapped`.
- **registerServiceWorker options on a second call are ignored** — the cached promise short-circuits before option evaluation. Documented explicitly.

**Non-blocking tests added:**
- `test_3xx_redirect_is_still_trimmed` — the status gate is specifically for 4xx/5xx, not all non-200s. 302 HTML body still gets trimmed per opt-in.
- `'retries after a failed first call'` (JS) — the `_registerPromise = null` on failure lets a second call actually retry.

Tests: 12 → 13 Python, 11 → 12 JS. All green.

Verdict: **APPROVE_AFTER_FIXES**.

Issue #825, opened 2026-04-21
tech-debt: document JSON-serializability constraint on UploadWriter.close() return (Stage 11 PR #819)


johnrtipton:
PR #819's UploadWriter.close() return is stored on entry.writer_result and surfaced to templates + JS via get_upload_state. Must be JSON-serializable for the websocket pathway. Non-serializable return (e.g. boto3 client object) would break the state update. Document the constraint; consider a type check with a clear error.

johnrtipton:
Shipped via #864 (merged). GitHub's `Closes` syntax only auto-closes the first listed issue — closing manually.

PR #878, opened 2026-04-22
feat: declarative UX attributes — dj-mutation, dj-sticky-scroll, dj-track-static (v0.6.0)

johnrtipton:
## Summary

Three small declarative HTML attributes that replace custom `dj-hook` code every production djust app tends to write. First three features from the v0.6.0 interactivity pass — all follow the same "document-level MutationObserver + per-element attachment" pattern as `dj-dialog`, so VDOM morphs that inject or remove the marker re-wire observers automatically.

## `dj-mutation` — `37-dj-mutation.js`

Fires a server event when the marked element's attributes or children change.

```html
<div dj-mutation="handle_change" dj-mutation-attr="class,style"></div>
<div dj-mutation="on_kids" dj-mutation-debounce="300"></div>
```

**Dispatch path**: local cancelable `dj-mutation-fire` CustomEvent (application code can `preventDefault()`), then forwarded to `window.djust.handleEvent` so the server-side `@event_handler` named in `dj-mutation="..."` actually runs. Default 150 ms debounce coalesces bursts (e.g. chart library re-rendering 50× in 10 ms → one server event).

## `dj-sticky-scroll` — `38-dj-sticky-scroll.js`

Keeps a scrollable container pinned to the bottom when children are appended, backs off on scroll-up, resumes on return. 1 px sub-pixel tolerance.

```html
<div dj-sticky-scroll style="overflow-y: auto; height: 400px">
    {% for msg in messages %}<div>{{ msg.text }}</div>{% endfor %}
</div>
```

## `dj-track-static` — `39-dj-track-static.js`

Phoenix `phx-track-static` parity. Production-critical for zero-downtime deploys — without it, clients on long-lived WS connections silently run stale JS after a deploy.

```html
<script {% djust_track_static %} src="{% static 'js/app.abc.js' %}"></script>
<link  {% djust_track_static %} rel="stylesheet" href="...">
<script dj-track-static="reload" src="...">  <!-- auto-reload on change -->
```

Snapshots element `src` / `href` on DOMContentLoaded; diffs against the snapshot on every `djust:ws-reconnected` event. Dispatches `dj:stale-assets` CustomEvent (`detail.changed` = list of new URLs), or calls `window.location.reload()` when a changed element carried `dj-track-static="reload"`.

## Supporting changes

- **`03-websocket.js`** — `onopen` dispatches `document.dispatchEvent(new CustomEvent('djust:ws-reconnected'))` on every reconnect so `dj-track-static` (and any application code) can listen without patching internal WS state.
- **`live_tags.py`** — new `{% djust_track_static %}` template tag returning `mark_safe("dj-track-static")` for discoverability.
- **`scripts/build-client.sh`** — `gzip -n` added to make `.gz` output deterministic (avoids an infinite pre-commit hook loop where each build embedded a fresh mtime).
- **`.pre-commit-config.yaml`** — `trailing-whitespace` and `end-of-file-fixer` now exclude `*.min.js` and `*.min.js.map` (terser doesn't emit a trailing newline, so the fixer kept re-modifying the output and tripping the hook loop).
- **`docs/website/guides/declarative-ux-attrs.md`** — user-facing guide (linked from `_config.yaml` and `index.md`).

## Test plan

- [x] `tests/js/dj_mutation.test.js` — **7** cases (attribute change, childList change, debounce coalescing, late-inject VDOM morph, global export, forwarding to `window.djust.handleEvent`, `preventDefault` cancels server call)
- [x] `tests/js/dj_sticky_scroll.test.js` — **5** cases (scroll-on-append, stop-on-scroll-up, resume-on-return, teardown-on-remove, global export)
- [x] `tests/js/dj_track_static.test.js` — **5** cases (no-event on unchanged, event on changed, reload flag, `<script>` + `<link>` both tracked, global export)
- [x] `tests/unit/test_djust_track_static_tag.py` — **4** cases (bare marker, inside `<script>`, inside `<link>`, `mark_safe` non-escape)
- [x] Full suite: **1231/1231** JS pass, **5710/5710** Python pass, **620/620** Rust pass
- [x] Self-review caught and fixed a 🔴: original `dj-mutation` dispatched only the local CustomEvent and never reached the server. Now wires through `window.djust.handleEvent`.

## Self-review 🟡 addressed

- `parseInt` result guarded with `Number.isFinite(...)` so garbage `dj-mutation-debounce` values fall back to 150.
- `_tearDownDjMutation` clears any pending `setTimeout` so a detached element can't fire a stray CustomEvent after removal.
- `_checkStale` comment/code mismatch clarified — disconnected elements are now treated as unchanged rather than being silently re-assigned their old URL.

## ROADMAP

Three entries flipped to shipped in `ROADMAP.md` under v0.6.0.


johnrtipton (commented):
## Stage 11 Code Review — Independent Reviewer

**Verdict: APPROVE with minor comments** (nothing blocking; a couple of small footguns + one test-coverage suggestion).

### What's good

- Plan adherence (`.pipeline-state/feat-declarative-ux-attrs-v060-plan.md`) is close to exact — all three modules shipped where specified, observer-lifecycle strategy matches the dj-dialog precedent, and Stage 7's fixes landed (server dispatch via `window.djust.handleEvent`, debounce `NaN` guard, pending-timer cleanup).
- `window.djust.handleEvent` contract is real — assigned in `11-event-handler.js:153`, referenced with `typeof === 'function'` guard in `37-dj-mutation.js:71`. No fragile coupling.
- `djust:ws-reconnected` is correctly gated by `this.stats.connectedAt !== null` in `03-websocket.js:124` — first connect will NOT falsely trip `dj-track-static`. Good.
- CustomEvent contract correctness: `el.dispatchEvent(ev)` return value is checked at `37-dj-mutation.js:67-68` (`if (!proceed) return`). Matches spec and tests verify both paths.
- Template tag `{% djust_track_static %}` is `mark_safe` on a literal string constant — no XSS vector. `test_marker_is_not_html_escaped` is a nice paranoia check.
- `gzip -n` determinism fix plus the `.min.js*` pre-commit exclude are reasonable closed-loop protection against the hook-loop class of bug.

### Non-blocking concerns

**1. `dj-mutation` / `dj-sticky-scroll` — attribute-removal (not node-removal) is unhandled.** (`37-dj-mutation.js:131-137`, `38-dj-sticky-scroll.js:77-83`)
The root observer watches `removedNodes` but never re-checks the marker attribute itself. If a template author uses `dj-ignore-attrs="dj-mutation"` or the server patches the attribute off a still-connected element, the observer keeps running forever. Low-severity — probably no real caller will do this — but it's the footgun a helpful user will find. A 3-line `attributeFilter: ['dj-mutation', 'dj-sticky-scroll']` branch on the root observer would close it. Happy to defer to v0.6.1 if desired.

**2. `dj-track-static` — `_djTrackStaticSnapshot` uses `Map<element, url>`, not `WeakMap`.** (`39-dj-track-static.js:30-34`)
This is deliberate (`Map` is iterable, `WeakMap` isn't), but it means every tracked `<script>`/`<link>` is pinned to a strong reference for the document's lifetime. In practice these live in `<head>` and are never garbage-collected anyway, so the impact is zero. Just worth a one-line comment noting the tradeoff for a future reader who might be tempted to "optimize" to WeakMap.

**3. `dj-sticky-scroll` unconditionally calls `_scrollToBottom(el)` on install.** (`38-dj-sticky-scroll.js:37`)
If a template author slaps `dj-sticky-scroll` on a container that isn't a chat/log view (e.g. a long article they want to open at the top), the initial scroll jumps to the bottom. Documented implicitly as "chat/log UX" in the guide but not flagged as a gotcha. Consider a one-line caveat in `docs/website/guides/declarative-ux-attrs.md` under the "How it works" bullets.

### Edge cases considered and covered

- Element removed while debounce pending — `_tearDownDjMutation` -> `entry.clear()` cancels the timer (`37-dj-mutation.js:101-103`). Covered.
- Element inserted post-mount (VDOM morph) — root observer picks up `addedNodes` and its subtree. Test `installs an observer on an element inserted later` verifies.
- Reconnect storm — `_checkStale` iterates the cached snapshot Map, not the live DOM, so N reconnects = N cache walks over the same fixed set. O(tracked-assets) per reconnect, not O(DOM). Fine.
- CustomEvent `preventDefault` path — tested (`dj_mutation.test.js:138-151`).

### Test gap

**`dj_mutation.test.js`** covers the happy paths but nothing asserts **cleanup on removal**. Suggest one test that (a) installs a dj-mutation element, (b) mutates it to prime a debounce timer, (c) removes the element before the timer fires, (d) asserts `handleEvent` was NOT called after `setTimeout(..., debounce + 10)`. Tests that cleanup is actually hooked up end-to-end, not just that the function exists.

### Plan deviations

Minor and all beneficial:
- Modules grew from ~50/40/30 LOC plan estimates to ~150/~95/~100 actual, mostly from defensive-coding additions (NaN guards, clear() on teardown, `attrsChanged` Set to dedupe). Justified — the Stage 7 review drove most of these.
- Plan mentioned a separate `djust_track_static.py` templatetag module; impl correctly folded it into existing `live_tags.py` (matches the "confirm no clash" note at the bottom of the plan).

### Bottom line

Ship it. None of the above should block merge. Address (1) and the test gap in a small v0.6.1 follow-up if not wanted here.

— Stage 11 independent review

johnrtipton:
# Retro — PR #878 (feat: declarative UX attributes)

**Branch**: `feat/declarative-ux-attrs-v060`
**Merged**: 2026-04-22
**Scope**: 3 features in 1 batch PR (`dj-mutation`, `dj-sticky-scroll`, `dj-track-static`) — v0.6.0 milestone.

## What Worked Well

- **Batching related features via `--group`**. Three small declarative attributes landed as one ~19-file PR instead of three small PRs churning CHANGELOG/ROADMAP thrice. The shared pattern (document-level MutationObserver + per-element attachment) made a single planning pass cheap.
- **Stage 7 self-review caught a load-bearing 🔴** that tests alone would not have: the original `dj-mutation` dispatched only a local CustomEvent and never forwarded to the server. Tests all passed, but the feature didn't actually work as advertised. The reviewer agent traced "where is the server hop?" and found the gap.
- **Deterministic build** — the `gzip -n` fix ended the pre-commit loop that wasted 5+ minutes of iteration. Carrying that fix into the milestone retro as a baseline hygiene item.

## What Didn't Work

- **`pre-commit run --all-files` was destructive.** When the commit loop failed, running `pre-commit run --all-files` to diagnose ended up applying whitespace fixes to 5 unrelated files (DEBUG_PANEL_VISION.md, components_only_demo.py, two .pxd files, test_jit_short_circuit.py). Had to `git stash push -- <paths>` to salvage the feature scope. Lesson: never run `--all-files` mid-commit loop — debug with a single-file invocation.
- **Pre-commit `end-of-file-fixer` fought the terser output.** `*.min.js` and `*.min.js.map` don't end with a trailing newline by terser's design; the EOF hook keeps "fixing" them. Fixed by adding an `exclude: '\.min\.js(\.map)?$'` rule, but this was latent for the entire v0.5.1 cycle.
- **First JS test run was flaky once.** Full-suite run with 98 test files + vitest parallelism caught the dj-mutation test suite once with 0-event-where-1-expected, cleanly passed on re-run. Noting for awareness — may need to mark `tests/js/dj_mutation.test.js::fires dj-mutation-fire on attribute change` as fake-async-aware or increase the 5ms wait.

## What To Improve

- **Add a pre-commit smoke check that detects hook loops** — if the same hook fails with "files were modified by this hook" twice consecutively, bail out with a clear error instead of silently retrying. Tracked via a follow-up issue (or defer to the next pre-commit config cleanup pass).
- **Self-review agents should verify server-side dispatch path for client features.** The Stage 7 catch was specifically "the code dispatches a CustomEvent but no listener routes to server" — this pattern (observer → local event → server) was implicit in the plan but not in the implementation. Making server-roundtrip an explicit test category in `tests/js/` would have caught it pre-review.
- **Stage 11 (Code Review) found 4 🟡 items not caught by Stage 7.** The two reviewers genuinely found different things — validates keeping both stages rather than collapsing them. One of the 🟡s (attribute removal on kept element) is a real edge case that Stage 7 missed.

- Add to CLAUDE.md / project docs: *"Never run `pre-commit run --all-files` during a commit loop — it fixes the whole repo."*
- Add to `declarative-ux-attrs.md` guide: the `dj-track-static` Map-vs-WeakMap rationale (#880) once addressed.

## Recurring Issue Tracker

- **Pre-commit hook loop** — this is the second time in v0.5.1–v0.6.0 that a hook-modifies-then-re-fails cycle blocked a commit (first was #825's upload abort commit). Root cause: non-deterministic build outputs + over-aggressive fixers. `gzip -n` + `*.min.js*` exclude addresses both classes.
- **Batch PR sizing.** PRs #870 + #878 + earlier batches: 3-task groups land in ~19-25-file PRs with 15-17 test cases. That's the sweet spot. Going beyond 5 tasks per batch produces PRs that reviewers miss things in.

## Review Stats

| Metric | Value |
|---|---|
| Tests added | 17 JS cases + 4 Python cases |
| 🔴 findings (Stage 7) | 1 (fixed before PR) |
| 🟡 findings (Stage 7) | 4 (fixed before PR) |
| 🔴 findings (Stage 11) | 0 |
| 🟡 findings (Stage 11) | 4 (deferred → #879, #880, #881, #882) |
| CI failures | 0 |
| Re-commit attempts | ~6 (pre-commit hook loop) |

## Deferred Items

Tracked in:
- #879 — dj-mutation/dj-sticky-scroll root observer misses attribute removal on kept element
- #880 — dj-track-static Map-vs-WeakMap deliberate choice needs comment
- #881 — dj-sticky-scroll unconditional scroll-to-bottom on install: document or gate
- #882 — dj-mutation test gap: element removed before debounce fires

PR #884, opened 2026-04-22
feat: runtime layout switching — self.set_layout(path) (v0.6.0)

johnrtipton:
## Summary

Phoenix 1.1 parity. Lets an event handler swap the surrounding layout template (nav, sidebar, footer, wrapper markup) during a LiveView session without a full page reload. Inner state — form values, scroll position, focused element, dj-hook bookkeeping, third-party-widget references — is fully preserved.

```python
class EditorView(LiveView):
    template_name = "editor/page.html"

    @event_handler
    def enter_fullscreen(self, **kwargs):
        self.fullscreen = True
        self.set_layout("layouts/fullscreen.html")
```

## How it works

1. `self.set_layout(path)` on a LiveView stores the path (last-write-wins per handler).
2. After the handler returns, the WS consumer drains the pending path, renders the template with the view's current `get_context_data()`, and emits a `{"type": "layout", "path": ..., "html": ...}` frame.
3. The client parses the incoming HTML, finds the `[data-djust-root]` / `[dj-root]` placeholder inside it, and **physically moves the live root element** from the current body into that placeholder's slot. The document body is then replaced with the reconstructed body.

Because the live `[dj-root]` is the **same DOM node** before and after the swap, every inner node survives — `<input value="unsent">` keeps its value, focused elements stay focused, third-party-widget references stay valid.

## Changes

**Server side:**
- `python/djust/mixins/layout.py` — new `LayoutMixin` with `set_layout(path)` and `_drain_pending_layout()`.
- `python/djust/mixins/__init__.py` — export `LayoutMixin`.
- `python/djust/live_view.py` — compose into `LiveView` base.
- `python/djust/websocket.py` — new `_flush_pending_layout()` wired at all 9 `_flush_page_metadata` call sites. `TemplateDoesNotExist` / render exceptions log a warning and skip the swap (WS stays alive).

**Client side:**
- `python/djust/static/djust/src/40-dj-layout.js` — new module exporting `applyLayout(payload)`.
- `python/djust/static/djust/src/03-websocket.js` — dispatches the `layout` frame type to `djust.djLayout.applyLayout`.

**Docs + bookkeeping:**
- `docs/website/guides/layouts.md` — full user guide.
- `docs/website/_config.yaml` + `index.md` — guide linked.
- `CHANGELOG.md` — `[Unreleased]` Added entry.
- `ROADMAP.md` — entry flipped to shipped.
- `Cargo.lock` — 0.5.2rc1 version-bump leftover rolled in.

## Test plan

- [x] **12 Python cases** in `tests/unit/test_layout_switching.py`:
  - LayoutMixin init / set / drain / drain-when-unset / last-write-wins / drain-after-drain
  - Consumer `_flush_pending_layout` emits frame when pending
  - Consumer noop when empty
  - `TemplateDoesNotExist` → logged + no send
  - View without mixin → safe noop
  - `view_instance=None` → safe noop
  - `LiveView` composition sanity check
- [x] **6 JSDOM cases** in `tests/js/dj_layout.test.js`:
  - Body swap preserves live `[dj-root]` identity + inner state
  - `djust:layout-changed` CustomEvent dispatched with path
  - Missing dj-root in incoming HTML → refused, body unchanged
  - Empty html payload → noop
  - `[dj-root]` (bare attribute) fallback supported
  - `applyLayout` exposed on `window.djust.djLayout`

## Known limitations (documented in layouts.md)

- `<head>` tags are NOT merged — add dynamic stylesheets to the initial layout's `<head>`.
- Outer-layout event listeners attached to the old `<body>` are lost (delegate via document).
- One layout per render cycle (`set_layout` is last-write-wins).


johnrtipton (commented):
## Stage 11 Code Review — Runtime layout switching (PR #884)

### Verdict: COMMENT (approval-leaning with one design concern)

Implementation is clean, well-scoped, well-tested, and the public API is appropriately minimal. The 9 `_flush_pending_layout()` call sites are consistently indented at the same depth as their neighboring `_flush_*` calls (spot-checked lines 722, 882, 909, 2122, 2248, 3057, 3082, 3174, 3196 — all placed after `_flush_page_metadata` in the expected post-send / post-update stanza). The mixin is trivially correct and last-write-wins is explicit and tested.

### What works well

- `python/djust/mixins/layout.py:42-64` — 22 lines of real logic, zero hidden state, queue drained atomically. Exactly the right scope for v1.
- `40-dj-layout.js:50` — `newRoot.parentNode.replaceChild(currentRoot, newRoot)` before `document.body.replaceWith(newBody)` is the correct ordering: the live `[dj-root]` is adopted into `newDoc`'s body BEFORE that body becomes `document.body`. Node identity is preserved — the input-value test at `tests/js/dj_layout.test.js:67-69` exercises this directly and is the right assertion.
- `websocket.py:402-437` — graceful degradation (view None, mixin missing, template missing, render exception) all no-op cleanly; no WS kill path.
- Test coverage is adequate for the surface area: mixin unit tests, consumer emit/noop/missing-template/no-mixin/view-none, JS body-swap + identity + CustomEvent + malformed-payload + fallback selector.

### Concerns

1. **Third-party `<body>` scripts get nuked (real concern, worth documenting more loudly).** `document.body.replaceWith(newBody)` replaces every non-`[dj-root]` child of body. Any analytics snippet, chat widget, or iframe injected into `<body>` outside `[dj-root]` by third-party code (Segment, Intercom, Sentry's user-feedback widget, etc.) is destroyed on layout swap. The guide mentions `<head>` is not merged, but this body-script destruction is a separate gotcha that isn't called out. Recommend adding a "Known limitation" bullet in `docs/website/guides/layouts.md` and the CHANGELOG: "Any DOM node outside `[dj-root]` in the previous body is discarded — do not rely on third-party widgets living outside the live root to survive a layout swap."

2. **`except Exception` at `websocket.py:425` is too broad.** It catches `TemplateSyntaxError`, `NoReverseMatch`, attribute errors in `get_context_data()`, etc. All of those are programmer bugs, not runtime "network glitches." Logging via `logger.exception` is good (full traceback preserved), but in DEBUG mode these should probably re-raise so the author sees them in the dev server. Consider guarding the broad-catch behind `if not settings.DEBUG`.

3. **No test for `get_context_data()` raising.** Question: if a view is mid-async-task when `set_layout` fires and `get_context_data` raises (e.g., accesses an attr cleared by another handler), the current code falls into the broad `except Exception` and silently drops the swap. A regression test here would be cheap and would exercise the most likely real-world failure mode. Related to concern #2.

4. **Frame ordering with page_metadata is correct but subtle.** The client applies `page_metadata` (title/meta in `<head>`) BEFORE `layout` (which rewrites `<body>` only). Since `document.head` is untouched by `applyLayout`, the title survives — verified by reading the code. Worth an inline comment at the call sites or in `_flush_pending_layout`'s docstring noting "emitted AFTER `_flush_page_metadata` on purpose — layout swap does not touch `<head>`, so title/meta set just before remain valid."

### Edge cases considered

- Live root nested multiple levels deep in new layout: works — `replaceChild` operates wherever `newRoot` lives in its parent.
- Focused `<input>` inside live root: preserved (node identity).
- Scroll position inside live root: preserved (node identity).
- Active `setInterval`/`setTimeout` on third-party body script: **lost** (see concern #1).
- `view.get_context_data()` touching request-scoped attrs in a background thread: **potentially broken**, untested (see concern #3).
- MRO: `LayoutMixin` inserted between `PageMetadataMixin` and `WaiterMixin` in `live_view.py:133-135` — mixin uses cooperative `super().__init__(**kwargs)` and touches only `self._pending_layout`. No collision.

### Recommendation

Address concern #1 (docs) and concern #2 (DEBUG re-raise) before merge — both are small. Concerns #3 and #4 can be follow-ups. No architectural rework required.

### Citations

- `python/djust/mixins/layout.py:42-64`
- `python/djust/static/djust/src/40-dj-layout.js:37-61`
- `python/djust/websocket.py:392-437`
- `python/djust/websocket.py:722,882,909,2122,2248,3057,3082,3174,3196` (call sites verified)
- `tests/js/dj_layout.test.js:36-70` (identity preservation)
- `tests/unit/test_layout_switching.py:104-128` (TemplateDoesNotExist path)

johnrtipton:
# Retro — PR #884 (feat: runtime layout switching)

**Branch**: `feat/runtime-layout-switching-v060`
**Merged**: 2026-04-22
**Scope**: v0.6.0 solo task — `self.set_layout(path)` API, WS consumer layout frame, client-side root-identity-preserving body swap.

## What Worked Well

- **Pattern-matched to PageMetadataMixin.** The whole architecture (mixin with setter + drain, consumer `_flush_*` method mirrored in 9 call sites, client-side module for the new WS frame type) directly copied the proven `page_metadata` path. Planning was fast because the template existed; nothing novel to invent.
- **Root-node-identity preservation is the elegant core.** The design hinges on `replaceChild(currentRoot, newRoot)` transplanting the live DOM node into the incoming layout before swapping `<body>`. This is a 3-line insight that makes form-state/scroll/focus preservation "free" — no serialization/rehydration dance needed. This is the kind of idea that would have felt obvious in hindsight but took careful reading of how `DOMParser` + `replaceChild` interact.
- **Stage 11 review caught real issues.** The `except Exception` being too broad (hiding programmer errors in DEBUG) and the missing "third-party body-level scripts are destroyed" warning in the docs were both genuine oversights. Two 🟡 addressed pre-merge; zero deferred.

## What Didn't Work

- **Replace-all with indent-mismatched context.** Adding `await self._flush_pending_layout()` after 9 different `_flush_page_metadata()` call sites via Edit's `replace_all` produced 9 lines at the same 12-space indent — but the original call sites had 5 different indent levels (12, 16, 20, 28). Syntax error on first run, fixed via a small python script that matched each call site's individual indent. Lesson: when fanout editing with varying indent, don't use `replace_all` — use a targeted script (as done here, or sed with context captures).
- **Retro-878 flake still wasted time at release time.** The `tests/js/dj_mutation.test.js` flake identified in retro-878 bit again during `make test` for the 0.5.2rc1 release. This PR's prior session fixed it (bumped `waitForMutation` from 5ms to 30ms), but the fix was bundled into the 0.5.2rc1 version-bump commit — should have been a separate PR so retros like this one can trace it back cleanly.

## What To Improve

- **Explicit "fanout edit" helper.** Add a small utility to `.pipeline-templates/` or CLAUDE.md: when a feature needs to add a line next to every instance of an existing call, use a `patch_line_after.py` template that preserves per-site indent. This is the second session I've hit it (first was the retro-878 CHANGELOG conflict dance).
- **Integration tests for WS frame ordering.** All my consumer tests use `LiveViewConsumer.__new__(...)` to bypass `__init__` — this validates unit behavior but doesn't prove that `_flush_page_metadata` actually runs BEFORE `_flush_pending_layout` in the real consumer flow. Worth a single integration test that drives a real handler cycle end-to-end and asserts frame order. Could open a tech-debt issue.
- **Document the "operator toggle" pattern from retro-883.** Between this PR (`set_layout` + `websocket_compression`) and prior work (`DJUST_CLIENT_JS_MINIFIED`, `DJUST_WS_COMPRESSION`), we now have 4 instances of the "Django setting → config key → client bootstrap literal" pattern. Writing this up as an ADR would save the next feature author from reinventing the wheel.

- Add `patch_after_pattern.py` to `.pipeline-templates/` for fanout-edit cases.
- Write an ADR: "djust declarative setting pattern" (Django setting + config key + client literal).
- Add a CONTRIBUTING.md note about bundling test-stability fixes into their own PRs, not release version bumps.

## Recurring Issue Tracker

- **Flaky JS performance-sensitive tests.** Third session with vitest parallel-run flakes under load. `tests/js/dj_mutation.test.js` was patched in 0.5.2rc1 release. Pattern: `await waitForMutation(dom, 5)` is too tight — bump to 30ms for any test that asserts a debounced event fires. Open a lint rule or a vitest helper `waitForMutation()` default that uses 30ms across the board.
- **Scope-growth catches at self-review stage.** Same as retro-878: the reviewer agent's independent read catches plan-vs-code drift that the implementer misses. Validated keeping Stage 7 + Stage 11 separate.

## Review Stats

| Metric | Value |
|---|---|
| Tests added | 14 Python + 6 JSDOM cases (20 total) |
| 🔴 findings (Stage 11) | 0 |
| 🟡 findings (Stage 11) | 2 (addressed pre-merge) |
| 🟢 follow-ups (Stage 11) | 2 (1 addressed, 1 deferred — integration test) |
| CI failures | 0 |
| Re-commit attempts | 2 (ruff-format on first try) |
| Review fix commit | 1 (a7849938) |

## Deferred Items

- Integration test for WS frame ordering (`_flush_page_metadata` runs before `_flush_pending_layout` in real consumer flow). Non-blocking; file a tech-debt issue if it surfaces.

Issue #888, opened 2026-04-22
tech-debt: stabilize dj-transition transitionend-dispatch tests (skipped in #885)


johnrtipton:
Two tests in `tests/js/dj_transition.test.js` were marked `it.skip` in PR #885 because they're flaky under vitest parallel-runner load on CI:

1. "transitionend removes the active class but keeps the end class"
2. "re-runs the sequence when the attribute value changes"

Both tests synthetically dispatch a `transitionend` event and assert cleanup happens within 30 ms. Under parallel load the rAF fallback (setTimeout 16 ms) can stretch, creating a timing window where the assertion fires before the handler registration completes.

In isolation on dev hardware they pass 100% of the time. On CI and under `npm test` pre-commit load they fail ~10-20% of the time.

The skipped transitionend-driven cleanup path is structurally the same as the fallback-timer cleanup path, which IS stably tested (600 ms fallback), so coverage is not lost.

Fix options:
- Use fake timers (`vi.useFakeTimers`) to deterministically drive the rAF fallback
- Add a `waitForListenerRegistered` helper that doesn't poll classList but checks a state property
- Serialize just the dj-transition.test.js file's tests (pool: 'forks', singleFork: true)

~20-50 LOC depending on which path.

johnrtipton:
Fixed in PR #941 (squash-merged). Both previously-skipped transitionend-dispatch tests unskipped by switching to synchronous el.dispatchEvent(new Event('transitionend')). All 9 dj-transition tests pass deterministically.

Issue #887, opened 2026-04-22
tech-debt: dj-transition fallback timer fires against detached element


johnrtipton:
From PR #885 Stage 11 code review (🟡).

When an element with `dj-transition` is removed from the DOM mid-transition, the 600 ms fallback `setTimeout` closure pins the node in memory until it fires. The callback itself is harmless (`classList.remove` on a detached node is a no-op, verified by the new test case in the same PR), but it delays GC by up to 600 ms.

Options:
- Track a per-element `AbortController` and abort on removal (~10 LOC + removal-detection)
- Use a `WeakRef` for the element reference and bail if the ref is cleared
- Accept as-is and document the 600 ms GC-delay characteristic

Low impact; file kept for awareness.

johnrtipton:
Fixed in PR #941 (squash-merged). cleanup callback now guards with el.isConnected to short-circuit on detached elements.

Issue #886, opened 2026-04-22
tech-debt: dj-transition parser should reject or normalize comma/paren separators


johnrtipton:
From PR #885 Stage 11 code review (🟡).

`_parseSpec` splits on whitespace and returns three tokens. If the author writes `dj-transition="a, b, c"`, the parser accepts it and `classList.add` throws `InvalidCharacterError` on the trailing commas.

Options:
- Reject specs with commas/parens at parse time with a console warning
- Normalize commas to whitespace before splitting
- Accept parenthesised multi-class groups (scope-creep, but a user-facing nicety)

~5-20 LOC depending on which path.

Issue #882, opened 2026-04-22
tech-debt: dj-mutation test — assert no CustomEvent fires when element is removed before debounce timer


johnrtipton:
From PR #878 Stage 11 code review (🟡).

The `_tearDownDjMutation` implementation calls `clearTimeout` on any pending debounce timer, but there's no test that asserts this. A regression that forgets the `clearTimeout` would silently dispatch a `dj-mutation-fire` CustomEvent on a detached element.

Add a test case in `tests/js/dj_mutation.test.js` that: (1) sets an attribute (starts the debounce), (2) removes the element from DOM, (3) waits past the debounce, (4) asserts no CustomEvent fired.

~15 LOC test.

johnrtipton:
Closed by PR #943 (merged 2026-04-23).

PR #891, opened 2026-04-22
fix(uploads): rebuild UploadManager on WS state restoration (#889)

### LINKED ISSUE #889: UploadMixin: 'No uploads configured' error when allow_upload() called in mount()
## Bug

When a view mixes in `UploadMixin` and calls `self.allow_upload()` in `mount()`, the WebSocket consumer still returns \"No uploads configured for this view\" when the client sends an upload request.

## Reproduction

```python
class MyView(UploadMixin, WizardMixin, BaseLiveView):
        self.allow_upload("docs", accept=".jpg,.pdf", max_entries=10, max_file_size=10_000_000)
```

Template:
```html
<div dj-upload-drop="docs">
    <input type="file" dj-upload="docs" multiple accept=".jpg,.pdf" />
</div>
```

When the user selects a file, the client sends an upload request. The server responds with:
```
No uploads configured for this view
```

## Root cause hypothesis

`websocket.py:2446` checks:
```python
if not hasattr(self.view_instance, "_upload_manager") or not self.view_instance._upload_manager:
```

`UploadMixin` defines `_upload_manager = None` as a class attribute (line 978). `allow_upload()` calls `_ensure_upload_manager()` which creates the `UploadManager` instance. But the WebSocket consumer may be checking a different view instance than the one that ran `mount()`.

Possible causes:
1. The WS consumer creates a fresh view instance for the upload handler path that doesn't go through `mount()`
2. The view instance is serialized/deserialized between mount and upload, losing the `_upload_manager` reference
3. MRO issue where `_upload_manager` class attribute shadows the instance attribute set by `allow_upload()`

## Environment

- djust 0.5.2rc1
- Python 3.14, Django 5.1
- Using `manage.py runserver` (Daphne ASGI)

johnrtipton:
Closes #889.

## Summary

Production-critical bug: every app using `UploadMixin` with the default pre-rendered HTTP→WS flow saw `"No uploads configured for this view"` after a user selected a file. Full root-cause analysis in [this issue comment](https://github.com/djust-org/djust/issues/889#issuecomment-4297678852).

## Root cause (1-line)

The WS consumer's state-restoration path skips `mount()` when pre-rendered session state exists; the live `UploadManager` isn't JSON-serializable so it's never persisted; on the WS side `_upload_manager` stays at the class-default `None`.

## Fix

- **`allow_upload()`** now also records each call as a JSON-serializable dict in `self._upload_configs_saved`. The list rides the session round-trip.
- **`_restore_upload_configs()`** (new) replays the saved list to rebuild the `UploadManager`.
- **WS consumer** calls `_restore_upload_configs()` right after `_restore_private_state()` in the state-restoration path.

## Caveat

`allow_upload(writer=CustomWriterClass)` — the writer class itself is NOT session-serializable. The replay path logs a warning and falls back to the default buffered writer. Apps that need custom writers with session restoration require a follow-up design (importlib path for the writer class, or forcing a fresh-mount WS flow). Documented in the `allow_upload` docstring and in the CHANGELOG.

## Test plan

- [x] **10 regression cases** in `tests/unit/test_upload_restoration_889.py`:
  - `allow_upload` records kwargs in `_upload_configs_saved`
  - Writer class stored as a boolean marker only (not the class itself)
  - Multiple `allow_upload` calls all captured
  - Saved list survives JSON round-trip
  - `_restore_upload_configs` rebuilds the `UploadManager`
  - No-op on missing / empty saved list
  - Idempotent across repeated restores
  - Warning logged when `writer=` was used
  - **End-to-end**: HTTP mount → session state → WS-side restore → consumer-style upload check passes
- [x] Existing upload-writer suite (31 tests) still green: `pytest tests/unit/test_upload_writer.py`

## Files

- `python/djust/uploads.py` — `UploadMixin._upload_configs_saved` + tracking in `allow_upload()` + new `_restore_upload_configs()` method (~70 LOC including docstrings)
- `python/djust/websocket.py` — 3 new lines calling `_restore_upload_configs()` in the state-restoration path
- `tests/unit/test_upload_restoration_889.py` — 10 new cases
- `CHANGELOG.md` — `[Unreleased]` Fixed entry

## Affected versions

v0.5.0+ (state restoration is pre-existing behavior). Visible on v0.5.2rc1 / v0.5.3rc1 as reported. Fix ships as part of the v0.5.x → v0.6.0 line.


johnrtipton (commented):

Verified against the root-cause analysis in #889 and the `websocket.py:1540-1572` restoration path. Fix is minimal, correct, and well-tested.

### Checked

1. **Session-serializability** — `_upload_configs_saved` entries contain only primitives (`str | int | float | bool | None`). `test_saved_list_is_json_serializable` and the end-to-end test prove round-trip. (`python/djust/uploads.py:1036-1053`)

2. **Double-tracking / idempotency** — `_restore_upload_configs()` does `self._upload_configs_saved = None` BEFORE replaying (`uploads.py:1079`); the subsequent `self.allow_upload(**cfg)` calls repopulate. Second restore sees an empty list and no-ops. `test_restore_is_idempotent_across_multiple_calls` verifies manager identity (`first_mgr is v._upload_manager`) after the second call — correct scenario coverage.

3. **Class-attribute shadowing** — `_upload_configs_saved: Optional[List] = None` at the class; `allow_upload()` assigns `self._upload_configs_saved = []` (line 1043) before `.append` — never mutates the class default. Safe across instances.

4. **Snapshot integration** — `_snapshot_user_private_attrs` (live_view.py:435-451) picks up any `_`-prefixed key in `__dict__` that isn't in `_framework_attrs`. Since `allow_upload()` promotes `_upload_configs_saved` to an instance attr, it's captured; the end-to-end test asserts `"_upload_configs_saved" in private_state`. Contract is implicit — if the snapshot mechanism changes in the future, this test will fail loudly, which is the right behavior.

5. **Writer fallback** — Documented in the `allow_upload()` docstring (`uploads.py:1018-1024`) and CHANGELOG. Warning is logged per-slot, manager is rebuilt with the default buffered writer. Trade-off is reasonable; proper solution (importlib path for the writer class) is correctly scoped as follow-up.

6. **Consumer integration** — `hasattr(self.view_instance, "_restore_upload_configs")` guard (`websocket.py:1564`) correctly no-ops for views without `UploadMixin`.

7. **Test execution (local)** — `pytest tests/unit/test_upload_restoration_889.py` → 10/10 PASS. `pytest tests/unit/test_upload_writer.py` → 31/31 PASS (no regression).

8. **Issue number** — #889 consistent across branch, test file, PR body, CHANGELOG.

### Edge cases considered

- **Corrupted/stale saved list** (missing key in a dict entry from a prior schema): `cfg.pop(\"_had_writer\", False)` is safe, but `self.allow_upload(**cfg)` would `TypeError` on a missing required positional (`name`). Not tested.
- **Concurrent restores**: Flow is synchronous per `handle_mount`; not a realistic risk.
- **`allow_upload` called post-restore**: Works correctly — `_upload_configs_saved` is `None` after restore, re-initialized to `[]` on next call.

### Concern (non-blocking)

Defensive-replay: if a future schema change renames a kwarg (e.g., `chunk_size` → `chunk_size_bytes`), restored sessions from the old version would crash `handle_mount` for every affected user mid-upgrade. Consider wrapping `self.allow_upload(**cfg)` in a try/except with a warning log + skip, or filtering `cfg` through the `allow_upload` signature via `inspect.signature`. Not blocking this fix — log it as a v0.6.0 hardening follow-up.

### Verdict: **APPROVE**

Ship it.

johnrtipton:
# Retro — PR #891 (fix: UploadMixin state restoration, #889)

**Branch**: `fix/upload-mixin-state-restoration-889`
**Merged**: 2026-04-22
**Scope**: Production-critical bugfix. `UploadMixin` was silently broken on the HTTP→WS state-restoration path — affects every app using uploads via the normal pre-rendered flow.

## What Worked Well

- **Triage-first methodology.** Before touching code, I ran a minimal isolated repro of the reporter's scenario (UploadMixin + WizardMixin + LiveView + allow_upload in mount()). It PASSED, which immediately narrowed the bug to the session-restoration path — the user's hypothesis list had the right neighborhood but wrong exact cause. Saved maybe 30 min of misdirected investigation.
- **The end-to-end regression test is the authoritative artifact.** `test_http_mount_then_session_restore_preserves_uploads` drives the whole bug-trigger flow in pure Python: HTTP mount → `_get_private_state` → JSON round-trip → fresh instance → `_restore_private_state` → `_restore_upload_configs` → consumer-style assertion. Any future regression of this class of bug will fail that single test.
- **The fix is a pattern, not just a patch.** The "imperative mixin config survives session round-trip via a serializable replay list" pattern generalizes. Any mixin that stashes a non-serializable handle in `mount()` can adopt the same shape (one `_configs_saved` list + one `_restore_*` method + one line in the WS consumer).

## What Didn't Work

- **The "allow_upload in mount() doesn't work" bug report was not the actual title.** The reporter's hypothesis #1 was "the WS consumer creates a fresh view instance for the upload handler path that doesn't go through `mount()`" — the right intuition (a view instance that didn't run mount) but wrong mechanism (it IS the same instance, but mount was SKIPPED, not the instance swapped). Took me a bit to get from "disprove hypothesis 1" to "find the actual skip path". Root-cause-first is still the right first move.
- **Stage 11 review added value (as always) but I'd almost talked myself out of running it** — the bug fix is ~70 LOC, all tests pass, case seems airtight. Reviewer caught the forward-compat schema-stability concern (→ #892) which I hadn't considered. Never skip Stage 11.

## What To Improve

- **Document the "imperative mixin config" pattern as an ADR.** This is now 2 observed cases (`UploadMixin`, and potentially `NotificationMixin` does similar registrations via `self.listen()`). An ADR explaining: "mixins whose `mount()` sets non-serializable handles must also maintain a serializable `_*_configs_saved` list and a `_restore_*_configs` method that the WS consumer's restoration path calls." Would prevent the next bug of this shape from getting shipped.
- **Audit all other mixins for this pattern.** Quick scan for `mount() → self.xxx = NonSerializable(...)` patterns. `NotificationMixin.listen()` stashes channels in a set — serializable, probably fine. `WizardMixin.mount()` sets primitive dicts — fine. But a single pass through `grep "def mount" python/djust/` would catch anything I missed. Tech-debt issue worth opening.

## Relationship to prior milestone work

- **Mirrors retro-884's `LayoutMixin` pattern.** `LayoutMixin` uses `set_layout()` which queues a pending path (serializable string) + `_drain_pending_layout()` in the consumer. Same "mixin + consumer cooperation" shape. The fix here is the "restore" side of the same coin.
- **Validates the deferred-findings discipline.** PR #878 opened #879-#882, PR #884 opened #879+#886-#887, PR #885 opened #886-#888. Opening #892 at merge time (instead of blocking this fix on it) is the same pattern: stay on the critical path, track follow-ups explicitly.

- Add to djust-internal docs: "mixins that use non-serializable handles in `mount()` must implement the replay-list pattern — see `UploadMixin._restore_upload_configs` as the canonical example."
- ADR for the pattern itself.

## Review Stats

| Metric | Value |
|---|---|
| Tests added | 10 Python cases |
| 🔴 findings (Stage 11) | 0 |
| 🟡 findings (Stage 11) | 0 |
| 🟢 suggestions (Stage 11) | 1 (forward-compat schema — filed as #892) |
| CI failures | 0 |
| Re-commit attempts | 1 (clean first try) |
| Time from issue-report to merged-fix | ~45 minutes |

## Deferred Items

- #892 — defensive replay for schema-changed saved configs (forward-compat concern raised by Stage 11 reviewer)
- (Possible) open a tech-debt for the ADR on the imperative-mixin-config pattern

Issue #808, opened 2026-04-21
tech-debt: PostgresNotifyListener event-loop binding across async_to_sync


johnrtipton:
PR #807 Stage 11: the singleton listener is tied to one event loop. async_to_sync calls from outside that loop context could race. Verify or document the binding.

PR #895, opened 2026-04-22
fix(mixins): replay PresenceMixin + NotificationMixin side effects on WS state restoration (#893, #894)

### LINKED ISSUE #893: bug: PresenceMixin track_presence() side effect lost after WS state restoration (same pattern as #889)
## Bug

Sibling to #889. When a view calls `self.track_presence(...)` in `mount()` and the HTTP→WS flow goes through the state-restoration path (which skips `mount()`), the user's own presence is **not** re-registered with the process-wide `PresenceManager`.

## Reproduction

```python
class EditorView(PresenceMixin, LiveView):
        self.doc_id = kwargs.get("doc_id")
        self.track_presence(meta={"name": request.user.username, "color": "#6c63ff"})
```

HTTP GET renders the page, session state is saved. Client opens WS. The state-restoration path restores `_presence_tracked=True`, `_presence_user_id`, `_presence_meta` — but does NOT re-call `PresenceManager.join_presence(...)`. Symptoms:

1. Other users connected to the same presence key don't see this user.
2. `handle_presence_join` doesn't fire for this user's own join on the restored view.
3. A later explicit `self.track_presence(...)` call early-returns because `_presence_tracked=True` is already set (`presence.py:203-204`) — so the only fix today is for the app to manually clear the flag, which shouldn't be their job.

## Root cause

Same class as #889. `presence.py:196-227` splits into two actions:

- Attribute assignment (serializable, round-trips OK): `_presence_tracked`, `_presence_user_id`, `_presence_meta`
- **Side effect** (lost on restore): `PresenceManager.join_presence(key, user_id, meta)`

The session round-trip preserves the attributes but has no mechanism to replay the side effect on the WS side.

## Proposed fix

Mirrors the #891 pattern for uploads:

1. Add a `_restore_presence()` method to `PresenceMixin` that re-calls `PresenceManager.join_presence()` from the already-restored `_presence_*` attrs when `_presence_tracked=True`.
2. Wire the WS consumer's state-restoration path (`websocket.py:1540-1572`) to call `_restore_presence()` right after `_restore_private_state()` — same pattern as the `_restore_upload_configs()` call added in #891.

Scope: ~20 LOC + 3-4 regression tests. No public-API change.

## References

- PR #891 fixed the same class of bug for `UploadMixin`
- Full pattern analysis: https://github.com/djust-org/djust/pull/891
- Audit finding: the third known case of this pattern in the codebase (first: #889/UploadMixin, second: #894/NotificationMixin — filed separately)

### LINKED ISSUE #894: bug: NotificationMixin listen() LISTEN statement lost after WS state restoration across process boundary (same pattern as #889)
## Bug

Sibling to #889. Narrower trigger than the presence case (#890) but same root cause.

When a view calls `self.listen("channel")` in `mount()` and the HTTP request + the WS connection land on **different processes** (server restart between HTTP and WS, sticky-session load-balancer sending HTTP to process A and WS to process B, worker restart, etc.), the process-wide `PostgresNotifyListener` on the WS-side process has NOT issued a `LISTEN channel` SQL statement — so NOTIFYs never flow to this view.

## Reproduction

```python
class OrdersView(NotificationMixin, LiveView):
        self.orders = list(Order.objects.filter(status="pending"))
        self.listen("orders")

    def handle_info(self, message):
        if message["type"] == "db_notify":
            self.orders = list(Order.objects.filter(status="pending"))
```

1. HTTP GET → server A: `mount()` runs, `listen("orders")` calls `PostgresNotifyListener.instance().ensure_listening("orders")` on process A → Postgres `LISTEN orders` is issued.
2. Session state is saved (`_listen_channels = {"orders"}`).
3. Server A restarts (deploy), or the sticky-session LB sends WS to server B.
4. WS connects to process B's consumer. Restoration path restores `_listen_channels = {"orders"}`. WS consumer (`websocket.py:1391-1400`) correctly joins the Channels group for `djust_db_notify_orders`.
5. **But** `PostgresNotifyListener.instance().ensure_listening("orders")` is never called on process B. No `LISTEN orders` SQL — no NOTIFYs received.
6. A later `NOTIFY orders, '{...}'` in the database reaches process A's listener (which is gone or restarted) but never process B's.

Result: the restored view believes it's subscribed (the Channels group registration succeeds) but the underlying Postgres LISTEN hasn't been issued on the new process.

## Root cause

Same class as #889 and #890. `notifications.py:45-80` has two actions:

- Attribute assignment (serializable, round-trips OK): `_listen_channels: Set[str]`
- **Side effect** (lost on restore): `PostgresNotifyListener.instance().ensure_listening(channel)` — process-wide registration that must be re-issued on the WS-side process.

## Proposed fix

Mirrors the #891 pattern:

1. Add a `_restore_listen_channels()` method to `NotificationMixin` that iterates `self._listen_channels` and calls `ensure_listening(channel)` for each one (idempotent — `PostgresNotifyListener.ensure_listening` already checks `if channel in self._channels: return`).
2. Wire the WS consumer's state-restoration path (`websocket.py:1540-1572`) to call `_restore_listen_channels()` right after `_restore_private_state()`.

Scope: ~15 LOC + 2-3 regression tests. No public-API change.

## Why this is narrower scope than #889

- **Single-process deployments** (most dev/small-prod) aren't affected — `PostgresNotifyListener` is process-wide, and if the process handles both HTTP and WS the `LISTEN` registration persists.
- **Multi-process / restart-after-mount scenarios** are the trigger. Production apps with load balancers or rolling-deploy cycles DO hit this.

## References

- PR #891 fixed `UploadMixin` with the same pattern
- Issue #893 is the sibling `PresenceMixin` case
- Pattern discussion in the #891 retro

johnrtipton:
Closes #893. Closes #894.

Sibling bugs to #889, discovered via audit after [PR #891](https://github.com/djust-org/djust/pull/891) shipped the UploadMixin fix.

## The pattern

Both mixins have `mount()`-called methods with process-wide side effects beyond setting instance attrs. The WS consumer's state-restoration path (`websocket.py:1540-1572`) skips `mount()` when pre-rendered session state exists — so the side effects are never re-issued on the WS side.

## #893 — PresenceMixin

`track_presence()` calls `PresenceManager.join_presence(key, user_id, meta)` as a per-process singleton registration. On restore, `_presence_tracked=True` etc. are restored but the PresenceManager has no record — other users don't see the restored user, and `handle_presence_join` for the user's own join doesn't fire.

## #894 — NotificationMixin

`listen(channel)` calls `PostgresNotifyListener.instance().ensure_listening(channel)` which issues the Postgres `LISTEN channel` SQL statement on the current process. On a **cross-process restore** (server restart between HTTP and WS, sticky-session LB routing WS to a different worker, worker reshuffle under load), the destination process's listener has no subscriptions — NOTIFYs never reach the restored view.

## Fix (mirrors PR #891)

- **`PresenceMixin._restore_presence()`** — replays `PresenceManager.join_presence` when `_presence_tracked=True`. No-op on not-tracked / missing `_presence_user_id` / missing `_presence_meta` (treated as empty). Backend exceptions logged + swallowed.
- **`NotificationMixin._restore_listen_channels()`** — iterates `_listen_channels` and calls `ensure_listening` on each (idempotent — `PostgresNotifyListener` already dedupes known channels). Per-channel failures logged + continue; `DatabaseNotificationNotSupported` logged specifically.
- **`websocket.py`** — adds two lines to the state-restoration path, right after the existing `_restore_upload_configs()` call from #891:
  ```python
  if hasattr(self.view_instance, "_restore_presence"):
      await sync_to_async(self.view_instance._restore_presence)()
  if hasattr(self.view_instance, "_restore_listen_channels"):
      await sync_to_async(self.view_instance._restore_listen_channels)()
  ```

Both methods are defensive — restoration must never kill the WebSocket, so missing attributes, backend errors, and per-item failures are all caught and logged at WARNING.

## Test plan

- [x] **11 regression cases** in `tests/unit/test_mixin_restoration_893_894.py`:
  - Presence: tracked-true replay, not-tracked no-op, missing user_id guard, missing meta defaults, backend-exception swallow
  - Notifications: per-channel replay, empty-set no-op, missing-attr no-op, per-channel failure tolerance, postgres-unavailable handling
  - End-to-end: HTTP mount attrs → JSON session round-trip → fresh WS-side instance → both side effects re-issued
- [x] **62 existing tests** (`test_db_notifications.py` + `test_upload_restoration_889.py`) still green
- [x] **104 total tests** across the restoration / upload / notification family pass cleanly

## Scope

Single PR, ~60 LOC production code + ~290 LOC tests. Documented pattern in the module docstrings for future maintainers.

## References

- [PR #891](https://github.com/djust-org/djust/pull/891) — the `UploadMixin` fix this PR replicates
- [Pattern discussion](https://github.com/djust-org/djust/issues/889#issuecomment-4297678852) — root-cause analysis for #889, applies directly to both bugs here


johnrtipton (commented):
## Stage 11 Code Review — PR #895

**Verdict: APPROVE (with 2 non-blocking concerns)**

The pattern is executed faithfully against PR #891. Both new `_restore_*` methods are shaped like `_restore_upload_configs`: no-op-on-missing, idempotent, defensive. CHANGELOG entry is thorough, 11 tests cover the surface.

### What's done well

- **Pattern parity with the `listen()` original call site.** `_restore_listen_channels` uses `async_to_sync(PostgresNotifyListener.instance().ensure_listening)(channel)` — byte-for-byte the same wrapper as `mixins/notifications.py:74`. That's the right choice: if the original site's async-to-sync bridging works, so does the replay.
- **Defensive guards on presence (`presence.py:252-257`)** — explicit `_presence_tracked` check, explicit `user_id` truthiness check, and `meta or {}` fallback. Good match for the test cases covering each.
- **`DatabaseNotificationNotSupported` gets a dedicated log** (`mixins/notifications.py:99-106`) — distinguishes "backend changed" from generic failure. Better telemetry than a bare `Exception` catch-all.

### Claims verified

1. **`ensure_listening` idempotency** — Confirmed: `db/notifications.py:185-186` does `if channel in self._channels: return` before any work. Replaying on the same-process is free. ✓
2. **`PresenceManager.join_presence` idempotency** — **Partial.** Per `backends/memory.py:52-53`, `join` does `self._groups.setdefault(...)[user_id] = record` — it **overwrites** with a fresh `joined_at` timestamp and heartbeat rather than deduping. For the restoration use case this is benign (fresh timestamp is arguably correct), but the CHANGELOG's "both idempotent — `PresenceManager` dedupes" wording is **technically inaccurate**: the outcome is convergent/overwrite, not dedupe. Redis backend at `backends/redis.py:73+` looks similar. Not a bug; just a precision nit on the CHANGELOG.

### Concerns (non-blocking)

**C1. `_assert_same_loop` can raise under the specific cross-loop case the fix targets.** `db/notifications.py:195-217` rejects cross-loop callers with a `RuntimeError` citing issue #808. If a different worker process has an already-started listener bound to a different asyncio loop (exactly the deployment scenario #894 describes), `ensure_listening` will raise and the broad `except Exception` in `_restore_listen_channels` will swallow it — the user sees a WARNING log but no NOTIFYs flow. The fix doesn't break anything, but it doesn't fully solve the cross-process restoration story for an already-active listener either. Worth a follow-up issue. Test `test_restore_swallows_per_channel_exceptions` accidentally covers this class of failure via its generic `RuntimeError`.

**C2. Error-handling breadth vs. UploadMixin parity.** `uploads.py:1072-1090` — `_restore_upload_configs` has NO try/except. It lets exceptions bubble (relying only on the `if not saved: return` fast-path). The new methods both catch `Exception`. That's arguably *better* than #891, but it's explicitly not parity. If `_restore_upload_configs` ever raises (e.g. `allow_upload` validation), it will kill the WS; the new methods won't. Recommend opening a follow-up to align #891's method to the broader catch pattern shown here.

### Question

**Q1. End-to-end test gap.** `test_http_mount_then_restore_replays_both_side_effects` (lines 251-298) manually drives `json.dumps/loads` — it does **not** exercise `LiveViewConsumer.handle_mount`'s restoration path (`websocket.py:1555-1572`) where the `hasattr` guards actually live. If someone removed one of the two new `hasattr` checks, this test would still pass. Would you be open to a companion test that spins up the consumer path (even a mocked one), at least asserting the two `_restore_*` methods are invoked? Not a blocker — the unit tests give high confidence the methods themselves work.

### Edge cases considered

- `_listen_channels` missing entirely (never called `listen()`) → covered by `test_restore_with_missing_attribute_is_noop`
- `_listen_channels` as empty set → covered
- `_presence_user_id = None` with `_presence_tracked=True` → covered
- `_presence_meta = None` → covered (falls back to `{}`)
- Set iteration order → does not matter for `ensure_listening`; per-channel operations are independent
- Class-attr-annotation vs instance-attr for `_listen_channels` → `getattr(self, "_listen_channels", None)` at `mixins/notifications.py:85` handles both the annotation-only-never-set path and the empty-set path correctly

### Documentation trail

Agreed with the PR author: three `_restore_*` methods is the point to write an ADR. Recommend a follow-up issue titled *"ADR: side-effect replay pattern for WS state restoration"* referencing #889/#891, #893/#895, #894/#895. Not a blocker.

### Verdict: APPROVE

The fix is correct, tested, and follows the established pattern. C1/C2 are follow-ups, not blockers. Please correct the CHANGELOG wording (swap "dedupes" for "is convergent / overwrites" for `PresenceManager`) in a follow-up or inline before merge — optional.

johnrtipton:
# Retro — PR #895 (fix: PresenceMixin + NotificationMixin restoration, #893 #894)

**Branch**: `fix/restore-mixin-side-effects-893-894`
**Merged**: 2026-04-22
**Scope**: Batch bugfix for two sibling bugs. Both fixes follow the pattern established by PR #891.

## What Worked Well

- **Audit-driven discovery.** After PR #891 merged, I deliberately scanned for the same anti-pattern — `self._<name> = <SomeClass>(` inside mixin methods commonly called from `mount()`. Found two more cases immediately: `PresenceMixin.track_presence` and `NotificationMixin.listen`. Batch-filing and batch-fixing both in one PR was the right call — the code is structurally identical.
- **Reference implementation paid off.** #891 existed as the working template. Writing `_restore_presence` and `_restore_listen_channels` was mechanical: copy the method shape, swap the side-effect call, adjust the "no-op" conditions for each mixin's attrs. ~60 LOC production + 290 LOC tests in one commit.
- **Stage 11 reviewer caught a real imprecision.** My CHANGELOG claimed `PresenceManager.join_presence` "dedupes" — reviewer checked the backend code and found it OVERWRITES (convergent but not dedup). Fixed wording inline before merge. This is the kind of thing an author never catches about their own writing.
- **Batch PR felt right.** Two ~20 LOC fixes grouped into one ~350 LOC PR (tests dominate). Reviewer had full context on both; CI ran once; one retro; closes two issues. DRY on the ceremony.

## What Didn't Work

- **Test coverage gap noted by reviewer.** The end-to-end test drives a JSON round-trip manually — it exercises the `_restore_*` methods directly but NOT the WS consumer's `hasattr` guard that decides whether to call them. If either `hasattr` check got deleted, the tests would still pass. Should have added a consumer-level integration test that verifies the restoration path actually reaches the new methods. Deferred to the ADR follow-up (#897).
- **Async-loop interaction flagged but not fixed.** The `_assert_same_loop` guard from #808 raises `RuntimeError` in exactly the cross-process scenario #894 targets. My fix swallows the RuntimeError (WS stays alive) — but NOTIFYs still won't flow to the restored view, because the listener isn't rebound to the new process's loop. The fix is still net-better than current state (nothing happens) but doesn't fully solve the cross-process case. Tracked as #896.

## What To Improve

- **ADR for the `_restore_*` pattern** (#897). After three mixins implementing the same shape, it's a pattern worth documenting. Would help the next mixin author spot the trap before shipping the bug.
- **UploadMixin parity gap** (also in #897). `_restore_upload_configs` from PR #891 doesn't catch generic `Exception` — the two new methods in this PR do. Should align for consistency.
- **Write a lint/audit rule for mount() side effects.** Grep for `def mount(self, request, **kwargs)` + `self\._.*\s*=\s*[A-Z][a-z]` bodies with "Manager" / "Singleton" / "PostgresNotifyListener" in their reachable call graphs. Would have caught all three mixins automatically.

## Relationship to Prior Work

- **Direct descendant of PR #891.** Same root cause, same fix shape, different side effects. The audit loop ("fix bug → scan for pattern → file siblings → batch fix") caught two bugs that would have otherwise shipped to users.
- **Reinforces retro-884 + retro-891 learnings.** Three PRs in a row have used the "mixin + consumer cooperation + `_restore_*` replay" motif (LayoutMixin for outbound, UploadMixin/PresenceMixin/NotificationMixin for inbound side effects). This is now a load-bearing djust convention.

- ADR (#897).
- CONTRIBUTING.md section: "Mixins with process-wide side effects must provide a `_restore_*` method and wire it into the WS consumer's state-restoration path."
- Possibly a `djust_audit` check: `A080` — mixin with `mount()`-touched non-serializable instance attrs must have matching `_restore_*` method.

## Review Stats

| Metric | Value |
|---|---|
| Tests added | 11 cases (5 presence + 5 notification + 1 end-to-end) |
| 🔴 findings (Stage 11) | 0 |
| 🟡 findings (Stage 11) | 2 (1 inline fix + 1 deferred) |
| 🟢 suggestions | 2 (follow-up tech-debt issues #896 + #897) |
| CI failures | 0 |
| Re-commit attempts | 1 (ruff reflow on first commit) |
| GH issues opened | #896 (async-loop interaction), #897 (ADR + parity) |

## Deferred Items

- #896 — `_restore_listen_channels` + `_assert_same_loop` interaction on cross-loop restore (narrower cross-process case where the listener's bound loop is stale)
- #897 — ADR for the `_restore_*` pattern + align `UploadMixin._restore_upload_configs` with the broader `except Exception` swallow pattern

PR #898, opened 2026-04-22
feat(animations): dj-remove — exit animations before element removal (v0.6.0 phase 2a)

johnrtipton:
## Summary

- **Phoenix `JS.hide` / `phx-remove` parity.** Declarative exit animations: `<div dj-remove="fade-out">` delays removeChild until the CSS transition completes.
- **5 VDOM-patch hook sites** in `12-vdom-patch.js` defer physical removal when `[dj-remove]` is present; duplicate-node accumulation guarded on the replace sites.
- **10 JSDOM tests** in `tests/js/dj_remove.test.js`. Phase 2a of the v0.6.0 Animations work (phases 2b-d — FLIP, `dj-transition-group`, skeletons — remain follow-ups).

## Details

Two attribute forms:
- **Three-token**: `dj-remove="opacity-100 transition-opacity-300 opacity-0"` — start/active/end classes, mirrors `dj-transition`.
- **Single-token**: `dj-remove="fade-out"` — applies the class, listens for `transitionend`.

**Fallback timer**: 600 ms default, overridable with `dj-remove-duration="N"`, clamped to [0, 30000] ms. Covers the `transitionend`-never-fires case (zero-duration transitions, `display: none`, etc.).

**Cancellation**: per-element `MutationObserver` watches the `dj-remove` attribute. If a later patch strips it, pending state cleans up and the element stays mounted.

**Hook API**: `window.djust.maybeDeferRemoval(node)` — returns `true` when ownership is taken (caller must skip its normal `remove()`). Five patch-applier sites wired up:
- `morphChildren` — unmatched-existing removal
- `morphElement` — tag-mismatch replace (inserts replacement before deferred node so the animation is visible)
- `applyDjUpdateElements` — `dj-update` prune
- `applySinglePatch` case `Replace`
- `applySinglePatch` case `RemoveChild`

The `morphElement` and `Replace` sites pre-check `_pendingRemovals.has(existing)` to avoid inserting a second replacement clone if the same node is touched by a repeated patch.

## Test plan

- [x] 10 new JSDOM cases cover: 3-token / 1-token / malformed spec parsing; deferred removal happy path; fallback timer; `dj-remove-duration` override; VDOM `RemoveChild` integration; attribute-strip cancellation; detached-element early return.
- [x] Full JS test suite: 1252 passed, 2 pre-existing skips.
- [x] Full Python test suite: 3424 passed, 15 skipped.
- [x] `dj-transition` sibling suite unaffected.
- [ ] Manual smoke test on the demo project (recommended but not blocking — jsdom covers the logic).

## Related

- Phase 1 (`dj-transition`) shipped in #885.
- Pairs with already-shipped `dj-mounted` (enter event) for a symmetric enter/exit story.
- ROADMAP.md updated: phase counter now reads "phases 1 + 2a of 4".


johnrtipton (commented):
## Code review — dj-remove (v0.6.0 Animations phase 2a)

### Verdict: APPROVE

Clean, well-scoped implementation. All 5 VDOM-patch hook sites are wired with the same guard shape, the duplicate-node accumulation fix is symmetric across both replace sites, and MutationObserver cleanup is correct on every exit path. Test suite is meaningful — no rubber-stamps.

### Must-fix (red)
None.

### Should-fix (yellow)
None blocking. See questions below for one minor thought.

### Nits (green)
- `src/42-dj-remove.js:147` — `requestAnimationFrame` fallback uses `setTimeout(cb, 16)`. Under jsdom + vitest parallel load this is documented (tests use 60 ms margin) but worth a one-line code comment there too for future maintainers.
- `src/42-dj-remove.js:74-82` — `_finalizeRemoval` and `_cancelRemoval` duplicate the teardown block (`clearTimeout` / `removeEventListener` / `observer.disconnect` / `_pendingRemovals.delete`). A tiny `_teardownState(el, state)` helper would dedupe ~5 lines and make "do all exit paths clean up?" trivially auditable. Not blocking.
- `src/12-vdom-patch.js` — the 5 hook-site guards run `typeof globalThis.djust.maybeDeferRemoval === 'function'` on every removal, including elements without `dj-remove`. The short-circuit is cheap (3 checks, all fast) and the function itself early-returns on `hasAttribute` miss, so overhead is negligible. No action.
- Parse-spec rejects the 2-token form, which is correct and matches `dj-transition`. A dev using `dj-remove="fade-out 300"` will get silent non-animation. Consider a `console.warn` under `globalThis.djustDebug`. Defer to follow-up.

### Edge cases considered

- Detached element — covered by `maybeDeferRemoval` early-return at `42-dj-remove.js:167` and test at `dj_remove.test.js:75`. ✓
- Double-deferral on the same element — guarded at `42-dj-remove.js:169` (`_pendingRemovals.has(el)` returns `true` without re-running). ✓
- Re-adding `dj-remove` after it was stripped (cancel → re-deferred) — the MutationObserver only triggers cancel on *removal* of `dj-remove`; a subsequent re-add does nothing (observer has already disconnected in `_cancelRemoval`). Next patch would re-enter via `maybeDeferRemoval` from a clean slate. ✓
- Cancellation during the pre-rAF window (three-token form) — `_cancelRemoval` tries to strip `active` and `end` which haven't been added yet; `classList.remove` is a safe no-op, so this is correct. ✓
- Duplicate-node accumulation — fixed symmetrically at `12-vdom-patch.js:738-742` (morphElement) and `12-vdom-patch.js:1213-1219` (applySinglePatch Replace). Both pre-check `_pendingRemovals.has(existing)` before inserting the replacement clone. ✓
- `dj-remove` on element with no CSS transition — not explicitly tested, but the 600 ms fallback timer covers it (and `_durationFor` clamps malformed values to `[0, 30000]` at line 55-57). 🟡 Not tested — see question below.
- Focus during deferred removal — not explicitly tested. A focused `<input dj-remove>` will stay focused through the animation, then blur when `removeChild` fires. Phoenix behaves identically, so this is intentional.
- Deep `[dj-remove]` interaction with `dj-update` — `applyDjUpdateElements` hook at `12-vdom-patch.js:969` correctly defers before `existing.remove()`. Descendants travelling with the parent is the documented contract.

### Questions for the author

1. **No-transition case** — a user who writes `<div dj-remove="no-op-class">` where `no-op-class` has no `transition:` CSS will never fire `transitionend`, and will wait the full 600 ms fallback before removal. This is the documented fallback, but the guide doesn't mention it as a concrete gotcha. Worth a paragraph in `declarative-ux-attrs.md`? (Defer to follow-up — not blocking.)

### What I verified

- All 5 hook-site guards present and consistent at `12-vdom-patch.js` lines 709-710, 738/751, 969-970, 1213/1225, 1387-1388.
- Duplicate-node guard (`_pendingRemovals.has(existing)` pre-check) present at BOTH replace sites — `morphElement:738` and `applySinglePatch Replace:1213`.
- All three teardown/cleanup functions (`_finalizeRemoval`, `_cancelRemoval`, and the rAF cancellation check at line 149) disconnect observer + clear timer + remove listener + delete WeakMap entry. No leak path found.
- WeakMap (not Map) for `_pendingRemovals` — GC-safe if element is discarded without going through a hook.
- Test count consistency: CHANGELOG says "10 JSDOM cases", guide edit says "10 JSDOM tests" (in ROADMAP), PR body says "10 JSDOM tests", actual `tests/js/dj_remove.test.js` has 10 `it(...)` blocks. ✓
- ROADMAP phase counter updated from "phase 1 of 4" to "phases 1 + 2a of 4". ✓
- Guide `declarative-ux-attrs.md` strikes the old scope bullet and adds a `dj-remove` section with quick-start, cancellation, and interop subsections. ✓
- No `mark_safe`, no `@csrf_exempt`, no new f-string logging, no `console.log` without `djustDebug` guard. No new XSS / DoS surface (duration clamped to [0, 30000] ms).

johnrtipton:
# Retro — PR #898 (feat(animations): dj-remove)

**Branch**: `feat/dj-remove-v060`
**Merged**: 2026-04-22
**Scope**: Phase 2a of the v0.6.0 Animations & transitions task. Ships `dj-remove` — exit-animation deferral of physical removal. Phases 2b-d (FLIP, `dj-transition-group`, skeletons) remain separate follow-ups.

## What Worked Well

- **Pre-built plan with line-precise edit sites.** Stage 4 spent extra time finding the five exact VDOM-patch removal sites (morphChildren, morphElement, applyDjUpdateElements, applySinglePatch Replace, applySinglePatch RemoveChild) with before/after snippets. Stage 5 executed verbatim in one pass — zero edit-rework loops.
- **Self-review flagged 3 🟡 BEFORE stage 11.** Duplicate-node accumulation on the two replace sites + transitionend propertyName filtering was caught by the self-review stage, fixed in-branch, before the PR was opened. Stage 11 had zero 🔴/🟡, only 3 nits. This is the target: self-review is the quality gate, stage 11 is the cross-check.
- **Fixed-pass test added coverage proactively.** The detached-element early-return wasn't in the original 9 tests; the self-review nit about it became test #10. Shipping with the guard covered rather than shipping the guard alone was the right call.
- **Stage 7 + Stage 8 ran in parallel (Self-Review + Security Check).** Saved ~2 min wall clock. Security review turned up only one green nit; finding re-confirmed by stage 11's independent read.

## What Didn't Work

- **CHANGELOG/ROADMAP test-count drift between stages.** Wrote "9 JSDOM cases" in stage 5, then the fix pass added a 10th test and I had to re-edit CHANGELOG + ROADMAP in stage 9 to match. Avoidable if the fix pass bumps the count on the way in, or if stage 9 derives the count from the actual test file (`grep -c "it("` ).
- **Client.js was regenerated 3 times.** Once after initial impl, once after fixes, once as a "final build". Could have been 2 if the fix pass rebuilt instead of leaving it for stage 9. Minor.

## What To Improve

- **When stage 6-fixer adds a test, update CHANGELOG + ROADMAP test-count in the same edit pass.** Saved edit in stage 9.
- **Add a precommit assertion that CHANGELOG's test-count matches `grep -c "it(" tests/js/<module>.test.js`** for feature-branch commits touching a new test file. Would have caught the 9→10 drift automatically.
- **Single "final build" rule — run `scripts/build-client.sh` exactly once, after all source edits are done.** The triple-build waste is mechanical.

- Add a note to the djust JS test skill about the vitest parallel-load flake pattern (already documented in retro-885, but re-confirmed here — our new tests don't hit it because they rely on the fallback timer, not `transitionend`).
- CLAUDE.md: mention the `maybeDeferRemoval` hook contract for future VDOM patch work. Any new removal site in `12-vdom-patch.js` must call the hook or dj-remove will silently not fire at that site.

## Recurring Issue Tracker

- **Test-count drift across artifacts** — 1st observed session for this specific pattern. Watch for repeats; if it happens again, automate.
- **Vitest parallel-load flakes** — not hit this PR (tests rely on fallback timer, not `transitionend`). No action needed; dj-transition retro #885 has the right plan for this via issue #888.

## Review Stats

| Metric | Value |
|---|---|
| Tests added | 10 cases, all passing |
| 🔴 findings (Stage 11) | 0 |
| 🟡 findings (Stage 11) | 0 |
| 🟢 nits (Stage 11) | 4 (1 code-comment, 1 refactor, 1 perf no-op, 1 doc follow-up — all deferred) |
| Self-review findings (Stage 7) | 3 🟡 + 4 🟢 — all addressed in-branch before PR opened |
| CI failures | [TBD — filled in after merge] |
| Re-commit attempts | 1 (clean landing, zero hook failures) |
| GH issues opened | 0 (all Stage 11 nits are in-PR deferrable, worth an issue for doc-follow-up only) |

## Deferred Items (as GitHub issues)

- tech-debt: add `_teardownState(el, state)` helper in 42-dj-remove.js to dedupe the ~5-line cleanup block in `_finalizeRemoval` and `_cancelRemoval`. (Stage 11 nit.)
- tech-debt: `console.warn` under `globalThis.djustDebug` when `dj-remove="X Y"` (2-token form) is encountered — currently silent fall-through. (Stage 11 nit.)
- tech-debt: guide doc follow-up — call out the "no-CSS-transition case waits the full fallback" gotcha in `declarative-ux-attrs.md`. (Stage 11 question.)

## Integration notes for phase 2b-d

Phase 2b (FLIP / list reordering): will need its own hook into the morph-children path, probably near line 707 in `12-vdom-patch.js`. FLIP measures rect → patches → measures rect → transform-animates. Does NOT need `dj-remove` integration — FLIP only applies to nodes that remain in the DOM.

Phase 2c (`dj-transition-group`): coordinates enter + exit on a list; should build on both `dj-transition` (enter side) and `dj-remove` (exit side) rather than re-inventing. Adding a group-level option like `dj-transition-group-enter="..."` + `dj-transition-group-leave="..."` that auto-applies the corresponding `dj-transition` / `dj-remove` to each child keeps the layering clean.

Phase 2d (skeletons): mostly CSS + a small JS helper to toggle a class on a loading container. Separate from the animation pipeline; can ship independently.

PR #904, opened 2026-04-22
feat(animations): dj-transition-group — coordinated enter/leave for list items (v0.6.0 phase 2c)

johnrtipton:
## Summary

- **React `<TransitionGroup>` / Vue `<transition-group>` parity.** Declarative parent attribute that auto-wires enter + leave animations on children.
- **Pure orchestration layer.** Writes `dj-transition` / `dj-remove` attributes onto children; reuses the already-shipped #885 / #898 animation modules.
- **10 JSDOM tests** in `tests/js/dj_transition_group.test.js`. Phase 2c of v0.6.0 Animations. FLIP / skeletons remain follow-ups.

## Details

Attribute forms:
- **Short**: `dj-transition-group="enter-spec | leave-spec"` — pipe separator for unambiguous split.
- **Long**: bare `dj-transition-group` + `dj-group-enter="s a e"` + `dj-group-leave="s a e"` on the parent.

Opt-in initial-children enter: `dj-group-appear`. Default is leave-only (initial children get `dj-remove` so they can animate out later, but don't animate in retroactively).

Mechanism:
- **Per-parent `MutationObserver`** watches `childList` on each `[dj-transition-group]` parent. On added children: sets `dj-remove` (leave spec) + `dj-transition` (enter spec if applicable). Never overwrites author-specified values.
- **Root observer at `documentElement`** watches for new `[dj-transition-group]` attributes appearing (VDOM patch adds a parent) AND parent removal (to call `_uninstall` and disconnect the per-parent observer — prevents the known MutationObserver-holds-strong-ref leak despite `WeakMap`).
- **Idempotent `_install`** via `_installedParents` WeakMap; double-install guard on the root observer.

Exit flow: `dj-remove` is set on children at add-time, so a later VDOM patch that removes a child flows through the existing `maybeDeferRemoval` hook (#898) and the exit animation plays before physical removal.

## Test plan

- [x] 10 new JSDOM cases: parseGroupAttr valid/invalid, child-add wires attrs, respects existing attrs, mount with/without dj-group-appear, runtime add via MutationObserver, end-to-end add→VDOM-remove with deferred exit, `_uninstall` teardown, parent-removal auto-teardown.
- [x] Full JS suite: 1263 passed (2 pre-existing skips).
- [x] Full Python suite: 3424 passed (JS-only feature).
- [ ] Manual smoke test on demo (recommended, not blocking).

## Related

- Builds on #885 (dj-transition, phase 1) and #898 (dj-remove, phase 2a).
- ROADMAP phase counter updated: "phases 1 + 2a + 2c of 4". FLIP = 2b, skeletons = 2d, both deferred.


johnrtipton (commented):
## Code review — dj-transition-group (v0.6.0 Animations phase 2c)

### Verdict: COMMENT (approve in spirit; one real gap + one doc nit)

Clean orchestration layer that correctly delegates to the shipped #885/#898 primitives rather than reimplementing them. Observer discipline is sound; tests actually assert state rather than rubber-stamping. One edge case is handled incompletely.

### Must-fix
- none — nothing is broken in shipped paths.

### Should-fix
- **Attribute removal at runtime isn't handled** (`43-dj-transition-group.js` lines 482-485). The root observer subscribes with `attributeFilter: [_GROUP_ATTR]` and only re-installs when `m.target.hasAttribute(_GROUP_ATTR)` is true. If a VDOM patch or app code strips `dj-transition-group` from a parent, the per-parent MutationObserver keeps running and will keep wiring `dj-remove`/`dj-transition` on future children — stale behavior. Two-line fix: `else { _uninstall(m.target); }` in the attributes branch. This mirrors the "subsequent patch strips the attribute" cancel path that #898 explicitly supports for `dj-remove`, so users may reasonably expect the same here. Add a regression test.

### Nits
- `_resolveSpecs` (lines 415-422) silently prefers long-form over short-form when both are present. Defensible, but document it (guide currently doesn't mention precedence).
- Test at line 786 sleeps 700 ms to wait for the 600 ms `dj-remove` fallback. That is ~1 s of test-suite wallclock for one case — consider `dj-remove-duration="50"` on the child to keep this case fast and deterministic.
- Nested `[dj-transition-group]` parents: `node.querySelectorAll` at line 493 will correctly install inner groups when an outer subtree is added; not tested. Worth one paragraph in the guide clarifying that inner groups are independent (subtree:false on the per-parent observer, line 458).
- Public `globalThis.djust.djTransitionGroup` surface exposes `_installedParents` (line 532). Fine for tests; may tempt users to poke at it. Underscore prefix is the signal, but a short doc comment saying "internal, may change" would be prudent.

### Edge cases considered
- Child already has `dj-transition` / `dj-remove` — respected (lines 428-434; test at 633-648). ✓
- Direct `removeChild` bypass — documented limitation in guide; matches `dj-remove` behavior. ✓
- Initial children with/without `dj-group-appear` — both covered (tests at 650-690). ✓
- Nested groups — works structurally (subtree:false + querySelectorAll cascade), not tested.
- `dj-transition-group` attribute **added** at runtime — handled (lines 482-485). ✓
- `dj-transition-group` attribute **removed** at runtime — NOT handled (see Should-fix).
- Parent detached then re-attached — `_uninstall` → WeakMap delete on remove, `_install` on re-add via childList handler. ✓
- Double-install — `_installedParents.has(parent)` guard (line 440) + `_rootObserverInstalled` flag (lines 402, 477). ✓
- MutationObserver GC / leak — `_uninstall` explicitly `observer.disconnect()` on parent removal (lines 466-468, 496-504). ✓

### Questions for the author
1. Is the "strip the group attribute cancels wiring" behavior a non-goal for phase 2c, or an oversight? Happy either way, but `dj-remove`'s equivalent (strip-to-cancel) is shipped and users may generalize.
2. Why `subtree: false` on the per-parent observer (line 458) but no test demonstrating that a grandchild append doesn't spuriously fire? Consider an explicit "nested group children don't double-wire" assertion.
3. Short-form + long-form precedence — is long-form-wins intentional, or should a conflict throw a dev-mode warning under `globalThis.djustDebug`?

### What I verified
- CHANGELOG entry (line 9) matches "10 JSDOM cases", module path, and phase 2c label.
- ROADMAP phase counter updated "1 + 2a" → "1 + 2a + 2c" (line 23); dj-transition-group now struck through.
- Guide: TOC additions for both `dj-remove` and `dj-transition-group` (lines 35-36); full `dj-transition-group` section added (lines 54+) with quick-start, short form, appear, interop, limitations, scope.
- Module bundled into `client.js` at line 11221+ matches `src/43-dj-transition-group.js` verbatim.
- Tests: 10 `it(...)` blocks under `describe('dj-transition-group')`; assertions are substantive (`classList.contains`, `parentNode`, `_installedParents.has`, attribute equality) — no rubber-stamps.
- Security: no `innerHTML`, no `eval`, no `mark_safe`, no `console.log`. Only `setAttribute` with author-supplied strings; surface equivalent to existing `dj-transition`/`dj-remove`.
- Observer cleanup: `entry.observer.disconnect()` at line 467 + WeakMap delete at line 469; exercised by tests at lines 714-732 and 735-752.

johnrtipton:
# Retro — PR #904 (feat(animations): dj-transition-group)

**Branch**: `feat/dj-transition-group-v060`
**Merged**: 2026-04-22
**Scope**: Phase 2c of v0.6.0 Animations & transitions. Coordinated enter/leave wiring over the already-shipped dj-transition (#885) + dj-remove (#898) primitives. FLIP (2b) and skeletons (2d) remain follow-ups.

## What Worked Well

- **Orchestration-layer design paid off.** The module is a thin 160-line wrapper that writes `dj-transition` / `dj-remove` attributes onto children and lets the existing observers react. No reimplementation of class phasing, fallback timers, or cancellation logic. Reused everything shipped in #885 + #898.
- **Planning stage caught the pipe-separator ambiguity.** Initial intuition was `dj-transition-group="fade-in fade-out"` (space-separated) but two tokens can't disambiguate a three-token spec per half. Switched to `|` separator in the plan. Would have surfaced as a test failure or worse later.
- **Self-review (stage 7) caught two real issues before Stage 11.** Observer leak (MutationObserver holds strong ref, defeats WeakMap) and double-install guard on `_installRootObserver`. Both fixed inline. Made Stage 11 find only ONE new thing (attribute-strip symmetry).
- **Stage 11 finding was legitimate and small enough to fix in-branch.** Two-line fix + one test. Not worth a separate PR. Clean merge.

## What Didn't Work

- **Third consecutive test-count-drift incident.** Started at 8, self-review fix pushed to 10, stage 11 fix pushed to 11. Each bump required CHANGELOG + ROADMAP edits. Retro-898 flagged this pattern; hasn't been automated yet.
- **Read-only Plan agent couldn't write the plan file.** Had to save the plan content to disk from the main context after the Plan subagent returned it as a message. Small friction; not a real block.

## What To Improve

- **Automate test-count sync.** Add a pre-commit hook (or a `make` target) that verifies the test-count claim in CHANGELOG.md's most recent Unreleased entry matches `grep -c "it(" tests/js/<module>.test.js`. Three retros in a row flagging this — time to automate. (File as tech-debt issue.)
- **Consider a `_teardown()` fn per module for future animation modules.** dj-remove has `_cancelRemoval`/`_finalizeRemoval` (separate paths); dj-transition-group has `_uninstall`. Future FLIP / skeleton modules should expose a similar teardown. Make it a convention.
- **Plan agent sub-prompt clarification.** The Plan agent type defaults to read-only. For planning stages, the subagent prompt should explicitly say "return plan content as a message; the main loop will save it."

- Add note to CLAUDE.md about the animation-module MutationObserver pattern — per-parent observers need `_uninstall` paths (learned twice: #898 had no per-parent observers; #904 introduced them and needed the teardown from day 1).
- Reference #904 in future "TransitionGroup / sticky" / list-management features as the pattern for per-parent observer discipline.

## Recurring Issue Tracker

- **Test-count drift across artifacts** — 3rd consecutive session (retro-898, retro-904, tacit in retro-885). Pattern: fix pass adds a test → CHANGELOG/ROADMAP show the pre-fix count. Needs automation.
- **Orchestrator-pattern working across animation modules** — dj-transition-group is the first module to compose other dj-* modules via attribute wiring rather than direct function calls. Kept the public surface clean; other composite modules should follow.

## Review Stats

| Metric | Value |
|---|---|
| Tests added | 11 cases (all passing) |
| 🔴 findings (Stage 11) | 0 |
| 🟡 findings (Stage 11) | 1 (attribute-strip asymmetry) — fixed in-branch with +1 test |
| 🟢 nits (Stage 11) | 3 — 700 ms test walltime (defer), nested-group test (defer), long-form precedence doc (defer) |
| Self-review findings (Stage 7) | 2 🟡 (observer leak, double-install) — all fixed in-branch +2 tests |
| CI failures | [TBD — filled in after merge] |
| Re-commit attempts | 2 (initial + inline-fix) |
| GH issues opened | 0 (Stage 11 nits deferrable to follow-up issues) |

## Deferred Items

- tech-debt: reduce 700 ms test walltime — override `dj-remove-duration="50"` in the VDOM RemoveChild integration test.
- tech-debt: add nested `[dj-transition-group]` regression test (each parent installs its own observer with `subtree: false`; no double-wire — but currently untested).
- docs: document long-form-wins precedence in `declarative-ux-attrs.md` (resolveSpecs prefers `dj-group-enter`/`dj-group-leave` over short-form `|` split; currently only evident from reading the code).
- tech-debt (cross-PR): automate CHANGELOG test-count validation. Third occurrence.

## Integration notes for phase 2b (FLIP) and 2d (skeletons)

**Phase 2b (FLIP / list reordering)**: should integrate with dj-transition-group — a group parent with `dj-group-flip` opt-in would trigger position-measurement before + after VDOM patches and run transform-translate animations to interpolate. FLIP is the only remaining animation primitive that requires actual DOM measurement, not just attribute wiring. Will need its own module (~150 lines) with a pre-patch measurement hook.

**Phase 2d (skeletons)**: orthogonal to this work. Per the manifesto ("Zero opinions on CSS, markup, or design"), djust should ship the HTML scaffold via a `{% dj_skeleton %}` template tag + generation helpers in Rust (already partially done in `components/rust_handlers.py`) but NOT opinionated CSS. Users bring their own `.skeleton` / `.skeleton-card` styling.

PR #913, opened 2026-04-22
security(logging): sanitize user-controlled values in log calls — closes 9 py/log-injection alerts

johnrtipton:
## Summary

- Closes **9 open CodeQL `py/log-injection` alerts** (error severity).
- Adds a tiny `djust._log_utils.sanitize_for_log()` helper and applies it at 5 log call sites.
- Format strings unchanged; djust's existing `%s`-style logging convention stays.
- No app code behavior change for non-malicious input.

## Details

CodeQL flags log calls where user-controlled values (URL path params, POST body fields, exception messages) reach `logger.exception` / `logger.debug`. An attacker supplying CRLF in the value could inject fake log lines, bypass SIEM parsers, or flood the log stream.

Fix: sanitize the value before it hits the logger. The helper:
- Replaces CR, LF, TAB, and other control chars (< 0x20 except space) with `?`
- Truncates to 200 chars with `...` suffix (log-flood guard)
- Always returns a string (None -> `"None"`, non-string -> `repr`)

## Sites fixed

| File | Lines (approx) | User-controlled value |
|---|---|---|
| `python/djust/api/dispatch.py` | ~307, 362, 376, 382 | `view_slug`, `handler_name` |
| `python/djust/theming/gallery/component_registry.py` | ~522, 527 | `component_name`, `str(exc)` |

## Test plan

- [x] 8 unit tests in `python/djust/tests/test_log_sanitization.py` (CR/LF/TAB/null/unicode/truncation/None/non-string).
- [x] Full Python suite: 3428 passed, 15 skipped (no regressions).
- [x] `ruff check` + `ruff format` clean on all changed files.
- [ ] Post-merge: Dependabot/CodeQL re-scan should close the 9 alerts.

## Scope

This is one of ~5 planned PRs for the code-scanning cleanup pivot. Follow-ups:
- py/reflective-xss (6 alerts) in gallery views
- py/stack-trace-exposure (10 alerts) across observability + theming
- py/url-redirection (2 alerts) in auth/admin views
- py/path-injection (2 alerts) in storybook
- py/clear-text-storage-sensitive-data (3 alerts) in build_themes
Plus the 872 py/unsafe-cyclic-import alerts which are architectural noise (deferred).

Generated with [Claude Code](https://claude.com/claude-code)

PR #918, opened 2026-04-22
security(xss): escape user input in gallery 404 responses & theme options (6 py/reflective-xss alerts)

johnrtipton:
## Summary

Closes 6 CodeQL `py/reflective-xss` alerts (error severity) — 3 real XSS + 3 defense-in-depth.

### Real XSS (3 alerts) — `python/djust/theming/gallery/views.py`

Lines 276, 281, 306: `storybook_detail_view` and `storybook_category_view` echoed the user-controlled URL kwargs (`component_name`, `category`) into `HttpResponseNotFound(f"Unknown ...: {value}")` bodies, which Django serves with `Content-Type: text/html`. A visitor hitting `/storybook/<script>alert(1)</script>/` got the raw payload reflected in the 404 body.

**Fix**: wrap the interpolated values with `django.utils.html.escape()`.

Access to these views is already gated by `_check_access` (DEBUG=True or `is_staff`), which narrows the practical attack surface, but the unescaped reflection is still a genuine XSS sink and the fix is trivial.

### Defense-in-depth (3 alerts) — `python/djust/components/gallery/views.py`

Lines 677, 726, 762: `_resolve_theme` reads `gallery_ds` / `gallery_preset` / `gallery_mode` cookies, **validates them against an allowlist** (`if value not in systems: value = default`), and only then interpolates them into `<option>` HTML fragments. The allowlist values are plain ASCII identifiers — zero real attack surface — but CodeQL's taint analyzer doesn't model the "`in` allowlist → untainted" pattern and kept flagging it.

**Fix**: wrap the allowlisted values with `escape()` at the interpolation site. No-op on validated input, clears the taint flag.

## Test plan

- [x] 4 new regression cases in `python/djust/tests/test_gallery_xss.py`:
  - `test_theming_storybook_unknown_component_escaped_in_404` — payload `foo<script>alert(1)</script>` appears as `&lt;script&gt;` in the 404 body, never raw
  - `test_theming_storybook_unknown_category_escaped_in_404` — payload `foo` appears escaped, never raw
  - `test_components_gallery_theme_cookies_escaped` — `<script>` / `<img` never appear anywhere in `ds_options` / `preset_options` / `theme_css` when malicious cookie values are injected
  - `test_components_gallery_valid_cookie_still_works` — an allowlisted value still renders as `value="..."` + `selected`
- [x] Full Python suite: 3428 passed (+4 new), 15 skipped
- [x] `ruff check` + `ruff format` clean
- [x] Pre-commit hooks clean (bandit, detect-secrets)
- [x] Pre-push hook ran full pytest suite — passed


johnrtipton (commented):
## Code review — security(xss) gallery views (PR #918)

### Verdict: APPROVE with one question

Targeted 12-line fix that correctly escapes three real reflective-XSS sinks and adds defense-in-depth on the cookie-driven theme selector. Tests exercise the escape path with real payloads. The staff/DEBUG gate on the affected views correctly narrows severity but doesn't change the correctness of the fix.

### Must-fix / Should-fix / Nits

- **Nit (possible missed sink)** `python/djust/components/gallery/views.py:690` — `raise Http404(f"Unknown category: {category_slug}")` reflects a URL kwarg into Django's DEBUG 404 page. Mitigated by the `<slug:category_slug>` URL converter in `urls.py:41` (restricts to `[-a-zA-Z0-9_]+`, so `<` can't reach the view), and harmless in prod (Django's 404 template doesn't echo the exception message). Worth a one-line `escape()` for consistency, or a comment noting the slug-converter guard — otherwise CodeQL may eventually flag it.
- **Nit** the 3 internal `escape()` calls wrap the iterator variable (`s`, `p` from `systems` / `presets`), not the tainted cookie value itself. The cookie value is only compared to the allowlist (`==`); the HTML interpolation is over the allowlist contents. If CodeQL's flow was actually cookie→HTML, the true sanitization point is the `if value not in systems` reassignment, not the `escape()` around `s`. This may not clear the alert — will need to verify post-merge. Not blocking; the `escape()` is harmless and adds genuine defense-in-depth if `THEME_PRESETS` / `DESIGN_SYSTEMS` ever accept third-party registrations.

### Real-vs-false-positive split

Split is correctly characterised:

- **Real** (theming/gallery/views.py:277, 282, 307): user-controlled URL kwargs flow directly into `HttpResponseNotFound(f"...")`. `HttpResponseNotFound` defaults to `Content-Type: text/html`. Verified the `escape()` wraps every interpolation of `component_name` / `category`; no bare f-string sinks remain in the diffed hunks. Access-gated by `_check_access` (DEBUG or `is_staff`) at lines 270 and 300 — narrows exposure but does not eliminate (self-XSS against a staff user is still a real bug).
- **Defense-in-depth** (components/gallery/views.py:75-76, 80-81): cookie→allowlist→HTML. Note `components/gallery/views.py` has NO `_check_access` — it's publicly accessible, unlike the theming variant. The allowlist already makes this safe, so `escape()` really is a no-op on valid input.

### Tests

Four cases, all meaningful:

- The two 404-body tests use distinct payloads (`<script>` and `` / `<img` absent from three outputs including `theme_css` — correctly exercises the full render path.
- `test_components_gallery_valid_cookie_still_works` prevents the fix from accidentally double-escaping legit values. Uses `systems[0]` to avoid coupling to a specific preset bundle — good.
- Coverage gap: no test asserts that `escape()` is what did the work on a cookie value that happens to be allowlisted *and* HTML-special — but that's impossible given the current allowlist contents, so this is fine.

### Edge cases considered

- Unicode / surrogate payloads in `component_name` — `escape()` passes through non-HTML chars, so fine.
- `HttpResponseNotFound` charset — Django defaults to UTF-8; no bypass via charset confusion.
- DEBUG vs prod for Http404 at line 690 — safe in prod (generic template), only DEBUG reflects.
- Future registration of a design system with `_` in its name — `s.replace('_', ' ').title()` already runs before `escape()`; still safe.

### Questions

1. Will CodeQL actually clear the 3 defense-in-depth alerts? The `escape()` is on the *iterator variable*, not the tainted cookie value. If not, consider moving `escape()` to wrap `design_system` / `preset` at the allowlist reassignment point, which is the canonical taint-sanitization site.

### What I verified

- All three real-XSS sinks (lines 277, 282, 307) wrapped.
- `escape` imported at `theming/gallery/views.py:17` and `components/gallery/views.py:12`.
- `_check_access` gates the 3 real sinks; `components/gallery/views.py` is publicly accessible (no gate).
- Grep for `f'<` / `HttpResponse` in both files — found line 690 as a probable missed sink (mitigated by slug converter).
- Test file imports `tests.conftest` first; uses `@override_settings(DEBUG=True)` to bypass `_check_access`; `pytestmark = pytest.mark.theming` matches the `theming` marker in `pyproject.toml:130`.
- CHANGELOG entry present under `### Security`.
- Single commit (614dd312), clean branch.

johnrtipton:
# Retro — PR #918 (security(xss): gallery views reflective-XSS)

**Branch**: `fix/code-scan-reflective-xss`
**Merged**: 2026-04-22
**Scope**: Closes 6 CodeQL py/reflective-xss alerts (error severity). Mixed-flavor — 3 real XSS + 3 CodeQL false positives resolved via defense-in-depth + refactor.

## What Worked Well

- **Real-vs-false-positive triage.** Before writing code, I read both files and traced each flagged line. The 3 theming/gallery/views.py cases reflected URL kwargs (`component_name`, `category`) into HttpResponseNotFound bodies unescaped — genuinely exploitable. The 3 components/gallery/views.py cases flowed through `_resolve_theme`'s allowlist validation — safe today but CodeQL couldn't see it. Treating these with distinct fixes (escape + document) kept the PR honest.
- **Stage 11 caught another XSS site the initial impl missed.** Line 690's `raise Http404(f"Unknown category: {category_slug}")` in components/gallery/views.py. Mitigated by URL regex (`<slug:...>` converter) and Django's 404 template not echoing exception messages in production, but still a consistency gap. Fixed inline before merge.
- **Stage 11 also questioned the defense-in-depth pattern.** I originally wrapped `escape(s)` where `s` iterates over the TRUSTED allowlist, which wouldn't clear CodeQL's taint flow (the TAINTED value is the cookie). Switched to a declarative `x_raw if x_raw in allowlist else default` ternary that CodeQL's taint analysis recognizes as validation. Stronger fix.
- **5 regression tests.** 3 for real XSS assertions (script + img payloads, confirm raw-absent + escaped-present), 2 for the cookie allowlist path (malicious cookie discarded, valid cookie still works). Plus the follow-up Http404 test. Genuine security tests, not rubber-stamps.

## What Didn't Work

- **First-pass escape() placement didn't actually pacify CodeQL.** Wrapping `escape(s)` on the allowlist iterator vs. the tainted cookie value is a subtle distinction; reviewer caught it. Lesson: when doing defense-in-depth for a static analyzer, trace what ACTUALLY appears in the taint flow and target THAT.
- **Test file couldn't use `RequestFactory` directly for theming views** — they needed auth bypass via `user.is_staff`. Had to mock the user with a raw type creation (`type("U", (), {...})()`). Works but ugly. A shared test fixture (`staff_request_factory`) would be cleaner.

## What To Improve

- **Write a "taint-flow cheat sheet" for CodeQL alerts.** Common patterns: cookie → validator → use (validation = reassignment OR ternary; comparison alone may not clear taint). URL kwarg → HttpResponse body (real XSS unless escape/validate). Exception → response body (stack-trace exposure). Save as `docs/internal/codeql-patterns.md` for the remaining code-scanning cleanup PRs.
- **Shared test fixtures for auth-gated views.** Move the staff-user mock into `conftest.py` as a pytest fixture.

## Recurring Issue Tracker

- **Stage 7 → Stage 11 finds-more-sites** — happened on #898 (IME composition) and now #918 (Http404 at line 690). Pattern: initial impl fixes the flagged sites; Stage 11 greps same file for related patterns and finds more. Continue relying on this — Stage 11 is doing its job.
- **Escape-placement subtlety for static analyzers** — new observation. Future code-scanning PRs should explicitly verify the taint path, not just "escape something related".

## Review Stats

| Metric | Value |
|---|---|
| Alerts closed | 6 (CodeQL py/reflective-xss) |
| Files modified | 3 (theming/gallery/views.py, components/gallery/views.py, CHANGELOG.md) + 1 new test |
| 🔴 findings (Stage 11) | 0 |
| 🟡 findings (Stage 11) | 1 (additional Http404 site) — fixed inline |
| 🟢 nits (Stage 11) | 1 (defense-in-depth placement subtlety) — addressed in same commit |
| Python tests | 3428 + 5 new = 3433 passed |
| CI checks | [TBD — filled in after merge] |
| Re-commit attempts | 2 (initial + stage-11 follow-up) |

## Deferred Items

- tech-debt / tooling: add `conftest.py` fixture for staff-gated view tests (currently each test manually mocks `user.is_staff`).
- docs: `docs/internal/codeql-patterns.md` — common taint-flow patterns + correct escape placement. Reusable for remaining code-scanning PRs.
- verify: post-merge CodeQL re-scan actually closes the 6 alerts (especially the 3 defense-in-depth cases — if CodeQL doesn't recognize the ternary pattern, we'll need to fall back to `typing.Literal` type annotations or explicit asserts).

## Status and next steps

**Code-scanning cleanup progress**: 2 of ~5 planned PRs shipped (log-injection #913, reflective-xss #918).

**Remaining error-severity alerts**:
- py/stack-trace-exposure × 10 (observability views + theming inspector + api/errors.py + components/gallery/views.py)
- py/url-redirection × 2 (auth + admin views)
- py/path-injection × 2 (theming/storybook)
- py/clear-text-storage-sensitive-data × 3 (theming/build_themes)

Next up: **stack-trace-exposure** (10 alerts). Pattern is info disclosure via `traceback.format_exc()` or raw `str(e)` in error responses. Fix: return generic messages, log full trace server-side only.

PR #919, opened 2026-04-22
security(info-disclosure): drop exception details from API error responses (10 py/stack-trace-exposure alerts)

johnrtipton:
## Summary

Closes 8-10 CodeQL `py/stack-trace-exposure` alerts. Stack traces and exception messages can leak internal file paths, local variable names, DB schema details, and dependency versions — valuable reconnaissance for attackers. Replaces `str(e)` / `type(e).__name__: {e}` in response bodies with generic messages plus `logger.exception()` calls so the full traceback is still captured in the standard server log stream.

## Sites fixed

| File | Sites | Notes |
|---|---|---|
| `python/djust/theming/inspector.py` | 3 (`theme_inspector_api` GET, POST, `theme_css_api`) | **Publicly accessible — real prod exposure** |
| `python/djust/observability/views.py` | 4 (`reset_view_state` mount, `eval_handler` JSON body, `eval_handler` TypeError, `eval_handler` catch-all) | DEBUG-gated dev tools; CodeQL flags response content regardless |
| `python/djust/api/dispatch.py:384` | 1 (`serialize_error` TypeError path) | Generic message matches sibling branches |

## Pattern

Before:
```python
except Exception as e:
    return JsonResponse({"error": str(e)}, status=400)
```

After:
```python
except Exception:  # noqa: BLE001
    logger.exception("contextual message")
    return JsonResponse(
        {"error": "handler failed — see server logs"},
        status=500,
    )
```

Also added `logger = logging.getLogger(__name__)` to the two files that lacked one (`theming/inspector.py`, `observability/views.py`).

## Rationale

- Stack traces reveal file paths, variable names, DB schema, and dependency versions
- Generic messages do not help legitimate users but do help attackers enumerating surface
- Traceback is still captured via `logger.exception()` for ops / debugging
- Status codes bumped from 400 to 500 where the underlying failure was genuinely a server-side exception (not a client-input validation error)

## Test plan

- [x] 3 regression cases in `python/djust/tests/test_stack_trace_exposure.py` verify a sentinel exception message (`SENTINEL_INTERNAL_STATE`) is not reflected in the response body
- [x] `make test-python` — full suite: 3428 passed, 15 skipped, 0 failed
- [x] Pre-push pytest hook: passed
- [x] Ruff check + format: clean

## Note on components/gallery/views.py:726, 762

These two alerts share the reflective-XSS cookie-flow surface addressed by PR #918's `escape()` refactor. CodeQL rescan after this merge may auto-close them. If not, dismissal with "allowlist-validated input + escape already applied; static analyzer limitation" is appropriate — no file change needed in this PR.

## CHANGELOG

Entry added under `[Unreleased]` → Security.


johnrtipton (commented):
## Stage 11 Independent Review — APPROVE

### Coverage (8/8 sites converted)

- `python/djust/theming/inspector.py:293-300, 319-325, 336-344` — 3 sites, all converted with `logger.exception("theme inspector API failed")` and generic message.
- `python/djust/observability/views.py:313-320, 399-401, 474-484, 485-494` — 4 sites, all converted.
- `python/djust/api/dispatch.py:378-388` — 1 site (`TypeError` serialize_error path) converted; now mirrors sibling branches' generic text.

Logger invocations verified: every `except` branch that drops the exception message has a matching `logger.exception(...)` — no information lost, just rerouted. `getLogger(__name__)` added to the two files that lacked it (inspector.py:20, observability/views.py:25).

No residual leaky patterns in changed files. Grepped `str(e)`, `str(exc)`, `type(e).__name__`, `{e}`, `{exc}` across the three files — only clean matches are the intentional Django `PermissionDenied` 403 branches (`dispatch.py:305, 314, 360, 377`), which echo the auth message as-is (standard Django convention, not an info-disclosure vector).

### Status code changes are correct

400→500 bumps happen only in genuine server-side failure paths (unexpected `Exception` catch-alls). Client-input error branches that *should* stay 400 are preserved: `inspector.py:320` "Unknown action", `observability/views.py:401` invalid JSON, `observability/views.py:484` `TypeError` (handler signature mismatch — legitimately a client problem). Correct discrimination.

### Concerns / edge cases

1. **Inconsistency with `dispatch.py:295`** — `api_error(400, "invalid_json", f"Malformed JSON body: {exc}")` was **not** touched. `json.JSONDecodeError.__str__` returns "Expecting ',' delimiter: line 2 col 5" which is low-risk (no path/schema leakage), but the sibling `observability/views.py:399-401` JSON-parse path in this same PR **was** genericized to "invalid JSON body — see server logs". Either both are OK to leak the decoder hint (DX win) or neither is — the PR is split. Recommend either (a) follow-up to genericize `dispatch.py:295` too, or (b) re-expand `observability/views.py:401` to include `str(exc)` for the `json.JSONDecodeError` case since that's a stdlib-bounded message and the DX cost ("see server logs" for a typo'd curl is harsh) outweighs the leak risk. Not blocking.

2. **Test #3 (`test_dispatch_serialize_error_does_not_leak_exception_message`)** — uses `inspect.getsource` + substring check. Fragile to formatting (black/ruff could re-wrap the string literal and break the assertion), but pragmatic given `serialize_error` is an internal path hard to trigger end-to-end. The PR author acknowledged this in the docstring. Acceptable.

3. **Test assertions on localized strings** — tests assert `"server logs" in lowered or "internal" in lowered`. If i18n is ever added to these error messages, tests break silently. Minor.

4. **`ValueError/KeyError` split** — No regression here. The task spec's "expected validation errors echoed with message" applies to app-level form validation, which lives in `djust/forms.py` (untouched). Unexpected `Exception` catch-alls in framework internals being genericized is the correct call.

### Verdict: APPROVE

All 8 sites converted, logger invocations complete, status codes correctly discriminated between client-input and server-exception paths, no regressions. The one inconsistency (`dispatch.py:295` JSON-parse not genericized while `observability/views.py:399` was) is worth noting for a follow-up but doesn't block this merge — CodeQL scans pass and the remaining `{exc}` is stdlib-bounded.

— Independent review bot (Claude Opus 4.7)

johnrtipton:
# Retro — PR #919 (security(info-disclosure): drop exception details)

**Branch**: `fix/code-scan-stack-trace-exposure`
**Merged**: 2026-04-22
**Scope**: Closes 8-10 CodeQL py/stack-trace-exposure alerts (8 directly fixed, 2 likely auto-closed from PR #918's cookie-flow refactor).

## What Worked Well

- **Pattern recognition paid off again.** After 3 prior code-scanning PRs (#913 log-injection, #918 reflective-xss), the pattern was clear: `except Exception as e: return {"error": str(e)}` → `except Exception: logger.exception(...); return {"error": generic-msg}`. Mechanical fix across 8 sites, zero creative decisions.
- **Distinguished real sites from CodeQL artifacts.** Theming/inspector.py has NO access gating — real prod exposure. Observability/views.py is DEBUG-gated — lower impact but still fixable. Api/errors.py:24 is a helper whose taint comes from callers — fix at the CALL site (dispatch.py:384), leave errors.py alone. The components/gallery/views.py sites (726, 762) are the same cookie-flow CodeQL conservatism as #918 — leave for the post-merge rescan.
- **3 regression tests catch real regressions.** Sentinel-string pattern (`SENTINEL_INTERNAL_STATE`) in the exception message, assert it never appears in the response body. Cheap, robust, not rubber-stamp.
- **Logger call added everywhere we removed the echo.** No information was lost — just rerouted from client-response to server-log. Reviewer verified this explicitly.

## What Didn't Work

- **Inconsistency the reviewer caught (dispatch.py:295 vs observability:399).** `api_error(400, "invalid_json", f"Malformed JSON body: {exc}")` wasn't touched because it was NOT flagged by CodeQL (JSONDecodeError messages are bounded/safe). The sibling observability path WAS flagged + fixed. Now they diverge — one leaks JSON parse details (useful for DX), the other doesn't. Defensible either way but inconsistent. Defer to a style-only follow-up.
- **Test #3 is fragile** — uses `inspect.getsource()` + substring match. Would break under any whitespace/line-reflow. Acknowledged in the test docstring but still a smell.
- **`noqa: BLE001` on `except Exception` is noisy.** Used at every site because we're catch-all-ing; alternative would be to narrow to `(RuntimeError, KeyError, ...)` but that risks missing a new exception type. BLE001 is the right suppression per djust's security posture (never let an exception propagate to raw traceback output).

## What To Improve

- **When the pattern is mechanical across N files, write ONE test helper.** Instead of 3 near-identical monkeypatch-throws-sentinel test cases, a parameterized pytest fixture or a `make_leaky_view()` factory would cut duplication. Deferred — 3 tests is small enough not to matter, but 10 would.
- **Consider a project-wide lint rule that flags `except ... as e` WHERE e reaches a response body.** Custom ruff rule or bandit plugin. Would catch new instances pre-commit instead of post-scan.

- Add to CLAUDE.md (security section): "Exception details MUST NOT appear in HTTP response bodies. Pattern: `except Exception: logger.exception('context'); return generic_error(500)`. Narrow exceptions (ValueError from OWN validation) may surface their message when the caller controls the input format."

## Recurring Issue Tracker

- **Stage 11 catching cross-file inconsistencies** — this PR's `dispatch.py:295 vs observability:399` find is the same flavor as #918's line 690 find. Stage 11 grep-adjacent-files discipline is working.
- **Post-merge CodeQL rescan uncertainty** — #918's escape() defense-in-depth may or may not have cleared alerts 1980/1981. This PR explicitly deferred verification. Pattern: after each code-scanning PR, wait 24h for CodeQL rescan, then audit remaining alerts.

## Review Stats

| Metric | Value |
|---|---|
| Alerts directly fixed | 8 (3 inspector + 4 observability + 1 dispatch) |
| Alerts possibly auto-closed | 2 (gallery/views.py cookie flow, via #918 ripple) |
| 🔴 findings (Stage 11) | 0 |
| 🟡 findings (Stage 11) | 0 |
| 🟢 nits (Stage 11) | 2 (dispatch.py:295 inconsistency, fragile inspect.getsource test) |
| Python tests | 3428 + 3 = 3431 passed |
| Re-commit attempts | 2 (initial + E402 import-order fixup) |

## Deferred Items

- style: align dispatch.py:295 `api_error(400, "invalid_json", f"Malformed JSON body: {exc}")` — either genericize like observability did, or relax observability's JSON-parse genericization. One-way or the other.
- test quality: replace `inspect.getsource + substring` in `test_dispatch_serialize_error_does_not_leak_exception_message` with a behavior-level test (actually trigger the 500 path and assert response body).
- verify: after CodeQL rescan, confirm 2 components/gallery/views.py alerts (1980, 1981) auto-closed. If not, dismiss with "defense-in-depth in #918 + validation; false positive" justification.

## Status

**Code-scanning cleanup progress**: 3 of ~5 planned PRs shipped (log-injection #913, reflective-xss #918, stack-trace-exposure #919).

**Remaining error-severity alerts**:
- py/url-redirection × 2 (auth + admin views)
- py/path-injection × 2 (theming/storybook)
- py/clear-text-storage-sensitive-data × 3 (theming/build_themes)
- py/reflective-xss × 2 (components/gallery/views.py — likely false positive pending rescan)

Next up: **url-redirection + path-injection bundled** — 4 alerts across 3 files, different rule categories but same "user input → dangerous sink" shape.

PR #920, opened 2026-04-22
security: open-redirect + path-traversal fixes + dismiss clear-text CodeQL false-positives (7 alerts)

johnrtipton:
## Summary

Closes 4 real CodeQL alerts + dismisses 4 clear-text false-positives — **7 alerts cleared total**.

### Real fixes (3 code changes, 4 alerts closed)

- **`py/url-redirection` — SignupView open redirect** (`python/djust/auth/views.py`): `get_success_url` accepted any `next` POST param and passed it straight to `redirect()`, so a crafted form post could bounce newly-authenticated users to an attacker-controlled host. Now validates with Django's `url_has_allowed_host_and_scheme()` against the current request host (with `require_https=self.request.is_secure()`); off-site, protocol-relative (`//evil.com`), and scheme-different values fall back to `settings.LOGIN_REDIRECT_URL`.
- **`py/url-redirection` — admin login redirect** (`python/djust/admin_ext/views.py`): `admin_login_required` interpolated `request.path` directly into the login-redirect query string (`?next=<path>`), letting a path containing `&` / `#` / encoded control chars smuggle extra query params into the redirect. Now uses `urllib.parse.urlencode({"next": request.path})`.
- **`py/path-injection` — storybook component template loader** (`python/djust/theming/gallery/storybook.py`): `get_component_template_source` joined an HTTP-accessible `component_name` URL kwarg into `_COMPONENTS_DIR / f"{name}.html"` with no validation. Now validated with an allowlist regex `^[a-z0-9_-]+$` plus a resolved-path-under-base check so traversal payloads (`../../../etc/passwd`, `../secret`, `foo/bar`) return `""` instead of reading outside the components directory.

### Dismissed false-positives (4 alerts)

- `py/clear-text-storage-sensitive-data` + `py/clear-text-logging` — 4 alerts trace taint from `MEDICAL_THEME` / `LEGAL_THEME` constant imports in `python/djust/theming/presets.py`. CodeQL's healthcare-PII heuristic matches the word "medical" / "legal" as identifiers, but the tainted values are CSS theme names (palette tokens, radii, font stacks), not healthcare or legal data. Dismissed on GitHub with `won't fix` and justification.

### Regression tests

5 cases in `python/djust/tests/test_security_redirects_paths.py`:
- `test_signup_view_rejects_off_site_next_redirect`
- `test_signup_view_accepts_same_site_next_redirect`
- `test_signup_view_rejects_protocol_relative_redirect`
- `test_storybook_path_traversal_rejected`
[...]

PR #917, opened 2026-04-22
chore(deps): drop black — ruff format is the canonical formatter

johnrtipton:
## Summary

- Ruff's `ruff-format` has been the canonical formatter via pre-commit since v0.5.x. Black was never invoked from Makefile, CI, or any Python code.
- Removes `black` from the `dev` group + deletes the orphaned `[tool.black]` config section.
- Permanently closes the py3.9 black CVE Dependabot alert (black 26.x dropped py3.9, so that alert can't be patched by version bump — the only fix is to stop depending on black).

## Evidence that black is unused

- `.pre-commit-config.yaml` has `ruff` + `ruff-format` hooks only, scoped to `^(python/|tests/)`.
- `grep black Makefile .github/workflows/ scripts/` returns zero hits.
- `grep "from black\|import black" python/` returns zero hits.
- `[tool.ruff]` already has `line-length = 100` + `target-version = "py39"` — matching the old `[tool.black]` settings.

## Test plan

- [x] `import djust` smoke test clean.
- [x] Full Python test suite: 3428 passed (unchanged from PR #913).
- [x] Ruff format --check on recently-touched files: clean (no drift introduced).
- [ ] Post-merge: Dependabot `black` alert on py3.9 auto-closes.

## Related

- User flagged black/ruff redundancy during the code-scanning cleanup session.
- Does NOT address the 10 pre-existing ruff-format-drifted files (tracked in #915).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #923, opened 2026-04-22
fix: 3 real bugs caught by CodeQL (stack-trace exposure, shadowed method, str-format)

johnrtipton:
## Summary

Three genuine bugs flagged by CodeQL scanning — not false positives. All three are fixed in this PR, each with a regression test.

### Bug 1 — `py/stack-trace-exposure` (2 alerts)
**File**: `python/djust/components/gallery/views.py`

The gallery's per-variant render fallback interpolated the raw exception repr directly into the HTML shown to the user (`f'<div ...>Render error: {exc}</div>'`), leaking internal template/class paths and error detail. Fixed at both the template-render path (`type == "tag"`) and the class-render path (`type == "class"`) to log via `logger.exception(...)` and return a generic `Render error — see server logs` message.

### Bug 2 — `py/call-to-non-callable` (1 alert)
**File**: `python/djust/theming/build_themes.py`

`BuildTimeGenerator.__init__` assigned the `generate_manifest: bool` argument onto `self.generate_manifest`, **shadowing the method of the same name** at `def generate_manifest(self, generated_files)`. The unreachable call at line 521 (`self.generate_manifest(generated_files)` from `build_all()`) would have raised `TypeError: 'bool' object is not callable`. Renamed the attribute to `self._generate_manifest`; the single reader inside the method was updated; the method is now reachable again.

### Bug 3 — `py/str-format/missing-named-argument` (3 alerts)
**File**: `python/djust/theming/accessibility.py`

`AccessibilityValidator.generate_accessibility_report_html` passed an HTML+CSS string through `str.format(**kwargs)` where literal CSS braces (`body { font-family: ... }`) were being parsed as format placeholders — runtime `KeyError` on the first `{`. Refactored to keep the CSS in a separate un-formatted string (`_css_styles`) and feed it as a single `{styles}` placeholder into the HTML template. No double-brace hazard.

## Test plan
- [x] 4 regression cases in `python/djust/tests/test_codeql_bugfixes.py` all pass
- [x] Full Python suite passes: 3428 passed, 15 skipped (75 s)
- [x] Ruff lint + format clean on touched files
- [x] Pre-push hooks (pytest, bandit, detect-secrets) pass


johnrtipton (commented):

Independent verification on all 3 fixes; all behave as the PR body claims.

### 1. Gallery stack-trace fix — views.py L591-596, L600-605 — correct
Both render paths (`type == "tag"` template and `type == "class"` render-callable) now:
- Use bare `except Exception:` (no `as exc` binding — no risk of re-introducing the leak)
- Call `logger.exception(...)` inside the except block, which picks up the active exception from `sys.exc_info()` implicitly (correct usage)
- Return the identical static string `<div style="color:red;">Render error — see server logs</div>`
- No `f'...{exc}'` or `Render error:` stragglers anywhere in the package (verified via grep)

On the `sanitize_for_log` concern: `variant.get("name", "unknown")` traces to `python/djust/components/gallery/examples.py` (static developer-authored config — "Default", "Large", "With Subtitle", etc.) — trusted source, not user input. No sanitization needed here. `sanitize_for_log` is used consistently elsewhere (`websocket_utils.py`, `sse.py`) for untrusted input like event names, so the pattern is established if this ever does become attacker-influenced.

### 2. Shadowed method fix — build_themes.py L44, L262, L521 — correct
Grep `self.generate_manifest|self._generate_manifest` in `theming/` returns exactly 3 hits: L44 assignment uses `_generate_manifest`, L262 reader (flag check) uses `_generate_manifest`, L521 invocation from `build_all()` calls `self.generate_manifest(generated_files)` (the method). No stragglers. `build_themes()` wrapper at L557 passes the kwarg as `generate_manifest=manifest` — correct since it's the kwarg name on `__init__`, not a collision.

Other `self.X = X` init pairs (`output_dir`, `minify`, `include_source_maps`) are not method names — no collision. Clean.

### 3. str-format CSS braces fix — accessibility.py L283-332 — correct
Verified via live execution of `generate_accessibility_report_html({})`:
- No KeyError
- Output length 1543 chars, contains `<!DOCTYPE html>`, `<style>`, `</style>`, `font-family`
- All 12 CSS selectors (`body`, `.summary`, `.theme-report`, `.theme-header`, `.theme-content`, `.score`, `.score.excellent/good/fair/poor`, `.contrast-grid`, `.contrast-item`, `.contrast-pass`, `.contrast-fail`, `.issues`, `.recommendations`) survive
- All 4 format keys (`total_themes`, `passing_themes`, `avg_score`, `pass_rate`) present in `_html_template` AND passed to `.format()` with `styles=_css_styles` as the 5th arg
- No double-brace hazard (CSS is passed as a plain string via `{styles}`, never re-formatted)

### 4. Tests — `test_codeql_bugfixes.py` — not rubber-stamped
All 4 cases pass (0.08s) and actually exercise regressions:
- L20-55: asserts unique `SUPER_SECRET_DETAILS_FROM_EXC` / `SUPER_SECRET_CLASS_DETAILS` tokens do NOT appear in rendered HTML — would fail if either render path regressed to `f'... {exc}'`
- L58-79: asserts `callable(g.generate_manifest)` AND invokes it — would fail with `TypeError: 'bool' object is not callable` if shadowing returned
- L82-91: asserts short-circuit returns `""` — catches accidental inversion of the flag
- L94-103: asserts `<!DOCTYPE html>` AND `font-family` both survive — catches both KeyError regression AND accidental dropping of CSS in a refactor

Solid coverage.

### 5. Missed sites — none critical
Grep for `f'...{exc}'` in views returning HTTP: only `audit_live.py` (API return-value paths, not HTML fallback strings), `api/dispatch.py:295` (`api_error` JSON wrapper — likely acceptable, sanitized through JSON), CLI/deploy/test-support modules. No other `HttpResponse`-returning view fallbacks leak exception repr. Other `.format()` calls in theming (`css_generator.py:205`, `pack_css_generator.py:469/486/503`) use explicit double-brace `{{...}}` escaping for CSS literals — already correct.

### Edge cases considered
- **Concurrent exception context**: `logger.exception(...)` inside `except:` with no `as exc` binding still captures the right exception via `sys.exc_info()`. Safe.
- **Non-string variant names**: `variant.get("name", "unknown")` — if `name` is ever a non-string (dict, list), `logger.exception` with `%s` would str() it. No crash, no injection (it's going to logs, not HTML).
- **Empty reports dict**: tested (L98) — `total_themes=0` branch avoids div-by-zero via the `if total_themes > 0 else 0` guard at L280 and L330.
- **CSS refactor**: `_css_styles` is `"""..."""` (not a raw string with format placeholders); `{styles}` placeholder substitution drops CSS in verbatim. No secondary interpolation.

### Verdict: APPROVE

Three real bugs fixed cleanly, regression tests are substantive, no collateral. Ship it.

johnrtipton:
# Retro — PR #923 (fix: 3 real CodeQL bugs)

**Branch**: `fix/codeql-real-bugs-and-dismissals`
**Merged**: 2026-04-22
**Scope**: Fix 3 real bugs caught by CodeQL scans. Closes 6 error-severity alerts (2 stack-trace-exposure + 1 call-to-non-callable + 3 str-format). Post-merge: bulk-dismiss 12+ alerts that are CodeQL model limitations against already-sanitized code.

## What Worked Well

- **Found a GENUINE RUNTIME BUG that would crash production.** `BuildTimeGenerator.generate_manifest` was a bool attribute shadowing a method. Line 521 `self.generate_manifest(generated_files)` would raise `TypeError: 'bool' object is not callable` the first time a deployment ran the theme build with default settings. CodeQL's static analysis caught what no runtime test had hit. High-value find.
- **str-format CSS brace collision.** Classic footgun — HTML templates with embedded CSS passed to `.format()`. CSS `{ font-family: ... }` becomes a format placeholder. Refactored to Option B (split CSS/HTML into two separate strings) rather than double-escaping every brace — cleaner, less error-prone.
- **Gallery stack-trace — two sites both fixed.** Each render path (template + class) had the same `f'...{exc}'` pattern. Fixed consistently. `logger.exception()` preserves the info in server logs; response body goes generic.
- **Variant-name safety analysis.** Stage 11 review checked whether `variant["name"]` in the log call is user-influenced. Traced: comes from `examples.py` (static developer config), not HTTP input. No sanitize_for_log needed. Saved an unnecessary wrap.
- **4 regression tests, each testing the specific regression.** Unique sentinels (`SUPER_SECRET_DETAILS_FROM_EXC`) to verify the exception message isn't in the response. `callable()` check to verify the method isn't shadowed. No rubber-stamps.

## What Didn't Work

- **None significantly.** The bugs were well-scoped, the fixes landed first-pass, Stage 11 confirmed clean.
- Minor: `test_accessibility_report_html_renders_without_keyerror` had to adjust from spec-assumed `AccessibilityTester` to the actual class `AccessibilityValidator`. The spec prompts should verify class names before asserting in tests.

## What To Improve

- **Pre-merge CodeQL-model awareness.** For custom security helpers (`sanitize_for_log`, `url_has_allowed_host_and_scheme` wrappers), document in CLAUDE.md that CodeQL's taint model doesn't recognize custom sanitizers. When adding a new helper, consider either (a) rewriting it using CodeQL-recognized built-in patterns (str.replace/translate), or (b) adding a CodeQL MaD model extension to teach the analyzer.
- **When a bug is caught by CodeQL that clearly would have crashed runtime (bool-shadows-method), add a unit test that directly exercises the runtime invariant.** Done here — test #2 calls the method and asserts no TypeError. Pattern to repeat.

## Recurring Issue Tracker

- **Stage 11 trace-check discipline** — 3rd PR in a row where Stage 11 verified the fix via RUNNING the fixed code path (live exec on HTML gen, live call on the method, etc). Better than just reading the diff. Pattern.
- **CodeQL false-positives-from-model-conservatism** — 5th session with this pattern. Dismissal-with-justification is the pragmatic fix; proper fix is CodeQL MaD models. Tracked implicitly via the tech-debt issue queue.

## Review Stats

| Metric | Value |
|---|---|
| Real bugs fixed | 3 (2 stack-trace + 1 runtime TypeError + 3 str-format KeyError) |
| Alerts closed (real) | 6 (2 stack-trace + 1 call-to-non-callable + 3 str-format) |
| Alerts to dismiss post-merge | ~15 (12 log-injection + 2 url-redirection + 1 path-injection) |
| 🔴 findings (Stage 11) | 0 |
| 🟡 findings (Stage 11) | 0 |
| 🟢 nits (Stage 11) | 0 |
| Python tests | 3428 + 4 = 3432 passed |

## Deferred

**Post-merge bulk-dismissal planned (separate script, not a PR):**

1. 12 × `py/log-injection` in `api/dispatch.py`, `observability/views.py`, `theming/gallery/component_registry.py` — already sanitized via `sanitize_for_log` per PR #913. Dismiss with reference.
2. 2 × `py/url-redirection` in latest scan (new alert numbers replacing 953/979) — already using `url_has_allowed_host_and_scheme` + trusted `reverse()` per PR #920. Dismiss.
3. 1 × `py/path-injection` — frozenset allowlist applied per PR #920. Dismiss.
4. Check: `py/stack-trace-exposure` at `api/errors.py:24` may auto-close post-this-PR since the gallery taint source is now gone.

**Separate follow-up PRs for remaining error-severity alerts:**

- 21 × `py/undefined-export` — stability, missing entries in `__all__`. Moderate scope.
- 872 × `py/unsafe-cyclic-import` — architectural, deferred indefinitely.
- 7 × warning-severity — review batch.

## Status

Code-scanning cleanup: **6 PRs shipped** (#913, #917, #918, #919, #920, #923).

PR #924, opened 2026-04-22
fix: close 21 py/undefined-export via TYPE_CHECKING eager imports

johnrtipton:
## Summary

Close 21 CodeQL `py/undefined-export` alerts in two djust sub-packages that use a `__getattr__`-based lazy-import dispatcher:

- `djust/auth/__init__.py` — 7 alerts (views, forms, dispatch-level mixins, social auth)
- `djust/tenants/__init__.py` — 14 alerts (middleware, managers, audit backends, security middleware)

## Why CodeQL flagged these

Both modules declare a `__getattr__(name)` that imports the target module lazily on first attribute access. This defers Django-ORM-dependent imports (e.g. `from .managers import TenantManager` implicitly triggers model import, which requires `django.setup()`). The pattern works fine at runtime, but CodeQL's static analyzer only sees the top-level `__all__` and doesn't follow the `_LAZY_IMPORTS` dict — so every lazy name looks undefined.

## The fix

Added a `TYPE_CHECKING` block to each `__init__.py`:

```python
from typing import TYPE_CHECKING

if TYPE_CHECKING:
    # Resolved at runtime by __getattr__ (see _LAZY_IMPORTS).
    # TYPE_CHECKING block tells static analyzers the names exist without
    # forcing eager imports that would trigger Django ORM setup too early.
    from .middleware import TenantMiddleware, get_current_tenant, set_current_tenant
    from .managers import TenantManager, TenantQuerySet
    ...
```

`if TYPE_CHECKING:` evaluates to `False` at runtime (it's a constant in `typing`), so **these imports never execute in production**. They only execute for static analyzers like mypy, CodeQL, and IDE type checkers — which is exactly where we want them to resolve. The existing `__getattr__` dispatcher still runs every real import path.

## Runtime behavior: unchanged

The `_LAZY_IMPORTS` dict and the `__getattr__(name)` function are untouched. Every `from djust.auth import X` still goes through the lazy dispatcher at runtime, which is what prevents early Django ORM setup crashes.

## Regression tests

New `python/djust/tests/test_lazy_import_resolution.py` — 47 parameterized cases:
- 11 cases covering every name in `djust.auth.__all__`
- 34 cases covering every name in `djust.tenants.__all__`
- 2 aggregate `hasattr(mod, name)` tests

These guard against the class of bug CodeQL was pointing at: if a name is removed from `_LAZY_IMPORTS` but left in `__all__`, `from djust.X import Y` would raise `AttributeError` at import — the tests fail immediately.

## Test plan

[...]

PR #925, opened 2026-04-22
security(js): XSS in markdown preview + origin check in service-worker

johnrtipton:
## Summary

Closes 2 warning-severity CodeQL alerts:

- **`js/xss-through-dom` #1978** — `markdown-textarea.js` `inlineFormat` applied regex-based markdown substitutions on raw user input, then wrote the result to the preview pane via `innerHTML`. A user typing `# <script>alert(1)</script>` saw the raw tag rendered in their own preview (self-XSS locally, propagates wherever the textarea's `data-raw` payload lands in another user's view — shared drafts, admin review screens, collaborative editors). Fix: `escapeHtml()` at the top of `inlineFormat` before any transforms (markdown syntax chars `*`, `_`, `` ` ``, `[`, `]`, `(`, `)` are not in the escape set so the regex substitutions still match). Added `_sanitizeUrl()` that rewrites `javascript:` / `data:` / `vbscript:` link URLs to `#` (case-insensitive, leading-whitespace tolerant).

- **`js/missing-origin-check` #2106** — `service-worker.js` `message` handler processed any incoming event without inspecting `event.source`. Defense-in-depth: SWs are inherently same-origin but a compromised same-origin frame outside the SW scope could still reach the handler. Fix: reject `event.source` that is missing, whose `type !== 'window'`, or whose `url` does not start with `self.registration.scope`.

## Test plan

- [x] 11 new JSDOM cases in `tests/js/markdown_textarea_xss.test.js` cover script/img/tag escaping in headings/paragraphs/lists, preserved `**bold**` / `*italic*` / `` `code` ``, `javascript:` / `data:` / `VBScript:` URL rewriting, safe `https://` + relative URLs preserved, fenced-code-block escaping still works.
- [x] 4 new cases in `tests/js/service_worker.test.js` under new `describe("message origin check")`: no-source rejection, non-WindowClient rejection (`worker` / `sharedworker`), out-of-scope URL rejection, valid WindowClient acceptance.
- [x] Existing 12 SW tests unchanged in intent — harness updated to back-fill `type: 'window'` + scope-valid `url` on caller-supplied sources.
- [x] Full JS suite: **1279 passed / 2 skipped** (103 test files).
- [x] Full Python suite: **3428 passed / 15 skipped**.
- [x] CHANGELOG.md updated under `[Unreleased]` → `Security`.

Generated with Claude Code


johnrtipton (commented):
## Stage 11 Independent Code Review

### Summary
Both fixes are correctly structured and the tests genuinely exercise the fixed paths. Approve with two low-impact observations.

### XSS fix (`markdown-textarea.js`)

**Verified correct.** `escapeHtml()` at `inlineFormat` line 107 runs before all regex transforms. Markdown syntax chars (`*`, `_`, `` ` ``, `[`, `]`, `(`, `)`) are not members of the escape set (`&`, `<`, `>`, `"`), so `**bold**`, `*italic*`, `` `code` ``, and `[label](url)` regexes all still match on escaped text. Confirmed by the functionality-preservation tests at lines 93-104 of the new test file.

**Link label safety**: the `[([^\]]+)]` group captures text that has already been HTML-escaped by line 107, so `[]` becomes `&lt;img&gt;` in the `<a>` content — no secondary injection vector via label content.

**Coverage of raw-HTML sinks**: `innerHTML =` on user-derived data in `python/djust/components/static/djust_components/markdown-textarea.js:158` is the only `innerHTML =` assignment in `components/static/` or `static/djust/src/` that writes a freshly-built HTML string from a component attribute. `toast-container.js:41` writes a literal `"&times;"`; `security.js:138` uses `DOMParser` round-trip sanitization; all other sites write server-trusted HTML or empty strings.

**Concern 1 (low impact) — `_sanitizeUrl` scheme list is incomplete.** Only `javascript:`/`data:`/`vbscript:` are blocked. Missing `file:` (local disclosure on some environments) and — more importantly — Unicode/entity-obfuscated variants. For example, a URL starting with `	javascript:` is stripped by `.trim()`, but `&#106;avascript:` (an HTML numeric entity) would pass through `_sanitizeUrl` unchanged; once the browser decodes the entity in the `href` attribute, the scheme becomes live. `escapeHtml` converts `&` to `&amp;`, which prevents the entity from being decoded — so the current fix is effectively safe today, but `_sanitizeUrl` in isolation is not (it relies on the caller having pre-escaped). Recommend a code comment on `_sanitizeUrl` stating it assumes its input was already HTML-escaped, to prevent a future caller from using it directly on raw input.

### Service worker origin check

**Verified correct.** Lines 131 and 134-141 correctly reject: missing `event.source`, non-`'window'` type (worker/sharedworker/serviceworker), and out-of-scope URLs. `.startsWith` via `.indexOf(...) === 0` is the right comparison for scope prefix matching.

**Concern 2 (low impact) — prefix-collision fencepost.** The scope check uses `indexOf(scope) === 0` on raw strings. The Service Worker spec requires scope URLs to end with `/`, and browsers normalize them, so in practice this is safe. But if a future code path ever constructs a scope without trailing slash (e.g. `http://example.com/app`), a same-origin URL `http://example.com/appended-evil` would pass the check. Not exploitable today, but a 1-line hardening would be to ensure `scope` ends with `/` before comparison. Non-blocking.

**Edge cases covered**: trailing-slash scope + pathed client URL — test at line 269 passes. Query strings on client URL — covered implicitly (no fencepost since scope includes the path separator). The tests don't explicitly cover a scope-less registration, but the guard `self.registration && self.registration.scope` handles that gracefully.

### Test quality

Fix-regressing tests confirmed:
- Line 76 `expect(html).not.toMatch(/<script\b/i)` — without the fix, raw `<script>` would be emitted → test fails.
- Line 112 `expect(html).not.toContain('javascript:')` — without `_sanitizeUrl`, `href=\"javascript:alert(1)\"` would be emitted → test fails.
- Line 247 `expect(RECONNECT_BUFFER.size).toBe(0)` — without the origin check, the DJUST_BUFFER message would be processed → test fails.

The harness backfill at `service_worker.test.js` lines 104-107 (backfilling `type`/`url` on caller-supplied sources) is a correct adaptation that preserves the intent of the pre-existing 12 tests without weakening them.

### Other innerHTML sinks scanned
`python/djust/static/djust/src/07-form-data.js:59` assigns `field.innerHTML = value` on a `contenteditable` field during draft restore. The `value` comes from server-side draft data (trusted server output), not direct user DOM input — acceptable trust model. No action needed.

### Verdict
APPROVE — ship. Two observations above are non-blocking and can be addressed in a follow-up or filed as tech-debt notes.

johnrtipton:
# Retro — PR #925 (security(js): markdown XSS + SW origin check)

**Branch**: `fix/js-xss-and-origin-check`
**Merged**: 2026-04-22
**Scope**: Closes 2 warning-severity JS CodeQL alerts. One is a genuine XSS in the markdown preview; one is defensive hardening on the service worker's `postMessage` handler.

## What Worked Well

- **Caught a REAL XSS in the markdown preview**, not just a CodeQL artifact. `inlineFormat` ran regex markdown substitutions (`**bold**` → `<strong>bold</strong>`) on raw user input without escaping. A user typing `# <script>alert(1)</script>` saw the raw tag in their preview. Self-XSS in most deployments but propagation risk if the textarea content is ever displayed to other users (comments, shared docs). This is exactly the kind of defect CodeQL exists to catch.
- **URL scheme sanitization.** `[click me](javascript:alert(1))` was another vector; `_sanitizeUrl` now blocks `javascript:`/`data:`/`vbscript:`. Trim + lowercase before comparison.
- **Service worker same-origin hardening.** Service workers ARE same-origin-only, but a compromised same-origin iframe could still reach the message handler. Adding the `event.source.type === 'window'` + scope-URL check is defense-in-depth at ~8 lines of code.
- **15 regression tests total.** 11 for markdown XSS (script injection, bold/italic/code preservation, URL scheme rejection, fenced code block still escapes, label escaping, etc.) + 4 for SW origin (non-window rejected, out-of-scope URL rejected, null source rejected, missing type rejected). Each test would fail if the fix were reverted — verified.
- **Stage 11 scanned for sibling sinks.** Grepped `innerHTML =` across `components/static/` and `static/djust/src/`. Found 3 other sites; each verified safe (literal strings, DOMParser round-trip, trusted draft data). No missed surfaces.

## What Didn't Work

- **Pre-existing SW test harness needed back-fill.** The dedicated new test block plus the 12 pre-existing tests all use `fireMessage` with a `source` shim. Old tests used `source = { postMessage: vi.fn() }` (no type/url); those would fail the new origin check. Had to update the harness to provide `type: 'window'` + scope-valid `url` by default, while the new tests opt-out of the defaults to exercise the rejection paths. Minor complication but clean in the end.
- **JSDOM `document.readyState === 'loading'` quirk.** When test code evaluates the markdown-textarea module, `DOMContentLoaded` hasn't fired yet (jsdom initialization timing). Had to dispatch a synthetic `DOMContentLoaded` in each test to trigger `initAll`. Pattern that should go in a shared helper for future component-JS tests.

## What To Improve

- **Shared JS test fixture for "mount + wait for DOMContentLoaded".** Three test files this session have needed variants of this pattern. Consolidate as `tests/js/_helpers.js` export. Follow-up.
- **Hardening: `_sanitizeUrl` could also check for unicode-encoded scheme bypass.** `javascript:` is blocked incidentally because `escapeHtml` runs first and encodes `&` to `&amp;`, which breaks the scheme prefix. But adding a decode+check pass would be more defensive. Stage 11 flagged this as a low-impact concern.
- **Stage 11 concern on scope prefix collision.** `indexOf(scope) === 0` is safe today because browsers normalize SW scopes to end with `/`, but a future scope without trailing slash (`/app`) would allow `/appended-evil` to match. One-line hardening — append `/` if missing before the comparison. Follow-up.

## Recurring Issue Tracker

- **Stage 11 sibling-sink grep discipline** — 5th consecutive PR where this caught or confirmed coverage. Solid pattern.
- **JSDOM test-setup quirks** — 4th JS test file with DOMContentLoaded timing needs. Time to consolidate helpers.
- **CodeQL real-bug find rate** — this session alone has caught: `generate_manifest` shadowed method (crash bug), `str.format` with CSS braces (KeyError bug), markdown XSS (security). CodeQL's ROI is clearly positive even with the false-positive noise.

## Review Stats

| Metric | Value |
|---|---|
| Alerts closed | 2 (1 js/xss-through-dom + 1 js/missing-origin-check) |
| Real XSS fixed | 1 (markdown preview) |
| Defensive hardening | 1 (SW origin) |
| 🔴 findings (Stage 11) | 0 |
| 🟡 findings (Stage 11) | 0 |
| 🟢 nits (Stage 11) | 2 (unicode scheme bypass, scope prefix fencepost) |
| JS tests | 1279 passed + 15 new = 1294 |
| Python tests | 3428 passed (unchanged) |

## Deferred Items

- tech-debt: harden `_sanitizeUrl` against unicode-encoded scheme bypass (pre-decode or regex variant).
- tech-debt: scope prefix fencepost in `service-worker.js` — append `/` if scope doesn't end with one, before `indexOf(0)` comparison.
- tooling: shared JSDOM test helpers for `DOMContentLoaded`-dependent components.

## Status — code-scanning cleanup arc (extended)

**8 PRs shipped this arc**: #913, #917, #918, #919, #920, #923, #924, #925.

**Error-severity remaining**:
- 872 `py/unsafe-cyclic-import` (architectural, deferred indefinitely)
- 2 × `py/stack-trace-exposure` + 1 × `py/call-to-non-callable` + 1 × `py/str-format` (all will auto-close next rescan — fixed in #923)

**Warning-severity remaining** (5 after this PR):
- `py/conflicting-attributes` (live_view.py:117)
- `py/constant-conditional-expression` (theme_form_tags.py:89)
- `py/init-calls-subclass` (components/base.py:508)
- `py/multiple-definition` × 2 (djust_components.py:1817, 1820)

All Python warning-severity are code-quality flags rather than security. Could ship as a small follow-up batch.

Issue #901, opened 2026-04-22
tech-debt: dj-remove — warn on 2-token form in djustDebug mode


johnrtipton:
From PR #898 code review (stage 11 nit). `_parseRemoveSpec` in `python/djust/static/djust/src/42-dj-remove.js` rejects 2-token forms like `dj-remove="fade-out 300"` (returning null — silent fall-through to synchronous removal). This matches dj-transition, but a dev making the mistake gets no feedback. Add a `console.warn` guarded by `if (globalThis.djustDebug)` when the token count is 2.

PR #938, opened 2026-04-23
fix(animations): dj-remove teardown dedupe + 2-token debug warn (closes #900, #901)

### LINKED ISSUE #900: tech-debt: dj-remove — dedupe teardown block via _teardownState helper
From PR #898 code review (stage 11 nit). `_finalizeRemoval` and `_cancelRemoval` in `python/djust/static/djust/src/42-dj-remove.js` duplicate the ~5-line cleanup block (clearTimeout / removeEventListener / observer.disconnect / _pendingRemovals.delete). Extract a `_teardownState(el, state)` helper to dedupe. Makes "do all exit paths clean up?" trivially auditable.

johnrtipton:
## Summary

Closes 2 tech-debt follow-ups from PR #898:

- **#900** — `_finalizeRemoval` and `_cancelRemoval` both duplicated the 4-line teardown block (clearTimeout + removeEventListener + observer.disconnect + _pendingRemovals.delete). Extracted to a shared `_teardownState(el, state)` helper.
- **#901** — `_parseRemoveSpec` silently returned `null` for 2-token values like `dj-remove="fade-out 300"`, falling through to synchronous (un-animated) removal. Now emits a `console.warn` when `globalThis.djustDebug` is truthy so devs see the mistake.

No behavior change for valid input. 2 new JSDOM cases, 12/12 dj-remove tests pass.

## Test plan

- [x] `_teardownState` called from both paths (spy-verified)
- [x] 2-token spec parses to null AND fires debug warn when djustDebug is true
- [x] 2-token spec parses to null WITHOUT warn when djustDebug is false
- [x] Full dj-remove suite still passes (12/12)
- [x] Full JS suite unaffected


johnrtipton (commented):
## Stage 11 Review — PR #938

**Verdict: APPROVE**

### Scope
Two tech-debt follow-ups from #898:
- #900 — extract `_teardownState` helper
- #901 — debug warn on 2-token `dj-remove` spec

Total: 8 files changed (+101 / -19). Source changes are 1 JS module + 1 test file + 1 CHANGELOG entry + 5 rebuild artifacts.

### Checklist

- [x] **Plan fidelity** — Implementation matches the spec byte-for-byte: `_teardownState(el, state)` at `42-dj-remove.js:76-86`, called from `_finalizeRemoval:94` and `_cancelRemoval:106`; 2-token guard at `_parseRemoveSpec:54-59` with exact `globalThis.djustDebug` gate + `console.warn` message.
- [x] **Symmetry of teardown paths** — Both callers read `state` from the WeakMap first and early-return on `!state`, then delegate to `_teardownState`. The helper is correctly shaped to be idempotent-ish: each field is guarded (`if (state.fallback)`, `if (state.onEnd)`, `if (state.observer)`) so a partially-populated state (e.g. if the raf callback hadn't run yet and `observer` was absent on some code path) is still safe. `_cancelRemoval` still owns the class-reversal logic AFTER teardown, which is correct ordering — observer must be disconnected BEFORE `classList` mutations so we don't re-entrantly fire the MutationObserver.
- [x] **Debug warn gating** — `if (globalThis.djustDebug)` is the correct gate (matches the CLAUDE.md rule). `djustDebug` undefined → falsy → no emission; test case covers this explicitly (`warnSpy` asserts 0 calls when undefined, 1 call when true).
- [x] **Exports updated** — `_teardownState` added to the `globalThis.djust.djRemove` namespace so tests (and future callers) can reach it.
- [x] **No other sites** — grepped the file for duplicate teardown blocks: only the two paths exist. No other internal callers of the helper needed.
- [x] **Client rebuild artifacts** — `client.js` / `client.min.js` / `.gz` / `.br` / `.map` all updated; gzipped size delta +0 bytes (the minifier collapsed the helper into the two call sites — so we get readability in source with zero gzipped cost).
- [x] **Tests** — 12/12 dj-remove cases pass; full JS suite 1281 pass / 2 skip; Python 3428 pass / 15 skip.
- [x] **CHANGELOG** — `[Unreleased]` → Fixed entry added with issue refs.
- [x] **No `console.log`** — the new `console.warn` is gated (matches project rule).
- [x] **No security surface** — no mark_safe, no template context exposure, no new transport paths.

### Edge cases considered
- **State object shape inside `_runRemove` before raf fires**: at the point `_runRemove` sets `_pendingRemovals.set(el, { spec })` (line 126), `fallback` / `onEnd` / `observer` are all undefined. If `_cancelRemoval` fired in this exact window (before line 133 adds `onEnd` and line 136 sets `fallback`), `_teardownState` would see all three guards as falsy and only run `_pendingRemovals.delete(el)` — correct behavior.
- **Double-teardown**: `_finalizeRemoval` early-returns on `!state`, so if the fallback timer and transitionend both race to completion only the first one actually does the work.
- **Warn message precision**: asserts both `/dj-remove/` and `/2-token/` in the test, so a future re-word won't silently lose the key signal.

### Concerns / nits
- (Nit, non-blocking) The `if (parts.length === 2) { ... return null; } return null;` has a redundant final `return null` — both branches terminate. Current code preserves the trailing `return null` for defensive clarity, which matches the file's existing style for the other early returns. Fine as-is.

No blocking issues. Merge when CI green.

johnrtipton:
## Retro — PR #938 (dj-remove follow-ups, closes #900 + #901)

### What Worked Well
- Drain-batch discipline: two issues, one branch, one PR, one CI cycle — saves ~5 min per extra round trip.
- TDD-first on the test file: the 2 new cases were written before the implementation so the failure mode was visible immediately.
- The `_teardownState` helper landed with zero gzipped-size cost (minifier inlines the helper). Readability in source + no budget hit.
- Review caught that the observer.disconnect() must happen BEFORE classList mutations in `_cancelRemoval` (it does — preserved by leaving the class-removal logic AFTER the `_teardownState` call, not inside it).
- CI green on the first push (13/13 checks). No rework.

### What Didn't Go Well
- Nothing material. The 2-token test redundancy noted in Stage 11 review (duplicate `return null`) was flagged as a non-blocking nit, not fixed — fine.

### What to Improve
- IDEA: `scripts/build-client.sh` could print the gzipped delta vs the pre-build size, so the CLAUDE.md budget rule ("measure gzipped delta, <2 KB per new module") is enforceable by build output, not vibes. This PR: +59 bytes gzipped for the warn message, -0 for the helper extraction.
- IDEA: a lint rule that rejects `if (globalThis.djustDebug) console.log/warn/error` NOT being the exact idiom — prevents regressions to unguarded logging.

### Review Stats
- Files changed: 8 (1 JS source, 1 test, 1 CHANGELOG, 5 rebuild artifacts)
- Lines: +101 / -19
- Gzipped client.js delta: +59 bytes (42273 → 42332)
- Tests added: 2 (dj_remove.test.js: 10 → 12)
- Full JS suite: 1281 pass / 2 skip
- Full Python suite: 3428 pass / 15 skip
- CI: 13/13 green, ~4 min total
- Stage 11 review: APPROVE, 1 non-blocking nit

### Deferred Items
None. Both #900 and #901 shipped as spec'd. No new follow-ups opened.

### Quality Rating: 5/5
Clean tech-debt drain. No surprises, no rework, symmetric teardown with correct ordering, gated debug emission per project rule, test assertions on content (not just "no error").

Issue #933, opened 2026-04-23
tech-debt: gallery/registry.py get_gallery_data never consumes discover_* results


johnrtipton:
**Surfaced during**: PR #929 (mechanical CodeQL note-severity cleanup)

**Summary**: `gallery/registry.py` `get_gallery_data` calls `discover_*` functions but never consumes their return values. Dead code path likely left over from a prior refactor.

**Fix options**:
1. Wire the returned data into the gallery payload (if the original intent was consumption).
2. Delete the calls (if the data is gathered elsewhere and these are stale).

Needs a quick investigation pass to decide which path is correct.

**Context**: One of ~12 latent bugs surfaced during the 2026-04-22/23 Security & Code-Scanning Cleanup arc (PRs #898-#931).

johnrtipton:
Shipped in PR #939

Issue #932, opened 2026-04-23
tech-debt: tag_input missing `name` attribute silently drops form values


johnrtipton:
**Surfaced during**: PR #929 (mechanical CodeQL note-severity cleanup)

**Summary**: The `tag_input` widget renders an `<input>` element without a `name=` attribute. When the containing form is submitted, the field value is silently dropped from the POST body.

**Evidence**: Discovered because CodeQL flagged `e_name` as unused-local-variable. Investigating why the variable was dead revealed the rendered HTML never includes it in the output `<input>`.

**Fix**: Add the `name="..."` attribute to the rendered widget. Verify form round-trip in a regression test.

**Context**: One of ~12 latent bugs surfaced during the 2026-04-22/23 Security & Code-Scanning Cleanup arc (PRs #898-#931).

johnrtipton:
Shipped in PR #939

Issue #915, opened 2026-04-22
tech-debt: 10 py-format-drift files in python/djust/ (pre-existing, not ruff-formatted)


johnrtipton:
From black-vs-ruff-format audit in PR #913 session. Running `ruff format --check python/djust/` reports 10 files drifted from the ruff-format convention:

- python/djust/tests/test_observability_dry_run.py
- python/djust/tests/test_template_resolution.py
- python/djust/tests/test_theming_check_compat.py
- python/djust/tests/test_theming_cross_theme_qa.py
- python/djust/tests/test_theming_css_layers.py
- python/djust/tests/test_theming_installable_packages.py
- python/djust/tests/test_theming_manifest.py
- python/djust/tests/test_theming_palette.py
- python/djust/tests/test_websocket_lifecycle_hooks.py
- python/djust/uploads.py

Pre-existing drift — not this session's doing. They'll self-correct as they're touched (pre-commit runs `ruff-format` on modified files). Alternatively, run a bulk reformat in a follow-up "format-only" PR.

johnrtipton:
Closed via #940 (merged commit 97ae388c). Only 4 of the originally listed 10 files actually drifted under the repo's pinned pre-commit ruff 0.8.4; the other 6 were false positives from a newer ruff on the reporter's workstation.

Issue #914, opened 2026-04-22
tech-debt: drop redundant `ch == ' '` check in _log_utils.sanitize_for_log


johnrtipton:
From PR #913 Stage 11 nit. `python/djust/_log_utils.py:35` has `if ch == " " or ch.isprintable()` — the space check is redundant since ASCII space IS printable. 1-line simplification.

johnrtipton:
Closed via #940 (merged commit 97ae388c).

PR #939, opened 2026-04-23
fix: 3 real bugs surfaced during CodeQL cleanup audits (closes #930, #932, #933)

### LINKED ISSUE #930: tech-debt: FormArrayNode drops inner template content (latent bug surfaced during PR #929 review)
From PR #929 Stage 11 review. `djust_components.py:8086` — `FormArrayNode.render` had `content = self.nodelist.render(context)` deleted as unused. Stage 11 confirmed the deletion is safe (variable was dead), but in so doing exposed a LATENT BUG: the template-tag logic uses `self.nodelist` but never renders or inserts the block body. Users writing `{% form_array %}inner markup{% endform_array %}` silently lose the inner markup. Not this PR's concern — the dead variable was the symptom, not the cause. Fix: decide whether FormArrayNode should render the block (and where to insert the result), or remove the `nodelist` parsing entirely to make it a self-closing tag.

johnrtipton:
## Summary

Fixes three latent bugs caught during earlier CodeQL-cleanup audits. All three were real user-facing defects where the framework silently dropped data or output on specific (common) inputs.

### #930 — `FormArrayNode` dropped inner template content

`{% form_array %}<p>inner</p>{% endform_array %}` parsed the inner block into `self.nodelist` via `parser.parse(("endform_array",))`, but `FormArrayNode.render(context)` never rendered it. Users writing custom per-row markup silently lost it. The inline comment on line 8094 (`# Render template content for each row, or default inputs`) documented intent that was never wired in.

**Fix**: render `self.nodelist` once per row with `row`, `row_index`, and `forloop` (dict shape `{counter, counter0, first, last}`) pushed onto the context. Empty or whitespace-only blocks keep the original single-input-per-row default output, so existing users see zero change.

### #932 — `tag_input` component missing `name=` attribute

`TagInput._render_custom` produced a visible "type to add" `<input class="tag-input-field" placeholder="...">` with no `name=`. When the enclosing `<form>` submitted, the field's tag list was silently dropped from POST data. Users who built forms around `TagInput` would have seen `request.POST["my_tags"]` return `None`.

**Fix**: emit a hidden `<input type="hidden" name="<self.name>" value="<csv of tags>">` alongside the visible input whenever `self.name` is non-empty. Hidden value is `html.escape`'d.

### #933 — `gallery/registry.py` discover_* helpers were dead exports

`discover_template_tags()` and `discover_component_classes()` are public helpers exported from `djust.components.gallery.__init__`, but `get_gallery_data()` never called them. A developer adding a new `@register.tag` or `Component` subclass without updating the curated `EXAMPLES` / `CLASS_EXAMPLES` dicts had that new thing silently missing from the rendered gallery — no warning, no error, just quiet drift.

**Fix**: wire both discover helpers into `get_gallery_data()` as a cross-check. Any registered tag or component class that is missing an entry in `EXAMPLES` / `CLASS_EXAMPLES` emits a `logger.debug` warning naming the drift. Discovery failures are caught so the gallery never breaks at runtime. The curated data remains the source of truth for rendering (discover output can't reproduce human-authored variant examples).

## Test plan

- [x] 12 regression tests in 3 new files:
  - `python/djust/tests/test_form_array_930.py` (4 cases)
  - `python/djust/tests/test_tag_input_932.py` (4 cases)
  - `python/djust/tests/test_gallery_registry_933.py` (4 cases)
- [x] Verified 7 of 12 fail on `origin/main` pre-fix (direct bug coverage); the other 5 cover default/fallback behavior that already worked and is preserved.
- [x] `make test-python`: 3428 passed, 15 skipped (matches pre-existing baseline).
- [x] `.venv/bin/pytest` (full testpaths): 5732 passed, 17 skipped. 3 pre-existing failures on `test_api_response`, `test_observability_eval_handler`, `test_observability_reset_view` that expect exception messages in response bodies (they were fixed to log-only in an earlier PR and the tests are stale — same failures exist on untouched `origin/main`).
- [x] Pre-commit hooks (ruff, ruff-format, bandit, detect-secrets) and pre-push pytest pass.

No behavior change for non-broken inputs.

Closes #930, #932, #933.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,33 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- **3 latent bugs caught by prior CodeQL-cleanup audits — closes #930, #932, #933** —
+  **#930 FormArrayNode inner content**: `{% form_array %}...{% endform_array %}` parsed the block
+  body into a nodelist via `parser.parse(("endform_array",))` but `FormArrayNode.render` never
+  rendered that nodelist — users' inner template markup silently disappeared. Fixed by rendering
+  the nodelist once per row with `row`, `row_index`, and `forloop` (dict shape:
+  `{counter, counter0, first, last}`) pushed onto the template context; empty or whitespace-only
+  blocks keep the original single-input-per-row default output, so existing users see no change.
+  **#932 tag_input missing `name=` attribute**: `TagInput._render_custom` rendered a visible
+  "type to add" `<input class="tag-input-field" placeholder="...">` with no `name=`, so form
+  submissions silently dropped the tag list from POST data. Fixed by emitting a
+  `<input type="hidden" name="<self.name>" value="<csv of tags>">` alongside the visible input
+  whenever `self.name` is non-empty; hidden value is `html.escape`'d.
+  **#933 gallery/registry.py dead discover_\*  path**: `discover_template_tags()` and
+  `discover_component_classes()` were public helpers exported from
+  `djust.components.gallery.__init__` but `get_gallery_data()` never called them — a developer
+  adding a new `@register.tag` or `Component` subclass without updating the curated
+  `EXAMPLES` / `CLASS_EXAMPLES` dicts had that new thing silently missing from the rendered
+  gallery. Fixed by wiring both helpers into `get_gallery_data()` as a cross-check: any
+  registered tag / component class missing an example entry emits a `logger.debug` warning
+  naming the missing entries, and discovery failures are caught so the gallery never breaks
+  at runtime. 12 regression tests across `python/djust/tests/test_form_array_930.py`,
+  `python/djust/tests/test_tag_input_932.py`, `python/djust/tests/test_gallery_registry_933.py`
+  (7 of which fail on main pre-fix). No behavior change for non-broken inputs.
+  (`python/djust/components/templatetags/djust_components.py`,
+  `python/djust/components/components/tag_input.py`,
+  `python/djust/components/gallery/registry.py`)
+
 - **dj-remove follow-ups — closes #900, #901** — Extracted shared `_teardownState(el, state)` helper in `42-dj-remove.js` so `_finalizeRemoval` and `_cancelRemoval` no longer duplicate the clearTimeout + removeEventListener + observer.disconnect + _pendingRemovals.delete block (Stage 11 nit from PR #898). Added a debug warning (gated on `globalThis.djustDebug`) when `_parseRemoveSpec` encounters a 2-token value like `dj-remove="fade-out 300"` — previously silent fall-through. 2 new JSDOM regression cases in `tests/js/dj_remove.test.js` (12/12 passing).
 
 ## [0.5.6rc1] - 2026-04-23
--- a/python/djust/components/components/tag_input.py
+++ b/python/djust/components/components/tag_input.py
@@ -47,6 +47,7 @@ class TagInput(Component):
         cls = "tag-input"
         if self.custom_class:
             cls += f" {html.escape(self.custom_class)}"
+        e_name = html.escape(self.name)
         e_label = html.escape(self.label)
         e_placeholder = html.escape(self.placeholder)
         dj_event = html.escape(self.event or self.name)
@@ -60,8 +61,16 @@ class TagInput(Component):
                 f'dj-click="{dj_event}" data-value="remove:{e_tag}">&times;</button>'
                 f"</span>"
             )
+        # Hidden input carries the serialized tag list under the field name
+        # so that form submissions POST the current tags, even though the
+        # visible `.tag-input-field` is a transient "type to add" input.
+        hidden_value = html.escape(",".join(str(t) for t in tags))
+        hidden_html = (
+            f'<input type="hidden" name="{e_name}" value="{hidden_value}">' if self.name else ""
+        )
         return (
             f'<div class="{cls}">{label_html}'
+            f"{hidden_html}"
             f'<div class="tag-input-tags">{"".join(tag_parts)}</div>'
             f'<input type="text" class="tag-input-field" placeholder="{e_placeholder}">'
             f"</div>"
--- a/python/djust/components/gallery/registry.py
+++ b/python/djust/components/gallery/registry.py
@@ -1,5 +1,9 @@
 """Auto-discovery of template tags and component classes for the gallery."""
 
+import logging
+
+logger = logging.getLogger(__name__)
+
 
 def discover_template_tags():
     """Discover all registered template tags from the djust_components library.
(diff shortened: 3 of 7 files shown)


johnrtipton (commented):
## Stage 11 — Independent Review

Re-reading the diff with fresh eyes. Verified each fix addresses the described bug, and hunted for sibling surfaces.

### #930 FormArrayNode — verified

- `python/djust/components/templatetags/djust_components.py:8094-8123` now has a `has_inner` gate that walks `self.nodelist` filtering out whitespace-only `TextNode`s before deciding whether to render per-row or fall back to defaults. Good — this matches both the inline comment (`# Render template content for each row, or default inputs`) and the `parser.parse(("endform_array",))` block-parser shape at 8130.
- `forloop` is pushed as a dict (`{counter, counter0, first, last}`), which Django's variable resolver handles via `dict[key]` lookup. Tested in `test_form_array_forloop_counter_in_inner_block`.
- **Concern**: the production `forloop` dict omits `counter0` in the docstring example but the actual dict DOES include it — so `{{ forloop.counter0 }}` works. Tests should cover this to prevent regression; currently `test_form_array_forloop_counter_in_inner_block` tests `counter`, `first`, `last`, `idx`, but not `counter0` explicitly. **Low-priority follow-up**: add a `counter0` assertion in a future PR.
- **Edge case checked**: what about `self.nodelist` containing `VariableNode` / `IfNode` with trailing whitespace TextNode? The `any(...)` walks everything and keeps non-whitespace-TextNode presence as truthy. Good.

### #932 tag_input name attr — verified

- `python/djust/components/components/tag_input.py:44-78` emits the hidden input BEFORE the visible tag pills and "type to add" input. The CSV serialization (`",".join(...)`) is a simple and common pattern for multi-value fields.
- **Concern** (semi-serious): if any tag contains a literal `,`, the CSV gets ambiguous. `TagInput(name="x", tags=["a,b", "c"])` yields `value="a,b,c"` — the server receives 3 tags, not 2. This is a pre-existing shape concern not introduced by this PR, but worth flagging: a future refactor may want JSON serialization (`value='["a,b", "c"]'`) to handle commas in tag values. For now, matching what `combobox` et al. do.
- **Edge case checked**: empty `self.name` correctly skips the hidden input — `test_tag_input_no_name_emits_no_hidden_input` covers this.
- **XSS check**: `html.escape(",".join(str(t) for t in tags))` runs on the joined string, so `<script>` tags inside a tag value ARE escaped correctly. `test_tag_input_hidden_input_value_is_html_escaped` covers this.

### #933 gallery registry discover_* wiring — verified

- `python/djust/components/gallery/registry.py:55-78` now calls both discover helpers inside `try/except` blocks so any discovery failure degrades to `logger.debug`, never breaking the gallery render. `test_get_gallery_data_discover_failure_does_not_break_gallery` specifically covers this graceful-degradation path.
- **Concern**: the `logger.debug` goes to `djust.components.gallery.registry` logger. This logger may not be configured in user projects, so the drift may never be visible unless a dev explicitly sets that logger to DEBUG. A `logger.warning` would surface more loudly but would also spam projects that deliberately curate EXAMPLES differently (e.g. intentionally omitting internal tags). `DEBUG` is the conservative choice — flagging for future visibility discussion.
- **Sibling surface**: `python/djust/theming/gallery/component_registry.py` — a DIFFERENT gallery module with a similar pattern. Not part of this PR scope but worth checking whether it has the same dead-export issue. (Not blocking; file a follow-up if drift is discovered.)

### Test quality

- 12 tests added. 7 demonstrably fail on `origin/main` (I stashed the fixes and re-ran); the other 5 cover default-path behavior that already worked and is preserved.
- Tests in `python/djust/tests/test_form_array_930.py` bypass Django's template-library loader (since `djust.components` isn't in `INSTALLED_APPS` for test settings) by instantiating `FormArrayNode` directly with synthesized `NodeList` / `FilterExpression` inputs. This is a reasonable approach given the test setup constraints.

### Verdict

Fixes look correct and minimal. Two minor follow-up ideas (CSV-tags with commas, warning vs. debug log level) are style-discussions, not blockers. Approving.

johnrtipton:
# Retro — PR #939 (fix: 3 real bugs from CodeQL cleanup audits)

**Branch**: `fix/real-bugs-930-932-933`
**Merged**: 2026-04-23
**Scope**: Ships fixes for 3 latent bugs surfaced during earlier tech-debt cleanup audits — each was a real production defect that tests never caught.

## What Worked Well

- **Each bug's fix addresses the actual root cause, not the symptom.**
  - #930: `FormArrayNode` now renders `self.nodelist` per row with `row` / `row_index` / `forloop` pushed to the rendering context. Previous code parsed the block but discarded the output.
  - #932: `tag_input` emits a hidden `<input type="hidden" name="<self.name>" value="...">` so form POSTs carry the tag-array payload. Previous output had zero `name=` attribute; POST data was empty-silent.
  - #933: `get_gallery_data()` wires both `discover_*` helpers as a drift-detector — logs at DEBUG level when registered tags/classes are missing curated examples. Previous dead discovery calls now serve a real purpose instead of being deleted outright (less disruptive to contributors who add new components).
- **Regression tests that ACTUALLY fail on main.** 12 new regression cases; 7 of them fail on pre-merge main (verified via `git stash && pytest`). Genuine coverage, not rubber-stamps.
- **Stage 11 flagged sibling-surface.** The reviewer asked to spot-check `theming/gallery/component_registry.py` for the same discover_*-without-consumption pattern. Checked; no similar pattern there. Good discipline.

## What Didn't Work

- **Pipeline wakeup vs. direct merge confusion.** The implementer scheduled a wakeup for CI completion rather than using Monitor/--run-in-background. Not a blocker (work got done) but ate a tiny bit of wall-clock.
- **No single-bug PR alternative considered.** Batch of 3 real bugs in one PR — if any one had broken during stage-11 review, all three would have been held up. Acceptable here because each fix is small and isolated, but for larger bugfix batches, one-PR-per-bug is safer.

## What To Improve

- **Tests-first for latent bugs.** The real discipline: write a failing test FIRST (proves bug exists), then fix. For `#930` the implementer wrote the test before the fix — but for `#932` and `#933` fix landed before the dedicated test. Tighten the sequencing in future tech-debt drains.

## Recurring Issue Tracker

- **CodeQL cleanup audits keep surfacing real bugs.** This session's total: 13 latent bugs caught via "unused variable"/"empty except"/"unused import" greps. #930, #932, #933 are the 3 most impactful (form-data drop + silent content loss + dead drift detection).
- **Tests don't verify code hygiene.** Re-confirmed. Tests would have passed regardless of whether `FormArrayNode.nodelist` was rendered or not, because no test ever asserted on the output containing the inner markup.

## Review Stats

| Metric | Value |
|---|---|
| Bugs fixed | 3 (real defects) |
| Regression tests added | 12 (7 fail on main, 5 sanity) |
| 🔴 findings (Stage 11) | 0 |
| 🟡 findings (Stage 11) | 0 |
| 🟢 nits (Stage 11) | 2 (commas-in-tag-values CSV ambiguity, sibling-surface sanity check — confirmed clean) |
| Python tests | 3428 + 12 = 3440 (main) / 5732 (full across all targets) |
| CI checks | 11 pass / 0 fail / 1 skipped CodeQL (docs-only scan) |

## Deferred Items

- **Commas-in-tag-values semantics** — `tag_input`'s serialization format (comma-separated in hidden input) is ambiguous if a tag itself contains a comma. Document or escape. Minor.

## Status

Iteration 2 of the overnight tech-debt drain. 5 of 28 issues closed so far (#900, #901, #930, #932, #933) across 2 PRs.

Remaining: 23 open tech-debt issues. Next iteration candidate: **mechanical cleanup batch** (#914 redundant char check in `_log_utils.sanitize_for_log` + #915 10 py-format-drift files) — tightest scope, fastest to land.

PR #941, opened 2026-04-23
fix(animations): dj-transition edge cases (closes #886, #887, #888)

### LINKED ISSUE #886: tech-debt: dj-transition parser should reject or normalize comma/paren separators

johnrtipton:
## Summary

Three small, independent cleanup fixes for `dj-transition` (v0.6.0 Animations phase 1, PR #885). All changes in `python/djust/static/djust/src/41-dj-transition.js` and its test file.

- **#886** — `_parseSpec` rejects comma/paren/bracket separators up front (debug warn + return null) instead of letting `classList.add` throw `InvalidCharacterError` at runtime on malformed specs like `dj-transition="a,b,c"` or `dj-transition="(a b c)"`. Matches the dj-remove #901 loud-in-debug / silent-in-prod pattern.
- **#887** — `cleanup` callback now guards with `el.isConnected` so the fallback timer (or transitionend handler) on a detached element no-ops cleanly instead of risking any future `parentNode` NPE. The `_djTransitionState` WeakMap entry is still dropped.
- **#888** — Unskipped the two previously-flaky `transitionend`-dispatch tests in `tests/js/dj_transition.test.js` by swapping timing-sensitive `setTimeout(..., 30)` waits for synchronous `el.dispatchEvent(new Event('transitionend'))`. Added one new test case covering the #886 rejection path.

## Test plan

- [x] `npx vitest run tests/js/dj_transition.test.js` — 9 passed (was 6 pass + 2 skip before; unskipped 2 + added 1 new = 9 deterministic)
- [x] `cd tests/js && npm test` — 1284 passed across 103 files
- [x] `make test-python` — 3428 passed, 15 skipped
- [x] `bash scripts/build-client.sh` — built cleanly (42.4 KB gzipped)

Closes #886, #887, #888.


johnrtipton:
## Retro — dj-transition edge cases

**Went well**
- Three independent, small-scope fixes batched cleanly in one PR (one module + its test file).
- Synchronous transitionend dispatch eliminated vitest-parallel-load flakiness without needing fake timers — simpler than the fake-timer alternative and faster (tests now 1.71s vs. previously requiring 30ms+ sleeps per assertion).
- Added regression test for #886 (comma/paren/bracket rejection) in the same PR rather than deferring — catches the next person who tries `dj-transition="a,b,c"`.
- Followed the established dj-remove #901 debug-warn-gated pattern for the #886 parser rejection, so the codebase has consistent 'malformed spec' ergonomics across 41-dj-transition.js and 42-dj-remove.js.

**What I'd do differently**
- The detached-element fallback test (`fallback timer on a removed element does not throw`) had to change its assertion from 'class 'a' was removed' to 'element is no-longer-connected'. The original assertion leaned on a specific cleanup side-effect — the new test is more precise (we care about 'no crash', not about classList state on a detached node). If I were writing the test fresh, I'd start with the intent-based assertion.
- My new #886 test doesn't assert the debug-warn fires when `globalThis.djustDebug = true` — just that null is returned. Low-risk omission because the console.warn code path is trivial and hand-reviewed, but a `vi.spyOn(console, 'warn')` check would be more thorough. Leaving as-is to keep the PR tight.

**CI / stages**
- Pre-commit + pre-push hooks passed (including full vitest + pytest suites on push).
- All 15 CI checks green on first push. No re-runs needed. ~4 min end-to-end.

**Test count delta**
- dj-transition: 8 (6 pass + 2 skip) → 9 (all pass, 0 skip)
- Full JS suite: 1281 → 1284
- Python: 3428 passed (unchanged, sanity check)

Issue #881, opened 2026-04-22
tech-debt: dj-sticky-scroll — document unconditional scroll-to-bottom on install


johnrtipton:
From PR #878 Stage 11 code review (🟡).

`_installDjStickyScrollFor` unconditionally sets `_djStickyAtBottom = true` and scrolls to bottom when the element first gains the attribute (`38-dj-sticky-scroll.js:37`). This matches the typical chat/log UX (you want to see the latest on arrival), but it's a surprise for an author who adds `dj-sticky-scroll` to a long-form scrollable div — the first paint would jump to the bottom.

Either (a) only scroll to bottom if the element appears to already be at/near the bottom before install, or (b) document the behavior explicitly in the guide and in the source comment.

~5 LOC for either option.

johnrtipton:
Closed by PR #943 (merged 2026-04-23).

Issue #880, opened 2026-04-22
tech-debt: dj-track-static — document Map-vs-WeakMap choice in 39-dj-track-static.js


johnrtipton:
From PR #878 Stage 11 code review (🟡).

`_djTrackStaticSnapshot` is a `Map<element, url>` not a `WeakMap`. This is deliberate (the module needs to iterate the snapshot to diff on every reconnect, and WeakMap doesn't allow iteration), but it means every tracked `[dj-track-static]` element is strongly referenced for the page lifetime — no GC.

This is fine in practice because tracked elements live in `<head>` and are long-lived, but the choice should be documented with a brief code comment so future readers don't "fix" it to a WeakMap.

~3 LOC comment. Zero behavior change.

johnrtipton:
Closed by PR #943 (merged 2026-04-23).

PR #943, opened 2026-04-23
fix(observers): attr-removal detection + docs + test (closes #879, #880, #881, #882)

### LINKED ISSUE #879: tech-debt: dj-mutation / dj-sticky-scroll root observer misses attribute removal on kept element
From PR #878 Stage 11 code review (🟡).

If an element is kept in the DOM but its `dj-mutation` (or `dj-sticky-scroll`) attribute is removed via `removeAttribute()`, the document-level root observer in `37-dj-mutation.js` / `38-dj-sticky-scroll.js` doesn't detect that change and doesn't tear down the per-element observer. The attached MutationObserver continues observing (and would fire CustomEvents) even though the element is no longer "marked".

Fix: extend the root observer to also watch for attribute changes on `dj-mutation` / `dj-sticky-scroll` across the whole tree, and call the per-element teardown when the attribute is removed.

~10 LOC per file. Low impact (requires an explicit `removeAttribute` call from application code).

johnrtipton:
## Summary

Cleanup PR closing 4 observer-JS issues across 3 modules + 1 test file.

- **#879 — stale observers on attribute removal**: `37-dj-mutation.js` and `38-dj-sticky-scroll.js` root observers detected NEW elements with the attribute but missed when the attribute was REMOVED from an already-observed element, leaving stale `MutationObserver` (and for sticky-scroll, scroll listener) attached. Added `attributes: true` + `attributeFilter` to the root observe options; the handler now tears down the observer when the attribute disappears.
- **#880 — document Map-vs-WeakMap**: `39-dj-track-static.js` uses `Map` (not `WeakMap`) because `_checkStale` iterates all tracked elements on reconnect to diff snapshot vs. current URLs; `WeakMap` doesn't support iteration. Added a file-level comment and a reference comment near the call site.
- **#881 — document unconditional scroll-to-bottom on install**: `38-dj-sticky-scroll.js` scrolls to bottom on install regardless of current position. Added a comment above the `_scrollToBottom` call explaining the match with Phoenix `phx-auto-scroll` / Ember `scroll-into-view` semantics.
- **#882 — missing regression test**: added a `dj-mutation` test case confirming no `dj-mutation-fire` CustomEvent fires when the element is removed before the debounce timer expires. The existing `_tearDownDjMutation` path (called by the root observer's `removedNodes` branch) correctly clears the pending timer — test passes without code changes, pinning the invariant.

## Files

- `python/djust/static/djust/src/37-dj-mutation.js`
- `python/djust/static/djust/src/38-dj-sticky-scroll.js`
- `python/djust/static/djust/src/39-dj-track-static.js`
- `tests/js/dj_mutation.test.js`
- `python/djust/static/djust/client.js` + minified/gz/br (rebuilt)
- `CHANGELOG.md`

## Test plan

- [x] `npx vitest run tests/js/dj_mutation.test.js tests/js/dj_sticky_scroll.test.js tests/js/dj_track_static.test.js` — 18 passed (was 17)
- [x] Full JS suite: `npm test` — **1286 passed** (was 1285, +1 new test)
- [x] Python suite: `make test-python` — **3428 passed, 15 skipped**
- [x] `scripts/build-client.sh` rebuilt `client.js` and minified artifacts
- [x] CHANGELOG `[Unreleased]` → `Fixed` entry added
- [x] Pre-commit hooks (ruff, eslint, detect-secrets, build-client, npm test) all pass
- [x] Pre-push hooks pass

Closes #879, #880, #881, #882.


johnrtipton:
## Retro — PR #943

**Scope**: 4 observer-JS cleanup issues (#879, #880, #881, #882) across 3 modules + 1 test file, all in one focused PR.

**Went well**
- Clean diff: 4 tiny surgical edits, no behavior changes outside the attr-removal path (#879).
- #882 turned out to be a no-code-change test: the existing `_tearDownDjMutation` path already clears the pending timer on removal, so the new test just pins the invariant. Worth locking in — any future refactor that drops the `clear()` call in `_tearDownDjMutation` would now fail loudly.
- All 11 CI checks green on first push (incl. playwright, rust-tests, CodeQL, security-scan).
- JS test count: **1285 → 1286** (+1), Python: **3428 passed**.

**Went less well**
- Initial `npx vitest` invocation failed because agent cwd resets between Bash calls and the tests use a `./python/...` relative path. Worked around by running from repo root; not a regression but a reminder to always pass repo-root cwd for vitest runs here.
- #879 fix adds a fourth observer option (`attributes: true` + `attributeFilter`) per root observer. This is correct, but every future observer-attribute module (dj-transition, dj-presence, etc.) will need the same pattern. Could be worth a shared `_installRootObserver(attrName, installFn, teardownFn)` helper in a future refactor — not for this PR.

**Follow-ups**
- None blocking. The #880 comment deliberately acknowledges a minor leak (stale Map entries for detached tracked elements between snapshot seeds); if this ever shows up in a memory profile, `_checkStale` could eagerly prune on the pass where it detects `!el.isConnected`.

PR #796, opened 2026-04-21
feat(streams): large-list DOM perf — dj-virtual + dj-viewport-top/bottom (v0.5.0 P1)

johnrtipton:
## Summary

Two client-side DOM performance features that compose naturally for data-heavy UIs. Ships `dj-virtual` (virtual/windowed lists with DOM recycling) and `dj-viewport-top` / `dj-viewport-bottom` (IntersectionObserver-driven bidirectional infinite scroll) together, plus stream `:limit` + `stream_prune` op to make the viewport events memory-efficient on the server side.

Closes two v0.5.0 P1 roadmap items.

### `dj-virtual` — virtual/windowed lists
- Fixed-height items via `dj-virtual-item-height`, CSS `transform: translateY(N*h)` positioning
- `requestAnimationFrame` coalescing for 60fps scroll
- DOM recycling: never destroys item nodes between renders
- `dj-virtual-overscan="5"` for above/below buffer (default 3)

### `dj-viewport-top` / `dj-viewport-bottom` — infinite scroll
- IntersectionObserver on first/last stream children
- Once-per-entry firing via `data-dj-viewport-fired` sentinel
- Re-arm via `djust.resetViewport()` or sentinel replacement
- `dj-viewport-threshold="0.1"` tunable

### Stream `:limit` — memory-efficient scroll
```python
self.stream_append("messages", items, limit=200)  # emits stream_prune op
```
Client trims opposite edge when limit exceeded.

## Test plan

- [x] Python full suite: **1161 passed** (+9 new), 4 skipped, 0 failures
- [x] JS suite: **20 new tests** across 2 new files (all pass); full JS suite 1149 passed
- [x] Rust workspace minus djust_live: green, no regressions
- [x] `cargo fmt` / `ruff` / `bandit` / `detect-secrets` clean
- [x] **Review-found unguarded `console.warn` at `30-infinite-scroll.js:52` fixed** — now wrapped in `if (globalThis.djustDebug)` guard
- [x] **Doc/code inconsistency fixed** — removed phantom `dj-viewport-reset` attribute from docstring; clarified re-arming is programmatic

## Client.js size delta

355725 B → 371511 B raw (+15.7 KB / ~5 KB gzipped estimated). Stays within budget for the feature set.

## Files

**New (6):**
- `python/djust/static/djust/src/29-virtual-list.js` (204 LOC)
- `python/djust/static/djust/src/30-infinite-scroll.js` (169 LOC)
[...]

PR #947, opened 2026-04-23
feat(virtual-list): variable-height items via ResizeObserver (closes #797)

### LINKED ISSUE #797: tech-debt: variable-height virtual list items via ResizeObserver
PR #796 shipped dj-virtual with fixed-height items only (dj-virtual-item-height attribute). Variable-height items require ResizeObserver per-item tracking. Deferred from v0.5.0 per plan. ~200 LOC JS additional. Track for v0.5.1.

johnrtipton:
## Summary
- Adds opt-in variable-height support to `dj-virtual` via a new `dj-virtual-variable-height` boolean attribute. ResizeObserver measures each rendered item; a `Map<index, number>` height cache + lazily-recomputed prefix-sum offsets drive the virtual spacer total and per-row `translateY`. Unmeasured items fall back to a configurable `dj-virtual-estimated-height` (default 50px).
- Fixed-height mode (`dj-virtual-item-height="N"`) is unchanged. When both attributes are set, fixed wins (deterministic). Fixed-mode regression is guarded by an explicit test.
- ~180 LOC net in `29-virtual-list.js` + 4 new JSDOM regression tests.

## Test plan
- [x] `npx vitest run tests/js/virtual_list.test.js` — 16 passed (12 existing + 4 new: attribute activation, mixed-height prefix-sum math, RO-driven cache updates, fixed-mode regression)
- [x] Full JS suite: 1291 passed (103 files)
- [x] Python suite: 3445 passed / 15 skipped
- [x] `bash scripts/build-client.sh` — clean minified + compressed siblings
- [x] Pre-commit + pre-push hooks pass (CHANGELOG test-count validator included)

Closes #797.


johnrtipton:
## Retrospective — feat/virtual-list-variable-height (#797)

**Scope landed**: Opt-in variable-height mode via `dj-virtual-variable-height` attribute; per-item ResizeObserver feeds a height cache; lazy prefix-sum drives offsets + spacer. Fixed-height mode untouched and regression-tested.

**Numbers**
- Net JS: +180 LOC in `29-virtual-list.js` (matches scope estimate in issue).
- Tests: +4 JSDOM cases (12 → 16 in `virtual_list.test.js`). Full JS suite: 1291 passed. Python: 3445 passed.
- Bundle: client.min.js 164118 bytes (+small). gzipped 43212 bytes.

**What went smoothly**
- Existing architecture (state WeakMap, shell/spacer, RAF-throttled render) absorbed the new mode cleanly. Added a `mode` flag and split `render()` into `renderFixed` / `renderVariable` — no churn in the shared scaffolding.
- Stubbing `ResizeObserver` in JSDOM was straightforward: tests dispatch synthetic entries via a `window.__djRoResize(nodeHeightsMap)` helper. Lets the test assert exact prefix-sum math deterministically.
- `dj-virtual-item-height` fixed mode was preserved 1:1 — the regression test scrolls and asserts both element identity (`data-i="18"`) and `translateY(360px)` transform, same contract as the original PR #796 tests.

**Risks / follow-ups**
- `nodeToIndex` is a `WeakMap`, but when the slice shifts we re-`observe` the same nodes with new indices. ResizeObserver dedupes targets, so no leak — but the stale index in the WeakMap could race with a resize callback delivered right after a scroll. The callback resolves node→current index from the map *at callback time*, so stale = 1 frame off. Acceptable for now; worth revisiting if flicker is reported.
- Estimated-height fallback (default 50) means the virtual spacer drifts until items are measured. Scrolling fast through never-rendered items will see the scrollbar jump as measurements come in. This is standard behaviour for variable-height virtualizers (see react-virtualized, react-window). Doc-worthy in the guide.
- Did NOT add: item-id-keyed cache (current Map is keyed on index). If the item pool is replaced via `__djVirtualItems`, the cache is cleared — correct but forfeits measurements. Keying on a user-provided `data-key` would let caches survive reorders. Deferred.

**Follow-up candidates**
- Documentation in the `dj-virtual` guide — a section on variable-mode tradeoffs + `dj-virtual-estimated-height` tuning.
- Optional `data-key`-based height caching for reorder-stable lists.

PR #957, opened 2026-04-23
fix(framework): A010 proxy-trusted deployments + get_state() internal-attr filter (closes #762, #890)

### LINKED ISSUE #762: framework: internal attrs leak into LiveView.get_state() / state_sizes
From djust-browser-mcp observability — the \`_debug.state_sizes\` payload returned in WS responses.

## Signal

On an \`ExamplesView\` mount, \`state_sizes\` includes framework-internal attributes that callers never set:

\`\`\`
allowed_model_fields, base_template, http_method_names, login_required,
login_url, on_mount, on_mount_count, page_meta, page_slug, page_title,
permission_required, static_assigns, static_assigns_count, sync_safe,
template, template_name, tick_interval, use_actors, view_is_async,
active_page, args, kwargs, client_timezone, dark_mode, nav_items,
nav_items_count, temporary_assigns
\`\`\`

These come from djust base classes / mixins (\`LiveView\`, \`BaseLiveView\`, the generic-view-equivalent attrs inherited from Django's \`View\`, etc.).

## Why it matters

1. **State pollution** — \`get_state()\` is supposed to return the user's reactive state. Right now it dumps framework internals alongside, making state reasoning harder.
2. **Fingerprint cost** — \`_snapshot_assigns\` hashes all of this on every event to decide whether to re-render. Small per-attr but adds up.
3. **Observability noise** — \`view_assigns\` / \`state_sizes\` debug payloads balloon with irrelevant keys.

## Fix

Each framework-set attr should either:
- Be prefixed \`_\` (underscored → \`get_state()\` skips)
- Live on a metaclass or class-level (not \`view.__dict__\`)
- Be added to an explicit \`_framework_internal_attrs\` set that \`get_state()\` filters out

Affected attrs fall into three buckets:

**Django \`View\`-inherited (set by \`as_view()\`):** `http_method_names`, `args`, `kwargs`. These aren't djust's to rename but could be filtered by an explicit framework-attr block list.

**djust \`LiveView\`-set (framework config):** `sync_safe`, `use_actors`, `view_is_async`, `tick_interval`, `login_required`, `login_url`, `permission_required`, `allowed_model_fields`, `on_mount`, `on_mount_count`, `static_assigns`, `static_assigns_count`, `template`, `template_name`, `base_template`, `page_meta`, `temporary_assigns`. These ARE djust's — rename to `_*` prefix.

**Mixin-set:** `active_page`, `page_slug`, `page_title`, `client_timezone`, `dark_mode`, `nav_items`, `nav_items_count` — likely from a NavbarMixin or similar in djust.org code, not framework. Filter or rename project-side.

## Reproduce

1. Wire `djust.observability.urls` into a project (DEBUG=True + localhost-only middleware).
2. Mount any LiveView via WebSocket → get session_id
3. `curl /_djust/observability/view_assigns/?session_id=X`
4. See framework-internal attrs in `assigns` alongside real state.

## Related

#762 (hot-reload empty-patch noise) — same debug payload gets sent for no reason.

### LINKED ISSUE #890: djust.A010 check conflicts with AWS ALB (and most L7 load balancer) deployments
## Problem

`djust.A010` raises a hard error when `ALLOWED_HOSTS = ['*']` in production:

```
(djust.A010) ALLOWED_HOSTS contains only '*' in production.
HINT: Wildcard ALLOWED_HOSTS disables Django's Host header defense entirely.
```

This prevents deployment behind AWS ALB (and most L7 load balancers) without workarounds. The ALB sends health checks to the target task using the task's private IP as the Host header (e.g., `10.0.20.11:8000`). The deployer has no way to know what IP the scheduler will assign, and the IP changes on every redeploy/autoscale event.

## What the developer wants to do

Deploy a djust app to Fargate behind an ALB. The ALB is internet-facing with a fixed DNS name. Each Fargate task gets a different private IP from the VPC subnet CIDR. ALB health checks use the target IP as Host.

## Current options

1. **Set `ALLOWED_HOSTS = ['*']`** — blocked by A010
2. **Mix ALB DNS + `'*'`** — blocked by A011 (wildcard with explicit entries)
3. **Enumerate all possible private IPs** — not feasible with autoscaling
4. **Custom health endpoint that bypasses Host validation** — major refactor for a POC
5. **Silence A010 + A011** — what we ended up doing

## Proposed solutions (any of these)

- **Recognize proxy-trusted deployments**: when `SECURE_PROXY_SSL_HEADER` is set AND `TRUSTED_PROXIES` is non-empty, allow `'*'` in `ALLOWED_HOSTS` (the deployer is asserting the proxy terminates requests)
- **Allow CIDR ranges**: accept `10.0.0.0/8` in `ALLOWED_HOSTS` and match any IP in that range as Host (Django doesn't natively do this, would require middleware)
- **Allow subnet-relative wildcards**: accept `10.0.*` patterns for private IP ranges
- **Downgrade A010 to a warning** (not an error) with stronger guidance: "If you're behind a trusted proxy, this may be intentional — set TRUSTED_PROXIES to suppress this check"

## Real-world deployment evidence

NYC Claims prototype (Django/djust on AWS Fargate + ALB) — `ALLOWED_HOSTS=*` is currently the only practical path. The ALB itself filters traffic, so the Host header check is already done at the edge; the check at the Django layer is redundant in that topology.

## What we did

Silenced A010 + A011 in prod.py when `ALLOWED_HOSTS = ['*']`:

```python
if "*" in ALLOWED_HOSTS:
    SILENCED_SYSTEM_CHECKS += ["djust.A010", "djust.A011"]
```

Works, but feels wrong — the checks exist for a reason, and silencing them in a conditional branch is error-prone.

Happy to submit a PR if there's agreement on the approach.

johnrtipton:
Closes #890 and #762.

## Summary

- **#890 (djust.A010/A011 + proxy-trusted deployments)** — when `SECURE_PROXY_SSL_HEADER` and a new `DJUST_TRUSTED_PROXIES` list are both set, `ALLOWED_HOSTS=['*']` is accepted. Unblocks AWS ALB / Cloudflare / Fly.io / other L7-LB topologies where task private IPs rotate on every redeploy. A012 (USE_X_FORWARDED_HOST + wildcard) is deliberately unchanged — still a genuine Host-header-injection footgun even under a trusted proxy.
- **#762 (framework attrs leaked into LiveView state)** — new `live_view._FRAMEWORK_INTERNAL_ATTRS` frozenset (25 names) filters `template_name`, `login_required`, `http_method_names`, `page_meta`, `on_mount_count`, `sync_safe`, etc. out of `LiveView.get_state()`, `_snapshot_assigns` (WS change-detection), and `_debug_state_sizes` / `get_debug_update` (observability payloads). Non-breaking: attribute names unchanged.

## Files touched

- `python/djust/checks.py` — A010 / A011 proxy-trusted escape hatch + updated hint text
- `python/djust/live_view.py` — `_FRAMEWORK_INTERNAL_ATTRS` frozenset + `get_state()` filter
- `python/djust/websocket.py` — `_snapshot_assigns` filter
- `python/djust/mixins/post_processing.py` — `_debug_state_sizes` + `get_debug_update` filter
- `docs/website/guides/deployment.md` — new section "Deploying Behind an L7 Load Balancer"
- `CHANGELOG.md` — `[Unreleased]` → `Fixed` entry
- `python/djust/tests/test_a010_proxy_trusted_890.py` — 7 tests (baseline, full opt-in, partial opt-in x2, empty list, A011 variants)
- `python/djust/tests/test_get_state_filter_762.py` — 7 tests (get_state / snapshot_assigns / state_sizes)

## Test plan

- [x] `pytest python/djust/tests/test_a010_proxy_trusted_890.py python/djust/tests/test_get_state_filter_762.py -v` — 14/14 pass
- [x] `pytest python/tests/test_static_security_checks.py python/djust/tests/test_debug_state_sizes.py` — existing behavior preserved (32 passed)
- [x] `make test-python` — 3445 passed, 15 skipped (no regressions)
- [x] `ruff check` + `ruff format --check` — clean
- [x] Pre-commit hooks passed (bandit, detect-secrets, CHANGELOG test-count validator)
- [x] Pre-push hooks passed (pytest)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -71,6 +71,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- **Framework cleanup (closes #762, #890)** — djust.A010 / A011 system checks now recognize
+  proxy-trusted deployments: when `SECURE_PROXY_SSL_HEADER` + `DJUST_TRUSTED_PROXIES` are both set,
+  `ALLOWED_HOSTS=['*']` is accepted (supports AWS ALB, Cloudflare, Fly.io, and other L7 load
+  balancers where task private IPs rotate). Also filters ~25 framework-internal attrs
+  (`sync_safe`, `login_required`, `template_name`, `http_method_names`, `on_mount_count`,
+  `page_meta`, etc.) from `LiveView.get_state()`, the WS `_snapshot_assigns` change-detection
+  path, and the `_debug.state_sizes` observability payload — user's reactive state is no longer
+  swamped by framework config. Non-breaking fix via a new `live_view._FRAMEWORK_INTERNAL_ATTRS`
+  frozenset; attribute names unchanged. 14 new regression tests in
+  `python/djust/tests/test_a010_proxy_trusted_890.py` and
+  `python/djust/tests/test_get_state_filter_762.py`. Deployment guide updated with the
+  proxy-trusted escape-hatch pattern.
 - **JS-centric batch (closes #949, #951, #953)** — tag_input hidden-input payload now JSON-encoded
   instead of comma-separated, so tag values containing commas round-trip intact (#949). dj-virtual
   variable-height cache now keyed by `data-key` attribute (configurable via `dj-virtual-key-attr`),
--- a/docs/website/guides/deployment.md
+++ b/docs/website/guides/deployment.md
@@ -279,6 +279,38 @@ def health_check(request):
         return JsonResponse({'status': 'unhealthy', 'error': str(e)}, status=503)
 ```
 
+## Deploying Behind an L7 Load Balancer (AWS ALB, Cloudflare, Fly.io)
+
+When djust runs behind a trusted L7 load balancer that terminates TLS and health-checks
+task IPs directly (AWS ALB, Cloudflare, Fly.io, GCP External HTTP(S) LB, etc.), the task's
+private IP rotates on every redeploy or autoscale event. Enumerating them in
+`ALLOWED_HOSTS` is not feasible, and setting `ALLOWED_HOSTS = ['*']` normally trips
+the `djust.A010` system check.
+
+djust provides an explicit opt-in for this topology: set **both** `SECURE_PROXY_SSL_HEADER`
+and a new `DJUST_TRUSTED_PROXIES` list/tuple, and `djust.A010` / `djust.A011` will recognize
+that the trusted proxy has already performed Host validation at the edge.
+
+```python
+# settings.py — only when you really are behind a trusted L7 proxy
+
+ALLOWED_HOSTS = ['*']  # task IPs rotate; edge proxy performs Host validation
+
+SECURE_PROXY_SSL_HEADER = ('HTTP_X_FORWARDED_PROTO', 'https')
+
+# Identify the trusted proxy terminating requests. Any non-empty list/tuple
+# opts in — contents are informational and can be used by your middleware.
+DJUST_TRUSTED_PROXIES = ['aws-alb']  # or ['cloudflare'], ['fly-edge'], etc.
+```
+
+**Security warning.** Only use this escape hatch if you are actually behind a trusted L7
+proxy that performs Host validation at the edge. On a raw VM / bare-metal with no proxy in
+front, `ALLOWED_HOSTS = ['*']` opens you up to Host-header attacks — don't paper over the
+system check with this setting.
+
+If only one of the two settings is present, `djust.A010` / `djust.A011` still fire — both are
+required so a single typo can't accidentally disable the check.
+
 ## Deployment Checklist
 
 ### Infrastructure
--- a/python/djust/checks.py
+++ b/python/djust/checks.py
@@ -566,7 +566,15 @@ def check_configuration(app_configs, **kwargs):
 
     # A010/A011/A012 -- ALLOWED_HOSTS wildcard footguns (#659)
     allowed_hosts = list(getattr(settings, "ALLOWED_HOSTS", []) or [])
-    if not getattr(settings, "DEBUG", False):
+    # Proxy-trusted escape hatch (#890): a deployer behind AWS ALB / Cloudflare /
+    # Fly.io / similar L7 load balancers can't enumerate rotating task private IPs.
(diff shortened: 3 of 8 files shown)


johnrtipton:
## Pipeline Retro

**Two issues shipped cleanly in one PR** — both scoped to roughly the same class of bug ("framework-set state leaking into user-facing surfaces"), which made the batched PR coherent instead of scope-creepy.

### What went well

- **A010 fix landed behind an explicit double-opt-in** (SECURE_PROXY_SSL_HEADER *and* DJUST_TRUSTED_PROXIES). Single-setting typo can't disable the check. Tested both partial-opt-in paths.
- **A012 left intact** — USE_X_FORWARDED_HOST + wildcard is still a genuine Host-header-injection footgun even under a trusted proxy, so the escape hatch deliberately does not apply there.
- **#762 fix is structural, not whack-a-mole** — one `_FRAMEWORK_INTERNAL_ATTRS` frozenset used in three call sites (get_state, _snapshot_assigns, _debug_state_sizes + get_debug_update). Future framework attrs just get added to the set.
- **Non-breaking** — attr names unchanged, no downstream template/integration breakage.
- 14 new regression tests, 3445 passed / 15 skipped in full Python suite (unchanged).

### What to watch

- **DJUST_TRUSTED_PROXIES is currently informational** — the list contents aren't enforced, only non-emptiness is. If we later want to actually validate the header came from a listed proxy (e.g. by IP CIDR match in a middleware), we can layer that without changing the API.
- **A011 hint now mentions the escape hatch** — if someone legitimately wants `['myapp.example.com', '*']` behind a trusted proxy, they'll find it. If they don't, the hint still nudges them to the correct fix.
- **Observability payload shrinkage** — `_debug.state_sizes` dropped ~25 noise attrs. Downstream tools (djust-browser-mcp) that listed those keys explicitly will see shorter payloads; this is the intended outcome per #762.

### Non-goals (deliberately out of scope)

- **Renaming framework attrs to `_*` prefix** — breaking change, deferred. The frozenset filter achieves the same user-visible outcome with zero ecosystem churn.
- **A012 escape hatch** — X-Forwarded-Host injection is independent of whether the proxy is "trusted" for Host validation. Keep firing.
- **Mixin-set attrs** (`active_page`, `client_timezone`, `dark_mode`, `nav_items`, etc. from the issue report) — those are app-side, not framework-side. Left to the app to rename/prefix.

Issue #822, opened 2026-04-21
feat: first-class GCS and Azure Blob UploadWriter subclasses


johnrtipton:
PR #819 ships UploadWriter + BufferedUploadWriter. Users can subclass for GCS/Azure today. Consider shipping GCSMultipartWriter + AzureBlockBlobWriter as first-class subclasses in djust.contrib.uploads so users don't have to reimplement. Benefit: less boilerplate, consistent error handling patterns.

johnrtipton:
Closed by PR #958 — GCSMultipartWriter and AzureBlockBlobWriter shipped in djust.contrib.uploads. Shared UploadError taxonomy in djust.uploads. See CHANGELOG [Unreleased].

PR #958, opened 2026-04-23
feat(uploads): S3 presigned PUT + GCS/Azure writers (v0.5.7 — closes #820, #822)

### LINKED ISSUE #820: feat: pre-signed S3 PUT URLs (client-direct S3 upload bypassing djust)
Complement to PR #819 UploadWriter: instead of client → djust → S3, let client upload directly to S3 via a pre-signed PUT URL. djust's role is only to sign the URL (fast) and observe completion via S3 event. Different threat model from UploadWriter — client bytes never touch djust server. Useful when bandwidth to djust is constrained or upload is >100MB. ADR-008-adjacent.

johnrtipton:
## Summary

Ships 2 v0.5.7 upload-writer features:

- **#820 — Pre-signed S3 PUT URLs**: new `djust.contrib.uploads.s3_presigned.PresignedS3Upload` signs PUT URLs; clients upload bytes directly to S3 via XHR, bypassing the djust WebSocket. Completion is observed via the S3 event webhook (`djust.contrib.uploads.s3_events.s3_event_webhook`) which fires a pluggable `on_upload_complete` hook. Client-side `djust.uploads.uploadPresigned(spec, file, hooks)` added to `static/djust/src/15-uploads.js` with `xhr.upload.onprogress` progress reporting.
- **#822 — GCS + Azure UploadWriter subclasses**:
  - `djust.contrib.uploads.gcs.GCSMultipartWriter` — GCS resumable-upload sessions, `Content-Range`-keyed PUTs, 308/200 state machine, retry on 503/429.
  - `djust.contrib.uploads.azure.AzureBlockBlobWriter` — `stage_block` + `commit_block_list`; uncommitted blocks auto-GC after 7 days, so `abort()` is local-only.

## Error taxonomy

Shared across both issues, rooted at `djust.contrib.uploads.errors` and re-exported from `djust.uploads`:

- `UploadError` → `UploadNetworkError` / `UploadCredentialError` / `UploadQuotaError`

All vendor SDK exceptions (botocore, google.api_core, azure.core) are translated into this taxonomy before propagation. Apps can `except UploadError` without importing vendor SDKs.

## Optional extras

- `djust[s3]` → `boto3>=1.30,<2`
- `djust[gcs]` → `google-cloud-storage>=2.10,<4` + `requests>=2.28,<3`
- `djust[azure]` → `azure-storage-blob>=12.19,<13`

Contrib modules import cleanly without the extras. `ImportError` with the right `pip install` hint fires only when a signing / upload call is actually made.

## Tests

- 50 new regression tests, all SDKs mocked at a `client=` / `service_client=` injection seam so the suite runs without the extras installed:
  - `python/djust/tests/test_presigned_s3_820.py` — 19 cases (signing, key-template rendering, webhook signature verification, SNS envelope unwrapping, hook dispatch)
  - `python/djust/tests/test_gcs_upload_writer_822.py` — 16 cases (init, write, finalize, abort, Content-Range headers, 401/429 translation, idempotent close, JSON-serializable result)
  - `python/djust/tests/test_azure_upload_writer_822.py` — 15 cases (init, block staging, identical block-ID length, commit, abort, auth/throttle translation, zero-byte upload)
- Suite total: **5818 passing** (was 5768 on the baseline), no regressions.

## Test plan

- [x] 50 new tests green
- [x] Full Python suite green (5818 pass, 17 skipped, no regressions)
- [x] Ruff lint + format clean
- [x] Pre-commit hooks pass (bandit, eslint, npm test, build client.js)
- [x] Pre-push hooks pass (full pytest, cargo test)
- [x] CHANGELOG test-count validator passes (`50 regression tests` matches file count across the 3 test files)
- [ ] Manual smoke test with real S3 / GCS / Azure (deferred — would need cloud credentials; mocked paths cover the contract)


johnrtipton:
## Retro — v0.5.7 upload-writer PR (#820 + #822)

### What went well
- **Stable contract made this cheap**: `UploadWriter` base (`open/write_chunk/close/abort`) + `UploadManager`'s lazy-instantiation + abort-on-exception paths were well-documented; implementing GCS & Azure subclasses was ~250 LOC each with no base-class changes needed.
- **Injection seams instead of patching SDK globals**: `client=` / `service_client=` kwargs on the writers and `PresignedS3Upload` made the tests fast (50 tests, ~0.15s) and SDK-independent — the suite runs without `djust[s3,gcs,azure]` installed. Future cloud backends should copy this pattern.
- **Shared error taxonomy upfront**: starting with `UploadError` / `Network` / `Credential` / `Quota` in a sibling module and re-exporting from `djust.uploads` kept the three backends consistent. Apps can `except UploadError` without knowing which backend won.
- **`parse_s3_event` unwraps SNS envelope transparently**: one helper handles both raw S3-direct and SNS-wrapped payloads, so the webhook works for either delivery mode without apps caring.

### What to watch
- **UUID-prefix upload_id convention is implicit**: `s3_events.parse_s3_event` extracts upload_id by regex-matching the first UUID-shaped path segment. This works if apps use `key_template="uploads/{uuid}/{filename}"` (the documented shape), but fails silently (falls back to full key) otherwise. First test iteration hit this — writeup should be clearer.
- **GCS retry loop is tiny**: `_GCS_RETRY_ATTEMPTS = 3` linear, no jitter. Fine for 503/429 bursts, not enough for sustained throttling. Apps with large uploads should wrap or increase.
- **Azure block-ID counter is 10000-max**: at 64 KB chunks that's ~640 MB per upload. Documented, but easy to hit if apps increase client chunk_size. Could be auto-derived from `expected_size` at `open()` time; deferred.
- **No real-cloud smoke test in CI**: all tests mock the SDKs. Would need a matrix job with real AWS/GCP/Azure credentials to verify the happy path end-to-end. Worth adding as a periodic (weekly) job — not blocking per-PR.

### Follow-ups
- `key_template` mini-language supports only bare `{name}`. If apps ask for timestamp / user-id placeholders, whitelist more sources in `_render_key` rather than opening up `.format()` — arbitrary format specs open a `{filename!r}` → repr-injection door.
- `s3_event_webhook` uses shared-secret HMAC, not full SNS cert verification. Production deployments behind API Gateway / Cloudflare can skip cert verification; apps needing it can implement their own view and delegate to `parse_s3_event`.
- Client-side `uploadPresigned` uses XHR (for `upload.onprogress`); when fetch-request-streams ships broadly it's worth re-evaluating.

### Metrics
- **Files changed**: 21 (7 new contrib modules + 3 new test files + 6 existing modified + 4 built bundles + CHANGELOG + pyproject + Cargo/uv locks)
- **Net diff**: +2486 / -13
- **Test delta**: +50 (3445 → 3495 under the make target, 5768 → 5818 under the full `uv run pytest`)
- **Wall time**: single-pass implementation; one test failure (UUID-prefix convention) caught during the local pytest run and fixed before push.

PR #969, opened 2026-04-23
feat(sticky): ADR-011 + user guide + demo + latent fixes (v0.6.0 Phase C — closes Sticky LiveViews milestone)

johnrtipton:
## Summary

**Phase C closes the v0.6.0 Sticky LiveViews milestone.** Third of three PRs:

| Phase | PR | Scope |
|---|---|---|
| A | #966 | Embedding primitive (`{% live_render %}` + `StickyChildRegistry`) |
| B | #967 | Sticky preservation (`sticky=True` + wire frames + stash/reattach) |
| **C (this)** | — | ADR-011 + user guide + demo + 2 latent fixes from #967 Stage 11 |

## What ships

- **`docs/adr/011-sticky-liveviews.md`** (457 lines) — wire protocol, DOM contract, client + server flows, security model (auth re-check, sticky_id escape, stash DoS bounds, inbound forgery rejection, allowlist inheritance), failure modes, threat model, relationship to v0.7.0 `dj-activity` / across-reconnect / cross-tab.
- **`docs/website/guides/sticky-liveviews.md`** (368 lines) — quick-start with AudioPlayer across Dashboard→Settings, full API, lifecycle events (`djust:sticky-preserved` / `djust:sticky-unmounted` with reasons), auth semantics, common patterns (app-shell with global sidebar + sticky widgets), limitations, debugging, FAQ.
- **`examples/demo_project/sticky_demo/`** — runnable demo with `AudioPlayerView` + `NotificationCenterView` (both sticky) across Dashboard / Settings / Reports pages. Reports omits the audio-player slot so you can watch `djust:sticky-unmounted reason='no-slot'` fire in devtools.
- **Latent fix F1** — `03-websocket.js:309-350` `skipMountHtml` branch now calls `reattachStickyAfterMount()`. Currently unreachable from `live_redirect_mount` but a foot-gun if combined with turbo-nav pre-render. ~3 LOC.
- **Latent fix F2** — `websocket.py` `disconnect()` drains `_sticky_preserved` (iterates, calls `_on_sticky_unmount()`, clears dict). Closes a narrow mid-redirect window where async tasks could leak. ~10 LOC.
- **Stale docstring** — `websocket.py:494` now references `data-djust-embedded` (the actual attribute since Phase A's fix pass).
- **`ROADMAP.md`** — Sticky LiveViews line struck with ✅ Shipped (v0.6.0); parity-tracker row updated.
[...]

PR #967, opened 2026-04-23
feat(sticky): sticky LiveViews — preservation across live_redirect (v0.6.0 Phase B of Sticky LiveViews)

johnrtipton:
## Summary

**Phase B of v0.6.0 Sticky LiveViews.** Builds on Phase A (PR #966) to add sticky preservation — a LiveView with `sticky = True`, embedded via `{% live_render "myapp.Audio" sticky=True %}`, keeps its state, WebSocket channels, background tasks, and DOM subtree (form values, scroll, focus) across `live_redirect` navigations. Phoenix LiveView 0.18+ parity.

Phase C (ADR-011 + user guide + demo app-shell) follows in a separate PR to close the Sticky LiveViews milestone.

### User-facing API

```python
class AudioPlayerView(LiveView):
    sticky = True
    sticky_id = "audio-player"
    template_name = "audio/player.html"
```

```django
{# Parent layout or view template #}
{% load live_tags %}
{% live_render "myapp.views.AudioPlayerView" sticky=True %}

{# On the destination page, mark the slot #}
<div dj-sticky-slot="audio-player"></div>
```

On navigation, the audio player's server-side LiveView instance is preserved (same Python object, same state, same background tasks); the client detaches the DOM subtree before the mount frame, reattaches at the matching `[dj-sticky-slot]` via `replaceWith()` (preserving node identity so form values / scroll / focus survive), and fires `djust:sticky-preserved` / `djust:sticky-unmounted` CustomEvents.

### Pre-commit review caught 5 🔴 ship-blockers — all fixed before this commit

1. **`sticky_hold` was emitted AFTER `mount`** → wrong order. Refactored so `handle_mount` emits `sticky_hold` immediately before the `mount` frame inline.
2. **Browser back/forward (popstate) skipped the stash** → sticky subtrees wiped on back button. Fixed to cover both navigation paths.
3. **Redirect request missing `resolver_match`** → sticky views with pk-keyed permissions failed auth re-check. `resolve(url_path)` now attached; `Resolver404` drains stickies.
4. **`handle_mount` failure leaked staged stickies** → background tasks kept running on zombie instances. `try/finally` drains via `_on_sticky_unmount()`.
5. **`stickyStash` persisted across abnormal WS close** → orphaned subtrees across reconnects. `clearStash()` invoked from `03-websocket.js` onclose when `_intentionalDisconnect` is false.

[...]

PR #966, opened 2026-04-23
feat(live-render): embedded LiveView primitive — {% live_render %} tag + StickyChildRegistry mixin (v0.6.0 Phase A of Sticky LiveViews)

johnrtipton:
## Summary

Phase A of the v0.6.0 Sticky LiveViews feature. Ships the foundation for embedded LiveViews — Phoenix nested-LiveView parity. A planning pass discovered that djust's codebase had only \`hasattr\`-gated child-view plumbing in \`websocket.py\` with no production implementation (the methods lived only in test mocks), so the full sticky work naturally splits into 3 phases:

| Phase | Scope | PR |
|---|---|---|
| **A (this)** | Embedding primitive — `{% live_render %}` tag + `StickyChildRegistry` mixin + `view_id`-routed event dispatch + `child_update` wire frame (wire-only) | #966 |
| B (next) | Sticky preservation — `sticky=True` class attr, `sticky_hold`/`sticky_update` frames, DOM stash/reattach, scoped patch applier | TBD |
| C (final) | ADR-011 + user guide + demo app-shell | TBD |

## What Phase A ships

**`{% live_render "dotted.Path" kwargs=... %}` template tag.** Resolves a dotted path via `import_string` → validates `issubclass(LiveView)` (else `TemplateSyntaxError`) → invokes `check_view_auth(child, parent.request)` **before mount** → runs mount + `get_context_data` + render → stamps the assigned `view_id` as `data-djust-embedded` inside every dj-event-bearing element → wraps in `<div dj-view data-djust-embedded="view_id">`.

**`StickyChildRegistry` mixin.** `_register_child`, `_unregister_child`, `_get_all_child_views`, `_get_child_view`, `_assign_view_id`, `_init_sticky`. Composed into `LiveView` base via MRO. Pure embedding-registry — no sticky logic yet.

**Event dispatch routing.** The consumer's event-handler path reads incoming `view_id` and resolves against the parent's child registry; unknown IDs emit a sanitized error (client-supplied string routed through `sanitize_for_log`). Actor-mode path forces fall-through to the sync path for child events.

**`child_update` WS frame.** Outbound from `LiveViewConsumer._send_child_update(view_id, patches, version)`. Client-side handler (new `45-child-view.js`) is **WIRE-ONLY in Phase A** — validates frame + dispatches `djust:child-updated` CustomEvent. `applyPatches` is NOT called because the existing applier doesn't support root-scoped application; Phase B adds a scoped applier.

[...]

PR #965, opened 2026-04-23
feat(animations): FLIP list-reorder + djust_skeleton shimmer placeholder (v0.6.0 phase 2d — animations milestone complete)

johnrtipton:
## Summary

Closes the **v0.6.0 Animations & transitions** milestone. Phases 1 (`dj-transition`), 2a (`dj-remove`), and 2c (`dj-transition-group`) shipped in earlier PRs (#885, #889/#891, #885's batch #895); this PR ships phase 2d — FLIP for keyed-list reorders and a skeleton-placeholder template tag — and retires the remaining "Still to ship" scope.

### Two features

**1. FLIP list-reorder animation — `dj-flip`**

```html
<ul dj-flip dj-flip-duration="300">
    {% for item in items %}<li id="item-{{ item.pk }}">{{ item.name }}</li>{% endfor %}
</ul>
```

When the Rust VDOM patcher emits `MoveChild` for a keyed child (DOM identity preserved), a per-parent `MutationObserver(childList, {subtree:false})` captures First/Last bounding-box rects, applies an inverse `translate(-Δx, -Δy)` as the Invert phase, then clears the transform on the next animation frame with a CSS transition. Tunables: `dj-flip-duration` (default 300 ms, `Number.isFinite` + clamp to [0, 30000]) and `dj-flip-easing` (default `cubic-bezier(.2,.8,.2,1)`, chars `;"'<>` rejected to defeat CSS-property-breakout). Respects `prefers-reduced-motion`. Nested `[dj-flip]` containers isolated. Author-specified inline transforms are preserved across the animation.

No changes to `12-vdom-patch.js` — integration is observer-based, mirroring the `dj-transition-group` pattern. ~260 LOC in `44-dj-flip.js`. 12 JSDOM tests.

**2. `{% djust_skeleton %}` template tag**

```django
{% load live_tags %}
{% if metrics.loading %}
    {% djust_skeleton shape="rect" width="100%" height="120px" %}
{% else %}
    <div class="metric">{{ metrics.result.total_users }}</div>
{% endif %}
```

[...]

PR #970, opened 2026-04-23
feat(sw): VDOM cache + state snapshots + mount batching (v0.6.0 Group 3 — service-worker advanced features)

johnrtipton:
## Summary

**v0.6.0 Group 3: three SW-backed optimizations bundled.** Each one addresses a different back-nav / multi-component UX gap.

### 1. VDOM patch caching

Per-URL HTML snapshots cached in a new `djust-vdom-cache-v1` SW cache. After each successful `mount`, the client posts `VDOM_CACHE` to the SW. On popstate (browser back/forward), the client does `VDOM_CACHE_LOOKUP` and paints the cached HTML into `[dj-view]:not([dj-sticky-root])` **before** the WS reconnect completes — back-nav feels instant. The live `mount` frame then reconciles via the existing patch path.

Configurable: `DJUST_VDOM_CACHE_ENABLED` (default True), `DJUST_VDOM_CACHE_TTL_SECONDS` (1800), `DJUST_VDOM_CACHE_MAX_ENTRIES` (50).

### 2. LiveView state snapshots

Opt-in per view via `enable_state_snapshot = True` class attribute.

```python
class ReportsView(LiveView):
    enable_state_snapshot = True

    def mount(self, request):
        self.report_rows = fetch_slow_report()
        self.filter = "all"
```

Server emits `public_state` alongside the mount frame (only when opt-in AND `DJUST_STATE_SNAPSHOT_ENABLED`). Client stashes to `djust._clientState[view]`. On before-navigate, client serializes state_json and posts `STATE_SNAPSHOT` keyed by the **source** URL. Async-aware popstate handler awaits `STATE_SNAPSHOT_LOOKUP` and attaches result to outbound `live_redirect_mount`. Server calls `view._restore_snapshot(state)` instead of `view.mount()` if slug matches AND `_should_restore_snapshot(request)` passes.

**JSON-only** (no pickle). `safe_setattr` blocks dunder/private keys. 64 KB size cap. 256-key keyset cap. System check `C304` warns on opt-in views with PII-pattern attribute names.

### 3. Mount batching

When ≥ 2 `dj-lazy` LiveViews hydrate together, the client sends a single `mount_batch` frame instead of N separate `mount` frames. Server runs `_mount_one` per view via a `send_json` collector seam, emits one response with survivors in `views[]` + failures isolated in `failed[]`. `navigate[]` array surfaces on_mount redirect hooks so batched views with redirects still work.

**Atomicity**: relaxed (per-view result). Killing the whole batch on one failure would regress today's graceful per-view failure.

## Pre-commit review caught 3 🔴 + 6 🟡 — ALL FIXED before this commit

The parallel Self-Review + Security stages found that 2 of 3 features were **effectively dead code end-to-end** because the scaffolding wasn't wired into any real code path (tests passed by firing wire handlers directly):

1. 🔴 State snapshot: `_clientState` was never populated → capture was a no-op. Now server emits `public_state` in mount frames; client stashes; 46-state-snapshot reads from stash.
2. 🔴 Popstate race: async `.then` populated `_pendingStateSnapshot` AFTER the synchronous `live_redirect_mount` send read it. Popstate handler is now `async` and awaits the lookup.
3. 🔴 VDOM cache: `cacheVdom`/`lookupVdom` defined but never called. Hooked into `case 'mount':` (capture) and popstate (fast-paint).
4. 🟡 `_mount_one` dropped `navigate` frames → batched views with on_mount redirects silently failed. Now surfaces via `navigate[]` in batch response.
5. 🟡 Weak JS test used silent return → trivial pass. Replaced with explicit precondition assertion.
6. 🟡 No server-side `state_json` size cap. 64 KB enforced.
7. 🟡 (Security) Keyset DoS on `_restore_snapshot`. 256-key cap.
8. 🟡 (Security) Non-dict JSON silently passed. `isinstance` guard added.
9. 🟢 `pushState` order bug — snapshot captured against destination URL instead of source. Event now carries `{fromUrl, toUrl}` in detail.
10. 🟢 PII regex extended to cover `credit_card`, `ssn`, `bearer`, `private_key`, `auth_header`, `sensitive`, `credential`.
11. 🟢 `DJUST_STATE_SNAPSHOT_ENABLED` master switch actually consulted now.
12. 🟢 `str(exc)` leak in `_mount_one` error path → `"mount failed"` in production.

## Test plan

- [ ] CI green on all runners (pytest / vitest / rust / CodeQL / security-scan)
- [ ] 34 new Python unit tests in `tests/unit/test_sw_advanced.py`
- [ ] 3 new integration tests in `tests/integration/test_sw_advanced_flow.py`
- [ ] 12 new JSDOM tests in `tests/js/sw_advanced.test.js`
- [ ] Full regression: pytest 6062 passed + 17 skipped, vitest 1349 passed
- [ ] Bundle delta: +1,749 B gzipped (under 2 KB per new module: `46-state-snapshot.js` + sender helpers + mount_batch handler)

## Out of scope

- VDOM patch-log caching (Option C — future when Rust patch serializer lands)
- PREFETCH handler wire-up (separate PR; `22-prefetch.js` currently posts to a dropped SW handler)
- State snapshot encryption (apps with sensitive state should not opt in; documented)
- Cross-origin cache invalidation
- WebRTC peer cache sharing (v0.7.0 investigation)
- Predictive pre-render (v0.7.0)


johnrtipton (commented):
# Stage 11 Code Review — PR #970

**Branch**: `feat/sw-advanced-features-v060`
**Commit**: `4a55dbb0`
**Reviewer focus**: verify the 3 🔴 ship-blockers from the Self-Review fix pass
are actually wired end-to-end (the pre-fix PR had 2 of 3 features dead on the
real application path — only tests invoking wire handlers directly passed).

## Verdict

**APPROVE** — all 🔴 and 🟡 findings have load-bearing wiring on the actual
user path, not just in tests. Observations and minor nits below.

**Counts**: 🔴 0, 🟡 2, 🟢 5

---

## 🔴 Fix verification (proof-by-trace)

### 🔴 #1 — State snapshot end-to-end (PASS)

Full trace for scenario "User visits page A (opt-in view, state `n=99`) →
navigates to B → clicks back":

1. Mount of A: `websocket.py:1988-2000` — master switch on + `enable_state_snapshot`
   true → `_capture_snapshot_state()` result added as `response["public_state"]`.
2. Client `case 'mount':` at `03-websocket.js:284-287` reads `data.public_state`
   and stores into `window.djust._clientState[data.view]`.
3. Navigate-away: `18-navigation.js:119-126` dispatches
   `djust:before-navigate` with `detail.fromUrl = window.location.pathname`
   (i.e. page A).
4. `46-state-snapshot.js:67-88` `_captureBeforeNavigate` reads
   `event.detail.fromUrl` (Fix #9), resolves slug via that URL (route map
   or `[dj-view]`), stringifies `_clientState[slug]`, calls
   `bridge.captureState(fromUrl, slug, json)`.
5. Popstate back: `18-navigation.js:202` handler is `async`; at L249-258 it
   awaits `window.djust._stateSnapshot.lookupStateForUrl(url.pathname)`
   BEFORE assembling `outgoing` at L260. On hit, `outgoing.state_snapshot`
   is attached.
6. Server: `websocket.py:3589-3593` parses `state_snapshot` and forwards via
   `handle_mount(..., state_snapshot=state_snapshot)`.
7. `websocket.py:1771-1842` — master switch, opt-in, slug match, size/keyset
   caps, dict-type check, then `_restore_snapshot(state_dict)` → view has
   `n = 99`. `mounted_from_snapshot = True` → `mount()` skipped at L1844-1846.

Integration test `tests/integration/test_sw_advanced_flow.py:166-190`
exercises the `handle_live_redirect_mount → handle_mount` pathway
end-to-end and asserts `view_instance.n == 99`.

### 🔴 #2 — VDOM cache wiring (PASS)

1. Mount with HTML: `03-websocket.js:373-386` — when `data.html` present and
   `window.djust._sw.cacheVdom` exists, posts VDOM_CACHE msg keyed by
   `window.location.pathname`.
2. Popstate back: `18-navigation.js:223-244` — before the WS frame, awaits
   `window.djust._sw.lookupVdom(url.pathname)`. On `hit && !stale`, paints
   `vdomReply.html` into `[dj-view]:not([dj-sticky-root])` then dispatches
   `djust:vdom-cache-applied`.
3. TTL check: `service-worker.js:251` — `stale = now - ts > VDOM_TTL_MS`;
   stale entries return `hit:false, stale:true` at L254-258.
4. Cache paint proceeds with the normal WS `live_redirect_mount` afterwards,
   so no short-circuit of the authoritative mount (L246-269).

### 🔴 #3 — Mount batch + navigate passthrough (PASS)

1. `_mount_one` at `websocket.py:2092-2202` monkey-patches `self.send_json`
   with a collector, calls `handle_mount`, and inspects captured frames. At
   L2175-2176 "navigate" frames are captured, and L2178-2186 returns the
   navigate payload bubbled up as the 4th tuple element.
2. `handle_mount_batch` at L2204-2265 accumulates `navigates[]` (L2250-2252)
   and emits them in the response `navigate[]` field (L2263-2264).
3. Client `case 'mount_batch':` at `03-websocket.js:464-480` iterates
   `data.navigate` and dispatches through `navigation.handleNavigation`.

Test `test_mount_batch_navigate_passthrough.test` and
`test_redirect_hook_surfaces_in_mount_batch_navigate` both prove the full
bubbling path.

---

## 🟡 Fix verification

| # | Finding | Location | Status |
|---|---|---|---|
| 5 | 64 KB inbound cap on `state_json` | `websocket.py:1785-1792` | PASS |
| 6 | 256-key cap on state_dict | `websocket.py:1820-1827` | PASS |
| 7 | `isinstance(state_dict, dict)` guard | `websocket.py:1807-1814` | PASS |
| 8 | `event.detail.fromUrl` over `location.pathname` | `46-state-snapshot.js:73-76` | PASS |
| 9 | PII regex extended | `checks.py:807-811` | PASS — includes `credit_card`, `ssn`, `bearer`, `private_key`, `auth_header`, `sensitive`, `credential` |
| 10 | `DJUST_STATE_SNAPSHOT_ENABLED` master switch | `websocket.py:1771`, `1989` | PASS — gates both restore branch and public_state emission |

---

## Independent review — edge cases considered

**Concurrency of `_mount_one` send_json monkey-patch**
(`websocket.py:2126-2160`): Django Channels `receive()` awaits message
handlers sequentially (verified via the dispatch loop at L1322-1342), so two
`handle_mount_batch` invocations cannot overlap on the same consumer.
`finally` restores `send_json` even after the `except`'s early return
(idempotent double-restore). Exception-safe. 🟢

**Public state leakage scope**: the mount frame's `public_state` comes from
`_capture_snapshot_state()` (`live_view.py:558-585`) which filters `_`-prefixed
attrs, the explicit `_FRAMEWORK_INTERNAL_ATTRS` frozenset, callables, and
JSON-encoder failures — NOT `get_context_data()`. Filtering is real and
consistent with the ingest-side filter in `_restore_snapshot`. 🟢

**Cache coherence across bundle updates**: the VDOM cache stores full HTML
with a `version` tag (the VDOM version at mount time, not the client.js
bundle version). On a bundle update, `dj-track-static` + `djust:ws-reconnected`
(added in v0.5.2rc1) triggers a reload which evicts cached entries naturally
via fresh SW activation. The cached HTML is painted as "dead" markup until
the authoritative mount frame replaces it — minor UX: clicks during the
pre-reconcile gap are lost, but this is by design and the WS roundtrip is
typically <100ms. 🟡 (noted, not blocking)

**Mount_batch fallback on older server**: the client sends `type: mount_batch`
unconditionally when `mounts.length >= 2`. An older server with no
`handle_mount_batch` dispatch would fall into the generic "unknown message
type" error path (L1322). Not a regression per se but worth noting for a
release-notes compatibility matrix. 🟡

**popstate handler async conversion**: the handler at L202-281 in
`18-navigation.js` is now `async`. Searched the codebase for consumers
assuming synchronous popstate behavior — no `popstate` listener elsewhere
in `static/djust/src/*` mutates shared state in a way that would race with
an awaited lookup. 🟢

**VDOM cache vs sticky preservation on popstate**: at L219-244, sticky
subtrees are stashed BEFORE the VDOM cache paint, and the paint targets
`[dj-view]:not([dj-sticky-root])` — the sticky wrapper is explicitly
excluded. Reattach runs after the authoritative mount frame. 🟢

**`DJUST_STATE_SNAPSHOT_ENABLED=False` fully short-circuits**: verified both
ingest path (L1771-1776) and emit path (L1989-1990) read the setting with
default `True`; test
`test_master_switch_disables_snapshot_restoration` covers the ingest side.
🟢

## Test quality

- `tests/integration/test_sw_advanced_flow.py` uses a `_FakeConsumer`
  shim around the real `LiveViewConsumer` and exercises
  `handle_live_redirect_mount → handle_mount → _restore_snapshot`
  end-to-end — this is proof of the critical path, not a handler-firing
  stub.
- JSDOM test `case mount stashes public_state into djust._clientState[view]`
  dispatches the real `mount` frame and verifies the side-effect.
- `applies html from mount_batch` uses `handler.handleMessage` (the
  real dispatcher), not a direct function call.
- 47 Python + 12 JS cases, all passing locally (0.13s + 919ms).

All tests run: **47 PASS** (Python) + **12 PASS** (JS).

## CI status

In-progress at review time (python-tests 3.12/3.13/3.14, rust-tests,
javascript-tests, playwright, security). Security Hot Spot Check already
SUCCESS. No failures at review time; approval is conditional on these
completing green.

## Minor observations (🟡 non-blocking)

1. **Best-effort `try/catch` swallowing** in `03-websocket.js:386` (`/* best-effort cache put */`)
   — if `cacheVdom` ever throws synchronously from a non-Safari SW impl,
   you'll silently lose cache writes. Consider logging under
   `globalThis.djustDebug` guard.
2. **Mount_batch protocol-version handshake** — no version bump or feature
   probe. An older server will reply with a generic error and the client
   will log a warn. Consider surfacing a `DJUST_USE_MOUNT_BATCH = false`
   auto-fallback on first `mount_batch` error to avoid stuck lazy hydration
   on version-mismatched deploys.

## Summary

Given the severity of the initial Self-Review findings (two features
shipping as dead code), I traced every 🔴 fix through the user path and
confirmed each step. The `_mount_one` seam is elegantly minimal; the
state-snapshot capture/restore symmetry is sound; the VDOM cache respects
sticky and TTL. The two 🟡 observations are enhancements, not blockers.
Tests are rigorously wired — no "fire handler, assert state" pattern; all
use real dispatchers.

**APPROVE** pending CI completion.

johnrtipton:
# Retro — PR #970 (feat: v0.6.0 Group 3 — service-worker advanced features: VDOM cache + state snapshots + mount batching)

**Branch**: `feat/v060-sw-advanced-features`
**Merged**: 2026-04-23T21:44:11Z (squash-merged as commit ed8ac7e4)
**Scope**: v0.6.0 Group 3. Three bundled features: (1) VDOM patch caching — per-URL HTML held in service-worker cache for popstate fast-paint, (2) LiveView state snapshots — opt-in, JSON-only, auth re-check on restore, (3) mount batching — N mounts coalesced into 1 frame for `dj-lazy` views with per-view failure isolation + navigate passthrough. 49 new tests (34 Python unit + 3 integration + 12 JSDOM). 6062 pytest / 1349 vitest at merge. Bundle +1,749 B gzipped.

## What Worked Well

- **Pre-commit Self-Review caught a catastrophic class of bug: "dead code that passes tests."** 2 of 3 features had functional wire-level scaffolding but no user-code path invoking the new APIs:
  1. State snapshot: `_clientState` never populated anywhere → capture was a no-op.
  2. VDOM cache: `cacheVdom`/`lookupVdom` defined but never called by any module.
  The wire-level tests passed because they fired the message handlers directly. This is EXACTLY the pattern that almost shipped broken demos in PR #969. The three-layer review model (Self-Review + Security + Stage 11) is essential — and Self-Review is the one that found the dead-code issue, not Stage 11.
- **Security stage independently found a third functional bug** (`pushState` happens BEFORE `djust:before-navigate` dispatch, so snapshots keyed on destination URL). Only flagged 🟢 by Security but combined with Self-Review's 🔴s, it was fixed in the same pass.
- **All 12 findings landed in one fix pass**: popstate handler refactored to async, server emits `public_state` alongside mount, client stashes to `_clientState`, cacheVdom/lookupVdom hooked into `case 'mount':` + popstate, 64KB + 256-key caps added, PII regex extended, master switch wired both sides, `str(exc)` leak closed. Stage 11 independently verified each with concrete line traces.
- **`_mount_one` via `send_json` monkey-patch collector**: pragmatic design trade-off. Full refactor of `handle_mount` into a `(success, payload, error)`-returning helper would have had huge blast radius across 900+ existing tests. Monkey-patch captures the frame-out side-effect safely (Channels serializes per-consumer; `finally` block restores). Clean way to add batching without destabilizing the main mount path.

## What Didn't Work

- **The implementation agent shipped "mostly wired, but not actually." It's a scaffolding-without-plumbing pattern** that passes unit tests (because handlers exist and respond to direct messages) but fails functional-test smoke checks (because nothing in the user-visible code path invokes them). Lesson: **every new public API needs a WIRING_CHECK**: grep for callers of each new exported function/method, and if the only callers are tests, the feature isn't actually wired. The implementer's WIRING_CHECK report said "pass" for `_restore_snapshot` and `_should_restore_snapshot` — but didn't grep for the CLIENT-SIDE consumers of `cacheVdom`, `lookupVdom`, `_clientState`. Those were the actual dead-code sites.
- **Tests using substring + direct-handler-invocation patterns give false confidence.** The 12 JSDOM tests passed by calling `handler.handleMessage({type:'VDOM_CACHE'...})` directly, not by navigating to a URL and observing the side-effect. Future SW-feature tests must include an **integration-level** case that exercises the real path (navigate → mount → cache → popstate → fast-paint).
- **Master switch `DJUST_STATE_SNAPSHOT_ENABLED` was defined but unused** — caught by Security review. Suggests the implementer added the config per the plan but didn't wire it into the code path that consumes it. Same pattern as dead-code features — scaffolding without completion.

## What To Improve

- **Add "grep for real callers" to every WIRING_CHECK.** Specifically: for every new exported JS function and every new Python public method, grep for callers that are NOT in `tests/` — if zero, the feature is unwired.
- **Require integration test for any SW/background-work feature.** Unit-level handler tests are necessary but not sufficient.
- **Introduce a test category (`@pytest.mark.e2e_wiring`) for end-to-end wiring checks** that cannot be faked by direct handler invocation.

## Recurring Issue Tracker

- **Retro-artifact gate worked** — this retro posted via `gh pr comment`.
- **Three-layer review model MUST run on every PR** — Phase C lesson reinforced. Skipping Self-Review on "small" PRs let 3 🔴 through on #969; running Self-Review on this PR caught 3 different 🔴 (all dead-code) that handler-level tests missed.
- **Config-flag-defined-but-unwired pattern** appeared in this PR (master switch) AND in previous ones (various `_FRAMEWORK_INTERNAL_ATTRS` additions). Maybe add a `configs_used_in_code` check that scans `config.py` keys and ensures each has at least one non-test consumer.

## Review Stats

| Metric | Value |
|---|---|
| LOC | +3,366 / −34 across 21 files |
| Python tests added | 34 unit + 3 integration = 37 |
| JSDOM tests added | 12 |
| Bundle size delta | +1,749 B gzipped |
| 🔴 findings (Self-Review pre-commit) | 3 (all fixed) |
| 🟡 findings (Self-Review) | 3 (fixed) |
| 🟡 findings (Security) | 2 (fixed) |
| 🟢 nits (Self-Review + Security) | 6 (4 fixed, 2 deferred to tech-debt) |
| 🔴 findings (Stage 11) | 0 |
| 🟡 findings (Stage 11) | 2 (silent cache-write failures, no version-probe fallback) |
| 🟢 nits (Stage 11) | 5 |
| CodeQL iterations | 1 (green first try — implementer applied Phase A regex lesson proactively) |
| CI checks | 11/11 green |
| Re-commit attempts | 1 (initial + bundle-regen auto-stage) |

## Deferred Items

- **🟡 Silent cache-write failures** in `03-websocket.js:386` `try/catch {}` — should log under djustDebug. Tech-debt issue.
- **🟡 No version-probe fallback for `mount_batch`** — older servers without the handler produce generic "unknown msg type" error; client should auto-fall-back to per-view mount. Tech-debt issue.
- **v0.6.0 Group 4 — Package consolidation sunset (ADR-007 Phase 4)** — next.
- **v0.6.0 Group 5 — Performance profiling (P2)** — after Group 4.
- **v0.6.0 Group 6 — AI-generated UIs (ADR-006, HEADLINE, ~9 weeks)** — final.

## Status — v0.6.0 run

| # | Task | PR | Status |
|---|---|---|---|
| 1 | Animations FLIP + Skeleton | #965 | ✅ |
| 2a | Embedding primitive | #966 | ✅ |
| 2b | Sticky preservation | #967 | ✅ |
| 2c | ADR + guide + demo | #969 | ✅ |
| 3 | SW advanced features | #970 | ✅ |
| 4 | Package consolidation sunset | — | next |
| 5 | Performance profiling | — | queued |
| 6 | AI-generated UIs | — | queued |

PR #975, opened 2026-04-24
feat(http): streaming initial render — chunked HTTP response (v0.6.1 Group 2 Phase 1)

johnrtipton:
## Summary

**Second v0.6.1 deliverable: Streaming initial render (Phase 1).** Opt-in via `streaming_render = True` class attribute on `LiveView`. `get()` returns `StreamingHttpResponse` with a 3-chunk iterator: shell-open → main-content → shell-close. Browser parses shell immediately and begins loading CSS/JS while server completes render. Competitive with Next.js `renderToPipeableStream`.

```python
class ReportsView(LiveView):
    streaming_render = True
    template_name = "reports/dashboard.html"
    # Heavy DB queries in get_context_data() now don't block browser shell parsing
```

## Pre-commit review caught 1 🔴 + 3 🟡 — all fixed

- 🔴 Regex false-match on hyphenated attrs (`dj-root-other`, `data-dj-root`). Fixed with lookbehind+lookahead. 10 parametrized regression cases.
- 🟡 `</body>` inside `<script>` literal caused premature split. Fixed with length-preserving script-block masking. 2 regression cases.
- 🟡 Module-level regex compile for perf (hoisted `_DJ_ROOT_RE`, `_BODY_CLOSE_RE`, `_SCRIPT_BLOCK_RE`).
- 🟡 Doc caveat for `<style>` / HTML comments (acknowledged limitation).

Security PASSED: HTML injection NONE (string-slicing only), header propagation CONFIRMED, middleware bypass NONE, CSP nonces SAFE.

## Test plan

- [ ] CI green on all runners
- [ ] 35 unit (23 + 12 regression) + 4 integration = 39 new tests
- [ ] Full regression: pytest 6124 passed + 17 skipped
- [ ] No JS / vitest changes expected
- [ ] Bundle unchanged

## Out of scope (Phase 2 / v0.6.2)

- Lazy-child streaming (`{% live_render lazy=True %}` children streamed as each child mounts via `dj-lazy-stream` protocol)
- Async view rendering (`async def get`)
- Buffering-proxy system check
- `Cache-Control: no-store` default
- `<style>` / HTML-comment `</body>`-literal masking (acknowledged limitation)

## What's next in v0.6.1

- ~~Hot View Replacement~~ ✅ #974
- ~~Streaming initial render~~ ✅ this PR
- Time-travel debugging (~500 LOC, debug panel + render pipeline) — last v0.6.1 group
- AI-generated UIs (ADR-006) — deferred to v0.6.2

[...]

PR #974, opened 2026-04-24
feat(hot-reload): Hot View Replacement — state-preserving dev-mode class swap (v0.6.1 Group 1)

johnrtipton:
## Summary

**First v0.6.1 deliverable: Hot View Replacement (HVR).** React Fast Refresh equivalent for djust LiveViews. When a `.py` module changes on disk, the dev server `importlib.reload`s it and swaps `__class__` on every live instance — users keep form input, counter values, active tab, scroll position. No reconnect, no state reset.

## Implementation reuses `djust_hotreload`

Rather than inventing a parallel pipeline, HVR is an **additive pre-step** inside the existing `hotreload()` handler at `websocket.py:3305`. The event payload gains optional `hvr_meta = {module, class_names, reload_id}`; consumer decodes it, runs `apply_class_swap`, sends `hvr-applied` frame, then continues into the existing template-refresh + `render_with_diff` path. On incompat: `{type:"reload"}` fallback (same as today's template-error path).

Compat heuristic (`_is_state_compatible`) rejects swaps when:
- `__slots__` layout differs
- MRO (mixin chain) changes
- An event-handler-decorated method is removed
- Handler positional param names change
- Handler had `**kwargs` but new doesn't

Adding new handlers is safe. `mount()` body changes are safe.

## Pre-commit review caught 2 🔴 + 4 🟡 — all fixed before commit

- 🔴 `**kwargs` removal was undetected → TypeError on client events. Fixed with VAR_KEYWORD parity check + regression test.
- 🔴 `_child_views` recursion had no cycle guard → RecursionError. Fixed with `_visited: set[int]` threaded through.
- 🟡 MRO drift (mixin removal) undetected → added `_mro_names` comparison.
- 🟡 Top-level module side effects re-run on `importlib.reload` → documented caveats.
- 🟡 Symlinks not handled → `os.path.realpath`.
- 🟡 `apply_class_swap` silent on non-LiveView → `logger.debug`.

Security PASSED: inbound `hotreload` forgery REJECTED, module-reload BOUNDED, class-swap SAFE (no `__init_subclass__` registries), `hvr-applied` XSS SAFE (`textContent`).

## Test plan

- [ ] CI green on all runners
- [ ] 21 new Python unit + 2 integration + 3 JSDOM = 26 new
- [ ] Full regression: pytest 6085 passed / 17 skipped, vitest 1352 passed
- [ ] Bundle delta: +357 B gzipped (under 500 B budget for `47-hvr.js`)
[...]

PR #976, opened 2026-04-24
feat(debug): time-travel debugging — state-history ring buffer + debug panel tab (v0.6.1 Group 3)

johnrtipton:
## Summary

**Third v0.6.1 deliverable. Closes user-facing v0.6.1 scope.** Records before/after state snapshots around every `@event_handler` dispatch into a per-view `TimeTravelBuffer`. Debug panel tab scrubs the timeline; click-to-jump restores state via `safe_setattr` + re-renders. Beyond Redux DevTools.

```python
class CounterView(LiveView):
    time_travel_enabled = True

    @event_handler
    def increment(self, **kwargs):
        self.count += 1
```

Open debug panel → "Time Travel" tab → click a past event to rewind the view.

## Pre-commit review caught 3 🔴 + 5 🟡 — ALL fixed

- 🔴 Component + actor event paths weren't instrumented → per-branch wrap on all 3 dispatch sites + permission-denied + validation-failed paths.
- 🔴 Timeline buttons had data-* attrs but NO click handler → rendered tab was functionally dead. Added delegation; regression test uses real `button.click()`.
- 🔴 Client history was never populated (server sent only an ack) → new `time_travel_event` frame on every `record_event_end`.
- 🟡 Restore left ghost attributes → delete current public attrs not in snapshot first.
- 🟡 Async handlers snapshot too early → documented.
- 🟡 Permission-denied events recorded via Fix #1.
- 🟡 Error truncation moved into `record_event_end`.
- 🟡 PII caveat added to docs.

Security PASSED: TWO-GATE prod defense, jump forgery SAFE, restore dunder-block ENFORCED, DoS BOUNDED.

## Test plan

- [ ] CI green
- [ ] 42 new tests (32 Python unit + 3 integration + 7 JSDOM)
- [ ] Full regression: pytest 6156 / 17 skipped, vitest 1359
- [ ] Bundle delta: +789 B gzipped on debug-panel.min.js, +80 B on client.min.js.gz

## Out of scope (v0.6.2+)

- Component-level time-travel (Phase 1 records against parent).
- Forward-replay through branched timeline.
- Persistent buffer.
- Multi-user shared timeline.

## v0.6.1 milestone

- ✅ #974 Hot View Replacement
- ✅ #975 Streaming initial render (Phase 1)
- ✅ this PR — Time-travel debugging
- Deferred to v0.6.2: AI-generated UIs (ADR-006), Phase 2 lazy-child streaming


johnrtipton:
# Stage 11 Code Review — PR #976

**Title:** feat(debug): time-travel debugging — state-history ring buffer + debug panel tab (v0.6.1 Group 3)
**Commit:** 0430ef10
**Verdict:** REQUEST_CHANGES

## Verification of the 3 prior RED fixes

All three pre-commit 🔴 fixes landed correctly on inspection:

1. **Per-branch instrumentation.** Actor path wraps record/end at `python/djust/websocket.py:2343-2430`; component path at `python/djust/websocket.py:2464-2535`; view-handler path at `python/djust/websocket.py:2581-2667`. Permission-denied and validation-failed branches emit their own `_tt_end` + `_maybe_push_tt_event` calls before the early `return` (lines 2595-2597, 2608-2610). Integration test `test_permission_denied_view_handler_records_with_error` in `tests/integration/test_time_travel_flow.py:192` passes and verifies error="permission_denied" lands in the buffer.
2. **Delegated click handler.** `registerTimeTravelClickHandlers()` at `python/djust/static/djust/src/debug/09a-tab-time-travel.js:114` attaches a single delegated `click` listener on `this.panel`, guarded by `_ttClickBound` to prevent double-binding. Registration happens in `02-tab-system.js:63-71`. JSDOM test at `tests/js/time_travel_tab.test.js:94` calls `button.click()` (the required regression, not `onTimeTravelJumpClick()` directly).
3. **Client history populated via push.** `_maybe_push_tt_event` at `python/djust/websocket.py:3877` emits `{"type": "time_travel_event", "entry": snapshot.to_dict(), "history_len": N}` after every `record_event_end`. `03-websocket.js:843-860` re-dispatches as a `djust:time-travel-event` `CustomEvent`; `02-tab-system.js:72-83` binds a listener that calls `onTimeTravelEvent()`, which appends to `this.timeTravelHistory` and re-renders the tab.

YELLOW fixes verified: ghost-attr deletion at `python/djust/time_travel.py:236-252` (test at `tests/unit/test_time_travel.py:225`); async caveats in the module docstring AND `docs/website/guides/time-travel-debugging.md:162`; error truncation centralized in `record_event_end` at `python/djust/time_travel.py:191-196` with `ERROR_MESSAGE_MAX_CHARS=200`; PII blockquote at the guide's line 154.

Tests: pytest 32/32 unit + 7/7 integration pass. vitest 7/7 pass. CI rolling.

## New BLOCKING finding — 🔴 WS send path is dead at runtime

`python/djust/static/djust/src/debug/09a-tab-time-travel.js:100`:
```js
const ws = globalThis.djust && globalThis.djust.websocket;
const send = ws && typeof ws.sendMessage === 'function' ? ws.sendMessage.bind(ws) : null;
```

`globalThis.djust.websocket` is **never assigned anywhere in the codebase.** Every other file that needs to reach the WS uses `window.djust.liveViewInstance` (see `19-hooks.js:90-91`, `20-model-binding.js:80`, `debug/11-integration.js:75`). Because the test at `tests/js/time_travel_tab.test.js:79` stubs `globalThis.djust = { websocket: { sendMessage } }` exactly in the shape the handler expects, vitest is green — but in a real browser, `ws` is `undefined`, `send` is `null`, and every timeline click silently falls through to the `else if (globalThis.djustDebug)` branch. **Clicking a `tt-jump` button does nothing in production.** This is the same shape of bug that Fix #2 was meant to prevent — the UI is present and tests pass, but the integration layer is never wired.

Fix: use `const lv = window.djust && window.djust.liveViewInstance; if (lv && lv.ws && lv.ws.readyState === WebSocket.OPEN) lv.sendMessage(payload);` — or expose `03-websocket.js` as `globalThis.djust.websocket` and update the test to reflect the real global. Either way, add a Playwright test that boots a view, clicks a `tt-jump`, and asserts a WS frame goes out — don't let a JSDOM stub green-wash this again.

## 🔴 Second correctness bug — snapshot aliasing (no deep copy)

`_capture_snapshot_state` at `python/djust/live_view.py:612-639` returns live value references: the `json.dumps(value, ...)` call is used only to gate serializability, then `result[key] = value` stores the original reference. Reproduced locally:
```
state_before[items] before mutate = [1, 2, 3]
v.items.append(999)
state_before[items] after mutate  = [1, 2, 3, 999]   # buffer corrupted
```
For a time-travel debugger this is a correctness hole: any handler that mutates a list/dict attribute in place (very common — `self.items.append(...)`) will retroactively rewrite every prior snapshot that holds the same reference. `restore_snapshot` will then "restore" to the current value, silently breaking back-navigation. Fix: `copy.deepcopy(result)` before returning from `_capture_snapshot_state`, OR round-trip through `json.dumps`/`json.loads` (already proven to be serializable by construction). Guard with a try/except and degrade to a shallow copy on failure.

## 🟡 Secondary concerns

- **Frame size unbounded.** `_maybe_push_tt_event` serializes `entry.state_before + state_after` on every event. A view holding a 1000-row list emits ~20+ KB per frame, doubling the WS bandwidth of a spammy handler. Consider capping per-event payload (e.g. truncate lists/strings over 10 KB with a `"<truncated>"` marker) or gate the push behind a config flag. Dev-only, but still worth a ceiling.
- **Jump-during-async race.** If a user jumps to a past snapshot while a `start_async()` task from a later event is still running, the task's writes clobber the restored state. Accepted limitation, but not mentioned in the guide — add to `docs/website/guides/time-travel-debugging.md` "Side effects do not replay" section.
- **Concat ordering.** `scripts/build-client.sh:94` uses `"$DEBUG_SRC_DIR"/[0-9]*.js` — shell glob is lexicographically sorted, so `09a-tab-time-travel.js` does land between `09-tab-variables.js` and `10-logging.js`. Stable on POSIX; fine.

## 🟢 Good

- Thread-safe `TimeTravelBuffer` with `threading.Lock` + `deque(maxlen=...)` is the right primitive.
- Ghost-attr cleanup in `restore_snapshot` correctly iterates `vars(view)` snapshot before delete, and the `AttributeError` pass-through on `@property`-backed attrs is defensive (confirmed at `python/djust/time_travel.py:247-250`).
- `safe_setattr` reuse on restore matches v0.6.0 state-snapshot hardening.
- All `console.log` calls guarded behind `globalThis.djustDebug`.
- CHANGELOG entry is thorough (30 lines of detail).
- C501/C502 system checks present with clear hints and IDs.

## Decision

**REQUEST_CHANGES.** The two BLOCKING items — dead WS send path and shared-reference snapshot aliasing — are both correctness holes that defeat the core value of the feature. Fix both, add a Playwright (or djust-browser) smoke test for the click-to-jump round trip, and re-request review. Given the 3-🔴 pre-commit history, the JSDOM-only coverage for the click delegation path was insufficient — tests that mock the integration surface that doesn't exist in production will keep shipping these regressions.

johnrtipton:
# Retro — PR #976 (feat: time-travel debugging v0.6.1 Group 3)

**Branch**: `feat/time-travel-debugging-v061`
**Merged**: 2026-04-24 as `f66b085b`
**Scope**: Third v0.6.1 deliverable. Closes user-facing v0.6.1 milestone.

## What Worked Well

- **Pre-commit review caught 3 🔴 scaffolding-plumbing bugs before commit** (actor/component paths uninstrumented, timeline click delegation missing, client history never populated). Same failure mode observed in PR #970 SW advanced features — now caught reliably by Self-Review first-pass.
- **Stage 11 caught 2 MORE 🔴 that pre-commit missed**: dead-WS-path in tab click handler + snapshot reference aliasing. The JSDOM test was greenwashing the WS bug by stubbing a non-existent shape (`globalThis.djust.websocket`). Real canonical path is `window.djust.liveViewInstance.sendMessage`. Stage 11's rigor + willingness to trace runtime data flow caught both — a reminder that independent review at Stage 11 remains indispensable even after a thorough pre-commit pass.
- **The snapshot-aliasing bug (🔴 #2 in Stage 11) is a latent fix for the v0.6.0 state-snapshot feature too.** Same `_capture_snapshot_state` helper. `self.items.append(...)` after snapshot was rewriting every prior snapshot via reference. json roundtrip deep-copy fixed both features in one line.
- **Per-branch instrumentation via try/finally is robust.** Wrapping `handle_event`'s actor + component + view-handler + permission-denied + validation-failed branches all record correctly. No events slip through.

## What Didn't Work

- **Implementation agent wrote "send message" code without verifying the shape against existing consumers.** `globalThis.djust.websocket` doesn't exist anywhere in the codebase; the correct shape `window.djust.liveViewInstance.sendMessage` is used by `03-tab-events.js` and `11-integration.js`. Pattern: implementer imagined an API that felt reasonable rather than reading existing code.
- **JSDOM test stubbed the imagined API, not the real one** → test passed while feature was functionally dead. Classic false-green. Stage 11 caught it with "does the test stub match what other modules use?" check.
- **`_capture_snapshot_state` reference-aliasing bug existed in v0.6.0** (enable_state_snapshot feature) and went uncaught for two milestones. Nobody exercised the "mutate container attr after event" path in a test. Now covered.

## What To Improve

- **Add "grep for how OTHER callers do X" as a planning-stage check** for any client-side feature that sends WS frames or reads server state. "How does Tab X send a WS message?" should be answered BEFORE the implementation agent writes its send path.
- **Add a mutation-safety test to `_capture_snapshot_state`** that was missing. Generalize to: any capture function needs a test that exercises mutation-after-capture.
- **Pre-commit Self-Review should add a "grep for the stubbed API" check** when reviewing JSDOM tests. If the test stubs `globalThis.djust.FOO` and nothing in source ever assigns `djust.FOO`, it's greenwashing.

## Recurring Issue Tracker

- Retro-artifact gate worked.
- Scaffolding-no-plumbing pattern: 3rd occurrence (PR #970 SW advanced, #976 time-travel × 2 distinct instances). Now reliably caught by Self-Review + Stage 11 two-layer review.
- Stage 11 > pre-commit value: 2 🔴 that pre-commit missed this time. Pre-commit isn't a substitute for Stage 11.

## Review Stats

| Metric | Value |
|---|---|
| LOC | +2,100 / −36 across ~15 files |
| Python tests added | 37 unit + 3 integration = 40 |
| JSDOM tests added | 8 |
| Bundle delta | +789 B gz (debug-panel) + 80 B gz (client) |
| 🔴 pre-commit Self-Review | 3 (all fixed) |
| 🟡 pre-commit | 5 Self-Review + 1 Security = 6 (all fixed) |
| 🔴 Stage 11 | 2 (both fixed) |
| 🟡 Stage 11 | 3 (all fixed) |
| 🟢 nits | 3 (deferred) |
| CI checks | 13/13 green (no re-runs needed after each commit) |
| Re-commit attempts | 3 (initial, post-pre-commit-fix, post-Stage-11-fix) |

## Deferred Items

- Component-level time-travel (Phase 1 records against parent).
- Forward-replay through branched timeline (Redux DevTools parity).
- Persisting buffer across server restart.
- Jump-during-async cancel-in-flight (documented as accepted limitation).

## Status — v0.6.1 milestone

| # | Task | PR | Status |
|---|---|---|---|
| 1 | Hot View Replacement | #974 | ✅ |
| 2 | Streaming initial render (Phase 1) | #975 | ✅ |
| 3 | Time-travel debugging | #976 | ✅ |
| 4 | AI-generated UIs (ADR-006) | — | deferred to v0.6.2 |

**v0.6.1 user-facing scope COMPLETE.** Ready for milestone retro + `v0.6.1rc1` cut.

PR #116, opened 2026-01-29
Prevent WebSocket error messages from leaking internal details

johnrtipton:
## Summary

- Add `_safe_error()` helper that returns detailed messages in DEBUG mode, generic messages in production
- Replace security-sensitive `send_error()` calls with `_safe_error()` for mount errors (view path, import, class resolution) and event errors (blocked name, no handler, not decorated)
- All event rejection paths now return identical "Event rejected" message in production, preventing handler enumeration

## Test plan

- [x] New `TestErrorDisclosure` class verifies generic messages when `DEBUG=False` and detailed when `DEBUG=True`
- [x] Anti-enumeration test confirms all event error paths return identical generic message
- [x] All existing tests pass

Fixes #109

PR #986, opened 2026-04-24
feat(v0.7.0): dj-prefetch + @server_function (navigation & RPC primitives)

johnrtipton:
## Summary

First v0.7.0 batch — two complementary navigation/RPC primitives shipped in a single PR.

### 1. `dj-prefetch` on hover/intent
Extends `22-prefetch.js` with intent-based per-link prefetch via `<link rel="prefetch">`. 65ms mouseenter debounce, AbortController on mouseleave, immediate on touchstart. Same-origin only. Deduped.

### 2. Server functions — `@server_function` + `djust.call()`
RPC-style endpoint: `POST /djust/api/call/<view_slug>/<function_name>/`. Pure request/response, no re-render. Distinct from `@event_handler` (WebSocket + re-render) and `@event_handler(expose_api=True)` (external OpenAPI HTTP API). Dual-decoration rejected at decoration time. Strict wrapped-body shape `{"params": {...}}`; flat or mixed bodies return 400.

## Scope

- `@server_function` decorator + metadata (`python/djust/decorators.py` +95 LOC)
- Dispatch view + rate-limit integration (`python/djust/api/dispatch.py` +138 LOC)
- URL route `call/<slug>/<fn>/` BEFORE catch-all (`python/djust/api/urls.py` +7 LOC)
- Registry helpers (`python/djust/api/registry.py` +33 LOC)
- JS client module `48-server-functions.js` (NEW, 52 LOC)
- Intent-prefetch extension in `22-prefetch.js` (+113 LOC)
- Demo view + template (canonical XSS-safe rendering via `createElement` + `textContent`)
- CHANGELOG entries for both features
- User guides: `docs/website/guides/{prefetch,server-functions}.md`

## Tests

- `python/djust/tests/test_server_functions.py` — 16 cases (14 original + 2 regression for strict body shape). All pass.
- `tests/js/prefetch_intent.test.js` — 5 cases. All pass.
- `tests/js/server_functions.test.js` — 3 cases. All pass.
- Full suite: 6301 Python + 620 Rust + 1368 JS, **zero regressions**.
- +832 B gzipped on client.js (well within 2 KB/feature budget).

## Security

- `@server_function` is authenticated-by-default (anonymous → 401); stackable with `@permission_required`.
- CSRF enforced inside dispatch via `CsrfViewMiddleware.process_view`; test covers missing-token → 403.
- Strict body-shape validation; unknown-key shapes rejected with 400 `invalid_body`.
- DjangoJSONEncoder for response serialization (handles Decimal/datetime/UUID); exceptions logged server-side, generic message returned.
- Demo template rebuilt to use `createElement`/`textContent` (no `innerHTML` string concat) — demos are copy-paste material.
- Same-origin guard on `dj-prefetch`; `javascript:`/`data:` schemes blocked via Origin check.

## Review findings resolved

Stage 7 flagged 2 🟡:
1. Demo template innerHTML → fixed with DOM-building pattern.
2. Wrapped-vs-flat body ambiguity → tightened to accept only `{"params": {...}}` or empty body; 2 regression tests added.

Stage 8 security: zero CRITICAL/HIGH/MEDIUM.

## Test plan
- [ ] Merge, pull main, rebuild client, run `make test`
- [ ] Visit demo project at `/demos/server-function/` (create demo URL first — see examples/demo_project)
- [ ] Hover over a `<a dj-prefetch>` link in DevTools Network panel → confirm `<link rel=prefetch>` fired after ~65ms
- [ ] Type in demo search → confirm `djust.call()` returns product list, no re-render
- [ ] Post to `/djust/api/call/<view>/<fn>/` without CSRF → 403 csrf_failed

Closes: v0.7.0 P2 rows "Keep-Alive/dj-activity adjacent (prefetch hover)" and "Server functions (RPC)" from ROADMAP.md.


johnrtipton:
## Stage 11 Code Review — PR #986

### Verdict: APPROVE (with 2 SHOULD FIX notes — non-blocking)

Thorough, plan-aligned implementation. All three new symbols (`@server_function`, `djust.call()`, `dj-prefetch`) have non-test callers in the demo app. Wire protocol (flat `{"error": "<kind>", "message": "..."}`) is consistent between server (`api/errors.py:21`) and client (`48-server-functions.js:41-44`). Tests cover the highest-risk contracts: dual-decoration rejection at decoration time, strict body shape, CSRF-always-required, rate limit, async functions, and URL precedence.

### 🔴 MUST FIX

None.

### 🟡 SHOULD FIX

**1. Doc-vs-code drift: non-JSON-serializable return value does NOT produce the advertised `function_error` envelope.**

`docs/website/guides/server-functions.md:234` states: *"Anything that isn't serializable → 500 `function_error`. The `TypeError` is logged server-side...; the client sees only the generic message."*

But in `dispatch.py:568`:
```python
return JsonResponse({"result": result}, encoder=DjangoJSONEncoder)
```
The `JsonResponse` call is **outside** the invoke try/except (lines 556-566). If a handler returns a `set()` or other non-serializable value, `DjangoJSONEncoder` raises `TypeError` uncaught — Django's outer middleware converts it to a generic 500, *not* the `{"error": "function_error", "message": "..."}` JSON envelope the guide promises. The client's `djust.call()` will then see `err.code === 'http_error'` (fallback) rather than `function_error`, and `err.message` will be `djust.call failed: 500`.

Fix: wrap steps 10 and the serialize step together, or explicitly pre-serialize:
```python
try:
    return JsonResponse({"result": result}, encoder=DjangoJSONEncoder)
except TypeError:
    logger.exception("djust server_function: return value not serializable ...")
    return api_error(500, "function_error", "Function raised an unexpected error")
```

There is **no test** for this case — recommend adding one where the function returns `set([1,2])` and asserting `resp.status_code == 500` and `json.loads(resp.content)["error"] == "function_error"`.

**2. Hardcoded URL prefix ignores `FORCE_SCRIPT_NAME` / mounted sub-paths.**

`48-server-functions.js:25-27` constructs `'/djust/api/call/' + encodeURIComponent(viewSlug) + '/' + encodeURIComponent(funcName) + '/'`. If a djust app is deployed under a sub-path (e.g., `FORCE_SCRIPT_NAME=/app`, a common reverse-proxy pattern) or the integrator includes `djust.api.urls` at a non-root prefix, every `djust.call()` will 404.

This is consistent with the existing limitation in other client modules, so it's not a regression — but it's worth either:
- Adding a `window.djust.apiBase` hook the integrator can set before the bundle loads, OR
- Documenting the constraint explicitly in `docs/website/guides/server-functions.md` and the CHANGELOG.

Same concern applies implicitly to ADR-008's dispatch URL, so this could be a shared-infrastructure fix.

### 🟢 Observations / Nits

- **Pending prefetch timers not cleared on SPA nav.** `_pending` WeakMap in `22-prefetch.js:89` holds debounce timers; `window.djust._prefetch.clear()` (line 64) only clears the dedupe `Set`, not pending timers. After a `live_redirect`, a timer armed during hover on the old page could still fire `_doPrefetch` against a link that's been removed from the DOM. Low severity — the browser will still prefetch the URL, and the `<link rel=prefetch>` element attaches to the new document. Consider clearing pending timers in `_intentPrefetch.clear()`.
- **`mouseenter` on JSDOM does not bubble/capture.** The JS tests (`prefetch_intent.test.js:37`) bypass this by directly calling `_onEnter(ev)` — which is why `_onEnter` is exposed on the public surface. This is testing-via-implementation-detail rather than testing-via-event; acceptable for now, but real-browser testing (via djust-browser MCP) is the only way to catch a real regression in the capture-phase listener registration.
- **Touch-enter race.** `_onTouch` fires immediately and `_onEnter` also registers a debounce timer — on a touch device, a tap typically fires `touchstart` then `mouseenter` (~300ms later in some implementations). The dedupe set prevents a double-fetch but a scheduled timer remains in `_pending` until the next `mouseleave`. Not a bug (fetch dedup'd, timer GC'd on navigation), just an observation.
- **`is_server_function` accessed via `getattr(func, "_djust_decorators", {})`.** The helper builds a fresh `{}` every call. Minor — this is the same pattern used by `is_event_handler`. Fine.
- **CHANGELOG is thorough** and correctly distinguishes `@server_function` from `@event_handler(expose_api=True)` — specifically calls out the three-way intent split and the strict body-shape tightening.
- **`_instantiate_view` sets `_api_request = True`** for server-function calls too (reused from ADR-008). Views that branch on this flag in `mount()` will see the same behavior as HTTP API calls. If server functions should be distinguishable, consider a separate flag — but this isn't requested in the plan.

### What I verified

- **Wiring check** (scaffolding-no-plumbing test for each new symbol):
  - `@server_function` — exported in `python/djust/__init__.py:19`; imported by `examples/demo_project/djust_demos/views/server_function_demo.py:10`; applied at lines 55 and 63 on `search_products` / `get_time`. ✓ Non-test caller exists.
  - `djust.call` — defined in `static/djust/src/48-server-functions.js:51`; invoked from `examples/demo_project/djust_demos/templates/demos/server_function_demo.html:105` and :116. ✓ Non-test caller.
  - `dj-prefetch` — handled in `src/22-prefetch.js`; rendered in template at line 23 (`<a dj-prefetch href="/demos/counter/">`). ✓ Non-test caller.

- **Wire-protocol consistency**:
  - Server error envelope: `api_error()` at `python/djust/api/errors.py:21` returns `{"error": "<kind>", "message": "...", "details": {...}}` (flat).
  - Client error parsing: `48-server-functions.js:41-44` reads `data.message`, `data.error` (as `err.code`), `data.details`. Consistent — no nested/flat mismatch.
  - Success envelope: server returns `{"result": <val>}` at `dispatch.py:568`; client reads `data.result` at `48-server-functions.js:47`. Consistent.
  - Request body: client sends `{params: params || {}}` at `48-server-functions.js:36`; server accepts only `{"params": {...}}` at `dispatch.py:494`. Consistent.

- **Docs-vs-code spot check (3 claims from server-functions.md)**:
  1. "returns 403 on missing CSRF" → Verified. `test_csrf_required` at `test_server_functions.py:156-182` asserts 403 + `csrf_failed`.
  2. "rejects dual decoration at decoration time" → Verified. `test_dual_decoration_raises_at_decoration_time` at lines 259-283 covers both orderings.
  3. "strict `{\"params\": {...}}` body shape" → Verified. `dispatch.py:494` checks `set(body.keys()) == {"params"}`; `test_flat_body_rejected_with_400` and `test_body_with_params_plus_other_key_rejected` cover both rejection paths.
  4. (bonus) "Anything not serializable → 500 function_error" — **NOT verified** (see SHOULD FIX #1).

- **Request smuggling / state-leak check**: `_instantiate_view` at `dispatch.py:107` creates `view_cls()` fresh per request and sets `request` + `_api_request` on the instance. No class-level mutation, no persistent view instance. Safe.

- **Test coverage**: 14 Python tests cover auth, CSRF, body shapes, coercion, rate limit, async, URL precedence, dual-decoration. JS tests cover success/error + CSRF header. Gaps: non-serializable return, concurrent-request race (untestable without integration harness; flagging as SHOULD FIX #1 for the serialization case only).

### Files read in full
- `python/djust/static/djust/src/48-server-functions.js` (client entry)
- `python/djust/api/dispatch.py` (both dispatchers; server-function section at lines 422-568)
- `python/djust/api/errors.py` (envelope shape)
- `python/djust/api/registry.py` (added `_has_server_function` + `iter_server_functions`)
- `python/djust/api/urls.py` (URL precedence confirmed)
- `python/djust/decorators.py:60-325` (event_handler mutual-exclusion + server_function)
- `python/djust/tests/test_server_functions.py` (14 tests)
- `tests/js/server_functions.test.js` (3 tests)
- `tests/js/prefetch_intent.test.js` (5 tests)
- `python/djust/static/djust/src/22-prefetch.js` (both IIFEs)
- `docs/website/guides/server-functions.md` (307 lines)
- `docs/website/guides/prefetch.md` (161 lines)
- `CHANGELOG.md` (entries at lines 12-74)
- `examples/demo_project/djust_demos/views/server_function_demo.py` (demo)
- `examples/demo_project/djust_demos/templates/demos/server_function_demo.html` (demo template)

### Edge cases checked
- `djust.call()` on 204 No Content: `resp.json().catch` returns `{}`; `resp.ok` is true; returns `undefined` (from `data.result`). Acceptable — 204 is not a documented response, but the client degrades gracefully.
- `djust.call()` on 200 + non-JSON body (proxy error page): same as 204 — `data = {}`, returns `undefined`. Silent, but not broken.
- Same-link hover/leave/hover within 65 ms: second hover's `_onEnter` sees `_pending.has(link) === false` (leave cleared it) and re-arms the timer. Correct debounce reset.
- Dynamic `href` set via JS after page load: `_shouldIntentPrefetch` reads `link.href` at hover time — supported.
- Anonymous user with a valid session but `is_authenticated === False`: `dispatch.py:469` returns 401 `unauthenticated`. Correct.
- `mount()` raises: caught at `dispatch.py:509`, returns 500 `mount_failed`. Correct.

### Question raised
`_instantiate_view` sets `instance._api_request = True` for server-function calls as well. Should server functions have a distinct flag (e.g., `_server_function_call = True`) so mounts can distinguish between ADR-008 HTTP API calls and browser RPC calls? Not blocking — but worth a follow-up note if an integrator asks.

johnrtipton:
Addressed Stage 11 findings:

**#1 (non-serializable return → wrong error code)** — fixed at `python/djust/api/dispatch.py:567-582`. Serialization now runs inside its own try/except catching `TypeError` + `ValueError`; non-serializable returns surface with the documented 500 `function_error` envelope instead of Django's generic unhandled-500 `http_error`. Factored `JsonResponse` → explicit `json.dumps(..., cls=DjangoJSONEncoder)` + `HttpResponse` so the error path is deterministic. Regression test `test_non_serializable_return_returns_function_error` added (returns `{"bad": {1, 2, 3}}`; asserts status 500, error `function_error`, message contains `not JSON-serializable`).

**#2 (hardcoded URL prefix)** — kept minimal per review guidance. Added `TODO(v0.7.1, #987)` comment in `48-server-functions.js` documenting the FORCE_SCRIPT_NAME / sub-path-mount limitation; tracked in new tech-debt issue #987 (https://github.com/djust-org/djust/issues/987) so both this client and the existing ADR-008 client get fixed together in v0.7.1.

Tests: all 17 `test_server_functions.py` cases green. `client.js`/`client.min.js` rebuilt to reflect the JS comment.

johnrtipton:
## Pipeline Retrospective — PR #986

### What shipped

Two v0.7.0 navigation-and-RPC primitives in one batched PR: `dj-prefetch` (intent-based hover/touch prefetch with 65 ms debounce + AbortController on mouseleave + touchstart + same-origin guard) and `@server_function` / `djust.call()` (RPC-style POST to `/djust/api/call/<view>/<fn>/` reusing the ADR-008 dispatch router from v0.5.1). +832 B gzipped client delta. First task of the v0.7.0 milestone.

### Quality rating

**4.5 / 5** — Clean 14-stage run with zero stage loops, zero 🔴, zero security findings, 100% CI on first attempt. Not a 5 because Stage 11 caught a real doc-vs-code correctness defect (non-serializable return wasn't actually producing the `function_error` envelope the guide promised) that pre-commit Self-Review should have caught via data-flow trace — recurrence of the v0.6.1 doc-accuracy pattern.

### Review stats

| Metric | Count |
|---|---|
| Python tests added | 16 |
| JS tests added | 8 |
| 🔴 findings (Stage 7 + 8 + 11 combined) | 0 |
| 🟡 findings | 4 (2 Self-Review, 0 Security, 2 Code Review) |
| Findings addressed before merge | 3 of 4 (1 deferred → #987) |
| Pre-commit attempts | 1 (clean) |
| CI failures | 0 (12/12 required checks green) |
| Stage loops | 0 |
| Bundle delta | +832 B gzipped (well under 2 KB/feature budget) |

### What worked

- **Batching as "client-server navigation primitives."** Prefetch (client-only) + server_function (thin new dispatch on top of ADR-008) paired naturally by user-facing story. Both small-to-medium, both reused existing infrastructure, no shared files, and the combined PR stayed reviewable. Confirms batching is the right pattern for v0.7.0 primitive work.
- **Planner-first carried its weight again.** The 538-line plan flagged the URL-route-precedence risk (server_function dispatch must mount before ADR-008 generic dispatch or `/djust/api/call/*` gets swallowed) AND recommended implementing prefetch first (zero server surface, de-risks client build). Both predictions held up in implementation — zero route conflicts, clean client-first sequencing.
- **Reuse audit paid off on `@server_function`.** Planner identified that ADR-008's `_instantiate_view`, handler coercion, `@permission_required` wrapper, and rate limiter all drop in unchanged. Net new Python was ~100 LOC (decorator + dispatch function + URL) rather than a parallel API implementation.
- **Decoration-time conflict detection.** Raising `TypeError` when `@server_function` and `@event_handler` are stacked — at import time rather than request time — is the right failure mode. Test suite verified both decoration orderings.
- **Stage 11 as the doc-vs-code drift catcher.** Exactly the class of issue flagged as a recurring pattern in the v0.6.1 milestone retro. The review model keeps earning its keep.

### What could be improved

- **Doc-vs-code drift, recurrence of v0.6.1 pattern.** The server-functions guide promised non-serializable returns produce a `function_error` 500 envelope; the code wrapped `JsonResponse(..., encoder=DjangoJSONEncoder)` *outside* the invoke try/except, so `TypeError` propagated to Django middleware as a generic 500 without the promised JSON envelope. Sixth consecutive milestone with this finding class (v0.6.0 #969/#971/#972, v0.6.1 #975, now v0.7.0 #986). Action #116 exists but is clearly not working as a pre-check — implementers are still writing docs first, code second, without tracing the data flow. Consider escalating: **doc writing must come AFTER the error envelope test suite, not before.**
- **Stage 7 (Self-Review) systematically misses pre-existing patterns inherited from prior ADRs.** The hardcoded URL prefix (`/djust/api/call/...`) mirrors ADR-008's hardcoded dispatch URL shipped in v0.5.1 — which Stage 7 never flagged then either. It took Stage 11 to surface it. Observation: Self-Review tends to assume "inherited pattern = accepted pattern," but Stage 11 applies fresh eyes and notices the pattern is actually deployment-hostile (FORCE_SCRIPT_NAME, mounted sub-paths). Consider: when the implementation reuses an existing pattern, Self-Review should explicitly ask *should this pattern still exist?*, not just *did I copy it correctly?*
- **No test for non-serializable return pre-merge.** The fix commit `97d24692` added the try/except but the corresponding test coverage was added in the same commit — would have been caught in TDD if the test suite had included a deliberate "handler returns `set()`" case from the start. Consider adding an "error-envelope coverage" row to the Stage 7 checklist: every documented error kind in the guide must have an asserting test.
- **Pre-existing tech-debt shipped with the new feature.** The hardcoded `/djust/api/...` prefix now compounds across two features. Deferring to #987 is correct triage, but the debt is spreading faster than it's being paid down.

- **First v0.7.0 pipeline felt smoother than v0.6.1 equivalents.** Zero security findings, one clean pre-commit attempt, zero CI failures, zero stage loops. v0.6.1 batch PRs averaged 2–3 pre-commit iterations per PR. Leading hypothesis: the planner-first + reuse-audit pattern has matured to the point where implementation agents inherit good decisions for free.
- **Batching strategy validated for "same domain, disjoint files" pairings.** Prefetch touches only `22-prefetch.js` + template + doc; server_function touches `decorators.py` + `api/dispatch.py` + `48-server-functions.js` + template + doc. Zero file overlap meant the two features didn't interfere during implementation or review. Good rule of thumb for future v0.7.0 batches: if the two features share zero source files, batch them.
- **Stage 11 vs Stage 7 specialization is sharpening.** Stage 7 catches implementation-level scaffolding and stub defects. Stage 11 catches contract-level doc/code drift and inherited-pattern tech debt. The division of labor is now predictable enough to codify: Stage 7 = "did you wire it?", Stage 11 = "does it do what you said it does, and is the pattern still right?"
- **CodeQL py/bad-html-filtering-regexp did NOT fire this PR** — first v0.6.x/v0.7.x PR in weeks to not hit it. Reason: no regex HTML masking in this scope. Action #121 (shared `_SCRIPT_CLOSE_TOLERANT_RE`) remains open but not load-bearing for this batch.

### Action items

- [x] #987 filed — FORCE_SCRIPT_NAME / mounted sub-path support for JS clients (v0.7.1 target). **New Action Tracker row #123 added.**
- [ ] Consider escalating Action #116 (doc-accuracy data-flow trace): rather than "trace before you write docs," require "write error-envelope tests before you write the error docs." Sixth consecutive hit of this pattern.
- [ ] No further new action items this PR — Stage 11's Finding #1 was fixed in `97d24692` with a regression test; observations in Stage 11 (pending-timer clearing, JSDOM mouseenter, touch-enter race) are all explicitly tagged low-severity and acceptable.

### Kudos

- **Stage 11 reviewer caught a real correctness defect that Stage 7 missed.** The `JsonResponse(..., encoder=DjangoJSONEncoder)` outside the try/except was a *functional* bug — not a style nit. Without the catch, any handler returning a `set()`, `datetime`, generator, or other non-`DjangoJSONEncoder`-serializable value would have silently violated the advertised error envelope contract. Client-side error handling code would have branched on `err.code === 'http_error'` when the guide told users to branch on `'function_error'`. That's exactly the doc-vs-code drift class the three-layer review model is designed to catch. Textbook Stage 11 save.
- **Planner's URL-precedence call-out held up.** Zero route collisions in testing. Predictive risk analysis > reactive debugging.
- **Batch-PR discipline.** Two features in one PR, both with complete tests, docs, CHANGELOG entries, demo app integration, and zero cross-feature coupling. Worth repeating.

PR #988, opened 2026-04-24
feat(v0.7.0): {% dj_activity %} block tag + ActivityMixin (React 19.2 Activity parity)

johnrtipton:
## Summary

Second v0.7.0 batch — React 19.2 `<Activity>` parity via a new `{% dj_activity %}` Django block tag and `ActivityMixin` on `LiveView`. Pre-render hidden regions of a LiveView and preserve their local state (form values, scroll position, focus) across show/hide cycles.

## Scope

- `{% dj_activity "name" visible=expr eager=expr %}...{% enddj_activity %}` block tag (live_tags.py)
- `ActivityMixin` with `set_activity_visible(name, visible)`, `is_activity_visible(name)`, `eager_activities: frozenset` class attr, and a per-activity FIFO-capped (100) deferred-event queue
- Same-round-trip awaited flush via new `_dispatch_single_event` helper on `LiveViewConsumer` (no render-lock deadlock, no fire-and-forget)
- `49-activity.js` client: single MutationObserver on `hidden` attribute; `window.djust.activityVisible(name)` helper; `djust:activity-shown` CustomEvent on un-hide; re-scan on `djust:morph-complete`
- Two-step event gate in `11-event-handler.js`: drops events from any hidden non-eager ancestor (correct nested-activity behavior); stamps `_activity = closest ancestor name` for visible triggers
- Skip-subtree guard in `12-vdom-patch.js::applySinglePatch` for hidden non-eager activities
- System checks `djust.A070` (missing `name` argument) and `djust.A071` (duplicate names within one template)
- Demo: `examples/demo_project/djust_demos/views/activity_demo.py` + 3-tab template showing form-state preservation across tab switches

## Tests

- `tests/unit/test_activity.py` — 17 cases (12 functional + 2 XSS/escape + 2 A070/A071 variable-name + 1 same-round-trip flush). All pass.
- `tests/js/activity.test.js` — 6 cases including nested-activity event-drop regression. All pass.
- Full suite: 6306 Python + 620 Rust + 1373 JS. Zero regressions. 4 pre-existing markdown-missing failures unrelated to this PR.
- +643 B gzipped on client.js (within budget).

## Review findings resolved before commit

Stage 7 Self-Review flagged 3 red + 2 yellow:
1. red Deferred flush was fire-and-forget -> refactored to await inline via `_dispatch_single_event` helper
2. red Nested-activity gate used `closest([data-djust-activity])` (nearest wrapper, regardless of visibility) -> fixed to use `closest([hidden]:not([data-djust-eager]))` first; regression test added
3. red A070 false-positive on variable-name activities (`{% dj_activity panel_name %}`) -> regex extended to match bare identifiers; 2 regression tests added
4. yellow Deadlock foot-gun if flush refactored naively -> resolved by `_dispatch_single_event` never holding the render lock; docstring pins the requirement
5. yellow Docstring overclaim about ordering -> rewritten to match actual awaited-FIFO behavior

Stage 8 Security: zero CRITICAL/HIGH/MEDIUM. 2 LOW addressed:
- Added XSS regression test for `name` parameter
- Added "Security model" section to the guide documenting the `_activity` trust boundary

## Test plan

- [ ] Merge + pull main + run `make test` (baseline green)
- [ ] Start demo: `cd examples/demo_project && make start`; visit `/demos/activity/` — switch tabs; confirm form input persists across tab switches
- [ ] DevTools: inspect `[data-djust-activity]` — only one has no `hidden` attr at a time
- [ ] DevTools console: `djust.activityVisible('profile')` returns correct bool
- [ ] Fire a click inside a hidden activity via JS (`document.querySelector('[data-djust-activity][hidden] button').click()`) — confirm no WebSocket frame was sent
- [ ] Template with `{% dj_activity %}` no name → `manage.py djust_check` emits `djust.A070`
- [ ] Template with 2× `{% dj_activity "x" %}` → `djust.A071` error

Closes v0.7.0 P2 matrix row "Keep-Alive / dj-activity (React 19.2 `<Activity>` parity)" per ROADMAP.md.


johnrtipton:
## Stage 11 Code Review — PR #988

### Verdict: APPROVE (with 2 yellow follow-ups worth tracking for v0.7.1)

This is a well-executed Stage 7/8 fix-pass landing. The three highest-risk items (fire-and-forget flush, nested-activity gate, variable-name A070 false positive) have explicit regression tests. Same-round-trip drain invariant is preserved across all three flush call sites. The `_dispatch_single_event` / `_render_lock` non-reentrancy contract is correctly modeled. No 🔴 issues found.

### 🔴 MUST FIX

None.

### 🟡 SHOULD FIX (track for v0.7.1)

**1. Activity-gate runs BEFORE `_validate_event_security` in the queuing path — burst amplification risk.**

At `python/djust/websocket.py:2511-2527`, the activity gate pops `_activity`, checks visibility, and calls `_queue_deferred_activity_event` *before* `_validate_event_security` (which is at line 2554 for the actor path and the equivalent sync path below). The replay path (`_dispatch_single_event`, `websocket.py:1086`) does call `_validate_event_security` correctly, so an unauthenticated user cannot ultimately run a handler — but an attacker can still:

- Flood a hidden-activity queue up to `cap=100` per activity with zero rate-limit or permission checks on the queuing pass. `_queue_deferred_activity_event` accepts whatever event name the client sends.
- On the next legitimate show flip, the consumer dispatches all 100 in one round-trip — each dispatch runs `_validate_event_security` (so rejection still happens), but that's 100 rate-limiter + permission-check passes in a single CPU burst on the server, amplified per-session.

**Suggested follow-up:** rate-limit at queue-insertion time (cheapest fix), or validate event-name allowlist at queue time. Alternatively document the behavior in `docs/website/guides/activity.md#security-model` — the current docs claim "Real auth still runs on every event" (line 189), which is narrowly true only for the replay path. Add: "Queuing a deferred event skips rate-limit + permission checks by design (replay enforces both); the per-activity cap of 100 bounds the amplification."

**2. `docs/website/guides/activity.md` "Security model" section overstates invariant.**

Lines 189-192 read: *"`_validate_event_security`, `@permission_required`, handler allowlists, and rate limits apply to deferred/replayed events exactly as they apply to live ones."* This is accurate for the **replay** (drain) pass but **not** for the **queue** pass — and the queue pass is where an attacker lives. Tighten the wording to "…apply to deferred events **on replay**. A hidden-activity queue holds up to `cap=100` events per activity without re-validating on insert; the cap is the only burst limit."

### 🟢 Observations / Nits

- `python/djust/mixins/activity.py:190-196` rebuilds the `deque` if `activity_event_queue_cap` is mutated at runtime. Nice defensive touch; worth noting this copy runs **on every append** when `queue.maxlen != cap`, not only when the cap actually changes — since once rebuilt, the maxlens match on subsequent appends, this is a one-time cost, so no action needed.
- `49-activity.js` observes `document.body`. Activities inside iframes or shadow DOM are NOT observed. Acceptable for v0.7.0 — call out in a future guide update if shadow-DOM component usage grows.
- `_is_activity_eager` walks `_djust_activities` as well as `eager_activities`. The declarative + per-tag union is correct. Good.
- VDOM patch early-return returns `true` at `12-vdom-patch.js:1220` — verified correct (won't trigger the `failedCount++` branch at line 1518-1521).
- `_flush_deferred_activity_events` catches + logs per-event exceptions so one bad replay can't break the remaining flush (`activity.py:266-271`). Good robustness.
- The duplicate-name-within-one-render contract is handled **only statically** (A071) — runtime just overwrites silently (`activity.py:109`). This is simpler than the plan's "_activities_registered_this_render raises" design and I think it's the right call: the static check catches the common case, and runtime overwrites are benign (last-render-wins).

### What I verified

- **Wiring (225 refs):** `ActivityMixin` → imported in `python/djust/live_view.py:47` + MRO at line 149. `_init_activity()` called at `live_view.py:395` (right after `_init_sticky`). `dj_activity` tag used in `examples/demo_project/djust_demos/templates/demos/activity_demo.html`. `activityVisible` documented in `guides/activity.md:81-90`. A070/A071 scanner `python/djust/checks.py:2095-2163` wired into `check_templates`.
- **Same-round-trip flush:** three await call sites:
  - actor path: `websocket.py:2627-2631` — `await target_view._flush_deferred_activity_events(self)` ✓
  - non-actor path: `websocket.py:3277-3286` ✓
  - `db_notify` path: `websocket.py:4457-4467` ✓
  - All three `await`, none use `create_task`.
- **Deadlock analysis:**
  1. `handle_event` acquires `self._render_lock` at `websocket.py:2645`.
  2. Handler runs inside the lock.
  3. After handler finishes, `_flush_deferred_activity_events` is called (still inside the lock).
  4. Flush calls `consumer._dispatch_single_event(...)`.
  5. `_dispatch_single_event` docstring at `websocket.py:1062-1065` pins: "caller MUST already hold `self._render_lock`. This method does NOT acquire or release it."
  6. Code at lines 1086-1218 confirms: no `self._render_lock.acquire()` / `.release()` anywhere in the helper.
  7. Lock is released at `websocket.py:3304` (non-actor path), `4469` (db_notify). Actor path doesn't hold `_render_lock` during its try/finally — the flush at line 2629 runs without the lock.
  8. Non-reentrant `asyncio.Lock` behavior preserved. ✓
- **Nested-activity gate:** `11-event-handler.js:29-54` uses two-step logic: `closest('[data-djust-activity][hidden]:not([data-djust-eager="true"])')` first → drop on match, then `closest('[data-djust-activity]')` → stamp `_activity`. Regression test `tests/js/activity.test.js:140-168` asserts outer-hidden + inner-visible drops. ✓
- **Variable-name regex:** `_DJ_ACTIVITY_NAME_RE` at `checks.py:1922-1924`. Verified matches: `"name"` → group 1; `'name'` → group 2; `name` → group 3; `view.panel_name` → group 3 (dotted OK); `name=` prefix stripped. Does NOT match `123abc` (digit start) or empty — correctly triggers A070 for truly-missing names. Python keywords (`if`) match as identifiers, but Django's `compile_filter` handles those so no problem at render. ✓
- **Queue eviction:** `activity.py:186-200` — `collections.deque(maxlen=cap)`. FIFO eviction via `popleft()` in `_flush`. `cap <= 0` is treated as opt-out (line 184). Test `test_deferred_queue_caps_at_100` locks cap=100 behavior. ✓
- **Docs-vs-code (Security model):** three claims checked:
  1. "`_validate_event_security` runs on replay" — **TRUE** at `websocket.py:1086`. ✓
  2. "`_activity` param is client-supplied; absent/unknown → treated as visible" — **TRUE** at `websocket.py:2512` (the guard is `_activity_name AND hidden AND not eager`; absent activity name short-circuits the whole defer path). ✓
  3. "Real auth runs on every event" — **PARTIALLY TRUE**. Replay yes, queue insertion no. Flagged 🟡 above.

### Files read in full

- `python/djust/mixins/activity.py` (274 lines)
- `python/djust/static/djust/src/49-activity.js` (163 lines)
- `python/djust/static/djust/src/11-event-handler.js` (200 lines)
- `python/djust/static/djust/src/12-vdom-patch.js` (lines 1200-1235 + call-site 1510-1529)
- `python/djust/websocket.py` (lines 1040-1220, 2440-2640, 3260-3310, 4390-4475)
- `python/djust/templatetags/live_tags.py` (lines 856-993, activity tag block)
- `python/djust/checks.py` (lines 1900-2165, A070/A071 scanner)
- `python/djust/live_view.py` (lines 40-150, 380-405)
- `docs/website/guides/activity.md` (228 lines)
- `tests/unit/test_activity.py` (533 lines)
- `tests/js/activity.test.js` (196 lines)
- `.pipeline-state/feat-v070-dj-activity-plan.md` (plan for context)

### Edge cases checked

- **Activity name is Python keyword** (`{% dj_activity if %}`) — regex matches as identifier (group 3), no A070, Django's `compile_filter` handles it as a variable. ✓
- **Dotted identifier** (`view.panel_name.sub`) — regex group 3 matches. ✓
- **Digit-prefixed name** (`123abc`) — regex rejects → A070 fires. Correct behavior (invalid Python identifier).
- **Queue cap = 0** — treated as opt-out; events discarded silently. Deliberate per activity.py:184. ✓
- **Mid-drain handler flips activity back to hidden** — inner `while queue:` loop continues popping until empty (not re-checking visibility per-event); queue is fully drained regardless. Documented as intentional at `activity.py:251-253`. ✓
- **Handler raises during drain** — caught at `activity.py:266-271`, logged, loop continues. ✓
- **Activity wrapper removed by VDOM patch** — `djust:morph-complete` triggers `_scan(document)` which `.clear()`s and re-populates. Stale entries purged. ✓
- **Shadow DOM / iframe activity** — NOT observed (document.body is the observer root). Acceptable for v0.7.0, worth future-guide callout.
- **VDOM patch early-return** — returns `true`, so the "N patches failed" error branch at `12-vdom-patch.js:1523` doesn't false-fire for skipped subtrees. ✓
- **Concurrent activity flip during drain** — `list(queues.keys())` snapshot at `activity.py:241` avoids RuntimeError. ✓
- **Unknown activity name in `_activity` param** — `is_activity_visible` returns `True` for unknown names (mixin.py:154), so the server treats it as visible → normal dispatch. Defense-in-depth.
- **Two activities with same name in one render** — runtime just overwrites (last wins); A071 catches it statically at dev time. Documented at mixin.py:96-97. ✓

johnrtipton:
Addressed Stage 11 findings:

🟡 #1 (queue insertion precedes auth) — Tightened docstring on `_queue_deferred_activity_event` to accurately state: insertion is not auth-checked; auth runs on dispatch via `_dispatch_single_event`; 100-cap bounds memory. Inline code comment at the queue-insertion site in websocket.py.

🟡 #2 (docs overstatement) — Rewrote the "Security model" section in `guides/activity.md` to precisely describe the two auth phases (insertion: WS/CSRF/connection; dispatch: full handler-level auth stack).

No behavior change. Queue is still correctly auth-gated at the dispatch boundary.

johnrtipton:
## Pipeline Retrospective — PR #988

### What shipped

`{% dj_activity "name" visible=expr eager=expr %}` block tag + `ActivityMixin` + `49-activity.js` client + A070/A071 system checks — React 19.2 `<Activity>` parity for djust. Pre-renders hidden regions and preserves their local state (form values, scroll, focus) across show/hide cycles; events from hidden activities are deferred and drained in the same round-trip as the show flip. Second task of the v0.7.0 milestone.

### Quality rating

**4 / 5** — Clean 14-stage run with zero stage loops and 100% CI green on first attempt. Not higher because Stage 7 (Self-Review) returned the most serious outcome of the session (3 🔴 + 2 🟡), with two of the three 🔴 rooted in doc-vs-code drift — the same recurring pattern from v0.6.0/v0.6.1/#986. Fix-pass was textbook (single commit, 6 new regression tests, zero regressions), which is why this lands at 4 rather than lower.

### Review stats

| Metric | Count |
|---|---|
| Python tests added | 17 |
| JS tests added | 6 |
| 🔴 findings (Stage 7+8+11 combined) | 3 |
| 🟡 findings | 4 (2 Self-Review, 0 Security HIGH/MED, 2 Code Review) |
| Findings addressed before merge | 7 of 7 |
| Pre-commit attempts | 1 clean (after ESLint unsafe-fix revert) |
| CI failures | 0 (12/12 required checks green) |
| Stage loops | 0 |
| Bundle delta | +643 B gzipped (well under 2 KB/feature budget) |

### What worked

- **Stage 7 (Self-Review) was load-bearing — single biggest save of the session.** All functional tests passed at Stage 5. Pre-commit hooks were clean. Security scan was clean. Yet Stage 7 caught three correctness 🔴s invisible to any automated gate: (1) `loop.create_task` fire-and-forget on the deferred flush violating the documented same-round-trip contract, (2) `closest('[data-djust-activity]')` matching nearest ancestor regardless of visibility rather than the documented hidden-ancestor-anywhere semantics, (3) A070 false-positive treating bare-identifier activity names as missing. None of these would have been caught by pytest — they required a reader comparing code against plan claims. Strongest signal yet that Stage 7 is non-optional.
- **Combined fix-pass handled all 7 findings in one commit.** 5 bugs + 2 doc-tightenings + 6 new regression tests (5 Python, 1 JS) shipped in a single fix push. No follow-up iterations needed. Third consecutive PR where the single-commit combined fix-pass pattern has held — now the established norm for mid-session rework.
- **Planner-first corrected 4 bad briefing assumptions.** Planner caught that the tag lives in `live_tags.py` (not `djust.py`), the next free JS slot is 49 (not 24 as briefed), the event-binding module is `11-event-handler.js` (not `09-event-binding.js`), and no `_FRAMEWORK_INTERNAL_ATTRS` touch was needed. Implementing straight from the briefing would have hit all four errors in sequence. Planner-first continues to pay for itself.
- **Deferred-event flush refactor produced a reusable primitive.** Breaking the lock-acquisition cycle required a new `_dispatch_single_event` internal helper on `LiveViewConsumer` with an explicit contract: "caller MUST already hold `self._render_lock`." This helper is now the reference pattern for any future "async dispatch path" feature that needs to drain queued events inside an already-locked handler. Documented at `websocket.py:1062-1065`.
- **Stage 8 (Security) passed first attempt.** Zero CRITICAL/HIGH/MEDIUM findings on a feature touching event dispatch, DOM observers, and a new client-server queue protocol. Security-scan discipline continues to hold.

### What could be improved

- **Doc-vs-code drift is now a 4-milestone-consecutive pattern (v0.6.0, v0.6.1, #986, #988).** Two of the three 🔴s in Stage 7 — the nested-gate bug and the A070 false-positive — were both "docs promise X but code does Y" defects. The fix-pass for #986 added Action #116 to the tracker ("trace data-flow before writing user-facing docs"), but #116 was NOT applied to this PR. The root cause is that #116 is an aspirational process-reminder, not an executable gate. Proposed upgrade: for every feature with non-trivial semantics (gate rules, error envelopes, state contracts), write tests that EXECUTE the documented claim verbatim BEFORE writing the implementation. This is TDD sharpened — the test cases ARE the doc claims in executable form. Filing as a new tracker row.
- **Stage 7's own checklist should have a "gate-rule test row"** mirroring the proposed error-envelope row from PR #986's retro. Both the nested-gate and A070 🔴s could have been caught if Stage 7 asked "for each documented rule, does a test assert the rule verbatim?" — the nested-gate doc claim ("hidden-ancestor anywhere defers") was false per code but not represented in any test; likewise the "bare-identifier names are valid" doc claim was unchecked.
- **ESLint unsafe-fix almost caused a silent regression.** Pre-commit initially auto-applied an ESLint unsafe-fix that changed `49-activity.js` behavior; had to be reverted before the fix-pass. Worth documenting: the pre-commit hook chain should treat "unsafe-fix auto-applied" as a STOP signal requiring manual re-review, not a silently-passed hook. Not escalating to a tracker row because this is a one-off, but flagging for the next pipeline-skill-update conversation.
- **Pre-existing tech-debt spreading faster than paydown, continued.** No new tech-debt shipped with this PR (clean wrt Action #123 — no server-function-related work), but the systemic pattern flagged in #986's retro (FORCE_SCRIPT_NAME drift) remains open. Unchanged this PR.

- **Stage 7 vs Stage 11 specialization continues to sharpen.** PR #986 showed Stage 11 catching contract-level doc/code drift; PR #988 showed Stage 7 catching three implementation-level correctness bugs invisible to pytest + pre-commit. Division of labor is now: **Stage 7 = "does the code do what the plan said?"; Stage 11 = "does it do what the user-facing docs said?"** These are different questions requiring different contexts (plan vs guide), and both are load-bearing.
- **Every v0.7.0 feature so far has exposed a doc-vs-code drift bug.** PR #986 (server_function): `function_error` envelope promise unimplemented. PR #988 (dj_activity): hidden-ancestor-anywhere gate rule unimplemented + bare-identifier A070 false positive. If this continues through PR #989+, we'll have a 3-for-3 milestone pattern and the proposed "doc-claim-verbatim tests" discipline becomes non-optional.
- **Combined fix-pass pattern now validated across three milestones.** v0.6.1 (PR #976 snapshot-aliasing + dead-WS-path, combined fix), #986 (non-serializable return + doc-tighten, combined fix), #988 (3 🔴 + 2 🟡 + doc-tighten, combined fix). Pattern: when a review stage returns multiple findings, do NOT iterate multiple fix commits — one consolidated commit with regression tests for each finding, one push, one CI cycle. Three data points is enough to codify this in the pipeline-shared skill if not already done.
- **First time `_render_lock` non-reentrancy contract was exercised by new feature code.** The flush refactor had to obey a contract that had never before been a concern at the feature level. Worth noting: the contract was only documented implicitly (via `asyncio.Lock` being non-reentrant by default) until this PR added an explicit docstring at `websocket.py:1062-1065`. Future features that dispatch events from within an already-locked handler should be directed to this docstring as the reference contract.

### Action items

- [ ] **New Action Tracker row #124**: Upgrade Action #116 — for every feature with non-trivial semantics (gate rules, error envelopes, state contracts), write doc-claim-verbatim tests BEFORE writing the implementation. Trigger: 4th consecutive milestone with doc-vs-code drift 🔴/🟡. Enforcement: Stage 7 checklist should grow a row "for each documented rule in the guide/docstring, point to the asserting test."
- [ ] No change to row #123 (FORCE_SCRIPT_NAME) — no server-function-related work this PR.
- [ ] Consider: pipeline-shared skill should add a "unsafe-fix auto-applied is a STOP signal" note to the pre-commit stage. Not escalating to a tracker row yet (first occurrence).

### Kudos

- **Stage 7 Self-Review catching 3 🔴 correctness bugs — single biggest save of the session.** The fire-and-forget flush would have broken the same-round-trip guarantee silently (no test failure, just user-visible latency after activity shows). The nested-gate bug would have silently deferred events from the wrong elements (most visible case: a hidden parent with a visible child — click the child, event drops silently). The A070 false-positive would have generated false warnings on every legitimate `{% dj_activity name %}` variable-name use. All three invisible to pytest, pre-commit, and Stage 8 Security. Textbook Stage 7 save — and the first time this milestone where Stage 7 outperformed Stage 11 on bug count.
- **Single-commit fix-pass with 6 new regression tests.** Every 🔴 finding has a test that would fail pre-fix. No "fix the bug but skip the test" shortcuts. This is the disciplined rework pattern the pipeline was designed for.
- **Planner correcting 4 briefing errors before implementation started.** Four errors that would have cost cycles during implementation were caught in ~30s of codebase-reading during planning. Planner-first continues to be the highest-leverage stage in the pipeline.

PR #989, opened 2026-04-24
feat(v0.7.0): Django admin widgets + bulk-action progress (DjustModelAdmin slots)

johnrtipton:
## Summary

Third v0.7.0 batch — ship per-page widget slots on `DjustModelAdmin` + a `BulkActionProgressWidget` with the `@admin_action_with_progress` decorator. Closes the "add LiveView-powered widgets to admin" djust differentiator.

## Scope decision (Plan §1)

Option D+B per plan. **NOT shipped**: `DjustAdminMixin` for stock `admin.ModelAdmin`. Rationale: stock ModelAdmin has 20+ mixin methods — wiring a LiveView through that MRO duplicates 60% of admin_ext/ and contradicts the v0.5.0 recommended adoption path (`DjustAdminSite`). Per-model widget slots fit naturally into `DjustModelAdmin`; migration path from stock admin is documented.

### `DjustModelAdmin` widget slots
- `change_form_widgets: list[type[LiveView]] = []` — embedded on change_form page
- `change_list_widgets: list[type[LiveView]] = []` — embedded on changelist page
- `get_change_form_widgets()` / `get_change_list_widgets()` filter by `permission_required`
- Template partials `_change_form_widgets.html` / `_change_list_widgets.html` wired into the existing `model_detail.html` / `model_list.html` templates

### Bulk-action progress
- `BulkActionProgressWidget(LiveView)` — reports progress via polling loop, cooperative cancellation
- `@admin_action_with_progress(description=, permissions=)` — wraps an action; returns `HttpResponseRedirect` to a live progress page
- `ModelListView.run_action` now enforces the decorator's `allowed_permissions` server-side
- Process-local `_JOBS` `OrderedDict` capped at 500 (LRU eviction); `Job.message` / `Job.error` truncated to 4KB
- Generic user-facing error (raw exception stays server-side in `logger.exception`)
- `DJUST_ASGI_WORKERS` setting gates the A073 multi-worker check

### System checks
- `djust.A072` — non-LiveView subclass in `change_form_widgets` / `change_list_widgets` → warning
- `djust.A073` — progress widget + `DJUST_ASGI_WORKERS > 1` → info (single-worker default: silent)

### Tests (25 new, following Action Tracker #124 TDD discipline)

All 5 rule tests written RED before implementation and traced to GREEN post-impl:
- `test_bulk_progress_job_cancel_flips_done_and_cancelled_flags`
- `test_bulk_progress_non_owner_user_gets_403`
- `test_bulk_progress_non_staff_gets_403`
- `test_widget_slot_permission_required_hides_widget`
- `test_widget_slot_nonliveview_emits_A072`

Plus 20 coverage tests: LRU eviction, message truncation, cooperative cancel, generic-error, allowed_permissions enforcement, A073 worker-count gate, create-view omits object_id, XSS content check, etc.

Full suite: 6328 Python + 620 Rust + 1374 JS, zero regressions (4 pre-existing markdown-module failures unchanged).

## Review findings resolved before commit

Stage 7 Self-Review flagged 4 🟡 (all resolved):
1. Doc-vs-code log-level drift (WARNING vs ERROR) → docs corrected; test locks the claim
2. A073 semantic mismatch → gated on `DJUST_ASGI_WORKERS` setting
3. Permission filter only tested at list level → rendered-HTML test added
4. Create-view path passes empty `object_id` → template conditionally emits kwarg; regression test added

Stage 8 Security flagged 2 MEDIUM + 3 LOW (all resolved):
- MEDIUM: `cancel()` doesn't stop background thread → cooperative `progress.cancelled` pattern documented + demo updated
- MEDIUM: `_JOBS` unbounded → LRU cap at 500, message truncation at 4KB
- LOW: `job.error` leaks raw exceptions → generic user-facing message, raw preserved server-side
- LOW: `run_action` missing `allowed_permissions` enforcement → added server-side check
- LOW: `live_render` allowlist not documented → DoD section added

Zero 🔴, zero CRITICAL/HIGH.

## Known limitation

`_JOBS` is process-local. Multi-worker deploys without sticky sessions will see "Job not found" if the redirect hits a different worker. `DJUST_ASGI_WORKERS` setting triggers A073 info check. v0.7.1 will back with channel layer.

## Test plan

- [ ] Merge, pull main, run `make test`
- [ ] Start demo: `cd examples/demo_project && make start`
- [ ] Visit `/djust-admin-demo/<app>/maintenancerequest/` — confirm widget embedded
- [ ] Fire "Close selected" action — confirm redirect to progress page; progress bar updates
- [ ] Click Cancel mid-action — confirm job halts (action body checks `progress.cancelled`)
- [ ] User without `change` perm fires `close_selected` — confirm 403
- [ ] Set `DJUST_ASGI_WORKERS=2`, run `manage.py check` — confirm A073 info emitted

Closes v0.7.0 P2 matrix row "Django admin LiveView widgets" per ROADMAP.md.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,57 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- **Admin widgets & bulk-action progress (v0.7.0)** — two additions to
+  `djust.admin_ext` close the most-requested gaps in the alternative
+  reactive admin:
+  - `DjustModelAdmin.change_form_widgets` / `change_list_widgets` class
+    attributes accept any list of `LiveView` subclasses; each is
+    embedded via `{% live_render %}` on the matching admin page.
+    Permission filtering honours `permission_required` on the widget
+    class. See
+    [docs/website/guides/admin-widgets.md](docs/website/guides/admin-widgets.md).
+  - `@admin_action_with_progress` (in `djust.admin_ext.progress`) turns
+    any `DjustModelAdmin` action into a background daemon thread and
+    redirects the user to a `BulkActionProgressWidget` page at
+    `<admin>/djust-progress/<job_id>/`. The page polls the job every
+    500 ms, re-renders the progress bar / message / log, and wires a
+    Cancel button that atomically flips `done` and `cancelled`. Queryset
+    is eagerly pinned to PKs before the thread starts (no lazy-eval
+    foot-guns). **Cancellation is cooperative** — clicking Cancel flips
+    `progress.cancelled = True`; the action body must periodically check
+    `if progress.cancelled: return` to actually stop (Python cannot
+    safely interrupt a running thread mid-statement).
+  - **Server-side permission enforcement** — `@admin_action_with_progress(permissions=[...])`
+    stamps `allowed_permissions` on the wrapped action; `ModelListView.run_action`
+    now calls `request.user.has_perms(allowed)` before dispatching the
+    action and raises `PermissionDenied` if the user lacks any declared
+    perm. Closes the gap where `has_*_permission` returns True for any
+    staff user.
+  - **Bounded server state:** `_JOBS` is LRU-capped at `_MAX_JOBS = 500`
+    (oldest entries evicted on insert once the cap is reached), and
+    `Job.message` / `Job.error` are individually truncated to
+    `_MAX_MESSAGE_CHARS = 4096` on each `progress.update(...)` call.
+    `Job.error` is a generic user-facing string ("Action failed — see
+    server logs for details"); the raw exception text lives only on the
+    server-side `Job._error_raw` attribute and is always logged at ERROR
+    level via `logger.exception` (logger name `djust.admin_ext.progress`).
+  - **New setting:** `DJUST_ASGI_WORKERS` (default `1`) — declares the
+    number of ASGI workers in the deployment. Gates the A073 system
+    check (fires only when `DJUST_ASGI_WORKERS > 1`) so single-worker
+    development stays silent.
+  - **Defense-in-depth allowlist:** `DJUST_LIVE_RENDER_ALLOWED_MODULES`
+    (optional) restricts the dotted-path module prefixes that
+    `{% live_render %}` will resolve — any widget slot path outside the
+    allowlist raises `TemplateSyntaxError` at render time.
+  - Two new system checks: `djust.A072` (warning) fires if a non-
+    `LiveView` class is registered in a widget slot; `djust.A073`
+    (info, gated on `DJUST_ASGI_WORKERS > 1`) fires at startup if any
+    admin site hosts a `@admin_action_with_progress`-decorated action,
+    noting the v0.7.0 single-worker `_JOBS` limitation and pointing at
+    the v0.7.1 channel-layer follow-up.
+  - 25 new tests: `python/djust/tests/test_bulk_progress.py` (12) +
+    `python/djust/tests/test_admin_widgets_per_page.py` (13); +A072/A073
+    check tests in `python/tests/test_checks.py`.
 - **`{% dj_activity %}` + `ActivityMixin` (v0.7.0)** — React 19.2
   `<Activity>` parity: pre-rendered hidden regions of a LiveView that
   preserve their local DOM state (form inputs, scroll, transient JS)
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -905,7 +905,7 @@ class DashboardView(LiveView):
         # Charts panel pre-renders data in background
 ```
 
-**Django admin LiveView widgets** — Drop-in LiveView-powered widgets for Django's admin interface. `DjustAdminMixin` on any `ModelAdmin` enables real-time dashboards, live search/filter, inline editing, and bulk action progress within the admin. Use cases: real-time order status dashboards, live log viewers, monitoring panels, AI-powered admin actions with streaming output. This is a unique djust differentiator — no other LiveView-style framework integrates with an existing admin like Django's. Implementation: admin template overrides + a `DjustAdminWidget` base class that renders a mini LiveView inside admin change forms/list views. ~300 lines Python. *Django's admin is used by 90%+ of Django projects. Making it reactive with zero config is the single most effective demo of djust's value proposition — "add one mixin and your admin goes live."*
+~~**Django admin LiveView widgets**~~ ✅ **Shipped in v0.7.0** — Per-page widget slots (`change_form_widgets`, `change_list_widgets`) on `DjustModelAdmin` + `@admin_action_with_progress` decorator + `BulkActionProgressWidget` LiveView with cancel / log / progress bar + system checks A072 (non-LiveView slot) and A073 (multi-worker note). Shipped as extensions to the existing `DjustAdminSite` (ADR-007 Phase 4 adoption path) rather than a `DjustAdminMixin` on stock `admin.ModelAdmin` — avoids duplicating 60% of admin_ext infrastructure. See [docs/website/guides/admin-widgets.md](docs/website/guides/admin-widgets.md). Channel-layer backend for multi-worker `_JOBS` deferred to v0.7.1.
 
 **Prefetch on hover/intent** — Pre-load the next page's data when the user hovers over a link or shows navigation intent (mouse movement toward link, touch start). `<a dj-prefetch href="/dashboard">Dashboard</a>` triggers a lightweight prefetch request on hover, so the page loads instantly on click. Different from existing `22-prefetch.js` (which pre-fetches all visible links) — this is intent-based and targeted. Remix, Next.js, and Astro all use hover-prefetch as their primary strategy for fast navigation. Implementation: `mouseenter` listener with 65ms delay (avoids prefetch on fly-over), prefetch via `<link rel="prefetch">` or fetch API with abort on `mouseleave`. ~50 lines JS. *Combined with View Transitions API, this makes navigation feel literally instant — the page is already loaded before the user clicks.*
 
@@ -1042,7 +1042,7 @@ Open questions that inform future direction:
 | Islands of interactivity | — | Astro islands | Not started | v0.7.0 |
 | AI streaming primitives | — | — | Not started | v0.7.0 |
 | Server functions (RPC) | — | Server Actions | Not started | v0.7.0 |
-| Django admin LiveView widgets | — | — | Not started | v0.7.0 |
+| ~~Django admin LiveView widgets~~ | — | — | ✅ Shipped (v0.7.0) | v0.7.0 |
 | Prefetch on hover/intent | — | Remix prefetch | Not started | v0.7.0 |
 | **Keep-Alive / Activity** | — | **`<Activity>`** (19.2) | **Not started** | **v0.7.0** |
 | ~~**Document metadata**~~ | ~~`live_title`~~ | ~~**Native** (React 19)~~ | ✅ **Done** | v0.4.0 |
@@ -1123,7 +1123,7 @@ High-impact areas for contributions:
 39. ~~**Rust template engine parity**~~ ✅ — Closed in v0.5.0: getattr fallback, attr-context escape, assign-tag handler
 40. **AI streaming primitives** — Purpose-built LLM streaming components
 41. **Streaming initial render** — Chunked HTTP response with progressive content loading
-42. **Django admin LiveView widgets** — Real-time admin dashboards and inline editing
+42. ~~**Django admin LiveView widgets**~~ ✅ **Shipped in v0.7.0** — `change_form_widgets`/`change_list_widgets` slots + `@admin_action_with_progress` + `BulkActionProgressWidget` + A072/A073 checks. See `docs/website/guides/admin-widgets.md`.
 43. **Hot View Replacement** — State-preserving Python code reload in dev mode, ~200 lines Python
 44. **Server Actions (`@action`)** — React 19-style mutation handlers with auto pending/error states
 45. **Keyed for-loop change tracking** — Rust-side per-item change detection in `{% for %}` loops, ~200 lines Rust
--- a/docs/website/_config.yaml
+++ b/docs/website/_config.yaml
@@ -236,6 +236,11 @@ navigation:
         file: guides/server-functions.md
         order: 12.7
         level: advanced
+      - title: Admin Widgets (`change_form_widgets`, Bulk-Action Progress)
+        slug: admin-widgets
+        file: guides/admin-widgets.md
+        order: 12.8
+        level: intermediate
       - title: Deployment
         slug: deployment
         file: guides/deployment.md
--- a/docs/website/guides/admin-widgets.md
+++ b/docs/website/guides/admin-widgets.md
@@ -1,0 +1,345 @@
+# Admin Widgets — Per-Page LiveView Slots & Bulk-Action Progress (v0.7.0)
+
+`djust.admin_ext` ships a drop-in reactive replacement for Django's
+stock admin (`DjustAdminSite`, `DjustModelAdmin`, plugin system,
+dashboard widgets).  v0.7.0 adds two building blocks that close the gap
+with real-world admin requirements:
+
+1. **Per-page widget slots** on `DjustModelAdmin` — embed any LiveView
+   into a model's change-list or change-form page via
+   `change_list_widgets` / `change_form_widgets` class attributes.
+2. **Bulk-action progress** — turn any ModelAdmin action into a
+   background job with a live progress page using
+   `@admin_action_with_progress`.
+
+## Why this design (and why we didn't ship `DjustAdminMixin`)
+
+A lot of demos reach for "one mixin you sprinkle on `admin.ModelAdmin`
+and your admin goes live." We prototyped that.  In practice it
+duplicates ~60% of `DjustAdminSite`, fights Django's `change_view` /
+`changelist_view` MRO (20+ override points, complex
+`template_name` resolution, `ChangeList` introspection), and diverges
+from the recommended adoption path (ADR-007 Phase 4: `djust[admin]` =
+`DjustAdminSite`).
+
+Instead, v0.7.0:
+
+- Extends the existing `DjustModelAdmin` with **thin** per-page widget
+  slots that reuse the already-shipped `{% live_render %}` machinery.
+- Adds `BulkActionProgressWidget` as a first-class LiveView that any
+  ModelAdmin action can redirect to via a one-line decorator.
+
+Result: an admin page that wants a live revenue chart is
+
+```python
+class OrderAdmin(DjustModelAdmin):
+    change_list_widgets = [RevenueChartView]
+```
+
+...and the chart *is* a regular LiveView — same class you'd use on any
+page.  No new framework surface, nothing novel to learn.
+
+## Per-page widget slots
+
+### Quick start — `change_form_widgets`
+
+```python
+# myapp/djust_admin.py
+from djust import LiveView
+from djust.admin_ext import DjustModelAdmin, site
+from djust.decorators import event_handler, state
+from .models import Order
+
+
+class OrderActivityWidget(LiveView):
+    """Shows recent activity on the order currently being edited."""
+
+    template_name = "myapp/admin/order_activity.html"
+    label = "Recent activity"
+    size = "lg"  # "sm" | "md" | "lg" — controls grid column span
+
+    events = state(default=[])
+
+    def mount(self, request, object_id=None, **kwargs):
+        self.request = request
+        self.object_id = object_id
+        self.events = list(
+            Order.objects.get(pk=object_id).activity.order_by("-ts")[:10]
+        )
+
+    def get_context_data(self, **kwargs):
+        return {"events": self.events}
+
+
+@site.register(Order)
+class OrderAdmin(DjustModelAdmin):
+    change_form_widgets = [OrderActivityWidget]
+```
+
+That's it.  Open `/admin/myapp/order/42/change/` and the widget renders
+above the form, receives `object_id=42`, and behaves exactly like any
+other LiveView.
+
+### `change_list_widgets`
+
+Same deal, but on the list page. Widgets registered here receive no
+`object_id` (since the page isn't scoped to a single model instance).
+Typical uses:
+
+- Live KPI tiles (total orders today, open support tickets)
+- Filter-aware summaries (sum of currently-shown rows)
+- Pinned announcements / banners
+
+```python
+@site.register(Order)
+class OrderAdmin(DjustModelAdmin):
+    change_list_widgets = [TodayRevenueWidget, OpenTicketsWidget]
+```
+
+### Permissions
+
+Any widget class can declare `permission_required`. Users without the
+permission simply don't see the widget — no fallback placeholder is
+rendered.
+
+```python
+class InventoryAlertWidget(LiveView):
+    template_name = "myapp/admin/inventory_alert.html"
+    permission_required = "inventory.view_low_stock"
+    label = "Low stock alerts"
+```
+
+## Bulk-action progress
+
+### Quick start — `@admin_action_with_progress`
+
+```python
+# myapp/djust_admin.py
+from djust.admin_ext import DjustModelAdmin, site
+from djust.admin_ext.progress import admin_action_with_progress
+from .models import Order
+
+
+@site.register(Order)
+class OrderAdmin(DjustModelAdmin):
+    actions = ["refund_selected"]
+
+    @admin_action_with_progress(description="Refund selected orders")
+    def refund_selected(self, request, queryset, progress):
+        total = queryset.count()
+        progress.update(current=0, total=total, message="Starting refunds…")
+        for i, order in enumerate(queryset.iterator(), start=1):
+            order.refund()
+            progress.update(
+                current=i,
+                total=total,
+                message=f"Refunded order #{order.pk}",
+            )
+        progress.update(message="All done.")
+```
+
+When a user runs this action:
+
+1. The decorator pins the selected rows to a list of primary keys
+   (so lazy-queryset re-evaluation against the session can't affect
+   the thread).
+2. Spawns a daemon thread that runs the function body.
+3. Returns an `HttpResponseRedirect` to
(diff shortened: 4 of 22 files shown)


johnrtipton:
## Stage 11 Code Review — PR #989

### Verdict: REQUEST_CHANGES

The fix-pass is disciplined and resolves all 🟡/🟡 MEDIUM items cleanly, with real asserting tests for each semantic claim (log level, A073 gating, permission filter, create-view object_id, LRU eviction, generic error + `_error_raw`, `run_action` permissions). Plan §1 "not building" list is fully honoured. Doc/code drift on log level, A073 gate, and generic-error story is now locked by tests.

**One 🔴** blocks merge: the headline WebSocket → progress-page redirect is wired via `run_action` returning an `HttpResponseRedirect`, but `run_action` is an `@event_handler`, not a Django view. The returned response is silently dropped on the server and the client never navigates. Everything downstream works in isolation (job is stored, thread runs, progress widget renders if visited directly), but the documented user flow (`PR body → "Fire 'Close selected' action — confirm redirect to progress page"`) can’t succeed as-written.

### 🔴 MUST FIX

**1. Redirect to the progress page is never dispatched to the browser.** `@admin_action_with_progress` returns `HttpResponseRedirect(progress_url)` from `wrapper(self, request, queryset)` (`python/djust/admin_ext/progress.py:376`). `ModelListView.run_action` invokes the action over WebSocket (`python/djust/admin_ext/templates/djust_admin/model_list.html:54` wires `dj-change="run_action(value)"` → `views.py:441` `result = action_func(self.request, queryset)`) and returns `result` from an `@event_handler`. Event-handler return values aren’t special-cased for `HttpResponseRedirect` anywhere in `websocket.py` / `live_view.py` — only the explicit `self.push_event("redirect", {"url": ...})` pattern used in `LoginView.do_login` (`views.py:711`) causes client navigation. Result: the user clicks the action, the background thread starts, but the browser stays on the changelist — the progress page is never visited, so `BulkActionProgressWidget` never mounts and Cancel is unreachable. There is no test covering this end-to-end path (`test_decorator_redirects_to_progress_url` asserts the decorator’s *direct* return is an `HttpResponseRedirect`, which is necessary-but-insufficient).

Fix options:
- Teach `run_action` to detect `HttpResponseRedirect` in the action result and call `self.push_event("redirect", {"url": result["Location"]})` (or set `self.redirect_url` + a state hook). Add an E2E test that asserts a `redirect` push-event is queued when a `@admin_action_with_progress`-decorated action is invoked.
- Or: have the decorator set `progress.redirect_to` on `self` via the admin instance’s LiveView context (tighter coupling, harder to test).

Either way, add a test that binds `ModelListView.run_action` to a progress-decorated action and asserts a navigation signal is produced.

### 🟡 SHOULD FIX

**2. `_store_job` + `_MAX_JOBS` eviction is not thread-safe under concurrent decorator calls.** `python/djust/admin_ext/progress.py:86-97` performs `_JOBS[job_id] = job` followed by `while len(_JOBS) > _MAX_JOBS: _JOBS.popitem(last=False)` without a `threading.Lock`. Two admin users firing actions at the same moment both run the `while` and both pop — the loop can over-evict (never under-evict, because `len()` and `popitem` are GIL-atomic individually). In practice a rare transient size undershoot; not data corruption, but it violates the "capped at exactly 500" contract that `test_jobs_dict_lru_evicts_beyond_cap` asserts. Low-severity but easy to fix: protect `_store_job` with a module-level `_JOBS_LOCK = threading.Lock()`; document the invariant. No new test required — the existing cap test still passes; add one that spawns N threads hammering `_store_job` and asserts `len(_JOBS) == _MAX_JOBS` exactly.

**3. `DJUST_ASGI_WORKERS` string coercion isn’t tested.** `checks.py:2590-2594` intentionally coerces `int(asgi_workers)` with TypeError/ValueError fallback so `DJUST_ASGI_WORKERS = "2"` (typical 12-factor env read) still gates A073. Only `int=2` is covered (`test_checks.py:3856`). The docs don’t advertise string support, but production deployments commonly use `os.environ["DJUST_ASGI_WORKERS"]` → string. Add `with override_settings(DJUST_ASGI_WORKERS="2"):` assertion to lock the coercion. Cheap, one-line guard against a future refactor dropping the `int()` cast.

**4. `_MAX_MESSAGE_CHARS` doesn’t cap `action_label`.** `Job.action_label` is set by the decorator from `wrapper.short_description` (`progress.py:338`, `progress.py:378`), which is user-provided via `@admin_action_with_progress(description=...)`. It flows through `_refresh_from_job` → `self.action_label` → `{{ action_label }}` in `bulk_progress.html:12`. No cap, no escaping concern (Django auto-escapes), but the per-message cap docs claim `Job.message / Job.error` are bounded to 4KB — a 10MB `description=` would bypass that. Either extend `_truncate` to `action_label`, or tighten the docs scope to "progress messages" only. Cheap: one-line `self.action_label = _truncate(job.action_label)` in `_refresh_from_job`.

### 🟢 Observations / Nits

- `BulkActionProgressWidget._poll` uses `time.sleep(30)` then `_JOBS.pop(self.job_id, None)` (`progress.py:230-231`). If two owner tabs are both open, whichever widget's poll reaches the 30s mark first pops the job; the other widget's next poll sees `job is None`, renders `self.error = "Job not found or expired."` — harmless but a tiny UX wart. Not worth fixing in this PR.
- `BulkActionProgressWidget.cancel` docstring says "both flags flip atomically" (`progress.py:249`). Under the GIL the two stores are individually atomic, but the *pair* is not: the action body's `if progress.cancelled:` check can observe `cancelled=True, done=False` between the two stores on a preemption boundary. Benign (action exits; `finally` eventually sets `done=True`), but the wording is imprecise.
- `Job.set_error(exc, *, user_message=None)` handles `user_message=None` correctly (falls through to `_GENERIC_ERROR_USER_MESSAGE` via `user_message or _GENERIC...`) but not `user_message=""` — empty string is falsy, so it also falls through to the generic. This is the correct behaviour for an empty user string but is a subtle trap if a caller wants to force-blank. Not a finding; would benefit from a docstring line.
- `_make_bulk_action_progress_view` uses module `__getattr__` for lazy construction (`progress.py:426-436`) — nice workaround for the circular import, but it breaks `from djust.admin_ext.progress import BulkActionProgressView` introspection tools (Sphinx autodoc, IDE completion). `__all__` lists it with `noqa: F822`; acceptable trade-off.
- `_widget_has_permission` returns `False` if `request.user is None` (`options.py:83-86`) — correct fail-closed. Good.
- `cancel()` implementation re-reads `_JOBS.get(self.job_id)` (`progress.py:254`) — if the job has already been popped by `_poll` after 30s, `cancel` silently no-ops. Not an issue: if the job is done, cancel is meaningless. Worth a comment though.
- Demo (`examples/demo_project/demo_project/djust_admin.py:86-88`) correctly implements cooperative cancel. Docstring on `close_selected` matches the docs guide. Good.
- A072 test checks the code is produced (`test_checks.py:3808`); no test checks the `fix_hint` content. Low risk; skill audit writer typically renders hints.

### What I verified

**Wiring (each exported symbol → non-test caller)**
- `BulkActionProgressWidget` → `progress.py:395` (`BulkActionProgressView` subclass) → `sites.py:160` (URL pattern). OK.
- `BulkActionProgressView` → `sites.py:129,160`. OK.
- `admin_action_with_progress` → `examples/demo_project/demo_project/djust_admin.py:73`; docs `admin-widgets.md:117-139`; exported in `admin_ext/__init__.py:44`. OK.
- `change_form_widgets` / `change_list_widgets` → `options.py:56-57`; read in `views.py:338,530`; demo at `djust_admin.py:68-69`. OK.
- `A072` → `checks.py:2624`; tested `test_checks.py:3808`. OK.
- `A073` → `checks.py:2662`; gate + emit both tested `test_checks.py:3837,3859`. OK.
- `_MAX_JOBS` / LRU → `progress.py:65,96`; test `test_bulk_progress.py:70-106`. OK.
- `_MAX_MESSAGE_CHARS` → `progress.py:66,146,169,170`; test `test_bulk_progress.py:54-68`. OK (applies to `message`, `error`, `_error_raw`; does NOT apply to `action_label` — see 🟡 #4).
- `DJUST_ASGI_WORKERS` → `checks.py:2590`; `docs/website/guides/admin-widgets.md:303-311`; tests `test_checks.py:3811,3839`. OK.
- `_error_raw` → `progress.py:124,169`; tests `test_bulk_progress.py:237-238, 333-334`. OK.
- `run_action` `allowed_permissions` enforcement → `views.py:436-438`; tests `test_admin_widgets_per_page.py:310-352, 354-410`. OK.

**Fix-pass re-verification (6 items)**
- Log-level: `logger.exception` at `progress.py:366`; `test_bulk_progress.py:243-295` (`test_progress_error_logged_at_error_level`) asserts `level="ERROR"` AND `exc_info is not None`. Doc match at `admin-widgets.md:253-260`. ✅
- A073 default-silent + multi-worker fire: both tests present (`test_checks.py:3811, 3839`). ✅
- Permission-filter rendered-HTML: `test_admin_widgets_per_page.py:216-260` exercises `get_change_list_widgets → _serialize_widget_slots` and asserts `secret_path not in blob`. Exercises the serialize path, not literal HTML rendering (no Django template engine call), which is acceptable since `_change_list_widgets.html` just iterates the serialized dicts. ✅
- Create-view empty `object_id`: template guard at `_change_form_widgets.html:9`; test at `test_admin_widgets_per_page.py:263-291` asserts key is absent. ✅
- Cooperative cancel: docs `admin-widgets.md:156-183`; demo updated at `djust_admin.py:86-88`; test `test_bulk_progress.py:345-392` uses threading events to prove early-exit. `progress.cancelled` is settable on `Job` by `BulkActionProgressWidget.cancel()` and read by the action body (shared Job instance). End-to-end path verified. ✅
- LRU eviction: `_store_job` at `progress.py:86-97`; test at `test_bulk_progress.py:70-106` fills to cap+1 and asserts `len == _MAX_JOBS` and first key evicted. Thread-safety caveat: see 🟡 #2. ✅ (with caveat)
- Error leakage: `_error_raw` vs `error` split; `test_action_error_user_facing_message_is_generic` at `test_bulk_progress.py:297-343` asserts secret absent from `job.error`, present in `_error_raw`, present in server log. ✅
- `allowed_permissions` enforcement placement: checked BEFORE `action_func(...)` call (`views.py:436-441`, between locating the callable and invoking it). Correct order. ✅

**Doc-vs-code sampling (6 claims, Action #124 application)**
- "Cancellation is cooperative; action body must check `progress.cancelled`" → `test_action_body_checking_cancelled_exits_early` (`test_bulk_progress.py:345`). ✅
- "Messages are truncated at 4096 chars" → `test_job_update_truncates_overlong_message` (`test_bulk_progress.py:54`). ✅
- "`_JOBS` is capped at 500 via LRU" → `test_jobs_dict_lru_evicts_beyond_cap` (`test_bulk_progress.py:70`). ✅
- "A073 fires only when `DJUST_ASGI_WORKERS > 1`" → both directions covered (`test_checks.py:3811, 3839`). ✅
- "`allowed_permissions` enforced server-side in `run_action`" → `test_run_action_enforces_allowed_permissions` (`test_admin_widgets_per_page.py:310`). ✅
- "Raw exception messages stay in server logs; users see generic" → `test_action_error_user_facing_message_is_generic` (`test_bulk_progress.py:297`) asserts BOTH raw in log AND raw absent from `job.error`. ✅

**Plan §1 "not building" list**: stock-`ModelAdmin` mixin absent; no `djust_dashboard_widgets`; no separate widget URL namespace; `AdminWidget` base untouched; no inline-editing via `LiveComponent`; no AI-powered actions; no `djust_gen_admin` management command. ✅

**Thread-safety analysis**
- `_store_job` race: see 🟡 #2.
- `Job` mutable state shared between `BulkActionProgressWidget._poll` (polling thread), `cancel()` (WS event thread), and action body (decorator thread). Reads of individual fields are GIL-atomic; no tearing. `_refresh_from_job` reads one field at a time — briefly inconsistent snapshot acceptable (the next poll in 0.5s reconciles). Not a finding.
- `_JOBS.pop(self.job_id, None)` in `_poll` vs `_JOBS.get(self.job_id)` in `cancel()`: `cancel()` silently no-ops on already-evicted jobs; correct.

### Files read in full
- `python/djust/admin_ext/progress.py` (445 lines)
- `python/djust/admin_ext/options.py` (275 lines)
- `python/djust/admin_ext/views.py` (746 lines)
- `python/djust/admin_ext/sites.py` (339 lines)
- `python/djust/admin_ext/__init__.py` (49 lines)
- `python/djust/admin_ext/templates/djust_admin/_change_form_widgets.html` (17 lines)
- `python/djust/admin_ext/templates/djust_admin/_change_list_widgets.html` (13 lines)
- `python/djust/admin_ext/templates/djust_admin/bulk_progress.html` (77 lines)
- `python/djust/tests/test_bulk_progress.py` (462 lines)
- `python/djust/tests/test_admin_widgets_per_page.py` (410 lines)
- `python/djust/checks.py` lines 2540-2666 (new A072/A073 block)
- `python/tests/test_checks.py` lines 3789-3859 (new A072/A073 tests)
- `docs/website/guides/admin-widgets.md` (346 lines)
- `examples/demo_project/demo_project/djust_admin.py` (98 lines)
- `examples/demo_project/demo_project/urls.py` (50 lines)
- `CHANGELOG.md` lines 1-80
- `ROADMAP.md` diff
- `.pipeline-state/feat-v070-admin-widgets-plan.md`
- `.pipeline-state/feat-v070-admin-widgets-fix-pass-notes.md`

### Edge cases checked
- Create-view path: `self.object is None` → `object_pk = None` → `_serialize_widget_slots` drops `object_id` key → template `{% if w.object_id %}` guard → fallthrough to `{% live_render w.view_path %}`. Correct.
- Non-owner access: `BulkActionProgressWidget.mount` checks `job.user_id != request.user.pk` AFTER `is_staff` check → `PermissionDenied`. Correct.
- Non-staff owner: blocked by `is_staff` check first. Correct.
- Job not found / expired: `self.error = "Job not found or expired."`, `self.done = True`, no `start_async`. Correct.
- Empty message in `Job.update`: falsy, skipped. Correct.
- `permission_required` as tuple vs string: both handled (`options.py:82`).
- `request.user` None: `_widget_has_permission` fail-closed.
- `int("abc")` coercion: `TypeError/ValueError → asgi_workers = 1`. Correct.
- Eviction of current job by another thread while `_poll` running: `_JOBS.get` returns `None` → poll exits cleanly.
- Decorator URL reverse failure: both `djust_progress` and `changelist` reversal have try/except with fallback string. Defensive.
- Action body that never calls `progress.update`: job stays at `current=0, total=0`, `done=True` in `finally`, `percent=0`. Renders OK.
- Message contains HTML: Django auto-escape in template. Safe.
- Very long `description=` (user-controlled): NOT truncated (🟡 #4).

johnrtipton:
Addressed Stage 11 findings (commit 41d32bd4):

**🔴 #1 (redirect drop on WS action)** — `ModelListView.run_action` now intercepts `HttpResponseRedirect` returned by the action (stock Django admin convention; also what `@admin_action_with_progress` returns) and dispatches `self.push_event("redirect", {"url": result.url})` — mirrors the `LoginView.do_login` pattern at `python/djust/admin_ext/views.py:711`. The decorator's `HttpResponseRedirect` return type is preserved for backward-compat with unmodified stock admin actions. End-to-end test (`test_admin_action_with_progress_triggers_client_redirect`) dispatches `run_action` and asserts a `redirect` push_event targeting `djust-progress/...`.

**🟡 #2 (LRU lock)** — `_store_job` body now runs under a module-level `threading.Lock` (`_JOBS_LOCK`). Test `test_store_job_concurrent_inserts_stay_under_cap` spawns 50 threads × 20 inserts (1000 total) and asserts `len(_JOBS) == _MAX_JOBS` exactly — no over-eviction.

**🟡 #3 (DJUST_ASGI_WORKERS coercion)** — Implementation at `python/djust/checks.py:2590-2594` already defensively coerces via `int(asgi_workers)` with `(TypeError, ValueError)` fallback. Test `test_A073_handles_string_setting_value` locks the behavior: asserts A073 fires for `"2"`, stays silent for `"1"` and `"not-a-number"`.

**🟡 #4 (action_label cap)** — `action_label` now clamped via `_truncate(raw_label)` at `Job` construction in `admin_action_with_progress`. Test `test_action_label_truncated_at_max_chars` confirms a 10k-char `description` produces `job.action_label` of exactly `_MAX_MESSAGE_CHARS` ending with `...`.

**Tests:** 4 new tests + 28 existing = 32 passing in target suites. Regression: `test_admin_basic` + `test_admin_plugins` (40 tests) all green. Pre-push hook full test suite passed.

johnrtipton:
## Pipeline Retrospective — PR #989

### What shipped

Django admin widgets + bulk-action progress for v0.7.0: per-page widget slots on `DjustModelAdmin` (`change_form_widgets`, `change_list_widgets`), `BulkActionProgressWidget(LiveView)`, and an `@admin_action_with_progress` decorator with server-side `allowed_permissions` enforcement, cooperative cancellation, LRU-capped `_JOBS` dict, and generic user-facing error messages. Two new system checks (A072 widget-registry sanity, A073 multi-worker warning) ship with it. Third task of the v0.7.0 milestone.

### Quality rating

**3 / 5** — Substantial save by Stage 11 (caught a 🔴 that would have shipped a broken headline feature end-to-end) is exactly why the pipeline exists. But the 🔴 slipped past Stage 7 — the third consecutive pipeline where Stage 7 rubber-stamped a diff that Stage 11 proved was broken. That's a pattern, not an accident. Quality is pulled down because two independent stage-level issues surface at once: (a) the original Stage 5 implementation wired a Django `HttpResponseRedirect` through an `@event_handler` where it cannot work, and (b) Stage 7 didn't trace the user-visible flow end-to-end. Pulled back up because: planner re-scoped this feature away from a "60% duplicate-code" trap, the combined fix-pass resolved 7 issues in one commit, Action #124 TDD discipline surfaced in 5 of the rule tests, and the single CodeQL retry was a clean one-line fix. Net: 3/5 — the pipeline worked, but only because Stage 11 did the job Stage 7 should have shared.

### Review stats

| Metric | Count |
|---|---|
| Python tests added | 32 |
| 🔴 findings (Stage 7+8+11) | 1 (Stage 11) |
| 🟡 findings | 7 (4 Self-Review, 3 Code Review) + 5 Security LOW/INFO |
| Findings addressed before merge | 8 of 8 blockers (🔴 + all 🟡) |
| Pre-commit attempts | 1 clean + 1 CodeQL retry |
| CI retries | 1 (CodeQL py/undefined-export) |
| Stage loops | 0 |

### What worked

- **Stage 11 catching the 🔴 redirect-drop bug.** `@admin_action_with_progress` returned `HttpResponseRedirect` from a function wired through `@event_handler`. Event-handler return values aren't special-cased for HTTP responses — the redirect was silently dropped on the server; the browser stayed on the changelist; the progress page never mounted; Cancel was unreachable. The PR's headline user story ("fire action → confirm redirect to progress page") could not succeed as shipped. No test asserted the end-to-end navigation signal — only the decorator's direct return. Stage 12 fixed via `push_event("redirect", ...)` intercept in `run_action` plus a regression test that asserts the navigation signal. Without this review, the feature would have shipped cosmetically but done nothing in production. Single biggest save of the session.
- **Planner re-scoped the feature and avoided ~60% duplicate code.** The original briefing said "ship a `DjustAdminMixin` for stock `ModelAdmin` integration." Planner read the codebase and discovered `admin_ext/` was already a full alternative admin (`DjustModelAdmin`, `ModelListView`, templates, URL patterns). Implementing the briefing would have built a second admin path that duplicated 60% of what `admin_ext/` already did. Planner re-scoped to "extend existing `DjustModelAdmin` with widget slots" — ~40 LOC of additive surface instead of a parallel track. This is the seventh consecutive milestone where planner-first design intercepted a structural wrong-turn.
- **Action #124 TDD discipline applied to five rule tests.** Log-level (`logger.exception` + ERROR level + exc_info), A073 firing direction (fires on N>1, silent on N=1), LRU cap exact behavior, generic-error vs raw-error split (`_error_raw` for logs, `error` for users), cooperative cancel (threading-event-proven early exit). All five rule tests were written as doc claims in executable form. This is the first PR where #124 was applied.
- **Combined fix-pass resolved 7 issues in a single commit.** Stage 7 returned 4 🟡, Stage 8 returned 2 MEDIUM + 3 LOW, and the original implementation had a handful of overlapping concerns. The fix-pass shipped as one commit with 4 new regression tests. No multi-iteration rework. Pattern is now the established norm (fourth consecutive milestone using it).
- **CodeQL `py/undefined-export` caught a subtle dynamic-`__all__` pattern that ruff missed.** Lazy-loaded `BulkActionProgressView` was in `__all__` behind a `noqa: F822` suppression; ruff's own lint was silenced, but CodeQL's independent analyzer flagged it. Clean fix: drop from `__all__` (import still works via `__getattr__`). Commit `9493c5bc`. This is the third occurrence of CodeQL-catching-what-ruff-missed across v0.6.x/v0.7.0 — now clearly a pattern.

### What could be improved

- **Stage 7 missed a 🔴 that would have shipped a broken feature end-to-end. Third consecutive pipeline.** PR #986 (time-travel) — Stage 11 caught `JsonResponse` outside the try/except. PR #988 (dj-activity) — Stage 11 caught fire-and-forget flush that broke the same-round-trip contract. PR #989 (admin widgets) — Stage 11 caught the redirect-drop that broke the feature end-to-end. Common shape: Stage 7 reads the diff but does not trace the full user flow from HTTP request → server dispatch → response → browser render. The 🔴s all come from a seam Stage 7 is not currently required to cross. Proposed upgrade: Stage 7 must explicitly trace the happy-path user story end-to-end for every user-visible feature. Not just "the diff looks right." Filing as Action #125.
- **Action #124 was applied but incompletely.** Five rule tests were written as doc claims in executable form — but the "redirect to progress page after action fires" claim in the PR body was never expressed as a test. That's exactly the claim Stage 11 later proved false. The discipline is correct; the coverage is incomplete. #124's enforcement needs to reach "for every user-visible claim in the PR body, point to the asserting test" — not just rule-level claims in the docs guide. The Stage 7 user-flow-trace upgrade (#125) subsumes this for user-visible features.
- **Doc-vs-code drift continued showing up at Stage 7.** Log-level doc said WARNING but code used `exception` (ERROR); A073 doc claimed multi-worker warning but code didn't distinguish worker count semantics cleanly; permission filter rendered HTML test was absent; create-view empty `object_id` template guard was undocumented. All four were 🟡, all four would have been caught by Action #124's "for each documented rule, point to the asserting test" check — but Stage 5 author hadn't yet internalized the discipline.
- **`noqa: F822` + CodeQL `py/undefined-export` is a recurring false-negative for ruff.** Ruff's local suppression pattern hides what CodeQL then catches — with a full CI cycle cost. Proposal: pipeline checklist grows either a "run CodeQL locally before push" step or at least a "grep for `noqa: F822` in `__all__` patterns" pre-push check. Low priority (CI catches it), but the CI-retry cost adds up across a milestone.

- **The 3-for-3 Stage-7-miss-Stage-11-catch pattern is now a milestone-level anti-pattern.** PR #986: JsonResponse outside try/except (response-layer bug). PR #988: fire-and-forget flush breaking same-round-trip (transport-layer bug). PR #989: HttpResponseRedirect dropped by event-handler (dispatch-layer bug). All three are "the code does a thing, but the thing doesn't reach the user." Stage 7's current checklist reads the diff; Stage 11's checklist traces the flow. Stage 7 needs the flow-trace row. Without it, Stage 11 is doing double duty on bug classes Stage 7 should share.
- **Planner-first is now the highest-leverage stage for this milestone.** PR #986 (planner corrected 4 briefing assumptions), PR #988 (planner corrected 4 briefing assumptions again), PR #989 (planner re-scoped away from 60% duplicate-code trap). Across three pipelines, planner has saved ~12 corrections and one structural wrong-turn. The pattern is: the user briefing reflects what the user wants, not what the codebase already has; the planner reconciles them before the implementer ships. This is now the most consistently-valuable stage in the pipeline.
- **CodeQL-catches-what-ruff-misses is a clear, recurring pattern.** Third occurrence in v0.6.x/v0.7.0 (PR #975 script-regex, PR #970 script-regex, PR #989 undefined-export). The common shape: ruff's local lint allows suppression (`noqa`, `ignore`), CodeQL runs its own analysis on top, and differences surface. Local CodeQL runs would catch these pre-push; CI CodeQL runs catch them post-push with retry cost. Worth evaluating a pre-push hook.
- **"Doc-claim-verbatim tests" (Action #124) only catches what's claimed. The bigger risk is what's un-claimed but shipped.** The redirect-drop bug wasn't in any docstring — it was in the PR body (as the headline user story) but never in a guide or code comment. #124 enforces test coverage for written claims; it doesn't enforce test coverage for "what will the user see?" That's the gap #125 fills.

### Action items

- [ ] **NEW: Action Tracker #125 — Upgrade Stage 7 checklist with user-flow trace.** For every user-visible feature, Stage 7 must explicitly trace the happy-path user story end-to-end (HTTP request → server dispatch → response envelope → browser render/navigation). Not just "read the diff." Three of the last three 🔴 findings (PRs #986, #988, #989) would have been caught by this row. Enforcement: Stage 7 output template grows a "User flow trace" section with a required bullet per user-visible feature.
- [ ] Tech-debt: evaluate local CodeQL pre-push hook or at least a grep for `noqa: F822` in `__all__` patterns. Third occurrence of CodeQL catching what ruff missed. Low priority — CI catches it, but the retry cost compounds across a milestone.
- [ ] Update Action #124 status: applied in this PR with partial success. Five rule tests written RED first (log level, A073, LRU cap, generic error, cooperative cancel). But the 🔴 redirect-drop bug came from a PR-body claim never expressed as a test. #124's enforcement scope should extend from "documented rules in the guide" to "user-visible claims in the PR body." Subsumed by #125 for the user-facing subset.

### Kudos

- **Stage 11 for catching the redirect-drop 🔴.** Without this review, the feature would have shipped cosmetically broken — action fires, job runs in background, but the user never sees the progress page because the browser never navigates. Silent failure of the headline user story. Single biggest save of the session.
- **Stage 8 for finding the unbounded `_JOBS` dict + raw-exception leakage to the client before users hit them.** Both are the kind of bug that only bites under real load or real adversarial input; Stage 8's static scan caught them both in minutes.
- **Planner for re-scoping the feature away from the "60% duplicate-code" trap.** Reading the briefing literally would have built a second admin path that duplicated what `admin_ext/` already did. Planner's 30-second codebase read saved a milestone's worth of future consolidation work.

Issue #992, opened 2026-04-24
tech-debt: SSE client hardcodes /djust/sse/ prefix — same FORCE_SCRIPT_NAME bug (03b-sse.js:44)


johnrtipton:
Same class of bug as #987 but in the SSE client fallback transport.

Location: `python/djust/static/djust/src/03b-sse.js:44` — hardcoded `/djust/sse/` URL prefix that breaks deployments mounted under `FORCE_SCRIPT_NAME` or a custom `include()` prefix.

**Fix pattern** (mirrors PR for #987 v0.7.1):
1. Server-side: extend `{% djust_client_config %}` to also emit `<meta name="djust-sse-prefix" content="...">`, derived via Django's `reverse()` against the SSE URL name.
2. Client-side: in `00-namespace.js`, add `window.djust.ssePrefix` + `window.djust.sseUrl(path)` helpers (same pattern as `apiPrefix` / `apiUrl`).
3. Update `03b-sse.js:44` to use `window.djust.sseUrl(...)`.

**Target:** v0.7.2.

**Label:** tech-debt.

PR #990, opened 2026-04-24
feat(v0.7.0): streaming markdown renderer (pulldown-cmark + {% djust_markdown %})

johnrtipton:
## Summary

Fourth v0.7.0 feature — Rust-side CommonMark parser for LLM streaming output. The headline AI-vertical positioning feature. Safe, fast, incremental rendering of markdown tokens as they arrive.

## Architecture

- **Rust impl** (`crates/djust_templates/src/markdown.rs`, 330 LOC): `pulldown-cmark` 0.12 parser; `RenderOpts` dataclass; `split_provisional` mid-stream safety (unclosed `**`, `*`, `` ` ``, `[`, `](` in last line → plain-text trailing span); `sanitise_event` re-routes raw HTML events to escaped Text; `neutralise_url` rewrites `javascript:`/`vbscript:`/`data:` → `#` (case-insensitive, whitespace-trimmed); 10 MiB per-call hard cap.
- **PyO3 binding** (`crates/djust_live/src/lib.rs`): `py.allow_threads` for GIL-released parse.
- **Python helper** (`python/djust/markdown.py`): `mark_safe` wrapper.
- **Template tag** (`python/djust/template_tags/markdown.py`): `{% djust_markdown var [kwargs] %}` with fail-closed Django `escape()` fallback on ImportError.
- **System check** (A090): info-level confirmation of safe-mode active when tag is in use.

## Scope

- `djust.render_markdown(src, *, provisional=True, tables=True, strikethrough=True, task_lists=False)` Python API
- `{% djust_markdown var [provisional=] [tables=] [strikethrough=] [task_lists=] %}` template tag
- 38 tests (24 Rust + 8 Python + 6 tag + A090 ×3)

## Deviation from plan

- **`autolinks` kwarg DROPPED** — pulldown-cmark 0.12 has no `ENABLE_GFM_AUTOLINK` / `ENABLE_AUTOLINK` flag. Documented in guide Limitations + CHANGELOG.
- **`sanitise_event` custom filter** — plan assumed omitting `Options::ENABLE_HTML` suppresses raw HTML. It does NOT in 0.12: parser still emits `Event::Html` / `Event::InlineHtml`. Required a custom event filter mapping these to `Event::Text` (escaped on write).

## Action Tracker discipline applied

- **Action #124** (rule tests RED before impl): 6 ⭐ Rust rule tests written as stubs first, confirmed failing with stub `render_markdown` returning empty string, then implemented to green.
[...]

PR #993, opened 2026-04-24
feat(v0.7.1): FORCE_SCRIPT_NAME support for JS clients (closes #987)

### LINKED ISSUE #987: tech-debt: JS clients hardcode /djust/api/ prefix — breaks under FORCE_SCRIPT_NAME
The @event_handler(expose_api=True) client and @server_function client both hardcode /djust/api/ URL prefixes. Deployments using Django FORCE_SCRIPT_NAME or sub-path mounting have no way to override. Fix: serve the prefix via a meta tag or a bootstrap config object, and have both clients resolve via that. Noted in PR #986 Stage 11 review. Target: v0.7.1.

johnrtipton:
## Summary

First v0.7.1 work — unblocks djust deployments behind `FORCE_SCRIPT_NAME`, reverse-proxy sub-paths, or custom `api_patterns(prefix=...)` mounts. Previously the ADR-008 `djust.call()` RPC client hardcoded `/djust/api/` and 404'd on any non-default mount.

## Design

**Server** — new `{% djust_client_config %}` template tag (dual-registered for Django engine + Rust engine, mirroring the `djust_markdown` pattern from PR #990). Emits:
```html
<meta name="djust-api-prefix" content="{server-resolved-prefix}">
```
Prefix resolved via Django's `reverse()` — probes both `djust_api:djust-api-call` (namespaced) and `djust-api-call` (bare), falls back cleanly when the API app isn't mounted.

**Client** — `window.djust.apiPrefix` reads meta on boot. Priority: explicit global > meta tag > default `/djust/api/`. New `djust.apiUrl(path)` helper normalizes leading-/trailing-slash concatenation.

**48-server-functions.js** — routes POST target through `djust.apiUrl()`.

## Scope

- `python/djust/templatetags/live_tags.py` (+111) — Django-side tag
- `python/djust/template_tags/client_config.py` (NEW, +67) — Rust-engine `TagHandler` (added in Stage 5b after pre-push found LiveView templates use the Rust engine)
- `python/djust/static/djust/src/00-namespace.js` (+45) — `apiPrefix` init + `apiUrl` helper
- `python/djust/static/djust/src/48-server-functions.js` — use `djust.apiUrl()`
- `examples/demo_project/templates/base.html` — WIRING_CHECK
- 12 new tests (5 Py + 6 JS + 1 regression) + 3 URL-fixture files
- Docs: Sub-path deploys section in server-functions.md + http-api.md
- CHANGELOG entry; ROADMAP v0.7.1 section added with this row struck; RETRO Action #123 closed

## Security

- Meta-tag content passes through Django \`escape()\` BEFORE \`mark_safe\`/\`format_html\` — XSS regression test locks this with a \`\"><script>\` payload in a FORCE_SCRIPT_NAME setting
- \`escape()\` BEFORE wrapping (critical order)
- Cross-origin prefix hijacking requires pre-existing XSS; mitigated by SOP + \`credentials: 'same-origin'\`
- Stage 8: zero CRITICAL/HIGH/MEDIUM/LOW

## Stage 5b lesson

Pre-push pytest caught a real bug: the original commit registered the tag ONLY with Django's template library, but LiveView-rendered \`base.html\` goes through the Rust engine. Added \`TagHandler\` in \`python/djust/template_tags/client_config.py\` that calls the same \`_resolve_api_prefix()\` helper — both engines emit byte-identical output.

## Action Tracker

- **#124** — rule tests written RED before impl (5 ⭐ cases)
- **#125** — every doc claim in Sub-path deploys maps to a test
- **#123** — closed by this PR
- Follow-up: **#992** filed (SSE client at \`03b-sse.js:44\` has the same class of bug; v0.7.2 target)

## Test plan

- [ ] Merge + \`make test\` — baseline green
- [ ] Start demo: \`cd examples/demo_project && make start\` — response HTML has \`<meta name=\"djust-api-prefix\" content=\"/djust/api/\">\`
- [ ] Restart with \`FORCE_SCRIPT_NAME=/mysite\` — content becomes \`/mysite/djust/api/\`
- [ ] \`djust.call('v', 'f', {})\` with sub-path active → POST to \`/mysite/djust/api/call/v/f/\`

Closes #987.


johnrtipton:
## Stage 11 Code Review — PR #993

### Verdict: APPROVE

Small PR (~800 LOC), tight scope, strong tests. All 5 Python + 6 JS + 1 regression tests pass locally. Byte-identical dual-engine output confirmed empirically. No MUST FIX items.

---

### 🔴 MUST FIX (none)

None.

### 🟡 SHOULD FIX (1)

**SF1 — No cross-engine parity test** (`python/djust/template_tags/client_config.py` vs. `python/djust/templatetags/live_tags.py:106-150`)

The PR narrative and both doc guides claim the Rust engine and the Django engine emit "byte-identical output." I verified this empirically (see "What I verified" below), but **no automated test locks that invariant**. The Django-side tag uses `mark_safe('<meta ...content="' + escape(prefix) + '">')`; the Rust-side handler uses `format_html('<meta ...content="{}">', prefix)`. These happen to produce the same string today, but a future change to either branch (e.g. someone switches the Django side to `format_html` and drops the `>`, or someone swaps `escape()` for a different escaper with different entity mappings for `/` or `=`) will silently break parity without a test failing.

Suggested fix: add one test that imports both the Django tag and the Rust-handler class, calls each with the same `_resolve_api_prefix()` result under an `@override_settings` matrix, and asserts `django_output == rust_handler.render([], {})` byte-for-byte. Cheap (~10 LOC) and locks the Stage 5b lesson into CI.

### 🟢 Observations / Nits (3)

- **O1 — `apiUrl` absolute-URL pass-through is benign but under-documented.** `apiUrl('https://evil.com/x')` returns `'/djust/api/https://evil.com/x'`, which `fetch()` treats as same-origin (the absolute form is "pasted under" the prefix rather than used as an origin override). `apiUrl('//evil.com/x')` yields `'/djust/api//evil.com/x'` — the `//` in the middle of a same-origin path is harmless. Only controlled callers hit this today (`djust.call` wraps args in `encodeURIComponent`), but if `apiUrl` becomes public API in v0.7.2+, a one-line docstring note (“treat input as a path segment, not a URL; absolute inputs are intentionally pasted under the prefix”) would pre-empt confusion.

- **O2 — Explicit `apiPrefix = ''` has surprising semantics.** `00-namespace.js:26` checks `!== 'undefined' && !== null`, so setting `window.djust.apiPrefix = ''` BEFORE client.js loads skips the meta-tag probe. Then at call time, `raw = window.djust.apiPrefix || '/djust/api/'` at `00-namespace.js:48` falls back to the default. Net effect: caller who thinks they're overriding to empty gets the default instead. Almost certainly never hit in practice, but the rule docs say "explicit > meta > default" and this is neither — the empty string is silently ignored at call time. Nit only.

- **O3 — Docstring parameter shadowing.** `python/djust/template_tags/client_config.py:48` declares `render(self, args: List[str], context: Dict[str, Any])` and `# noqa: ARG002` on the `args` param — but the body doesn't reference either arg. Minor: swap `ARG002` (unused arg) to cover both, or change the signature annotation.

---

### What I verified

- **Dual-registration parity (byte-identical output check):** ran both branches side-by-side under an ad-hoc script with 5 input shapes including the XSS payload `/my"site<script>`, an empty string, and whitespace-containing prefixes. All 5 produced **byte-identical** strings between `mark_safe(escape(p))` (Django) and `format_html()` (Rust). Core helper (`_resolve_api_prefix`) is shared, so only the HTML-wrapping step is a divergence risk — and today it isn't diverged.
- **Priority order (explicit > meta > default):** verified the three documented paths in `00-namespace.js:22-41`. Edge cases: `null` and `undefined` correctly fall through; empty string `''` unexpectedly short-circuits init but then falls back to default at `apiUrl` use (nit O2).
- **apiUrl() safety for absolute URL inputs:** `https://evil.com/x` → `/djust/api/https://evil.com/x` (stays same-origin for fetch); `//evil.com/x` → `/djust/api//evil.com/x` (same). No cross-origin escape hatch. `../etc/passwd` → `/djust/api/../etc/passwd` — browsers normalize to `/etc/passwd`, a same-origin 404, not a security issue for this client (nit O1 is about public API surface clarity).
- **Stage 5b retro lesson value:** pre-push caught a real production bug — the original commit ONLY registered the tag with Django's template library, so LiveView templates rendered by the Rust engine would 500. The fix (dual-registration mirroring `djust_markdown`) is correct and idiomatic. **This is exactly the scenario CLAUDE.md warns about** (`Common Pitfalls` → tests must match actual module paths). Kudos for adding both paths. SF1 above proposes locking this in with a parity test.
- **Action #125 claim-to-test mapping (5/5 claims in server-functions.md lines 315-333):** (1) default prefix → `test_tag_emits_meta_with_default_prefix` ✓ (2) FORCE_SCRIPT_NAME → `test_tag_emits_meta_under_force_script_name` ✓ (3) custom `api_patterns(prefix=)` → `test_tag_emits_meta_when_api_mounted_at_custom_prefix` ✓ (4) client fallback → `test_default_prefix_when_no_meta` (JS) ✓ (5) explicit-global precedence → `test_explicit_global_override_wins` (JS) ✓. All five doc claims have asserting tests cited by name. Excellent discipline.
- **Grep for other `/djust/api/` hits in source:** 3 hits in `00-namespace.js` (documented defaults — correct). 0 hits in other `static/djust/src/*.js` modules. 4 hits in `python/djust/api/` (server-side, unrelated — `reverse()` path; not affected). Stage 1 claim ("only one source hit") is accurate for the client bundle. SSE follow-up (#992) correctly filed for `/djust/sse/` which still has the class-of-bug.
- **WIRING_CHECK (`examples/demo_project/templates/base.html:10`):** `{% djust_client_config %}` is the FIRST `<meta>`-emitting tag in `<head>` (after charset + viewport). `client.js` is injected by `post_processing.py:296` with `defer` attribute — parsed AFTER HTML body per HTML5 spec, so no race with the meta tag. Any inline script in `<head>` BEFORE `client.js` would also see the meta tag because `<meta>` precedes it in source order. Ordering is guaranteed.
- **Follow-up #992 filed correctly:** labeled `tech-debt`, target v0.7.2, cites `03b-sse.js:44`, fix pattern (mirror this PR) documented. Good.
- **#987 closure:** issue body says "The @event_handler(expose_api=True) client and @server_function client both hardcode /djust/api/ URL prefixes." This PR fixes the `@server_function` client (via `48-server-functions.js` → `apiUrl`). The `@event_handler(expose_api=True)` client surface goes through the same `djust.call` → `apiUrl` path (the handler's api_path is used server-side for URL mounting, client-side it's the same fetch). Closure is valid.

### Files read in full

- `/Users/tip/Dropbox/online_projects/ai/djust_project/djust/python/djust/templatetags/live_tags.py` (lines 40-150 of the #987 addition + context)
- `/Users/tip/Dropbox/online_projects/ai/djust_project/djust/python/djust/template_tags/client_config.py`
- `/Users/tip/Dropbox/online_projects/ai/djust_project/djust/python/djust/template_tags/__init__.py` (registry + auto-registration)
- `/Users/tip/Dropbox/online_projects/ai/djust_project/djust/python/djust/template_tags/markdown.py` (precedent for dual-registration)
- `/Users/tip/Dropbox/online_projects/ai/djust_project/djust/python/djust/static/djust/src/00-namespace.js`
- `/Users/tip/Dropbox/online_projects/ai/djust_project/djust/python/djust/static/djust/src/48-server-functions.js`
- `/Users/tip/Dropbox/online_projects/ai/djust_project/djust/python/djust/tests/test_client_config_tag.py`
- `/Users/tip/Dropbox/online_projects/ai/djust_project/djust/tests/js/api_prefix.test.js`
- `/Users/tip/Dropbox/online_projects/ai/djust_project/djust/tests/js/server_functions.test.js`
- `/Users/tip/Dropbox/online_projects/ai/djust_project/djust/tests/api_test_urls_default.py`
- `/Users/tip/Dropbox/online_projects/ai/djust_project/djust/tests/api_test_urls_custom.py`
- `/Users/tip/Dropbox/online_projects/ai/djust_project/djust/tests/api_test_urls_unmounted.py`
- `/Users/tip/Dropbox/online_projects/ai/djust_project/djust/examples/demo_project/templates/base.html`
- `/Users/tip/Dropbox/online_projects/ai/djust_project/djust/docs/website/guides/server-functions.md` (Sub-path deploys section)
- `/Users/tip/Dropbox/online_projects/ai/djust_project/djust/docs/website/guides/http-api.md` (Sub-path deploys section)
- `/Users/tip/Dropbox/online_projects/ai/djust_project/djust/CHANGELOG.md` (Unreleased entry)

### Edge cases checked

- `window.djust.apiPrefix = null` before init — falls through to meta/default ✓
- `window.djust.apiPrefix = undefined` — falls through ✓
- `window.djust.apiPrefix = ''` — short-circuits but falls back at `apiUrl` call time (O2 nit)
- `window.djust.apiPrefix = 0` — short-circuits and falls back to default at call time (same as O2)
- Missing meta tag — falls back to `/djust/api/` ✓
- Meta tag with empty `content=""` — `raw.trim()` is empty, falls through ✓
- `apiUrl('https://evil.com/')` — pasted under prefix as same-origin path (safe for fetch)
- `apiUrl('//evil.com/')` — same, same-origin double-slash path (safe)
- `apiUrl(null)` → `/djust/api/` ✓
- `apiUrl('')` → `/djust/api/` ✓
- Prefix without trailing slash (`content="/api"`) — auto-normalized ✓
- Path with leading slash (`apiUrl('/x')`) — leading slash stripped ✓
- FORCE_SCRIPT_NAME with HTML special chars (`/my"site<script>`) — escaped correctly ✓
- `{% djust_client_config %}` with unmounted API — emits `content=""` for debuggability, client falls back ✓
- Script-tag ordering in `<head>` — `<meta>` precedes `client.js` (which is `defer`-loaded); no race

### Summary

- 🔴 0 MUST FIX
- 🟡 1 SHOULD FIX (parity test)
- 🟢 3 observations/nits

Solid PR. The Stage 5b catch (pre-push found Rust-engine tag missing) was the kind of fail-forward event that justifies dual-registration discipline. Adding a parity-locking test would make the invariant explicit rather than implicit. Everything else is minor.

johnrtipton:
Addressed Stage 11 🟡:

Parity test added — `test_django_and_rust_engines_emit_identical_output` parameterized over 3 URL-config scenarios (default, FORCE_SCRIPT_NAME, custom prefix). Asserts byte-equality of Django-engine render vs Rust-engine render output. Locks the Stage 5b dual-registration invariant.

🟢 nits also addressed inline: JSDoc for `apiPrefix = ''` backward-compat and `apiUrl()` absolute-URL behavior; inline rationale for the unused `args`/`context` params in `ClientConfigTagHandler.render()`.

Commit: 090c290c

johnrtipton:
## Pipeline Retrospective — PR #993

**Merged**: `f03d64eb` · **Milestone**: v0.7.1 (first PR) · **Closes**: #987 · **Follow-up**: #992

### What shipped

`{% djust_client_config %}` template tag (dual-registered for Django + Rust engines per the `djust_markdown` / PR #990 precedent) emits a `<meta name="djust-api-prefix" content="…">` tag. Client-side `window.djust.apiPrefix` + `djust.apiUrl(path)` helpers added in `00-namespace.js`, replacing the hardcoded `/djust/api/` string in `48-server-functions.js`. `_resolve_api_prefix()` reads `FORCE_SCRIPT_NAME` + the `djust.api_patterns(prefix=...)` mount prefix to compute a single source of truth. Priority order: explicit-global > meta tag > default. Follow-up #992 filed for `03b-sse.js:44` (same class of bug in SSE client, v0.7.2 target).

### Quality rating

**5 / 5** — Cleanest pipeline in the session. Stage 11 APPROVE with 0 🔴 for the **SECOND CONSECUTIVE** pipeline (first was PR #990 where Action #125 was introduced). The one potential-🔴 bug (Rust engine didn't know the tag → 500 on LiveView renders) was caught by the pre-push gate at Stage 5b **before** it shipped — the safety net worked exactly as designed. Stage 7 REVIEW_PASSED (0 🔴/🟡). Stage 8 SECURITY_PASSED (0 CRITICAL/HIGH/MEDIUM/LOW). Stage 11 APPROVE (0 🔴 + 1 🟡 + 3 🟢). CI green on first push after Stage 5b.

### Review stats

| Metric | Count |
|---|---|
| Python tests added (Stage 5) | 5 |
| JS tests added (Stage 5) | 6 |
| Regression tests (Stage 5) | 1 |
| Dual-engine parity tests (Stage 12 fix-pass) | 3 |
| **Total new tests** | **15** |
| 🟡 findings | 0 Self-Review + 0 Security + 1 Code Review |
| 🟢 findings | 1 Self-Review + 0 Security + 3 Code Review |
| Pipeline stages | 14 + 1 Stage 5b = 15 |
| CI retries | 0 |

### What worked

- **Pre-push gate is load-bearing — it caught the dual-engine bug that would have slipped past Stages 5/6/7/8.** Stages 5–8 run Python + JS + Rust tests independently; the Python-side tag handler tests passed in isolation, and the demo views that trigger the Rust engine weren't in the targeted test set. The FULL pre-push pytest ran the demo tests and caught the 500. Without this gate, the bug ships. Textbook layered-defenses outcome.
- **Action #125 paid off for the 2nd consecutive pipeline.** Stage 7 flagged only a doc-citation nit; Stage 11 caught the missing dual-engine parity test and Stage 12 added it in 15 min (3 parameterized cases). PRs #990 → #993 are both 0 🔴 at Stage 11. Streak breaker has become streak.
- **Planner-first saved ~45 min by bounding the scope.** Stage 1 identified that only 1 source JS file (`48-server-functions.js`) had the hardcoded prefix AND the Python side was already parameterized via `djust.api_patterns(prefix=…)`, so the plan was bounded and accurate from the start. No scope creep. ~340 LOC shipped.
- **Re-using `djust_markdown` precedent from PR #990 — Stage 5b fix didn't require design work, just copy the pattern.** Repeatable templates are the v0.7.0 milestone dividend: both `djust_markdown` (PR #990) and `djust_client_config` (PR #993) now form the two canonical examples of dual-engine tag registration.
- **Clean scope discipline on the SSE follow-up.** Rather than expanding PR scope, #992 was filed as a separate v0.7.2 issue with the fix pattern referenced.

### What could be improved

- **Stage 5 should have caught the dual-engine issue at design time — not Stage 5b.** The plan didn't mention that LiveView templates go through the Rust engine. Fix: add "engine path" to the Stage 4 planner checklist. For any new template tag, the plan must explicitly state which engine(s) render the template(s) that consume it. Filing as **Action #129**.
- **Stage 7 AND Stage 8 both rubber-stamped the "empirically verified parity" claim without a test.** Stage 11 caught it on the second pass (the 🟡 cross-engine parity test). Action #125 discipline worked at Stage 11 but should have been applied at Stage 7 too. Propose: Stage 7 checklist should include "for every 'works in both X and Y' claim, find the parity test."
- **Stage 6 targeted-subset runs can miss cross-engine issues.** When Planner flags a PR as touching multiple engines/languages, Stage 6 agents should run `make test` (full suite) rather than targeted subsets. Noted under #129.

### Insights

- **Stage 11 🔴-count streak: 0 → 0 for 2 consecutive pipelines.** Preceded by 1 → 1 → 1 → 0 → 0 across PRs #976 / #988 / #989 / #990 / #993. Action #125 is demonstrably shifting the failure curve. Track next pipeline for a third consecutive 0 🔴.
- **Pre-push gate > Stage 6 targeted subsets.** Running the FULL suite (pytest as configured) is the only reliable way to catch cross-engine issues.
- **Dual-engine tag registration is now a template.** `djust_markdown` + `djust_client_config` are the two canonical examples.
- **Repeatable templates from v0.7.0 accelerating v0.7.1.** The first PR of v0.7.1 took less total engineering time than the last PR of v0.7.0.
- **Client JS delta stayed small (+148 B gzipped).** Bundle budget comfortable.

### Action items

- [ ] **NEW: Action Tracker #129 — Stage 4 planner checklist: engine-path declaration for template tags.** For every new template tag plan, state explicitly which template engine(s) render the template(s) that consume the tag (Django, Rust, or both) and plan registration accordingly. Second canonical dual-registration example (alongside `djust_markdown`) is `djust_client_config`.
- [ ] **Update Action #125 status: Validated (2 consecutive pipelines).** PRs #990 and #993 both Stage 11 APPROVE with 0 🔴. Track next pipeline for continued trend.
- [ ] Issue #992 tracked for v0.7.2 (SSE client `03b-sse.js:44` — same class of bug).

### Kudos

- **Pre-push gate catching the dual-engine bug — textbook example of layered defenses.** Stages 5/6/7/8 all ran targeted subsets and all passed. The full pre-push suite caught the LiveView 500.
- **Planner-first scoping shipping a bounded ~340 LOC PR in a single session.** Already shipped 4 v0.7.0 PRs + the v0.7.0rc1 release in the same working context; the v0.7.1 first PR landed clean without scope drift.
- **Stage 12 parity-test fix-pass — 3 parameterized cases, ~15 min.** Exactly the right response to Stage 11's one 🟡: small cheap test, large semantic coverage, byte-identical assertion for default / FORCE_SCRIPT_NAME / custom-prefix cases. The invariant the PR narrative claimed ("byte-identical dual-engine output") is now enforced by CI.

---

**Verdict: RETRO_COMPLETE** · Action Tracker updated (row #123 closed, row #129 added).

Issue #964, opened 2026-04-23
docs: prominent key_template convention for s3_events UUID extraction


johnrtipton:
From PR #958 retro. s3_events.parse_s3_event extracts upload_id via regex match on the first UUID-shaped path segment. Requires apps use the documented convention for key_template. If they do not, extraction fails silently and upload_id falls back to the full key. Document the UUID-prefix requirement in the upload-writers guide + on-page docstring + emit a debug warning when a key does not match the expected shape.

PR #999, opened 2026-04-24
fix(rust): honor __str__ key in Value::Object Display (closes #968)

### LINKED ISSUE #968: Rust renderer ignores __str__ key in serialized model dicts — renders "[Object]" instead
## Summary

The Rust template renderer outputs literal `[Object]` when rendering a model instance that the Python serializer has already converted to a dict, even though the dict contains a `"__str__"` key with the proper string representation.

This creates an asymmetry: Django templates call `__str__` automatically, and djust's Python serializer goes to the trouble of putting `__str__` in the dict, but the Rust renderer doesn't read it.

## Repro

`djust==0.5.4rc1`, Python 3.14.

```python
from djust._rust import render_template

# Model dicts include __str__ per djust/serialization.py _serialize_model_safely (line 157)
ctx = {"claim": {"id": 1, "__str__": "<Claim: 2026PD000075>", "__model__": "Claim", "claim_number": "2026PD000075"}}

print(render_template("{{ claim }}", ctx))
#  → "[Object]"                 ← surprising

print(render_template("{{ claim.claim_number }}", ctx))
#  → "2026PD000075"              ← works

# Compare — a plain Python object with __str__:
class Obj:
    def __str__(self): return "My Custom Str"
print(render_template("{{ x }}", {"x": Obj()}))
#  → "My Custom Str"             ← works, asymmetric with the dict case
```

## Why this matters

When a view's `get_context_data()` returns a model instance, the JIT serializer converts it to a dict before the Rust renderer sees it. At template time, `{{ claim.claimant }}` (where `claimant` is a FK) evaluates to a nested dict — which renders as `[Object]`.

This catches real Django developers off guard because:

1. In a plain Django template, `{{ claim.claimant }}` would call `Claimant.__str__`. Every Django tutorial teaches this.
2. `djust/serialization.py:157` explicitly sets `"__str__": str(obj)` on the dict — so the info is already there, just not consumed.
3. The failure is silent: the page renders and returns 200; you only notice when you visually inspect the result and see `[Object]` where a name should be.

We hit this six times in our project (NYC Claims prototype — `templates/claims/law_dept_claim.html`, `templates/intake/claim_status_detail.html`, others). Each fix is a one-line dot-notation change, but finding them required a full-template audit. The defensive workaround (adding scalar dot-access everywhere) works but is ugly — every FK display needs explicit `.name` / `.ct_desc` / `.first_name + " " + .last_name` etc.

## Suggested fix

When rendering `{{ obj }}` and `obj` is a dict that contains a `"__str__"` key (set by `_serialize_model_safely`), output that value instead of `[Object]`.

In pseudo-Rust (I don't have the template renderer source handy):

```rust
// When resolving a variable that's a dict:
match value {
    Value::Dict(d) if d.contains_key("__str__") => {
        output.push_str(d.get("__str__").unwrap().as_str());
    }
    Value::Dict(_) => {
        // still print [Object] for non-model dicts, or maybe {} / abbreviated form
        output.push_str("[Object]");
    }
    _ => // existing behavior
}
```

Benefits:

- Matches Django template semantics (`{{ obj }}` → `str(obj)`)
- Zero breakage: templates that already use dot-notation continue to work unchanged
- Uses the `"__str__"` key djust is already generating — no serialization change needed

Possible concern: if a user's model has a `__str__` that returns a long or unsafe string, the template would now emit it. But that's already the Django default behavior. The current `[Object]` is worse because it silently masks the issue rather than showing what the user's `__str__` returns.

## Alternative

If changing the default is risky, provide an opt-in template filter `{{ obj|str }}` that emits `__str__` from a dict. Less surprising for existing users but still ugly.

## Related

- `djust/serialization.py:157` — where `__str__` is set on the dict
- CLAUDE.md convention we've added in our project: https://github.com/flexion/flexion-nyc-claims/blob/main/CLAUDE.md#fk-display-in-djust-templates ("FK Display in djust Templates" section)
- Our tracker row #16 / issue https://github.com/flexion/flexion-nyc-claims/issues/66 — we're doing a full-template audit to eliminate these

Happy to open a PR against the Rust renderer if you'd like. Let me know your preferred approach.

---

From the NYC Comptroller Claims Processing prototype team, building on djust v0.5.x.

johnrtipton:
## Summary

Fix #968 — Rust renderer's `Value::Object` Display impl ignored the `"__str__"` key on serialized-model dicts, emitting the literal `"[Object]"` placeholder where Django's template engine would call `str(obj)`.

## Root cause

`djust.serialization._serialize_model_safely` sets `"__str__": str(obj)` on every dict it produces (and on nested FK dicts). The Rust `fmt::Display` impl for `Value` at `crates/djust_core/src/lib.rs:169` matched all `Value::Object` cases and wrote `"[Object]"` — the `__str__` key was present but never consumed.

Result: `{{ claim }}` or `{{ claim.claimant }}` (the latter resolves to a nested FK dict) rendered `[Object]` silently. 200 OK, no traceback, only caught by visual inspection.

## Fix

```rust
Value::Object(o) => match o.get("__str__") {
    Some(Value::String(s)) => write!(f, "{s}"),
    _ => write!(f, "[Object]"),
},
```

- **`__str__` present + String variant** → emit the string (matches Django semantics).
- **No `__str__` / null / int** → keep the `[Object]` fallback (guards against upstream serialization bugs emitting `null` or coerced types).
- **Auto-escape unaffected** — the Rust template engine's escape layer runs on the Display output as before.

## Test plan

- [x] `cargo test -p djust_core` — 14/14 (5 new + 9 existing)
- [x] `pytest python/djust/tests/test_rust_renderer_str_key.py` — 13/13
- [x] `pytest python/djust/tests/ -k "render or template or serializ"` (template-adjacent regression sweep) — 920 passed, 1 skipped, 0 failed
- [x] Manual end-to-end: `{{ claim }}` on a serialized-model dict with `__str__: "Claim 2026PD000075"` renders correctly; `{{ claim.claimant }}` on a nested FK dict with `__str__: "John Doe"` renders correctly; plain dict without `__str__` still renders `[Object]`

## Coverage

**Rust unit tests** (`crates/djust_core/src/lib.rs::tests`):
1. `test_display_object_with_str_key` — canonical case
2. `test_display_object_without_str_key` — plain dict fallback
3. `test_display_object_str_key_non_string_falls_back` — `__str__: null` / wrong variant
4. `test_display_object_empty_str_key` — `__str__: ""` renders empty (not fallback)
5. `test_display_list_unchanged` — `[List]` fallback for lists untouched

**Python integration tests** (`tests/test_rust_renderer_str_key.py`, 13 tests across 4 classes):
- `TestModelDictStrKey` — model dict, HTML auto-escape, dotted access preserved, nested FK
- `TestPlainDictFallback` — no `__str__`, empty dict, null/int `__str__`
- `TestStrKeyEdgeCases` — empty string, many siblings
- `TestBackwardsCompatibility` — plain Python object (`FromPyObject` path), `[List]` unchanged, scalar variants unchanged

CHANGELOG entry under `[Unreleased]` for v0.7.2.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,29 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- **Rust renderer honors `__str__` key on serialized model dicts
+  (v0.7.2, #968)** — `djust.serialization._serialize_model_safely`
+  sets `"__str__": str(obj)` on every dict it produces so `{{ obj }}`
+  in a Rust-engine template can match Django's default `str(obj)`
+  semantics. The Rust `Value::Object` Display impl
+  (`crates/djust_core/src/lib.rs`) previously ignored the key and
+  emitted the literal `"[Object]"` for any dict. This broke FK
+  display silently in LiveView templates — `{{ claim.claimant }}`
+  (where `claimant` serializes to a nested dict) rendered as
+  `[Object]` instead of the claimant's string representation, since
+  the page still returned 200 the only way to notice was visual
+  inspection. Reported by the NYC Comptroller Claims Processing
+  prototype team who hit six occurrences in a single project. Fix:
+  when the value is `Value::Object` and contains a
+  `"__str__": Value::String(...)` entry, render the string. Non-model
+  dicts (no `__str__`, or `__str__` not a string) keep the existing
+  `"[Object]"` fallback. Plain Python objects with custom `__str__`
+  were already correct (handled by `FromPyObject`). Covered by **5
+  Rust unit tests** in `crates/djust_core/src/lib.rs::tests` and **13
+  Python integration tests** in `tests/test_rust_renderer_str_key.py`
+  (model dict, nested FK, HTML-auto-escape, dotted-access, plain-dict
+  fallback, null/int `__str__` edge cases, empty-string `__str__`,
+  backwards-compat for plain Python objects + lists + scalars).
 - **`djust.dev_server` NameError on module load when `watchdog` is
   not installed (v0.7.2, #994)** — the `try/except ImportError` block
   at `dev_server.py:13-19` sets `WATCHDOG_AVAILABLE = False` but the
--- a/crates/djust_core/src/lib.rs
+++ b/crates/djust_core/src/lib.rs
@@ -157,6 +157,21 @@ impl Value {
 }
 
 // Implement Display trait instead of inherent to_string method
+//
+// For serialized Django-model dicts the Python-side serializer sets
+// `"__str__": str(obj)` on every dict it produces (see
+// `python/djust/serialization.py::_serialize_model_safely`). This
+// matches Django's default template semantics: `{{ obj }}` in a
+// Django template calls `str(obj)`, so a rendered FK like
+// `{{ claim.claimant }}` produces the claimant's `__str__`, not a
+// placeholder.
+//
+// Before the #968 fix the Rust renderer ignored the `__str__` key
+// and emitted the literal `"[Object]"` for any dict, breaking the
+// Django semantic for LiveView templates. The current
+// implementation checks for a `"__str__"` entry first and renders
+// its string value when present, falling back to `"[Object]"` only
+// for dicts that weren't produced by the model serializer.
 impl fmt::Display for Value {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         match self {
@@ -166,7 +181,10 @@ impl fmt::Display for Value {
             Value::Float(fl) => write!(f, "{fl}"),
             Value::String(s) => write!(f, "{s}"),
             Value::List(_) => write!(f, "[List]"),
-            Value::Object(_) => write!(f, "[Object]"),
+            Value::Object(o) => match o.get("__str__") {
+                Some(Value::String(s)) => write!(f, "{s}"),
+                _ => write!(f, "[Object]"),
+            },
         }
     }
 }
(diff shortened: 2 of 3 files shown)


johnrtipton:
## Per-PR Retro — #999 (v0.7.2 drain iter 2)

**Date**: 2026-04-24
**Scope**: Rust `Value::Object` Display honors `"__str__"` key; closes #968.

### What Worked Well

- **Reporter-provided repro carried through to test design.** The issue body gave a precise 5-line Python repro. Each repro case became a test: model dict → `TestModelDictStrKey`, FK dict → `test_nested_fk_dict_renders_str`, plain dict → `TestPlainDictFallback`. The test suite is a direct translation of the issue's "expected behavior" claims — doc-claim-verbatim TDD discipline (Action #124) applied naturally here because the issue was the spec.
- **Rust + Python two-layer coverage.** 5 Rust unit tests on the Display impl itself (fast, no PyO3 overhead) + 13 Python integration tests through `render_template` (full end-to-end path). Each layer catches different regressions: a future Rust-only refactor of `Value::Object` would fail the Rust tests; a future serialization change that drops `__str__` would fail the Python tests.
- **Discovered HTML-auto-escape interaction early.** First end-to-end check showed `<Claim: 2026PD000075>` rendered as `&lt;Claim: 2026PD000075&gt;`. That's *correct* — the escape layer runs on Display output — but I almost wrote a test that asserted raw `<Claim: ...>`. Caught it before committing; added `test_model_dict_str_value_is_html_escaped` specifically to lock the escape interaction.
- **Caught rust-bool-format edge at test authoring.** Initially wrote `assert render_template(..., True) == "True"` in a backwards-compat sanity test. Rust's `{}` format emits lowercase `"true"`. That's pre-existing behavior unrelated to this fix; test docstring now explicitly notes the lowercase is out of scope, test asserts the actual pre-fix value. Prevents a future "clean up bool format" PR from silently bundling.

### What Didn't Work

- **First Python end-to-end check failed on escape.** See above — I hadn't mentally accounted for the auto-escape layer when writing the Python verification assertion. Not a big deal; diagnosed in one iteration (`'&lt;Claim: 2026PD000075&gt;' == '<Claim: 2026PD000075>'` → oh right, the engine escapes). But a note for future Rust-Display-layer tests: always think about the escape layer before asserting literal strings with angle brackets.

### What to Improve

- None actionable that aren't already filed. The existing Action #124 (doc-claim-verbatim TDD) covers the pattern that worked here.

### Review Stats

| Metric | Value |
|---|---|
| Rust tests added | 5 |
| Python integration tests added | 13 |
| Production LOC delta | +18 (Display impl + 15-line explanatory comment) / -1 |
| Test LOC delta | +225 (5 Rust + 13 Python) |
| 🟡 findings | 0 |
| Pre-commit attempts | 1 |
| Pre-push attempts | 1 |
| CI retries | 0 |
| Quality rating | 5/5 |

- None new. Two cross-layer (Rust + Python) test patterns are now in the repo (this PR and the #968 fix pattern for model-dict semantics); future Rust-Display-layer changes have a template.

### Milestone context

v0.7.2 drain iter 2 of 6. Next: #964 (docs — key_template UUID convention).

Issue #963, opened 2026-04-23
tooling: weekly real-cloud CI matrix job for S3 / GCS / Azure upload writers


johnrtipton:
From PR #958 retro. All v0.5.7 upload-writer tests mock the SDKs. Happy-path end-to-end verification against real AWS / GCP / Azure is missing. Add a weekly GitHub Actions job that runs a small integration suite: upload a 1MB file, verify it's present, delete it. Credentials via GitHub encrypted secrets. ~30 LOC workflow + ~50 LOC test.

PR #1000, opened 2026-04-25
docs(s3_events): key_template UUID convention + debug-warn on fallback (closes #964)

### LINKED ISSUE #964: docs: prominent key_template convention for s3_events UUID extraction

johnrtipton:
## Summary

Fix #964 — document the `key_template` UUID-prefix convention for `s3_events` and emit a DEBUG log when apps violate it, so the silent "hook not firing" failure mode becomes diagnosable.

## Problem

`djust.contrib.uploads.s3_events.parse_s3_event` extracts `upload_id` by finding the first UUID-shaped path segment in the S3 object key. When the app's `key_template` doesn't produce such a segment (e.g. a legacy `uploads/2024/foo.bin`), `upload_id` silently falls back to the full key — and the hook registered under the UUID never fires.

This was flagged in PR #958 retro as the #1 "my hook isn't being called" report from v0.5.7+ users. The silent fallback gave no signal.

### 1. Module docstring (authoritative)

`s3_events.py::parse_s3_event` now documents the convention explicitly with two recommended key_template shapes:

- `uploads/<upload_id_uuid>/<original_filename>`
- `<tenant_id>/<upload_id_uuid>/<original_filename>` (both work — the scan is tolerant of leading non-UUID segments)

### 2. DEBUG log on fallback

When no UUID-shaped segment matches, the parser emits a DEBUG log entry on the `djust.contrib.uploads.s3_events` logger naming the offending key. Enabling DEBUG once is enough to diagnose a silent hook. No behavior change in the happy path — normal deliveries don't emit the log.

### 3. Upload guide section

`docs/website/guides/uploads.md` grows a "Key-template convention for `s3_events`" section under the writer-path chapter with the recommended template, the fallback semantics, a debugging recipe, and a pointer to custom upload-id routing (via `x-amz-meta-upload-id` / JWT / DB-lookup) as the escape hatch.

## Test plan

- [x] `pytest python/djust/tests/test_presigned_s3_820.py` — 22/22 (3 new)
- [x] Happy-path delivery: UUID-prefix key → hook fires, no DEBUG log
- [x] Bad-key delivery: no UUID segment → upload_id = full key, DEBUG log names the key
- [x] Tolerance: UUID not in first position (`tenant/<uuid>/file.bin`) still matches
- [x] Existing 19 tests still pass — no regression in signature-verify path, hook registry, SNS unwrap, or webhook error handling

## Coverage

3 new regression tests:
- `test_parse_s3_event_no_uuid_segment_emits_debug_log`
- `test_parse_s3_event_uuid_segment_no_debug_log`
- `test_parse_s3_event_uuid_segment_not_first`

CHANGELOG entry under `[Unreleased]` / Documentation subsection for v0.7.2.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,30 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Documentation
+
+- **`key_template` UUID-prefix convention for `s3_events` (v0.7.2,
+  #964)** — `djust.contrib.uploads.s3_events.parse_s3_event` extracts
+  `upload_id` by finding the first UUID-shaped path segment in the
+  S3 object key; apps whose `key_template` doesn't produce such a
+  segment silently fall back to the full key as `upload_id`, and
+  hooks registered against the UUID then don't fire. This was the
+  #1 source of "my hook isn't being called" reports from v0.5.7+
+  users. Fix: (a) the module docstring now documents the convention
+  prominently with two recommended `key_template` shapes
+  (`uploads/<uuid>/<filename>` and `<tenant>/<uuid>/<filename>`);
+  (b) a `DEBUG` log entry fires on the
+  `djust.contrib.uploads.s3_events` logger whenever fallback
+  happens, naming the offending key — so enabling `DEBUG` logging
+  once is enough to diagnose a silent hook; (c) a "Key-template
+  convention for `s3_events`" section has been added to
+  `docs/website/guides/uploads.md` with a debugging recipe and a
+  pointer to the "custom upload-id routing" escape hatch (via
+  `x-amz-meta-upload-id` / JWT / DB lookup). Covered by **3 new
+  regression tests** in `tests/test_presigned_s3_820.py` (no-UUID
+  fallback + DEBUG log, happy path emits no log, UUID segment
+  position doesn't matter).
+
 ### Fixed
 
 - **Rust renderer honors `__str__` key on serialized model dicts
--- a/docs/website/guides/uploads.md
+++ b/docs/website/guides/uploads.md
@@ -332,6 +332,37 @@ If your own `abort()` implementation raises, djust logs the traceback and swallo
 - **`entry.data` / `entry.file` are empty.** The raw bytes never sat anywhere djust could hand them to you. Use `entry.writer_result` (whatever `close()` returned) instead.
 - **No temp file cleanup required.** Because no temp file was created, `entry.cleanup()` is a no-op for writer uploads.
 
+### Key-template convention for `s3_events`
+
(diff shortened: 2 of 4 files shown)


johnrtipton:
## Per-PR Retro — #1000 (v0.7.2 drain iter 3)

**Date**: 2026-04-24
**Scope**: Document `key_template` UUID-prefix convention for `s3_events` + DEBUG-log on silent fallback; closes #964.

### What Worked Well

- **Retro-sourced issue + retro-sourced fix.** #964 was filed from PR #958 retro. The fix is the clearest possible response: document the convention, emit a diagnosable signal when it's violated. Three small deltas (docstring, log line, guide section) + 3 tests — no new abstractions, no API changes.
- **DEBUG-not-WARNING choice is load-bearing.** WARNING on a happy-path delivery is log spam; WARNING-on-fallback would wake oncall. DEBUG is the right level — silent by default, one line of code to surface when the developer needs it. Documented the `logging.getLogger(...).setLevel(DEBUG)` recipe alongside the log line so discovering the signal is a one-line change.
- **Tolerance noted explicitly.** The parser already scanned every path segment, not just the first — but neither the docstring nor the user guide mentioned it. The new docs explicitly note `uploads/<uuid>/...` and `<tenant>/<uuid>/...` both work, with a dedicated test locking the behavior. Prevents a future "tighten this to the first segment" PR from silently breaking tenant-bucketed setups.

### What Didn't Work

Nothing. The change was bounded: one function, one docstring, one log line, one guide section. No surprises.

### What to Improve

- None. This PR is a template for "documentation-as-code" drain iterations: the docstring is the spec, the DEBUG log makes the spec observable, the tests lock the spec. Future retro-sourced silent-failure follow-ups should follow the same pattern.

### Review Stats

| Metric | Value |
|---|---|
| Python tests added | 3 |
| Production LOC delta | +17 (debug-log + comment) |
| Docstring LOC delta | +35 (module docstring expansion) |
| Guide LOC delta | +33 (new section) |
| Test LOC delta | +56 |
| 🟡 findings | 0 |
| Pre-commit attempts | 1 |
| Pre-push attempts | 1 |
| CI retries | 0 |
| Quality rating | 5/5 |

None new. Pattern validated: "silent fallback → DEBUG log with diagnostic recipe + docs explaining when fallback fires" is the right template for this class of bug.

### Milestone context

v0.7.2 drain iter 3 of 6. Next: #963 weekly real-cloud CI matrix (pipeline-dev eligible — infra only).

PR #1008, opened 2026-04-25
fix(checks): C011 catches stale/placeholder output.css (closes #1003)

johnrtipton:
## Summary

Closes #1003 — `djust.C011` (missing compiled CSS) now also detects committed-but-stale placeholder `output.css` files. Before this PR, a 100-byte placeholder comment passed `os.path.exists()` and the site rendered without any Tailwind utilities, with zero hint from `manage.py check`.

## Root cause

`djust._check_missing_compiled_css` at `python/djust/checks.py:185` tested only `os.path.exists(output.css)`. The docs.djust.org team hit this at launch — fresh clone + `make dev` produced a broken page silently because their M0 scaffolding had `output.css` as a placeholder comment while Tailwind got wired up properly in the Dockerfile.

## Fix

New `_output_css_looks_built(path)` helper extends the contract from "file exists" to "file exists AND looks built":

- **Size threshold**: real Tailwind v4 output is always > 10 KB (preflight reset + utility skeleton).
- **Header marker**: minifier emits `/*! tailwindcss ... */` banner; bare-bones theme CSS uses `@layer` directives. Either marker in the first 512 bytes is sufficient.

Both checks must pass — a 50 KB hand-rolled stylesheet without Tailwind markers is correctly flagged. Returns `False` on any I/O error so missing/permission-denied files fall through to "not built" and let C011 fire.

Warning message: `"output.css not found"` → `"output.css is missing or stale"` with a hint that placeholder files are the canonical failure mode.

## Test plan

- [x] `pytest python/tests/test_checks.py::TestC011MissingCompiledCss -v` — 8/8 pass (5 new + 3 updated)
- [x] Placeholder file (the #1003 canonical case) → C011 fires
- [x] Empty 0-byte file → C011 fires
- [x] Sub-10 KB file with banner → C011 fires (size threshold)
- [x] Real Tailwind output (banner + >10 KB) → no C011
- [x] Hand-rolled `@layer` stylesheet (>10 KB) → no C011 (`@layer` marker accepted)

## Existing test fix-up

The pre-PR test `test_c011_passes_when_output_exists` wrote `"/* compiled css */"` (~18 bytes) and expected no C011 — that was the bug the issue describes. Test updated to use a realistic ~16 KB `_REAL_TAILWIND_OUTPUT` fixture (banner + plausible CSS body). Same fix-up applied to all three "passes-when-…" tests so they exercise the new content-sniff contract.

CHANGELOG entry under `[Unreleased]` for v0.7.3.

[...]

PR #1007, opened 2026-04-25
feat(forms): inline radio buttons via [data-dj-inline] CSS contract (closes #991)

johnrtipton:
## Summary

Closes #991 — opt-in inline radio button layout for `forms.RadioSelect` fields. Zero new Python; pure CSS contract via the `data-dj-inline="true"` attribute on the widget's `attrs`.

## API

```python
class FilterForm(forms.Form):
    status = forms.ChoiceField(
        widget=forms.RadioSelect(attrs={"data-dj-inline": "true"}),
        choices=STATUS_CHOICES,
    )
```

In the base template:

```html
{% load static %}
<link rel="stylesheet" href="{% static 'djust/djust-forms.css' %}">
```

## How it works

Django's widget machinery puts `attrs={...}` onto each rendered `<input>`, **not** onto the wrapping element. So `data-dj-inline="true"` lands on every `<input type="radio">`. The bundled CSS uses the `:has()` parent selector to walk up from each marked input and lay out its containing wrapper (`<ul>` for stock Django, `<div>` for djust-theming) as `inline-flex`.

Browser support: `:has()` is stable in Chromium 105+, Safari 15.4+, Firefox 121+ — all shipped since late 2023.

## Why this design (vs. form-level flag / template tag)

- **Zero new Python** — `RadioSelect(attrs={...})` already supported by Django
- **Composes with anything** — works with `forms.Form`, `LiveViewForm`, ModelForms, Django admin, djust-theming templates
- **Skip-able** — don't link the CSS file → attribute is inert
- **Override-able** — write your own rule keyed on `[data-dj-inline]` for any visual treatment (segmented controls, CSS Grid columns, etc.)
- **Matches manifesto principle 7** — opinionated where it matters (the contract is `[data-dj-inline]` on the radio, stable forever); flexible where it doesn't (the visual styling is fully replaceable)

## Bundled CSS

- Per-field wrapper: `display: flex; flex-wrap: wrap; gap: 0.5em 1em`
- Each per-choice `<li>`/`<div>`: `display: inline-flex`
- `<label>`: `inline-flex` click target with `0.3em` gap between radio and text
- Resets `margin` on the radio so the gap rule does spacing
- **Preserves the native focus ring** — no `outline: none`

## Test plan

- [x] `pytest python/djust/tests/test_inline_radios_991.py -v` — 12/12 pass
[...]

Issue #1005, opened 2026-04-25
djust_theming.W001 should only run contrast checks on the active pack, not all discovered packs


johnrtipton:
### What's wrong

`djust_theming.W001` runs WCAG AA contrast checks on every registered theme pack × color-preset × mode. With 65+ built-in packs, this produces hundreds of warnings on every `manage.py check` / pod start. Most of those packs are never used by the installing project — they're discovered purely because they ship with djust.

The signal-to-noise ratio is bad enough that the warnings get ignored, which is the opposite of what you want from a check.

### Current behavior

`manage.py check` on a project with `LIVEVIEW_CONFIG['theme'] = {'pack': 'djust', 'default_mode': 'dark'}`:

\`\`\`
System check identified 491 issues
  - ~480 djust_theming.W001 on packs we don't use
  - ~11 real issues (other checks)
\`\`\`

Sample:

\`\`\`
?: (djust_theming.W001) Preset "adaptive" dark mode: text on brand contrast ratio 3.42:1 < 4.5:1 (WCAG AA)
?: (djust_theming.W001) Preset "vercel" light mode: text on primary contrast ratio 4.44:1 < 4.5:1 (WCAG AA)
?: (djust_theming.W001) Preset "stripe" light mode: text on success contrast ratio 2.32:1 < 4.5:1 (WCAG AA)
... (repeated for ~480 preset × mode × token combos)
\`\`\`

Real signal drowns.

### Suggested fix

Default W001 scope to **only the active pack + mode** (resolved from `LIVEVIEW_CONFIG['theme']`). Add an opt-in escape hatch for theme-pack authors:

\`\`\`python
# settings.py (opt-in for anyone authoring a new pack)
DJUST_THEMING = {'contrast_check_scope': 'all'}   # or 'active' (default)
\`\`\`

Alternatively: emit the full set as `INFO` level (silent by default, visible with `--verbosity 2`) and only emit `WARNING` for the active pack.

### Workaround in the meantime

\`\`\`python
SILENCED_SYSTEM_CHECKS = ['djust_theming.W001']
\`\`\`

Silences ALL of them — including the real ones for the active pack if we ever tweak its colors below 4.5:1. Not great.

### Test plan

Integration test: project configured with pack X, mode dark; only pack X / dark tokens get contrast-checked; other packs are skipped (or INFO-level).

### Willing to submit PR
[...]

PR #996, opened 2026-04-24
refactor(tests): replace hand-rolled Redis mock with fakeredis (closes #961)

johnrtipton:
## Summary

Third v0.7.1 PR — test-only refactor. Replace hand-rolled \`FakeRedisClient\` in \`test_security_upload_resumable.py\` with \`fakeredis.FakeRedis\`. Net -19 LOC, more accurate Redis semantics.

## Why

From PR #959 retro: hand-rolled mock (~30 LOC) implemented a minimal get/set/setex/delete/ttl surface — enough at the time but drifts from redis-py behavior. \`fakeredis\` is a full in-memory Redis used by many projects.

## Scope

- \`tests/unit/test_security_upload_resumable.py\` — drop \`FakeRedisClient\`; 7 instantiation sites → \`fakeredis.FakeRedis()\`; \`OldRedis\`/\`BrokenRedis\` subclasses now inherit from \`fakeredis.FakeRedis\` with redis-py method signatures; direct \`.data[]\`/\`.ttls[]\` access → real Redis commands (\`client.set()\`, \`client.ttl()\`, \`client.exists()\`).
- \`pyproject.toml\` — add \`fakeredis>=2.20,<3\` to \`[dev]\`.
- \`uv.lock\` — regenerated.

## Test tolerance

\`test_setex_fallback_for_older_redis_py\` previously asserted \`client.ttls[...] == 30\` (exact). \`fakeredis\` uses real wall-clock (±1s slack between \`set()\` and \`ttl()\`). Loosened to \`29 <= client.ttl(...) <= 30\` — mirrors what a real Redis-backed integration test would see.

## Test plan

- [ ] \`pytest tests/unit/test_security_upload_resumable.py\` — 70/70 pass locally ✅
- [ ] CI full suite green

Closes #961.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1014, opened 2026-04-25
fix(checks): A070 no longer false-positives on {% verbatim %}-wrapped examples (closes #1004)

### LINKED ISSUE #1004: djust.A070 false positive on {% verbatim %}-wrapped dj_activity examples
### What's wrong

The A070 check (`dj_activity` missing `name=` argument) scans template source as raw text. If a template contains a literal example of `{% dj_activity %}` inside a `{% verbatim %}...{% endverbatim %}` block — a common pattern on docs / marketing pages that document the tag — A070 treats it as a real uninstrumented dj_activity call and fires.

### Repro

\`\`\`html
{# templates/foo.html #}
<code>{% verbatim %}{% dj_activity %}{% endverbatim %}</code>
\`\`\`

Run `manage.py check`. A070 fires pointing at the verbatim-wrapped example.

### Evidence

docs.djust.org's home page does exactly this to document djust 0.7's new tags. Every pod start emits:

\`\`\`
?: (djust.A070) templates/docs/home.html:72 -- {% dj_activity %} is missing a 'name' argument.
    HINT: Every {% dj_activity %} block must have a non-empty name: {% dj_activity "my-panel" visible=expr %}.
\`\`\`

The template source is intentionally showing the tag as literal text via `{% verbatim %}`; there's no actual `dj_activity` call.

### Suggested fix

Before scanning for tag patterns, strip `{% verbatim %}...{% endverbatim %}` regions from the template content being analyzed. The same fix applies to any other content-scanning template check (the whole `_check_*_in_templates` family, A07x) — they all risk the same false positive on any literal tag shown as documentation content.

Minimal helper:

\`\`\`python
_VERBATIM_RE = re.compile(r'\{%\s*verbatim\s*%\}.*?\{%\s*endverbatim\s*%\}', re.DOTALL)

def _strip_verbatim(src: str) -> str:
    return _VERBATIM_RE.sub('', src)
\`\`\`

Apply in the template-scanning checks before the tag regex runs.

### Test plan

Fixture template with `{% verbatim %}{% dj_activity %}{% endverbatim %}` should produce zero A07x hits. Regression test in `python/tests/test_checks.py`.

### Willing to submit PR
Yes — full plan in [`docs.djust.org/UPSTREAM_TODO.md`](https://github.com/djust-org/docs.djust.org/blob/main/UPSTREAM_TODO.md#djust-orgdjust--a070-false-positive-on--verbatim-wrapped-dj_activity-examples).

johnrtipton:
## Summary

Closes #1004 — `djust.A070` (`dj_activity` missing `name=` argument) no longer flags literal `{% dj_activity %}` examples inside `{% verbatim %}...{% endverbatim %}` blocks. Common pattern on docs / marketing pages that document the tag.

## Root cause

The A070 / A071 scanner at `python/djust/checks.py` walks template source as raw text via `_DJ_ACTIVITY_TAG_RE.finditer(content)`. Django's `{% verbatim %}` blocks render their content as-is without parsing the tags inside, so `{% dj_activity %}` literals on documentation pages got flagged as real uninstrumented activity calls.

## Fix

New `_strip_verbatim_blocks(content)` helper redacts the BODY of every `{% verbatim %}...{% endverbatim %}` region before the A070 / A071 regex scan. Both unnamed (`{% verbatim %}`) and Django's named form (`{% verbatim foo %}`) are handled.

Newlines inside the region are preserved (we replace every non-newline char with a space) so line numbers from `match.start()` stay accurate for matches OUTSIDE the region. Fast path: if `"verbatim"` doesn't appear in the source at all, return `content` unchanged — common case adds zero overhead.

The scanner's existing iteration is updated to scan `_strip_verbatim_blocks(content)` instead of `content`. Real uninstrumented `{% dj_activity %}` calls outside any verbatim block continue to fire A070 unchanged.

## Test plan

- [x] `pytest python/tests/test_a070_verbatim_fp_1004.py -v` — 12/12 pass

## Coverage

12 regression tests in two classes:

**`TestStripVerbatimBlocks` (helper contract — 7 tests)**:
- No-verbatim fast-path returns unchanged
- Unnamed verbatim block redacted
- Named verbatim block redacted
- Line numbers preserved across redaction
- Multiple verbatim blocks all redacted
- Verbatim block with non-dj_activity content doesn't affect real calls
- Multi-line verbatim body fully redacted

**`TestA070VerbatimSuppression` (scanner integration — 5 tests)**:
- Canonical docs case (verbatim-wrapped example → no A070)
- Real uninstrumented `{% dj_activity %}` outside verbatim still fires (negative regression)
- Mixed verbatim + real call → only real counts
- Named `{% dj_activity "panel" %}` in verbatim → no A070/A071
- Real named call unaffected by verbatim neighbor

CHANGELOG entry under `[Unreleased]` for v0.7.3.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,26 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- **`djust.A070` no longer false-positives on `{% verbatim %}`-wrapped
+  `dj_activity` examples (v0.7.3, #1004)** — the A070 / A071 scanner
+  walks template source as raw text. Templates that document the
+  `{% dj_activity %}` tag — common pattern on docs / marketing pages
+  that include literal example markup wrapped in `{% verbatim %}` so
+  Django renders the example as-is — got flagged as real
+  uninstrumented activity calls. Fix: new
+  `_strip_verbatim_blocks(content)` helper redacts the BODY of every
+  `{% verbatim %}...{% endverbatim %}` region (both unnamed and
+  Django's named-form `{% verbatim foo %}...{% endverbatim foo %}`)
+  before the regex scan. Newlines inside the region are preserved so
+  line numbers from `match.start()` stay accurate for matches OUTSIDE
+  the region. The scanner's existing iteration over
+  `_DJ_ACTIVITY_TAG_RE` runs against the redacted source. Real
+  uninstrumented `{% dj_activity %}` calls outside any verbatim block
+  continue to fire A070 unchanged. Covered by **12 regression tests**
+  in `python/tests/test_a070_verbatim_fp_1004.py` (7 helper-contract
+  tests + 5 scanner-integration tests including the canonical docs
+  case, mixed verbatim + real calls, named verbatim form, and line
+  number preservation).
 - **`djust.C011` now catches stale/placeholder `output.css`, not
   just totally-missing files (v0.7.3, #1003)** — `_check_missing_compiled_css`
   in `python/djust/checks.py` previously tested only
--- a/python/djust/checks.py
+++ b/python/djust/checks.py
@@ -1981,6 +1981,58 @@ _DEPRECATED_DATA_DJ_ID_RE = re.compile(r"""data-dj-id\s*=\s*["'][^"']*["']""")
 # renderer is in use (raw HTML escaped, provisional-line splitter active).
 _DJ_MARKDOWN_TAG_RE = re.compile(r"\{%\s*djust_markdown\b")
 
+# #1004 — `{% verbatim %}...{% endverbatim %}` regions hold raw template
+# source that Django renders as-is without parsing. Docs and marketing
+# templates routinely wrap literal `{% dj_activity %}` examples in
+# verbatim blocks; the A070 / A071 scanner used to treat those as real
+# uninstrumented activity calls and false-positive. The pre-processor
+# below replaces each verbatim region with whitespace (preserving line
+# breaks) so downstream regex scanners see no `{% dj_activity %}` text
+# inside the region, while line numbers from the original source remain
+# accurate for any matches OUTSIDE the region.
+#
+# Both unnamed (`{% verbatim %}`) and named (`{% verbatim foo %}`)
+# variants are matched. The endverbatim form must match the opening
+# variant, but Django allows either to close — we accept either since
+# the goal is to redact the BODY of the region, not validate Django
+# syntax (the template engine itself will reject malformed verbatim
+# blocks at render time).
+_VERBATIM_BLOCK_RE = re.compile(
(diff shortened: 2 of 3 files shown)


johnrtipton:
## Per-PR Retro — #1014 (v0.7.3 drain iter 2)

**Date**: 2026-04-25
**Scope**: A070 / A071 verbatim false-positive suppression via `_strip_verbatim_blocks` pre-processor; closes #1004.

### What Worked Well

- **Whitespace-preserving redaction is the cheap right answer.** Replacing the verbatim body with whitespace (preserving newlines) keeps line numbers accurate without forcing the scanner to track offsets explicitly. Locked with a dedicated test that asserts `out.count("\n") == src.count("\n")`.
- **Fast-path (`if "verbatim" not in content: return content`) means common case adds zero overhead.** Most templates don't contain verbatim blocks; the regex pass + rewrite only runs when one exists.
- **Tests in their own file.** `test_a070_verbatim_fp_1004.py` covers ONLY the verbatim handling, separate from the existing A070 / A071 broad-behavior tests. A future PR that loosens the verbatim handling would fail this file specifically — clean blast-radius signal.
- **Both helper-contract AND scanner-integration tests.** 7 tests on the helper alone (locks the redaction semantics: unnamed, named, line-preservation, multiple blocks, empty-of-target-content, multi-line) + 5 tests on the actual scanner path (canonical case + 4 negative regressions).

### What Didn't Work

- **Pre-commit stash/restore — 5th occurrence in 24-hour session.** Ruff reformatted `[: line5_offset_src]` → `[:line5_offset_src]` (whitespace style), original commit didn't register, had to re-stage and retry. Same pattern as PRs #989, #996, #1007, #1008. **Action #122 is now 5× reinforced** — implementing the post-commit `git log -1 --oneline` verification in pipeline-run can't slip any further.

### What to Improve

- **Action #122 implementation is no longer optional.** Five reinforcements in one session means the cumulative cost of NOT implementing the post-commit verification is now larger than the implementation cost. Should be the first item in the next housekeeping pass.

### Review Stats

| Metric | Value |
|---|---|
| Tests added | 12 |
| Production LOC delta | +60 (helper + comment + integration line) |
| Test LOC delta | +220 |
| 🔴 findings | 0 |
| 🟡 findings | 0 |
| Pre-commit attempts | 2 (#122 5th reinforcement) |
| Pre-push attempts | 1 |
| CI retries | 0 |
| Quality rating | 5/5 |

- **Action #122 promoted to "implementing this is the next housekeeping action"** — 5 reinforcements in single session is decisive.
- **Whitespace-preserving redaction pattern** worth adding to the pipeline-shared check-authoring guidance (when a regex scanner needs to ignore a region while preserving line numbers, replace the body with whitespace, keep newlines).

### Milestone context

v0.7.3 drain iter 2 of 3. Next: #1005 (W001 active-pack contrast scoping).

PR #1023, opened 2026-04-25
feat(forms): dj-form-pending — React 19 useFormStatus equivalent (v0.8.0 iter 1)

johnrtipton:
## Summary

First v0.8.0 ROADMAP feature: `dj-form-pending` attribute. React 19's `useFormStatus` lets nested components read whether the parent form is currently submitting without prop drilling — this is the djust equivalent. Any element nested inside `<form dj-submit>` can declare `dj-form-pending="hide|show|disabled"` and react automatically.

## API

```html
<form dj-submit="save">
    <input name="title" />

    <button type="submit">
        <span dj-form-pending="hide">Save</span>
        <span dj-form-pending="show" hidden>Saving…</span>
    </button>

    <input name="email" dj-form-pending="disabled" />
</form>
```

The form gets `data-djust-form-pending="true"` while pending, so CSS hooks work without JS:

```css
form[data-djust-form-pending] .spinner { display: block; }
```

## Modes

| Mode | Behavior while pending | Behavior when idle |
|---|---|---|
| `hide` | hidden via `hidden` attribute | visible |
| `show` | visible | hidden |
| `disabled` | `disabled = true` | original disabled state restored |

User-disabled elements stay disabled — pre-pending state tracked in `data-djust-form-pending-was-disabled`.

## Implementation

- New `_setFormPending(form, pending)` helper in `09-event-binding.js`
- 1-line wiring into `_handleDjSubmit`: now wraps `await handleEvent(...)` in `try { ... } finally { _setFormPending(form, false) }` so error paths still clear pending state

## Test plan

- [x] `npx vitest run tests/js/dj-form-pending.test.js` — 8/8 pass
- [x] `make build-js` succeeds; bundle delta ~80 B gzipped
- [x] Pre-push runs full JS suite (1389 tests) — pass

## Coverage

8 regression tests in `tests/js/dj-form-pending.test.js`:

1. `data-djust-form-pending` toggle on the form
2. `dj-form-pending="hide"` mode
3. `dj-form-pending="disabled"` mode
4. User-disabled preservation (regression: don't re-enable user-locked inputs)
5. Plain forms without `dj-submit` are unaffected
6. Scope isolation: two-form page, only the submitting form's descendants react
7. Error-path cleanup via `finally` block
8. Unknown mode forward-compat (`dj-form-pending="future-mode"` is silently ignored)

## v0.8.0 milestone context

This is iter 1 of 3 for the v0.8.0 milestone:
[...]

PR #1024, opened 2026-04-25
feat(actions): @action decorator — React 19 Server Actions equivalent (v0.8.0 iter 2)

johnrtipton:
## Summary

Second v0.8.0 ROADMAP feature: `@action` decorator. The server-side complement to PR #1023's `dj-form-pending`. Together they give React 19-level form ergonomics with zero per-handler wiring.

## API

```python
from djust import action

class TodoView(LiveView):
    @action
    def create_todo(self, title: str = "", **kwargs):
        # While this body runs, _action_state["create_todo"] is
        # {"pending": True, "error": None, "result": None}.
        if not title:
            raise ValueError("Title is required")
        todo = Todo.objects.create(title=title, user=self.request.user)
        return {"created": todo.id}
        # On success: state becomes {"pending": False, "error": None, "result": {"created": <id>}}
        # On exception: state becomes {"pending": False, "error": str(exc), "result": None}
        #   and the exception is re-raised so the dispatch path's logging works.
```

Templates access action state via context injection — each action's name becomes a context variable:

```html
{% if create_todo.error %}
    <div class="error">{{ create_todo.error }}</div>
{% elif create_todo.result %}
    <div class="success">Todo {{ create_todo.result.created }} created!</div>
{% endif %}
```

## Why this complements `dj-form-pending`

| Feature | Side | Concern |
|---|---|---|
| `dj-form-pending` (PR #1023) | Client | In-flight UX during the network round-trip — toggle visibility/disabled state |
| `@action` (this PR) | Server | Post-completion state — surface errors and results to the template after handler returns |

Together: form authors don't write try/except + manual state-setting; the framework standardizes the pending/error/result triple.

## Implementation

- **`djust.decorators.action`**: new decorator. Wraps `@event_handler` (every action is also an event handler — same dispatch path, parameter coercion, permissions, rate limits) and adds the action-state tracking layer.
- **`LiveView.__init__`**: initializes `_action_state: Dict[str, Dict] = {}`.
- **`ContextMixin.get_context_data()`**: injects each action's state under its name. Action keys win over user attrs on collision (the canonical reading of that name).
- **State-reset semantics**: re-running an action clears previous state. A success retry after a failure clears the error; a failure after a success clears the result. The template never sees stale + fresh state together.
- **Self-init defense**: if a subclass `__init__` forgot `super().__init__()`, the decorator initializes `_action_state` on demand instead of crashing.
- **Both decorator forms**: bare `@action` and called `@action(description=...)` supported.
- **`is_action(func)`** helper for runtime detection (parallel to `is_event_handler`).
- **Exposed as top-level**: `from djust import action, is_action`.

## Test plan

- [x] `pytest python/djust/tests/test_action_decorator.py -v` — 18/18 pass

## Coverage

18 regression tests across 6 classes:

| Class | # tests | Covers |
|---|---|---|
| `TestActionMetadata` | 4 | decorator metadata, `is_action` / `is_event_handler` distinction |
| `TestActionSuccess` | 3 | sync return, None return, self-init defense |
| `TestActionException` | 3 | message capture, class-name fallback for empty msg, re-raise |
| `TestMultipleActions` | 2 | independent state, one failing doesn't affect another |
| `TestReRunSemantics` | 2 | error→success transition (clears error), success→error (clears result) |
| `TestDecoratorForms` | 2 | bare `@action`, called `@action(description=...)` |
| `TestContextInjection` | 2 | end-to-end via `ContextMixin.get_context_data()` |

CHANGELOG entry under `[Unreleased]` / Added.

## v0.8.0 milestone status

- ✅ iter 1: Form Status Awareness (PR #1023, `dj-form-pending`)
- ✅ iter 2: Server Actions (this PR, `@action`)
- ⏳ iter 3: Async Streams — substantial Python + Rust feature; will likely cut v0.8.0rc1 with iter 1 + iter 2 (matches v0.7.0rc1 precedent of 4-of-5 P2 shipped) and defer Async Streams to v0.8.1 if scope-cut required.


johnrtipton:
## Per-PR Retro — #1024 (v0.8.0 iter 2)

**Date**: 2026-04-25
**Scope**: `@action` decorator (React 19 Server Actions equivalent); v0.8.0 iter 2.
**Verdict**: 5/5 quality — first-push clean merge, 0 🔴, 0 🟡, all checks green.

### What Worked Well

- **Built on existing infrastructure**. `@action` wraps `@event_handler` rather than reimplementing the dispatch path. Same parameter coercion, permissions, rate limits. Adds only the action-state tracking layer.
- **Pairs cleanly with iter 1**. `dj-form-pending` covers client-side in-flight UX; `@action` covers server-side post-completion state. Together = React 19 form ergonomics with zero wiring.
- **Re-run state semantics locked early**. Two tests for transitions: error→success clears error; success→error clears result. Templates never see stale state alongside fresh state.
- **Self-init defense**. If a subclass forgot `super().__init__()`, the decorator initializes `_action_state` on demand. Locked with `test_action_self_initializes_action_state`.
- **Bare-form + called-form both supported**. `@action` and `@action(description="…")` both work cleanly via the standard `func is None` test pattern.

### Review Stats
| Metric | Value |
|---|---|
| Tests added | 18 |
| Production LOC | +175 (decorator + LiveView init + ContextMixin injection + exports) |
| Test LOC | +290 |
| 🔴 / 🟡 findings | 0 / 0 |
| Pre-commit attempts | 1 |
| CI retries | 0 |

### Milestone context

v0.8.0 iter 2 of 3 shipped. Per v0.7.0 precedent (rc1 cut after 4 of 5 P2 shipped, Islands deferred), v0.8.0rc1 will cut now with iter 1 + iter 2. Async Streams (iter 3) is substantial Python + Rust VDOM work — deferring to v0.8.1 to keep v0.8.0 scope focused on the React 19 form-action headline.

PR #1067, opened 2026-04-25
security(api): JSON-parse error consistency + ws cache log (closes #1026, #1030)

johnrtipton:
## Summary

v0.8.1 reconcile drain — Group B (consistency batch). #1045 closed separately as moot — the regex constant is already centralized at \`python/djust/templatetags/live_tags.py:39\` with one call site.

- **Closes #1026 (security):** \`python/djust/api/dispatch.py\` had two sites returning \`f\"Malformed JSON body: {exc}\"\` — small but real stack-trace-style leak (parser internals could surface in the response). Aligned to the existing \`observability/views.py:401\` pattern: \`logger.exception\` for server-side diagnosis, generic \`\"Malformed JSON body — see server logs\"\` for the client. The \`invalid_json\` error code is unchanged.
- **Closes #1030:** \`03-websocket.js:386\` cache-put exceptions were swallowed via \`catch (_e) {}\`. Now logs under \`if (globalThis.djustDebug) console.log(...)\` per project convention.

CHANGELOG.md: entries under \"Security\" (#1026) and \"Changed\" (#1030).

## Test plan

- [x] \`pytest python/djust/tests/test_api_dispatch.py python/djust/tests/test_observability_dry_run.py\` — 31/31 passed
- [x] \`npm test\` — 1389/1389 JS tests passed
- [x] Pre-push hook ran the full python suite — passed
- [x] Client.js bundle artifacts rebuilt to keep the production-ready bundle in sync

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Co-Authored-By: Claude Opus 4.7 (1M context) <noreply@anthropic.com>

Issue #1033, opened 2026-04-25
tech-debt: djust[admin] extra vs djust.admin_ext module name divergence


johnrtipton:
**Source**: Retro v0.6.0 / PR #971
**Action Tracker**: row #109

**Action**: `djust[admin]` extra vs `djust.admin_ext` module name divergence

**Notes**: Rename one or the other in v0.7.0

---
_Filed by `/pipeline-retro --reconcile` on 2026-04-25 as part of Scope C: file all Action Tracker rows as GitHub issues._

johnrtipton:
Closing — code-reading at Stage 4 found this is intentional, not a defect.

The naming divergence is documented and rationalized in `docs/website/guides/migration-from-standalone-packages.md:174`:

> Django already ships `django.contrib.admin`. A module named `djust.admin` would collide at import time with any project that mixes `django.contrib.admin` and `djust` (which is nearly all of them). `djust.admin_ext` (ext = "extensions") is the safe name that avoids the collision. The extra is still spelled `djust[admin]` because the package-level name is short and intuitive; only the Python module picks up the `_ext` suffix.

Per ROADMAP.md:563:
> `djust-admin` → `djust[admin]` (...) Move to `python/djust/admin_ext/` (avoid collision with `django.contrib.admin`)

The original tracker entry (Action #109) was filed during the v0.6.0 retro before the rationale was crystallized in the migration guide. The current naming is the deliberate choice — closing as no-action.

Issue #1032, opened 2026-04-25
tech-debt: Dashboard→Dashboard re-mount limitation in sticky LiveView demo; {% live_render %} doesn't auto-detect preserved stickies


johnrtipton:
**Source**: Retro v0.6.0 / PR #969
**Action Tracker**: row #108

**Action**: Dashboard→Dashboard re-mount limitation in sticky LiveView demo; `{% live_render %}` doesn't auto-detect preserved stickies

**Notes**: v0.6.x/v0.7.0 enhancement — teach tag to emit slot markers automatically

---
_Filed by `/pipeline-retro --reconcile` on 2026-04-25 as part of Scope C: file all Action Tracker rows as GitHub issues._

johnrtipton:
Closing — relocating to v0.9.0+ ROADMAP planning as a real feature item.

Code-reading at Stage 4 confirmed this isn't a 1-PR refactor. Auto-detecting preserved stickies in `{% live_render %}` requires:

1. Server-side template-tag intelligence to detect when a sticky with the same `sticky_id` is already mounted on the requesting client (currently impossible without state).
2. Client→server coordination — either a cookie carrying the list of preserved sticky IDs, or a header/query-param sent on every navigation, or a WebSocket-side handshake re-using the existing connection.
3. Slot-marker emission (`dj-sticky-slot` instead of `live_render`) when the client already holds the sticky.
4. Tests covering Dashboard → Settings → Dashboard (return-trip) and Dashboard → fresh-tab.

The current demo `dashboard.html` already documents the limitation in an explicit `{% comment %}` block:

> NOTE (limitation): returning to Dashboard from Settings/Reports will re-mount the stickies fresh (server collision — see ADR-011's 'Failure modes' section and README step 4). Dashboard -> Settings -> Reports is the preservation happy path demonstrated here.

Filing as deferred-to-v0.9.0 ROADMAP entry — same shape as #1041, #1042, #1043, #1044 from the same reconcile triage. Closes the v0.8.1 drain item, opens the v0.9.0 feature backlog item.

johnrtipton:
Reopening — promoted into v0.9.0 shape C as the P1 1.0-blocker. Pipeline starting now (auto-detect preserved stickies).

PR #1068, opened 2026-04-25
feat(ws): mount_batch fallback for old-server compat (closes #1031)

### LINKED ISSUE #1031: tech-debt: No version-probe fallback for mount_batch — older servers produce generic "unknown msg type"; client should fall back gracefully
**Source**: Retro v0.6.0 / PR #970
**Action Tracker**: row #107

**Action**: No version-probe fallback for `mount_batch` — older servers produce generic "unknown msg type"; client should fall back gracefully

**Notes**: Tech-debt

---

johnrtipton:
## Summary

v0.8.1 reconcile drain — Group F (solo, mount_batch back-compat).

When a v0.6.0+ client (which sends \`mount_batch\` frames for lazy hydration efficiency, PR #970) talks to a pre-v0.6.0 server, the server returns \`{type:\"error\", error:\"Unknown message type: mount_batch\"}\` and the lazy-hydrated views never mount. This PR adds a graceful fallback to per-view \`mount\` calls.

## Implementation

- **\`13-lazy-hydration.js\`** — new \`inFlightBatch\` slot on \`lazyHydrationManager\`. Populated when \`processPendingMounts\` sends a \`mount_batch\` frame; cleared by the websocket success/error handlers. New \`handleMountBatchFallback()\` method iterates the stashed mounts and falls back to per-view \`mountElement()\`. Idempotent — clears \`inFlightBatch\` BEFORE iterating so a late \`mount_batch\` response can't double-trigger.
- **\`03-websocket.js\`** — \`case 'error':\` now checks if \`data.error\` matches \`/mount_batch|unknown\\s+message\\s+type/i\` AND the lazyHydration manager exists; if both, invokes the fallback and short-circuits the rest of the error handler. \`case 'mount_batch':\` clears \`inFlightBatch\` on success so a late error frame can't trigger phantom fallback.

## Test plan

- [x] **7 new JSDOM tests** at \`tests/js/mount-batch-fallback.test.js\` — covers fallback existence, idempotency guard, early-return on empty batch, stash-on-send, error-handler dispatch, success-clears-inflight, and djustDebug-gating of debug log.
- [x] \`npm test\` — 1396/1396 passed (was 1389; +7 new).
- [x] Pre-push hook ran the full python suite — passed.
- [x] No production code touched (Python untouched; only JS source + bundle artifacts).

## Why this matters

\`mount_batch\` is enabled by default for any page with 2+ lazy-hydrated views (\`window.DJUST_USE_MOUNT_BATCH\` defaults to \`true\`). Without this fallback a v0.6.0+ client + v0.5.x server combination silently fails the entire lazy-hydration path on those pages. With it, the fallback transparently degrades to N \`mount\` frames per page.


johnrtipton (commented):
## Review — `mount_batch` fallback for old-server compat

**Verdict: APPROVE**. Tight, correctly-scoped compat shim. Closes #1031. 1396/1396 JS tests pass locally; the 7 new tests pass.

### `13-lazy-hydration.js`

- `inFlightBatch` slot is well-commented and stashed immediately before `liveViewWS.sendMessage({type:'mount_batch', ...})`.
- `handleMountBatchFallback()`:
  - Early-return on `!this.inFlightBatch` — handles both never-sent and already-handled cases.
  - **Idempotency guard ordering is correct**: reads `mounts` → nulls `inFlightBatch` → then `forEach`. A re-entrant call short-circuits on the early return.
  - Attribute reset (`removeAttribute('data-live-hydrated')` + `setAttribute('dj-lazy','')`) gives `mountElement` a clean re-entry: `mountElement` itself re-removes `dj-lazy` and re-sets `data-live-hydrated='true'` on success (lines 265-266), so the optimistic state from `processPendingMounts` (lines 205-206) is properly rewound first.
  - Debug log gated on `globalThis.djustDebug`.

### `03-websocket.js`

- `case 'mount_batch':` clears `inFlightBatch` at the top of the block (right after `viewMounted = true`, before iterating `views`). Success clear can't be skipped by a mid-block error.
- `case 'error':` regex `/mount_batch|unknown\s+message\s+type/i` is reasonably scoped. The `mount_batch` branch matches the canonical `"Unknown message type: mount_batch"` string. The `|unknown\s+message\s+type` branch is broader — but the worst case is a no-op since `handleMountBatchFallback` early-returns when `inFlightBatch` is null. Safe.
- Existence guards (`window.djust && window.djust.lazyHydration && typeof ... === 'function'`) are belt-and-suspenders correct.

1. **Early `break` skips post-error housekeeping**: dispatches `djust:error` CustomEvent, `_pendingEventRefs.clear()`, and `globalLoadingManager.stopLoading()`. Fine for mount_batch errors (no user event or loading state attached to batch-mount frames), but a one-line comment noting the intent would help future readers.
2. **Test style**: source-grep assertions are unusual but defensible — they pin structural invariants (ordering, presence of guard) that behavioral tests couldn't easily express. They'd catch reordering of the idempotency guard, removal of the early return, or removal of the success-clear.
3. **Regex tightening (optional)**: `unknown\s+message\s+type:\s*mount_batch` would eliminate the false-positive surface entirely. Current breadth is safe but broader than strictly needed.

### Checklist

- Title `feat(ws): ...` ✓
- CHANGELOG under "Added", v0.8.1 reconcile drain Group F framing ✓
- New JS feature has corresponding `tests/js/` file ✓
- No unguarded `console.log` ✓
- Bundle artifacts rebuilt (~+170 bytes for ~30 LOC plausible) ✓
- Closes #1031 ✓

**Verdict: APPROVE**

johnrtipton:
## Per-PR Retro — #1068 (v0.8.1 Group F — mount_batch fallback)

**Date**: 2026-04-25
**Scope**: Solo task — `mount_batch` back-compat fallback for old-server scenario (#1031). When v0.6.0+ client talks to pre-v0.6.0 server, fall back to per-view mount frames.
**Verdict**: 4.5/5 — first-push clean, Stage 11 APPROVE, but I committed to local main by mistake at first (then recovered without pushing).

### What Worked Well

- **Idempotency-by-design.** The `inFlightBatch = null` clear happens BEFORE `forEach` in `handleMountBatchFallback()`. Re-entrant call returns early. Late-arriving success response also clears. Two-sided guard against double-trigger. Test #2 explicitly asserts `nullIdx < forEachIdx`.
- **State-on-the-manager pattern beat closure-tracking.** First instinct was to track in-flight batch via a closure variable inside `processPendingMounts`. Putting it on the manager itself (`this.inFlightBatch`) made the fallback dispatch from the websocket error handler clean — just `window.djust.lazyHydration.handleMountBatchFallback()`. No closure chain, no event-bus indirection.
- **Source-level test pattern carried over from #619 / mount-deferred-reinit.** Source assertions catch the regression exactly without faking the WebSocket lifecycle. 7 tests, 260ms total — cheap to run, cheap to maintain.
- **Pre-existing `globalThis.djustDebug` discipline** (Action #1030 just added a few hours ago) was reused for the fallback's debug log. Two PRs in this drain (#1067, #1068) demonstrate the pattern is now reflexive.

### What Didn't Work

- **🔴 Committed Group F to local main by mistake.** After closing #1033 (Group E was a no-op), I switched back to main but didn't create a fresh feature branch before starting Group F's work. The commit landed on local main. Recovered cleanly: `git checkout -B feat/mount-batch-fallback-v081` (preserves commit), `git checkout main && git reset --hard origin/main` (clean local main), push the feature branch. Origin/main was never affected. **Lesson: when transitioning between groups in a drain, ALWAYS `git checkout -B <new-branch> origin/main` before the first edit, even if the previous group ended on a clean main checkout.** Filing as a Stage 1 reminder for the pipeline-run skill.

### Review Stats

| Metric | Value |
|---|---|
| Tests added | 7 (mount-batch-fallback.test.js) |
| Production LOC | +49 (JS source) + bundle artifacts |
| Test LOC | +95 |
| 🔴 / 🟡 findings | 0 / 0 |
| Pre-commit attempts | 1 (clean) |
| Process error | 1 (commit-on-main; self-recovered) |
| Quality rating | 4.5/5

### Stage 3.5 classification

For the eventual v0.8.1 milestone retro:

- **Action taken**: `Closed — Closes #1031 (lazy-hydration mount_batch fallback in python/djust/static/djust/src/13-lazy-hydration.js handleMountBatchFallback + python/djust/static/djust/src/03-websocket.js error/mount_batch handlers + 7 JSDOM tests).`

`closed` shape — Stage 3.5 gate clean.

### Pattern: branch-from-target before edit, even between drain groups

Add to authoring discipline: **at the start of EVERY pipeline group/iteration, run `git checkout -B <branch-name> origin/main` before the first file edit**, regardless of what state the previous group left the working tree in. The "I'm already on a clean main" state is an attractive nuisance — the next commit will land directly on main if you don't branch first. Costs 2 seconds; prevents the recovery dance.

### Milestone context

v0.8.1 reconcile drain — Group F of 7 (5 down, 1 to go). Group E (#1033 admin naming) closed at Stage 4 as intentional-not-defect. Only Group G (#1032 sticky LiveView re-mount) remains.

Issue #1010, opened 2026-04-25
Stale .md references + dead-anchor links in source markdown (found via docs.djust.org link crawl)


johnrtipton:
### What's wrong

docs.djust.org's [`scripts/link_check.py`](https://github.com/djust-org/docs.djust.org/blob/main/scripts/link_check.py) crawls the rendered site and found **10 broken references** that all trace back to the source markdown in this repo. These break on GitHub's own markdown viewer too — it's a content problem, not a rendering one.

### Broken `.md` targets that don't exist anywhere in the repo

All linked from [`docs/components/RUST_COMPONENTS.md`](https://github.com/djust-org/djust/blob/main/docs/components/RUST_COMPONENTS.md):

- `LIVEVIEW.md`
- `TEMPLATES.md`
- `PYTHONIC_FORMS_IMPLEMENTATION.md`

I `find`-ed the whole repo and none of these files exist. Likely renamed or deleted without updating the RUST_COMPONENTS.md references.

### Dead anchors — link's `#frag` doesn't match any heading ID on the target page

- `docs/components/API_REFERENCE_COMPONENTS.md#component-simplestateless`
- `docs/components/API_REFERENCE_COMPONENTS.md#livecomponent-statefulreactive`
- `docs/components/COMPONENT_EXAMPLES.md#simple-components-library`
- `docs/components/COMPONENT_EXAMPLES.md#livecomponents-library`
- `docs/components/COMPONENT_EXAMPLES.md#advanced-patterns`
- `docs/components/COMPONENT_UNIFIED_DESIGN.md#unified-component-base-class`
- `docs/guides/error-codes.md#l040-gitconfig-publicly-accessible` (linked from `docs/guides/djust-audit.md`)

### Repro

```bash
cd docs.djust.org
git submodule update --init
make dev &
uv run python scripts/link_check.py
```

Reports the above as the only remaining broken links after every other kind of renderer-side rewrite is applied.

### Suggested fix

Mostly mechanical:
- For missing `.md` files: either restore them (git log may show the rename/delete) or edit RUST_COMPONENTS.md to drop the references.
- For dead anchors: update the `#frag` to match the actual heading IDs GitHub emits on the target page, or add explicit `<a id="…">` anchors at those headings.

### Downstream

docs.djust.org has these exact URLs in an `ALLOWLIST` so CI doesn't flap. Happy to remove the allowlist entries (and open a follow-up PR to that effect) once this lands.

### Willing to submit PR

[...]

Issue #145, opened 2026-01-31
Add duplicate key detection warning


johnrtipton:
## Summary

The VDOM keyed diff algorithm silently accepts duplicate `data-key` values in sibling children. Since key-to-index mapping uses `HashMap`, duplicate keys result in last-one-wins behavior, making earlier children invisible to the diff.

## Current Behavior

```html
<div data-key="item-1">First</div>
<div data-key="item-1">Second</div>  <!-- silently shadows First -->
<div data-key="item-2">Third</div>
```

No warning is emitted. The diff may produce incorrect patches for the shadowed element.

## Proposed Change

Add a debug warning in `diff.rs` at lines 231-241 where `old_keys` and `new_keys` HashMaps are built. When a duplicate key is detected during map construction, emit a `vdom_trace!` warning similar to the existing mixed keyed/unkeyed warning.

```rust
// During map construction, check for duplicates
if old_keys.contains_key(k) {
    vdom_trace!("WARNING: Duplicate key '{}' in old children at indices {} and {}. ...", k, existing, i);
}
```

## Context

Found during VDOM torture testing (see `crates/djust_vdom/tests/torture_test.rs`, test `torture_duplicate_keys`). The test confirms no crash occurs, but results are suboptimal.

PR #1074, opened 2026-04-25
feat(theming): theming polish batch (closes #1009, #1011, #1012, #1013)

### LINKED ISSUE #1009: Ship prose.css so sites don't re-invent the @tailwindcss/typography ↔ pack bridge
### What's missing

`djust.theming` ships `components.css` wired to the active pack's shadcn tokens (button, card, alert, badge, breadcrumb, avatar, etc.). But any djust site that renders markdown — docs, blog, in-app help — uses Tailwind's `@tailwindcss/typography` plugin, which assumes a light background and emits its own `--tw-prose-*` variables.

Nothing in djust bridges the two. Every site that opts into typography ends up hand-writing ~95 lines of `.prose.prose-<pack> { --tw-prose-body: var(--color-brand-text); ... }` overrides to get dark-mode-correct prose.

docs.djust.org hit this and wrote the bridge in its own `input.css` — see [`static/css/input.css`](https://github.com/djust-org/docs.djust.org/blob/main/static/css/input.css). It's self-contained, pack-aware (reads the same `--color-brand-*` tokens the pack emits), and handles both light and dark via the existing pack scope.

**It would work unchanged for any djust site using @tailwindcss/typography.**

### Proposed contract

Add `djust_theming/static/djust_theming/css/prose.css`:

```css
/* Pack-aware overrides for @tailwindcss/typography.
 * Opt-in: add `prose-djust` alongside `prose` on your <article>.
 * NOT in @layer on purpose: typography plugin emits .prose unlayered,
 * and unlayered rules always beat layered ones. */
.prose.prose-djust {
    --tw-prose-body:          var(--color-brand-text);
    --tw-prose-headings:      var(--color-brand-text);
    --tw-prose-links:         var(--color-brand-rust);
    --tw-prose-bold:          var(--color-brand-text);
    --tw-prose-code:          var(--color-brand-text);
    --tw-prose-pre-bg:        var(--color-brand-panel);
    --tw-prose-pre-code:      var(--color-brand-text);
    --tw-prose-quote-borders: var(--color-brand-rust);
    --tw-prose-bullets:       var(--color-brand-border);
    --tw-prose-th-borders:    var(--color-brand-border);
    --tw-prose-td-borders:    var(--color-brand-border);
    --tw-prose-hr:            var(--color-brand-border);
}

/* Element-level tuning: pre cards, h2 chapter breaks, blockquote callouts.
 * See the full version in the docs.djust.org sketch. */
```

### Distribution

Two options:

1. **Ship `djust_theming/css/prose.css`**, opt in via `<link rel="stylesheet" href="/static/djust_theming/css/prose.css">`. Matches how `components.css` is distributed.
2. **Auto-include via `{% theme_head %}`** behind `DJUST_THEMING['include_prose'] = True`.

Option 1 is lower-risk and doesn't affect sites that don't use typography.

### PR plan

Willing to submit. Draft branch: `feat/ship-prose-css` on `djust-org/djust`. Full element-level rules already written and tested — I'd port them straight across. Adding `--color-brand-headings` as an optional pack token would let packs customize heading color vs body text (falls back to `--color-brand-text` via `var(..., fallback)`).

### Downstream win

docs.djust.org drops ~95 lines of custom CSS the moment this ships — we replace the block with a single `<link>`. Every other djust docs-like site would benefit identically.

### LINKED ISSUE #1011: .card / .alert in components.css should set overflow: hidden for clean rounded corners
### What's wrong

`djust_theming/static/djust_theming/css/components.css` gives `.card` and `.alert` a `border-radius` but leaves `overflow` at the default `visible`. When a child has a full-width background or border — the canonical `.card-header { border-bottom: 1px solid hsl(var(--border)); }` is the most visible case — it renders to the parent's content-box width, which in the curved-corner zone sticks out past the rounded edge.

Result: a visible 1–2 px notch at the card's top-left and top-right where the header's straight bottom border crosses the parent's rounded arc. Affects every theme pack, not palette-specific.

### Repro

```html
<div class="card">
  <div class="card-header">
    <h3 class="card-title">Settings</h3>
  </div>
  <div class="card-body">body</div>
</div>
```

Render against any pack. Zoom the top-left corner. The header's `border-bottom` visibly cuts through the rounded arc.

### Suggested fix

One line each on `.card` and `.alert` in `djust_theming/static/djust_theming/css/components.css`:

```css
.card,
.alert {
    overflow: hidden;
}
```

Possibly also `.modal` — same topology (header above body, rounded container).

### Workaround

docs.djust.org applied the same fix site-side earlier, then dropped it after we restructured our card markup to avoid the `.card-header` / `.card-body` split entirely. The bug is still latent for any site that uses the canonical header/body structure.

### Willing to submit PR

Yes, tiny. Branch: `fix/card-alert-overflow-hidden`.

### LINKED ISSUE #1012: djust_theming theme.css Cache-Control insufficient — Chrome ignores Vary: Cookie, pack switching silently fails
### What's wrong

`djust.theming.views.theme_css_view` sets `Cache-Control: max-age=3600, private` + `Vary: Cookie` + an ETag based on theme state. That's the textbook-correct pattern for per-cookie content, but Chrome's Vary: Cookie handling is well-known to be unreliable for dynamic per-cookie responses.

After `window.djustTheme.setPack('dracula')` sets the `djust_theme_pack` cookie and reloads, Chrome often serves the PRIOR pack's CSS from its own HTTP cache. The site keeps rendering with the old palette even though the cookie flipped and the server is (correctly) returning the new pack's CSS. User-visible symptom: **pack switching appears to do nothing**.

### Evidence

docs.djust.org hit this across consecutive sessions on Chrome 139 / macOS. Server-side everything was correct: cookie received, state resolved, CSS body for the new pack returned with 200. But rendered pages kept the stale palette until we cleared the browser cache manually.

### Workaround deployed on docs.djust.org

Append cache-busting params to the stylesheet URL: `<link href="/_theming/theme.css?p=djust&m=dark">`. Different pack or mode = different URL = guaranteed fresh fetch. Values come from a context processor that reads the same `ThemeManager.get_state()` the view itself reads, so the link URL and served body stay in lockstep. Source: [`docs_app/context_processors.py`](https://github.com/djust-org/docs.djust.org/blob/main/docs_app/context_processors.py) + [`templates/docs/base.html`](https://github.com/djust-org/docs.djust.org/blob/main/templates/docs/base.html).

### Suggested fix upstream

Preferably:

**Option A — ship a helper tag.** Add `{% theme_css_link %}` to `theme_tags.py`, emits a `<link>` with the cache-busting params baked in:

```python
@register.simple_tag(takes_context=True)
def theme_css_link(context):
    request = context.get("request")
    state = get_theme_manager(request).get_state()
    url = reverse("djust_theming:theme_css")
    pack = state.pack or ""
    mode = state.resolved_mode or "dark"
    return format_html(
        '<link rel="stylesheet" href="{}?p={}&m={}">',
        url, pack, mode,
    )
```

Sites drop in one tag, get correct behavior without knowing any of this.

**Option B — switch to `Cache-Control: no-store`.** Simpler, less defensible (loses caching on a ~33 KB file). But one line in the view.

Option A preserves caching when theme state doesn't change. That's the right trade-off.

### PR plan

Willing to submit. Branch: `fix/theme-css-cache-bust`. Changes in:
- `python/djust/theming/templatetags/theme_tags.py` — new tag
- `python/djust/theming/templates/djust_theming/theme_head.html` — use new tag in place of the current `<link>`
- `docs/guides/theming.md` or equivalent — note the cache-bust rationale

`theme_css_view` untouched (backward-compat).

### LINKED ISSUE #1013: djust.theming cookie priority overrides LIVEVIEW_CONFIG.theme.pack — cross-project bleed on localhost
### What's wrong

`djust.theming.manager.ThemeManager.get_state()` reads `djust_theme_pack` / `djust_theme_preset` / `djust_theme` from request cookies **before** consulting `LIVEVIEW_CONFIG['theme']`, the session, or config defaults. That's the right priority for a site with a user-facing theme switcher — user preference persists across reloads. But it creates a cross-site bleed on localhost, where every djust project answers on `localhost:80xx` and therefore shares a cookie jar.

### Repro

1. Visit `localhost:8001` (any djust site with a theme switcher). Pick pack `nyc_core` + preset `rose_pine`. The browser now has `djust_theme_pack=nyc_core` scoped to `localhost`.
2. Visit `localhost:8003` — a different project whose settings explicitly pin `LIVEVIEW_CONFIG['theme'] = {'pack': 'djust', 'default_mode': 'dark'}`.
3. docs.djust.org's pages render with the `nyc_core` palette — pink buttons, purple cards — despite the settings-level pin to `djust`.

Not a production issue (per-subdomain cookie scope isolates real sites). But it's a real dev trap — two devs reproducing issues on localhost will see different palettes unless they remember to clear cookies between projects.

### Evidence

Hit this while setting up docs.djust.org. Stale cookies from an earlier project on localhost pinned `nyc_core + rose_pine` despite `pack='djust'` in settings. Took a while to diagnose because the cookie header is silent — nothing in the rendered HTML or logs flagged the override. Filed in docs.djust.org's [UPSTREAM_TODO.md](https://github.com/djust-org/docs.djust.org/blob/main/UPSTREAM_TODO.md).

### Suggested fix

Either of:

**Option A — opt-in cookie override.** Read cookies only when `LIVEVIEW_CONFIG['theme']['enable_client_override']: True` (default `True` for back-compat). Sites that don't ship a user-facing switcher can set `False` and they become un-hijackable from cookies.

```python
# docs_project/settings.py
LIVEVIEW_CONFIG = {
    'theme': {
        'pack': 'djust',
        'default_mode': 'dark',
        'enable_client_override': False,  # force-pin
    },
}
```

**Option B — name cookies per project.** Key by `SESSION_COOKIE_NAME` or a new `DJUST_THEME_COOKIE_PREFIX` setting so localhost:8001 and localhost:8003 don't share a cookie jar.

Option A is simpler to explain and covers docs-site-style use cases cleanly. Option B covers them too but also fixes any legitimate multi-project localhost scenario.

### Workaround

Currently: manually clear `djust_theme*` cookies in DevTools when switching between djust projects during dev. No supported settings-level pin exists.

### Willing to submit PR

Yes. Branch: `fix/theming-cookie-scope-dev`. Changes:
1. Add `LIVEVIEW_CONFIG['theme']['enable_client_override']` (default `True` for back-compat).
2. In `theming/manager.py::get_state()`, skip cookie read when `False`.
3. Docs note in `docs/guides/theming.md` or equivalent.

johnrtipton:
## Summary

v0.8.2 drain — Group T (4 theming/CSS issues bundled). All 4 surfaced by docs.djust.org's testing.

- **Closes #1011** (bug, P1) — \`.card\` / \`.alert\` in components.css now set \`overflow: hidden\` so child borders stay inside the rounded corners. Affects every theme pack.
- **Closes #1012** (bug, P1) — new \`{% theme_css_link %}\` simple_tag in theme_tags.py. Emits \`<link href=\"/_theming/theme.css?p=<pack>&m=<mode>&r=<preset>\">\`. Different pack/mode = different URL = guaranteed fresh fetch even when Chrome ignores \`Vary: Cookie\`.
- **Closes #1009** (enhancement, P2) — new \`djust_theming/static/djust_theming/css/prose.css\` (91 lines). Pack-aware overrides for \`@tailwindcss/typography\` plugin's \`--tw-prose-*\` variables. Opt-in via \`prose-djust\` class alongside \`prose\`.
- **Closes #1013** (enhancement, P2) — \`ThemeManager.get_state()\` now respects \`LIVEVIEW_CONFIG['theme']['enable_client_override']\` (default \`True\` for back-compat). Set \`False\` to ignore \`djust_theme_*\` cookies — prevents localhost cross-project bleed.

CHANGELOG.md: 3 \"Added\" entries + 1 \"Fixed\" entry.

## Test plan

- [x] **7 new tests** in \`python/djust/tests/test_theming_v082_drain.py\` — covers all 4 issues with both source-level (CSS file content checks) and behavior-level (template render, ThemeManager.get_state()) assertions.
- [x] Broader theming suite: **1780 passed**, 1 skipped, 0 failures.
- [x] Pre-commit hooks ran clean (after one auto-format pass).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,42 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- **`{% theme_css_link %}` cache-busting helper tag (v0.8.2 drain — Group T, closes #1012)** —
+  Chrome's `Vary: Cookie` handling is unreliable for per-cookie dynamic CSS;
+  after a pack switch the browser often serves the prior pack's stylesheet
+  from its own HTTP cache and the page renders with stale palette. The new
+  `{% theme_css_link %}` tag in `djust.theming.templatetags.theme_tags`
+  emits `<link href="/_theming/theme.css?p=<pack>&m=<mode>&r=<preset>">`
+  with cache-busting query params derived from the same `ThemeManager.get_state()`
+  the view itself reads. Different pack/mode = different URL = guaranteed
+  fresh fetch. Usage: `<link rel="stylesheet" href="{% theme_css_link %}">`.
+
+- **`prose.css` for `@tailwindcss/typography` ↔ pack bridge (v0.8.2 drain — Group T, closes #1009)** —
+  new `djust_theming/static/djust_theming/css/prose.css` ships pack-aware
+  overrides for the typography plugin's `--tw-prose-*` variables. Opt in
+  by adding `prose-djust` alongside `prose` on your `<article>`. Reads
+  `--color-brand-*` tokens the active pack emits, so flipping packs at
+  runtime updates prose without a stylesheet swap. Includes both light-mode
+  and dark-mode invert variables. Pulled from docs.djust.org's reference
+  implementation. ~95 lines.
+
+- **`enable_client_override` flag for `LIVEVIEW_CONFIG['theme']` (v0.8.2 drain — Group T, closes #1013)** —
+  `ThemeManager.get_state()` reads `djust_theme_pack` / `djust_theme_preset`
+  cookies with priority over config defaults. Default behavior unchanged
+  (back-compat `True`). Sites without a user-facing theme switcher can set
+  `LIVEVIEW_CONFIG['theme']['enable_client_override']: False` to ignore
+  cookie reads — prevents cross-project bleed on localhost where multiple
+  djust apps share a cookie jar.
+
+### Fixed
+
+- **`.card` / `.alert` overflow:hidden for clean rounded corners (v0.8.2 drain — Group T, closes #1011)** —
+  `djust_theming/static/djust_theming/css/components.css` `.card` and
+  `.alert` selectors now set `overflow: hidden`. Without this, child
+  borders (e.g. `.card-header { border-bottom: ... }`) cross the
+  parent's rounded arc and produce a visible 1-2 px notch at the
+  corners. Affects every theme pack.
+
 - **`mount_batch` fallback for old-server compat (v0.8.1 reconcile drain — Group F, closes #1031)** —
   the `mount_batch` WebSocket frame was added in v0.6.0 (PR #970) for lazy-
   hydration efficiency. A v0.6.0+ client talking to a pre-v0.6.0 server
--- a/python/djust/tests/test_theming_v082_drain.py
+++ b/python/djust/tests/test_theming_v082_drain.py
@@ -1,0 +1,167 @@
+"""Regression tests for v0.8.2 drain — Group T (theming polish).
+
+Covers:
+- #1011 — `.card` / `.alert` overflow:hidden in components.css
+- #1012 — `{% theme_css_link %}` cache-busting tag
+- #1013 — `enable_client_override` flag in `LIVEVIEW_CONFIG['theme']`
+- #1009 — `prose.css` shipped for `@tailwindcss/typography` ↔ pack bridge
+"""
+
+from __future__ import annotations
+
+from pathlib import Path
+
+import pytest
+from django.template import Context, Template
+from django.test import RequestFactory
+
+
+# Path to the package's static dir so we can inspect shipped CSS files
+THEMING_PKG = Path(__file__).resolve().parent.parent / "theming"
+STATIC_CSS = THEMING_PKG / "static" / "djust_theming" / "css"
+
+
+# ---------------------------------------------------------------------------
+# #1011 — .card / .alert overflow:hidden
+# ---------------------------------------------------------------------------
+
+
+def test_components_css_card_has_overflow_hidden():
+    """The `.card` selector must set `overflow: hidden` so child borders
+    don't poke through the rounded corners (#1011)."""
+    text = (STATIC_CSS / "components.css").read_text()
+    # Find the .card { ... } block (the canonical one near "Card" comment).
+    start = text.find("\n.card {\n")
+    assert start > 0, "could not find .card selector"
+    end = text.find("\n}\n", start)
+    block = text[start:end]
+    assert "overflow: hidden" in block, f".card block missing overflow: hidden\n---\n{block}\n---"
+
+
+def test_components_css_alert_has_overflow_hidden():
+    """The `.alert` selector must set `overflow: hidden` (#1011)."""
+    text = (STATIC_CSS / "components.css").read_text()
+    start = text.find("\n.alert {\n")
+    assert start > 0, "could not find .alert selector"
+    end = text.find("\n}\n", start)
+    block = text[start:end]
+    assert "overflow: hidden" in block, f".alert block missing overflow: hidden\n---\n{block}\n---"
+
+
+# ---------------------------------------------------------------------------
+# #1012 — {% theme_css_link %} cache-busting tag
+# ---------------------------------------------------------------------------
+
+
+@pytest.mark.django_db
+def test_theme_css_link_emits_url_with_pack_and_mode_params():
+    """The {% theme_css_link %} tag emits a URL with `?p=<pack>&m=<mode>`
+    so different pack/mode = different URL → guaranteed fresh fetch
+    even when Chrome ignores Vary: Cookie (#1012)."""
+    rf = RequestFactory()
+    req = rf.get("/")
+    # No cookies set → falls back to defaults from settings
+    tmpl = Template("{% load theme_tags %}{% theme_css_link %}")
+    rendered = tmpl.render(Context({"request": req})).strip()
+    assert rendered.startswith("/_theming/theme.css") or "/theme.css" in rendered, (
+        f"unexpected URL: {rendered!r}"
+    )
+    # Either pack or mode (or both) should appear as a query param
+    assert "?" in rendered, (
+        f"theme_css_link should emit cache-busting query string; got {rendered!r}"
+    )
+    # At least one of the canonical keys
+    assert "p=" in rendered or "m=" in rendered, (
+        f"theme_css_link should include p= or m=; got {rendered!r}"
+    )
+
+
+@pytest.mark.django_db
+def test_theme_css_link_url_changes_with_pack_cookie():
+    """Different pack cookie → different URL. This is the actual cache-bust
+    contract — same URL for stale cache → guaranteed fresh fetch on switch."""
+    rf = RequestFactory()
+    tmpl = Template("{% load theme_tags %}{% theme_css_link %}")
+
+    # Render with one pack cookie
+    req1 = rf.get("/")
+    req1.COOKIES["djust_theme_pack"] = "djust"
+    url1 = tmpl.render(Context({"request": req1})).strip()
+
+    # Render with a different pack cookie
+    req2 = rf.get("/")
+    req2.COOKIES["djust_theme_pack"] = "nyc_core"
+    url2 = tmpl.render(Context({"request": req2})).strip()
+
+    assert url1 != url2, (
+        f"theme_css_link must produce different URLs for different packs; got {url1!r} and {url2!r}"
+    )
+
+
+# ---------------------------------------------------------------------------
+# #1013 — enable_client_override flag
+# ---------------------------------------------------------------------------
+
+
+@pytest.mark.django_db
+def test_get_state_reads_cookies_when_enable_client_override_true():
+    """Default behavior (enable_client_override=True): cookies override
+    config. Existing back-compat (#1013)."""
+    from djust.theming.manager import ThemeManager
+
+    rf = RequestFactory()
+    req = rf.get("/")
+    req.COOKIES["djust_theme_pack"] = "nyc_core"
+    req.session = {}
+
+    mgr = ThemeManager(request=req)
+    # Pin config defaults relevant to this test (others come from settings)
+    mgr.config = dict(mgr.config, pack="djust", enable_client_override=True)
+    state = mgr.get_state()
+    # Cookie wins over config (back-compat default)
+    assert state.pack == "nyc_core", (
+        f"expected cookie pack to win when enable_client_override=True; got pack={state.pack!r}"
+    )
+
+
+@pytest.mark.django_db
+def test_get_state_ignores_cookies_when_enable_client_override_false():
+    """When `enable_client_override=False`, config wins — sites without a
+    user-facing switcher won't get hijacked by stale localhost cookies (#1013)."""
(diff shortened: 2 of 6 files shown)


johnrtipton (commented):
## Review: PR #1074 — v0.8.2 theming polish (Group T drain)

**Verdict: APPROVE.** Clean drain bundle — 4 surgical fixes, 7 new tests all pass, 1780 broader theming tests green, no regressions.

### What was checked

**#1011 (`.card` / `.alert` overflow:hidden)**
- Both selectors fixed at the right offsets (lines ~25 and ~235).
- `.card-header` at line 240 has `border-bottom: 1px solid hsl(var(--border))` — exactly the failure mode the comment describes. Fix is on-point.
- Spot-check for missed siblings: `.modal-content` (validation modal at line 1488) already has `overflow: hidden`. No third selector with the `rounded-corner + child-bottom-border` shape needs the same fix in this file.

**#1012 (`{% theme_css_link %}` simple_tag)**
- Confirmed `ThemeState` dataclass at `manager.py:58–67` exposes `pack`, `preset`, `mode`, `resolved_mode` — attribute access in the tag is correct.
- `getattr(state, "resolved_mode", None) or getattr(state, "mode", None)` is appropriately defensive (both keys exist on the dataclass, but ordering preference for the resolved value is right).
- `NoReverseMatch` → literal `/_theming/theme.css` fallback is the correct sentinel for test-env URL-conf isolation.
- `mark_safe(full)` is safe here: query params are derived from `ThemeManager.get_state()` which validates pack/theme/preset against the registry (`registry.has_theme`, `registry.has_preset` at lines 330–335). Cookie input is filtered before reaching the URL string. No XSS surface.

**#1013 (`enable_client_override` flag)**
- Verified all 4 cookies (`djust_theme`, `djust_theme_preset`, `djust_theme_pack`, `djust_theme_layout`) are inside the same `if self.request and enable_client_override:` block at `manager.py:291–298`. Skip is atomic — no partial-skip bug.
- Default `bool(self.config.get(\"enable_client_override\", True))` preserves back-compat exactly.
- Debug log (`Cookies skipped — LIVEVIEW_CONFIG.theme.enable_client_override=False`) fires only on the skip branch via `elif self.request:` — diagnostically useful and won't spam the success path.

**#1009 (`prose.css`)**
- All 19 light-mode + 19 dark-mode `--tw-prose-*` invert variables present. Without the invert pair, `.dark .prose` ancestors would render washed out — important detail that's correctly handled.
- No hardcoded colors. Every value is a `var(--color-brand-*)` token with a sane `hsl(var(--border))` fallback for `--*-hr`, `--*-th-borders`, `--*-td-borders`. Single `color-mix` (line 81) uses `var(--color-brand-rust)` — pack-aware.
- The unlayered-on-purpose comment at the top is exactly right — typography plugin emits `.prose` unlayered, so `@layer` would lose specificity.

**CHANGELOG.md**
- 4 entries (3 Added + 1 Fixed) match the diff. Each links its issue and explains the *why*, not just the *what*. Good.

### Test results
- `test_theming_v082_drain.py`: 7 passed, 0.21s.
- Broader `test_theming_*.py`: 1780 passed, 1 skipped, 27 warnings (pre-existing `LiveViewForm` deprecation + Django 6.0 URLField warnings, unrelated).

### Nits (non-blocking)
- `theme_css_link` could `urlencode` the params for paranoid completeness, but registry validation already constrains them to known-safe identifiers, so manual `f\"p={pack}\"` is fine.
- `prose.css` `--tw-prose-pre-bg: var(--color-brand-bg)` for both light and dark relies on the pack scope flipping `--color-brand-bg` itself — confirm packs that ship in v0.8.2 do this. (Not blocking; if a pack regresses, it's a pack issue, not this file's.)

Ready to merge.

johnrtipton:
## Per-PR Retro — #1074 (v0.8.2 Group T — theming polish)

**Date**: 2026-04-25
**Scope**: 4 theming/CSS fixes bundled per `--group` mode (#1009 prose.css, #1011 overflow:hidden, #1012 theme_css_link, #1013 enable_client_override). All from docs.djust.org's testing.
**Verdict**: 5/5 — clean Stage 11 APPROVE with 0 🔴/🟡, single fix-pass during commit (ruff auto-format).

### What Worked Well

- **Pulling reference impl from a downstream consumer.** `prose.css` (#1009) was lifted nearly verbatim from docs.djust.org's `static/css/input.css` — the ↔ pack bridge had already been battle-tested against three theme packs in production. Time-to-implement was a fraction of "design from scratch". **Pattern: when an issue cites a downstream consumer's solution, lift the reference impl FIRST, generalize SECOND.**
- **`getattr(state, ..., None)` on the ThemeState dataclass beat `state.get(...)`.** First implementation used `state.get("pack")` because the in-conversation memory said "ThemeState has dict-like API" — it's actually a dataclass. Caught at Stage 5 smoke-test (the new test failed with `AttributeError: 'ThemeState' object has no attribute 'get'`). Stage 5 smoke-test discipline (Action #145 from v0.8.1) caught the bug before commit.
- **Test-spec coverage matched the issue spec exactly.** Each test maps 1:1 to an issue's claim:
  - #1011 → 2 tests asserting CSS file has `overflow: hidden`
  - #1012 → 2 tests asserting URL has cache-busting params + URL-changes-with-pack contract
  - #1013 → 2 tests asserting cookie-read with both flag values
  - #1009 → 1 test asserting file shipped + key tokens present
  Stage 11 reviewer noted "tests would actually fail if implementation regressed" — that's the doc-claim-verbatim TDD discipline (Action #124).
- **`mark_safe` audit at Stage 11.** Reviewer specifically checked whether the URL string passing through `mark_safe` introduces any XSS surface. Confirmed: cookie inputs are filtered through `registry.has_theme/has_preset` validation in `get_state()` before reaching the URL — XSS-safe.

### What Didn't Work

- **First commit attempt failed pre-commit's ruff auto-format.** `import os` was unused (left over from an early test draft). One auto-fix pass cleared it. No-op cosmetically but a trip through the pre-commit cycle. Stage 5 smoke-test wouldn't have caught this — ruff F401 is post-commit. Pattern: keep test imports tight from the start, even in draft files.

### Review Stats

| Metric | Value |
|---|---|
| Tests added | 7 (test_theming_v082_drain.py) |
| Production LOC | +273 (CSS + Python) + 91 (prose.css new) = +364 / -2 |
| 🔴 / 🟡 findings | 0 / 0 |
| Pre-commit attempts | 2 (1 ruff F401 auto-fix) |
| CI retries | 0 |
| Quality rating | 5/5

### Stage 3.5 classification

For the eventual v0.8.2 milestone retro:

- **Action taken**: `Closed — Closes #1009 (prose.css for @tailwindcss/typography ↔ pack bridge), #1011 (.card/.alert overflow:hidden in components.css), #1012 ({% theme_css_link %} cache-busting tag in theme_tags.py), #1013 (enable_client_override flag in ThemeManager.get_state). All 4 shipped in this PR.`

`closed` shape — Stage 3.5 gate clean.

### Pattern: lift-from-downstream FIRST

Add to authoring discipline: **when an issue cites a downstream consumer's working solution (e.g. "docs.djust.org wrote the bridge in its own input.css"), lift the reference impl verbatim FIRST, generalize SECOND.** Skips a design-from-scratch phase that risks producing an incompatible variant. PR #1074's `prose.css` is the canonical example — 91 lines lifted near-verbatim, generalized only via the `prose-djust` opt-in class wrapper.

### Milestone context

v0.8.2 drain — Group T of 2 (1 down, 1 to go). Solo (#1010 docs link cleanup) remains.

PR #1092, opened 2026-04-26
feat(client): View Transitions API integration in applyPatches (closes Quick Win #23)

johnrtipton:
## Summary

- Opt-in browser-native cross-fade animation between WebSocket VDOM patches via `document.startViewTransition()`.
- Activates only when **both** conditions hold:
  1. Browser supports `document.startViewTransition()` (Chrome 111+, Edge 111+, Safari 18+; Firefox in active development).
  2. Body element carries `<body dj-view-transitions>`.
- Closes **Quick Win #23** + **Phoenix LV Parity Tracker** *View Transitions API* row in `ROADMAP.md`.

## Why opt-in

Per-WS-update animation would be jarring on rapid updates (keystrokes, streaming markdown, cursor presence). The body-attribute opt-in matches djust's existing `dj-cloak` / `dj-prefetch` convention — explicit, discoverable, removable per page.

Per-element control via the standard `view-transition-name` CSS property is fully supported and orthogonal to the opt-in:

```css
.hero-image { view-transition-name: hero-image; }
```

When the opt-in is active, any element with `view-transition-name` gets shared-element animation between states. Otherwise the browser-default cross-fade fires.

## Implementation

`python/djust/static/djust/src/12-vdom-patch.js`:

- **5-line check** at the top of `applyPatches` wraps the patch loop in `document.startViewTransition()` when both conditions hold.
- Inner patch logic **extracted as `_applyPatchesInner`** so the wrap and fallback paths share one implementation (no duplication of focus save/restore, 4-phase sort, autofocus handling).
- **Graceful degradation**: if API missing OR attribute absent, falls through to direct apply with no behavior change.

```js
if (
    typeof document !== 'undefined' &&
    typeof document.startViewTransition === 'function' &&
    document.body &&
    document.body.hasAttribute('dj-view-transitions')
) {
    let innerResult = true;
    document.startViewTransition(() => {
        innerResult = _applyPatchesInner(patches, rootEl);
    });
    return innerResult;
}
return _applyPatchesInner(patches, rootEl);
```

The spec guarantees the callback runs synchronously before `startViewTransition()` returns (animation phase is deferred), so the boolean return value is observable without awaiting `transition.finished`.

## Tests

5 JSDOM cases in `tests/js/view-transitions.test.js`:

- `wraps applyPatches in startViewTransition when body has dj-view-transitions AND API is supported`
- `does NOT wrap when body lacks dj-view-transitions, even if API is supported`
- `does NOT wrap when API is unsupported, even if body has the opt-in attribute` (no throw)
- `still applies patches successfully when wrapped` (inner contract preserved)
- `opting in dynamically (post-mount) is honored on subsequent patches` — `document.body.setAttribute('dj-view-transitions', '')` is honored on the next `applyPatches` call (no startup latch — the check happens at every call)

## Bundled assets

`make build-js` regenerated. All hand-written changes are confined to `12-vdom-patch.js` (~50 lines including JSDoc); bundle deltas are mechanical regeneration:

- `client.js` (bundled source)
- `client.min.js` (minified)
- `client.min.js.gz` / `client.min.js.br` (compressed)
- `client.min.js.map` (sourcemap)

## ROADMAP

- Quick Win #23 → ✅ Shipped
- Phoenix LV Parity Tracker `View Transitions API` row → ✅ Shipped (was v0.5.0 milestone target — 3 milestones late but clean scope; deferred items audited in PR #1087's staleness sweep)

## Test plan

- [x] `npx vitest run tests/js/view-transitions.test.js` — 5 passed
- [x] `make build-js` — bundle regenerates cleanly
- [x] Pre-commit hooks — all green (build-client, npm test, eslint, secrets, CHANGELOG-test-counts)


johnrtipton:
Withdrawing per Stage 11 review of the wrap-and-capture-return pattern. The sync-callback claim contradicts the W3C spec (callback runs in a microtask after frame capture, not synchronously) and the optimistic `return innerResult` was always returning `true`. JSDOM tests passed because the stub lied.

Re-design captured in ADR-013 — see PR for the new design (Option A: make `applyPatches` async + `Promise<boolean>` return). Implementation will resume on `feat/view-transitions-api-v2` after ADR review.

Issue #1098, opened 2026-04-26
tech-debt: handleMessage interleaving after async migration (PR-A follow-up)


johnrtipton:
After the async migration in #PR-A (`feat/view-transitions-api-v2`), `LiveViewWebSocket.handleMessage` and `LiveViewSSE.handleMessage` are now `async`. The `onmessage` arrow callback fires-and-forgets each one with `.catch()`. With multiple inbound frames arriving in quick succession, two `handleMessage` invocations can interleave their `await handleServerResponse` calls.

The `_pendingEventRefs` / `_tickBuffer` flush logic at `03-websocket.js:561-568` reads `_pendingEventRefs.size` AFTER an `await`, so an in-flight second message could mutate the set between the check and the flush — leading to a buffered tick draining out-of-order versus its expected version, or being applied twice.

The previous synchronous code guaranteed handlers ran to completion before `onmessage` fired again. Same concern on SSE at `03b-sse.js:77`.

**Suggested fix**: per-transport message queue (`await this._inflight` chain) before the next PR (PR-B) widens the await window with the View Transitions wrap.

**Source**: PR-A Stage 8 (Security Check) finding 1 — bumped from 🟡 to a real follow-up because PR-B's View Transitions wrap will widen the at-risk window meaningfully.

**Audit trail**: this issue must be resolved or explicitly accepted before PR-B (View Transitions wrap) lands.

PR #1099, opened 2026-04-26
refactor(vdom): make applyPatches async — View Transitions API foundation (PR-A)

johnrtipton:
## Summary

Foundational refactor preparing for View Transitions API integration per ADR-013. Converts `applyPatches(patches, rootEl)` from synchronous `boolean` to `async function ... -> Promise<boolean>`. The patch-loop body itself is byte-identical — this is a **signature-only migration**.

PR-B (View Transitions wrap) will layer on top of this foundation without further signature changes.

## Why split

The original PR #1092 attempt tried to ship the async refactor + the `startViewTransition()` wrap together. Stage 11 review found the wrap shape was wrong (the W3C spec runs the callback in a microtask, not synchronously). PR-A here is the corrected foundation: get the signature right, ship it, then layer the wrap in PR-B against a sound base.

## Caller chain migrated

- `02-response-handler.js`: `handleServerResponse` → `async`; `await applyPatches`
- `03-websocket.js`: `LiveViewWebSocket.handleMessage` → `async`; `setTimeout` jitter + direct fire-and-forget invocations wrapped in `.catch()`; `await handleChildUpdate` / `handleStickyUpdate`
- `03b-sse.js`: `LiveViewSSE.handleMessage` → `async`; `EventSource.onmessage` arrow callback (which can't itself be `async`) wraps the call in `.catch()`
- `11-event-handler.js`: `await applyPatches` + `await handleServerResponse` in event-cache replay path
- `45-child-view.js`: `_applyScopedPatches`, `handleChildUpdate`, `handleStickyUpdate` all `async`

## JSDOM async-hoisting fix

Async function declarations don't hoist to the `eval` host scope the way sync function declarations do. Added explicit `globalThis.djust.applyPatches = applyPatches` at the end of `12-vdom-patch.js` so test files can address it as `dom.window.djust.applyPatches`. New public-surface export documented in CHANGELOG.

## Test surface migrated

7 JS test files updated: `batch_insert_before_remove`, `event_sequencing`, `sse`, `tab_switch_real_repro`, `vdom-autofocus`, `vdom_patch_errors`, `vdom_recovery`. Added `await` to API calls; `async` to test callbacks.

## Pipeline review verdicts

- **Stage 7 (Self-Review)**: REVIEW_PASSED — zero 🔴, all six 🟡s addressed (un-awaited test calls fixed; CHANGELOG scope verified; mock staleness pre-existing) or accepted (style nit on redundant `await` in `45-child-view.js:114`).
- **Stage 8 (Security Check)**: SECURITY_PASSED — zero 🔴. One 🟡 (handleMessage interleaving widening with PR-B's `await` window) filed as #1098 — must be resolved before PR-B lands.
- **Tests**: 1396 JS pass, 4230 Python pass.

## Test plan

- [ ] Behavior parity with sync signature confirmed by existing patch test suite (`vdom_patch_errors.test.js`, `vdom_recovery.test.js`, `tab_switch_real_repro.test.js`, `event_sequencing.test.js`, `batch_insert_before_remove.test.js`, `vdom-autofocus.test.js`, `sse.test.js`)
- [ ] CI green
- [ ] Browser smoke test: `make start` + click through demo project (counter, forms) — verify no regression in WebSocket message handling
- [ ] No external API change for view authors (Python views unchanged)

## Public surface change

`window.djust.applyPatches` returns `Promise<boolean>` instead of `boolean`. Documented under `### Changed` in CHANGELOG as a signature change.

## Follow-up

- #1098 — `handleMessage` interleaving from `await` widening (Stage 8 Security finding) — must resolve before PR-B
- PR-B (next) — View Transitions wrap with corrected microtask semantics

Refs: `docs/adr/013-view-transitions-api-integration.md` (Option A)


johnrtipton (commented):
## Stage 11 Code Review — PR #1099 (PR-A: applyPatches async signature)

### Verdict: REQUEST_CHANGES

One must-fix and one nit; the rest is solid. Trace details below.

### 🔴 Must-fix before merge

1. **`tests/js/vdom_recovery.test.js:81, 98, 117` — three `ws.handleMessage(...)` calls dropped without `await`** (and without `.catch()`). After the migration `handleMessage` returns a `Promise`, so any rejection from these three test bodies is now an unhandled-rejection. The `html_recovery` branch happens to be sync today (just `morphChildren` + `reinitAfterDOMUpdate`), so the test assertions immediately after still see the mutated DOM — but the promise is dropped. Two specific risks: (a) any future change to that branch (e.g., adding a single await) would silently make these three tests race; (b) Node 18+ surfaces unhandled rejections as test failures depending on jest config. Add `await`. The other 4 awaited handleMessage calls in the same file are already correct, so this is just a missed sweep.

### 🟡 Should-fix

2. **`tests/js/vdom-autofocus.test.js:144` — `expect(() => applyPatches(...)).not.toThrow()` is now a no-op assertion.** After the migration, errors become rejections, not synchronous throws; the wrapper only catches sync throws from the call site (which would only fire if `applyPatches` weren't a function at all). Replace with `await expect(dom.window.djust.applyPatches(patches)).resolves.not.toThrow()` or assert the resolved value. Same migration sweep gap as #1.

3. **handleMessage interleaving across the new `await` boundary** — flagged correctly in #1098, referenced in PR body, and bumped to a PR-B blocker. I confirmed the concrete sites (`03-websocket.js:561, 706`) read `_pendingEventRefs.size` after an `await`. Today the body of `applyPatches` is byte-identical sync code, but the `await` still introduces a microtask boundary so back-to-back inbound frames CAN interleave. The existing JS test suite doesn't reproduce a corruption case, so I won't block PR-A — but the hardening must land before PR-B widens this window. Issue tracking is correctly in place.

### 🟢 Nits

4. **`45-child-view.js:114` — `return await applyPatches(...)`** — a redundant `await` (the outer function is `async`, returning the promise directly is equivalent and avoids one extra microtask hop). Style only; the self-review accepted this. Fine to defer.

5. **CHANGELOG copy** — the entry says "No external API change for view authors — VDOM internals only," which is true for Python view authors but understates that `window.djust.applyPatches` is now newly exposed as a public symbol (was previously not on the namespace at all per the diff). Strictly speaking that's an addition, not a breakage, so the copy is technically correct; consider a one-line note that the symbol is freshly exposed for hook integration parity.

### Verifications passed

- 🟢 Patch-loop body in `12-vdom-patch.js` is **byte-identical** — only the function signature, JSDoc block, and the new `globalThis.djust.applyPatches` export changed.
- 🟢 Every `applyPatches` callsite in `python/djust/static/djust/src/` is `await`ed (verified: 02-response-handler.js:109, 11-event-handler.js:122, 45-child-view.js:114).
- 🟢 Every `handleServerResponse` callsite is `await`ed (03-websocket.js:552, 567, 709; 03b-sse.js:179, 185; 11-event-handler.js:193).
- 🟢 Both `onmessage` arrows correctly handle the async `handleMessage` via `.catch()` — `03-websocket.js:258, 263` (jitter setTimeout + direct path) and `03b-sse.js:77`.
- 🟢 `handleChildUpdate` and `handleStickyUpdate` calls in 03-websocket.js are `await`ed (lines 742, 770).
- 🟢 No unguarded `console.log` introduced; the three new lines are `console.error`, which is unguarded by project convention.
- 🟢 CHANGELOG has a `### Changed` entry that documents the signature change with rationale.
- 🟢 #1098 is correctly filed and referenced in PR body as a PR-B prerequisite.
- 🟢 6 of 7 modified test files correctly migrated to `await` + `async` callbacks.
- 🟢 Diff against `origin/main` (not stale local main) is **JS-only** as the brief required — 19 files, all under `static/djust/`, `tests/js/`, or CHANGELOG.

### Summary

Solid foundation refactor — signature change is clean, propagation is thorough, the one structural concern (interleaving) is correctly tracked off-PR. Two test-side migration misses that should be cleaned up before merge — both small. After fixing #1, this is approve-on-resubmit.

johnrtipton (commented):
## Stage 13 Re-Review (PR #1099 @ af6d9abe)

Verified all Stage 11 findings against `git show af6d9abe`:

🔴 **Must-fix #1 (vdom_recovery.test.js)** — **ADDRESSED**
- Line 81: `await ws.handleMessage({` ✓
- Line 98: `await ws.handleMessage({` ✓
- Line 117: `await ws.handleMessage({` ✓

🟡 **Should-fix #1 (vdom-autofocus.test.js:144)** — **ADDRESSED**
- Was: `expect(() => dom.window.djust.applyPatches(patches)).not.toThrow();`
- Now: `await expect(dom.window.djust.applyPatches(patches)).resolves.not.toThrow();`

🟢 **Nit #1 (45-child-view.js:114)** — **ADDRESSED**
- Was: `return await applyPatches(patches, rootEl);`
- Now: `return applyPatches(patches, rootEl);`

🟢 **Nit #2 (CHANGELOG.md)** — **ADDRESSED**
- New entry explicitly calls out "**Newly exposed public surface**: `window.djust.applyPatches` is now explicitly assigned via `globalThis.djust.applyPatches = applyPatches`" with hook-author guidance.

### Verification
- `npx vitest run tests/js/vdom_recovery.test.js tests/js/vdom-autofocus.test.js`: 2 files, 8 tests passed (723ms)
- `gh pr checks 1099`: Detect Security Hot Spots passed; remaining checks pending (re-triggered by latest push)

### Verdict

**RE_REVIEW_PASSED** — all Stage 11 findings addressed. Ready to merge once pending CI completes.

johnrtipton:
# Retro: PR #1099 — applyPatches async (View Transitions PR-A)

## Quality rating: 8/10

Foundation refactor landed clean — zero 🔴 in re-review, follow-up tracker filed for the one structural concern (#1098). Two-PR split was the right call.

## What worked

1. **Splitting PR-A (foundation) from PR-B (View Transitions wrap) on user input.** The first attempt (#1092) bundled them and the wrap pattern straddled sync/async wrong — caught only by retroactive Stage 11. Splitting the signature change into a verifiable foundation, then layering the wrap on a sound base in PR-B, removes the entire class of issue that broke #1092.

2. **Stage 11 + Stage 13 caught real bugs.** Stage 11 found 3 un-awaited `ws.handleMessage` calls in `vdom_recovery.test.js` (lines 81, 98, 117) — silent unhandled-rejection risk. Stage 13 verified the fix actually landed at those line numbers, not just claimed-fixed in the commit message. The "trust but verify" gate paid for itself.

3. **JSDOM async-hoisting workaround is reusable.** Async function declarations don't hoist to `eval` host scope the way sync ones do. Explicit `globalThis.djust.applyPatches` namespace export is the right escape hatch — and it's cleaner than the implicit hoisting we relied on before. Future async refactors should use this pattern from the start.

4. **Pre-commit hooks held.** Both commits passed all hooks first try (formatters, eslint, secrets, CHANGELOG test counts) — no commit-bounce loop, no `git commit` retries. The pre-commit-scoped-to-staged-files pattern in the skill paid off again.

## What didn't work

1. **Initial scope estimate undershot.** ADR-013 estimated "~5 caller sites" for the migration. Actual: 6 caller sites + 7 test files + bundle regen + JSDOM hoisting fix. Estimate was off by ~3x in test-file count. Test migration was the long pole, not the production code.

2. **First test-migration attempt left un-awaited calls in `vdom_recovery.test.js`.** The bulk regex pass that added `await` to `applyPatches`/`handleServerResponse` calls didn't catch `ws.handleMessage(...)` because the regex only targeted top-level functions, not method invocations. Caught by Stage 11 — but should have been caught by self-review.

## What to improve

1. **For multi-file regex migrations, validate completeness with a follow-up grep.** After the first regex pass, run something like: `grep -nE '(^|[^t])(handleMessage|applyPatches|handleServerResponse)\\(' tests/js/*.js` and visually scan for hits inside `async` test bodies that don't have `await`. Cheap discipline that closes the gap regex misses.
   **Action taken**: tracked in Action Tracker (open) — generic enough to canonicalize as a CLAUDE.md note for the next async migration.

2. **Estimate test-migration scope by counting test files that import the changed module, not by intuition.** ADR-013 said ~5 sites; actual was 13 (6 production + 7 test). Next ADR for an async-style migration should run `grep -lr "applyPatches\\|handleMessage\\|handleServerResponse" tests/js/` upfront and put the count in the plan.
   **Action taken**: tracked in Action Tracker (open) — generic ADR-quality lesson.

3. **#1098 must resolve before PR-B.** PR-B widens the await window (the View Transitions wrap inside `applyPatches` adds another await layer). Multi-frame interleaving across `await` boundaries is already a latent risk today — PR-B will widen the at-risk window. Sequence: fix #1098 → ship PR-B.
   **Action taken**: filed as #1098 with explicit "PR-B blocker" note in the issue body.

- CLAUDE.md (or pipeline checklist): "After bulk regex migration, run a completeness-grep on the targeted patterns and visually scan for hits inside async bodies that lack `await`."
- ADR scoping guidance: "For async-style migrations, count test files that import/eval the module via `grep -lr` upfront — test scope is often 2-3× the production scope."

## Review stats

- 🔴 Findings (Stage 11): 1 (un-awaited handleMessage in 3 test sites — same root cause)
- 🟡 Findings (Stage 11): 2 (no-op assert, interleaving — latter filed as #1098)
- 🟢 Findings (Stage 11): 2 (redundant await, CHANGELOG namespace note)
- All addressed before merge: yes
- Tests: 1396 JS pass, 4230 Python pass
- Time from PR open to merge: ~30 min including review address

PR #1102, opened 2026-04-26
feat(wizard): wizard_input_event + dom_event kwarg for as_live_field (closes #1095)

### LINKED ISSUE #1095: feat: as_live_field() / WizardMixin — add option for dj-input event binding
## Root cause

`as_live_field()` in `WizardMixin` always generates `dj-change="<handler>"` on text inputs. `dj-change` fires on the `change` DOM event (blur/unfocus).

## Symptom

In multi-step wizards where fields are pre-filled (e.g. demo autofill), a user who edits a text field and immediately clicks the Next button — without first tabbing away — never triggers `dj-change`. Their edit is silently discarded, and the wizard review page shows the stale pre-filled value. This is a footgun for any wizard that pre-fills fields.

## Workaround (current)

We post-process `field_html` with a regex injection method (`_inject_dj_input`) that adds `dj-input="validate_field"` to text/textarea inputs after the fact. This is fragile and shouldn't be required.

## Requested change

Add an `event` kwarg to `as_live_field()` so callers can specify `dj-input` instead of (or in addition to) `dj-change`:

```python
# Single event
as_live_field("first_name", event="dj-input")

# Both events
as_live_field("first_name", event=["dj-change", "dj-input"])
```

Or add a class-level config on `WizardMixin`:

```python
class MyWizard(WizardMixin, ...):
    wizard_input_event = "dj-input"  # default: "dj-change"
```

`dj-input` fires on every keystroke (300ms debounce). This ensures edited values are captured regardless of whether the user explicitly blurs before clicking Next.

johnrtipton:
## Summary

Closes #1095. `WizardMixin.as_live_field()` previously emitted `dj-change` unconditionally on text/textarea/select/checkbox/radio inputs. `dj-change` fires only on blur, so a user who edits a pre-filled field and clicks **Next** without tabbing away has their edit silently discarded — the wizard review page shows the stale pre-filled value.

Replaces the regex post-process workaround downstream consumers (nyc-claims PR #185) had to maintain.

## API

**Class-level default** (most common case):

```python
class MyWizard(WizardMixin, LiveView):
    wizard_input_event = "dj-input"   # default: "dj-change"
```

**Per-call override**:

```python
view.as_live_field("email", dom_event="dj-input")
```

`dj-input` fires on every keystroke (300ms client-side debounce already in `09-event-binding.js`), so edits land regardless of whether the user blurs first.

## Implementation

- `python/djust/wizard.py` — class attribute + `kwargs.setdefault("dom_event", self.wizard_input_event)` before forwarding to adapter
- `python/djust/frameworks.py` — 5 sites: `_render_input` (text/textarea/select), `_render_checkbox`, `_render_radio` — read `kwargs.get("dom_event", "dj-change")` instead of hardcoded `"dj-change"`
- `python/tests/test_wizard_input_event.py` — 11 new regression cases

Default behavior unchanged. Strictly additive opt-in.

## Test plan

- [x] 11 new tests in `test_wizard_input_event.py` cover default behavior, `dom_event=` override, class-attr propagation, and per-call wins-over-class
- [x] Existing `test_live_field_widget_attrs.py` (21 cases) still passes — widget attrs preserved
- [x] Full Python suite: 2060 pass
- [ ] CI green

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,12 +44,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   `_render_checkbox`, `_render_radio` — read `kwargs.get("dom_event",
   "dj-change")` instead of hardcoding `"dj-change"`).
 
-  11 regression cases in `python/tests/test_wizard_input_event.py`
+  14 regression cases in `python/tests/test_wizard_input_event.py`
   cover: default class attribute is `"dj-change"`; default rendering on
-  text/textarea/select/checkbox emits `dj-change`; per-call
+  text/textarea/select/checkbox/radio emits `dj-change`; per-call
   `dom_event="dj-input"` swaps to `dj-input` and removes `dj-change`;
   `wizard_input_event = "dj-input"` flows through `as_live_field()`;
-  per-call kwarg overrides class attribute.
+  per-call kwarg overrides class attribute; `dom_event=None` coalesces
+  to the class attr instead of producing `attrs[None]`.
 
 - **`self.defer(callback, *args, **kwargs)` — Phoenix-style post-render
   callback scheduling** — new method on `AsyncWorkMixin` (and therefore on
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,48 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- **`WizardMixin.wizard_input_event` class attribute + `dom_event` kwarg on
+  `as_live_field()` — configurable DOM event for live-field validation
+  binding (closes #1095)** — `WizardMixin.as_live_field()` previously emitted
+  `dj-change="<handler>"` unconditionally on text/textarea/select/checkbox/
+  radio inputs. `dj-change` fires only on blur, so a user who edits a
+  pre-filled field and clicks Next without tabbing away has their edit
+  silently discarded. Wizards with autofill or pre-filled-from-database
+  fields hit this routinely.
+
+  New API: a class-level default and a per-call override.
+
+  Class default::
+
+      class MyWizard(WizardMixin, LiveView):
+          wizard_input_event = "dj-input"   # default: "dj-change"
+
+  Per-call override::
+
+      view.as_live_field("email", dom_event="dj-input")
+
+  ``dj-input`` fires on every keystroke (300ms client-side debounce
+  already in `09-event-binding.js`), so edits land regardless of whether
+  the user blurs first. Per-call kwarg wins over the class attribute.
+
+  Default behavior unchanged (``"dj-change"``), so this is a strictly
+  additive opt-in — existing wizards see no behavior change. Replaces
+  the regex post-process workaround that downstream consumers (e.g.
+  nyc-claims PR #185) had to maintain.
+
+  Files: `python/djust/wizard.py` (class attribute, `as_live_field`
+  forwards `dom_event` through `kwargs.setdefault`), `python/djust/
+  frameworks.py` (5 sites — `_render_input` text/textarea/select,
+  `_render_checkbox`, `_render_radio` — read `kwargs.get("dom_event",
+  "dj-change")` instead of hardcoding `"dj-change"`).
+
+  11 regression cases in `python/tests/test_wizard_input_event.py`
+  cover: default class attribute is `"dj-change"`; default rendering on
+  text/textarea/select/checkbox emits `dj-change`; per-call
+  `dom_event="dj-input"` swaps to `dj-input` and removes `dj-change`;
+  `wizard_input_event = "dj-input"` flows through `as_live_field()`;
+  per-call kwarg overrides class attribute.
+
 - **`self.defer(callback, *args, **kwargs)` — Phoenix-style post-render
   callback scheduling** — new method on `AsyncWorkMixin` (and therefore on
   every `LiveView`) that schedules a callback to run **once, after the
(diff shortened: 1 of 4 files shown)


johnrtipton (commented):
## Stage 11 — Code Review

### 🟢 Verified

**Default preserved (3 sites):** `frameworks.py:237` (text/textarea/select via `_render_input`), `:289` (checkbox), `:345` (radio). All use `kwargs.get("dom_event", "dj-change")` — legacy callers see no change. The PR's "5 sites" count refers to widget types; the actual write locations are 3 functions sharing the pattern. No stragglers — all `dj-change` literals remaining in `frameworks.py` are comments/docstrings.

**Forwarding chain is correct:** `wizard.py:175` `kwargs.setdefault("dom_event", self.wizard_input_event)` → `render_field(**kwargs)` (`frameworks.py:74-78`) → `_render_input` / `_render_checkbox` / `_render_radio`. Per-call kwarg wins over class attr (test `test_per_call_kwarg_overrides_class_attr` confirms).

**CHANGELOG:** Under `[Unreleased] / Added`, accurate, references files. ✓

**Tests:** All 88 local tests pass (`test_live_field_widget_attrs.py` 21 + `test_live_input_tag.py` 56 + new `test_wizard_input_event.py` 11). No auto-reject patterns (`mark_safe(f"...")`, `print()`, bare `except:`, f-string logging) in changed code.

### 🟡 Should-fix (non-blocking)

1. **Radio path lacks dedicated test coverage.** `_render_radio` (`frameworks.py:345`) is migrated but no test in `test_wizard_input_event.py` exercises it — only text/textarea/select/checkbox are covered. Mechanically identical to the others, so risk is low, but coverage should match the claim. Adding a `_Form` field with `forms.ChoiceField(widget=forms.RadioSelect)` and one assertion would close this.

2. **`dom_event=None` silently renders broken HTML.** `kwargs.setdefault("dom_event", self.wizard_input_event)` (`wizard.py:175`) does not overwrite an explicit `None`, and `kwargs.get("dom_event", "dj-change")` on the adapter side (`frameworks.py:237/289/345`) also returns `None`. Result: `attrs[None] = "validate_field"` → `_build_tag` (`frameworks.py:215`) renders `None="validate_field"` in the output. Not a crash, not a security issue, but the wizard validation binding silently breaks. Consider `if kwargs.get("dom_event") is None: kwargs["dom_event"] = self.wizard_input_event` in `as_live_field()`, or validate at the adapter site.

3. **Attr-key escape gap (theoretical).** `_build_tag` (`frameworks.py:215`) escapes attr **values** but not attr **keys**: `f'{k}="{escape(str(v))}"'`. Since `dom_event` is a developer-supplied kwarg (never user input), there is no XSS path today. Worth a comment near the kwarg surface noting the contract, or a runtime `re.fullmatch(r"[a-zA-Z][a-zA-Z0-9-]*", dom_event)` guard. Not blocking.

### 🟢 Nits

- PR description says "5 sites" but the diff has 3 hunks (3 write locations covering 5 widget types). Minor wording.

---

**COMMENT** — no 🔴 must-fix. Default behavior is preserved, the adapter forwarding chain is correct, CHANGELOG and tests are in order. The 🟡 items (radio test, `None` setdefault edge, attr-key escape) are worth addressing but do not block merge.

johnrtipton:
# Retro: PR #1102 — wizard_input_event (closes #1095)

## Quality rating: 8/10

Small, focused PR. Default behavior preserved, opt-in API for the dj-input case. Stage 11 caught the `dom_event=None` edge before it shipped.

## What worked

1. **Mechanical 5-site swap held up.** Replacing each hardcoded `attrs["dj-change"]` with `attrs[kwargs.get("dom_event", "dj-change")]` was uniform and easy to audit. Stage 11 verified all sites covered, no lingering hardcodes.

2. **Setdefault → explicit None coalesce.** Stage 11 caught a real edge: `kwargs.setdefault("dom_event", self.wizard_input_event)` doesn't overwrite a caller-passed `None`, which would propagate to `attrs[None]` and produce `None="validate_field"` in the HTML. Switched to explicit `if kwargs.get("dom_event") is None` check. Saved a 1-line bug from shipping.

3. **Class attribute + per-call kwarg pattern.** Standard Django-mixin shape — wizard authors set the wizard-wide default once, override per-field where needed. Discoverable in IDEs, testable in isolation.

## What didn't work

1. **Initial test pass missed RadioSelect.** I tested text/textarea/select/checkbox but skipped radio. The radio site (`frameworks.py:345`) is mechanically identical to the others, but identical-looking ≠ tested. Stage 11 flagged it; added 2 cases.

2. **Inline self-review and security review for small PRs.** Skipped subagent for Stages 7-8 because the change was 12 lines of mechanical attr-key swaps. This was OK here, but I missed both the None edge AND the radio gap that Stage 11 then caught. Inline review hits the obvious; subagent review (with fresh eyes) hits the non-obvious. The Stage 11 catch validates the skill's "never skip Stage 11" rule.

## What to improve

1. **For mechanical-replacement PRs, write tests for ALL replacement sites, not just the representative few.** Test surface should mirror the diff scope. If 5 sites change, 5 sites need coverage.
   **Action taken**: tracked in Action Tracker (open) — generic enough to canonicalize.

2. **For Python kwargs-forwarding, prefer `is None` coalescing over `setdefault`.** `setdefault` has the surprising property of not overwriting caller-supplied `None`. The two are NOT interchangeable.
   **Action taken**: tracked in Action Tracker (open) — worth a CLAUDE.md note.

- CLAUDE.md (Python patterns): "When a method forwards kwargs and a falsy default-from-class semantic is wanted, use `if kwargs.get('x') is None` rather than `setdefault('x', default)` — the latter does not overwrite a caller-passed `None`."
- PR checklist: "If the change touches N similar sites in a file, the test suite should hit all N (not just a subset that 'looks representative')."

## Review stats

- 🔴 Stage 11: 0
- 🟡 Stage 11: 3 (radio coverage, dom_event=None edge, _build_tag key-escape note)
- All 🟡 addressed before merge: yes (radio test added, None coalesce explicit, key-escape comment added inline)
- Tests: 14 pass; full suite 2060 pass

PR #1105, opened 2026-04-26
fix(checks): T012 partial-template marker + global suppress wiring (closes #1096)

### LINKED ISSUE #1096: bug: djust.T012 false positive on {% include %} partial templates
## Root cause

`manage.py check` emits `djust.T012` for any template that uses `dj-*` event directives but has no `dj-view` attribute — even when the template is an intentional partial designed to be `{% include %}`d from a LiveView root.

## Symptom

Step templates included via `{% include "intake/steps/vpd_step_vehicle.html" %}` produce:

```
?: (djust.T012) templates/intake/steps/vpd_step_vehicle.html -- template uses dj-* event directives but has no dj-view attribute.
```

These templates are fragments. `dj-view` belongs on the parent wizard base template, not on each included step. We have 15+ T012 warnings that clutter `manage.py check` output and make it harder to spot real issues.

## Requested change

One of the following:

1. **Suppression annotation** — support a `{# djust: partial #}` comment to opt a template out of T012.

2. **Smarter heuristic** — don't fire T012 on templates that have no root-level element (fragment templates by definition can't have `dj-view`).

3. **Global config key** — document and implement `DJUST_CONFIG = {'suppress_checks': ['T012']}` (T012's hint already mentions this key but it does not appear to be implemented or documented).

Any of these would let teams with large wizard UIs use partial templates without noisy false-positive warnings.

johnrtipton:
## Summary

Closes #1096. `djust.T012` (template uses `dj-*` event directives but missing `dj-view`) fired unconditionally on fragment templates intentionally `{% include %}`d from a parent LiveView root. Wizards with 15+ step partials produced a 15-warning wall on every `manage.py check`.

## Two opt-out paths

**Per-template marker** — case-insensitive, whitespace-flexible:
```html
{# djust:partial #}
<fieldset>
    <input dj-input="validate_field" name="vin" />
</fieldset>
```

**Global suppression** via `DJUST_CONFIG` (infrastructure existed; T012 just wasn't wired to it):
```python
DJUST_CONFIG = {"suppress_checks": ["T012"]}  # short or qualified id
```

T012's hint now mentions both opt-out paths.

## Implementation

- `python/djust/checks.py` — new `_DJ_PARTIAL_MARKER_RE`; T012 guard now reads partial marker AND `_is_check_suppressed("djust.T012")`. Component templates (`dj-component`) continue to bypass T012 — pre-existing behavior preserved.
- `python/tests/test_checks.py` — 5 new cases in `TestT012EventDirectivesWithoutView`.

## Test plan

- [x] 5 new test cases in `TestT012EventDirectivesWithoutView` cover partial marker (default + case-insensitive), global suppress (short id + qualified id), and hint text mentions both paths
- [x] Existing 3 T012 cases still pass — no regression in detect/with-view/component-template behavior
- [x] Full `test_checks.py` (180 cases) passes
- [ ] CI green

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -147,6 +147,36 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- **`djust.T012` false positive on `{% include %}` partial templates (closes
+  #1096)** — `T012` (template uses `dj-*` event directives but missing
+  `dj-view`) fired unconditionally for any template containing `dj-click`,
+  `dj-input`, etc., even when the file was an intentional fragment included
+  from a parent LiveView root. Wizards with 15+ step partials produced a
+  noisy 15-warning wall in `manage.py check`.
+
+  Two opt-out paths now silence T012 for legitimate fragments:
+
+  1. **Per-template marker**: add `{# djust:partial #}` (case-insensitive,
+     whitespace flexible) anywhere in the template. The marker is the right
+     choice when most fragments in a project don't need the check but a few
+     full-page templates do.
+  2. **Global suppression**: `DJUST_CONFIG = {"suppress_checks": ["T012"]}`
+     in `settings.py`. Right when the project never uses T012's intended
+     diagnostic (e.g. component-only architectures).
+
+  T012's hint now mentions both options. Component templates (`dj-component`
+  present) continue to bypass T012 as before — pre-existing behavior
+  unchanged.
+
+  Files: `python/djust/checks.py` (new `_DJ_PARTIAL_MARKER_RE`, T012 guard
+  reads partial marker AND `_is_check_suppressed("djust.T012")` —
+  previously the global suppression infrastructure existed but T012 wasn't
+  wired in). New cases added to `TestT012EventDirectivesWithoutView` in
+  `python/tests/test_checks.py` cover: partial marker silences T012;
+  case-insensitive matching; global suppression via short ID (`"T012"`)
+  and qualified ID (`"djust.T012"`); hint text mentions both opt-out
+  paths.
+
 - **`scripts/check-changelog-test-counts.py` regex missed `async def test_*`** —
   the test-counter pre-push hook's `PY_TEST_FN_RE` matched only `def test_*`,
   silently undercounting pytest-asyncio test files (any module-level
--- a/python/djust/checks.py
+++ b/python/djust/checks.py
@@ -1975,6 +1975,11 @@ _DJ_EVENT_DIRECTIVES_RE = re.compile(
     r"dj-(click|input|change|submit|blur|focus|keydown|keyup|mouseenter|mouseleave|window-\w+|document-\w+|click-away|shortcut)="
 )
 _DJ_COMPONENT_RE = re.compile(r"dj-component")
+# #1096 — opt-out marker for fragment templates that are intentionally
+# {% include %}d from a parent LiveView root. Fragment authors annotate
+# the file with `{# djust:partial #}` (case-insensitive, optional surrounding
+# whitespace) to silence T012 without introducing a global suppression.
+_DJ_PARTIAL_MARKER_RE = re.compile(r"\{#\s*djust\s*:\s*partial\s*#\}", re.IGNORECASE)
 _DEPRECATED_DATA_DJ_ID_RE = re.compile(r"""data-dj-id\s*=\s*["'][^"']*["']""")
 # A090 — scanner for {% djust_markdown %} (v0.7.0). Fires info-level once
 # per project when the tag is detected, confirming the Rust-side safe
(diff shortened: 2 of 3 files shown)


johnrtipton (commented):
## Stage 11 Code Review — PR #1105

Verified on `fix/t012-partial-suppress-1096`. 180/180 tests in `python/tests/test_checks.py` pass; 8/8 in `TestT012EventDirectivesWithoutView`.

### Verification

1. **Regex correctness** (`python/djust/checks.py:1979`) — `\{#\s*djust\s*:\s*partial\s*#\}` with `IGNORECASE` validated empirically:
   - `{# djust:partial #}` matches
   - `{#  djust : partial  #}` matches (whitespace-flexible)
   - `{#djust:partial#}` matches (no spaces)
   - `{# Djust: Partial #}` matches (case-insensitive)
   - `<!-- djust:partial -->` does NOT match (HTML form — acceptable)
   - `{% comment %}djust:partial{% endcomment %}` does NOT match (acceptable; Django comment is canonical)
   - `{# djust:partial-extra #}` does NOT match (bounded by `\s*#\}`)

2. **T012 guard scope** (`python/djust/checks.py:2168-2193`) — the partial marker check is inside T012's `if`-expression only. T013/T014/A070/A071 each have their own guards downstream, so the marker cannot accidentally suppress other checks. Structural, not test-asserted, but the surrounding code makes a regression here visually obvious in any future diff.

3. **Backwards compatibility preserved** — pre-existing 3 cases (`detects_events_without_view`, `passes_with_view`, `passes_for_component_template`) still pass. Component template bypass via `_DJ_COMPONENT_RE` retained as second clause in the new conjunction.

4. **Hint text** (`checks.py:2185-2188`) — mentions both opt-outs with correct syntax: backtick-quoted `{# djust:partial #}` and `DJUST_CONFIG = {'suppress_checks': ['T012']}`.

5. **`_is_check_suppressed` reuse** — helper at `checks.py:72` accepts both short (`T012`) and qualified (`djust.T012`) forms; both code paths covered by the two new global-suppress tests.

6. **No auto-reject patterns** — no `mark_safe`, no `print()`, no bare `except: pass`, no f-string logging. CHANGELOG updated (`CHANGELOG.md:150-178`).

### Findings

🔴 None.

🟡 None.

🟢 **Nit** — could add a future test asserting the marker doesn't suppress T013/T014 when present in the same file (defense-in-depth against refactors that might lift the marker check out of T012's `if`). Not blocking.

🟢 **Nit** — `{% verbatim %}` blocks containing the marker would still suppress T012, but this risk is shared by every other regex check in the file (pre-existing pattern, explicitly called out in PR scope).

PR scope is clean, tightly scoped to #1096, well-tested, documented. CHANGELOG entry is thorough.

**APPROVE** (posted as comment — self-authored PR cannot be self-approved via the GitHub review API).

— Stage 11, pipeline-run

johnrtipton:
# Retro: PR #1105 — T012 partial-template suppression (closes #1096)

## Quality rating: 9/10

Smallest of the three nyc-claims gap-fix PRs. Clean review (zero 🔴/🟡), minimal diff, behavior-preserving.

## What worked

1. **Existing infrastructure reuse.** `_is_check_suppressed()` was already implemented for 6 other checks; T012 just wasn't wired in. The fix was 1 line of guard logic + the partial-marker regex. The user-reported "T012 hint mentions this key but it does not appear to be implemented" was technically wrong — the helper existed, only the call site was missing. Reading the surrounding code first paid off.

2. **Two opt-out paths.** Per-template marker covers the "few partials in a project that mostly uses T012" case. Global `suppress_checks` covers the "project never uses the diagnostic at all" case. Different shapes for different needs. The hint mentions both so users can pick.

3. **Existing test class made test placement obvious.** `TestT012EventDirectivesWithoutView` already had 3 cases. Adding 5 more in the same class kept the file organized — no cargo-culted new test class.

## What didn't work

1. **CHANGELOG count phrase tripped the test-count drift hook.** "5 regression cases in `python/tests/test_checks.py`" tripped the check because that file already has 175 cases unrelated to this fix. Solution: reword to "New cases added to TestT012EventDirectivesWithoutView" — referring to the test class, not the file. The hook regex matches "N regression cases in path/to/file.py" exactly. Took 1 commit retry.

## What to improve

1. **CHANGELOG phrasing convention for additions to existing test files**: when adding tests to a file with pre-existing tests, reference the test CLASS or the FUNCTION COUNT NEW, not the total file. The hook's regex assumes the count refers to total tests in the named file — anything else trips drift.
   **Action taken**: tracked in Action Tracker (open) — generic enough to canonicalize as a CHANGELOG style note in CLAUDE.md or the PR checklist.

- CLAUDE.md (CHANGELOG conventions): "When adding tests to an existing test file, reference the test CLASS by name in CHANGELOG entries rather than 'N regression cases in <file>'. The pre-push test-count drift hook treats the latter as a claim about the entire file."

## Review stats

- 🔴 Stage 11: 0
- 🟡 Stage 11: 0
- 🟢 Stage 11: 2 nits (cross-check isolation test, verbatim-block edge — pre-existing pattern)
- All findings addressed before merge: N/A (zero blocking findings)
- Stage 13 (Re-Review): skipped per skip_if (zero 🔴/🟡)
- Tests: 8 in T012 class pass, 180 in checks.py pass, full suite green

PR #1107, opened 2026-04-26
perf(wizard): wizard_rendered_fields opt-in skips field_html for unused fields (closes #1097)

### LINKED ISSUE #1097: perf: WizardMixin renders field_html for all step fields, including fields not referenced in the template
## Root cause

`WizardMixin.get_context_data()` calls `self.as_live_field(fname)` for every field declared in the current step's form class, regardless of whether that field is actually referenced in the template.

## Symptom

On our VPD wizard, `VPDVehicleInfoForm` has ~20 fields, including conditional owner-info fields that are only shown when `is_vehicle_owner == "no"`. All 20 inputs are fully rendered on every LiveView event, even when only 5–8 are visible.

Benchmarks from our djust performance panel:
- Template rendering: **115ms** (threshold: 50ms)
- **47 VDOM patches** on autofill

This hurts both Python-side rendering time and VDOM diff size.

## Requested change

Consider lazy / on-demand `field_html` generation — only render `field_html` entries for fields that are actually referenced in the current template.

The JIT serializer already scans templates for used fields; applying the same optimization to `field_html` generation would reduce both server CPU and the number of patches sent to the client.

A simpler intermediate fix would be a `wizard_rendered_fields` class attribute that lets the developer explicitly declare which fields need `field_html`, skipping the rest:

```python
class VPDVehicleInfoView(WizardMixin, ...):
    wizard_rendered_fields = ["first_name", "last_name", "vehicle_vin"]
    # remaining fields only rendered when explicitly listed
```

This gives developers an escape hatch while a smarter template-scan approach is designed.

johnrtipton:
## Summary

Closes #1097. `WizardMixin.get_context_data()` unconditionally pre-rendered `field_html` for **every** field on the current step's form, regardless of whether the template referenced that field. Wizards with conditional fields (e.g. owner-info hidden behind `is_vehicle_owner == \"no\"`) paid the rendering cost on every event for fields nobody ever sees.

Reported impact (nyc-claims VPD wizard):
- Template render: **115ms** (threshold: 50ms)
- **47 VDOM patches** on autofill, most for invisible inputs

## API (default behavior unchanged)

**Class-level**:
```python
class VPDVehicleInfoView(WizardMixin, LiveView):
    wizard_rendered_fields = [\"first_name\", \"last_name\", \"vehicle_vin\"]
```

**Per-step override** (wins over class-level):
```python
wizard_steps = [
    {\"name\": \"vpd\", \"form_class\": VPDVehicleInfoForm,
     \"rendered_fields\": [\"first_name\", \"vehicle_vin\"]},
]
```

`None` (default) renders all fields — pre-#1097 behavior preserved.

## Important: only `field_html` is filtered

`form_data`, `form_required`, and `form_choices` are NOT filtered — all fields remain part of validation state. Only the (expensive) HTML rendering is opt-in skipped. Excluded field names produce no `field_html[fname]` entry; templates that reference them via `{{ field_html.unused|safe }}` render empty.

## Implementation

- `python/djust/wizard.py` — `wizard_rendered_fields` class attribute; `get_context_data()` reads `current_step.get(\"rendered_fields\", self.wizard_rendered_fields)` and only calls `as_live_field()` when the field passes the filter
- `python/tests/test_wizard_rendered_fields.py` — 8 new cases across 3 classes

## Future direction

A template-scan-driven version (similar to the JIT serializer's used-field detection) is the next iteration. This PR ships the explicit escape hatch first per the issue's \"intermediate fix\" request.

## Test plan

- [x] 8 new cases cover default-None renders all; class-attr filters; empty list renders nothing; unknown names silently ignored; per-step override; per-step wins over class-attr; form_data/required/choices still contain ALL fields
- [x] `test_wizard_input_event.py` (the #1095 sibling) still passes — no regression
- [x] Full Python suite: 2076 pass
- [ ] CI green

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,39 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- **`WizardMixin.wizard_rendered_fields` opt-in skips `field_html` rendering
+  for fields not in the list (closes #1097)** — `WizardMixin.get_context_data()`
+  unconditionally pre-rendered `field_html` for **every** field on the
+  current step's form, regardless of whether the template referenced that
+  field. Wizards with conditional fields (e.g. owner-info hidden behind
+  `is_vehicle_owner == "no"`) paid the rendering cost on every event for
+  fields nobody ever sees. Reported impact on the nyc-claims VPD wizard:
+  115ms template render (threshold: 50ms), 47 VDOM patches per autofill —
+  most for invisible inputs.
+
+  New API (default behavior unchanged — `None` renders all):
+
+  - **Class-level**: `wizard_rendered_fields = ["first_name", "vin", ...]`
+    on the wizard view limits `field_html` to that subset across every step.
+  - **Per-step override**: a step dict can include
+    `{"name": "...", "form_class": ..., "rendered_fields": [...]}` to scope
+    the filter to that step. Wins over the class-level default.
+
+  `form_data`, `form_required`, and `form_choices` are NOT filtered — all
+  fields remain part of validation/state. Only the (expensive) HTML rendering
+  is opt-in skipped. Excluded field names produce no `field_html[fname]`
+  entry; templates that reference them via `{{ field_html.unused|safe }}`
+  render empty (the dict-key absence is intentional and visible).
+
+  Files: `python/djust/wizard.py` (class attribute, per-step lookup +
+  filter in `get_context_data()`); `python/tests/test_wizard_rendered_fields.py`
+  with 8 cases in `DefaultRendersAllFieldsTest`,
+  `ClassAttributeFiltersTest`, `PerStepOverrideTest`.
+
+  Future direction: a smarter automatic template-scan (similar to the JIT
+  serializer's used-field detection) could drive this without explicit
+  developer wiring. This PR ships the explicit escape hatch first.
+
 - **`WizardMixin.wizard_input_event` class attribute + `dom_event` kwarg on
   `as_live_field()` — configurable DOM event for live-field validation
   binding (closes #1095)** — `WizardMixin.as_live_field()` previously emitted
--- a/python/djust/wizard.py
+++ b/python/djust/wizard.py
@@ -91,6 +91,15 @@ class WizardMixin:
     #: pre-filled and submitted without an intermediate blur. Closes #1095.
     wizard_input_event: str = "dj-change"
 
+    #: Optional opt-in to skip ``field_html`` rendering for fields not in this
+    #: list. Default ``None`` renders ALL fields in the current step's form
+    #: (pre-#1097 behavior). Set to a list of field names to render only those
+    #: — useful when a step's form has many fields but the template only
+    #: references a subset (e.g. conditional owner-info fields hidden behind
+    #: ``is_vehicle_owner == "no"``). Per-step overrides via the step dict's
+    #: ``"rendered_fields"`` key. Closes #1097.
+    wizard_rendered_fields: list | None = None
+
     @property
     def _steps(self) -> list:
         """Read wizard_steps from the CLASS definition, not the instance.
@@ -225,6 +234,9 @@ class WizardMixin:
             form_instance = (
                 form_class(data=current_step_data) if current_step_data else form_class()
             )
+            # #1097: opt-in field_html filter. Per-step "rendered_fields" wins
+            # over the class-level default. None = render all (legacy behavior).
+            rendered_filter = current_step.get("rendered_fields", self.wizard_rendered_fields)
             for fname, field in form_instance.fields.items():
                 val = current_step_data.get(fname, field.initial or "")
                 form_data[fname] = val if val is not None else ""
@@ -233,7 +245,8 @@ class WizardMixin:
                     form_choices[fname] = [
                         {"value": str(k), "label": str(v)} for k, v in field.choices if str(k)
                     ]
-                field_html[fname] = self.as_live_field(fname, event_name="validate_field")
+                if rendered_filter is None or fname in rendered_filter:
+                    field_html[fname] = self.as_live_field(fname, event_name="validate_field")
 
         # Expose choices as flat top-level vars too (e.g. borough_choices)
         # so templates can use either form_choices.borough or borough_choices.
--- a/python/djust/wizard.py
+++ b/python/djust/wizard.py
@@ -92,13 +92,18 @@ class WizardMixin:
     wizard_input_event: str = "dj-change"
 
     #: Optional opt-in to skip ``field_html`` rendering for fields not in this
-    #: list. Default ``None`` renders ALL fields in the current step's form
-    #: (pre-#1097 behavior). Set to a list of field names to render only those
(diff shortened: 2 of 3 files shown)


johnrtipton (commented):
## Stage 11 Code Review

### 🔴 Must-fix
None.

### 🟡 Should-fix

**1. Test gap: per-step `rendered_fields=[]` (render none) is not covered.**
`python/tests/test_wizard_rendered_fields.py` covers empty-list at the class-attr level (`test_empty_list_renders_no_fields`) but the per-step `[]` path is not exercised. Given the `dict.get(key, default)` shape returns `[]` (not the class fallback) when the key is present with value `[]`, this is a meaningful branch worth a test.

**2. Test gap: per-step `rendered_fields=None` semantic.**
`current_step.get("rendered_fields", self.wizard_rendered_fields)` returns `None` (not the class attr) when the step explicitly sets the key to `None` — meaning per-step `None` reverts to render-all, *overriding* a class-level filter. This is defensible (explicit None = "this step opts out of filtering") but undocumented and untested. Either add a test pinning the behavior or document it in the class-attr docstring.

**3. Test hygiene: class-level mutation in `_W.__init__`.**
`python/tests/test_wizard_rendered_fields.py:54-65` mutates `type(self).wizard_steps` from `__init__`. It works today only because every `_W()` call rewrites it, but the order-dependence is a smell — a future `_W()` that doesn't pass `per_step_rendered` will inherit whatever the previous test left on the class. Prefer instance-level override (`self.wizard_steps = [...]`) or a `setUp`/`tearDown` reset.

### 🟢 Nits

**4. Type annotation could be tighter.**
`wizard_rendered_fields: list | None = None` (`python/djust/wizard.py:101`) — the membership check `fname in rendered_filter` works for any iterable container (tuple, set, frozenset). If you intended only `list`, fine; if any-iterable is supported, `Iterable[str] | None` or `Sequence[str] | None` documents it. Not load-bearing.

**5. Docstring could note the surprising-to-developers behavior.**
The CHANGELOG explicitly calls out that templates referencing an excluded field via `{{ field_html.unused|safe }}` render empty. The class-attr docstring at `python/djust/wizard.py:94-100` is silent on this. A one-line note ("excluded fields produce no `field_html` entry; templates rendering them get empty output") would save the next reader a CHANGELOG hunt.

### Verified

- ✅ Default `None` preserves legacy behavior (line 248: `if rendered_filter is None or fname in rendered_filter`)
- ✅ Only `field_html[fname]` is gated; `form_data`, `form_required`, `form_choices` are NOT filtered (wizard.py:241-247 unchanged)
- ✅ Per-step wins over class-attr via `dict.get(key, default)` semantics
- ✅ Tests pass: 8/8 in 0.16s
- ✅ No auto-reject patterns (no `mark_safe(f'...')`, `print()`, bare except, f-string logging)
- ✅ CHANGELOG entry accurate and detailed
- ✅ Conventional commit prefix (`perf:`)

### Decision

**APPROVE** — the must-fix list is empty. Should-fix items (test gaps + test hygiene) are worth a follow-up commit on this branch but don't block merge. Performance escape-hatch is correctly scoped, tests prove the load-bearing behavior, and the API is cleanly opt-in.

APPROVE

johnrtipton:
# Retro: PR #1107 — wizard_rendered_fields opt-in (closes #1097)

## Quality rating: 8/10

Last of the 3-PR nyc-claims gap-fix drain. Smallest blast radius (single Python method modified), highest-leverage perf fix (115ms → expected 30-50ms when wired up downstream). Stage 11 caught real test-hygiene improvements.

## What worked

1. **Per-step + class-level dual API.** Class-level for the wizard-wide common case ("most of my wizard's fields don't need rendering"); per-step for the precise case ("step 3 only needs these 2 of its 8 fields"). The `current_step.get("rendered_fields", self.wizard_rendered_fields)` lookup gives both for free with one line.

2. **Filter only `field_html`, not `form_data`/`form_required`/`form_choices`.** Filtering the latter would break validation. Stage 11 explicitly verified this — easy to get wrong, important to get right.

3. **Test-hygiene refactor.** Original tests used `type(self).wizard_steps = [...]` in `__init__` to vary per-step config — mutation-via-shared-class. Stage 11 flagged it; refactor uses `type("...", (WizardMixin, _StubLiveView), {...})` to dynamically build a fresh subclass per call. Cleaner; fixed 8 → 12 cases.

## What didn't work

1. **Initial type annotation was too narrow.** `wizard_rendered_fields: list | None = None` would have rejected tuple/set filters — but the membership-check shape (`fname in rendered_filter`) accepts any iterable. Removed the annotation entirely; added 2 tests proving tuple + set work. Lesson: when the contract is "any iterable supporting `in`", don't annotate as `list | None`.

2. **Per-step `None` semantic was undefined initially.** Did `"rendered_fields": None` mean "fall through to class attr" or "render all"? The implementation made it "render all" by accident (None-filter branch). Now documented and tested as the deliberate semantic — explicit None at step level is the per-step opt-OUT from any class-level filter.

## What to improve

1. **For "filter shape" parameters, document and test the iterable contract.** Annotation `list | None` implies list-only; reality is "any iterable supporting `in`". For future similar parameters, write the docstring around the contract (`Iterable[str]` semantically) and test multiple iterable shapes.
   **Action taken**: tracked in Action Tracker (open) — generic enough for CLAUDE.md.

2. **For tests that need varying class-level state, use dynamic `type()` subclasses, not `type(self).x = ...` mutation.** `type(self).x = y` mutates a shared object; tests can leak state. `type("Name", (Base,), {"x": y})` returns a fresh subclass every call.
   **Action taken**: tracked in Action Tracker (open) — pattern worth canonicalizing.

- CLAUDE.md (Python patterns): "When a parameter accepts any iterable for membership checks, document as 'iterable of T' / `Iterable[T]` semantically — don't narrow with `list[T]`. Test at least one non-list shape (tuple or set) to lock in the contract."
- CLAUDE.md (testing patterns): "When a test fixture needs to vary class-level state, use `type(name, bases, dict)` to build a fresh subclass per call rather than mutating `type(self).x` — the latter leaks across tests via the shared class object."

## Review stats

- 🔴 Stage 11: 0
- 🟡 Stage 11: 3 (per-step empty-list test, per-step None semantic, test-hygiene class mutation)
- 🟢 Stage 11: 2 (type annotation, docstring)
- All findings addressed before merge: yes (12 tests, dynamic subclass refactor, doc-string + iterable nit)
- Stage 13 (Re-Review): inline-verified (all addressed)
- Tests: 12 in test_wizard_rendered_fields.py, 22 across both wizard test files, 2076 full suite

## Cross-PR learnings (3-PR nyc-claims gap-fix arc)

PRs #1102 (#1095), #1105 (#1096), #1107 (#1097) all shipped today as scoped, focused fixes for nyc-claims-reported framework gaps.

**Repeated pattern**: each PR was a small mechanical change with a Stage 11 catch on a non-obvious edge case (None coalescing, partial-marker scope, iterable contract / class mutation). The "small change → no Stage 11 catch" assumption is the failure mode; reality across all three: small change → 0-3 🟡 every time.

**Action #1098** (handleMessage interleaving) remains open as a PR-B blocker for the View Transitions wrap. The nyc-claims gap-fix arc is complete.

Issue #1100, opened 2026-04-26
tech-debt: completeness-grep for async-migration regex passes


johnrtipton:
**Source**: PR #1099 retro (https://github.com/djust-org/djust/pull/1099#issuecomment-4322596132)

After a bulk regex pass that adds `await` to migrated function calls, the regex commonly misses method invocations (`ws.handleMessage(...)`) when it's targeting top-level identifiers (`handleMessage(...)`). PR #1099 left 3 un-awaited `ws.handleMessage(...)` calls in `tests/js/vdom_recovery.test.js` (lines 81, 98, 117) that Stage 11 caught.

**Suggested check** to run after any async-migration regex pass:
```bash
grep -nE '(^|[^t])(handleMessage|applyPatches|handleServerResponse)\(' tests/js/*.js
```
Visually scan for hits inside `async` test bodies that don't have `await`.

**Promotion**: this is generic enough to canonicalize as a CLAUDE.md note (under 'Async migrations' or similar) so the next migration starts with this discipline.

Issue #1106, opened 2026-04-26
tech-debt: CHANGELOG conventions for additions to existing test files


johnrtipton:
**Source**: PR #1105 retro (https://github.com/djust-org/djust/pull/1105#issuecomment-4322703024)

The pre-push hook `scripts/check-changelog-test-counts.py` matches CHANGELOG claims of the form `N regression cases in path/to/file.py` and verifies N matches the total test count in that file. When a PR adds K new tests to a file that already has M tests, writing `K regression cases in path/to/file.py` trips drift (claims K, actual M+K).

**Convention**: when adding tests to an existing test file, write CHANGELOG entries that reference the test CLASS or the new-test count without binding it to the file:

- Wrong: `5 regression cases in python/tests/test_checks.py`
- Right: `5 new cases in TestT012EventDirectivesWithoutView (python/tests/test_checks.py)`

**Action**: add to CLAUDE.md (CHANGELOG conventions) or PR checklist.

johnrtipton:
Canonicalized in PR #1116 (CLAUDE.md / Process canonicalizations from PR retros section).

PR #1112, opened 2026-04-26
fix(transport): serialize handleMessage via _inflight Promise chain (closes #1098)

### LINKED ISSUE #1098: tech-debt: handleMessage interleaving after async migration (PR-A follow-up)

johnrtipton:
## Summary

Closes #1098 — the v0.8.5rc1 PR-A follow-up. PR-A made both transports' `handleMessage` async without serializing the inbound frame queue. Two adjacent inbound frames could fire-and-forget `_handleMessageImpl` concurrently and interleave their `await handleServerResponse` calls — racing on shared state like `_pendingEventRefs` / `_tickBuffer`.

Latent today; would have been meaningfully worse when **PR-B (View Transitions wrap)** widens the await window inside `applyPatches` itself.

## Fix

Per-transport `_inflight` Promise chain. Each public `handleMessage(data)` invocation chains onto the prior in-flight promise:

```javascript
handleMessage(data) {
    const prev = this._inflight || Promise.resolve();
    const next = prev
        .then(() => this._handleMessageImpl(data))
        .catch((err) => {
            console.error('[LiveView] handleMessage threw:', err);
        });
    this._inflight = next;
    return next;
}
```

- Existing async `handleMessage` body renamed to `_handleMessageImpl` (private)
- Both `LiveViewWebSocket` and `LiveViewSSE` apply the same pattern
- Errors are isolated per-frame via `.catch()` — a single bad frame doesn't poison the chain

## Test plan

- [x] New file `tests/js/handlemessage_serialization.test.js` (6 cases) — rapid-fire ordered drain, error isolation, both WS+SSE serialize
- [x] 4 existing test files updated to `await` handleMessage calls (`dj-cloak`, `hvr`, `sse-transport`, `sw_advanced`) — same un-awaited-call pattern Stage 11 caught on PR #1099
- [x] 1402 JS tests pass; 2080 Python tests pass
- [ ] CI green

## What this unblocks

PR-B (View Transitions wrap on top of PR-A foundation, ADR-013) is now safe to ship. Without this fix, PR-B would amplify the interleaving window inside `applyPatches` itself.

## Pipeline review note

The 4 caller-side test files needing `await` is exactly the regression class Stage 11 caught on PR #1099. Tracker #1100 (completeness-grep for async-migration regex passes) is the canonicalized version of this lesson — applying that grep would have caught the 4 files before the test run.


johnrtipton (commented):
# Stage 11 Code Review

Reviewed against the 8 verification points. All hard checks pass.

## Verified

1. **`_handleMessageImpl` body byte-identical to old `handleMessage`.** Confirmed via `diff` of the function bodies (with names normalized). No drift in either transport.
   - `python/djust/static/djust/src/03-websocket.js` — only delta is the new wrapper + rename
   - `python/djust/static/djust/src/03b-sse.js` — same

2. **No deadlock risk from internal self-calls.** Grepped both impl bodies for `this.handleMessage` / `this._handleMessageImpl` — zero matches inside either `_handleMessageImpl`. The 2 callsites at `03-websocket.js:258,263` and `03b-sse.js:77` are inside the `onmessage` arrow callback (external invocations), not internal self-calls.

3. **Error path doesn't leak rejection as `_inflight`.** The wrapper's `.catch((err) => console.error(...))` swallows errors and returns a resolved promise. The next `handleMessage` chains onto the resolved promise — chain continues past a bad frame. Test `'continues draining even when one message rejects'` (`handlemessage_serialization.test.js:123-150`) proves this exactly.

4. **Both transports tested independently.** New file covers both `LiveViewWebSocket` (3 cases) and `LiveViewSSE` (2 cases) plus 1 shared exposure check — separate `_inflight` per instance, no cross-pollution.

5. **All 4 caller-side test files correctly migrated.** Every `handleMessage(...)` in changed files now has `await`; their containing `it(...)` callbacks are marked `async`. Wider `grep` across `tests/js/` confirms no other un-awaited callers exist (`event_sequencing`, `vdom_recovery`, `sse.test.js`, etc. were already awaited from earlier work). The `await expect(...).resolves.not.toThrow()` form on `dj-cloak.test.js:121-123` correctly awaits.

6. **CHANGELOG entry** under `[Unreleased] / Fixed` — accurate, references PR-A / v0.8.5rc1, mentions both transports, calls out the test-migration gap and Tracker #1100 lesson.

7. **No auto-reject patterns.** No unguarded `console.log` in changed JS. `console.error` in the wrappers is fine (rule applies only to `console.log`); matches existing convention (12 prior `console.error` callsites in these files).

8. **Tests run green locally.** All 5 affected files: 43/43 pass. Full JS suite: 1402/1402 pass. CI: javascript-tests / playwright / analyze / security all green; python tiers still pending.

9. **Bundle artifacts updated.** `client.js`, `client.min.js{,.map,.gz,.br}` all rebuilt. Bundle delta ~50 B gzipped — well under the 2 KB-per-feature budget.

## Findings

🟢 **Nit** — `03-websocket.js:284-285` JSDoc says "Errors propagate through `.catch()` to preserve unhandled-rejection visibility." The wrapper's `.catch()` actually *swallows* errors (logs and returns undefined) — they don't propagate. The behavior is correct (chain continues past failures, which is what we want); just the wording is slightly misleading. The SSE docstring (`03b-sse.js:115-120`) is fine. Optional follow-up; not a blocker.

🟢 **Nit** — `03-websocket.js:258,263` and `03b-sse.js:77` retain an outer `.catch(...)` chained on `this.handleMessage(data)`. With the new wrapper's internal `.catch`, the outer one is now unreachable (wrapper returns a resolved promise even on impl rejection). Harmless defense-in-depth; the comment at `03-websocket.js:254-256` / `03b-sse.js:74-76` rationale still reads correctly. Optional cleanup; leaving it is also defensible.

🟢 **Nit (latent, out of scope)** — Latency-simulation path (`03-websocket.js:251-261`) wraps `handleMessage` in `setTimeout`, so under jittered debug runs two adjacent frames could enqueue out of order. Pre-existing, debug-only, not in scope here.

## Verdict

Zero 🔴, zero 🟡, three 🟢 nits.

The fix is the minimal correct shape: per-transport `_inflight` Promise chain, byte-identical impl body, error isolation per-frame, both transports treated symmetrically. Tests prove the three properties that matter (ordering, error isolation, drain semantics) and cover both transports.

**APPROVE** — ready to merge once python-tests tiers complete.

johnrtipton:
# Retro: PR #1112 — handleMessage interleaving fix (closes #1098)

## Quality rating: 9/10

Cleanest review of any of the recent PRs — Stage 11 returned APPROVE with zero 🔴 and zero 🟡, only three non-blocking 🟢 nits. Tight scope, clear shape, immediately actionable.

## What worked

1. **Naming the fix shape upfront in the state file's task_description.** The plan was specific: rename `handleMessage` → `_handleMessageImpl`, add new public `handleMessage` that does Promise-chain enqueue. Implementation was mechanical against that spec. Stage 4 (Planning) didn't need a separate agent because the shape was already locked.

2. **Foundation lessons from earlier PRs paid off.** The 4 caller-side test files (`dj-cloak`, `hvr`, `sse-transport`, `sw_advanced`) needing `await` is exactly the pattern Stage 11 caught on PR #1099 (PR-A). I knew to check for it this time and ran the full suite immediately after the implementation change. Caught all 8 failures in one batch, fixed in a single pass.

3. **Mirroring the wrapper between WS and SSE.** Same pattern, same `_inflight` semantics, both transports. The serialization invariant is a per-instance property — it doesn't matter if WS and SSE are running concurrently; they each have their own queue.

4. **Cheap CHANGELOG-hook fix.** When the test-count drift hook tripped (claiming "6 cases in 5 files" when only 1 of those files is new), rephrasing to "New regression file ... covers ... Caller-side test migration: 4 existing test files updated" cleared it without inflating the test count or shrinking the description. Direct application of #1106's lesson from PR #1105 retro.

## What didn't work

1. **First commit rejected by the test-count hook.** I phrased the CHANGELOG entry as "6 regression cases in `tests/js/handlemessage_serialization.test.js`, `tests/js/dj-cloak.test.js`, ..." which the hook interpreted as a claim of 6 total across all 5 files. Lost ~1 minute reformatting. The lesson is now twice-burned — should canonicalize: when a CHANGELOG entry references a NEW test file AND existing files updated for caller-migration, treat the new file's count separately from the migration delta. **Updates #1106** (CHANGELOG conventions for additions to existing test files) — this is now the second occurrence of that exact failure mode in 24 hours.

## What to improve

1. **For caller-migration PRs (signature change ripples through callers), the CHANGELOG should always have two test-related sentences: "New regression file …" and "Caller-side migration: N existing files updated to …". Never a single bullet that lumps both.** Avoids the hook's regex.
   **Action taken**: this lesson is the second occurrence; #1106 is upgraded from "should canonicalize" to "must canonicalize before next caller-migration PR".

2. **The 🟢 nit about redundant outer `.catch()` at `onmessage` callsites** is a real code-cleanup opportunity. The wrapper's internal `.catch()` swallows the rejection, so `this.handleMessage(data).catch(...)` at lines 258/263 of `03-websocket.js` is now dead code. Defense-in-depth, harmless, but adds reader confusion. Future PR-B work touches this file — pick it up there.

- CLAUDE.md (CHANGELOG conventions) — explicit "two-sentence rule" for caller-migration PRs (one for new tests, one for migration deltas), with a warning that the test-count drift hook treats `N regression cases in path/to/file.py` as a claim about the file's TOTAL count, not the new count. Reference both PR #1105 and PR #1112 retros as evidence.

## Review stats

- 🔴 Stage 11: 0
- 🟡 Stage 11: 0
- 🟢 Stage 11: 3 (JSDoc clarity, redundant outer catch, pre-existing latency-sim ordering edge case — all out of scope for PR-B)
- Stage 13 (Re-Review): skipped per `skip_if` (zero 🔴/🟡)
- All 🟢 worth tracking: outer-`.catch` cleanup → fold into PR-B
- Tests: 1402 JS + 2080 Python + new 6 cases all green
- CI: 12/12 green
- Time PR-open → merge: ~10 min (CI was the long pole)

## Cross-PR / cross-arc context

PR-A → PR (this) → PR-B is the View Transitions arc. PR-A's signature change introduced the latent interleaving regression that this PR fixes. Without this fix, PR-B would amplify the at-risk window inside `applyPatches` itself. Now safe to ship PR-B.

The arc demonstrates the value of Stage 8 security review's "🟡 follow-up tracker" mechanism — #1098 was filed at PR-A merge time (Stage 8 finding), explicitly labeled PR-B blocker, then resolved here before PR-B. The audit trail is unbroken.

Issue #1093, opened 2026-04-26
tech-debt: SSE-side test for legacy-view hasattr guard in _flush_deferred_to_sse


johnrtipton:
## Surfaced by Stage 13 re-review of PR #1091

Stage 13 review of #1091 ([approve verdict](https://github.com/djust-org/djust/pull/1091)) flagged one non-blocking gap:

> WS-side has \`test_flush_deferred_handles_view_without_drain_method\` (\`python/djust/tests/test_defer.py:236-244\`) for legacy views. SSE-side only covers \`view_instance=None\`. The \`hasattr\` guard at \`python/djust/sse.py:565\` is uncovered.

The WS-side guard at \`python/djust/websocket.py:535\` has a parallel test:

```python
async def test_flush_deferred_handles_view_without_drain_method():
    class _LegacyView:
        pass  # No _drain_deferred attribute
    await _FakeConsumer(_LegacyView())._flush_deferred()  # no-throw
```

The SSE-side equivalent is missing. Add to \`python/djust/tests/test_defer.py\`:

\`\`\`python
@pytest.mark.asyncio
async def test_sse_flush_deferred_handles_view_without_drain_method():
    \"\"\"SSE-side parity with \`test_flush_deferred_handles_view_without_drain_method\` —
    a view that doesn't subclass AsyncWorkMixin (no \`_drain_deferred\`)
    is silently skipped, mirroring the \`hasattr\` guard at \`sse.py:565\`.\"\"\"
    from djust.sse import _flush_deferred_to_sse

    class _LegacyView:
        pass

    await _flush_deferred_to_sse(_LegacyView())  # no-throw
\`\`\`

## Acceptance criteria

- [ ] Test added to \`python/djust/tests/test_defer.py\` in the SSE section
- [ ] Test passes against \`main\`
- [ ] CHANGELOG \`[Unreleased]\` claim of "19 regression cases" updated to "20"

Single-test addition, ~15 lines including docstring. Quick-win-sized.

## References

- PR #1091 — \`self.defer(callback)\` Phoenix-style post-render scheduling (merged)
- Stage 13 re-review verdict — approved with this as non-blocking suggestion
- \`python/djust/sse.py:565\` — the \`hasattr\` guard line
- \`python/djust/websocket.py:535\` — WS-side equivalent guard (covered by existing test)

🤖 Filed via Stage 13 re-review follow-up

johnrtipton:
[...]

Issue #1090, opened 2026-04-26
tech-debt: debug-log when |date / |time filter parse fails


johnrtipton:
## Surfaced by #1081

The #1081 investigation took 4 rounds of reopens to land at: the reporter's app was passing strings with embedded literal `\"` characters into `update_state` (likely from a `json.dumps(date.isoformat())` call in their custom serialization), the `|date` filter silently failed to parse them, and HTML-escape correctly converted the `\"` chars to `&quot;` — surfacing as `&quot;Apr 25, 2026&quot;` in the rendered DOM.

The whole arc would have collapsed to a 5-minute diagnosis if the `|date` filter had logged a single line at parse-failure time:

```
DEBUG djust.templates.filters: |date filter could not parse value '\"2026-04-25\"' (type=str) — passing through unchanged. If you expected formatting, check whether the value was JSON-encoded upstream.
```

## Proposed change

`crates/djust_templates/src/filters.rs` — the `|date` and `|time` arms currently:

```rust
\"date\" => {
    let format_str = arg.or(default_format).unwrap_or(\"N j, Y\");
    let datetime_str = value.to_string();
    match format_date(&datetime_str, format_str) {
        Ok(formatted) => Ok(Value::String(formatted)),
        Err(_) => Ok(value.clone()),  // ← silent passthrough
    }
}
```

The `Err(_) => Ok(value.clone())` branch should also emit a tracing event:

```rust
Err(parse_err) => {
    tracing::debug!(
        target: \"djust.templates.filters\",
        \"|date filter parse failed for value={:?} format={:?} err={:?}\",
        datetime_str, format_str, parse_err,
    );
    Ok(value.clone())
}
```

Same shape for the `|time` arm.

## Open questions for the implementer

- **Severity**: `debug` (not `warn` or `error`) — parse failure may be intentional in some flows (e.g., `|default` chained after `|date` to handle null/empty strings). `debug` keeps the noise off by default but makes the failure mode discoverable when the user enables djust DEBUG logging.
- **Tracing crate vs eprintln**: the Rust crate already uses `tracing::*` macros elsewhere — match that. Confirm via `grep -r tracing crates/djust_templates/src/` before adding.
- **Python-side log forwarding**: djust's Python code uses standard `logging` — the Rust `tracing` events need a subscriber that forwards to Python logging. If that's not already wired, the log line won't reach the user. Check `crates/djust/src/lib.rs` for any `tracing_subscriber::fmt()` or similar setup.
[...]

Issue #1089, opened 2026-04-26
tech-debt: Expand release wheel matrix to cp313 + cp314


johnrtipton:
## Surfaced by #1081

Long investigation of #1081 ended at: reporter's installed `_rust.cpython-314-darwin.so` is a source build, not a published wheel. PyPI for `djust==0.8.3rc1` ships only cp310 / cp311 / cp312. Reporter on Python 3.14 fell back to compiling the sdist at install time, producing an untested binary whose runtime behavior diverged from our CI-tested cp312 binary.

We can't tell users "your bug is in the binary you compiled" without first having shipped a tested cp314 binary. This issue files the work.

## Concrete change

`.github/workflows/release.yml` matrix is currently:

```yaml
python-version: ['3.10', '3.11', '3.12']
```

Expand to:

```yaml
python-version: ['3.10', '3.11', '3.12', '3.13', '3.14']
```

Plus the per-OS / per-arch matrix entries below the main one. Verify each wheel installs and the test suite passes against it (the `djust-monitor` smoke-test pipeline is the natural place to gate this — runs on every published wheel).

## Open questions for the implementer

- **PyO3 version**: confirm our pinned PyO3 supports cp314 ABI (releases 0.22+ support 3.13; 0.23+ should support 3.14, but verify against actual cp314 build).
- **GitHub Actions runner availability**: `actions/setup-python` lists supported versions per runner. Python 3.14 is current as of late 2025 — should be available on `ubuntu-latest` / `macos-latest` / `windows-latest` runners.
- **Drop cp310?**: Python 3.10 reaches EOL Oct 2026. Dropping it now would simplify the matrix; deferring to a future release is also fine. Out of scope for this issue.
- **abi3-stable wheel**: longer-term option — produce a single `cp310-abi3-...` wheel that works on cp310+. Reduces wheel count from ~12 to ~3 (one per OS). Bigger refactor; file as a separate consideration if pursued.

## Verification

After the matrix expansion:

1. Cut a fresh RC release (e.g., v0.8.4rc1).
2. Confirm PyPI listing includes cp313 + cp314 wheels for Linux x86_64, macOS x86_64+arm64, Windows.
3. Smoke-test `pip install djust==0.8.4rc1` on a fresh Python 3.14 venv → confirms binary wheel is selected, no source compilation triggered.
4. Run the 26-case regression test from #1086 against the cp314 wheel.

## References

- #1081 — surfacing user report (reopened multiple times)
[...]

Issue #1081, opened 2026-04-25
bug: `|date` filter on model DateField produces literal JSON-quoted string (`"Apr 25, 2026"`) in Rust-rendered output


johnrtipton:
## Observed behavior

djust 0.8.0rc1, Django 5.2.13.

Template:

```django
{{ claim.filed_date|date:"M d, Y" }}
```

where `claim` is a Django model instance passed in context and `filed_date` is a `DateField` with value `datetime.date(2026, 4, 25)`.

**Expected HTML**: `Apr 25, 2026`
**Actual HTML**: `&quot;Apr 25, 2026&quot;` (i.e. a literal double-quote-wrapped string, HTML-escaped)

Happens consistently for both `DateField` and `DateTimeField` values on model instances. Not observed for plain `str` values passed explicitly in context.

## Repro

```python
# views.py
class MyView(BaseLiveView):
    def get_context_data(self, **kwargs):
        ctx = super().get_context_data(**kwargs)
        ctx["claim"] = Claim.objects.first()  # has DateField `filed_date`
        return ctx
```

```django
{# template #}
{{ claim.filed_date|date:"M d, Y" }}      {# renders as `&quot;Apr 25, 2026&quot;` #}
{{ claim.filed_date }}                     {# renders as `2026-04-25` (no quotes — raw date __str__) #}
```

Applies to:
- `|date:"M d, Y"`
- `|date:"F j, Y"`
- `|date:"M d, Y H:i"` (DateTimeField — same pattern, quoted)

Also happens with `|default:"—"` chained on an empty date — the fallback `—` renders as `&quot;—&quot;`.

## Workaround (what I did locally)

Pre-compute the string in Python and pass it as a plain str:

```python
ctx["claim_filed_str"] = claim.filed_date.strftime("%b %-d, %Y") if claim.filed_date else "—"
```

```django
{{ claim_filed_str }}    {# renders correctly as `Apr 25, 2026` #}
```

## Guess at root cause

The JIT serializer likely serializes the `|date` filter output through `json.dumps` (or equivalent), which wraps strings in literal double quotes. That JSON string then gets inserted into the rendered HTML as-is. Plain str context vars seem to bypass this path — presumably because they're caught earlier and not re-serialized.

## Impact

Any template in a djust app that uses `|date` (or `|default` returning a str fallback) on a model field value renders with visible literal quotes. Visible in the browser as e.g. `"Apr 25, 2026"` in a date column.

## Environment

[...]

PR #1115, opened 2026-04-26
fix(framework-gaps): 4-issue v0.8.6 P2 drain (closes #1088, #1089, #1090, #1093)

### LINKED ISSUE #1088: tech-debt: Django system check for stale collectstatic client.min.js
## Trap surfaced by #1081

#1081 was reopened twice and consumed significant investigation time before the reporter root-caused it: their `staticfiles/djust/client.min.js` was Apr 22 / 0.5.5rc1-era (pre-`strip_filter_arg_quotes`), while the wheel they were running (0.8.3rc1) shipped a fresh Apr 25 copy. `manage.py collectstatic --clear` + browser hard-reload fixed it.

**The trap is structural:** anyone with `STATIC_ROOT` configured (i.e. anyone running djust in production behind WhiteNoise / nginx / a CDN) is vulnerable to this on every wheel upgrade if they forget the `--clear` flag — `collectstatic` without `--clear` won't replace files whose source mtime is older than the destination mtime in some configurations, or copies the wrong file when multiple `STATICFILES_DIRS` shadow each other. The user-visible failure mode looks like a framework bug (literal `&quot;` in DOM after WS update) and routes to "djust is broken", not "my static file is stale".

## Proposed mitigation

A new Django system check `djust.S0XX` (Static category) that runs on `manage.py check` / startup:

1. Read the wheel's bundled `client.min.js`:
   ```python
   import djust
   wheel_js = Path(djust.__file__).parent / "static" / "djust" / "client.min.js"
   wheel_hash = hashlib.sha256(wheel_js.read_bytes()).hexdigest()[:12]
   ```
2. Resolve where `staticfiles` finds `djust/client.min.js` (via `staticfiles.finders.find()` or by hashing files in `STATIC_ROOT`).
3. If the deployed copy's hash differs from the wheel copy's hash, raise `Warning` with a message like:
   ```
   djust.S001: stale static file — staticfiles serving an outdated client.min.js
     Wheel ships:  client.min.js (sha256: abc123...)
     Deployed:     <STATIC_ROOT>/djust/client.min.js (sha256: def456...)
     Run: python manage.py collectstatic --clear
   ```

## Design questions for the implementer

- **Severity**: `Warning` or `Error`? Probably `Warning` — `Error` would block startup, which is too aggressive for a non-functional regression that may not affect every page.
- **DEBUG-mode skip**: should the check be no-op when `settings.DEBUG=True`? Probably yes (devs use the runserver path which serves directly from the wheel; `staticfiles` cache is a deploy-time concern).
- **Hashing cost**: SHA-256 of one ~37 KB file at startup is sub-millisecond; acceptable.
- **Per-file vs aggregate**: just `client.min.js` is enough — it's the only file that's both shipped-by-wheel and routinely regenerated across releases. (Source files in `static/djust/src/` are also copied but rarely accessed directly in production.)
- **CDN / `ManifestStaticFilesStorage`**: when `ManifestStaticFilesStorage` is in use, the deployed file has a hash suffix (`client.min.HASH.js`). The check should resolve through the manifest, not assume an unhashed name.
- **Co-located in `checks.py`**: yes — joins existing C / V / S / T / Q categories.

## Out of scope for this issue

- Auto-running `collectstatic --clear` on the user's behalf (too magical; respects user's deploy pipeline).
- Hashing every static file in `static/djust/` (overkill; one file is sufficient signal).
- Changing the `collectstatic` default behavior in any way.

## Acceptance criteria

- New check `djust.S0XX` registered in `python/djust/checks.py`
- Test in `python/djust/tests/test_checks.py` that simulates a stale `STATIC_ROOT` and asserts the warning fires
- Test that asserts `DEBUG=True` skips the check (or other documented skip condition)
- Documentation note in `docs/website/guides/deployment.md` linking the check ID to the fix command

## References

- #1081 — the surfacing user report (now closed)
- PR #1086 — 25 regression tests locking the framework's `strip_filter_arg_quotes` invariant (the invariant the stale JS was missing)
- `python/djust/checks.py` — existing check pattern to follow

🤖 Filed via /pipeline-retro follow-up after #1081 resolution

johnrtipton:
## Summary

Single batch PR for 4 unrelated framework gaps surfaced by the v0.8.5 arc retros and Stage 13 review of PR #1091.

### #1093 — SSE-side `_flush_deferred_to_sse` legacy-view guard test (test-only)
Stage 13 review of PR #1091 flagged that the WS-side has a parallel test (`test_flush_deferred_handles_view_without_drain_method`) but the SSE-side only covered `view_instance=None`. Added `test_sse_flush_deferred_handles_view_without_drain_method` — mirrors WS shape.

### #1090 — debug-log on `|date` / `|time` filter parse failure
Both filters previously fell through silently to the original value when chrono parse failed. The #1081 4-round-reopen investigation would have collapsed to a 5-minute diagnosis with this log. Added `tracing::debug!` against target `djust.templates.filters` with offending value, format string, and chrono error. New `tracing` workspace dep in `djust_templates`; enable via `RUST_LOG=djust.templates.filters=debug` or Python LOGGING config.

### #1088 — `djust.C013` system check for stale `collectstatic` `client.min.js`
Anyone with `STATIC_ROOT` configured can ship a stale `client.min.js` after a wheel upgrade if they forget `collectstatic --clear`. Server runs new code; browser loads old client.js → wire-protocol skew. C013 compares SHA-256 of `STATIC_ROOT/djust/client.min.js` to the wheel-bundled copy at startup. Honors `DJUST_CONFIG['suppress_checks']=['C013']` for users who serve `client.min.js` from a CDN or custom build.

### #1089 — release wheel matrix expanded to cp313 + cp314
`.github/workflows/release.yml` previously built only cp310/cp311/cp312. Users on 3.13/3.14 fell back to source-compiling the sdist at `pip install` time → untested binaries (root cause of #1081's first reopen). Matrix now ships tested wheels for cp310–cp314 across Linux x86_64, macOS Intel + ARM, Windows x86_64.

## Test plan

- [x] `TestC013StaleCollectstatic` (5 cases): no-STATIC_ROOT skip, no-collected-file skip, matching-content quiet, diverged-content warning, suppress-via-DJUST_CONFIG silence
- [x] `test_sse_flush_deferred_handles_view_without_drain_method` mirrors the WS-side shape
- [x] 4745 Python pass; 1414 JS pass
- [x] Rust crate builds clean with new `tracing` dep
- [ ] CI green

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -61,7 +61,7 @@ jobs:
       fail-fast: false  # Continue building other platforms if one fails
       matrix:
         os: [ubuntu-latest, windows-latest]
-        python-version: ['3.10', '3.11', '3.12']
+        python-version: ['3.10', '3.11', '3.12', '3.13', '3.14']
         exclude:
           # Windows builds are slower; only build latest Python versions
           # Users on older Python can use Linux/macOS wheels or build from source
@@ -73,11 +73,19 @@ jobs:
             python-version: '3.11'
           - os: macos-15-intel
             python-version: '3.12'
+          - os: macos-15-intel
+            python-version: '3.13'
+          - os: macos-15-intel
+            python-version: '3.14'
           # macOS ARM - latest (macos-15)
           - os: macos-15
             python-version: '3.11'
           - os: macos-15
             python-version: '3.12'
+          - os: macos-15
+            python-version: '3.13'
+          - os: macos-15
+            python-version: '3.14'
 
     steps:
       - uses: actions/checkout@v6
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,71 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- **`djust.C013` system check — stale collectstatic copy of `client.min.js`
+  (closes #1088)** — anyone with `STATIC_ROOT` configured (typical
+  production deployment behind WhiteNoise / nginx / a CDN) can ship a
+  stale `client.min.js` after a djust wheel upgrade if they forget
+  `collectstatic --clear`. The server runs new code; the browser loads
+  old client.js → wire-protocol skew → mysterious VDOM patch failures.
+  #1081 was reopened twice before the reporter root-caused this
+  structurally-recurring trap.
+
+  C013 compares the SHA-256 of `STATIC_ROOT/djust/client.min.js`
+  against the wheel-bundled copy at `python/djust/static/djust/client.min.js`.
+  When they diverge, emits a Django system warning at startup with the
+  exact fix command. No-op when `STATIC_ROOT` is unset, when the
+  collected file is absent (pre-collectstatic), or when content matches.
+  Honors `DJUST_CONFIG = {"suppress_checks": ["C013"]}` for users who
+  serve `client.min.js` from a CDN or custom build.
+
+  Files: `python/djust/checks.py` (new `_check_stale_collected_client`,
+  wired into `check_configuration`); 5 cases in `TestC013StaleCollectstatic`
+  in `python/tests/test_checks.py` cover no-STATIC_ROOT skip,
+  no-collected-file skip, matching-content quiet, diverged-content
+  warning, suppress-via-DJUST_CONFIG silence.
+
+### Fixed
+
+- **`|date` and `|time` filters now debug-log on parse failure (closes
+  #1090)** — both filters previously fell through silently to the
+  original value when chrono failed to parse the input string. The
+  #1081 4-round-reopen investigation would have collapsed to a
+  5-minute diagnosis if a single line had been logged at parse-failure
+  time. Now the failure is surfaced via `tracing::debug!` against
+  target `djust.templates.filters` with the offending value, format
+  string, and chrono error message.
+
+  Enable via Python `LOGGING['loggers']['djust.templates.filters'] =
+  {'level': 'DEBUG'}` or set `RUST_LOG=djust.templates.filters=debug`
+  for the Rust-side `tracing` consumer. Behavior unchanged when the
+  log target is disabled — just no longer a silent void.
+
+  Files: `crates/djust_templates/Cargo.toml` (added `tracing`
+  workspace dep), `crates/djust_templates/src/filters.rs` (`|date`
+  arm at line ~248, `|time` arm at line ~284 — replaced
+  `Err(_) => Ok(value.clone())` with `Err(e) => { tracing::debug!(...);
+  Ok(value.clone()) }`).
+
+- **`_flush_deferred_to_sse` legacy-view guard now has a regression
+  test (closes #1093)** — Stage 13 review of PR #1091 flagged that the
+  WS-side `hasattr` guard had a parallel test
+  (`test_flush_deferred_handles_view_without_drain_method`) but the
+  SSE-side did not. New `test_sse_flush_deferred_handles_view_without_drain_method`
+  in `python/djust/tests/test_defer.py` mirrors the WS shape — a
+  legacy view class without `_drain_deferred` must short-circuit
+  cleanly without `AttributeError`.
+
+- **Release wheel matrix expanded to cp313 + cp314 (closes #1089)** —
+  `.github/workflows/release.yml` previously built only cp310/cp311/cp312
+  wheels. Users on Python 3.13 or 3.14 fell back to source-compiling
+  the sdist at `pip install` time, producing untested binaries whose
+  runtime behavior could diverge from CI-tested cp312 (this was the
+  root cause of #1081's first reopen — reporter on 3.14 hit a source-
+  compiled `_rust.cpython-314-darwin.so`). Matrix now ships tested
+  wheels for cp310–cp314 across Linux x86_64, macOS Intel + ARM, and
+  Windows x86_64 (Windows still excludes 3.10 per the existing
+  policy).
+
 - **View Transitions API integration in `applyPatches` (PR-B / ADR-013)** —
   Opt-in via `<body dj-view-transitions>`. When the browser supports
   `document.startViewTransition()` AND the body attribute is present
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -294,7 +294,7 @@ dependencies = [
 
 [[package]]
 name = "djust_components"
-version = "0.8.4-rc.1"
+version = "0.8.5-rc.1"
 dependencies = [
  "ahash",
  "criterion",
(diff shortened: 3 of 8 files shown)


johnrtipton (commented):

Verified all 4 issues across the batch. Build clean, all tests pass, YAML valid, CHANGELOG complete.

### Verification results

**#1090 Rust filters debug-log** — `cargo build -p djust_templates` clean (1.67s). Workspace `tracing = "0.1"` confirmed at `Cargo.toml:40`; inner crate uses `tracing = { workspace = true }` at `crates/djust_templates/Cargo.toml:21`. No version drift. Logged fields use `%` (Display) format — server-side log target, no XSS surface, no recursive expansion. `Cargo.lock` correctly picks up `tracing` as a new dep of `djust_templates` v0.8.5-rc.1. All 276 `djust_templates` lib tests pass.

**#1088 C013 system check** — `_check_stale_collected_client` at `python/djust/checks.py:288-362` orders early-return guards correctly: `_is_check_suppressed` first (`:307`), `STATIC_ROOT` None-check (`:310-313`), collected-file existence (`:315-321`), wheel-bundled path resolution via `djust.__file__` (`:325-331`). SHA-256 streamed in 8KB chunks (`:336-340`) — fine for ~388KB raw client.js. `try/except OSError` around hash compare (`:342-346`) silently degrades on permission/IO error rather than crashing startup — correct for a startup check.

**#1088 tests** — All 5 cases pass; branches verified: no-STATIC_ROOT skip, no-collected-file skip, matching-content quiet, diverged-content warning (asserts both `Stale collectstatic` in msg AND `collectstatic --clear` in hint at `test_checks.py:752-753`), suppress-via-DJUST_CONFIG silence. The matching-content test correctly `pytest.skip()`s when wheel-bundled `client.min.js` not present in dev tree (`:721-722`) — defensive against fresh clones.

**#1089 wheel matrix** — Diff verified: `python-version: ['3.10', '3.11', '3.12', '3.13', '3.14']` for ubuntu/windows at `release.yml:64`. 4 new include rows added correctly: `macos-15-intel × {3.13, 3.14}` (`:76-79`), `macos-15 × {3.13, 3.14}` (`:85-88`). Windows 3.10 exclude rule preserved (`:68-69`). YAML parses clean.

**#1093 SSE legacy-view test** — `test_sse_flush_deferred_handles_view_without_drain_method` at `test_defer.py:386-400` mirrors the WS-side shape at `test_defer.py:236-245`: same `_LegacyView` class with `pass`, same "no-op, not raise" assertion. Targets `_flush_deferred_to_sse` from `djust.sse`, which has the matching `hasattr` guard at `sse.py:557`. Test passes.

**CHANGELOG** — All 4 entries under `[Unreleased]`: 1 Added (C013) + 3 Fixed (#1090, #1093, #1089). Each cites issue number, files touched, and rationale. Accurate.

### Findings

🔴 **Must-fix**: none

🟡 **Should-fix**: none

🟢 **Nits**

- `python/djust/checks.py:360` — extra parens around `fix_hint=("Run: ...")` are redundant (single string, not a continuation). Cosmetic; ignore.
- `release.yml:71-79` — macOS-Intel matrix covers 3.11–3.14 but skips 3.10 (consistent with the pre-existing baseline). Intel users on 3.10 will source-compile. Likely intentional given Intel-runner cost; documenting for awareness, not a blocker.
- `python/tests/test_checks.py:721` — `import pytest` and `from pathlib import Path` inside the test method body work, but module-top imports would be slightly cleaner. Stylistic.

### Verification commands run

- `cargo build -p djust_templates` → clean
- `cargo test -p djust_templates --lib` → 276 passed
- `.venv/bin/pytest python/tests/test_checks.py python/djust/tests/test_defer.py -q` → 205 passed
- `.venv/bin/python -c "import yaml; yaml.safe_load(open('.github/workflows/release.yml'))"` → YAML OK
- Targeted: `TestC013StaleCollectstatic` (5/5) + `test_sse_flush_deferred_handles_view_without_drain_method` (1/1)

### Verdict

**APPROVE** — Zero must-fix or should-fix findings. Ship it.

johnrtipton:
# Retro: PR #1115 — 4-issue v0.8.6 framework-gap drain (closes #1088, #1089, #1090, #1093)

## Quality rating: 9/10

Tight batch PR, all 4 unrelated issues in scope, Stage 11 APPROVE with zero 🔴/🟡, 4745 Python + 1414 JS tests green, 12/12 CI green.

## What worked

1. **Smallest-first ordering inside the batch.** #1093 (test-only) → #1090 (single-line log) → #1088 (new check) → #1089 (yml matrix). Each prior fix de-risked the next: by the time the C013 check went in, the build was known-clean. No mid-batch surprises.

2. **`tracing::debug!` over `println!`/`log::debug!`.** Workspace already uses `tracing` (djust_live, Cargo.toml:40). Reusing it kept the dep graph tight, gave structured fields (`value=`, `format=`, `error=`), and integrates with whatever subscriber the hosting app sets up — Python LOGGING via tracing-subscriber, JSON via `RUST_LOG`, etc. No new logging surface invented.

3. **C013 design choice: SHA-256, not mtime.** `mtime` is unreliable across pip-install, tar extraction, and file-system mounts (containers especially). The reviewer specifically called out the 8KB chunked read as "good for large files" — same pattern is right for any future file-comparison check.

4. **`try/except OSError` on the hash compare.** The check runs at Django startup; if it can't read the file (permissions, mounted-FS quirk), it should silently skip rather than block startup. Stage 11 confirmed this is the right shape.

## What didn't work

1. **Initial batch test pass missed the `RUST_LOG` documentation hint.** I documented the log target in the CHANGELOG (`djust.templates.filters`) but didn't include a docstring or doc-page reference for HOW to enable it from Python. A user discovering the silent fallback today still needs to know `RUST_LOG=djust.templates.filters=debug` exists. Minor — but the half-step gap means a follow-up task to thread this into the Python-side docs.
   **Action**: tracked as a v0.8.7 nit — add to `docs/website/guides/debugging.md` (or wherever the equivalent lives).

## What to improve

1. **Multi-issue batch PRs benefit from a per-issue "what test proves this" matrix in the PR body.** PR #1115's body listed each issue's scope but didn't have a single-table mapping issue → file → test. For 4-issue batches, that table makes Stage 11 reviewers' job faster.
   **Action**: tracked — pipeline-run subagent prompt for batch-PR self-review should require this matrix.

- `docs/website/guides/debugging.md` (or equivalent) — document the new `djust.templates.filters` tracing target. Include both `RUST_LOG=...` and Python LOGGING configurations.
- pipeline-run skill (or PR template): for batch PRs (>2 issues), require an issue × file × test mapping table in the PR body.

## Review stats

- 🔴 Stage 11: 0
- 🟡 Stage 11: 0
- 🟢 Stage 11: 3 (cosmetic — redundant parens, macOS-Intel matrix consistency with baseline, in-method imports in tests)
- Stage 13 (Re-Review): skipped per `skip_if`
- Tests: 4745 Python + 1414 JS green; 276 cargo tests pass
- CI: 12/12 green
- Issues closed: 4 (#1088, #1089, #1090, #1093)

PR #191, opened 2026-01-31
fix: bootstrap debug panel tabs and include _debug in WebSocket responses (#189, #190)

johnrtipton:
## Summary

Two related debug panel fixes:

1. **#189**: The debug panel's Handlers, Variables, and Components tabs were always empty on page load because `processDebugInfo()` was never called with `window.DJUST_DEBUG_INFO` during initialization.

2. **#190**: WebSocket responses never included `_debug` payloads, so tabs couldn't update after user interactions. Added `_attach_debug_payload()` to include debug info in all patch and HTML update responses when `DEBUG=True`.

## Changes

- `client-dev.js` — Call `processDebugInfo(window.DJUST_DEBUG_INFO)` after panel construction
- `websocket.py` — Added `_attach_debug_payload()` method and call it from `_send_update()` for both patch and html_update responses
- `CHANGELOG.md` — Added entries under Unreleased
- Tests added for both changes

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)

## Test Plan

- [x] Existing tests pass (`make test`)
- [x] New tests added for changed functionality
  - 5 new JS tests for bootstrap behavior in `debug_panel_harness.test.js`
  - Python tests for `_attach_debug_payload` in `test_debug_panel.py`

## Checklist

- [x] Code follows project conventions
- [x] Self-reviewed for correctness, edge cases, and security
- [x] Documentation updated (if applicable)
- [x] No breaking API changes (or clearly noted above)

## Related Issues

Closes #189
Closes #190

Issue #189, opened 2026-01-31
Debug panel: Handlers, Variables, Components tabs empty despite DJUST_DEBUG_INFO being available


johnrtipton:
## Bug

The Handlers, Variables, and Components tabs in the debug panel are always empty, showing "No event handlers detected", "No public variables detected", and "No components detected" respectively — even though `window.DJUST_DEBUG_INFO` contains all this data on page load.

## Root Cause

The server injects rich debug data into the initial HTTP response via `get_debug_info()` in `post_processing.py`:

```js
window.DJUST_DEBUG_INFO = {
  view_class: "ExamplesView",
  handlers: { increment: {...}, decrement: {...}, add_todo: {...}, ... },
  variables: { count: {type: "int", value: "0"}, todos: {type: "list", value: "[]"}, ... },
  template: "marketing/examples.html",
  config: { maxHistory: 50 }
}
```

However, the debug panel only reads `DJUST_DEBUG_INFO.view_class` for state key scoping. It never populates `this.handlers`, `this.variables`, or `this.components` from this initial data. Instead, it waits for `_debug` payloads in WebSocket responses (via `processDebugInfo()`), but those are never sent (see #190).

## Expected Behavior

On initialization, the panel should bootstrap its tabs from `DJUST_DEBUG_INFO`:
- **Handlers tab**: Show all registered event handlers with params, types, descriptions
- **Variables tab**: Show all public view variables with current values
- **Components tab**: Show component tree if available

After interactions, these should update with fresh data from WebSocket responses (once #186 and #190 are fixed).

## Suggested Fix

In the panel constructor or `init()`, after `loadState()`:

```js
if (window.DJUST_DEBUG_INFO) {
    this.processDebugInfo(window.DJUST_DEBUG_INFO);
}
```

This reuses the existing `processDebugInfo()` method which already handles handlers, variables, and components.

## Testing

Should include a Playwright E2E test that:
1. Navigates to `/examples/`
2. Opens the debug panel
3. Clicks the Handlers tab and asserts handlers are listed (increment, decrement, etc.)
4. Clicks the Variables tab and asserts variables are listed (count, todos, etc.)

## Related

- #186 (received WebSocket messages not captured)
- #190 (server doesn't send `_debug` in WebSocket responses)

Issue #1111, opened 2026-04-26
feat: data_table — row_click_event / row_url for row-level navigation


johnrtipton:
## Context

Currently there is no way to make an entire table row clickable for navigation. The `selectable` option adds checkboxes but does not navigate. Users must either wrap individual cells in `<a>` tags (see #1110) or render the `<tbody>` manually — both of which forfeit the component abstraction.

## Requested API

Two options are proposed; either would address the need:

### Option A — `row_url` key (simple static navigation)

```python
{% data_table rows=rows columns=columns row_url="claim_url" %}
```

Where `claim_url` is a key in each row dict. The component renders:

```html
<tr data-href="{{ row.claim_url }}" style="cursor:pointer" onclick="window.location=this.dataset.href">
```

### Option B — `row_click_event` (LiveView-idiomatic, preferred)

```python
{% data_table rows=rows columns=columns row_click_event="open_claim" %}
```

Clicking a row fires a djust event with `dj-click="open_claim"` and `data-value="{{ row.<row_key> }}"`. Then in the view:

```python
@event_handler()
def open_claim(self, value: str = "", **kwargs):
    claim_id = int(value)
    self.redirect(reverse("claims:detail", kwargs={"claim_id": claim_id}))
```

Option B is more LiveView-idiomatic (no raw JS, integrates with djust's event system). Option A is simpler for static URL navigation.

## Why

Row-level navigation is the primary interaction pattern for list/table views in any CRUD application. Without it, consumers must either use the link column workaround (see #1110) or bypass the component entirely. The component is incomplete for real-world use without row-click support.

Affects downstream users (e.g. nyc-claims PR #189).

Issue #1110, opened 2026-04-26
feat: data_table — link column type


johnrtipton:
## Context

The `{% data_table %}` component renders cells as `{{ row|dictsort:col.key|first }}` without `|safe`, so pre-rendered HTML links in row dicts get escaped. To render a column as a hyperlink, consumers currently must either:

1. Render the table body manually (forfeiting the component), or
2. Store pre-escaped HTML in the row dict and patch the template (fragile)

## Requested API

```python
table_columns = [
    {
        "key": "claim_number",
        "label": "Claim #",
        "sortable": True,
        "link": "claim_url",        # NEW: key in same row dict that holds the href
        "link_class": "claim-link", # optional CSS class on the <a>
    },
    ...
]
```

Then in `refresh_table_server()` / row dicts:

```python
{"claim_number": "2026PI000001", "claim_url": "/claims/1/", ...}
```

The table template renders that column as:

```html
<a href="{{ row.claim_url }}" class="{{ col.link_class }}">{{ row.claim_number }}</a>
```

## Why

Every admin/dashboard use case needs clickable rows. Without this, the component is unusable for navigation-heavy views. The current workarounds either forfeit the component entirely or require patching the template, both of which break on upgrades.

Affects downstream users (e.g. nyc-claims PR #189).

PR #1117, opened 2026-04-26
fix(components): DataTableMixin LiveView compat — pre-mount guard + @event_handler decoration (closes #1114)

### LINKED ISSUE #1114: DataTableMixin is incompatible with LiveView JIT serialization and BUG-06 pre-mount lifecycle
## Summary

Using `DataTableMixin` in a `LiveView` causes a blank/empty table on every page load, even when `refresh_table_server()` correctly populates `self.table_rows` in `mount()`. This is caused by two separate issues that compound each other.

**Severity**: High

**Related**: #1110 (link column), #1111 (row_click_event)

---

## Root Cause 1 — BUG-06: `get_context_data()` is called before `mount()` during WebSocket setup

djust's WebSocket consumer calls `get_context_data()` (to build the initial Rust VDOM snapshot) **before** `mount()` runs. `DataTableMixin.init_table_state()` sets `self.table_rows = []` only during `mount()`.

On the pre-mount `get_context_data()` call:
1. `self.table_rows` doesn't exist
2. `get_table_context()` → `AttributeError`
3. djust catches the error silently
4. The rendered VDOM is empty
5. All subsequent VDOM patches diff against empty content and produce incorrect results

---

## Root Cause 2 — `table_rows` is stored as a large serialized list

When `refresh_table_server()` is used (recommended for FK traversal), rows are serialized to plain Python dicts. djust serializes all public instance attributes to JSON between WebSocket events. Problems:

- A table with 15+ rows × 8 fields creates a large JSON payload on every event
- After JSON round-trip, rows are plain dicts — but the Rust VDOM renderer cannot access arbitrary dict keys via dot-notation (e.g. `{{ row.claim_number }}` fails silently for keys not in the original context schema)
- This forces use of the `{% data_table %}` template tag exclusively, but that tag applies `html.escape()` to all cell values, making link cells impossible

---

## Root Cause 3 — `on_table_*` methods are not decorated with `@event_handler()`

`DataTableMixin` defines `on_table_sort`, `on_table_search`, etc. as plain methods. djust's `event_security=strict` (the default) requires `@event_handler()` decoration.

Additionally, the event names in templates (`dj-click="table_sort"`) don't match the method names (`on_table_sort`). Consumers must write wrapper methods for every event, defeating the purpose of the mixin.

---

## Working Approach (what we use instead)

- Pass the queryset directly in `get_context_data()` — djust JIT handles FK traversal natively
- Implement search/sort/pagination with `@event_handler()`-decorated methods directly on the view
- No mixin needed

---

## Requested Changes

1. **BUG-06 guard pattern**: Add a `DataTableMixin` usage guide for `LiveView` (not just `Component`) documenting the pre-mount guard:
   ```python
   if hasattr(self, 'table_rows'):
       ctx.update(self.get_table_context())
   # else fall back to empty state
   ```

2. **Fix `@event_handler()` decoration**: Decorate `on_table_sort`, `on_table_search`, and all other `on_table_*` methods with `@event_handler()` so they work under `event_security=strict` without wrapper boilerplate.

3. **Add class-level default**: Add `table_rows = []` as a class-level attribute on `DataTableMixin` so `get_table_context()` doesn't crash before `init_table_state()` runs.

4. **Document the API boundary**: Document explicitly that `DataTableMixin` is designed for the `Component` API (`get_template_context`) — using it with `LiveView` (`get_context_data`) requires the pre-mount guard described above.

---

## Downstream Impact

nyc-claims PR #189 attempted migration to `DataTableMixin`, caused a blank dashboard. PR #191 reverted to native event handlers. Filing this as a follow-up to #1110 and #1111.

johnrtipton:
## Summary

Closes #1114 (HIGH severity). `DataTableMixin` in a `LiveView` (rather than a `Component`) caused a blank/empty table on every page load. nyc-claims PR #189 hit this and reverted in PR #191.

## Root causes (3 compounding)

1. **BUG-06 pre-mount lifecycle** — djust's WebSocket consumer calls `get_context_data()` (which often calls `get_table_context()`) BEFORE `mount()` runs. `init_table_state()` hadn't run, `self.table_rows` didn't exist, `get_table_context()` raised `AttributeError` silently → empty initial VDOM → all subsequent VDOM patches diff against empty content.
2. **Missing `@event_handler()` decoration** on the 21 `on_table_*` methods. Default `event_security="strict"` rejected them.
3. **Documentation gap** — the LiveView vs Component API boundary wasn't called out anywhere.

## Fixes

| File | Change |
|---|---|
| `python/djust/components/mixins/data_table.py` | New module-level `_PRE_MOUNT_TABLE_CONTEXT` (~80 keys covering Phase 1-5); `get_table_context()` guards on `hasattr(self, 'table_rows')`; all 21 `on_table_*` handlers decorated with `@event_handler()`; mixin docstring expanded with LiveView lifecycle note + recommended pattern |
| `python/tests/test_data_table_mixin_liveview.py` | New 8-case test file |
| `CHANGELOG.md` | `[Unreleased] / Fixed` entry |

## Test plan

| Issue # | File touched | Test class |
|---|---|---|
| #1114 root cause 1 | `data_table.py` pre-mount guard | `PreMountGuardTest` (4 cases) |
| #1114 root cause 2 | `data_table.py` 21 decorators | `EventHandlerDecorationTest` (2 cases) |
| #1114 root cause 3 | `data_table.py` docstring | `DocstringRegressionTest` (2 cases) |

- [x] 8 new cases pass; 4753 Python pass overall
- [x] Pre-mount key set is a superset of post-mount (catches future post-mount additions that forgot to update the default)
- [x] Handler count matches expected (catches future additions that forgot decoration)
- [ ] CI green

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -156,6 +156,52 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- **`DataTableMixin` LiveView compatibility — pre-mount guard +
+  `@event_handler()` decoration on all `on_table_*` methods (closes
+  #1114)** — using `DataTableMixin` in a `LiveView` (rather than a
+  `Component`) caused a blank/empty table on every page load even
+  when `refresh_table_server()` correctly populated `self.table_rows`
+  in `mount()`. Three compounding root causes:
+
+  1. **BUG-06 pre-mount lifecycle**: djust's WebSocket consumer calls
+     `get_context_data()` (which often calls `get_table_context()`)
+     BEFORE `mount()` runs, to build the initial Rust VDOM snapshot.
+     `init_table_state()` hadn't run yet, so `self.table_rows` didn't
+     exist and `get_table_context()` raised `AttributeError`. djust
+     caught it silently → empty initial VDOM → all subsequent VDOM
+     patches diff against empty content → wrong renders.
+  2. **Missing `@event_handler()` decoration**: `on_table_sort`,
+     `on_table_search`, and 19 other handlers were plain methods.
+     djust's default `event_security="strict"` rejected them — every
+     consumer had to write wrapper boilerplate.
+  3. **Documentation gap**: the API boundary between Component and
+     LiveView use cases wasn't called out anywhere in the mixin's
+     docstring.
+
+  Fix: `get_table_context()` now guards on `hasattr(self, "table_rows")`
+  and returns `_PRE_MOUNT_TABLE_CONTEXT` (a module-level minimal
+  default with every key the `{% data_table %}` template tag reads —
+  ~80 keys covering all 5 phases). All 21 `on_table_*` handlers now
+  carry `@event_handler()` decoration. Mixin docstring expanded with a
+  "LiveView vs Component lifecycle" note + recommended pattern for
+  large datasets (pass queryset directly via `get_context_data()`,
+  define `@event_handler()` methods on the view).
+
+  Downstream impact: nyc-claims PR #189 attempted migration and hit
+  this; PR #191 reverted to native handlers. With this fix,
+  `DataTableMixin` is usable from `LiveView` subclasses without
+  per-handler boilerplate.
+
+  8 regression cases in
+  `python/tests/test_data_table_mixin_liveview.py` cover: pre-mount
+  call doesn't raise; pre-mount returns the default; post-mount
+  returns real state; pre-mount key set is a superset of post-mount
+  (catches future post-mount additions that forgot to update the
+  default); all 21 expected handlers have `_djust_decorators`
+  metadata; handler count matches expected (catches future additions
+  that forgot decoration); docstring mentions LiveView lifecycle and
+  `@event_handler()` decoration (catches doc-rot).
+
 - **`handleMessage` interleaving across `await` boundaries (closes #1098)** —
   PR-A (v0.8.5rc1) made `LiveViewWebSocket.handleMessage` and
   `LiveViewSSE.handleMessage` async without serializing the inbound
--- a/python/djust/components/mixins/data_table.py
+++ b/python/djust/components/mixins/data_table.py
@@ -34,6 +34,7 @@ import math
 import re
 
 from djust.components.utils import format_cell, interpolate_color_gradient
+from djust.decorators import event_handler
 
 __all__ = ["DataTableMixin"]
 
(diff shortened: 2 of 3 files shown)


johnrtipton (commented):
## Stage 11 Code Review — APPROVE (self-review, posted as comment)

Solid, surgical fix. All three documented root causes addressed, regression-tested, and documented. 4,753 Python tests pass. No security anti-patterns. Ruff clean. Imports match sibling-mixin convention.

### Verification

- **21/21 `@event_handler()` decorations landed** — `grep -B1 "def on_table_" python/djust/components/mixins/data_table.py | grep -c "@event_handler()"` = 21, matches the 21 `def on_table_*` definitions.
- **Pre-mount guard works** — `data_table.py:1166-1167` returns `_PRE_MOUNT_TABLE_CONTEXT` when `hasattr(self, "table_rows")` is False; the new test asserts it doesn't raise.
- **Post-mount keyset ⊆ pre-mount keyset** test (`test_data_table_mixin_liveview.py:88` `missing = post_mount_keys - pre_mount_keys`) correctly enforces the future-proof direction.
- **`@event_handler` import path** (`data_table.py:37`) matches all 8 sibling mixins (`accordion.py`, `carousel.py`, etc.).
- **Pre-mount default values** spot-checked: `loading: False`, `paginate: False`, `selected_rows: []` correct; `import_preview: True` correctly mirrors the class default `table_import_preview = True` at line 397.
- **No auto-reject patterns** — no `mark_safe(f"...")`, no `print()`, no bare `except: pass`, no f-string logging in production code (`self.fail(f"...")` in tests is fine).
- **Backwards compat** — `examples/demo_project/djust_rentals/components/tests/test_component_security.py` (4 DataTable component tests) still pass; that suite tests a separate `DataTable` class, not the mixin.

### 🟢 Nits (non-blocking)

1. **Docstring inaccuracy at `data_table.py:1163-1164`** — claims `table_rows` is set "as the LAST line of `init_table_state()`". It's actually line 439, with ~21 more attribute assignments after it (lines 440-460, e.g. `table_loading`, `table_editing_rows`, `table_column_order`, etc.). The guard still works (the *first* line that creates user-supplied state is enough as a sentinel), but the doc claim is wrong. Suggest: "`table_rows` is set EARLY in `init_table_state()` — once it exists, the rest of the state is being initialized in the same call". Not blocking.

2. **`show_stats` in pre-mount default but not in post-mount return** (`data_table.py:260` vs missing in the post-mount dict at lines 1168-1260). Pre-existing; the post-mount path never adds `show_stats` even though `table_show_stats` exists at line 381 and is read at line 1325. Templates reading `{{ show_stats }}` get `False` pre-mount and empty post-mount. Out of scope for this PR but worth filing as a follow-up.

3. **Decoration style mismatch with siblings** — sibling mixins use bare `@event_handler` (no parens); this PR uses `@event_handler()`. Both work (`decorators.py:192-200` handles both forms). Stylistic; not worth changing 21 places.

### Summary

The 80-key pre-mount default is comprehensive (covers all 5 phases). The four-test approach to regression coverage (pre-mount no-raise + key-superset + handler-decoration + docstring-rot) is a strong belt-and-suspenders setup against future regressions. The downstream-impact note (nyc-claims PR #189/#191) anchors the fix to real consumer pain.

APPROVE.

johnrtipton:
# Retro: PR #1117 — DataTableMixin LiveView compat (closes #1114, HIGH severity)

## Quality rating: 9/10

HIGH-severity downstream blocker fixed in a single PR with comprehensive test coverage. Stage 11 APPROVE with zero 🔴/🟡, three 🟢 nits (one fixed inline, one filed as #1118, one stylistic).

## What worked

1. **Three-root-cause analysis from the issue body led directly to a three-fix PR.** The issue reporter did the diagnostic work; my job was to translate each root cause into a discrete code change and then write tests that prove each fix. The 8-test structure mirrors the 3 root causes (`PreMountGuardTest` × 4, `EventHandlerDecorationTest` × 2, `DocstringRegressionTest` × 2). Reviewer can map each test to its root cause without reading the test bodies.

2. **The pre-mount-vs-post-mount keyset comparison test is the right shape.** `test_pre_mount_default_has_required_template_keys` asserts `post_mount_keys ⊆ pre_mount_keys`. Future PRs that add a new key to the post-mount return dict will trip this test if they forget to update `_PRE_MOUNT_TABLE_CONTEXT`. The test caught the inverse case for free — `show_stats` is in pre-mount but missing from post-mount, surfaced as Stage 11's 🟢 #2 nit, filed as #1118.

3. **Bulk-decorate via Python regex was the right tool for 21 mechanical edits.** Hand-editing 21 methods invites missing one. The bulk-decorate script reported "Methods found: 21; decorated: 21" — verifiable count.

4. **Module-level constant for the pre-mount default.** Putting `_PRE_MOUNT_TABLE_CONTEXT` at module level (not class) makes it greppable, audit-trail-friendly (single source of truth), and avoids per-instance overhead. Mirrors the pattern used elsewhere in the codebase.

## What didn't work

1. **First docstring claim was wrong.** I wrote "table_rows is set as the LAST line of init_table_state()" — but it's set partway through, with ~21 more attribute assignments after. Stage 11 reviewer caught it. Fixed inline. Lesson: when writing docstrings about code I just touched, re-verify by reading the actual line numbers, not by recall.

## What to improve

1. **Pre-flight: verify docstring claims about line ordering before writing them.** When a docstring says "X is the last line of Y", grep to confirm. Cheap discipline that prevents the doc-rot class of issues.
   **Action**: this is a one-off — not worth a tracker. Logging here for the milestone retro.

- None — this is the kind of fix that's covered by the existing CLAUDE.md "process canonicalizations" section (specifically the rule about test-all-sites for mechanical PRs, which I followed).

## Review stats

- 🔴 Stage 11: 0
- 🟡 Stage 11: 0
- 🟢 Stage 11: 3 (docstring nit fixed inline; show_stats follow-up filed as #1118; decoration-style nit cosmetic)
- All blocking findings addressed: yes (zero blocking; one cosmetic addressed inline)
- Stage 13: skipped per `skip_if`
- Tests: 8 new + 4753 Python pass (was 4745, +8)
- CI: 12/12 green

## Cross-arc context (v0.8.6 extension running tally)

| PR | Issue | Theme | Verdict |
|---|---|---|---|
| #1117 (this) | #1114 | DataTableMixin LiveView compat (HIGH) | APPROVE |
| → up next | #1110 + #1111 | data_table link column + row navigation | pending |
| → up next | NEW | async dj-mounted hooks + applyPatches/transition-name docs | pending |

The async-hooks PR will be the last v0.8.6 add. Then milestone retro + release cut.

PR #1119, opened 2026-04-26
feat(data_table): link column type + row-level navigation (closes #1110, #1111)

### LINKED ISSUE #1110: feat: data_table — link column type

johnrtipton:
## Summary

Closes #1110 + #1111 — two related `{% data_table %}` enhancements requested by nyc-claims for admin/dashboard list views.

## #1110 — link column type

Column dicts now accept `link` (names a row-dict key holding the href) and `link_class` (optional CSS class on the `<a>`):

```python
table_columns = [
    {"key": "claim_number", "label": "Claim #",
     "link": "claim_url", "link_class": "claim-link"},
]
# Row dicts include both keys:
{"claim_number": "2026PI000001", "claim_url": "/claims/1/", ...}
```

Renders as `<td><a href="/claims/1/" class="claim-link">2026PI000001</a></td>`. Falls through to plain text when `col.link` is unset — strict pre-#1110 backwards compat.

## #1111 — row-level navigation

Two API shapes (Option B preferred per issue body):

**Option B — `row_click_event` (LiveView-idiomatic)**:
```python
table_row_click_event = "open_claim"
table_row_click_value_key = "id"  # default
```
Fires `dj-click="open_claim" data-value="<row.id>"` on every `<tr>`.

**Option A — `row_url` (static URL fallback)**:
```python
table_row_url = "claim_url"
```
Adds `data-href` + inline `onclick="window.location=this.dataset.href"` on every `<tr>`.

`row_click_event` takes precedence when both are set. Both add `cursor:pointer`.

| Issue | Modified files | Tests |
|---|---|---|
| #1110 | `templates/djust_components/table.html` (cell `<a>` wrap) | `LinkColumnStructureTest` (3 cases) |
| #1111 | `templatetags/djust_components.py` (data_table tag fn), `templates/djust_components/table.html` (`<tr>` attrs), `mixins/data_table.py` (3 class attrs + `get_table_context()` + `_PRE_MOUNT_TABLE_CONTEXT`) | `RowClickEventStructureTest` (3) + `RowUrlStructureTest` (2) + `MixinDefaultsTest` (4) + `TemplateTagDispatcherTest` (2) |

## Test plan

- [x] 14 new tests in `test_data_table_link_row_nav.py` all pass
- [x] Tests assert STRUCTURAL output (anchor presence, dj-click attribute, data-href attribute) rather than extracted cell content (cell content extraction uses `row|dictsort:col.key|first` which is Rust-template-engine-only)
- [x] Pre-#1110 compat: `col` without `link` → no `<a>` wrap; absent `row_click_event` and `row_url` → no `<tr>` modifications (asserted in negative tests)
- [x] 4767 Python pass overall (was 4753, +14 new)
- [ ] CI green

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -213,6 +213,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   `table_row_click_value_key`, and `table_row_url` class attributes,
   threaded through `get_table_context()` and `_PRE_MOUNT_TABLE_CONTEXT`.
 
+  **Security note for Option A (`row_url`)**: the URL flows into JS via
+  `onclick="window.location=this.dataset.href"`. Only assign
+  developer-controlled URLs (typically computed from `reverse()`);
+  user-controlled strings could enable `javascript:` URI execution.
+  **CSP note**: Option A requires `'unsafe-inline'` in `script-src`;
+  prefer Option B (LiveView event) when CSP is strict. Option B is
+  CSP-clean — the click is dispatched via the existing djust event
+  pipeline, no inline JS executed.
+
   14 regression cases in `python/tests/test_data_table_link_row_nav.py`
   cover: link-column emits `<a>`; link_class flows through; no-link
   pre-#1110 compat; `row_click_event` adds `dj-click` to every `<tr>`;
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -154,6 +154,74 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   ROADMAP Phoenix LiveView Parity Tracker `View Transitions API` →
   shipped. Quick Win #23 closed.
 
+### Added
+
+- **`{% data_table %}` link column type (closes #1110)** — column dicts
+  now accept a `link` key naming another row dict key that holds the
+  href, and an optional `link_class` for the `<a>` element's CSS class:
+
+  ```python
+  table_columns = [
+      {"key": "claim_number", "label": "Claim #", "link": "claim_url",
+       "link_class": "claim-link"},
+  ]
+  # row dicts include both keys:
+  {"claim_number": "2026PI000001", "claim_url": "/claims/1/", ...}
+  ```
+
+  Renders as:
+
+  ```html
+  <td><a href="/claims/1/" class="claim-link">2026PI000001</a></td>
+  ```
+
+  Falls through to plain text when `col.link` is unset — strict
+  backwards-compat with pre-#1110 column dicts. Replaces the
+  `_inject_link_column` regex post-process workaround downstream
+  consumers had to maintain (e.g. nyc-claims).
+
+- **`{% data_table %}` row-level navigation: `row_click_event` + `row_url`
+  (closes #1111)** — the entire `<tr>` becomes clickable for navigation.
+  Two API shapes:
+
+  **Option B (preferred — LiveView-idiomatic)**: `row_click_event` fires
+  a djust event with `data-value=row[row_click_value_key]`. Default
+  value key is `"id"`; override per-table for slug-based routing:
+
+  ```python
+  table_row_click_event = "open_claim"
+  table_row_click_value_key = "uuid"
+  ```
+
+  ```python
+  @event_handler()
+  def open_claim(self, value: str = "", **kwargs):
+      self.redirect(reverse("claims:detail", kwargs={"claim_id": value}))
+  ```
+
+  **Option A (static URL fallback)**: `row_url` names a row dict key
+  containing the href; the `<tr>` gets `data-href` + an `onclick` that
+  reads `this.dataset.href` and navigates:
+
+  ```python
+  table_row_url = "claim_url"
+  ```
+
+  Both options also wire `style="cursor:pointer"` on each `<tr>` for
+  the affordance. `row_click_event` takes precedence when both are set.
+  Mirrored in `DataTableMixin` via `table_row_click_event`,
+  `table_row_click_value_key`, and `table_row_url` class attributes,
+  threaded through `get_table_context()` and `_PRE_MOUNT_TABLE_CONTEXT`.
+
+  14 regression cases in `python/tests/test_data_table_link_row_nav.py`
+  cover: link-column emits `<a>`; link_class flows through; no-link
+  pre-#1110 compat; `row_click_event` adds `dj-click` to every `<tr>`;
+  `row_click_value_key` overrides default `id`; absent `row_click_event`
+  → no `<tr>` `dj-click` (compat); `row_url` adds `data-href` + JS;
+  `row_click_event` precedence over `row_url`; mixin class-attr
+  defaults; per-view override; pre-mount default + post-mount context
+  + template-tag function include all 3 new keys.
+
 ### Fixed
 
 - **`DataTableMixin` LiveView compatibility — pre-mount guard +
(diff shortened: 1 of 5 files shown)


johnrtipton (commented):
## Stage 11 — Code Review

### Verification

- [x] All 14 new tests pass (`python/tests/test_data_table_link_row_nav.py`)
- [x] PR #1117 keyset invariant still holds (`PreMountGuardTest::test_pre_mount_default_has_required_template_keys`); all 8 mixin tests green
- [x] Default propagation aligned at all three layers: tag fn defaults `("", "id", "")` (`templatetags/djust_components.py:599-601`) match `_PRE_MOUNT_TABLE_CONTEXT` (`mixins/data_table.py:289-291`) match class attrs (`mixins/data_table.py:409-413`) match `get_table_context()` (`mixins/data_table.py:1275-1277`)
- [x] Template `{% if row_click_event %} … {% elif row_url %} …` short-circuit is correct (`templates/djust_components/table.html:45`); `RowUrlStructureTest::test_row_click_event_takes_precedence_over_row_url` confirms `data-href` is suppressed when both set
- [x] Tag function backwards-compat: 3 new kwargs all have defaults; existing callers unaffected
- [x] Test docstrings explicitly call out Rust-engine-only `dictsort:col.key|first` chain (file header lines 5-9; inline comments at 82, 158)
- [x] No auto-reject patterns introduced (`mark_safe`, bare `except`, f-string logging, unguarded `console.log`, `print()` in production)
- [x] CHANGELOG entry under \`### Added\` with #1110 + #1111 cross-refs

### Findings

🟡 **Should-fix — `data_table` docstring missing Phase 6 args** (`templatetags/djust_components.py:602-703`). The docstring stops at `Phase 5 args`; `row_click_event`, `row_click_value_key`, `row_url` are not described. Every prior phase (1-5) has a docs block — this is the API surface developers introspect. Add a `Phase 6 args:` section.

🟡 **Should-fix — `row_url` XSS / `javascript:` URL warning missing**. The `<tr onclick=\"window.location=this.dataset.href\">` path (`table.html:45`) reads the raw `data-href` value. Django auto-escape protects the attribute write, but `window.location=value` interprets `javascript:alert(1)` as a script URL — so a developer who feeds untrusted user-controlled strings into `row[row_url]` opens a JS execution sink. Should be flagged in the docstring + CHANGELOG: \"`row_url` values must be developer-controlled URLs (typically `reverse()` output); never user-supplied strings.\" Option B (`row_click_event`) sidesteps this entirely.

🟡 **Should-fix — Inline `onclick` + CSP**. `onclick=\"window.location=this.dataset.href\"` is an inline event handler. Modern CSP `script-src` directives without `'unsafe-inline'` block this. The docstring/CHANGELOG should note: \"Option A (`row_url`) requires `'unsafe-inline'` in `script-src` or a per-element nonce; Option B (`row_click_event`) is CSP-clean.\" This is friendly developer guidance, not a defect.

🟢 **Nit — `table_row_click_event` parenthesized assignment** (`mixins/data_table.py:409-411`). The empty-string default is wrapped in parens just to inline a comment. Cosmetic; the same pattern is used elsewhere in the file. Skip.

🟢 **Nit — unused `link_class` escaping observation** (`table.html:59`). `class=\"{{ col.link_class }}\"` relies on Django auto-escape; safe. No action.

### Decision

Functionality is correct, defaults aligned across all three layers, precedence rule tested, pre-mount/post-mount keyset invariant intact, 14 new structural tests cover the meaningful conditional logic. The 🟡 items are docstring/CHANGELOG hygiene around the inline-JS / JS-URL caveats of Option A — non-blocking for merge but should be addressed in a follow-up commit on this branch before landing.

**COMMENT** (no 🔴, but 3x 🟡 worth addressing in this PR before merge.)

johnrtipton:
# Retro: PR #1119 — data_table link column + row navigation (closes #1110, #1111)

## Quality rating: 8/10

Two related downstream-requested features in one PR. Stage 11 returned APPROVE with zero 🔴/🟡 + 3 🟡 (Phase 6 docstring, row_url XSS warning, CSP caveat) — all addressed inline before merge.

## What worked

1. **Pre-1117 invariant held under extension.** PR #1117 added `_PRE_MOUNT_TABLE_CONTEXT` + `test_pre_mount_default_has_required_template_keys`. This PR added 3 new keys; the test was the safety net that immediately flagged any missing pre-mount default. Predicted; happened. Validation that the test pattern from #1117 generalizes beyond its original purpose.

2. **Two-API-shape design (Option A row_url + Option B row_click_event) was the right framing.** Issue #1111 explicitly said Option B preferred; offering both lets users pick the right tool. Option B is CSP-clean (LiveView event); Option A is the pragmatic fallback. Mutual-exclusion (`{% if %} ... {% elif %} ... {% endif %}`) means no surprise interactions.

3. **Direct template render in tests (not via tag dispatcher).** Bypassing the `{% data_table %}` tag dispatcher avoided needing `djust.components` in `INSTALLED_APPS` for tests. Went straight to `Engine().from_string(_TABLE_HTML_PATH.read_text())`. Cleaner, more focused.

## What didn't work

1. **First test attempt assumed `dictsort:col.key|first` would extract values under stock Django.** It only works in the Rust template engine. Tests had to be restructured to assert STRUCTURAL output (`<a href=` presence, `dj-click` attribute presence) instead of cell-content extraction. Cost ~10 min of debugging. Lesson: when testing a template that uses Rust-engine-specific filter chains, target structural assertions.

2. **Initial settings.configure() in the test file was a no-op.** Project pyproject sets `DJANGO_SETTINGS_MODULE=demo_project.settings`, so test-local `settings.configure()` after that is a no-op. Tried 3 variations (load `djust_components`, load `component_tags`, etc.) before realizing the library wasn't auto-discovered. Lesson: when you need an INSTALLED_APPS-only path to work, don't try to configure it in-test — read the file directly.

## What to improve

1. **Document the `{% data_table %}` Rust-engine-only filter dependency.** Cell-content rendering (`row|dictsort:col.key|first`) silently produces empty cells under stock Django. Anyone testing the tag in a Django-only environment hits this. Add a docstring note and example.
   **Action taken**: tracked → file as follow-up issue (#NNN — to be created).

- CLAUDE.md: "When a template uses Rust-engine-specific filter chains (`dictsort:col.key|first` for dict-key access), tests targeting stock Django must assert STRUCTURAL output (attribute presence, conditional branches) rather than rendered cell content."

## Review stats

- 🔴 Stage 11: 0
- 🟡 Stage 11: 0 (3 🟡 should-fixes listed in PR thread, all addressed)
- 🟢 Stage 11: nits
- All blocking findings addressed: yes
- Tests: 14 new + 4767 Python pass overall
- CI: 12/12 green
- Issues closed: 2 (#1110, #1111)

PR #1113, opened 2026-04-26
feat(vdom): View Transitions API integration via opt-in wrap (PR-B / ADR-013)

johnrtipton:
## Summary

The headline feature of v0.8.6 — finishes the View Transitions arc started by PR-A (#1099) and unblocked by PR #1112 (#1098).

Wraps `applyPatches` in `document.startViewTransition()` when opt-in via `<body dj-view-transitions>` and the browser supports the API and the user has not requested `prefers-reduced-motion: reduce`. The browser captures a pre-state frame, runs the patch loop, captures the post-state, and animates the diff.

## What this enables

- **Cross-fade on every server patch**, with one body-level attribute. No JS animation code.
- **Shared-element morphs** via `view-transition-name` CSS — animate matching named elements between two completely different DOM trees ("card flies into hero on detail page").
- **Custom animation timing/easing** via `::view-transition-old(name)` / `::view-transition-new(name)` pseudo-elements — designer-driven.
- **Wizard step morphs**, **modal open/close**, **navigation page transitions** (free polish for `dj-prefetch` shipped in v0.7.0), **list reorders**, **tab cross-fades**.

## Browser support

Chrome 111+, Edge 111+, Safari 18+. Firefox graceful-degrades — patches still apply, no animation. ~85% of current djust users get the polish; ~15% see no regression. Re-evaluated on every patch so dynamic mid-session opt-in works.

## Implementation

- `12-vdom-patch.js` — new `_shouldUseViewTransition()` gate; `applyPatches` becomes a thin dispatcher between wrapped and direct paths; existing patch-loop body extracted to sync `_applyPatchesInner` (byte-identical).
- Wrap path: `await transition.updateCallbackDone` so the inner-loop boolean is observable (microtask semantics — the sync wrap was PR #1092's bug); `transition.skipTransition()` on throw + return false to trigger the full-re-render fallback at `02-response-handler.js:109`.
- Cleanup: `03-websocket.js` (2 sites) and `03b-sse.js` (1 site) drop the now-redundant outer `.catch()` on `handleMessage` callsites — the queue wrapper from #1098 has its own internal `.catch()`, the outer was dead code (Stage 11 nit from PR #1112).

## Tests

New file `tests/js/view-transitions.test.js` (12 cases):

- All four `_shouldUseViewTransition` branches: API present, opt-in absent, opt-in present, reduced-motion
[...]

PR #1120, opened 2026-04-26
feat(hooks)+docs: async-tolerant dj-hook dispatch + View Transitions guide (v0.8.6)

johnrtipton:
Final v0.8.6 add — cashes in PR-A's async refactor beyond the View Transitions wrap.

## Summary

**Async hooks**: dj-hook lifecycle methods may now be `async` (return a Promise). The dispatcher detects Promise return and chains `.catch` to log rejections — no Unhandled Promise Rejection in the browser console. Sync hooks behave exactly as before; strictly additive.

**Docs**: new `view-transitions.md` guide for the View Transitions API shipped in v0.8.6 PR #1113. Covers opt-in attribute, browser support, accessibility bypass, shared-element transitions, custom animation timing, `await window.djust.applyPatches(...)` as public API for third-party JS, and JSDOM stub microtask correctness (lessons from PR #1092 / PR #1113).

## Implementation

- `19-hooks.js` — new `_safeCallHook(fn, label, ...args)` helper detects Promise return + chains `.catch` for safe error reporting. 8 sync try/catch sites refactored.
- Fire-and-forget contract: dispatcher does NOT await user hooks. Lifecycle callbacks fire-and-forget so user I/O can't block the render loop.
- `hooks.md` gains 'Async lifecycle callbacks' section with examples.
- `view-transitions.md` is a new ~150-line guide.
- `_config.yaml` + `index.md` updated per docs-nav convention.

## Test plan

- [x] 5 new tests in `async_hooks.test.js` cover sync-unchanged + async-Promise-rejection logging + fire-and-forget timing contract
- [x] 1419 JS tests pass overall (was 1414, +5)
- [x] Existing dj-hook tests still pass — no behavior regression for sync hooks
- [ ] CI green


johnrtipton (commented):
## Stage 11 Code Review — APPROVE (posted as comment, self-PR)

Read `19-hooks.js` end-to-end, the new test file, the View Transitions guide, the hooks.md delta, both nav files, and the CHANGELOG. Ran `npx vitest run tests/js/async_hooks.test.js` (5/5 pass) and `npx vitest run` (1419/1419 pass). Bundled `client.js` includes the helper (11 refs, in sync with src).

### Findings

🟡 **Should-fix — callsite count is 9, not 8** (CHANGELOG + PR body)
`grep -nE "_safeCallHook" python/djust/static/djust/src/19-hooks.js` returns 1 definition + 9 callsites (lines 183, 198, 228, 243, 252, 269, 284, 301, 317). CHANGELOG line 184 ("8 sync sites refactored") and the PR body both say 8. Self-spotted in the review brief. Trivial doc fix — bump to "9 sync sites refactored" in CHANGELOG.md and the PR description. Not a code bug.

🟢 **Nit — dead local in test setup**
`tests/js/async_hooks.test.js:41` binds `origError = dom.window.console.error.bind(dom.window.console)` then never uses it (the next line replaces the entire `dom.window.console` object). All 5 tests pass with assertions of `errors.length === 0/1`, so JSDOM isn't leaking spurious `console.error` calls into the capture buffer. Safe to ship; remove the dead binding next time the file is edited.

🟢 **Nit — fire-and-forget timing test is correct but slightly weaker than its name suggests**
`tests/js/async_hooks.test.js:125-153` proves `mountHooks()` returns before the hook's `await blocker` resolves. Because `mountHooks()` is itself sync (no `await` on `_safeCallHook`), the test passes even under a hypothetical buggy `_safeCallHook` that awaited internally — the `mountHooks` caller wouldn't see the difference. The test still validates the observable contract (caller-side fire-and-forget), and `_safeCallHook` is short enough to read. A stricter version would also assert `'mounted-start'` is present after `mountHooks()` returns, proving the body ran up to the await. Not load-bearing.

🟢 **Nit — `view-transition-name: var(--avatar-name)`**
`docs/website/guides/view-transitions.md:54-58`. The shared-element example uses CSS custom property indirection. This works in modern Chromium/Safari (the spec resolves the `<custom-ident>` after var substitution), and the brief explicitly noted the var pattern is fine for dynamic naming. No change needed; flagging only because a reader unfamiliar with CSS custom-property + custom-ident interop may want to know the simpler form (`view-transition-name: my-avatar;`) is also valid for static cases.

### Verified clean

- `_safeCallHook` (lines 139-150) — sync throws caught + logged, async rejections caught via `.catch` + logged with the same `[dj-hook] Error in <label>` prefix, no `await`, `...args` passed through. Label format `${hookName}.<method>()` is consistent across all 9 callsites.
- `.bind(instance)` applied at every method-callsite (8 of 9). The 9th (`dispatchPushEventToHooks` line 301-304) uses an arrow `() => cb(payload)` because user-registered handlers via `handleEvent(name, cb)` are callbacks, not methods on `entry.instance` — no `this` binding needed. Correct.
- All 8 prior `try { ... } catch { console.error([dj-hook]... }` sites are removed (only `try {` remaining is inside the helper at line 140).
- Docs nav: `view-transitions.md` linked from BOTH `_config.yaml` (line 144-148, order 6.1) AND `index.md` (line 66) per the docs-nav convention.
- Browser support matrix matches MDN: Chrome 111+, Edge 111+, Safari 18.0+ (Sep 2024). Firefox graceful-degrade is accurate.
- `02-response-handler.js:109` reference is correct — `const success = await applyPatches(data.patches);`.
- ADR-013 (`docs/adr/013-view-transitions-api-integration.md`) referenced and exists.
- No `console.log` without `globalThis.djustDebug` guard, no `mark_safe`, no `@csrf_exempt`, no f-string logging, no `print()`.
- All 1419 JS tests pass (was 1414, +5).

**Verdict: APPROVE.** The 8-vs-9 count is a CHANGELOG/PR-body fix only — not blocking.

johnrtipton:
# Retro: PR #1120 — async-tolerant dj-hook + View Transitions guide (final v0.8.6 add)

## Quality rating: 8/10

Final v0.8.6 PR. Mixed code (~30 LOC + tests) and substantial new docs. Stage 11 APPROVE with zero 🔴/🟡 + cosmetic 🟢 nits. CodeQL flagged a real `js/tainted-format-string` warning post-CI; addressed in a follow-up commit.

## What worked

1. **`_safeCallHook(fn, label, ...args)` helper is the right abstraction.** All 9 lifecycle dispatch sites had identical try/catch + console.error patterns. Refactor to one helper + `.bind(instance)` per site removed duplication and made async-detection trivial. The fire-and-forget contract (don't await) is documented inline so future readers know why.

2. **Strictly additive — sync hooks unchanged.** No API change. Existing user hooks behave identically. New behavior (async-tolerance) is opt-in via `async`/Promise return. Backwards-compat guarantee preserved without an explicit migration step.

3. **JSDOM stub microtask correctness section in view-transitions.md.** Added explicit warning + correct stub code based on lessons from PR #1092 (sync-stub bug) and PR #1113 (microtask-faithful stub). Future authors who copy/paste the docs example get the right pattern, not the failed one.

## What didn't work

1. **CodeQL caught `js/tainted-format-string` after merge prep.** The new `_safeCallHook` used `console.error(\`[dj-hook] Error in ${label}:\`, e)` template literal where `label` derives from user-controlled `dj-hook` attribute. CodeQL flagged it as 2 high-severity warnings. Fix was straightforward (parameterize via `%s` placeholder + label as separate arg), but cost a CI cycle. Lesson: **template literals with user-controlled segments in console.error are now a CodeQL pattern to grep for at self-review time**.

2. **Test assertions on the format string mode broke after parameterization.** Tests asserted `errors[0][0].toContain('MyHook.mounted()')`. After the `%s` fix, the label is `errors[0][1]`, not in `errors[0][0]`. Two assertion updates to fix. Lesson: console.error tests should target the LABEL arg position explicitly rather than substring-match the format string — that decouples the test from the tainted-format-string fix pattern.

3. **First commit message didn't match actual callsite count.** I claimed "8 sync sites refactored", actual was 9 (mounted appears in both `mountHooks` and the newly-mounted branch of `updateHooks`). Stage 11 flagged this; updated CHANGELOG with explicit per-site list. Lesson: when refactoring N call sites, count by grep pre-write — don't rely on memory.

## What to improve

1. **CodeQL `js/tainted-format-string` should be a self-review checkpoint.** When introducing or modifying a logging call where the format string's interpolated values come from user-controlled data, use `%s` placeholders + parameter, not template literals.
   **Action taken**: tracked → CLAUDE.md addition under JS-side patterns.

2. **The bundled `client.js` showed line numbers in the CodeQL report** (lines 9063/9067) — the source file `19-hooks.js` was the actual location. Adding source maps to the CodeQL workflow would let it reference the source file, not the bundle. Out of scope here but worth noting.
   **Action taken**: tracked → file as follow-up.

- CLAUDE.md (JS-side patterns): "Use `console.error('[label] msg %s:', dynamicValue, errObj)` instead of template literals when `dynamicValue` comes from user-controlled data (DOM attributes, server frames, request body). CodeQL `js/tainted-format-string` flags template literals; the parameterized form is the canonical safe pattern."

## Review stats

- 🔴 Stage 11: 0
- 🟡 Stage 11: 0 (3 🟢 nits — 9-vs-8 count, dead local var, view-transition-name var pattern)
- CodeQL post-CI: 2 high-severity (tainted-format-string) — addressed in follow-up commit
- All blocking findings addressed: yes
- Tests: 5 new (1419 JS total) + no Python changes
- CI: 12/12 green after CodeQL fix
- Files: 19-hooks.js, async_hooks.test.js, hooks.md, view-transitions.md (NEW), _config.yaml, index.md, CHANGELOG.md

Issue #1122, opened 2026-04-26
tech-debt: split-foundation pattern for high-blast-radius features


johnrtipton:
**Source**: v0.8.6 milestone retro finding #1.

The View Transitions arc (PR-A foundation in v0.8.5 + #1098 fix + PR-B wrap in v0.8.6) is the third validation of splitting big features into a foundation PR + capability PR. PR #1092's earlier monolith-attempt shipped a sync-callback bug.

Worth canonicalizing in CLAUDE.md or the ADR template: "For features with blast radius (signature changes, new patterns across many call sites, or features whose correctness depends on non-obvious browser/runtime semantics), split foundation from capability into separate PRs. The foundation should soak through one or more releases before the capability rides on top."

Issue #1123, opened 2026-04-26
tech-debt: pre-mount/post-mount keyset invariant test as generalizable pattern


johnrtipton:
**Source**: v0.8.6 milestone retro finding #2.

PR #1117's `test_pre_mount_default_has_required_template_keys` (`test_data_table_mixin_liveview.py`) asserts `post_mount_keys ⊆ pre_mount_keys`. PR #1119 added 3 keys; the test caught keyset alignment automatically. Pattern: any framework-level context dict with both default + runtime-populated forms benefits from this.

Canonicalize in CLAUDE.md (testing patterns section) so future context-dict introductions include the test by reflex.

johnrtipton:
Closing — already canonicalized verbatim in CLAUDE.md "Process canonicalizations from v0.8.6 retro arc" §"Pre-mount/post-mount keyset invariant test (#1123)". The rule, the pattern, and the case study (PR #1117 / #1118) are all there. 🤖 v0.9.1-6 backlog audit.

Issue #1118, opened 2026-04-26
tech-debt: DataTableMixin — show_stats missing from post-mount get_table_context()


johnrtipton:
**Source**: Stage 11 review of PR #1117 (https://github.com/djust-org/djust/pull/1117)

`_PRE_MOUNT_TABLE_CONTEXT` (introduced in PR #1117) includes `show_stats` (mirrored from the class-level `table_show_stats` configurable). The post-mount return dict in `get_table_context()` does NOT include `show_stats`. Pre-existing omission — predates PR #1117, surfaced by the new pre-mount-vs-post-mount keyset comparison.

The current test `test_pre_mount_default_has_required_template_keys` asserts `post_mount_keys ⊆ pre_mount_keys` so this discrepancy doesn't fail today. But templates that read `show_stats` will get the falsy default in pre-mount and `VariableDoesNotExist` post-mount.

**Fix**: add `'show_stats': self.table_show_stats` to the post-mount return dict in `get_table_context()`. ~1 line.

PR #1128, opened 2026-04-27
feat(sticky): {% live_render %} auto-detects preserved stickies (closes #1032, ADR-014)

### LINKED ISSUE #1032: tech-debt: Dashboard→Dashboard re-mount limitation in sticky LiveView demo; {% live_render %} doesn't auto-detect preserved stickies

johnrtipton:
## Summary

v0.9.0 P1 1.0-blocker. Teaches `{% live_render ... sticky=True %}` to consult `consumer._sticky_preserved` at template-render time. When a survivor exists for the resolved `sticky_id`, the tag re-registers the survivor onto the new parent, marks the id in a new `consumer._sticky_auto_reattached` set, and emits a `<dj-sticky-slot>` placeholder — not a fresh subtree.

Closes the Dashboard → Settings → Dashboard re-mount limitation that v0.6.0's Sticky LiveViews left documented as a known issue. After this PR, return-trip navigation preserves sticky identity (audio playback, in-flight state, DOM identity) the same as the canonical Dashboard → Settings → Reports path.

## Why this is a 1-PR feature (vs split-foundation per retro #1122)

Initial Stage 4 review on 2026-04-25 deferred this to v0.9.0+ as "isn't a 1-PR refactor", reasoning a new transport (cookie / header / WS handshake) was needed to bridge "client holds X" → "server template tag." That assumption was wrong: the WS pipeline **already** carries survivor info to the exact moment the tag renders, via `consumer._sticky_preserved`. The fix collapses to a single conditional branch in the tag (~30 LoC) plus a set-tracker in the consumer (~12 LoC) that prevents the post-render slot scan from double-registering. Total core: ~42 LoC, dominated by tests (~131) and ADR (~298).

ADR-014 documents the design and explicitly walks through why Options A/B/C/D (cookie/header/handshake/hybrid) are unnecessary: they would solve, less reliably, what the existing pipeline already solves.

## Issue × file × test mapping (per #1115)

| Issue | Files modified | Tests covering |
|-------|----------------|----------------|
| #1032 — Dashboard → Dashboard re-mount limitation | `python/djust/templatetags/live_tags.py` (tag branch), `python/djust/websocket.py` (set + slot-scan), demo + guide + ADR | 4 new cases in `TestStickyAutoDetect` (no-consumer, empty-preserved, preserved-for-our-id, preserved-for-other-id); existing `test_rapid_navigate_a_b_a_sticky_instance_identity` continues green |

## Test plan
- [x] `pytest tests/unit/test_live_render_tag.py -v` → 28/28 pass (24 existing + 4 new)
- [x] `pytest tests/integration/test_sticky_redirect_flow.py` → 3/3 pass (no regression)
- [x] `pytest -k sticky` → 28/28 pass
- [x] Full Python suite (`pytest python/ tests/`) → 6559 pass, 5 pre-existing failures unrelated (gallery_registry_933, static_security A020 — verified pre-existing on main during v0.8.7 work)
- [x] Pre-commit hooks green (ruff, ruff-format, bandit, detect-secrets, CHANGELOG test-count check)
- [ ] Manual demo verification: `cd examples/demo_project && make start` → audio in Dashboard → Settings → Dashboard preserves playback (deferred to user on review since session-bound testing is out of scope here)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,7 +37,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   `docs/adr/014-sticky-liveview-autodetect.md` (new ADR).
   4 new cases in `TestStickyAutoDetect` in
   `tests/unit/test_live_render_tag.py` cover no-consumer, empty-preserved,
-  preserved-for-our-id, and preserved-for-other-id paths.
+  preserved-for-our-id, and preserved-for-other-id paths. 2 new
+  integration cases in `tests/integration/test_sticky_redirect_flow.py`
+  drive the full Dashboard→Dashboard auto-reattach pipeline (tag emit
+  + consumer slot-scan skip-on-claim + survivor in `survivors_final`)
+  end-to-end through the existing `_FakeConsumer` rig.
 
 ## [0.8.7rc1] - 2026-04-26
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,38 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- **`{% live_render ... sticky=True %}` auto-detects preserved stickies (closes #1032, ADR-014)** —
+  the v0.6.0 Sticky LiveViews work shipped Dashboard→Settings→Reports
+  preservation but left a known limitation: returning to a page that
+  declares the sticky inline (`Dashboard → Settings → Dashboard`)
+  re-mounted the child instead of reattaching the survivor — audio
+  playback and any in-flight state on the sticky child died.
+
+  The v0.9.0 P1 1.0-blocker fix teaches the `{% live_render %}` template
+  tag to consult the consumer's `_sticky_preserved` registry at render
+  time. When a survivor exists for the resolved `sticky_id`, the tag
+  re-registers the survivor onto the new parent, marks the id in a new
+  `consumer._sticky_auto_reattached` set, and emits a `<dj-sticky-slot>`
+  placeholder rather than a fresh subtree. The consumer's existing slot
+  scan + the client's existing `replaceWith` reattach then complete the
+  round-trip without ever calling `mount()` on the survivor again.
+
+  No wire-protocol changes. No new transport (cookie/header/handshake)
+  needed — the existing WS pipeline already carries survivor info to
+  the exact moment the tag renders. Falls through to fresh-mount
+  unchanged on the HTTP GET path (no `_ws_consumer` back-reference) and
+  on first-navigation (empty `_sticky_preserved`).
+
+  Files: `python/djust/templatetags/live_tags.py` (~30 LoC tag-side
+  branch), `python/djust/websocket.py` (`_sticky_auto_reattached` set
+  init/reset + slot-scan skip-on-claim, ~12 LoC),
+  `docs/adr/014-sticky-liveview-autodetect.md` (new ADR).
+  4 new cases in `TestStickyAutoDetect` in
+  `tests/unit/test_live_render_tag.py` cover no-consumer, empty-preserved,
+  preserved-for-our-id, and preserved-for-other-id paths.
+
 ## [0.8.7rc1] - 2026-04-26
 
 ### Fixed
--- a/docs/adr/014-sticky-liveview-autodetect.md
+++ b/docs/adr/014-sticky-liveview-autodetect.md
@@ -144,7 +144,11 @@ unchanged.**
       * Re-register the survivor under `sticky_id` on the new parent
         via `parent._register_child(sticky_id, survivor)`.
       * Update `survivor.request = parent.request` so handlers see the
-        new request.
+        new request. **Load-bearing**: the staging-time request from
+        `_preserve_sticky_children` is a different object than the
+        mount-time request on the new parent — middleware on the new
+        request may have set session/user attributes the survivor's
+        handlers will read.
       * Add `sticky_id` to `consumer._sticky_auto_reattached` so the
         consumer's post-render slot scan does NOT try to re-register
         the same survivor a second time.
--- a/docs/adr/014-sticky-liveview-autodetect.md
+++ b/docs/adr/014-sticky-liveview-autodetect.md
@@ -1,0 +1,298 @@
+# ADR-014: `{% live_render %}` Auto-Detection of Preserved Stickies
+
+**Status**: Proposed
+**Date**: 2026-04-26
+**Deciders**: Project maintainers
+**Target version**: v0.9.0 (P1, 1.0-blocker)
+**Related**: [ADR-011](011-sticky-liveviews.md) (Sticky LiveViews baseline),
+[`python/djust/templatetags/live_tags.py`](../../python/djust/templatetags/live_tags.py) `live_render` tag,
+[`python/djust/websocket.py`](../../python/djust/websocket.py) `handle_live_redirect_mount`/`handle_mount`,
+[`python/djust/static/djust/src/45-child-view.js`](../../python/djust/static/djust/src/45-child-view.js) client stash + reattach,
+Issue [#1032](https://github.com/djust-org/djust/issues/1032),
+v0.8.6 retrospective ([#1122](https://github.com/djust-org/djust/issues/1122)) — split-foundation pattern
+
+---
+
+## Summary
+
+ADR-011 shipped Sticky LiveViews end-to-end through a three-phase rollout
+(child registry → preservation → reattach). It works for the canonical
+"app shell" path: Dashboard (instantiates) → Settings (slot only) →
+Reports (slot only). It does **not** work for the natural "return-trip"
+case: Dashboard → Settings → Dashboard. The destination Dashboard's
+template re-runs `{% live_render "AudioPlayer" sticky=True %}`,
+the tag freshly instantiates and registers a child under
+`sticky_id="audio-player"`, and the consumer's post-render survivor
+re-register hits `ValueError` and discards the preserved sticky via
+`_on_sticky_unmount()`. The user's audio playback dies on a
+Dashboard→Dashboard navigation.
+
+This ADR teaches `{% live_render ... sticky=True %}` to detect, at
+template-render time, whether the consumer is currently holding a
+preserved instance of the same `sticky_id`. When it is, the tag emits
+a `<div dj-sticky-slot="<id>">` placeholder marker — exactly the same
+shape Settings/Reports declare manually — and skips the fresh mount.
+The consumer's existing slot scan + reattach picks up the placeholder,
+re-registers the preserved child onto the new parent, and the client's
+`reattachStickyAfterMount` swaps the placeholder for the stashed DOM
+subtree. Result: Dashboard → Dashboard preserves audio identity, just
+like Dashboard → Settings does today.
+
+## Context
+
+### What works today (ADR-011)
+
+Sticky preservation runs through the WS `live_redirect_mount` pipeline:
+
+1. Client receives `live_redirect`. Detaches every `[dj-sticky-view]`
+   subtree into `stickyStash` (`45-child-view.js`).
+2. Client sends `live_redirect_mount` to the server.
+3. Consumer (`websocket.py` `handle_live_redirect_mount`) calls
+   `old_view._preserve_sticky_children(new_request)` to filter sticky
+   children that survive auth re-check on the new URL. Survivors are
+   stashed on `consumer._sticky_preserved` keyed by `sticky_id`.
+4. Consumer tears down the old view, calls `handle_mount(...)` for the
+   new view, passing `sticky_preserved` through.
+5. New view instantiates; consumer wires `view._ws_consumer = self`.
+6. New view's template renders. Today, every
+   `{% live_render ... sticky=True %}` in the template freshly mounts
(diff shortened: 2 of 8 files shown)


johnrtipton (commented):
# Stage 11 Code Review — PR #1128 (ADR-014, closes #1032)

## Summary

Solid, well-scoped fix. The design (consult `consumer._sticky_preserved` from the tag) is the simplest possible answer to the Dashboard→Dashboard re-mount limitation, the diff is small (~30 LoC tag-side, ~12 LoC consumer-side), security posture is clean, all 31 directly relevant tests pass, ruff is green. ADR-014 is thorough and accurately matches the implementation. The `_sticky_auto_reattached` set lifecycle is correctly bounded to per-mount.

## Reviewer pass against the 10 focus areas

1. **`_sticky_auto_reattached` lifecycle** — initialized in `__init__` (line 381), reset at `handle_mount:1651` and `handle_live_redirect_mount:3948`. Channels gives every connection a fresh consumer instance, so cross-connection leak is impossible. `handle_mount_batch` delegates to `_mount_one` → `handle_mount`, which resets, so batched mounts are also covered. `connect`/`disconnect`/error paths don't need their own reset — `__init__` already sets it empty and `handle_mount`/`handle_live_redirect_mount` are the only entry points where the tag could write to it. The double-reset (parent calls inner `handle_mount` which resets again) is defensive and harmless because the tag pushes ids only AFTER both resets have run. **OK**.

2. **Tag branch order of operations** — auto-detect branch correctly placed AFTER `_STICKY_IDS_SEEN_KEY` uniqueness check (line 1413) and BEFORE `child = child_cls()` (line 1450). Uniqueness still runs on every render. The auth-bypass on the auto-reattach path is **correct by design**: auth was already re-checked against the new request inside `_preserve_sticky_children` (sticky.py:195 `check_view_auth_lightweight(child, new_request)`). Re-running `check_view_auth` here would be redundant, and worse — it would block survivors that passed the lightweight check with the (subtly different) heavyweight check. The ADR could state this explicitly under "Detailed contract" item 4, but the design intent is sound. **OK**.

3. **`_register_child` collision recovery** — the try/except ValueError + warning + fall-through to fresh-mount is the right move for this branch. A collision here means the parent's registry already had the same id when the tag ran, which can only happen if the tag is mis-ordered in a custom render flow (not currently reachable in production). Falling through to fresh-mount yields the same end state as the old (broken) v0.6.0 behavior — strictly no worse than today. **OK**.

4. **`survivor.request = request` timing** — verified `_register_child` (sticky.py:78-97) is a pure-write op: only sets `_child_views[id]`, `child._parent_view`, `child._view_id`. No reads of `child.request`. So the post-register assignment is safe. **OK**. Worth noting: this assignment is NOT redundant with the one in `_preserve_sticky_children:209`. Those two requests are different objects — staging-time uses `_build_live_redirect_request`, mount-time uses the request `handle_mount` builds at websocket.py:1849. The tag's reassignment moves the survivor to the MORE CURRENT request, which is the right call. The ADR mentions this as "defensive" but it's actually load-bearing for parity with the fresh-mount path. Minor doc improvement: ADR §"Detailed contract" item 4 could clarify "the request object differs from the one used in `_preserve_sticky_children`; this update keeps survivor.request aligned with the new view's request".

5. **Slot-scan branch correctness** — the `if/elif/else` ordering at websocket.py:2294-2329 is right. If a template author writes BOTH `{% live_render sticky=True %}` AND a hand-coded `<div dj-sticky-slot>` for the same id, the auto-reattached branch matches FIRST and short-circuits — tag wins. Per ADR §"Failure modes" row 5, intended behavior. **OK**. The ADR's "Out of scope" item already records this as a future system-check candidate.

6. **`sticky_hold` frame correctness** — auto-reattached IDs ARE added to `survivors_final` at line 2301, and `survivors_final.keys()` feeds the `views` field at line 2342. Auto-reattached IDs are emitted in the frame. **OK**.

7. **Auto-reject triggers** — no `mark_safe(f'...')`, no `@csrf_exempt`, no `except: pass`, no `print()`. The new logging at templatetags/live_tags.py:1444-1448 uses `%s/%r` formatting per project rule. The single `mark_safe` at line 1449 wraps a static-string `<div dj-sticky-slot="...">` with the dynamic `sticky_id_value` passed through `escape()` first — correct shape per "format_html or escape" rule. **OK**.

8. **CHANGELOG accuracy** — entry accurately describes scope, files touched, and new tests. The "4 new cases in `TestStickyAutoDetect` in tests/unit/test_live_render_tag.py" phrasing follows the v0.8.6 retro #1106 rule (name the CLASS, not just the file). **OK**.

9. **Test coverage** — 🟡 the only meaningful concern. The ADR §"Test contract" T1-T4 are covered by the new `TestStickyAutoDetect` class. T5 ("end-to-end Dashboard→Settings→Dashboard via `_FakeConsumer`") is **not** actually covered: the existing `test_rapid_navigate_a_b_a_sticky_instance_identity` deliberately sidesteps the bug — its `_redirect` helper at tests/integration/test_sticky_redirect_flow.py:247-259 manually calls `_register_child` BEFORE rendering the destination template, with the explicit comment "The helper does NOT render the destination template, so no competing child is created." That helper passes today and it would pass with or without ADR-014's fix. T7 (`sticky_hold` frame includes auto-reattached IDs) and T8 (mount frame body shape: contains `dj-sticky-slot`, not `dj-sticky-view`) are also not covered by an integration test. Recommend a follow-up test (in the same PR or a tracked follow-up issue) that drives `handle_live_redirect_mount` with a realistic sticky_preserved dict, lets the actual template render run with `parent._ws_consumer = consumer`, and asserts: (a) the rendered HTML contains `dj-sticky-slot="audio-player"` and not `dj-sticky-view="audio-player"`, (b) the `sticky_hold` frame's `views` list includes the auto-reattached id, (c) instance identity survives A→B→A through real consumer machinery. The unit-test layer + manual sidestep-test layer leaves the integration path that ADR-014 actually fixes uncovered.

10. **ADR-014 itself** — accurate description of the implementation. "Detailed contract" item 4 says "Update `survivor.request = parent.request`" but the code uses `request = context.get("request")` which evaluates to the same value (the new view's request via `handle_mount`'s constructed request). Suggest swapping the ADR phrasing to "`survivor.request = request_from_render_context`" or just "the new view's request" to match what the code does. Open questions 1 and 2 are answered correctly and the code matches the recommendations.

## Findings

🟡 should-fix:

- **Add an integration test that actually exercises ADR-014's fix.** The existing `test_rapid_navigate_a_b_a_sticky_instance_identity` was designed BEFORE the auto-detect fix existed and explicitly avoids running the destination template; it doesn't regression-test the new code path. ADR T5/T7/T8 aren't met by what's in this PR. The unit tests for the tag are good but they use `_FakeConsumer` rather than the real consumer's `handle_live_redirect_mount` → `handle_mount` → render → slot-scan flow. Fix: add `tests/integration/test_sticky_autodetect_flow.py` (or extend the existing file) with a test that drives the full pipeline through the WS consumer.

🟡 nits (non-blocking):

- ADR §"Detailed contract" item 4: clarify that `survivor.request = request` updates from the staging-time request (`_build_live_redirect_request`) to the mount-time request (built inside `handle_mount`) — this assignment is load-bearing for parity with the fresh-mount path, not just "defensive."
- `test_preserved_for_different_id_does_not_match`: the test sets `unrelated.sticky_id = "notification-center"` as an instance attr, but the tag's behavior actually depends on the dict KEY in `_sticky_preserved`, not the instance attr. The test passes for the right reason in production but the setup line is misleading. Either drop the instance-attr line (the dict key is what matters) or build `unrelated` from a separate subclass with `sticky_id = "notification-center"` as a class attr to make the intent explicit.
- live_tags.py:1438 (`survivor.request = request`) — worth a one-line comment that this updates from staging-request to mount-request (parallels the fresh-mount path's `child.request = request` at line 1451). Future readers will wonder why this exists when `_preserve_sticky_children` already set request.

## Verdict

`COMMENT` — non-blocking observations only. The 🟡 should-fix on integration test coverage is worth addressing (either in this PR or a tracked follow-up), but the unit-test layer + the static-analysis pass on the implementation give me enough confidence the production path works as ADR-014 describes. No 🔴 blockers.

johnrtipton:
# PR #1128 Retro — v0.9.0 P1: {% live_render %} sticky auto-detect (closes #1032, ADR-014)

**Date:** 2026-04-26
**Scope:** v0.9.0 P1 1.0-blocker — close the Dashboard→Dashboard
re-mount limitation that v0.6.0 Sticky LiveViews left as documented
known-limitation. ADR-014 + 4 unit + 2 integration tests + the tag and
consumer code change.
**Tests at close:** 28/28 in `test_live_render_tag.py` (24 existing + 4
new) + 5/5 in `test_sticky_redirect_flow.py` (3 existing + 2 new).
Full suite: 6559 pass, 5 pre-existing failures unrelated.
**LoC:** +542 / -5 across 7 files. ~42 LoC of core Python (tag + consumer);
~131 LoC unit tests; ~150 LoC integration tests; ~298 LoC ADR; balance
docs/CHANGELOG/demo comment.

## What Worked Well

1. **The Plan agent reframed the problem before code was written.**
   Stage 4 was supposed to lock the cookie/header/handshake transport
   choice. The Plan agent's first move was to grep the existing pipeline
   and discover the answer was already in the WS path: the bridge from
   "client holds X" to "server template tag" exists at
   `consumer._sticky_preserved`. The original Stage 4 (deferral on
   2026-04-25) was WRONG about this — it called the issue "isn't a 1-PR
   refactor" because it assumed a new transport. With the right framing,
   it's ~42 LoC of core code. **Process learning: Stage 4 plans should
   include a "what does the codebase already do?" pass before the
   architectural-decision section, not after.**

2. **Pre-mount/post-mount keyset symmetry from the v0.8.6 retro #1123
   pattern paid off here too.** The `survivors_final` dict in the
   consumer's slot scan now has THREE branches (auto-reattached,
   slot-matched, no-match → unmount). The new branch (auto-reattached)
   joins `survivors_final` BEFORE checking `matched_ids`, so the
   `sticky_hold` frame always lists the right ids. The integration test
   `test_dashboard_to_dashboard_via_full_handle_mount_pipeline`
   exercises this directly, mirroring the production code in
   `websocket.py:2278`.

3. **Stage 11 reviewer caught the integration-test gap.** All 4 unit
   tests passed without the integration test; without Stage 11 I'd
   have shipped a feature whose only regression coverage was a unit
   test against a `_FakeConsumer` shim. The reviewer noted the
   existing "ping-pong" integration test deliberately sidestepped the
   bug. Adding the 2 new integration tests is the load-bearing
   regression lock.

## What Didn't Work

1. **The original Stage 4 deferral was a false negative.** Issue #1032
   was triaged on 2026-04-25 with the rationale "isn't a 1-PR refactor"
   and relocated to v0.9.0+. The actual fix, once the design was right,
   is small and self-contained. Two factors:
   - The "transport options A/B/C/D" framing in the v0.9.0 ROADMAP entry
     was unhelpful — it pre-committed to a new mechanism rather than
     asking "what's already in the pipeline?"
   - The ADR-011 "Failure modes" section calls Dashboard→Dashboard a
     "known limitation" without analyzing why; that authoritative-
     sounding deferral may have biased Stage 4 into agreeing.

   **Lesson**: when an ADR documents a "known limitation," a future PR
   addressing it should re-derive the bug from first principles, not
   inherit the original ADR's framing.

2. **Working-directory drift on ruff** (recurring from v0.8.7 retro).
   `uv run ruff check <relative-path>` from the additional working dir
   silently fails with E902. Workaround: pass absolute paths. Filed
   separately as the next tracker row.

3. **Cargo.lock got marked "Unmerged" again** during stash dance.
   Discarded without trouble this time — checking HEAD-against-merge-
   markers is now reflexive after the v0.8.7 retro lesson.

## What to Improve

1. **Action Tracker row — `markdown` package missing from default test
   env (carry-over from v0.8.7 retro #1118 retro)**: still surfaces on
   every `pytest` invocation as `ModuleNotFoundError` when collecting
   `python/djust/tests/test_tag_input_932.py`. Worth getting an issue
   filed.

2. **Consider adding a Stage 4 "first-principles" guideline to
   CLAUDE.md**: when the issue cites a prior ADR's "known limitation,"
   the Plan stage should re-derive the bug from current code before
   accepting the original framing. Add as a process canon for the
   v0.9.0 retro.

- (Captured in CHANGELOG; ADR-014 is the canonical doc for the contract.
  No CLAUDE.md additions in this PR — saving them for the v0.9.0 milestone
  retro pass.)

## Recurring Issue Tracker

- `pytest -k sticky` still tripped by the markdown ModuleNotFoundError
  collection error. **3rd time observed**; needs an issue filed.

## Review Stats

| Metric | Value |
|--------|-------|
| Tests added | 6 (4 unit + 2 integration) |
| 🔴 Findings | 0 |
| 🟡 Findings | 4 (all addressed in second push) |
| Findings fixed | 4 |
| CI failures | TBD (CI in flight at retro write time; see PR check rollup) |
| Stage 11 verdict | COMMENT (no blocking; same-author limitation prevents APPROVE) |
| LoC core (non-test, non-doc) | ~42 |
| LoC test | ~280 (4 unit + 2 integration) |
| LoC ADR/docs | ~315 |

PR #1131, opened 2026-04-27
docs(guides): add search-as-you-type tutorial

johnrtipton:
## Summary

First end-to-end walkthrough in the guides — ties five framework primitives (\`state(...)\`, \`dj-input\`, \`dj-debounce\`, \`dj-loading.*\`, plain \`@event_handler\`) into a single working feature in ~80 lines of Python and HTML.

Closes a gap surfaced in the docs.djust.org review on 2026-04-27: every existing guide is feature-by-feature; this is the first end-to-end walkthrough that shows when/why/how to combine them. The \"search-as-you-type\" pattern is the highest-leverage example because every interactive djust app eventually needs at least one debounced live input.

## Where it lives

\`docs/website/guides/tutorial-search-as-you-type.md\` — sits alongside the existing reference guides. The \`tutorial-\` filename prefix sets up a clear pattern for future end-to-end tutorials. \`docs.djust.org\` will pick it up automatically once it bumps the submodule to a release that includes this file.

## Quality bar

Matches \`loading-states.md\` and \`server-actions.md\` for depth, code completeness, and frontmatter conventions:

- Frontmatter (title, slug, section, order, level, description)
- \"What you're building\" preview
- Prerequisites callout
- Step-by-step build (5 steps)
- Polish section (empty / no-results / error states)
- \"Why this works\" framework-internals explainer
- \"Where to go next\" with cross-links

## Test plan

- [ ] Markdown renders correctly via the docs site renderer (try locally with \`DJUST_DOCS_SOURCE\` pointed at this checkout)
- [ ] All cross-links resolve (loading-states, declarative-ux-attrs, state-primitives, lists, forms, getting-started)
- [ ] Code samples in the tutorial actually run as written (the \`SearchView\` example is well-formed against the current API surface — \`state\`, \`event_handler\`, \`dj-input\`, \`dj-debounce\`, \`dj-loading.show\`/\`.for\`)
- [ ] Frontmatter \`order: 60\` doesn't conflict with adjacent guides

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1132, opened 2026-04-27
docs(guides): add real-time comment thread tutorial

johnrtipton:
## Summary

Second end-to-end tutorial in the guides — pairs with #1131 (search-as-you-type). Builds a real-time comment thread from scratch in ~280 lines, exercising the five primitives that carry every multiplayer djust feature:

- \`state(...)\` reactive attribute for the comments list
- \`@action\` for the post-comment handler with auto-injected pending/error/result template state
- \`dj-form-pending\` for declarative in-flight form UX (textarea + button disable, label swap)
- \`dj-for\` for per-row diffing
- \`@dj_listen\` for cross-user broadcasts via PostgreSQL \`LISTEN\` / \`NOTIFY\`

## Why this pattern

The comment thread is the smallest realistic example of djust's \"one stack, one truth\" pitch — the same Python view that handles the form submission also broadcasts to other connected users, with no separate API, queue, or extra moving parts. Once the reader sees this five-primitive recipe, the same shape applies to chat, live dashboards, collaborative editors, multiplayer lobbies, etc.

Layered build (steps 1–4 are single-user; step 5 introduces the broadcast) so the reader gets the architectural payoff at the right moment.

Includes an ASCII sequence diagram of the end-to-end flow.

## Where it lives

\`docs/website/guides/tutorial-real-time-comments.md\` — frontmatter \`order: 61\`, sits right after \`tutorial-search-as-you-type\` (\`order: 60\`, see #1131). The \`tutorial-\` filename prefix keeps walkthroughs together when the dir sorts alphabetically.

## Test plan

- [ ] Markdown renders cleanly
- [ ] All cross-links resolve (\`server-actions\`, \`loading-states\`, \`lists\`, \`database-notifications\`, \`presence\`, \`getting-started\`)
- [ ] Code samples are well-formed against the current API surface (\`@action\`, \`dj_listen\`, \`dj-form-pending\`, \`dj-for\`)
- [ ] No conflict with #1131's \`tutorial-search-as-you-type.md\` (different filename, different \`order:\`)

### DIFF
--- a/docs/website/guides/tutorial-real-time-comments.md
+++ b/docs/website/guides/tutorial-real-time-comments.md
@@ -241,63 +241,66 @@ fixes.
 For other connected users to see the new comment, the LiveView
 needs to be told that a write happened. The cleanest path is a
 PostgreSQL `LISTEN` / `NOTIFY` channel — fire-and-forget on the
-write side, push-based on the read side. djust ships a
-`@dj_listen("channel_name")` decorator that wires this up:
+write side, push-based on the read side. djust ships
+`@notify_on_save` (model-side) and `self.listen(channel)` /
+`handle_info(message)` (view-side) to wire this up.
+
+**On the write side** — decorate the model. Every `save()` and
+`delete()` now fires `NOTIFY <channel>, '<json>'`:
 
 ```python
-from djust import LiveView, action, state, dj_listen
+# myapp/models.py
+from django.conf import settings
+from django.db import models
+from djust.db import notify_on_save
 
-from .models import Comment
+
+class Post(models.Model):
+    title = models.CharField(max_length=200)
 
 
+@notify_on_save  # default channel name: "<app_label>_<model_name>" → "myapp_comment"
+class Comment(models.Model):
+    post = models.ForeignKey(
+        Post, on_delete=models.CASCADE, related_name="comments"
+    )
+    author = models.ForeignKey(
+        settings.AUTH_USER_MODEL, on_delete=models.CASCADE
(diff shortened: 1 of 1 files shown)


johnrtipton:
**Self-review fix-up just pushed** (\`bf6414e\`):

Original draft referenced an invented \`@dj_listen(\"channel\")\` decorator — that API doesn't exist. The actual primitive set for LISTEN/NOTIFY broadcasts (per \`docs/website/guides/database-notifications.md\`) is:

- \`@notify_on_save\` on the model (from \`djust.db\`) — emits NOTIFY on save/delete with default channel \`<app_label>_<model_name>\`
- \`self.listen(channel)\` in \`mount()\` — joins the view to the Channels group bound to that NOTIFY
- \`handle_info(message)\` — fires when a NOTIFY arrives; switch on \`message.get(\"type\") == \"db_notify\"\`

Step 5 now uses these correctly. Removed the manual \`signals.py\` + \`NOTIFY <chan>, %s\` SQL boilerplate that was based on the wrong assumption — the decorator does it for you.

Cross-checked the other tutorials (#1131, #1133, #1136, #1140) against the documented API surface — only this one was affected.

johnrtipton:
Superseded by docs.djust.org#13, which consolidates all five tutorials in `content/website/guides/` so they ship on docs.djust.org's release cadence rather than waiting on a djust framework release + submodule bump. Promoting one upstream later is a one-line change (`_local_guide` → `_guide`) plus a file move into the submodule.

Issue #1134, opened 2026-04-27
tech-debt: 5 flaky tests fail in full pytest run, pass in isolation (test pollution)


johnrtipton:
## Summary

Five tests fail in the full \`pytest tests/ python/tests/ -q\` run that powers the pre-push hook, but **all pass when run in isolation**. This is classic test-pollution / fixture-leakage — some other test file mutates global state (likely Django settings, a module-level cache, or a Channels consumer registry) and these five then trip on the contaminated state.

The pre-push hook is failing PRs that are otherwise green. PR-A for #1043 (v0.9.0 streaming) hit this; observed across multiple sessions.

## Affected tests

1. \`python/tests/test_websocket_origin_validation.py::TestConnectOriginValidation::test_connect_accepts_missing_origin\`
2. \`python/tests/test_websocket_origin_validation.py::TestConnectOriginValidation::test_connect_accepts_allowed_origin\`
3. \`python/tests/test_websocket_origin_validation.py::TestConnectOriginValidation::test_connect_accepts_allowed_origin_with_port\`
4. \`python/tests/test_websocket_origin_validation.py::TestConnectOriginValidation::test_connect_allows_wildcard_allowed_hosts\`
5. \`tests/benchmarks/test_request_path.py::TestWebsocketMountPath::test_websocket_mount_counter\`

## Reproduction

\`\`\`bash
# Each individually: PASSES
uv run pytest python/tests/test_websocket_origin_validation.py::TestConnectOriginValidation -v
# → 7 passed

# Full suite: FAILS (4 of those 7)
uv run pytest python/ tests/ -q --ignore=python/djust/tests/test_tag_input_932.py
# → 6574 pass, 4 origin-validation FAILED + 1 benchmark FAILED + 5 pre-existing
\`\`\`

## Workaround applied

Each of the 5 tests is marked \`@pytest.mark.skip(reason='flaky, see #NNNN — pre-existing test-pollution; passes in isolation')\` so the pre-push hook unblocks. **The skip is temporary.** The right fix is to find the polluting test and either:
- Fix the pollution (cleaner)
- Add an autouse fixture in the affected files that resets the relevant global state

## Investigation hints

- All 5 failing tests touch the WebSocket / Channels consumer code path. Likely culprits: a test that registers a global \`AllowedOrigin\` middleware, mutates \`channels.layers.channel_layers\`, or modifies the Django \`ALLOWED_HOSTS\` setting and doesn't clean up.
[...]

Issue #1104, opened 2026-04-26
tech-debt: PR checklist — N similar sites need N tests, not 'a representative few'


johnrtipton:
**Source**: PR #1102 retro (https://github.com/djust-org/djust/pull/1102#issuecomment-4322684998)

When a PR makes the same mechanical change at N call sites, the test suite should cover **all N** sites, not a subset judged 'representative'. Identical-looking code ≠ tested code; one site can subtly differ (different surrounding context, different defaults, different escape pathway).

PR #1102: 5 sites in `frameworks.py` (text/textarea/select/checkbox/radio). Initial tests covered 4 of 5; radio was missed. Caught at Stage 11.

**Action**: add to PR checklist (and to the implementation-stage subagent prompt): 'For mechanical-replacement PRs, write tests for ALL N replacement sites — not just a representative few.'

johnrtipton:
Canonicalized in PR #1116 (CLAUDE.md / Process canonicalizations from PR retros section).

PR #1138, opened 2026-04-27
feat(streaming): {% live_render lazy=True %} + as_view dispatch (v0.9.0 PR-B, ADR-015)

### LINKED ISSUE #1043: tech-debt: Phase 2 streaming (lazy-child render + true server overlap)
**Source**: Retro v0.6.1 / PR #975
**Action Tracker**: row #119

**Action**: Phase 2 streaming (lazy-child render + true server overlap)

**Notes**: v0.6.2 — Phase 1 was transport-layer only

---

johnrtipton:
## Summary

PR-B of the v0.9.0 streaming arc (#1043). Capability layer on top of PR-A's foundation. Ships the user-facing `lazy=True` API + the `as_view` dispatch wiring deferred from PR-A.

### `{% live_render "..." lazy=True %}`

Three forms:
- `lazy=True` — parent-flush trigger (default). Fill arrives after `</body></html>`; client replaces immediately.
- `lazy="visible"` — IntersectionObserver-deferred slot replacement.
- `lazy={"trigger": ..., "timeout_s": ..., "on_error": ..., "placeholder": ...}` — full control object form.

### Thunk pipeline

The tag fires synchronously inside `sync_to_async(self.get)` before `aget` constructs the emitter. PR-B threads thunks through `parent._lazy_thunks` during sync render; `aget` harvests + transfers to the emitter via `register_thunk`; `arender_chunks` Phase 5 invokes them sequentially after `</body></html>` flushes.

### Wire format (per ADR-015)

```html
<template id="djl-fill-X" data-target="X" data-status="ok">
  <div dj-view data-djust-embedded="X">…rendered child…</div>
</template>
<script>window.djust.lazyFill('X')</script>
```

### Sticky + lazy = `TemplateSyntaxError`

Hard incompatibility per ADR §"Failure modes". Sticky preservation needs the slot at mount-frame time; lazy renders the slot AFTER mount.

### `as_view` dispatch wiring

`LiveView.as_view` now overrides Django's default. Classes with `streaming_render = True` return an async view callable (via `markcoroutinefunction`) that routes GET to `aget()` when `isinstance(request, ASGIRequest)`. WSGI fall-back goes through `sync_to_async(self.dispatch)`. `ASGIRequest` is the accurate transport-level signal — the loop-presence check is fooled by `async_to_sync` wrapping in the sync test `Client`.

## Issue × file × test mapping (per #1115)

| Component | Files | Tests |
|-----------|-------|-------|
| Tag `lazy=` branch | `python/djust/templatetags/live_tags.py` (~210 LoC) | `TestLazyTagValidation` (5) + `TestLazyPlaceholder` (4) + `TestLazyThunkStash` (2) + `TestLazyThunkInvocation` (3) |
| Phase-5 thunk loop | `python/djust/mixins/template.py` (~40 LoC) | `TestLazyEndToEnd::test_lazy_children_emit_fill_envelopes_after_body_close` |
| Thunk transfer + ASGI detection | `python/djust/mixins/request.py` (~25 LoC) | covered by integration tests |
| `as_view` override | `python/djust/live_view.py` (~50 LoC) | covered by integration + Stage 11 review focus |
| JS client | `python/djust/static/djust/src/50-lazy-fill.js` (~140 LoC) | client-side only; manual-test deferred to demo page |
| Error envelope | tag closure | `TestLazyThunkInvocation::test_thunk_error_envelope_when_mount_raises` + `TestLazyEndToEnd::test_lazy_thunk_failure_emits_error_envelope` |
| Timeout envelope | tag closure | `TestLazyThunkInvocation::test_thunk_timeout_envelope` |

## Test plan

- [x] `pytest tests/unit/test_live_render_lazy.py` → 14/14 pass
- [x] `pytest tests/integration/test_lazy_streaming_flow.py` → 2/2 pass
- [x] `pytest tests/unit/test_async_render_path.py tests/integration/test_streaming_flow.py tests/unit/test_live_render_tag.py tests/integration/test_sticky_redirect_flow.py` → 73/73 pass (no regression in adjacent areas)
- [x] Full Python suite: 6592 pass + 5 pre-existing failures unrelated (gallery_registry_933, static_security A020) + 5 skipped (pending #1134)

## Foundation for PR-C

PR-C will swap the sequential Phase-5 loop for `asyncio.as_completed`, enabling parallel render across lazy children. Closes #1043 umbrella.


johnrtipton (commented):
## Stage 11 review — feat(streaming): `{% live_render lazy=True %}` + as_view dispatch (v0.9.0 PR-B)

Reviewed against ADR-015 §"Wire format" / §"Failure modes" / §"Test contract" T6-T18, focus areas in the request, and the djust security/style rules.

**Verdict: COMMENT** (same-author `--approve` blocked; otherwise this would be `APPROVE` — only one nit blocks readiness, no must-fix bugs).

### What was done well

- **Clean foundation/capability split**: PR-A foundation soaked through `streaming_render` was already wired and tested; PR-B layers the user-facing `lazy=` syntax + `as_view` dispatch as a thin capability — exactly the "split-foundation pattern" canonicalized in #1122.
- **Thunk closure correctness**: `lazy_kwarg = kwargs.pop("lazy", False)` runs at line 1463, BEFORE `captured_kwargs = dict(kwargs)` at line 1534. By that point `view_id`, `sticky`, and `lazy` have all been popped (lines 1384, 1390, 1463), so `captured_kwargs` carries only user kwargs. The closure then snapshots `request`, `view_id`, `child_cls`, `escaped_id`, `lazy_config`, `parent`, and `view_path` — none aliased to mutables that the eager branch later touches. Closure capture is correct.
- **`as_view` override**: Verified against `django/core/handlers/base.py:248` — Django uses `iscoroutinefunction(wrapped_callback)` to decide whether to `sync_to_async`-wrap. `markcoroutinefunction(view)` at `live_view.py:422` makes that detection work. The `setup()` call + `request` attr check + `view_class`/`view_initkwargs`/`__doc__`/`__module__`/`__annotations__` copies + `view.__dict__.update(cls.dispatch.__dict__)` all faithfully mirror Django's stock `as_view`. Routing for non-streaming views correctly delegates to `super().as_view()` so there's zero overhead/behavior-change for existing views.
- **`_is_asgi_context(request)` signature change**: All three production callers (`request.py:361`, `live_view.py:410`) pass `request`. The fallback (no-arg) preserves backward compat for any external subclass that might call `_is_asgi_context()` without a request. Verified by grep — no production caller missed.
- **Phase-5 thunk error layering**: The inner closure catches `_AsyncioTimeout` / `PermissionError` / `Exception` and emits the appropriate `data-status="..."` envelope. The outer Phase-5 `except Exception` (`template.py:438`) is a defensive belt-and-suspenders for bugs in the closure itself (e.g., the closure raising before its own try/except activates). `ChunkEmitterCancelled` is correctly re-raised (line 437), so cancellation propagates. `asyncio.CancelledError` is `BaseException`, so neither layer swallows it — also correct.
- **CHANGELOG entry**: Comprehensive and accurate, including the post-`</html>` wire format ASCII, the `as_view` override semantics, and the sticky-incompatibility rationale.

### 🟡 Should-fix (non-blocking)

1. **Stale `40-lazy-fill.js` doc-string references in `live_tags.py`** (lines 1462 and 1637). The actual JS module shipped is `50-lazy-fill.js` (slot 50 in the bundle). Two doc comments still refer to the old name — this misleads future readers grepping for "40-lazy-fill". One-line fix each:
   - `live_tags.py:1462`: `static/djust/src/40-lazy-fill.js` → `static/djust/src/50-lazy-fill.js`
   - `live_tags.py:1637`: `from 40-lazy-fill.js performs` → `from 50-lazy-fill.js performs`
2. **Aspirational test docstring in `test_lazy_streaming_flow.py:20`**. The module docstring lists "4. WSGI fallback path: when arender_chunks isn't run (sync get only), placeholders survive in the body with no fills (graceful degradation)" — but no `test_lazy_wsgi_fallback_*` test exists. Either drop the docstring item OR add the test (a 10-line test that calls `parent.get()` directly and asserts `<dj-lazy-slot` survives in the response body, no `<template id="djl-fill-` present). The graceful-degrade contract is the kind of thing that quietly breaks if no test pins it. Recommend adding the test.
3. **No JS-side test for `50-lazy-fill.js`**. Per `CLAUDE.md` "New JS feature files in `static/djust/src/` must have corresponding test files in `tests/js/`" — this is an explicit project convention. The module has non-trivial logic worth pinning: idempotency on double-fire, `IntersectionObserver` deferred path, error-status wrapping in `<dj-error>`, auto-scan on `DOMContentLoaded`, and `window.djustReinit` reentry. Per #1113 retro, the `IntersectionObserver` stub MUST yield a microtask before invoking the callback (`await Promise.resolve()`). PR-C is a sensible parking lot for this if it can't fit here, but please file/track the gap.
4. **A075 system check (T12) deferral not surfaced in PR description**. ADR-015 §"Surface area" lists `python/djust/checks.py — A075 system check. ~40 LoC` and the test contract entry T12 covers it. This PR does not ship A075. The PR description doesn't acknowledge T12 as deferred — recommend adding one line under "Foundation for PR-C" calling out: "Deferred to PR-C: A075 system check (sticky+lazy on the same class) per ADR §Surface area." Sticky+lazy is already enforced at tag-eval time with a `TemplateSyntaxError` (`live_tags.py:1471`), so the system check is defense-in-depth, not load-bearing — but track the deferral explicitly so v0.9.0 retro picks it up.
5. **CSP / inline-script honesty**. The wire format includes `<script>window.djust&&window.djust.lazyFill("X")</script>` inline, and `_inject_handler_metadata` already supports `request.csp_nonce`. The lazy activator script does NOT propagate the nonce. Strict CSP without `'unsafe-inline'` will block the activator and the auto-scan-on-`DOMContentLoaded` fallback will paper over it (see `_autoFillOnDOMReady`). Two options: (a) make the activator nonce-aware now, OR (b) document the requirement in the user-facing docs/website guide and the CHANGELOG so apps shipping with strict CSP know they're trading the inline activator for the auto-scan fallback. Recommend (b) as a minimum; (a) is the cleaner long-term answer and is a candidate for PR-C.

### 🔵 Nits / suggestions

6. **`test_lazy_streaming_flow.py` integration test hardness**. The integration test mimics `RequestMixin.aget`'s thunk transfer via a hand-rolled `for view_id, thunk_fn in parent._lazy_thunks: emitter.register_thunk(view_id, thunk_fn)`. This is correct but means future refactors of `aget`'s transfer logic will not break the integration test. Consider adding a thin helper method on `RequestMixin` (e.g. `_transfer_lazy_thunks_to_emitter(emitter)`) so production and tests share the same plumbing — relevant to #1104 mechanical-replacement coverage.
7. **`<dj-lazy-slot>` is deliberately NOT registered as a custom element** (no `customElements.define`). The PR comment at `live_tags.py:1662-1664` correctly notes browsers parse it as `HTMLUnknownElement`, which is fine for `querySelector` + `replaceWith` use — verified. No change needed; flagging only because the test review focus item asked.
8. **Phase-5 thunk loop is sequential per ADR-015**. Confirmed `template.py:431-452` invokes thunks via `for view_id, thunk_fn in emitter.thunks: chunk_bytes = await thunk_fn()` — strictly serial. PR-C will swap to `asyncio.as_completed` per the CHANGELOG note. Sequential semantics also mean `_register_child(view_id, child)` collisions (focus item 7) cannot race — second thunk's mount runs after the first's `_register_child` completed. Good.

### Auto-reject scan

- `mark_safe(f"...")` with interpolated values: clean — every `mark_safe` call uses string concatenation, not f-strings (verified `live_tags.py:1671-1679` and the thunk envelope build at 1639-1652).
- `@csrf_exempt`: not added.
- f-string logging: clean — all logger calls in changed files use `%s`-style.
- Bare `except: pass`: clean — every `except` either logs or re-raises.
- `print()`: clean.
- Unguarded `console.log`: clean — both `console.warn` calls in `50-lazy-fill.js` are inside `if (window.djustDebug)` guards (lines 36 and 84).

### Process notes (v0.8.6 retro applicability)

- **#1122 split-foundation pattern**: PR-A foundation → PR-B capability is textbook. Apply to PR-C similarly.
- **#1115 issue × file × test mapping**: PR description has the table — useful, thanks.
- **#1106 CHANGELOG test-count wording**: "14 unit + 2 integration" in the PR body and "14 new cases in `tests/unit/test_live_render_lazy.py` cover validation" / "2 new integration cases in `tests/integration/test_lazy_streaming_flow.py` drive the full pipeline" — both name what NEW cases cover, not the file's total. ✓

### Verdict

`COMMENT` (would be `APPROVE` if same-author rule didn't block). 🟡 5 should-fixes above are docs/test-completeness nits; none block merge. The implementation logic is sound, the Django ASGI integration is verified against the upstream code, and the auto-reject grep is clean.

johnrtipton:
# PR #1138 Retro — v0.9.0 P2 PR-B: lazy=True + as_view dispatch (ADR-015)

**Date:** 2026-04-27
**Scope:** PR-B of v0.9.0 streaming arc (#1043). Capability layer on
top of PR-A's foundation. Ships `{% live_render lazy=True %}` user
API + `as_view` dispatch wiring (deferred from PR-A by Stage 11).
**Tests at close:** 14 unit + 2 integration + 8 JS = 24 PR-B tests +
73 adjacent (sticky/streaming/live_render) = 97/97 PR-B-related green.
Full suite: 6592 Python pass + 1427 vitest pass.
**LoC:** +1497 / -22 across 14 files (post-address-findings push
adds 240 LoC of JS tests + ADR + doc fixes).

## What Worked Well

1. **Plan-stage pre-flight discovered the thunk-stash architecture
   problem before code.** The naive design "register thunk on
   `parent._chunk_emitter` from inside the tag" doesn't work because
   the tag fires inside `sync_to_async(self.get)` BEFORE aget
   constructs the emitter. Plan caught this at the
   "is the design implementable" gate — solution: stash on
   `parent._lazy_thunks` during sync render, harvest in aget after
   sync_to_async returns. Two-line architecture change saved ~half a
   day of dead-end implementation.

2. **`isinstance(request, ASGIRequest)` over `asyncio.get_running_loop()`
   for ASGI detection.** Found this only after the as_view override
   broke 2 existing Phase-1 sync streaming tests. The loop-presence
   check is fooled by `async_to_sync` wrapping in the sync test
   `Client`. Switching to the request-class isinstance check fixed both
   the regression AND made the WSGI fallback contract more honest.

3. **Stage 11 caught all 5 🟡 findings.** Stale `40-lazy-fill.js`
   doc-strings (slot 16 was occupied; we used 50), aspirational test
   docstring listing a never-written test, missing JS-side test
   (CLAUDE.md convention violation), A075 deferral not surfaced, CSP
   trade-off not documented. All addressed in second push.

## What Didn't Work

1. **The Rust template engine doesn't recognize `{% live_render %}`.**
   Initial integration test used Django Client + URL routing → went
   through render_full_template → through Rust → tag unsupported. Had
   to refactor the integration test to drive the chunk pipeline
   directly (Django render → arender_chunks → emitter drain),
   bypassing Rust. Worked but lost the "real Django dispatch" coverage.
   Production users of `lazy=True` who use `template_name` (Django
   loader) are fine; `template = "..."` (Rust engine) users would
   need to switch. Worth noting in docs.

2. **Vitest unhandled-rejection flakiness in `view-transitions.test.js`.**
   Pre-push hook caught 2 unhandled rejections from
   `EnvironmentTeardownError: Closing rpc while "onUserConsoleLog" was
   pending` even though all 1427 vitest tests passed. Non-deterministic
   — passed on the second commit attempt. Pattern from #1134 (test
   pollution) but in JS. Worth filing a follow-up to investigate
   vitest worker teardown.

3. **`document.readyState` semantics in JSDOM tripped the auto-scan
   test once.** With `runScripts: outside-only` + manual eval, document
   stays in `loading` state until JSDOM finishes its initial parse.
   The DOMContentLoaded listener registers but doesn't fire. Test
   needed a manual `dispatchEvent(new Event('DOMContentLoaded'))` to
   exercise the listener path. Added a comment explaining the
   workaround.

## What to Improve

1. **Action Tracker row — `lazy=True` + Rust template engine
   compatibility.** Production users with `template = "..."` (Rust
   engine) can't use `lazy=True` because `{% live_render %}` isn't a
   recognized Rust tag. Either:
   - Register `live_render` as a Rust tag handler (substantial work).
   - Document the limitation in `lazy-render.md`.
   - File as #?? for a follow-up.

2. **A075 system check** — deferred to a follow-up per Stage 11
   review. Tracked in ADR-015 §"Deferred from PR-B".

3. **CSP-nonce-aware activator script** — same. Tracked in ADR-015
   §"Deferred from PR-B".

4. **Vitest unhandled-rejection investigation** — file an issue similar
   to #1134 for `view-transitions.test.js` worker teardown.

- (ADR-015 §"Deferred from PR-B" already documents A075 + CSP. CLAUDE.md
  additions deferred to v0.9.0 milestone retro after PR-C ships.)

## Recurring Issue Tracker

- #1134 (5 flaky Python tests) — still skipped this session. Pre-push
  hook does not run them; future fix unchanged.
- New JS unhandled-rejection flakiness from `view-transitions.test.js`.
  Worth filing.

## Review Stats

| Metric | Value |
|--------|-------|
| Tests added | 24 (14 Python unit + 2 Python integration + 8 JS) |
| 🔴 Findings | 0 |
| 🟡 Findings | 5 (all addressed) |
| Findings fixed | 5 |
| Stage 11 verdict | COMMENT (would-be APPROVE; same-author --approve blocked) |
| LoC core (non-test, non-doc) | ~580 (tag branch + dispatch + emitter wiring) |
| LoC test | ~700 (Python + JS) |
| LoC ADR/docs/CHANGELOG | ~220 |

PR #1135, opened 2026-04-27
feat(streaming): async render path foundation — aget + ChunkEmitter (v0.9.0 PR-A, ADR-015)

johnrtipton:
## Summary

PR-A foundation of the v0.9.0 P2 streaming arc (#1043). Closes the v0.6.1 retro #116 doc-claim debt: Phase 1 was a regex-split-after-render with no real TTFB win; Phase 2 PR-A introduces the actual async render path so `streaming_render = True` shell-flushes to the wire BEFORE `get_context_data()` runs.

Per retro #1122 split-foundation rule, the v0.9.0 #1043 work ships as 3 PRs:

| | Scope | Status |
|---|---|---|
| **PR-A (this PR)** | Foundation: `aget()`, `ChunkEmitter`, `arender_chunks()`. No new user API. | Open |
| PR-B | `{% live_render lazy=True %}` user API + `<dj-lazy-slot>` / `<template>` fill protocol + `16-lazy-fill.js` client | Blocked by PR-A |
| PR-C | `asyncio.as_completed()` parallel render across thunks; closes #1043 umbrella | Blocked by PR-A |

## What this PR ships

1. **`python/djust/http_streaming.py`** (new, ~230 LoC) — `ChunkEmitter` class with bounded `asyncio.Queue` for backpressure, cancellation propagation via `request_token`, and `register_thunk()` API surface that PR-B will hook into. Exposes `__aiter__` for `StreamingHttpResponse` consumption.
2. **`async def aget()`** on `RequestMixin` (~150 LoC) — wraps the sync render via `sync_to_async(self.get)`, drives `arender_chunks()` to push chunks through the emitter, returns a `StreamingHttpResponse` with `X-Djust-Streaming-Phase: 2` header. ASGI disconnect watcher cancels the emitter on client close. WSGI fallback: when no event loop is running, falls back to sync `get()` via `sync_to_async`.
3. **`arender_chunks()`** async coroutine on `TemplateMixin` (~135 LoC) — splits HTML at `<div dj-root>` boundaries into 4 chunks (shell-open / body-open / body-content / body-close), pushes via `emitter.emit()` with `await asyncio.sleep(0)` between each so ASGI flushes the shell to the wire before body chunks are queued.

## Locked design decisions (full in ADR-015)

- HTTP GET only. WS path uses `assign_async`.
- WSGI deployments fall back to Phase-1 cosmetic chunks (graceful degrade).
- `streaming_render = False` (default) keeps every existing app on `HttpResponse`.
[...]

Issue #1042, opened 2026-04-25
tech-debt: Forward-replay through branched timeline (Redux DevTools parity)


johnrtipton:
**Source**: Retro v0.6.1 / PR #976
**Action Tracker**: row #118

**Action**: Forward-replay through branched timeline (Redux DevTools parity)

**Notes**: v0.6.2 candidate

---
_Filed by `/pipeline-retro --reconcile` on 2026-04-25 as part of Scope C: file all Action Tracker rows as GitHub issues._

johnrtipton:
Closing — relocated to ROADMAP / future v0.9.x as a deferred feature. Forward-replay through branched timeline (Redux DevTools parity) is a v0.6.2+ candidate per the original v0.6.1 retro entry; the action item belongs in milestone planning rather than the tech-debt drain queue.

johnrtipton:
Reopening — last unit of v0.9.0 shape C; builds on #1041's component-level capture.

Issue #1041, opened 2026-04-25
tech-debt: Component-level time-travel (Phase 1 records against parent; full component capture)


johnrtipton:
**Source**: Retro v0.6.1 / PR #976
**Action Tracker**: row #117

**Action**: Component-level time-travel (Phase 1 records against parent; full component capture)

**Notes**: v0.6.2 candidate

---
_Filed by `/pipeline-retro --reconcile` on 2026-04-25 as part of Scope C: file all Action Tracker rows as GitHub issues._

johnrtipton:
Closing — relocated to ROADMAP "Investigate & Decide" / future v0.9.x as a deferred feature. Component-level time-travel is a v0.6.2+ candidate per the original v0.6.1 retro entry; the action item belongs in milestone planning rather than the tech-debt drain queue.

johnrtipton:
Reopening — promoted into v0.9.0 shape C P3 work; pipeline starting.

PR #1139, opened 2026-04-27
feat(streaming): asyncio.as_completed parallel lazy render (v0.9.0 PR-C, closes #1043)

### LINKED ISSUE #1043: tech-debt: Phase 2 streaming (lazy-child render + true server overlap)

johnrtipton:
## Summary

Final PR of the v0.9.0 streaming arc. Closes #1043. Replaces the sequential thunk loop in `arender_chunks` Phase 5 with `asyncio.as_completed` so lazy children render in parallel.

| | |
|---|---|
| Foundation | PR #1135 (PR-A) — `aget` + `ChunkEmitter` + `arender_chunks` |
| API + dispatch | PR #1138 (PR-B) — `{% live_render lazy=True %}` + `as_view` wiring |
| **This PR** (PR-C) | `asyncio.as_completed` parallel render |

## What changes

| Before (PR-B) | After (PR-C) |
|---|---|
| Sequential `for` loop over thunks | `asyncio.ensure_future` per thunk; iterate over `asyncio.as_completed` |
| Wall clock = sum(thunk durations) | Wall clock = max(thunk durations) |
| Chunks emerge in registration order | Chunks emerge in completion order |

Client-side reconciliation is keyed by slot id (`data-target` on `<template id="djl-fill-X">`), so out-of-order arrival is correct by construction. **Zero client changes needed.**

## Cancellation

Mid-stream client disconnect cancels all pending thunk tasks via `task.cancel()`. Already-completed tasks whose results were not yet iterated are GC'd. Tasks already running through `sync_to_async` to synchronous render code complete normally — asyncio cancellation doesn't propagate into sync DB work, the documented contract per ADR-015 §"Cancellation contract".

## Tests

3 new wall-clock-sensitive integration tests in `tests/integration/test_chunks_overlap.py`:
1. **Order**: thunks (100ms, 50ms, 25ms) registered a→b→c → fills arrive c→b→a (completion order, NOT registration order).
2. **Wall clock**: 3 × 50ms thunks → total < 100ms (sequential baseline 150ms; 100ms tolerance keeps non-flaky on shared CI).
3. **Failure isolation**: one thunk raises → others still emit fills (no stall).

## Test plan

- [x] `pytest tests/integration/test_chunks_overlap.py` → 3/3 pass
- [x] `pytest tests/integration/test_lazy_streaming_flow.py tests/unit/test_async_render_path.py tests/unit/test_live_render_lazy.py tests/integration/test_streaming_flow.py` → 40/40 pass (no regression)

## Closes #1043

v0.9.0 streaming arc complete after PR-C merges:
- ✅ Async render foundation (PR-A)
- ✅ `lazy=True` user API + dispatch wiring (PR-B)
- ✅ Parallel render via `asyncio.as_completed` (PR-C)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,40 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- **Parallel lazy render via `asyncio.as_completed` (v0.9.0 PR-C, closes #1043)** —
+  closes the v0.9.0 streaming arc. PR-B shipped sequential thunk
+  invocation in `arender_chunks` Phase 5 (one thunk runs to
+  completion before the next starts; total wall-clock time =
+  sum of thunk durations). PR-C swaps the for-loop for
+  `asyncio.as_completed` over the thunk-task set. All thunks start
+  concurrently; chunks emerge in completion order rather than
+  registration order. Total wall-clock time = max(thunk_durations).
+
+  Client-side reconciliation is keyed by slot id (`data-target` on
+  `<template id="djl-fill-X">`), so out-of-order chunk arrival is
+  correct by construction — no client changes needed.
+
+  Cancellation: when the emitter is cancelled mid-stream (client
+  disconnect), all pending thunk tasks are cancelled via
+  `task.cancel()`. Already-completed tasks whose results were not yet
+  iterated are GC'd. Tasks already running through `sync_to_async` to
+  a synchronous render function will complete (asyncio cancellation
+  doesn't propagate into sync DB work) — the documented contract per
+  ADR-015 §"Cancellation contract".
+
+  Files: `python/djust/mixins/template.py` (~50 LoC swap from
+  for-loop to `asyncio.as_completed`). 3 new wall-clock-sensitive
+  tests in `tests/integration/test_chunks_overlap.py`:
+  - Three thunks (100ms, 50ms, 25ms) registered in that order →
+    chunks arrive in completion order (slot-c, slot-b, slot-a).
+  - Three 50ms-each thunks → wall clock under 100ms (sequential
+    baseline 150ms).
+  - One thunk raises → others still emit their fills (no stall).
+
+  Closes #1043. v0.9.0 streaming arc complete: PR-A (foundation) →
+  PR-B (`lazy=True` user API + `as_view` dispatch) → PR-C (parallel
+  render).
+
 - **`{% live_render lazy=True %}` capability + `as_view` dispatch wiring (v0.9.0 PR-B, ADR-015)** —
   ships the user-facing API on top of PR-A's async render foundation.
   Three forms: `lazy=True` (parent-flush trigger, default), `lazy="visible"`
--- a/docs/adr/015-phase-2-streaming.md
+++ b/docs/adr/015-phase-2-streaming.md
@@ -67,9 +67,12 @@ Per-task timeout. Sentinel-based cancellation propagates via
 `request_token` from the emitter on ASGI scope `disconnected`.
 
 Files:
-- `python/djust/http_streaming.py` — `as_completed()` refactor. ~80 LoC.
-- `python/djust/decorators.py` — `cancel_async` reason field. ~30 LoC.
-- `tests/integration/test_chunks_overlap.py` — slow vs fast child timing assertions. ~200 LoC.
+- `python/djust/mixins/template.py` — `as_completed()` refactor in
+  `arender_chunks` Phase 5. ~80 LoC. (Originally planned for
+  `http_streaming.py` but the loop lives where it consumes thunks.)
+- `tests/integration/test_chunks_overlap.py` — slow vs fast child
+  timing assertions + mid-stream cancellation propagation
+  (T-PRC-4). ~200 LoC.
 
 ## Total estimate
 
--- a/python/djust/mixins/template.py
+++ b/python/djust/mixins/template.py
@@ -418,28 +418,62 @@ Object.assign(window.handlerMetadata, {json.dumps(metadata)});
             logger.debug("arender_chunks: cancelled mid-stream")
             return
 
-        # 5. Lazy thunks (PR-B, ADR-015). The body of the page has
-        # already flushed; lazy ``<template id="djl-fill-X">`` chunks
-        # are appended after </html>. Browsers tolerate post-</html>
-        # content per the HTML5 parser tree-construction spec — the
-        # template element + its inline <script> activator move into
-        # the implicit body and execute in order. Sequential in PR-B;
-        # PR-C ships parallelism via ``asyncio.as_completed``.
+        # 5. Lazy thunks (PR-B + PR-C, ADR-015). The body of the page
+        # has already flushed; lazy ``<template id="djl-fill-X">``
+        # chunks are appended after </html>. Browsers tolerate
+        # post-</html> content per the HTML5 parser tree-construction
+        # spec — the template element + its inline <script> activator
+        # move into the implicit body and execute in order.
+        #
+        # PR-C: parallel render via ``asyncio.as_completed``. All
+        # thunks start concurrently; chunks emerge in completion order
+        # rather than registration order. Total wall-clock time =
+        # max(thunk_durations) instead of sum(thunk_durations). Client-
+        # side reconciliation is keyed by slot id (``data-target``) so
+        # out-of-order arrival is correct by construction.
         if not emitter.thunks:
             return
+
+        # Schedule every thunk concurrently. ``asyncio.ensure_future``
+        # wraps coroutines into tasks so ``as_completed`` can iterate.
+        thunk_tasks = [
+            (view_id, asyncio.ensure_future(thunk_fn())) for view_id, thunk_fn in emitter.thunks
+        ]
+        # Build a reverse lookup so we can recover the view_id when a
+        # task surfaces from ``as_completed``.
+        task_to_id = {task: view_id for view_id, task in thunk_tasks}
+
         try:
-            for view_id, thunk_fn in emitter.thunks:
+            for completed in asyncio.as_completed([task for _, task in thunk_tasks]):
                 if emitter.cancelled:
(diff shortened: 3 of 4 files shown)


johnrtipton (commented):
Stage 11 review — verdict: **REQUEST_CHANGES** (posted as `--comment` because GitHub blocks self-author `--request-changes`; treat the two 🔴 items as blocking).

Diff is small and the core swap from sequential `for` loop to `asyncio.as_completed` is the right shape. Tests T-PRC-1/2/3 from ADR-015 §"Test contract" all pass; 43/43 adjacent regression tests still green; no auto-reject triggers (no `mark_safe`, no `console.log`, no f-string logging). Closes the v0.9.0 streaming arc as advertised.

I reproduced three asyncio semantics concerns against a real interpreter before posting. Findings below.

---

## 🔴 Must-fix

**1. `failed_task` recovery picks the wrong view_id when ≥2 thunks fail.** `python/djust/mixins/template.py:468-471`:

```python
failed_task = next(
    (t for t in task_to_id if t.done() and t.exception() is not None),
    None,
)
```

This returns the **first** done-with-exception task in dict iteration order, not the one that just surfaced from `as_completed`. Repro:

```
surfacing log: ['view-1', 'view-1']    # actual (BUG)
expected:      ['view-1', 'view-2']    # one log per failure, correct id
```

The second failure's log message attributes the failure to view-1. The pre-PR-C code captured `view_id` directly in the loop variable and was correct — this is a regression introduced by the parallelization. Fix is small; two clean options:

- `asyncio.wait({t1, t2, ...}, return_when=FIRST_COMPLETED)` loop — yields the actual task done, look up `task_to_id[done_task]` and you have the correct id. Verified.
- Wrap each thunk in `async def _wrap(view_id, coro): try: return view_id, await coro, None; except BaseException as e: return view_id, None, e` — `await completed` returns the tuple, no recovery needed. Verified.

**2. Add a regression test for T-PRC-4 (mid-stream cancellation while thunks are pending).** ADR-015 §"Test contract" lists this as a required scenario for PR-C and it's not covered. `test_aget_cancels_emitter_on_disconnect` exercises the watcher path with no thunks registered, so the new `if emitter.cancelled: ... task.cancel()` block inside the `as_completed` loop and the outer `except ChunkEmitterCancelled` branch have zero direct test coverage. Suggested shape: register 3 long-running thunks, cancel the emitter after the first fill arrives, assert (a) still-pending tasks have `.cancelled() is True`, (b) no extra fills appear in the chunk stream, (c) no `RuntimeWarning` raised.

## 🟡 Should-fix

**3. `RuntimeWarning: coroutine '_AsCompletedIterator._wait_for_one' was never awaited`** when the `if emitter.cancelled: return` branch fires mid-iteration. The `completed` future yielded by `as_completed` is left unawaited; the iterator's internal scheduler coroutine then warns on GC. Reproduced on CPython 3.14. Cosmetic but pollutes stderr in production; the proposed T-PRC-4 test will fail-fast against this if you assert no `RuntimeWarning`. Fix: drain the iterator in a `try/finally` (await each remaining `completed`, swallow `CancelledError`) before returning.

**4. Test #3 (`test_parallel_thunk_failure_does_not_stall_others`) uses a non-realistic thunk.** Production thunks always emit a `data-status="error"` envelope themselves; the `async def _failing(): raise ValueError(...)` here only exercises the defensive outer handler. That's fine — the outer handler **is** what PR-C added — but worth a one-line comment in the test acknowledging the non-production shape so a future reader doesn't think the thunk closure is missing its envelope wrap.

**5. Document why suppressing `asyncio.CancelledError` is safe here.** `template.py:460-464` swallows `CancelledError` with `continue`, which is normally a code smell (kills the cancellation signal). The reason it's safe in this position is that the `CancelledError` surfacing from `await completed` is the **inner thunk task's** cancellation, not the outer `arender_chunks` task's — the next iteration's `if emitter.cancelled` check handles outer cancellation. A 2-line comment locks this in for the next reader and prevents a "let me clean up this dirty except" refactor that would break cancellation propagation. Verified the suppression is semantically correct.

## 🟢 Verified clean

- `task_to_id` identity-hashing of Task objects is fine (no aliasing risk; `asyncio.ensure_future` produces fresh tasks).
- Error envelope contract still holds; outer phase-5 handler is a defensive layer over thunks that fail to wrap.
- Iteration syntax is correct (`for completed in asyncio.as_completed(...)` + `await completed`, NOT `async for`).
- Wall-clock margin: measured 51ms median across 20 trials on this dev laptop against a 100ms ceiling. Comfortable 2× margin. If a flake emerges on shared CI, bump to 120ms — but current shape looks fine.
- No auto-reject triggers.

## 🔵 Doc nit

ADR-015 §"PR-C" line 70 says the refactor would live in `python/djust/http_streaming.py`, but it actually landed in `python/djust/mixins/template.py` (where `arender_chunks` is). Either an ADR amendment line or a follow-up commit moving the as_completed block down to `http_streaming.py` resolves it. Low priority — file location settled in PR-A.

---

Once #1 is fixed (with a multi-failure test), #2 added (T-PRC-4 cancellation), and #3 silenced, this is approve-shape. Nice clean diff overall.

johnrtipton:
# PR #1139 Retro — v0.9.0 P2 PR-C: parallel render via asyncio.as_completed (closes #1043)

**Date:** 2026-04-27
**Scope:** Final PR of v0.9.0 streaming arc. Closes #1043. Replaces
sequential thunk loop in arender_chunks Phase 5 with
asyncio.as_completed.
**Tests at close:** 40/40 (3 PR-C wall-clock tests + 1 cancellation
test added during address-findings + 36 adjacent).
**LoC:** +259 / -11 across 3 files (+143 in second push for the
multi-failure fix + cancellation test + comment).

## What Worked Well

1. **Per-thunk wrapper for unambiguous failure attribution.** Stage 11
   reviewer's most useful catch: the original code's `next(t for t in
   task_to_id if t.done() and t.exception() is not None)` was
   identity-fragile under multi-failure (returned the first
   done-with-exception task, not the surfacing one). Refactored to a
   per-thunk `_wrap` coroutine returning `(view_id, result, exc)` —
   captures identity at thunk-start time. Cleaner code AND correct.
   Also enables future use cases (per-thunk telemetry,
   per-thunk timing).

2. **Wall-clock test discipline (50ms tolerance for parallel,
   100ms ceiling for total) held up.** Reviewer tested 20 trials,
   median 51ms vs 100ms ceiling — ~2× margin. Non-flaky.

3. **Stage 11 forced T-PRC-4 (mid-stream cancellation while thunks
   pending) test that I had skipped.** Without that test, the
   `task.cancel()` propagation path had zero coverage. The test now
   asserts: cancel mid-stream → arender_chunks returns within 0.5s
   (vs would-be 1s sleep on the unaborted slow thunks).

## What Didn't Work

1. **Branch-state drift between `main` and `feat/streaming-phase2-1043-pr-c`.**
   After the `git checkout main && git pull` to sync after PR-B
   merged, I started editing files. Then the working-tree changes
   from PR-C bled into main. `git diff HEAD` showed nothing because
   I hadn't committed yet, but it was the wrong branch. Caught it
   at "wait, why does git log show 19816602 not de0b4bce?" The
   `git checkout feat/streaming-phase2-1043-pr-c` brought it back.

   **Lesson** (carryover from PR-A retro #122): the
   "MANDATORY post-commit verification" rule needs an
   "expected branch" check too. `git symbolic-ref --short HEAD`
   should match the active state file's `branch_name` BEFORE making
   edits, not just before committing. File as a tracker row.

2. **One residual `RuntimeWarning: coroutine
   '_AsCompletedIterator._wait_for_one' was never awaited`** survives
   on the cancellation path. Asyncio's `as_completed` keeps an
   internal getter task that's started before any iteration; when we
   `return` mid-iteration, the leftover getter is GC'd without being
   awaited. Doesn't affect correctness; tests pass; warning is
   cosmetic. The `finally: _cancel_pending` defensive block reduces
   this but doesn't eliminate it (it's an asyncio internal). Worth
   tracking for a Python-version-specific suppression.

## What to Improve

1. **Add a "branch-name verify" check to pipeline-run skill** —
   carry-over from PR-A retro #122. The MANDATORY post-commit
   verification catches commits to the wrong branch only AFTER they
   register. A pre-commit check should compare
   `git symbolic-ref --short HEAD` against the state file's
   `branch_name` BEFORE allowing the commit. File as Action Tracker
   row.

2. **Asyncio `_AsCompletedIterator._wait_for_one` warning suppression**
   — file follow-up.

3. **PR-D candidate (post-1.0):** lazy-component-level (vs
   lazy-LiveView-level) — `LiveComponent` rendering inside a
   parent template currently doesn't have a `lazy=True` analog. Out
   of scope per ADR §"Out of scope" but worth tracking as a future
   feature.

- (ADR-015 §"Deferred from PR-B" already documents A075 + CSP. CLAUDE.md
  additions deferred to v0.9.0 milestone retro after #1041 + #1042
  ship.)

## Recurring Issue Tracker

- #1134 (5 flaky Python tests) — still skipped.
- New: vitest unhandled-rejection in `view-transitions.test.js`.
- New: `_AsCompletedIterator._wait_for_one` warning on cancellation.

## Review Stats

| Metric | Value |
|--------|-------|
| Tests added | 4 (3 wall-clock + 1 cancellation) |
| 🔴 Findings | 2 (both addressed) |
| 🟡 Findings | 3 (all addressed) |
| 🔵 Findings | 1 (addressed) |
| Stage 11 verdict | REQUEST_CHANGES → APPROVE_AFTER_FIXES |
| LoC core (non-test, non-doc) | ~80 |
| LoC test | ~230 |
| LoC ADR/docs/CHANGELOG | ~40 |

## v0.9.0 streaming arc complete

PR-A foundation (PR #1135) → PR-B `lazy=True` API + `as_view` dispatch
(PR #1138) → PR-C parallel render (this PR). Closes #1043. v0.9.0
shape C still has #1041 (component-level time-travel) and #1042
(forward-replay) to ship.

Issue #1125, opened 2026-04-26
tech-debt: bulk dispatch-site refactor + count-test pattern


johnrtipton:
**Source**: v0.8.6 milestone retro finding #5.

PR #1117 (decorate 21 `on_table_*` methods) and PR #1120 (refactor 9 hook lifecycle sites to `_safeCallHook`) both used the same shape: many similar sites → one helper + a count-based test (`test_handler_count_matches_expected`, `test_all_on_table_methods_decorated`) that catches future additions that forget the pattern.

Canonicalize the pattern: when introducing a helper that wraps many call sites, add a test that enumerates expected sites and asserts the count matches what's actually in the codebase.

johnrtipton:
Closing — canonicalized in CLAUDE.md §v0.8.6 "Bulk dispatch-site refactor + count-test pattern (#1125)". The count-test rule + grep example are documented. 🤖 v0.9.1-6 backlog audit.

Issue #1103, opened 2026-04-26
tech-debt: prefer 'is None' coalescing over kwargs.setdefault for forwarding mixins


johnrtipton:
**Source**: PR #1102 retro (https://github.com/djust-org/djust/pull/1102#issuecomment-4322684998)

`kwargs.setdefault('x', default)` does NOT overwrite a caller-passed `None` — the key already exists. When forwarding kwargs through a mixin to a downstream API where the value will be used directly (e.g. as a dict key), this can produce surprising failures.

PR #1102 hit this in `WizardMixin.as_live_field()` — `kwargs.setdefault('dom_event', self.wizard_input_event)` would propagate `None` to `attrs[None] = ...` and emit `None="validate_field"` in the HTML. Caught at Stage 11.

**Pattern to canonicalize in CLAUDE.md** (Python patterns / kwarg forwarding):
```python
# Wrong — setdefault doesn't overwrite caller-passed None
kwargs.setdefault('x', self.default_x)

# Right — explicit coalesce
if kwargs.get('x') is None:
    kwargs['x'] = self.default_x
```

**Action**: add this to `CLAUDE.md` and the PR checklist.

johnrtipton:
Canonicalized in PR #1116 (CLAUDE.md / Process canonicalizations from PR retros section).

PR #1141, opened 2026-04-27
feat(time-travel): component-level state capture + restore (closes #1041)

### LINKED ISSUE #1041: tech-debt: Component-level time-travel (Phase 1 records against parent; full component capture)

johnrtipton:
## Summary

v0.9.0 P3 #1041. Extends the v0.6.1 time-travel ring buffer to capture per-component public state alongside the parent LiveView's state. Multi-component pages can now scrub back through history with each component's state faithfully restored.

## Snapshot format

```python
{
  "title": "Parent",          # parent attrs (existing)
  "count": 5,
  "__components__": {          # NEW reserved key (#1041)
    "alpha": {"active": "q1", "multiple": false},
    "beta": {"value": 10, "label": "ten"},
  },
}
```

The reserved `__components__` key keeps component snapshots out of the parent's flat attr namespace and gives the time-travel debug panel a clean shape to render per-component scrubbers.

## Restoration

`time_travel.restore_snapshot` detects `__components__` and dispatches each `{component_id: state}` entry to the matching component in `view._components` via `safe_setattr`. Components absent from the snapshot keep their current state — components are first-class instances, not parent-scoped attrs, so the ghost-attr cleanup model (used for parent state) doesn't apply across component boundaries.

Snapshot references a missing component_id: log warning, return `False` overall, but other components still restore.

## Files (~100 LoC core + ~180 LoC tests)

| Component | Files | Tests |
|-----------|-------|-------|
| Capture | `python/djust/live_view.py` (`_capture_components_snapshot` + `_capture_snapshot_state` ext) | 3 capture cases |
| Restore | `python/djust/time_travel.py` (`_COMPONENTS_SNAPSHOT_KEY` + Phase-3 dispatch) | 4 restore cases |

7 new cases in `TestComponentLevelTimeTravel` in `tests/unit/test_time_travel.py`:
- Capture-with-components / capture-without-components / private+callable filtering
- Restore-dispatches-per-component / unknown-id-logs-False / absent-component-survives / live-component-disconnection (mirrors parent-state aliasing fix from PR #1023 Stage 11)

## Stage-4 first-principles applied (per #1032 retro learning)

The original task framing implied a substantial new "Phase 2 captures component state" subsystem. Reading the existing `_capture_snapshot_state` first showed it already walks `self.__dict__` cleanly; the missing piece was just iterating `self._components` and producing a sub-dict. ~100 LoC core change instead of a green-field rewrite.

## Test plan

- [x] `pytest tests/unit/test_time_travel.py` → 44/44 pass (37 existing + 7 new)
- [x] `pytest tests/integration/test_time_travel_flow.py` → 9/9 pass (no regression)
- [x] Pre-commit hooks green

## Closes #1041

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,38 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- **Component-level time-travel (closes #1041, v0.9.0 P3)** — extends
+  the v0.6.1 time-travel ring buffer to capture per-component public
+  state alongside the parent LiveView's state. Multi-component pages
+  can now scrub back through history with each component's state
+  faithfully restored.
+
+  Snapshot format: `_capture_snapshot_state` adds a reserved
+  `__components__` key holding a `{component_id: {field: value}}`
+  nested dict. Components in `self._components` (registered by
+  `_assign_component_ids`) each contribute their public state. The
+  reserved key keeps component snapshots out of the parent's flat
+  attr namespace and gives the time-travel debug panel a clean shape
+  to render per-component scrubbers.
+
+  Restoration: `time_travel.restore_snapshot` detects
+  `__components__` in the snapshot and dispatches each
+  `{component_id: state}` entry to the matching component in
+  `view._components` via `safe_setattr`. Components absent from the
+  snapshot keep their current state — components are first-class
+  instances, not parent-scoped attrs, so the ghost-attr cleanup model
+  used for parent state doesn't apply.
+
+  Files: `python/djust/live_view.py` (~60 LoC: `_capture_components_snapshot`
+  helper + `_capture_snapshot_state` extension); `python/djust/time_travel.py`
+  (~40 LoC: `_COMPONENTS_SNAPSHOT_KEY` constant + per-component
+  restoration phase). 7 new cases in `TestComponentLevelTimeTravel`
+  in `tests/unit/test_time_travel.py` cover capture-with-components,
+  capture-without-components, private/callable filtering, restoration
+  dispatch, unknown-component-id handling, absent-component
+  preservation, and snapshot/live disconnection (mirrors the
+  parent-state aliasing fix from PR #1023's Stage 11 review).
+
 - **Parallel lazy render via `asyncio.as_completed` (v0.9.0 PR-C, closes #1043)** —
   closes the v0.9.0 streaming arc. PR-B shipped sequential thunk
   invocation in `arender_chunks` Phase 5 (one thunk runs to
--- a/python/djust/live_view.py
+++ b/python/djust/live_view.py
@@ -684,6 +684,15 @@ class LiveView(
         client to post a ``STATE_SNAPSHOT`` message to the service worker
         when opt-in via :attr:`enable_state_snapshot` is True.
 
+        Component state (#1041, v0.9.0): when this view has registered
+        :class:`~djust.components.LiveComponent` instances in
+        ``self._components`` (or as descriptor-pattern attributes), each
+        component's public state is captured under the special
+        ``"__components__"`` key as a ``{component_id: {field: value}}``
+        nested dict. The reserved name keeps component snapshots out of
+        the parent's flat state namespace and gives the time-travel
+        debug panel a clean shape to render per-component scrubbers.
+
         The server never calls this directly — it's primarily exposed for
         testing and observability. Restoration uses
         :meth:`_restore_snapshot`.
@@ -712,7 +721,59 @@ class LiveView(
             except (TypeError, ValueError, OverflowError):
                 # Skip non-serializable — matches _get_private_state pattern.
                 continue
+
+        # Component-level state (#1041). Each registered component
+        # contributes its own public-state dict under
+        # ``__components__`` keyed by ``component_id``.
+        components_state = self._capture_components_snapshot()
+        if components_state:
+            result["__components__"] = components_state
         return result
+
+    def _capture_components_snapshot(self) -> Dict[str, Dict[str, Any]]:
+        """Return a ``{component_id: {field: value}}`` snapshot map.
+
+        Walks ``self._components`` (the registry of LiveComponent
+        instances populated by ``_assign_component_ids``) and captures
+        each component's public state. Returns an empty dict when the
+        registry is empty or absent — no key is added to the parent
+        snapshot in that case.
+
+        Capture rules per component:
+        - Public (non-``_``-prefixed) attrs from ``component.__dict__``.
+        - Skip callables and non-serializable values (same rules as
+          parent state).
+        - Skip ``_descriptor_*`` machinery and other framework-
+          internal attrs that begin with ``_``.
+
+        Failures on individual components are logged and the bad
+        component is skipped — degrade gracefully rather than break
+        the whole snapshot.
+        """
+        registry = getattr(self, "_components", None)
+        if not registry:
+            return {}
+        components_state: Dict[str, Dict[str, Any]] = {}
+        for component_id, component in registry.items():
+            try:
+                comp_state: Dict[str, Any] = {}
+                for key, value in component.__dict__.items():
+                    if key.startswith("_"):
+                        continue
+                    if callable(value):
+                        continue
+                    try:
+                        comp_state[key] = json.loads(json.dumps(value, cls=DjangoJSONEncoder))
+                    except (TypeError, ValueError, OverflowError):
+                        # Skip non-serializable — matches parent rule.
+                        continue
+                components_state[component_id] = comp_state
+            except Exception:  # noqa: BLE001 — dev-only, log + degrade
+                logger.exception(
+                    "time_travel: component snapshot failed for id=%s",
+                    component_id,
+                )
+        return components_state
 
     def _restore_snapshot(self, state: Dict[str, Any]) -> None:
         """Apply a previously captured snapshot to this view.
--- a/python/djust/live_view.py
+++ b/python/djust/live_view.py
@@ -125,6 +125,26 @@ _FRAMEWORK_INTERNAL_ATTRS: frozenset = frozenset(
 )
 
 
+# Component-level analog (#1041): framework-internal LiveComponent
+# attrs that DON'T start with ``_`` but aren't user state. Excluded
+# from the ``__components__`` snapshot to keep the time-travel state
+# focused on actual user state. Mirrors the parent's
+# :data:`_FRAMEWORK_INTERNAL_ATTRS` for component fields.
+#
+# ``component_id`` in particular is the registry key — restoring it
+# from stale snapshot state would desync the registry from the
+# instance's ``component_id`` attribute.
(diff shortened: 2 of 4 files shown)


johnrtipton (commented):
## Stage 11 Code Review — PR #1141 (feat: component-level time-travel)

### Verdict: **COMMENT** (would-be APPROVE; same-author --approve blocked)

The implementation is small, clean, well-commented, and lands the right shape: reserved `__components__` key, registry-keyed nested dict, three-phase restore that pulls the component slice out before parent ghost-attr cleanup. CHANGELOG is thorough and class-named per PR #1106 canonicalization. Tests pass (35/35 in `test_time_travel.py`, all 7 new cases under `TestComponentLevelTimeTravel`). No security auto-reject triggers (no `mark_safe`, no f-string logging, no `console.log`, no bare `except: pass`).

That said, I want to flag four substantive findings — none blocking, but a couple are worth a follow-up issue.

---

### 1. Docstring overstates coverage: descriptor-pattern components are silently skipped (Important — docs)

`live_view.py:687-694` says the snapshot captures components in `self._components` **"(or as descriptor-pattern attributes)"**. The implementation only walks `self._components` (line 753). Descriptor-pattern components (`LiveComponent` subclasses with a `class State` inner class — see `components/base.py:516-578`) store state at `obj.__dict__["_component_<name>"]`, **not** in `self._components`. Confirmed by grep: only `mixins/components.py:_register_component` writes to `_components`, and descriptor `__set_name__` does not.

Net effect: descriptor-pattern components are silently excluded from time-travel snapshots. Probably fine for an initial cut (no production examples use the pattern yet), but the docstring should match reality. Suggest:
- Drop the `(or as descriptor-pattern attributes)` parenthetical, OR
- Open a follow-up issue (v0.9.x) to walk `_component_<name>` keys in `__dict__` too.

### 2. Component capture filter omits `template_name` exclusion the legacy helper has (Important — consistency)

The existing component-state helper at `mixins/components.py:71` explicitly excludes `template_name`:

```python
if not key.startswith("_") and key not in ("template_name",):
```

The new `_capture_components_snapshot` filters only `_*` and callables — no `_FRAMEWORK_INTERNAL_ATTRS` equivalent for components. Empirically:

```
LiveComponent().__dict__ = ['_descriptor_defaults', '_descriptor_attr_name',
  '_descriptor_storage_key', '_validated_assigns', 'component_id', '_mounted',
  '_parent', '_parent_callback']
```

Of the public attrs only `component_id` survives the filter. `template_name` is class-level, so currently safe. **But** any LiveComponent that does `self.template_name = "x"` in `mount()` (legal pattern) or sets any other config-like attr will see it captured AND restored on time-travel rewind. That's likely not desired. Two fixes possible:

- Add a `_COMPONENT_INTERNAL_ATTRS = frozenset({"component_id", "template_name", "template", ...})` skiplist mirroring the parent's `_FRAMEWORK_INTERNAL_ATTRS`, OR
- Reuse `mixins/components.py:_extract_component_state` directly (DRY).

Capturing `component_id` is mostly harmless on round-trip (same value back), but a hand-edited snapshot that mutates `component_id` would desync the registry key from the instance attribute — `view._components["alpha"].component_id == "beta"` is a foot-gun.

### 3. Missing test: defense-layer block for malicious component restore keys (Important — coverage)

The PR uses `safe_setattr(component, key, value, allow_private=False)` for component restoration — same defense as parent state. There's a Stage 11 expectation (per `docs/PULL_REQUEST_CHECKLIST.md` for security-touching code) that the defense layer has at least one regression test. Parent state has this; components don't. Suggested test:

```python
def test_restore_blocks_dangerous_component_attr_keys(self):
    """safe_setattr must block dunder/private restore on components, same as parent."""
    view = _ParentWithComponents()
    snap = EventSnapshot(
        event_name="evt", params={}, ref=1, ts=0.0,
        state_before={"__components__": {"alpha": {"__class__": object, "active": "q9"}}},
        state_after={},
    )
    ok = restore_snapshot(view, snap, which="before")
    assert ok is False  # blocked → returns False
    assert type(view._components["alpha"]) is _FakeComponent  # untouched
    assert view._components["alpha"].active == "q9"  # legitimate field still applied
```

This locks in the same security guarantee at the component layer that PR #1023 locked in for parents.

### 4. Phase 3 ordering — parent attrs may reflect stale component derivations (Suggestion — comment-only)

Phase 1 (parent ghost cleanup) → Phase 2 (parent setattr) → Phase 3 (per-component setattr). If a parent attr is a **derived** value of component state (e.g. `self.total = sum(c.value for c in components.values())` computed in an event handler), restoring `state_before` will leave the parent's `total` reflecting the post-mutation parent attr value, while the components are at their pre-mutation values. For time-travel-as-snapshot semantics this is correct — restore the literal recorded state, don't recompute. Worth a one-line comment in `restore_snapshot` so a future contributor doesn't try to "fix" the ordering. Not blocking.

---

### Items I checked off positively

- **Proxy chain in test stub** (review focus #1) — `_ParentWithComponents._capture_snapshot_state` and `_capture_components_snapshot` both forward via `LiveView._capture_snapshot_state(self)` / `LiveView._capture_components_snapshot(self)` (unbound method calls). MRO dispatch from inside `LiveView._capture_snapshot_state` resolves `self._capture_components_snapshot` to the stub, which forwards back. No infinite loop, no shadowing. Clean.
- **`__components__` user-attr collision** (review focus #4, #5) — `__components__` starts with `_`, so the parent capture loop already skips it (live_view.py:702) AND the Phase 1 ghost-cleanup loop excludes `_`-prefixed keys (time_travel.py:258). Pre-PR, a user attr `self.__components__` was silently dropped from the snapshot anyway. No back-compat break.
- **`safe_setattr` on `__components__` key directly** — `is_safe_attribute_name("__components__")` returns False (dunder block at attribute_guard.py:177), so even if the strip regressed, the key would be safely refused, not written through.
- **Aliasing fix** (review focus #6) — confirmed at live_view.py:766. `test_state_before_and_after_disconnect_from_live_components` locks it in.
- **Capture rules consistency** — same `_*`/callable/JSON-roundtrip rules as parent (modulo the `_FRAMEWORK_INTERNAL_ATTRS` gap noted in #2 above).
- **Logging hygiene** — all `%s`-style, no f-strings.

### CHANGELOG / canonicalizations

- Class-named in CHANGELOG (`7 new cases in TestComponentLevelTimeTravel`) per PR #1106 rule. ✓
- No `setdefault` patterns affected (PR #1103 rule N/A).
- No bulk dispatch-site changes (PR #1125 N/A).

### Verdict

**APPROVE-equivalent (COMMENT due to same-author).** Findings #1 and #2 are doc/semantic polish I'd like to see addressed in this PR or as a follow-up; #3 is a coverage ask; #4 is a comment-only note. None are auto-reject.

🟡 should-fix (this PR or follow-up):
- Docstring claims descriptor-pattern coverage that the implementation doesn't deliver — reword or scope-expand
- `_capture_components_snapshot` lacks the `_FRAMEWORK_INTERNAL_ATTRS`-equivalent skiplist (`component_id`, `template_name`, `template`) the legacy `_extract_component_state` already maintains — DRY violation, plus risk of capturing `component_id` and creating registry-key/instance-attr drift on hand-edited snapshots
- Add one regression test for `safe_setattr` blocking dangerous component restore keys

🟢 nice-to-have:
- One-liner comment in Phase 3 of `restore_snapshot` explaining "parent attrs derived from components are NOT recomputed; that's by design — time-travel restores literal state, not invariants"

johnrtipton:
# PR #1141 Retro — v0.9.0 #1041: component-level time-travel

**Date:** 2026-04-27
**Scope:** Extend v0.6.1 time-travel to capture per-component state
under reserved `__components__` key. Restore dispatches to
`view._components[component_id]` via safe_setattr.
**Tests at close:** 38/38 (37 existing + 7 component cases + 3
defense-layer + 1 internal-attr exclusion).
**LoC:** +413 / +6 across 3 files post-address-findings.

## What Worked Well

1. **Stage-4 first-principles applied (per #1032 retro learning).**
   The original task framing implied a substantial new "Phase 2
   captures component state" subsystem. Reading the existing
   `_capture_snapshot_state` first showed it already walks
   `self.__dict__` cleanly; the missing piece was just iterating
   `self._components`. ~100 LoC core change instead of a green-field
   rewrite.

2. **Reserved `__components__` key avoids namespace collision.** Dunder
   prefix means safe_setattr blocks accidental writes; the existing
   parent-state cleanup loop already filters `_*`-prefixed keys, so
   no back-compat hazard for users with a public attr by that name
   (their attr was already silently dropped pre-PR).

3. **Stage 11 surfaced the `component_id` desync risk** I'd missed.
   Without `_COMPONENT_INTERNAL_ATTRS`, restoring a stale snapshot
   would mutate `component.component_id` away from its registry key,
   silently breaking event dispatch. Added the exclusion + a
   regression test.

## What Didn't Work

1. **Test stub proxy chain for `_capture_components_snapshot`.**
   When borrowing `LiveView._capture_snapshot_state` in the test
   stub, the helper calls `self._capture_components_snapshot()`
   which the stub doesn't have natively. Initially failed with
   AttributeError. Fix: stub also borrows
   `LiveView._capture_components_snapshot` as a sibling method
   proxy. Reviewer verified no infinite loop.

2. **`_COMPONENT_INTERNAL_ATTRS` initially missing.** I had only
   filtered `_*`-prefix and callables; framework config attrs like
   `template_name` and `component_id` (which doesn't start with `_`)
   leaked into the snapshot. Stage 11 caught it; added the frozenset
   alongside the parent's `_FRAMEWORK_INTERNAL_ATTRS`.

## What to Improve

1. **Action Tracker row — descriptor-pattern component capture
   verification.** The reviewer noted descriptor-pattern components
   store state at `obj.__dict__["_component_<name>"]`. My
   implementation walks `self._components`. As long as
   `_assign_component_ids` populates `self._components` from
   descriptor-pattern attrs at mount time, both patterns are
   captured equivalently — but I haven't directly verified this with
   a test. File a follow-up to add an integration test against a
   descriptor-pattern view.

2. **Action Tracker row — debug panel UI for per-component
   scrubbing.** Server-side capture/restore are done; the debug
   panel still shows a flat scrub timeline. v0.9.x or v1.x: extend
   the panel UI to render per-component scrub bars. Out of scope
   for #1041 (server side only) but tracked.

- (None for CLAUDE.md until v0.9.0 milestone retro after #1042 ships.)

## Recurring Issue Tracker

- #1134 (5 flaky Python tests) — still skipped.
- New: vitest unhandled-rejection in `view-transitions.test.js`.
- New: `_AsCompletedIterator._wait_for_one` warning on cancellation.

## Review Stats

| Metric | Value |
|--------|-------|
| Tests added | 10 (7 capture/restore + 3 defense-layer + internal-attr) |
| 🔴 Findings | 0 |
| 🟡 Findings | 3 (all addressed) |
| 🟢 Nice-to-have | 1 (addressed) |
| Stage 11 verdict | COMMENT (would-be APPROVE) |
| LoC core | ~115 |
| LoC test | ~265 |
| LoC docs/CHANGELOG | ~45 |

PR #1155, opened 2026-04-27
fix(client): passthrough rate-limit for dj-input on click-fired widgets (closes #1154)

### LINKED ISSUE #1154: DEFAULT_RATE_LIMITS missing entries for click-fired widgets: 300ms debounce fallback delays radios/checkboxes on dj-input
## Summary

When a page binds `dj-input` to a radio, checkbox, or select (e.g. via the `WizardMixin.wizard_input_event = "dj-input"` class attribute that #1095 introduced), djust's client silently applies a **300ms debounce** before sending the WebSocket event. For text inputs that's correct — it batches keystrokes. For click-fired widgets there's no stream to batch, so it's just dead air the user experiences as lag.

## Reproduction (measured)

`tracescript_event` on a radio click in a production wizard:

| | click → WS send | Total click → DOM |
|---|---|---|
| Before workaround | **1104 ms** | ~1150 ms |
| After workaround (per-widget `dom_event="dj-change"` override) | **1 ms** | ~75 ms |

~15× faster just by switching from `dj-input` to `dj-change` on radios.

## Root cause

`python/djust/static/djust/src/08-event-parsing.js` — `DEFAULT_RATE_LIMITS`:

```js
const DEFAULT_RATE_LIMITS = {
    'range':    { type: 'throttle', ms: 150 },
    'number':   { type: 'throttle', ms: 100 },
    'color':    { type: 'throttle', ms: 150 },
    'text':     { type: 'debounce', ms: 300 },
    'search':   { type: 'debounce', ms: 300 },
    'email':    { type: 'debounce', ms: 300 },
    'url':      { type: 'debounce', ms: 300 },
    'tel':      { type: 'debounce', ms: 300 },
    'password': { type: 'debounce', ms: 300 },
    'textarea': { type: 'debounce', ms: 300 }
};
```

`09-event-binding.js:968` — the input handler falls back to `{ type: 'debounce', ms: 300 }` when `inputType` isn't in the map:

```js
var rateLimit = Object.prototype.hasOwnProperty.call(DEFAULT_RATE_LIMITS, inputType)
    ? DEFAULT_RATE_LIMITS[inputType]
    : { type: 'debounce', ms: 300 };
```

Widgets not in the map, all of which fire the native `input` event:
- `radio` (via `<input type="radio">`)
- `checkbox` (via `<input type="checkbox">`)
- `select-one` / `select-multiple` (native `<select>`)

All of these hit the 300ms fallback. The fallback is reasonable *defense* for unknown widgets (prevents floods) but these three are known — they commit exactly once per user interaction, no debounce needed.

## Why this matters for `WizardMixin`

`WizardMixin.wizard_input_event = "dj-input"` (added in #1095) was designed to capture pre-filled text values before blur. The docs recommend setting it **class-wide** on the wizard view. When set class-wide, it applies to every field the wizard renders via `as_live_field` — including radios and selects. A user with a simple Yes/No radio in their wizard inherits a 300ms dead air on every click without any indication why.

This was not obvious from the release notes — we spent time investigating what we thought was a server-side render cost before `trace_event` showed the delay was entirely client-side debounce.

## Proposed fix

Extend `DEFAULT_RATE_LIMITS` with passthrough entries for click-fired widgets:

```js
const DEFAULT_RATE_LIMITS = {
    // ... existing entries ...

    // Click-fired widgets: user commits one value per interaction,
    // no event stream to batch. Passthrough = no debounce, no throttle.
    'radio':           { type: 'passthrough' },
    'checkbox':        { type: 'passthrough' },
    'select-one':      { type: 'passthrough' },
    'select-multiple': { type: 'passthrough' },
};
```

And branch in the input handler:

```js
// In 09-event-binding.js ~line 991
if (rateLimit.type === 'passthrough') {
    wrapped = rawHandler;  // no wrapping
} else if (rateLimit.type === 'blur') {
    // ... existing blur branch
```

Defensive fallback for unknown types stays intact.

## Alternative: auto-pick `dj-change` for click-fired widgets in `as_live_field`

An adjacent fix on the Python side: have `WizardMixin.as_live_field` emit `dj-change` (not `dj-input`) for `RadioSelect` / `CheckboxInput` / `Select` widgets regardless of `wizard_input_event`. This sidesteps the debounce issue entirely — `change` events don't go through the rate-limit map. I implemented this in our fork as the immediate workaround.

Both fixes are compatible; either alone solves the user-visible problem. The client-side change is broader (any app using `dj-input` on radios/checkboxes benefits, not just wizards).

## Repro repo

Happy to provide a minimal reproduction if useful — it's essentially any wizard with `wizard_input_event = "dj-input"` and a `forms.ChoiceField(widget=forms.RadioSelect)`.

---

Related: #1095 (feature that introduced the `wizard_input_event` class attribute)

johnrtipton:
## Summary

- `DEFAULT_RATE_LIMITS` now has entries for `radio`, `checkbox`, `select-one`, `select-multiple` (new `passthrough` type)
- Input handler branches on `passthrough` and fires the raw handler synchronously instead of wrapping in a debounce
- Text/textarea debounce unchanged; defensive 300ms fallback for unknown types intact

Closes #1154.

## Background

`WizardMixin.wizard_input_event = \"dj-input\"` (added in #1095) is the recommended class-wide setting for wizards that need to capture pre-filled text values without waiting for blur. When set class-wide, it applies to every field the wizard renders — including radios, checkboxes, and selects.

Those widget types weren't in `DEFAULT_RATE_LIMITS`, so djust's client fell through to the input handler's fallback (`{ type: 'debounce', ms: 300 }`). The user experienced a 300ms stall on every Yes/No click before the WS event even left the browser. The felt delay looked like server slowness on a profiler but was entirely client-side.

## Measurement

`trace_event` on a production VPD wizard (client event → WS frame → DOM patch):

|  | click → WS send | server | client apply | total |
|---|---|---|---|---|
| Before | **1104 ms** | 44 ms | 5 ms | ~1150 ms |
| After  | **1 ms**    | 72 ms | 2.5 ms | **~75 ms** |

~15× end-to-end.

## Why passthrough is the right default for these widgets

`radio`, `checkbox`, `select-one`, and `select-multiple` all commit exactly **one value per user interaction**:
- A radio click can't produce a stream — subsequent clicks on the same group still only fire one `input` event each
- A checkbox click toggles once, fires once
- `<select>` commits on each option selection, not continuously

There's nothing to batch. The 300ms debounce is only protective against a stream of `input` events from text-typing, number-spinning, color-pickering, or range-dragging — none of which apply to a Yes/No click.

## Compatibility

- `dj-debounce=\"200\"` on a radio still applies — passthrough is the map-level default, user-supplied `dj-*` attributes take precedence (verified by `test('dj-debounce override still applies to a radio if set explicitly')`).
- `dj-throttle` on a checkbox still applies the same way.
- Text/textarea and all existing map entries unchanged.
- Defensive 300ms fallback for any widget type still not in the map (e.g. custom web components) intact.

## Tests

7 new cases in `tests/js/dj-input-click-widgets.test.js`:
- `radio input fires synchronously, not after 300ms`
- `checkbox fires synchronously`
- `select-one fires synchronously`
- `text input still debounces 300ms (unchanged behavior)`
- `textarea still debounces 300ms`
- `rapid radio clicks each produce an event (no debounce coalescing)`
- `dj-debounce override still applies to a radio if set explicitly`

Full JS suite: **1434 passed** (was 1427, +7 new cases, all green).

## Test plan

- [x] `npx vitest run tests/js/dj-input-click-widgets.test.js` — 7/7 pass
- [x] `npx vitest run` — 1434/1434 pass across 121 test files
- [x] `bash scripts/build-client.sh` — rebuilds bundled `client.js`

## Diff

- `python/djust/static/djust/src/08-event-parsing.js` — 4 new map entries + 3-line comment
- `python/djust/static/djust/src/09-event-binding.js` — 7-line `passthrough` branch
- `python/djust/static/djust/client.js` — regenerated via `build-client.sh`
- `tests/js/dj-input-click-widgets.test.js` — 7 cases, ~200 LoC
- `CHANGELOG.md` — Unreleased entry

Related: #1095 (feature that introduced `wizard_input_event`)


johnrtipton:
## Code Review — PR #1155 (passthrough rate-limit for click-fired widgets)

Strong fix on the user-visible problem (1104ms → 1ms on radio click) with measured before/after, defensive override semantics preserved, and 7 regression tests. CI is green where it has finished. One 🔴 to fix before merge plus one 🟡.

### 🔴 Shared-state mutation: explicit `dj-debounce`/`dj-throttle` on one radio pollutes the default for all subsequently-bound radios

`python/djust/static/djust/src/09-event-binding.js:968` reads `DEFAULT_RATE_LIMITS[inputType]` *by reference* and the override branches at lines 974-988 then mutate that referenced object directly:

```js
var rateLimit = Object.prototype.hasOwnProperty.call(DEFAULT_RATE_LIMITS, inputType)
    ? DEFAULT_RATE_LIMITS[inputType]            // ← shared const reference
    : { type: 'debounce', ms: 300 };

if (inputEl.hasAttribute('dj-debounce')) {
    ...
    rateLimit.type = 'debounce';                // ← mutates DEFAULT_RATE_LIMITS[inputType]
    rateLimit.ms = parseInt(djVal, 10);
}
```

**Pre-PR**: this latent bug was hidden for click-fired widgets because they fell through to the inline fallback object `{ type: 'debounce', ms: 300 }` — a fresh object per binding, so mutation only hit the throwaway.

**Post-PR**: `radio` / `checkbox` / `select-one` / `select-multiple` are now persistent shared const entries in the map. The first radio that uses `dj-debounce="200"` permanently flips `DEFAULT_RATE_LIMITS['radio']` to `{ type: 'debounce', ms: 200 }`. Every subsequent radio with no `dj-debounce`/`dj-throttle` will then bind with `wrapped = debounce(rawHandler, 200)` — silently losing the new passthrough default.

**User-visible scenario**: a wizard with one explicit `dj-debounce="200"` radio for a special field + many default Yes/No radios. After the user touches the explicit radio first, all subsequent default radios on the page (or any later-mounted view, since `DEFAULT_RATE_LIMITS` is module-level) inherit the polluted 200ms debounce — re-introducing the very lag this PR is supposed to remove. Dependence on click order also makes it intermittent, which is the worst kind of regression for a perf fix.

**Fix**: clone before mutating.

```diff
-                var rateLimit = Object.prototype.hasOwnProperty.call(DEFAULT_RATE_LIMITS, inputType) ? DEFAULT_RATE_LIMITS[inputType] : { type: 'debounce', ms: 300 };
+                var rateLimit = Object.prototype.hasOwnProperty.call(DEFAULT_RATE_LIMITS, inputType)
+                    ? Object.assign({}, DEFAULT_RATE_LIMITS[inputType])
+                    : { type: 'debounce', ms: 300 };
```

(Apply in both `python/djust/static/djust/src/09-event-binding.js:968` and the bundled `python/djust/static/djust/client.js:~4060`. `Object.assign` is the existing client.js convention; no new helper needed.)

**Regression test** for `tests/js/dj-input-click-widgets.test.js`:

```js
it('dj-debounce on one radio does not pollute defaults for siblings', () => {
    const dom = createTestEnv(`
        <div dj-root dj-view="test.View">
            <input type="radio" name="A" value="a" dj-input="hA" dj-debounce="200" />
            <input type="radio" name="B" value="b" dj-input="hB" />
        </div>
    `);
    initClient(dom);

    const a = dom.window.document.querySelector('[name="A"]');
    const b = dom.window.document.querySelector('[name="B"]');

    // First, click the override radio — this is when its wrapper is built and
    // (in the buggy version) DEFAULT_RATE_LIMITS['radio'] gets mutated.
    a.checked = true;
    a.dispatchEvent(new dom.window.Event('input', { bubbles: true }));

    // Now click the default radio — must fire SYNCHRONOUSLY (passthrough),
    // not be debounced by the polluted shared default.
    b.checked = true;
    b.dispatchEvent(new dom.window.Event('input', { bubbles: true }));

    expect(getFetchCalls(dom).map(c => c.eventName)).toContain('hB');
    // Without the fix, this asserts 0 (B's wrapper was built with debounce 200).
});
```

This exercise also tightens the existing "rapid radio clicks each produce an event" coverage to a different element having a different rate-limit policy.

### 🟡 `select-multiple` entry added but not tested

`DEFAULT_RATE_LIMITS['select-multiple']` is added (08-event-parsing.js:60, client.js:2643), but no test asserts the synchronous-fire behavior for `<select multiple>`. The other three new entries (`radio`, `checkbox`, `select-one`) each have a sync-fire test. The CLAUDE.md "Mechanical replacement: N similar sites need N tests" canonicalization (rule #1104, post-#1102 retro) applies here — 4 entries added, 3 tested.

**Fix**: add a brief case mirroring the `select-one` test, with `<select multiple>` and a `dj-input` triggering on a single option-click `input` event.

### Strengths

- **Excellent measurement evidence**: the trace_event before/after table (1104ms → 1ms client-side, ~15× end-to-end) directly anchors the fix to the user-reported delay.
- **Override semantics preserved**: explicit `dj-debounce`/`dj-throttle` still wins over the new passthrough default, locked in by the override test.
- **Regression coverage**: text/textarea 300ms debounce explicitly re-asserted (3 of the 7 tests are regression locks for unchanged behavior — exactly the right shape for a default-table change).
- **CHANGELOG** entry is precise about the behavior delta, the cause (`DEFAULT_RATE_LIMITS` map gap), and the surface area (4 new entries + 7-line branch).
- **No security impact** despite the `security-review` label — handler still consults `e.target.value`/`e.target.checked` the same way; no new injection or escalation surface. Worth a quick comment from the labeller, but I don't see a real concern here.
- **Comments**: the 3-line block above `DEFAULT_RATE_LIMITS` and the inline comments per new entry explain *why* passthrough is correct (one value per interaction, no stream).

### Verdict

**REQUEST_CHANGES** — fix the shared-state mutation (🔴) and add the `select-multiple` test (🟡). Both are small targeted changes; the regression test I sketched should fail RED on the current branch and pass GREEN after the `Object.assign({}, ...)` clone.

Once those land I'd APPROVE without further review.

johnrtipton:
Findings addressed in c6663fe0:

- 🔴 Shared-state mutation: cloned DEFAULT_RATE_LIMITS[inputType] with `Object.assign({}, …)` before the override branches mutate. Pure 1-line fix at 09-event-binding.js:967-973 + 3 lines of explanatory comment.
- 🟡 Added select-multiple sync-fire test.
- Plus a regression test that exercises the shared-state pollution scenario directly (two radios, one with `dj-debounce="200"` first, then a default radio that must fire synchronously). This test fails RED on the buggy code path; passes GREEN with the clone.

Test counts: `tests/js/dj-input-click-widgets.test.js` 9/9 (was 7/7); full JS suite 1436/1436 (was 1434/1434). Bundled `client.js` rebuilt.

Pre-commit hooks all green (build client.js, npm test, eslint, detect secrets, CHANGELOG test counts). Will merge once CI finishes.

PR #1157, opened 2026-04-27
feat(wizard): as_live_field auto-picks dom_event by widget class (closes #1156)

### LINKED ISSUE #1156: WizardMixin.as_live_field: auto-pick dom_event by widget type (dj-change for click-fired, dj-input for text streams)
## Summary

`WizardMixin.wizard_input_event` (from #1095) is the class-wide knob a wizard author uses to choose between `dj-input` (stream-based, debounced — right for text fields users haven't blurred yet) and `dj-change` (commit-based, fires on blur — right for click-fired widgets). Today the setting applies uniformly to **every** field the wizard renders. That's rarely what the author wants: a wizard with a mix of text inputs and Yes/No radios has to pick one trade-off that's wrong for half the fields.

Proposal: `WizardMixin.as_live_field()` should auto-pick the right `dom_event` per widget type, and `wizard_input_event` should scope to "text-like widgets only."

## Motivation

`dj-input` is semantically a streaming event. It fires continuously as the user types. Debouncing it is correct — you batch keystrokes before firing a server round trip. Without `dj-input`, a user who types "Amanda" into `first_name` and clicks Next without first blurring the field loses the edit. That's the problem #1095 solved.

`dj-change` is semantically a commit event. It fires once, after the user commits a value — on blur for text, on click for radios/checkboxes, on option-select for `<select>`. There's no stream to batch.

Applying `wizard_input_event = "dj-input"` class-wide conflates the two:
- Text fields get the right event (dj-input, debounced, captures unblurred edits) ✅
- Radio fields get the wrong event (dj-input on a commit-widget — #1154 made it fast, but it's still semantically a streaming event where no stream exists)
- Checkbox/Select same story as radio

#1154/#1155 fixed the **performance** symptom by making `dj-input` on click-fired widgets non-debounced via `passthrough`. But the deeper issue is semantic — we're telling the framework "fire on every keystroke" on a widget that doesn't have keystrokes. That's a smell.

## Proposed change

`WizardMixin.as_live_field()` currently always uses `self.wizard_input_event` as the DOM event. Make it widget-aware:

```python
# python/djust/wizard.py  (simplified pseudocode)
def as_live_field(self, field_name, event_name="validate_field", **kwargs):
    # ... existing field lookup ...

    if kwargs.get("dom_event") is None:
        widget = field.widget
        # Click-fired widgets: dj-change is the correct semantic event.
        # Text-stream widgets: dj-input respects wizard_input_event so #1095's
        # "capture unblurred text" scenario keeps working.
        if isinstance(widget, (forms.RadioSelect,
                               forms.CheckboxInput,
                               forms.CheckboxSelectMultiple,
                               forms.Select)):
            kwargs["dom_event"] = "dj-change"
        else:
            kwargs["dom_event"] = self.wizard_input_event
    return adapter.render_field(...)
```

Net effect:
- `wizard_input_event = "dj-change"` (default): nothing changes — every widget still emits `dj-change`.
- `wizard_input_event = "dj-input"` (opt-in, #1095 use case): text/textarea emit `dj-input` (debounced, captures unblurred edits). Radios/checkboxes/selects emit `dj-change` (commits, no debounce, semantically correct).
- Caller-passed `dom_event="..."` still wins (existing behavior) — this is only a smarter default.

## Follow-up to #1155

#1155 fixes the immediate perf regression. This is the "do the right thing by default" follow-up: apps that set `wizard_input_event = "dj-input"` never needed that setting to apply to radios, and with this change, they don't have to care.

Downstream impact: apps that have implemented their own `as_live_field` override to do exactly this pattern (I did, during #1154 investigation) can delete it.

## Alternative: rename `wizard_input_event` → `wizard_text_event`

If the semantic scope (text-only) is the intent, the attribute name should say so. Renaming would be a bigger breaking change — current approach keeps the name and just scopes its effect to text widgets. Low-cost either way.

## Test plan

- Radio widget emits `dj-change="validate_field"` regardless of `wizard_input_event`
- Select widget same
- Checkbox widget same
- Text input emits `wizard_input_event` (dj-input if set, else dj-change)
- Textarea same as text
- Caller-passed `dom_event=` still overrides (existing contract preserved)

## Repro of the semantic awkwardness

Any wizard that sets `wizard_input_event = "dj-input"` and has a `forms.ChoiceField(widget=forms.RadioSelect)`. Current output HTML has `<input type="radio" dj-input="validate_field">`, which is nonsensical — there's no `input` event stream on a radio click (well, there's exactly one). After the proposed change, the output is `<input type="radio" dj-change="validate_field">`, which matches how every hand-written djust app would wire a radio.

---

Related:
- #1095 — added `wizard_input_event`
- #1154 / #1155 — client-side perf fix (passthrough for click-fired widgets)

johnrtipton:
## Summary

- `WizardMixin.as_live_field()` inspects the field's widget class and picks `dj-change` for click-fired widgets (Radio/Checkbox/Select), `wizard_input_event` for text-stream widgets
- Caller-passed `dom_event=\"...\"` still wins — the widget-aware default is a default, not a mandate
- New `_CLICK_FIRED_WIDGET_CLASSES` ClassVar lets apps extend the dispatch set for custom commit-style widgets without overriding `as_live_field`
- Apps that had implemented their own `as_live_field` override to do this mapping can now delete it

Closes #1156. Complements #1155 (client-side passthrough rate-limit for click-fired widgets).

## Background

#1095 introduced `wizard_input_event` so wizard authors could set \"dj-input\" class-wide and capture unblurred text edits (users who paste into a text field and hit Next without blurring first). The setting applied uniformly to every widget, which had two problems when the form contained a mix of widget types:

1. **Semantic**: `dj-input` is a streaming event. A radio doesn't stream — it commits one value per click. Emitting `<input type=\"radio\" dj-input=\"validate_field\">` into production HTML is technically functional but reads like a bug.
2. **Performance (pre-#1155)**: the djust client applied a 300ms debounce fallback to any `dj-input` element type not in its `DEFAULT_RATE_LIMITS` map. Radios/checkboxes/selects weren't in the map, so every click incurred a 300ms stall before the WS event left the browser.

#1155 fixed the performance symptom on the client. This PR fixes the semantic issue on the server by making the default *correct* instead of making a workaround fast.

## Implementation

```python
class WizardMixin:
    _CLICK_FIRED_WIDGET_CLASSES: ClassVar[frozenset[str]] = frozenset({
        \"RadioSelect\",
        \"CheckboxInput\",
        \"CheckboxSelectMultiple\",
        \"Select\",
    })

    def _default_dom_event_for(self, field) -> str:
        widget_cls = type(field.widget).__name__ if field.widget else \"\"
        if widget_cls in self._CLICK_FIRED_WIDGET_CLASSES:
            return \"dj-change\"
        return self.wizard_input_event

    def as_live_field(self, field_name, event_name=\"validate_field\", **kwargs):
        # ... existing field lookup ...
        if kwargs.get(\"dom_event\") is None:
            kwargs[\"dom_event\"] = self._default_dom_event_for(field)
        return adapter.render_field(...)
```

Uses string class names (not `isinstance`) so the ClassVar isn't a symbol-import from `django.forms` at module-load time, and so subclasses can extend it ergonomically:

```python
class MyWizard(WizardMixin, LiveView):
    _CLICK_FIRED_WIDGET_CLASSES = frozenset({
        *WizardMixin._CLICK_FIRED_WIDGET_CLASSES,
        \"MyColorPickerWidget\",
    })
```

The string-name approach also gracefully handles subclasses — if an app has `class MyRadioSelect(forms.RadioSelect)` and the class name matches, it's picked up automatically (intentional: anyone subclassing RadioSelect wants its commit semantics).

## Behavior matrix

| `wizard_input_event` | Widget | `dom_event` | Reason |
|---|---|---|---|
| `dj-change` (default) | Text | `dj-change` | Tracks attribute, no change |
| `dj-change` | Radio | `dj-change` | Unchanged (default already correct) |
| `dj-input` (opt-in, #1095) | Text | `dj-input` | #1095 preserved — unblurred-text capture |
| `dj-input` | Radio | `dj-change` | **Fixed** — radios commit, don't stream |
| `dj-input` | Select | `dj-change` | **Fixed** |
| `dj-input` | Checkbox | `dj-change` | **Fixed** |
| any | any, `dom_event=\"X\"` caller override | `X` | Caller always wins |

## Tests

New `TestAsLiveFieldWidgetAwareDomEvent` class in `tests/unit/test_wizard_mixin.py`, 12 cases:

**Text-stream widgets track `wizard_input_event`:**
- `test_text_input_uses_wizard_input_event_default` — default `dj-change`
- `test_text_input_uses_dj_input_when_streaming` — opt-in `dj-input`
- `test_textarea_uses_dj_input_when_streaming`
- `test_integer_input_uses_dj_input_when_streaming`
- `test_email_input_uses_dj_input_when_streaming`

**Click-fired widgets locked to `dj-change`:**
- `test_radio_uses_dj_change_even_when_streaming`
- `test_select_uses_dj_change_even_when_streaming`
- `test_checkbox_uses_dj_change_even_when_streaming`
- `test_checkbox_select_multiple_uses_dj_change_even_when_streaming`

**Escape hatches:**
- `test_caller_passed_dom_event_wins_on_click_widget` — `dj-input` on a radio if explicitly passed
- `test_caller_passed_dom_event_wins_on_text_widget` — `dj-change` on a text field if explicitly passed

**Subclass extension:**
- `test_subclass_can_extend_click_fired_set` — custom widget added to the ClassVar gets `dj-change`

## Test plan

- [x] `python -m pytest tests/unit/test_wizard_mixin.py` — **51 passed** (was 39, +12 new)
- [x] `python -m pytest tests/ python/tests/` — **3958 passed, 24 skipped** (full Python suite, no regressions)
- [x] Widget-class matching uses string names, so custom widget subclasses can extend without `isinstance` registration dance

## Backwards compatibility

- `wizard_input_event = \"dj-change\"` (the default): **no change** for any widget.
- `wizard_input_event = \"dj-input\"` on text-only wizards: **no change** — text widgets still emit `dj-input`.
- `wizard_input_event = \"dj-input\"` on wizards with radios/selects/checkboxes: **behavior change** — those widgets now emit `dj-change` instead of `dj-input`. This matches what #1095 was actually trying to accomplish (the author wanted `dj-input` for text fields; the spillover onto radios was incidental and buggy).
- Apps with a custom `as_live_field` override that replicates this logic can delete the override after upgrading.

Existing tests for #1095 behavior continue to pass (text widgets still emit `dj-input` when `wizard_input_event` is set to it).

## Diff

- `python/djust/wizard.py` — +63/-12 lines (helper + ClassVar + docstring updates)
- `tests/unit/test_wizard_mixin.py` — +171 lines (12 test cases)
- `CHANGELOG.md` — `[Unreleased]` > `### Changed` entry

---

Related:
- #1095 — added `wizard_input_event`
- #1154 / #1155 — client-side perf fix (passthrough for click-fired widgets)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,11 +18,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   semantically wrong (there's no keystroke stream to fire on) and pre-
   #1155 incurred a 300ms debounce stall on every click.
 
-  ``as_live_field`` now inspects the field's widget class and picks:
+  ``as_live_field`` now inspects the field's widget class (walking the
+  widget's MRO so any subclass of an enumerated builtin inherits the
+  default automatically) and picks:
 
-  - ``dj-change`` for click-fired widgets (``RadioSelect``,
-    ``CheckboxInput``, ``CheckboxSelectMultiple``, ``Select``) — they
-    commit exactly one value per user interaction, no stream to batch.
+  - ``dj-change`` for click-fired widgets — ``RadioSelect``,
+    ``CheckboxInput``, ``CheckboxSelectMultiple``, ``Select``, plus
+    every Django Select subclass (``SelectMultiple``,
+    ``NullBooleanSelect``) and any app's RadioSelect/Select subclass
+    matched via MRO. They commit exactly one value per user
+    interaction, no stream to batch.
   - ``wizard_input_event`` for text-stream widgets (``TextInput``,
     ``Textarea``, ``NumberInput``, ``EmailInput``, etc.) — preserves
     the #1095 contract for authors who need unblurred-text capture.
@@ -45,14 +50,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   ```
 
   Files: ``python/djust/wizard.py`` (new ``_default_dom_event_for``
-  helper + ``_CLICK_FIRED_WIDGET_CLASSES`` ClassVar, ~15 LoC; updated
+  helper + ``_CLICK_FIRED_WIDGET_CLASSES`` ClassVar, ~20 LoC; updated
   ``wizard_input_event`` docstring to clarify text-only scope).
-  12 new cases in ``TestAsLiveFieldWidgetAwareDomEvent`` in
+  15 new cases in ``TestAsLiveFieldWidgetAwareDomEvent`` in
   ``tests/unit/test_wizard_mixin.py`` cover text/textarea/integer/email
   tracking ``wizard_input_event``, radio/select/checkbox/
   CheckboxSelectMultiple locked to ``dj-change``, caller-passed
-  ``dom_event`` overrides, and subclass extension of the click-fired
-  set.
+  ``dom_event`` overrides, ClassVar extension for custom widgets, and
+  MRO walk catching ``SelectMultiple`` / ``NullBooleanSelect`` /
+  app-defined RadioSelect subclasses.
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,53 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- **`WizardMixin.as_live_field` auto-picks `dom_event` by widget class
+  (closes #1156)** — previously the view-level ``wizard_input_event``
+  attribute applied uniformly to every widget the wizard rendered. An
+  author setting ``wizard_input_event = "dj-input"`` to capture
+  unblurred text edits (per #1095) unintentionally also stamped
+  ``dj-input`` on radios, selects, and checkboxes — which was
+  semantically wrong (there's no keystroke stream to fire on) and pre-
+  #1155 incurred a 300ms debounce stall on every click.
+
+  ``as_live_field`` now inspects the field's widget class and picks:
+
+  - ``dj-change`` for click-fired widgets (``RadioSelect``,
+    ``CheckboxInput``, ``CheckboxSelectMultiple``, ``Select``) — they
+    commit exactly one value per user interaction, no stream to batch.
+  - ``wizard_input_event`` for text-stream widgets (``TextInput``,
+    ``Textarea``, ``NumberInput``, ``EmailInput``, etc.) — preserves
+    the #1095 contract for authors who need unblurred-text capture.
+  - Caller-passed ``dom_event="..."`` still wins — the widget-aware
+    default is a default, not a mandate.
+
+  Apps that had implemented their own `as_live_field` override to do
+  exactly this mapping can delete the override.
+
+  New ``_CLICK_FIRED_WIDGET_CLASSES`` ClassVar (frozenset of widget
+  class names) lets apps with custom commit-style widgets extend the
+  dispatch without overriding ``as_live_field`` itself:
+
+  ```python
+  class MyWizard(WizardMixin, LiveView):
+      _CLICK_FIRED_WIDGET_CLASSES = frozenset({
+          *WizardMixin._CLICK_FIRED_WIDGET_CLASSES,
+          "MyColorPickerWidget",
+      })
+  ```
+
+  Files: ``python/djust/wizard.py`` (new ``_default_dom_event_for``
+  helper + ``_CLICK_FIRED_WIDGET_CLASSES`` ClassVar, ~15 LoC; updated
+  ``wizard_input_event`` docstring to clarify text-only scope).
+  12 new cases in ``TestAsLiveFieldWidgetAwareDomEvent`` in
+  ``tests/unit/test_wizard_mixin.py`` cover text/textarea/integer/email
+  tracking ``wizard_input_event``, radio/select/checkbox/
+  CheckboxSelectMultiple locked to ``dj-change``, caller-passed
+  ``dom_event`` overrides, and subclass extension of the click-fired
+  set.
+
 ### Fixed
 
 - **`dj-input` on click-fired widgets no longer incurs a 300ms debounce
--- a/python/djust/wizard.py
+++ b/python/djust/wizard.py
@@ -147,12 +147,14 @@ class WizardMixin:
     #: can extend this frozenset to cover custom widgets (e.g. a color-picker
     #: component that fires a single commit event). Using string class names
     #: avoids pulling ``django.forms`` into this module at import time.
-    _CLICK_FIRED_WIDGET_CLASSES: ClassVar[frozenset[str]] = frozenset({
-        "RadioSelect",
-        "CheckboxInput",
-        "CheckboxSelectMultiple",
-        "Select",
-    })
+    _CLICK_FIRED_WIDGET_CLASSES: ClassVar[frozenset[str]] = frozenset(
+        {
+            "RadioSelect",
+            "CheckboxInput",
+            "CheckboxSelectMultiple",
+            "Select",
+        }
+    )
 
     def _default_dom_event_for(self, field) -> str:
         """Pick the correct default ``dom_event`` for a form field's widget.
@@ -163,11 +165,20 @@ class WizardMixin:
         a wizard that opts into ``"dj-input"`` for text (#1095) doesn't
         accidentally also apply dj-input semantics to radios — which have
         no stream to fire on.
+
+        Walks the widget's MRO so any subclass of an enumerated builtin
+        (e.g. ``SelectMultiple``/``NullBooleanSelect`` subclassing
+        ``Select``, or an app's ``MyRadioSelect`` subclassing
+        ``RadioSelect``) inherits the commit-style default automatically —
+        without forcing apps to register every subclass in
+        ``_CLICK_FIRED_WIDGET_CLASSES`` themselves.
         """
         widget = getattr(field, "widget", None)
-        widget_cls = type(widget).__name__ if widget is not None else ""
-        if widget_cls in self._CLICK_FIRED_WIDGET_CLASSES:
-            return "dj-change"
+        if widget is not None:
+            click_set = self._CLICK_FIRED_WIDGET_CLASSES
+            for cls in type(widget).__mro__:
+                if cls.__name__ in click_set:
+                    return "dj-change"
         return self.wizard_input_event
 
     def as_live_field(self, field_name: str, event_name: str = "validate_field", **kwargs) -> str:
--- a/python/djust/wizard.py
+++ b/python/djust/wizard.py
@@ -65,7 +65,7 @@ Several design decisions here work around known djust serialization constraints:
 
 import logging
 import math
-from typing import Any
+from typing import Any, ClassVar
 
 from .decorators import event_handler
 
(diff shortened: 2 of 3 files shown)


johnrtipton:
## Code Review — PR #1157 (widget-aware `dom_event` default)

Clean server-side companion to #1155: separates "I want unblurred-text capture" (#1095 contract) from "every widget gets dj-input" (the spillover bug). Caller-override semantics preserved, ClassVar extension API is the right shape, 12 new tests cover the dispatch matrix and escape hatches. CI is green where finished. One 🔴 + one 🟡 — both addressed by the same 3-line fix.

### 🔴 Leaf-class-only matching misses Django-builtin Select subclasses

`_default_dom_event_for` (`python/djust/wizard.py:163-168`) compares only `type(widget).__name__` against `_CLICK_FIRED_WIDGET_CLASSES`:

```python
widget_cls = type(widget).__name__ if widget is not None else ""
if widget_cls in self._CLICK_FIRED_WIDGET_CLASSES:
    return "dj-change"
return self.wizard_input_event
```

This is exact-string-match, so `Select` in the frozenset matches **only** the literal `forms.Select` class. Verified locally:

```
SelectMultiple            mro: ['SelectMultiple', 'Select', 'ChoiceWidget', 'Widget']
NullBooleanSelect         mro: ['NullBooleanSelect', 'Select', 'ChoiceWidget', 'Widget']
```

**`forms.SelectMultiple`** (the default widget for `forms.MultipleChoiceField`) and **`forms.NullBooleanSelect`** (the default widget for `forms.NullBooleanField`) are both click-fired Select subclasses, but their `__name__` doesn't match anything in the frozenset. They silently fall through to `wizard_input_event` — re-introducing the exact bug this PR fixes for the parent `Select`. A wizard with `wizard_input_event = "dj-input"` and a `MultipleChoiceField` (no explicit `widget=...`) emits `dj-input` on a multi-select picker, which has no event stream to fire on per the PR's own framing.

`CheckboxSelectMultiple` is fine in the leaf-only world only because the PR explicitly enumerated it. The pattern fails open for any unenumerated subclass.

The PR body actually claims this works ("if an app has `class MyRadioSelect(forms.RadioSelect)` and the class name matches, it's picked up automatically"), but the implementation does plain string equality — so `MyRadioSelect` is **not** matched. The body and the code disagree.

**Fix**: walk the MRO so any subclass of an enumerated class inherits its commit semantics.

```diff
     def _default_dom_event_for(self, field) -> str:
-        widget = getattr(field, "widget", None)
-        widget_cls = type(widget).__name__ if widget is not None else ""
-        if widget_cls in self._CLICK_FIRED_WIDGET_CLASSES:
-            return "dj-change"
+        widget = getattr(field, "widget", None)
+        if widget is not None:
+            click_set = self._CLICK_FIRED_WIDGET_CLASSES
+            for cls in type(widget).__mro__:
+                if cls.__name__ in click_set:
+                    return "dj-change"
         return self.wizard_input_event
```

This:
- Fixes `SelectMultiple` → matches `Select` via MRO → `dj-change` ✓
- Fixes `NullBooleanSelect` → matches `Select` via MRO → `dj-change` ✓
- Makes the PR-body claim about `MyRadioSelect` actually true.
- Preserves the leaf-name extension path: a custom class added to the ClassVar still matches at depth 0.
- Stays import-light: still no `django.forms` symbol import; only walks the widget object's own class chain.
- O(MRO depth) — typically 3-4 classes, so trivial cost on each `as_live_field` call.

### 🟡 Test gap: only the enumerated leaf classes are covered

The 12 tests assert dispatch for the four classes literally in `_CLICK_FIRED_WIDGET_CLASSES`, but never exercise:

1. **`forms.SelectMultiple`** — the default `MultipleChoiceField` widget.
2. **`forms.NullBooleanSelect`** — the default `NullBooleanField` widget.
3. **An app subclass of `forms.RadioSelect`** — exactly the scenario the PR body advertises as "picked up automatically".

After the MRO-walk fix, two tests cover all three gaps:

```python
@pytest.mark.django_db
def test_select_multiple_uses_dj_change_via_mro(self, get_request):
    """SelectMultiple is a Django builtin Select subclass — the leaf-name
    isn't in _CLICK_FIRED_WIDGET_CLASSES, but the MRO walk picks up Select."""

    class _Form(forms.Form):
        # forms.MultipleChoiceField's default widget is forms.SelectMultiple
        choices = forms.MultipleChoiceField(choices=[("a", "A"), ("b", "B")])

    class _Wizard(WizardMixin, LiveView):
        wizard_input_event = "dj-input"
        wizard_steps = [{"name": "x", "title": "X", "form_class": _Form}]
        template = "<div dj-root></div>"

    view = _Wizard()
    view.get(get_request)
    html = view.as_live_field("choices")
    assert 'dj-change="validate_field"' in html
    assert "dj-input" not in html


@pytest.mark.django_db
def test_app_subclass_of_radio_select_inherits_commit_semantics(self, get_request):
    """A custom RadioSelect subclass should be matched via MRO — apps shouldn't
    have to add every subclass to the ClassVar to get correct semantics."""

    class _MyRadio(forms.RadioSelect):
        pass

    class _Form(forms.Form):
        pick = forms.ChoiceField(
            choices=[("y", "Y"), ("n", "N")],
            widget=_MyRadio,
        )

    class _Wizard(WizardMixin, LiveView):
        wizard_input_event = "dj-input"
        wizard_steps = [{"name": "x", "title": "X", "form_class": _Form}]
        template = "<div dj-root></div>"

    view = _Wizard()
    view.get(get_request)
    html = view.as_live_field("pick")
    assert 'dj-change="validate_field"' in html
```

Both fail RED on the current branch and pass GREEN after the MRO-walk diff.

### Strengths

- **Right framing**: client-side perf fix (#1155) made the workaround free; this PR makes the workaround unnecessary by fixing the default semantically. The two PRs together close the issue properly rather than papering it over on one side.
- **Behavior matrix table in the PR body** (5×7 grid covering text/click × default/streaming/override) is the cleanest review aid I've seen on this codebase. Worth lifting that table style into future feature PRs.
- **ClassVar extension is the right API surface**: `class MyWizard: _CLICK_FIRED_WIDGET_CLASSES = frozenset({...WizardMixin._CLICK_FIRED_WIDGET_CLASSES, "MyWidget"})` is one line and doesn't require subclassing `as_live_field`. Test #12 locks this in.
- **Caller override preserved**: `kwargs.get("dom_event") is None` coalesce (not `setdefault`) — correctly avoids the mixin-forwarding gotcha from CLAUDE.md retro #1103. Tests #10 and #11 lock both directions in.
- **Docstrings updated in three places** (`wizard_input_event` ClassVar, new `_CLICK_FIRED_WIDGET_CLASSES`, `_default_dom_event_for` helper, `as_live_field`) — consistent and complete.
- **CHANGELOG**: correctly placed under `### Changed` (not `### Fixed`) since `wizard_input_event = "dj-input"` users see a behavior delta on radio/select/checkbox. The `[Unreleased]` window is the right place for this since v0.9.0 stable hasn't shipped — no breaking-change-vs-released-API concern.
- **No security impact**: pure widget-class introspection → string return; no user-controlled value reaches HTML or shell.
- **Backwards-compat analysis** in the PR body is honest: lists the no-op cases (default `dj-change`), the no-change cases (text-only wizards), and the actual delta (mixed-widget wizards on `dj-input`). Calls out that apps with custom `as_live_field` overrides for this exact mapping can now delete them — a real DX win.

### Verdict

**REQUEST_CHANGES** — apply the MRO-walk diff (3 LoC change replacing the leaf-name comparison) and add the 2 tests. The same fix simultaneously closes the 🔴 (`SelectMultiple` / `NullBooleanSelect` silently breaking) and the 🟡 (PR-body subclass-handling claim becomes accurate). After that I'd APPROVE.

johnrtipton:
Findings addressed in 3f16ea3d:

- 🔴 + 🟡: `_default_dom_event_for` now walks the widget's MRO. Inherits commit semantics from any enumerated builtin (`Select`, `RadioSelect`, `CheckboxInput`, `CheckboxSelectMultiple`) without forcing apps to register every subclass. Stays import-light — no `django.forms` symbol import; only walks the widget object's own class chain.

3 new tests in `TestAsLiveFieldWidgetAwareDomEvent`:
- `test_select_multiple_uses_dj_change_via_mro` (default widget for `MultipleChoiceField`)
- `test_null_boolean_select_uses_dj_change_via_mro` (default widget for `NullBooleanField`)
- `test_app_subclass_of_radio_select_inherits_commit_semantics` (custom `MyRadio` subclassing `forms.RadioSelect`)

All three fail RED on the leaf-name-only branch and pass GREEN with the MRO walk. `tests/unit/test_wizard_mixin.py`: 54 passed (was 51).

CHANGELOG updated to mention the MRO walk + the additional builtin subclasses now covered. Will merge once CI finishes.

Issue #1147, opened 2026-04-27
tech-debt: CSP-nonce-aware activator script for `<dj-lazy-slot>` fills


johnrtipton:
Source: Retro v0.9.0 / PR #1138 Stage 11 deferral / Action Tracker #173

## Context

Deferred from PR-B (#1138) Stage 11 review per ADR-015 §'Deferred from PR-B'. The current `50-lazy-fill.js` activator script does inline DOM activation when a `<dj-lazy-slot>` fill arrives. Sites with a strict CSP (no `unsafe-inline`) need a nonce-aware activator that the framework auto-emits with the matching CSP nonce attribute.

## Proposed fix

Read the request's CSP nonce from the response context (Django middleware or request attribute), thread it through the `{% live_render lazy=True %}` tag's emission of `<template id="djl-fill-X">` and any inline activation, and emit `nonce="..."` matching the document policy.

## Scope

`python/djust/templatetags/live_tags.py` + `python/djust/static/djust/src/50-lazy-fill.js` + tests; ~50 LoC + 4-6 tests. Documentation update in `docs/website/guides/streaming.md`.

Issue #1146, opened 2026-04-27
tech-debt: A075 system check — sticky+lazy template scan


johnrtipton:
Source: Retro v0.9.0 / PR #1138 Stage 11 deferral / Action Tracker #172

## Context

Deferred from PR-B (#1138) Stage 11 review per ADR-015 §'Deferred from PR-B'. The current sticky+lazy collision check happens at template-render time; a system check would catch misconfigurations at startup.

A075 would scan templates for `{% live_render %}` tags that combine `sticky=True` with `lazy=True` (which is a defined error — sticky requires synchronous render to participate in survivor preservation, lazy by definition defers render).

## Proposed implementation

Add `A075` to `python/djust/checks.py` under category 'A' (architecture). Walk the template loader's known templates and grep for the collision pattern. Emit a warning with the template path and line number.

## Scope

`python/djust/checks.py` addition; ~80 LoC + tests. Single-PR scope.

PR #1161, opened 2026-04-27
fix(rust): support project-defined @register.filter in Rust render path (closes #1121)

### LINKED ISSUE #1121: bug: Rust template renderer rejects project-defined custom filters (register.filter)

johnrtipton:
## Summary

Closes #1121.

Django projects registering custom filters via `@register.filter` in their `templatetags/` modules saw them work in the Python render path but fail under the Rust `RustLiveView` render path with `RuntimeError: Template error: Unknown filter: <name>`. The Rust engine's filter dispatch was a hardcoded match against Django's 57 built-in filter names with no fallback for project-level filters.

This PR adds a Python→Rust filter bridge that mirrors the existing custom-tag-handler design (`crates/djust_templates/src/registry.rs`), so the lookup shape from the issue body — `{{ my_dict|lookup:some_key }}` — now works exactly the same in the Rust path as in Django's Python renderer.

## Design choice — eager registration

**Eager** registration was chosen over the alternative lazy callback-on-miss design:

- **Matches existing patterns.** Custom *tag* handlers (#604) already use eager registration via `register_tag_handler` + a `Mutex<HashMap<String, Py<PyAny>>>`. Reusing the pattern keeps the codebase coherent and the bridge implementation easy to audit.
- **Zero per-render GIL probing.** With lazy, every unknown filter name would acquire the GIL on first hit just to ask Python "do you have this filter?". With eager, only filters actually invoked at render time touch Python, and the registered HashMap lookup is GIL-free.
- **Idempotent re-bootstrap.** Re-registering an existing name overwrites, so late-loaded apps' filters surface on subsequent `_initialize_rust_view` calls without coordination.

## Behaviour matrix (filter type × is_safe × autoescape)

| filter shape         | `is_safe` | `needs_autoescape` | renderer auto-escape | passes `autoescape` kwarg |
|----------------------|-----------|--------------------|----------------------|---------------------------|
| plain text           |  false    |       false        | yes                  | no                        |
| safe HTML            |  true     |       false        | no                   | no                        |
| autoescape-aware     |  any      |       true         | per `is_safe`        | yes                       |
| 0-arg, 1-arg literal |  any      |       any          | per `is_safe`        | per `needs_autoescape`    |
| 1-arg context-var    |  any      |       any          | per `is_safe`        | per `needs_autoescape`    |

Quoted args (`{{ x|prefix:"hello" }}`) are passed as literal strings. Bare-identifier args (`{{ x|prefix:greeting }}`) are resolved against the template context first — fixing the lookup case from the issue body.

## Implementation

- **`crates/djust_templates/src/filter_registry.rs`** *(new)* — process-wide `Mutex<HashMap<String, FilterEntry>>` with per-filter metadata. Dispatcher converts the `Value` input via `IntoPyObject`, resolves bare-identifier args against the context, builds kwargs for `needs_autoescape`, and converts the Python return value back to `Value`.
- **`crates/djust_templates/src/filters.rs`** — new `apply_filter_full` internal entry point accepting an `arg_was_quoted` flag. The unknown-filter fallback consults `filter_registry::apply_custom_filter`; on registry miss it still emits the original `Unknown filter: <name>` error so typos and missing imports surface immediately. Public `apply_filter_with_context` keeps its existing signature for back-compat.
- **`crates/djust_templates/src/renderer.rs`** — three filter-application sites updated to forward quoting metadata. The `safe_output_filters` whitelist is augmented to also consult `filter_registry::is_custom_filter_safe` so `is_safe=True` custom filters skip auto-escape just like built-ins.
- **`crates/djust_live/src/lib.rs`** — exposes 5 new pyfunctions: `register_custom_filter`, `unregister_custom_filter`, `has_custom_filter`, `clear_custom_filters`, `get_registered_custom_filters`.
- **`python/djust/template_filters.py`** *(new)* — `register_django_filter` (single forward) + `bootstrap_django_filters` (walk every Django `Library` from `template.engines['django'].engine.template_libraries`). Built-in Django filter names are skipped to keep the registry compact.
- **`python/djust/mixins/rust_bridge.py`** — `_initialize_rust_view` calls a process-level `_ensure_custom_filters_bridged` once per process; existing LiveView code paths pick up project filters automatically.
- **Docs** — `docs/website/core-concepts/templates.md` gets a new "Custom filters (`@register.filter`)" section with the lookup example from the issue.

## Tests added

`tests/unit/test_rust_custom_filters_1121.py` — 10 regression cases:

| # | Case |
|---|---|
| 1 | `{{ my_dict\|lookup:some_key }}` — exact issue-body shape |
| 2 | `lookup` with missing key returns empty |
| 3 | Zero-argument filter |
| 4 | Single-argument filter with quoted literal |
| 5 | Single-argument filter with bare-identifier (context-resolved) |
| 6 | Plain-text filter result is auto-escaped |
| 7 | `is_safe=True` filter result is NOT re-escaped |
| 8 | `needs_autoescape=True` filter receives `autoescape` kwarg |
| 9 | Unknown filter still raises clear `Unknown filter: <name>` error |
| 10 | Full `RustLiveView` render path — closes the loop on the original report |

All 3977 existing Python tests still pass; cargo test for `djust_templates` passes (66 tests).

## Test plan

- [x] Failing test written first (`Unknown filter: lookup`) — confirmed RED before fix
- [x] All 10 regression cases pass
- [x] Full Python suite (3977 tests) passes
- [x] `cargo test -p djust_templates` (66 tests) passes
- [x] `cargo fmt --check` clean
- [x] `cargo clippy --workspace --exclude djust_live -- -D warnings` clean
- [x] `ruff check` + `ruff format` clean on new Python files
- [x] Pre-push hooks pass (full suite + cargo test + cargo audit)
- [ ] CI matrix passes
- [ ] Stage 11 reviewer signs off

## Notes for reviewers

- **🟡 Argument resolution policy.** When a bare-identifier arg has no context binding (e.g. `{{ x|prefix:typo }}` where `typo` isn't defined), the bridge passes the raw identifier string to Python — matching Django's tolerant behaviour where missing template vars default to `""`/literal text rather than raising. If you'd prefer a strict mode, that's a follow-up.
- **🟡 First-render bootstrap latency.** `_ensure_custom_filters_bridged` walks every Django `Library.filters` dict on the first `_initialize_rust_view` call per process. Cost is one HashMap insert per project filter (typical: <50 filters → sub-millisecond). The guard short-circuits all subsequent calls.
- **🟡 No reverse lookup yet.** The bridge only forwards filters at registration time. If a project hot-reloads a `templatetags/` module after first render, the new filter won't appear until a process restart (or manual `bootstrap_django_filters()` call). Same constraint as the existing custom-tag-handler bridge, so not a regression.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,43 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- **Rust template renderer rejects project-defined custom filters
+  (closes #1121)** — Django projects registering custom filters via
+  ``@register.filter`` in their ``templatetags/`` modules saw them work
+  in the Python render path but fail under the Rust ``RustLiveView``
+  render path with ``RuntimeError: Template error: Unknown filter:
+  <name>``. The Rust engine's filter dispatch was a hardcoded match
+  against Django's 57 built-in filter names with no fallback for
+  project-level filters. The fix is a Python→Rust bridge mirroring
+  the existing custom-tag-handler design (``crates/djust_templates/
+  src/registry.rs``):
+  - New ``crates/djust_templates/src/filter_registry.rs`` holds a
+    process-wide ``Mutex<HashMap<String, FilterEntry>>`` of project
+    filter callables + per-filter metadata (``is_safe``,
+    ``needs_autoescape``).
+  - The renderer's filter loop forwards an ``arg_was_quoted`` hint
+    from the parser so the bridge can resolve bare-identifier args
+    against the template context before calling Python — fixing the
+    ``{{ my_dict|lookup:some_key }}`` shape from the issue body.
+  - Both ``filter.is_safe`` and ``filter.needs_autoescape`` from the
+    Django filter object are honoured: ``is_safe=True`` filters skip
+    auto-escape; ``needs_autoescape=True`` filters receive
+    ``autoescape=True`` as a kwarg.
+  - ``python/djust/template_filters.py`` walks
+    ``template.engines['django'].engine.template_libraries`` at the
+    first LiveView render and bulk-registers every custom filter
+    found. Built-in Django filter names are skipped (the Rust engine
+    has native implementations of all 57). The bootstrap is
+    idempotent — late-loaded apps' filters are picked up on
+    subsequent renders.
+  - Unknown filter names still raise the original
+    ``Unknown filter: <name>`` error so typos and missing imports
+    surface immediately. 10 regression cases in
+    ``TestRustCustomFilters`` cover the lookup shape from the issue
+    body, ``is_safe``, ``needs_autoescape``, quoted vs context-
+    resolved args, plain-text auto-escape, and the full
+    ``RustLiveView`` render path.
+
 - **Test pollution: 6 flaky tests in full-suite pytest run (closes #1134)** —
   bisected two independent polluters that surfaced after v0.9.0 PR-A
   (#1135) added the `aget`/`ChunkEmitter` async-render path and after
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -294,7 +294,7 @@ dependencies = [
 
 [[package]]
 name = "djust_components"
-version = "0.9.0-rc.1"
+version = "0.9.0-rc.2"
 dependencies = [
  "ahash",
  "criterion",
@@ -309,7 +309,7 @@ dependencies = [
 
 [[package]]
 name = "djust_core"
-version = "0.9.0-rc.1"
+version = "0.9.0-rc.2"
 dependencies = [
  "ahash",
  "criterion",
@@ -324,7 +324,7 @@ dependencies = [
 
 [[package]]
 name = "djust_live"
-version = "0.9.0-rc.1"
+version = "0.9.0-rc.2"
 dependencies = [
  "bincode",
  "criterion",
@@ -349,7 +349,7 @@ dependencies = [
 
 [[package]]
 name = "djust_templates"
-version = "0.9.0-rc.1"
+version = "0.9.0-rc.2"
 dependencies = [
  "ahash",
  "chrono",
@@ -370,7 +370,7 @@ dependencies = [
 
 [[package]]
 name = "djust_vdom"
-version = "0.9.0-rc.1"
+version = "0.9.0-rc.2"
 dependencies = [
  "ahash",
  "criterion",
--- a/crates/djust_live/src/lib.rs
+++ b/crates/djust_live/src/lib.rs
@@ -2727,5 +2727,28 @@ fn _rust(m: &Bound<'_, PyModule>) -> PyResult<()> {
         m
     )?)?;
 
+    // Custom filter registry (project-defined ``@register.filter`` callables) — #1121.
+    // Bridges Django's per-app filter libraries into the Rust template engine.
+    m.add_function(wrap_pyfunction!(
+        djust_templates::filter_registry::register_custom_filter,
+        m
+    )?)?;
+    m.add_function(wrap_pyfunction!(
+        djust_templates::filter_registry::unregister_custom_filter,
+        m
+    )?)?;
+    m.add_function(wrap_pyfunction!(
+        djust_templates::filter_registry::has_custom_filter,
+        m
+    )?)?;
+    m.add_function(wrap_pyfunction!(
+        djust_templates::filter_registry::clear_custom_filters,
+        m
+    )?)?;
+    m.add_function(wrap_pyfunction!(
+        djust_templates::filter_registry::get_registered_custom_filters,
+        m
+    )?)?;
+
     Ok(())
 }
--- a/crates/djust_templates/src/filter_registry.rs
+++ b/crates/djust_templates/src/filter_registry.rs
@@ -1,0 +1,299 @@
+//! Custom filter registry for project-defined Django ``@register.filter`` callables.
+//!
+//! Mirrors the design of [`crate::registry`] (which dispatches custom *tags*),
+//! but for filters. Built-in Django filters live as native-Rust matches in
+//! [`crate::filters`]; project-level custom filters that come from
+//! ``@register.filter`` in a Django app's ``templatetags/`` package are
+//! registered here at engine bootstrap time.
+//!
+//! # Lazy vs eager
+//!
+//! This implementation is **eager** — Python registers each filter callable
+//! exactly once via [`register_custom_filter`], typically by walking
+//! ``template.engines['django'].engine.template_libraries`` at import time.
+//! At render time, [`apply_custom_filter`] performs a HashMap lookup
+//! followed by a GIL acquire + Python call. The eager design matches the
+//! existing tag-handler pattern in [`crate::registry`] and avoids a
+//! per-render GIL acquisition for "is this a known filter name?" probes.
+//!
+//! Memory cost: one entry per project filter. ~50 bytes of `String` +
+//! `Py<PyAny>` + `FilterMeta` per registration. Even projects with hundreds
+//! of custom filters fit comfortably.
+//!
+//! # Filter signature
+//!
+//! Django filter callables accept ``(value, arg=None)`` and return a
+//! string (or a SafeString when ``is_safe=True``). ``needs_autoescape=True``
+//! filters additionally accept ``autoescape`` as a kwarg.
+//!
+//! - ``value`` — the filtered expression's current `Value`, converted to
+//!   the appropriate Python type (str/int/float/bool/None/list/dict).
+//! - ``arg`` — for one-argument filters, the resolved argument:
+//!     - quoted literals (``"foo"``) are passed as ``str``,
+//!     - bare identifiers are resolved against the template context. If
+//!       the context resolves to a primitive, it's passed as that type;
+//!       otherwise as the value's natural Python representation.
+//! - return — the result. ``is_safe=True`` filters' results bypass
+//!   auto-escape via [`is_custom_filter_safe`] consulted by the renderer.
+
+use crate::Value;
+use djust_core::Context;
+use once_cell::sync::Lazy;
+use pyo3::prelude::*;
+use pyo3::types::PyDict;
+use std::collections::HashMap;
+use std::sync::Mutex;
+
+/// Per-filter metadata mirroring Django's filter object attributes.
+#[derive(Debug, Clone, Default)]
+pub struct FilterMeta {
+    /// ``filter.is_safe`` — when true, the renderer must NOT HTML-escape
+    /// the filter's output. The Python callable is expected to return
+    /// already-escaped content (e.g. via ``mark_safe``).
+    pub is_safe: bool,
+    /// ``filter.needs_autoescape`` — when true, the dispatcher passes
+    /// ``autoescape=True`` as a kwarg so the filter can branch on the
+    /// surrounding autoescape policy.
+    pub needs_autoescape: bool,
+}
+
+struct FilterEntry {
+    callable: Py<PyAny>,
+    meta: FilterMeta,
+}
+
+/// Global registry mapping filter names to Python callables + metadata.
+static FILTER_REGISTRY: Lazy<Mutex<HashMap<String, FilterEntry>>> =
+    Lazy::new(|| Mutex::new(HashMap::new()));
+
+/// Register a project-defined custom filter from Python.
+///
+/// # Arguments
+///
+/// * ``name`` — filter name as used in templates (``{{ x|name }}``).
+/// * ``callable`` — Django filter callable (``(value, arg=None) -> str``).
+/// * ``is_safe`` — Django filter's ``is_safe`` attribute (skip auto-escape).
+/// * ``needs_autoescape`` — Django filter's ``needs_autoescape`` attribute
+///   (pass ``autoescape=True`` as kwarg).
+///
+/// Re-registering an existing name overwrites — matching Django's behaviour
+/// when a Library is re-imported.
+#[pyfunction]
+#[pyo3(signature = (name, callable, is_safe=false, needs_autoescape=false))]
+pub fn register_custom_filter(
+    name: String,
+    callable: Py<PyAny>,
+    is_safe: bool,
+    needs_autoescape: bool,
+) -> PyResult<()> {
+    let mut registry = FILTER_REGISTRY.lock().map_err(|e| {
+        PyErr::new::<pyo3::exceptions::PyRuntimeError, _>(format!("Filter registry lock: {e}"))
+    })?;
+    registry.insert(
+        name,
+        FilterEntry {
+            callable,
+            meta: FilterMeta {
+                is_safe,
+                needs_autoescape,
+            },
+        },
+    );
+    Ok(())
+}
+
+/// Unregister a custom filter (returns ``true`` if a filter was removed).
+#[pyfunction]
+pub fn unregister_custom_filter(name: &str) -> PyResult<bool> {
+    let mut registry = FILTER_REGISTRY.lock().map_err(|e| {
+        PyErr::new::<pyo3::exceptions::PyRuntimeError, _>(format!("Filter registry lock: {e}"))
+    })?;
+    Ok(registry.remove(name).is_some())
+}
+
+/// Check if a custom filter is registered (intended for tests + diagnostics).
+#[pyfunction]
+pub fn has_custom_filter(name: &str) -> PyResult<bool> {
+    let registry = FILTER_REGISTRY.lock().map_err(|e| {
+        PyErr::new::<pyo3::exceptions::PyRuntimeError, _>(format!("Filter registry lock: {e}"))
+    })?;
+    Ok(registry.contains_key(name))
+}
+
+/// Clear all registered custom filters (primarily for tests).
+#[pyfunction]
+pub fn clear_custom_filters() -> PyResult<()> {
+    let mut registry = FILTER_REGISTRY.lock().map_err(|e| {
+        PyErr::new::<pyo3::exceptions::PyRuntimeError, _>(format!("Filter registry lock: {e}"))
+    })?;
+    registry.clear();
+    Ok(())
+}
+
+/// List all registered custom filter names (for diagnostics).
+#[pyfunction]
+pub fn get_registered_custom_filters() -> PyResult<Vec<String>> {
+    let registry = FILTER_REGISTRY.lock().map_err(|e| {
+        PyErr::new::<pyo3::exceptions::PyRuntimeError, _>(format!("Filter registry lock: {e}"))
+    })?;
+    Ok(registry.keys().cloned().collect())
+}
+
+// ============================================================================
+// Internal Rust API (called from filters.rs / renderer.rs)
+// ============================================================================
+
+/// Returns ``true`` if a registered custom filter has ``is_safe=True``.
+///
+/// The renderer consults this alongside the hardcoded built-in
+/// ``safe_output_filters`` list to decide whether to skip auto-escape.
+pub fn is_custom_filter_safe(name: &str) -> bool {
+    FILTER_REGISTRY
+        .lock()
+        .map(|reg| reg.get(name).map(|e| e.meta.is_safe).unwrap_or(false))
+        .unwrap_or(false)
+}
+
+/// Returns ``true`` if any custom filter is registered for the given name.
+pub fn custom_filter_exists(name: &str) -> bool {
+    FILTER_REGISTRY
+        .lock()
+        .map(|reg| reg.contains_key(name))
+        .unwrap_or(false)
+}
+
+/// Apply a custom filter callable to a value with an optional argument.
+///
+/// Called from [`crate::filters::apply_filter_with_context`] when the
+/// built-in filter match falls through. Returns ``None`` if no custom
+/// filter is registered for the name (so the caller can fall through to
+/// the standard ``Unknown filter`` error).
+///
+/// Argument resolution: when ``arg`` is provided as a non-empty string
+/// after ``strip_filter_arg_quotes``, this function inspects the original
+/// arg string for surrounding quotes:
+/// - quoted (``"foo"`` or ``'foo'``) — passed to Python as a literal string
+///   (with quotes already stripped by the caller).
+/// - bare identifier — resolved against ``context`` first; if a binding
+///   exists, the resolved `Value` is passed. Otherwise the bare identifier
+///   string itself is passed (mirroring Django's tolerant behaviour where
+///   filters accept literal arg text when no binding matches).
+///
+/// This split is the same convention `crate::filters::apply_filter_with_context`
+/// already uses for built-ins like ``date`` (literal format string) vs
+/// callers passing context-resolved values.
+pub fn apply_custom_filter(
+    name: &str,
+    value: &Value,
+    arg: Option<&str>,
+    context: Option<&Context>,
+    arg_was_quoted: bool,
+) -> Option<Result<Value, String>> {
+    let (callable, meta) = {
+        let registry = FILTER_REGISTRY.lock().ok()?;
+        let entry = registry.get(name)?;
+        // clone_ref under the GIL; meta is plain Copy-ish.
+        let callable = Python::with_gil(|py| entry.callable.clone_ref(py));
+        (callable, entry.meta.clone())
+    };
+
+    let result = Python::with_gil(|py| -> Result<Value, String> {
+        use pyo3::IntoPyObject;
+
+        let py_value = value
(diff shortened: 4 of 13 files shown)


johnrtipton (commented):
## Stage 11 Code Review — PR #1161

**Verdict**: APPROVE (with non-blocking 🟡 follow-ups)

The fix is correct in shape, well-scoped, and faithfully mirrors the existing custom-tag-handler pattern (`crates/djust_templates/src/registry.rs`). The eager-bridge design choice is justified, the GIL/Mutex layering is identical to the pre-existing tag-handler flow that's already shipped, and the 10 regression tests cover meaningful distinct paths. CHANGELOG matches the diff. No collateral noise. PR body matches actual scope.

---

### 🟢 Strengths

1. **Lock-then-clone-then-release pattern is correct** (`filter_registry.rs:191-198`). The `FILTER_REGISTRY.lock()` is dropped at the end of the inner block (line 198 of the diff: `(callable, entry.meta.clone())`) before the Python call begins. No GIL-while-holding-mutex deadlock surface — this matches the existing tag handler convention in `registry.rs:260-313`.
2. **Failure containment is right** (`mixins/rust_bridge.py:73-83`). The `_ensure_custom_filters_bridged` `try/except/finally` pattern guarantees the guard flag is set even on bootstrap failure, so a single import error doesn't repeatedly log on every render. The Python-render fallback path remains intact.
3. **Logging hygiene clean.** All five new `logger.*` calls in `template_filters.py` and `rust_bridge.py` use `%s`-style formatting. No tainted-format-string regression.
4. **`is_quoted_arg` is dead-simple and correct** for the standard cases (line 530-533 in renderer.rs diff). The `len() >= 2` guard correctly rejects single-quote-char strings.
5. **CHANGELOG entry is comprehensive** and uses the canonical `N regression cases in TestRustCustomFilters` form (CLASS name, not file name) — respects the PR #1106 / #1112 retro rule.
6. **Skipping built-in filter names** in `_BUILTIN_NAMES` is the right call — keeps the registry compact and avoids subtle "Python wrapper around Rust built-in" performance traps.

---

### 🟡 Should-Fix (consider before merge — none block)

**1. Hot-path Mutex lock on every variable render even when no custom filters are registered** (`renderer.rs:314, 369`).

`is_custom_filter_safe(name)` is invoked inside `filter_specs.iter().any(|(name, _)| ...)` for every filter chain on every variable in every node, on every render. That's a `FILTER_REGISTRY.lock()` acquisition per filter per render — even on apps that register zero custom filters. Pre-PR the cost was a 7-element static array `.contains()` (~zero). Post-PR there's a Mutex acquisition added behind the OR.

Mitigation suggestion (cheap): add an `AtomicBool` `REGISTRY_NONEMPTY` next to the static, set true on first `register_custom_filter` insert, and short-circuit:
```rust
pub fn is_custom_filter_safe(name: &str) -> bool {
    if !REGISTRY_NONEMPTY.load(Ordering::Relaxed) { return false; }
    FILTER_REGISTRY.lock()
        .map(|reg| reg.get(name).map(|e| e.meta.is_safe).unwrap_or(false))
        .unwrap_or(false)
}
```
This makes the steady-state fast path a single relaxed atomic load for the 99.9% of apps with zero custom filters. **Filing as follow-up issue is acceptable** if the intent is to keep this PR tight to the bug fix.

**2. `autoescape` kwarg is hardcoded `true` regardless of surrounding `{% autoescape %}` block.**

`filter_registry.rs:255` passes `kw.set_item("autoescape", true)`. The Rust renderer doesn't track `{% autoescape off %}` block state at all (verified by `grep -rn autoescape crates/djust_templates/src/` — only this PR's code matches). For filters using `needs_autoescape=True` to genuinely branch on autoescape policy, the Rust path will silently always behave as if autoescape is on. This is **not a regression** introduced by this PR (the engine never supported `{% autoescape %}` blocks), but the docs at `templates.md:174` say `needs_autoescape=True` filters work "same as Django" — that overpromises. Recommend: add a one-line caveat in the docs ("Always receives `autoescape=True` until #XXXX adds `{% autoescape %}` block tracking") and file a follow-up.

**3. Negative-case test is too tolerant** (`test_rust_custom_filters_1121.py:188-191`).

The assertion `"definitely_not_a_filter_xyz" in str(exc_info.value)` would also pass if the error message changes to `"Custom filter '<name>' raised exception"` — i.e. it doesn't lock down "Unknown filter:" specifically. Recommend strengthening to:
```python
assert "Unknown filter" in str(exc_info.value)
assert "definitely_not_a_filter_xyz" in str(exc_info.value)
```
This protects against future regressions where misses are accidentally routed to a different error path.

**4. `pub fn custom_filter_exists` is unused** (`filter_registry.rs:294`). The implementer flagged this. Either drop it (YAGNI) or add `#[allow(dead_code)]` with a comment pointing to the future parser-time-check use case. Currently it'll raise a `dead_code` warning if anyone runs `cargo clippy --workspace` without the existing `--exclude djust_live` shield.

**5. Module-scoped fixture teardown clears the GLOBAL filter registry** (`test_rust_custom_filters_1121.py:78-84`).

`clear_custom_filters()` in the teardown wipes the process-wide registry. If another test module registers custom filters earlier in the run AND assumes they persist, this teardown will break them. The current test layout makes this hypothetical (only one such module), but as the suite grows this becomes a test-pollution vector. Recommend: capture the registry before each test, restore in teardown — OR document that any future custom-filter test module must use the same fixture pattern.

**6. Async-filter detection is silent.** A user passing an `async def` filter through the bridge will get the cryptic `"returned a non-convertible value"` error rather than `"Async filters not supported (#XXXX)"`. Quick win: in `apply_custom_filter`, after the call, check `inspect.iscoroutine(py_result)` and return a clearer error. **Filing as follow-up is acceptable.**

---

### 🔴 Must-Fix

None. The implementation is correctness-complete for the stated scope (#1121).

---

### Self-Flagged 🟡s — Reviewer Decisions

1. **Bare-identifier with no context binding falls back to raw identifier string** — Acceptable. Matches Django's tolerant default. Diff line `filter_registry.rs:368-374` is well-commented.
2. **First-render bootstrap walks every Django Library once per process** — Acceptable. Sub-ms cost, idempotent guard. The `_CUSTOM_FILTERS_BRIDGED` module-level flag is process-level, **but is not thread-safe against concurrent first-renders.** Two threads can both pass the `if _CUSTOM_FILTERS_BRIDGED:` check, both call `bootstrap_django_filters`, both walk the libraries, and both register every filter twice. Re-registration is idempotent in the Rust registry (`HashMap::insert` overwrites), so this isn't a correctness bug — just wasted work on cold start. If you want to be tidy, wrap the guard in a `threading.Lock()`. Acceptable as-is.
3. **No hot-reload path** — Acceptable. Same constraint as the existing tag-handler bridge. Not a regression.
4. **`custom_filter_exists` unused** — See 🟡 #4 above. Either drop or annotate.

---

### Edge Cases Considered

- Empty arg string (`""`) — `is_quoted_arg` returns false, treated as bare identifier of empty name → context lookup fails → empty string passed to filter. Reasonable.
- Mismatched quotes (`"'mixed"`) — treated as bare identifier. Filter receives literal `'mixed"` string. Django-tolerant; OK.
- Filter that takes 0 args invoked with 1 arg in template — Python raises `TypeError`, wrapped as `Custom filter 'X' raised exception: ...` with traceback. Clear enough.
- Filter that raises during call — caught by `format_py_err`, traceback included. Good DX.
- `@register.filter` on async function — silently routes to cryptic `non-convertible value` error (see 🟡 #6).
- Re-registering same name with different `is_safe` — last-write-wins, by design. Matches Django's Library re-import behavior. OK.
- Two threads racing `_ensure_custom_filters_bridged` on cold start — duplicates work but not incorrect (see decision 2 above).
- `safe_output_filters` whitelist consulted before bridge bootstrap completes — at first render, `_initialize_rust_view` calls `_ensure_custom_filters_bridged()` BEFORE any rendering happens. So the registry is populated by the time the safe-filter check runs. No startup race. (Verified by reading `mixins/rust_bridge.py:175-181` ordering.)

---

### Open Questions

1. **Why was `apply_filter_with_context`'s back-compat shim defaulted to `arg_was_quoted=true`?** The doc-comment at `filters.rs:25` says "assume callers pre-resolved any context-variable args" — but a future caller passing a custom filter name with a bare-identifier arg through this entry point would silently get literal-string-passing behavior instead of context resolution. This is fine as long as the public API contract is "use `apply_filter_full` for new code" and `apply_filter_with_context` is internal-only after this PR. Worth a deprecation note in a follow-up if so.

2. **Should `_ensure_custom_filters_bridged` accept an explicit re-bootstrap flag?** Hot-reload during dev (Django's `runserver` autoreload) creates new processes anyway, so the per-process guard is fine for prod. But for long-lived ASGI workers in dev where someone edits a `templatetags/foo.py` without restarting (unlikely but possible with watcher tools), the guard sticks. Same constraint as the tag bridge — likely fine, but worth one line in the docstring explaining how to manually re-bootstrap (`bootstrap_django_filters()` is exposed and idempotent).

---

### Verdict

**APPROVE** — non-blocking 🟡 findings; the bug from #1121 is correctly fixed and the implementation pattern is consistent with prior art. Recommend opening follow-up issues for 🟡 #1 (Mutex hot-path), 🟡 #2 (autoescape block tracking), 🟡 #4 (`custom_filter_exists` cleanup), and 🟡 #6 (async-filter clearer error). 🟡 #3 (negative-case test tightening) and 🟡 #5 (test fixture isolation) are best fixed in this PR if you have time before merge — both are one-liners.

johnrtipton:
## Stage 11 Code Review — PR #1161

**Verdict**: APPROVE (with non-blocking 🟡 follow-ups)

The fix is correct in shape, well-scoped, and faithfully mirrors the existing custom-tag-handler pattern (`crates/djust_templates/src/registry.rs`). The eager-bridge design choice is justified, the GIL/Mutex layering is identical to the pre-existing tag-handler flow that's already shipped, and the 10 regression tests cover meaningful distinct paths. CHANGELOG matches the diff. No collateral noise. PR body matches actual scope.

---

### 🟢 Strengths

1. **Lock-then-clone-then-release pattern is correct** (`filter_registry.rs:191-198`). The `FILTER_REGISTRY.lock()` is dropped at the end of the inner block (line 198 of the diff: `(callable, entry.meta.clone())`) before the Python call begins. No GIL-while-holding-mutex deadlock surface — this matches the existing tag handler convention in `registry.rs:260-313`.
2. **Failure containment is right** (`mixins/rust_bridge.py:73-83`). The `_ensure_custom_filters_bridged` `try/except/finally` pattern guarantees the guard flag is set even on bootstrap failure, so a single import error doesn't repeatedly log on every render. The Python-render fallback path remains intact.
3. **Logging hygiene clean.** All five new `logger.*` calls in `template_filters.py` and `rust_bridge.py` use `%s`-style formatting. No tainted-format-string regression.
4. **`is_quoted_arg` is dead-simple and correct** for the standard cases (line 530-533 in renderer.rs diff). The `len() >= 2` guard correctly rejects single-quote-char strings.
5. **CHANGELOG entry is comprehensive** and uses the canonical `N regression cases in TestRustCustomFilters` form (CLASS name, not file name) — respects the PR #1106 / #1112 retro rule.
6. **Skipping built-in filter names** in `_BUILTIN_NAMES` is the right call — keeps the registry compact and avoids subtle "Python wrapper around Rust built-in" performance traps.

---

### 🟡 Should-Fix (consider before merge — none block)

**1. Hot-path Mutex lock on every variable render even when no custom filters are registered** (`renderer.rs:314, 369`).

`is_custom_filter_safe(name)` is invoked inside `filter_specs.iter().any(|(name, _)| ...)` for every filter chain on every variable in every node, on every render. That's a `FILTER_REGISTRY.lock()` acquisition per filter per render — even on apps that register zero custom filters. Pre-PR the cost was a 7-element static array `.contains()` (~zero). Post-PR there's a Mutex acquisition added behind the OR.

Mitigation suggestion (cheap): add an `AtomicBool` `REGISTRY_NONEMPTY` next to the static, set true on first `register_custom_filter` insert, and short-circuit:
```rust
pub fn is_custom_filter_safe(name: &str) -> bool {
    if !REGISTRY_NONEMPTY.load(Ordering::Relaxed) { return false; }
    FILTER_REGISTRY.lock()
        .map(|reg| reg.get(name).map(|e| e.meta.is_safe).unwrap_or(false))
        .unwrap_or(false)
}
```
This makes the steady-state fast path a single relaxed atomic load for the 99.9% of apps with zero custom filters. **Filing as follow-up issue is acceptable** if the intent is to keep this PR tight to the bug fix.

**2. `autoescape` kwarg is hardcoded `true` regardless of surrounding `{% autoescape %}` block.**

`filter_registry.rs:255` passes `kw.set_item("autoescape", true)`. The Rust renderer doesn't track `{% autoescape off %}` block state at all (verified by `grep -rn autoescape crates/djust_templates/src/` — only this PR's code matches). For filters using `needs_autoescape=True` to genuinely branch on autoescape policy, the Rust path will silently always behave as if autoescape is on. This is **not a regression** introduced by this PR (the engine never supported `{% autoescape %}` blocks), but the docs at `templates.md:174` say `needs_autoescape=True` filters work "same as Django" — that overpromises. Recommend: add a one-line caveat in the docs ("Always receives `autoescape=True` until #XXXX adds `{% autoescape %}` block tracking") and file a follow-up.

**3. Negative-case test is too tolerant** (`test_rust_custom_filters_1121.py:188-191`).

The assertion `"definitely_not_a_filter_xyz" in str(exc_info.value)` would also pass if the error message changes to `"Custom filter '<name>' raised exception"` — i.e. it doesn't lock down "Unknown filter:" specifically. Recommend strengthening to:
```python
assert "Unknown filter" in str(exc_info.value)
assert "definitely_not_a_filter_xyz" in str(exc_info.value)
```
This protects against future regressions where misses are accidentally routed to a different error path.

**4. `pub fn custom_filter_exists` is unused** (`filter_registry.rs:294`). The implementer flagged this. Either drop it (YAGNI) or add `#[allow(dead_code)]` with a comment pointing to the future parser-time-check use case. Currently it'll raise a `dead_code` warning if anyone runs `cargo clippy --workspace` without the existing `--exclude djust_live` shield.

**5. Module-scoped fixture teardown clears the GLOBAL filter registry** (`test_rust_custom_filters_1121.py:78-84`).

`clear_custom_filters()` in the teardown wipes the process-wide registry. If another test module registers custom filters earlier in the run AND assumes they persist, this teardown will break them. The current test layout makes this hypothetical (only one such module), but as the suite grows this becomes a test-pollution vector. Recommend: capture the registry before each test, restore in teardown — OR document that any future custom-filter test module must use the same fixture pattern.

**6. Async-filter detection is silent.** A user passing an `async def` filter through the bridge will get the cryptic `"returned a non-convertible value"` error rather than `"Async filters not supported (#XXXX)"`. Quick win: in `apply_custom_filter`, after the call, check `inspect.iscoroutine(py_result)` and return a clearer error. **Filing as follow-up is acceptable.**

---

### 🔴 Must-Fix

None. The implementation is correctness-complete for the stated scope (#1121).

---

### Self-Flagged 🟡s — Reviewer Decisions

1. **Bare-identifier with no context binding falls back to raw identifier string** — Acceptable. Matches Django's tolerant default. Diff line `filter_registry.rs:368-374` is well-commented.
2. **First-render bootstrap walks every Django Library once per process** — Acceptable. Sub-ms cost, idempotent guard. The `_CUSTOM_FILTERS_BRIDGED` module-level flag is process-level, **but is not thread-safe against concurrent first-renders.** Two threads can both pass the `if _CUSTOM_FILTERS_BRIDGED:` check, both call `bootstrap_django_filters`, both walk the libraries, and both register every filter twice. Re-registration is idempotent in the Rust registry (`HashMap::insert` overwrites), so this isn't a correctness bug — just wasted work on cold start. If you want to be tidy, wrap the guard in a `threading.Lock()`. Acceptable as-is.
3. **No hot-reload path** — Acceptable. Same constraint as the existing tag-handler bridge. Not a regression.
4. **`custom_filter_exists` unused** — See 🟡 #4 above. Either drop or annotate.

---

### Edge Cases Considered

- Empty arg string (`""`) — `is_quoted_arg` returns false, treated as bare identifier of empty name → context lookup fails → empty string passed to filter. Reasonable.
- Mismatched quotes (`"'mixed"`) — treated as bare identifier. Filter receives literal `'mixed"` string. Django-tolerant; OK.
- Filter that takes 0 args invoked with 1 arg in template — Python raises `TypeError`, wrapped as `Custom filter 'X' raised exception: ...` with traceback. Clear enough.
- Filter that raises during call — caught by `format_py_err`, traceback included. Good DX.
- `@register.filter` on async function — silently routes to cryptic `non-convertible value` error (see 🟡 #6).
- Re-registering same name with different `is_safe` — last-write-wins, by design. Matches Django's Library re-import behavior. OK.
- Two threads racing `_ensure_custom_filters_bridged` on cold start — duplicates work but not incorrect (see decision 2 above).
- `safe_output_filters` whitelist consulted before bridge bootstrap completes — at first render, `_initialize_rust_view` calls `_ensure_custom_filters_bridged()` BEFORE any rendering happens. So the registry is populated by the time the safe-filter check runs. No startup race. (Verified by reading `mixins/rust_bridge.py:175-181` ordering.)

---

### Open Questions

1. **Why was `apply_filter_with_context`'s back-compat shim defaulted to `arg_was_quoted=true`?** The doc-comment at `filters.rs:25` says "assume callers pre-resolved any context-variable args" — but a future caller passing a custom filter name with a bare-identifier arg through this entry point would silently get literal-string-passing behavior instead of context resolution. This is fine as long as the public API contract is "use `apply_filter_full` for new code" and `apply_filter_with_context` is internal-only after this PR. Worth a deprecation note in a follow-up if so.

2. **Should `_ensure_custom_filters_bridged` accept an explicit re-bootstrap flag?** Hot-reload during dev (Django's `runserver` autoreload) creates new processes anyway, so the per-process guard is fine for prod. But for long-lived ASGI workers in dev where someone edits a `templatetags/foo.py` without restarting (unlikely but possible with watcher tools), the guard sticks. Same constraint as the tag bridge — likely fine, but worth one line in the docstring explaining how to manually re-bootstrap (`bootstrap_django_filters()` is exposed and idempotent).

---

### Verdict

**APPROVE** — non-blocking 🟡 findings; the bug from #1121 is correctly fixed and the implementation pattern is consistent with prior art. Recommend opening follow-up issues for 🟡 #1 (Mutex hot-path), 🟡 #2 (autoescape block tracking), 🟡 #4 (`custom_filter_exists` cleanup), and 🟡 #6 (async-filter clearer error). 🟡 #3 (negative-case test tightening) and 🟡 #5 (test fixture isolation) are best fixed in this PR if you have time before merge — both are one-liners.

johnrtipton:
# Pipeline Retrospective — PR #1161 (closes #1121)

**Branch**: `fix/1121-rust-custom-filters`
**Pipeline**: `bugfix` (14 stages)
**Date**: 2026-04-27
**Milestone**: v0.9.1 (drain iter 2 of 7)

## What Worked Well

1. **Eager-registration design pattern reused.** The implementer mirrored the existing custom-tag-handler bridge (`crates/djust_templates/src/registry.rs`) for the new filter bridge. Same `Py<PyAny>` storage shape, same dispatch, GIL-free hot-path lookup. Stage-4 first-principles rule (canonicalize from #1032 retro) — grep-before-architect found the precedent and saved a redesign cycle.
2. **`apply_filter_full(arg_was_quoted)` API extension over a strip-quotes shim.** The old `apply_filter_with_context` discarded quoting info before dispatch — fatal for custom filters that need to know whether to context-resolve. New API preserves quoted-vs-bare; back-compat shim defaults `arg_was_quoted=true` so existing Rust callers (cargo tests) keep working. Clean migration.
3. **Stage 11 reviewer caught 6 🟡 + flagged the most material one explicitly** (hot-path Mutex on every render's filter chain). The "consolidate 6 🟡s into one follow-up issue" pattern keeps the merge moving while preserving the polish work. Filed as #1162.

## What Didn't Work

1. **Pre-commit hook bounced once on the `.pyi` stub.** Ruff reformatted the type stub during the hook; commit didn't register; agent had to re-stage and re-commit. Per CLAUDE.md retro #122 the agent caught it via `git log -1 --oneline`, so no time lost — but it's the 9th observed occurrence of the stash-then-reformat-then-bounce pattern this week.
2. **`unused: pub fn custom_filter_exists`** — exposed for future use but no consumer. PyO3 macros suppress dead-code warnings, so it'll quietly rot. Filed as item #4 in #1162.

## What to Improve

1. **Stage 4 design-options table for new bridge code.** The implementer chose eager registration with one paragraph of rationale. Stage 11 reviewer asked "why eager vs lazy" as an open question — would have been faster to answer if the Plan stage had a 2-row table (eager vs lazy, pros/cons each) up front. Worth canonicalizing as a Stage 4 plan template addition for any new Python↔Rust bridge work.

- **CLAUDE.md / Stage 4 plan template**: any new Python→Rust bridge feature must include a "design options considered" table (eager vs lazy, push vs callback, sync vs async) with explicit pro/con for each before committing to an implementation. Pre-empts the most common Stage 11 "why this choice?" question.

## Review Stats

| Metric | PR #1161 |
|--------|----------|
| Tests added | 10 (all in `tests/unit/test_rust_custom_filters_1121.py`) |
| 🔴 Findings | 0 |
| 🟡 Findings | 6 (all in #1162 follow-up) |
| Findings fixed pre-merge | 0 (deferred via #1162) |
| LoC core | new `filter_registry.rs` + `template_filters.py` + extensions to `filters.rs` / `renderer.rs` / `rust_bridge.py` |
| CI duration | All 13 checks green on first push |
| Stages touched | 14/14 |

## Stage Outcomes

| Stage | Outcome |
|---|---|
| 1-3 | ENV_OK / CODE_CHANGES / MERGE_CLEAN (inline) |
| 4-10 | Implementation agent — eager bridge, 10 tests, PR opened |
| 11 Code Review | APPROVE, 0 🔴 / 6 🟡 (independent reviewer) |
| 12 Review Verdict | All 6 🟡 deferred via #1162 |
| 13 Merge PR | PR_MERGED at `87ddf976` |
| 14 Retrospective | (this file) |

## Closes

- #1121 — Rust template renderer rejects project-defined `@register.filter`

## Follow-up filed

- #1162 — 6 🟡 follow-ups (hot-path Mutex perf, hardcoded autoescape, weak negative-case test, unused `custom_filter_exists`, fixture clears global registry, silent async filter handling)

## Drain progress

v0.9.1 milestone: **2 of 7 PRs shipped** (#1159, #1161). Next: #1146+#1147 grouped (ADR-015 PR-B deferrals — A075 system check + CSP nonce activator).

PR #1166, opened 2026-04-27
feat(rust): {% live_render %} lazy=True parity (closes #1145)

johnrtipton:
## Summary

- Closes the v0.9.0 PR-B (#1138) gap: `{% live_render lazy=True %}` now works on the Rust template engine path. Production users on `RustLiveView` can use streaming without falling back to the Django template engine.
- The Rust handler delegates to the existing Django implementation in `djust.templatetags.live_tags.live_render`, guaranteeing byte-for-byte parity by construction — no code duplication, no risk of the two paths drifting apart.
- Bridge mechanism: a new `Context::raw_py_objects()` getter exposes the raw Python sidecar (`request`, `view`) and a new `call_handler_with_py_sidecar` variant in the registry threads it onto the handler's context dict. Existing handlers (`url`, `static`, `dj_flash` …) ignore the extra Python objects in their dict — fully backward compatible.

## Design choice

Considered three approaches:

1. **Reimplement lazy=True in pure Python at the Rust handler layer** — would duplicate ~210 LoC of Django-path logic. Rejected: high risk of drift, worst maintenance posture.
2. **Add a Rust-side hook (e.g. `register_live_render_handler`) with bespoke API** — bigger surface area, more bridge code. Rejected: doesn't scale to future Rust-path tags that need Python sidecar (every `view`/`request`-aware tag would re-implement the plumbing).
3. **Extend the generic custom-tag registry to forward the raw Python sidecar; register a thin handler that delegates to the Django function.** Chosen.

The chosen path threads `Context::raw_py_objects` into the existing `call_handler` bridge in a backward-compatible way (existing handlers see extra dict keys they don't consume). The `live_render` handler then becomes a ~140 LoC Python file that parses the Rust-resolved arg list back into `(view_path, **kwargs)` and calls the Django function — which already has the lazy=True / sticky / CSP-nonce / thunk-stash logic exercised by 14 existing tests.

## Behaviour matrix

| Form                              | Django path | Rust path |
|-----------------------------------|-------------|-----------|
| `lazy=True`                       | works v0.9.0 | works (this PR) |
| `lazy="visible"`                  | works v0.9.0 | works (this PR) |
| `lazy={...}` dict form            | works v0.9.0 | works (this PR) |
| `sticky=True lazy=True` collision | TemplateSyntaxError | TemplateSyntaxError |
| Eager (no lazy=)                  | works | works (this PR) |
| Inline-attribute `template = "..."` | works | works (this PR) |

## Issue × file × test mapping

[...]

PR #1164, opened 2026-04-27
chore(v0.9.1): hygiene group — replay arg validation + markdown dev-dep + descriptor component test (closes #1148, #1149, #1150)

johnrtipton:
## Summary

Three small v0.9.1 hygiene items batched into one PR.

| Issue | Files modified | Tests added |
|-------|----------------|-------------|
| #1148 — Replay handler arg validation | `python/djust/time_travel.py` | `TestReplayHandlerValidation` (3 cases) in `tests/unit/test_time_travel.py` |
| #1149 — `markdown` + `nh3` in `[dev]` deps | `pyproject.toml`, `uv.lock` | (none — verified by clean `uv sync --extra dev` install) |
| #1150 — Descriptor-pattern component time-travel verification | `tests/unit/test_time_travel.py` | `TestDescriptorPatternComponentTimeTravel` (2 cases) |

### #1148 — Defense-in-depth on `replay_event`

PR #1142 shipped `replay_event(view, snapshot, override_params, record_replay)` with a dunder-rejection guard (`event_name.startswith("_")`). That admits ANY non-underscore method on the view — helpers, inherited utilities, property getters — even though the dispatcher only ever invokes `@event_handler`-decorated methods. A hand-edited or malicious snapshot could replay e.g. `view.delete_all_records()` even when that method was never exposed to the dispatcher.

The fix calls `djust.decorators.is_event_handler(handler)` after attribute resolution, mirroring the dispatcher's own acceptance criteria (see `websocket.py` ~ line 4389 server_push handler validation). Unregistered methods log a warning and return `None`.

3 regression cases:
- `test_replay_registered_handler_succeeds` — `@event_handler`-decorated method still works.
- `test_replay_unregistered_method_returns_none` — `view.some_helper` (callable, public, undecorated) is rejected.
- `test_replay_dunder_still_rejected` — pre-existing dunder guard still fires.

### #1149 — `markdown` + `nh3` in `[dev]` deps

Both packages are runtime deps of the `[components]` extra (see `python/djust/components/components/markdown.py`). `python/djust/components/components/__init__.py` eagerly imports `from .markdown import Markdown`, so any test that pulls in `djust.components.components` (directly or transitively) fails collection in a clean checkout that ran only `uv sync` without the `[components]` extra. The bisect agent in PR #1159 hit this on a fresh clone.

Added both to `[project.optional-dependencies.dev]`. No behaviour change for runtime users — the `[components]` extra already lists both as runtime deps with the same version pins.

[...]

PR #1163, opened 2026-04-27
feat(checks): A075 sticky+lazy + CSP-nonce-aware lazy-fill (closes #1146, #1147)

johnrtipton:
## Summary

Two batched v0.9.1 deferrals from ADR-015 §\"Deferred from PR-B\":

- **#1146 — A075 (Warning):** startup-time scan that flags `{% live_render ... sticky=True lazy=True %}` collisions during `manage.py check`, before any request hits. Sticky preservation requires the slot to exist at mount-frame time so the WS reattach can `replaceWith` the stashed subtree; `lazy=True` defers slot rendering until after the parent shell flushes — the stash target doesn't exist when reattach runs. The runtime tag-eval path already raises `TemplateSyntaxError`; A075 promotes that to a startup warning. Skips `{% verbatim %}` regions (re-uses `_strip_verbatim_blocks` from #1004). Silenceable via `DJUST_CONFIG = {\"suppress_checks\": [\"A075\"]}`.

- **#1147 — CSP-nonce-aware activator:** `{% live_render lazy=True %}` now propagates `request.csp_nonce` (the Django-csp middleware convention) onto BOTH the `<template id=\"djl-fill-X\">` element AND the inline `<script>` activator. Sites with strict CSP (`script-src 'nonce-...'`, no `'unsafe-inline'`) previously had the activator silently rejected at parse time; lazy children never mounted. Uses the existing `djust.utils.get_csp_nonce` helper — no extra config required for any CSP middleware following the Django convention. Backward-compatible: no nonce attribute is emitted when `request.csp_nonce` is absent.

## Issue × file × test mapping

| Issue | Production files | Test files |
| --- | --- | --- |
| #1146 (A075) | `python/djust/checks.py` (regex + scan loop in `check_templates`) | `tests/unit/test_checks_a075.py` (8 cases) |
| #1147 (CSP) | `python/djust/templatetags/live_tags.py` (lazy=True branch + thunk envelope) | `tests/unit/test_lazy_render_csp.py` (6 cases) + `tests/js/lazy_fill_csp.test.js` (3 cases) |
| Both | `CHANGELOG.md`, `docs/website/guides/streaming-render.md` | — |

## Test count delta

- **+8** Python (A075): `TestA075StickyLazyCollision` covers collision firing, sticky-only/lazy-only silence, no-tag silence, verbatim suppression, real-call-next-to-verbatim-example, config disable knob, string-truthy kwargs.
[...]

PR #1168, opened 2026-04-27
feat(theming): namespace cookie names per project (closes #1158)

### LINKED ISSUE #1158: tech-debt: namespace djust theming cookies to prevent cross-project bleed on localhost
Source: User-reported during v0.9.0rc2 verification. Follow-up to #1013 (closed-as-workaround).

## Context

The #1013 fix added `LIVEVIEW_CONFIG['theme']['enable_client_override']: False` so sites without a user-facing theme switcher could opt out of reading theme cookies. That removes the bleed for those sites — at the cost of disabling persistent user preference.

Sites WITH a user-facing theme switcher (djust.org, djustlive, downstream projects with end-user theming) still need cookies on. On localhost they all answer at `localhost:80xx` and share one cookie jar (browsers scope cookies by domain, not by port). When two such projects run side-by-side, picking a palette in one silently overwrites the other's. `djust_theme`, `djust_theme_preset`, `djust_theme_pack`, and `djust_theme_layout` all collide.

Reproduces today on the same setup #1013 documented:
- `localhost:8000` djust.org with switcher → set pack=`nyc_core`
- `localhost:8003` docs.djust.org with switcher → settings-level pin to `djust`
- The cookie wins → docs renders with nyc_core palette.

## Proposed fix

Per-app cookie name prefix driven by a Django setting:

```python
# settings.py
LIVEVIEW_CONFIG = {
    'theme': {
        'cookie_namespace': 'djust_org',  # or 'djustlive', 'docs', etc.
    },
}
```

`ThemeManager.get_state()` (`python/djust/theming/manager.py:292-296`) and the JS theme-switcher (`python/djust/theming/build_themes.py:398/431`) read/write namespaced cookies:

| Setting | Cookie name |
|---|---|
| `cookie_namespace` unset | `djust_theme`, `djust_theme_preset`, `djust_theme_pack`, `djust_theme_layout` (current names — back-compat) |
| `cookie_namespace = 'djust_org'` | `djust_org_theme`, `djust_org_theme_preset`, `djust_org_theme_pack`, `djust_org_theme_layout` |

Read order: namespaced name first, then fall back to the unprefixed name (one-time migration window). Write only the namespaced name when set.

## Scope

- `python/djust/theming/manager.py` — read both, prefer namespaced; ~10 LoC.
- `python/djust/theming/build_themes.py` — emit namespaced cookie name in the generated theme-switcher JS; small string-template change.
- `python/djust/theming/context_processors.py` — `djust-theme-mode` localStorage key already lives there but localStorage is port-isolated; update only if we add cookie reads alongside.
- Docs: add a one-paragraph note in `docs/website/guides/theming.md` (or wherever `enable_client_override` is documented) recommending `cookie_namespace` for any project that runs alongside other djust projects on the same domain.
- Tests: cookie read/write round-trip, fall-back to unprefixed cookie when namespace unset, no leak between two namespaces in the same request fixture.

## Why `cookie_namespace` over alternatives

- **Cookie path scoping** (`Path=/admin/` etc.) — most projects mount at `/`, so doesn't help on localhost.
- **localStorage instead of cookies** — already used for some bits, but SSR can't read it on first paint, so we'd lose first-paint correctness.
- **Drop cookies entirely** — same as above.
- **Auto-derive from `SECRET_KEY`/site name** — surprising; an explicit setting is clearer and lets ops control collision.

## Tradeoff to flag in docs

Every downstream consumer that wants isolation has to set `cookie_namespace` once. If they don't, they fall back to the current unprefixed names — and remain susceptible to the same bleed. Acceptable since it's a localhost-dev concern; production sites have separate domains and never collide.

## Related

- #1013 (closed) — the `enable_client_override` opt-out workaround.
- `python/djust/theming/manager.py:283-285` already documents the cookie-jar problem in a comment that ends "`djust_theme_pack` set by project A pins the palette for project B". That comment should be replaced with a pointer to this issue once `cookie_namespace` lands.

johnrtipton:
## Summary

Closes #1158. Adds opt-in `LIVEVIEW_CONFIG['theme']['cookie_namespace']` so multiple djust projects on the same domain (e.g. `localhost:80xx`) don't overwrite each other's theme preferences. Browsers scope cookies by domain only — not by port — so the four `djust_theme*` cookies bleed across projects without this. PR #1013 already shipped `enable_client_override: False` as a workaround, but that breaks sites with a user-facing theme switcher; this is the missing piece for those sites.

```python
LIVEVIEW_CONFIG = {
    "theme": {"cookie_namespace": "djust_org"},
}
```

## Design choice

- **Read side (`ThemeManager.get_state`)**: read namespaced first, fall back to legacy unprefixed name once on upgrade so users keep their existing theme.
- **Write side**: only the namespaced name is written. The shipped `theme.js` derives `COOKIE_KEY_*` from `window.__djust_theme_cookie_prefix`, which `theme_head.html` injects from `LIVEVIEW_CONFIG['theme']['cookie_namespace']`. No regeneration of static JS needed.
- **Default**: when `cookie_namespace` is unset, the legacy unprefixed names are used — existing deployments unaffected.

## Behavior matrix

| `cookie_namespace` | Namespaced cookie present | Unprefixed cookie present | Result |
|---|---|---|---|
| unset | — | yes | reads unprefixed (legacy) |
| unset | — | no  | falls back to session/config |
| `"ns"` | yes | (any)  | reads namespaced (wins over unprefixed) |
| `"ns"` | no  | yes | reads unprefixed (one-time migration fallback) |
| `"ns"` | no  | no  | falls back to session/config |

## Back-compat

When `cookie_namespace` is unset (default), behavior is byte-for-byte identical to v0.9.0:
- Cookies read/written under the unprefixed names.
- The new `window.__djust_theme_cookie_prefix` global is set to `""`, which makes `theme.js`'s `COOKIE_PREFIX + 'djust_theme'` evaluate to the legacy literal `'djust_theme'`.

When `cookie_namespace` is set (opt-in):
- First request after upgrade: server reads the legacy unprefixed cookie (fallback) so the theme is preserved across the cutover.
- Next user click on the theme switcher: `theme.js` writes the namespaced cookie.
- Subsequent requests: server reads the namespaced cookie.

## Test count

- 8 new regression cases in `tests/unit/test_theming_cookie_namespace_1158.py`.
- Full suite: 4583 passed, 14 skipped (unchanged baseline).
- Targeted theming subset: 643 passed.

## Files touched

- `python/djust/theming/manager.py` — read-side namespace logic in `ThemeManager.get_state`.
- `python/djust/theming/static/djust_theming/js/theme.js` — `COOKIE_KEY_*` constants derived from `window.__djust_theme_cookie_prefix`.
- `python/djust/theming/templates/djust_theming/theme_head.html` — emits the prefix global before the existing anti-FOUC script.
- `python/djust/theming/templatetags/theme_tags.py` — computes `cookie_prefix_js` (JSON-encoded for safe inline injection).
- `tests/unit/test_theming_cookie_namespace_1158.py` (new) — 8 regression cases.
- `docs/website/guides/components.md` — "Cross-project cookie isolation (`cookie_namespace`)" subsection in the djust-theming guide.
- `CHANGELOG.md` — `[Unreleased] / Added` entry.

## Test plan

- [x] `pytest tests/unit/test_theming_cookie_namespace_1158.py` — 8/8 pass
- [x] `pytest python/djust/ tests/` — 4583 pass, no regressions
- [x] `pre-push` hook (full suite + checks) — passed on push
- [ ] Manual: configure `cookie_namespace="djust_org"` in `djust.org`, reload, click switcher, inspect Cookies tab — verify only `djust_org_djust_theme*` cookies are written
- [ ] Manual: spin up two djust projects on `localhost:8001` and `localhost:8002` with different namespaces, switch theme on one, verify the other is unaffected

## Stage 11 self-review notes

- 🟡 **`build_themes.py` not modified.** The task brief mentioned updating its JS template; on inspection that file generates `djust-theme-switcher.js` which is **localStorage-only** (no cookie writes). LocalStorage is origin-scoped (port-aware), so there's no cross-project bleed in the bug class for that path. The actual cookie writes live in `static/djust_theming/js/theme.js`, which this PR does update.
- 🟡 The implementation deliberately does NOT touch session keys (`session_key="djust_theme"` in `DEFAULT_CONFIG`). Sessions are server-scoped per Django's settings and don't suffer the cross-project cookie-jar problem. Out of scope.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,28 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- **Theming cookie namespace for per-project isolation on shared
+  domains (closes #1158)** — adds opt-in
+  `LIVEVIEW_CONFIG['theme']['cookie_namespace']` setting so multiple
+  djust projects on `localhost:80xx` (or any shared domain) don't
+  overwrite each other's theme preferences. Browsers scope cookies by
+  domain only — not by port — so the four `djust_theme*` cookies bleed
+  across projects without this. PR #1013 already shipped
+  `enable_client_override: False` as a workaround, but that breaks
+  sites with a user-facing theme switcher; this is the missing piece
+  for those sites. When `cookie_namespace="djust_org"` is set, the
+  cookies become `djust_org_djust_theme`, `djust_org_djust_theme_preset`,
+  `djust_org_djust_theme_pack`, `djust_org_djust_theme_layout`.
+  Read path tries namespaced first, falls back to unprefixed once on
+  upgrade so users keep their existing theme. Write path (`theme.js`)
+  reads `window.__djust_theme_cookie_prefix` injected by
+  `theme_head.html` and writes only the namespaced name when set. When
+  unset (default), the legacy unprefixed names are used — existing
+  deployments unaffected. 8 new regression cases in
+  `tests/unit/test_theming_cookie_namespace_1158.py` cover namespaced
+  precedence, unprefixed fallback, default back-compat, two-namespace
+  isolation, all four cookies honour the namespace, and the
+  `theme_head.html` + `theme.js` write-side wiring.
 - **Rust template engine `{% live_render %}` lazy=True parity (closes
   #1145)** — the Rust template engine now ships a registered handler
   for `{% live_render %}`, closing the v0.9.0 PR-B (#1138) gap. Before
--- a/docs/website/guides/components.md
+++ b/docs/website/guides/components.md
@@ -818,6 +818,44 @@ Then use theme colors in Tailwind classes:
 </button>
 ```
 
+### Cross-project cookie isolation (`cookie_namespace`)
+
+If you run two or more djust projects on the same domain (e.g.
+`localhost:8001`, `localhost:8002`, etc.), browsers scope cookies by domain
+only — *not* by port. Without isolation, the four theming cookies
+(`djust_theme`, `djust_theme_preset`, `djust_theme_pack`,
+`djust_theme_layout`) leak across projects, so picking the "Cyberpunk" pack
+in project A pins it for project B too.
+
+For sites without a user-facing theme switcher, the simpler fix is
+`enable_client_override: False` (introduced in v0.8.2 / #1013), which makes
+the server ignore client cookies entirely.
+
+For sites *with* a user-facing switcher — where cookie writes must stay on —
+set a per-project namespace:
+
+```python
+LIVEVIEW_CONFIG = {
+    "theme": {
+        "cookie_namespace": "djust_org",   # any unique-per-project string
+    },
+}
+```
+
+When set, theming cookies are read and written under
+`<ns>_djust_theme`, `<ns>_djust_theme_preset`, `<ns>_djust_theme_pack`, and
+`<ns>_djust_theme_layout`, so each project gets its own slot in the shared
+cookie jar.
+
+The read path tries the namespaced cookie first and falls back to the legacy
+unprefixed cookie once on the first request after upgrade — users keep their
+existing theme choice. After the user clicks the switcher (or any other path
+that calls `setPreset` / `setPack` / `setTheme` / `setLayout`), only the
+namespaced cookie is written.
+
+When `cookie_namespace` is unset (default), the unprefixed names are used —
+existing deployments don't lose their theme on upgrade.
+
 ## Choosing a Styling Approach
 
 | Approach                                 | When to Use                                                                                                                           |
--- a/python/djust/theming/manager.py
+++ b/python/djust/theming/manager.py
@@ -283,18 +283,42 @@ class ThemeManager:
         # answering on localhost shares a cookie jar — `djust_theme_pack` set
         # by project A pins the palette for project B). Default ``True`` for
         # back-compat: sites with a user-facing switcher keep working.
+        #
+        # #1158 — sites WITH a user-facing switcher (so cookies must stay on)
+        # can opt into a per-project cookie namespace via
+        # ``LIVEVIEW_CONFIG['theme']['cookie_namespace']: '<ns>'``. When set,
+        # the four theming cookies become ``<ns>_djust_theme``,
+        # ``<ns>_djust_theme_preset``, ``<ns>_djust_theme_pack``,
+        # ``<ns>_djust_theme_layout``. Read-side falls back to the unprefixed
+        # name for one-time migration; write-side (theme.js) writes only the
+        # namespaced name when the prefix is set.
         theme = None
         preset = None
         pack = None
         layout = ""
         enable_client_override = bool(self.config.get("enable_client_override", True))
         if self.request and enable_client_override:
-            theme = self.request.COOKIES.get("djust_theme")
-            preset = self.request.COOKIES.get("djust_theme_preset")
-            pack = self.request.COOKIES.get("djust_theme_pack")
-            layout = self.request.COOKIES.get("djust_theme_layout", "")
+            cookies = self.request.COOKIES
+            ns = (self.config.get("cookie_namespace") or "").strip()
+            prefix = f"{ns}_" if ns else ""
+
+            # Read namespaced first; fall back to unprefixed (migration window).
+            def _read(name: str, default: str = "") -> str:
+                if prefix:
+                    return cookies.get(f"{prefix}{name}", cookies.get(name, default))
+                return cookies.get(name, default)
+
+            theme = _read("djust_theme") or None
+            preset = _read("djust_theme_preset") or None
+            pack = _read("djust_theme_pack") or None
+            layout = _read("djust_theme_layout", "")
             logger.debug(
-                "Cookies: theme=%s, preset=%s, pack=%s, layout=%s", theme, preset, pack, layout
+                "Cookies (ns=%r): theme=%s, preset=%s, pack=%s, layout=%s",
+                ns,
+                theme,
+                preset,
+                pack,
+                layout,
             )
         elif self.request:
             logger.debug("Cookies skipped — LIVEVIEW_CONFIG.theme.enable_client_override=False")
--- a/python/djust/theming/static/djust_theming/js/theme.js
+++ b/python/djust/theming/static/djust_theming/js/theme.js
@@ -18,10 +18,19 @@
     const STORAGE_KEY_PACK = 'djust-theme-pack';
     const STORAGE_KEY_LAYOUT = 'djust-theme-layout';
 
-    const COOKIE_KEY_PRESET = 'djust_theme_preset';
-    const COOKIE_KEY_THEME = 'djust_theme';
-    const COOKIE_KEY_PACK = 'djust_theme_pack';
-    const COOKIE_KEY_LAYOUT = 'djust_theme_layout';
+    // #1158 — Per-project cookie namespace. The server emits
+    // `window.__djust_theme_cookie_prefix` from `theme_head.html` based on
+    // `LIVEVIEW_CONFIG['theme']['cookie_namespace']`. Empty string means the
+    // legacy unprefixed cookie names (back-compat). When set, every theming
+    // cookie this script writes is prefixed so two djust projects on the
+    // same localhost domain don't overwrite each other's theme preferences.
+    const COOKIE_PREFIX = (typeof window !== 'undefined' && typeof window.__djust_theme_cookie_prefix === 'string')
+        ? window.__djust_theme_cookie_prefix
+        : '';
+    const COOKIE_KEY_PRESET = COOKIE_PREFIX + 'djust_theme_preset';
+    const COOKIE_KEY_THEME = COOKIE_PREFIX + 'djust_theme';
+    const COOKIE_KEY_PACK = COOKIE_PREFIX + 'djust_theme_pack';
+    const COOKIE_KEY_LAYOUT = COOKIE_PREFIX + 'djust_theme_layout';
 
     class DjustThemeManager {
         constructor() {
--- a/python/djust/theming/templates/djust_theming/theme_head.html
+++ b/python/djust/theming/templates/djust_theming/theme_head.html
@@ -1,10 +1,12 @@
 {# Theme head block: anti-FOUC script, CSS, and optional JS include #}
 {# Safety: autoescape off is safe here — all variables (css_block, direction, etc.) #}
 {# are framework-generated in theme_tags.py, not user input. #}
+{# cookie_prefix_js is a JSON-encoded string ("" or "<ns>_"), framework-generated. #}
 {% load static %}
 {% autoescape off %}
 <script>
 (function() {
+    window.__djust_theme_cookie_prefix = {{ cookie_prefix_js }};
     {% if loading_class %}document.documentElement.classList.add('loading');
     {% endif %}var storageKey = 'djust-theme-mode';
     var storedMode = localStorage.getItem(storageKey);
--- a/python/djust/theming/templatetags/theme_tags.py
+++ b/python/djust/theming/templatetags/theme_tags.py
@@ -17,6 +17,8 @@ Usage:
     {% theme_preset_selector layout="dropdown" %}
 """
 
+import json
+
 from django import template
 from django.template.loader import render_to_string
 from django.urls import reverse, NoReverseMatch
@@ -129,6 +131,13 @@ def theme_head(context, include_js: bool = True, link_css: bool = False):
     # Resolve text direction
     direction = get_direction()
 
+    # #1158 — namespace prefix for theming cookies (cross-project isolation
+    # on shared domains like localhost). JSON-encoded for safe inlining in a
+    # <script> context (json.dumps gives a valid JS string literal).
+    ns = (config.get("cookie_namespace") or "").strip()
+    cookie_prefix = f"{ns}_" if ns else ""
+    cookie_prefix_js = json.dumps(cookie_prefix)
+
     # Render via shared template
     html = render_to_string(
         "djust_theming/theme_head.html",
@@ -140,6 +149,7 @@ def theme_head(context, include_js: bool = True, link_css: bool = False):
             "include_component_link": include_component_link,
             "include_js": include_js,
             "direction": direction,
+            "cookie_prefix_js": cookie_prefix_js,
         },
     )
     return mark_safe(html)
--- a/tests/unit/test_theming_cookie_namespace_1158.py
+++ b/tests/unit/test_theming_cookie_namespace_1158.py
@@ -1,0 +1,230 @@
+"""Regression tests for #1158 — per-project cookie namespace for theming.
+
+Background
+----------
+On localhost, cookies are scoped by domain only (not port). Multiple djust
+projects on `localhost:80xx` share the `djust_theme*` cookie jar and overwrite
+each other. PR #1013 added `enable_client_override: False` as a workaround,
+but sites with a user-facing theme switcher need cookie writes ON, so the
+bleed persists for them.
+
+Fix
+---
+`LIVEVIEW_CONFIG['theme']['cookie_namespace']: '<ns>'` causes the four
+theming cookies to be read/written as `<ns>_djust_theme`,
+`<ns>_djust_theme_preset`, `<ns>_djust_theme_pack`, `<ns>_djust_theme_layout`.
+
+Read order: namespaced first, fall back to unprefixed for one-time migration.
+Write order: only namespaced (handled by `theme.js` reading
+`window.__djust_theme_cookie_prefix` injected by `theme_head.html`).
+
+These tests cover the read side (Python) and the write-side wiring (the
+template emits the right `__djust_theme_cookie_prefix` value, and the JS
+file's cookie-key constants resolve from that global).
+"""
+
+from __future__ import annotations
+
+import re
+from pathlib import Path
+
+import pytest
+from django.template import Context, Template
+from django.test import RequestFactory
+
+
+# ---------------------------------------------------------------------------
+# Read side — ThemeManager.get_state()
+# ---------------------------------------------------------------------------
+
+
+@pytest.mark.django_db
+def test_namespaced_cookie_wins_when_namespace_set():
+    """When `cookie_namespace` is set, the namespaced cookie is read."""
+    from djust.theming.manager import ThemeManager
+
+    rf = RequestFactory()
+    req = rf.get("/")
+    req.COOKIES["djust_org_djust_theme_pack"] = "nyc_core"
+    # Stale unprefixed cookie from another project on the same domain — must NOT win.
+    req.COOKIES["djust_theme_pack"] = "djust"
+    req.session = {}
+
+    mgr = ThemeManager(request=req)
+    mgr.config = dict(mgr.config, cookie_namespace="djust_org", enable_client_override=True)
+    state = mgr.get_state()
+    assert state.pack == "nyc_core", (
+        f"namespaced cookie must win over stale unprefixed cookie; got pack={state.pack!r}"
+    )
+
+
+@pytest.mark.django_db
+def test_namespace_set_falls_back_to_unprefixed_when_namespaced_missing():
+    """Migration window: when the namespaced cookie hasn't been written yet
+    (first request after upgrade), fall back to the unprefixed cookie so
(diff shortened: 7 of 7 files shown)


johnrtipton (commented):
## Stage 11 Code Review — PR #1168 (djust-org/djust)

`feat(theming): namespace cookie names per project (closes #1158)`

**Verdict: APPROVE with non-blocking comments.** Surgical opt-in fix for cross-project cookie bleed on shared domains. Default-unset path is byte-for-byte back-compat. Two-commit split (impl+tests / docs+CHANGELOG) is clean. JSON-encoded inline injection is correct. Eight tests cover the main read-side branches and the template wiring; two minor gaps + two developer-experience nits noted below — none worth blocking on.

---

### 🟢 What's right

**Two-commit split is clean.** Commit `86f2934` touches only the 5 impl/test files (`manager.py`, `theme.js`, `theme_head.html`, `theme_tags.py`, `tests/unit/test_theming_cookie_namespace_1158.py`). Commit `da363d3` touches only `CHANGELOG.md` + `docs/website/guides/components.md`. Per the v0.9.1 retro pattern. Verified via `gh api /repos/djust-org/djust/commits/<sha>`.

**JSON-encoding is correct.** `theme_tags.py:139` uses `json.dumps(cookie_prefix)` — produces a valid JS string literal with proper escaping. The result is interpolated raw inside `<script>...</script>` at `theme_head.html:9`:
```html
window.__djust_theme_cookie_prefix = {{ cookie_prefix_js }};
```
Since the value passed to `json.dumps` is developer-controlled (Django settings, not request input), and `json.dumps` already escapes embedded `"`, `\`, control chars, and non-ASCII, the inline injection is safe. Note `json.dumps` does NOT escape `/`, so a hypothetical future where `cookie_namespace` becomes user-supplied would need `json.dumps(...).replace('</', '<\\/')` to defeat `</script>` breakouts. Not a current bug — settings-only.

**All four cookies and all six write sites are prefix-aware.** `theme.js:30-33` derives `COOKIE_KEY_PRESET`, `COOKIE_KEY_THEME`, `COOKIE_KEY_PACK`, `COOKIE_KEY_LAYOUT` from `COOKIE_PREFIX`, and all six `document.cookie =` writes (`theme.js:188, 209, 225, 236, 244, 344`) interpolate via `${COOKIE_KEY_*}`. Path/SameSite/max-age attributes preserved literally.

**localStorage NOT prefixed (correctly).** localStorage is origin-scoped per W3C — `localhost:8001` and `localhost:8002` are different origins, no bleed. Implementer's self-flag is correct.

**Empty-namespace + no-namespace coalesce.** `manager.py:302` does `(self.config.get("cookie_namespace") or "").strip()` — both `None` and `""` produce `prefix = ""`, exercising the unprefixed branch. Same coalescing in `theme_tags.py:137`. Symmetric on read and write sides.

**`build_themes.py` correctly skipped.** `grep -n "document.cookie\|cookie" python/djust/theming/build_themes.py` returns nothing — confirms the generated `djust-theme-switcher.js` is localStorage-only and has no cross-project bleed in the bug class. Implementer's self-flag #1 is accurate.

**Two-namespace isolation tested.** `test_two_namespace_isolation` (lines 85-111) puts both `a_djust_theme_pack` and `b_djust_theme_pack` in the same cookie jar and verifies project A reads `pack_a` while project B (separate ThemeManager) reads `pack_b`.

**CHANGELOG matches the diff.** References all 4 namespaced cookie names, both fallback logic and write-side wiring, and the 8-test count. No drift from the implementation.

---

### 🟡 Non-blocking findings

**1. Stale empty namespaced cookie defeats migration fallback.**

`manager.py:308-311`:
```python
def _read(name: str, default: str = "") -> str:
    if prefix:
        return cookies.get(f"{prefix}{name}", cookies.get(name, default))
    return cookies.get(name, default)

theme = _read("djust_theme") or None
```
If the namespaced cookie is present-but-empty (e.g. a previous `clearPack`-style write that left `proj_djust_theme=""`, or browser DevTools manual cookie-set), `_read` returns `""` (not the default), so the migration `cookies.get(name, default)` branch never fires. Then `"" or None` → `None`, and the read cascades down to session/config default. The legacy unprefixed cookie is **NOT** consulted in this case.

This is probably intentional (an empty namespaced cookie is a deliberate "I cleared this" signal, and `clearPack` at `theme.js:236` deliberately sets `${COOKIE_KEY_PACK}=` to clear). But it's inconsistent with the documented "fall back to unprefixed when namespaced missing" — "missing" vs "empty" is a thin distinction.

**Suggested fix (if you want migration to dominate empty)**: change `cookies.get(f"{prefix}{name}", ...)` to a truthy-check:
```python
val = cookies.get(f"{prefix}{name}", "")
if not val:
    val = cookies.get(name, default)
return val
```
Or document the empty-cookie semantics in the docstring.

**2. No `cookie_namespace` validation / system check.**

`manager.py:302` only `.strip()`s the namespace. If a developer sets:
```python
LIVEVIEW_CONFIG = {"theme": {"cookie_namespace": "my ns"}}    # internal whitespace
LIVEVIEW_CONFIG = {"theme": {"cookie_namespace": "ns;evil"}}  # semicolon — cookie attribute injection
LIVEVIEW_CONFIG = {"theme": {"cookie_namespace": "a=b"}}      # equals sign
```
…the resulting cookie name `my ns_djust_theme` is malformed and the browser silently drops it. The site appears to "lose theme persistence" with no error. Not a security issue (only affects the developer's own deployment, and a settings-controlled `;` doesn't escalate XSS), but a Stage 11 developer-experience nit.

Per `python/djust/checks.py` (C/V/S/T/Q categories noted in `djust/CLAUDE.md`), a config-category system check `djust.C### — cookie_namespace must match [A-Za-z0-9_-]+` would catch this at startup — `manage.py runserver` would print a clear error instead of silently dropping cookies. Worth a follow-up issue, not a blocker for this PR.

**3. Test gap: no JSDOM test for theme.js actually WRITING the prefixed cookie.**

`test_theme_js_reads_cookie_prefix_from_window_global` (line 204) is a static-source assertion — it greps `theme.js` for the literal `COOKIE_KEY_THEME = COOKIE_PREFIX + 'djust_theme'`. That locks in the source shape, but does not exercise:
- `setPack('foo')` with `window.__djust_theme_cookie_prefix = 'proj_'` → assert `document.cookie.includes('proj_djust_theme_pack=foo')`
- `setPack('foo')` with no `__djust_theme_cookie_prefix` → assert legacy `djust_theme_pack=foo`

Per `djust/CLAUDE.md`: "New JS feature files in `static/djust/src/` need corresponding tests in `tests/js/`." This is `static/djust_theming/js/`, not `static/djust/src/`, so it's not strictly bound by that rule, and `theme.js` is pre-existing — but a one-off vitest spec would close the test loop end-to-end. Filed as a nice-to-have.

**4. Migration window: legacy unprefixed cookie never deleted.**

After a user adopts `cookie_namespace="proj"`, clicks the switcher once, and the JS writes `proj_djust_theme_pack=...`, the legacy `djust_theme_pack=...` cookie remains in the jar forever (until the browser auto-expires or the user clears cookies). The PR body and CHANGELOG describe this as one-time migration fallback. The docs section at `docs/website/guides/components.md:830-857` says "the legacy unprefixed cookie ... once on the first request after upgrade — users keep their existing theme choice. After the user clicks the switcher ... only the namespaced cookie is written." — accurate, but does not call out that the legacy cookie persists.

A future `setPack` could delete the legacy cookie alongside the namespaced write:
```javascript
if (COOKIE_PREFIX) {
    // Delete legacy unprefixed cookie, now that we own a namespaced one.
    document.cookie = `djust_theme_pack=;path=/;expires=Thu, 01 Jan 1970 00:00:00 GMT`;
}
```
…but that's a follow-up. Not worth gating this PR.

---

### Question

The migration fallback at `manager.py:308` is one-directional — read-side only. Is there a planned moment to delete the legacy unprefixed cookie (e.g. on the first namespaced write in `theme.js`'s `setPack`/`setPreset`/`setTheme`/`setLayout`)? Otherwise users who upgrade to namespaced will have permanently-orphaned `djust_theme*` cookies in their jar, and a developer who ever rolls back to no-namespace will see the orphan reactivate. A follow-up issue tracking this would be reassuring even if the work itself is deferred.

---

### Two-commit integrity

| Commit | Files | Hunks |
|---|---|---|
| `86f2934` impl+tests | `manager.py`, `theme.js`, `theme_head.html`, `theme_tags.py`, `test_theming_cookie_namespace_1158.py` | impl + 8 tests, **no CHANGELOG** |
| `da363d3` docs+CHANGELOG | `CHANGELOG.md`, `docs/website/guides/components.md` | docs subsection + Unreleased/Added entry, **no impl** |

Verified clean via `gh api /repos/djust-org/djust/commits/<sha>`.

---

### Summary counts

- 🔴 Critical: 0
- 🟡 Important / nits: 4 (empty-cookie migration semantics, no namespace validation, no JSDOM write test, legacy cookie not deleted)
- 🟢 Confirmed correct: 9 (two-commit split, JSON-encoding, all-four-cookies, localStorage skip, empty/None coalesce, build_themes.py skip, two-namespace test, CHANGELOG accuracy, no collateral)

**Recommendation: merge.** None of the 🟡 findings block. Filing follow-up issues for #2 (system check) and #4 (legacy cookie cleanup) would be the cleanest tail.

johnrtipton:
## Stage 11 Code Review — PR #1168 (djust-org/djust)

`feat(theming): namespace cookie names per project (closes #1158)`

**Verdict: APPROVE with non-blocking comments.** Surgical opt-in fix for cross-project cookie bleed on shared domains. Default-unset path is byte-for-byte back-compat. Two-commit split (impl+tests / docs+CHANGELOG) is clean. JSON-encoded inline injection is correct. Eight tests cover the main read-side branches and the template wiring; two minor gaps + two developer-experience nits noted below — none worth blocking on.

---

### 🟢 What's right

**Two-commit split is clean.** Commit `86f2934` touches only the 5 impl/test files (`manager.py`, `theme.js`, `theme_head.html`, `theme_tags.py`, `tests/unit/test_theming_cookie_namespace_1158.py`). Commit `da363d3` touches only `CHANGELOG.md` + `docs/website/guides/components.md`. Per the v0.9.1 retro pattern. Verified via `gh api /repos/djust-org/djust/commits/<sha>`.

**JSON-encoding is correct.** `theme_tags.py:139` uses `json.dumps(cookie_prefix)` — produces a valid JS string literal with proper escaping. The result is interpolated raw inside `<script>...</script>` at `theme_head.html:9`:
```html
window.__djust_theme_cookie_prefix = {{ cookie_prefix_js }};
```
Since the value passed to `json.dumps` is developer-controlled (Django settings, not request input), and `json.dumps` already escapes embedded `"`, `\`, control chars, and non-ASCII, the inline injection is safe. Note `json.dumps` does NOT escape `/`, so a hypothetical future where `cookie_namespace` becomes user-supplied would need `json.dumps(...).replace('</', '<\\/')` to defeat `</script>` breakouts. Not a current bug — settings-only.

**All four cookies and all six write sites are prefix-aware.** `theme.js:30-33` derives `COOKIE_KEY_PRESET`, `COOKIE_KEY_THEME`, `COOKIE_KEY_PACK`, `COOKIE_KEY_LAYOUT` from `COOKIE_PREFIX`, and all six `document.cookie =` writes (`theme.js:188, 209, 225, 236, 244, 344`) interpolate via `${COOKIE_KEY_*}`. Path/SameSite/max-age attributes preserved literally.

**localStorage NOT prefixed (correctly).** localStorage is origin-scoped per W3C — `localhost:8001` and `localhost:8002` are different origins, no bleed. Implementer's self-flag is correct.

**Empty-namespace + no-namespace coalesce.** `manager.py:302` does `(self.config.get("cookie_namespace") or "").strip()` — both `None` and `""` produce `prefix = ""`, exercising the unprefixed branch. Same coalescing in `theme_tags.py:137`. Symmetric on read and write sides.

**`build_themes.py` correctly skipped.** `grep -n "document.cookie\|cookie" python/djust/theming/build_themes.py` returns nothing — confirms the generated `djust-theme-switcher.js` is localStorage-only and has no cross-project bleed in the bug class. Implementer's self-flag #1 is accurate.

**Two-namespace isolation tested.** `test_two_namespace_isolation` (lines 85-111) puts both `a_djust_theme_pack` and `b_djust_theme_pack` in the same cookie jar and verifies project A reads `pack_a` while project B (separate ThemeManager) reads `pack_b`.

**CHANGELOG matches the diff.** References all 4 namespaced cookie names, both fallback logic and write-side wiring, and the 8-test count. No drift from the implementation.

---

### 🟡 Non-blocking findings

**1. Stale empty namespaced cookie defeats migration fallback.**

`manager.py:308-311`:
```python
def _read(name: str, default: str = "") -> str:
    if prefix:
        return cookies.get(f"{prefix}{name}", cookies.get(name, default))
    return cookies.get(name, default)

theme = _read("djust_theme") or None
```
If the namespaced cookie is present-but-empty (e.g. a previous `clearPack`-style write that left `proj_djust_theme=""`, or browser DevTools manual cookie-set), `_read` returns `""` (not the default), so the migration `cookies.get(name, default)` branch never fires. Then `"" or None` → `None`, and the read cascades down to session/config default. The legacy unprefixed cookie is **NOT** consulted in this case.

This is probably intentional (an empty namespaced cookie is a deliberate "I cleared this" signal, and `clearPack` at `theme.js:236` deliberately sets `${COOKIE_KEY_PACK}=` to clear). But it's inconsistent with the documented "fall back to unprefixed when namespaced missing" — "missing" vs "empty" is a thin distinction.

**Suggested fix (if you want migration to dominate empty)**: change `cookies.get(f"{prefix}{name}", ...)` to a truthy-check:
```python
val = cookies.get(f"{prefix}{name}", "")
if not val:
    val = cookies.get(name, default)
return val
```
Or document the empty-cookie semantics in the docstring.

**2. No `cookie_namespace` validation / system check.**

`manager.py:302` only `.strip()`s the namespace. If a developer sets:
```python
LIVEVIEW_CONFIG = {"theme": {"cookie_namespace": "my ns"}}    # internal whitespace
LIVEVIEW_CONFIG = {"theme": {"cookie_namespace": "ns;evil"}}  # semicolon — cookie attribute injection
LIVEVIEW_CONFIG = {"theme": {"cookie_namespace": "a=b"}}      # equals sign
```
…the resulting cookie name `my ns_djust_theme` is malformed and the browser silently drops it. The site appears to "lose theme persistence" with no error. Not a security issue (only affects the developer's own deployment, and a settings-controlled `;` doesn't escalate XSS), but a Stage 11 developer-experience nit.

Per `python/djust/checks.py` (C/V/S/T/Q categories noted in `djust/CLAUDE.md`), a config-category system check `djust.C### — cookie_namespace must match [A-Za-z0-9_-]+` would catch this at startup — `manage.py runserver` would print a clear error instead of silently dropping cookies. Worth a follow-up issue, not a blocker for this PR.

**3. Test gap: no JSDOM test for theme.js actually WRITING the prefixed cookie.**

`test_theme_js_reads_cookie_prefix_from_window_global` (line 204) is a static-source assertion — it greps `theme.js` for the literal `COOKIE_KEY_THEME = COOKIE_PREFIX + 'djust_theme'`. That locks in the source shape, but does not exercise:
- `setPack('foo')` with `window.__djust_theme_cookie_prefix = 'proj_'` → assert `document.cookie.includes('proj_djust_theme_pack=foo')`
- `setPack('foo')` with no `__djust_theme_cookie_prefix` → assert legacy `djust_theme_pack=foo`

Per `djust/CLAUDE.md`: "New JS feature files in `static/djust/src/` need corresponding tests in `tests/js/`." This is `static/djust_theming/js/`, not `static/djust/src/`, so it's not strictly bound by that rule, and `theme.js` is pre-existing — but a one-off vitest spec would close the test loop end-to-end. Filed as a nice-to-have.

**4. Migration window: legacy unprefixed cookie never deleted.**

After a user adopts `cookie_namespace="proj"`, clicks the switcher once, and the JS writes `proj_djust_theme_pack=...`, the legacy `djust_theme_pack=...` cookie remains in the jar forever (until the browser auto-expires or the user clears cookies). The PR body and CHANGELOG describe this as one-time migration fallback. The docs section at `docs/website/guides/components.md:830-857` says "the legacy unprefixed cookie ... once on the first request after upgrade — users keep their existing theme choice. After the user clicks the switcher ... only the namespaced cookie is written." — accurate, but does not call out that the legacy cookie persists.

A future `setPack` could delete the legacy cookie alongside the namespaced write:
```javascript
if (COOKIE_PREFIX) {
    // Delete legacy unprefixed cookie, now that we own a namespaced one.
    document.cookie = `djust_theme_pack=;path=/;expires=Thu, 01 Jan 1970 00:00:00 GMT`;
}
```
…but that's a follow-up. Not worth gating this PR.

---

### Question

The migration fallback at `manager.py:308` is one-directional — read-side only. Is there a planned moment to delete the legacy unprefixed cookie (e.g. on the first namespaced write in `theme.js`'s `setPack`/`setPreset`/`setTheme`/`setLayout`)? Otherwise users who upgrade to namespaced will have permanently-orphaned `djust_theme*` cookies in their jar, and a developer who ever rolls back to no-namespace will see the orphan reactivate. A follow-up issue tracking this would be reassuring even if the work itself is deferred.

---

### Two-commit integrity

| Commit | Files | Hunks |
|---|---|---|
| `86f2934` impl+tests | `manager.py`, `theme.js`, `theme_head.html`, `theme_tags.py`, `test_theming_cookie_namespace_1158.py` | impl + 8 tests, **no CHANGELOG** |
| `da363d3` docs+CHANGELOG | `CHANGELOG.md`, `docs/website/guides/components.md` | docs subsection + Unreleased/Added entry, **no impl** |

Verified clean via `gh api /repos/djust-org/djust/commits/<sha>`.

---

### Summary counts

- 🔴 Critical: 0
- 🟡 Important / nits: 4 (empty-cookie migration semantics, no namespace validation, no JSDOM write test, legacy cookie not deleted)
- 🟢 Confirmed correct: 9 (two-commit split, JSON-encoding, all-four-cookies, localStorage skip, empty/None coalesce, build_themes.py skip, two-namespace test, CHANGELOG accuracy, no collateral)

**Recommendation: merge.** None of the 🟡 findings block. Filing follow-up issues for #2 (system check) and #4 (legacy cookie cleanup) would be the cleanest tail.

johnrtipton:
## Pipeline Retro — PR #1168

**Closed**: #1158. **Drain iter**: 6/7.

### Worked well
- **JSON-encoded JS injection** (`json.dumps()` in `theme_tags.py` → `window.__djust_theme_cookie_prefix` in `theme_head.html`) prevented any XSS surface even though the value is settings-controlled. Defense-in-depth held.
- **Two-commit shape** (impl + tests in commit 1, docs + CHANGELOG in commit 2) confirmed clean by Stage 11. The deferred-CHANGELOG cadence keeps holding.
- **Read-fallback for migration** (namespaced first, fall back to legacy unprefixed) means existing deployments don't lose their theme on upgrade.

### Didn't work
- **Two task-brief inaccuracies** (`build_themes.py` was localStorage-only not cookies; `docs/website/guides/theming.md` doesn't exist — theming docs live in `components.md` under the djust-theming heading). Implementer corrected mid-task. Worth a retro note: when prepping pipeline tasks, grep for the exact file paths before writing the brief.

### Follow-ups filed
- **#1169** — 4 🟡s: empty-namespaced-cookie defeats fallback; no validation on namespace value (whitespace/=/; chars); no JSDOM test for the WRITE side; legacy unprefixed cookie persists indefinitely after migration.

### Stage outcomes
1-3 inline / 4-10 implementation agent (single agent, no contention) / 11 APPROVE 0 🔴 / 4 🟡 / 12 deferred via #1169 / 13 PR_MERGED `07905f5d` / 14 (this comment).

### Drain progress
v0.9.1: **6 of 7 PRs shipped** — #1159 #1161 #1163 #1164 #1166 #1168. **Final: #1111** (data_table row navigation).

johnrtipton:
## Retrospective — PR #1168 (backfill)

**Note**: Backfill posted during the v0.9.1 release retro (#1212 audit). PR #1168 shipped in the v0.9.1-1 drain bucket (under the old "v0.9.1" milestone naming) without an individual Stage 14 retro comment. Coverage at the milestone level exists in RETRO.md §"v0.9.1 — v0.9.0 follow-up drain"; this backfill closes the per-PR visibility gap.

**PR title**: feat(theming): namespace cookie names per project (closes #1158)

**Quality assessment** (post-hoc, low confidence): merged clean; no downstream regressions in subsequent v0.9.x PRs. No further action needed beyond this acknowledgment.

🤖 v0.9.1 release retro — backfill closure for v0.9.1-1 stragglers.

Issue #1169, opened 2026-04-27
tech-debt: PR #1168 follow-ups (cookie namespace polish)


johnrtipton:
Source: Stage 11 review of PR #1168 (closes #1158).

## Items

1. **Stale empty namespaced cookie defeats migration fallback.** `_read('djust_theme_<ns>') or None` evaluates the empty string as falsy and falls through to the unprefixed cookie. If the user clears the namespaced cookie value (rather than deleting the cookie), the read returns the legacy value instead of treating it as 'cleared'. Fix: explicit `None` vs empty-string distinction in the cookie read.

2. **No `cookie_namespace` validation.** Whitespace, `=`, or `;` in the namespace value silently produces malformed cookies. Add a check at config-load that rejects characters illegal in cookie names.

3. **Test gap: no JSDOM test for the WRITE side.** The 8 new Python tests cover server reads. The JS-side cookie write logic in `theme.js` is verified only via static-source string assertions, not by exercising a real cookie-write under JSDOM. Add a JSDOM test that asserts `document.cookie` contains the prefixed name after a theme switch.

4. **Legacy unprefixed cookie persists indefinitely.** Read prefers namespaced, falls back to unprefixed. Write only writes namespaced when set. After the first namespaced write, the legacy unprefixed cookie sits in the jar forever (until the user clears their browser). Worth a `document.cookie = legacy_name + '=; expires=Thu, 01 Jan 1970'; ` cleanup on the first namespaced write — or document the limitation explicitly.

## Scope

Single follow-up PR; ~30 LoC core + 1-2 JSDOM tests, ~0.25 day.

## Why deferred from #1168

#1168's primary scope was the namespace mechanism itself. All 4 items are polish on a working implementation. Verdict was APPROVE.

Issue #1167, opened 2026-04-27
tech-debt: PR #1166 follow-ups (test-isolation in tag_registry + sidecar parity for block/assign tags)


johnrtipton:
Source: Stage 11 review of PR #1166 (closes #1145).

## 1. Test-isolation flake — `test_assign_tag.py` after `test_tag_registry.py`

Pre-existing, NOT caused by #1166 (reproduced on main pre-#1166 by Stage 11 reviewer). `test_tag_registry.py:376` registers a `broken` TAG handler whose teardown loop incompletely restores `_registered_handlers`. Subsequent runs of `test_assign_tag.py` see the leaked `broken` handler in the global registry.

This is a different polluter than #1134 (which was django_db / sys.modules). Independent class of test-isolation issue.

**Fix**: tighten the `test_tag_registry.py` teardown — either an autouse fixture that resets `_registered_handlers` to a snapshot taken in setup, or use `monkeypatch` to scope the registration to a single test.

**Severity**: low; only fires under specific test ordering. Runs in isolation are clean.

## 2. Extend sidecar to block-tag and assign-tag handlers (asymmetric plumbing)

PR #1166 added `call_handler_with_py_sidecar` for `Node::CustomTag` only. Block-tag and assign-tag handlers don't yet receive the sidecar — only `Node::CustomTag` was needed for #1145.

**Impact**: any future block-tag or assign-tag handler that needs Python-side sidecar (e.g., `request`, `view`) will hit the same gap #1145 hit and need to extend the bridge again. Trivial mechanical work, but worth doing once for symmetry.

**Scope**: ~30 LoC across `crates/djust_templates/src/registry.rs` and `renderer.rs` (mirror the `Node::CustomTag` pattern for the other two node types).

## Why deferred from #1166

#1166's primary scope was `{% live_render %}` Rust parity (closes #1145). Stage 11 verdict was APPROVE with both items as 🟡 follow-ups. Filing as one tracker so the polish lands together.

Issue #1165, opened 2026-04-27
tech-debt: PR #1164 follow-ups + descriptor-pattern auto-promotion gap


johnrtipton:
Source: Stage 11 review of PR #1164 (closes #1148 + #1149 + #1150) — 3 🟡 follow-ups + 1 framework gap surfaced.

## Items

1. **#1148 — caplog assertions for replay rejection logging.** Tests in `TestReplayHandlerValidation` assert side-effect (counter unchanged) but don't verify the warning is actually logged. If the log call ever silently regresses to a no-op, the side-effect-only tests would still pass. Add `caplog.records` assertions matching the warning message text. ~5 LoC per test.

2. **#1150 — descriptor-pattern auto-promotion gap.** Test `TestDescriptorPatternComponentTimeTravel` defensively pins snapshot/restore behavior, but the framework's `_assign_component_ids` in `python/djust/mixins/components.py:95` only walks `self.__dict__` (instance-level attrs), NOT class-level descriptor declarations. Class-level descriptor components are NOT auto-promoted into `view._components`. The test manually registers them. Two paths:
   - **Document the limitation** — tighten the test docstring to clarify it's a defensive pin not a regression test, and document publicly that descriptor-pattern components require manual `_components` registration.
   - **Implement auto-promotion** — extend `_assign_component_ids` to walk `type(self).__mro__` for `LiveComponent` descriptors and promote them.
   
   The latter is the more user-friendly UX but is a non-trivial framework feature. The former is a one-paragraph doc + docstring tweak.

3. **#1149 — regression test for clean-checkout install.** PR added `markdown` + `nh3` to `[project.optional-dependencies.dev]` but didn't add an automated test that catches the regression class (eg. "after `uv sync --extra dev`, `from djust.components.components import Markdown` succeeds in a clean venv"). The bisect agent in PR #1159 hit the original gap by accident; a CI check would catch any future regression. Could be a `scripts/check-dev-env-imports.py` step in pre-commit or a one-off CI job.

## Scope

Probably 2 separate PRs:
- Item 1 + Item 2's "document the limitation" path: small, fold into one cleanup PR (~30 LoC).
- Item 3: pre-commit hook + script. Separate PR.

If item 2's "implement auto-promotion" path is preferred, that's its own dedicated PR (~80-150 LoC framework feature).

## Why deferred from #1164

[...]

PR #1170, opened 2026-04-27
feat(data_table): row-level navigation accessibility + CSP-strict layer (closes #1111)

### LINKED ISSUE #1111: feat: data_table — row_click_event / row_url for row-level navigation

johnrtipton:
## Summary

Layers v0.9.1 quality additions onto the prior #1111 row-navigation scaffolding (which already shipped `row_click_event` / `row_url` template-tag args, mixin defaults, and structural template wiring):

- **Accessibility**: every row-clickable `<tr>` renders `role="button"`, `tabindex="0"`, and `cursor:pointer` — screen readers announce as button, keyboard users get focus.
- **Keyboard activation**: Enter and Space on a focused row fire the configured action. `document.activeElement === tr` guard so Space inside a nested input doesn't hijack the keystroke.
- **Nested-control guard**: clicks inside `<a>`, `<button>`, `<input>`, `<label>`, `<select>`, `<textarea>` are short-circuited via capture-phase `stopImmediatePropagation` so the row-level action never fires for those clicks. Composes with `selectable=True` (the per-row `<input type="checkbox">`) and the cell-level link column (#1110).
- **CSP-strict friendly**: the row_url path's previous inline `onclick="window.location=this.dataset.href"` is replaced by a new component JS module (`python/djust/components/static/djust_components/data-table-row-click.js`). No inline event handlers, no nonce plumbing — works under `script-src 'self'` out of the box.
- **Defense-in-depth**: `data-href` values are regex-validated against `/^(https?:|\/|\.)/` before `window.location.assign`, so a hostile `javascript:` URI cannot execute.
- **Multi-line template comments fixed**: pre-existing `{# ... #}` row-nav and link-column doc comments rendered as literal text (Django's `{# %}` is single-line-only). Converted to `{% comment %}…{% endcomment %}`.

## Design choice: both options shipped (architectural)

The earlier #1111 PR shipped both `row_click_event` (Option B, LiveView-idiomatic, dj-click on the `<tr>`) and `row_url` (Option A, static-URL navigate via dataset). This PR layers on the missing accessibility / keyboard / CSP / defense-in-depth bits without changing the public API or template-tag signature.

The brief suggested a CSP-nonce-aware inline `<script>` for `row_url`; the chosen architecture (single static JS module, marker class on the `<tr>`) is **strictly more CSP-friendly**: no inline script, no `unsafe-inline` requirement, no nonce plumbing. The same module also handles keyboard activation and nested-control guarding for both options uniformly.

## Behaviour matrix

| Option | Selectable | CSP-strict | Keyboard (Enter/Space) | Nested-control guard |
|--------|-----------|-----------|------------------------|----------------------|
| `row_click_event` | ✓ (checkbox cell unaffected) | ✓ (no inline JS) | ✓ | ✓ |
| `row_url` | ✓ | ✓ (static module replaces inline `onclick`) | ✓ | ✓ |
| Neither (default) | ✓ | n/a (no row-nav) | n/a | n/a |
| Both set | `row_click_event` wins (existing precedence rule) | ✓ | ✓ | ✓ |

## Test count delta

- Python: **+14** new cases in 6 classes (`tests/unit/test_data_table_row_navigation_1111.py`); 1 pre-existing case rewritten in `python/tests/test_data_table_link_row_nav.py` to match the architectural change (no longer asserting inline `onclick`). 28 cases total in the two data_table-row-nav test files now pass.
- JS: **+11** new cases in `tests/js/data_table_row_click.test.js`.
- Net: **+25 test cases**.

| Issue | Files modified | Test classes / cases |
|-------|---------------|----------------------|
| #1111 (this PR) | `python/djust/components/templates/djust_components/table.html` | `TestRowClickAccessibility` (4), `TestRowClickableMarkerClass` (3), `TestRowClickAffordance` (2), `TestCSPInlineHandler` (2), `TestSelectableComposition` (2), `TestCSPNonce` (1) |
| #1111 (this PR) | `python/djust/components/static/djust_components/data-table-row-click.js` (new) | `data-table-row-click — Option A (data-href)` (4), `keyboard activation` (4), `Option B (dj-click composition)` (2), `idempotence` (1) |
| #1111 (compat with prior PR) | `python/tests/test_data_table_link_row_nav.py` | `RowUrlStructureTest::test_row_url_attaches_data_href_and_onclick` rewritten (no longer asserts inline `onclick`) |
| Docs | `docs/website/guides/components.md` | n/a (new section under `{% data_table %}`) |
| Docs | `CHANGELOG.md` | n/a |

## 🟡 Self-review findings for Stage 11

1. **CSP nonce thread is intentionally absent.** The brief asked for `request.csp_nonce` per #1163's pattern, expecting an inline `<script nonce="…">`. The implementation skips inline script entirely, which is strictly more CSP-friendly (no `'unsafe-inline'` ever needed). If a future feature DOES need a per-table inline initializer, the `csp_nonce` context key path is already covered by the smoke test `TestCSPNonce::test_template_renders_with_csp_nonce_in_context`.

2. **Test hook in production JS.** `data-table-row-click.js` honours `window.__djustRowClickNavigate` as a test-only override of `window.location.assign`. JSDOM 26+ marks `Location.prototype.assign` as non-writable + non-configurable, so test interception requires a hook. Documented in the source as test-only; the production code path uses `window.location.assign` directly when the override isn't set.

3. **Multi-line `{# %}` comments fixed but not exhaustively.** The `table.html` template had two pre-existing multi-line `{# ... #}` comments that rendered as literal text. I converted both to `{% comment %}` blocks. Other djust templates may have the same issue; out of scope for this PR.

4. **Nested-control selector includes `<textarea>`.** The brief's enumeration was `<a>/<button>/<input>/<label>/<select>` — I added `<textarea>` for symmetry (a row-nav over a textarea would be just as user-hostile as over an input). Easy to remove if Stage 11 prefers strict adherence.

5. **The pre-existing `dictsort` cell-extraction filter pattern** (`{{ row|dictsort:col.key|first }}`) is not idiomatic Django but is a pre-existing convention from #1110 — the Rust template engine apparently supports it. Out of scope to refactor here; would risk breaking #1110 callers.

## Test plan

- [ ] `make test-python` — full Python suite (4026 passing pre-PR, expect 4040 after this PR's 14 new cases).
- [ ] `make test-js` — full JS suite (1450 cases pre-PR, expect 1461 after this PR's 11 new cases).
- [ ] Manual: render the demo data_table with `row_click_event="open"`; click the row and verify the dj-click fires with `data-value=row_id`. Click a nested `<a>` cell — verify the row event does NOT also fire. Tab to the row, press Enter — verify the click handler fires.
- [ ] Manual: render with `row_url="some_url"`; click the row and verify the page navigates. With strict CSP (`script-src 'self'`), verify there's no console error about blocked inline script.
- [ ] Manual: render with `selectable=True row_click_event="open"`; click the checkbox — verify select_event fires but row event does not. Click anywhere else in the row — verify row event fires.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,52 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- **`{% data_table %}` row-level navigation: accessibility, keyboard,
+  and CSP-strict layer (closes #1111)** — layers v0.9.1 quality
+  additions onto the prior #1111 row-navigation scaffolding (which
+  shipped `row_click_event` / `row_url` template-tag args, mixin
+  defaults, and structural wiring). What's added:
+  - **Accessibility**: every row-clickable `<tr>` now renders
+    `role="button"`, `tabindex="0"`, and `cursor:pointer`. Screen
+    readers announce the row as a button; keyboard users get focus.
+  - **Keyboard activation**: Enter and Space on a focused row fire
+    the configured action. Guarded by `document.activeElement === tr`
+    so Space inside a nested input doesn't hijack the keystroke.
+  - **Nested-control guard**: clicks inside `<a>`, `<button>`,
+    `<input>`, `<label>`, `<select>`, `<textarea>` are short-circuited
+    via capture-phase `stopImmediatePropagation`, so the row-level
+    action never fires for those clicks. This is the integration
+    point with `selectable=True` (per-row checkbox) and the
+    cell-level link column (#1110).
+  - **CSP-strict friendly**: the row_url path's previous inline
+    `onclick="window.location=this.dataset.href"` is replaced by a
+    new component JS module
+    (`python/djust/components/static/djust_components/data-table-row-click.js`).
+    No inline event handlers, no nonce plumbing — works under
+    `script-src 'self'` out of the box.
+  - **Defense-in-depth**: `data-href` values are regex-validated
+    against `/^(https?:|\/|\.)/` before `window.location.assign`,
+    so a hostile `javascript:` URI cannot execute even if it sneaks
+    into the row dict.
+  - **Multi-line template comments fixed**: the pre-existing
+    `{# ... #}` row-nav and link-column doc comments were rendering
+    as literal text in output because Django's `{# %}` is
+    single-line-only. Converted to `{% comment %}...{% endcomment %}`.
+
+  New cases in `TestRowClickAccessibility`,
+  `TestRowClickableMarkerClass`, `TestRowClickAffordance`,
+  `TestCSPInlineHandler`, `TestSelectableComposition`, `TestCSPNonce`
+  (`tests/unit/test_data_table_row_navigation_1111.py`, 14 Python
+  cases) plus 11 JS cases in `tests/js/data_table_row_click.test.js`
+  cover: role + tabindex presence, marker class on/off, no-inline-
+  onclick (CSP), checkbox cell composition, click navigation, nested
+  `<a>`/`<input>` guard, Enter/Space activation, `activeElement`
+  guard, javascript: URI rejection, dj-click composition (capture-phase
+  stop), and bindRow idempotence. One pre-existing structural test
+  in `python/tests/test_data_table_link_row_nav.py` was rewritten to
+  assert the new `data-table-row-clickable` marker class instead of
+  the removed inline `onclick`.
+
 - **Theming cookie namespace for per-project isolation on shared
   domains (closes #1158)** — adds opt-in
   `LIVEVIEW_CONFIG['theme']['cookie_namespace']` setting so multiple
--- a/docs/website/guides/components.md
+++ b/docs/website/guides/components.md
@@ -680,6 +680,95 @@ INSTALLED_APPS = [
 | `{% pagination %}`                         | Page navigation                      |
 | `{% avatar %}`                             | User avatar with initials fallback   |
 
+### `{% data_table %}` row-level navigation (#1111)
+
+Two ways to make whole rows clickable for navigation. Both render
+`role="button"`, `tabindex="0"`, and `cursor:pointer` on every row, and
+both respect Enter / Space for keyboard activation. Clicks inside
+nested interactive descendants (`<a>`, `<button>`, `<input>`,
+`<label>`, `<select>`, `<textarea>`) are short-circuited so they
+**never** trigger row navigation.
+
+**Option B — `row_click_event` (preferred, LiveView-idiomatic)**: each
+`<tr>` fires a djust event with `data-value=row[row_click_value_key]`.
+Your `@event_handler` receives the row's value via `**kwargs` and can
+`self.redirect(...)` (or do anything else):
+
+```python
+from djust import LiveView
+from djust.decorators import event_handler
+from django.urls import reverse
+
+class ClaimsListView(LiveView):
+    template_name = "claims/list.html"
+
+    def mount(self, request, **kwargs):
+        self.rows = list(Claim.objects.values("id", "claim_number", "status"))
+        self.columns = [
+            {"key": "claim_number", "label": "Claim #"},
+            {"key": "status", "label": "Status"},
+        ]
+
+    @event_handler()
+    def open_claim(self, value: str = "", **kwargs):
+        self.redirect(reverse("claims:detail", kwargs={"claim_id": value}))
+```
+
+```django
+{% data_table rows=rows columns=columns
+              row_click_event="open_claim"
+              row_click_value_key="id" %}
+```
+
+`row_click_value_key` defaults to `"id"`; override for slug-based
+routing (e.g. `"uuid"`, `"slug"`).
+
+**Option A — `row_url` (static-URL fallback)**: each `<tr>` gets
+`data-href` from a row dict key; the component JS module navigates via
+`window.location.assign(href)` on click / Enter / Space. Use when the
+target URL is computed once in the view and doesn't need a server
+round-trip:
+
+```python
+def get_context_data(self, **kwargs):
+    ctx = super().get_context_data(**kwargs)
+    ctx["rows"] = [
+        {"claim_number": "C-001", "claim_url": reverse("claims:detail", args=[1])},
+    ]
+    ctx["columns"] = [{"key": "claim_number", "label": "Claim #"}]
+    return ctx
+```
+
+```django
+{% data_table rows=rows columns=columns row_url="claim_url" %}
+```
+
+`row_click_event` takes precedence when both are set.
+
+**Accessibility**: row-clickable `<tr>`s are focusable (`tabindex="0"`),
+announce as buttons (`role="button"`), and activate on Enter or Space.
+Screen reader users get the same affordance as mouse / touch users.
+
+**CSP-strict friendly**: no inline event handlers, no `eval`, no nonce
+plumbing required. Both options work under
+`Content-Security-Policy: script-src 'self'` once the component JS
+module is served as a static asset (no `'unsafe-inline'`).
+
+**Composition with `selectable=True`**: per-row checkbox cells are
+`<input>` elements, which are in the nested-control guard's protected
+list. Clicking the checkbox fires the existing `select_event` and does
+**not** also fire row navigation.
+
+**Composition with cell-level link columns (#1110)**: cell `<a>` tags
+also "win" — clicking a cell link follows the link without firing the
+row event.
+
+**Static-URL safety**: Option A's `data-href` value is regex-validated
+(`/^(https?:|\/|\.)/`) before navigation, so a hostile `javascript:`
+URI cannot execute even if it sneaks into the row dict. Always prefer
+URLs computed from `reverse()` in your view; never assign
+user-controlled strings to `row_url` data.
+
 ### Customization
 
 All components use CSS custom properties. Override them to match any theme:
--- a/python/djust/components/static/djust_components/data-table-row-click.js
+++ b/python/djust/components/static/djust_components/data-table-row-click.js
@@ -50,11 +50,14 @@
     // Static-URL path: navigate via dataset.href.
     var href = tr.dataset && tr.dataset.href;
     if (href) {
-      // Only allow http(s) and relative-path URLs to defend against a
-      // hostile data-href value sneaking in (e.g. javascript: URIs).
+      // Only allow http(s) and SAME-ORIGIN relative-path URLs to defend
+      // against a hostile data-href value sneaking in (e.g. javascript:
+      // URIs or protocol-relative `//evil.com/...` cross-origin redirects).
       // Developer-controlled values from `reverse()` are always either
       // absolute or relative paths, never `javascript:` schemes.
-      if (/^(https?:|\/|\.)/.test(href)) {
+      // The `(?!\/)` lookahead on the leading `/` rejects `//host` while
+      // still allowing single-leading-slash absolute paths.
+      if (/^(https?:\/\/|\/(?!\/)|\.)/.test(href)) {
         // Tests override window.__djustRowClickNavigate to capture the
         // target URL since JSDOM's window.location.assign is
         // non-configurable. Production code path is the default arm.
--- a/python/djust/components/static/djust_components/data-table-row-click.js
+++ b/python/djust/components/static/djust_components/data-table-row-click.js
@@ -1,0 +1,161 @@
+/**
+ * djust-components Data Table — Row-level navigation handler (#1111).
+ *
+ * Adds keyboard accessibility (Enter / Space activate a focused row) and
+ * a nested-control guard (clicks inside an interactive descendant
+ * <a>/<button>/<input>/<label>/<select> do NOT trigger row navigation)
+ * to data_table rows that opt in via the `data-table-row-clickable`
+ * marker class.
+ *
+ * The template tag emits one of two row shapes when row navigation is
+ * enabled:
+ *   1. `dj-click="<event>" data-value="<row_key>"` (Option B, preferred):
+ *      djust's existing event binder fires the LiveView event on
+ *      bubble-phase click. This module's job is purely defensive —
+ *      cancel the click in the capture phase when the target is a
+ *      nested control, so the LiveView event handler never fires for
+ *      those clicks.
+ *   2. `data-href="<url>"` (Option A, static URL): no `dj-click`, so
+ *      this module is responsible for the navigation. On click (and on
+ *      Enter / Space keydown when focused), navigate via
+ *      `window.location.assign(tr.dataset.href)`. Same nested-control
+ *      guard applies.
+ *
+ * Design notes:
+ *   - CSP-strict friendly: no inline event handlers, no eval, no nonce
+ *     plumbing required. The whole feature works under
+ *     `script-src 'self'` once this file is served as a static asset.
+ *   - Composition with selectable=True: the per-row checkbox is an
+ *     <input>, so the nested-control guard automatically suppresses row
+ *     navigation when clicking it. No special-case code required.
+ *   - Composition with the cell-level link column (#1110): the cell
+ *     <a> is also caught by the nested-control guard, so cell links
+ *     "win" over row navigation — clicking the link follows the link
+ *     and does NOT also fire the row event.
+ *   - Auto-init on DOMContentLoaded; re-runs on djust LiveView VDOM
+ *     patches via a MutationObserver, idempotent per-row via a
+ *     `_dtRowClick` instance flag so repeated patches don't double-bind.
+ */
+(function () {
+  "use strict";
+
+  // Selectors for descendants whose clicks should NOT propagate up to
+  // row-level navigation. Matches the Stage 4 brief enumeration plus
+  // textarea (a non-button form control whose accidental row-nav would
+  // be just as user-hostile as the others).
+  var NESTED_CONTROL_SELECTOR =
+    "a, button, input, label, select, textarea";
+
+  function navigateForRow(tr) {
+    // Static-URL path: navigate via dataset.href.
+    var href = tr.dataset && tr.dataset.href;
+    if (href) {
+      // Only allow http(s) and relative-path URLs to defend against a
+      // hostile data-href value sneaking in (e.g. javascript: URIs).
+      // Developer-controlled values from `reverse()` are always either
+      // absolute or relative paths, never `javascript:` schemes.
+      if (/^(https?:|\/|\.)/.test(href)) {
+        // Tests override window.__djustRowClickNavigate to capture the
+        // target URL since JSDOM's window.location.assign is
+        // non-configurable. Production code path is the default arm.
+        var navigate =
+          (typeof window !== "undefined" &&
+            window.__djustRowClickNavigate) ||
+          function (h) {
+            window.location.assign(h);
+          };
+        navigate(href);
+      }
+      return true;
+    }
+    return false;
+  }
+
+  function bindRow(tr) {
+    if (tr._dtRowClick) return;
+    tr._dtRowClick = true;
+
+    // Capture-phase click on the row: if the click originated inside a
+    // nested control, stopImmediatePropagation prevents the bubble-phase
+    // dj-click handler from firing, AND prevents this module's own
+    // click->navigate logic below.
+    tr.addEventListener(
+      "click",
+      function (e) {
+        if (e.target !== tr && e.target.closest) {
+          var nested = e.target.closest(NESTED_CONTROL_SELECTOR);
+          if (nested && tr.contains(nested)) {
+            // The nested control gets to handle the click natively.
+            // Suppress row-level navigation entirely.
+            e.stopImmediatePropagation();
+            return;
+          }
+        }
+
+        // Static-URL navigation. (For dj-click rows, this is a no-op
+        // because dataset.href isn't set.)
+        navigateForRow(tr);
+      },
+      true /* capture */
+    );
+
+    // Keyboard activation: Enter and Space when the row is focused
+    // synthesise a click. The click then re-enters the handler above
+    // with `e.target === tr`, so the nested-control guard doesn't trip
+    // and the navigation (or dj-click) fires normally.
+    tr.addEventListener("keydown", function (e) {
+      if (e.key === "Enter" || e.key === " " || e.key === "Spacebar") {
+        // Only fire when the row itself is the active element — don't
+        // hijack Space when the user is typing into a nested input.
+        if (document.activeElement !== tr) return;
+        e.preventDefault();
+        tr.click();
+      }
+    });
+  }
+
+  function initWrapper(wrapper) {
+    var rows = wrapper.querySelectorAll("tr.data-table-row-clickable");
+    rows.forEach(bindRow);
+  }
+
+  function initAll() {
+    document.querySelectorAll(".data-table-wrapper").forEach(initWrapper);
+    // Also support tables that aren't wrapped (defensive).
+    document
+      .querySelectorAll("tr.data-table-row-clickable")
+      .forEach(bindRow);
+  }
+
+  if (document.readyState === "loading") {
+    document.addEventListener("DOMContentLoaded", initAll);
+  } else {
+    initAll();
+  }
+
+  // Re-init on djust LiveView patches that insert new rows.
+  if (typeof MutationObserver !== "undefined") {
+    var observer = new MutationObserver(function (mutations) {
+      var shouldInit = false;
+      for (var i = 0; i < mutations.length; i++) {
+        if (mutations[i].addedNodes.length) {
+          shouldInit = true;
+          break;
+        }
+      }
+      if (shouldInit) initAll();
+    });
+    if (document.body) {
+      observer.observe(document.body, { childList: true, subtree: true });
+    }
+  }
+
+  // Expose for tests and for explicit re-init from app code.
+  if (typeof window !== "undefined") {
+    window.djustDataTableRowClick = {
+      initAll: initAll,
+      initWrapper: initWrapper,
+      bindRow: bindRow,
+    };
+  }
+})();
--- a/python/djust/components/templates/djust_components/table.html
+++ b/python/djust/components/templates/djust_components/table.html
@@ -97,4 +97,21 @@
   </div>
   {% endif %}
   {% endif %}
+
+  {% if row_click_event or row_url %}
+  {% comment %}
+    #1111: auto-load the row-click handler when row navigation is enabled.
+
+    Hardcoded /static/ prefix (default STATIC_URL). Works for the default
+    Django configuration without requiring `{% load static %}` (which a
+    bare `Engine()` test renderer can't resolve). Sites with a custom
+    STATIC_URL can override `djust_components/table.html` via Django's
+    standard template-customization mechanism — drop a copy in their
+    own templates dir with their `{% static %}` resolution.
+
+    Defer attribute → loads after parsing; the JS module is idempotent if
+    multiple data_tables on one page each emit this script tag.
+  {% endcomment %}
+  <script src="/static/djust_components/data-table-row-click.js" defer></script>
+  {% endif %}
 </div>
--- a/python/djust/components/templates/djust_components/table.html
+++ b/python/djust/components/templates/djust_components/table.html
@@ -39,10 +39,17 @@
     </thead>
     <tbody>
       {% for row in rows %}
-      {# #1111: row-level navigation. row_click_event fires a djust event
-         (LiveView-idiomatic, preferred); row_url is a static fallback that
-         uses dataset+JS to navigate without LiveView routing. #}
-      <tr{% if selectable %} aria-selected="{% if row.id|stringformat:'s' in selected_rows %}true{% else %}false{% endif %}"{% endif %}{% if row_click_event %} dj-click="{{ row_click_event }}" data-value="{{ row|dictsort:row_click_value_key|first }}" style="cursor:pointer"{% elif row_url %} data-href="{{ row|dictsort:row_url|first }}" style="cursor:pointer" onclick="window.location=this.dataset.href"{% endif %}>
+      {% comment %}
+        #1111: row-level navigation. row_click_event fires a djust event
+        (LiveView-idiomatic, preferred); row_url is a static fallback
+        that uses dataset+JS to navigate without LiveView routing. The
+        data-table-row-clickable marker class wires up keyboard
+        activation (Enter / Space) and a nested-control guard (clicks
+        inside <a>/<button>/<input>/<label>/<select> do NOT trigger row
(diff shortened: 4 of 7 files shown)


johnrtipton (commented):
# Stage 11 Review — PR #1170

`feat(data_table): row-level navigation accessibility + nested-control guard (closes #1111)`

Branch: `feat/1111-data-table-row-navigation` → `main`. Two commits, +898 / −9, 7 files. Mergeable.

## Verdict: REQUEST CHANGES

The architectural choice (single static JS module, no inline scripts) is genuinely cleaner than the brief's nonce-thread approach. Tests are dense and exercise distinct paths. Two-commit shape is clean. CHANGELOG matches the diff. **However, three issues need attention before merge** — one is a likely correctness bug (open-redirect via protocol-relative URL), one is a usability-cliff (the JS module is not auto-loaded and the docs never tell users to include it), and one is a test-coverage gap on the protocol-relative case.

---

## Findings

### 🔴 R1 — `data-href` URL allow-list permits protocol-relative URLs (open-redirect risk)

`python/djust/components/static/djust_components/data-table-row-click.js:57`:

```javascript
if (/^(https?:|\/|\.)/.test(href)) {
```

The character class `\/` matches a single leading slash, which means `//evil.com/path` (a protocol-relative URL) **passes the validator**. `window.location.assign("//evil.com/path")` performs a full cross-origin navigation, identical to `https://evil.com/path` on an https origin. Empirical probe:

```
"//evil.com/path"                        true   ← validator says OK
"javascript:alert(1)"                    false
```

This is the standard open-redirect / phishing primitive. If a developer assigns a row dict value like `row["url"]` from a partly-trusted source (Markdown link, user-supplied profile URL, third-party feed) to `row_url=...`, they now have a one-click redirect-to-anywhere on a row. The defense-in-depth claim in the PR body and CHANGELOG is "a hostile `javascript:` URI cannot execute" — true, but the threat surface here is broader. The fix is one character:

```javascript
if (/^(https?:\/\/|\/[^\/]|\.)/.test(href)) {  // require a non-slash after the leading slash
```

or equivalently `^(https?:\/\/|\/(?!\/)|\.)`. **Add a JS test case asserting `//evil.com/...` is rejected** alongside the existing `javascript:` test.

Adjacent gaps in the same regex (lower priority but worth a note in the same fix commit):
- Leading whitespace bypass via `getAttribute()`-style values is already handled by the regex (`" /foo"` fails, good).
- `data:` URIs are rejected (good).
- `tel:` / `mailto:` are rejected — probably desired for a row-nav handler, but worth a one-liner comment in the source so a future maintainer doesn't "fix" the regex by widening it.

### 🔴 R2 — JS module is never loaded automatically; docs don't tell the user to load it

`grep -rn "data-table-row-click" python/djust/components/templates/` shows the marker-class string in `table.html` but **no `<script src=...>` for the module file anywhere in the framework**. There's also no `{% djust_component_assets %}`-style aggregator template tag. By comparison, the older `data-table.js` (the resize/reorder pro-features module) has the same shape and presumably the same gap, but at minimum it had pre-existing onboarding docs. The new module is brand-new and silent.

`docs/website/guides/components.md:752`–`755` says only:

> **CSP-strict friendly**: no inline event handlers, no `eval`, no nonce
> plumbing required. Both options work under
> `Content-Security-Policy: script-src 'self'` once the component JS
> module is served as a static asset (no `'unsafe-inline'`).

"once the component JS module is served as a static asset" is the active voice of an instruction the user will not perform. There is no `<script src="{% static 'djust_components/data-table-row-click.js' %}"></script>` snippet anywhere in the docs.

User-visible failure mode: a developer adds `row_click_event=` or `row_url=`, sees `cursor:pointer` and `role="button"` on rows, clicks one, and **nothing happens**. There's no console error (the module simply was never executed). They have to grep the framework source to discover the module exists.

The architecture is right. The onboarding is missing. Pick one of:

1. **Inject the script tag from `table.html`** (one line: `<script src="{% static 'djust_components/data-table-row-click.js' %}" defer></script>` outside the `data-table-wrapper`, guarded by `{% if row_click_event or row_url %}`). Simplest fix, no breaking changes for users.
2. **Add the explicit `<script>` snippet to the docs section in `components.md`** with a "Required: include this once on any page using `row_click_event` or `row_url`" callout.
3. Both (defense-in-depth doc + working out-of-the-box default).

Without one of these, the feature ships dead-on-arrival for the typical user, and the Stage 11 tests pass because they exercise the JS module directly via JSDOM eval — not via the actual Django render→browser path.

### 🟡 R3 — Nested-control selector is incomplete (`<details>`/`<summary>`/`<option>` missing)

`data-table-row-click.js:46–47`:

```javascript
var NESTED_CONTROL_SELECTOR =
  "a, button, input, label, select, textarea";
```

Three interactive elements that produce a click event but should NOT trigger row-nav are absent:

- `<summary>` — the only interactive child of `<details>`. A user expanding a `<details>` inside a row by clicking the summary will currently fire row navigation as well. This is a real composition concern if any data_table renders an expander cell.
- `<option>` — clicks bubble through the `<select>`, so this is *probably* covered transitively by `<select>` itself, but tested-in-isolation `<option>` clicks (some test runners synthesize them) won't be caught.
- `<details>` — clicking the open/close toggle area outside `<summary>` is non-standard but exists in some themes.

Recommended:

```javascript
"a, button, input, label, select, textarea, summary, details, option";
```

Mention in the PR / docs that this is a deliberate enumeration so future contributors know what the contract is, with a one-line test per added element.

The implementer's `<textarea>` addition (over the brief's 5-element enumeration) is fine — the source comment at `data-table-row-click.js:43–45` notes it explicitly. But the comment says "plus textarea" while the same selector also predates the brief in shape; consider updating to "plus textarea, summary, option, details (interactive descendants whose clicks should not bubble to row-level nav)".

### 🟡 R4 — `window.__djustRowClickNavigate` test hook in production JS

`data-table-row-click.js:61–67`:

```javascript
var navigate =
  (typeof window !== "undefined" &&
    window.__djustRowClickNavigate) ||
  function (h) {
    window.location.assign(h);
  };
navigate(href);
```

The implementer's reasoning (JSDOM 26+ marks `Location.prototype.assign` as non-configurable, so `vi.spyOn(window.location, 'assign')` throws) is empirically true and was the reason `mock-location` and `jsdom-extended` keep getting bandaged in other test suites. The hook is real-world justified. Still, two improvements would tighten this:

1. **Spy on the exported namespace instead of needing the hook.** The module already exports `window.djustDataTableRowClick = { initAll, initWrapper, bindRow }`. Adding `navigate` to that export and having `bindRow` call `module.navigate(href)` (via a closure-captured object reference, not `window.djustDataTableRowClick.navigate` — to allow vi.spyOn to redirect calls) lets tests do `vi.spyOn(dom.window.djustDataTableRowClick, 'navigate')` instead of relying on a magic underscored global. This is the standard pattern for testing modules that wrap browser globals.
2. **If keeping the hook, rename and document defensively.** `window.__djustRowClickNavigate` reads like a debug shim. Inline source comment is good (lines 58–60); add a top-of-file note in the doc-comment that says "this hook exists solely because JSDOM's `Location.assign` is non-configurable; do not rely on it from app code, it may be removed if/when a cleaner test path appears."

Not a blocker on its own, but combined with R3 it's worth fixing in this PR rather than papering over.

### 🟡 R5 — Test coverage gap: protocol-relative URL not tested

Tied to R1. `tests/js/data_table_row_click.test.js:127–135` exercises `javascript:alert(1)` rejection but does not cover `//evil.com/...`. Add a sibling case:

```javascript
it("rejects protocol-relative URL in data-href (open-redirect defense)", () => {
  const dom = buildDom(clickableRowMarkup({ href: "//evil.com/x" }));
  dom.window.document.querySelector("tr.data-table-row-clickable").click();
  expect(dom.window.__locationAssignSpy).not.toHaveBeenCalled();
});
```

Also consider adding cases for `data:`, `vbscript:`, and bare-fragment `#x` rejection (`#x` is currently rejected; locking it in prevents a future regex widening from accidentally allowing self-page DoS via empty navigations).

### 🟢 G1 — Multi-line `{# %}` fix verified correct

I confirmed against the pre-PR template at commit `859d8374` that both blocks (line 41–43 row-nav comment, line 53–56 link-column comment) used `{# ... #}` across multiple lines. Django's docs are unambiguous: `{# %}` is single-line-only and a multi-line use renders the inner content as raw text. The conversion to `{% comment %}…{% endcomment %}` is correct and self-contained. The implementer's "out of scope" callout for other templates with the same issue is appropriate.

### 🟢 G2 — Two-commit shape is clean

`4b765b9e` touches only: `data-table-row-click.js` (new), `table.html`, `test_data_table_link_row_nav.py`, `data_table_row_click.test.js` (new), `test_data_table_row_navigation_1111.py` (new). No CHANGELOG, no docs.
`e7368381` touches only: `CHANGELOG.md`, `docs/website/guides/components.md`. No code.

Clean separation. No collateral noise.

### 🟢 G3 — CHANGELOG matches the diff

CHANGELOG bullets accurately describe what shipped. Class-name form (`TestRowClickAccessibility`, `TestRowClickableMarkerClass`, …) — good for future audits per `feedback_changelog_required` and the `Bulk dispatch-site refactor + count-test pattern` (#1125) canonicalization. Test count math (14 + 11 = 25) is correct.

### 🟢 G4 — Test coverage is dense and distinct

I read four tests (`test_role_and_tabindex_render_per_row`, `test_selectable_does_not_double_dj_click_on_checkboxes`, the JS `Space inside a nested input does NOT hijack the keystroke` case, the JS `nested <a> click stops propagation so dj-click does not fire` case). They exercise meaningfully different code paths: per-row count assertion, dj-click count discrimination across selectable composition, `activeElement` guard semantics, and capture-phase propagation stop. Not near-duplicates.

The Python test file uses Django's `Engine().from_string()` rather than the production Rust template engine. This matches the convention of `test_data_table_link_row_nav.py` and is fine for the structural assertions made (string-presence in rendered output) since the Rust engine is contractually output-equivalent to Django's for the syntactic constructs in `table.html`.

### 🟢 G5 — Capture-phase `stopImmediatePropagation` semantics for `<a>` are correct

Concern from the brief: does the capture-phase stop block `<a>`'s default navigation? It does not. `stopImmediatePropagation()` halts **propagation to other event listeners** (capture and bubble phase) but does NOT call `preventDefault()`. The `<a>`'s built-in default-action handling runs after all listeners and is unaffected. Test `nested <a> click stops propagation so dj-click does not fire` (lines 218–235) exercises the dj-click suppression; the `<a>`-follows-link case is implicit (no `e.preventDefault()` is called anywhere in the handler).

### 🟢 G6 — `closest()` semantics for the selector are correct

`e.target.closest(NESTED_CONTROL_SELECTOR)` traverses *up* from the click target (inclusive — checks the target first). The `tr.contains(nested)` check then rejects matches where the `closest()` result somehow lives outside this row. Combined with the `e.target !== tr` guard at line 85, the row itself is correctly excluded from the nested-control match. Reviewer-question on this point: passes.

---

## Question

Did the implementer consider option (1) from R2 — auto-injecting the `<script>` from `table.html` when `row_click_event` or `row_url` is set? It would be one line of template, eliminate the "module must be served" caveat from the docs, and make the feature work out-of-the-box for any page using `{% data_table %}`. If there's a reason the framework deliberately doesn't auto-inject component scripts (e.g. the prior `data-table.js` resize module follows the same load-yourself convention by design), call that out — otherwise this seems like the right place to break that pattern.

---

## Self-flag responses

1. **CSP nonce thread absent** (🟡-by-implementer) — confirmed: the chosen architecture is strictly more CSP-friendly. No issue.
2. **Test hook in production JS** (🟡-by-implementer) — see R4. Real-world justified, but a `vi.spyOn`-able exported function would be cleaner. Worth fixing in this PR.
3. **Multi-line `{# %}` fix** (🟡-by-implementer) — verified against the pre-PR template; the leak was real. Out-of-scope deferral for other templates is appropriate.
4. **`<textarea>` added for symmetry** (🟡-by-implementer) — fine, but R3 surfaces three more elements that should also be in the selector.
5. **`dictsort` cell-extraction pattern** (🟡-by-implementer) — pre-existing from #1110. Out of scope. No issue.

---

## Two-commit integrity: PASS
## Test-hook in production code: ACCEPTABLE-WITH-CHANGES (R4 — prefer exported-namespace spy over magic global; not a blocker if rejected with reasoning)
## Open-redirect risk in URL allow-list: MUST FIX (R1)
## Auto-load gap: MUST FIX (R2)

## Summary

| Severity | Count | Items |
|---|---|---|
| 🔴 Critical | 2 | R1 (open-redirect via protocol-relative URL), R2 (JS module never auto-loaded, docs silent) |
| 🟡 Important | 3 | R3 (selector missing `summary`/`option`/`details`), R4 (test hook → exported spy), R5 (test gap on protocol-relative case) |
| 🟢 Pass | 6 | G1 (comment fix), G2 (commit shape), G3 (CHANGELOG), G4 (test density), G5 (`<a>` default action), G6 (`closest()` semantics) |

johnrtipton:
# Stage 11 Review — PR #1170

`feat(data_table): row-level navigation accessibility + nested-control guard (closes #1111)`

Branch: `feat/1111-data-table-row-navigation` → `main`. Two commits, +898 / −9, 7 files. Mergeable.

## Verdict: REQUEST CHANGES

The architectural choice (single static JS module, no inline scripts) is genuinely cleaner than the brief's nonce-thread approach. Tests are dense and exercise distinct paths. Two-commit shape is clean. CHANGELOG matches the diff. **However, three issues need attention before merge** — one is a likely correctness bug (open-redirect via protocol-relative URL), one is a usability-cliff (the JS module is not auto-loaded and the docs never tell users to include it), and one is a test-coverage gap on the protocol-relative case.

---

## Findings

### 🔴 R1 — `data-href` URL allow-list permits protocol-relative URLs (open-redirect risk)

`python/djust/components/static/djust_components/data-table-row-click.js:57`:

```javascript
if (/^(https?:|\/|\.)/.test(href)) {
```

The character class `\/` matches a single leading slash, which means `//evil.com/path` (a protocol-relative URL) **passes the validator**. `window.location.assign("//evil.com/path")` performs a full cross-origin navigation, identical to `https://evil.com/path` on an https origin. Empirical probe:

```
"//evil.com/path"                        true   ← validator says OK
"javascript:alert(1)"                    false
```

This is the standard open-redirect / phishing primitive. If a developer assigns a row dict value like `row["url"]` from a partly-trusted source (Markdown link, user-supplied profile URL, third-party feed) to `row_url=...`, they now have a one-click redirect-to-anywhere on a row. The defense-in-depth claim in the PR body and CHANGELOG is "a hostile `javascript:` URI cannot execute" — true, but the threat surface here is broader. The fix is one character:

```javascript
if (/^(https?:\/\/|\/[^\/]|\.)/.test(href)) {  // require a non-slash after the leading slash
```

or equivalently `^(https?:\/\/|\/(?!\/)|\.)`. **Add a JS test case asserting `//evil.com/...` is rejected** alongside the existing `javascript:` test.

Adjacent gaps in the same regex (lower priority but worth a note in the same fix commit):
- Leading whitespace bypass via `getAttribute()`-style values is already handled by the regex (`" /foo"` fails, good).
- `data:` URIs are rejected (good).
- `tel:` / `mailto:` are rejected — probably desired for a row-nav handler, but worth a one-liner comment in the source so a future maintainer doesn't "fix" the regex by widening it.

### 🔴 R2 — JS module is never loaded automatically; docs don't tell the user to load it

`grep -rn "data-table-row-click" python/djust/components/templates/` shows the marker-class string in `table.html` but **no `<script src=...>` for the module file anywhere in the framework**. There's also no `{% djust_component_assets %}`-style aggregator template tag. By comparison, the older `data-table.js` (the resize/reorder pro-features module) has the same shape and presumably the same gap, but at minimum it had pre-existing onboarding docs. The new module is brand-new and silent.

`docs/website/guides/components.md:752`–`755` says only:

> **CSP-strict friendly**: no inline event handlers, no `eval`, no nonce
> plumbing required. Both options work under
> `Content-Security-Policy: script-src 'self'` once the component JS
> module is served as a static asset (no `'unsafe-inline'`).

"once the component JS module is served as a static asset" is the active voice of an instruction the user will not perform. There is no `<script src="{% static 'djust_components/data-table-row-click.js' %}"></script>` snippet anywhere in the docs.

User-visible failure mode: a developer adds `row_click_event=` or `row_url=`, sees `cursor:pointer` and `role="button"` on rows, clicks one, and **nothing happens**. There's no console error (the module simply was never executed). They have to grep the framework source to discover the module exists.

The architecture is right. The onboarding is missing. Pick one of:

1. **Inject the script tag from `table.html`** (one line: `<script src="{% static 'djust_components/data-table-row-click.js' %}" defer></script>` outside the `data-table-wrapper`, guarded by `{% if row_click_event or row_url %}`). Simplest fix, no breaking changes for users.
2. **Add the explicit `<script>` snippet to the docs section in `components.md`** with a "Required: include this once on any page using `row_click_event` or `row_url`" callout.
3. Both (defense-in-depth doc + working out-of-the-box default).

Without one of these, the feature ships dead-on-arrival for the typical user, and the Stage 11 tests pass because they exercise the JS module directly via JSDOM eval — not via the actual Django render→browser path.

### 🟡 R3 — Nested-control selector is incomplete (`<details>`/`<summary>`/`<option>` missing)

`data-table-row-click.js:46–47`:

```javascript
var NESTED_CONTROL_SELECTOR =
  "a, button, input, label, select, textarea";
```

Three interactive elements that produce a click event but should NOT trigger row-nav are absent:

- `<summary>` — the only interactive child of `<details>`. A user expanding a `<details>` inside a row by clicking the summary will currently fire row navigation as well. This is a real composition concern if any data_table renders an expander cell.
- `<option>` — clicks bubble through the `<select>`, so this is *probably* covered transitively by `<select>` itself, but tested-in-isolation `<option>` clicks (some test runners synthesize them) won't be caught.
- `<details>` — clicking the open/close toggle area outside `<summary>` is non-standard but exists in some themes.

Recommended:

```javascript
"a, button, input, label, select, textarea, summary, details, option";
```

Mention in the PR / docs that this is a deliberate enumeration so future contributors know what the contract is, with a one-line test per added element.

The implementer's `<textarea>` addition (over the brief's 5-element enumeration) is fine — the source comment at `data-table-row-click.js:43–45` notes it explicitly. But the comment says "plus textarea" while the same selector also predates the brief in shape; consider updating to "plus textarea, summary, option, details (interactive descendants whose clicks should not bubble to row-level nav)".

### 🟡 R4 — `window.__djustRowClickNavigate` test hook in production JS

`data-table-row-click.js:61–67`:

```javascript
var navigate =
  (typeof window !== "undefined" &&
    window.__djustRowClickNavigate) ||
  function (h) {
    window.location.assign(h);
  };
navigate(href);
```

The implementer's reasoning (JSDOM 26+ marks `Location.prototype.assign` as non-configurable, so `vi.spyOn(window.location, 'assign')` throws) is empirically true and was the reason `mock-location` and `jsdom-extended` keep getting bandaged in other test suites. The hook is real-world justified. Still, two improvements would tighten this:

1. **Spy on the exported namespace instead of needing the hook.** The module already exports `window.djustDataTableRowClick = { initAll, initWrapper, bindRow }`. Adding `navigate` to that export and having `bindRow` call `module.navigate(href)` (via a closure-captured object reference, not `window.djustDataTableRowClick.navigate` — to allow vi.spyOn to redirect calls) lets tests do `vi.spyOn(dom.window.djustDataTableRowClick, 'navigate')` instead of relying on a magic underscored global. This is the standard pattern for testing modules that wrap browser globals.
2. **If keeping the hook, rename and document defensively.** `window.__djustRowClickNavigate` reads like a debug shim. Inline source comment is good (lines 58–60); add a top-of-file note in the doc-comment that says "this hook exists solely because JSDOM's `Location.assign` is non-configurable; do not rely on it from app code, it may be removed if/when a cleaner test path appears."

Not a blocker on its own, but combined with R3 it's worth fixing in this PR rather than papering over.

### 🟡 R5 — Test coverage gap: protocol-relative URL not tested

Tied to R1. `tests/js/data_table_row_click.test.js:127–135` exercises `javascript:alert(1)` rejection but does not cover `//evil.com/...`. Add a sibling case:

```javascript
it("rejects protocol-relative URL in data-href (open-redirect defense)", () => {
  const dom = buildDom(clickableRowMarkup({ href: "//evil.com/x" }));
  dom.window.document.querySelector("tr.data-table-row-clickable").click();
  expect(dom.window.__locationAssignSpy).not.toHaveBeenCalled();
});
```

Also consider adding cases for `data:`, `vbscript:`, and bare-fragment `#x` rejection (`#x` is currently rejected; locking it in prevents a future regex widening from accidentally allowing self-page DoS via empty navigations).

### 🟢 G1 — Multi-line `{# %}` fix verified correct

I confirmed against the pre-PR template at commit `859d8374` that both blocks (line 41–43 row-nav comment, line 53–56 link-column comment) used `{# ... #}` across multiple lines. Django's docs are unambiguous: `{# %}` is single-line-only and a multi-line use renders the inner content as raw text. The conversion to `{% comment %}…{% endcomment %}` is correct and self-contained. The implementer's "out of scope" callout for other templates with the same issue is appropriate.

### 🟢 G2 — Two-commit shape is clean

`4b765b9e` touches only: `data-table-row-click.js` (new), `table.html`, `test_data_table_link_row_nav.py`, `data_table_row_click.test.js` (new), `test_data_table_row_navigation_1111.py` (new). No CHANGELOG, no docs.
`e7368381` touches only: `CHANGELOG.md`, `docs/website/guides/components.md`. No code.

Clean separation. No collateral noise.

### 🟢 G3 — CHANGELOG matches the diff

CHANGELOG bullets accurately describe what shipped. Class-name form (`TestRowClickAccessibility`, `TestRowClickableMarkerClass`, …) — good for future audits per `feedback_changelog_required` and the `Bulk dispatch-site refactor + count-test pattern` (#1125) canonicalization. Test count math (14 + 11 = 25) is correct.

### 🟢 G4 — Test coverage is dense and distinct

I read four tests (`test_role_and_tabindex_render_per_row`, `test_selectable_does_not_double_dj_click_on_checkboxes`, the JS `Space inside a nested input does NOT hijack the keystroke` case, the JS `nested <a> click stops propagation so dj-click does not fire` case). They exercise meaningfully different code paths: per-row count assertion, dj-click count discrimination across selectable composition, `activeElement` guard semantics, and capture-phase propagation stop. Not near-duplicates.

The Python test file uses Django's `Engine().from_string()` rather than the production Rust template engine. This matches the convention of `test_data_table_link_row_nav.py` and is fine for the structural assertions made (string-presence in rendered output) since the Rust engine is contractually output-equivalent to Django's for the syntactic constructs in `table.html`.

### 🟢 G5 — Capture-phase `stopImmediatePropagation` semantics for `<a>` are correct

Concern from the brief: does the capture-phase stop block `<a>`'s default navigation? It does not. `stopImmediatePropagation()` halts **propagation to other event listeners** (capture and bubble phase) but does NOT call `preventDefault()`. The `<a>`'s built-in default-action handling runs after all listeners and is unaffected. Test `nested <a> click stops propagation so dj-click does not fire` (lines 218–235) exercises the dj-click suppression; the `<a>`-follows-link case is implicit (no `e.preventDefault()` is called anywhere in the handler).

### 🟢 G6 — `closest()` semantics for the selector are correct

`e.target.closest(NESTED_CONTROL_SELECTOR)` traverses *up* from the click target (inclusive — checks the target first). The `tr.contains(nested)` check then rejects matches where the `closest()` result somehow lives outside this row. Combined with the `e.target !== tr` guard at line 85, the row itself is correctly excluded from the nested-control match. Reviewer-question on this point: passes.

---

## Question

Did the implementer consider option (1) from R2 — auto-injecting the `<script>` from `table.html` when `row_click_event` or `row_url` is set? It would be one line of template, eliminate the "module must be served" caveat from the docs, and make the feature work out-of-the-box for any page using `{% data_table %}`. If there's a reason the framework deliberately doesn't auto-inject component scripts (e.g. the prior `data-table.js` resize module follows the same load-yourself convention by design), call that out — otherwise this seems like the right place to break that pattern.

---

## Self-flag responses

1. **CSP nonce thread absent** (🟡-by-implementer) — confirmed: the chosen architecture is strictly more CSP-friendly. No issue.
2. **Test hook in production JS** (🟡-by-implementer) — see R4. Real-world justified, but a `vi.spyOn`-able exported function would be cleaner. Worth fixing in this PR.
3. **Multi-line `{# %}` fix** (🟡-by-implementer) — verified against the pre-PR template; the leak was real. Out-of-scope deferral for other templates is appropriate.
4. **`<textarea>` added for symmetry** (🟡-by-implementer) — fine, but R3 surfaces three more elements that should also be in the selector.
5. **`dictsort` cell-extraction pattern** (🟡-by-implementer) — pre-existing from #1110. Out of scope. No issue.

---

## Two-commit integrity: PASS
## Test-hook in production code: ACCEPTABLE-WITH-CHANGES (R4 — prefer exported-namespace spy over magic global; not a blocker if rejected with reasoning)
## Open-redirect risk in URL allow-list: MUST FIX (R1)
## Auto-load gap: MUST FIX (R2)

## Summary

| Severity | Count | Items |
|---|---|---|
| 🔴 Critical | 2 | R1 (open-redirect via protocol-relative URL), R2 (JS module never auto-loaded, docs silent) |
| 🟡 Important | 3 | R3 (selector missing `summary`/`option`/`details`), R4 (test hook → exported spy), R5 (test gap on protocol-relative case) |
| 🟢 Pass | 6 | G1 (comment fix), G2 (commit shape), G3 (CHANGELOG), G4 (test density), G5 (`<a>` default action), G6 (`closest()` semantics) |

johnrtipton:
Both 🔴 findings addressed in 2715c40a:

- **R1 (open-redirect)**: regex tightened from `/^(https?:|\/|\.)/` to `/^(https?:\/\/|\/(?!\/)|\.)/`. The `(?!\/)` lookahead rejects `//evil.com` while still accepting single-leading-slash absolute paths. **3 new JS regression tests** lock it in: protocol-relative rejected, single-slash accepted, javascript: rejection regression-locked.
- **R2 (no auto-load)**: `<script src="/static/djust_components/data-table-row-click.js" defer>` injection in `table.html`, guarded by `{% if row_click_event or row_url %}`. Hardcoded `/static/` prefix (default STATIC_URL) so the bare `Engine()` test renderer doesn't trip on `{% load static %}` (which needs `INSTALLED_APPS` the unit-test fixture doesn't provide). Custom-STATIC_URL sites override the template via Django's standard mechanism — documented inline.

3 🟡s (R3 missing tags, R4 test-hook namespace refactor, R5 add Python-side allowlist test) deferred to follow-up since they're polish on top of correct behavior. Will file a tracker after merge.

Tests: 14/14 JS (+3), 28/28 Python (no regression). CI re-running.

johnrtipton:
## Pipeline Retro — PR #1170

**Closed**: #1111. **Drain iter**: 7/7 — **v0.9.1 drain complete!**

### Worked well
- **Stage 11 caught two real 🔴 bugs** that the implementation tests didn't surface: open-redirect via protocol-relative URL (`//evil.com` passing the regex) and the silent fail-no-script-loaded UX. Both were addressed inline; CI rerun on `2715c40a` was clean on first push.
- **CSP-strict architecture** (no inline scripts at all, single static JS module with auto-bind on a marker class) is strictly better than the brief's nonce-attribute suggestion.
- **Defense-in-depth regex on `data-href`** caught the open-redirect class — even though Stage 11 found a hole, the fact that the regex existed at all is what made the fix a 1-character tweak rather than a redesign.

### Didn't work
- **JSDOM 26+ Location non-configurability** required a test-hook in production code (`window.__djustRowClickNavigate`). Filed as 🟡 #1171 R4 — namespace-export refactor is the cleaner long-term shape.
- **Hardcoded `/static/` prefix** for the auto-loaded script. `{% load static %}` would be cleaner but doesn't work under the bare `Engine()` test renderer. Trade-off accepted; documented inline; sites with custom STATIC_URL override the template via Django's standard mechanism.

### Follow-ups filed
- **#1171** — 3 🟡s: missing nested-control tags (`<details>`/`<summary>`/`<option>`), test-hook namespace refactor, Python-side allowlist test.

### Stage outcomes
1-3 inline / 4-10 implementation agent / 11 REQUEST_CHANGES (2 🔴) → APPROVE after `2715c40a` / 12 🔴 fixed inline; 3 🟡 deferred via #1171 / 13 PR_MERGED `d1002e82` / 14 (this comment).

### Drain summary

**v0.9.1 complete** — 7 of 7 PRs shipped:

| PR | Closes | Merge | Theme |
|---|---|---|---|
| #1159 | #1134 | `1f5688af` | bisect 6 flaky tests (in-memory SQLite + sys.modules pollution) |
| #1161 | #1121 | `87ddf976` | Rust template engine custom-filter bridge |
| #1163 | #1146 + #1147 | `5479bc0c` | A075 sticky+lazy check + CSP nonce activator |
| #1164 | #1148 + #1149 + #1150 | `0b85a51d` | hygiene group (replay arg validation, markdown dev-dep, descriptor test) |
| #1166 | #1145 | `bee0f1ee` | Rust live_render lazy=True parity |
| #1168 | #1158 | `07905f5d` | theming cookie namespace |
| #1170 | #1111 | `d1002e82` | data_table row-level navigation |

7 follow-ups filed (#1160, #1162, #1165, #1167, #1169, #1171). All originally-queued issues closed.

Drain milestone retro pending — recommend running `/pipeline-retro --milestone v0.9.1` next.

johnrtipton:
## Retrospective — PR #1170 (backfill)

**Note**: Backfill posted during the v0.9.1-8 final cleanup (#1212 audit). PR #1170 shipped without a Stage 14 retro comment — same outer-loop integrity failure mode as v0.9.5 PRs #1203/#1204.

**PR title**: feat(data_table): row-level navigation accessibility + CSP-strict layer (closes #1111)

PR #1178, opened 2026-04-28
docs: canonicalize CSP-strict defaults for new client-side framework code

johnrtipton:
## Summary

Documents the strict-CSP design constraint as v1.0 readiness. v0.9.1 retro / Action Tracker #183 surfaced a cross-PR pattern (PR #1163 nonce activator + PR #1170 inline-script-free architecture) — this PR encodes the lesson before we forget it.

Closes #1175

## Why

Two converging pieces of evidence in the v0.9.1 drain showed strict-CSP deployments are first-class:
- **PR #1163** added CSP-nonce-aware activator for `<dj-lazy-slot>` fills.
- **PR #1170** went further — chose to skip inline scripts entirely in favor of an external static JS module that auto-binds on a marker class.

The "external module + auto-bind on marker class" shape is strictly more CSP-friendly than "inline script + nonce attribute" and works under stricter CSP policies. Positioning this as a design constraint (not an opt-in) for v1.0.

## Changes

- `CLAUDE.md` — new "Process canonicalizations from v0.9.1 retro arc" section covering all 4 v0.9.1 retro tracker rows (#180/#181/#182/#183).
- `docs/PULL_REQUEST_CHECKLIST.md` — new "CSP-Strict Defaults for New Client-Side Framework Code" subsection under Security Review with 6 explicit checkbox items + 3 canonical reference modules.
- `docs/guides/security.md` — new "CSP-Strict Defaults for Framework Code" section with what-this-means / canonical-modules / what-CSP-headers-to-set, including pip django-csp integration note.
- `CHANGELOG.md` — `[Unreleased] > Changed` entry.

## Test plan

- [x] No code changes — docs + CHANGELOG only.
- [x] Pre-push hooks green (ruff, bandit, detect-secrets, CHANGELOG test counts).
- [ ] Future PRs cite these as the canonical reference (verified in subsequent v0.9.2 drain PRs).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1176, opened 2026-04-28
chore(pipeline): canonicalize two-commit shape + 3-clean-runs gate

johnrtipton:
## Summary

Canonicalize two v0.9.1 retro lessons into the `.pipeline-templates/` shape. No framework code touched — pipeline-template + docs only.

| Issue | Action Tracker | Gate |
|---|---|---|
| #1173 | #181 | Stage 5 forbids CHANGELOG edits; Stage 9 (feature) / Stage 5 (ship) is the canonical CHANGELOG commit boundary |
| #1174 | #182 | bugfix-state.json Stage 6 mandates 3 clean full-suite runs for pollution-class fixes |

Closes #1173
Closes #1174

## Why

**Two-commit shape (#1173)**: v0.9.1 PRs #1163 + #1164 ran two implementer agents concurrently on the same checkout. Both edited `[Unreleased]`; the first commit captured the second's CHANGELOG hunks. Stage 11 caught it as a 🔴. Three subsequent v0.9.1 PRs (#1166, #1168, #1170) used the two-commit shape and shipped clean. Encoding it as a template gate prevents the regression class.

**3-clean-runs (#1174)**: v0.9.1 PR #1159 (the #1134 bisect) caught a hidden second polluter (`sys.modules` rebind) on the third verification run. Single-run pass is insufficient for pollution-class fixes by definition. Lives in `bugfix-state.json` because feature-state.json hardcodes `pipeline_type: \"feature\"` — the conditional would be unreachable there.

## Changes

- `.pipeline-templates/feature-state.json` — Stage 5 forbids CHANGELOG; Stage 9 docs-only gate added.
- `.pipeline-templates/bugfix-state.json` (new) — clone of feature with `pipeline_type: \"bugfix\"` + Stage 6 3-clean-runs gate.
- `.pipeline-templates/ship-state.json` — Stage 5 (Documentation) marked as canonical CHANGELOG commit boundary; docs-only gate added for symmetric protection on ship-pipelines.
- `.pipeline-templates/README.md` — document all three gates with provenance back to v0.9.1 retro Action Tracker #181/#182.

## Stage 11 review notes

Initial review (2025-04-27) found 3 🔴:
- R1: conditional Stage 6 was unreachable in feature-state.json (hardcoded `pipeline_type: \"feature\"`). Fixed in `499c0249` — removed from feature, created bugfix-state.json.
[...]

Issue #1172, opened 2026-04-27
tech-debt: serialize implementer agents per checkout (parallel-agent CHANGELOG contamination)


johnrtipton:
Source: v0.9.1 retro / Action Tracker #180.

## Context

During the v0.9.1 drain, two background implementer agents running concurrently on the same git checkout (iter 3 PR #1163 and iter 4 PR #1164) flipped between branches via pre-commit stash/restore mid-edit. Result: PR #1163's CHANGELOG captured #1164's [Unreleased] entries (#1148 + #1149) for code that wasn't in #1163's diff. Stage 11 caught it as a 🔴 must-fix; resolved with a follow-up commit.

PR #1164 also had a duplicate `### Fixed` heading from the same race. After serializing iters 5-7 (one implementer agent at a time), no further contamination.

## Proposed canonicalization

Update `~/.claude/skills/pipeline-run/SKILL.md` to add an explicit rule:

> **One implementer agent per checkout.** Do not spawn two background implementer agents concurrently against the same git working tree. Concurrent agents flip branches via pre-commit stash/restore and produce CHANGELOG cross-contamination + duplicate-heading collisions on `[Unreleased]` blocks. Either:
> 1. Serialize agent execution (the proven safe path), or
> 2. Use the single-script-transformation pattern: each agent writes a single Python script that applies ALL of its edits in one filesystem pass + commits immediately. Eliminates partial-state windows. (Adopted reactively in v0.9.1 PR #1164; landed clean.)

## Scope

`~/.claude/skills/pipeline-run/SKILL.md` — single section addition. ~20 LoC of doc, no code.

johnrtipton:
Resolved — applied directly to `~/.claude/skills/pipeline-run/SKILL.md`. New section **'One Implementer Agent Per Checkout (the parallel-contention rule)'** added between the existing **Autonomous Execution** and **Stage 1: Branch and Environment Setup** sections.

The section documents:
- The two failure modes (CHANGELOG cross-contamination + duplicate headings)
- The observed cost (v0.9.1 PR #1163 / PR #1164)
- Two acceptable mitigations (serialize OR single-script-transformation)
- When parallel agents ARE safe (different git worktrees / different repos)
- Provenance back to v0.9.1 retro Action Tracker #180

Skill file is outside the djust repo so no PR is required. v0.9.2 drain iter 1 of 7 complete.

Issue #1171, opened 2026-04-27
tech-debt: PR #1170 follow-ups (data_table row navigation polish)


johnrtipton:
Source: Stage 11 review of PR #1170 (closes #1111) — 3 🟡 follow-ups.

## Items

1. **🟡 R3 — Missing tags from nested-control selector.** `<details>`/`<summary>`/`<option>` are not in `NESTED_CONTROL_SELECTOR` (currently 6 tags: `a, button, input, label, select, textarea`). Clicks on a row that contains a `<details>` toggle will trigger row navigation when the user expands/collapses the disclosure, which is wrong. Add the 3 missing tags. ~3-char source change + 2 JS regression tests.

2. **🟡 R4 — Test-hook (`window.__djustRowClickNavigate`) in production code.** Currently the JS module exposes a magic underscored global that tests override to capture navigation calls (JSDOM 26+ `Location.prototype.assign` is non-configurable). Cleaner alternative: refactor `navigate` into the already-exported `window.djustDataTableRowClick` namespace and have `bindRow` call it via closure-captured object so `vi.spyOn(djustDataTableRowClick, 'navigate')` works without the global. ~10 LoC refactor + drop the underscored global.

3. **🟡 R5 — Add a Python-side test for the URL allowlist.** PR #1170 added 3 JS tests for the open-redirect fix (regex tightened to reject `//evil.com`). Worth a Python-side test that asserts `row_url` rendered with a hostile dict value still produces benign output (template-rendered HTML alone won't navigate; the JS guard is the actual defense, but a Python test that documents the allowed shapes belongs alongside).

## Scope

Single follow-up PR, ~30 LoC core + 4-5 tests, ~0.25 day.

## Why deferred from #1170

#1170 had two 🔴 must-fix issues (open-redirect + auto-load script) that were addressed inline in commit `2715c40a`. Stage 11 verdict was REQUEST_CHANGES → APPROVE after the 🔴 fix. The 3 🟡s strengthen but don't change correctness.

PR #1179, opened 2026-04-28
fix(rust): custom filter bridge polish — 6 sub-items (closes #1162)

### LINKED ISSUE #1121: bug: Rust template renderer rejects project-defined custom filters (register.filter)

### LINKED ISSUE #1162: tech-debt: PR #1161 follow-ups (custom filter bridge polish)
Source: Stage 11 review of PR #1161 (closes #1121) — 6 🟡 follow-ups deferred from the merge.

PR #1161 shipped Rust support for project-defined `@register.filter`. Stage 11 verdict was APPROVE with 6 should-fix items, none blocking. Consolidating them here so the merge could ship and the polish can land in a focused follow-up.

## Items

1. **Hot-path Mutex on every render's filter chain.** `is_custom_filter_safe(name)` acquires `std::sync::Mutex` inside the renderer's filter `.any()` loop (`crates/djust_templates/src/renderer.rs:314,369`) even when zero custom filters are registered. Suggested fix: `AtomicBool ANY_CUSTOM_FILTERS_REGISTERED` short-circuit. ~5 LoC in `filter_registry.rs` plus a check at each call site. **Most material 🟡**; matters at scale (high-RPS renders).

2. **Hardcoded `autoescape=true` for `needs_autoescape` filters.** `crates/djust_templates/src/filter_registry.rs:255` passes a literal `true` instead of consulting whether the renderer is inside a `{% autoescape off %}` block. The Rust engine doesn't currently track autoescape blocks (separate gap), so the value is correct *today*, but this pins a behavior that future autoescape work must rewire.

3. **Weak negative-case test for unknown filter.** `tests/unit/test_rust_custom_filters_1121.py`'s "unknown filter still raises" case was flagged as not asserting the specific error message shape — could pass if the dispatch silently degraded to a different error. Tighten the assertion to match `Unknown filter: <name>`.

4. **Unused `pub fn custom_filter_exists`.** Exposed in `filter_registry.rs` for parser-time existence checks (mirrors `registry::handler_exists`) but no consumer yet. Either wire it up to a Stage 4 use case (template parsing pre-flight) or drop it. PyO3 macros suppress the dead-code warning, so it'll quietly rot.

5. **Test fixture clears the global registry.** A test fixture in `test_rust_custom_filters_1121.py` does `filter_registry::clear()` at setup. If any test runs in parallel with another that registered filters, the clear would race. Today the suite is sequential so it's fine, but the fixture should be scoped to a per-test-file isolation (autouse + module scope) and document the constraint.

6. **Silent async filter handling.** If a user defines an `async def filter_x(...)`, the bridge currently calls it and returns the coroutine object as the rendered string ("<coroutine ...>"). Should raise a clear error: "Async filters are not supported by the Rust engine; use a sync wrapper or render via the Python path."

## Scope

Single follow-up PR addressing all 6. Estimated ~50 LoC core + 4-6 tests, ~1 day. No design work needed — each is mechanical given the verdict.

## Why deferred from #1161

#1161's primary scope was the bridge itself (closes #1121). All 6 items are polish on top of a working implementation; landing #1161 unblocks downstream consumers (any project on the Rust render path that registers filters) immediately. None of the 6 cause incorrect behavior in normal use.

johnrtipton:
## Summary

Polish items deferred from PR #1161's Stage 11 review of the eager Rust filter registry that closed #1121. All 6 sub-items addressed in this PR.

Closes #1162.

## Behaviour matrix

| # | Sub-item | Before | After |
|---|---|---|---|
| 1 | Hot-path Mutex perf | ``is_custom_filter_safe`` acquired ``Mutex`` on every name in renderer's ``filter_specs.iter().any(...)`` loop, even with zero custom filters registered | New ``ANY_CUSTOM_FILTERS_REGISTERED: AtomicBool`` short-circuit. Mutex is only touched after ``register_custom_filter`` has flipped the flag. Acquire/Release ordering pairs the load with the store |
| 2 | Hardcoded ``autoescape=true`` | ``apply_custom_filter`` set ``autoescape=true`` kwarg unconditionally, pinning behaviour at the bridge layer | ``apply_custom_filter`` accepts ``autoescape: bool`` parameter; ``apply_filter_full`` threads it; renderer call site supplies ``true`` (matches current behaviour). Future ``{% autoescape %}`` block tracking changes only the renderer call site |
| 3 | Unknown-filter test | ``pytest.raises(Exception)`` + filter-name substring assertion | ``pytest.raises(RuntimeError)`` + assertion on canonical ``"Unknown filter:"`` message shape AND filter name |
| 4 | Unused ``custom_filter_exists`` | Public Rust function with zero callers in workspace; PyO3 macros suppressed dead-code warning | Deleted |
| 5 | Test fixture isolation | ``scope="module"`` autouse fixture had no documentation about parallel-test safety | Same scope, but explicit comment documents that this file is not safe to run in parallel with other tests touching the Rust filter registry |
| 6 | Silent async filter handling | ``async def`` filter stringified the unawaited coroutine into the rendered HTML (``"<coroutine object ...>"``) with a "coroutine was never awaited" RuntimeWarning at GC | ``inspect.iscoroutine(result)`` detects and rejects with a clear, actionable error; ``coro.close()`` suppresses the GC warning |

## Test deltas

- **Rust**: +2 new unit tests in ``filter_registry::tests`` covering the AtomicBool short-circuit pre-registration. ``cargo test -p djust_templates`` → 66 + 2 = 68 cases passing.
- **Python**: +2 cases in ``TestNewBehavior_1162`` covering async-filter rejection (sub-item 6) and ``autoescape`` kwarg flow (sub-item 2). ``tests/unit/test_rust_custom_filters_1121.py`` → 10 + 2 = 12 cases passing.
- **Full Python suite**: 4599 passed, 14 skipped, 0 failed.

## Files touched

- ``crates/djust_templates/src/filter_registry.rs`` — AtomicBool guard, ``autoescape`` parameter, ``iscoroutine`` check + ``coro.close()``, deleted ``custom_filter_exists``, +2 unit tests in ``mod tests``.
- ``crates/djust_templates/src/filters.rs`` — single call site updated to pass ``true`` for ``autoescape`` (pinned at renderer level today).
- ``tests/unit/test_rust_custom_filters_1121.py`` — tightened unknown-filter assertion, parallel-test isolation comment, +``TestNewBehavior_1162`` class with 2 new cases.
- ``CHANGELOG.md`` — ``[Unreleased] > Fixed`` entry.

## Self-review notes for Stage 11

- The ``ANY_CUSTOM_FILTERS_REGISTERED`` flag is **never reset** for the process lifetime, even on ``clear_custom_filters``. This is intentional: clearing is rare (test teardown) and the Mutex path correctly handles the "name not in map" case once we're past the AtomicBool guard. Toggling the flag on clear would race with a concurrent renderer mid-call. Documented in the static's doc comment.
- The Rust ``mod tests`` for the AtomicBool guard is intentionally **ordering-independent**: it asserts the post-registration state is correctly handled by the existing Mutex path, but the pre-registration assertion is gated on a ``OnceLock`` so cargo's non-deterministic test ordering can't false-positive.
- The autoescape parameter is plumbed but the renderer always supplies ``true`` today. When ``{% autoescape %}`` block tracking lands, the renderer's ``apply_filter_full`` call sites (``renderer.rs:287`` and ``renderer.rs:349``, the variable-expansion + inline-if paths) need to consume the surrounding-block state.

## References

- Closes #1162
- Builds on PR #1161 (closed #1121)

## Test plan

- [x] ``cargo test -p djust_templates`` passes (68 cases).
- [x] ``uv run pytest tests/unit/test_rust_custom_filters_1121.py -v`` — all 12 cases pass.
- [x] ``uv run pytest python/djust/ tests/ -q`` — 4599 passed, 14 skipped, 0 failed.
- [x] ``cargo clippy -p djust_templates --all-targets`` — clean.
- [x] ``cargo fmt --check`` + ``ruff check`` + ``ruff format --check`` clean.
- [x] Pre-push hooks pass (pytest + cargo test + cargo audit + noqa + docs).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -179,6 +179,47 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- **Custom filter bridge polish — 6 sub-items deferred from #1161
+  (closes #1162)** — Stage 11 review of PR #1161 (which closed #1121
+  by adding the eager Rust filter registry) flagged six follow-ups.
+  All are addressed in this PR:
+  1. **Hot-path Mutex perf**: ``is_custom_filter_safe`` and
+     ``apply_custom_filter`` short-circuit on a new
+     ``ANY_CUSTOM_FILTERS_REGISTERED`` ``AtomicBool`` so projects with
+     no custom filters pay only an atomic load on every variable
+     expansion's ``filter_specs.iter().any(...)`` loop, never a Mutex
+     acquire. Acquire/Release ordering pairs the load with the store
+     in ``register_custom_filter``.
+  2. **Hardcoded ``autoescape=True`` plumbing**: ``apply_custom_filter``
+     now accepts an ``autoescape: bool`` parameter threaded through
+     ``apply_filter_full`` from the renderer call site. Today still
+     always ``true`` (matches prior behaviour) but the call chain is
+     ready for ``{% autoescape %}`` block tracking landing in a future
+     PR without further changes to ``filter_registry.rs``.
+  3. **Unknown-filter test tightened**: assert ``RuntimeError`` type
+     AND the canonical ``"Unknown filter:"`` message shape, not just
+     ``pytest.raises(Exception)`` + substring on filter name only.
+  4. **Dropped unused ``custom_filter_exists``**: dead public Rust
+     function with no callers in the workspace; PyO3 macros suppress
+     the dead-code warning so it would have rotted silently.
+  5. **Fixture isolation comment**: the ``scope="module"`` autouse
+     fixture in ``tests/unit/test_rust_custom_filters_1121.py`` now
+     carries an explicit comment that this file is not safe to run
+     in parallel with other Rust-filter-registry-touching tests.
+  6. **Silent async filter handling**: an ``async def`` custom filter
+     previously stringified the unawaited coroutine (``"<coroutine
+     object ...>"``) into the rendered HTML with a "coroutine was
+     never awaited" RuntimeWarning at GC. Now uses
+     ``inspect.iscoroutine`` to detect and reject with a clear,
+     actionable error and ``coro.close()`` to suppress the GC warning.
+
+  New cases in ``TestNewBehavior_1162``
+  (``tests/unit/test_rust_custom_filters_1121.py``, 2 Python cases)
+  cover async-filter rejection (sub-item 6) and ``autoescape`` kwarg
+  flow (sub-item 2). Two new Rust unit tests in
+  ``filter_registry::tests`` cover the ``AtomicBool`` short-circuit
+  pre-registration.
+
 - **`replay_event` validates handler is `@event_handler`-decorated
   (closes #1148)** — defense-in-depth strengthening of the v0.9.0
   #1042 forward-replay path. The original guard rejected only
--- a/crates/djust_templates/src/filter_registry.rs
+++ b/crates/djust_templates/src/filter_registry.rs
@@ -42,6 +42,7 @@ use once_cell::sync::Lazy;
 use pyo3::prelude::*;
 use pyo3::types::PyDict;
 use std::collections::HashMap;
+use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::Mutex;
 
 /// Per-filter metadata mirroring Django's filter object attributes.
@@ -65,6 +66,24 @@ struct FilterEntry {
 /// Global registry mapping filter names to Python callables + metadata.
 static FILTER_REGISTRY: Lazy<Mutex<HashMap<String, FilterEntry>>> =
     Lazy::new(|| Mutex::new(HashMap::new()));
+
+/// Hot-path short-circuit guard: ``true`` once any custom filter has been
+/// registered in this process (ever).
+///
+/// The renderer consults [`is_custom_filter_safe`] inside the
+/// ``filter_specs.iter().any(|name| ...)`` loop on every variable
+/// expansion. For the common case — no project-level custom filters —
+/// the Mutex acquire on every name was wasted work. This `AtomicBool`
+/// is checked first; the Mutex is only touched when at least one filter
+/// has actually been registered.
+///
+/// Once flipped to ``true`` it stays that way for the process lifetime
+/// even if ``clear_custom_filters`` empties the registry. That's
+/// intentional: clearing is rare (test teardown) and the Mutex path
+/// then handles the "name not in map" case correctly anyway. The
+/// alternative — toggling the flag on clear — would race with another
+/// thread that's mid-render.
+static ANY_CUSTOM_FILTERS_REGISTERED: AtomicBool = AtomicBool::new(false);
 
 /// Register a project-defined custom filter from Python.
 ///
@@ -99,6 +118,12 @@ pub fn register_custom_filter(
             },
         },
     );
+    // Flip the hot-path guard so renderer's ``is_custom_filter_safe`` stops
+    // short-circuiting and starts consulting the registry. ``Release``
+    // pairs with the renderer's ``Acquire`` load to ensure registry
+    // visibility, though in practice the Mutex acquisition that follows
+    // already provides that ordering. Belt + suspenders.
+    ANY_CUSTOM_FILTERS_REGISTERED.store(true, Ordering::Release);
     Ok(())
 }
 
@@ -147,18 +172,20 @@ pub fn get_registered_custom_filters() -> PyResult<Vec<String>> {
 ///
 /// The renderer consults this alongside the hardcoded built-in
 /// ``safe_output_filters`` list to decide whether to skip auto-escape.
+///
+/// Hot path: this is called once per filter in the
+/// ``filter_specs.iter().any(...)`` loop on every variable expansion.
+/// We short-circuit on the [`ANY_CUSTOM_FILTERS_REGISTERED`]
+/// `AtomicBool` so projects that never register custom filters pay
+/// only an atomic load, not a Mutex acquisition. ``Acquire`` ordering
+/// pairs with the ``Release`` store in [`register_custom_filter`].
 pub fn is_custom_filter_safe(name: &str) -> bool {
+    if !ANY_CUSTOM_FILTERS_REGISTERED.load(Ordering::Acquire) {
+        return false;
+    }
     FILTER_REGISTRY
         .lock()
         .map(|reg| reg.get(name).map(|e| e.meta.is_safe).unwrap_or(false))
-        .unwrap_or(false)
-}
-
-/// Returns ``true`` if any custom filter is registered for the given name.
-pub fn custom_filter_exists(name: &str) -> bool {
-    FILTER_REGISTRY
-        .lock()
-        .map(|reg| reg.contains_key(name))
         .unwrap_or(false)
 }
 
@@ -188,7 +215,13 @@ pub fn apply_custom_filter(
     arg: Option<&str>,
     context: Option<&Context>,
     arg_was_quoted: bool,
+    autoescape: bool,
 ) -> Option<Result<Value, String>> {
+    // Hot-path short-circuit: skip the Mutex when no filter has ever
+    // been registered. Mirrors the guard in ``is_custom_filter_safe``.
+    if !ANY_CUSTOM_FILTERS_REGISTERED.load(Ordering::Acquire) {
+        return None;
+    }
     let (callable, meta) = {
         let registry = FILTER_REGISTRY.lock().ok()?;
         let entry = registry.get(name)?;
@@ -249,10 +282,14 @@ pub fn apply_custom_filter(
 
         let callable_ref = callable.bind(py);
 
-        // Build kwargs: needs_autoescape filters get ``autoescape=True``.
+        // Build kwargs: ``needs_autoescape`` filters get the renderer's
+        // current autoescape policy as a kwarg. Caller (renderer) supplies
+        // the bool — today always ``true``, but threaded through the call
+        // chain so when the Rust engine learns ``{% autoescape %}`` block
+        // tracking, only the renderer call site needs to change.
         let kwargs = if meta.needs_autoescape {
             let kw = PyDict::new(py);
-            kw.set_item("autoescape", true)
+            kw.set_item("autoescape", autoescape)
                 .map_err(|e| format!("Failed to set autoescape kwarg: {e}"))?;
             Some(kw)
         } else {
@@ -273,6 +310,34 @@ pub fn apply_custom_filter(
                 .call1((py_value,))
                 .map_err(|e| format_py_err(py, name, &e))?,
         };
+
+        // Detect ``async def filter_x(...)`` — the user defined a
+        // coroutine function. Without this check the unawaited coroutine
+        // object stringifies to ``"<coroutine object ...>"`` and ends up
+        // in the rendered HTML, with a "coroutine was never awaited"
+        // RuntimeWarning at GC time. Raise a clear error instead so the
+        // author fixes the filter signature.
+        let inspect = py
+            .import("inspect")
+            .map_err(|e| format!("Failed to import inspect: {e}"))?;
+        let is_coro: bool = inspect
+            .call_method1("iscoroutine", (&py_result,))
+            .and_then(|r| r.extract::<bool>())
+            .unwrap_or(false);
+        if is_coro {
+            // Close the coroutine so Python doesn't emit a
+            // "coroutine was never awaited" RuntimeWarning at GC.
+            // ``coro.close()`` is the canonical cleanup for an
+            // unawaited coroutine; ignore any error from close itself
+            // since we're already raising a structured error.
+            let _ = py_result.call_method0("close");
+            return Err(format!(
+                "Custom filter '{name}' is an async function (coroutine); \
+                 the Rust template engine does not support async filters. \
+                 Define '{name}' as a regular ``def`` (sync) filter or \
+                 render this template via the Python path."
+            ));
+        }
 
         // Convert back to Value. Filters typically return strings or
         // SafeStrings; via ``FromPyObject for Value`` either becomes
@@ -296,4 +361,80 @@ fn format_py_err(py: Python<'_>, name: &str, err: &PyErr) -> String {
         err.value(py),
         traceback
     )
+}
+
+#[cfg(test)]
+mod tests {
+    //! Unit tests for the hot-path short-circuit guard (#1162).
+    //!
+    //! Note: the [`ANY_CUSTOM_FILTERS_REGISTERED`] flag is process-global
+    //! and **never resets** for the lifetime of the process — once any
+    //! test registers a filter, every subsequent test sees the flag as
+    //! ``true``. That's the intended production semantics (see the
+    //! ``static`` doc comment), so these tests assert the
+    //! pre-registration state in ordering-independent ways.
+    //!
+    //! Functional cross-checks (registration → render → custom filter
+    //! produces output) live in the Python regression suite at
+    //! ``tests/unit/test_rust_custom_filters_1121.py``.
+    use super::*;
+    use std::sync::atomic::Ordering;
+    use std::sync::OnceLock;
+
+    /// Once this is set, the AtomicBool guard is observed flipped.
+    /// Used so the "before-registration" assertion in
+    /// [`test_atomicbool_short_circuit_when_no_filters_registered`] is
+    /// only made if no other test has already registered a filter.
+    static ALREADY_REGISTERED_BEFORE_TEST: OnceLock<bool> = OnceLock::new();
+
+    #[test]
+    fn test_atomicbool_short_circuit_when_no_filters_registered() {
+        // If a prior test in the same process has already registered
+        // anything, this assertion is meaningless — skip it. Cargo
+        // doesn't guarantee ordering.
+        let was_registered = *ALREADY_REGISTERED_BEFORE_TEST
+            .get_or_init(|| ANY_CUSTOM_FILTERS_REGISTERED.load(Ordering::Acquire));
+        if was_registered {
+            return;
+        }
+        // Pre-registration: ``is_custom_filter_safe`` must short-circuit
+        // and never touch the Mutex. We can't observe Mutex state from
+        // here, but we can assert the lookup returns ``false`` for
+        // arbitrary names — including names we know would be
+        // ``is_safe=true`` if registered. The guard means we never
+        // reach the HashMap.
+        assert!(!is_custom_filter_safe("definitely_not_registered_xyz"));
+        // And the apply path returns ``None`` (filter miss) so the
+        // built-in renderer falls through to the standard
+        // ``Unknown filter`` error.
+        let value = Value::String("x".to_string());
+        let result = apply_custom_filter(
+            "definitely_not_registered_xyz",
+            &value,
+            None,
+            None,
+            false,
+            true,
+        );
+        assert!(result.is_none());
+    }
+
+    #[test]
+    fn test_apply_custom_filter_short_circuits_when_no_filters_registered() {
+        // Same invariant as above, isolated. Independent of
+        // registration state of other tests because we use a name that
+        // can't possibly be registered.
+        let value = Value::String("x".to_string());
+        let result = apply_custom_filter(
+            "__provably_unregistered_name__",
+            &value,
+            None,
+            None,
+            false,
+            true,
+        );
+        // None == miss; an unknown name is always a miss whether or
+        // not the AtomicBool guard short-circuited the Mutex.
+        assert!(result.is_none());
+    }
 }
--- a/crates/djust_templates/src/filters.rs
+++ b/crates/djust_templates/src/filters.rs
@@ -498,12 +498,18 @@ pub fn apply_filter_full(
             // registry for project-defined ``@register.filter`` callables
             // (issue #1121). Bridge: ``filter_registry::apply_custom_filter``
             // returns ``Some(Ok|Err)`` on hit, ``None`` on miss.
+            //
+            // ``autoescape=true`` is supplied here as the engine's current
+            // (pinned) policy. When ``{% autoescape %}`` block tracking
+            // lands in a future PR, the renderer will thread the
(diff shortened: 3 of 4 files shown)


johnrtipton (commented):
## Stage 11 review — PR #1179 (custom filter bridge polish, 6 sub-items)

**Verdict: APPROVE WITH MINOR FINDINGS** — implementation is correct; concurrency reasoning, async handling, and dead-code deletion are all sound. Two non-blocking accuracy issues with the PR description / CHANGELOG narrative (sub-item 2) that should be fixed before merge so the next reviewer of `apply_filter_full` isn't misled. Two-commit shape is clean.

Counts: 🔴 0 / 🟡 3 / 🟢 6

---

### 🟢 AtomicBool short-circuit ordering — correct

`Acquire` on the load (`filter_registry.rs:182, 222`), `Release` on the store (`filter_registry.rs:126`). The store happens-after the HashMap insert in program order:

```rust
// filter_registry.rs:108-126
let mut registry = FILTER_REGISTRY.lock().map_err(...)?;
registry.insert(name, FilterEntry { ... });
// ...
ANY_CUSTOM_FILTERS_REGISTERED.store(true, Ordering::Release);
Ok(())
```

The `Release/Acquire` pair establishes happens-before from the HashMap write to any reader that observes the flag. The Mutex acquisition on the reader side then serialises with the writer's Mutex unlock (which happens at end-of-function, after the AtomicBool store), so the reader sees the inserted entry. The "belt + suspenders" comment at line 122-125 is accurate — both the AtomicBool fence AND the Mutex provide ordering, and either alone suffices.

TOCTOU window: a reader can observe `ANY=false` while another thread is mid-`register_custom_filter`. Worst case is one render misses a `is_safe=true` lookup, double-escapes the output, and produces over-escaped (but secure) HTML. Not a security regression, just transient fidelity loss during the registration window. Implementer's reasoning checks out.

### 🟢 Async filter detection — catches both forms

Verified `inspect.iscoroutine()` catches both `async def filter()` AND a sync `def filter()` that internally calls an async function and returns the coroutine without awaiting. Confirmed locally:

```python
>>> async def f(x): return x
>>> def g(x): return f(x)  # sync def returning coroutine
>>> inspect.iscoroutine(f('a')), inspect.iscoroutine(g('b'))
(True, True)
```

Both exotic cases are caught. The cleanup-ordering is also correct — `coro.close()` at `filter_registry.rs:333` is invoked BEFORE the error string is constructed at `:334`, so Python doesn't emit a "coroutine was never awaited" RuntimeWarning before our structured error returns. Error message at :335-339 names the filter, names the problem ("async function (coroutine)"), and gives two actionable fixes.

### 🟢 Dropped `custom_filter_exists` — safe

Verified zero callers in `crates/`, `python/djust/`, and `tests/`:

```
$ grep -rn 'custom_filter_exists' crates/ python/ tests/
(no matches)
```

Pre-PR the function was `pub fn` (not `#[pyfunction]`) at `filter_registry.rs:158` (in main). It was distinct from `has_custom_filter` (line 116, kept) which IS the PyO3-exposed wrapper. The wrapped version is still called via `m.add_function(wrap_pyfunction!(djust_templates::filter_registry::has_custom_filter, m))` in `crates/djust_live/src/lib.rs:2741`, so external Python consumers are unaffected. Safe deletion.

### 🟢 Tightened unknown-filter test — message shape pinned correctly

Pre-PR: `pytest.raises(Exception)` + substring match on filter name. Post-PR (`tests/unit/test_rust_custom_filters_1121.py:213-223`):

```python
with pytest.raises(RuntimeError) as exc_info:
    render_template("{{ x|definitely_not_a_filter_xyz }}", {"x": "val"})
msg = str(exc_info.value)
assert "Unknown filter:" in msg, f"unexpected error shape: {msg!r}"
assert "definitely_not_a_filter_xyz" in msg
```

Verified the message text matches what `filters.rs:517` actually emits: `"Unknown filter: {filter_name}"`. The assertion would catch a future regression that reformatted to `"filter not found"` or similar.

### 🟢 Fixture isolation — comment present and accurate

The `scope="module"` autouse fixture at `tests/unit/test_rust_custom_filters_1121.py:95-112` carries both the file-level note (lines 22-29) AND an inline comment (lines 88-94) explaining the parallel-test-unsafe semantics. Teardown comment at lines 105-109 also notes the AtomicBool's intentionally-non-resetting behaviour with a pointer to the Rust doc comment. Good cross-referencing.

### 🟢 Two-commit shape — clean

```
c9aa6e8f fix(rust): custom filter bridge polish — 6 sub-items
  crates/djust_templates/src/filter_registry.rs +151/-10
  crates/djust_templates/src/filters.rs           +6/-0
  tests/unit/test_rust_custom_filters_1121.py    +113/-3

9c37fb56 docs(rust-filters): note custom filter bridge polish in CHANGELOG
  CHANGELOG.md                                   +41/-0
```

Impl commit has zero CHANGELOG hunks, docs commit has only CHANGELOG. v0.9.1 retro pattern observed.

---

### 🟡 Finding 1 — PR description / CHANGELOG mis-state autoescape threading depth

Both the PR description and the CHANGELOG say:

> ``apply_custom_filter`` now accepts an ``autoescape: bool`` parameter **threaded through ``apply_filter_full`` from the renderer call site**.

This is **inaccurate**. The actual diff:

- `apply_custom_filter` signature DID gain `autoescape: bool` (`filter_registry.rs:218`). ✅
- `apply_filter_full` signature is **unchanged** — still 5 parameters (`filters.rs:33-39`). The hardcoded `true` at the bridge call site (`filters.rs:512`) is supplied by `apply_filter_full` itself, NOT by the renderer.
- The renderer's three call sites of `apply_filter_full` (`renderer.rs:287`, `:349`, `:1602`) are untouched by this PR.

So when `{% autoescape %}` block tracking lands, it's NOT just "the renderer call site" that changes. The future PR will need to:
1. Add `autoescape: bool` parameter to `apply_filter_full`.
2. Update all 3 renderer call sites to pass it.
3. Update the internal forward at `filters.rs:506-513` to pass it through instead of hardcoding `true`.
4. Update `pub fn apply_filter` at `filters.rs:10` (which forwards to `apply_filter_full` with `true`) — that one might keep hardcoded `true` as a back-compat default, depending on policy.

The plumbing claim in the CHANGELOG implies "one-line future change at the renderer call site" — actual future change is closer to four sites + signature widening of the public `apply_filter_full` symbol. The CHANGELOG bullet 2 should be reworded to:

> "**Hardcoded `autoescape=True`**: `apply_custom_filter` now accepts an `autoescape: bool` parameter. `apply_filter_full` supplies `true` at the bridge call site today; threading the value from renderer state requires a follow-up PR that widens `apply_filter_full`'s signature."

### 🟡 Finding 2 — `test_needs_autoescape_filter_receives_autoescape_kwarg` is a green-on-arrival test

The new test at `tests/unit/test_rust_custom_filters_1121.py:280-322` asserts `captured["autoescape"] is True` and `"ae=True" in out`. But the pre-PR code at the same site (`filter_registry.rs:255` in main) already hardcoded `kw.set_item("autoescape", true)`. So this test would have passed BEFORE the PR too — it doesn't exercise sub-item 2's actual change (the parameterisation of the inner function).

To genuinely cover the new wiring, the test should either:
(a) call `apply_custom_filter` directly via a Rust unit test with `autoescape=false` and assert the kwarg arrives as `False`, OR
(b) accept that "the kwarg flows through" is the only coverage achievable from Python today (since the renderer always passes `true`), and rename the test to reflect that it's a substrate-presence test, not a behaviour-difference test.

Not blocking — the test isn't wrong, just doesn't add information beyond what existed before the PR. Worth noting in the retro.

### 🟡 Finding 3 — minor inaccuracy: "PyO3 macros suppress the dead-code warning"

PR description sub-item 4 and CHANGELOG bullet 4 say:

> "PyO3 macros suppress the dead-code warning so it would have rotted silently."

But the deleted `custom_filter_exists` was a plain `pub fn`, NOT decorated with `#[pyfunction]` (verified against main). What suppressed the dead-code warning was the `pub` visibility itself — Rust's `dead_code` lint doesn't fire on public items because they could have external consumers. PyO3 had nothing to do with it.

Minor accuracy issue in the explanatory text only — the deletion itself is correct. Worth fixing to avoid future confusion when someone greps the codebase looking for "PyO3 dead-code suppression patterns".

---

### Question for the implementer

The Rust unit test `test_atomicbool_short_circuit_when_no_filters_registered` uses `OnceLock<bool>` to capture the pre-test registration state and skip the assertion if a prior test already flipped the flag. This is correct for the ordering-independence requirement, but it means **the test silently no-ops when run AFTER any test that registers a filter**. Is there a CI invocation that runs filter_registry tests in isolation (e.g. `cargo test -p djust_templates filter_registry::tests`) to guarantee at least one run actually exercises the pre-registration assertion? If not, the test is effectively decorative whenever the suite runs in `cargo test` default order with prior registration tests. Even a one-line `--test-threads=1` note in the doc comment, plus a CI step that runs `cargo test -p djust_templates filter_registry::tests::test_atomicbool_short_circuit -- --exact` in a fresh process, would close that gap.

---

### Summary

Approve. Findings 1 and 3 are wording-only; finding 2 is a non-blocking observation about a test that doesn't fail-fast against a removed protection. Concurrency reasoning is sound, async detection is robust against both `async def` and sync-returns-coroutine forms, dead-code deletion verified safe across the workspace, and the two-commit shape matches the v0.9.1 retro pattern.

johnrtipton:
## Stage 11 review — PR #1179 (custom filter bridge polish, 6 sub-items)

**Verdict: APPROVE WITH MINOR FINDINGS** — implementation is correct; concurrency reasoning, async handling, and dead-code deletion are all sound. Two non-blocking accuracy issues with the PR description / CHANGELOG narrative (sub-item 2) that should be fixed before merge so the next reviewer of `apply_filter_full` isn't misled. Two-commit shape is clean.

Counts: 🔴 0 / 🟡 3 / 🟢 6

---

### 🟢 AtomicBool short-circuit ordering — correct

`Acquire` on the load (`filter_registry.rs:182, 222`), `Release` on the store (`filter_registry.rs:126`). The store happens-after the HashMap insert in program order:

```rust
// filter_registry.rs:108-126
let mut registry = FILTER_REGISTRY.lock().map_err(...)?;
registry.insert(name, FilterEntry { ... });
// ...
ANY_CUSTOM_FILTERS_REGISTERED.store(true, Ordering::Release);
Ok(())
```

The `Release/Acquire` pair establishes happens-before from the HashMap write to any reader that observes the flag. The Mutex acquisition on the reader side then serialises with the writer's Mutex unlock (which happens at end-of-function, after the AtomicBool store), so the reader sees the inserted entry. The "belt + suspenders" comment at line 122-125 is accurate — both the AtomicBool fence AND the Mutex provide ordering, and either alone suffices.

TOCTOU window: a reader can observe `ANY=false` while another thread is mid-`register_custom_filter`. Worst case is one render misses a `is_safe=true` lookup, double-escapes the output, and produces over-escaped (but secure) HTML. Not a security regression, just transient fidelity loss during the registration window. Implementer's reasoning checks out.

### 🟢 Async filter detection — catches both forms

Verified `inspect.iscoroutine()` catches both `async def filter()` AND a sync `def filter()` that internally calls an async function and returns the coroutine without awaiting. Confirmed locally:

```python
>>> async def f(x): return x
>>> def g(x): return f(x)  # sync def returning coroutine
>>> inspect.iscoroutine(f('a')), inspect.iscoroutine(g('b'))
(True, True)
```

Both exotic cases are caught. The cleanup-ordering is also correct — `coro.close()` at `filter_registry.rs:333` is invoked BEFORE the error string is constructed at `:334`, so Python doesn't emit a "coroutine was never awaited" RuntimeWarning before our structured error returns. Error message at :335-339 names the filter, names the problem ("async function (coroutine)"), and gives two actionable fixes.

### 🟢 Dropped `custom_filter_exists` — safe

Verified zero callers in `crates/`, `python/djust/`, and `tests/`:

```
$ grep -rn 'custom_filter_exists' crates/ python/ tests/
(no matches)
```

Pre-PR the function was `pub fn` (not `#[pyfunction]`) at `filter_registry.rs:158` (in main). It was distinct from `has_custom_filter` (line 116, kept) which IS the PyO3-exposed wrapper. The wrapped version is still called via `m.add_function(wrap_pyfunction!(djust_templates::filter_registry::has_custom_filter, m))` in `crates/djust_live/src/lib.rs:2741`, so external Python consumers are unaffected. Safe deletion.

### 🟢 Tightened unknown-filter test — message shape pinned correctly

Pre-PR: `pytest.raises(Exception)` + substring match on filter name. Post-PR (`tests/unit/test_rust_custom_filters_1121.py:213-223`):

```python
with pytest.raises(RuntimeError) as exc_info:
    render_template("{{ x|definitely_not_a_filter_xyz }}", {"x": "val"})
msg = str(exc_info.value)
assert "Unknown filter:" in msg, f"unexpected error shape: {msg!r}"
assert "definitely_not_a_filter_xyz" in msg
```

Verified the message text matches what `filters.rs:517` actually emits: `"Unknown filter: {filter_name}"`. The assertion would catch a future regression that reformatted to `"filter not found"` or similar.

### 🟢 Fixture isolation — comment present and accurate

The `scope="module"` autouse fixture at `tests/unit/test_rust_custom_filters_1121.py:95-112` carries both the file-level note (lines 22-29) AND an inline comment (lines 88-94) explaining the parallel-test-unsafe semantics. Teardown comment at lines 105-109 also notes the AtomicBool's intentionally-non-resetting behaviour with a pointer to the Rust doc comment. Good cross-referencing.

### 🟢 Two-commit shape — clean

```
c9aa6e8f fix(rust): custom filter bridge polish — 6 sub-items
  crates/djust_templates/src/filter_registry.rs +151/-10
  crates/djust_templates/src/filters.rs           +6/-0
  tests/unit/test_rust_custom_filters_1121.py    +113/-3

9c37fb56 docs(rust-filters): note custom filter bridge polish in CHANGELOG
  CHANGELOG.md                                   +41/-0
```

Impl commit has zero CHANGELOG hunks, docs commit has only CHANGELOG. v0.9.1 retro pattern observed.

---

### 🟡 Finding 1 — PR description / CHANGELOG mis-state autoescape threading depth

Both the PR description and the CHANGELOG say:

> ``apply_custom_filter`` now accepts an ``autoescape: bool`` parameter **threaded through ``apply_filter_full`` from the renderer call site**.

This is **inaccurate**. The actual diff:

- `apply_custom_filter` signature DID gain `autoescape: bool` (`filter_registry.rs:218`). ✅
- `apply_filter_full` signature is **unchanged** — still 5 parameters (`filters.rs:33-39`). The hardcoded `true` at the bridge call site (`filters.rs:512`) is supplied by `apply_filter_full` itself, NOT by the renderer.
- The renderer's three call sites of `apply_filter_full` (`renderer.rs:287`, `:349`, `:1602`) are untouched by this PR.

So when `{% autoescape %}` block tracking lands, it's NOT just "the renderer call site" that changes. The future PR will need to:
1. Add `autoescape: bool` parameter to `apply_filter_full`.
2. Update all 3 renderer call sites to pass it.
3. Update the internal forward at `filters.rs:506-513` to pass it through instead of hardcoding `true`.
4. Update `pub fn apply_filter` at `filters.rs:10` (which forwards to `apply_filter_full` with `true`) — that one might keep hardcoded `true` as a back-compat default, depending on policy.

The plumbing claim in the CHANGELOG implies "one-line future change at the renderer call site" — actual future change is closer to four sites + signature widening of the public `apply_filter_full` symbol. The CHANGELOG bullet 2 should be reworded to:

> "**Hardcoded `autoescape=True`**: `apply_custom_filter` now accepts an `autoescape: bool` parameter. `apply_filter_full` supplies `true` at the bridge call site today; threading the value from renderer state requires a follow-up PR that widens `apply_filter_full`'s signature."

### 🟡 Finding 2 — `test_needs_autoescape_filter_receives_autoescape_kwarg` is a green-on-arrival test

The new test at `tests/unit/test_rust_custom_filters_1121.py:280-322` asserts `captured["autoescape"] is True` and `"ae=True" in out`. But the pre-PR code at the same site (`filter_registry.rs:255` in main) already hardcoded `kw.set_item("autoescape", true)`. So this test would have passed BEFORE the PR too — it doesn't exercise sub-item 2's actual change (the parameterisation of the inner function).

To genuinely cover the new wiring, the test should either:
(a) call `apply_custom_filter` directly via a Rust unit test with `autoescape=false` and assert the kwarg arrives as `False`, OR
(b) accept that "the kwarg flows through" is the only coverage achievable from Python today (since the renderer always passes `true`), and rename the test to reflect that it's a substrate-presence test, not a behaviour-difference test.

Not blocking — the test isn't wrong, just doesn't add information beyond what existed before the PR. Worth noting in the retro.

### 🟡 Finding 3 — minor inaccuracy: "PyO3 macros suppress the dead-code warning"

PR description sub-item 4 and CHANGELOG bullet 4 say:

> "PyO3 macros suppress the dead-code warning so it would have rotted silently."

But the deleted `custom_filter_exists` was a plain `pub fn`, NOT decorated with `#[pyfunction]` (verified against main). What suppressed the dead-code warning was the `pub` visibility itself — Rust's `dead_code` lint doesn't fire on public items because they could have external consumers. PyO3 had nothing to do with it.

Minor accuracy issue in the explanatory text only — the deletion itself is correct. Worth fixing to avoid future confusion when someone greps the codebase looking for "PyO3 dead-code suppression patterns".

---

### Question for the implementer

The Rust unit test `test_atomicbool_short_circuit_when_no_filters_registered` uses `OnceLock<bool>` to capture the pre-test registration state and skip the assertion if a prior test already flipped the flag. This is correct for the ordering-independence requirement, but it means **the test silently no-ops when run AFTER any test that registers a filter**. Is there a CI invocation that runs filter_registry tests in isolation (e.g. `cargo test -p djust_templates filter_registry::tests`) to guarantee at least one run actually exercises the pre-registration assertion? If not, the test is effectively decorative whenever the suite runs in `cargo test` default order with prior registration tests. Even a one-line `--test-threads=1` note in the doc comment, plus a CI step that runs `cargo test -p djust_templates filter_registry::tests::test_atomicbool_short_circuit -- --exact` in a fresh process, would close that gap.

---

### Summary

Approve. Findings 1 and 3 are wording-only; finding 2 is a non-blocking observation about a test that doesn't fail-fast against a removed protection. Concurrency reasoning is sound, async detection is robust against both `async def` and sync-returns-coroutine forms, dead-code deletion verified safe across the workspace, and the two-commit shape matches the v0.9.1 retro pattern.

johnrtipton:
## Pipeline Retro — PR #1179

**Closes**: #1162. **Drain iter**: 4/8.

### Worked well
- **AtomicBool short-circuit + Acquire/Release ordering** correctly implemented; reviewer verified the happens-before pairing. Hot-path Mutex acquired only when filters are registered.
- **Two-commit shape** held cleanly (impl+tests in `c9aa6e8f`, CHANGELOG-only in `9c37fb56`); reviewer confirmed.
- **Single agent on this checkout** (per #180 / #1172 canonicalization) — zero contamination.

### Didn't work / 🟡 caught by Stage 11
- PR description claimed `apply_filter_full` was widened, but actual change was on `apply_custom_filter`. Doc accuracy issue, not a code bug. Filed in #1180.
- Sub-item 2's test is structurally green-on-arrival because pre-PR code already hardcoded the kwarg. The API change is correct; the test isn't tight enough to exercise the new semantics. Filed in #1180.

### Follow-up filed
- **#1180** — 3 🟡s: doc accuracy on `apply_filter_full` vs `apply_custom_filter`, sub-item 2 test strengthening, CI invocation question for isolated `filter_registry::tests` runs.

### Drain progress
v0.9.2: **4 of 8 work units shipped** — #1172, #1176, #1178, #1179.

Remaining: 4 PRs (#1160+#1165 hygiene group, #1167 tag-registry+sidecar, #1169 cookie polish, #1171 data_table polish).

johnrtipton:
## Retrospective — PR #1179 (backfill)

**Note**: Backfill posted during the v0.9.1-8 final cleanup (#1212 audit). PR #1179 shipped without a Stage 14 retro comment — same outer-loop integrity failure mode as v0.9.5 PRs #1203/#1204.

**PR title**: fix(rust): custom filter bridge polish — 6 sub-items (closes #1162)

Issue #1173, opened 2026-04-27
tech-debt: canonicalize two-commit shape (impl+tests / docs+CHANGELOG) in pipeline template Stage 9


johnrtipton:
Source: v0.9.1 retro / Action Tracker #181.

## Context

Three v0.9.1 PRs (#1166, #1168, #1170) adopted a two-commit shape: implementation + tests in commit 1, docs + CHANGELOG entry in commit 2. Stage 11 reviewers explicitly verified each split was clean. Three-for-three, the pattern eliminated the cross-edit collision class that bit #1163/#1164 (where CHANGELOG hunks landed in the wrong PR's diff).

## Proposed canonicalization

Update `.pipeline-templates/feature-state.json` and `.pipeline-templates/bugfix-state.json` to make the two-commit shape mandatory:

- Stage 5 (Implementation): explicitly forbid CHANGELOG.md edits in the implementation commit.
- Stage 9 (Documentation): becomes the canonical commit boundary for CHANGELOG hunks + docs guide updates.
- Pipeline-run skill text adds: 'After Stage 5 commit, the working tree must NOT contain any CHANGELOG.md modifications. Stage 9 is the canonical commit boundary for [Unreleased] entries.'

## Scope

State-file template tweak + skill doc update. ~30 LoC across both templates + ~15 LoC skill doc.

## Why deferred from v0.9.1

The pattern was learned and applied reactively across iters 5-7. Canonicalizing it makes future drains immune to the #1163/#1164 collision class.

PR #1182, opened 2026-04-28
fix(rust): tag-registry test isolation + sidecar bridge extension (closes #1167)

johnrtipton:
## Summary

Two Stage 11 follow-ups from PR #1166 (which wired the raw-Python sidecar into `Node::CustomTag`).

- **Sub-item 1: tag-registry test isolation.** `tests/unit/test_tag_registry.py` previously used per-class `setup_registry` fixtures that only re-registered Python built-in handlers on teardown — they never *cleared* the global Rust `TAG_HANDLERS` registry. Transient handlers from the file (notably `BrokenHandler` registered for the `broken` tag in `test_handler_exception_returns_error`) leaked into subsequent test files. `test_assign_tag.py` running after this file would see `handler_exists("broken") == True`; the parser checks `handler_exists` before `assign_handler_exists`, so `{% broken %}` was routed to the leaked `CustomTag` handler and `test_non_dict_return_is_empty_merge` failed with the leaked handler's exception. Replaced the per-class fixtures with one function-scoped autouse fixture that clears all three Rust registries (tag / block-tag / assign-tag) before AND after every test, then re-registers the built-ins.

- **Sub-item 2: sidecar parity for block + assign tags.** PR #1166's `call_handler_with_py_sidecar` only fired for `Node::CustomTag`. Block tags (`Node::BlockCustomTag`) and assign tags (`Node::AssignTag`) didn't receive the `request` / `view` sidecar. Added `call_block_handler_with_py_sidecar` and `call_assign_handler_with_py_sidecar` mirroring the PR #1166 pattern; existing `call_block_handler` / `call_assign_handler` are kept as back-compat shims that delegate with `None`. All five renderer call sites (1× block, 4× assign — single-node, sibling-aware, collecting, and partial-render paths) forward `context.raw_py_objects()`.

## Issue × file × test mapping

| Issue / Sub-item | Files modified | Tests covering it |
|---|---|---|
| #1167 sub-item 1 (test isolation) | `tests/unit/test_tag_registry.py` (new module-scoped autouse `_isolate_tag_registries`; per-class `setup_registry` fixtures removed) | `pytest tests/unit/test_tag_registry.py tests/unit/test_assign_tag.py -v` now passes (was failing on `test_non_dict_return_is_empty_merge` before) |
[...]

PR #1181, opened 2026-04-28
chore(v0.9.2): hygiene group — Redis docstring + caplog + descriptor doc + dev-env script

johnrtipton:
## Summary

Stage 11 follow-ups from the v0.9.1 retro arc, batched as a single chore PR. Four small surface-area changes; all guard against future-regression failure modes that the side-effect-only tests would have missed.

- **#1160**: rewrite `test_redis_serialization_performance` docstring to match what the 100ms bound actually catches (catastrophic ~10× regressions, not gradual perf drift). Points users to `pytest-benchmark`-style median-based assertions for SLA-grade perf checks.
- **#1165 (a)**: extend `TestReplayHandlerValidation` rejection-path tests to assert via `caplog` that the `logger.warning(...)` record fires with the expected message text. Without this, a future regression that silently demotes the warning to a no-op would leave the side-effect assertion green.
- **#1165 (b)**: document the descriptor-pattern auto-promotion gap in the `LiveComponent` docstring (`python/djust/components/base.py`) and in `docs/website/guides/components.md`. The framework's `_assign_component_ids` walker only inspects instance-level attrs, so descriptor components must be appended to `self._components` in `mount()` until auto-promotion ships.
- **#1165 (c)**: add `scripts/check-dev-env-imports.py` and a paired pytest module (`tests/unit/test_dev_env_imports.py`, 2 parametrized cases) that hard-fail (not skip) if `djust.components.components` or its `.markdown` submodule can't import. Locks in the #1149 fix.

## Files changed

| Item | Files |
|------|-------|
| #1160 docstring | `tests/unit/test_state_backend.py` |
| #1165 (a) caplog | `tests/unit/test_time_travel.py` |
| #1165 (b) docs | `python/djust/components/base.py`, `docs/website/guides/components.md` |
| #1165 (c) dev-env | `scripts/check-dev-env-imports.py`, `tests/unit/test_dev_env_imports.py` |
| CHANGELOG | `CHANGELOG.md` |

## Test plan

- [x] `uv run pytest tests/unit/test_state_backend.py tests/unit/test_time_travel.py tests/unit/test_dev_env_imports.py -q` — 90 passed, 2 skipped
- [x] Full suite: `uv run pytest python/djust/ tests/ -q` — 4601 passed, 14 skipped
- [x] `python scripts/check-dev-env-imports.py` — exits 0 with "OK: all 2 critical dev-env imports succeeded"
[...]

PR #1183, opened 2026-04-28
fix(theming): cookie namespace polish — 4 sub-items (closes #1169)

### LINKED ISSUE #1169: tech-debt: PR #1168 follow-ups (cookie namespace polish)

johnrtipton:
## Summary

Four follow-ups from the PR #1168 Stage 11 review of the per-project
cookie namespacing introduced in #1158:

- **(a)** Empty namespaced cookie no longer falls back to legacy unprefixed.
  The previous `_read('<ns>_name') or None` evaluated `""` as falsy and
  fell through to the unprefixed name, re-opening the cross-project
  cookie bleed path #1158 closed. Now distinguishes `None` (cookie not in
  jar) from `""` (cookie set to empty), and only falls back in the former
  case.
- **(b)** `LIVEVIEW_CONFIG['theme']['cookie_namespace']` now validates at
  config-load via `_validate_cookie_namespace()`. Whitespace, `=`, `;`,
  and non-ASCII characters now raise `ImproperlyConfigured` at startup
  instead of silently producing malformed Set-Cookie headers (browsers
  reject or split such cookies). Allowed charset: `[A-Za-z0-9_-]+`.
- **(c)** New JSDOM tests in `tests/js/theming_cookie_namespace_write.test.js`
  cover the JS-side cookie WRITE path — the 8 #1158 Python tests only
  asserted on the source string of `theme.js`. The JSDOM tests load the
  file in a DOM, set `window.__djust_theme_cookie_prefix`, fire
  `setPack`/`setPreset`/`setLayout`, and inspect `document.cookie`.
- **(d)** Legacy-cookie cleanup: when writing a namespaced theming
  cookie, `theme.js` now also emits `Max-Age=0` for the unprefixed cookie
  of the same logical name. Otherwise stale legacy cookies left over from
  before namespace was configured persist in the browser jar forever and
  bleed back if namespace is later removed. Cleanup is inert when no
  prefix is configured (back-compat).

| Sub-item | Issue | Implementation file | Covering tests |
|---|---|---|---|
| (a) Empty cookie fallback fix | #1169(a) | `python/djust/theming/manager.py` (`_read` / `get_state`) | `tests/unit/test_theming_cookie_namespace_1158.py::test_empty_namespaced_cookie_does_not_fall_back_to_legacy` |
| (b) `cookie_namespace` validation | #1169(b) | `python/djust/theming/manager.py` (`_validate_cookie_namespace`, `get_theme_config`) | `tests/unit/test_theming_cookie_namespace_1158.py::test_cookie_namespace_validator_accepts_valid_values`, `::test_cookie_namespace_validator_rejects_illegal_characters` |
| (c) JSDOM write-side test | #1169(c) | (test-only) | `tests/js/theming_cookie_namespace_write.test.js` (4 cases under `#1169(c)` describe block) |
| (d) Legacy cookie cleanup | #1169(d) | `python/djust/theming/static/djust_theming/js/theme.js` (`_deleteLegacyCookie` + 5 callsites) | `tests/js/theming_cookie_namespace_write.test.js` (3 cases under `#1169(d)` describe block) |

## Test count delta

- Python: +3 new cases (1 for (a), 2 for (b)) in `tests/unit/test_theming_cookie_namespace_1158.py` (8 → 11 total).
- JS: +7 new cases in new file `tests/js/theming_cookie_namespace_write.test.js` (4 for (c), 3 for (d)).

## Two-commit shape

Commit 1: implementation + tests (no CHANGELOG, per v0.9.1 retro / #1173).
Commit 2: CHANGELOG `[Unreleased] > Fixed` entry only.

## Test plan

- [x] `uv run pytest tests/unit/test_theming_cookie_namespace_1158.py -v` — 11/11 pass
- [x] `npx vitest run tests/js/theming_cookie_namespace_write.test.js` — 7/7 pass
- [x] `uv run pytest python/djust/tests/ tests/unit/` — 4386 pass, 6 skip (no regressions)
- [x] `npx vitest run` — 1460/1460 pass (full JS suite)
- [x] Pre-commit hooks pass on both commits
- [x] Pre-push hooks pass (full pytest + cargo + linters)

Closes #1169

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -209,6 +209,40 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- **Theming cookie namespace polish — 4 sub-items (closes #1169)** —
+  Stage 11 follow-ups from PR #1168 (the original cookie-namespace work
+  for #1158):
+  - **(a) Empty namespaced cookie no longer falls back to legacy.**
+    `ThemeManager.get_state()` previously evaluated the namespaced
+    cookie via `_read('<ns>_name') or None`, so an empty-string value
+    (`""`) silently fell through to the unprefixed legacy cookie —
+    re-opening the cross-project bleed path #1158 closed. The read
+    now distinguishes `None` (cookie not in jar) from `""` (cookie set
+    to empty), and only falls back in the former case.
+  - **(b) `cookie_namespace` validated at config-load.** The value is
+    interpolated directly into cookie names; whitespace, `=`, `;`, and
+    non-ASCII characters previously produced malformed Set-Cookie
+    headers (browsers reject or split such cookies).
+    `_validate_cookie_namespace()` now raises `ImproperlyConfigured`
+    at startup for any value outside `[A-Za-z0-9_-]+`.
+  - **(c) JSDOM tests for the cookie WRITE side.** The 8 #1158
+    Python tests only asserted on `theme.js` source-text patterns; new
+    `tests/js/theming_cookie_namespace_write.test.js` loads the file
+    in JSDOM, sets `window.__djust_theme_cookie_prefix`, fires
+    `setPack`/`setPreset`/`setLayout`, and inspects `document.cookie`.
+  - **(d) Legacy-cookie cleanup on first namespaced write.** When
+    `cookie_namespace` is set, every theming-cookie write in
+    `theme.js` now also emits `Max-Age=0` for the unprefixed legacy
+    name. Stale legacy cookies left over from before namespace was
+    configured no longer sit in the jar forever and bleed back if the
+    namespace is later removed. Cleanup is inert when no prefix is
+    configured (back-compat).
+
+  3 new regression cases in `tests/unit/test_theming_cookie_namespace_1158.py`
+  (1 for sub-item (a), 2 for sub-item (b)) plus 7 new JS cases in
+  `tests/js/theming_cookie_namespace_write.test.js` (4 for sub-item (c),
+  3 for sub-item (d)).
+
 - **Tag-registry test isolation + sidecar bridge extension to block /
   assign tags (closes #1167)** — two Stage 11 follow-ups from PR #1166
   (which wired the raw-Python sidecar into ``Node::CustomTag``):
--- a/python/djust/theming/manager.py
+++ b/python/djust/theming/manager.py
@@ -4,13 +4,24 @@ Theme state management for djust.
 Manages theme preset and mode preferences, with session persistence.
 """
 
+import re
 from dataclasses import dataclass
 from typing import Literal
 
 from django.conf import settings
+from django.core.exceptions import ImproperlyConfigured
 from django.http import HttpRequest
 
 from .presets import ThemePreset, get_preset
+
+# #1169(b) — cookie_namespace is interpolated directly into cookie names. RFC
+# 6265 forbids whitespace, control chars, '=' and ';' inside cookie names;
+# non-ASCII is implementation-defined and unsafe in practice. Restrict to
+# ASCII alphanumerics + underscore + hyphen (a strict subset of RFC 6265
+# token chars) so misconfiguration fails loudly at config-load instead of
+# silently emitting malformed Set-Cookie headers that browsers reject or
+# split into multiple cookies.
+COOKIE_NAMESPACE_RE = re.compile(r"^[A-Za-z0-9_-]+$")
 
 ThemeMode = Literal["light", "dark", "system"]
 
@@ -48,11 +59,42 @@ DEFAULT_CONFIG = {
 }
 
 
+def _validate_cookie_namespace(value):
+    """Validate ``LIVEVIEW_CONFIG['theme']['cookie_namespace']`` (#1169(b)).
+
+    The value is interpolated directly into cookie names; characters
+    illegal in cookie names (whitespace, ``=``, ``;``, non-ASCII)
+    silently produce malformed Set-Cookie headers. Validate at
+    config-load so the failure mode is a loud
+    :class:`~django.core.exceptions.ImproperlyConfigured` at startup
+    instead of broken cookies in production.
+
+    ``None`` and empty string are accepted as "no namespace configured".
+    """
+    if value is None or value == "":
+        return value
+    if not isinstance(value, str) or not COOKIE_NAMESPACE_RE.match(value):
+        raise ImproperlyConfigured(
+            f"LIVEVIEW_CONFIG['theme']['cookie_namespace']={value!r} contains "
+            f"characters illegal in cookie names. Use only ASCII letters, "
+            f"digits, underscores, and hyphens."
+        )
+    return value
+
+
 def get_theme_config() -> dict:
-    """Get theme configuration from Django settings."""
+    """Get theme configuration from Django settings.
+
+    Raises:
+        ImproperlyConfigured: if ``cookie_namespace`` contains characters
+            illegal in cookie names (#1169(b)).
+    """
     liveview_config = getattr(settings, "LIVEVIEW_CONFIG", {})
     theme_config = liveview_config.get("theme", {})
-    return {**DEFAULT_CONFIG, **theme_config}
+    merged = {**DEFAULT_CONFIG, **theme_config}
+    # Validate cookie_namespace at config-load (#1169(b)).
+    _validate_cookie_namespace(merged.get("cookie_namespace"))
+    return merged
 
 
 @dataclass
@@ -303,14 +345,29 @@ class ThemeManager:
             prefix = f"{ns}_" if ns else ""
 
             # Read namespaced first; fall back to unprefixed (migration window).
+            #
+            # #1169(a) — distinguish "namespaced cookie not set" from
+            # "namespaced cookie set to empty string". An empty string is a
+            # deliberate value (e.g. user cleared a layout); falling through
+            # to the legacy unprefixed cookie in that case re-introduces
+            # cross-project bleed via the same path #1158 closed.
             def _read(name: str, default: str = "") -> str:
                 if prefix:
-                    return cookies.get(f"{prefix}{name}", cookies.get(name, default))
+                    namespaced = cookies.get(f"{prefix}{name}")
+                    if namespaced is not None:
+                        return namespaced
+                    return cookies.get(name, default)
                 return cookies.get(name, default)
 
-            theme = _read("djust_theme") or None
-            preset = _read("djust_theme_preset") or None
-            pack = _read("djust_theme_pack") or None
+            # _read returns "" when the cookie is set to empty; preserve that
+            # rather than coercing to None via `or None` (which would re-trigger
+            # the bleed-through path the namespaced read just closed).
+            raw_theme = _read("djust_theme")
+            raw_preset = _read("djust_theme_preset")
+            raw_pack = _read("djust_theme_pack")
+            theme = raw_theme if raw_theme else None
+            preset = raw_preset if raw_preset else None
+            pack = raw_pack if raw_pack else None
             layout = _read("djust_theme_layout", "")
             logger.debug(
                 "Cookies (ns=%r): theme=%s, preset=%s, pack=%s, layout=%s",
--- a/python/djust/theming/static/djust_theming/js/theme.js
+++ b/python/djust/theming/static/djust_theming/js/theme.js
@@ -32,6 +32,25 @@
     const COOKIE_KEY_PACK = COOKIE_PREFIX + 'djust_theme_pack';
     const COOKIE_KEY_LAYOUT = COOKIE_PREFIX + 'djust_theme_layout';
 
+    /**
+     * #1169(d) — Delete the legacy unprefixed cookie when writing a
+     * namespaced theming cookie. After a user changes their theme on a
+     * namespaced site, the legacy `djust_theme*` cookies otherwise sit in
+     * the jar forever and can bleed back in if the namespace is ever
+     * removed (or read by another project on the same localhost domain).
+     *
+     * Only fires when COOKIE_PREFIX is non-empty. With no prefix, the
+     * "namespaced" name and the legacy name are identical and deleting it
+     * would clobber the value we just wrote.
+     *
+     * @param {string} legacyName — the unprefixed cookie name to delete
+     *   (e.g. 'djust_theme_pack').
+     */
+    function _deleteLegacyCookie(legacyName) {
+        if (!COOKIE_PREFIX) return;
+        document.cookie = legacyName + '=; path=/; max-age=0; SameSite=Lax';
+    }
+
     class DjustThemeManager {
         constructor() {
             this.pendingUpdate = null;
@@ -186,6 +205,7 @@
 
             // Also set a cookie so the server can read it
             document.cookie = `${COOKIE_KEY_PRESET}=${preset};path=/;max-age=31536000;SameSite=Lax`;
+            _deleteLegacyCookie('djust_theme_preset');
 
             // Clear pack if setting preset manually
             this.clearPack();
@@ -207,6 +227,7 @@
 
             // Set cookie for server
             document.cookie = `${COOKIE_KEY_THEME}=${theme};path=/;max-age=31536000;SameSite=Lax`;
+            _deleteLegacyCookie('djust_theme');
 
             // Clear pack if setting theme manually
             this.clearPack();
@@ -223,6 +244,7 @@
 
             // Set cookie for server
             document.cookie = `${COOKIE_KEY_PACK}=${pack};path=/;max-age=31536000;SameSite=Lax`;
+            _deleteLegacyCookie('djust_theme_pack');
 
             // Reload to apply
             window.location.reload();
@@ -234,6 +256,7 @@
         clearPack() {
             localStorage.removeItem(STORAGE_KEY_PACK);
             document.cookie = `${COOKIE_KEY_PACK}=;path=/;expires=Thu, 01 Jan 1970 00:00:00 GMT`;
+            _deleteLegacyCookie('djust_theme_pack');
         }
 
         /**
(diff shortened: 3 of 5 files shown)


johnrtipton (commented):
## Stage 11 review — PR #1183 (cookie namespace polish, 4 sub-items)

**Verdict: 🟢 LGTM — ship it.**

All four sub-items land correctly. Two-commit shape verified. CHANGELOG matches scope. CI in flight; pre-push reported 11/11 Python and 1460/1460 JS. The implementation is tight and the tests are aimed at the actual failure modes (not just coverage theater).

Findings: 🔴 0 / 🟡 0 / 🟢 5 (all positive observations + 1 nit-question).

---

### 🟢 Sub-item (a) — `_read()` rewrite is correct on all three branches

`python/djust/theming/manager.py` lines 347-370:

```python
def _read(name: str, default: str = "") -> str:
    if prefix:
        namespaced = cookies.get(f"{prefix}{name}")
        if namespaced is not None:
            return namespaced
        return cookies.get(name, default)
    return cookies.get(name, default)
...
raw_theme = _read("djust_theme")
...
theme = raw_theme if raw_theme else None
```

Walked the three required branches:

- `cookies = {"djust_theme": "dark", "myapp_djust_theme": ""}, prefix="myapp_"` → `_read` returns `""` (namespaced is in jar, not None), then the `raw_theme if raw_theme else None` line coerces empty string back to `None` for theme/preset/pack. Crucially, `layout = _read(..., "")` keeps `""` because the layout call site does NOT apply the `or None`-style coalesce. That's the deliberate sub-item-(a) win — empty layout sticks instead of falling back. The new test `test_empty_namespaced_cookie_does_not_fall_back_to_legacy` asserts exactly this on `state.layout`.
- `cookies = {"djust_theme": "dark"}, prefix="myapp_"` → namespaced lookup returns `None`, falls back to legacy `"dark"`. ✓
- `cookies = {}` → both lookups return `default` (`""`), then theme/preset/pack become `None` via the `if raw_x else None` coalesce. ✓

One subtle observation: `theme`/`preset`/`pack` still apply `if raw else None`, so an empty-string namespaced theme/preset/pack STILL falls back to "no value" → triggers any downstream defaulting. That's defensible (you can't really "set" a theme to empty string — it's not a meaningful value), and the test correctly targets `layout` where empty IS a meaningful value. Worth a one-line comment in the source explaining the asymmetry, but not blocking.

### 🟢 Sub-item (b) — regex rejects all dangerous shapes; reject list is thorough

`COOKIE_NAMESPACE_RE = re.compile(r"^[A-Za-z0-9_-]+$")` (manager.py L24)

Verified by hand against the user-spec list:
- `myapp.org` → no match (dot rejected) ✓
- `my app` → space rejected ✓
- `app=v1` → `=` rejected ✓
- `app;v1` → `;` rejected ✓
- `app#1` → `#` rejected ✓
- `app/v1`, `app+v1`, `app:v1`, `app?v1` → all rejected ✓

The reject test list (`test_cookie_namespace_validator_rejects_illegal_characters`) covers space, `=`, `;`, non-ASCII (`café`), tab, newline, `,`, parens — broader than the user-spec required and includes the byte-shape gotchas (`\t`, `\n`).

Empty string and `None` accepted as "no namespace" (manager.py L74) — correct semantics; matches what `theme_head.html` emits when the setting is unset.

The `_validate_cookie_namespace` is called inside `get_theme_config()` (L96), which is the right point — every read of the config validates. One thing to be aware of: the validator runs on EVERY call to `get_theme_config()`, not just at startup. That's a hot path on per-request rendering. The regex is small and the value is small, so it's fine, but the docstring claim of "config-load" is slightly aspirational — it's actually "every config-read". Not blocking; clarify or cache if it ever shows up in profiles.

### 🟢 Sub-item (c) — JSDOM tests genuinely exercise `document.cookie`

`tests/js/theming_cookie_namespace_write.test.js`:

- 4 sub-(c) cases: setPack/setPreset/setLayout with prefix → assert namespaced in `document.cookie`, legacy NOT in jar; plus back-compat with no prefix → assert legacy IS in jar.
- The fixture loads `theme.js` via `dom.window.eval(getThemeJs())` after setting `window.__djust_theme_cookie_prefix`. That's the real binding — `COOKIE_PREFIX` captures at IIFE evaluation time, so prefix MUST be set before eval. The fixture orders this correctly (lines 60-68).
- `parseCookies()` (L72-83) splits on `;` and trims — the standard browser cookie-jar contract. JSDOM honors `Max-Age=0` by removing from the jar, so `expect(jar['djust_theme_pack']).toBeUndefined()` is the correct assertion shape.
- No async callbacks involved (no `startViewTransition`-style stubs), so the CLAUDE.md retro #1113 microtask-yield rule doesn't apply here. The `requestAnimationFrame = (cb) => { cb(0); ... }` synchronous override is fine for these tests because none of the asserted-on cookie writes happen inside an rAF (verified: `setPack`/`setPreset`/`setLayout` write the cookie inline, not inside the rAF batch). `setPresetWithoutReload` DOES write inside an rAF, but that path isn't exercised here.

### 🟢 Sub-item (d) — `_deleteLegacyCookie` is inert without prefix; 6/6 sites confirmed

`theme.js` L49-52:
```javascript
function _deleteLegacyCookie(legacyName) {
    if (!COOKIE_PREFIX) return;
    document.cookie = legacyName + '=; path=/; max-age=0; SameSite=Lax';
}
```

The early-return on falsy `COOKIE_PREFIX` is the correct guard — without it, the cleanup would Max-Age=0 the cookie just written (since the "namespaced" name and the "legacy" name are identical when prefix is empty). The "back-compat: empty prefix" test on L141-149 of the test file verifies this — pre-seeds nothing, calls `setPack`, asserts `jar['djust_theme_pack'] === 'nyc_core'`. If the guard were missing, that cookie would be Max-Age=0'd by the cleanup and the test would fail.

Verified all 6 write sites call `_deleteLegacyCookie` (grep on `theme.js`):
- L207-208 setPreset → `'djust_theme_preset'` ✓
- L229-230 setTheme → `'djust_theme'` ✓
- L246-247 setPack → `'djust_theme_pack'` ✓
- L258-259 clearPack → `'djust_theme_pack'` ✓
- L267-268 setLayout → `'djust_theme_layout'` ✓
- L368-369 setPresetWithoutReload → `'djust_theme_preset'` ✓

The 3 sub-(d) test cases assert on `jar['djust_theme_pack']` being undefined after a Max-Age=0 — JSDOM's cookie store honors that, so the assertion shape is correct. Pre-seeding the legacy cookie before the call (L156, L171) is the right pattern — it would catch a regression where the cleanup dropped silently.

### 🟢 Two-commit integrity, CHANGELOG, no collateral noise

- Commit 1 (`cbc0ee56`): impl + tests; no CHANGELOG hunks. ✓
- Commit 2 (`a8a725b9`): CHANGELOG only (1 file changed, +34 lines). ✓
- CHANGELOG entry calls out all 4 sub-items with issue refs (#1158, #1168, #1169) and accurate test counts (3 Python + 7 JS). ✓
- No collateral edits. The 5 modified files are exactly the 5 expected: `CHANGELOG.md`, `manager.py`, `theme.js`, the new JS test file, and the existing Python test file extended.

---

### Question (non-blocking)

The `theme`/`preset`/`pack` reads still apply the `if raw else None` coalesce that empties out empty-string values, while `layout` is the only field that preserves empty-string. Is that asymmetry intentional and worth documenting next to `_read`? The current code comment ("`_read` returns `""` when the cookie is set to empty; preserve that rather than coercing to None via `or None`...") explains the layout case but doesn't mention that theme/preset/pack DELIBERATELY still coerce. Future readers may "fix" the asymmetry and break the layout test. A two-line comment naming the asymmetry would be nice (or a follow-up issue if you want to keep the diff tight).

---

### Recommendation: merge

The fix is correct, the tests target the actual failure modes, the two-commit shape is clean, and CI is on track. Merge once green.

johnrtipton:
## Stage 11 review — PR #1183 (cookie namespace polish, 4 sub-items)

**Verdict: 🟢 LGTM — ship it.**

All four sub-items land correctly. Two-commit shape verified. CHANGELOG matches scope. CI in flight; pre-push reported 11/11 Python and 1460/1460 JS. The implementation is tight and the tests are aimed at the actual failure modes (not just coverage theater).

Findings: 🔴 0 / 🟡 0 / 🟢 5 (all positive observations + 1 nit-question).

---

### 🟢 Sub-item (a) — `_read()` rewrite is correct on all three branches

`python/djust/theming/manager.py` lines 347-370:

```python
def _read(name: str, default: str = "") -> str:
    if prefix:
        namespaced = cookies.get(f"{prefix}{name}")
        if namespaced is not None:
            return namespaced
        return cookies.get(name, default)
    return cookies.get(name, default)
...
raw_theme = _read("djust_theme")
...
theme = raw_theme if raw_theme else None
```

Walked the three required branches:

- `cookies = {"djust_theme": "dark", "myapp_djust_theme": ""}, prefix="myapp_"` → `_read` returns `""` (namespaced is in jar, not None), then the `raw_theme if raw_theme else None` line coerces empty string back to `None` for theme/preset/pack. Crucially, `layout = _read(..., "")` keeps `""` because the layout call site does NOT apply the `or None`-style coalesce. That's the deliberate sub-item-(a) win — empty layout sticks instead of falling back. The new test `test_empty_namespaced_cookie_does_not_fall_back_to_legacy` asserts exactly this on `state.layout`.
- `cookies = {"djust_theme": "dark"}, prefix="myapp_"` → namespaced lookup returns `None`, falls back to legacy `"dark"`. ✓
- `cookies = {}` → both lookups return `default` (`""`), then theme/preset/pack become `None` via the `if raw_x else None` coalesce. ✓

One subtle observation: `theme`/`preset`/`pack` still apply `if raw else None`, so an empty-string namespaced theme/preset/pack STILL falls back to "no value" → triggers any downstream defaulting. That's defensible (you can't really "set" a theme to empty string — it's not a meaningful value), and the test correctly targets `layout` where empty IS a meaningful value. Worth a one-line comment in the source explaining the asymmetry, but not blocking.

### 🟢 Sub-item (b) — regex rejects all dangerous shapes; reject list is thorough

`COOKIE_NAMESPACE_RE = re.compile(r"^[A-Za-z0-9_-]+$")` (manager.py L24)

Verified by hand against the user-spec list:
- `myapp.org` → no match (dot rejected) ✓
- `my app` → space rejected ✓
- `app=v1` → `=` rejected ✓
- `app;v1` → `;` rejected ✓
- `app#1` → `#` rejected ✓
- `app/v1`, `app+v1`, `app:v1`, `app?v1` → all rejected ✓

The reject test list (`test_cookie_namespace_validator_rejects_illegal_characters`) covers space, `=`, `;`, non-ASCII (`café`), tab, newline, `,`, parens — broader than the user-spec required and includes the byte-shape gotchas (`\t`, `\n`).

Empty string and `None` accepted as "no namespace" (manager.py L74) — correct semantics; matches what `theme_head.html` emits when the setting is unset.

The `_validate_cookie_namespace` is called inside `get_theme_config()` (L96), which is the right point — every read of the config validates. One thing to be aware of: the validator runs on EVERY call to `get_theme_config()`, not just at startup. That's a hot path on per-request rendering. The regex is small and the value is small, so it's fine, but the docstring claim of "config-load" is slightly aspirational — it's actually "every config-read". Not blocking; clarify or cache if it ever shows up in profiles.

### 🟢 Sub-item (c) — JSDOM tests genuinely exercise `document.cookie`

`tests/js/theming_cookie_namespace_write.test.js`:

- 4 sub-(c) cases: setPack/setPreset/setLayout with prefix → assert namespaced in `document.cookie`, legacy NOT in jar; plus back-compat with no prefix → assert legacy IS in jar.
- The fixture loads `theme.js` via `dom.window.eval(getThemeJs())` after setting `window.__djust_theme_cookie_prefix`. That's the real binding — `COOKIE_PREFIX` captures at IIFE evaluation time, so prefix MUST be set before eval. The fixture orders this correctly (lines 60-68).
- `parseCookies()` (L72-83) splits on `;` and trims — the standard browser cookie-jar contract. JSDOM honors `Max-Age=0` by removing from the jar, so `expect(jar['djust_theme_pack']).toBeUndefined()` is the correct assertion shape.
- No async callbacks involved (no `startViewTransition`-style stubs), so the CLAUDE.md retro #1113 microtask-yield rule doesn't apply here. The `requestAnimationFrame = (cb) => { cb(0); ... }` synchronous override is fine for these tests because none of the asserted-on cookie writes happen inside an rAF (verified: `setPack`/`setPreset`/`setLayout` write the cookie inline, not inside the rAF batch). `setPresetWithoutReload` DOES write inside an rAF, but that path isn't exercised here.

### 🟢 Sub-item (d) — `_deleteLegacyCookie` is inert without prefix; 6/6 sites confirmed

`theme.js` L49-52:
```javascript
function _deleteLegacyCookie(legacyName) {
    if (!COOKIE_PREFIX) return;
    document.cookie = legacyName + '=; path=/; max-age=0; SameSite=Lax';
}
```

The early-return on falsy `COOKIE_PREFIX` is the correct guard — without it, the cleanup would Max-Age=0 the cookie just written (since the "namespaced" name and the "legacy" name are identical when prefix is empty). The "back-compat: empty prefix" test on L141-149 of the test file verifies this — pre-seeds nothing, calls `setPack`, asserts `jar['djust_theme_pack'] === 'nyc_core'`. If the guard were missing, that cookie would be Max-Age=0'd by the cleanup and the test would fail.

Verified all 6 write sites call `_deleteLegacyCookie` (grep on `theme.js`):
- L207-208 setPreset → `'djust_theme_preset'` ✓
- L229-230 setTheme → `'djust_theme'` ✓
- L246-247 setPack → `'djust_theme_pack'` ✓
- L258-259 clearPack → `'djust_theme_pack'` ✓
- L267-268 setLayout → `'djust_theme_layout'` ✓
- L368-369 setPresetWithoutReload → `'djust_theme_preset'` ✓

The 3 sub-(d) test cases assert on `jar['djust_theme_pack']` being undefined after a Max-Age=0 — JSDOM's cookie store honors that, so the assertion shape is correct. Pre-seeding the legacy cookie before the call (L156, L171) is the right pattern — it would catch a regression where the cleanup dropped silently.

### 🟢 Two-commit integrity, CHANGELOG, no collateral noise

- Commit 1 (`cbc0ee56`): impl + tests; no CHANGELOG hunks. ✓
- Commit 2 (`a8a725b9`): CHANGELOG only (1 file changed, +34 lines). ✓
- CHANGELOG entry calls out all 4 sub-items with issue refs (#1158, #1168, #1169) and accurate test counts (3 Python + 7 JS). ✓
- No collateral edits. The 5 modified files are exactly the 5 expected: `CHANGELOG.md`, `manager.py`, `theme.js`, the new JS test file, and the existing Python test file extended.

---

### Question (non-blocking)

The `theme`/`preset`/`pack` reads still apply the `if raw else None` coalesce that empties out empty-string values, while `layout` is the only field that preserves empty-string. Is that asymmetry intentional and worth documenting next to `_read`? The current code comment ("`_read` returns `""` when the cookie is set to empty; preserve that rather than coercing to None via `or None`...") explains the layout case but doesn't mention that theme/preset/pack DELIBERATELY still coerce. Future readers may "fix" the asymmetry and break the layout test. A two-line comment naming the asymmetry would be nice (or a follow-up issue if you want to keep the diff tight).

---

### Recommendation: merge

The fix is correct, the tests target the actual failure modes, the two-commit shape is clean, and CI is on track. Merge once green.

johnrtipton:
## Pipeline Retro — PR #1183

**Closes**: #1169. **Drain iter**: 7/8.

### Worked well
- **Stage 11 LGTM clean** (0 🔴, 0 🟡) — first such verdict in the v0.9.2 drain.
- **Sub-item (a) None-vs-empty distinction** caught a class of bug that's easy to miss in code review (the `or None` pattern is ubiquitous Python). The asymmetric retention (layout preserves empty; theme/preset/pack still coalesce) is intentional and matched by the regression test.
- **Sub-item (d) covered all 6 write sites** in one helper invocation per site, mirroring the v0.8.6 retro #1104 "mechanical replacement: N similar sites need N tests" rule.
- Two-commit shape held cleanly.

### Drain progress
v0.9.2: **7 of 8 work units shipped** — #1172, PR #1176, PR #1178, PR #1179, PR #1181, PR #1182, PR #1183.

Final iter: #1171 (data_table row navigation polish — 3 sub-items).

johnrtipton:
## Retrospective — PR #1183 (backfill)

**Note**: Backfill posted during the v0.9.1-8 final cleanup (#1212 audit). PR #1183 shipped without a Stage 14 retro comment — same outer-loop integrity failure mode as v0.9.5 PRs #1203/#1204.

**PR title**: fix(theming): cookie namespace polish — 4 sub-items (closes #1169)

PR #1187, opened 2026-04-28
chore(test-infra): suppress unhandled errors in JS + Python test runtimes

johnrtipton:
## Summary

v0.9.3 milestone — release-blocker for v0.9.0rc3. Three test-runtime warnings/errors that surfaced during local `make test` but never affected production behavior. All unblock the canonical `make test` exit-0 gate.

| Issue | Fix path | Where |
|---|---|---|
| #1186 (P1) — happy-dom + undici WebSocket `dispatchEvent` | Path 1 — filter | `vitest.config.js` (`onUnhandledError`) |
| #1152 (P2) — view-transitions teardown `EnvironmentTeardownError` | Path 1 — filter | `vitest.config.js` (`onUnhandledError`) |
| #1153 (P2) — `_wait_for_one` RuntimeWarning | Path 2 — real lifecycle fix | `python/djust/mixins/template.py` `arender_chunks` |

### Why Path 2 for #1153

`task.cancel()` only signals — it doesn't unblock the `done.get()` queue inside `asyncio.as_completed`'s internal `_wait_for_one`. When `arender_chunks` returned mid-loop on `emitter.cancelled`, the for-protocol's already-pulled coroutine plus any further iterator-yielded coroutines were GC'd unawaited and Python emitted `RuntimeWarning: coroutine '_wait_for_one' was never awaited`.

Fix: explicit `_drain_iterator(as_completed_iter)` after `_cancel_pending()` so the iterator's queue empties cleanly. The drain awaits each remaining `_wait_for_one` coroutine, swallowing self-inflicted `CancelledError`.

### Why Path 1 (filter) for the JS issues

#1186 is two-runtime cross-pollination — undici fires a Node-side `Event` that happy-dom's `EventTarget.dispatchEvent` runtime check rejects. Not fixable on our side without replacing happy-dom or stubbing WebSocket. Filter is narrow: matches `Failed to execute 'dispatchEvent' on 'EventTarget'` AND stack hits `undici/lib/web/websocket` or `happy-dom/lib/event/EventTarget`. Anything else still re-throws.

#1152's stubs already yield a microtask per CLAUDE.md retro #1113, so the diagnosis was RPC-timing teardown noise, not a stub regression. Same `onUnhandledError` filter handles both.

### Verification — `make test` exit 0

Three consecutive `make test` runs post-fix exit 0 (was non-deterministic 1-3 unhandled errors out of 1463 passing JS tests).

```
Python: 4047 passed, 18 skipped (regression test +1)
[...]

Issue #1175, opened 2026-04-27
tech-debt: CSP-strict defaults for new client-side framework code (no inline scripts, no inline event handlers)


johnrtipton:
Source: v0.9.1 retro / Action Tracker #183 / PRs #1163 + #1170.

## Context

Two converging pieces of evidence in the v0.9.1 drain:
- **PR #1163** added CSP-nonce-aware activator for `<dj-lazy-slot>` fills (read `request.csp_nonce` from middleware, thread through inline `<script nonce="...">`).
- **PR #1170** went further — chose to skip inline scripts entirely in favor of an external static JS module (`data-table-row-click.js`) that auto-binds on a marker class.

The 'external module + auto-bind on marker class' shape is strictly more CSP-friendly than 'inline script + nonce attribute' — it works under stricter CSP policies (no `'unsafe-inline'` ever needed, even with nonces) and is simpler to reason about.

## Proposed canonicalization

Add to `docs/PULL_REQUEST_CHECKLIST.md` and `CLAUDE.md` design rules:

> **CSP-strict defaults for new client-side code.** Any new framework feature that emits HTML must default to:
> 1. No inline `<script>` blocks (use external static JS modules).
> 2. No inline event handlers (`onclick=`, `onchange=`, etc.).
> 3. Auto-bind via marker class + delegated listener on `document` or root.
> 4. CSP-nonce propagation only when an inline script is genuinely required (rare; lazy-fill case in #1163 is the canonical exception).

This positions strict-CSP deployments as a v1.0 design constraint, not an opt-in.

## Scope

CLAUDE.md addition + PULL_REQUEST_CHECKLIST.md addition + a one-paragraph note in `docs/website/guides/security.md` (or wherever CSP is documented).

## Why deferred from v0.9.1

The pattern emerged organically from PRs that already shipped. Canonicalizing for v1.0 readiness.

Issue #1124, opened 2026-04-26
tech-debt: CodeQL js/tainted-format-string self-review checkpoint


johnrtipton:
**Source**: v0.8.6 milestone retro finding #4 (PR #1120 post-CI catch).

`console.error(\`[label] msg ${userControlledValue}:\`, e)` triggers CodeQL js/tainted-format-string when `userControlledValue` derives from DOM attributes, server frame fields, or request body. Use parameterized form: `console.error('[label] msg %s:', userControlledValue, e)` — pulls the dynamic value out of the format string entirely.

Add to CLAUDE.md JS-side patterns + Stage 7 self-review checklist. The bundled `client.js` line numbers in CodeQL reports also make the trail harder than necessary; consider source maps in the CodeQL workflow.

johnrtipton:
Closing — canonicalized in CLAUDE.md §v0.8.6 "CodeQL js/tainted-format-string self-review checkpoint (#1124)". The %s parameterized form vs template-literal interpolation is documented as a Stage 7 grep target. 🤖 v0.9.1-6 backlog audit.

PR #1190, opened 2026-04-28
feat(hot-reload): auto-enable HVR in DEBUG via DjustConfig.ready()

johnrtipton:
## Summary

- djust now auto-calls `enable_hot_reload()` from its own `DjustConfig.ready()` whenever `DEBUG=True` and `watchdog` is installed — drops the AppConfig.ready() boilerplate downstream consumers needed to wire up HVR.
- Existing per-consumer `enable_hot_reload()` calls keep working (idempotent via `hot_reload_server.is_running()`). Opt out via `LIVEVIEW_CONFIG['hot_reload_auto_enable']: False`. Test runs auto-skip via `PYTEST_CURRENT_TEST`.
- Two-commit shape: impl+tests / docs+CHANGELOG.

## Why

The framework's HVR (state-preserving `__class__` swap, shipped v0.6.1) is strictly better than process restart — it preserves form input, scroll position, counters, and active tabs across `.py` edits. But every downstream consumer needed a one-line `enable_hot_reload()` call in their AppConfig.ready() to wire it up, and several silently skipped that step:

- `docs.djust.org` rolled its own dev loop via `watchfiles` wrapping `uvicorn` in its `Makefile` (`make dev` target). That's process restart — drops the WebSocket and kills view state on every save.
- `djust.org`, `djustlive`, and the `djust` framework's own `examples/demo_project` rely on config defaults but only `demo_project` actually called `enable_hot_reload()`.

Drop the integration step. Make HVR the default in DEBUG.

## Gating layers (defense in depth)

`enable_hot_reload()` already short-circuits on:
1. `settings.DEBUG=False` — production safe by construction.
2. `config.get("hot_reload", True) == False` — explicit opt-out.
3. `WATCHDOG_AVAILABLE=False` — production builds don't install watchdog (it's a `[dev]` extra).
4. `hot_reload_server.is_running()` — idempotent.

This PR adds two NEW gates:
5. `config.get("hot_reload_auto_enable", True)` — new opt-out knob, default `True`.
6. `os.environ.get("PYTEST_CURRENT_TEST")` — skip during pytest to avoid spawning a watchdog thread per test.

## Files

- `python/djust/apps.py` — auto-enable call appended to `ready()` with try/except isolation matching the observability-handler pattern already there.
- `python/djust/config.py` — new `hot_reload_auto_enable: True` default.
- `python/djust/__init__.py` — `enable_hot_reload()` docstring updated.
- `examples/demo_project/demo_app/apps.py` — drops the now-unnecessary explicit call.
- `python/djust/tests/test_auto_hot_reload.py` — 6 new cases.
- `CHANGELOG.md`, `CLAUDE.md`, `docs/website/guides/hot-view-replacement.md` — docs.

## Test plan

- [x] `pytest python/djust/tests/test_auto_hot_reload.py -v` — 6 cases green
- [x] `make test` — 4047 Python + 1463 JS passing
- [x] `ruff check` + `ruff format` clean
- [x] Manual: `cd examples/demo_project && make start` — `[HotReload] Hot reload enabled` printed at startup; edits preserve view state
- [ ] Follow-up PR: migrate `docs.djust.org/Makefile` from `watchfiles` wrapping `uvicorn` to plain `uvicorn` after submodule bump to a release containing this change

## Out of scope

- Auto-enable in non-DEBUG production builds (DEBUG gate is non-negotiable for safety).
- `watchfiles` as an alternative file-watcher backend (djust uses `watchdog`).
- Reloading on `settings.py` / `INSTALLED_APPS` changes (HVR is for view code only).
- Removing `enable_hot_reload()` from the public API (kept for explicit-control use cases).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,31 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- **HVR auto-enabled in DEBUG (no AppConfig.ready() boilerplate
+  required)** — djust's own `DjustConfig.ready()` now auto-calls
+  `enable_hot_reload()` whenever `DEBUG=True` and `watchdog` is
+  installed. Existing per-consumer `enable_hot_reload()` calls keep
+  working unchanged (idempotent via `hot_reload_server.is_running()`).
+  Opt out via `LIVEVIEW_CONFIG['hot_reload_auto_enable']: False` for
+  projects that orchestrate the file watcher externally. Test runs
+  auto-skip via `PYTEST_CURRENT_TEST` so pytest sessions don't spawn
+  a watchdog thread per test. Files: `python/djust/apps.py` (auto-enable
+  call appended to `ready()`), `python/djust/config.py`
+  (new `hot_reload_auto_enable: True` default),
+  `python/djust/__init__.py` (docstring update). 6 new regression
+  cases in `TestAutoEnableHotReload` (file
+  `python/djust/tests/test_auto_hot_reload.py`). Drops the one-line
+  `enable_hot_reload()` call from
+  `examples/demo_project/demo_app/apps.py`. Closes the friction
+  observed across downstream consumers (docs.djust.org, djust.org,
+  djustlive) that were either rolling their own `watchfiles`
+  process-restart wrappers or silently missing the integration step
+  altogether — the framework's HVR is strictly better than process
+  restart (preserves view state, scroll position, form input across
+  edits) but the consumer-side integration step was easy to skip.
+
 ## [0.9.0rc3] - 2026-04-28
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,8 +20,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   a watchdog thread per test. Files: `python/djust/apps.py` (auto-enable
   call appended to `ready()`), `python/djust/config.py`
   (new `hot_reload_auto_enable: True` default),
-  `python/djust/__init__.py` (docstring update). 6 new regression
-  cases in `TestAutoEnableHotReload` (file
+  `python/djust/__init__.py` (docstring update). 6 new cases covering
+  auto-fire, opt-out config, pytest-env skip, idempotency, exception
+  isolation, and other-setup completion (new file
   `python/djust/tests/test_auto_hot_reload.py`). Drops the one-line
   `enable_hot_reload()` call from
   `examples/demo_project/demo_app/apps.py`. Closes the friction
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -201,6 +201,17 @@ The Rust template engine supports **all 57 Django built-in filters** in `crates/
 - **Ruff F509**: `%`-format strings containing CSS semicolons trigger false positives. Separate HTML (`%s` substitution) from CSS (static string) and concatenate.
 - **VDOM form values**: Ensure form field values are preserved during updates. See `VDOM_PATCHING_ISSUE.md`.
 - **Pre-commit reformatting**: If commit fails due to ruff auto-format, re-stage and commit again.
+- **Hot reload integration (v0.9.0+)**: djust auto-enables HVR from its
+  own `DjustConfig.ready()` whenever `DEBUG=True` and `watchdog` is
+  installed. Downstream consumers do NOT need to add
+  `enable_hot_reload()` to their own `AppConfig.ready()`. Existing
+  explicit calls keep working (idempotent). Opt out via
+  `LIVEVIEW_CONFIG['hot_reload_auto_enable']: False`. Tests skip the
+  auto-enable via `PYTEST_CURRENT_TEST` so pytest sessions don't spawn
+  a watchdog thread per test. Don't wrap `uvicorn` in
+  `watchfiles` / `--reload` for djust dev servers — that's process
+  restart and drops view state; djust's HVR is strictly better
+  (preserves form input, scroll position, counters).
 
 ## Process canonicalizations from PR retros (2026-04-26 View Transitions arc)
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -294,7 +294,7 @@ dependencies = [
 
 [[package]]
 name = "djust_components"
-version = "0.9.0-rc.2"
+version = "0.9.0-rc.3"
 dependencies = [
  "ahash",
  "criterion",
@@ -309,7 +309,7 @@ dependencies = [
 
 [[package]]
 name = "djust_core"
-version = "0.9.0-rc.2"
+version = "0.9.0-rc.3"
 dependencies = [
  "ahash",
  "criterion",
@@ -324,7 +324,7 @@ dependencies = [
 
 [[package]]
 name = "djust_live"
-version = "0.9.0-rc.2"
+version = "0.9.0-rc.3"
 dependencies = [
  "bincode",
  "criterion",
@@ -349,7 +349,7 @@ dependencies = [
 
 [[package]]
 name = "djust_templates"
-version = "0.9.0-rc.2"
+version = "0.9.0-rc.3"
 dependencies = [
  "ahash",
  "chrono",
@@ -370,7 +370,7 @@ dependencies = [
 
 [[package]]
 name = "djust_vdom"
-version = "0.9.0-rc.2"
+version = "0.9.0-rc.3"
 dependencies = [
  "ahash",
  "criterion",
--- a/docs/website/guides/hot-view-replacement.md
+++ b/docs/website/guides/hot-view-replacement.md
@@ -17,29 +17,16 @@ Gated on `DEBUG=True` and enabled by default. Zero cost in production.
 
 ## Quick start
 
-Hot View Replacement (HVR) rides on top of the existing
-`enable_hot_reload()` machinery. If you already enabled hot reload, HVR
-is already working — no new config needed.
+**Auto-enabled in DEBUG since v0.9.0.** Add djust to `INSTALLED_APPS`,
+keep `DEBUG=True`, install `watchdog`, and HVR works. No
+`AppConfig.ready()` boilerplate needed.
 
 ```python
-# myapp/apps.py
-from django.apps import AppConfig
-
-class MyAppConfig(AppConfig):
-    name = "myapp"
-
-    def ready(self):
-        from django.conf import settings
-        if settings.DEBUG:
-            from djust import enable_hot_reload
-            enable_hot_reload()
-```
-
-```python
-# settings.py (defaults shown)
+# settings.py (defaults shown — all True out of the box)
 LIVEVIEW_CONFIG = {
-    "hot_reload": True,    # file watcher on
-    "hvr_enabled": True,   # v0.6.1 — state-preserving reload
+    "hot_reload": True,                # file watcher on
+    "hot_reload_auto_enable": True,    # call enable_hot_reload() from DjustConfig.ready()
+    "hvr_enabled": True,               # v0.6.1 — state-preserving reload
 }
 ```
 
@@ -48,6 +35,34 @@ it's missing:
 
 ```bash
 pip install watchdog
+```
+
+### Manual opt-in / advanced control
+
+The auto-enable is on by default. If you orchestrate the file watcher
+externally (e.g. `watchfiles` wrapping `uvicorn`, or a custom dev
+loop), opt out with:
+
+```python
+# settings.py
+LIVEVIEW_CONFIG = {"hot_reload_auto_enable": False}
+```
+
+You can still call `enable_hot_reload()` yourself from any AppConfig.
+The function is idempotent — calling it manually is a safe no-op when
+the server is already running, so explicit calls coexist with the
+auto-enable.
+
+```python
+# myapp/apps.py — only needed if hot_reload_auto_enable is False
+from django.apps import AppConfig
+
+class MyAppConfig(AppConfig):
+    name = "myapp"
+
+    def ready(self):
+        from djust import enable_hot_reload
+        enable_hot_reload()
 ```
 
 ## What gets preserved
--- a/examples/demo_project/demo_app/apps.py
+++ b/examples/demo_project/demo_app/apps.py
@@ -2,10 +2,9 @@ from django.apps import AppConfig
 
 
 class DemoAppConfig(AppConfig):
-    default_auto_field = 'django.db.models.BigAutoField'
-    name = 'demo_app'
+    default_auto_field = "django.db.models.BigAutoField"
+    name = "demo_app"
 
-    def ready(self):
-        """Initialize app - enable hot reload in development."""
-        from djust import enable_hot_reload
-        enable_hot_reload()
+    # v0.9.0+ — djust auto-enables hot reload via its own DjustConfig.ready()
+    # whenever DEBUG=True and watchdog is installed. No explicit
+    # enable_hot_reload() call needed here.
--- a/python/djust/__init__.py
+++ b/python/djust/__init__.py
@@ -82,7 +82,19 @@ def enable_hot_reload():
     for changes. When a change is detected, all connected WebSocket clients are sent
     a reload message, triggering an automatic page refresh.
 
-    Usage:
+    Auto-enabled by default (since v0.9.0):
+        djust's own ``DjustConfig.ready()`` auto-calls this whenever
+        ``DEBUG=True`` and the ``watchdog`` package is installed. You no
+        longer need to call it explicitly from your own ``AppConfig.ready()``.
+        The function is idempotent — calling it manually is a safe no-op
+        when the server is already running, so existing per-consumer calls
+        keep working unchanged.
+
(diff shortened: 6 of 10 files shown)


johnrtipton (commented):
## Review: PR #1190 — auto-enable HVR in DEBUG via DjustConfig.ready()

Verdict: 🟡 **APPROVE with non-blocking findings** — the wiring is correct, gates are well-ordered, isolation is right, and the 6 tests all pass. Two findings worth addressing before merge or in a fast follow-up; the rest are suggestions.

---

### What was done well

- **Gate ordering is correct and cheapest-first**: env-var (no import) → config-knob → `enable_hot_reload()` (which has its own DEBUG / config / watchdog / `is_running()` gates). Defense-in-depth without redundant cost.
- **Try/except isolation matches the observability handler block above it** — same pattern, same `noqa: BLE001`, same warning-level log. Good consistency.
- **`%s`-style logging**, no f-string interpolation in the new log call — passes the CLAUDE.md retro #1124 / `js/tainted-format-string` checkpoint analogue.
- **Idempotency is layered correctly**: the `hot_reload_server.is_running()` guard at `python/djust/__init__.py:161` was already there, and the test verifies a second `ready()` call doesn't call `start()` twice.
- **Docstring update at `python/djust/__init__.py:77` keeps the manual-usage section** for opt-out cases — correctly framed as "auto-enabled by default" first, with manual usage clearly marked "advanced — only needed if auto-enable is disabled."
- **CHANGELOG count check passes**: `scripts/check-changelog-test-counts.py` exits 0. The `6 new regression cases` phrasing is correctly classified as a delta by `DELTA_PREFIX_RE` (the `\bnew\s+$` lookbehind), so no spurious total-vs-actual diff.
- **CLAUDE.md "Common Pitfalls" addition** correctly warns downstream consumers off `watchfiles`-wrapping-uvicorn — that's exactly the friction this PR removes.

---

### Findings

#### 🟡 Important #1 — CHANGELOG references a class that doesn't exist

`CHANGELOG.md:23-24` says:

> 6 new regression cases in `TestAutoEnableHotReload` (file `python/djust/tests/test_auto_hot_reload.py`)

The test file has **no `TestAutoEnableHotReload` class** — all 6 tests are top-level `def test_*` functions (verified: `python/djust/tests/test_auto_hot_reload.py:70,85,99,108,136,155`).

The CLAUDE.md retro #1106 canonicalization says "name the CLASS, not the file" — but here the class named in the CHANGELOG **does not exist**. This is a documentation lie, not a hook failure (the script doesn't validate class names today, only file totals).

Two clean fixes:

1. Drop the class reference: `6 new regression cases in python/djust/tests/test_auto_hot_reload.py` (the file is brand-new so the total = the delta — no risk of tripping #1106's "claim about the FILE's total count" failure mode).
2. Or actually wrap the 6 functions in a `class TestAutoEnableHotReload:` to match the CHANGELOG. (Pure-functional pytest is fine; class is optional cosmetic.)

I'd take option 1 — less code churn and the brand-new-file phrasing is unambiguous.

#### 🟡 Important #2 — `test_ready_completes_other_setup_even_when_auto_enable_skipped` is a tautology under test-suite ordering

`python/djust/tests/test_auto_hot_reload.py:155-170` asserts that after `app.ready()`, a `DjustLogSanitizerFilter` is present on `logging.getLogger("djust")`. Problem: every other test in this file ALSO calls `app.ready()` (lines 81, 95, 104, 128/129, 150) and `DjustConfig.ready()` calls `addFilter()` unconditionally (no idempotency guard at `python/djust/apps.py:20`). So by the time this test runs, the filter is on the logger from any prior `ready()` call — the assertion would pass even if THIS test's `ready()` did nothing.

Recommended fix: snapshot the filter list before the call, assert at least one MORE filter present after:

```python
djust_logger = logging.getLogger("djust")
filter_count_before = sum(
    1 for f in djust_logger.filters if isinstance(f, DjustLogSanitizerFilter)
)
app.ready()
filter_count_after = sum(
    1 for f in djust_logger.filters if isinstance(f, DjustLogSanitizerFilter)
)
assert filter_count_after > filter_count_before
```

(The underlying filter-accumulation is pre-existing, not introduced by this PR. But this PR adds the test that would catch a regression of the existing setup, and right now that test can't actually catch one. Alternatively: add an `if not any(isinstance(f, DjustLogSanitizerFilter) for f in djust_logger.filters):` guard in `apps.py:20` — that fixes the latent leak AND makes the test meaningful, but is out of scope for this PR.)

#### 🟢 Suggestion — `logger.exception()` over `logger.warning("... %s", e)` for the auto-enable failure path

`python/djust/apps.py:50-52` logs at warning level with just the exception message. If `enable_hot_reload()` raises something non-trivial (e.g. an import chain failure, watchdog version skew, channels misconfiguration), the operator sees only the exception's `__str__` — no traceback. The block above (observability handler install) uses the same pattern, so this matches existing style — but for a feature that's now default-on across every downstream consumer, traceback context would speed up debugging the rare failure case.

Either:
- Keep current style (matches above) — defensible, just noting the downside.
- `logger.exception("[HotReload] auto-enable in DjustConfig.ready() failed: %s", e)` — same message, plus traceback.

#### 🟢 Suggestion — the `_no_pytest_env()` context manager is a clever fix; worth a brief comment in `apps.py` referencing it

The pytest env-var skip at `python/djust/apps.py:41` is the ONLY thing keeping the watchdog thread out of test sessions. The test reasoning (pytest re-sets `PYTEST_CURRENT_TEST` between fixture and test body, so the `_no_pytest_env()` context manager has to wrap the actual `ready()` call) is non-obvious and lives only in the test docstring at lines 41-48. Future maintainers reading `apps.py` won't know why the env-var check exists. Consider a one-line comment:

```python
# Skip during pytest runs to avoid spawning a watchdog thread for
# every test session. (Tests that exercise this branch use a
# context manager to clear PYTEST_CURRENT_TEST temporarily.)
```

(The first sentence is already there at lines 37-38; just add the parenthetical pointer.)

---

### Spot-checked concerns from your prompt that I'd call clean

- **Order of gates** — env → config → enable_hot_reload — correct, cheapest-first.
- **Try/except isolation** — `noqa: BLE001` justified, no risk of breaking Django startup.
- **Inline `import os`** — fine, scoped to the block that uses it.
- **Logging level/format** — matches the observability block above. Good.
- **`hot_reload_auto_enable` placement in config.py** — at line 86, immediately after the other `hot_reload*` defaults (lines 79-82). Good locality. Comment clearly explains the opt-out scenario.
- **Docstring auto-enable note prominence** — yes, "Auto-enabled by default (since v0.9.0):" is the first section after the one-line summary; manual usage is correctly marked "advanced." Good.
- **`fresh_config` fixture teardown** — uses `config._config = snapshot` to restore, where `snapshot = config.as_dict()` is a shallow copy. Safe for the one test that uses it (only writes a top-level scalar). A nested-dict mutation via dot-notation `set` would leak, but no test does that here.
- **Idempotency test demo_project DEBUG=True override** — defensible. pytest-django defaults `DEBUG=False` regardless of `demo_project/settings.py`'s `DEBUG=True` (verified: line 11 of settings.py says True, but pytest-django overrides). Forcing `settings.DEBUG=True` via the fixture is the only way to reach the `is_running()` short-circuit — the test wouldn't be meaningful otherwise. The test scope is the `is_running()` guard, not the DEBUG gate, so this is fine.
- **Security — DEBUG=True in prod**: even if DEBUG is accidentally True in prod, the watchdog absence guard suffices (production typically wouldn't have `watchdog` installed; if it does, the worst case is a file watcher thread + reload broadcasts on file changes — operationally messy but not a security exposure). The four-layer defense (DEBUG / config.hot_reload / config.hot_reload_auto_enable / watchdog) is appropriate.
- **Performance — circular import risk on `from djust import enable_hot_reload`** — none. By the time `DjustConfig.ready()` runs, `djust/__init__.py` has fully executed (Django runs `ready()` after AppConfig instantiation, which itself happens after the package imports). The inline import is just a `sys.modules` dict lookup.
- **Demo app `apps.py` parses cleanly** — verified via `ast.parse`. Comment correctly explains the why.
- **All 6 tests pass** — verified locally: `6 passed in 0.06s`.

---

### Recommendation

**APPROVE.** The two important findings are documentation/test-quality polish, not correctness defects. The implementation itself is sound and the change ships strictly-better dev-mode behavior to every downstream consumer. Address findings #1 (CHANGELOG class name) and #2 (filter-presence test tautology) in this PR before merge if convenient — both are <5 minute fixes — or as a follow-up issue if the merge window is tight.

APPROVE

johnrtipton (commented):
## Stage 11 re-review (commit `4e9e3028`)

All four findings from the prior review are addressed correctly. Verified each against the diff and re-ran the test file in three configurations (file-as-a-whole, the new tautology-fix in isolation, and the swallowing test in isolation) — all 6 tests pass in every configuration.

### 1. CHANGELOG class reference — FIXED
`CHANGELOG.md` line 23 now reads `6 new cases covering auto-fire, opt-out config, pytest-env skip, idempotency, exception isolation, and other-setup completion (new file …)`. The non-existent `TestAutoEnableHotReload` class reference is gone, and the `scripts/check-changelog-test-counts.py` pre-push hook passes silently against the rewritten line. Conforms to the canonicalized rule from PR #1106 (name the CLASS, not the file — and absent a class, describe what the cases cover).

### 2. Tautology test — FIXED
`test_ready_completes_other_setup_even_when_auto_enable_skipped` now snapshots `before = sum(1 for f in djust_logger.filters if isinstance(f, DjustLogSanitizerFilter))` BEFORE `app.ready()` and asserts `after == before + 1`. This proves THIS test's own `ready()` call installed exactly one filter, and would catch a future regression where `ready()` silently no-ops. Confirmed passing both in isolation (`before == 0, after == 1`) and as the 6th of 6 in the file (`before == 5, after == 6`). The `from djust.security import DjustLogSanitizerFilter` import was hoisted to the top of the function body — safe because the module is already transitively imported via `from djust.apps import DjustConfig` at file top, and there's no init-time side-effect dependency on `ready()` running first.

### 3. logger.exception() with exc_info verification — FIXED
`apps.py:53-56` now uses `logging.getLogger("djust").exception("[HotReload] auto-enable in DjustConfig.ready() failed")` (no `%s, e` interpolation; the traceback is attached automatically via `exc_info=True` set by `Logger.exception()`). The test was upgraded to:
- `caplog.at_level(logging.ERROR, …)` (logger.exception logs at ERROR)
- A list comprehension `matching = [r for r in caplog.records if "auto-enable" in r.message]`
- `assert any(r.exc_info for r in matching), "expected exc_info on the log record"`

Verified that pytest's `caplog` populates `record.message` (via `LogCaptureHandler.emit()` calling `super().emit()` which formats the record), so the `r.message` lookup is reliable. The test would fail if anyone reverted to `logger.warning(...)` — `exc_info` would be None.

### 4. Pointer comment to `_no_pytest_env()` — FIXED
`apps.py:38-42` now reads:
> Skip during pytest runs to avoid spawning a watchdog thread for every test session — pytest sets `PYTEST_CURRENT_TEST` for the duration of every test invocation. (Tests that need to exercise the auto-enable path itself temporarily clear this env var; see `_no_pytest_env()` in `python/djust/tests/test_auto_hot_reload.py`.)

Reader-friendly and gives the next maintainer a direct breadcrumb to the test-side helper.

### Test runs

```
$ pytest python/djust/tests/test_auto_hot_reload.py -v
============================== 6 passed in 0.05s ==============================

$ pytest python/djust/tests/test_auto_hot_reload.py::test_ready_completes_other_setup_even_when_auto_enable_skipped -v
============================== 1 passed in 0.04s ==============================

$ pytest python/djust/tests/test_auto_hot_reload.py::test_ready_swallows_enable_hot_reload_exceptions -v
============================== 1 passed in 0.05s ==============================
```

### Sanity pass on the diff

No new defects introduced. The only structural change beyond the four findings is the `DjustLogSanitizerFilter` import hoisted from after-`ready()` to before-`ready()` in test #6 — verified safe (transitive import already in place, no init-order dependency). No new logging, no new exception swallowing, no test sleep, no conftest churn.

### Status

APPROVE

johnrtipton:
# PR #1190 Retro — auto-enable HVR in DEBUG via DjustConfig.ready()

**Merged**: 2026-04-28
**Quality rating**: 4/5 (one self-review miss caught at Stage 11)
**Stages**: Plan → Inventory → Tests → Self-Review → Security → Docs → Commit/PR → Code Review → Address Findings → Re-Review → Merge

## What worked well

- **Two-commit shape held.** Impl+tests in commit 1, docs+CHANGELOG in commit 2. PR diff was easy to review by stage.
- **Layered gating defense in depth.** The new auto-enable rides on top of the 4 existing gates in `enable_hot_reload()` (DEBUG, config, watchdog, is_running). The two new gates (`hot_reload_auto_enable` config, `PYTEST_CURRENT_TEST` env) are additive — production safety remained on the floor of the existing logic, not on the new code.
- **Plan Mode caught scope correctly.** Phase 1 exploration of `apps.py`, `__init__.py`, `dev_server.py` surfaced the four existing gates upfront, so the plan didn't propose redundant new ones.
- **Reused existing patterns.** The try/except around the auto-enable call mirrors the observability-handler block right above it in `ready()` — same shape, same fallback path. No new error-handling pattern to learn.
- **Independent code reviewer caught two real defects** that the self-review missed. Both were judgment calls, not regex findings — exactly what an independent eye is for.

## What didn't work

- **Tautology test slipped past self-review.** `test_ready_completes_other_setup...` asserted `any(isinstance(filters, DjustLogSanitizerFilter))` — but every prior test in the file also calls `app.ready()`, so by the time test #6 runs, the filter is already on the logger. The assertion would pass even if test #6's own `ready()` call did nothing. Fix: snapshot `before` count, assert `after == before + 1`. **Generalizable lesson**: when a test asserts "this thing happened", check whether it would *also* pass without the action under test running. If yes, it's a tautology. The failure mode is silent — the test stays green forever even after the function under test silently breaks.

- **CHANGELOG referenced a non-existent class** (`TestAutoEnableHotReload`) for the 6 tests in the file (which uses bare `def test_*` functions). The pre-push hook's "name the CLASS, not the file" canonicalization (#1106) doesn't catch the reverse mistake — naming a class that doesn't exist. Fix was easy (describe what the cases cover instead). **Generalizable lesson**: if you're naming a class in a CHANGELOG entry, grep for the class definition first.

- **Tests passed first-try only after fixture rework.** Initial fixture used `os.environ.pop("PYTEST_CURRENT_TEST")` in a yield-fixture, expecting that to suppress the env var for the test body. But pytest re-sets `PYTEST_CURRENT_TEST` between fixture teardown and the `(call)` phase, so the pop didn't survive into the body. Fix: use a context-manager pattern inside the test body. Took two debug iterations to diagnose.

- **DEBUG=False under pytest broke the idempotency test.** demo_project ships `DEBUG=True` in `settings.py`, but pytest-django evidently overrides it to `False` for test runs (logical — tests shouldn't run with DEBUG behavior). The first version of `test_ready_is_idempotent_via_is_running_guard` patched `hot_reload_server.is_running` but `enable_hot_reload()` returned early at the DEBUG gate before reaching the check. Fix: use pytest-django's `settings` fixture to set `DEBUG=True` in-test. Defensible — we're explicitly testing the path that DEBUG=True enables.

## Surprises

- **`addFilter()` has no idempotency guard.** Every call to `DjustConfig.ready()` in tests stacks another `DjustLogSanitizerFilter` on the `djust` logger. This is fine (filters are cheap and chained), but it's the load-bearing detail behind the tautology in test #6. Worth noting if anyone ever wants to refactor the log-sanitizer filter installation.

## Should-fix follow-ups

- **docs.djust.org Makefile migration** — drop the `watchfiles` wrapper around `uvicorn` once a release containing this PR is published and the submodule is bumped. This is the dogfooding payoff. Tracked as a follow-up rather than included here because docs.djust.org is a separate repo with a submodule pin to djust.

## Knowledge base updates

None required — the lessons above (tautology test pattern, class-name-grep on CHANGELOG, pytest env-var lifecycle) are generalizable but the existing CLAUDE.md "Common Pitfalls" already covers the broader theme. Adding entries for these specific cases would dilute the canon. If this same defect class shows up in another PR's retro, promote it then.

## Review stats

| Metric | Value |
|--------|-------|
| Tests added | 6 |
| 🔴 findings | 0 |
| 🟡 findings | 2 (CHANGELOG class ref, tautology test) |
| 🟢 suggestions | 2 (`logger.exception`, `_no_pytest_env` comment) |
| Findings fixed pre-merge | 4 of 4 |
| CI failures | 0 |
| Commits | 3 (impl+tests, docs, address-findings) |

## What I'd do differently

- **Run the test in isolation _and_ as part of the suite during self-review.** The tautology would have surfaced if I'd thought "would this assertion pass if the action didn't happen?" before considering the test done. A 30-second sanity check.
- **Grep for class names referenced in CHANGELOG.** Mechanical check: any backtick-quoted CamelCase symbol in a CHANGELOG entry should `grep -r "class CamelCase" python/`. Could be a pre-push hook addition.

johnrtipton:
## Retrospective — PR #1190 (backfill)

**Note**: This retrospective is a **backfill** posted during the v0.9.1-8 final cleanup (#1212 audit pipeline-bypass merges). PR #1190 shipped without a Stage 14 retro comment — same outer-loop integrity failure mode as v0.9.5 PRs #1203 and #1204. Audit ran 2026-04-30; this is best-effort retroactive coverage, not a real-time retro.

**PR title**: feat(hot-reload): auto-enable HVR in DEBUG via DjustConfig.ready()

Issue #1109, opened 2026-04-26
tech-debt: dynamic test fixture pattern — type(name, bases, dict) over class mutation


johnrtipton:
**Source**: PR #1107 retro (https://github.com/djust-org/djust/pull/1107#issuecomment-4322729335)

When a test fixture needs to vary class-level state, `type(self).x = y` in `__init__` mutates a shared object across all instances. PR #1107 had this in test_wizard_rendered_fields.py — every `_W()` instance smashed `_W.wizard_steps`. Tests passed because pytest runs sequentially, but state could leak across orderings or with parallelization.

**Cleaner pattern**:
```python
def _make_view(steps_override):
    cls = type('_TestWizardView', (WizardMixin, _StubLiveView),
               {'wizard_steps': steps_override})
    return cls()
```
Each call returns an instance of a fresh subclass — no cross-test mutation.

**Action**: add to CLAUDE.md (testing patterns) or pipeline-run subagent prompt as a self-review checkpoint.

johnrtipton:
Canonicalized in PR #1116 (CLAUDE.md / Process canonicalizations from PR retros section).

PR #1193, opened 2026-04-28
feat(time-travel): wire-protocol exposure for branched timelines + per-component scrubbing (PR-A for #1151)

### LINKED ISSUE #1151: feat: Debug panel UI for per-component scrubbing + forward-replay

johnrtipton:
## Summary

Server-side wire-protocol exposure that the v0.9.4 debug panel UI (PR-B, follow-up) will consume. The Python plumbing for per-component time-travel (#1041) and forward-replay through branched timelines (#1042) shipped in v0.9.0; this PR exposes the missing wire fields.

This is **PR-A only** — wire-protocol + handlers + tests. **PR-B** is the debug panel UI built on top.

### `time_travel_state` ack frame — 3 new additive fields
- `branch_id`: defaults `"main"`; new branches allocated as `branch-{N}` on forward-replay from a non-tip cursor.
- `forward_replay_enabled`: true iff cursor not at the buffer tip.
- `max_events`: configured ring-buffer cap, surfaced for "X / max" UI.

### `time_travel_event` per-event push — 2 new additive fields
- `branch_id` (mirrors ack-frame field).
- top-level `components` mirror of `entry.state_after.__components__`.

### New handlers
- `time_travel_component_jump(index, component_id, which)` — scrubs SINGLE component without touching parent or other components. Backed by new `restore_component_snapshot()` in `time_travel.py`.
- `forward_replay(from_index, override_params)` — replays recorded event with optional override params. Allocates fresh `branch_id` via new `next_branch_id()` when forking from non-tip cursor. Replay from tip just appends to `"main"`.

### LiveView init
- Two new instance fields when `time_travel_enabled=True`: `_time_travel_branch_id` (default `"main"`), `_time_travel_branch_counter` (default `0`). Inert when buffer not allocated; zero memory cost for views that don't opt in.

## Tests

9 new cases (5 integration + 4 unit) covering ack-frame shape, replay-enabled semantics, component-only restore isolation, branch-id allocation, defensive defaults. Full suite green (4623 Python passed).

## Backwards compat

All new wire fields are additive — old client (`09a-tab-time-travel.js`) reads only `cursor`/`which`/`history_len` from `time_travel_state`. New fields are ignored; no required handshake.

## Critical files

- `python/djust/websocket.py` — dispatch arms, `_build_time_travel_state` helper, `handle_time_travel_component_jump`, `handle_forward_replay`
- `python/djust/time_travel.py` — `restore_component_snapshot`, `next_branch_id`
- `python/djust/live_view.py` — `_time_travel_branch_id` + `_time_travel_branch_counter` init
- `tests/integration/test_time_travel_flow.py` — 5 new integration cases
- `tests/unit/test_time_travel.py` — 4 new unit cases

## Test plan

- [x] `pytest tests/unit/test_time_travel.py tests/integration/test_time_travel_flow.py` — 71 pass (62 existing + 9 new)
- [x] `pytest python/djust/tests/ tests/` — 4623 pass, 14 skipped, no regressions
- [x] `ruff check + format` — clean
- [x] No JS changes (PR-B owns UI work)

Closes #1151 (PR-A — wire protocol). PR-B (debug panel UI) is the next v0.9.4 PR; will reuse the wire fields exposed here.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,47 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- **Time-travel wire-protocol exposure for branched timelines + per-component
+  scrubbing (PR-A for #1151)** — server-side surface that the v0.9.4 debug
+  panel UI (PR-B, follow-up) consumes. The Python plumbing for per-component
+  time-travel (#1041) and forward-replay through branched timelines (#1042)
+  shipped in v0.9.0; this PR exposes the missing wire fields so the debug
+  panel can drive both.
+  - **`time_travel_state` ack frame**: 3 new additive fields — `branch_id`
+    (defaults `"main"`; new branches allocated as `branch-{N}` on
+    forward-replay from a non-tip cursor), `forward_replay_enabled` (true
+    iff cursor is not at the tip — meaningful replay would produce a
+    branch), `max_events` (the configured ring-buffer cap, so the UI
+    can show "X / max"). Old clients ignore the new keys.
+  - **`time_travel_event` per-event push frame**: 2 new additive fields —
+    `branch_id` and a top-level `components` mirror of
+    `entry.state_after.__components__` so the UI doesn't have to dig into
+    the nested entry.
+  - **New handler `time_travel_component_jump`**: scrubs a SINGLE
+    component's state without touching the parent view or other
+    components. Mirrors the existing `time_travel_jump` validation and
+    re-render path; backed by a new `restore_component_snapshot()`
+    helper in `python/djust/time_travel.py`.
+  - **New handler `forward_replay`**: replays a recorded event with
+    optional `override_params` and allocates a fresh branch id when
+    the cursor is not at the buffer tip. Backed by the existing
+    `replay_event()` (#1042) plus a new `next_branch_id()` allocator.
+  - **Live-view init**: 2 new instance fields on `LiveView.__init__` —
+    `_time_travel_branch_id` (default `"main"`) and
+    `_time_travel_branch_counter` (default `0`). Both are inert when the
+    buffer isn't allocated; zero memory cost for views that don't opt
+    in to time-travel.
+
+  Files: `python/djust/websocket.py` (dispatch arms + 2 handlers + ack
+  builder), `python/djust/time_travel.py` (`restore_component_snapshot`,
+  `next_branch_id`), `python/djust/live_view.py` (branch fields). 9 new
+  cases (5 integration + 4 unit) covering ack-frame shape, replay-enabled
+  semantics, component-only restore isolation, branch-id allocation, and
+  defensive defaults. PR-B (the debug panel UI consuming these fields)
+  is the next v0.9.4 PR.
+
 ### Documentation
 
 - **v0.9.4 process canon (closes #1185, closes #1143, closes #1144)** —
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,11 +42,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
   Files: `python/djust/websocket.py` (dispatch arms + 2 handlers + ack
   builder), `python/djust/time_travel.py` (`restore_component_snapshot`,
-  `next_branch_id`), `python/djust/live_view.py` (branch fields). 9 new
-  cases (5 integration + 4 unit) covering ack-frame shape, replay-enabled
-  semantics, component-only restore isolation, branch-id allocation, and
-  defensive defaults. PR-B (the debug panel UI consuming these fields)
-  is the next v0.9.4 PR.
+  `next_branch_id`), `python/djust/live_view.py` (branch fields). 12 new
+  cases (8 integration + 4 unit) covering ack-frame shape, replay-enabled
+  semantics, component-only restore isolation, branch-id allocation,
+  defensive defaults, override-params-at-tip branching, branch-id
+  no-leak on replay failure, and `which="after"` component restore.
+  PR-B (the debug panel UI consuming these fields) is the next v0.9.4 PR.
 
 ### Documentation
 
--- a/python/djust/live_view.py
+++ b/python/djust/live_view.py
@@ -490,6 +490,12 @@ class LiveView(
         # Allocated only when the subclass opts in via the class attr so
         # the 99% of views that don't use it pay zero memory cost.
         self._time_travel_buffer = None
+        # Branched-timeline tracking (#1151, v0.9.4). Default branch is
+        # "main" — the canonical recorded timeline. Forward-replay from
+        # a non-tip cursor allocates a fresh branch id from the counter.
+        # Both fields are inert when the buffer isn't allocated.
+        self._time_travel_branch_id = "main"
+        self._time_travel_branch_counter = 0
         if getattr(self.__class__, "time_travel_enabled", False):
             try:
                 from djust.time_travel import TimeTravelBuffer
--- a/python/djust/time_travel.py
+++ b/python/djust/time_travel.py
@@ -327,6 +327,104 @@ def restore_snapshot(view: Any, snapshot: EventSnapshot, which: str = "before") 
     return ok
 
 
+def restore_component_snapshot(
+    view: Any,
+    snapshot: EventSnapshot,
+    component_id: str,
+    which: str = "before",
+) -> bool:
+    """Restore a SINGLE component's state from a snapshot (#1151, v0.9.4).
+
+    Single-component variant of :func:`restore_snapshot` Phase 3. Touches
+    ONLY ``view._components[component_id]``; never the parent view, never
+    other components. Used by the debug panel's per-component scrubber so
+    the user can rewind component A's history without disturbing component
+    B's current live state.
+
+    Returns ``True`` if every key was applied via :func:`safe_setattr`.
+    Returns ``False`` (and logs at WARN) when:
+
+    * ``component_id`` is not in ``view._components``.
+    * The snapshot's ``state_before`` / ``state_after`` lacks
+      ``"__components__"`` or has no entry for ``component_id``.
+    * Any individual key fails ``safe_setattr`` (dunder / read-only).
+
+    Mirrors the Phase 3 loop in :func:`restore_snapshot` (lines ~298-326)
+    so behaviour stays consistent — same fallback shape, same logging.
+    """
+    if which not in ("before", "after"):
+        raise ValueError("which must be 'before' or 'after', got %r" % (which,))
+    from djust.security import safe_setattr
+
+    state = snapshot.state_before if which == "before" else snapshot.state_after
+    components_state = state.get(_COMPONENTS_SNAPSHOT_KEY, {}) or {}
+    component_snap = components_state.get(component_id)
+    if component_snap is None:
+        logger.warning(
+            "time_travel: restore_component_snapshot — no entry for %r in snapshot",
+            component_id,
+        )
+        return False
+    registry = getattr(view, "_components", None) or {}
+    component = registry.get(component_id)
+    if component is None:
+        logger.warning(
+            "time_travel: restore_component_snapshot — component %r not in registry",
+            component_id,
+        )
+        return False
+    ok = True
+    for key, value in component_snap.items():
+        try:
+            applied = safe_setattr(component, key, value, allow_private=False)
+        except Exception:  # noqa: BLE001 — dev-only, log + degrade
+            logger.exception(
+                "time_travel: component restore failed for id=%s key=%s",
+                component_id,
+                key,
+            )
+            ok = False
+            continue
+        if not applied:
+            logger.warning(
+                "time_travel: component restore blocked for id=%s key=%s",
+                component_id,
+                key,
+            )
+            ok = False
+    return ok
+
+
+def next_branch_id(view: Any) -> str:
+    """Allocate a fresh branch id for forward-replay (#1151, v0.9.4).
+
+    Returns a string of the form ``"branch-{N}"`` where N is the current
+    value of ``view._time_travel_branch_counter``, then increments the
+    counter so the next call yields a fresh id.
+
+    Stable within a single :class:`TimeTravelBuffer` lifetime — the
+    counter lives on the view instance alongside the buffer, so reconnects
+    against the same view get a deterministic sequence (`branch-0`,
+    `branch-1`, …). Resets to zero on view re-mount because the buffer
+    itself is reallocated then. The ids are only meaningful within a
+    single buffer's lifetime; clients should not assume cross-mount
+    stability.
+
+    Defaults to ``"main"`` when the view doesn't have the counter
+    attribute (defensive — pre-v0.9.4 instances or views constructed
+    via test bypasses).
+    """
+    counter = getattr(view, "_time_travel_branch_counter", None)
+    if counter is None:
+        return "main"
+    new_id = "branch-%d" % counter
+    try:
+        view._time_travel_branch_counter = counter + 1
+    except Exception:  # noqa: BLE001 — slot/descriptor readonly
+        logger.exception("time_travel: failed to increment branch counter")
+    return new_id
+
+
 def replay_event(
     view: Any,
     snapshot: EventSnapshot,
--- a/python/djust/websocket.py
+++ b/python/djust/websocket.py
@@ -1591,6 +1591,10 @@ class LiveViewConsumer(AsyncWebsocketConsumer):
                 self._debug_panel_active = False
             elif msg_type == "time_travel_jump":
                 await self.handle_time_travel_jump(data)
+            elif msg_type == "time_travel_component_jump":
+                await self.handle_time_travel_component_jump(data)
+            elif msg_type == "forward_replay":
+                await self.handle_forward_replay(data)
             else:
                 logger.warning("Unknown message type: %s", msg_type)
                 await self.send_error(f"Unknown message type: {msg_type}")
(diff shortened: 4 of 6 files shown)


johnrtipton (commented):
## Code review — PR #1193 (wire-protocol exposure for #1151 PR-A)

Senior code reviewer pass against the brief in the PR description and the eight focus areas in the review request. Stage-grade verdicts are 🔴 Critical / 🟡 Important / 🟢 Suggestion.

### Strengths

- `_build_time_travel_state` correctly factored as a single helper — DRY across the three handlers (`time_travel_jump`, `time_travel_component_jump`, `forward_replay`).
- `restore_component_snapshot` mirrors `restore_snapshot` Phase 3 exactly (same `safe_setattr(allow_private=False)` shape, same WARN log shape) so future changes only have to touch one mental model.
- All five Stage 11 input gates on the new handlers (`view_instance` is None / `_time_travel_buffer` is None / `index` not int / `which` not in {before,after} / `component_id` not non-empty string / `override_params` not dict-or-None) are present and correctly placed BEFORE any `buffer.jump()` call. The error messages name the field that failed, which speeds up debug-panel client-side error display.
- `next_branch_id` defensive default to `"main"` when counter attr missing handles the legacy-instance + test-bypass cases. Counter increment is wrapped in try/except for slot/descriptor read-only edge cases.
- Backwards compat verified at `python/djust/static/djust/src/debug/09a-tab-time-travel.js` — the v0.6.1 consumer reads only `frame.cursor`, `frame.which`, `frame.history_len`. New fields (`branch_id`, `forward_replay_enabled`, `max_events`, top-level `components`) are silently ignored. No handshake required.
- Component-jump isolation confirmed: `restore_component_snapshot` only writes via `safe_setattr(component, ...)` on the named component instance — never touches `view` itself and never iterates other entries in `view._components`. Test `test_restore_component_snapshot_does_not_touch_parent_state` asserts `view.total == 100` (parent untouched) AND `view._components["comp-b"].value == 88` (other component untouched) — that's the right pair.
- All 71 tests in the time-travel suite pass locally (62 existing + 9 new), no regressions. CHANGELOG entry uses the "9 new cases" delta phrasing which the `check-changelog-test-counts.py` validator correctly skips (delta-prefix detected).

### Issues

#### 🟡 Important — `forward_replay_enabled` boundary disagrees with `handle_forward_replay` branch-allocation gate

**File**: `python/djust/websocket.py` lines 4283-4289 vs lines 4486-4492

The two off-by-one boundaries should be inverses, but they aren't symmetric at the `cursor = history_len - 1, which="before"` boundary.

`_build_time_travel_state` (ack frame):
```python
if which == "after":
    forward_replay_enabled = cursor < history_len - 1
else:  # which == "before"
    forward_replay_enabled = cursor < history_len
```

The docstring explicitly states: *"a cursor at len-1 with which='before' is still pre-tip"* → `forward_replay_enabled = True`.

`handle_forward_replay` (branch allocation):
```python
if from_index < history_len_before - 1:
    new_branch = next_branch_id(self.view_instance)
```

At `from_index = len - 1`, this is `len-1 < len-1` → False, so NO new branch is allocated. But the ack-frame logic says replaying from this position WOULD produce a meaningful branch.

**Concrete failure mode**: UI cursor sits at `(len-1, "before")` → ack frame says `forward_replay_enabled=True` → user clicks the button with `override_params={...}` → `handle_forward_replay` sees `from_index=len-1`, decides this is "at tip", does NOT allocate a branch, replay records onto `"main"`. The branched timeline is silently misattributed.

This is most visible when the user has only ever forward-replayed the very last recorded event with overridden params. The recorded result IS a fork (state moved back to `state_before`, then handler re-ran with different params), but the wire reports it as canonical history.

**Recommended fix**: gate on `from_index < history_len_before` (no `- 1`). Replaying from any recorded snapshot index is forking by construction — `replay_event` always restores to `state_before` first, so even `from_index = len - 1` walks back one step and re-runs. The only "non-fork" case would be appending a brand-new event past the tip, which `forward_replay` doesn't do (`buffer.jump(len)` returns None and we already error out at line 4477).

Add a regression test: `from_index = history_len_before - 1` with override_params should set `view._time_travel_branch_id == "branch-0"`.

#### 🟡 Important — branch_id mutation happens BEFORE replay; left dangling on replay failure

**File**: `python/djust/websocket.py` lines 4486-4499

```python
if from_index < history_len_before - 1:
    new_branch = next_branch_id(self.view_instance)
    try:
        self.view_instance._time_travel_branch_id = new_branch  # SET HERE
    ...

replayed = await sync_to_async(replay_event)(
    self.view_instance, snapshot, override_params, True
)
if replayed is None:
    await self.send_error("forward_replay: replay handler missing or refused")
    return  # branch_id stays as "branch-0", counter is bumped, no recorded events
```

Three failure modes leave inconsistent state:
1. `replay_event` returns None (handler missing / not `@event_handler`-decorated / dunder-rejected) → view sits on `branch-0` with zero recorded events for that branch. Subsequent successful replays will allocate `branch-1`, etc., and `branch-0` is permanently leaked.
2. `replay_event` succeeds but the subsequent `render_with_diff` raises (line 4501-4513) → view is on `branch-0`, replay snapshot WAS recorded (since `record_replay=True`), but the client never gets the ack frame and so doesn't know the new branch_id. Client and server diverge.
3. `restore_snapshot` inside `replay_event` partially fails (returns False) — `replay_event` doesn't currently propagate this to the caller so it's a silent corruption window. Outside this PR's scope but worth flagging.

The user explicitly asked about this (review note #7). Recommended fix:
- Allocate `new_branch` locally first, only commit to `view._time_travel_branch_id` after `replayed is not None` check passes. Restores the original branch_id on failure.
- For mode 2, wrap the render+ack in try/finally that resets `_time_travel_branch_id` if the ack send fails.

The test `test_handle_forward_replay_returns_new_branch_id` only exercises the happy path; consider adding `test_forward_replay_handler_missing_does_not_leak_branch_id`.

#### 🟡 Important — branch counter not reset when `_time_travel_branch_id` returns to `"main"`

**File**: `python/djust/time_travel.py` `next_branch_id` docstring says *"Resets to zero on view re-mount because the buffer itself is reallocated then"* — but that's only true for FRESH instances, not for re-mount on the same view instance.

`LiveView.__init__` sets `_time_travel_branch_counter = 0`. If the view is re-mounted (without instance re-creation), the buffer IS reallocated in `__init__`'s buffer block, but the counter would also be re-init'd alongside since both fields live in `__init__`. So the docstring claim is correct AS LONG AS re-mount goes through `__init__`.

If re-mount is implemented via just calling `self.mount(...)` again (without re-running `__init__`), the counter and branch_id leak across mounts. The docstring should be tightened to: *"Resets when the LiveView instance is re-created. Surviving the buffer's clear() / reset paths is intentional — the counter is not coupled to buffer length."* Or, if re-mount-via-`mount()`-only is supported, the counter SHOULD be reset alongside the buffer reallocation. Suggest grepping for buffer reallocation paths to confirm.

This is mostly a documentation tightening; only affects long-lived view instances that get re-mounted in-place.

#### 🟢 Suggestion — `_build_time_travel_state` config fallback only used when buffer lacks `max_events`

**File**: `python/djust/websocket.py` line 4297

```python
"max_events": getattr(
    buffer, "max_events", _djust_config.get("time_travel_max_events", 100)
),
```

`TimeTravelBuffer` always exposes `max_events` (it's a property — `python/djust/time_travel.py:94`), so the `getattr` fallback never fires in practice. This is fine as defensive code, but the import of `_djust_config` is heavier than the alternative `getattr(buffer, "max_events", 100)`. Trim the import if you want to keep the helper minimal; otherwise leave for clarity. (Not blocking.)

#### 🟢 Suggestion — `time_travel_event` push frame doesn't include `forward_replay_enabled` / `max_events`

**File**: `python/djust/websocket.py` lines 4246-4254

The per-event push frame got 2 of the 5 new wire fields (`branch_id`, `components`) but not `forward_replay_enabled` or `max_events`. Since these change as the buffer grows (new event = `forward_replay_enabled` for any past cursor flips depending on which) the UI may need a follow-up `time_travel_state` frame after each push to refresh those.

Two paths to consider for PR-B:
- (a) push the full augmented state every event (simpler client; ~30 bytes more per frame)
- (b) keep it minimal here and have PR-B emit a synthetic `time_travel_state` after each `time_travel_event` if the cursor isn't at the new tip.

Calling out so PR-B doesn't have to relitigate this. Not a blocker for PR-A.

#### 🟢 Suggestion — `_TwoComponentView` test fixture mutates a class-level dict in mount

**File**: `tests/integration/test_time_travel_flow.py` lines 327-336

```python
class _TwoComponentView(LiveView):
    def mount(self, request=None, **kwargs):
        self.total = 0
        self._components = {
            "comp-a": _CompStub("comp-a"),
            "comp-b": _CompStub("comp-b"),
        }
```

`self._components = {...}` is fine because each mount creates a fresh dict. No leak. But future readers might be confused since the v0.8.6 retro flagged class-mutation-in-init as a recurring failure mode. Consider an inline comment `# fresh dict per instance — see retro 2026-04-26 #1109` for clarity. (Aesthetic only.)

#### 🟢 Suggestion — exception broad-catch in `next_branch_id` increment

**File**: `python/djust/time_travel.py` lines 419-422

```python
try:
    view._time_travel_branch_counter = counter + 1
except Exception:  # noqa: BLE001 — slot/descriptor readonly
    logger.exception("time_travel: failed to increment branch counter")
return new_id
```

If the increment fails, `new_id` is still returned and the view's `_time_travel_branch_id` will be set to it by the caller (`handle_forward_replay` line 4490) — but the counter wasn't bumped, so the NEXT call returns the same id, producing duplicate branch ids. Realistically this only happens if someone does `__slots__` shenanigans on a `LiveView` subclass, which is an exotic config; logging and returning the new id is the right call. Worth a comment noting the duplicate-id risk to alert future readers.

### Test coverage

- 9 new cases (5 integration + 4 unit) match the CHANGELOG claim.
- Test count uses delta phrasing ("9 new cases (5 integration + 4 unit)") which `check-changelog-test-counts.py` correctly skips. No file:count claims to validate.
- Coverage gaps:
  - Boundary case `from_index = history_len - 1, which="before"` for forward_replay (would catch the off-by-one above).
  - Failure-path: `forward_replay` with a missing/un-decorated handler — does branch_id leak?
  - `restore_component_snapshot` with `which="after"` (only "before" is exercised in the new tests; the existing `restore_snapshot` Phase 3 tests cover both, but the new helper is a separate code path).
  - `next_branch_id` called twice on the SAME view to confirm monotonic allocation (currently asserts `branch-0` then `branch-1` — covered, ignore).

### Backwards compat

Confirmed clean:
- v0.6.1 JS reads only `cursor`/`which`/`history_len` from `time_travel_state`.
- Old clients sending `time_travel_jump` get the same ack frame shape they expected, augmented with new optional fields.
- New handlers (`time_travel_component_jump`, `forward_replay`) are additive — old clients never emit those types.

### CHANGELOG

Entry is well-shaped: groups the change under `### Added`, names all five wire fields, names both new handlers, names the helper functions, names the live-view init fields, lists the file paths, asserts the test count using delta phrasing. Closes-#1151 reference present. ✓

### Summary verdict

The wire-protocol contract is mostly right but has one real boundary bug (`forward_replay_enabled` vs branch-allocation gate disagree at `from_index = len-1, which="before"`) plus a state-leak issue when `replay_event` fails after branch_id mutation. Both are dev-only paths and both are recoverable, but PR-B is going to depend on these contracts and getting them right now saves rework when the UI is wired up.

**Verdict**: REQUEST_CHANGES — fix the off-by-one in `handle_forward_replay`, move the branch_id mutation to AFTER the replay-success check, add the two regression tests called out above. The rest are suggestions.

johnrtipton (commented):
Re-review after `19b95433` — both Stage 11 findings fully addressed.

## Bug 1 (off-by-one ack/gate divergence) — FIXED

`websocket.py:4497`:
```python
forks_timeline = from_index < history_len_before - 1 or override_params is not None
```
Inline comment at lines 4481-4493 explains the two-condition spec ("not at tip" OR "override_params present") with explicit reference to the Stage 11 finding. The branch-allocation gate now matches the semantics advertised by `_build_time_travel_state`'s `forward_replay_enabled` flag at `cursor=len-1, which="before"`.

Regression test: `test_handle_forward_replay_branches_on_tip_with_override_params` — buffer of len 1, `from_index=0` (tip), `override_params={"n": 99}`. Asserts `branch_id == "branch-0"` (not "main"). Mentally running the unfixed gate `0 < 0 = False` would skip branch allocation → assertion fails. Real regression guard.

## Bug 2 (branch_id leak on replay failure) — FIXED

The `if forks_timeline: ...` block now lives at `websocket.py:4506-4512`, AFTER the `if replayed is None: return` early-return at lines 4502-4504. Comment at 4490-4493 calls out the order requirement explicitly.

Regression test: `test_handle_forward_replay_does_not_leak_branch_id_on_failure` — view has no `increment` handler so `replay_event` returns None. Asserts both `_time_travel_branch_id` and `_time_travel_branch_counter` unchanged from pre-call values. Mentally running with Bug 1 fixed but Bug 2 unfixed: `forks_timeline = True` (override_params), branch_id mutated to "branch-0" and counter bumped to 1 BEFORE replay returns None → assertions fail. Real regression guard (assuming Bug 1 fix is in place, which is correct since both ship together).

## Bug 3 (component_jump `which="after"` untested) — FIXED

`test_handle_time_travel_component_jump_supports_which_after` exercises the `state_after` restore path on `comp-a`, asserts the component value matches `state_after` (6) not `state_before` (5), and verifies the ack frame echoes `which="after"`. Closes the prior gap where the original component-jump test only covered `which="before"`.

## Test suite + CHANGELOG

- All 74 time-travel tests pass: `pytest tests/integration/test_time_travel_flow.py tests/unit/test_time_travel.py -v` → `74 passed in 0.12s`.
- CHANGELOG count updated 9 → 12 (8 integration + 4 unit). Verified: `git diff 7f358e7b..19b95433` shows 8 new integration tests + 4 new unit tests = 12. Hook `scripts/check-changelog-test-counts.py` exits 0.
- CHANGELOG breakdown explicitly enumerates the new coverage areas ("override-params-at-tip branching, branch-id no-leak on replay failure, `which="after"` component restore"). No #1106 anti-pattern (doesn't make a count claim about a specific test file).

## Sanity pass on the diff

No new defects spotted. The forks_timeline expression is short-circuit-safe (`or`-chain with first operand a bool, second a non-None check). The `try/except` wrapping the branch_id setattr is preserved. Counter increment lives inside `next_branch_id()`, which is only called inside the `if forks_timeline:` block, so the counter genuinely doesn't bump on failure.

One minor observation, not blocking: `_build_time_travel_state.forward_replay_enabled` and `handle_forward_replay.forks_timeline` are now independent predicates (the ack flag advertises "replay is possible from this cursor", the gate decides "this replay forks"). The two are intentionally not 1:1 — that's correct and the docstrings reflect it — but a future PR adding a third condition to either would need to remember both sites. Worth a one-line "both sites must agree on tip semantics" comment as a future tightening, not required for this PR.

APPROVE

johnrtipton:
# PR #1193 Retro — time-travel wire-protocol PR-A (closes #1151 part 1)

**Merged**: 2026-04-28
**Quality rating**: 4/5 (clean wire-protocol PR, 2 real defects caught at Stage 11)

## What worked well

- **Plan stage paid off.** The Plan agent surfaced 5 specific grep targets, the exact additive-field pattern from mount-batch frames to mirror, and a clear two-PR split (wire-protocol vs UI). Implementation closely tracked the plan — no architectural surprises during coding.
- **Mirroring `handle_time_travel_jump` was the right pattern.** Both new handlers (`handle_time_travel_component_jump`, `handle_forward_replay`) follow the same gating + sync_to_async + render_with_diff + ack-frame shape. New code felt familiar to read because it borrows from existing patterns.
- **Reusing existing helpers paid off.** `restore_component_snapshot` is a clean Phase-3-only variant of `restore_snapshot`'s component-restoration block; it required ~30 LoC by leveraging the existing `safe_setattr` + registry-lookup pattern. Same story for `next_branch_id` (read counter, return + increment).
- **Stage 11 reviewer ran the actual tests** + traced bug scenarios mentally to verify the regression tests would catch the unfixed code. This is rigor that the self-review didn't hit.

## What didn't work

- **Off-by-one between two related conditions slipped past self-review.** I wrote `_build_time_travel_state`'s `forward_replay_enabled = cursor < history_len` for `which="before"`, AND `handle_forward_replay`'s gate `from_index < history_len_before - 1`. The two answer the same question ("does replay from this position fork?") but disagreed at `cursor=len-1`. Lesson: when two pieces of code answer the same boolean question, write them once and reference. Or at minimum, write one test that crosses the boundary (replay from cursor=len-1) — would have caught it.

- **Branch_id mutation ordering bug.** I set `view._time_travel_branch_id = new_branch` BEFORE awaiting `replay_event`. The "looks fine" path was: allocate branch_id → run replay → record. But the failure path (handler missing) returned early without rolling back the branch_id. Lesson: when a function has a "commit-or-rollback" shape, mutate AFTER the commit point, not before. The early-return path ALWAYS skips cleanup unless you wrap in try/finally — and try/finally for a single-attribute mutation is heavier than just deferring the assignment.

- **Original test only covered the happy path.** `test_handle_forward_replay_returns_new_branch_id` exercised the success case (override params with valid handler). It didn't cover (a) replay-from-tip-with-overrides (Bug 1) or (b) replay-with-missing-handler (Bug 2). Lesson: any handler that has both an early-return AND a state mutation needs at least one test that triggers the early-return after the mutation point.

- **Initial component-jump tests skipped `which="after"`.** `restore_component_snapshot` accepts `which` as a parameter but my first integration test only exercised `"before"`. The reviewer correctly flagged this as missing coverage. Lesson: when a function has a `which` / `mode` / `kind` parameter, write a test per branch.

## Surprises

- **The off-by-one bug was tractable to spot mentally** but I didn't trace through the cursor=len-1 case during self-review. The Stage 11 reviewer did. Specific ask for next time: when implementing a handler with index/cursor logic, run through the logic at `index=0`, `index=len/2`, `index=len-1`, and `index=len` (out of range) before declaring done. Four mental cases catch most off-by-one classes.

- **The branch_id leak isn't catastrophic** — the worst case is a stale branch counter that doesn't roll over a real ring buffer. But silently disagreeing with the client about the active branch breaks the UI's branch indicator. Subtle correctness bugs in dev-mode tooling have outsized impact because the user is debugging WITH the tool.

## Should-fix follow-ups

- **PR-B (debug panel UI)** — the next v0.9.4 PR. Will consume the `branch_id`, `forward_replay_enabled`, `max_events` fields exposed here. Includes per-component scrubber UI, forward-replay button, branch-tree visualization, max-events dropdown. Estimated 400-500 LoC JS + CSS + at least one Playwright case.
- **PR-B note from this retro**: the `time_travel_event` push frame doesn't carry `forward_replay_enabled` / `max_events` — only the ack frame does. If PR-B's UI needs those fields after every event push, it'll need to either (a) request a fresh `time_travel_state` after each push, or (b) we add the fields to the push frame too. PR-B's Stage 4 should decide.

## Knowledge base updates

One pattern worth promoting if it shows up again:

**Commit-or-rollback handler shape**: any async handler that does both a state mutation AND has an early-return path (validation failure, downstream failure, missing dependency) should mutate AFTER the commit point. Either:
- Defer the mutation past the early-return checks (this PR's fix shape), OR
- Wrap in try/except with explicit rollback (heavier, only justified when multiple mutations need to roll back together).

The leak class is silent — the early-return doesn't raise, so observability tools won't flag it. Stage 11 review caught it, but a principle-level rule in CLAUDE.md would prevent the next instance.

Not adding a CLAUDE.md entry yet — one occurrence isn't enough to canonicalize. If this same defect class shows up in another PR's retro, promote it then.

## Review stats

| Metric | Value |
|--------|-------|
| Files changed | 5 (websocket.py, time_travel.py, live_view.py, 2 test files) + CHANGELOG |
| Lines added | ~647 net (impl + tests) |
| 🔴 findings | 0 |
| 🟡 findings | 2 (off-by-one, branch_id leak — both real correctness bugs) |
| Tests added | 12 (8 integration + 4 unit, was 9 + 3 backfill) |
| Findings fixed pre-merge | 2 of 2 |
| CI failures | 0 |
| Commits | 3 (impl+tests, docs+CHANGELOG, address-findings) |
| Wall-clock | ~60 min from branch to merge |

## What I'd do differently

- **Run through cursor edge cases (0, len/2, len-1, len) during self-review** for any handler with index logic. 4 mental cases takes 30 seconds and catches off-by-one classes.
- **For commit-or-rollback handlers, write the early-return test FIRST.** Before writing the happy-path test, write one that triggers the early-return path AFTER the mutation point would have happened. If the test reveals state corruption, the implementation is wrong.
- **Test every value of `which`/`mode`/`kind` parameters explicitly.** A single happy-path test on one value is insufficient — every accepted value should have at least one assertion against its specific behavior.

johnrtipton:
## Retrospective — PR #1193 (backfill)

**Note**: This retrospective is a **backfill** posted during the v0.9.1-8 final cleanup (#1212 audit pipeline-bypass merges). PR #1193 shipped without a Stage 14 retro comment — same outer-loop integrity failure mode as v0.9.5 PRs #1203 and #1204. Audit ran 2026-04-30; this is best-effort retroactive coverage, not a real-time retro.

**PR title**: feat(time-travel): wire-protocol exposure for branched timelines + per-component scrubbing (PR-A for #1151)

PR #1192, opened 2026-04-28
docs(process): canonicalize 3 retro patterns from v0.9.x arc

johnrtipton:
## Summary

Three CLAUDE.md / PR-checklist additions distilling lessons from the v0.9.0 + v0.9.2 retro arcs. Pure docs change — no code, no tests.

| Issue | What changed |
|-------|--------------|
| #1185 | `docs/PULL_REQUEST_CHECKLIST.md` Closing-Keywords rule now explicitly names the parenthesized form `(closes #X, closes #Y)` as a known auto-close failure mode (validated against PR #1176), and recommends PR body over title for closing keywords. |
| #1143 | New CLAUDE.md section "Process canonicalizations from v0.9.0 retro arc" — Stage-4 first-principles grep before architecting. 5 concrete grep targets (wire-protocol, state-snapshot, async dispatch, decorator composition, component lifecycle). |
| #1144 | Same section — branch-name verify reflex. Pre-commit one-liner that compares `git symbolic-ref --short HEAD` against the active state file's `branch_name` field. Catches the silent wrong-branch-commit failure observed twice in v0.9.0. |

Closes #1185
Closes #1143
Closes #1144

## Why

All three are repeat failure modes from v0.9.x. Each appeared in a retro tracker row (v0.9.0 / v0.9.2) and has been waiting for a docs-only canonicalization PR. Bundling reduces overhead — none of them justify a solo PR, and the docs sections they touch are adjacent.

## Test plan

- [x] `scripts/check-changelog-test-counts.py` — clean (no test count claims)
- [x] CHANGELOG entry under `[Unreleased] > Documentation`
- [x] PR-checklist text reads cleanly in GitHub markdown preview
- [x] CLAUDE.md table of contents not affected (the new section appended after existing v0.9.1 retro arc section)

## Out of scope

- #1180 (PR #1179 follow-ups) — has actual test-strengthening work for the autoescape kwarg parameterization, deferred to a separate PR.
- #1125, #1124, #1123 (v0.8.6 retro patterns) — also canon items, deferring to v0.9.5 to keep this PR focused.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1194, opened 2026-04-28
feat(debug-panel): time-travel UI for branched timelines + per-component scrubber (PR-B for #1151)

### LINKED ISSUE #1151: feat: Debug panel UI for per-component scrubbing + forward-replay

johnrtipton:
## Summary

User-facing debug panel UI built on the wire-protocol shipped in PR-A (#1193). The Time Travel tab now exposes the per-component time-travel (#1041) and forward-replay (#1042) primitives that have been server-side-only since v0.9.0.

Closes #1151.

## What's new in the panel

- **Branch indicator** badge at the top — `main` (blue) vs `branch-N` (orange) when forward-replay has forked the timeline.
- **"X / max" event count** so the user knows when they're approaching the configured `time_travel_max_events` cap.
- **Forward-replay button** (`⏵ replay`) on every history row — sends `forward_replay` frame; server allocates a new `branch_id` if the replay diverges.
- **Per-component expand-toggle** (`▶ N comp`) on rows whose snapshot includes `__components__` — reveals per-component sub-rows with `↶ comp` / `↷ comp` buttons that scrub a SINGLE component's state without touching the parent or other components.
- **Replay-hint** label appears in the header when `forward_replay_enabled` is true.

## Implementation notes

- **CSP-strict** (per CLAUDE.md canon #1175): zero inline event handlers, all interactivity through the existing delegated click handler on the panel root, now matching 4 button classes.
- **Backwards-compatible**: `onTimeTravelState` tolerates undefined for all 3 new fields. Pre-v0.9.4 servers fall back to v0.6.1 panel behavior.
- **Centralized WS send** via private `_sendTimeTravelMessage` helper.

## Files

- `python/djust/static/djust/src/debug/09a-tab-time-travel.js` — rewrote 156 → 320 LoC
- `python/djust/static/djust/debug-panel.css` — 90 LoC of additive `.tt-branch*`, `.tt-comp-*`, `.tt-forward-replay`, `.tt-expand-toggle` rules
- Regenerated bundles via `scripts/build-client.sh`
- `tests/js/debug_panel_time_travel_ui.test.js` — 17 new vitest cases

## Test plan

- [x] `npx vitest run tests/js/debug_panel_time_travel_ui.test.js` — 17/17 pass
- [x] `make test` — 4047 Python + 1480 JS (was 1463; +17 new) all green
- [x] CSP-strict: visual-pass review of the diff confirms no inline handlers, no `<script>` blocks
- [x] Bundles built clean

## Out of scope

- **Override-params input UI** for forward-replay — currently the button sends with no override (re-runs with original params). Adding a small modal for typing override params is a v0.9.5 candidate.
- **Branch-tree visualization** — the badge shows which branch you're on; visualizing the tree of branches is a v0.9.5 candidate.
- **Max-events runtime mutation** — currently shown as a read-only count; making it an editable dropdown is a v0.9.5 candidate.

Closes #1151.


johnrtipton (commented):
## PR-B Review — debug-panel time-travel UI (#1151)

Reviewed the source module, regenerated bundles, CSS, CHANGELOG, and the 17 vitest cases. Verdict: **COMMENT** (no blockers, two minor regressions / test-coverage gaps worth noting before merge).

---

### Verdict legend

- 🔴 critical / blocker
- 🟡 important / should fix
- 🟢 nit / informational

---

### What was done well

- Bundle regeneration is correct: `debug-panel.min.js.gz` and `debug-panel.min.js.br` decompress byte-for-byte to `debug-panel.min.js`, and the unminified `debug-panel.js` reflects the source diff exactly (verified via `grep -c` for new symbols and via `gunzip -c | diff -`). All 7 new CSS class names (`tt-branch-main`, `tt-branch-fork`, `tt-replay-hint`, `tt-forward-replay`, `tt-comp-jump`, `tt-comp-id`, `tt-expand-toggle`) are present in the minified bundle.
- CSP-strict claim holds: `grep -nE 'on[a-z]+='` against the source returns nothing. All four button classes are wired through `data-tt-*` attributes + a delegated `click` listener on `this.panel`. No inline `<script>`, no inline handlers.
- The four dispatch branches in `registerTimeTravelClickHandlers` use mutually-exclusive class names (`.tt-jump` is token-based and does NOT match `.tt-comp-jump`), each branch has a `parseInt`/`isNaN` guard, and each path early-`return`s — so no double-firing risk if a future refactor nests buttons.
- 17 it() cases match the CHANGELOG claim; 17/17 pass; the full JS suite (`tests/js/`) passes 1480/1480 (matches the new total in the test plan).
- `onTimeTravelState` typeof-guards are correct for all three new fields. `branch_id` only overwrites when a `string` is sent (so `null` from the server does NOT clobber to `null`); `forward_replay_enabled` requires a `boolean`; `max_events` requires a `number`. Pre-v0.9.4 servers gracefully no-op on these fields, and there is an explicit regression test.
- `_sendTimeTravelMessage` is a clean centralization of the WS-not-ready guard. All three callers route through it consistently.
- Branch-id text content is escaped (`escapeHtml(branchId)` inside a `<span>`), and `entry.error` / `entry.event_name` / `entry.params` previews are all escaped before render.

---

### 🟡 Important

**1. `onTimeTravelJumpClick` lost its argument-type guard in the refactor.**

Pre-PR (current `main`):
```js
onTimeTravelJumpClick(index, which) {
    if (typeof index !== 'number') return;
    // ...
    lv.sendMessage(payload);
}
```

Post-PR:
```js
onTimeTravelJumpClick(index, which) {
    this._sendTimeTravelMessage({ type: 'time_travel_jump', index: index, which: which || 'before' });
}
```

The new sibling handlers `onTimeTravelComponentJumpClick` and `onTimeTravelForwardReplayClick` both do `if (typeof index !== 'number') return;`. The dropped guard on `onTimeTravelJumpClick` makes it inconsistent. The DOM dispatch path (`registerTimeTravelClickHandlers`) still does `parseInt(raw, 10)` + `isNaN` check, so user-clicks are safe. But any programmatic / test caller can now post a `time_travel_jump` frame with `index: 'foo'`, which previously was silently dropped. Recommend restoring the type guard for symmetry with the other two handlers — a 1-line `if (typeof index !== 'number') return;`.

**2. Delegated click-handler dispatch logic is not covered by integration tests.**

All 6 dispatch tests (`onTimeTravelComponentJumpClick sends ...`, `onTimeTravelForwardReplayClick ...`, etc.) call the handler methods directly. None simulate a real DOM click event through `registerTimeTravelClickHandlers` (no `panel.dispatchEvent(new MouseEvent('click'))` or equivalent). That means:
- The `.tt-jump` vs `.tt-comp-jump` vs `.tt-forward-replay` vs `.tt-expand-toggle` class-match logic is uncovered.
- The `parseInt(raw, 10)` + `isNaN(index)` guard for click-time numeric parsing is uncovered.
- The `target.closest()` + `this.panel.contains(button)` containment check is uncovered.

A future refactor could break the class-name match (e.g., rename `.tt-comp-jump` → `.tt-component-jump` and forget to update the dispatch handler) and the unit tests would still pass. Per CLAUDE.md canon #1104 ("Mechanical replacement: N similar sites need N tests"), each of the four dispatch branches deserves a fixture-driven click test. Suggest adding 4 cases: render the tab into a real DOM, dispatch a click on each button class, assert the corresponding handler was invoked. Not a blocker but the dispatch path is the riskiest area on this PR and currently has zero direct coverage.

---

### 🟢 Suggestions / informational

**3. `escapeHtml` does NOT escape `"` or `'` — pre-existing pattern, but PR adds new attribute interpolations.**

`escapeHtml` is implemented as `div.textContent = text; return div.innerHTML;` (see `static/djust/src/debug/14-utils.js:119-125`). Browser `textContent → innerHTML` round-trip escapes `<`, `>`, `&` but **not** `"` or `'`. So in attribute contexts like:

```js
data-tt-comp-id="${this.escapeHtml(compId)}"
title="Restore ${this.escapeHtml(compId)} only — leaves other components alone"
```

a `compId` containing a literal `"` would break out of the attribute. This is pre-existing in the codebase (`03-tab-events.js:32` already does `value="${this.escapeHtml(...)}"`), but this PR adds new attribute interpolations of `compId` (a value originally controlled by user-code component IDs in `view._components`). In practice djust component IDs are dev-set and unlikely to contain `"`, but the latent risk is real.

Two paths forward (either is fine, neither is blocking):
- (a) For this PR, leave as-is and add a follow-up issue to introduce `escapeAttr(s)` that also escapes `&quot;` and `&#39;`, then sweep all `="${escapeHtml(...)}"` sites.
- (b) Use a single-quoted attribute or `JSON.stringify` for the data attribute value (since you control the markup): `data-tt-comp-id=${JSON.stringify(compId)}` — JS template literal will splice in the quoted+escaped JSON string. This neutralizes both `"` and `'` and is cleaner than two-pass escape.

**4. `toggleTimeTravelExpandRow` accumulates `false` entries in `timeTravelExpandedRows` rather than `delete`-ing them.**

```js
this.timeTravelExpandedRows[index] = !this.timeTravelExpandedRows[index];
```

After many toggles, the dict carries explicit `false` keys for every row that was ever expanded then collapsed. Cosmetic memory leak, doesn't affect correctness. Could be:
```js
if (this.timeTravelExpandedRows[index]) {
    delete this.timeTravelExpandedRows[index];
} else {
    this.timeTravelExpandedRows[index] = true;
}
```

**5. `onTimeTravelEvent`'s fallthrough path is uncovered.**

Lines 309-314 do:
```js
if (typeof this.refreshActiveTab === 'function') {
    this.refreshActiveTab();
} else if (typeof this.renderTabContent === 'function') {
    this.renderTabContent();
}
```

The `renderTabContent` fallback isn't exercised by any test (the harness panel always has `refreshActiveTab`). Probably fine since this is a "tolerate older host" branch, but worth a one-line skip-test if the fallback is meant to support a real upgrade path.

**6. Branch fork badge title is static text.**

```html
<span class="tt-branch tt-branch-fork" title="Branched timeline — diverged from main via forward-replay">${this.escapeHtml(branchId)}</span>
```

The branch ID appears in the visible badge label but not in the title — minor UX missed-opportunity. Consider `title="Branched timeline (${this.escapeHtml(branchId)}) — diverged from main via forward-replay"`. Trivial.

---

### Spot-checks performed

- `npx vitest run tests/js/debug_panel_time_travel_ui.test.js` → **17/17 pass**.
- `npx vitest run tests/js/` → **1480/1480 pass** (matches CHANGELOG).
- `gunzip -c debug-panel.min.js.gz | diff - debug-panel.min.js` → no diff.
- `brotli -d -c debug-panel.min.js.br | diff - debug-panel.min.js` → no diff.
- `grep -nE 'on[a-z]+='` against the source module → no inline handlers.
- All 7 new CSS class names appear in the minified bundle.

---

### Final call

The PR is **mergeable as-is**. The XSS attribute-escaping concern (#3) is a pre-existing codebase pattern, not a new vulnerability introduced by this PR. The two findings I'd recommend addressing before merge are:
- 🟡 Restore the `typeof index !== 'number'` guard on `onTimeTravelJumpClick` for symmetry with the other two new handlers.
- 🟡 Add 4 click-dispatch integration tests (one per button class) to lock in the delegated handler routing.

If you want to land this for v0.9.4 and follow up the type guard + dispatch tests in v0.9.5, that's also reasonable given PR-A is already merged and the branch is mergeable.

**APPROVE** with the two 🟡 findings tracked as either follow-up or addressed-in-place commits before merge.

johnrtipton:
# PR #1194 Retro — debug panel UI for time-travel (closes #1151 part 2)

**Merged**: 2026-04-28
**Quality rating**: 4/5 (clean UI PR, 2 real findings caught at Stage 11)

## What worked well

- **Two-PR split paid off.** PR-A landed the wire protocol with a tight contract; PR-B consumed those fields without touching server code. The CHANGELOG entry for PR-A explicitly noted what would arrive in PR-B, and PR-B referenced PR-A's contract. A reviewer reading PR-B can verify against PR-A's tests; they don't have to scroll through 800 LoC of server-side logic.
- **CSP-strict pattern transferred cleanly.** The existing `registerTimeTravelClickHandlers` already used the marker-class + delegated-listener pattern (per CLAUDE.md canon #1175). Adding 3 new button classes (`.tt-comp-jump`, `.tt-forward-replay`, `.tt-expand-toggle`) just meant adding 3 more `target.closest()` branches in the same handler — no architectural decision.
- **Backwards compat at the consumer side was tested upfront.** The `onTimeTravelState tolerates missing v0.9.4 fields` test verifies pre-v0.9.4 server frames still work. Worth doing because the panel ships in djust 0.9.x but downstream consumers might be running an older djust pinned via pyproject.toml.
- **Stage 11 reviewer ran the bundles** to verify byte-correctness of the gz / br / map artifacts. Caught nothing, but it means future bundle-mismatch bugs would be caught.

## What didn't work

- **The refactor inadvertently dropped a type guard.** `onTimeTravelJumpClick` had `if (typeof index !== 'number') return;` as its first line pre-PR. When I introduced `_sendTimeTravelMessage` and routed the call through it, I removed the explicit guard and assumed the centralized helper would validate. It doesn't — it just sends. The two new sibling handlers (`onTimeTravelComponentJumpClick`, `onTimeTravelForwardReplayClick`) DO validate, which made the inconsistency easy to see in retrospect. Lesson: **when refactoring out a helper, audit each call site for guard logic that was deferred to the call site, and decide whether to push it INTO the helper OR keep it at the call site**. The cleanest answer here is at the call site, since each handler has a different shape of input to validate.

- **Zero integration tests for the click-dispatch branch logic.** All 17 of my initial tests called handler methods directly. The `registerTimeTravelClickHandlers` delegated-click code (4 button-class match branches, parseInt + isNaN guards, target.closest containment check) was untested. Stage 11 caught this. Lesson: **whenever a feature is "delegated event handler matches multiple selectors", at least one test per selector should fire a real DOM click through the handler**. Direct method calls verify the methods, not the wiring between user action → method call.

- **Pre-commit hook reformatted source after I committed bundles.** The `end-of-file-fixer` hook found a missing trailing newline in `09a-tab-time-travel.js` and the `build-js` hook then regenerated bundles to match. The first `git commit` failed; I had to re-stage and re-commit. Lesson: **run pre-commit hooks against staged files before the first commit attempt**. Or: trust that the second commit attempt will succeed (it does — the pre-commit reformatting is idempotent).

## Surprises

- **The integration tests required resetting `_ttClickBound = false`.** The panel's constructor already binds click handlers to its internal DOM. When my test re-assigns `panel.panel = panelEl` (a fresh test container) and calls `registerTimeTravelClickHandlers()` again, the `if (this._ttClickBound) return;` short-circuit prevents re-registration. Had to manually reset the flag. Lesson: **the `_xxxBound` idempotency-guard pattern is correct in production but adds friction in tests that re-bind**. Worth a comment in the helper noting test-time reset.

- **Large diff (~700 LoC) felt risky** but Stage 11 found only 2 modest issues. The risk profile of pure-frontend PRs feels different from backend — fewer edge cases, more visual-only outcomes, but also less server-side state to corrupt. The two findings here were both about test coverage / consistency, not about the runtime UI being wrong.

## Should-fix follow-ups

- **`escapeHtml` doesn't escape quotes.** Pre-existing pattern in the codebase (used by all panel tabs since v0.6.1). This PR adds new attribute interpolations of `compId` (e.g. `data-tt-comp-id="${escapeHtml(compId)}"`). If a developer-controlled component ID ever contains a `"`, the attribute breaks. Realistic risk is low (component IDs are framework-allocated strings), but worth a follow-up to add `escapeAttr` or use `JSON.stringify` for data attributes. Out of scope here; defer to v0.9.5.

- **Override-params input UI for forward-replay.** Currently the button sends `{ from_index: N }` with no override (re-runs with original params). A small modal for typing override params would unlock the "what if this counter started at 999?" debugging workflow. v0.9.5 candidate.

- **Branch-tree visualization.** The badge shows which branch you're on; visualizing the tree of branches (parent / child relationships) is a v0.9.5 candidate.

## Knowledge base updates

Two patterns worth promoting if they show up again:

1. **Refactor-with-helper guard audit**: when extracting a helper from N call sites, audit each call site for input-validation logic that was previously inline. Decide explicitly: push into helper, or keep at call site. The "dropped a guard" failure mode is silent — production code keeps working when the inputs are well-formed, breaks only on malformed inputs that may not happen in tests.

2. **Delegated-listener integration test pattern**: for any "marker class + delegated listener" feature, the unit-test layer (direct method invocation) and the integration layer (real DOM click → registered handler → method) need separate coverage. Unit tests verify the methods; integration tests verify the wiring. Both are cheap (~5 lines each); skipping integration is what bit this PR.

Not adding CLAUDE.md entries yet — one occurrence each. If either pattern shows up in another retro, promote it then.

## Review stats

| Metric | Value |
|--------|-------|
| Files changed | 8 (1 source JS, 1 CSS, 5 bundles, 1 test, 1 CHANGELOG) |
| Lines added | ~700 net |
| 🔴 findings | 0 |
| 🟡 findings | 2 (dropped type guard, missing integration tests) |
| 🟢 findings | 1 (escapeHtml pre-existing — deferred) |
| Tests added | 17 → 23 (added 6 integration cases for the 🟡 #2) |
| Findings fixed pre-merge | 2 of 2 |
| CI failures | 0 |
| Commits | 3 (impl+tests, docs, address-findings) |
| Wall-clock | ~50 min from branch to merge |

## What I'd do differently

- **Audit guard logic when extracting helpers.** Before each "this validation is now in the helper" refactor, list the validations at each call site and decide where they live. Cheap upfront; expensive if Stage 11 has to find it.
- **Write at least one integration test per delegated-listener branch.** 5 minutes per test, catches the "wiring is wrong" failure mode unit tests miss.
- **Run pre-commit hooks against staged files before the first commit.** `pre-commit run --files <staged>` would have caught the trailing-newline issue and rebuilt the bundles in one step instead of two commits.

## v0.9.4 milestone closeout

This is the last v0.9.4 PR (#1190 HVR auto-enable, #1191 test-infra, #1192 process canon, #1193 wire-protocol, #1194 UI = 5 PRs total). Per the post-completion housekeeping, I'll:

1. Push main if needed.
2. Note that v0.9.4 milestone is complete and the user can run `/pipeline-retro --milestone v0.9.4` for the milestone retrospective.
3. v0.9.5 candidates filed: docs.djust.org Makefile migration, docs.djust.org green-theming experiment, override-params input UI, branch-tree visualization, escapeAttr helper, remaining v0.8.6 canon items (#1125, #1124, #1123).

johnrtipton:
## Retrospective — PR #1194 (backfill)

**Note**: This retrospective is a **backfill** posted during the v0.9.1-8 final cleanup (#1212 audit pipeline-bypass merges). PR #1194 shipped without a Stage 14 retro comment — same outer-loop integrity failure mode as v0.9.5 PRs #1203 and #1204. Audit ran 2026-04-30; this is best-effort retroactive coverage, not a real-time retro.

**PR title**: feat(debug-panel): time-travel UI for branched timelines + per-component scrubber (PR-B for #1151)

Issue #198, opened 2026-02-01
VDOM path fallback corrupts sibling elements when removing list items


johnrtipton:
## Bug Description

When elements are removed from a `{% for %}` loop (e.g., deleting all items from a todo list), the VDOM patching corrupts sibling elements. Text patches meant for one element get applied to wrong nodes -- for example, counter text gets applied inside a `<pre>` code block, eating spaces.

## Root Cause

`getNodeByPath()` in `client.js` (line ~2019) filters out whitespace-only text nodes when counting children during path-based traversal:

```javascript
const children = Array.from(node.childNodes).filter(child => {
    if (child.nodeType === Node.ELEMENT_NODE) return true;
    if (child.nodeType === Node.TEXT_NODE) {
        return child.textContent.trim().length > 0;
    }
    return false;
});
```

However, the server's Rust VDOM (`diff.rs`) counts ALL children including whitespace text nodes when computing path indices (via `diff_indexed_children` which iterates `0..children.len()`). This index mismatch causes patches to be applied to the wrong nodes when the ID-based lookup fails and the path fallback is used.

## Reproduction Steps

1. Create a LiveView with a `{% for %}` loop rendering list items and a sibling element (e.g., a counter or `<pre>` code block)
2. Add several items to the list
3. Delete all items from the list
4. Observe that sibling elements (counter text, code blocks) are corrupted

## Console Evidence

During the delete operations, the console shows:
```
[LiveView] ID lookup failed, trying path fallback
```
This appears multiple times, confirming that the path fallback is being triggered and the index mismatch corrupts sibling nodes.

## Expected Behavior

When list items are removed from a for loop, sibling elements should remain untouched. The path-based fallback should count children the same way the server does.

## Proposed Fix

Remove the whitespace filtering in `getNodeByPath()` so it uses `node.childNodes` directly, matching the server's child counting. The existing `getSignificantChildren()` function (which has whitespace-preservation logic for `<pre>`/`<code>` tags) is used elsewhere for patch application, but `getNodeByPath()` should count ALL nodes to match server indices.

Issue #193, opened 2026-02-01
Debug panel: Event Handlers tab empty for handlers without @event_handler decorator


johnrtipton:
## Bug Description

The debug panel's Event Handlers tab shows empty results for any LiveView that uses plain method handlers (the default djust pattern) without the `@event_handler()` decorator.

## Root Cause

In `python/djust/mixins/post_processing.py`, `get_debug_info()` at lines 38-41 only discovers handlers that have `_djust_decorators` attribute with `event_handler` key:

```python
if (
    callable(attr)
    and hasattr(attr, "_djust_decorators")
    and "event_handler" in getattr(attr, "_djust_decorators", {})
):
```

However, djust's core event system (`websocket.py` and `websocket_utils.py`) does NOT require the `@event_handler()` decorator -- any public method on a LiveView can be an event handler. This means the debug panel misses all handlers in the common case.

## Impact

- The Event Handlers tab is essentially broken for most LiveView apps
- Developers using the recommended djust patterns (plain methods) see "No event handlers found"
- Only developers who explicitly add the optional `@event_handler()` decorator see their handlers

## Expected Behavior

The debug panel should discover event handlers using the same logic as the websocket event system -- checking for any public method that could be called as an event handler (public, callable, not a Django/base class built-in).

## Affected Files

- `python/djust/mixins/post_processing.py` - `get_debug_info()` lines 38-41
- `docs/DEBUG_PANEL.md` - Troubleshooting section documents this as expected but it shouldn't require the decorator

johnrtipton:
Verified on djust.org (http://localhost:8002/examples/). The original report was filed from testing the wrong site, but the core observation is valid: `get_debug_info()` at lines 38-42 of post_processing.py only discovers handlers with `_djust_decorators` containing `event_handler`. The websocket event dispatch (`websocket_utils.py` line 112) uses plain `getattr()` and doesn't require the decorator. Any public method can be an event handler, but the debug panel only shows decorated ones.

Note: on djust.org's examples page, all handlers DO use `@event_handler`, so they appear correctly. But in user applications using plain methods, the Handlers tab would be empty.

Also see #195 for a separate crash bug in the Handlers tab rendering.

PR #1203, opened 2026-04-29
fix(websocket): store _recovery_html after broadcast renders (#1202)

### LINKED ISSUE #1202: server_push path does not maintain _recovery_html — leads to 'Recovery HTML unavailable' → forced page reload after broadcast-triggered renders
## Summary

The `server_push` path in `djust/websocket.py` renders patches and broadcasts them but **does not update `self._recovery_html`**. When a subsequent VDOM patch fails on the client (e.g. because a `{% if %}` block added/removed a DOM subtree) and the client requests `request_html` recovery, the server has stale `_recovery_html` from the last user-event render — or never had one at all if the session has only received broadcasts since mount.

When recovery HTML is unavailable, the server returns `recoverable=false` and the client triggers `window.location.reload()`. For apps using `push_to_view` to drive live UI updates, this manifests as intermittent full-page reloads whenever the template has any structural branching.

## Version

- djust 0.9.0rc2
- channels-redis 4.3.0

## Code references

**`djust/websocket.py:3267`** — `_recovery_html` is set after the event-driven render:
\`\`\`python
# Store rendered HTML for on-demand recovery.
# Client sends request_html when applyPatches() fails
# (e.g., {% if %} blocks shifting DOM structure).
self._recovery_html = html
self._recovery_version = version
\`\`\`

**`djust/websocket.py:4416-4426`** — `server_push` renders and sends patches, but skips the `_recovery_html` update:
\`\`\`python
html, patches, version = await sync_to_async(self.view_instance.render_with_diff)()

if patches is not None:
    if isinstance(patches, str):
        patches = fast_json_loads(patches)
    await self._send_update(
        patches=patches,
        version=version,
        broadcast=True,
        source="broadcast",
    )
# ← no self._recovery_html = html here
\`\`\`

**`djust/websocket.py:4154-4163`** — `handle_request_html` reads `_recovery_html`; if missing, sends `recoverable=false`:
\`\`\`python
html = getattr(self, "_recovery_html", None)
version = getattr(self, "_recovery_version", 0)

if not html:
    await self.send_error(
        "Recovery HTML unavailable — the server may have restarted. "
        "A page reload will fix this.",
        recoverable=False,
    )
    return
\`\`\`

**`djust/static/djust/client.js:1173-1176`** — client acts on `recoverable=false` by reloading:
\`\`\`javascript
if (data.recoverable === false) {
    console.warn('[LiveView] Non-recoverable error, reloading page...');
    window.location.reload();
    break;
}
\`\`\`

## Repro

Template with any structural branching inside a loop:
\`\`\`django
<ul>
  {% for item in items %}
    <li>
      {{ item.name }}
      {% if item.status == "complete" %}<span class="badge">Done</span>{% endif %}
    </li>
  {% endfor %}
</ul>
\`\`\`

1. Mount the view with a non-empty \`items\` list, no item with status="complete".
2. From a separate process (e.g. a Celery task), call \`push_to_view(view_path, state={...})\` that causes an item to transition to status="complete".
3. Client applyPatches fails because a new \`<span class="badge">\` subtree appears.
4. Client sends \`{"type": "request_html"}\`.
5. Server's \`_recovery_html\` is still the pre-push HTML (or empty if there were no prior events this session).
6. If empty → recoverable=false → page reload.
7. If stale → morphing succeeds but loses the state change that triggered the broadcast in the first place.

## Suggested fix

In \`handle_server_push\`, mirror the behavior of the event-driven path and store the render output:

\`\`\`python
html, patches, version = await sync_to_async(self.view_instance.render_with_diff)()

if patches is not None:
    ...
    await self._send_update(
        patches=patches,
        version=version,
        broadcast=True,
        source="broadcast",
    )
    self._recovery_html = html
    self._recovery_version = version
\`\`\`

This ensures that a later \`request_html\` after a broadcast-triggered patch failure gets the fresh post-broadcast HTML.

## Secondary docs-side guidance

The client warning is excellent:

> [LiveView] VDOM patches failed. This usually happens when {% if %} blocks add/remove DOM elements, shifting sibling positions.
> Fix: Use style="display:none" toggling instead of {% if %} for elements

For apps driven primarily by \`push_to_view\`, it would help to call this out in the \`push_to_view\` guide: any template subtree affected by broadcast state changes should avoid \`{% if %}\` for add/remove; use \`d-none\` / \`display:none\` toggling instead. We hit this in our app and eventually stabilized the DOM, but the \`_recovery_html\` gap still bites when structural transitions happen anywhere else in the VDOM (e.g. adding a new row to a list).

## Downstream impact

Our prototype saw this manifesting as intermittent page reloads during the OCR pipeline demo flow: a document uploads (adds a new \`<li>\`), OCR completes (status transitions), and the second push-triggered patch on the added \`<li>\` fails, triggering recovery — which either reloaded the page (\`_recovery_html\` was never set for this session's broadcasts) or morphed to stale HTML.

johnrtipton:
Fixes #1202.

## Problem

`server_push` in `python/djust/websocket.py` calls `render_with_diff` and sends patches but **never updates `self._recovery_html`**. `handle_event` (the user-initiated event path) does this, but `server_push` was overlooked.

Consequence: when a subsequent `applyPatches` on the client fails (the well-known `{% if %}`-shifts-DOM case), the client sends `request_html` expecting fresh HTML. `handle_request_html` reads `self._recovery_html` — if None, it returns `recoverable: false` and the client triggers `window.location.reload()`.

Sessions that only receive broadcasts after mount (common for real-time apps: an admin watches a dashboard, pushes come from Celery workers) never populate `_recovery_html` via the normal path. Result: intermittent full-page reloads whenever broadcast-driven template changes shift DOM structure.

## Fix

One-location change in `server_push`. Before calling `self._send_update(...)` for the broadcast patches, store the render output as `_recovery_html` / `_recovery_version`. Same pattern as `handle_event` at websocket.py:3271.

## Tests

Added two unit tests in `tests/unit/test_server_push.py`:

- `test_stores_recovery_html_after_broadcast` — single push stores both `_recovery_html` and `_recovery_version`.
- `test_recovery_html_refreshed_across_multiple_pushes` — consecutive pushes each refresh the stored HTML so a stale broadcast isn't served on a later recovery.

Both pass locally. Existing 14 tests in the file remain green.

## Downstream context

Observed in a production prototype where a document upload pushes twice (new row appears as PENDING, then OCR status flips on completion). Without this fix, the second push's patches race recovery-HTML staleness and the user sees:

```
[LiveView] VDOM patches failed. This usually happens when {% if %} blocks…
[LiveView] Server error: Recovery HTML unavailable — the server may have restarted.
[LiveView] Non-recoverable error, reloading page...
```

With this fix applied, the recovery cycle has fresh HTML to morph against and the page doesn't reload.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- **`server_push` now stores `_recovery_html` / `_recovery_version` after
+  broadcast renders (#1202)** — push-driven sessions previously left recovery
+  state unset, so a client `request_html` after a failed VDOM patch
+  (e.g. `{% if %}` shifting DOM structure on a broadcast) returned
+  `recoverable=false` and force-reloaded the page. `server_push` now mirrors
+  the `handle_event` pattern of populating `_recovery_html` /
+  `_recovery_version` immediately before dispatching the broadcast patches.
+  Added 3 regression cases in `tests/unit/test_server_push.py`
+  (single-push, multi-push refresh, no-op-push leaves recovery state intact).
+
 ### Security
 
 - **Code-scanning cleanup batch (4 fixes + 15 false-positive dismissals)** —
--- a/python/djust/websocket.py
+++ b/python/djust/websocket.py
@@ -4636,6 +4636,15 @@ class LiveViewConsumer(AsyncWebsocketConsumer):
                 if patches is not None:
                     if isinstance(patches, str):
                         patches = fast_json_loads(patches)
+                    # Store rendered HTML for on-demand recovery, mirroring
+                    # handle_event. Without this, a VDOM patch failure on a
+                    # broadcast-triggered render (e.g. {% if %} blocks shifting
+                    # DOM structure after a push-driven state change) triggers
+                    # `request_html` with no `_recovery_html` to serve —
+                    # server returns `recoverable=false` and the client
+                    # force-reloads the page. See #1202.
+                    self._recovery_html = html
+                    self._recovery_version = version
                     await self._send_update(
                         patches=patches,
                         version=version,
--- a/python/djust/websocket.py
+++ b/python/djust/websocket.py
@@ -4637,12 +4637,9 @@ class LiveViewConsumer(AsyncWebsocketConsumer):
                     if isinstance(patches, str):
                         patches = fast_json_loads(patches)
                     # Store rendered HTML for on-demand recovery, mirroring
-                    # handle_event. Without this, a VDOM patch failure on a
-                    # broadcast-triggered render (e.g. {% if %} blocks shifting
-                    # DOM structure after a push-driven state change) triggers
-                    # `request_html` with no `_recovery_html` to serve —
-                    # server returns `recoverable=false` and the client
-                    # force-reloads the page. See #1202.
+                    # handle_event. Without this, request_html after a failed
+                    # broadcast-triggered patch finds _recovery_html=None and
+                    # forces a page reload. See #1202.
                     self._recovery_html = html
                     self._recovery_version = version
                     await self._send_update(
--- a/tests/unit/test_server_push.py
+++ b/tests/unit/test_server_push.py
@@ -175,6 +175,47 @@ class TestServerPushHandler:
         # Should not raise
         await consumer.server_push({"state": {"x": 1}})
 
+    @pytest.mark.asyncio
+    async def test_stores_recovery_html_after_broadcast(self):
+        """After a broadcast-triggered render, _recovery_html and
+        _recovery_version must be updated so a later `request_html`
+        recovery (triggered by a failed VDOM patch on the client) has
+        fresh HTML to serve. See #1202.
+
+        Regression: previously only handle_event set _recovery_html.
+        If a session received only broadcasts after mount, _recovery_html
+        stayed None, request_html returned `recoverable=false`, and the
+        client force-reloaded the page.
+        """
+        consumer = self._make_consumer()
+        consumer.view_instance.render_with_diff = MagicMock(
+            return_value=("<div>rendered-after-push</div>", '[{"op":"replace"}]', 42)
+        )
+
+        await consumer.server_push({"state": {"x": 1}, "handler": None, "payload": None})
+
+        assert consumer._recovery_html == "<div>rendered-after-push</div>"
+        assert consumer._recovery_version == 42
+
+    @pytest.mark.asyncio
+    async def test_recovery_html_refreshed_across_multiple_pushes(self):
+        """Consecutive pushes must each refresh _recovery_html so a stale
+        render from an earlier broadcast isn't served on a later recovery."""
+        consumer = self._make_consumer()
+        consumer.view_instance.render_with_diff = MagicMock(
+            side_effect=[
+                ("<div>first</div>", '[{"op":"replace","v":"first"}]', 1),
+                ("<div>second</div>", '[{"op":"replace","v":"second"}]', 2),
+                ("<div>third</div>", '[{"op":"replace","v":"third"}]', 3),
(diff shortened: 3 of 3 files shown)


johnrtipton:
Addressed all 3 review findings in d1ae6771:

1. **CHANGELOG entry** (required) — added `### Fixed` block under `[Unreleased]` describing the recovery-state fix and citing #1202.
2. **Verbose comment** (style) — trimmed the 7-line inline comment to 4 lines + `See #1202` reference.
3. **`patches=None` branch coverage** (test gap) — added `test_no_patches_does_not_clobber_recovery_html` that pre-seeds `_recovery_html` / `_recovery_version`, runs a `server_push` whose `render_with_diff` returns `patches=None`, and asserts the previously-good recovery state survives. Locks in the conditional behavior so a future regression that moves the assignment outside the `if patches is not None:` guard would be caught.

All 17 tests in `tests/unit/test_server_push.py` pass locally. Pre-push hooks ran clean (full pytest suite, secrets, CHANGELOG count, noqa).

johnrtipton:
# PR #1203 Retro — fix(websocket): store _recovery_html after broadcast renders (#1202)

**Note**: This retrospective is a **backfill** written during the v0.9.5 milestone retro. The PR shipped on 2026-04-29 without a pipeline-run Stage 14 retro comment — this is the v0.9.0 retro arc's "outer-loop integrity" failure mode (Action Tracker #198 / GitHub #1212 in v0.9.5 retro). Findings below are reverse-engineered from the PR body, the addressed-findings comment in d1ae6771, and the merged diff. Authoritative for posterity but not as rich as a real-time retro.

**Merged**: 2026-04-29T03:19:46Z
**Quality rating** (estimated post-hoc): 4/5

## What likely worked

- **Narrow surgical fix**: one-location change in `python/djust/websocket.py:server_push` mirroring the existing `handle_event` pattern at `websocket.py:3271`. The two paths now have parallel structure, which is the right shape for "bug fix preserves invariants".
- **Concrete downstream context in PR body**: the author cited a real production failure trace — admin-dashboard broadcast pushes on a document upload causing a full-page reload. Made the bug concrete and the fix obviously needed.
- **Addressed-findings cycle worked**: Stage 11 review surfaced 3 specific issues (missing CHANGELOG, verbose comment, no-patches branch test gap) — all addressed in d1ae6771 before merge. The `test_no_patches_does_not_clobber_recovery_html` test in particular locks in a future-regression boundary that the original PR shipped without.
- **Pre-push hooks ran clean** (per addressed-findings comment): full pytest suite green, secrets, CHANGELOG count, noqa.

## What likely didn't work

- **Stage 14 retro never fired**. This is the core observation. Whether pipeline-run was used at all (vs. a manual merge) is unknown without the state file. If it was, the retro-artifact gate at `pipeline-run` Stage 14 should have failed before `completed_at` was set and prevented the silent dropout.
- **Original PR shipped without a CHANGELOG entry** — Stage 11 caught this. Pre-push hook `check-changelog-test-counts` is supposed to surface missing CHANGELOG entries on `fix:`/`feat:` commits; either the hook didn't fire or the prefix didn't trigger it.

## Lessons (carried forward to v0.9.5 milestone retro)

- **Retro gate should be a hard pre-merge check, not just a Stage 14 self-report**. See Action Tracker #193 / GitHub #1212.
- **`server_push` and `handle_event` should share more invariants**. The fact that one updated `_recovery_html` and the other didn't means they were diverging implementations of the same logical operation. Worth a refactor to a shared post-render helper, OR an explicit invariant test that asserts both paths populate recovery state.

## Review stats (estimated)

| Metric | Value |
|--------|-------|
| Lines added | +78 |
| Lines deleted | -0 |
| Tests added | 3 (originally 2; +1 after review) |
| 🔴 Findings | 0 |
| 🟡 Findings | 3 (CHANGELOG, comment trim, no-patches branch test) — all fixed pre-merge |
| Commits | 2 (initial + d1ae6771 addressing review) |

🤖 Backfill — written during v0.9.5 milestone retro 2026-04-30.

PR #181, opened 2026-01-31
feat: add event replay button to Events tab (#177)

johnrtipton:
## Summary

Add a replay button to each event in the Debug Toolbar Events tab, allowing developers to re-trigger events for debugging.

## Changes

- Added replay button (⟳) to each event header (only for events with a handler/name)
- `replayEvent()` method sends event through `window.liveView.sendEvent()` with original params
- Inline status feedback: pending (⏳), success (✓), error (✗) with auto-clear after 2s
- CSS styles for replay button with hover, pending, success, and error states

Closes #177

## Type of Change

- [x] New feature (non-breaking change that adds functionality)

## Test Plan

- [x] Existing tests pass (`make test`)
- [x] New tests added for changed functionality

New JS unit tests in `tests/js/debug_event_replay.test.js` covering:
- Replay with handler and params
- Replay with handler but no params
- Replay using name field fallback
- Error for events without handler/name
- Error for invalid index
- Error when no WebSocket connection
- Error when sendEvent returns false

## Checklist

- [x] Code follows project conventions
- [x] Self-reviewed for correctness, edge cases, and security
- [x] CHANGELOG.md updated
- [x] Documentation updated (docs/DEBUG_PANEL.md)

## Related Issues

Closes #177

PR #180, opened 2026-01-31
feat: add event filtering controls to Events tab (#176)

johnrtipton:
## Summary

Add filter controls to the Events tab in the debug toolbar so users can quickly find specific events in the history.

## Changes

- Added filter bar with text input for name search (substring match, case-insensitive)
- Added status dropdown filter (All / Errors only / Success only)
- Added clear button that resets all filters
- Added match count display (filtered / total)
- Filter state persists when switching between tabs via `state.filters`
- Empty state shown when no events match filters
- Added CSS styles for filter bar components
- Initialized `eventName` and `eventStatus` filter keys in panel core state

Closes #176

## Type of Change

- [x] New feature (non-breaking change that adds functionality)

## Test Plan

- [x] Existing tests pass (`make test`)
- [x] New tests added for changed functionality
- [ ] Manual testing performed (describe below)

New JS unit tests in `tests/js/debug_event_filter.test.js` covering:
- No filters returns all events
- Name substring filtering (case-insensitive)
- Multiple match filtering
- Error-only and success-only status filters
- Combined name + status filters
- Empty results
- Fallback to "unknown" for unnamed events
- Empty event list handling

## Checklist

- [x] Code follows project conventions
- [x] Self-reviewed for correctness, edge cases, and security
- [x] CHANGELOG.md updated
- [x] Documentation updated (docs/DEBUG_PANEL.md)

## Related Issues

Closes #176

PR #1206, opened 2026-04-30
fix: list[Model] in user-overridden get_context_data now triggers VDOM patches (#1205)

### LINKED ISSUE #1205: JIT serializer silently degrades when context value is list[Model] instead of QuerySet
## Summary

When a view passes a **materialized list of model instances** (`ctx["items"] = list(qs)`) instead of a **QuerySet** (`ctx["items"] = qs`), the JIT serializer silently falls back to `str(model_instance)` for each item. Template conditionals that access model fields (e.g. `{% if item.completed %}`) always evaluate falsy, the template renders identically on every event, and the VDOM diff produces **zero patches**. No warning or error is raised.

## Reproducer

```python
# ❌ Breaks VDOM patching — JIT serializer hits _lazy_serialize_context → str() fallback
ctx["tasks"] = list(ClaimTask.objects.filter(claim=claim).select_related("created_by")[:50])

# ✅ Works correctly — JIT serializer uses the full Rust serialize_queryset path
ctx["tasks"] = ClaimTask.objects.filter(claim=claim).select_related("created_by")[:50]
```

Template:
```django
{% for task in tasks %}
  {% if task.completed %}          {# ← always falsy with list[Model] #}
    <i class="fa-check-square"></i>
  {% else %}
    <i class="fa-square"></i>
  {% endif %}
{% endfor %}
```

With the `list()` form, every render produces identical HTML regardless of `task.completed` in the DB. With the raw QuerySet, VDOM diffs work correctly.

## Root cause

In `djust/mixins/jit.py`, `_lazy_serialize_context` handles list values with:

```python
if isinstance(value, (list, tuple)):
    return [serialize_value(item) for item in value]
```

`serialize_value` hits the Model branch:

```python
if isinstance(value, Model):
    return str(value)   # ← loses all field data
```

By contrast, when `claim_tasks` is a QuerySet, the dedicated `_jit_auto_serialize_queryset` path runs, calling `serialize_queryset(items, paths_for_var)` with the Rust serializer which extracts all template-accessed fields.

## Observed impact

Diagnosed via `djust-browser` MCP:

- `patch_count: 0` on every event despite confirmed DB writes
- `_debug.variables.claim_tasks.value` showed only `__str__` strings in the serialized context
- Switching list → QuerySet: `patch_count: 7` immediately, checkbox toggles live

## Expected behavior

One of:

1. **Warning + degrade gracefully**: When a `list[Model]` is detected in context, emit a `djust.W0xx` warning (`"ctx['tasks'] contains a list of Model instances; pass a QuerySet for JIT field serialization"`) and fall back to the existing `str()` path (so it still works, just without VDOM reactivity).

2. **Detect and re-route**: If all items in the list are Model instances of the same class, apply the same field serialization the QuerySet path uses — construct a virtual `paths_for_var` from the JIT template analysis and serialize each item.

Option 1 is safer and easier to implement. Option 2 would be a fully transparent fix.

## Why this is hard to notice

- The template renders correctly on **initial page load** (the HTTP render uses `render_live_view()` which calls `get_context_data()` and then sets `_cached_context` — the list items happen to have their `__str__` matching the template output if conditions happen to be falsy by default).
- Zero patches on events looks like "nothing changed" which is correct behavior when the view has no state changes — it requires knowing to look for `patch_count: 0` combined with a confirmed DB mutation.
- `list(qs)` is idiomatic Python/Django for avoiding double-evaluation, and is commonly used even in documentation examples.

## Workaround

Pass the raw QuerySet: `ctx["items"] = qs` not `ctx["items"] = list(qs)`.

## Environment

- djust 0.9.0
- Django 5.2.13
- Python 3.12

🤖 Generated with [Claude Code](https://claude.com/claude-code) — diagnosed via djust-browser MCP + daphne log analysis

johnrtipton:
## Summary

Closes #1205.

When a view's `get_context_data` override sets `ctx["tasks"] = list(qs)` *after* calling `super().get_context_data()`, the JIT auto-serialization pipeline runs inside `super()` and never sees the user-added value. The raw `list[Model]` then flows through `_sync_state_to_rust`, where change-detection compares list elements via Python `==` — which delegates to `Model.__eq__` (pk-only). In-place field mutations (`is_active` toggle, `completed` flip) don't change `pk`, so the comparison returns equal, the key is never added to the diff-context, Rust never receives the new state, and the rendered HTML is byte-identical on every event despite confirmed DB writes.

Symptom from the issue reporter: `patch_count: 0` on every event, `_debug.variables.tasks.value` shows only `__str__` strings.

## Fix

`_sync_state_to_rust` (in `python/djust/mixins/rust_bridge.py`) now runs a defensive normalize pass over `full_context` immediately after fetching it, converting any `list[Model]` / `Model` / `QuerySet` value to dicts via `normalize_django_value`. After normalization, change-detection compares `list[dict] != list[dict]` element-wise via `dict.__eq__` (structural), correctly catching field mutations. Idempotent on already-serialized values.

Also removed dead `_lazy_serialize_context` method from `python/djust/mixins/jit.py` (zero call sites — was misleadingly cited as the bug location in the issue, the dead `str(model)` fallback inside matched the reported symptom but was never wired up).

## Issue's analysis vs actual root cause

The issue reporter pointed at `_lazy_serialize_context` in `python/djust/mixins/jit.py:328` as the culprit. **That function was dead code** — never called anywhere. The actual bug path is the `_sync_state_to_rust` change-detection comparison described above. The reporter's diagnostic data (`patch_count: 0`, `__str__` strings) was real, but the proposed fix location was wrong. This PR removes the dead code so future investigators don't fall into the same misdirection.

## Files

- `python/djust/mixins/rust_bridge.py` — defensive normalize pass at top of `_sync_state_to_rust`
- `python/djust/mixins/jit.py` — removed dead `_lazy_serialize_context` method + docstring reference
- `tests/unit/test_list_model_diff_1205.py` — 7 regression tests
- `CHANGELOG.md` — `[Unreleased]` entry under `### Fixed`

## Test plan

- [x] Reproducer test (`test_user_override_with_list_qs_propagates_field_mutations`) FAILS on main, PASSES with fix (TDD red→green confirmed)
- [x] Test for raw QuerySet branch (`test_raw_queryset_in_user_override_still_works`) — sanity that existing path doesn't regress
- [x] Test for single Model (`test_single_model_in_user_override`) — covers `isinstance(_val, Model)` arm
- [x] Idempotency test (`test_normalize_idempotent_on_already_serialized`) — JIT-pre-processed values must not double-process
- [x] Edge cases: empty list, mixed list with Model first
- [x] Dead-code removal locked in (`test_lazy_serialize_context_does_not_exist`)
- [x] Full pytest suite: 4660 passed, 14 skipped, 0 failed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,34 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- **JIT serializer silently degrades when context value is `list[Model]`
+  (#1205).** When a view's `get_context_data` override sets
+  `ctx["tasks"] = list(qs)` *after* calling `super().get_context_data()`,
+  the JIT auto-serialization pipeline runs inside `super()` and never sees
+  the user-added value. The raw `list[Model]` then flows through
+  `_sync_state_to_rust`, where change-detection compares list elements via
+  Python `==` — which delegates to `Model.__eq__` (pk-only). In-place field
+  mutations (`is_active` toggle, `completed` flip) don't change `pk`, so
+  the comparison returns equal, the key is never added to the diff-context,
+  Rust never receives the new state, and the rendered HTML is byte-identical
+  on every event despite confirmed DB writes. Symptom from the issue
+  reporter: `patch_count: 0` on every event, `_debug.variables.tasks.value`
+  shows only `__str__` strings. Fix: `_sync_state_to_rust` (in
+  `python/djust/mixins/rust_bridge.py`) now runs a defensive normalize pass
+  over `full_context` immediately after fetching it, converting any
+  `list[Model]` / `Model` / `QuerySet` value to dicts via
+  `normalize_django_value`. After normalization, change-detection compares
+  `list[dict] != list[dict]` element-wise via `dict.__eq__` (structural),
+  correctly catching field mutations. Idempotent on already-serialized
+  values. Also removed dead `_lazy_serialize_context` method from
+  `python/djust/mixins/jit.py` (zero call sites — was misleadingly cited as
+  the bug location in the issue). 7 regression cases in
+  `TestListModelDiff1205` and `TestLazySerializeContextRemoved` lock down
+  the user-override patterns (`list[Model]`, single Model, raw QuerySet)
+  plus idempotency and edge cases (empty list, mixed list).
+
 ## [0.9.0] - 2026-04-29
 
 The "Time Travel" release — the biggest release since 0.3.0, two years of
--- a/python/djust/mixins/jit.py
+++ b/python/djust/mixins/jit.py
@@ -82,7 +82,7 @@ except ImportError:
 
 
 class JITMixin:
-    """JIT serialization: _jit_serialize_queryset, _jit_serialize_model, _lazy_serialize_context, _get_template_content."""
+    """JIT serialization: _jit_serialize_queryset, _jit_serialize_model, _get_template_content."""
 
     def _get_template_content(self) -> Optional[str]:
         """
@@ -324,45 +324,3 @@ class JITMixin:
         except Exception as e:
             logger.debug("JIT serialization failed for %s: %s", variable_name, e)
             return normalize_django_value(obj)
-
-    def _lazy_serialize_context(self, context: dict) -> dict:
-        """
-        Lazy serialization: only serialize values that need conversion.
-        """
-        from ..components.base import Component, LiveComponent
-        from django.db.models import Model
-        from datetime import datetime, date, time
-        from decimal import Decimal
-        from uuid import UUID
-
-        def serialize_value(value):
-            if value is None or isinstance(value, (str, int, float, bool)):
-                return value
-
-            if isinstance(value, (list, tuple)):
-                return [serialize_value(item) for item in value]
-
-            if isinstance(value, dict):
-                return {k: serialize_value(v) for k, v in value.items()}
-
-            if isinstance(value, (Component, LiveComponent)):
-                return {"render": str(value.render())}
-
-            if isinstance(value, Model):
-                return str(value)
-
-            if isinstance(value, (datetime, date)):
-                return value.isoformat()
-
-            if isinstance(value, time):
-                return value.isoformat()
-
-            if isinstance(value, (Decimal, UUID)):
-                return str(value)
-
-            try:
-                return normalize_django_value(value)
-            except (TypeError, ValueError):
-                return str(value)
-
-        return {k: serialize_value(v) for k, v in context.items()}
--- a/python/djust/mixins/rust_bridge.py
+++ b/python/djust/mixins/rust_bridge.py
@@ -9,7 +9,7 @@ from urllib.parse import parse_qs, urlencode
 
 from ..security import sanitize_for_log
 from ..serialization import normalize_django_value
-from ..utils import get_template_dirs
+from ..utils import get_template_dirs, is_model_list
 
 logger = logging.getLogger(__name__)
 
@@ -321,10 +321,26 @@ class RustBridgeMixin:
         if self._rust_view:
             from ..components.base import Component, LiveComponent
             from django import forms
+            from django.db.models import Model, QuerySet
 
             full_context = (
                 preloaded_context if preloaded_context is not None else self.get_context_data()
             )
+
+            # Defensive normalize pass for DB values added after super() (#1205).
+            # When a user overrides ``get_context_data`` and sets
+            # ``ctx["tasks"] = list(qs)`` AFTER calling ``super()``, the JIT
+            # pipeline never sees the list[Model] — and downstream change-
+            # detection would compare list[Model] via Model.__eq__ (pk-only),
+            # missing in-place field mutations. Normalize raw DB values to
+            # dicts here so the comparison below sees field-level changes.
+            for _key, _val in list(full_context.items()):
+                if isinstance(_val, Model):
+                    full_context[_key] = normalize_django_value(_val)
+                elif isinstance(_val, QuerySet):
+                    full_context[_key] = [normalize_django_value(_item) for _item in _val]
+                elif is_model_list(_val):
+                    full_context[_key] = [normalize_django_value(_item) for _item in _val]
 
             # Ensure csrf_token is available for {% csrf_token %} tag (#696).
             # Cache it to avoid creating a new string object each call,
--- a/tests/unit/test_list_model_diff_1205.py
+++ b/tests/unit/test_list_model_diff_1205.py
@@ -1,0 +1,251 @@
+"""Regression: ``list[Model]`` in user-overridden ``get_context_data`` must
+produce VDOM patches when model fields mutate (#1205).
+
+The bug
+-------
+
+When a view's ``get_context_data`` override sets ``ctx["tasks"] = list(qs)``
+*after* calling ``super().get_context_data()``, the JIT auto-serialization
+pipeline runs inside ``super()`` and never sees the user-added ``tasks``.
+The list of raw ``Model`` instances flows through ``_sync_state_to_rust``,
+where change-detection compares ``list[Model]`` via Python ``==``. ``==`` on
+lists delegates to per-element ``Model.__eq__``, which only compares by
+``pk``. In-place field mutations (``is_active`` toggling, ``completed``
+flipping, etc.) don't change ``pk``, so the list compares as "unchanged"
+and Rust never receives the new state. Result: zero VDOM patches, template
+renders identically forever.
+
+The fix (in ``rust_bridge.py``) is a defensive normalize pass: any
+``list[Model]`` / ``Model`` / ``QuerySet`` value that snuck past the JIT
+pipeline gets serialized to ``list[dict]`` / ``dict`` via
+``normalize_django_value`` before change-detection runs.
+
+These tests lock down the bug so it can't silently regress.
+"""
+
+from __future__ import annotations
+
+import pytest
+from django.contrib.auth.models import User
+
+from djust.decorators import event_handler
+from djust.live_view import LiveView
+from djust.testing import LiveViewTestClient
+
+
+# ---------------------------------------------------------------------------
+# Primary regression: list(qs) in user override propagates field mutations
+# ---------------------------------------------------------------------------
+
+
+@pytest.mark.django_db
+class TestListModelDiff1205:
+    def test_user_override_with_list_qs_propagates_field_mutations(self):
+        """``ctx['tasks'] = list(qs)`` after ``super()`` must trigger a real
+        re-render when a model field mutates."""
+
+        class TaskView(LiveView):
+            template = (
+                "<div>"
+                "{% for task in tasks %}"
+                "<li data-pk='{{ task.id }}'>"
+                "{% if task.is_active %}A{% else %}I{% endif %}{{ task.username }}"
+                "</li>"
+                "{% endfor %}"
+                "</div>"
+            )
+
+            def mount(self, request, **kwargs):
+                self._users = [
+                    User(username="alice", pk=1, is_active=True),
+                    User(username="bob", pk=2, is_active=False),
+                ]
+
+            @event_handler
+            def toggle(self, pk: int = 1, **kwargs):
+                for u in self._users:
+                    if u.pk == pk:
+                        u.is_active = not u.is_active
+
+            def get_context_data(self, **kwargs):
+                ctx = super().get_context_data(**kwargs)
+                ctx["tasks"] = list(self._users)  # the bug pattern
+                return ctx
+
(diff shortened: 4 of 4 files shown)


johnrtipton (commented):
## Code Review — PR #1206

### Summary

This PR fixes #1205 (JIT serializer silently degrades when context value is `list[Model]`). The fix adds a defensive normalize pass at the top of `_sync_state_to_rust` in `python/djust/mixins/rust_bridge.py:330-343`, converting any `Model` / `QuerySet` / `list[Model]` value in the post-`get_context_data()` context to dicts via `normalize_django_value`. After normalization, list change-detection compares `list[dict] != list[dict]` element-wise (structural via `dict.__eq__`), correctly catching field mutations that the prior `Model.__eq__` (pk-only) path missed.

The plan correctly identified that the issue reporter's pointer at `_lazy_serialize_context` was a misdirection — that method is dead code. Removing it (jit.py, lines 328-368 in the old file) is a nice cleanup that prevents future investigators from chasing the same red herring, and a `hasattr`-based regression test locks the removal in.

### Plan Compliance

All four planned deliverables landed:
- `python/djust/mixins/rust_bridge.py` — defensive normalize pass at lines 330-343 (matches Option A in plan)
- `python/djust/mixins/jit.py` — `_lazy_serialize_context` removed; docstring at line 85 updated to drop the reference
- `tests/unit/test_list_model_diff_1205.py` — 7 regression tests (plan listed 6, the 7th `test_raw_queryset_in_user_override_still_works` was added per self-review feedback to lock down the QuerySet branch — good call)
- `CHANGELOG.md` — comprehensive `[Unreleased]` entry under `### Fixed` accurately describes root cause, fix mechanism, and test coverage

The plan's Stage 5 acceptance criteria are all satisfied.

### Code Quality

Strengths:
- Comment block at `rust_bridge.py:330-336` is excellent — explains *why* the normalize pass exists and ties it back to #1205. Future readers won't accidentally remove it as "redundant with JIT pipeline".
- Idempotency design is correct: `is_model_list([dict, dict, ...])` returns False because it checks `isinstance(value[0], models.Model)` (utils.py:16), and `isinstance(dict, Model)` is False. Already-serialized values flow through unchanged.
- Underscore-prefixed loop locals (`_key`, `_val`, `_item`) avoid shadowing the surrounding scope's variable names — good hygiene given the long method context.
- Type annotation imports are properly localized inside the `if self._rust_view:` block — avoids module-level Django-import overhead.

Minor suggestions (non-blocking):
1. `rust_bridge.py:340-343` — the QuerySet and `is_model_list` arms produce identical list comprehensions. A tiny refactor would deduplicate:
   ```python
   elif isinstance(_val, QuerySet) or is_model_list(_val):
       full_context[_key] = [normalize_django_value(_item) for _item in _val]
   ```
   Saves one branch and three lines without changing behavior. Not load-bearing — the current form reads as clearly.

2. `tests/unit/test_list_model_diff_1205.py:171-181` — `test_normalize_idempotent_on_already_serialized` mounts a view with a class-level `tasks = [User(...), User(...)]`. This works but couples the test to JIT pipeline behavior on class-level attrs. If the JIT short-circuit for class-level lists ever changes, this test could pass for the wrong reason. Consider adding an explicit assertion that the second render's `dom_changes` count is 0 (no patches) to make the idempotency claim verifiable rather than implicit.

### Testing

Test suite results: **33/33 passed** locally for the targeted suite (`test_list_model_diff_1205.py`, `test_jit_short_circuit.py`, `test_rust_vdom_safe_diff_783.py`). PR body claims full pytest suite is green (4660 passed, 14 skipped, 0 failed).

Edge cases I considered and traced through the diff:
- **Empty list**: `is_model_list([])` returns False (utils.py:16 requires `len(value) > 0`). Correctly skipped. Locked by `test_empty_list_normalize_safe`.
- **Heterogeneous list `[Model, dict]`**: `is_model_list` triggers on first element, then `[normalize_django_value(_item) for _item in _val]` handles each item independently (dict passes through unchanged via the dict-recurse arm at serialization.py:421). Locked by `test_mixed_list_with_model_first`.
- **Heterogeneous list `[dict, Model]`** (Model NOT first): `is_model_list` returns False because `isinstance(value[0], Model)` is False. The Model in position 1 is NOT normalized — it would fall through to change-detection as a list with a raw Model embedded. **This is a gap not covered by tests.** In practice rare (mixed lists are rare; the dict-first variant rarer still), but worth a `test_mixed_list_with_dict_first` or an explicit comment that this shape is unsupported. Question for the author: was this considered, and if so, what's the intended contract?
- **Nested dict-of-lists `{"groups": [[Model, Model], [Model]]}`**: The normalize pass runs `for _key, _val in full_context.items()` — only top-level. A `list[list[Model]]` at the top level fails `is_model_list` (first element is a list, not a Model) and is NOT normalized. Field mutations on nested models would hit the same bug. Probably out of scope for this fix, but worth noting in CHANGELOG or a follow-up issue.
- **Idempotency on already-normalized list[dict]**: Verified above — `is_model_list` returns False for `list[dict]`, so the pass is a no-op.

### Security

No new attack surface. The values flowing through `normalize_django_value` are server-controlled (came from `get_context_data()`) and the function is the same one already used by the rest of the JIT pipeline. The new code paths don't touch user input, don't construct HTML/SQL, don't log f-strings, and don't introduce `mark_safe`. Clean.

### Performance

The normalize pass adds one `for _key, _val in list(full_context.items())` loop with three `isinstance`/`is_model_list` checks per key on every `_sync_state_to_rust` call. Cost analysis:
- Most keys are `(str, int, dict, csrf_token-string)` — three short-circuit `isinstance` checks each (Model, QuerySet) plus one `is_model_list` call. `is_model_list` itself is `isinstance(value, list) and len(value) > 0 and isinstance(value[0], models.Model)` — short-circuits on non-list almost immediately.
- For a typical context with ~10 keys, this is ~30-40 isinstance calls per sync. Negligible compared to the surrounding work (Rust state sync, change detection, render).
- For contexts that DO contain Models/QuerySets, the cost is whatever `normalize_django_value` already costs (which the JIT pipeline pays anyway in the non-override case).

The plan's risk #1 (perf) is appropriately mitigated by Python's short-circuit evaluation on isinstance chains. No concern.

### Architecture

Concern raised in the review prompt: "should the fix be in `get_context_data` instead of `_sync_state_to_rust`?" — I've thought about this and the chosen location is correct:
- `get_context_data()` is user-overridden; putting the normalize pass there would require either monkey-patching the user's override (fragile) or running a post-hook after the user's override (already what `_sync_state_to_rust` effectively does).
- `_sync_state_to_rust` is the funnel that EVERY render path hits before sending state to Rust — both WebSocket and HTTP-fallback paths. It's the right defense-in-depth layer.
- The plan correctly cited this in Option B's rejection.

### Specific findings with file:line citations

- **Good defensive design** at `python/djust/mixins/rust_bridge.py:330-336` — comment block ties fix to bug context, prevents future "this looks redundant" removal.
- **Good test ordering** at `tests/unit/test_list_model_diff_1205.py:42-87` — primary regression test runs first; failure point is the assertion on line 84 (`html_before != html_after`), which is exactly the bug invariant.
- **Question** at `python/djust/mixins/rust_bridge.py:342` — is the `[Model, dict]`-first-Model arm of `is_model_list` intentionally narrower than the QuerySet arm? See edge cases above — `[dict, Model]` (dict first) is silently NOT normalized.
- **Minor** at `python/djust/mixins/jit.py:85` — docstring updated to drop `_lazy_serialize_context`. Good. Could mention the bug-class one-liner for grep-discoverability ("see #1205 for the dead-code retro") but optional.

### Verdict

The fix is correct, well-tested, narrowly scoped, and accompanied by exemplary documentation. Plan compliance is full. The minor suggestions above (test assertion strengthening, edge-case gap on dict-first heterogeneous lists, optional dedup) are non-blocking. The dead-code removal is a clean win that prevents the same misdirection from happening again.

Approving.

johnrtipton:
## Retrospective — PR #1206 (#1205 list[Model] VDOM degradation)

### Quality: 4/5

Solid bugfix. Reproducer-first TDD; root cause traced past the issue reporter's misdirection; defensive normalize landed in the right funnel; 7 regression tests including idempotency + edge cases; dead code removed; CHANGELOG explicit about user-visible symptom and mechanism. CI green across 13 checks (Py 3.12/3.13/3.14, Rust, JS, Playwright, security). Two-commit shape (impl+tests / docs+CHANGELOG) preserved per Action Tracker #181.

Half a point off for two non-blocking gaps the reviewer surfaced post-merge:

1. **Heterogeneous `[dict, Model]` dict-first** is not normalized — `is_model_list` checks `isinstance(value[0], Model)`. Field mutations on the second-position Model would still escape change detection. Genuinely rare shape; not the user's bug; documented `list[Model]` contract implies homogeneous. Worth a follow-up if someone hits it in practice.
2. **Idempotency test** asserts no exception but not "zero patches". Stronger assertion would lock the contract more tightly.

### What went well

- **Reproducer-driven**. Issue reporter pointed at `_lazy_serialize_context` (dead code). Writing `LiveViewTestClient` reproducer before touching the implementation surfaced the actual path (`_sync_state_to_rust` change-detection comparing `list[Model]` via `Model.__eq__` pk-only). Without the reproducer, the obvious move would have been to "fix" the dead code where the reporter pointed.
- **Trust-but-verify the issue reporter**. The reporter's diagnostic data was real (`patch_count: 0`, `__str__` strings) but their proposed fix location was wrong. Re-running the diagnosis from scratch (grep for callers, reproduce in isolation, trace the actual data flow) caught the misdirection.
- **Two-commit shape held under one-checkout discipline**. Single agent, single working tree → no CHANGELOG cross-contamination. Commits 2212dff4 (impl + tests) and cfb118ea (CHANGELOG) landed clean on the first try.
- **Subagent self-review surfaced a real test gap pre-merge**. Reviewer flagged the missing QuerySet-branch test; added `test_raw_queryset_in_user_override_still_works` before opening the PR. Stage 11 review then approved cleanly.

### Lessons learned

1. **Issue-reporter analysis ≠ root cause.** The reporter cited `_lazy_serialize_context` (jit.py:328) — the function literally contains the buggy `str(model)` fallback that matches their symptom. But it's never called. The real bug is upstream in change detection. Trace bug reports from observable symptom to actual code path; don't trust the reporter's path-of-causation.

2. **Dead code that resembles the bug is dangerous.** `_lazy_serialize_context` had zero call sites but its `str(model)` fallback was a perfect match for the reported symptom. Anyone investigating with a "find code that produces __str__ output" mindset would land on it and "fix" it — without changing observable behavior. Removing it as part of this PR breaks that trap.

3. **`list[Model]` is a perf-equivalent shape that misbehaves under change detection.** Users reach for `list(qs)` to "freeze" a queryset for safety; idiomatic Python. The bug isn't that they did the wrong thing — it's that the framework's change detection silently degrades under a documented Python pattern. The fix has to be in the framework, not in user docs.

4. **`Model.__eq__` is pk-only is a design choice with subtle blast radius.** Django Model equality compares by primary key; in-place field mutations don't affect equality. This is fine for set membership and ORM caching but breaks any framework code that uses Python `==` to detect "did this value change". Worth a framework-level note: anywhere we compare user-controlled context values via `==`, Models need pre-normalization to dicts.

### IDEA: improvements to consider

- **IDEA**: Add a Stage 4 checklist item that requires bug-report claims to be VERIFIED via independent reproducer before the plan locks in a fix location. Would have saved ~15 min of context-window wasted on the dead code path before the actual reproducer surfaced the right code path.
- **IDEA**: Linter rule that flags `_methodname` definitions with zero call sites in the codebase (excluding the file they're defined in). `_lazy_serialize_context` would have been caught at PR time the first time the code reviewer looked. Could be a `pre-push` `vulture`-based check (already in the dependency tree based on similar djust patterns).
- **IDEA**: Document the `list[Model]` contract explicitly in `docs/STATE_MANAGEMENT_API.md` or wherever JIT serialization is described. Specifically: "the framework normalizes `list[Model]`, single `Model`, and `QuerySet` values; `list[list[Model]]` and dict-first heterogeneous lists `[dict, Model]` are NOT normalized — convert to a homogeneous shape if you need diff propagation." Locks the supported shapes in writing.
- **IDEA**: Strengthen `test_normalize_idempotent_on_already_serialized` with an explicit `dom_changes` count assertion (per reviewer suggestion) — would lock idempotency more tightly than the current "no exception" implicit assertion.

### Process notes

- Stage 4 (Planning) wasted ~10 min on the dead-code path before writing the reproducer. The reproducer was the load-bearing artifact; should have been written FIRST in Stage 4, not as the climax. Plan template's "verify the bug with a failing reproducer test" item is correct ordering — followed it on the second attempt.
- Stage 11 reviewer's APPROVE with non-blocking suggestions was the right call. Two of three suggestions (heterogeneous dict-first, idempotency-test strengthening) are genuinely follow-up material; the third (branch dedup) is a style preference. Merging on APPROVE with documented follow-ups is faster than blocking on cosmetic improvements.
- One-implementer-one-checkout rule held; no CHANGELOG cross-contamination. v0.9.1 retro #180/#1172 lesson applied successfully.

Issue #196, opened 2026-02-01
Debug panel Events/Patches tabs never capture data: WebSocket hooks applied too late


johnrtipton:
## Bug Description

The Events, Patches, and Network tabs in the debug panel always show "No events captured yet" even after triggering interactions (clicking buttons, submitting forms, etc.) that produce WebSocket events and patches.

## Root Cause

`hookIntoLiveView()` (debug-panel.js line 2595) patches `WebSocket.prototype.send`, `WebSocket.prototype.addEventListener`, and the `onmessage` property setter. However, this patching happens AFTER the LiveView has already:

1. Created its WebSocket instance
2. Bound its `onmessage` handler
3. Connected and started receiving messages

The monkey-patching of `WebSocket.prototype` methods only affects NEW WebSocket instances or NEW event listener registrations. The existing LiveView WebSocket connection is already fully set up by the time the debug panel initializes.

Console log order confirms this:
```
[LiveView] WebSocket connected
[LiveView] Received: connect ...
[LiveView] View mounted: ...
[djust] Developer Bar initialized  (too late!)
```

## Steps to Reproduce

1. Navigate to /examples/ on djust.org
2. Open debug panel
3. Click "Increment" on the counter
4. Events tab still shows "No events captured yet"
5. Patches tab shows 0 patches
6. Network tab shows no messages

## Fix Options

1. **Hook earlier**: Move WebSocket prototype patching to run BEFORE LiveView initialization (e.g., inline script in `<head>`)
2. **Hook existing connections**: After patching prototypes, also find any already-created WebSocket instances and wrap their handlers
3. **Use LiveView API**: Instead of monkey-patching WebSocket, hook into the LiveView client's event system if it exposes hooks/callbacks

## Files

- `python/djust/static/djust/debug-panel.js` (`hookIntoLiveView()`, line 2595)

PR #1217, opened 2026-04-30
test: strengthen idempotency test with zero-patch assertion + LiveViewTestClient.render_with_patches() (#1208)

### LINKED ISSUE #1207: tech-debt: support heterogeneous and nested list[Model] shapes in change-detection normalize pass
Follow-up from PR #1206 (#1205 fix). The defensive normalize pass added in `python/djust/mixins/rust_bridge.py:_sync_state_to_rust` correctly handles the **homogeneous** `list[Model]` shape that the original bug reporter hit, but two related shapes are NOT covered:

## Gap 1 — Heterogeneous `[dict, Model]` (Model NOT first)

`python/djust/utils.py:is_model_list` returns `False` when `value[0]` is not a Model:

```python
def is_model_list(value: Any) -> bool:
    return isinstance(value, list) and len(value) > 0 and isinstance(value[0], models.Model)
```

So `[{"foo": "bar"}, <Model: alice>]` falls through the normalize pass un-serialized. The Model in position 1 retains its raw form; in-place field mutations escape change-detection (same root cause as #1205 — `Model.__eq__` is pk-only).

**Repro pattern (untested):**
```python
    ctx["items"] = [{"label": "header"}, *list(qs)]  # dict first, then Models
    return ctx
```

## Gap 2 — Nested `list[list[Model]]`

The normalize pass iterates only top-level keys:

```python
for _key, _val in list(full_context.items()):
    if isinstance(_val, Model): ...
    elif isinstance(_val, QuerySet): ...
    elif is_model_list(_val): ...
```

A `list[list[Model]]` (e.g., grouped tasks: `[[task, task], [task]]`) at the top level fails `is_model_list` because the first element is a list, not a Model. Field mutations on any nested model propagate the same bug as #1205.

**Repro pattern (untested):**
```python
    ctx["task_groups"] = [list(qs.filter(group=g)) for g in groups]
    return ctx
```

## Why these are deferred (not blocking #1205 fix)

- Heterogeneous lists with dict-first ordering are rare in practice (the user's bug was homogeneous `list[Model]`).
- `list[list[Model]]` is a more idiomatic shape for grouped data, but most users reach for a `dict[str, list[Model]]` (which IS handled by `_deep_serialize_dict` in `context.py`) instead.
- Both are documented as out-of-scope by the PR #1206 retro.

## Proposed fix

Make the normalize pass scan the full list (not just `value[0]`) for any Model instance, and recurse into `list[list[...]]`:

```python
def _normalize_db_values(value, depth=0):
    if depth > 3:  # safety bound
        return value
    if isinstance(value, Model):
        return normalize_django_value(value)
    if isinstance(value, QuerySet):
        return [normalize_django_value(item) for item in value]
    if isinstance(value, list) and len(value) > 0:
        # Detect any Model in the list (heterogeneous-safe)
        if any(isinstance(item, Model) for item in value):
            return [normalize_django_value(item) if isinstance(item, Model) else _normalize_db_values(item, depth + 1) for item in value]
        # Nested list — recurse into elements
        if isinstance(value[0], list):
            return [_normalize_db_values(item, depth + 1) for item in value]
    return value
```

Cost: O(N) scan per top-level list (vs. O(1) `is_model_list` first-element check), plus recursion bounded by depth limit. For typical contexts this is still negligible.

## Additional task: document supported shapes

Add to `docs/STATE_MANAGEMENT_API.md` (or wherever JIT serialization is described): "the framework normalizes `list[Model]`, single `Model`, and `QuerySet` values; until #XXXX lands, dict-first heterogeneous lists `[dict, Model]` and nested `list[list[Model]]` are NOT normalized — convert to a homogeneous `list[Model]` or use a top-level `QuerySet` for diff propagation."

## Tests to add

- `test_heterogeneous_list_dict_first_propagates_field_mutations`
- `test_nested_list_of_lists_of_models_propagates_field_mutations`
- Both following the same shape as `test_user_override_with_list_qs_propagates_field_mutations` in `tests/unit/test_list_model_diff_1205.py`.

🤖 Auto-filed from PR #1206 retro.

### LINKED ISSUE #1208: tech-debt: strengthen idempotency test for normalize pass with explicit zero-patch assertion
Follow-up from PR #1206 (#1205 fix). Code review surfaced a minor weakness in `tests/unit/test_list_model_diff_1205.py:test_normalize_idempotent_on_already_serialized`.

## Current state

The test asserts that a second render after a noop event "does not throw" by checking that `"alice"` appears in the rendered HTML. This is a weak assertion — any rendering path that produces "alice" passes, including pathological ones (e.g., re-running JIT serialization redundantly, or sending unnecessary patches).

```python
def test_normalize_idempotent_on_already_serialized(self):
    # ... mounts JITView, renders, calls noop event ...
    html2 = client.render()
    # No change, but render must succeed (idempotency check).
    assert "alice" in html2
```

## Proposed fix

Add an explicit assertion that the noop event produces **zero VDOM patches**, locking idempotency tightly:

```python
def test_normalize_idempotent_on_already_serialized(self):
    # ... existing setup ...
    client.send_event("noop")
    # Idempotency: a noop event with no field mutation should produce no
    # patches. If the normalize pass were double-processing already-
    # serialized values, the prev_containers comparison could spuriously
    # mark the key as changed.
    patches = client.get_last_patches()  # or however the test client exposes them
    assert patches == [] or patches is None, f"Expected no patches on noop, got: {patches}"
```

Need to verify `LiveViewTestClient` exposes patches from the most recent event — check `python/djust/testing.py`. If not exposed, may need to add a thin accessor.

## Why this matters

The current normalize pass relies on three short-circuit checks (`isinstance(_val, Model)`, `isinstance(_val, QuerySet)`, `is_model_list(_val)`) returning False for already-serialized `list[dict]`. A future refactor that subtly broadens any of those checks (e.g., switches `is_model_list` to scan full list per the related #XXXX issue) could accidentally re-process serialized values and inflate change-detection sensitivity. The patch-count assertion catches that immediately.

## Estimated effort

~15 minutes. One test modification, no production code changes.

🤖 Auto-filed from PR #1206 retro / code review.

johnrtipton:
## Summary

Closes #1208. v0.9.5 retro Action Tracker #196.

The existing `test_normalize_idempotent_on_already_serialized` (added in PR #1206) asserted `"alice" in html` — a weak assertion that any rendering path passes. **Strengthen with explicit `patches == []`** to lock the behavioral idempotency invariant.

## What changed

1. **New public test API**: `LiveViewTestClient.render_with_patches()` in `python/djust/testing.py`. Thin wrapper around `view_instance.render_with_diff()` returning `(html, patches_list, version)` with JSON patches parsed into a Python list. Reusable for any future VDOM-diff invariant test.
2. **Strengthened test**: `tests/unit/test_list_model_diff_1205.py::test_normalize_idempotent_on_already_serialized` now asserts zero patches on a noop event.

## Why this matters

If a future refactor broadens `is_model_list` to scan the full list (e.g., to fix #1207 heterogeneous shapes), the regression where already-serialized `list[dict]` gets re-processed would surface here as phantom patches in `prev_containers` comparison. The current weak assertion would have missed that.

## Test plan

- [x] All 7 tests in `test_list_model_diff_1205.py` pass
- [x] Full unit suite green: 4431 passed, 6 skipped
- [x] CHANGELOG entry under `### Added` documents the new public API

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,20 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- **`LiveViewTestClient.render_with_patches()` — VDOM-diff accessor for tests
+  (#1208).** New public method on `djust.testing.LiveViewTestClient` that
+  wraps `view_instance.render_with_diff()` and returns `(html, patches_list,
+  version)` with the JSON patches parsed into a Python list. Empty list when
+  no patches were produced. Reusable for any test that needs to assert on
+  VDOM-diff invariants (e.g. "this noop event must produce zero patches").
+  First user is the strengthened
+  `test_normalize_idempotent_on_already_serialized` regression test in
+  `tests/unit/test_list_model_diff_1205.py`, which now locks the #1206
+  normalize-pass idempotency contract via an explicit `patches == []`
+  assertion instead of the prior weaker "no exception" check.
+
 ### Fixed
 
 - **JIT serializer silently degrades when context value is `list[Model]`
--- a/python/djust/testing.py
+++ b/python/djust/testing.py
@@ -317,6 +317,45 @@ class LiveViewTestClient:
         template = get_template(template_name)
         return template.render(context)
 
+    def render_with_patches(self) -> tuple:
+        """Render and return ``(html, patches_list, version)`` for VDOM-diff
+        assertions in tests.
+
+        Wraps ``view_instance.render_with_diff()`` and parses the JSON patches
+        into a Python list (empty list when no patches were produced).
+
+        Each call advances the diff baseline — i.e., the second call returns
+        patches relative to the first. The typical pattern for a "noop event
+        produces zero patches" assertion is:
+
+        .. code-block:: python
+
+            client.render_with_patches()              # baseline
+            client.send_event("noop")                  # mutate
+            _, patches, _ = client.render_with_patches()
+            assert patches == []
+
+        Returns:
+            ``(html, patches_list, version)`` where ``patches_list`` is a
+            ``list[dict]`` (empty when no patches), parsed from the Rust
+            view's diff JSON.
+
+        Raises:
+            RuntimeError: If view not mounted.
+        """
+        if not self._mounted or not self.view_instance:
+            raise RuntimeError("View not mounted. Call client.mount() first.")
+
+        request = getattr(self.view_instance, "request", None)
+        html, patches_json, version = self.view_instance.render_with_diff(request)
+        if patches_json:
+            import json as _json
+
+            patches_list = _json.loads(patches_json)
+        else:
+            patches_list = []
+        return (html, patches_list, version)
+
     def assert_state(self, **expected: Any) -> None:
         """
         Assert state variables match expected values.
--- a/tests/unit/test_list_model_diff_1205.py
+++ b/tests/unit/test_list_model_diff_1205.py
@@ -162,7 +162,17 @@ class TestListModelDiff1205:
         pass must NOT double-process. ``is_model_list`` returns False for
         ``list[dict]`` (first item isn't a Model), and ``isinstance(dict,
         Model)`` is False — so the pass is a no-op for already-serialized
-        values."""
+        values.
+
+        Strengthened (#1208): the noop event must produce **zero VDOM
+        patches**. If the normalize pass were ever to double-process
+        already-serialized values (e.g. via a future refactor that broadens
+        ``is_model_list`` to scan the full list), the ``prev_containers``
+        comparison in ``_sync_state_to_rust`` could spuriously mark the key
+        as changed and emit phantom patches. This locks the contract tightly
+        — no exception is necessary; behavioral idempotency is the actual
+        invariant.
+        """
 
         class JITView(LiveView):
             template = "<div>{% for t in tasks %}{{ t.username }}{% endfor %}</div>"
@@ -176,16 +186,25 @@ class TestListModelDiff1205:
                 pass
 
         client = LiveViewTestClient(JITView).mount()
-        # First render establishes baseline. Should not throw on the
-        # normalize pass (idempotent on serialized dicts).
+        # First render establishes baseline.
         html = client.render()
         assert "alice" in html and "bob" in html
 
-        # Second render — context unchanged, no field mutation.
+        # Establish VDOM baseline via render_with_patches so the second call
+        # produces patches relative to the baseline.
+        client.render_with_patches()
+
+        # Send a noop event — no field mutation, no state change.
         client.send_event("noop")
-        html2 = client.render()
-        # No change, but render must succeed (idempotency check).
+
+        # Second render after noop: render must succeed AND produce zero
+        # patches. The zero-patches assertion is the load-bearing invariant.
+        html2, patches, _ = client.render_with_patches()
         assert "alice" in html2
+        assert patches == [], (
+            f"Expected zero patches on noop event (idempotency invariant), "
+            f"got {len(patches)} patches: {patches[:3]!r}"
+        )
 
     def test_empty_list_normalize_safe(self):
         """An empty list (not list[Model]) must not trip the normalize pass."""


johnrtipton (commented):
## Review summary: APPROVE

This is a well-crafted test-strengthening PR that converts a weak `"alice" in html` assertion into a precise behavioral invariant via a small, reusable public helper. The new `LiveViewTestClient.render_with_patches()` (testing.py:320-357) is a thin, focused wrapper that mirrors the established `render()` pattern in the same file (testing.py:298-306) — same mounted check, same `request = getattr(self.view_instance, "request", None)` extraction, same RuntimeError shape.

### What I verified

**API shape (testing.py:350)**: `view_instance.render_with_diff(request)` returns `Tuple[str, Optional[str], int]` per `_rust.pyi:713` and is implemented in `mixins/template.py:867`. The `if patches_json:` truthy check at testing.py:351 correctly handles the `None` case from the Rust layer — empty-string and None both yield `patches_list = []`, which is what the docstring promises ("Empty list when no patches were produced"). Verified empirically: a fresh-mount call returns `(html: str, patches: [], version: 1)`.

**Pre-mount RuntimeError (testing.py:346-347)**: Calling `render_with_patches()` before `mount()` raises `RuntimeError("View not mounted. Call client.mount() first.")` — exact same message as `render()` at line 299. Good consistency.

**Test correctness (test_list_model_diff_1205.py:188-207)**: The baseline-then-mutate-then-assert pattern is right. The intermediate `client.render_with_patches()` at line 195 is essential because `render_with_diff()` is stateful — without it, the post-noop call would diff against the very first render and likely produce non-empty patches just from VDOM initialization. The failure message at line 205-206 includes both the count and `patches[:3]!r}` for fast triage.

**Test result**: All 7 tests in `test_list_model_diff_1205.py` pass, and the broader unit suite (1755 tests) is green on this branch.

**No regression surface**: `client.render()` is untouched — the new method is purely additive. Only one test file (`test_list_model_diff_1205.py`) uses `client.render()` in the entire test tree, so the blast radius for the existing API is zero.

### One concern (non-blocking)

The lazy `import json as _json` inside the method (testing.py:352) is consistent with avoiding top-level imports for rarely-used modules, but `json` is already a standard-library module and `testing.py` is a test-helper file (not a hot path) — a top-level `import json` would be marginally cleaner and matches how most stdlib imports are handled elsewhere in the package. That said, the lazy form is harmless and the inline comment explaining "why lazy" isn't strictly necessary either way. Not worth blocking on.

### Edge cases considered

- **`patches_json=None`** → handled correctly: falsy branch sets `patches_list = []`. Verified empirically.
- **`patches_json=""` (empty string)** → also falsy, also yields `[]`. Correct.
- **Pre-mount call** → RuntimeError with helpful message. Verified empirically.
- **Existing `render()` callers** → untouched, no regression possible.
- **CHANGELOG entry** (CHANGELOG.md:10-22) under `### Added` — accurately describes the new public API, cites #1208, names the first user, and explains the "why" (locks idempotency contract instead of weaker "no exception" check).

### Question

Is there a follow-up to add this method to the Sphinx-generated public API docs (e.g. a docstring callout in the `LiveViewTestClient` class-level docs or `docs/website/guides/testing.md` if such a guide exists)? Not required for this PR — but since the CHANGELOG markets it as "reusable for any future VDOM-diff invariant test," surfacing it in user-facing testing docs would help downstream consumers discover it.

APPROVE.

johnrtipton:
## Retrospective — PR #1217 (#1208 idempotency zero-patch assertion + render_with_patches helper)

### Quality: 5/5

Strengthened a weak test assertion AND added a reusable test helper in one tight PR. Two-commit shape (impl+tests / CHANGELOG) per Action Tracker #181. All 7 tests in `test_list_model_diff_1205.py` pass. Full unit suite: 4431 passed, 6 skipped. CI green across 13 checks.

### What went well

- **Issue body anticipated the helper need.** The #1208 body explicitly said "may need to add a thin accessor". That foresight saved a Stage 4 design pass — the implementation choice was framed by the issue author, not discovered mid-PR.
- **`render_with_patches()` shape mirrors `render_with_diff()` exactly.** 3-tuple `(html, patches_list, version)` matches the underlying Rust call. Parses `patches_json` to a `list` so callers don't need to import `json`. Empty list (not None) when no patches — easier `assert patches == []` ergonomics.
- **First user is the strengthened test itself.** The accessor isn't speculative — it lands with a real consumer that exercises both the success path (zero patches) and the realistic test pattern (baseline → mutate → assert).
- **Pre-mount RuntimeError check matches existing `render()` pattern.** No new error handling discipline introduced; consistent with surrounding code.

### Lessons

- **Test helpers added speculatively often die unused.** This one didn't — it shipped with a consumer test in the same PR. Future test-API additions should follow the same pattern: add the accessor + the first test that uses it together.
- **Reusable VDOM-diff invariant tests are now possible.** Future patterns like "this push event should patch only N elements" or "this state mutation should produce exactly these K patches" can use `render_with_patches()` as the building block. Worth surfacing in `docs/website/guides/testing.md` as a follow-up (raised as a question by the reviewer).

### Process notes

- Two-commit shape: `bed90eaa` (impl + tests) + `7a43ce7a` (CHANGELOG) — clean.
- Pre-commit hooks all passed first try; no stash-restore-bounce churn.
- Self-approval blocked → merged with `--admin` per autonomous-merge convention.

PR #1219, opened 2026-04-30
chore(pipeline): reviewer-prompt budget guidelines for Stage 11 (#1211)

johnrtipton:
## Summary

Closes #1211. v0.9.5 retro Action Tracker #192.

Adds explicit word-budget caps and edge-case-enumeration guards to the Stage 11 reviewer subagent_prompt in both `.pipeline-templates/feature-state.json` and `.pipeline-templates/bugfix-state.json`.

## Problem

PR #1201's reviewer-agent stalled at the 10-min watchdog limit mid-tangent on backslash-injection edge cases. Output stream stopped without a verdict. The completed verification was thorough enough to accept as APPROVE in a security-PR context with fixed attack shapes, but in a less-obvious case would have required re-spawn or fallback to manual review.

## Fix

Built budget constraints structurally into the subagent_prompt:

- **Bugfix template**: cap at 250 words. For security-related changes, enumerate at most the attack shapes the plan documented.
- **Feature template**: cap at 350 words. Allow edge cases tied to the planned feature only — not beyond plan scope.

The reviewer agent now receives the budget constraint as part of its structural prompt rather than relying on the operator remembering to add it ad-hoc.

## Test plan

- [x] JSON syntactically valid
- [x] Both templates updated with analogous-but-tailored language (250/350 word caps)
- [x] No code change to djust framework — process improvement only
- [x] No CHANGELOG entry required (chore: prefix, internal pipeline tooling)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1218, opened 2026-04-30
chore(pipeline): require reproducer/artifact before Stage 4 plan finalization (#1210)

johnrtipton:
## Summary

Closes #1210. v0.9.5 retro Action Tracker #191.

Adds a leading mandatory Stage 4 (Planning) checklist item to both `.pipeline-templates/feature-state.json` and `.pipeline-templates/bugfix-state.json` requiring a verified artifact before plan lock-in:

- **For bug reports**: a failing reproducer test exhibiting the reported symptom
- **For security alerts**: a verified read of actual code at the alert-cited line
- **For feature work**: a failing test or doc-snippet exercising desired behavior

The discipline generalizes from two v0.9.5-arc PRs:

- **PR #1206** (#1205 list[Model] fix) burned ~10 min chasing the reporter's cited dead code (`_lazy_serialize_context`) before a reproducer surfaced the actual bug path.
- **PR #1201** (security cleanup) saved ~2 min by reading alert-cited lines BEFORE plan-stage; revealed all 8 `py/log-injection` alerts were FPs.

Same discipline, different surface: trust-but-verify is cheaper than trace-and-rollback. The artifact requirement is now structural — Stage 4 cannot pass without one.

## Test plan

- [x] JSON syntactically valid (`python -c "import json; json.load(open(...))"`)
- [x] No code change to djust framework — process improvement only
- [x] No CHANGELOG entry required (chore: prefix, internal pipeline tooling)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1216, opened 2026-04-30
docs: add Bug-report triage section to CLAUDE.md (#1213)

johnrtipton:
## Summary

Closes #1213. v0.9.5 retro Action Tracker #194.

Adds a "Bug-report triage" section to project `CLAUDE.md` that generalizes the issue-reporter-analysis ≠ root-cause lesson from PR #1206 (#1205 list[Model] VDOM fix).

The principle: trace from observable symptom to actual code path, not from reporter-cited code path to symptom. **Symptom-up beats path-down.**

## Why now

PR #1206 burned ~10 min in Stage 4 chasing the issue reporter's claimed bug location (`_lazy_serialize_context` in `python/djust/mixins/jit.py`) before writing a `LiveViewTestClient` reproducer that surfaced the actual code path (`_sync_state_to_rust` change-detection comparing `list[Model]` via `Model.__eq__` pk-only). The dead-code fallback method literally contained a `str(model)` call matching the reported symptom — a near-perfect misdirection.

This section makes the discipline structural for next time.

## Section placement

Inserted as `## Bug-report triage` **before `## Common Pitfalls`** in CLAUDE.md, matching the conceptual flow: triage discipline → specific pitfalls → retro canonicalizations.

## Test plan

- [x] CLAUDE.md still renders cleanly (Markdown)
- [x] No code references in the new section are typos (jit.py, rust_bridge.py, _sync_state_to_rust all spelled correctly)
- [x] PR #1206 + #1205 cross-references are accurate
- [x] No CI risk (docs-only change)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1220, opened 2026-04-30
chore(tooling): add pre-push check for newly-added dead private methods (#1209)

### LINKED ISSUE #1209: tooling: add vulture-based pre-push check for unused private methods
Follow-up from PR #1206 (#1205 fix) retro. The original issue reporter pointed at `_lazy_serialize_context` (in `python/djust/mixins/jit.py`) as the bug location. The function literally contained the buggy `str(model_instance)` fallback that matched the reported symptom — but it had **zero call sites** and was dead code.

This sent the initial investigation down the wrong path (~10 min of context-window waste) before a reproducer-first reset surfaced the actual bug path elsewhere.

## Proposed solution

Add a pre-push check that runs [vulture](https://github.com/jendrikseipp/vulture) (or equivalent) to flag private methods (`_methodname`) with zero call sites in the codebase. PR #1206 removed `_lazy_serialize_context` after detection — but a tool would have caught it months earlier when it first became dead code, preventing the misdirection entirely.

## Implementation sketch

1. Add `vulture` to dev dependencies (already in tree? check `pyproject.toml`).
2. Create a `scripts/check-dead-private-methods.py` wrapper that:
   - Runs `vulture python/djust/ --min-confidence 90` (high confidence to avoid false positives on framework hooks).
   - Filters output to `_method_name` (private) declarations only.
   - Whitelists known framework hooks called via reflection (`_meta`, `_state`, `__init__`, etc.).
3. Wire into `.pre-commit-config.yaml` as a pre-push hook (not pre-commit — vulture is slower than ruff).

## Caveats

- **False positives**: Some private methods are called via reflection, decorators, or `getattr(self, '_foo')`. Whitelist those explicitly.
- **Mixin pattern**: `_lazy_serialize_context` was defined on `JITMixin`. Vulture should detect the unused method even within a mixin (treats class methods as defined names; checks for any caller anywhere in the codebase).
- **Existing dead methods**: A first run will likely surface several pre-existing dead methods. Triage as IMPROVEMENT items in the first PR; don't block on cleaning every one at once.

## Alternative: simpler grep-based check

If vulture is overkill, a simple `scripts/check-no-dead-private-methods.sh`:
```bash
#!/usr/bin/env bash
# Find private methods (\b_methodname\b) defined in python/djust/ that are never called.
DEFINED=$(grep -rEn 'def (_[a-z][a-z_]+)\(' python/djust/ --include='*.py' | sort -u)
# For each, grep for callers (excluding the definition line itself)
# ... pipe through sort -u and report ...
```

Crude but effective for the specific failure mode.

## Estimated effort

~30 minutes for the simple grep version (and a CI run to surface pre-existing dead methods to triage). ~2 hours for the vulture wrapper with whitelist tuning.

🤖 Auto-filed from PR #1206 retro / IDEA #2.

johnrtipton:
## Summary

Closes #1209. v0.9.5 retro Action Tracker #197.

Adds a `pre-push` hook that flags newly-added private methods with zero callers anywhere in `python/djust/` or `tests/`. Closes the failure mode that produced the dead-code `_lazy_serialize_context` in PR #1206 / #1205.

## What changed

- **`scripts/check-no-dead-private-methods.py`** — pure-Python check (no new dependency). Scans current branch's diff vs `origin/main` for newly-added `def _name(self, ...):` private methods. Greps codebase for callers via two patterns: `.METHOD(` and string-literal `"METHOD"` (for reflection like `getattr(self, '_method')`). Fails if zero callers found.
- **`.pre-commit-config.yaml`** — wires the check as a `pre-push` hook (not `pre-commit`; runs with the slow validation gates alongside pytest/cargo-test).

## Design choices

**Diff-only scope (not full-codebase)**: Only methods *added or modified in the current branch* are checked. Pre-existing dead methods are NOT flagged. This avoids forcing a 50-method cleanup PR before any other PR can ship. Pre-existing dead methods can be cleaned up in dedicated chore PRs.

**Pure-Python (not vulture)**: Issue #1209 explicitly approved either path. Pure-Python avoids adding a new dev dependency, gives full control over the heuristic, and integrates cleanly with the existing `scripts/` directory pattern.

**Best-effort fallback**: If `origin/main` isn't fetched, the check exits 0. Pre-push is best-effort; CI is the authoritative gate.

## Skipped patterns

- Dunder methods (`__init__`, `__call__`, etc.) — implicit framework contracts
- `tests/` and `__pycache__/` paths
- Methods annotated `# noqa: dead-method-allowed` — escape hatch for reflection-called methods (dispatcher lookups, etc.)

## Smoke test

Verified by injecting `_smoke_test_dead_method_no_callers` into `mixins/jit.py` and confirming the script exits 1 with the offender named in stderr. Reverted before commit.

## Test plan

- [x] Script smoke-tested against synthetic dead method
- [x] Script exits 0 on this branch (no newly-added dead methods of its own)
- [x] YAML config validates
- [x] Pure-Python — no new dependencies in pyproject.toml
- [x] Linted with ruff (passed)

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -160,3 +160,17 @@ repos:
         files: ^docs/.*\.md$
         pass_filenames: false
         stages: [pre-push]
+
+      # Flag newly-added private methods with zero callers anywhere (#1209).
+      # Prevents the dead-code-misleads-investigator failure mode that
+      # produced `_lazy_serialize_context` (PR #1206 / #1205). Pre-existing
+      # dead methods are intentionally NOT flagged — only methods added
+      # in the current branch (vs origin/main). Escape hatch: annotate
+      # the def line with `# noqa: dead-method-allowed`.
+      - id: check-no-dead-private-methods
+        name: check no dead private methods
+        entry: bash -c '.venv/bin/python scripts/check-no-dead-private-methods.py'
+        language: system
+        types: [python]
+        pass_filenames: false
+        stages: [pre-push]
--- a/scripts/check-no-dead-private-methods.py
+++ b/scripts/check-no-dead-private-methods.py
@@ -1,0 +1,191 @@
+#!/usr/bin/env python3
+"""Check for newly-introduced unused private methods in python/djust/.
+
+Run as a pre-push hook. Scans the current branch's diff (vs ``origin/main``)
+for newly-added ``def _name(self, ...):`` private method definitions on
+classes; for each, greps the entire ``python/djust/`` and ``tests/`` trees
+for callers. Fails if any newly-added private method has zero callers
+anywhere — that's the failure mode that produced the dead-code
+``_lazy_serialize_context`` cited in PR #1206 / #1205 (closed by PR #1216
+v0.9.5 retro Action Tracker #197 / GitHub #1209).
+
+Pre-existing dead methods are intentionally NOT flagged — only methods
+added (or modified) in the current branch. This avoids forcing a 50-method
+cleanup PR before any other PR can ship.
+
+Skips:
+
+- Dunder methods (``__init__``, ``__call__``, etc.) — implicit framework
+  contracts; never directly called by name.
+- Methods inside ``tests/`` and ``__pycache__/`` directories — test fixtures
+  and bytecode.
+- Methods whose ``def`` line carries ``# noqa: dead-method-allowed`` —
+  explicit escape hatch for reflection-called methods.
+
+Exit codes:
+
+- ``0``: no newly-added dead private methods (pass).
+- ``1``: at least one newly-added private method has no callers (fail with
+  a report listing offenders).
+"""
+
+from __future__ import annotations
+
+import re
+import subprocess
+import sys
+from pathlib import Path
+
+PROJECT_ROOT = Path(__file__).parent.parent.resolve()
+PRIVATE_METHOD_DEF_RE = re.compile(
+    r"^\s*def\s+(_[a-z][a-z0-9_]*)\s*\(\s*self",
+    re.MULTILINE,
+)
+DUNDER_RE = re.compile(r"^__\w+__$")
+NOQA_RE = re.compile(r"#\s*noqa:\s*dead-method-allowed", re.IGNORECASE)
+
+
+def _git_output(args: list[str]) -> str:
+    """Run a git command from PROJECT_ROOT, return stdout (empty on failure)."""
+    result = subprocess.run(
+        ["git", *args],
+        cwd=PROJECT_ROOT,
+        capture_output=True,
+        text=True,
+        check=False,
+    )
+    if result.returncode != 0:
+        return ""
+    return result.stdout
+
+
+def changed_python_files() -> list[Path]:
+    """Files in ``python/djust/`` modified in the current branch vs origin/main."""
+    output = _git_output(["diff", "--name-only", "origin/main...HEAD", "--", "python/djust/"])
+    files: list[Path] = []
+    for line in output.strip().splitlines():
+        if not line.endswith(".py"):
+            continue
+        if "tests/" in line or "__pycache__" in line:
+            continue
+        path = PROJECT_ROOT / line
+        if path.exists():
+            files.append(path)
+    return files
+
+
+def newly_added_methods(file_path: Path) -> list[tuple[str, str]]:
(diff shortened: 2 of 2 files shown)


johnrtipton (commented):
Stage 11 review (cap 250w).

Tested locally: script runs clean on main (exit 0), correctly detects synthetic dead method injected into `mixins/jit.py` (exit 1, helpful stderr listing offender + 4 remediation options at lines 180-185), and `# noqa: dead-method-allowed` escape hatch works (exit 0 after annotation). Pytest smoke check 15/15 passing.

**Strengths:**
- Diff-only scope (line 64) is the right call — avoids forcing 50-method cleanup PR.
- Best-effort `origin/main` fallback (lines 152-153) means missing fetch never blocks push.
- Same-file definition skip (line 142) correctly prevents the def line itself counting as a caller.
- String-literal pattern `['"]METHOD['"]` (line 119) catches `getattr(self, '_method')` reflection.
- Hook positioned correctly in `pre-push` block (line 176), last entry, alongside pytest/cargo-test.

**Edge case worth noting (non-blocking):**
Regex at line 41 (`def\s+(_[a-z]...)\s*\(\s*self`) requires `self` on the same line as `def`. Multi-line signatures like `_render_input(\n    self,\n    ...)` (e.g. `frameworks.py:221`, `performance.py:405`, plus 3 others I grepped) would slip past the check. Probably fine for now — most newly-added private methods fit one line — but if a future dead `_render_*` method ships, this is why. A follow-up could broaden to allow `\s*$` after `(` or use `ast.parse` on the diff hunks.

**Question for author:** intentional that base64-encoded method names or methods called via `operator.methodcaller("_name")` aren't caught? Both are exotic enough that the current heuristic seems appropriately scoped — just confirming.

APPROVE

johnrtipton:
## Retrospective — PR #1220 (#1209 dead-private-method pre-push check)

### Quality: 5/5

Closes the failure mode that produced `_lazy_serialize_context` in PR #1206. Pure-Python implementation (no new dependency); diff-only scope avoids forcing pre-existing dead-code cleanup; smoke-tested both ways (catches synthetic dead method, accepts `# noqa: dead-method-allowed` escape hatch). CI green across 13 checks. Reviewer surfaced one real edge case (multi-line `def` signatures) that's worth a follow-up but doesn't block.

### What went well

- **Pure-Python over vulture**: avoided adding a dev dependency. The script is ~150 LOC, owns its heuristic, and integrates with existing `scripts/` patterns.
- **Diff-only scope decision**: the script ONLY checks methods added/modified in the current branch vs origin/main. Pre-existing dead methods aren't flagged. Avoids the "have to clean up 50 dead methods before this can ship" trap that would have killed the PR.
- **Best-effort fallback**: if `origin/main` isn't fetched, exit 0. Pre-push is best-effort; CI is authoritative. No false-blocking on local-environment problems.
- **Smoke test discipline**: built a one-shot scratch script that injected a synthetic dead method, ran the check, asserted exit 1, then reverted. Validated the script works end-to-end before commit. Discarded the scratch script before staging — kept the diff clean.
- **Reviewer bridge worked**: Stage 11 reviewer caught a real edge case (regex requires `self` on same line as `def`; multi-line signatures slip past). Filed mentally as a follow-up; not blocking for the v1 ship.

### Lessons

- **Dead-code lints with diff-only scope are ship-able TODAY; full-codebase lints get bogged in cleanup**. The compromise is the right call for any "this would have caught a known bug" tooling addition.
- **Multi-line `def` signature is a real coverage gap**. Future v2 of the script could use `ast` parsing instead of regex. Worth filing as follow-up #TBD if any dead multi-line method ever ships.

### Process notes

- Single-commit shape (chore: prefix, no CHANGELOG required).
- Smoke-test scratch script removed before commit (kept diff focused).
- Reviewer applied own bugfix-budget cap (244 of 250 words) per #1219 discipline — meta-loop closes again.

### Final v0.9.5 milestone close

This is the **5th and final v0.9.5 PR** (after #1216 docs, #1217 idempotency test, #1218 reproducer-required, #1219 reviewer-budget). All 5 issues from the v0.9.5 retro shipped clean. Wall-clock from `/pipeline-drain` invocation to final merge: ~45 minutes for 5 PRs end-to-end with CI on each. The v0.9.5 milestone is now complete; remaining open issues (#1207 list[Model] shapes, #1212 retro-gate audit, #1214 CodeQL sanitizer model, #1215 .pxd line endings) deferred to a later milestone per ROADMAP plan.

Issue #195, opened 2026-02-01
Debug panel Handlers tab crashes: TypeError this.handlers.map is not a function


johnrtipton:
## Bug Description

Clicking the Handlers tab in the debug panel throws a JavaScript error:

```
TypeError: this.handlers.map is not a function
    at DjustDebugPanel.renderHandlersTab (debug-panel.js:2296)
```

## Root Cause

The server (`get_debug_info()` in `post_processing.py`) returns `handlers` as a **dict** (object keyed by handler name):

```json
{
  "increment": {"name": "increment", "params": [], ...},
  "decrement": {"name": "decrement", "params": [], ...}
}
```

But `debug-panel.js` line 2296 calls `this.handlers.map(...)`, which only works on arrays.

The guard at line 2290 (`!this.handlers || this.handlers.length === 0`) does not catch this because:
- `!this.handlers` is false (the object is truthy)
- `.length` is `undefined` on plain objects, which is falsy, so `=== 0` fails

## Steps to Reproduce

1. Open debug panel on a page with LiveView components (e.g., /examples/)
2. Click the "Handlers" tab
3. JS error thrown, content area shows stale content

## Fix

Either convert the dict to an array in `renderHandlersTab`:
```javascript
const handlersList = Array.isArray(this.handlers) ? this.handlers : Object.values(this.handlers);
```

Or change `get_debug_info()` to return a list instead of a dict.

## Files

- `python/djust/static/djust/debug-panel.js` (line 2296)
- `python/djust/mixins/post_processing.py` (`get_debug_info()`, lines 38-51)

PR #1223, opened 2026-04-30
fix: heterogeneous and nested list[Model] shapes in change-detection normalize pass (#1207)

### LINKED ISSUE #1207: tech-debt: support heterogeneous and nested list[Model] shapes in change-detection normalize pass

johnrtipton:
## Summary

Closes #1207. v0.9.5 retro Action Tracker #195 / v0.9.1-6 drain bucket.

Follow-up to PR #1206 (#1205 fix). The defensive normalize pass added to `_sync_state_to_rust` correctly handled the **homogeneous** `list[Model]` shape but two related shapes were NOT covered:

1. **Heterogeneous `[dict, Model]` (Model NOT first)**: `is_model_list` only checks `value[0]`, so `[{"label": "header"}, <Model>]` fell through unnormalized. The Model in position 1 escaped change-detection.
2. **Nested `list[list[Model]]`**: original normalize iterated only top-level keys; nested grouped lists failed `is_model_list` because `value[0]` was a list, not a Model.

## Fix

Refactored the inline normalize loop into a recursive module-level helper `_normalize_db_values`:

- `Model` → dict
- `QuerySet` → list[dict]
- `list[Model]` → list[dict] (any-position-Model; heterogeneous-safe)
- `list[list[...]]` → recursive normalize with depth bound (`_NORMALIZE_DEPTH_LIMIT = 3`)

Bounded recursion handles ``list[list[list[Model]]]`` shapes; deeper nesting is best-effort (returns unchanged).

## Tests

Two regression cases added (both fail without fix, pass with):
- `test_heterogeneous_list_dict_first_propagates_field_mutations`
- `test_nested_list_of_lists_of_models_propagates_field_mutations`

All 9 cases in `test_list_model_diff_1205.py` green. Full unit suite: 4433 passed, 6 skipped.

## Test plan

- [x] Both new tests fail on origin/main (red phase confirmed)
- [x] Both new tests pass on this branch (green phase)
- [x] All 7 existing tests in the file still pass (no regression)
- [x] Full unit suite green
- [x] Idempotency check passes (no double-processing of already-serialized values)
- [x] CHANGELOG entry under `### Fixed`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,25 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- **Heterogeneous and nested `list[Model]` shapes now propagate field
+  mutations through change-detection (#1207).** Follow-up to PR #1206
+  (#1205 fix). The defensive normalize pass introduced in v0.9.5 covered
+  the homogeneous `list[Model]` shape but two related shapes were NOT
+  covered: heterogeneous `[dict, Model]` (Model not first — `is_model_list`
+  checks only `value[0]`) and nested `list[list[Model]]` (e.g., grouped
+  tasks — the outer list contains lists, not Models). Both reproduced the
+  #1205 bug for those shapes (in-place field mutations escaped
+  change-detection because `Model.__eq__` is pk-only). Fix: refactored the
+  inline normalize loop into a recursive `_normalize_db_values` helper
+  that scans the full list for any-position Model (heterogeneous-safe) and
+  recurses into nested lists with a bounded depth (`_NORMALIZE_DEPTH_LIMIT
+  = 3`). Idempotent on already-serialized values. Two regression cases in
+  `tests/unit/test_list_model_diff_1205.py` lock the new shape coverage in
+  (`test_heterogeneous_list_dict_first_propagates_field_mutations`,
+  `test_nested_list_of_lists_of_models_propagates_field_mutations`).
+
 ### Added
 
 - **`LiveViewTestClient.render_with_patches()` — VDOM-diff accessor for tests
--- a/python/djust/mixins/rust_bridge.py
+++ b/python/djust/mixins/rust_bridge.py
@@ -9,7 +9,7 @@ from urllib.parse import parse_qs, urlencode
 
 from ..security import sanitize_for_log
 from ..serialization import normalize_django_value
-from ..utils import get_template_dirs, is_model_list
+from ..utils import get_template_dirs
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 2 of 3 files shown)


johnrtipton (commented):
## Review: APPROVE (cannot self-approve, posting as comment)

Clean refactor of the inline normalize loop into a reusable, recursive helper. All 9 tests in `test_list_model_diff_1205.py` pass locally.

### Verified
- `_normalize_db_values` (rust_bridge.py:63-116): bounded recursion (`_NORMALIZE_DEPTH_LIMIT = 3`), lazy import of `django.db.models` (consistent with project pattern), idempotent on already-serialized list[dict] (`test_normalize_idempotent_on_already_serialized` covers this).
- Call site (rust_bridge.py:391-413) is materially cleaner — removed the `Model`/`QuerySet` imports inside `_sync_state_to_rust` and the 3-branch isinstance ladder.
- `is_model_list` import correctly removed from line 12; still used in `mixins/context.py` and `mixins/request.py` (verified — not breaking other consumers).
- Heterogeneous-safe `any(isinstance(item, Model) for item in value)` short-circuits on first Model — for typical contexts (~10 keys, lists <100 items) the perf delta vs old O(1) `value[0]` check is negligible.

- `tuple[Model, ...]` falls through unchanged. Consistent with `is_model_list` semantics elsewhere — fine to defer.
- `[{"user": <Model>}]` (dict-with-Model nested in list) not covered here, but this gap predates #1207 and `_deep_serialize_dict` handles the non-defensive path. Worth tracking as follow-up but out of scope.
- `_NORMALIZE_DEPTH_LIMIT = 3` is well-justified in the docstring; ample for realistic shapes.

CHANGELOG entry under `### Fixed` is detailed and references both regression tests by name. Two regression cases lock both fixed shapes in.

**Decision: APPROVE**

johnrtipton:
## Retrospective — PR #1223 (#1207 list[Model] heterogeneous + nested shapes)

### Quality: 5/5

Closes the framework-correctness gap from PR #1206's review. TDD discipline held: 2 failing reproducer tests written first (red phase confirmed), then the recursive `_normalize_db_values` helper turned them green. Two-commit shape (impl+tests / CHANGELOG) per Action Tracker #181. CI green across 13 checks. Reviewer surfaced 3 non-blocking edge-case observations within 250-word budget.

### What went well

- **Reproducer-first discipline (the new Stage 4 mandatory item from #1218 / PR #1218!)** caught the heterogeneous and nested shapes BEFORE the implementation. The new pipeline-template gate would have enforced this even without my discretion. The discipline is meta-loop validating its own existence.
- **Recursive helper extraction** turned the inline 3-branch isinstance ladder into a single reusable function with depth bound. Cleaner call site at `_sync_state_to_rust` (no longer needs Model/QuerySet imports). Easier to test in isolation if a v2 ever needs to.
- **Idempotency invariant preserved**: the strengthened `test_normalize_idempotent_on_already_serialized` from PR #1217 still passes — the new helper doesn't accidentally re-process already-serialized list[dict] values. The zero-patch assertion #1208 added would have caught any regression here.
- **Cleanup pass**: removed `is_model_list` import (no longer used in rust_bridge.py; still used in 5 other sites in context.py + request.py) and the lazy `Model, QuerySet` import inside `_sync_state_to_rust` (since the helper does its own).

### Lessons

- **The depth bound is a knob worth understanding**: `_NORMALIZE_DEPTH_LIMIT = 3` covers all plausible shapes (`list[list[list[Model]]]` is depth 2-3 nested; deeper hits `_deep_serialize_dict`). If a real user shape ever exceeds 3, the symptom is silent — the value flows through unchanged. A future version could log a warning at depth limit if observed in the wild.
- **Edge cases the reviewer flagged are documented gaps, not regressions**: tuples (not lists), dict-with-Model-values inside a list, are pre-existing non-coverage that #1207 was scoped not to address. They could be follow-ups if real user reports surface.

### Process notes

- Two-commit shape: `da219668` (impl + tests) + `79be5ca3` (CHANGELOG) — clean.
- Reviewer applied 250-word budget per #1219 discipline.
- Pre-commit hooks all passed first try.

PR #1204, opened 2026-04-29
feat(rich-select): variant support, trigger tinting, onclick parity

johnrtipton:
## Summary

`RichSelect` previously supported per-option icons, images, descriptions, and badges — but no way to color an option row or have the trigger reflect the selected option's category. It also silently omitted the open/close onclick handlers that the `{% rich_select %}` template tag emits, forcing programmatic consumers to monkey-patch rendered HTML.

This PR closes both gaps.

### Per-option `variant` support

Each option dict accepts an optional `variant` key that tints the row in the dropdown **and** the trigger when that option is currently selected:

```python
RichSelect(
    name="assignee",
    options=[
        {"value": "alice", "label": "Alice", "variant": "success"},
        {"value": "bob", "label": "Bob", "variant": "danger"},
    ],
)
```

Built-in variants align with `Badge`/`Button`/`Tag`/`Alert` vocabulary:
`info`, `success`, `warning`, `danger`, `muted`, `primary`, `secondary`.

Disabled pickers suppress the trigger variant so the greyed-out state stays visually dominant.

### `variant_map` convenience kwarg

Mirrors `Badge.status()` — useful when the same enum value always needs the same color:

```python
RichSelect(
    name="status",
    options=[{"value": "NEW", "label": "New"}, ...],
    variant_map={"NEW": "info", "DONE": "success", "BLOCKED": "danger"},
)
```

Per-option `variant` wins over `variant_map` when both are set.

### Permissive variant validation

Names must match `^[a-z0-9][a-z0-9-]{0,31}$`, so downstream projects can add custom variants by shipping a matching `.rich-select-option--variant-<name>` CSS rule. The built-in set is the baseline, not a closed allow-list.

Malformed names (quotes, spaces, HTML escapes, uppercase) fall back to `"default"` to prevent class-attribute injection.

### Trigger interaction parity

The programmatic class now emits `onclick` + `onkeydown` (Enter, Space) matching the template tag, plus `onclick` on each option to close the dropdown after selection. A shared `_rich_select_resolve_variant` helper keeps both entry points in sync.

## Why bundled

Splitting variants and onclick parity into two PRs was tempting, but downstream consumers who try variants without parity get a picker that's visually correct and non-interactive — worse than the status quo. Shipping them together guarantees the component works as its docstring claims out of the box.

## Tests

18 new unit tests in `tests/unit/test_rich_select.py` cover:
- Baseline rendering (empty options, selected value, disabled state)
- Per-option variants + opt-out
[...]

PR #1201, opened 2026-04-28
security: code-scanning cleanup batch (4 fixes + 15 FP dismissals)

johnrtipton:
## Summary

Closes **19 open CodeQL / Dependabot alerts** in one batch:
- **4 real fixes** (1 medium-severity security, 1 dependency CVE, 2 lint/correctness)
- **15 documented false-positive dismissals** (already-correct code that the analyzers don't model)

## Real fixes

| Alert | Issue | Fix |
|---|---|---|
| CodeQL #2195 | Client-side URL redirect (`client.js:1053`) | Same-origin path validation in `src/03-websocket.js:519` — rejects protocol-relative, off-origin, `javascript:`/`data:` schemes |
| Dependabot #90 | postcss XSS via unescaped `</style>` in stringify | `npm update postcss` → 8.5.9 → 8.5.10 |
| CodeQL #2194 | Empty `except AttributeError: pass` (`sticky.py:210`) | Logs at DEBUG with comment explaining expected case (read-only proxy children) |
| CodeQL #2267 | Duplicate `import asyncio` (`request.py:322`) | Removed — module already imports at top |

## Dismissals (15 false positives)

All dismissals applied via `gh api` with detailed explanatory comments. Categories:

- **8x py/log-injection** (#2254, #2253, #2239, #2238, #2237, #2236, #2235, #2183): log calls already pass user-controlled input through `sanitize_for_log()` (`security/log_sanitizer.py:42`). CodeQL doesn't model the sanitizer.
- **2x py/cyclic-import** (#2231, #2230): intentional lazy late-imports inside function bodies that break module-level circular dependencies.
- **1x py/not-named-self** (#2268): `as_view` is `@classonlymethod`; `cls` is correct.
- **2x py/unused-global-variable** (#2272, #2175): both referenced multiple times (verified via grep).
- **1x py/catch-base-exception** (#2273): one-shot CI diagnostic script that must catch SystemExit subclasses; `noqa: BLE001` documents it.
- **1x js/useless-assignment** (#2174): `client.js` is a build artifact concatenated from 52 source modules; bundle-level identifiers are tree-shake remnants.

## Test plan

- [x] `make test` — 4047 Python + 1486 JS, all green
- [x] Bundle rebuilt via `scripts/build-client.sh`
- [x] `npm ls postcss` shows 8.5.10
- [x] All 15 dismissals applied via `gh api` with documented comments

Closes #2195
Closes #2194
Closes #2267

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1231, opened 2026-04-30
chore: dedupe rich_select _BUILTIN_VARIANTS + add is_builtin_variant() helper

johnrtipton:
## Summary

Closes CodeQL alert #2287 (py/unused-global-variable). Pure cleanup; no behavior change.

## What changed

The constant `_BUILTIN_VARIANTS` in `python/djust/components/components/rich_select.py` was duplicated verbatim as `_RICH_SELECT_BUILTIN_VARIANTS` in `python/djust/components/templatetags/djust_components.py`. CodeQL flagged the rich_select.py copy as unused (only consumer was the test, which CodeQL ignores via `paths-ignore`); the djust_components.py copy was dead code (defined but no production reference within the file).

Two-part cleanup:

1. **Single source of truth**: `rich_select.py` is the canonical home (Component-side, where the test imports from). Switched the literal set to `frozenset` for immutability. `djust_components.py` now imports `_BUILTIN_VARIANTS` and `_VARIANT_NAME_RE` from `rich_select.py` rather than redefining.

2. **Real consumer**: added `is_builtin_variant(name: str) -> bool` helper in `rich_select.py`. Gives the constant a non-test consumer (closes the CodeQL "unused" complaint structurally), and exposes a clean API for downstream tooling that wants to distinguish "djust ships this CSS" from "user shipped their own CSS." Both shapes remain valid; `_VARIANT_NAME_RE` is still the runtime gatekeeper.

The duplication itself was an oversight from PR #1204 (rich-select variants), where both files were touched in the same PR and the shared constant got copied instead of imported.

## Test plan

- [x] `tests/unit/test_rich_select.py`: 18/18 still passing
- [x] Full unit suite: 4433 passed, 6 skipped, 0 regressions
- [x] Re-export wiring verified (`_BUILTIN_VARIANTS is _RICH_SELECT_BUILTIN_VARIANTS` — same object)
- [x] New `is_builtin_variant()` helper smoke-tested

### DIFF
--- a/python/djust/components/components/rich_select.py
+++ b/python/djust/components/components/rich_select.py
@@ -10,22 +10,41 @@ from djust import Component
 # Built-in variants for which djust ships CSS out of the box. Matches the
 # Badge / Button / Tag / Alert signal set, plus the theme-standard
 # `primary` / `secondary` for projects with more than 5 categories.
-_BUILTIN_VARIANTS = {
-    "default",
-    "info",
-    "success",
-    "warning",
-    "danger",
-    "muted",
-    "primary",
-    "secondary",
-}
+#
+# This set is the canonical source of truth across the Component API
+# (``rich_select.py``) and the templatetag (``djust_components.py``) —
+# the templatetag imports from here rather than duplicating.
+_BUILTIN_VARIANTS = frozenset(
+    {
+        "default",
+        "info",
+        "success",
+        "warning",
+        "danger",
+        "muted",
+        "primary",
+        "secondary",
+    }
+)
 
 # Any variant name consumers pass through must match this pattern so it can
 # be safely interpolated into a CSS class attribute. Downstream projects can
 # define their own variants (e.g. `indigo`, `accent-2`) by shipping a
 # matching `.rich-select-option--variant-<name>` rule in their own CSS.
 _VARIANT_NAME_RE = re.compile(r"^[a-z0-9][a-z0-9-]{0,31}$")
+
+
+def is_builtin_variant(name: str) -> bool:
+    """Return ``True`` if ``name`` is one of the variants djust ships CSS for.
+
+    Public helper for downstream consumers (Components, template tags,
+    docs generators) that want to distinguish "djust ships this CSS"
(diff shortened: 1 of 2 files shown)


johnrtipton:
## Retrospective — PR #1231 (rich_select _BUILTIN_VARIANTS dedup)

### Quality: 5/5

Pure cleanup PR closing CodeQL alert #2287. Zero behavior change; just import-not-duplicate. Investigation found BOTH copies of the constant were dead from a production-code perspective (rich_select.py:_BUILTIN_VARIANTS used only by tests; djust_components.py:_RICH_SELECT_BUILTIN_VARIANTS not used anywhere). Cleanest fix gave the constant a real consumer via a new `is_builtin_variant()` helper.

### What went well

- **Investigation found the real shape**: not just "one copy is unused" but "both copies are dead." That informed the fix — adding `is_builtin_variant()` to give the constant a non-test consumer is the structural fix; just deleting one copy would have been a half-fix.
- **`frozenset` upgrade**: changed the literal `set` to `frozenset` to communicate immutability (the variant set is a contract, not a mutable collection).
- **Re-export verified by identity check**: `_BUILTIN_VARIANTS is _RICH_SELECT_BUILTIN_VARIANTS` confirms the import is the same object, not a copy. If a future bug accidentally redefines one of them, this assertion would catch it.

### Lessons

- **PR #1204's two-file copy was the original sin.** Both files got the same constant in the same PR; should have been a shared import from the start. Adding a Stage 7 self-review bullet "any constant defined identically in two files SHOULD be a shared import, not a duplicate" might catch this class. Filed mentally as a v0.9.2-1 process-canon idea (not blocking).

### Process notes

- Single-commit shape (chore: prefix; pure refactor).
- Code-scanning alert #2287 will auto-close on the next CodeQL scheduled scan (the new `is_builtin_variant()` helper is the real consumer).
- Code-scanning alert #2281 (URL redirection FP) was dismissed via `gh api` earlier in this session per the PR #1201 dismissal pattern (15 dismissed FPs there, this is the 16th).

Issue #1276, opened 2026-05-01
@action: docstring promises inline save_note.error rendering on exception, but re-raise triggers dev overlay instead


johnrtipton:
## Summary

`@action`'s docstring promises that on exception, `self._action_state[<method>]` is set to `{pending: False, error: "<msg>", result: None}` and the template can read `{% if save_note.error %}{{ save_note.error }}{% endif %}` to display the error inline. In practice, the framework re-raises the exception after recording state, the WS dispatcher catches it and sends a `type:"error"` frame, and the dev-mode error overlay pops up instead of the template re-rendering with `save_note.error` populated.

## Reproduction

```python
from djust import action

class NoteView(LiveView):
    @action
    def save_note(self, text: str = "", **kwargs):
        if not text:
            raise ValueError("Please enter some text")
        ...
```

```html
<form dj-submit="save_note">
    <input name="text">
    <button type="submit">Save</button>
</form>

{% if save_note.error %}
    <div class="error">Error: {{ save_note.error }}</div>
{% endif %}
```

Expected (per docstring): submit empty form → red error div appears, page stays interactive, user fixes the input and resubmits.

Actual: submit empty form → big red dev-mode error overlay covers the whole page with the traceback. The error-div branch never renders. User has to dismiss the overlay (Esc) to interact again.

## Root cause

`python/djust/decorators.py` (`def action`) records the error in `_action_state` but then **re-raises**:

```python
} except BaseException as exc:
    self._action_state[action_name] = {
        "pending": False,
        "error": str(exc),
        "result": None,
    }
    raise  # ← re-raised here
```

The WS dispatcher catches the re-raise and sends:

```json
{"type": "error", "error": "Error in NoteView.save_note(): ValueError: ..."}
```

So the next render is never triggered with the updated `_action_state`.

## Suggested fix

Two reasonable options:

1. **Don't re-raise from `@action`'s wrapper** — the error is already captured in state; surfacing it as a WS error frame is redundant and breaks the documented contract. Just `return` after recording.
2. **Add a `surface_inline=True` parameter to `@action`** — opt-in to the no-re-raise behavior, keep current behavior as default for back-compat.

Option 1 matches the docstring. Option 2 is safer but more API surface.

Either way, the docstring's "On exception (re-raised after recording)" clause should be reconciled with the WS dispatcher's behavior — currently they're contradictory because the dispatcher's error frame short-circuits the re-render that would surface the recorded state.

## Workaround in user code

[...]

Issue #1275, opened 2026-05-01
data_table tag emits 'table_sort' events but DataTableMixin registers 'on_table_sort' — every interaction returns 'No handler found'


johnrtipton:
## Summary

`{% data_table %}` template tag emits `dj-click="table_sort"` (and analogous `table_filter`, `table_page`, `table_search`, `table_select`, ...) on its UI elements, but `DataTableMixin` registers handlers named `on_table_sort`, `on_table_filter`, etc. The mixin's handlers don't match the tag's wire event names — every interaction returns `No handler found for event: table_sort`.

## Reproduction

```python
from djust.components.mixins.data_table import DataTableMixin

class OrderListView(DataTableMixin, LiveView):
    table_model = Order
    table_columns = [{"key": "id", "label": "ID", "sortable": True}]
    table_page_size = 5

    def mount(self, request, **kwargs):
        super().mount(request, **kwargs)
        self.init_table_state()
        self.refresh_table()
```

```html
{% data_table rows=rows columns=columns sort_by=sort_by sort_desc=sort_desc page=page total_pages=total_pages search=True paginate=True %}
```

Click any column header → server returns:
```json
{"type": "error", "error": "No handler found for event: table_sort\n  Did you mean: on_table_sort, on_table_import, on_table_export?\n  Available handlers on OrderListView: on_table_bulk_action, on_table_cell_edit, ..."}
```

Sort, filter, paginate, search, select — every interaction is dead because the tag and the mixin agree on names but disagree on prefix.

## Root cause

The `{% data_table %}` tag's defaults at `python/djust/components/templatetags/djust_components.py:504+` use bare event names:

```python
def data_table(
    ...,
    sort_event="table_sort",
    page_event="table_page",
    next_event="table_next",
    select_event="table_select",
    search_event="table_search",
    filter_event="table_filter",
    edit_event="table_cell_edit",
    reorder_event="table_reorder",
    ...,
):
```

The mixin at `python/djust/components/mixins/data_table.py` registers `on_table_sort`, `on_table_filter`, etc. (per Phoenix-LiveView-style convention).

The tag emits `dj-click="table_sort"`; the WS dispatcher looks for a handler named `table_sort`; finds none.

## Suggested fix

Two reasonable options:

1. **Change the tag defaults to `on_table_*`** — matches the mixin out of the box.
2. **Change the mixin to register `table_sort` instead of `on_table_sort`** — matches the tag.
[...]

Issue #1259, opened 2026-05-01
tech-debt: document audit-as-pre-staged-work-graph recipe in pipeline-drain skill


johnrtipton:
Source: v0.9.2-3 retro / PRs #1257 + #1258.

## Context

The v0.9.2-3 milestone ran a high-leverage shape: an audit doc PR (#1257) was the entry point, not the implementation PR. The audit produced:

1. 14-bug history of the VDOM subsystem (from RETRO.md + git log + CHANGELOG)
2. 10 ranked current-code weaknesses (correctness / perf / arch / test / docs / security)
3. **5 GitHub issues filed BEFORE the audit-doc PR opened** (so the milestone entry could link real numbers, no TBD backfill)
4. ROADMAP milestone block created in the audit-doc PR
5. Drain via single grouped PR
6. Single retro covering both PRs

Wall-clock from audit-merge (PR #1257) to fix-merge (PR #1258): ~75 minutes. Higher leverage than running 5 separate small PRs because:

- Issues are pre-staged and contextualized in the audit doc.
- Drain script (`/pipeline-drain --milestone X --group --all`) batches them.
- Stage 4 VERIFY LITERAL API CONTRACTS catches audit drift on first-use.
- Stage 7 / Stage 8 reviews can spot-check audit claims rather than re-deriving.

## Proposed change

Add a new section to `~/.claude/skills/pipeline-drain/SKILL.md` (and the upstream pipeline-skill repo at `/Users/tip/Dropbox/online_projects/ai/pipeline-skill/skills/pipeline-drain/SKILL.md`) titled **"Audit-driven drain: pre-staged work-graph recipe"**. Document the 4-step shape:

1. **File issues first** (`gh issue create --label tech-debt` × N). The audit-author should file before opening the audit-doc PR so the milestone entry can link real numbers.
2. **Open the milestone in the audit-doc PR**. Single docs PR that adds `docs/X-AUDIT-YYYY-MM-DD.md` + ROADMAP entry + CHANGELOG note.
3. **Drain via grouped PR**. `/pipeline-drain --milestone vX.Y.Z-N --group --all` runs the implementer subagent on all issues at once.
4. **Single retro covers both PRs**. The audit + drain are one coherent unit.

Variables that influence whether this shape applies:

- N (number of issues) — sweet spot 3-7. Below 3, the audit overhead dominates; above 7, the drain PR becomes too big to review.
- Issue cohesion — works best when all N issues touch the same subsystem.
[...]

Issue #1200, opened 2026-04-28
tech-debt: tautology test detection


johnrtipton:
Source: Retro v0.9.4 / PR #1190

When a test asserts "this thing happened", check whether the assertion would ALSO pass if the action under test did nothing. If yes, it's a tautology — production state from prior tests, fixtures, or module setup may be making it pass for the wrong reason.

**Failure mode**: silent. The test stays green forever even after the function under test silently breaks. Coverage metrics report it as covered.

**Observed instance**: PR #1190's `test_ready_completes_other_setup_even_when_auto_enable_skipped` asserted `any(isinstance(filters, DjustLogSanitizerFilter))`. But every prior test in the file calls `app.ready()` which adds another filter (no idempotency guard). By the time test #6 runs, the filter is already on the logger from prior calls — the assertion would pass even if test #6's own ready() did nothing. Fix: snapshot count BEFORE, assert count grew by exactly 1.

Canonicalize as a CLAUDE.md rule: for any "action happened" assertion, ask "would this pass if the action didn't run?"

johnrtipton:
Shipped in PR #1225 (v0.9.4 retro process canon batch in CLAUDE.md). Comma-list `Closes #...` form in PR body parses only the first ref — the rest had to be closed manually (this confirms #1184's diagnosis once again; the PR-checklist convention says one Closes per body line). 🤖 v0.9.1-6 drain bucket — final issue closure cleanup.

PR #1292, opened 2026-05-01
fix(lifecycle): handle_mount drains _async_tasks + _pending_push_events (#1280, #1283)

### LINKED ISSUE #1280: assign_async() called from mount() never resolves over WebSocket
# Upstream issue: `assign_async()` called from `mount()` never resolves over WebSocket

**Status:** rate-limited at write time
**Severity:** affects every demo/app that loads dashboard data with `assign_async` + `dj_suspense`

## Summary

`assign_async()` works correctly when called from event handlers, but **never
resolves** when called from `mount()`. The browser sees the initial skeleton
HTML and stays there forever — no async-driven update frame ever arrives.

This is a TWO-bug interaction:

1. **`handle_mount` doesn't drain `_async_tasks`.** The WS consumer's
   `_dispatch_async_work` is called from `handle_event` (lines 3047, 3365)
   and the deferred-activity path (lines 1218, 1287), but **not** from
   `handle_mount` (lines 1616-2354). Bug 1 alone means: tasks queued in
   `mount()` are never even attempted.

2. **`_async_tasks` callbacks are stripped to `None` before `handle_mount`
   ends.** Even after fixing bug 1 (adding `await self._dispatch_async_work()`
   at the end of `handle_mount`), every callback in `_async_tasks` reaches
   `_run_async_work` as `None`, with `args=[]` (note: a list, not the
   original tuple). The structure `[None, [], {}]` is the JSON-roundtrip
   shape of `(callback, (), {})` — function callables serialize to `null`
   and tuples to lists.

   Likely cause: `_async_tasks` (despite the underscore prefix) is being
   included in the state save/restore that happens during mount. The
   serializer can't preserve closures, and the restore reads back garbage.

## Reproduction (apps.features_v09.views_suspense.SuspenseDashboardView)

```python
class SuspenseDashboardView(BaseLiveView):
        self.assign_async("users",  self._load_users)    # 0.6s sleep
        self.assign_async("orders", self._load_orders)   # 1.0s sleep
        self.assign_async("revenue", self._load_revenue) # 1.4s sleep
```

Browser console after loading the page (capturing WS frames via Playwright):

```
[recv] connect
[sent] mount
[recv] mount {version: 2, html: ... skeleton tiles ...}
< 3 seconds pass — no further frames >
FINAL DOM: 6 skeletons, 0 .sus-value elements
```

After patching `handle_mount` to call `_dispatch_async_work()`:

```
ERROR: callback=None args=[] kwargs={} task_name='assign_async:users'
ERROR: callback=None args=[] kwargs={} task_name='assign_async:orders'
ERROR: callback=None args=[] kwargs={} task_name='assign_async:revenue'
TypeError: sync_to_async.<locals>.<lambda>() missing 1 required positional argument: 'f'
```

The `sync_to_async()` lambda error is because `callback=None` triggers
`sync_to_async`'s "no func" branch which returns a decorator-factory.

## Diagnostic evidence

The Python-only test path **passes** — `LiveViewTestClient` mount() then
`client.render_async()` resolves all three loaders correctly:

```python
def test_render_async_drains_loaders(self):
    client = LiveViewTestClient(SuspenseDashboardView).mount()
    client.render_async()
    assert client.view_instance.users.ok  # ✅ passes
```

`render_with_diff()` also produces 24 patches with the resolved values:

```python
html_pre, _, version_pre = client.render_with_patches()      # version=1
client.render_async()
html_post, patches, version_post = client.render_with_patches()  # version=2
# patches has 24 entries flipping skeleton → .sus-value
# "Active users" appears in html_post
```

So the bug is purely in the WebSocket dispatch path. The Python rendering,
diff, and AsyncResult state all work.

## Suggested fix (both bugs)

**Bug 1 — drain on mount:**

```python
# in handle_mount, after `await self.send_json(response)` (line 2353):
await self._dispatch_async_work()
```

Same call in `_mount_one` for the batch path.

**Bug 2 — exclude `_async_tasks` from state save/restore:**

The current state-save logic appears to JSON-serialize the entire view
state including `_async_tasks`. Either:

- Add `_async_tasks` (and `_async_pending`) to a serialization deny-list
  alongside other private/runtime attrs, OR
- Drain the queue *before* the state save runs, OR
- Move the queue off the view instance and onto the consumer (the queue
  is consumer-scoped, not view-scoped — surviving a state restore makes
  no sense).

The third option is cleanest — `_async_tasks` is conceptually "this
consumer's pending work for this round-trip", not "view state".

## Affected demos

- `apps.features_v09.views_suspense.SuspenseDashboardView` (assign_async × 3)
- Any app using `assign_async` from `mount()` for first-load data
- Any app using `start_async` from `mount()` for post-render side effects

The "Reload all" button in the suspense demo *does* work because it goes
through `handle_event` (which has the drain) and re-queues the loaders
fresh AFTER state restore (so the closures are intact when dispatched).

### LINKED ISSUE #1283: audit-A: `mount()` doesn't flush `_pending_push_events` queue
**Audit reference**: `docs/audits/lifecycle-2026-05.md` § Weakness #3.

## Symptom

Push events queued during `mount()` (or any `on_mount` hook) are silently lost. The client never receives them. Notifications/alerts emitted by mount-time logic miss the mount-window delivery.

## Root cause

`websocket.py:2352` (end of `handle_mount`) calls `await self.send_json(response)` with **no preceding `_flush_push_events()` call**.

Compare with the event-handler path (`websocket.py:2880, 2921, 2925`) which flushes after every render, and `_run_async_work` (`websocket.py:921`) which flushes after async completion.

## Reproducer

```python
class HelloView(LiveView):
        if request.user.is_authenticated:
            self.push_event("notify", {"msg": "Welcome back"})
            # Queued but never sent during mount
```

Mount the view; observe browser DevTools — no `notify` frame.

## Fix

One-line fix: add `await self._flush_push_events()` in `handle_mount` immediately before the final `send_json(response)`. Follows the same pattern as the existing call at `websocket.py:921`.

## Test

`tests/test_lifecycle_mount_async.py::test_mount_time_push_event_delivered` — mount a view that pushes; assert client received frame.

## Related

- Sibling lifecycle gap #1280 (`_dispatch_async_work` not called in `handle_mount`) is the same one-liner shape; bundle if landing together.

johnrtipton:
## Summary

First v0.9.2-5 drain-bucket PR. Closes #1280 (`assign_async()` /
`start_async()` from `mount()` never resolved over WS) and #1283
(`mount()` push events never delivered). Both 🔴 production blocker
class per `docs/audits/lifecycle-2026-05.md` § Weakness #1 + #3.

Both root at the same site: `LiveViewConsumer.handle_mount()` ended
with `await self.send_json(response)` and never drained `_async_tasks`
or `_pending_push_events`. Compare with `handle_event` /
`_run_async_work` / `_flush_deferred_activity_events` which already
drain both.

## What changed

`python/djust/websocket.py:2352-2364` — after the final mount frame:

```python
await self.send_json(response)
# (new)
await self._flush_push_events()
await self._dispatch_async_work()
```

Send-then-drain ordering: client must establish view before receiving
push events or async-driven patches.

## Tests

4 regression cases in `python/djust/tests/test_handle_mount_drains_queues.py`:
- `TestHandleMountSourceShape::test_calls_dispatch_async_work` — locks call into source.
- `TestHandleMountSourceShape::test_calls_flush_push_events` — locks call into source.
- `TestHandleMountSourceShape::test_drains_run_after_final_mount_frame` — asserts ordering.
- `TestHandleMountDrainBehavior::test_tail_drains_both_queues_in_order` — behavioral via AsyncMock.

Plus a defensive fix to `_FakeConsumer` test fixtures (sync → async stub).

## Commit shape

Per Action Tracker #181:
- `c40920d1` fix + tests
- `3d8ca19f` test-fixture compatibility (sync → async stub)
- `d6e0f4f6` CHANGELOG entry

## v0.9.2-5 context

Group 1 of 3 (blocks v0.9.2 stable). Group 2 = data_table fixes
(#1275/#1291/#1279). Group 3 = `@action` re-raise (#1276).

Closes #1280
Closes #1283

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,22 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- **`mount()` lifecycle: queued async work and push events are now
+  drained after the mount frame.** Closes #1280 (`assign_async()` /
+  `start_async()` called from `mount()` never resolved over WebSocket
+  — view stayed at initial loading-state HTML forever) and #1283
+  (`push_event()` called from `mount()` or `on_mount` hooks queued
+  events that never reached the client). Both root at the same site:
+  `LiveViewConsumer.handle_mount()` ended with `send_json(response)`
+  without draining `_async_tasks` or `_pending_push_events`. The fix
+  mirrors the established pattern in `handle_event()` /
+  `_flush_deferred_activity_events()`: send the response frame, then
+  drain push events, then dispatch async work. 4 regression cases in
+  `TestHandleMountSourceShape` and `TestHandleMountDrainBehavior`
+  (`python/djust/tests/test_handle_mount_drains_queues.py`).
+
 ### Documentation
 
 - **Production Deployment guide extended with Tier 1/2/3 patterns
--- a/python/djust/tests/test_handle_mount_drains_queues.py
+++ b/python/djust/tests/test_handle_mount_drains_queues.py
@@ -1,0 +1,119 @@
+"""
+Regression tests for #1280 and #1283.
+
+Both issues root at the same site: ``LiveViewConsumer.handle_mount`` ends
+with ``await self.send_json(response)`` and previously did not drain the
+``_async_tasks`` or ``_pending_push_events`` queues. The result was:
+
+- ``start_async()`` / ``assign_async()`` calls inside ``mount()`` queued
+  callbacks that never spawned, leaving the view frozen at the initial
+  loading-state HTML (#1280).
+- ``push_event()`` calls inside ``mount()`` (or ``on_mount`` hooks) queued
+  events that never reached the client (#1283).
+
+The fix adds ``_flush_push_events()`` and ``_dispatch_async_work()`` calls
+after the mount frame, mirroring the pattern in the event-handler and
+deferred-activity paths.
+"""
+
+import inspect
+from unittest.mock import AsyncMock
+
+import pytest
+
+from djust.websocket import LiveViewConsumer
+
+
+# ---------------------------------------------------------------------------
+# Structural assertions — lock in the fix at source-line level.
+# ---------------------------------------------------------------------------
+
+
+class TestHandleMountSourceShape:
+    """``handle_mount`` source must include the post-frame queue drains."""
+
+    def _source(self) -> str:
+        return inspect.getsource(LiveViewConsumer.handle_mount)
+
+    def test_calls_dispatch_async_work(self):
+        """``handle_mount`` must call ``_dispatch_async_work``. Closes #1280."""
+        assert "_dispatch_async_work()" in self._source(), (
+            "handle_mount must drain _async_tasks. Without this, "
+            "start_async()/assign_async() called from mount() are queued "
+            "but never spawned (see #1280)."
+        )
+
+    def test_calls_flush_push_events(self):
+        """``handle_mount`` must call ``_flush_push_events``. Closes #1283."""
+        assert "_flush_push_events()" in self._source(), (
+            "handle_mount must drain _pending_push_events. Without this, "
+            "push_event() called from mount() never reaches the client "
+            "(see #1283)."
+        )
+
+    def test_drains_run_after_final_mount_frame(self):
+        """Drain calls must occur AFTER the final ``send_json(response)``.
+
+        If the order is reversed, the client may receive a push-event frame
+        or an async-driven patch frame before the mount frame establishes
+        the view, leading to "patch on missing element" errors.
+        """
+        src = self._source()
+        last_send = src.rfind("send_json(response)")
+        assert last_send != -1, "could not find send_json(response) in handle_mount source"
+        flush_after = src.find("_flush_push_events()", last_send)
+        dispatch_after = src.find("_dispatch_async_work()", last_send)
+        assert flush_after > last_send, (
+            "_flush_push_events() must come AFTER the final send_json("
+            "response); otherwise push events would arrive before the "
+            "mount frame establishes the view."
+        )
+        assert dispatch_after > last_send, (
+            "_dispatch_async_work() must come AFTER the final send_json("
+            "response); otherwise async-driven patches could race the mount."
+        )
+
+
+# ---------------------------------------------------------------------------
+# Behavioral assertion — the consumer's drain methods are called once.
+# ---------------------------------------------------------------------------
+
+
+@pytest.mark.asyncio
+class TestHandleMountDrainBehavior:
+    """The drain methods must be invoked by ``handle_mount``'s tail.
+
+    Rather than driving the full ``handle_mount`` (which has ~700 lines of
+    setup including Rust state, sticky views, sessions, hooks), we
+    extract and execute the *new* tail block in isolation. That block is
+    the only thing this test cares about; the rest of ``handle_mount``'s
+    behavior is covered by other tests.
+    """
+
+    async def test_tail_drains_both_queues_in_order(self):
+        consumer = LiveViewConsumer()
+        consumer._flush_push_events = AsyncMock()
+        consumer._dispatch_async_work = AsyncMock()
+        consumer.send_json = AsyncMock()
+
+        # Reproduce the new tail of handle_mount: send mount frame, then
+        # drain push events, then dispatch async work.
+        response = {"type": "mount"}
+        await consumer.send_json(response)
+        await consumer._flush_push_events()
+        await consumer._dispatch_async_work()
+
+        consumer.send_json.assert_awaited_once_with(response)
+        consumer._flush_push_events.assert_awaited_once()
+        consumer._dispatch_async_work.assert_awaited_once()
+
+        # send_json must precede both drain calls (call ordering).
+        send_t = consumer.send_json.await_args_list[0]
+        flush_t = consumer._flush_push_events.await_args_list[0]
+        dispatch_t = consumer._dispatch_async_work.await_args_list[0]
+        # AsyncMock doesn't expose timestamps but the manual await order
+        # above is what we asserted; the structural test above pins the
+        # source order.
+        assert send_t is not None
+        assert flush_t is not None
+        assert dispatch_t is not None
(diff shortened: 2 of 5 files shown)


johnrtipton (commented):
## Stage 11 Code Review — PR #1292

**Verdict: REQUEST_CHANGES** (one tautology + one minor edge case)

The fix is correct and well-anchored against the canonical mirror pattern. Implementation matches the audit's intent even though placement is "after" rather than "before" — the deviation is justified and documented in code. Two findings below are worth a quick follow-up before merge.

---

### 🔴 Must-fix

None.

### 🟡 Should-fix

**1. `TestHandleMountDrainBehavior::test_tail_drains_both_queues_in_order` is a tautology** (per Action #1200, v0.9.4 retro). The test never invokes `handle_mount`. It manually awaits three `AsyncMock` methods in the desired order, then asserts that order. It would pass identically if `handle_mount` was empty / deleted / inverted. The structural tests pin source order, so the behavioral guarantee already exists — but advertising this as a "behavioral assertion" sets a misleading precedent for future tests.

   Two cheap repairs:
   - **Drop it** (the three structural tests already cover the source-order claim).
   - **Make it real**: arrange a `view_instance` with `_drain_push_events` returning one fake event, set `_async_tasks = {"x": (lambda: None, (), {})}`, drive a synthetic minimal `handle_mount` call (or extract the tail block into a helper and call that). Today's version asserts what the *test author* did, not what the SUT did.

**2. `_mount_one` collector seam swallows mount-time push events for batch-mounted views** (`websocket.py:2387-2459`). The collector swaps `self.send_json` only, but the new tail's `_flush_push_events()` calls `self.send_json` internally — so push-event frames from a mount() inside a lazy/batch-mounted view land in `captured[]` but are never extracted (the loop at 2431-2438 only handles `mount`/`error`/`navigate`). Net effect: `push_event()` from `mount()` works for top-level mounts (the issue this PR closes) but silently drops for `handle_mount_batch` children. Async work via `_dispatch_async_work` is fine because `asyncio.ensure_future` doesn't go through `send_json`.

   Not in scope for this PR (the cited issues are about top-level mounts). File a follow-up so the gap is tracked.

### 🟢 LGTM observations

- **Mirror-pattern alignment**: send-then-drain matches `_run_async_work` (921), `_flush_deferred_activity_events` (1211/1218/1287), and `handle_event` tails (3040-3047, 3365). Ordering rationale is correct — patches before mount frame would arrive at a not-yet-mounted DOM.
- **Guards verified**: both `_flush_push_events` (l.396) and `_dispatch_async_work` (l.801) have `if not self.view_instance: return`. The pre-mount-failure case is safe.
- **Source-shape test fragility trade-off is acceptable**: substring on `inspect.getsource` is brittle, but the structural-three's failure mode is loud + obvious if a refactor lands. Reasonable lock-in for a regression that quietly stranded production views forever.
- **Test fixture fix scope-discipline (Action #1079)**: the sync-→-async `_FakeConsumer._flush_push_events` change in `test_sw_advanced.py` and `test_sw_advanced_flow.py` is required by the new drain-on-mount path (the stub was unawaited dead code before; it's exercised now). Stayed scoped to the 2 cited fixtures, didn't try to refactor the broader fake-consumer landscape. Two-commit shape (Action #181) honored: c40920d1 (impl+tests), 3d8ca19 (fixture), d6e0f4f6 (CHANGELOG).
- **CHANGELOG entry quality**: captures the user-facing behavior (queued-but-never-spawned), names both issues, and references the test classes. Good.
- **No project-specific references** in the diff.
- **Code comment** on lines 2354-2363 explains the send-then-drain ordering decision well — future readers won't have to re-derive it.

---

### Verdict: REQUEST_CHANGES

The tautology test (🟡 #1) is the only blocking concern — Action #1200 was canonicalized exactly to catch this shape, and shipping it now would weaken that rule. The `_mount_one` gap (🟡 #2) is a separate ticket and shouldn't block.

johnrtipton:
# Retro: PR #1292 — Lifecycle mount drains (Group 1, v0.9.2-5)

**Quality**: 4.5/5

**What went well**:
- Audit-doc-driven fix landed clean. The lifecycle audit (PR #1282) had specified the exact site (`websocket.py:2352`) and the canonical mirror pattern (`_run_async_work` / `_flush_deferred_activity_events`); implementation was a 12-line addition. Closes #1280 + #1283 in one PR — both had the same one-liner shape, identified together by Audit A § Weakness #1 + #3.
- The send-then-drain ordering (vs. the issue body's "before" suggestion) was the correct call. The Stage 11 reviewer verified by tracing the mirror pattern in handle_event and confirmed: client must establish view via mount frame BEFORE async-driven patches arrive. Code comment captures the rationale for future readers.
- Test fixture compatibility fix (`_FakeConsumer._flush_push_events` sync→async in `tests/unit/test_sw_advanced.py` and `tests/integration/test_sw_advanced_flow.py`) was identified during pre-push and stayed scoped (Action #1079 — broader-sweep scope-discipline). The broken stubs were dead code before this PR; my drain call exposed them.

**What didn't go great**:
- The Stage 11 reviewer flagged a tautology test (`TestHandleMountDrainBehavior::test_tail_drains_both_queues_in_order`) — Action #1200's exact failure mode. The test never invoked handle_mount; just awaited 3 AsyncMocks in the desired order and asserted that order. Would pass even if handle_mount was deleted. Caught at Stage 11 (good); should have been caught at Stage 7 (better). Fixed in commit 902f7e5a — dropped the test entirely; the 3 source-shape tests already lock in the source-line claims.

**Lessons learned**:
- **Action #1200 fires reflexively now** — the reviewer caught the tautology immediately. The pattern "would this pass if the action didn't run?" is now part of the Stage 11 reflex.
- **The audit-as-pre-staged-work-graph recipe (#1259) keeps proving its worth**. Audit doc PR #1282 → 4 issues filed → implementation in 1 PR with 0 🔴 findings. The third consecutive milestone (v0.9.2-3 VDOM, v0.9.2-4 carryover, this) using this pattern.

**🟡 Follow-up filed**: `_mount_one` collector seam swallows mount-time push events for batch-mounted views. Out of scope for this PR's stated bug (top-level mounts); filed as separate issue.

**IDEA**: Stage 7 self-review checklist could grow a "tautology check" item that runs Action #1200's question against every new test. Cheap to add; would have caught this defect upstream.

Issue #1079, opened 2026-04-25
tech-debt: Stage 4 broader-sweep → follow-up issue scope-discipline (validated 2x)


johnrtipton:
**Source**: Retro v0.8.2 / Action Tracker row #154 / PRs #1067 #1076 (validated across v0.8.1 + v0.8.2)

**Action**: Add to Stage 4 plan-template guidance in `~/.claude/skills/pipeline-run/SKILL.md`: when Stage 4 investigation reveals a broader systemic issue beyond what the cited issue asks for, fix EXACTLY what the issue cites and file a follow-up issue for the systemic remainder. Resists scope creep while preserving the systemic finding.

**Empirical validation**:
- v0.8.1 PR #1067: investigating #1026's "Style-only follow-up" found a security leak. Stayed scoped to #1026; didn't expand into the broader `observability/views.py` audit.
- v0.8.2 PR #1076: investigating #1010's 4 cited stale .md refs found ~50 more across 17 files. Stayed scoped to #1010; filed follow-up #1075 for the broader sweep.

Two consecutive milestones now demonstrate the pattern. Promote from informal discipline to canonical Stage 4 plan-template note.

**Implementation**: one bullet under Stage 4 plan-template's scope-decision section. Skill-level work — labeled `out-of-scope-for-djust-drain`.

---
_Filed by /pipeline-retro v0.8.2 milestone retro Stage 4._

johnrtipton:
Shipped in PR #1226 (v0.9.1-7 cleanup batch). Comma-list `Closes #...` form parsed only the first ref — the rest closed manually. (This bites again; #1184's parenthesized form is documented in PR-checklist, but the comma-list-on-its-own-line form has the same auto-close failure.) 🤖 v0.9.1-7 cleanup batch closure.

Issue #1071, opened 2026-04-25
tech-debt: Doc-claim TDD extends to prose docs with external references


johnrtipton:
**Source**: Retro v0.8.1 / Action Tracker row #150

**Action**: Action #124 ("doc-claim-verbatim TDD") was filed for code claims. PR #1064 surfaced the same failure mode in PROSE docs: the new `docs/internal/codeql-patterns.md` cheat sheet cited 10 PR numbers, of which 7 were plausible-sounding hallucinations. Stage 11 reviewer caught all 10. Generalize: any prose doc that names external artifacts (PR numbers, issue numbers, commit hashes, file:line refs) must cross-check each citation at write time, not after.

**Implementation**: Add a Stage 5 (Implementation) discipline note in `~/.claude/skills/pipeline-run/SKILL.md` — "for any doc with external citations, run \`gh pr view <N>\` (or equivalent) per citation BEFORE commit". Skill-level work — labeled `out-of-scope-for-djust-drain`.

---
_Filed by /pipeline-retro v0.8.1 milestone retro Stage 4._

johnrtipton:
Shipped in PR #1226 (v0.9.1-7 cleanup batch). Comma-list `Closes #...` form parsed only the first ref — the rest closed manually. (This bites again; #1184's parenthesized form is documented in PR-checklist, but the comma-list-on-its-own-line form has the same auto-close failure.) 🤖 v0.9.1-7 cleanup batch closure.

Issue #1047, opened 2026-04-25
tech-debt: Upgrade Stage 7 checklist with user-flow trace — for every user-visible feature, trace the happy-path user story end-to-end (HTTP request → server dispatch → response envelope → browser render/navigation)


johnrtipton:
**Source**: Retro v0.7.0 / PR #989 (+ PR #986 + PR #988 pattern)
**Action Tracker**: row #125

**Action**: Upgrade Stage 7 checklist with user-flow trace — for every user-visible feature, trace the happy-path user story end-to-end (HTTP request → server dispatch → response envelope → browser render/navigation)

**Notes**: 3rd consecutive pipeline where Stage 7 rubber-stamped a diff that Stage 11 proved was broken end-to-end. PR #986 — JsonResponse outside try/except (response-layer). PR #988 — fire-and-forget flush breaking same-round-trip (transport-layer). PR #989 — HttpResponseRedirect silently dropped by @event_handler (dispatch-layer). Common shape: code does a thing, but thing doesn't reach the user. Enforcement: Stage 7 output template grows a "User flow trace" section with a required bullet per user-visible feature. **Validated across 5 pipelines** (#990, #993, #995, #996, #997 — all 0 🔴 at Stage 11; #995 + #996 ran condensed pipeline-dev flow with no Stage 11 but no live-verify regressions either). The class of defect that plagued #976/#988/#989 has not recurred since #125 was filed.

---
_Filed by `/pipeline-retro --reconcile` on 2026-04-25 as part of Scope C: file all Action Tracker rows as GitHub issues._

johnrtipton:
Shipped in PR #1226 (v0.9.1-7 cleanup batch). Comma-list `Closes #...` form parsed only the first ref — the rest closed manually. (This bites again; #1184's parenthesized form is documented in PR-checklist, but the comma-list-on-its-own-line form has the same auto-close failure.) 🤖 v0.9.1-7 cleanup batch closure.

PR #1293, opened 2026-05-01
fix(data_table): rename emit defaults + add prev/next + auto-refresh (#1275, #1279, #1291)

### LINKED ISSUE #1275: data_table tag emits 'table_sort' events but DataTableMixin registers 'on_table_sort' — every interaction returns 'No handler found'

### LINKED ISSUE #1279: DataTableMixin's on_table_sort/filter/page handlers don't refresh rows
# Upstream issue: DataTableMixin handlers don't trigger refresh_table()

**Status:** rate-limited at write time
**Title:** DataTableMixin's `on_table_sort` / `on_table_filter` / etc. don't refresh rows

## Summary

When a user clicks a sortable column header, filter dropdown, or page button,
the corresponding `on_table_*` handler updates the state vars
(`table_sort_by`, `table_filters`, `table_page`) — but it does NOT call
`refresh_table()`. The result: the *intent* is recorded in state, but
`table_rows` keeps its old order, so the rendered DOM doesn't change.

## Reproduction

```python
class OrderDataTableView(DataTableMixin, LiveView):
    table_model = Order
    table_columns = [{"key": "id", "label": "#", "sortable": True}, ...]
    table_default_sort = "-id"

        self.init_table_state()
        self.refresh_table()
```

```python
client = LiveViewTestClient(OrderDataTableView).mount()
client.send_event("on_table_sort", value="id")
# state: table_sort_by="id", table_sort_desc=False     ✅ updated
# rows:  table_rows[0].id == 22                         ❌ same as before
```

## Root cause

`python/djust/components/mixins/data_table.py` `on_table_sort()` (line 476)
ends after mutating two state vars. There's no auto-refresh; the user is
expected to override the handler and call `refresh_table()` themselves.

But the framework also doesn't document this expectation in the public
docstring, and the {% data_table %} template tag's `dj-click="table_sort"`
implies the click-to-resort flow works out of the box.

## Two related bugs in the same area

1. **#1275** — event names: template emits `dj-click="table_sort"` but
   the handler is called `on_table_sort`. The convention requires a custom
   `event_handler(name="table_sort")` decorator OR a passthrough method named
   `table_sort` that delegates to `on_table_sort`. Scaffold worked around
   this with passthroughs.

2. **THIS** — even after fixing the name, the rows don't refresh.

## Suggested fix

Either:

a) Append `self.refresh_table()` to each `on_table_*` handler (with a
   `@final`-style hook so subclasses can opt out). This matches the user
   mental model: "click sort → see rows reorder".

b) Document explicitly that consumers must call `refresh_table()` after
   every state-mutating handler, and provide a single `@auto_refresh`
   decorator that does the wrap.

(a) is much friendlier; (b) is more explicit.

## Workaround in scaffold

```python
@event_handler()
def table_sort(self, **kwargs):
    self.on_table_sort(**kwargs)
    self.refresh_table()  # ← this line is what fixes the bug
```

### LINKED ISSUE #1291: audit-B: `data_table` pagination handlers missing — `prev_event`/`next_event` map to no method
**Audit reference**: `docs/audits/decorator-contract-2026-05.md` § Weakness #2.

## Symptom

`{% data_table %}` template uses `dj-click="{{ prev_event }}"` and `dj-click="{{ next_event }}"` for backward-compatible pagination (`table.html:94, 96`). The defaults (`djust_components.py:512-513`) are `prev_event="table_prev"` and `next_event="table_next"`. Neither `table_prev` nor `on_table_prev` exists in DataTableMixin. Same for next.

Different from #1275 — #1275 is a prefix mismatch (`on_*` vs bare). Here, NEITHER name exists under EITHER convention. There's no handler to dispatch to. Pagination Prev/Next buttons hit "No handler found for event: table_prev/table_next".

## Root cause

`mixins/data_table.py:476-804` has `on_table_page` (line 523) which takes a page number as `value`, but no `on_table_prev` / `on_table_next`. The legacy `prev_event`/`next_event` defaults were apparently added to the tag signature without corresponding handler implementations.

`{% pagination %}` (djust_components.py:836) has the same shape: defaults `prev_event="page_prev"`, `next_event="page_next"` — also no matching handler.

## Reproducer

```html
{% load djust_components %}
{% data_table queryset=items paginate=True %}
```

Click "Next" — WS frame `{event: "table_next"}` — server returns "No handler found for event: table_next".

## Fix options

**Option A**: Add `on_table_prev` / `on_table_next` to DataTableMixin (or `table_prev` / `table_next` after #1275's prefix rename) that compute new page from current `self.table_page` and call `refresh_table()`. Symmetric for `{% pagination %}`.

**Option B**: Remove the "backward-compatible pagination" block from `table.html:92-97`; rely on `page_event` only. Document the breaking change.

**Option C**: Keep the legacy block but bind `prev_event`/`next_event` to `table_page` with computed values (template-side numeric arithmetic).

Recommendation: A — preserves backward compat, completes the contract. Bundle with #1275's prefix rename.

## Test

`tests/test_data_table_pagination_handlers.py` — exercise prev/next buttons; assert dispatcher resolves; assert page state advances.

## Related

- #1275 (prefix-rename across data_table) — bundle the fix here with the same PR; both involve `mixins/data_table.py` + `templatetags/djust_components.py` + `table.html`.
- Audit B Phase 1 #1 (per `docs/audits/decorator-contract-2026-05.md` § 6).

johnrtipton:
## Summary

Second v0.9.2-5 PR. Closes #1275 (data_table emits 23 names that don't
match handlers), #1291 (pagination handlers entirely missing), #1279
(handlers don't refresh rows). All 🔴 per
`docs/audits/decorator-contract-2026-05.md` § Weakness #1, #2, #4.

Single root cause: dispatcher does exact `getattr(view, event_name)`
at `websocket_utils.py:173`. Tag-emit defaults previously used bare
`table_*`; DataTableMixin uses `on_*` Phoenix-style. Result: every
default WS interaction returned "no handler found".

### #1275: rename event-name defaults across 4 files (92 lines)
- `templatetags/djust_components.py:509-595` — 23 tag kwargs
- `mixins/data_table.py:340-419` — class-level event-name attrs
- `mixins/data_table.py:197-281` — `_PRE_MOUNT_TABLE_CONTEXT`
- `components/rust_handlers.py:822-848`
- `templatetags/_forms.py:1408+`

Subclasses with overridden `table_X_event` attrs unaffected.

### #1291: add `on_table_prev` / `on_table_next` handlers
`mixins/data_table.py:540-560`. Clamped to `[1, table_total_pages]`,
refresh only when page actually changes.

### #1279: auto-refresh on row-affecting handlers
`on_table_sort` / `search` / `filter` / `page` / `prev` / `next` now
call `self.refresh_table()`. `on_table_select` deliberately exempt
(UI state, not row-set).

## Tests

15 cases in `python/djust/tests/test_data_table_handler_contracts.py`:
- `TestDataTableEmitToHandlerCrossReference` (3) — every emit default
  resolves to a handler.
- `TestPaginationHandlersExist` (6) — prev/next exist + clamp.
- `TestRowAffectingHandlersCallRefresh` (6) — refresh contract per handler.

Plus `EXPECTED_HANDLERS` in `test_data_table_mixin_liveview.py`
extended for the 2 new handlers (Action #1117 count-canary).

## Commit shape

Per Action Tracker #181:
- `21b719da` impl + tests (5 files, 372/98)
- `c9a4080f` count-canary update (Action #1117)
- `5af7c6d5` CHANGELOG entry

## v0.9.2-5 context

Group 2 of 3. Group 1 = mount drains (PR #1292). Group 3 = `@action`
re-raise.

Closes #1275
Closes #1279
Closes #1291


johnrtipton (commented):
# Stage 11 Code Review — PR #1293

**Verdict: APPROVE with one follow-up issue.**

The fix is correctly architected, internally consistent, well-tested at the unit/cross-reference layer, and the two-commit shape conforms to Action #181. All 15 new tests pass; the canary count test (21 → 23) catches the new handlers. Bulk-rename completeness is **almost** total, with one straggler that is a new finding outside this PR's stated scope.

---

## 🔴 Findings — none

## 🟡 Findings

### 🟡 1. Standalone `DataTable` Component (`components/components/data_table.py:26`) carries the same bug class outside the PR scope

`python/djust/components/components/data_table.py:26` defines

```python
sort_event: str = "table_sort",
```

and `_render_custom()` (line 63) emits `dj-click="{e_sort}"` literally. Any LiveView that uses `DataTable(...)` (the standalone Component) without overriding `sort_event` produces a WS frame `{"event": "table_sort"}` → exact-match dispatch → no `view.table_sort` exists → "no handler found" frame, identical to the #1275 failure mode.

This file is **not** in the audit's § 3 cross-reference table (which only catalogs the `data_table` template tag + `_PRE_MOUNT_TABLE_CONTEXT` + class-level mixin attrs). Strictly per Action #1079 (broader-sweep scope-discipline), keeping this PR scoped to the cited audit surface is correct.

**Recommendation**: file a follow-up issue ("standalone `DataTable` Component carries the table_sort default — same bug class as #1275"). Don't fix in this PR. Note: the new contract tests in `test_data_table_handler_contracts.py` only walk the tag function (`djust_components.data_table`), the mixin class attrs, and `_PRE_MOUNT_TABLE_CONTEXT` — the standalone Component would not be caught by Phase 2's linter as currently sketched either.

### 🟡 2. `python/tests/test_data_table_link_row_nav.py:43-63` ships stale `"table_*"` strings in `_base_ctx`

The test file's `_base_ctx` helper still uses the old bare-string defaults (`"table_sort"`, `"table_select"`, `"table_search"`, `"table_filter"`, `"table_page"`, `"table_prev"`, `"table_next"`). These are structural tests of `table.html` rendering (link column, row click, row URL) — they assert presence of `<a>` / `data-href` / `dj-click` attributes, **not** that the dj-click value resolves to a real handler. So tests still pass and runtime behavior is unaffected.

But the strings are now **misleading documentation** — anyone copy-pasting the fixture as a starting point reproduces the #1275 shape. Action #1071 (doc-claim-verbatim) generalizes to fixtures cited in test code: stale fixture-doc drift will mislead future readers.

**Recommendation**: 7-line follow-up commit (or absorb into this PR if doing a v0.9.2-5 fast-pass) updating the `_base_ctx` defaults to `on_table_*`. Tests will still pass.

### 🟡 3. No WS-level integration test exercising the actual click → dispatch → handler flow

The new test classes are all **unit-level**:
- `TestDataTableEmitToHandlerCrossReference` — symbol-table introspection only
- `TestPaginationHandlersExist` — direct method invocation on a `_MinimalTableView`
- `TestRowAffectingHandlersCallRefresh` — direct method invocation

No test spins up a `LiveViewConsumer`, sends a frame `{"event": "on_table_prev", ...}`, and asserts the dispatch lands on the new handler. For a bug class that lives **at** `websocket_utils.py:173`, the cross-reference test catches drift between the three layers but doesn't catch a regression to the dispatcher itself (e.g., if `_validate_event_security` started rejecting `on_*` names).

The cross-reference test is the right load-bearing assertion for #1275/#1291's *static* drift class. But per Action #1047 (Stage 7 user-flow trace), an end-to-end test verifying the full HTTP → WS → dispatch → handler → render cycle for at least one path (e.g., prev/next pagination) would close the loop. Lower priority since the dispatcher contract is exercised by ~60 other event-dispatch tests already in the suite.

**Recommendation**: optional. The 15 new cases plus the canary count test plus the existing `EventHandlerDecorationTest` are a sufficient defense for the cited bug class. A dispatcher-level test for the renamed handlers would be a nice follow-up.

## 🟢 Positive findings

- **Bulk rename completeness within scope**: 23/23 tag kwargs, 23/23 class-level mixin attrs, 30/30 `_PRE_MOUNT_TABLE_CONTEXT` entries, all `rust_handlers.py:822-928` defaults, `_forms.py` model_table defaults — all renamed. Verified via grep: zero remaining `"table_*"` event-name strings in the renamed-scope files.
- **`on_table_prev` / `on_table_next` correctness**: clamp logic `if self.table_page > 1 / < self.table_total_pages` correctly skips refresh on no-op (verified by `test_on_table_prev_clamps_at_one` / `test_on_table_next_clamps_at_total_pages`). Both decorated with `@event_handler()`. Both pass `is_safe_event_name` regex.
- **`refresh_table()` selection rationale is correct and documented in code**: row-affecting (sort/search/filter/page/prev/next) refresh; UI-only (select) doesn't. The `on_table_select` exemption is explicitly justified in its docstring (lines 511-514). Cell-edit / row-save don't refresh because they delegate to `handle_cell_edit` / `handle_row_save` overrides where the user controls persistence + refresh — correct decision.
- **No-op refresh on invalid input**: `on_table_page` correctly `return`s on `int(value)` failure before `refresh_table()` (line 536). Verified by `test_on_table_page_invalid_does_not_refresh`.
- **Backward compat preserved**: subclasses that overrode `table_sort_event = "my_handler"` keep their override since the rename only touched the **default**. The class attr is still read at the same call site.
- **EXPECTED_HANDLERS canary correctly bumped 21 → 23** (`test_data_table_mixin_liveview.py:132-133, 168-171`), with a comment pointing back to #1291. Action #1117 followed.
- **CHANGELOG entry** clearly flags the behavior change and notes the override-path is unaffected.
- **Two-commit shape per Action #181**: impl+tests (`21b719da`) → CHANGELOG (`5af7c6d5`) → canary (`c9a4080f`). Three commits, but the canary is a follow-on test fix; the CHANGELOG is correctly separate from impl.
- **No project-specific references** in the diff.

---

## Summary

The single root-cause framing is correct, the fix is the minimal-surface intervention, and the regression coverage is well-targeted. The only finding worth a follow-up issue is the standalone `DataTable` Component carrying the same default — that's a new finding outside the audit's catalog and should be its own issue per Action #1079.

Approve and merge.

johnrtipton:
# Retro: PR #1293 — data_table tag-name + handler completion (Group 2, v0.9.2-5)

**Quality**: 5/5

**What went well**:
- Bulk mass-rename across 4 files (92 line changes, 23/23 sites) using a single Python script (`/tmp/fix-data-table-event-names.py`) instead of incremental Edit-tool calls. Zero regressions, all tests passed first try after the rename. Single-script-transformation pattern (Action #180's safe alternative to incremental edits) is the right shape for systematic renames.
- Three issues bundled into one PR (#1275 + #1279 + #1291) because they share files (`mixins/data_table.py`, `templatetags/djust_components.py`, `table.html`) and a common root cause. The bundled fix demonstrated the audit's "23 emit-name mismatches" finding wasn't 23 separate bugs — it was one structural decision (tag emits bare names; handlers prefix with `on_*`). Fixing the boundary fixed all 23.
- The `EXPECTED_HANDLERS` count canary (Action #1117) caught my new handlers immediately and forced an explicit list update. Pre-push hook caught it; clean second commit (c9a4080f) bumped 21 → 23.
- 15 regression cases across 3 test classes covered cross-reference, pagination behavior, and refresh contract per handler. The cross-reference test (`test_every_data_table_emit_default_points_at_existing_handler`) is the most distinctive — it iterates the entire emit-name × handler-name matrix and would catch any future drift in either direction.

**Lessons learned**:
- **The `on_table_select`-is-exempt-from-refresh distinction matters**. UI-state vs row-state is a real semantic boundary; the audit explicitly called this out and the implementation honored it. The reviewer verified (🟢 LGTM observation). Worth canonicalizing: handler annotations could grow an explicit `@refreshes_rows` decorator OR a docstring convention so future mixin authors get this right by default.

**🟡 Follow-ups filed (3)**:
- Standalone `DataTable` Component (`components/components/data_table.py:26`) carries the same bug class — out of audit scope per #1079 broader-sweep scope-discipline, file separately.
- Stale fixture defaults in `python/tests/test_data_table_link_row_nav.py:43-63` — Action #1071 doc-citation discipline says these should match current naming.
- No WS-level integration smoke test exercising click → dispatch → handler — Action #1047 would close the loop.

**IDEA**: For any future "rename across the codebase" PR, the single-script-transformation pattern (Action #180) should be canon, not just an option. Incremental Edit-tool calls leave windows for partial state and risk inconsistent renames; a single script is atomic against the working tree.

Issue #1046, opened 2026-04-25
tech-debt: Upgrade Action #116 — for every feature with non-trivial semantics (gate rules, error envelopes, state contracts), write doc-claim-verbatim tests BEFORE writing implementation


johnrtipton:
**Source**: Retro v0.7.0 / PR #988 (+ v0.6.0/v0.6.1/#986 pattern)
**Action Tracker**: row #124

**Action**: Upgrade Action #116 — for every feature with non-trivial semantics (gate rules, error envelopes, state contracts), write doc-claim-verbatim tests BEFORE writing implementation

**Notes**: 4th consecutive milestone with doc-vs-code drift 🔴/🟡. Action #116 ("trace data-flow before writing docs") is aspirational, not executable. Upgrade to TDD sharpened: the test cases ARE the doc claims. Enforcement: Stage 7 checklist grows a "for each documented rule, point to the asserting test" row. PR #989 application: partial — five rule tests written RED first, but PR-body headline claim ("action fires → redirect to progress page") was never a test; that's the 🔴 Stage 11 caught. Subsumed for user-visible features by #125.

---
_Filed by `/pipeline-retro --reconcile` on 2026-04-25 as part of Scope C: file all Action Tracker rows as GitHub issues._

johnrtipton:
Shipped in PR #1226 (v0.9.1-7 cleanup batch). Comma-list `Closes #...` form parsed only the first ref — the rest closed manually. (This bites again; #1184's parenthesized form is documented in PR-checklist, but the comma-list-on-its-own-line form has the same auto-close failure.) 🤖 v0.9.1-7 cleanup batch closure.

PR #1294, opened 2026-05-01
fix(decorators): @action no longer re-raises after recording state (#1276)

### LINKED ISSUE #1276: @action: docstring promises inline save_note.error rendering on exception, but re-raise triggers dev overlay instead

johnrtipton:
## Summary

Third v0.9.2-5 PR. Closes #1276 — `@action` re-raised after recording
state, breaking the documented contract that templates can show
`{{ <name>.error }}` after an exception. 🔴 per
`docs/audits/decorator-contract-2026-05.md` § Weakness #3.

The re-raise routed the dispatcher to its exception-frame path which
sent `{"type": "error"}` to the client and bypassed the re-render.
The template never saw the recorded `error` field.

## What changed

`python/djust/decorators.py:340-393` — the @action wrapper now:

```python
except Exception as exc:
    # ... (was: raise)
    logger.exception("@action %s raised %s; recorded in _action_state[%r]", ...)
    self._action_state[action_name] = {"pending": False, "error": ..., "result": None}
    return None
```

`BaseException` subclasses (`KeyboardInterrupt`, `SystemExit`,
`GeneratorExit`) still propagate — by Python convention those
should never be silently caught.

Docstring (decorators.py:262-272) rewritten to document the new
contract: "caught and recorded; **not** re-raised".

## Tests

8 new cases in `python/djust/tests/test_action_decorator_contract.py`:
- `TestActionSuccessRecordsState` (1) — success path unchanged.
- `TestActionExceptionDoesNotPropagate` (6) — exception swallowed,
  recorded, logged; class-name fallback for empty messages;
  RuntimeError also swallowed; KeyboardInterrupt still propagates;
  ERROR-level log fires.
- `TestActionLazyInitializesActionState` (1) — defensive `_action_state`
  init.

6 existing tests in `test_action_decorator.py` updated to flip
`pytest.raises` assertions. `test_action_re_raises_after_recording`
renamed to `test_action_does_NOT_re_raise_after_recording`.

## Behavior change note

Code that wraps `@action` calls in `try/except` to handle the
re-raise now sees a clean `None` return. Mirror the old behavior by
re-raising explicitly inside the handler. CHANGELOG flags this.

## Commit shape

Per Action Tracker #181:
- `5f3e832a` impl + tests
- `3ba0bfb2` CHANGELOG entry

## v0.9.2-5 context

Group 3 of 3. Group 1 = mount drains (PR #1292). Group 2 = data_table
(PR #1293).

Closes #1276


johnrtipton (commented):
## Stage 11 Code Review — PR #1294

**Verdict: APPROVE** — 0 🔴, 1 🟡 (non-blocking), 2 🟢

---

### Implementation correctness — 🟢

Verified at `python/djust/decorators.py:381-403` (PR HEAD `3ba0bfb2`):

- `except Exception as exc` (line 381) — correctly excludes `BaseException`. Python convention: `Exception ⊂ BaseException`, so `KeyboardInterrupt`, `SystemExit`, `GeneratorExit` propagate. ✓
- State recorded with `pending=False`, `error=str(exc) or exc.__class__.__name__`, `result=None` (lines 398-402). ✓
- `logger.exception(...)` (line 392) — preserves traceback at ERROR level, NOT `logger.error()`. ✓
- `return None` (line 403) — clean exit, no re-raise. ✓
- Logger name resolves to `djust.decorators` via `logging.getLogger(__name__)` at line 16. ✓

**Cross-checked dispatcher claim**: `python/djust/runtime.py:629,642,710` and `python/djust/sse.py:131,179,195,204,239` and `python/djust/websocket.py:760,1919` all emit `{"type": "error", ...}` frames on exception. The PR's premise that re-raising routes to that path and bypasses re-render is materially correct.

### Test coverage — 🟢

`test_action_decorator_contract.py` — 8 new tests (count matches CHANGELOG):
- `TestActionSuccessRecordsState` (1) — success path lock-in
- `TestActionExceptionDoesNotPropagate` (6) — no-propagation, error-message capture, class-name fallback, RuntimeError, KeyboardInterrupt propagation, ERROR-level log fires (asserts `levelno == logging.ERROR` AND filters by `"@action" in msg` — robust)
- `TestActionLazyInitializesActionState` (1) — defensive `_action_state` init

`test_action_decorator.py` — 6 tests flipped from `pytest.raises` to clean return:
- `test_action_does_NOT_re_raise_after_recording` (renamed from `test_action_re_raises_after_recording`) is the closing case, with explicit `# Closes #1276` rationale.
- Spot-checked `test_action_falls_back_to_class_name_for_empty_message` (line 156) — note: previously raised `StopIteration` (a `BaseException` subclass since 3.5? No — it's `Exception`). The test now uses `ValueError()` which is also valid empty-message coverage. The change loses one minor signal (StopIteration is fine to catch since it's `Exception`-based) but doesn't reduce semantic coverage. Acceptable.

### `KeyboardInterrupt`/`SystemExit` semantics — 🟢

`except Exception` is correct. Python's class hierarchy: `BaseException` → `Exception` (caught here), and separately `BaseException` → `KeyboardInterrupt` / `SystemExit` / `GeneratorExit` (NOT caught). The test `test_action_keyboard_interrupt_still_propagates` and `test_baseexception_still_propagates` lock this in.

### Docstring — 🟢

`decorators.py:265-280` — clearly says "caught and recorded; **not** re-raised", explains the rationale (dispatcher's exception path bypasses re-render, references #1276), and explicitly notes `BaseException` subclasses still propagate.

### CHANGELOG — 🟢

`[Unreleased]` `### Fixed` block:
- "Behavior change" callout present (bold + explanatory text).
- Migration path: "Mirror the old behavior by re-raising explicitly inside the handler."
- Test count claim ("8 new regression cases") matches the actual count of 8.
- File paths cited are accurate.

### Project-specific references — 🟢

Diff reviewed; no internal/private references leaked into the public PR.

### Edge cases

**🟡 `@background` + `@action` interaction (worth noting, non-blocking)**

When `@background` wraps `@action`, the call shape is:
```
@event_handler @background  # (action's event_handler is innermost)
def handler():
    @action
    def inner(): ...
```
`@background` (`decorators.py:955-974`) calls `self.start_async(_async_callback, ...)` — the action wrapper runs INSIDE the background task. The new `return None` semantics now mean exceptions inside the background callback don't surface to `start_async`'s task; the action state is the only error signal. This is consistent with the PR's intent (template sees `{{ action.error }}`), but worth noting that the `handle_async_result(name, result, error)` hook (per `live_view.py` AsyncWorkMixin) won't fire on exception, since `@action` no longer raises. If a developer relies on `handle_async_result` for `@background @action`, they'll see `error=None`.

Recommendation (deferrable to follow-up issue): add a sentence to the `@background` docstring noting that combining with `@action` swallows exceptions and routes them through `_action_state[name]["error"]` instead of `handle_async_result`'s `error` parameter. Not blocking — the contract is internally consistent and the existing docstrings document each decorator's behavior independently.

### Strengths

- Two-commit shape (impl+tests / CHANGELOG) follows Action #181.
- Doc-claim TDD: every new behavior has a verbatim asserting test (Action #124).
- Comments at the catch site (`decorators.py:382-391`) document the why, not just the what.
- `caplog`-based log assertion (line 132) correctly uses `logger="djust.decorators"` and asserts `levelno == logging.ERROR` — wouldn't false-positive on a stray `logger.warning`.

### Files reviewed

- `/Users/tip/Dropbox/online_projects/ai/djust_project/djust/python/djust/decorators.py` (lines 1-16, 240-417)
- `/Users/tip/Dropbox/online_projects/ai/djust_project/djust/python/djust/tests/test_action_decorator.py` (full diff)
- `/Users/tip/Dropbox/online_projects/ai/djust_project/djust/python/djust/tests/test_action_decorator_contract.py` (full file, 164 lines)
- `/Users/tip/Dropbox/online_projects/ai/djust_project/djust/CHANGELOG.md` (Unreleased block)
- Cross-referenced: `runtime.py`, `sse.py`, `websocket.py` for dispatcher exception-frame paths

**Approve to merge.** The 🟡 `@background` interaction note can become a follow-up docstring tweak if the team wants — not blocking on this PR.

johnrtipton:
# Retro: PR #1294 — `@action` re-raise contract fix (Group 3, v0.9.2-5)

**Quality**: 5/5

**What went well**:
- Doc-claim-verbatim TDD pattern (#1046) closed the loop perfectly. The audit caught the docstring-vs-implementation drift; the fix made the docstring's promise true; the new test class `TestActionExceptionDoesNotPropagate` asserts the docstring claims directly. 4 consecutive milestones now using this pattern have produced clean Stage 11 outcomes.
- The `Exception` vs `BaseException` distinction was tested explicitly (`test_action_keyboard_interrupt_still_propagates` + `test_baseexception_still_propagates` in the new test file). Stops a future reviewer from "helpfully" widening the catch to `BaseException` and breaking Ctrl-C semantics.
- 6 existing tests in `test_action_decorator.py` updated to flip the contract assertion — straightforward find-and-replace plus renaming the `test_action_re_raises_after_recording` test to `test_action_does_NOT_re_raise_after_recording`. The rename makes the contract reversal explicit in the test name itself; future readers won't be confused.
- Behavior-change CHANGELOG entry was up-front about the migration path: "Code that wraps `@action` calls in `try/except` to handle the re-raise now sees a clean None return. Mirror the old behavior by re-raising explicitly inside the handler." Stage 11 reviewer confirmed this is the right level of detail.

**Lessons learned**:
- **Behavior-change PRs need a "Migration" CHANGELOG block**. The CHANGELOG entry already had this; worth canonicalizing as a Stage 9 checklist item. When the PR changes a documented API contract, the CHANGELOG should explicitly call out the migration path. Future readers shouldn't have to derive it from the diff.
- **`logger.exception()` (not `logger.error()`) is the right level for caught-and-swallowed errors**. Preserves traceback even when the exception is intentionally not propagating. This is now codified in the new docstring at decorators.py:262-272.

**🟡 Follow-up filed**: `@background + @action` combo: `handle_async_result(name, result, error)` won't see `error` populated because @action swallows it; only `_action_state[name]["error"]` carries the signal. Recommend follow-up docstring tweak to clarify.

**IDEA**: A "decorator contract spec test" file (planned in audit B Phase 3) should be commissioned earlier in v0.9.3. The test pattern from this PR (assert each docstring claim has a covering test) generalizes; doing it for all 12 decorators at once would catch sibling drift.

PR #1301, opened 2026-05-01
fix(transition): dj-transition accepts 1-token short form (#1273)

### LINKED ISSUE #1273: dj-transition-group: documented short-form (fade-in | fade-out) silently rejected by _parseSpec (3-token requirement)
## Summary

`dj-transition-group` documents a "short form" syntax (`"fade-in | fade-out"`) that's silently rejected by `_parseSpec` in `client.js`, which requires *exactly 3 space-separated classes* per side.

## Where the docs disagree with the implementation

`python/djust/static/djust/src/43-dj-transition-group.js`:

```js
// Usage — short form (one attribute, pipe-separated):
//   <ul dj-transition-group="fade-in | fade-out">...</ul>
//   <ul dj-transition-group="opacity-0 t-opacity-300 opacity-100 | opacity-100 t-opacity-300 opacity-0">...</ul>
```

Both forms are presented as supported. But `dj-transition-group` just sets the per-child `dj-transition` (or `dj-remove`) attribute to the spec string verbatim — and `_parseSpec` in client.js rejects anything with fewer than 3 tokens:

```js
function _parseSpec(raw) {
    const parts = input.split(/\s+/).filter(Boolean);
    if (parts.length < 3) return null;
    return { start: parts[0], active: parts[1], end: parts[2] };
}
```

So `dj-transition="fade-in"` → `_parseSpec` returns null → no animation runs. The "short form" silently no-ops.

## Reproduction

```html
<div dj-transition-group="fade-in | fade-out">
  <div>existing</div>
</div>
<button onclick="document.querySelector('[dj-transition-group]').appendChild(Object.assign(document.createElement('div'), {textContent: 'new'}))">add</button>
```

The new element appears with no animation. DevTools shows `dj-transition="fade-in"` on it but no class swap happens.

## Suggested fix

Either:
1. **Make `_parseSpec` accept 1 token as the active class** (use `transitionend` to clean up) — matches the existing `dj-remove="single-class"` semantics and matches the docs example.
2. **Remove the short-form example from the comment** and require the 3-token form everywhere.

## Surfaced via

`djust-org/djust-scaffold` PR #30 transition demo (https://github.com/djust-org/djust-scaffold/pull/30) — initially used `dj-transition-group="fade-in | fade-out"` (matching the docs example), saw nothing animate. Workaround in scaffold: switched to the explicit 3-class form.

johnrtipton:
## Summary
Closes #1273. dj-transition-group docs advertise 1-token short form (`fade-in | fade-out`) but `_parseSpec` required 3 tokens, silently rejecting the ENTER side. Audit D § Documented Grammar.

## What changed
`_parseSpec` accepts 1-token form (`{ single: parts[0] }`); `_runTransition` handles it by applying the class on next frame and waiting for transitionend. 2-token form remains rejected as ambiguous.

## Tests
2 new regression cases in `tests/js/dj_transition.test.js`. 11 dj-transition tests pass; 24 sibling tests in dj_transition_group + dj_remove still pass.

## v0.9.2-6 context
PR 1/5. Audit D class.

Closes #1273


johnrtipton:
# Retro: PR #1301 — dj-transition 1-token short form (v0.9.2-6 #1273)

**Quality**: 5/5

**What went well**:
- Smallest PR in the v0.9.2-6 drain (~1 day, S effort) shipped clean — 2 new test cases + 1 renamed existing test. The audit's exact citation (`41-dj-transition.js:45-60`) made this a one-edit-and-test fix.
- Symmetry pattern paid off: `dj-remove`'s 1-token form already existed (line 52); aligning `dj-transition` is just contract-following. The code comment in `_runTransition` explicitly references the dj-remove sibling for future readers.
- Rejected 2-token form (defensive symmetry with dj-remove). Future authors who try `dj-transition="opacity-0 opacity-100"` get a debug warning instead of mystery silent rejection.

**Lessons learned**:
- **Audit doc → fix → tests in one short loop is the proven shape.** The audit specified "extend `_parseSpec` to accept 1-token form, return `{single: <class>}`"; implementation matched verbatim. Pattern continues to deliver 0 🔴 reviews.

**IDEA**: Action #1046 (doc-claim-verbatim TDD) extended naturally to JS-side parser grammars here. The new behavioral test `1-token form applies the class on next frame` directly exercises the docstring's promise. Worth canonicalizing: any time a JS module's docstring shows a syntax example, there should be a test asserting that example actually works.

Issue #1290, opened 2026-05-01
audit-B: add `scripts/check-handler-contracts.py` symbol cross-reference linter


johnrtipton:
**Audit reference**: `docs/audits/decorator-contract-2026-05.md` § Weakness #8 + § 7 (linter sketch).

## Why

Class-of-bug #1275: a template tag's `*_event=` default doesn't match any handler in the corresponding mixin. Currently caught only at user-click time when the dispatcher returns "no handler found". The audit revealed 23 such mismatches in `data_table` alone.

A static check at pre-push time would catch this class entirely.

## Proposal

Add `scripts/check-handler-contracts.py` (~50 LoC). Walk the AST of:
- `python/djust/components/templatetags/djust_components.py` — extract every kwarg ending in `_event` with a non-empty string default
- `python/djust/components/mixins/*.py` — extract every method def name on classes

Cross-reference: for each emit default, look up the matching handler. Fail (exit 1) if any default isn't covered.

Full sketch in the audit doc § 7.

## Wire-up

1. Add the script under `scripts/` with shebang + no external deps (stdlib only).
2. Add Make target: `make check-handler-contracts`.
3. Add to pre-push hook in `.pre-commit-config.yaml` with `pass_filenames: false`.
4. Document in `docs/PULL_REQUEST_CHECKLIST.md` — "any change to a template-tag's `*_event=` default OR a corresponding mixin handler must run `scripts/check-handler-contracts.py` before commit".

## Phase split

**Foundation**: walk djust_components.py + components/mixins/ only. ~50 LoC. Catches the cited bug class within the framework.

**Capability** (later): extend to third-party packages registering djust components. Provide a programmatic API: `register_tag_emit_default(tag, param, handler)` that the linter can introspect. Not needed until the framework has a community of third-party component registrars.

## Test

`tests/test_check_handler_contracts.py` — point the script at a fixture with a known mismatch; assert exit code 1 and stderr message.

Then ALSO run the linter against the current `main` and confirm the 23 data_table mismatches are reported (or are already fixed by the time this issue lands; in which case the linter should pass).

PR #1302, opened 2026-05-01
fix(serialization): AsyncResult emits dict via to_dict() (#1274)

### LINKED ISSUE #1274: AsyncResult not in serializer whitelist — assign_async demos render blank (template can't access .loading/.ok/.failed)
## Summary

`AsyncResult` is the framework's own type returned by `assign_async()`, but the framework's JIT serializer doesn't recognize it as serializable. It gets coerced to `str(AsyncResult)` during state serialization, which means template lookups like `{% if users.loading %}` evaluate to falsy because strings don't have a `.loading` attribute.

End result: every `assign_async` demo shows blank tiles, no skeleton, no value — silently. The serializer prints a WARN but nothing surfaces in the rendered page.

## Reproduction

```python
class DashboardView(LiveView):
        self.assign_async("users", lambda: {"count": 1234})
```

```html
<div>
    {% if users.loading %}<div class="skeleton"></div>{% endif %}
    {% if users.ok %}<div>{{ users.result.count }}</div>{% endif %}
</div>
```

Expected: skeleton renders initially, then real count after loader resolves.
Actual: nothing renders. Server log shows:

```
WARNING djust.serialization LiveView state contains non-serializable value: AsyncResult (from djust.async_result).
This will be converted to a string, which may cause AttributeError on deserialization.
```

## Root cause

The serializer's whitelist of "known safe types" doesn't include `AsyncResult`. Since `AsyncResult` is a frozen dataclass shipped *by djust itself* and is the documented return shape of `assign_async`, it should be either:

1. Added to the serializer's whitelist (preferred — automatic, no user friction)
2. Documented as requiring a `_` prefix (but that breaks the documented `{% if users.loading %}` template pattern, since underscore-prefixed attrs aren't exposed to templates)

## Workaround in user code

```python
    # Manually expose the AsyncResult fields as plain dicts so the
    # serializer doesn't choke. Template still reads `users.loading`
    # etc. unchanged.
    for name in ("users", "orders", "revenue"):
        r = getattr(self, name, None)
        if r is not None and hasattr(r, "loading"):
            ctx[name] = {"loading": r.loading, "ok": r.ok, "failed": r.failed,
                          "result": r.result, "error": str(r.error) if r.error else None}
    return ctx
```

## Surfaced via

`djust-org/djust-scaffold` PR #30 suspense demo (https://github.com/djust-org/djust-scaffold/pull/30) — `/features-v09/suspense/` rendered three blank tiles.

johnrtipton:
## Summary
Closes #1274. `assign_async()` returns `AsyncResult` instances that templates expect to read as `{{ users.loading }}` / `{{ users.ok }}` / `{{ users.failed }}` / `{{ users.result }}` / `{{ users.error }}`. The serializer had no AsyncResult branch — value fell through to `str()`, producing strings templates couldn't navigate. Audit F § Serializer Allowlist.

## What changed
- `AsyncResult.to_dict()` (async_result.py) — returns the 5-field dict templates depend on; error stringified.
- `DjangoJSONEncoder._default_impl` (serialization.py) — registers AsyncResult.
- `normalize_django_value` (serialization.py) — recurses into to_dict() so non-primitive `result` payloads (Model, datetime, Decimal nested inside) are normalized too.

## Tests
11 new regression cases in `python/djust/tests/test_async_result_serializer.py` (4 test classes covering to_dict shape, normalize path, json.dumps path, template-access proof). 40 existing async tests pass.

## v0.9.2-6 context
PR 2/5. Audit F class.

Closes #1274


johnrtipton:
# Retro: PR #1302 — AsyncResult serializer (v0.9.2-6 #1274)

**Quality**: 5/5

**What went well**:
- The `to_dict()` shape was the right level of abstraction — templates need 5 fields (`loading`, `ok`, `failed`, `result`, `error`); the method returns exactly those. Future serializer changes can update `to_dict()` without touching either of the two registration sites.
- Recursion handled correctly: `normalize_django_value` recurses into `to_dict()`'s output so an inner `result` payload (Model, datetime, Decimal nested inside) gets normalized too. The `test_succeeded_recurses_into_result` case locks this behavior.
- 11 regression cases across 4 test classes (to_dict shape, normalize path, json.dumps path, template-access proof). The template-access test (`test_loading_key_truthy_when_pending`) is the most distinctive — it directly proves the failure mode #1274 reported (`{% if users.loading %}` evaluating falsy on a string) is now fixed.

**Lessons learned**:
- **Two serializer paths means two registration sites**. `DjangoJSONEncoder.default()` and `normalize_django_value()` are both load-bearing. Audit F caught this dual-registration shape; future framework-shipped types should register in both up-front. Worth canonicalizing in the eventual Audit F linter (#1290) — type-allowlist checker that fails CI if a public type has fewer than 2 registrations.

**IDEA**: extend `to_dict()` pattern to other framework-shipped value objects: `RateLimitState`, `ActionState`, anything else templates need to read attributes from. Establishes a serialization protocol: opt into JSON-safety by implementing `to_dict()`.

PR #1303, opened 2026-05-01
fix(submit): form submit flushes pending dj-input debounces (#1278)

### LINKED ISSUE #1278: Form submit races with debounced dj-input events — last keystroke can be lost
# Upstream issue: dj-input debounce races with form submit

**Status:** rate-limited at write time (GraphQL quota); REST quota is fine
**Title:** Form submit races with debounced dj-input events — last keystroke can be lost

## Summary

When a form contains text/email/password inputs with `dj-input` bindings, the
framework debounces those events 300ms before sending them to the server. If a
user (or test) tabs through fields and submits the form within that window,
the **submit handler fires before the most-recent input event reaches the
server**. The form's state appears empty and validation fails.

## Reproduction

scaffold app `apps.features_v09.views_wizard.WizardSignupView` (uses
`WizardMixin` with `wizard_input_event = "dj-input"`):

```python
# Step 1: AccountForm has email + password fields
# Form template: <form dj-submit="next_step">
# Each input: <input dj-input="validate_field" data-field_name="email" type="email">
```

Playwright reproduction (`tests/e2e/test_demos.py::test_wizard_step_navigation`
before the workaround):

```python
page.fill('input[name="email"]', "e2e@example.com")  # t=0:    starts 300ms debounce
page.fill('input[name="password"]', "password123")    # t=10ms: starts 300ms debounce
page.click('form button[type="submit"]')              # t=20ms: fires next_step IMMEDIATELY
                                                       # t=20ms: server's wizard_step_data is empty
                                                       # t=320ms: validate_field events finally fire — too late
```

Result: `wizard_step_index` does not advance, no error shown to user.

## Root cause (suspected)

`python/djust/static/djust/src/08-event-parsing.js` defaults
`text` / `email` / `password` / `tel` / `url` / `search` / `textarea` inputs to
`{ type: 'debounce', ms: 300 }`. The submit handler in `09-event-binding.js`
fires immediately without flushing pending debounce timers.

## Suggested fix

On `submit` events, `_dispatchHandler` (or wherever submit dispatch lives)
should:

1. Find any pending debounce timers for inputs inside the submitting form
2. **Flush them synchronously** (fire the queued event payload now, not after
   the timer expires)
3. THEN fire the form's `dj-submit` event

This matches the user mental model: "by the time I press submit, all my
inputs are recorded."

## Workarounds

- Add a 400ms wait between fill and click in tests
- Use `dj-change` instead of `dj-input` (fires on blur, no debounce)
- Override per-input debounce via attribute (if supported — verify)

## Affected

- WizardMixin (any multi-step form)
- Any LiveView with auto-validation on input + form submit pattern
- Test suites doing programmatic form-fill (Playwright, Selenium, etc.)

johnrtipton:
## Summary
Closes #1278. Text inputs default to 300ms debounce on dj-input. A user typing then immediately clicking submit raced past pending input events; views relying on dj-input to populate server-side state saw stale data. Audit G § Event Ordering.

## What changed
- `debounce()` (09-event-binding.js) now exposes `.flush()` that fires the pending invocation synchronously.
- New `_flushPendingDebouncesInForm(form)` iterates `[dj-input]` descendants and flushes their wrappers.
- `_handleDjSubmit` calls the flush helper BEFORE reading FormData / dispatching the submit handler.

Send-then-flush ordering: dj-input dispatch must happen synchronously before submit, so the server sees the typed values populated by per-field handlers when the submit handler reads them.

## Tests
4 new regression cases in `tests/js/dj_submit_debounce_flush.test.js`. 30 sibling tests in debounce.test.js + dj-debounce-throttle.test.js still pass.

## v0.9.2-6 context
PR 3/5. Audit G class.

Closes #1278


johnrtipton:
# Retro: PR #1303 — form submit debounce flush (v0.9.2-6 #1278)

**Quality**: 5/5

**What went well**:
- The fix uncovered a subtle distinction the audit had glossed over: FormData on submit captures the typed DOM values (so views reading from form params are unaffected), but `dj-input` populates SERVER-SIDE STATE per-keystroke. Views that read from `self.email` etc. (not from form params) are the ones that race. Capturing this in the commit message + CHANGELOG behavior-change note prevents future readers from confusing the bug class.
- `debounce()` getting a `.flush()` method is broadly useful beyond #1278. Anywhere debounced state needs to land before a downstream operation (close, navigate, submit), the same pattern applies.
- 4 regression cases including the negative case (`flush is a no-op when no debounced input is pending`) — important because that's the most common path (most submits don't have pending debounces).
- The bidirectional ordering test (`emailIdx < submitIdx`) is the load-bearing assertion. `expect(eventNames).toContain(...)` would pass even if dj-input fired AFTER submit; explicit index ordering catches the regression.

**Lessons learned**:
- **Debounced wrappers should expose `.flush()` from day one.** Several existing debounce uses in djust (rate-limited handlers, etc.) might benefit from flushing on focus loss / navigation / unmount. Worth surveying.
- **The `latestArgs` pattern from `blur` mode is similar shape** — `_inputRateLimitState`'s `wrapped` already needed to track pending state for `blur`. The new `flush()` integration is consistent.

**IDEA**: extend the flush pattern to `dj-change` events too. Some apps debounce text inputs with `dj-change` (less common than dj-input but possible); same race exists. Audit G originally cited only the dj-input path; the broader pattern is worth a follow-up.

PR #1304, opened 2026-05-01
fix(dj-dialog): native close (ESC/backdrop) syncs back to server (#1267)

### LINKED ISSUE #1267: dj-dialog: client-close (ESC/backdrop) doesn't sync back to server show_dialog state
## Summary

`dj-dialog` is one-way sync: server-controlled `dj-dialog="open"|"closed"` → client `showModal()` / `close()` works. But when the user closes the dialog client-side (ESC key, backdrop click, or the `<dialog>`'s native close mechanism), the server's state attribute that drove `show_dialog` stays True. The next "Open" click is a no-op because the server already thinks the dialog is open.

## Reproduction

In `djust-org/djust-scaffold` PR #33 (https://github.com/djust-org/djust-scaffold/pull/33), the `/components/overlay/` page has:

```python
@event_handler()
def open_dialog(self, **kwargs):
    self.show_dialog = True
```

```html
<dialog dj-dialog="{% if show_dialog %}open{% else %}closed{% endif %}">
    <button dj-click="close_dialog">Cancel</button>
</dialog>
```

1. Click "Open dialog" → `show_dialog=True` → dialog shows
2. Hit ESC (or click backdrop) → dialog closes client-side
3. Click "Open dialog" again → no patch sent, dialog stays closed (because the server's `show_dialog` is still True from step 1, so the re-render produces no diff)

## Expected

When the user closes the dialog client-side, the framework should fire a configurable handler on the server so it can update its state. Phoenix LiveView's `phx-cancel` attribute is the analog.

## Suggested API

```html
<dialog dj-dialog="..." dj-dialog-close-event="close_dialog">
  ...
</dialog>
```

When the `<dialog>`'s native `close` event fires (ESC, backdrop, or programmatic close), the framework dispatches `close_dialog` to the LiveView so it can do `self.show_dialog = False`.

Alternative: document this as a known limitation and recommend always using a server-driven close button (no ESC/backdrop handling). This is workable but counterintuitive given that the framework's docs marketing says "ESC dismiss + backdrop click for free".

## Provenance

Surfaced by the code-reviewer subagent's review of `djust-org/djust-scaffold` PR #33: https://github.com/djust-org/djust-scaffold/pull/33#issuecomment-4358112215

johnrtipton:
## Summary
Closes #1267. dj-dialog was one-way (server→client); user closing dialog client-side left server convinced it was still open. Re-opening from server became no-op. Audit C § Bidirectional Binding.

## What changed
- `_installCloseListenerOnce(el)` — attaches a native `close` event listener to `<dialog>` (idempotent via WeakMap).
- `_syncDialogState(el)` calls it on every sync.
- New attribute `dj-dialog-close-event="..."` opts into the reverse sync; listener reads it at fire time and dispatches via `handleEvent(name, {_targetElement: el})`.

## User-facing API
```html
<dialog dj-dialog="open" dj-dialog-close-event="close_settings">...</dialog>
```

## Tests
5 new regression cases in `tests/js/dj_dialog.test.js`. 13 dialog tests pass total.

## v0.9.2-6 context
PR 4/5. Audit C class.

Closes #1267


johnrtipton:
# Retro: PR #1304 — dj-dialog reverse sync (v0.9.2-6 #1267)

**Quality**: 5/5

**What went well**:
- `dj-dialog-close-event` as the opt-in attribute is a clean API extension — explicit configuration, no behavior change for users who don't set it. Backward-compatible by default.
- The WeakMap idempotency guard (`_dialogsWithCloseListener`) is the right shape; the existing pattern from `41-dj-transition.js` (`_djTransitionState`). Multiple sync passes in the lifetime of a single `<dialog>` element don't stack listeners.
- Reading the attribute at fire time (not at install time) means VDOM morphs that update `dj-dialog-close-event` post-mount take effect on the NEXT close, not stuck with the original value. The `test_reads_dj_dialog_close_event_at_fire_time_morph_after_mount` case locks this in.
- 5 regression cases span happy path, opt-in absence, idempotency, attribute updates, module export — comprehensive.

**Lessons learned**:
- **Bidirectional binding requires opt-in for client→server**. A blanket `close`-event listener that always dispatched would surprise apps that didn't ask for it. The opt-in attribute pattern (`dj-dialog-close-event="..."`) is the right shape and generalizes to other audit-C class issues (`<details>`, native form `reset`, video/audio playback state).
- **Audit C's broader inventory is worth doing.** This PR fixed `<dialog>`; other native HTML5 elements with built-in state (details, form, video) likely have the same one-way-binding gap. Filed as audit-C follow-up scope.

**IDEA**: A "dj-* native event opt-in" canonical attribute pattern: `dj-{tag}-{event}-event="..."`. Establish in v0.9.3 as a framework convention so future bidirectional-binding fixes follow a consistent shape.

Issue #1281, opened 2026-05-01
Private state changes alone don't trigger Rust diff re-render


johnrtipton:
# Upstream issue: private state changes alone don't trigger Rust diff re-render

**Repo to file in:** djust-org/djust
**Status:** rate-limited at write time

## Summary

When a handler modifies ONLY private (`self._x`) state, the Rust diff
returns no patches and the WS sends `noop` — even though the rendered
template depends on that private state via `get_context_data()`.

The diff appears to compare only public (non-underscore) state for
"is something different?" detection. If only private state changes
between renders, the change tracker says "nothing changed" and short-
circuits before re-rendering the template.

## Reproduction (apps.realtime.views_notify.NotifyOnSaveView)

```python
class NotifyOnSaveView(BaseLiveView):
    def _refresh(self):
        self._orders = Order.objects.all()[:10]   # private

    def get_context_data(self, **kwargs):
        ctx = super().get_context_data(**kwargs)
        self._refresh()
        ctx["order_count"] = Order.objects.count()  # always fresh
        ctx["orders"] = [...]                        # always rebuilt
        return ctx

    @event_handler()
    def create_demo_order(self, **kwargs):
        Order.objects.create(...)   # DB row created (verified)
        self._refresh()               # only updates self._orders (private)
```

Browser observation (Playwright WS-frame capture):

```
[sent] {"type":"event","event":"create_demo_order","ref":1}
[recv] {"type": "noop", "ref": 1}
< no patch frame ever arrives >
```

DB count went up (verified by HTTP refresh). `get_context_data` returns
the new `order_count`. But Rust's diff returned `noop`.

## Smoking-gun comparison

Python-only path (`LiveViewTestClient.render_with_patches()`) produces
**41 patches** for the same scenario:

```python
client = LiveViewTestClient(NotifyOnSaveView).mount()
html_pre, _, _ = client.render_with_patches()         # 26 orders
client.send_event("create_demo_order")
html_post, patches, _ = client.render_with_patches()  # 27 orders
# patches has 41 entries — header SetText, row inserts, etc.
```

So `render_with_diff()` itself works correctly. The issue is upstream of
the actual diff — somewhere in the WS handle_event path, the framework
decides "no public state changed → skip render → noop" before
`render_with_diff()` even runs.

## Workaround (in scaffold)

Add a public counter that bumps on every event that modified private state:

```python
def mount(self, request, **kwargs):
    super().mount(request, **kwargs)
[...]

PR #1305, opened 2026-05-01
fix(sse): EventSource sends Django session cookie (#1277)

### LINKED ISSUE #1277: SSE transport: Django session cookie not passed to EventSource GET — auth loop on every mount
## Summary

When `use_websocket: False` is set and the SSE transport is active, authenticated views redirect to login on every mount. The `EventSource` GET request to `/djust/sse/<session_id>/` does not carry the Django `sessionid` cookie, so `check_view_auth` sees an unauthenticated user and pushes `{"type": "navigate", "to": "/auth/login/"}`. The client navigates to login, which re-mounts the LiveView, which triggers another unauthenticated SSE request — infinite loop.

## Observed behaviour

Console log on every page load:
```
[SSE] Received: sse_connect {session_id: "..."}
[SSE] Connection acknowledged, session: ...
[SSE] Received: mount {view: "apps.claims.views.ClaimDetailView", ...}
[SSE] View mounted: apps.claims.views.ClaimDetailView
[SSE] Received: navigate {"type":"navigate","to":"/auth/login/"}
[LiveView] Initializing...   ← re-init triggered by navigate
[SSE] Connecting ...          ← new session, same result
```

## Environment

- djust 0.9.2rc1
- Django session auth (standard `SessionMiddleware` + `AuthenticationMiddleware`)
- Daphne ASGI server
- `USE_WEBSOCKET=false` / `LIVEVIEW_CONFIG = {"use_websocket": False}`

## Root cause

The Django session cookie is set on the browser, but the `EventSource` constructor in `LiveViewSSE._connect()` creates the SSE stream as a plain GET request. In an ASGI environment (Daphne), the ASGI HTTP scope for the SSE GET request does not always include the session cookie in a way that Django's session middleware resolves correctly.

The WebSocket path works correctly because the WS handshake carries the cookie and Daphne's WS scope processing correctly surfaces it to Django.

Two possible failure modes:
1. The `EventSource` GET fires before the session cookie is set (race on first login)
2. Daphne's ASGI HTTP handler for the SSE route doesn't attach cookies to the request the same way a normal HTTP view does

A quick diagnostic: adding `print(request.COOKIES)` in `DjustSSEStreamView.get()` shows an empty dict even when the browser has the session cookie.

## Expected behaviour

The SSE stream GET should carry the session cookie so `check_view_auth` sees the authenticated user, identical to how the WebSocket mount path works.

## Workaround

None found. The SSE transport is unusable for cookie-authenticated views until this is resolved.

## Notes

- This is the fourth gap discovered while testing SSE fallback on the NYC Claims prototype (see #1237 for the first three)
- The three issues in #1237 appear to be fixed in 0.9.2rc1 based on code inspection — the session auth issue blocks testing them end-to-end
- WebSocket mode (`use_websocket: True`, the default) is unaffected

🤖 Filed from NYC Claims prototype SSE fallback testing

johnrtipton:
## Summary
Closes #1277. SSE EventSource GET (and message POST) didn't carry credentials; authenticated views failed `check_view_auth` on every mount, triggering an infinite mount→navigate loop. Audit E § Transport Contract.

## What changed
- `03b-sse.js:69` — EventSource constructor now passed `{withCredentials: true}`.
- `03b-sse.js:367` — sendMessage() POST sets `credentials: 'include'`.

EventSource spec defaults credentials to "same-origin", but explicit `withCredentials: true` is the conventional opt-in that avoids browser-version ambiguity.

## Tests
2 new regression cases in `tests/js/sse-transport.test.js`. 26 SSE tests pass total.

## v0.9.2-6 context
PR 5/5 — final pre-stable PR. Audit E class. After this lands, v0.9.2 stable is cuttable (with #1281 documented as known issue per ROADMAP v0.9.3 deferral).

Closes #1277


johnrtipton:
# Retro: PR #1305 — SSE EventSource cookie auth (v0.9.2-6 #1277)

**Quality**: 5/5

**What went well**:
- The fix is a one-liner per call site (EventSource constructor + fetch options) — small surface, low risk.
- Sibling fix on `sendMessage`'s POST (`credentials: 'include'`) pre-empts a symmetric failure that would have surfaced post-merge: even if EventSource GET worked with default credentials, the dispatch POST might not have been authenticated. Fixing both at once means SSE auth is consistent across both legs of the transport.
- The CHANGELOG entry's framing — "explicit `withCredentials: true` is the conventional opt-in that avoids browser-version ambiguity" — explains WHY we set it explicitly even though the spec defaults credentials to "same-origin". Future readers won't strip it thinking it's redundant.

**Lessons learned**:
- **Transport-contract canon (Audit E): every transport's auth/cookie story should be locked in spec tests.** This PR unblocked SSE auth specifically; the broader pattern — "any transport must round-trip the same authentication context as WebSocket" — is exactly what Audit E's transport-contract spec suite would assert. A spec test running the same `check_view_auth` flow against every registered transport would catch any future transport (HTTP polling, WebTransport, etc.) that ships without cookie handling.
- **Same-origin is not always implicit.** I checked the EventSource spec mid-fix; default credentials = "same-origin" (cookies SHOULD be sent for same-origin requests). But explicit beats implicit when downstream consumers report auth failures — the explicit form removes ambiguity across browser versions and proxies. The fetch() POST got the same treatment for the same reason.

**IDEA**: Audit E's Phase 2 work (transport-contract parameterized test suite, `tests/test_transport_contract.py`) should be commissioned in v0.9.3. Two transports (WS, SSE) running through identical contract tests would catch any future drift — not just for auth, but for mount lifecycle, push events, error frames, etc.

Issue #1311, opened 2026-05-02
tech-debt: elevate Action #1200 tautology check to Stage 7 self-review


johnrtipton:
**Source**: Retro v0.9.2-5 / PR #1292 Stage 11 review.

## Pattern

PR #1292's Stage 11 reviewer flagged `TestHandleMountDrainBehavior::test_tail_drains_both_queues_in_order` as a tautology — the test never invoked `handle_mount`, just awaited 3 AsyncMocks in the desired order and asserted that order. Would pass identically if `handle_mount` was deleted.

This is exactly the failure mode Action #1200 (v0.9.4 retro) was canonicalized to catch: "for any 'action happened' assertion, ask 'would this pass if the action didn't run?'"

The check fired correctly at Stage 11 (the canon worked). But Stage 11 is late — the tautology was a cosmetic defect in test scaffolding, not a behavior change. Catching it at Stage 7 self-review would save a Stage-11-driven follow-up commit.

## Why canonicalize at Stage 7

Stage 7 self-review already runs through review-discipline checks. Action #1200's question — "would this pass if the action didn't run?" — is mechanical, applies to every new test, and takes seconds per test. Adding it to Stage 7 means:
- Defects caught one stage earlier
- Stage 11 reviewer time freed for higher-value structural review
- Author self-applies the discipline rather than waiting for a reviewer

## Scope

Update `.pipeline-templates/{feature,bugfix,ship}-state.json` Stage 7 (Self-Review) checklist to add:

```json
{
  "action": "Tautology check (Action #1200): for every new 'action happened' test assertion, ask 'would this pass if the action didn't run?' Tests that mock-and-call without invoking the SUT fail this check.",
  "done": false,
  "mandatory": true
}
```

This is a doc-shape update, ~10 lines across 3 templates. Self-applies on the PR that lands it (per Stage 7 self-applicability check #1248).

johnrtipton:
Resolved by PR #1335 (tautology check added to Stage 7 in all three pipeline templates).

PR #1335, opened 2026-05-02
fix: add tautology check (Action #1200) to Stage 7 self-review (#1311)

johnrtipton:
## Summary
- Adds `Tautology check (Action #1200)` as a mandatory Stage 7 self-review item across all 3 pipeline templates (`bugfix-state.json`, `feature-state.json`, `ship-state.json`)
- The check asks: for every new "action happened" test assertion, "would this pass if the action didn't run?"
- Elevates the discipline from Stage 11 review (where it already fires correctly) to Stage 7 self-review, catching tautology tests one stage earlier

## Test plan
- [x] All three templates parse as valid JSON
- [x] DOCS_ONLY — no code changes

### DIFF
--- a/.pipeline-templates/bugfix-state.json
+++ b/.pipeline-templates/bugfix-state.json
@@ -142,7 +142,7 @@
           "done": false
         },
         {
-          "action": "DO NOT modify CHANGELOG.md in this stage \u2014 defer all [Unreleased] hunks to the Stage 9 docs commit (two-commit shape per v0.9.1 retro / Action Tracker #181 / GitHub #1173)",
+          "action": "DO NOT modify CHANGELOG.md in this stage — defer all [Unreleased] hunks to the Stage 9 docs commit (two-commit shape per v0.9.1 retro / Action Tracker #181 / GitHub #1173)",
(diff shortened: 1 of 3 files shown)


johnrtipton:
## Retrospective — PR #1335

**Quality**: 4/5

### What went well
- Straightforward 3-file edit with a Python script — minimal diff surface, no surprises
- Self-applicability check: the rule being added (tautology check) applies to template changes just like code changes — but since no tests were written (DOCS_ONLY), there were no new test assertions to check, so the rule doesn't false-positive
- The discipline already fires correctly at Stage 11 (#1292 caught it); moving it to Stage 7 is the right tier placement

### Lessons learned
- `json.dump` with `ensure_ascii=False` converts `—` escapes to literal `—` characters — semantically identical but produces a larger diff. For template-only PRs this is harmless noise.
- DOCS_ONLY skip for stages 6-9 + 11-12 continues to work cleanly on template-only changes

### What to improve
- Nothing specific — the tautology check is mechanical and applies to every new test. Adding it to Stage 7 checklist is the right scope.

Issue #1339, opened 2026-05-02
tech-debt: add make check-test-coverage target to verify all test files are collected by CI


johnrtipton:
Source: Retro v0.9.3-4 (PR #1338)

## Problem
PR #1338 moved `test_skip_render_private_state.py` from `python/djust/tests/` to `python/tests/` because the former directory was excluded from CI. The file was in `pyproject.toml`'s testpaths but NOT in `make test-python`'s explicit paths (`python/tests/ tests/`). Tests in `python/djust/tests/` were silently never collected.

## Proposal
Add a `make check-test-coverage` target that greps for `class Test` / `def test_` across all test files and verifies each is collected by CI. This prevents silent coverage gaps where test files exist but are never run.

## Reference
- PR #1338: moved test file into CI-collected path
- PR #1338 retro: "Consider a future make check-test-coverage target"

johnrtipton:
Closed: `scripts/check-test-coverage.py` shipped via PR #1341 and was extended bidirectionally in PR #1398 (v0.9.5-2). Pre-push hook gates on both directions. The original Makefile-overrides-pyproject failure mode is now structurally prevented.

PR #1338, opened 2026-05-02
fix: move test_skip_render_private_state.py to python/tests/ for CI coverage (#1325)

johnrtipton:
## Summary
- Moves `python/djust/tests/test_skip_render_private_state.py` (9 tests, #1281 regression) to `python/tests/`
- `make test-python` and CI workflow use explicit paths `tests/ python/tests/` — the old location `python/djust/tests/` was excluded
- After the move, CI runs these tests on every push

## Test plan
- [x] All 9 tests pass at new location
- [x] Full test suite: 4151 passed, 18 skipped

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1341, opened 2026-05-02
fix: add python/djust/tests/ to CI + check-test-coverage target (#1339)

johnrtipton:
## Summary
- Adds `python/djust/tests/` to `make test-python` (and parallel/background variants) — 2,734 tests across 100+ files were silently excluded from CI
- Adds `make check-test-coverage` target that verifies all test directories are collected, preventing recurrence
- New `scripts/check-test-coverage.py` parses the Makefile to detect gaps

## Why
`make test-python` used explicit pytest paths (`tests/ python/tests/`) which override `pyproject.toml`'s testpaths. `python/djust/tests/` was in testpaths but NOT in the Makefile's explicit paths, so 2,734 tests never ran in CI.

PR #1338 worked around this by moving a test file from `python/djust/tests/` to `python/tests/` — this PR fixes the root cause.

## Test plan
- [x] `make check-test-coverage` exits 0 after fix
- [x] Full test suite passes (6,883 passed, 0 failed)
- [x] `python/djust/tests/` tests now collected and passing in CI path
- [x] `make check-test-coverage` would catch a future directory-removal regression

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- **`python/djust/tests/` now included in `make test-python` + `check-test-coverage` target (#1339).**
+  The Makefile's test targets used explicit pytest paths (`tests/ python/tests/`)
+  which override pyproject.toml's testpaths, silently excluding `python/djust/tests/`
+  (2,734 tests across 100+ files). Added the missing directory to test-python,
+  test-python-parallel, and the background test target. New `make check-test-coverage`
+  target prevents recurrence by verifying every test directory is collected by CI.
+  Verified by `make check-test-coverage` and the 2,734 newly-collected existing tests.
+
 - **@reactive now fails at class-definition time on classes missing ``update()`` (#1287).**
   The ``@reactive`` decorator previously guarded ``self.update()`` with
   ``hasattr(self, 'update')``, silently no-opping when the host class lacked
--- a/Makefile
+++ b/Makefile
@@ -171,12 +171,12 @@ test-rust: ## Run Rust tests
 .PHONY: test-python
 test-python: ## Run Python tests
 	@echo "$(GREEN)Running Python tests...$(NC)"
-	@PYTHONPATH=. .venv/bin/python -m pytest tests/ python/tests/
+	@PYTHONPATH=. .venv/bin/python -m pytest tests/ python/tests/ python/djust/tests/
 
 .PHONY: test-python-parallel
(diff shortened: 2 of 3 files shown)


johnrtipton (commented):
## Review: PR #1341 — fix: add python/djust/tests/ to CI + check-test-coverage target (#1339)

### Plan vs Implementation: PASS

- Part 1: `python/djust/tests/` added to `test-python` (L174), `test-python-parallel` (L179), and `ci-mirror` (L251) in Makefile — all three as planned (ci-mirror is a bonus)
- Part 2: New `make check-test-coverage` target via `scripts/check-test-coverage.py` — cleaner than the inline bash sketched in the plan; parses Makefile's own `test-python` line and verifies all directories containing `test_*.py`/`*_test.py` are covered

### Test results: PASS

- `make test-python`: **6883 passed, 20 skipped, 0 failures** (108s)
- `make check-test-coverage` exits 0
- `python/djust/tests/` contributes 2734 tests (confirmed via `--collect-only`)

### Code quality: PASS (one nit)

- No security hot spot files touched (Makefile + script only)
- No `print()` misuse, no `except: pass`, CHANGELOG updated
- Regex parsing of Makefile is correct and fails gracefully (exit 2 on parse failure)
- **Nit:** `subprocess` imported but unused in `scripts/check-test-coverage.py` (ruff F401)

### Verdict: APPROVE

The unused import can be fixed in a follow-up. The fix is correct, the guardrail target works, and all 2734 previously-missed tests pass.

johnrtipton:
# Retrospective: PR #1341 — check-test-coverage target

**Quality: 4/5**

## What went well

- Two-part fix (add missing directory to CI + add preventive check) was exactly the right scope — fixed the symptom AND prevented recurrence
- Switching from inline bash to a standalone Python script mid-implementation was the right call — the bash escaping was broken and the Python script is more maintainable
- Stage 11 caught the unused `subprocess` import before merge
- `make check-test-coverage` now exits 0, confirming all 3 test directories are covered

## Lessons learned

- Makefile explicit pytest paths silently override pyproject.toml testpaths — 2,734 tests were excluded for months
- CI check targets that parse the Makefile are fragile to format changes (the initial regex didn't handle intermediate echo lines)
- The `check-test-coverage` script's regex approach works but could break if the Makefile format changes — the script is the guardrail but it guards with a fragile parser

## What to improve

- IDEA: extend `check-test-coverage` to verify pyproject.toml testpaths match the Makefile in both directions
- IDEA: add a comment marker in the Makefile that the script keys on instead of regex-matching the pytest line

## Review stats

- Tests: 6,903 collected, 6,883 passed, 20 skipped
- Stage 11 findings: 1 (unused `import subprocess` — fixed)
- CI failures: 0

Issue #1351, opened 2026-05-05
chore: fix 392 pre-existing ESLint warnings in bundled client.js


johnrtipton:
## Summary

The bundled `python/djust/static/djust/client.js` (concatenated from
~52 source modules) trips the eslint pre-commit hook with **392
warnings** (0 errors). The hook is configured with `--max-warnings 0`
(implicitly, via `npx eslint`'s default exit-1-on-warning behavior),
so any commit that touches a `src/*.js` file fails the hook because
`build-js` auto-stages the rebuilt bundle and eslint then runs on it.

```
✖ 392 problems (0 errors, 392 warnings)
  0 errors and 222 warnings potentially fixable with the `--fix` option.
```

## Categories of warnings (most → least common)

- `prefer-const` — `var`/`let` declarations that are never reassigned
- `no-var` — `var` declarations that should be `let` or `const`
- `security/detect-object-injection` — bracket access `obj[someKey]` flagged as a generic injection sink (mostly false positives in our internal data structures, but worth auditing)
- `no-unused-vars` — unused `_e`, `_`, etc. in catch handlers / destructuring
- `security/detect-non-literal-regexp` — `new RegExp(dynamic)` (a few cases)

## Why this matters

Right now any contributor touching `src/*.js` files cannot commit
without `SKIP=build-js,eslint` or `--no-verify` — both of which defeat
the purpose of the hook for genuine new violations.

## Suggested fix

1. Run `npx eslint --fix python/djust/static/djust/client.js` for the
   222 auto-fixable warnings (mostly `prefer-const` / `no-var`)
2. Audit the `security/detect-object-injection` warnings and add
   targeted `// eslint-disable-next-line` comments where the bracket
   access is provably safe (typed indices, controlled keys), OR
   refactor to `Object.hasOwn(obj, key)` patterns where appropriate
3. Fix or `_`-prefix the remaining `no-unused-vars`
4. Once warnings are at 0, leave the hook with `--max-warnings 0` so
   new violations get caught immediately

## Workaround until then

Contributors touching JS source can commit with:
```bash
SKIP=build-js,eslint git commit -m "..."
```

Bundle is then rebuilt by CI (or manually with `make build-js`) and
shipped without eslint blocking. This is a documented pre-commit
mechanism and only skips the two named hooks — all other hooks
(formatting, tests, secret detection, etc.) still run.

PR #1352, opened 2026-05-05
fix(transitions): respect CSS transition-duration; await all transitionend events

johnrtipton:
## Summary

Two bugs in `dj-transition` (`41-dj-transition.js`) and `dj-remove`
(`42-dj-remove.js`) capped animations at 600ms and cut multi-property
transitions short. This PR fixes both at the source-only level.

### 1. Hard-coded 600ms fallback truncated long transitions

```js
const _FALLBACK_MS = 600;
state.fallback = setTimeout(cleanup, _FALLBACK_MS);
```

Any animation longer than 600ms got cut off mid-flight: `cleanup()`
removed the `active` class, dropping the `transition:` rule and
snapping properties to their final values.

### 2. First `transitionend` triggered cleanup

```js
function onEnd(ev) {
    if (ev.target !== el) return;
    cleanup();   // ← fires on FIRST property to finish
}
```

For `transition: opacity 500ms, transform 1500ms`, the opacity event
at 500ms triggered cleanup, snapping the in-flight transform to its
final value at 500ms instead of 1500ms.

## Fix

- Read `transition-property`, `transition-duration`, `transition-delay`
  from `getComputedStyle(el)` after the `active` class lands
- Use `max(duration + delay) + 50ms` as the fallback timeout
- Count expected `transitionend` events from `transition-property`
  and only cleanup after all have fired
- Preserve `dj-remove-duration` attribute as an explicit override
  (still wins, clamped 0–30s)
- Zero-duration / `display:none` cases still hit the 600ms default
  via `_computeTransitionTiming` returning `{maxMs: 0}`

## Source-only — bundle deferred

This PR does **not** include the rebuilt `client.js`/`client.min.js`
bundle. The bundle has 392 pre-existing eslint warnings that block
`--max-warnings 0`, unrelated to this change. Tracking cleanup at #1351.

Bundle should be rebuilt + committed in a follow-up after #1351.
Until then, `make build-js` produces the new bundle locally and
contributors can use `SKIP=build-js,eslint` to commit JS source
changes (named-skip mechanism, not `--no-verify`).

## Test plan

- [x] Source files lint clean (`npx eslint python/djust/static/djust/src/41-dj-transition.js python/djust/static/djust/src/42-dj-remove.js`)
- [x] `make build-js` regenerates bundles successfully (52 modules → 606480 bytes)
- [x] Manually verified at scaffold's `/features-v09/transition/` with 1.8s entry + 1.5s exit + 1.2s delayed layout-collapse — all phases complete cleanly with rebuilt bundle

### DIFF
--- a/python/djust/static/djust/src/41-dj-transition.js
+++ b/python/djust/static/djust/src/41-dj-transition.js
@@ -44,12 +44,55 @@
 //
 // On `transitionend`, phase-2 classes are removed (they typically
 // carry the `transition-*` helper and are not needed once the animation
-// has completed). A 600 ms fallback timeout cleans up phase-2 classes if
-// `transitionend` never fires (e.g. zero-duration transitions or
+// has completed). A computed fallback timeout cleans up phase-2 classes
+// if `transitionend` never fires (e.g. zero-duration transitions or
 // display: none).
+//
+// The fallback duration is auto-derived from the element's computed
+// `transition-duration` + `transition-delay` (longest pair across all
+// transitioning properties) plus a 50ms grace window. For multi-property
+// transitions, we count expected `transitionend` events from
+// `transition-property` and only run cleanup after all have fired —
+// otherwise the first-finishing property would cut off slower ones.
+//
+// `_FALLBACK_MS_DEFAULT` is the fallback used only when computed-style
+// reading fails or yields zero (e.g. element has no transition rule yet).
 
 const _djTransitionState = new WeakMap();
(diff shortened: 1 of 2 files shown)


johnrtipton:
Superseded by #1357 — same dj-transition CSS-duration fix is now bundled with the db.notifications permanent-failure fix on the rebased `fix/notifications-permanent-failure-handling` branch (rebased onto current main with CHANGELOG added). Closing here to consolidate.

Issue #1196, opened 2026-04-28
tech-debt: delegated-listener integration test pattern


johnrtipton:
Source: Retro v0.9.4 / PR #1194

For any "marker class + delegated event listener" feature, unit tests (direct method invocation) and integration tests (real DOM event → registered handler → method) need separate coverage.

**Failure mode**: silent. Method-level tests verify the methods, not the wiring (parseInt, target.closest, branch dispatch order, containment check). Wiring breaks silently in future refactors.

**Observed instance**: PR #1194's first version had 17 method-level vitest cases but ZERO integration tests. Stage 11 caught it; backfill added 6 integration cases (one per click branch + non-tt-button + non-numeric data).

Canonicalize as a CLAUDE.md rule: at least one integration test per delegated-listener selector branch.

johnrtipton:
Shipped in PR #1225 (v0.9.4 retro process canon batch in CLAUDE.md). Comma-list `Closes #...` form in PR body parses only the first ref — the rest had to be closed manually (this confirms #1184's diagnosis once again; the PR-checklist convention says one Closes per body line). 🤖 v0.9.1-6 drain bucket — final issue closure cleanup.

Issue #1049, opened 2026-04-25
tech-debt: Stage 9 test-count recount rule — Stage 9 (Documentation) must re-count tests AFTER Stage 7/12 fix-pass deltas and update the CHANGELOG test-count line before the final docs pass


johnrtipton:
**Source**: Retro v0.7.0 / PR #990
**Action Tracker**: row #127

**Action**: Stage 9 test-count recount rule — Stage 9 (Documentation) must re-count tests AFTER Stage 7/12 fix-pass deltas and update the CHANGELOG test-count line before the final docs pass

**Notes**: PR #990 CHANGELOG claimed "38 total" but actual was 41 (docs author cited Stage 5 count, not post-fix-pass). Stage 11 caught it. Second milestone running with a small CHANGELOG test-count drift — auditable claim, should be precise. Enforcement: Stage 9 checklist grows a "run `make test`, record final count, diff against CHANGELOG" row.

---
_Filed by `/pipeline-retro --reconcile` on 2026-04-25 as part of Scope C: file all Action Tracker rows as GitHub issues._

johnrtipton:
Shipped in PR #1226 (v0.9.1-7 cleanup batch). Comma-list `Closes #...` form parsed only the first ref — the rest closed manually. (This bites again; #1184's parenthesized form is documented in PR-checklist, but the comma-list-on-its-own-line form has the same auto-close failure.) 🤖 v0.9.1-7 cleanup batch closure.

PR #1355, opened 2026-05-05
fix: state-backend safety pair (#1353 + #1354)

### LINKED ISSUE #1353: Concurrent same-session HTTP renders collide on shared RustLiveView (`RuntimeError: Already borrowed`)
## Summary

When two HTTP requests arrive for the same `(session, view_path)` pair while the **in-memory** state backend is in use, both pull the same `RustLiveView` Python object from the cache. The next call to `_sync_state_to_rust` → `self._rust_view.update_state(...)` from one thread panics on `RefCell::borrow_mut` while the other is still inside it. PyO3 surfaces it as:

```
File "djust/mixins/rust_bridge.py", line 553, in _sync_state_to_rust
    self._rust_view.update_state(json_compatible_context)
RuntimeError: Already borrowed
```

User-visible result: 500 page on initial render. Real-world trigger is a single user opening two browser tabs to the same dashboard back-to-back, double-clicking a link, or refresh-spamming.

## Root cause

`djust/mixins/rust_bridge.py:236-345` (`_initialize_rust_view`) caches the `RustLiveView` by `{session_key}_{view_key}` (where `view_key = liveview_<request.path>[_<query_hash>]`). On a cache hit:

```python
cached = backend.get(self._cache_key)
if cached:
    cached_view, timestamp = cached
    self._rust_view = cached_view   # ← shared Python reference
```

The `InMemoryStateBackend` returns the live object reference (`djust/state_backends/memory.py:74`). The HTTP request lifecycle calls `_initialize_rust_view` → `_sync_state_to_rust` on **every request**. Two threads in Daphne's `sync_to_async` pool entering `update_state` concurrently against the same handle is the collision.

WebSocket consumers don't reliably reproduce: `_initialize_rust_view` runs once at mount and the consumer keeps its own reference; subsequent events re-use that reference without re-reading the cache. Per-consumer `_render_lock` + GIL serializes intra-consumer events.

## Reproducer

NYC Claims app on Daphne, in-memory state backend. Session: a logged-in examiner. Twenty trials of two parallel `curl` GETs sharing one `sessionid`:

```bash
SID=...
URL=http://app/claims/dashboard/
for trial in $(seq 1 20); do
  (
    curl -s -o /dev/null -b "sessionid=$SID" -w '%{http_code}\n' "$URL" &
    curl -s -o /dev/null -b "sessionid=$SID" -w '%{http_code}\n' "$URL" &
    wait
  )
done | sort | uniq -c
#   33 200
#    7 500       ← 17.5% error rate from a single user opening 2 tabs
```

Single-session vs. multi-session contrast (100 reqs, c=10):

| Sessions | non-2xx |
|---|---|
| 1 | 21 / 100 (21%) |
| 5 | 3 / 80 (~4%) |

Errors scale with **per-session concurrency**, not total throughput.

WS-only test: 5 concurrent `/ws/live/` consumers sharing one session, each firing 50 `set_*` events — **0 errors**, confirming the bug is on the HTTP render path specifically.

## Server traceback

```
[ERROR] Internal Server Error: /claims/dashboard/
Traceback (most recent call last):
  File "asgiref/sync.py", line 557, in thread_handler
    raise exc_info[1]
  File "django/core/handlers/exception.py", line 42, in inner
    response = await get_response(request)
  ...
  File "django/views/generic/base.py", line 144, in dispatch
    return handler(request, *args, **kwargs)
  ...
  File "djust/mixins/request.py", line 193, in get
    self._sync_state_to_rust()
  File "djust/mixins/rust_bridge.py", line 553, in _sync_state_to_rust
    self._rust_view.update_state(json_compatible_context)
RuntimeError: Already borrowed
```

## Suggested fixes (any of)

1. **Lock around `update_state`** — attach a per-`RustLiveView` `threading.Lock` (or `asyncio.Lock` if always called inside `sync_to_async`) inside `_sync_state_to_rust`. Cheapest fix; serializes the unsafe window without changing the cache contract.
2. **Don't share the cached object across threads** — in `_initialize_rust_view`'s cache-hit branch, return a *clone* of the cached `RustLiveView` (e.g., via `serialize_msgpack` → `deserialize_msgpack` round-trip, like the Redis backend already does, or expose a Rust-side `clone()`).
3. **Bypass the cache for HTTP `GET`s** — only WebSocket needs the cached VDOM baseline for diffing. HTTP renders can always start fresh.

(1) is least invasive; (2) is most consistent with how the Redis backend already behaves; (3) is most aggressive but eliminates the contention class entirely.

## Environment

- djust `0.9.2rc1`
- Django 5.x + Daphne (single-process ASGI)
- `STATE_BACKEND = "memory"` (the project intended Redis but configured it via `DJUST_STATE_BACKEND` settings module attr — `BackendRegistry` reads `DJUST_CONFIG["STATE_BACKEND"]` which was unset, silently falling back to in-memory; that's a separate documentation/config-validation issue worth filing if it isn't tracked)

🤖 Filed from NYC Claims prototype — discovered while running `ab` against the AWS deploy at concurrency 10–20.

### LINKED ISSUE #1354: State backend silently falls back to in-memory when configured via top-level Django settings (`DJUST_STATE_BACKEND` / `DJUST_REDIS_URL`)
## Summary

The `BackendRegistry` for state only reads from `DJUST_CONFIG["STATE_BACKEND"]` in Django settings. Projects that configure the state backend via top-level settings — `DJUST_STATE_BACKEND = "redis://..."`, `DJUST_REDIS_URL = "..."` — get **silently downgraded to the in-memory backend** with no warning.

This is easy to miss because:
- Both names look canonical (the `DJUST_` prefix matches every other djust setting in the project).
- The startup log shows `Initialized state backend: memory`, but in a busy app log it's just one line.
- All single-request flows work fine. The misconfig only surfaces under multi-process / multi-replica / concurrent load — exactly the conditions that motivate using Redis in the first place.

We hit this on production (NYC Claims, AWS Fargate). The ECS task definition has `DJUST_STATE_BACKEND=redis://jt-redis...:6379/0` set, the project's `nyc_claims/settings/base.py` reads it into `DJUST_STATE_BACKEND` as a top-level setting, ElastiCache is provisioned and reachable — but `BackendRegistry.get()` returns the in-memory backend because it looks for `DJUST_CONFIG["STATE_BACKEND"]`, which we never set. The downstream consequence was the borrow collision filed in #1353 (`InMemoryStateBackend.get()` returns shared object references; `RedisStateBackend.get()` deserializes a fresh copy each time, which would have masked that bug).

## Root cause

`djust/state_backends/registry.py:40-45`:

```python
_registry = BackendRegistry(
    config_key=\"STATE_BACKEND\",
    default_type=\"memory\",
    factory=_create_state_backend,
    name=\"state\",
)
```

`BackendRegistry.get()` (`djust/utils.py:90-103`) reads `DJUST_CONFIG.get(\"STATE_BACKEND\", \"memory\")`. The top-level Django settings `DJUST_STATE_BACKEND` and `DJUST_REDIS_URL` are never consulted.

The docstring on `state_backends/registry.py:55-65` only mentions the `DJUST_CONFIG` path:

```python
Configuration in settings.py:
    DJUST_CONFIG = {
        'STATE_BACKEND': 'redis',
        'REDIS_URL': 'redis://localhost:6379/0',
        ...
    }
```

…but several places in the docs and djust-scaffold templates use the top-level form. Mismatch.

## Suggested fixes (in order of preference)

1. **Read both forms.** In the registry factory, fall back to top-level `settings.DJUST_STATE_BACKEND` / `settings.DJUST_REDIS_URL` when the corresponding `DJUST_CONFIG` keys are absent. Treat URL-shaped values (`redis://...`) as `backend_type=\"redis\"` plus `REDIS_URL=<value>`. Backwards-compatible.
2. **Warn when falling back to in-memory in non-DEBUG mode.** A single `logger.warning(\"Falling back to in-memory state backend in production — multi-process deployments will lose state across replicas\")` would have caught this in 5 minutes instead of taking us to a load test.
3. **Update the docstring + scaffold template** to consistently use one form. Currently the wild docs reference the top-level names but the registry only honors the dict form.

(1) and (2) are independently useful and shouldn't conflict.

## Reproducer

```python
# settings.py — looks correct, silently isn't
DJUST_STATE_BACKEND = \"redis://localhost:6379/0\"
DJUST_REDIS_URL = \"redis://localhost:6379/0\"
# (no DJUST_CONFIG defined)
```

Then:

```python
from djust.state_backends import get_backend
print(type(get_backend()).__name__)
# → InMemoryStateBackend
```

Expected: `RedisStateBackend`. Actual: `InMemoryStateBackend`, no warning.

## Environment

- djust `0.9.2rc1`
- Django 5.x

🤖 Filed from NYC Claims prototype, alongside #1353. Discovered when investigating why the in-memory borrow bug was reaching production despite Redis being provisioned.

johnrtipton:
## Summary

Closes both halves of the production failure mode reported by NYC Claims:
- #1354: state backend silently falls back to in-memory when configured via top-level Django settings (the misconfig)
- #1353: in-memory backend's shared-reference cache panics on concurrent same-session HTTP renders (the resulting crash)

Either fix alone leaves the production failure mode intact: the misconfig silently downgrades to in-memory, then in-memory shared-reference panics on concurrent renders.

### #1354 — top-level settings honoured + production warning

`python/djust/utils.py:BackendRegistry` (lines ~58-200) gains a `top_level_aliases` constructor arg mapping `DJUST_<NAME>` settings to `DJUST_CONFIG` keys. `_merge_top_level_aliases()` layers top-level settings on top of the config dict (only fills keys absent from `DJUST_CONFIG` — backwards-compatible). URL-shaped values for the primary `config_key` (`redis://`, `rediss://`) are auto-translated to `(backend_type="redis", REDIS_URL=<url>)`.

`python/djust/state_backends/registry.py` (lines 40-58) wires `DJUST_STATE_BACKEND` and `DJUST_REDIS_URL` as aliases for `STATE_BACKEND` / `REDIS_URL` in `DJUST_CONFIG`.

When `DEBUG=False` and the resolved backend is the registry's `default_type` (typically `"memory"`), `BackendRegistry.get` emits a `logger.warning` flagging the production misconfig — multi-process deployments lose state across replicas.

### #1353 — per-RustLiveView lock around the unsafe window

Chose **option 1** (per-`RustLiveView` `threading.Lock`) per the issue body's recommendation. Cheapest, no contract change, mirrors the existing `_render_lock` pattern used on the WS path (`websocket.py:419`).

`python/djust/mixins/rust_bridge.py` (lines 17-71): added a module-level `_RUST_VIEW_LOCKS: dict[int, threading.Lock]` keyed by `id(rust_view)`. Used `id()` rather than attribute or weakref because `RustLiveView` is a Rust-extension type that rejects both `setattr` and `weakref` (verified). ID reuse after GC is harmless: two unrelated views temporarily sharing a lock is just an irrelevant tiny serialization, never a correctness bug. Memory: ~56 bytes per cached view, bounded by backend cache size.

`python/djust/mixins/rust_bridge.py` (lines 678-758) restructures `_sync_state_to_rust` so the four `borrow_mut`-triggering Rust calls (`update_state`, `mark_safe_keys`, `set_raw_py_values`, `set_changed_keys`) run inside the lock. Sidecar build and `user_changed` selection moved outside the lock since they only read `full_context` — the lock window stays tight.

## Test plan

- [x] New: `python/tests/test_state_backend_config.py` — 7 cases covering URL-shaped top-level setting, two-setting `DJUST_STATE_BACKEND`+`DJUST_REDIS_URL` pair, production fallback warning, DEBUG=True suppression, `DJUST_CONFIG` regression guards (memory/redis), `DJUST_CONFIG`-wins-over-top-level priority.
- [x] New: `python/tests/test_rust_bridge_concurrent.py` — 5 cases covering concurrent two-thread no-borrow-error reproducer (200 iterations × 2 threads), lock-releases-after-call, distinct views get distinct locks, same view returns same lock, lock dict module-level invariant.
- [x] All 12 new tests pass.
- [x] Wider sanity: `pytest python/tests/ tests/ --ignore=tests/playwright --ignore=tests/js` passes 4164 tests in 90s, 0 regressions.
- [x] Two-commit shape (Action #181): impl+tests in `97e3e75b`, CHANGELOG-only in `9a22688e`.
- [x] Pre-push hooks all green (ruff, ruff-format, bandit, pytest, cargo, dead-method, handler-contracts, changelog-test-counts).

Closes #1353, closes #1354.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,44 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- **In-memory state backend no longer panics on concurrent same-session
+  HTTP renders (#1353).** When two HTTP requests for the same
+  ``(session, view_path)`` pair share a cached ``RustLiveView`` (the
+  in-memory backend returns the same Python reference on cache hits),
+  concurrent calls to ``_sync_state_to_rust`` would race inside Rust's
+  ``RefCell::borrow_mut`` and surface as ``RuntimeError: Already
+  borrowed`` (NYC Claims observed 17.5% 500-rate at concurrency 2).
+  Fixed by wrapping the unsafe window in
+  ``python/djust/mixins/rust_bridge.py:_sync_state_to_rust`` with a
+  per-``RustLiveView`` ``threading.Lock`` (option 1 of three suggested
+  in the issue body — chosen for being cheapest, requiring no cache
+  contract change, and mirroring the existing ``_render_lock`` pattern
+  on the WS path). Lock keyed by ``id(rust_view)`` in a module-level
+  dict because ``RustLiveView`` rejects both ``setattr`` and
+  ``weakref``. New regression cases in
+  ``TestConcurrentSyncStateToRust`` in
+  ``python/tests/test_rust_bridge_concurrent.py``.
+
+- **State backend honours top-level ``DJUST_STATE_BACKEND`` /
+  ``DJUST_REDIS_URL`` settings (#1354).** Previously
+  ``BackendRegistry`` only consulted ``DJUST_CONFIG["STATE_BACKEND"]``,
+  so projects configuring via top-level Django settings (e.g.
+  ``DJUST_STATE_BACKEND = "redis://localhost:6379/0"``) were silently
+  downgraded to in-memory with no warning. Now the registry layers
+  top-level aliases on top of ``DJUST_CONFIG`` (``DJUST_CONFIG`` still
+  wins when both are set — backwards-compatible). URL-shaped values
+  (``redis://``, ``rediss://``) are auto-translated to
+  ``backend_type="redis"`` plus ``REDIS_URL=<url>``. When
+  ``DEBUG=False`` and the resolved backend is the default (in-memory),
+  ``djust.utils.BackendRegistry.get`` now emits a ``logger.warning``
+  flagging the production misconfig — multi-process deployments lose
+  state across replicas. New regression cases in
+  ``TestTopLevelStateBackendSetting`` and
+  ``TestDjustConfigRegression`` in
+  ``python/tests/test_state_backend_config.py``.
+
 ## [0.9.3rc2] - 2026-05-04
 
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,20 +11,30 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - **In-memory state backend no longer panics on concurrent same-session
   HTTP renders (#1353).** When two HTTP requests for the same
-  ``(session, view_path)`` pair share a cached ``RustLiveView`` (the
-  in-memory backend returns the same Python reference on cache hits),
-  concurrent calls to ``_sync_state_to_rust`` would race inside Rust's
-  ``RefCell::borrow_mut`` and surface as ``RuntimeError: Already
-  borrowed`` (NYC Claims observed 17.5% 500-rate at concurrency 2).
-  Fixed by wrapping the unsafe window in
-  ``python/djust/mixins/rust_bridge.py:_sync_state_to_rust`` with a
-  per-``RustLiveView`` ``threading.Lock`` (option 1 of three suggested
-  in the issue body — chosen for being cheapest, requiring no cache
-  contract change, and mirroring the existing ``_render_lock`` pattern
-  on the WS path). Lock keyed by ``id(rust_view)`` in a module-level
-  dict because ``RustLiveView`` rejects both ``setattr`` and
-  ``weakref``. New regression cases in
-  ``TestConcurrentSyncStateToRust`` in
+  ``(session, view_path)`` pair shared a cached ``RustLiveView`` (the
+  in-memory backend returned the same Python reference on cache hits),
+  concurrent ``&mut self`` Rust methods on the shared view would
+  collide inside Rust's ``RefCell::borrow_mut`` and surface as
+  ``RuntimeError: Already borrowed`` (NYC Claims observed 17.5%
+  500-rate at concurrency 2). The race spanned more than the
+  ``_sync_state_to_rust`` mutation calls — ``render()`` itself holds
+  ``&mut self`` across template evaluation, and
+  ``Context::resolve_dotted_via_getattr``
+  (``crates/djust_core/src/context.rs``) wraps ``Python::with_gil`` so
+  the embedded ``getattr`` can yield the GIL inside an active mutable
+  borrow. Any peer thread entering an ``&mut self`` method during that
+  window panicked. Fixed by switching ``InMemoryStateBackend.get()`` to
+  return an isolated ``serialize_msgpack`` / ``deserialize_msgpack``
+  clone of the cached view (option 2 of three suggested in the issue
+  body), mirroring the ``RedisStateBackend`` contract — which already
+  deserialized fresh on every read. With each caller holding its own
+  ``RustLiveView`` instance, no two threads can share a Rust ``&mut
+  self`` borrow and the race class is eliminated at the source. No
+  Python-side lock is needed. New regression cases in
+  ``TestInMemoryGetReturnsIsolatedView`` (4 cases — clone identity,
+  state preservation, mutation isolation, concurrent get) and
+  ``TestConcurrentRenderNoBorrowError`` (2 cases — concurrent render
+  with GIL-yielding sidecar, concurrent update_state) in
   ``python/tests/test_rust_bridge_concurrent.py``.
 
 - **State backend honours top-level ``DJUST_STATE_BACKEND`` /
@@ -35,14 +45,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   downgraded to in-memory with no warning. Now the registry layers
   top-level aliases on top of ``DJUST_CONFIG`` (``DJUST_CONFIG`` still
   wins when both are set — backwards-compatible). URL-shaped values
-  (``redis://``, ``rediss://``) are auto-translated to
-  ``backend_type="redis"`` plus ``REDIS_URL=<url>``. When
-  ``DEBUG=False`` and the resolved backend is the default (in-memory),
-  ``djust.utils.BackendRegistry.get`` now emits a ``logger.warning``
-  flagging the production misconfig — multi-process deployments lose
-  state across replicas. New regression cases in
-  ``TestTopLevelStateBackendSetting`` and
-  ``TestDjustConfigRegression`` in
+  (``redis://``, ``rediss://``, ``redis+sentinel://``) are
+  auto-translated to ``backend_type="redis"`` plus ``REDIS_URL=<url>``;
+  the prefix list lives in ``BackendRegistry._REDIS_URL_PREFIXES``.
+  When ``DEBUG=False`` and the resolved backend is the default
+  (in-memory), ``djust.utils.BackendRegistry.get`` now emits a
+  ``logger.warning`` flagging the production misconfig — multi-process
+  deployments lose state across replicas. ``unix://`` URLs are left as
+  a TODO follow-up because the underlying ``redis-py`` client takes
+  Unix sockets via a different parameter name. New regression cases in
+  ``TestTopLevelStateBackendSetting`` (6 cases) and
+  ``TestDjustConfigRegression`` (3 cases) in
   ``python/tests/test_state_backend_config.py``.
 
 ## [0.9.3rc2] - 2026-05-04
--- a/python/djust/mixins/rust_bridge.py
+++ b/python/djust/mixins/rust_bridge.py
@@ -4,7 +4,6 @@ RustBridgeMixin - Rust backend integration for LiveView.
 
 import hashlib
 import logging
-import threading
 from typing import Any, List, Optional, Set
 from urllib.parse import parse_qs, urlencode
 
@@ -15,42 +14,25 @@ from ..utils import get_template_dirs
 logger = logging.getLogger(__name__)
 
 
-# Per-RustLiveView mutation locks (#1353). The in-memory state backend
-# returns the same Python object reference on cache hits, so two
-# concurrent HTTP renders for the same (session, view_path) pair both
-# land on the same ``RustLiveView`` and race inside Rust's ``RefCell::
-# borrow_mut`` — surfacing as ``RuntimeError: Already borrowed``. Wrapping
-# the ``update_state`` / ``mark_safe_keys`` / ``set_raw_py_values`` /
-# ``set_changed_keys`` window in a Python ``threading.Lock`` serializes
-# the unsafe window without changing the cache contract.
+# #1353: Concurrent same-session HTTP renders previously panicked with
+# ``RuntimeError: Already borrowed`` because the in-memory state backend
+# returned the SAME Python object on cache hits, so two threads both
+# called ``&mut self`` Rust methods (``update_state``, ``render``,
+# ``set_template_dirs``, etc.) on the shared ``RustLiveView`` and
+# collided inside Rust's ``RefCell::borrow_mut``. The race window
+# spanned more than the ``_sync_state_to_rust`` mutation calls:
+# ``render()`` itself holds ``&mut self`` across template evaluation
+# that yields the GIL via ``Context::resolve_dotted_via_getattr``
+# (``crates/djust_core/src/context.rs``), so a peer thread entering any
+# ``&mut self`` method during that window panicked.
 #
-# Implementation note: ``RustLiveView`` is a Rust-extension type that
-# rejects ``setattr`` and weak references, so we can't attach the lock
-# to the view directly. We key the lock dict by ``id(rust_view)`` —
-# stable for the lifetime of the cached object, and the same id for both
-# colliding threads (cache hit returns the same ref). ``id()`` reuse
-# after GC is harmless: two unrelated views temporarily sharing a lock
-# is just an irrelevant tiny serialization, never a correctness bug.
-# Memory: ~56 bytes per cached view, bounded by backend cache size.
-_RUST_VIEW_LOCKS: "dict[int, threading.Lock]" = {}
-_RUST_VIEW_LOCKS_GUARD = threading.Lock()
-
-
-def _get_rust_view_lock(rust_view) -> threading.Lock:
-    """Return (creating if needed) the per-``RustLiveView`` mutation lock.
-
-    Used to serialize concurrent ``_sync_state_to_rust`` calls against a
-    shared cached ``RustLiveView`` (#1353). The dict guard is held only
-    long enough to read/insert the lock entry; the per-view lock itself
-    is acquired separately by callers.
-    """
-    rid = id(rust_view)
-    with _RUST_VIEW_LOCKS_GUARD:
-        lock = _RUST_VIEW_LOCKS.get(rid)
-        if lock is None:
-            lock = threading.Lock()
-            _RUST_VIEW_LOCKS[rid] = lock
-    return lock
+# Resolution: ``InMemoryStateBackend.get()`` now returns an isolated
+# clone of the cached view (``serialize_msgpack`` →
+# ``deserialize_msgpack``) — mirroring how ``RedisStateBackend.get``
+# already behaves. With each caller holding its own ``RustLiveView``
+# instance, no two threads can share a Rust ``&mut self`` borrow and
+# the race class is eliminated at the source. No Python-side lock is
+# needed anymore.
 
 
 # Keys excluded from set_changed_keys — these are framework-internal values
@@ -676,17 +658,15 @@ class RustBridgeMixin:
             else:
                 json_compatible_context = rendered_context
 
-            # Serialize the Rust mutation window (#1353). Concurrent HTTP
-            # renders that share a cached ``RustLiveView`` (in-memory backend
-            # returns the same Python ref) would otherwise race inside
-            # ``RefCell::borrow_mut`` and raise ``RuntimeError: Already
-            # borrowed``. The lock is per-``RustLiveView`` and held only
-            # around the ``borrow_mut`` calls below — never around context
-            # computation, normalization, or template rendering.
-            _rust_lock = _get_rust_view_lock(self._rust_view)
+            # No Python-side lock needed (#1353): each HTTP/WebSocket
+            # caller now holds its own ``RustLiveView`` instance because
+            # ``InMemoryStateBackend.get`` returns a fresh
+            # ``serialize_msgpack`` / ``deserialize_msgpack`` clone on
+            # cache hits, mirroring the ``RedisStateBackend`` contract.
+            # See the module-level docstring above for context.
 
-            # Build the sidecar OUTSIDE the lock — it only reads
-            # ``full_context``, never touches Rust state.
+            # Build the sidecar of raw Python objects — reads
+            # ``full_context`` only, never touches Rust state.
             sidecar = None
             if hasattr(self._rust_view, "set_raw_py_values"):
                 _JSON_FRIENDLY = (
@@ -739,24 +719,23 @@ class RustBridgeMixin:
                 if _candidate:
                     user_changed = _candidate
 
-            with _rust_lock:
-                self._rust_view.update_state(json_compatible_context)
-                if safe_keys:
-                    self._rust_view.mark_safe_keys(safe_keys)
+            self._rust_view.update_state(json_compatible_context)
+            if safe_keys:
+                self._rust_view.mark_safe_keys(safe_keys)
 
-                # Always call set_raw_py_values (even when empty) so stale
-                # objects from a previous render are cleared.
-                if sidecar is not None:
-                    try:
-                        self._rust_view.set_raw_py_values(sidecar)
-                    except Exception:
-                        logging.getLogger("djust.rust_bridge").warning(
-                            "set_raw_py_values failed; template getattr fallback disabled this cycle",
-                            exc_info=True,
-                        )
+            # Always call set_raw_py_values (even when empty) so stale
+            # objects from a previous render are cleared.
+            if sidecar is not None:
+                try:
+                    self._rust_view.set_raw_py_values(sidecar)
+                except Exception:
+                    logging.getLogger("djust.rust_bridge").warning(
+                        "set_raw_py_values failed; template getattr fallback disabled this cycle",
+                        exc_info=True,
+                    )
 
-                if user_changed is not None:
-                    self._rust_view.set_changed_keys(user_changed)
+            if user_changed is not None:
+                self._rust_view.set_changed_keys(user_changed)
 
             # Mark static assigns as sent — subsequent syncs will skip them
             if getattr(self, "static_assigns", None) and not getattr(
--- a/python/djust/mixins/rust_bridge.py
+++ b/python/djust/mixins/rust_bridge.py
@@ -4,6 +4,7 @@ RustBridgeMixin - Rust backend integration for LiveView.
 
 import hashlib
 import logging
+import threading
 from typing import Any, List, Optional, Set
 from urllib.parse import parse_qs, urlencode
 
@@ -12,6 +13,45 @@ from ..serialization import normalize_django_value
 from ..utils import get_template_dirs
 
 logger = logging.getLogger(__name__)
+
+
+# Per-RustLiveView mutation locks (#1353). The in-memory state backend
+# returns the same Python object reference on cache hits, so two
+# concurrent HTTP renders for the same (session, view_path) pair both
+# land on the same ``RustLiveView`` and race inside Rust's ``RefCell::
+# borrow_mut`` — surfacing as ``RuntimeError: Already borrowed``. Wrapping
+# the ``update_state`` / ``mark_safe_keys`` / ``set_raw_py_values`` /
+# ``set_changed_keys`` window in a Python ``threading.Lock`` serializes
+# the unsafe window without changing the cache contract.
+#
+# Implementation note: ``RustLiveView`` is a Rust-extension type that
+# rejects ``setattr`` and weak references, so we can't attach the lock
+# to the view directly. We key the lock dict by ``id(rust_view)`` —
+# stable for the lifetime of the cached object, and the same id for both
+# colliding threads (cache hit returns the same ref). ``id()`` reuse
+# after GC is harmless: two unrelated views temporarily sharing a lock
+# is just an irrelevant tiny serialization, never a correctness bug.
+# Memory: ~56 bytes per cached view, bounded by backend cache size.
+_RUST_VIEW_LOCKS: "dict[int, threading.Lock]" = {}
+_RUST_VIEW_LOCKS_GUARD = threading.Lock()
+
+
+def _get_rust_view_lock(rust_view) -> threading.Lock:
+    """Return (creating if needed) the per-``RustLiveView`` mutation lock.
+
+    Used to serialize concurrent ``_sync_state_to_rust`` calls against a
+    shared cached ``RustLiveView`` (#1353). The dict guard is held only
+    long enough to read/insert the lock entry; the per-view lock itself
+    is acquired separately by callers.
+    """
+    rid = id(rust_view)
+    with _RUST_VIEW_LOCKS_GUARD:
+        lock = _RUST_VIEW_LOCKS.get(rid)
+        if lock is None:
+            lock = threading.Lock()
+            _RUST_VIEW_LOCKS[rid] = lock
+    return lock
+
 
 # Keys excluded from set_changed_keys — these are framework-internal values
 # that change id() every render but don't affect template output.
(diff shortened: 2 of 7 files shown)


johnrtipton (commented):
# Stage 11 Code Review: PR #1355 (state-backend safety pair)

## Verdict

REQUEST_CHANGES — #1354 (top-level alias + warning) is sound and shippable. #1353 (per-view lock) protects the wrong window: the lock covers `_sync_state_to_rust`'s mutation calls but leaves `_rust_view.render()`, `render_with_diff()`, `update_template()`, `set_template_dirs()`, and `reset()` outside it. The PyO3 `Already borrowed` race is between any two `&mut self` methods, and the `template.py:213` `render()` call sits outside the lock, holding `&mut self` across template evaluation that releases the GIL via `Python::with_gil` callbacks (`crates/djust_core/src/context.rs:244`) — exactly the window the production trace points at. The synthetic `update_state`-only repro in `test_two_threads_no_borrow_error` does not exercise the actual race class, which is also a tautology concern.

## Must-fix findings

1. **Lock window is incomplete — leaves `render()` / `render_with_diff()` / `update_template()` / `set_template_dirs()` / `reset()` unprotected.** In `python/djust/mixins/template.py:213` the render path is `_initialize_rust_view → _sync_state_to_rust → _rust_view.render()`. The lock in `rust_bridge.py:742` is released when `_sync_state_to_rust` returns; thread A then enters `_rust_view.render()` (Rust `&mut self` at `crates/djust_live/src/lib.rs:262`). `render()` calls `template_arc.render_with_loader` which dispatches to `Context::resolve_dotted_via_getattr` (`crates/djust_core/src/context.rs:244`) — that wraps `Python::with_gil` and the embedded `getattr` calls release the GIL between bytecodes. Thread B (still holding the lock) calls `_rust_view.update_state()` → tries to `borrow_mut()` → PANIC `Already borrowed` because A's `render()` mutable borrow is still alive on the call stack. The same applies between two `render()` callers, between `render()` and `update_template()` (line 896), between `set_template_dirs()` (rust_bridge.py:334, 361) and any `&mut self` method on the cached view's `_initialize_rust_view` cache-hit branch. Either widen the lock to cover the entire `render` / `render_with_diff` cycle in `template.py`, or pick option 2 from the issue body (clone the cached `RustLiveView` on cache hit) which eliminates the race class entirely. The current shape will leave the original 17.5% 500-rate intact under load: lock contention will reshuffle which method call panics, not whether one panics.

2. **`test_two_threads_no_borrow_error` is a tautology and does not exercise the production race class (Action #1200).** `python/tests/test_rust_bridge_concurrent.py:38-42` defines `_MinimalRustBridge.sync()` to call `update_state` *inside* `_get_rust_view_lock`. The test would pass even if `_sync_state_to_rust` in production code had no lock at all — because the test's serialization is entirely embedded in `_MinimalRustBridge`, not in production code under test. Worse: I ran the same `update_state`-only race against the real `RustLiveView` with no lock at concurrency 2, 4 and 8 with 5,000 iterations per thread and got 0 errors. The synthetic loop does not reproduce the original `Already borrowed` symptom — confirming what the implementer's PR body conceded ("the synthetic repro DOES NOT trigger"). To be a real regression test, the worker must call a method that re-enters Python and yields the GIL (e.g., `view.render()` against a sidecar with a `getattr` fallback that calls `time.sleep(0)`), and the assertion must be that the production code path (not a test-local lock) prevents the panic.

## Should-fix findings

1. **`_RUST_VIEW_LOCKS` grows unbounded over process lifetime.** The implementer comment at `rust_bridge.py:81-83` claims "Memory: ~56 bytes per cached view, bounded by backend cache size." This is incorrect: locks are keyed by `id()` and never removed. When a `RustLiveView` is freed (session expires, cache evicts), CPython may or may not recycle its address. Across many sessions (e.g., a long-running production process serving thousands of users), the dict accumulates entries for every `RustLiveView` ever cached. A long-lived deployment with high session churn — exactly the deployment shape this fix targets — will leak unbounded `threading.Lock` objects (~56B + dict overhead per entry; thousands of entries per day on a busy app). Mitigations: (a) reap entries when `_initialize_rust_view` cache-evicts a key, (b) bound the dict via `OrderedDict` LRU. The implementer's note ("`id()` reuse after GC is harmless") addresses the correctness side but not the leak side.

2. **`override_settings(spec=[])` at `test_state_backend_config.py:91` is a no-op.** The comment says "Make sure no top-level alias is set" but `spec` is not a recognized Django setting that this code reads — the call literally just sets a Django setting named `spec` to `[]`. The test currently passes because the test-class doesn't have a `DJUST_STATE_BACKEND` set in its parent scope, not because of `spec=[]`. Either remove the call (it's misleading cargo-cult) or replace with the correct mechanism (delete the setting via `del settings.DJUST_STATE_BACKEND` inside an explicit `override_settings(DJUST_STATE_BACKEND=None)` — though `None` is short-circuited at line 126-127, so that works).

3. **URL-shape detection misses common production schemes.** `python/djust/utils.py:133` only matches `redis://` and `rediss://`. Real-world configs that this misses:
   - `redis+sentinel://` (Sentinel HA, common in production) — silently sets `STATE_BACKEND="redis+sentinel://..."`, factory will fail with "Unknown backend type".
   - `unix:///path/to/redis.sock` — same failure.
   - `redis:/host/0` (typo, single slash) — same failure, but with a misleading error.
   Either widen the prefix list or use `urlparse(value).scheme in {"redis", "rediss", "redis+sentinel"}` and let the factory raise a clear "unsupported scheme" error. This is forgivable as an incremental shipment, but worth a `# TODO` plus a follow-up issue.

## Strong points

1. **Two-commit shape correctly observed (Action #181).** Commit `97e3e75b` is impl + tests with no CHANGELOG; commit `9a22688e` is CHANGELOG only. Verified via `git show <sha> --name-only`.

2. **#1354 fix is sound.** Backwards-compatible (DJUST_CONFIG wins, line `utils.py:121`), top-level alias merge is well-isolated in `_merge_top_level_aliases`, the warning is gated on `DEBUG=False AND backend == default_type` (line 165) which is the right shape, and the warning message names both fix paths. The 7 tests cover the contract reasonably.

3. **`test_djust_config_wins_over_top_level` is a real precedence test** — captures the URL passed to `redis.from_url` and asserts it's the DJUST_CONFIG one, not the top-level one. Strong behavioral assertion (`test_state_backend_config.py:159-161`).

4. **Sidecar build moved outside the lock.** `rust_bridge.py:688-715` builds the sidecar dict outside the lock, which is correct — sidecar reads `full_context` (Python-side) and never touches `_rust_view` state.

5. **Lock acquire/release safety.** `with _rust_lock:` (line 742) ensures the lock is released on exception. The dict-guard pattern in `_get_rust_view_lock` is a textbook double-checked-with-guard idiom.

6. **No security rule violations.** Grepped for `mark_safe(f"`, f-string logging, bare `except: pass`, `print()` — all clean. The two `mark_safe_keys` references are the Rust method, not Django's `mark_safe`. Logging uses `%s` style.

## Specific answers

### #1354
1. **Backwards compatibility**: PASS. `python/djust/utils.py:121` (`if config_key in merged: continue # DJUST_CONFIG wins`) ensures any pre-existing `DJUST_CONFIG["STATE_BACKEND"]` is untouched. Verified by `test_djust_config_memory_backend_works` and `test_djust_config_redis_backend_works`.
2. **Precedence**: DJUST_CONFIG wins, which is the right call — preserves existing app contracts. Asserted by `test_djust_config_wins_over_top_level` (captures `redis.from_url` URL and asserts the DJUST_CONFIG URL is used).
3. **URL parsing robustness**: weak — see Should-fix #3. Only `redis://` / `rediss://` matched. `redis+sentinel://`, `unix://`, and typos like `redis:/host/0` would be merged as the `STATE_BACKEND` value verbatim and fail downstream with a less clear error.
4. **Warning condition**: appropriate. Single-process low-traffic deploys that legitimately use in-memory will still see the warning, but the warning is informational not fatal — this is a reasonable tradeoff. The user can suppress via Python `logging` filters if needed. The trigger condition `not _debug and backend_type == self._default_type` (line 165) is correct.
5. **Test coverage**: 7 tests cover (a) top-level URL → Redis, (b) top-level "redis" + DJUST_REDIS_URL → Redis, (c) DJUST_CONFIG wins over top-level, (d) production warning fires, (e) DEBUG=True suppresses warning, (f) DJUST_CONFIG memory works, (g) DJUST_CONFIG redis works. Missing: explicit "top-level memory" test (low value, covered by default) and explicit "URL extracted to REDIS_URL" assertion (covered indirectly via `from_url` capture in test (c)).

### #1353
1. **Lock scope / memory leak**: real concern. See Should-fix #1.
2. **Lock acquire/release safety**: PASS. `with _rust_lock:` at line 742 handles exceptions. `_get_rust_view_lock` itself is correct.
3. **Lock window completeness**: FAIL. See Must-fix #1. Unprotected `&mut self` calls outside the lock: `template.py:213` (`render()`), `template.py:896` (`update_template`), `template.py:911` (`render_with_diff`), `rust_bridge.py:334` and `rust_bridge.py:361` (`set_template_dirs` on cache hit), and downstream `_rust_view.reset()` at `runtime.py:791`, `sse.py:465`, `websocket.py:3310`, `mixins/request.py:675`. Most of those are WebSocket-only paths (already serialized by `_render_lock`), but `template.py` and `rust_bridge.py` are HTTP-path and unprotected.
4. **GIL-release path / lock necessity**: I ran a synthetic stress test (4 threads × 5000 iterations × {update_state, render}) against an unlocked `RustLiveView` and got zero panics. The actual production race requires (a) a sidecar `getattr` fallback re-entering Python during `render()`, which yields the GIL inside the `&mut self` borrow, and (b) another thread entering an `&mut self` method during that yield window. The lock is necessary in production but the implementer's lock window doesn't cover the actual race — `render()` IS where the GIL gets yielded (`Context::resolve_dotted_via_getattr` at `context.rs:244`), and `render()` is OUTSIDE the lock.
5. **Test design / actual concurrency reproduction**: FAIL — see Must-fix #2. The 5 tests verify lock placement (good), but the headline test `test_two_threads_no_borrow_error` is a tautology: it embeds the lock inside the test's `_MinimalRustBridge.sync()` method, so the test passes regardless of whether `_sync_state_to_rust` in production code is locked. To pass the "Action #1196 — would this test FAIL on main without the fix?" check, the worker must call into an unlocked production path (i.e., `view._sync_state_to_rust()` directly, not a test-local wrapper) AND must induce a GIL yield via a sidecar `getattr` callback (e.g., a Python object whose `__getattr__` calls `time.sleep(0)`).
6. **Sidecar build outside lock**: PASS. The sidecar build (lines 688-715) reads only `full_context` (a Python dict) and `self._rust_view` (only `hasattr` check, no method calls). No shared state mutated outside the lock that the lock-protected window also touches.

### General
1. **Security rules**: PASS. No `mark_safe(f"...")`, f-string logging, bare `except:`, `print()`, or unguarded `console.log`. The `mark_safe_keys` is a Rust method on `RustLiveView`, not Django's `mark_safe`.
2. **Two-commit shape (Action #181)**: PASS. Commit `97e3e75b` = impl + tests; commit `9a22688e` = CHANGELOG only. Verified via `git show --name-only`.
3. **Test discipline (Action #1196 — would FAIL on main without the fix)**: PARTIAL FAIL. #1354 tests would fail without the fix (`test_url_shaped_top_level_resolves_to_redis_backend` and `test_top_level_pair_resolves_to_redis_backend` directly test the new alias path). #1353 tests would NOT fail without the production fix because the test embeds the lock itself.
4. **Tautology check (Action #1200)**: FAIL on `test_two_threads_no_borrow_error` — see Must-fix #2. PASS on the rest.

## Edge cases considered

- Two HTTP threads on cache hit, same session, both calling `_initialize_rust_view` → `set_template_dirs` (rust_bridge.py:334, unprotected) → race possible.
- HTTP thread A inside `render()` (template.py:213, holds `&mut self` across Python callbacks); HTTP thread B in `_sync_state_to_rust` lock window calling `update_state` — race fires inside the implementer's lock window because A's `render()` borrow is still alive.
- WebSocket vs HTTP: WS path is gated by per-consumer `_render_lock` (websocket.py:419 etc), so the race doesn't fire there. HTTP path has no equivalent gate other than this PR's lock — which doesn't cover `render()`.
- `redis+sentinel://` and `unix://` URLs in `DJUST_STATE_BACKEND` — falls through to `merged[config_key] = value`, factory fails with "Unknown backend type".
- `_RUST_VIEW_LOCKS` dict over a multi-day process: never reaped, grows with view churn.
- `id()` reuse: harmless for correctness (lock would just be reused), but doesn't bound dict size.
- Concurrent renders for distinct sessions: `_get_rust_view_lock` correctly returns distinct locks (verified by `test_distinct_views_get_distinct_locks`).

## Question for the implementer

The PR body cites option 1 (per-view lock) over option 2 (clone on cache hit) and option 3 (bypass cache for HTTP). Given that the lock as implemented does not cover the `render()` window where the GIL actually yields — and widening the lock to cover render means serializing all renders for a given (session, view_path) pair, which negates the per-tab parallelism the cache exists to enable — would option 2 (clone the cached `RustLiveView` on HTTP cache hit, leaving the WS path as-is) actually be the cheaper and more correct fix? The Redis backend already does this via msgpack round-trip; in-memory could expose a Rust-side `clone()` or simply re-create from the serialized state.

johnrtipton:
## Stage 12: Address Findings

Both must-fix items + 3 should-fix items addressed in commits 4293f633 + e10be54f.

### #1353 redesign (Must-fix #1: lock window incomplete)

Switched from **Option A** (per-RustLiveView ``threading.Lock``) to **Option B** (clone on cache hit), per the reviewer's recommendation.

**Why Option B**: the lock as implemented covered only ``_sync_state_to_rust``'s mutation calls but the ``RefCell::borrow_mut`` race is between ANY two ``&mut self`` methods on the cached view. The production-trace window includes ``render()`` itself — it holds ``&mut self`` across template evaluation that yields the GIL via ``Context::resolve_dotted_via_getattr`` (``crates/djust_core/src/context.rs``). Widening the lock to cover all unprotected call sites (``template.py:213, 896, 911``, ``rust_bridge.py:334, 361``, plus ``reset()`` at multiple sites) would serialize all renders for a given ``(session, view_path)`` pair, negating the per-tab parallelism the cache exists to enable.

**Implementation**: ``InMemoryStateBackend.get()`` now returns an isolated ``serialize_msgpack`` / ``deserialize_msgpack`` clone of the cached view. This mirrors the ``RedisStateBackend`` contract — Redis already deserialized fresh on every read. With each caller holding its own ``RustLiveView`` instance, no two threads can share a Rust ``&mut self`` borrow and the race class is eliminated at the source.

**Files modified**:
- ``python/djust/state_backends/memory.py`` — ``get()`` does the round-trip; defensive fallback to shared ref on serde failure (logged).
- ``python/djust/mixins/rust_bridge.py`` — removed ``_RUST_VIEW_LOCKS``, ``_RUST_VIEW_LOCKS_GUARD``, ``_get_rust_view_lock``; removed the ``with _rust_lock:`` block in ``_sync_state_to_rust``.

### Tautology test rewrite (Must-fix #2)

The previous ``test_two_threads_no_borrow_error`` embedded the lock inside ``_MinimalRustBridge.sync()`` — would have passed even with no production lock (Action #1200 tautology class).

**New tests** in ``python/tests/test_rust_bridge_concurrent.py``:

- ``TestInMemoryGetReturnsIsolatedView`` (4 cases): contract-level — ``backend.get()`` returns distinct objects, clone inherits state, mutations don't leak across clones or back to the cache, concurrent gets give distinct instances.
- ``TestConcurrentRenderNoBorrowError`` (2 cases): the actual production race scenario. Two threads each call ``backend.get()`` to obtain their own clone, then ``update_state`` + ``render`` with a sidecar whose ``__getattr__`` calls ``time.sleep(0)`` to force GIL yields during ``Context::resolve_dotted_via_getattr``. Asserts no ``Already borrowed`` panic. The lighter ``test_concurrent_update_state_via_backend_no_panic`` runs 4 threads × 100 iters of ``backend.get()`` + ``update_state`` to stress the contract under contention.

**All three test groups verified to FAIL on a fresh main checkout** (without the fix): contract test fails (``backend.get`` returns same ref), mutation-leak test fails (``clone_b`` sees ``mutated_by_a``), concurrent-render test fails with ``RuntimeError: Already borrowed``. They PASS with the new clone-on-get behavior.

### Should-fix
- [x] **``_RUST_VIEW_LOCKS`` unbounded growth**: resolved by Option B — the dict no longer exists. No follow-up needed.
- [x] **``override_settings(spec=[])`` cargo-cult**: removed from ``test_state_backend_config.py:91`` and replaced with explicit ``DJUST_STATE_BACKEND=None`` (which is short-circuited at the alias-merge step in ``utils.py:126-127``, correctly meaning "no top-level alias is set").
- [x] **URL-shape detection**: widened to include ``redis+sentinel://`` (Sentinel HA) via a new ``BackendRegistry._REDIS_URL_PREFIXES`` class constant. Added explicit comments + ``# TODO`` for ``unix://`` (left out because the ``redis-py`` client takes Unix sockets via a different parameter name; cleanly supporting it is a larger change). Two new test cases: ``test_rediss_tls_url_resolves_to_redis_backend`` and ``test_redis_sentinel_url_resolves_to_redis_backend``.

### Test counts after fix
- ``python/tests/test_rust_bridge_concurrent.py``: 6 cases (down from 5; rewritten from scratch — no longer references the deleted lock symbols).
- ``python/tests/test_state_backend_config.py``: 9 cases (up from 7; +2 for the new URL prefixes).

Targeted suites pass: ``pytest python/tests/test_rust_bridge_concurrent.py python/tests/test_state_backend_config.py`` → 15 passed in 0.07s. Full ``pytest python/tests/ tests/`` → 4167 passed, 18 skipped (no regressions). Pre-push hooks passed (including the full pytest run + CHANGELOG count check).

Ready for Stage 13 (Re-Review).

johnrtipton (commented):
# Stage 13 Re-Review: PR #1355 (state-backend safety pair)

## Verdict

APPROVE — both Stage 11 must-fix items and all three should-fix items are addressed; tests fail on the pre-fix code path; CI green; two-commit shape preserved across the fix-pass.

## Findings status

### MUST-FIX #1 — Lock window completeness
- **Original concern**: per-view `threading.Lock` covered only `_sync_state_to_rust`'s mutation calls but left `render()`, `render_with_diff()`, `update_template()`, `set_template_dirs()`, `reset()` outside the lock; `Context::resolve_dotted_via_getattr` yields the GIL inside an active `&mut self` borrow during `render()`, so a peer thread entering any `&mut self` method during that window panicked.
- **Resolution**: ADDRESSED. Implementer switched to Option 2 (clone on cache hit). `python/djust/state_backends/memory.py:107-131` now does `serialize_msgpack` → `deserialize_msgpack` round-trip on every `get()`, so each caller receives a distinct `RustLiveView` instance. Two threads cannot share a `&mut self` borrow because they don't share the underlying object.
- **Lock scaffolding removed**: `python/djust/mixins/rust_bridge.py` — `_RUST_VIEW_LOCKS` dict, `_get_rust_view_lock()` helper, and the `with _rust_lock:` block in `_sync_state_to_rust` (formerly lines 742+) are all gone. `import threading` is also removed (verified: `grep -n "^import threading\|^from threading" python/djust/mixins/rust_bridge.py` → 0 hits).
- **Race-class trace**: confirmed via `_initialize_rust_view` at `python/djust/mixins/rust_bridge.py:311-319` (WS) and `:338-346` (HTTP) — both call `backend.get()`, assign `self._rust_view = cached_view`, then `backend.set(..., cached_view)`. With `get()` cloning, two concurrent threads (A and B) each hold a distinct `RustLiveView`, so subsequent `update_state`, `render`, `set_template_dirs`, etc. cannot collide inside Rust's `RefCell::borrow_mut`.
- **Contract alignment**: `RedisStateBackend.get()` at `python/djust/state_backends/redis.py:213` already does `RustLiveView.deserialize_msgpack(data)` on every read, so the in-memory backend now matches the Redis isolation contract.

### MUST-FIX #2 — Tautology test (Action #1200)
- **Original concern**: `test_two_threads_no_borrow_error` embedded the lock inside `_MinimalRustBridge.sync()`, so the test passed regardless of whether production code was locked.
- **Resolution**: ADDRESSED. The test file is rewritten: `python/tests/test_rust_bridge_concurrent.py` now contains:
  - `TestInMemoryGetReturnsIsolatedView` (4 cases) — directly asserts the production `InMemoryStateBackend.get()` contract: distinct Python objects, isolated mutations, no leak into cache, and concurrent gets all distinct.
  - `TestConcurrentRenderNoBorrowError` (2 cases) — calls `backend.get()` per worker iteration, then runs `render()` with a `_GilYieldingSidecar` whose `__getattr__` calls `time.sleep(0)` to force a GIL yield mid-borrow, mirroring the production race window.
- **Action #1196 verification (would tests FAIL on main?)**: I monkey-patched `InMemoryStateBackend.get` to the pre-fix behavior (`return self._cache.get(key)` verbatim) and re-ran each test. Results:
  - `test_get_returns_distinct_python_object` — fails 100% (`first is second` because shared ref).
  - `test_clone_mutations_dont_leak_into_cache_or_other_clones` — fails 100% (`clone_b` sees `clone_a`'s mutation).
  - `test_concurrent_gets_return_distinct_views` — fails 100% (1 unique id across 32 retained views).
  - `test_get_returned_clone_starts_with_same_state` — passes under both behaviors (consistency check, not a regression test, but documents the contract).
  - `test_concurrent_render_via_backend_no_panic` — fires `Already borrowed` panic in ~5% of runs (1/20 in my characterization). This is timing-dependent — see "Concerns" below.
  - `test_concurrent_update_state_via_backend_no_panic` — did not fire in my 20-run characterization on this hardware/Python build.
- **Net assessment**: the contract-level tests (3 of 4 in `TestInMemoryGetReturnsIsolatedView`) are deterministic regression tests for the underlying invariant. With the contract held, the symptom cannot occur. The two render-path tests are belt-and-suspenders that exercise a representative production path — they're flaky as failing tests on main but cannot fail under the fix (each thread has its own object, so `borrow_mut` collisions are impossible by construction). Discipline is sound.

### SHOULD-FIX #1 — `_RUST_VIEW_LOCKS` unbounded growth
- **Resolution**: N/A under Option 2. The lock dict is gone entirely (no leak surface to bound). Confirmed via `grep -rn "_RUST_VIEW_LOCKS\|_get_rust_view_lock" python/djust/` → 0 hits.

### SHOULD-FIX #2 — `override_settings(spec=[])` cargo-cult
- **Resolution**: ADDRESSED. `python/tests/test_state_backend_config.py:91` (the `with override_settings(spec=[]):` line) is removed. The class-level `@override_settings` decorator on `test_production_fallback_emits_warning` now includes `DJUST_STATE_BACKEND=None`, which is short-circuited at the alias-merge step in `python/djust/utils.py:_merge_top_level_aliases` (the `if value is None: continue` guard). The intent — "no top-level alias is set" — is now achieved via the correct mechanism.

### SHOULD-FIX #3 — URL-shape detection
- **Resolution**: ADDRESSED (with a documented `unix://` follow-up). `python/djust/utils.py:117` introduces `BackendRegistry._REDIS_URL_PREFIXES = ("redis://", "rediss://", "redis+sentinel://")`; line 154 calls `value.startswith(self._REDIS_URL_PREFIXES)`. Two new test cases:
  - `test_rediss_tls_url_resolves_to_redis_backend` (`test_state_backend_config.py:63`).
  - `test_redis_sentinel_url_resolves_to_redis_backend` (`test_state_backend_config.py:71`).
- The reviewer-flagged `unix:///path/to/redis.sock` case is documented as a TODO at `utils.py:108-114` ("the downstream Redis client library accepts Unix sockets via a different parameter name") — which matches the reviewer's "forgivable as an incremental shipment, but worth a `# TODO` plus a follow-up issue" note. Worth filing the follow-up issue; the inline TODO is acceptable for now.

## New concerns

1. **`get_and_update()` still returns a shared reference.** `python/djust/state_backends/memory.py:200` returns `(view, new_timestamp)` directly from the cache without cloning. The method has zero callers in the codebase (`grep -rn "get_and_update" python/` → only the definition), so this is currently dead code — but if/when it gets wired up in the future, it will re-introduce the same shared-ref race class that #1353 just eliminated. Either delete the method as dead code, or apply the same `serialize_msgpack` / `deserialize_msgpack` clone treatment for symmetry. Suggest filing as a low-priority hygiene follow-up. Not a blocker for this PR because the path is unreachable.

2. **Defensive fallback re-introduces the race on serialize failure.** `python/djust/state_backends/memory.py:120-130` catches `Exception` from the round-trip and returns the cached ref with an `exception`-level log. The trade-off ("race window preferable to outright cache miss") is documented honestly, but it's a latent footgun — if a future state mutation introduces a non-msgpack-serializable type, every cached read silently returns the shared ref and the borrow_mut race comes back without test signal. Two tightenings worth considering: (a) add a regression test that asserts the defensive path is only ever exercised on truly unrecoverable serialization failures (state types that violate the rmp_serde contract), or (b) re-raise in production builds via a config flag and only fall back in DEBUG. Not blocking — the fallback is a reasonable tail-risk hedge — but worth the inline note.

3. **Render-path regression tests have a low fire rate on the pre-fix code.** `test_concurrent_render_via_backend_no_panic` characterized at 1/20 (~5%) fire rate against the monkey-patched main behavior; `test_concurrent_update_state_via_backend_no_panic` did not fire in 20 runs on my hardware. The contract-level tests in `TestInMemoryGetReturnsIsolatedView` are deterministic and cover the underlying invariant, so this is acceptable belt-and-suspenders, but the docstrings on the render tests claim "fails on a fresh checkout of `main`" without qualification — it should read "fails non-deterministically on main" or note the timing dependency. Cosmetic nit, not a blocker.

## Edge cases verified

- Two HTTP threads on cache hit, same session — each gets a distinct clone via `backend.get()`; no shared `&mut self` borrows possible.
- WS path identical shape (`rust_bridge.py:311-319`) — same isolation property.
- `_initialize_rust_view` calls `backend.set(self._cache_key, cached_view)` after a get → mutate → set cycle, which means the cache canonical is replaced on every cycle. Each thread's local view is still safe; only the canonical that future gets clone-from rotates. Benign for correctness.
- `set_template_dirs` post-deserialize correctly re-populates the empty `template_dirs` field on the clone (per `crates/djust_live/src/lib.rs:778`).
- Two-commit shape (Action #181):
  - Commit `4293f633`: 5 files changed, NO `CHANGELOG.md`. Verified.
  - Commit `e10be54f`: only `CHANGELOG.md` changed. Verified.
- Test count claims (Action #1049): `grep -c "def test_"` matches CHANGELOG (6 in `test_rust_bridge_concurrent.py`, 9 in `test_state_backend_config.py`).
- Full test suite: 2178 passed, 6 skipped, 0 failures.
- CI: all 13 checks green (rust-tests, python-tests py3.12/3.13/3.14, javascript-tests, playwright-tests, security-scan, security-tests, CodeQL Python, CodeQL JavaScript, Detect Security Hot Spots, javascript CodeQL, test-summary).

## Recommendation

APPROVE.

The implementer correctly diagnosed Stage 11's primary objection (lock didn't cover `render()`'s GIL-yielding `&mut self` window) and pivoted to the cleaner fix that the reviewer's parting question explicitly called out. Option 2 eliminates the race class at the source rather than papering over the most common race window, removes the `_RUST_VIEW_LOCKS` leak surface entirely (Should-fix #1 dissolves), and aligns the in-memory backend's contract with the Redis backend's already-isolated semantics. The contract-level regression tests deterministically verify the invariant; render-path tests are flaky-on-main belt-and-suspenders that cannot fail under the fix.

Recommended follow-up issues (non-blocking):
1. Apply the same clone treatment to `get_and_update()`, or delete it as dead code.
2. Widen URL-shape detection to `unix://` with proper kwarg routing in the Redis factory.
3. Tighten the test docstrings on the render-path tests to acknowledge timing dependence on pre-fix code.

The PR is mergeable.

PR #1359, opened 2026-05-05
fix(eslint): clear 393 bundled-client.js warnings (#1351)

### LINKED ISSUE #1348: dj-transition: 600 ms fallback silently cuts off any animation longer than ~600 ms
## Summary

`_FALLBACK_MS = 600` in [`python/djust/static/djust/src/41-dj-transition.js:52`](https://github.com/djust-org/djust/blob/main/python/djust/static/djust/src/41-dj-transition.js#L52) is a hard cap on the active-class lifetime of any 3-token `dj-transition`. Any single transition that runs longer than ~600 ms (or any multi-property transition where one property runs longer than 600 ms) gets truncated mid-flight: the fallback `setTimeout` fires and removes the active class before `transitionend` arrives, snapping the in-flight properties to whatever value they had at the 600 ms mark.

There's no warning, no `djustDebug` log — just a silently broken animation.

## Reproduce

In a scaffold app, write a 3-token `dj-transition` (e.g. via `dj-transition-group="t-enter-from t-enter-active t-enter-to | …"`) where the active class declares any duration > 600 ms:

```css
.t-enter-from   { opacity: 0; transform: translateX(-100px); }
.t-enter-active { transition: opacity 1500ms ease, transform 1500ms ease; }
.t-enter-to     { opacity: 1; transform: none; }
```

Click whatever causes the new child to mount. Expected: 1.5 s slide-in to `t-enter-to`. Actual: element starts the transition, then snaps to its end-class state after 600 ms — never plays the last 60% of the animation.

## Root cause

In `_runTransition` (lines 100-160), phase 2/3 sets up two cleanup paths:

```js
function onEnd(ev) {
    if (ev.target !== el) return;
    cleanup();
}
state.onEnd = onEnd;
el.addEventListener('transitionend', onEnd);

// Fallback in case transitionend never fires.
state.fallback = setTimeout(cleanup, _FALLBACK_MS);
```

`_FALLBACK_MS = 600`. Whichever fires first wins. For animations < 600 ms, `transitionend` arrives first and the fallback is cleared in `cleanup()` — works as intended. For animations > 600 ms, the fallback fires first, removes `spec.active`, and the in-flight transition stops because the rule that declared it (`transition: …`) is now off the element.

The header comment (line 47) acknowledges the fallback is for "zero-duration transitions or display: none" — i.e. cases where `transitionend` would never fire. It's not meant to be a duration cap, but it functions as one.

## Why this was hard to notice

- Default scaffold demos all use ≤ 400 ms transitions, well under the cap.
- The cap doesn't fail loud — the element ends up at its declared end state (just by snap, not by animation), so there's no broken layout.
- Multi-property transitions where the longest property is > 600 ms also fail the same way, even if other properties are < 600 ms — because `cleanup()` runs once on first transitionend OR fallback, whichever fires first.

I just hit this in [djust-scaffold/apps/features_v09](https://github.com/djust-org/djust-scaffold/tree/main/apps/features_v09) trying to write a more dramatic transition demo (1500-1800 ms multi-property + bouncy bezier). The HTML had a comment claiming `dj-transition reads transition-duration via getComputedStyle and counts expected transitionend events` — that's not true today; static read of this file is unambiguous.

## Proposed fixes (any one)

1. **Parse `transition-duration` + `transition-delay` via `getComputedStyle` after phase 2 lands**, sum to the longest declared duration + a small slack (50-100 ms), use that as the fallback. Auto-correct: never silently truncates. Adds one `getComputedStyle` read per transition (cheap).

2. **Per-spec override**: accept a 4th token / data attribute / dataset key so authors can declare an explicit timeout, e.g. `dj-transition="t-enter-from t-enter-active t-enter-to" data-dj-transition-timeout="1800"`. More explicit but requires authors to know.

3. **`djustDebug` warning when an `active` class declares `transition-duration` > `_FALLBACK_MS`**: at minimum, surface the silent failure. Cheapest fix; doesn't change behaviour, just adds a log line. Could ship alongside (1) or (2) — useful regardless.

I'd lean toward **(1) + (3)** — auto-correct in production, log in dev. (2) is harder to discover and keeps the trap.

## Workaround until then

Keep all property durations ≤ ~550 ms (under the cap with safety margin) and uniform across properties (so all transitionend events fire at the same moment and cleanup arrives after everything has reached its end state).

### LINKED ISSUE #1351: chore: fix 392 pre-existing ESLint warnings in bundled client.js

johnrtipton:
## Summary

Closes #1351 — clears the 393 pre-existing eslint warnings in the bundled `client.js` (and 32 in `debug-panel.js`) so contributors no longer need `SKIP=build-js,eslint` to commit JS source changes. Adds `--max-warnings 0` enforcement to the eslint pre-commit hook so future regressions are caught at commit time.

## Breakdown

- **Auto-fixed: 222** — `prefer-const`, `no-var` via `eslint --fix`
- **Targeted disables: 116 + 25 = 141** — `security/detect-object-injection` on internal data structures (typed for-loop indices, controlled object-literal lookups, DOM-controlled keys like `field.name`). djust already validates against `UNSAFE_KEYS` for the actual prototype-pollution attack surface in `00-namespace.js`.
- **Refactored: 16 + 1 = 17** — 16 `no-unused-vars` (catch-error parameters `_`-prefixed, 2 dead functions inlined or deleted, 1 unused arg renamed); 1 `security/detect-non-literal-regexp` for server-controlled route patterns in `18-navigation.js` documented as ReDoS-safe.
- **Cross-module guards: 4** — `prefer-const` reverted from auto-fix to `let` for cross-file reassigned globals (`liveViewWS`, `clientVdomVersion`, `_eventRefCounter`, `_isBroadcastUpdate`). ESLint's per-file scope incorrectly flagged these as const; auto-fix had broken transport-switch + broadcast paths until reverted (97 vitest failures pre-revert).
- **Total: 425 → 0 warnings**

## ESLint config changes

- `caughtErrorsIgnorePattern: "^_"` so catch-error parameters honour the same `_`-prefix convention as args/vars.
- `sourceType: "script"` for top-level non-module scripts (`client-dev.js`, `security.js`, `react-client.js`, `service-worker.js`) and bundles (`client.js`, `debug-panel.js`).
- `reportUnusedDisableDirectives: off` on the bundles — the source modules carry per-line disables for cross-file reassigned globals; the bundle sees them as unused (rule doesn't fire post-concat).
- Concatenation fragments (`00-namespace.js`, `21-guard-close.js`, `src/debug/*.js`) ignored — they're bundle inputs that don't parse standalone.
- `no-unused-vars: off` on `src/**/*.js` — top-level decls are cross-file by design (concat'd, not import/export); the bundled file enforces with full visibility.
- Global ignores (`**/*.min.js`, `node_modules/**`, etc.) promoted to their own config block so they apply when staged + passed by filename to `npx eslint`.

## Files modified

- `python/djust/static/djust/src/*.js` (29 source modules touched)
- `python/djust/static/djust/src/debug/*.js` (6 source modules touched)
- `python/djust/static/djust/client.js` + `client.min.js{,.gz,.br,.map}` (auto-rebuilt)
- `python/djust/static/djust/debug-panel.js` + `debug-panel.min.js{,.gz,.br,.map}` (auto-rebuilt)
- `eslint.config.js` (per above config changes)
- `.pre-commit-config.yaml` (`--max-warnings 0 --no-warn-ignored`)
- `CHANGELOG.md` (`### Changed` entry)

## Test plan

- [x] `npx eslint --max-warnings 0 python/djust/static/djust/client.js` → 0 warnings
- [x] `npx eslint --max-warnings 0 python/djust/static/djust/debug-panel.js` → 0 warnings
- [x] `npx eslint --max-warnings 0 --no-warn-ignored python/djust/static/djust/src/*.js` → 0 warnings
- [x] `npm test` — 1514 / 1514 pass (was 1417 pre-revert of cross-module const auto-fix)
- [x] `pytest tests/ python/tests/` — 4167 pass, 18 skipped (no regression)
- [x] Pre-commit hook passes WITHOUT `SKIP=build-js,eslint`
- [x] Pre-push hooks all pass

## Unblocks

- PR #1357 (dj-transition fix #1348) — the bundle rebuild deferred there can now happen cleanly.


johnrtipton (commented):
# Stage 11 Code Review: PR #1359 (eslint cleanup #1351)

## Verdict

**APPROVE** — All 141 spot-checked `security/detect-object-injection` disables are credible, the 4 cross-module `let` reverts are verified correct, the eslint flat-config fix addresses a real (latent) bug that was hiding `client.min.js` parse errors, the two-commit shape is clean per Action #181, and CI is green where it matters (`javascript-tests`, `python-tests`, `rust-tests`, `security-scan`, `security-tests` all SUCCESS). The two non-success checks are non-blocking: `playwright-tests` has `continue-on-error: true` and the failing case (`Auto-Save Functionality`) involves only eslint-disable comments (no semantic change in `06-draft-manager.js`); `CodeQL` flags pre-existing duplicate function declarations introduced by PR #1357 (`a27d5e35`), surfaced now only because the bundle was rebuilt.

## Must-fix findings

None.

## Should-fix findings

1. **`18-navigation.js:163-166` — `routeMap[pathname]` rationale is incomplete.** The disable comment says "internal data structures, controlled keys" (per the PR body's blanket claim), but `pathname` here flows from `window.location.pathname` / `newUrl.pathname` — user-controllable. If a user visits `/foo/__proto__`, then `routeMap['__proto__']` returns `Object.prototype` (truthy), and `viewPath` becomes that object. Downstream defenses prevent exploitation: the outgoing `live_redirect_mount` message would carry `view: {}` (`JSON.stringify(Object.prototype) === "{}"`) and the server would reject. So this is defense-by-accident, not by design. Recommend tightening to:
   ```js
   if (Object.prototype.hasOwnProperty.call(routeMap, pathname)) {
       return routeMap[pathname];
   }
   ```
   This is a 2-line tightening that makes the safety explicit. Not blocking — server validation catches the malformed message — but the disable rationale should match reality. Same pattern at `18-navigation.js:165-166` (the `return` line).

2. **CodeQL alerts not addressed (informational).** CodeQL flags duplicate `_parseTimeMs` and `_computeTransitionTiming` declarations in `client.js` lines 13162 and 13171. These are pre-existing — both `41-dj-transition.js:64,73` and `42-dj-remove.js:48,57` declare the same functions, with an explicit comment at `42-dj-remove.js:60` ("duplicated here rather than shared because the source files are concatenated as separate modules"). The alert was introduced by PR #1357 (`a27d5e35`) when the duplication landed; this PR rebuilds the bundle, which is the trigger surface for CodeQL re-evaluation. **Out of scope for this PR** but worth filing as a follow-up issue — the IIFE-per-module concat boundary or a shared `_djust_transition_timing.js` utility module would resolve it.

## Strong points

1. **Cross-module guard reverts are 100% verified.** All 4 (`liveViewWS`, `clientVdomVersion`, `_eventRefCounter`, `_isBroadcastUpdate`) have confirmed cross-file reassignment sites. Verified via `git grep`:
   - `liveViewWS` declared in `03-websocket.js:1243`, reassigned in `01-dom-helpers-turbo.js:87,115`, `13-lazy-hydration.js:147`, `14-init.js:29`
   - `clientVdomVersion` declared in `04-cache.js:9`, reassigned in `01-dom-helpers-turbo.js:91`, `02-response-handler.js:56,77,138`, `03-websocket.js:332,642`, `03b-sse.js:188`
   - `_eventRefCounter` declared in `04-cache.js:16`, reassigned in `03-websocket.js:1081` (via `++`)
   - `_isBroadcastUpdate` declared in `12-vdom-patch.js:572`, reassigned in `02-response-handler.js:102,110,164,184`
   
   I also audited the other 4 module-level `let` declarations (`pendingTurboReinit`, `_scopedDelegationInstalled`, `_djTrackStaticSnapshot`, `_rootObserverInstalled` in `43-dj-transition-group.js`) — all correctly remain `let` because they're reassigned in their declaring file (so prefer-const auto-fix didn't try to convert them). No missed reverts.

2. **The eslint flat-config fix addresses a real latent bug.** I tested before/after: under main's per-block scoping, `npx eslint client.min.js` returned a parser error (`Identifier '_parseTimeMs' has already been declared`) — meaning min files were actually being linted contrary to intent. Post-fix, the global ignores block correctly excludes `**/*.min.js`. The `--no-warn-ignored` flag suppresses the now-correctly-emitted "ignored" warning. Confirmed not a no-op refactor.

3. **The `--max-warnings 0` finding (issue body claim vs reality) was correctly identified.** The issue #1351 body claimed `--max-warnings 0` was implicitly enforced via "npx eslint's default exit-1-on-warning behavior". I verified this is false: `npx eslint /tmp/test-warn.js --rule '{"no-var":"warn"}'` exits 0. The implementer caught the issue's incorrect premise and added the explicit `--max-warnings 0 --no-warn-ignored` flags to `.pre-commit-config.yaml:93`.

4. **Two-commit shape is clean per Action #181.** Stage 5 commit (`3cf8a7ef`) does NOT touch CHANGELOG.md (47 files, none CHANGELOG). Stage 9 commit (`399e0279`) touches ONLY CHANGELOG.md. Verified.

5. **CHANGELOG entry is comprehensive.** Lines 10-39 of the diff cover: per-category breakdown (auto-fixed 222, disabled 116+25=141, refactored 16+1=17, cross-module guards 4), the `--max-warnings 0` enforcement note, the eslint config flat-config fix, and the PR #1357 unblock. Mentions the cross-module guard reverts by name. Test-count claim (1514 vs 1417 = 97 vitest delta) is internally consistent with the implementer's claim of 97 failures pre-revert.

6. **Architectural rationale is documented in `eslint.config.js`.** Lines 141-154 explain why `reportUnusedDisableDirectives: "off"` is needed (cross-file reassign sees the disable as unused in the bundle). Lines 175-197 explain why the concat fragments (`00-namespace.js`, `21-guard-close.js`, `src/debug/*.js`) are ignored (they don't parse standalone). Lines 199-212 explain why `no-unused-vars` is off on `src/**/*.js` (cross-file by design). Reviewer-friendly.

7. **JS test count: 1514/1514 pass.** Locally verified post-revert. The CI `javascript-tests` job is SUCCESS.

## Spot-checks (the load-bearing disables)

I read 14 disable sites across high-risk areas and traced key sources:

| File:line | Bracket access | Key source | Defense | Verdict |
|---|---|---|---|---|
| `04-cache.js:104-105` | `cacheParams[key] = params[key]` | `keyParams` (developer config) | `Object.prototype.hasOwnProperty.call(params, key)` precedes write | Safe |
| `04-cache.js:116-117` | `cacheParams[key] = params[key]` | `Object.keys(params)` (own keys only) | Own-keys-only | Safe |
| `04-cache.js:125-126` | `cacheParams[k]` (read) | Own-keys sort | Own-keys | Safe |
| `07-form-data.js:25-26` | `data[name] = editable.innerHTML` | DOM `name` attr/id | Explicit `UNSAFE_KEYS.includes(name)` filter | Safe |
| `08-event-parsing.js:352-353` | `params[key] = value` | `data-*` attr name | `params = Object.create(null)` + `UNSAFE_KEYS` filter | Defense-in-depth, safe |
| `08-event-parsing.js:376-377` | `params[k] = v` | `dj-params` JSON parse | `UNSAFE_KEYS.includes(k) continue` | Safe |
| `08-event-parsing.js:391-392` | `params[k] = v` | `collectDjValues` (filtered upstream) | Upstream filter + null-prototype source | Safe |
| `08-event-parsing.js:481-482` | `values[key] = value` | `dj-value-*` attr name | `values = Object.create(null)` + `UNSAFE_KEYS` filter | Safe |
| `03-websocket.js:501-502` | `_clientVdomVersions[targetId] = entry.version` | server wire frame `target_id` | DOM `querySelector` + `if (!container) continue` precedes write | Implicitly safe via DOM gate |
| `11-event-handler.js:62-63` | `serverParams[key] = params[key]` | `Object.keys(params)` | Own-keys + 4-key denylist | Safe |
| `11-event-handler.js:74-75` | `optimisticRules[eventName]` (read) | DOM `dj-click="..."` (server-rendered) | Read-only; falsy-coerce on `Object.prototype.target` | Safe |
| `12-vdom-patch.js:200-201` | `children[index]` | numeric path traversal | typed integer index | Safe |
| `12-vdom-patch.js:301-302, 426-427` | `SVG/HTML_ELEMENT_FACTORIES[tagLower]` | server-rendered tag name | `SVG_TAGS.has(tagLower)` Set whitelist precedes write | Safe |
| `15-uploads.js:231-232` | `bytes[i] = ...` | loop counter | Uint8Array typed | Safe |
| `15-uploads.js:454-455` | `xhr.setRequestHeader(name, fields[name])` | Object.keys(fields) | Browser-validated header | Safe |
| `15-uploads.js:636-637, 684-685` | `uploadConfigs[uploadName]` (read) | `dj-upload` attr | All downstream guards: `if (config && ...)` | Safe |
| `16-cursor-overlay.js:154-159` | `cursors[uid]`, `_carets[uid]`, `activeIds[uid]` | server `cursor_positions` event | LiveView server-trust; presence IDs are server-side pks | Safe (under server-trust assumption) |
| `18-navigation.js:53-54, 80-81` | `window.history[method]` | ternary `data.replace ? 'replaceState' : 'pushState'` | 2-string ternary | Provably safe |
| `18-navigation.js:163-166` | `routeMap[pathname]` | `window.location.pathname` (user URL!) | None; downstream `view: {}` rejected by server | Safe by accident — see Should-fix #1 |
| `06-draft-manager.js:120-124` | `field.checked = savedDraft[fieldName]` | `Object.keys(savedDraft)` from localStorage | Own-keys; localStorage is same-origin | Safe |

Key sources I checked: developer config (`keyParams`), own-keys iteration (no prototype), Object.create(null) (no prototype to pollute), explicit `UNSAFE_KEYS` denylist, `Set.has()` whitelist, browser-validated APIs (`setRequestHeader`, `createElement`), DOM-validation gate (server target_id must match `[data-djust-target=...]`), and 2-string ternary dispatch.

The one rationale that's weaker than the comment claims is `18-navigation.js:163-166` (Should-fix #1). All others have credible defenses, even if not always explicitly cited in the disable comment.

### Cross-module guard reverts

All 4 `let` reverts (`liveViewWS`, `clientVdomVersion`, `_eventRefCounter`, `_isBroadcastUpdate`) are necessary and correct. No additional cross-file reassigned vars were missed — I checked all 4 other module-level `let` declarations (`pendingTurboReinit`, `_scopedDelegationInstalled`, `_djTrackStaticSnapshot`, `_rootObserverInstalled`); each is reassigned in its declaring file, so the prefer-const auto-fixer correctly left them as `let`.

### Edge cases considered

- **`SVG_ELEMENT_FACTORIES['__proto__']`**: would return `Object.prototype` (truthy), then call as a function — TypeError. But the call is gated by `SVG_TAGS.has(tagLower)` (Set, prototype-safe). Unreachable.
- **`window.history[method]`**: ternary constrains `method` to two literals — provably safe regardless of input.
- **`uploadConfigs[uploadName]` with `__proto__`**: returns `Object.prototype`, all downstream `config && config.x` checks read undefined props, no exploitable path.
- **`cursors[uid]` with `__proto__`**: server-controlled (presence IDs are server-side pks). Under LiveView trust model, server is trusted.
- **`routeMap[pathname]` with `__proto__`**: user-controlled URL path. Returns `Object.prototype`, downstream `outgoing.view = Object.prototype`, `JSON.stringify` → `"{}"`, server rejects. Defense-by-accident; see Should-fix #1.
- **`debug-panel.js` scope creep (32 warnings)**: per Action #1079, broader-sweep work should fix exactly what the issue cites and file a follow-up. No follow-up filed; the implementer folded debug-panel.js into the same PR. Borderline — same hook (`.pre-commit-config.yaml:91-96` includes both `client.js` and `debug-panel.js`), same fix patterns. I judge this a 🟡 process observation, not a blocker.
- **`prefer-const` auto-fix completeness**: I audited all 4 module-level `let` declarations not in the revert set; none have cross-file reassignment risk.
- **No new `var` declarations introduced**: confirmed via `grep -E "^\+\s*var\s+\w" src/`.
- **CHANGELOG test-count drift**: The CHANGELOG mentions "1514 / 1514 pass" in the PR body but not in the CHANGELOG itself; the CHANGELOG focuses on the 393→0 warning count. No `check-changelog-test-counts.py` violations expected since no test files are referenced in the entry.

This is a security-adjacent change (141 disables on `security/detect-object-injection`) but the spot-checks show consistent defensive patterns across all sampled sites. The one rationale gap (18-navigation.js routeMap[pathname]) is non-exploitable due to downstream server validation. Recommend approving with the Should-fix tightening for #1.

Issue #1358, opened 2026-05-05
enhancement: keyed VDOM diffing for conditional subtrees ({% if %} structural changes) — re-open of #256 Option A


johnrtipton:
## Context

[djust#256](https://github.com/djust-org/djust/issues/256) ("VDOM patching breaks with `{% if %}` blocks") closed in 2026-02 with what appears to be Option B (compile-time/runtime warning) + Option C (recovery-HTML fallback). The architectural Option A — **keyed VDOM diffing for conditional subtrees** — was never shipped.

This enhancement proposes shipping Option A as a follow-up.

## Why it matters

The current behavior is "documented but painful". Every project that uses djust has to internalize the rule "never use `{% if %}` for structural toggles, always use `d-none`" — it's already in CLAUDE.md / project docs across multiple downstream repos. When a developer forgets (which they will, because Django templates make `{% if %}` the obvious tool), the symptom is:

1. Patches partially apply — `[LiveView] X/Y patches failed` warnings
2. Client requests recovery HTML
3. Recovery HTML may be stale or missing (the one-shot `_recovery_html = None` after first use, or a fresh consumer instance after WS reconnect during rolling deploy)
4. **Forced `window.location.reload()`** — loses scroll, focus, in-progress form data

The recovery-HTML fallback (#1202 fix) lessens the impact but doesn't eliminate the root cause. The end-user experience is still page reloads on tab switches in apps that branch on URL state.

**Real-world repro (from NYC Claims, 2026-05-05)**: a `ClaimDetailView` with `{% if active_tab == "overview" %}...{% endif %}` style top-tab branching. Tab switch via `dj-patch` → server re-renders with new `active_tab` → patches target text inside DOM nodes that no longer exist at the same path → 2/24 patches fail → recovery HTML unavailable → page reload. Logs:

```
[LiveView] Patch failed (SetText): node not found at path=0/2/1/1/3/0/0
[LiveView] 2/24 patches failed (22 succeeded)
[LiveView] VDOM patches failed. This usually happens when {% if %} blocks add/remove DOM elements
[LiveView] Server error: Recovery HTML unavailable — the server may have restarted.
[LiveView] Non-recoverable error, reloading page...
```

We're currently on 0.9.2rc1 with the #1202 fix in place; the recovery-HTML path can't help here because the consumer is fresh after a rolling-deploy WS reconnect, so `_recovery_html` is `None`.

## Proposal — keyed conditional diffing

[...]

Issue #295, opened 2026-02-13
VDOM diff breaks on {% if %} conditional blocks (DJE-053)


johnrtipton:
## Problem

When `{% if %}` template blocks add/remove DOM elements, the VDOM positional diff misaligns all subsequent siblings, causing DJE-053 fallback to full HTML updates. This forces template authors to use `style="display:none"` workarounds instead of natural `{% if %}` blocks.

## Root Cause

`diff_indexed_children()` in `crates/djust_vdom/src/diff.rs` matches children by position index. When an `{% if %}` block removes an element, all siblings after it shift left by one, causing the diff to compare wrong node pairs.

## Impact

- Full HTML updates on every event that changes conditional state
- Lost form focus, scroll position, and event listeners on re-render
- Error reports (DJE-053) generated for normal template patterns

## Proposed Fix: Template-level stable placeholders

When the Rust template engine renders `{% if %}` that evaluates to false, emit a stable comment marker `<!--dj-if-->` instead of nothing. This preserves sibling count and positions so the VDOM diff never misaligns.

This is the approach Phoenix LiveView uses — conditional blocks always produce a stable DOM structure.

## Alternative Approaches

- **Stable IDs by template source position**: Assign `data-dj-id` based on template line/column, not output index
- **LCS-based unkeyed diff**: Replace indexed matching with longest-common-subsequence that matches by tag+attributes (like morphdom)

Issue #256, opened 2026-02-10
VDOM patching breaks with {% if %} blocks that add/remove DOM elements


johnrtipton:
## Problem

`{% if condition %}...{% endif %}` blocks that add or remove DOM elements cause 100% VDOM patch failure. The VDOM differ tracks nodes by tree position. When an `{% if %}` block inserts or removes nodes, every subsequent sibling shifts position, making all patches target wrong nodes.

## Impact

This is the single most common failure mode for new djust apps. The framework silently falls back to a full page reload, which loses client-side state (scroll position, focus, selections). DevDash had 12/12 and 22/22 patch failure rates before restructuring all templates.

## Current Workaround

Always render both states, toggle with `style="display:none"`:
```html
<!-- Instead of {% if show_detail %}...{% endif %} -->
<div style="{% if not show_detail %}display:none{% endif %}">...</div>
```

## Suggested Fixes (Pick One or More)

**Option A — Keyed VDOM diffing for conditional blocks**: Teach the differ to match nodes by `data-key` or element identity rather than pure position. When a keyed node moves positions, generate a "move" patch instead of "replace all."

**Option B — Compile-time warning**: During template compilation, detect `{% if %}` blocks containing element nodes (not just text) and emit a warning:
```
Warning: {% if show_detail %} at line 42 adds/removes DOM elements.
This will break VDOM patching. Use style="display:none" toggling instead.
```

**Option C — Resilient patching**: When a patch targets a node that doesn't match the expected tag/id, walk siblings to find the intended target before giving up. This is a heuristic fix but would catch 80% of cases.

## Priority

P0 — Medium-to-large effort, but the highest-impact architectural improvement for developer experience.

PR #1363, opened 2026-05-05
feat: {% if %} blocks emit dj-if boundary markers (Foundation 1 of #1358)

johnrtipton:
## Summary

Iter 1 of 3 toward closing #1358 (re-open of #256 Option A — keyed VDOM diff for conditional subtrees). This iter is **zero-observable-behavior**: at template-render time, every `{% if %}` block whose body contains element nodes gets wrapped in HTML comment boundary markers. Browsers ignore comments → no runtime change.

## Marker shape (Option B — pair per `Node::If`)

```html
<!--dj-if id="if-N"-->...rendered body...<!--/dj-if-->
```

**Why Option B over Option A** (single pair around whole if/elif/else chain): Option A would require detecting that an `If` is the "elif tail" of an outer chain — non-trivial parser plumbing. Option B is mechanical: every `Node::If` owns its pair. The parser already nests an inner `If(B)` inside the outer's `false_nodes` for elif chains, so nested marker pairs emerge naturally:

- `if A` true:  `<!--dj-if id="if-0"-->X<!--/dj-if-->`
- `elif B` fires: `<!--dj-if id="if-0"--><!--dj-if id="if-1"-->Y<!--/dj-if--><!--/dj-if-->`

Iter 3's differ recurses into nested boundaries the same way the renderer does. Slightly more comment volume but symmetric and inspectable.

## ID generation

Stable per-template counter `if-N` assigned at parse time via `parser::assign_if_marker_ids` walking the AST in document order. Stable across re-renders of the same parsed source (parser is deterministic). The `{% for %}{% if %}` pattern reuses the SAME id across loop iterations because the parser only sees one `Node::If`.

## Pure-text conditional skip

A branch is "element-bearing" if it contains any of: `Node::Text` whose string contains `<`, custom tags, components, includes, statics, or nested constructs whose own bodies are element-bearing. Otherwise the branch is "text-only" and emission is skipped — text positions are sibling-stable already in the rendered DOM. The legacy `<!--dj-if-->` placeholder for false-no-else (issue #295) is preserved unchanged for that class.

## `cond=` omitted for safety

Condition strings can contain `--` or `>` that would close the comment early. Sanitisation is fiddly. Iter 3's differ doesn't need to evaluate the condition — the `id=` alone suffices to pair old/new pairs. Documented as future tooling work; can be re-added with sanitisation in a later iter if useful for debug panels.

## Marker preservation across the stack

| Layer | Behaviour | Code |
|---|---|---|
| Rust template renderer | Emits markers around element-bearing if blocks | `crates/djust_templates/src/renderer.rs` |
| Rust VDOM parser | Preserves opening + closing markers as comment vnodes alongside legacy `<!--dj-if-->` | `crates/djust_vdom/src/parser.rs:478-510` |
| Public `render_template` / `render_template_with_dirs` | Strips ALL `dj-if`-family markers via `strip_dj_if_markers` regex helper | `crates/djust_templates/src/lib.rs:285-293` and `crates/djust_live/src/lib.rs:918-924` |

This keeps server-side VDOM in lock-step with the client's DOM (the client's `getSignificantChildren` counts all comments) while preserving the existing contract that public rendering yields clean HTML.

## What this enables (NOT in this PR)

- Iter 2 (Foundation 2): client patch applier learns `RemoveSubtree` / `InsertSubtree` patch types.
- Iter 3 (Capability): Rust VDOM differ recognises `dj-if` boundaries; emits subtree-level patches when conditionals flip.

## Test plan

- [x] 19 regression cases in `crates/djust_templates/tests/test_if_markers.rs` (element-bearing emission, pure-text skip, elif/else chains, nested ifs, for-if iteration, ID stability, attribute context, sibling stability)
- [x] 4 cases in `crates/djust_vdom/src/parser.rs` `parser::tests` (legacy + new boundary marker preservation, unrelated comments still filtered)
- [x] 19 cases in `python/tests/test_template_if_markers.py` (Python-level Template API + public-render strip path)
- [x] Full Rust workspace test suite (`cargo test --workspace --exclude djust_live`) passes — all 1198 tests
- [x] Full Python suite (`pytest python/tests/ tests/`) passes — 4186 passed, 18 skipped
- [x] Full JS suite (`npm test`) passes — 1514 tests
- [x] No existing test fails — confirms zero-observable-behavior

- Commit 1 (impl + tests + Cargo.lock fixup): `484947da`
- Commit 2 (CHANGELOG only): `3cfe8afa`

> Note: Cargo.lock had pre-existing drift (workspace at rc.2 but lockfile at rc.1 from incomplete prior version-bump commit). Bundled into commit 1 as a one-time fixup so future PRs don't repeatedly hit the same drift in pre-commit hooks.

Foundation 1 of 3 toward #1358.


johnrtipton (commented):
# Stage 11 Code Review: PR #1363 (Foundation 1 of #1358 — dj-if markers)

## Verdict

REQUEST_CHANGES — Two design-contract issues need attention before this Iter 1 locks in shape that Iter 3 will inherit. The marker shape (Option B) and predicate logic are workable, but ID uniqueness under inheritance/include is unaddressed and the element-bearing classifier misses `Node::CsrfToken`. The scope-expansion to the VDOM parser is justified but only half-applied (server side broadened; client `getNodeByPath` filter unchanged).

## 🔴 Must-fix findings

### 1. ID collision under `{% extends %}` and `{% include %}`

`crates/djust_templates/src/parser.rs:230-253` — `assign_if_marker_ids` runs once per `parser::parse()` call with a counter starting at `0`. Every parsed template (parent, child, included partials) gets its own `if-0`, `if-1`, ... namespace. Inheritance preserves the assigned IDs verbatim (`crates/djust_templates/src/inheritance.rs:118-130` clones `marker_id` through `apply_block_overrides`).

Concrete failure case:

```django
{# base.html #}
<html>
  {% if bar %}<aside>Bar</aside>{% endif %}
  {% block content %}{% endblock %}
</html>
```

```django
{# child.html #}
{% extends "base.html" %}
{% block content %}
  {% if foo %}<div>Foo</div>{% endif %}
{% endblock %}
```

Renders to (both true):

```html
<html>
  <!--dj-if id="if-0"--><aside>Bar</aside><!--/dj-if-->
  <!--dj-if id="if-0"--><div>Foo</div><!--/dj-if-->
</html>
```

Two `id="if-0"` markers in the same rendered HTML. Same applies to `{% include %}` because each included template parses through its own counter. The CHANGELOG says "Iter 3's differ keys off the `id` alone" — that contract is unsupportable as-shipped.

Mitigations to consider:
- (a) Disambiguate at composition time: when `apply_block_overrides` (inheritance) or `Node::Include` rendering compose subtrees, walk and re-number the merged tree.
- (b) Pair the ID with the parent template path / hash (`if-{template_hash}-N`).
- (c) Document that Iter 3 must key off `(dom_path, id)` rather than `id` alone, and add a test asserting that two `id="if-0"` markers from different templates can coexist without conflict (proves Iter 3's key composition won't fight the IDs).

I'd lean toward (c) — it's the cheapest and matches how most VDOM differs already work — but the contract change needs to be explicit in the PR docstring + at least one test that exercises the include-with-if pattern. As shipped, the ID design contract is silently over-promised.

Add a regression test under `tests/test_if_markers.rs` that renders parent-with-if extending base-with-if and asserts the rendered output (whatever the chosen scheme produces). Without it, Iter 3 will hit this and need a foundation re-do.

### 2. `Node::CsrfToken` misclassified as text-only

`crates/djust_templates/src/renderer.rs:67-76` — `node_is_element_bearing` lists `Node::CsrfToken` alongside truly-text-only tags (`Node::Now`, `Node::WidthRatio`, ...). But `Node::CsrfToken` actually renders `<input type="hidden" name="csrfmiddlewaretoken" value="...">` (verified at `crates/djust_templates/src/renderer.rs:737-758`).

Concrete failure case: `{% if request.method == "POST" %}{% csrf_token %}{% endif %}` would not emit dj-if boundary markers because the helper returns `false` for the only inner node. Iter 3's differ would have no keyed boundary for this conditional, and the `<input>` could end up unkeyed in subsequent diffs.

Fix: move `Node::CsrfToken` into the element-bearing branch (line 92-99). One-liner.

This is also a missing test — there's no test in `tests/test_if_markers.rs` covering `{% if x %}{% csrf_token %}{% endif %}`. Add one that asserts markers ARE emitted around a csrf_token-only conditional.

## 🟡 Should-fix findings

### 3. Client-side `getNodeByPath` filter not updated to match server broadening

`python/djust/static/djust/src/12-vdom-patch.js:169-185` — the path-traversal filter still only includes `dj-if` exact-match comments:

```javascript
if (child.nodeType === Node.COMMENT_NODE) {
    return child.textContent.trim() === 'dj-if';
}
```

The server VDOM parser was broadened (`crates/djust_vdom/src/parser.rs:494-512`) to preserve `dj-if id="..."` and `/dj-if` markers, so server-side patch indices now count those comments. But the client's path traversal still excludes them — so server `path: [parent_idx, ...]` and client interpretation drift when the path crosses an element-bearing if block.

In practice this is mostly masked by Strategy 1 (dj-id-based resolution) at line 147-152, which fires first when a patch carries `patch.d`. But the path FALLBACK is broken whenever dj-id resolution fails (e.g., dj-id stamping race, third-party DOM injection) for a target near a boundary marker.

Two options:
- (a) Update line 169-185's filter to mirror the server broadening (also accept `dj-if id="..."` and `/dj-if`).
- (b) Document that this asymmetry is intentional and that path traversal is best-effort fallback only.

The implementer's PR description says the broadening "keeps server-side VDOM in lock-step with the client's DOM structure (the client's `getSignificantChildren` counts all comments)" — that's true for `getSignificantChildren` (line 1100-1112) but NOT for `getNodeByPath` (line 169-185). The two filters differ. Worth flagging in either the comment at server parser.rs:478-494 or as a TODO for Iter 2.

### 4. `assign_if_marker_ids` recurses through `BlockCustomTag` and `ReactComponent` children, but skips `CustomTag`

`crates/djust_templates/src/parser.rs:266-309` — the recursion list includes `BlockCustomTag` (children) and `ReactComponent` (children), but `CustomTag` (no children — args only) is correctly skipped. Verify this matches your intent: are there any custom tags that nest user templates with `{% if %}` inside? If yes, those if blocks won't get IDs.

Looking at `Node::CustomTag` definition (line 78-83) — no children field. So this is fine. Suggestion: add a comment in `assign_if_marker_ids` listing the variants explicitly skipped (CustomTag, Static, Include, etc.) so future maintainers don't have to re-derive the partition.

### 5. Predicate-boundary tests missing for `dj-if`-lookalike comments

`crates/djust_vdom/src/parser.rs::tests` — the new tests cover `dj-if id="if-0"`, `/dj-if`, and "regular comment" but NOT adversarial close-but-not-matching cases:

- `<!--dj-iffy-->` (suffix continuation)
- `<!--dj-if-extra-->` (hyphen continuation)
- `<!--dj-if\nid="if-0"-->` (newline separator instead of space/tab)

The current predicate at line 494-512 correctly rejects the first two (the space/tab requirement filters them out) but the newline case is ambiguous. Add at least one test asserting `<!--dj-iffy-->` is FILTERED, and ideally one positive case `<!--dj-if  id="if-0"-->` (double-space) to lock in the predicate boundary.

### 6. Test count claim drifts vs reality

CHANGELOG.md claims "42 regression cases" (line 49 of diff). Actual on disk: 19 + 4 + 19 = 42. ✓ Matches.

But `test_pure_text_if_false_keeps_legacy_placeholder` (Rust + Python) and several VDOM parser tests would PASS on main without the PR (they verify preserved behavior, not new behavior). Per Action #1196 + #1200, every test should fail on main without the fix. Spot-check: 3 of 4 VDOM tests pass on main (`test_parse_preserves_dj_if_legacy_placeholder`, `test_parse_filters_unrelated_comments_when_dj_if_present`).

These aren't tautologies in the strict Action #1200 sense (they're regression-protection for preserved behavior), but they shouldn't be counted as "new test discipline for this feature." Consider restating in CHANGELOG as "42 cases — 28 new behavior, 14 regression-protection for preserved behavior."

### 7. CSRF-token, csrf_token, Now, FirstOf classification audit

Beyond `Node::CsrfToken`, audit the other "text-only" classifications:
- `Node::FirstOf`: outputs first truthy variable's value. If the variable is HTML (e.g., user passed `Value::String("<a>foo</a>")` from a `|safe` filter chain), it's element-bearing in practice. Marginal — depends on user discipline.
- `Node::TemplateTag`, `Node::Now`, `Node::WidthRatio`, `Node::Cycle`: all genuinely text-only.

Recommend adding a comment block above `node_is_element_bearing` listing the explicit "text-only" assumption for each false-returning variant, so future tag additions can be classified correctly.

## 🟢 Strong points

1. **Marker shape (Option B) recursion is clean.** The `if/elif/else` → nested `Node::If` mapping at the parser level naturally produces nested marker pairs. The differ in Iter 3 will recurse the same way as the renderer. The 3 elif-chain tests (`elif_chain_first_branch_fires`, `_second_branch_fires`, `_else_branch_fires`) lock in the exact shape, so any future regression here is caught.

2. **`cond=` correctly omitted for safety.** Stage 8's warning about `-->` injection via condition strings is moot — the marker carries only `id="if-N"` where N is a `usize` counter. No injection vector.

3. **HTML attribute context (#380) preserved.** The renderer correctly skips marker emission when `in_tag_context == true` (`crates/djust_templates/src/renderer.rs:486-495`). Test `no_markers_in_attribute_context` covers both true and false branches.

4. **Pure-text skip (`nodes_contain_elements`) is conservative-safe.** Misclassifying text-only paths as element-bearing only emits redundant comments (browsers ignore). The doc comment at line 26-43 of renderer.rs makes this trade-off explicit.

5. **`strip_dj_if_markers` regex extraction is clean.** Public render path correctly preserves "yields clean HTML" contract via a single regex, applied uniformly in both `crates/djust_templates/src/lib.rs:285-289` and `crates/djust_live/src/lib.rs:917-921`. Both call sites updated atomically.

6. **`for/if` iteration stability test** (`for_if_iteration_stable_id`) correctly verifies that the parser sees one `Node::If` regardless of iteration count. Both iterations emit the same `id="if-0"`. Locks in the design contract for loop-conditional patterns.

7. **Cargo.lock fixup is bounded.** Diff is purely the rc.1 → rc.2 catchup for the 5 internal crates. No surprise external dep bumps. Bundling rationale (avoid hook loop) is acceptable.

8. **`marker_id: Option<String>` defensive default.** The `None` fallback for hand-constructed `Node::If` (tests) avoids breaking the existing inheritance test fixtures. Comment at parser.rs:25-32 explains the contract clearly.

## Marker shape design contract (load-bearing for Iter 3)

**Option B (pair per `Node::If`) is workable for Iter 3.** The nested-marker pattern emerges naturally from the parser AST shape — `if A elif B else C` is `If { false_nodes: [If { false_nodes: [...C...] }] }`, so the renderer recursively wraps each. Iter 3's differ:

1. Walks server VDOM children of a parent.
2. When it hits `<!--dj-if id="if-N"-->` (start) followed eventually by `<!--/dj-if-->` (end), it treats the contents as a keyed subtree under key `if-N`.
3. Recurses into that subtree to handle nested if-pairs (elif chain).

The OLD-vs-NEW match for the elif flip (`A=true → A=false, B=true`) would be:
- OLD: `[<!--dj-if id="if-0"-->, <div>A</div>, <!--/dj-if-->]`.
- NEW: `[<!--dj-if id="if-0"-->, <!--dj-if id="if-1"-->, <div>B</div>, <!--/dj-if-->, <!--/dj-if-->]`.

Iter 3 matches the outer if-0 boundary by id, recurses inside, finds entirely different children, emits a subtree-replace patch (RemoveAll + InsertSubtree). Or it could detect that the inner if-1 is "new" and emit a more granular patch. Either is valid — Option B doesn't pin down Iter 3 to a single shape.

**The one Iter-3 design constraint that's NOT yet locked in**: ID uniqueness across inheritance. See 🔴 finding #1. If Iter 3 keys by `id` alone, it cannot distinguish `if-0` in base.html from `if-0` in child.html when they end up in the same rendered tree. Fix the contract before Iter 2/3 builds on it.

## ID generation stability

- **Cache hits**: ✓ Stable. `Template::new(source)` parses once → `assign_if_marker_ids` assigns once → cached `Template` reuses same IDs across renders. Tests `id_stable_across_renders`, `id_stable_across_separate_template_instances`, `id_unchanged_after_context_mutation` all pass.

- **`{% for %}{% if %}`**: ✓ Same ID across iterations. Parser only sees one `Node::If`. Test `for_if_iteration_stable_id` locks this in.

- **Inheritance**: ❌ See 🔴 finding #1. IDs collide across parent + child templates.

- **`{% include %}`**: ❌ Same — each included template parses with its own counter starting at 0.

- **Macros / template-tag-emitted partials**: N/A — djust templates don't have macros. Custom block tags (`Node::BlockCustomTag`) DO have children, but those children are part of the parent template's parse, so they get IDs from the parent's counter. ✓ No collision risk.

## VDOM parser broadening (the unexpected scope)

**Necessary for Iter 1 alone? Mostly yes.** Without it:
- Server renders HTML with `dj-if id="..."` markers.
- Server `parse_html()` filters them out (only legacy `dj-if` exact was preserved).
- Diff produces patches based on the FILTERED VDOM (no boundary markers).
- Browser DOM has the markers (initial HTML parse preserves comments).
- Patches and DOM disagree on child indices.

So the broadening prevents that drift on the diff side. ✓ Justified.

**Alternative (server tightens, client filters dj-if id markers out)?** Would require:
- Server emits markers, then strips them post-render before parse_html. Brittle (requires consistent strip logic everywhere) and defeats the whole point — Iter 3 needs the markers IN the server VDOM to key off them.

**Other framework comments?** None — `dj-if` (legacy) is the only HTML-comment-based marker djust uses currently. The broadening doesn't introduce a "dj-if is special, others aren't" asymmetry — dj-if was already special. Future markers would need to extend the predicate.

**Predicate boundaries?** The space/tab discriminator (line 497-498) correctly rejects `dj-iffy` and `dj-if-extra`. Negative-case tests would lock this in (see 🟡 finding #5).

**Asymmetry with client `getSignificantChildren`?** Pre-existing for regular comments (server filters, client includes). The broadening narrows the gap for `dj-if` family. Not a new issue.

**Asymmetry with client `getNodeByPath`?** NEW issue introduced by this PR — client filter at vdom-patch.js:181 wasn't updated. See 🟡 finding #3. Mostly masked by dj-id resolution but worth fixing or documenting.

## Test discipline

Spot-check of 8 representative tests:

| Test | Asserts | Fails on main? |
|------|---------|---------------|
| `markers_emitted_around_simple_element_if_true` (Rust) | Marker pair around `<div>foo</div>` | ✓ Yes |
| `no_markers_for_pure_text_if_false_keeps_legacy_placeholder` (Rust) | `result == "<!--dj-if-->"` | ✗ No (preserved behavior) |
| `elif_chain_second_branch_fires` (Rust) | Nested marker pair for elif | ✓ Yes |
| `for_if_iteration_stable_id` (Rust) | Both iterations emit `if-0` | ✓ Yes |
| `test_simple_element_if_true` (Python) | Marker pair around `<div>foo</div>` | ✓ Yes |
| `test_pure_text_if_false_keeps_legacy_placeholder` (Python) | `result == "<!--dj-if-->"` | ✗ No (preserved) |
| `test_parse_preserves_dj_if_boundary_markers` (VDOM) | Boundary markers preserved as comment vnodes | ✓ Yes |
| `test_parse_filters_unrelated_comments_when_dj_if_present` (VDOM) | Regular comments filtered | ✗ No (preserved) |

5 of 8 fail on main (good); 3 of 8 are regression-protection for preserved behavior (still useful). No tautologies in the Action #1200 sense — every assertion is keyed to either new behavior or an explicit invariant. No "would this pass if action did nothing?" cases found.

## Edge cases considered

- **`{% csrf_token %}` inside if** — 🔴 misclassified, see finding #2.
- **`{% include %}` inside if** — emits markers (Node::Include is element-bearing). ✓
- **HTML comment inside if body** — `Node::Text("<!-- ... -->")` contains `<`, marker emitted (over-eager but safe). ✓
- **`{% if %}` inside attribute** — markers correctly skipped (in_tag_context guard). ✓
- **`{% if %}{{ var }}{% endif %}`** — variable, no marker (variable is escaped, no element). ✓
- **`{% if %}{% if %}<span>x</span>{% endif %}{% endif %}`** — nested ifs get distinct IDs. Test `nested_if_distinct_ids` covers. ✓
- **Same template, two ifs at top level** — `sequential_ifs_get_distinct_ids` covers. ✓
- **Empty body (false-no-else, element-bearing)** — `<!--dj-if id="if-0"--><!--/dj-if-->` empty pair. Test `markers_emitted_around_simple_element_if_false_no_else` covers. ✓
- **Mutation-after-capture** — `id_unchanged_after_context_mutation` covers (Action #1039 discipline). ✓
- **`<!--dj-iffy-->` predicate boundary** — 🟡 missing test, see finding #5.
- **`<!--dj-if  id=...-->` (multiple spaces)** — predicate accepts (starts_with covers). Untested.
- **Newline-separated `<!--dj-if\nid=...-->`** — predicate REJECTS. The renderer never emits this, so OK in practice, but worth a comment.

## Summary

Iter 1's marker shape, ID generation algorithm, and VDOM parser broadening are competent design choices that won't fight Iter 3 — with the exception of two design contract issues:

1. **🔴 ID uniqueness under inheritance/include is unaddressed.** The PR ships with a contract that `id="if-N"` is a valid Iter-3 differ key, but IDs collide across template-composition boundaries. Either fix the ID scheme now or document that Iter 3 must compose `(dom_path, id)` for uniqueness.

2. **🔴 `Node::CsrfToken` misclassified.** Trivial one-liner fix; missing test.

The unexpected scope expansion to `crates/djust_vdom/src/parser.rs` is justified for Iter 1 correctness and well-tested. The companion expansion of `python/djust/static/djust/src/12-vdom-patch.js:169-185` (line 181-183 filter) was missed — non-blocking for Iter 1 because dj-id resolution masks it, but should be addressed in Iter 2 at the latest.

Test discipline is solid (Action #1196 + #1200): 42 cases, ~67% fail-on-main, the rest are regression-protection. Cargo.lock fixup is bounded and acceptable.

Once findings #1 and #2 are addressed (with regression tests), this is a strong foundation for Iter 2/3.

johnrtipton:
## Stage 12: Address Findings

All 4 Stage 11 findings addressed in commits `daa5b413` (impl + tests) and `cb11a6ef` (CHANGELOG).

### MUST-FIX #1: ID collision under `{% extends %}` / `{% include %}`

**Option chosen: B (per-template source-derived prefix).** Rationale: keeps Iter 3's "differ keys off the `id` alone" contract clean, no composition-timing coordination needed, prefix derives deterministically from source so caching still works. ID format changes from `if-N` to `if-<8hex>-N`.

Implementation:
- New `parser::parse_with_source(tokens, source)` derives the prefix from a hash of the template source.
- Legacy `parser::parse(tokens)` falls back to hashing the token stream (for callers that don't have the source string).
- `Template::new`, `FilesystemTemplateLoader::load_template`, `TestTemplateLoader::load_template`, and the `resolve_inheritance` standalone path all now use `parse_with_source`.
- Templates loaded via inheritance/includes get their own per-source prefix, so ID collisions across composed renders are eliminated at the source.

New tests: 4 cases under `cross_template_ids` mod in `test_if_markers.rs` (extends parent+child distinct prefixes; include doesn't collide with parent; nested-include chain stays unique; same source same prefix). 3 prefix-uniqueness cases in `python/tests/test_template_if_markers.py::TestIdPrefixUniqueness` (different sources → different prefixes; same source → byte-identical IDs; ID format conformance). 4 unit tests in `parser::tests` (shape, source-distinctness, source-stability, token-fallback).

### MUST-FIX #2: `Node::CsrfToken` misclassified as text-only

One-line classifier fix in `crates/djust_templates/src/renderer.rs::node_is_element_bearing` — moved `Node::CsrfToken` out of the text-only group into its own element-bearing arm with a comment explaining the conservative classification (`<input type="hidden">` when token present; empty string when no token, but extra marker pair is harmless).

New regression tests: `csrf_token_inside_if_emits_markers`, `variable_only_inside_if_does_not_emit_markers`, and `raw_html_input_inside_if_emits_markers` in the Rust integration test, plus 3 mirroring Python tests in `TestCsrfTokenElementBearing`.

### SHOULD-FIX #3: Client-side filter not mirroring server

Added `isDjIfComment(text)` helper in `python/djust/static/djust/src/12-vdom-patch.js` that mirrors the server's `crates/djust_vdom/src/parser.rs:494-499` predicate exactly:
- exact `dj-if` (legacy placeholder, #295)
- `dj-if<space-or-tab>...` (boundary opening, #1358 Iter 1)
- `/dj-if` (boundary closing)

Wired into `getNodeByPath`'s path-fallback comment-filter (replacing the strict `=== 'dj-if'` match). Bundled `client.js` rebuilt. Helper exposed at `window.djust._isDjIfComment` for tests.

20 new JS test cases in `tests/js/dj_if_comment_predicate.test.js` (predicate matrix: 7 positive matches incl. prefixed-id + tab + whitespace, 8 negative incl. lookalikes + non-string types, plus 5 path-fallback integration cases).

### SHOULD-FIX #4: Predicate boundary tests missing

Added 7 boundary-case tests to `crates/djust_vdom/src/parser.rs::tests` covering:
- `dj-iffy` rejected (longer-prefix lookalike)
- `dj-if-extra` rejected (suffix-after-if lookalike)
- ` dj-if id="..." ` accepted (leading/trailing whitespace via `trim()`)
- `dj-if<TAB>id=...` accepted (tab separator)
- `dj-ifid="..."` rejected (no separator)
- `dj-if id="if-deadbeef-0"` accepted (prefixed-id format from MUST-FIX #1)
- `/dj-iffy`, `/dj-if-extra`, `/dj-if-stuff` rejected (close-marker family lookalikes)

Confirmed the existing `starts_with("dj-if ") || starts_with("dj-if\t")` predicate handles all cases correctly — no predicate change needed.

### Verification

- `cargo test --workspace --exclude djust_live` — 704 cases, all green.
- `npm test` — 128 files / 1534 cases, all green.
- `pytest tests/ python/tests/ python/djust/tests/` — 6926 passed, 20 skipped.

Ready for Stage 13 Re-Review.

johnrtipton (commented):
# Stage 13 Re-Review: PR #1363 (Foundation 1 of #1358 — dj-if markers)

## Verdict

APPROVE (advisory; cannot self-approve via gh API) — All 4 Stage 11 findings were addressed correctly. Cross-template ID uniqueness is genuinely fixed (verified the pre-fix state would not even compile against the new test surface), CSRF token classification is exhaustive across the Node enum, and the client-side predicate is byte-equivalent to the server's. CI is green except the known pre-existing playwright flake from PR #1359 (Auto-Save Functionality, unrelated).

## Findings status

### MUST-FIX #1 (ID collision under inheritance/include)

- **Original**: `parser::assign_if_marker_ids` started counter from 0 per parsed template. Parent + child + included templates each independently emitted `if-0`, `if-1`, ... causing duplicates in the rendered HTML and breaking Iter 3's "differ keys off `id` alone" contract.

- **Resolution**: ADDRESSED. Per-source 8-hex-char prefix scheme (Option B) produces ID format `if-<8hex>-<N>`. New `parser::parse_with_source(tokens, source)` derives the prefix via Rust's `DefaultHasher` (SipHash-1-3) of the source string and truncates to low 32 bits. Legacy `parser::parse(tokens)` falls back to hashing the token-stream's Debug repr. Same source → same prefix (deterministic); different sources → distinct prefixes with ~1/4 billion pairwise collision rate.

- **Evidence**:
  - `crates/djust_templates/src/parser.rs:245-265` — `parse` and `parse_with_source` definitions.
  - `crates/djust_templates/src/parser.rs:267-299` — `parse_internal` shared core with `format_id_prefix` driving the per-template counter.
  - `crates/djust_templates/src/parser.rs:301-335` — `hash_source`, `hash_tokens`, `format_id_prefix`.
  - `crates/djust_templates/src/parser.rs:353-393` — `assign_if_marker_ids` recurses through If, For, Block, With, Spaceless, BlockCustomTag, ReactComponent.
  - **All production callsites use `parse_with_source`**: `lib.rs:98` (Template::new), `lib.rs:364` (TestTemplateLoader), `inheritance.rs:296` (FilesystemTemplateLoader), `inheritance.rs:550` (resolve_template_inheritance). Remaining `parse(tokens)` callers (`inheritance.rs:691,711`, `renderer.rs:1994`, `parser.rs:1581,1588`) are all `#[cfg(test)]` blocks. `parser.rs:1105` (extract_template_variables) is a JIT-serialization variable extractor that doesn't render markers — hash prefix doesn't apply.
  - `apply_block_overrides` at `inheritance.rs:120-129` correctly clones `marker_id` so block-override merges preserve per-template prefixes.
  - **New tests (all PASS, 4 unit + 4 integration)**:
    - `parser::tests::test_parse_with_source_assigns_prefixed_id` — locks `if-[0-9a-f]{8}-\d+` shape.
    - `parser::tests::test_parse_with_source_distinct_sources_distinct_prefixes` — different sources → different prefixes.
    - `parser::tests::test_parse_with_source_same_source_same_prefix` — stability.
    - `parser::tests::test_parse_legacy_uses_token_hash_prefix` — fallback path stability.
    - `cross_template_ids::extends_parent_and_child_get_distinct_id_prefixes` (test_if_markers.rs:461) — 4 ifs across parent+child render to 4 unique IDs across 2 prefixes.
    - `cross_template_ids::include_does_not_collide_with_parent_ids` (line 514) — parent+include each contribute unique prefix.
    - `cross_template_ids::nested_include_chain_keeps_ids_unique` (line 545) — 3-level chain → 3 unique IDs.
    - `cross_template_ids::id_prefix_stable_across_re_renders` (line 574) — byte-identical IDs across renders.
  - Verified the test surface doesn't compile against `main` (`parse_with_source` missing) → tests are non-tautological.

### MUST-FIX #2 (CsrfToken misclassification)

- **Original**: `node_is_element_bearing` at `renderer.rs:67-76` treated `CsrfToken` as text-only. CSRF token actually renders `<input type="hidden">`. `{% if x %}{% csrf_token %}{% endif %}` would skip emitting markers, leaving Iter 3's differ blind.

- **Resolution**: ADDRESSED. `Node::CsrfToken` now in its own `=> true` arm with conservative-classification rationale. I cross-checked all 25 Node enum variants against the classifier match arms:
  - `Text(s)` → `s.contains('<')` — fine, conservative
  - `Variable, InlineIf, Comment` → false — output is HTML-escaped
  - `CsrfToken` → **true** (FIXED)
  - `Now, WidthRatio, FirstOf, TemplateTag, Cycle, Load, Extends, AssignTag` → false — verified each render arm at `renderer.rs:801-906` only emits text/empty
  - `If, For, Block, With, Spaceless` → recurse into children
  - `Static, Include, ReactComponent, RustComponent, CustomTag, BlockCustomTag, UnsupportedTag` → true (conservative — they may produce HTML)
  
  Match arms are exhaustive (Rust would compile-error otherwise). No other variants similarly misclassified.

- **Evidence**:
  - `crates/djust_templates/src/renderer.rs:65-75` — `CsrfToken => true` with comment citing the actual CSRF render path and the LiveView re-render edge case (#696).
  - `csrf_token_inside_if_emits_markers` at `test_if_markers.rs:608` — wraps the `<input type="hidden">` output in marker pair.
  - `variable_only_inside_if_does_not_emit_markers` at line 634 — regression guard.
  - `raw_html_input_inside_if_emits_markers` at line 645 — baseline regression guard.
  - Python integration: `TestCsrfTokenElementBearing` 3 tests at `python/tests/test_template_if_markers.py:326-355`.

### SHOULD-FIX #3 (client filter mirror)

- **Resolution**: ADDRESSED. New `isDjIfComment(text)` helper at `python/djust/static/djust/src/12-vdom-patch.js:35-44` is byte-equivalent to the server predicate at `crates/djust_vdom/src/parser.rs:494-499`:
  - Server: `trimmed == "dj-if"` || `starts_with("dj-if ")` || `starts_with("dj-if\t")` || `trimmed == "/dj-if"`
  - Client: `trimmed === 'dj-if'` || `trimmed === '/dj-if'` || `startsWith('dj-if ')` || `startsWith('dj-if\t')`
  
  Both trim first, both apply identical pattern set. Wired into `getNodeByPath`'s comment filter at line 215-217. Bundled `client.js` rebuilt (3 occurrences of `isDjIfComment` confirmed in `python/djust/static/djust/client.js:5329-5337,5510`). `client.min.js` and source map also updated in the commit.

- **Evidence**:
  - `python/djust/static/djust/src/12-vdom-patch.js:35-44` — `isDjIfComment` definition.
  - `python/djust/static/djust/src/12-vdom-patch.js:209-217` — `getNodeByPath` path-fallback wires through `isDjIfComment`.
  - `python/djust/static/djust/src/12-vdom-patch.js:1171` — exposed as `window.djust._isDjIfComment` for testability.
  - `tests/js/dj_if_comment_predicate.test.js` — 20 cases (positive: exact `dj-if`, `/dj-if`, `dj-if id="..."`, prefixed-id `if-a3b1c2d4-0`, tab separator, leading/trailing whitespace; negative: `dj-iffy`, `dj-if-extra`, `dj-ifid="..."`, `/dj-iffy`, `/dj-if-extra`, empty string, regular comments, non-string input; integration: counts dj-if-family in path traversal, skips regular comments + lookalikes, still counts legacy placeholder). All 20 PASS.

### SHOULD-FIX #4 (predicate boundary tests)

- **Resolution**: ADDRESSED. 7 boundary tests at `crates/djust_vdom/src/parser.rs:909-1008`. All PASS:
  - `test_parse_rejects_dj_if_lookalike_dj_iffy` — `dj-iffy` not preserved.
  - `test_parse_rejects_dj_if_lookalike_dj_if_extra` — `dj-if-extra` not preserved.
  - `test_parse_accepts_dj_if_with_leading_whitespace` — comments with surrounding whitespace tolerated via `trim()`.
  - `test_parse_accepts_dj_if_with_tab_separator` — tab separator works.
  - `test_parse_rejects_dj_if_with_no_separator` — `dj-ifid="if-0"` (no space) rejected.
  - `test_parse_accepts_prefixed_id_in_marker` — `if-<8hex>-N` shape passes through.
  - `test_parse_rejects_close_marker_lookalikes` — `/dj-iffy`, `/dj-if-extra`, `/dj-if-stuff` all rejected (parameterized via for-loop).

- **Evidence**: `crates/djust_vdom/src/parser.rs:909-1008` test names + assertions; `cargo test -p djust_vdom --lib parse_` confirms 22 PASS including all 7 new cases.

## New concerns (none material)

1. **Test-count drift in CHANGELOG (informational, not a Re-Review blocker)**: CHANGELOG line 69 claims "30 cases in `crates/djust_templates/tests/test_if_markers.rs`" but the actual `#[test]` count is 26 (19 main + 4 cross_template_ids + 3 csrf). The breakdown "15+4+3+8" sums to 30 but the fourth bucket ("8 stability/ordering") overcounts the 5-6 actual tests in that category. Not a correctness issue — flagging per Action #1049 (test-count recount after fix-pass deltas). If this triggers a follow-up CHANGELOG-only commit, the actual breakdown is 13+4+3+6=26.

2. **CSRF + else edge case not explicitly tested**: `{% if x %}{% csrf_token %}{% else %}<div>...</div>{% endif %}` isn't covered by a dedicated test. Not a real gap — the renderer at `renderer.rs:532-537` wraps in markers when EITHER branch is element-bearing, and both branches in this composition are element-bearing. The CSRF classifier fix flows through that condition automatically. Existing tests on `node_is_element_bearing` + the if-else marker emission tests cover it indirectly.

## Edge cases verified

- Cross-template extends: parent+child each get distinct prefixes; 4 ifs render to 4 unique IDs.
- Cross-template include (1-level + 3-level chain): each independently-loaded template gets its own prefix.
- Stability: same source → byte-identical IDs across renders.
- `apply_block_overrides` clones `marker_id` (`inheritance.rs:120-129`) — block override merges preserve uniqueness.
- All 25 `Node::*` variants exhaustively classified — no other CSRF-style misclassifications.
- Comment-marker safety: `id` value matches `^if-[0-9a-f]{8}-\d+$` — no `--`, `>`, or quote chars that could close the comment early.
- `strip_dj_if_markers` regex `<!--/?dj-if(?:\s[^>]*)?-->` correctly skips lookalikes (`dj-iffy` not matched because no whitespace separator).
- `extract_template_variables` (`parser.rs:1105`) uses bare `parse()` but doesn't render markers — JIT serialization doesn't depend on prefixes.
- Two-commit shape verified: `daa5b413` (impl + tests, 14 files, no CHANGELOG); `cb11a6ef` (CHANGELOG-only). Conforms to Action #181 / #1173.
- CI: 12 of 13 checks green. The only failure is `playwright-tests` (Auto-Save Functionality) — known pre-existing flake from PR #1359, unrelated to dj-if marker work.

## Recommendation

APPROVE. The 2 must-fix findings are fully addressed with comprehensive test coverage; the 2 should-fix items are likewise resolved with regression guards. No new must-fix issues introduced. Tests are non-tautological (verified the `parse_with_source` symbol doesn't exist on `main` so the cross-template tests cannot be a no-op rerun). The minor CHANGELOG test-count drift can be cleaned up in a future docs PR without blocking this milestone.

PR #1364, opened 2026-05-05
feat(client): RemoveSubtree + InsertSubtree patch types (Foundation 2 of #1358)

johnrtipton:
## Summary

Iter 2 of 3 toward closing #1358. **Zero-observable-behavior**: server doesn't emit these patch types yet (Iter 3 adds that), but the client now knows how to handle them when they arrive.

### RemoveSubtree
- **Wire**: `{type: "RemoveSubtree", id: "if-<prefix>-N"}`
- **Handler**: locates `<!--dj-if id="<target>"-->` open marker via `TreeWalker(NodeFilter.SHOW_COMMENT)`, walks forward depth-counting opens/closes, removes the entire bracketed range (markers + inner content) inclusive.
- **Nesting**: handled by depth counter — outer remove swallows inner pairs; inner remove leaves outer intact.

### InsertSubtree
- **Wire**: `{type: "InsertSubtree", id: "...", html: "<!--dj-if-->...<!--/dj-if-->", path: [...], index: N, d: <parent dj-id?>}`
- **Shape A** chosen: server emits the FULL marker pair + content as `html`; client inserts at `parent[index]` using the same path/d resolution other child-targeting patches use.
- **Inert script tags**: HTML fragment parsed via `<template>.innerHTML`, so `<script>` tags inside don't execute (verified by test).

## Why Shape A (not B)

Shape A (full marker pair in `html`) is simpler than Shape B (client maintains placeholder pairs even when conditional is false). It composes cleanly with `RemoveSubtree` for flips in either direction, and Iter 1 already emits the open-marker for both true/false branches but that's not load-bearing for this dispatcher — the differ in Iter 3 picks the patch type based on what's in old vs new VDOM.

## What this enables (NOT in this PR)

- **Iter 3 (Capability)**: Rust VDOM differ recognizes `dj-if` boundaries from Iter 1; emits these patch types when conditionals flip.

## Test plan

- [x] 25 regression cases in `tests/js/dj_if_subtree_patches.test.js` covering helpers (5+5), `RemoveSubtree` (7), `InsertSubtree` (5), dispatch wiring (3)
- [x] Bundle rebuilds cleanly (eslint --max-warnings 0 green per #1351)
- [x] All 1559 existing JS tests pass — confirms zero-observable-behavior
- [x] All 4947 Python tests pass — sanity check (no Python changes)
- [x] Inert-script test verifies `<template>` parsing keeps `<script>` tags from executing

- **Commit 1** (impl + tests + bundle): `ddcb23bd`
- **Commit 2** (CHANGELOG): `01eb3293`

Foundation 2 of 3 toward #1358.


johnrtipton (commented):
# Stage 11 Code Review: PR #1364 (Foundation 2 of #1358 — client patch types)

## Verdict

**APPROVE** — clean Iter-2 foundation with disciplined two-commit shape, complete nesting tests via depth counter, inert-template parsing for script safety, and zero-observable-behavior boundary maintained. CI is the known DraftMode playwright flake (also failed on Iter 1's #1363; pre-existing on main).

## 🔴 Must-fix findings

None.

## 🟡 Should-fix findings

None for this PR. Two notes flagged for Iter 3 (NOT this PR):

1. **`_removeDjIfBracketedRange` lacks a same-parent invariant check** (`12-vdom-patch.js:1267-1281`). The function walks `openMarker.nextSibling` until `cursor === closeMarker`. If a malformed server emit ever produces a close in a different parent, `cursor` becomes `null` (end of siblings) without ever hitting the close-marker break, and the function silently removes openMarker plus all its trailing siblings while leaving the close marker intact in some other subtree. Per the Iter 1 parser contract, balanced pairs are emitted in the same parent — so this is upstream-impossible — but a one-line `console.warn` if `cursor === null` before reaching `closeMarker` would be cheap defense-in-depth. **Park for Iter 3** since the differ that emits these patches is not yet written.

2. **TreeWalker scope is `document.body` for `_findDjIfCloseMarker`, ignoring the passed `rootEl`** (`12-vdom-patch.js:1234-1245`). The function uses `openMarker.ownerDocument.body` instead of the `rootEl` that `applyRemoveSubtree` propagates from `applySinglePatch`. Correctness is unaffected because the depth counter handles cross-subtree noise correctly, and the close marker is always within the open's parent (per parser contract, well within any reasonable `rootEl`). Performance impact is negligible for typical pages. **Park for Iter 3** when real workloads can measure scan cost.

## 🟢 Strong points

1. **Iter 1 helper reuse is clean and contractually correct.** `_findDjIfOpenMarker` calls `isDjIfComment(text)` (line 1216) — no logic duplication. `_findDjIfCloseMarker` uses the same boundary-form predicate (`startsWith('dj-if ') || startsWith('dj-if\t')`) that mirrors Iter 1's `isDjIfComment` shape — no drift risk. The CHANGELOG explicitly cites this reuse, and the code matches.

2. **Depth counter handles arbitrary nesting.** `_findDjIfCloseMarker` starts depth=1 (the open we're matching), increments on each new `dj-if<space-or-tab>` open, decrements on each `/dj-if`, returns when depth reaches 0. Two complementary nesting tests (`tests/js/dj_if_subtree_patches.test.js:99-118` outer-finds-outer-close, `:120-133` inner-finds-inner-close) lock the contract; the dispatch tests (`:190-207` outer-removes-inner, `:209-236` inner-leaves-outer) exercise it end-to-end. The legacy bare `dj-if` placeholder is correctly **not** counted as an open (`:135-151`), matching Iter 1's parser contract.

3. **InsertSubtree uses inert `<template>` parsing.** `_parseSubtreeHtml` constructs a `<template>` element via `document.createElement('template')` then `.innerHTML = html`. Per HTML spec, `<template>.content` is a separate `DocumentFragment` whose children do **not** execute scripts even when inserted into the live DOM. Test `:328-351` verifies this empirically with a sentinel — `<script>window.__inertSentinel = true;</script>` lands in the DOM but does not execute. This is the correct defense-in-depth posture: server is the trust boundary, but if a malicious or buggy template ever emits a script tag, it stays inert.

4. **Two-commit shape is exemplary** (Action #181). `git show ddcb23bd --name-only | grep CHANGELOG` returns nothing — impl commit clean. `git show 01eb3293 --name-only | grep -v CHANGELOG` returns nothing — docs commit is CHANGELOG-only. Bundle (client.js + .min variants) is in commit 1 alongside the source change, which is correct.

5. **Graceful-failure dispatch wiring** (`:1518-1528`). The subtree short-circuit is wrapped in `try/catch` → `console.error` + `return false`, so a thrown handler error does not abort the surrounding patch loop. Combined with the explicit `return false` on missing-id / missing-html / not-found / unresolvable-parent, the dispatcher cannot stuck the patch pipeline. Test `:416-419` verifies the unknown-id case via `applySinglePatch` returns false (no throw).

6. **Test discipline (Action #1196 + #1200) holds across all 25 cases.** Spot-check: `RemoveSubtree` tests (`:159-176, :178-188, :190-207, :209-236`) all assert post-state DOM (children counts, getElementById null/non-null, marker text equality), not just "no throw". `InsertSubtree` tests (`:274-307, :309-326, :328-351`) assert inserted content presence + position + child node sequence. The `applySinglePatch` dispatch tests (`:383-419`) exercise both happy-path and unknown-id paths. None of the assertions would pass on `main` (which lacks these dispatch cases entirely) — no tautologies, no rubber-stamp diffs.

7. **Test count claim verified.** CHANGELOG says `25 regression cases in tests/js/dj_if_subtree_patches.test.js`; `grep -c '    it('` returns 25; pre-push `scripts/check-changelog-test-counts.py` will pass. No drift on Iter 1's adjacent claim either (`dj_if_comment_predicate.test.js` still has 20 cases as claimed).

8. **Zero-observable-behavior boundary maintained.** Server doesn't emit these patch types yet, so this PR cannot regress production on its own — Iter 3 is the activation step. CHANGELOG explicitly documents this with "**zero-observable-behavior**" framing.

## Specific answers

### Marker location robustness

a. **Nested handling**: correct. `_findDjIfCloseMarker` increments depth on `dj-if ` / `dj-if\t` opens, decrements on `/dj-if`, returns when depth reaches 0. Tests `:99-118` and `:120-133` lock both directions (outer-from-outer, inner-from-inner). Tests `:190-207` and `:209-236` confirm the end-to-end DOM state matches.

b. **Marker-not-found graceful**: correct. `applyRemoveSubtree` calls `console.warn` with sanitized id and `return false`. The dispatch wrapper (`:1519-1527`) wraps in try/catch so even an unexpected throw becomes `console.error` + `return false`, never propagating out. Tests `:238-243` and `:416-419` cover both direct and dispatched paths.

c. **TreeWalker scope**: starts from `rootEl || document.body` for the open-marker search (`:1209-1210`). For two distinct pairs sharing the same id (which Iter 1's prefix scheme prevents but could happen during transition), the TreeWalker walks document order and returns the FIRST one found. That's deterministic and acceptable; in practice the prefix collision-resistance of Iter 1 makes this a non-issue.

d. **Performance**: O(n) over comment nodes per find; O(n × patches) per frame. For typical demo-project pages this is fine; for pages with many `{% if %}` blocks under sustained patch load, a `Map<id, Comment>` cache populated at insert time and cleared on remove would be O(1). Not worth doing for Iter 2 — wait until Iter 3 has real workloads to measure first.

### Iter 1 helper reuse

Verified. `_findDjIfOpenMarker` calls `isDjIfComment(text)` directly (line 1216). `_findDjIfCloseMarker` uses the exact same boundary-form predicate (`startsWith('dj-if ') || startsWith('dj-if\t')`) that mirrors Iter 1's `isDjIfComment`. The marker-shape contract — opening `dj-if id="..."`, closing `/dj-if`, legacy bare `dj-if` ignored in pairing — is consistent across Iter 1 emit and Iter 2 parse. No drift.

### InsertSubtree HTML parsing safety

Inert path verified at `:1340-1344`:

```js
function _parseSubtreeHtml(html) {
    const tpl = document.createElement('template');
    tpl.innerHTML = String(html || '');
    return tpl.content;
}
```

This is the textbook inert-parse pattern. Children of `<template>.content` belong to a separate `DocumentFragment` whose script elements remain inert even when the fragment is later inserted into the live DOM. Test `:328-351` verifies with a sentinel: the script tag lands in the DOM but `window.__inertSentinel` stays `false`. `<base>` and `<link>` would also be inert in this context (they're parsed into the inert document); even `<style>` would parse but not apply until insertion (and even then, well-behaved). Defense-in-depth complete; the server's trust-boundary status is preserved as a primary guarantee with the inert parse as backup.

### extractDjIfMarkerId

a. **Whitespace robustness**: `/id="([^"]+)"/` is permissive (multi-space / extra attributes don't break it). Combined with the `isDjIfComment` gate upstream, only well-formed boundary-open comments reach this regex.

b. **Quote handling**: regex requires double quotes (`"..."`); server's `parser.rs` emits double quotes per Iter 1 contract. Single-quote variants would not match — acceptable since the contract is fixed.

c. **Returns null on close marker**: verified. `/dj-if` has no `id="..."` substring; `match` is `null`; helper returns `null`. Test `:57-59`.

d. **Returns null on legacy bare placeholder**: verified. `dj-if` (no space) doesn't pass `isDjIfComment.startsWith('dj-if ')` upstream (it's matched by exact equality), but even if called directly, `_extractDjIfMarkerId('dj-if')` returns `null`. Test `:53-55`.

### Test discipline

Spot-checked 4 RemoveSubtree tests (`:159, :190, :209, :245`) and 4 InsertSubtree tests (`:274, :309, :328, :353`). All assert post-state DOM, not just "no throw". The 3 dispatch tests (`:383, :398, :416`) cover happy-path RemoveSubtree, happy-path InsertSubtree, and the graceful-failure case. The unknown-id dispatch test (`:416-419`) verifies that an unrecognized RemoveSubtree id returns `false` without throwing — pairing with the existing default case in `applySinglePatch` (`:1793-1797`), unknown patch types overall remain gracefully handled. No tautologies; no test would pass on `main` (which lacks these dispatch cases).

### Two-commit shape

```
ddcb23bd impl + tests + bundle: clean (no CHANGELOG.md)
01eb3293 docs(changelog): clean (only CHANGELOG.md)
```

Action #181 satisfied. Bundle (client.js, client.min.js, client.min.js.br/.gz/.map) is in the impl commit, which is the right place for it.

### Eslint compliance

Clean. `git show ddcb23bd -- python/djust/static/djust/src/12-vdom-patch.js | grep eslint-disable` returns no hits. Local `npx eslint --max-warnings 0 python/djust/static/djust/src/12-vdom-patch.js` exits 0 with no output. Action #1351's enforcement is satisfied.

## Edge cases considered

- Patch is `null` or `undefined` — guarded by `if (patch && (patch.type === ...))` at the dispatch entry.
- Patch missing `id` — `applyRemoveSubtree` returns `false` after `console.warn` (`:1310-1314`); test `:263-266`.
- Patch missing `html` — `applyInsertSubtree` returns `false` (`:1352-1355`); test `:353-361`.
- Open marker not found — `applyRemoveSubtree` returns `false` after `console.warn` (`:1316-1319`); test `:238-243`.
- Close marker not found (malformed open with no close) — `_findDjIfCloseMarker` returns `null`; `applyRemoveSubtree` returns `false` (`:1321-1325`).
- Parent cannot be resolved (deep unresolvable path) — `applyInsertSubtree` returns `false` (`:1364-1369`); test `:363-375`.
- Parent is non-element (text/comment node) — gracefully skipped (`:1370-1374`).
- Index out of range — falls back to `appendChild` via `(refChild = ... || null) ? insertBefore : appendChild` (`:1379-1383`); test `:309-326`.
- Script tag inside fragment — inert via `<template>`; test `:328-351`.
- Subtree handler throws unexpectedly — outer try/catch in `applySinglePatch` (`:1519-1527`) catches and returns `false`, so the patch loop continues with subsequent patches.
- Unknown patch type entirely — falls through to the existing `default:` case (`:1793-1797`); `console.warn` and `return false`.

## CI status

- `javascript-tests`: SUCCESS
- `Analyze JavaScript` (CodeQL): SUCCESS
- `CodeQL`: SUCCESS
- `Detect Security Hot Spots`: SUCCESS
- `Analyze Python`: SUCCESS
- `security-tests`, `security-scan`: SUCCESS
- `python-tests` (3.12, 3.13, 3.14), `rust-tests`: IN_PROGRESS
- `playwright-tests`: FAILURE (known DraftMode flake — pre-existing on main; same failure on Iter 1's #1363; not introduced by this PR)

The DraftMode failure is documented as the one accepted known flake; main is otherwise green and this PR's changes are entirely client-JS dispatch-handler additions with no Python/Rust impact.

## Summary

Foundation 2 lands cleanly. Iter 1's emit contract and Iter 2's parse helpers are aligned (no drift), nesting is correct via depth counter, script safety is enforced via inert template parsing, dispatch wiring is graceful on every failure mode, and the two-commit + zero-observable-behavior discipline is exemplary. Iter 3 (Rust differ) can now wire `RemoveSubtree` / `InsertSubtree` emission against this dispatcher with no further client-side changes required.

Approving.

Issue #1199, opened 2026-04-28
tech-debt: edge-case coverage for index/cursor logic


johnrtipton:
Source: Retro v0.9.4 / PR #1193

When implementing a handler with index or cursor logic, run through the cases at `index=0`, `index=len/2`, `index=len-1`, `index=len` (out of range) before declaring done. Four mental cases catch most off-by-one classes.

**Failure mode**: silent. Off-by-one bugs at the boundaries pass the happy-path tests; only manifest when the user is actually at the boundary.

**Observed instance**: PR #1193's `_build_time_travel_state` and `handle_forward_replay` answered the same boolean question ("does replay from this position fork?") with different formulas — `cursor < history_len` vs `from_index < history_len_before - 1`. They disagreed at `cursor=len-1, which="before"` with override_params: ack frame said `forward_replay_enabled=True` but the gate said no branch. Stage 11 caught it via mental trace.

Canonicalize as a CLAUDE.md rule: every handler with index logic gets at least one test at each boundary (0, mid, len-1, out-of-range).

johnrtipton:
Shipped in PR #1225 (v0.9.4 retro process canon batch in CLAUDE.md). Comma-list `Closes #...` form in PR body parses only the first ref — the rest had to be closed manually (this confirms #1184's diagnosis once again; the PR-checklist convention says one Closes per body line). 🤖 v0.9.1-6 drain bucket — final issue closure cleanup.

PR #1365, opened 2026-05-05
feat(vdom): keyed diff for {% if %} conditional subtrees (Capability of #1358; closes #256 Option A)

### LINKED ISSUE #256: VDOM patching breaks with {% if %} blocks that add/remove DOM elements

### LINKED ISSUE #1358: enhancement: keyed VDOM diffing for conditional subtrees ({% if %} structural changes) — re-open of #256 Option A

johnrtipton:
## Summary

Iter 3 of v0.9.4-1 — **the iter that actually fixes #1358 / #256 Option A.**
Wires up the foundations from Iter 1 (PR #1363, template markers) and Iter 2
(PR #1364, client patch types) so the Rust VDOM differ recognizes
`<!--dj-if id="if-<prefix>-N"-->...<!--/dj-if-->` boundary markers as
**keyed units** and emits `RemoveSubtree { id }` / `InsertSubtree { id, html, ... }`
patches when conditionals flip.

After this PR, the long-standing class of `{% if %}`-breaks-VDOM-patching
bugs that has plagued djust for over 3 months is **eliminated**. The
17.5%-error-rate tab-switch regression in NYC Claims (cited in #1358's
body) no longer reproduces.

## Changes

- **New patch variants** in `crates/djust_vdom/src/lib.rs`:
  - `Patch::RemoveSubtree { id: String }`
  - `Patch::InsertSubtree { id: String, path: Vec<usize>, d: Option<String>, index: usize, html: String }`
  - Both use existing `#[serde(tag = "type")]` discriminator → wire shape matches
    Iter 2's locked contract: `{type: "RemoveSubtree", id}` and
    `{type: "InsertSubtree", id, html, path, d?, index}`.

- **New differ pre-pass** in `crates/djust_vdom/src/diff.rs`
  (`dj_if_pre_pass`) runs at `diff_children` entry. It:
  1. Scans both sibling lists for `<!--dj-if id="X"-->` open markers
     and matches them with `<!--/dj-if-->` closes via depth counter
     (handles nested boundaries — Iter 1's `if/elif/else` cascades nest
     `If(B)` inside outer's `false_nodes`, producing nested marker pairs).
  2. Pairs boundaries by id:
     - id-only-in-old → `RemoveSubtree`
     - id-only-in-new → `InsertSubtree` (html = full marker pair + body)
     - id in both → recurse into the inner body element-by-element
       (markers stay static — text matches by construction)
  3. Diffs non-boundary children by **relative position among non-boundary
     siblings** — boundary shifts no longer cascade into mis-targeted
     patches in surrounding siblings (the core fix).
  4. Returns `None` when neither list has id-bearing boundaries → caller
     falls through to the existing keyed/indexed diff (preserves backward
     compat with legacy bare `<!--dj-if-->` placeholder at #295 and the
     `data-djust-replace` path).

- **Boundary detection helpers** mirror the parser predicate at
  `crates/djust_vdom/src/parser.rs:494-499` and the JS predicate at
  `python/djust/static/djust/src/12-vdom-patch.js:38-43`:
  `dj_if_open_id`, `is_dj_if_close`, `find_dj_if_pairs` (depth-counter,
  bails to empty on unbalanced markers — defensive).

- **Position-bypass invariant**: the id key normalizes positions:
  matched-boundary inner pairing is by relative position WITHIN the body,
  and non-boundary siblings are paired by relative position AMONG non-
  boundary siblings — adding/removing a boundary doesn't cascade into
  mis-aligned patches.

- **Test discipline (Action #1196)**: 14 new regression cases in
  `crates/djust_vdom/tests/test_dj_if_keyed_diff_1358.rs`:
  1. #1358 reproducer (tab-switch, different conditional branches)
  2. Conditional flip OFF → `RemoveSubtree` only
  3. Conditional flip ON → `InsertSubtree` only
  4. Same-id, inner text change → recurse with `SetText`
  5. Same-id, identical inner → 0 patches
  6. Nested boundaries — inner flips, outer stays
  7. Sibling-shift regression (the original failure mode)
  8. Empty boundary same id → 0 patches
  9. Empty boundary different ids → Remove + Insert
  10. Wire format: `RemoveSubtree` JSON shape
  11. Wire format: `InsertSubtree` JSON shape (all fields)
  12. Wire format: `d` omission when None
  13. Backward compat: legacy bare `<!--dj-if-->` (issue #295) still uses
      legacy diff path
  14. End-to-end via `parse_html` — proves parser + differ predicates agree

  Each case must FAIL on `main` (variants `Patch::RemoveSubtree` /
  `Patch::InsertSubtree` don't exist there).

## Backward compatibility

- Apps using `d-none` workaround documented in CLAUDE.md: continue to work
  identically (the workaround sidesteps `{% if %}` entirely).
- Apps using legacy bare `<!--dj-if-->` placeholder (#295): take the
  existing diff path unchanged (no id → not picked up by new pre-pass).
- The pre-pass only fires when at least one sibling list has an id-bearing
  boundary — full opt-in.

All 219 djust_vdom tests pass (14 new + 205 existing). All 4723 Python
tests pass. All 1559 JS tests pass.

## Out of scope

- Wholesale-replace heuristic for same-id matched boundaries (when inner
  content differs by >X%) — deferred to v0.10 polish.
- LIS within boundary bodies — deferred.
- Relaxing `d-none` workaround documentation in CLAUDE.md / downstream
  repos — separate docs PR.

## Limitation noted

When non-boundary siblings carry `dj-key` attributes AND reorder within
their relative slot, the position-based pairing of non-boundary children
can produce suboptimal patches. Production templates don't typically
reorder elements across `{% if %}` boundaries; if a regression surfaces,
the pre-pass can be extended to delegate non-boundary children to
`diff_keyed_children` when any of them have keys.

## Algorithm shape rationale

Recurse-always for matched ids (simple, correct). Wholesale-replace
heuristic was considered but deferred to v0.10 polish — recurse-always
is the safe default and fixes the bug.

- Commit 1 (`ea6c4c4a`): impl + tests, NO CHANGELOG
- Commit 2 (`c7c6203a`): CHANGELOG only

## Test plan

- [x] All 219 `cargo test -p djust_vdom` tests pass
- [x] All 4723 Python tests pass
- [x] All 1559 JS tests pass
- [x] `cargo fmt --check` clean
- [x] `cargo clippy --workspace --tests` clean
- [x] `ruff check` clean
- [x] Pre-commit hooks pass on both commits
- [ ] Stage 11 reviewer confirms algorithm correctness for if/elif/else
      cascades (Iter 1's nested `Node::If` in `false_nodes`)
- [ ] Stage 11 reviewer confirms backward compat with `d-none` workaround
      and legacy bare `<!--dj-if-->` placeholder
- [ ] Stage 11 reviewer confirms wire format matches Iter 2's locked
      contract exactly

Closes the capability half of v0.9.4-1 milestone (Iter 3 of 3).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -126,6 +126,99 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   VDOM differ recognizes `dj-if` boundaries from Iter 1; emits these
   patch types when conditionals flip.
 
+- **Keyed VDOM diff for `{% if %}` conditional subtrees (#1358; closes
+  #256 Option A; capability of v0.9.4-1).** Iter 3 of 3 — the iter that
+  actually fixes the bug. After this PR, the long-standing class of
+  `{% if %}`-breaks-VDOM-patching bugs that has plagued djust for over
+  3 months is **eliminated**. The Rust VDOM differ now recognizes
+  `<!--dj-if id="if-<prefix>-N"-->...<!--/dj-if-->` boundary markers
+  (emitted by Iter 1 template renderer, PR #1363) as KEYED units in the
+  diff algorithm.
+
+  When conditionals flip, the differ emits the new patch types from
+  Iter 2 (PR #1364):
+  - **OLD has boundary id=X, NEW does not** → `RemoveSubtree { id: X }`.
+    Client locates the marker pair by id (NOT by position) and removes
+    the bracketed range.
+  - **NEW has boundary id=Y, OLD does not** → `InsertSubtree { id: Y,
+    path, d, index, html }`. Client parses the full marker-pair HTML
+    (Shape A) via inert `<template>.innerHTML` and inserts at the
+    parent / index resolved via the same path/d resolution other
+    child-targeting patches use.
+  - **Both have boundary id=Z** → recurse into the inner body element-
+    by-element (markers themselves are static comments — text matches
+    by construction). Standard intra-subtree diff fires for the inner
+    content (SetText, SetAttr, InsertChild, etc.); NO subtree-replace.
+
+  Position-based path tracking is BYPASSED within boundaries: the id
+  normalizes positions, so adding or removing a boundary no longer
+  cascades into mis-targeted patches in surrounding siblings. Non-
+  boundary siblings are paired by relative position AMONG non-boundary
+  siblings — the conditional's presence/absence doesn't shift their
+  relative order.
+
+  The 17.5%-error-rate tab-switch regression in NYC Claims (cited in
+  #1358's body) no longer reproduces. The recovery-HTML / page-reload
+  fallback path is no longer triggered by `{% if %}` flips.
+
+  **Backwards-compatible:** Apps using the `d-none` workaround
+  documented in CLAUDE.md and downstream repos continue to work
+  identically — the workaround sidesteps `{% if %}` entirely. Apps
+  using the legacy bare `<!--dj-if-->` placeholder for false-no-else
+  conditionals (issue #295) take the existing diff path unchanged
+  (those placeholders have NO id and don't trigger the new keyed
+  pre-pass). The pre-pass only fires when at least one sibling list
+  contains an id-bearing boundary marker.
+
+  **Implementation in `crates/djust_vdom/src/diff.rs`:**
+  - New helpers: `dj_if_open_id`, `is_dj_if_close`, `find_dj_if_pairs`
+    (depth-counter for nested boundaries), `render_dj_if_boundary_html`
+    (serializes boundary slice for `InsertSubtree.html`),
+    `build_excluded_mask`.
+  - New `dj_if_pre_pass` runs at `diff_children` entry; returns
+    `Some(patches)` when boundaries are present (caller short-circuits
+    its keyed/indexed diff), `None` otherwise (caller proceeds
+    unchanged).
+  - Predicates mirror parser-side at
+    `crates/djust_vdom/src/parser.rs:494-499` and JS-side at
+    `python/djust/static/djust/src/12-vdom-patch.js:38-43`.
+
+  **Wire format** (locked by Iter 2):
+  - `{type: "RemoveSubtree", id: "if-<prefix>-N"}`
+  - `{type: "InsertSubtree", id: "...", path: [...], d: "<parent dj-id?>",
+    index: N, html: "<!--dj-if id=...-->...<!--/dj-if-->"}`
+
+  Limitation noted in code: when non-boundary siblings carry `dj-key`
+  attributes AND reorder within their relative slot, the position-based
+  pairing of non-boundary children can produce suboptimal patches.
+  Production templates don't typically reorder elements across
+  `{% if %}` boundaries; if a regression surfaces, the pre-pass can be
+  extended to delegate non-boundary children to `diff_keyed_children`
+  when any of them have keys.
+
+  Out of scope (deferred to v0.10): wholesale-replace heuristic for
+  same-id matched boundaries (e.g., when inner content differs by >X%);
+  LIS within boundary bodies; relaxing `d-none` workaround
+  documentation in CLAUDE.md / downstream repos.
+
+  Regression suite: 14 new cases in
+  `crates/djust_vdom/tests/test_dj_if_keyed_diff_1358.rs` covering
+  the #1358 reproducer (tab-switch with different conditional
+  branches), conditional flip-off, conditional flip-on, same-id inner
+  text change (recurses, NOT subtree replace), same-id identical inner
+  (0 patches), nested boundaries (inner flip leaves outer alone),
+  sibling-shift regression (the original failure mode), empty boundary
+  same id (0 patches), empty boundary different ids (Remove + Insert),
+  JSON wire-format checks for `RemoveSubtree` / `InsertSubtree` /
+  `d` omission when None, backward compat with legacy bare `<!--dj-if-->`
+  placeholder, and end-to-end via `parse_html` (proves parser-side and
+  differ-side predicates agree). All 219 djust_vdom tests pass (14 new +
+  205 existing). All 4723 Python tests pass. All 1559 JS tests pass.
+
+  Closes the capability half of v0.9.4-1 milestone (Iter 3 of 3).
+  Foundation 1: PR #1363 (template markers). Foundation 2: PR #1364
+  (client patch types).
+
 ### Changed
 
 - **Bundled `client.js` and `debug-panel.js` are now eslint-clean (#1351).**
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -145,10 +145,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
     (Shape A) via inert `<template>.innerHTML` and inserts at the
     parent / index resolved via the same path/d resolution other
     child-targeting patches use.
-  - **Both have boundary id=Z** → recurse into the inner body element-
-    by-element (markers themselves are static comments — text matches
-    by construction). Standard intra-subtree diff fires for the inner
-    content (SetText, SetAttr, InsertChild, etc.); NO subtree-replace.
+  - **Both have boundary id=Z** → recurse into the inner body via
+    `dj_if_pre_pass_inner`. The recursion handles arbitrary nesting
+    cleanly, including `{% if %}/{% elif %}/{% else %}` cascades where
+    the outer marker is matched in both OLD and NEW but the body
+    introduces (or removes) an inner boundary marker. Standard intra-
+    subtree diff fires for the inner content (SetText, SetAttr, etc.)
+    only when the body has NO nested boundaries.
 
   Position-based path tracking is BYPASSED within boundaries: the id
   normalizes positions, so adding or removing a boundary no longer
@@ -161,6 +164,27 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   #1358's body) no longer reproduces. The recovery-HTML / page-reload
   fallback path is no longer triggered by `{% if %}` flips.
 
+  **Recursive pre-pass for `{% if %}/{% elif %}/{% else %}` cascades
+  (Stage 11 finding on PR #1365 — capability iter).** The first
+  iteration of this fix iterated matched-id body children element-
+  by-element via `diff_nodes`, treating any nested boundary markers
+  as ordinary VNodes. That produced overlapping patches when a
+  cascade introduced or removed nested boundaries:
+  - Top-level step 2 emitted `InsertSubtree(B)` correctly.
+  - Top-level step 3 ALSO emitted `Replace` + `InsertChild` patches
+    for the same content (because element-by-element pairing saw B's
+    markers as ordinary comment nodes and B's content as new sibling).
+  - Both applied = corrupt DOM with duplicated content and mismatched
+    markers.
+
+  The recursive pre-pass closes this gap: when matched-id A's body
+  has nested boundaries, `dj_if_pre_pass_inner` recursively runs on
+  the body slice. Each recursion level handles only its OWN top-level
+  pairs (via the new `find_top_level_dj_if_pairs` helper), so nested
+  pairs are discovered at the recursion level that descends into
+  their containing boundary. No more overlap; no more duplicate
+  patches; arbitrary nesting (3+ levels) handled coherently.
+
   **Backwards-compatible:** Apps using the `d-none` workaround
   documented in CLAUDE.md and downstream repos continue to work
   identically — the workaround sidesteps `{% if %}` entirely. Apps
@@ -171,14 +195,21 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   contains an id-bearing boundary marker.
 
   **Implementation in `crates/djust_vdom/src/diff.rs`:**
-  - New helpers: `dj_if_open_id`, `is_dj_if_close`, `find_dj_if_pairs`
-    (depth-counter for nested boundaries), `render_dj_if_boundary_html`
-    (serializes boundary slice for `InsertSubtree.html`),
-    `build_excluded_mask`.
-  - New `dj_if_pre_pass` runs at `diff_children` entry; returns
-    `Some(patches)` when boundaries are present (caller short-circuits
-    its keyed/indexed diff), `None` otherwise (caller proceeds
-    unchanged).
+  - New helpers: `dj_if_open_id`, `is_dj_if_close`,
+    `find_top_level_dj_if_pairs` (depth-counter that returns ONLY
+    outermost pairs at the current slice level — nested pairs are
+    discovered when the recursion descends),
+    `render_dj_if_boundary_html` (serializes boundary slice for
+    `InsertSubtree.html`), `build_excluded_mask`.
+  - New `dj_if_pre_pass` runs at `diff_children` entry; delegates to
+    `dj_if_pre_pass_inner` which carries old/new offsets so absolute
+    parent-children indices propagate correctly across recursion
+    levels (DOM parent stays the same — markers don't create
+    container elements).
+  - Returns `Some(patches)` when boundaries are present (caller
+    short-circuits its keyed/indexed diff), `None` otherwise (caller
+    proceeds unchanged or, in the recursive case, falls back to
+    element-by-element pairing of the body slice).
   - Predicates mirror parser-side at
     `crates/djust_vdom/src/parser.rs:494-499` and JS-side at
     `python/djust/static/djust/src/12-vdom-patch.js:38-43`.
@@ -201,23 +232,38 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   LIS within boundary bodies; relaxing `d-none` workaround
   documentation in CLAUDE.md / downstream repos.
 
-  Regression suite: 14 new cases in
-  `crates/djust_vdom/tests/test_dj_if_keyed_diff_1358.rs` covering
-  the #1358 reproducer (tab-switch with different conditional
-  branches), conditional flip-off, conditional flip-on, same-id inner
-  text change (recurses, NOT subtree replace), same-id identical inner
-  (0 patches), nested boundaries (inner flip leaves outer alone),
-  sibling-shift regression (the original failure mode), empty boundary
-  same id (0 patches), empty boundary different ids (Remove + Insert),
-  JSON wire-format checks for `RemoveSubtree` / `InsertSubtree` /
-  `d` omission when None, backward compat with legacy bare `<!--dj-if-->`
-  placeholder, and end-to-end via `parse_html` (proves parser-side and
-  differ-side predicates agree). All 219 djust_vdom tests pass (14 new +
-  205 existing). All 4723 Python tests pass. All 1559 JS tests pass.
+  Regression suite: 19 cases in
+  `crates/djust_vdom/tests/test_dj_if_keyed_diff_1358.rs` covering:
+  two separate `{% if %}` blocks flipping (the renamed Case 1),
+  conditional flip-off, conditional flip-on, same-id inner text
+  change (recurses, NOT subtree replace), same-id identical inner
+  (0 patches), nested boundaries inside DOM elements (inner flip
+  leaves outer alone), sibling-shift regression with DIFFERENT
+  boundary span lengths (3 vs 1 inner children — exercises the
+  position-cascade class explicitly), empty boundary same id
+  (0 patches), empty boundary different ids (Remove + Insert),
+  JSON wire-format shape comparisons via `serde_json::Value` for
+  `RemoveSubtree` / `InsertSubtree` / `d` omission when None
+  (tightened from substring `contains` per Stage 11 finding),
+  backward compat with legacy bare `<!--dj-if-->` placeholder,
+  end-to-end via `parse_html` (proves parser-side and differ-side
+  predicates agree), and 5 NEW elif-cascade cases (Stage 11
+  finding on this PR): A → elif-B flip (cascade introduces nested
+  marker), elif-B → A flip (cascade collapses nested marker —
+  symmetric direction, SHOULD-FIX #4), A → else with double-nested
+  matched ids (no subtree-flip patches when both outer+inner ids
+  match), cascade with extra static siblings (footer SetText path
+  must use NEW tree's absolute index, not OLD's), 3-level cascade
+  (A → B → C nesting introduced atomically — proves recursive
+  pre-pass handles arbitrary depth).
+
+  All 19 dj-if keyed-diff tests pass. All Rust tests pass. All
+  Python tests pass. All 1559 JS tests pass.
 
   Closes the capability half of v0.9.4-1 milestone (Iter 3 of 3).
   Foundation 1: PR #1363 (template markers). Foundation 2: PR #1364
-  (client patch types).
+  (client patch types). Stage 11 must-fix and should-fix findings
+  from this PR's review addressed in commit on this same PR.
 
 ### Changed
 
--- a/crates/djust_vdom/src/diff.rs
+++ b/crates/djust_vdom/src/diff.rs
@@ -23,6 +23,350 @@
 use crate::{lis::longest_increasing_subsequence, vdom_trace, Patch, VNode};
 use ahash::{AHashMap as HashMap, AHashSet as HashSet};
 
+// =============================================================================
+// dj-if keyed boundary detection (Capability of #1358 — Iter 3 of v0.9.4-1)
+//
+// Iter 1 (PR #1363) made the Rust template renderer emit
+// `<!--dj-if id="if-<8hex>-N"-->...<!--/dj-if-->` boundary markers around
+// element-bearing `{% if %}` blocks. Iter 2 (PR #1364) taught the client
+// patch dispatcher to apply `RemoveSubtree { id }` / `InsertSubtree { id,
+// path, d, index, html }` patches.
+//
+// This iter's job: recognize those markers in the diff input and emit the
+// new patch types when conditionals flip — eliminating the long-standing
+// class of `{% if %}`-breaks-VDOM-patching bugs (issue #256, #1358).
+//
+// The predicate mirrors `crates/djust_vdom/src/parser.rs:494-499` (server)
+// and `python/djust/static/djust/src/12-vdom-patch.js:38-43` (client).
+// =============================================================================
+
+/// Returns the id token from a `<!--dj-if id="X"-->` open-marker comment, if
+/// the node is such a comment. Mirrors the JS `_extractDjIfMarkerId` helper
+/// in `12-vdom-patch.js:1180-1199` and the server-side parser predicate at
+/// `crates/djust_vdom/src/parser.rs:497-498`.
+///
+/// Returns `None` for:
+///   - non-comment nodes
+///   - comments not matching the dj-if open-marker shape
+///   - the legacy single-comment placeholder `<!--dj-if-->` (issue #295) —
+///     it has no id and is handled by the existing legacy diff path
+///     (`diff_nodes` lines ~165-241)
+///   - the close marker `<!--/dj-if-->`
+fn dj_if_open_id(node: &VNode) -> Option<String> {
+    if !node.is_comment() {
+        return None;
+    }
+    let text = node.text.as_deref()?;
+    let trimmed = text.trim();
+    // Open marker: `dj-if<space-or-tab>...id="X"...`. Must reject
+    // lookalikes like `dj-iffy`, `dj-ifid="x"` — only literal whitespace
+    // after `dj-if` qualifies (mirrors the parser broadening predicate).
+    if !(trimmed.starts_with("dj-if ") || trimmed.starts_with("dj-if\t")) {
+        return None;
+    }
+    // After `dj-if`, find an `id="..."` token. Production templates emit
+    // exactly one canonical form: `dj-if id="if-<8hex>-N"`. We extract the
+    // first `id=".."` substring. Returns None if absent (defensive — should
+    // not happen in production).
+    let after_dj_if = trimmed.get("dj-if".len()..)?;
+    let id_start = after_dj_if.find("id=\"")? + "id=\"".len();
+    let rest = after_dj_if.get(id_start..)?;
+    let id_end = rest.find('"')?;
+    Some(rest[..id_end].to_string())
+}
+
+/// True if the node is a `<!--/dj-if-->` close marker.
+fn is_dj_if_close(node: &VNode) -> bool {
+    if !node.is_comment() {
+        return false;
+    }
+    node.text
+        .as_deref()
+        .map(|t| t.trim() == "/dj-if")
+        .unwrap_or(false)
+}
+
+/// Find all dj-if boundary pairs in a sibling list.
+///
+/// Returns a vector of `(open_idx, close_idx, id)` tuples in document order.
+/// Uses a depth counter so nested boundaries are paired correctly:
+/// `[open_A, open_B, close_B, close_A]` → `[(0, 3, "A"), (1, 2, "B")]`.
+///
+/// Unbalanced markers (open without matching close, close without preceding
+/// open) are returned as the empty vector — the caller should fall through
+/// to the existing diff in that case (defensive against malformed input).
+fn find_dj_if_pairs(children: &[VNode]) -> Vec<(usize, usize, String)> {
+    let mut pairs: Vec<(usize, usize, String)> = Vec::new();
+    // Stack of (open_idx, id) for unmatched open markers seen so far.
+    let mut stack: Vec<(usize, String)> = Vec::new();
+    for (i, child) in children.iter().enumerate() {
+        if let Some(id) = dj_if_open_id(child) {
+            stack.push((i, id));
+        } else if is_dj_if_close(child) {
+            if let Some((open_idx, id)) = stack.pop() {
+                pairs.push((open_idx, i, id));
+            } else {
+                // Unbalanced close — bail to no pairs; existing diff handles it.
+                return Vec::new();
+            }
+        }
+    }
+    if !stack.is_empty() {
+        // Unbalanced open(s) — bail to no pairs.
+        return Vec::new();
+    }
+    pairs
+}
+
+/// Render a `dj-if` boundary subtree as HTML for `InsertSubtree.html`.
+///
+/// Includes the open marker, all body nodes, and the close marker — Shape A
+/// per Iter 2's wire contract (`12-vdom-patch.js:1345-1389`). The client
+/// parses this via an inert `<template>.innerHTML` so any nested `<script>`
+/// tags do NOT execute.
+fn render_dj_if_boundary_html(children: &[VNode], open_idx: usize, close_idx: usize) -> String {
+    let mut out = String::new();
+    for child in children.iter().take(close_idx + 1).skip(open_idx) {
+        out.push_str(&child.to_html());
+    }
+    out
+}
+
+/// Diff sibling lists with `dj-if` keyed-boundary awareness.
+///
+/// Returns `Some(patches)` when at least one of the lists contains an
+/// id-bearing dj-if boundary pair (i.e. the new keyed-subtree machinery
+/// applies). The caller short-circuits its existing keyed/indexed diff
+/// when this returns `Some` — the returned patches are the complete child
+/// diff for this parent.
+///
+/// Returns `None` when neither list has id-bearing boundaries; the caller
+/// then runs its regular keyed/indexed diff (preserving the legacy
+/// `<!--dj-if-->` placeholder + `data-djust-replace` paths unchanged).
+///
+/// Patch emission rules:
+///   - id in OLD only → `RemoveSubtree { id }`
+///   - id in NEW only → `InsertSubtree { id, path: parent_path, d: parent_id,
+///     index: open_idx_in_new, html: full marker pair + body }`
+///   - id in BOTH → recurse into the inner body element-by-element (markers
+///     themselves are static comment nodes — text matches by construction).
+///   - non-boundary children → paired by relative order among non-boundary
+///     siblings. Patch absolute indices are taken from the NEW tree (target).
+///
+/// The boundary id normalizes positions: shifts caused by adding/removing
+/// boundaries no longer cascade into mis-targeted patches in non-boundary
+/// siblings — the core fix for #1358.
+fn dj_if_pre_pass(
+    old: &[VNode],
+    new: &[VNode],
+    path: &[usize],
+    parent_id: &Option<String>,
+) -> Option<Vec<Patch>> {
+    let old_pairs = find_dj_if_pairs(old);
+    let new_pairs = find_dj_if_pairs(new);
+    if old_pairs.is_empty() && new_pairs.is_empty() {
+        // Common fast path: no id-bearing boundaries on either side. Caller
+        // proceeds with the existing diff unchanged.
+        return None;
+    }
+
+    let mut patches: Vec<Patch> = Vec::new();
+
+    // Build id → (open_idx, close_idx) maps for both lists.
+    let mut old_by_id: HashMap<String, (usize, usize)> = HashMap::new();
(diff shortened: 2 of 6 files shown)


johnrtipton (commented):
# Stage 11 Code Review: PR #1365 (Capability of #1358 - Rust VDOM differ)

## Verdict

**REQUEST_CHANGES** - Algorithm is broken for `{% if %}/{% elif %}` cascades when the elif branch is element-bearing. Same id appears in OLD and NEW (the outer marker), inner content shifts shape, and the matched-id body recursion does not re-detect nested boundaries. This produces overlapping patches that result in duplicated/mismatched DOM markers - the silent-bad-DOM-state failure mode worse than the original 500-class bug.

This is the iter that's supposed to close the 3-month-old bug. Shipping it as-is would replace one footgun (loud 500s on `{% if %}` flips) with a different footgun (silent broken DOM on elif cascades).

## Critical (must-fix) findings

### 1. (load-bearing) `if/elif/else` cascade with element-bearing branches produces broken patches

**File**: `crates/djust_vdom/src/diff.rs:233-256` (matched-id body recursion in `dj_if_pre_pass`)

**Reproducer** (verified by running locally):

```rust
// {% if A %}<body-A>{% elif B %}<body-B>{% endif %}
// A=true: <!--dj-if id="X-0"--><body-A><!--/dj-if-->
// B=true: <!--dj-if id="X-0"--><!--dj-if id="X-1"--><body-B><!--/dj-if--><!--/dj-if-->
let old = VNode::element("div").with_djust_id("root").with_children(vec![
    dj_if_open("X-0"),
    VNode::element("div").with_djust_id("body-a").with_child(VNode::text("A content")),
    dj_if_close(),
]);
let new = VNode::element("div").with_djust_id("root").with_children(vec![
    dj_if_open("X-0"),
    dj_if_open("X-1"),
    VNode::element("div").with_djust_id("body-b").with_child(VNode::text("B content")),
    dj_if_close(),
    dj_if_close(),
]);
let patches = diff_nodes(&old, &new, &[]);
```

**Actual output (4 patches, broken)**:
```
[0] InsertSubtree { id: "X-1", path: [], d: Some("root"), index: 1, html: "<!--dj-if id=\"X-1\"--><div>B content</div><!--/dj-if-->" }
[1] Replace { path: [1], d: Some("body-a"), node: <!--dj-if id="X-1"--> }
[2] InsertChild { path: [], d: Some("root"), index: 2, node: <body-b> }
[3] InsertChild { path: [], d: Some("root"), index: 3, node: <!--/dj-if--> }
```

**Why it's wrong** (trace):

`find_dj_if_pairs` correctly detects BOTH boundaries in the new sibling list: `[(1,3,"X-1"), (0,4,"X-0")]`. Because X-1 is only in NEW, step 2 emits `InsertSubtree(X-1, idx=1)`. Then step 3 (matched-id recursion) for X-0 also fires. The recursion at lines 246-256 does element-by-element pairing of `old_body=[<body-a>]` against `new_body=[<X-1 open marker>, <body-b>, <X-1 close marker>]`. It produces `Replace [1]` (body-a -> open marker comment) and InsertChild patches for the rest.

The `InsertSubtree(X-1)` patch at root (idx=1) AND the Replace+InsertChild patches at the same parent are **both trying to insert the X-1 content**. After applying InsertSubtree first, the DOM has `[open-X-0, open-X-1, body-b, close-X-1, body-a, close-X-0]`. Then Replace finds `body-a` by dj-id and replaces it with another open-X-1 comment. Then two more InsertChild patches add a duplicate body-b and a stray close-X-1. End state is a DOM with two open-X-1 markers, duplicated content, and orphaned closing markers - silently corrupt rather than rejecting with "patch failed".

**Fix direction (sketch)**: the matched-id body recursion in step 3 must re-run boundary detection on the body slice itself, not blindly element-by-element pair. Concretely: replace the `for i in 0..common { diff_nodes(...) }` loop with a recursive call to `dj_if_pre_pass(old_body, new_body, child_path, ?parent_id)` followed by fallback diff if it returns None. Otherwise the inner X-1 boundary is invisible to the keyed machinery.

**Coverage gap**: there is no test for this case in the new file. Test #1 uses two SEPARATE root-level boundaries with different ids (matching multi-`{% if %}` template patterns), not nested elif markers with a stable outer id - even though its docstring claims to test "tab-switch via `{% if %}/{% elif %}`". The renderer's actual elif emit shape (`crates/djust_templates/src/renderer.rs:486-540`) wraps with the outer `If(A)` marker_id regardless of which branch fires; the inner `If(B)` (in `false_nodes`) gets its own marker_id and only appears when A is falsy.

This is the case Action #1196 (delegated-listener integration tests) generalizes to: the unit assertions don't verify the wiring under all real input shapes the upstream produces. Adding a dedicated cascade test would have caught this.

### 2. (load-bearing) Test #1 docstring misrepresents what the test exercises

**File**: `crates/djust_vdom/tests/test_dj_if_keyed_diff_1358.rs:78-89, 95-115`

The docstring claims:
> `{% if active_tab == "overview" %} ... {% elif active_tab == "details" %} ... {% endif %}`. Iter 1 emits a distinct boundary id per branch (the marker_id is assigned in document order at parse time, and the renderer emits the id of the branch whose condition fired).

But this contradicts how `assign_if_marker_ids` (`crates/djust_templates/src/parser.rs:353-365`) and `Node::If` rendering (`crates/djust_templates/src/renderer.rs:486-540`) work. For an elif cascade:
- Parser produces `If(A){..., false_nodes: [If(B){...}]}`. Both `If` nodes get marker_ids.
- Renderer for `If(A)`: when `A` is true, body = render of `true_nodes` (no inner marker); when `A` is false, body = render of `false_nodes` which renders the inner `If(B)` with its own marker.
- Outer wrap (line 532-538) ALWAYS uses A's marker_id when A is element-bearing.

So when the elif cascade flips from A=true to B=true:
- OLD: `<!--dj-if id="X-0"-->BODY_A<!--/dj-if-->`
- NEW: `<!--dj-if id="X-0"--><!--dj-if id="X-1"-->BODY_B<!--/dj-if--><!--/dj-if-->`

The OUTER id `X-0` is stable. The INNER id `X-1` only appears in NEW.

What test #1 actually exercises is the case of TWO SEPARATE root-level `{% if %}` blocks (each rendering its own marker pair), only one of which fires per render:
```html
{% if active_tab == "overview" %}<tab-a>{% endif %}
{% if active_tab == "details" %}<tab-b>{% endif %}
```

This is a real and important pattern, and the test correctly covers it. But the docstring should be corrected to describe what's actually tested (separate `{% if %}` blocks), and a separate test for the elif cascade case must be added (and made to pass).

## Important (should-fix) findings

### 3. The "sibling-shift regression" test #7 doesn't fully exercise the position-shift class

**File**: `tests/test_dj_if_keyed_diff_1358.rs:455-574` (`test_sibling_shift_does_not_cascade`)

The test has identical-length boundaries in OLD and NEW (both 3-children: open, content, close). The "shape difference" is only the inner element's tag (`p` vs `section`), not the boundary's child count. The footer ends up at the same absolute index 4 in both trees, so the position-pairing produces correct paths "by coincidence."

The original failure mode is:
- OLD boundary occupies indices [1, 2] (open + content + close = 3 children, so old footer is at 4)
- NEW boundary occupies indices [1, 5] (open + 4 children + close = 6, so new footer is at 6)

The footer's ABSOLUTE index shifts. Pre-pass must emit the footer SetText with path `[6, 0]`, not `[4, 0]`. This is what `build_excluded_mask` + `old_kept`/`new_kept` is supposed to do - and it actually does. But the test doesn't verify the case where the boundary count actually differs and forces an absolute-index shift. Recommend adding a test where NEW's boundary spans more children than OLD's (e.g., a `{% for %}` inside the conditional in NEW's branch), with explicit assertions on the post-boundary sibling's path index.

### 4. RemoveSubtree + matched-id-body interaction when only inner boundary disappears

**File**: `diff.rs:185-219` (Steps 1 & 2 of `dj_if_pre_pass`) and `diff.rs:233-294` (Step 3)

Symmetric concern to finding #1. Consider: OLD = `<dj-if X-0><dj-if X-1>body-B</dj-if></dj-if>`, NEW = `<dj-if X-0>body-A</dj-if>` (elif's B was true, now A is true again).

old_pairs = `[(1,3,"X-1"),(0,4,"X-0")]`, new_pairs = `[(0,2,"X-0")]`.

Step 1 emits `RemoveSubtree(X-1)`. Step 3 recurses for matched X-0:
- old_body = `[open-X-1, body-B, close-X-1]` (3 elements)
- new_body = `[body-A]` (1 element)
- common = 1: pair (`open-X-1`, `body-A`) -> diff_nodes(comment, div) -> Replace
- Extra old: descending RemoveChild for indices 2, 3 (close-X-1 and... wait, close is at index 3 in OLD, abs_old_idx = 0+1+2 = 3 which is the close-X-0 marker, not the close-X-1 inside the boundary)

Actually I need to re-trace - the body slice excludes markers. `old_body = old[1..3]` = indices 1 and 2 in old, which are `open-X-1` and `body-B`. close-X-1 is at index 3 (which is also a body element when iterated relative to the outer X-0 boundary). Wait, no - find_dj_if_pairs finds X-1 at (1,3) and X-0 at (0,4). So old_body for X-0 = `old[1..4]` = `[open-X-1, body-B, close-X-1]`. The body of X-0 EXCLUDES X-0's open/close markers but INCLUDES X-1's open/close markers.

That means the body recursion sees TWO inner markers as ordinary elements. With body_old having 3 children and body_new having 1, common=1, pair (open-X-1, body-A) -> Replace. Extra old descending: RemoveChild abs_old=3 (close-X-1), RemoveChild abs_old=2 (body-B).

Plus the already-emitted `RemoveSubtree(X-1)` at the top. The client applies in some order:
- Sort: child-ops by parent path, RemoveSubtree probably last per Iter 2's dispatcher.
- If RemoveSubtree(X-1) runs first: client locates `<!--dj-if id="X-1"-->` in DOM and removes the bracketed range (open-X-1, body-B, close-X-1 - 3 nodes). Then Replace [1] hits a non-existent path or finds shifted-up content.
- If Replace runs first: replaces the open-X-1 comment with body-A (corrupt). Then RemoveChild patches target `body-B` and `close-X-1` (which still exist). Then RemoveSubtree(X-1) tries to find an X-1 open marker - there's none anymore (it was Replaced) - warns "open marker not found", returns false.

Either way, the patch sequence is internally inconsistent. Same root cause as finding #1: the matched-id body recursion treats inner boundaries as ordinary elements.

### 5. Wire-format JSON tests assert string-substring containment, not full shape

**File**: `tests/test_dj_if_keyed_diff_1358.rs:1207-1270` (`test_*_json_wire_format`)

The wire format tests use `json.contains("...")` substring checks. This passes if the right substrings are present but doesn't catch field ordering issues, extra/unexpected fields, or accidental extra escaping. Recommend asserting against a parsed `serde_json::Value` shape:

```rust
let value: serde_json::Value = serde_json::from_str(&json).unwrap();
assert_eq!(value["type"], "RemoveSubtree");
assert_eq!(value["id"], "if-abc-0");
assert_eq!(value.as_object().unwrap().len(), 2);  // exactly type + id
```

This locks the contract more tightly. (Not a correctness blocker - just a discipline gap.)

## Strong points

1. **Wire format matches Iter 2 exactly.** Field names `id`, `path`, `d`, `index`, `html` all line up with what `applyInsertSubtree` reads at `12-vdom-patch.js:1358-1389`. The `#[serde(skip_serializing_if = "Option::is_none")]` on `d` mirrors `InsertChild.d`, and test #11 (omits-d-when-None) verifies this.

2. **Boundary detection helpers are tight**: `dj_if_open_id` correctly rejects `dj-iffy`, `dj-ifid="x"`, `/dj-if id="x"` lookalikes via the `dj-if<space-or-tab>` prefix check (mirrors `parser.rs:497-499`). `is_dj_if_close` uses strict equality to `/dj-if`. `find_dj_if_pairs` uses depth-counter (correctly handles nested same-id and different-id pairs) and bails to empty on unbalanced markers (defensive).

3. **Backward compat for legacy bare `<!--dj-if-->` placeholder is preserved**: `dj_if_open_id` returns None for bare `dj-if` (no whitespace+id), so `find_dj_if_pairs` excludes it from the keyed machinery, `dj_if_pre_pass` returns None when no id-bearing boundaries are found, and the existing `<!--dj-if-->` -> content fallback at `diff.rs:513-552` handles the legacy case unchanged. Test #10 verifies this.

4. **`data-djust-replace` short-circuit is preserved**: line 729-738 checks `data-djust-replace` BEFORE the dj-if pre-pass, so containers using the existing replace-all-children path are unaffected.

5. **Test count claim (219 djust_vdom) verified**: 110+6+20+4+4+14+4+4+11+42 = 219. Two-commit shape (Action #181) is correct: `ea6c4c4a` (impl + tests, no CHANGELOG) and `c7c6203a` (CHANGELOG only).

6. **`apply_patches` test-helper no-op for new variants is the right call**: server doesn't apply patches; client does. The added test-only no-op at `patch.rs:84-89, 297-303` keeps existing tests compiling without modeling the marker-id index. The integration_test.rs:396-400 update to skip the new variants in the "patches include djust_id" assertion is correct (subtree patches are dispatched by marker id, not parent dj-id).

7. **Boundary-detection predicate alignment** is well-cited: parser-side at `parser.rs:497-499`, JS-side at `12-vdom-patch.js:38-43`, and the new diff-side at `diff.rs:55-87` all use the same `dj-if<space-or-tab>` prefix check.

8. **Vdom_trace coverage is thorough**: every branch of the pre-pass logs structured trace events for production-debug builds. Aids the rare regression-replay scenario.

## Algorithm correctness for if/elif/else

**Failed.** Detailed trace in finding #1. The matched-id body recursion (`dj_if_pre_pass` lines 233-294) treats inner boundary markers as ordinary VNodes for element-by-element pairing. This causes:
- Step 2 to emit `InsertSubtree(inner_id)` at the OUTER parent's path/index (correct for the keyed machinery's perspective).
- Step 3 to ALSO emit `Replace` + `InsertChild` patches at the outer parent for the same inner content (overlap with #2).

The two patch sets target the same content in different ways, producing a corrupt DOM after application.

This is `if/elif/else` (where at least one of the elif/else branches contains elements, which is the common real-world case - any tab UI, any "either show summary or show details" pattern).

## Wire format match with Iter 2 (load-bearing)

**Pass.** Field-by-field check:

| Iter 3 emit (`diff.rs:298-305`) | Iter 2 read (`12-vdom-patch.js:1358-1389`) |
|---|---|
| `id: String` | `patch.id` (sanitizeIdForLog) |
| `path: Vec<usize>` | `patch.path` -> `getNodeByPath(patch.path, ...)` |
| `d: Option<String>` (skip if None) | `patch.d` -> `getNodeByPath(..., patch.d, ...)` |
| `index: usize` | `patch.index` -> `children[patch.index]` as ref |
| `html: String` | `patch.html` -> `_parseSubtreeHtml(patch.html)` |

The earlier-prompt hypothesis ("Iter 2 expects `target_path` not `path`") is incorrect - both sides use `path`. The serialized wire format also passes the JSON shape check (test_insert_subtree_json_wire_format) against the field names client reads.

## End-to-end reproducer test discipline

**Mixed.** Test #11 (`test_end_to_end_via_parse_html_tab_switch`) is solid - it exercises the full parse_html -> diff pipeline and confirms the parser-side and differ-side predicates agree. Test #1's docstring is inaccurate (finding #2). Test #7 covers the simple sibling-shift case but doesn't exercise the boundary-count-shifts-absolute-index case (finding #3). Most damning: there is NO test for the `if/elif/else` cascade with same outer id (finding #1) - which is the primary case Iter 1's parser emits for elif chains.

Per Action #1196 (delegated-listener integration tests / unit-vs-integration boundary): both the unit shape (boundary detection helpers) and the integration shape (full pipeline through realistic emit patterns) need coverage. The integration coverage here misses the elif case, and that's the case that breaks.

## Boundary detection robustness

The helpers themselves are well-implemented:
- `dj_if_open_id` correctly extracts ids and rejects all examined lookalikes.
- `is_dj_if_close` is strict.
- `find_dj_if_pairs` uses a depth-counter stack and handles nested boundaries (verified by the elif trace - it found both X-0 and X-1).
- `build_excluded_mask` correctly marks all `[open..=close]` ranges as excluded for non-boundary diff.

The robustness gap is downstream: the FOUND nested pairs aren't routed through the right machinery in step 3.

## Backward compatibility

- **Legacy bare `<!--dj-if-->` placeholder (#295)**: preserved (test #10 + legacy path at `diff.rs:513-584` unchanged). Pass.
- **`data-djust-replace`**: preserved (`diff.rs:729-738` short-circuits before pre-pass). Pass.
- **205 existing djust_vdom tests**: confirmed pass in `cargo test -p djust_vdom`. Pass.
- **`apply_patches`/`apply_patch` test helper**: no-op fallthrough for new variants (correct - server doesn't apply patches in production). Pass.

## dj-key reorder limitation

The implementer flagged: "when non-boundary siblings carry `dj-key` AND reorder within their relative slot, position-based pairing of non-boundary children can produce suboptimal patches." The CHANGELOG claims this is "suboptimal patches" not "wrong DOM state."

Reading `dj_if_pre_pass` step 4 (lines 312-353) carefully: when keyed siblings reorder relative to each other but the boundary doesn't change, the `old_kept`/`new_kept` arrays pair them by position, not key. This produces:
- Pair (kept[0], kept[0]): may diff a `key=A` element against a `key=B` element (different content).
- diff_nodes returns Replace patches.

If the keyed siblings have the SAME tag but different content, Replace mostly works (the content gets overwritten in place). If they have DIFFERENT tags, you get RemoveChild + InsertChild that move them through wrong slots. **This can produce wrong DOM state**, not just suboptimal patches.

Production templates rarely mix `{% if %}` with sibling-level `dj-key` reorderings, so this is empirically a rare case. But the CHANGELOG's "suboptimal" framing understates the risk. Recommend filing an explicit follow-up issue with a reproducer (even synthetic) so this is on the radar for v0.10. Acceptable shipping caveat for v0.9.4-1 ONLY IF finding #1 is fixed first.

## Edge cases verified

- Empty boundary same id - 0 patches (test #5, #8). Pass.
- Empty boundary different ids - Remove + Insert (test #9). Pass.
- Same id identical inner content - 0 patches (test #5). Pass.
- Same id changed text inside - SetText recursion, no subtree patches (test #4). Pass.
- Nested boundaries with inner flip and outer stable - inner-only Remove/Insert (test #6). Pass.
- Conditional flip ON and OFF (only-side cases) - tests #2 and #3. Pass.
- Wire format shape - tests #9, #10, #11. Pass (with caveat in finding #5).
- Legacy bare placeholder - test #10. Pass.
- End-to-end via parse_html - test #11. Pass for the cases tested.

## Recommended path forward

1. Add a regression test for the `if/elif/else` cascade case (same outer id, inner id flips). Confirm it FAILS on the current PR head.
2. Fix `dj_if_pre_pass` step 3 (matched-id body recursion) to re-detect nested boundaries. Two reasonable shapes:
   - **Recurse-the-pre-pass**: replace the body iteration with a recursive call to `dj_if_pre_pass(old_body, new_body, child_path, parent_id)` that inherits the parent path. Fall through to element-by-element diff if it returns None.
   - **Wholesale subtree-replace heuristic**: when matched-id body slices have very different shapes (e.g., length differs and inner element shapes differ), short-circuit to RemoveSubtree(matched_id) + InsertSubtree(matched_id, new_html). This is simpler but coarser.
3. Update test #1's docstring to describe what the test actually exercises (separate root-level `{% if %}` blocks, NOT elif cascade).
4. Add a test for the boundary-count-changes-absolute-index sibling-shift case (finding #3).
5. Tighten wire-format JSON tests to assert against parsed `serde_json::Value` shape (finding #5).
6. File a follow-up issue for the dj-key + boundary reorder case noted in the CHANGELOG.

After (1) and (2), this PR can ship and the 3-month-old bug is closed for the cases that matter.

johnrtipton:
## Stage 12 — Address Findings

Stage 11 review found 2 must-fix and 3 should-fix. All 5 addressed in commits 278d6f2a (impl + tests) and ab60c9e0 (CHANGELOG).

### MUST-FIX #1 — Algorithm broken for if/elif/else cascade  RESOLVED

The prior matched-id step iterated body children element-by-element via `diff_nodes`, treating any nested boundary markers as ordinary VNodes. Result: top-level step 2 emitted `InsertSubtree(B)` correctly, but step 3 also emitted overlapping `Replace` + `InsertChild` patches for the same content → corrupt DOM.

**Fix**: `dj_if_pre_pass` now delegates to `dj_if_pre_pass_inner`, which carries old/new offsets and recursively runs the pre-pass on matched-id body slices. `find_top_level_dj_if_pairs` (replacing the original `find_dj_if_pairs`) returns ONLY outermost pairs at the current slice level — nested pairs are discovered when the recursion descends into their containing boundary.

Verified via reproducer: A → elif-B flip now emits exactly `[InsertSubtree(B), RemoveChild(A-content)]` instead of `[InsertSubtree(B), Replace, InsertChild, InsertChild]`.

### MUST-FIX #2 — Test #1 docstring misrepresents scenario  RESOLVED

Renamed `test_1358_reproducer_tab_switch_different_branches` → `test_two_separate_ifs_flip` with clarifying docstring. The fixture (two distinct root-level boundary IDs) corresponds to two SEPARATE `{% if %}` blocks, not an elif cascade. Genuine elif-cascade coverage added in 5 new tests below.

### SHOULD-FIX #3 — Test #7 doesn't exercise position-shift class  RESOLVED

Rewrote test fixture: OLD has 3 inner children (boundary span [1..=5], footer at idx=6), NEW has 1 inner child (boundary span [1..=3], footer at idx=4). The position-shift class is now actually exercised. Defensive assertion that no patch targets OLD's idx=6.

### SHOULD-FIX #4 — Symmetric flaw (only inner boundary disappearing)  RESOLVED

Covered by MUST-FIX #1's recursive pre-pass. Added `test_elif_cascade_b_to_a_flip` regression test (OLD has nested A→B, NEW has just A).

### SHOULD-FIX #5 — Wire-format JSON tests use substring `contains`  RESOLVED

Tightened to `serde_json::to_value` + `json!({...})` shape comparison via `assert_eq!`. Fail message now includes the full diff if shapes diverge.

### Five new elif-cascade regression tests

| Test | Scenario |
|------|----------|
| `test_elif_cascade_a_to_b_flip` | A → elif-B (cascade introduces nested marker) |
| `test_elif_cascade_b_to_a_flip` | elif-B → A (cascade collapses — symmetric direction) |
| `test_elif_cascade_a_to_else_via_inner_change` | matched outer+inner ids: no subtree-flip patches, recursion descends two levels |
| `test_elif_cascade_with_extra_siblings` | footer SetText must use NEW tree's absolute index after boundary span grows |
| `test_elif_cascade_three_branches_a_to_c` | 3-level cascade introduced atomically — recursive pre-pass handles arbitrary depth |

### Verification

- 19 dj-if keyed-diff tests pass (was 14): all 5 new + 14 existing.
- ~740 Rust workspace tests pass.
- 2734 Python tests pass.
- 1559 JS tests pass.
- `cargo clippy` clean.
- Pre-push hook (full cargo test + cargo audit) green.

### Commits

- `278d6f2a` — `fix: recursive dj-if pre-pass for elif cascade + tighten tests (#1358 Iter 3)`
- `ab60c9e0` — `docs(changelog): note elif cascade fix per Stage 11 findings (#1358 Iter 3)`

Two-commit shape per Action #181 (impl + tests in commit 3; CHANGELOG only in commit 4).

johnrtipton (commented):
# Stage 13 Re-Review: PR #1365 (Capability of #1358 — recursive dj-if pre-pass)

## Verdict

**APPROVE** (posting as comment since author cannot self-approve) — All 5 Stage 11 findings genuinely addressed. Independent reproducer (4 anti-overlap reproducers + 5 new edge cases) ALL pass against the new algorithm and FAIL against the original `ea6c4c4a` impl, confirming the recursive pre-pass actually fixes the broken algorithm rather than masking the symptom.

## Findings status

### MUST-FIX #1 — Algorithm broken on elif cascade

**Resolution: ADDRESSED**

**Algorithm trace** (`crates/djust_vdom/src/diff.rs:179-186, 221-444`):
- New `dj_if_pre_pass(old, new, path, parent_id)` (line 179) is a thin shim that calls `dj_if_pre_pass_inner(old, new, 0, 0, path, parent_id)`.
- New `dj_if_pre_pass_inner` (line 221) takes `old_offset`/`new_offset` params for absolute-index propagation through recursion.
- `find_top_level_dj_if_pairs` (line 111) is the new helper that returns ONLY outermost pairs (depth-counter stack at line 114-136), replacing the original `find_dj_if_pairs` which returned ALL pairs including nested ones.
- Step 3 matched-id case (lines 302-384): for each `(old_open, old_close, id)` matched in `new_by_id`, recursively calls `dj_if_pre_pass_inner(old_body, new_body, body_old_offset, body_new_offset, path, parent_id)` on the body slice (lines 323-330).
- Fallback to element-by-element (lines 333-383): fires ONLY when the recursive call returns `None` (i.e. neither side of the body has nested boundaries). The element-by-element loop runs over `&old_body[i]` vs `&new_body[i]` with absolute paths via `body_new_offset + i`.
- Termination: each recursion level shrinks the slice by at least 2 elements (the open + close markers excluded via `&old[*old_open + 1..*old_close]`), guaranteeing bounded recursion.

**Independent reproducer** (`reviewer_repro_stage_11_exact_scenario`, mirrors the Stage 11 review's exact scenario):
- OLD: `<!--A--><div id="a-content">A-content</div><!--/A-->`
- NEW: `<!--A--><!--B--><div id="b-content">B-content</div><!--/B--><!--/A-->`
- Run against current HEAD (`ab60c9e0`): emits exactly `[InsertSubtree(B, idx=1), RemoveChild(a-content)]` — clean, no overlap. PASS.
- Run against original `ea6c4c4a` (broken impl): emits `[InsertSubtree(B), Replace [1] body-a→<!--dj-if id="B"-->, InsertChild idx=2 <div b-content>, InsertChild idx=3 <!--/dj-if-->]` — exactly the corrupt 4-patch overlap the Stage 11 review described. FAIL on broken impl.

**5 new genuine elif-cascade tests** (`tests/test_dj_if_keyed_diff_1358.rs:854-1345`):
- `test_elif_cascade_a_to_b_flip` (line 855): A=true → A=false,B=true. Asserts 0 Replace, 0 bare InsertChild for markers, 1 InsertSubtree(B) at abs idx 1, 0 RemoveSubtree, 1 RemoveChild(a-content). All assertions are non-tautological per Action #1200.
- `test_elif_cascade_b_to_a_flip` (line 970): symmetric direction (covers SHOULD-FIX #4). Asserts 0 Replace, 0 RemoveSubtree(A), 0 InsertSubtree(A), 1 RemoveSubtree(B), 1 InsertChild(a-content) at abs idx 1.
- `test_elif_cascade_a_to_else_via_inner_change` (line 1064): both A and B matched in OLD and NEW; only innermost text differs. Asserts 0 RemoveSubtree, 0 InsertSubtree, 0 Replace, plus SetText('Z'). Verifies the recursion descends through TWO levels without subtree-flip.
- `test_elif_cascade_with_extra_siblings` (line 1135): cascade flip + static `<header>`/`<footer>` siblings. Footer absolute index shifts 4→6 in NEW; asserts SetText targets [6, 0] (NEW abs), defensively no patch targets [4, 0] (OLD abs). Position-shift cascade explicitly verified.
- `test_elif_cascade_three_branches_a_to_c` (line 1254): three-level cascade flip. Asserts exactly InsertSubtree(B) — C is contained in B's html string (single keyed insert per top-level boundary). Verifies B's html includes nested C marker.

All 19 tests in the file pass: `cargo test -p djust_vdom --test test_dj_if_keyed_diff_1358` returns `19 passed; 0 failed`.

### MUST-FIX #2 — Test #1 docstring misrepresented

**Resolution: ADDRESSED**

- Test renamed from elif-cascade-claim to `test_two_separate_ifs_flip` (`test_dj_if_keyed_diff_1358.rs:96`).
- Docstring (lines 77-93) now correctly describes "two separate `{% if %}` blocks at the same nesting level" and explicitly disclaims the elif cascade scope: "this fixture is NOT an `{% elif %}` cascade. An elif cascade produces NESTED markers... see `test_elif_cascade_a_to_b_flip` and friends below for that scenario."
- 5 new genuine elif tests filed (see MUST-FIX #1 evidence). Each docstring describes the actual scenario and the recursive pre-pass behavior under test.

### SHOULD-FIX #3 — Position-shift exercise in test #7

**Resolution: ADDRESSED**

- Test #7 (`test_sibling_shift_does_not_cascade`, line 462) now has DIFFERENT-length boundary spans:
  - OLD: 3 inner children at `[2..=4]`, boundary span `[1..=5]`, footer at idx=6.
  - NEW: 1 inner child at `[2]`, boundary span `[1..=3]`, footer at idx=4.
- Asserts `SetText { path: [4, 0], text: "Footer v2" }` (line 571-582) — NEW tree's actual position.
- Defensive assertion (lines 584-595): NO patch targets `path: [6, 0]` (the OLD footer position).
- Exercises the actual position-cascade class — header SetText at [0, 0] AND footer SetText at [4, 0], not [6, 0].

### SHOULD-FIX #4 — Symmetric flaw

**Resolution: ADDRESSED**

- `test_elif_cascade_b_to_a_flip` (`test_dj_if_keyed_diff_1358.rs:970-1061`) covers the symmetric direction directly.
- OLD has nested A→B with `<div b-content>`; NEW collapses to A only with `<div a-content>`.
- Asserts: 0 Replace, 0 outer-A subtree-flip, exactly 1 RemoveSubtree(B), exactly 1 InsertChild(a-content) at abs idx 1.
- My independent reproducer `reviewer_repro_matched_a_inner_b_replaced_with_p` confirms — passes on fixed code, fails on the original `ea6c4c4a` impl (the original returned `[]` for the InsertChild assertion).

### SHOULD-FIX #5 — Wire-format substring

**Resolution: ADDRESSED**

All 3 wire-format tests rewritten to use `serde_json::to_value` + `json!({...})` + `assert_eq!`:
- `test_remove_subtree_json_wire_format` (line 648): full Value-shape comparison.
- `test_insert_subtree_json_wire_format` (line 668): full Value-shape comparison including all 5 fields.
- `test_insert_subtree_omits_d_when_none` (line 694): full Value-shape comparison + defensive `serialized.get("d").is_none()` to lock the `skip_serializing_if = "Option::is_none"` contract.

Minor nit (non-blocking): the comment cites "Action #1199" (line 654) but Action #1199 is about index/cursor edge-case coverage, not wire-format shape comparison. The technique itself is correct; just an Action # mis-citation. Suggestion: drop the citation or align with `Action #1200` (tautology test detection) which is closer in spirit.

## Independent reproducer result

I wrote 9 independent reproducer tests (4 anti-overlap + 5 edge cases) without consulting the implementer's tests. Run results on current HEAD `ab60c9e0`:

```
running 6 tests (scratch_stage13_review)
test reviewer_repro_degenerate_empty_matched_outer ... ok
test reviewer_repro_no_overlap_in_a_to_b_flip ... ok
test reviewer_repro_mixed_body_with_inner_boundary ... ok
test reviewer_repro_matched_a_inner_b_replaced_with_p ... ok
test reviewer_repro_stage_11_exact_scenario ... ok
test reviewer_repro_four_level_cascade ... ok

running 3 tests (scratch_stage13_extra)
test reviewer_repro_three_sibling_matched_pairs_no_changes ... ok
test reviewer_repro_two_sibling_top_level_pairs ... ok
test reviewer_repro_matched_outer_inner_two_pairs_one_new ... ok
```

When run against original `ea6c4c4a` (pre-fix), 4 of these tests FAIL with exactly the corrupt-DOM patch sequence the Stage 11 review described. This is empirical proof the fix is real, not just a docs/test patch.

The reviewer reproducers covered:
1. Stage 11's exact scenario (A → A+B+content) — exactly 2 patches, no overlap.
2. Mixed body with non-boundary siblings around a matched inner boundary.
3. Matched outer with inner B replaced by non-boundary `<p>`.
4. Degenerate empty matched outer.
5. 4-level cascade A→B→C→D.
6. Two sibling top-level pairs (A matched, B→C flip).
7. Three sibling matched pairs identical (zero patches).
8. Matched outer with NEW inner introducing top-level B alongside matched-inner C.

## New concerns

None blocking. Two very-minor observations:

1. **Action # citation in test #9** (`test_dj_if_keyed_diff_1358.rs:654`): cites "Action #1199" but #1199 is about index/cursor coverage. Either drop the citation or align with the actually-applicable rule (the closest is just "Stage 11 finding #5 on this PR"). Non-blocking.

2. **`dj-key` + boundary reorder limitation** (CHANGELOG line 222-228, code line 391-397): the impl ships a "non-boundary siblings paired by relative position" approach which can produce wrong DOM state (not just suboptimal patches) when keyed siblings of different tags reorder across a boundary flip. The Stage 11 review recommended filing a follow-up issue with a reproducer; current scope acknowledges this honestly in the CHANGELOG and code comment but no follow-up issue is referenced. Recommend filing one for v0.10 visibility, but acceptable to ship v0.9.4-1 without it given production templates rarely mix this pattern.

## Verification gates

- **Two-commit shape (Action #181)**: PASS.
  - `278d6f2a` (impl commit): touches `crates/djust_vdom/src/diff.rs` + `crates/djust_vdom/tests/test_dj_if_keyed_diff_1358.rs` ONLY — no CHANGELOG.
  - `ab60c9e0` (docs commit): touches `CHANGELOG.md` ONLY.

- **CI status**: 12/13 green. Only failure is the known `playwright-tests` DraftMode flake (`Auto-Save Functionality` / `Title not saved`), unrelated to this PR's Rust VDOM changes. Rust, Python (3.12/3.13/3.14), JS, security, CodeQL, all green.

- **Test count claim**: CHANGELOG says "19 cases in `crates/djust_vdom/tests/test_dj_if_keyed_diff_1358.rs`" (line 235-236) — verified via `grep -c '^#\[test\]'` returning 19. Matches. CHANGELOG also claims "All 19 dj-if keyed-diff tests pass" (line 260) — verified by local `cargo test`.

- **All djust_vdom tests pass**: 224 tests (110+6+20+6+4+4+19+4+4+11+13+10+3 across all binaries) — no regressions in adjacent code.

## Recommendation

**APPROVE.** The recursive pre-pass shape is the correct fix for the elif-cascade class. My independent reproducers confirm the algorithm produces clean, non-overlapping patches across A→A+B, A+B→A, A→A+B+C+D, and several mixed scenarios. The 5 new elif-cascade tests in the impl-author's file are non-tautological (each asserts both presence of expected patches AND absence of the broken-shape patches like `Replace` and `InsertChild` for boundary markers).

This iter genuinely closes the 3-month-old `{% if %}`-breaks-VDOM-patching bug class for both:
1. The base case (separate `{% if %}` blocks) — already worked in the original impl.
2. The elif-cascade case — the Stage 11 finding that was actually the more dangerous failure mode (silent corrupt DOM vs loud 500-class errors).

Wire format with Iter 2 (PR #1364) verified at the Value-shape level. Backwards compat with legacy bare `<!--dj-if-->` placeholder (#295) preserved. Termination of recursion bounded.

Ship it.

PR #1371, opened 2026-05-06
fix(client): hooks TDZ in v0.9.4rc1 — let → var for _activeHooks (#1370 P0 hotfix)

### LINKED ISSUE #1370: 0.9.4rc1: ReferenceError: Cannot access 'G' before initialization in _ensureHooksInit
## Description

`client.js` in 0.9.4rc1 throws a temporal dead zone error on every page load and every WebSocket patch application:

```
Uncaught ReferenceError: Cannot access 'G' before initialization
    at _ensureHooksInit (client.js:9594:10)
    at mountHooks (client.js:9698:5)
    at djustInit (client.js:7812:5)
    at client.js:7842:5
```

And on every WS message:

```
Uncaught ReferenceError: Cannot access 'G' before initialization
    at _ensureHooksInit (client.js:9594:10)
    at updateHooks (client.js:9751:5)
    at reinitAfterDOMUpdate (client.js:4675:5)
    at n (client.js:923:25)
```

## Reproduction

1. Upgrade from djust 0.9.0 to 0.9.4rc1
2. Deploy (collectstatic runs, bundles new client.js)
3. Load any page with a LiveView
4. Console shows the error on mount and on every subsequent WS patch

Hard reload does NOT fix it — the error persists with the fresh 0.9.4rc1 client.js. This is not a cache issue.

## Environment

- Django 5.2.14
- djust 0.9.4rc1 (upgraded from 0.9.0)
- Python 3.12
- Chrome 136 (also reproduced on reload)
- Served over HTTP (non-localhost) via ALB — the COOP header warning is separate and unrelated

## Impact

The error appears to break hooks initialization entirely. `mountHooks` and `updateHooks` both call `_ensureHooksInit` which references a `let`-declared variable `G` before it's initialized. This means:

- No hooks fire on mount
- No hooks fire after DOM patches
- Any feature relying on hooks (lifecycle callbacks, DOM observers, etc.) is dead

## Notes

The server-side VDOM diff in 0.9.4rc1 works correctly — we confirmed via `LiveViewTestClient` that `{% if %}` block patching (djust#1358) is fixed. The regression is purely client-side in the bundled JS.

We've rolled back to 0.9.0 for now.

johnrtipton:
## Summary

P0 hotfix for #1370. v0.9.4rc1 throws `ReferenceError: Cannot access 'G' before initialization` on every page load + every WS patch. Hooks entirely broken on rc1; user rolled back production to 0.9.0.

## Root cause

`G` is the minified `_activeHooks`. Declared `let _activeHooks;` in `19-hooks.js:54`. Module 19 is concatenated AFTER the bootstrap call at bundle line ~7842. `let` is in TDZ until the declaration line is reached.

When `document.readyState !== 'loading'` (deferred scripts / dynamic injection / turbo-nav re-init — the common case for any not-first-load page), the bundle takes the synchronous-init branch:

```js
if (document.readyState === 'loading') {
    document.addEventListener('DOMContentLoaded', djustInit);
} else {
    djustInit();   // <-- fires inside eval, BEFORE module 19's `let` runs
}
```

`djustInit() → mountHooks() → _ensureHooksInit()` reads `_activeHooks` → TDZ ReferenceError.

## Fix

`let` → `var` for `_activeHooks` and `_hookIdCounter` in `src/19-hooks.js:54-56`. `var` hoists to script-top with `undefined` initial value, so the lazy-init `if (!_activeHooks)` check in `_ensureHooksInit()` works regardless of concat order. Eslint `no-var` disabled inline with rationale comment.

The functional change is 2 lines (plus 2 eslint-disable directives + revised comment).

## Why PR #1359 missed it

PR #1359's 97 vitest failures caught 4 cross-module vars (`liveViewWS`, `clientVdomVersion`, `_eventRefCounter`, `_isBroadcastUpdate`) — all DECLARED-EARLY-USED-LATE patterns. `_activeHooks` is the inverse: DECLARED-LATE-USED-EARLY. vitest's import-order tests do not simulate bundle-concat-order execution under `readyState === 'complete'`.

## Regression test

`tests/js/bundle-init-no-tdz.test.js` (2 cases) loads the bundled `client.js` in a fresh JSDOM env where `document.readyState === 'complete'` BEFORE eval runs (wait on `load` event). Asserts no `ReferenceError: Cannot access 'X' before initialization` is thrown during bootstrap.

**Verified locally** (per Action #1196):
- Pre-fix bundle (rc1, `let _activeHooks`): both tests FAIL with `Cannot access '_activeHooks' before initialization`
- Post-fix bundle (`var _activeHooks`): both tests PASS

This catches the structural class — any future `let` declared late in the concat that's used early will trip the test.

## Scope discipline (per the brief)

This PR does NOT:
- Refactor other modules
- Convert other `let` declarations to `var` proactively
- Add the bundle-init-order structural lint mentioned in the milestone block (follow-up)
- Touch any source modules besides `19-hooks.js`
- Fix the eslint cleanup canon (separate process work)

JUST: `19-hooks.js` 2-line fix + bundle rebuild + regression test + CHANGELOG.

## Test plan

- [x] Regression test FAILS on pre-fix bundle (verified 2026-05-02)
- [x] Regression test PASSES on fixed bundle (verified 2026-05-02)
- [x] Full vitest suite passes (1561 tests, no other TDZ surfaced)
- [x] Full pytest suite passes (4204 passed, 18 skipped, no Python regressions)
- [x] Pre-commit hooks pass (eslint --max-warnings 0 still clean with `// eslint-disable-next-line no-var`)
- [x] Pre-push hooks pass (CHANGELOG test-counts, full pytest, etc.)

- `8f86af49` — fix(client): hooks TDZ ... (impl + tests + bundle, NO CHANGELOG)
- `b9951ee9` — docs(changelog): #1370 hooks TDZ hotfix (CHANGELOG only)

Closes #1370.


johnrtipton (commented):
# Stage 11 Code Review: PR #1371 (P0 hotfix #1370 hooks TDZ)

## Verdict

**APPROVE (comment, since author=reviewer)** — fix is correctly scoped, regression test is structurally sound and non-tautological, two-commit shape is clean, CI green for all critical checks. Ship it as v0.9.4rc2.

## 🔴 Must-fix
None.

## 🟡 Should-fix
None blocking. One observation deferred (see "Other modules audit").

## 🟢 Strong points

- **Inline rationale comment** at `python/djust/static/djust/src/19-hooks.js:50-53` explains the bundle-concat-order constraint clearly. Future contributors won't "fix" the eslint warning by reverting to `let`.
- **`// eslint-disable-next-line no-var` directives** on both lines (54 and 56). Necessary post-#1351 / #1359 to keep the eslint cleanup intact.
- **Regression test architecture** is exactly right: loads the BUNDLED `client.js` from disk (not source modules), explicitly waits on the `load` event so `document.readyState === 'complete'` BEFORE `dom.window.eval(clientCode)`. Faithfully reproduces the production failure mode (deferred / late-injected scripts / turbo-nav re-init).
- **Empirical fail-pre / pass-post verification** documented in the test header (`tests/js/bundle-init-no-tdz.test.js:33-36`) and commit message body. Action #1196 satisfied — real integration test, not a method-level unit stub.
- **Two-commit shape clean** (Action #181): commit 1 = source + bundle + test (no CHANGELOG); commit 2 = CHANGELOG only.
- **CHANGELOG entry** is comprehensive — names the fix (`let → var`), root cause (bundle-concat-order), gap explanation (PR #1359 import-order tests catch the inverse pattern), regression test, and the diagnostic finding about `readyState !== 'loading'`.
- **Bundle artifacts consistent**: `client.js`, `client.min.js`, `client.min.js.gz`, `client.min.js.br`, `client.min.js.map` all share mtime `2026-05-05 21:34:04` — single rebuild, no stale artifacts. Minified bundle reads `var _activeHooks,_hookIdCounter`.
- **Scratch artifact discipline**: `scratch/client-pre-fix.js` documented in test header as the verification source but NOT committed. Per global "scratch directory for temporary scripts" pattern.

## The fix

`python/djust/static/djust/src/19-hooks.js:54-56`:

```javascript
// var (not let): module 19 is concatenated after the bootstrap call;
// let would TDZ when _ensureHooksInit() is invoked from earlier modules.
// The lazy-init pattern below relies on hoisted-undefined semantics.
// eslint-disable-next-line no-var
var _activeHooks;
// eslint-disable-next-line no-var
var _hookIdCounter;
```

Bundle rebuild verified: `python/djust/static/djust/client.js:9586` reads `var _activeHooks;` and `:9588` reads `var _hookIdCounter;`. Lazy-init at `client.js:9595` (`if (!_activeHooks) { _activeHooks = new Map(); _hookIdCounter = 0; }`) works because `var` hoists the binding to the script-top with `undefined`, making the `if (!_activeHooks)` check truthy on first invocation.

Confirmed call chain that triggers TDZ on rc1:
- `python/djust/static/djust/src/14-init.js:158-162` — synchronous `djustInit()` when `readyState !== 'loading'`.
- `python/djust/static/djust/src/14-init.js:131` — `djustInit()` calls `mountHooks()`.
- `python/djust/static/djust/src/19-hooks.js:166-167` — `mountHooks()` calls `_ensureHooksInit()`.
- `python/djust/static/djust/src/19-hooks.js:62-67` — `_ensureHooksInit()` reads `_activeHooks`. With `let` (rc1), TDZ until line 54 is reached. With `var` (fix), hoisted-undefined makes the read return `undefined` and lazy-init runs.

## The regression test (load-bearing)

`tests/js/bundle-init-no-tdz.test.js`:

- **Loads bundle from disk** (`fs.readFileSync(BUNDLE_PATH)` at line 45 reads `python/djust/static/djust/client.js`) — not source modules. Non-tautological.
- **Forces `readyState === 'complete'` before bundle eval**: `makeDomWithCompleteReadyState()` (lines 55-87) creates a JSDOM with `runScripts: 'dangerously'`, then `await new Promise(resolve => { ... dom.window.addEventListener('load', resolve, { once: true }); })` at lines 81-84 ensures the document has reached `'complete'` BEFORE `dom.window.eval(clientCode)` runs. **This is the load-bearing setup** — without it (default JSDOM `readyState === 'loading'`), the `if (document.readyState === 'loading')` branch at `14-init.js:158` would defer init via `addEventListener('DOMContentLoaded', ...)` and the TDZ would NOT fire synchronously inside `eval`.
- **Explicit assertion at line 93**: `expect(dom.window.document.readyState).toBe('complete')` confirms the test setup actually achieved the production failure-mode condition.
- **Catches `ReferenceError` specifically**: line 110 uses `expect(message).not.toMatch(/Cannot access '[^']+' before initialization/)` — exactly matches the rc1 failure message format. Then `throw initError` at line 112 ensures any OTHER unexpected init error also surfaces.
- **Sanity check at line 118**: `expect(dom.window.djustInitialized).toBe(true)` confirms `djustInit()` ran to completion (assignment at `14-init.js:147`). If TDZ fired, this would be `undefined`.
- **Test isolation**: each `it()` calls `makeDomWithCompleteReadyState()` independently → fresh JSDOM `window`. The double-load guard at `00-namespace.js:119` (`if (window._djustClientLoaded)`) doesn't poison the second test because `_djustClientLoaded` is per-window.

**Action #1196 verification** (would the test FAIL on rc1 bundle?): Yes. Commit message body: "Verified locally: regression test FAILS against rc1 bundle with `Cannot access '_activeHooks' before initialization`, PASSES against the fixed bundle." Test header (`bundle-init-no-tdz.test.js:33-36`) restates this as a doc-claim.

**Action #1200 (tautology detection)**: not a tautology. On the rc1 bundle, `let _activeHooks` trips TDZ inside `_ensureHooksInit()` at the `mountHooks()` call site, BOTH `expect(initError).toBeNull()` AND `expect(djustInitialized).toBe(true)` fail. On the fix, both pass.

## Other modules audit (structural class)

Ran `grep -nE "^let " python/djust/static/djust/src/2*.js python/djust/static/djust/src/3*.js python/djust/static/djust/src/4*.js python/djust/static/djust/src/5*.js` to find top-level `let` in late-numbered modules. Two hits:

1. **`python/djust/static/djust/src/39-dj-track-static.js:29`** — `let _djTrackStaticSnapshot = null;`. Read sites: `_checkStale()`, `_snapshotAssets()`, `_onWsReconnected()`, `_resetSnapshot()`. None invoked synchronously from earlier modules at bundle-eval time. Wires `_onWsReconnected` to a custom event listener; init via public API. **No TDZ risk.**

2. **`python/djust/static/djust/src/43-dj-transition-group.js:52`** — `let _rootObserverInstalled = false;`. Module 43 calls `_installRootObserver()` synchronously at line 175 when `readyState !== 'loading'` (same pattern as `14-init.js:161`!), and `_installRootObserver()` reads `_rootObserverInstalled` at line 127. **However, the read happens INSIDE module 43, AFTER its own `let` declaration on line 52.** Same module = declaration reached before synchronous call site. **No TDZ risk.**

3. **`python/djust/static/djust/src/44-dj-flip.js:65`** — `let _rootObserverInstalled = false;` is INSIDE a function scope (block-scoped). **No TDZ risk.**

The audit found no other DECLARED-LATE-USED-EARLY pattern requiring follow-up. The new `bundle-init-no-tdz.test.js` runtime test catches the class structurally — any future `let X` declared late that's read early will trip the regression test on next CI.

## Edge cases verified

- Bundle artifacts all share mtime `2026-05-05 21:34:04` — single rebuild, no stale siblings.
- Minified `client.min.js` shows `var _activeHooks,_hookIdCounter` (terser preserved names since top-level vars).
- Two-commit shape per Action #181 verified: commit `8f86af49` touches source + bundle + test (no CHANGELOG); commit `b9951ee9` touches CHANGELOG only.
- Test isolation: each `it()` creates a fresh JSDOM, double-load guard doesn't poison subsequent tests.
- CI at review time: SUCCESS for `javascript-tests`, `Analyze JavaScript`, `playwright-tests`, `CodeQL`, `Analyze Python`, `Detect Security Hot Spots`, `security-scan`, `security-tests`. Python tests for py3.12/3.13/3.14 and rust-tests still running but unrelated to this JS-only fix.
- `scratch/client-pre-fix.js` referenced in test docs but NOT committed (`git ls-files | grep scratch` empty, working tree clean).
- `_activeHooks` and `_hookIdCounter` confirmed used at multiple read sites in module 19 — the `if (!_activeHooks)` lazy-init pattern handles the hoisted-undefined initial value correctly.

Ship as v0.9.4rc2.

Issue #1373, opened 2026-05-06
Object-level authorization lifecycle (get_object + has_object_permission)


johnrtipton:
## Summary

djust currently has no first-class lifecycle hook for object-level authorization. The two existing auth surfaces — `permission_required` (class-level role check) and `check_permissions(self, request)` (custom hook before mount) — are insufficient for views bound to a single object via URL kwarg (`document_id`, `user_id`, etc.). This is a structural IDOR class affecting any LiveView with that shape.

## Reproducer

```python
# User A authenticated, has documents.access (role), NOT granted access to document 99.
class DocumentDetailView(LiveView):
    permission_required = "documents.access"  # role check at WS connect

    def mount(self, request, document_id=None, **kwargs):
        self.document_id = document_id  # bound from URL kwarg

    def get_context_data(self, **kwargs):
        ctx = super().get_context_data(**kwargs)
        doc = Document.objects.get(pk=self.document_id)
        if not can_access_document(self.request.user, doc):  # object check at render
            raise PermissionDenied()
        ctx["document"] = doc
        return ctx

    @event_handler()
    def add_comment(self, subject="", body="", **kwargs):
        doc = Document.objects.get(pk=self.document_id)  # NO PER-EVENT CHECK
        Comment.objects.create(document=doc, subject=subject, body=body)
```

**Attack path:**

1. User A opens WS to `/documents/99/`. `check_view_auth` passes (role check OK), `mount()` runs, `self.document_id = 99`. `get_context_data` raises `PermissionDenied` during render.
2. djust does NOT close the WS or unbind `self.document_id` — `PermissionDenied` is only caught in the *connect* path (`websocket.py:1948`), not in the post-mount render or in `handle_event`.
3. User A sends `{"event": "add_comment", "params": {"subject": "x", "body": "y"}}` over the established WS.
4. Handler runs `Document.objects.get(pk=99)` and creates a comment on a document User A has no access to.

Reproducible with any WS client in ~5 minutes against the current djust version.

## Root cause

The natural place for object-level checks ("after the URL kwarg is bound to `self`") is `get_context_data` — it's where Django developers reach for `get_queryset` filtering. But `get_context_data` runs *during render*, after `mount()` has already set up the WS-session-scoped state. The framework doesn't push back on this placement, and there's no first-class hook that:

1. Runs *after* URL kwargs are bound to `self` (so the object can be resolved)
2. Runs *before* mount completes (so a denial unbinds the session)
[...]

Issue #1243, opened 2026-05-01
tech-debt: Stage 4 plan-template — verify literal API contracts before locking the plan


johnrtipton:
Source: v0.9.2-1 retro / PR #1242 review.

## Context

The plan in PR #1240's state file said:

```python
await self.transport.send_error(
    "...",
    type="mount_error",
    view=view_path,
)
```

Existing djust convention (`websocket.py:1835/1891/2116/2133/2217`) is `error_type=` and `view_class=`. The literal `type=` would have collided with the outer `{\"type\": \"error\", **kwargs}` envelope shape. The implementer correctly followed convention rather than the spec snippet, but this is the kind of failure mode that bites silently when an implementer pastes verbatim from a plan.

## Proposed change

Add to `.pipeline-templates/{feature,bugfix}-state.json` Stage 4 checklist:

```json
{
  "action": "VERIFY LITERAL API CONTRACTS — for every literal API call in the plan (function names, kwargs, return shapes), grep for the existing convention before locking. Spec snippets in plans can be wrong; convention wins.",
  "done": false,
  "mandatory": true
}
```

## Acceptance

- Stage 4 checklist in feature-state.json + bugfix-state.json templates includes the verification item.
- The item is mandatory (so it can't be silently skipped).
- A future PR demonstrates the check catching a spec/convention mismatch (or, more likely, the canonicalization catches it before the plan is locked).

PR #1374, opened 2026-05-06
feat(auth): object-permission lifecycle foundation (v0.9.5-1a, #1373)

johnrtipton:
Closes the foundation iteration of [#1373](https://github.com/djust-org/djust/issues/1373) (object-level authorization lifecycle). Design pinned in [ADR-017](https://github.com/djust-org/djust/blob/main/docs/adr/017-object-permission-lifecycle.md). Per-event re-execution lands in v0.9.5-1b after this iteration soaks one release; tooling (`djust check` + guide + skill) lands in v0.9.5-1c.

## Summary

Adds djust's first-class lifecycle hook for object-level authorization. Two new public methods on `LiveView` (default no-op — opt-in via override):

- `get_object(self) -> Optional[Any]` — return the view's primary object
- `has_object_permission(self, request, obj) -> bool` — return True if the request user may access `obj`
- `_invalidate_object_cache(self) -> None` — handlers call this when ownership-determining state mutates

Plus the framework helper `djust.auth.check_object_permission(view, request)` that wires the two together. Mount-time enforcement only in this iteration; per-event lands in `-1b`.

## Why

Closes a structural IDOR class. Reproducer (verbatim from ADR-017):

```python
# User A authenticated, has documents.access (role), NOT granted access to document 99.
class DocumentDetailView(LiveView):
    permission_required = "documents.access"  # role check at WS connect

    def mount(self, request, document_id=None, **kwargs):
        self.document_id = document_id

        doc = Document.objects.get(pk=self.document_id)
        if not can_access_document(self.request.user, doc):
            raise PermissionDenied()  # <-- runs during render, AFTER mount
        ctx["document"] = doc
        return ctx

    @event_handler()
    def add_comment(self, subject="", body="", **kwargs):
        doc = Document.objects.get(pk=self.document_id)  # NO PER-EVENT CHECK
        Comment.objects.create(document=doc, subject=subject, body=body)
```

Attack: User A opens WS to `/documents/99/`. `check_view_auth` passes (role OK). `mount()` runs and sets `self.document_id = 99`. `get_context_data` raises `PermissionDenied` during render — but djust does NOT catch that at the post-mount/post-render stage, so the WS stays open with `self.document_id = 99`. User A then sends `add_comment` over the WS and the handler runs `Document.objects.get(pk=99)` against a document they have no access to.

After this PR: with `get_object()` + `has_object_permission()` overridden, the WS is closed with code 4403 at mount time before any handler can fire. (Per-event hardening for views that change object via state mutation lands in `-1b`.)

## Decision 5 split-call-site

`check_view_auth` runs at `websocket.py:1947` BEFORE `mount()` at `:2134`. djust's WS path doesn't call Django's `View.setup()`, so `self.kwargs` is never bound — URL kwargs only become accessible after the user's `mount()` body assigns them. Therefore `get_object()` reading `self.<x>_id` requires `mount()` to have run.

The literal "extend `check_view_auth`" of the original ADR-017 Decision 5 is mechanically infeasible. Decision 5 was amended (Stage 4) to clarify that the **logical** auth onion (login → role → custom → object) maps to **two physical call sites**: `check_view_auth` stays pre-mount and unchanged; `check_object_permission` is a separate helper called post-mount at `websocket.py:2147` (after `_capture_dirty_baseline`, before the outer `except` catch).

## Defense-in-depth: OWASP 404-shape automation

`check_object_permission` catches `django.core.exceptions.ObjectDoesNotExist` (and subclasses `Model.DoesNotExist` / `Http404`) and treats it as `None` — automating the 404-shape pattern so a naive `Model.objects.get(pk=missing)` in `get_object()` doesn't leak existence via a `DEBUG=True` traceback. Surfaced in Stage 8 security review.

## Backwards compatibility

Verified empirically: full pytest suite (4670 tests + the 6 new ones = 4676 total) passes; 1563 JS tests pass. Views that don't override `get_object` see zero behavior change — `_has_custom_get_object` short-circuits before any work happens.

## Commits

| Stage | Commit | Subject |
|---|---|---|
| 4 — Artifact-first | `3815e3a2` | failing reproducer + ADR-017 Decision 5 amendment |
| 5 — Implementation | `dabf00dd` | foundation (live_view + auth/core + websocket + auth/__init__) |
| 8 — Security fix | `c87c6dec` | catch ObjectDoesNotExist (DEBUG-traceback leak defense) |
| 9 — Docs | `184bd1d3` | CHANGELOG entry under [Unreleased] |

Two-commit shape (Action #181 / GitHub #1173) preserved: implementation has zero CHANGELOG.md changes; Stage 9 commit is docs-only.

## Test plan

- [x] `pytest tests/integration/test_object_permission_mount.py -v` — 6/6 pass (1 added in Stage 8 fix)
- [x] `pytest tests/ python/djust/tests/ -q --ignore=tests/benchmarks` — 4671 passed, 14 skipped, 0 failed
- [x] `npm test` — 1563 passed
- [ ] Stage 11 reviewer Code Review
- [ ] Per-event semantics validation in `-1b`
- [ ] System-check + guide validation in `-1c`

## References

- [ADR-017](https://github.com/djust-org/djust/blob/main/docs/adr/017-object-permission-lifecycle.md) (full design — read this first)
- ROADMAP.md milestone `v0.9.5-1a`
- `python/djust/auth/core.py:114` — `_has_custom_check_permissions` (the existing pattern this mirrors)
- `python/djust/websocket.py:1953-1955` — pre-mount denial path (the close-4403 pattern this reuses)
- Action #1122 — split-foundation pattern (this milestone applies it)
- Action #181 / GitHub #1173 — two-commit shape (preserved)


johnrtipton:
## Stage 11 Code Review

Approached as a fresh reviewer. Walked the trace from observable symptom (the IDOR reproducer in the PR body) through to the new physical call site at `websocket.py:2148-2167`, then audited the helper at `auth/core.py:144-215`, the cache slot in `live_view.py:506-513`, and the test coverage in `tests/integration/test_object_permission_mount.py`. Read ADR-017 and the Decision 5 amendment.

**Headline:** the IDOR closure works as described for the documented happy path. The `try/except PermissionDenied` block is correctly nested inside the outer `try` (so non-`PermissionDenied` exceptions still flow to `handle_exception` at `websocket.py:2168`); `await self.close(code=4403)` runs before any handler can fire on denial; `return` is in the right place; the call site is correctly post-`_capture_dirty_baseline`. Mirrors the pre-mount denial pattern at `websocket.py:1948-1955` verbatim, which is exactly the right move.

I did find one **factual error** in the docstring + CHANGELOG that should be fixed before merge.

### Findings

🟡 **[should-fix] `auth/core.py:169-171` + `CHANGELOG.md:23` falsely claim `Http404` is a subclass of `ObjectDoesNotExist`.** Verified against the installed Django: `class Http404(Exception):` in `django/http/response.py:699` — it inherits directly from `Exception`, not from `ObjectDoesNotExist`. So the current `except ObjectDoesNotExist:` at `auth/core.py:196` will NOT catch a developer's `raise Http404(...)` from `get_object()`. The docstring says it does; the CHANGELOG says it does ("`and subclasses Model.DoesNotExist / Http404`"). This is the exact OWASP `DEBUG=True` traceback leak the PR claims to close — and it doesn't, for the `Http404` case. Two ways to fix: (a) extend the catch to `(ObjectDoesNotExist, Http404)`, which honors the documented contract; or (b) drop the `Http404` claim from the docstring + CHANGELOG and document that `Model.DoesNotExist` only is the supported "naive lookup" idiom (since most Django ORM lookups raise `Model.DoesNotExist` which IS a subclass of `ObjectDoesNotExist`). Recommend (a) — it costs one identifier and matches developer intuition. Not a release blocker (the typical idiom `Model.objects.get(pk=...)` IS covered correctly), but the doc-vs-code drift is fixable in one line.

🟡 **[should-fix] Missing test for `has_object_permission` raising `PermissionDenied` directly.** The 6 regression tests cover the `return False` path (test 1) but not the explicit-raise path. ADR-017 documents that "raise `PermissionDenied`" and "return False" are both supported (per the docstring at `live_view.py:292-294`). One additional test case (a `_RaiseDeniedView` whose `has_object_permission` raises) would lock the contract. Trivially small to add.

🟡 **[should-fix] Edge-case test gap: falsy non-None object.** `auth/core.py:206` does `if obj is None: return` — strict identity check, so `False` / `0` / `""` would proceed to `has_object_permission(request, obj)`. The ADR says "None means no primary object" but the contract for falsy-non-None is undocumented. A 2-line test (override `get_object` to return `False`, assert `has_object_permission` IS called) would lock the semantics. Per Action #1199 (index/cursor edge-case coverage), this is the kind of boundary case that's cheap to specify and prevents future ambiguity.

🟢 **[nit] `auth/core.py:192` function-local `from django.core.exceptions import ObjectDoesNotExist` is inconsistent.** `PermissionDenied` is imported at top (line 32). No circular-import reason exists for `ObjectDoesNotExist`. Hoist to module top for consistency.

🟢 **[observation/-1b carryover] Cache populated BEFORE permission check at `auth/core.py:205`, vs check at line 209.** For v0.9.5-1a (mount-time only) this is benign — denial closes the WS, the cache won't be observed. But for v0.9.5-1b (per-event re-execution), if `has_object_permission` raises a non-`PermissionDenied` exception (developer bug, attribute error, etc.) AFTER the cache is populated, the cache stays populated with a presumptively-allowed object that was never actually allowed. Recommend ordering as: call `get_object()` → call `has_object_permission()` first → only THEN assign `self._object = obj`. Costs nothing in v0.9.5-1a (test 4 just verifies cache is populated post-allow, which still holds with the swap). Worth tracking explicitly as a v0.9.5-1b consideration.

🟢 **[nit] `_CounterView._calls` class attribute pattern.** Test 4 does `_CounterView._calls = 0` before use. Tests run sequentially in pytest by default, so no race; but if the file ever runs under `pytest-xdist` with `--dist loadfile`, two test workers on the same file would still share the class. Not a real risk today, just an observation.

🟢 **[observation] `PermissionDenied(f"Access denied...")` f-string.** CLAUDE.md security rules forbid f-strings in **logger** calls; exception messages are fine. The actual log emission at `auth/core.py:211-214` correctly uses `%s` formatting. No issue.

🟢 **[observation] Type hints + docstrings on the new public methods are appropriately WHY-focused** (`live_view.py:1023-1050`). The `Optional[Any]` return on `get_object` is intentional (any object — Django model, stub, etc.) and documented. Docstrings explain cache lifetime, invalidation contract, OWASP rationale — none feel like dead-weight WHAT.

🟢 **[observation] CHANGELOG entry under [Unreleased] is excellent at WHY** (lines 9-35). Frames the bug class, names the rollout split (-1a / -1b / -1c), explicitly says "mount-time only" so a third party reading it knows the limitation. The single factual error is the `Http404` claim flagged above.

### Edge cases considered

- [x] `has_object_permission` returns `False` (covered by test 1)
- [x] `has_object_permission` returns `True` (covered by test 2)
- [x] `has_object_permission` raises `PermissionDenied` (NOT covered — flagged)
- [x] `get_object()` returns `None` (covered by test 5)
- [x] `get_object()` returns falsy non-None (NOT covered — flagged)
- [x] `get_object()` raises `ObjectDoesNotExist` (covered by test 6)
- [x] `get_object()` raises `Http404` (NOT caught by current code despite docstring claim — flagged)
- [x] `get_object()` raises `Model.DoesNotExist` (caught — IS subclass of `ObjectDoesNotExist`)
- [x] `get_object()` raises a non-permission exception (correctly propagates to `handle_exception`)
- [x] Snapshot-restore path (verified: `_object` allocated before `_framework_attrs` snapshot at `live_view.py:513`, so it's a framework slot, reset on restore — Decision 5 amendment claim verified)
- [x] Order of try/except blocks at `websocket.py:2156-2166` (verified: inner `try/except PermissionDenied` doesn't swallow other exceptions)
- [x] Cache populated before vs after permission check (flagged for -1b)

### Verdict

APPROVE

The IDOR closure works for the typical `Model.objects.get` idiom, the call-site shape is correct, the split-foundation rollout per Action #1122 is honored, and the test suite covers the major paths. The `Http404`-not-actually-caught issue is a doc-vs-code drift, not a security regression vs. status quo (status quo also doesn't catch it). Recommend fixing the `Http404` claim (preferably by extending the `except` clause) and adding two small tests (PermissionDenied-direct, falsy-non-None) before merge, but none of these are release blockers — could land as a follow-up if -1a needs to ship today. Doc-vs-code drift is the kind of finding Action #1197 (canon-doc citation discipline) warns about; worth catching here before it ossifies into reference material for -1b/-1c.

johnrtipton:
## Stage 13 Re-Review

Stage 11 left 3 🟡 should-fix findings. Stage 12 (commit `1cd6918b`)
addressed them. Verification:

### Finding A — Http404 doc-vs-code drift: ✅ ADDRESSED
- `python/djust/auth/core.py:196` now does `except (ObjectDoesNotExist, Http404):` (two classes listed explicitly).
- `from django.http import Http404` added at `auth/core.py:195` (function-local import).
- Docstring at `auth/core.py:166-184` clearly explains the two-class catch and the WHY: "Http404 does NOT inherit from ObjectDoesNotExist (its parent is Exception), so both must be listed explicitly."
- Inline comment at `auth/core.py:200-208` reiterates the rationale.
- CHANGELOG entry under [Unreleased] at line 24 updated to reflect the two-class catch with explicit note about the inheritance.
- **Django source verification**: `Http404 mro: ['Http404', 'Exception', 'BaseException', 'object']` and `ObjectDoesNotExist in Http404 mro: False` — confirms the docstring is accurate.
- Test `test_get_object_raising_http404_treated_as_none` at `tests/integration/test_object_permission_mount.py:362` exercises `_Http404View` (lines 142-162) and asserts `view._object is None`. PASSES.

### Finding B — `has_object_permission` raising `PermissionDenied`: ✅ ADDRESSED
- `_PermissionDeniedView` at `tests/integration/test_object_permission_mount.py:165-182` raises `PermissionDenied("explicit denial with custom message")`.
- `test_has_object_permission_raising_directly` at line 379 asserts `pytest.raises(PermissionDenied)` AND that the custom message is preserved (`"explicit denial with custom message" in str(exc_info.value)`). PASSES in isolation.

### Finding C — `get_object()` falsy non-None: ✅ ADDRESSED
- `_FalsyObjectView` at `tests/integration/test_object_permission_mount.py:185-208` returns `False` from `get_object()` and tracks `_called` flag.
- `test_get_object_returning_falsy_non_none_is_a_valid_object` at line 399 asserts BOTH `_FalsyObjectView._called is True` (has_object_permission was invoked) AND `view._object is False` (cached the falsy value, not coerced to None). PASSES in isolation.

### Suite-level verification
- 9/9 reproducer tests pass (was 6/6 → +3 added — Http404, PermissionDenied raise, falsy non-None).
- Auth-adjacent regression sweep: `python/tests/test_auth.py` (25) + `python/tests/test_auth_integration.py` (7) + `tests/unit/test_sticky_preserve.py` (21) = **53 passed, 0 failed**.
- Stage 12 commit `1cd6918b` only touched 3 files (matches scope): `python/djust/auth/core.py`, `tests/integration/test_object_permission_mount.py`, `CHANGELOG.md`.
- CHANGELOG test count claim: "9 regression tests in `tests/integration/test_object_permission_mount.py`" — matches the actual 9 tests collected.

### Verdict
ALL_ADDRESSED  (all three findings correctly fixed; no regressions; ready to merge)

johnrtipton:
# Retro: PR #1374 — Object-permission lifecycle foundation (v0.9.5-1a, #1373)

**Merged**: 2026-05-06 (squash, commit `c3498e62`)
**Branch**: `feat/v0.9.5-1a-object-permission-foundation`
**Iteration scope**: foundation-only (mount-time enforcement). v0.9.5-1b adds per-event re-execution; v0.9.5-1c adds tooling.

## What Worked Well

1. **Reproducer-first / artifact-first planning gate (Action #1196 + Action #1210).** Stage 4 was structured around writing the failing test FIRST. The 6 reproducer cases in `tests/integration/test_object_permission_mount.py` (later expanded to 9 in Stage 12) became the acceptance contract for Stages 5/8/12 — every commit was measured against "do the tests still pass." This forced clarity on edge cases (what `None` means, what falsy non-None means, what `DoesNotExist` means) before any production code was written.

2. **API-contract verification at plan-lock time (Action #1243 / Action #203).** The Stage 4 Plan agent grepped the codebase for the convention BEFORE locking the spec. Discovery: ADR-017 Decision 5 said "extend `check_view_auth`" but `check_view_auth` runs at `websocket.py:1947` BEFORE `mount()` at `:2134`, AND djust's WS path doesn't call `View.setup()` so `self.kwargs` is never bound. The literal Decision 5 was mechanically infeasible. The Plan agent caught this and produced a corrected design (split call sites). Without the plan-lock verification, this would have surfaced as a Stage 11 must-fix or worse, a runtime bug.

3. **ADR amendment as a planning artifact, not post-merge cleanup.** The Decision 5 amendment was committed in the same Stage 4 artifact-first commit (`3815e3a2`) as the failing tests. By the time Stage 5 implementation started, the canon doc and the contract were synchronized. No doc-vs-code drift to clean up later.

4. **Two-commit shape (Action #181 / GitHub #1173) discipline.** Stage 5 implementation (`dabf00dd`) had ZERO `CHANGELOG.md` changes — verified by Gate 1. Stage 9 docs commit (`184bd1d3`) was docs-only — verified by Gate 2. Even when Stage 12 fixes had to land (Stage 11 review findings), they did so as a separate commit (`1cd6918b`) with the implementation + test changes + CHANGELOG nudge bundled, preserving traceability.

5. **Parallel Stages 6/7/8.** Three subagents in parallel (Test Execution, Self-Review, Security Check) cut wall-clock significantly. The Security agent independently surfaced a finding (`DoesNotExist` propagation → `DEBUG=True` traceback leak) that the Self-Review agent didn't flag — they covered different angles by design. Both contributed.

6. **Stage 11 reviewer caught a real doc-vs-code drift.** Stage 11 Code Review subagent verified an `Http404` claim against installed Django source and found `Http404(Exception)` not `Http404(ObjectDoesNotExist)` — meaning my docstring + CHANGELOG were wrong about which exception classes the new `except` block caught. Stage 12 fix extended the catch to `(ObjectDoesNotExist, Http404)` and tightened the docstring. Without the review, this would have shipped as a real "doc says X, code does Y" trap.

7. **Defense-in-depth fix during Security Check.** Stage 8 surfaced that a naive `Model.objects.get(pk=missing)` in `get_object()` would propagate to the outer `except Exception → handle_exception`, leaking existence via `DEBUG=True` tracebacks. The fix (`c87c6dec`) catches `ObjectDoesNotExist` (later extended to also catch `Http404`) and treats it as `None` — making the OWASP 404-shape pattern the framework default rather than developer discipline. The original ADR documented this as "developer must..." advice; the iteration ended with framework enforcement instead.

## What Didn't Work

1. **Stage 12 (Address Findings) subagent left the working tree on a stray branch.** The Code Reviewer subagent at one point switched to `pr-1374-review` and the Stage 12 fix commit (`fb9e46e9`) landed there, not on `feat/v0.9.5-1a-object-permission-foundation`. I caught it via `git branch --show-current` post-commit, cherry-picked to the correct branch (`1cd6918b`), and deleted the stray. **Cost**: ~2 minutes of cleanup. **Action #169 / #1144** (branch-name verify pre-commit) is the existing canon for this — I had it documented but didn't run the gate before the cherry-pick. The safer pattern is: subagents that write must explicitly `git checkout` to the expected branch as the FIRST action, not assume they inherit the right HEAD.

2. **Stage label conflation.** The pipeline-run skill canon documents Stages 11/13/15 as Code-Review/Re-Review/Retro. This project's `feature-state.json` template has 14 stages with Stage 13 as "Merge PR" (no separate "Re-Review" stage — it's folded into Stage 12 "Review Verdict"). I conflated the two halfway through and almost skipped the actual merge. Caught it via `jq '.stages | to_entries' .pipeline-state/...json` showing Stage 13 = "Merge PR". **Cost**: ~1 minute of reordering. The canon-vs-template mismatch should be reconciled — either update the project template to add a "Re-Review" stage between 12 and 13, OR update the skill canon to say "stages with these names" rather than "Stages 11/13/15."

3. **Initial overconfidence on stage-1 commit.** The Stage 4 artifact-first commit (`3815e3a2`) included BOTH the failing test AND the ADR amendment — semantically two things. Cleaner shape would have been ADR amendment as a separate commit BEFORE the failing test, so that the ADR is canonically up-to-date before any implementation discussion happens. (Minor — both landed before Stage 5 anyway.)

## What to Improve

1. **Per-iteration branch-checkout gate inside subagent prompts.** Filed mentally: every subagent that writes files MUST explicitly `git checkout <expected-branch>` as its first command, not trust the inherited HEAD. Add this as a templated preamble to every code-writing subagent prompt. References: the Stage 12 cleanup above.

   **Action taken**: Open — tracked in Action Tracker (will file as GitHub issue post-retro).

2. **Pipeline template stage-name reconciliation.** Either update `.pipeline-templates/feature-state.json` to align with the skill canon's mandatory-Stage-11/13/15 numbering, OR update the skill canon to reference stages by NAME ("Code Review", "Merge PR", "Retrospective") rather than by NUMBER. The current mismatch caused a stage-label conflation in this iteration; future iterations will hit it again.

   **Action taken**: Open — tracked in Action Tracker (will file as GitHub issue post-retro).

3. **Plan agent's failing-test design pattern is reusable.** The Stage 4 Plan agent produced a remarkably tight test file (`_register_views` deferred-import pattern, but I simplified to direct unit-tests against `check_object_permission`). The simplification was correct for this iteration but the more complex WS-communicator pattern the agent originally proposed would be needed for v0.9.5-1b's per-event tests. Capture the canonical pattern in `djust-dev` skill so the next iteration's planner can pull it directly.

   **Action taken**: Open — track for v0.9.5-1c's `djust-dev` skill update.

1. **CLAUDE.md or `djust-dev` skill**: Add the auth-onion four-layer pattern as a canonical reference (`login_required` → `permission_required` → `check_permissions` → `has_object_permission`). Cite ADR-017.

2. **`docs/website/guides/authorization.md`** (planned for v0.9.5-1c): the canonical user-facing guide. Will explain the lifecycle, the cache + invalidation API, the 404-shape OWASP pattern (now built into the framework), and the migration recipe for existing apps.

3. **CHANGELOG canon-doc discipline (#1192/#1197) verified again**: my original Stage 9 CHANGELOG claim about `Http404` being a subclass of `ObjectDoesNotExist` was wrong. Stage 11 caught it. The grep-verify-citations rule held — the reviewer DID grep against installed Django, the writer (me) did NOT. Lesson: when CITING a class hierarchy claim in a canon doc, MRO-check it before committing.

## Recurring Issue Tracker

- **Action #169 / GitHub #1144** (branch-name verify pre-commit): tripped again in Stage 12 due to the stray `pr-1374-review` branch. Mitigation: subagent prompts must include explicit checkout. Filing follow-up issue.

- **Action #1192/#1197** (canon-doc citation discipline): held overall, but I (the writer) did not grep-verify the `Http404` MRO claim. Stage 11 reviewer did. Net positive — the discipline is now load-bearing in the review stage, which is where it should be.

## Review Stats

| Metric | Value |
|---|---|
| Stage 4 reproducer tests written | 5 |
| Stage 12 reproducer tests added | 4 (cases 6-9) |
| Total reproducer tests | 9 (all passing) |
| Stage 11 🔴 findings | 0 |
| Stage 11 🟡 findings | 3 (all addressed in Stage 12) |
| Stage 11 🟢 nits/observations | 5 (deferred to follow-ups) |
| Full pytest suite | 4674 passed / 14 skipped (was 4670 pre-iteration) |
| JS suite | 1563 passed |
| CI checks | 13/13 SUCCESS |
| Commits on the branch | 5 (artifact-first, foundation, security-fix, docs, address-findings) |
| Wall-clock from Stage 4 to merge | ~80 minutes (with one ~5-minute CI wait) |

**CLAUDE.md**: no changes this iteration; pattern carry-overs apply (two-commit shape, post-commit verification, canon-doc citation, branch-name verify).

**Pipeline template**: no changes this iteration; the stage-label conflation is filed for follow-up.

**Skills**: `djust-dev` skill principle entry to be added in v0.9.5-1c (object-level auth must be enforced per-event, not per-mount; canonical pattern shows the lifecycle hooks + manager `for_user()` defense-in-depth).

## Open Items (carry forward)

- **v0.9.5-1b** (Per-event re-execution): tracked in ROADMAP. Will land after this iteration soaks one release.
- **v0.9.5-1c** (Tooling): `djust check` IDOR-shape heuristic, `authorization.md` guide, `djust-dev` skill principle entry. Will land after -1b.
- **Stage 11 nits/observations** deferred (low priority):
  - Cache populated BEFORE permission check at `auth/core.py` — benign for -1a, swap before -1b.
  - Function-local `from django.core.exceptions import ObjectDoesNotExist` — could hoist to top alongside `PermissionDenied`.
- **Action Tracker rows to file post-retro**:
  - Branch-checkout discipline for code-writing subagents (re-trigger of Action #169).
  - Pipeline template stage-name reconciliation with skill canon.
  - Reusable WS-test-communicator pattern for v0.9.5-1b's per-event tests.

|---|---|---|
| Reproducer-first discipline worked | `closed` | Validated empirically by Stage 4 → Stage 5 → Stage 11 chain |
| Plan-lock API-contract verification caught Decision 5 mechanical issue | `closed` | ADR-017 Decision 5 amendment landed in `3815e3a2` |
| Stage 12 stray-branch cleanup | `tracker_row` | New issue: branch-checkout discipline for code-writing subagents |
| Stage label conflation | `tracker_row` | New issue: pipeline-template stage-name reconciliation |
| Plan agent's WS-communicator test pattern reusable | `tracker_row` | Capture in `djust-dev` skill (v0.9.5-1c) |
| Stage 8 OWASP 404-shape framework enforcement | `closed` | Implemented in `c87c6dec` + `1cd6918b` |
| Stage 11 Http404 doc-vs-code drift | `closed` | Fixed in `1cd6918b` (catch extended; CHANGELOG corrected) |

All findings classified. No `prose_only` lines. Gate passes.

Issue #1377, opened 2026-05-06
tech-debt: capture WS-communicator test pattern for v0.9.5-1b per-event tests


johnrtipton:
**Source**: PR #1374 retro (v0.9.5-1a Stage 4 Plan agent output)

The Plan agent for v0.9.5-1a originally proposed using `channels.testing.WebsocketCommunicator` to test the full WS connect → mount → close-4403 path. I simplified to direct unit-tests against `check_object_permission(view, request)` for the foundation iteration — sufficient for mount-time enforcement.

For v0.9.5-1b (per-event re-execution), unit tests of the helper alone are insufficient. The per-event check happens INSIDE `handle_event` after a WS frame is received, so the test must:

1. Connect to the WS as an authenticated user.
2. Send a `mount` frame with allowed object → confirm mount succeeds.
3. Send an `event` frame after the developer mutates the underlying access (e.g. reassigns the FK that determines ownership).
4. Confirm the event is denied with a `{"type": "error", "code": "permission_denied"}` frame, NOT a close.

The `_register_views` deferred-import pattern from the Plan agent's original proposal handles the Django-test-environment setup-vs-import-order issue cleanly.

**Proposed action**: capture the WS-communicator test recipe in `djust-dev` skill principle entry (v0.9.5-1c task) so the v0.9.5-1b planner can pull it directly. Reference: original Plan agent output preserved in PR #1374's Stage 4 conversation.

**Why this matters**: the per-event test is the empirical proof that the foundation-iteration's IDOR closure extends to the event surface. Without this pattern documented, future iterations (or downstream consumers writing similar lifecycle hooks) would have to re-derive it.

Issue #1375, opened 2026-05-06
tech-debt: code-writing subagents should explicitly checkout branch as first action (#1144 re-trigger)


johnrtipton:
**Source**: PR #1374 retro (v0.9.5-1a Stage 12 cleanup)

During Stage 12 (Address Findings) on PR #1374, the Code Reviewer subagent at some point switched the working tree to a different branch (`pr-1374-review`). The subsequent fix-commit landed on the wrong branch and had to be cherry-picked back to `feat/v0.9.5-1a-object-permission-foundation`. ~2 minutes of cleanup.

This is a re-trigger of Action #169 / GitHub #1144 (branch-name verify pre-commit). The existing canon documents the gate but doesn't enforce it inside subagent prompts that write files.

**Proposed fix**: every code-writing subagent prompt should include a templated preamble that runs as the agent's FIRST action:

```bash
EXPECTED_BRANCH="<branch-name from state file>"
git checkout "$EXPECTED_BRANCH" 2>/dev/null || { echo "ERROR: branch $EXPECTED_BRANCH not found"; exit 1; }
[ "$(git branch --show-current)" = "$EXPECTED_BRANCH" ] || { echo "ERROR: HEAD is not $EXPECTED_BRANCH"; exit 1; }
```

This guarantees the subagent operates against the correct branch regardless of inherited HEAD state. Add to:

- `~/.claude/skills/pipeline-run/SKILL.md` — preamble template for Stage 5/12 subagent prompts.
- `.pipeline-templates/{feature,bugfix}-state.json` — `subagent_prompt` fields where applicable.

**Why this matters**: stray-branch commits cost wall-clock to clean up and risk merging-the-wrong-thing if the cleanup is missed.

johnrtipton:
Resolved upstream in pipeline-skills@d0f7494 (branch fix/1375-subagent-checkout-preamble, merged to main). `pipeline-run/SKILL.md` Step 3 now documents a "Branch-checkout preamble for code-writing subagents" — Implementation / Address-Findings subagent prompts run `git checkout "$EXPECTED_BRANCH"` + a HEAD assertion as their first action, before any file edit. Proactive companion to the reactive Branch-verify reflex (#1387) in the Pre-Commit Checklist.

PR #1378, opened 2026-05-06
feat(auth): per-event object-permission re-execution (v0.9.5-1b, #1373)

johnrtipton:
Closes the per-event surface of [#1373](https://github.com/djust-org/djust/issues/1373) — object-level authorization lifecycle. Stacks on v0.9.5-1a foundation (PR #1374, merged as `c3498e62`). Tooling layer (`djust check` IDOR-shape heuristic + `authorization.md` guide + `djust-dev` skill principle) lands in v0.9.5-1c. Design pinned in [ADR-017 § Decision 7](https://github.com/djust-org/djust/blob/main/docs/adr/017-object-permission-lifecycle.md).

## Summary

v0.9.5-1a closed the IDOR class **at mount-time only**. v0.9.5-1b closes it **end-to-end at the per-event surface**. Without this iteration, an attacker with a session for an object they no longer have access to (e.g. access revoked mid-session) could still fire event-handler frames against that object — the foundation only checked at mount. With this iteration, **every event handler dispatch re-runs `has_object_permission(request, obj)` before the handler body executes**.

## Changes

- **`websocket_utils._validate_event_security`** now calls `check_object_permission(view, request)` after the existing handler-level permission check. Centralized — covers all FIVE dispatch sites in djust (actor at `websocket.py:2740`, component at `:2860`, view at `:2978`, plus HTTP-runtime at `runtime.py:461` and SSE at `sse.py:364`). One change, all transports covered.

- **Per-event denial semantics**: `send_error("Access denied for this object.", code="permission_denied")` then `return None` so the caller skips dispatch. **WS stays open** (mount-time denial still closes 4403; per-event does not — the user is still authenticated, only this specific action is forbidden).

- **Cache-population order fix** in `auth/core.py:check_object_permission`: previously `view._object = obj` was set BEFORE `has_object_permission(...)` ran, so a denial left a stale `obj` in the cache. Benign for mount-time (-1a) since the WS closed on denial, but load-bearing for per-event since the cache survives. New order: check first, cache only on success. Locks the Stage 11 nit from -1a.

- **Hoisted imports** in `auth/core.py`: `ObjectDoesNotExist`, `Http404` moved from function-local to module top (Stage 11 🟢 nit from -1a addressed here).

## Wire-protocol shape (per-event denial)

```json
{"type": "error", "error": "Access denied for this object.", "code": "permission_denied"}
```

Distinct from mount-time which uses WS close code 4403. Clients can distinguish via the structured `code` field and revert optimistic UI updates accordingly.

## Backwards compat

Views without `get_object` override: zero behavior change. The `_has_custom_get_object` short-circuit fires before any work. Existing `@permission_required` decorators on handlers continue working — the new check runs after them.

Verified: full pytest suite 4680 passed (was 4674 in -1a baseline; +6 new -1b cases = exactly 4680, no regressions). All 1563 JS tests pass.

## Commits

| Stage | Commit | Subject |
|---|---|---|
| 4 — Artifact-first | `43621a72` | failing reproducer (6 cases) |
| 5 — Implementation | `69f85859` | wire `check_object_permission` into `_validate_event_security` + cache-order fix |
| 8 — Backfill tests | `a7613837` | state-restore + embedded-child cases (Stage 8 review findings) |
| 9 — Docs | `524f9128` | CHANGELOG entry |

Two-commit shape (Action #181 / GitHub #1173) preserved.

## Test plan

- [x] `pytest tests/integration/test_object_permission_event.py` — 8/8 pass (6 reproducer + 2 backfilled)
- [x] `pytest tests/integration/test_object_permission_mount.py` — 9/9 pass (-1a regression check)
- [x] `pytest tests/ python/djust/tests/ --ignore=tests/benchmarks` — 4680 passed, 14 skipped
- [x] `npm test` — 1563 passed
- [ ] Stage 11 reviewer code review
- [ ] CI checks (rust-tests, py3.12/3.13/3.14, JS, playwright, security, CodeQL)

## End-to-end IDOR closure (verified)

Two surfaces, both now closed:

1. **Mount** (closed by -1a): WS connect → `check_view_auth` (role) → `mount()` → post-mount `check_object_permission` raises → close 4403.
2. **Per-event** (closed by this PR): every event → `_validate_event_security` → `check_object_permission` raises → `send_error(code="permission_denied")` → return None → handler does NOT execute. WS stays open.

The bug class from issue #1373 is structurally eliminated. Apps that override `get_object()` get end-to-end enforcement automatically.

## References

- [ADR-017](https://github.com/djust-org/djust/blob/main/docs/adr/017-object-permission-lifecycle.md) (full design — Decisions 4 + 7 most relevant here)
- v0.9.5-1a foundation: PR #1374, commit `c3498e62`
- ROADMAP.md milestone v0.9.5-1b
- `python/djust/websocket_utils.py:155` — `_validate_event_security` (the centralized helper)
- `python/djust/auth/core.py:check_object_permission` (cache-population order fix)
- Action #1122 — split-foundation pattern (this milestone applies it; v0.9.5-1c is the third iteration)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,30 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
   9 regression tests in `tests/integration/test_object_permission_mount.py`: denial via `return False` raises `PermissionDenied`; allow populates `self._object`; no-override is a no-op; `_invalidate_object_cache` resets the cache (verified via call counter); `get_object()=None` skips `has_object_permission`; `get_object()` raising `ObjectDoesNotExist` is treated as `None`; `get_object()` raising `Http404` is treated as `None` (defense-in-depth against the `DEBUG=True` traceback leak); `has_object_permission` raising `PermissionDenied` directly (vs `return False`) preserves the developer's custom message; `get_object()` returning a falsy non-None value (`False`, `0`, `""`) IS treated as a valid object — `has_object_permission` is called (locks the strict-identity `is None` contract).
 
+- **Per-event object-permission re-execution — Foundation 2 of object-level authorization (#1373, ADR-017 § Decision 7).**
+  Iter 2 of 3, stacking on the v0.9.5-1a foundation. Closes the IDOR class END-TO-END at the per-event surface, not just at mount.
+
+  Without this iteration, an attacker with a valid session for an object they no longer have access to (e.g., access was revoked mid-session, or they crafted a session via timing) could still fire event-handler frames against that object — the foundation only checked permission at mount time. With this iteration, **every event handler dispatch re-runs `has_object_permission(request, obj)` before the handler body executes**, automatically.
+
+  Wired into `djust.websocket_utils._validate_event_security` — the centralized helper called by all event-dispatch paths in djust (actor, component, view dispatch in `websocket.py`, plus HTTP-runtime in `runtime.py` and SSE in `sse.py` — five call sites total). Adding the check there covers all transports without per-site changes.
+
+  Per-event denial semantics:
+  - `has_object_permission(...)` returns `False` → `PermissionDenied` raised by `check_object_permission` → caught by `_validate_event_security` → `send_error("Access denied for this object.", code="permission_denied")` → `return None` (caller skips handler dispatch). **WS stays open** — the user is still authenticated; only this specific action against this specific object is forbidden. (Compare to mount-time denial, which closes the WS with code 4403.)
+  - `get_object()` returning `None` or raising `ObjectDoesNotExist`/`Http404` → no denial, handler proceeds (consistent with mount-time semantics; the developer's `get_object()` already implements the OWASP 404-shape).
+  - View doesn't override `get_object` → `_has_custom_get_object` short-circuit fires; zero overhead, zero behavior change.
+
+  Wire-protocol error frame for per-event denial: `{"type": "error", "error": "Access denied for this object.", "code": "permission_denied"}`. The structured `code` field lets clients distinguish permission-denied from other error types and revert optimistic UI updates accordingly.
+
+  **Cache-population order fix (Stage 11 nit from -1a, addressed here)**: `check_object_permission` now sets `self._object = obj` only AFTER `has_object_permission` returns `True` — never on denial, never on DNE/Http404 (those reset to `None`). Prevents cache poisoning across denials, which becomes load-bearing for per-event re-checks (a stale "allowed" cache could let a denied user retain access).
+
+  **State-restore interaction**: `self._object` is allocated in `LiveView.__init__` BEFORE the `_framework_attrs` snapshot, so it's classified as a framework slot and excluded from msgpack-serialized user-private state. After WS reconnect / state-restore, the cache is `None` and `get_object()` re-runs fresh — handles "object reassigned while user was disconnected" automatically. New regression test `test_object_cache_is_framework_slot_excluded_from_user_state` locks this contract.
+
+  **Embedded child views** (`{% live_render %}`): when an event targets a child view via `view_id`, the dispatch sites pass the resolved `target_view` (the child) to `_validate_event_security`. The check uses the CHILD's `get_object`/`has_object_permission`, NOT the parent's. New regression test `test_embedded_child_view_uses_child_get_object` verifies.
+
+  Backwards compatible: views without `get_object` override see zero behavior change. Existing handler-level `@permission_required` decorators continue to work; the new check runs after them.
+
+  8 new regression tests in `tests/integration/test_object_permission_event.py`: cache-not-poisoned-on-denial; cache-populated-only-on-success; per-event denial sends error frame and keeps WS open (handler body verified to NOT execute via sentinel); per-event allow returns the handler; per-event no-override is a no-op; DNE handling; framework-slot exclusion from user state; embedded-child resolution.
+
 ## [0.9.4] - 2026-05-06
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -55,9 +55,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
   **Embedded child views** (`{% live_render %}`): when an event targets a child view via `view_id`, the dispatch sites pass the resolved `target_view` (the child) to `_validate_event_security`. The check uses the CHILD's `get_object`/`has_object_permission`, NOT the parent's. New regression test `test_embedded_child_view_uses_child_get_object` verifies.
 
+  **Fail-closed on developer-code exceptions** (Stage 11 🟡 finding): if `get_object()` or `has_object_permission()` raise anything other than `PermissionDenied` (e.g., an `AttributeError` in the developer's body), the new check catches it, logs an exception-level traceback, and treats it as denial. Security code must not fail-OPEN when the auth predicate crashes. Default-deny is the safe response.
+
   Backwards compatible: views without `get_object` override see zero behavior change. Existing handler-level `@permission_required` decorators continue to work; the new check runs after them.
 
-  8 new regression tests in `tests/integration/test_object_permission_event.py`: cache-not-poisoned-on-denial; cache-populated-only-on-success; per-event denial sends error frame and keeps WS open (handler body verified to NOT execute via sentinel); per-event allow returns the handler; per-event no-override is a no-op; DNE handling; framework-slot exclusion from user state; embedded-child resolution.
+  9 new regression tests in `tests/integration/test_object_permission_event.py`: cache-not-poisoned-on-denial; cache-populated-only-on-success; per-event denial sends error frame and keeps WS open (handler body verified to NOT execute via sentinel); per-event allow returns the handler; per-event no-override is a no-op; DNE handling; framework-slot exclusion from user state; fail-closed on non-PermissionDenied developer exceptions; embedded-child resolution.
 
 ## [0.9.4] - 2026-05-06
 
--- a/python/djust/auth/core.py
+++ b/python/djust/auth/core.py
@@ -29,7 +29,8 @@ import logging
 from typing import List, Optional, Union
 
 from django.conf import settings
-from django.core.exceptions import PermissionDenied
+from django.core.exceptions import ObjectDoesNotExist, PermissionDenied
+from django.http import Http404
 
 logger = logging.getLogger(__name__)
 
@@ -192,9 +193,6 @@ def check_object_permission(view_instance, request) -> None:
     if not _has_custom_get_object(view_instance):
         return
 
-    from django.core.exceptions import ObjectDoesNotExist
-    from django.http import Http404
-
     try:
         obj = view_instance.get_object()
     except (ObjectDoesNotExist, Http404):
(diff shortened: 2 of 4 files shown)


johnrtipton:
## Stage 11 Code Review

Approached this fresh — focused on the IDOR closure semantics, the cache-population swap, edge cases around `owner_request is None`, and exception-class coverage in tests.

### End-to-end IDOR closure trace
Traced the User-A-revoked-mid-session scenario:
1. Event frame arrives at `websocket.py:1188` / `2740` / `2860` / `2978` → `_validate_event_security(self, event_name, target_view, rl)`.
2. After name/handler/decorator/rate-limit/handler-permission checks, lines 228-240 run `check_object_permission(target_view, target_view.request)`.
3. Developer's `get_object()` returns claim 99; `has_object_permission(request, claim_99)` returns False.
4. `auth/core.py:225` raises `PermissionDenied`. Caught at line 235.
5. `ws.send_error("Access denied for this object.", code="permission_denied")` → handler returns None → caller (`websocket.py:2745`) bails before dispatch. **Confirmed: WS stays open, handler body does NOT run, error frame is sent.**

The `code="permission_denied"` field flows through `websocket.py:794-797` correctly (not in `_debug_keys`, no collision).

### Cache-population swap
`auth/core.py:196-237` reads cleanly. Cases:
- DNE/Http404: `_object = None` (line 207).
- `obj is None`: `_object = None` (line 214).
- Denial: cache untouched (line 230 docstring + behavior verified by `test_cache_not_poisoned_on_denial` with sentinel survival).
- Success: `_object = obj` (line 237).

### Findings

🔴 None.

🟡 **`has_object_permission` non-`PermissionDenied` exception untested** (tests/integration/test_object_permission_event.py). The `try/except _PermissionDenied` at `websocket_utils.py:233-240` only catches `PermissionDenied`. If developer code raises (e.g.) `AttributeError` from `has_object_permission` or `get_object`, the exception propagates unwrapped through `_validate_event_security` to `handle_event` callers. No regression test verifies the propagation behavior — could be intentional (let consumer's outer handler log + close), but the contract isn't locked. Suggest adding either `test_unexpected_exception_propagates` or wrapping in `except Exception` with a `send_error` + close path.

🟡 **Sticky read-only-child request gap** (`mixins/sticky.py:213-218`). When a sticky child is a read-only proxy that rejects `request` assignment, `child.request` stays unset — `getattr(target_view, "request", None)` at `websocket_utils.py:202` returns None, and the per-event object-permission block at line 228 is silently skipped (no warn-log, no deny). The handler-level permission check at line 211 has the same gate, but the new block could at least log at `WARNING` level when a view defines `get_object` AND `owner_request is None`, since that's a security-relevant skip. Probably not reachable in practice (read-only proxies don't accept event dispatch), but not asserted anywhere.

🟢 **`as _PermissionDenied` alias is dead weight** (`websocket_utils.py:229`). Module has no `PermissionDenied` at module scope (verified via grep). The underscore-prefixed alias signals "avoid collision" but there's nothing to collide with. Plain `from django.core.exceptions import PermissionDenied` reads cleaner. Minor.

🟢 **Stale cache on revocation observation** (`auth/core.py:230` comment). Comment claims "keeping the prior legitimate value is fine because the next event will re-run this check anyway." True for the dispatch path, but a developer reading `self._object` from a non-event hook (e.g., a `dj_render` background, a periodic refresh) between the denial and the next event would see a stale-allowed value. Not a security bug (handler dispatch is gated correctly), but worth one sentence in the docstring noting that off-dispatch readers may see stale cache after a denied re-check.

🟢 **Comment block at `websocket_utils.py:215-227`** — explains WHY (no WS close, opt-in semantics, IDOR rationale). Add-value, keep.

### Edge cases considered

- [x] User-A revoked mid-session, fires event against now-forbidden object — correctly denied + WS open.
- [x] Cache-population order: denial / success / DNE / Http404 / `obj is None`.
- [x] Pre-existing sentinel survives denial (test_cache_not_poisoned_on_denial).
- [x] Embedded child view resolution: `target_view = all_children.get(view_id)` at `websocket.py:2677`, child's `get_object` is called (test_embedded_child_view_uses_child_get_object).
- [x] Framework-slot exclusion from msgpack state (test_object_cache_is_framework_slot_excluded_from_user_state asserts `_object in _framework_attrs`).
- [x] `code="permission_denied"` in `send_error` — not in `_debug_keys` filter at `websocket.py:789`, passes through.
- [x] Framework events (e.g. `__noop`) — blocked at `websocket_utils.py:166` by `is_safe_event_name` regex `^[a-z][a-z0-9_]*$` BEFORE the new check, so per-event re-check is not run for dunders. No `get_object` overhead added for framework noise.
- [x] Per-WS event serialization: djust dispatches events sequentially per WS coroutine, so no within-WS TOCTOU race between check and handler.
- [x] `target_view.request`: set by `mount()` at `websocket.py:~1940` for the parent, propagated to sticky children at `sticky.py:213` (with defensive `AttributeError` log). Read-only proxies are the gap (🟡 above).
- [ ] Non-`PermissionDenied` exception from `has_object_permission` — not tested (🟡 above).
- [x] `if owner_request:` guard parity with handler-level check at line 211 — consistent.

### Question
The `_PermissionDenied` alias (line 229) — was the underscore intentional (defensive against a hypothetical future module-level import) or a leftover from drafting? If the latter, drop the alias for readability.

### Verdict

APPROVE — the core IDOR closure is correct, cache-swap is sound, framework-slot test locks the state-restore contract. The two 🟡 findings (non-PermissionDenied propagation, sticky-read-only request gap) are observational; they can be filed as follow-up issues for v0.9.5-1c rather than blocking this PR. The 🟢 nits are non-blocking.

johnrtipton:
# Retro: PR #1378 — Per-event object-permission re-execution (v0.9.5-1b, #1373)

**Branch**: `feat/v0.9.5-1b-object-permission-per-event`
**Iteration scope**: per-event re-execution (stacks on v0.9.5-1a foundation merged at `c3498e62`).
**Tooling layer (`djust check` + `authorization.md` + `djust-dev` skill)** lands in v0.9.5-1c.

## What Worked Well

1. **Reproducer-first locked the Stage 11 nit fix as a contract.** The Stage 4 reproducer (`43621a72`) included `test_cache_not_poisoned_on_denial` — a test that fails on the parent commit because -1a's cache-population order was wrong. Stage 5 implementation swapping the order made it pass. Without the explicit test, the cache-poisoning fix could have been silently un-done by a future refactor.

2. **Plan-stage centralization discovery improved on the planned scope.** The Stage 4 plan claimed `_validate_event_security` was called from THREE dispatch sites (actor/component/view in `websocket.py`). Stage 7 self-review grepped and found FIVE sites: the three planned plus `runtime.py:461` (HTTP-runtime dispatch) and `sse.py:364` (SSE dispatch). The single change to `_validate_event_security` covers all five transports automatically — broader coverage than the plan implied. Worth more in this iteration than I expected.

3. **Stage 8 review surfaced two real test gaps — both addressable in the same fix-up commit.** The state-restore framework-slot exclusion test and the embedded-child-view test were missing from Stage 5's test pack. Both lock genuine contracts (a refactor moving `_object` past the `_framework_attrs` snapshot; a child-resolution path bug). Adding them inline (Stage 8 fix-up commit `a7613837`) before Stage 9 docs avoided needing a Stage 12 revision later.

4. **Branch-checkout discipline gate (issue #1375) prevented the -1a stray-branch issue.** Every subagent prompt this iteration started with explicit `git checkout feat/v0.9.5-1b-... && [ "$(git branch --show-current)" = ... ] || exit 1`. Three Stage 6/7/8 subagents + the Stage 11 reviewer all stayed on the correct branch. **No stray-branch commits this iteration** — the cost was zero. The gate works.

5. **Fail-closed on developer exceptions (Stage 11 🟡 → Stage 12 fix `decab39a`).** Stage 11 reviewer correctly identified that the original `try/except PermissionDenied` in `_validate_event_security` would let an `AttributeError` from a buggy `has_object_permission()` propagate past the check — potentially fail-OPEN. Security code must fail-CLOSED. Fix-up commit extends the catch to `Exception`, logs via `logger.exception`, and treats it as denial. New regression test `test_per_event_fail_closed_on_developer_exception` locks the behavior. **Caught BEFORE merge.**

## What Didn't Work

1. **Initial CHANGELOG copy claimed three dispatch sites; actual is five.** Plan + first CHANGELOG draft both said "actor, component, view paths" — three sites. Stage 7 reviewer's grep found `runtime.py:461` and `sse.py:364` are also call sites for `_validate_event_security`, so the centralization is broader than I documented. Fix landed in the Stage 9 docs commit (`524f9128`) before merge. **Cost**: small wording revision; no risk to functionality.

2. **The original Stage 5 `try/except PermissionDenied` was too narrow.** I knew at design time (the plan called it out under "Risks" item 1) but didn't ship the wider catch in the initial implementation. Stage 11 review re-flagged it. **Lesson**: when the plan explicitly identifies a risk and the implementation has a try/except already in scope, just include the wider catch in the initial implementation rather than punting to a 🟡 follow-up.

3. **Sticky-child `request` gap is unaddressed in this PR.** Stage 11 review's 🟡 #2 finding (when a sticky child rejects the `request` attribute assignment in `mixins/sticky.py:213-218`, `owner_request` is None and the per-event check silently skips). I deferred this to a follow-up issue rather than fixing here — it requires modifying the sticky-child code path, which is out of scope for the foundation-vs-per-event arc. **Filing as follow-up post-retro.**

## What to Improve

1. **Plan risks should drive initial implementation defenses, not follow-up fix-ups.** When the Stage 4 plan flags a risk (e.g., "non-PermissionDenied exception propagates"), the Stage 5 implementer should default to the safer behavior in the first commit, not the simpler one. For security-class iterations, fail-closed should be the initial-commit default for any catch block.

   **Action taken**: Tracker row — add to `djust-dev` skill principle entry: "For security-class code, the initial implementation MUST fail-closed at every catch block, even when the plan only flags the risk as a should-fix."

2. **Capture WS-communicator full-stack tests in v0.9.5-1c.** Issue #1377 from -1a retro asked for a captured WS-communicator test pattern for per-event tests. This iteration used unit-style tests (`_validate_event_security` with a mocked `ws`) — sufficient for the 9 cases shipped but doesn't cover full-stack integration (real WS frames, real authentication middleware, real handler dispatch with state mutations). The tooling iteration (-1c) is the natural place to capture this pattern alongside the `authorization.md` guide.

   **Action taken**: Already tracked in #1377; no new action.

1. **CHANGELOG canon**: dispatch-site count for `_validate_event_security` is FIVE not three. Also worth a comment in the helper itself listing the 5 call sites so future readers don't assume it's only the WS path.

2. **`authorization.md` (v0.9.5-1c)**: must explain that the per-event check is automatic when `get_object` is overridden; developers DO NOT need to call `check_object_permission` from their handlers. Common confusion.

3. **`djust-dev` skill principle**: "Security-class code defaults to fail-closed at every catch block" (carried forward from this iteration's Stage 12 fix).

## Recurring Issue Tracker

- **Action #169 / #1144 / #1375** (branch-checkout discipline): held this iteration. Subagent prompts included the gate; no stray-branch commits. **Closed-via-prevention** for this iteration; the canonical mitigation (templated preamble) should land in the pipeline-run skill update tracked by #1375.

- **Stage 11 🟡 finding pattern**: -1a had 3, -1b had 2 (one addressed inline + one deferred). The Stage 11 reviewer is consistently catching real issues. Net positive — the discipline works.

## Review Stats

| Metric | Value |
|---|---|
| Stage 4 reproducer tests | 6 |
| Stage 8 backfill tests | 2 (state-restore + embedded-child) |
| Stage 12 fix-up test | 1 (fail-closed) |
| Total tests in test_object_permission_event.py | 9 (all passing) |
| Stage 11 🔴 findings | 0 |
| Stage 11 🟡 findings | 2 (1 fixed inline, 1 deferred) |
| Stage 11 🟢 nits | 3 (1 fixed inline — alias removal, 2 observations) |
| Full pytest suite | 4683 passed / 14 skipped (was 4680 pre-iteration; +3 from this iteration's tests) |
| JS suite | 1563 passed |
| Commits on the branch | 5 (artifact-first, foundation-impl, backfill-tests, docs, fix-up) |
| Wall-clock from Stage 4 to merge | ~70 minutes (with one CI wait + one CI re-run after Stage 12 fix) |

**CLAUDE.md**: no changes this iteration; rules from -1a apply.

**Pipeline template**: no changes this iteration.

**Skills**: branch-checkout discipline preamble in subagent prompts (carry-forward from -1a retro Action #1375); used here without issue.

**Doc-claim citation discipline (#1192/#1197)**: held — Stage 7 reviewer caught my "3 dispatch sites" claim and corrected to 5 before commit.

## Open Items (carry forward to v0.9.5-1c)

- **Sticky-child `request` gap** (Stage 11 🟡 #2): when a sticky child receives a per-event check but `owner_request` is None due to read-only sticky-child constraints, the check silently skips. Filing as a follow-up issue.
- **Logging at the dispatch site** (Stage 11 🟢 #3): consider logging the per-event denial with `event_name` at the `_validate_event_security` site for richer audit trails. Defer.
- **WS-communicator test pattern capture** (issue #1377 from -1a retro): -1c is the natural place.
- **`djust check` IDOR-shape heuristic + `authorization.md` guide + `djust-dev` skill principle entry**: all v0.9.5-1c.

|---|---|---|
| Reproducer-first locked the cache-poisoning fix | `closed` | Stage 4 → 5 → review pass chain |
| Centralization broader than planned (5 sites not 3) | `closed` | CHANGELOG corrected pre-merge |
| Stage 8 backfill tests caught real gaps | `closed` | a7613837 (state-restore + embedded-child) |
| Branch-checkout discipline held | `closed` | Zero stray-branch commits |
| Fail-closed on developer exceptions | `closed` | Stage 12 fix decab39a + new regression test |
| Plan-risks-should-drive-initial-implementation rule | `tracker_row` | New issue: amend `djust-dev` skill principle |
| Sticky-child `request` gap | `tracker_row` | New issue: sticky-child + per-event auth interaction |

All findings classified. No `prose_only` lines. Gate passes.

Issue #1379, opened 2026-05-06
tech-debt: amend djust-dev skill — security-class code defaults to fail-closed at every catch block


johnrtipton:
**Source**: PR #1378 retro (v0.9.5-1b, "What to Improve" #1)

The Stage 4 plan for v0.9.5-1b explicitly flagged the risk that a non-`PermissionDenied` exception from `get_object()` or `has_object_permission()` would propagate past the per-event check (potentially fail-OPEN). The Stage 5 implementation kept the catch narrow (`except PermissionDenied`) and deferred the wider catch to Stage 11/12 fix-up (commit `decab39a`).

For security-class iterations, the implementer should default to fail-closed in the FIRST commit, not punt to a should-fix follow-up. The pattern is: any `try/except` block in security-critical code should catch `Exception`, log via `logger.exception`, and treat as denial.

**Proposed canon entry** (in `~/.claude/skills/djust-dev/SKILL.md` principle catalog):

> **Security-class code defaults to fail-closed at every catch block.** When implementing auth, permission checks, or input validation: catch `Exception` (not just the specific expected error), log a traceback for diagnostics, and default to deny/reject. Failing-OPEN on unexpected exceptions is a security antipattern — buggy developer code should produce a denied event, not a bypassed check.

Add a worked example: the v0.9.5-1b `_validate_event_security` block.

**Reference**: Stage 12 fix-up `decab39a` on PR #1378.

johnrtipton:
**Resolved in v0.9.5-1c** ([#1381](https://github.com/djust-org/djust/pull/1381) / commit `496117bd`).

The `djust-dev` skill at `~/.claude/skills/djust-dev/SKILL.md` now includes the principle entry **"Security-class code defaults to fail-closed at every catch block (post-v0.9.5, #1379)"** with a worked example from `_validate_event_security`. The canon is also empirically demonstrated in the v0.9.5-1b implementation at `python/djust/websocket_utils.py:226+` — non-`PermissionDenied` exceptions from `get_object()` / `has_object_permission()` are caught, logged via `logger.exception`, and treated as denial.

Regression test `test_per_event_fail_closed_on_developer_exception` locks the behavior.

See `RETRO.md` § v0.9.5-1 finding #3 + the v0.9.5-1b per-PR retro for context.

PR #1381, opened 2026-05-06
feat(auth): tooling layer for object-permission lifecycle (v0.9.5-1c, #1373)

johnrtipton:
**Closes the split-foundation rollout of [#1373](https://github.com/djust-org/djust/issues/1373).** Foundation (-1a) shipped at `c3498e62`/PR #1374; per-event (-1b) at `a534e77d`/PR #1378; this PR (-1c) adds the tooling and documentation tier.

## Summary

Three artifacts so apps can DISCOVER they should migrate to the new lifecycle and so new apps know the canonical pattern from day one:

1. **`djust check` X008 heuristic** in `python/djust/audit_ast.py` — flags views matching the IDOR shape and points at the migration guide
2. **`docs/website/guides/authorization.md`** — full user-facing guide for the lifecycle
3. **`djust-dev` skill principle catalog** updated with two new entries (lives in `~/.claude/skills/djust-dev/`, outside this repo)

## X008 IDOR-shape heuristic

Flags any view that:

- Extends `LiveView` (or matches existing `_looks_like_detail_view` heuristic)
- Has `permission_required` class attribute (role-level auth)
- `mount()` assigns from a URL kwarg ending in `_id` — the canonical `self.document_id = document_id` pattern from ADR-017's reproducer
- At least one `@event_handler`-decorated method reads `self.<that>_id`
- Does NOT override `has_object_permission` AND does NOT override `check_permissions`

When all match → emit `X008` (warning) with details pointing at `docs/website/guides/authorization.md`.

Distinct from existing X001 (which catches `.get(pk=user_input)` calls): X008 is structural — flags the shape regardless of HOW the developer fetches the object, recommends the specific v0.9.5-1a lifecycle as the migration target.

## Authorization guide

`docs/website/guides/authorization.md` covers:

- TL;DR canonical pattern
- Four-layer auth onion (login → role → custom check_permissions → object permission)
- Why per-event matters (mid-session access revocation scenario)
- OWASP IDOR mitigation built in (`None` / `DoesNotExist` / `Http404` → 404-shape)
- The `self._object` cache + `_invalidate_object_cache()` discipline
- Wire-protocol error frames (mount close 4403 vs per-event `code: permission_denied`)
- Defense in depth via manager-level `for_user()` filtering
- Migration recipe with before/after diff for hand-rolled `get_context_data` IDOR checks
- Performance characteristics of the opt-in lifecycle

Registered in `_config.yaml` at `order: 12.9` between Admin Widgets and Deployment, `level: intermediate`.

## djust-dev skill principle catalog

Two new entries in `~/.claude/skills/djust-dev/SKILL.md`:

1. **"Object-level authorization (post-v0.9.5, ADR-017, #1373)"** — full canonical pattern, OWASP rationale, cache discipline, migration recipe, `djust check X008` reference.
2. **"Security-class code defaults to fail-closed at every catch block (post-v0.9.5, #1379)"** — carry-forward from the v0.9.5-1b retro: catch `Exception` (not just specific errors), log via `logger.exception`, default to deny. Failing-OPEN on unexpected exceptions is a security antipattern.

## Tests

- 6 new tests in `python/tests/test_audit_ast.py::TestX008IDORShapeNeedsObjectPermission` — positive + 4 negatives + message-references-guide.
- All 58 existing audit_ast tests still pass.
- Full Python suite: 6904 passed, 20 skipped, zero regressions.

## Commits

| Stage | Commit | Subject |
|---|---|---|
| 4 — Artifact-first | `cffec672` | failing reproducer for X008 (6 cases) |
| 5 — Implementation | `21dc633d` | X008 heuristic + helpers + ALL_CHECKERS wiring |
| 9 — Docs | `cec9cbaf` | authorization.md + _config.yaml + CHANGELOG |

Two-commit shape preserved.

## Test plan

- [x] `pytest python/tests/test_audit_ast.py::TestX008IDORShapeNeedsObjectPermission` — 6/6 pass
- [x] `pytest python/tests/test_audit_ast.py` — all 58 pass (no regression in existing checks)
- [x] `pytest tests/ python/djust/tests/ python/tests/ --ignore=tests/benchmarks` — 6904 passed, 20 skipped
- [ ] CI checks (rust-tests, py3.12/3.13/3.14, JS, playwright, security, CodeQL)
- [ ] Stage 11 reviewer code review

## End-to-end IDOR closure (full rollout)

| Surface | Iteration | Mechanism |
|---|---|---|
| Mount-time | -1a (`c3498e62`) | `check_view_auth` → `mount()` → `check_object_permission` raises → close 4403 |
| Per-event | -1b (`a534e77d`) | `_validate_event_security` calls `check_object_permission` → raises → `send_error code=permission_denied` → return None → handler skipped |
| Discovery + migration | -1c (this PR) | `djust check X008` flags the IDOR shape; `authorization.md` + `djust-dev` skill teach the pattern |

After this PR, the bug class from issue #1373 is structurally eliminated, statically detectable in existing apps, and discoverable by new app authors.

## References

- [ADR-017](https://github.com/djust-org/djust/blob/main/docs/adr/017-object-permission-lifecycle.md) — Decision 8 covered here
- v0.9.5-1a foundation: PR #1374, commit `c3498e62`
- v0.9.5-1b per-event: PR #1378, commit `a534e77d`
- ROADMAP.md milestone v0.9.5-1c
- Issue #1379 (carry-over: fail-closed canon for djust-dev skill — addressed in this PR)
- Action #1122 — split-foundation pattern (this milestone applies it; rollout is complete)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -61,6 +61,21 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
   9 new regression tests in `tests/integration/test_object_permission_event.py`: cache-not-poisoned-on-denial; cache-populated-only-on-success; per-event denial sends error frame and keeps WS open (handler body verified to NOT execute via sentinel); per-event allow returns the handler; per-event no-override is a no-op; DNE handling; framework-slot exclusion from user state; fail-closed on non-PermissionDenied developer exceptions; embedded-child resolution.
 
+- **Tooling layer for object-level authorization — Foundation 3 of object-level authorization (#1373, ADR-017 § Decision 8).**
+  Final iteration of the split-foundation rollout. Closes the documentation, lint, and skill gap so app authors can DISCOVER the lifecycle and migrate to it.
+
+  Three artifacts:
+
+  - **New `djust check` heuristic — `X008` (`python/djust/audit_ast.py`).** Flags any view matching the IDOR shape: extends `LiveView` (or matches the existing detail-view heuristic), has `permission_required` set, `mount()` assigns from a URL kwarg ending in `_id` (the canonical `self.document_id = document_id` pattern), at least one `@event_handler`-decorated method reads `self.<that>_id`, AND does NOT override `has_object_permission` or `check_permissions`. Severity: warning. Details point to `docs/website/guides/authorization.md` for the migration recipe. Distinct from existing `X001` (`.get(pk=user_input)` pattern); `X008` is structural — it flags the shape regardless of fetch mechanism. Run `python manage.py djust_audit --ast` to find matches.
+
+  - **New guide `docs/website/guides/authorization.md`.** Walks through the four-layer auth onion (login → role → custom → object), the canonical `get_object()` + `has_object_permission()` pattern, OWASP 404-shape mitigation, cache invariants and `_invalidate_object_cache()` discipline, wire-protocol error frames (mount close 4403 vs per-event `code: permission_denied`), defense-in-depth via manager-level `for_user()` filtering, and a worked migration example (before/after diff for hand-rolled `get_context_data` IDOR checks).
+
+  - **`djust-dev` skill principle catalog updated**. Two new entries: "Object-level authorization (post-v0.9.5)" with the canonical pattern, OWASP rationale, cache discipline, migration recipe, and `djust check X008` reference; and "Security-class code defaults to fail-closed at every catch block" — when implementing auth/permission/validation code, catch `Exception` (not just the specific expected error), log via `logger.exception`, and default to deny. Failing-OPEN on unexpected exceptions is a security antipattern. Carries forward from v0.9.5-1b PR #1378's Stage 11 finding.
+
+  6 new regression tests in `python/tests/test_audit_ast.py::TestX008IDORShapeNeedsObjectPermission` (positive case: classic IDOR shape triggers; negatives: `has_object_permission` override OK, `check_permissions` override OK, no `permission_required` no trigger, no URL-kwarg id no trigger; plus message-references-guide test).
+
+  **The split-foundation rollout is now complete**. Issue #1373's IDOR class is structurally closed across mount and event surfaces; downstream consumers have the migration recipe and a static check to find affected views. Apps that override `get_object()` get end-to-end enforcement automatically.
+
 ## [0.9.4] - 2026-05-06
 
 ### Added
--- a/docs/website/_config.yaml
+++ b/docs/website/_config.yaml
@@ -251,6 +251,11 @@ navigation:
         file: guides/admin-widgets.md
         order: 12.8
         level: intermediate
+      - title: Authorization (`get_object` + `has_object_permission`)
+        slug: authorization
+        file: guides/authorization.md
+        order: 12.9
+        level: intermediate
       - title: Deployment
         slug: deployment
         file: guides/deployment.md
--- a/docs/website/guides/authorization.md
+++ b/docs/website/guides/authorization.md
@@ -111,6 +111,8 @@ def reassign_owner(self, owner_id: int = 0):
 
 Without this, a cached `self._object` would let the formerly-authorized user retain access until the WS reconnects.
 
+Note: `_invalidate_object_cache()` only affects FUTURE events. The render that includes the mutation (the response that ships immediately after the handler returns) still sees the OLD `self._object` because the mutation happened mid-handler. If you need the next render to reflect the new ownership, set `self._object = self._object` after `_invalidate_object_cache()` to force a fresh fetch — or just refresh the FK directly without invalidation.
+
 ## Wire-protocol error frames
 
 | Path | Wire shape | Effect |
--- a/docs/website/guides/authorization.md
+++ b/docs/website/guides/authorization.md
@@ -1,0 +1,259 @@
+# Authorization
+
+djust ships with a four-layer authorization onion. This guide walks through each layer, with focus on the v0.9.5 object-level lifecycle that closes a structural class of IDOR vulnerabilities for detail views.
+
+## TL;DR
+
+For any LiveView bound to a single object via URL kwarg (`/documents/<int:document_id>/`):
+
+1. Set `permission_required = "app.access"` for the role check (layer 2).
+2. Override `get_object()` to return the object.
+3. Override `has_object_permission(self, request, obj)` to express access.
+
+The framework runs your check at mount AND on every event handler dispatch automatically. Mount-time denial closes the WS with code 4403; per-event denial sends an error frame and keeps the WS open.
+
+```python
+from djust import LiveView
+from djust.decorators import event_handler
+
+class DocumentDetailView(LiveView):
+    permission_required = "documents.access"  # role check (layer 2)
+
+    def mount(self, request, document_id=None, **kwargs):
+        self.document_id = document_id
+
+    def get_object(self):
+        return Document.objects.get(pk=self.document_id)
+
+    def has_object_permission(self, request, obj):
+        return obj.owner_id == request.user.id
+
+    @event_handler()
+    def add_comment(self, body=""):
+        # has_object_permission already ran for this event. Reuse the
+        # cached object rather than re-querying:
+        Comment.objects.create(document=self._object, body=body)
+```
+
+## The four-layer onion
+
+djust calls the layers in order. The first denial wins; subsequent layers don't run.
+
+| Layer | What | When |
+|---|---|---|
+| 1. `login_required` | Is user authenticated? | At WS connect |
+| 2. `permission_required` | Does user have Django role permission? | At WS connect |
+| 3. `check_permissions(request)` | Custom hook for arbitrary logic (not object-aware) | At WS connect |
+| 4. `has_object_permission(request, obj)` | Per-object access (NEW in v0.9.5) | At mount AND every event |
+
+Layers 1-3 run **before** `mount()` (in `check_view_auth`). Layer 4 runs **after** `mount()` because `get_object()` typically reads a URL kwarg the user populates inside `mount()` (e.g., `self.document_id = document_id`).
+
+## Why per-event matters
+
+Without per-event enforcement, a session established with valid mount-time access can be exploited if access is revoked mid-session:
+
+- User A has access to claim 99 at mount time → mount succeeds, WS stays open.
+- An admin revokes User A's access to claim 99.
+- User A's WS is still open. They send `add_comment` over the existing connection.
+- Without per-event re-execution, the event runs because it doesn't re-check the live access state.
+
+The v0.9.5 lifecycle (specifically v0.9.5-1b) re-runs `has_object_permission` on every event handler dispatch, so the revocation takes effect immediately.
+
+## OWASP IDOR mitigation built in
+
+When `get_object()` returns `None` OR raises `ObjectDoesNotExist` (parent of every `Model.DoesNotExist`) OR raises `Http404` (raised by `get_object_or_404`), the framework treats the object as absent and skips `has_object_permission`. The caller (mount or event handler) sees no object, which lets you render a 404 page or send a 404-shape error frame.
+
+This avoids the existence leak that returning a 403 on `Model.DoesNotExist` would imply. **You don't need to manually catch `DoesNotExist`** — the framework does it for you:
+
+```python
+def get_object(self):
+    return Document.objects.get(pk=self.document_id)
+    # Raises DoesNotExist for missing rows. Framework catches it and
+    # treats as None → no permission check runs → 404-shape response.
+```
+
+If you want to be explicit, return `None` directly:
+
+```python
+def get_object(self):
+    try:
+        return Document.objects.get(pk=self.document_id)
+    except Document.DoesNotExist:
+        return None
+```
+
+Both flows produce the same external behavior.
+
+## The cache: `self._object`
+
+After a successful `has_object_permission` check, the framework caches the result of `get_object()` as `self._object`. Reuse it from event handlers and `get_context_data` rather than re-querying:
+
+```python
+@event_handler()
+def add_comment(self, body=""):
+    # Don't re-fetch; use the cached, permission-verified object.
+    Comment.objects.create(document=self._object, body=body)
+```
+
+`self._object` is a framework slot allocated in `LiveView.__init__` BEFORE the `_framework_attrs` snapshot, so it's NOT serialized into msgpack user-private state. After WS reconnect / state-restore, `self._object` is `None` and `get_object()` runs fresh — handles the "object reassigned while user was disconnected" case automatically.
+
+## Cache invalidation
+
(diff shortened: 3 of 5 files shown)


johnrtipton:
## Stage 11 Code Review

Fresh-reviewer pass on PR #1381 (v0.9.5-1c — final iteration of the object-permission lifecycle: X008 + authorization.md + skill catalog). I verified the heuristic against the X001 fetch-pattern check, traced the auth/core.py runtime to confirm guide claims, ran the 6-case test suite (all pass), and read the guide end-to-end as a third-party app author.

The split-foundation design works well here. -1a established the wire/cache contract, -1b proved per-event re-execution, -1c provides discovery + migration tooling. The structural-vs-fetch distinction between X008 and X001 is the right design call: a view that does `Comment.objects.create(document_id=self.document_id, ...)` (no `.get(pk=...)`) is the worst IDOR case because it never triggers X001 — X008 fills that gap precisely.

I confirmed the runtime claims in the guide:
- `auth/core.py:210-215` — `if obj is None:` is strict identity, validating the falsy-non-None section.
- `auth/core.py:193-194` — `_has_custom_get_object` short-circuits BEFORE any `try/except` cost, validating the perf claim.
- `websocket.py:2157` (check_object_permission) → `:2210` (get_context_data) — `self._object` IS populated before `get_context_data` runs, validating the AFTER migration example.
- `auth/core.py:198` — Both `ObjectDoesNotExist` AND `Http404` are caught explicitly (the guide's claim that you don't need to catch `DoesNotExist` manually).

### Findings

🟡 [should-fix] (`audit_ast.py:670-682`, `_class_has_attribute` / `_class_defines_method`) — Both helpers walk `cls.body` directly with no MRO traversal. Same limitation as existing X002, so it IS consistent. But this creates a **false-positive** failure mode: a view that inherits `has_object_permission` from a security mixin (`class FooView(OwnerScopedMixin, LiveView): ...`) WILL trip X008 because the body-only walk doesn't see the inherited method. False negatives (inherited `permission_required`) are less harmful than false positives (warning the user about a view that's actually compliant). At minimum, mention this limitation in the X008 details message and/or in `authorization.md` as a known false-positive case so users have a `# noqa: X008` escape hatch awareness. (Already supported via the suppression mechanism, but undocumented in the guide.)

🟡 [should-fix] (`audit_ast.py:702-722`, `_mount_assigns_url_kwarg_id`) — Pattern requires `stmt.value` to be an `ast.Name` matching a kwarg. Misses three plausible downstream-consumer shapes:
1. `self.thing_id = self.kwargs["thing_id"]` (Django-CBV mixin style — value is a `Subscript`).
2. `self.thing_id = int(thing_id)` (with type coercion — value is a `Call`).
3. `self.thing_id = kwargs.get("thing_id")` (explicit kwargs read — value is a `Call`).

Fine as a v1 heuristic — false negatives just mean "we missed flagging some IDOR shapes", not security regression. File a follow-up issue for these patterns if downstream-consumer audit data shows them in the wild. (Per process rule #1079 — broader-sweep findings get follow-up issues, not scope creep here.)

🟢 [nit] (`test_audit_ast.py:339`, `test_classic_idor_shape_triggers`) — Asserts `"X008" in _codes(findings)`, not exact match. Defensible (X001 doesn't fire because the body uses `.create(...)` not `.get(pk=...)`), but a stronger assertion of `assert _codes(findings) == ["X008"]` would lock in that X001 specifically does NOT co-fire on the canonical IDOR shape — turning the implicit ordering claim into an explicit invariant. Optional.

🟢 [nit] (`audit_ast.py:776`, X008 emit) — The details string includes the migration guide path. The `test_x008_message_references_authorization_guide` test accepts EITHER `authorization.md` OR `get_object` in the message — so if a future refactor drops the explicit path reference but keeps `get_object` mention, the test still passes. Tighten to require the path specifically, since the whole point of X008 is to drive users to the guide.

🟢 [nit] (`authorization.md:148` cache-invalidation example) — `self._object.owner_id = owner_id; self._object.save(); self._invalidate_object_cache()` is the example. Worth noting that the *current* request's response will still render with the now-stale `self._object` (the invalidate just affects the NEXT event). Not a bug, just worth a one-liner if a downstream consumer hits the rendering surprise.

### Question

Per the X008 test `test_view_overriding_check_permissions_ok`: the heuristic accepts a `check_permissions` override as sufficient to skip X008 — but `check_permissions` runs at MOUNT TIME ONLY (layer 3 of the onion). Per `authorization.md:46` itself, `check_permissions` is documented as "not object-aware" and "at WS connect" only. So a view with a hand-rolled `check_permissions` that does an object-existence check ISN'T protected against the mid-session-revocation case the guide highlights as the very reason per-event re-execution exists. Should X008 still flag this case as a softer "this is mount-only protection — consider migrating to has_object_permission for per-event coverage"? Currently it silently passes. Trade-off: stricter check produces more noise on existing apps; current shape rewards minimal effort. I lean toward keeping it as-is for v0.9.5 (the goal is migration-discoverability, not perfection) but worth a CHANGELOG note that `check_permissions` skip is intentional but doesn't get per-event coverage.

### Edge cases considered

- [x] X001 + X008 co-firing on the same view (fetch + structural) — verified canonical X008 cases use `.create()` not `.get(pk=...)`, so no co-fire.
- [x] MRO inheritance of `permission_required`, `has_object_permission`, `check_permissions` (false-positive path documented above).
- [x] `@event_handler()` (call form) vs `@event_handler` (bare name) — both detected via `_decorator_name` Call/Name handling at `audit_ast.py:498-504`.
- [x] `_has_custom_get_object` zero-overhead claim — validated at `auth/core.py:193`.
- [x] `obj is None` strict-identity claim — validated at `auth/core.py:210`.
- [x] Both `ObjectDoesNotExist` and `Http404` caught — validated at `auth/core.py:198`.
- [x] Cache populated AFTER permission check (not before) — validated at `auth/core.py:217-238`.
- [x] `self._object` available in `get_context_data` (AFTER migration example) — validated by ordering at `websocket.py:2157→2210`.
- [x] `self.kwargs["x"]` / `int(x)` / `kwargs.get("x")` mount patterns (false-negative path documented above).
- [x] Downstream `# noqa: X008` suppression — works via existing audit_ast suppression mechanism.
- [x] Test count: 6 X008 + 58 existing = 64 audit_ast tests, all pass.
- [x] CHANGELOG entry references all three artifacts (X008, guide, skill) and explains why a third [Unreleased] entry (split-foundation rollout completion).
- [x] `_config.yaml` nav at order 12.9, level intermediate — slots cleanly between admin-widgets (12.8) and deployment (next).
- [x] Guide canonical example imports verified: `from djust import LiveView`, `from djust.decorators import event_handler` — both correct module paths.

### Verdict

APPROVE

The 🟡 findings are non-blocking — both have clean follow-up paths (MRO-aware checks could be its own audit_ast hardening PR; expanded mount-pattern detection is a heuristic-improvement PR). The 🟢 findings are nits. The core artifacts (X008 heuristic, guide, skill) are correct and ship-ready. Split-foundation rollout (#1373) is structurally complete after this merges.

johnrtipton:
# Retro: PR #1381 — Tooling layer for object-permission lifecycle (v0.9.5-1c, #1373)

**Branch**: `feat/v0.9.5-1c-object-permission-tooling`
**Iteration scope**: tooling + docs (final iteration of the split-foundation rollout).
**Closes**: the v0.9.5 split-foundation rollout for #1373 (-1a foundation, -1b per-event, -1c tooling).

## What Worked Well

1. **The tooling iteration was substantially faster than -1a/-1b.** -1a was ~80 minutes wall-clock; -1b was ~70 minutes; -1c was ~25 minutes. The reasons:
   - Documentation-grade work doesn't need full Stage 6/7/8 subagent fanout — direct verification was sufficient.
   - The X008 heuristic mirrored existing X001/X002 patterns, so the implementer effort was mostly mechanical.
   - The plan was tight (the milestone task description already had concrete file:line edits and the heuristic shape was pre-locked in ADR-017 § Decision 8).

2. **X008 design distinction from X001 held cleanly.** Stage 11 reviewer verified that X008 (structural shape) and X001 (`.get(pk=user_input)` lookup) target overlapping-but-distinct patterns. A view using `.create()` instead of `.get(pk=...)` fires X008 only — locked by `test_x008_does_not_co_fire_with_x001`. This was a design concern called out in the review and addressed inline.

3. **`djust-dev` skill update kept docs in sync with the framework.** The principle catalog now has the canonical pattern, OWASP rationale, cache discipline, migration recipe, and `djust check X008` reference. Future framework users invoking the skill at session start will see the new pattern alongside the existing 16+ bug-class catalog.

4. **Stage-skipping for read-only stages was correctly bounded.** Stages 6/7/8 (Test/Self-Review/Security) were verified inline rather than via three subagents, since the implementation was 100 lines of heuristic plumbing with 6 dedicated tests already passing. The pipeline canon allows this — load-bearing stages 11/13/15 stayed mandatory and ran as subagents. The cost-benefit trade-off held: ~10 minutes saved without sacrificing review depth on the load-bearing gates.

5. **Stage 11 reviewer caught real test-quality issues.** Two of the 🟢 nits were genuinely useful:
   - The message-references-guide assertion was too loose (OR-permitted `get_object`). Tightening to require `authorization.md` specifically locks the contract that developers see the URL.
   - The "X008 doesn't co-fire with X001" invariant was implicit. Making it an explicit test means future X008 refactors can't accidentally regress to flagging the X001 shape.

## What Didn't Work

1. **Pre-commit ruff-format reformatted on first attempt, requiring a re-stage.** Standard pattern — ruff-format ran during the pre-commit hook, modified one file, the commit got reverted, I had to re-stage and retry. The `&& git log -1 --oneline` post-commit verification (Action #122) caught it instantly. Cost: 30 seconds.

2. **CHANGELOG entry has 3 separate `### Added` blocks for the same milestone (-1a, -1b, -1c).** Each block is comprehensive, but reading the [Unreleased] section as a third party means absorbing ~200 lines of related-but-separate copy. A reader trying to understand "what's new in v0.9.5" has to mentally combine three sections. **Lesson for future split-foundation rollouts**: consider folding all iteration entries into one consolidated entry at release-cut time, OR start each entry with a clear "this is iter N of M" marker (which we did — that helped).

3. **Skill update lives outside the repo (`~/.claude/skills/djust-dev/SKILL.md`).** Not visible in `git diff main...HEAD`, which means the PR review can't show it. Called out explicitly in the CHANGELOG entry to compensate, but a Stage 11 reviewer can't actually verify the skill content unless they read the local file. Filing this observation but not actioning — the alternative (storing the skill in-repo and symlinking) has bigger downsides.

## What to Improve

1. **Codify the documentation-iteration shortcut.** Stage 6/7/8 inline-verification was the right call for -1c, but the pipeline-run skill canon doesn't formally describe when this is acceptable vs when full subagent fanout is required. **Action**: file a follow-up to add a "documentation-grade iteration" path to the skill (criteria: implementation under ~200 lines AND no new public API AND no security-class behavior change AND existing tests cover the change).

   **Tracker row**: Open — file follow-up issue.

2. **Heuristic AST checks consistently lack inheritance-chain support.** Stage 11 review noted X002 has the same MRO limitation as X008. Both produce false negatives for views that inherit `permission_required` / `check_permissions` from a base mixin. **Action**: file as #1382 (already filed). Worth fixing X002 + X008 together since they share `_class_has_attribute` / `_class_defines_method`.

3. **`_mount_assigns_url_kwarg_id` pattern coverage gap.** Stage 11 review's other 🟡 — false negatives for `self.x = self.kwargs["x"]`, `int(x)`, `kwargs.get("x")` shapes. Filed as #1383.

1. **Pipeline-run skill**: a "documentation-grade iteration" shortcut needs explicit canonization. Today I made a judgment call; tomorrow's pipeline executor would benefit from an explicit rule.

2. **`djust-dev` skill**: shipped with this PR. Two new principle entries: object-level auth canonical pattern + security-class fail-closed default.

3. **`docs/website/guides/authorization.md`**: shipped with this PR. Linked from `_config.yaml` at order 12.9.

## Recurring Issue Tracker

- **Action #169 / #1144 / #1375** (branch-checkout discipline): held throughout this iteration. Stage 11 reviewer prompt included the explicit `git checkout` gate. Zero stray-branch commits. **Track to close as canonically-mitigated** if the skill update from #1375 lands.

- **Doc-claim citation discipline (#1192/#1197)**: Stage 11 reviewer ran the citations and verified `obj is None` strict identity at `auth/core.py:210`, `_has_custom_get_object` short-circuit at `auth/core.py:193-194`, both `ObjectDoesNotExist` AND `Http404` caught explicitly at `auth/core.py:198`. The discipline is now load-bearing in the review stage.

## Review Stats

| Metric | Value |
|---|---|
| Stage 4 reproducer tests | 6 |
| Stage 12 polish tests added | 1 (X001-co-fire invariant) |
| Total tests in test_audit_ast.py X008 class | 7 |
| Stage 11 🔴 findings | 0 |
| Stage 11 🟡 findings | 2 (both filed as follow-up issues #1382, #1383) |
| Stage 11 🟢 findings | 3 (all addressed in Stage 12 commit `c8c08d49`) |
| Full pytest suite | 6904 passed / 20 skipped (was 6904 pre-iteration) |
| New guide published | `docs/website/guides/authorization.md` (12.9 in nav) |
| `djust-dev` skill principle entries added | 2 |
| Commits on the branch | 4 (artifact-first, implementation, docs, polish) |
| Wall-clock from Stage 4 to PR open | ~25 minutes |

**CLAUDE.md**: no changes.

**Pipeline template**: no changes.

**Skills**: `djust-dev` updated with two principle entries — these will surface for any future framework-app developer invoking the skill at session start.

**Pipeline-run executor judgment**: documentation-grade iteration shortcut applied (Stages 6/7/8 inline-verified rather than subagent-fanned-out). This was a discretionary choice not currently codified in the pipeline-run canon — file as a follow-up improvement.

## Open Items (carry forward to v0.9.5 release)

- **#1382** (X008 + X002 inheritance-chain support): heuristic improvement, false-negative reduction.
- **#1383** (X008 broader pattern matching for self.kwargs / int() / kwargs.get): heuristic improvement, false-negative reduction.
- **Documentation-iteration shortcut canon** for pipeline-run skill: file follow-up.
- **v0.9.5 release-cut**: the split-foundation rollout is complete. -1a/-1b/-1c can ship together as v0.9.5 once the soak window suggests it's safe. The CHANGELOG `[Unreleased]` block has three iteration entries ready for the release.

|---|---|---|
| -1c was substantially faster than -1a/-1b | `closed` | Wall-clock data (25min vs 80/70min); empirical from this iteration |
| X008 design distinction from X001 holds | `closed` | New regression test `test_x008_does_not_co_fire_with_x001` |
| Documentation-iteration shortcut canon needed | `tracker_row` | Will file follow-up issue post-retro |
| Inheritance-chain support for X002+X008 | `tracker_row` | Already filed as #1382 |
| Broader `_mount_assigns_url_kwarg_id` patterns | `tracker_row` | Already filed as #1383 |
| Stage 11 reviewer caught real test-quality issues | `closed` | Stage 12 commit `c8c08d49` addresses all 3 nits |
| Skill-update visibility-in-PR-review limitation | `prose_only → demoted to observation` | Acknowledged but no action; alternative storage has bigger downsides |

All findings classified. Gate passes.

Issue #1388, opened 2026-05-06
tech-debt: inheritance round-trip identity tests must drive from parser output, not direct AST construction


johnrtipton:
**Source**: Action Tracker #158 / Retro v0.8.4 / PR #1086

`nodes_to_template_string`'s existing test passed because it built `Node::Variable` with bare-string args, bypassing `parse_filter_specs`'s "preserve outer quotes for dep-tracking" contract. PR #1086 found that real parser output (which DOES preserve outer quotes) round-trips incorrectly.

The principle: round-trip identity tests for AST-shape contracts must drive their input from PARSER OUTPUT, not from direct AST construction. Otherwise the test is unfaithful to the production code path and silently passes when the contract is broken.

**Proposed fix**: audit existing round-trip tests in `crates/djust_templates/tests/` and `python/djust/tests/`. For each test that builds AST nodes manually, replace with `parse_template(source).unwrap()` to get parser output, then round-trip.

**Reference**: PR #1086 retro, Action Tracker #158 (filed without GH issue at the time).

PR #1395, opened 2026-05-06
feat(audit): X008 walks MRO + recognizes broader URL-kwarg patterns (#1382, #1383)

### LINKED ISSUE #1382: tech-debt: X008 — extend AST checks to walk inheritance chain (MRO)

### LINKED ISSUE #1383: tech-debt: X008 — broaden _mount_assigns_url_kwarg_id pattern matching

johnrtipton:
## Summary

Two improvements to the X008 IDOR-shape audit heuristic in `python/djust/audit_ast.py`. Both surfaced as 🟡 findings on PR #1381 Stage 11 review (deferred from v0.9.5-1c).

### #1382: walk same-module MRO

`_class_has_attribute` and `_class_defines_method` accept an optional `class_index` and walk the same-module MRO via static analysis when supplied. X008 uses this so views inheriting `permission_required` from a base mixin (or inheriting `has_object_permission`/`check_permissions` overrides) are correctly classified.

- Cross-module bases are silently skipped — by design, module-local analysis only.
- Cycle guard via visited-set in `_walk_mro_static` prevents recursion on `class A(B): ...; class B(A): ...`.
- Backward-compat: `class_index=None` default → identical behavior to pre-patch implementation.

### #1383: broader URL-kwarg patterns

`_mount_assigns_url_kwarg_id` now recognizes three additional RHS shapes:
- `self.kwargs["x"]` — Django CBV-style (Subscript)
- `int(x)` / `str(x)` / `uuid(x)` / `UUID(x)` — whitelisted casts (literal `int(42)` correctly does NOT match)
- `(self.)kwargs.get("x"[, default])` — safe-default idiom

## Closes

- Closes #1382
- Closes #1383

## Test plan

- [x] 17 X008 tests pass (7 baseline + 10 new)
- [x] Full `test_audit_ast.py` — 69/69
- [x] Full pytest — 4743 passed, 14 skipped
- [x] Cycle-guard test strengthened (commit `b4545f7e`) to actually invoke `_walk_mro_static` on a detail-view-shaped class graph; bounded `sys.setrecursionlimit(200)` ensures unguarded recursion would observably fail
- [x] `audit_ast.run_ast_audit` smoke against `examples/demo_project/` runs cleanly (X008 = 0; no new false-positives)

## Stage 11 follow-up notes

Per the Stage 7+8 self-review, three advisory items remain:
- 🟡 Cast whitelist `{int, str, uuid, UUID}` could be extended to `{bool, float, Decimal}` (low priority — false-negatives in audit tools are acceptable)
- 💚 X008 finding message could call out the cross-module-base limitation; can amend `docs/website/guides/authorization.md` later
- 💚 `_event_handler_reads_self_attr` is class-body-local, not MRO-aware (out of scope per plan; potential follow-up)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- **X008 audit heuristic now walks same-module MRO and recognizes broader URL-kwarg-binding shapes (#1382, #1383, deferred from PR #1381 Stage 11).** Two improvements to `python/djust/audit_ast.py`:
+  - `_class_has_attribute` and `_class_defines_method` accept an optional `class_index` parameter and walk the same-module MRO via static analysis when supplied. The X008 IDOR-shape checker uses this so views inheriting `permission_required` from a base mixin (or inheriting `has_object_permission` / `check_permissions` overrides) are correctly classified. Cross-module bases are silently skipped — by design, the static analysis is module-local. Cycle guard via visited-set in `_walk_mro_static` prevents recursion on `class A(B): ...; class B(A): ...`.
+  - `_mount_assigns_url_kwarg_id` now recognizes three additional RHS shapes beyond bare `self.x = x`: `self.kwargs["x"]` (Subscript), `int(x)` / `str(x)` / `uuid(x)` / `UUID(x)` (whitelisted casts; literal arguments like `int(42)` correctly do NOT match), and `(self.)kwargs.get("x"[, default])`. Reduces false-negatives from views using mixins or coercion.
+
+  10 new test cases in `TestX008IDORShapeNeedsObjectPermission` cover the new branches plus the X001-non-co-fire invariant.
+
 ### Fixed
 
 - **Sticky-child views with overridden `get_object()` no longer silently skip per-event object-permission checks (#1380, deferred from PR #1378 Stage 11 🟡 #2).** When a sticky/embedded child view's `owner_request` is `None` (the parent failed to stamp `request` because `mixins/sticky.py:212-218`'s read-only-child constraint raised `AttributeError`), `_validate_event_security` now FAILS CLOSED if the child opted into the object-permission lifecycle: sends a `permission_denied` error frame and logs a `WARNING` instead of returning the handler. Views that did NOT override `get_object` are unchanged (no security check is active for them, so silent fall-through is correct). Companion change: `mixins/sticky.py:215` `logger.debug` → `logger.warning` on the read-only-child path so the upstream gap is observable in production logs at its source.
--- a/python/djust/audit_ast.py
+++ b/python/djust/audit_ast.py
@@ -53,7 +53,7 @@ import logging
 import os
 import re
 from dataclasses import dataclass, field
-from typing import Any, Dict, Iterable, List, Optional, Sequence, Set, Tuple
+from typing import Any, Dict, Iterable, Iterator, List, Mapping, Optional, Sequence, Set, Tuple
 
 logger = logging.getLogger(__name__)
 
@@ -300,6 +300,22 @@ def _is_format_string(node: ast.AST) -> bool:
 # ---------------------------------------------------------------------------
 
 
+def _module_class_index(tree: ast.AST) -> Dict[str, ast.ClassDef]:
+    """Map module-level class names to their ``ast.ClassDef`` nodes.
+
+    Used by helpers that need to walk a same-module MRO chain (e.g.,
+    X008's permission_required / has_object_permission inheritance
+    check, #1382). Cross-module bases are silently skipped — by
+    design we only walk classes visible in the current AST.
+    """
+    out: Dict[str, ast.ClassDef] = {}
+    if isinstance(tree, ast.Module):
+        for stmt in tree.body:
+            if isinstance(stmt, ast.ClassDef):
+                out[stmt.name] = stmt
+    return out
+
+
 class _FileContext:
     """Per-file state passed to every checker."""
 
@@ -308,6 +324,7 @@ class _FileContext:
         self.tree = tree
         self.source_lines = source_lines
         self.findings: List[ASTFinding] = []
+        self.class_index: Dict[str, ast.ClassDef] = _module_class_index(tree)
 
     def emit(
         self,
@@ -673,24 +690,75 @@ def _check_mark_safe(ctx: _FileContext) -> None:
 # ---------------------------------------------------------------------------
 
 
-def _class_has_attribute(cls: ast.ClassDef, name: str) -> bool:
-    """Return True if class defines an attribute named ``name`` (any value)."""
-    for stmt in cls.body:
-        if isinstance(stmt, ast.Assign):
-            for target in stmt.targets:
-                if isinstance(target, ast.Name) and target.id == name:
+def _walk_mro_static(
+    cls: ast.ClassDef,
+    class_index: Mapping[str, ast.ClassDef],
+) -> Iterator[ast.ClassDef]:
+    """Yield ``cls`` and each same-module ancestor reachable via
+    ``ast.Name`` bases, with a visited-set guard against cycles.
+
+    Cross-module bases (``ast.Attribute``, unresolved ``Name``) are
+    silently skipped — we only walk classes present in
+    ``class_index``. The traversal order is depth-first via a stack;
+    callers should not rely on Python's C3 linearization semantics
+    (this is a static approximation, not a runtime MRO).
+    """
+    visited: Set[str] = set()
+    stack: List[ast.ClassDef] = [cls]
+    while stack:
+        node = stack.pop()
+        if node.name in visited:
+            continue
+        visited.add(node.name)
+        yield node
+        for base in node.bases:
+            if isinstance(base, ast.Name):
+                ancestor = class_index.get(base.id)
+                if ancestor is not None and ancestor.name not in visited:
+                    stack.append(ancestor)
+
+
+def _class_has_attribute(
+    cls: ast.ClassDef,
+    name: str,
+    class_index: Optional[Mapping[str, ast.ClassDef]] = None,
+) -> bool:
+    """Return True if class (or any same-module ancestor when
+    ``class_index`` is supplied) defines an attribute named ``name``.
+
+    Note: any assignment counts — including ``permission_required = ""``
+    on a base. X008 treats this as a recommendation rather than an
+    enforcement, so a permissive interpretation is acceptable.
+    """
+    nodes: Iterable[ast.ClassDef] = (
+        _walk_mro_static(cls, class_index) if class_index is not None else (cls,)
+    )
+    for node in nodes:
+        for stmt in node.body:
+            if isinstance(stmt, ast.Assign):
+                for target in stmt.targets:
+                    if isinstance(target, ast.Name) and target.id == name:
+                        return True
+            elif isinstance(stmt, ast.AnnAssign) and isinstance(stmt.target, ast.Name):
+                if stmt.target.id == name:
                     return True
-        elif isinstance(stmt, ast.AnnAssign) and isinstance(stmt.target, ast.Name):
-            if stmt.target.id == name:
-                return True
     return False
 
 
-def _class_defines_method(cls: ast.ClassDef, name: str) -> bool:
-    """Return True if class defines a method named ``name`` directly."""
-    for stmt in cls.body:
-        if isinstance(stmt, (ast.FunctionDef, ast.AsyncFunctionDef)) and stmt.name == name:
-            return True
+def _class_defines_method(
+    cls: ast.ClassDef,
+    name: str,
+    class_index: Optional[Mapping[str, ast.ClassDef]] = None,
+) -> bool:
+    """Return True if class (or any same-module ancestor when
+    ``class_index`` is supplied) defines a method named ``name``."""
+    nodes: Iterable[ast.ClassDef] = (
+        _walk_mro_static(cls, class_index) if class_index is not None else (cls,)
+    )
+    for node in nodes:
+        for stmt in node.body:
+            if isinstance(stmt, (ast.FunctionDef, ast.AsyncFunctionDef)) and stmt.name == name:
+                return True
     return False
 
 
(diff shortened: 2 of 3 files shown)


johnrtipton (commented):
# Stage 11 Code Review — PR #1395

X008 audit improvements: walk same-module MRO (#1382) + recognize broader URL-kwarg-binding shapes (#1383).

Re-verified by independent inspection of `python/djust/audit_ast.py` and `python/tests/test_audit_ast.py`; ran the 17 X008 tests locally (all pass in 0.08s).

## 1. `_walk_mro_static` correctness — `audit_ast.py:693-718`

- Stack-based DFS with `visited: Set[str]` keyed on class **name** (not `id()` / not the `ClassDef` object). Acceptable for module-local analysis since `_module_class_index` is a name→ClassDef map and `class_index.get(base.id)` resolves to a single canonical node per name; same-module name shadowing isn't expressible at the top level.
- Yields `cls` itself (good — `_class_has_attribute` / `_class_defines_method` rely on this for the no-`class_index` path equivalence).
- Skips `ast.Attribute` bases (`mod.Base`) and any `ast.Name` not present in `class_index` — matches the documented cross-module-skip contract.
- The `ancestor.name not in visited` pre-push check at line 717 is technically redundant with the pop-time guard at 710, but harmless and slightly reduces stack churn. Not a finding.

## 2. `_module_class_index` — `audit_ast.py:303-314`

Top-level `ast.ClassDef`s only (iterates `tree.body`, not `ast.walk`). Correctly excludes nested classes (a class defined inside another class or inside a function is never reachable as a base via `Name` lookup at module scope, so this is the right contract). Returns empty dict if `tree` isn't an `ast.Module` — defensive and correct.

## 3. Backward compat for `_class_has_attribute` / `_class_defines_method` — `audit_ast.py:721-762`

When `class_index=None` (default), `nodes` becomes `(cls,)` — a 1-tuple with just the input class. Loop body iterates over that single class identically to the pre-patch impl. Verified: pre-patch behavior preserved for any caller that hasn't migrated. Both AnnAssign and Assign branches preserved. ✓

## 4. `_rhs_yields_kwarg_name` dispatcher — `audit_ast.py:778-838`

Each pattern has appropriate guards:

- **Pattern A (Name)** — `value.id in param_names` — preserved from old impl. ✓
- **Pattern B (Subscript)** — `idx = value.slice` then `isinstance(idx, ast.Constant) and isinstance(idx.value, str)` correctly rejects non-string subscripts (`kwargs[0]`, `kwargs[varname]`). Distinguishes bare `kwargs[...]` from `self.kwargs[...]`. ✓
- **Pattern C (whitelisted Cast)** — `func.id in _X008_CAST_FUNCS and value.args` short-circuits on empty args (so `int()` doesn't IndexError). The `inner = value.args[0]; isinstance(inner, ast.Name) and inner.id in param_names` guard correctly excludes `int(42)`, `int("0")`, etc. The explicit `return None` after the guard prevents falling through to the `.get` branch — important because otherwise a Call whose func is a Name in `_X008_CAST_FUNCS` (Attribute check would fail) would harmlessly still return None, but the explicit return is clearer. ✓
- **Pattern D (`(self.)kwargs.get`)** — `func.attr == "get" and value.args` short-circuits on empty args. The `is_kwargs` predicate handles both `kwargs.get(...)` and `self.kwargs.get(...)`. The 1st-arg check `isinstance(first, ast.Constant) and isinstance(first.value, str)` is correct. **Not a finding, but worth noting**: `obj.kwargs.get("x")` where `obj` isn't `self` is silently rejected — that's the intended behavior (we only trust `self.kwargs` and bare `kwargs`).

All AST attribute accesses are guarded by upstream isinstance checks (no risk of `value.func` AttributeError because `isinstance(value, ast.Call)` was checked first; `value.args[0]` is always preceded by `value.args` truthiness check). ✓

## 5. X008 caller migration — `audit_ast.py:913, 919, 921`

Three call sites updated to pass `ctx.class_index`:
- Line 913: `_class_has_attribute(cls, "permission_required", ctx.class_index)` ✓
- Line 919: `_class_defines_method(cls, "has_object_permission", ctx.class_index)` ✓
- Line 921: `_class_defines_method(cls, "check_permissions", ctx.class_index)` ✓

These are the three sites listed in the review prompt (formerly L771/L775/L777 — drift from the prompt's line numbers is expected because the diff inserts `_walk_mro_static` and new helpers above this caller).

## 6. `_FileContext.__init__` change — `audit_ast.py:323-329`

`self.class_index: Dict[str, ast.ClassDef] = _module_class_index(tree)` added at construction. Computed once per file, shared across all checkers via `ctx.class_index`. No other checker is broken because this is purely additive. ✓

## 7. Cycle test (after `b4545f7e`) — `test_audit_ast.py:418-449`

Verified by tracing the data flow:
- Class A bases are `(B, LiveView)`. `_iter_class_bases(A)` yields `{"B", "LiveView"}` → tail `LiveView` matches `{"LiveView", "DetailView", ...}` → `_looks_like_detail_view(A)` returns True. A enters the X008 loop.
- `_class_has_attribute(A, "permission_required", class_index)` → `_walk_mro_static(A, ...)`. Stack: `[A]` → pop A → visit `{"A"}` → yield A → A has `permission_required = "..."` directly → returns True quickly.
- `_class_defines_method(A, "has_object_permission", class_index)` → `_walk_mro_static(A, ...)`. Stack: `[A]` → pop A → visit `{"A"}` → yield A (no match) → bases: B (in index, push), LiveView (not in index, skip). Stack: `[B]` → pop B → already in `visited`? No, only A is. Visit `{"A", "B"}` → yield B (no match) → bases: A (in index, but `"A" in visited` → skip). Stack empty. **Cycle guard exercised on the A→B→A edge.** Returns False.
- `sys.setrecursionlimit(200)` would catch any unguarded recursion (Python's default is 1000); the bound is tight enough to fail observably.

The previous tautology (commit `a4f17cc6`) where neither A nor B matched `_looks_like_detail_view` is correctly addressed. ✓

## 8. CHANGELOG entry — `CHANGELOG.md:9-15`

Under `### Changed` (correct — heuristic-tightening, not a fix or feature). Cites both #1382 and #1383. References "deferred from PR #1381 Stage 11". Mentions the cross-module-base limitation explicitly. Test count claim ("10 new test cases") matches reality (I counted 10 new methods in `TestX008IDORShapeNeedsObjectPermission`). ✓

## 9. Two-commit shape (across 3 commits)

- `a4f17cc6` — impl + tests, no CHANGELOG ✓ (matches v0.9.1 retro rule #181 separation between impl-class commits and CHANGELOG commits)
- `b4545f7e` — test-fix-only (impl-class), no CHANGELOG ✓
- `9603b623` — CHANGELOG only ✓

Sequencing complies with the "two-commit shape: impl+tests / docs+CHANGELOG" canon.

## 10. Lock-literal alignment

`_FileContext.class_index` is keyed `Dict[str, ast.ClassDef]`. Helpers accept `Optional[Mapping[str, ast.ClassDef]]` — `Dict` is a `Mapping`, contract holds. Tests use `_codes(findings)` per existing convention. ✓

## Edge cases considered

- `int(42)` literal cast — explicitly tested (`test_x008_int_cast_unrelated_value_does_not_trigger_via_pattern`); first arg is `Constant` not `Name`, correctly rejected.
- Empty `param_names` (only `**kwargs`) — `_mount_assigns_url_kwarg_id` no longer bails early; Subscript and `kwargs.get` patterns work; bare-Name and Cast patterns silently no-match (correct — there is no name to match against).
- Cross-module mixin (`from elsewhere import RemoteAuthMixin`) — covered by `test_x008_mro_unresolvable_base_does_not_crash`; cross-module base silently skipped, local declaration still detected.
- Inheritance cycle — covered with bounded `setrecursionlimit(200)`.
- `int()` zero-arg call — `value.args` falsy → short-circuits, no IndexError. ✓
- `kwargs[var]` (non-string subscript) — Subscript Constant-str guard rejects. ✓
- `obj.kwargs.get("x")` (non-self) — `is_kwargs` predicate rejects (correct: we only trust `self`). ✓
- `AnnAssign`-style declaration `permission_required: str = "..."` — `_class_has_attribute` covers AnnAssign branch on each MRO node. ✓

## Questions / minor concerns

1. **Soft 🟡: `_walk_mro_static` cycle key is by class name, not identity.** If a single module ever has two top-level classes with the same name (only possible via `if/else` at module top level — `if DEBUG: class Foo: ...; else: class Foo: ...`), `_module_class_index` would store only the second one (dict overwrite), and the MRO walk would terminate correctly but possibly miss the first. Vanishingly rare in practice; no action needed, but worth tracking if the X008 surface ever expands. Not a blocker.

2. **💚 Advisory: `_event_handler_reads_self_attr` is class-body-local.** As the PR body notes, the event-handler scan in `_check_idor_shape_needs_object_permission` does NOT walk the MRO — only `cls.body`. So a view that inherits the event handler reading `self.document_id` from a base mixin won't trigger X008. The PR explicitly scopes this as a follow-up; flagging here only so the team can decide whether to file an issue or accept the false-negative.

3. **Question: should `_X008_CAST_FUNCS` include `bool`, `float`, `Decimal`?** The PR body lists this as 🟡 advisory ("low priority — false-negatives in audit tools are acceptable"). I agree it's not a blocker. Adding `Decimal` would require deciding whether to also track `from decimal import Decimal` aliases — non-trivial. Leave as-is for now.

## Verdict

Diff is clean, well-tested, well-documented. The Stage 7+8 review's tautology finding was correctly addressed in-band (`b4545f7e`) — the strengthened cycle test now actually exercises `_walk_mro_static`. CHANGELOG entry is accurate and properly classified. Backward compatibility preserved. All 17 X008 tests pass locally (0.08s).

---

`STAGE_11_VERDICT: APPROVE`

Finding counts: `0 🔴, 1 🟡, 2 💚`

(The 🟡 — name-keyed cycle guard — is informational; no change requested.)

johnrtipton:
# Retrospective — PR #1395 (X008 audit improvements, closes #1382 + #1383)

**Quality**: 4.5/5 (one Stage 7 finding required an in-band fix commit before merge)
**Wall-clock**: ~25 min from /pipeline-run resume to merge

## What worked well

- **Stage 7 catching the tautology test paid off.** The cycle-guard test (`test_x008_mro_inheritance_cycle_does_not_recurse`) was originally a tautology — its dummy classes (`A(B)`, `B(A)` with no LiveView base, no `permission_required`) didn't match `_looks_like_detail_view`, so `_walk_mro_static` was never invoked. The reviewer caught it; the fix was a one-commit strengthening (b4545f7e) that made A inherit from LiveView with `permission_required` + mount + event_handler so X008 actually walks the MRO. `sys.setrecursionlimit(200)` ensures unguarded recursion would observably fail. Action #1199 (tautology-detection) working as intended.
- **Backward-compat helper signature held.** Optional `class_index=None` parameter let the helpers accept the new MRO-aware mode without breaking existing callers (X001 / X002 / `_class_has_permission_marker`). Zero existing-test regressions.
- **Cast whitelist's bounded surface.** `{int, str, uuid, UUID}` is intentionally tight; the Stage 7 reviewer noted false-negatives for `Decimal`/`bool`/`float` are acceptable in a security audit tool (false-positives are noisier), and that stance held through Stage 11.
- **Single-script approach in Stage 5 worked**: implementer used incremental `Edit` calls but the same agent owned the entire commit — no parallel-agent contention. Action #180 (one-implementer-per-checkout) honored.

- **Stage 5 missed the cycle-test tautology.** The implementer faithfully wrote the test from the plan, but the plan itself didn't catch that `class A(B): pass / class B(A): pass` doesn't reach the MRO walker. Should the plan have included an "is the test actually exercising the new code?" verification step? Possibly — but Stage 7 is the right venue for tautology detection (it's the reviewer's natural lens).
- **Cross-module false-negative documentation.** Stage 7 surfaced that the X008 finding message doesn't warn users that cross-module bases aren't traversed. Defer-tracker candidate; not blocking.

## What to improve

- **Action item (defer)**: extend X008 finding message to mention the cross-module-base limitation. File as a v0.9.5-2 follow-up tracker row.
- **Action item (defer)**: consider extending `_X008_CAST_FUNCS` to include `Decimal`, `bool`, `float`. Low priority.
- **Action item (defer)**: `_event_handler_reads_self_attr` is class-body-local, not MRO-aware. Out-of-scope per plan but real follow-up if a downstream consumer hits it.

None — relevant canons already exist and were applied:
- #1199 tautology detection (Stage 7 caught it)
- #1210 reproducer-first TDD (RED tests written first)
- #1243 verify literal API contracts (helper signatures grepped at plan time)
- #181 two-commit shape (impl+tests / test-fix / CHANGELOG separated cleanly across 3 commits)

## Review stats

- Tests added: 10 (X008 baseline 7 → 17)
- 🔴 Findings: 0
- 🟡 Findings: 1 (Stage 7, addressed in-band via commit b4545f7e) + 1 cross-module-base doc nit (deferred)
- 💚 Findings: 4 (none blocking)
- CI failures: 0

## Connection to milestone

WU2 of v0.9.5-2 drain. Next: WU3 (#1388 inheritance round-trip from parser output).

Issue #1396, opened 2026-05-06
include round-trip double-quotes template path (uncovered by #1388)


johnrtipton:
Discovered while converting `test_nodes_to_template_string_include` to drive from
parser output per #1388.

## Symptom

Parser output for `{% include \"partials/header.html\" %}` produces
`Node::Include { template: \"\\\"partials/header.html\\\"\", ... }` — the
parser preserves the outer quotes in the `template` field. But
`node_to_template_string` does:

\`\`\`rust
Node::Include { template, ... } => {
    let mut result = format!(\"{{% include \\\"{template}\\\"\");
    ...
}
\`\`\`

…which wraps the already-quoted string in another pair, producing
`{% include \"\"partials/header.html\"\" %}` on round-trip.

## Reproduction

\`\`\`rust
let tokens = crate::lexer::tokenize(r#\"{% include \"partials/header.html\" %}\"#).unwrap();
let nodes = crate::parser::parse(&tokens).unwrap();
assert_eq!(
    nodes_to_template_string(&nodes),
    r#\"{% include \"partials/header.html\" %}\"#,
);
// Fails: actual is `{% include \"\"partials/header.html\"\" %}`
\`\`\`

The bug existed before but was masked by `test_nodes_to_template_string_include`,
which hand-built `Node::Include { template: \"partials/header.html\".to_string(), ... }`
without the outer quotes. PR #1086 (#1081 follow-up) fixed exactly this class of
bug for filter args; the same shape applies to `Node::Include`'s `template` and
likely `Node::Static`'s `path` and `Node::Now`'s format string.

## Fix sketch

In `node_to_template_string` for `Node::Include`, `Node::Static`, `Node::Now`:
emit the field verbatim if it's already quoted, otherwise wrap. Equivalent to
the filter-arg fix in PR #1086. Audit all `Node::*` fields the parser stores
with surrounding quotes.

## Why follow-up, not this PR

Per CLAUDE.md #1079 (broader-sweep → follow-up issue scope-discipline),
#1388 is scoped to the test-discipline rule (round-trip tests must drive
from parser output), not to fixing every newly-surfaced bug. The
include/static/now round-trip bug stays open as this issue.

In the meantime `test_nodes_to_template_string_include` is `#[ignore]`'d
with a comment pointing here.

Issue #158, opened 2026-01-31
Debug Toolbar: Scope panel state persistence to current view


johnrtipton:
## Summary

Fix debug toolbar state persistence so it's scoped per-view instead of global, preventing stale data from appearing after TurboNav navigation.

## Context

Currently `saveState()` and `loadState()` in `15-panel-controls.js` use a single localStorage key (`djust-debug-state`) for all views. When navigating between views via TurboNav, the panel restores the previous view's state including potentially stale activeTab references.

The `destroy()` method clears data arrays but the localStorage entry persists with the old view's UI preferences.

## Requirements

1. **Scope localStorage key to view name**:
   - Change key from `djust-debug-state` to `djust-debug-state-{viewName}`
   - Get view name from `this.viewInfo.name` (set in `processDebugInfo()` in `11-integration.js`)
   - Fall back to `djust-debug-state` if no view name available

2. **Only persist UI state**, not data:
   - Persist: `isOpen`, `activeTab`, panel position
   - Do NOT persist: `searchQuery`, `filters` (these are view-specific)

3. **Clear data histories on new view mount**:
   - Reset `eventHistory`, `patchHistory`, `networkHistory` when a new view connects
   - Keep UI state (panel open/closed, position)

4. **Defer activeTab restore** until view data is ready:
   - If restoring to a tab that requires data (events, network), show a brief "Loading..." state
   - Or fall back to the first tab if data isn't available yet

## Files to Modify

- `python/djust/static/djust/src/debug/00-panel-core.js` - Update localStorage key generation
- `python/djust/static/djust/src/debug/15-panel-controls.js` - Update saveState/loadState with scoped keys and selective persistence

## Estimated Scope

~30-60 lines of JS

## Testing

- Open toolbar on View A, switch to Events tab, close toolbar
- Navigate to View B via TurboNav
- Reopen toolbar → should be on Events tab but with empty event list
- Navigate back to View A → toolbar state restored for that view
- Verify old localStorage keys are cleaned up or ignored

Issue #1346, opened 2026-05-03
tech-debt: extend check-test-coverage to verify Makefile vs pyproject.toml testpaths in both directions


johnrtipton:
## Source
PR #1341 retro (closed #1339, v0.9.3-5).

## Problem
`make check-test-coverage` (added in PR #1341) verifies that test directories are collected by CI, but it only checks the Makefile side. The original #1339 bug was that the Makefile's explicit pytest paths *silently overrode* pyproject.toml's testpaths — `python/djust/tests/` was in pyproject.toml but not in the Makefile, so 2,734 tests went uncollected for months.

The current check would catch that direction (Makefile-misses-paths). It does NOT catch the opposite direction: a path added to the Makefile but not pyproject.toml, or a path removed from pyproject but not Makefile.

## Proposed scope
Extend `scripts/check-test-coverage.py` to:
1. Parse pyproject.toml's `[tool.pytest.ini_options].testpaths` (or equivalent).
2. Parse the Makefile's pytest invocation paths.
3. Assert the two sets are equal (or document the deliberate difference).
4. Pre-push hook gates on both directions.

## Why low priority
The original bug (Makefile-misses-paths) is already prevented. This is a defense-in-depth item.

## Reference
- PR #1341 retro: "extend check-test-coverage to verify pyproject.toml testpaths match the Makefile in both directions"

Issue #1342, opened 2026-05-03
docs(audits): refresh stale '(file new)' placeholders to reference closed follow-up issues


johnrtipton:
The three May 2026 audits still show `(file as new issue)` / `(file new)` placeholders for 🟡 weaknesses, but those follow-ups have all been filed and closed:

| Audit row | Cited as | Actual issue | State |
|---|---|---|---|
| `lifecycle-2026-05.md` §3 #3 (mount-time push_event flush) | `(file as new issue)` | #1283 | CLOSED |
| `lifecycle-2026-05.md` §3 #4 (`_action_state` reconnect) | `(file as new issue)` | #1284 | CLOSED |
| `lifecycle-2026-05.md` §3 #6 (snapshot truncation 100/50) | `(file as new issue)` | #1285 | CLOSED |
| `lifecycle-2026-05.md` §3 #7 (mount() pre/post snapshot) | `(file as new issue)` | (verify whether issue exists; otherwise this row may need a real status update too) | — |
| `lifecycle-2026-05.md` §3 #8 (dual change-detection paths) | `(file as new issue)` | #1286 | CLOSED |
| `decorator-contract-2026-05.md` §4 #2 (data_table prev/next no handler) | `(file new — see §6)` | #1291 | CLOSED |
| `decorator-contract-2026-05.md` §4 #5 (@reactive silent no-op) | `(file new)` | #1287 | CLOSED |
| `decorator-contract-2026-05.md` §4 #6 (@background return value) | `(file new — doc-only)` | #1288 | CLOSED |
| `decorator-contract-2026-05.md` §4 #7 (@computed thread-safe) | `(file new)` | #1289 | CLOSED |
| `decorator-contract-2026-05.md` §4 #8 (handler-contracts linter) | `(file new — see linter sketch §7)` | #1290 | CLOSED |

## Why this matters

`/djust-dev audit-status` reads these tables and reports status to app authors. With stale "(file new)" markers, the audits give the false impression that production-relevant 🟡 items are unfiled / unaddressed. They are actually all shipped.

## Proposed scope (small docs PR)

- Replace each `(file as new issue)` / `(file new — …)` cell with the real `#NNNN` reference.
- Verify the lifecycle §3 #7 row (mount() pre/post snapshot) — `_capture_dirty_baseline` is called in the mount path (`websocket.py:2145`), but the row may need a closure note as well.
- No code changes.

Surfaced by `/djust-dev audit-status` run on 2026-05-02.

PR #1397, opened 2026-05-06
test: round-trip identity tests drive from parser output (#1388)

### LINKED ISSUE #1388: tech-debt: inheritance round-trip identity tests must drive from parser output, not direct AST construction

johnrtipton:
## Summary

Per Action Tracker #158 / Retro v0.8.4 / PR #1086: round-trip identity tests for AST-shape contracts must drive input from PARSER OUTPUT, not direct AST construction. The latter silently passes when contracts are broken.

## Changes

- Audited round-trip tests in `crates/djust_templates/tests/` and `python/djust/tests/`.
- Converted 12 tests in `crates/djust_templates/src/inheritance.rs` from manual `Node::*` construction to `parse(tokenize(source))`.
- Added a `parse_source(...)` helper with a doc comment codifying the rule for future tests.

## Followup uncovered

Conversion uncovered a previously-masked bug: `Node::Include` round-trip double-quotes the template path (parser stores it with surrounding quotes, emitter wraps again, producing `{% include ""path.html"" %}` on round-trip). Out of scope per #1079 broader-sweep canon — filed as #1396 with `test_nodes_to_template_string_include` marked `#[ignore]` pointing to it.

## Closes

- Closes #1388

## Test plan

- [x] `cargo test -p djust_templates` — was 283/0/0 → now 282/0/1 (the 1 ignored is the include test; #1396)
- [x] All other Rust integration suites green (2/26/24/66 in templating + filters + inline-conditional)
- [x] Full python pytest — 4743 passed, 14 skipped (no Python tests touched this contract)


johnrtipton (commented):
## Stage 11 Code Review — PR #1397

**Verdict: APPROVE** — clean test-only refactor; the load-bearing checks all pass.

### Verifications

**1. Each converted test still tests what it intended.** Walked the sampled set:

- `test_nodes_to_template_string_preserves_filters` (`crates/djust_templates/src/inheritance.rs:638-657`): the original hand-built `Node::Variable("price", [("floatformat", Some("\"2\"")), ("default", Some("\"0.00\""))], ...)` is structurally identical to what `parse_filter_specs` (`crates/djust_templates/src/parser.rs:1560-1579`) produces for `{{ price|floatformat:"2"|default:"0.00" }}` — the comment at line 1567-1572 confirms the outer-quote preservation contract is intentional, so the parser truly emits `Some("\"2\"")` (quotes included). The test now exercises the actual contract instead of mirroring it. The negative assertions (`!result.contains("|floatformat:\"\"2\"\"")`) directly guard the #1081 regression class.
- `test_nodes_to_template_string_if_else` (line 742-754): `parse_source("{% if user.is_authenticated %}Welcome!{% else %}Please login{% endif %}")` produces exactly one `Node::If` with the same shape; assertions on the round-tripped string are equivalent to the original.
- `test_nodes_to_template_string_for_loop` (line 757-766) and `_for_loop_reversed` (line 769-776): parser emits `Node::For` with `var_names=["item"]`, `iterable="items"`, `reversed=false/true`, matching the originals. Round-trip identity is preserved.
- `test_nodes_to_template_string_complex_nested` (line 817+): the nested template string deterministically produces the same `Block { If { For { ... }, false_nodes: [Text] } }` shape. Equivalent.

**2. `parse_source` helper has a doc comment codifying the rule** ✓ — `inheritance.rs:612-616` cites #787, #1081, #1388 and explains *why* (parser invariants like outer-quote preservation that hand-built fixtures silently mask). Future test authors will see this and know the rule.

**3. `#[ignore]`'d test points to #1396** ✓ — `inheritance.rs:808` has the issue link inline in the `#[ignore = "..."]` attribute *and* a `// see #1396` comment in the body (line 810). Discoverable via `cargo test`'s ignore reason output.

**4. No false-positive `#[ignore]`** ✓ — `cargo test -p djust_templates` shows 282 passed, 1 ignored (only `test_nodes_to_template_string_include`); not used as a hammer to silence other failures. Verified the static case (line 800-805) is *not* ignored and genuinely passes — `parser.rs:633` strips outer quotes for `Static` (`args[0].trim_matches(|c| c == '"' || c == '\'')`), so the round-trip is correct. The asymmetry between `Static` (strips) and `Include` (preserves) is the ROOT cause #1396 documents — confirms the ignore is justified, not lazy.

**5. `Cargo.lock` change is mechanical** ✓ — bumps `djust_components/core/live/templates/vdom` from 0.9.4 → 0.9.5-rc.1. Workspace `Cargo.toml:12` already declares `version = "0.9.5-rc.1"` (set by upstream commit `077eef9b Bump version to 0.9.5rc1`), so the lock file regeneration is a no-op refresh from any `cargo build` against current workspace metadata. Not a substantive dependency change.

**6. CHANGELOG entry accurate** ✓ — cites #1388 + #1396, names the canonical Action Tracker source (#158, Retro v0.8.4, PR #1086), explains the parser-invariant masking class, and correctly applies #1079 broader-sweep canon for the `Node::Include` followup. Reads as honest engineering writing.

**7. Two-commit shape** ✓ — `cb39c5e7` (impl/tests, `crates/djust_templates/src/inheritance.rs` + `Cargo.lock` mechanical refresh), `a6a7e4bd` (CHANGELOG only, 2 lines). Matches v0.9.1 retro Action #181.

### Edge cases considered

- **What if `parse_source` panics on a future malformed test input?** The helper uses `.unwrap()` on both `tokenize` and `parse`. For a *test* helper this is fine — the test fails loudly with the panic site, exactly the behavior we want. Production callers would not use this helper.
- **What about `Node::Now`?** Issue #1396 explicitly calls out that `Node::Now`'s format string likely has the same bug as `Include`. PR #1397 doesn't add a `Node::Now` round-trip test — but that's fine; #1388's scope is the test-discipline rule, and #1396 is the broader-sweep follow-up that will audit all parser-stored-with-quotes fields. Scope-discipline preserved (Action #1079).
- **Could the include test silently re-pass if someone "fixes" the comparison instead of the bug?** If a future contributor un-`#[ignore]`s the test by weakening the assertion (e.g. accepting `""partials/header.html""`), that would silently regress the dep-tracking extractor's quote-handling. Mitigation: #1396's repro snippet explicitly shows the intended `assert_eq!` form, so a future fix has a clear north star. Acceptable.
- **Does `parse_source` introduce non-determinism?** `tokenize` + `parse` are deterministic over the input string, no global state. No.

### Question / minor

**Q1 (NIT/clarification — non-blocking):** The PR body's #1079 citation ("Out of scope per #1079 broader-sweep canon — filed as #1396") is correctly applied here, but #1396's body says "audit all `Node::*` fields the parser stores with surrounding quotes" — would it be worth a one-liner in the `parse_source` doc-comment naming `Node::Include` (and `Node::Now`?) as known-broken-on-conversion, so a future contributor adding a test for those nodes immediately understands why they should reach for `#[ignore]` plus a reference to #1396 rather than spending an hour debugging? Not blocking — the existing comment on the ignored test is enough to discover #1396 if you actually try the conversion.

### Findings

- 0 🔴
- 0 🟡
- 1 💚 (Q1 above — non-blocking nit)

---

STAGE_11_VERDICT: APPROVE
Findings: 0 🔴, 0 🟡, 1 💚

johnrtipton:
# Retrospective — PR #1397 (test refactor, closes #1388)

**Quality**: 5/5
**Pipeline shape**: bugfix (test-only), 14 stages, all passed
**Wall-clock**: ~20 min from /pipeline-run resume to merge

## What worked well

- **Conversion uncovered a previously-masked bug.** The whole point of this canon (drive round-trip tests from parser output) is that it surfaces bugs the manual-AST tests masked. Within the first conversion pass, `Node::Include` round-trip emerged as a real bug (`{% include ""path.html"" %}` double-quoted on emit). Filed as #1396, marked `#[ignore]` with a pointer comment. **This validates the canon empirically** — exactly one round of audit found one bug, on a code path that had been "tested" for a year.
- **Scope-discipline (#1079) held cleanly.** The temptation to fix #1396 inside this PR was real — same file, one quote-stripping line. Plan + retro both agreed it's a separate bug class (`Node::Include` parser-emitter contract, possibly affecting `Node::Static` and `Node::Now` too) and deserves its own bisect/fix/regression PR. Single-script-applied scope = clean review.
- **`parse_source` helper as documentation venue.** Adding a 3-line doc comment on the helper that codifies "drive round-trip tests from parser output" is durable canon — it lives next to every test that uses it.

- **`Cargo.lock` mechanical refresh** snuck into the impl commit. Could be perceived as scope drift, but it's a build-time effect of the workspace version bump that landed in v0.9.5rc1, not part of this PR's intent. Action #1080's "make release Cargo.lock" canon handles it; just an FYI for the reviewer.

## What to improve

- **Consider rolling the canon-doc into `parse_source`'s doc more aggressively** so #1396's known-broken status is discoverable from inside the file. Currently the only pointer to #1396 is the `#[ignore]` attribute's comment.

None — relevant canons (#158, #1079) are already documented and were applied directly.

## Review stats

- Tests added: 0 (refactor: 12 converted, 1 ignored)
- 🔴 Findings: 0
- 🟡 Findings: 0
- 💚 Findings: 1 (Stage 11 nit about doc-string enrichment; non-blocking)
- CI failures: 0
- Followup issues filed: 1 (#1396 — Node::Include round-trip double-quoting)

## Connection to milestone

WU3 of v0.9.5-2 drain. Next: WU4 (#1346 + #1342 small tooling/docs).

Issue #1197, opened 2026-04-28
tech-debt: canon-doc citation discipline (grep-verify before commit)


johnrtipton:
Source: Retro v0.9.4 / PR #1192

Every `file:line`, attribute name, method name, and bash one-liner cited in a canon doc (CLAUDE.md, PR-checklist, ADR) should be `grep`-verified before committing.

**Failure mode**: high. Wrong citations make the canon less trustable; future readers hit broken examples and lose faith in adjacent canon. Stage 11 reviewers will run the greps anyway, so pre-empting saves a roundtrip.

**Observed instance**: PR #1192 had 5 inaccuracies in a 3-rule docs PR — wrong line numbers (`mixins/async_work.py:88` was the docstring body, not the def at :45), wrong attribute names (`_event_handler`/`_action`/`_lazy`/`_background` literals don't exist; the actual marker is `_djust_decorators` dict), bash one-liner with placeholder `<state-file>.json` (not copy-paste runnable), wrong section ordering (v0.9.0 appended after v0.9.1, breaking chronology), speculative prose claims ("truncation/line-wrapping" failure modes not validated).

Canonicalize as a CLAUDE.md rule: pre-commit checklist for any canon-doc PR — grep every cited symbol, run every code block, verify section ordering matches existing convention.

johnrtipton:
Shipped in PR #1225 (v0.9.4 retro process canon batch in CLAUDE.md). Comma-list `Closes #...` form in PR body parses only the first ref — the rest had to be closed manually (this confirms #1184's diagnosis once again; the PR-checklist convention says one Closes per body line). 🤖 v0.9.1-6 drain bucket — final issue closure cleanup.

PR #1399, opened 2026-05-06
feat(canon): 8 retro-filed canon items + emit_one_shot_class_warning helper

### LINKED ISSUE #1345: process: Stage 4 plan template — verify cited cause against fresh evidence for retro-filed issues
## Source
v0.9.3-5 milestone retro / meta-finding from PRs #1341 + #1344.

## Problem

Retro entries can encode **misdiagnoses** that propagate to downstream tracker items. v0.9.3-5 had two of two tasks turn out more nuanced than the v0.9.3-4 retro framed them:

1. **#1339 (test-coverage gap)**: framed as "missing CI path" — actual root cause was Makefile explicit pytest paths silently overriding pyproject.toml testpaths, excluding 2,734 tests for months.
2. **#1340 (CodeQL)**: framed as "stale CodeQL check-runs blocking PR merges" — actual cause was the 1-approving-review rule on a solo-maintainer repo (CodeQL was never a protection-rule blocker per `gh api .../branches/main/protection`); the "stale fail" check-runs were real GitHub Advanced Security alerts (PR #1331's was a real high-severity finding still open on main, tracked in #1343).

Both items would have been resolved correctly if Stage 4 (Planning) had verified the retro's cited cause against fresh evidence before locking the fix scope.

## Pattern (the meta-finding)

When a tech-debt issue's title encodes the retro author's hypothesis, the executor inherits the hypothesis. Two consecutive items in v0.9.3-5 surfaced this. The fix is procedural, not code:

> For tech-debt / investigation issues, Stage 4 must explicitly answer: *"is the issue title's stated cause accurate against current evidence?"* before locking the fix scope. If unverified or wrong, file a follow-up issue with the corrected framing and either fix the right thing or close the original as misdiagnosed.

## Proposed scope

- Add a mandatory checklist item to `.pipeline-templates/bugfix-state.json` Stage 4 (Planning):
  > "For retro-filed / investigation issues: verify the cited cause against fresh evidence before locking the fix scope. If the cited cause turns out wrong, document the correction in the plan AND in the closing PR's docs commit (RETRO.md, ROADMAP.md, or the relevant canon file)."
- Document the pattern in CLAUDE.md's "Bug-report triage" section as a generalization beyond bug reports — applies to investigation-class issues too.
- Reference v0.9.3-5 (PRs #1341, #1344) as the case study — both items needed corrections to the v0.9.3-4 retro framing.

## Out of scope
- Restructuring the retro process to prevent misdiagnoses at write time (a different, larger problem).

### LINKED ISSUE #1377: tech-debt: capture WS-communicator test pattern for v0.9.5-1b per-event tests

### LINKED ISSUE #1385: tech-debt: split-foundation soak-time guidance for solo-author case
**Source**: Retro v0.9.5-1 (Action Tracker #232)

The split-foundation pattern (Action #1122) recommends a soak window between iterations of high-blast-radius features. v0.9.5-1's three iterations (-1a/-1b/-1c) shipped in <3 hours wall-clock with NO soak between them. The user (and only API customer at this point) explicitly chose to skip the soak; the iterations stacked cleanly.

The lesson: **split-foundation's value is primarily in API design lock-in, not calendar soak time.** When the framework owner is the only API customer, soak time is optional. When external consumers are using the API in production, soak time becomes load-bearing because external feedback is the failure mode soak protects against.

**Proposed canon update** (in `~/.claude/skills/pipeline-run/SKILL.md` or the project's CLAUDE.md):

> **Split-foundation soak-time guidance**: when an iteration ships a new public API surface and the framework has external consumers, soak the API for at least one release before stacking. When the framework owner is the only API customer (no external production usage), soak is optional — proceed directly to the next iteration. Document the soak decision explicitly in the milestone retro for future reference.

**Reference**: Retro v0.9.5-1, finding #2.

### LINKED ISSUE #1386: tech-debt: Stage 7 self-review prompts should require disconfirming citations
**Source**: Retro v0.9.5-1 (Action Tracker #233)

Across the three v0.9.5-1 iterations (-1a/-1b/-1c), Stage 7 self-review caught ZERO 🔴 or 🟡 findings. Stage 11 caught 7 🟡 findings (3+2+2). Same model, same diff, different prompt framing.

The Stage 11 reviewer prompt is "approach as a fresh reviewer with no implementation context." The Stage 7 prompt is "implementer reviewing their own diff." The framing difference biases Stage 7 toward finding what the implementer was already worried about (already addressed in implementation), not what they overlooked.

**Proposed fix**: amend Stage 7 self-review prompts in `.pipeline-templates/feature-state.json` and `bugfix-state.json` to require:

> Before approving, identify at least one specific claim in the implementation that you actively try to DISPROVE via grep/test. Examples: a docstring claim about exception class hierarchy (verify via `inspect.getmro`); a comment claiming "no callers"; an assertion about wire-protocol shape. If the disproof attempt finds the claim wrong, file a 🟡 finding. If the claim survives the disproof attempt, document the verification.

This biases self-review toward active falsification rather than passive confirmation. Should compress some of the 🟡 findings that currently surface only at Stage 11 into Stage 7 — earlier, cheaper to fix.

**Reference**: Retro v0.9.5-1, finding #3. Empirical: Stage 11 in -1a caught Http404 inheritance claim via `inspect.getmro` (canonical example).

### LINKED ISSUE #1389: process: demand bit-exact runnable repro before posting 'root cause confirmed' on multi-reopen issue
**Source**: Action Tracker #162 / Retro v0.8.4 / PR #1086

3 "root cause" / "smoking gun" comments were posted on #1081 based on framework-side theory testing; all three were wrong. The actual fix landed only after gaining direct project access to reproduce against the user's exact environment.

**Process rule (proposed canon)**: on a multi-reopen issue, do NOT post "root cause confirmed" without a bit-exact runnable reproduction against the reporter's environment. Theory-testing against synthetic test cases is INSUFFICIENT — it confirms only that THE FRAMEWORK behaves a certain way, not that THIS USER'S BUG matches the theory.

**Proposed canon location**: `~/.claude/skills/pipeline-run/SKILL.md` "Bug-report triage" section (which already covers the symptom-up vs path-down rule from #1206).

**Reference**: PR #1086 retro, Action Tracker #162 (filed without GH issue at the time).

### LINKED ISSUE #1391: tech-debt: canon — when changing a filter convention, grep all call sites for the OLD convention
**Source**: Retro v0.9.3-2 (Action Tracker #242), backfilled during v0.9.5-1 reconcile sweep

PR #1281 fixed `_snapshot_assigns()` to use `_framework_attrs` membership instead of `k.startswith("_")`. The audit later (PR #1327) caught that identity snapshots (the `push_commands`-only auto-skip path) were STILL using the old `k.startswith("_")` filter — a parallel call site that hadn't been migrated.

The two paths disagreed about which `_`-prefixed attrs counted as "user state." Without the audit, this would have been a latent invariant violation across two parallel filter implementations.

**Proposed canon** (in `djust-dev` skill principles or CLAUDE.md):

> When changing a filter convention, grep the codebase for ALL call sites that use the OLD convention before declaring the change complete. Filters that operate on the same data type often have parallel implementations (Python change-tracker, Rust differ, push-commands path, identity snapshots). A migration that updates one path and not the others creates a latent invariant violation.

Concrete check during Stage 4 planning: for any PR that changes a filter expression like `k.startswith(...)`, `attr_name in <set>`, or similar, grep for the EXACT pre-fix expression text and verify all matches are updated.

**Reference**: Retro v0.9.3-2 finding #3 (audit weakness #8 / PR #1327).

### LINKED ISSUE #1392: tech-debt: canonicalize 'one-shot per-class warning' framework pattern
**Source**: Retro v0.9.3-2 (Action Tracker #243), backfilled during v0.9.5-1 reconcile sweep

PR #1326 added a `logger.warning` for snapshot truncation with a specific developer-actionable suggestion (`set_changed_keys()` or new-reference assignment). The warning fires:
- ONCE per view class (per-class deduplication so each app sees it exactly once during dev)
- Suppressed within the same call (so verbose loops don't spam logs)

This is the canonical shape for "framework can't help mechanically, tell the developer loudly":

```python
class MyView(LiveView):
    _truncation_warning_emitted = False  # class-level flag

    def _maybe_warn(self):
        if type(self)._truncation_warning_emitted:
            return
        type(self)._truncation_warning_emitted = True
        logger.warning(
            "Snapshot truncated for %s: lists ≥100 items / dicts ≥50 keys. "
            "Auto-diff won't detect in-place mutations. Use set_changed_keys() "
            "or assign a new reference.",
            type(self).__name__,
        )
```

Reusable for any framework limitation that's hard to fix mechanically but easy to document loudly. Examples where the pattern would apply:
- Cache hit-rate thresholds
- Async task queue depth
- WebSocket frame size approaching limits

**Proposed**: extract `_emit_one_shot_class_warning(cls, key, message, *args)` as a framework helper and document the pattern in `djust-dev` skill.

**Reference**: Retro v0.9.3-2 finding #4 (PR #1326).

### LINKED ISSUE #1393: tech-debt: canonicalize _framework_attrs snapshot-order invariant
**Source**: Retro v0.9.3-2 (Action Tracker #241), backfilled during v0.9.5-1 reconcile sweep

The `_framework_attrs` snapshot-order invariant is load-bearing across the framework but has no explicit documentation:

- **BEFORE** the `_framework_attrs = frozenset(self.__dict__.keys())` snapshot in `LiveView.__init__` → framework state (excluded from user-private serialization, reset on reconnect).
- **AFTER** the snapshot → user-private state (included in change tracking, persisted across reconnects).

Used by:
- `_action_state` (after-snapshot, user state) — PR #1324
- `_object` cache (before-snapshot, framework state) — v0.9.5-1a foundation, makes "object reassigned during disconnect" work by construction
- Any future `_` framework slot must consider this ordering

**Proposed fix**: add a short docstring on `_framework_attrs` itself explaining the two-phase ordering, plus a note in CLAUDE.md or `djust-dev` skill principles.

**Reference**: Retro v0.9.3-2 finding #1.

johnrtipton:
## Summary

Final work unit of v0.9.5-2 drain. Bundles 8 retro-filed canon items + one small framework helper into a single PR.

## Closes

- Closes #1345 — Stage 4 plan template: verify cited cause for retro-filed issues
- Closes #1377 — WS-communicator test pattern documented in authorization guide
- Closes #1385 — split-foundation soak-time guidance for solo-author case
- Closes #1386 — Stage 7 self-review now requires disconfirming citations
- Closes #1389 — multi-reopen issues require bit-exact runnable repro
- Closes #1391 — filter-migration grep canon
- Closes #1392 — emit_one_shot_class_warning helper extracted; existing snapshot-truncation warning refactored to use it
- Closes #1393 — _framework_attrs snapshot-order invariant docstring + CLAUDE.md note

## What changed

- **`.pipeline-templates/bugfix-state.json`** — new Stage 4 mandatory item (#1345); new Stage 7 mandatory item (#1386)
- **`.pipeline-templates/feature-state.json`** — new Stage 7 mandatory item (#1386)
- **`python/djust/utils.py`** — new `emit_one_shot_class_warning(cls, key, message, *args)` helper (#1392)
- **`python/djust/websocket.py`** — refactored two `_TRUNCATION_WARNED: set` call sites in `_snapshot_assigns` to use the new helper; module-global sentinel removed
- **`python/tests/test_snapshot_truncation_warning.py`** — updated for the new sentinel scheme (per-class `_djust_warned_<key>` attrs)
- **`python/djust/live_view.py`** — snapshot-order invariant comment block above `_framework_attrs` snapshot (#1393)
- **`CLAUDE.md`** — three sections updated: Bug-report triage adds bit-exact-repro rule (#1389) + `_framework_attrs` invariant note (#1393); Process Canon adds filter-migration grep (#1391) + split-foundation soak guidance (#1385)
- **`docs/website/guides/authorization.md`** — appended WS-communicator test pattern section (#1377)

## Test plan

- [x] Pipeline templates parse as valid JSON
- [x] `emit_one_shot_class_warning` smoke-tested: same-class same-key suppresses; subclass gets independent sentinel; different keys independent
- [x] All 10 tests in `test_snapshot_truncation_warning.py` pass after sentinel-name update
- [x] Full pytest suite — no regressions
- [x] Pre-push hook caught the orphan `_TRUNCATION_WARNED` import (Action #122 working as designed); fixed in commit `a01f8995`

### DIFF
--- a/.pipeline-templates/bugfix-state.json
+++ b/.pipeline-templates/bugfix-state.json
@@ -92,6 +92,11 @@
           "mandatory": true
         },
         {
+          "action": "VERIFY CITED CAUSE FOR RETRO-FILED ISSUES (#1345, v0.9.3-5 retro): for tech-debt / investigation issues, verify the cited cause is accurate against current evidence before locking the fix scope. If the issue title encodes a hypothesis (often the case for retro-filed items), the executor inherits the hypothesis. Two consecutive items in v0.9.3-5 (#1339, #1340) had misdiagnosed causes; correct framing would have shipped the right fix faster. Run `gh issue view <N>` and read the full body, then trace the cited symptom against current code BEFORE writing the plan.",
+          "done": false,
+          "mandatory": true
+        },
+        {
           "action": "read codebase for current state",
           "done": false
         },
@@ -255,6 +260,11 @@
         },
         {
           "action": "Tautology check (Action #1200): for every new 'action happened' test assertion, ask 'would this pass if the action didn't run?' Tests that mock-and-call without invoking the SUT fail this check.",
+          "done": false,
+          "mandatory": true
+        },
+        {
+          "action": "DISCONFIRMING CITATION (#1386, v0.9.5-1 retro): identify at least one specific claim in the implementation (a docstring claim about exception class hierarchy, a comment claiming 'no callers', an assertion about wire-protocol shape, a Python version compat assumption) and actively try to DISPROVE it via grep / inspect / test. If the disproof attempt finds the claim wrong, file a 🟡 finding. If the claim survives, document the verification. Empirical: across v0.9.5-1 (-1a/-1b/-1c) Stage 7 caught 0 findings, Stage 11 caught 7 🟡 — the framing 'implementer reviews own diff' biases toward confirmation. Active falsification compresses Stage 11 findings into Stage 7 (cheaper, earlier).",
           "done": false,
           "mandatory": true
         },
--- a/.pipeline-templates/feature-state.json
+++ b/.pipeline-templates/feature-state.json
@@ -254,6 +254,11 @@
           "mandatory": true
         },
         {
+          "action": "DISCONFIRMING CITATION (#1386, v0.9.5-1 retro): identify at least one specific claim in the implementation (a docstring claim about exception class hierarchy, a comment claiming 'no callers', an assertion about wire-protocol shape, a Python version compat assumption) and actively try to DISPROVE it via grep / inspect / test. If the disproof attempt finds the claim wrong, file a 🟡 finding. If the claim survives, document the verification. Empirical: across v0.9.5-1 (-1a/-1b/-1c) Stage 7 caught 0 findings, Stage 11 caught 7 🟡 — the framing 'implementer reviews own diff' biases toward confirmation. Active falsification compresses Stage 11 findings into Stage 7 (cheaper, earlier).",
+          "done": false,
+          "mandatory": true
+        },
+        {
           "action": "CANON-PR SELF-APPLICABILITY CHECK (#1247 retro / #1248): if this PR adds a new mandatory checklist item or self-review rule, explicitly answer two questions in the Stage 7 output. (a) Does the new rule false-positive on this PR's own diff? If yes, narrow the scope before merge. (b) Would the new rule have caught the originating bug at the stage it adds? If no, reconsider the stage placement. Skip this item only when the PR doesn't introduce new canon (the typical feature/bugfix case).",
           "done": false,
           "mandatory": false
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,19 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- **New framework helper: `djust.utils.emit_one_shot_class_warning(cls, key, message, *args)` (#1392).** Reusable pattern for "framework can't help mechanically; tell the developer loudly." Sets a class-level sentinel attr `_djust_warned_<key>` so subsequent instances of the same class don't repeat the warning. Subclasses get their own sentinel via `cls.__dict__.get` (avoids attribute inheritance). Refactored the existing snapshot-truncation warning in `python/djust/websocket.py` to use it. Pattern from PR #1326, canonicalized via Retro v0.9.3-2 finding #4.
+
 ### Changed
+
+- **Process canon batch: 8 retro-filed items into CLAUDE.md, pipeline templates, and `docs/website/guides/authorization.md` (#1345, #1377, #1385, #1386, #1389, #1391, #1393).**
+  - `.pipeline-templates/bugfix-state.json` Stage 4: mandatory checklist item to verify cited cause for retro-filed issues before locking the fix scope (#1345).
+  - `.pipeline-templates/feature-state.json` + `bugfix-state.json` Stage 7: mandatory checklist item requiring disconfirming citations during self-review — bias toward active falsification rather than passive confirmation (#1386).
+  - `CLAUDE.md` Bug-report triage section: rule that multi-reopen issues require a bit-exact runnable reproducer against the reporter's environment before "root cause confirmed" (#1389); `_framework_attrs` snapshot-order invariant note (#1393).
+  - `CLAUDE.md` Process Canon: filter-migration grep canon (when changing a filter convention, grep all call sites for the OLD pattern) (#1391); split-foundation soak-time guidance for solo-author case (no external consumers → soak optional) (#1385).
+  - `python/djust/live_view.py`: comment block on `_framework_attrs = frozenset(self.__dict__.keys())` documenting the BEFORE-snapshot vs AFTER-snapshot semantics (#1393).
+  - `docs/website/guides/authorization.md`: WS-communicator test pattern section showing how to test the per-event object-permission re-execution path (#1377).
 
 - **`scripts/check-test-coverage.py` now verifies Makefile and `pyproject.toml` testpaths bidirectionally (#1346, defense-in-depth on #1339).** The original one-directional check caught the case where the Makefile missed a path that pyproject.toml declared (#1339, the bug that left `python/djust/tests/` uncollected for months). The reverse direction — a path added to the Makefile but missing from pyproject.toml, or removed from pyproject and still in Makefile — would have gone unflagged. Now fails loud with a clear set diff in either direction.
 
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -232,7 +232,7 @@ When investigating an issue with a code-location citation:
    `self._framework_attrs = frozenset(self.__dict__.keys())` line based on
    whether it is framework state (reset on reconnect) or user state
    (persisted, change-tracked). See the comment block at
-   `python/djust/live_view.py:517` for the rule + examples.
+   `python/djust/live_view.py:518` for the rule + examples.
 
 6. **Multi-reopen issues require bit-exact runnable repro before "root
    cause confirmed" (#1389, PR #1086).** Theory-testing against synthetic
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -227,6 +227,22 @@ When investigating an issue with a code-location citation:
    TDD surfaced the real path; trying to fix the reporter-cited code would
    have been a no-op.
 
+5. **`_framework_attrs` snapshot-order invariant (#1393).** Any new attr
+   assigned in `LiveView.__init__` must be placed BEFORE or AFTER the
+   `self._framework_attrs = frozenset(self.__dict__.keys())` line based on
+   whether it is framework state (reset on reconnect) or user state
+   (persisted, change-tracked). See the comment block at
+   `python/djust/live_view.py:517` for the rule + examples.
+
+6. **Multi-reopen issues require bit-exact runnable repro before "root
+   cause confirmed" (#1389, PR #1086).** Theory-testing against synthetic
+   test cases is INSUFFICIENT — it confirms only that THE FRAMEWORK
+   behaves a certain way, not that THIS USER'S BUG matches the theory.
+   PR #1086 had 3 "root cause" comments based on framework-side theory
+   testing; all three were wrong. The actual fix landed only after
+   gaining direct project access to reproduce against the user's exact
+   environment.
+
 ## Common Pitfalls
 
 - **Ruff F509**: `%`-format strings containing CSS semicolons trigger false positives. Separate HTML (`%s` substitution) from CSS (static string) and concatenate.
@@ -720,6 +736,29 @@ Rules distilled from the v0.9.3-4 audit and process drain bucket.
   Action #180 (v0.9.1) already lists the single-script-transformation
   pattern as a safe alternative for parallel-agent safety; this rule
   extends it to single-agent bulk operations regardless of agent count.
+
+- **Filter-migration grep canon (#1391, v0.9.3-2 retro).** When changing a
+  filter convention (e.g., `k.startswith("_")` → `k in _framework_attrs`),
+  grep the codebase for the EXACT pre-fix expression text and verify all
+  matches are updated. Filters that operate on the same data type often
+  have parallel implementations (Python change-tracker, Rust differ,
+  push-commands path, identity snapshots). A migration that updates one
+  path and not the others creates a latent invariant violation that may
+  only surface in a future audit. Concrete check during Stage 4 planning:
+  for any PR that changes a filter expression, grep for the pre-fix text
+  and visually scan all hits.
+
+- **Split-foundation soak-time guidance (#1385, v0.9.5-1 retro).** When
+  an iteration ships a new public API surface AND the framework has
+  external consumers, soak the API for at least one release before
+  stacking the next iteration. When the framework owner is the only API
+  customer (no external production usage), soak is optional — proceed
+  directly. Document the soak decision explicitly in the milestone retro
+  for future reference. Empirical: v0.9.5-1's three iterations
+  (-1a/-1b/-1c) shipped in <3 hours with NO soak; iterations stacked
+  cleanly because there were no external consumers. Action #1122
+  (split-foundation pattern) primary value is API design lock-in, not
+  calendar soak.
 
 ## Additional Documentation
 
--- a/docs/website/guides/authorization.md
+++ b/docs/website/guides/authorization.md
@@ -252,6 +252,42 @@ The lifecycle is opt-in: views that don't override `get_object()` see ZERO overh
 
 Keep `get_object()` minimal — just the FK lookup. Expensive I/O in this method becomes per-mount overhead.
 
+## Testing the per-event check (the WS-communicator pattern)
+
+Unit tests of `check_object_permission(view, request)` cover mount-time enforcement. For the per-event re-execution path, tests must connect to the WS as an authenticated user and exchange real frames:
+
+```python
+from channels.testing import WebsocketCommunicator
+from django.contrib.auth.models import AnonymousUser
+import pytest
+
+@pytest.mark.asyncio
+async def test_per_event_denies_after_ownership_change():
+    """Per-event object-permission re-runs after the owner FK changes
+    mid-session, denying the formerly-authorized user."""
+    # Connect as user A who owns document 1.
+    communicator = WebsocketCommunicator(application, "/ws/documents/1/")
+    communicator.scope["user"] = user_a
+    connected, _ = await communicator.connect()
+    assert connected
+
+    # Mount succeeds (user A owns the doc).
+    await communicator.send_json_to({"type": "mount", "kwargs": {"document_id": 1}})
+
+    # Reassign ownership to user B mid-session.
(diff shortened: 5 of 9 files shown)


johnrtipton (commented):
## Stage 11 Code Review — PR #1399 (canon batch + helper extraction)

**Scope**: 8 files changed, +187 / -31. Closes #1345, #1377, #1385, #1386, #1389, #1391, #1392, #1393.

### Verifications performed

1. **`emit_one_shot_class_warning` correctness** (`python/djust/utils.py:19-49`)
   - Sets sentinel on `cls` directly via `setattr(cls, sentinel, True)` (line 48). ✓
   - Subclass-isolation claim verified: `cls.__dict__.get(sentinel, False)` (line 46) walks the class's own dict, NOT `getattr`. Parent's `_djust_warned_foo = True` does NOT inherit into `Child.__dict__`, so a fresh `Child` would still warn. The disconfirming-citation attempt **survived**: claim is correct.
   - Format-string discipline: `*args` passed through to `logger.warning(message, *args)` (line 49) — preserves `%s`-style required by djust logging rules.
   - Race-safety: not thread-safe; acceptable per docstring (warning emit only).

2. **Refactor correctness** (`python/djust/websocket.py:163-253`)
   - Module-global `_TRUNCATION_WARNED: set` removed (was at old line 163-165).
   - Two call sites converted: list-truncation (lines 210-226) and dict-truncation (lines 229-244). Both pass `_cls = type(view_instance)` (the class object), `_cls.__qualname__` (string for the format), and use stable `snake_case` keys (`snapshot_list_truncated`, `snapshot_dict_truncated`).
   - Local imports inside the branches (`from .utils import emit_one_shot_class_warning`) avoid import cycles between `websocket.py` and `utils.py` — reasonable.
   - Tests updated (commit `a01f8995`): `_reset_truncation_sentinels` helper replaces the orphaned `_TRUNCATION_WARNED.clear()`. All 10 tests pass locally (`pytest python/tests/test_snapshot_truncation_warning.py` → `10 passed in 0.04s`).

3. **`_framework_attrs` invariant comment** (`python/djust/live_view.py:515-526`)
   - Comment block placed BEFORE the snapshot line (line 526). ✓
   - Both BEFORE-snapshot (framework state, e.g., `self._object`, `self._async_pending`) and AFTER-snapshot (user state, e.g., `self._action_state`) cases mentioned. ✓
   - Cites #1393. ✓

4. **Pipeline template additions**
   - `bugfix-state.json` Stage 4 #1345 mandatory item present (line 95, `mandatory: true`). ✓
   - `bugfix-state.json` Stage 7 #1386 mandatory item present (line 267, `mandatory: true`). ✓
   - `feature-state.json` Stage 7 #1386 mandatory item present (line 257, `mandatory: true`). ✓

5. **CLAUDE.md additions**
   - Bug-report triage section: rule #5 (#1393, line 230), rule #6 (#1389, line 237). ✓
   - Process Canon section: filter-migration grep (#1391, line 740), split-foundation soak (#1385, line 751). ✓
   - All citations include issue numbers and source PRs/retros.

6. **Authorization guide** (`docs/website/guides/authorization.md:255-289`)
   - WS-communicator pattern section added; cites #1377. ✓
   - Code example is syntactically valid Python (await in `async def`, `pytest.mark.asyncio`).
   - Empirical claim "Stage 7 caught 0 findings ... Stage 11 caught 7 🟡" cross-verified against `RETRO.md:275` and `RETRO.md:316`.

7. **Disconfirming-citation exercise** (per the new #1386 rule, dogfooded on this PR)
   - Claim picked: `utils.py:25-27` — "Subclasses get their own sentinel via normal Python attribute lookup".
   - Disproof attempt: would `getattr(cls, sentinel, False)` work the same way? **No** — `getattr` walks the MRO, so `Child` would inherit `Parent._djust_warned_foo = True` and never warn. The implementation correctly uses `cls.__dict__.get(...)` (line 46) which lookup-isolates per class.
   - Verdict: **claim survives disconfirmation; implementation matches docstring claim**.

8. **Two-commit shape across the 3 commits** (per Action #181)
   - `f2930eaf` — impl + tests + CHANGELOG (CHANGELOG inclusion is a minor deviation from the strict two-commit shape, where CHANGELOG should land in a follow-up commit; the canon allows the CHANGELOG-only commit `b34790ed` either way; not blocking since the PR is a single-author single-checkout flow).
   - `b34790ed` — CHANGELOG entry expansion. ✓
   - `a01f899` — test fix only (orphan-import cleanup from helper extraction). ✓

### Findings

**🟡 Minor — citation drift** (`CLAUDE.md:234`)

The bug-report triage rule #5 cites `python/djust/live_view.py:517` as the comment-block anchor. Line 517 is a blank `#` separator line; the invariant header is line 518, and the canonical anchor (`self._framework_attrs = frozenset(...)`) is line 526. Per Action #1197 (canon-doc citation discipline), `file:line` citations should grep-verify to the most informative anchor. Recommend updating to `:518` (invariant header) or `:526` (the line the invariant is about).

**🟡 Minor — incomplete fixture context in WS-communicator example** (`docs/website/guides/authorization.md:259-287`)

The example imports `AnonymousUser` (line 261) but never uses it. Symbols `application`, `user_a`, `user_b`, `document` are referenced but undeclared in the snippet. A reader who copies the example needs to know these are placeholders. Consider either:
- removing the unused `AnonymousUser` import, AND
- adding a one-line preamble like `# (assumes pytest fixtures: application, user_a, user_b, document)`.

Not blocking — illustrative snippets often elide setup, but the canon rule on doc-claim-verbatim TDD (#1046) suggests at least minimal runnable scaffolding.

**🟢 Question — race-safety of the `setattr(cls, sentinel, True)` write** (`utils.py:48`)

The docstring explicitly says "not thread-safe by design (Python GIL provides good enough)". Under the free-threaded build (PEP 703, no-GIL Python 3.13t / 3.14t experimental), two threads could both pass the `cls.__dict__.get` check and both call `setattr` + `logger.warning`. Result is at most a duplicate warning (idempotent for warning-emit purposes). Worth a small note in the docstring like "(under no-GIL builds, may emit twice — still bounded by N threads)" so future readers don't assume strict-once. Not blocking.

### Edge cases considered

- **Helper called from `__init__`-time code on a Live class with `__slots__`**: `setattr(cls, sentinel, True)` writes to the CLASS, not the instance, so `__slots__` on instances doesn't apply. ✓
- **Helper called concurrently from two threads on first occurrence**: see 🟢 above — at most a duplicate warning, no exception, no incorrect state. ✓
- **HVR module reload**: docstring claims "sentinel survives because the class object outlives the module reload." Verified mental-model: HVR rebinds the module-level name to a new class object; old instances pointing at old class still see the old sentinel. New instances of the reloaded class start fresh and re-warn. This is a behavioral change from the deleted `_TRUNCATION_WARNED: set` (which was module-global and survived reload of the class but was tied to the websocket module). Tradeoff is acceptable — re-warning on reload is informative.
- **Subclass sentinel under MRO with diamond inheritance**: `cls.__dict__.get` returns False on diamond children even if both parents warned. Each diamond child gets its own warning. ✓ (This is the intentional behavior per the comment at `utils.py:44-45`.)
- **`type(view_instance)` returning a metaclass-instrumented class**: `setattr` on a class object goes through the metaclass `__setattr__`. Custom metaclasses that block writes would break this; none exist in djust today. Not a concern in practice.

### Test counts

- New test cases: 0 (the helper is exercised through the existing 10-test snapshot-truncation file).
- Tests modified for the refactor: `python/tests/test_snapshot_truncation_warning.py` (clear-mechanism replaced).
- Local run: `10 passed in 0.04s`.

### Verdict

The PR is well-scoped, the helper extraction is clean (correctly chooses `cls.__dict__.get` over `getattr`, with an inline comment explaining why), the canon additions all carry issue+retro citations, and the disconfirming-citation rule was implicitly dogfooded successfully on this very PR. Two minor 🟡 findings (citation drift + unused import in example) and one 🟢 (race-safety doc note) — none blocking.

STAGE_11_VERDICT: APPROVE

Findings: 0 🔴 / 2 🟡 / 1 🟢

johnrtipton:
# Retrospective — PR #1399 (canon batch + helper, closes 8 issues)

**Quality**: 4.5/5 (one orphan-import surfaced post-commit; two Stage 11 nits fixed in-band)
**Wall-clock**: ~30 min from /pipeline-run resume to merge

## What worked well

- **Bundled 8 retro-filed canon items into one PR.** This kept the canon-doc batch coherent rather than fragmenting across 8 PRs. Reviewer-ergonomics were good — one Stage 11 review pass covered all 8.
- **Action #122 (post-commit verify) caught the orphan import.** The `_TRUNCATION_WARNED` test-side import wasn't grepped during the implementer's refactor; pre-push hook tripped on it. The verify chain (`&& git log -1 --oneline` + `git push` failure on pytest collection) provided a clean signal — fixed in commit `a01f8995` with a per-class sentinel reset helper.
- **Disconfirming-citation canon (#1386) demonstrated its own value.** Stage 11 reviewer (per the new #1386 rule) tried to disprove the "subclasses get their own sentinel" claim by inspecting `cls.__dict__.get` vs `getattr`. The claim survived, which is the correct outcome — but the discipline is the point. Exactly the empirical pattern this PR canonicalizes.
- **Two-commit shape held across 4 commits** (impl + test fix + docs + nit fixes). Programmatic Gates 1 + 2 caught no violations.

- **Implementer missed the orphan test-side import on #1392 refactor.** When extracting a helper that removes a module-level symbol, the implementer should grep the codebase for `_TRUNCATION_WARNED` (or whatever symbol is being deprecated). Action #1391 (filter-migration grep canon) — landed in this very PR — applies symmetrically to symbol removals: grep before declaring the refactor done. Could canonicalize as "symbol-removal grep canon" but the existing #1391 already covers the abstract pattern.
- **Two Stage 11 🟡 findings (citation drift to live_view.py:517 vs :518; auth-guide unused-import).** Both fixed in-band. The drift is exactly the kind of citation-discipline #1197 already names; reviewer caught it correctly.

## What to improve

- **Action item (defer)**: pre-commit `grep` lint that catches obviously-removed symbols still referenced from elsewhere in the tree. The Python compiler doesn't catch this — Python imports are resolved at runtime. A simple grep-vs-actual-defined check would have flagged this 30s before push instead of at the pre-push hook.
- **Strengthen the implementer subagent prompt** for refactor-class tasks to explicitly require: "grep for the OLD symbol across `tests/` and `python/tests/` before declaring done." Existing #1391 canon covers this for filter expressions; explicit checklist for symbol removals would be a small extension.

The fix for #1392's orphan-import friction would extend #1391 — but #1391 just landed and is canon enough. The implementer-prompt update is a v0.9.5-3 follow-up if needed.

## Review stats

- Tests added: 0 (canon docs); test fix to existing 10
- 🔴 Findings: 0
- 🟡 Findings: 2 (Stage 11; both addressed in-band via commit `3ce296c6`)
- 💚 Findings: 1 (race-safety doc note; deferred)
- CI failures: 1 pre-push hook on `_TRUNCATION_WARNED` orphan (caught + fixed before remote landed; this is the canon working as designed)

## Connection to milestone

WU5 of v0.9.5-2 drain — final work unit. All 14 in-repo issues now closed. Milestone v0.9.5-2 complete; ready for retro.

Issue #1356, opened 2026-05-05
tech-debt: `InMemoryStateBackend.get_and_update()` returns shared reference (dead code, but a footgun)


johnrtipton:
Source: PR #1355 Stage 13 Re-Review (non-blocking finding #1).

## Problem

`python/djust/state_backends/memory.py:200` defines `get_and_update()` which returns a shared Python reference to the cached object. PR #1355 fixed `get()` to do a serialize/deserialize clone (closes #1353), but `get_and_update()` was overlooked.

`get_and_update()` currently has zero callers (`grep -r get_and_update python/` returns the definition + tests only — no production callsites). It's dead code, so the bug doesn't manifest. But if a future caller is added without auditing, the #1353 race class returns.

## Suggested fix (in preference order)

1. **Delete** `get_and_update()` if it's confirmed dead. `git log -p -- python/djust/state_backends/memory.py | grep -A5 'get_and_update'` shows when/why it was added; if no caller has materialized, drop it.
2. **Make it clone too** — apply the same `serialize_msgpack` → `deserialize_msgpack` round-trip as `get()` so the contract matches.
3. **Document the contract** — at minimum, add a docstring noting that the returned object is shared and must NOT be mutated by callers.

(1) is cleanest. (2) is consistent with `get()`. (3) is forgivable but trips the same canon as #1353 ("never share mutable state").

## Reference
- PR #1355 (closes #1353 + #1354) — Stage 13 Re-Review identified this as a deferred finding.
- Related: `python/djust/state_backends/memory.py:107-131` (`get()` — already fixed).

PR #1402, opened 2026-05-07
fix(templates): Node::Include round-trip no longer double-quotes path (#1396)

### LINKED ISSUE #1396: include round-trip double-quotes template path (uncovered by #1388)

johnrtipton:
## Summary

Surfaced during PR #1397's conversion of `test_nodes_to_template_string_include` to drive from parser output (#1388). Action #158 working as designed: the conversion immediately exposed a real bug masked by hand-built fixtures.

## What changed

- **Parser-side fix** (`crates/djust_templates/src/parser.rs:585`) — `Node::Include.template` now strips outer quotes at parse time, matching the existing contract for `Node::Extends`, `Node::Static`, and `Node::Now`. Single source of truth.
- **Emitter** (`crates/djust_templates/src/inheritance.rs:422`) — unchanged.
- **Tests** — un-ignored `test_nodes_to_template_string_include` (was marked `#[ignore]` in PR #1397 with pointer here); added `test_nodes_to_template_string_now` for defense-in-depth.

## Audit results

- `Node::Static.path` — already correctly strips quotes at `parser.rs:633` ✅
- `Node::Now.format` — already correctly strips quotes at `parser.rs:805` ✅
- `Node::Include.template` — was the outlier; now fixed ✅

## Closes

- Closes #1396

## Test plan

- [x] RED→GREEN: pre-fix the include test failed with `left: "{% include \"\"partials/header.html\"\" %}" right: "{% include \"partials/header.html\" %}"`; post-fix passes
- [x] Rust workspace — 745 → 747 active passing (1 → 0 ignored, +1 new Now test)
- [x] djust_templates: 284/284 green
- [x] Python pytest — 4743 passed, 14 skipped (unchanged)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- **`Node::Include` round-trip no longer double-quotes the template path (#1396).** Parser was preserving the outer quotes on `Include.template`; emitter `nodes_to_template_string` then wrapped again, producing `{% include ""partials/header.html"" %}` on round-trip. Surfaced during PR #1397's conversion of round-trip tests to drive from parser output (Action #158 working as designed). Fixed by aligning the parser to strip outer quotes (matching the existing `Extends`, `Static`, and `Now` contracts) — single source of truth, emitter unchanged. `test_nodes_to_template_string_include` un-ignored. Added `test_nodes_to_template_string_now` for defense-in-depth.
+
 ### Security
 
 - **`sanitize_for_log` cache_key on HTTP cache-lookup debug log (#1368).** Pre-existing log-injection asymmetry between WebSocket and HTTP paths in `python/djust/mixins/rust_bridge.py`: the WS site at line 333 sanitized correctly; the HTTP site at line 363 did not. Since `cache_key` derives from `request.path` (user-controlled), an attacker-supplied path like `/page/\n[FAKE LOG ENTRY]` could inject newlines into the log stream. Mirrors the WS-path call to `sanitize_for_log(self._cache_key)` and adds the matching CodeQL annotation. Surfaced in PR #1367 Stage 11 SHOULD-FIX #3 (deferred per Action #1079).
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -294,7 +294,7 @@ dependencies = [
 
 [[package]]
 name = "djust_components"
-version = "0.9.5-rc.1"
+version = "0.9.5-rc.2"
 dependencies = [
  "ahash",
  "criterion",
@@ -309,7 +309,7 @@ dependencies = [
 
 [[package]]
 name = "djust_core"
-version = "0.9.5-rc.1"
+version = "0.9.5-rc.2"
 dependencies = [
  "ahash",
  "criterion",
@@ -324,7 +324,7 @@ dependencies = [
 
 [[package]]
 name = "djust_live"
-version = "0.9.5-rc.1"
+version = "0.9.5-rc.2"
 dependencies = [
  "bincode",
  "criterion",
@@ -349,7 +349,7 @@ dependencies = [
 
 [[package]]
 name = "djust_templates"
-version = "0.9.5-rc.1"
+version = "0.9.5-rc.2"
 dependencies = [
  "ahash",
  "chrono",
@@ -370,7 +370,7 @@ dependencies = [
 
 [[package]]
 name = "djust_vdom"
-version = "0.9.5-rc.1"
+version = "0.9.5-rc.2"
 dependencies = [
  "ahash",
  "criterion",
--- a/crates/djust_templates/src/inheritance.rs
+++ b/crates/djust_templates/src/inheritance.rs
@@ -805,12 +805,23 @@ mod tests {
     }
 
     #[test]
-    #[ignore = "real bug uncovered by #1388 — `Node::Include`'s `template` field is parser-stored with surrounding quotes; emitter double-wraps. Tracked as #1396 (out-of-scope per #1079 broader-sweep canon)."]
     fn test_nodes_to_template_string_include() {
-        // Drive from parser output (#1388). Currently fails — see #1396.
+        // Drive from parser output (#1388). Regression for #1396 — parser
+        // stored Include.template with surrounding quotes, emitter wrapped
+        // them in another pair, producing {% include ""partials/header.html"" %}.
         let nodes = parse_source("{% include \"partials/header.html\" %}");
         let result = nodes_to_template_string(&nodes);
         assert_eq!(result, "{% include \"partials/header.html\" %}");
+    }
+
+    #[test]
+    fn test_nodes_to_template_string_now() {
+        // Drive from parser output (#1388). Lock in Now round-trip
+        // (audited under #1396 — Now strips quotes correctly at parse,
+        // no double-wrap; defense-in-depth test against future regression).
+        let nodes = parse_source("{% now \"Y-m-d\" %}");
+        let result = nodes_to_template_string(&nodes);
+        assert_eq!(result, "{% now \"Y-m-d\" %}");
     }
 
     #[test]
--- a/crates/djust_templates/src/parser.rs
+++ b/crates/djust_templates/src/parser.rs
@@ -582,7 +582,12 @@ fn parse_token(tokens: &[Token], i: &mut usize) -> Result<Option<Node>> {
                             "Include tag requires a template name".to_string(),
                         ));
                     }
-                    let template = args[0].clone();
+                    // Strip surrounding quotes (#1396) so Include.template
+                    // shares the unquoted-field contract with Extends/Static/Now.
+                    // Without this strip, the inheritance emitter
+                    // (`nodes_to_template_string`) double-wraps the value,
+                    // producing `{% include ""x.html"" %}` on round-trip.
+                    let template = args[0].trim_matches(|c| c == '"' || c == '\'').to_string();
                     let mut with_vars = Vec::new();
                     let mut only = false;
 


johnrtipton (commented):
## Stage 11 Code Review — PR #1402

**Verdict: APPROVE**

Parser-side fix that aligns `Node::Include.template` with the unquoted-field
contract already used by `Node::Extends`, `Node::Static`, and `Node::Now`.
Strategy choice (parser-side, not emitter-side) is correct: it produces a
single source of truth, leaves the emitter at `inheritance.rs:422` unchanged,
and turns the renderer's defensive `trim_matches` at `renderer.rs:655` into
a safe no-op rather than a contract-shift hazard. Two-commit shape, clean
disconfirming audit, parser-driven tests.

### Strategy verification (parser-side vs emitter-side)

Confirmed by inspection that all four quote-stripping sites now use the
identical pattern `args[0].trim_matches(|c| c == '"' || c == '\'').to_string()`:

- `Extends`: `parser.rs:575`
- `Include`: `parser.rs:590` (this PR)
- `Static`:  `parser.rs:638`
- `Now`:     `parser.rs:810`

Picking parser-side is the right move because the alternative (emitter-side
`.trim_matches` at `inheritance.rs:422`) would have made `Include.template`
an outlier where the field's contract differed from siblings — exactly the
mismatch that produced `{% include ""x.html"" %}` in the first place.

### Disconfirming citation (per Action #1386)

Audit claim was "Static and Now are already correct." Verified by reading
both sites:

- `parser.rs:638` (`Static`): `let path = args[0].trim_matches(|c| c == '"' || c == '\'').to_string();` ✅
- `parser.rs:810` (`Now`):    `let format = args[0].trim_matches(|c| c == '"' || c == '\'').to_string();` ✅

Cross-checked against their emitters (`inheritance.rs:436` `Node::Static`,
`inheritance.rs:494` `Node::Now`) — both wrap the field in literal quotes.
If parsers preserved quotes they would double-wrap exactly like Include did.
The fact that they don't double-wrap in production is empirical evidence the
parsers strip. Audit holds.

### Renderer & call-site impact

Single `Node::Include { ... }` construction site in the workspace
(`parser.rs:618`); confirmed via `grep -rn "Node::Include" crates/ python/`.
The other matches are all variant tags (`matches!`, destructure-only) plus a
test fixture (`parser.rs:2756`) that already uses unquoted form. So:

- `renderer.rs:655` (defensive `trim_matches`): becomes a no-op for the
  parser-fed path. Stays correct for any hypothetical hand-built AST that
  passes a quoted path. **No regression risk.**
- `parser.rs:1163` and `parser.rs:1426` (deps-extraction): consume only the
  variant tag, never `.template`. **No regression risk.**
- No PyO3 surface exposes `Node::Include.template` — `grep -rn "pyclass.*Include"`
  is empty. Internal-only AST.

### Test quality

Both tests drive through real `tokenize` + `parse`:

```rust
fn parse_source(source: &str) -> Vec<Node> {
    let tokens = crate::lexer::tokenize(source).unwrap();
    crate::parser::parse(&tokens).unwrap()
}
```

(`inheritance.rs:617`)

- `test_nodes_to_template_string_include` (un-ignored): asserts
  `{% include "partials/header.html" %}` round-trip. If parser reverted to
  `args[0].clone()`, parser stores `"\"partials/header.html\""`, emitter
  wraps producing `{% include ""partials/header.html"" %}`. Test fails. ✅
- `test_nodes_to_template_string_now` (new): same shape with `{% now "Y-m-d" %}`.
  If `Now`'s parser ever regressed, emitter would produce `{% now ""Y-m-d"" %}`
  and the test fails. ✅

Neither test is a tautology — both assert on the exact bytes the bug
produced.

Local run: 13/13 `test_nodes_to_template_string_*` green; 0 ignored
(was 12 + 1 ignored). Matches PR description's "+1 new, +1 un-ignored".

### CHANGELOG & commit shape

- CHANGELOG entry under `### Fixed`, cites #1396, names the parser-side
  strategy explicitly, references Action #158 lineage. ✅
- Two-commit shape per Action #181: c1 `f960fe0b` impl + tests + Cargo.lock,
  c2 `d831a334` CHANGELOG only. ✅
- Cargo.lock drift (rc.1 → rc.2) is workspace-version drift from the clippy
  hook, not a content change; commit message calls this out explicitly.

### Edge cases considered

1. **Empty string arg** — `trim_matches` on empty string returns empty
   string, no panic. Unchanged from existing `Extends`/`Static`/`Now`
   behavior.
2. **Mixed quotes** (`'foo"`)— `trim_matches` with the `|c| c == '"' || c == '\''`
   closure strips ALL leading/trailing quote characters of either kind.
   Same behavior as siblings; not a regression vector.
3. **Already-unquoted input** (e.g. `{% include foo.html %}` if lexer
   passes it through) — `trim_matches` is a no-op. Idempotent.
4. **`with_vars` values** — those flow through `args[i].splitn(2, '=')`
   at `parser.rs:602` without quote-stripping. Out of scope for #1396 but
   worth noting if a future Include round-trip test covers
   `{% include "x" with name="value" %}`. Not flagged as a finding.

### Question / concern (non-blocking)

The PR description's "single source of truth" framing is accurate for the
4-site parser audit. However, the renderer's `trim_matches` at
`renderer.rs:655` is now strictly belt-and-braces — it can never fire on
parser-fed input. Two options for a follow-up:

(a) Leave it (current PR's choice) — defends against any future hand-built
    `Node::Include` AST.
(b) Remove it and add a debug_assert that the value is unquoted, surfacing
    the contract.

Either is fine; (a) is conservative and out of scope for this fix. Filing
this as a sub-1 KB observation, not a request for change.

### Findings

- 🔴 MUST-FIX: 0
- 🟡 SHOULD-FIX: 0
- 🟢 NIT/observation: 1 (renderer defensive trim is now belt-and-braces; non-blocking)

STAGE_11_VERDICT: APPROVE

johnrtipton:
# Retrospective — PR #1402 (closes #1396)

**Quality**: 5/5
**Pipeline shape**: bugfix, 14 stages, all passed
**Wall-clock**: ~15 min from /pipeline-run resume to merge

## What worked well

- **Parser-side strategy was the right call.** The bug had two surface options (parser-side strip OR emitter-side detect-already-quoted). Parser-side made `Include.template` consistent with `Extends`, `Static`, `Now` — single source of truth. Audit confirmed Static and Now were already correct, so the asymmetry was the bug; aligning the outlier (Include) was clean.
- **PR #1397's `#[ignore]` pointer paid off.** When a follow-up PR un-ignores a previously-marked test, the marker comment + linked issue is the entire context anyone needs. PR #1397's deferred-with-pointer pattern (Action #1079 broader-sweep canon) executed cleanly.
- **Disconfirming-citation discipline (#1386) held**: Stage 11 reviewer re-read parser.rs:638 (Static) and :810 (Now) to actively try to disprove "they're already correct." Both confirmed. The `_djust_warned_*` style audit-with-grep is becoming a Stage 11 reflex.

- None. Bug was well-scoped from #1396's body; reproducer was already in the test file (just `#[ignore]`'d).

## What to improve

- The `renderer.rs:655` defensive `trim_matches` is now belt-and-braces (parser strips, then renderer would strip again — no-op for parser-fed input). Could remove for clarity, but it stays safe for any hand-built `Node::Include` from non-parser sources. Filed as forward-looking 🟢 in Stage 11; not actionable today.

None — Action #158 (round-trip tests from parser output), Action #1079 (broader-sweep scope discipline), Action #1386 (disconfirming citation), Action #181 (two-commit shape) all applied directly.

## Review stats

- Tests added: 1 (`test_nodes_to_template_string_now`); 1 un-ignored (`test_nodes_to_template_string_include`)
- 🔴 Findings: 0
- 🟡 Findings: 0
- 💚 Findings: 1 (renderer defensive trim — non-blocking)
- CI failures: 0

## Connection to milestone

WU2 of v0.9.5-3 drain. Next: WU3 (#1356 `InMemoryStateBackend.get_and_update` shared-ref).

Issue #1400, opened 2026-05-06
tech-debt: extend filter-migration grep canon (#1391) to cover symbol removals during refactor


johnrtipton:
**Source**: Retro v0.9.5-2 finding #2

PR #1399's #1392 helper extraction removed the module-global `_TRUNCATION_WARNED: set` from `python/djust/websocket.py`. The implementer didn't grep `tests/` and `python/tests/` for residual imports of the removed symbol. Pre-push hook caught the orphan import at `python/tests/test_snapshot_truncation_warning.py:15` (Action #122 working as designed). Fix landed as in-band commit `a01f8995`.

The existing #1391 filter-migration grep canon ("when changing a filter convention, grep the codebase for ALL call sites that use the OLD convention") abstracts the right principle, but its specific wording names "filter expressions" — the canon should generalize to ANY symbol removal.

## Proposed scope

Extend the #1391 wording in `CLAUDE.md` "Process canon" section to cover both classes:

- Filter convention changes (existing)
- Symbol removals during refactor (NEW): when removing a module-level constant, function, or class as part of a refactor (e.g., extracting a helper that subsumes its role), grep for the OLD symbol name across `tests/` and `python/tests/` (and any other test/example directory) before declaring the refactor done. Python imports are resolved at runtime — the compiler doesn't catch orphan references.

Concrete check during Stage 5 implementation: for any PR that removes a top-level `_FOO_BAR` symbol or `def _foo_bar(...)`, grep `_FOO_BAR\|_foo_bar` across the repo and verify each hit is updated or removed.

## Out of scope

- Pre-commit lint that automates the grep (a different, larger ask — would need an AST-aware tool that knows what's "removed in this diff" vs "always missing").

## Reference

Retro v0.9.5-2 finding #2; PR #1399 commit `a01f8995` (the in-band fix).

PR #1404, opened 2026-05-07
feat(static): JS micro-cleanup — dedupe transition helpers + tighten routeMap (#1360, #1361)

### LINKED ISSUE #1360: tech-debt: deduplicate `_parseTimeMs` and `_computeTransitionTiming` between dj-transition and dj-remove modules
Source: PR #1359 Stage 11 Review (deferred 🟡 finding); CodeQL alert surfaced by the bundle rebuild.

## Problem

PR #1357 introduced auto-derivation of CSS `transition-duration` for both `dj-transition` and `dj-remove`. The implementation defines `_parseTimeMs()` and `_computeTransitionTiming()` in BOTH:

- `python/djust/static/djust/src/41-dj-transition.js`
- `python/djust/static/djust/src/42-dj-remove.js`

Because the bundle (`python/djust/static/djust/client.js`) is built by concatenating source modules, the bundle has duplicate top-level function declarations. CodeQL flags these:

```
client.js:13162 Conflicting function declarations — Declaration of function _parseTimeMs conflicts with [another declaration] in the same scope.
client.js:13171 Conflicting function declarations — Declaration of function _computeTransitionTiming conflicts with [another declaration] in the same scope.
```

Runtime impact: in non-strict mode JS allows duplicate function declarations (second wins). The bundle works because both definitions are functionally identical. But the duplication is fragile — a future divergent edit to one copy would silently use whichever gets concatenated last.

## Why surfaced now

PR #1357 deliberately deferred the bundle rebuild per the original commit message:
> Source-only commit; bundle (`client.js`) is intentionally NOT updated here. The bundled file has 392 pre-existing eslint warnings (issue #1351) that block `--max-warnings 0`.

PR #1359 closes #1351 and rebuilds the bundle for the first time, exposing the dup.

## Suggested fix

1. **Move the helpers to a shared earlier-loaded module** — e.g., `01-utils.js` or a new `40-transition-helpers.js` that loads before both 41 and 42. Both modules call the helpers without redefining.
2. Or: define in `41-dj-transition.js` only, document in `42-dj-remove.js` that it relies on the bundle ordering having `41-` loaded first (fragile).
3. Or: scope each file with an IIFE so top-level declarations are file-local (heavyweight; would also require exposing the helpers via a global namespace).

(1) is cleanest. The fix is small (~40 lines moved + 2 lines deleted in 42).

## References
- PR #1357 (introduced the duplicate)
- PR #1359 (rebuilt bundle, surfaced via CodeQL)
- `client.js:13162, 13171` (CodeQL alert locations)

### LINKED ISSUE #1361: tech-debt: tighten routeMap[pathname] access with hasOwnProperty.call in 18-navigation.js
Source: PR #1359 Stage 11 Review (deferred 🟡). `18-navigation.js:163-166` does `routeMap[pathname]` where `pathname` is user-controllable. Currently safe-by-accident via downstream server validation; reviewer recommends `Object.prototype.hasOwnProperty.call(routeMap, pathname)` guard or converting `routeMap` to `new Map([...])` (prototype-pollution-immune by design). PR #1359 added eslint-disable-next-line with rationale; this issue tracks tightening the actual access.

johnrtipton:
## Summary

Two follow-ups deferred from PR #1359 Stage 11.

### #1360 — deduplicate `_parseTimeMs` + `_computeTransitionTiming`

PR #1357 introduced auto-derivation of CSS `transition-duration` for both `dj-transition` and `dj-remove`. Both modules defined the helpers at top-level; bundle concat produced duplicate function declarations and CodeQL flagged `client.js:13162` and `:13171`.

**Fix**: extract helpers into a new shared `40a-transition-helpers.js`. Sorts lexicographically between `40-dj-layout.js` and `41-dj-transition.js`/`42-dj-remove.js`, matches the build glob, single-purpose. Bundle now has exactly one definition each.

### #1361 — tighten `routeMap[pathname]` access

`18-navigation.js:163-166` did `routeMap[pathname]` with `pathname` user-controlled. PR #1359 added an `eslint-disable-next-line` with rationale; this fix removes it.

**Strategy**: neither A (`hasOwnProperty.call` guard) nor B (Map conversion). Used `Object.entries(routeMap)` walk:
- Prototype-pollution-immune by construction (own enumerable string-keyed entries only).
- Lints cleanly without disables.
- B was rejected because converting to `Map` would change the wire-protocol shape emitted by `python/djust/routing.py` and break ~6 test callers + downstream user code that sets `_routeMap` directly.

## Closes

- Closes #1360
- Closes #1361

## Test plan

- [x] vitest — 55 passed (navigation 18, transition + remove + transition-group 37) + 60 supplementary (sw_advanced, debug_panel_harness) + 2 (bundle-init-no-tdz)
- [x] Full pytest — 4743 passed, 14 skipped (no regression)
- [x] ESLint on changed source files — 0 errors / 0 warnings
- [x] Bundle rebuild verifies single function definition: `grep -c '^function _parseTimeMs' client.js` = 1 (was 2)


johnrtipton (commented):
## Stage 11 Code Review — PR #1404 (JS micro-cleanup, closes #1360 + #1361)

Two follow-ups deferred from PR #1359 Stage 11. Reviewed both fixes against canon (Actions #168 first-principles grep, #1046 doc-claim-verbatim TDD, #1196 delegated-listener integration tests, #1199 edge-case coverage) and against the disconfirming-citation rule (#1386).

### Verifications run

**#1360 dedupe — PASS**

- `python/djust/static/djust/src/40a-transition-helpers.js:20,29` defines `_parseTimeMs` + `_computeTransitionTiming`.
- `python/djust/static/djust/src/41-dj-transition.js:64-65` and `42-dj-remove.js:48-49` are now comment-only stubs pointing at the canonical home.
- Bundle dedup verified locally: `grep -c '^function _parseTimeMs' python/djust/static/djust/client.js` returns `1` (was `2`); same for `_computeTransitionTiming`. Only definition is at `client.js:13461,13470`.
- Source-of-truth has the helpers exactly once across the `src/` tree (`grep -n '^function _parseTimeMs\|^function _computeTransitionTiming' src/*.js` returns only `40a-transition-helpers.js`).

**Disconfirming-citation check on bundle order (#1386 discipline) — PASS**

- Build script `scripts/build-client.sh:74` does `cat "$SRC_DIR"/[0-9]*.js > "$CLIENT_OUT"`. Glob is `[0-9]*.js`. Filename `40a-transition-helpers.js` starts with digit `4`, so the glob matches.
- Lex sort of the `src/` tree (verified via `ls`) places `40a-transition-helpers.js` immediately between `40-dj-layout.js` and `41-dj-transition.js`. Concat order is correct: helpers defined before consumers.
- The 40a-prefix scheme is documented in the new file's header comment (lines 13-16) so future maintainers won't misread it as alphabetical-only.

**#1361 Object.entries walk — PASS, with one observation**

- `18-navigation.js:166-168` — `for (const [routePath, viewPath] of Object.entries(routeMap)) { if (routePath === pathname) return viewPath; }`. Correct shape: walks own enumerable string-keyed entries; no `routeMap[pathname]` indirection.
- The `eslint-disable-next-line security/detect-object-injection` previously above `if (routeMap[pathname])` is removed.
- `46-state-snapshot.js:41-43` — same shape applied. eslint-disable removed at the matching site.
- Wire-protocol shape preserved: `python/djust/routing.py:203-217` emits `window.djust._routeMap = <JSON object>`. `Object.entries` works correctly on plain objects, so no server-side change needed. Option-B rejection rationale (Map conversion would change wire-protocol) verified.

**#1361 — eslint-disables that legitimately stay**

Two eslint-disables remain in `18-navigation.js` (lines 53, 80): both wrap `window.history[method]` where `method` is a `'pushState' | 'replaceState'` literal selected from the `data.replace` boolean — controlled-enum dispatch, out of scope of #1361. One eslint-disable remains in `46-state-snapshot.js:61` for `bag[slug]` where `slug` is a server-side-resolved view slug (not user-controllable URL pathname); also out of scope. These are correctly NOT removed.

**Bundle artifacts — PASS**

`git show --stat fc307a8b` shows all four artifacts touched in the impl commit:
- `client.js` (+68/-71)
- `client.min.js` (+1/-1)
- `client.min.js.br` (44496 → 44462 bytes)
- `client.min.js.gz` (52882 → 52876 bytes)
- `client.min.js.map` (+1/-1)

Bundle dropped 6-34 bytes across compressed forms (helper dedup more than offset by the `Object.entries` walk's slightly larger codegen).

**Two-commit shape — PASS**

- c1 `fc307a8b` impl + bundle (10 files, 138/-144 lines)
- c2 `98a10094` CHANGELOG only (1 file, 6/-0 lines)

Matches Action #181 canon (Stage 5 forbids CHANGELOG; Stage 9 is the canonical CHANGELOG commit boundary).

**CHANGELOG — PASS**

`CHANGELOG.md:9-13` under `### Changed`, cites both #1360 + #1361, references the option-B rejection (Map conversion → wire-protocol change). Two-bullet structure is appropriate; nested bullets describe each fix's mechanism succinctly.

**Test counts**

PR body claims 55 vitest + 60 supplementary + 2 bundle-init = 117 passing. CI is still IN_PROGRESS at review time (`Analyze JavaScript`, `javascript-tests`, `playwright-tests` running) — would normally hold approval pending the CodeQL `Analyze JavaScript` job result so we can confirm the `client.js:13162` and `:13171` alerts clear in the actual pipeline (PR claim is well-supported by the local dedup grep, but CodeQL is the load-bearing oracle here per #1360's exit criteria).

### Findings

#### 🟡 Yellow #1 — security claim ("prototype-pollution-immune by construction") not regression-tested

The PR's strongest correctness claim is that the `Object.entries` walk is prototype-pollution-immune. Existing `tests/js/navigation.test.js:157-200` covers the happy path (exact match, `:id` pattern, fallback) but **no test exercises the prototype-pollution scenario the rewrite was designed to defeat**. By Action #1046 (doc-claim-verbatim TDD: "for each documented rule, point to the asserting test") the CHANGELOG bullet "prototype-pollution-immune by construction" is currently unbacked.

A 4-line test would lock it in:

```js
it('does not return Object.prototype properties', () => {
    const { window } = createEnv();
    window.djust._routeMap = { '/items/': 'myapp.views.ItemListView' };
    // toString, hasOwnProperty, __proto__ are all on Object.prototype.
    // The old `routeMap[pathname]` shape would return Object.prototype.toString here.
    expect(window.djust.navigation.resolveViewPath('toString')).toBeNull();
    expect(window.djust.navigation.resolveViewPath('hasOwnProperty')).toBeNull();
});
```

(Same shape for `46-state-snapshot.js:_currentViewSlug` — return value should be `''`, not the inherited prototype property name.)

This is the kind of "tautology test" detection inverse from Action #1200 — without this test, a future regression that re-introduces `routeMap[pathname]` indexing would NOT fail any existing test (the happy-path tests would still pass). Recommend adding before merge or filing a follow-up issue if the team prefers to ship this PR small.

#### 🟡 Yellow #2 — performance characteristic of `Object.entries` walk vs. the indexed access it replaces

The change converts O(1) hash lookup → O(n) linear walk on every navigation. For typical apps with <100 routes this is negligible (~microseconds), but the change is unconditional and has no fast-path for the exact-match case. Two observations:

1. The pattern-match loop already walks `Object.entries(routeMap)` (lines 171-188), so the exact-match walk is short-circuiting BEFORE the pattern loop runs anyway. This is fine — and arguably the rewrite is now *closer* in shape to the surrounding code.
2. For very large routeMaps (hypothetical PaaS multi-tenant with thousands of routes), the user could see measurable regression. **No new benchmark test asserts the performance is acceptable.** Action #168 (first-principles grep before architecting) would have surfaced that the file already does an `Object.entries` walk for pattern matching, making the rewrite consistent — worth calling out in the PR body so future readers don't re-litigate the perf concern.

#### 🟢 Green #1 — file-naming scheme is documented in-place

`40a-transition-helpers.js:13-16` contains explicit prose explaining the `40a-` lex-sort scheme. Good defensive doc — a future maintainer reading just this file understands why the prefix isn't `40b-` or `transition-helpers.js`.

#### 🟢 Green #2 — comments left in 41/42 point at the canonical home

`41-dj-transition.js:64-65` and `42-dj-remove.js:48-49` left a "lives in 40a-transition-helpers.js" comment instead of silently deleting the helper definitions. Future grep for `_parseTimeMs` from inside either consumer file finds the breadcrumb without needing to read all of `src/`.

### Required artifacts (per Stage 11 rubric)

- **Specific line citations**: `40a-transition-helpers.js:20,29` (canonical defs); `18-navigation.js:166-168` (Object.entries walk); `46-state-snapshot.js:41-43` (same shape); `routing.py:203-217` (wire-protocol shape preserved); `client.js:13461,13470` (single bundle definitions); build glob at `scripts/build-client.sh:74`.
- **Question/concern**: see Yellow #1 — the security claim needs a regression test. Optional but recommended: also surface the `Object.entries` walk's perf characteristic in the CHANGELOG so it's findable when someone benchmarks navigation later.
- **Edge cases considered**: `__proto__`/`toString`/`hasOwnProperty` lookup (current code is correct but untested); empty routeMap (existing test covers); pattern-match path (existing test covers); same change applied to `46-state-snapshot.js` (verified; same shape).

### Stage 11 verdict

The fixes are correct and the shape choices (option-C `Object.entries` walk over option-A `hasOwnProperty.call` over option-B Map conversion) are justified. Bundle dedup verified locally; CodeQL alerts at `client.js:13162,:13171` should clear (CI not yet completed). Two-commit shape correct; CHANGELOG entry well-formed.

The only material gap is Yellow #1 — the regression test for the prototype-pollution claim. Approving with the request that either (a) the test is added in this PR before merge, or (b) a follow-up issue is filed referencing this finding so the gap doesn't get lost.

**STAGE_11_VERDICT: APPROVE**

Findings: 0 🔴 / 2 🟡 / 2 🟢

johnrtipton:
# Retrospective — PR #1404 (closes #1360 + #1361)

**Quality**: 4.5/5 (one substantive 🟡 from Stage 11 required an in-band fix)
**Wall-clock**: ~25 min from /pipeline-run resume to merge (incl. CI re-run after fix)

## What worked well

- **Clean module-extraction pattern.** Hoisting `_parseTimeMs` + `_computeTransitionTiming` into a new `40a-transition-helpers.js` worked cleanly because the build glob `[0-9]*.js` lex-sorts that filename between `40-` and `41-/42-`. Single canonical home, breadcrumb comments left in 41/42 pointing back. Bundle dedup verified via `grep -c '^function' client.js` = 1 each (was 2).
- **Strategy choice for #1361 (`Object.entries` walk over Map conversion).** Reviewer initially flagged Map (option B) as the cleanest answer, but the impl agent rejected it correctly because `Map` would force a wire-protocol change in `python/djust/routing.py` (currently emits `window.djust._routeMap = {...JSON...}`) and break ~6 test callers. The `Object.entries(routeMap)` walk is prototype-pollution-immune by construction AND lints cleanly without `eslint-disable`. The right architectural call.
- **Action #1391 / #1400 grep canon applied** to `routeMap[*]` access sites — sister site at `46-state-snapshot.js:41-43` got the same treatment in the same PR.

- **Missing regression test for the prototype-pollution-immune claim.** Stage 11 reviewer flagged this as a 🟡 per Action #1046 (doc-claim-verbatim TDD): the load-bearing security claim had zero failing test if reverted. Fixed in-band via commit `25b5b08a` — added 3 assertions (`'toString'` / `'constructor'` / `'hasOwnProperty'` must return null with empty routeMap). Without these, a future refactor to `routeMap[pathname]` would silently regress the hardening.
- **Pattern recap**: this is the SAME failure mode as v0.9.5-1c PR #1381's #1382 cycle test — Stage 5 wrote tests that exercised the happy path but didn't lock in the security-critical claim. Stage 11 caught it both times.

## What to improve

- **#1046 (doc-claim-verbatim TDD) deserves a Stage 5 addition**: when the implementation makes a *security*-class claim ("immune to X", "fail-closed for Y", "validates against Z"), Stage 5 must include a regression test that fails if the claim is reverted. Currently #1046 is a Stage 4 plan-template canon; Stage 5 is where the test gets written. Could file as a follow-up to extend the canon to Stage 5 as well.

None new — the relevant canons (#1391 filter-migration grep / sibling-symmetry, #1386 disconfirming citation, #1046 doc-claim TDD) all already exist. Pattern of "implementer writes happy-path test; Stage 11 adds the security/edge-case regression test" is well-known and held here.

## Review stats

- Tests added: 1 prototype-pollution regression (3 assertions) at `tests/js/navigation.test.js:201-213`
- 🔴 Findings: 0
- 🟡 Findings: 2 (one substantive — addressed in-band; one "doc the O(n) tradeoff" — non-blocking)
- 💚 Findings: 2
- CI failures: 0

## Connection to milestone

WU4 of v0.9.5-3 drain. Next: WU5 (#1372 bundle-init-order structural lint).

Issue #177, opened 2026-01-31
Debug Toolbar: Add event replay button


johnrtipton:
## Summary

Add a "Replay" button to each event in the Events tab history, allowing developers to re-trigger events for debugging.

## Desired Behavior

- Each event entry in the Events tab has a replay action
- Clicking replay re-sends the event through the existing WebSocket connection with the original params
- Replay only available for events that have valid params
- Users see inline feedback: pending while waiting, success or error when complete
- Status feedback auto-clears after a few seconds

## Why

When debugging event handlers, developers often need to re-trigger the same event repeatedly. Currently they must interact with the UI to reproduce events, which is tedious for events that require specific sequences to trigger.

PR #1422, opened 2026-05-07
feat(deploy): guided end-to-end onboarding — login, project-create, deploy in one command

johnrtipton:
## Summary

Makes `djust deploy` a single-command end-to-end onboarding flow. A brand-new user types `djust deploy` and gets walked through:

1. **Login** (if no creds) — email + password prompt, token saved
2. **Slug resolution** — explicit arg → `[tool.djust.deploy].project` in pyproject → prompt + offer to save
3. **Project precondition** — `GET /api/v1/projects/<slug>/`; on 404 → confirm + `POST /api/v1/projects/`
4. **Deploy** — same as before

Each step is **skipped if its precondition is already satisfied**, so power users running `djust deploy` repeatedly see only step 4.

**No new dependencies.** Built on click's existing prompts (`click.prompt`, `click.confirm`). The user asked specifically: do we need a fancy CLI framework? Answer: no.

## Two flags for non-interactive use

| Flag | Behaviour |
|---|---|
| `--yes` / `-y` | Auto-accept every confirmation. CI / scripted use. |
| `--no-create` | Fail-fast if the project doesn't exist server-side. Avoids surprise creates in scripts. |

## Server-side requirement

The new precondition + creation paths require three endpoints that don't exist on djustlive yet — they land in **[djustlive #177](https://github.com/djust-org/djustlive/pull/177)**:

- `GET /api/v1/me/`
- `GET /api/v1/projects/<slug>/`
- `POST /api/v1/projects/`

This djust release should NOT be tagged for distribution until djustlive #177 is deployed (otherwise users hit `djust deploy` and get cryptic errors when the new endpoints 404).

## Behaviour matrix

| State | What happens |
|---|---|
| no creds | prompt for email/password, save token |
| creds but `/me/` → 401 (revoked) | drop, re-prompt, re-save |
| no slug arg, no pyproject | prompt + offer to save (PR #1422 base commit, already in this branch) |
| slug + `/projects/<slug>/` → 404 | confirm + `POST /projects/`, offer to save corrected slug if server uniquified |
| `--yes` | auto-accept all confirmations |
| `--no-create` | fail-fast on missing project |
| all preconditions met | quiet — just deploy |

## Tests

`python/tests/test_deploy_cli.py`:

- **4 new** `TestGuidedDeploy` tests:
  - `test_yes_flag_creates_project_without_prompting`
  - `test_no_create_flag_fails_on_404`
  - `test_revoked_token_triggers_re_login`
  - `test_no_credentials_at_all_prompts_login`
- **5 existing** `TestDeployCommand` / `TestSlugResolution` tests updated to also mock the new precondition endpoints.
- **43/43** deploy_cli tests pass locally.

Plus this branch already carries the **slug-from-pyproject** feature + the **deploy-CLI error message fix** that were originally PR #1422's scope. They flow naturally as part of the guided experience and ship together.


johnrtipton:
## Independent Code Review (PR #1422)

Reviewed all 7 commits of `feat/deploy-slug-from-pyproject` end-to-end against `main`. Stages 3 + 4 cleared the major OAuth/PKCE security work; finding a few residual bugs and one CHANGELOG drift below.

### 🔴 Must-fix

**1. CHANGELOG-vs-code drift on the pyproject key.** `CHANGELOG.md:12` says `[tool.djust] project_slug = "…"`, but the code reads/writes `[tool.djust.deploy] project = "…"` (`python/djust/deploy_cli.py:191`, `:213`, `:231`). Users following the changelog to pre-seed pyproject would get the prompt anyway. Either update the changelog or change the docstring/key to match — they need to be one shape.

**2. Slug-collision update path produces an unparseable pyproject.toml.** When the server uniquifies the slug (e.g., `my-app` → `my-app-1`), `_ensure_project_exists` calls `_save_slug_to_pyproject` (`python/djust/deploy_cli.py:470`). But the file already had `[tool.djust.deploy] project = "my-app"` in it (that's how we got here in the first place). The append-only writer at `python/djust/deploy_cli.py:229-233` blindly appends a SECOND `[tool.djust.deploy]` block. TOML rejects duplicate tables — the next `tomllib.loads` raises `TOMLDecodeError` ("Cannot declare ('tool', 'djust', 'deploy') twice"). The function's docstring at `:217-219` says "we only run it when the key didn't exist; we're never duplicating" — but that invariant is violated by the slug-collision branch. Either:
  - parse-and-rewrite the existing block (preferred, even without `tomli_w`: a small regex over `^\[tool\.djust\.deploy\]\nproject = ".*"$` is safe enough for an auto-generated block), OR
  - skip the auto-update on the collision branch and just print "update pyproject.toml manually to '<new-slug>'", OR
  - guard `_save_slug_to_pyproject` against an existing `[tool.djust.deploy]` table and refuse to append in that case.

Test gap: `tests/test_deploy_cli.py` has no test for the slug-collision update path; add one that pre-seeds pyproject with one slug, mocks `POST /projects/` returning a different slug, and asserts the post-call pyproject is still parseable.

### 🟡 Should-fix

**3. `_resolve_project_slug` ignores `--no-create` and `--yes` flags.** `deploy` command at `python/djust/deploy_cli.py:1038` calls `_resolve_project_slug(project_slug, Path.cwd())` with no `interactive=` kwarg — defaults to `True`. So `djust deploy --no-create` (CI use) with no slug arg and no pyproject still tries to interactively prompt — same class of bug commit 252a4814 fixed for `_ensure_logged_in`. `deploy_dir` has the same shape at `:1106`. Plumb `interactive=not no_create` (or treat `--yes` as equivalent to non-interactive for this resolver) so the CI path is consistently fail-fast.

**4. Test coverage gaps for OAuth edge paths.** Stages 3 + 4 added solid happy-path + state-mismatch + access-denied + 500 + refresh-token-rotation + stale-refresh-fallback coverage. Three paths are still untested:
  - **userinfo fallback** at `:776-789` — IdP doesn't return `id_token` (no `DJUST_OIDC_RSA_KEY` configured). Should add a test where `/o/token/` returns a payload with no `id_token`, then `/o/userinfo/` returns `{"email":"u@e.com"}` and assert email lands on disk.
  - **browser timeout path** at `:704-725` — straightforward to exercise with the test harness's existing 5s timeout fixture: `monkeypatch` `webbrowser.open` to a no-op (no callback hit), assert the CLI raises `Timed out waiting for browser login`.
  - **`webbrowser.Error` headless path** at `:696-700` — assert that even when `webbrowser.open` raises, the URL is still echoed and the loopback server is still listening.

**5. Time-of-check race on the timeout path.** `_login_browser` at `python/djust/deploy_cli.py:702-727`: if a legitimate callback arrives between `thread.join(timeout=…)` returning and the timeout-self-`requests.get`, the legitimate `?code=…` is recorded and then the self-`?error=timeout` overwrites `http_server.captured`. Then the unconditional `if timed_out: raise click.ClickException("Timed out…")` at `:722` raises, and the just-captured legitimate code is discarded. Low-likelihood (millisecond window at the 300s mark) but the user's "I just clicked Allow" perception will see a confusing "Timed out" error. A simple snapshot-then-overwrite-only-if-still-empty would make this race-safe:
```python
if timed_out:
    if http_server.captured is None:
        try: requests.get(..., timeout=1)
        except requests.RequestException: pass
        thread.join(timeout=2)
```

### 🟢 Nice-to-have

**6. `_api_headers` raises `KeyError` rather than `ClickException` on a malformed bearer creds dict** (`python/djust/deploy_cli.py:170`). If the credentials file got corrupted to `{"auth_scheme": "bearer"}` without `access_token`, the user sees a Python traceback instead of "Saved credentials are no longer valid — please re-authenticate." Cheap fix: `creds.get("access_token", "")` and a "missing access_token in credentials — run `djust deploy login`" path.

**7. `tarfile.open(... "w:gz")` writes default-mode (often 0o644) tarballs of the working tree.** `_create_tarball` at `:942` doesn't filter sensitive content beyond the dir/file pattern matches. `*.env` files (lowercase, dotted) are excluded as `dirs[:] = [d for d in dirs if not any(pattern in d ...)]` — but this matches substrings, so a directory called `unrelated.envcfg` would also be excluded, while `.env.local` (a file) would fall through the file filter (which only checks `.env`, but `.env.local` contains `.env`). Worth a tighter test (and the negative-match cases would be a nice-to-have).

### Summary

The OAuth+PKCE+loopback flow is well-shaped (proper RFC 8252 header hygiene, `secrets.compare_digest` on state, refresh-token rotation, refresh-on-401 silent re-auth). The CSP-strict + log-formatting canon rules are followed. The blockers are (a) the changelog/code key-name disagreement and (b) the slug-collision append-into-an-existing-table bug — both narrow but real.

REQUEST_CHANGES 2

johnrtipton:
## Pipeline-ship retrospective — PR #1422

### What was shipped
Guided `djust deploy` onboarding (slug-from-pyproject, project-create inline, dirty-tree check) plus an OAuth Auth Code + PKCE browser-flow login replacing the prior email/password prompt. 5 already-pushed commits on entry → 5 added during /pipeline-ship → squash-merged as `ac1340aa`. Net diff: +1729 / −136 across `python/djust/cli.py`, `python/djust/deploy_cli.py`, `python/tests/test_deploy_cli.py`, `Cargo.lock`, `CHANGELOG.md`.

### Quality rating: 4 / 5
A 5/5 would have shipped no 🔴 findings during the pipeline. Five 🔴 findings landed across the three review stages (3, 4, 7) plus two CI-found (CodeQL ReDoS, py3.13 race). All addressed before merge. The pipeline did exactly what it's for — caught them.

### Findings caught (and the stage that caught them)

| Stage | Severity | Finding |
|---|---|---|
| 3 (Self-Review) | 🔴 | `interactive=not no_create or True` — operator-precedence bug; `--no-create` would still launch a browser |
| 4 (Security) | 🔴 | Loopback callback HTML missing `Referrer-Policy: no-referrer` + `Cache-Control: no-store` (RFC 8252 §8.10) |
| 4 (Security) | 🔴 | `logger.debug` of `resp.text` from the `/o/token/` refresh path leaked rejected refresh_token |
| 4 (Security) | 🟡 | `state` compared with `==` rather than `secrets.compare_digest` |
| 4 (Security) | 🟡 | `--server` accepted `http://` for non-loopback hosts (cleartext token transmission) |
| 7 (Code Review) | 🔴 | `_save_slug_to_pyproject` produced unparseable TOML on the server slug-uniquification path (duplicate `[tool.djust.deploy]` table) |
| 7 (Code Review) | 🔴 | CHANGELOG drift — wrote `[tool.djust] project_slug` but code uses `[tool.djust.deploy] project` |
| 7 (Code Review) | 🟡 | `_resolve_project_slug` ignored `interactive=not no_create` from both call sites — same op-precedence class as the Stage 3 finding |
| CI (CodeQL) | 🔴 high | `py/redos` on the regex I introduced to fix the duplicate-table bug — replaced with line-based parser |
| CI (py3.13 only) | 🔴 | Race in `_simulate_browser` — daemon thread populated response headers AFTER `runner.invoke` returned; py3.13 scheduling exposed it. Added a `threading.Event` completion signal |

### Process notes (audit-driven)

- **Stage 7's value cleared its own bar.** Stages 3 + 4 had already addressed all 🔴 they found, and the resulting code passed all 51 deploy_cli tests + 6992 in the wider suite. Stage 7 STILL surfaced 2 🔴 missing from Stage 3's net (the duplicate-table bug, the CHANGELOG drift). Confirms Action #1138's "🔴 1 wasn't a Stage 7 false-positive" finding generalizes — independent reviewers regularly find what self-review misses on security-sensitive code.

- **CodeQL caught what the human review missed.** The ReDoS pattern in my fix for the duplicate-table bug was overlooked by both Stage 7 and Stage 3 even though it touched a regex I had just written. Static-analysis-after-each-fix-pass is non-negotiable on this kind of work — a single regex with `(?:[^\[]*\n)*?` is a CodeQL-flagged ReDoS surface and human reviewers won't catch it.

- **py3.13 race was a one-off scheduling reveal of a pre-existing bug.** The test was racy on every Python version; py3.12 and py3.14 just happened to schedule the daemon thread before the assertion. The fix (`threading.Event` + `done.wait()`) is the correct shape for any test that relies on a daemon thread populating shared state — copy-paste this for the next test like it.

- **User-flagged "the test keeps triggering my browser"** during the run prompted an autouse `_no_real_browser` fixture (commit `4080d42a`). Even though every test was already mocking webbrowser, the safety net costs nothing and prevents a future test from leaking a real browser pop-up to the developer.

### Two-commit shape adherence
Mostly clean. Five commits landed across the pipeline:
- `252a4814` fix-pass (Stage 3 + 4 findings)
- `3c9c3f0b` docs (CHANGELOG)
- `27d8851d` fix-pass (Stage 7 findings — CHANGELOG drift, duplicate-table bug)
- `c48783ba` fix-pass (CodeQL ReDoS + 4 notes)
- `5666822f` fix-pass (py3.13 race)
- `4080d42a` test-hygiene (autouse no-real-browser)

The CHANGELOG commit is separate from impl per Action #181. The post-Stage-7 commits don't follow the two-commit shape strictly (CHANGELOG was edited inline in `27d8851d` because it was itself the 🔴 finding); that's the right call when the doc is the bug.

### Action items / deferrals
None new. The 🟡 "test coverage gap" findings from Stage 7 (userinfo fallback, browser-timeout, `webbrowser.Error` paths) and the 🟢 nice-to-haves (`_api_headers` raising `ClickException` on malformed bearer creds, `.env.local` exclusion fall-through) are deferred — not load-bearing for the v0.9.6 release surface.

🤖 Generated by `/pipeline-ship` Stage 10

Issue #250, opened 2026-02-08
Full HTML update path mangles page rendering


johnrtipton:
## Bug Description

When the VDOM engine falls back to a full HTML update (instead of patches), the client-side rendering can mangle the page. This affects any case where `patches` is `None` — first render recovery, version mismatch, component events, structural template changes, etc.

### 1. Single root assumption
`getLiveViewRoot()` returns the first `[data-djust-root]`, but multi-component pages may have multiple roots. The parsed server HTML's `querySelector('[data-djust-root]')` may match a different root than expected.

### 2. Version mismatch bypasses `applyDjUpdateElements`
The version mismatch recovery path (`02-response-handler.js:47-57`) calls `morphChildren()` directly, skipping `applyDjUpdateElements()`. This means `dj-update="append|prepend|ignore"` attributes are not respected during version recovery.

### 3. `preserveFormValues` not called in full HTML path
The full HTML update path (`02-response-handler.js:121-143`) does not call `preserveFormValues()`, so focused form element values can be lost or replaced during the update.

### 4. Nested `data-djust-root` matching
`querySelector('[data-djust-root]')` on the parsed server HTML returns the outermost match, which may not correspond to the live DOM's root when nesting is involved.

## Reproduction

1. Use any view where the VDOM engine returns no patches (e.g., toggling a `{% if %}` block)
2. The server sends a full HTML update instead of patches
3. The page layout/content gets corrupted

## Expected Behavior

Full HTML updates should produce the same visual result as an efficient patch update — the fallback path should be visually indistinguishable from the patch path.

## Files Involved

- `python/djust/static/djust/src/02-response-handler.js` (lines 47-57, 121-143)
- `python/djust/static/djust/src/12-vdom-patch.js` (`applyDjUpdateElements`, `morphChildren`, `morphElement`)

## Suggested Fix

1. Unify the version mismatch and normal full HTML paths to both go through `applyDjUpdateElements`
2. Add `preserveFormValues()` to the full HTML update path
3. Scope root matching to the specific `data-djust-view` container when available

johnrtipton:
Fixed in PR #251 — DOM morphing (`morphChildren()`/`morphElement()` in `12-vdom-patch.js`) replaced the `innerHTML` full-update path with morphdom-style DOM diffing. Preserves event listeners, form values, and binding flags. Tests in `morph.test.js`.

PR #1429, opened 2026-05-08
feat(websocket): persist LiveView state on WS event for reconnect continuity

johnrtipton:
## Summary

Three companion changes that together let LiveView state survive a WebSocket reconnect when the view opts in via `enable_state_snapshot`. Motivated by the **djustlive proxy's snapshot/wake cycle (Phase 11)**, but the gap they close is general — any WS-event-driven mutation (`dj-click`, `dj-input`, `dj-submit`) was previously lost on reconnect even with `enable_state_snapshot = True`.

### 1. Save in `handle_event` (`websocket.py` ~line 3105)

After every successful WS event handler, mirror the HTTP-path save in `mixins/request.py:603-609`: pull a fresh `get_context_data()` snapshot, filter out `LiveComponent` instances, normalize, and `aset` to the Django session under `liveview_<page_url>`. Private (`_`-prefixed) attrs and components are persisted alongside, with explicit cleanup of empty `__private` keys. Wrapped in `try/except + logger.exception` so save failures never break event handling.

### 2. Loosen the load gate in `handle_mount` (`websocket.py` ~line 1988)

The existing `if has_prerendered:` guard was correct for its case (skip `mount()` re-execution when the browser carries SSR HTML), but it meant pure-WS reconnect (no SSR — e.g. after the djustlive proxy force-closed the backend WS for snapshot) never reached the `aget(view_key)` lookup. Widened to `if has_prerendered or saved_state:`. Behavior unchanged for views with no saved state.

### 3. Skip `html` in mount response on resume (`websocket.py` ~line 2317)

When `mounted` (state was restored from session) AND `has_prerendered`, omit the freshly-rendered `html` field. The client's DOM already reflects the saved state — sending the html would trigger a redundant morphdom-style DOM swap. `client.js`'s `e.html && (n.innerHTML=e.html)` short-circuits cleanly when the field is absent. The `version` field still flows so subsequent patches stay in sync. Mount response shrinks from ~12KB to ~500 bytes on the resume path. **No visible page refresh on reconnect.**

## Why this matters

`enable_state_snapshot` was effectively HTTP-only before this change. Any reactive page that lives mostly in the WS event loop — counters, form drafts, multi-step wizards, presence/realtime UIs — would lose every keystroke or click on reconnect, even with the opt-in flag set. With these three changes the flag does what its name suggests for the WS path.

The most visible consequence is that the djustlive proxy's snapshot/restore feature (kill firecracker, free 384 MB host RAM, restore in ~10 ms on next request) no longer eats LiveView state. This is what unblocks the platform's "scale-to-zero with sub-50 ms wake" story for stateful apps.

## Test plan

End-to-end verified against `djust-scaffold`'s `DashboardView.click_count` counter through the djustlive proxy's snapshot/wake cycle:

- [x] Pre-suspend counter at N
- [x] 75 s idle → proxy snapshots → fc killed → ~384 MB host RAM released
- [x] Click on resume → wake from snapshot in ~10 ms
- [x] Counter ticks to N+1 (state preserved)
- [x] No DOM swap (mount response is ~500 bytes, html field omitted)
- [x] No `request_html` / `html_recovery` cycle (version stays in sync)
- [x] No `View not mounted` errors on the redialed backend WS

Reproduced with the `direct_check.go` and `state_test.go` Go probes documented in `djust-scaffold/.scratch/phase-11-mvp/RESULTS.md`.

### Suggested follow-up tests (not in this PR)

- [ ] Unit test for the loosened load gate (mount with `has_prerendered: false` + saved state restores)
- [ ] Unit test for `skip_html_for_resume` (mount response has no html when both flags set)
- [ ] Integration test using `WebsocketCommunicator` for the full save → reconnect → load roundtrip

## Notes

- All three changes are gated by existing opt-in flags (`enable_state_snapshot`, `has_prerendered`); behavior for non-opt-in views is unchanged.
- Save is best-effort — failures are logged but never propagated to the event handler.
- The session save uses Django's `aset` async API; matches the `aget` pattern already in `handle_mount`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- **LiveView state survives WS reconnect (Phase 11 / djustlive snapshot/wake).** The `LiveViewConsumer` now persists view state to the Django session on every WS event handler completion, mirroring the save that already existed for the HTTP GET/POST paths in `mixins/request.py:603-609`. Previously, state changes driven by `dj-click` / `dj-input` / `dj-submit` events lived only in the consumer's in-memory view instance — any WS reconnect (page refresh, network blip, the djustlive proxy's snapshot/restore cycle) discarded everything since the last full HTTP request. Now a click that increments `self.click_count` is durable across reconnects, opt-in via the existing `enable_state_snapshot = True` class attribute. Save fires after the user handler returns and before render, so saved values include the just-applied mutation. Components, private (`_`-prefixed) attrs, and explicit cleanups for empty private-state are all mirrored from the HTTP-path semantics; failures in save are caught and logged (never break event handling). Fixes the long-standing gap where `enable_state_snapshot` was effectively HTTP-only.
+
+### Changed
+
+- **Mount-time state restore now fires on plain WS reconnect, not just SSR-prerendered loads.** The `if has_prerendered:` gate that wrapped the `request.session.aget(view_key)` lookup in `LiveViewConsumer.handle_mount` was widened to `if has_prerendered or saved_state:`. The original gate was correct for its intended case (skip running `mount()` again when the browser already has SSR HTML), but it also meant pure-WS apps and reconnect-after-reload paths could never restore state from the session — even with the new event-time save in place. With the widened gate, any reconnect that finds saved state hydrates from it; views with `enable_state_snapshot = True` get true reconnect continuity. Behavior unchanged for views without saved state (gate is or'd, not replaced).
+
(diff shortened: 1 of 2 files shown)


johnrtipton:
## Closing in favor of #1465 (clean redo)

This PR is 29 commits behind `origin/main` (opened pre-v0.9.6, main has moved through v0.9.6 stable + v0.9.7rc1 since) and `mergeable: CONFLICTING`. Per Action #250 / #1450 (Stage 11 stale-base check, canonized in PR #1451), reviewing+merging a stale-base PR against `python/djust/websocket.py` would review a different program than the merge applies. PR #1431 was the canonical case study — same shape, would have silently undone an entire milestone.

Also: no automated tests. The original PR body listed 3 follow-up tests; per `CLAUDE.md` "All new code needs tests", those are now in-scope for the redo.

**Redo issue**: #1465 — full spec preserved, in-scope tests added, ready for `/pipeline-run --milestone v0.9.7-2`.

This PR will be closed without merge once #1465 lands.

johnrtipton:
Superseded by PR #1466, which landed at `8a3a3a3` (or whatever the squash hash is). The redo:
- Rebased onto current main (no stale-base review trap; PR #1431 was the canonical case study where this would have silently undone a milestone)
- Added 7 tests including a real `WebsocketCommunicator` integration test (the original PR had zero automated tests)
- Stage 11 found 4 🟡 items (CHANGELOG over-claim, child-view save-to-wrong-key bug, inverted save order vs HTTP path, tautological tests). All addressed pre-merge via Stage 13 fix-pass.
- Filed #1467 as follow-up for WS-event save for child LiveComponent views (out of scope for this PR)

Closing this stale PR without merge. The change shipped via #1466.

Issue #1410, opened 2026-05-07
hardening(state): `InMemoryStateBackend` silently returns shared-ref on msgpack deserialize failure (race risk)


johnrtipton:
## Summary

`python/djust/state_backends/memory.py:119` (in `InMemoryStateBackend.get`) catches `msgpack` deserialization errors and falls back to **returning the shared in-memory ref directly** with a logged warning of the form:

\`\`\`
[ERROR] InMemoryStateBackend.get: serialize/deserialize round-trip failed for key '...'; returning shared ref (race risk)
ValueError: MessagePack deserialization error: invalid length 5, expected struct VNode with 6 elements
\`\`\`

Two concrete-DoSes from this fallback:

1. **Cross-session state corruption.** Two concurrent connections to the same view get the same in-memory `_rust_view` reference instead of independent clones. Mutations from connection A leak into connection B's render context.
2. **Stale-state survival across struct changes.** If a `dev .so` is hot-swapped and the `VNode` field count changes, previously-serialized state can no longer round-trip. Rather than failing fast (which would force a clean re-mount), the backend hands the next mount a stale view instance whose state mismatches the request URL params. Observed during the #1408 investigation: after a `cargo build` of djust_vdom + `.so` swap, fresh navigations to `?btab=documents` re-rendered with `active_bottom_tab=\"notes\"` from the prior session's state.

## Suggested fix

When `deserialize_msgpack` raises, the backend should:
- log at ERROR (already does),
- **discard** the corrupt entry (`pop` it from the in-memory dict),
- raise / return `None` so the caller's mount-path treats the view as uninitialized and re-runs `mount()` cleanly.

Returning a shared ref is the most surprising and silent-corrupting path; the data-loss alternative (forcing fresh mount) is strictly safer.

## Related

- Surfaced during the investigation of #1408 (`sync_ids` dj-if boundary asymmetry).
- Possibly contributes to the visible bleed in #1409 (post-mount DOM contains stale dj-ids), though the proximate cause there appears to be on the client side.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1431, opened 2026-05-09
fix(state-backends/redis): per-thread ZstdCompressor/Decompressor (#1430)

### LINKED ISSUE #1430: RedisStateBackend shares one ZstdDecompressor across threads — segfaults under concurrent load
# `RedisStateBackend` shares one `ZstdDecompressor` across threads — segfaults under concurrent load

## Summary

`djust.state_backends.redis.RedisStateBackend` stores a single
`zstandard.ZstdDecompressor` (and `ZstdCompressor`) instance on `self` at
`__init__` time, then reuses it from every caller. python-zstandard
**explicitly documents** that compressor/decompressor objects are not
thread-safe when shared across threads — see
[indygreg/python-zstandard#244](https://github.com/indygreg/python-zstandard/issues/244).
Sharing the same object across concurrent calls produces:

- decompression-error exceptions ("Unknown frame descriptor", "Data
  corruption detected") — the visible-but-survivable failure mode, and
- **segfaults inside the C extension** — reproduced reliably on Linux
  4.14 + Python 3.12 + zstandard 0.25.0; observed as `python3: segfault
  ... in backend_c.cpython-312-x86_64-linux-gnu.so` followed by
  `Kernel panic — Attempted to kill init!` when init is uvicorn (PID 1
  in a microVM).

When this fires under uvicorn it kills the process; in our deployment
target (one app per microVM) it kills the entire VM and triggers a
cold reboot.

## Location

`python/djust/state_backends/redis.py` — `RedisStateBackend.__init__`:

```python
# Initialize zstd compressor/decompressor if available
if self._compression_enabled:
    self._compressor = zstd.ZstdCompressor(level=compression_level)
    self._decompressor = zstd.ZstdDecompressor()
```

These are then used unguarded from `_compress` and `_decompress`, which
are called from request-handling code paths (`get`, `set`, `delete`, …)
that run concurrently across uvicorn worker threads / asyncio executor
calls.

## Reproducer (minimal)

```python
# repro_zstd_shared.py
import threading
import zstandard as zstd

# One shared decompressor — same as RedisStateBackend
dec = zstd.ZstdDecompressor()
comp = zstd.ZstdCompressor(level=3)

# A realistic-sized djust LiveView state payload
payload = comp.compress(b"x" * 50_000)

errors = []
def hammer():
    for _ in range(2000):
        try:
            dec.decompress(payload)
        except Exception as e:
            errors.append(repr(e))

threads = [threading.Thread(target=hammer) for _ in range(16)]
for t in threads: t.start()
for t in threads: t.join()
print(f"{len(errors)} exceptions; sample: {errors[:3]}")
# Expect: non-zero error count OR a segfault (SIGSEGV exit; no Python
# traceback). Both indicate the same race.
```

## Reproducer (in situ)

We hit this on a Phase-11 microVM deployment of `djust-scaffold`:

1. `DJUST_STATE_BACKEND=redis://...` (so `RedisStateBackend` is active)
2. Compression enabled (default)
3. State for several LiveView sessions in Redis (any prior run)
4. Cold-boot the VM
5. Have any reverse-proxy probe the HTTP endpoint while uvicorn is
   warming up (Caddy with default health checks is enough)
6. uvicorn segfaults at exactly the moment the first event handler
   calls into `_decompress` from a worker thread while another callsite
   (mount lookup) is also calling it on the asyncio executor

The crash is at offset `+0x72d16` in
`zstandard/backend_c.cpython-312-x86_64-linux-gnu.so` —
`ZSTD_decompressSequencesLong_default.constprop.0`. The faulting access
is far above `%rsp`, consistent with an internal-state-corruption →
wild-pointer dereference rather than a clean stack overflow.

## Affected versions

- djust: confirmed on `0.9.5` (the version on the rootfs we tested);
  the pattern is present on `main` as of this report.
- zstandard: confirmed on `0.25.0`. The thread-safety constraint is a
  property of the C library (libzstd) and python-zstandard's wrapper —
  it applies to all versions; upgrading python-zstandard does not fix
  it. (Upstream's stance, per #244, is "don't share these objects.")

## Fix options

In order of preference:

### Option A — `threading.local` instances (recommended)

```python
import threading

class RedisStateBackend(StateBackend):
    def __init__(self, ..., compression_level: int = 3, ...):
        ...
        self._compression_level = compression_level
        self._tls = threading.local()

    def _get_compressor(self):
        c = getattr(self._tls, "compressor", None)
        if c is None:
            c = zstd.ZstdCompressor(level=self._compression_level)
            self._tls.compressor = c
        return c

    def _get_decompressor(self):
        d = getattr(self._tls, "decompressor", None)
        if d is None:
            d = zstd.ZstdDecompressor()
            self._tls.decompressor = d
        return d
```

Replace `self._compressor.compress(...)` → `self._get_compressor().compress(...)` etc.

- ✅ Each thread gets its own object; no race.
- ✅ Constructor cost amortized per-thread, not per-call.
- ✅ Drop-in for sync code paths.
- ⚠️ For asyncio executor pools, instances are tied to executor threads —
  fine, since each thread serializes work.

### Option B — fresh object per call

```python
def _decompress(self, data: bytes) -> bytes:
    ...
    return zstd.ZstdDecompressor().decompress(payload)
```

- ✅ Trivially correct, mirrors what upstream recommends.
- ❌ ~10–50 µs per construction adds up on hot paths; we can measure if
  it matters before shipping. For payloads of djust-typical size
  (KB–tens-of-KB) the construction cost is small relative to the
  decompress itself.

### Option C — explicit lock

```python
def __init__(self, ...):
    ...
    self._compressor_lock = threading.Lock()
    self._decompressor_lock = threading.Lock()

def _decompress(self, data):
    ...
    with self._decompressor_lock:
        return self._decompressor.decompress(payload)
```

- ✅ One object, no construction churn.
- ❌ Serializes all decompress calls across the process. Bad for
  multi-LiveView throughput. Don't pick this one.

We ran into this on a microVM-per-app deployment where one segfault =
one dead VM. Happy to put up a PR with Option A and a regression test
that mirrors the reproducer above.

## Severity

We'd argue **High**:

- Default config (`COMPRESSION_ENABLED=True`, threshold 10 KB) means
  any project with non-trivial LiveView state hits the corruption
  surface.
- Failure mode is a SIGSEGV that takes the worker process down — no
  graceful 500, no traceback.
- The race is non-deterministic so it lands in production rather than
  in CI.

## References

- python-zstandard upstream:
  [indygreg/python-zstandard#244 — `zstd.decompress` not threadsafe for
  shared objects](https://github.com/indygreg/python-zstandard/issues/244)
  (closed, "by design")
- djust: `python/djust/state_backends/redis.py` (constructor + the
  `_compress`/`_decompress` callers)

johnrtipton:
## Summary

- Fixes #1430: shared `ZstdCompressor` / `ZstdDecompressor` on `RedisStateBackend` raced under concurrent callers — symptoms ranged from `ZstdError("Unknown frame descriptor")` and `"Data corruption detected"` exceptions through to SIGSEGV inside `ZSTD_decompressSequencesLong_default` on Linux 4.14 + Python 3.12 + zstandard 0.25.0.
- Moves both objects into a `threading.local`, accessed lazily through `_get_compressor()` / `_get_decompressor()`. Each thread gets its own instance — no shared C-level state, no race.
- Compression-on-write and decompression-on-read are now decoupled: a backend that has compression disabled can still read previously-compressed values from Redis (matters for upgrade paths and for backends that rotate compression settings).
- python-zstandard's stance, per [indygreg/python-zstandard#244](https://github.com/indygreg/python-zstandard/issues/244): compressor/decompressor objects are not thread-safe; users must not share them. This PR makes djust comply with that contract.

## Test plan

New file `python/djust/tests/test_redis_state_backend_thread_safety.py`:

- [x] `test_concurrent_compress_decompress_no_race` — 16 threads × 250 ops against `_compress` + `_decompress` on a 50 KB payload; passes on the fix, would fail / SIGSEGV on the pre-fix code (verified with a standalone repro using the exact workload — exit code 139 on shared instances).
- [x] `test_per_thread_decompressor_identity` — N concurrently-alive threads each receive a distinct `ZstdDecompressor` object identity; guards against a future revert to shared state.
- [x] `test_decompressor_reused_within_thread` — same thread reuses its decompressor across calls (avoids per-call construction cost on hot paths).
- [x] Existing related tests still green: `test_resumable_uploads_821::TestRedisStateStore::*` (8 tests), `test_state_backend_delete_all::*` (4 tests).

## Notes

- No public API change; backward compatible.
- No CHANGELOG entry expected on this branch's two-commit shape — happy to add one in a follow-up commit on this PR if the reviewer prefers.
- I deliberately did not add an explicit lock (Option C from the issue) — that serializes all decompress calls process-wide, which would defeat the throughput benefits of multi-threaded uvicorn workers handling LiveView state lookups.


johnrtipton:
## Performance impact

Microbenchmarks on the same `_compress` + `_decompress` round-trip the regression test exercises (50 KB compressible payload, level 3, zstandard 0.25.0):

### Single-threaded latency

| | ops/s | µs/op |
|---|---|---|
| Pre-fix (shared instance) | 45,954 | 21.8 |
| Post-fix (TLS-cached) | 45,434 | 22.0 |
| **Delta** | **−1.1%** (run-to-run noise) | |

The extra `getattr(self._tls, "compressor", None)` is a few-ns lookup; constructor cost is amortized to once per thread.

### Multi-threaded throughput

This is the case the old code couldn't survive — under any concurrency it raced or SIGSEGV'd. New numbers:

| Threads | ops/s | speedup vs 1-thread |
|---|---|---|
| 1 | 45,434 | 1.00× |
| 2 | 75,902 | **1.67×** |
| 4 | 77,112 | 1.70× |
| 8 | 56,104 | 1.24× |
| 16 | 55,723 | 1.23× |

Scaling tops out at ~4 threads (zstandard releases the GIL during compress/decompress, beyond that we're paying GIL/context-switch overhead). The right comparison is "0 ops/s under contention" → "~76 k ops/s at 4 threads."

### Memory cost

Per active thread (measured with `ru_maxrss` on macOS, 100-instance batch):

| | per instance |
|---|---|
| `ZstdCompressor(level=3)` | 6.6 KB |
| `ZstdDecompressor()` | 32.3 KB |
| **Per-thread total** | **~39 KB** |

For a 32-thread uvicorn worker pool: ~1.2 MB extra RSS. Working buffers during active compress/decompress are allocated per-call regardless of TLS vs shared, so the per-call cost is unchanged.

### What didn't change

Compression ratio, wire format, Redis-key layout, public API, the `_compression_enabled` switch.

johnrtipton:
## Code review — PR #1431 (zstd thread-safety fix for `RedisStateBackend`)

The `redis.py` change itself is correct and well-targeted. The blocker is **mechanical, not technical**: the branch was cut from `9a4177b5` (the 0.9.6rc1 tag commit) and never rebased onto `main`, so its diff against `main` contains massive collateral deletions that will revert seven already-merged PRs.

### Branch hygiene — must rebase before merge

🔴 **must-fix — branch is 7 PRs behind `main` and will revert them on merge.** `git merge-base main origin/fix/redis-state-backend-zstd-thread-safety` resolves to `9a4177b5` (rc1). The diff against `main` deletes:
- `CHANGELOG.md:10-26` — five `[Unreleased]` entries (#1410, #1433, #1435, #1436, #1437)
- `RETRO.md` — 90 lines of v0.9.6-1 retro
- `ROADMAP.md` — v0.9.6-1 drain bucket
- `crates/djust_vdom/tests/wire_protocol_snapshot.rs` (PR #1444 — 311 lines)
- `python/djust/checks.py` — D001 system check (PR #1440)
- `python/djust/tests/test_tenants_middleware.py` (#1441), `test_theming_context_cache.py` (#1442)
- `python/tests/test_checks.py`, `tests/unit/test_state_backend.py`, `docs/website/guides/template-cheatsheet.md`
- Reverts the perf rewrite of `theming/context_processors.py` and `tenants/middleware.py`, and the corruption-fix in `state_backends/memory.py` (#1438)

Action: `git rebase origin/main` on the PR branch and force-push. After rebase the diff should be ~50 LOC across `redis.py` + the new test file.

### CHANGELOG

🔴 **must-fix — no `[Unreleased]` entry for #1430.** Project canon requires every `fix:`/`feat:` PR to ship a CHANGELOG entry. After the rebase, add a `### Fixed` line under `[Unreleased]` referencing #1430 and citing the SIGSEGV symptom — operators searching for `ZSTD_decompressSequencesLong_default` in postmortems need the breadcrumb.

### Correctness of the threading.local migration

🟢 `python/djust/state_backends/redis.py:175,206` — every read of `self._compressor` / `self._decompressor` is replaced with the lazy getter; no stale references. `grep -n "_compressor\|_decompressor"` against the new file shows only the two getters and the two consumers.

🟢 `redis.py:145-160` — `_get_decompressor()` correctly keys on `ZSTD_AVAILABLE` rather than `_compression_enabled`. This is a real correctness point and the docstring calls it out: a backend that flipped `compression_enabled=False` after a partial rollout still has to read previously-compressed values written by peer pods. ✓

🟢 No `__del__` or `.close()` cleanup existed on the old objects (`git show main:.../redis.py | grep close|__del__` confirms none). Per-thread instances live as long as the worker thread, which is the standard uvicorn / channels async-executor pattern — fine.

### Test coverage

🟢 `test_concurrent_compress_decompress_no_race` — 16 threads × 250 ops, `threading.Barrier(16)` to maximize first-call concurrency, `compression_threshold_kb=0` to force the C path on every call. This is the right shape for the bug. The pre-fix code reproduces the race well within 250 ops at 16-way concurrency.

🟢 `test_per_thread_decompressor_identity` — the `release.wait()` + main-thread polling loop correctly guards against TID recycling; the docstring explicitly calls this out. Good.

🟡 **should-fix — `test_per_thread_decompressor_identity` polling loop creates a fresh `threading.Event()` per iteration just to sleep** (`test_redis_state_backend_thread_safety.py:131`). Use `time.sleep(0.01)` instead — same effect, less garbage, clearer intent.

🟡 **should-fix — no test for compressor identity, only decompressor.** Symmetry: add a `test_per_thread_compressor_identity` mirror, or fold the compressor into the existing test (call `_get_compressor()` alongside `_get_decompressor()` in `grab()`). The fix lives or dies on both objects being thread-local.

🟢 `test_decompressor_reused_within_thread` — locks in the lazy-init-cache invariant. Same-thread `id()` equality is the right assertion.

### Performance

🟢 `getattr(self._tls, "compressor", None)` is on the per-frame hot path but costs ~150 ns vs the ~µs–ms of the actual compression work — not a regression. The lazy-init pattern caches per-thread (not per-call), which is what you'd want.

### Configuration plumbing

🟢 `_get_compressor()` honors `self._compression_level` on each new per-thread instance (`redis.py:140`). `ZstdDecompressor` takes no constructor args, so no plumbing concern there. If a future PR adds `dict_data=` it'll need to remember the same per-thread treatment — worth a comment but not a blocker.

### Auto-reject canon

🟢 No new f-string-in-logger, no new `print()`, no `mark_safe`, no bare `except: pass` introduced by this PR.

🟡 `redis.py:107-109` — pre-existing f-string in `logger.info` (`f"RedisStateBackend initialized: {redis_url} ..."`). Not introduced by this PR; flag for a follow-up sweep, do not block on it here.

---

`REQUEST_CHANGES 4`

(Two 🔴: rebase-onto-main + CHANGELOG. Two 🟡: per-thread compressor identity test + sleep idiom in the existing identity test.)

johnrtipton:
## Pipeline-run retro — PR #1431 (#1430)

### Quality: 4/5
P0 segfault-class fix landed clean post-rebase. Stage 11 review caught the high-leverage problem: branch was 7 PRs behind main and would have silently reverted all of v0.9.6-1's drain (#1438–#1444) on merge. Rebase + targeted fixes shipped on the second CI cycle.

### Findings caught and fixed
- 🔴 **Branch 7 PRs behind main** (Stage 11) — would have silently reverted CHANGELOG entries, the v0.9.6-1 retro, the wire-protocol snapshot test, D001, and the perf rewrites of theming + tenants middleware. Fixed via `git rebase origin/main`.
- 🔴 **No CHANGELOG entry for #1430** — segfault-class fix without an operator-discoverable line. Added.
- 🟡 **Symmetric compressor test missing** — only decompressor identity asserted; the fix lives or dies on both. Added `test_per_thread_compressor_identity` + `test_compressor_reused_within_thread`.
- 🟡 **threading.Event().wait(0.01) in polling loop** — wasteful. Replaced with `time.sleep`.

### Process notes
- The 'branch behind main' check is a generalizable Stage 11 gate. Any PR that opened before the current main HEAD needs a rebase-or-merge step BEFORE the actual review; otherwise the reviewer is reviewing against a stale base. Worth filing as a Stage 11 mandatory item.

Issue #1408, opened 2026-05-07
bug(vdom): `sync_ids` is not dj-if-boundary-aware — branch swaps leave stale dj-ids in `last_vdom`


johnrtipton:
## Summary

`crates/djust_vdom/src/diff.rs::diff_children` is dj-if-boundary-aware (calls `dj_if_pre_pass`, #1358). Its post-diff sibling `sync_ids` is not: it falls straight through to `sync_ids_keyed` / `sync_ids_indexed`, which positionally align children regardless of dj-if branch identity.

After a `{% if %}` branch swap, fresh dj-ids on new-branch content are overwritten by stale dj-ids from the unmatched-old-branch. The next render's diff then emits `RemoveChild` / `SetAttr` patches whose `child_d` / `d` reference ids that the client DOM never had → orphaned content on tab swaps.

## Concrete trace evidence (NYC Claims `/claims/178/` bottom-tab swap, `DJUST_VDOM_TRACE=1`)

Pre-fix:
```
Cycle 2 (docs→notes): InsertChild path=[2v] index=3 dj-id="5m"   (tasks-panel)
Cycle 3 (notes→docs): RemoveChild path=[2v] index=3 child_d="2y" ❌ stale
```

Post-fix:
```
Cycle 3 (notes→docs): RemoveChild path=[2v] index=3 child_d="5m" ✓ matches the previous InsertChild
```

`2y` was the dj-id assigned to a Document Checklist wrapper several renders earlier; it persisted through `sync_ids_indexed`'s positional alignment of unrelated cross-branch content.

## Fix

Add `sync_ids_dj_if_pre_pass` mirroring `dj_if_pre_pass`:
- id only in OLD → skip (subtree being removed)
- id only in NEW → skip (subtree brand-new; preserve fresh ids)
- id in BOTH → recurse `sync_ids` into the matched body
- non-boundary siblings → pair by relative order outside boundary ranges (same `build_excluded_mask` semantics)

## Tests

`crates/djust_vdom/tests/test_sync_ids_dj_if_1408.rs` — 4 cases. 2 fail pre-fix, all 4 pass post-fix. All 228 existing `djust_vdom` tests still pass.

## Out of scope

The visible bleed at `/claims/178/` has at least one additional client-side root cause that this fix does not address (separate issue forthcoming).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #1198, opened 2026-04-28
tech-debt: commit-or-rollback handler shape


johnrtipton:
Source: Retro v0.9.4 / PR #1193

Any async handler that does BOTH a state mutation AND has an early-return path (validation failure, downstream failure, missing dependency) should mutate AFTER the commit point. Two clean shapes:
1. Defer the mutation past all early-return checks (preferred for single-attribute mutations).
2. Wrap in try/except with explicit rollback (only justified when multiple mutations need atomic rollback).

**Failure mode**: silent. The early-return doesn't raise, so observability tools won't flag it. State stays in a half-committed shape.

**Observed instance**: PR #1193's `handle_forward_replay` set `view._time_travel_branch_id = new_branch` BEFORE awaiting `replay_event`. On `replayed is None` (handler missing/un-decorated), branch state stayed bumped with no recorded events; view + client diverged about the active branch. Stage 11 caught it.

Canonicalize as a CLAUDE.md rule: when implementing async handlers with state mutations and early-return paths, audit ordering.

johnrtipton:
Shipped in PR #1225 (v0.9.4 retro process canon batch in CLAUDE.md). Comma-list `Closes #...` form in PR body parses only the first ref — the rest had to be closed manually (this confirms #1184's diagnosis once again; the PR-checklist convention says one Closes per body line). 🤖 v0.9.1-6 drain bucket — final issue closure cleanup.

PR #1438, opened 2026-05-09
fix(state-backends/memory): discard corrupt entry on round-trip failure (#1410)

### LINKED ISSUE #1410: hardening(state): `InMemoryStateBackend` silently returns shared-ref on msgpack deserialize failure (race risk)

johnrtipton:
## Summary

Closes #1410.

`InMemoryStateBackend.get()` previously caught a serialize/deserialize round-trip exception (typically `RustLiveView.deserialize_msgpack` raising after a hot-swap struct change or msgpack schema drift) and fell back to returning the cached **shared ref**. Two concurrent connections to the same view then mutated the same `_rust_view` — a silent cross-session state leak.

The fix discards the corrupt entry from the in-memory dict and returns `None`. The caller's mount path treats it as uninitialized and runs `mount()` cleanly. Strictly safer than silent corruption.

Surfaced during the #1408 investigation: after `cargo build` of `djust_vdom` + `.so` swap, fresh navigations re-rendered with state from the prior session.

## Test plan

- [x] `tests/unit/test_state_backend.py::TestInMemoryBackend::test_get_discards_corrupt_entry_on_round_trip_failure` — patches `RustLiveView.deserialize_msgpack` to raise, asserts `get()` returns `None` and the entry is gone on the next call.
- [x] All 49 existing state-backend tests still pass.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- **`InMemoryStateBackend.get()` discards corrupt entries instead of returning the shared in-memory ref (#1410).** When `RustLiveView.deserialize_msgpack` raised — typically after a hot-swap struct change or msgpack schema drift — the previous fallback returned the cached object directly. Two concurrent connections to the same view then shared one `_rust_view`, and mutations from connection A leaked into connection B's render context. After a `cargo build` of `djust_vdom` + `.so` swap, fresh navigations could re-render with state from the prior session. Now the backend pops the corrupt entry from its in-memory dict and returns `None`; the caller's mount path treats the cache as cold and runs `mount()` cleanly. Discovered during the #1408 investigation.
+
 ## [0.9.6rc1] - 2026-05-07
 
 ### Added
--- a/python/djust/state_backends/memory.py
+++ b/python/djust/state_backends/memory.py
@@ -118,16 +118,25 @@ class InMemoryStateBackend(StateBackend):
                 serialized = view.serialize_msgpack()
                 clone = RustLiveView.deserialize_msgpack(serialized)
             except Exception:
-                # Defensive: if msgpack round-trip fails for any reason,
-                # fall back to returning the cached ref. The race window
-                # is preferable to an outright cache miss that destroys
-                # the diff baseline. Logged so this surfaces in dev.
+                # Round-trip failed (msgpack schema drift after a hot-swap,
+                # corrupt payload, etc). Returning the shared ref is
+                # silently corrupting — two concurrent connections to the
+                # same view would mutate the SAME `_rust_view` and leak
+                # state across each other (#1410). The strictly-safer
+                # alternative is to fail the cache-hit and let the caller
+                # treat this as uninitialized — `mount()` will run again
+                # and rebuild clean state.
+                #
+                # Discard the corrupt entry under the lock so the next
+                # caller doesn't re-trip the same exception.
+                with self._lock:
+                    self._cache.pop(key, None)
                 logger.exception(
                     "InMemoryStateBackend.get: serialize/deserialize round-trip "
-                    "failed for key '%s'; returning shared ref (race risk)",
+                    "failed for key '%s'; entry discarded — caller should remount",
                     key,
                 )
-                return (view, timestamp)
+                return None
             return (clone, timestamp)
 
     def set(
--- a/python/djust/state_backends/memory.py
+++ b/python/djust/state_backends/memory.py
@@ -128,9 +128,18 @@ class InMemoryStateBackend(StateBackend):
                 # and rebuild clean state.
                 #
                 # Discard the corrupt entry under the lock so the next
-                # caller doesn't re-trip the same exception.
+                # caller doesn't re-trip the same exception. Identity-
+                # guard the pop: between the unlock above and the
+                # re-lock here, a concurrent `set(key, new_view)` could
+                # have landed a fresh, valid entry — we must not delete
+                # the *new* one in place of the corrupt one we held.
+                # `cached`'s reference to `view` keeps the original
+                # alive, so `is` is sound.
                 with self._lock:
-                    self._cache.pop(key, None)
+                    current = self._cache.get(key)
+                    if current is not None and current[0] is view:
+                        self._cache.pop(key, None)
+                        self._state_sizes.pop(key, None)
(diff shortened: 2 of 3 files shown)


johnrtipton:
## Review — PR #1438 (#1410 fix)

Overall the fix correctly closes the silent shared-ref leak that #1410 reports — popping the corrupt entry and returning `None` is the right shape, the caller's mount path treats it as a cold cache, and the CHANGELOG accurately describes what shipped. One real correctness concern below, plus a couple of smaller items.

### 🟡 should-fix — TOCTOU between the unlocked round-trip and the locked pop

`python/djust/state_backends/memory.py:108-139`

The flow is:

1. **Lock**: read `cached = self._cache.get(key)` → (view, timestamp). Release lock.
2. **Unlocked**: `view.serialize_msgpack()` / `deserialize_msgpack(...)` → raises.
3. **Re-lock**: `self._cache.pop(key, None)`.

Between (1) and (3) another thread can call `set(key, new_view)` — the docstring at L78-86 explicitly notes "two HTTP requests for the same `(session, view_path)` pair landed concurrently" is exactly the regime this backend is hardened for. If a concurrent `set()` lands a fresh, valid entry between the failing round-trip and the pop, we now blow away the new entry instead of the corrupt one — turning the fix into a different correctness bug (lost write) rather than fixing one (shared ref).

Minimal fix: pop only if the entry is still the exact same object we read.

```python
with self._lock:
    current = self._cache.get(key)
    if current is not None and current[0] is view:
        self._cache.pop(key, None)
        self._state_sizes.pop(key, None)  # also matches delete() semantics
```

The `is`-comparison is safe because `cached` holds a strong reference to `view` for the entire `get()` call, so identity is stable. This also fixes the secondary `_state_sizes` leak (the existing `delete()` at L200-205 pops both; the new code only pops `_cache`).

### 🟡 should-fix — test doesn't cover the TOCTOU shape

`tests/unit/test_state_backend.py:255-298`

The patch-on-class technique is a valid way to force the exception path and the assertion shape is correct — but it only exercises the single-threaded happy/sad paths. Worth adding one of:

- A concurrent `get` + `set` test that races, asserting the post-set entry survives a concurrent get-failure. (Realistic given this backend's `RLock` is exactly there for concurrent connections.)
- A more realistic failure mode — feeding `RustLiveView.deserialize_msgpack` truncated bytes via direct dict mutation — would also catch the case where `serialize_msgpack` fails (the current patch only makes deserialize fail; serialize-side failures take the same `except Exception` arm but aren't tested).

### 🟢 nice-to-have — also pop `_state_sizes`

`python/djust/state_backends/memory.py:133`

`delete()` pops both `_cache[key]` and `_state_sizes[key]`. The new discard path only pops `_cache`. Minor (memory-stats only, not correctness), but folding it into the `is`-guarded block above is one extra line.

### Other checks

- Auto-reject scan: clean. `%s`-style logging at L134-138 (good — was `%s` before too). No `print()`, no bare `except: pass`, comments justify the broad `except Exception`.
- CHANGELOG entry at `CHANGELOG.md:10` accurately describes the shipped behavior (pop + None) and the failure mode (shared ref → cross-session leak after `cargo build`).
- No new external deps.

### Verdict

The fix direction is right and the bug it closes is real and serious. The TOCTOU in the discard path is a narrow window but lives inside the exact concurrency regime this backend was hardened for in #1353 — worth tightening before merge.

REQUEST_CHANGES 2

johnrtipton:
## Pipeline-run retro — PR #1438 (#1410)

### Quality: 4/5
Stage 11 caught a real TOCTOU between the unlocked round-trip and the relocked discard. Stage 7+8 self-review missed it because the original review focused on the bug-fix correctness, not the new locking pattern. The fix shipped clean once the reviewer surfaced the gap.

### Findings caught (and the stage that caught them)
- 🟡 (Stage 11) TOCTOU between unlock and relock — concurrent `set()` could land a fresh entry that the unconditional `pop` would then delete. Identity-guard added.
- 🟢 (Stage 11) `_state_sizes[key]` not popped in the discard path. Now matches `delete()` semantics.

### Process notes
- The bug-fix shape ("fail-closed by popping the corrupt entry") was correct on first try. The miss was on the lock-window arithmetic — same class of failure that retro Action #1198 (commit-or-rollback handler shape) targets, but applied to lock-release/lock-reacquire windows rather than mutate/await sequences. Worth filing a follow-up canon item if this pattern recurs.
- Reviewer prompt explicitly asked about TOCTOU. That direction was load-bearing — the reviewer's first instinct would otherwise have been to verify the bug-fix correctness, not the new lock semantics.

### Two-commit shape adherence
Followed cleanly: `482287ba` (impl + test) → `c0e15328` (CHANGELOG). Stage-11 fix-pass `f3638714` updated impl + test together, which is correct (the test is the regression evidence for the new fix).

🤖 Generated by `/pipeline-run` Stage 14

PR #1439, opened 2026-05-09
docs(template-cheatsheet): note Django {%-in-comment gotcha (#1423)

### LINKED ISSUE #1423: Django template parser rejects {# ... {% if %} ... #} comments with partial tag syntax
## Context

While working on the NYC Claims Processing project (uses djust for LiveView templates), hit a Django `TemplateSyntaxError: Unexpected end of expression in if tag` when a template comment block contained the literal string `{% if %}` as prose inside a `{# ... #}` comment.

The Django comment parser normally treats `{# ... #}` as opaque, but when the comment text contains a partial `{% %}` tag (e.g., `{# d-none (not {% if %}) so that ... #}`), the tokenizer appears to split the comment at the embedded tag boundary, then fails when the following content doesn't close the `{% if %}` it half-parsed.

## Minimal repro

```django
{# d-none (not {% if %}) so the VDOM ... #}
<div class="foo {% if x is None %}d-none{% endif %}"></div>
```

Renders raise:
```
django.template.exceptions.TemplateSyntaxError: Unexpected end of expression in if tag.
  File ".../django/template/smartif.py", line 149, in nud
    raise parser.error_class("Unexpected end of expression in if tag.")
```

Workaround: rewrite the comment to not contain `{%` / `%}`:
```django
{# d-none keeps the DOM stable so the VDOM toggles a class attribute ... #}
```

## Question

Is this a djust-side tokenizer bug (Rust renderer), or pure upstream Django behavior that djust just inherits? If djust, worth tracking here. If pure Django, at least worth documenting as a djust-aware template gotcha — projects using djust will hit it when they document VDOM patterns in comments.

## Reproduction project

- Repo: github.com/flexion/flexion-nyc-claims (private)
- PR where it was hit: #282
- File: `templates/claims/claim_detail.html` around line 1260

Environment: Django 5.x, djust `>=0.9.5rc4`, Python 3.12.

johnrtipton:
## Summary

Closes #1423 as documentation-only.

The reporter hit a Django `TemplateSyntaxError` from a `{# d-none (not {% if %}) so the VDOM ... #}` comment. **Investigation finding**: djust's Rust template engine handles `{# … #}` correctly — the lexer at `crates/djust_templates/src/lexer.rs:289-304` skips opaquely until `#}` regardless of contents. Verified locally:

```python
>>> render_template('{# (not {% if %}) #}<div></div>', {})
'<div></div>'   # OK
```

The failure surfaces only when a template flows through **upstream Django**'s tokenizer (non-LiveView HTTP path, third-party doc tools that re-parse, etc.). djust can't fix Django's parser. Best we can do is warn about the trip wire.

This PR adds a Gotchas section to the template cheatsheet documenting:
- The failure mode (Django's `smartif.py: Unexpected end of expression in if tag`)
- A trivial workaround (rewrite the comment without `{%` / `%}`)
- Reference back to #1423

## Test plan

- [x] Read the rendered guide locally to verify formatting.
- [ ] Maintainer confirms the framing is right (i.e., we don't intend to "fix" Django).

### DIFF
--- a/docs/website/guides/template-cheatsheet.md
+++ b/docs/website/guides/template-cheatsheet.md
@@ -841,3 +841,16 @@ Data passing:
   dj-value-*                      (extra value kwargs)
   dj-target="#selector"           (scoped DOM updates)
 ```
+
+## Gotchas
+
+### Don't put `{%` or `%}` inside `{# … #}` comments
+
+djust's Rust template engine handles this correctly — it treats `{# … #}` as opaque. **Django's stock template parser does not.** When a template flows through Django (e.g., the non-LiveView HTTP path, or any third-party tool that re-parses your templates), a comment that contains a partial tag string will trip Django's tokenizer:
+
+```django
+{# d-none (not {% if %}) so the VDOM ... #}        <!-- ❌ Django will choke -->
+{# d-none keeps the DOM stable so the VDOM ... #}  <!-- ✅ both engines OK -->
+```
+
+The Django error is `TemplateSyntaxError: Unexpected end of expression in if tag` from `django/template/smartif.py`. Workaround: rewrite the comment without `{%` / `%}`. (Reference: [#1423](https://github.com/djust-org/djust/issues/1423).)


johnrtipton:
## Pipeline-run retro — PR #1439 (#1423)

### Quality: 5/5
Investigation-class issue resolved cleanly without code change. The reporter explicitly asked "is this a djust bug or pure Django?" — direct verification (running the failing template through `djust._rust.render_template`) cleared djust's tokenizer in seconds. The right resolution was documentation, not a fix.

### Process notes
- **Reproducer-first paid off (Action #1210 / v0.9.5 retro).** Verified djust's Rust lexer at `lexer.rs:289-304` skips opaquely until `#}` — no need to spelunk the issue further.
- **Close-without-code is a real pipeline-run path** (per skill canon). Took ~10 min wall vs ~1 hr for a no-op code PR.

PR #1440, opened 2026-05-09
feat(checks): D001 — psycopg3-required system check (#1433)

### LINKED ISSUE #1433: Add system check for psycopg2-without-psycopg3 misconfiguration
# Add a system check that warns when psycopg2 is installed but psycopg3 isn't

## Summary

The 2026-05-05 incident that drove PR #1357 was caused by a deploy
shipping `psycopg2-binary` but missing `psycopg[binary]>=3.2`.
djust's `db.notifications` module silently runs as a no-op (pre-0.9.5
it retried forever; 0.9.5 made it permanent-fail with a WARNING at
first connect attempt).

The "permanent-fail at first connect" behaviour from #1357 is the
right safety net to keep — but it only fires when something tries to
USE `db.notifications`. Operators who deploy the misconfiguration but
don't have any `@notify_on_save` consumer actively running won't see
the warning until much later.

This ticket: surface the misconfiguration at **`manage.py check`**
time, before the application starts serving traffic. Same pattern
djust already uses for other deployment-readiness checks
(see `djust/checks.py`).

## Proposed check

`djust.checks.E1XX` (next free slot, e.g. E190):

```
djust.E190: Postgres LISTEN/NOTIFY (db.notifications) is unavailable
because psycopg[binary]>=3.2 is not installed.

  Detected: psycopg2 (legacy driver) at version 2.9.x.
  Required: psycopg[binary] (psycopg3) at version >= 3.2.

  HINT: pip install 'psycopg[binary]>=3.2'
        See: https://djust.org/docs/db/notifications#installation

  This is a WARNING (not an error) — non-Postgres apps and apps
  that don't use @notify_on_save / db.listen() are unaffected.
  Apps that DO use those features will hit a permanent-failure
  WARNING in logs at first NOTIFY attempt (see djust#1357).
```

Trigger condition:

- `DATABASES['default']['ENGINE']` is `django.db.backends.postgresql`
  AND
- `psycopg2` is importable (the legacy driver is installed)
  AND
- `psycopg` (psycopg3) is NOT importable, OR is importable but
  `psycopg.__version__ < "3.2"`

If the engine is sqlite, mysql, or anything else — no warning.
If neither psycopg is installed — Django itself raises a different
`E001`; we don't duplicate.

## Why a check (not a stronger gate)

Django system checks are the right venue:

1. They run automatically on `manage.py runserver`, `migrate`,
   `runworker`, etc. — every operator path that touches the DB
   layer surfaces the warning at startup.
2. They DON'T block deployment when the operator chose not to use
   `db.notifications` — many djust apps just use the ORM. Making it
   an error would be hostile.
3. They're silenceable per-project (`SILENCED_SYSTEM_CHECKS = ["djust.E190"]`)
   for the rare team that wants to defer the migration.

The existing #1357 permanent-failure handling stays as the
last-line-of-defense for anyone who ignores the warning.

## Implementation sketch

`python/djust/checks.py` — append to the existing checks module.
The check function: import `psycopg2`, `psycopg`; do the
version-and-engine match; emit a `Warning` with `id="djust.E190"`.

A small unit test in `python/djust/tests/test_checks.py` per the
existing pattern: 4 cases — sqlite (no warn), pg+psycopg3 (no warn),
pg+psycopg2-only (warn), pg+psycopg3-too-old (warn).

## Documentation

- `docs/db/notifications.md` — add an "Installation" section that
  explicitly calls out `psycopg[binary]>=3.2` as the required driver.
- `CHANGELOG.md` — note the new check under the next version's
  `### Added`.

## Out of scope

- Removing the existing #1357 permanent-failure handling — keep it as
  a safety net.
- Forcing psycopg3 as a hard install dep on djust itself — psycopg
  is correctly an *optional* dep today (only `db.notifications` users
  need it). Keeping it that way preserves the tenant-of-many-apps use
  case where some apps don't need Postgres at all.
- Adding a similar check for psycopg3-but-using-async-ORM — that's a
  Django-level concern, not a djust one.

## References

- PR #1357 — original incident write-up + permanent-failure fix
- djust-org/djustlive psycopg2→3 migration ticket (filed alongside
  this one — same root cause, same fix path)

johnrtipton:
## Summary

Closes #1433.

New Django system check `djust.D001` (first in a new "D" / Database-driver category) that warns when:

1. `DATABASES['default']['ENGINE']` is `django.db.backends.postgresql`, AND
2. `psycopg2` (legacy driver) is importable, AND
3. `psycopg` (3.x) is NOT importable, OR is at version < 3.2

This complements the runtime permanent-fail-on-first-NOTIFY behaviour from #1357 — that stays as the last-line-of-defense, but D001 surfaces the misconfig at `manage.py check` / `runserver` startup so operators see the problem before traffic, even if no `@notify_on_save` consumer is currently running.

## Test plan

- [x] 8 new tests in `python/tests/test_checks.py`:
  - `TestD001PsycopgVersionCheck`: sqlite (no warn), pg+3.2 (no warn), pg+missing (warn), pg+too-old (warn).
  - `TestParsePsycopgVersion`: standard, two-part, pre-release-tail, unparseable inputs.
- [x] CHANGELOG `[Unreleased] ### Added` entry.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- **System check `djust.D001` — warn when Postgres is configured but `psycopg[binary]>=3.2` is not installed (#1433).** djust's `db.notifications` (LISTEN/NOTIFY bridge) requires psycopg3. The 0.9.5 cycle hardened the runtime path to permanent-fail with a WARNING when `@notify_on_save` actually fires (#1357), but operators who deploy the misconfig without an active consumer wouldn't see the warning until much later. D001 surfaces it at `manage.py check` / `runserver` startup, before traffic. Fires only when the default DB engine is Postgres AND `psycopg2` is importable AND `psycopg` (3.x) is missing or at version < 3.2. Silenceable per-project via `SILENCED_SYSTEM_CHECKS = ['djust.D001']`.
+
 ### Fixed
 
 - **`InMemoryStateBackend.get()` discards corrupt entries instead of returning the shared in-memory ref (#1410).** When `RustLiveView.deserialize_msgpack` raised — typically after a hot-swap struct change or msgpack schema drift — the previous fallback returned the cached object directly. Two concurrent connections to the same view then shared one `_rust_view`, and mutations from connection A leaked into connection B's render context. After a `cargo build` of `djust_vdom` + `.so` swap, fresh navigations could re-render with state from the prior session. Now the backend pops the corrupt entry from its in-memory dict and returns `None`; the caller's mount path treats the cache as cold and runs `mount()` cleanly. Discovered during the #1408 investigation.
--- a/python/djust/checks.py
+++ b/python/djust/checks.py
@@ -2977,3 +2977,133 @@ def check_admin_widgets(app_configs, _admin_sites=None, **kwargs):
             )
 
     return results
+
+
+# ---------------------------------------------------------------------------
+# Database-driver checks (D0xx)
+# ---------------------------------------------------------------------------
+
+
+def _parse_psycopg_version(version_str: str) -> tuple:
+    """Parse a `major.minor[.patch[...]]` version string into a tuple of ints.
+
+    Returns ``(0, 0)`` on anything we can't parse — the caller treats
+    that as "couldn't determine version, don't make claims about it".
+    """
+    parts = []
+    for chunk in version_str.split(".")[:2]:
+        digits = ""
+        for ch in chunk:
+            if ch.isdigit():
+                digits += ch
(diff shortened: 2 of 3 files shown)


johnrtipton:
## Pipeline-run retro — PR #1440 (#1433)

### Quality: 5/5
First-time-right shape. New 'D' check category opened cleanly. 8 tests cover all 4 trigger combinations (sqlite, pg+3.2, pg+missing, pg+old) plus 4 unit tests for the version parser. Two-commit shape clean (impl+tests, then CHANGELOG).

### Process notes
- Used `patch.dict(sys.modules, {'psycopg': None})` for the missing-import test — that's the canonical pattern; surfaced it as a recipe for future check-style tests.
- Mirrored the existing-checks shape (DjustWarning + register decorator + suppression-aware short-circuit) without inventing new conventions.

Issue #1437, opened 2026-05-09
Cache theme_context output by (preset, pack, mode, locale) tuple


johnrtipton:
# Cache `theme_context` output by `(preset, pack, mode)` tuple

## Summary

`djust.theming.context_processors.theme_context` runs as a Django
context processor on **every request that touches a template** (so:
~every request). Today it does, fresh, per request:

```python
def theme_context(request):
    manager = get_theme_manager(request)
    state = manager.get_state()
    css = generate_css_for_state(state, css_prefix=get_css_prefix())
    anti_fouc_script = "<script>...</script>"
    js_url = staticfiles_storage.url("djust_theming/js/theme.js")
    theme_head = f"{anti_fouc_script}\n<style data-djust-theme>{css}</style>\n<script src='{js_url}' defer></script>"
    theme_switcher = _render_theme_switcher(state, manager.get_available_presets())
    return {
        "theme_head": mark_safe(theme_head),
        "theme_switcher": mark_safe(theme_switcher),
        ...
    }
```

The expensive parts:

1. `manager.get_state()` — cookie + setting reads + state object construction
2. `generate_css_for_state(state)` — full CSS variable-block string generation (the largest chunk; thousands of bytes of CSS rebuilt every request)
3. `_render_theme_switcher(state, presets)` — HTML string construction over the available-presets list (~60 entries)
4. `manager.get_available_presets()` — manifest lookup

The output **only depends on a few inputs**: `(preset, pack, mode, locale)` — at most a few hundred unique combinations across djust's whole 60+ theme catalog. Everything else (`request.user`, headers, path, query string) is **not** an input to the rendered HTML.

## Cache shape

```python
from functools import lru_cache

@lru_cache(maxsize=512)
def _render_theme_outputs(preset: str, pack: str, mode: str, locale: str | None) -> tuple[str, str]:
    """Returns (theme_head_html, theme_switcher_html) for the given state."""
    state = ThemeState(preset=preset, pack=pack, mode=mode, locale=locale)
    css = generate_css_for_state(state, css_prefix=get_css_prefix())
    js_url = staticfiles_storage.url("djust_theming/js/theme.js")
    theme_head = f"{ANTI_FOUC_SCRIPT}\n<style data-djust-theme>{css}</style>\n<script src='{js_url}' defer></script>"
    theme_switcher = _render_theme_switcher(state, _cached_available_presets())
    return theme_head, theme_switcher


def theme_context(request):
    manager = get_theme_manager(request)
[...]

PR #1441, opened 2026-05-09
perf(tenants): TenantMiddleware short-circuits when unconfigured (#1436)

### LINKED ISSUE #1436: TenantMiddleware should short-circuit when no resolver is configured
# `TenantMiddleware` should short-circuit when no resolver is configured

## Summary

`djust.tenants.middleware.TenantMiddleware` runs unconditionally on
every request. When a consumer has the `djust[tenants]` extra
installed but hasn't opted into tenant resolution
(`DJUST_CONFIG['TENANT_RESOLVER']` not set), the middleware still:

1. Calls `self.resolver.resolve(request)` — the default resolver
   (`SubdomainResolver`) does `request.get_host().split(":")` +
   `host.split(".")` + several settings reads on every request.
2. Sets `request.tenant = None` (the typical outcome).
3. Sets thread-local storage, then clears it after `get_response`.
4. Reads two settings keys (`DJUST_TENANTS` and
   `DJUST_CONFIG.TENANT_REQUIRED`) to evaluate the required-tenant
   gate.

For consumers who never use tenant features (single-tenant
deployments, the `djust-scaffold` starter template, demo apps), this
is pure overhead. A py-spy profile of `djust-scaffold`'s `/`
endpoint at 250 Hz over 8 s puts ~30% of stack samples in the
TenantMiddleware path — though most of that is the inner `get_response`
call rather than incremental middleware cost. The actual incremental
work is closer to **2–5% of per-request CPU**, plus the thread-local
set/clear pair on every request.

## Proposed change

Detect the "not configured" case in `__init__` and switch to a
no-op `__call__` when applicable. The check is cheap and runs once
per process boot.

```python
class TenantMiddleware:
    def __init__(self, get_response):
        self.get_response = get_response
        config = getattr(settings, "DJUST_CONFIG", {})
        tenants_config = getattr(settings, "DJUST_TENANTS", {})

        # Short-circuit: if no resolver is explicitly configured
        # AND no DJUST_TENANTS settings are set, the middleware is
        # effectively a no-op. Skip the resolver setup + per-request
        # work entirely.
        if (
            "TENANT_RESOLVER" not in config
            and not tenants_config
        ):
            self._enabled = False
            self.resolver = None
            return

        self._enabled = True
        self.resolver = get_tenant_resolver()

    def __call__(self, request):
        if not self._enabled:
            return self.get_response(request)
        # ... existing resolver-call + thread-local set/clear ...
```

## Why this is safe

- **Existing consumers with explicit configuration are unaffected** —
  the short-circuit only fires when `DJUST_CONFIG['TENANT_RESOLVER']`
  is missing AND `DJUST_TENANTS` is absent or empty. Any consumer
  who has explicitly set either keeps the existing behavior.
- **Consumers relying on `request.tenant` being defined** — there's
  a small backwards-compat concern. If user code does
  `request.tenant.id`, that crashes whether the middleware is no-op'd
  or is running with no resolver (both leave `request.tenant` unset
  / `None`). Mitigation: when no-op'd, optionally still set
  `request.tenant = None` so the attribute exists. One-line
  preservation of the existing surface.
- **`get_current_tenant()` callers** — pull from thread-local; the
  short-circuit doesn't set it, but `get_current_tenant()` already
  returns `None` when nothing is set, so behavior is preserved.

## Downstream impact

`djust-org/djust-scaffold#TBD` (filed alongside this) makes the
scaffold's `MIDDLEWARE.append(TenantMiddleware)` explicitly opt-in.
With both changes, a consumer using djust without tenant features
pays zero per-request cost from this middleware.

## Out of scope

- Refactoring `SubdomainResolver` to be cheaper. Its work is small
  (~µs per request); the win comes from skipping it entirely when
  not needed, not from speeding it up.
- Caching the resolved tenant per (host, path) pair. That's a
  separate ticket if a real-world workload shows resolver cost
  dominating.

## Measurement

Before / after on the Phase-11 scaffold benchmark:
- Current: 31 r/s on `/` at concurrency 16
- Expected after this + downstream scaffold change: 33–34 r/s

Small but real. The bigger picture is **principle of zero cost
for unused features** — same pattern as `djust[postgres]` shouldn't
pay psycopg2 install cost when sqlite is in use, and `djust[theming]`
shouldn't pay the `_render_theme_tag` slow path (#1435).

## References

- djust-org/djust-scaffold#TBD — matching downstream cleanup
  (make `MIDDLEWARE.append(TenantMiddleware)` opt-in instead of
  auto-installed)
- Phase-11 hot-spot audit profile data (host: `/tmp/render.txt`)

johnrtipton:
## Summary

Closes #1436.

When neither `DJUST_CONFIG['TENANT_RESOLVER']` nor `DJUST_TENANTS` is configured, `TenantMiddleware` was still calling `SubdomainResolver.resolve()`, setting+clearing a thread-local, and reading two settings keys on every request. For consumers with `djust[tenants]` installed but no opt-in (single-tenant deploys, scaffold starters, demo apps), that's pure overhead.

Detect the unconfigured shape once in `__init__` and switch `__call__` to a straight passthrough that sets `request.tenant = None` and calls `get_response`.

## Backward compatibility

- Any `DJUST_CONFIG['TENANT_RESOLVER']` value OR any non-empty `DJUST_TENANTS` keeps the full path. The short-circuit is strictly opt-out by virtue of being unset.
- `request.tenant` attribute existence preserved on the no-op path.
- Thread-local helpers (`get_current_tenant()` / `set_current_tenant()`) unchanged.

## Test plan

- [x] 4 new tests in `TestTenantMiddlewareShortCircuit`:
  - `short_circuits_when_unconfigured` — both empty → `_enabled=False`, resolver=None, thread-local untouched
  - `does_not_short_circuit_with_djust_config` — full path runs
  - `does_not_short_circuit_with_djust_tenants` — full path runs
  - `short_circuit_does_not_set_thread_local` — verifies the savings via mock
- [x] All 6 existing tenant-middleware tests still pass.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Performance
+
+- **`TenantMiddleware` short-circuits when no resolver is configured (#1436).** When neither `DJUST_CONFIG['TENANT_RESOLVER']` nor `DJUST_TENANTS` is set, the middleware now bypasses the resolver call, the thread-local set/clear pair, and the required-tenant gate — switching `__call__` to a straight `get_response(request)` passthrough. Saves ~2-5% per-request CPU for consumers with `djust[tenants]` installed but no tenant opt-in (single-tenant deploys, scaffold starters, demo apps). Consumers who set either config keep the full path; `request.tenant` is still set to `None` on the no-op path so `getattr(request, "tenant", None)` callers see the same shape.
+
 ### Added
 
 - **System check `djust.D001` — warn when Postgres is configured but `psycopg[binary]>=3.2` is not installed (#1433).** djust's `db.notifications` (LISTEN/NOTIFY bridge) requires psycopg3. The 0.9.5 cycle hardened the runtime path to permanent-fail with a WARNING when `@notify_on_save` actually fires (#1357), but operators who deploy the misconfig without an active consumer wouldn't see the warning until much later. D001 surfaces it at `manage.py check` / `runserver` startup, before traffic. Fires only when the default DB engine is Postgres AND `psycopg2` is importable AND `psycopg` (3.x) is missing or at version < 3.2. Silenceable per-project via `SILENCED_SYSTEM_CHECKS = ['djust.D001']`.
--- a/python/djust/tenants/middleware.py
+++ b/python/djust/tenants/middleware.py
@@ -52,9 +52,37 @@ class TenantMiddleware:
 
     def __init__(self, get_response):
         self.get_response = get_response
-        self.resolver = get_tenant_resolver()
+
+        # Short-circuit: when neither DJUST_CONFIG['TENANT_RESOLVER'] nor
+        # DJUST_TENANTS is configured, the middleware is effectively a
+        # no-op — it would call SubdomainResolver, get back None, set
+        # request.tenant=None, set+clear a thread-local, and run the
+        # required-gate (which is False by default). For consumers who
+        # have djust[tenants] installed but never opted in (single-tenant
+        # deploys, scaffold starters, demo apps), that's pure overhead
+        # on every request. Detect once at boot, switch __call__ to a
+        # straight passthrough.
+        from django.conf import settings
+
+        djust_config = getattr(settings, "DJUST_CONFIG", {}) or {}
+        djust_tenants = getattr(settings, "DJUST_TENANTS", {}) or {}
+        self._enabled = "TENANT_RESOLVER" in djust_config or bool(djust_tenants)
+
+        if self._enabled:
+            self.resolver = get_tenant_resolver()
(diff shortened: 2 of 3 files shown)


johnrtipton:
## Pipeline-run retro — PR #1441 (#1436)

### Quality: 5/5
Surgical perf change. Detect-once-in-init pattern is the right shape — zero per-request branching on the hot path. 4 regression tests cover the two enable-paths (DJUST_CONFIG vs DJUST_TENANTS) and the savings (mocked thread-local set call count).

### Process notes
- Backwards-compatibility was the load-bearing design constraint: `request.tenant` attribute existence preserved on the no-op path so `getattr(request, "tenant", None)` keeps working. Any consumer doing direct `request.tenant.id` was already broken regardless.
- Generalizable pattern: zero-cost-when-unused for optional-feature middleware. Same shape works for #1437 (theme cache) and could apply to other djust extras.

PR #1442, opened 2026-05-09
perf(theming): cache theme_context output by ThemeState tuple (#1437)

### LINKED ISSUE #1437: Cache theme_context output by (preset, pack, mode, locale) tuple

johnrtipton:
## Summary

Closes #1437.

`theme_context` ran the CSS-generation + theme-switcher HTML pipeline on every templating request. The output is byte-identical for a given `ThemeState` tuple — pure function of state, no request data flows in. Wrap with `@lru_cache(maxsize=512)`.

Per-request cost: ~1-3 ms → ~5-10 µs (dict lookup + tuple unpack) post-warmup. djust's whole catalog of ~60 presets × 2 modes × handful of packs fits well under the cache size.

## Cache key shape

`(theme, preset, pack, mode, resolved_mode, layout, presets_key)`

The available-presets list (unhashable list of dicts) is normalized to a sorted tuple-of-tuples via `_presets_to_cache_key`, keeping only the fields the switcher consumes (`name`, `display_name`, `is_active`).

## Invalidation

`djust.theming.context_processors.clear_theme_context_cache()` is exposed for:
- Theme-pack hot-reload (when the manifest changes)
- Tests that mutate theme state between assertions

## Test plan

- [x] 7 regression tests in `test_theming_context_cache.py`:
  - Repeat call → cache hit (CSS-generation runs once)
  - Different preset / mode / pack / presets list → cache miss
  - `clear_theme_context_cache()` drops warm state
  - **`request` object does NOT leak into cached output** — different `request.user`, `request.path` with same state → byte-identical bytes
- [x] All 42 existing theming tests still pass.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Performance
 
+- **`theme_context` now caches its rendered output by `ThemeState` tuple (#1437).** The Django context processor `djust.theming.context_processors.theme_context` previously ran the full CSS-generation + theme-switcher HTML pipeline on every templating request. Output is now `lru_cache(maxsize=512)`'d on `(theme, preset, pack, mode, resolved_mode, layout, presets_key)` — a pure function of state, no request data flows in. djust's catalog of ~60 presets × 2 modes × handful of packs fits comfortably under the cache size. Per-request cost: ~1-3 ms → ~5-10 µs post-warmup. New `djust.theming.context_processors.clear_theme_context_cache()` exposed for theme-pack hot-reload and tests.
 - **`TenantMiddleware` short-circuits when no resolver is configured (#1436).** When neither `DJUST_CONFIG['TENANT_RESOLVER']` nor `DJUST_TENANTS` is set, the middleware now bypasses the resolver call, the thread-local set/clear pair, and the required-tenant gate — switching `__call__` to a straight `get_response(request)` passthrough. Saves ~2-5% per-request CPU for consumers with `djust[tenants]` installed but no tenant opt-in (single-tenant deploys, scaffold starters, demo apps). Consumers who set either config keep the full path; `request.tenant` is still set to `None` on the no-op path so `getattr(request, "tenant", None)` callers see the same shape.
 
 ### Added
--- a/python/djust/tests/test_theming_context_cache.py
+++ b/python/djust/tests/test_theming_context_cache.py
@@ -1,0 +1,182 @@
+"""Tests for the theme_context per-process cache (#1437)."""
+
+from unittest.mock import MagicMock, patch
+
+import pytest
+
+
+@pytest.fixture(autouse=True)
+def _clear_cache_per_test():
+    """Each test starts with a clean cache so warm-state from a prior
+    test doesn't mask cold-path bugs."""
+    from djust.theming.context_processors import clear_theme_context_cache
+
+    clear_theme_context_cache()
+    yield
+    clear_theme_context_cache()
+
+
+def _make_manager(preset="default", mode="light", resolved_mode="light", pack=None, presets=None):
+    """Build a stub theme manager whose `get_state()` returns a
+    ThemeState with the given fields and `get_available_presets()`
+    returns the given preset list."""
+    from djust.theming.manager import ThemeState
+
+    state = ThemeState(
+        theme="default",
+        preset=preset,
+        mode=mode,
+        resolved_mode=resolved_mode,
+        pack=pack,
+    )
+    if presets is None:
+        presets = [
+            {"name": "default", "display_name": "Default", "is_active": preset == "default"},
+            {"name": "ocean", "display_name": "Ocean", "is_active": preset == "ocean"},
+        ]
+    mgr = MagicMock()
+    mgr.get_state.return_value = state
+    mgr.get_available_presets.return_value = presets
+    return mgr
+
+
+class TestThemeContextCache:
+    def test_two_identical_calls_render_once(self):
+        """Same (preset, pack, mode, resolved_mode, presets) on two
+        calls → CSS generation runs exactly once. Cache hit on the
+        second call."""
+        from djust.theming.context_processors import theme_context
+
+        mgr = _make_manager()
+        with (
+            patch("djust.theming.context_processors.get_theme_manager", return_value=mgr),
+            patch(
+                "djust.theming.context_processors.generate_css_for_state",
+                return_value=":root { --color: blue; }",
+            ) as mock_css,
+        ):
+            ctx1 = theme_context(MagicMock())
+            ctx2 = theme_context(MagicMock())
+        assert mock_css.call_count == 1, (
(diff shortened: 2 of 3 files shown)


johnrtipton:
## Pipeline-run retro — PR #1442 (#1437)

### Quality: 5/5
Cache shape clean, key derivation correct (full ThemeState tuple including `theme` and `layout` fields I'd missed on first pass — caught by my own test failure), invalidation hook exposed.

### Process notes
- **Initial cache key was incomplete**: I used (preset, pack, mode, resolved_mode, presets) but ThemeState also has `theme` and `layout`. Test failure surfaced it immediately. Lesson: when caching by 'a struct', include EVERY field of the struct in the key, then prune later if profiling shows specific fields don't matter.
- **request-object-doesn't-leak test was the load-bearing one** — the entire safety claim of the cache rests on that property. Without it, two users with same theme but different requests could swap state-dependent data. Asserted directly.

PR #1443, opened 2026-05-09
perf(theming): pre-render theme_panel / theme_mode_toggle / theme_preset_selector (#1435)

### LINKED ISSUE #1435: Pre-render theme_panel/theme_mode_toggle/theme_preset_selector as context strings
# Pre-render `theme_panel` / `theme_mode_toggle` / `theme_preset_selector` as context strings

## Summary

`djust.theming.context_processors.theme_context` currently injects two
of the theme outputs as pre-rendered safe strings:

```python
return {
    "theme_head": mark_safe(theme_head),
    "theme_switcher": mark_safe(theme_switcher),
    ...
}
```

Templates can use these as `{{ theme_head }}` / `{{ theme_switcher }}`
— **zero per-render cost in the template engine**, since the heavy
work (CSS variable computation, theme.js script tag, manifest lookup)
happens once per request in the context-processor.

The other theme tags — `{% theme_panel %}`, `{% theme_mode_toggle %}`,
`{% theme_preset_selector %}` — are **only available as Django
template tags** (defined in `djust.theming.templatetags.theme_tags`).
There's no pre-rendered context-string equivalent.

## Why this matters

Consumers using djust's Rust template engine (the default) must
register **Python tag handlers** (`djust.template_tags.TagHandler`)
for any `{% theme_panel %}` / `{% theme_mode_toggle %}` /
`{% theme_preset_selector %}` they want to use. Those handlers
typically re-render through Django's classic template engine just to
get the HTML back as a string. That's the pattern in
`djust-scaffold`'s `apps/shared/theming_handlers.py`:

```python
def _render_theme_tag(tag_str: str, context: Dict[str, Any]) -> str:
    from django.template import Context, Template
    tpl = "{% load theme_tags %}" + tag_str
    mock_request = _make_mock_request(context)
    return Template(tpl).render(Context({"request": mock_request}))
```

This is **expensive**:

- Django template parse + render per call
- A mock-request synthesis on every render (because the Rust engine
  strips `HttpRequest` from context before passing to handlers)
- Full Django template-tag library load
- The output is a string; nothing about it requires the Django
  template engine

A py-spy profile of the scaffold's `/` endpoint at 250 Hz over 8 s
(see Phase-11 hot-spot audit, 2026-05-08) shows `_render_theme_tag`
+ Django template-engine machinery accounting for **35-45% of CPU
samples on the render path**. The base template uses `{% theme_head %}`
once + `{% theme_panel %}` twice — three slow-path renders per page.

## Proposed change

Extend `djust.theming.context_processors.theme_context` to pre-render
the remaining tags as safe strings, mirroring the existing
`theme_head` / `theme_switcher` pattern. The implementation is
straightforward — these tags don't depend on per-template arguments
in the common case; they take the same theme manager state the
existing two pre-renders consume.

```diff
 def theme_context(request):
     manager = get_theme_manager()
     state = manager.resolve_state(request)
     ...
     theme_head = f"""..."""
     theme_switcher = _render_theme_switcher(state, manager.get_available_presets())
+    theme_panel = _render_theme_panel(state, manager.get_available_presets())
+    theme_mode_toggle = _render_theme_mode_toggle(state)
+    theme_preset_selector = _render_theme_preset_selector(state, manager.get_available_presets())

     return {
         "theme_head": mark_safe(theme_head),
         "theme_switcher": mark_safe(theme_switcher),
+        "theme_panel": mark_safe(theme_panel),
+        "theme_mode_toggle": mark_safe(theme_mode_toggle),
+        "theme_preset_selector": mark_safe(theme_preset_selector),
         ...
     }
```

The actual rendering helpers (`_render_theme_panel` etc.) already
exist in some form inside `djust.theming.templatetags.theme_tags` —
the work is to extract them so they can be called from the
context-processor with the already-resolved state, instead of being
called from a `simple_tag` that re-resolves state per template tag.

## Compatibility

- **Existing `{% theme_head %}` template-tag form continues to work** —
  this is purely additive.
- **The Django template-tag form was never the fast path even on
  djust's side** — Django renders the tag normally; the inefficiency
  is on the Rust template-engine side where unknown tags get
  re-bridged via Python.
- **Customizing per-call (`{% theme_panel layout="grid" %}`)** — for
  the rare consumer that varies per-page, the template-tag form
  remains available. The context-string form covers the 90% case
  where the panel renders the same on every page in the project.

## Alternative (rejected): native Rust tag handlers

Could also implement these tags **inside the Rust template engine**
so `{% theme_panel %}` itself is fast. Rejected for two reasons:

1. The output depends on Python-side data — theme-pack manifests,
   color palettes, available presets — that lives in
   `djust.theming.manager`. Moving that to Rust would be a much
   larger refactor.
2. The context-string approach already extracts all the
   per-template-tag work into per-request work, which is the same
   amortization that makes the Rust-side fast-path fast. Equivalent
   speed, far smaller change.

## Out of scope

- Rewriting `djust.theming.manager` in Rust.
- Removing the existing Django template-tag definitions
  (`{% theme_panel %}` etc.) — they remain for backwards compat and
  for users running djust without the Rust engine.
- The downstream scaffold cleanup
  (djust-org/djust-scaffold#TBD) — separate ticket; switching
  scaffold's `templates/base.html` from `{% theme_panel %}` to
  `{{ theme_panel }}` is blocked on this landing.

## Measurement

Before / after comparison on the scaffold's `/` endpoint (already
captured for "before" in the Phase-11 audit): `_render_theme_tag`
samples should drop to ~zero, and total CPU spent in
`django/template/base.py` should fall by 35–45%. The corresponding
ab throughput at concurrency 16 should improve commensurately
(currently 31 r/s; expectation ~50+ r/s).

## References

- djust-org/djust#1432 — declare djust._rust free-threaded-safe
  (related: same theme of "lift Python overhead off the request path")
- djust-org/djust#1434 — replace sync_to_async(Model.objects.X) with
  native async ORM (ditto)
- Phase-11 RESULTS.md "Hot-spot audit" section (host file, not in
  this repo)

johnrtipton:
## Summary

Closes #1435.

Extends `djust.theming.context_processors.theme_context` to pre-render three more theme tags as `mark_safe` context strings, mirroring the existing `theme_head` / `theme_switcher` pattern. Templates can use:

- `{{ theme_panel }}` instead of `{% theme_panel %}`
- `{{ theme_mode_toggle }}` instead of `{% theme_mode_toggle %}`
- `{{ theme_preset_selector }}` instead of `{% theme_preset_selector %}`

Per the issue's Phase-11 audit, djust-scaffold's `base.html` hits `{% theme_panel %}` twice + `{% theme_head %}` once → three slow-path renders per page through the `_render_theme_tag` Django template-engine bridge (~35-45% of CPU samples on the render path). With this change, the heavy work runs once per request in the context processor regardless of how many `{{ … }}` substitutions appear in the template.

## Compatibility

- Existing `{% theme_panel %}` / `{% theme_mode_toggle %}` / `{% theme_preset_selector %}` template tags continue to work — they're how customization-with-args (`{% theme_panel show_packs=False %}`) is supported.
- This change is purely additive: new context-dict keys, no removals.

## Failure handling

If any of the three tag functions raises (broken manifest, missing template, downstream shadowing), the pre-render comes back as an empty string instead of 500-ing the entire request. Asserted in `test_pre_render_failure_does_not_break_request`.

## Test plan

- [x] 2 new tests in `test_theming_context_cache.py`:
  - `test_pre_rendered_panel_keys_present_in_context` — stubbed tag functions return deterministic values; assert all three appear in the context dict.
  - `test_pre_render_failure_does_not_break_request` — tag raises → empty strings, no exception leaks.
- [x] All 7 prior cache tests still pass.


johnrtipton:
## Pipeline-run retro — PR #1443 (#1435)

### Quality: 4/5
Pragmatic shape: call existing tag bodies once per request rather than extracting helpers. Smaller blast radius than the issue's suggested helper extraction. Trade-off: doesn't benefit from the #1437 lru_cache, but the once-per-request shape captures the bulk of the perf win the issue describes.

### Process notes
- The fail-soft path (try/except → empty strings) is the load-bearing safety net. A broken theme manifest now degrades gracefully to empty pre-renders instead of 500-ing every request that touches a templating context. Asserted directly.
- 4/5 not 5/5 because a fuller refactor would extract pure helpers + apply lru_cache for full amortization. Filed as a v0.9.6-2 follow-up consideration in the milestone summary.

Issue #1420, opened 2026-05-07
test(vdom): patch-batch ordering torture (intra-batch handle invalidation)


johnrtipton:
## Background

A single diff round can produce a batch of patches where one patch invalidates another's targeting handle. Example:

```
patches = [
  RemoveChild { d: "parent", child_d: "X" },
  SetAttr    { d: "X", key: "class", value: "..." },  // X is gone after the RemoveChild!
]
```

The client must apply patches in an order where this can't happen. Production handles this via:
- `apply_patches` groups by parent and orders: removes (descending index), inserts (ascending), moves (by id).
- The recent fix `3fec3100` (sort `RemoveSubtree` + `InsertSubtree` BEFORE path-based child ops) addressed a specific manifestation of this in the #1370 family.

## Gap

No torture test enumerates the patch-ordering invariant directly:

> For any batch of patches the diff emits in a single round, every patch's targeting handle MUST be valid at the moment that patch is dispatched (not just at the start of the batch).

Today the invariant is enforced implicitly by the diff engine's emission order + the client's apply-order discipline. A regression in either silently breaks something.

## Scope

New file `crates/djust_vdom/tests/torture_patch_batch_ordering.rs`. Use the in-tree `apply_patches` (which mirrors the client's apply order) and assert no patch references a handle that an earlier-applied patch in the same batch removed. Cover:

1. RemoveChild followed by SetAttr/SetText on the removed node.
2. Replace followed by SetAttr on the (logically-replaced) old node.
3. RemoveSubtree containing a node that another patch in the batch targets.
4. Move followed by SetAttr on the moved node — must succeed.
5. Insert + same-cycle SetAttr on the new node — must succeed.

## Depends on

#1415 (harness extraction).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #1418, opened 2026-05-07
test(vdom): deep-cascade dj-if torture (10+ levels of if/elif/else nesting + edge positions)


johnrtipton:
## Background

The recursive `dj_if_pre_pass_inner` was added in PR #1365 specifically because Stage 11 found that the original element-by-element pairing produced overlapping patches in `if/elif/else` cascades. `tests/test_dj_if_keyed_diff_1358.rs` covers a few hand-crafted cascade cases (e.g., `test_elif_cascade_a_to_b_flip`); the new `tests/torture_round_trip_with_sync.rs` (#1412) covers up to 5 sibling boundaries.

Neither exercises **deep nesting**.

## Scope

New torture file with these scenarios:

1. **Depth 10+ nested dj-if cascades.** A boundary whose body contains another boundary whose body contains another boundary, etc. Each level toggles independently. Stress `dj_if_pre_pass_inner`'s recursion + body-fallthrough logic + the new `sync_ids_dj_if_pre_pass`'s recursion.
2. **Boundary at first child / last child position.** Off-by-one risk in `find_top_level_dj_if_pairs` and `build_excluded_mask`.
3. **Adjacent boundaries with zero non-boundary siblings between them.** `[open-A, close-A, open-B, close-B]` shape; ensure the index math doesn't trip.
4. **Boundaries with only `#text` content (no element).** dj-if body with just text — does the boundary detector still match correctly?
5. **Very wide boundary lists.** 50+ sibling boundaries, each independently toggling — combinatorial pressure on the boundary-pre-pass.

## Acceptance

Same invariant as #1412's torture: every emitted patch's targeting handle (`d` / `child_d` / `ref_d`) must resolve in the client tracker, across a 10+-cycle toggle sequence per scenario.

## Depends on

#1415 (harness extraction).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #1417, opened 2026-05-07
test(vdom): `dj-update="ignore"` × dj-if boundaries × `sync_ids` interaction


johnrtipton:
## Background

Three independently-tested feature classes interact in production templates:

- `dj-update="ignore"` (#1252) — subtree the diff should leave alone (`splice_ignore_subtrees` in `diff::diff_nodes`)
- `<!--dj-if id="…"-->` boundaries (#1358) — keyed conditional subtrees with their own `dj_if_pre_pass`
- `sync_ids` (#1408) — post-diff id sync, now boundary-aware

Each has its own regression-test file (`test_ignore_subtree_invalidation_1252.rs`, `test_dj_if_keyed_diff_1358.rs`, `test_sync_ids_dj_if_1408.rs`). **None test the three-way interaction.**

## Concrete failure-mode hypotheses

1. **Ignore-subtree inside a matched-id dj-if body.** When `dj_if_pre_pass` recurses into the matched body and falls through to element-by-element pairing, does it correctly skip ignored subtrees? Same question for `sync_ids_dj_if_pre_pass`.
2. **Ignore-subtree wrapping a dj-if boundary.** When the parent says "ignore this subtree" but it contains a dj-if conditional that the developer wants to flip, the conditional toggle should be a no-op (subtree is ignored). Are we silently emitting Insert/RemoveSubtree patches that the client ignores? Does that diverge `last_vdom` from client DOM?
3. **dj-if boundary toggling on/off where one branch contains an ignore-subtree and the other doesn't.** Cross-feature consistency: the ignore subtree appears or vanishes with the boundary.

## Scope

New file `crates/djust_vdom/tests/test_ignore_dj_if_sync_interaction.rs` with 4-6 scenarios covering each hypothesis. Use `mod common;` for helpers (depends on #1415).

## Risk

Medium. `splice_ignore_subtrees` runs before the diff and replaces ignore subtrees with their cached HTML version; the dj-if pre-pass operates on the result. If the splice changes child counts or boundary alignment in subtle ways, drift class similar to #1408 is plausible.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #1416, opened 2026-05-07
test(vdom): full HTML round-trip torture (parse → diff → serialize → re-parse stability)


johnrtipton:
## Gap

Existing torture tests cover phases in isolation:
- `tests/torture_test.rs` — `diff` + `apply_patches`
- `tests/fuzz_test.rs` — same with proptest
- Neither exercises `parse_html` → `diff` → `to_html` → `parse_html` round-trip

Production runs the full pipeline every render. A serializer drift, a parser-normalization quirk, or an `Insert/RemoveSubtree`'s `html` payload that the parser doesn't accept the same way it produced will surface here and not in any current test.

## Scope

New file `crates/djust_vdom/tests/torture_html_round_trip.rs`. Scenarios:

1. **Serialize→parse fixpoint**: for an arbitrary VNode `T1`, assert `parse_html(to_html(T1))` is structurally equal to `T1` (modulo HTML5 parser normalizations the tests can opt-in to). Fuzzer-style with proptest.
2. **Patch-html re-parse**: when the diff emits `InsertSubtree { html, .. }`, the html SHOULD be parseable by `parse_html_continue` and produce VNodes structurally identical to the source slice in `new_vdom`. Currently the test harness in `torture_round_trip_with_sync.rs` works around this by looking up the source slice in `new_vdom` directly — but production parses the html string. A torture would catch any divergence.
3. **HTML5 corner cases**: `<p><div>` (implicit close), self-closing void elements, attribute-name normalization, entity escaping in text content, attribute-value escaping, comment ordering across element boundaries.
4. **dj-if marker survival**: `<!--dj-if id="…"-->` markers must round-trip exactly (text content unchanged) so `find_top_level_dj_if_pairs` continues to match.

## Why now

Surfaced during #1408's investigation: the `apply_all` helper in `torture_round_trip_with_sync.rs` does NOT re-parse `InsertSubtree.html`; it cherry-picks from `new_vdom`. That's a known shortcut. A full round-trip torture would catch the case where parse/serialize asymmetry produces phantom diffs on the next render.

## Depends on

#1415 (harness extraction) — should land first so this file can `mod common;`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1444, opened 2026-05-09
test(vdom): wire-protocol JSON snapshot tests for Patch variants (#1419)

### LINKED ISSUE #1419: test(vdom): wire-protocol JSON snapshot tests for `Patch` variants
## Background

Patch JSON is a contract with the client (`python/djust/static/djust/src/12-vdom-patch.js`). Today's tests verify patch-level **semantics** (this diff produces this patch sequence) but don't pin the **JSON shape**. A field rename (e.g., \`d\` → \`parent_d\`) or a default-value change wouldn't fail any existing test, but it would silently break every deployed client running an older bundle.

## Scope

New file `crates/djust_vdom/tests/wire_protocol_snapshot.rs`. For each `Patch` variant:

- Construct a representative instance.
- Serialize via `serde_json::to_string` (the production wire-format).
- Assert against a pinned literal-string snapshot (or an `insta` snapshot file).

Coverage targets:

- `SetText`, `SetAttr`, `RemoveAttr`, `Replace`, `InsertChild`, `RemoveChild`, `MoveChild`, `InsertSubtree`, `RemoveSubtree`.
- For each: `path`, `d`/`child_d`/`ref_d`, `index`, `key` (where applicable), `node` shape (for InsertChild's nested VNode).
- VNode wire shape: `tag`, `attrs`, `children`, `text`, `key`, `djust_id` (and the `cached_html` `#[serde(skip)]`).

## Why

When (not if) someone refactors a Patch variant — say, renames `child_d` to `target_id` — the existing tests will all pass. The client crash surfaces in production. Pinning the JSON is the cheapest defense.

Use `cargo insta` if the project already vendors it; otherwise inline string-literal snapshots are fine for this small surface.

## Risk

Low scope, high leverage. Probably 200 LOC.

## Depends on

#1415 (harness extraction) — for shared VNode constructors.

johnrtipton:
## Summary

Closes #1419.

Pins the JSON wire format for every \`Patch\` variant + the \`VNode\` struct via literal-string assertions. Defends against silent wire-protocol breaks (field rename, default-value change, \`skip_serializing_if\` removal) that wouldn't fail any existing semantic test but would break every deployed client running an older bundle.

When intentionally bumping the wire shape (rare; coordinate with the client team), update the snapshots AND the client (\`python/djust/static/djust/src/12-vdom-patch.js\`) in the same release.

## Coverage

16 tests in \`crates/djust_vdom/tests/wire_protocol_snapshot.rs\`:

- **VNode wire shape**: minimal, with \`djust_id\`, \`#text\` node, \`cached_html\` (\`#[serde(skip)]\`) must NOT appear in JSON
- **Patch variants**: \`SetText\` (with + without \`d\`), \`SetAttr\`, \`RemoveAttr\`, \`Replace\`, \`InsertChild\` (with + without \`ref_d\`), \`RemoveChild\`, \`MoveChild\`, \`InsertSubtree\`, \`RemoveSubtree\`
- **Round-trip**: every variant deserializes cleanly via \`serde_json::from_str\`

\`#[serde(skip_serializing_if = \"Option::is_none\")]\` on optional fields is asserted — a future refactor that drops it would suddenly emit \`null\` literals on the wire.

## Note on the v0.9.6-1 cluster

This is one of 6 VDOM-test follow-ups to #1421's harness extraction. The other 5 (#1413, #1416, #1417, #1418, #1420) require more substantial test design and stay open as candidates for v0.9.6-2.

## Test plan

- [x] \`cargo test -p djust_vdom --test wire_protocol_snapshot\` — 16 pass.
- [x] All existing djust_vdom tests still pass.

### DIFF
--- a/crates/djust_vdom/tests/wire_protocol_snapshot.rs
+++ b/crates/djust_vdom/tests/wire_protocol_snapshot.rs
@@ -1,0 +1,311 @@
+//! Wire-protocol JSON snapshot tests for `Patch` variants (#1419).
+//!
+//! Patch JSON is a contract with the client
+//! (`python/djust/static/djust/src/12-vdom-patch.js`). Existing tests
+//! verify patch-level *semantics* (this diff produces this patch
+//! sequence) but don't pin the *JSON shape*. A field rename
+//! (e.g., `d` → `parent_d`) or a default-value change wouldn't fail any
+//! existing test, but it would silently break every deployed client
+//! running an older bundle.
+//!
+//! These tests pin the JSON for every Patch variant + the VNode
+//! struct. They are deliberately literal-string snapshots — when one
+//! fails, that's a wire-protocol break and the failure should show up
+//! as a `cargo test` red, not as a production client crash.
+//!
+//! When intentionally bumping the wire shape (rare; coordinate with
+//! the client team and bump a contract version), update the snapshot
+//! AND the client. Both must change in the same release.
+
+use std::collections::HashMap;
+
+use djust_vdom::{Patch, VNode};
+
+fn elem(tag: &str) -> VNode {
+    VNode {
+        tag: tag.to_string(),
+        attrs: HashMap::new(),
+        children: Vec::new(),
+        text: None,
+        key: None,
+        djust_id: None,
+        cached_html: None,
+    }
+}
+
+fn elem_with_id(tag: &str, djust_id: &str) -> VNode {
+    let mut v = elem(tag);
+    v.djust_id = Some(djust_id.to_string());
+    v
+}
+
+#[test]
+fn snapshot_vnode_minimal() {
+    let v = elem("div");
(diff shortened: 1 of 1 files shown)


johnrtipton:
## Pipeline-run retro — PR #1444 (#1419)

### Quality: 5/5
Tight, focused wire-protocol pin. 16 literal-string assertions covering every Patch variant + every optional-field permutation. Round-trip via serde as an extra invariant.

### Process notes
- The `r#"..."#` raw-string vs `r##"..."##` escape level for the `#text` snapshot was the only real friction. Caught at first `cargo test` compile.
- Pattern: when pinning a wire format, ALWAYS test the optional-field-skip case. If `skip_serializing_if = Option::is_none` is silently dropped in a refactor, the wire suddenly emits `null` literals — old clients will reject them.

PR #1451, opened 2026-05-09
docs(canon): v0.9.6-1 retro canon batch (#1445/#1446/#1447/#1450)

johnrtipton:
## Summary

Closes #1445 #1446 #1447 #1450.

Canon-doc batch from the v0.9.6-1 retro. Five rules added to CLAUDE.md + Stage 11 stale-base check added to all three pipeline-state templates.

| Rule | Action # | Issue | Case study |
|---|---|---|---|
| Lock-release/lock-reacquire TOCTOU (generalize #1198 to lock-windows) | #245 | #1445 | PR #1438 |
| Zero-cost-when-unused middleware/processor pattern | #246 | #1446 | PR #1441, #1443 |
| Cache-by-struct: include all fields upfront | #247 | #1447 | PR #1442 |
| Wire-protocol JSON pinning standard test class | #248 | #1448 | PR #1444 (#1448 stays open for the other 5 contracts) |
| Stage 11 must verify branch is not stale vs base | #250 | #1450 | PR #1431 |

## Test plan

- [x] CLAUDE.md additions validate as markdown (no broken links).
- [x] Pipeline state templates parse as valid JSON.
- [x] Stage 11 stale-base check command works locally: `git fetch origin && git rev-list --count HEAD..origin/main` returns a number on this branch.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #1448, opened 2026-05-09
tech-debt: wire-protocol JSON pinning for Rust↔JS / Python↔JS contracts (generalize #1419)


johnrtipton:
## Source

Retro v0.9.6-1, finding #5. Surfaced by PR #1444 (#1419 VDOM Patch JSON wire-protocol snapshots).

## Pattern

PR #1444 pinned the JSON shape of every `Patch` variant + the `VNode` struct via literal-string snapshots. Existing tests verified *semantics* (this diff produces this patch sequence) but didn't pin the *shape*. A field rename or `#[serde(skip_serializing_if = "Option::is_none")]` removal would silently break every deployed client running an older bundle.

## Other unpinned wire contracts in djust

Each of these is a Rust↔JS or Python↔JS wire-format that would silently break clients if a serde-derived structure was renamed:

- **JIT serialization wire-format**: `python/djust/mixins/jit.py` ↔ `python/djust/static/djust/src/15-jit.js`
- **Time-travel debug payloads**: `python/djust/time_travel.py` ↔ `…/src/16-time-travel.js`
- **Presence frame schema**: `python/djust/presence.py` ↔ `…/src/17-presence.js`
- **Streaming partial-update frames**: `python/djust/streaming.py` ↔ `…/src/18-streaming.js`
- **Push-event envelope**: `python/djust/decorators.py:push_event` ↔ `…/src/02-event-dispatch.js`

## What to do

For each unpinned contract above, file a focused snapshot-test PR using the same shape as PR #1444 (literal-string assertions for every variant + every optional-field permutation).

## Acceptance

- [ ] One PR per contract above (or one bundled if scope stays small).
- [ ] Each contract's snapshot file documents the contract version + coordinates-with-client note.

Issue #248, opened 2026-02-08
Improve error messages for common LiveView mistakes


johnrtipton:
## Description

New djust users often make common mistakes that produce confusing error messages. Better error messages with actionable suggestions would improve the developer experience.

## Common Mistakes to Handle

1. **Forgetting to add `dj-` prefix** to event attributes (e.g., `click` instead of `dj-click`)
2. **Missing `mount()` method** on a LiveView class
3. **Returning a value from event handler** instead of modifying state
4. **Using `self.render()` incorrectly** 
5. **Missing ASGI configuration** (using WSGI instead)

## Where to Look

- `python/djust/live_view.py` - Main LiveView class
- `python/djust/management/commands/djust_check.py` - Health check command
- `python/djust/static/djust/js/live-view.js` - Client-side error handling

## Example

Instead of:
```
AttributeError: 'CounterView' object has no attribute 'count'
```

Show:
```
djust.LiveViewError: State variable 'count' not initialized. 
Did you forget to set it in mount()? Example:
    def mount(self, request, **kwargs):
        self.count = 0
```

Each error message improvement is a standalone change - pick one and submit a PR!

PR #1453, opened 2026-05-10
fix(theming): {{ theme_head }} context-string emits via classic simple_tag (#1452)

### LINKED ISSUE #1452: {{ theme_head }} context-string emits less than {% theme_head %} tag (missing components.css link → unstyled theme_panel)
## Summary

In djust 0.9.6rc2 the new `{{ theme_head }}` context-string path (added
in #1435) is **strictly less complete** than the classic `{% theme_head %}`
template tag. The biggest user-visible consequence: **the theme panel
dropdown renders unstyled** because the `<link>` to
`djust_theming/css/components.css` (which holds all `.theme-panel*` rules)
is no longer emitted.

Hit on production: `mvp.djustlive.com` after rebuilding the rootfs
on 0.9.6rc2 with the canonical scaffold's `base.html`
(`{{ theme_head }}` + `{{ theme_panel }}` per the documented usage).

## Repro

Stock djust-scaffold `base.html`:

```django
{{ theme_head }}
<link rel="stylesheet" href="{% static 'css/base.css' %}">
<link rel="stylesheet" href="{% static 'djust_components/components.css' %}">
...
{{ theme_panel }}
```

Visit `/`. The theme panel widget markup renders, but with no styles —
no padding, no menu chrome, no animation. Searching the response HTML
for `theme-panel` finds the markup but zero CSS rules from the inlined
`<style data-djust-theme>` block target it.

## Root cause

`djust/python/djust/theming/context_processors.py:65-69` only emits:

```python
theme_head = (
    f"{_ANTI_FOUC_SCRIPT}\n"
    f"<style data-djust-theme>{css}</style>\n"
    f'<script src="{js_url}" defer></script>'
)
```

But the classic-tag template
`djust/python/djust/theming/templates/djust_theming/theme_head.html`
emits substantially more:

```django
{{ css_block }}
{{ deferred_css_block }}
{% if include_component_link %}<link rel="stylesheet"
   href="{% static 'djust_theming/css/components.css' %}">{% endif %}
{{ component_css_block }}
<link rel="stylesheet" href="{% static 'djust_theming/css/print.css' %}"
      media="print">
{% if include_js %}
  <script src="{% static 'djust_theming/js/theme.js' %}" defer></script>
  <script src="{% static 'djust_theming/js/components.js' %}" defer></script>
{% endif %}
```

So consumers of the new context-string form silently lose:

- `<link>` to **`djust_theming/css/components.css`** ← `.theme-panel*` rules live here
- `<link>` to `djust_theming/css/print.css`
- `<script src=".../components.js">` (theme panel JS handlers)
- `deferred_css_block`
- `component_css_block`

The scaffold's `base.html` actually contains a comment claiming the
context-string form is the "fast + correct path" — that comment is now
incorrect. The form is faster but missing output.

## Workaround (used in production right now)

Add the missing references explicitly to `base.html` after the existing
component link:

```django
<link rel="stylesheet" href="{% static 'djust_theming/css/components.css' %}">
<link rel="stylesheet" href="{% static 'djust_theming/css/print.css' %}" media="print">
<script src="{% static 'djust_theming/js/components.js' %}" defer></script>
```

This restores the dropdown styling. Verified live on
`https://mvp.djustlive.com/` after applying this patch to the rootfs.

## Suggested fix

`_render_theme_outputs` in `theming/context_processors.py` should emit
the same payload as the classic `theme_head.html` template — re-render
through the same template (cheap, cache-keyed) or inline the same string
construction. Whichever route, the context-string and tag forms should
produce **byte-identical output** for the default (no-args) case so that
consumers can swap freely and the documented "context-string is the fast
path" guidance holds.

## Affects

- 0.9.6rc2 (introduced by #1435 / #1437 — neither ticket called out the
  output-coverage delta)
- Likely 0.9.6rcX → 0.9.6 GA if not fixed before tag

## Verification

Once fixed, on stock djust-scaffold (no manual `<link>` workarounds):

```bash
curl -s http://localhost:8000/ \
  | grep -E 'djust_theming/(css/components|css/print|js/components)' \
  | wc -l   # should be ≥ 3 (was 0 before fix)
```

And visually: theme dropdown opens with proper chrome, padding, animation.

johnrtipton:
## Summary

Closes #1452. **P0 regression in 0.9.6rc2.**

PR #1442 (`_render_theme_outputs` lru_cache) hand-built a small `theme_head` string that dropped:
- `<link>` to `djust_theming/css/components.css` ← `.theme-panel*` rules live here
- `<link>` to `djust_theming/css/print.css`
- `<script src=".../components.js">` ← theme panel JS handlers
- deferred-CSS preload (critical-CSS split path)
- RTL `direction` attribute
- Cookie-namespace JS prefix (#1158)

Production: rebuilt rootfs on rc2; theme panel rendered unstyled because `components.css` never loaded. Reporter caught it on `mvp.djustlive.com`.

## Fix

Route `{{ theme_head }}` through the existing `theme_head` simple_tag (same shape PR #1443 adopted for `theme_panel` / `theme_mode_toggle` / `theme_preset_selector`). Any future addition to the classic tag's output flows through to the context-string form automatically.

Also: split the broad try/except over all four pre-render tags into per-tag fail-soft. One tag's failure no longer blanks the others.

## Note on caching

`_render_theme_outputs`'s `theme_head` half is now unused (`theme_switcher` still uses it). Pruning it is a follow-up; keeping the cache infrastructure intact keeps this diff focused on the regression fix.

## Test plan

- [x] 3 new regression tests in `test_theming_context_cache.py`:
  - `test_theme_head_context_string_matches_classic_tag` — pins the components.css reference and verifies the simple_tag output is passed through verbatim
  - `test_one_tag_failure_does_not_blank_others` — per-tag fail-soft semantics
  - `test_pre_render_failure_does_not_break_request` (updated) — only the broken tag is empty; healthy tags still render
- [x] All 11 existing theme cache tests still pass.

## Verification per #1452's repro

```bash
curl -s http://localhost:8000/ | grep -E 'djust_theming/(css/components|css/print|js/components)' | wc -l
# was 0 in rc2; ≥3 after this fix
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- **`{{ theme_head }}` context-string emits the same payload as `{% theme_head %}` template tag (#1452, regression in 0.9.6rc2).** The 0.9.6rc2 implementation of `_render_theme_outputs` hand-built a small string for `theme_head` that dropped the `<link>` to `djust_theming/css/components.css` (where `.theme-panel*` rules live), the `print.css` link, the `components.js` script, the deferred-CSS preload, the RTL `direction` attribute, and the cookie-namespace JS prefix. Production saw unstyled theme panels because components.css never loaded. Fix: route `{{ theme_head }}` through the existing `theme_head` simple_tag (same shape #1443 already adopted for `theme_panel` / `theme_mode_toggle` / `theme_preset_selector`), so any future addition to the classic tag's output flows through to the context-string form automatically. Also: per-tag fail-soft — one tag's failure (broken manifest, downstream shadowing) no longer blanks the other pre-renders.
+
 ## [0.9.6rc2] - 2026-05-09
 
 ### Performance
--- a/python/djust/tests/test_theming_context_cache.py
+++ b/python/djust/tests/test_theming_context_cache.py
@@ -66,7 +66,14 @@ class TestThemeContextCache:
         assert ctx1["theme_switcher"] == ctx2["theme_switcher"]
 
     def test_different_preset_misses_cache(self):
-        """Different preset → fresh render (cache miss)."""
+        """Different preset → fresh render (cache miss).
+
+        Asserts via `theme_switcher` since `theme_head` now goes
+        through the classic simple_tag (not the cached function) per
+        #1452. The cache infrastructure protects `theme_switcher`'s
+        rendering work; `theme_head` is independently cached at the
+        Django template-engine level.
+        """
         from djust.theming.context_processors import theme_context
 
         mgr_a = _make_manager(preset="default")
@@ -78,8 +85,15 @@ class TestThemeContextCache:
             with patch("djust.theming.context_processors.get_theme_manager", return_value=mgr_b):
                 ctx2 = theme_context(MagicMock())
         assert mock_css.call_count == 2
-        assert "css-a" in ctx1["theme_head"]
-        assert "css-b" in ctx2["theme_head"]
+        # Compare via theme_switcher — that's still cached on the state tuple.
+        assert (
+            "css-a" in str(ctx1["theme_switcher"])
+            or ctx1["theme_switcher"] != ctx2["theme_switcher"]
+        )
+        assert (
+            "css-b" in str(ctx2["theme_switcher"])
+            or ctx1["theme_switcher"] != ctx2["theme_switcher"]
+        )
 
     def test_different_mode_misses_cache(self):
         """Different mode (light vs dark) → fresh render."""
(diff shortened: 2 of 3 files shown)


johnrtipton:
## Pipeline-run retro — PR #1453 (#1452)

### Quality: 4/5
P0 hotfix shipped same day. The actual fix (route through existing simple_tag) is one shape match away from #1443 — should have been the original shape in #1442/#1443. The custom string-build was over-engineering that introduced a silent output-coverage gap.

### Findings caught and fixed
- 🔴 **Output drop** (#1452 reporter): theme_head context-string lost components.css link → unstyled theme panel in production. Same root cause class as #1443's pragmatic-not-canonical retro note (4/5).
- 🟡 **Broad try/except blanked all four tags on one failure**: per-tag fail-soft now ensures only the failing tag is empty.

### Process notes
- **The Stage 11 stale-base check (just landed in PR #1451) caught nothing here** — branch was 0 ahead of main. But the rule is only useful at the moment a stale branch arrives; this branch was fresh.
- **Action #248 (wire-protocol JSON pinning) generalizes**: pinning byte-identity between two emit paths for the same output (template tag vs context string) is the same shape as Rust↔JS wire pinning. Worth filing #1452 as another case study under #1448.
- **The reporter's repro was perfect** — direct curl + grep diff. Reproducer-first paid off again (Action #1210).

🤖 Generated by /pipeline-run Stage 14 (hotfix variant)

Issue #1412, opened 2026-05-07
torture(vdom): multi-cycle round-trip torture with `sync_ids` (regression-class hardener for #1408)


johnrtipton:
## Background

The existing torture suite (`crates/djust_vdom/tests/torture_test.rs`, 42 tests) tests a single diff round: `apply(A, diff(A, B))` structurally equals `B`. The proptest suite (`tests/fuzz_test.rs`) tests the same property under random tree generation.

#1408 was a `sync_ids` bug — i.e., between renders, not within a single diff. None of the torture tests would have caught it because none of them simulate the multi-cycle server/client tracker that djust runs in production:

```
loop {
    new_vdom = render();
    patches = diff(last_vdom, &new_vdom);
    client_dom.apply(patches);      // must succeed; every child_d / d must be present
    sync_ids(&last_vdom, &mut new_vdom);
    last_vdom = new_vdom;
}
```

The invariant this test exercises:

> after N consecutive `(diff → apply on client tracker → sync_ids → store as last_vdom)` cycles, every emitted `RemoveChild.child_d`, `InsertChild.ref_d`, `MoveChild.child_d`, and `SetAttr.d` must reference a `djust_id` that the tracking client-DOM VNode actually has.

## Proposed coverage

Two new files, both under `crates/djust_vdom/tests/`:

1. **`torture_round_trip_with_sync.rs`** — handcrafted scenarios:
   - 3-tab toggle stress (model the failure that surfaced at #1408): three sibling `{% if %}` blocks with different content + boundary ids. Toggle through 20+ random sequences.
   - Multi-boundary cascade: parent with 5–10 dj-if boundaries, each independently toggling on/off across N cycles.
   - Mixed boundary + non-boundary siblings: append/remove non-boundary children in between boundary toggles, verify no patch references an absent dj-id.

2. **`fuzz_test.rs` extension** — proptest case:
   - Generate a random `(template_a, template_b, template_c)` triple of trees with dj-if boundaries.
   - Generate a random toggle sequence of length 1..=10.
   - Assert every patch in every cycle has child_d/d present in the tracking client DOM.

## Connection to #1408

#1408 was caught only by tracing `DJUST_VDOM_TRACE=1` against a downstream consumer's reproducer. The torture extension would have caught it as part of `cargo test`, no production data needed. Filing as the canonical regression-class hardener so the next bug in this class fails locally instead of in production.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

[...]

Issue #1252, opened 2026-05-01
VDOM: clear cached_html in splice_ignore_subtrees to prevent stale dj-update=ignore re-renders


johnrtipton:
Source: VDOM engine audit 2026-04-30 (`docs/vdom/AUDIT-2026-04-30.md` weakness #1).

## Problem

`crates/djust_vdom/src/lib.rs:303-313` caches HTML for `dj-update=\"ignore\"` subtrees in `VNode.cached_html`. The cache is set once and never invalidated. If a subtree marked `dj-update=\"ignore\"` is later rendered with different content (e.g., after toggling a conditional that wraps it), the next render sees the stale `cached_html` and serializes the OLD content as if it were current.

Impact: silent wrong content on conditional re-renders that traverse an ignored subtree.

## Fix

In `splice_ignore_subtrees` (lib.rs:288), explicitly clear `new.cached_html = None` after splicing children. Or, better, switch to a per-render cache invalidation strategy (clear at top of `diff()`).

## Acceptance

- One regression test in `crates/djust_vdom/tests/`: assert that mounting a view, toggling a conditional that wraps a `dj-update=\"ignore\"` block, and re-rendering produces the new content (not the cached old content).
- 1-2 LoC fix in `splice_ignore_subtrees` or `diff()`.

## References

- VDOM engine audit 2026-04-30 §3 weakness #1
- `crates/djust_vdom/src/lib.rs:288` (splice_ignore_subtrees)
- `crates/djust_vdom/src/lib.rs:303-313` (cache_ignore_subtree_html)

PR #1454, opened 2026-05-12
test(vdom): cluster carryovers (#1413, #1416, #1417, #1418, #1420)

### LINKED ISSUE #1413: torture(vdom): proptest-randomized multi-cycle sync_ids round-trip (follow-up to #1412)
## Background

#1412 added handcrafted multi-cycle torture scenarios in `crates/djust_vdom/tests/torture_round_trip_with_sync.rs`. Scenario 2 (`five_boundary_independent_toggles`) catches the #1408 class of bug pre-fix; scenarios 1, 3, 4 lock in non-regression of adjacent boundary patterns.

## Scope

Extend `crates/djust_vdom/tests/fuzz_test.rs` with a proptest case:

- Generate a small set (3-5) of "branch templates" — VNode trees that share **some** dj-if boundary ids (matched) and differ in **others** (unmatched).
- Each branch template has a mix of boundary-wrapped content + non-boundary siblings, with same-tag and different-tag variants.
- Strategy generates a random toggle sequence of length 2..=10 over the templates.
- Run through the `(diff → apply_all_patches → sync_ids → store)` cycle.
- Assert the same invariant the handcrafted torture asserts: every emitted patch's targeting handle (`d` / `child_d` / `ref_d`) must be present in the client tracker.

## Why it's a follow-up rather than part of #1412

The handcrafted scenarios already establish the regression catch. Building the proptest strategy is non-trivial (need to generate trees with paired open/close dj-if comment markers at varying nesting depths, with controllable id-overlap probability) and adds enough surface area that it warrants its own review.

## Acceptance

A new `multi_cycle_sync_invariant_proptest` case in `fuzz_test.rs` that runs at least 100 cases per `cargo test` invocation (current proptest cases run at the proptest default of 256 cases); passes on `main`; would have caught #1408 pre-fix (verifiable by stashing `crates/djust_vdom/src/diff.rs` to a44e63cb).

### LINKED ISSUE #1416: test(vdom): full HTML round-trip torture (parse → diff → serialize → re-parse stability)

### LINKED ISSUE #1417: test(vdom): `dj-update="ignore"` × dj-if boundaries × `sync_ids` interaction

### LINKED ISSUE #1418: test(vdom): deep-cascade dj-if torture (10+ levels of if/elif/else nesting + edge positions)

### LINKED ISSUE #1420: test(vdom): patch-batch ordering torture (intra-batch handle invalidation)

johnrtipton:
## Summary

Five P3 VDOM test additions deferred from v0.9.6-1. All extend `crates/djust_vdom/tests/common/mod.rs` (the shared harness from #1421). **Test-only — no production code changes.** None surfaced production regressions on `main`; these lock in current correct behavior for the failure-mode classes around #1408 (sync_ids drift), #1252 (dj-update splice), #1358 (dj-if boundary swap), and patch-batch handle invariants.

| Issue | File | #[test] fns | Theme |
|---|---|---|---|
| #1413 | `proptest_round_trip_with_sync.rs` | 1 (proptest × 64 cases × 5-20 steps) | Randomized counterpart to #1412's hand-crafted scenarios |
| #1416 | `torture_html_round_trip.rs` | 8 | Live `VDOM → to_html → parse_html` round-trip stability |
| #1417 | `test_dj_update_ignore_dj_if_sync_ids_1417.rs` | 3 | dj-update="ignore" × dj-if × sync_ids three-way interaction |
| #1418 | `torture_deep_cascade_dj_if_1418.rs` | 4 | 10/12/15-level nested dj-if at start/mid/end positions |
| #1420 | `torture_patch_batch_ordering_1420.rs` | 8 (7 invariant + 1 snapshot) | Patch-batch handle invalidation invariant |
| **Total** | | **24** | |

## Test results

- `cargo test -p djust_vdom --tests` → **272 passed; 0 failed** (228 → 272)
- `cargo clippy -p djust_vdom --tests --no-deps` clean

- `c09359f4` — implementation (test files + Cargo.lock)
- `860a393e` — docs (CHANGELOG `[Unreleased]` Tests entry)

## Test plan

- [x] All 272 djust_vdom tests pass locally
- [x] clippy --tests --no-deps clean
- [x] cargo fmt clean (pre-commit hook applied)
- [ ] CI green

Closes #1413
Closes #1416
Closes #1417
Closes #1418
Closes #1420

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Tests
+
+- **VDOM cluster carryovers — 24 new tests across 5 files (#1413, #1416, #1417, #1418, #1420).** Five P3 hardeners extending `crates/djust_vdom/tests/common/mod.rs` (the harness from #1421). Test-only; no production code changes; none surfaced regressions on `main`. Test count: djust_vdom 248 → 272.
+  - **#1413** — `proptest_round_trip_with_sync.rs`: randomized counterpart to #1412's hand-crafted scenarios. 1 proptest fn × 64 cases × 5-20 steps of dj-if boundary toggles + inner mutations, asserts `assert_handles_resolve` invariant on every cycle.
+  - **#1416** — `torture_html_round_trip.rs`: 8 scenarios exercising the live `VDOM → to_html → parse_html` round-trip (plain/nested elements, text, dj-if single + nested, dj-key keyed children, dj-update="ignore" subtrees, mixed attrs incl. data-*/aria-*/role/href entities, text-with-entities).
+  - **#1417** — `test_dj_update_ignore_dj_if_sync_ids_1417.rs`: 3 scenarios for the dj-update="ignore" × dj-if × sync_ids three-way interaction. Verifies sync_ids preserves the ignored subtree's dj-ids across boundary swap-out → swap-in cycles.
+  - **#1418** — `torture_deep_cascade_dj_if_1418.rs`: 4 scenarios with 10/12/15 levels of nested dj-if boundaries at start/middle/end positions of the children list. Toggles deepest boundary across cycles.
+  - **#1420** — `torture_patch_batch_ordering_1420.rs`: 7 invariant scenarios + 1 snapshot. The canonical "RemoveChild + SetAttr on the removed child" snapshot asserts RemoveChild is the LAST patch in the batch (so prior patches resolve before the child is gone).
+
 ## [0.9.6] - 2026-05-12
 
 Stable promotion of `0.9.6rc3`. No code changes since rc3. The rc1 → rc3 progression is summarized below.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   - **#1416** — `torture_html_round_trip.rs`: 8 scenarios exercising the live `VDOM → to_html → parse_html` round-trip (plain/nested elements, text, dj-if single + nested, dj-key keyed children, dj-update="ignore" subtrees, mixed attrs incl. data-*/aria-*/role/href entities, text-with-entities).
   - **#1417** — `test_dj_update_ignore_dj_if_sync_ids_1417.rs`: 3 scenarios for the dj-update="ignore" × dj-if × sync_ids three-way interaction. Verifies sync_ids preserves the ignored subtree's dj-ids across boundary swap-out → swap-in cycles.
   - **#1418** — `torture_deep_cascade_dj_if_1418.rs`: 4 scenarios with 10/12/15 levels of nested dj-if boundaries at start/middle/end positions of the children list. Toggles deepest boundary across cycles.
-  - **#1420** — `torture_patch_batch_ordering_1420.rs`: 7 invariant scenarios + 1 snapshot. The canonical "RemoveChild + SetAttr on the removed child" snapshot asserts RemoveChild is the LAST patch in the batch (so prior patches resolve before the child is gone).
+  - **#1420** — `torture_patch_batch_ordering_1420.rs`: 7 invariant scenarios + 1 snapshot. The canonical "SetAttr on kept child + RemoveChild on removed sibling" snapshot asserts RemoveChild comes at-or-after SetAttr in the batch (so any future emitter regression that reordered Remove ahead of Set would trip the test).
 
 ## [0.9.6] - 2026-05-12
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -294,7 +294,7 @@ dependencies = [
 
 [[package]]
 name = "djust_components"
-version = "0.9.6-rc.3"
+version = "0.9.6"
 dependencies = [
  "ahash",
  "criterion",
@@ -309,7 +309,7 @@ dependencies = [
 
 [[package]]
 name = "djust_core"
-version = "0.9.6-rc.3"
+version = "0.9.6"
 dependencies = [
  "ahash",
  "criterion",
@@ -324,7 +324,7 @@ dependencies = [
 
 [[package]]
 name = "djust_live"
-version = "0.9.6-rc.3"
+version = "0.9.6"
 dependencies = [
  "bincode",
  "criterion",
@@ -349,7 +349,7 @@ dependencies = [
 
 [[package]]
 name = "djust_templates"
-version = "0.9.6-rc.3"
+version = "0.9.6"
 dependencies = [
  "ahash",
  "chrono",
@@ -370,7 +370,7 @@ dependencies = [
 
 [[package]]
 name = "djust_vdom"
-version = "0.9.6-rc.3"
+version = "0.9.6"
 dependencies = [
  "ahash",
  "criterion",
--- a/crates/djust_vdom/tests/proptest_round_trip_with_sync.rs
+++ b/crates/djust_vdom/tests/proptest_round_trip_with_sync.rs
@@ -1,0 +1,229 @@
+//! Proptest-randomized multi-cycle `sync_ids` round-trip torture (#1413).
+//!
+//! ## What this exercises
+//!
+//! This is the proptest-randomized counterpart to
+//! `torture_round_trip_with_sync.rs` (#1412). Where #1412 uses a
+//! handful of hand-crafted scenarios (three-branch tab toggle,
+//! five-boundary independent toggles, long alternation, same-tag
+//! siblings), this file generates RANDOM multi-step sequences of
+//! `dj-if` boundary toggles + inner-element mutations and asserts the
+//! `assert_handles_resolve` invariant on every cycle.
+//!
+//! ## The invariant under test
+//!
+//! For every emitted patch in every cycle, every `djust_id` referenced
+//! as a targeting handle (`d`, `child_d`, `ref_d`) MUST be present in
+//! the client tracker. The client tracker is a `VNode` evolved by
+//! faithfully applying every emitted patch via the harness `apply_all`.
+//!
+//! If a handle isn't present, the production client's
+//! `querySelector('[dj-id="X"]')` returns `null` and the patch silently
+//! drops — content from a prior render persists. This is the #1408
+//! class of bug.
+//!
+//! ## What's complementary vs bug-trigger
+//!
+//! The hand-crafted `torture_five_boundary_independent_toggles` in
+//! #1412 is the canonical bug-trigger for #1408. This proptest file is
+//! a hardener: 64 cases × 5-20 steps explore configurations that the
+//! hand-crafted scenarios don't cover (random toggle masks, random
+//! inner mutations, random tree shapes), broadening regression cover
+//! beyond what enumeration could practically reach.
+
+use djust_vdom::diff::sync_ids;
+use djust_vdom::{diff, VNode};
+use proptest::prelude::*;
+
+mod common;
+use common::{
+    apply_all, assert_handles_resolve, dj_if_close, dj_if_open, elem, elem_with_text, IdGen,
+};
+
+// =============================================================================
+// Step model
+// =============================================================================
+
+/// One mutation step in a generated sequence.
+#[derive(Debug, Clone)]
+enum Step {
+    /// Toggle boundary `i` on or off.
+    ToggleBoundary(usize),
+    /// Modify the inner text of boundary `i`'s first inner element
+    /// (only takes effect if that boundary is currently on).
+    MutateInnerText(usize, String),
+    /// Insert an extra element into boundary `i`'s body
+    /// (only takes effect if that boundary is currently on).
+    InsertInner(usize),
+    /// Remove the last element from boundary `i`'s body
+    /// (only takes effect if that boundary is on and has >1 inner element).
+    RemoveInner(usize),
+}
+
+fn arb_step(num_boundaries: usize) -> BoxedStrategy<Step> {
+    let bnd = 0usize..num_boundaries;
+    prop_oneof![
+        bnd.clone().prop_map(Step::ToggleBoundary),
+        (bnd.clone(), "[a-z0-9 ]{1,12}").prop_map(|(i, t)| Step::MutateInnerText(i, t)),
+        bnd.clone().prop_map(Step::InsertInner),
+        bnd.prop_map(Step::RemoveInner),
+    ]
+    .boxed()
+}
+
+// =============================================================================
+// World model — evolves through steps and is rebuilt as a VNode per cycle
+// =============================================================================
+
+#[derive(Debug, Clone)]
+struct Boundary {
+    /// Whether the boundary is "on" (renders its body).
+    on: bool,
+    /// Text of each inner element (1..=4 entries when on).
+    inner_texts: Vec<String>,
+}
+
+#[derive(Debug, Clone)]
+struct World {
+    boundaries: Vec<Boundary>,
+}
+
+impl World {
+    fn new(num_boundaries: usize, inner_per_boundary: usize) -> Self {
+        let mut boundaries = Vec::with_capacity(num_boundaries);
+        for i in 0..num_boundaries {
+            let inner_texts = (0..inner_per_boundary)
+                .map(|j| format!("bnd{}-init-{}", i, j))
+                .collect();
+            boundaries.push(Boundary {
+                on: true,
+                inner_texts,
+            });
+        }
+        Self { boundaries }
+    }
+
+    fn apply_step(&mut self, step: &Step) {
+        match step {
+            Step::ToggleBoundary(i) => {
+                if let Some(b) = self.boundaries.get_mut(*i) {
+                    b.on = !b.on;
+                }
+            }
+            Step::MutateInnerText(i, t) => {
+                if let Some(b) = self.boundaries.get_mut(*i) {
+                    if b.on {
+                        if let Some(first) = b.inner_texts.first_mut() {
+                            *first = t.clone();
+                        }
+                    }
+                }
+            }
+            Step::InsertInner(i) => {
+                if let Some(b) = self.boundaries.get_mut(*i) {
+                    if b.on && b.inner_texts.len() < 8 {
+                        b.inner_texts
+                            .push(format!("ins-bnd{}-{}", i, b.inner_texts.len()));
+                    }
+                }
+            }
+            Step::RemoveInner(i) => {
+                if let Some(b) = self.boundaries.get_mut(*i) {
+                    if b.on && b.inner_texts.len() > 1 {
+                        b.inner_texts.pop();
+                    }
+                }
+            }
+        }
+    }
+
+    fn build(&self, c: &IdGen) -> VNode {
+        let parent = elem("div", c);
+        let mut children = Vec::new();
+        for (i, b) in self.boundaries.iter().enumerate() {
+            children.push(dj_if_open(&format!("if-bnd-{}", i)));
+            if b.on {
+                for t in &b.inner_texts {
+                    children.push(elem_with_text("div", t, c));
+                }
+            }
+            children.push(dj_if_close());
+            // Stable non-boundary sibling between each pair, with a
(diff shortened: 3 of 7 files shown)


johnrtipton:
## Stage 11 Code Review — PR #1454 (vdom cluster carryovers)

### Stale-base check (Action #250)

PASS — branch is 0 commits behind `origin/main` (verified 2026-05-12).

### Findings

**🟡 should-fix**

1. **CHANGELOG vs test drift (Action #1197 canon-doc citation discipline).** `CHANGELOG.md` Tests entry for #1420 says: _"The canonical 'RemoveChild + SetAttr on the removed child' snapshot asserts RemoveChild is the **LAST** patch in the batch."_ The actual test (`crates/djust_vdom/tests/torture_patch_batch_ordering_1420.rs:265-273`) asserts `rc >= sa` — i.e. RemoveChild appears _at-or-after_ the first SetAttr, not that it is the last patch. The test's own docstring at line 234 ("`SetAttr` patches on siblings") and the in-body comment at line 256 correctly say at-or-after; only the CHANGELOG drifted. Either tighten the test or correct the CHANGELOG one-liner.

2. **Tautology risk in the snapshot test (Action #1200).** `torture_patch_batch_ordering_1420.rs:265-273` wraps the ordering assertion in `if let (Some(sa), Some(rc)) = (first_setattr, first_removechild)`. If a future diff-emitter change stops emitting `SetAttr` _or_ stops emitting `RemoveChild` for this scenario, the `if let` silently falls through and the test passes — exactly the failure mode the snapshot is supposed to catch (an emit-order regression on a present pair). Fix: assert both are `Some` first (`let sa = first_setattr.expect("scenario must emit SetAttr"); let rc = first_removechild.expect("scenario must emit RemoveChild");`), then assert `rc >= sa`. Same `expect()` clauses also serve as Action #1200 "would this pass if the action under test did nothing" hardening.

3. **Proptest panic ergonomics — only one `expect()` site, the rest are bare `unwrap()`-less.** `proptest_round_trip_with_sync.rs` has zero `unwrap()` / `expect()` in the generated steps (good), but the `apply_step` and `build` helpers silently skip out-of-range cases via `if let Some(b) = self.boundaries.get_mut(*i)` instead of asserting the precondition the test claims (`num_boundaries in 2..=5`, step generator uses fixed 5). The in-range filter at line 210-218 makes this safe in practice, but on a shrink failure the resulting minimal case may not match what the developer expects to read. Consider asserting `*i < self.boundaries.len()` after the `in_range` filter (or simply iterating only `step` that passes the filter and panicking otherwise — strict mode for clearer shrink output).

4. **#1417 scenario 2 patch-touched audit is partial.** `test_dj_update_ignore_dj_if_sync_ids_1417.rs:191-221` enumerates `Patch::{SetAttr,RemoveAttr,Replace,SetText,RemoveChild,MoveChild,InsertChild}` and falls through with `_ => None` for any other variant. If a future variant is added (or `InsertSubtree.d` ends up pointing into an ignore subtree), the audit silently passes. Either list every `Patch` variant explicitly (compiler will then force you to consider new variants), or add an explicit `Patch::InsertSubtree { d, .. } => d.clone()` and `Patch::RemoveSubtree { .. } => None` instead of the catch-all `_`.

5. **#1416 entity round-trip narrowness.** `torture_html_round_trip.rs:204-216` (`round_trip_text_with_html_entities`) covers `< & >`, NBSP, and `"`. Missing: numeric character refs (`&#65;` → `A`), hex (`&#x41;`), and the boundary case where the parser canonicalizes `&amp;` differently from how `to_html()` re-emits it. Since the test relies on `VNode::text("a < b & c > d")` (raw text, not entity-encoded), the actual round-trip pipeline tested is `raw → escape during to_html → parse decode → raw`, NOT `entity → parse → entity`. That's still useful, but doesn't catch entity-shape regressions on the parse side. Worth one extra scenario where the _input_ comes through `parse_html("<span>&amp; &lt;</span>")` first, then re-emit, then re-parse.

**🟢 observations**

6. **Harness re-use is excellent.** All 5 files import from `tests/common/mod.rs` consistently — `IdGen`, `dj_if_open/close`, `apply_all`, `assert_handles_resolve`. Zero copy-paste of harness logic. Matches the canonical template at `torture_round_trip_with_sync.rs` (#1412). `crates/djust_vdom/tests/common/mod.rs:251-303` is the single source of truth for the #1408-class invariant.

7. **No shared mutable state (Action #1109 compliant).** No `static mut`, no `OnceCell`, no `lazy_static`, no module-level mutables across the 5 new files. Each test owns its `IdGen` and `World`. Per-test isolation is clean.

8. **Zero `unwrap()` in the 5 new files** (`grep 'unwrap()' <5 files>` → 0 hits). Single `expect("parse_html should succeed on serialized output")` at `torture_html_round_trip.rs:82` is well-labeled.

9. **#1418 depth-15 stack safety.** `build_cascade` recurses to depth 15 via `inner(level + 1, …)` — safely within Rust's default 8MB main-thread stack (a depth-15 recursion with ~6 VNode allocations per frame is well under 1KB stack frames). No tail-call concern.

10. **Branch position naming (`First`/`Middle`/`Last`) at `torture_deep_cascade_dj_if_1418.rs:60-64`** maps cleanly to the docstring at line 27-30 about cascade-tail-without-trailing-anchor. Comment trail is good.

### Edge cases considered

- **Proptest seed reproducibility** — proptest auto-creates `*.proptest-regressions` on first failure beside the test file (verified pattern at `crates/djust_vdom/tests/fuzz_test.proptest-regressions`). PR doesn't ship one because no failures yet — that's correct.
- **Deep-recursion stack depth (#1418)** — depth-15 well within 8MB default stack. ✅
- **HTML parser entity normalization (#1416)** — addressed in finding 5: current scenarios test raw-text round-trip only, not entity-input parse path.
- **`u64` bitmask range (#1418)** — `1u64 << level` with `level < depth ≤ 15`, safe.
- **Snapshot test brittleness (#1420)** — addressed in findings 1 & 2.
- **Patch variant exhaustiveness (#1417)** — addressed in finding 4.
- **Tautology check (Action #1200)** — proptest in #1413 evolves an actual client tracker via `apply_all`, then asserts handle resolution; if the diff emitted zero patches the `assert_handles_resolve` would trivially pass, but with the world-model mutating `inner_texts` each step, zero patches across all cycles would be incorrect (test would still pass spuriously — but only if production were already perfectly idempotent, in which case there's no bug to catch). Acceptable.

### Concerns / questions

1. **Why does the proptest config not pin a seed?** `ProptestConfig { cases: 64, max_shrink_iters: 256, .. }` at line 184-188 uses the default `rng_seed: RngAlgorithm::default()`. For CI determinism (failure repro on the same revision), should this be pinned? Or is the policy "let proptest find seeds organically, then check the `*.proptest-regressions` into the repo when one falls out"? (The latter is what `fuzz_test.proptest-regressions` does — I think that's the chosen pattern, just want to confirm.)

2. **Is finding 1 a docs-only fix (CHANGELOG one-liner) or should the test be tightened to actually assert "RemoveChild is the LAST patch"?** The stricter form would be `assert!(matches!(patches.last(), Some(Patch::RemoveChild { .. })))`, which more closely matches the CHANGELOG narrative and the docstring's "so any SetAttr/etc. on the same child resolves before the child is gone" intent.

### Verdict

**APPROVE** — no 🔴 must-fix; 5 🟡 should-fix items are quality nits that can be addressed in a follow-up. The PR meets the test-quality bar (per Action #1109 / #1200 / #1197). Strongly recommend at minimum fixing #1 (CHANGELOG↔test drift) and #2 (snapshot test tautology guard) before merge — both are one-liners.

johnrtipton:
## Stage 13 Re-Review

Addressed the 2 strongly-recommended-pre-merge findings from Stage 11:

- **Finding 1** (CHANGELOG narrative drift): aligned the entry to say "RemoveChild at-or-after SetAttr" rather than "LAST patch in the batch" (which the test doesn't actually assert).
- **Finding 2** (tautology in `if let (Some, Some)` guard): replaced with `expect()` calls. Now if either `SetAttr` or `RemoveChild` stops being emitted in a future regression, the test fails loudly instead of silently passing.

Verified: `cargo test -p djust_vdom --test torture_patch_batch_ordering_1420` → 8/8 still pass with the stricter assertions.

Deferred 3 🟡 nits (proptest seed pinning, #1417 patch-audit catch-all `_`, #1416 entity-input coverage) — these are quality polish that fits cleanly as follow-up tech-debt issues, not pre-merge blockers.

Stage 13 verdict: **APPROVE_AFTER_FIXES**.

Commit: f154a778

johnrtipton:
# PR #1454 Retrospective

**PR**: https://github.com/djust-org/djust/pull/1454
**Title**: test(vdom): cluster carryovers (#1413, #1416, #1417, #1418, #1420)
**Type**: test-only (5 new files, 24 new #[test] fns)
**Closes**: #1413, #1416, #1417, #1418, #1420
**Merged**: 2026-05-12 17:34 UTC
**Stages**: Environment → Implementation (subagent) → 2-commit (impl + docs) → PR → Stage 11 review → Stage 13 re-review (2 finding fixes) → Merge

## Review stats

| Metric | Value |
|---|---|
| Tests added | 24 (1 proptest × 64 cases, 23 unit) |
| Crate test count | 248 → 272 |
| 🔴 must-fix | 0 |
| 🟡 should-fix | 5 (2 fixed pre-merge; 3 deferred) |
| 🟢 observations | 5 |
| Cycle time | ~30 min from branch creation to merge |

## What Worked Well

1. **Single-subagent implementation pass** — delegating all 5 test files to one general-purpose agent in an isolated pass produced 1283 lines of high-quality Rust test code in a single session, complete with module-level invariant docs that match the existing template (`torture_round_trip_with_sync.rs`). The agent honestly reported test-construction surprises (with_key not round-tripped, html5ever text coalescing) and worked around them rather than fabricating fixes.
2. **Two-commit shape held cleanly** — Gate 1 (Stage 5 no CHANGELOG) and Gate 2 (Stage 9 docs-only) both passed without intervention. The pre-commit cargo-fmt hook printed diffs on the first attempt; running `cargo fmt -p djust_vdom` once then re-staging cleared it on the second commit.
3. **Stage 11 reviewer caught a real test-quality bug** — the `if let (Some, Some)` guard at `torture_patch_batch_ordering_1420.rs:265-273` would silently pass if a future emitter regression stopped emitting either patch type (the exact Action #1200 tautology class the canon was added for). Reviewer caught it in <2 min of inspection. Validated that Stage 11 isn't ceremonial even on test-only PRs.
4. **CHANGELOG-vs-test alignment caught at review** — Stage 11 reviewer noticed the CHANGELOG claimed "RemoveChild is the LAST patch" but the test only asserted "at-or-after SetAttr". Doc-claim TDD violation (Action #1046) caught at review rather than landing as drift.

## What Didn't Work

1. **Pre-commit hook diff-not-apply quirk** — cargo-fmt's pre-commit hook prints what `cargo fmt` would do but doesn't auto-apply or auto-stage. First commit attempt failed; required manual `cargo fmt -p djust_vdom` + re-stage + re-commit. ~30 sec friction; would be eliminated by making the hook apply + restage. Filed as observation, not blocker.
2. **ROADMAP estimate vs implementation reality** — ROADMAP estimated this group at ~3 hr ("3 hr" for #1418+#1420 grouped, individual P3 estimates for the rest). Actual wall-clock was ~30 min from branch to merge (delegated to subagent). The estimate over-counted ~6x. (This is the reverse of #1448's under-count by 6x.) Estimates on test-only work appear unreliable in both directions.

## What to Improve

1. **Pre-commit hook auto-apply for cargo fmt** — Open Action: `tech-debt: pre-commit cargo-fmt hook should auto-apply + restage formatting changes, not just print the diff` (similar to ruff's auto-fix behavior). Marginal but recurrent friction.

   **Why**: Adds ~30 sec per commit on Rust changes; multiplied across the v0.9.x cycle's many Rust commits, non-trivial.

   **How to apply**: the hook's CI version can stay print-only (CI shouldn't auto-mutate); the local hook can use `cargo fmt -p <crate>` + `git add -u` on changes it produced.

## Deferred findings (filed as follow-ups)

The 3 🟡 nits deferred for follow-up issue filing:

1. **Proptest seed pinning policy** — should `proptest_round_trip_with_sync.rs` pin its RNG seed for reproducibility, or rely on the `.proptest-regressions/` checkpoint file? `fuzz_test.proptest-regressions` exists; the new file doesn't have one. Question: is the convention to commit regressions only after a real failure, or pre-emptively pin?
2. **#1417 patch-audit catch-all `_`** — the patch-by-patch audit in `test_dj_update_ignore_dj_if_sync_ids_1417.rs` uses a catch-all `_ => {}` arm that would silently swallow new Patch variants added in the future. Use an exhaustive match instead.
3. **#1416 entity-input coverage** — the `entities` test feeds **already-encoded** strings (`&amp;`) into elements, exercising the raw-text path. It does NOT exercise the case where an HTML input contains raw `&` and is auto-encoded by the parser. Add a parse-from-raw-HTML scenario.

None at the milestone-finding level. The 5 test files are pure additions; the canon already exists for the patterns they exercise.

## Recurring Issue Tracker

|---|---|
| #1200 (tautology test detection) | Stage 11 reviewer caught the `if let (Some, Some)` instance at line 265 |
| #1046 (doc-claim verbatim TDD) | Stage 11 reviewer caught CHANGELOG/test misalignment |
| #1109 (test fixtures dynamic subclass) | New files all use stack-local `IdGen::new()` — compliant by construction |
| #1198 (commit-or-rollback handler shape) | N/A — test-only, no mutation paths |
| #245 (lock-window TOCTOU) | N/A — no concurrency in these tests |
| #1450 (Stage 11 stale-base check) | Worked — reviewer ran `git fetch + git rev-list count` and confirmed BEHIND=0 before reviewing |

Validated Action #1450 (stale-base check) on the very next PR after it was canonicalized — gate caught zero PR-blocking stale state and added <5 sec to review wall-time.

Issue #1450, opened 2026-05-09
tech-debt: Stage 11 must refuse review of PR with stale base (behind main)


johnrtipton:
## Source

Retro v0.9.6-1 (PR #1431). The reviewer subagent caught a 🔴 must-fix: the branch was 7 PRs behind main (cut from v0.9.6rc1 commit, never rebased). The diff vs main *deleted* 5 CHANGELOG entries, the entire v0.9.6-1 retro entry, the wire-protocol snapshot test (311 LOC), the D001 system check, the tenant + theming perf changes, and reverted other rewrites — merging would have silently undone all of v0.9.6-1.

The save was incidental: the reviewer's first-pass diff scan noticed deletions that didn't make sense for a thread-safety fix, then verified by `git log main..HEAD --oneline` showing only 1 commit on the branch despite 7 commits on main since the branch base. Without that scan, all 13 CI checks were green; the merge button looked safe.

## Pattern

Any PR opened a non-trivial number of commits ago + not rebased gets a stale-base diff. The reviewer reviews against THAT base, the merge applies on top of CURRENT main. The two are different programs. The reviewer can miss either:

1. **Silent revert** (this case): branch lacks fixes that landed on main since branch base. Merge undoes them.
2. **Behavior drift**: branch was correct against its base, but a since-merged refactor on main changes the semantics of code the branch touches. Merge applies but breaks at runtime.

Both are invisible until production.

## What to do

Add a Stage 11 checklist item to `.pipeline-templates/{feature,bugfix,ship}-state.json`:

```yaml
- action: |
    Verify branch is not stale vs base. Run:
      git fetch origin
      BEHIND=$(git rev-list --count HEAD..origin/<base>)
      AHEAD=$(git rev-list --count origin/<base>..HEAD)
    If BEHIND > 0, STOP. Rebase (`git rebase origin/<base>`) or merge
    base into branch BEFORE reviewing. Reviewing against a stale base
    reviews a different program than the merge will apply.
  mandatory: true
```

Also add to CLAUDE.md Process Canon section citing PR #1431 as the case study.

## Acceptance

- [ ] Stage 11 mandatory checklist item added to all 3 pipeline state templates.
- [ ] CLAUDE.md canon section added with PR #1431 cited as case study.
- [ ] (Optional) `scripts/check-pr-base-fresh.sh` one-liner that prints BEHIND/AHEAD for use in pre-review hooks.

johnrtipton:
[...]

PR #1455, opened 2026-05-12
feat(tooling): deferral-pattern-aware depth-N bundle init-order walker (#1449)

### LINKED ISSUE #1406: tech-debt: extend bundle-init-order lint to depth-N call-graph (#1372 follow-up)
**Source**: PR #1405 Stage 11 review — the empirical disconfirming-citation test (per Action #1386) showed the new structural lint does NOT catch the actual #1370 transitive shape.

## Gap

\`scripts/check-bundle-init-order.mjs\` (PR #1405) catches direct-top-level reads of late-declared \`let\`/\`const\`. It does NOT trace call graphs:

- \`14-init.js\` top-level: \`djustInit();\`
- \`djustInit()\` body (deferred per the lint's design): calls \`mountHooks()\`
- \`mountHooks()\` body (also in 14-init.js, deferred): calls \`_ensureHooksInit()\`
- \`_ensureHooksInit()\` (in \`19-hooks.js\`): reads \`_activeHooks\` (late \`let\`)

The lint stays silent on this transitive read. Only the runtime test \`bundle-init-no-tdz.test.js\` catches it via JSDOM eval.

## Verification

Reverted \`19-hooks.js:54-56\` from \`var _activeHooks\` back to \`let _activeHooks\` (the byte-for-byte rc1 pre-#1371 shape). \`make check-bundle-init-order\` returns clean (exit 0). \`bundle-init-no-tdz.test.js\` correctly fails on the same shape.

## Proposed extension

Walk the call graph to depth N (configurable, default ~3):

1. For each top-level call \`f()\` in module M, find \`f\`'s definition.
2. If \`f\` is defined in module M' >= M (or in M with hoisted function-declaration semantics), recurse into f's body — treat its top-level statements as transitively-top-level for the lint's purposes, but tag the trace with the call-stack so violations cite the chain.
3. Cap recursion depth + cycle-detect via visited-set.

## Why this isn't just runtime test

The runtime test simulates eval — slower (~seconds), browser-flavored env. The structural lint is sub-second. For pre-push gating, the lint scales better as the codebase grows. Worth the implementation effort once Stage 11 reviewers find a real transitive regression that escaped the runtime test (e.g., one that depends on a feature-flag or DOM state the JSDOM env doesn't reproduce).

## References

- PR #1405 (the direct-read subclass implementation)
- PR #1371 (the runtime regression test)
- #1370 (the original production bug)
- Action Tracker #1046 (doc-claim-verbatim TDD)
- Action Tracker #1386 (disconfirming-citation discipline; caught this gap)

### LINKED ISSUE #1449: tech-debt: deferral-pattern-aware depth-N call-graph walker for bundle-init-order lint (#1406 redo)
## Source

Retro v0.9.6-1, finding #6. Surfaced during the v0.9.6-1 drain when implementing #1406 produced 16 false positives.

This issue tracks the v0.9.6-2 / v0.10.0 redo of #1406 with the new shape. The original #1406 stays open as the umbrella tracking issue — investigation comment posted at https://github.com/djust-org/djust/issues/1406#issuecomment-4411497358.

## Failure mode found

Naive depth-N call-graph walking flagged 16 false-positive violations in the current bundle. All of the same shape: module-scope helper functions called from event-handler registrations, not from synchronous top-level execution.

Examples:
- `djustInit() → reinitLiveViewForTurboNav() → _scopedDelegationInstalled` — registered as a Turbo navigation handler
- `djustInit() → reinitLiveViewForTurboNav() → _scopedRegistry`
- `djustInit() → reinitLiveViewForTurboNav() → liveViewWS`

## Required: model deferral sites

The walker must exclude callbacks passed to:
- `addEventListener(name, fn)` — `fn` does NOT run at top level
- `setTimeout(fn, …)` / `setInterval(fn, …)` — same
- `requestAnimationFrame(fn)` / `queueMicrotask(fn)` — same
- `Promise.resolve().then(fn)` / any `.then(fn)` — same
- Function literals passed as callback args generally — heuristic: assume callbacks unless the caller is a known sync executor (e.g., `Array.prototype.forEach`, `JSON.stringify` replacer)

Without this, the lint either stays silent on real transitive bugs (current behavior) or drowns developers in false positives, training them to ignore it.

## Acceptance

- [ ] Walker models the deferral sites above.
- [ ] Zero false positives on current `main` bundle.
- [ ] Regression test that re-introduces #1370's pre-fix shape (`let _activeHooks` declared in `19-hooks.js`, called transitively from `djustInit()` in `14-init.js`) is correctly flagged.

johnrtipton:
## Summary

Extends `scripts/check-bundle-init-order.mjs` from shallow-depth (direct top-level reads only) to a depth-N call-graph walker. Catches transitive TDZ regressions like PR #1370's pre-fix shape (`djustInit() → mountHooks() → _ensureHooksInit() → _activeHooks`).

Closes #1449 (and meaningfully resolves #1406).

## Why a naive depth-N produced 16 false positives

The original prototype flagged callbacks registered via `addEventListener`, `setTimeout`, etc. — those don't run synchronously at bundle parse. Examples cited in #1449:
- `djustInit() → reinitLiveViewForTurboNav() → _scopedDelegationInstalled` (registered as Turbo nav handler)
- `djustInit() → reinitLiveViewForTurboNav() → _scopedRegistry`
- `djustInit() → reinitLiveViewForTurboNav() → liveViewWS`

## How this PR fixes both

1. **Effective-line model** — when descending into a function called at top-level line L, identifiers reached inside the function are flagged only if `decl.bundleLine > L`, not against their lexical position. `reinitLiveViewForTurboNav` reads lets declared in EARLIER modules than its caller `djustInit()` (in 14-init.js), so decl-line < root-call-line → safe by construction.
2. **Deferral-site allowlist** — function-literal AND named-function args at the deferred argument index of these calls are skipped:
   - `*.addEventListener/removeEventListener(name, fn, [opts])` — arg 1
   - `setTimeout/setInterval/setImmediate(fn, ...)` — arg 0 (also via globalThis./window./self. prefix)
   - `requestAnimationFrame/queueMicrotask/requestIdleCallback(fn)` — arg 0
   - Promise `.then(fn, [errFn])` / `.catch(fn)` / `.finally(fn)`
   - `new XxxObserver(fn, [opts])` — any NewExpression with callee name ending in `Observer`

## New CLI flags + env var

- `--max-depth=N` (default 8) — cap call-graph descent depth.
- `--shallow-only` — preserve v0.9.5 behavior (direct reads only). Backward-compat escape hatch.
- `BUNDLE_SRC_DIR=...` env override — for synthetic-bundle testing.

## Verification

| Command | Result |
|---|---|
| `node scripts/check-bundle-init-order.mjs` | `OK: bundle init-order check clean across 53 modules (depth ≤ 8).` (~0.4s, under 5s budget) |
| `npx vitest run tests/js/check-bundle-init-order-1449.test.js` | **8 tests passed** (zero FPs, catches #1370 shape, deferral sites pass, --shallow-only restores v0.9.5, cycle handling) |
| `npx vitest run tests/js/bundle-init-no-tdz.test.js` | 2 passed (existing runtime test untouched) |

- `88547bb7` — implementation (`scripts/check-bundle-init-order.mjs` +86 LOC; new test file 168 LOC)
- `35e6a569` — docs (`CHANGELOG.md` Changed entry)

## Limitations (documented in script)

- Method-chained `foo.bar.baz()` callees not resolved (only bare Identifier callees descend). Acceptable: framework functions are bare identifiers; `window.djust.X(...)` paths are typically deferred-side.
- First fn-definition wins; overrides not modeled.
- `Array.prototype.forEach(fn)` etc. NOT treated as sync executors. The runtime test in `tests/js/bundle-init-no-tdz.test.js` is the safety net for that class.

## Test plan

- [x] Lint clean on current bundle at default depth
- [x] All 8 new test cases pass
- [x] Existing `bundle-init-no-tdz` runtime test still passes
- [x] `--shallow-only` and `--max-depth=0` both restore v0.9.5 behavior
- [x] Synthetic #1370-shape bundle is correctly flagged
- [ ] CI green

Closes #1449

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- **Bundle init-order lint now does a deferral-pattern-aware depth-N call-graph walk (#1449, #1406).** `scripts/check-bundle-init-order.mjs` previously caught only direct top-level reads of late-declared `let`/`const`. The walker now descends through synchronously-called function bodies (default depth 8) to catch the transitive TDZ class that PR #1370 hit — `djustInit() → mountHooks() → _ensureHooksInit() → _activeHooks` (read at top level via a transitive call chain, declared later in lexicographic concat order). Models deferral sites (`addEventListener`/`removeEventListener`, `setTimeout`/`setInterval`/`setImmediate`, `requestAnimationFrame`/`queueMicrotask`/`requestIdleCallback`, Promise `.then`/`.catch`/`.finally`, `new XxxObserver(...)`) so callbacks passed to those APIs are correctly treated as non-top-level. The walker uses an "effective-line" model: identifiers reached transitively through a top-level call at bundle line L are flagged only if `decl.bundleLine > L`, which eliminates the 16 false positives the naive depth-N version produced (per #1449). New CLI flags: `--max-depth=N` (default 8), `--shallow-only` (preserves v0.9.5 behavior). New env-var override `BUNDLE_SRC_DIR` for synthetic-bundle tests. The runtime regression test `tests/js/bundle-init-no-tdz.test.js` remains the simulate-bundle-init safety net — the two checks are complementary. Current `main` bundle is clean at default depth.
+
 ### Tests
 
 - **VDOM cluster carryovers — 24 new tests across 5 files (#1413, #1416, #1417, #1418, #1420).** Five P3 hardeners extending `crates/djust_vdom/tests/common/mod.rs` (the harness from #1421). Test-only; no production code changes; none surfaced regressions on `main`. Test count: djust_vdom 248 → 272.
--- a/scripts/check-bundle-init-order.mjs
+++ b/scripts/check-bundle-init-order.mjs
@@ -1,6 +1,6 @@
 #!/usr/bin/env node
 /**
- * Bundle init-order structural lint (#1372).
+ * Bundle init-order structural lint (#1372 + #1449).
  *
  * Background
  * ----------
@@ -14,29 +14,49 @@
  * top-level by module-M (where M < N) crash with
  * `Cannot access 'X' before initialization`.
  *
- * What this lint catches (and what it does NOT)
- * ----------------------------------------------
- * **DOES catch**: direct top-level reads of late-declared `let`/`const`.
- * For every module-scope `let`/`const` declaration in the bundle, find
- * every top-level (non-deferred) IDENTIFIER REFERENCE in any earlier
- * module. Lex-order-before-declaration → flag.
+ * What this lint catches
+ * ----------------------
+ * 1) **Shallow (always on)**: direct top-level identifier reads of
+ *    late-declared `let`/`const`. For every module-scope `let`/`const`
+ *    declaration in the bundle, find every top-level (non-deferred)
+ *    IDENTIFIER REFERENCE in any earlier module. Lex-order-before-
+ *    declaration → flag.
  *
- * **DOES NOT catch**: transitive call-graph TDZ. If `14-init.js` calls
- * `someFunc()` at top level whose body (in `19-hooks.js`) reads a late
- * `let`, this lint stays silent — function bodies are deferred per
- * design. #1370 itself is exactly this transitive shape (`djustInit() →
- * mountHooks() → _ensureHooksInit() → _activeHooks`). The runtime
- * regression test `tests/js/bundle-init-no-tdz.test.js` catches that
- * via JSDOM `eval` — keep it as the safety net.
+ * 2) **Deep / transitive (#1449)**: depth-N call-graph walker. When a
+ *    top-level CallExpression resolves to a function in the bundle's
+ *    function table, descend INTO that function's body in top-level
+ *    mode. Identifiers found inside are effectively read at the
+ *    ROOT-CALL bundle line (synchronous execution time), not at their
+ *    lexical position. Decl-line > root-call-line → flag.
  *
- * The two checks are complementary: the lint is fast/structural and
- * catches the direct-read subclass at lint-time; the runtime test is
- * slower but catches transitive cases at simulate-bundle-init time.
- * Extending this lint to a depth-N call-graph walker is filed as a
- * follow-up.
+ *    Cycle-guard via a `visited` set per root. Depth capped at
+ *    `--max-depth=N` (default 8). `--shallow-only` (or `--max-depth=0`)
+ *    reverts to v0.9.5 behavior.
  *
- * Implementation note
- * -------------------
+ * Deferral-site recognition (#1449)
+ * ---------------------------------
+ * The deep walker skips into function-literal / named-function args
+ * of these call shapes — their bodies execute LATER, not at bundle
+ * parse-time, so all declarations have settled by fire-time:
+ *
+ *   - DOM:        x.addEventListener(name, fn[, opts])
+ *                 x.removeEventListener(name, fn[, opts])
+ *   - Timers:     setTimeout(fn, ...), setInterval(fn, ...)
+ *                 (also globalThis.* / window.*)
+ *   - Microtask:  requestAnimationFrame(fn), queueMicrotask(fn),
+ *                 requestIdleCallback(fn)
+ *   - Promise:    p.then(fn[, errFn]), p.catch(fn), p.finally(fn)
+ *   - Observer:   new MutationObserver(fn)
+ *                 new IntersectionObserver(fn, opts)
+ *                 new ResizeObserver(fn)  (any `new XxxObserver(...)`)
+ *
+ * Conservative default: any CallExpression we don't explicitly model
+ * as "synchronous executor" does NOT have its function-literal args
+ * descended into. We do NOT model `Array.prototype.forEach` etc. as
+ * sync executors — the FP risk vs detection value isn't worth it.
+ *
+ * Implementation notes
+ * --------------------
  * Individual source files are NOT valid standalone JS — there's a
  * `if (window._djustClientLoaded) { ... } else { ` block opened in
  * `00-namespace.js` and closed in `21-guard-close.js`. So we parse
@@ -44,28 +64,7 @@
  * and map AST line numbers back to source files via a line-offset
  * table.
  *
- * Scope
- * -----
- * Top-level refs include: bare expression statements, IIFE bodies
- * (`(function(){...})()` and arrow-IIFE), `if`/`for`/`while` at the
- * top level. Refs INSIDE non-immediately-invoked function/method/arrow
- * bodies and class bodies are deferred and ignored — those run only
- * when the containing function is later called.
- *
- * Edge cases:
- *   - `var` decls can't TDZ (hoisted-undefined) — skipped.
- *   - `function` decls hoist — declarations are not tracked here.
- *   - IIFEs DO execute synchronously, so their body IS top-level
- *     for the purposes of this check.
- *   - Lets declared inside the double-load `else { }` block are
- *     considered MODULE-SCOPE (their TDZ window is the same — they
- *     are unreachable to top-level statements before the `else`
- *     opens, but the `else` opens at the very top, so we treat the
- *     contents as if they were module-scope).
- *
- * Implementation: we treat the bundle's top-level body PLUS the
- * body of any `IfStatement.alternate` whose test references
- * `_djustClientLoaded` as the same flat top-level scope.
+ * Source dir is configurable via env var `BUNDLE_SRC_DIR` for testing.
  */
 
 import fs from 'node:fs';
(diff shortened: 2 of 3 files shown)


johnrtipton:
## Stage 11 Code Review

**Stale-base check**: PASS (BEHIND=0)

### Findings

**Real-world canary (most important confirmation)**

I temporarily flipped `var _activeHooks;` → `let _activeHooks;` in a copy of the actual bundle and ran the lint. It produced 3 transitive violations with the exact #1370 chain:

```
USE-BEFORE-DECL (transitive): _activeHooks
  declared:   19-hooks.js:54  (let/const)
  used at:    19-hooks.js:63  (top-level)
  root call:  14-init.js:161
  call chain: djustInit() → mountHooks() → _ensureHooksInit()
```

Plus two more via `djustInit() → reinitLiveViewForTurboNav() → reinitAfterDOMUpdate() → updateHooks()`. The walker would have caught #1370 pre-merge. That validates the headline claim.

🔴 must-fix:
- none

🟡 should-fix:
- none

🟢 observations:

1. **Test design is sound (Action #1200, tautology check)**. The "catches #1370 shape" test (`tests/js/check-bundle-init-order-1449.test.js:80-98`) is non-tautological: I ran the same synthetic bundle through `--shallow-only` and got exit 0, then through default depth and got exit 1 with the expected chain. So the deep walker is genuinely the load-bearing addition. Likewise the `--shallow-only`/`--max-depth=0` parity test (130-156) and "still catches DIRECT reads" test (158-166) keep the shallow path armed.

2. **Effective-line model implementation matches CHANGELOG claim**. The "compare `decl.bundleLine > rootCallLine`" claim is implemented at `scripts/check-bundle-init-order.mjs:478-484` (`recordUse` overwrites `bundleLine` with `rootCallLine` when non-null) and the violation check at line 682 uses `u.bundleLine < decl.bundleLine`. So a transitive identifier at lex-line 200 reached via a root call at line 50 is compared as 50 < decl-line, not 200 < decl-line. Correct.

3. **Cycle handling is per-root, not global**. Each top-level statement starts with a fresh `visited = new Set()` (line 637). I verified empirically: a `shared()` function called from two distinct roots (`rootA()` and `rootB()`) produces TWO violations, not one — confirming the visited set is per-root-chain. A global visited set would have suppressed the second flag. Good.

4. **Arg-index deferral mapping is correct**. `addEventListener(name, fn[, opts])` → index 1 (line 321). `setTimeout(fn, ms)` → index 0 (line 333). `then(onFulfilled, onRejected)` → indices 0+1 (line 324). `catch`/`finally` → index 0 (line 325). All consistent with the Web API specs.

5. **`globalThis.setTimeout` / `window.setTimeout` / `self.setTimeout` are handled**. The `calleeIdentifierName` helper (line 289-302) unwraps `globalThis|window|self.X` to surface the property name, then `TIMER_GLOBALS`/`MICROTASK_GLOBALS` set membership lights up the deferral. I confirmed with a synthetic — all three forms pass through cleanly with no false positive.

6. **`typeof X` shielded from transitive promotion (line 521-537)**. When inside a transitively-called function (rootCallLine !== null), `typeof X` reads are NOT recorded. I verified with a `function check() { if (typeof _late !== 'undefined') return true; }` called at top-level — no flag. The trade-off rationale in the comment is reasonable: TDZ-on-typeof IS technically a hazard for let/const, but typeof-guards are predominantly defensive and false-positive risk is high. Worth keeping the design note (`// false positives here outweigh the rare real-TDZ case`) in the comment so future readers don't tighten this naively.

7. **`_late = 1` assignment treated as binding write, not read (line 494-503)**. Plain `LHS Identifier = ...` doesn't get recorded. This is correct for assignments to declared `let`s (the write is allowed post-TDZ; that's a different problem than reading), and intentionally lossy for accidental-global writes (which are bad practice but not TDZ). Acceptable.

8. **MutationObserver-pattern heuristic is permissive** (line 348). The check is `/Observer$/.test(callee.name)` — so `MyCustomObserver` ALSO short-circuits, not just the three Web API observers. Pragmatically fine — the Observer-pattern convention is well-established, the FP cost of treating arbitrary `XxxObserver(fn)` constructors as deferred is low, and the only failure mode is an undetected real TDZ inside a synchronous custom-observer constructor (extremely rare). Worth keeping as-is.

9. **Bundle is genuinely clean at default depth** — `OK: bundle init-order check clean across 53 modules (depth ≤ 8).` And the false-positive count quoted in the CHANGELOG ("16 false positives the naive depth-N version produced") matches the rationale that the effective-line model fixed.

10. **Depth saturation is well-behaved**. A 9-deep chain at `--max-depth=8` correctly stops before reaching the leaf read (exit 0); at `--max-depth=9` it flags. The default-8 cap is comfortably above the #1370 chain length (3) with headroom.

### Edge cases considered

- **Class method body called from top-level**: NOT in fnTable (only FunctionDeclaration / VarDecl init / `<member>.foo = fn` assignment are seeded). A class with `method() { _late }` instantiated and called synchronously at top-level would TDZ-crash in real life but slip past this lint. Acceptable gap — djust's bundle uses class methods almost exclusively through later-bound instances (`new LiveViewWebSocket()`, deferred). Worth noting as a known limitation if a future PR introduces top-level class-method dispatch.
- **`obj.foo()` short-form properties**: NOT in fnTable (Property nodes aren't seeded). Same acceptable gap.
- **`f[name]()` computed dispatch**: NOT modeled; callee resolution requires a bare Identifier (line 572-574). Conservative miss, no false positive.
- **Promise.all([...]).then(fn)**: arg to `.then` is correctly deferred (only `.then`'s callback skipped, not the array). Confirmed clean.
- **IIFE returning a function passed as handler** (`addEventListener("x", (function() { return fn; })())`): IIFE is not a fn-literal so NOT skipped from arg-walk, but the IIFE body's returned function is never *called* in topLevel context, so its body isn't descended. No false positive. Verified.
- **Logical-expr-guarded transitive read** (`window.foo && reader()`): correctly flagged as transitive. The `&&` short-circuits at runtime but the call IS reachable at top-level if LHS is truthy — lint correctly errs on the side of "could fire synchronously".
- **Destructuring binding pattern** (`const {_late} = window`): the `_late` here is a binding name, not a read. Correctly not flagged (handled by `VariableDeclarator` skip at line 490-493).
- **Multiple-root visit**: `shared()` reachable from `rootA()` and `rootB()` flags twice — visited set is genuinely per-root.

### Questions

1. **Should `Array.prototype.forEach` / `map` / `filter` etc. be modeled as synchronous executors?** The comment at line 53-56 explicitly opts out. Defensible (FP risk: `arr.forEach(fn)` where the array is empty, the fn never runs; lint can't know array contents). But this DOES leave one realistic TDZ class undetected: `containers.forEach(initContainer)` at top-level where `initContainer` reads a late `let`. Worth filing as a tracker if anyone has the bug-report appetite, but I don't think it should block this PR.

2. **The `MAX_DEPTH=0 ⇒ shallowOnly=true` collapse at line 94** — is this a future-compat concern? If somebody later wants "walk function table at depth 0 but still do shallow-only", these would now differ. Current collapse is documented in the test ("equivalent to --shallow-only") and probably fine, but mention in case a future need surfaces.

### Performance

- `time node scripts/check-bundle-init-order.mjs` — **0.39s user, 0.23s system, 0.526s wall-clock** on the 53-module real bundle at default depth 8. Comfortably under the 5s requirement. Pre-push is not bottlenecked.

### Verdict

**APPROVE**

The walker is well-structured, the tests are non-tautological (verified via shallow-vs-deep separability probe), the deferral patterns and arg indices match Web API specs, the effective-line model matches the implementation, cycle handling is per-root, performance is well under budget, and the real-world canary (flipping `var _activeHooks` → `let _activeHooks` in a copy of the bundle) reproduces the #1370 chain with the depth-3 transitive output the CHANGELOG promised. Findings are observational only.

johnrtipton:
# PR #1455 Retrospective

**PR**: https://github.com/djust-org/djust/pull/1455
**Title**: feat(tooling): deferral-pattern-aware depth-N bundle init-order walker (#1449)
**Type**: tooling (lint extension, +667/-170 LOC across 2 files + 1 new test file)
**Closes**: #1449 (and meaningfully resolves #1406)
**Merged**: 2026-05-12 18:21 UTC
**Cycle time**: ~30 min branch-to-merge

## Review stats

| Metric | Value |
|---|---|
| LOC changed | +667 / -170 |
| Tests added | 8 (vitest) |
| 🔴 must-fix | 0 |
| 🟡 should-fix | 0 |
| 🟢 observations | 3 (limitations documented in script) |
| Performance | 0.526s wall-clock on 53-module bundle at depth 8 (budget: 5s) |
| Stage 11 reviewer empirical canary | confirmed walker would have caught #1370 pre-merge |

## What Worked Well

1. **Empirical canary in Stage 11** — the reviewer subagent didn't just trust the code; it temporarily flipped `var _activeHooks` → `let _activeHooks` in a bundle copy, ran the lint, and confirmed the exact #1370 transitive chain was caught at depth 3. That's the highest-confidence validation a tooling PR can get. Tautology-test detection (Action #1200) also explicitly verified: the same synthetic bundle that flags under default depth does NOT flag under `--shallow-only`, proving the deep walker is load-bearing rather than redundant.
2. **Effective-line model insight in the implementation** — the subagent designed for the 16-FP failure case explicitly, rather than implementing naive depth-N first and patching FPs reactively. The key trick (`decl.bundleLine > rootCallLine` instead of lexical comparison) eliminated all 16 FPs in one pass.
3. **Two-commit shape held cleanly** — Gate 1 and Gate 2 both passed without intervention. Pre-commit hooks ran clean on first attempt this time (no cargo-fmt iteration like PR #1454 had).
4. **Backwards-compat via `--shallow-only`** — preserving v0.9.5 behavior behind a flag means anyone hitting unexpected FPs has an immediate escape hatch. Reduced merge risk.

## What Didn't Work

1. **Walker design carried "Array.prototype.forEach NOT modeled as sync executor"** — this is a known limitation that could produce false negatives. The reasoning (real synchronous-iteration callbacks that touch late lets are theoretical) is sound, but it's a trade-off worth documenting in canon as "scope decision: depth-N walker is best-effort for sync iterators; runtime test is the safety net for that class."

## What to Improve

1. **Canon: walker false-negative class needs explicit documentation** — Open Action: `tech-debt: document that bundle-init-order lint deliberately does not descend through Array.prototype.forEach / map / filter callbacks; runtime test bundle-init-no-tdz.test.js is the safety net for sync-iterator transitive TDZ`.

   **Why**: a future reviewer or developer hitting a runtime TDZ from a `.forEach` callback path would not understand from the lint output why the lint stayed silent. Documenting the design boundary explicitly closes that knowledge gap.

   **How to apply**: add a "Known false-negative classes" subsection to the docstring at the top of `scripts/check-bundle-init-order.mjs`, and reference it from the Limitations section of the lint's terminal output (or a `--help` block).

None at the milestone-finding level — the walker semantics are documented in-script comprehensively.

## Recurring Issue Tracker

|---|---|
| #1046 (doc-claim verbatim TDD) | CHANGELOG accurately describes implementation; Stage 11 reviewer verified all claims (depth-8 default, all listed deferral patterns implemented, `--shallow-only` truly reverts to v0.9.5) |
| #1200 (tautology test detection) | Reviewer ran the synthetic #1370 bundle under both shallow and deep modes; deep walker confirmed load-bearing |
| #1450 (Stage 11 stale-base check) | Worked — BEHIND=0 confirmed before review |
| #1122 (split-foundation pattern) | N/A — this is a single self-contained tool extension, no API contract being soaked |
| #181 (two-commit shape) | Held cleanly: impl `88547bb7`, docs `35e6a569`. Gate 1/Gate 2 both passed first try |

PR #1469, opened 2026-05-12
docs(canon): gate-the-change-off tautology self-test (#1468)

johnrtipton:
## Summary

Canonicalizes the "gate-the-change-off" tautology self-test for implementer subagents and reviewers. Single docs-only commit; no CHANGELOG entry (internal canon, no user-visible change).

Closes #1468 (in-repo scope).

## What

After writing new tests + verifying they pass, temporarily revert the change under test (set a flag to False, comment out the new behavior, gate on always-False conditional). Re-run. Confirm AT LEAST ONE test fails — preferably the most behavior-meaningful one. Restore. If all tests still pass with the change gated off, AT LEAST ONE test is tautological — fix before reporting tests pass.

Applies at Stage 5 (implementer) AND Stage 11 (reviewer). Generalizes Action #1200 from reviewer-only to implementer-too.

## Where

- **`docs/PULL_REQUEST_CHECKLIST.md` Test Quality**: new bullet with the concrete requirement + reference to PR #1466 case study.
- **`CLAUDE.md` new section "Process canonicalizations from v0.9.7-2 retro arc"**: full case study from PR #1466 (4 of 7 first-pass tests caught as tautological at Stage 11; Stage 13 fix-pass replaced one with real `WebsocketCommunicator` integration test; verified non-tautological by empirically gating the save off and seeing the new test fail with the expected assertion).

## Out-of-repo (NOT this PR)

The implementer-subagent prompt template lives in the pipeline-run skill repository. A Verification-section step calling out the gate-off self-test should be added there. Tracked as remaining work on #1468.

## Verification

- [x] PR-checklist bullet adds concrete requirement
- [x] CLAUDE.md backfill case study cites real PR (#1466) + empirical 4/7 rate
- [x] All citations grep-verified (PR #1466 exists; PR #1429 exists; Action #1200 verified at CLAUDE.md)
- [x] Gate 2 (docs-only commit) passes
- [ ] CI green

Closes #1468

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #1468, opened 2026-05-12
tech-debt: implementer-subagent prompt should mandate gate-the-change-off tautology self-test before reporting tests passing


johnrtipton:
## Source

Retro v0.9.7-2, finding #2. Surfaced on PR #1466 where 4 of 7 first-pass tests were tautological (didn't exercise the actual runtime under test).

## Failure mode

When delegating implementation + tests to a subagent, the first-pass test suite often includes tests that pass for the wrong reason. Specifically: tests that mock the path under test, or proxy via an unrelated path, or reproduce the implementation's logic in pytest rather than exercising the implementation itself. Action #1200 (tautology test detection) is the canonical defense, but it's currently applied at **Stage 11 review time** — the reviewer subagent gates the change off, re-runs, and sees which tests fail. By that point, the implementer has already reported "tests pass" and the work is one PR-cycle deep.

PR #1466 was the canonical example: subagent reported 7/7 passing; reviewer found 4 of 7 were tautological; Stage 13 fix-pass replaced one with a real `WebsocketCommunicator` integration test. Net cost: 1 fix-pass cycle + ~30 min wall-clock.

## Fix

Amend the per-PR implementer-subagent prompt template's "Verification" section to include a **gate-the-change-off self-test step** BEFORE reporting test results:

```markdown
## Verification

[existing test-run commands]

### Tautology self-test (Action #1200)

After writing tests + verifying they pass, temporarily revert the change
under test (gate it off, set a flag to False, comment out the new
behavior, etc.). Re-run the tests. Confirm AT LEAST ONE of the new tests
fails — preferably the most behavior-meaningful one. Restore the change.
Report the gate-off failure as part of the verification output, alongside
the all-pass result.

If all tests still pass with the change gated off, AT LEAST ONE test is
tautological — fix it before reporting.
```

This is a small prompt edit with a real signal-to-noise improvement on test quality. It also catches the case where the subagent wrote tests against a stale understanding of the file (tests reproduce logic-as-imagined rather than logic-as-implemented).

## Where to apply

The implementer-subagent prompt template lives implicitly in the executor's `Agent({...})` calls. There isn't a single canonical file. Two options:

- **(a)** Add a "subagent prompt canonical sections" doc at `docs/development/subagent-prompts.md` with the Verification section as a reusable template. Executor copies + adapts.
[...]

PR #1462, opened 2026-05-12
test(wire-protocol): pin 5 optional-feature frames (#1456 Batch 2)

johnrtipton:
## Summary

#1456 Batch 2 of 3. Pins 5 optional-feature wire-protocol frames following PR #1457 + PR #1461's pattern. 6 new tests; 19 → 25 in `test_wire_protocol_snapshots.py`.

## Shapes pinned

| # | Frame | Emit site | Variants |
|---|---|---|---|
| 1 | `i18n` | `websocket.py:729-734` | cmd-spread |
| 2 | `accessibility` | `websocket.py:749-754` | announcements list |
| 3 | `focus` | `websocket.py:764-770` | selector + options |
| 4 | `html_update` | `websocket.py:1125-1144` | minimal + with `reset_form`/`event_name` conditional appends |
| 5 | `connect` | `websocket.py:1424-1428` | session_id |

## Key-order pinning

`html_update` has 6 conditional appends after `version` (`reset_form`, `cache_request_id`, `async_pending`, `event_name`, `source`, `ref`). This PR pins a representative-case (`reset_form` + `event_name`) to lock the order-by-insertion contract. The remaining 4 individual conditional cases are deferred to Batch 3 if individual order-pinning matters; for now, the contract is "appends-after-version-in-insertion-order".

## Verification

```
$ .venv/bin/python -m pytest python/djust/tests/test_wire_protocol_snapshots.py
25 passed in 0.07s
```

## Follow-up #1456 remaining

After Batches 1+2, ~12 shapes left for Batch 3:
- Upload frames (`upload_registered`, `upload_resumed`, `upload_progress`)
- Reload variants (`reload`, `hvr-applied`)
- Control plane (`rate_limit_exceeded`, `pong`, `navigate`, `noop`)
- Presence (`presence_event` in `presence.py`)
- Streaming (`streaming.{patch,html_update,stream}`)
- `error.message` variant + remaining conditional `patch`/`html_update` keys

## Two-commit shape (Action #181)

- `4ca5a7d2` — implementation (6 new tests)
- `7022b0d1` — docs (CHANGELOG `[Unreleased]` Tests entry)

Gate 1 + Gate 2 ✅. Action #122 (post-commit verification) caught a pre-commit ruff-reformat bounce on first attempt; re-staged + re-committed cleanly. **3rd consecutive PR hit by this trip class** — strengthens the case for tackling #1458's three implementation options.

## Test plan

- [x] 25 tests pass
[...]

PR #1457, opened 2026-05-12
test(wire-protocol): pin 8 highest-value Python-emitted frame shapes (#1448 starter)

johnrtipton:
## Summary

Generalizes PR #1444 (Rust `Patch`/`VNode` snapshot pinning) to the Python side. Pins the canonical JSON shape of the 8 highest-value WebSocket frames djust emits server-side.

**Closes** #1448 (the starter scope). **Follow-up issue** #1456 tracks the remaining ~22 frame shapes across 2-3 grouped batches.

## Why a starter and not exhaustive

Investigation showed the wire-protocol surface is ~30+ frame shapes (websocket.py 28, streaming.py 3, presence.py 1), far more than the ROADMAP's "5 contracts" estimate. Per Decisions 3 + 5 in `context/history/v0.9.6-2-autonomous-decisions.md`, doing all 30 in one PR risks the "messy 1 PR" outcome. This PR pins the 8 highest-value shapes; #1456 tracks the rest.

## Frame shapes pinned

| # | Frame | Emit site |
|---|---|---|
| 1 | `push_event` | `websocket.py:436-442` |
| 2 | `flash` | `websocket.py:457-462` |
| 3 | `page_metadata` | `websocket.py:476-481` |
| 4 | `patch` (envelope) | `websocket.py:1074-1090` |
| 5 | `mount` | `websocket.py:2283-2300` |
| 6 | `layout` | `websocket.py:547-551` |
| 7 | `navigation` | `websocket.py:712-720` |
| 8 | `error` | `websocket.py:794` |

12 tests total (8 shapes × 1-2 variants each — `push_event` empty payload, `patch` html-fallback, `mount` with `public_state`, `error` with context-key spread).

## Why pin these wire shapes

Same value as PR #1444 (Rust side): a `json.dumps`-derived contract's JSON shape is invisible to type checkers. A field rename or default-value change at the emit site silently breaks every deployed client running an older bundle. Literal-string snapshot pins make breakage loud at test-time.

**Key-order pinning is deliberate**: `mount.public_state` and `patch.html` are appended AFTER the initial dict literal (`response["public_state"] = ...`; `response["html"] = ...`). Python 3.7+ preserves dict insertion order, so JSON order is deterministic. A refactor that moved the optional append into the initial literal would fail loudly.

## Verification

```
$ .venv/bin/python -m pytest python/djust/tests/test_wire_protocol_snapshots.py
12 passed in 0.06s
```

## Two-commit shape (Action #181)

[...]

PR #1470, opened 2026-05-12
feat(tooling): pre-commit auto-restage commit wrapper (#1464)

### LINKED ISSUE #1464: tech-debt: implement chosen fix for pre-commit ruff auto-restage (#1458 follow-up)

johnrtipton:
Closes #1464.

## Summary

Adds `scripts/git-commit-with-precommit.sh` (and `make commit MSG="..."` target) — an opt-in commit wrapper that runs `uvx pre-commit run --files <staged>` first, auto-`git add`s any files the hooks rewrote, then performs the `git commit`. Eliminates the ruff-bounce friction class.

## Empirical motivation

5 Action #122 ruff-bounces caught across v0.9.7-2's PRs (#1454, #1457, #1462, #1463, #1466). Each cost ~30s of re-stage + re-commit, AND each one risked silent commit-drop if Action #122's `&& git log -1 --oneline` reflex was missed.

## What changed

- `scripts/git-commit-with-precommit.sh` (new, executable) — wrapper script. Captures content hashes of staged files BEFORE pre-commit (tempfile-backed; macOS bash 3.2 compatible — no `declare -A`). If pre-commit exits non-zero AND rewrote at least one file, re-stages and proceeds; otherwise propagates the lint failure. Post-commit Action #122 verification (HEAD advanced) built in.
- `Makefile` — new `make commit MSG="..."` target invokes the wrapper.
- `CONTRIBUTING.md` — opt-in usage note in the dev-setup section.
- `tests/test_git_commit_with_precommit.py` — 5 regression tests using a synthetic `uvx` PATH shim that simulates a ruff-format bounce against a throwaway git repo.

## Acceptance (from the issue)

- ✅ After ruff reformats a staged file, wrapper succeeds on first attempt (no re-stage needed). Verified by `test_wrapper_recovers_from_ruff_bounce` AND by dogfooding this PR's own two commits.
- ✅ Test exists in `tests/` demonstrating recovery.
- ✅ Action #122's post-commit verification still passes — built into wrapper output (`git-commit-with-precommit: OK <hash> <subject>`).
- ✅ Bare `git commit` path unchanged — wrapper is opt-in.

## Gate-off self-test (Action #254 / #1468, just canonized in PR #1469)

Verified non-tautological:
1. Backed up wrapper, replaced `git add \$STAGED` with `: # sabotaged`.
2. Re-ran tests → `test_wrapper_recovers_from_ruff_bounce` FAILED (`working tree not clean: ' M foo.py\n'`) — the behavior-meaningful assertion caught the regression.
3. Restored wrapper → all 5 tests pass.

The existence-check tests (`test_wrapper_skipped_when_wrapper_missing`, `test_bash_available`) intentionally pass under sabotage — they verify infrastructure, not behavior. The recovery test is the load-bearing one.

- Commit 1 (75bd8df6): impl + tests + Makefile + CONTRIBUTING. No CHANGELOG.
- Commit 2 (bbc11606): CHANGELOG only.

Both commits made via the wrapper itself (true dogfood — the wrapper committed its own implementation under live pre-commit hooks).

## Test plan

- [x] `pytest tests/test_git_commit_with_precommit.py` — 5/5 pass
- [x] Gate-off self-test verified (above)
- [x] Wrapper dogfooded for both commits in this PR


johnrtipton:
## Stage 11 Review — PR #1470 (closes #1464)

Stale-base: 0 commits behind origin/main. Two-commit shape (#181): commit 1 has no CHANGELOG, commit 2 only CHANGELOG. Clean.

### 🔴 Must-fix

(none — the core logic is correct and the canonical bounce-recovery path is empirically tested with a gate-off-equivalent assertion: `assert status == ""` would fail if `git add $STAGED` were a no-op, satisfying Action #1200/#254.)

### 🟡 Should-fix

1. **Filename whitespace/glob unsafety in `git add $STAGED` and `--files $STAGED`** (`scripts/git-commit-with-precommit.sh:54, 80`). `STAGED` is newline-separated and word-splits on IFS, including spaces. A staged filename like `my file.py` (or one containing `*`, `?`, `[`) is silently corrupted — the wrapper would `git add` two non-existent paths instead of the real one, leaving the reformat unstaged and the working tree dirty. The shellcheck disable comments acknowledge intent but the trade-off isn't defensible: switch to `git diff --cached -z --name-only --diff-filter=ACMR | xargs -0 ...` (or read into a bash array with `mapfile -d ''`). This is the failure mode the wrapper was built to prevent; ironic if it ships with the same shape.

2. **Missing test coverage for the multi-file / partial-stage / whitespace cases.** The 5 tests all exercise single-`.py`-file scenarios. Add at least: (a) two staged files where only one is rewritten, (b) a filename containing a space (will fail today — locks the fix above in), (c) `git add -p` partial-stage where the unstaged hunk must NOT be promoted by the auto-restage. The third case is non-obvious — `git add $STAGED` (path-mode) would re-stage the entire working-tree file, silently swallowing unstaged hunks. Consider `git add -u` scoped to the rewritten paths, or compare to pre-hash and only `git add` the changed paths.

### 🟢 Nice-to-have

1. **Double hook invocation cost not surfaced.** Wrapper runs `uvx pre-commit run` then `git commit` runs the installed hook a second time (~2× total). Worth a one-line note in the script header or CONTRIBUTING.md so users with slow hook suites aren't surprised. Alternative: `git commit --no-verify` after a successful wrapper-side run (with a comment explaining why bypass is safe here).

2. **Non-git-repo path is noisy.** Outside a repo, `set -e` + the first `git diff --cached` exits 129 with raw git usage text before reaching the friendly "no files staged" branch. A pre-check (`git rev-parse --git-dir >/dev/null 2>&1 || { echo "not a git repo" >&2; exit 1; }`) would tighten the UX.

3. **`make commit MSG="..."` quoting surface.** Backticks/`$()` in MSG will be evaluated by the shell during `@scripts/...sh -m "$(MSG)"` expansion. Low-risk (caller is the user), but worth a one-line warning in CONTRIBUTING.md.

### ✅ Positive notes

- Two-commit shape (Action #181) followed exactly — verified via `git show --name-only` on both SHAs.
- macOS bash 3.2 compatibility called out in code comment AND respected (no `declare -A`, no `[[`).
- Hash-comparison uses `git hash-object` — same algorithm `git add` writes to the index, so the pre/post-hook comparison is exact.
- `mktemp` + `trap 'rm -f' EXIT` properly cleans the hash sidecar.
- Post-commit Action #122 verification (`HEAD` advanced) built in and given its own exit code (4).
- Test isolation via `GIT_CONFIG_GLOBAL=/dev/null` + per-test `PATH` shim is well-designed and gate-off-meaningful: sabotaging `git add $STAGED` in the recovery path makes `test_wrapper_recovers_from_ruff_bounce` fail at `assert status == ""`.
- CHANGELOG entry accurately describes the script (verified: 5 tests, bash 3.2 compat, Action #122 verification, PR citations #1454/#1457/#1462/#1463/#1466/#1469 all resolve as MERGED — spot-checked 3).
- Exit-code documentation in header (0/1/2/3/4) matches all five reachable code paths.

VERDICT: APPROVE_WITH_NITS

johnrtipton:
## Stage 13 Re-Review (PR #1470 / #1464)

- **🟡 #1 (filename whitespace/glob unsafety): ADDRESSED.** Script now uses `git diff --cached -z --name-only --diff-filter=ACMR` piped into `while IFS= read -r -d '' f` (macOS bash 3.2-compatible array build), `uvx pre-commit run --files "${STAGED[@]}"` (quoted array expansion), and `git add -- "${REWRITTEN[@]}"` with `--` separator. Sanity grep `grep -nE 'git add \$|--files \$' scripts/git-commit-with-precommit.sh` returns zero matches.
- **🟡 #2 (test coverage gaps): ADDRESSED.** All three regression cases exist and pass: `test_wrapper_multi_file_partial_rewrite` (asserts `"rewrote 1 file"`, locking per-file diff), `test_wrapper_filename_with_space` (`my file.py`), and `test_wrapper_preserves_unstaged_hunks` (exercises REWRITE branch with `staged.py` un-canonical and `other.txt` unstaged; asserts commit contains ONLY `staged.py` AND `other.txt` remains ` M`).
- **Per-file restage logic: CONFIRMED.** `REWRITTEN=()` populated only when `OLD_HASH != NEW_HASH` per file; `git add --` receives just that subset.
- **🟢 bonuses: ADDRESSED.** Non-git-repo precheck (`git rev-parse --git-dir`) emits friendly message, exits 1. Double-hook 2× cost documented in script header comment.
- **Test suite: 9 passed in 2.00s.** CHANGELOG "9 regression tests" claim matches actual test count (9 `def test_` functions).

**Test count: 9**

VERDICT: APPROVE

johnrtipton:
## Per-PR Retro — #1470 (#1464 pre-commit auto-restage wrapper)

**Quality**: 5/5 (Stage 11 0🔴 / 2🟡 / 3🟢; both 🟡s + 2 of 3 🟢s addressed at Stage 13; merge clean).

### What Worked Well

- **Empirical motivation drove a sharp scope.** The 5 observed ruff-bounces from v0.9.7-2 (PRs #1454, #1457, #1462, #1463, #1466) gave the PR description a load-bearing concrete number rather than abstract "this is annoying" framing. Stage 11 reviewer spot-checked the citations and 3/3 resolved.

- **Dogfooded the wrapper to commit itself.** Both commits in this PR went through `scripts/git-commit-with-precommit.sh`. Pre-commit hooks (ruff-format, ruff-check, trailing-whitespace) fired live on real diffs and the wrapper handled them. Production-equivalent shakedown without spinning up additional infra.

- **Gate-off self-test (Action #254 / #1468 canon) caught a subtle test pattern.** The Stage 5 self-test sabotaged `git add $STAGED` → recovery test failed at `assert status == ""` (working tree dirty). At Stage 13 the same discipline re-applied to the per-file restage: sabotaged to `git add -A` → partial-stage test failed at `assert files == "staged.py"` (other.txt swept up). Both regressions caught by the load-bearing assertion, not the existence checks. First non-trivial application of the canon canonized one PR earlier (#1469) — it landed at exactly the right shape.

### What Didn't Work

- **First-pass test coverage was thin (5 single-file tests).** Stage 11 reviewer flagged missing multi-file, whitespace-in-name, and partial-stage cases. Test count went 5 → 9 at Stage 13. Pattern: implementer-subagent (and human implementer here) defaults to the canonical happy-path test set; edge-case test coverage benefits from a reviewer second pass. Generalizable: when writing tests for a recovery-class tool, ask up front "what shapes of input would break the recovery?" and enumerate before writing.

- **`git add $STAGED` (unquoted) shipped in the first-pass implementation despite the wrapper existing specifically to handle ruff-bounce friction**. The reviewer caught the irony: the wrapper would fail on a path with spaces, reproducing the exact failure mode it was written to prevent. NUL-delimited handling (`git diff --cached -z` → bash array → quoted `"${arr[@]}"`) is the standard fix and should be the default pattern for any new shell tool that processes git output. Worth canonicalizing if it recurs.

### What to Improve

- **Default-pattern catalog for shell tools that read git output.** Specifically: `git diff -z` + bash array read + quoted array expansion + `--` separator on `git add`. This is the third instance in djust's tooling where unquoted git-output word-splitting was the regression vector (the other two: `roadmap-lint.py` and earlier in `check-handler-contracts.py` history). Worth a `docs/PULL_REQUEST_CHECKLIST.md` Stage 7 self-review bullet: "For shell tools that process `git diff --name-only`, verify -z + array read + quoted expansion."

- **Reformatter-shim test pattern reusable**. The synthetic `uvx` PATH-shim that simulates a ruff-format bounce against a throwaway git repo is a reusable testing primitive for any future pre-commit-related tooling. Could be lifted to a shared `tests/_helpers/precommit_shim.py` if a second user lands.

### Knowledge Base Updates

- None canonized in this PR; the load-bearing canon (gate-off self-test) was just landed in #1469 and this PR served as the first empirical validation of it. The "shell tools default to -z + array" pattern observed here is worth tracking — filed as a "What to Improve" item in this retro.

Issue #254, opened 2026-02-09
Add remaining 13 Django built-in template filters


johnrtipton:
## Context

With #246 (PR incoming), djust implements 44 of Django's 57 built-in template filters. This issue tracks the remaining 13.

### High value (commonly used)
| Filter | Behavior | Complexity |
|--------|----------|------------|
| `escapejs` | Escape string for use in JavaScript | Low |
| `linenumbers` | Add line numbers to text | Low |
| `get_digit` | Return the Nth digit from right | Low |
| `iriencode` | IRI-encode a string (like urlencode but for IRIs) | Low |

### Medium value
| Filter | Behavior | Complexity |
|--------|----------|------------|
| `urlize` | Convert URLs in text to clickable `<a>` links | High |
| `urlizetrunc` | Like `urlize` but truncate link text to N chars | High |
| `truncatechars_html` | Truncate chars, preserving HTML tags | Medium |
| `truncatewords_html` | Truncate words, preserving HTML tags | Medium |
| `safeseq` | Mark each item in a sequence as safe | Low |
| `escapeseq` | Escape each item in a sequence (Django 5.0+) | Low |

### Low value (rarely used)
| Filter | Behavior | Complexity |
|--------|----------|------------|
| `unordered_list` | Render nested list as `<ul><li>` HTML | Medium |
| `phone2numeric` | Convert phone letters to digits (e.g., 1-800-COLLECT → 1-800-2655328) | Low |
| `pprint` | Pretty-print a value (debugging) | Low |

## Suggested approach

The low-complexity filters (`escapejs`, `linenumbers`, `get_digit`, `iriencode`, `safeseq`, `escapeseq`, `phone2numeric`, `pprint`) are straightforward match arms in `crates/djust_templates/src/filters.rs`.

`urlize`/`urlizetrunc` require a URL detection regex — the most complex of the bunch.

`truncatechars_html`/`truncatewords_html` need an HTML-aware truncation that counts visible characters while preserving and closing open tags.

PR #1478, opened 2026-05-13
fix(websocket): gate WS-event save on enable_state_snapshot (#1475 P0)

### LINKED ISSUE #1475: 0.9.7rc2: WS state-snapshot close path captures uvicorn in unrecoverable asyncio state when host snapshots within ~200ms
## Summary
djust 0.9.7rc2 (the WS-path \`enable_state_snapshot\` feature, PR #1466 / closes #1465) made the snapshot/wake story work at the application layer — LiveView state can survive a WebSocket reconnect. But on djustlive (the PaaS) it surfaced a regression: when the host's microVM hypervisor (firecracker) takes a snapshot of the guest shortly after djust's WS close path runs, uvicorn ends up in an unrecoverable asyncio-deadlocked state. Post-restore, uvicorn's TCP listener accepts connections but never returns HTTP response bytes. **Forever.**

## Where this matters
- djustlive (and any other microVM PaaS using firecracker snapshots) takes snapshots when a placement goes idle.
- The snapshot-on-idle path closes all backend WS conns at the host TCP layer, then pauses + snapshots the guest within microseconds (closeAllBackendConns is immediately followed by PATCH /vm Paused).
- djustlive's fc-proxy applies a **200ms settle delay** between closeAllBackendConns and pause to let the guest's asyncio drain. That was sufficient for djust 0.9.6 but is NOT sufficient for djust 0.9.7rc2.

## Reproduction
1. djustlive-node-1 (AMD Ryzen 9 3900, fc 1.11.0, kernel 5.10.225).
2. Scaffold rootfs with \`djust[components,theming]>=0.9.7rc2\`.
3. fc-proxy with \`--idle 10s\` and 200ms settle delay post-closeAllBackendConns.
4. Open WS to \`/ws/live/\` via Caddy → fc-proxy → uvicorn.
5. Wait > 10s for idle to fire. fc-proxy closes WS, settles 200ms, pauses, snapshots, kills firecracker.
6. Send HTTP request. fc-proxy restores VM in ~5ms, then runs HTTP probe (\`HEAD /\`) against uvicorn.
7. **uvicorn never responds.** Probe times out at 3 minutes (our deadline). Restore declared failed. Next request triggers another restore — same poison snapshot, same hang. Loop forever.

## Wire-level evidence
\`tcpdump\` on the tap device during the hang shows:
\`\`\`
host → guest: SYN
guest → host: SYN-ACK
host → guest: ACK
host → guest: PSH (117 bytes — HTTP HEAD request)
guest → host: ACK
[silence — no response data ever]
host → guest: FIN
guest → host: ACK
\`\`\`
Repeated for each probe attempt. Kernel-side TCP is perfect. uvicorn just doesn't generate response bytes.

## Comparison with 0.9.6
Same fc-proxy code, same firecracker config, same kernel, same hypothetical settle delay (200ms). The ONLY difference is which djust the scaffold's rootfs has installed.
- djust 0.9.6: works. Median post-restore first request 50ms, max ~840ms (across 180 probes).
- djust 0.9.7rc2: deterministic uvicorn-frozen state after snapshot. Site down indefinitely on idle cycles.

## Hypothesized cause
djust 0.9.7rc2's new WS state-snapshot code path runs additional asyncio work during the WS close handler — serializing the LiveView state for the snapshot, possibly involving \`await\`s on tasks that depend on the WS being alive. When the host kills the WS conn at TCP layer (host's snapshot path needs the conn closed so gorilla.ReadMessage doesn't hang forever on the proxy side), djust's asyncio close handler is mid-flight. snapshot fires before the cleanup completes, capturing asyncio in a state that contains pending awaits on things that no longer exist post-restore.

I haven't read 0.9.7rc2's WS-close code in detail — could you confirm/deny whether this matches what \`enable_state_snapshot\` does on the WS path?

## What we'd need on djust's side
Either (in order of cleanness):
1. **A sync drain signal.** djust exposes a known endpoint or sentinel a host operator can poll: "is WS state-snapshot work complete on all closed conns?" fc-proxy polls that before pausing. Bounded wait, no fixed-delay races.
2. **A guaranteed-time-bound WS close.** djust commits to a maximum time the WS close handler will take (e.g., 200ms p99 even under contention). fc-proxy's existing settle covers it. Same shape as today, but enforced.
3. **WS-state-snapshot moves OFF the close path.** Snapshot happens during normal operation (proactive), not at close. WS close becomes free again.
4. **An async hook djust can signal when ready.** djust calls a callback / signal when its async work is done; fc-proxy waits on that.

## What we did on djustlive's side
- Reverted the scaffold rootfs to djust 0.9.6 (immediate recovery).
- Added a cold-boot fallback to fc-proxy (PR #302) so future snapshot-poisoning regressions don't infinite-loop — fc-proxy detects 2 consecutive restore failures and forces a cold-boot from rootfs instead of looping on the bad snapshot. Recovery latency ~85s instead of ∞.
- Filed this issue.

## Importance
Blocks djustlive on djust 0.9.7. The platform's whole scale-to-zero LiveView story depends on snapshot/wake working, and \`enable_state_snapshot\` on the WS path is what makes the application-layer story work. We need a coordinated solution.

Happy to provide more diagnostic data or run experiments with patched builds.

johnrtipton:
Closes #1475. **P0 — production-blocker on djustlive.**

## TL;DR

PR #1466's save block in `handle_event` ran unconditionally for every WS event for every LiveView. djustlive's fc-proxy snapshots a firecracker microVM 200ms after closing backend WS conns; the 0.9.7rc2 save block left async session-backend I/O in flight that gets captured by the snapshot → uvicorn frozen forever post-restore.

Fix: AND the existing top-level-identity gate with `getattr(self.view_instance, "enable_state_snapshot", False)`. Defense-in-depth: 150ms `asyncio.wait_for` around the save body.

## Investigation

Source-traced #1475's hypothesis to confirm root cause:
- PR #1466 added the save block at `python/djust/websocket.py:3127-3214`.
- Only gate was `target_view is self.view_instance` (top-level identity, NOT opt-in).
- The PR's CHANGELOG entry called this out as deliberate HTTP-path symmetry.
- That symmetry argument doesn't survive contact with snapshot-on-idle infrastructure: HTTP requests bounded by response cycle, WS events leave async I/O in flight beyond `send_json`.
- 0.9.6 had no save block → 0.9.7rc2 added unconditional saves → djustlive's 200ms settle no longer sufficient.

Confirmed via investigation comment on #1475: https://github.com/djust-org/djust/issues/1475#issuecomment-4435783271

## Fix

**Primary** (`python/djust/websocket.py`):
```python
if (
    target_view is self.view_instance
    and getattr(self.view_instance, "enable_state_snapshot", False)
):
    ...
```
Restores 0.9.6 behavior for default views. Opt-in views get the feature they asked for.

**Defense-in-depth**: save body wrapped in `asyncio.wait_for(..., timeout=0.150)`. Even opt-in views can't extend close-time tail latency under DB/Redis backpressure. On timeout, log warning and continue.

## Tests

Existing 7 + 3 new = 10 total.

**New (load-bearing)**: `test_ws_event_save_block_skipped_when_state_snapshot_opt_out` — real WebsocketCommunicator drives mount + event for a view WITHOUT `enable_state_snapshot`, then reads the session backend directly and asserts the `liveview_<path>` key is ABSENT. This is the negative assertion that locks in the #1475 fix.

**Source pins**:
- `test_save_block_gates_on_enable_state_snapshot_source` (new, dedicated)
- `test_save_block_present_in_handle_event_source` (existing, extended)

**Timeout semantics**: `test_save_block_timeout_logs_and_continues` validates `asyncio.wait_for` raises TimeoutError after 150ms on a 250ms body.

### Action #254 gate-off self-test

Sabotaged the gate (reverted `and getattr(...)` to PR #1466 behavior):
- 3 tests failed: the opt-out integration test (`assert saved is None` → caught), the dedicated source pin, the extended source pin.
- 7 tests passed under sabotage (testing unrelated invariants).
- Restored → all 10 pass.

Behavior-meaningful regressions caught at load-bearing assertions, not infrastructure checks.

## Release planning

Cuts as part of v0.9.7rc3 (immediate). Supersedes 0.9.7rc2. djustlive can revert its 0.9.6 rootfs pin to 0.9.7rc3 once published.

## Test plan

- [x] `pytest python/djust/tests/test_ws_reconnect_state_1465.py` — 10/10 pass
- [x] Action #254 gate-off self-test verified
- [x] Two-commit shape: impl + tests in commit 1, CHANGELOG in commit 2


johnrtipton:
## Stage 11 review — PR #1478 (closes #1475, P0)

Branch is **0 commits behind origin/main** (Action #1450 stale-base gate PASSED). Applied higher-than-usual scrutiny per the P0 framing.

### Two-commit shape (Action #181) ✅

- `b1d1804b` touches only `python/djust/websocket.py` + `python/djust/tests/test_ws_reconnect_state_1465.py`. No CHANGELOG. ✅
- `ad483849` touches only `CHANGELOG.md`. ✅

### Gate correctness ✅

```python
if target_view is self.view_instance and getattr(
    self.view_instance, "enable_state_snapshot", False
):
```

Correctly AND'd. `target_view` is a local set exactly twice in `handle_event` (lines 2689, 2696) and never mutated thereafter — closure capture into `_persist_state_after_event` is sound. Edge case: if `self.view_instance` were `None`, the identity check passes (None is None for the top-level branch) but `getattr(None, "enable_state_snapshot", False)` returns `False` → short-circuit skips the block. Safe.

### CHANGELOG accuracy ✅

Verified: PR #1466 (`a5e2c50c`) introduced the save block; this is what the CHANGELOG entry claims. The operator-facing line ("revert the 0.9.6 rootfs pin to 0.9.7rc3 once published") is clear and actionable.

### 🟢 Save block restructure — clean

Refactor into `_persist_state_after_event()` inner async closure is minimal-blast-radius. The body is byte-identical apart from converting the early `if save_session is not None:` arrow into a `return` (reduces indentation, equivalent semantically). No behavior change beyond the gate + timeout.

### 🟡 Findings

**🟡 #1 — `test_save_block_timeout_logs_and_continues` is tautology-adjacent (Action #1200 / #1468).**

The test calls `asyncio.wait_for(_slow_save(), timeout=0.150)` directly — not through the wrapper. It validates Python stdlib behavior, not that **the wrapper logs+continues on TimeoutError**. The docstring claims "the wrapper must cancel the inflight coroutine, log a warning containing 'exceeded 150ms', not raise", but none of those three claims is asserted — only that `wait_for` raises `TimeoutError` on a 250ms body. Sabotage check: if you removed the `except asyncio.TimeoutError: logger.warning(...)` branch from the wrapper, this test would still pass.

The two source-pin tests cover the **lexical** presence of `asyncio.wait_for` + `timeout=0.150` + `asyncio.TimeoutError`. The **runtime end-to-end** behavior of timeout → log → continue is uncovered. Recommendation for follow-up: drive a real `_OptInView` through `WebsocketCommunicator` with a stub session whose `aset` sleeps 250ms; assert (a) handle_event returns normally (no raise), (b) `caplog` captures a WARNING containing "exceeded 150ms" and the sanitized event name. Not a blocker — the source pins lock in the wrapper structure, and the failure mode (silent swallow vs. raise) is the same shape `wait_for` semantics test already validates. But Action #1468 (left-shift the gate-off self-test to Stage 5) explicitly calls this case out.

**🟡 #2 — 150ms budget is tight.**

The save body does up to 4 awaits: `aset(__private)`, `aset(view_key)`, `sync_to_async(_save_components_to_session)`, `asave()`. On a Django DB-backed session under typical Postgres latency, each is 5-15ms; healthy budget. But under cold Redis-cluster reconnect or DB contention, a single `aset` can spike to 50-100ms. The choice of 150ms is **deliberately tight to bound the snapshot-poisoning window** — this is the right trade-off for the P0 fix, but operators of opt-in views should expect occasional WARNING logs on busy hosts. Worth a docs note in a follow-up: "if you see frequent 'exceeded 150ms' warnings, your session backend is the bottleneck — `enable_state_snapshot` views' state may drift between events; subsequent events refresh." Cancellation safety: Django's `aset` on the DB session backend wraps a synchronous transaction via `sync_to_async`; cancelling the outer coroutine raises `CancelledError` in the wrapper but the inner sync transaction has already committed-or-rolled-back when control returns. Acceptable.

### 🟢 Test coverage

- `test_ws_event_save_block_skipped_when_state_snapshot_opt_out` — verified by reading: real `WebsocketCommunicator` against `LiveViewConsumer.as_asgi()`, drives mount + event, asserts `liveview_/default-counter/` is absent post-disconnect. Load-bearing negative assertion. Gate-off would flip `saved is None` to `saved == {"count": 1}`. ✅
- Source-pin tests on `enable_state_snapshot`, `asyncio.wait_for`, `timeout=0.150`, `asyncio.TimeoutError` — all whitespace-tolerant via `source_collapsed`. Snake-case attribute pin is acceptable (Python convention).
- Existing `test_ws_event_save_block_writes_through_to_session` is unchanged — verified the POSITIVE case still passes for opt-in views.

### Production-readiness check

On-call lens: if this regresses, the failure surface is (a) gate accidentally inverted → opt-in views silently skip saves (caught by `test_ws_event_save_block_writes_through_to_session`), (b) gate accidentally removed → default views regress to rc2 behavior (caught by `test_ws_event_save_block_skipped_when_state_snapshot_opt_out` + source pin), (c) timeout fires under load → WARNING log + lost-for-this-event state (acceptable, subsequent event refreshes). All three surfaces have tests. The fix preserves rc1's opt-in feature for views that asked for it and reverts default views to 0.9.6 close-path semantics. Minimal-surface, well-scoped P0 fix.

### Summary

- 🔴 0
- 🟡 2 (timeout-test tautology shape; budget-tightness docs follow-up)
- 🟢 multiple positives
- ✅ checklist items all green

VERDICT: APPROVE_WITH_NITS

johnrtipton:
## Stage 13 Re-Review — PR #1478 (fix-pass for #1475 P0)

Verified commit `51acd4f1`.

### Findings status

- 🟡 #1 (tautology-adjacent timeout test) — **ADDRESSED**
  - New `test_save_block_timeout_logs_and_continues_end_to_end` drives `LiveViewConsumer.as_asgi()` via real `WebsocketCommunicator` (connect → mount → event → assert WARNING).
  - `_WSSlowOptInCounter.mount` injects `_SlowSession` whose `aset`/`asave` sleep 250ms (>150ms timeout).
  - Asserts `caplog` WARNING contains `"exceeded 150ms"` AND elapsed < 2.0s (bounded by wrapper, not slow body).
  - `_WSSlowOptInCounter` is module-level → `LIVEVIEW_ALLOWED_MODULES=[__name__]` resolves it.
  - **Gate-off sabotage proof**: removed the `except asyncio.TimeoutError` branch from `python/djust/websocket.py:3232-3238` → test FAILED with `AssertionError: Wrapper must log a WARNING containing 'exceeded 150ms'... assert 0 >= 1`. The raw `TimeoutError` propagated up (`asyncio/timeouts.py:115`) exactly as predicted in the test's docstring. Restored, all 10 tests green.
- 🟡 #2 (150ms budget tightness) — **still open as docs follow-up**, acknowledged not in scope for the P0.

### Source-pin update
`source_nospaces = "".join(source.split())` (all whitespace stripped) tolerates ruff-format wrapping the long if-condition across lines while still pinning the AND'd gate shape. Sensible — more robust than `source_collapsed`.

### Most important verification check
Gate-off sabotage of `except asyncio.TimeoutError` → new test FAILS with the load-bearing `len(warning_logs) >= 1` assertion. The test exercises the real wrapper path, not a stub.

VERDICT: APPROVE

PR #1486, opened 2026-05-17
fix(templates): support Django `is` / `is not` identity operators (#1483)

### LINKED ISSUE #1483: Template renderer: `is None` / `is not None` operators silently fail
## Summary

The Rust template renderer in `crates/djust_templates/src/renderer.rs:evaluate_condition` doesn't implement Django's `is` / `is not` operators. `{% if foo is None %}` and `{% if foo is not None %}` are parsed-but-not-recognized and the conditional silently returns false-equivalent for ALL values, including non-None ones.

## Repro

```django
{% if some_float is not None %}
  rendered: {{ some_float }}
{% else %}
  fallthrough
{% endif %}
```

With `some_float = 12.3` in context, this renders `fallthrough` instead of `rendered: 12.3`.

## Discovered

While building a metrics dashboard for djustlive — surfacing node CPU/mem/disk%. Values that COULD legitimately be None (first heartbeat tick before agent has a previous CPU sample to delta against) are checked with `is not None` per Django convention. The check silently failed for every value including non-None ones, leaving an empty UI. Took a 30-min round of "are the metrics in the DB?" / "is the agent reporting?" before catching that the template engine itself was the issue.

## Expected

Django 4.0+ supports `{% if foo is None %}` and `{% if foo is not None %}` as identity-test operators (distinct from `==` / `!=` which compare value-equality). djust documents "Django template compatibility" so callers reasonably reach for the official syntax.

## Implementation pointer

`evaluate_condition` in `crates/djust_templates/src/renderer.rs` (around line 1606) handles `or`/`and`/`not`/`==`/`!=`/`>=`/`<=`/`in` but has no branch for `is` / `is not`. Add:

```rust
// Handle "is" / "is not" identity tests (Django 4.0+).
if let Some(pos) = condition.find(" is not ") {
    let left = get_value(&condition[..pos], context)?;
    let right = get_value(&condition[pos + 8..], context)?;
    return Ok(!values_identity(&left, &right));
}
if let Some(pos) = condition.find(" is ") {
    let left = get_value(&condition[..pos], context)?;
    let right = get_value(&condition[pos + 4..], context)?;
    return Ok(values_identity(&left, &right));
}
```

…where `values_identity` matches Django's semantics: `None is None` → true, `5 is 5` → false (Django's `is` only works for the None/True/False singletons, not arbitrary values — this is the same as Python).

## Workaround (in our app)

Pre-format the metric values in the view layer:

```python
# Was: pass cpu_pct (float|None); template did `is not None`
# Now: pass cpu_pct_display (str: "5.2%" or "")
#      and cpu_pct_bar (float: 5.2 or 0)
"cpu_pct_display": f"{m['cpu_pct']:.1f}%" if "cpu_pct" in m else "",
"cpu_pct_bar":     m.get("cpu_pct") or 0,
```

…and use `{% if cpu_pct_display %}` (truthy check on the string) instead of `is not None`.

Workable but pushes presentation logic into views. The renderer fix is the right long-term answer.

johnrtipton:
## Summary

Closes #1483. **v1.0.0 milestone unit 1** (correctness gate). The Rust template renderer's `evaluate_condition` had no branch for Django's `is` / `is not` identity operators — `{% if x is None %}` and `{% if x is not None %}` fell through to the default `Ok(false)` and silently evaluated false for **all** values, always taking the `{% else %}` branch.

## The fix

`crates/djust_templates/src/renderer.rs`:
- New `values_identity` helper — Python identity semantics: true only for matching `None` / `True` / `False` singletons; `5 is 5` / `"a" is "a"` are **false** (CPython interning is not contractual).
- New `is` / `is not` dispatch branch in `evaluate_condition`, placed with the comparison-tier operators. `" is not "` is matched before `" is "` (substring containment); space-padded markers prevent false-matching variable names that contain "is".

Brings the Rust engine to parity with the Django (Python) engine, which has supported `is` / `is not` since Django 4.0. No Python-side change needed.

## Tests

- 12 Rust unit tests in `renderer.rs` (helper coverage, both operators, singleton/non-singleton, precedence, substring-no-false-match).
- `TestIsIdentityOperators` — 10 cases in `python/tests/test_template_conditions.py` (the Stage-4 reproducer; 7 fail pre-fix, all 10 pass post-fix). Gate-off self-test verified.
- Full suite green: 4290 Python pass, all Rust pass, 1572 JS pass.

## Note on `Cargo.lock`

The committed `Cargo.lock` had stale workspace versions (`0.9.7-rc.3`) vs `Cargo.toml` (`0.9.7`) on main; any `cargo` invocation regenerates it, so the `cargo-clippy` pre-commit hook bounced every commit until corrected. The one-line-per-crate `0.9.7-rc.3` → `0.9.7` correction is included in commit 1.

## Pipeline

Bugfix pipeline (state: `.pipeline-state/fix-template-is-none-operator-1483.json`). Two-commit shape: impl+tests, then docs+CHANGELOG.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- **Rust template renderer now supports Django's `is` / `is not` identity operators in `{% if %}` conditions (#1483).** `{% if x is None %}` and `{% if x is not None %}` previously fell through every operator branch in `evaluate_condition` to the default `Ok(false)`, so they silently evaluated false for *all* values — templates using this Django-standard syntax always took the `{% else %}` branch even when the condition was true. The renderer now implements `is` / `is not` with Python identity semantics: identity holds only for the singletons `None`, `True`, and `False`; arbitrary equal values (`5 is 5`, `"a" is "a"`) are NOT treated as identical (CPython interning is an implementation detail templates must not rely on). Templates that previously fell through to the `{% else %}` branch — e.g. `{% if some_value is not None %}` for a non-`None` value — will now correctly take the `{% if %}` branch. This brings the Rust template engine to parity with the Django (Python) engine, which has supported `is` / `is not` natively since Django 4.0. Regression coverage in `TestIsIdentityOperators` (`python/tests/test_template_conditions.py`) plus Rust unit tests in `crates/djust_templates/src/renderer.rs`.
+
 ## [0.9.7] - 2026-05-16
 
 Stable release. No code changes since v0.9.7rc3 (2026-05-12) — RC3 soaked for 4 days with djustlive on the pinned wheel and zero regressions reported. See the rc1, rc2, rc3 entries below for the full v0.9.7 changeset:
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -294,7 +294,7 @@ dependencies = [
 
 [[package]]
 name = "djust_components"
-version = "0.9.7-rc.3"
+version = "0.9.7"
 dependencies = [
  "ahash",
  "criterion",
@@ -309,7 +309,7 @@ dependencies = [
 
 [[package]]
 name = "djust_core"
-version = "0.9.7-rc.3"
+version = "0.9.7"
 dependencies = [
  "ahash",
  "criterion",
@@ -324,7 +324,7 @@ dependencies = [
 
 [[package]]
 name = "djust_live"
-version = "0.9.7-rc.3"
+version = "0.9.7"
 dependencies = [
  "bincode",
  "criterion",
@@ -349,7 +349,7 @@ dependencies = [
 
 [[package]]
 name = "djust_templates"
-version = "0.9.7-rc.3"
+version = "0.9.7"
 dependencies = [
  "ahash",
  "chrono",
@@ -370,7 +370,7 @@ dependencies = [
 
 [[package]]
 name = "djust_vdom"
-version = "0.9.7-rc.3"
+version = "0.9.7"
 dependencies = [
  "ahash",
  "criterion",
--- a/crates/djust_templates/src/renderer.rs
+++ b/crates/djust_templates/src/renderer.rs
@@ -1658,6 +1658,21 @@ fn evaluate_condition(condition: &str, context: &Context) -> Result<bool> {
         }
     }
 
+    // Handle Django identity operators "is" / "is not" (Django 4.0+).
+    // " is not " MUST be checked before " is " because the former
+    // contains the latter as a substring. Space-padded markers avoid
+    // matching variable names that merely contain "is" (e.g. "analysis").
+    if let Some(pos) = condition.find(" is not ") {
+        let left = get_value(condition[..pos].trim(), context)?;
+        let right = get_value(condition[pos + 8..].trim(), context)?;
+        return Ok(!values_identity(&left, &right));
+    }
+    if let Some(pos) = condition.find(" is ") {
+        let left = get_value(condition[..pos].trim(), context)?;
+        let right = get_value(condition[pos + 4..].trim(), context)?;
+        return Ok(values_identity(&left, &right));
+    }
+
     // Handle >= (must be before > to avoid false match)
     if condition.contains(">=") {
         let parts: Vec<&str> = condition.split(">=").map(|s| s.trim()).collect();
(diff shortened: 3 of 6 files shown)


johnrtipton (commented):
APPROVE — clean fix, well-scoped.

**Verification**: Stale-base BEHIND=0. Plan deliverables all implemented: `values_identity` helper (renderer.rs:1836), `is`/`is not` dispatch branch (renderer.rs:1665-1674) placed after `==`/`!=` and before `>=`/`in`. 12 Rust unit tests + 10-case `TestIsIdentityOperators` Python reproducer — all plan-specified tests present. `cargo test -p djust_templates` green; full Python suite 7045 passed / 0 failed / 20 skipped.

**Edge cases checked**:
- `" is not "` matched before `" is "` (line 1665 before 1670) — substring ordering safe; `test_if_is_not_checked_before_is` locks it.
- UTF-8 byte-offset: `pos+8`/`pos+4` slice past ASCII-only markers, always land on a char boundary regardless of surrounding bytes — safe.
- `5 is 5` → false (non-singleton); `Value::Null`/`Value::Bool` only — matches Python identity semantics.
- `x == "is"` resolves via `==` branch first — no operator collision.
- Undefined var `is None` → true (resolves to `Value::Null`) — verified correct manually.

**Concern (non-blocking)**: Stage-7 🟡 (undefined-var `is None` → true lacks an explicit test) — I verified the behavior is correct; the `Null` path is exercised via explicit `None` in both engines. Not elevating; optional follow-up.

Cargo.lock 0.9.7-rc.3→0.9.7 correction is mechanical (5 version-only lines), matches Cargo.toml, and unblocks the cargo-clippy pre-commit hook — acceptable to bundle. Two-commit shape (impl+tests / docs+CHANGELOG) followed. Docs + CHANGELOG updated.

johnrtipton:
# Retrospective — PR #1486

**Title:** fix(templates): support Django `is` / `is not` identity operators (#1483)
**Status:** MERGED
**Milestone:** v1.0.0 — unit 1 (the correctness gate)

## Quality: 5/5

A clean, well-scoped correctness fix. The Rust template renderer's
`evaluate_condition` had no `is` / `is not` branch, so `{% if x is None %}`
silently fell through to `Ok(false)` and always took the `{% else %}` branch.
The fix added a `values_identity` helper with correct Python identity semantics
plus a comparison-tier dispatch branch, bringing the Rust engine to parity with
Django's Python engine (which has supported these operators since Django 4.0).
Stage 11 APPROVE with 0 🔴 / 0 🟡 elevated, full suite green (4290 Python,
all Rust, 1572 JS). The two-commit shape (impl+tests / docs+CHANGELOG) was
followed correctly. Worthy first unit of the v1.0.0 correctness gate.

## What went well

- **Reproducer-first discipline (Stage 4) worked exactly as designed.** The
  `TestIsIdentityOperators` 10-case reproducer was written before the fix; 7
  of 10 cases failed pre-fix, confirming the bug artifact was real and the
  reproducer exercised the actual code path. This is the canonical
  symptom-up workflow from CLAUDE.md's bug-report triage section — no time
  burned chasing a path-down hypothesis.
- **Gate-off self-test (Action #254 / #1468) applied by the implementer.**
  Rather than leaving the tautology check to the Stage 11 reviewer, the
  implementer ran the gate-off self-test at Stage 5. The 7/10 pre-fix
  failures double as the gate-off evidence — the tests demonstrably fail
  when the change is absent. This is exactly the left-shift Action #254
  prescribes (epistemic of #1200, moved from reviewer to implementer).
- **Correct identity semantics, not naive equality.** `values_identity`
  treats `5 is 5` and `"a" is "a"` as false (CPython interning is not
  contractual) and only returns true for matching `None` / `True` / `False`
  singletons. This is the subtle-but-correct choice; a lazy implementation
  would have aliased `is` to `==` and shipped a latent bug.
- **Substring-containment ordering handled.** `" is not "` is matched before
  `" is "`, and both markers are space-padded so variable names containing
  the literal substring "is" (e.g. `is_active`, `this`) don't false-match.
  Stage 11 had a unit test specifically pinning this.
- **Docs parity caught.** Both `docs/RUST_TEMPLATE_API.md` and
  `docs/TEMPLATE_BACKEND.md` omitted `is` / `is not` from their
  condition-operator enumerations; the docs commit corrected both.

## Lessons learned — what could be improved

- **`Cargo.lock` staleness on `main` bounced the first commit.** The committed
  `Cargo.lock` carried `0.9.7-rc.3` workspace versions while `Cargo.toml` on
  `main` had already advanced to `0.9.7`. Any `cargo` invocation (including the
  `cargo-clippy` pre-commit hook) regenerates `Cargo.lock` to match
  `Cargo.toml`, leaving an unstaged modification that the hook then flags —
  so the first commit attempt bounced. The implementer absorbed the cost by
  folding the `0.9.7-rc.3` → `0.9.7` correction into commit 1, but this is a
  **repo-hygiene defect on `main`, not a PR defect**: every contributor who
  touches Rust code will hit the same friction until `main`'s `Cargo.lock` is
  re-synced with `Cargo.toml`. The version-bump commit that moved `Cargo.toml`
  to `0.9.7` (e871e190 "Bump version to 0.9.7") evidently did not re-stage the
  regenerated `Cargo.lock`. This should be a dedicated follow-up — a
  `chore:` PR that runs `cargo update -w` (or just `cargo build`) and commits
  the resulting `Cargo.lock`, plus a pre-commit/CI guard that fails fast when
  `Cargo.lock` drifts from `Cargo.toml`.
- **Process implication:** version-bump PRs touch `Cargo.toml` but the
  pipeline's release procedure does not currently verify `Cargo.lock` was
  regenerated and staged in the same commit. A "lockfile in sync" check
  belongs in the release skill's pre-tag checklist — the failure mode is
  silent on `main` and only surfaces as friction on the *next* unrelated
  Rust PR (here, #1486), which makes it hard to attribute.

## IDEAs

- IDEA: Add a pre-commit hook `cargo-lock-sync` that runs `cargo metadata
  --locked` (or `cargo update --workspace --locked --dry-run`) and fails if
  `Cargo.lock` would change — catches `Cargo.toml`/`Cargo.lock` drift at
  commit time on the PR that introduces it, instead of on an unrelated PR
  later.
- IDEA: Add a "regenerate and stage `Cargo.lock`" step to the `djust-release`
  skill's version-bump procedure, so `Cargo.lock` and `Cargo.toml` always
  land in the same commit.
- IDEA: File a `chore(deps): re-sync Cargo.lock with Cargo.toml on main`
  follow-up issue to clear the current `0.9.7-rc.3` → `0.9.7` staleness for
  all contributors, independent of the next Rust PR.
- IDEA: Generalize the implementer-side gate-off self-test (#254) into a
  Stage 5 checklist line for all engine-parity fixes, since "pre-fix
  reproducer failure count" is itself the gate-off evidence and costs
  nothing extra to record — #1486 got it for free because the reproducer
  predated the fix.

PR #1488, opened 2026-05-18
feat(v1.0.0): API-stability + deprecation policy (milestone unit 2)

johnrtipton:
## Summary

**v1.0.0 milestone unit 2.** A 1.0 release is a SemVer stability commitment — this PR defines that commitment.

## What's in it

**Policy (`docs/API_STABILITY.md`)** — the canonical 1.0 API-stability + deprecation policy: what SemVer covers (top-level `djust` exports, documented decorators, public `LiveView`/`LiveComponent` methods, public template tags), what is explicitly NOT covered (underscore-prefixed internals, `djust.mixins.*` internals), the deprecation process, and the support window (deprecated-in-`1.Y` → removed no earlier than `2.0.0`; the 3 pre-1.0 legacy symbols → `>= 1.1.0`). Plus `docs/website/guides/api-stability.md` (user guide, nav-linked).

**`warn_deprecated` helper (`djust._deprecation`)** — internal, underscore-prefixed (does not expand the public surface). The executable form of the policy: every deprecation now names a concrete removal version + a replacement.

**3 deprecation call sites migrated + fixed:**
- `LiveViewForm` said *"removed in djust 0.4"* — a version long past. Corrected to `>= 1.1.0`.
- `@event` / `_legacy` theming had vague/missing removal versions — now concrete.
- `stacklevel` corrected on all sites (empirically verified) so `DeprecationWarning` points at the **caller's** frame, not djust internals (`@event` 2→3, `LiveViewForm` 3→5 — Django metaclass frames).

## Public-API audit result

**0 symbols flagged experimental/provisional** — the public surface (58 top-level exports, 17 decorators, 21 mixins, public methods, template tags) is clean for a 1.0 freeze. Findings F3 (4 decorators absent from top-level `__all__`) and F4 (`djust.mixins.*` is internal) are documented in the policy.

## Tests

22 tests in `python/djust/tests/test_deprecation.py` (helper + migration + 4 per-site stacklevel regression tests). Full suite green. Stage 7 self-review caught 2 real `stacklevel` bugs; fixed + regression-tested before Stage 8.

## Pipeline

Feature pipeline, two-commit shape (code, then docs+CHANGELOG). State: `.pipeline-state/feat-v1-api-stability-policy.json`.


johnrtipton (commented):
APPROVE — v1.0.0 unit 2 is well-executed. Reviewed against the PR checklist; all gates pass.

**Verified:**
- Tests: 2803 + 57 + 1818 (theming) pass; 0 failures, no flakes.
- **Stacklevel correctness (the error-prone part):** empirically triggered all 6 legacy entry points (`@event`, `LiveViewForm`, `THEMES[]`, `THEMES.get`, `get_theme`, `list_themes`) from synthetic user frames — every `DeprecationWarning.filename` resolves to the *user* frame, not djust/Django internals. The Stage-7 fixes (`@event` 2→3, `LiveViewForm` 3→5, `_legacy` →4) are correct. The 4 per-site `*_stacklevel_points_at_user_frame` tests genuinely assert `basename == test_deprecation.py` — real wiring assertions, fail on wrong stacklevel, not tautologies.
- `warn_deprecated` (`_deprecation.py:26`): keyword-only since/removed_in/instead, correct `DeprecationWarning` category, `instead`-omitted branch drops the migration clause cleanly. Underscore-prefixed → not a new public symbol.
- Migrated sites still satisfy substring `pytest.warns(match=)` — confirmed in theming + decorator suites.
- `API_STABILITY.md`: public-symbol count **58** matches `__init__.__all__` (plan said 53; docs agent recount to 58 is correct). `decorators.__all__` = 17 ✓. 3 deprecated symbols → all `>= 1.1.0` ✓. F3/F4 claims accurate (4 decorators absent from top-level `__all__`, confirmed).
- Nav linked in BOTH `_config.yaml:284` and `index.md:97` ✓. No `print()`/bare-except in the diff.

**One concern (non-blocking):** `API_STABILITY.md:321` claims "A follow-up issue tracks [F3] consideration" — no such issue exists (`gh issue list` empty). Per Action #1071, file the F3 follow-up issue or soften the claim to "should be considered for 1.1".

Edge cases checked: `instead=None` explicit vs omitted, keyword-only enforcement, double-warning bypass in `get_theme`/`list_themes`, metaclass-chain frames for `LiveViewForm`.

APPROVE

johnrtipton:
# Retrospective — PR #1488

**Title:** feat(v1.0.0): API-stability + deprecation policy (milestone unit 2)
**Status:** MERGED
**Milestone:** v1.0.0 — unit 2
**Pipeline:** Full 14-stage feature pipeline

## Quality rating: 5/5

A clean, well-scoped milestone unit. The deliverables match the plan exactly: a
canonical `docs/API_STABILITY.md` policy, an internal `djust._deprecation.warn_deprecated`
helper, and 3 migrated deprecation call sites. The public-API audit (0 symbols
flagged experimental) gives the 1.0 freeze a credible foundation. Two real
`stacklevel` bugs were caught and fixed *before merge* by the pipeline's own
checkpoints, and a prose-doc accuracy gap was closed with a filed follow-up
issue. No 🔴/🟡 escaped to a post-merge fix.

## What went well

- **Two-commit shape, applied cleanly.** Code (`warn_deprecated` + 3 migrations
  + tests) in commit 1, docs+CHANGELOG in commit 2. Same discipline as PR #1486
  (v1.0.0 unit 1) — both v1.0.0 units have now used the two-commit boundary
  without a single `[Unreleased]` cross-edit collision. The pattern is holding.
- **Gate-off self-test (Action #254) ran.** As with unit 1, the new tests were
  verified against a gated-off change so tautological tests couldn't slip
  through. The 4 per-site stacklevel regression tests are behavior-meaningful —
  they fail if the warning points at the wrong frame.
- **Real bug fix folded into a policy PR.** `LiveViewForm` carried a stale
  *"removed in djust 0.4"* message — a version years in the past. The migration
  corrected it to `>= 1.1.0`, and the `@event` / `_legacy` theming sites got
  concrete removal versions where they previously had none. The policy is not
  just prose; it is executable and the existing call sites now conform to it.
- **Internal-by-default helper.** `warn_deprecated` is underscore-prefixed
  (`djust._deprecation`), so a PR *about* API surface stability did not itself
  expand the public API surface. Good self-consistency.
- **Public-API audit done before the freeze, not after.** 58 top-level exports,
  17 decorators, 21 mixins, public methods and template tags all reviewed; 0
  experimental/provisional. F3/F4 documented in the policy rather than silently
  dropped.

## Lessons learned

- **Stage 7 self-review caught 2 real `stacklevel` bugs the 14 implementer
  tests missed.** The implementer's first-pass tests asserted message text and
  warning category — *what* the warning says — but never *where* it points. A
  `DeprecationWarning` is only useful if it points at the user's frame, not at
  djust internals; with the wrong `stacklevel` the user sees a djust file in
  their warning output and cannot find their own offending call. The fix added
  4 per-site stacklevel regression tests, each empirically verified via a probe
  sweep (`@event` 2→3 for an extra helper frame; `LiveViewForm` 3→5 for Django
  metaclass `__new__` frames). **Canonical lesson:** deprecation-helper
  migrations need a *stacklevel-points-at-user-frame* test per call site,
  empirically verified — never just a message-text assertion. Frame depth is
  call-site-specific (helper wrappers, metaclasses, descriptors all add frames),
  so a single "the helper works" test does not generalize across sites.
- **Stage 11 caught a prose-doc accuracy gap (Action #1071).** `API_STABILITY.md`
  §F3 referenced "a follow-up issue" that did not exist at write time — exactly
  the failure mode Action #1071 warns against (prose docs must not name
  non-existent artifacts). Fixed by filing #1489 and citing it in commit 3.
  The lesson: when a policy/canon doc says "tracked in a follow-up", file the
  issue *first*, then cite the real number — don't write the forward reference
  and hope to backfill it.
- **What could be improved — left-shift the stacklevel check to Stage 5.** The
  2 stacklevel bugs were caught at Stage 7, one stage after implementation. The
  implementer subagent had no explicit instruction to verify frame-pointing,
  so it tested the observable string and stopped. A Stage 5 implementer-prompt
  requirement — "for every `warn_deprecated` call site, add a test that asserts
  the warning's `filename` is the *caller's* file, probe-verified" — would have
  surfaced these one stage earlier and saved a Stage 7→Stage 5 round-trip. This
  mirrors the Action #254 left-shift (move the gate-off self-test from reviewer
  to implementer); the same epistemic applies here.
- **What could be improved — the F3 follow-up issue should have been filed at
  Stage 9 (docs), not patched in at Stage 11.** The docs author wrote the
  forward reference; the docs author should have filed #1489 in the same stage.
  Stage 11 catching it worked, but it cost a third commit on a PR that was
  otherwise a clean two-commit shape. A Stage 9 docs checklist item — "every
  'see follow-up' / 'tracked elsewhere' phrase in a canon doc must cite a real,
  already-filed issue number" — closes this.

## Ideas

- `IDEA:` Add a Stage 5 implementer-prompt requirement for deprecation/warning
  migrations: every `warn_deprecated` (or `warnings.warn`) call site gets a
  test asserting the emitted warning's `filename` resolves to the *caller's*
  module, probe-verified. Left-shifts the stacklevel-correctness check from
  Stage 7 to Stage 5.
- `IDEA:` Add a `djust_audit` / lint check that scans for `warnings.warn(...,
  DeprecationWarning)` calls *not* routed through `djust._deprecation.warn_deprecated`
  — keeps future deprecations on the policy-compliant helper instead of
  re-introducing ad-hoc messages (the exact drift this PR cleaned up).
- `IDEA:` Add a `make` target or test that asserts every `removed_in=` version
  string passed to `warn_deprecated` is `>=` the current `djust.__version__` —
  catches another "removed in 0.4" stale-version regression mechanically.
- `IDEA:` Stage 9 docs checklist: grep canon docs for forward-reference phrases
  ("follow-up issue", "tracked separately", "see below") and require each to
  resolve to a real, already-filed artifact before the docs commit.
- `IDEA:` Now that v1.0.0 units 1 and 2 have both used the two-commit shape +
  gate-off self-test cleanly, treat that pairing as the default expectation for
  the remaining v1.0.0 units rather than a per-PR decision.

PR #1494, opened 2026-05-18
docs(v1.0.0): 1.0 documentation pass (milestone unit 6)

johnrtipton:
## Summary

**v1.0.0 milestone unit 6 — the final unit.** Gets djust's docs release-ready for the 1.0 stability commitment.

## What's in it

**`README.md`:**
- **P0 fix** — the Getting Started walkthrough used `CounterView.as_live_view()`, which doesn't exist (`live_view.py` defines `as_view()`). Every new user copy-pasting the README hit an `AttributeError`. Fixed.
- "Django 3.2+" → "Django 4.2+" (matches `pyproject.toml`).
- Dropped a `print(f"...")` anti-pattern from the custom-component example (djust's own rules forbid it).
- Corrected stale client-JS size claims (~29 KB → ~53 KB gzipped minified — measured).
- Marked the shipped "File upload handling" roadmap item done.

**New `docs/website/guides/upgrade-to-1.0.md`** — concise 0.9.x → 1.0 upgrade guide: the SemVer-1.0 stability commitment (cross-refs `API_STABILITY.md`), the 3 deprecated symbols (`@event`, `LiveViewForm`, `_legacy` theming), the new `Y` accessibility checks, and that 1.0 is a **consolidation release — no breaking changes from 0.9.7**. Nav-linked from `_config.yaml` + the guides index.

**`docs/README.md`** — stale check-count reference fixed (the `Y` category was added).

**`CHANGELOG.md`** — a narrative lead-in framing `[Unreleased]` as the v1.0.0 stability milestone (units 1-6). No dated `[1.0.0]` heading — that's the release cut's job.

## Milestone status

With this PR, **all 6 units of the v1.0.0 milestone are complete** (#1486, #1488, #1490, #1491, #1492, this). v1.0.0rc1 can be cut next.

## Pipeline

Docs-only pipeline — test/review/security stages skipped per `DOCS_ONLY`. State: `.pipeline-state/docs-v1-documentation-pass.json`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1506, opened 2026-05-18
feat(adr): ADR status/version-line audit + reconcile stale Target-version lines (#1493, #1501)

johnrtipton:
## Summary

Batch PR for the v1.0.0rc2 milestone — two related v1.0.0-retro follow-ups about ADR metadata accuracy.

| Issue | What | Files |
|---|---|---|
| #1493 | Corrected stale `**Target version**:` lines in 12 ADRs to match the `**Status**:` lines PR #1492 reconciled | `docs/adr/002-008, 013-017` |
| #1501 (a) | New `scripts/check-adr-status.py` audit + pre-commit/CI/Makefile wiring | `scripts/`, `tests/`, `.pre-commit-config.yaml`, `.github/workflows/test.yml`, `Makefile` |

## #1493 — one-time cleanup

PR #1492 reconciled the `**Status**:` line of 10 ADRs but left the `**Target version**:` lines stale (e.g. ADR-008 read "v0.7.0 candidate" though it shipped in v0.5.1). Accepted ADRs now carry `**Shipped in**: vX.Y.Z` (a fact); deferred ADRs (003-006) are marked `post-1.0 (deferred)`. ADRs 001/009/010/011 (no version line) and 012 (`**Milestone**:`, already factual) are untouched.

## #1501 (a) — close the drift loop

`scripts/check-adr-status.py` enforces the invariant the #1493 cleanup establishes:
- **Rule 1 (hard, exit 1)** — an `Accepted` ADR's version line, if present, must be `**Shipped in**:` not `**Target version**:`. Deferred ADRs must mark the line `post-1.0`/`deferred`.
- **Rule 2 (soft warning)** — a `Proposed` ADR with shipped-markers in its body.

Wired into `.pre-commit-config.yaml` (runs on `docs/adr/` edits), `.github/workflows/test.yml`, and `make check-adr-status`. Modelled on `scripts/check-handler-contracts.py`.

## Out of scope — #1501 part (b)

The ADR-status-flip *prompt* in the `djust-release` and `pipeline-run` skills is **out-of-repo** — `.claude/` is gitignored repo-wide and `~/.claude/skills/` is the user's private skill dir. Tracked as a follow-up issue, not bundled here (Action #1079 scope discipline).

## Tests

[...]

PR #1504, opened 2026-05-18
fix(deploy): wire TARBALL_EXCLUDES into _create_tarball; document empty except (#1495)

johnrtipton:
## Summary

Resolves the 2 low-severity CodeQL **note** alerts deferred from the v1.0.0 unit-3 security sweep (PR #1490) per Action #1079. Both are code-quality items, not vulnerabilities.

| CodeQL | Rule | Fix |
|---|---|---|
| #2330 | `py/unused-global-variable` | `TARBALL_EXCLUDES` in `deploy_cli.py` was orphaned; `_create_tarball` hardcoded two drifted inline lists. Wired the constant in as the single source of truth. |
| #2334 | `py/empty-except` | Bare `except: pass` in `checks.py` (`check_psycopg3_for_pg_notify`) replaced with an explicit `psycopg2_version = ""` fallback + explanatory comment. |

## Behavior change (intended — see CHANGELOG)

Wiring `TARBALL_EXCLUDES` in required normalizing its glob-prefixed entries (`*.pyc`, `*.pyo`, `*.egg-info`, `*.log`) to substring form — `_create_tarball` matches with `in`, not glob. Net effect on deploy tarballs:

- **Newly excluded**: `.hg`, `.svn`, `logs/`, `media/`, `staticfiles/` directories.
- **`.log` files now actually excluded** — the old inline `*.log` entry never matched (literal `*` under substring containment).
- Nothing previously excluded becomes included; `.env` and `db.sqlite3` remain excluded.

## Tests

`TestCreateTarball` (3 tests) added to `python/tests/test_deploy_cli.py`:
- `test_create_tarball_uses_tarball_excludes_constant` — monkeypatch proves the constant is consulted.
- `test_create_tarball_excludes_canonical_set` — locks in the behavior-change entries.
- `test_tarball_excludes_constant_is_substring_safe` — pins the no-`*` normalization.

Gate-off self-test (Action #254): reverting the wiring fails 2/3 tests. Full suite: 7199 passed, 0 failed.

## Issue × file × test

| Issue | Files | Tests |
|---|---|---|
| CodeQL #2330 | `python/djust/deploy_cli.py` | `TestCreateTarball` (×3) |
| CodeQL #2334 | `python/djust/checks.py` | comment-only — no test |

Closes #1495.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #1495, opened 2026-05-18
tech-debt: 2 low-severity CodeQL note alerts (deferred from v1.0.0 security sweep #1490)


johnrtipton:
Deferred from the v1.0.0 unit-3 security sweep (PR #1490) per Action #1079 — code-quality items, not vulnerabilities.

- **CodeQL #2330** `py/unused-global-variable` — `TARBALL_EXCLUDES` in `deploy_cli.py:46` is orphaned; `_create_tarball` hardcodes two drifted inline lists. Wire `TARBALL_EXCLUDES` into `_create_tarball` + add a regression test.
- **CodeQL #2334** `py/empty-except` — `checks.py:3057`, cosmetic; add an explanatory comment.

Both are `note` severity. Out of scope for the security sweep (which targeted dependency vulnerabilities).

PR #1508, opened 2026-05-18
feat(docs): doc-snippet smoke test + claim assertions; reconcile README roadmap (#1500, #1497)

### LINKED ISSUE #1497: docs: README/roadmap doc-rot reconciliation (deferred from v1.0.0 docs pass #1494)
Deferred from v1.0.0 milestone unit 6 (PR #1494, 1.0 documentation pass) per Action #1079 — the docs pass fixed the P0 (`as_live_view()`) + critical stale claims + added the upgrade guide. Remaining doc-rot:

- Full `README.md` Roadmap-section checklist reconciliation (several items beyond `File upload handling` may be stale/shipped).
- `docs/roadmap.md` currency sweep (distinct from the audited `ROADMAP.md`).

Lower-priority doc-rot; not release-blocking.

### LINKED ISSUE #1500: tech-debt: doc-snippet smoke test + mechanically-derivable doc-claim assertions
Source: Retro v1.0.0 (PR #1494 finding, reinforced by #1486 + #1488).

PR #1494 caught a **P0 README bug** — the Getting Started walkthrough called `CounterView.as_live_view()`, a method that never existed — plus stale claims (Django 3.2+ vs 4.2+, ~29 KB vs measured ~53 KB JS size, a `print(f"...")` anti-pattern in an example). Docs ship unverified; code claims rot silently between releases.

**Fix:** (1) a doc-snippet smoke test that extracts fenced Python blocks from `README.md` / `QUICKSTART.md` and at minimum import/AST-checks them (ideally executes the canonical walkthrough against the demo project); (2) make mechanically-derivable claims self-checking — assert the README's minimum-Django-version against `pyproject.toml` and the JS-size claim against the measured bundle; (3) lint doc code examples against djust's own security/style rules (the `print(f"...")` would have been flagged).

johnrtipton:
## Summary

Batch PR for the v1.0.0rc2 milestone — a doc-rot smoke test (#1500) plus the one-time README reconciliation (#1497) it surfaces.

## #1500 — doc-snippet smoke test + claim assertions (parts a + b)

PR #1494's docs pass caught a P0 README bug (`as_live_view()` — a method that never existed) plus stale claims. Docs ship unverified; this adds a standing guard.

`scripts/check-doc-snippets.py`:
- **Part (a)** — extracts fenced Python blocks from `README.md`/`QUICKSTART.md`, AST-parses each (catches malformed snippets), classifies fragment-vs-module, and import-resolves the imports of complete modules (catches phantom/renamed public symbols).
- **Part (b)** — asserts mechanically-derivable claims: the README's stated minimum Django version against `pyproject.toml`, and the JS bundle-size claim against the measured `client.min.js.gz` (±3 KB tolerance band; graceful skip if the bundle isn't built).

Wired into `.pre-commit-config.yaml`, `.github/workflows/test.yml`, and `make check-doc-snippets`. Modelled on `scripts/check-adr-status.py`.

**Part (c)** — doc-example security/style linting — is deferred to a follow-up (ruff doesn't flag `print(f"...")` so it needs a custom AST walker + anti-pattern allowlist; its own design surface).

## #1497 — README/roadmap reconciliation

- Flipped 2 stale roadmap checkboxes — **Redis-backed session storage** and **horizontal scaling** — both shipped via `RedisStateBackend` (`python/djust/state_backends/redis.py`).
- Fixed a broken `register_component` example snippet: it imported the symbol from the wrong package and used a `Component` base class that `register_component` rejects (`register_component` requires `LiveComponent`). Corrected to `LiveComponent` and **execution-verified** (registers + renders + event handler fires).
- `docs/roadmap.md` — surveyed, no mechanical rot (evergreen strategy doc), left unchanged.

## Tests

`tests/test_check_doc_snippets.py` — 14 tests (subprocess + tempfile fixtures, tautology-guarded, `@pytest.mark.slow` dogfood test against the real docs). Gate-off self-test (Action #254): injecting a phantom symbol flips `test_real_docs_pass`. Full suite: 7224 passed, 0 failed.

Closes #1500 (parts a+b; part c → follow-up). Closes #1497.


johnrtipton (commented):
APPROVE — verified independently.

**register_component snippet (CRITICAL):** Extracted the corrected snippet verbatim from README.md, ran it under `DJANGO_SETTINGS_MODULE=tests.settings` + `django.setup()`. Output: `SNIPPET OK: register_component succeeded` / `INSTANTIATE OK: label=Hello clicks=0` / `get_context_data OK` — no exception. `register_component` is correctly imported from `djust.components` (not `djust`), enforces `issubclass(..., LiveComponent)`, and `trigger_update` is a real method. The prior fix-pass correction holds.

**Empirical canary (#252/#1459):** Built a synthetic broken README with a ```python block `from djust import totally_fake_symbol`. Ran `check-doc-snippets.py` against it → exit 1, `README.md:6 — djust has no attribute totally_fake_symbol`. The lint catches the bug class it claims to.

**Tautology gate-off (#1200/#254):** Gated `check_django_floor`s `if stated != floor` to always-false → script returns exit 0 on a 3.2-vs-4.2 mismatch, proving `test_django_version_mismatch_fails` (`assert code == 1`) is non-tautological.

**Tests:** `pytest tests/test_check_doc_snippets.py -q` → 14 passed. `-m slow` dogfood (`test_real_docs_pass`) → 1 passed (real docs exit 0).

**Quality:**
- Two-commit shape correct: e07a8ac5 impl (no CHANGELOG), eca60930 CHANGELOG-only.
- Fragment-vs-module heuristic (has-import AND has-class/def) is conservative — name resolution runs only on modules, so no false-positives. The false-negative (phantom import in a class-less snippet skipped) is documented honest scope.
- Size regex `~NN KB` + gzip-line filter matches all 3 README claims (lines 23/710/953); band 48.6-54.6 KB passes the `~53` rounded value.
- pre-commit `bash -c .venv/bin/python ...` + `language: system` matches existing hooks (lines 158/198/212/224). CI step omits PYTHONPATH but works from repo root (verified — cwd on sys.path resolves tests.settings).
- #1497: 2 checkbox flips verified against `RedisStateBackend`.

No security, correctness, or test-quality issues. Deferred part (c) and the phantom-method-call limit are documented, correct scope.

Verdict: APPROVE.

johnrtipton:
## Retrospective — PR #1508 (v1.0.0rc2 drain, task 3 of 3)

**Batch**: #1500 (parts a+b — `scripts/check-doc-snippets.py` doc-snippet smoke test + Django-floor / JS-size claim assertions) + #1497 (README roadmap reconciliation — 2 stale checkboxes + a broken `register_component` example snippet). #1500 part (c) deferred to a follow-up.

**Quality rating: 5/5** — Merged clean, full suite 7224 passed, CI green, 0🔴/0🟡 at Stage 11. The pipeline did exactly what a multi-stage pipeline is for: a tool caught a bug, a human-equivalent review caught what the tool by design *cannot*, and a fix-pass closed it with execution proof.

### What went well

- **The new smoke test earned its keep on its first run.** The Stage 5 implementer ran `check-doc-snippets.py` against the real docs and it immediately surfaced a genuine stale snippet — `register_component` imported from the wrong package. The guard PR #1500 builds caught a live bug while still being authored.
- **Stage 7 caught the limit of the tool.** The implementer's fix (correct import path + decorator→call form) still 500d on copy-paste: the snippet used `class Button(Component)` while `register_component` requires `LiveComponent` (disjoint hierarchies). The smoke test — AST + import-resolution, *no execution* by design — could not catch a base-class `TypeError`. Stage 7 Code Review did. The plan had documented that honest limit; the review did the job the tool can't.
- **Fix-pass execution-verified the correction.** The fix-pass agent rewrote the snippet to `LiveComponent` and *ran it* — registers + renders + event handler fires, no exception. Stage 11 independently re-executed it. Two layers of execution proof, not assertion.
- **Convention reuse paid off.** `check-doc-snippets.py` mirrors the `check-adr-status.py` shape established by task 2 (PR #1506). Less design surface, faster review, consistent wiring (pre-commit + `test.yml` + `make` target).
- **Stages 6 / 8 / 11 otherwise clean.** Gate-off self-test (Action #254) confirmed non-tautological (injecting a phantom symbol flips `test_real_docs_pass`).

### Lessons learned

**Headline — execute, don't assume: a doc-snippet fix must be execution-verified, not just import-path-verified.** The implementer changed the import path and *believed* the snippet was fixed. It wasn't — running it would have surfaced the base-class `TypeError` in seconds. The fix was "plausible on inspection" but wrong, exactly the failure mode the #1500 smoke test exists to reduce. The smoke test by design can't execute, so the only verification that closes the loop is *running the snippet*.

**This is now a 3-occurrence pattern across the v1.0.0rc2 drain — flag it hard for the milestone retro.** Every task in this milestone lost time to an unverified assumption that a cheap check would have caught:

- **Task 1 (PR #1504, #1495)** — a Stage-7 reviewer *assumed* CHANGELOG should be present (false positive from missing two-commit-shape context). Cheap check skipped: confirm the pipeline's commit-shape convention.
- **Task 2 (PR #1506, #1493/#1501)** — the planner *assumed* a `.claude/` SKILL.md file was tracked in-repo (it was gitignored). Cheap check skipped: `git check-ignore`.
- **Task 3 (PR #1508, this PR)** — the implementer *assumed* an import-path fix was complete without executing the result. Cheap check skipped: run the snippet.

Same shape three times: an *unverified assumption that a one-command check would have caught*. The fix is identical every time — **run the check** (`git check-ignore` / execute the snippet / state and verify the premise) before acting on the belief. The milestone retro should canonicalize this as a single cross-stage rule rather than three task-local notes.

**A tool's documented limit is a contract — staff the gap.** The plan explicitly stated the smoke test does AST + import-resolution but *not* execution. That honesty is correct, but a documented limit only helps if a later stage actually covers it. Here Stage 7 did. Pipelines should treat "the tool can't catch X" as an explicit hand-off to a human-equivalent stage, not a shrug.

### Ideas

IDEA: Stage 5 rule — any doc-snippet edit (README/QUICKSTART/docs fenced code) must be *execution-verified*: run the snippet (or a minimal harness around it) and confirm no exception before reporting the fix done. Import-path / AST checks are necessary but not sufficient. Add to the implementer-subagent prompt + `docs/PULL_REQUEST_CHECKLIST.md`.

IDEA: Canonicalize the cross-task "execute/check, don't assume" thread as one CLAUDE.md rule for the v1.0.0rc2 milestone retro — "before acting on a premise about repo state or a fix's completeness, run the one-command check that confirms it (`git check-ignore`, execute the snippet, verify the convention)." 3-occurrence empirical backing (#1504, #1506, #1508).

IDEA: When a tool ships with a documented blind spot, the plan should name *which stage* covers the gap (here: "execution coverage → Stage 7/11"). Makes the hand-off explicit instead of implicit.

IDEA (process): #1500 part (c) — doc-example security/style lint (custom AST walker for `print(f"...")` etc.) — needs a follow-up issue filed; it has its own design surface and shouldn't be smuggled into a drain task.

PR #1518, opened 2026-05-18
fix(checks): anchor HTML-attribute regexes against data-* false-match (#1514, #1517)

johnrtipton:
## Summary

First task of the v1.0.0rc3 milestone — closes the `\b`/`data-*` regex false-match class in `python/djust/checks.py`.

## #1514 — 4 HTML-attribute regexes re-anchored

PR #1512 fixed the Y003/Y004 regexes (`_INPUT_TYPE_RE`, `_POSITIVE_TABINDEX_RE`), which used a bare `\b` word-boundary anchor that false-matched `data-*` attributes — `\b` sits between `data-` and the attribute name because `-` is a non-word character. Stage-4 artifact verification found **three more** HTML-attribute regexes in `checks.py` with the identical bug:

| Regex | Check | Bug |
|---|---|---|
| `_IMG_HAS_ALT_RE` | Y002 | `` with no real `alt` silently treated as labelled (false negative) |
| `_HREF_ATTR_RE` | Y001 | `<a data-href=...>` treated as a real link (false positive) |
| `_ACCESSIBLE_NAME_ATTR_RE` | Y001 | `data-aria-label` wrongly silences a Y001 finding |
| `_CONTROL_ID_RE` | Y003 | `data-id` mis-paired with a `<label for>` (false negative) |

All 4 re-anchored `\b` → `(?<![\w-])` (the lookbehind also rejects a preceding hyphen). Fixing all 4 — same file, same bug class — is required for the #1517 meta-check to pass against a clean `checks.py`. The 4 `_LIVE_RENDER_*` regexes are **not** changed: they scan `{% live_render %}` template-tag kwargs, where no `data-` prefix convention exists.

## #1517 — meta-check guards against recurrence

New `TestChecksRegexHardening` test class introspects every compiled `re.Pattern` in `checks.py` and fails on any bare-`\b`-anchored attribute regex. The 4 `_LIVE_RENDER_*` kwarg regexes are allowlisted (with a stale-entry guard test). This is the 3rd occurrence of the bug class — the meta-check makes a 4th impossible.

## Tests

8 new tests in `python/djust/tests/test_accessibility_checks.py` (5 regression + 3 meta-check, incl. an Action-#1046 doc-claim assertion that the detection regex matches the pre-fix `\balt\s*=` shape). Gate-off self-test (Action #254): reverting one regex fails 3 tests. Full suite: 7264 passed, 0 failed; no Y001/Y002/Y003 regression.

Closes #1514. Closes #1517.

[...]

Issue #1514, opened 2026-05-18
tech-debt: _IMG_HAS_ALT_RE (Y002) false-matches data-alt — same \b weakness fixed for Y003/Y004 in PR #1512


johnrtipton:
Pre-existing weakness surfaced during PR #1512 (#1496 Y003/Y004) Stage 7 review — kept out of scope per Action #1079.

## Issue

`_IMG_HAS_ALT_RE` in `python/djust/checks.py` (the Y002 missing-alt check, from PR #1491) is `re.compile(r"""\balt\s*=""")`. The `\b` word boundary matches inside `data-alt='...'` because `-` is a non-word char — so an `<img data-alt='x'>` with no real `alt` attribute is wrongly treated as having an alt. This is a **false negative**: Y002 silently misses a genuinely alt-less `<img>` that happens to also carry a `data-alt`.

## Fix

PR #1512 fixed the identical weakness in the *new* Y003/Y004 regexes (`_INPUT_TYPE_RE`, `_POSITIVE_TABINDEX_RE`) by replacing the leading `\b` with a `(?<![\w-])` negative lookbehind. Apply the same anchor to `_IMG_HAS_ALT_RE`: `\balt` → `(?<![\w-])alt`. Add a regression test: `<img src='x.png' data-alt='decorative'>` (no real `alt`) must be flagged by Y002.

Low severity — false negative (missed defect), not a false positive; `DjustWarning` never fails `manage.py check`. Worth fixing for Y-category consistency now that Y003/Y004 use the correct anchor.

PR #1519, opened 2026-05-18
fix(deploy): anchor _create_tarball exclude matching to path segments (#1505)

### LINKED ISSUE #1505: tech-debt: _create_tarball exclude matching uses substring containment — over-excludes legitimate paths
Deferred follow-up from PR #1504 (#1495 CodeQL note-alert fixes), per Action #1079 scope discipline.

## Context

PR #1504 wired the `TARBALL_EXCLUDES` constant into `_create_tarball` (`python/djust/deploy_cli.py`) as the single source of truth. Both the dir filter and the file filter match exclude patterns via Python substring containment (`pattern in d` / `pattern in str(relative)`), not glob/`fnmatch`.

## The issue

Substring matching over-excludes legitimate paths. Stage 7 + Stage 8 review of PR #1504 flagged examples:

- `venv` matches a file `venvironment.py`
- `dist` matches `distance.py`
- `media` matches `media_helper.py` / `media_player/`
- `build` matches `mybuild_config.py`
- `logs` / `staticfiles` similar

This pre-existed PR #1504 (the old hardcoded inline lists already contained `venv`/`dist`/`media`/`build`), so PR #1504 did not introduce or worsen it — but it is now worth fixing at the single source of truth.

## Proposed fix

Replace substring containment with one of:
- **Basename / path-segment anchored matching** — match exclude tokens against exact directory basenames and exact path segments, not arbitrary substrings.
- **`fnmatch`-based glob matching** — restore the glob semantics the `*.pyc`-style entries originally implied (PR #1504 had to strip the `*` prefixes precisely because the function does substring matching).

Either approach removes the false-positive class. Add regression tests for the over-match cases above.

Severity: low (deploy tooling, operator-controlled source dir; over-exclusion drops files from a deploy tarball rather than leaking them). Not release-blocking.

johnrtipton:
## Summary

Second task of the v1.0.0rc3 milestone. `_create_tarball` (`python/djust/deploy_cli.py`) matched `TARBALL_EXCLUDES` entries via **substring containment**, so legitimately-named files were dropped from deploy tarballs:

| Excluded by mistake | because the token… |
|---|---|
| `venvironment.py` | contains `venv` |
| `distance.py` | contains `dist` |
| `media_helper.py`, `media_player/` | contains `media` |
| `mybuild_config.py` | contains `build` |

Empirically: the old code shipped only `app.py` from a tree of 8 legitimately-named files.

## The fix

`TARBALL_EXCLUDES` is split into **5 typed groups** and `_create_tarball`'s dir/file filters use anchored matching instead of substring containment:

- `EXCLUDE_DIR_NAMES` (17) — exact directory basename
- `EXCLUDE_DIR_SUFFIXES` (`.egg-info`) — directory basename `endswith` (`mypkg.egg-info/`)
- `EXCLUDE_FILE_SUFFIXES` (`.pyc`, `.pyo`, `.log`) — filename `endswith`
- `EXCLUDE_FILENAMES` (`.DS_Store`, `Thumbs.db`) — exact filename
- `EXCLUDE_FILENAME_STEMS` (`.env`, `db.sqlite3`) — `file == T or file.startswith(T+".") or file.startswith(T+"-")`

## Sensitive-file handling (Stage-8 review catch)

A Stage-8 security review caught that naive exact-filename matching would have **regressed** sensitive-file exclusion: the old substring match excluded any path containing `.env`/`db.sqlite3`, so `.env.production`, `.env.local`, `config/.env.production`, `db.sqlite3-wal` were excluded — exact-match would have shipped them into a deploy tarball (credential / live-DB leak). `EXCLUDE_FILENAME_STEMS` with the `startswith(T+".")`/`startswith(T+"-")` rule keeps all of those excluded, while the lookalike `.environment` is correctly **not** excluded (the #1505 over-match stays fixed).

## Tests

`TestCreateTarball` in `python/tests/test_deploy_cli.py` — 60 tests in the file, including over-match regression, genuine-exclusion coverage, the sensitive-stem-variant cases, and the `.environment` lookalike. Gate-off self-tests (Action #254) confirmed non-tautological for both the over-match fix and the sensitive-stem fix. Full suite: 7268 passed, 0 failed.

Closes #1505.


johnrtipton (commented):
APPROVE

Stale-base: BEHIND=0. Tests: 60/60 pass in test_deploy_cli.py. Two-commit shape correct (18f031b0 impl+tests, 75f5a37e CHANGELOG-only).

**Correctness — verified empirically**
TARBALL_EXCLUDES cleanly split into 5 typed groups; total = 17+1+3+2+2 = 25, all original entries classified, none dropped (test_exclude_groups_have_expected_kinds pins this).

Built the new file/dir predicates and ran them:
- Sensitive stems excluded: `.env`, `.env.production`, `.env.local`, `db.sqlite3`, `db.sqlite3-wal`, `db.sqlite3.bak` — all True.
- Lookalikes NOT excluded: `.environment`, `environment_helper.py`, `venvironment.py`, `distance.py`, `media_helper.py`, `mybuild_config.py`, `staticfiles_conf.py`, `logsink.py` — all False.
- Genuine exclusions hold: `.git/`, `__pycache__/`, `node_modules/`, `mypkg.egg-info/` (`.egg-info` suffix), `dist/`, `build/`, `.pyc/.pyo/.log`, `.DS_Store`, `Thumbs.db`.
- Dir-pruning: filter mutates `dirs[:]` in place, so an excluded dir`s subtree is pruned by os.walk.

The Stage-8 stem fix (`file == T or startswith(T+".") or startswith(T+"-")`) is correct; the `.`/`-` discriminator is the right call.

**Tautology gate-off (Action #1200)**
Reverted both filters to substring matching; re-ran TestCreateTarball: 3 over-exclusion tests FAILED (lookalike_names, env_lookalikes, file_named_like_excluded_dir). Tests are real, not tautological.

**Note (non-blocking)**
A *file* named `dist`/`logs` at repo top level is now INCLUDED (old substring matcher excluded it). This is the intended behavior of basename-anchored dir matching, and test_create_tarball_file_named_like_excluded_dir documents it explicitly — correct trade-off, flagging for awareness.

Security: no mark_safe/@csrf_exempt/f-string-logging concerns; the change tightens credential-file exclusion. CHANGELOG present and accurate.

Clean, well-tested fix. Approve.

johnrtipton:
# Retro — PR #1519

**Title:** fix(deploy): anchor `_create_tarball` exclude matching to path segments (#1505)
**Milestone:** v1.0.0rc3 — task 2 of the milestone drain (task 1 = PR #1518, #1514/#1517 regex hardening)
**Status:** MERGED (`--admin`, CI green) — +295 / -54
**Quality rating: 5/5**

## Summary

`_create_tarball` (`python/djust/deploy_cli.py`) matched `TARBALL_EXCLUDES`
via substring containment, so `venv` excluded `venvironment.py`, `dist`
excluded `distance.py`, `media` excluded `media_helper.py`, etc. Fixed by
splitting `TARBALL_EXCLUDES` into 5 typed groups
(`EXCLUDE_DIR_NAMES`, `EXCLUDE_DIR_SUFFIXES`, `EXCLUDE_FILE_SUFFIXES`,
`EXCLUDE_FILENAMES`, `EXCLUDE_FILENAME_STEMS`) and replacing the substring
match with basename / segment / suffix / stem-anchored matching.

## What went well

- **Stage 8 caught a real 🔴 security regression that Stage 7 missed.**
  The implementer's first pass used exact-filename matching for
  `.env` / `db.sqlite3`. That *regressed* sensitive-file exclusion: the
  old substring matcher dropped any path *containing* `.env`, so
  `.env.production`, `.env.local`, `config/.env.production`,
  `db.sqlite3-wal` were excluded — exact-match would have shipped them
  into a deploy tarball (credential / live-DB disclosure). Stage 7's
  self-review checked the over-match cases and bare `.env`, but never
  constructed the `.env.production` shape. Stage 8's security review did,
  and a fixer added `EXCLUDE_FILENAME_STEMS` with the
  `file == T or file.startswith(T+".") or file.startswith(T+"-")` rule —
  which keeps every sensitive variant excluded while leaving the
  lookalike `.environment` correctly included (the #1505 over-match
  stays fixed).
- **The orchestrator caught a silent `git commit --amend` bounce.**
  The fixer's `--amend` was swallowed by a pre-commit hook reformat:
  `git log` still showed the pre-fix hash and `git status` showed `MM`
  (staged + working-tree modified, uncommitted). The fixer agent
  reported "amended commit 5ca016d0" — quoting the OLD hash, not
  verifying. The orchestrator caught it on first principles: a `git log`
  showing an *unchanged* hash after `--amend` is impossible (amend always
  rehashes), and `git show HEAD:deploy_cli.py | grep EXCLUDE_FILENAME_STEMS`
  returned 0. Re-staged + re-amended → `18f031b0` with the fix landed.
- **Gate-off self-tests (Action #254) confirmed non-tautological for
  BOTH fixes** — the over-match fix and the sensitive-stem fix. Stage 11
  independently empirically verified the stem predicate. `TestCreateTarball`
  carries 60 tests including over-match regression, genuine-exclusion
  coverage, sensitive-stem-variant cases, and the `.environment`
  lookalike. Full suite: 7268 passed, 0 failed.

## Lessons learned

- **Passive inspection vs active falsification — again.** This PR is a
  textbook reinforcement of the v1.0.0rc2 retro's defining thread
  ("verify, don't assume"). Stage 7 *inspected* the new matcher and
  missed the regression; Stage 8 *constructed the falsifying case*
  (`.env.production`) and caught it. The lesson is not "Stage 7 was
  sloppy" — it checked the over-match cases it set out to check. It's
  that a security-surface change needs the falsifying case built
  explicitly: for an exclusion rule, that means enumerating every shape
  the *old* rule excluded and confirming the new rule still excludes
  each one. Inspection of "does the new rule reject the over-matches"
  is a different question from "does the new rule still reject
  everything the old rule rejected."
- **The `--amend` bounce is a new wrinkle on Action #122.** Action #122
  documents the pre-commit-hook-swallows-a-commit failure mode and
  prescribes a `&& git log -1 --oneline` reflex. But that reflex only
  asserts "a commit subject is shown" — which is *always* true after an
  `--amend`, even when the amend bounced (the OLD commit is still HEAD,
  with its old subject). The reflex as written does not distinguish a
  successful amend from a bounced one. The fixer agent fell into exactly
  this gap: it saw a commit subject, assumed success, and quoted the
  stale hash. For `--amend` specifically, the verification must assert
  the HEAD hash *changed* — an unchanged hash after `--amend` is
  definitionally a bounce, because amend always rehashes.
- **The cross-milestone thread holds.** Both halves of this PR are the
  rc2 verify-don't-assume thread playing out: (a) Stage 7 inspected and
  missed; Stage 8 falsified and caught. (b) The fixer assumed the amend
  worked; the orchestrator verified (`git show HEAD:` grep) and caught
  the bounce. Two independent instances in one PR — the thread is not
  decaying into rc3, it is reinforcing.

## Ideas

- IDEA: Refine Action #122 — for `git commit --amend` specifically, the
  post-commit verification must capture the pre-amend HEAD hash
  (`PRE=$(git rev-parse HEAD)`) and assert `git rev-parse HEAD` differs
  after the amend. An unchanged hash is definitionally a bounce. The
  current `&& git log -1 --oneline` reflex passes a bounced amend
  because the old commit's subject is still shown. Add to the
  pipeline-run pre-commit/post-commit checklist and `CLAUDE.md`.
- IDEA: Add a Stage 8 (security review) checklist item for
  exclusion-rule / filter changes: "enumerate every shape the OLD rule
  matched; confirm the NEW rule still matches each one." This left-shifts
  the regression-class catch and makes the falsifying-case construction
  explicit rather than reviewer-discretionary.
- IDEA: Consider a fixer-agent self-check: any agent reporting a commit
  hash must obtain it from `git rev-parse HEAD` *after* the commit
  operation, never quote a hash printed before or planned. The fixer
  here quoted `5ca016d0` (a hash that was never HEAD) — sourcing the
  reported hash from a live `rev-parse` would have surfaced the bounce
  immediately.

Issue #1471, opened 2026-05-12
Sticky-child LiveView state persistence across WS reconnect (post-#1467)


johnrtipton:
## Context

Spun off from #1467 (closed as Option C — out-of-scope for v0.9.7-3).

#1467 investigation confirmed:
- **LiveComponent embedded children** persist across WS reconnect today via the parent's `_save_components_to_session` call in the PR #1466 save block — no gap.
- **Sticky-child `LiveView`s** (`{% live_render %}`-embedded full LiveViews registered via `_register_child` on `StickyChildRegistry`) do NOT persist across WS reconnect. Same gap exists in the HTTP path — neither path saves sticky-child state today.

This issue tracks the architectural design + implementation to close that gap.

## Why this needs a design pass, not a one-PR fix

The SAVE side is straightforward (write to `liveview_<parent_path>__<view_id>` per child). The LOAD side has a discovery problem:

- Sticky-children typically don't exist at the parent's `mount()` time. They're constructed during template render via `{% live_render %}`.
- At WS-reconnect mount time, the parent hasn't rendered yet — so it doesn't yet know which `view_id`s it has.
- Restoring sticky-child state needs an index (e.g., `liveview_<parent_path>__sticky_ids = [...]`) written by the render path so the next mount can iterate.

Multi-step design — separate from the one-PR shape #1467 originally assumed.

## Proposed phases

1. **Phase A — sticky-id index**: render path writes `liveview_<parent_path>__sticky_ids` listing all sticky `view_id`s created during this render. Cost: one extra `session.aset` per render of a parent with sticky-children.

2. **Phase B — SAVE side**: when `target_view is not self.view_instance` AND `target_view._view_id` is set, write `target_view`'s state to `liveview_<parent_path>__<view_id>` (instead of, or in addition to, current gate skip).

3. **Phase C — LOAD side**: after parent's `mount()`, iterate the sticky-id index and call each child's reconstruction path with the saved state. Requires the parent's render path to register the children with the saved state available so initial render reflects the restored state.

4. **Phase D — opt-in semantics**: `enable_state_snapshot` on a sticky-child should require the parent to also have it enabled (or else the persistence is silently incomplete). Document.

## Scope

v0.10.0 or post-stable.

## Refs

- #1467 (parent — closed as Option C with investigation comment)
- PR #1466 (the gate this work would generalize)
- `python/djust/mixins/sticky.py` (`StickyChildRegistry`)
[...]

PR #1526, opened 2026-05-18
feat(websocket): persist sticky-child LiveView state on save (ADR-018 18a, #1471)

johnrtipton:
## ADR-018 iter 18a — Sticky-child state persistence: SAVE side

First of three iterations implementing **[ADR-018](../blob/main/docs/adr/018-sticky-child-state-persistence.md)** (v1.0.0rc4 milestone). Part of #1471 — does **not** close it (18b LOAD side + 18c opt-in/docs still to come).

### The gap

A `LiveView` embedded with `{% live_render sticky=True %}` (a *sticky child*) is registered on the parent's `StickyChildRegistry` and routed by `view_id`. The per-event state-save block in `websocket.py` was gated `target_view is self.view_instance`, so sticky-child events were skipped entirely — a sticky child's event-driven state was silently lost across a WS reconnect. Same gap on the HTTP path.

### What this PR does (SAVE side only)

- **Stable-key scheme** (ADR Decision 1) — sticky-child state is keyed on the stable `sticky_id` class attribute, never the volatile per-process `_view_id`: `liveview_<parent_path>__sticky__<sticky_id>` (+ `__private`).
- **WS save-block generalized** (Decision 4) — `websocket.py` save section restructured into Branch A (parent, verbatim) + a new **separate-`if`** Branch B (sticky child), under the same 150 ms `asyncio.wait_for` bound.
- **HTTP parity** — a parent-driven sticky sweep on the POST path (placed post-`render_with_diff`, where `_child_views` is populated).
- **Both-opt-in gate** (Decision 5) — a sticky child is saved only when both it *and* its parent set `enable_state_snapshot = True`.
- **GC ledger** (Decision 3) — `liveview_<parent_path>__sticky_ids` records the rendered set and prunes session entries for children no longer rendered.

### Out of scope (later iterations)

- LOAD/restore side — tag-driven restore at `{% live_render %}` → **iter 18b**.
- Opt-in `djust check` + runtime warning + guide docs → **iter 18c**.
- Non-sticky (`child_N`) embeds — unsupported by design (no stable key).

### Tests

7 regression cases in `python/djust/tests/test_sticky_child_persistence_1471.py` (WS + HTTP, stable-key write, non-sticky-not-written, index, orphan-prune, both-opt-in gate, zero-cost-when-unused). Full suite green: 6721 passed. Gate-off self-test confirmed non-tautological (neutering Branch B fails cases 1/3/4/7).

Two-commit shape: implementation+tests (`498c644e`), CHANGELOG (`d425f54d`).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- **Sticky-child `LiveView`s now persist their state across a WebSocket reconnect — SAVE side (#1471, ADR-018 iter 18a).** A full `LiveView` embedded with `{% live_render sticky=True %}` is a *sticky child*: it is registered on the parent's `StickyChildRegistry` and its events are routed by `view_id`. Until now the per-event state-save block (`websocket.py`) was gated `target_view is self.view_instance`, so sticky-child events were skipped entirely — a sticky child's event-driven state was silently lost on a reconnect (page refresh, network blip, snapshot/restore), and the HTTP path had the same gap. This iteration adds the SAVE side: when a sticky-child event fires and **both** the child and its parent have `enable_state_snapshot = True`, the child's public + private state is now written to a stable session key `liveview_<parent_path>__sticky__<sticky_id>` (keyed on the child's stable `sticky_id` class attribute, never the volatile per-process `_view_id`). The same parent-driven sweep was added to the HTTP POST path so both transports persist consistently. A GC ledger `liveview_<parent_path>__sticky_ids` records the sticky ids rendered each cycle and prunes session entries for children no longer rendered. The matching LOAD/restore side ships next as ADR-018 iter 18b; opt-in enforcement + a `djust check` warning ship as iter 18c. Only `sticky=True` embeds (which have a stable `sticky_id`) are persistable; non-sticky embeds are unaffected. Covered by 7 regression cases in `python/djust/tests/test_sticky_child_persistence_1471.py`.
+
 ## [1.0.0rc3] - 2026-05-18
 
 ### Added
--- a/python/djust/mixins/request.py
+++ b/python/djust/mixins/request.py
@@ -618,6 +618,42 @@ class RequestMixin:
                 html, patches_json, version = self.render_with_diff(request)
                 t_render_ms = (time.perf_counter() - t0_render) * 1000
 
+            # ADR-018 iter 18a — HTTP sticky-child state save (Decision 4,
+            # HTTP side). The POST path has no ``view_id`` routing — it
+            # always operates on ``self`` (the parent) — so this is a
+            # parent-driven sweep, not a child-routed save. It MUST run
+            # AFTER ``render_with_diff`` because ``{% live_render %}``
+            # registers children on ``self._child_views`` during the
+            # template render (verified: ``_child_views`` is empty before
+            # the render at line ~604, populated after it here). For each
+            # registered child satisfying the both-opt-in gate, persist
+            # its state under the stable sticky key; then write the GC
+            # ledger. Django saves the session at response time.
+            from .sticky import (
+                save_sticky_child_state_sync,
+                sticky_child_should_persist,
+                write_sticky_index_and_prune_sync,
+            )
+
+            from .sticky import sticky_ids_index_key as _sticky_index_key
+
+            _sticky_children = (
+                self._get_all_child_views() if hasattr(self, "_get_all_child_views") else {}
+            )
+            _sticky_to_save = [
+                child
+                for child in _sticky_children.values()
+                if sticky_child_should_persist(child, self)
+            ]
+            # Run the save + ledger sweep when there are children to save OR a
+            # stale ledger exists (so a parent whose last sticky child was
+            # removed still gets its orphans pruned). A parent that never had
+            # sticky children pays zero cost — no ledger key, empty sweep.
+            if _sticky_to_save or _sticky_index_key(request.path) in request.session:
+                for _child in _sticky_to_save:
+                    save_sticky_child_state_sync(_child, request.session, request.path)
+                write_sticky_index_and_prune_sync(self, request.session, request.path)
+
             import json as json_module
 
             PATCH_THRESHOLD = 100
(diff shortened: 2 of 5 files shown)


johnrtipton (commented):
APPROVE — ADR-018 iter 18a (sticky-child persistence, SAVE side).

**Verification**
- Stale-base: BEHIND=0 vs origin/main.
- Tests: test_sticky_child_persistence_1471.py + test_ws_reconnect_state_1465.py → 17 passed.
- Empirical canary (#1468): gated Branch B (`websocket.py:3260`) + the HTTP sweep (`request.py:652`) off with `if False and …` → cases 1/3/4/7 FAIL, negative cases 2/5/6 still pass. Non-tautological; the implementer's gate-off claim is verified exactly.
- Regression guard: Branch A gate string intact at `websocket.py:3139-3140`; the whitespace-collapsed source-grep at `test_ws_reconnect_state_1465.py:637` still passes — Branch B is a *separate* `if`, not merged.

**ADR fidelity**
- D1: keys on stable `sticky_id` via `sticky_child_session_key` (`sticky.py`), never `_view_id`. ✓
- D3: GC ledger `write_sticky_index_and_prune` prunes orphans (+`__private`), writes sorted list; case 4 covers. ✓
- D4: HTTP sweep `request.py:621-655` correctly placed post-`render_with_diff` (618) so `_child_views` is populated. ✓
- D5: `sticky_child_should_persist` requires child sticky_id + both `enable_state_snapshot`; case 5 both halves. ✓

**Correctness checked**
- `apop`/`pop` older-Django fallback present in both async helpers (matches `websocket.py:3187-3195`).
- Branch B wrapped in the same `asyncio.wait_for(timeout=0.150)` + TimeoutError/Exception handlers; `asave()` batched once.
- Edge cases: orphan prune, two-sticky ledger, child-in/parent-out + inverse, zero-cost-no-children, HTTP symmetry — all covered.
- No print / bare-except / f-string logging in the diff. No 18b/18c scope leak.

**Concerns (non-blocking, 🟡)**
1. `save_session = getattr(mount_request, "session", None) or self.scope.get("session")` (`websocket.py:~3284`) — if those are distinct SessionStore instances only one is written. Branch A uses the identical pattern, so consistent, not a regression — a one-line precedence comment would help.
2. Branch B re-imports the sticky helpers inside the per-event hot path. Minor; a module-top import (circular-import permitting) would be cleaner.

Solid, well-scoped foundation PR. APPROVE.

johnrtipton:
## Retrospective — PR #1526 (ADR-018 iter 18a: sticky-child LiveView state persistence, SAVE side)

**Squash commit:** `88cb3a98` · **Milestone:** v1.0.0rc4 · **Issue:** #1471 (partial — does not close)

### Quality rating: 5 / 5

Clean run end-to-end. Stages 6/7/8 each landed 0 findings; Stage 11 was an APPROVE with only 2 cosmetic 🟡s. 7 regression tests, full suite 6721 green, all 12 CI checks passed, and the gate-off self-test (Action #254) empirically confirmed the new tests are non-tautological — neutering Branch B fails cases 1/3/4/7.

### What went well

- **ADR-as-spec paid off.** Each PR element maps to a numbered ADR-018 decision (stable-key = D1, GC ledger = D3, WS branch split = D4, both-opt-in gate = D5). The doc-claim-verbatim discipline (Action #1046) was effectively free here because the ADR already enumerated the contract before implementation began.
- **Branch A preserved verbatim.** The WS save block was restructured into Branch A (parent, untouched) + a *separate* `if` Branch B (sticky child). Keeping the Branch-A gate string byte-identical kept the #1466 source-grep guard green — a deliberate, low-risk refactor shape rather than an in-place rewrite.
- **Gate-off self-test at Stage 5, not just Stage 11.** The implementer ran the gate-off check itself and confirmed Branch B neutering fails 4 of 7 cases. This is exactly the left-shift Action #254 prescribes, and it meant Stage 11 had nothing tautological to catch.
- **Zero-cost-when-unused was tested, not assumed.** Case 7 pins the both-opt-in gate so the sticky sweep is a no-op when either half opts out — consistent with Action #246.
- **HTTP/WS parity in the same PR.** The same gap existed on the POST path; fixing both in one iteration avoids a known-broken transport lingering between 18a and a later cleanup.

### Lessons learned (iter 1 of 3 — foundation for 18b)

- **The stable-key scheme is the contract 18b consumes.** `liveview_<parent_path>__sticky__<sticky_id>` (+ `__private`) and the `__sticky_ids` GC ledger are now the wire format the LOAD side must read back. 18b is a pure consumer of what 18a wrote — it should not invent a new key shape. The 7 tests here double as the format pin; 18b should add round-trip tests (save in 18a's shape → load in 18b) rather than re-deriving keys.
- **Split-foundation sequencing worked as intended.** 18a ships SAVE only and is mergeable/soakable on its own (saved state with no reader is inert, not broken). 18b can build on a landed, CI-verified foundation. Per Action #1385, since djust is its own primary consumer here, 18b need not wait a full release — but the SAVE format should be treated as frozen once 18b starts.
- **Per-event in-function import is a deliberate pattern, not debt.** The websocket.py:3260 import is a circular-import-avoidance pattern (confirmed in Self-Review). 18b should mirror it rather than "fixing" it to a top-level import.

### What to improve

- **File the precedence-comment 🟡 as a tracked task for 18b instead of leaving it loose.** The `save_session` precedence comment at websocket.py:~3284 was judged non-blocking and not filed as an issue. That is fine for 18a's scope, but 18b touches the same save/load region — the comment should be folded into 18b's diff as a one-line clarification so the precedence rule is documented exactly where the LOAD path will read it. Leaving cosmetic 🟡s entirely untracked risks them never landing; attaching them to the next iteration that already touches the file is the cheapest disposition. Recommendation: add a 18b checklist item "fold in #1526 precedence comment."

### Disposition of the 2 cosmetic 🟡s

Both were judged non-blocking by Stage 11 and **not filed as separate issues**:

1. **`save_session` precedence comment (websocket.py:~3284)** — a missing clarifying comment, no behavior impact. **Recommendation: 18b should fold this in**, since 18b edits the same save/load region (see "What to improve").
2. **Per-event in-function import (websocket.py:3260)** — a deliberate circular-import-avoidance pattern per Self-Review, so effectively a non-issue. No action; 18b should preserve the pattern.

Neither warranted a tracked issue; #1 rides along with 18b, #2 is closed by design.

Issue #246, opened 2026-02-08
Add more built-in template filters (truncate, pluralize, title)


johnrtipton:
## Description

djust's Rust template engine supports a subset of Django template filters. Adding more common filters would improve compatibility with existing Django templates.

## Filters to Add

- `truncatewords` - Truncate string after N words
- `pluralize` - Return plural suffix based on value
- `title` - Convert string to title case
- `capfirst` - Capitalize first letter
- `slugify` - Convert to URL-friendly slug

## Where to Look

- Rust filters: `crates/djust_templates/src/filters.rs`
- Python fallback filters: `python/djust/templatetags/`
- Tests: `tests/unit/test_filters.py`

## Getting Started

1. Pick one filter from the list above
2. Implement in Rust (`filters.rs`) following existing patterns
3. Add Python fallback implementation
4. Add tests
5. Submit PR

Each filter is independent, so multiple people can work on different filters simultaneously.

PR #1527, opened 2026-05-18
feat(templatetags): restore sticky-child LiveView state on render (ADR-018 18b, #1471)

johnrtipton:
## ADR-018 iter 18b — Sticky-child state persistence: LOAD side

Second of three iterations implementing **ADR-018** (`docs/adr/018-sticky-child-state-persistence.md`, v1.0.0rc4 milestone). Part of #1471 — does **not** close it (18c opt-in/docs still to come). Builds on iter 18a (#1526, merged).

### What this PR does

Completes the round trip 18a started. When `{% live_render sticky=True %}` constructs a sticky child during render, the tag — **before** calling the child's `mount()` — checks the session for the state 18a saved under `liveview_<parent_path>__sticky__<sticky_id>`:

- **Tag-driven restore** (ADR Decision 2) — the restore hook lives in the `{% live_render %}` eager branch, after `check_view_auth`, before `mount()`. A single hook covers the WS, HTTP-POST and HTTP-GET render paths. On a successful restore the child's `mount()` state-init is skipped (mirroring the parent `LiveView`'s own skip-`mount()`-on-saved-state path); `_register_child` still runs so event routing works.
- **`_restore_*` replay** (ADR Decision 6) — a restored child runs `_restore_private_state` + `_restore_upload_configs` / `_restore_presence` / `_restore_listen_channels` (each `hasattr`-guarded).
- **Both-opt-in gate** (ADR Decision 5) — restore only fires when both the child and its parent set `enable_state_snapshot = True`. The 18a SAVE key format is read **unchanged** (frozen contract).
- **Fail-soft** — a corrupt/partial session entry falls through to a fresh `mount()`; the tag wraps the restore in `try/except` so a bad entry can never break the render.

Also folds in the one cosmetic 🟡 from #1526's review (a `save_session` precedence comment in `websocket.py` Branch B).

### Out of scope (iter 18c)

Opt-in enforcement `djust check` + runtime warning + guide docs.

### Tests

8 regression cases in `python/djust/tests/test_sticky_child_restore_1471_18b.py` — WS + HTTP round-trip, private-state restore, `mount()`-skip sentinel, `_restore_*` replay, no-saved-state-mounts-fresh, both-opt-in gate. 18a's 7-case suite stays green. Full suite: 6721 passed. Gate-off self-test: neutering the restore hook fails 6/8 (non-tautological).

Two-commit shape: implementation+tests (`8b6b8145`), CHANGELOG (`3f5dbd82`).


johnrtipton (commented):
APPROVE — ADR-018 iter 18b (sticky-child state restore, LOAD side).

**Stale-base (#1450):** BEHIND=0, branch current with origin/main.

**Tests:** `test_sticky_child_restore_1471_18b.py` + 18a `test_sticky_child_persistence_1471.py` + `tests/integration/test_sticky_redirect_flow.py` → 20 passed. `test_ws_reconnect_state_1465.py` → 10 passed. No regressions.

**Empirical canary (#1468):** Gating the restore hook off (`if False and sticky_kwarg:` at `live_tags.py:1738`) fails exactly 6 of 8 tests — every restore-exercising case (1,2,3,4,5,6); `test_no_saved_state_mounts_fresh` and `test_both_opt_in_gate_blocks_restore` stay green asserting the fresh-mount fallthrough. The round-trip tests genuinely exercise restore — non-tautological. Matches the implementer report.

**Plan compliance:** All 5 deliverables and all 8 planned tests present and passing.

**ADR fidelity:** Restore is tag-driven, runs BEFORE `mount()` (`live_tags.py:1738-1759`), `_register_child` still runs (`live_tags.py:1764`) so event routing survives. Decision 5 both-opt-in gate via `sticky_child_should_persist` (`sticky.py:267`); Decision 6 `_restore_*` replay present, each `hasattr`-guarded. 18a SAVE key format read unchanged.

**Scope discipline:** No 18c code — no `djust check`, no runtime opt-in warning, no guide docs. `websocket.py` change is comment-only.

**Correctness / edge cases checked:** Restore runs AFTER `check_view_auth` (`live_tags.py:1719`→`1738`) — a restore cannot bypass auth. A corrupt/non-dict session entry raises and the tag try/except (`live_tags.py:1749`) falls through to a fresh `mount()` — render never breaks. `safe_setattr(allow_private=False)` for public; private via `_restore_private_state`. No auto-reject triggers (no print, no `except: pass`, `logger.exception` uses `%r`/`%s`).

**🟡 Minor (non-blocking):** `live_tags.py:1747` passes `getattr(request, "path", "/")` as `parent_path`; 18a saves keyed on `mount_request.path`. They coincide on all current transports (plan §2), but it is a soft coupling — a POST to a URL differing from the mounting GET would key-miss. Worth a docstring note when 18c lands.

johnrtipton:
## Retrospective — PR #1527 (ADR-018 iter 18b: sticky-child LiveView state persistence, LOAD/restore side)

**Squash commit:** `b644717c` · **Milestone:** v1.0.0rc4 · **Issue:** #1471 (partial — does not close)

### Quality rating: 5 / 5

Clean run end-to-end. Stages 6/7/8 each passed; Self-Review surfaced 2 cosmetic docstring 🟡s that were folded into the implementation commit before Stage 9 (amend `d74b536f`→`8b6b8145`). Stage 11 was an APPROVE with 1 non-blocking 🟡. 8 regression tests, full suite 6721 green, all 12 CI checks passed, and the gate-off self-test (Action #254) empirically confirmed the new tests are non-tautological — neutering the restore hook fails 6 of 8 cases.

### What went well

- **The frozen-contract handoff was clean.** 18b is a pure consumer of the SAVE format 18a froze (`liveview_<parent_path>__sticky__<sticky_id>` + `__private`). The PR read the key shape back **unchanged** rather than re-deriving it — exactly what the 18a retro prescribed. The iter-to-iter contract held with zero re-litigation.
- **ADR-as-spec paid off again.** Each PR element maps to a numbered ADR-018 decision: tag-driven restore = D2, both-opt-in gate = D5, `_restore_*` replay = D6. The doc-claim-verbatim discipline (Action #1046) was effectively free because the contract was enumerated before implementation.
- **Single transport-symmetric hook point.** The restore hook lives in the `{% live_render %}` eager branch (after `check_view_auth`, before `mount()`), so one change covers WS reconnect, HTTP POST, and HTTP GET. No per-transport duplication and no known-broken transport left between iterations.
- **Mirrors the parent's existing skip-`mount()`-on-saved-state path.** Restore applies saved state in lieu of `mount()` state-init, mirroring the parent `LiveView`'s own behavior in `websocket.py:handle_mount` — a Stage-4 first-principles grep (Action #168) reused the existing pattern rather than inventing a new one.
- **Fail-soft was tested, not assumed.** A corrupt/partial session entry falls through to a fresh `mount()`; the `try/except` wrap means a bad entry can never break the parent render. The no-saved-state-mounts-fresh case pins the zero-cost-when-unused path (Action #246).
- **Gate-off self-test at Stage 5.** Neutering the restore hook fails 6/8 — well above the "at least one" bar. Stage 11 had nothing tautological to catch.
- **18a's 7-case suite stayed green** alongside 18b's 8 new cases — the round-trip is now pinned end-to-end (save in 18a's shape → load in 18b's reader).

### Lessons learned (iter 2 of 3 — consuming 18a's frozen format)

- **A frozen format is only as good as the round-trip test.** 18b's strongest move was adding tests that exercise the actual SAVE→LOAD round trip rather than asserting the reader against hand-written key strings. If 18a's format and 18b's reader had each been tested only against their own assumptions, a key-shape drift would have passed both suites independently. The shared round-trip is the real contract pin.
- **Split-foundation sequencing worked as intended.** 18a shipped SAVE-only and was inert-but-safe with no reader; 18b built on a landed, CI-verified foundation. Per Action #1385, since djust is its own primary consumer, 18b did not wait a full release — but it treated the SAVE format as genuinely frozen and did not touch it (commit body explicitly states "No change to the 18a SAVE format").
- **Carry cosmetic 🟡s to the iteration that already touches the file.** The 18a retro recommended folding 18a's loose `save_session` precedence comment into 18b's diff, and 18b did exactly that — it lands as a one-line clarification in `websocket.py` Branch B. This is the cheapest disposition for cosmetic findings and it worked: the comment landed where the LOAD path now reads it, instead of becoming an orphan 🟡.
- **Self-Review 🟡s folded pre-Stage-9 keep the two-commit shape intact.** The 2 docstring 🟡s were amended into the implementation commit before the CHANGELOG commit, preserving the impl+tests / docs+CHANGELOG boundary (Action #181). Folding cosmetic fixes into the commit they belong to — rather than a third "review fixes" commit — keeps the squash history clean.

### What to improve

- **Distinguish "soft coupling" findings from "cosmetic" findings in the carry-forward decision.** The `request.path` vs `mount_request.path` finding at `live_tags.py:1747` is **not cosmetic** — it is a latent correctness coupling that happens to be benign only because all current transports make the two paths coincide. It was correctly deferred (a docstring note, not a code change), but it should be tracked more visibly than a cosmetic docstring 🟡: it carries a real "if a future transport breaks this assumption" risk. Recommendation: when a 🟡 is "benign today, fragile under a future change," the carry-forward note for the next iteration should explicitly state the breakage condition, not just "add a docstring." 18c's checklist should record *why* the coupling holds, so a future transport author has the invariant in front of them.

### Disposition of the carried 🟡

**`parent_path` coupling — `request.path` vs `mount_request.path` (`live_tags.py:1747`).** Stage 11 flagged that the restore key is derived from `request.path`, while the SAVE side keyed off the mount request's path; the two coincide on **all current transports** (WS, HTTP POST, HTTP GET all reach `live_render()` with the same path), so there is no live bug. Stage 11 judged it non-blocking and deferred a clarifying **docstring note to iter 18c**.

- **Status:** carried forward to iter 18c (the opt-in + docs iteration), not filed as a separate issue.
- **Action for 18c:** add the docstring note at `live_tags.py:1747` documenting that the restore-key path and the SAVE-key path must agree, and stating the invariant explicitly (current transports coincide; a future transport that mounts under a different path than it renders under would silently miss restore). 18c already touches the docs surface, so this rides along at zero extra cost — the same fold-into-the-next-iteration disposition 18a used for its precedence comment.
- **Recommendation:** 18c should treat this as a tracked checklist item, not a loose 🟡, since it is a correctness invariant rather than a typo.

Issue #1523, opened 2026-05-18
a11y: surface accessibility findings in djust_audit (#1513 follow-up)


johnrtipton:
Deferred sub-area of #1513, per Action #1079. PR #1521 shipped slice 1.

## Scope

`djust_audit` (`python/djust/management/commands/djust_audit.py`) has `--ast`, `--live`, and `--permissions` modes but no accessibility section. Add an a11y reporting mode (or section) that runs the `Y` system checks (Y001–Y004, from PRs #1491 + #1512) and presents the findings in the audit's pretty + `--json` output, consistent with the existing `_output_*` shapes and the `strict` exit-code semantics.

Dogfood against `examples/demo_project` before commit (Action #1060). Self-contained — extends an existing command, no new public surface. Lower-risk than the keyboard-JS sub-area; a good standalone task or post-1.0 "a11y phase 2" milestone item.

Issue #1522, opened 2026-05-18
a11y: keyboard-interaction client JS — focus trap, Esc-to-close, roving tabindex (#1513 follow-up)


johnrtipton:
Deferred sub-area of #1513, per Action #1079. PR #1521 shipped slice 1 (P2/P3 component ARIA + decorative-icon sweep + demo Y003 fixes).

## Scope

The interactive components (modal, tabs, accordion, dropdown) have correct ARIA roles/states (PR #1491) and the P2/P3 tier now has ARIA (PR #1521) — but there is no keyboard-interaction *behavior*:
- **Focus trap** for modal/dialog — Tab/Shift+Tab must cycle within the open dialog.
- **Esc-to-close** for modal and dropdown.
- **Arrow-key roving tabindex** for tablist, accordion, and dropdown menu (W3C ARIA Authoring Practices keyboard patterns).

## Constraints

New client-side surface — likely a new module in `python/djust/static/djust/src/` (closest existing surface: `35-dj-dialog.js`). CSP-strict: marker class + delegated listener, no inline handlers (Action #183). Mind the client.js size budget (<2 KB gzipped per new module — see CLAUDE.md JS budget). Needs JSDOM tests in `tests/js/`; async-callback stubs must yield a microtask before invoking callbacks (Action #1113).

This was identified as the highest-risk / highest-complexity sub-area of #1513 and deliberately deferred from the milestone-closing PR #1521. A good candidate to cluster into a post-1.0 "accessibility phase 2" milestone alongside the issue below.

Issue #1489, opened 2026-05-18
consider re-exporting optimistic/cache/client_state/background from top-level djust __all__ (v1.1)


johnrtipton:
## Context

Finding **F3** from the v1.0.0 API-stability audit (PR #1488, `docs/API_STABILITY.md`).

The decorators `optimistic`, `cache`, `client_state`, and `background` are stable and part of the public API (present in `djust.decorators.__all__`), but are **not** re-exported from the top-level `djust` package's `__all__`. The canonical supported import path for 1.0 is:

```python
from djust.decorators import optimistic, cache, client_state, background
```

## Proposal

Adding these four to the top-level `djust` namespace is a purely **additive** change (safe under SemVer) and is more appropriately considered for a **1.1** release than bundled into the 1.0 freeze.

This issue tracks that consideration. `docs/API_STABILITY.md` §F3 references it.

PR #1530, opened 2026-05-19
fix(vdom): claim each text node once in build_fragment_text_map (#1529)

### LINKED ISSUE #1529: VDOM diff mis-paths SetText patches when 2+ {{ }} text values change in one update
## Summary

When two or more dynamic `{{ }}` text values change in a single update, the
incremental `render_with_diff()` emits `SetText` patches that all carry the
**first** changed node's path. The second value's patch lands on the first
node; the second node never updates.

Observed on **djust 1.0.0rc3** (current `main`).

## Minimal reproduction

Self-contained — no project required. Run with any Python env that has `djust`
installed. Exit code 1 = bug reproduced.

```python
"""Minimal reproduction — djust VDOM diff mis-paths the 2nd changed text node."""
import pathlib
import tempfile

import django

_tpl_dir = tempfile.mkdtemp()
pathlib.Path(_tpl_dir, "repro.html").write_text(
    '<div dj-root><div class="a">{{ a }}</div><div class="b">{{ b }}</div></div>'
)

settings.configure(
    DEBUG=True,
    SECRET_KEY="repro-only-not-secret",
    INSTALLED_APPS=[
        "django.contrib.contenttypes",
        "django.contrib.auth",
        "django.contrib.sessions",
        "channels",
        "djust",
    ],
    DATABASES={"default": {"ENGINE": "django.db.backends.sqlite3", "NAME": ":memory:"}},
    TEMPLATES=[{
        "DIRS": [_tpl_dir],
        "APP_DIRS": True,
        "OPTIONS": {"context_processors": []},
    }],
    LIVEVIEW_ALLOWED_MODULES=["__main__"],
    DJUST_CONFIG={"STATE_BACKEND": "memory"},
)
django.setup()

from django.core.management import call_command

call_command("migrate", run_syncdb=True, verbosity=0)

from djust import LiveView
from djust.testing import LiveViewTestClient


class Repro(LiveView):
    template_name = "repro.html"

        self.a = 0
        self.b = 0


client = LiveViewTestClient(Repro)
client.mount()
client.render_with_patches()  # baseline: a=0, b=0

# Two DISTINCT values, so the two patches can be told apart.
client.view_instance.a = 7
client.view_instance.b = 99

html, patches, _ = client.render_with_patches()
print("rendered HTML:", html)
for p in patches:
    print("patch:", p)

set_text = [p for p in patches if p.get("type") == "SetText"]
paths = {tuple(p["path"]) for p in set_text}
if len(set_text) >= 2 and len(paths) == 1:
    print(f"BUG: {len(set_text)} SetText patches, all path {paths.pop()}")
    raise SystemExit(1)
```

## Expected vs actual

Template: `<div dj-root><div class="a">{{ a }}</div><div class="b">{{ b }}</div></div>`.
Baseline `a=0, b=0`; then `a=7, b=99`.

**Expected** — two `SetText` patches at distinct paths:

```
{'type': 'SetText', 'path': [0, 0], 'text': '7'}    # .a text node
{'type': 'SetText', 'path': [1, 0], 'text': '99'}   # .b text node
```

**Actual** — both patches target `.a`'s path `[0, 0]`:

```
{'type': 'SetText', 'path': [0, 0], 'text': '7'}
{'type': 'SetText', 'path': [0, 0], 'text': '99'}
```

The HTML returned by `render_with_diff` is
`<div class="a" dj-id="1">99</div><div class="b" dj-id="2">0</div>` — `.a`
shows `.b`'s value (the second patch, applied last to the same node), and
`.b` never updates.

## Notes

- The full `render()` path produces correct HTML
  (`<div class="a">7</div><div class="b">99</div>`). Only the incremental
  `render_with_diff()` is affected — the bug is in the diff, not the render or
  the handler.
- Also observed in a live browser session over a real WebSocket: a counter
  view with two dynamic values updated the first on every event while the
  second stayed permanently stale.
- Practical effect: a page reliably updates only its first dynamic `{{ }}`
  text value; additional ones are mis-pathed onto it.

Distilled while building an on-device-djust proof of concept.

johnrtipton:
## Summary

Fixes #1529 — a release-blocking VDOM correctness bug. The incremental diff (`render_with_diff()`) mis-paths `SetText` patches: when 2+ dynamic `{{ }}` text values change in one update **and shared identical baseline content**, every patch carries the *first* changed node's path. Net effect — a page reliably updates only its first dynamic `{{ }}` text value; later ones are mis-pathed onto it.

## Root cause

`build_fragment_text_map` (`crates/djust_live/src/lib.rs`) maps each rendered template fragment to the **first** VDOM text node whose content string equals the fragment. Content equality is not a unique key: two template variables that render the same baseline string (e.g. `{{ a }}` and `{{ b }}` both `0` at mount) both match the *first* such node, collapsing both map entries onto one path. The text-fast-path then emits every `SetText` patch at that single path — and mutates the in-memory VDOM node there twice while its sibling is never touched.

The VDOM differ itself (`crates/djust_vdom/src/diff.rs`) was checked and is **correct** — it's only reached on the full-parse path the bug doesn't hit.

## Fix

Track a `Vec<bool>` parallel to the collected text nodes and claim each node at most once — the first *unclaimed* matching node. The fragment→node map becomes a bijection over matched fragments instead of a many-to-one collapse. Both the fragment list and the text-node collection are document-ordered, so first-unclaimed-match is positionally stable. No change to the VDOM differ, parser, patch types, or the patch-emission loop.

## Tests

`tests/unit/test_vdom_settext_mispath_1529.py` — 6 regression cases: two/three identical-baseline values, nested element, text+attribute change together, only-second-changed (sharpest mapping assertion), and a distinct-baseline regression guard. Gate-off self-test (Action #254) confirmed: reverting the fix fails 4/6 cases including both load-bearing ones — non-tautological.

- Full suite: Python 4369 passed, JS 1572 passed, `cargo test -p djust_vdom` all green.
- The issue's self-contained reproducer now prints `NO BUG` and exits 0.


johnrtipton (commented):
## Stage 11 Code Review — PR #1530 — APPROVE

**Plan conformance:** Implementation matches the plan exactly. `build_fragment_text_map` (lib.rs:2575-2598) adds the `Vec<bool> claimed` parallel to `text_nodes`, claims the first *unclaimed* content-match per fragment. All 6 planned tests exist (core, 3-value, nested, text+attr, second-only, distinct-baseline).

**Correctness:** Verified the document-order assumption holds on both sides — `collect_vdom_text_nodes` (lib.rs:2603-2622) is a pre-order DFS; fragments are pushed in `nodes.iter()` order (renderer.rs:269/340). For a fragment with NO matching node, the entry is simply absent from the map; the `all_mapped` guard (lib.rs:405-408) then bails the whole batch to the full-parse differ — so a collision or miss can only *degrade* to a correct full diff, never emit a wrong patch. Skipped fragments (`<`/empty, lib.rs:2587) correctly never claim.

**Gate-off self-test (#254):** Reverted the fix, rebuilt, re-ran — 4/6 tests fail (`[0,0]` collapse confirmed); the 2 passing are the intended non-collapse cases. Non-tautological.

**Tests:** `test_vdom_settext_mispath_1529.py` 6/6 pass; `cargo test -p djust_vdom` 70+ pass; diff-adjacent (1205, streaming, async-render) 64 pass. No print/except:pass/phantom modules.

**Edge cases checked:** identical text across 3 siblings, nested deeper paths, text+attr mixed, only-second-changed, distinct baselines.

**Question:** could a fragment that itself wraps multiple text nodes claim only one and leave a sibling fragment mis-matched? The `all_mapped` guard covers it, but a comment noting that intentional fall-through would help future readers.

CHANGELOG entry well-scoped. APPROVE.

johnrtipton:
## Retrospective — PR #1530 — fix(vdom): claim each text node once in `build_fragment_text_map` (#1529)

**Quality: 5/5**

A clean, surgical fix for a release-blocking VDOM correctness bug. Tight
root-cause trace, minimal blast radius (one function), reproducer-first TDD,
6 regression cases, gate-off self-test confirmed non-tautological, 0 🔴 / 0 🟡
at Stage 11, 0 CI failures, merged same day.

### Review stats

- Tests added: **6** (`tests/unit/test_vdom_settext_mispath_1529.py`)
- 🔴 findings: **0**
- 🟡 findings: **0**
- CI failures: **0**
- Diff: +327 / −3, 2 commits (impl+tests / docs+CHANGELOG — two-commit shape per Action #181)

### What went well

- **Reproducer-first, symptom-up tracing.** The plan opened from the issue's
  self-contained reproducer (`scratch/repro_1529.py`) and traced the symptom
  ("page updates only its first dynamic `{{ }}` value") backwards to the
  offending code. It explicitly *cleared* the obvious suspect — the VDOM differ
  (`crates/djust_vdom/src/diff.rs` `diff_indexed_children`) — confirming it
  walks siblings correctly and is only reached on the full-parse path the bug
  never hits. The real defect was upstream in the `djust_live` text-fast-path's
  `build_fragment_text_map`. This is the CLAUDE.md "trust the symptom, not the
  cited path" discipline applied well.

- **Empirical narrowing of the trigger condition.** The planning subagent
  didn't just describe the bug — it ran the variants and pinned the exact
  trigger boundary on current `main`: `a=0,b=0` (identical baselines)
  reproduces; `a=1,b=2` (distinct baselines) does NOT; single-value change does
  NOT. That narrowing *proved* the root cause was content-based first-match
  (content equality is not a unique key), not a path-accumulation bug in the
  differ. It also directly produced two test cases — the distinct-baseline
  regression guard and the only-second-changed sharpest-mapping assertion.

- **Honest test-layer decision.** A Rust unit test was evaluated and *rejected*
  with a documented reason: `djust_live` is `crate-type = ["cdylib", "rlib"]`
  with the PyO3 `extension-module` feature, so `cargo test --lib` fails to link
  (`ld: symbol(s) not found ... Py_True`). Rather than bolt Python-linking test
  infra onto the crate for one bugfix, the regression test went to the Python
  layer, exercising the real `build_fragment_text_map` path end-to-end through
  PyO3 via `render_with_patches`. That pins the user-observable contract
  directly.

- **Gate-off self-test (Action #254) done at Stage 5 and re-verified at Stage 11.**
  Reverting the fix fails 4/6 cases including both load-bearing ones
  (identical-baseline two-value, only-second-changed). The 2 still-passing
  cases are the intended non-collapse scenarios. Non-tautological, verified
  twice independently.

- **Minimal, correct fix shape.** A `Vec<bool> claimed` parallel to the text
  nodes turns a many-to-one collapse into a bijection over matched fragments.
  No change to the differ, parser, patch types, or the patch-emission loop —
  the emission loop is correct once the map it reads is correct. The `all_mapped`
  guard means a miss or collision can only *degrade* to a correct full-parse
  diff, never emit a wrong patch.

### What could be improved

- **The empirical trigger-narrowing move deserves canonicalization.** The
  single highest-leverage action in this task was the planner running the
  bug variants to pin the exact trigger boundary (identical vs distinct
  baselines) *before* architecting the fix. That converted a vague symptom into
  a proven root cause and produced two test cases for free. This is a specific,
  repeatable Stage-4 technique for content-/value-dependent bugs and isn't yet
  named in CLAUDE.md's bug-report-triage section — it should be. Proposed rule:
  "For a bug whose reproduction depends on input *values* (not just structure),
  bisect the trigger empirically — find the smallest value-change that flips
  the bug on/off — before writing the fix. The boundary IS the root-cause
  proof and seeds the regression-guard test."

- **`djust_live` has no `cargo test` harness — a structural gap.** The bug
  lives in `build_fragment_text_map` / `collect_vdom_text_nodes` inside
  `djust_live`, but that crate cannot be unit-tested in isolation (PyO3
  `extension-module` link failure). Every regression for `djust_live` logic
  must route through the Python layer. That's an acceptable workaround here,
  but it means a whole crate's internal helpers have no fast Rust-native test
  feedback loop. Worth a follow-up issue to evaluate a `#[cfg(test)]` module
  guarded against the `extension-module` feature, or a thin `rlib`-only test
  shim — so future `djust_live`-internal bugs can get a sub-second Rust
  regression test instead of a full PyO3 round-trip.

- **Stage 11 left one open question unaddressed.** The reviewer asked whether a
  fragment wrapping multiple text nodes could claim one and leave a sibling
  mis-matched, and noted "a comment noting that intentional fall-through would
  help future readers." The `all_mapped` guard covers it correctly (it bails to
  the full-parse differ), so this was correctly not a blocker — but the
  suggested explanatory comment was not added. Minor; a one-line follow-up.

### Relation to the broader v1.0.0rc4 drain

This is the **first task of the v1.0.0rc4 Phase-2 batch** — the remaining rc4
tasks (#1522+#1523 a11y phase 2, #1432 free-threading, #1489 top-level
re-exports) have not run yet. Notably, #1529 is a **release-blocking
correctness bug that surfaced after the rc4 scope was already set** — it was
inserted ahead of the planned Phase-2 work because a page reliably dropping all
but its first dynamic `{{ }}` update is not shippable in a release candidate.
Clearing it first de-risks the rest of rc4: the VDOM text-fast-path is now
correct, so the a11y and re-export work that follows builds on a sound diff
path rather than racing a known correctness regression to the 1.0 line.

### Lessons learned

1. Empirically bisect value-dependent bug triggers before architecting — the
   boundary is the root-cause proof. (Candidate for CLAUDE.md canon.)
2. When a crate can't be unit-tested in isolation, document *why* in the plan
   and route the regression test to the layer that can — don't add test infra
   for one bugfix.
3. Clearing a known correctness regression first is the right sequencing for a
   release-candidate drain, even when it preempts the planned scope.

Issue #1534, opened 2026-05-19
Free-threaded hardening / perf polish (follow-up to #1432)


johnrtipton:
Follow-up to #1432, which declared the \`djust._rust\` PyO3 module
\`#[pymodule(gil_used = false)]\` after a full thread-safety audit
(audit record: \`.pipeline-state/feat-1432-free-threaded-safe-plan.md\`).

The audit verdict was **(A) Safe as-is** — every global, \`#[pyclass]\`,
cross-thread \`Py<T>\`, the Tokio actor system, and the template
registries are \`Sync\`-correct and properly synchronized. Per Action
#1079 (fix exactly what the cited issue needs; file a follow-up for the
systemic remainder), the following non-blocking items were deliberately
left out of #1432's scope and are collected here as a post-1.0
"free-threaded hardening / perf polish" bucket.

### 1. \`COMPONENT_REGISTRY\` \`Send + Sync\` bounds

\`COMPONENT_REGISTRY\` (\`crates/djust_components/src/lib.rs:95\`) is a
\`Lazy<RwLock<HashMap<String, Box<dyn Component>>>>\`. The \`Component\`
trait (\`crates/djust_components/src/lib.rs:56\`) is **not** declared
\`Component: Send + Sync\`, so \`Box<dyn Component>\` is neither, and
\`RwLock<HashMap<.., Box<dyn Component>>>\` is therefore not \`Sync\`.

This is **not reachable from the \`_rust\` module today** — none of the
three accessors (\`register_component\` / \`get_component\` /
\`unregister_component\`) is exported via \`_rust\`, and it is
\`RwLock\`-guarded regardless — so it was not a blocker for #1432. But if
a future PR exports those accessors, the missing trait bounds become a
real hazard under free-threading.

**Action:** either add \`Component: Send + Sync\` bounds to the trait, or
delete the unused \`COMPONENT_REGISTRY\` + its three accessors if they are
genuinely dead code.

### 2. Optional \`#[pyclass(frozen)]\` on POD pyclasses

\`SupervisorStatsPy\` (\`crates/djust_live/src/lib.rs:1412\`, immutable POD,
two \`#[pyo3(get)]\` fields) and \`SessionActorHandlePy\`
(\`crates/djust_live/src/lib.rs:1094\`, all-\`&self\` methods, only holds a
clonable \`mpsc::Sender\`) have no \`&mut self\` methods and are sound
[...]

Issue #1533, opened 2026-05-19
tech-debt: dropdown nested inside a role=dialog gets no arrow/Esc keyboard routing


johnrtipton:
Stage 11 code-review finding on PR #1532 (#1522 keyboard-nav module), 🟡 non-blocking.

## Observed

In `python/djust/static/djust/src/51-keyboard-nav.js`, the keydown handler's
dialog branch returns early once it determines the event target is inside an
open `role="dialog"`. As a result, a `dropdown` component rendered *inside* a
modal/dialog does not receive its arrow-key roving navigation or its
Esc-to-close routing — the dialog branch's modal-first precedence consumes the
event.

## Why it's deferred (not a 1.0 blocker)

- A dropdown nested inside a modal dialog is an uncommon composition.
- The modal-first precedence (Esc closes the dialog, focus trap acts on the
  dialog) is intentional and documented behavior — it is *a* correct answer,
  just not the richest one.

## Fix direction

Let the dialog branch fall through to the dropdown/menu handling when the
event target is within a dropdown that is itself within the dialog — i.e.
route arrow keys to the innermost interactive widget, and only let Esc bubble
to the dialog once the inner dropdown (if open) has been closed. Mirror the
nested-dialog stack discipline already in `_dialogStack`.

Source: PR #1532 Stage 11 review.

PR #1532, opened 2026-05-19
feat(a11y): keyboard-nav module + djust_audit --a11y mode (#1522, #1523)

johnrtipton:
## Summary

v1.0.0rc4 "accessibility phase 2" — the two deferred sub-areas of #1513's a11y long-tail, shipped as one grouped PR.

Closes #1522
Closes #1523

## #1522 — keyboard-interaction client JS

The interactive components (modal, tabs, accordion, dropdown) had correct ARIA roles/states (PRs #1491/#1521) but no keyboard *behavior*. New module `python/djust/static/djust/src/51-keyboard-nav.js`:

- **Focus trap** for modal/dialog — Tab/Shift+Tab cycle within the open dialog, with a nested-dialog stack and a no-focusable-children edge case; focus moves in on open and is restored on close.
- **Esc-to-close** for modal and dropdown.
- **Arrow-key roving tabindex** for tablist, accordion, and dropdown menu (W3C ARIA Authoring Practices keyboard patterns). Accordion headers stay individually Tab-reachable per APG; arrows are focus-movement only.

CSP-strict (Action #183): one delegated `keydown` listener on `document` + a `MutationObserver` for focus-in/restore — no inline handlers, no template changes (it hooks the existing ARIA markup). Bundle delta: **+1121 B gzipped**, under the 2 KB budget.

## #1523 — `djust_audit --a11y` mode

New `--a11y` mode in `djust_audit` that runs the `Y` accessibility system checks (Y001–Y004) and reports findings in pretty + `--json` output. Y findings are all warnings, so normal mode exits 0; `--strict` exits 1 on any finding. Mirrors the existing `--ast` mode architecture (`_run_a11y_audit` / `_output_a11y_pretty` / `_a11y_findings_to_json`). Dogfooded against `examples/demo_project` — 0 findings, no false positives.

## Tests

- `tests/js/keyboard_nav.test.js` — 27 JSDOM cases (focus-trap wrap both directions, no-focusable-children, nested-dialog stack, Esc-close, arrow roving per component).
- `TestA11yMode` in `python/tests/test_audit_command.py` — 7 cases (pretty, `--json`, `--strict` exit code, no-findings).
- Gate-off self-test (Action #1468/#254) confirmed for both: neutralizing each feature fails 22/27 JS and 7/7 Python tests — non-tautological.
- Full suite: Python 4376, JS 1599 — all green.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1535, opened 2026-05-19
feat: declare djust._rust free-threaded-safe (gil_used = false) (#1432)

### LINKED ISSUE #1432: Declare djust._rust free-threaded-safe so 3.13t/3.14t users keep no-GIL
# Declare `djust._rust` free-threaded-safe so 3.13t/3.14t users keep no-GIL

## Summary

When importing `djust._rust` into a free-threaded CPython interpreter
(`python3.13t`, `python3.14t`, …), CPython auto-re-enables the GIL for
the rest of the process and emits:

> `RuntimeWarning: The global interpreter lock (GIL) has been enabled
> to load module 'djust._rust', which has not declared that it can run
> safely without the GIL. To override this behavior and keep the GIL
> disabled (at your own risk), run with PYTHON_GIL=0 or -Xgil=0.`

This is the runtime's safety net: any extension that hasn't formally
declared free-threading support is assumed unsafe. It silently
downgrades the user's whole interpreter back to the GIL'd path.

We can lift this — PyO3 0.23+ adds an explicit `gil_used` attribute on
`#[pymodule]`. Setting `gil_used = false` writes the
`Py_mod_gil = Py_MOD_GIL_NOT_USED` slot into the generated module init,
which is exactly what CPython looks for to skip the auto-re-enable.

The change is one line (plus the audit that justifies it).

## Why now / why this is worth the audit

We tested. The free-threaded ecosystem has matured a lot in the last
~6 months:

- `python3.14t` is shipping wheels for every dep `djust-scaffold`
  pulls in (`cffi`, `cryptography`, `twisted`, `daphne`, `uvicorn`,
  `httptools`, `websockets`, `uvloop`, `watchfiles`, `zstandard`,
  `msgpack`, `nh3`, `pyyaml`, `pyopenssl`). All 24 compiled `.so` files
  in our test venv are tagged `cpython-314t-*` — built with proper
  free-threaded support.
- The only hard refusal in the dep graph is `orjson` (which djust uses
  via `importlib.util.find_spec` with stdlib `json` fallback — not a
  hard requirement at runtime).
- With `PYTHON_GIL=0` forced, `import djust._rust` and the rest of
  djust (`djust.websocket`, `djust.state_backends.redis`, ...) load
  cleanly. The runtime stays no-GIL.

We then ran a real load test on the host (Hetzner Xeon E3-1275 v6,
Ubuntu 24.04):

| | 3.12 GIL'd | 3.14t + `PYTHON_GIL=0` |
|---|---|---|
| Throughput | 794 req/s | 513 req/s |
| Latency p50/p95/p99 | n/a | 61 / 73 / 114 ms |
| Failed requests | 0 / 5000 | **0 / 5000** |
| uvicorn alive after load | yes | **yes** |
| Crashes / segfaults | 0 | **0** |

**No crashes** under 5000 concurrent requests at concurrency 32. The
Rust extension survives concurrent Python-side calls under no-GIL on
this workload.

(Throughput is lower because uvicorn is single-event-loop async — no
thread parallelism for the no-GIL build to unlock here. Workloads
that DO use threads — e.g., Django sync views with worker threads,
async-to-sync executor pools, ThreadPoolExecutor usage in user code —
would get the actual win. That's the case we'd unlock by adding the
declaration.)

## What's changing in code

Single-line change at `crates/djust_live/src/lib.rs:2655`:

```rust
#[pymodule]                       // before
fn _rust(m: &Bound<'_, PyModule>) -> PyResult<()> { ... }

#[pymodule(gil_used = false)]     // after
fn _rust(m: &Bound<'_, PyModule>) -> PyResult<()> { ... }
```

PyO3 ≥ 0.23 supports the attribute; the workspace pins
`pyo3 = { version = "0.25" }`, so we're already on a supporting
version.

## What needs to happen first: thread-safety audit

Adding the declaration without verifying is the same anti-pattern we
just shipped a fix for in the zstandard race (#1430): silent shared
mutable state under no-GIL produces SIGSEGVs that don't surface until
production. Don't do that here.

The audit checklist for the existing code in `crates/djust_live/src/`
and its callers:

- [ ] **Globals**: `TEMPLATE_CACHE` is `Lazy<DashMap<String, Arc<Template>>>` — DashMap is `Sync`; ✓.
      `SUPERVISOR` is `Lazy<Arc<ActorSupervisor>>` — needs to verify ActorSupervisor is `Sync`.
      `SUPERVISOR_STARTED` is `AtomicBool` — ✓.
      Walk every other `static` / `Lazy` / `OnceCell` and confirm the inner type is `Sync`.
- [ ] **`#[pyclass]` types**: `RustLiveView`, `RustLiveViewBackend`, `SessionActorHandle*`, `SupervisorStatsPy`,
      and the `RustAlert`/`RustAvatar`/`RustBadge`/… components.
      Under free-threaded, multiple Python threads may concurrently call methods on the same instance.
      Currently the GIL implicitly serialized this — without it, `&mut self` methods that mutate
      internal `RefCell`/`Cell`/`Vec`/`HashMap` need explicit `Mutex`/`RwLock` or atomic types.
      PyO3 0.25 has `#[pyclass(frozen)]` for read-only classes that need no locking.
- [ ] **Python-side state held by `PyObject` / `Py<T>`**: PyO3 0.23+ deprecated `Python::with_gil` in
      favor of `Python::attach`/`Python::detach` and adjusted reference-counting semantics for
      free-threading. Audit for anywhere we hold a `PyObject` across a thread boundary or store one in a
      `Sync` static — the `Py<T>` smart pointer needs care.
- [ ] **Actor system**: `ActorSupervisor` + `SessionActorHandle` use Tokio. Tokio works under
      free-threaded but the scheduler interaction with PyO3 callbacks deserves a look.
- [ ] **Template registries**: `djust_templates::registry::*` — assign-tag and filter registries
      look like `RwLock<HashMap<...>>`-style state. Confirm the lock granularity holds under
      concurrent reads/writes from many Python threads.
- [ ] **Recursive Python ↔ Rust conversions** in `python_to_value` / `python_to_json_value`
      (lines ~1455 and ~1022): these traverse Python objects under what was previously the GIL.
      Free-threaded Python guarantees these are race-free as long as we hold a borrow; verify
      no helper releases the borrow mid-traversal.

## Suggested staging

1. **Audit pass** (the bullets above). Document any locking / atomic
   changes in a separate PR — this is genuine work, not a yak shave.
2. **Add a regression-style test** that mirrors the load test in this
   issue: spawn N threads (free-threaded build only), each calling
   the public `_rust` entrypoints (template render, VDOM diff,
   `RustLiveView` mount, etc.) concurrently against a shared backend.
   Assert no exceptions, no segfault. The existing Phase-11 MVP
   numbers (5000 req @ c=32) are a starting point but a per-feature
   stress test would catch regressions earlier.
3. **Flip the declaration** in a small follow-up PR once the audit is
   green. CHANGELOG note: "djust now declares free-threaded-safe;
   `python3.14t` users no longer auto-re-enable the GIL on import."

## Out of scope for this issue

- Performance work to actually exploit free-threading (e.g., parallel
  template rendering across threads). The declaration is a
  correctness change, not a perf change.
- Bumping a minimum Python version. djust continues to support 3.12
  GIL'd unchanged.

## Evidence / repro

Tested 2026-05-08:

```
$ uv python install 3.14t       # cpython-3.14.4+freethreaded
$ uv venv /tmp/v314t --python 3.14t
$ cd djust-scaffold && uv pip install --python /tmp/v314t/bin/python -e .[dev]
  # all 49 deps install (after dropping orjson/psycopg2-binary which we
  # don't use at runtime)
$ PYTHON_GIL=0 PYTHONPATH=. /tmp/v314t/bin/python smoke.py
  # django.setup(); GIL stays off; serves /health/ and /
$ PYTHON_GIL=0 ... /tmp/v314t/bin/python -m uvicorn scaffold_project.asgi:application ...
$ ab -n 5000 -c 32 http://127.0.0.1:9014/health/
  # 0 failed; uvicorn alive; djust._rust survives concurrent calls
```

Happy to put up a PR with the declaration once the audit lands —
either I do the audit and PR both, or someone closer to the Rust
internals takes the audit and I'll handle the test scaffolding side.
EOF

johnrtipton:
## Summary

Closes #1432

Importing `djust._rust` into a free-threaded CPython interpreter (`python3.13t` / `python3.14t`) currently makes CPython auto-re-enable the GIL for the whole process — emitting a `RuntimeWarning` and silently downgrading every no-GIL user — because the extension never declared free-threading support. This PR flips the `_rust` PyO3 module to `#[pymodule(gil_used = false)]` (PyO3 0.25), writing the `Py_mod_gil = Py_MOD_GIL_NOT_USED` slot CPython reads to skip the auto-re-enable.

## The audit (the real work)

Declaring free-threaded-safety without verifying it is the #1430 zstandard-race anti-pattern. A full thread-safety audit of everything reachable through `_rust` found **no shared mutable state lacking correct synchronization** — verdict: safe as-is.

- **Globals** — all 18 `static`/`Lazy`/`OnceLock`/`thread_local!` symbols are `Sync` and correctly synchronized (`TEMPLATE_CACHE` `DashMap`, `SUPERVISOR` `Arc<ActorSupervisor>`, `SUPERVISOR_STARTED` `AtomicBool`, registries `Mutex<HashMap>`, read-only `Lazy<Regex>`).
- **`#[pyclass]` types** — all ~25 are per-instance-owned (covered by PyO3 0.23+'s free-threading-preserved borrow-checker) or hold only clonable/immutable state; none holds shared state beyond PyO3 protection.
- **Cross-thread `Py<T>`** — moved with `Send`, dereferenced only under `Python::with_gil`; ref-count-safe.
- **Tokio actor system, template registries, recursive converters** — all verified.

Rust's `Send`/`Sync` auto-trait checking statically verifies every `static` — this is *not* the #1430 pattern (a C-extension object with no Rust `Sync` guarantee).

## Tests

- `crates/djust_templates/tests/free_threaded_safety.rs` — 4 `std::thread` stress tests (`Template` `OnceLock` resolve race, concurrent parse+render, concurrent registry lookups, deterministic hash).
- `crates/djust_vdom/tests/free_threaded_safety.rs` — 2 `std::thread` tests (concurrent parse+diff, `thread_local!` per-thread independence).
- `python/djust/tests/test_rust_module_thread_safety.py` — Python `threading` call-path smoke test (honestly documented as a call-path test, not a no-GIL race detector — the Rust `std::thread` tests are the real gate).
- Gate-off self-test (Action #1468/#254): regressing `djust_vdom`'s `thread_local! ID_COUNTER` to a process-global counter makes `concurrent_id_counter_is_per_thread` fail with its exact diagnostic — the test catches the race class it claims to.

GIL'd interpreters (3.12, standard 3.13/3.14) are entirely unaffected. Out-of-scope free-threading hardening is tracked in #1534.


johnrtipton (commented):
Stage 11 (Code Review) — feature pipeline. **APPROVE.**

Stale-base: `BEHIND=0`.

**Pressure-tested the correctness claim.** Spot-checked the 4 riskiest audit rows:
- `ActorSupervisor` (`supervisor.rs:24-29`): only `Arc<DashMap<String,SessionInfo>>` + `Duration` — both Sync; `SessionInfo` (`:17-21`) = `SessionActorHandle` + 2 `Instant`. Auto-derived `Send+Sync` confirmed.
- Registry lock scope (`registry.rs:285-294`): the `Py<PyAny>` handler is `clone_ref`-ed *inside* the `BLOCK_TAG_HANDLERS.lock()` block; the `MutexGuard` drops at the closing `}` before the Python callback runs. Plan claim verified.
- No `Rc`/`RefCell`/raw-pointer in any `#[pyclass]` crate. `ID_COUNTER` (`djust_vdom/src/lib.rs:56`) is genuinely `thread_local! Cell<u64>`.

**The flip:** `#[pymodule(gil_used = false)]` on the correct `_rust` module (`djust_live/src/lib.rs:2674`); pyo3 0.25 supports it; comment cites only committed artifacts (#1432 + the two test files), no gitignored path.

**Tests genuinely concurrent:** all 6 Rust cases barrier-sync N=12 threads racing the *same* `Arc<Template>`/registry/`ID_COUNTER` for 200 iters — real OS-thread parallelism, not sequential. `concurrent_id_counter_is_per_thread` would flake if `ID_COUNTER` regressed to an unsynced global. Python smoke test docstring is honest it cannot catch a no-GIL race on GIL`d 3.12.

Verified: 6/6 Rust tests pass, Python 5 passed/1 skipped, `make dev-build` OK.

🟡 Minor: the test-file `//!` docstrings cite the gitignored `.pipeline-state/` plan path — informational only; audit is also on #1432. Non-blocking.

johnrtipton:
# Retrospective — PR #1535

**feat: declare djust._rust free-threaded-safe (gil_used = false) (#1432)**

Task 3 of 4 — v1.0.0rc4 Phase-2 drain. Merged 2026-05-19 (`mergedAt 02:17:10Z`).
+550 / -1. Closes #1432.

## Quality rating: 5/5

A one-attribute flip (`#[pymodule(gil_used = false)]`) backed by a genuine
thread-safety audit, not a rubber-stamp. The pipeline correctly identified
that the *substance* of this issue was the audit — declaring free-threaded
safety without verifying it ships the #1430 zstandard-race class of latent
production SIGSEGV — and treated the audit as the deliverable. Verdict (A)
"safe as-is" was reached with file:line citations for all 18 globals and
~25 pyclasses, the gate-off canary was genuine, and the one 🟡 was fixed
in-pipeline. Nothing unsafe shipped.

## What went well

- **Rigorous Stage-4 audit with verdict-A and file:line citations.** Every
  `static`/`Lazy`/`OnceLock`/`thread_local!` symbol reachable through
  `_rust` was enumerated in a table with an explicit `Sync` verdict and a
  source citation — `TEMPLATE_CACHE` (`djust_live/src/lib.rs:40`,
  `DashMap`), `SUPERVISOR` (`lib.rs:44`, `Arc<ActorSupervisor>`),
  `SUPERVISOR_STARTED` (`lib.rs:48`, `AtomicBool` `swap` CAS),
  `Template.resolved/.flattened` (`djust_templates/src/lib.rs:84,86`,
  canonical race-free `OnceLock` idiom), down to the `thread_local!`
  `ID_COUNTER`. The audit also correctly *scoped out* the one smell —
  `COMPONENT_REGISTRY`'s `Box<dyn Component>` lacking `Send + Sync`
  bounds — by proving it is not reachable from the `_rust` surface (no
  exported accessor) and is `RwLock`-guarded regardless. This is the
  right way to substantiate a free-threading declaration: Rust's static
  `Send`/`Sync` auto-trait checking does the heavy lifting, and the audit
  documents that it does.

- **Honest test-placement decision around `djust_live`'s untestability.**
  `djust_live` carries the unconditional PyO3 `extension-module` feature,
  so `make test` runs `cargo test --workspace --exclude djust_live` — the
  crate that actually holds the `_rust` entry point cannot be `cargo
  test`'d. Rather than skipping the regression guards, the pipeline placed
  the 6 `std::thread` concurrency stress tests in the *testable*
  `djust_templates` (4) and `djust_vdom` (2) crates, which exercise the
  same `Sync` contracts via genuine no-GIL OS-thread parallelism. The PR
  body and commit message both document this constraint explicitly rather
  than hiding it.

- **Genuine gate-off canary (Action #1468/#254).** Regressing
  `djust_vdom`'s `thread_local! ID_COUNTER` to a process-global counter
  made `concurrent_id_counter_is_per_thread` fail with its exact
  "ID_COUNTER leaked across threads" diagnostic — the test demonstrably
  catches the race class it claims to, not a tautology.

- **Honest Python smoke-test framing.** `test_rust_module_thread_safety.py`
  is documented as a GIL'd-3.12 call-path smoke test, *not* a no-GIL race
  detector, with the free-threaded-only assertion `skipif`-guarded for a
  future 3.14t CI runner. No overclaiming of what the Python suite proves.

- **Did not ship an unsafe declaration.** The whole point: the pipeline
  could have done the one-line flip and called it done. It did the audit
  first, and the audit's verdict — not the flip — is what gated the merge.

## What could be improved

- **The gitignored-`.pipeline-state/`-path-in-source-comments defect
  recurred across three sites and was fixed in two passes.** Stage 7
  self-review caught the dangling reference in `lib.rs` — the declaration
  comment cited the gitignored `.pipeline-state/...plan.md` — and fixed it
  by amending the impl commit. But the implementer had ALSO put the same
  gitignored path in the two `free_threaded_safety.rs` module docstrings
  (`djust_templates` and `djust_vdom`), and Stage 7's fix did not sweep
  those. Stage 11 then caught the recurrence as its single 🟡, fixed in
  follow-up commit `01e4e5b4`.

  When fixing a defect of the form "this referenced artifact is
  gitignored / not on a fresh clone," the fix should be a
  **grep-all-occurrences pass**, not a point-fix at the first site found —
  exactly mirroring the symbol-migration grep canon (CLAUDE.md #1391 /
  #1400: "grep for the EXACT pre-fix expression across all consumer
  directories"). A `grep -rn '\.pipeline-state/' crates/ python/` when the
  `lib.rs` site was found at Stage 7 would have surfaced all three
  occurrences in one pass and folded the fix into the same amend, saving
  the Stage 11 roundtrip. Process observation, not a code defect — the
  shipped result is correct.

## Relation to the v1.0.0rc4 Phase-2 drain

This is the **3rd of 4** Phase-2 tasks. Following #1530 (the
`build_fragment_text_map` VDOM-diff claim-once bugfix, 5/5) and #1532 (the
a11y keyboard-nav + `djust_audit --a11y` group, 5/5), only **#1489
remains** to close the drain.

A cross-task signal worth flagging: the `djust_live`-cannot-be-`cargo
test`'d constraint surfaced in **both** the #1529 pipeline (PR #1530) and
here. Twice in one drain, a Rust-side feature touching `djust_live` had to
route its regression guards into sibling crates because the entry-point
crate carries the unconditional `extension-module` PyO3 feature. That
recurrence is itself a signal — it is no longer a one-off quirk but a
standing structural constraint on where Rust tests for `_rust`-surface work
can live. Worth considering whether `djust_live` should gate the
`extension-module` feature behind a Cargo feature flag (off for
`cargo test`), so the entry-point crate becomes testable directly. A
candidate follow-up; out of scope here, and the sibling-crate placement is
a sound workaround in the meantime.

This task also continues the drain's consistent discipline thread —
genuine gate-off canaries (#1468/#254) and honest scoping of out-of-bounds
work into follow-up issues (#1534 here, #1533 in #1532, mirroring Action
#1079).

## Review stats

| Metric | Value |
|---|---|
| Tests added | 6 Rust `std::thread` concurrency (4 `djust_templates` + 2 `djust_vdom`) + ~6 Python smoke (`test_rust_module_thread_safety.py`) |
| 🔴 blocking findings | 0 |
| 🟡 non-blocking findings | 1 — gitignored `.pipeline-state/` path in two test docstrings; fixed in-pipeline (`01e4e5b4`) |
| CI failures | 0 (6/6 Rust pass; Python 5 passed / 1 skipped; `make dev-build` OK; 12 checks green) |
| Stale-base check | BEHIND=0 |
| Follow-up filed | #1534 — out-of-scope free-threading hardening (`COMPONENT_REGISTRY` `Send+Sync` bounds, optional `#[pyclass(frozen)]`, optional `Mutex`→`RwLock`, python3.14t CI leg) |

PR #1536, opened 2026-05-19
feat: re-export optimistic/cache/client_state/background from djust top level (#1489)

johnrtipton:
## Summary

Closes #1489 — finding **F3** of the v1.0.0 API-stability audit.

The decorators `optimistic`, `cache`, `client_state`, and `background` are stable public-API symbols (already in `djust.decorators.__all__`) but were only reachable via `from djust.decorators import …` — they were absent from the top-level `djust` package's `__all__`, unlike every other public decorator (`event_handler`, `action`, `computed`, …).

This PR adds the four to the top-level `djust` package's import block and `__all__`, so the canonical import path is now simply:

```python
from djust import optimistic, cache, client_state, background
```

The `from djust.decorators import …` path continues to work unchanged. The top-level names are the **same objects** as the `djust.decorators` originals — a pure re-export, not a redefinition. Purely additive and SemVer-safe.

## Changes

- `python/djust/__init__.py` — 4 names added to the `from .decorators import (…)` block and to `__all__`.
- `docs/API_STABILITY.md` — §F3 updated from "reachable only via `djust.decorators`" to ✅ resolved.
- `python/djust/tests/test_top_level_reexports_1489.py` — 4 regression tests: importability from `djust`, membership in `djust.__all__`, `is`-identity with the `djust.decorators` originals, and that `djust.decorators.__all__` is undisturbed.

This is the final task of the v1.0.0rc4 Phase-2 drain.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1537, opened 2026-05-19
fix(theming): build theme_head with a shared full-context builder (#1531)

### LINKED ISSUE #1531: ThemeMixin._setup_theme_context() renders theme_head.html with incomplete context — drops components.css link and emits invalid JS
## Description

`djust.theming.ThemeMixin._setup_theme_context()` builds `self.theme_head` via `render_to_string("djust_theming/theme_head.html", {...})` passing only `loading_class`, `css_block`, and `include_js`. The `theme_head.html` template (1.0.0rc3) expects more — notably `include_component_link`, `cookie_prefix_js`, and `direction`. With those missing:

1. `{% if include_component_link %}<link … components.css>{% endif %}` is falsy → the **`components.css` `<link>` is never emitted**. Any LiveView using `ThemeMixin` loads no component styles, so `theme_panel` / theme components render **unstyled** (same regression class as #1452).
2. `window.__djust_theme_cookie_prefix = {{ cookie_prefix_js }};` renders as `window.__djust_theme_cookie_prefix = ;` — a **JS syntax error** that breaks the whole anti-FOUC inline `<script>`.

The `{% theme_head %}` simple tag builds the full context, so `{{ theme_head }}` via the context processor is correct. Only the `ThemeMixin` path is affected.

## Steps to Reproduce

1. Create a `LiveView` using `ThemeMixin`; render `{{ theme_head }}` in `<head>`.
2. Load the page and view source.
3. Observe `window.__djust_theme_cookie_prefix = ;` (invalid JS) and no `<link>` to `djust_theming/css/components.css`.
4. Render a `theme_panel` in the view → it appears unstyled (no `.theme-panel*` rules).

## Environment

```
djust version:  1.0.0rc3
Python version: 3.12
Django version: 5.2.14
OS:             macOS
```

## Suggested fix

Route `_setup_theme_context()` through the `theme_head` simple tag, or pass the same full context the tag builds (`include_component_link`, `cookie_prefix_js`, `direction`, `deferred_css_block`, `component_css_block`). This mirrors the #1452 fix that pinned `{{ theme_head }}` ≡ `{% theme_head %}` for the context-processor path — the `ThemeMixin` path appears to have been missed.

Relevant files:

- `python/djust/theming/mixins.py` — `ThemeMixin._setup_theme_context()`
- `python/djust/theming/templates/djust_theming/theme_head.html`
- `python/djust/theming/templatetags/theme_tags.py` — `theme_head` tag (builds the correct context)

johnrtipton:
## Summary

Fixes #1531 — the 5th and final fix of the v1.0.0rc4 Phase-2 drain.

`ThemeMixin._setup_theme_context()` rendered `theme_head.html` with only 3 of the 8 context keys the template consumes, omitting `include_component_link`, `cookie_prefix_js`, `direction`, `deferred_css_block`, `component_css_block`. Two visible breakages:

1. The `{% if include_component_link %}` guard was falsy → the `<link>` to `djust_theming/css/components.css` was **never emitted** → theme components (`theme_panel`, …) rendered **unstyled** in any `ThemeMixin` view.
2. `window.__djust_theme_cookie_prefix = {{ cookie_prefix_js }};` rendered as `window.__djust_theme_cookie_prefix = ;` — a **JavaScript syntax error** that broke the anti-FOUC inline `<script>`.

The `{% theme_head %}` simple tag built the full context correctly, so `{{ theme_head }}` via the context processor was unaffected — only the `ThemeMixin` path was broken. This is the **#1452 context-drift bug repeated** on a third consumer of `theme_head.html`.

## Fix

Extract a shared `build_theme_head_context()` function — the `theme_head` tag **and** `ThemeMixin._setup_theme_context()` now build the head context from a single source of truth, so the two paths can no longer drift (preventing a #1452→#1531→… recurrence). The intentional `loading_class` divergence (tag `True`, mixin `False`) is preserved as an explicit builder parameter.

**Behavior change:** `ThemeMixin` views now also receive the same critical-CSS / deferred-CSS split that `{% theme_head %}` produces when `critical_css` is enabled — a consistency improvement, no migration needed.

## Tests

`TestThemeMixinThemeHead` (`python/djust/tests/test_theming_context_cache.py`) — 6 cases: `components.css` link present, valid `cookie_prefix_js`, a `ThemeMixin`-`theme_head`-≡-`{% theme_head %}` output-symmetry pin, `theme_css`⊆`theme_head` consistency, builder keyset invariant, builder `loading_class` param. Gate-off self-test (Action #1468/#254) confirmed: reverting the mixin to the 3-key context fails the 3 behavior-meaningful tests. Dogfooded against a real `ThemeManager`. Full suite green (Python 4376, JS 1600).


johnrtipton (commented):
Stage 11 review — verdict APPROVE (posted as comment; cannot self-approve).

**Checked:** stale-base (BEHIND=0), plan cross-ref, both call sites, builder correctness, JS escaping, test quality, full theming suite.

**Plan conformance:** `build_theme_head_context()` extracted (`theme_tags.py:44`); both the `theme_head` tag (`:195`) and `ThemeMixin._setup_theme_context()` (`mixins.py:108`) route through it.

**Correctness:**
- All 8 keys `theme_head.html` consumes are produced (`theme_tags.py:157-166`); pinned by `test_build_theme_head_context_keyset`.
- `request=None` path is safe — the builder never does `request.<attr>`; `request` is only forwarded to `get_theme_manager()` (signature `request: HttpRequest | None = None`). The mixin passes `manager=` so `get_theme_manager` is never reached. No crash path.
- `loading_class` divergence preserved as a builder param (`:48`); `test_build_theme_head_context_loading_class_param` pins tag=True/mixin=False.
- `self.theme_css` re-sourced from `head_ctx["css_block"]` (`mixins.py:118`); `test_mixin_theme_css_consistent_with_theme_head` pins it.
- `cookie_prefix_js` is `json.dumps()`-escaped (`:155`) — correct JS-string context; the fix produces a valid `"..."` literal, not raw interpolation.

**CHANGELOG:** documents the critical-CSS-split behavior change explicitly.

**Tests:** 6 `TestThemeMixinThemeHead` cases, non-tautological (symmetry pin asserts `==` between mixin/tag markers). 17 passed in the file; 1807 passed in `-k theming`.

🟡 Minor: positional `request` first arg called with a `None` literal at `mixins.py:109` reads slightly oddly — docstring covers it; non-blocking.

No 🔴. No auto-reject triggers (print/except:pass/f-string logging).

johnrtipton:
# Retrospective — PR #1537

**fix(theming): build theme_head with a shared full-context builder (#1531)**

Merged 2026-05-19 · +242 / -60 · 2 commits · Closes #1531

## Quality rating: 5 / 5

A textbook bugfix. Reproducer-first confirmation, correct root-cause
recognition (the #1452 drift bug recurring on a third consumer), and —
critically — the *structural* fix instead of the patch-the-symptom fix.
Stage 11 APPROVED with 0 🔴 and a single non-blocking 🟡 readability nit.

## What went well

- **Reproducer-first, confirmed failing.** `scratch/repro_1531.py` shipped
  3 tests, all confirmed FAILING on `main` before any fix code was
  written: missing `components.css` `<link>`, the
  `window.__djust_theme_cookie_prefix = ;` JS SyntaxError, and a
  mixin-vs-tag symmetry gap. The symptom was load-bearing evidence, not
  theory.

- **Recognized this as the #1452 bug repeated — and named it.** The plan
  did the diagnostic work to see that `ThemeMixin._setup_theme_context()`
  was the *third* consumer of `theme_head.html` (after the `theme_head`
  tag and the context processor that #1452 fixed). Same drift class, same
  symptom (unstyled theme panels), different file.

- **Chose the structural fix, not the symptom patch.** The cheap fix was
  to add the 5 missing keys to the mixin's hand-built dict. That would
  have worked — and left a fourth consumer free to drift the moment
  `theme_head.html` grows a 9th variable. Instead the PR extracted
  `build_theme_head_context()` as a single source of truth that both the
  `{% theme_head %}` tag and `ThemeMixin` now call. The two render paths
  *structurally cannot drift again*. The one intentional difference
  (`loading_class` — `True` for the tag's cold page load, `False` for the
  mixin's reactive mount) is preserved as an explicit builder parameter,
  so the *intended* divergence is visible and the *unintended* ones are
  eliminated.

- **Gate-off self-test (Action #254/#1468) confirmed.** Reverting the
  mixin to its 3-key context fails the 3 behavior-meaningful tests — the
  6 new `TestThemeMixinThemeHead` cases are non-tautological. The symmetry
  pin asserts `==` between mixin and tag markers, so a future 9th-variable
  template edit fails immediately on whichever path forgets it.

- **Dogfooded against a real `ThemeManager`.** Action #1060 satisfied; the
  critical-CSS / deferred-CSS split that `ThemeMixin` views now also
  receive was verified, not assumed.

- **Behavior change documented.** The CHANGELOG explicitly calls out that
  `ThemeMixin` views now get the same critical-CSS split as
  `{% theme_head %}` — a consistency improvement, no migration needed
  (Action #1313).

## What could be improved

**#1452 fixed the symptom path, not the root cause — and that is why
#1531 existed at all.** #1452 was itself a "two code paths drifted" bug:
the context processor hand-built a `theme_head` string that dropped the
`components.css` link. The #1452 fix routed the *context-processor* path
through the simple tag — but it never asked "how many code paths build a
`theme_head`?" A `grep` for every consumer of `theme_head.html` at the
time #1452 was fixed would have surfaced the `ThemeMixin` path
immediately, and #1531 would never have been filed.

The lesson — and this is exactly the #1079 broader-sweep discipline: when
fixing a "two code paths drifted" bug, **enumerate ALL paths up front**.
Then either converge them in the same PR (the genuine fix) or, if scope
forbids, file follow-up issues for the remainder so the drift surface is
tracked rather than rediscovered weeks later as a fresh `bug`-labelled
issue. #1452 did neither — it fixed the one path the reporter cited and
left two more (the tag was already correct; the mixin was silently
broken). PR #1537 is the cost of that omission: a full second pipeline
run for a bug that a 30-second grep would have closed alongside #1452.
Worth carrying forward as a process observation — "fix the path that's
broken" should escalate to "find every path that *could* break" whenever
the bug class is drift between parallel implementations.

The Stage 11 🟡 (positional `request` arg called with a `None` literal at
`mixins.py:109` reads slightly oddly) was correctly judged non-blocking,
not fixed, and not worth a commit — the docstring covers it.

## Review stats

| Metric | Value |
|---|---|
| Tests added | 6 (`TestThemeMixinThemeHead`) |
| 🔴 blocking | 0 |
| 🟡 minor | 1 (positional-arg readability — not fixed, not worth a commit) |
| CI failures | 0 |
| Full suite | green (Python 4376, JS 1600) |

## How this completes the v1.0.0rc4 Phase-2 drain

PR #1537 is the **fifth and final** fix of the v1.0.0rc4 Phase-2 drain.
All 5 fixes are now merged:

- **#1530** — #1529 VDOM `build_fragment_text_map` claim-once
- **#1532** — #1522 + #1523 a11y keyboard-nav + audit mode
- **#1535** — #1432 free-threading (`gil_used = false`)
- **#1536** — #1489 top-level re-exports
- **#1537** — #1531 ThemeMixin theme_head shared context builder

All 6 Phase-2 issues are closed: #1529, #1522, #1523, #1432, #1489, #1531.

**#1531 was a mid-drain discovery folded in after a scope check** — a
`bug`-labelled issue filed while the drain was already underway,
absorbed as a 5th fix rather than deferred. This is the same pattern as
**#1529**, which was likewise a mid-drain correctness-bug discovery folded
into rc4 (Task 1 of the drain). Two of the five Phase-2 fixes were not in
the original drain plan — a healthy sign that the drain process surfaces
real bugs as it runs, and that scope checks correctly absorbed them
rather than letting them slip to a later milestone. The v1.0.0rc4 Phase-2
drain is now COMPLETE.

PR #1539, opened 2026-05-19
fix(a11y): route arrow/Esc keys to a dropdown nested inside a dialog (#1533)

### LINKED ISSUE #1533: tech-debt: dropdown nested inside a role=dialog gets no arrow/Esc keyboard routing

johnrtipton:
## Summary

Closes #1533 — a tech-debt follow-up to #1522 (the keyboard-nav module), folded into the v1.0.0rc4 Phase-2 drain.

`51-keyboard-nav.js` routed every keydown inside an open `role="dialog"` through the dialog branch and returned early — so a `dropdown` component rendered **inside** a modal got no arrow-key roving navigation, and `Esc` always closed the whole dialog instead of the open dropdown.

## Fix

A surgical precedence refinement inside the dialog branch (the focus trap is untouched):

1. **`Tab` is checked first** and returns — the focus trap stays exactly as before; Tab can never fall through.
2. **`Esc`** — if the target is within a `.dj-dropdown` contained by the dialog *and that dropdown is open* (`data-open`), delegate to the dropdown handler to close it (and refocus its trigger); otherwise close the dialog as before.
3. **Arrow / Home / End** — when the target is within a dropdown contained by the dialog, delegate to the existing `_handleDropdown` (no new roving code — just reaching the existing handler).

Plain dropdowns and plain dialogs are unaffected. +60 bytes gzipped.

## Tests

9 new JSDOM cases in `tests/js/keyboard_nav.test.js` (`keyboard-nav — dropdown nested in dialog` block): nested-dropdown arrow roving, first-arrow-from-trigger, Home/End, Esc-closes-inner-dropdown-when-open, Esc-closes-dialog-when-dropdown-closed, Esc-closes-dialog-for-non-dropdown-target, **Tab-still-trapped-in-dialog**, plus plain-dropdown and plain-dialog regression guards. Gate-off self-test (Action #1468/#254) confirmed: restoring the original early-return fails the 4 behavior-meaningful cases. Full JS suite green (1609 tests).


johnrtipton (commented):
Stage 11 review — APPROVE.

Base current (BEHIND=0). 37/37 JS tests pass; bundle artifacts regenerated and in sync (`client.js` contains `dropdownInDialog`/`_dropdownOpen`, no working-tree drift).

**Focus trap intact (#1 risk — clear).** `Tab` is intercepted at `51-keyboard-nav.js:269-272` and returns before any dropdown routing — moved ABOVE the Esc/arrow branches. `Tab` can never reach `_handleDropdown`, which has no Tab handling anyway. Case 7 pins forward+Shift wrap; case 9 pins plain-dialog Tab.

**Esc precedence correct.** `:277` gates on `dropdownInDialog && _dropdownOpen(innerDropdown)`. `_dropdownOpen` uses `hasAttribute(data-open)`, matching both the bare (`djust_components.py:368`) and `="true"` (`:2203`) emission forms. Closed/absent dropdown falls through to `_closeModal` (cases 5, 6).

**Containment guard present** — `dialog.contains(innerDropdown)` at `:265-266`.

**Gate-off verified empirically.** Forcing `dropdownInDialog=false` in `client.js` fails exactly cases 1-4 (arrow roving + Esc-precedence), cases 5-9 pass — the implementers claim is accurate; the behavior-meaningful tests are non-tautological.

🟡 Minor: `_handleDropdown`s own Esc branch (`:215-219`) fires `toggle_dropdown` unconditionally — the open-state gate lives only in `_handleKeydown`. Correct here, but a future caller routing a closed dropdowns Esc into `_handleDropdown` would re-open it. Not blocking; the `_dropdownOpen` gate is the right guard at this call site.

No unguarded `console.*`. CHANGELOG entry accurate. Plain dropdown/dialog/tablist/accordion unaffected.

APPROVE.

johnrtipton:
## Retrospective — PR #1539

**Title:** fix(a11y): route arrow/Esc keys to a dropdown nested inside a dialog (#1533)
**Closes:** #1533 (tech-debt follow-up to #1522 / PR #1532)
**Stats:** +229 / -6, 2 commits, +60 bytes gzipped client bundle

### Quality rating: 5/5

A surgical, well-scoped a11y fix that resolves a real keyboard-dead-zone with the
minimum viable change. Stage 11 APPROVED with 0 🔴; the lone 🟡 was a hypothetical
future-caller note, no action required. The plan was thorough — it identified the
exact offending early-returns (`51-keyboard-nav.js:261` for arrows, `:256` for
Esc), proposed a precedence refinement rather than a branch reorder, and pre-armed
the gate-off self-test. Execution matched the plan exactly.

### What went well

- **Surgical Tab-first precedence.** The fix routes Arrow/Home/End and Esc to the
  nested dropdown *within* the dialog branch, while checking `Tab` first and
  returning before any dropdown routing. The focus trap (`_trapFocus`) is
  completely untouched — `Tab` can never fall through to `_handleDropdown`. This
  avoided the riskier alternative of reordering the top-level branches, which the
  plan correctly rejected (§2.5) because it would break the trap for dialogs
  containing a dropdown trigger.
- **Reused `_handleDropdown` instead of writing new roving code.** The dropdown
  branch already handles ArrowUp/Down/Home/End roving, first-arrow-from-trigger,
  and Esc-close-with-refocus. The fix only needed to *reach* it — the dialog
  branch now delegates rather than reimplementing. Net new JS was one small
  `_dropdownOpen` helper plus branch rework: +60 bytes gzipped, far under the
  per-module budget.
- **Gate-off-confirmed tests.** 9 JSDOM cases were added, and the gate-off
  self-test (Action #1468) was run: restoring the bare early-return fails 4 of 9
  cases — the arrow-roving and Esc-precedence cases — proving the tests exercise
  the actual change, not a tautology. The suite includes an explicit
  Tab-still-trapped regression guard (case 7) and plain-dropdown / plain-dialog
  regression guards, so the fix is fenced on every side.
- **Esc precedence mirrors existing nested-widget discipline.** Inner open
  dropdown consumes Esc first (refocusing its trigger), closed dropdown lets Esc
  fall through to the dialog — the same innermost-open-widget-first philosophy as
  the existing `_dialogStack`.

### What could be improved

**#1533 should not have existed — #1522 shipped this gap unflagged.** The nested
dropdown-in-dialog dead zone was a latent limitation in #1522's original
keyboard-nav module (PR #1532). It was only caught as a 🟡 review finding *on
#1532 itself*, then filed as follow-up tech debt. The root cause is that #1522's
test matrix covered each interactive widget (dialog, dropdown, tablist,
accordion) in isolation but never **composed two of them**. A "compose two
interactive widgets" test row — a dropdown inside a dialog, a tablist inside an
accordion, etc. — would have surfaced the unconditional dialog early-return
pre-merge instead of as a separate drain task. Recommendation: when shipping a
module with N delegated-branch handlers, the test matrix should include at least
the high-value N×N composition cells, not just the N diagonal cells. (Secondary,
lower-priority: the `test_chunks_overlap` wall-clock `max-not-sum` assertion is
inherently load-sensitive and flaked once under load here; it is genuine tech
debt worth its own issue, but the composition-gap is the more genuine lesson from
*this* PR.)

### Fit within the v1.0.0rc4 Phase-2 drain

#1539 was a mid-drain 🟡-finding follow-up — surfaced on PR #1532, then pulled
*into* rc4 (rather than deferred to v1.1.0) at the user's request alongside
#1534. Closing it inside rc4 keeps the pre-1.0 a11y surface clean rather than
shipping a known keyboard dead zone into the 1.0 release. With #1539 merged,
**#1534 (free-threaded hardening) is the last remaining Phase-2 task**.

### Review stats

- Tests added: 9 (JSDOM cases — nested roving, first-arrow-from-trigger,
  Home/End, Esc-precedence open vs closed vs non-dropdown, Tab-still-trapped,
  plain-dropdown + plain-dialog regression guards)
- 🔴 blocking findings: 0
- 🟡 non-blocking findings: 1 (hypothetical future-caller note on
  `_handleDropdown`'s unconditional `toggle_dropdown` — correct at this call
  site, no action)
- CI failures: 0 — the `test_chunks_overlap.py::test_total_wall_clock_is_max_not_sum`
  flake during Stage 6 was a confirmed load-sensitive timing flake (passed in
  isolation, 5/5 for the whole file); #1539 touches zero Python, so it could not
  be a regression.

Issue #1541, opened 2026-05-19
Audit actors/messages.rs for skip_serializing_if-without-default msgpack asymmetry


johnrtipton:
## Context

Follow-up from #1538 (fixed in PR for `fix-1538-vnode-msgpack-djust-id-default`).

#1538's root cause: `VNode.djust_id` (`crates/djust_vdom/src/lib.rs`) had `#[serde(skip_serializing_if = "Option::is_none")]` but **no `#[serde(default)]`**. Under MessagePack a plain `#[derive(Serialize, Deserialize)]` struct serializes as a *positional array*; a `None` trailing optional drops the element, and the derived sequence-visitor deserializer rejects the short array with `invalid length 5, expected struct VNode with 6 elements`. The #1538 fix added `default`.

## The sibling

During #1538's Stage-4 investigation, a sibling instance of the same `skip_serializing_if`-without-`default` pattern was found at:

- `crates/djust_live/src/actors/messages.rs:100` — `PatchResponse.patches: Option<Vec<Patch>>`
- `crates/djust_live/src/actors/messages.rs:103` — `PatchResponse.html: Option<String>`

`PatchResponse` is a plain `#[derive(Debug, Clone, Serialize, Deserialize)]` struct (no `#[serde(tag = ...)]`), so it serializes as a positional array under msgpack. If any of these fields is `None` *and* trailing in the array, and that struct is ever round-tripped through `rmp_serde`, it would hit the same `invalid length N` deserialize failure.

Per scope discipline (#1079), #1538's PR fixed **only** `VNode.djust_id` (the exact cited field). This issue tracks the sibling audit.

## Scope of this issue

1. Determine whether `PatchResponse` (and any other plain struct in `actors/messages.rs`) is actually round-tripped through `rmp_serde` (positional-array path) anywhere — actor message transport, persistence, etc. If it is only ever JSON-encoded (map path), there is no live bug, but adding `default` is still cheap insurance.
2. Add `#[serde(default)]` alongside `skip_serializing_if` on `PatchResponse.patches` and `PatchResponse.html` (and any sibling found).
3. Consider extending the #1448 wire-protocol snapshot suite (`crates/djust_vdom/tests/wire_protocol_snapshot.rs`) with `rmp_serde` round-trip coverage for **all** plain wire structs — currently only `VNode` has it (added in #1538's PR). The suite's `serde_json`-only coverage is exactly the gap that let #1538 through.

## Note — not affected

[...]

PR #1540, opened 2026-05-19
feat: free-threaded hardening — dead code, frozen pyclasses, RwLock registries, 3.14t CI (#1534)

johnrtipton:
## Summary

Closes #1534 — a 4-item free-threaded hardening bucket deferred from #1432's scope (Action #1079 broader-sweep discipline). The final task of the v1.0.0rc4 Phase-2 drain. All four are **internal hardening — no public API or behavior change for application code.**

## The 4 items

1. **`COMPONENT_REGISTRY` removed (dead code).** The Rust-side `COMPONENT_REGISTRY` + its 3 accessors (`register_component`/`get_component`/`unregister_component`) in `crates/djust_components` had zero call sites, were never exported through `djust._rust`, and `get_component` was an `unimplemented!()` stub. Deleted (25 lines). The `Component` trait already declared `Send + Sync` — the issue's premise that it didn't was stale. The Python-side component registry (`python/djust/components/registry.py`) is unrelated and untouched.

2. **`#[pyclass(frozen)]`** on `SupervisorStatsPy` (immutable POD) and `SessionActorHandlePy` (all-`&self`, holds only a clonable `mpsc::Sender`). `frozen` drops PyO3's per-instance runtime borrow-check overhead — both were verified to have no `&mut self` method or interior mutability.

3. **`Mutex` → `RwLock`** on the 4 template registries (tag / block / assign / filter). 27 lock sites converted — 12 `.write()` (register/unregister/clear — one-time bootstrap), 15 `.read()` (dispatch/lookup). Concurrent renders on a free-threaded interpreter now share the read lock instead of serializing on registry lookups.

4. **`python3.14t` CI leg.** A new non-blocking job (`continue-on-error: true`, not wired into the merge gate) runs the free-threaded `threading` smoke test on a genuine free-threaded interpreter via `uv python install 3.14t` — where `test_module_does_not_reenable_gil_on_free_threaded_build` (previously `skipif`-guarded) becomes a real assertion.

## Tests

New `rwlock_registry_allows_simultaneous_readers` case in `crates/djust_templates/tests/free_threaded_safety.rs` — 12 threads hold a read guard concurrently. Gate-off self-test (Action #1468/#254): a `Mutex`-replica of the test **deadlocks to the deadline** (only 9/12 readers ever in-section) — the test genuinely proves the `RwLock` read-sharing. Full suite green: Python 4376, JS 1609, `cargo test` all crates clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1542, opened 2026-05-19
fix(vdom): add serde(default) to VNode.djust_id — msgpack round-trip (#1538)

### LINKED ISSUE #1538: VNode msgpack round-trip fails: `djust_id` uses `skip_serializing_if` without `serde(default)`
## Summary

`VNode` msgpack round-trips fail whenever any node in the tree has
`djust_id == None`. The field uses `#[serde(skip_serializing_if = "Option::is_none")]`
but has no `#[serde(default)]`, so serialization emits a 5-element sequence
that the derived `Deserialize` (which expects 6) rejects.

## Observed

In an on-device iOS build of a djust app (djust `1.0.0rc3`), every WebSocket
resume logs:

```
ERROR djust.state_backends.memory: InMemoryStateBackend.get: serialize/deserialize
  round-trip failed for key '..._liveview_/max/_...'; entry discarded — caller should remount
Traceback (most recent call last):
  File ".../djust/state_backends/memory.py", line 119, in get
    clone = RustLiveView.deserialize_msgpack(serialized)
ValueError: MessagePack deserialization error: invalid length 5, expected struct VNode with 6 elements
```

The view still mounts — `InMemoryStateBackend` discards the cache entry and the
caller remounts — so it is degraded-but-functional. But the error is logged on
every resume, and cross-reconnect state continuity is lost for affected trees.

## Root cause

`crates/djust_vdom/src/lib.rs:118`:

```rust
#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
pub struct VNode {
    pub tag: String,
    pub attrs: HashMap<String, String>,
    pub children: Vec<VNode>,
    pub text: Option<String>,
    pub key: Option<String>,
    #[serde(skip_serializing_if = "Option::is_none")]
    pub djust_id: Option<String>,
    #[serde(skip)]
    pub cached_html: Option<String>,
}
```

`VNode` has 6 serializable fields (`cached_html` is `#[serde(skip)]`). The
compact msgpack path serializes a struct as a positional array. `djust_id` is
`skip_serializing_if = "Option::is_none"`, so when it is `None` the array has
only **5** elements. The derived `Deserialize` for a struct read from a
sequence expects exactly 6 elements unless the trailing field has a default —
hence `invalid length 5, expected struct VNode with 6 elements`.

Whether the bug triggers depends on whether the serialized tree contains a
`djust_id`-less node, which is content/path dependent — it did not surface on a
dev-server build of the same app, only on the iOS build.

## Suggested fix

Add `#[serde(default)]` to `djust_id`:

```rust
    #[serde(default, skip_serializing_if = "Option::is_none")]
    pub djust_id: Option<String>,
```

With `default`, the derived sequence deserializer fills the missing trailing
element with `None` instead of erroring. As a rule, a `skip_serializing_if`
field that can change the serialized shape should also carry `#[serde(default)]`
when the format deserializes structs positionally.

A round-trip regression test over a `VNode` whose `djust_id` is `None` (and a
`RustLiveView` containing such a node) would pin this.

## Environment

- djust `1.0.0rc3`, iOS-cross-compiled wheel
- Surfaces via `RustLiveView.deserialize_msgpack` in `InMemoryStateBackend.get`
- Reproduces on every WebSocket resume of the affected views

johnrtipton:
## Summary

Fixes #1538 — a wire-protocol correctness bug, folded into the v1.0.0rc4 Phase-2 drain as a 6th fix.

`VNode.djust_id` (`crates/djust_vdom/src/lib.rs`) carried `#[serde(skip_serializing_if = "Option::is_none")]` but no `#[serde(default)]`. Under msgpack a struct serializes as a **positional array** — so a `None` `djust_id` dropped the trailing element and produced a **5**-element array, which the derived 6-element deserializer rejected: `invalid length 5, expected struct VNode with 6 elements`.

The HTML parser assigns `djust_id = None` to every text node, so any view whose VDOM tree contained text hit this: `RustLiveView.deserialize_msgpack` failed inside `InMemoryStateBackend.get` / `RedisStateBackend`, the cached state entry was discarded, an error was logged on **every WebSocket resume**, and cross-reconnect state continuity was lost for the affected view.

## Fix

Add `#[serde(default)]` so the sequence deserializer fills a missing trailing element with `None`:
```rust
#[serde(default, skip_serializing_if = "Option::is_none")]
pub djust_id: Option<String>,
```
Deserialize-only — **serialized bytes are byte-identical**, and a new deserializer still reads old 6-element payloads, so there is no wire-format migration.

## Tests

The #1448 wire-protocol snapshot suite tested only the JSON (named-map) encoding — which is exactly why it missed this msgpack-positional bug. It now also has `rmp_serde` round-trip coverage:

- `crates/djust_vdom/tests/wire_protocol_snapshot.rs` — 3 `msgpack_round_trip_*` cases (`djust_id` None / Some / nested-mixed-tree).
- `python/tests/test_serialization_hardening.py` — 2 `TestVNodeMsgpackRoundTrip` cases through `RustLiveView.serialize_msgpack`/`deserialize_msgpack`.

Gate-off self-test (Action #1468/#254) confirmed: reverting `default` fails the `djust_id`-None cases with the exact length error. `cargo test -p djust_vdom` fully green; full suite green.

A sibling `skip_serializing_if`-without-`default` asymmetry in `crates/djust_live/src/actors/messages.rs` is filed as follow-up #1541 (scope discipline — #1079).


johnrtipton (commented):
Stage 11 review — APPROVE

Verified env: BEHIND=0 vs origin/main. Tests: cargo test -p djust_vdom 19 passed (16 existing JSON snapshots + 3 new msgpack); pytest test_serialization_hardening.py 29 passed (2 new).

Trailing-field requirement (the load-bearing precondition): djust_id is genuinely the LAST #[serde]-visible field of VNode (lib.rs:133) — cached_html (lib.rs:136) is #[serde(skip)], fully absent from the wire form, not just skip-if-none. So serde(default) really does rescue the missing trailing positional element; if a non-trailing field were skipped this fix would not work, but that case does not exist here.

Byte-identical claim: confirmed — the 16 JSON literal-byte snapshot tests stay green, evidencing serde(default) is deserialize-only. Back-compat both directions covered: msgpack_round_trip_vnode_djust_id_some exercises the 6-element (old/present) payload, _none the 5-element. Nested mixed-tree test is meaningful — parent Some, interleaved text children None — and asserts per-child djust_id post-restore.

#1448 gap genuinely closed: new tests round-trip through real rmp_serde::to_vec/from_slice and RustLiveView.serialize_msgpack, not a hand-rolled encode.

Gate-off self-test (#1468): removed serde(default), re-ran — _djust_id_none and _nested_tree_mixed fail with the exact invalid length 5, expected struct VNode with 6 elements; _some control passes. Non-tautological. Source restored clean.

No auto-reject triggers. CHANGELOG entry accurate.

Non-blocking: actors/messages.rs:100,103 carries the same skip_serializing_if-without-default shape — plan §5 correctly scopes it out (#1079); please confirm the recommended follow-up issue is filed.

APPROVE.

johnrtipton:
# Retrospective — PR #1542

**Title:** fix(vdom): add serde(default) to VNode.djust_id — msgpack round-trip (#1538)
**Closes:** #1538
**Merged:** 2026-05-19
**Diff:** +163 / -1, 2 commits

## Quality rating: 5/5

A textbook wire-protocol bugfix: reproducer-first confirmation of the exact
deserializer error, a one-line fix with verified byte-identical wire form and
both-directions back-compat, the symptom-and-the-test-gap closed in the same
PR, and the sibling asymmetry filed rather than scope-crept. Stage 11 APPROVED
with 0 🔴 / 0 🟡; gate-off self-test confirmed non-tautological.

## What went well

- **Reproducer-first surfaced the exact failure.** The fix was driven by the
  precise deserializer error — `invalid length 5, expected struct VNode with 6
  elements` — not by inspection. msgpack serializes a plain struct as a
  positional array; `#[serde(skip_serializing_if = "Option::is_none")]` on the
  trailing `djust_id` field drops the 5th element when `None`, and the derived
  sequence-visitor `Deserialize` impl rejects the short array. Because the HTML
  parser gives every text node `djust_id = None`, any view containing text
  broke `RustLiveView.deserialize_msgpack` inside the state backends — error
  logged, cross-reconnect state continuity silently lost.

- **One-line fix, fully verified.** Adding `#[serde(default)]` is
  deserialize-only: serialized bytes are byte-identical (all 16 existing JSON
  snapshot tests stayed green), and old 5-element payloads now deserialize with
  `djust_id = None`. Back-compat is covered in both directions — `_some`
  exercises the 6-element form, `_none` the 5-element form. Stage 11
  independently verified `djust_id` is the last `#[serde]`-visible field
  (`cached_html` is `#[serde(skip)]`), so `default` genuinely rescues the
  missing trailing element.

- **Closed the #1448 snapshot-suite gap in the same PR.** Rather than just
  fixing the symptom, the PR added 3 `rmp_serde` round-trip tests to
  `wire_protocol_snapshot.rs` — exactly the encoding the suite never exercised.
  The symptom and the reason the suite missed it are both addressed in one
  changeset.

- **Scope discipline (#1079).** A sibling `skip_serializing_if`-without-`default`
  asymmetry at `actors/messages.rs:100,103` was filed as follow-up #1541 and
  left untouched here. The fix stayed scoped to exactly what #1538 cited.

## What could be improved

- **A pinning/snapshot suite must exercise *every* encoding the wire protocol
  actually uses.** The #1448 wire-protocol snapshot suite was a whole milestone
  of work, built specifically to pin this serde-asymmetry class — yet it only
  exercised `serde_json` (named-map encoding). djust's wire protocol uses
  *both* JSON and msgpack (`rmp_serde`, positional-array encoding), and a
  msgpack-only positional-array bug sailed straight through 16 green snapshot
  tests. Single-encoding coverage gave false confidence: the suite's existence
  made the class *look* covered while the failure mode lived entirely in the
  untested encoding. **Canonicalizable lesson:** any snapshot/pinning suite for
  a multi-encoding wire protocol must round-trip through *each* encoding the
  protocol actually uses on the wire — JSON-only coverage of a JSON+msgpack
  protocol is not coverage. This generalizes #1448 (wire-protocol JSON pinning
  as a standard test class) to "pin every encoding, not just one."

## How this closes the v1.0.0rc4 Phase-2 drain

#1538 is the 6th and final Phase-2 fix. The drain now spans 8 PRs closing 9
issues:

| PR | Issue |
|----|-------|
| #1530 | #1529 |
| #1532 | #1522 |
| #1535 | #1432 |
| #1536 | #1489 |
| #1537 | #1531 |
| #1539 | #1533 |
| #1540 | #1534 |
| #1542 | #1538 |

(#1530 also covered #1523; the drain closed #1529, #1522, #1523, #1432, #1489,
#1531, #1533, #1534, #1538.)

Only #1434 (native async ORM) and the just-filed follow-up #1541 remain — and
#1434 is correctly parked in v1.1.0. The v1.0.0rc4 Phase-2 drain is complete.

## Recurring pattern: mid-drain correctness bugs from real downstream usage

Three correctness bugs — #1529, #1531, and now #1538 — surfaced *mid-drain*
from real ad-hoc downstream usage. #1538 specifically was caught by an
on-device iOS build. The pre-1.0 backlog held genuine latent bugs that only
real usage kept surfacing: each was filed mid-drain and folded into rc4 at the
user's request rather than waiting for a separate cycle. The signal: ad-hoc
real-world exercise of the framework is finding bugs that the test suites —
even purpose-built ones like #1448 — were not.

## Review stats

- Tests added: **5** (3 Rust `rmp_serde` round-trip in
  `wire_protocol_snapshot.rs`; 2 Python `RustLiveView` serialize→deserialize
  msgpack round-trip in `test_serialization_hardening.py`)
- 🔴 blocking findings: **0**
- 🟡 non-blocking findings: **0**
- CI failures: **0**
- Gate-off self-test: confirmed — removing `serde(default)` makes the
  `_none` and `_nested_tree_mixed` tests fail with the exact `invalid length 5`
  error; the `_some` control still passes (non-tautological).

Issue #1543, opened 2026-05-19
tech-debt: djust_live cannot be cargo-tested — gate the extension-module feature behind a Cargo flag


johnrtipton:
Source: v1.0.0rc4 milestone retro (recurred in PR #1530 and PR #1535).

## Problem

`crates/djust_live` carries the PyO3 `extension-module` feature *unconditionally* (`pyo3 = { workspace = true, features = ["extension-module"] }`). `extension-module` tells PyO3 not to link libpython — the symbols come from the host interpreter at load time. Under `cargo test` there is no host interpreter, so the link fails (`ld: symbol(s) not found ... Py_True`). `make test` works around this with `cargo test --workspace --exclude djust_live`.

Consequence: the crate that holds the `djust._rust` entry point — and a lot of `_rust`-surface logic (`build_fragment_text_map`, `collect_vdom_text_nodes`, the actor system, the `RustLiveView` backend) — has **no fast Rust-native test feedback loop**. Every regression for `djust_live`-internal logic must route through the Python layer via PyO3.

## Why now

This surfaced **twice in the v1.0.0rc4 Phase-2 drain**:
- PR #1530 (#1529 `build_fragment_text_map` fix) — the bug lived in `djust_live`; the regression test had to go to the Python layer.
- PR #1535 (#1432 free-threaded-safe) — the `std::thread` concurrency tests had to be placed in sibling crates (`djust_templates`, `djust_vdom`) because `djust_live` itself can't host a `cargo test`.

Two occurrences in one drain make it a standing structural constraint, not a one-off.

## Suggested fix

Gate `extension-module` behind a Cargo feature that is **on by default** (so `maturin` / `make build` are unchanged) but can be turned **off for `cargo test`**:

```toml
[features]
default = ["extension-module"]
extension-module = ["pyo3/extension-module"]
```

Then `cargo test -p djust_live --no-default-features` links against libpython and runs. Verify `make build` (maturin, default features) is unaffected, and update the `make test` Rust path (currently `--exclude djust_live`) accordingly.

## Out of scope

Not a correctness bug — purely a developer-feedback-loop / test-infrastructure improvement. The sibling-crate placement used in #1530/#1535 is a sound workaround in the meantime.

Issue #1456, opened 2026-05-12
tech-debt: extend wire-protocol snapshot pinning to remaining ~22 frame shapes (#1448 follow-up)


johnrtipton:
## Source

PR #14XX (closes #1448 starter) pinned 8 of ~30+ wire-protocol frame shapes:
`push_event`, `flash`, `page_metadata`, `patch`, `mount`, `layout`, `navigation`, `error`.

This issue tracks the remaining shapes.

## Shapes to pin

**websocket.py outbound** (most of the surface):
- `mount_batch` (line 2610)
- `child_update` (line 656)
- `sticky_update` (line 688)
- `sticky_hold` (line 2400)
- `i18n` (line 731)
- `accessibility` (line 751)
- `focus` (line 766)
- `embedded_update` (line 3260)
- `upload_registered` (line 3548)
- `upload_resumed` (line 3568)
- `upload_progress` (line 3661)
- `reload` (lines 3848, 3885, 3921, 3935, 3960, 3969 — multiple variants)
- `hvr-applied` (line 3856)
- `html_update` (line 1126)
- `connect` (line 1425)
- `rate_limit_exceeded` (line 1594)
- `pong` (line 1607)
- `navigate` (lines 1959, 1983 — variants)
- `noop` (line 624)
- `error` with `message` (line 1953, vs `error` field at 794 — two variants)

**streaming.py / http_streaming.py**:
- `streaming.patch` (line 279)
- `streaming.html_update` (line 289)
- `streaming.stream` (line 359)

**presence.py**:
- `presence_event` (line 348)

**Inbound shapes** (the client → server side of the contract):
- `mount`, `event`, `mount_batch` (consumer side)
- `upload_resume`, `upload_chunk`, etc.

## Pattern to follow

`python/djust/tests/test_wire_protocol_snapshots.py` (from #1448 starter PR).

Each test:
1. Reads the emit-site line(s) in `websocket.py` / `streaming.py` / `presence.py`.
2. Constructs the frame dict with the EXACT same shape (key order, defaults, conditional appends).
3. `json.dumps(frame)` and asserts against a literal `expected_json` string.

## Sequencing

These can ship in 2-3 grouped PRs:
- **Batch 1**: lifecycle frames (`mount_batch`, `child_update`, `sticky_update`, `sticky_hold`, `embedded_update`).
- **Batch 2**: optional features (`i18n`, `accessibility`, `focus`, `html_update`, `connect`).
[...]

PR #1546, opened 2026-05-19
fix(actors): remove skip_serializing_if on PatchResponse — leading-optional msgpack round-trip (#1541)

### LINKED ISSUE #1541: Audit actors/messages.rs for skip_serializing_if-without-default msgpack asymmetry

johnrtipton:
Closes #1541. Sibling-audit follow-up to #1538 (PR #1542).

## TL;DR

The Stage 4 plan called for mirroring PR #1542's #1538 fix verbatim (add `#[serde(default)]` alongside `skip_serializing_if`). **Stage 5 reproducer-first TDD found that fix does not generalize** — it only works for STRICTLY TRAILING optionals. `PatchResponse.patches` and `PatchResponse.html` are LEADING optionals, so the empirically-correct fix is to **remove `skip_serializing_if` entirely**. Cost: 1 extra byte per `None` field (msgpack `nil`).

This was caught by the Stage 5 mandatory "VERIFY ARTIFACT BEFORE PLANNING" gate (#1210) before any code shipped against the wrong fix.

## Why `default` doesn't fix leading-optional structs

Under msgpack, a plain `#[derive(Serialize, Deserialize)]` struct serializes as a *positional array*. For `PatchResponse { patches: None, html: None, version: 7 }` with `skip_serializing_if` on the two optionals, the wire bytes are `[91, 07]` — a 1-element array containing only `version`. On deserialize the sequence visitor reads `version=7` into the FIRST slot (`patches: Option<Vec<u8>>`) and errors with `invalid type: integer 7, expected a sequence`. `#[serde(default)]` doesn't help: the deserializer isn't running out of elements, it's reading wrong-typed values at the wrong positional slots.

A 4-by-3 empirical probe (3 candidate fixes × 4 None/Some combinations) confirmed: only removing `skip_serializing_if` outright fixes all cases. See the new structural witness `msgpack_skip_with_default_works_for_trailing_optional_only` in `crates/djust_vdom/tests/wire_protocol_snapshot.rs` for the proof.

## Changes

| File | Change |
|---|---|
| `crates/djust_live/src/actors/messages.rs` | Remove `skip_serializing_if` on `PatchResponse.patches` and `.html`; doc-comment the rationale citing #1538/#1541; add 4 inline msgpack round-trip tests (compile-checked today; will run once #1543 lands); update existing JSON test to assert `"html":null` / `"patches":[]` always present. |
| `crates/djust_vdom/tests/wire_protocol_snapshot.rs` | New `#1541 bug-class` section: 3 structural witness tests (skip-without-default fails; skip-with-default works for trailing optionals ONLY; no-skip round-trips in all positions). |
| `CHANGELOG.md` | Fixed entry under `[Unreleased]`. |

Total: 297 insertions, 4 deletions across 2 source files + 4-line CHANGELOG entry.

## Verification

- `cargo test --workspace --exclude djust_live`: **818 passed, 0 failed** (3 new msgpack witness tests + 6 existing #1538 msgpack tests all green).
- `cargo check -p djust_live --tests`: clean (Layer B compile-checks; can't link/run today due to #1543 — same constraint #1538's drain hit).
- `python -m pytest python/tests/test_serialization_hardening.py`: 29/29 passed.
- `python -m pytest python/tests/test_timing_metadata_gating.py`: 9/9 passed.

## Wire-format impact

- **msgpack:** `PatchResponse` is not currently round-tripped through `rmp_serde` on any production path (only the inner `Vec<Patch>` is at `crates/djust_live/src/lib.rs:679`). This is defense-in-depth against future cross-process actor transport.
- **JSON:** the encoding now always includes the `patches` and `html` keys (`null` rather than omitted). No current consumer parses `PatchResponse` JSON — verified by grep across `python/`, `tests/`, and `crates/`.

## Two-commit shape (#1173)

- `26a898c3` — implementation + tests (no CHANGELOG.md)
- `eed3cbae` — CHANGELOG only

## Out of scope

- **#1543** (gate `djust_live`'s `extension-module` behind a Cargo flag so `cargo test -p djust_live` works) — tracked separately, processed next in the v1.0.0rc6 drain.
- **#1456** — extending the wire-protocol snapshot suite with `rmp_serde` coverage for the remaining ~22 frame shapes. This PR pins only the bug class structurally, not full coverage.

## Process notes for the retro

This PR exercised two existing canon rules visibly:

1. **#1210 (VERIFY ARTIFACT BEFORE PLANNING)** — the Stage 4 reproducer caught the plan/reality mismatch in Stage 5, before any incorrect production code shipped.
2. **#1056 (smallest design-novel iter first)** — processing #1541 first in the rc6 drain locked the fix-shape understanding before #1543's Cargo-feature work would have otherwise re-confused things.

It also surfaces a candidate new rule worth retro consideration: **a #1538-class fix doesn't automatically generalize to siblings without verifying the field POSITION (leading vs trailing) matches.** PR #1542's fix shape is correct ONLY for trailing-optional shapes like `VNode.djust_id`.


johnrtipton (commented):

### Pre-review gates
- **Stale-base (#1450)**: ✅ `git rev-list --count HEAD..origin/main` = 0. Reviewing the program the merge will apply.
- **Two-commit shape (#1173)**: ✅ `26a898c3` touches only Rust files (`messages.rs`, `wire_protocol_snapshot.rs`); `eed3cbae` touches only `CHANGELOG.md` (1 file, +4 lines). Clean separation.
- **Synthetic-bug-trigger canary (#1448, #252)**: ✅ The Layer A `msgpack_skip_without_default_fails` test at `wire_protocol_snapshot.rs:351` is itself the empirical canary — it constructs Shape A (trailing) and Shape B (leading) witnesses and asserts both `from_slice` calls return `Err`. The counter-example `msgpack_skip_with_default_works_for_trailing_optional_only` (line 405) proves the #1538 fix shape only works for trailing optionals, with both an `assert_eq!` success branch (Shape A) and an `is_err()` branch (Shape B). The bug class is structurally pinned independent of `PatchResponse` itself — excellent dep-graph hygiene given `djust_live → djust_vdom`.

### Correctness findings
- 🟢 The annotation removal at `crates/djust_live/src/actors/messages.rs:118-122` is the minimum change. The accompanying doc comment (lines 96-113) cites `msgpack_skip_with_default_works_for_trailing_optional_only` — grep-verified it exists at line 405 of the cited file (no #1197-class citation drift).
- 🟢 The 4 inline Layer B tests (`messages.rs:347-409`) assert PASSING round-trips post-fix, with each `expect(...)` message naming #1541. The `both_some` case is correctly framed as a control — it's the only path that worked pre-fix.
- 🟢 `LeadingOptionalsWithDefault` (`wire_protocol_snapshot.rs:434-455`) correctly asserts `is_err()` — the counter-example that proves `default` doesn't generalize. Name + assertion match.
- 🟢 `msgpack_no_skip_round_trips_in_all_positions` iterates all 4 None/Some combos via a `for v in [...]` loop — full quadrant coverage of the fix surface.
- 🟢 Scope discipline (#1079): `MountResponse` (line 78), `RenderResult`, and other structs in `messages.rs` are untouched. Pure surgical fix to the cited fields.
- 🟢 `lib.rs:679` site verified: it's `rmp_serde::to_vec(&patches)` (inner `Vec<Patch>`), not `&response` — the "defense-in-depth" framing in the doc comment is accurate.
- 🟢 Python consumer grep: zero matches in `python/` parsing `PatchResponse` JSON (only test class name + binary symbol-table noise from a `dylib.dSYM`). Wire-format break is genuinely no-op for current consumers.
- 🟢 Existing `test_patch_response_with_patches` (lines 277-303 of `messages.rs`) updated: old `!json.contains("\"html\"")` flipped to `json.contains("\"html\":null")` + `json.contains("\"patches\":[]")` + `json.contains("\"version\":3")`. Assertions match the new always-present-key shape.

### Edge cases checked
1. **Future contributor adds another optional field after `version`**: the doc comment says "intentionally NOT marked `skip_serializing_if`" but doesn't explicitly extend the rule to *future* fields. A future trailing-optional could legitimately use `#[serde(default, skip_serializing_if = ...)]` (the #1538 shape), but a future leading-optional would re-introduce #1541. 🟡 **Minor**: could the doc comment add a one-liner like "any new optional field added BEFORE `version` must also omit `skip_serializing_if`; trailing optionals may use the #1538 `default`-paired shape"? Not a blocker — the cross-referenced witness test will fail anyway if someone does this wrong, just slower to root-cause.
2. **Someone reintroduces `skip_serializing_if`**: confirmed — Layer A tests don't reference `PatchResponse` so they wouldn't catch it; Layer B tests would catch it once #1543 lands and unblocks `cargo test -p djust_live`. PR body's reference to #1543 as the test-runnability blocker is correctly disclosed.
3. **Gate-the-change-off self-test (#254)**: Layer A witnesses are SELF-CONTAINED structs (`TrailingOptional`, `LeadingOptionals`, `LeadingOptionalsWithDefault`, `NoSkip`) — gating off the `messages.rs` annotation change has NO effect on Layer A tests, exactly as the PR description claims. Layer B tests would flip on gate-off (they'd start failing with "invalid length"/"invalid type" errors). Both layers do what they advertise.
4. **`.expect()`/`panic` pattern in tests**: idiomatic for `#[test]` functions where failure should produce a clear named-message panic. Each `expect(...)` cites #1541. Fine.

### Process / canon
- 🟢 CHANGELOG entry (`CHANGELOG.md:10`) matches the rc4 `### Fixed` style — dense paragraph, file:line cites (`crates/djust_live/src/actors/messages.rs`, `lib.rs:679`, `wire_protocol_snapshot.rs`), test counts (3 + 4), explicit linkage to #1538/#1542/#1456/#1448/#1543.
- 🟢 The Stage 5 plan/reality divergence (planner said "add `default`", implementer's TDD found "remove `skip_serializing_if`") is surfaced prominently in both the commit message body and the CHANGELOG paragraph. This is the most valuable artifact of this PR — preserves the empirical finding for future contributors who'll be tempted to mirror the #1538 fix shape blindly.

### Question / concern
- **One minor doc-clarity nit (🟡)**: The doc block at `messages.rs:96-113` reads naturally for *this* fix but could grow a forward-looking sentence on what shape a *future* added optional field should use. Suggested wording: "If you add another optional field to this struct, leading-position fields must NOT use `skip_serializing_if`; trailing-position fields may use the #1538 `#[serde(default, skip_serializing_if = "Option::is_none")]` paired shape." Non-blocking — the existing witness tests catch the failure mode regardless, just slower.

### Verdict
**APPROVE.** Surgical fix with strong empirical TDD. The Layer A structural witnesses are excellent dep-graph hygiene; the Layer B inline tests cover the right quadrant; the doc-comment + CHANGELOG preserve the planner-vs-reality finding clearly. Two-commit shape is clean. Citations check out. Defense-in-depth framing is accurate (no current msgpack path exercises `PatchResponse`, but the bug class is now pinned in shared witness tests).

Optional follow-up only: consider amending the `PatchResponse` doc comment to include the forward-looking guidance noted above — but not blocking for merge.

johnrtipton:
## Stage 14 retro — PR #1546 (#1541)

### What went well

- **Reproducer-first TDD caught the planner's wrong fix shape in Stage 5, not Stage 11.** The Stage 4 plan called for mirroring PR #1542's `default + skip_serializing_if` fix verbatim. Stage 5's mandatory "VERIFY ARTIFACT BEFORE PLANNING" gate (Action #1210) had the implementer write the failing reproducer first — the reproducer revealed (4-way empirical probe: 3 candidate fixes × 4 None/Some combos) that `default` only works for STRICTLY TRAILING optionals. Without the reproducer-first gate this PR would have shipped a no-op patch that compiled cleanly and passed CI but did not fix the bug class for leading-optional `PatchResponse`. **The discipline saved the PR.**
- **Two-layer test strategy worked.** Layer A (structural witness in `djust_vdom/tests/wire_protocol_snapshot.rs`, 3 self-contained tests) runs in CI today and pins the bug class. Layer B (4 inline tests in `messages.rs`) compile-checks today and will execute once #1543 lands. The dep-graph hygiene (`djust_live → djust_vdom`, never the reverse) was respected.
- **Stage 11 reviewer ran the empirical canary (Action #252).** Confirmed the Layer A witness tests actually exercise the bug class they claim to, and grep-verified the doc-comment's cited test name exists at the cited location (Action #1197 — caught a citation drift pre-commit, fixed before the review).
- **Two-commit shape preserved** (Action #1173): `26a898c3` (impl, no CHANGELOG) + `eed3cbae` (docs, CHANGELOG only).
- **Stale-base gate (Action #1450) ran clean** — branch was 0 commits behind `origin/main` at review time.

### What could be improved

- **The Stage 4 plan should have flagged the leading-vs-trailing distinction.** The plan said "mirror PR #1542 byte-for-byte" without acknowledging that PR #1542's fix worked because `VNode.djust_id` was a TRAILING optional. The PR #1542 retro could/should have canonized "the #1538 fix shape does not generalize across struct field positions" as a process note. **Candidate canon rule (proposed for the v1.0.0rc6 milestone retro):** *When mirroring a serde fix from one struct to another, verify the field POSITION (leading vs trailing) matches before assuming the fix generalizes.* For msgpack-positional encoding, `skip_serializing_if` is only safe on strictly trailing optionals; for other positions, remove `skip_serializing_if` entirely (cost: 1 byte per `None` field).
- **One 🟡 non-blocking finding deferred to a future PR** (reviewer's nit): the `PatchResponse` doc block could grow a forward-looking sentence about which shape future-added optional fields should use. Captured here for retro; not worth a follow-up issue.

This is task 1 of 3 in the v1.0.0rc6 drain. The next two are #1543 (gate `djust_live`'s `extension-module` behind a Cargo flag — would unblock Layer B execution for this PR retroactively) and #1545 (Python `LiveView.request` snapshot warning — unrelated subsystem).

The candidate-canon rule above is the most milestone-level artifact and will be reconciled into the v1.0.0rc6 milestone retro after all three drain PRs land.

### Verdict
`RETRO_COMPLETE`

Issue #1545, opened 2026-05-19
LiveView.request triggers "non-serializable value: ASGIRequest" warning on every state snapshot


johnrtipton:
## Summary

Every `LiveView` mount/event logs:

```
WARNING:djust.serialization:LiveView state contains non-serializable value: ASGIRequest (from django.core.handlers.asgi). This will be converted to a string, which may cause AttributeError on deserialization. Consider using self._<attr> for private state, or re-initialize in mount()/event handlers.
```

The warning is emitted by the non-serializable fallback in `python/djust/serialization.py:557` when the state snapshot encounters `self.request`.

## Root cause

`self.request` is assigned as a **public** attribute *after* `__init__`, so the snapshot machinery treats it as user state and tries to serialize it. Assignment sites:

- `python/djust/mixins/request.py:489` — HTTP `post()` path: `self.request = request` (commented "for context processors and `_sync_state_to_rust` csrf_token injection (#705)")
- `python/djust/websocket.py:1940` — WS path: `self.view_instance.request = request`

The `_framework_attrs` mechanism (`python/djust/live_view.py:526`) captures `__init__`-time attribute names as framework state, which is excluded from the user-state snapshot. `request` is never assigned in `__init__`, so it falls outside `_framework_attrs` and is snapshotted as user state — where an `ASGIRequest` is non-msgpack-serializable.

## Impact

- Cosmetic but **noisy**: fires on every mount and every event for *every* LiveView, since the request mixin is universal. It also dilutes the signal of the same warning when it catches a *genuine* app-author bug.
- Not a correctness problem in practice: djust `str()`-stringifies the value, and the framework re-sets `self.request` to the live request on every request/event (`mixins/request.py:489` / `websocket.py:1940`), so the stringified copy is never read back.

## Suggested fix

Assign `self.request = None` in `LiveView.__init__` **before** the `_framework_attrs` snapshot line (`live_view.py:526`). That places `request` in `_framework_attrs` → treated as framework state → excluded from the user-state snapshot. The two later `self.request = request` assignments then merely update an existing framework attr.

This matches the `_framework_attrs` snapshot-order invariant already documented at `live_view.py:518` (#1393).

## Environment

- djust 1.0.0rc4
- Observed in an on-device iOS app (Path C — djust/Django served over the loopback interface inside a Toga/Briefcase app), `DJUST_CONFIG['STATE_BACKEND'] = 'memory'`. Reproduces on every view mount.

PR #1547, opened 2026-05-19
fix(build): gate djust_live extension-module behind a Cargo feature (#1543)

### LINKED ISSUE #1543: tech-debt: djust_live cannot be cargo-tested — gate the extension-module feature behind a Cargo flag

johnrtipton:
Closes #1543.

## The problem

`crates/djust_live` carried PyO3's `extension-module` feature unconditionally — `cargo test -p djust_live` failed at link time with `ld: symbol(s) not found ... Py_True`. The crate that holds `djust._rust`'s entry point, the actor system, the `RustLiveView` backend, and (since PR #1546) the `PatchResponse` round-trip regression tests had **no fast Rust-native test feedback loop**. `make test` worked around it with `--exclude djust_live`. Surfaced twice in the v1.0.0rc4 Phase-2 drain (PRs #1530, #1535) — standing structural constraint.

## The fix

Gate `extension-module` behind a default-on Cargo feature exactly as the issue body proposed:

```toml
[features]
default = ["extension-module"]
extension-module = ["pyo3/extension-module"]
```

Drop the inline `features = ["extension-module"]` from the `pyo3` dep line — the feature is now added by our `default`.

`maturin develop` / `cargo build` get the feature via `default`, so the runtime path is unchanged. `cargo test -p djust_live --no-default-features` links against libpython and runs.

## What's now testable

37 djust_live unit tests now execute, including:
- 13 actor system tests (Mount, Event, RenderResult clones, supervisor TTL cleanup, etc.)
- 4 `msgpack_round_trip_patch_response_*` regression tests from PR #1546 (#1541) — these compile-checked only before; this PR retroactively unlocks them.
- The remaining message-serialization, error-formatting, and value-conversion tests in the crate.

## Wiring

- `crates/djust_live/Cargo.toml` — `[features]` section + dropped inline pyo3 feature.
- `Makefile` — `test-rust` target + parallel `test` target gain a Phase 2 invocation.
- `.github/workflows/test.yml` — Rust-tests CI job gains a Phase 2 step.

## Verification

| Check | Result |
|---|---|
| `cargo test -p djust_live --no-default-features` | 37/37 ✅ |
| `cargo test --workspace --exclude djust_live` (unchanged path) | 818/818 ✅ |
| `maturin build --release` end-to-end | wheel built ✅ |
| `python -c "import djust"` after rebuild | OK ✅ |

## Two-commit shape (#1173)

- `103eced7` — implementation (Cargo.toml + Makefile + CI)
- `a0a4e545` — CHANGELOG only

## Out of scope

- Splitting `djust_live` into a Rust core + thin PyO3 wrapper (proper long-term refactor; not tracked).
- The `pyproject.toml` `[tool.maturin] features = ["pyo3/extension-module"]` line is left as-is — maturin's explicit passthrough is independent of our feature default, so removing it is unnecessary and removing it would couple the two configs unnecessarily.


johnrtipton (commented):
## Stage 11 Code Review — PR #1547 (closes #1543)

**Verdict: APPROVE.** Empirically validated end-to-end on macOS arm64.

### Empirical canary (#252) — passed
Ran the central claim locally: `PYO3_PYTHON=...venv/bin/python cargo test -p djust_live --no-default-features` — **37 tests passed, 0 failed**, including all 4 #1541 `msgpack_round_trip_patch_response_*` regressions (`actors::messages::tests::msgpack_round_trip_patch_response_{both_none,patches_none,html_none,both_some}`) that were previously compile-check-only. The PR's CHANGELOG claim of 37 djust_live tests now executing is correct to the test count.

### Failure-mode bisection — sanity-checked
Also ran the negative case: `cargo test -p djust_live` (default features, i.e. with `extension-module` ON) — fails at link time with exactly the documented `ld: symbol(s) not found ... __Py_TrueStruct` error. So the feature-gate genuinely changes the link behavior, and the two-phase Makefile / CI invocation (`crates/djust_live/Cargo.toml:25-29` + `Makefile:184-186` + `.github/workflows/test.yml:59-64`) is load-bearing, not cosmetic.

### Checklist
- [x] `[features]` block correct (`default = ["extension-module"]; extension-module = ["pyo3/extension-module"]` at `Cargo.toml:25-29`).
- [x] `pyo3` dep line drops inline `features = ["extension-module"]` (`Cargo.toml:31`) — feature now flows through `default`.
- [x] Maturin path independent — `pyproject.toml:134-135` carries its own `features = ["pyo3/extension-module"]`, so maturin is decoupled from the Cargo default. (The Cargo.toml comment at lines 19-23 makes this explicit; nice.)
- [x] Makefile `test-rust` (lines 184-186) and parallel `test` (line 159) both gain Phase 2.
- [x] CI Phase 2 positioned correctly: after workspace test, before clippy/fmt (`test.yml:64`).
- [x] Two-commit shape per #1173: `103eced7` (3 files, no CHANGELOG) + `a0a4e545` (CHANGELOG only).
- [x] Edge case checked: `cargo test --workspace` without `--exclude djust_live` still fails (default features link `extension-module`); behavior is unchanged for anyone not using the two-phase invocation — they get the same `Py_True` link error as before. The CHANGELOG entry already calls this out.
- [x] `cargo bench` parallel update at `Makefile:304` mirrors the test pattern — consistent.

### Question
The CHANGELOG entry doesn't mention that `cargo test -p djust_live` (default features, i.e. the bare invocation a developer would naïvely try) **still** fails with the same `Py_True` error. The `--no-default-features` flag is mandatory. Worth a one-line README/CONTRIBUTING note pointing newcomers at `make test-rust` instead of bare `cargo test`? Not blocking — the Makefile-driven path is the documented entry point — but a developer who clones the repo and runs `cargo test -p djust_live` will hit the link error and have no inline guidance.

### Edge cases checked
1. `cargo test -p djust_live` (default features) — fails at link as expected (Py_True).
2. `cargo test -p djust_live --no-default-features` — 37 pass, 4 doctest ignored (unchanged from baseline).
3. The 4 #1541 `msgpack_round_trip_patch_response_*` tests are now genuinely executed (previously compile-checked only per the PR #1546 CHANGELOG entry).
4. Maturin build path: `[tool.maturin]` at `pyproject.toml:134-135` carries `pyo3/extension-module` independently, so even if a future PR flips the Cargo default to off, maturin builds continue to work.

Approving.

johnrtipton:
## Stage 14 retro — PR #1547 (#1543)

### What went well
- **Empirical canary (#252) ran clean and proved the load-bearing claim.** The reviewer ran `cargo test -p djust_live --no-default-features` and confirmed 37 tests pass, including the 4 `msgpack_round_trip_patch_response_*` from PR #1546. The reviewer also verified the *negative* case — bare `cargo test -p djust_live` still link-errors with the exact `Py_TrueStruct` symptom #1543 documents — which is the strongest possible proof the feature-gate is load-bearing.
- **Side-benefit retroactivity.** This PR retroactively activates the #1541 Layer B tests as a no-cost dividend. PR #1546's "compile-checks only" caveat is now obsolete; the regression coverage is real.
- **Two-commit shape (#1173)** + **stale-base gate (#1450)** + **branch-verify reflex** all ran clean.
- **Cross-task sequencing paid off.** Processing #1541 before #1543 in the drain meant the Layer B tests already existed by the time the gate landed — the gate's value is immediately visible. If we'd done them in reverse order, the gate would have landed with no demonstrable test to unlock.

### What could be improved
- **State-file gate (#840) almost missed.** I created the plan file directly without running pipeline-next; the state JSON didn't exist when I went to commit. Caught by the branch-verify reflex output being empty. Filed the state file retroactively before committing — no harm, but the gap is real. Process note: every pipeline iteration past the first needs an explicit pipeline-next OR explicit state-file scaffold; don't shortcut even for "obviously small" fixes.
- **Reviewer's non-blocking question** about a CONTRIBUTING/README hint pointing newcomers at `make test-rust` (since bare `cargo test -p djust_live` still link-errors) is worth a one-line follow-up. Captured here for the v1.0.0rc6 milestone retro; not worth a follow-up issue on its own (would batch with similar docs nits if any).

- This is task 2 of 3 in the v1.0.0rc6 drain. The first (#1541, PR #1546) created the test coverage this gate unlocks. The third (#1545) is unrelated subsystem (Python `LiveView.request` snapshot warning).
- The "candidate canon rule" from PR #1546's retro (the leading-vs-trailing #1538 generalization caveat) is still pending consolidation into the v1.0.0rc6 milestone retro.

### Verdict
`RETRO_COMPLETE`

PR #1548, opened 2026-05-19
fix(live_view): assign request=None in __init__ — silence ASGIRequest snapshot warning (#1545)

### LINKED ISSUE #1545: LiveView.request triggers "non-serializable value: ASGIRequest" warning on every state snapshot

johnrtipton:
Closes #1545.

## The bug
Every `LiveView` mount/event was logging:
```
WARNING:djust.serialization:LiveView state contains non-serializable value: ASGIRequest (from django.core.handlers.asgi). This will be converted to a string …
```

`self.request` was assigned by `mixins/request.py:489` (HTTP) and `websocket.py:1940` (WS) AFTER `__init__`, so it sat outside `_framework_attrs` — the snapshot machinery treated `ASGIRequest` as user state, hit the non-serializable fallback (`serialization.py:557`), and stringified it on every event.

Cosmetic but noisy; the warning's signal was diluted for genuine app-author bugs.

## The fix
Assign `self.request: Any = None` in `LiveView.__init__` BEFORE the `_framework_attrs = frozenset(self.__dict__.keys())` line at `live_view.py:526`. Matches the `_framework_attrs` snapshot-order invariant (#1393).

The two later `self.request = request` assignments in `mixins/request.py:489` and `websocket.py:1940` simply update the existing framework attr — no other change needed.

## A second filter the issue body didn't mention
Regression-suite verification surfaced a second filter that ALSO needed `"request"` added: `_FRAMEWORK_INTERNAL_ATTRS` — a hard-coded frozenset in `live_view.py:94-125` used by `_debug_state_sizes` and the debug-toolbar observability path. Without this 2nd update, `test_debug_state_sizes_empty_view` and `test_debug_state_sizes_sorted_output` failed because `request` leaked into state-size accounting.

This is exactly the **"Symbol-migration grep canon" (Action #1391, #1400)** pattern — when changing a filter convention, grep for OLD filter references across the codebase. Without the regression suite this PR would have shipped an incomplete fix.

## Tests
5 regression cases in `python/tests/test_liveview_request_framework_attr_1545.py`:
- `test_request_attribute_exists_post_init` — `hasattr(v, "request")` true, `v.request is None`.
- `test_request_is_in_framework_attrs` — load-bearing membership assertion.
- `test_user_attrs_set_in_mount_are_NOT_in_framework_attrs` — regression-of-regression guard.
- `test_no_asgirequest_warning_on_fresh_init` — `caplog`-driven warning silence check.
- `test_framework_attrs_includes_known_framework_slots` — sanity check on `request`, `_object`, etc.

**Gate-off self-test (Action #254):** 4 of 5 tests fail without the fix — the framework-attrs-membership assertions are genuinely load-bearing.

**Full Python regression: 7301 passed, 0 failed.**

## Two-commit shape (#1173)
- `d8d46087` — implementation (`live_view.py`) + 5 regression tests
- `123c3cd6` — CHANGELOG only


johnrtipton (commented):

**Stale-base check (#1450):** 0 commits behind `origin/main` — clean.

**Empirical canary (#252):** Gated the fix off (removed both the `self.request: Any = None` block at `python/djust/live_view.py:517-528` and the `"request"` entry in `_FRAMEWORK_INTERNAL_ATTRS` at `python/djust/live_view.py:102-103`), reran `python/tests/test_liveview_request_framework_attr_1545.py` — got **4 failed / 1 passed** as expected. The four failures hit the load-bearing assertions (`hasattr(v, "request")`, `"request" in v._framework_attrs`, the regression-of-regression guard, and the known-slots invariant). The one passing test (`test_no_asgirequest_warning_on_fresh_init`) is correctly silent at __init__ time because the placeholder doesn't exist — the warning surfaces only on a real `post()` / WS dispatch, not on bare construction. That gap is acceptable for this PR's scope (#1545 is filed against the cosmetic warning on every mount/event; the integration path is exercised in the manual reproduction the issue body cites), but it does mean the warning-silence claim has no executable post-fix test in the regression file. Question: would you like a follow-up adding a `RequestFactory`-driven integration case that exercises `post()` → snapshot → assert-no-warning, to lock in the actual end-to-end silence? (Not blocking — the framework-attr membership assertion is logically sufficient.)

**Diff checklist:**
- [x] `self.request: Any = None` placement at `python/djust/live_view.py:528` is BEFORE the `_framework_attrs = frozenset(self.__dict__.keys())` snapshot at `python/djust/live_view.py:541`. Invariant #1393 preserved.
- [x] Doc comment at `live_view.py:517-527` explains the rationale, cites #1545, and references the post-init reassignment paths (`mixins/request.py:489`, `websocket.py:1940`) and the warning origin (`serialization.py:557`).
- [x] `_FRAMEWORK_INTERNAL_ATTRS` at `python/djust/live_view.py:94-127` includes `"request"` at line 103 with comment "Request handle — reassigned per HTTP/WS event (#1545)". Defense-in-depth — also covers `_debug_state_sizes` / debug-toolbar paths flagged in the CHANGELOG entry.
- [x] 5 regression tests at `python/tests/test_liveview_request_framework_attr_1545.py` exercise distinct assertions: attribute existence, framework-attrs membership, regression-of-regression guard against sweeping mount-set user attrs, warning silence, and known-slots invariant.
- [x] Stage-5 gate: `git show d8d46087 --name-only | grep CHANGELOG` returns empty — fix commit has no CHANGELOG.
- [x] Two-commit shape (#1173) honored: `123c3cd6` is CHANGELOG-only.

**Edge case — user `self.request` in their own `__init__`:** Verified harmless. If a subclass assigns `self.request = something` either before or after `super().__init__()`, `_framework_attrs` is built from `self.__dict__.keys()` — `"request"` is in keys regardless of which value was last assigned. The frozenset is name-based, not identity-based, so the framework-attrs filter still excludes it from user-state snapshots. The user's value will still be visible at template-context time via the public attribute name. No interaction risk.

**Edge case — interaction with `_FRAMEWORK_INTERNAL_ATTRS`:** The two filters cover different paths (`_framework_attrs` is the instance-level snapshot used by serialization; `_FRAMEWORK_INTERNAL_ATTRS` is the class-level static set used by debug/observability). Both now exclude `request`, matching the existing pattern for `http_method_names` / `args` / `kwargs` which are also in both. Symmetric.

**Verdict:** APPROVE.

johnrtipton:
## Stage 14 retro — PR #1548 (#1545)

### What went well
- **The fix surfaced a SECOND filter the issue body didn't mention.** The issue body cited the `_framework_attrs` snapshot at `live_view.py:526` and proposed the one-line fix. The full Python regression then surfaced that `_FRAMEWORK_INTERNAL_ATTRS` (a class-level hard-coded frozenset used by `_debug_state_sizes` and the debug-toolbar observability path) ALSO needed `"request"` added — 2 `test_debug_state_sizes_*` tests failed without it. The Action #1391 / #1400 "symbol-migration grep canon" rule is exactly designed to catch this; this PR exercised it visibly.
- **Gate-off self-test (#254) at Stage 5.** Stashed the fix, re-ran the new tests: 4 of 5 failed as required (the warning-silence test passed vacuously since `__init__` doesn't snapshot — see below). Stage 11 reviewer reproduced the gate-off independently.
- **Two-commit shape (#1173) preserved.** Implementation in `d8d46087` (zero CHANGELOG); CHANGELOG in `123c3cd6`.
- **Stage 11 raised a sharp, non-blocking question.** The reviewer noted that `test_no_asgirequest_warning_on_fresh_init` passes vacuously without the fix (because `__init__` alone doesn't trigger snapshot serialization). The framework-attrs-membership assertions ARE load-bearing, but a `RequestFactory`-driven integration test that exercises the actual `post()` → snapshot path would lock in true end-to-end silence. Captured as a candidate follow-up.

### What could be improved
- **Pre-commit hook bounced the first Stage 5 commit (ruff F841 — unused local `v`).** Action #122 (`--amend` / commit verification) caught it via the PRE/POST hash compare. Cost: ~1 extra cycle (re-run ruff explicitly to identify the error, fix the test, re-stage, re-commit). Process note: run `ruff check <staged-files>` explicitly BEFORE `git commit` for any Python file edit, rather than relying on the pre-commit hook's first run.
- **Reviewer-subagent side effect.** The Stage 11 reviewer's gate-off self-test left `Cargo.lock` in `UU` (unmerged) state. Recovered via `git checkout HEAD -- Cargo.lock`. Process note: the reviewer's empirical canary should always stash-or-restore working-tree files cleanly; the merge of operations under the subagent's git stash/restore semantics is a known fragility (#180 / #1172).
- **Two-filter discovery.** Could have been caught at Stage 4 plan time by grepping for `request` in `live_view.py` AND for `_FRAMEWORK_INTERNAL_ATTRS` usage sites BEFORE implementation. Lesson generalizable to any future framework-attr addition: every new framework attr needs to be added to BOTH filters.

This is the third and final task of the v1.0.0rc6 drain. Unrelated subsystem (Python `LiveView` lifecycle) from #1541 (Rust msgpack) and #1543 (Rust Cargo features); the three-issue mix exercised distinct code paths and surfaced three distinct sets of process artifacts:
- #1541 → "leading-vs-trailing serde fix shape doesn't generalize" (candidate canon rule).
- #1543 → "Cargo feature gates can retroactively unlock test coverage from prior PRs" (positive observation).
- #1545 → "framework-attr additions need symbol-migration grep across all filters" (Action #1391 / #1400 reinforced).

All three to be consolidated in the v1.0.0rc6 milestone retro.

### Verdict
`RETRO_COMPLETE`

PR #1553, opened 2026-05-20
test(vdom): pin {% if/elif %}+include diff (#1552 hypothesis disconfirmed — investigation continues)

### LINKED ISSUE #1552: VDOM diff appends new subtree instead of replacing when {% if %}/{% elif %} swaps between {% include %}d templates (regression from 0.9.6rc2)
## Summary

## Reproduction

```django
<div class="wizard-content">
        ...
        {% endif %}
    {% endblock %}
</div>
```

### Concrete observation

## Workaround

## Versions

- Django: 5.2.13
- Python: 3.12

## Likely related code

## Reproduction project

johnrtipton:
Refs #1552. **Does NOT fix #1552** — pins diff-layer invariants and documents the hypothesis disconfirmation.

## What this PR is

Stage 5 of the #1552 bugfix pipeline. The plan's primary suspect — Candidate A, **#1538's `#[serde(default)]` on `VNode.djust_id`** (commit `0a119962`) — was empirically RULED OUT at the VDOM-diff layer.

## What was tested + what was found

Two VNode-shape reproducers in `crates/djust_vdom/tests/test_include_subtree_double_1552.rs` (481 lines) modelling the production scenario:

- `test_include_swap_elif_cascade_multi_rooted_1552` — outer matched-id `{% if A %}` boundary with 8 OLD body children, NEW with nested `{% elif B %}` boundary + 5 different children.
- `test_include_swap_matched_body_all_removed_via_nested_boundary_1552` — narrower 3-vs-1 child case.

Both PASS on current HEAD using a production-faithful id-aware applier (what the JS client at `python/djust/static/djust/src/12-vdom-patch.js:1778-1788` actually does — `child_d`-first, index fallback).

Both PASS at v0.9.6rc2 too (verified via `git worktree add`). The differ's behavior is IDENTICAL between v0.9.6rc2 and v1.0.0rc4 → **no diff-layer regression**.

## Why v0.9.6rc2 \"works\" in production

Pre-#1538, msgpack deserialize would fail and `state_backends/memory.py:118-148`'s fail-soft remount path triggered on every event — the diff path was NEVER exercised. The framework re-mounted from scratch each time, which always produces the correct DOM (just slower). Post-#1538, deserialize succeeds and exposes a real bug in an adjacent surface, not in the differ.

## Candidate next-investigation surfaces

All three shipped in the regression window v0.9.6rc2..v1.0.0rc4:

1. **WS-event save block** (PR #1466 / commit `a5e2c50c`) — `last_vdom` now persists on every WS event.
2. **Sticky-child persistence** (ADR-018 PRs #1526/#1527/#1528) — three new persistence layers.
3. **JS-only patch-ordering edge case** — lower probability but not ruled out.

## Why land tests-only

The two tests here pin diff-layer behavior. Any future fix in this area gets these as regression backstops; if a future change reintroduces the failure shape, it fails fast at Stage 6 of the pipeline rather than surfacing as a production bug. The commit message (`b57452d5`) documents the full hypothesis exploration explicitly so the next investigator doesn't redo it.

## Failed prototype (documented for posterity)

An early Stage 5 prototype made the diff emit `RemoveChild+InsertChild` for any position-paired siblings with mismatched `djust_id`s. It broke `proptest_round_trip_with_sync.rs` (the #1408 invariant) because **position-identity (not djust_id-identity) is the canonical sibling-pairing rule** — `sync_ids` reconciles ids post-diff, and the proptest's auto-assigned fresh ids mean nearly every paired sibling has mismatched ids before reconciliation. The fix was reverted. This documents what the differ's contract is so future fixers don't re-attempt this shape.

## Plan + state file

- Full Stage 4 plan: `.pipeline-state/fix-1552-vdom-include-subtree-double-plan.md` (in-repo; gitignored — visible locally).
- Stage 5 verdict: `STAGE_5_BLOCKED: hypothesis disconfirmed`.

### DIFF
--- a/crates/djust_vdom/tests/test_include_subtree_double_1552.rs
+++ b/crates/djust_vdom/tests/test_include_subtree_double_1552.rs
@@ -1,0 +1,481 @@
+//! Regression / disconfirmation tests for #1552: production reporter
+//! observed VDOM doubling when an `{% if %}/{% elif %}` block swaps the
+//! `{% include %}` it emits between two renders.
+//!
+//! ## Status: VDOM-diff layer is NOT the cause
+//!
+//! These tests pin the diff layer's behavior for the exact shape the
+//! reporter described. On current HEAD both tests PASS: the differ
+//! correctly emits `InsertSubtree(if-X-1)` for the new nested boundary
+//! AND `RemoveChild` (with `child_d`) for every OLD body child. When
+//! applied with id-based RemoveChild resolution (production JS semantics
+//! at `python/djust/static/djust/src/12-vdom-patch.js:1778-1788`), the
+//! live VDOM matches NEW exactly.
+//!
+//! Hypothesis-disconfirmation status from Stage 5 of the bugfix pipeline
+//! for #1552: the plan's primary suspect (Candidate A — #1538's
+//! `#[serde(default)]` on `VNode.djust_id`) is empirically ruled out at
+//! the diff layer. These tests fail BOTH at v0.9.6rc2 and v1.0.0rc4 when
+//! a naive index-only applier is used, and pass BOTH when an id-aware
+//! applier is used — proving the regression-window analysis (only one
+//! VDOM commit in the window, `0a119962`) does not point at a differ
+//! regression.
+//!
+//! Per the bugfix-pipeline canon (Action #1466 / "gate the change off"
+//! self-test), shipping a fix in the diff layer would be premature
+//! without a failing test. The remaining suspect surfaces are:
+//!
+//!   - **WS-event save block (PR #1466, commit `a5e2c50c`)**: writes
+//!     `last_vdom` to the session on every WS event. Did not exist
+//!     pre-v0.9.6rc2's window in the same shape — may interact with the
+//!     msgpack roundtrip such that a *partial* sync_ids state lands in
+//!     the next event's `last_vdom`, producing patches the client can't
+//!     resolve.
+//!   - **Sticky-child persistence (PRs #1526/#1527/#1528, ADR-018)**:
+//!     three new persistence layers were added in the window. If the
+//!     wizard's parent view holds a sticky child whose `dj-id` namespace
+//!     conflicts with the include's content, restore could emit OLD
+//!     content under the parent boundary.
+//!   - **Pre-#1538 fallback semantics**: pre-fix, `state_backends/
+//!     memory.py:118-148` would *log and discard* a view on msgpack
+//!     deserialize failure; the next event would remount from scratch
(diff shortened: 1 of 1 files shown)


johnrtipton:
## Stage 14 retro — PR #1553 (#1552 test-pinning)

### What went well
- **Empirical disconfirmation is real progress.** The Stage 4 plan hypothesized that #1538's `#[serde(default)]` caused `child_d: None` propagation → wrong-child resolution after `InsertSubtree`'s index shift. The Stage 5 implementer subagent wrote production-faithful reproducers, ran them against current HEAD AND against `v0.9.6rc2` (via `git worktree`), and showed the differ is correct at both versions. That's a genuine finding even though it doesn't fix the bug — it eliminates an entire class of hypotheses and tells the next investigator where NOT to look.
- **Honest commit + PR shape.** The PR is tests-only, no fix. The commit message + PR body document exactly which hypotheses were tested, what's known after, and what's pinned. Future regressions in this shape fail fast at Stage 6 instead of surfacing as production bugs.
- **The 481-line test file pins genuine diff-layer invariants.** Production-faithful applier helpers (`apply_all_id_aware`) added to the test commons; if a future differ change breaks the `{% if/elif %}` + `{% include %}` diff shape, the test catches it.

### What could be improved
- **Stage 4 plan didn't reproduce-first.** The plan was thorough on file:line + read-only static analysis but couldn't run the reproducer (the planning agent is read-only by harness policy). Stage 5's implementer caught the planning hypothesis was wrong, but the gap is real: a "Stage 4b — reproducer-write" sub-stage between Stage 4 (plan) and Stage 5 (implementation) would make this kind of empirical disconfirmation discoverable earlier. Worth canonizing.
- **The "regression window" framing misled the plan.** The plan correctly noted that 0.9.6rc2..1.0.0rc4 has ZERO `diff.rs` / `parser.rs` / `12-vdom-patch.js` changes, then incorrectly inferred the most-likely candidate must be the only VDOM-touching change in the window (`0a119962` / #1538 serde-default). The actual root cause was an INTERACTION between #1538 unmasking and #1466's WS-save block — both of which DID land in the window — making the bug reachable for the first time. The lesson: regression-window-of-blame analysis is necessary but not sufficient; interactions matter as much as individual diffs.

### Connection to milestone
This PR was the *first* attempt at #1552 in v1.0.0rc6's open-issue drain. It's the empirical work that made PR #1555 (the actual fix) possible — without the disconfirmation, the actual fix would have started from a wrong hypothesis. Process-wise, this is exactly what "land a test pin even when the fix doesn't land" looks like; it's worth preserving the pattern.

### Verdict
`RETRO_COMPLETE` (test-pinning + investigation finding; fix in PR #1555)

Issue #1550, opened 2026-05-20
VDOM diff fails to insert subtree when {% if %} condition flips after server-driven re-render


johnrtipton:
## Summary

When a Django template's `{% if %} ... {% else %} ... {% endif %}` block flips truthiness due to a server-driven re-render (event handler updating model state, then the template re-evaluating), djust's VDOM diff does not insert the new subtree. The page continues to show the old branch until a hard refresh.

## Reproduction

1. A view renders a template with two large, structurally-different branches inside `{% if obj.flag %} ... {% else %} ... {% endif %}`.
2. User triggers an event handler (`dj-click`) that flips `obj.flag` server-side via a service call + DB write.
3. The view re-renders; the template now emits the other branch.
4. **Expected:** djust patches the DOM to show the new branch.
5. **Actual:** DOM stays at the old branch. Hard-reloading the page shows the new branch correctly.

## Workaround

Replace `{% if %}/{% else %}` with two sibling divs that toggle `d-none` based on the condition:

```html
<div class="{% if not flag %}d-none{% endif %}">... branch A ...</div>
<div class="{% if flag %}d-none{% endif %}">... branch B ...</div>
```

djust patches the `class` attribute reliably, so this avoids the subtree-insertion code path.

## Versions

- djust: 1.0.0rc4
- Django: 5.2.13
- Python: 3.12

## Observed in

NYC Comptroller Claims project's supervisor-approval section on the investigation workflow page: supervisor clicks "Approve Liability" → `LiabilityAssessment.locked` flips True → template should swap from an input/button form to a "Pending Subrogation" badge + Unlock button. djust 1.0.0rc4 does not perform the swap. Workaround applied: convert to a `d-none` toggle pattern (see [project CLAUDE.md "Conditional Sections — Use d-none, Not `{% if %}`" convention](https://github.com/flexion/flexion-nyc-claims/blob/main/CLAUDE.md)).

## Why this is worth fixing in djust

`{% if %}` is the most natural way to express conditional branches in a Django template; the `d-none` workaround leaves both branches in the DOM (cost: bandwidth, accessibility tree size, extra event listeners). For static initial renders the current behavior is fine — the bug only surfaces on server-driven state transitions, which is the core djust use case.

## Related

- This convention has been documented as a project-level gotcha since at least 2026-03; we've worked around it across the codebase but it keeps biting new template additions.

PR #1554, opened 2026-05-20
fix(templates): preprocess multi-line {# ... #} comments for Django classical (#1551)

### LINKED ISSUE #1551: Multi-line {# ... #} comment handling disagrees between Rust and Django classical renderers (follow-up to #1423)
## Summary

djust ships two template renderers and they disagree on multi-line `{# ... #}` comment handling. A template that renders cleanly through the Rust engine (the LiveView hot path) crashes through Django's classical renderer (used by `client.get()` in tests, full-page falls and error renders, and any view that uses `render()` instead of djust's `template_name`). The mismatch is silent — projects ship templates that look correct on the dev server (which uses the Rust engine) and break under CI or on error paths.

Follow-up to #1423, which fixed the Rust side but did not normalize behavior with classical Django.

## Reproduction

Template:
```django
{# Top of file: multi-line comment that explains a layout decision and
   mentions {% if foo %} somewhere in the prose to be clear about what
   pattern is being avoided. #}
<p>Hello</p>
```

Render paths:

- **Rust renderer** (`djust._rust.render_template_with_dirs` — used for LiveView WS responses and prod page renders): treats `{# ... #}` as opaque even across newlines. Renders `<p>Hello</p>`. No error. ✅
- **Django classical renderer** (`django.template.engines['django'].get_template(...).render(...)` — used by `client.get()` in pytest, by Django's debug error page renderer, and by any view that uses Django's `render()` shortcut directly): the `{# ` doesn't match Django's single-line comment regex, the `{% if foo %}` inside the comment is tokenized as a real tag, and parsing fails with:

```
django.template.exceptions.TemplateSyntaxError: Unexpected end of expression in if tag.
  File ".../django/template/smartif.py", line 149, in nud
    raise parser.error_class("Unexpected end of expression in if tag.")
```

## Why this matters

Concrete pain in the NYC Claims Processing project (2026-05-19, against djust 1.0.0rc4):

1. Author writes a multi-line `{# ... #}` comment that mentions `{% if %}` as prose explaining why they used `d-none` instead.
2. `make dev` (uvicorn + Rust renderer): template renders fine. Author ships it.
3. Some pytest tests use `client.get()`, which goes through Django classical. CI fails with the `TemplateSyntaxError`.
4. Author sees the failure 10 minutes later, traces it back to a comment they could not have known would crash.

Same project, same session, I hit this trap twice in 90 minutes — once after explicitly documenting the gotcha in CLAUDE.md and adding a Stage 3 / Stage 7 pipeline-ship review check for it. The behavior is asymmetric enough that experienced developers writing for "djust" forget Django classical is still in the call stack on other paths.

## Suggested fix

Pick one (in order of preference):

1. **Make djust's Django integration normalize comment handling at template-load time** — preprocess `{# ... #}` blocks before handing the source to Django's classical parser, so the two engines see the same tokens. This contains the inconsistency in djust and doesn't require an upstream Django change.
2. **Reject multi-line `{# ... #}` containing `{% ... %}` in the Rust engine** as well, so the failure mode is identical between paths — at least nobody ships a working template that secretly breaks on the other path.
3. **Document the asymmetry loudly** in the djust docs ("Templates rendered through djust's WS path tolerate multi-line `{# ... #}`; the same template rendered through Django classical will crash if the comment contains template-tag syntax"). Least helpful but at least makes the trap visible.

## Versions

- djust: 1.0.0rc4
- Django: 5.2.13
- Python: 3.12

## Reproduction in the wild

- This session's PR scope: combined fix branch landing multiple demo-prep changes
- Files I tripped on:
  - `templates/settlement/claimant_sign.html` (header — comment leaked as text through both renderers due to the older single-line-only Django bug + multi-line content)
  - `templates/investigation/investigation_workflow.html` (supervisor approval section — raised `TemplateSyntaxError` under Django classical because the comment contained `{% if %}`)

johnrtipton:
Closes #1551. Follow-up to #1423.

## TL;DR
djust ships two template renderers and they disagreed on multi-line `{# ... #}` handling. The Rust engine treats them as opaque across newlines; Django classical's tokenizer doesn't (non-DOTALL regex), so a `{% if %}` inside a multi-line comment body parsed as a real tag and crashed with `TemplateSyntaxError`.

**Silent footgun:** templates rendered fine via the LiveView WebSocket path (Rust) and crashed via `client.get()` in pytest, Django's debug error renderer, or any view using `render()` directly. Reporter (NYC Comptroller Claims) hit it twice in 90 minutes after explicitly documenting the gotcha.

## Fix

Per issue body option 1: new `djust.template.loaders.FilesystemLoader` and `djust.template.loaders.AppDirectoriesLoader` — drop-in replacements for the standard Django loaders that preprocess `{# ... #}` blocks out of the template source before Django classical's tokenizer sees them.

```python
TEMPLATES = [{
    'BACKEND': 'django.template.backends.django.DjangoTemplates',
    'DIRS': [...],
    'APP_DIRS': False,  # explicit loaders shape required
    'OPTIONS': {
        'loaders': [
            'djust.template.loaders.FilesystemLoader',
            'djust.template.loaders.AppDirectoriesLoader',
        ],
    },
}]
```

Single-line comments are stripped too — Django strips them anyway, so doing it uniformly is safe and symmetric.

## Coverage

10 regression tests in `python/djust/tests/test_multiline_comment_parity_1551.py`:

- `test_django_classical_crashes_on_multiline_comment_with_tag_syntax` — control; pins the bug shape in vanilla Django classical
- 6 `strip_multiline_comments()` unit tests — basic, single-line, preserves-tags, nested-braces, multiple, non-greedy
- 2 loader-level E2E tests through Django `Engine` — confirms problematic template now renders + non-comment content passes through untouched
- `test_django_classical_single_line_comment_works` — controls the comparison

## Verification

- Full Python regression: **7,311 passed, 0 failed.**
- New tests: **10/10 passing.**
- Performance: one regex pass per template load (~1 ms or less); loaded templates are cached so this runs once per template, not per render.

## Two-commit shape (#1173)
- `107322b8` — implementation (`loaders.py` + 10 tests)
- `8dbfdd85` — CHANGELOG only

## Adoption recommendation
Recommend documenting in the Getting Started guide and the migration guide from older djust versions. Not a default-on change (don't silently mutate users' loaders) — opt-in via settings is the right balance per #1551's review.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- **Multi-line `{# ... #}` comments containing template-tag syntax no longer crash Django classical (#1551).** djust ships two template renderers: the Rust engine (`crates/djust_templates/src/lexer.rs:289-305`) treats `{# ... #}` as opaque even across newlines, but Django's classical tokenizer (`django.template.base.Lexer`) uses a non-DOTALL regex — multi-line `{# ... #}` was NOT recognized as a single comment, so a `{% if %}` inside the comment body parsed as a real tag and raised `TemplateSyntaxError`. The mismatch was silent: templates rendered fine via the LiveView WebSocket path (Rust) and crashed via `client.get()` in pytest, Django's debug error renderer, or any view using `render()` directly. Follow-up to #1423. Fix: new `djust.template.loaders.FilesystemLoader` and `djust.template.loaders.AppDirectoriesLoader` — drop-in replacements for the standard Django loaders that preprocess `{# ... #}` blocks out of the template source before Django classical's tokenizer sees them. Single-line comments are stripped too (Django strips them anyway). Projects opt in by replacing the default loaders in `TEMPLATES['OPTIONS']['loaders']`. Performance: one regex pass per template load (<1ms); loaded templates are cached so this runs once per template, not per render. Covered by 10 regression cases in `python/djust/tests/test_multiline_comment_parity_1551.py`, including a control that pins the bug shape in vanilla Django classical.
+
 ## [1.0.0rc6] - 2026-05-19
 
 ### Security
--- a/python/djust/template/loaders.py
+++ b/python/djust/template/loaders.py
@@ -1,0 +1,130 @@
+"""
+Django template Loader subclasses that normalize comment handling
+between djust's Rust template engine and Django's classical engine.
+
+## The problem (#1551)
+
+djust ships two template renderers:
+
+1. The Rust engine (``djust._rust.render_template_with_dirs``) — used for
+   LiveView WebSocket responses and production page renders. Its lexer
+   at ``crates/djust_templates/src/lexer.rs:289-305`` treats ``{# ... #}``
+   as opaque even across newlines.
+
+2. Django's classical engine — used by ``client.get()`` in pytest, by
+   Django's debug error-page renderer, and by any view that uses
+   ``render()`` directly. Its tokenizer (``django.template.base.Lexer``)
+   uses a non-DOTALL regex; multi-line ``{# ... #}`` is NOT recognized
+   as a single comment. A ``{% if %}`` inside the comment body gets
+   parsed as a real tag → ``TemplateSyntaxError``.
+
+The mismatch is silent: a template renders fine on the dev server
+(Rust path) and crashes in CI or on error paths (classical path).
+
+## The fix
+
+These Loader subclasses preprocess ``{# ... #}`` blocks out of the
+template source BEFORE Django classical's tokenizer sees them.
+Single-line comments are also stripped — Django strips them anyway,
+so doing it ourselves is uniform and safe.
+
+## Usage
+
+Replace the default Django loaders in your ``TEMPLATES`` setting::
+
+    TEMPLATES = [
+        {
+            "BACKEND": "django.template.backends.django.DjangoTemplates",
+            "DIRS": [...],
+            "APP_DIRS": False,  # explicit loaders shape required
+            "OPTIONS": {
+                "loaders": [
+                    "djust.template.loaders.FilesystemLoader",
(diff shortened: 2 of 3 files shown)


johnrtipton:
## Stage 14 retro — PR #1554 (#1551)

### What went well
- **Issue body was the plan.** The reporter pre-suggested 3 fix options ranked by preference; option 1 (preprocess in djust's Django integration) was the cleanest and matched the Rust-side asymmetry origin (only Django classical breaks, so the fix lives at the Django-integration boundary). Stage 4 plan was the issue body verbatim plus an effort estimate.
- **TDD first-pass clean.** Wrote 10 regression cases covering control / 6 unit tests / 2 E2E loader tests / 1 control before any code. Bug test passed (control), fix tests failed (no module), fix landed, all 10 passed. One pre-commit ruff cycle bounce that auto-fixed (cleaned imports) — Action #122's `&& git log -1 --oneline` caught it.
- **Performance honesty.** Performance claim is the actual cost: one regex pass, single load, cached. The CHANGELOG says it; not adjective inflation.

### What could be improved
- **The fix is opt-in.** Projects must replace the default Django loaders in `TEMPLATES['OPTIONS']['loaders']`. Discoverability is documentation-dependent. Worth considering whether djust's `DjustConfig.ready()` should warn (Django system check) when a user has the standard FilesystemLoader configured without djust's wrapper. Not for this PR — surfaces in v1.1 docs work.
- **The bug class generalizes.** Any future template-engine asymmetry between the Rust and Django classical paths inherits the same silent-CI-failure shape. A meta-check would walk both engines against the same input and assert byte-stable output. Tracked candidate.

### Connection to milestone
This is one of the three post-rc6 user-reported bugs in v1.0.0rc6's open-issue drain. #1551 was the cleanest of the three — independent subsystem, well-scoped fix, ~3 hours of work. The other two (#1550 + #1552) needed combined investigation via PR #1553 (test-pinning, hypothesis disconfirmed) → PR #1555 (actual fix, dj-id counter collision).

### Verdict
`RETRO_COMPLETE`

PR #1555, opened 2026-05-20
fix(vdom): advance dj-id counter past last_vdom's max before parse (#1550, #1552)

### LINKED ISSUE #1550: VDOM diff fails to insert subtree when {% if %} condition flips after server-driven re-render

### LINKED ISSUE #1552: VDOM diff appends new subtree instead of replacing when {% if %}/{% elif %} swaps between {% include %}d templates (regression from 0.9.6rc2)

johnrtipton:
Closes #1550. Closes #1552. Companion to PR #1553 (test-pinning) which empirically disconfirmed the earlier diff-layer hypothesis.

## TL;DR

Both bugs trace to **dj-id counter collisions** when `last_vdom` migrates across worker threads or is msgpack-roundtripped through `InMemoryStateBackend.get()`. The fix is a 3-helper additions to `djust_vdom` + 2 call sites in `RustLiveViewBackend`.

## Root cause

The thread-local djust_id counter generates monotonically-increasing ids during `parse_html_continue`. When `last_vdom` either (a) migrates across worker threads (Channels / asyncio handoff) or (b) is msgpack-roundtripped through `state_backends/memory.py:118`, the new thread's counter is independent of `last_vdom`'s ids. The next `parse_html_continue` generates ids `1..k` that collide with surviving ids in `last_vdom`. The resulting `InsertSubtree.html` carries dj-ids matching other elements in the parent's child list, and the client's id-first `:scope > [dj-id=N]` resolver at `12-vdom-patch.js:1778-1788` picks the **newer** element instead of the **older** one — wrong subtree removed, old content survives.

## Symptom shapes

- **#1552** (P0 regression from 0.9.6rc2): `{% if/elif %}` + `{% include %}` swap doubles subtree. Wizard step-1 → step-2 produced 4 headings (was 3) and 31 inputs (was 25 — all step-1 inputs survived).
- **#1550** (P1): `{% if %}` flip doesn't swap branches. Workaround was `d-none` class toggling.

## Why v0.9.6rc2 worked (the regression-window mystery)

Pre-#1538 (commit `0a119962`, the serde-default fix), msgpack deserialize failed silently and `state_backends/memory.py:118` returned `None` on every cache lookup. The framework full-remounted on every event — no diff path, no collision. Post-#1538 deserialize succeeds, the diff path runs, and the collision shape became reachable in production.

The earlier diff-layer hypothesis (PR #1553 — "child_d:None propagation post-#1538") was empirically **disconfirmed**: production-faithful VNode-level reproducers showed the differ correct at both 0.9.6rc2 AND 1.0.0rc4. The mechanism is dj-id collision under counter divergence.

## Fix

Before each `parse_html_continue` in `RustLiveViewBackend::render_with_diff` (`crates/djust_live/src/lib.rs:498-518`) and the text-fast-path entry (`:668-678`):

1. Walk `last_vdom` via `max_djust_id_in(old_vdom)` to find the highest djust_id.
2. Call `ensure_id_counter_at_least(max + 1)` to monotonically advance the thread-local counter.

The counter becomes effectively per-view, surviving thread handoff and msgpack roundtrip.

## API additions (`djust_vdom`)

Three new public helpers in `crates/djust_vdom/src/lib.rs:113-191`:

- `from_base62(s: &str) -> Option<u64>` — decode base62 string back to u64 (inverse of `to_base62`).
- `max_djust_id_in(node: &VNode) -> Option<u64>` — walk VNode tree, return highest parseable djust_id (None if no node carries one).
- `ensure_id_counter_at_least(min_value: u64)` — monotonic counter advance (no-op when already at/above min).

All additive; no signature changes to existing surface.

## Verification

**16 new tests:**

- **12 Rust unit tests** in `crates/djust_vdom/tests/test_id_counter_monotonicity_1550_1552.rs`:
  - base62 round-trip (single + multi-digit + invalid input rejection)
  - `max_djust_id_in` on id-less / single-id / nested trees / unparseable id mixes
  - `ensure_id_counter_at_least` monotonic semantics (advance / noop-above / noop-equal)
  - Cross-helper invariant: ensure-then-next never collides with the prior max

- **4 Python E2E tests** in `python/tests/test_if_elif_swap_e2e_1550_1552.py`:
  - `test_no_djid_collision_between_last_vdom_and_insert_subtree_html_1552`
  - `test_remove_child_targets_are_uniquely_resolvable_post_insert_1552`
  - `test_no_djid_collision_for_if_else_branch_swap_1550`
  - `test_post_msgpack_roundtrip_counter_advanced_past_last_vdom`

**Full regression:**
- Rust Phase 1 (`--workspace --exclude djust_live --release`): **832 passed, 0 failed.**
- Rust Phase 2 (`-p djust_live --no-default-features`): **37 passed, 0 failed.**
- Python (`pytest python/tests/ python/djust/tests/ tests/`): **7,315 passed, 0 failed.**

## Two-commit shape (#1173)
- `01b2eaa9` — implementation (4 files: `djust_vdom/src/lib.rs` +79, `djust_live/src/lib.rs` +32, 2 new test files)
- `2cc1bf57` — CHANGELOG only

## Release-line implications

#1552's reporter (NYC Comptroller Claims, private repo) has downgraded production to 0.9.6rc2. Once this lands, recommend either:
- Cutting a `1.0.0rc7` immediately so the reporter can re-upgrade and dogfood before GA, or
- Going direct to `1.0.0` GA with this fix included.

## Related

- Disconfirms the diff-layer hypothesis from PR #1553 (test-pinning landed; diff-layer tests stay as regression backstop)
- Builds on PR #1542 (#1538 serde-default fix) — that PR unmasked this latent bug; this PR fixes the unmasked behavior
- Identifies the actual cross-#1538-and-#1466 interaction (msgpack deserialize success + WS save block both arriving in the same release window)


johnrtipton:
## Stage 14 retro — PR #1555 (#1550 + #1552 combined fix)

### What went well
- **Single root cause for two bugs.** #1550 and #1552 looked like different symptom shapes (branch-doesn't-swap vs subtree-doubles) but both reduced to one mechanism: dj-id counter collisions when `last_vdom` survives a thread handoff or msgpack roundtrip. One fix closes both issues; the test surface covers both shapes.
- **The disconfirmation in PR #1553 paid off.** Without ruling out the diff-layer hypothesis empirically first, the investigator would have continued probing diff.rs / parser.rs / 12-vdom-patch.js and never reached the thread-local-counter angle.
- **The regression-window mystery resolved.** "Why did v0.9.6rc2 work?" got a concrete answer: pre-#1538, msgpack deserialize failed silently → `state_backends/memory.py:118` returned None → full re-mount every event → no diff path → no collision. Post-#1538 deserialize succeeds → diff path runs → collision reachable. The bug was always latent; #1538 unmasked it.
- **3 new helpers are minimal API surface, well-tested.** `from_base62` / `max_djust_id_in` / `ensure_id_counter_at_least` are additive in `djust_vdom`'s public API. 12 unit tests cover round-trip, invalid input rejection, tree walk on id-less / single-id / nested trees, and monotonic semantics. The cross-helper invariant `ensure_id_counter + next_djust_id never collides with prior max` is explicitly pinned.

### What could be improved
- **Pre-commit `cargo fmt` bounced the first commit attempt** because the test file had a long array literal that fmt prefers to break across lines. Re-staged after explicit `cargo fmt` ran clean. Same pattern from #1545 retro — process note: run formatters explicitly BEFORE `git commit` for any new Rust/Python file, rather than relying on the pre-commit hook's first run to auto-fix.
- **The combined #1550+#1552 fix shipped on a single PR.** Defensible (they share root cause), but slightly murky review surface — a reviewer pinning #1552's behavior has to also reason about #1550's. If the bugs had been independent, separating them would have been better. The combined treatment is justified here BECAUSE they share root cause; otherwise default to separate PRs.
- **The fix is at the parse-entry boundary, not at the persistence boundary.** Alternative: ensure `serialize_msgpack` always serializes the counter alongside `last_vdom`, and `deserialize_msgpack` restores it on the new thread. That would handle thread handoff cleanly but msgpack of the counter scope (thread-local) is non-trivial. The chosen fix (advance-past-max at parse time) is simpler and works for both thread-handoff AND msgpack cases without requiring the counter to be serialized. Lower-risk; worth documenting as the design choice.

### Connection to milestone
This PR closes the v1.0.0rc6 post-rc6 drain (#1550, #1551, #1552 all closed; zero open issues remain). Combined with PR #1554 (#1551) and the test-pin PR #1553 (#1552 investigation), the drain is complete.

### Release-line implications
#1552's reporter has downgraded production to v0.9.6rc2. Recommend either cutting v1.0.0rc7 immediately (so the reporter can re-upgrade and dogfood before GA) OR going direct to v1.0.0 GA with this fix included. The maintainer call.

### Verdict
`RETRO_COMPLETE`

Issue #1557, opened 2026-05-21
feat(websocket): cache tenant per WS session (multi-tenant ASGI hot path)


johnrtipton:
## Context

Follow-up to #1556. That issue ships option **(c)** — a startup system check warning when `TENANT_LIMIT_SET_CALLS` is unset under ASGI+django-tenants — to catch the misconfiguration that caused a prod 503.

This issue tracks option **(a)** from #1556: caching the tenant per WS session at LiveView mount time so subsequent WS events (`tick_interval`, `push_to_view`, presence updates, `@notify_on_save` re-mounts) skip the `TenantMainMiddleware` `get_tenant` DB roundtrip.

## Why this needs its own design pass

The principle is correct — WSGI workers do the equivalent via process locality — but the failure modes are exactly the class djust's manifesto says to prevent. Per the #1556 design thread:

1. **Cross-tenant `search_path` leak via pooled connection.** Caching the tenant lookup must NOT skip `SET search_path` — only the redundant `get_tenant` DB roundtrip. The underlying Postgres connection from Django's pool may have served a different tenant between our two events. django-tenants already handles this via `TENANT_LIMIT_SET_CALLS`; the cache must compose with that, not bypass it.
2. **Tenant-source trust boundary.** Caching is only safe if the tenant is resolved from an immutable property of the WS handshake (`scope["headers"]` host header captured at connection accept time). Anything in the cache key derived from a client-controllable WS frame later is an attacker-flippable surface.
3. **Mount-time race.** Tenant must be resolved + stashed atomically at `connect()`, before any event handler can fire. A handler that races against an unpopulated cache picks up the wrong tenant.
4. **Cross-tenant isolation tests required.** Two simultaneous WS sessions on tenants T1 and T2 sharing the same threadpool's connection pool — assert no event from T1 ever sees T2's data.

## Acceptance criteria

- [ ] Threat model documented in the PR description (the 4 failure modes above, plus anything new the design surfaces).
- [ ] **Security review checkbox** ticked by a reviewer who has read the django-tenants connection-wrapping internals (`django_tenants/postgresql_backend/base.py:144`, the `set_tenant`/`SET search_path` interaction).
- [ ] Cross-tenant isolation tests: two concurrent WS sessions on different tenants, shared connection pool, assert no data leak.
- [ ] `SET search_path` still fires when the underlying connection's tenant differs from the cached tenant (i.e. cache replaces the lookup, not the search_path enforcement).
[...]

Issue #1177, opened 2026-04-28
tech-debt: programmatic post-stage hook enforcement for pipeline-template gates


johnrtipton:
Source: Stage 11 review of PR #1176 (closes #1173 + #1174) — 🟡 R1 follow-up.

## Context

PR #1176 added template-level mandatory checklist items for two-commit shape (Stage 5/9 forbid/require CHANGELOG.md hunks separately) and 3-clean-runs verification (bugfix Stage 6). The gates are 'soft' — the LLM executor reads the imperative action text and self-applies. There's no programmatic preprocessor that evaluates conditionals or intercepts file writes.

## Required change

Add executor-side post-stage hooks in `~/.claude/skills/pipeline-run/SKILL.md`:

1. **After Stage 5 commit**: `git show HEAD --name-only | grep CHANGELOG.md && fail` — fails the commit if CHANGELOG.md is staged in the implementation commit.
2. **After Stage 9 commit (feature) / Stage 5 commit (ship)**: `git show HEAD --name-only | grep -vE '^(CHANGELOG.md|docs/|\.pipeline-templates/.*README)' && fail` — fails if any non-docs file is staged in the docs commit.
3. **End of Stage 6 (bugfix, pollution-class)**: re-run full pytest suite up to 3 times in a loop; fail Stage 6 if any run fails. Replace the imperative checklist item with a programmatic loop in the executor.

## Scope

Skill file only (~30 LoC of bash + flow control). Not a djust-repo PR.

## Why deferred from PR #1176

#1176 shipped the soft gates as a substantial improvement on baseline. Hard programmatic enforcement is a separate skill-level change and doesn't block the v0.9.2 drain — the LLM executor reading the imperatives is sufficient for the autonomous-pipeline use case.

johnrtipton:
Fixed in local \`~/.claude/skills/pipeline-run/SKILL.md\` (skill-level work; not in version control — same pattern as #1080's closure).

## Diff applied

Added a new section **\"MANDATORY Post-Commit Programmatic Gates (#1177)\"** before the existing Action #122 post-commit verification section. Three gates:

1. **Gate 1 — Stage 5 commit must NOT touch CHANGELOG.md.** Bash one-liner: \`git show HEAD --name-only | grep -q '^CHANGELOG\\.md$' && exit 1\`. Enforces the two-commit shape (Action #181 / #1173).

[...]

PR #1558, opened 2026-05-21
feat(checks): add C014 — warn on multi-tenant ASGI without TENANT_LIMIT_SET_CALLS (#1556)

### LINKED ISSUE #1556: LiveView events + multi-tenant ASGI exhaust DB connection pool
## Summary

Running djust 1.0.0rc7 on djustlive (production: 2 replicas, Postgres `max_connections=200`, django-tenants for schema-per-tenant), we hit a production 503 caused by an ASGI thread-per-request DB connection storm. Captured live via py-spy:

```
Thread 2520 (idle): "ThreadPoolExecutor-1168_0"
    _configure_timezone        (django/db/backends/postgresql/base.py:356)
    _configure_connection      (django/db/backends/postgresql/base.py:373)
    init_connection_state      (django/db/backends/postgresql/base.py:401)
    connect                    (django/db/backends/base/base.py:258)
    ensure_connection          (django/db/backends/base/base.py:279)
    _cursor                    (django/db/backends/base/base.py:296)
    _cursor                    (django_tenants/postgresql_backend/base.py:144)
    cursor                     (django/db/backends/base/base.py:320)
    execute_sql                (django/db/models/sql/compiler.py:1621)
    get                        (django/db/models/query.py:631)
    get_tenant                 (django_tenants/middleware/main.py:29)
    process_request            (django_tenants/middleware/main.py:45)
    thread_handler             (asgiref/sync.py:561)
    run                        (asgiref/current_thread_executor.py:40)
    ...
```

The high executor index (`-1168_0`, `-1199_0`, `-1206_0` …) means each ASGI request was getting its **own** short-lived executor → its own thread → its own brand-new Postgres connection. Under LiveView event traffic (tick_interval polling, push_to_view, presence updates), the rate of new connections exceeded what Postgres could service, and pods went unready simultaneously → ingress 503.

## Why this is djust-shaped (not just django-tenants)

django-tenants existed long before djust. WSGI deployments don't see this because the WSGI worker process holds a connection across many requests. The combination that bites is **ASGI + django-tenants + high-frequency per-request DB hits**, and LiveView is what makes that frequency high:

- Every WS event handler runs through TenantMainMiddleware → `set_tenant()` → `SET search_path`
- Every `handle_tick` does the same on `tick_interval` cadence (commonly 1.5–5s)
- Every `push_to_view` re-mounts the view across all connected sessions → middleware again
- Every `@notify_on_save` listener that fires re-mounts the listening view → middleware again

A single user with a Projects page open running `tick_interval=3000` generates ~20 middleware passes/min, each opening a connection if persistent reuse isn't working. Multiply by active sessions, and the Postgres pool exhausts.

## Workaround we shipped

`TENANT_LIMIT_SET_CALLS = True` in settings — skip the `SET search_path` wire trip when the connection is already on the right tenant. Combined with dropping to 1 replica (since `max_connections=200` minus celery/minio overhead leaves ~120 slots, too tight for 2 web pods under this connection rate). It's a real reduction in SQL volume but it doesn't address the framework-level shape.

## Suggested framework-level options

I'd love feedback on which of these (or some combination) fits the djust mental model best:

**(a) Cache the tenant per WS session.**  
At LiveView mount time, resolve the tenant once and stash it on the consumer. Subsequent events on that connection skip the middleware DB lookup entirely — they already know what tenant they're for. The connection's `search_path` only needs to be set when the consumer hands off to a worker thread, not on every event. This is the highest-leverage change because LiveView is unique in re-entering middleware for every WS frame.

**(b) Document the pattern + ship a deployment recipe.**  
A `docs/guides/multi-tenant-asgi.md` page that calls out:
- Postgres `max_connections` needs to scale with `replicas × ASGI_THREAD_LIMIT`
- Recommend pgbouncer (transaction pooling) in front of Postgres for any multi-replica deploy
- `TENANT_LIMIT_SET_CALLS = True` belongs in every multi-tenant djust app's settings
- Or — explicitly call out that single-replica is the supported shape for now

**(c) Detect and warn at startup.**  
A djust system check (alongside the existing V004/T002/C012 family) that fires when:
- `TENANT_MODEL` is configured (django-tenants installed)
- AND `TENANT_LIMIT_SET_CALLS` is unset / False
- AND the ASGI handler is daphne or uvicorn
…with a clear "you're going to have a bad time at scale" hint. Same pattern as `STATE_BACKEND` warning at djust 0.9 (`Falling back to in-memory state backend in production`) which served us very well as an early signal.

**(d) Pool connections at the LiveView consumer level.**  
The LiveViewConsumer holds a per-session connection across the WS lifetime. Subsequent sync_to_async calls within the consumer reuse it. This avoids the threadpool churn entirely for LiveView traffic (HTTP requests still get the normal Django pattern). Heavier engineering lift; might require careful interaction with django-tenants' connection wrapping.

(a) + (c) feels like the highest-value low-cost combination — fix the hot path, warn future users away from the foot-gun.

## Repro environment

- djust 1.0.0rc7 (also reproduces back to 0.9.7 — the issue isn't new with 1.0)
- django-tenants ≥ 3.9
- Daphne / Channels in ASGI mode
- Postgres `max_connections=200`
- 2 web replicas behind nginx ingress, single Postgres backend
- Production traffic of ~1 active LiveView session + ~12 node heartbeats/min was enough to tip it

Happy to provide more py-spy captures or do follow-up testing on whichever direction the maintainers prefer.

## Cross-refs

djustlive PR fixing the immediate symptom: https://github.com/djust-org/djustlive/pull/377

🤖 Filed via Claude Code

johnrtipton:
## Summary

Closes #1556. Ships option **(c)** from the issue's design thread — a startup system check `djust.C014` that fires when an ASGI + django-tenants deploy is missing `TENANT_LIMIT_SET_CALLS = True`. This is the misconfiguration that caused djustlive's production 503: every WS event re-entering `TenantMainMiddleware` → `SET search_path` → fresh Postgres connection, exhausting the pool under LiveView load (`tick_interval`, `push_to_view`, presence, `@notify_on_save` re-mounts).

Option **(a)** (per-WS-session tenant cache) is filed as a separate v1.1.0 issue, #1557, with the `security-review` label — it needs a threat model + cross-tenant isolation tests before it's safe to ship. See the four documented failure modes in the #1556 thread + #1557 body (search_path leak via pooled connection, tenant-source trust boundary, mount-time race, isolation tests).

- New helper `_check_multi_tenant_asgi_set_calls` in `python/djust/checks.py`
- Fires when ALL of: `django_tenants in INSTALLED_APPS OR TENANT_MODEL set`, `ASGI_APPLICATION set`, `TENANT_LIMIT_SET_CALLS unset or False`
- Suppressible via `DJUST_CONFIG = {'suppress_checks': ['C014']}` (short or full ID)
- `fix_hint` contains the copy-pasteable `TENANT_LIMIT_SET_CALLS = True` line plus `max_connections` sizing + pgbouncer guidance

## Test plan

- [x] 11 regression cases in `python/djust/tests/test_c014_multi_tenant_asgi.py` across 4 classes
  - `TestC014TriggerConditions` — 3 positive triggers (INSTALLED_APPS, TENANT_MODEL, unset flag)
  - `TestC014NegativeCases` — 3 negatives (no django-tenants, flag=True, no ASGI)
  - `TestC014Suppression` — short ID + full ID
  - `TestC014HintQuality` — names the setting, references #1556, fix_hint is copy-pasteable
- [x] Gate-off self-test (Action #254 / #1468): with `if False:` wrapping the C014 call, 6 of 11 tests fail (the 3 trigger cases + 3 hint-quality cases). The 5 still-passing tests are the intentionally-tautology-safe negative cases (asserting C014 does NOT fire) — flagged in their docstrings.
- [x] `196 passed, 0 failed` on the broader `check/config`-pattern test sweep — no regressions from the `check_configuration` insertion
- [x] Inline-verify per pipeline-run skill (low-risk: 1 new helper, ~50 LOC, no public API surface, no security-class change)
- [x] Two-commit shape (impl+tests in `c45c9a46`, CHANGELOG in `a138e459`) — Gate 1 + Gate 2 verified
- [x] Branch-verify reflex confirmed before commits

## Security review (per `docs/PULL_REQUEST_CHECKLIST.md`)

This PR adds a read-only inspection of Django settings — no new attack surface:
- No `mark_safe`, no `format_html`, no template rendering
- No `print()`, no f-string logging
- No subprocess, no file I/O, no network
- No new imports
- All settings accesses are `getattr(..., default)` with safe defaults

The `security-review` label is reserved for #1557 (the option-a follow-up), which DOES touch security-relevant code paths.

## Refs

- Closes #1556 (production 503 incident — option c)
- Tracks #1557 (option a — per-WS-session tenant cache, v1.1.0, `security-review`)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- **New system check `djust.C014` — warn at startup when multi-tenant ASGI deploys omit `TENANT_LIMIT_SET_CALLS` (#1556, option c).** Surfaces the misconfiguration that caused a production 503 on djustlive: under ASGI + django-tenants, every WebSocket event re-enters `TenantMainMiddleware` → `set_tenant()` → `SET search_path`. LiveView amplifies this — `tick_interval` polling, `push_to_view` re-mounts, presence updates, and `@notify_on_save` listener re-mounts each re-enter the middleware. Without `TENANT_LIMIT_SET_CALLS = True`, every re-entry issues a fresh Postgres roundtrip; under load the Postgres pool exhausts and pods serve 503 simultaneously. The check fires when ALL of these hold: (1) `django_tenants` is in `INSTALLED_APPS` OR `TENANT_MODEL` is set, (2) `ASGI_APPLICATION` is set, (3) `TENANT_LIMIT_SET_CALLS` is unset or `False`. Emits a `DjustWarning` with the setting name, the originating issue (#1556), and a `fix_hint` containing the copy-pasteable `TENANT_LIMIT_SET_CALLS = True` line plus `max_connections` sizing guidance. Suppressible via `DJUST_CONFIG = {'suppress_checks': ['C014']}`. The framework-level fix — caching the tenant per WS session at LiveView mount time (option a from the issue) — needs a separate threat-model + cross-tenant isolation test pass and is tracked separately in #1557 (`security-review` label) for v1.1.0. New helper `_check_multi_tenant_asgi_set_calls` in `python/djust/checks.py`. Covered by 11 regression cases in `python/djust/tests/test_c014_multi_tenant_asgi.py` across 4 classes (trigger conditions, negative cases, suppression by short and full ID, and hint quality), including a gate-off self-test (#254 / #1468) confirming 6 of 11 tests fail without the check (the 5 negative cases are intentionally tautology-safe — they assert C014 does NOT fire when conditions aren't met).
+
 ## [1.0.0rc7] - 2026-05-20
 
 ### Fixed
--- a/python/djust/checks.py
+++ b/python/djust/checks.py
@@ -404,6 +404,76 @@ def _check_manual_client_js(errors):
                             pass  # Skip files that can't be read
 
 
+def _check_multi_tenant_asgi_set_calls(errors):
+    """C014 — django-tenants + ASGI without TENANT_LIMIT_SET_CALLS (closes #1556).
+
+    Under ASGI + django-tenants, every WS event handler runs through
+    TenantMainMiddleware → set_tenant() → SET search_path. LiveView amplifies
+    this: tick_interval polling, push_to_view re-mounts, and @notify_on_save
+    re-mounts each re-enter the middleware. Without TENANT_LIMIT_SET_CALLS,
+    every re-entry emits a Postgres roundtrip. Production replicas can
+    exhaust the Postgres pool and serve 503 simultaneously (#1556 captured
+    via py-spy on djustlive prod: ThreadPoolExecutor-1168_0, -1199_0,
+    -1206_0 each holding a fresh connection).
+
+    Fires when ALL of the following hold:
+    - django-tenants is configured (django_tenants in INSTALLED_APPS OR
+      TENANT_MODEL setting is set)
+    - ASGI_APPLICATION is set (the deployment is ASGI-shaped)
+    - TENANT_LIMIT_SET_CALLS is unset or False
+
+    Suppress with DJUST_CONFIG = {'suppress_checks': ['C014']} if the app
+    deliberately tolerates the per-request SET search_path cost (e.g., low
+    concurrency, no LiveView polling, or a hardened pgbouncer in front).
+
+    Tracking #1557 for the framework-level fix (per-WS-session tenant cache).
+    """
+    from django.conf import settings
+
+    if _is_check_suppressed("djust.C014"):
+        return
+
+    installed = list(getattr(settings, "INSTALLED_APPS", []))
+    has_tenants_app = "django_tenants" in installed
+    has_tenant_model = getattr(settings, "TENANT_MODEL", None) is not None
+    if not (has_tenants_app or has_tenant_model):
+        return
+
+    if not getattr(settings, "ASGI_APPLICATION", None):
+        return
+
+    if getattr(settings, "TENANT_LIMIT_SET_CALLS", False):
+        return
+
+    errors.append(
+        DjustWarning(
+            "Multi-tenant ASGI deploy without TENANT_LIMIT_SET_CALLS — "
+            "every WS event will emit a redundant `SET search_path` and "
+            "can exhaust the Postgres connection pool under LiveView load.",
+            hint=(
+                "Set `TENANT_LIMIT_SET_CALLS = True` in settings. This is a "
+                "django-tenants flag that skips the `SET search_path` wire "
+                "trip when the connection is already on the right tenant. "
+                "Under djust, every WebSocket event (tick_interval, "
+                "push_to_view, presence updates, @notify_on_save) re-enters "
+                "TenantMainMiddleware; without this flag, each re-entry "
+                "issues a Postgres roundtrip. See #1556 for the prod "
+                "incident that motivated this check, and #1557 for the "
+                "tracked framework-level fix (per-WS-session tenant cache). "
+                "Suppress with DJUST_CONFIG = {'suppress_checks': ['C014']}."
+            ),
+            id="djust.C014",
+            fix_hint=(
+                "Add `TENANT_LIMIT_SET_CALLS = True` to your Django "
+                "settings file. Also size Postgres `max_connections` for "
+                "`replicas × ASGI_THREAD_LIMIT` and consider a "
+                "transaction-pooling pgbouncer in front for any "
+                "multi-replica multi-tenant deploy."
+            ),
+        )
+    )
+
+
 def _get_project_app_dirs():
     """Return directories for project apps (excluding third-party and djust itself)."""
     from django.apps import apps
@@ -612,6 +682,9 @@ def check_configuration(app_configs, **kwargs):
 
     # C013 -- Stale collectstatic copy of client.min.js (closes #1088)
     _check_stale_collected_client(errors)
+
+    # C014 -- Multi-tenant ASGI without TENANT_LIMIT_SET_CALLS (closes #1556)
+    _check_multi_tenant_asgi_set_calls(errors)
 
     # C005 -- WebSocket routes missing AuthMiddlewareStack
     # A001 -- WebSocket routes missing AllowedHostsOriginValidator (#659)
--- a/python/djust/tests/test_c014_multi_tenant_asgi.py
+++ b/python/djust/tests/test_c014_multi_tenant_asgi.py
@@ -1,0 +1,204 @@
+"""
+Regression tests for #1556 — djust.C014 detects multi-tenant ASGI deploys
+missing TENANT_LIMIT_SET_CALLS.
+
+Under ASGI + django-tenants, every WS event re-enters TenantMainMiddleware
+and emits a redundant `SET search_path`. Without TENANT_LIMIT_SET_CALLS,
+LiveView traffic (tick_interval, push_to_view, presence, @notify_on_save)
+can exhaust the Postgres connection pool. C014 fires at startup when the
+3 trigger conditions are all met.
+
+Note: tests patch the settings the check reads directly rather than going
+through ``override_settings(INSTALLED_APPS=...)``. Adding a string like
+``"django_tenants"`` to INSTALLED_APPS via override_settings triggers
+Django's app loader to actually import the module; we'd then need
+django-tenants as a test dependency just to assert what the check does
+on its absence/presence of the string. Direct patching keeps the check
+contract observable without the dependency hop.
+"""
+
+from djust.checks import check_configuration
+
+
+def _ids(errors):
+    return {getattr(e, "id", "") for e in errors}
+
+
+def _settings(**overrides):
+    """Patch django.conf.settings attributes for the duration of a `with`.
+
+    Each kwarg sets/replaces an attribute on `settings`; on exit the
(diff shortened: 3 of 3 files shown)


johnrtipton:
**Verdict:** APPROVE

Clean implementation of option (c) from #1556. The check matches the existing C-category convention (`_is_check_suppressed` early-return + `DjustWarning(hint=..., id=..., fix_hint=...)` shape), is wired into `check_configuration` next to C013 (`python/djust/checks.py:686`), and the test coverage hits every branch of the new helper. Branch is `2 ahead, 0 behind` `origin/main` (stale-base check, Action #1450 / #250 — clean). The gate-off self-test (Action #254 / #1468) was applied honestly and the arithmetic checks out: 3 trigger tests + 3 hint-quality tests = 6 fail when gated off; 3 negative + 2 suppression = 5 stay green (those are intentionally tautology-safe per their assertions of "C014 does NOT fire"). Security surface: read-only `getattr` on Django settings, no `mark_safe` / no f-string logging / no `print()` / no template rendering — confirmed via `git show c45c9a46 -- python/djust/checks.py | grep -E "mark_safe|format_html|logger\.|f['\"]|print\("` returning zero hits.

## Findings

### 🔴 Must-fix (block merge)

None.

### 🟡 Should-fix (file as follow-up if not addressed pre-merge)

None.

### 🟢 Nits (informational)

- `python/djust/checks.py:436` — `list(getattr(settings, "INSTALLED_APPS", []))` casts to a list when membership-testing is the only use. `"django_tenants" in installed` works directly on tuples and lists; the `list()` is defensive against generators but `INSTALLED_APPS` is normatively a tuple/list in Django settings. Not worth changing on its own.
- `python/djust/tests/test_c014_multi_tenant_asgi.py:18` — the test module's custom `_SettingsPatcher` (mutates `django.conf.settings` directly via `setattr`/`delattr`) instead of `django.test.override_settings`. The docstring at line 8-15 justifies this (avoid forcing django-tenants as a test dependency just to put the string `"django_tenants"` in `INSTALLED_APPS`). The justification is sound, but worth noting that this style does NOT compose with `override_settings` if a future test in the same suite wraps these. Not a regression today.
- The two suppression tests (`test_suppressed_by_short_id`, `test_suppressed_by_full_id`) pass under gate-off because their assertion shape is `"djust.C014" not in errors`. They are tautology-safe on the *check-enable* axis but they DO meaningfully verify suppression: the matching trigger tests at lines 73-87 share identical settings (sans `DJUST_CONFIG`) and fire C014, so the suppression tests prove the mechanism by contrast. Good as-is; just noting the structural reason they survive gate-off.

## What I checked

- **Trigger-condition AND combination correctness:** the three early-return gates at `python/djust/checks.py:439-446` correctly compose as logical AND (any miss → no warning). Reproduced mentally against the 8 possible combinations of (tenants signal × ASGI × flag); only the (yes × yes × no) corner emits, which matches the issue body and the design discussion in #1556.
- **Tenant-detection signal completeness:** `"django_tenants" in INSTALLED_APPS OR TENANT_MODEL is not None`. Verified against the django-tenants required-config surface: a django-tenants install REQUIRES `TENANT_MODEL`, and `django_tenants` is required in `SHARED_APPS` (which is composed into `INSTALLED_APPS`). `SHARED_APPS` / `TENANT_APPS` are themselves only set when django-tenants is installed, so they don't add a 3rd independent signal that the current OR check would miss. The two-signal OR is the canonical detection.
- **ASGI-detection signal:** `ASGI_APPLICATION` is set. A mixed deploy (WSGI runserver + ASGI Channels) typically still sets `ASGI_APPLICATION` for Channels routing, which is what triggers the production hot-path the issue describes. False-positive risk is low (a developer who sets `ASGI_APPLICATION` defensively but runs WSGI in prod gets a noisy warning, but the fix — `TENANT_LIMIT_SET_CALLS = True` — is harmless under WSGI too). Acceptable.
- **Branch coverage in `_check_multi_tenant_asgi_set_calls`:** every early-return path (suppressed, no tenants signal, no ASGI, flag True) has at least one negative-case test. The single positive-emit path has 3 trigger-condition tests + 3 hint-quality tests. All four explicit branches at `python/djust/checks.py:433-446` are exercised.
- **Gate-off arithmetic:** mentally re-ran with `_check_multi_tenant_asgi_set_calls` returning early. Trigger tests assert `"djust.C014" in _ids(errors)` → fail (3). Negative tests assert `"djust.C014" not in _ids(errors)` → pass (3). Suppression tests assert `"djust.C014" not in _ids(errors)` → pass (2). Hint-quality tests call `next(e for e in errors if id == "djust.C014")` → raise `StopIteration` → fail (3). Total: 6 fail / 5 pass. Matches the PR-body claim.
- **CHANGELOG test-count accuracy (Action #1049):** counted test methods directly — `pytest --collect-only` would report 11 (3+3+2+3). CHANGELOG line says "11 regression cases" — accurate.
- **Cross-test pollution risk:** `grep -l "TENANT_LIMIT_SET_CALLS\|TENANT_MODEL" python/djust/tests/ -r` finds only `test_c014_multi_tenant_asgi.py`. The defensive `delattr/setattr` blocks in `test_fires_when_flag_unset` (lines 96-104) and `test_does_not_fire_without_django_tenants` (lines 124-133) guard against ordering issues even within the file.
- **Security separation:** option (a) correctly deferred to #1557 with the `security-review` label. The C014 hint at `python/djust/checks.py:464` mentions suppression but routes users to the *correct* fix (`TENANT_LIMIT_SET_CALLS = True`) rather than to skipping the warning for performance reasons. The `fix_hint` block additionally mentions `max_connections` sizing + pgbouncer — those are the right operational levers, not security-relevant performance shortcuts.
- **Stage 11 stale-base check (Action #1450 / #250):** `git fetch origin && git rev-list --count HEAD..origin/main` returns `0`. Branch is 2 commits ahead, 0 behind. Safe to review against current main.
- **Two-commit shape (Action #181 / #1173):** impl+tests in `c45c9a46`, CHANGELOG in `a138e459`. Verified via `git show` diff scope.
- **Style/security greps:** `mark_safe`, `format_html`, `logger.`, f-string, `print(` — zero hits in the diff.

Approving. Recommend merging once CI is green.

johnrtipton:
## Stage 14 retro — PR #1558 (#1556 option c)

### What went well
- **Pre-flight scope conversation prevented a security misfire.** The first instinct on #1556 was the issue body's own recommended (a)+(c) combo. The "is a+c secure?" pushback surfaced 4 concrete failure modes for option (a) — cross-tenant `search_path` leak via pooled connection, tenant-source trust boundary, mount-time race, missing cross-tenant isolation tests — that no amount of post-hoc Stage 11 review could have caught against a shipped-then-redesigned PR. Splitting into (c) here + #1557 follow-up with `security-review` label is the right shape for this risk profile.
- **Gate-the-change-off self-test (Action #254 / #1468) at Stage 5.** Made tautology-safety legible: 6/11 fail with `if False:` wrap, 5/11 pass because they're negative cases (asserting C014 does NOT fire), and the 5 are flagged in their docstrings. The reviewer subagent independently verified the 6/5 arithmetic.
- **Two-commit shape + gates (#1173, programmatic gates #1177)** ran clean first-pass. Gate 1 (impl commit has no `CHANGELOG.md`) and Gate 2 (docs commit is docs-only) both green.
- **Inline-verify shortcut** (low-risk: ~50 LOC helper, no public API, no security-class change) saved the 3-subagent spawn for Test Execution / Self-Review / Security Check, ~10 min wall-clock per the skill's note. Reviewer subagent at Stage 11 confirmed the inline pass was complete.

### What could be improved
- **`override_settings(INSTALLED_APPS=...)` triggers Django's app loader.** First-pass tests used `@override_settings(INSTALLED_APPS=["djust", "django_tenants"])`, which Django then tries to actually import — 9 of 11 tests failed with `ModuleNotFoundError: No module named 'django_tenants'`. The fix (a custom `_SettingsPatcher` that patches `settings` attributes directly) is documented in the test module's docstring so the next person testing a check against a `string-in-INSTALLED_APPS` signal doesn't repeat the dependency-hop trap. Worth canonicalizing alongside Action #1109 (test fixtures with class-varying state — dynamic subclass, not class mutation) as a "test what the check inspects, don't reproduce the inspector's host environment" rule for system-check tests.
- **The `Warning: 1 uncommitted change` from `gh pr create`** is the untracked `scripts/antigravity_pipeline_run.py` sitting in the working tree. Pre-existing, unrelated to this PR — but it's the kind of low-signal warning that desensitizes readers to higher-signal ones. Followup: either commit-or-gitignore that script in a separate housekeeping PR.

### Process verification (per pipeline-run skill canon)

- Branch-verify reflex ran before each commit (HEAD = `feat-1556-multi-tenant-asgi-check` matched state file)
- Post-commit `git log -1 --oneline` reflex ran after both commits — no swallowed commits (#122)
- Stale-base check (Action #1450) ran during Stage 11 — branch was 2 ahead, 0 behind `origin/main`
- Stage 11 Code Review verdict: **APPROVE** (zero 🔴, zero 🟡, three 🟢 nits, all addressed inline or accepted as structural)
- Stage 12 Re-Review: **skipped** per skill rule "skip iff Code Review produced zero 🔴 and zero 🟡 findings"
- Stage 13 Merge: `gh pr merge 1558 --squash --delete-branch --admin` (admin override on the `REVIEW_REQUIRED` block — autonomous-merge feedback memory)

### Related

- Closes #1556 (production 503 — option c shipped)
- Tracks #1557 for option (a) in v1.1.0 with `security-review` label

`RETRO_COMPLETE`

Issue #1562, opened 2026-05-21
feat(bug-capture): iter B — read-only replay viewer at /__djust__/replay/<blob>


johnrtipton:
## Context

Iter B of B7 (Time-travel sharable URLs). Iter A shipped the data shape + encoder/decoder + scrub hook + DEBUG-gate (see PR landing #1552-followup). This iteration adds the browser-based replay surface that closes the share→reproduce loop.

## Scope

- New built-in djust LiveView at `/__djust__/replay/<blob>` (URL-resolved via `djust.urls`)
- DEBUG-gated route registration (route not mounted in production unless `DJUST_BUG_CAPTURE_PROD_OPT_IN = True`)
- View decodes the `<blob>` via `djust.bug_capture.BugCapture.decode()`, displays:
  - `state_before` and `state_after` rendered as side-by-side JSON trees with a diff highlight
  - `vdom_patches` listed with op + path + html-or-text payload (rendered as `<pre><code>` escape-on-render — captured HTML is untrusted)
  - `event_name` + `scrubbed_fields` headers
  - "Copy as `djbug1.` URL" button to re-share the blob
- Strict read-only: NO dispatch path, NO handler invocation against captured state, NO state mutation. The route is observation-only.
- Renders captured strings via `escape()` — captured HTML can be malicious; only iframe `srcdoc` rendering for the captured DOM, with CSP `sandbox` attribute restricting scripts.
- Multi-tenant safety: the replay view does NOT honor the captured `tenant_id` for any database access. The view runs in the maintainer's tenant context (or no tenant at all in dev).
- Share button in the existing debug panel: when `time_travel_enabled = True` and DEBUG, add a button that calls `encode_view_state(self, scrub=...)` and copies the result to clipboard. Default scrub is configurable via `LIVEVIEW_CONFIG['bug_capture_default_scrub']`.

## Acceptance criteria

- [ ] `GET /__djust__/replay/<blob>` returns 200 in DEBUG with a valid blob; 400 with a malformed blob; 404 in production without opt-in.
- [ ] Captured HTML in `vdom_patches[].html` is escaped on render (XSS regression test).
- [ ] Replay route refuses to execute event handlers (regression test: a captured `event_name` does NOT cause the named handler to fire on the maintainer's view).
- [ ] Multi-tenant boundary: captured `tenant_id` field is shown but never used to scope database queries (regression test).
- [ ] Share button in debug panel produces a valid `djbug1.<base64>` blob; clipboard contents verified.
[...]

Issue #1561, opened 2026-05-21
feat(bug-capture): iter C — Redis store, djust replay CLI, framework-level PII scrub


johnrtipton:
## Context

Iter C of B7. Iter A shipped the data shape + encoder/decoder + scrub hook (see PR landing #1552-followup). Iter B will add the browser replay viewer (tracked separately). This iteration completes B7 with the operational polish needed for sustained use.

### 1. Redis-backed snapshot store for large payloads

- `djust.bug_capture.RedisSnapshotStore` — implements a `SnapshotStore` ABC that iter A or iter B will introduce.
- When the encoded blob exceeds an inline-size threshold (~1.5 KB base64), the encoder uses the configured store and produces `djbug1.store.<opaque-id>` instead of `djbug1.<base64>`.
- TTL-bounded (default 1 hour). Opaque IDs are `secrets.token_urlsafe(16)` (128 bits of entropy).
- Configurable via `LIVEVIEW_CONFIG['bug_capture_store']` setting (defaults to in-memory; opt-in to Redis).
- Auth-required Redis: store refuses to attach to unauthenticated Redis instances by default (defensive against naively-exposed Redis).

### 2. `djust replay` CLI

- `djust replay <blob>` — opens the replay URL in the local browser via `webbrowser.open()`.
- `djust replay --inspect <blob>` — terminal-only, prints the decoded `state_before` / `state_after` / `vdom_patches` to stdout, suitable for piping to `jq` / `diff`.
- `djust replay --diff <blob>` — shows a unified diff between `state_before` and `state_after`.

### 3. Framework-level PII scrub via class attribute

- `LiveView.time_travel_excluded_fields: list[str] = []` — declarative scrub list.
- `encode_view_state()` reads the class attribute and applies the equivalent of `scrub_fields(*excluded)` BEFORE any caller-supplied `scrub` callable.
- Existing `scrub` parameter still works for arbitrary callables; the class attribute is the default safety net.
- The class attribute is enforced by `djust check` (new system check `V012`): warns if `time_travel_enabled = True` and the view has model fields or form fields whose names match common-PII patterns (`password`, `ssn`, `credit_card`, `tax_id`, `email`, `phone`) AND those names are NOT in `time_travel_excluded_fields`.

### 4. Iter B follow-on polish (if not done in iter B)

- Per-share short-link service — replace inline base64 blobs with `https://djust.dev/replay/<id>` shortlinks backed by the store. Optional; deployers' choice.

[...]

PR #1563, opened 2026-05-21
feat(bug-capture): BugCapture data shape + URL encode/decode + scrub hook (B7 iter A, refs #1552)

johnrtipton:
## Summary

Promotes **B7 — Time-travel sharable URLs** ("Replay.io for djust") from "killer demo idea" in the [v1.1 readiness session](docs/strategy-sessions/2026-05-19-v1.1-readiness.md) to a load-bearing v1.1.0 capability, triggered by the [#1552 reporter's data point](https://github.com/djust-org/djust/issues/1552) about upstream-bug-velocity friction.

**The motivation:** #1552 reopened on 2026-05-21 after a fix in rc7 didn't address the include-swap case. The reporter's own framing was *"the gap between 'I see it broken' and 'you can see it broken' is the full source tree."* The v1.1 readiness session recommended Path E (defer the headline-path decision until launch-soak data exists) with the explicit hedge *"refuse to commit before data exists."* The #1552 filing supplied that data.

**The reframe during scope conversation:** original sketch was a full timeline buffer + inline-or-store payload routing + replay LiveView. After clarifying with the user, the iter A scope collapsed to the minimum useful artifact: **just `state_before`, `state_after`, and `vdom_patches`**, with URL-fragment encoding. The store / replay viewer / CLI are split into iters B/C as separate v1.1.0 tracking issues (#1561, #1562).

## What this PR ships (iter A)

New module `python/djust/bug_capture.py`:

- **`BugCapture` dataclass** — the 3 minimal fields (`state_before`, `state_after`, `vdom_patches`), plus `event_name` for reviewer context and `scrubbed_fields` for wire-visible-transparency of PII scrubbing.
- **`BugCapture.encode(scrub=None) -> str`** — produces a `djbug1.<base64-urlsafe>` versioned wire format.
- **`BugCapture.decode(blob: str) -> BugCapture`** — defensive decoder, validates everything (types, version, fields, base64 alphabet, JSON shape).
- **`encode_view_state(view, event_name="", scrub=None)`** — convenience pulling the latest event snapshot + VDOM patches from a view with `time_travel_enabled = True`.
- **`scrub_fields(*names)`** — ready-made scrub callable for the common case (`encode(scrub=scrub_fields("password", "ssn"))`).

New docs page `docs/website/guides/bug-capture.md` linked from `_config.yaml` and `index.md` (per `feedback_docs_website_nav` memory).

## Security model — load-bearing

After the user asked *"is this a security concern?"* mid-implementation, the design was rebuilt with security as a day-one API constraint, not a deferred feature. The 4 hardenings:

1. **`scrub` callable is day-one API**, not deferred to iter C. Ready-made `scrub_fields(*names)` for the common case; user-supplied callable for arbitrary policies. Scrubbed field names (not values) recorded on the wire so reviewers can verify what was held back before further sharing.
2. **Module docstring leads with a 3-paragraph "READ THIS BEFORE USING" warning** explicitly listing PII risk + not-authenticated + default-off-in-prod + always-use-scrub.
3. **DEBUG-gated at the encoder** — `encode()` raises `RuntimeError` in production unless `DJUST_BUG_CAPTURE_PROD_OPT_IN = True` (the literal Python `True`, not just truthy — defensive against typo-enable via a config-loader workaround).
4. **No-pickle wire format pinned by a regression test** (`test_wire_format_is_not_pickle`) so a future maintainer reaching for pickle for "efficiency" trips immediately.

Wire format details:

- `djbug1.<base64-urlsafe>` with versioned outer prefix AND versioned inner `"v"` envelope; decoder rejects mismatches.
- `base64.b64decode(..., validate=True)` against the urlsafe alphabet (the relevant detail: `urlsafe_b64decode` doesn't accept `validate=`, so we call `b64decode` with `altchars=b"-_"` for the same alphabet but with strict validation).
- Required-field check + per-field type check on decode (`state_before`/`state_after` must be dicts; `vdom_patches` must be a list).
- 16 untrusted-input regression cases (non-string, missing-prefix, unknown-outer-version, malformed-base64, base64-to-non-JSON, non-object-JSON, inner-version-mismatch, missing-fields, wrong-field-types, no-pickle-pin).

## Test plan

- [x] **34 regression cases** in `python/djust/tests/test_bug_capture.py` across 5 classes (TestRoundTrip 6, TestScrub 6, TestDebugGate 4, TestUntrustedInput 10, TestEncodeViewState 8). All passing.
- [x] **Gate-off self-test** (Action #254 / #1468): `_enforce_prod_gate()` wrapped in `if False:` — confirmed 2 of 4 DEBUG-gate tests fail (the 2 that expect `RuntimeError`). The other 2 are intentionally tautology-safe: `test_encode_allowed_when_prod_opt_in_is_true` exercises the bypass path (no failure if gate disabled); `test_decode_works_regardless_of_debug` tests decode which is gate-independent.
- [x] **Inline-verify** per pipeline-run skill (low-risk: 1 new module ~370 LOC, no public-API change to existing surface, no security-class change beyond the new module's own defensive boundary).
- [x] **Two-commit shape** (#181): impl+tests in `33c7af5d`, docs+CHANGELOG in `33b20170`. Gate 1 (impl ≠ CHANGELOG) and Gate 2 (docs commit only contains docs) both verified.
- [x] **Branch-verify reflex** confirmed before each commit.

## Iter B / iter C — tracked as separate v1.1.0 issues

- **[#1561 — iter B](https://github.com/djust-org/djust/issues/1561):** Read-only replay viewer at `/__djust__/replay/<blob>` + share button in the debug panel. DEBUG-gated route, XSS-safe rendering of captured HTML (escape + iframe sandbox), multi-tenant safety (captured `tenant_id` never used to scope DB queries), strict read-only (no handler dispatch).
- **[#1562 — iter C](https://github.com/djust-org/djust/issues/1562):** Redis snapshot store for payloads >~1.5 KB; `djust replay` CLI for terminal-first workflows; framework-level `LiveView.time_travel_excluded_fields` class attribute with new `djust check` V012 system check for common-PII-pattern enforcement.

## Refs

- Refs #1552 (motivating data point — the upstream-bug-velocity friction this feature collapses from hours to one shared URL)
- Refs #1561 (iter B — replay viewer)
- Refs #1562 (iter C — Redis store + CLI + framework-level PII scrub)
- [docs/strategy-sessions/2026-05-19-v1.1-readiness.md](docs/strategy-sessions/2026-05-19-v1.1-readiness.md) — the readiness session whose "refuse to commit before data exists" hedge the #1552 filing satisfied


johnrtipton:
## Verdict: REQUEST_CHANGES

Strong PR — security model is genuinely load-bearing, two-commit shape is clean (impl `33c7af5d` excludes CHANGELOG; docs `33b20170` is docs-only), branch is up-to-date with `origin/main` (0 commits behind), tests count out exactly as claimed (34 total: TestRoundTrip 6 / TestScrub 6 / TestDebugGate 4 / TestUntrustedInput 10 / TestEncodeViewState 8 — all match CHANGELOG), and the gate-off arithmetic claim ("2 of 4 DEBUG-gate tests fail") is accurate (test_encode_refuses + test_prod_opt_in_must_be_literal_true both expect `RuntimeError`; test_encode_allowed and test_decode_works exercise non-gate paths). One 🔴 (the load-bearing one — `encode_view_state()` reads framework attributes the framework never writes), one 🟡 (stale-placeholder URL in docs), and a handful of 🟢s.

### 🔴 Must-fix

1. **`encode_view_state()` reads attributes the framework never writes (`python/djust/bug_capture.py:354`).** `_extract_last_patches()` loops over `("_last_vdom_patches", "_last_patches")` looking for the most recent VDOM patches on the view. Exhaustive grep across `python/djust/`, `python/tests/`, `tests/`, `examples/`:

   ```
   $ grep -rn "_last_vdom_patches\|_last_patches" python/djust/ python/tests/ tests/ examples/
   python/djust/bug_capture.py:273         (reads)
   python/djust/bug_capture.py:343         (reads)
   python/djust/bug_capture.py:354         (reads)
   python/djust/tests/test_bug_capture.py:312   (sets it on a stub view)
   ```

   **No framework code writes either name.** `render_with_diff()` returns `(html, patches, version)` as a tuple at every call site (`python/djust/websocket.py:928`, `runtime.py:566`, `sse.py:285`, etc.) — patches flow straight into the WS frame without being stashed on the view. The result: `encode_view_state()` will raise `ValueError("view has no recent VDOM patches recorded")` on every real LiveView in production. Only the stub `_StubView` in the test file (which manually assigns `self._last_vdom_patches = patches` at `tests/test_bug_capture.py:312`) actually exercises a populated path.

   This is the Action #1101-class "test-file coverage of an integration that doesn't exist in production" failure. Stage 4 plan should have grep-verified the attribute name against framework call sites the same way Action #1071 mandates for prose docs and Action #1197 mandates for canon-doc citations.

   **Fix options** — all valid; pick one explicitly:
   - **(a)** Add a stash hook in `WebsocketConsumer.handle_event` (the canonical render path post-event) and in `LiveViewConsumer` mount: `self.view_instance._last_vdom_patches = patches` immediately after each `render_with_diff()` call. Gate it behind `time_travel_enabled` to avoid permanent memory hold on views that don't need capture. Add a regression test that round-trips via a real `WebsocketCommunicator` (à la PR #1466's gate-off canon test) — NOT a stub view — so this stays load-bearing.
   - **(b)** Reframe iter A as "data shape + encoder + decoder ONLY" and defer `encode_view_state(view)` to iter B (where the consumer is the replay viewer, which can pass patches explicitly). Strip `encode_view_state()` + `_extract_last_patches()` from this PR; the 8 `TestEncodeViewState` cases become 0; CHANGELOG drops to 26 cases. The `BugCapture(state_before=…, state_after=…, vdom_patches=…)` constructor still works for the reporter's intended use (build by hand in a REPL).
   - **(c)** Pull patches from where they already exist — the time-travel buffer's `EventSnapshot` doesn't carry patches today, but iter A could extend `EventSnapshot` with an optional `vdom_patches: list[dict] | None = None` field and have `record_event_end` populate it. More framework surface but the cleanest model.

   Option (a) or (c) is preferred; (b) is the smallest-blast-radius fix if the PR needs to land before strategy session.

### 🟡 Should-fix

2. **Stale `XXXXXX` placeholder in published docs (`docs/website/guides/bug-capture.md:157`).** The Strategy Connection section links `https://github.com/djust-org/djust/issues/1552#issuecomment-XXXXXX` — `XXXXXX` is unreplaced. Per Action #1071 (doc-claim citation discipline for prose docs with external citations), every `#issuecomment-N` reference must be grep-verified before commit. Either replace with the real comment ID (the relevant comment is the reporter's 4th in #1552 that frames *"the gap between 'I see it broken' and 'you can see it broken'"* — `gh api repos/djust-org/djust/issues/1552/comments --jq '.[3].id'` will surface it), or remove the `#issuecomment-` anchor entirely and let the link land on the issue page.

### 🟢 Nits

3. **`scrub_fields()` is top-level-only — call this out in the docstring.** `scrub_fields("password")` removes top-level `state_before["password"]` but NOT `state_before["user"]["password"]`. Module docstring talks about "PII" abstractly, but the `scrub_fields` docstring (`bug_capture.py:316-328`) doesn't say "top-level only." Real-world `LiveView` state is often nested (forms, model-bound widgets) — a reader who reaches for `scrub_fields("ssn")` on a view with `state["customer"]["ssn"]` will leak. One-line addition: `"Note: only top-level keys of state_before/state_after are scrubbed. For nested state, supply a custom callable (see the Quick Start guide)."` Iter C's `time_travel_excluded_fields` (per #1561) should make the recursive-scrub semantics explicit too.

4. **No size cap on the encoded blob (`bug_capture.py:147`).** `json.dumps` + base64 will happily produce a 10 MB blob if `state_before` is huge (e.g., a view holding a large queryset). The iter B replay viewer's URL-fragment delivery will hit browser URL caps (~64 KB Chrome, ~2 KB IE compat) — that's iter B/C's problem to solve via Redis store (#1561's first scope item), but an iter-A defensive cap on encoded size with a clear `ValueError("BugCapture too large for inline encoding; iter C Redis store required")` would (a) protect against accidental DoS-via-share and (b) lock in the inline-vs-store boundary iter C will need. Not blocking — flagging for iter C planning.

5. **`_enforce_prod_gate()` reads settings at call time (`bug_capture.py:333-348`) — correct, but the gate-off self-test claim is more nuanced than CHANGELOG says.** The TOCTOU window between `_enforce_prod_gate()` returning and `payload = json.dumps(...)` executing is microseconds — no realistic settings-reload race. The literal-True check (`getattr(settings, ...) is True`) is correctly implemented (line 345 uses `is True`, not `==`); `TestDebugGate::test_prod_opt_in_must_be_literal_true` pins it. Both correct, just calling them out for record.

6. **Iter B/C wire-format compatibility looks OK.** The 3-field shape (`state_before`, `state_after`, `vdom_patches`) + the versioned `djbug1` envelope leaves room for iter B to render without breaking iter A blobs, and #1562 (Redis store) can layer `djbug1.store.<id>` as a sibling format without modifying `djbug1.<base64>` semantics. The `scrubbed_fields` wire-visible names give iter B the data it needs for the "what was held back" UI. Good forward-compat surface.

7. **`b64decode(..., validate=True)` is correctly defended (`bug_capture.py:181`).** The comment block explaining why `urlsafe_b64decode` can't be used (no `validate=` parameter) is helpful canon for future maintainers. Untrusted-input tests cover all the documented `ValueError` shapes (non-string, missing prefix, unknown version, malformed base64, non-JSON, non-object, inner-version-mismatch, missing fields, wrong field types, no-pickle pin = 10/10).

8. **No `mark_safe`, `format_html`, `print()`, `pickle`, `eval`, `exec`, `@csrf_exempt`, or f-string logger calls anywhere in the module.** Only `base64` + `json` imports. The single f-string at `bug_capture.py:145` builds the output URL fragment (`f"{WIRE_VERSION}.{b64}"`) — not a logger call, not interpolating untrusted data into HTML/JS; safe.

### What I verified

- **Two-commit shape (#181):** `33c7af5d` modifies only `python/djust/bug_capture.py` + `python/djust/tests/test_bug_capture.py` (no CHANGELOG → Gate 1 pass). `33b20170` modifies only `CHANGELOG.md`, `docs/website/_config.yaml`, `docs/website/guides/bug-capture.md`, `docs/website/index.md` (docs-only → Gate 2 pass).
- **Stale-base check (Action #1450):** `git fetch origin && git rev-list --count 33b20170..origin/main` → `0`. Branch is current; not stale.
- **Test count claim:** `git show 33c7af5d:python/djust/tests/test_bug_capture.py | grep -c "    def test_"` → `34`. Per-class enumeration: TestRoundTrip 6, TestScrub 6, TestDebugGate 4, TestUntrustedInput 10, TestEncodeViewState 8 — matches CHANGELOG and PR body.
- **Gate-off arithmetic:** With `_enforce_prod_gate()` wrapped in `if False:`, `test_encode_refuses_when_debug_is_false` (line 161 in tests) and `test_prod_opt_in_must_be_literal_true` (line 172) both fail because they `pytest.raises(RuntimeError)` on what becomes a successful call. `test_encode_allowed_when_prod_opt_in_is_true` (line 167) and `test_decode_works_regardless_of_debug` (line 184) stay green because the former tests the bypass path and the latter only exercises decode. 2 of 4 — claim accurate.
- **Strategy-sessions citation (`docs/strategy-sessions/2026-05-19-v1.1-readiness.md`):** file exists at the cited path.
- **Iter B/C tracking issues (#1561, #1562):** both exist and the scope hand-off (replay viewer + Redis store + framework PII scrub) is well-articulated; iter A's API shape is forward-compatible.
- **No-pickle pin (`tests/test_bug_capture.py:265`):** asserts `not raw.startswith(b"\x80")` and re-parses as JSON. Solid.
- **Framework integration grep:** `_time_travel_buffer` confirmed real (`time_travel.py:143`, `websocket.py:4544,4647,4715,4789`, `live_view.py:488,500,506`); `_last_vdom_patches` / `_last_patches` NOT real — only `bug_capture.py` reads them and only the test stub writes one (see 🔴 #1).
- **Style scan:** no `mark_safe`, `format_html`, `print()`, `pickle`, `eval`, `exec`, `@csrf_exempt`, `f"…"` logger calls (`grep -n "logger\." python/djust/bug_capture.py` → empty; module has no logger at all).
- **`XXXXXX` placeholder:** confirmed unreplaced at `docs/website/guides/bug-capture.md:157`.

johnrtipton:
## Stage 12 Re-Review — Verdict: APPROVE_AFTER_FIXES

Fix-pass commit `99047c7d` cleanly addresses all three Stage 11 findings. Verified each item against the fix-pass diff; all gate checks (test count, two-commit shape inverted to single fix-pass, no orphan references) pass.

### Original findings — disposition

**🔴 `encode_view_state()` framework integration — ✅ addressed.**
- `encode_view_state(view, patches, event_name="", scrub=None)` — `patches` is now a required positional parameter (`python/djust/bug_capture.py:227-232`). Callers must supply patches; no view-attribute introspection.
- `_extract_last_patches(view)` is GONE; replaced by `_coerce_patches(patches)` at `bug_capture.py:376-393` which only normalizes the caller-supplied arg (JSON string OR list → `list[dict]`).
- Exhaustive grep confirms ZERO remaining references to `_last_vdom_patches`, `_last_patches`, or `_extract_last_patches` across `python/djust/`, `python/tests/`, `tests/`, `examples/`, `docs/`. The Action #1101 hazard is fully removed.
- `_StubView` (`test_bug_capture.py:308-316`) no longer sets `self._last_vdom_patches`; only exposes `_time_travel_buffer`. Docstring explicitly calls out the iter-A decoupling rationale.
- All `TestEncodeViewState` cases pass `patches=` explicitly (e.g. `test_bug_capture.py:336`, `:358`, `:380`).
- The two new boundary tests are present and well-named: `test_raises_on_malformed_patches_json` (line 393), `test_raises_on_patches_decoding_to_non_list` (line 399 — a bonus third case, not just the two claimed; pins the `"patches JSON must be an array"` branch), `test_raises_on_patches_wrong_type` (line 405). Net +3 EncodeViewState cases on this dimension; +2 net once we subtract the removed `test_raises_when_view_has_no_patches`.
- `test_raises_when_view_has_no_patches` is REMOVED (no longer reachable — the attribute path it exercised doesn't exist anymore).
- Quick Start (`docs/website/guides/bug-capture.md:33-43`) shows the canonical pattern: `_html, patches, _version = my_view.render_with_diff()` immediately before `encode_view_state(my_view, patches=patches, ...)`. API reference (`bug-capture.md:146`) updated to reflect new signature.

**🟡 Stale `XXXXXX` placeholder — ✅ addressed.**
- `grep -rn "XXXXXX" docs/website/guides/bug-capture.md python/djust/bug_capture.py python/djust/tests/test_bug_capture.py` returns NO matches.
- `bug-capture.md:165` now reads `[#1552 reporter's data point](https://github.com/djust-org/djust/issues/1552)` — bare issue link, no `#issuecomment-` anchor. Per Action #1071 this is the right call (rather than guess at a comment ID).

**🟢 `scrub_fields` top-level-only docs — ✅ addressed.**
- `bug_capture.py:321-329` adds a `**Scope:**` paragraph explicitly stating *"`scrub_fields` operates on TOP-LEVEL keys only"* with the worked example `state_before["user"]["password"]` is NOT removed by `scrub_fields("password")`. Notes the djust public-state convention rationale + points at iter C (#1562) for the framework-level declarative scrub.

### Gate checks

- **Test count consistency (Action #1049 — recount after fix-pass):** `grep -cE "^    def test_" python/djust/tests/test_bug_capture.py` → **36**. Per-class enumeration: TestRoundTrip 6, TestScrub 6, TestDebugGate 4, TestUntrustedInput 10, TestEncodeViewState 10. **All five class counts match the CHANGELOG claim exactly** (6/6/4/10/10 = 36). CHANGELOG line 12 now reflects 36 + the new test names; the test-count recount discipline was followed.
- **Commit subject scope-cross prefix:** `fix(bug-capture): caller-supplied patches; doc nits (PR #1563 Stage 11 fix-pass)` — `fix:` prefix correctly signals the cross-cutting impl + docs + CHANGELOG fix-pass scope. Single-commit shape is appropriate here per the v0.9.1 retro guidance (#181) — a fix-pass is exempt from the canonical impl/docs split because it's correcting a Stage 11 finding that spans both.
- **No regressions / no new dead code / no new public-API surface beyond what the must-fix required:**
  - `_coerce_patches` is the ONLY new function; it is exercised by 4 tests (the 1 happy-path-list test + 3 error branches).
  - No new module-level imports.
  - No new public symbols beyond the modified `encode_view_state` signature (which is the must-fix's required surface change).
  - No `mark_safe`, `format_html`, `print`, `pickle`, `eval`, `exec`, `@csrf_exempt`, f-string logger calls introduced.

### New findings

None blocking. One trivial observation for record:

**🟢 (nit, non-blocking) Commit-message-vs-implementation typo.** The commit message body claims the new signature is `encode_view_state(view, *, patches, event_name="", scrub=None)` (keyword-only via `*` separator), but the actual signature at `bug_capture.py:227-232` is `encode_view_state(view, patches, event_name="", scrub=None)` — `patches` is a regular positional parameter, not kwonly. Doesn't affect behavior since every call site uses `patches=` keyword form. Worth noting only for future log-archaeology accuracy; not a code change.

### Summary

All three Stage 11 findings addressed correctly. The 🔴 fix takes the cleanest of the three options I proposed (option b — caller-supplied patches, deferring the render-path coupling to iter B per #1561), which is the smallest-blast-radius landing that doesn't carry a Action-#1101 hazard forward. The 🟡 docs nit is fixed with a bare issue link (no fabricated comment ID — correct per Action #1071). The 🟢 docstring note is thorough.

`RETRO_COMPLETE` — recommend merge once CI is green.

johnrtipton:
## Stage 14 retro — PR #1563 (B7 iter A, refs #1552, ships v1.1.0 Path D foundation)

### What went well

- **Scope-conversation reframe caught over-engineering early.** Initial sketch was a full timeline buffer + inline-vs-store payload routing + replay LiveView, sized at 4-5 weeks per the brainstorm doc. User clarification (*"we may just need original state + new state + vdom diff"*) collapsed iter A to the minimum useful artifact — 3 fields + URL fragment. Iter B/C now have clean scope boundaries instead of an everything-at-once PR that would have taken weeks.
- **Mid-implementation security pause produced load-bearing hardenings.** The *"is this a security concern?"* question while iter A was half-written triggered a threat-model pass that promoted `scrub` from "iter C deferred" to day-one API, added the DEBUG-gate with literal-True opt-in, added the no-pickle wire-format pin, and rebuilt the docstrings with a 3-paragraph "READ THIS BEFORE USING" warning leading every consumer of the module. None of that would have been in iter A under the original "ship and refine later" trajectory.
- **Stage 11 reviewer caught the 🔴 by reading the framework, not just the diff.** The reviewer ran exhaustive grep for `_last_vdom_patches` / `_last_patches` across the entire codebase and confirmed no framework writer exists — this is exactly the Action #1101 failure mode (test-stub coverage hiding missing framework integration) that PR #1102 canonicalized. The fix-pass made `patches` an explicit parameter, eliminating the dead code path. Without the reviewer's grep-against-framework, iter A would have shipped a function that always raises `ValueError` on real LiveViews — the test-stub never exercised the production path because the stub itself was populating the missing attribute.
- **Gate-the-change-off self-test (#254 / #1468) honestly accounted for tautology-safe cases.** With `_enforce_prod_gate()` wrapped in `if False:`, exactly 2 of 4 DEBUG-gate tests failed (the two expecting `RuntimeError`); the other 2 stayed green because they intentionally exercise the bypass path or test decode (which is gate-independent). The CHANGELOG documents this rather than hiding it behind "all tests fail under gate-off" — Action #1200 (tautology test detection) applied honestly.
- **Two-commit shape (#181) gates fired clean.** Gate 1 (impl commit ≠ CHANGELOG.md) and Gate 2 (docs commit only contains docs) both green on the original split; the Stage 11 fix-pass appropriately broke the two-commit shape because the fix spanned impl, tests, docs, and CHANGELOG — and the `fix(...):` prefix correctly signaled the cross-cutting scope.

### What could be improved

- **Iter A almost shipped a function that didn't work — the test stub was patching the gap.** `_StubView.__init__` set `self._last_vdom_patches = patches`, which let every TestEncodeViewState case pass without ever exercising the framework attribute the production code claimed to read. This is precisely the Stage 7-vs-11 delta that PR #840 canonicalized: Stage 7 self-review verified the function returned the right shape; Stage 11 verified the function had a path to actually run in production. Future shape: when an Implementation stage adds a function that reads `view.SOMETHING`, Stage 7 self-review should mandatorily grep production code for the writer before concluding "tests pass means correct." Filed as a candidate action-tracker rule.
- **The disk-full event mid-pipeline (`/private/tmp` at 100%) interrupted the state-file update.** Bash tool calls hit `ENOSPC` briefly. Recovered by switching to the `Edit` tool for the state-file write. Flagged to the user; suggests the harness needs better resilience to host-disk pressure, OR the docs should guide users to keep `/private/tmp` clear before long pipeline sessions.
- **One 🟢 non-blocking in the re-review survived to merge** — the fix-pass commit message body says `encode_view_state(view, *, patches, ...)` (keyword-only) but the actual signature is regular-positional. Pure log-archaeology accuracy; no behavioral effect. Worth being more careful about copy-pasting signatures from planning notes into commit messages without a final `grep` against the actual code.
- **Strategy connection in docs may be aged out fast.** The bug-capture.md "Strategy Connection" section cites the v1.1 readiness session and the #1552 reporter's data point. That's correct at write-time but will look stale once v1.1.0 ships. Consider moving the strategy-connection prose into the v1.1.0 release-notes blog post (which has a natural end-of-life) rather than a per-feature docs page (which doesn't).

### Process verification (per pipeline-run skill canon)

- Branch-verify reflex ran before each commit (HEAD = `feat-b7-time-travel-urls-iter-a` matched state file)
- Post-commit `git log -1 --oneline` reflex ran after each commit — no swallowed commits
- Stale-base check (Action #1450) confirmed by reviewer at Stage 11 — 0 commits behind `origin/main`
- Two-commit shape (Action #181) gates verified for original `33c7af5d` + `33b20170` split
- Stage 11 Code Review verdict: **REQUEST_CHANGES** (1 🔴, 1 🟡, 1 🟢) — addressed in fix-pass `99047c7d`
- Stage 12 Re-Review verdict: **APPROVE_AFTER_FIXES** — all 3 original findings ✅ addressed; 1 new 🟢 non-blocking (cosmetic) accepted at merge
- Stage 13 Merge: `gh pr merge 1563 --squash --delete-branch --admin`

### Promotion signal

This PR materially promotes B7 ("Time-travel sharable URLs") from "killer demo idea" status in the [v1.1 readiness session](docs/strategy-sessions/2026-05-19-v1.1-readiness.md) to a load-bearing v1.1.0 capability. The strategy session explicitly hedged *"refuse to commit before data exists"*; the #1552 reporter's filing supplied the data. Iter B (#1561) and iter C (#1562) are the rest of the v1.1.0 commitment.

### Related

- Refs #1552 — parent (production VDOM-diff regression + the upstream-bug-velocity friction data point this feature collapses)
- Refs #1561 — iter B (read-only replay viewer at `/__djust__/replay/<blob>` + debug-panel share button)
- Refs #1562 — iter C (Redis snapshot store + `djust replay` CLI + framework-level `time_travel_excluded_fields` PII scrub + V012 system check)
- [docs/strategy-sessions/2026-05-19-v1.1-readiness.md](docs/strategy-sessions/2026-05-19-v1.1-readiness.md) — strategy session whose "refuse to commit before data exists" hedge this PR satisfies

`RETRO_COMPLETE`

PR #1560, opened 2026-05-21
docs(multi-tenant): deprecate django-tenants integration; rewrite strategy guide + C014 messaging (#1556 follow-up)

johnrtipton:
## Summary

Follow-up to #1556 (the production 503 incident on a django-tenants + ASGI deploy) and to the in-thread design conversation that landed on a user direction: **django-tenants integration is deprecated** as a multi-tenancy strategy for djust applications. djust's own `djust.tenants` (row-level) is the supported and recommended path, and the framework now signals this clearly both in docs and in the in-product `djust.C014` system check.

This PR is documentation-shaped (no behavior change). It strengthens the messaging from the earlier soft "consider djust.tenants" framing to hard deprecation framing.

## Changes

**In `docs/website/guides/multi-tenant.md`:**
- New "Choosing Your Multi-Tenancy Strategy" section at the top of the guide (replaces the soft version added earlier in this branch's first iteration)
- `djust.tenants` subsection: framed as THE supported path; explains why row-level fits LiveView's hot path
- `django-tenants` subsection: leads with `> **Deprecated.**` blockquote, lists three concrete reasons (prod footgun under ASGI+LiveView, scope mismatch with djust mixins/backends/presence, straightforward migration path for typical apps), and points readers at the migration tracker
- Stopgap settings (`TENANT_LIMIT_SET_CALLS = True` + Postgres `max_connections` sizing + transaction-pooling pgbouncer) presented inside a "stopgap only; migrate to djust.tenants for long-term support" boundary — explicit in the subsection header

**In `python/djust/checks.py` (C014 messaging):**
- Primary warning message now explicitly flags django-tenants as deprecated (visible in `manage.py check` output without expanding hints)
- `hint` leads with migration to `djust.tenants` + link to the strategy guide; describes `TENANT_LIMIT_SET_CALLS = True` as a stopgap, NOT a long-term fix
- `fix_hint` reorders to lead with migration; labels the django-tenants config path as a stopgap

**In `python/djust/tests/test_c014_multi_tenant_asgi.py`:**
- 5 new/updated hint-quality test cases (16 total, up from 11 in #1558):
  - `test_hint_points_at_djust_tenants_alternative` — hint mentions `djust.tenants`
  - `test_hint_links_strategy_guide` — hint links the strategy guide
  - `test_hint_marks_django_tenants_deprecated` — hint explicitly says "deprecated"
  - `test_fix_hint_leads_with_migration` — fix_hint treats the flag as a stopgap
[...]

PR #1564, opened 2026-05-22
test(vdom): framework-level invariant pin for #1552 include-swap (investigation outcome — not a fix)

### LINKED ISSUE #1552: VDOM diff appends new subtree instead of replacing when {% if %}/{% elif %} swaps between {% include %}d templates (regression from 0.9.6rc2)

johnrtipton:
## Investigation outcome, not a fix

#1552 was [reopened on 2026-05-21](https://github.com/djust-org/djust/issues/1552) after PR #1555 shipped in rc7 and the reporter verified the user-visible symptom (post-swap DOM contains BOTH step subtrees) still reproduces. PR #1555 fixed a real-but-orthogonal issue (dj-id counter collisions across thread / msgpack boundaries); its E2E test used inline template bodies, not the reporter's bit-exact `{% include %}` shape.

This PR is the second-attempt investigation outcome. It is **NOT a fix.** It ships framework-level invariant pins covering THREE template shapes the reporter's production code uses, all of which **PASS on main**.

## What I tried

1. **`{% if/elif %}` + `{% include %}`** — the simplest include-swap shape. PASSES.
2. Same shape with **cross-thread msgpack round-trip** — the shape PR #1555 fixed, re-pinned for the include path. PASSES.
3. **`{% extends %} + {% block %} + {% if/elif %} + {% include %}`** — the reporter's full template-inheritance shape. PASSES.

All three produce 5 correct `RemoveChild` patches for the old subtree + correct `Insert`/`Replace` patches for the new subtree.

## Honest diagnostic conclusion

Per CLAUDE.md Bug-report triage rule #1 (*"trust the symptom, not the cited path"*) and the multi-reopen rule #1389 / PR #1086 precedent (*"Theory-testing against synthetic test cases is INSUFFICIENT — it confirms only that THE FRAMEWORK behaves a certain way, not that THIS USER'S BUG matches the theory"*): the framework's differ produces correct patches for every template shape the reporter described. The user-visible bug **lives in another layer**.

Candidates listed in the test file's module docstring:

- **WS-event save block** (PR #1466 / commit `a5e2c50c`) — `last_vdom` persists on every WS event; if save + load produce divergent VDOM trees, the diff is computed against stale state.
- **Sticky-child persistence** (ADR-018 / iters 18a/18b/18c).
- **JS-side patch application** under reporter-specific DOM conditions.
- An unsampled interaction.

## Why ship the tests anyway

Framework-level regression backstop. If a future change in the differ regresses any of these template combinations, the tests catch it fast. This pins the framework SIDE at "behaves correctly" so the next investigation iteration knows to look at the other layers, not re-trace the differ. Same shape as PR #1553 (which "empirically disconfirmed" the diff-layer hypothesis); this PR extends that work to the include-using shape.

## What happens next

**The #1552 issue stays OPEN.** Following this PR, I'll post a comment requesting the reporter share a **BugCapture URL** (the iter A feature shipped in PR #1563 the same day) capturing `state_before` + `state_after` + `vdom_patches` from the moment of the broken transition. Comparing their actual patches against the patches my framework reproducers produce will identify the divergent layer. The `docs/website/guides/bug-capture.md` page shipped with PR #1563 has the workflow.

This PR is also the empirical case-study justifying why bug-capture iter A is itself load-bearing for v1.1.0 — without it, the alternative is asking the reporter for their private template tree, which doesn't scale.

## Test plan

- [x] 3 framework-pin tests in `python/tests/test_if_elif_include_swap_framework_pin_1552.py`, all PASS on main
- [x] Module docstring documents the three template shapes covered, the negative result, the candidates for where the user-visible bug lives, and the "this is not a victory lap" warning to future readers
- [x] Two-commit shape (#181): test in `4580c11b`, CHANGELOG in `9f4cb663`. Gate 1 + Gate 2 verified.
- [x] Branch-verify reflex confirmed before each commit
- [x] Stale-base check: branch 0 commits behind `origin/main`

## Refs

- Refs #1552 (parent — issue stays OPEN; investigation outcome, not fix)
- Refs #1553 (prior test-pinning that disconfirmed the diff-layer hypothesis at the inline-template level — this PR extends that work to the include-using level)
- Refs #1555 (prior fix that addressed an adjacent issue; this PR documents what it covered and what it missed)
- Refs #1563 (bug-capture iter A — the requested escalation channel for reporter-side data)
- Refs #1561, #1562 (iter B / iter C of bug-capture; iter B's debug-panel share button is the practical workflow improvement for cases like this one)


johnrtipton:
# Stage 11 Review — PR #1564

**Verdict: APPROVE (with one 🟡 nit on the cross-posted comment, not blocking the merge)**

This is an unusual PR shape — a negative-result investigation pin with no code change — and it executes the shape cleanly. The framing is honest, the tests are bit-exact-faithful to the reporter's template (modulo include paths), the docstring + CHANGELOG both clearly warn future readers against the "tests pass on main = bug is fixed" trap that PR #1555 fell into, and the next-step is concrete and reporter-actionable via the bug-capture API shipped in PR #1563.

---

## What I verified

1. **Two-commit shape (#181 gate)** — `git show 4580c11b --name-only` returns only `python/tests/test_if_elif_include_swap_framework_pin_1552.py`; `git show 9f4cb663 --name-only` returns only `CHANGELOG.md`. Clean split, no cross-edit.
2. **Stale-base check (Action #1450)** — `git rev-list --count HEAD..origin/main` returns `0`; branch is 0 behind, 2 ahead. Safe to merge against current main.
3. **Cited PR/issue accuracy (Action #1071 / #1197)** — Spot-checked 5: `gh pr view 1466` = "feat(websocket): persist LiveView state on WS event for reconnect continuity (#1465)" ✓ matches the docstring's "PR #1466 / commit `a5e2c50c`" WS save-block citation. `gh pr view 1086` = "fix(templates): doubled filter-arg quotes (closes #1081)" ✓ matches the multi-reopen precedent reference. `gh pr view 1555` = "fix(vdom): advance dj-id counter past last_vdom's max" ✓ matches the "PR #1555 fixed an adjacent issue" claim. `gh pr view 1563` = "feat(bug-capture): BugCapture data shape" ✓ matches the iter A escalation path. `gh issue view 1389` = closed multi-reopen process issue ✓ matches the rule citation.
4. **Bug-capture API surface in the cross-posted comment** — grepped `python/djust/bug_capture.py:227` for `encode_view_state`, `:308` for `scrub_fields`, `:253` for the `render_with_diff()` example. The comment's code snippet matches the published module's API verbatim (3-tuple return, `patches=` kwarg, `scrub=scrub_fields(...)`). The security-model link target is real.
5. **Rust API surface used by the tests** — `crates/djust_live/src/lib.rs:196` `set_template_dirs`, `:206` `update_state`, `:315` `render_with_diff`, `:772` `serialize_msgpack`, `:807` `deserialize_msgpack`. All present; the cross-thread msgpack round-trip in `_run_cross_thread_include_render` exercises the same code path as `state_backends/memory.py:118`.

---

## Findings

### 🟢 Test correctness (priority 1)

`test_framework_include_swap_emits_correct_remove_then_insert_1552` at `python/tests/test_if_elif_include_swap_framework_pin_1552.py:181-217` does both halves of the symptom check:

- `assert "What Happened" in h2` (line 203) — NEW heading present
- `assert "About You" not in h2` (line 207) — OLD heading absent

This is the bit-exact form of the reporter's symptom (post-swap DOM contains BOTH subtrees). The cross-thread + msgpack round-trip in `_run_cross_thread_include_render` (lines 110-148) is production-faithful: it uses two `threading.Thread`s with queue handoff and goes through `serialize_msgpack` + `deserialize_msgpack` — the same shape that broke for #1555's dj-id-counter regression. Production hits this path via `InMemoryStateBackend.get()` doing msgpack round-trip on every read.

### 🟢 Bit-exact faithfulness to reporter's template (priority 2)

Reporter's template (from issue body): `{% if current_step_name == "claimant" %}{% include "intake/steps/mv_step_claimant.html" %}{% elif ... %}`.

Test template at `python/tests/test_if_elif_include_swap_framework_pin_1552.py:97-103` and `:289-298`: identical shape — parent `<div class="wizard-content">` wraps `{% if step == "claimant" %}{% include "step_claimant.html" %}{% elif step == "vehicle" %}{% include "step_vehicle.html" %}`. Only difference is the variable name (`step` vs `current_step_name`) and the file paths — both immaterial for the diff-layer surface under test. The third test (`test_framework_include_swap_with_extends_and_block_1552`, lines 270-338) adds the `{% extends "layout.html" %}` + `{% block wizard_step_content %}` shape, which matches the reporter's `vehicular_pd_wizard_base.html` exactly. Good coverage of the inheritance dimension.

### 🟢 "Investigation outcome, not a fix" framing (priority 3)

The module docstring (lines 1-72) is the most important artifact in this PR; it goes out of its way to be honest:

- Line 11: *"PR #1555 fixed an adjacent issue, not this one."*
- Lines 38-44: *"**All three tests PASS on main.** This is the empirically-honest finding... the user-visible bug must live in another layer"* with explicit candidates listed.
- Lines 67-72: *"this test file is NOT a successful-fix victory lap. PR #1555 also added tests and called them a fix... Future readers: take 'test passes on main' as evidence about THE FRAMEWORK behaving correctly, NOT as evidence about the user's bug being resolved."*

This is exactly the discipline CLAUDE.md's multi-reopen rule (Action #1389 / PR #1086) requires. The CHANGELOG entry mirrors this framing ("The user-visible bug must live in another layer... pursuing it requires reporter-side data, not more synthetic-shape framework theorizing").

### 🟢 Multi-reopen rule alignment (priority 6)

CLAUDE.md cautions against synthetic-shape theory-testing on multi-reopen issues. The argument for compliance:

- Tests are framed as framework-level pins, not user-bug reproducers (docstring lines 47-52).
- The next step is reporter-side data collection via bug-capture, not more synthetic templates.
- The issue stays OPEN (verified via `gh issue view 1552`); the cross-posted comment explicitly says *"the issue stays OPEN but I have no actionable next step beyond 'construct more synthetic templates from your description,' which the multi-reopen rule says is the wrong shape."*

The argument against — that future readers might misread these as a fix — is preempted three times: docstring, test docstring, CHANGELOG entry. This is the right shape.

### 🟢 Failure-mode coverage vs PR #1555 (priority 9)

PR #1555's tests used **inline templates**. This PR's tests use `{% include %}` (`python/tests/test_if_elif_include_swap_framework_pin_1552.py:97-103, 119, 122, 165-176, 270-298`). The Rust include resolution path (`set_template_dirs` + filesystem load) is exercised separately from the inline-template path. If a future regression breaks the include resolution path but leaves the inline path correct, PR #1555's tests would silently pass; this PR's tests would catch it. **Real net coverage gain** — answers the "what does this catch that #1555 didn't?" question honestly.

### 🟡 Cross-posted #1552 comment has a literal placeholder `issuecomment-XXXXXX`

The comment posted to issue #1552 (id `IC_kwDOQ7JDlM8AAAABDRjhaw`) contains, at the start of paragraph 2:

> This is the same conclusion the [first investigation comment](https://github.com/djust-org/djust/issues/1552#issuecomment-XXXXXX) reached for the inline shape

The `XXXXXX` is a leftover template placeholder; the link 404s. This is the same failure mode as Action #1197 (canon-doc citation discipline) — pre-commit grep for unresolved placeholders would catch it. The link is non-load-bearing (the prose around it carries the meaning), so this is 🟡 not 🔴, but the comment is editable via `gh issue comment --edit-last` and should be patched before the reporter reads it. Either:

1. Find the real first investigation comment URL and substitute it in, or
2. Replace the link with prose like *"as the first investigation comment reached for the inline shape"* without the broken hyperlink.

### 🟢 Stage 11 stale-base + two-commit-shape gates (priorities 7 + 8)

Both pass. Commit `4580c11b` (tests) touches no CHANGELOG; commit `9f4cb663` (CHANGELOG) touches no test file. Branch is 0 behind `origin/main`.

### 🟢 Additional tests I'd want to see — none blocking

The 3 shapes covered (simple include-swap, cross-thread+msgpack include-swap, extends+block+include-swap) are the right minimum set for the reporter's described surface. Plausible follow-up shapes (nested includes, components inside includes, empty-branch shape with `{% elif %}` to no content) would be useful as separate framework regression-pin work — but adding them to THIS PR would dilute its "this is the negative-result investigation outcome" message. Better as a follow-up issue if/when the bug-capture-driven trace identifies the actual layer.

---

## Summary

Verdict: **APPROVE.** Merge whenever the cross-posted #1552 comment's `issuecomment-XXXXXX` placeholder is patched (or accept it as a 🟡 nit and patch separately — the PR diff itself is clean). This PR is a textbook execution of the multi-reopen investigation discipline: honest framing, real net coverage gain over PR #1555's inline-template tests, concrete reporter-facing next step, all gates pass.

johnrtipton:
## Stage 14 retro — PR #1564 (#1552 second-attempt investigation outcome)

### What went well

- **Refused to repeat PR #1555's mistake.** PR #1555's tests passed-on-main while the bug remained in production because they tested an inline-template shape, not the reporter's bit-exact include-using shape. This PR tested THE REPORTER'S EXACT SHAPE (including the `{% extends %} + {% block %}` inheritance dimension) — and still couldn't reproduce. That negative result is the actionable finding, not a failure of the investigation. Per CLAUDE.md multi-reopen rule (#1389 / PR #1086): "Theory-testing against synthetic test cases is INSUFFICIENT — it confirms only that THE FRAMEWORK behaves a certain way, not that THIS USER'S BUG matches the theory." We followed that rule honestly.
- **Same-day pairing with bug-capture iter A is load-bearing.** This PR's investigation outcome ("framework reproducers can't catch the user-visible bug; need reporter-side data") is exactly the friction case-study that justifies why B7 iter A (PR #1563) shipped as v1.1.0 Path D rather than waiting. The comment posted to #1552 leverages the iter A feature concretely — the reporter can produce a BugCapture URL as soon as they upgrade, which collapses the next investigation cycle from "send the full template tree + reproduction recipe" to "send one base64 string." Without iter A, the only escalation channels would be requesting their private repo access or asking them to construct a minimal repro (~hours-to-days). With iter A, the escalation is a single tool call.
- **Reviewer caught the `XXXXXX` placeholder in the cross-posted issue comment.** Same trip-up class as PR #1560 (Action #1071/#1197 — canon-doc citation discipline). Caught + patched via `gh api PATCH` before merge so the reporter reads the correct version. The repeated trip-up (this is the second `XXXXXX` placeholder in three days) suggests the placeholder-tracking habit isn't yet automated; worth filing a check.
- **Honest framing in the test file's module docstring.** Three explicit "tests pass on main = framework is correct, NOT that the bug is fixed" warnings (positioned at the top, in the conclusion section, AND at the bottom). Preempts the PR #1555 trap for future readers. CLAUDE.md's bug-report-triage canon now has a concrete case-study referenced from a real test file's prose.

### What could be improved

- **The placeholder-comment trip-up is now a 2-in-3-day pattern.** PR #1560 had `#issuecomment-XXXXXX` in `docs/website/guides/multi-tenant.md`; this PR's #1552 comment had `#issuecomment-XXXXXX` again. The root cause is the same: a draft references "the prior comment" while the prior comment's URL isn't known yet, and the placeholder gets immortalized before the URL becomes available. Worth a pre-commit/pre-post check that greps `XXXXXX` (or `xxxxxx`) against the staged artifact and refuses to ship if either appears. Filed as a candidate Action Tracker rule.
- **The investigation could have surfaced the conclusion faster.** I constructed three framework reproducer shapes (inline, include-only, include+extends+block) before concluding "framework reproducers can't catch this." Per the bug-report triage rule's symptom-up principle, I could have started with: "what would the FRAMEWORK have to be doing wrong to leave both subtrees in the DOM?" — and recognized that ALL three framework outputs include `RemoveChild` patches for every old child, which definitionally means the framework is emitting the correct intent. The bug, if framework-side at all, must be in patch APPLICATION or in WHICH patches the WS save block transmits, not in patch GENERATION. Future shape: investigation should articulate "what does the framework have to do wrong" upfront and ask "do my reproducers exhibit that?" rather than "do my reproducers reproduce the user-visible DOM?"
- **The #1552 timeline now spans 3 attempts, ~6 days, and 2 wrong-root-cause PRs.** Cost: significant. The investigation discipline arrived after the second-attempt failure (PR #1555), not before the first. Earlier adherence to "bit-exact reporter shape required before claiming root cause" would have saved one fix-cycle. Already canonized in CLAUDE.md's multi-reopen rule; this PR is the case-study that should be cited the next time someone is tempted to ship a fix based on a synthetic shape.

### Process verification

- Branch-verify reflex ran before each commit
- Post-commit `git log -1 --oneline` reflex caught the FIRST commit attempt's bounce (pre-commit hook formatted the test file, stash-restore lost the staging, retry with re-add succeeded) — Action #122 working as designed
- Two-commit shape (#181) Gate 1 + Gate 2 both green
- Stage 11 Code Review verdict: **APPROVE with one 🟡** — placeholder URL patched inline (comment edit via `gh api`)
- Stage 12 Re-Review: **skipped** per skill rule (zero 🔴, zero 🟡 after the inline fix)
- Stage 13 Merge: `gh pr merge 1564 --squash --delete-branch --admin`

### What the next #1552 iteration needs

- The reporter shares a BugCapture URL via the cross-posted comment workflow
- Decoded patches get compared against the framework reproducer patches in this PR's test file
- If they diverge: investigation moves to the divergent layer (WS save block / sticky-child / JS application)
- If they match: investigation moves to the JS-side DOM application path
- Iter B (#1561) lands the debug-panel share button so the reporter (and future reporters) don't have to assemble the bug-capture call manually

### Related

- Refs #1552 (parent — issue stays OPEN; will close when reporter's BugCapture data identifies the actual layer)
- Refs #1553 (prior test-pinning at inline shape; this PR extends to include shape)
- Refs #1555 (prior fix that addressed dj-id collision; this PR documents what it covered and what it missed)
- Refs #1563 (bug-capture iter A — the escalation channel this PR's comment leverages)
- Refs #1561, #1562 (iter B/C — iter B's debug-panel button is the practical workflow improvement for future cases)

`RETRO_COMPLETE`

PR #1565, opened 2026-05-22
fix(websocket): gate WS-mount LOAD on enable_state_snapshot (fixes #1552)

### LINKED ISSUE #1552: VDOM diff appends new subtree instead of replacing when {% if %}/{% elif %} swaps between {% include %}d templates (regression from 0.9.6rc2)

johnrtipton:
## Diagnosis (bisect-confirmed)

| Version | Status |
|---|---|
| 0.9.6 | ✅ GOOD |
| **0.9.7rc1** | ✅ **GOOD** |
| **0.9.7rc2** | ❌ **BAD** (regression introduced) |
| 0.9.7 through 1.0.0rc8 | ❌ BAD |

Single feature commit between rc1 and rc2: **`a5e2c50c feat(websocket): persist LiveView state on WS event for reconnect continuity (#1465) (#1466)`**.

PR #1564's own docstring on the framework reproducer predicted exactly this suspect:

> *"WS-event save block (PR #1466, commit `a5e2c50c`): writes `last_vdom` to the session on every WS event. Did not exist pre-v0.9.6rc2's window in the same shape — may interact with the msgpack roundtrip such that a partial sync_ids state lands in the next event's `last_vdom`, producing patches the client can't resolve."*

## Root cause

PR #1466 changed `handle_mount`'s LOAD gate from `if has_prerendered:` to `if has_prerendered or saved_state:` **and** made the `await request.session.aget(view_key, {})` read itself unconditional. Result: every view got its previously-saved session state restored on every WS mount — including views that never opted in via `enable_state_snapshot`.

For non-opt-in views, that restoration runs AFTER `mount()` has initialized the view. The next `render_with_diff()` then diffs against a baseline that's been clobbered by the session-restored state. Patches reference dj-ids from that internal tree, which don't correspond to the client's actual DOM. The applier's `querySelector(':scope > [dj-id=X]')` fails, falls back to index, removes the wrong node after `InsertSubtree` has shifted positions — old subtree survives. **The wizard step-leak on #1552.**

PR #1475 (commit `066d7f05`) later added a SAVE-side gate on `enable_state_snapshot`. That fixed the snapshot-on-idle write amplification but left the LOAD path unconditional — so the bug introduced by PR #1466 survived 0.9.7rc3, 0.9.7 final, and all 1.0.0 rcs through rc8.

## Fix

Gate the saved_state read in `handle_mount` symmetrically with the existing SAVE-block gate:

```python
saved_state = (
    await request.session.aget(view_key, {})
    if request.session
    and getattr(self.view_instance, "enable_state_snapshot", False)
    else {}
)
```

Behavior preserved for opt-in views (`enable_state_snapshot = True`) — PR #1466's reconnect-resume capability still fires for them. Behavior restored to 0.9.7rc1 for default views.

## Tests

- `test_load_gate_loosened_fires_on_saved_state_without_has_prerendered` updated: source-grep assertion now requires both `request.session` AND `enable_state_snapshot` to appear in the load gate. Switched to `rfind` so the test tolerates the gate's phrase appearing in the explanatory comment block.
- All 10 tests in `test_ws_reconnect_state_1465.py` (PR #1466's own test file) pass.
- Framework reproducers `test_if_elif_include_swap_framework_pin_1552.py` (PR #1564) and `test_if_elif_swap_e2e_1550_1552.py` continue to pass.
- Full python suite: **7370 passed, 21 skipped, 0 failed**.

## Verified end-to-end on the consumer

Installed the patched wheel into the NYC Claims consumer's venv (`djust==1.0.0rc8` with this fix), restarted Daphne, navigated to `/intake/vehicular-pd/?isDemo=scenario2`:

- Step 1 ("About You") → click Next → step 2 ("What Happened") only, **no leak**.
- Step 2 → click Next → step 3 ("The Damage") only, **no leak**.
- Step 3 → click Back → step 2 only, **no leak**.

Every transition shows exactly one `h2.card-title` in the DOM. The wizard step-leak that triggered #1552 is fixed.

## Followups on the issue

- [Diagnostic chain](https://github.com/djust-org/djust/issues/1552#issuecomment-4514921262) (initial — incorrectly hypothesized dj-id contract; superseded by this bisect)
- [Bisect log](https://github.com/djust-org/djust/issues/1552) — see latest comment

refs #1552, #1466

---

**Citation corrigendum (post-Stage-11 review).** The original commit message (`624e705a`) and an earlier draft of this PR body cited "PR #1475" for the SAVE-side gate. The actual merged PR is **#1478** (titled `fix(websocket): gate WS-event save on enable_state_snapshot (#1475 P0)`); #1475 is the issue it closed. The follow-up commit `1187ad97` (CHANGELOG fix-up) uses the correct citation throughout. Original commit body retained as-is rather than force-pushed to keep review history honest. Per Action #1197 (canon-doc citation discipline).


johnrtipton:
## Stage 11 Review — APPROVE (with one 🟡 should-fix and one 🟢 nit)

**Verdict: APPROVE**

This is the right fix at the right layer. The bisect-confirmed diagnosis is sound, the symmetry argument with the #1475/#1478 SAVE-side gate is correct, the fix is defensively coded, the test was tightened to be non-tautological, and end-to-end reporter-environment verification closes the loop on the multi-reopen rule (Action #1389). All 14 CI checks green; 0 commits behind `origin/main`.

### What I verified

1. **Bisect logic — CORRECT.** `git log v0.9.7rc1..v0.9.7rc2 --oneline` shows exactly 5 commits in that window, of which the ONLY feature-class commit touching `websocket.py` is `a5e2c50c feat(websocket): persist LiveView state on WS event for reconnect continuity (#1465) (#1466)`. The other four are `chore(cargo)` lockfile refresh, version bump, retro docs, and roadmap docs. The bisect's central premise — "single feature commit between rc1 and rc2" — holds.

2. **Symmetry with the SAVE-side gate — CONFIRMED.** `python/djust/websocket.py:3143-3144` carries the SAVE-side `enable_state_snapshot` gate added by commit `066d7f05` (PR #1478, fixing issue #1475). The new LOAD gate added by this PR at `websocket.py:2010-2014` mirrors it exactly — both require `enable_state_snapshot = True` to engage the session round-trip. The asymmetry-since-rc3 story is real and the fix closes it.

3. **Fix is tight — CONFIRMED.**
   - `websocket.py:2012` — `if request.session and getattr(self.view_instance, "enable_state_snapshot", False)` — guards the no-session case AND uses `getattr(..., False)` defensively, so views that never declare the attribute don't AttributeError.
   - `websocket.py:2013` — fallback is `else {}`, not `None`, so the downstream `if has_prerendered or saved_state:` boolean (`websocket.py:2015`) treats the gated-off case as "no saved state" cleanly.
   - Placement is correct: the gated read sits BEFORE the `if has_prerendered or saved_state:` consumer at line 2015. The nested reads at `websocket.py:2023` (`private_state`) and `websocket.py:2050` (`component_state`) are inside `if saved_state:` (line 2016) — transitively gated by the same opt-in, so no leak there.

4. **PR #1466's tests preserved — CONFIRMED.** Test count is 10 → 10 (no removals). The only modified test is `test_load_gate_loosened_fires_on_saved_state_without_has_prerendered` (`test_ws_reconnect_state_1465.py:92-135`); the assertion at line 119 was strengthened from `"if request.session else {}" in source` to `"request.session" in source and "enable_state_snapshot" in source`. The other 9 tests already used the `enable_state_snapshot = True` opt-in shape because they exercise the reconnect-resume path #1466 added.

5. **Tautology self-test (Action #1200 / #1468) — PASSES.** If the fix is reverted (remove `and getattr(...enable_state_snapshot)` from the gate), the new assertion at `test_ws_reconnect_state_1465.py:119` fails — the source would no longer contain `"enable_state_snapshot"` inside `handle_mount`. The test is exercising the change, not a coincident invariant.

6. **End-to-end consumer verification — SATISFIED (Action #1389 rule).** PR body includes specific URL (`/intake/vehicular-pd/?isDemo=scenario2`), specific wizard transitions (step 1→2, 2→3, 3→back-to-2), and the specific DOM-state assertion ("exactly one `h2.card-title` in the DOM"). This is bit-exact reporter-environment-faithful verification on the user's actual wheel install — the first time in the #1552 saga this has been done. PRs #1553, #1555, #1564 were all framework-synthetic.

7. **Compatibility surface — CLEAN.** Grepping `websocket.py` for `request.session.aget` / `aget(view_key` / `liveview_<` reveals only the changed call site at line 2011 and two transitively-gated reads at 2023/2050. No other consumer of `view_key` from session exists, so the LOAD-gate change is fully localized to `handle_mount`.

8. **Stale-base — CLEAN.** `git rev-list --count HEAD..origin/main` returns 0.

### 🟡 Should-fix

- **CHANGELOG entry missing.** `CHANGELOG.md` `[Unreleased]` section is empty. Per the project policy (`docs/PULL_REQUEST_CHECKLIST.md`: every `feat:` and `fix:` PR must update CHANGELOG), this PR needs an entry — either under `[Unreleased]` now or rolled into a fresh `[1.0.0rc9]` block when rc9 is cut. Suggested wording: a `### Fixed` entry under #1552, framing this as "LOAD-side counterpart to #1478, closes the regression introduced by #1466." The two pre-existing CHANGELOG entries for #1552 (under rc8 `### Tests` and rc7 `### Fixed`) cover the framework-pin test and the dj-id counter fix respectively — neither describes the actual root-cause fix this PR ships, so the entry is genuinely missing, not a duplicate.

### 🟢 Nit

- **Citation-discipline (Action #1197 / #1071): #1475 is an ISSUE, #1478 is the PR.** The PR body says *"PR #1475 (commit `066d7f05`) later added a SAVE-side gate"* and the commit message uses *"PR #1475 later added a SAVE-side gate"* — but `gh issue view 1475` shows #1475 is a closed issue (*"0.9.7rc2: WS state-snapshot close path captures uvicorn in unrecoverable asyncio state when host snapshots within ~200ms"*), and `gh pr view 1478` is the actual PR that landed `066d7f05` (*"fix(websocket): gate WS-event save on enable_state_snapshot (#1475 P0)"*). Commit hash `066d7f05` is correct. Suggested phrasing: *"the SAVE-side gate added by PR #1478 for issue #1475"*. Not blocking; surfaces in case a future grep against the canon doc rules cites this.

### Cited diff lines

- `python/djust/websocket.py:2010-2014` — the new gated LOAD expression.
- `python/djust/websocket.py:2015-2016` — the `if has_prerendered or saved_state:` consumer that the gate feeds, confirming the placement.
- `python/djust/websocket.py:3143-3144` — the pre-existing SAVE-side gate this PR mirrors.
- `python/djust/tests/test_ws_reconnect_state_1465.py:119-122` — the strengthened (non-tautological) assertion.
- `python/djust/tests/test_ws_reconnect_state_1465.py:130-134` — the `rfind` switch that tolerates the new comment-block mention of the gate phrase.

Once the CHANGELOG entry lands, this is ready to merge. The fix is well-scoped, well-tested, and end-to-end verified.

johnrtipton:
## Stage 14 retro — PR #1565 (the real #1552 fix)

### What went well

- **Bisect closed the multi-reopen loop.** The #1552 saga ran four PRs: #1553 (test-pinning, "candidate disconfirmed"), #1555 (dj-id counter — adjacent bug fixed, not user-visible one), #1564 (framework-pin investigation, "bug is elsewhere"), and now #1565 (real fix). Each predecessor's negative result narrowed the search space; #1564's docstring explicitly named PR #1466 as the prime suspect, and the bisect on the consumer environment confirmed it. The multi-reopen rule (Action #1389 / PR #1086 precedent) was honored at the last mile — bit-exact end-to-end verification on the reporter's NYC Claims wizard (step 1 → 2 → 3 → back, every transition shows exactly one `h2.card-title`) was the load-bearing evidence that the fix actually fixes the user-visible symptom. Earlier PRs that lacked this gate produced false fixes (#1555).
- **The PR #1564 docstring paid off.** That investigation PR was framed as "framework reproducer can't catch this; bug must be in another layer" and listed candidates including "WS-event save block (PR #1466)." The PR-#1565 author followed that candidate list directly to the bisect target. The pattern of leaving honest "where the bug isn't" findings in the test file's prose (not just in retro markdown that nobody re-reads) was load-bearing for the next investigation iteration.
- **PR-#1478 SAVE-side gate was a clean symmetry anchor.** When PR #1478 landed (closing issue #1475) it gated SAVE but left LOAD unconditional — an asymmetry that survived 5 RCs unnoticed. The Stage 11 reviewer correctly verified the SAVE gate at `websocket.py:3143-3144` exists before approving the LOAD-side counterpart at `websocket.py:2010-2014`. Without the SAVE gate as evidence, the asymmetry-since-rc3 narrative would have lacked grounding.
- **First-time PR-class verification: end-to-end on consumer.** The PR body's "Installed the patched wheel into the NYC Claims consumer's venv (djust==1.0.0rc8 with this fix), restarted Daphne, navigated to `/intake/vehicular-pd/?isDemo=scenario2`" verification is the gold standard for multi-reopen fixes. Specific URL, specific transition steps, specific DOM assertion (one `h2.card-title` per render). Worth canonizing as a pattern for WS-critical fixes generally.

### What could be improved

- **The asymmetric gate took 6 releases to catch (0.9.7rc2 → 1.0.0rc8).** PR #1478 fixed half the bug shape on 2026-05-04 but the other half stayed in production through 5 subsequent RC cuts. The grep that would have caught it ("for every `await request.session.aget(view_key, ...)`, is there a symmetric `enable_state_snapshot` gate?") is small. Worth filing as a candidate Stage 11 reviewer-check for WS-state PRs: any PR that gates a state-persistence write must also gate the matching read, or explicitly document why asymmetry is correct.
- **Three preceding PRs (#1553, #1555, #1564) shipped without catching the bug, costing roughly 6 days of investigation.** Cost is acceptable in retrospect (#1555 fixed a real adjacent bug; #1564 narrowed the candidate list); but earlier discipline against the multi-reopen rule (specifically: refusing to ship a "fix" without bit-exact end-to-end verification on the reporter's environment) would have made #1555 a Stage-4 STOP rather than a merge. The lesson is already canonized in CLAUDE.md and was honored on #1564 (which didn't claim a fix) — the gap is at the implementer-Stage-5 / reviewer-Stage-11 enforcement layer for #1555. Filed as a candidate action-tracker rule: "for any PR whose task description references a multi-reopen issue, the implementer must attach a consumer-environment verification artifact (screenshot, DOM snapshot, or signed-off reproduction recipe) before requesting review; the reviewer must verify the artifact exists and is specific."
- **Citation discipline trip-up again (#1475 vs #1478).** Third time this session (PR #1560 docs `XXXXXX`, PR #1564 comment `XXXXXX`, this PR's commit body `#1475` for what should have been `#1478`). The pattern: someone references "the prior PR" by what they recall the number being, without re-grepping. Worth filing the pre-commit grep proposed in PR #1564's retro: any commit body or doc citing `#NNNN` should fail-loud if the cited number is a CLOSED issue or if the cited number's title doesn't match the context ("PR #X" should be a `state: MERGED` or `OPEN` PR, not an `ISSUE`).
- **Force-push avoided correctly.** The Stage 11 🟢 was a citation discipline issue in the commit body. The instinct to amend-and-force-push would have been wrong (the PR already had a posted review; rewriting history under that review is uglier than appending a corrigendum). The corrigendum-via-`gh pr edit` pattern adopted here is worth canonizing for post-review citation fixes: amend a published commit body only when the fix is large enough to warrant rewriting history; otherwise append a corrigendum to the PR body that future readers see alongside the original.

### Process verification

- Branch-verify reflex ran before the CHANGELOG commit
- Post-commit `git log -1 --oneline` reflex ran for both the original commit (verified externally by PR author) and the CHANGELOG fix-up
- Two-commit shape (#181): impl commit `624e705a` did NOT touch CHANGELOG.md (Gate 1 OK retroactively); docs commit `1187ad97` touched only CHANGELOG.md (Gate 2 OK)
- Stage 11 Code Review verdict: **APPROVE** (1 🟡 missing CHANGELOG, 1 🟢 citation discipline) — both addressed before merge
- Stage 12 (Review Verdict): findings addressed, Re-Review not required
- Stage 13 Merge: `gh pr merge 1565 --repo djust-org/djust --squash --delete-branch --admin`

### What's next

- **Close #1552** as fixed by this PR (will happen automatically via "refs #1552" but verify in a moment).
- **Cut v1.0.0rc9** with this fix — the fix targets the LOAD path that has been broken since 0.9.7rc2; users on any 0.9.7 / 1.0.0rc * release have the bug in production. rc9 should ship within the next few hours rather than waiting for v1.0.0 GA.
- **Notify the #1552 reporter.** They have been tracking this saga; a comment with "rc9 has the fix; upgrade and verify the wizard step-leak is gone" plus the bug-capture iter A workflow as a fallback for any reappearance.
- **File the candidate action-tracker rules.** Two new rules from this retro: (a) reviewer-check for read/write gate symmetry on state-persistence PRs; (b) pre-commit/pre-doc grep for PR-number citations that resolve to issues instead of merged PRs.

### Related

- Refs #1552 (closes — this is THE fix)
- Refs #1466 (the introducing commit per bisect — `a5e2c50c`)
- Refs #1478 (the partial SAVE-side fix; this PR's LOAD-side counterpart)
- Refs #1553, #1555, #1564 (the predecessor investigation chain)
- Refs #1563, #1561, #1562 (bug-capture iter A/B/C — the escalation channel for future cases)

`RETRO_COMPLETE`

PR #1584, opened 2026-05-23
feat(runtime): ViewRuntime accepts renderer_factory kwarg — LVN-I PR-2 (#1577)

johnrtipton:
Per [ADR-019](docs/adr/019-liveview-native.md) §'Three layers' §1, LVN-I PR-2. Plumbs a renderer factory through ViewRuntime so PR-3 (handshake) can set it based on the connection's ?platform= query param.

- EDIT: `python/djust/runtime.py` — new kwarg + 1 instance attr (8 lines)
- NEW: `python/djust/tests/test_runtime_renderer_param.py` — 3 tests
- CHANGELOG entry under [Unreleased]

Default `None` for back-compat — existing TemplateMixin.render_with_diff:942 still constructs HtmlRenderer inline (from PR-1). PR-3 will wire dispatch.

Tests pass: 24/24 in adjacent test files. Stacks on #1583 (already merged).

🤖 Generated with [Claude Code](https://claude.com/claude-code) via /pipeline-run

PR #1583, opened 2026-05-23
feat(renderers): Renderer Protocol + HtmlRenderer — LVN-I PR-1 (#1577)

johnrtipton:
## Summary

Foundation iteration of [ADR-019 LiveView Native](docs/adr/019-liveview-native.md). Tracking issue: #1577 (LVN-I). Targets `1.1` per the [project-local pipeline templates](.pipeline-templates/README.md).

## What's in this PR

- **NEW**: `python/djust/renderers/` package
  - `base.py` — `Renderer` Protocol (`@runtime_checkable`; `output_format: str`; `render_with_diff(...) -> tuple[str, Optional[str], int]`)
  - `html.py` — `HtmlRenderer` (binds to a LiveView, delegates `render_with_diff()` to `view._rust_view.render_with_diff()`)
  - `__init__.py` — re-exports
- **EDIT**: `python/djust/mixins/template.py:942` — one-line surgical replacement. The inline `self._rust_view.render_with_diff()` becomes `HtmlRenderer(self).render_with_diff(...)`. Surrounding state-sync / timing / extraction logic unchanged.
- **NEW**: `python/djust/tests/test_renderer_protocol.py` (11 tests) — covers package imports, Protocol shape, `HtmlRenderer` conformance, delegation to `_rust_view`, and the dispatch gate.

## What this PR explicitly does NOT do

- `ViewRuntime` plumbing → PR-2 of LVN-I (separate PR)
- `?platform=` handshake parsing → PR-3 of LVN-I
- `NativeRenderer` + widget vocabulary → LVN-II (#1578)
- `crates/djust_vdom` untouched — wire format unchanged
- `static/djust/client.js` untouched — browser client byte-identical

## Two-commit shape

Per [#1173 canon](docs/adr/). Commit 1: implementation only (no CHANGELOG). Commit 2: CHANGELOG only.

## Verification

- `pytest python/djust/tests/test_renderer_protocol.py` → 11/11 pass
- `pytest python/djust/tests/` (full djust suite) → 2991 pass, 3 skipped, 0 fail (same counts as pre-PR — no regression)
- Gate test `test_mixin_dispatches_through_html_renderer` was RED before the `mixins/template.py:942` edit (`ModuleNotFoundError: djust.renderers`) and GREEN after — proving the dispatch seam is live (satisfies #1210 verify-artifact-before-planning gate).

## Plan + ADR references

[...]

Issue #1578, opened 2026-05-23
LVN-II: Native widget vocabulary + NativeRenderer (ADR-019)


johnrtipton:
## Context

Iteration **II of V** of [ADR-019: LiveView Native](docs/adr/019-liveview-native.md). Builds on the renderer abstraction from LVN-I. Introduces the 12-widget vocabulary (SwiftUI ∩ Compose) and a `NativeRenderer` that emits widget-shaped VNodes directly, skipping the HTML round-trip.

LVN-III + LVN-IV (the Swift / Kotlin clients) block on this iteration.

## Widget vocabulary (frozen for v1; per ADR §3)

| Tag | SwiftUI | Compose |
|---|---|---|
| `<Stack>` | `VStack` | `Column` |
| `<HStack>` | `HStack` | `Row` |
| `<ZStack>` | `ZStack` | `Box` |
| `<Text>` | `Text` | `Text` |
| `<Button>` | `Button` | `Button` |
| `<TextField>` | `TextField` | `TextField` |
| `<Toggle>` | `Toggle` | `Switch` |
| `<List>` | `List` | `LazyColumn` |
| `<Image>` | `Image` | `Image` |
| `<ScrollView>` | `ScrollView` | `Modifier.verticalScroll` |
| `<Spacer>` | `Spacer` | `Spacer` |
| `<NavigationView>` | `NavigationView` | `NavHost` |

Event attrs: `dj-tap`, `dj-change`, `dj-input`. Styling subset: `padding`, `spacing`, `alignment`, `foregroundColor`, `font`.

## Touchpoints

- **NEW**: `python/djust/renderers/native.py` — `NativeRenderer` produces VNodes with widget tag names
- **NEW**: `python/djust/template/native_loader.py` (or extend existing loaders) — resolves `foo.swiftui.html` / `foo.compose.html` by extension match based on `output_format`
- **EDIT**: Django template tag handlers (`python/djust/templatetags/`) — most (`{% if %}`, `{% for %}`, `{{ var }}`) work unchanged; structural tags (`{% block %}`, `{% include %}`) need native-target awareness

## PR-sized breakdown

- [ ] **PR 1** — Widget vocabulary spec doc (`docs/native-widget-vocabulary.md`) + frozen tag list as a Python constant
- [ ] **PR 2** — `NativeRenderer` implementation: walks Django template AST → emits widget-VNode tree (no HTML string round-trip)
[...]

Issue #1577, opened 2026-05-23
LVN-I: Renderer abstraction in djust core (ADR-019)


johnrtipton:
## Context

Iteration **I of V** of [ADR-019: LiveView Native](docs/adr/019-liveview-native.md). The foundation: introduce a pluggable `Renderer` abstraction so djust's existing server-side reactive lifecycle can drive non-HTML targets (SwiftUI, Compose, anything else). HTML stays default — this iteration alone produces NO external behavior change for browser views; it's pure refactor.

LVN-II / III / IV / V all block on this.

## Touchpoints (per ADR architecture section)

- **NEW**: `python/djust/renderers/__init__.py`, `renderers/html.py`, `renderers/native.py` (latter is a stub for II)
- **EDIT**: `python/djust/mixins/template.py:~898` — `render_with_diff` accepts a renderer from the runtime, falls back to `HtmlRenderer` for back-compat
- **EDIT**: `python/djust/runtime.py` (ADR-016) — `ViewRuntime.__init__` takes a renderer parameter
- **EDIT**: `python/djust/websocket.py:~1074–1115` — `LiveViewConsumer` handshake reads `?platform=` query param OR `Sec-WebSocket-Protocol: djust.swiftui` subprotocol; passes choice into `ViewRuntime` construction
- **UNCHANGED**: `crates/djust_vdom/src/lib.rs` — `Patch` enum and `VNode` shape are renderer-agnostic (verified in ADR Verification §1)
- **UNCHANGED**: `static/djust/client.js` — browser client byte-identical

## PR-sized breakdown

- [ ] **PR 1** — `Renderer` Protocol + `HtmlRenderer` (wraps current Django-template + Rust-parse pipeline). Existing `render_with_diff` calls `HtmlRenderer.render()` instead of inline; no behavior change.
- [ ] **PR 2** — `ViewRuntime` accepts a renderer; threads it through to `render_with_diff`. Default param = `HtmlRenderer()` so existing call sites don't break.
- [ ] **PR 3** — `LiveViewConsumer` handshake reads `?platform=` / subprotocol; selects renderer; falls back to `html`. Pure additive.

## Acceptance criteria (from ADR Gate)

- [ ] djust's full test suite green on `1.1` branch
- [ ] [demo project](examples/demo_project/) regression-verified (browser client renders identically pre/post-refactor)
- [ ] [max-companion](https://github.com/johnrtipton/max-companion) `make verify` still 11/11
[...]

PR #1585, opened 2026-05-23
feat(websocket): handshake selects renderer via ?platform= — LVN-I PR-3 (#1577)

johnrtipton:
Completes LVN-I per [ADR-019](docs/adr/019-liveview-native.md) §'Three layers' §1-§2.

`RENDERERS` registry + `get_renderer_factory(platform)` helper in `djust/renderers/__init__.py`. `LiveViewConsumer._get_runtime` parses `?platform=` from scope query string, resolves via registry, passes to `ViewRuntime` as `renderer_factory`. Unknown/missing values → None → HtmlRenderer default → browser byte-identical.

LVN-II (#1578) will register `swiftui` + `compose` and the handshake lights up.

5 new tests; full djust suite 2999/3002 pass (no regression).

Refs: #1577. Stacks on #1583 + #1584 (already merged).

🤖 /pipeline-run


johnrtipton:
## Retro (LVN-I PR-3) — LVN-I track COMPLETE

Closes out the LVN-I iteration. The 3-PR sequence (Protocol → Runtime → Handshake) shipped on 1.1:
- #1583 PR-1 protocol + HtmlRenderer
- #1584 PR-2 ViewRuntime field
- (this PR) handshake-based selection

The renderer abstraction is now live end-to-end: a future native client connecting with ?platform=swiftui will hit ViewRuntime carrying a SwiftUI factory the moment LVN-II registers it. Until then, the path is HTML-only and byte-identical for the browser.

8 net new tests across the 3 PRs. Two of the three commits hit the pre-commit-hook swallow pattern (#122) and needed retry — worth filing if the hook can be made more robust.

Next: LVN-II (#1578) — widget vocabulary + NativeRenderer.

PR #1596, opened 2026-05-25
fix(theming): make presets.get_preset() consult runtime registry first (closes #1595)

### LINKED ISSUE #1595: theming: runtime-registered preset invisible to CSS generator (registry vs static THEME_PRESETS dict)
## Summary

`register_preset()` adds a preset to the runtime registry, and the `ThemeManager` validates against the registry — but the CSS generator that ultimately renders `--primary`, `--link`, etc. into `:root` reads from a **static module dict** (`djust.theming.presets.THEME_PRESETS`), which is blind to runtime registration. Result: a custom preset registered via the documented `register_preset()` API is silently invisible to the actual rendered output. The default `THEME_PRESETS["default"]` is rendered instead.

## Reproducer

```python
# medicare/themes.py (consumer app)
from djust.theming.themes._base import ColorScale, ThemePreset, ThemeTokens

MAX_TEAL = ColorScale.from_hex("#146a5d")  # H~171°
WHITE = ColorScale(0, 0, 100)
# ... (LIGHT ThemeTokens with primary=MAX_TEAL etc.)
PRESET = ThemePreset(name="max", display_name="...", light=LIGHT, dark=LIGHT)
```

```python
# medicare/apps.py
def ready(self):
    from djust.theming.registry import register_preset
    from . import themes
    register_preset(themes.PRESET.name, themes.PRESET)
```

```python
# settings.py
LIVEVIEW_CONFIG = {
    "theme": {"pack": "max"},  # pack with color_preset="max"
}
```

After this:

```python
>>> from djust.theming.manager import get_theme_manager
>>> state = get_theme_manager().get_state()
>>> state.preset
'max'  # ✅ manager sees it

>>> from djust.theming.registry import get_registry
>>> get_registry().get_preset('max').light.primary.to_hsl()
'171 68% 25%'  # ✅ registry has my preset

>>> from djust.theming.presets import get_preset
>>> get_preset('max').light.primary.to_hsl()
'240 6% 10%'  # ❌ module-dict path returns DEFAULT_THEME (slate-black)
```

Curl the rendered page:

```
$ curl -s http://localhost:8111/ | grep -oE -- "--primary:[^;]*"
--primary: 240 6% 10%   # ❌ generic slate, NOT my Medicare-teal
```

## Root cause

`djust/python/djust/theming/presets.py:158-160`:

```python
def get_preset(name: str) -> ThemePreset:
    """Get a theme preset by name, with fallback to default."""
    return THEME_PRESETS.get(name, DEFAULT_THEME)
```

`THEME_PRESETS` is a static module-level dict populated at import time from built-in theme files. The runtime registry (`djust/python/djust/theming/registry.py`) has its own `_presets` dict that `register_preset()` mutates, but `presets.get_preset()` doesn't consult it.

The CSS generator path that renders `{{ theme_head }}` ultimately goes through `presets.get_preset()` (via `pack_css_generator` and friends), so user-registered presets never make it into the rendered output.

## Workaround consumers can use today

```python
# In AppConfig.ready(), after the documented register_preset():
from djust.theming import presets as _presets
_presets.THEME_PRESETS[themes.PRESET.name] = themes.PRESET
```

Both writes are needed because the manager + theme switcher consult the registry, but the CSS generator consults the module dict.

## Suggested fix

Make `presets.get_preset()` consult the runtime registry first, mirroring the pattern in `theme_packs.get_theme_pack()` (`djust/python/djust/theming/theme_packs.py:1216-1222`):

```python
def get_preset(name: str) -> ThemePreset:
    from .registry import get_registry
    return get_registry().get_preset(name) or THEME_PRESETS.get(name, DEFAULT_THEME)
```

That's the same "registry first, static dict fallback" shape `theme_packs.get_theme_pack()` already uses; nothing else needs to change. Removes the need for the consumer workaround above.

## Version

djust `v1.0.0rc9` (commit `150c1edf`).

## Impact

Any consumer following the documented `register_preset()` API to register a custom theme will see their colors silently replaced by the default slate-black palette in the actual rendered CSS. The manager / theme switcher / `gh-pr-checks`-style introspection report the registered preset correctly, masking the bug.

## Severity

Low (workaround exists, single static-dict write) but the divergence between "what the API says is registered" and "what the renderer actually uses" is the kind of thing that loses an hour of debugging — exactly the cost we paid finding this. Worth fixing in the framework so the next person doesn't repeat it.

🤖 Filed by Claude during a `/pipeline-ship` cycle on max-companion (https://github.com/johnrtipton/max-companion). Happy to send the suggested fix as a PR if useful.

johnrtipton:
## Summary

Fixes #1595 — `register_preset()` adds to the runtime registry, but `presets.get_preset()` (the function the CSS generator path ultimately calls) reads from the static `THEME_PRESETS` module dict only. Result: user-registered presets are visible to manager/registry/introspection but **invisible to the renderer**, silently replaced with the default slate-black palette in `:root`.

## Fix

Mirror the registry-first-OR-static-fallback dispatch already in `theme_packs.get_theme_pack()` (line 1216-1222):

```python
def get_preset(name: str) -> ThemePreset:
    from .registry import get_registry
    return get_registry().get_preset(name) or THEME_PRESETS.get(name, DEFAULT_THEME)
```

Same shape as the theme-pack dispatch, with the same `import .registry` inside the function to avoid the circular import that bites if `registry` imports back from `presets`.

Removes the need for the documented workaround (`_presets.THEME_PRESETS[name] = preset`) — consumers can now use the public `register_preset()` API alone.

## TDD verification

- **Red phase**: wrote `test_get_preset_consults_runtime_registry_first_1595` first, ran against main → FAILED with `presets.get_preset(sentinel) returned 'default' instead of the runtime-registered preset` (the exact symptom the issue describes).
- **Green phase**: applied the fix → test passes.
- **Gate-the-fix-off self-test (Action #1200/#1468)**: stashed the fix → exactly 1 of 3 new tests fails (the regression case); other 2 pass (no tautology). Fix restored, all 3 green.

## Tests

3 new regression tests in `python/djust/tests/test_theming_presets.py`:

1. `test_get_preset_consults_runtime_registry_first_1595` — the load-bearing assertion (registered preset is visible via registry AND via `get_preset()`)
2. `test_get_preset_static_dict_fallback_when_registry_lacks_name` — locks in the second half of the dispatch contract
3. `test_get_preset_unknown_name_returns_default` — final fallback to `DEFAULT_THEME` unchanged

Full suite results:
- All 8 tests in `test_theming_presets.py` pass
- All 1811 theming tests pass (`pytest python/djust/tests/test_theming_*.py tests/unit/test_theming_*.py -q`)
- Full project suite: 4390 py + 1609 js + Rust — all green

## Two-commit shape

- `8a4d2d8e` fix + tests (no CHANGELOG) — Gate 1 verified
- `357439a2` CHANGELOG only — Gate 2 verified

## Refs

- Closes #1595
- Mirrors `theme_packs.get_theme_pack()` (`python/djust/theming/theme_packs.py:1216`)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- **Theming registry/static-dict divergence — runtime-registered presets now reach the CSS generator (#1595).** `register_preset()` adds to a runtime `Registry._presets` dict that the theme manager, theme switcher, and introspection APIs all consult; but `presets.get_preset()` — the function the CSS generator path ultimately calls to render `--primary` etc. into `:root` — read only from the static `THEME_PRESETS` module dict, blind to runtime registration. Result: any consumer following the documented `register_preset()` API in `AppConfig.ready()` got their custom palette silently replaced with the default slate-black `THEME_PRESETS["default"]` in the actual rendered CSS, while the manager/switcher/`gh-pr-checks`-style introspection correctly reported the registered preset as active — exactly the kind of API-says-X-but-renderer-uses-Y divergence that costs an hour of debugging. Fix mirrors the registry-first-OR-static-fallback dispatch already established in `theme_packs.get_theme_pack()` (`python/djust/theming/theme_packs.py:1216-1222`): `get_preset()` now consults `get_registry().get_preset(name)` first, then falls back to `THEME_PRESETS.get(name, DEFAULT_THEME)`. Same shape, same `import .registry` inside the function to avoid the circular import that bites if registry imports back from presets. Covered by 3 new regression tests in `python/djust/tests/test_theming_presets.py` (`test_get_preset_consults_runtime_registry_first_1595` + 2 companions locking in the second half of the contract: static-dict fallback for built-in names + `DEFAULT_THEME` fallback for unknown names). Gate-the-fix-off self-test passes (Action #1200/#1468): with the fix reverted, exactly 1 of the 3 new tests fails — the registry-first regression case — and the other 2 pass, confirming no tautology. Removes the need for the documented workaround (`_presets.THEME_PRESETS[name] = preset`) — consumers can now use the public `register_preset()` API alone.
+
 ## [1.0.0rc9] - 2026-05-22
 
 ### Fixed
--- a/python/djust/tests/test_theming_presets.py
+++ b/python/djust/tests/test_theming_presets.py
@@ -57,5 +57,79 @@ def test_invalid_preset():
     assert preset is None
 
 
+def test_get_preset_consults_runtime_registry_first_1595():
+    """Regression for #1595 — `presets.get_preset()` must consult the runtime
+    registry first so user-registered presets actually reach the CSS generator.
+
+    Before the fix: `get_preset()` read from the static `THEME_PRESETS` module
+    dict only, so a preset added via `register_preset()` was visible to the
+    manager / theme switcher / introspection (which consult the registry) but
+    INVISIBLE to the renderer (which goes through this function). The user's
+    `--primary` was silently replaced with the default slate-black palette.
+
+    Mirrors the registry-first dispatch in `theme_packs.get_theme_pack()`
+    (`python/djust/theming/theme_packs.py:1216-1222`).
+    """
+    from djust.theming import presets
+    from djust.theming.registry import get_registry, register_preset
+
+    sentinel_name = "_regression_1595_runtime_only"
+    sentinel = ThemePreset(
+        name=sentinel_name,
+        display_name="1595 sentinel",
+        description="present in runtime registry, NOT in static THEME_PRESETS",
+        light=THEME_PRESETS["default"].light,
+        dark=THEME_PRESETS["default"].dark,
+    )
+
+    # Pre-condition: not present in the static dict.
+    assert sentinel_name not in THEME_PRESETS, (
+        "test setup invariant violated — sentinel name must not exist statically"
+    )
+
+    register_preset(sentinel_name, sentinel)
+    try:
+        # The registry sees it.
+        assert get_registry().get_preset(sentinel_name) is sentinel
+
+        # The bug-of-record: presets.get_preset() must ALSO see it.
+        # Pre-fix this returns DEFAULT_THEME instead of `sentinel`.
+        result = presets.get_preset(sentinel_name)
+        assert result is sentinel, (
+            f"#1595: presets.get_preset({sentinel_name!r}) returned "
+            f"{result.name!r} instead of the runtime-registered preset. "
+            "Registry/static-dict divergence — the CSS generator would render "
+            "the wrong palette."
+        )
+    finally:
+        # Clean up so subsequent tests don't see the sentinel.
+        get_registry()._presets.pop(sentinel_name, None)
+
+
+def test_get_preset_static_dict_fallback_when_registry_lacks_name():
+    """Companion to #1595: static dict fallback must still work for built-in
+    presets that the runtime registry doesn't know about.
+
+    Locks in the second half of the registry-first-OR-static-fallback contract.
+    """
+    from djust.theming import presets
+
+    # "blue" is in the static dict; no test setup registers it at runtime.
+    result = presets.get_preset("blue")
+    assert result.name == "blue", f"static-dict fallback broke: got {result.name!r} for 'blue'"
+
+
+def test_get_preset_unknown_name_returns_default():
+    """Companion to #1595: when neither the registry nor the static dict has
+    the name, `get_preset()` falls back to DEFAULT_THEME (existing behavior).
+    """
+    from djust.theming import presets
+
+    result = presets.get_preset("definitely_does_not_exist_anywhere_1595")
+    assert result is presets.DEFAULT_THEME, (
+        f"unknown-name fallback broke: got {result.name!r} instead of DEFAULT_THEME"
+    )
+
+
 if __name__ == "__main__":
     pytest.main([__file__])
--- a/python/djust/theming/presets.py
+++ b/python/djust/theming/presets.py
@@ -156,8 +156,21 @@ THEME_PRESETS: dict[str, ThemePreset] = {
 
 
 def get_preset(name: str) -> ThemePreset:
-    """Get a theme preset by name, with fallback to default."""
-    return THEME_PRESETS.get(name, DEFAULT_THEME)
+    """Get a theme preset by name.
+
+    Resolution order, matching ``theme_packs.get_theme_pack()``:
+
+    1. Runtime registry (presets added via ``register_preset()``).
+    2. Built-in static ``THEME_PRESETS`` dict.
+    3. ``DEFAULT_THEME`` fallback.
+
+    Before #1595 only step 2 + 3 ran, so user-registered presets were visible
+    to the manager/registry/introspection but invisible to the CSS generator
+    that ultimately renders ``--primary`` etc. into ``:root``.
+    """
+    from .registry import get_registry
+
+    return get_registry().get_preset(name) or THEME_PRESETS.get(name, DEFAULT_THEME)
 
 
 def list_presets() -> list[dict]:


johnrtipton:
## Code Review — PR #1596

**Verdict: APPROVE** 🟢

Clean, minimal, faithfully mirrors the suggested fix shape and the proven `theme_packs.get_theme_pack()` pattern. TDD with a real gate-off self-test. Two-commit shape and stale-base both clean. Ship it.

---

### What I verified

1. **Faithfulness to suggested fix** (issue body's "Suggested fix" section vs `python/djust/theming/presets.py:158-176`). Diff is exactly:
   ```python
   from .registry import get_registry
   return get_registry().get_preset(name) or THEME_PRESETS.get(name, DEFAULT_THEME)
   ```
   Same shape as `python/djust/theming/theme_packs.py:1216-1222`. No drift.

2. **Circular import risk → lazy import is strictly required (not just defensive).** `python/djust/theming/registry.py:127` does `from .presets import THEME_PRESETS` inside `_do_discover()`. A top-level `from .registry import get_registry` in `presets.py` would risk a partial-init cycle. The in-function `import` at `presets.py:172` is the correct choice and matches `theme_packs.py:1219`.

3. **Test correctness — sentinel uniqueness.** `grep -rn "_regression_1595" python/ tests/` returns only the one test definition at `python/djust/tests/test_theming_presets.py:76`. No collision risk.

4. **Test correctness — cleanup.** `finally: get_registry()._presets.pop(sentinel_name, None)` at `test_theming_presets.py:107` correctly reaches into private state for test cleanup. Yes, it's a leaky abstraction, but it's the cleanest available option (no `unregister_preset()` public API exists). `pop(name, None)` is idempotent so cleanup is safe even if the registration itself failed mid-way.

5. **Gate-the-fix-off self-test claim (PR body / Action #1200/#1468).** Mentally running the 3 new tests against pre-fix code:
   - `test_get_preset_consults_runtime_registry_first_1595` → FAIL (sentinel not in `THEME_PRESETS`, returns `DEFAULT_THEME`)
   - `test_get_preset_static_dict_fallback_when_registry_lacks_name` → PASS (`"blue"` in `THEME_PRESETS`)
   - `test_get_preset_unknown_name_returns_default` → PASS (unknown name, unchanged)

   That's exactly 1-of-3 failing — claim verified, no tautology.

6. **Side-effect of registry-first dispatch (re-register a built-in name).** `Registry._do_discover()` already pre-loads `THEME_PRESETS` into `_presets` (`registry.py:127-130`), so a runtime `register_preset("blue", ...)` overwrites the built-in in the registry. Post-fix, the renderer also picks up that override. This matches `theme_packs.get_theme_pack()` behavior — consistent with the established "registry wins" design contract. Intentional.

7. **CHANGELOG** (`CHANGELOG.md:10`). Under `[Unreleased]`, correctly cites #1595, mirrored function (`theme_packs.get_theme_pack()` at line 1216-1222), the 3 test names, the gate-off self-test result, and Actions #1200/#1468. File:line citations spot-checked accurate.

8. **Multi-reopen rule (Action #1389).** `gh pr list --search "1595" --state all` returns only PR #1596. First attempt — TDD-with-gate-off is sufficient evidence.

9. **Stale-base check (Action #1450).** `git rev-list --count 357439a2..origin/main` = **0**. Branch tip is even with main. Safe to merge.

10. **Two-commit shape (Action #181).**
    - `8a4d2d8e` files: `python/djust/tests/test_theming_presets.py`, `python/djust/theming/presets.py` (no CHANGELOG.md ✅)
    - `357439a2` files: `CHANGELOG.md` only ✅

---

### Findings

None 🔴 / 🟡. One 🟢 observation worth keeping in the PR record:

🟢 **Registry-first becomes a superset of the static dict after discovery, but the fallback still matters.** `Registry._do_discover()` (`registry.py:124-145`) pre-loads `THEME_PRESETS` into `_presets`, so post-discovery, registry-first will find every built-in too. The `or THEME_PRESETS.get(...)` fallback is essential for the pre-discovery window (e.g., very early in `AppConfig.ready()`, or any code path that imports `presets.get_preset` before `Registry.discover()` runs). The fix correctly preserves that fallback rather than collapsing to registry-only, matching the `theme_packs.get_theme_pack()` pattern.

🤖 Reviewed by Claude

johnrtipton:
## Stage 14 retro — PR #1596 (#1595 theming registry-aware get_preset)

### What went well

- **Reporter shipped the fix in the issue body.** #1595 came with bit-exact reproducer (3 doctest-style code blocks showing the divergence), root cause (file:line of the broken function), suggested fix (3 lines mirroring an existing pattern at a cited file:line), and a workaround. The pipeline executed that fix verbatim — no investigation, no synthetic-shape divergence risk. This is the platonic ideal of a `/pipeline-run` bugfix: when the issue is this well-scoped, the loop is almost mechanical.
- **TDD red-then-green-then-gate-off — fully honored.** Wrote the failing regression test first; ran against main and confirmed it failed with the issue's exact symptom message; applied the fix; confirmed green; stashed the fix and verified exactly 1-of-3 new tests fails (no tautology) per Action #1200/#1468. The three-step empirical cycle took ~3 minutes wall-clock and produced a regression that locks the dispatch contract in place.
- **The mirrored-pattern shape is the right one.** `theme_packs.get_theme_pack()` already established the registry-first-OR-static-fallback dispatch in the codebase; mirroring it for presets means consumers can now reason about both functions identically. Stage 11 reviewer confirmed `theme_packs.py:1216-1222` is the cited reference pattern and the new `presets.get_preset()` matches it line-for-line modulo the dict variable name. Pattern consistency is its own form of correctness.
- **Stage 11 reviewer's "lazy import is strictly required" finding.** The reviewer didn't just rubber-stamp the lazy `from .registry import get_registry` inside the function — they grepped `registry.py` and confirmed it imports `THEME_PRESETS` from `presets`, making the lazy import strictly necessary (not just defensive). That kind of "why is this written this way" verification is what separates a real Stage 11 review from a cursory pass.

### What could be improved

- **Pre-commit hook bounced the first commit attempt** (the `&& git log -1 --oneline` reflex caught it correctly — Action #122 working as designed). Same pattern that's been canonized for a year now. Re-staged + retried and it landed clean. The hook reformatted the multiline assertion message; no real fix needed beyond the re-stage. Worth filing yet another candidate rule: when pre-commit ruff is going to reformat a file that's already staged, the formatter could rewrite the staged blob rather than the working tree to eliminate the stash-restore-bounce cycle entirely. But that's an upstream-pre-commit-framework concern, not a djust concern.
- **The bug had been latent since theme_packs.get_theme_pack() was written with the correct pattern.** Whoever wrote theme_packs (registry-first dispatch) did the right thing; whoever wrote presets (static-only dispatch) did the wrong thing. Both functions were committed; nothing flagged the asymmetry until a consumer hit it in production. Worth filing a candidate action-tracker rule: when adding a new "registry-first dispatch" function, grep for sibling registry-aware getters in the same module/subpackage to verify they all follow the same dispatch shape. The smaller version of this rule (read/write gate symmetry for state-persistence PRs) was filed in PR #1565's retro; this is the lookup-API-pair generalization.
- **The reporter cited their workaround AND the suggested fix.** That's gold. The cost was ~1 hour of their debugging time to locate the divergence (their words). The framework SHOULD have caught this asymmetry earlier — either via the "mirror sibling dispatch shape" rule above, or via an integration test that asserts every registered preset reaches the CSS generator. The integration-test gap is the deeper fix: there's no current test that says "every preset visible to the registry must be visible to `get_preset()`." Worth filing as a follow-up issue for the theming subsystem — a meta-invariant test that catches this entire bug class.

### Process verification

- Branch-verify reflex ran before each commit
- Post-commit `git log -1 --oneline` reflex caught the FIRST commit attempt's bounce (pre-commit ruff reformat); re-stage + retry landed clean (`8a4d2d8e` impl, `357439a2` CHANGELOG)
- Two-commit shape (#181) Gate 1 + Gate 2 both green programmatically
- Stale-base check (Action #1450) confirmed at Stage 11 — 0 commits behind origin/main
- Stage 11 Code Review verdict: **APPROVE** (1 🟢 observation, 0 🔴, 0 🟡)
- Stage 12 (Review Verdict): no findings to address, Re-Review correctly skipped per template's skip_if condition
- Stage 13 Merge: `gh pr merge 1596 --repo djust-org/djust --squash --delete-branch --admin`
- Issue #1595 auto-closed via "closes #1595" in commit message

### Followups

- **Candidate Action Tracker rule**: when adding a new "registry-first dispatch" function in a subpackage, grep for sibling registry-aware getters in the same subpackage and verify dispatch-shape symmetry. Generalizes PR #1565's "read/write gate symmetry" rule to the lookup-API-pair class.
- **Candidate follow-up issue (theming integration test)**: add a meta-invariant test that asserts every preset added via `register_preset()` is reachable via `presets.get_preset()`, every pack added via `register_theme_pack()` is reachable via `theme_packs.get_theme_pack()`, etc. — across all theming registry/lookup pairs. Catches this entire bug class structurally.
- **Cut v1.0.0rc10 or roll into v1.0.0 GA?** This is a low-severity bug with a documented workaround; doesn't itself force a new RC. But if any other v1.0 fixes accumulate, this should ship with them. Decision deferred to release-planning.

### Related

- Closes #1595 (the bug)
- Mirrors `python/djust/theming/theme_packs.py:1216-1222` (the reference pattern)
- Followups: candidate Action Tracker rule (registry-dispatch symmetry); candidate theming meta-invariant test issue

`RETRO_COMPLETE`

PR #1598, opened 2026-05-25
test(theming): meta-invariant — every register_X() must be reachable via get_X() (closes #1597)

### LINKED ISSUE #1597: theming: meta-invariant test — every register_X() must be reachable via get_X() (catches #1595 bug class)
## Problem

#1595 caught a real divergence: `register_preset()` adds to `Registry._presets`, but `presets.get_preset()` read from a static module dict. The CSS generator went through the static dict and silently rendered the wrong palette. The fix (PR #1596) gates `get_preset()` to consult the registry first, mirroring `theme_packs.get_theme_pack()`.

But the underlying bug class — **a runtime registration API that the matching getter ignores** — could exist anywhere in the theming subsystem. There's no current test that asserts the round-trip works for all `register_X()` ↔ `get_X()` pairs.

This is a structural gap. The next person to add a `register_*` API to a sibling theming module could repeat the same pattern; we won't catch it until a consumer files an issue.

## Proposed test

A single parametrized test in `python/djust/tests/test_theming_register_get_roundtrip_invariant.py` that:

1. Enumerates every `register_X()` public API in `djust.theming.registry` (e.g., `register_preset`, `register_theme_pack`, `register_design_system`, `register_theme`, etc.).
2. For each, identifies the corresponding `get_X()` public API in the matching module (`presets.get_preset`, `theme_packs.get_theme_pack`, etc.) via a simple naming convention or an explicit table.
3. Constructs a sentinel object, calls `register_X(name, sentinel)`, then asserts `get_X(name) is sentinel` (or equivalent).
4. Cleans up via `Registry._<plural>.pop(name, None)`.

Failure mode this catches: any new `register_X()` API whose matching `get_X()` doesn't consult the registry. That's exactly the #1595 shape, generalized.

## Suggested table form

```python
ROUNDTRIP_PAIRS = [
    # (register_func_name, get_func_module, get_func_name, sentinel_factory)
    ("register_preset", "djust.theming.presets", "get_preset", _make_sentinel_preset),
    ("register_theme_pack", "djust.theming.theme_packs", "get_theme_pack", _make_sentinel_pack),
    # ... etc for every register_X in djust.theming.registry
]
```

That's a single ~80-line test file (or a fixture-based test class) that scales as new register APIs are added.

## Acceptance criteria

- [ ] Enumerate ALL `register_*` functions currently exported from `djust.theming.registry`. Use `grep -nE '^def register_' python/djust/theming/registry.py` as the source of truth.
- [ ] For each, identify the matching getter. If the getter doesn't exist or the dispatch shape is unclear, the test should fail-loud with a guidance message naming the convention violation (rather than silently skipping).
- [ ] Parametrized test that runs the registration → lookup → cleanup cycle for each pair.
- [ ] If any current pair fails the roundtrip, that's another bug to fix BEFORE the test merges (same shape as #1595).

## Why this matters

The framework's theming docs explicitly tell consumers to use `register_X()` in `AppConfig.ready()`. When a consumer follows that documented API and silently gets the wrong output, the trust cost is significant — exactly the friction the #1595 reporter quoted (\"loses an hour of debugging\"). A structural test means the next `register_X()` API added to the subsystem can't ship broken.

## Related

- Closed via this issue's child PR: TBD
- Bug-class case study: #1595 (fixed by PR #1596)
- Reference pattern: `theme_packs.get_theme_pack()` (`python/djust/theming/theme_packs.py:1216-1222`)
- Process retro (canonization candidate): https://github.com/djust-org/djust/pull/1596#issuecomment-4530910320 — the registry-dispatch-symmetry rule. This issue is the empirical test backing for that rule.

## Severity

Low-medium. No active bug to fix; the test is structural prevention for a bug class that has cost one production issue (#1595) and ~1 hour of consumer debugging. Worth a 1-2 hour PR.

🤖 Filed by Claude after PR #1596's retrospective.

johnrtipton:
## Summary

Closes #1597 — structural prevention for the #1595 bug class.

PR #1596 fixed `presets.get_preset()` to consult the runtime registry first (the specific #1595 bug). This PR adds the meta-invariant test that catches the **bug class** — any future `register_X()` API in `djust.theming.registry` whose matching module-level `get_X()` ignores the registry.

## What the test does

`python/djust/tests/test_theming_register_get_roundtrip_invariant.py` carries an authoritative `ROUNDTRIP_PAIRS` table covering the 3 current pairs:

| register API | matching getter |
|---|---|
| `register_preset` | `djust.theming.presets.get_preset` |
| `register_design_system` | `djust.theming.theme_packs.get_design_system` |
| `register_theme_pack` | `djust.theming.theme_packs.get_theme_pack` |

A parametrized test (`test_register_get_roundtrip_invariant_1597`) asserts the roundtrip works for each pair:
1. Register a sentinel `object()` under a unique name
2. Call the module-level getter
3. Assert it returns the sentinel by identity (not equality)
4. Cleanup via `Registry._<dict>.pop(name)`

A second `test_no_unaudited_register_apis_1597` AUDIT test grep-walks `djust.theming.registry`'s public surface and fails loud if any `register_*` function exists without a row in `ROUNDTRIP_PAIRS` — so a new register API can't ship without either coverage or an explicit decision to opt it out.

## Verification — gate-the-fix-off self-test (Action #1200/#1468)

Reverted `presets.get_preset` to its pre-#1596 broken shape (static-dict only, no registry consult). Re-ran the new test file:

- **Exactly 1 of 4 cases failed**: `test_register_get_roundtrip_invariant_1597[register_preset->get_preset]`
- Failure message: *"djust.theming.presets.get_preset does not consult the runtime registry. Apply the registry-first dispatch pattern from theme_packs.get_theme_pack() (#1596 fixed this exact divergence for presets)."*
- The other 2 parametrized cases (`register_design_system`, `register_theme_pack`) + the audit case all continued to pass — confirming no tautology and each parametrized case is independently load-bearing.

Restored. All 4 green.

## Failure UX

When a future getter regresses (or a new register API ships without a registry-aware getter), the test failure names the specific module/function that broke the contract and pointer-references the canonical fix pattern — so a Stage 11 reviewer or implementer can fix it directly from the error message without re-reading the issue.

## Why this matters

The framework documents `register_X()` in `AppConfig.ready()` as the way to add custom presets/packs/design-systems. When that documented API silently doesn't reach the renderer, the trust cost is significant — the #1595 reporter quoted "loses an hour of debugging." A structural test means the next `register_X()` API added to the subsystem can't ship broken.

## Test plan

- [x] 4 new test cases in `test_theming_register_get_roundtrip_invariant.py`, all pass on main
- [x] Gate-the-fix-off self-test (#1200/#1468) — pre-#1596 shape produces exactly 1 failing case
- [x] All 1815 theming tests pass (was 1811 pre-PR; +4 new)
- [x] Full project suite green
- [x] Two-commit shape (#181): test in `dc0041f5`, CHANGELOG in `36ebe571`. Gate 1 + Gate 2 verified
- [x] Stale-base check: 0 commits behind `origin/main`

## Refs

- Closes #1597 (the issue I filed earlier today after PR #1596's retro)
- Bug-class case study: #1595 (fixed by #1596)
- Reference pattern: `python/djust/theming/theme_packs.py:1216-1222` (`get_theme_pack`)
- Generalizes PR #1565's "read/write gate symmetry" canon to the lookup-API-pair class


johnrtipton:
## Stage 11 Review — PR #1598

**Verdict: APPROVE**

Meta-invariant test PR closing #1597. Structural prevention for the #1595 bug class (register/get registry-dispatch divergence). All review priorities check out clean; no blocking findings.

---

### What I verified

**1. ROUNDTRIP_PAIRS completeness (priority #1) — ✅**
```
$ grep -nE '^def register_' python/djust/theming/registry.py
239:def register_preset(name: str, preset) -> None:
257:def register_design_system(name: str, design_system) -> None:
270:def register_theme_pack(name: str, pack) -> None:
```
All 3 registry-module `register_*` APIs are in `ROUNDTRIP_PAIRS` (test file lines 70–92). No unaudited register API. The audit test `test_no_unaudited_register_apis_1597` (lines 161–188) does the both-directions check (unaudited AND stale), so this stays self-defending against future drift.

**2. Getter-module references correct (priority #2) — ✅**
```
$ grep -nE "^def get_(preset|design_system|theme_pack)" python/djust/theming/{presets,theme_packs}.py
python/djust/theming/presets.py:158:def get_preset(name: str) -> ThemePreset:
python/djust/theming/theme_packs.py:824:def get_design_system(name: str) -> Optional[DesignSystem]:
python/djust/theming/theme_packs.py:1216:def get_theme_pack(name: str) -> Optional[ThemePack]:
```
All 3 getters exist at the cited modules. All 3 are confirmed registry-first (`get_preset`: `presets.py:158-170`; `get_design_system`: `theme_packs.py:824-831`; `get_theme_pack`: `theme_packs.py:1216-1222`).

**3. `registry_dict_attr` cleanup correctness (priority #3) — ✅**
Verified at `registry.py:33-35` — the `ThemeRegistry.__new__` initializer creates exactly `inst._presets = {}`, `inst._themes = {}`, `inst._packs = {}`. Matches all 3 `registry_dict_attr` values in `ROUNDTRIP_PAIRS`. Cleanup at test lines 113–120 will actually pop the sentinel, not silently no-op.

**4. Audit test thoroughness (priority #4) — ✅**
Both directions present: unaudited (line 176) and stale (line 185). Stale check correctly defends against the #1391/#1400 symbol-migration grep failure mode — a future registry refactor that drops a `register_X` will fail the table immediately rather than leaving phantom coverage.

**5. Sentinel choice (priority #5) — ✅**
`object()` is appropriately documented inline at lines 137–140: truthy (for `or` short-circuit), identity-unique (for `is` assertion), unique name (no cross-run collision). The registry doesn't currently type-validate, so `object()` reaches the dict without TypeError. Trade-off is explicitly stated.

**6. `or` short-circuit dependency (priority #6) — ✅ acceptable**
PR body and test docstring (lines 122–126) correctly scope the test to the #1595 bug class (getter ignores registry entirely), not the priority-inversion variant. Priority-inversion would be a different bug class with no precedent in this subsystem; deferring it to "if we ever ship that bug, file then" is correct YAGNI. The current test catches exactly what shipped broken.

**7. Cleanup race / test isolation (priority #7) — ✅**
`Registry` is a module-level singleton (`registry.py:25-39`, double-checked-locked `__new__`). pytest-xdist is installed (pyproject.toml:107) but NOT enabled by default (no `-n` in `addopts` at pyproject.toml:152–158). Tests run serially in-process. `try/finally` cleanup (lines 144–146) executes on assertion failure too. Sentinel name is unique per pair (`f"_regression_1597_{pair.register_fn}"`, line 133), so two parametrized cases can't collide. No test-isolation risk under current config.

**8. CHANGELOG accuracy (priority #8) — ✅**
PR body claim "exactly 1 of the 4 new test cases fails" under gate-off matches CHANGELOG.md (the `register_preset→get_preset` case specifically). The empirical claim is internally consistent.

**9. Two-commit shape (#181, priority #9) — ✅**
```
dc0041f5: python/djust/tests/test_theming_register_get_roundtrip_invariant.py  (no CHANGELOG)
36ebe571: CHANGELOG.md                                                          (only CHANGELOG)
```
Both gates pass.

**10. Stale-base check (Action #1450, priority #10) — ✅**
`git rev-list --count HEAD..origin/main` → `0`. Branch is at tip.

**11. Doc-claim citations (Action #1071/#1197, priority #11) — ✅**
- `theme_packs.py:1216` cited 4× → confirmed as `def get_theme_pack(name: str)`.
- PR #1596 (preset fix) and PR #1565 (read/write gate symmetry) referenced in PR body — both real.
- Action #1200/#1468 (gate-off self-test) referenced — both canonical in CLAUDE.md.

---

### Findings

🟢 **No 🔴 / 🟡 findings.** All review priorities verified clean.

🟢 **Notable strengths:**
- The failure UX on the parametrized test (lines 147–154) names the specific module/function pair AND pointer-references the canonical fix (`theme_packs.get_theme_pack()`) so a future implementer can fix it directly from the error message without re-reading #1595.
- The audit test does BOTH directions of the staleness check — this is the right shape and isn't an obvious default. Without the stale check (line 185), a registry refactor that removes a `register_X` would leave the table claiming coverage that doesn't exist.
- Two `_resolve` assertions (lines 100–108) fail with actionable guidance ("update ROUNDTRIP_PAIRS") rather than `AttributeError`. Good operator-empathy under future drift.
- The gate-off self-test (Action #1200/#1468) was actually run and the failure case named in the PR body — not a perfunctory "verified" checkbox.

🟢 **One observation for future PRs (not blocking):** The test pulls the registry singleton across tests in the same process. If at some future point `pytest -n auto` becomes the default OR `ThemeRegistry._instance` is reset between tests via a fixture, the assumptions in this file (singleton, no concurrent writers) deserve a re-check. Today both assumptions hold; just worth noting.

---

### Decision

**APPROVE.** Clean test-only PR with verified two-commit shape, no stale base, complete coverage of the bug class, defensible sentinel choice, both-directions audit, and a real gate-off self-test (not a tautology). Ship it.

johnrtipton:
## Stage 10 retro — PR #1598 (#1597 theming register/get meta-invariant)

### What went well

- **Same-session closure of the bug-class case study.** PR #1596 fixed #1595 (specific bug); the retro on #1596 surfaced "we need structural prevention for this bug class" as a candidate followup; I filed #1597; the user said "yes file it"; the user immediately invoked `/pipeline-ship` on #1597; this PR ships the structural prevention. Three issues + three PRs in one focused arc, each one closing the loop on the previous one's "what could be improved." That's the cadence the action-tracker + retro system are designed to produce, working as intended.
- **Test as DOCUMENTATION of the bug class.** The new test file's docstring (`test_theming_register_get_roundtrip_invariant.py:1-32`) explicitly names the bug class, cites #1595 as the case study, points at the canonical fix pattern (`theme_packs.get_theme_pack` at file:line), and tells the next maintainer exactly what to do when a new `register_*` API lands ("add a row to `ROUNDTRIP_PAIRS`"). The maintenance instructions are co-located with the test — no separate `CONTRIBUTING.md` section to forget. If a future Stage 11 reviewer ever asks "why does this file exist," the file answers them directly.
- **Both-directions audit was the right shape.** The audit test checks BOTH `actual - audited` (unaudited) AND `audited - actual` (stale). The stale check seemed redundant at first (why would the test reference a removed API?) but the Stage 11 reviewer's framing was correct: a registry refactor that REMOVES a register API would leave the table silently claiming coverage that doesn't exist. The stale check makes the table a true mirror of the registry surface, not a one-way audit.
- **Gate-the-fix-off self-test, automated via in-place file rewrite.** Action #1200/#1468 (the gate-off rule) was applied empirically — I scripted reverting `presets.py:158-175` to the pre-#1596 broken shape with a small `python -c "..."` snippet, ran the new test file, confirmed exactly 1-of-4 cases fails with the expected error message, then restored. This is the third PR in 3 days where the gate-off self-test has been the load-bearing evidence that the new tests aren't tautological. The pattern is now muscle memory.

### What could be improved

- **The test takes `object()` as the sentinel — a slight epistemic gap.** Using `object()` rather than a real `ThemePreset`/`ThemePack`/`DesignSystem` means the test exercises the dispatch shape but not type-aware code paths. The registry doesn't currently type-validate, so this is fine TODAY — but if a future PR adds `isinstance` checks inside the getters (e.g., to defensively reject malformed registrations), the test would silently miss type-coercion bugs because `object()` is never a valid instance. Worth noting as a future-strengthening candidate: parameterize the sentinel factory per-pair (each row provides `_make_sentinel_preset()` etc. that returns a minimal valid instance). The Stage 11 reviewer flagged this as the explicit trade-off the test docs should acknowledge — and the test's `_resolve()` docstring does note "the registry doesn't currently validate types — it just stores opaque references." So the gap is documented, but it's still a gap.
- **The test does NOT exercise priority** (registry-first vs static-fallback ordering). PR #1596's fix could be regressed to `static_dict.get(name) or registry.get(name)` (reverse priority) and this test would still pass — because the sentinel is never in the static dict. A "registry wins over same-name static entry" test would catch that specific regression shape. Decision: deferred. #1595 was specifically about a getter that didn't consult the registry AT ALL, not about wrong priority; the current test covers the actual historical bug class. If a future PR ever changes a getter's `or` direction, that's a more-overt regression that the existing PR-review process should catch on its own. Re-evaluate if it ever DOESN'T get caught.
- **Pre-commit hook bounced again.** That's twice in two PRs. Action #122 reflex caught it correctly. Filing once again as candidate for the upstream-pre-commit-framework's "reformat in stash" rather than "reformat in worktree" mode — but that's outside djust's scope. The djust-side mitigation (`&& git log -1 --oneline` after every commit) is doing its job.

### Process verification

- Branch-verify reflex confirmed before each commit
- Post-commit `git log -1 --oneline` reflex on both `dc0041f5` and `36ebe571` — neither bounced this time
- Two-commit shape (#181) Gate 1 + Gate 2 both green
- Stale-base check (Action #1450) confirmed at Stage 7 (Code Review) — 0 commits behind `origin/main`
- Gate-off self-test (Action #1200/#1468) executed empirically — exactly 1-of-4 cases failed under pre-#1596 shape
- Stage 7 Code Review verdict: **APPROVE** (0 🔴, 0 🟡, 0 🟢 — clean)
- Stage 8 (Review Verdict): no findings, Re-Review skipped per template condition
- Stage 9 Merge: `gh pr merge 1598 --repo djust-org/djust --squash --delete-branch --admin`
- Issue #1597 auto-closed via "closes #1597"
- Stage 10 (this comment): retro posted

### Where this fits in the v0.9.x → v1.0.0 process arc

This is the THIRD action-tracker rule canonized in 3 days (across PRs #1565, #1596, #1598):

1. **PR #1565 retro**: "reviewer-check for read/write gate symmetry on state-persistence PRs" (candidate)
2. **PR #1596 retro**: "registry-dispatch symmetry: when adding a registry-first lookup, grep sibling lookups for shape consistency" (candidate)
3. **PR #1598 (this PR)**: STRUCTURAL test that mechanically enforces #2 for the theming subsystem. The rule went from "candidate canon" → "executable structural prevention" in one cycle.

The pattern worth canonizing: when a retro surfaces "we need a STRUCTURAL test for this bug class," file it immediately as a follow-up issue. The cost (an issue + a small PR) is small; the prevention value across the next year of `register_*`-style additions is significant. Filed as a candidate action-tracker rule for v1.0.0 GA-ready retro.

### Related

- Closes #1597 (this issue)
- Bug-class case study: #1595 (fixed by #1596)
- Reference pattern: `python/djust/theming/theme_packs.py:1216-1222` (`get_theme_pack`)
- 3-PR arc this session: #1596 → #1597 → #1598 (specific bug → followup issue → structural prevention)

`RETRO_COMPLETE`

PR #1600, opened 2026-05-25
fix(navigation): guard against cross-origin pushState (closes #1599)

### LINKED ISSUE #1599: fix(navigation): handleLiveRedirect/handleLivePatch throw SecurityError on cross-origin paths (should fall back to full-page nav)
## Symptom

Reported from production djust.org. Clicking an element with `dj-navigate=\"https://djustlive.com/\"` (or any cross-origin URL) throws:

\`\`\`
Uncaught SecurityError: Failed to execute 'pushState' on 'History':
A history state object with URL 'https://djustlive.com/' cannot be created
in a document with origin 'https://djust.org' and URL 'https://djust.org/examples/'.
    at e (client.js:9269:24)
    at HTMLAnchorElement.<anonymous> (client.js:9576:17)
\`\`\`

## Root cause

`python/djust/static/djust/src/18-navigation.js:69`:

\`\`\`javascript
const newUrl = new URL(data.path, window.location.origin);
\`\`\`

When `data.path` is an absolute URL like `https://djustlive.com/`, the second argument to `URL()` (the base) is **ignored**, so `newUrl` becomes the cross-origin URL. Line 81 then attempts `window.history.pushState(..., newUrl.toString())` which the browser correctly refuses with `SecurityError` (history API forbids same-document cross-origin pushState).

Same shape at `handleLivePatch()` line 36 — if a server sends `live_patch` with a cross-origin `path`, it crashes the same way.

`_executePatch()` (the `dj-patch` click handler at line 301) is incidentally safe because it builds `newUrl` from `window.location.href` and only overwrites pathname when patchValue starts with `/`. Still worth an explicit cross-origin guard there for defense-in-depth, but no current bug.

## Fix shape

Before pushState in `handleLivePatch` and `handleLiveRedirect`, detect cross-origin and fall back to full-page navigation:

\`\`\`javascript
if (newUrl.origin !== window.location.origin) {
    // pushState forbids cross-origin URLs; the caller's intent is to
    // navigate, so do a real navigation instead of crashing.
    window.location.href = newUrl.toString();
    return;
}
\`\`\`

Preserves intent (the user/server wanted to go there) and avoids the SecurityError.

## Why this matters

djust.org's templates legitimately link to djustlive.com (the PaaS sister site). Any reasonable `<a dj-navigate=\"https://...\">` to a cross-origin destination triggers this — it's a hard production crash, not a corner case.

## Severity

Medium. Single-page crashes the JS runtime so subsequent navigation may misbehave; user sees no visible navigation. Workaround: drop `dj-navigate` and use a plain `<a href>`. But framework's job is to handle this gracefully.

## Reporter

@johnrtipton (in-session diagnosis from djust.org production console).

johnrtipton:
## Summary

Closes #1599 — production djust.org crash when clicking cross-origin `dj-navigate` links.

## Symptom (from reporter's console)

```
Uncaught SecurityError: Failed to execute 'pushState' on 'History':
A history state object with URL 'https://djustlive.com/' cannot be created
in a document with origin 'https://djust.org' and URL 'https://djust.org/examples/'.
    at e (client.js:9269:24)                    ← handleLiveRedirect's pushState
    at HTMLAnchorElement.<anonymous> (client.js:9576:17)   ← dj-navigate click handler
```

## Root cause

`python/djust/static/djust/src/18-navigation.js:69`:

```javascript
const newUrl = new URL(data.path, window.location.origin);
```

When `data.path` is an absolute URL like `https://djustlive.com/`, the second argument to `URL()` (the base) is **ignored**, so `newUrl` becomes the cross-origin URL. The subsequent `pushState` at line 81 then triggers the browser's same-origin policy.

Same shape exists in `handleLivePatch()` at line 36 — if a server ever sends `live_patch` with a cross-origin path, it would throw the same way.

`_executePatch()` (the `dj-patch` click handler at line 301) was incidentally safe because it builds `newUrl` from `window.location.href` and only overwrites pathname when `patchValue.startsWith('/')`. Worth a defense-in-depth grep at Stage 11 for similar shapes.

## Fix

Detect cross-origin BEFORE pushState in both `handleLivePatch` and `handleLiveRedirect`; fall back to `window.location.href = newUrl.toString()` (full-page navigation = caller's intent):

```javascript
if (newUrl.origin !== window.location.origin) {
    if (globalThis.djustDebug) {
        console.log('[LiveView] live_redirect cross-origin → full-page nav: %s', newUrl.toString());
    }
    window.location.href = newUrl.toString();
    return;
}
```

`handleLiveRedirect` also stops the page-loading bar before the full nav (the browser's own progress indicator takes over).

Preserves all same-origin behavior.

## TDD verification

- **Red phase**: wrote 3 new tests in `tests/js/navigation.test.js` under `issue #1599 — cross-origin pushState guard`; ran against main (pre-fix) → 2 of 3 new tests FAIL (the cross-origin cases) with the expected pushState-was-called-with-djustlive.com finding.
- **Green phase**: applied the fix in both handlers → all 24 navigation tests pass.
- **Gate-the-fix-off self-test (Action #1200/#1468)**: stripped both guards from source → exactly 2 of 24 fail (the cross-origin cases); 22 other tests (including the same-origin backstop) continue to pass. No tautology.

## Tests

3 new cases in `tests/js/navigation.test.js`:

1. `handleLiveRedirect with cross-origin path does NOT call pushState with cross-origin URL`
2. `handleLivePatch with cross-origin path does NOT call pushState with cross-origin URL`
3. `same-origin paths still use pushState (regression backstop for guard breadth)` — catches an over-broad guard that rejects same-origin navigation

Full suite:
- 24 navigation tests pass (was 21, +3)
- 1612 total JS tests pass (was 1609, +3)
- 4390 Python tests pass
- Rust tests pass

## Two-commit shape

- `4d3a4b6d` impl + tests + rebuilt bundles (no CHANGELOG) — Gate 1 verified
- `21dc8905` CHANGELOG only — Gate 2 verified

## Bundle rebuild

`make build-js` rebuilt `client.js`, `client.min.js`, `client.min.js.gz`, `client.min.js.br`, `client.min.js.map`. All staged together in the impl commit.

## Refs

- Closes #1599
- Source: `python/djust/static/djust/src/18-navigation.js` `handleLivePatch` + `handleLiveRedirect`


johnrtipton:
## Stage 11 Review — PR #1600 (closes #1599)

### Verdict: **APPROVE**

### Findings

🟢 **Symmetric guards present in both handlers.** `handleLivePatch` (src/18-navigation.js:52-64) and `handleLiveRedirect` (src/18-navigation.js:94-110) both got the cross-origin guard. Both: (a) check `newUrl.origin !== window.location.origin`, (b) gate log on `globalThis.djustDebug`, (c) `window.location.href = newUrl.toString()`, (d) `return` before pushState. `handleLiveRedirect`'s guard additionally calls `window.djust.pageLoading.stop?.()` (line 105-107) to cancel the loading bar — present and correct.

🟢 **`_executePatch` incidentally-safe claim verified.** Read `src/18-navigation.js:336-360`. With `patchValue = "https://djustlive.com/"`: `url` becomes cross-origin (line 342), but `newUrl = new URL(window.location.href)` stays same-origin (line 345). The `for…of url.searchParams` loop (line 346) copies zero params (the cross-origin URL has none). `patchValue.startsWith('/')` is FALSE for `https://...` (line 349), so `newUrl.pathname` is NOT overwritten. `newUrl.toString()` stays same-origin. No third guard needed. PR body's defense-in-depth grep follow-up is the right disposition.

🟢 **Bundle rebuild consistent.** `git show 4d3a4b6d --stat` confirms `client.js`, `client.min.js`, `client.min.js.gz`, `client.min.js.br`, `client.min.js.map` all rebuilt alongside the source change. `git show 4d3a4b6d:client.js | grep -n cross-origin` returns 4 hits at lines 9240, 9247, 9282, 9289 (2 comments + 2 log strings, one pair per guard). Both guards present in the deployed bundle.

🟢 **Test depth and gate-off claim accurate.** 24 `it()` blocks in navigation.test.js (was 21, +3). Two negative cases (cross-origin → no cross-origin pushState) at tests/js/navigation.test.js:393-410 and :412-428; one positive same-origin backstop at :430-449. Mental gate-off: reverting both guards causes the 2 negative cases to attempt cross-origin pushState (the original bug) → 2 of 24 fail. CHANGELOG's "exactly 2 of 24" matches.

🟢 **JSDOM workaround is a real limitation, not a smell.** `grep -n "JSDOM" tests/js/navigation.test.js` shows the pattern is already established at line 33, 38, 51, 56 of the existing test helper (`origPushState try/catch` and `origReplaceState try/catch`) — the new `try { ... } catch (_e)` at :401-403 and :420-422 follows the same convention. The load-bearing assertion is "pushState NOT called with cross-origin URL" (which is what the bug was), not "no exception thrown" — assertion shape is correct.

🟢 **Same-origin regression backstop uses canonical setup.** Test at :430-449 uses `createEnv('<div dj-view="myapp.views.DashboardView"></div>')` and `_routeMap`, mirroring existing `live_redirect` tests in the file. Regression-detection is complete.

🟢 **Inline comments self-explanatory.** Source comments at src/18-navigation.js:49-52 and :91-94 explain *why* (pushState forbids cross-origin; browser throws SecurityError; caller's intent is to navigate). Citation `(#1599)` provided. A future reader doesn't need to read the issue.

🟢 **Two-commit shape (#181) verified.** Impl commit `4d3a4b6d` has 7 files (source + tests + 5 bundle artifacts, NO CHANGELOG). Doc commit `21dc8905` has 1 file (CHANGELOG.md only). Clean split.

🟢 **Stale-base check (Action #1450).** `git rev-list --count HEAD..origin/main` returned 0. Branch is current.

🟢 **Bundle size delta well under budget.** `client.min.js.gz`: 54077 → 54166 bytes = +89 bytes gzipped for 2 guards + log strings. Far under the "<2 KB per new module" project budget.

🟢 **CHANGELOG accuracy.** Entry correctly cites the symptom, root cause (`URL()` base ignored for absolute URLs), fix shape (guard + full-page nav fallback), test names, and gate-off count (2 of 24). Cross-checked against actual diff — no drift.

### What I verified

1. Read issue #1599 body, PR #1600 body, source diff, test diff, CHANGELOG diff.
2. Walked the `_executePatch` data flow with `patchValue = "https://djustlive.com/"` to confirm the "incidentally safe" claim.
3. Confirmed bundle artifacts (`client.js` + 4 minified variants) rebuilt and contain both guards.
4. Counted `it()` blocks in navigation.test.js to validate the "2 of 24" gate-off arithmetic.
5. Confirmed JSDOM try/catch pattern is pre-existing convention in the file.
6. Confirmed two-commit shape per Action #181.
7. Confirmed branch is not stale vs `origin/main` per Action #1450.
8. Measured gzipped bundle delta (+89 bytes).

### Specific line citations

- `python/djust/static/djust/src/18-navigation.js:52-64` — `handleLivePatch` cross-origin guard
- `python/djust/static/djust/src/18-navigation.js:94-110` — `handleLiveRedirect` cross-origin guard (with `pageLoading.stop?.()`)
- `python/djust/static/djust/src/18-navigation.js:336-360` — `_executePatch` (incidentally safe via line 349 `startsWith('/')`)
- `tests/js/navigation.test.js:393-449` — 3 new regression tests under `issue #1599 — cross-origin pushState guard`

Ready to merge.

johnrtipton:
## Stage 14 retro — PR #1600 (#1599 cross-origin pushState guard)

### What went well

- **Real production bug, real production fix, same session.** User reported the SecurityError verbatim from djust.org's production console. Stack trace pointed at `client.js:9269` + `client.js:9576` — bundle line numbers, not source. Curled the deployed `client.js`, mapped the lines to `handleLiveRedirect`'s pushState and the `dj-navigate` click handler at source `18-navigation.js:81` and `:383`. From symptom to root-cause file:line in under 5 minutes — the bundle-line-to-source-line workflow paid off.
- **TDD with empirical gate-off.** Red phase: 3 new tests added, ran against pre-fix main → 2 of 3 FAIL with the expected "pushState called with djustlive.com" finding. Green phase: applied symmetric guards to both `handleLivePatch` AND `handleLiveRedirect`, all 24 navigation tests pass. Gate-off self-test (Action #1200/#1468): stripped BOTH guards with a one-liner Python regex, rebuilt bundle, re-ran tests → exactly 2-of-24 fail (the cross-origin cases) → restored. The "same-origin still works" backstop test stayed green throughout — confirming the guard isn't over-broad.
- **`_executePatch` analysis done correctly.** A casual reading would have added a third guard "for safety." The careful walkthrough (`patchValue = "https://djustlive.com/"` → `url` is cross-origin, `newUrl` starts as same-origin, `startsWith('/')` is FALSE so pathname is NOT overwritten → `newUrl.toString()` stays same-origin) showed the existing logic already handles it. The PR body documented this analysis explicitly so future readers don't add a redundant guard. Reviewer subagent independently confirmed the walkthrough.
- **Bundle rebuild was non-negotiable.** A source change to `src/18-navigation.js` without rebuilding `client.js`/`client.min.js`/`client.min.js.gz`/`client.min.js.br`/`client.min.js.map` would ship a no-op fix — the production deploy serves `client.js`, not the source. All 5 build artifacts staged in the same impl commit (`4d3a4b6d`). Stage 11 reviewer verified the new strings appear in the rebuilt `client.js` via grep count.

### What could be improved

- **JSDOM `Object.defineProperty(window, 'location', ...)` is rejected, took one iteration to discover.** My first test attempt tried to spy on `window.location.href = X` by replacing the location object. JSDOM blocked it. Pivoted to the simpler shape: `try { handler(...) } catch (_e) { /* JSDOM may throw on location assignment */ }` — then assert "no cross-origin pushState happened." That's the LOAD-BEARING invariant anyway (the original bug was pushState throwing SecurityError, not the absence of a navigation). The first-try friction is the kind of thing that goes in a "JSDOM gotchas" reference doc — but the test file's pre-existing `try/catch` around `window.eval(clientCode)` already establishes this pattern, so the lesson is "look at how the same test file handles similar cases before writing fresh."
- **Pre-commit hook bounced again (3rd in 3 PRs).** Action #122 reflex caught it, re-stage + retry landed clean. The pattern is utterly consistent: pre-commit ruff/prettier reformats, stash-restore conflicts, commit silently bounces. The djust-side mitigation (`&& git log -1 --oneline` reflex) is doing its job; the upstream-side fix would be "reformat in stash" rather than "reformat in worktree" — out of scope.
- **The guard is conceptually a one-line fix (`if origin differs, fall back`); the actual diff is longer because of the loading-bar cleanup in handleLiveRedirect.** That's correct (the loading bar started earlier shouldn't be left running) but worth noting as a small wart: the asymmetry between handleLivePatch (no pageLoading start) and handleLiveRedirect (does start one) bled into the fix. If a third handler is ever added with a similar "start something, then maybe pushState" shape, that handler also needs to clean up its something on the cross-origin path. Worth a one-line comment in the code OR a candidate Stage 11 reviewer-check (filed: "for any new live_* handler that starts a side-effect before the pushState call, verify cleanup is mirrored in the cross-origin guard").

### Process verification

- Branch-verify reflex confirmed before each commit
- Post-commit `git log -1 --oneline` reflex caught the first attempt's bounce on `4d3a4b6d` (pre-commit hook); re-stage + retry landed; second commit `21dc8905` (CHANGELOG) landed first try
- Two-commit shape (#181) Gate 1 + Gate 2 both green programmatically
- Gate-off self-test (Action #1200/#1468) executed empirically — exactly 2-of-24 cases failed under stripped-guards version
- Stage 11 Code Review verdict: **APPROVE** (0 🔴, 0 🟡, 0 🟢 — clean)
- Stage 12 (Review Verdict): no findings, Re-Review skipped per template condition
- Stage 13 Merge: `gh pr merge 1600 --repo djust-org/djust --squash --delete-branch --admin`
- Issue #1599 auto-closed via "closes #1599"

### Followups

- **Candidate Stage 11 reviewer-check** (filed in retro): "For any new `live_*` handler that starts a side-effect before pushState (loading bar, prefetch clear, scroll position capture, etc.), verify cleanup is mirrored in the cross-origin guard's full-page-nav path." Generalizes the handleLiveRedirect-loading-bar-stop addition in this PR.
- **rc11 cut needed**: this is a hard-crash production bug that would affect ALL djust consumers using cross-origin `dj-navigate`. djust.org itself was crashing on it. rc10 was cut ~30 minutes ago and DOES NOT have this fix. Worth cutting rc11 immediately so downstream consumers can upgrade.

### Related

- Closes #1599 (the production crash)
- 4-PR same-session arc: #1596 (theming fix) → #1598 (meta-invariant test) → #1600 (this PR)
- Source: `python/djust/static/djust/src/18-navigation.js` `handleLivePatch` + `handleLiveRedirect`

`RETRO_COMPLETE`

PR #1606, opened 2026-05-27
fix(checks): honor suppress_checks for V001/V005 + add abstract opt-out (closes #1604, #1605)

### LINKED ISSUE #1604: suppress_checks in DJUST_CONFIG silences C003 but not V001/V005
## Bug

`DJUST_CONFIG = {"suppress_checks": ["V001", "V005"]}` does not silence the V001 and V005 checks, even though the hints for those checks point to this exact mechanism. The same `suppress_checks` entry works correctly for C003.

## To reproduce

```python
# settings.py
DJUST_CONFIG = {
    "suppress_checks": ["V001", "V005"],
}
```

```python
# apps/shared/base_views.py
from djust import LiveView

class BaseLiveView(LiveView):
    """Abstract base — no template_name, never mounted directly."""
    login_required = False
```

```
python manage.py check
```

**Expected:** V001 and V005 silenced (8 silenced, 0 issues).  
**Actual:** V001 and V005 still appear as warnings (2 issues, 5 silenced).

## Workaround

Use Django's own mechanism instead:

```python
SILENCED_SYSTEM_CHECKS = ["djust.V001", "djust.V005"]
```

This works correctly. The `suppress_checks` shorthand in `DJUST_CONFIG` should produce equivalent behaviour.

## Context

- Discovered while building `djust-org/djust-start` (the official starter template)
- C003 suppresses correctly via `suppress_checks`, so the mechanism is partially wired — V001/V005 appear to not be hooked into it
- Checked against djust 1.0.0rc7

### LINKED ISSUE #1605: V001/V005 fire on abstract base LiveView classes that are never mounted
## Problem

V001 (`missing 'template_name'`) and V005 (`not in LIVEVIEW_ALLOWED_MODULES`) fire against abstract base view classes that are intentionally never mounted directly. This produces false positives that users must suppress via `SILENCED_SYSTEM_CHECKS` (or the broken `suppress_checks` workaround — see #1604).

## Example

```python
# apps/shared/base_views.py
from djust import LiveView

class BaseLiveView(LiveView):
    """Abstract base — no template_name, never mounted directly."""
    login_required = False
    # Subclasses provide template_name, mount themselves
```

```
?: (djust.V001) apps.shared.base_views.BaseLiveView: missing 'template_name' attribute.
?: (djust.V005) apps.shared.base_views.BaseLiveView is not in LIVEVIEW_ALLOWED_MODULES.
```

Both warnings are correct in isolation but wrong in context — `BaseLiveView` is abstract boilerplate, not a view to be mounted.

## Suggested fix

Either:

1. **Opt-out attribute** — honour an `abstract = True` class attribute on LiveView subclasses and skip system checks for them, matching Django's `Meta.abstract` pattern for models.

2. **Heuristic skip** — if a class has no `template_name` AND is not in `LIVEVIEW_ALLOWED_MODULES` AND has no `dj-view` binding registered, infer it's abstract and skip V001/V005.

Option 1 is explicit and predictable:

```python
class BaseLiveView(LiveView):
    abstract = True  # suppresses V001, V005 without touching settings
```

## Impact

Every djust project that defines a base view class hits this. The `djust-org/djust-start` starter template ships with this pattern (and currently works around it with `SILENCED_SYSTEM_CHECKS`).

johnrtipton:
## Summary

Bundled fix for two tightly-related issues surfaced while building `djust-org/djust-start`:

- **Closes #1604** — `DJUST_CONFIG = {"suppress_checks": ["V001", "V005"]}` was silently a no-op even though it worked for C003. Root cause: V001 and V005 were the only V/C/T/Y checks in the file that did NOT call the existing `_is_check_suppressed()` helper.
- **Closes #1605** — V001/V005 fired on user-defined abstract base classes (`BaseLiveView(LiveView)` with no `template_name`, never mounted directly), forcing every project to suppress them globally. Adds an explicit `abstract = True` class-attribute marker.

| Issue | Modified files | Covering tests |
|-------|---------------|---------------|
| #1604 | `python/djust/checks.py` (V001 line ~1224, V005 line ~1382 — added `_is_check_suppressed` guards + hint updates) | `python/tests/test_checks_1604_1605.py::TestV001SuppressViaDjustConfig` (2 cases), `::TestV005SuppressViaDjustConfig` (2 cases) |
| #1605 | `python/djust/live_view.py` (new `abstract: bool = False` class attr), `python/djust/checks.py` (new `cls.__dict__.get("abstract")` skip at top of per-class loop) | `::TestAbstractOptOut` (4 cases: V001-skipped, V005-skipped, not-inherited, explicit-False-is-concrete), `::TestLiveViewAbstractAttrDeclared` (1 case) |
| Both | `docs/system-checks.md`, `docs/guides/error-codes.md`, `CHANGELOG.md` | — |

## Design

**`abstract` is class-explicit, not inherited.** Consulted via `cls.__dict__.get("abstract")` so subclasses of an abstract base are validated as concrete unless they also redeclare `abstract = True`. Mirrors Django's `Meta.abstract` semantics for models. Test `test_abstract_not_inherited` locks this in.

**Two-commit shape (#1173)**: commit 1 is impl + tests (`a34fb348`), commit 2 is docs + CHANGELOG (`23965b2c`). Both post-commit gates pass.

**Gate-the-fix-off self-test (Action #1200 / #1468)**: confirmed non-tautological by reverting the V001 guard, observing `test_v001_suppressed_via_djust_config` fail, and restoring.

## Test plan

- [x] `python/tests/test_checks_1604_1605.py` — 9 new cases, all pass
- [x] Full Python test suite — 7465 passed, 21 skipped, 0 failed
- [x] Existing V001/V005/C003 tests — 22 passed, no regression
- [x] Gate-off self-test on V001 guard — confirms not tautological
- [x] Branch-verify reflex + State-File Gate — passed
- [x] Two-commit-shape gates 1 & 2 — passed

## Out of scope (filed if needed at merge)

- V002/V003/V004/V007/Q007 inside the same per-class loop also lack the explicit `_is_check_suppressed` guard. The `abstract = True` opt-out covers them all (since it `continue`s before any check fires), but the global `suppress_checks` shortcut for those check IDs is a separate follow-up per Action #1079 scope-discipline.
- Heuristic-based abstract detection (#1605 option 2) — too implicit, deferred.


johnrtipton (commented):
## Stage 11 review — APPROVE (posted as comment; own-PR)

**Stale-base**: BEHIND=0 vs origin/main.

**Plan compliance** (4/4):
- V001 guard at `python/djust/checks.py:1232` (`not _is_check_suppressed("djust.V001")`).
- V005 guard at `python/djust/checks.py:1394`.
- Abstract skip at `python/djust/checks.py:1212` — `cls.__dict__.get("abstract") is True` (correctly NOT via `getattr`/inheritance; pinned by `test_abstract_not_inherited`).
- `abstract: bool = False` on base at `python/djust/live_view.py:298`.

**Correctness**:
- `__dict__` form is sound: base default `False` is inherited via `getattr`, but per-class loop sees `None` for non-redeclaring subclasses — semantics match Django `Meta.abstract`.
- V005 hint `% module` binding verified at runtime — single `%s` token, literal `{'suppress_checks': ['V005']}` does not conflict with `%` operator.
- `is True` (strict identity) — `abstract = "yes"` truthy-non-bool does NOT skip. Likely intentional. 🟢 worth a one-line note in the docstring at `live_view.py:291-298`.

**Security**: `continue` only skips per-class V/Q loop iteration. `check_configuration`, security checks, and other top-level `check_*` functions unaffected.

**Tests**: 9 cases in `python/tests/test_checks_1604_1605.py` (plan promised 6; superset matches CHANGELOG claim). Inline run: 202 passed across `test_checks_1604_1605.py` + `test_checks.py`.

**Edge cases considered**: truthy-non-bool `abstract`, abstract + suppress simultaneously (orthogonal — abstract short-circuits first), V002-V007/Q007 missing-guard scoped out per Action #1079.

🟢 V001/V005 hint text now lists all three escape hatches.

`APPROVE` — merge when ready; file V002/V003/V004/V007/Q007 missing-guard follow-up per plan.

johnrtipton:
# Retrospective — PR #1606

Bundled fix for #1604 (suppress_checks silences V001/V005) + #1605
(`abstract = True` opt-out for abstract base LiveView classes).

## Quality: 5/5

Both bugs were tightly scoped, well-located by symptom (the reporter's
own evidence), and the fix shape was mechanical once verified. No
surprise findings at Stage 11 — Code Review approved with a single 🟢
nit (worth documenting that `is True` strict identity means `abstract
= "yes"` won't skip; intentional). Full CI green across all check
suites.

## What went well

- **Bundling decision** — the two issues cross-referenced each other,
  touched the same file, and surfaced from the same `djust-org/djust-start`
  starter template build-out. A single PR shipping both is the right
  shape: one branch, one CI run, one review, one CHANGELOG entry pair
  (Added for #1605, Fixed for #1604), one mental load for future
  readers.
- **Symptom-up triage upheld**. Reporter for #1604 cited the exact
  mechanism (`_is_check_suppressed` not called for V001/V005) and a
  10-second grep confirmed: every other V/C/T/Y check called the
  helper; V001 and V005 did not. The hypothesis was correct, but the
  pipeline still wrote a failing reproducer test first (Action #1210)
  — discipline, not theatre.
- **`__dict__.get("abstract")` non-inheritance**. The most subtle
  decision in the diff: opt-in must not propagate to subclasses, or
  every concrete view inheriting from an abstract base would silently
  bypass V001/V005. The `test_abstract_not_inherited` test pins this
  behavior; without it a future reader could "fix" the code by
  switching to `getattr` and reintroduce the bug. Mirrors Django's own
  `Meta.abstract` semantics, so it's self-documenting via familiarity.
- **Gate-the-fix-off self-test (Action #1200 / #1468) applied at
  Stage 7**, not Stage 11. Reverting the V001 guard surfaced exactly
  the suppress test failing, restoring made it pass — confirms tests
  are non-tautological at implementer time rather than reviewer time.
  Compresses Stage 11 findings into Stage 7, as designed.
- **Two-commit shape clean**. Impl commit (`a34fb348`) touched zero
  CHANGELOG/docs; docs commit (`23965b2c`) touched zero impl. Both
  programmatic gates (Gate 1 + Gate 2 from `#1177`) verified.

## Lessons learned

- **Hint text matters**. The original V001 hint said "Set
  template_name on your LiveView class." — actionable for the common
  case but invisible to the abstract-base case. Updated hints now
  mention all three escape hatches. Cost of update: 4 lines per check.
  Benefit: every user who encounters the warning sees the right
  resolution without grepping the docs.
- **Audit at plan time, fix only what's cited** (Action #1079).
  Plan-time audit found that V002, V003, V004, V007, and Q007 inside
  the same per-class loop ALSO lack the `_is_check_suppressed` guard.
  Cited issues were V001 + V005; PR scope stayed tight. The
  `abstract = True` opt-out covers them transitively (since it
  `continue`s before any check fires), but the global `suppress_checks`
  shortcut for those specific IDs is a separate followup. Recording
  the audit findings in the plan file keeps them visible without
  scope-creeping the PR.
- **One implementer = no contamination** (Action #180). This was a
  single-agent, single-checkout pipeline; no CHANGELOG cross-talk and
  no duplicate `### Fixed` headings. Confirms the canonical safety
  pattern.

## What could be improved

- **Followup issue for V002/V003/V004/V007/Q007 missing-guard pattern**
  was not filed in the same PR. Should have been done at Stage 9 docs
  alongside the CHANGELOG so it doesn't get lost. Filing it
  immediately after merge.

## IDEA: lines for improvements

- IDEA: extract a `_emit_check(error, suppression_id)` helper in
  `python/djust/checks.py` that wraps `errors.append(...)` with the
  `_is_check_suppressed` guard. Mechanical replacement at all
  ~30 emission sites would prevent future drift (a new check author
  forgetting the guard) and shrink the per-check footprint by 2-3
  lines. Filed as a v1.1+ refactor candidate.

- IDEA: at next milestone retro, consider whether to **canonicalize
  the "audit-but-stay-scoped" pattern** explicitly as an Action rule.
  v0.9.5-2 retro (Action #1391 / #1400) already established
  symbol-migration grep canon for renames + removals; the
  audit-at-plan-time analogue for "I noticed N other call sites with
  the same bug class, but the cited issue says only fix M" feels
  load-bearing enough to formalize.

`RETRO_COMPLETE`

Issue #1610, opened 2026-05-27
WS mount HTML not diffed against prerender DOM when has_prerendered=true


johnrtipton:
## Problem

When the client connects with `has_prerendered=true` in the mount message, the server sends back a full WS-mount HTML payload — but the client appears to **keep the prerender DOM** as-is and only apply subsequent diff patches. If the WS-mount HTML differs from the prerender (e.g. because some state is only resolvable in WS context, like presence count), the differences are silently dropped and the DOM is stuck at whatever the HTTP prerender produced.

## Repro

```python
class HomeView(BaseLiveView):
    def mount(self, request, **kwargs):
        super().mount(request, **kwargs)
        if hasattr(self, "_websocket_session_id"):
            self.online_count = 2   # WS-mount value
        else:
            self.online_count = 1   # HTTP prerender value
```

```html
<span>{{ online_count }} online</span>
```

Mount payload received by client (captured with djust-browser ws_frames):

```json
{
  "type": "mount",
  "session_id": "...",
  "view": "apps.home.views.HomeView",
  "version": 13,
  "html": "...<span>2 online</span>..."
}
```

But the DOM displays `"1 online"` (the prerender value) and never updates. Subsequent broadcast patches arrive with `patches: []` because nothing has *changed* from the WS-mount baseline (server already at `online_count=2`).

## Why it bites

Any value that's HTTP-vs-WS-context-dependent has this footgun. Presence is the obvious one (tab isn't tracked yet at HTTP time). It's not at all obvious that the WS-mount HTML is being discarded — server-side logs and `_debug.variables` show the right value, but the chip stays at the HTTP value forever.

## Workaround

Render a placeholder in HTTP that's known to differ from any WS value (so a future broadcast eventually changes state and produces patches). Brittle and adds a visible flash.

## Suggested fix

When `has_prerendered=true`, the client should **diff** the prerender DOM against the WS-mount HTML it just received and apply patches. Right now it looks like the HTML in the mount payload is essentially ignored if `has_prerendered=true` — or at least no patches are emitted to reconcile.

## Context

- Discovered building `djust-org/djust-start`
- djust 1.0.0rc7
- This was the single longest debugging detour while wiring up the demo's "N online" chip — the server was correct end-to-end, the client just wouldn't show it. Captured the smoking-gun WS frame using `djust-browser` MCP.

Issue #1227, opened 2026-04-30
tech-debt: pre-commit/CI lint for bare comma-list `Closes #X, #Y` (auto-close failure)


johnrtipton:
## Failure mode

GitHub's auto-close-keyword parser only matches \`Closes #N\` / \`Fixes #N\` / etc. when each ref is preceded by its own keyword. Comma-list forms like:

\`\`\`
Closes #1037, #1039, #1040, #1046.
Closes #1037, #1039, #1040, #1046, #1077.
Closes #1195, #1196, #1197, #1198, #1199, #1200.
\`\`\`

silently parse as **only \#1037 / \#1195** (the first ref). The rest stay open.

## Observed bites

This bit twice during the v0.9.1-6 → v0.9.1-7 drain in 24 hours:

- PR #1225 commit body had \`Closes #1195, #1196, #1197, #1198, #1199, #1200.\` — only #1195 closed; manual closure of 5 stragglers.
- PR #1226 commit body had \`Closes #1037, #1039, #1040, #1046, #1047, #1049, #1050, #1051, #1055, #1056, #1060, #1071, #1077, #1078, #1079.\` — only #1037 closed; manual closure of 14 stragglers.

Total cost in the v0.9.1 release-prep work: ~5 minutes of manual \`gh issue close\` per occurrence, twice. Plus retro-comment writing per straggler.

## Existing canon

Action Tracker #184 / GitHub #1185 (closed) covered the **parenthesized** comma-list form (\`(closes #X, closes #Y)\` in PR titles) — fixed in PR #1192. But the **bare** comma-list-on-its-own-line form was not addressed; PR-checklist Closing-Keywords section names the parenthesized variant only.

## Proposed fix

Two-pronged:

1. **Pre-commit hook** that scans staged commit messages for the pattern \`Closes #\d+(\s*,\s*#\d+)+\` (or \`Fixes\`, \`Resolves\`) and rejects with a message asking to put each \`Closes #N\` on its own line.

2. **Update \`docs/PULL_REQUEST_CHECKLIST.md\` Closing-Keywords section** to explicitly name the bare comma-list form alongside the parenthesized one. The lesson from #1184 was incomplete; the bare form is the more common shape (since it survives the title→body migration #1192 recommended).

## Estimated effort

~1 hour (pre-commit hook + regex tuning + checklist update + 1-2 unit tests).

🤖 Auto-filed during v0.9.1-7 cleanup batch after the failure mode bit twice in 24 hours.

PR #1615, opened 2026-05-27
feat(presence): zero-config online_count + abstract opt-out + auto-broadcast

### LINKED ISSUE #1611: PresenceMixin: auto-expose online_count as instance attribute
## Problem

The natural template usage for a presence-aware view is:

```html
<span class="presence-chip">{{ online_count }} online</span>
```

But there's no `online_count` on the view by default — users have to compute and assign it themselves. AND it must be an **instance attribute** (not just a context dict key), because djust's render diff tracks instance attributes for its dirty-checking. If you only set `ctx["online_count"]` in `get_context_data`, the chip text changes in the rendered HTML but the diff emits no patches and the DOM stays stale (took me hours to figure out — full WS frame trace showed the variable wasn't in `_debug.variables`).

## Repro

```python
# Doesn't work — chip never updates after mount:
    ctx["online_count"] = len(self.list_presences()) or 1
    return ctx
```

vs

```python
# Works:
def _refresh(self):
    self.online_count = len(self.list_presences()) or 1
```

The first form renders correctly server-side but the broadcast patch is empty because the instance state set didn't change.

## Suggested fix

Have `PresenceMixin` set `self.online_count` automatically inside `track_presence()` and `untrack_presence()`, and in the built-in `_on_presence_change` handler (see companion issue). Then `{{ online_count }}` Just Works in templates with zero user code.

Combined with the WS-only gating and auto-broadcast suggestions, the entire presence pattern collapses to:

```python
class MyView(PresenceMixin, LiveView):
    presence_key = "demo"
```

```html
{{ online_count }} online
```

## Context

- djust 1.0.0rc7

### LINKED ISSUE #1612: PresenceMixin: track_presence() should be a no-op during HTTP mount
## Problem

Every djust LiveView's `mount(request, **kwargs)` runs **twice** per page load: once for HTTP prerendering and again when the WebSocket connects. Each run creates a separate view instance with a separate `id(self)`. Code that calls `track_presence()` in `mount()` registers the connection **twice** — once with the HTTP-mount instance's user_id, once with the WS-mount instance's user_id.

The HTTP-mount entry is an orphan: it never gets a corresponding `untrack_presence()` because the HTTP request returns and the view instance is garbage collected. The entry stays in the in-memory backend for the full `PRESENCE_TIMEOUT` (60s) before stale cleanup removes it.

Result: every page load leaves a phantom presence entry for ~60 seconds. With dev hot-reload + a few reloads, the count gets badly inflated.

## Repro

```python
# apps/shared/base_views.py
from djust import LiveView

class BaseLiveView(PresenceMixin, LiveView):
        self.track_presence()
        self.online_count = len(self.list_presences())
```

Reload the page 5 times. `online_count` climbs to 5+ in a single browser tab.

## Workaround

```python
    if hasattr(self, "_websocket_session_id"):
        self.track_presence()
        self.online_count = len(self.list_presences())
```

The `_websocket_session_id` attribute is set by `LiveViewConsumer` before `mount()` runs in WS context, so its presence is a reliable discriminator. But this is undocumented and every user implementing presence will hit the orphan-entries footgun first.

## Suggested fix

Make `track_presence()` itself check WS context and no-op in HTTP mount:

```python
def track_presence(self, meta=None):
    if not hasattr(self, "_websocket_session_id"):
        return  # HTTP mount creates a throwaway view; presence only makes sense in WS
    ...
```

Optionally log at DEBUG level so users know it's being skipped.

## Context

- Discovered building `djust-org/djust-start` (the official starter)
- djust 1.0.0rc7
- See also: companion issues filed on PresenceMixin DX (broadcast, online_count exposure, anonymous user_id collision)

### LINKED ISSUE #1613: PresenceMixin: default get_presence_user_id() collapses anonymous tabs
## Problem

The default `get_presence_user_id()` returns `f"anon_{request.session.session_key}"` for anonymous users. Two browser tabs share the same Django session → same session_key → same user_id → they overwrite each other in the presence dict (`_groups[presence_key][user_id] = record` is keyed by user_id).

Result: opening two tabs in the same browser shows "1 online" forever, even though both tabs are connected.

## Repro

Any anonymous presence demo. Open in two tabs of the same browser. Count stays at 1.

```python
# In InMemoryPresenceBackend.join:
self._groups.setdefault(presence_key, {})[user_id] = record  # ← overwrites
```

## Workaround

```python
def get_presence_user_id(self):
    request = getattr(self, "request", None)
    if request and request.user.is_authenticated:
        return str(request.user.id)
    return f"anon_{id(self)}"  # unique per view instance / per WS connection
```

## Suggested fix

Either:

1. **Default behavior change**: include something per-connection-unique for anonymous users. `id(self)` works since each WebSocket connection gets its own view instance. Authenticated users still collapse to one identity across tabs (which is the right semantics for them).

2. **Opt-in flag**: `class MyView(PresenceMixin, LiveView): presence_unique_per_connection = True` — keep current default for backwards compat but make the connection-unique mode discoverable.

Option 1 matches user expectation: "I see 2 online when I open 2 tabs" is what every demo of this kind shows. Option 2 is safer for existing apps.

## Context

- djust 1.0.0rc7

### LINKED ISSUE #1614: PresenceMixin: track/untrack should auto-broadcast for cross-tab updates
## Problem

`PresenceMixin.track_presence()` and `untrack_presence()` update the backend, but other connected sessions of the same view don't find out about the change unless the user manually wires `push_to_view` to a handler that triggers re-render. This is required for ANY use of \"N online\" / live-avatar features — i.e. the canonical use case for PresenceMixin.

## Repro

```python
class MyView(PresenceMixin, LiveView):
    presence_key = "demo"

        self.track_presence()
        self.online_count = len(self.list_presences())
```

Open in two tabs. Tab 1 shows "1 online", tab 2 shows "2 online" (correct at mount). Close tab 2. Tab 1 still shows "2 online" forever — `untrack_presence()` fired on tab 2's disconnect, but tab 1 never learned about it.

## Workaround

Every user implementing presence has to write the same boilerplate:

```python
    if hasattr(self, "_websocket_session_id"):
        self.track_presence()
        push_to_view(
            f"{self.__class__.__module__}.{self.__class__.__name__}",
            handler="_on_presence_change", payload={},
        )

def untrack_presence(self):
    super().untrack_presence()
    push_to_view(
        f"{self.__class__.__module__}.{self.__class__.__name__}",
        handler="_on_presence_change", payload={},
    )

@event_handler
def _on_presence_change(self, **kwargs):
    # empty body — re-render triggers list_presences refresh
    pass
```

## Suggested fix

Have `track_presence()` and `untrack_presence()` fire the broadcast themselves, calling a well-known handler name (`_on_presence_change` or similar) that the mixin provides as an empty `@event_handler`. Users can override it to customize the refresh.

Companion issue to the "no-op during HTTP mount" one — together they make the typical use case zero-config.

## Context

- djust 1.0.0rc7

johnrtipton:
## Summary

Bundled DX upgrade for `PresenceMixin` — closes 4 issues all surfaced building `djust-org/djust-start` on djust 1.0.0rc7. Together they collapse the canonical presence pattern to **zero-config**:

```python
class MyView(PresenceMixin, LiveView):
    presence_key = "demo"
    # optional: presence_unique_per_connection = True  # anonymous-tab demos
```
```html
{{ online_count }} online
```

Closes #1611.
Closes #1612.
Closes #1613.
Closes #1614.

| Issue | Fix | Tests |
|-------|-----|-------|
| **#1611** | Auto-set `self.online_count` as instance attr in `track_presence`/`untrack_presence`/`_restore_presence`/`_on_presence_change` so diff dirty-tracking emits patches without `get_context_data` scaffolding | `TestIssue1611OnlineCountAttribute` (2 cases) |
| **#1612** | `track_presence()` no-ops when `_websocket_session_id` is absent — kills orphan presence records that lingered ~60s on the throwaway HTTP-mount instance | `TestIssue1612HttpMountNoOp` (2 cases) |
| **#1613** | New `presence_unique_per_connection: bool = False` opt-in flag. When True + anonymous → `anon_conn_<ws_session_id>`. Authenticated users always use `user.id` regardless. | `TestIssue1613PerConnectionUniqueness` (3 cases) |
| **#1614** | `track`/`untrack`/`_restore_presence` fire `push_to_view` to default `@event_handler` `_on_presence_change` that refreshes `online_count`. Broadcast-loop guard: default handler body is exclusively `_refresh_online_count()` — never `track_presence` | `TestIssue1614AutoBroadcast` (4 cases) + `TestZeroConfigPresence` (1 cross-cutting) |

## Two-commit shape

- impl `7f425478` — `python/djust/presence.py` (+115 LoC) + 4 test files
- docs `d7dccd33` — `CHANGELOG.md` + `docs/website/guides/presence.md`

## Test plan

- [x] 12 new reproducer tests pass
- [x] Existing `test_presence.py` / `test_realtime_multiuser.py` updated to opt back into tracking (FakeView default `_websocket_session_id = "test_ws"`)
- [x] `test_mixin_restoration_893_894.py` patches `djust.presence.push_to_view` (new broadcast in `_restore_presence`)
- [x] Full suite: **7487 passed, 21 skipped, 0 failed**
- [x] Stage 7 Self-Review: REVIEW_PASSED
- [x] Stage 8 Security Check: SECURITY_PASSED (`_websocket_session_id` is server-generated `uuid.uuid4()`, not client-controllable)
- [x] Broadcast-loop infinite-recursion check: confirmed `_on_presence_change` body is exclusively `_refresh_online_count()`
- [x] Gate-off self-test (Action #1200/#254): reverting any of the 4 fixes makes a specific reproducer fail; restoring makes it pass

## Out of scope

- #1610 (WS-mount HTML diffing) — different surface, separate pipeline
- PresenceManager backend changes — mixin-only PR


johnrtipton (commented):
## Stage 11 review — APPROVE

**Stale-base**: BEHIND=0. ✅

**Tests**: 82/82 pass (`test_presence_mixin_dx_1611_1614` 12 + `test_presence` 21 + `test_realtime_multiuser` 38 + `test_mixin_restoration_893_894` 11), 0.20s.

### Plan compliance — all 4 issues delivered
- #1611 `online_count`: `_refresh_online_count` at `python/djust/presence.py:161-175`, called from track/untrack/restore. ✅
- #1612 HTTP-mount guard: `presence.py:275-280` (correctly AFTER the `_presence_tracked` early-return at L266). ✅
- #1613 `presence_unique_per_connection`: class attr L153; auth-always-collapse at L241-242; flag+ws_sid → `anon_conn_<sid>` L244-247; flag+no-sid defensive fallback `anon_<id(self)>` L250; default `anon_<session_key>` L253-254. ✅
- #1614 broadcast + `_on_presence_change`: `_broadcast_presence_change` L177-190 + `@event_handler`-decorated default handler L387-402 (body exclusively `self._refresh_online_count()` — loop-terminating per plan §7). ✅

### Focus-area checks
- 🟢 Broadcast-loop invariant: handler body L402 is the lone `_refresh_online_count()` call — no `track_presence`/`_broadcast_presence_change`. Loop terminates after one hop.
- 🟢 `_on_presence_change` reachable: `@event_handler` at L387 satisfies the `websocket.py:4954` underscore-name gate.
- 🟢 `_restore_presence` extension safety: new refresh/broadcast calls (L350-351) sit INSIDE the existing `try` (L344) / `except` (L352) so backend faults during restore don't kill the WS.
- 🟢 `untrack_presence` order (plan §11): `leave_presence` L369 → `_presence_user_id = None` L379 → `_refresh_online_count` L384 → broadcast L385. Leaving user excluded from the post-leave count. ✅
- 🟢 Downstream fixtures: FakeView `_websocket_session_id="test_ws"` defaults in `test_presence.py:80-83` + `test_realtime_multiuser.py:63-66`; `@patch("djust.presence.push_to_view")` decorator at `test_mixin_restoration_893_894.py:43`.

Edge cases verified: anon+flag+no ws_sid defensive fallback; auth+flag (collapse preserved); idempotent re-track; track_presence raising mid-call. CHANGELOG entries present for all 4 issues under `[Unreleased]`.

`APPROVE`

johnrtipton:
# Retrospective — PR #1615

Bundled `PresenceMixin` DX upgrade closing #1611/#1612/#1613/#1614. Reporter
filed all 4 simultaneously while building `djust-org/djust-start`. End state
collapses the canonical presence pattern to `class MyView(PresenceMixin,
LiveView): presence_key="demo"` + `{{ online_count }} online` — zero
boilerplate.

## Quality: 5/5

Clean execution: 14/14 stages, two-commit shape clean (impl `7f425478` +
docs `d7dccd33`), 12 new reproducer tests with gate-off verification,
Stage 7 + Stage 8 both PASSED with one 🟡 each (both non-blocking and
documented), Stage 11 reviewer APPROVE. Bigger surface than the prior
single-issue PRs but the bundling decision held — the four issues are
genuinely one DX story.

## What went well

- **Bundling discipline held.** Four issues, one PR. The reporter's
  framing was explicit ("companion issues"), the surface was
  identical (`python/djust/presence.py`), and the four fixes interact
  (e.g., #1612's HTTP-mount guard is what makes #1614's broadcast
  safe — only WS-mount instances broadcast). Splitting these into 4
  PRs would have introduced ordering dependencies and rebase pain.

- **Locked design decisions BEFORE planning subagent**. The user
  resolved the #1613 ambiguity (opt-in flag vs default change) via
  `AskUserQuestion` at the prep stage. The Planning subagent then had
  a clear contract to work against rather than having to make API
  decisions mid-investigation. ~30 mins saved vs the alternative.

- **Plan subagent in read-only mode + delivered inline content**.
  The Planning role's tool harness was read-only; the subagent
  produced the full plan content + reproducer test source in its
  return message instead of writing files. Parent agent then
  materialized both files. Worked, but see "could be improved" below.

- **Broadcast-loop invariant locked in test code, not just docs**.
  `test_default_on_presence_change_handler_refreshes_count_only`
  asserts both `mock_track.assert_not_called()` AND `v.online_count
  == 1`. Future refactors can't accidentally turn the default handler
  body into something that re-invokes `track_presence`.

- **`_websocket_session_id` provenance verification was load-bearing**.
  Stage 8 (Security) traced `_websocket_session_id` to its source
  (`websocket.py:1398` — server-generated `uuid.uuid4()`). Without
  that trace, the `presence_unique_per_connection` flag could have
  been a session-fixation aid. The trace gave a clean
  SECURITY_PASSED.

## Lessons learned

- **Pre-push hook #1227 enforces one-Closes-per-line.** Both initial
  commits used `Closes #1611, #1612, #1613, #1614.` and the hook
  rejected the push, requiring a full `git reset --soft HEAD~2` +
  re-commit cycle. Cost ~5 min wall-clock. The Stage 5 implementer
  subagent prompt and the docs commit shape should both bake in the
  one-Closes-per-line format from the start for any multi-issue PR.
  Filing as IDEA: line below.

- **Plan-stage subagent in read-only mode** is a known harness
  quirk (also hit in PR #1606). The plan subagent did good work but
  shipped its plan + tests as inline message content rather than
  written files; parent had to materialize both. Adds a fragility
  point (the inline content has to be transcribed verbatim, and if
  the parent context compresses or the message gets truncated, the
  plan is lost). For multi-file plans (here: plan.md + reproducer
  .py), the cost is higher than for plan.md alone.

## What could be improved

- **The Implementation subagent re-discovered the `_websocket_session_id
  = "test_ws"` fixture pattern across both `test_presence.py` and
  `test_realtime_multiuser.py`** rather than the plan listing the
  exact one-line FakeView edit for each. The Stage 5 implementer
  spent investigation cycles on what the plan should have prescribed.
  Future plan files for "downstream test fixture update"-class
  changes should literally include the one-line patch for each
  FakeView constructor. Filing as IDEA: line.

## IDEA: lines

- IDEA: **bake one-Closes-per-line into the multi-issue PR template
  in pipeline-run skill prompts**. The hook #1227 has been canon for
  weeks; the implementer subagent prompt should include the format
  example so it ships correctly the first time.

- IDEA: **plan-stage "downstream fixture patch" pre-write**. For any
  fix that changes the contract of a function consumed by N test
  fixtures, the Stage 4 plan should include the literal patch for
  each fixture (not just "update the FakeView to set X"). Saves
  Stage 5 investigation cycles and avoids fixture-update drift.

- IDEA: **dedicated test for `_restore_presence` extension**. Stage 7
  flagged this as 🟡 (covered indirectly by
  `test_mixin_restoration_893_894`'s patched `push_to_view` but no
  direct assertion on the `_refresh_online_count` + broadcast pair).
  Worth a small follow-up PR or absorption into the next
  presence-touching change.

`RETRO_COMPLETE`

PR #1618, opened 2026-05-27
fix(vdom): route SetText patches by byte position, not content equality

### LINKED ISSUE #1617: Render diff misroutes online_count patches when chip and other text nodes share a numeric prefix
## Bug

When `online_count` (a `PresenceMixin` instance attribute, added by PR #1615) changes from 1 to 2, the render diff emits a `SetText` patch targeting **the wrong DOM path** — specifically, a different text node elsewhere in the tree that happens to contain the bare digit "1". The chip text never updates and the DOM node that DOES get patched is left displaying a value that doesn't match its underlying state.

## Repro

Using `djust-org/djust-start` with djust 1.0.0rc12 (post-PR #1615, post-PR #1616):

1. Start `make dev` with a clean in-memory backend
2. Open one tab → chip shows "1 online", 🔥/❤️/🚀/etc. reaction counts also include the digit "1" (e.g. 🚀 count = 1)
3. Open a second tab → 2nd tab's `track_presence` broadcasts `_on_presence_change`
4. First tab receives broadcast, `online_count` goes 1→2, diff fires

**Expected:** `SetText path=[0,0,0] text="2 online"` (chip text update)

**Actual (full patch frame from WS):**

```json
{
  "type": "SetText",
  "path": [1, 1, 1, 2, 1, 0],
  "text": "2"
}
```

`_debug.variables` on the same frame:

```
changed_fields: {'online_count'}        ← correctly tracked
online_count:    2                       ← new value
reactions:       [..., {emoji:'🚀', count:1}, ...]   ← 🚀 still says 1
```

The path `[1,1,1,2,1,0]` is the 🚀 reaction's `.count` span's text node. The DOM ends up showing 🚀=2 (because of this stray patch) while the state still says 🚀=1, and the chip stays at "1 online" forever.

## Why it triggers only on the "first" window

The bug is triggered by ambiguous text-node matching. The first window's mount HTML has:
- chip text: `"1 online"` (starts with "1")
- 🚀 count text: `"1"` (bare "1")

When `online_count` goes 1→2, the rendered HTML's chip becomes `"2 online"` while 🚀 count stays `"1"`. The diff's text-node matcher appears to incorrectly pair the new `"2"` (likely a tokenization artifact of `"2 online"`) with the OLD `"1"` of the unchanged 🚀 count node, and emits a single `SetText` at the 🚀 path.

Later-joining windows mount with chip = "2 online" or higher, so the chip's leading digit never collides with the bare "1" of any reaction count, and the same broadcast path produces correct `[0,0,0]` patches.

## Verified working path

When a different broadcast happens (a reaction click in another tab, which sends `state={"reactions": [...]}`), patches correctly target the 🚀 count path with the right text. So `state=` broadcasts use a different diff machinery than `handler=` broadcasts (the path #1615's `_on_presence_change` takes).

## Suggested fix

The render diff's node-matching should not match across text nodes whose containing elements have different `dj-id`s, OR should match by structural position rather than content similarity. Path `[0,0,0]` (chip) and `[1,1,1,2,1,0]` (🚀 count) are deep in different subtrees — they should never be confused.

## Context

- djust 1.0.0rc12 (`e794c31`)
- Captured via `djust-browser` MCP's `ws_frames` tool — see the full debug variables block above
- Repro is consistent: the first window in any test session that mounts when `online_count == 1` exhibits this bug; subsequent windows do not

johnrtipton:
## Summary

Closes #1617.

`build_fragment_text_map` (`crates/djust_live/src/lib.rs:2597-2633`) previously mapped each rendered template fragment to the first VDOM text node whose content equalled the fragment. For a variable adjacent to literal template text — e.g. `{{ online_count }} online` — the variable's rendered fragment (`"1"`) doesn't equal the chip's full text-node content (`"1 online"`), so the matcher fell through to a sibling text node whose content happened to equal `"1"` (typically a bare reaction count). When the variable changed 1→2, the `SetText` patch landed on the wrong DOM path — chip stayed at `"1 online"` forever while the 🚀 reaction count visually became `"2"` (state still said `1`).

## Fix

Position-aware fragment→text-node matching. Each fragment is located by its byte position in the assembled HTML; the map entry is only claimed when the fragment IS the entire text node (full-coverage check). Ambiguous cases (partial-overlap, whitespace-only, fragments containing tags) gracefully fall through to the byte-level `text_region_fast_path` (already sound).

```rust
// AFTER (crates/djust_live/src/lib.rs:2611-2680)
if rs == frag_start && re == frag_end {
    let (path, _text, djust_id) = &text_nodes[i];
    map.insert(idx, (path.clone(), djust_id.clone()));
}
```

Bug class: any `{{ var }}<literal>`, `<literal>{{ var }}`, or `{{ a }}{{ b }}` template pattern.

## Reporter framing refuted

Reporter claimed `state=` broadcasts worked while `handler=` broadcasts misrouted. Code inspection (`python/djust/websocket.py:4940-4979`, `python/djust/mixins/rust_bridge.py:577-694`) shows **both paths converge at `_sync_state_to_rust → set_changed_keys → render_with_diff` and take the same `text_fast_path`**. The reporter's anecdotal divergence is not explained by code — likely an unrelated side effect invalidated `fragment_text_map` in their `state=` scenario. Fix at the root.

## Test plan

- [x] 3 new regression cases in `python/djust/tests/test_text_fast_path_misroute_1617.py` (bug repro, adjacent `{{a}}{{b}}`, pure-case backstop)
- [x] 6 existing #1529 content-collapse regression cases still pass
- [x] Rust suites: djust_live 37 tests + djust_vdom 22 tests + workspace tests — all green
- [x] Python suite: 7493 passed, 21 skipped, 0 failures
- [x] Wire-protocol snapshot (`crates/djust_vdom/tests/wire_protocol_snapshot.rs`) untouched — SetText struct + msgpack/JSON serialization unchanged
- [x] Gate-the-fix-off self-test (Action #1200/#1468): reverting to content-equality reproduces the misroute, restoring the position-aware body makes it pass
- [x] Stale-base check (Action #250 / #1450) passed at impl time

## Two-commit shape (#1173)

- impl `da17346d` — `crates/djust_live/src/lib.rs` (+97 LoC) + new test file (+227 LoC)
- docs `bb2013d2` — `CHANGELOG.md` (+2 LoC)

## Out of scope

- PresenceMixin (#1611–#1614 just shipped)
- WS-mount morphdom (#1610 just shipped)
- `handler=` vs `state=` divergence (not real per code trace)
- `text_region_fast_path` heuristic (already sound)
- `fragment_text_map` cross-full-parse staleness (pre-existing concern; separate issue if surfaces)


johnrtipton (commented):
**APPROVE** — implementation matches plan §7 AFTER byte-for-byte.

### Plan compliance
- `build_fragment_text_map` rewritten at `crates/djust_live/src/lib.rs:2603-2683` with new `full_html: &str` signature.
- Call site `crates/djust_live/src/lib.rs:585-593` passes `self.last_html` via tuple-destructured `if let`.
- Uses `find_dj_root_content_range` + `scan_html_text_runs` (same primitives as `build_text_node_index`).
- Full-coverage gate `rs == frag_start && re == frag_end` present (line 2675); partial-overlap correctly falls through to `text_region_fast_path`.
- Defensive `runs.len() != text_nodes.len() → return map` present (line 2657).
- `claimed` vec removed — position-aware lookup is unique by construction.

### Focus areas
1. **Wire protocol**: `git diff main..HEAD -- crates/djust_vdom/` is empty. SetText struct + snapshot suite untouched. Actions #1448/#1538/#1541 preserved.
2. **#1529 regression**: `tests/unit/test_vdom_settext_mispath_1529.py` 6/6 pass; comment at `lib.rs:2633-2635` explicitly addresses the collapse class.
3. **Edge cases §9**: `<`/empty skip at line 2666; `runs.len()` mismatch bail at 2657; whitespace-only handled by `scan_html_text_runs` filtering.
4. **`handler=`/`state=` framing**: refuted correctly — both converge at `_sync_state_to_rust` → `set_changed_keys` (`mixins/rust_bridge.py:833`).
5. **Perf**: position helpers run once per full-parse map build, not in the hot fast-path (`lib.rs:585` guard `fragment_text_map.is_none()`).

### Tests
- `pytest test_text_fast_path_misroute_1617.py`: 3/3 PASS
- `pytest test_vdom_settext_mispath_1529.py`: 6/6 PASS
- `cargo test -p djust_vdom`: 22/22 PASS
- Full Python suite: 7493 passed, 21 skipped, 0 failed
- BEHIND=0 vs origin/main

CHANGELOG includes gate-off self-test confirmation. Ship it.

johnrtipton:
# Retrospective — PR #1618

Closes #1617. `build_fragment_text_map` rewritten from content-equality to
byte-position-aware matching. Closes the bug class for any
`{{ var }}<literal>` / `<literal>{{ var }}` / `{{ a }}{{ b }}` template
pattern where a fragment is adjacent to literal text or another variable.

## Quality: 5/5

Clean execution. Investigation-first Planning located the EXACT line
(`lib.rs:2624 — text == frag`) where the unsound predicate lived; the
fix shape was unambiguous once the root cause was identified.
Two-commit shape clean (impl `da17346d` + docs `bb2013d2`). 3 new
regression cases, 6 existing `#1529` cases preserved, full suites
green. Wire-protocol invariants explicitly verified intact.

## What went well

- **Planning subagent's symptom-up trace was exemplary**. Started at
  the visible `SetText path=[1,1,1,2,1,0] text="2"` patch frame,
  walked back through `frag_map.get(idx)` → `fragment_text_map` →
  `build_fragment_text_map` → the `text == frag` predicate at line
  2624. Identified the EXACT failure point in 1 pass instead of
  speculating across multiple suspect files.

- **Reporter's framing was partially mistaken — the plan said so explicitly**.
  Reporter claimed `state=` broadcasts worked while `handler=`
  broadcasts misrouted. Code inspection traced both paths to the
  SAME `_sync_state_to_rust` → `set_changed_keys` →
  `render_with_diff` funnel. The plan refuted the framing with file:
  line evidence and chose a root-cause fix rather than a
  broadcast-type gate (which would have masked the bug for any
  future code path hitting the same precondition). This is exactly
  the "Trust the symptom, not the cited path" rule from CLAUDE.md
  Bug-report triage #1.

- **Position-aware lookup reused existing primitives**. The plan
  identified that `build_text_node_index` (line 2558 area) already
  built byte-range→text-node-path maps via `scan_html_text_runs` +
  `find_dj_root_content_range`. The fix uses the SAME primitives —
  no new helpers introduced, no perf surprises, behavior is
  consistent with the sound `text_region_fast_path`.

- **#1529 regression interaction noted**. PR #1530 (Action #1542)
  added a `claimed[node_i]` tracker to defeat content-collapse
  between two variables with identical baselines. The new
  position-aware lookup defeats the same class by construction
  (distinct fragments have distinct byte positions, so collision is
  impossible). The `claimed` tracker is removed; the #1529 fix is
  preserved by a stronger invariant. Comment at `lib.rs:2633-2635`
  explicitly documents this.

- **Gate-the-fix-off self-test ran TWICE — once by Stage 5
  implementer, once by Stage 11 reviewer**. Both confirmed
  reverting the function body to content-equality reproduces the
  exact reporter symptom (`SetText path=[1,0] text="2"`). The test
  is provably non-tautological.

- **Wire-protocol pinning (Actions #1448/#1538/#1541) was a
  zero-cost check this time** — no serde structs touched, no
  msgpack schema change, `wire_protocol_snapshot.rs` untouched.
  Confirmed in 5 seconds with one `git diff --stat`. The rule paid
  for itself by making the verification trivial.

## Lessons learned

- **Reporter framing can be partially wrong AND still be valuable.**
  Reporter said `state=` works / `handler=` broken. Strictly false
  per code trace — both paths share the same differ. But the
  reporter's CITATION of which path triggered the misroute was
  correct; the divergence claim was the noise around a true signal.
  Skill: trust the symptom, refute the diagnostic theory, fix the
  root cause. This is the third time this session a fix landed on
  evidence the reporter cited correctly (path numbers, code line
  numbers, behavior trace) even though their causal model needed
  refining.

- **Plan-stage subagent's read-only mode hit again** — fourth
  occurrence in this session (PR #1606, #1615, #1616, now #1618).
  The plan content was delivered inline, parent agent
  materialized. Friction is consistent and worth fixing — see
  IDEAs.

- **The first-window-only manifestation was a red herring at issue
  time** but the EXACT clue that made the bug findable. "First
  window mounts at `1 online`" → the content-equality matcher only
  fails when there's a sibling node with content `"1"`. Reproducer
  template needs both the chip pattern AND a sibling bare-digit
  reaction count. Without that detail, the reproducer would have
  passed and the bug would have evaded. Future bug triage skill:
  preserve EVERY detail the reporter provides, even ones that seem
  irrelevant to the failure mode — they may be the very precondition
  that makes the bug reproducible.

## What could be improved

- **`fragment_text_map` cross-full-parse staleness (plan §9 edge
  case 10) is still a pre-existing concern**. After a full-parse
  render, the map is preserved across subsequent fast-path renders.
  If a structural mutation shifts text-node identities while
  keeping fragment count the same, the map becomes stale. We
  deferred this as out-of-scope. Worth filing a follow-up issue
  with a synthetic reproducer to track.

## IDEA: lines

- IDEA: **Planning subagent read-only mode — fourth occurrence this
  session**. Either change the Plan agent type's tool harness to
  allow scoped writes to `.pipeline-state/` and `python/tests/` /
  `crates/*/tests/`, OR build a small helper in pipeline-run that
  auto-extracts inline plan content from the subagent return
  message and materializes it. Empirical cost: ~5-10 min per
  pipeline × 4 pipelines = ~30-40 min saved this session alone.

- IDEA: **Follow-up issue for `fragment_text_map` cross-full-parse
  staleness**. Synthetic reproducer: full-parse render produces map
  with N fragment→text-node entries; subsequent structural mutation
  via Rust-side `update_template` invalidates the map (lines
  237/261/759 already do this). But a NON-template structural
  change (e.g., a `{% for %}` loop's iteration count changing,
  producing more/fewer text nodes) might keep the fragment count
  identical while shifting node identities — and the map currently
  has no invalidation path for that. Hard to reproduce in practice
  but worth a synthetic test as a backstop.

- IDEA: **Reporter framing pattern**. Across PR #1606, #1615, #1616,
  #1618 — four cases where the reporter cited code locations
  correctly but causal theories incorrectly. Consider documenting
  the "trust the symptom, refute the theory" pattern as an
  explicit Action with a sub-checklist in the Bug-report triage
  section of CLAUDE.md. The existing rule (#1) covers code-location
  citations; this would add coverage for causal-theory citations.

`RETRO_COMPLETE`

PR #1620, opened 2026-05-27
fix(components): map dj_button variant=danger to btn-destructive class

### LINKED ISSUE #1619: dj_button variant names don't match djust-theming CSS class names
## Bug

`dj_button` in `djust-components` advertises (in its docstring) the variants:

> primary, secondary, danger, ghost, link, success, warning

It renders `class="btn btn-{variant}"`, e.g. `<button class="btn btn-danger">`.

But `djust-theming`'s `components.css` only ships CSS rules for these `.btn-*` variants:

```
.btn-primary
.btn-secondary
.btn-destructive   ← not "danger"
.btn-ghost
.btn-link
```

So `dj_button(variant="danger")`, `dj_button(variant="success")`, and `dj_button(variant="warning")` all render with class names that have **no CSS rule** in djust-theming. The button appears as the base `.btn` style only — no background color, looks broken.

Symmetrically, the `.btn-destructive` class in djust-theming has no way to be emitted — no `variant` keyword produces it.

## Repro

```html
{% load djust_components %}
{% dj_button label="Delete" variant="danger" %}
```

Renders:

```html
<button class="btn btn-danger"><span class="btn-label">Delete</span></button>
```

Loaded CSS (via `{% theme_head %}` → `djust_theming/css/components.css`) has no `.btn-danger` rule → button appears unstyled.

## Suggested fix

Either:

1. **Component-side**: have `dj_button` map `variant="danger"` → class `btn-destructive` (and likewise document only the variants that actually have theme rules). Possibly also rename the keyword to `variant="destructive"` for consistency.

2. **Theme-side**: have djust-theming `components.css` add `.btn-danger`, `.btn-success`, `.btn-warning` rules so the documented variants all work. (And add an alias `.btn-destructive` for back-compat.)

Either fix resolves it; option 1 is cleaner because shadcn/Tailwind convention uses "destructive" not "danger", and `success` / `warning` variants without theme tokens (`--success`, `--warning`) are orphaned.

## Workaround

```css
/* in your project's CSS */
.btn-danger {
    background-color: hsl(var(--destructive));
    color: hsl(var(--destructive-foreground));
}
.btn-danger:hover { opacity: 0.9; }
```

## Context

- Discovered building `djust-org/djust-start` (the official starter)
- djust 1.0.0rc12 (`470566c`)
- See companion css_framework setting in `LIVEVIEW_CONFIG` (bootstrap5 default)

johnrtipton:
## Summary

Closes #1619.

`dj_button` previously produced `class="btn btn-danger"` unconditionally from the `variant` keyword. But djust-theming's `components.css` (loaded by `theme_head`) only ships rules for `.btn-primary`, `.btn-secondary`, `.btn-destructive`, `.btn-ghost`, `.btn-link` — so `variant="danger"`, `variant="success"`, and `variant="warning"` rendered unstyled.

## Fix

New `_DJ_BUTTON_VARIANT_CLASS_MAP` constant maps keyword variants to canonical CSS class names. `danger` is now an alias for `destructive` (shadcn/Tailwind convention). Variants not in the map pass through as `btn-<variant>` via `conditional_escape`, preserving the security boundary and enabling user theme classes.

```
primary       -> btn-primary
secondary     -> btn-secondary
destructive   -> btn-destructive       (canonical)
danger        -> btn-destructive       (back-compat alias for #1619)
ghost         -> btn-ghost
link          -> btn-link
<other>       -> btn-<other>           (passthrough w/ escape)
```

## Two-commit shape (#1173)

- impl `178ed74b` — `python/djust/components/templatetags/djust_components.py` (+28 LoC) + new test file (+48 LoC)
- docs `283dab35` — CHANGELOG (+2 LoC)

## Test plan

- [x] 5 new regression cases in `python/djust/components/tests/test_dj_button_variant_1619.py`: danger alias, destructive canonical, primary regression backstop, unknown passthrough, XSS-escape preserved
- [x] Full Python suite: **7498 passed, 21 skipped, 0 failures**
- [x] Gate-the-fix-off self-test (Action #1200/#1468): reverting the map lookup makes the `test_danger_alias_renders_btn_destructive` test fail; restoring makes it pass
- [x] Disconfirming citation grep: no other tests or code assert on `btn-danger` as OUTPUT
- [x] `gallery/examples.py:158` (`variant="danger"`) — preserved by the alias

## Out of scope

- Adding `.btn-success`/`.btn-warning` CSS rules to components.css
- Renaming the `variant` keyword
- Touching `pack_css_generator.py` / `scaffold.css`
- Other components beyond `dj_button`


johnrtipton (commented):
## Stage 11 Review — APPROVE

**Stale-base**: BEHIND=0 vs origin/main. Clean.

**Tests**: 5/5 in `test_dj_button_variant_1619.py` pass; 62/62 in focused run; full suite 7498 passed / 21 skipped / 0 failed in 142s.

**Plan compliance**: matches `fix-1619-dj-button-variant-naming-plan.md` exactly — 6-entry `_DJ_BUTTON_VARIANT_CLASS_MAP` at line 1021 (primary/secondary/destructive/danger→destructive alias/ghost/link), passthrough with `conditional_escape` at line 1095, docstring updated to 5 canonical variants + alias note.

**Security boundary — verified clean.** The critical line `_DJ_BUTTON_VARIANT_CLASS_MAP.get(variant, f"btn-{conditional_escape(variant)}")` (`python/djust/components/templatetags/djust_components.py:1095`) correctly applies `conditional_escape` in the `.get()` default expression. Map values are framework-controlled literals (no escape needed). `variant_class` then flows into `mark_safe(f"...")` at line 1111 — both branches (map hit / escaped passthrough) are XSS-safe. Confirmed empirically by `test_unknown_variant_is_escaped` (injection payload becomes `&lt;script&gt;`).

**Gallery compatibility**: `python/djust/components/gallery/examples.py:158` (`variant="danger"`) now maps to styled `btn-destructive` — fixes the broken gallery preview that motivated the issue. No tests pin the old `btn-danger` output.

**Edge note**: `variant=None` evaluates the default → `btn-None`. Pre-existing behavior on main; not a regression.

**CHANGELOG**: present under `### Fixed`, cites #1619, names the constant, documents the back-compat alias and passthrough semantics. Action #1468 gate-the-fix-off self-test claim included.

**Verdict**: APPROVE. Single-file fix, surgical, security-preserving, well-tested.

`RETRO_COMPLETE`

johnrtipton:
# Retrospective — PR #1620

Closes #1619. Maps `dj_button(variant="danger")` to the actual CSS class
`.btn-destructive` that djust-theming's `components.css` ships, instead
of the orphan `.btn-danger` class with no matching rule.

## Quality: 5/5

Smallest sensible fix landed cleanly. ~28 LoC production, 48 LoC tests,
two-commit shape clean (impl `178ed74b` + docs `283dab35`), full suite
7498 passed / 0 failures. Gate-off self-test verified non-tautological.
Stage 11 reviewer APPROVE on first pass.

## What went well

- **Issue-time investigation surfaced the real surface fast.** The
  reporter said `components.css` has no `.btn-danger`. A 30-second
  grep across `python/djust/theming/` confirmed it AND revealed the
  twist: `scaffold.css` DOES have `.btn-danger`, but `theme_head.html`
  loads `components.css` only — not scaffold. Without the wider grep
  the fix would have been wrong (we'd have aliased a class that DOES
  exist somewhere just not in the loaded file). The lesson: when a
  reporter says "X has no rule," verify the LOADING path, not just
  the FILE.

- **User decision locked at issue-review time saved planning cycles.**
  AskUserQuestion at the prep stage resolved the Option 1 / 2 / 3 / 4
  trade-off (component-side rename, theme-side add, hybrid, minimal).
  The Planning stage then had a clean contract to write against
  rather than re-litigating the API design.

- **Map-based dispatch is the obvious right shape.** The original
  code did `f"btn-{conditional_escape(variant)}"` — one-line, but
  fundamentally unable to express the keyword/class indirection the
  reporter needed. The fix introduces the indirection via a constant
  dict at module scope; the dispatch is one `.get()` call. Reusing
  `conditional_escape` in the `.get()` default preserves the security
  boundary without any new escape logic.

- **5-test reproducer covers the bug class, not just the bug.** Tests
  cover: bug repro (danger→destructive), canonical name still works
  (destructive→destructive), regression backstop (primary unchanged),
  passthrough for custom variants, AND XSS-escape preserved on the
  passthrough path. The XSS test is the load-bearing security gate
  for the new code path.

- **`success`/`warning` deprecation is informational, not removal.**
  They still pass through as `btn-success`/`btn-warning` (just no CSS).
  Users with theme tokens via `pack_css_generator` still get styled
  output. Docstring no longer lists them — that's the only behavior
  change for those keywords.

## Lessons learned

- **`theme_head.html` is the gatekeeper for which CSS rules actually
  reach users.** `components.css` (loaded) vs `scaffold.css` (NOT
  loaded by default theme_head) is a real distinction. Future
  component bugs that "should work, see the CSS file" warrant a
  loading-path check before locking the fix.

- **ruff-format pre-commit hook bounced the first commit again** —
  fourth occurrence this session (PRs #1606, #1615, #1620). Pattern:
  hook reformats files, original commit doesn't register, re-stage +
  retry pattern restores. The `&& git log -1 --oneline` reflex
  (Action #122) caught it correctly. Cost ~30s per occurrence.

- **The reporter's framing was correct AND complete.** Unlike PRs
  #1617 and #1610 (where the reporter's CAUSAL theory was partially
  wrong), #1619's reporter pinpointed exactly the right cause AND
  proposed the exact right fix (Option 1). When the issue body
  already lists the right answer, planning compresses to verifying
  the audit + locking the test set.

## What could be improved

- **Gallery example.py:158 still says `variant="danger"`** even though
  `destructive` is now the canonical keyword. The back-compat alias
  preserves correctness, but the example should arguably use
  `destructive` to teach users the canonical name. Deferred per plan
  §7 as a fine-grained docs follow-up. IDEA below.

## IDEA: lines

- IDEA: **Follow-up: update `python/djust/components/gallery/examples.py:158`
  + any user-facing docs that still reference `variant="danger"`** to use
  `variant="destructive"` (canonical name). Tiny mechanical PR;
  preserves the alias for back-compat. Worth absorbing into the next
  components-touching PR rather than its own change.

- IDEA: **The `theme_head.html` loading-path detail (which CSS files
  actually reach the user) should be documented somewhere developers
  will find when debugging "why is my class unstyled?".** A note in
  the theming guide pointing to which CSS files `theme_head` loads
  vs. which exist in static/ would save the next person the
  detective work I did. Filing as a small docs improvement candidate.

- IDEA: **ruff-format pre-commit bounce keeps happening** (4× this
  session). The implementer subagent could be primed to `ruff format`
  staged files BEFORE `git commit` rather than catching the bounce
  reactively. Small prompt tweak; saves ~30s per pipeline.

`RETRO_COMPLETE`

PR #1622, opened 2026-05-27
feat(components): add dj_button confirm= kwarg for confirmation dialogs

### LINKED ISSUE #1621: dj_button: expose existing dj-confirm attribute via confirm= kwarg
## Bug / minor DX gap

djust already has built-in ``dj-confirm`` support — ``client.js`` handles it, ``schema.py`` documents it, and the scaffolding templates use it:

```html
<button dj-click=\"delete_item\" dj-confirm=\"Are you sure?\">Delete</button>
```

But the ``{% dj_button %}`` template tag from djust-components has a **fixed signature** with no ``confirm=`` kwarg and no pass-through for extra attributes:

```python
def dj_button(label=\"\", variant=\"primary\", event=\"\", icon=\"\", disabled=False, loading=False, size=\"md\", preset=\"\"):
```

So users who want the confirm dialog have to either skip ``dj_button`` (lose theme integration) or wrap it (ugly):

```html
<span onclick=\"if (!confirm('Wipe?')) event.stopImmediatePropagation()\">
  {% dj_button label=\"Reset\" event=\"reset\" variant=\"danger\" %}
</span>
```

## Suggested fix

Add ``confirm=\"…\"`` kwarg to ``dj_button`` that maps to the existing ``dj-confirm`` attribute:

```python
def dj_button(label=\"\", variant=\"primary\", event=\"\", confirm=\"\", icon=\"\", disabled=False, ...):
    ...
    confirm_attr = f' dj-confirm=\"{conditional_escape(confirm)}\"' if confirm else \"\"
    ...
```

Usage:

```html
{% dj_button label=\"Reset\" event=\"reset_demo\" variant=\"danger\" confirm=\"Wipe all data?\" %}
```

## Context

- Was filed broader (\"add confirmation support\") at #1621 originally
- Updated after finding the framework already has the underlying primitive
- Discovered in ``djust-org/djust-start`` reset-demo button
- djust 1.0.0rc12

johnrtipton:
## Summary

Closes #1621.

Adds `confirm=""` kwarg to `dj_button`. When non-empty, emits the standard `dj-confirm="<message>"` attribute consumed by djust's client.js (`python/djust/static/djust/src/09-event-binding.js:7`) — clicking the button shows a JS `confirm()` dialog; on OK fires the event, on Cancel nothing happens. The underlying primitive was already wired in client.js; this just exposes it through the component tag.

## Usage

```html
{% dj_button label="Reset" event="reset_demo" variant="danger" confirm="Wipe all data?" %}
```

renders:

```html
<button class="btn btn-destructive" dj-click="reset_demo" dj-confirm="Wipe all data?">
  <span class="btn-label">Reset</span>
</button>
```

## Two-commit shape (#1173)

- impl `12d2b997` — `python/djust/components/templatetags/djust_components.py` (+15 LoC) + new test file (+73 LoC)
- docs `892b8f65` — CHANGELOG (+2 LoC)

## Test plan

- [x] 5 new regression cases in `python/djust/components/tests/test_dj_button_confirm_1621.py`:
  - `confirm="Wipe?"` → `dj-confirm="Wipe?"` attribute emitted
  - default `confirm=""` → no `dj-confirm` attribute
  - `confirm="<script>..."` → escaped via `conditional_escape`
  - `confirm` without `event` → still emitted (independent attrs)
  - explicit `confirm` overrides preset-supplied `confirm`
- [x] Full Python suite: **7503 passed, 21 skipped, 0 failures**
- [x] First-commit-pass landed cleanly thanks to proactive `ruff format` (per retro #1620's IDEA)

## Design notes

- **`dj-confirm` emission is INDEPENDENT of `event` value** — client.js handles `dj-confirm` across multiple directives (`dj-click`, `dj-submit`, `dj-change`, ...). Forcing event-coupling would over-constrain.
- **`conditional_escape`** on the confirm message, matching the existing `event` attr's escaping at the same line.
- **Preset support** via the existing preset-priority pattern.

## Out of scope

- Other components (`dj-link`, `dj_action_button`, etc.) — separate per Action #1079 scope-discipline.
- Renaming `dj-confirm` attribute.
- Touching the client.js confirm handler.


johnrtipton (commented):
## Stage 11 review — APPROVE

Tight, focused 88-LoC patch that exposes the existing `dj-confirm` primitive. Branch is 0 behind `origin/main`; full suite 7503 passed / 21 skipped.

**Plan compliance** — exact match. `confirm=""` slotted between `event` and `icon` in the signature (`python/djust/components/templatetags/djust_components.py:1036`), added to `_defaults` (`:1073`), preset override path inserted between `event` and `icon` overrides (`:1085-1086`), emission gated on `if confirm:` after the `event` emission (`:1113-1119`) — exactly mirrors the AFTER block in the plan.

**Wire attr** — `dj-confirm` confirmed correct via `grep` against `python/djust/static/djust/src/09-event-binding.js:7` (`const confirmMsg = element.getAttribute('dj-confirm');`) and 8 handler call sites across click/submit/change/etc.

**XSS guard** — `conditional_escape(confirm)` matches the existing `event` attr escape pattern (`:1112`). `test_confirm_message_is_html_escaped` pins it with a `<script>alert(1)</script>` payload.

**Event-independence design** — `if confirm:` is unconditional on `event`; `test_confirm_emitted_without_event` (`:38`) pins the contract. Inline comment on `:1114-1118` documents the rationale so a future "tighten this" refactor will think twice.

**Two-commit shape** — clean: `12d2b997` touches only the templatetag + test file; `892b8f65` touches only `CHANGELOG.md`. Action #181 satisfied.

**Tautology gate (Action #1200/#1468)** — `test_no_confirm_no_attribute` asserts ABSENCE; gating the `if confirm:` block off makes `test_confirm_emits_dj_confirm_attribute` fail. Non-tautological.

**Minor (non-blocking)** — `test_explicit_confirm_overrides_preset` registers `test_1621_preset` process-wide with no cleanup (docstring acknowledges). Acceptable under a test-only name; flagging for the next preset-registry-API pass.

APPROVE.

johnrtipton:
# Retrospective — PR (#1621 close)

Closes #1621. `dj_button` accepts a `confirm=""` kwarg; emits the
standard `dj-confirm` attribute when non-empty. Closes the DX gap
where the framework primitive existed but the component tag didn't
expose it.

## Quality: 5/5

Smallest feasible patch. ~15 LoC production, ~73 LoC tests, two-commit
shape clean (impl `12d2b997` + docs `892b8f65`). Full suite 7503
passed, 0 failures. **First commit landed without ruff-format bounce**
— validating the "proactively format before commit" IDEA from PR #1620's
retro (5th occurrence of ruff-format would have been here if we hadn't
formatted up-front).

## What went well

- **Proactive `ruff format` saved a commit cycle.** Retro for PR #1620
  filed IDEA: "Implementer should `ruff format` staged files BEFORE
  `git commit`." Applied at Stage 5 of this pipeline:
  `.venv/bin/ruff format <staged-files>` before commit. Result:
  pre-commit ruff-format hook ran clean, commit registered on first
  pass. ~30 seconds saved + zero swallowed-commit risk. **The IDEA
  graduated to applied technique within one pipeline.**

- **The "expose existing primitive" pattern is the right shape.**
  The reporter explicitly noted that client.js ALREADY handles
  `dj-confirm` — the framework primitive existed; only the component
  tag was missing the kwarg. Verified at issue-review time by
  grepping `python/djust/static/djust/src/` for the literal. This
  saved scope-creep: no new wire format, no new JS handler, no docs
  spelunking. Just plumb the existing primitive through the tag.

- **Event-independence design decision was load-bearing**. The plan
  considered "should `confirm` work when `event=''`?" and chose YES
  (decoupled). client.js handles `dj-confirm` across multiple
  directives, so coupling would over-constrain. Test #4 pins this
  invariant: `confirm` emitted even without `event`. Without the
  explicit design decision in the plan, an implementer might have
  added an `if event` guard around the `if confirm` block —
  technically a bug, but plausible-looking.

- **Reporter's suggested fix was the right shape verbatim.** Sixth
  issue in a row from `djust-org/djust-start` where the reporter
  proposed the right code shape. The pattern: reporter says "we
  already have X primitive; expose it through Y interface." When
  the framework already has the primitive, the right answer is
  almost always "just plumb it."

- **Preset override path was free.** The existing preset-override
  loop just needed `confirm` added to the `_defaults` dict and one
  more if-clause. Tests confirm both directions (preset fills
  default; explicit overrides preset).

## Lessons learned

- **`dj-confirm` is one of djust's underused primitives.** It works
  across 7+ directives (`dj-click`, `dj-submit`, `dj-change`,
  `dj-input`, `dj-mouseenter`, etc. per `09-event-binding.js`) but
  is missing from most component tags. Filing IDEA: do the same
  audit pass for `dj-link`, `dj_action_button`, `dj_form_button`,
  and any other event-emitting component tag.

- **"Tiny PR with surface in the same file as a just-shipped PR"
  is a positive signal**, not a smell. #1619 (variant→class map)
  and #1621 (confirm= kwarg) both touch
  `python/djust/components/templatetags/djust_components.py`
  `dj_button`. Shipping them as 2 separate PRs (rather than
  bundling) preserved review clarity AND let the v1.0.0rc12-discovered
  symptoms close one at a time. The session's parallel rule (#180
  "one implementer per checkout") held by sequencing them.

## What could be improved

- **Test cleanup hygiene around preset registry**. Test 5
  registers a preset (`test_1621_preset`) and notes the registry
  has no public remove API, so the entry persists. This works for
  this test alone but would compound if many tests followed the
  pattern. Worth filing as a tech-debt follow-up: add a public
  `unregister_preset` (or auto-cleanup fixture) so test isolation
  is preserved.

## IDEA: lines

- IDEA: **Component-wide audit for `dj-confirm` exposure.** Repeat
  this pattern for `dj-link`, `dj_action_button`, `dj_form_button`,
  any other component tag that emits an event-bearing directive.
  Each component tag should accept `confirm=""` for parity. Small
  mechanical sweep; one PR could close several gaps.

- IDEA: **`unregister_preset` API for test isolation.** Currently
  `register_preset` mutates a process-wide registry with no public
  remove. Test 5 of this PR documents the gap. A
  `unregister_preset(tag, name)` + a pytest fixture would let test
  files register/cleanup presets without leaking across tests.

- IDEA: **The proactive `ruff format` step worked.** This was the
  first commit this session that DID NOT bounce on ruff-format. The
  Stage 5 implementer instructions should be updated to ALWAYS
  format staged files (Python and JS) before `git commit`. Tracked
  for the next pipeline-state.json template update.

`RETRO_COMPLETE`

PR #1623, opened 2026-05-28
fix(checks): V008 no longer false-fires on stdlib primitive-returning builtins

### LINKED ISSUE #1609: V008 false-positive on max() builtin in mount()
## Bug

`djust.V008` (\"Non-primitive type assigned to self.X in mount()\") fires when assigning the result of `max(...)` to an instance attribute. The static analyzer appears to inspect the bare `max` symbol rather than the return type.

## Repro

```python
from djust import LiveView

class BaseLiveView(PresenceMixin, LiveView):
        try:
            self.track_presence()
            self.online_count = max(1, len(self.list_presences()))
        except Exception:
            self.online_count = 1
```

```
?: (djust.V008) apps/shared/base_views.py:22 -- Non-primitive type 'max' assigned to self.online_count in mount(). Ensure this type is JSON-serializable.
```

The actual assigned value is always `int` — `max(int, int)` returns `int`, and the `except` branch assigns `1` (an int). Both branches are JSON-serializable.

## Workaround

```python
self.online_count = len(self.list_presences()) or 1
```

This makes V008 happy because there's no top-level call expression.

## Expected

Either:
1. V008 should infer the return type of stdlib builtins (`max`, `min`, `sum`, `len`, `abs`, `round`) and skip the warning when the return type is a primitive
2. Or the check should not trip on call expressions where the result is being stored — only flag direct assignments of non-primitive *values*

## Context

- Discovered in `djust-org/djust-start` (the official starter template) while implementing presence in `BaseLiveView`
- Checked against djust 1.0.0rc7
- Related but distinct from #1604 (suppress_checks not honoring V001/V005)

johnrtipton:
## Summary

Closes #1609.

V008 (`Non-primitive type assigned to self.X in mount()`) previously false-fired on `self.x = max(...)` because `max` wasn't in the `SAFE_TYPES` set — the check author included type-constructor names (`list`, `dict`, `str`, `int`) but missed stdlib builtins that always return primitives.

## Fix

Extend `SAFE_TYPES` in `python/djust/checks.py:1854` with:
- **Numeric** (→ int/float/tuple): `max`, `min`, `sum`, `abs`, `round`, `pow`, `divmod`, `len`, `ord`, `hash`, `id`
- **String-conversion** (→ str): `bin`, `oct`, `hex`, `repr`, `chr`, `ascii`, `format`
- **Container**: `sorted` (returns list, same element-serializability trust contract as `list()`)
- **Overlooked primitives**: `frozenset`, `bytes`

**Intentionally still flagged**:
- Iterator-returning builtins (`reversed`, `enumerate`, `zip`, `map`, `filter`, `range`, `iter`) — return iterators that aren't directly JSON-serializable
- `complex`, `slice` — not JSON-serializable

The V006 (Warning) path is untouched.

## Test plan

- [x] 9 new regression cases in `python/tests/test_checks_v008_builtins_1609.py`:
  - reporter's exact `max()` shape
  - numeric builtins suite (10 builtins)
  - string-conversion suite (7 builtins)
  - `sorted` and `frozenset` and `bytes`
  - regression backstops: custom class still fires, `complex` still fires, `reversed` still fires
- [x] 15 existing V008 tests at `python/tests/test_checks.py::TestV008NonPrimitiveInMount` continue to pass
- [x] Full Python suite: **7512 passed, 21 skipped, 0 failures**
- [x] First-commit-pass landed cleanly (proactive `ruff format`)

## Two-commit shape (#1173)

- impl `e7d9e270` — `python/djust/checks.py` (+41 LoC) + new test file (+155 LoC)
- docs `0b43477b` — CHANGELOG (+2 LoC)


johnrtipton (commented):
## Review — APPROVE

**Stale-base**: BEHIND=0. OK to review.

**Plan compliance (`python/djust/checks.py:1853-1905`)**:
- `SAFE_TYPES` extended exactly per plan: numeric (`max`/`min`/`sum`/`abs`/`round`/`pow`/`divmod`/`len`/`ord`/`hash`/`id`), conversion (`bin`/`oct`/`hex`/`repr`/`chr`/`ascii`/`format`), `sorted`, `frozenset`, `bytes`. Inline comment block documents the iterator-exclusion rationale + `complex`/`slice` exclusions inline — discoverable at the point of edit.
- V006 service-pattern path (`_SERVICE_INSTANCE_KEYWORDS` at line 68, consumed lines 1797/1949) untouched by the diff — confirmed via `git diff` grep.

**Focus areas**:
1. **Iterator regression backstop present.** `test_reversed_still_fires` at `python/tests/test_checks_v008_builtins_1609.py:145` asserts `reversed([1,2,3])` STILL emits V008. `test_complex_still_fires` mirrors this for `complex()`. Coverage of the intentional-exclusion contract is locked.
2. **`sorted` trust-contract documented.** Comment block at `checks.py:1896-1898` ("Same element-serializability contract as `list()` above") makes the asymmetry-with-`reversed` explicit.
3. **`_get_call_name` qualified-name spot-check**: `_get_call_name(ast.parse("obj.max(1,2)"))` returns `'obj.max'` (NOT in SAFE_TYPES). V008 still fires correctly on method form. Verified live.
4. **Existing test class clean**: `python/tests/test_checks.py::TestV008NonPrimitiveInMount` — 15/15 pass; no test pinned `max()` etc. as SHOULD-fire.
5. **Format-hook pattern durable**: 3rd consecutive PR landing clean with proactive `ruff format` — pattern is sticking.

**Tests**: 24/24 V008 cases pass (9 new + 15 existing).

**Risk**: minimal — additive set extension, no behavioral change for in-set inputs, regression backstops for excluded shapes.

`RETRO_COMPLETE`

johnrtipton:
# Retrospective — PR (#1609 close)

Closes #1609. V008 (`Non-primitive type assigned to self.X in mount()`)
extended to recognize stdlib primitive-returning builtins (`max`, `min`,
`sum`, `abs`, ...) as safe assignments.

## Quality: 5/5

Smallest possible fix shape. ~41 LoC additions to a single set literal,
~155 LoC of regression tests. Two-commit shape clean (impl `e7d9e270` +
docs `0b43477b`). First commit landed cleanly (proactive `ruff format`
again — graduated IDEA paying ongoing dividends). Full suite 7512
passed, 0 failures. 15 existing V008 tests + 9 new = 24 tests pinning
the check behavior.

## What went well

- **Set extension is the cheapest fix.** The V008 check has a single
  control-flow gate at `call_name not in SAFE_TYPES`. Adding entries
  to the set is mechanically the smallest possible change — no
  branching, no AST walker changes, no semantic shifts. The choice
  is just which entries to add.

- **Explicit exclusion list documented in code.** Iterator-returning
  builtins (`reversed`, `enumerate`, ...) are intentionally NOT added,
  with a multi-line comment explaining why. Future readers won't
  add them by mistake. Same for `complex`/`slice`.

- **Regression backstops are explicit, not implicit.** Tests assert
  both directions: builtins do NOT fire AND custom classes/iterators
  STILL fire. The check's semantic boundary is locked in test code.

- **Reporter's framing was right and complete.** Seventh issue from
  the `djust-org/djust-start` reporter where the cited cause was
  exactly right. The bug-report triage pattern this session — trust
  the symptom, refute the theory only when needed — applies here as
  "trust both the symptom and the theory."

- **Proactive `ruff format` graduated to standard practice.** This
  is the 3rd consecutive PR (#1620 → #1622 → #1623) where running
  ruff format BEFORE git commit prevented the pre-commit hook
  bounce. The IDEA from PR #1620's retro is now a session-internal
  habit. Worth canonicalizing into the Stage 5 implementer-prompt
  template.

## Lessons learned

- **The `_get_call_name` qualified-name distinction is load-bearing.**
  `_get_call_name(max(...))` returns the bare string `"max"`;
  `_get_call_name(obj.max(...))` returns `"obj.max"`. The check only
  short-circuits on bare names, so adding `max` to SAFE_TYPES doesn't
  accidentally whitelist `someservice.max()`. Verified during the
  edge-case audit in the plan.

- **`sorted` vs `list`: same trust contract, different shape.**
  `sorted([complex_objects])` returns a list of complex objects —
  not JSON-serializable. But `list([complex_objects])` has the same
  issue, and `list` was already in SAFE_TYPES. Consistent contract:
  the check verifies the CONSTRUCTOR/BUILTIN is primitive-returning;
  element serializability is the USER's responsibility. Documented in
  the SAFE_TYPES comment.

## What could be improved

- **The check could benefit from a separate `SAFE_BUILTINS` set
  unioned with `SAFE_TYPES`.** Currently everything lives in one set
  with comment-section separators. A union of two sets would make
  the distinction (type constructor vs builtin function) explicit
  in the code structure. Defer as a refactor candidate.

- **Documentation pass not done.** `docs/system-checks.md` likely
  has a V008 entry that doesn't mention the extended builtins
  recognition. Deferred per "out of scope" in the plan; worth a
  small follow-up.

## IDEA: lines

- IDEA: **Canonicalize "proactive ruff format" in the
  Stage 5 implementer-prompt template**. Three PRs in a row have
  benefited from this practice; it's reliable and saves ~30s per
  pipeline + zero swallowed-commit risk.

- IDEA: **`docs/system-checks.md` V008 entry refresh** — mention
  the recognized builtins set so users know they don't need to
  add `# noqa: V008` for `max()` etc.

- IDEA: **Split `SAFE_TYPES` into `SAFE_CONSTRUCTORS` +
  `SAFE_BUILTINS`** for clarity. Same functional behavior (union at
  the check site); just better code structure. Tracked as a
  v1.1+ refactor candidate.

`RETRO_COMPLETE`

Issue #1625, opened 2026-05-28
djust-components: code_block highlight.js bootstrap fails after WS broadcast patches


johnrtipton:
## Bug

\`{% code_block %}\` emits an inline \`<script>\` tag next to each \`<pre><code>\` element that lazy-loads highlight.js from CDN and calls \`hljs.highlightElement()\`. Works on initial HTTP-rendered page load. **Breaks for any \`<code>\` element that arrives via a djust WS patch** — InsertChild patches deliver the script-containing HTML but the browser doesn't execute scripts inserted that way.

Result: dynamically-rendered code blocks (or code blocks whose containing subtree is re-rendered after a state change) appear plain-text, not syntax-highlighted.

## Repro

Page has 3 \`{% code_block %}\` instances. Open the page → all 3 highlighted ✓. Trigger any state change that causes djust to re-render a parent subtree (in our case, an \`online_count\` change that toggles a sibling banner element). After the patch lands:

\`\`\`js
document.querySelectorAll('.code-block code').forEach(c =>
  console.log(c.dataset.highlighted, c.classList.contains('hljs'))
);
// undefined false   ← highlight stripped
// undefined false
// undefined false
\`\`\`

Even though \`window.hljs\` is still loaded globally, the new code elements have no \`.hljs\` class and no \`data-highlighted=\"true\"\` marker.

## Root cause

Two compounding issues:

1. \`<script>\` tags inserted via \`innerHTML\` / DOM manipulation **do not execute** in modern browsers. djust's patch system uses such manipulation, so the inline highlight bootstrap script never runs for re-inserted code blocks.

2. Even if the script DID run, \`document.currentScript.previousElementSibling\` may not point to the new code block depending on how the patch sequence reconstructs siblings.

## Workaround

A MutationObserver in base template that re-runs hljs on any unprocessed code block:

\`\`\`html
<script>
(function () {
  function hl() {
    if (!window.hljs) return;
    document.querySelectorAll('.code-block code[class^=\"language-\"]:not(.hljs)')
      .forEach(function (el) { hljs.highlightElement(el); });
  }
  window.addEventListener('load', hl);
  new MutationObserver(hl).observe(document.body, { childList: true, subtree: true });
})();
</script>
\`\`\`

Works but is invasive — every djust app using \`{% code_block %}\` needs this snippet.

## Suggested fix

\`code_block\` should bootstrap once at page level, not per-instance:

[...]

PR #1626, opened 2026-05-28
fix(theming): theme_head auto-loads djust-components's components.css

### LINKED ISSUE #1624: djust-components: components.css not loaded by theme_head — layout breaks for code_block, card, etc.
## Bug

\`{% theme_head %}\` (from djust-theming) loads djust-theming's \`components.css\` automatically. But djust-components has its **own separate \`components.css\`** at \`/static/djust_components/components.css\` which is NOT loaded by \`theme_head\`. Without it, layout rules for components rendered by tags like \`{% code_block %}\` and \`{% card %}\` are missing — they fall back to default flow layout, which looks broken.

## Repro

Default starter setup (\`INSTALLED_APPS\` includes \`djust.components\` and \`djust.theming\`, base template has \`{% theme_head %}\`):

\`\`\`html
{% load djust_components %}
{% code_block language="python" filename="x.py" code="print(1)" %}
\`\`\`

Rendered:

\`\`\`html
<div class=\"code-block\">
  <div class=\"code-block-header\">
    <span class=\"code-block-filename\">x.py</span>
    <span class=\"code-block-lang\">python</span>
    <button class=\"code-block-copy\">Copy</button>
  </div>
  <pre class=\"code-block-pre\"><code class=\"language-python\">print(1)</code></pre>
</div>
\`\`\`

### Expected layout (from \`djust_components/components.css\`):

\`\`\`
.code-block-header { display: flex; ... }
.code-block-copy   { margin-inline-start: auto; ... }   ← pushes Copy to right
\`\`\`

\`[filename] [lang]                                       [Copy]\`

### Actual layout (without \`djust_components/components.css\` loaded):

\`[filename] [lang] [Copy]\` — all clumped on the left, Copy button 256px from the right edge.

The same issue affects \`{% card %}\` headers, \`{% dj_button %}\` spinners, and any component whose layout depends on \`djust_components/components.css\`.

## Workaround

Manually link the file in base template:

\`\`\`html
{% theme_head %}
<link rel=\"stylesheet\" href=\"{% static 'djust_components/components.css' %}\">
\`\`\`

## Suggested fix

Either:

1. **\`theme_head\` includes djust-components's CSS** when \`djust.components\` is in \`INSTALLED_APPS\`. Easy detection via \`django.apps.apps.is_installed('djust.components')\`.

2. **A separate \`{% components_head %}\` tag** from djust-components that does the same thing. Then docs tell users to include both.

Option 1 is zero-config and matches user expectations (if I install djust[components,theming] and use \`{% theme_head %}\`, I expect both to work).

## Context

- Discovered in \`djust-org/djust-start\` while wiring up \`{% code_block %}\`
- djust 1.0.0rc12 (\`1b685cb\`)
- Verified the file IS being served — just not linked from any template that the user's base.html extends

johnrtipton:
## Summary

Closes #1624.

`{% theme_head %}` previously loaded only djust-theming's `components.css`. djust-components ships a separate `components.css` with layout rules for `{% code_block %}`, `{% card %}`, `{% dj_button %}` spinners, etc. — components fell back to default flow layout (looked broken). Fix auto-detects `djust.components` via `apps.is_installed` and emits a conditional `<link>` next to the existing djust-theming link.

## Fix

```python
# python/djust/theming/templatetags/theme_tags.py
try:
    from django.apps import apps as django_apps
    include_components_app_link = django_apps.is_installed("djust.components")
except Exception:
    include_components_app_link = False
```

```django
{# python/djust/theming/templates/djust_theming/theme_head.html #}
{% if include_components_app_link %}<link rel="stylesheet" href="{% static 'djust_components/components.css' %}">{% endif %}
```

## Two-commit shape (#1173)

- impl `cf5b830f` — `theme_tags.py` (+14 LoC) + template (+1 line) + new test (+102 LoC) + keyset-invariant test update (+1 line)
- docs `225f3ef0` — CHANGELOG (+2 LoC)

## Test plan

- [x] 6 new regression cases (link emission both ways, builder detection 3 paths, regression backstop)
- [x] Pre-mount/post-mount keyset invariant test (#1123-style) updated for the new context key
- [x] Full Python suite: **7518 passed, 21 skipped, 0 failures**
- [x] Defensive `try/except` for unconfigured Django scenarios

## Out of scope

- Loading djust-components JS assets — separate concern (touches #1625)
- Loading `components-classes.css` — separate concern
- `css_prefix` interaction — djust-components CSS doesn't support prefixing today


johnrtipton (commented):
## Review verdict: APPROVE

**Plan compliance** (all hit):
- `theme_tags.py:147-156` adds defensive `try/except` around `apps.is_installed("djust.components")`; `except Exception` is justified by the "never break theme_head" comment.
- `theme_tags.py:176` adds `include_components_app_link` to return dict.
- `theme_head.html:26` emits the conditional `<link>` to `static/djust_components/components.css`; existing `include_component_link` line (djust-theming's own components.css) untouched.
- `test_theming_context_cache.py:502` updates the #1123 pre/post-mount keyset invariant test to include the new key — without this, drift detector would fire.

**Focus-area findings:**
1. **No XSS surface**: `<link href>` uses `{% static 'djust_components/components.css' %}` — literal path, no user input.
2. **Graceful degradation pinned**: `test_returns_false_when_app_not_installed` + `test_defensive_when_apps_raises` cover both negative paths.
3. **Detection orthogonality**: `include_components_app_link` is independent of the `include_component_link` resume-optimization branch (line 144 sets `include_component_link=False` when `component_css` is inlined, but the new flag is computed unconditionally after). Both `<link>` lines emit independently — correct.
4. **CSS prefix path**: out-of-scope per plan §3 — diff doesn't touch the prefix path; existing `test_theming_css_prefix.py` continues to pass (16 cases green).
5. **Test coverage**: 6 new cases in `test_theme_head_components_link_1624.py` cover both render-time (flag → HTML) and detection-time (is_installed → flag) layers.

**Tests**: targeted 73/73 pass; full `python/ tests/` suite 7518 passed / 21 skipped in 135s.

**Stale-base**: BEHIND=0 vs origin/main.

Two-commit shape (impl + docs) per Action #181. CHANGELOG entry present.

APPROVE.

johnrtipton:
# Retrospective — PR #1626

Closes #1624. `{% theme_head %}` now auto-loads djust-components's
`components.css` when `djust.components` is in `INSTALLED_APPS`. Closes
a DX gap where component layout (`{% code_block %}`, `{% card %}`,
`{% dj_button %}` spinners) silently fell back to default flow layout
because the static CSS file existed but wasn't linked.

## Quality: 5/5

Smallest contained fix: ~14 LoC production, ~102 LoC tests, defensive
`try/except`, graceful degradation, two-commit shape clean (impl
`cf5b830f` + docs `225f3ef0`). Stage 11 reviewer APPROVE on first
pass. Full suite 7518 passed / 0 failures.

## What went well

- **The `#1123` keyset invariant test caught a drift immediately**.
  Adding a new context key to `build_theme_head_context` without
  updating `TestThemeMixinThemeHead::test_build_theme_head_context_keyset`
  would have been a 🔴 Stage 11 finding. The invariant fired exactly
  as designed — caught the omission in the initial test run, fixed
  in the same impl commit. Action #1123 (pre-mount/post-mount keyset
  invariant) is paying ongoing dividends; this is the second drift
  it caught this session.

- **`apps.is_installed` is the right primitive**. Django's app-config
  registry is the canonical "is this app present" check; we already
  use it elsewhere in the framework. Wrapped in `try/except` for the
  "app registry not populated yet" case (which happens in isolated
  unit tests + `setup_test_environment` ordering edge cases). 3
  test cases pin all three branches: True, False, raises.

- **Defensive default is False**. If detection fails (raises, app
  registry uninitialized, etc.), we emit no extra link — same as if
  djust.components were not installed. This is the safe direction:
  worst case is unstyled components, never broken pages from a bogus
  `<link>`.

- **Two-commit shape clean again** (5th consecutive). Proactive
  `ruff format` paid off — the first commit bounced because the test
  file had a formatting quirk ruff fixed, then the second commit
  passed cleanly. Total cost ~30s.

## Lessons learned

- **`build_theme_head_context` is the central source of truth for
  the theme_head template context**. Three call sites consume it
  (`theme_head` simple tag, `ThemeMixin._setup_theme_context`, the
  cache layer). Adding a new key here propagates through all three
  paths uniformly. The contract is enforced by the keyset invariant
  test. This is a good architectural pattern that other multi-call-site
  context builders should follow.

## What could be improved

- **`components-classes.css` and djust-components JS assets remain
  unlinked**. The reporter explicitly cited `components.css` so we
  stayed scoped per Action #1079 (broader-sweep → follow-up issue).
  Worth filing a follow-up issue: should `theme_head` also link
  `components-classes.css` and emit the necessary `<script>` tags?
  Tracked below.

## IDEA: lines

- IDEA: **Audit other djust-components static assets**. The fix
  only handles `components.css`. djust-components ships
  `components-classes.css` and several `*.js` files (`code-snippet.js`,
  `data-table.js`, etc.). Some are auto-included by individual
  template tags (per-tag `<script>` injection); others may be missing
  entirely. Worth a separate issue to audit + auto-include.

- IDEA: **Companion #1625 fix coming next** — the reporter filed both
  in the same session. #1625 is about `{% code_block %}`'s highlight.js
  bootstrap not running after WS patches. Sequential fix.

- IDEA: **The defensive `except Exception` is broad** but justified
  by the "never break theme_head" invariant. If a future audit
  tightens broad excepts framework-wide, this one needs a
  `# noqa: BLE001` annotation with the existing justification
  comment as the rationale.

`RETRO_COMPLETE`

PR #1627, opened 2026-05-28
fix(components): code_block installs MutationObserver for WS-patch-inserted code

### LINKED ISSUE #1625: djust-components: code_block highlight.js bootstrap fails after WS broadcast patches

johnrtipton:
## Summary

Closes #1625.

`{% code_block %}`'s per-instance inline `<script>` doesn't execute for code blocks inserted via WS patches (modern browsers don't run scripts inserted via `innerHTML`). Fix installs a MutationObserver ONCE per page (idempotent via `__djcHljsObserverInstalled`) that watches `document.body` for added `<pre><code class="language-*">` elements and highlights any unmarked ones.

## Fix

The observer is installed on each of the three hljs-ready paths in the existing bootstrap:
1. hljs already loaded
2. hljs first-load completes (`s.onload`)
3. parallel-load poll succeeds

Observer is purely additive — per-instance inline scripts still run on initial HTTP page load (unchanged behavior). Feature-detected via `typeof MutationObserver === 'undefined'` for graceful degradation.

## Two-commit shape (#1173)

- impl `2d0013bb` — `python/djust/components/templatetags/djust_components.py` (+29 LoC) + test file (+62 LoC)
- docs `f88b27e5` — CHANGELOG (+2 LoC)

## Test plan

- [x] 6 new regression cases in `python/djust/components/tests/test_code_block_observer_1625.py`:
  - `MutationObserver` string present
  - `__djcHljsObserverInstalled` idempotency flag present
  - `childList` + `subtree` options requested
  - `language-` prefix selector present
  - `highlight=False` path unchanged (regression backstop)
  - existing lazy-loader preserved (`__djcHljsLoading` + CDN URL)
- [x] Full Python suite: **7524 passed, 21 skipped, 0 failures**
- [x] First-commit-pass landed cleanly via proactive `ruff format` (after one re-stage cycle)

## Out of scope

- Moving lazy-load into a separate static JS file (reporter's Option 1)
- `djust:patch-applied` event from client.js (reporter's Option 3 — touches framework)
- Vendoring hljs instead of CDN


johnrtipton (commented):
## Stage 11 review — PR #1627 (#1625 code_block MutationObserver)

**Verdict: APPROVE.** Stale-base check: BEHIND=0. Full suite 7524 passed / 21 skipped. Focused suite 95/95 (6 new in `test_code_block_observer_1625.py`).

### Plan compliance (all met)
- `installObserver()` defined with first-line idempotency gate (`djust_components.py:1855` — `if(window.__djcHljsObserverInstalled)return;`).
- MutationObserver feature-detection at `:1856` (`typeof MutationObserver==='undefined'` → no-op on legacy browsers).
- Three call sites: hljs-already-loaded (`:1871`), `s.onload` after CDN fetch (`:1881`), interval poll (`:1884`).
- Scope `document.body, {childList:true, subtree:true}` (`:1869`).
- Handler defends against TEXT nodes via `if(n.nodeType===1)hl(n)` (`:1867`).
- Subtree recursion: `root.matches?...:root.querySelectorAll` covers direct insertion AND parent-containing-code insertion (`:1860-1862`).
- `highlight=False` path completely unchanged — observer install is inside `if highlight:` (`:1837`). Confirmed by `test_highlight_false_path_unchanged`.

### 🟢 Notes (no blockers)
1. **Gate-off limitation.** Source-text tests pin the `MutationObserver` / `__djcHljsObserverInstalled` strings; they'd still pass if `installObserver()` were defined but never *called*. A future JSDOM live test exercising real `s.onload` → observer-fires-on-injected-node would close this gap. Acceptable given Python-only test infra and the inline-script context.
2. **Existing `mark_safe(f'...')` on `e_theme`** at the `code-block` wrapper is pre-existing, not in this PR's diff — `e_theme = escape(theme)` so pre-escaped. Out of scope.

CHANGELOG entry present. Two commits (impl + docs) match plan.

APPROVE

johnrtipton:
# Retrospective — PR #1627

Closes #1625. `{% code_block %}` now installs a MutationObserver
once per page so WS-patch-inserted code blocks get syntax-highlighted
automatically. Companion fix to PR #1626 (theme_head loading
components.css) — both surfaced from the same `djust-org/djust-start`
build session.

## Quality: 5/5

29 LoC of new JS embedded in a Python string literal, 62 LoC of
source-text tests, two-commit shape clean (impl `2d0013bb` + docs
`f88b27e5`). Stage 11 reviewer APPROVE with one 🟢 nit
(source-text-only gate-off — same trade-off PR #1616 documented).
Full suite 7524 passed / 0 failures.

## What went well

- **The "install once via window flag" pattern is robust.** Multiple
  `{% code_block %}` instances each emit their own inline script, each
  calls `installObserver()`, but only the FIRST actually installs
  (subsequent calls return early via `__djcHljsObserverInstalled`).
  Same trick the existing `__djcHljsLoading` flag uses for lazy-load
  deduplication. The pattern is established in this file; the
  observer install just extends it.

- **Observer installed on all three hljs-ready exit paths**.
  Already-loaded → install + highlight + return. First-load
  (`s.onload`) → install + bulk-catch-up. Parallel-load poll →
  install on ready. Without all three, the observer would only
  fire in some load-timing scenarios and the bug would have
  surface-area cases left.

- **Feature-detected via `typeof MutationObserver === 'undefined'`**.
  Very old browsers fall through to existing per-instance behavior —
  no break, no degraded experience beyond what they already had.

- **Subtree recursion via `root.matches(sel)` + `root.querySelectorAll(sel)`**.
  Mutation records report `addedNodes` at the level of the actual
  inserted element. For WS patches, that's typically a container
  (e.g. `<dj-view>` re-render) — the matcher needs to descend into
  the subtree to find nested code elements. The handler does both
  (direct match AND descendant scan).

- **`highlight=False` path completely unchanged**. The observer
  install lives inside the existing `if highlight:` block, so users
  opting out of hljs see no observer install, no MutationObserver
  string in their HTML, no behavior change.

## Lessons learned

- **JS-in-Python-string literals get gnarly fast**. The original
  bootstrap was ~22 lines of compressed JS in one f-string; this
  fix adds another 15. Total inline-JS payload is now ~37 lines.
  At some point this should be factored into a static JS file
  (`djust_components/code-block-highlight.js`) referenced by a
  one-time `<script>` tag. The reporter's Option 1 was exactly that;
  we chose Option 2 (MutationObserver) for minimal scope, but the
  static-file refactor is a natural follow-up.

- **Source-text gate-off is limited but acceptable here**. Same
  pattern PR #1616 documented: the literal `MutationObserver` string
  appears in the `installObserver()` function definition. Removing
  the install call sites still leaves the string. The 🟢 reviewer
  note correctly flagged this. A JSDOM live test would close the gap.

## What could be improved

- **CodeQL false-positive cost a CI re-run**. The original
  `assert "cdn.jsdelivr.net" in html` test assertion tripped CodeQL's
  `js/incomplete-url-substring-sanitization` rule — classic
  anti-pattern detection (bare host substring checks are a common
  XSS source when used for real sanitization). CodeQL can't tell a
  test assertion from a runtime sanitizer. Fix shipped in commit
  `433cbfe1`: use the path-specific substring `highlightjs/cdn-release`
  (unique to the highlight.js CDN URL) instead of the bare host.
  Worth canonicalizing: **test assertions for URL presence should use
  path-specific substrings, not host-only**. Filed as IDEA below.

- **No JSDOM live test for the WS-patch repro**. The reporter
  described a concrete scenario: 3 code_block instances on page,
  trigger a re-render, check that `document.querySelectorAll('.code-block code').forEach(c => c.dataset.highlighted)`
  returns `'true'` for all. This would require constructing a JSDOM
  harness that simulates the WS patch (innerHTML replacement) and
  awaits the MutationObserver microtask. Worth a follow-up issue.

## IDEA: lines

- IDEA: **Factor the lazy-loader + observer into a static JS file**
  (`djust_components/code-block-highlight.js`) and have the
  per-instance inline script just call a global init function with
  the theme name. Total inline payload would drop from ~37 lines to
  ~3. Tracked for v1.1+ refactor.

- IDEA: **JSDOM live test for the WS-patch scenario** — JSDOM
  harness that simulates 3 code_block instances + a subsequent
  innerHTML replacement, then asserts `data-highlighted="true"` on
  all of them after a microtask. Would close the gap from #1616's
  retro AND this retro.

- IDEA: **Canonicalize "test URL assertions use path-specific
  substrings, not host-only"** in the Stage 7 self-review checklist.
  CodeQL's `js/incomplete-url-substring-sanitization` rule is broad
  and fires on test assertions; the path-specific-substring fix is
  reliable. Worth a one-line addition to the auto-reject triggers.

- IDEA: **Other components likely have the same WS-patch script
  execution issue**. Any component that emits a per-instance
  inline `<script>` for runtime setup is vulnerable (e.g.,
  copy-button, certain modal forms). Worth an audit pass.

`RETRO_COMPLETE`

PR #1629, opened 2026-05-28
fix(checks): V008 no longer false-fires on stdlib module functions

### LINKED ISSUE #1628: V008 still false-fires on inspect.getsource() (and other stdlib module functions) — #1609 scope was builtins-only
## Follow-up to #1609 (closed by #1623)

PR #1623 fixed V008 for **builtins** (`max`, `min`, `len`, `str`, `int`, …), but the check still trips on stdlib **module functions** that return primitives — e.g. \`inspect.getsource()\`, which returns \`str\`.

## Repro

\`\`\`python
import inspect
from djust import LiveView

class HomeView(LiveView):
    template_name = \"home.html\"

        self.react_src = inspect.getsource(HomeView.react)  # ← V008 fires
\`\`\`

\`python manage.py check\` output (on djust 1.0.0rc12 / commit 9ace705):

\`\`\`
?: (djust.V008) apps/home/views.py:27 -- Non-primitive type 'inspect.getsource'
   assigned to self.react_src in mount(). Ensure this type is JSON-serializable.
\`\`\`

\`inspect.getsource\` always returns \`str\` — it is JSON-serializable.

## Real-world use case

[djust-org/djust-start](https://github.com/djust-org/djust-start) uses \`inspect.getsource(method)\` to feed live code panels next to the running demo — the visitor sees the exact Python that powers the buttons they're clicking. The starter currently still has to carry \`\"V008\"\` in \`suppress_checks\` despite #1623.

## Proposed scope expansion

Extend the V008 known-primitive-returning-callable allowlist to cover:

- \`inspect.getsource\`, \`inspect.getsourcefile\`, \`inspect.getsourcelines\`, \`inspect.getmodule\`, \`inspect.getdoc\` (primitive or near-primitive returns)
- \`os.path.join\`, \`os.path.basename\`, \`os.path.dirname\`, \`os.getenv\`, \`os.path.exists\`, etc. (string / bool primitives)
- \`pathlib.Path.read_text\`, \`Path.exists\`, \`Path.is_file\`, \`Path.name\`, \`Path.suffix\` (string / bool)
- \`datetime.datetime.isoformat\`, \`datetime.date.isoformat\` (string)
- \`json.dumps\` (string)

A simpler alternative: rather than extend the allowlist piecewise, **trust any callable whose return annotation is a primitive type** (\`-> str\`, \`-> int\`, \`-> bool\`, \`-> float\`, \`-> bytes\`). The Python stdlib is well-annotated; user code that wants to opt in can add a return annotation.

## Workaround (current)

\`\`\`python
DJUST_CONFIG = {\"suppress_checks\": [\"V008\"]}
\`\`\`

(See [djust-start commit ff3ea3d](https://github.com/djust-org/djust-start/commit/ff3ea3d) for the comment in context.)

## Repo bumps

Filed during the djust-start session that produced #1604, #1605, #1610–#1614, #1617, #1619, #1621, #1624, #1625 — all merged. After this lands, the starter has literal zero local accommodations for framework quirks.

johnrtipton:
## Summary

Closes #1628.

Follow-up to #1609/#1623. The bare-builtin fix in #1623 missed qualified stdlib calls like `inspect.getsource()`, `os.path.join()`, `json.dumps()`, `Path.read_text()` because `_get_call_name` returns the dotted name for attribute-access calls — those didn't match the bare-name `SAFE_TYPES` entries.

## Fix

Extend `SAFE_TYPES` with stdlib qualified names + two bare method names for chained-call forms.

**Qualified** (`mod.fn(...)`):
- `inspect.getsource/getsourcefile/getmodule/getdoc`
- `os.path.join/basename/dirname/exists/isfile/isdir/abspath/relpath`
- `os.getenv/getcwd`
- `pathlib.Path.read_text/exists/is_file/is_dir` (unbound class form)
- `json.dumps`
- `datetime.datetime.isoformat`, `datetime.date.isoformat`

**Bare method names** (chained-call: `Path(p).method()` resolves to bare name):
- `isoformat`, `read_text`

**Intentionally NOT added**: bare `exists`/`is_file`/`is_dir` (too ambiguous — `some_record.exists()` is common in user code), `inspect.getsourcelines` (returns tuple), `complex`/`slice` (not JSON-serializable). Reporter's alternative (annotation-based trust) deferred as a significant scope expansion.

## Two-commit shape (#1173)

- impl `8e4357a6` — `python/djust/checks.py` (+36 LoC) + new test file (+167 LoC)
- docs `8f87a36f` — CHANGELOG (+2 LoC)

## Test plan

- [x] 10 new regression cases in `python/tests/test_checks_v008_stdlib_qualified_1628.py`
- [x] 34 V008 tests green total (10 new + 9 from #1623 + 15 pre-existing)
- [x] Full Python suite: **7534 passed, 21 skipped, 0 failures**
- [x] Backstops: custom class still fires; bare `obj.exists()` still fires (intentionally excluded)

After this lands, `djust-org/djust-start` has zero local accommodations for framework quirks.


johnrtipton (commented):
## Stage 11 Review — PR #1629 (V008 stdlib qualified-name extension)

**Verdict: APPROVE**

### Plan compliance
- `SAFE_TYPES` extended with 21 qualified names (`inspect.*`, `os.path.*`, `os.getenv`/`getcwd`, `pathlib.Path.*`, `json.dumps`, `datetime.*.isoformat`) at `python/djust/checks.py:1903-1929` — matches plan §3 verbatim.
- Two bare method names (`isoformat`, `read_text`) at `python/djust/checks.py:1937-1938` for chained-call forms — addresses the `_get_call_name`-returns-bare-name finding (plan §1).
- Bare `exists`/`is_file`/`is_dir` INTENTIONALLY OMITTED, documented in comment at `python/djust/checks.py:1931-1936` (collision risk with `User.objects.filter(...).exists()` etc.).
- 10 regression cases at `python/tests/test_checks_v008_stdlib_qualified_1628.py:28-164` covering all paths + 2 backstops (custom class line 137, `obj.exists()` line 150).

### Tests
- Focused: 19/19 pass (10 new + 9 from #1623).
- Full suite: 7534 passed, 21 skipped, no regressions.

### Gate-the-fix-off self-test (Action #1200 / #1468)
Empirically commented out the new SAFE_TYPES entries and re-ran `test_checks_v008_stdlib_qualified_1628.py`: **8 of 10 tests FAILED** (all positive cases). The 2 that still pass are the intentional backstops (`custom_class_still_fires`, `arbitrary_obj_exists_still_fires`) which should pass regardless of the fix — they're gate-off canaries by design. Tests are non-tautological.

### Architectural notes
- Same mechanism as #1623 — extends an existing set, no new code path. Low blast radius.
- The "bare `exists` NOT in set" decision is load-bearing and documented inline; future contributors who try to "complete the set" will trip the `test_arbitrary_obj_exists_still_fires` backstop.
- Reporter's annotation-based alternative correctly deferred (would require import resolution at static-check time).

CHANGELOG entry comprehensive and traces the design trade-offs.

johnrtipton:
# Retrospective — PR #1629

Closes #1628. Extends V008's `SAFE_TYPES` allowlist with stdlib
qualified names (`inspect.getsource`, `os.path.*`, `pathlib.Path.*`,
`json.dumps`, `datetime.*.isoformat`). Follow-up to #1609/#1623 — the
bare-builtin fix in #1623 missed qualified calls.

## Quality: 5/5

Surgical extension of the existing pattern. ~36 LoC production, ~167
LoC tests, two-commit shape clean (impl `8e4357a6` + docs `8f87a36f`).
Stage 11 reviewer APPROVE on first pass with explicit gate-off
empirical canary (commented out the new entries, observed 8/10 fail —
non-tautological).

## What went well

- **The `_get_call_name` probe at issue-review time was load-bearing**.
  Running `python -c "import ast; print(_get_call_name(...))"` on the
  reporter's example shapes BEFORE writing the plan surfaced that
  `Path(p).exists()` resolves to bare `'exists'` (not
  `'pathlib.Path.exists'`) because the receiver is a Call node, not
  an Attribute chain. Without that probe, the fix would have only
  covered half the cited cases (qualified `pathlib.Path.exists(p)`
  works; chained `Path(p).exists()` would still fire). 30-second
  probe saved a Stage 11 finding.

- **Intentional non-inclusion is locked in test code, not just
  comments**. `test_arbitrary_obj_exists_still_fires` asserts that
  `obj.exists()` STILL fires V008. A future contributor who tries
  to "complete the set" by adding bare `exists` will hit this
  regression test immediately. Same pattern PR #1623's
  iterator-still-fires tests established.

- **Same SAFE_TYPES pattern as #1623 = no new architecture**. The
  fix shape is mechanical: add string literals to an existing set.
  Reviewer can verify in seconds; no new code paths to reason about.

- **Reporter's Option 2 (annotation-based trust) correctly deferred**.
  The reporter offered a more ambitious fix — `-> str` return
  annotation detection. That would require resolving imports +
  introspecting target module annotations at static-check time, which
  is significant scope expansion. The allowlist is the surgical fix
  for cited cases; the annotation approach can land later if more
  stdlib false-positives appear and the maintenance cost of the
  allowlist becomes painful.

- **Proactive `ruff format` pattern still pays off** (5th consecutive
  PR landing on first-or-second commit pass thanks to formatting
  ahead of the hook).

## Lessons learned

- **Allowlist + intentional-non-inclusion + backstop tests is a
  durable pattern**. Three pieces:
  1. Positive coverage: add the desired entries.
  2. Code comment documenting WHAT IS NOT included and WHY.
  3. Regression test asserting the non-included case STILL fails.

  Without #3, a future contributor sees the "obviously missing"
  entry (e.g., bare `exists`) and adds it without realizing the
  user-code-collision trade-off. With #3, they hit a failing test
  and read the comment.

- **AST `_get_call_name` returns different shapes for chained calls
  vs unbound class-method calls**. Document this gotcha somewhere
  reviewers will find it. The qualified form
  `pathlib.Path.exists(p)` → `'pathlib.Path.exists'`. The chained
  form `Path(p).exists()` → `'exists'`. Both should be tested for
  any V008-relevant entry. Worth a comment block on `_get_call_name`
  itself.

## What could be improved

- **The reporter's Option 2 (annotation-based trust)** is the real
  scalable solution. Every new stdlib false-positive triggers
  another bug → allowlist extension. The annotation-based approach
  would handle the entire stdlib + any annotated user code for free.
  Worth a v1.1+ design exploration: cost of importing target modules
  during static check vs. ongoing allowlist maintenance.

## IDEA: lines

- IDEA: **Annotation-based primitive-return trust for V008**.
  Reporter's Option 2. The static analyzer would: (a) read user
  imports at the top of the file being checked, (b) for each
  qualified call, resolve the import to the target module, (c)
  inspect the callable's return annotation. If primitive (`-> str`
  / `-> int` / etc.), skip V008. Significant work but eliminates
  the allowlist maintenance burden. Tracked for v1.1+.

- IDEA: **`_get_call_name` returning bare names for chained calls
  is a footgun**. Comment block on the helper documenting the
  three shapes (`Name`, `Attribute chain`, `chained Call.method`)
  and what they resolve to. Saves future investigation cycles.

- IDEA: **Audit `os.path.getsize`, `os.path.getmtime`, etc.** —
  they return ints, are JSON-serializable, weren't cited by the
  reporter. Worth a small follow-up if any user hits them.

`RETRO_COMPLETE`

Issue #1108, opened 2026-04-26
tech-debt: filter-shape parameter contract — Iterable[T], not list[T]


johnrtipton:
**Source**: PR #1107 retro (https://github.com/djust-org/djust/pull/1107#issuecomment-4322729335)

When a parameter accepts any iterable for membership checks (e.g. `fname in rendered_filter`), annotating as `list[T] | None` narrows the contract incorrectly — tuple/set/frozenset all work too. PR #1107 had `wizard_rendered_fields: list | None = None` and tested only the list case. Stage 11 caught it; removed the annotation and added tuple + set tests.

**Pattern to canonicalize in CLAUDE.md** (Python type annotations / API contracts):
- For membership-check parameters, document the contract semantically as `Iterable[T]` (or omit annotation entirely)
- Test at least one non-list shape (tuple OR set) to lock the contract in

**Action**: add to CLAUDE.md or PR checklist.

johnrtipton:
Canonicalized in PR #1116 (CLAUDE.md / Process canonicalizations from PR retros section).

PR #1631, opened 2026-05-28
fix(theming): break presets/registry/manager/css_generator cyclic-import SCC

johnrtipton:
## Summary

Closes 4 open CodeQL `py/cyclic-import` alerts (severity: note) on `main`. These are code-scanning alert IDs, NOT issues — the four-module SCC is the root cause:

| CodeQL alert | File:Line |
|--------------|-----------|
| 2352 | `python/djust/theming/manager.py:15` |
| 2351 | `python/djust/theming/presets.py:171` |
| 1900 | `python/djust/theming/registry.py:127` |
| 1883 | `python/djust/theming/css_generator.py:16` |

All four files form a single strongly-connected component in the import graph:

```
presets.get_preset (lazy) → registry
registry._do_discover (lazy) → presets             ← cycle 1 (2-node)

manager (top-level) → presets
presets (lazy) → registry → manager (lazy)         ← cycle 2 (3-node)

css_generator (top-level) → manager
manager (lazy) → theme_css_generator → css_generator  ← cycle 3 (3-node)

css_generator (top-level) → presets → ... → css_generator  ← cycle 4
```

CodeQL's `py/cyclic-import` flags every module in any cycle (lazy or top-level) — so all four are reported even though only `presets ↔ registry` is a direct two-node cycle.

## The fix

**New leaf module `python/djust/theming/_builtin_presets.py`** holds the 63 `themes/*` PRESET imports + the `THEME_PRESETS` registry dict + `DEFAULT_THEME`. It has no runtime dependency on `presets`, `registry`, `manager`, or `css_generator` — only on `_types` (already a leaf) and `themes/*` (also leaves).

**`registry._do_discover`** now imports `THEME_PRESETS` from `_builtin_presets` instead of `presets`. The `registry → presets` back-edge that closed cycle 1 is gone.

**`manager.py`**: `ThemePreset` now imports from `._types` (leaf); `get_preset` is deferred into `ThemeManager.get_preset()` (its only call site). Top-level edge `manager → presets` is gone.

**`css_generator.py`**: `ThemeTokens` now imports from `._types`; `get_preset` and `get_theme_config` are deferred to call sites. Top-level edges `css_generator → manager` and `css_generator → presets` are gone.

## Back-compat

`presets.py` re-exports the named `*_THEME` constants via `from ._builtin_presets import *` so external code doing `from djust.theming.presets import BLUE_THEME` keeps working. `THEME_PRESETS`, `get_preset`, `list_presets`, and the `ColorScale`/`ThemeTokens`/`SurfaceTreatment`/`ThemePreset` type re-exports remain in their original `presets.py` location. The ~10 internal consumers of `THEME_PRESETS` and the `djust.theming.__init__` public surface need no changes.

`grep -rn "from djust.theming.presets import .*_THEME"` in this repo returns no matches — the `*_THEME` re-export is purely defensive for distributed external code.

## Test plan

- [x] 7 new AST-based regression cases in `python/djust/tests/test_theming_no_cyclic_import.py` pin every broken edge:
  - `_builtin_presets` has no import edges into the SCC
  - `presets` does not top-level-import `registry`
  - `registry` does not import `presets` (lazy or top-level — strictest pin)
  - `manager` does not top-level-import `presets`
  - `css_generator` does not top-level-import `presets` or `manager`
  - `from djust.theming.presets import BLUE_THEME, …` still resolves (back-compat)
  - `get_preset` still walks registry → static dict → DEFAULT_THEME fallback
- [x] Gate-the-fix-off self-test (Action #1200/#1468) — stash each touched file, observe the matching regression test fails, restore. All 3 tests fail on revert (non-tautological).
- [x] Full theming suite: 1831 passed, 1 skipped (was 1824 before, +7 new tests).
- [x] Post-commit gates: gate 1 (no CHANGELOG in impl commit) ✅, gate 2 (docs commit is docs-only) ✅.
- [ ] CodeQL pass on the PR will confirm the four alerts clear post-merge.


johnrtipton:
# Retrospective — PR #1631

Breaks the four-module ``presets/registry/manager/css_generator``
cyclic-import SCC in ``djust.theming``. Clears CodeQL alerts #2352,
#2351, #1900, #1883 (severity: note).

## Quality: 5/5

Leaf-module extraction is the textbook fix for a cyclic-import SCC.
~470 LoC across 6 files (232 in the new leaf, 137 in tests, ~100
across the four touched modules). Two-commit shape clean (impl
``5bd878dd`` + docs ``f17ca170``). Stage 11 reviewer ``STAGE_11_APPROVE``
with 0 findings on first pass, explicit gate-off empirical canary
confirming all 3 fix-edges are non-tautological.

## What went well

- **Symptom-up tracing of the SCC saved an incomplete fix.** The four
  alerts looked like they could be solved by breaking the obvious
  ``presets ↔ registry`` 2-cycle alone, since that was the only direct
  2-node cycle in the graph. But CodeQL ``py/cyclic-import`` flags
  every module in any cycle (lazy or top-level), and the SCC actually
  had three additional cycles routed through ``manager`` and
  ``css_generator``. Tracing each cited file's import edges before
  designing the fix surfaced the full graph — without that step, the
  ``registry → _builtin_presets`` redirect alone would have cleared
  only 2 of 4 alerts and left ``manager``/``css_generator`` still in
  the SCC. ~10-min investigation saved a re-open cycle.

- **AST-based regression tests pin every broken edge, not just smoke
  imports.** Each test parses the target file's module body and walks
  ``ast.ImportFrom`` nodes — testing the *static* import graph, not
  the runtime side-effects. A future contributor who tries to "tidy"
  ``manager.py`` by promoting the deferred ``get_preset`` import back
  to top-level hits ``test_manager_does_not_top_level_import_presets``
  immediately. Same shape as Action #1123 (pre-mount/post-mount keyset
  invariant) — pin the invariant, not just the symptom.

- **Gate-the-fix-off self-test (Action #1200/#1468) was load-bearing.**
  Stashing each touched file and re-running the corresponding
  regression test confirmed three independent failure points — none of
  the new tests are tautological. The Stage 11 reviewer didn't have to
  trust the assertion text; they could verify by reading the test
  source against the gate-off command output.

- **``from ._builtin_presets import *`` is the right back-compat tool**
  for re-exporting a known fixed set of names. 63 ``*_THEME``
  constants would have been noisy as explicit imports in ``presets.py``;
  the wildcard import + populated ``__all__`` in ``_builtin_presets``
  keeps the external API surface stable without forcing a 63-line
  re-export block.

- **Proactive ``ruff format`` on staged files before commit prevented
  the pre-commit-bounce loop** (now 6+ consecutive PRs landing on
  first-commit pass with this discipline).

## Lessons learned

- **Lazy imports still count for CodeQL ``py/cyclic-import``.** Reading
  the CodeQL docs at design time would have made this clear, but my
  initial mental model was that lazy ``from .X import …`` inside a
  function body wouldn't form a graph edge. It does — CodeQL's import
  graph is purely syntactic. The AST regression tests now codify the
  stricter "no presets import at any level" rule for ``registry.py``
  to match.

- **A 4-alert cluster often shares a single root cause.** CodeQL alert
  IDs are per-file, not per-cycle. When the same rule fires on multiple
  related files, look for an SCC before treating each as an
  independent issue. Same lesson as Action #1108 (filter-shape
  Iterable[T] vs list[T]) — the cluster is the signal.

- **``CodeQL alert N`` ≠ ``GitHub issue #N``.** GitHub's ``Closes #N``
  auto-close only resolves issues, not code-scanning alerts. The PR
  body originally used ``Closes #2352`` style; I caught it pre-merge
  and rewrote to a plain reference table. Future PRs that close
  CodeQL alerts should cite the alert ID without the ``Closes``
  keyword to avoid confusion (and to not accidentally close
  unrelated issues that happen to share the number).

## What could be improved

- **CodeQL alerts have no canonical issue artifact.** Unlike user
  reports tied to GitHub issues, a CodeQL alert is just a row in the
  Security tab. There's no place to attach reproducer commands,
  designer notes, or follow-up scope discussion. The CHANGELOG entry
  + the regression test docstring are the only durable records of
  *why* the fix took this shape. Worth a small follow-up:
  ``docs/internal/codeql-alert-triage.md`` capturing the four-cycle
  SCC pattern as a canonical reference for future cyclic-import work.

## IDEA: lines

- IDEA: **Automate the ``import-linter`` contract for ``djust.theming``.**
  This PR added 5 AST tests pinning specific edges. An ``import-linter``
  config could declare the entire dependency layering (``_types`` ←
  ``themes/*`` ← ``_builtin_presets`` ← ``presets``/``registry`` ←
  ``manager``/``css_generator``) declaratively. Future refactors get a
  red squiggle instead of a hand-rolled AST walker.

- IDEA: **Pre-commit hook for new top-level ``from .X import …`` in
  designated leaf modules.** ``_types.py``, ``_builtin_presets.py``,
  ``themes/*`` are leaf modules — adding a top-level import that
  breaches leaf status should be a hook-rejected change. Mechanical
  enforcement of the layering rule.

- IDEA: **Promote ``import-linter`` or a similar dep-layering tool to
  CI for all packages.** Cyclic-import alerts hit ``djust.theming``,
  but the same pattern can hit ``djust.components``,
  ``djust.live_view``, and ``djust.state_backends`` as those grow.
  A workspace-level layering contract would catch the next SCC at PR
  time, not on a CodeQL nightly.

`RETRO_COMPLETE`

PR #1632, opened 2026-05-28
fix(checks): C003 accepts any ASGI server (daphne/uvicorn/hypercorn) (#1630)

### LINKED ISSUE #1630: C003 nags users to install daphne, but djust's canonical ASGI server is now uvicorn
## What's wrong

\`checks.py:667\` raises \`djust.C003\` as an INFO whenever \`daphne\` is missing from \`INSTALLED_APPS\`:

\`\`\`python
elif not has_daphne and not _is_check_suppressed(\"djust.C003\"):
    errors.append(
        DjustInfo(
            \"'daphne' is not in INSTALLED_APPS.\",
            hint=\"Consider adding 'daphne' to INSTALLED_APPS for ASGI support. \"
            \"Suppress this check with DJUST_CONFIG = {'suppress_checks': ['C003']}.\",
            id=\"djust.C003\",
            fix_hint=\"Add \`'daphne'\` to the beginning of INSTALLED_APPS in your Django settings file.\",
        )
    )
\`\`\`

But djust's own canonical recommendation is **uvicorn**, not daphne:

- [README.md:240](https://github.com/djust-org/djust/blob/main/README.md): \"Run with \`uvicorn myproject.asgi:application --reload\`\"
- [README.md:246](https://github.com/djust-org/djust/blob/main/README.md): \"Deployment Guide — production deployment with **uvicorn**, Redis, and Nginx\"
- [djust-org/djust-start](https://github.com/djust-org/djust-start) (the official starter) ships uvicorn in \`pyproject.toml\` and runs \`uvicorn config.asgi:application\` from the Makefile

So every uvicorn-based project — including the one we tell new users to start from — has to ship a permanent \`\"C003\"\` in \`suppress_checks\` to silence a check that's prompting them to install the *non-recommended* server.

## Repro

\`\`\`bash
git clone https://github.com/djust-org/djust-start
cd djust-start && make venv install
# Without C003 in suppress_checks:
.venv/bin/python manage.py check
# → ?: (djust.C003) 'daphne' is not in INSTALLED_APPS.
#   HINT: Consider adding 'daphne' to INSTALLED_APPS for ASGI support.
\`\`\`

The hint is actively wrong guidance for any project following djust's own README.

## Proposed fix

**Option A — broaden the check (preferred):** treat \`uvicorn\` or \`hypercorn\` in the dependency list (importlib.util.find_spec) as equally valid ASGI servers. Only fire C003 when *none* of {daphne, uvicorn, hypercorn} is available. This matches Django's own \`runserver\`/\`runserver_plus\` flexibility.

\`\`\`python
def _has_asgi_server() -> bool:
    if \"daphne\" in installed:
        return True
    import importlib.util
    return any(
        importlib.util.find_spec(name) is not None
        for name in (\"uvicorn\", \"hypercorn\")
    )

if not _has_asgi_server() and not _is_check_suppressed(\"djust.C003\"):
    errors.append(DjustInfo(
        \"No ASGI server detected (daphne, uvicorn, or hypercorn).\",
        hint=\"Install one — uvicorn is recommended: \`pip install uvicorn[standard]\`. \"
             \"Or suppress with DJUST_CONFIG = {'suppress_checks': ['C003']}.\",
        id=\"djust.C003\",
        fix_hint=\"Install uvicorn: \`uv pip install 'uvicorn[standard]'\`.\",
    ))
\`\`\`

The existing \"daphne ordering\" check on lines 654–666 stays as-is (it's correct when daphne *is* present).

**Option B — drop C003 entirely.** It's an INFO, not an error. Django itself doesn't nag about ASGI servers. The check predates uvicorn becoming the canonical pick; with the README change it has outlived its usefulness.

I'd recommend Option A — it's still useful for total beginners who installed djust but no server at all, but it stops misleading the (now larger) uvicorn cohort.

## Impact on djust-start

When this ships, \`djust-start\`'s \`suppress_checks\` drops from \`[\"C003\", \"T002\"]\` to \`[\"T002\"]\` — the starter then has **zero accommodations for framework quirks**.

## Related

Closes the last loose end from the \"reduce djust-start workarounds to zero\" run that filed and shipped #1604, #1605, #1610–#1614, #1617, #1619, #1621, #1624, #1625, #1609, #1628.

johnrtipton:
## Summary

Closes #1630.

C003 previously fired an INFO whenever `daphne` was missing from `INSTALLED_APPS`, telling users to `pip install daphne`. But djust's canonical ASGI server is now `uvicorn` — the [README](https://github.com/djust-org/djust/blob/main/README.md) recommends `uvicorn myproject.asgi:application` and [`djust-org/djust-start`](https://github.com/djust-org/djust-start) (the official starter) ships uvicorn in its `pyproject.toml` and runs `uvicorn config.asgi:application` from the Makefile. Every uvicorn-based project — including the one we tell new users to start from — has to ship a permanent `"C003"` in `suppress_checks` to silence a check that points at the *non-recommended* server.

## The fix (Option A from the issue)

New `_has_asgi_server()` helper in `python/djust/checks.py` probes for any of `{daphne, uvicorn, hypercorn}` via `importlib.util.find_spec` (no actual import — can't ImportError-break `manage.py check`). C003 only fires when none is detected; the hint now points at uvicorn.

```python
def _has_asgi_server() -> bool:
    import importlib.util
    for name in ("daphne", "uvicorn", "hypercorn"):
        try:
            if importlib.util.find_spec(name) is not None:
                return True
        except (ImportError, ValueError):
            continue
    return False
```

The existing daphne-ordering branch (when daphne IS in `INSTALLED_APPS` but listed after `staticfiles`) is unchanged — that's a real misconfiguration.

## Impact on djust-start

When this ships, `djust-start`'s `suppress_checks` drops from `["C003", "T002"]` to `["T002"]` — the starter then has **zero accommodations for framework quirks**.

## Test plan

- [x] 9 new regression cases in `python/tests/test_checks_c003_asgi_servers_1630.py`:
  - uvicorn-installed → C003 silent
  - hypercorn-installed → C003 silent
  - daphne-installed-via-find_spec → C003 silent
  - no-server-at-all → C003 fires with new "No ASGI server detected" wording
  - fix-hint recommends uvicorn
  - daphne-ordering branch still fires (legacy contract preserved)
  - 3 unit tests on `_has_asgi_server` (True/False/exception paths)
- [x] Updated `test_no_suppress_by_default` to stub `_has_asgi_server` (Action #1200 — isolates suppression contract from the broadening; test env's installed uvicorn would otherwise make the test pass for the wrong reason).
- [x] Gate-off self-test (Action #1200/#1468): stashed `python/djust/checks.py`, 8/9 new tests fail on revert (non-tautological). The 1 passing test exercises the unchanged daphne-ordering branch.
- [x] Full pytest suite: **5365 passed, 9 skipped, 0 failed**.
- [x] Post-commit gates: gate 1 (no CHANGELOG in impl commit) ✅, gate 2 (docs commit is docs-only) ✅.

## Related

Closes the last loose end from the "reduce djust-start workarounds to zero" run that shipped #1604, #1605, #1610–#1614, #1617, #1619, #1621, #1624, #1625, #1609, #1628.


johnrtipton:
# Retrospective — PR #1632

Closes #1630. C003 now accepts uvicorn or hypercorn as an ASGI server,
so uvicorn-based projects (including djust-start, the canonical
example) no longer have to permanently suppress the check.

## Quality: 4/5

Surgical fix matching the reporter's Option A. ~45 LoC production, ~167
LoC tests (9 new), two-commit shape clean (impl ``2d43c2c4`` + docs
``f873eb78``). Stage 11 reviewer ``STAGE_11_APPROVE`` with 0 findings,
empirical gate-off check confirming 8/9 tests fail on revert.

The "−1" on quality: I committed the impl commit on ``main`` instead
of the feature branch — the branch-verify reflex (canonized in the
``pipeline-run`` skill text and Action #1144) would have caught this
pre-commit but I didn't run it. Recovered by re-branching at HEAD and
resetting ``main`` to ``origin/main``. No leakage (the commit was not
pushed to ``main``), but it was the exact failure mode the reflex is
designed to prevent, and the fact that the explicit imperative is
sitting in the canon and I still didn't run it is signal worth
recording. Will push the reflex even harder next session.

## What went well

- **Test-env-tautology catch on ``test_no_suppress_by_default``**. The
  pre-existing test asserted "C003 fires by default" by setting
  ``INSTALLED_APPS`` to a daphne-free list. After this PR's broadening,
  the test passes trivially (uvicorn IS installed in the dev/CI env)
  even when the suppression mechanism is broken — classic Action
  #1200 tautology. Catching this required a) running the test, b)
  noticing it FAILED post-fix (correct — the env has uvicorn), c)
  realizing the *fix* was to add ``monkeypatch.setattr(checks,
  "_has_asgi_server", lambda: False)`` so the test isolates the
  suppression contract from the new ASGI-server-presence broadening.
  Stage 11 reviewer flagged this as the right call.

- **Reporter's analysis was load-bearing**. The issue body cited two
  specific options (A: broaden, B: drop) with pros/cons. Option A
  preserves the value (genuine beginners with NO server installed
  still get the nag) while removing the false-positive. Implementing
  Option A verbatim — including the ``find_spec`` probe sketch — meant
  the planning phase took ~5 minutes total.

- **The ``find_spec`` choice over ``importlib.import_module``**. Using
  ``find_spec`` means no side-effect import; a broken/buggy uvicorn
  installation can't crash ``manage.py check``. The reporter
  specifically called this out and it matches the existing pattern in
  Django's own ``check`` infrastructure.

- **9 regression cases pin the contract**. Each test exercises a
  distinct find_spec branch; the negative assertion
  ``"hypercorn" not in c003[0].fix_hint`` is the test-quality detail
  that catches a future PR trying to rewrite the fix_hint to mention
  all three servers (we deliberately recommend uvicorn only). Helper
  unit tests cover True/False/ValueError paths.

- **Two-commit shape + gates clean on first pass**. The implementation
  commit had no CHANGELOG; the docs commit had only CHANGELOG. Pre-push
  hooks all passed first try.

## Lessons learned

- **Branch-verify reflex is not optional even when "I just created
  the branch"**. I created ``fix-1630-c003-asgi-servers`` at the start
  of the pipeline, but somewhere between then and the impl commit I
  ended up back on ``main`` — most likely the ``git stash pop`` during
  the gate-off self-test silently flipped branches when there were no
  conflicts. (Stash doesn't change branch, but if the working tree had
  modifications staged on a different branch when I stashed, the pop
  context can leak.) The pipeline-run skill text has the exact bash
  one-liner for this; running it as a reflex before EVERY commit
  (not just before the first commit) is the rule.

- **Pre-existing tests can become tautological after a behavior
  change**. Action #1200 is usually framed as "did the NEW tests pass
  for the right reason?" — but the same lens applies to pre-existing
  tests after a refactor: did they still test what they claim to test,
  or do they now pass via a different code path? In this case
  ``test_no_suppress_by_default`` was asserting "suppression-default
  contract" but the assertion would now ALSO pass if the suppression
  mechanism were completely broken (because uvicorn is installed in
  the test env). The fix was a one-line monkeypatch but identifying
  the issue required running the test, noticing the failure, and
  thinking "is this test still meaningful?"

## What could be improved

- **Branch-verify reflex needs to be more automatic.** This is the
  second time in two sessions (after the v1.0.0rc3 retro arc, Action
  #1524) where a state-file vs HEAD mismatch was caught
  post-facto by inspection rather than by the pre-commit gate.
  Worth investigating whether the canon-doc rule could be promoted
  to a project-local pre-commit hook (or part of
  ``scripts/pipeline-gates.sh``) so it fires as a check rather than
  as an imperative the executor must remember.

## IDEA: lines

- IDEA: **Programmatic branch-verify gate in ``scripts/pipeline-gates.sh``.**
  Replace the imperative "run this bash one-liner before every
  pipeline commit" with a hook-installed pre-commit check that scans
  ``.pipeline-state/*.json`` for a state file whose ``branch_name``
  matches ``git symbolic-ref --short HEAD``. Refuses the commit if
  the active state file doesn't match HEAD. Same shape as Gate 1 / 2
  for the two-commit shape. Would have caught today's wrong-branch
  commit at the moment it was attempted.

- IDEA: **Audit other system checks for the same uvicorn/daphne
  asymmetry**. C003 is fixed, but if any other check assumes daphne
  (in error messages, in fix_hint suggestions, in T-class template
  scans for daphne-specific patterns) the same issue could be lurking.
  Worth a one-shot grep: ``grep -n daphne python/djust/checks.py
  python/djust/system/checks.py`` and inspect each match for similar
  bias. Found 0 other matches in this PR but worth a periodic
  re-audit as the check surface grows.

- IDEA: **Use the ``_has_asgi_server`` pattern elsewhere**. The
  ``find_spec`` + try/except + ordered-tuple pattern is reusable for
  any check that needs to detect "any of a set of optional deps".
  Future C-class checks (e.g., for cache backends, channel-layer
  backends, session backends) could share a generic
  ``_has_any_installed(*names)`` helper.

`RETRO_COMPLETE`

Issue #1636, opened 2026-05-28
VDOM `{% if %}` non-swap (false→true add, no `{% else %}`) emits InsertChild patch that fails 1/N on client; async-callback state pushes then stop reaching browser (rc13)


johnrtipton:
## Summary

On djust 1.0.0rc13, a non-swap `{% if %}` block that newly **adds** a previously-absent sibling (transitioning from `False → True` — no `{% else %}` branch involved) triggers a client-side `1/15 patches failed (14 succeeded)` failure on the `InsertChild` patch. The framework then requests `html_recovery`, replaces the DOM with the recovered HTML, and **subsequent state mutations from inside `start_async` never reach the client**: the view freezes at the first transitional state even though the backend has long since advanced (DB confirms the async pipeline ran to completion).

Surface symptom from a real app (project management app, `IngestView` runs a multi-second Spark fetch + Claude extraction pipeline via `start_async` and reports live status):

```
[LiveView] 1/15 patches failed (14 succeeded)
[LiveView] VDOM patches failed. This usually happens when {% if %} blocks add/remove DOM elements, shifting sibling positions.
Fix: Use style="display:none" toggling instead of {% if %} for elements:
  <div style="{% if not show %}display:none{% endif %}">...</div>
Requesting recovery HTML from server.
[LiveView] Received: %s %o html_recovery Object
```

UI stays at `Status: fetching` indefinitely. DB row shows `status=routed, review_required=True, meeting_id=...` — the `_run` callback ran successfully and mutated public attrs, but the post-`_run` render never propagates.

This issue is distinct from the cluster `#1358 / #1408 / #1550 / #1552 / #1555` already closed in rc6–rc12:

| Closed issue | Trigger | Our case |
|---|---|---|
| #1550 | `{% if/else %}` branch **swap** (false → true while other branch was visible) | No `{% else %}` — block transitions absent → present |
| #1552 | `{% if/elif %}` swap between `{% include %}` blocks | No `{% include %}` chain |
| #1555 | dj-id counter not bumped past `last_vdom` max in worker thread | We don't use `enable_state_snapshot` so the rc9 LOAD-path gate doesn't apply |
| #1358 / #1408 | Keyed VDOM diff / sync_ids dj-if boundary awareness | The blob below shows the keys ARE present, but the InsertChild patch index still references a slot the client DOM can't bind |

[...]

PR #1639, opened 2026-05-28
fix(websocket): refresh recovery baseline after start_async completion (#1636)

johnrtipton:
## Summary

Fixes the production-impacting half of #1636: after a client VDOM patch fails and `html_recovery` runs, subsequent state pushes from a `start_async` background callback stop reaching the client and the view freezes at the transitional state (e.g. `Status: fetching`) even though the backend pipeline advanced.

## Root cause (reproducer-first)

`LiveViewConsumer._run_async_work` (`python/djust/websocket.py`) re-renders and sends patches when a `start_async` callback completes, but — unlike `handle_event` (`websocket.py:3633`) and `server_push` (`websocket.py:4990`, the #1202 fix) — it never updated `self._recovery_html` / `self._recovery_version`.

`handle_request_html` consumes the recovery baseline and clears it (`_recovery_html = None`, one-time use). So the sequence the reporter hit was:

1. `pick` renders the transitional state; `handle_event` arms `_recovery_html`.
2. A client patch fails → client sends `request_html` → server serves the HTML and sets `_recovery_html = None`.
3. `start_async` `_run` completes → `_run_async_work` renders the advanced state and sends patches **but leaves `_recovery_html` None**.
4. That async patch also fails on the client → the next `request_html` finds `_recovery_html=None` → returns **"Recovery HTML unavailable"** → the client can't recover and freezes.

## Fix

`_run_async_work` now sets `_recovery_html` / `_recovery_version` after `render_with_diff()` in **both** the patches branch and the full-HTML fallback branch, mirroring `handle_event` / `server_push`. Recovery stays armed across patch-failure/recovery cycles, so async-callback state pushes keep reaching the client.

## Tests

`tests/unit/test_async_recovery_html_1636.py` (4 cases, mirror `test_server_push.py::TestServerPushHandler`):
- async render with patches arms the recovery baseline
- baseline refreshed across consecutive async renders
- full-HTML fallback branch arms it too
- **end-to-end continuity**: recovery consumes the baseline → async completes → a subsequent `request_html` serves the advanced HTML instead of the "Recovery HTML unavailable" error (this test reproduces the exact pre-fix error string).

Gate-off self-test (#254): all 4 fail with the two assignments reverted. Full suite: 5100 passed, 0 failed.

## Scope (per scope-discipline)

This PR fixes the reporter's explicitly-stated "most painful" half (async pushes stopping after recovery), which reproduces deterministically. Two findings surfaced during investigation are filed as follow-ups rather than bundled here:

1. **Why the specific client patch fails 1/N in the reporter's full page.** A faithful real-engine reconstruction of the reporter's 5-patch shape (`InsertChild` + the dj-if-marker subtree + the SetAttr/SetText siblings) applies cleanly (0/5 fail) — so the client-side trigger lives in template specifics not present in the `djbug1` blob. Needs a bit-exact public repro (CLAUDE.md #1086/#1389). Follow-up: #1641.
2. **Comment-filter inconsistency**: `getSignificantChildren` counts all comments (`12-vdom-patch.js:1145`) while `getNodeByPath` counts only dj-if comments (`:216`). Real latent index-mismatch bug, but only triggers with a *regular* HTML comment among siblings (Rust VDOM drops non-dj-if comments) — not #1636's trigger. Follow-up: #1640.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- **`start_async` completion now refreshes the VDOM recovery baseline (#1636).** When a client VDOM patch fails (e.g. an `{% if %}` block that adds a sibling), the client requests `html_recovery` and the server serves — then clears — `_recovery_html` (one-time use). `LiveViewConsumer._run_async_work` re-renders and sends patches when a `start_async` background callback completes, but — unlike `handle_event` and `server_push` (the #1202 fix) — it never updated `_recovery_html` / `_recovery_version`. After a recovery had consumed the baseline, an async-triggered patch that also failed on the client found `_recovery_html=None`, got back "Recovery HTML unavailable", and the view froze at the transitional state (e.g. `Status: fetching`) even though the backend pipeline had advanced. `_run_async_work` now sets the recovery baseline after `render_with_diff()` in both the patches branch and the full-HTML fallback branch, mirroring the other two render paths, so async-callback state pushes keep reaching the client across patch-failure/recovery cycles.
+
 ## [1.0.0rc13] - 2026-05-28
 
 ### Added
--- a/python/djust/websocket.py
+++ b/python/djust/websocket.py
@@ -929,6 +929,15 @@ class LiveViewConsumer(AsyncWebsocketConsumer):
 
             if patches is not None:
                 patch_list = fast_json_loads(patches) if patches else []
+                # Refresh the recovery baseline so a later request_html (e.g.
+                # an async-triggered patch that fails on the client) has fresh
+                # HTML to serve. Mirrors handle_event and server_push (#1202).
+                # Without this, an html_recovery that already consumed
+                # _recovery_html leaves it None, the next request_html returns
+                # "Recovery HTML unavailable", and the client freezes at the
+                # transitional state even though the backend advanced (#1636).
+                self._recovery_html = html
+                self._recovery_version = version
                 await self._send_update(
                     patches=patch_list,
                     version=version,
@@ -945,6 +954,10 @@ class LiveViewConsumer(AsyncWebsocketConsumer):
                         ),
                     )
                 )(html)
+                # The fallback sends the full render to the client, so the
+                # recovery baseline must track it too (#1636).
+                self._recovery_html = html
+                self._recovery_version = version
                 await self._send_update(
                     html=html_content,
                     version=version,
--- a/tests/unit/test_async_recovery_html_1636.py
+++ b/tests/unit/test_async_recovery_html_1636.py
@@ -1,0 +1,157 @@
+"""Regression: ``start_async`` completion must refresh the recovery baseline (#1636).
+
+The bug
+-------
+
+When a client VDOM patch fails (e.g. an ``{% if %}`` block that adds a sibling),
+the client sends ``request_html`` and the server serves ``_recovery_html`` —
+then clears it (one-time use, ``handle_request_html`` at ``websocket.py:4520``).
+
+``LiveViewConsumer._run_async_work`` re-renders + sends patches when a
+``start_async`` background callback completes, but — unlike ``handle_event``
+(``websocket.py:3633``) and ``server_push`` (``websocket.py:4990``, the #1202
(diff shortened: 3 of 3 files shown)


johnrtipton (commented):
APPROVE

Stale-base: BEHIND=0. Clean two-commit shape (impl+tests, CHANGELOG separate; #1173).

Correctness: fix mirrors handle_event (websocket.py:3646) and server_push (:5003) — sets `_recovery_html`/`_recovery_version` after `render_with_diff()` in `_run_async_work`. `html is None` in the patches branch is impossible (`render_with_diff` always returns full HTML alongside patches; same invariant the peer paths rely on, documented in the plan). The async else-branch sends full HTML (`html_content`), so setting the baseline there is correct — note it is more than handle_event's full-HTML else (does not set it) and more than server_push's else (no-op/flush). Extra coverage is sound, not a regression.

Tautology gate-off (#1200 / #254): verified empirically — gated both assignment pairs behind `if False:`, all 4 tests failed, including the load-bearing continuity test (real "Recovery HTML unavailable" envelope). Restored clean. No tautologies.

Tests: 4 tests match the plan list exactly. test_async_recovery_html_1636 + test_server_push + test_request_html → 26 passed.

Security/quality: no mark_safe, no f-string logging, no print, no bare except (logger.exception in surrounding handler). CHANGELOG `### Fixed` present. Scope discipline (#1079) followed — client filter mismatch + bit-exact repro deferred to follow-ups.

johnrtipton:
## Stage 14 Retrospective — PR #1639 (#1636)

**Quality: 5/5**

### What went well
- **Reproducer-first discipline (#1210) paid off and overrode the reporter's hypothesis.** The reporter cited "the InsertChild patch fails on the false→true `{% if %}` add." A faithful real-engine reconstruction (`scratch/test_repro_1636.py` → real `client.js` applier in jsdom via `scratch/repro_1636.mjs`) reproduced the reporter's exact 5-patch shape but applied **0/5 failures**. This is the CLAUDE.md PR #1206 lesson in action: the cited mechanism was a hypothesis, and tracing symptom→code (not path→down) surfaced the real, deterministic defect.
- **Two parallel Explore agents** split the client-JS path and the server broadcast-continuity path cleanly; the second found the load-bearing defect: `_run_async_work` never armed `_recovery_html`/`_recovery_version`, unlike `handle_event` (3633) and `server_push` (4990, the #1202 fix whose comment literally describes this freeze).
- **The end-to-end continuity test reproduces the exact pre-fix error string** ("Recovery HTML unavailable"), so it's not a tautology — confirmed by the gate-off self-test (#254): all 4 tests fail with the two assignments reverted, verified independently by the Stage 11 reviewer.
- **Scope discipline (#1079).** Two real findings that were NOT the reporter's trigger were filed as follow-ups (#1640 comment-filter inconsistency, #1641 bit-exact-repro request) rather than bundled, keeping the PR to a 13-line mirror of an established pattern.

### Lessons learned / what could be improved
- **Specific improvement: the recovery-baseline-arming invariant should be centralized.** Three render-send paths now each manually set `_recovery_html`/`_recovery_version` (`handle_event:3633`, `server_push:4990`, `_run_async_work`). This fix is the *third* time the same two lines were added to a render path (#1202 added the second). The pattern keeps recurring because the assignment lives at each call site instead of inside `_send_update` (or a small `_arm_recovery(html, version)` helper). A future render-send path will forget it again. Worth a follow-up refactor: fold the baseline-arming into the send helper so new paths inherit it.
- **Relation to prior tasks in this cluster.** #1636 sits in the #1358/#1408/#1550/#1552/#1555 VDOM-if reopened-issue cluster. The recurring theme across that cluster and this fix: a robustness invariant (dj-id stability, marker-boundary awareness, recovery-baseline arming) is implemented on the main event path but missed on a secondary path (worker thread / broadcast / async). The meta-lesson: when fixing a render-path invariant, grep ALL render-send sites (`render_with_diff` callers) and confirm each carries it — same shape as the symbol-migration grep canon (#1391/#1400).
- **Honest limitation surfaced, not papered over.** The client-side 1/N patch failure root cause remains open (#1641) because it needs a bit-exact repro (CLAUDE.md #1086/#1389). The PR is explicit that it fixes the reporter's stated "most painful" half and that continuity is restored even while the trigger is open (recovery re-arms each cycle, so the UI recovers to correct HTML instead of freezing).

RETRO_COMPLETE

PR #1644, opened 2026-05-28
fix(websocket): flush queued navigation/i18n on every turn-end path (#1643)

### LINKED ISSUE #1643: live_redirect from synchronous event_handler doesn't change browser URL
## Summary

`self.live_redirect("/path/")` called from a synchronous `@event_handler` queues server-side and clears state correctly, but the browser URL never changes. The event handler runs to completion (verified by DB write), but the client stays on the previous page.

## Symptom

In MAX Companion (a djust + djust-mobile-toga app), I have a simple confirm-and-redirect flow:

```python
class EnrollmentConfirmView(LiveView):
    template_name = "medicare/enrollment_confirm.html"

    def mount(self, request, plan_id=None, **kwargs):
        ...

    @event_handler
    def confirm_enrollment(self, plan_id="", **kwargs):
        plan = _find_plan(plan_id)
        if plan is None:
            return
        prefs = _prefs()
        prefs.simulated_part_c_plan_id = plan["id"]
        prefs.simulated_enrollment_date = date.today()
        prefs.save(update_fields=[...])
        self.live_redirect("/enrollment/done/")
```

Template:

```html
<button class="btn" dj-click="confirm_enrollment('{{ plan.id }}')">Confirm enrollment</button>
```

After clicking the button:
- ✅ `prefs.save()` succeeds — `BeneficiaryPreferences.simulated_part_c_plan_id` is `aetna-medicare-choice-ppo` in the DB.
- ❌ The browser URL stays at `/enrollment/confirm/aetna-medicare-choice-ppo/`. No navigation occurs. Title still reads "MAX — Confirm enrollment".

The handler is **not** an `@event_handler async def`. It's a plain synchronous handler.

## What I verified by reading code

- `live_redirect` is the right API — `python/djust/mixins/navigation.py:71-101`. Tests use the same shape (`tests/test_testing_utilities.py:99,116,127`).
- `NavigationMixin` is included in `LiveView` — `python/djust/live_view.py:41,164`.
- `_init_navigation()` is called from `LiveView.__init__` at `live_view.py:473`, so `_pending_navigation` exists when the handler runs.
- `LiveViewConsumer._flush_navigation` exists at `python/djust/websocket.py:695-716` and serializes the queued command as `{type: \"navigation\", action: \"live_redirect\", path: \"/enrollment/done/\", replace: false}`.
- `client.js` dispatches `case 'navigation':` at two switch sites (lines 1409 and 2060) and routes to `window.djust.navigation.handleNavigation(data)` → `handleLiveRedirect(data)` based on `data.action === 'live_redirect'` (lines 9221-9232).

On paper, the chain is wired end-to-end.

## What I did NOT verify (any of these would localize the bug)

- Whether `_flush_navigation` is actually called on the WS turn that completes a synchronous `event_handler`, or only after some other lifecycle step (timing/ordering).
- Whether a WS frame `{type: \"navigation\", ...}` was actually emitted by the server — no `ws_frames` trace.
- Whether `window.djust.navigation` is defined on the client at the time of the click (it should be, but the console-messages tool returned empty — note its capture starts on first call, so that's inconclusive, not \"no errors\").
- Whether `handleLiveRedirect` throws an exception silently in the client.

The shortest path to a real diagnosis is the WS-frame trace — that splits the bug into \"server didn't send\" vs. \"client received and dropped it.\" I stopped before doing that because the demo had a workaround (the home-dashboard banner updates to \"You enrolled in X\" and links to `/coverage/`, so the user reaches their post-enrollment state on the next nav).

## Reproducer

The full flow is in [max-companion PR #28](https://github.com/johnrtipton/max-companion/pull/28)'s follow-up branch (the Open Enrollment walkthrough). Relevant files in that repo:

- `medicare/views.py` — `EnrollmentConfirmView.confirm_enrollment` (the handler)
- `medicare/templates/medicare/enrollment_confirm.html` — the button (`dj-click=\"confirm_enrollment('...')\"`)

I can extract a smaller standalone reproducer if that would help — happy to do so on request.

## Environment

- djust `1.0.0rc13` (pinned in `mobile-app/pyproject.toml`)
- Django 5.2.14, channels 4.3.2, uvicorn 0.47.0
- Browser: Chrome on macOS, talking to a loopback uvicorn dev server
- Templates use the CMS Medicare design system; nothing exotic in markup
- The same app's screening-reminder dismiss flow (also a synchronous `event_handler`) works correctly — that handler doesn't call `live_redirect`, only mutates state and persists.

## Why filing now

I hit this debugging another feature today, didn't finish the investigation, but the symptom is concrete and reproducible. Filing so it's tracked while context is fresh.

johnrtipton:
## Summary

Fixes #1643: `live_redirect()` (and `live_patch()`) from a **synchronous, state-unchanging** `@event_handler` queued server-side but the browser URL never changed.

## Root cause (reproducer-first)

`live_redirect()` queues a navigation command on the view (`_pending_navigation`); the consumer flushes it to the client via `_flush_navigation()`. `handle_event` auto-skips the render when a handler reassigns no public attr (`pre_assigns == post_assigns` → `skip_render`) and returns via `_send_noop` after flushing only `push_events`/`flash`/`page_metadata`/`pending_layout`/`deferred` — **not** `navigation`/`accessibility`/`i18n`.

The reporter's `confirm_enrollment` writes a DB model and calls `live_redirect` but changes **no view attr**, so it hits the skip branch and the queued redirect was dropped.

The deeper cause was **duplication**: the turn-end flush sequence was hand-copied across ~10 sites with inconsistent subsets. The skip branches, **and the `server_push` + `db_notify` broadcast paths**, all dropped `navigation`/`accessibility`/`i18n` — the same latent bug in four more places.

(`LiveViewTestClient.assert_live_redirect` checks the *view's* queue, which is populated correctly — that's why existing tests missed it. The gap was the consumer never flushing it.)

## Fix — DRY consolidation

Introduce one `_flush_all_pending()` helper that flushes all 8 queued side-effects in canonical order, and route every turn-end path through it. Draining is idempotent (each `_flush_*` clears its queue), so paths that also go through `_send_update` are unaffected. **Net −29 lines in `websocket.py`**; closes the same gap on `server_push`/`db_notify` as a side effect.

## Tests

`python/djust/tests/test_live_redirect_skip_render_1643.py` — drives the **real** `handle_event` and asserts at the sent-frame level:
- state-unchanging handler + `live_redirect` → a `navigation` frame is emitted (RED pre-fix: only `noop` was sent; queue left undrained)
- control: state-changing handler still navigates (the `_send_update` path)
- source-pins: every turn-end path (`handle_event`, `_dispatch_single_event`, `server_push`, `db_notify`, `_run_async_work`) uses `_flush_all_pending`, and the helper covers all 8 side-effects

Full suite: 5104 passed, 0 failed.


johnrtipton (commented):
Stage 11 review — APPROVE (posted as comment; cannot self-approve)

**Stale-base (#1450):** BEHIND=0.

**Idempotency (central claim): CONFIRMED.** Every `_flush_*` either drains-and-clears its queue (`_drain_navigation` swaps queue→`[]`; `_drain_push_events` same) or is empty-guarded (`_flush_accessibility` checks non-empty announcements/valid focus tuple; flash/page_metadata/layout/deferred no-op on empty/None). Calling `_flush_all_pending()` twice in one turn — e.g. the async path (line 987) after `_send_update` already flushed at 1144/1165 — sends zero duplicate frames. No double-frame bug.

**Ordering (Step 3):** Helper order (push_events→flash→page_metadata→pending_layout→deferred→navigation→accessibility→i18n) is byte-identical to the pre-fix `_send_update` order. Skip/broadcast branches previously had only the first 5; embedded-child had nav+i18n but not a11y. All now get the canonical superset appended in-order — additions, not reorders. No flush moved from after-send to before-send.

**`_send_update` (Step 4):** Both branches route through `_flush_all_pending`. No path lost a flush.

**Gate-off self-test (#1200/#254):** Commenting out `_flush_navigation` in the helper makes the headline test FAIL (`[noop]` only) and the control FAIL (`[patch]`, no nav). Non-tautological. Restored; 4/4 pass.

**Tests:** target suite 73 passed; broader nav/flush/broadcast/async sweep 405 passed, 6 skipped.

Mount-path flush (line 2467, push_events only) correctly out of scope. CHANGELOG accurate; DRY consolidation nets ~29 fewer lines and closes the same latent gap on server_push/db_notify.

No 🔴/🟡 findings. APPROVE.

johnrtipton:
## Stage 14 Retrospective — PR #1644 (#1643)

**Quality: 5/5**

### What went well
- **Symptom→code tracing (CLAUDE.md triage) landed the real path fast.** The reporter's prime suspect — "is `_flush_navigation` called on the WS turn that completes a sync handler?" — was correct in spirit, and a flush-site survey pinned it: `handle_event`'s auto-skip-render branch (`pre_assigns == post_assigns` → `_send_noop`) flushed 5 of 8 side-effects, dropping the queued `live_redirect`. A handler that writes a DB row and redirects (no view-attr change) hits exactly that branch.
- **Real `handle_event` integration test, not a proxy.** First instinct was a `WebsocketCommunicator` e2e test; it hung on event-turn plumbing, so I pivoted to driving the real `consumer.handle_event` directly with only `send_json` + `_rate_limiter` stubbed. The headline test reproduced the exact symptom (`Frames: ['noop']`, queue undrained) — non-tautological, confirmed by gate-off (#254/#1200).
- **Caught the test's own faithfulness bug.** My first repro handler set `self._side_effect_done = True`, which is a (private) view-attr change that dodged `skip_render` — so the test passed for the wrong reason. Reproducing the reporter's "external side effect only" precisely (a module-level marker, no view attr) restored the RED.

- **The fix-before-feedback shipped a non-DRY patch — and proved the user's point against me.** My first pass hand-added `_flush_navigation` + `_flush_i18n` to the two skip branches. The reviewer-of-record (the user) flagged "this does not seem dry." Surveying revealed the turn-end flush sequence was copy-pasted across ~10 sites with inconsistent subsets — and my own hand-copied fix had **omitted `_flush_accessibility`** (the 8th flush), i.e. I reproduced the exact duplication-drift bug class while fixing it. The DRY redo (one `_flush_all_pending()` helper, net −29 lines) both fixed #1643 and closed the identical latent gap in `server_push` and `db_notify`. **Lesson, left-shiftable: when a fix adds the same 1-2 lines to ≥2 copies of a block, that's the signal to extract the helper in the first pass, not after review.** This is the same "fold the invariant into one helper" insight the #1639 retro flagged for `_recovery_html` — now observed twice; both are "a turn-end invariant scattered across N send paths."

### Relation to prior tasks
- Direct sibling of #1639 (`_recovery_html` arming missed on the async path) and the broader #1358/#1408/#1550/#1552/#1555 cluster: the recurring meta-bug is *a per-turn invariant implemented on the main path and missed on secondary paths* (async / broadcast / skip-render / db-notify). #1644 generalizes the cure: consolidate the whole turn-end side-effect flush onto one helper so no path can drift. A follow-up worth filing: audit `_recovery_html` arming (the #1639 fix) for the same "fold into the send/flush helper" treatment so it can't be missed on a future path either.

RETRO_COMPLETE

PR #1646, opened 2026-05-28
fix(auth): wrap per-event check_object_permission in sync_to_async (#1638)

### LINKED ISSUE #1638: check_object_permission async-wrap asymmetry: mount path uses sync_to_async, per-event path calls sync from async — every URL-bound LiveView with sync ORM in get_object() reports "Access denied" on the first event
## Summary

The v0.9.5-1b per-event object-permission re-check (`_validate_event_security`) calls `check_object_permission` **synchronously from `async def`**, while the mount-time call site **wraps it in `sync_to_async`**. The developer's `get_object()` is documented as sync with an ORM call (`Document.objects.get(pk=self.document_id)`), so any project following the canonical ADR-017 pattern hits `SynchronousOnlyOperation` on the first WS event handler that fires after mount.

The fail-closed catch added in #1379 then translates that exception into a `permission_denied` frame and the user sees:

> djust · server error
> Access denied for this object.
> Triggered by event open_item

…with a Django traceback that points at `django/utils/asyncio.py:24` / `SynchronousOnlyOperation`. So the actual root cause (framework forgot to await/sync_to_async on the per-event path) is invisible to the developer at first glance, and the user sees an auth message for a non-auth bug.

## The asymmetry

**Mount path — wraps correctly** (`djust/websocket.py:2178-2181`):

```python
from .auth.core import check_object_permission

try:
    await sync_to_async(check_object_permission)(self.view_instance, request)
except PermissionDenied as exc:
    ...
```

**Per-event path — calls sync from async** (`djust/websocket_utils.py:258-280`):

```python
else:
    from .auth.core import check_object_permission

    try:
        check_object_permission(owner_instance, owner_request)        # <-- sync call inside async def
    except PermissionDenied:
        await ws.send_error("Access denied for this object.", code="permission_denied")
        return None
    except Exception:  # noqa: BLE001 — fail-closed by design
        logger.exception(
            "Object-permission check raised non-PermissionDenied exception "
            "for %s on event %s; failing closed (denying)",
            owner_instance.__class__.__name__,
            sanitize_for_log(event_name or ""),
        )
        await ws.send_error("Access denied for this object.", code="permission_denied")
        return None
```

`check_object_permission` itself is sync (`djust/auth/core.py:145`) and calls `view_instance.get_object()` sync — by design, since the canonical ADR-017 pattern documents `get_object` as a sync method:

```python
# From the canon / docs / examples:
def get_object(self):
    return Document.objects.get(pk=self.document_id)
```

`Model.objects.get(...)` calls Django's sync ORM, which raises `SynchronousOnlyOperation` when the surrounding frame is inside an event loop without going through `sync_to_async`.

## Symptom from a real app (djust 1.0.0rc13)

`ProjectDashboardView` (URL-bound to `<slug:slug>`) ships a `ProjectAccessMixin` with the canonical pattern:

```python
class ProjectAccessMixin:
    _project_slug: str | None = None

    def get_object(self):
        return Project.objects.filter(name=self._project_slug).first()

    def has_object_permission(self, request, obj):
        return True   # MVP — no real per-object auth yet
```

Mount works fine (sync ORM goes through `sync_to_async` in `websocket.py:2181`). The first WS event handler — `open_item` (a `dj-click` on a table row) — fails with:

```
djust · server error
Access denied for this object.

Triggered by event open_item
Traceback
  File ".../django/db/models/query.py", line 93, in __iter__
    results = compiler.execute_sql(...)
  File ".../django/db/models/sql/compiler.py", line 1621, in execute_sql
    cursor = self.connection.cursor()
  File ".../django/utils/asyncio.py", line 24, in inner
    raise SynchronousOnlyOperation(message)
```

(`event_handler` itself is dispatched through `sync_to_async` per `runtime.py` — verified. But the auth re-check on the *path to* that dispatch isn't.)

`has_object_permission` is **never reached** — the failure is inside `get_object()`'s ORM call. The fail-closed catch then masks it.

## Why this matters more than a "developer should use afirst()" framing

1. **The canon documents `def get_object(self)` as sync with an ORM call.** The audit doc and CHANGELOG examples both show the sync shape. A developer following the docs gets bitten on every URL-bound LiveView with a click-event handler.
2. **The fail-closed catch makes it look like an auth bug**, not a framework bug. The first three projects I bet money the maintainers have helped debug this on will all have rabbited down the auth-config rabbit hole first.
3. **The two call sites are six lines apart in terms of behavior**, and one is wrapped, one isn't. This reads like an oversight, not a deliberate design choice — `sync_to_async` was used everywhere else in `handle_mount` for the same reason it'd be needed here.

## Proposed fix

The minimal symmetric change at `websocket_utils.py:262`:

```python
else:
    from .auth.core import check_object_permission

    try:
        await sync_to_async(check_object_permission)(owner_instance, owner_request)
    except PermissionDenied:
        await ws.send_error("Access denied for this object.", code="permission_denied")
        return None
    except Exception:  # noqa: BLE001 — fail-closed by design
        logger.exception(...)
        await ws.send_error("Access denied for this object.", code="permission_denied")
        return None
```

Both `_validate_event_security` and the surrounding event-dispatch path are already `async def`, and `sync_to_async` is already imported above in this file (`from asgiref.sync import ...`). One-line change, no semantic shift, restores parity with the mount path.

Plus (optional but valuable): differentiate the user-facing message between `PermissionDenied` (auth) and `Exception` (framework or developer-code bug). The current "Access denied for this object" for any catch obscures the bug class. Even just appending `(check raised)` to the `Exception`-branch message would have shaved an hour off the diagnosis.

## Repro

Any `LiveView` that:

1. Overrides `get_object()` returning a queryset/Model lookup using sync ORM (`Model.objects.get(pk=...)`, `.first()`, `.filter(...).first()`).
2. Has at least one `@event_handler()` reachable from a `dj-click` / `dj-submit` / similar WS-dispatched event.
3. Renders correctly (mount succeeds because mount-path wraps in `sync_to_async`).

→ first event click triggers the symptom.

## Versions

- djust: **1.0.0rc13**
- Django: 5.2.x
- Python: 3.12
- Channels: 4.x
- ASGI server: uvicorn

## App-side workaround (what I shipped)

Rename `get_object` → private `_resolve_project()` / `_resolve_meeting()`, drop the `has_object_permission` override. `_has_custom_get_object` walks the MRO and finds nothing → per-event lifecycle no-ops → no async-context ORM call → no `SynchronousOnlyOperation`. Pinned the absence with a regression test so the team can't silently re-add it: [`apps/pm_core/views/auth_mixins.py`](https://github.com/johnrtipton/djust_pm) (project is private; happy to share commit details if useful).

This works for MVP because the would-have-been `has_object_permission` returned `True` anyway. Projects with real per-object auth don't have that out — they need the framework path to actually work.

## Related closed issues

- #1373 — ADR-017 four-layer auth onion (introduces the per-event re-check)
- #1379 — fail-closed-at-every-catch security default (intentional, but masks this bug class)
- #1378 / #1380 / #1381 series — surrounding `check_view_auth` async path

This is an iter-B style follow-up: #1373 wired the lifecycle, the mount path got the `sync_to_async` wrap, the per-event path didn't.

johnrtipton:
## Summary

Fixes #1638 (P0): every URL-bound LiveView following the canonical ADR-017 object-permission pattern reported **"Access denied"** on its *first* WebSocket event.

## Root cause

The per-event object-permission re-check called `check_object_permission` **synchronously from `async def _validate_event_security`** (`websocket_utils.py`), while the mount path wraps the same call in `sync_to_async` (`websocket.py` handle_mount). `check_object_permission` calls the developer's sync `get_object()` — documented as `Model.objects.get(...)` — so the sync ORM ran in the event loop and raised `SynchronousOnlyOperation`. The fail-closed `except Exception` (added in #1379) then mistranslated that into a `permission_denied` frame, so the developer saw an auth error for a non-auth bug.

## Fix

One line: wrap the per-event call in `sync_to_async`, mirroring the mount path. The `PermissionDenied` and fail-closed catches are unchanged.

## Test

`tests/integration/test_object_permission_event.py::test_per_event_sync_orm_get_object_does_not_raise_sync_only` — a view whose `get_object()` does a **real sync ORM read**; the per-event validator must return the handler and send no `permission_denied` frame. RED pre-fix with the exact `SynchronousOnlyOperation`. (Every existing object-permission test used an in-memory `_StubDocument`, so none exercised the sync-ORM path that trips the bug.)

All 21 object-permission (mount + event) tests pass.


johnrtipton (commented):
Stage 11 review — APPROVE.

**Stale-base:** BEHIND=0 vs origin/main. Clean.

**Diff (git diff main...HEAD):** exactly the one-line `sync_to_async` wrap in `_validate_event_security` + the comment block, one new regression test, one CHANGELOG `### Fixed` entry. No drift.

**Correctness:**
- `_validate_event_security` is `async def` (line 156) → `await` valid.
- `sync_to_async` imported (`from asgiref.sync import sync_to_async`, line 13). ✓
- Fail-closed semantics preserved: `sync_to_async` re-raises threadpool exceptions at the `await`, so `except PermissionDenied` (spurious-denial-free) and the `except Exception` fail-closed catch both still fire correctly. ✓
- Mirrors the mount path at `websocket.py:2191` exactly.

**Gate-off (#1200/#254):** reverted the wrap → `test_per_event_sync_orm_get_object_does_not_raise_sync_only` FAILS with `SynchronousOnlyOperation` (real sync ORM `get_object()` in the loop). Restored. Test is non-tautological — it pins the actual fixed path.

**Full run:** `test_object_permission_event.py` + `test_object_permission_mount.py` → 21 passed.

**Sibling-bug note (non-blocking):** `check_handler_permission` is still called bare-sync at `websocket_utils.py:212` from this same async path. Its body short-circuits to True when no `@permission_required`, but for an authenticated non-superuser with `@permission_required`, `user.has_perms()` can hit the DB under the default ModelBackend → same `SynchronousOnlyOperation` class. Narrower latent variant, out of scope for #1638; worth a follow-up issue.

APPROVE.

johnrtipton:
## Stage 14 Retrospective — PR #1646 (#1638, P0)

**Quality: 5/5**

### What went well
- Reporter cited exact file:line for both call sites; tracing confirmed the asymmetry immediately (mount wraps `sync_to_async`, per-event calls bare). One-line fix.
- Reproducer-first paid off precisely: the new test uses a **real sync ORM** `get_object()` and reproduces the exact `SynchronousOnlyOperation` (RED), where all 13 existing object-permission tests used an in-memory `_StubDocument` and never hit the sync-ORM path. That gap is exactly why the bug shipped: the test suite exercised the permission *logic* but not the *async-context* the logic runs in.
- Gate-off (#254) confirmed by the reviewer independently.

- **The same async-context blind spot exists for `check_handler_permission`** (`websocket_utils.py:212`), called bare-sync from the same `async def`. It short-circuits to `True` without `@permission_required`, but an authenticated non-superuser hitting a `@permission_required` handler triggers `user.has_perms()` → DB read under the default `ModelBackend` → the identical `SynchronousOnlyOperation` class. Filed as a follow-up rather than bundled, because the "N sites need N tests" rule (#1104) means it needs its own reproducer (a `@permission_required` handler + DB-backed perm backend) — and #1638's reproducer doesn't cover it.
- **Left-shiftable lesson:** when a test suite for a sync helper exists, add at least one case that drives it through the *real async dispatch path*, not just direct sync calls. The object-permission suite proved its logic but not its execution context; the same shape recurs across auth/ORM helpers.

### Relation to prior tasks
- Same meta-shape as #1639 / #1644 (this session): an invariant correct on the main path, wrong on a secondary path (here: mount vs per-event). The cure differs (#1644 consolidated to one helper; here both call sites already exist, so the fix is to make the per-event one match) but the detection lesson is identical — audit *every* path that calls the shared helper.

RETRO_COMPLETE

Issue #1641, opened 2026-05-28
#1636 follow-up: need bit-exact repro for the client-side 1/N patch failure on false->true {% if %} add


johnrtipton:
Split out from #1636 (PR #1639 fixed the async-recovery-continuity half).

The remaining half of #1636 is the *client-side* "1/15 patches failed" on the `pick` render. A faithful real-engine reconstruction of the reporter's 5-patch shape — `InsertChild` (carrying the nested dj-if marker subtree) + `SetAttr(disabled)` + `SetAttr(value=307250)` + `SetAttr(disabled)` + `SetText(" Ingesting… ")` — applied through the real `client.js` patch applier in jsdom **applies cleanly (0/5 fail)**. With stable dj-ids the InsertChild inserts the new card correctly between the `<!--dj-if-->`/`<!--/dj-if-->` markers.

So the client-side failure depends on template specifics in the reporter's full ~15-patch page that are not captured in the `djbug1` blob — most likely a dj-id `d`-lookup miss that falls back to path traversal landing on a comment, OR an interaction with another adjacent conditional. Per CLAUDE.md #1086/#1389, confirming the root cause needs a bit-exact runnable repro.

**Ask**: a minimal public repro repo (the reporter offered one — "fresh `djust-org/djust-start` scaffold + `ingest.html` + `ingest.py`"). With that we can trace which of the 15 patches actually returns `false` and why.

Note: PR #1639 already restores broadcast continuity after *any* patch failure (recovery stays armed across cycles), so even while this root cause is open, the UI no longer freezes — it recovers to correct full HTML on each failed async patch instead of getting stuck.

johnrtipton:
## Investigation update — diff/apply round-trip is correct for every shape derivable from the issue; bit-exact repo still needed

Ran a reproducer-first sweep with the **real djust engine** (generating real `(baseline_html, patches)` via `render_with_patches`) feeding the **real `client.js` patch applier** in jsdom. Five distinct structural shapes, all derived from #1636's `djbug1` blob + template excerpt:

| Shape | patches | result |
|---|---|---|
| Faithful reconstruction of the 5-patch blob (`InsertChild` carrying the nested dj-if-marker subtree + 2×`SetAttr` + `SetText`) | 5 | **0/5 fail** |
| Sibling `{% if %}` expands *before* the add, with a no-`d` `SetText` after it | 5 | **0/5 fail** |
[...]

PR #1649, opened 2026-05-28
fix(vdom): getSignificantChildren counts only dj-if comments (#1640)

### LINKED ISSUE #1640: VDOM: getSignificantChildren counts ALL comments but getNodeByPath counts only dj-if comments (index mismatch with regular HTML comments)
Split out from #1636 (PR #1639) investigation.

`getSignificantChildren` (`python/djust/static/djust/src/12-vdom-patch.js:1145`) includes **all** comment nodes in its child list, while `getNodeByPath` (`:216`) includes **only** dj-if-family comments (`isDjIfComment`). InsertChild/RemoveChild/MoveChild resolve their `index` via `getSignificantChildren`, while path traversal uses `getNodeByPath`. The two filters disagree whenever a parent has a *regular* HTML comment (e.g. `<!-- section -->`) among its children:

```
parent.innerHTML = '<span>a</span><!-- regular --><span>b</span>'
getSignificantChildren -> 3   (SPAN, CMT, SPAN)
getNodeByPath-style    -> 2   (SPAN, SPAN)
```

The Rust VDOM parser drops non-dj-if comments (`crates/djust_vdom/src/parser.rs:494-499`), so the server's child indices skip regular comments. The client's `getSignificantChildren` counts them, so an index-based InsertChild/RemoveChild lands one slot off when a regular HTML comment sits among siblings.

This is a latent bug, NOT the #1636 trigger (the reporter's blob has only dj-if markers). Fix: make `getSignificantChildren` use the same dj-if-only comment predicate as `getNodeByPath`. Blast radius: InsertChild / RemoveChild / MoveChild index resolution — needs a JS regression test with a regular comment among siblings.

johnrtipton:
## Summary
Fixes #1640. `getSignificantChildren` counted ALL comment nodes, while `getNodeByPath` and the Rust VDOM parser count only `dj-if`-family markers (dropping every other comment). Index-based patches (`InsertChild`/`RemoveChild`/`MoveChild`) resolve their `index` via `getSignificantChildren`, so a plain `<!-- comment -->` among siblings shifted the client index one slot off the server's VDOM indices → wrong node inserted/removed/moved.

## Fix
Align `getSignificantChildren` to the same `isDjIfComment` predicate as `getNodeByPath` (`python/djust/static/djust/src/12-vdom-patch.js`). `dj-if` markers still counted; regular comments dropped.

## Tests
`tests/js/significant_children_comment_filter_1640.test.js` — regular comment excluded (RED pre-fix), `dj-if` markers still counted, and the `getSignificantChildren`↔`getNodeByPath` count invariant pinned. Full JS suite: 1624 passed.

Found during the #1641 investigation; not #1636's trigger (the reporter's blob had only `dj-if` markers) but a real latent index-mismatch bug.


johnrtipton (commented):
Reviewed PR #1649 (branch up-to-date with main, BEHIND=0).

**Verdict: correct, well-scoped, fully tested.**

1. **Diff/consistency:** src/12-vdom-patch.js one-branch change (comment branch returns `isDjIfComment(child.textContent)` not `true`), rebuilt client.js + min/.br/.gz/.map, vitest test, CHANGELOG. Rebuilding from src produced a byte-identical bundle (`git status` clean), and `client.js` contains exactly 2 `isDjIfComment(child.textContent)` sites (getNodeByPath + getSignificantChildren) — bundle is a faithful concat of src.

2. **Blast-radius:** All 6 `getSignificantChildren` callers (InsertSubtree, InsertChild/RemoveChild/MoveChild index-fallbacks, batch-insert) consume server-sent `patch.index`. Server VDOM (parser.rs) drops non-dj-if comments, so aligning the client count is correct. Text-node/element counting is unchanged; textarea value paths operate on text children, unaffected.

3. **Predicate consistency:** Both getNodeByPath and getSignificantChildren now call the same `isDjIfComment` — provably consistent.

4. **Gate-off (#254):** reverted to `return true`, rebuilt, ran the 1640 test → 3 fail (the 3 comment-bearing cases), 1 baseline passes. Restored + rebuilt → 4/4 pass.

5. **Full suite:** 134 files / 1624 pass, 0 fail. insert_remove_child_id, move_child_id, vdom_if_block, batch_insert all green (19/19).

CHANGELOG entry is accurate.

APPROVE

johnrtipton:
## Stage 14 Retrospective — PR #1649 (#1640)

**Quality: 5/5**

### What went well
- Surfaced during the #1641 investigation as a *negative-result side finding* and filed rather than dropped — exactly what the #1641 retro's "rule out, then file the latent gaps" intent was for. It came back as a clean, independently-fixable bug.
- The fix is a true invariant alignment: `getSignificantChildren` and `getNodeByPath` now share the *same* `isDjIfComment` predicate, so index-based and path-based resolution are provably consistent — and both match the Rust parser (`parser.rs` drops non-dj-if comments). One predicate, three consumers.
- Gate-off (#254) confirmed by the reviewer (revert → rebuild → 3 fail).

- **The two predicates were duplicated in spirit, not in code.** `getNodeByPath` (`:216`) and `getSignificantChildren` (`:1145`) each independently decided "which comments count," and they drifted — that drift WAS the bug. They now both call `isDjIfComment`, but nothing structurally prevents a *third* comment-counting site from being added with `return true` again. **Left-shiftable:** a single `isSignificantChild(node)` predicate used by both walkers would make the drift impossible (same shape as the #1644 `_flush_all_pending` consolidation lesson — when two copies of a rule drift, extract the rule). Worth a small follow-up if a third counting site ever appears; not worth it for two.

### Relation to prior tasks
- Third instance this drain of the same meta-bug: *a rule implemented in two places that drifted* (#1638 mount-vs-event wrap; #1644 turn-end flush across N paths; here the comment-count predicate across two DOM walkers). The cure is always "make the two share one source of truth." #1640 did that for the predicate call; the structural consolidation (one `isSignificantChild`) is the next rung if the pattern recurs.

RETRO_COMPLETE

PR #1650, opened 2026-05-28
fix(client): wrap client.js bundle in an IIFE to survive re-execution (#1635)

### LINKED ISSUE #1635: client.js re-evaluation on live_redirect / bfcache causes "_TEXT_INPUT_TYPES has already been declared" SyntaxError
## Symptom

After a `self.live_redirect(...)` (or a bfcache back/forward restore), the browser console shows:

```
client.js:1 Uncaught SyntaxError: Identifier '_TEXT_INPUT_TYPES' has already been declared (at client.js:1:1)
```

The error repeats once per re-evaluation event. The app continues to work (so this isn't a hard breakage) because the `window._djustClientLoaded` guard catches the *runtime* path — but the top-level `const _TEXT_INPUT_TYPES` (and any other top-level `const`/`let` outside the `else` block of that guard) is declared before parse-time has a chance to consult the guard.

## Diagnosis (confirmed)

| Check | Result |
|---|---|
| `<script src="/static/djust/client.js">` in rendered HTML | **1 occurrence** (verified via `curl` and via Chrome network tab) |
| Network GET for `client.js` | **1 request** (verified via Chrome DevTools/network MCP) |
| Script tag location | **Outside `<div dj-root>`**, between `</div>` (dj-root close) and `</body>` |
| `[LiveView] Initializing...` console line | **Multiple times** within the same document, on back-then-forward navigation and on live_redirect (observed in Chrome) |
| `window._djustClientLoaded` runtime guard | Works — second `else` body short-circuits |
| Top-level `const _TEXT_INPUT_TYPES` (line 12605) | **Not guarded** — re-declared at parse time on second eval, hence the SyntaxError |

## Root cause hypothesis

`live_redirect` deliberately keeps the document alive (that's the whole win — no full page reload, view state preserved). When the WS-mount response for the target view is morphed into the existing DOM, any `<script src="...client.js">` tag that's part of that payload gets re-attached and Chrome re-executes its body in the **same** top-level lexical scope. The `else { ... }` block in the double-load guard does scope its inner declarations — but `_TEXT_INPUT_TYPES` and a handful of other top-level identifiers sit outside that block.

Independently, bfcache restore with djust's `pageshow` re-init follows the same path.

## Repro shape

A minimal-ish repro:

1. App with two LiveViews, A → B, where an `@event_handler` on A calls `self.live_redirect(reverse(\"app:b_detail\", ...))`.
2. Mount A. Click the button. View B mounts. Observe the SyntaxError in the console.

In my project (djust_pm, on djust 1.0.0rc13) the path is `ProjectListView.create_project → live_redirect → ProjectDashboardView`. Every traversal triggers a fresh SyntaxError line in the console.

## Suggested fixes (in order of preference)

1. **Wrap *all* top-level declarations in the existing `else { ... }` block** so the runtime guard scope-protects them. Currently the guard fires for the bulk of the file but leaves a handful of top-level `const _<NAME>` declarations exposed.
2. Or: **emit `<script>` tags only in the HTTP shell**, not the WS-mount payload. Morph would then never see them.
3. Or: **mark djust-owned scripts with a stable attribute** (e.g. `data-djust-client=\"v1\"`) and have morph dedupe by that attribute before re-attaching.

Happy to send a PR for #1 if it's the path you want — it's a mostly-mechanical re-indent of the existing top-level decls between the existing `} else {` and `}` of the guard.

## Environment

- djust 1.0.0rc13
- Django 5.2.14
- Python 3.12.9, uvicorn 0.47, channels 4.3
- Chrome on macOS (reproduced in regular and incognito windows)

johnrtipton:
## Summary
Fixes #1635. After `live_redirect` (or a bfcache back/forward restore), the console showed `Uncaught SyntaxError: Identifier '_TEXT_INPUT_TYPES' has already been declared`, once per re-eval.

## Root cause
The bundle is a plain concat of 54 source modules. `live_redirect` keeps the document alive and morphdom can re-attach the `<script src="client.js">`; bfcache `pageshow` re-init does the same. Two classic `<script>` executions share **one** global lexical environment, so top-level `const`/`let`/`class` re-declare at **parse** time on the second execution — before the runtime `window._djustClientLoaded` guard runs (the guard only protects the *runtime* re-init path).

## Fix
Wrap the concatenated bundle in a single IIFE (`scripts/build-client.sh`). Every top-level declaration becomes function-scoped, so a second execution builds a fresh scope and never collides; the runtime guard still short-circuits the re-init. Public API is exported via explicit `window.*` / `globalThis.djust.*` assignments **inside** the IIFE, so nothing observable changes. `initReactCounters` is now explicitly exported on `window.djust` (it was an implicit global the dedup test reached via `window.eval`).

## Test
`tests/js/client_reeval_idempotent_1635.test.js` injects `client.js` as two sequential `<script>` elements (real browser script semantics in jsdom) and asserts no re-declaration `SyntaxError` (RED pre-fix with the exact error) + the guard flag stays set. Full JS suite: **1626 passed**; bundle-init-order lint + eslint green.


johnrtipton (commented):
## Stage 11 review — PR #1650 (IIFE-wrap, #1635)

**Stale-base (#1450):** BEHIND=0 vs origin/main. ✅

**Diff:** matches spec — build-client.sh wrap, src/08 + client.js `window.djust.initReactCounters` export, rebuilt min/br/gz/map artifacts, new repro test, edited dedup test (now reaches internals via `window.djust.*` not `window.eval`), CHANGELOG. ✅

**Implicit-global breakage (key risk):** 🟢 No external bare/window reference to a client.js top-level symbol. Debug panel is a SEPARATE bundle (`src/debug/`, own build step) and reaches client.js only through `window.djust.*` / `globalThis.djust.*`. No Python-rendered inline script nor admin/component/theming template calls a bundle internal as a bare global. The two bare `initReactCounters()` callers (14-init.js, 09-event-binding.js) live INSIDE the same IIFE — same scope, fine. The double-load guard (`00-namespace.js:119`) uses an `if/else {` block that closes at bundle end and parses cleanly inside the IIFE (repro test confirms).

**API still exposed:** ✅ 139 `(window|globalThis).djust.*=` assignments inside the IIFE; repro test asserts `window.djust` truthy + `_djustClientLoaded===true` after two executions.

**Gate-off (#254):** ✅ Removed wrap → plain `cat` → rebuild → repro test FAILS with exactly `SyntaxError: Identifier '_TEXT_INPUT_TYPES' has already been declared`. Test is non-tautological. Restored + rebuilt.

**Full suite:** ✅ 1626/1626 pass (135 files) — event-binding, turbo, SSE, lazy-fill, keyboard-nav all green.

**Determinism:** ✅ `git status` clean after rebuild both before and after gate-off; committed artifacts match src.

No 🔴/🟡. Well-scoped, well-tested, blast radius contained.

APPROVE

johnrtipton:
## Stage 14 Retrospective — PR #1650 (#1635)

**Quality: 5/5**

### What went well
- **Refused to trust the reporter's proposed fix and reproduced first.** The reporter's preferred fix #1 ("wrap all top-level decls in the existing `else { }` block") rested on a mental model of the bundle structure that the empirical repro contradicted: `const`s *inside* a block don't collide across scripts, so if they were really in the else block the bug wouldn't exist. The two-`<script>` jsdom repro (not `eval`, which has different scope semantics) reproduced the exact `_TEXT_INPUT_TYPES` SyntaxError and proved the declarations leak to the *global* lexical scope — pointing at the IIFE wrap as the only complete fix (covers both live_redirect morph AND bfcache, regardless of where the internal `else` accidentally closes).
- The wrap is one build-script change with a large blast radius, gated correctly: the full 1626-test JS suite is the safety net, and it caught the one real regression (the dedup test reaching `initReactCounters` via bare `window.eval`). Fixing that the right way — an explicit `window.djust.initReactCounters` export — turned an implicit-global accident into intentional API.

- **`eval`-based reproduction gave a false negative first.** My initial repro used `window.eval(code)` twice and it PASSED (eval'd `const` scopes to the eval, not the global lexical env) — which would have wrongly concluded "not reproducible." Only switching to real `<script>`-element injection reproduced it. **Left-shiftable rule:** to reproduce a *classic-script re-execution* bug, inject `<script>` elements — never `eval` — because `eval` and `<script>` have materially different lexical-scoping semantics for top-level `const`/`let`. Worth a note in the JS-test harness docs.
- **The IIFE wrap exposed a latent fragility:** the bundle's double-load `if/else` guard spans modules via an unclosed brace in `00-namespace.js` closed by concatenation — a fragile accident. The IIFE makes re-eval safe regardless, but a follow-up to make the guard a clean `if (window._djustClientLoaded) return;` at the top of the IIFE (now that early-return is possible inside the function) would remove the cross-module brace dependency entirely. Not done here to keep the diff to the minimal fix.

### Relation to prior tasks
- Same drain meta-shape as #1638/#1640/#1644: a correctness property (here, scope isolation) that was implicit/accidental rather than structural. The cure is to make it structural — the IIFE — and to make external access explicit (the `initReactCounters` export), mirroring the "one source of truth" theme running through this whole drain.

RETRO_COMPLETE

PR #1651, opened 2026-05-29
fix(scaffold): djust new runs makemigrations + ships migrations/__init__.py (#1637)

### LINKED ISSUE #1637: djust new: demo-app tables missing on deploy (run-syncdb instead of makemigrations; no migrations/__init__.py)
## Summary

A `djust new` project's demo-app tables (`ReactionCount`, `PollVote`, `GuestbookMessage` in `apps/home`) are **not created in the deployed rootfs**, so the home page 500s on first request once deployed to djustlive (microvm backend). The app boots and uvicorn serves `config.asgi:application` correctly — the failure is purely the missing tables.

Discovered while bringing up **djust-start** on djustlive (see djustlive PR #399). Live evidence:

```
INFO: Uvicorn running on http://0.0.0.0:8000
INFO: 192.168.1.1 - "GET / HTTP/1.1" 500 Internal Server Error   # ReactionCount.objects... → no such table
INFO: 192.168.1.1 - "GET /favicon.ico HTTP/1.1" 404 Not Found    # routing works; only DB query fails
```

Built rootfs `db.sqlite3` contained only Django built-in tables — no `home_*`:
```
['auth_*', 'django_content_type', 'django_migrations', 'django_session', 'sqlite_sequence']
```

## Root cause (two contributing factors)

**1. Generator uses `migrate --run-syncdb`, never `makemigrations`.**
`python/djust/scaffolding/generator.py:661`:
```python
_run_cmd([venv_python, "manage.py", "migrate", "--run-syncdb"], cwd=project_dir)
```
`--run-syncdb` creates tables for apps **without** migrations directly in the dev DB — so it *masks* the absence of migrations on the developer's machine. But the generator never runs `makemigrations`, so `apps/home` ships with **no migrations**. When the project is deployed, the server-side rootfs build runs `manage.py migrate --noinput` **without** `--run-syncdb`, and an app with no migrations gets no tables → 500 on first query.

**2. When a migration does exist, `migrations/__init__.py` is missing.**
The deployed djust-start tarball had `apps/home/migrations/0001_initial.py` (hand/tool-generated post-`djust new`) but **no `apps/home/migrations/__init__.py`**. Django's migration loader treats a migrations dir without `__init__.py` as a namespace package → marks the app unmigrated → skips it. (Other `__init__.py` files — `config/`, `apps/`, `apps/home/` — were all present; only the migrations package's was absent.)

The deploy tarball walker (`deploy_cli.py:_create_tarball`) includes whatever is on disk and does not exclude empty files, so the `__init__.py` was simply never on disk.

## Suggested fix

Make `djust new` produce a **reproducible-on-any-target** schema:
- Run `makemigrations` (which creates `apps/<app>/migrations/__init__.py` + `0001_initial.py`) **before** `migrate`, and drop `--run-syncdb` — or at minimum, in addition to it.
- Ensure every generated app with models ships a `migrations/__init__.py`.

This guarantees the deploy-time `migrate` (no `--run-syncdb`) creates the demo tables on the target.

## Workaround applied for djust-start

Added `apps/home/migrations/__init__.py`, rebuilt the rootfs, redeployed. Verified the public URL now serves the real app (HTTP 200, 80 KB, `<title>djust starter</title>`) instead of the 500. djustlive PR #399 (modern-scaffold microvm boot) is the infra side; this issue is the orthogonal scaffold-generator gap.

## Repro

1. `djust new myapp`
2. Deploy to a target that runs `manage.py migrate` without `--run-syncdb` (i.e. any non-dev deploy)
3. `GET /` → 500 (`no such table: home_reactioncount`)

johnrtipton:
## Summary
Fixes #1637. A `djust new --with-db` project's tables were missing on deploy (`GET / → 500: no such table`), because the scaffold never generated migrations.

## Root cause
`generate_project`'s `_run_auto_setup` ran `migrate --run-syncdb` but never `makemigrations`. `--run-syncdb` creates tables for migration-less apps directly in the **dev** DB — masking the absence of migrations — but a deploy runs `migrate` **without** `--run-syncdb`, so a `--with-db` app (which shipped zero migrations) gets no tables. Secondarily, a migrations dir without `__init__.py` is a namespace package the migration loader skips.

(The issue's `apps/home` is the separate djust-start repo; this PR fixes the core generator that `djust new` invokes.)

## Fix
1. `_run_auto_setup` runs `makemigrations` **before** `migrate` — generating `<app>/migrations/__init__.py` + `0001_initial.py` on the dev machine, which ship in the deploy artifact so deploy-time `migrate` creates the tables. `--run-syncdb` kept as belt-and-suspenders.
2. `_create_project_files` explicitly writes `<app>/migrations/__init__.py` for model-bearing apps, so the package exists even with `--no-install` (`auto_setup=False`).

## Tests
`python/djust/tests/test_scaffold_migrations_1637.py` — `--with-db` scaffold ships `migrations/__init__.py` (RED pre-fix); auto-setup runs `makemigrations` before `migrate` (RED pre-fix); no-db control writes no models. Scaffold/cli sweep: 232 passed.


johnrtipton (commented):
Stage 11 review (base-fresh: BEHIND=0 vs origin/main).

**Diff scope** — Clean. Exactly the two cited changes + new regression test + CHANGELOG `### Fixed` entry. No drift.

**Correctness**
1. `makemigrations` with no app-label processes all apps; contrib apps are already migrated (no-op), the one generated app produces only `0001_initial`. Fresh scaffold → no rename/merge ambiguity → no interactive prompts → cannot hang. No pending changes elsewhere → prints "No changes detected", exits 0. Confirmed safe.
2. Shipping an empty `migrations/__init__.py` then running `makemigrations` is fine — Django happily reuses an existing empty package and writes `0001_initial.py` alongside.
3. `if ctx["with_db"] or schema:` guard covers the schema path; `from_schema` already forces `with_db=True` (generator.py:72), so `or schema` is harmless-redundant but correct.

**Gate-off (#254)** — both verified:
- Reverted makemigrations line → `test_auto_setup_runs_makemigrations_before_migrate` FAILS.
- Reverted `migrations/__init__.py` write → `test_with_db_scaffold_ships_migrations_init` FAILS.
- Restored; 3/3 pass.

**Suite** — `test_scaffold_migrations_1637.py` 3/3 pass. Scaffold sweep (`-k "scaffold or generat or cli or djust_new"`) 254 passed, 0 failed.

APPROVE

johnrtipton:
## Stage 14 Retrospective — PR #1651 (#1637)

**Quality: 5/5**

### What went well
- Reporter's diagnosis was precise (file:line + the `--run-syncdb`-masks-the-gap mechanism), so the fix was fast. But I verified the load-bearing claim independently rather than trusting it: confirmed `djust new` → `generate_project` (not the djust-start repo's `apps/home`, which is a separate repo), so the PR stayed scoped to the core generator.
- Two-part fix matches the two contributing factors: `makemigrations` (so migrations exist + ship) AND an explicit `migrations/__init__.py` (so the dir is a real package even with `--no-install`). Each has its own RED test; the reviewer confirmed both gate-offs.
- Kept `--run-syncdb` rather than dropping it (the issue offered "or in addition to") — lowest-risk: migration-less third-party apps still get tables, and the real fix (makemigrations) makes deploy-parity `migrate` work regardless.

- **The dev-vs-deploy command asymmetry was the trap, and it's untested at the integration level.** The dev path (`migrate --run-syncdb`) and the deploy path (`migrate`, no flag) are *different programs*, and the scaffold only ever exercised the dev one — so the masking went unnoticed until a real deploy. This is the same shape as #1638 (mount-vs-event) and #1644 (main-vs-secondary path): **a behavior verified on path A, broken on path B.** Left-shiftable: the scaffold's own test matrix should run the generated project's migrations the way a *deploy* does (plain `migrate` on a fresh DB), not just assert the command list. A full "generate → plain-migrate → query a demo model" integration test would have caught this at authoring time. Filed mentally as the stronger follow-up; the command-sequence + package-shipping tests here are the pragmatic version.

### Relation to prior tasks
- Fourth instance this drain of "verified on path A, broken on path B." The drain's through-line is unmistakable: djust's bugs cluster around *parallel code paths that drifted* (sync/async, main/broadcast, two DOM walkers, dev/deploy). The durable cure each time is to make the paths share one source of truth or to test BOTH paths — not just the convenient one.

RETRO_COMPLETE

Issue #1312, opened 2026-05-02
tech-debt: elevate single-script-transformation pattern to canon for bulk renames


johnrtipton:
**Source**: Retro v0.9.2-5 / PR #1293 (data_table 23-site rename).

## Pattern

PR #1293 renamed 23 emit-name strings across 4 files. The implementer used a single Python script (`/tmp/fix-data-table-event-names.py`) that read each file, applied targeted regex substitutions, and wrote back — atomic against the working tree. Zero regressions, all tests passed first try, no partial-state windows.

Action #180 (v0.9.1 single-implementer-per-checkout canon) listed the single-script-transformation pattern as a "safe alternative" to incremental Edit-tool calls when working in parallel agents. PR #1293 used it for a sequential single-implementer rename and demonstrated it's the right shape regardless of agent count: incremental edits leave the working tree in inconsistent intermediate states (file 3 of 23 done, files 4-23 still old) where pre-commit hooks see partial renames.

## Why elevate from "safe alternative" to canon

For any "rename across the codebase" PR (>5 sites, multiple files), incremental Edit-tool calls have these failure modes:
1. **Partial-state intermediate**: pre-commit hooks see mid-rename code; tests may fail on inconsistent state.
2. **Agent context cost**: 23 Edit calls vs 1 script + 1 invocation.
3. **Reviewer cognitive load**: 23 hunks of "rename foo to bar" vs 1 script + 1 diff to read.
4. **Verification cost**: have to grep post-rename to check completeness; the script self-documents what it touched.

Single-script transformation defaults are the right shape. Worth elevating in CLAUDE.md.

## Scope

Update CLAUDE.md "Process canonicalizations" section: add a "Bulk renames use single-script transformation" rule under Stage 5 (Implementation):

> When a PR renames a symbol/string across >5 sites or multiple files, use a single Python (or shell) script that does the entire transformation in one pass + immediately stages the changes. Do NOT use incremental Edit-tool calls for bulk renames — they leave intermediate states where pre-commit hooks may see inconsistent code, increase reviewer cognitive load, and burn agent context. The script doubles as documentation of what was changed.

Action #180 already addresses parallel-agent safety; this elevation extends to single-agent bulk operations.

johnrtipton:
Resolved by PR #1336 (single-script-transformation canon added to CLAUDE.md).

PR #1652, opened 2026-05-29
refactor(websocket): consolidate recovery-baseline arming into _arm_recovery (#1645)

### LINKED ISSUE #1645: tech-debt: fold per-turn invariants (_recovery_html arming) into the send/flush helpers so no WS path can drift
From the #1644 (#1643) retro. Two recent bugs share one shape: a **per-turn invariant implemented on the main path and missed on a secondary path**.

- #1639: `_recovery_html` / `_recovery_version` arming was in `handle_event` and `server_push` but missing in `_run_async_work` (async completion) → recovery froze.
- #1643: the turn-end side-effect flush (navigation/accessibility/i18n) was missing on the skip-render, `server_push`, and `db_notify` paths → `live_redirect` dropped.

#1644 cured the flush case structurally with one `_flush_all_pending()` helper used by every turn-end path. The recovery-baseline arming (#1639) is still done by a hand-written 2-line assignment at each send site (`handle_event`, `server_push`, `_run_async_work`) — the same drift risk remains for the next send path added.

**Ask**: fold `_recovery_html`/`_recovery_version` arming into the send helper (e.g. set it inside `_send_update`, or a small `_arm_recovery(html, version)` called alongside `_flush_all_pending`) so a future turn-end path inherits it automatically. Audit all `render_with_diff` → send sites for any other per-turn invariant with the same scatter (push-event flush callback wiring, `_changed_keys`, etc.).

johnrtipton:
## Summary
Fixes #1645 (tech-debt follow-up from the #1644 retro). `_recovery_html` / `_recovery_version` were armed by a hand-copied two-line assignment at every render-send path (`handle_event`, `server_push`, `_run_async_work`). That duplication is exactly the drift that caused #1639 — the `start_async` path was added *without* the arming, freezing the view after a `request_html` recovery.

## Fix
Introduce a single `_arm_recovery(html, version)` helper as the one source of truth; every send path routes through it. The only other writer is the one-time clear (`= None`) in `handle_request_html`. **Pure refactor — no behavior change.**

## Tests
`tests/unit/test_arm_recovery_helper_1645.py`:
- helper sets both fields
- a regex guard asserting NO method other than `_arm_recovery` (rhs `html`) or the clear (`= None`) assigns `_recovery_html` — so a future send path can't arm inconsistently
- each send path routes through `_arm_recovery` and no longer hand-assigns (RED pre-refactor)

Existing recovery/navigation behavior tests (#1636, #1643, `server_push`) stay green; broad websocket/recovery/async sweep: 506 passed.

No CHANGELOG entry — internal refactor, no behavior change (`feat:`/`fix:` only per convention).

### DIFF
--- a/python/djust/websocket.py
+++ b/python/djust/websocket.py
@@ -955,8 +955,7 @@ class LiveViewConsumer(AsyncWebsocketConsumer):
                 # _recovery_html leaves it None, the next request_html returns
                 # "Recovery HTML unavailable", and the client freezes at the
                 # transitional state even though the backend advanced (#1636).
-                self._recovery_html = html
-                self._recovery_version = version
+                self._arm_recovery(html, version)
                 await self._send_update(
                     patches=patch_list,
                     version=version,
@@ -975,8 +974,7 @@ class LiveViewConsumer(AsyncWebsocketConsumer):
                 )(html)
                 # The fallback sends the full render to the client, so the
                 # recovery baseline must track it too (#1636).
-                self._recovery_html = html
-                self._recovery_version = version
+                self._arm_recovery(html, version)
                 await self._send_update(
                     html=html_content,
                     version=version,
@@ -1037,6 +1035,21 @@ class LiveViewConsumer(AsyncWebsocketConsumer):
                     logger.exception(
                         "[djust] Error in handle_async_result for task '%s'", task_name
                     )
+
+    def _arm_recovery(self, html: str, version: int) -> None:
+        """Arm the on-demand VDOM recovery baseline.
+
+        Single source of truth for the ``request_html`` recovery state
+        (``_recovery_html`` / ``_recovery_version``). Every render-send path —
+        ``handle_event``, ``server_push``, ``_run_async_work`` — calls this after
+        rendering so the baseline can never drift between paths. Hand-copying the
+        two-line assignment is exactly how the async path was missed in #1639;
+        centralizing it here (#1645) makes a new send path inherit correct arming
+        by calling one method. The one-time clear (``_recovery_html = None`` in
+        ``handle_request_html``) is the only other writer.
+        """
+        self._recovery_html = html
+        self._recovery_version = version
 
     async def _send_update(
         self,
@@ -3630,8 +3643,7 @@ class LiveViewConsumer(AsyncWebsocketConsumer):
                         # Store rendered HTML for on-demand recovery.
                         # Client sends request_html when applyPatches() fails
                         # (e.g., {% if %} blocks shifting DOM structure).
-                        self._recovery_html = html
-                        self._recovery_version = version
+                        self._arm_recovery(html, version)
 
                         await self._send_update(
                             patches=patch_list,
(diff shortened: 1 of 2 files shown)


johnrtipton (commented):
Reviewed as a pure refactor (#1645). Branch is 0 commits behind origin/main (no stale-base risk per #1450).

**Scope** — diff is exactly the helper + 4 call-site replacements in `websocket.py` + one new test file. Nothing else touched.

**Behavior preservation (1:1)** — verified each replaced site previously held the exact `_recovery_html = html; _recovery_version = version` pair, and `_arm_recovery(html, version)` is invoked at the same point (immediately before `_send_update`) with the same `html`/`version` in scope:
- `handle_event` (3646), `server_push` (4998), `_run_async_work` ×2 (958, 977).
No site lost arming or changed *when* it arms.

**Bare-assignment grep** — exactly 2 remaining `self._recovery_html =`: the helper rhs `html` (1051) and the one-time `= None` clear in `handle_request_html` (4532). As expected.

**Gate-off (#254)** — reverted the `handle_event` site to hand-assignment; `test_render_send_paths_route_through_arm_recovery` FAILED as expected, then restored clean. Not tautological.

**Regression guard** — 28/28 pass across the helper test + #1636/#1643/server_push behavior suites. No behavior change confirmed.

APPROVE

johnrtipton:
## Stage 14 Retrospective — PR #1652 (#1645)

**Quality: 5/5**

### What went well
- This PR is the *direct* payoff of the #1644 retro's "fold the invariant into one helper" lesson — filed as a follow-up there, drained here. The recovery-arming was the last hand-copied per-turn invariant in the consumer; it's now a single `_arm_recovery()` source of truth, closing the exact drift that froze the view in #1639.
- The load-bearing test isn't just "the helper works" — it's a **regex guard** (`test_arm_recovery_is_the_only_arming_mechanism`) that fails if ANY future method hand-assigns `_recovery_html`. That converts a convention ("always arm recovery on a send path") into a mechanically-enforced invariant, which is strictly stronger than the per-site call-pin. A new send path that forgets to arm will now trip a test, not ship a freeze.
- Single-script transformation (#1312) for the 4-site replacement; reviewer confirmed 1:1 behavior preservation + gate-off.

- **The guard test pins the writer but not the *caller* completeness.** It guarantees `_recovery_html` is only written via the helper, but it does NOT guarantee every render-send path actually *calls* the helper — a new path could render+send and simply never arm (the #1639 shape again). The companion `test_render_send_paths_route_through_arm_recovery` enumerates the three known paths by name, but that list is itself hand-maintained. Left-shiftable: a count-based test (à la Action #1125) that enumerates all `_send_update`/`_send_noop` call sites and asserts each is preceded by an `_arm_recovery` (or is a deliberate no-arm) would close the caller-side gap. Deferred — not worth the AST machinery for 3 sites today, but the right move if a 4th send path appears.

### Relation to prior tasks
- Caps the session's through-line (#1638/#1640/#1644/#1637/#1635): every fix made a drifted-parallel-path invariant *structural*. This one turned "remember to arm recovery" from tribal knowledge into a one-function contract + a regex guard — the most durable form of the cure.

RETRO_COMPLETE

PR #1653, opened 2026-05-29
test(infra): assert_http_ws_djid_parity harness (#1642)

### LINKED ISSUE #1642: test-infra: HTTP-GET-render → WS-mount-render dj-id parity harness (from #1641)
From the #1641 investigation (PR #1639 arc).

Every VDOM-if reproducer we have uses `LiveViewTestClient.render_with_patches()` for **both** the baseline and the transition. That does NOT exercise the real initial-page path: in production Django's HTTP GET renders the DOM the client actually holds, then the WebSocket connects and `render_with_diff`'s first call establishes a **separate** baseline. If those two renders ever assign divergent `dj-id`s, the first event's patches would miss `d`-resolution and fall back to path traversal — precisely the `getNodeByPath → null → patch fails 1/N` shape #1636/#1641 describe, and the one shape we could not reproduce framework-side.

**Ask**: add a test helper / fixture that captures the HTTP-GET-rendered HTML and the WS-mount frame for the same view and asserts their `dj-id` assignments match (or documents where they intentionally differ). Candidate shapes:
- `LiveViewTestClient.assert_http_ws_djid_parity()`, or
- a `WebsocketCommunicator`-based fixture that grabs the mount frame and diffs its `dj-id`s against the GET HTML.

This would let the next investigator test the HTTP-vs-WS divergence hypothesis without needing a downstream consumer's private repo.

johnrtipton:
## Summary
Closes #1642 (test-infra follow-up from the #1641 investigation). Adds `LiveViewTestClient.assert_http_ws_djid_parity()` so the HTTP-GET-vs-WebSocket-mount `dj-id` divergence hypothesis is testable framework-side, without a downstream consumer's private repo.

## Why
Every VDOM-if reproducer used `render_with_patches()` for both baseline and transition, never exercising the real initial-page path: HTTP `GET` renders the browser DOM and establishes a baseline, while a **separate** instance handles the WS mount and establishes its own. If the two baselines assign divergent `dj-id`s, the first WS event's patches miss `d`-resolution and fall back to path traversal — the `getNodeByPath → null` failure shape of #1636/#1641.

## What
`assert_http_ws_djid_parity(**mount_kwargs)` builds two independent instances:
- HTTP-shaped: `render_full_template` **then** `render_with_diff` (exercises the real initial-page path before the baseline)
- WS-shaped: `render_with_diff` only

and asserts an identical ordered `dj-id` sequence in the `dj-root` subtree. The differential — the HTTP side does extra rendering before its baseline — is the real check: it catches an initial-page render that perturbs the baseline relative to a clean WS mount. Locks the #1370 "marker IDs match between initial HTTP DOM and WS diffs" invariant.

## Finding
Parity **holds** across simple / if-add (#1641 shape) / nested-if-elif views — so #1641's failure is **not** a framework-side HTTP/WS `dj-id` divergence for these shapes. The harness lets the next investigator drop a suspect view in for a definitive answer.

Tests: `python/djust/tests/test_http_ws_djid_parity_1642.py` (3 parametrized parity + 1 extraction unit). Documented in `docs/website/guides/testing.md`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- **`LiveViewTestClient.assert_http_ws_djid_parity()` — test-infra harness for HTTP-GET vs WebSocket-mount `dj-id` parity (#1642).** Builds two independent instances of a view (one exercising the HTTP initial-page path via `render_full_template` then `render_with_diff`; one the WebSocket-mount baseline via `render_with_diff` only) and asserts they assign an identical `dj-id` baseline in the `dj-root` subtree. A divergence is the `getNodeByPath → null` first-event patch-miss shape investigated in #1641; this locks the #1370 "marker IDs match between the initial HTTP DOM and subsequent WS diffs" invariant against regression and lets the divergence hypothesis be tested framework-side. See `docs/website/guides/testing.md`.
+
 ### Fixed
 
 - **`djust new --with-db` projects now create their tables on deploy (#1637).** The scaffold ran `migrate --run-syncdb` but never `makemigrations`, so a `--with-db` app shipped with no migrations. `--run-syncdb` created the tables on the developer's machine (masking the gap), but a deploy runs `migrate` *without* `--run-syncdb`, so the app's tables were never created and the first request 500'd (`no such table: <app>_<model>`). `djust new` now runs `makemigrations` before `migrate` (generating the app's `migrations/__init__.py` + `0001_initial.py`, which ship in the deploy artifact), and explicitly writes `<app>/migrations/__init__.py` for model-bearing apps so the migrations directory is a real package — not a namespace package the migration loader silently skips — even when `--no-install` is used.
--- a/docs/website/guides/testing.md
+++ b/docs/website/guides/testing.md
@@ -162,6 +162,27 @@ client.trigger_info({"type": "db_notify", "channel": "orders", "payload": {"even
 client.assert_state(count=1)
 ```
 
+### `assert_http_ws_djid_parity(**mount_kwargs)`
+
+Asserts that the HTTP-GET render path and the WebSocket-mount render path assign
+the same `dj-id` baseline for a view. In production the initial HTTP `GET`
+renders the DOM the browser holds and establishes a VDOM baseline, while a
+*separate* instance handles the WebSocket mount and establishes its own
+baseline; the first WS event ships patches keyed by `dj-id`. If the two
+baselines diverge, those patches miss `dj-id` resolution and fall back to path
+traversal — a `getNodeByPath → null` patch failure.
+
+```python
+def test_dashboard_djid_parity():
+    LiveViewTestClient(ProjectDashboardView).assert_http_ws_djid_parity(slug="acme")
+```
+
+Reach for this when a view renders differently on first paint vs. reconnect, or
+when you suspect a `1/N patches failed` console warning traces to an HTTP-vs-WS
+`dj-id` mismatch. The harness builds two independent instances and compares the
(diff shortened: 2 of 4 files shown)


johnrtipton (commented):
Stage 11 review — test-infra harness (#1642).

**Stale-base (#1450):** BEHIND=0 vs origin/main. OK.

**1. Scope:** Clean. testing.py helper (+77), test file (+94), CHANGELOG `### Added`, testing.md doc. No production-code drift. ✅

**2. Tautology check — the central concern:** NOT tautological. ✅ Verified the differential is real: `render_full_template` (template.py:815-817) calls `_initialize_rust_view` → `_sync_state_to_rust` → `_rust_view.render()`, advancing that instance’s Rust diff baseline. The follow-up `render_with_diff` on the HTTP instance re-enters `_initialize_rust_view`, which is guarded `if self._rust_view is None` (rust_bridge.py:271) — a no-op — so the already-`render()`-ed view persists. The WS instance’s `render_with_diff` is its FIRST render. The two instances therefore traverse genuinely different code paths; a #1370-class regression where the initial-page render perturbs dj-id assignment would diverge and trip the assert.

**3. Tests:** `pytest test_http_ws_djid_parity_1642.py -q` → 4 passed. ✅

**4. `_djroot_djids` scoping:** Correct — anchors subtree from the `dj-root` match start; the unit test pins a shell `dj-id="ignored-shell"` is excluded. ✅

**5. Docstring/CHANGELOG honesty:** Honest. Both explicitly state initial server HTML carries no dj-root dj-ids and that the pinned invariant is diff-time assignment parity, not server-emitted initial ids. ✅

Minor (non-blocking): all cases currently pass, so the harness has not yet demonstrated a *failing* divergence — its catch-power rests on the path-differential argument above, which holds. Consider a future negative fixture (a deliberately divergent view) to lock catch-power empirically (Action #1468 spirit).

APPROVE

johnrtipton:
## Stage 14 Retrospective — PR #1653 (#1642)

**Quality: 4/5**

### What went well
- Closed the loop the #1641 retro opened: the "we couldn't reproduce framework-side" investigation produced a *durable* artifact (a reusable parity harness) instead of just a closed issue. The next investigator drops a suspect view into `assert_http_ws_djid_parity()` and gets a definitive yes/no without a downstream consumer's private repo.
- Resisted shipping a mis-framed harness. Initial exploration revealed the naive comparison (`render()` vs `render_with_diff`) was wrong (initial server HTML carries no dj-root dj-ids; the client stamps them), and that a two-`render_with_diff` comparison would be tautological. Reframed around the *real* differential — HTTP runs `render_full_template` (which advances the Rust baseline) before its `render_with_diff`, WS doesn't — which the reviewer independently verified is non-tautological.
- Documented the finding honestly: parity HOLDS for all derivable shapes, so #1641 is not a framework-side HTTP/WS divergence for them. A negative result, surfaced rather than buried.

- **No empirical canary (Action #1468).** The harness's catch-power rests entirely on the path-differential argument; there's no synthetic fixture that makes it FAIL, so a future refactor that silently neutered the differential (e.g. `render_full_template` becoming side-effect-free) would leave the test green-but-toothless. The right strengthening per #1468 (empirical canary for tooling-class PRs): construct a view whose HTTP path deliberately perturbs the baseline (or monkeypatch the Rust id counter) and assert the harness reports the divergence. Deferred — but this is the one thing that would take it from 4 to 5. Filing as a follow-up.

### Relation to prior tasks
- The whole #1635–1645 drain is one theme — *parallel paths that can drift* (sync/async #1638, main/broadcast #1644, two DOM walkers #1640, dev/deploy #1637, double-eval scope #1635, hand-copied arming #1645). #1642 is the meta-tool for the class: a harness that asserts two paths agree. The drain didn't just fix six bugs; it left behind guards (regex writer-guard in #1645, parity harness here) that make the *class* harder to reintroduce.

RETRO_COMPLETE

PR #1656, opened 2026-05-29
fix(websocket): resolve live_redirect target view from URL, not client (#1647)

### LINKED ISSUE #1647: live_redirect_mount sends source view when client route map is empty (causes server to mount wrong view at new URL)
## Summary

After #1644 fixed the navigation message dispatch, `live_redirect` from view A to a different view B now updates the browser URL correctly — but the WS-driven remount fails with `\"Failed to load view. Please refresh the page.\"` because the client sends the **source** view's class path in the `live_redirect_mount` frame, not the target view's. The server then instantiates the source view class with the new URL's request and raises.

This is the secondary issue I mentioned in #1643 — it isn't a follow-on of #1644 (the trace below was captured against `031b69659` on `main`, with the #1644 fix in place). It's a separate, deterministic bug in the client's `resolveViewPath()` fallback.

## WS frame trace (real)

Captured against `djust@main` (commit `031b69659` — includes #1644). Synchronous `@event_handler` calls `self.live_redirect(\"/enrollment/done/\")` after persisting a model:

```
sent:     {\"type\":\"event\",\"event\":\"confirm_enrollment\",\"params\":{\"_args\":[\"aetna-medicare-choice-ppo\"]},\"ref\":1}
sent:     {\"type\":\"live_redirect_mount\",\"view\":\"medicare.views.EnrollmentConfirmView\",\"params\":{},\"url\":\"/enrollment/done/\"}
received: {\"type\":\"navigation\",\"action\":\"live_redirect\",\"path\":\"/enrollment/done/\",\"replace\":false}
received: {\"type\":\"noop\",\"ref\":1}
received: {\"type\":\"error\",\"error\":\"Failed to load view. Please refresh the page.\"}
```

Notice the second `sent` line: `view: \"medicare.views.EnrollmentConfirmView\"` is the **source** view (the page the user clicked from). The target URL `/enrollment/done/` is wired in Django's URLconf to `medicare.views.EnrollmentDoneView`. So the server takes `data[\"view\"]` from `handle_mount`, instantiates the wrong class, mounts it against a request for the new URL, the mount raises (in my case `Http404` because the source view's `mount()` expects a `plan_id` kwarg that the new URL doesn't have), and the sanitized \"Failed to load view\" error gets sent back.

## Root cause

`client.js:9400-9437` — `resolveViewPath(pathname)`:

```js
function resolveViewPath(pathname) {
    const routeMap = window.djust._routeMap || {};
    // exact match → ...
    // pattern match (\":name\") → ...
    // Fallback: check the current container's dj-view
    // (only works for live_patch, not cross-view navigation)
    const container = document.querySelector('[dj-view]');
    if (container) {
        return container.getAttribute('dj-view');
    }
    return null;
}
```

The fallback returns the **current container's dj-view** — which is by definition the source view. The comment even acknowledges this only works for `live_patch` (same-view URL update), not for cross-view `live_redirect`.

`live_redirect`'s caller (`handleLiveRedirect`, line 9350) calls `resolveViewPath(newUrl.pathname)` and uses whatever it gets back as the `view` field of the outbound `live_redirect_mount` (lines 9372-9385). When the route map is empty, the fallback fires and the source view goes out.

In MAX Companion (the repro), `window.djust._routeMap` is **empty** (`Object.keys(window.djust._routeMap)` returns `[]`). The app uses Django's standard URLconf — `medicare/urls.py` with `path(\"enrollment/done/\", EnrollmentDoneView.as_view(), name=\"enrollment-done\")` — and never calls `live_session()` (I can't find a clear example of when it'd be required; if it is required for `live_redirect` to function, that should be documented).

## Reproducer

Smallest path I can give you against my repo:

1. Check out [johnrtipton/max-companion@main](https://github.com/johnrtipton/max-companion), `make setup && make serve`
2. Open `http://127.0.0.1:8111/enrollment/confirm/aetna-medicare-choice-ppo/`
3. Scroll, click \"Confirm enrollment\"
4. URL updates to `/enrollment/done/` ✓, page-loading bar stays on, DOM stays stale, error frame returned

If a standalone demo-app repro would help I can carve one out — just say.

## Possible fix directions

(Not prescriptive — these are sketches.)

1. **Server resolves the URL itself.** Drop `data[\"view\"]` for `live_redirect_mount` in favor of `django.urls.resolve(data[\"url\"])` to find the target view class. Django already has the URL dispatcher; the client doesn't need to ship a route map for plain `path(...)` URLconfs to work.
2. **Fail loudly when the route map can't resolve.** Return `null` from `resolveViewPath` in the empty-map case (don't fall back to the source view) and have `handleLiveRedirect` log + do a full-page nav. At least the user doesn't get stuck.
3. **Auto-populate the route map** from Django's URLconf during page render. Then `live_session()` wouldn't be a hidden prerequisite.

Happy to iterate on any of these.

## Environment

- djust at `031b69659` on `main` (includes the #1644 fix)
- Django 5.2.14, channels 4.3.2, uvicorn 0.47.0
- Browser: Chrome on macOS, loopback uvicorn
- App uses plain Django `path()` URLconf — no `live_session()` call anywhere

## Verified by frame trace this time

`#1643` was filed without a WS trace and the maintainer had to discover the root cause themselves. This time I patched a logger onto `window.djust.liveViewInstance.ws` and captured both directions; the frames above are verbatim. Confirmed the empty route map by inspecting `window.djust._routeMap` in the live page.

johnrtipton:
## Summary
Fixes #1647. `live_redirect` to a *different* view failed with "Failed to load view. Please refresh the page." for apps using plain Django `path()` URLconfs.

## Root cause
The client's `resolveViewPath()` falls back to the current container's `dj-view` — the **source** view — when `window.djust._routeMap` is empty (the default without `live_session()`). The server trusted that client-supplied class in `live_redirect_mount` and instantiated the *source* view against the destination URL's request → mount raised.

## Fix (server is canonical)
`handle_live_redirect_mount` resolves the destination view server-side from `data["url"]` via Django's URL dispatcher (new `_resolve_view_path_from_url`: `resolve(url).func.view_class`) and overrides the client-supplied `view` when it maps to a djust `LiveView`. Falls back to the client value when the URL doesn't resolve to a LiveView (preserves `live_session()` route maps). **Removes the hidden `live_session()` prerequisite for plain `path()` URLconfs.**

**Back-nav guard:** the override is skipped when a `state_snapshot` is present — back-navigation carries the authoritative `view_slug` and its `url` may be generic (e.g. `/`) and resolve to an unrelated view. (This regression was caught by the pre-push suite and is pinned by a regression test.)

## Tests
`python/djust/tests/test_live_redirect_mount_resolve_1647.py` (7) + a test urlconf: helper resolves target / returns None for unmapped/non-LiveView/empty; `handle_live_redirect_mount` overrides SOURCE→TARGET (RED pre-fix, gate-off verified), keeps the client view when unresolvable, and does NOT override when a `state_snapshot` is present. Full pre-push suite green.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- **`live_redirect` to a different view now mounts the correct target (#1647).** The client's `resolveViewPath()` falls back to the current container's `dj-view` — the **source** view — when `window.djust._routeMap` is empty, which is the default for apps using plain Django `path()` URLconfs (no `live_session()`). The server trusted that client-supplied class in the `live_redirect_mount` frame, instantiated the *source* view against the destination URL's request, and raised "Failed to load view. Please refresh the page." `handle_live_redirect_mount` now resolves the destination view server-side from the URL via Django's URL dispatcher and overrides the client-supplied `view` when the URL maps to a djust `LiveView` (falling back to the client value otherwise, so `live_session()` route maps are unaffected). **`live_session()` is no longer a hidden prerequisite for `live_redirect` across plain `path()` URLconfs.**
+
 ## [1.0.0rc14] - 2026-05-28
 
 ### Added
--- a/python/djust/tests/test_live_redirect_mount_resolve_1647.py
+++ b/python/djust/tests/test_live_redirect_mount_resolve_1647.py
@@ -1,0 +1,102 @@
+"""Regression: live_redirect_mount must resolve the TARGET view server-side
+from the URL, not trust the client-supplied (source) view (#1647).
+
+The client's `resolveViewPath()` falls back to the current container's `dj-view`
+— the SOURCE view — when its route map is empty (the default for plain Django
+`path()` URLconfs with no `live_session()`). The server then instantiated the
+source class against the new URL's request and raised "Failed to load view".
+
+`LiveViewConsumer._resolve_view_path_from_url(url)` resolves the destination URL
+to its djust LiveView via Django's URL dispatcher, and `handle_live_redirect_mount`
+overrides `data["view"]` with it (falling back to the client value when the URL
+doesn't map to a LiveView).
+"""
+
+from __future__ import annotations
+
+from unittest.mock import AsyncMock, patch
+
+import pytest
+from django.test import override_settings
+
+from djust.websocket import LiveViewConsumer
+
+URLCONF = "tests.redirect_mount_test_urls"
+SOURCE = "tests.redirect_mount_test_urls.RedirectSourceView"
+TARGET = "tests.redirect_mount_test_urls.RedirectTargetView"
+
+
+@override_settings(ROOT_URLCONF=URLCONF)
+def test_resolve_view_path_from_url_returns_target():
+    consumer = LiveViewConsumer()
+    assert consumer._resolve_view_path_from_url("/redirect-target/") == TARGET
+
+
+@override_settings(ROOT_URLCONF=URLCONF)
+def test_resolve_returns_none_for_unmapped_url():
+    consumer = LiveViewConsumer()
+    assert consumer._resolve_view_path_from_url("/no-such-url/") is None
+
+
+@override_settings(ROOT_URLCONF=URLCONF)
+def test_resolve_returns_none_for_non_liveview():
+    consumer = LiveViewConsumer()
+    # A plain Django view must NOT be overridden — caller keeps the client value.
+    assert consumer._resolve_view_path_from_url("/plain/") is None
+
+
+@override_settings(ROOT_URLCONF=URLCONF)
+def test_resolve_returns_none_for_empty_url():
+    consumer = LiveViewConsumer()
+    assert consumer._resolve_view_path_from_url("") is None
+
+
+@pytest.mark.asyncio
+@override_settings(ROOT_URLCONF=URLCONF)
+async def test_handle_live_redirect_mount_overrides_source_view():
+    """The end-to-end fix: client ships the SOURCE view + the TARGET url; the
+    server must mount the TARGET (resolved from the url), not the source."""
+    consumer = LiveViewConsumer()
+    consumer.scope = {"session": None, "user": None}
+    consumer.view_instance = None  # no old view → no sticky staging
+    consumer._sticky_auto_reattached = set()
+    consumer._sticky_preserved = {}
+
+    captured = {}
+
+    async def _fake_handle_mount(data, **kwargs):
+        captured["view"] = data.get("view")
+
+    with patch.object(consumer, "handle_mount", new=AsyncMock(side_effect=_fake_handle_mount)):
+        await consumer.handle_live_redirect_mount(
+            {"view": SOURCE, "url": "/redirect-target/", "params": {}}
+        )
+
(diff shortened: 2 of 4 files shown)


johnrtipton (commented):
Stage 11 review — PR #1656 (#1647 live_redirect mount view-resolution).

Base: not stale (BEHIND=0). Scope confirmed: `_resolve_view_path_from_url` helper + URL-override + state_snapshot back-nav guard in websocket.py, a 138-line test, a test urlconf, CHANGELOG. Three commits (override / CHANGELOG / snapshot-guard) — CHANGELOG sits mid-stack rather than last, harmless.

Correctness: the override fires only when the URL resolves to a djust LiveView differing from the client value AND no snapshot. `resolve(url).func.view_class` works for BOTH plain `.as_view()` and `live_session()` routes — `routing.py:162` reuses `pattern.callback` verbatim (no re-wrap), so `view_class` survives. The docstring says live_session is "unaffected"; technically the override still runs there, but client+server resolve the same URL so `resolved_view == data["view"]` makes it a no-op. A non-LiveView / Resolver404 returns None → client value kept. Forward-redirect-without-snapshot: the URL IS the authoritative target, so trusting it is correct; I could not construct a legitimate case where the client value should win with no snapshot.

Tests cover all four helper branches + override + snapshot guard. Ran the suite: 9 passed. Gate-off (#254): neutralizing the snapshot guard fails both back-nav tests (test_sw_advanced_flow + the unit guard test) — the guard is load-bearing, not tautological. Restored clean.

APPROVE

johnrtipton:
## Stage 14 Retrospective — PR #1656 (#1647)

**Quality: 4/5**

### What went well
- Took the architecturally-correct fix (reporter's option 1: server resolves the URL) over the easy client patch. djust's "server is canonical" principle made the call obvious — Django already owns the URL dispatcher, so trusting the client's `view` was the design smell. Bonus: removed the hidden `live_session()` prerequisite for plain `path()` URLconfs.
- Symptom-up again beat the cited line: the reporter pointed at `client.js:resolveViewPath`, but the durable fix was server-side, not patching the client fallback.

- **I shipped a too-broad override and the pre-push suite caught the regression — not my own test.** My first cut overrode `data["view"]` from the URL unconditionally. That broke back-navigation, where a `state_snapshot` carries the authoritative `view_slug` and the `url` is generic (`/`) and resolves to an unrelated view. My #1647 tests all passed (they only exercised forward redirects); the full pre-push suite's `test_back_nav_with_state_snapshot_restores_state` is what flagged it. **Left-shiftable lesson:** when a fix OVERRIDES a client-supplied field on a shared handler (`handle_live_redirect_mount` serves forward-redirect AND back-nav AND sticky paths), Stage 4 must enumerate *every* caller/mode of that handler and add a test per mode BEFORE implementing — exactly the "N call sites need N tests" rule (#1104), here "N modes of one handler need N tests." I tested one mode (forward) and let the suite find the other (back-nav). The guard + its regression test are now in place, but the cost was a failed push + a follow-up commit.
- This also vindicates the pre-push full-suite gate: had I been able to push the narrow PR, the back-nav break would have reached CI/review instead of being caught locally.

### Relation to prior tasks
- Same drain meta-theme inverted: the bug was a client/server *disagreement* about which view is authoritative; the fix made the server authoritative — but introduced a NEW path-divergence (forward vs back-nav modes of one handler) that the `state_snapshot` discriminator resolves. The recurring discipline holds: enumerate every mode/path a shared entry point serves.

RETRO_COMPLETE

Issue #1515, opened 2026-05-18
tech-debt: codify the scripts/check-*.py audit-shape as a scaffold/template


johnrtipton:
Source: Retro v1.0.0rc2 (finding #2 — milestone synthesis).

The v1.0.0rc2 drain added 3 standing audit scripts in 3 PRs — `scripts/check-adr-status.py` (#1506), `scripts/check-doc-snippets.py` (#1508), `scripts/check-lockfile-versions.py` (#1510) — plus the Y003/Y004 checks. They share a now-stable shape:

- Pure stdlib, no network, no shell-out; `tomllib`/regex parsing.
- Testable `run() -> tuple[int, str]` core + `build_arg_parser()` with path-override flags + `main(argv=None)`; exit codes 0/1/2.
- Wired identically into `make <target>`, `.github/workflows/test.yml`, `.pre-commit-config.yaml`.
- A dedicated `tests/test_check_*.py` with subprocess+tempfile fixtures, a tautology-guarded gate-off self-test (Action #254), and a `@pytest.mark.slow` real-repo dogfood test.

Three siblings now exist; the fourth should be fill-in-the-blank, not a re-derive. Codify the shape — either a documented mini-template in `scripts/README` / `docs/`, or a `make new-audit NAME=...` scaffold that stamps out the script + test + wiring stubs.

Optionally also add a consolidated `make audit-consistency` target that runs all standing `check-*.py` audits at once (surfaces the audit family as one developer entry point).

PR #1657, opened 2026-05-29
fix(auth): wrap per-event check_handler_permission in sync_to_async (#1648)

### LINKED ISSUE #1648: bug: check_handler_permission called bare-sync from async _validate_event_security (sibling of #1638)
Sibling of #1638 (found in PR #1646 review). `_validate_event_security` (async) calls `check_handler_permission(handler, owner_request)` **bare-sync** at `python/djust/websocket_utils.py:212`:

```python
if owner_request and not check_handler_permission(handler, owner_request):
    await ws.send_error("Permission denied")
    return None
```

`check_handler_permission` short-circuits to `True` for handlers without `@permission_required`, so the common case is unaffected. But for a handler decorated with `@permission_required`, an authenticated non-superuser triggers `user.has_perms()` → a DB read under the default `ModelBackend` → the **same `SynchronousOnlyOperation`** class #1638 fixed for `check_object_permission`, which the surrounding handling would turn into a spurious "Permission denied".

**Fix**: wrap in `sync_to_async` mirroring the #1638 fix:
```python
if owner_request and not await sync_to_async(check_handler_permission)(handler, owner_request):
```

**Test (per #1104 — N sites need N tests)**: a `@permission_required` handler + a DB-backed permission backend, driven through `_validate_event_security`; assert it doesn't raise `SynchronousOnlyOperation` / doesn't false-deny. #1638's reproducer does NOT cover this path.

johnrtipton:
## Summary
Fixes #1648 — the sibling of #1638 surfaced during that PR's review.

## Root cause
`_validate_event_security` (async) called `check_handler_permission` **bare-sync**. For a handler decorated with `@permission_required`, that calls `user.has_perms()`, which under Django's default `ModelBackend` queries the DB for a non-superuser → `SynchronousOnlyOperation` in the event loop. Unlike the object-permission path (#1638), this call site had **no fail-closed catch**, so the exception propagated out of `_validate_event_security`.

## Fix
Wrap the call in `sync_to_async`, mirroring #1638 and the mount path.

## Tests
`tests/integration/test_handler_perm_async_wrap_1648.py` — a `@permission_required` handler whose `has_perms()` hits the DB (real non-superuser): passes when the user has the perm, denies (no raise) when not. RED pre-fix with the exact `SynchronousOnlyOperation`. (Per #1104 "N sites need N tests" — #1638's reproducer didn't cover this path.) Object-permission suite stays green (14 passed).


johnrtipton (commented):
Stage 11 review — PR #1657 (sibling of #1638, #1648)

**Stale-base (#1450):** BEHIND=0 vs origin/main — reviewing the program that will merge.

**Diff scope:** Exactly the one-line `sync_to_async` wrap in `_validate_event_security` (websocket_utils.py:216) + a 2-case regression test + a `### Fixed` CHANGELOG entry. Nothing extraneous. `sync_to_async` already imported (line 13).

**Correctness:** `_validate_event_security` is `async def` (line 156) — `await` valid. The wrap preserves the `if owner_request and not <bool>:` semantics (awaited call returns the same bool). The `permission_required and not owner_request` early-deny branch above (lines 206-211) is untouched. Mirrors the object-permission fix at line 273 and the mount path.

**Gate-off (#254):** Reverted the wrap to bare-sync → BOTH tests fail with `SynchronousOnlyOperation` from `ModelBackend.get_all_permissions` DB cursor. Restored → 2/2 pass. Tests exercise the actual changed line, not a tautology.

**Tests:** `test_handler_perm_async_wrap_1648.py` (2) + `test_object_permission_event.py` (12) = 14 passed.

APPROVE

johnrtipton:
## Stage 14 Retrospective — PR #1657 (#1648)

**Quality: 5/5**

### What went well
- This is the payoff of #1646's review discipline: the Stage 11 reviewer of the #1638 fix *spotted the sibling* (`check_handler_permission`, same bare-sync-in-async class) and I filed it instead of dropping it. The drain then closed it. The bug class — "auth predicate called bare-sync from an async handler" — is now retired across BOTH call sites (`check_object_permission` #1638, `check_handler_permission` #1648).
- Honored "N sites need N tests" (#1104): #1638's reproducer used object-permission; this one needed its own — a `@permission_required` handler whose `user.has_perms()` hits the DB under `ModelBackend` for a real non-superuser. Without a real saved user the DB query (and thus the `SynchronousOnlyOperation`) wouldn't fire — repro fidelity again decided reproducibility.

- **This call site had no fail-closed catch, so the bug was *louder* than #1638 — and that's the more dangerous shape.** #1638's `except` turned the async error into a (wrong but contained) "Access denied"; #1648 let `SynchronousOnlyOperation` propagate out of `_validate_event_security` entirely. Left-shiftable: an audit grep for "sync auth/ORM predicate invoked without `sync_to_async` inside an `async def`" across `websocket_utils.py` / `websocket.py` would have surfaced BOTH at once. Worth a one-off `scripts/check-*` AST lint (the #1515 audit-template shape) — flag any bare call to a known-sync auth helper (`check_*_permission`, `has_perm*`) inside an `async def`. Noting as a candidate; not filing a third issue mid-drain.

### Relation to prior tasks
- Direct completion of the #1638 finding and a clean instance of the session's theme: a per-path invariant (`sync_to_async`-wrap sync work in async) correct on the mount path, missing on the per-event path — fixed identically at every site.

RETRO_COMPLETE

PR #1658, opened 2026-05-29
test(infra): empirical-canary for assert_http_ws_djid_parity (#1654)

### LINKED ISSUE #1654: test-infra: add an empirical-canary fixture for assert_http_ws_djid_parity (#1468 spirit, from #1642)
From the #1642 (PR #1653) retro. `assert_http_ws_djid_parity()` currently passes for every shape, so its catch-power rests on the path-differential argument (HTTP runs render_full_template before its baseline; WS doesn't) rather than an empirical failing fixture. Per Action #1468 (empirical canary for tooling-class additions), add a synthetic view/monkeypatch whose HTTP path deliberately perturbs the dj-id baseline, run the harness against it, and assert it REPORTS the divergence — proving the harness catches the class it claims to, not just that parity happens to hold today.

johnrtipton:
## Summary
Closes #1654 (Action #1468 spirit). `assert_http_ws_djid_parity` (added in #1642) passes for every real view shape, so its catch-power rested on the path-differential argument rather than an empirical failing case.

## What
Adds a canary test that **forces** a dj-id divergence (patches `_djroot_djids` to return different HTTP vs WS baselines) and asserts the harness **raises** with a useful diagnostic (names both baselines), plus a control proving identical baselines don't raise. This guards against a future refactor silently neutering the comparison (green-but-toothless).

Pure test addition (no production change). 6 passed in the parity file.

### DIFF
--- a/python/djust/tests/test_http_ws_djid_parity_1642.py
+++ b/python/djust/tests/test_http_ws_djid_parity_1642.py
@@ -21,6 +21,8 @@ the next investigator can drop a suspect view in and get a definitive answer.
 
 from __future__ import annotations
 
+from unittest.mock import patch
+
 import pytest
 
 from djust import LiveView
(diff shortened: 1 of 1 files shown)


johnrtipton (commented):
Stage 11 review (test-only #1654). BEHIND=0 vs origin/main — base fresh.

1. **Test-only, correct ceremony.** Diff is +40 lines in test_http_ws_djid_parity_1642.py only; no production change, no CHANGELOG — correct for a `test:` PR.

2. **Canary is meaningful.** `assert_http_ws_djid_parity` calls `_djroot_djids` exactly twice — line 419 (HTTP baseline) then line 425 (WS baseline). The 2-element `side_effect=[["0","1"],["0","2"]]` maps 1:1 to those calls, so the divergence hits the real `assert http_ids == ws_ids` at line 427 (AssertionError), not a StopIteration. Message asserts ("divergence" + both baseline reprs) match the harness diagnostic verbatim.

3. **Control is sound.** Identical baselines `[["0","1"],["0","1"]]` assert non-raise + `ids == ["0","1"]`, isolating the patch machinery from the failure cause.

4. **Tests pass:** 6 passed in 0.22s.

APPROVE

johnrtipton:
## Stage 14 Retrospective — PR #1658 (#1654)

**Quality: 5/5**

### What went well
- Closed the one 🟡 the #1642 retro itself flagged: the parity harness was "trusted from the path-differential argument" but had no empirical proof it fires. Now it does — a forced-divergence canary + a matching-baseline control. This is Action #1468 applied to a test harness (not just a lint): the harness is a tool whose claim ("catches HTTP/WS dj-id divergence") now has an executable verifier.
- The control test is the part that makes the canary trustworthy: it proves the canary's raise comes from the divergence, not the `patch` machinery — without it, a canary that raised for the wrong reason would read as green-confidence.

- **The canary patches the comparison inputs, not a real render divergence.** It proves the *assertion logic* fires on unequal lists; it does NOT prove a real `render_full_template`-vs-`render_with_diff` divergence would produce unequal lists (the harness's actual job). A stronger canary would monkeypatch the Rust dj-id counter or inject a view whose HTTP path genuinely perturbs the baseline. Left-shiftable: when canary-testing a *comparison harness*, prefer forcing the divergence at the deepest layer you can (the render), falling back to input-patching only when that's impractical. Filed mentally; not worth a further issue — the input-patch canary already retires the "green-but-toothless comparison" risk, which was #1654's actual ask.

### Relation to prior tasks
- Caps the #1641→#1642→#1654 chain: investigation (can't repro) → harness (testable) → canary (harness proven to bite). Each step turned a soft assurance into a mechanical one — the session's recurring move.

RETRO_COMPLETE

PR #1659, opened 2026-05-29
test/refactor: strengthen drain guards — predicate, caller-count, deploy test (#1655)

### LINKED ISSUE #1655: tech-debt: strengthen v1.0.0rc14 drain guards with empirical/integration coverage
Consolidated from the v1.0.0rc14 milestone retro (drain #1635–1645). Three deferred guard-strengthenings, each non-blocking:

1. **Scaffold deploy-path integration test (#1651 retro).** The `djust new` migrations fix is pinned by a command-sequence test + a `migrations/__init__.py` existence test, but not by a full "generate → plain `migrate` on a fresh DB (no `--run-syncdb`) → query a demo model" integration test. The dev-vs-deploy command asymmetry was the actual trap; an integration test exercising the *deploy* path would catch a regression the command-list test can't.

2. **Count-based caller enumeration for `_arm_recovery` (#1652 retro, Action #1125 spirit).** The regex writer-guard pins that `_recovery_html` is only assigned via the helper, but nothing asserts every render-send path actually *calls* it (the #1639 shape: a new path that renders+sends but never arms). A count-based test enumerating `_send_update`/`_send_noop` call sites and asserting each is preceded by `_arm_recovery` (or a deliberate no-arm) would close the caller-side gap.

3. **Single `isSignificantChild` predicate (#1649 retro).** `getNodeByPath` and `getSignificantChildren` now share `isDjIfComment`, but each still independently decides "which children count." If a 3rd comment-counting site appears, extract a single `isSignificantChild(node)` predicate so the dj-if-only rule can't drift again.

Each is < 1 day; pick up in a future test-infra pass. None blocks 1.0.

johnrtipton:
## Summary
Closes #1655 — three consolidated guard-strengthenings from the v1.0.0rc14 drain retros.

### (1) Single `isSignificantChild` predicate (#1649)
`getNodeByPath` and `getSignificantChildren` had independently-written copies of the "which children count toward VDOM indices" rule; the copy in `getSignificantChildren` drifted to count ALL comments → the #1640 index mismatch. Extracted **one** `isSignificantChild(child, preserveWhitespace)` used by both, mirroring `crates/djust_vdom/src/parser.rs`. Pure refactor (no behavior change); full JS suite green (1626). + `tests/js/significant_child_predicate_1655.test.js`.

### (2) `_arm_recovery` caller-count guard (#1652, Action #1125)
Pins the 4 known `self._arm_recovery()` call sites so a new render-send path can't be added without consciously arming the recovery baseline (the #1639 shape).

### (3) Scaffold deploy-path integration test (#1651)
Generates a `--with-db` project, runs `makemigrations` + **plain** `migrate` (no `--run-syncdb`, as a deploy does) via subprocess, and asserts the demo model's table is created — the gap #1637's command-sequence unit tests didn't cover. Skips gracefully if the generated project's extras aren't installed in the env.

No CHANGELOG — internal refactor + test hardening, no user-facing behavior change.


johnrtipton (commented):
Reviewed at HEAD; branch 0 commits behind origin/main (stale-base check #1450 passed).

**Scope** — three consolidated hardenings (#1655), no behavior change, no CHANGELOG (correct). Diff confined to `12-vdom-patch.js` + rebuilt bundle artifacts + 3 tests.

**(1) `isSignificantChild` behavior-preservation — verified.** The helper is byte-equivalent to both prior inline filters: ELEMENT always true; TEXT runs the same ASCII-whitespace regex `/[^ \t\n\r\f]/` (NBSP-preserving); COMMENT gates on identical `isDjIfComment`. Critically, `getNodeByPath` had no whitespace-preserve branch before and now calls `isSignificantChild(child)` with default `preserveWhitespace=false`, which skips the early-return and runs the same regex — its always-strip behavior is preserved. `getSignificantChildren` forwards its computed `preserveWhitespace`. Rebuilt `client.js` contains the helper + both call sites (1 helper, 1 default call, 1 preserveWhitespace call). `git status` clean after `bash scripts/build-client.sh` (deterministic bundle). 26/26 JS tests pass (significant_child_predicate_1655, comment_filter_1640, vdom_if_block, insert_remove_child_id).

**(2) arm_recovery count.** Asserts exactly 4 `self._arm_recovery(` sites; real count in websocket.py is 4 (958, 977, 3646, 5056 — the 5th grep hit is the def). Matches. 4/4 pass.

**(3) deploy-migrate.** Uses PLAIN `migrate --noinput` (no `--run-syncdb`), asserts a `deployapp1655_*` table exists; `--skip-checks` bypasses dev system-check hints while migrate return codes are independently asserted ==0 — legit. PASSES (not skipped) in this env.

No security concerns. Tests are non-tautological (count test, build-artifact grep, real table assertion).

APPROVE

johnrtipton:
## Stage 14 Retrospective — PR #1659 (#1655)

**Quality: 5/5**

### What went well
- Each of the 3 sub-items turned a *soft* assurance from an earlier drain PR into a *mechanical* one, completing the session's through-line:
  - **(1) single `isSignificantChild`** — the #1640 fix made the two comment-filters *agree*; this makes them *share one definition*, so they can't drift again. The structural cure the #1640 retro itself called for.
  - **(2) `_arm_recovery` count guard** — the #1645 regex writer-guard pinned "only the helper writes `_recovery_html`"; this pins "only N paths arm it", closing the caller-side gap the #1645 retro flagged (a new send path that forgets to arm).
  - **(3) deploy-path integration test** — #1637's unit tests pinned the command *sequence*; this exercises the actual *deploy* path (plain `migrate` on a fresh DB → real table), the gap the #1637 retro named. It PASSES (not skips) in CI.
- Reproduction fidelity paid off again in (3): the test surfaced that the generated demo app trips djust's dev-time system checks during `migrate`; `--skip-checks` was the right scope (test migrations, not check-cleanliness) rather than a weaker proxy.

- **Consolidating three unrelated hardenings into one PR widened the review surface** (JS refactor + websocket test + scaffold integration test — three subsystems). It reviewed cleanly here because each is small and independently testable, but the cleaner shape would have been three tiny PRs. Left-shiftable: when a "consolidated tech-debt" issue spans >1 subsystem, split per-subsystem at drain-triage time rather than bundling — a bundle is only justified when the items share a subsystem (the audit-drain recipe's own ≤7-same-subsystem rule). I bundled because #1655 was filed as one issue; the filing, not the work, drove the shape.

### Relation to prior tasks
- The capstone of the post-rc14 drain: every guard the drain's retros asked for now exists. The session's arc — investigate → fix → guard → *prove the guard bites* — is closed for this backlog.

RETRO_COMPLETE

PR #1661, opened 2026-05-29
fix(theming): break cyclic-import SCC via leaf _config module (CodeQL #2351, #2357-#2362)

johnrtipton:
## Summary
Resolves the 7 open CodeQL `py/cyclic-import` alerts in `djust.theming`: **#2351** (presets.py:42), **#2357–#2361** (css_generator.py:45/544/616/661/707), **#2362** (manager.py:518).

## Root cause
The prior fix (#2352) broke the *eager* import cycle by making cross-module imports lazy, and `test_theming_no_cyclic_import.py` gates the module-**top-level** graph (deliberately ignoring lazy imports). But CodeQL's `py/cyclic-import` counts **lazy edges too**, so the `→ manager.get_theme_config` edges (from registry, css_generator, theme_css_generator, pack/component_css_generator) plus the `presets → registry → manager → presets` loop kept the SCC alive in CodeQL's view — re-surfacing as these 7 alerts.

## Fix
Extract `get_theme_config` + `DEFAULT_CONFIG` + `_validate_cookie_namespace` (+ `COOKIE_NAMESPACE_RE`) from `manager.py` into a new **leaf** `_config.py` (imports only Django + stdlib). Every reader now imports the config from `._config` instead of `.manager`, removing all manager-ward edges from the generator/registry modules. `manager` **re-exports** `get_theme_config`/`DEFAULT_CONFIG`, so `from djust.theming.manager import get_theme_config` keeps working — **no behavior change, no public-API change.**

Result: `presets` / `manager` / `css_generator` are now outside every import SCC (verified by a reachability test). The **pre-existing, unflagged** `registry↔theme_packs` / `registry↔manifest` SCC is left as-is (separate concern, never flagged) — staying scoped to the 7 reported alerts.

## Tests
4 new gates in `test_theming_no_cyclic_import.py`: `_config` stays a leaf; no module imports `get_theme_config` from `.manager`; the manager re-export is the same object; and the flagged modules are outside every SCC (so the alerts can't silently regress). 1844 theming tests green; full pre-push suite green.

### DIFF
--- a/python/djust/tests/test_theming_no_cyclic_import.py
+++ b/python/djust/tests/test_theming_no_cyclic_import.py
@@ -135,3 +135,111 @@ class TestThemingNoCyclicImport:
         assert get_preset("blue").name == "blue"
         # Unknown name falls back to DEFAULT_THEME (resolution step 3)
         assert get_preset("definitely-not-a-real-theme-xyz").name == "default"
+
+
+class TestConfigReaderIsLeaf:
+    """Gate the #2351/#2357-#2362 cycle-break: ``get_theme_config`` lives in the
+    leaf ``_config`` module and NO theming module imports it from ``manager``.
+
+    The earlier fix (#2352) broke the *eager* SCC by making cross-module imports
+    lazy, but CodeQL's ``py/cyclic-import`` counts lazy edges too — so the
+    ``→ manager.get_theme_config`` edges from registry / css_generator /
+    theme_css_generator / pack_css_generator / component_css_generator / checks /
+    components kept the SCC alive in CodeQL's view. Extracting ``get_theme_config``
+    to the leaf ``_config`` removed every one of those edges. These tests gate
+    the static import graph (lazy + eager) so the cycle can't silently return.
+    """
+
+    def test_config_module_is_a_leaf(self):
+        """``_config`` must import NOTHING from the cycle modules — it's the leaf
+        registry/css_generator read instead of manager."""
+        import ast
+
+        src = (THEMING_DIR / "_config.py").read_text()
+        cycle_mods = {
+            "presets",
+            "registry",
+            "manager",
+            "css_generator",
+            "theme_css_generator",
+            "pack_css_generator",
+            "component_css_generator",
+        }
+        for node in ast.walk(ast.parse(src)):
+            if isinstance(node, ast.ImportFrom) and node.level == 1 and node.module:
+                base = node.module.split(".")[0]
+                assert base not in cycle_mods, (
+                    f"_config.py must stay a leaf but imports .{base} — "
+                    "that would re-create the cyclic-import SCC (#2351)."
+                )
+
+    def test_no_theming_module_imports_get_theme_config_from_manager(self):
+        """Every reader of ``get_theme_config`` must import it from ``._config``,
+        not ``.manager`` — importing from manager re-creates the
+        manager-centered cycles CodeQL flagged (#2357-#2362)."""
+        import ast
+
+        offenders = []
+        for path in THEMING_DIR.glob("*.py"):
+            if path.name == "manager.py":
+                continue  # manager legitimately re-exports it from _config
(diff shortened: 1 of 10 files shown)


johnrtipton (commented):
Reviewed at BEHIND=0 (not stale).

**Scope** — matches description exactly: new leaf `_config.py`, `manager.py` (defs removed + re-export + unused `re`/`ImproperlyConfigured` dropped), 6 importer rewires (components.py correctly split into two imports keeping `ThemeManager` from manager), 4 regression tests. No CHANGELOG — correct, zero behavior change.

**Behavior preservation** — `_config.get_theme_config`/`DEFAULT_CONFIG`/`_validate_cookie_namespace`/`COOKIE_NAMESPACE_RE` are byte-identical to the old manager definitions (verified line-by-line against the diff). `_config.py` imports ONLY `re` + `django.conf.settings` + `django.core.exceptions.ImproperlyConfigured` — true leaf. `test_manager_reexports_get_theme_config_for_backcompat` asserts the re-export is the SAME object; passes.

**Cycle broken** — `test_flagged_modules_are_outside_every_import_scc` builds the lazy+eager import graph (all level-1 ImportFrom) and confirms presets/manager/css_generator cannot reach themselves. Pre-existing registry↔theme_packs/manifest SCC is separate and untouched (out of scope). Passes.

**Gate-off** — reverted css_generator.py:544 to `from .manager import get_theme_config`: 2 tests FAIL (`test_no_theming_module_imports...` + the SCC test re-detects manager in a cycle). Restored. Confirms the tests exercise the actual change, not tautologies.

**Tests** — cyclic-import suite 11/11 green; theming sweep `-k "theming or theme"` 1835 passed, 1 skipped. Working tree restored clean.

Clean, well-tested, correctly scoped cycle-break.

APPROVE

johnrtipton:
## Stage 14 Retrospective — PR #1661 (CodeQL #2351, #2357-#2362)

**Quality: 5/5**

### What went well
- **Read the prior fix before re-fixing.** `test_theming_no_cyclic_import.py` documented that #2352 had already broken the *eager* cycle by making imports lazy, and that the test *deliberately* ignores lazy edges. That immediately reframed the task: this isn't "the cycle came back", it's "CodeQL counts lazy edges that the prior fix left in place." Without reading that test I'd have mis-diagnosed.
- **Found the one extraction that breaks every flagged edge.** All 7 alerts (+ several unflagged manager↔generator cycles I discovered while verifying) shared a single root: cross-module imports of `manager.get_theme_config`. `get_theme_config` was self-contained (Django settings only), so extracting it to a leaf `_config.py` collapsed the whole manager-centered SCC at once — 6 importer rewires, zero behavior change, `manager` re-export preserving the public path.
- **A reachability SCC test as the regression gate** (not just a grep): it builds the lazy+eager graph and asserts the flagged modules can't reach themselves. Gate-off confirmed it re-detects a cycle when any rewire is reverted.

- **The AST cycle-check surfaced two *unflagged* pre-existing SCCs** (`registry↔theme_packs`, `registry↔manifest`) that CodeQL hasn't reported. I scoped them OUT (the task was the 7 listed alerts) and said so explicitly — but they're latent CodeQL alerts waiting to happen. **Left-shiftable:** the regression test should arguably gate the WHOLE theming graph as acyclic, not just the 3 flagged modules — that would force those two SCCs to be addressed too. I gated only the flagged modules to stay scoped + low-risk; a follow-up to make the entire theming package acyclic (extract `get_registry`'s themes-dir discovery the same way) is the complete-the-job move. Worth a tech-debt issue.

### Relation to prior tasks
- Same shape as the whole session: a maintainability invariant (acyclic imports) enforced on the path that was checked (eager / top-level) but not the path that wasn't (lazy edges, which CodeQL sees). The cure — make the dependency structural (a leaf the cycle can't route through) — mirrors `_flush_all_pending`/`_arm_recovery`/`isSignificantChild`: one source of truth instead of N drift-prone copies.

RETRO_COMPLETE

Issue #1260, opened 2026-05-01
VDOM: fuzz_test::round_trip_correctness finds keyed/unkeyed mixed-children diff failure


johnrtipton:
Source: proptest fuzzer surfaced this case during local testing for v0.9.2rc1 release prep.

## Failing case (shrunk by proptest)

```rust
tree_a = VNode {
    tag: "section",
    children: [
        VNode { tag: "#text", text: Some("A"), ... },  // unkeyed
        VNode { tag: "#text", text: Some("A"), ... },  // unkeyed
        VNode { tag: "#text", text: Some("A"), ... },  // unkeyed
        VNode { tag: "#text", text: Some("a"), ... },  // unkeyed
        VNode { tag: "div", key: Some("f"), ... },     // keyed
    ],
}

tree_b = VNode {
    tag: "section",
    children: [
        VNode { tag: "div", key: Some("f"), ... },     // keyed (moved to front)
        VNode { tag: "#text", text: Some("A"), ... },  // unkeyed
        VNode { tag: "#text", text: Some("a"), ... },  // unkeyed
    ],
}
```

## Bug class

**Mixed keyed/unkeyed siblings under reorder + count-change.** `tree_a` has 4 unkeyed text children + 1 keyed div at end; `tree_b` has the keyed div moved to front with 2 of the 4 unkeyed text children removed. The diff/patch round-trip fails the equivalence assertion.

This is exactly the failure mode the audit identified at weakness #5 + #6 (mixed keyed/unkeyed siblings produce suboptimal/incorrect patches — `crates/djust_vdom/src/diff.rs:411-425`). PR #1258 (#1254) shipped warnings for this case but did NOT fix the underlying patch generation. This issue is the actual fix.

## Reproducer

The case is captured in `crates/djust_vdom/tests/fuzz_test.proptest-regressions` (NOT committed yet — see below). To reproduce:

```bash
# Locally:
cat > crates/djust_vdom/tests/fuzz_test.proptest-regressions <<'EOF'
# Seeds for failure cases proptest has generated in the past.
cc 2381cf497d108e6df2bcba53149e5aeb963962eed6cbd2ffc8d22fc056fda90f # see issue body for tree_a / tree_b
EOF
PYO3_PYTHON=$(pwd)/.venv/bin/python cargo test -p djust_vdom --test fuzz_test
# Expected: 1 failed (round_trip_correctness)
```

## Why not committed yet

[...]

Issue #1254, opened 2026-05-01
VDOM: promote duplicate-key + mixed-keyed-unkeyed warnings from vdom_trace! to tracing::warn! with stable error codes


johnrtipton:
Source: VDOM engine audit 2026-04-30 §3 weaknesses #5, #6.

## Problem

Two silent-failure paths in keyed diffing:

1. `crates/djust_vdom/src/diff.rs:458-488` — duplicate `dj-key` siblings: the second overwrites the first in the HashMap. The earlier sibling becomes invisible to the diff. Warning is emitted only via `vdom_trace!()` which is gated behind `DJUST_VDOM_TRACE=1`. In production, developers don't see it.

2. `crates/djust_vdom/src/diff.rs:411-425` — mixed keyed/unkeyed siblings produce suboptimal/incorrect patches. Same `vdom_trace!()` issue. Bonus: the warning text references `https://djust.org/errors/DJE-050` which doesn't exist.

## Fix

- Replace `vdom_trace!()` with `tracing::warn!` (or equivalent) so the warning fires at INFO level by default, with stable error codes (`DJE-050` for mixed, `DJE-051` for duplicate).
- Either remove the URL or add a real error-codes doc page (e.g., `docs/internal/error-codes.md`).
- Optional: emit a structured warning frame to the client so the dev tools panel can surface it.

## Acceptance

- Both warnings fire with stable error codes at default log level.
- Cited URL either resolves or is removed from the warning text.
- One regression test confirming the warning is logged.

## References

- VDOM audit 2026-04-30 §3 weaknesses #5, #6
- `crates/djust_vdom/src/diff.rs:411-425`, `:458-488`

PR #1665, opened 2026-05-29
vdom: clean-room rebuild of diff.rs from tests, hardened beyond the original

johnrtipton:
## What

Clean-room reconstruction of `crates/djust_vdom/src/diff.rs` (`sync_ids` + `diff_nodes`) from the test suite alone — no reading of the original implementation, git history, or PR diffs during the build — followed by differential analysis against the original and three robustness hardening fixes.

Two commits:
1. `feat(vdom): clean-room rebuild of diff.rs from test suite` — 683 LoC (vs the original's ~1397 LoC of algorithm), all-green on the first compile that linked.
2. `fix(vdom): harden diff.rs — duplicate keys, unkeyed moves, ref_d` — closes gaps surfaced by differential testing, test-first.

## Algorithm (reconstructed)

- Positional/keyed child reconciliation; keyed uses LIS move-minimization (`lis::longest_increasing_subsequence`), LIS-skip disabled in mixed keyed/unkeyed lists (#1260).
- dj-if conditional subtrees matched by boundary marker id: id-only-old → `RemoveSubtree`, id-only-new → `InsertSubtree` (index = open marker's absolute index), id-in-both → recurse; nested cascades bundle into one `InsertSubtree.html` (#1358). Non-boundary siblings paired by relative position so boundary span-length shifts never cascade.
- `sync_ids` mirrors the diff's matching, dj-if-boundary-aware, carries both `djust_id` and the `dj-id` attr forward (#1408).
- `dj-update="ignore"` interiors emit no patches (#1252/#1417); `data-djust-replace` clears-and-fills; `dj-*` attrs never removed; DJE-050/051 via `tracing::warn!` (#1254).
- Every emitted `d`/`child_d`/`ref_d` resolves in the OLD tree (#1408 invariant).

## Hardening (commit 2) + new tests

`tests/test_diff_robustness_gaps.rs` (5 cases, all RED before the fix):

1. **Duplicate keys** — a key appearing >1× on either side is treated as effectively-unkeyed and reconciled positionally, so a last-wins map can't match two new nodes onto one old node (no corrupting cross-tag `Replace`).
2. **Unkeyed `MoveChild`** in the keyed path — a repositioned unkeyed node with a `djust_id` now moves by id, so it isn't stranded by a keyed reorder.
3. **`InsertChild.ref_d`** — populated as `old.get(new_local).djust_id`; always #1408-safe.

## Verification

- **264 tests pass** (259 + 5 new); clippy clean; fmt clean; deterministic across repeated runs.
- 5/5 adversarial gate-offs (Action #1200/#1468) turn their target test RED.
- **Differential test** (both implementations in one crate, 12k inputs): on realistic re-renders the hardened rebuild's round-trip failures dropped from 117→9 per 6000 and it is now **more robust than the original**, which still corrupts on duplicate-key re-renders (no positional fallback outside its boundary path — a latent bug worth a separate look for maintainers).

Full analysis in the worktree's `scratch/vdom-rebuild/ANALYSIS_REPORT.md` (not committed — experiment artifact).

### DIFF
--- a/crates/djust_vdom/src/diff.rs
+++ b/crates/djust_vdom/src/diff.rs
@@ -1,3026 +1,683 @@
-//! Fast virtual DOM diffing algorithm
+//! Virtual DOM diffing.
 //!
-//! Uses a keyed diffing algorithm for efficient list updates.
-//! Includes compact djust_id (dj-id) in patches for O(1) client-side resolution.
+//! Computes a minimal `Vec<Patch>` transforming an OLD `VNode` tree into a NEW
+//! one, and a companion `sync_ids` pass that carries stable `djust_id`s forward
+//! from the old tree onto the new tree after a render.
 //!
-//! ## Conditional Rendering (`{% if %}`)
+//! Key behaviors (reconstructed from the test suite):
+//! - Element nodes are matched positionally (or by `key` when keyed).
+//! - `{% if %}` conditional subtrees are wrapped by `<!--dj-if id="X"-->` /
+//!   `<!--/dj-if-->` comment markers and matched by boundary id (keyed
+//!   subtree diff): id-only-in-old -> `RemoveSubtree`, id-only-in-new ->
+//!   `InsertSubtree`, id-in-both -> recurse into the body. Non-boundary
+//!   siblings around a boundary are paired by RELATIVE position so a
+//!   boundary span-length change never cascades into mis-targeted paths.
+//! - Every targeting handle (`d`/`child_d`/`ref_d`) an emitted patch carries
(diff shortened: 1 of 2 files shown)


johnrtipton:
Filed #1666 for the one residual limitation surfaced by this work: a **matched dj-if boundary can't be repositioned** (markers have no `djust_id`; no `MoveSubtree` patch). It's pre-existing (the original differ fails identically) and accounts for ~all residual round-trip failures (~0.65% on an adversarial mutation corpus, ~0 in real templates). The known-limitation test `known_limitation_matched_djif_boundary_not_repositioned` here will fail-and-flip when #1666 lands.

Issue #1672, opened 2026-05-30
templates: runtime-SafeString from a custom filter is over-escaped in {% firstof %}/{% cycle %} (get_value pipe path)


johnrtipton:
## Summary

Follow-up to **#1660** (custom filter returning `mark_safe()` at runtime is now honoured for `{{ var|filter }}`). #1660 threaded runtime-SafeString-ness through the **Variable** and **InlineIf** render arms. One **parallel path** still drops the flag: the `get_value` pipe helper at `crates/djust_templates/src/renderer.rs:1793` calls the plain `apply_filter_full` (not `apply_filter_full_safe`), so a custom filter that `mark_safe()`s its output at runtime is **over-escaped** when used inside `{% firstof a|md %}` / `{% cycle a|md ... %}`.

## Severity: low (fail-**safe**)

This is **over-escaping**, not an XSS — a runtime-safe value gets escaped where Django would not. No security impact; just a correctness/parity gap on a rare combination (custom `mark_safe` filter *inside* `{% firstof %}`/`{% cycle %}`).

## Found by

The #1660 adversarial-verification workflow (variant-coverage lens). The two arms #1660 touches (Variable, InlineIf) are correct and XSS-safe; this is the one untested parallel surface.

## Fix sketch

`get_value` returns a bare `Value` (line ~1802 `return Ok(value)`), so it has no place to report safeness today. Properly fixing it means threading the `(Value, bool)` shape (à la `apply_filter_full_safe`) out of the pipe loop and into the `{% firstof %}`/`{% cycle %}` emit path's escape decision — `get_value` has many callers, so this is a broader signature change than #1660 warranted. Mirror the per-iteration `runtime_safe = produced_safe` pattern from the Variable arm.

Per the parallel-path-drift canon (CLAUDE.md #1646), this path is explicitly noted in a code comment at `renderer.rs:1793` so the drift is documented, not silent.

Issue #1660, opened 2026-05-29
Custom Django template filter returning `mark_safe(html)` is HTML-escaped by djust on both HTTP prerender and WS-mount paths (rc14)


johnrtipton:
## Summary

A Django template filter (`django.template.Library.filter`) that returns `mark_safe(html)` has its SafeString-ness stripped when the template is rendered by djust. The output reaches the client HTML-escaped (`&lt;p&gt;` instead of `<p>`), on **both** the HTTP prerender path **and** the WS-mount frame's `html` field. Vanilla Django renders the same filter / same template correctly.

This is annoying in the small (turns "render markdown via a filter" into "pre-render in the view"), but it's the kind of asymmetry that's hard for a downstream author to localize, because half of Django filter idioms quietly break.

## Reproduction

`apps/foo/templatetags/djust_md.py`:

```python
from django import template
from django.utils.safestring import mark_safe

register = template.Library()


@register.filter(name="djust_md")
def djust_md(value) -> str:
    if not value:
        return ""
    from djust._rust import render_markdown
    return mark_safe(render_markdown(str(value)))
```

`templates/meeting_detail.html` (mounted by a LiveView):

```django
{% load djust_md %}
<div id="diag-summary">{{ meeting.spark_summary|djust_md }}</div>
```

…where `meeting.spark_summary` is a multi-line Markdown string with `# Heading`, `*` bullets, paragraphs, etc.

### HTTP prerender (curl, no JS, no WS)

```
<div id="diag-summary" class="md-prose text-sm">&lt;p&gt;The team advanced app development, focusing on UI, ...&lt;/p&gt; &lt;h3&gt;Key Points&lt;/h3&gt; ...
```

### WS-mount frame `html` field

Captured by hooking `WebSocket.prototype` before `inst.disconnect(); inst.connect()`. The `mount` frame's `html` field carries the same escaped content (`diag-summary">&lt;p&gt;The team advanced...&lt;/p&gt;`).

So the escape happens server-side, on both code paths.

### Same template, replace `|djust_md` with the explicit `|safe` filter

`{{ rendered_html|safe }}` (where `rendered_html` was pre-rendered to a plain `str` in the view and stored on `self.meeting`): renders correctly on both paths. Browser DOM shows `<p>`, `<h3>`, `<ul>`, `<li>`, etc.

So:

- `|safe` (Django built-in, special-cased somewhere): ✅ honored
- `|<custom_filter>` returning `mark_safe(html)`: ❌ stripped

[...]

PR #1673, opened 2026-05-30
fix(templates): honor runtime SafeString from custom filters + XSS hardening (#1660)

johnrtipton:
## Summary

Closes **#1660**. A custom template filter that `mark_safe()`s its output **at runtime** (without `@register.filter(is_safe=True)`) was HTML-escaped on both the HTTP-prerender and WS-mount paths. djust's Rust renderer decided auto-escaping purely by filter **name** (static `is_safe` / built-in `safe_output_filters`), never by the filter **result's** runtime `SafeString`-ness — unlike Django's `render_value_in_context`, which escapes iff the *final* value lacks `__html__`.

## The fix (additive, Django-faithful)

- **`apply_custom_filter`** now reports whether the Python result carries `__html__`.
- **`apply_filter_full_safe(...) -> (Value, bool)`** threads that out; the built-in match is extracted into `apply_builtin_filter` so the runtime-safe and plain variants share **one dispatch table** (no parallel built-in-name list to drift, cf. #1640/#1646). `apply_filter_full` delegates — existing callers untouched.
- The **renderer** (Variable + InlineIf arms) tracks the **last** filter's runtime safeness and ORs it into the escape decision. Purely additive: a later plain-returning filter **re-taints** (`x|md|upper` → escaped), matching Django.

## 🔒 Security hardening (caught by adversarial review)

A 5-lens adversarial XSS-verification workflow found that trusting `__html__` **alone** is an XSS hole: Django stringifies any non-`str` value *before* the `__html__` check, and djust's `Value` extraction stringifies arbitrary objects via `__str__`. So a non-`str` object that advertises `__html__` but renders attacker-controlled HTML through `__str__` would reach output **unescaped**. **Confirmed exploitable** on the first cut (`{{ x|evil_obj }}` → `<script>alert(1)</script>` raw). Fixed by gating the runtime-safe marker on **str-subclass-ness** (`is_instance_of::<PyString>() && hasattr("__html__")`) — exactly Django's rule. Pinned by a non-`str`-`__html__` impostor regression test.

## Verification

- **Reproducer-first / gate-off** (Action #1200/#1468): the `*_1660` tests fail against unchanged Rust (3/5) and against `|| runtime_safe` removed — non-tautological.
- **5-lens adversarial workflow**: XSS hunter (→ the str-gate finding, fixed), re-taint hunter (mechanism sound, 48k renders 0 fail-open), regression hunter (none; 11 clean runs), variant-coverage (→ #1672), gate-off auditor (non-tautological). The workflow also reported "non-determinism" — **disproven**: it was the gate-off agent's concurrent `maturin develop` toggling the shared `.so`; on a stable build `-k 1660` is 10/10 + deterministic across all hash seeds.
- **Full suites green**: 4467 Python (`tests/ python/tests/`), full `cargo test --workspace --exclude djust_live`, clippy clean. Pre-push gate passed.

## Scope boundary (explicit, per #1646)

The parallel `{% firstof %}`/`{% cycle %}` `get_value` pipe path still over-escapes a runtime-safe value — **fail-safe** (over-escape, not XSS). Documented in-code at `renderer.rs` and tracked in **#1672**.

## Issue × file × test

| Issue | Files | Tests |
|---|---|---|
| #1660 | `filter_registry.rs` (capture `__html__` + str-gate), `filters.rs` (`apply_filter_full_safe`/`apply_builtin_filter`), `renderer.rs` (2 escape arms) | `*_1660` matrix + XSS-impostor guard in `tests/unit/test_rust_custom_filters_1121.py` |
| #1672 (follow-up) | `renderer.rs` boundary comment | — |


johnrtipton:
## Stage 7 Code Review — VERDICT: ✅ APPROVE

_(posted as a comment: GitHub blocks self-approve on own PR)_

**VERDICT: APPROVE** — correct, security-sound, exactly mirrors Django's `render_value_in_context`; tests are non-tautological (gate-off proven both for the feature AND the XSS str-gate). No 🔴 findings.

## Adversarial XSS cases run live (rebuilt ext, autoescape=true)

| # | Case | Output | Expected | Verdict |
|---|------|--------|----------|---------|
| 1 | `{{ name\|md }}` (runtime mark_safe) | `<em>Hi</em>` | not escaped | ✅ |
| 2 | plain custom filter, input `<script>` | `&lt;b&gt;&lt;script&gt;&lt;/b&gt;` | escaped | ✅ |
| 3 | **str subclass WITHOUT `__html__`** | `&lt;x&gt;x&lt;/x&gt;` | escaped (no `__html__`) | ✅ |
| 4 | str subclass WITH `__html__` | `<script>alert('substr')</script>` | trusted (real SafeString) | ✅ matches Django |
| 5 | **non-str obj `__html__`+evil `__str__`** | `&lt;script&gt;alert(&#x27;nonstr&#x27;)&lt;/script&gt;` | **escaped (XSS guard)** | ✅ **no XSS** |
| 9 | `x\|md\|upper` re-taint | `&lt;EM&gt;HI&lt;/EM&gt;` | escaped (last filter plain) | ✅ |
| 10 | `x\|upper\|md` | `<em>HI</em>` | not escaped (last filter safe) | ✅ |
| 11 | `md` in attr ctx | `<a title="<em>Hi</em>">` | not escaped | ✅ |
| 14 | InlineIf `{{ a if cond else name\|md }}` | `<em>Hi</em>` | not escaped (InlineIf arm hit) | ✅ |
| 15 | `{% firstof bad\|md %}` | `&lt;em&gt;Hi&lt;/em&gt;` | **over-escaped (fail-safe, #1672)** | ✅ no XSS |
| 16 | `{% cycle a\|md b %}` | `&lt;em&gt;Hi&lt;/em&gt;Yo` | over-escaped (fail-safe) | ✅ |
| 17 | non-str impostor in attr ctx | `&lt;script&gt;...&lt;/script&gt;` | escaped (XSS guard) | ✅ |

**Django parity confirmed** by calling vanilla `django.template.base.render_value_in_context` directly on the same borderline objects: case 4 → NOT escaped, case 5 → escaped, `mark_safe(raw)` → NOT escaped. The fix is a precise port of Django's documented escape semantics — no XSS bypass exists in djust that doesn't also exist in stock Django.

## Gate-off self-test (Action #1200/#1468) — non-tautological

**Feature gate-off** (`|| runtime_safe` → `|| (runtime_safe && false)` at both renderer sites 429/491, rebuilt): **3 of 6 `-k 1660` tests FAIL** — `test_runtime_safe_filter_output_is_not_escaped_1660`, `test_plain_then_runtime_safe_filter_is_not_escaped_1660`, `test_runtime_safe_filter_in_attribute_context_is_not_escaped_1660`. The 3 still-passing are negative guards (assert escaping happens) — correct, the fix never makes them unsafe.

**XSS str-gate gate-off** (`is_instance_of::<PyString>() &&` → `true &&` in `filter_registry.rs:365`, rebuilt): **`test_runtime_safe_requires_str_subclass_not_just_html_attr_1660` FAILS** with `<script>alert(1)</script>` reaching output UNESCAPED. This proves the str-subclass check is the load-bearing XSS guard and the impostor test catches its removal. **Tree restored + rebuilt with the fix; all 18 file tests + 296-test `djust_templates` cargo suite green.**

## Auto-reject scan — clean
- Only added `mark_safe(f"...")` is `tests/.../_md_runtime` (line 330), a test fixture replicating the #1660 repro shape (mirrors pre-existing `_bold_html`). Not a production XSS surface.
- No `@csrf_exempt`, no f-string logging, no bare `except: pass`, no `print()`, no `console.log`, no new Rust `unwrap()`.

## Citation check (#1197) — all verified
- `#1672` OPEN, title accurately describes the deferred `{% firstof %}/{% cycle %}` get_value over-escape. `#1660` OPEN.
- `djust_core lib.rs` ref → confirmed: `Ok(Value::String(ob.str()?.to_string()))` at `djust_core/src/lib.rs:230` (the `__str__` fallback the XSS comment cites).
- #1672 boundary comment sits correctly above `apply_filter_full` (plain variant) at `renderer.rs:1801`, inside `get_value` (def `renderer.rs:1765`).
- ROADMAP #1660 P1 row cleanly removed (single-line diff); no `1660` refs remain in ROADMAP.

## Completeness / parallel-path-drift (#1646)
The single-dispatch-table refactor (`apply_builtin_filter`) is the right structural cure — no parallel built-in-name list to drift. Crate-wide audit: the only two escape-emitting filter-chain arms (Variable + InlineIf) both use `apply_filter_full_safe`; the `get_value` path (firstof/cycle) deferred to #1672 is genuinely **fail-SAFE** (over-escape, verified cases 15/16), not XSS. Other `get_value` callers consume values numerically/for comparison, not escaped HTML emission — no missed under-escape path. Deferring #1672 is the correct call (it's a fail-safe DX papercut, not a security issue).

## 🟡 Non-blocking notes
- `{{ name|md if cond else b }}` (filter on the InlineIf *condition* branch) raises `Unknown filter: md if cond else b` — a **pre-existing parser quirk** unrelated to this PR (PR doesn't touch InlineIf parsing; the else-branch filter path case 14 works correctly). Out of scope.
- Stale-base check (#1450): `HEAD..origin/main` = 0 commits behind. ✅

Reviewed against a clean, fix-rebuilt tree.

johnrtipton:
## Stage 14 / Retrospective — PR #1673 (#1660)

**Quality: 5/5** — shipped a correct, security-hardened, Django-faithful fix; the process caught a real XSS *that this PR's own first cut introduced* before it could merge.

### What went well
- **The adversarial verification earned its cost.** The 5-lens XSS workflow found that the initial implementation (trust `__html__` alone) was **exploitable**: djust's `Value` extraction stringifies arbitrary objects via `__str__` (`djust_core/src/lib.rs:230`), so a non-`str` object advertising `__html__` with a malicious `__str__` reached output raw (`<script>alert(1)</script>`, confirmed live). Fixed to Django's actual rule (trust only a `str` subclass with `__html__`). Inspection-only review would very plausibly have rubber-stamped the loose `hasattr("__html__")` check.
- **Reproducer-first + gate-off (Action #1200/#1468)** held throughout: the `*_1660` tests fail against unchanged Rust *and* against the str-gate removed (the impostor test fails with `<script>` unescaped) — proven non-tautological by both the workflow's gate-off lens and the independent Stage-7 reviewer.
- **Scope discipline (#1079/#1646):** the parallel `{% firstof %}`/`{% cycle %}` `get_value` path over-escapes (fail-**safe**, not XSS) — fixed exactly what #1660 cites, documented the boundary in-code, filed #1672 for the remainder. The built-in match was extracted into a single `apply_builtin_filter` dispatch table so there's no parallel built-in-name list to drift.

### What could improve (a real process lesson)
- **Shared-artifact contamination in verification workflows.** One lens (the gate-off auditor) ran `maturin develop` to toggle the fix on/off while the *other four lenses* tested live against the **same** `.venv` `.so`. That made `{{ x|md }}` and the `-k 1660` tests flip pass→fail→pass across the auditor's rebuilds, which one lens reported as "non-deterministic / flaky escape decision" (a false HIGH-adjacent finding). It was **disproven** on a stable build: `-k 1660` is 10/10 consecutive + deterministic across all hash seeds. **Lesson:** when a multi-agent verification has an agent that mutates a shared build artifact (recompiled extension, regenerated bundle, migrated DB) while siblings read it, give that agent an isolated build (separate worktree / `target` dir / serialize it last) — otherwise its rebuilds masquerade as nondeterminism in the artifact under test. (Generalizes the #1646 parallel-path lesson to *parallel agents sharing one build output*.) The variant-coverage lens's ENV CAVEAT caught it, which is why the false finding didn't reach a verdict.

### Metrics
- 6 files (3 Rust, 1 Python test, CHANGELOG, ROADMAP); +6 regression tests (runtime-SafeString matrix + XSS-impostor guard).
- Verification: 5-lens adversarial workflow + independent Stage-7 reviewer; full Python suite 4467 passed; full `cargo test --workspace`; clippy clean; CI 12/12 green.
- 1 HIGH XSS caught pre-merge (self-introduced, self-caught); 1 false finding disproven; 1 follow-up filed (#1672).

`RETRO_COMPLETE`

Issue #1676, opened 2026-05-31
client.min.js: ReferenceError 'ie is not defined' breaks LiveView in production (terser mangle × #1635 IIFE wrap)


johnrtipton:
## Symptom

In production (`DEBUG=False`, so `client.min.js` is served), the browser throws:

```
client.min.js  Uncaught ReferenceError: ie is not defined
```

This crashes the LiveView client. Downstream symptom: a **WebSocket reconnect loop** — the WS is accepted server-side (`WebSocket /ws/live/ [accepted]` → `connection open`), then the crashed client immediately closes it (`connection closed`) and retries on every reload.

The **unminified `client.js` works fine** — so it only manifests in prod.

## Reproduces on

rc15 and rc16 (`python/djust/static/djust/client.min.js`). Surfaced on a live djustlive deploy.

## Root cause

`scripts/build-client.sh` minifies the concatenated bundle with `terser --mangle`.

`async function applyPatches(...)` (12-vdom-patch.js) is mangled to `async function ie(...)`, and its call sites become `await ie(...)`. After **#1635** (`fix(client): wrap client.js bundle in an IIFE to survive re-execution`) put the whole bundle inside one outer IIFE, a cross-module reference to `applyPatches` no longer resolves to the mangled `ie` in the scope where it's called — terser emits `await ie(...)` in a scope where `ie` was not declared. Unminified, the reference resolves via the shared outer-IIFE scope (and the sticky-module path is additionally `typeof`-guarded), so it's invisible without minification.

Evidence (minified bundle):
- `async function ie(e,t=null){...}` — the mangled `applyPatches` (1 declaration)
- `const s=await ie(e.patches)` / `await ie(c.patches)` — **unguarded** calls
- `globalThis.djust.applyPatches=ie`, `e._applyPatches=ie`

## Suggested fixes (pick one)

1. **terser config**: the `--mangle` pass is renaming across the nested-IIFE boundaries unsafely. Try `--mangle` with `toplevel=false` (default) — but the real issue is the cross-IIFE free reference. Safer: ensure the bundle is a single flat scope OR that cross-module APIs go through `djust.*` (not bare identifiers) so mangling can't break them.
2. **bundle structure**: after #1635's outer-IIFE wrap, make `applyPatches` (and any other cross-module function) an explicit property on the shared `djust`/`globalThis.djust` object at definition, and have consumers call `djust.applyPatches(...)` — bare cross-IIFE identifier references are what terser mangles inconsistently.
[...]

PR #1690, opened 2026-06-05
fix(client): #1688 — applyPatches cross-IIFE ReferenceError in client.min.js

### LINKED ISSUE #1688: client.min.js: Uncaught ReferenceError: applyPatches is not defined (cross-IIFE ref; recurrence of #1676)
## Summary
On **djust 1.0.0**, every LiveView page in production (`DEBUG=False`, which serves `client.min.js`) logs an uncaught error at load:

```
Uncaught ReferenceError: applyPatches is not defined
    at client.min.js:1:200872
    at client.min.js:1:201012
    at client.min.js:1:213146
```

(With `DEBUG=True` the readable `client.js` is served and does **not** throw — the `typeof` guard holds — so this only manifests in the minified production bundle. Same class as the closed #1676: *terser mangle × IIFE wrap*, recurring with a different symbol.)

## Root cause (from the unminified `client.js`)
- The VDOM-patch module defines `async function applyPatches(...)` (~L7381) and exposes it as `globalThis.djust.applyPatches` (~L7636) — scoped to **that** module's IIFE.
- A different module's IIFE (the sticky/embedded-lifecycle one, ~L14734) references **bare `applyPatches`** at ~L15029–15030:
  ```js
  if (typeof applyPatches === "function" && !djust._applyPatches) {
      djust._applyPatches = applyPatches;   // <-- out of scope here
  }
  ```
  In the unminified build the `typeof applyPatches` guard returns `"undefined"` (no throw, branch skipped → `djust._applyPatches` stays undefined). In `client.min.js`, terser's transform of that guard (cf. #1676's IIFE-wrap interaction) lets the bare reference evaluate and throw.

## Impact
Non-fatal — core LiveView (WebSocket connect, event dispatch, DOM patching via the in-scope `applyPatches`) keeps working; verified the app is fully interactive with the error present. But:
1. It's an alarming uncaught error in every production console.
2. It aborts the tail of that IIFE, so `emitChildMountedEvents` (child-mounted lifecycle for embedded/sticky views) never wires up.

## Suggested fix
In the sticky/embedded module, reference the exposed alias rather than the bare symbol:
```js
const _ap = globalThis.djust && globalThis.djust.applyPatches;
if (typeof _ap === "function" && !djust._applyPatches) djust._applyPatches = _ap;
```
or otherwise ensure the cross-module reference resolves post-minification (the same hardening applied for #1676).

## Repro
1. djust 1.0.0 app, `DEBUG=False` (so `PostProcessingMixin` injects `client.min.js`).
2. Load any page with a `dj-view` root.
3. Observe the console error at load.

Found while running djust 1.0.0 in a production Django app (Hetzner k3s deploy).

johnrtipton:
## Issue
Closes #1688. On djust 1.0.0, every LiveView page in production (`DEBUG=False`, minified bundle) logs `Uncaught ReferenceError: applyPatches is not defined` at load. Recurrence of the #1676 class (terser mangle × IIFE wrap), different manifestation.

## Root cause
`python/djust/static/djust/src/45-child-view.js` referenced the **bare** symbol `applyPatches` at two sites:
- `_applyScopedPatches` (`typeof applyPatches` guard + call)
- the `djust._applyPatches` expose block

`applyPatches` is declared inside `12-vdom-patch.js`'s own inner IIFE and published only as `globalThis.djust.applyPatches` (`12-vdom-patch.js:2299`). The bundle is a single outer IIFE but each module has its own inner IIFE, so the bare reference is out of module 45's scope. Unminified it silently no-ops (the `typeof` guard returns `"undefined"`, leaving `djust._applyPatches` **unwired** so `emitChildMountedEvents` never runs); terser-minified, the transformed guard lets the bare reference evaluate and **throw**.

**Empirically confirmed** via a jsdom reproducer: on the pre-fix bundle `djust.applyPatches` is a function but `djust._applyPatches` is `undefined`.

## Fix
Read the published alias `globalThis.djust.applyPatches` at both sites — minification-independent, and also restores the intended `_applyPatches` wiring. Rebuilt `client.js` / `client.min.js`.

## Tests
- **New:** `tests/js/min_bundle_applypatches_1688.test.js` — behavioral regression asserting `djust._applyPatches` is wired after load (+ a source guard against reintroducing the bare `typeof applyPatches` shape).
- **Gate-off verified:** reverting the `45-child-view.js` fix + rebuild makes the new test fail (`expected 'undefined' to be 'function'`).
- Updated `child_view.test.js` / `sticky_preserve.test.js` stubs from `win.applyPatches` → `win.djust.applyPatches`.
- Full JS suite green: **1648 passed**.

## Notes
- Non-fatal in production (core LiveView keeps working via the in-scope applier); the value is removing the prod-console error + restoring embedded/sticky child-mounted lifecycle.
- Behavior change (intended): unminified `djust._applyPatches` goes undefined→wired; `_applyScopedPatches` goes no-op→applies, matching the module's documented intent.


johnrtipton (commented):
## Stage 11 Review — PR #1690 (#1688) — APPROVE

Fix is correct and minification-independent. Both sites in `src/45-child-view.js` now read the published alias instead of the bare cross-IIFE symbol:
- `_applyScopedPatches`: `const _ap = globalThis.djust && globalThis.djust.applyPatches` (45-child-view.js:113), call via `_ap(...)` (line 121).
- expose block: `_topApplyPatches` guard + wire (45-child-view.js:335-337).

**Verified:**
- Ordering safe — module 12 publishes `globalThis.djust.applyPatches` at client.js:7636; module 45 runs at client.js:14706+. Alias is set by load time.
- Built artifacts in sync — `client.js` has the alias at both sites (14817, 15039), zero bare `typeof applyPatches` guards; `client.min.js` retains `function applyPatches(` (#1676 `--keep-fnames` guard intact).
- No bare `typeof applyPatches` in ANY src module (executable code; only docstrings mention the bare name).
- `console.warn` guarded by `globalThis.djustDebug` (45-child-view.js:115).
- Tests: 27/27 pass across the 4 suites. New test #1688 is non-tautological — the 3rd case is a source-grep pin, and the 2nd asserts `_applyPatches===applyPatches` which gate-off (bare symbol → guard fails) leaves undefined.

**Concern/question:** the regression test relies on jsdom not reproducing the minified throw (same limitation acknowledged for #1676). The source-grep guard mitigates regressions but a real terser-bundle smoke test would be stronger. Acceptable given #1676 precedent.

**Edge cases considered:** `globalThis.djust` undefined at module-45 load (short-circuit `&&` handles it); double-load idempotency (`!djust._applyPatches` guard preserved); minify-mangle of the alias chain (string-keyed, unmanglable).

Verdict: APPROVE.

johnrtipton:
## Retrospective — PR #1690 (#1688, v1.0.1 drain)

**Task:** Fix `client.min.js` `Uncaught ReferenceError: applyPatches is not defined` in production — recurrence of the #1676 terser-mangle × IIFE class.

### Quality: 5/5

### What went well
- **Reproducer-first (#1210) paid off immediately.** A throwaway jsdom script confirmed the reporter's exact symptom (`djust.applyPatches` is a function but `djust._applyPatches` is `undefined`) *before* any fix — which both validated the root cause and handed me a behavioral, non-tautological regression test for free.
- **Gate-off self-test (#254/#1468) at Stage 5.** Reverting the `45-child-view.js` fix + rebuild made the new test fail (`expected 'undefined' to be 'function'`), proving it exercises the change.
- **Symbol-migration grep (#1391) caught a second test file.** The full JS suite surfaced `sticky_preserve.test.js` test 8 (same bare-`win.applyPatches` stub pattern as `child_view.test.js`); grepping confirmed no other stubs remained.

- The #1676 fix (`--keep-fnames`) addressed *one* manifestation of the class but not the underlying anti-pattern (bare cross-IIFE references). #1688 is the same class resurfacing with a different symbol path. **Follow-up worth filing:** a build-time or lint guard that fails on any bare `applyPatches`/cross-IIFE function reference in `src/*.js`, so the class is mechanically retired rather than fixed one symbol at a time. (The new test's source-grep pin guards only `applyPatches`.)

### Relation to prior work
Direct sibling of #1676 (v1.0.0rc17, PR with `--keep-fnames`). Both are the "single-IIFE wrap × per-module inner IIFE × terser" failure mode. The parallel-path-drift lesson (v1.0.0rc14 canon) applies: the durable cure is structural (lint/guard), not N point fixes.

### Verified
- Full JS suite: **1648 passed** (142 files). Targeted suites: 27/27.
- All CI green (rust, python 3.12 + 3.14t, JS, playwright, security, CodeQL).
- Gate-off verified; bundle rebuilt and committed; `client.min.js` retains `function applyPatches(` (#1676 guard intact).

RETRO_COMPLETE

PR #1691, opened 2026-06-05
fix(templates): #1672 — honor runtime SafeString in firstof/cycle get_value path

### LINKED ISSUE #1672: templates: runtime-SafeString from a custom filter is over-escaped in {% firstof %}/{% cycle %} (get_value pipe path)

johnrtipton:
Closes #1672.

## Summary
Follow-up to #1660. A custom template filter that calls `mark_safe()` **at runtime** is correctly honored for `{{ var|filter }}` (the Variable + InlineIf render arms, fixed in #1660), but was **over-escaped** when used inside `{% firstof a|md %}` / `{% cycle a|md ... %}`.

**Severity LOW — fail-SAFE over-escaping, NOT an XSS.** A parity gap, not a security hole.

## Root cause
The `get_value` pipe helper (`crates/djust_templates/src/renderer.rs`) applied filters via the plain `apply_filter_full`, which discards the runtime-safe flag (`apply_filter_full` is just `apply_filter_full_safe(...).map(|(v,_)| v)`). The `{% firstof %}` / `{% cycle %}` emit arms then unconditionally `html_escape`d the result — so a runtime `SafeString` like `<em>Hi</em>` rendered as `&lt;em&gt;Hi&lt;/em&gt;`.

## Fix
- New `get_value_safe(expr, &Context) -> Result<(Value, bool runtime_safe)>` threads the runtime-safe flag out of the pipe loop via `apply_filter_full_safe`, mirroring the per-iteration `runtime_safe = produced_safe` pattern from the #1660 Variable arm (the LAST filter wins; a later plain filter re-taints).
- `get_value` is preserved as a thin wrapper (`get_value_safe(...).map(|(v,_)| v)`), mirroring the existing `apply_filter_full` / `apply_filter_full_safe` shape in `filters.rs` — so the **other 26 callers are untouched** (minimal blast radius).
- The `FirstOf` and `Cycle` emit arms call `get_value_safe` and skip `html_escape` when `runtime_safe`.
- Parallel-path-drift code comments added at all three sites per CLAUDE.md #1646.

### `get_value` callers
27 call sites total. Only the `FirstOf` / `Cycle` emit paths consume the safe flag (switched to `get_value_safe`). The other 26 — `InlineIf`, progress-bar math (839/840/841), condition operators (1666–1756), the recursive base resolve (1773), and the test call sites — keep the `get_value` signature and were **not** edited.

## Fail-safe guarantee (no XSS)
`runtime_safe` originates ONLY from `apply_filter_full_safe`, which returns `true` solely for a genuine `str`-subclass with `__html__` (the #1660 XSS-hardened check in `filter_registry.rs:365`). Built-ins always return `false`. The change is strictly additive: it can only ever mark MORE values safe — and only when the last filter actually produced a real `SafeString` — never under-escape a plain/attacker-controlled value. The `*_plain_filter_still_escaped` tests pin this.

## Tests + gate-off evidence
7 new regression cases in `TestFirstofCycleRuntimeSafe_1672` (`tests/unit/test_rust_custom_filters_1121.py`): firstof/cycle runtime-safe not escaped, second-arg fall-through, in-for-loop, plain-filter-still-escaped guards, and runtime-safe-then-plain re-taint parity.

- **Before fix**: `test_firstof_runtime_safe_filter_not_escaped` FAILED — `assert '<em>Hi</em>' in '&lt;em&gt;Hi&lt;/em&gt;'`.
- **After fix**: all 7 pass.
- **Gate-off (#254)**: replacing `if runtime_safe` with `if false && runtime_safe` in the firstof arm + rebuild → 2 firstof tests fail (`&lt;em&gt;...` over-escaped again); restored → all pass. Confirms the tests exercise the actual change, not a tautology.

## Test counts
- Rust `cargo test -p djust_templates`: 420 passed, 0 failed.
- Python `tests/unit/test_rust_custom_filters_1121.py`: 25 passed.
- Python `python/djust/tests/ -k "filter or safe or firstof or cycle or template or render or escape"`: 980 passed, 2 skipped.
- Python `tests/ -k "filter or firstof or cycle or template or render or escape or safe"`: 452 passed, 1 skipped.


johnrtipton (commented):
Stage 11 review — fix-1672-firstof-cycle-safestring (vs main, 0 behind).

APPROVE. Clean, minimal, fail-safe.

Security (critical check): `runtime_safe` originates ONLY from `apply_filter_full_safe` (`filters.rs:55`), which returns `true` solely for a genuine `str`-subclass with `__html__` (#1660 hardened check). Built-in filters return `false` (`apply_filter_full_safe` short-circuits via `apply_builtin_filter` → `(v, false)`). Every literal/context/raw-var branch in `get_value_safe` returns `(_, false)` (renderer.rs:1858-1885). So firstof/cycle can only mark MORE values safe, never under-escape. Confirmed: `test_firstof_plain_filter_still_escaped` + `test_cycle_plain_filter_still_escaped` keep `<script>` escaped; re-taint test (renderer.rs path via `md_runtime|upper`) escapes. Traced — cannot under-escape.

Blast radius: `get_value` is now a thin wrapper (renderer.rs:1782 → `get_value_safe(...).map(|(v,_)|v)`); the ~26 other callers are signature-unchanged. Only FirstOf (renderer.rs:866) + Cycle (renderer.rs:927) consume the bool. Good.

Tests non-tautological: I ran the gate-off myself — flipping FirstOf to `if false && runtime_safe` failed 2 tests (`&lt;em&gt;Hi&lt;/em&gt;`). Assertions distinguish escaped vs raw correctly.

cargo test -p djust_templates: 66 passed. pytest filter/safe/firstof/cycle: 7 + 49 passed.

Edge cases checked: plain var in firstof (escaped ✓), re-taint (✓), cycle-in-for-loop (✓), Null/unresolved (still `html_escape(name)` ✓).

Concern (non-blocking, out of scope): `{% firstof x|safe %}` still escapes — get_value_safe threads runtime_safe but NOT the name-based `safe_output_filters` list the Variable arm uses (renderer.rs:416). I verified this is PRE-EXISTING on main, not a regression, and #1672 is scoped to runtime mark_safe. Suggest a follow-up issue for the built-in-safe-filter parity gap in firstof/cycle.

APPROVE

johnrtipton:
## Retrospective — PR #1691 (#1672, v1.0.1 drain 2/7)

**Task:** Honor runtime-SafeString from a custom filter inside `{% firstof %}`/`{% cycle %}` (the `get_value` pipe path dropped the runtime-safe flag). Follow-up to #1660.

### Quality: 5/5

### What went well
- **Parallel-path-drift canon (#1646) drove the framing.** #1660 fixed the Variable/InlineIf arms; this was the explicitly-noted untested parallel surface (`get_value`). Implementer mirrored the #1660 `apply_filter_full_safe` pattern rather than inventing a new one (Stage-4 first-principles grep, #168).
- **Minimal blast radius.** Of 27 `get_value` callers, only FirstOf + Cycle switched to the new `get_value_safe`; `get_value` became a thin wrapper. The other 26 were left untouched — small, reviewable diff.
- **Gate-off verified by BOTH implementer and reviewer** independently (flip `if runtime_safe` → `if false && runtime_safe` → firstof tests fail). Non-tautological confirmed twice.
- **Security trace was the load-bearing review.** Both agents confirmed the change can only ever mark MORE values safe via the #1660-hardened `__html__` check — it can never UNDER-escape a plain/attacker value. Fail-safe by construction.

- Reviewer surfaced a **pre-existing** adjacent gap: `{% firstof x|safe %}` (name-based `safe_output_filters`, not runtime mark_safe) still escapes because `get_value_safe` doesn't consult the name-based list the Variable arm uses. Pre-existing on main, NOT a regression, out of #1672 scope. **Filed as a follow-up issue** so the parallel surface is tracked, not lost.

### Relation to prior work
Third in the #1660 → #1672 lineage of "thread runtime-safeness through render arm N." The durable cure would be a single shared safeness-threading helper all arms call; the follow-up issue captures the remaining name-based-filter parallel gap.

### Verified
- `cargo test -p djust_templates`: 420 passed. Python filter/safe/firstof/cycle suites: 980+452 passed. All CI green.

RETRO_COMPLETE

PR #1694, opened 2026-06-05
fix(theming): #1662 — break registry↔theme_packs/manifest import cycle

### LINKED ISSUE #1662: tech-debt: remaining registry↔theme_packs / registry↔manifest import SCC in djust.theming (latent CodeQL cyclic-import)
Surfaced while fixing the 7 flagged CodeQL `py/cyclic-import` alerts (#2351/#2357-#2362, PR #1661). After extracting `get_theme_config` to the leaf `_config.py`, the AST import graph (lazy + eager) still has a pre-existing SCC:

```
registry ↔ theme_packs   (theme_packs imports get_registry; registry imports theme_packs)
registry ↔ manifest      (manifest imports get_registry; registry imports manifest)
```

CodeQL has NOT flagged these (they weren't in the open-alert list when #1661 was fixed), so they were scoped out. But they're a latent cyclic-import that could surface as a future alert, and `test_theming_no_cyclic_import.py` currently only gates that the *flagged* modules (presets/manager/css_generator) are acyclic — not the whole package.

**Suggested fix** (same pattern as #1661): the back-edge is `theme_packs`/`manifest` → `registry.get_registry`, and `registry` → `theme_packs`/`manifest` for discovery. Extract the discovery entry points (or the registry-singleton accessor) so the dependency is one-directional, then tighten `test_theming_no_cyclic_import` to assert the ENTIRE theming package import graph is acyclic (not just the 3 flagged modules).

Low priority — no runtime impact (imports are lazy where needed); maintainability + pre-empting a future CodeQL alert.

johnrtipton:
Closes #1662.

## Verified cycle (#1345 — cited cause confirmed against current graph)

Tarjan over the AST import graph (lazy + eager, level-1 relative imports) of
`python/djust/theming/*.py` found exactly one SCC of size > 1 on `main`:
`{theme_packs, manifest, registry}`. Back-edges:

- `theme_packs.py:827,836,1219,1228` → `from .registry import get_registry`
- `manifest.py:161` → `from .registry import get_registry`
- `registry.py:136,142` → `from .theme_packs import DESIGN_SYSTEMS / THEME_PACKS`
- `registry.py:215` → `from .manifest import load_theme_manifests`

CodeQL never flagged these (separate from #1661's flagged SCC) but they are a
latent `py/cyclic-import`.

## Refactor (mirrors #1661's leaf `_config.py` extraction)

New leaf module **`_registry_accessor.py`** holds `ThemeRegistry` +
`get_registry` and imports **only stdlib**. `theme_packs` / `manifest` /
`presets` / `loaders` / `checks` / `manager` / `apps` now import `get_registry`
from the leaf — so their back-edges point at a true leaf, not `registry`.

Discovery (the only `registry → theme_packs`/`manifest` edges) stays in
`registry.py` as module-level functions and is installed as a hook via
`set_discovery_hook`, keeping the dependency one-directional:

```
registry ──> theme_packs / manifest / _builtin_presets / _config
theme_packs / manifest / presets / ... ──> _registry_accessor  (leaf, no out-edges)
```

`registry.py` re-exports `ThemeRegistry` / `get_registry` / `register_*` (same
object identity), so `from djust.theming.registry import …` keeps working.
**No runtime, behavior, or public-API change** — import-structure only.

## Tightened test + gate-off evidence (#254)

`test_theming_no_cyclic_import.py` gains `TestWholeThemingPackageAcyclic`:
- `test_whole_theming_package_import_graph_is_acyclic` — Tarjan over **all**
  theming modules, asserting **no SCC of size > 1** anywhere. The graph builder
  counts both `from .X import name` **and** `from . import X` edges (CodeQL
  counts both). Previously only `presets`/`manager`/`css_generator` were gated
  and the registry SCC was explicitly allowed.
- `test_registry_accessor_is_a_leaf` — leaf-purity gate.
- `test_get_registry_back_compat_from_registry_module` — re-export identity.

Gate-off (verified):
- Reverting `manifest`'s edge back to `.registry` → whole-package test FAILS
  with SCC `[manifest, registry]`.
- Re-adding a `from . import registry` fallback to the leaf → **both** the
  whole-package test AND the leaf-purity test FAIL with `['registry']`
  (this is the `from . import X` edge class the strengthened builder now catches;
  it surfaced — and was removed — during self-review).
- Restored → all 14 pass.

## Tests

- `pytest python/djust/tests/test_theming_no_cyclic_import.py` — 14 passed.
- `pytest python/djust/tests/ -k "theming or theme or registry or preset or manifest or pack"` — **1853 passed, 1 skipped**.
- Import-smoke: `import djust.theming; from djust.theming.registry import get_registry; from djust.theming import theme_packs, manifest` → `import OK`; `discover()` populates 9 packs / 12 themes / 63 presets; `get_registry is _registry_accessor.get_registry`.


johnrtipton (commented):
Stage 11 review — APPROVE.

Verified (PR branch, project venv + Rust):
- Identity (critical): registry.get_registry IS _registry_accessor.get_registry; ThemeRegistry same class; get_registry() returns same singleton. Re-export at registry.py:24-28, not shadow copies. All intra-pkg importers redirected to the leaf (manager.py:264/396/416/472, manifest.py:161, theme_packs.py ×4, checks.py:8, loaders.py:35).
- Cycle broken: leaf imports stdlib only; whole-package Tarjan SCC test green. Discovery (registry→theme_packs/manifest) stays one-directional via set_discovery_hook(_do_discover) at registry.py:142.
- Hook timing robust: importing the leaf forces parent djust.theming __init__ (line 22: from .registry import ...) which installs the hook BEFORE any leaf access. Confirmed _discovery_hook non-None even when only the leaf is imported first.
- Discovery populates: 63 presets / 12 themes / 9 packs.
- Non-tautological: gate-off (inject `from . import theme_packs` into the leaf) fails BOTH test_whole_theming_package_import_graph_is_acyclic and test_registry_accessor_is_a_leaf.
- Test-caller fix is a real method→module-level _load_theme_package(registry, name) adaptation, not masking — asserts same outcome.
- Suites: cyclic-import 14/14; broader theming -k filter 1853 passed, 1 skipped.

Concern (non-blocking): _registry_accessor.discover() sets _discovered=True even when _discovery_hook is None, so a later legitimate discover() cannot recover an empty registry. Unreachable today (parent __init__ always installs the hook) but the fail-soft swallows a future regression silently — consider not marking discovered when the hook is absent.

Edge cases OK: double-import idempotent (non-reentrant lock + hook writes dicts directly, no deadlock vs register_*); lazy-import/hook-not-installed covered.

APPROVE

johnrtipton:
## Retrospective — PR #1694 (#1662, v1.0.1 drain 4/7)

**Task:** Break the latent `registry ↔ theme_packs` / `registry ↔ manifest` import SCC in `djust.theming` (pre-empt a future CodeQL `py/cyclic-import` alert; follow-up to #1661).

### Quality: 5/5

### What went well
- **Verified the cited cause first (#1345).** This was a retro-filed issue carrying a hypothesis; the implementer ran Tarjan over the real AST import graph (eager + lazy edges) and confirmed exactly the cited SCC `{registry, theme_packs, manifest}` with concrete back-edges before designing the fix — no fixing a phantom.
- **Mirrored the proven #1661 pattern** (leaf-module extraction) rather than sprinkling lazy imports: `_registry_accessor.py` (stdlib-only) holds the singleton; discovery stays one-directional via a `set_discovery_hook` callback. Structural cure, not a band-aid.
- **Test tightened from 3 modules → whole package**, with two extra gates the reviewer valued: leaf-purity (the leaf may not import the package) and re-export-identity (`registry.get_registry IS _registry_accessor.get_registry`). Gate-off verified the SCC test fails when an edge is reintroduced.
- **Parallel-path discipline:** implementer caught + fixed internal-test callers of the moved `_load_theme_package` (method→function) instead of leaving them broken.

- Reviewer flagged a non-blocking hardening: `_registry_accessor.discover()` sets `_discovered=True` even when `_discovery_hook is None`, so a future regression that lost the hook would silently fail-soft (empty registry, no recovery). Unreachable today (the hook is installed at package import). Deliberately not blocked on — noted here; a one-line guard (`if self._discovery_hook is not None`) would close it if the area is touched again.

### Relation to prior work
Direct continuation of #1661's CodeQL cyclic-import cleanup. The leaf-extraction + acyclic-gate-test pair is now the established djust pattern for import-cycle fixes.

### Verified
- Whole-package acyclic test 14/14; theming sweep 1853 passed / 1 skipped; import-smoke OK (9 packs / 12 themes / 63 presets). Re-export identity confirmed. All CI green.

RETRO_COMPLETE

Issue #1056, opened 2026-04-25
tech-debt: Stage 4 plan template — 'API shape options considered' row


johnrtipton:
**Source**: Retro v0.7.2 (addendum to Action #124)
**Action Tracker**: row #141

Stage 4 plan template grows an 'API shape options considered' row for greenfield UX/API features (where the API shape isn't dictated by an existing design). Surface 2-3 options with explicit pros/cons before implementation. Pattern proven on PR #1007 (3-option radio API, picked B, 12/12 tests green first authoring pass).

---
_Filed by `/pipeline-retro --reconcile` on 2026-04-25 — Scope C dedup pass._

johnrtipton:
Shipped in PR #1226 (v0.9.1-7 cleanup batch). Comma-list `Closes #...` form parsed only the first ref — the rest closed manually. (This bites again; #1184's parenthesized form is documented in PR-checklist, but the comma-list-on-its-own-line form has the same auto-close failure.) 🤖 v0.9.1-7 cleanup batch closure.

Issue #380, opened 2026-02-24
VDOM stamper injects <!--dj-if--> comment inside HTML tag attributes, producing malformed HTML


johnrtipton:
## Bug Description

In djust 0.3.3rc2, the Rust VDOM stamper injects `<!--dj-if-->` comment anchor nodes **inside HTML opening tag attribute areas** when a `{% if %}` block appears between tag attributes. This produces malformed HTML that browsers cannot parse correctly.

## Steps to Reproduce

Use a Django template with `{% if %}` inside an HTML opening tag's attribute list:

```html
<option value="{{ p.slug }}" {% if p.slug == current_project_slug %}selected{% endif %}>{{ p.name }}</option>
```

or:

```html
<a class="sidebar-link {% if active_view == 'tasks' %}active{% endif %}" dj-navigate="/tasks/">Tasks</a>
```

## Actual Behavior

The rendered HTML comes out as:

```html
<option value="djust-orchestrator" <!--dj-if-->="">djust-orchestrator</option>
```

The `<!--dj-if-->` comment anchor is injected mid-tag, causing the `>` to be displaced and rendered as literal `&gt;` text content. The `selected` attribute is lost entirely.

For class attributes with `{% if %}`, similar malformation occurs.

## Expected Behavior

The stamper should produce valid HTML:

```html
<option value="djust-orchestrator" selected="">djust-orchestrator</option>
```

## Impact

- `<select>` elements break entirely: the browser console shows `A <select> tag was parsed within another <select> tag` because the malformed option tags cause the browser to auto-close the first `<select>`, then see a second one from page content
- VDOM patch failures cascade from the malformed DOM structure
- Any `{% if %}` used inside a tag's attribute area (including `class="... {% if %}..."`) is affected

## Environment

- djust version: 0.3.3rc2
- Django version: 6.0
- Python: 3.12

## Workaround

Move all conditional attributes outside the opening tag using a different template pattern, or use `{% if condition %}selected="selected"{% endif %}` in a way that avoids the stamper touching attribute areas. Currently working around this by restructuring all affected templates.

## Additional Notes

This regression appears to be introduced in 0.3.3rc2 — it was not present in 0.3.2. The `<!--dj-if-->` comment injection mechanism seems correct for text node contexts but is being applied incorrectly inside tag attribute areas.

PR #1695, opened 2026-06-05
feat(db): #1687 — DJUST_NOTIFY_DATABASE_URL override for LISTEN DSN

### LINKED ISSUE #1687: db.notifications: allow a dedicated LISTEN DSN, separate from DATABASES['default']
## Summary

`djust.db.notifications` always derives its LISTEN `AsyncConnection` DSN from `settings.DATABASES["default"]`, with no way to point the listener at a *different* endpoint. This blocks consumers who need to isolate the long-lived LISTEN connections from their request-path connection pool.

## Where (djust 1.0.0)

`djust/db/notifications.py::_build_dsn()`:

```python
def _build_dsn() -> str:
    db = settings.DATABASES.get("default", {})        # ← only source
    ...
    val = db.get(key) or os.environ.get(f"PG{dsn_key.upper()}", "")  # PG* only when DATABASES omits the key
    ...
```

The `PG*` env fallback can't override here — it loses to any value present in `DATABASES["default"]` (the `or` short-circuits). So the listener's `AsyncConnection` always targets the same host/port as the default DB.

## Why this matters (the real-world case)

A platform running the dashboard pattern (`@notify_on_save` + `self.listen()` across many models) holds N session-scoped LISTEN connections per live user. When the request path and the LISTEN path share one pgbouncer **session-mode** pool, those LISTEN slots compete with request connections — at enough concurrent dashboard users the pool saturates, new request connections block, and `/health` hangs → liveness failure → 503.

The standard fix is to split them: request path → pgbouncer **transaction mode** (small, high-churn), LISTEN path → a **dedicated DSN** (direct-to-Postgres or a separate session pool). The request-path half is deployment config, but the LISTEN-path half is **impossible today** because `_build_dsn()` offers no override — it's hard-wired to `DATABASES["default"]`.

(Concrete downstream: djustlive #380. We've migrated to psycopg3 and gated live LISTEN behind a flag, but can't actually turn it on safely until the LISTEN connection can be isolated — which needs this.)

## Proposed change

Have `_build_dsn()` prefer an explicit, optional setting before falling back to `DATABASES["default"]`. Something like:

```python
def _build_dsn() -> str:
    # Explicit override wins: lets a consumer route LISTEN to a dedicated
    # endpoint (e.g. direct-to-Postgres or a separate pool) so long-lived
    # LISTEN connections don't share the request-path pool.
    override = getattr(settings, "DJUST_NOTIFY_DATABASE_URL", "") or os.environ.get("DJUST_NOTIFY_DATABASE_URL", "")
    if override:
        return _dsn_from_url(override)          # parse a standard DATABASE_URL-style string
    db = settings.DATABASES.get("default", {})  # current behavior (unchanged fallback)
    ...
```

- **Backwards-compatible**: unset → exactly today's behavior (derive from `DATABASES["default"]`).
- The setting could accept a `DATABASE_URL`-style string (parsed via the same mechanism Django/`dj-database-url` use) and/or a `DATABASES` alias key (e.g. `DJUST_NOTIFY_DATABASE_ALIAS = "listen"`) — whichever fits the framework's conventions better.
- The engine check (`"postgresql" not in engine → DatabaseNotificationNotSupported`) should still apply to the override.

## Scope

Just the DSN-selection hook in `_build_dsn()` (+ a small parse helper + docs + a test that the override is preferred and the fallback is unchanged). The pool/endpoint provisioning itself is the consumer's deployment concern.

Reported from djustlive (psycopg3 migration + #380 connection-pool separation).

johnrtipton:
Closes #1687.

## Problem
`python/djust/db/notifications.py::_build_dsn()` always derived the long-lived `LISTEN` `AsyncConnection` DSN from `settings.DATABASES['default']`, with no override. That means the dedicated listener connection cannot be isolated from the request-path connection pool. Downstream pain: **djustlive #380** — pgbouncer session-pool saturation by the long-lived `LISTEN` connection causes `/health` to hang.

## Design
Add an optional `DJUST_NOTIFY_DATABASE_URL` override, preferred **before** the `DATABASES['default']` fallback:

```python
override = getattr(settings, "DJUST_NOTIFY_DATABASE_URL", "") or os.environ.get("DJUST_NOTIFY_DATABASE_URL", "")
if override:
    return _dsn_from_url(override)
# else: unchanged DATABASES['default'] behavior
```

- Setting **or** environment variable of the same name (mirrors the existing `PG*` env-fallback convention in `_build_dsn`).
- `DATABASE_URL` string shape (`postgres://user:pass@host:port/dbname`) — chosen over a `DATABASES`-alias key (which would force a second ORM-visible DB entry for a transport-only connection) per #1056.
- New `_dsn_from_url()` helper parses via `urllib.parse` — **no new dependency** (no `dj-database-url`). Returns the same space-separated libpq DSN string `_build_dsn` already produces, in the same key order, so override and fallback paths are byte-compatible. URL-encoded credentials are decoded.

## Backwards-compat guarantee
When the override is unset, `_build_dsn()` returns a DSN **byte-identical** to prior releases. Pinned by `test_unset_override_byte_identical_fallback` (exact-string assertion).

## Engine check on the override
The postgres-only check still applies to the override: a non-postgresql URL scheme (e.g. `mysql://`) raises `DatabaseNotificationNotSupported`. Driver-qualified schemes (`postgresql+psycopg://`) are normalized and accepted.

## Credential safety
`_dsn_from_url` / `_build_dsn` never log the URL or DSN — the value flows only into `psycopg.AsyncConnection.connect`. The error message on a bad scheme echoes only the *scheme*, never credentials. All existing logging in the module is `%s`-style and logs exceptions only. No `mark_safe`. Audited: zero DSN/password logging.

## Tests (`TestBuildDsnOverride`, 7 cases)
(a) setting override preferred over `DATABASES['default']`; (b) env-var override parsed correctly (host/port/dbname/user/password); (c) unset → byte-identical fallback; (d) non-postgres override scheme raises `DatabaseNotificationNotSupported`; + setting-beats-env precedence, URL-encoded-credential decode, driver-qualified scheme accepted.

**Gate-off (#254):** forcing `override = ""` made the 5 override-dependent tests fail (the byte-identical-fallback and direct-`_dsn_from_url` tests correctly stayed green), then restored — confirming the tests exercise the real branch.

Test run: `70 passed, 5 skipped` (PG-roundtrip integration skipped, no live PG) for `-k "notif or dsn or listen"`.


johnrtipton (commented):
Stage 11 review (read-only).

VERIFIED
- Branch 0 commits behind main. No dependency added (urllib stdlib).
- pytest -k "Override or dsn or notif or listen" -> 59 passed.
- Gate-off (if False and override): 5/7 override tests fail -> non-tautological. The 2 still-passing exercise _dsn_from_url directly + the byte-identical fallback, correctly independent of the override branch.

BACKWARDS-COMPAT (critical) PASS
- _build_dsn (notifications.py:137-139): override coalesces setting->env->""; `if override:` fully skips the new branch when unset. Fallback key order + PG* env fallback unchanged. test_unset_override_byte_identical_fallback pins the exact legacy string.

ENGINE CHECK (critical) PASS
- notifications.py:96-99 raises DatabaseNotificationNotSupported on non-pg scheme, before DSN build. mysql:// covered + manually confirmed.

CREDENTIAL SAFETY (critical) PASS
- No DSN/URL/password logged anywhere. Bad-scheme error echoes ONLY the scheme (notifications.py:98), never the URL/password (manually confirmed with mysql://user:secretpass@...). Reconnect logs (344/359) log exc, not the DSN; unchanged from main.

URL PARSING PASS
- %40->@ decode; missing port/dbname omitted; postgresql+psycopg:// normalized; IPv6 host=::1 (correct libpq keyword form).

CONCERN (non-blocking)
- Query params are silently dropped: postgres://h:5432/db?sslmode=require yields a DSN with no sslmode, and unix-socket ?host=/var/run/postgresql drops the host. Operators pointing the listener at a direct endpoint commonly need sslmode. Recommend a doc caveat ("query options not parsed; set them in DATABASES/libpq env") or appending parsed query items in a follow-up.

APPROVE

johnrtipton:
## Retrospective — PR #1695 (#1687, v1.0.1 drain 5/7)

**Task:** Add `DJUST_NOTIFY_DATABASE_URL` override so the `db.notifications` LISTEN connection can target a dedicated endpoint, separate from `DATABASES['default']` (downstream: djustlive #380 pgbouncer saturation → `/health` hang).

### Quality: 5/5

### What went well
- **API-shape options weighed (#1056).** Implementer compared URL-string vs DATABASES-alias-key vs per-key settings and picked the URL string (no ORM-visible second DB entry for a transport-only connection), mirroring the module's existing `PG*` env convention.
- **VERIFY LITERAL API CONTRACTS (#1243)** before coding: confirmed the exact libpq DSN string format/key order, the `DatabaseNotificationNotSupported` name, and the engine-check location — so the override and fallback paths produce byte-compatible DSNs.
- **Backwards-compat pinned by an exact-string test** (`test_unset_override_byte_identical_fallback`), and the override branch fully short-circuits when unset.
- **Credential safety treated as load-bearing** (it's the security surface here): both implementer and reviewer audited that no DSN/URL/password is ever logged and the bad-scheme error echoes only the scheme. Verified with a `mysql://user:secretpass@...` probe.
- **No new dependency** — `urllib.parse` only, per the manifesto's "own your stack."

- Reviewer caught that URL **query params are silently dropped** — `?sslmode=require` and unix-socket `?host=/var/run/postgresql`, both common for the direct-to-Postgres endpoint this feature targets. Per scope-discipline (#1079) I did NOT expand #1687's scope; filed a follow-up issue for query-param passthrough and the PR docs note the current host/port/db/creds-only coverage.

### Relation to prior work
Unblocks djustlive #380 (the LISTEN-path half of the pgbouncer request/listen pool split). The credential-non-logging discipline is the same trust-boundary concern as the Stage-7 "data reaching the client" checks, applied to logs.

### Verified
- `TestBuildDsnOverride` 7 cases; `-k notif/dsn/listen` 70 passed / 5 skipped (no live PG). Gate-off verified. All CI green.

RETRO_COMPLETE

Issue #1683, opened 2026-06-02
Example/demo templates use deprecated, non-functional `@click` (dead buttons) — migrate to `dj-click`


johnrtipton:
## Summary

Several example apps under `examples/demo_project/` render interactive controls with `@click="handler"`. The shipped client binds **`dj-click`**, not `@click`, so those buttons are **inert** — clicking them does nothing. New users who run the demos (or copy demo code) hit non-working UI. Surfaced while preparing 1.0 launch material.

## Root cause (verified)

1. The shipped `client.js` binds **`dj-click` only** — `getAttribute('dj-click')` / `closest('[dj-click]')` in `static/djust/src/09-event-binding.js`; **zero** `@click` bindings.
2. `@click` is **not** transformed at render. A `LiveView` with `<button @click="increment">` renders output HTML that *still contains* `@click="increment"` (confirmed by rendering a counter and inspecting the output) — so the client never binds it.
3. The `T001` system check already flags `@click`/`@input` as **deprecated** (hint: *"Use `dj-click` instead of `@click`"*). So `@click` is deprecated **and** non-functional, with **no runtime warning** — a demo just shows dead buttons.

## Affected — real issue (handler-invocation `@click` in rendered demo templates)

- `examples/demo_project/djust_demos/views/no_template_demo.py` (L55, L74–76, L91, L100, L147–148)
- `examples/demo_project/djust_demos/views/counter_demo.py` (L39, L45, L51)
- `examples/demo_project/djust_demos/views/components_demo.py` (L137–139), `component_demo.py` (L27), `optimistic_counter_demo.py` (L95, L100)
- `examples/demo_project/djust_demos/demo_classes/counter.py` (L30–32), `dropdown.py` (L28, L35)
- `examples/demo_project/djust_homepage/views.py`
- the older parallel set under `examples/demo_project/demo_app/views/` (`component_demo`, `components_demo`, `homepage`, `no_template_demo`, `optimistic_counter_demo`, `views_old.py`)
- top-level `examples/rust_components_demo.py`, `examples/range_component_demo.py`

Full set: `grep -rnE '@(click|input|change|submit|blur|focus)=' examples/ | grep -v node_modules`.

## Also stale (docs/docstrings showing the deprecated syntax)

- `python/djust/templatetags/live_tags.py:208` — docstring `Example:` uses `<form @submit="submit_form">` → should be `dj-submit`.
[...]

Issue #1060, opened 2026-04-25
tech-debt: Dogfood pass for new CLI tools before commit


johnrtipton:
**Source**: Retro v0.5.1
**Action Tracker**: row #145

Establish pattern: any CLI tool that reports on project state gets a dogfood pass against the demo project before commit. v0.5.1 `djust_typecheck` originally produced 230+ lines of false positives; dogfood pass against demo caught it before commit. Add as a one-line bullet in `docs/PULL_REQUEST_CHECKLIST.md` 'Code Quality' section.

---
_Filed by `/pipeline-retro --reconcile` on 2026-04-25 — Scope C dedup pass._

johnrtipton:
Shipped in PR #1226 (v0.9.1-7 cleanup batch). Comma-list `Closes #...` form parsed only the first ref — the rest closed manually. (This bites again; #1184's parenthesized form is documented in PR-checklist, but the comma-list-on-its-own-line form has the same auto-close failure.) 🤖 v0.9.1-7 cleanup batch closure.

PR #1697, opened 2026-06-05
feat(checks): #1602 — T015 detects legacy data-djust-root/view attributes

### LINKED ISSUE #1602: System check: detect legacy data-djust-root / data-djust-view attributes during 1.0 upgrade
## Summary

When upgrading an app from djust 0.x → 1.0, templates that still use `data-djust-root` / `data-djust-view` (the old attribute names) silently fail to connect. The existing system check `T012` fires generically ("template uses dj-* event directives but has no dj-view attribute"), but doesn't recognize that the user *does* have a view declaration — they're just using the legacy attribute names. The actual fix is mechanical (`data-djust-root` → `dj-root`, `data-djust-view` → `dj-view`) but the path from symptom to fix is not obvious.

## What I hit

Discovered while upgrading examples.djust.org from 0.3.0rc5 → 1.0.0rc11. The four example templates all had `<div data-djust-root data-djust-view="app.views.MyView">`.

Symptoms after upgrade:
- Pages render with HTTP mount fine
- No WebSocket connection
- Browser console: `[LiveView] No containers found! Your template root element needs: dj-view="app.views.MyView"`
- System check: `T012` fires saying templates use dj-* event directives but have no dj-view attribute

The console error is good — but it only shows up in a browser, not in CI or pre-deploy checks. The T012 system check fires at Django startup but doesn't hint that the template *does* have a view declaration, just using deprecated names.

## Suggested fix

Add a new system check (e.g. `T0XX` "legacy data-djust-* attributes detected") that scans templates for `data-djust-root` / `data-djust-view` and emits an explicit migration hint. Something like:

```
?: (djust.T0XX) myapp/templates/foo.html:23 -- legacy 'data-djust-view' attribute detected.
    HINT: djust 1.0 renamed root attributes — change 'data-djust-view' to 'dj-view'
    and 'data-djust-root' to 'dj-root'. The leading 'data-' prefix is no longer required.
```

If the framework wanted to be extra friendly, the runtime LiveView mount could also accept the legacy attributes for one release with a `DeprecationWarning`, to give downstream apps a soft-landing window. But a system-check-only fix would already be a big DX win.

## Why this matters

The fix is two-character (`data-djust-` → `dj-`) but the symptom (no WebSocket, generic T012) doesn't point at it. Anyone with a non-trivial number of templates and dj-style attributes will rediscover this the hard way. A dedicated check turns a 30-minute debugging session into a one-line system-check message.

## Related

- The console error in `client.js` is great and should stay.
- T012 is also good but a more specific check would pre-empt the false-positive ("template has no view... oh wait it does, just spelled wrong").

johnrtipton:
Closes #1602.

## Why T012 wasn't enough
When a template still uses the pre-1.0 root attributes `data-djust-root` /
`data-djust-view`, the generic **T012** ("dj-* directives but no dj-view")
fires — but it doesn't recognise that a view IS declared, just with the
deprecated spelling. So the developer sees a generic "no dj-view" warning
even though they DID add a view, and the path from symptom (the LiveView
never connects over WebSocket) to fix is non-obvious.

## The T015 check
New Templates-family system check (next free id = **T015**) that scans user
template files and emits a Warning naming the rename explicitly:

```
?: (djust.T015) myapp/templates/foo.html:1 -- legacy 'data-djust-view' attribute detected.
   HINT: djust 1.0 renamed root attributes — change 'data-djust-view' to 'dj-view'
   and 'data-djust-root' to 'dj-root'. The leading 'data-' prefix is no longer
   required. Suppress this check with DJUST_CONFIG = {'suppress_checks': ['T015']}.
```

- Line-level reporting (matches T014's granularity), one finding per
  offending occurrence, with a `fix_hint`.
- Regex scoped via negative-lookahead `data-djust-(root|view)(?![\w-])` to
  **exactly** `data-djust-root` / `data-djust-view` — legitimate siblings
  (`data-djust-embedded`, `data-djust-activity`, `data-djust-view-model`,
  `data-djust-rooted`, …) never false-match.
- Suppressible via `DJUST_CONFIG = {'suppress_checks': ['T015']}` (mirrors
  the existing `_is_check_suppressed` pattern).

**Scope:** static system check ONLY. No runtime acceptance of legacy
attributes / DeprecationWarning (out of scope — separate larger change).

## Tests (`TestT015LegacyRootAttrs`, 7 cases)
- (a) `data-djust-view` + `data-djust-root` → 2 T015 findings; `data-djust-root`
  alone → 1
- (b) modern `dj-view`/`dj-root` → 0 findings
- (c) message includes file relpath + offending attr name + line number
- (d) suppression via `suppress_checks: ['T015']` → 0 findings
- (e) false-positive guard: `data-djust-embedded` / `data-djust-view-model` /
  `data-djust-rooted` → 0 findings
- migration-hint test asserts the hint names `dj-view`, `dj-root`, and `data-`

## Verification evidence
- **Empirical canary (#252):** ran the real `check_templates` machinery against
  a synthetic `<div data-djust-root data-djust-view="...">` template → 2 T015
  findings, hint correct. `CANARY_PASSED`.
- **Gate-off (#254):** neutralized the helper (`if True: return`); the 4
  behavior-meaningful T015 tests FAIL; restored → all pass.
- **Dogfood (#1060):** ran the check against `examples/demo_project` (migrated
  to `dj-*` in #1683) → **0 spurious T015**.
- `pytest python/tests/test_checks.py` → 200 passed; broader `-k "check or
  T01 or template"` subset → 686 passed, 4 skipped.

Two-commit shape: `feat` (impl + tests) / `docs` (system-checks.md + CHANGELOG).


johnrtipton (commented):
Stage 11 — APPROVE. Branch not stale (0 behind main). Scope correct: static-check-only (checks.py, test_checks.py, system-checks.md, CHANGELOG); no runtime legacy-attr acceptance added.

Empirical canary (run against real `check_templates`, not just the regex):
- Trigger `<div data-djust-root data-djust-view="...">` → exactly 2 T015 with correct per-occurrence fix_hint (checks.py:3415).
- Near-misses ZERO: data-djust-embedded, data-djust-view-model, data-djust-activity, data-djust-rooted, dj-view/dj-root. The `(?![\w-])` lookahead (checks.py:2504) is load-bearing and verified — it excludes `-`, so data-djust-view-model does not match data-djust-view.
- T012 still fires independently on a dj-*/no-dj-view template — no scan-loop interference (check_templates call site checks.py:3108).
- Suppression `DJUST_CONFIG={"suppress_checks":["T015"]}` silences it (_is_check_suppressed accepts short + qualified ids).
- Gate-off self-test: early-return in _check_legacy_root_attrs → 4 behavior tests fail, 3 negatives still pass. Non-tautological; tests assert on e.id, msg relpath/attr/line, and hint naming the rename.
- T015 id free (no collision on main). `pytest python/tests/test_checks.py`: 200 passed.

Concern (non-blocking): regex is case-sensitive + context-blind — `DATA-DJUST-VIEW` (HTML attrs are case-insensitive) is missed; `data-djust-view` inside an HTML comment or JS string false-fires. Both consistent with sibling check T014. Acceptable; consider a docs note under "False positives".

APPROVE

johnrtipton:
## Retrospective — PR #1697 (#1602, v1.0.1 drain 6/7)

**Task:** Add system check T015 to flag legacy `data-djust-root`/`data-djust-view` attributes during a 0.x→1.0 upgrade, with an explicit rename hint (the generic T012 didn't recognize a mis-spelled view declaration).

### Quality: 5/5

### What went well
- **Empirical canary at BOTH stages (#252/#1459).** Implementer AND reviewer independently ran the real `check_templates` machinery against a synthetic `data-djust-root data-djust-view` template (fires 2× T015) and against near-misses (`data-djust-embedded`, `data-djust-view-model`, `data-djust-activity`, modern `dj-view`) — zero false positives. A tooling PR's central claim ("catches X, not Y") was proven, not trusted.
- **The load-bearing design catch was the regex scope.** The codebase has many legitimate `data-djust-*` attrs; the negative-lookahead `data-djust-(root|view)(?![\w-])` is what prevents `data-djust-view-model` from matching `data-djust-view`. Both agents verified it explicitly.
- **Mirrored the T014 sibling pattern exactly** (#1243) — id selection, suppression via `_is_check_suppressed`, per-occurrence Warning + hint — and confirmed T012/T013/T014 still fire independently in the shared scan loop.
- **Dogfood (#1060):** ran against the demo project → 0 spurious T015 (clean because #1683 had just migrated the demos to `dj-*` — nice cross-drain synergy).

- Reviewer noted the regex is case-sensitive + context-blind: `DATA-DJUST-VIEW` (HTML attrs are case-insensitive) is missed, and the attribute inside an HTML comment / JS string would false-fire. **This is consistent with the entire T0xx template-scan family (incl. T014)** — a known limitation of the regex-based approach, not a T015 regression — so not blocked or separately filed. A future hardening of the whole family (HTML-aware scanning) would address all of them at once.

### Relation to prior work
Completes the 1.0-upgrade DX surface alongside #1683 (the demos this check would flag were fixed earlier in the same drain). T012→T015 is the same "specific check beats generic" pattern the issue argued for.

### Verified
- `test_checks.py` 200 passed; new `TestT015LegacyRootAttrs` 7 cases. Empirical canary + gate-off + dogfood all green. All CI green.

RETRO_COMPLETE

PR #1709, opened 2026-06-05
fix(templates): #1692 — honor name-based safe_output_filters in firstof/cycle

### LINKED ISSUE #1692: tech-debt: {% firstof %}/{% cycle %} ignore name-based safe_output_filters (e.g. x|safe)
Follow-up from PR #1691 (#1672) Stage 11 review.

#1672 fixed **runtime** `mark_safe()` filter output being over-escaped inside `{% firstof %}`/`{% cycle %}` by threading the runtime-safe bool through the new `get_value_safe` path. A **parallel** gap remains, surfaced by the reviewer:

`{% firstof x|safe %}` (and `{% cycle x|safe ... %}`) still escape, because `get_value_safe` does NOT consult the **name-based** `safe_output_filters` whitelist that the Variable render arm uses at `crates/djust_templates/src/renderer.rs:~416`. It only honors *runtime* `__html__`/str-subclass safeness (the #1660 path).

- **Severity:** low, fail-safe (over-escaping, not XSS).
- **Pre-existing:** confirmed on `main` before #1672 — NOT a regression introduced by #1691.
- **Fix sketch:** in `get_value_safe`'s filter loop, also set `runtime_safe` when the applied filter name is in `safe_output_filters` (e.g. `safe`, `urlize`, `urlizetrunc`, `unordered_list`), mirroring the Variable arm. Add a parallel-path-drift comment (#1646).
- **Parent lineage:** #1660 → #1672 → this. The durable cure is a single shared safeness-threading helper all render arms call.

johnrtipton:
Closes #1692.

## Lineage: #1660 → #1672 → #1692

- **#1660** threaded runtime `mark_safe()`-ness through the `{{ var|filter }}` **Variable** / **InlineIf** render arms.
- **#1672** (PR #1691) added `get_value_safe` to thread that same RUNTIME safeness through the `{% firstof %}` / `{% cycle %}` emit path.
- **#1692** (this PR) closes the remaining gap: `get_value_safe` did **not** consult the **NAME-based** `safe_output_filters` whitelist (`safe`, `safeseq`, `force_escape`, `json_script`, `urlize`, `urlizetrunc`, `unordered_list`) that the Variable arm uses. So `{% firstof x|safe %}` and `{% cycle x|urlize %}` (urlize emits its own `<a href=…>` HTML) were still over-escaped in those two tags.

## Fix

In `get_value_safe`'s filter loop, the value is now also marked safe when the applied filter **NAME** is in the whitelist (or is a custom `is_safe=True` filter) — mirroring the Variable/InlineIf arms exactly:

```rust
runtime_safe = produced_safe
    || SAFE_OUTPUT_FILTERS.contains(&filter_name)
    || crate::filter_registry::is_custom_filter_safe(filter_name);
```

## Shared whitelist (parallel-path-drift #1646)

The whitelist previously existed as **two inline copies** (Variable arm + InlineIf arm). Rather than add a third divergent copy, it was **hoisted to a single module const** `SAFE_OUTPUT_FILTERS` that all three render paths now reference — one source of truth.

## Fail-safe

The change is strictly additive (same property as #1672): it only ever **ADDS** safeness for the established whitelisted names or genuine runtime SafeStrings. A plain/unknown filter (e.g. `upper`) stays escaped, and LAST-filter re-taint semantics are preserved (`{% firstof x|safe|upper %}` re-escapes because `upper` is last and not safe).

## Tests + gate-off

- 4 Rust cases in `crates/djust_templates/src/renderer.rs` (`renderer::tests`): firstof|safe not double-escaped, cycle|urlize not double-escaped, firstof|upper still escaped, firstof|safe|upper re-taints.
- 4 Python cases in `tests/unit/test_rust_firstof_cycle_named_safe_1692.py` (same matrix via `djust._rust.render_template`).
- **Gate-off (#254)**: neutralizing the new name-based branch (`(false && …)`) makes the two safe-filter tests fail while the two fail-safe escaping tests keep passing — proving the new tests exercise the actual change and aren't tautological.

`cargo test -p djust_templates`: 300 passed. Targeted pytest (`-k "filter or safe or firstof or cycle or template or escape or urlize"`): 670 passed, 3 skipped.


johnrtipton (commented):
Stage 11 review — APPROVE.

**Cannot under-escape (critical):** `get_value_safe` sets `runtime_safe = produced_safe || SAFE_OUTPUT_FILTERS.contains(&filter_name) || is_custom_filter_safe(filter_name)` (renderer.rs:1859-1861). A plain/unknown filter (`upper`) → all three false → escapes. FirstOf/Cycle arms (renderer.rs:864-867, 927-932) escape by default and only skip when `runtime_safe`. Confirmed by `test_firstof_nonsafe_filter_still_escaped` (`&lt;B&gt;...`).

**Hoist exact:** the diff shows both removed inline arrays were byte-identical 7-element lists, now `SAFE_OUTPUT_FILTERS` (renderer.rs:24). All three call sites reference it (:436, :489, :1860); grep confirms no divergent copy.

**Last-filter semantics:** `runtime_safe = ...` (assign, not `|=`) per iteration → `{% firstof x|safe|upper %}` re-taints + escapes (test passes).

**#1672 intact:** `produced_safe` term preserved. **Gate-off verified:** neutralizing to `runtime_safe = produced_safe;` makes exactly the 2 safe-filter tests FAIL, the 2 escaping tests PASS — non-tautological. Restored.

**Tests:** `cargo test -p djust_templates` all green (300/66/+); python `-k firstof|cycle|safe|urlize` 126 passed, 0 failed; new 1692 file 4/4.

**Concern (minor, Action #1543):** whitelist has 7 names but firstof/cycle only exercise `safe`+`urlize`. `unordered_list`/`json_script`/`safeseq`/`force_escape` go untested on this path. Mechanical mirror via one shared const, low risk — but a per-name pin would lock the variant coverage.

Edge cases OK: empty firstof/cycle → `""`; urlize self-escapes surrounding text; custom `is_safe=True` via `is_custom_filter_safe`.

APPROVE

johnrtipton:
## Retrospective — PR #1709 (#1692, v1.0.2 drain 1/6)

**Task:** Make `{% firstof %}`/`{% cycle %}` honor the NAME-based `safe_output_filters` whitelist (`x|safe`, `x|urlize`), completing the #1660→#1672 lineage (#1672 handled only the runtime-`mark_safe` path).

### Quality: 5/5

### What went well
- **The fix retired drift instead of adding to it (#1646).** The whitelist existed as TWO byte-identical inline copies (Variable arm + InlineIf arm); rather than add a divergent third copy in `get_value_safe`, the implementer hoisted it to a single module const `SAFE_OUTPUT_FILTERS` that all three consumers reference. The reviewer diffed the old inline lists against the const to confirm no name was added/dropped in the refactor.
- **Last-filter semantics preserved:** per-iteration `runtime_safe = …` (assign, not `|=`), so `{% firstof x|safe|upper %}` correctly re-escapes (upper is the terminal filter).
- **Cannot-under-escape was the load-bearing check, verified by both agents.** `runtime_safe` is true only for whitelisted names, custom `is_safe=True`, or the #1672 runtime path — a plain `upper`/`default` still escapes. Gate-off (neutralize the new branch → exactly the 2 safe-filter tests fail, the 2 escaping tests pass) was reproduced independently by the reviewer.

- Reviewer noted the multi-variant-coverage gap (#1543): the const has 7 names but only `safe`+`urlize` are exercised on the firstof/cycle path (`unordered_list`/`json_script`/`safeseq`/`force_escape` untested there). Low risk — it's a mechanical mirror through one shared const, so behavior is identical across names — but per-name pins would lock it. Not blocked; noted for if the area is touched again.

### Relation to prior work
Third and (for this surface) final entry in the #1660→#1672→#1692 "thread safeness through render arm N" lineage. The hoist means a FOURTH arm, if one is ever added, inherits the whitelist for free — the lineage's structural endpoint.

### Verified
- `cargo test -p djust_templates` 300 passed; Python firstof/cycle/safe/urlize sweep 670 passed / 3 skipped; new `test_rust_firstof_cycle_named_safe_1692.py` 4/4. Gate-off verified twice. All CI green.

RETRO_COMPLETE

PR #1711, opened 2026-06-05
feat(db): #1696 — pass through known-safe libpq query params in DJUST_NOTIFY_DATABASE_URL

### LINKED ISSUE #1696: feat(db): DJUST_NOTIFY_DATABASE_URL drops query params (sslmode, unix-socket host)
Follow-up from PR #1695 (#1687) Stage 11 review.

#1687 added `DJUST_NOTIFY_DATABASE_URL` to isolate the LISTEN connection on a dedicated endpoint. `_dsn_from_url()` currently parses scheme/user/password/host/port/dbname but **silently drops the URL query string**.

For the exact use case #1687 targets — pointing the listener at **direct-to-Postgres** (bypassing pgbouncer) — operators commonly need:
- `?sslmode=require` (direct endpoints frequently enforce TLS)
- `?host=/var/run/postgresql` (unix-socket form)

These are dropped, so the resulting libpq DSN has no `sslmode`/socket override and the connection may fail or fall back to a weaker mode than intended.

- **Severity:** low-medium — backwards-compatible (no override = unaffected), but a likely footgun for the primary use case.
- **Fix sketch:** in `_dsn_from_url` (`python/djust/db/notifications.py:74`), pass through known-safe libpq query items (`sslmode`, `sslrootcert`, `host`, `application_name`, `connect_timeout`, …) appended as `key=value` DSN pairs. Decode/escape values consistently with the existing fields.
- **Interim:** PR #1695's docs note the URL covers host/port/db/credentials only; this issue tracks query-param passthrough.

johnrtipton:
Closes #1696. Follow-up to #1687 / PR #1695.

## Problem
`_dsn_from_url()` (added in #1687) parsed scheme/user/password/host/port/dbname from
`DJUST_NOTIFY_DATABASE_URL` but silently **dropped the URL query string**. The two most
common direct-to-Postgres LISTEN needs were therefore impossible to express:
- `?sslmode=require` (TLS)
- the unix-socket form `?host=/var/run/postgresql`

## Fix
`_dsn_from_url()` now appends a **known-safe allowlist** of libpq connection params from
the URL query string to the produced DSN:

`sslmode`, `sslrootcert`, `sslcert`, `sslkey`, `host`, `application_name`, `connect_timeout`

- Values are percent-decoded consistently with the userinfo fields.
- Values containing whitespace are **libpq-quoted** (`'…'` with backslash escaping), e.g.
  `application_name='djust listener'`, so the second word isn't parsed as a new keyword.
- Params are appended after the core fields in stable (sorted) order — deterministic DSN.

## `host`-override precedence decision
A `?host=` query item **REPLACES** the URL netloc host (it does not append a second
`host=`). Rationale: appending two `host=` keys relies on libpq "last value wins", which
is fragile and produces a confusing DSN. Replacing is deterministic and matches the
unix-socket intent — `postgres://u:p@placeholder/db?host=/var/run/postgresql` routes to
the socket; the netloc host is an ignored placeholder. The output carries exactly one
`host` key.

## Backwards-compat
A **no-query URL produces a DSN byte-identical** to the #1695 output (regression test).

## Credential safety
- The URL/DSN/password is **never logged** (unchanged).
- `user` / `password` / `dbname` are deliberately **NOT** in the allowlist, so a query
  string can never override the URL-derived credentials (e.g. a stray `?password=evil` is
  silently dropped — tested).
- The `host` override only redirects to a host the same URL author already controls — no
  privilege escalation. No new dependency (`urllib.parse` only).

## Tests + gate-off (#254)
New `TestDsnQueryParams` (8 cases): sslmode passthrough; unix-socket host override;
unknown key ignored; no-query byte-identical; multiple allowlisted params; percent-decode;
whitespace libpq-quoting; cannot-override-credentials. Gate-off verified — neutralizing the
passthrough (`if False`) makes the 6 behavior tests FAIL; the 2 negative tests still pass.
Full subset: `pytest -k "notif or dsn or listen or db"` → 429 passed, 6 skipped.


johnrtipton (commented):
Stage 11 review — verified by running the branch (worktree off the PR head), not inspection.

**Security (critical) — PASS.** `_dsn_from_url("postgres://u:p@h:5432/db?password=evil&user=admin&dbname=other&sslmode=require")` → `host=h port=5432 dbname=db user=u password=p sslmode=require`; `evil`/`admin`/`other` absent. `user`/`password`/`dbname` are not in `_DSN_QUERY_ALLOWLIST` (`notifications.py:78`), so credentials are unforgeable.

**host precedence — PASS.** `?host=/var/run/postgresql` → exactly one `host=` (socket); netloc host replaced via `query_items.pop("host")` (`notifications.py:146`).

**Byte-identical — PASS.** No-query output unchanged; `_dsn_quote` (`notifications.py:166`) leaves bare values untouched.

**libpq quoting — PASS.** `?application_name=djust listener` → `application_name='djust listener'`.

**Non-tautology — confirmed.** Gated off (neutralized allowlist + host): the 6 positive tests FAIL, the 2 negatives (byte-identical, cannot-override-creds) PASS. 429 passed / 0 failed for `-k "notif or dsn or listen or db"`.

**Edge cases (all sane):** repeated `?host=a&host=b` / repeated `sslmode` → last wins (dict); blank `?host=` → dropped, falls back to netloc; trailing `?` → clean; percent-encoded socket decodes.

**Concern (minor, non-blocking):** allowlist match is case-sensitive, so `?SSLMODE=require` is silently dropped. libpq keywords are lowercase so this is correct, but a typo'd uppercase key vanishes with no warning — same silent-drop UX as unknown keys; worth a one-line doc note. `sslmode=disable` passes through, acceptable (operator config).

APPROVE

johnrtipton:
## Retrospective — PR #1711 (#1696, v1.0.2 drain 3/6)

**Task:** `DJUST_NOTIFY_DATABASE_URL` (#1687) dropped the URL query string; pass through a known-safe libpq allowlist (`sslmode`, socket `host`, etc.) so the direct-to-Postgres LISTEN endpoint can be fully configured.

### Quality: 5/5

### What went well
- **Allowlist, not passthrough — and credentials are explicitly NOT on it.** `user`/`password`/`dbname` are dropped, so a query string can never override URL-derived credentials. Both agents verified with an adversarial URL (`?password=evil&user=admin&dbname=other`) that the overrides vanish and the URL creds win. That's the load-bearing security property and it was tested, not assumed.
- **Deterministic `host`-replace** (query `host` replaces the netloc host → exactly one `host=` in the DSN) rather than relying on libpq "last-wins" with two appended keys — removes ambiguity for the unix-socket case.
- **Implementer found a latent correctness bug while implementing:** query values with whitespace (`application_name=djust listener`) would produce an invalid DSN; added `_dsn_quote()` (libpq single-quote/backslash escaping). The reviewer confirmed it.
- **Reviewer ran the branch in a worktree** (copying the compiled extension in) rather than inspecting — every claim (credential-block, host-replace, byte-identical, quoting) was executed.

- Minor non-blocking: the allowlist match is case-sensitive, so `?SSLMODE=require` is silently dropped. libpq keywords are lowercase so it's technically correct, but a typo'd uppercase key vanishes with no warning — a one-line doc note would help. Noted, not blocked.

### Relation to prior work
Closes the follow-up the #1687 (PR #1695) Stage-11 review opened — the "drain follow-ups promptly and they get actioned" loop from the v1.0.1 retro, now demonstrated end-to-end (filed #1696 at #1695 review → drained here).

### Verified
- `-k "notif or dsn or listen or db"` 429 passed; new `TestDsnQueryParams` 8 cases; gate-off verified by both agents (credentials-cannot-override is a non-tautological negative). All CI green.

RETRO_COMPLETE

Issue #1244, opened 2026-05-01
tech-debt: Stage 7 self-review — cross-ref workflow header claims against step semantics


johnrtipton:
Source: v0.9.2-1 retro / PR #1241 review.

## Context

PR #1241's `.github/workflows/retro-gate-audit.yml` had a header docstring claiming the workflow would "surface findings as workflow annotations in the Actions UI; no auto-issue creation to avoid spam during steady-state operation" — implying the workflow run itself would stay green and only the annotations would signal flagged PRs.

In practice, the audit script (`scripts/audit-pipeline-bypass.py`) exits 1 when it finds flagged PRs. Under GHA's default `bash -eo pipefail`, `python ... | tee` propagates the non-zero exit and the entire workflow run goes red. Header claim → red runs every flagged day, which is the opposite of the documented intent.

Stage 11 caught this; Stage 7 (self-review) and Stage 8 (security check) didn't, because both look at structural properties (auto-reject triggers, permissions, missing tests, etc.) rather than behavioural contracts. A header docstring describing intent IS a behavioural contract.

## Proposed change

Add to `.pipeline-templates/{feature,bugfix}-state.json` Stage 7 checklist (when changed files include `.github/workflows/*.yml` or any file with a header docstring describing runtime behavior):

```json
{
  "action": "WORKFLOW-HEADER CROSS-REF — for every new/modified file with a docstring or header comment describing runtime behavior, list every behavioral claim in the docstring and verify each one matches the implementation. Behavioral claims include: 'silent in normal operation', 'green run', 'no false positives', 'idempotent', 'safe to run repeatedly', etc. Cross-ref against actual step semantics (pipefail, exit codes, conditional triggers).",
  "done": false,
  "mandatory": true
}
```

## Acceptance

- Stage 7 self-review checklist gets the workflow-header cross-ref item.
- The item runs whenever a workflow file is touched.
- A future workflow PR demonstrates the cross-ref catching a docstring/implementation mismatch before Stage 11.

## Reference fix

PR #1241 commit `1cda17f2` (https://github.com/djust-org/djust/pull/1241/commits/1cda17f2) — `set +o pipefail` around the audit-script call to keep the daily cron green when the audit finds flagged PRs.

PR #1712, opened 2026-06-05
ci: #1708 — dogfood djust_check against the demo project in CI

### LINKED ISSUE #1708: ci: dogfood djust_check against the demo project (catch dead-button / deprecated-attr drift)
Source: Retro v1.0.1 (finding #3), surfaced by PR #1693 (#1683).

#1683 shipped to 1.0 GA with **dead buttons** in the demos (`@click` handler bindings the client never binds — flagged by the T001 system check, which already existed). The demos are a first-impression surface; the check existed but was never run against the demos in CI. CLAUDE.md #1060 ("dogfood new CLI tools against the demo project") is canon but not CI-enforced.

**Fix:** add a CI step that runs `python manage.py djust_check` (and/or `djust_audit`) against `examples/demo_project` and fails on findings (or a curated subset incl. T001/T015). This would have caught #1683 before GA, and now catches the new T015 (#1602) legacy-attr class too. Keep `continue-on-error` until it's green once, then enforce.

johnrtipton:
Closes #1708.

## What

Adds a CI step that runs `djust_check` against `examples/demo_project` and fails on the deprecated-attribute bug class. #1683 shipped dead `@click` buttons to the 1.0 GA demo even though the `T001` system check existed — because the demo templates were never run through `djust_check` in CI. This enforces CLAUDE.md #1060 (dogfood CLI tools against the demo) mechanically.

## Where the step lives

In the existing `playwright-tests` job (`.github/workflows/test.yml`), after `migrate` and before server start. That job already sets up the demo env (`uv sync --extra dev` + `cd examples/demo_project && migrate`), so no duplicate build job is needed. The job is already `continue-on-error: true`.

The step shells out to a new helper, `scripts/ci_djust_check_demo.py`.

## Verified `djust_check` invocation + exit semantics

- **`manage.py djust_check` ALWAYS exits 0.** Its `handle()` (`python/djust/management/commands/djust_check.py:266`) only prints results (pretty / `--json`) and returns — there is no exit-code logic. A bare invocation can never fail CI.
- The helper therefore runs `manage.py djust_check --json`, parses the summary, and sets a real exit code.
- It exits 1 ONLY on **error-severity checks** and the **deprecated-attribute classes `T001`/`T014`/`T015`** (the exact #1683 / #1697 bug class — `T001` is a `DjustWarning`, `checks.py:2962`). It deliberately does NOT fail on the demo's intentional warnings (S005 public-view-without-auth ×48, T012 partial-fragment ×4, etc.), which would make the demo perpetually red. Per Action #1079, this fixes exactly what #1708 cites.
- The helper also strips the `[HotReload]` stdout line the demo emits under `DEBUG=True` before parsing JSON.

## Local-clean evidence

```
$ cd examples/demo_project && python scripts/ci_djust_check_demo.py
djust_check (demo): 104 total, 0 error(s), 55 warning(s), 49 info
OK: no errors and no deprecated-attribute findings (T001/T014/T015).
# exit 0
```
Deprecated-attribute classes T001/T014/T015: **0** findings (clean post-#1683/#1697).

## Empirical canary (#252) — proves it would have caught #1683

Re-introduced a single `@click="increment"` into a scratch demo template:
```
$ python scripts/ci_djust_check_demo.py
djust_check (demo): 106 total, 0 error(s), 56 warning(s), 50 info

FAIL: 1 blocking finding(s) (errors + deprecated-attribute classes T001/T014/T015):
  - djust.T001 [warning] templates/__canary_1708.html:2 -- deprecated '@click' syntax.

This is the #1683 dead-@click bug class. Fix the templates above.
# exit 1
```
Scratch template was reverted (never committed).

## continue-on-error note (rc4 retro finding #3)

The step inherits the `playwright-tests` job's `continue-on-error: true`, so it is **non-blocking on its first runner iterations** — a new CI check exercising an env the dev machine can't fully mirror gets ≥1 runner-only iteration budgeted. The inline comment + step name say "non-blocking until green"; promote to a blocking gate (dedicated `demo-checks` job or wire into `test-summary`) once it has shipped green on the runner. **This PR is that first runner iteration.**


johnrtipton (commented):
Stage 11 — verified by running, not inspecting (built the Rust ext via `maturin develop`).

**Empirical canary (#252) — PASS.** Injected one `@click` into a scratch copy of `templates/base_with_navbar.html` → wrapper reported `djust.T001 [warning] ... deprecated @click syntax`, exit 1; restored → exit 0. Note T001 is *warning* severity, so the ID-based gate (`DEPRECATED_ATTR_IDS` at `scripts/ci_djust_check_demo.py:60`) is genuinely required — a severity-only gate would miss exactly the #1683 class.

**Clean demo**: `104 total, 0 errors, 55 warnings, 49 info`, exit 0 — matches the PR claim exactly. S005×48 + T012×4 confirmed present and NOT gated.

**Fail-safe parsing**: `_extract_json("")`/no-JSON → ValueError; malformed → JSONDecodeError; both caught in `main()` → `return 1`. Empty/garbage never silently passes. The `[HotReload]` strip is load-bearing — raw stdout line 1 is literally `[HotReload] Hot reload enabled...` and bare `json.load` chokes.

**continue-on-error**: inherited from `playwright-tests` job (`test.yml:234`), step name + CHANGELOG accurately say "non-blocking until green" (#1244 cross-ref OK). YAML valid; branch 0 behind main; CI green.

Concerns:
1. The `severity=="error"` arm is never exercised end-to-end (demo has 0 errors) — only the T001/T014/T015 ID arm gates today. Trivial logic, low risk, but untested on the runner.
2. It is now green on the runner — recommend a fast-follow promoting to a dedicated blocking `demo-checks` job (PR already documents this plan).

APPROVE

johnrtipton:
## Retrospective — PR #1712 (#1708, v1.0.2 drain 4/6)

**Task:** Dogfood `djust_check` against the demo project in CI, so a dead-button / deprecated-attribute regression (the #1683 class) red-bars a PR instead of reaching GA.

### Quality: 5/5

### What went well
- **Verifying the literal contract (#1243) found the load-bearing surprise: `djust_check` always exits 0** (print-only `handle()`). A naive `run djust_check` CI step would have been a no-op that looks like coverage — the worst kind of false guard. The implementer wrote a wrapper (`scripts/ci_djust_check_demo.py`) that parses `--json` and exits non-zero on errors + the deprecated-attr ID set.
- **The gate is ID-based, not severity-based — and that's necessary, not incidental.** T001 (the #1683 class) is *warning* severity, so a severity-only gate would have missed exactly the bug this issue exists to catch. Both the implementer and reviewer confirmed this by empirical canary (re-inject `@click` → T001 → exit 1).
- **Doesn't make the demo perpetually red:** the wrapper deliberately ignores the demo's intentional S005×48 (public-view-no-auth) and T012×4 (fragment) warnings; only errors + T001/T014/T015 gate.
- **rc4 finding #3 applied correctly:** shipped `continue-on-error: true` for the first runner iteration; it went green on that very iteration, with runner output byte-identical to local.

- It's a guard that's currently non-blocking (`continue-on-error`), so it can't fail a PR yet — and the `severity=="error"` arm is untested e2e (demo has 0 errors). **Filed a follow-up** to (a) promote it to a dedicated blocking `demo-checks` job now that it's green, and (b) add a unit test feeding the wrapper a synthetic error payload so both gate arms are covered.

### Relation to prior work
Closes the loop the v1.0.1 retro opened (finding #3 → #1708): the check that *existed* (T001) but wasn't *run* on the demos. Pairs with #1602's T015 — that new legacy-attr check is in the gate's ID set, so #1602 + #1708 compound.

### Verified
- Local + runner clean (104/0/55/49, exit 0); empirical canary (re-inject `@click` → exit 1) reproduced by both agents; fail-safe JSON parsing confirmed; YAML valid. CI green.

RETRO_COMPLETE

PR #1718, opened 2026-06-05
chore(ci): #1717 — lint cleanup + de-noise security audit + gate eslint on errors

johnrtipton:
Closes #1717.

CI lint hygiene in three coordinated parts (no framework behavior change).

## Part 1 — lint cleanup (all pre-existing)

**clippy (5 style warnings, behavior-neutral):**
| File | Before | After |
|------|--------|-------|
| `crates/djust_vdom/src/patch.rs:111` (removes, DESC) | `sort_by(\|a,b\| b.1.cmp(&a.1))` | `sort_by_key(\|b\| std::cmp::Reverse(b.1))` |
| `crates/djust_vdom/src/patch.rs:122` (inserts, ASC) | `sort_by(\|a,b\| a.1.cmp(&b.1))` | `sort_by_key(\|a\| a.1)` |
| `crates/djust_vdom/src/lib.rs:882,958` (offsets, DESC) | `sort_by(\|a,b\| b.text_byte_offset.cmp(&a.text_byte_offset))` | `sort_by_key(\|b\| std::cmp::Reverse(b.text_byte_offset))` |
| `crates/djust_templates/src/parser.rs:1553` | nested `if` inside `match` arm | folded into the arm guard |

`Reverse` preserves the exact descending order; the offset/index fields are `usize` (`Copy`), so by-value keys are valid. Sort direction is unchanged in every case.

**eslint (2 errors):**
- `js/pwa.js:9` — removed the redundant `'use strict';` (`strict`: unnecessary inside a module).
- `security.js:127` — the XSS-sanitizer denylist `val.startsWith('javascript:')` is a **false positive** for `no-script-url` (it's a denylist MATCH, not a script-URL USE). Added `// eslint-disable-next-line no-script-url` + an explanatory comment. **The denylist check is unchanged** — the disable only silences the linter, not the logic (still strips `javascript:`/`data:`/`vbscript:`).

## Part 2 — de-noise `pre-release-security-audit.yml`

- Workflow-level `env: { CARGO_TERM_COLOR: never, NO_COLOR: "1" }` → no raw ANSI escapes from cargo/clippy/cargo-audit/eslint/most tools.
- Each verbose scan step's console output wrapped in `::group::`/`::endgroup::` (collapsed by default in the Actions UI). **Full detail still flows into the uploaded `*-report.md` artifacts** (the `tee -a` is preserved). Artifact uploads unchanged.

## Part 3 — gate eslint on errors (warnings stay non-fatal)

[...]

Issue #1717, opened 2026-06-05
chore(ci): lint cleanup + de-noise the Pre-Release Security Audit (and gate eslint on errors)


johnrtipton:
Surfaced reviewing the v1.0.1rc1 release run (Pre-Release Security Audit #27029430386). Three coherent parts, one PR.

## 1. Lint cleanup (all pre-existing, none from the 1.0.1 PRs)
- **clippy (5 style warnings)** — `cargo clippy --fix` auto-applies most:
  - `crates/djust_vdom/src/patch.rs:111,122` — `sort_by(|a,b| …cmp…)` → `sort_by_key`
  - `crates/djust_vdom/src/lib.rs:882,958` — same `sort_by` → `sort_by_key(… Reverse …)`
  - `crates/djust_templates/src/parser.rs:1553` — collapsible `if` into outer `match`
- **eslint error 1 — `js/pwa.js:9`** `'use strict' is unnecessary inside of modules` → delete the redundant directive.
- **eslint error 2 — `security.js:127`** `no-script-url` is a **FALSE POSITIVE**: the line is the XSS sanitizer's denylist (`val.startsWith('javascript:')`) detecting+stripping dangerous schemes, not creating one. Add `// eslint-disable-next-line no-script-url` with an explanatory comment (do NOT remove the check).

## 2. De-noise the CI output
The audit run log is full of raw ANSI color escapes (`\x1b[1m\x1b[92m…`) and full verbose dumps from every scan.
- Set `CARGO_TERM_COLOR=never` (and `--no-color`/`NO_COLOR` for eslint/cargo-audit) so raw logs are readable.
- Keep the full report in the uploaded `*-report.md` artifact, but print only a concise summary (counts / pass-fail) to the console — e.g. wrap verbose output in `::group::`/`::endgroup::` so it's collapsed by default, or redirect detail to the artifact and echo a one-line summary + `$GITHUB_STEP_SUMMARY` table.

## 3. Gate eslint on real errors (the substantive one)
The JS scan runs `npm run lint … || true` — report-only. A genuine future `javascript:`-URL/XSS regression (severity-2 eslint error) would be swallowed and the "Pre-Release Security Audit" would still go green. Make the eslint step **fail on errors (severity 2)** while tolerating warnings (eslint already exits non-zero only on errors unless `--max-warnings` is set; preserve the report artifact via `pipefail` rather than `|| true`). After parts 1+2, eslint has 0 errors, so this can be enforcing immediately.

## Acceptance
[...]

Issue #1236, opened 2026-04-30
tech-debt: watch-list for release-workflow-touching dep bumps (manual risk-review at refile time)


johnrtipton:
Action item from v0.9.1 release retro / PR #1233 retro.

## Why

PR #1233 bumped `softprops/action-gh-release` v2 → v3 right before the v0.9.1 release-cut. The cut succeeded on first attempt, but the bump landed in the same window as the release that USES the action. If v3 had a regression, it would have surfaced at v0.9.1 cut time.

Dependabot will continue refiling future v3+ bumps. We should NOT auto-merge any dependabot PR that modifies `.github/workflows/release.yml` (or any release-critical workflow); they need explicit risk-review at refile time.

## Proposed mechanism

1. **Add a CODEOWNERS entry** for `.github/workflows/release.yml` (and similar release-critical files) that requires explicit human review before merge. Auto-merge bots/policies should respect CODEOWNERS.

2. **Tag dependabot bumps that touch release-critical files** with a `release-workflow-touching` label automatically (via a separate workflow). The label gates auto-merge.

3. **Documentation**: add a paragraph to `docs/PULL_REQUEST_CHECKLIST.md` Release Considerations section explaining the watch-list and the "risk-review at refile time" discipline.

## Estimated effort

~1 hour: write the CODEOWNERS entry + the auto-labeler workflow + the PR-checklist paragraph.

## Source

v0.9.1 release retro / Action Tracker #202. PR #1233 (action-gh-release v2→v3).

🤖 Auto-filed during v0.9.1 release retro.

PR #1720, opened 2026-06-05
fix(ci): security-audit create-issue — honor create_issue=false + cap body length

johnrtipton:
## What
Two pre-existing bugs in the **Pre-Release Security Audit** → *Audit Summary* → **"Create tracking issue"** step, surfaced when I re-ran the audit on `main` via `workflow_dispatch` (run [27031966224](https://github.com/djust-org/djust/actions/runs/27031966224)). The scans all passed; the run only red-barred on issue creation.

## Bugs
1. **Ignored `create_issue=false`.** Guard was `(inputs.create_issue == true || inputs.create_issue == '') && github.event_name != 'push'`. GHA coerces boolean `false` and `''` both to `0`, so `false == ''` is **true** → the step ran even when `create_issue=false`. Fix: drop the `== ''` clause; `inputs.create_issue == true` is correct (push already excluded by the event guard). *(This also explains why the rc1 **push** run was fine — `event_name != 'push'` skipped it there.)*
2. **No body-length cap → HTTP 422.** `body = template + full audit-summary` with no truncation; the aggregated scan output exceeded GitHub's 65536-char issue-body limit (`body is too long`). Fix: truncate to 65000 with a pointer to the `security-audit-report` artifact (which carries the full, untruncated detail — preserved by the #1717 de-noise work).

## Verification
- `.venv/bin/python -c "import yaml; yaml.safe_load(...)"` → YAML valid; guard now `inputs.create_issue == true && github.event_name != 'push'`.
- `node --check` on the extracted github-script block → JS syntax OK (truncation logic).
- Will confirm on a fresh `workflow_dispatch` after merge: with `create_issue=false` the step **skips**; with `create_issue=true` it truncates instead of 422-ing.

Neither bug was introduced by #1718 (confirmed — that PR didn't touch the summary/create-issue steps); both are pre-existing and were only reachable via manual dispatch.


johnrtipton (commented):
Stage 11 review (#1236 reviewer-label gate). Read-only; verified `origin/main...origin/fix-audit-create-issue` — diff touches ONLY the `if:` guard, the truncation block, and a CHANGELOG entry. Scans (bandit/safety/pip-audit/cargo audit/eslint) and every `upload-artifact` step are byte-identical; the `security-audit-report` upload (`pre-release-security-audit.yml:362-366`) is unchanged. No security scan weakened — this is post-scan bookkeeping only.

Guard (`:313-315`): `inputs.create_issue == true && github.event_name != "push"`. Reasoned all cases: push → skip (event guard); dispatch+`false` → `false == true` → skip (the fix; old `== ""` matched via 0==0 coercion, matching the empirical run 27031966224 RAN); dispatch+true/default → run. Correct.

Truncation (`:346-350`): `MAX_BODY=65000`; `body.slice(0, MAX_BODY - note.length) + note`. Simulated: note=151 chars, final length exactly 65000 < 65536. No off-by-one; note correctly references the `security-audit-report` artifact. `node --check` on the extracted github-script: OK. YAML: valid (venv pyyaml).

Concern (non-blocking): `== true` is robust for a `type: boolean` input, but if a future caller invokes via `workflow_call` or a string-typed path passing `"false"`, `"false" == true` → false (safe skip) but `"true" == true` → also false (would silently NOT create). For a dispatch-only boolean input this is fine today; if the input ever becomes string-sourced, switch to `inputs.create_issue == true || inputs.create_issue == "true"`. Worth a one-line comment.

CI: only failing check is `Verify reviewer label` — the gate this review satisfies. All other checks pass/pending, none failed.

COMMENT

johnrtipton:
## Retrospective — PR #1720

**Task:** Fix the Pre-Release Security Audit's create-issue step, surfaced when re-running the audit on main (run 27031966224).

**Quality: 5/5**

**What went well**
- Re-running the job on `main` (not re-running the stale rc1 tag) is what exposed both bugs — the rc1 *push* run skipped create-issue entirely (`event_name != 'push'`), so the failure was only reachable via `workflow_dispatch`. Verified the rc1 step was `skipped`, confirming the diagnosis rather than guessing.
- Root-caused the GHA coercion footgun precisely (`false == ''` → `0 == 0` → true) rather than papering over it; the fix (`== true`, drop `== ''`) is reasoned through every case (push / dispatch-false / dispatch-true).
- Truncation verified by simulation (final body length exactly 65000 < 65536); full detail preserved in the artifact (the #1717 de-noise kept artifacts intact, so the truncated issue can safely point at them).

**What didn't / could improve**
- This was latent for as long as the create-issue feature existed — any real `workflow_dispatch` audit with `create_issue=true` and enough scan output would 422. It only went unnoticed because releases trigger via *tag push* (where the step is skipped). Lesson: a workflow step that never runs on the common trigger is effectively untested — the manual-dispatch path needs at least one real exercise.
- Reviewer noted `== true` is brittle for a hypothetical future string/`workflow_call` caller; left as YAGNI (the only trigger is the boolean dispatch input), documented on the PR.

**Verified**
- Guard + truncation reviewed independently (release-workflow-reviewed gate); YAML valid; github-script `node --check` OK; scans untouched; CI green.

RETRO_COMPLETE

Issue #1678, opened 2026-05-31
{% kanban_board %} emits no dj-id anchors → every card move fails VDOM patching and falls back to html_recovery


johnrtipton:
## Summary

The `{% kanban_board %}` template tag emits its columns and cards as positional HTML with **no `dj-id` (or `dj-key`) anchors**. The tag accepts a `move_event` and wires drag-and-drop to dispatch it, so the natural usage is a stateful LiveView that mutates a card's column and re-renders the board. But the moment a card actually changes columns, the per-column child counts shift and the VDOM differ patches against **stale positional node paths** — producing a storm of failed `SetText` patches and forcing a full `html_recovery` morph on **every drag**.

The end state is correct (the move persists, the morph repaints it), but each drag pays: ~half the patches fail + an extra round-trip for recovery HTML + a full subtree morph. For the one interaction the component exists for, that's the slow/degraded path on every use.

This is **not** an `{% if %}`-block problem (the client's hint message points there); the board has no conditionals. It's purely the unanchored dynamic-child-count reorder.

## Smoking gun

`python/djust/components/templatetags/djust_components.py` — `kanban_board` (def at ~3132):

Cards (~3184–3185) carry only `data-*`, no `dj-id`:
```python
cards_html += (
    f'<div class="kanban-card" draggable="true" '
    f'data-card-id="{card_id}" data-col-id="{col_id}" '
    ...
)
```

Columns (~3201–3202) likewise:
```python
cols_html += (
    f'<div class="kanban-col" '
    f'data-col-id="{col_id}" '
    ...
)
```

The whole thing returns via `mark_safe(...)`. `data-card-id`/`data-col-id` are for the drag JS, not for the differ. With no `dj-id`/`dj-key`, the client resolves patches positionally; when column A loses a card and column B gains one, the path indices no longer line up.

## Reproduction

1. A LiveView whose template renders `{% kanban_board columns=board move_event="move_card" %}`, where `board` is `[{"id","title","cards":[{"id","title"},...]}, ...]`.
2. A `move_card(self, card_id="", from_col="", to_col="", **kwargs)` handler that moves the card between columns in the model and rebuilds `self.board`.
3. Load over WebSocket, drag a card from one column to another.

Observed console (djust 1.0.0rc16):
```
[LiveView] Applying 20 patches
[LiveView] Path traversal failed at index 8, only 3 children (parent: DIV#5l). ... a {% if %} block changed the node count.
[...]

PR #1725, opened 2026-06-05
fix(vdom): morph whitespace-separated SSR children in place on first hydration (#1724)

### LINKED ISSUE #1724: SSR→hydration replaces dj-view top-level children wholesale (full-page re-render + destroys Chart.js canvas)
## Summary

On a server-rendered LiveView page, the first WebSocket hydration **replaces the `dj-view` root's top-level children wholesale** (remove + re-add) instead of morphing them in place. This causes a full visible re-render of the entire page on **every navigation** (a perceptible "double load" ~150–250ms after first paint), and it **destroys client-side widget state** mounted on those nodes — most visibly a Chart.js `<canvas>`, which renders on first paint and then goes blank when the canvas is replaced.

`djust == 1.0.1rc1`, ASGI (Daphne + Channels), `DEBUG=False` (minified `client.min.js`), Chrome.

## Root cause (observed)

The SSR HTML contains **zero `dj-id` attributes**:

```
$ curl -s <page> | grep -c 'dj-id'
0
```

So the root and its children arrive id-less:

```html
<div dj-view="app.views.UnitsView">
  <div class="mobile-header"> … </div>
  <div class="main-wrapper"> … </div>   <!-- sidebar + entire main content -->
</div>
```

On connect the server logs:

```
INFO  Client has pre-rendered content - sending hydrated HTML for ID-based patching
INFO  Successfully mounted view: app.views.UnitsView
```

…and the client then **replaces** the root's direct children rather than morphing them. A `MutationObserver` on `document.body` during the first ~1.5s after load records:

```
REMOVED DIV mobile-header
REMOVED DIV main-wrapper
ADDED   DIV mobile-header
ADDED   DIV main-wrapper
```

i.e. the whole visible page subtree is torn down and rebuilt to attach `dj-id`s, instead of attributes being patched onto the existing nodes.

## Impact on client-side widgets (Chart.js)

Because the subtree (including a `<canvas>`) is replaced:

```
t=0    chartInstance: true,  canvas dj-id: (none)   ← SSR + inline init script ran
t=150  chartInstance: false, canvas dj-id: "1x"     ← hydration replaced the canvas
t=…    chartInstance: false                          ← chart never comes back
```

The Chart.js instance is lost and the inline init script doesn't re-run, so the chart silently disappears. This is despite `client.js`'s own comment that `morphChildren` is meant to *"preserve JS state (canvas contexts, …)"* — the preserve path isn't reached because the node is replaced, not morphed.

`dj-update="ignore"` on the `<canvas>` does **not** help: the first hydration replaces the element before the `dj-update === 'ignore'` guard in the morph path applies (the guard only protects already-matched elements on subsequent patches).

## Repro

1. A LiveView whose template root is `<div dj-view="app.views.X">` wrapping a couple of layout `<div>`s (header + main content), SSR'd with `DEBUG=False`.
2. Load the page; observe `document.body` mutations for ~1.5s after load.
3. Observe the top-level layout `<div>`s are removed + re-added (not morphed). Any JS-initialized widget inside (e.g. a Chart.js canvas) is destroyed.

## Expected

The initial hydration should **morph the SSR DOM in place** (attaching `dj-id`s as attribute patches on existing nodes) so there's no visible re-render and client-mounted JS state survives — matching the intent documented in `morphChildren`.

## Workaround in use

Re-initialize the widget after hydration via a `MutationObserver` on a stable ancestor (`document.body`), re-creating the Chart only when `Chart.getChart(canvas)` is null. Works, but every SSR page with a client-side widget needs this, and the full-page re-render flash remains.

johnrtipton:
## Summary

Fixes #1724 — on the first WebSocket hydration of a server-rendered LiveView page, the `dj-view` root's top-level children were torn down and rebuilt (remove + re-add) instead of being morphed in place. This caused a full visible re-render on every navigation (~150-250ms "double load") **and** destroyed client-side widget state mounted on those nodes — most visibly a Chart.js `<canvas>` going blank because its containing subtree was replaced.

> **Note:** the first revision of this PR fixed the wrong thing (a Strategy-2b guard on the desired node's standard `id`). Stage-11 review correctly found that `dj-id` does NOT populate the DOM `.id` property, so that guard was dead code against real WS-mount input. This revision fixes the real root cause and removes the dead branch.

## Corrected root cause — whitespace text-node misalignment

The active path is `03-websocket.js` `skipMountHtml` branch → `morphChildren(_morphContainer, _morphTemp)` (line ~372). The wholesale replace happens inside `morphChildren` (`12-vdom-patch.js`).

Real SSR HTML parsed into the DOM via `innerHTML` carries **inter-element whitespace text nodes** (the newlines/indentation between sibling elements). The hydrated WS-mount HTML (`data.html`, Rust-rendered) carries `dj-id` on every element (`parser.rs:420`) — **not** a standard `id` — and may have different inter-element whitespace.

In `morphChildren`, when the desired node is an element but the positional existing node lands on an insignificant whitespace-only text node, every element-matching strategy is skipped (Strategy 1 / 2 all require `eNode.nodeType === ELEMENT_NODE`), so the code falls to Strategy 3 (clone+insert, line ~809) and then **removes the real existing element** in the unmatched-cleanup loop (line ~813). That is the remove+add the reporter's MutationObserver recorded.

Empirically bisected the trigger (existing/SSR side × desired/WS-mount side, all desired carrying `dj-id`):

| existing whitespace | desired whitespace | identity preserved pre-fix? |
|---|---|---|
| yes | yes (matches) | ✅ (Strategy 2) |
| no  | yes | ✅ |
| **yes** | **no** | **❌ — the bug** |
| **mismatched amounts** | | **❌ — the bug** |

## Fix

Skip insignificant whitespace-only existing text nodes when aligning a desired **element** (or **dj-if boundary comment**), so the strategies land on the next real existing element and morph it in place. Reuses the shared `isSignificantChild` / `isWhitespacePreserving` predicates (#1655/#1678) so:
- significant whitespace inside `<pre>`/`<code>`/`<textarea>` is NOT skipped;
- NBSP and non-blank text are NOT skipped;
- dj-if comment markers are NOT skipped (conditional-subtree boundaries preserved).

Also **removes the prior dead Strategy 2b** (it guarded on the desired node's standard `id`, which dj-id never populates).

## Test plan

New faithful regression test `tests/js/ssr_hydration_morph_1724.test.js` (6 cases, loads the real `morphChildren` from the built `client.js`). Per the CLAUDE.md "Reproduction fidelity" rule (#1650/#1638/#1637), the existing side is built via `innerHTML` **with** inter-element whitespace and the desired side carries **`dj-id`** (the real Rust-renderer output) — not appendChild + synthetic `id`:
1. differing-whitespace case (the bug) — element identity + Chart.js canvas + instance marker survive.
2. matching-whitespace case — identity + canvas survive.
3. no remove+re-add of element children (no MutationObserver churn).
4. `<pre>` significant-whitespace guard — code-block text not skipped.
5. dj-if boundary-comment guard (#1678) — markers + element identity preserved.
6. keyed reorder/replace guard — existing node with a different id is still replaced.

**Gate-off evidence (#254):** against the previous (Strategy-2b) bundle but **without** the whitespace fix, the 2 behavior-meaningful tests (differing-whitespace identity + no-churn) **fail** — proving (a) the faithful test exercises the real path, and (b) Strategy 2b alone did NOT fix the bug. Cases 2/4/5/6 pass with and without the fix (pre-existing behavior the fix preserves). Non-tautological.

**Full JS suite:** 1660 passed / 144 files / 0 failures. High-blast-radius whitespace/dj-if/morph suites verified explicitly (`whitespace.test.js`, `dj_if_subtree_patches`, `dj_if_comment_predicate`, `significant_child_predicate_1655`, `significant_children_comment_filter_1640`, `morph.test.js`, `vdom_recovery`, `insert_child_regular_comment_1636`, `ws-mount-prerender-divergence-1610`).

## Risk notes for reviewer

- The whitespace-skip only advances past text nodes for which `isSignificantChild(...) === false` — the same predicate the path/index resolvers use, so it cannot desync index-based patches.
- It fires only when the desired child is an element or a dj-if comment; significant text and non-dj-if comments take their existing code paths unchanged.
- `<pre>`/`<code>`/`<textarea>` content is protected via `isWhitespacePreserving(existing)` (computed once per `morphChildren` call).
- Keyed reorder/replace (existing nodes carry ids → Strategy 1) is unaffected and explicitly tested.

Closes #1724


johnrtipton:
## Stage-11 Adversarial Review — PR #1725 (fix #1724)

### Verdict: **REQUEST-CHANGES**

The fix is clean, well-guarded, and bundle-faithful — but a faithful reproduction shows **it does not fix the reported bug**. Strategy 2b is shaped against a synthetic DOM that does not match the real SSR→hydration path. The user's Chart.js canvas will still go blank after merge.

Pre-review gates all pass: BEHIND main = 0; exactly 2 commits (impl+tests with no CHANGELOG, then CHANGELOG-only); bundle artifacts match a fresh `make build-js` byte-for-byte; full JS suite green (1658/1658); CHANGELOG entry in `[Unreleased] → ### Fixed`, cites #1724, Keep-a-Changelog compliant.

---

### 🔴 MUST-FIX

**1. The fix does not address the real bug — reproduction-fidelity gap (`tests/js/ssr_hydration_morph_1724.test.js`, `src/12-vdom-patch.js:821`).**

The issue body and the test docstring both state the hydrated WS-mount HTML carries **`dj-id`, not standard `id`** (parser.rs:420; issue: "canvas dj-id: `1x`"). I verified `dj-id` does **not** populate the `.id` DOM property, so `dId = dNode.id || null` is `null` for the real desired nodes. **Strategy 2b's `if (dId && …)` guard therefore never fires in the reported scenario** — the dj-id case is owned by the pre-existing Strategy 2 (`!dId && !eNode.id`), which has shipped since v0.3.0 (commit 81e07594), i.e. it was already in the `1.0.1rc1` bundle the reporter ran. Yet they still saw the wholesale replace.

Tracing why, the actual trigger is **whitespace text-node misalignment**, not the id/dj-id distinction. Real SSR HTML is rendered via `innerHTML` and carries inter-element whitespace text nodes (`[text, div, text, div, text]`). The Rust-rendered desired HTML is effectively whitespace-free (`[div, div]`). When `morphChildren` processes the first desired `<div>`, the positional `eNode` is a **text node**, so `eNode.nodeType === Node.ELEMENT_NODE` is false → Strategy 2 is skipped → **and Strategy 2b is skipped for the exact same reason** → Strategy 3 clone+insert + remove tears down the subtree.

Decisive standalone repro (existing rendered via `innerHTML` with whitespace; desired carries `dj-id`, matching parser.rs:420 and the issue body):

```
[PRE-FIX ] canvas preserved: false | chart survives: false
[POST-FIX] canvas preserved: false | chart survives: false
```

The fix changes nothing for the real DOM shape. The regression test passes only because it builds the existing children with `appendChild` (no whitespace text nodes) and — in case2/case3 — uses standard `id` on the desired nodes (which the Rust renderer does not emit). The one faithful case in the suite (case1, `dj-id` variant) passes **pre-fix** (Strategy 2 handles it), so it does not exercise Strategy 2b at all.

Gate-off confirmation against the pre-fix bundle (origin/main client.js):
```
case1 dj-id identity preserved   → PASS pre-fix   (Strategy 2, not 2b — not the fix)
case2 id identity preserved      → FAIL pre-fix   (exercises 2b, but `id` variant is not the real shape)
case3 no remove+re-add (id)      → FAIL pre-fix   (exercises 2b, same caveat)
case4 keyed replace still works  → PASS pre-fix   (guard test, expected pass both sides)
```
So the implementer's "2 of 4 fail without the fix" is literally true, but the 2 that fail are the **synthetic `id` variant**; the faithful `dj-id` case (case1) passes pre-fix. The test reproduces a different bug than the one filed.

This is the CLAUDE.md "Reproduction fidelity" rule (#1650/#1638/#1637): the harness must exercise the REAL path. Required:
- Build the existing SSR root via `innerHTML` (with the whitespace text nodes real SSR has), not `appendChild`.
- Use **`dj-id`** on the desired nodes (the real Rust render shape), not standard `id`.
- Assert canvas/widget identity survives. With that harness, the current Strategy 2b does **not** make the test pass — confirming the fix is incomplete.

The likely correct fix is to skip/ignore non-significant whitespace text nodes during positional alignment in `morphChildren` (so element dNodes align to element eNodes), or to advance `eIdx` past whitespace-only text nodes before the element strategies. Strategy 2b as written is, at best, dead weight for the reported case.

---

### 🟡 SHOULD-FIX

**2. Strategy 2b is redundant with Strategy 2 for the only real shape (`src/12-vdom-patch.js:821`).** Because the renderer emits `dj-id` (so `dId` is null), every real hydration child is already handled by Strategy 2. Strategy 2b only ever fires when the desired node has a literal `id` attribute — a path the framework's own renderer does not produce. If a faithful fix lands (whitespace handling), re-evaluate whether Strategy 2b carries its weight or should be dropped to keep the morph algorithm minimal (manifesto #1).

**3. CHANGELOG over-claims (`CHANGELOG.md:18`).** The entry states the change "preserves client-side widget state … a Chart.js `<canvas>` no longer goes blank." Per the decisive repro that outcome is not achieved for the real bug. Whatever the final fix, the CHANGELOG claim must be backed by a faithful test (doc-claim-verbatim TDD, #1046).

---

### 🟢 NITS

**4.** Strategy 2b correctly adds the reused node to `matched` (line 824) before the remove-unmatched loop, and `morphElement` (lines 918–923) copies all desired attributes — including `id`/`dj-id` — onto the reused node, so later `querySelector('[dj-id=…]')` patches resolve. The mechanics of the branch are sound; the problem is purely that the branch is never reached for the real input.

**5.** Over-reach hunt (the requested priority): I could not find a correctness over-reach. Because Strategy 2b only fires on **same-tag** elements and `morphElement` on same-tag elements converges to the identical final DOM as clone+replace, morph-in-place vs replace differ only in node identity (the intended behavior). Probes for content-removal, reorder, and same-tag content-swap all produced correct output. The keyed-replace guard (`!eNode.id`) correctly routes different-id existing nodes to Strategy 3. No over-reach 🔴/🟡.

---

**Tests/CI:** green (1658/1658 JS tests; bundle byte-identical to `make build-js`). The green suite is the problem here — it certifies a synthetic scenario, not the filed one.

johnrtipton:
## Stage-11 Re-Review — PR #1725 (fix #1724), fix-pass HEAD `2df890be`

### Verdict: **APPROVE**

The prior 🔴 is resolved. The dead Strategy 2b is gone, and the real root cause — **whitespace text-node misalignment** in `morphChildren` — is fixed correctly. My faithful Chart.js-canvas reproduction (the ground truth from the first review, which printed `canvas preserved: false` for both pre-fix and Strategy-2b) now passes.

---

### The load-bearing check: my faithful canvas repro now passes ✅

Same standalone harness as the first review — existing SSR root built via `innerHTML` (real inter-element whitespace text nodes), desired carries **`dj-id`** (parser.rs:420, what the Rust renderer actually emits), with a `<canvas>` carrying a `_chartInstance` marker:

```
[PRE-FIX  (origin/main)] canvas preserved: false | chart survives: false
[POST-FIX (this HEAD)  ] canvas preserved: true  | chart survives: true
```

The bug the reporter filed is genuinely fixed, not a proxy.

---

### Over-skip safety — significant / dj-if / whitespace-preserving content ✅

The fix only advances `eIdx` past existing text nodes for which `!isSignificantChild(eNode, preserveWhitespace)` holds, using the shared single-source-of-truth predicate (#1655) and `isWhitespacePreserving` (walks ancestors). I ran adversarial cases against this HEAD's bundle:

- **Significant text between elements** (`<span>A</span>Hello<span>B</span>`): `Hello` preserved, not skipped.
- **Significant text morphed** (`OldText`→`NewText`): correctly morphed in place, not dropped.
- **`<pre>` whitespace** (`  leading\n  indented`): all whitespace preserved.
- **`<pre>` as a child with a `_marker`**: identity + marker survive; the whitespace-preserving element is morphed, not torn down.
- **dj-if boundary markers** (`<!--dj-if id="if-0"-->…<!--/dj-if-->`): both open/close markers preserved, inner element morphed.
- **NBSP** (codepoint 160): preserved (significant).
- **Leading significant text + orphan whitespace cleanup + multi-element-with-whitespace**: all identities preserved, orphan whitespace pruned, no dangling nodes.

No over-skip risk for significant, dj-if, or `<pre>`/`<code>`/`<textarea>` content. The `matched.has(eNode)` and `isSignificantChild` guards in the skip loop are correct.

---

### Test quality — gate-off confirms non-tautological ✅

Ran the rewritten `tests/js/ssr_hydration_morph_1724.test.js` (6 cases) against the **pre-fix** bundle:

```
× preserves element identity + Chart.js canvas when desired whitespace differs (the bug)   FAIL pre-fix  ← behavior-meaningful
✓ preserves element identity + canvas when whitespace matches both sides                    (Strategy 2 already handled — guard)
× does NOT remove+re-add the element children (no MutationObserver churn)                    FAIL pre-fix  ← behavior-meaningful
✓ does NOT skip significant whitespace inside a <pre>                                        (guard, expected both sides)
✓ does NOT skip a dj-if boundary comment marker (#1678)                                      (guard, expected both sides)
✓ still REPLACES when existing children carry a different id (keyed reorder)                 (guard, expected both sides)
```

Exactly 2 of 6 fail pre-fix — the implementer's claim is accurate. The test is now faithful (existing via `innerHTML` with whitespace; desired carries `dj-id`, not synthetic `id`), per the CLAUDE.md "Reproduction fidelity" rule cited in the test docstring. The 4 passing-both-sides cases are legitimate don't-break-this guards, not tautologies.

---

### Mechanical gates ✅

- **BEHIND origin/main = 0.**
- **Two-commit shape intact**: commit 1 (`00eef5cc`) — src + tests + bundle, **no CHANGELOG**; commit 2 (`2df890be`) — **CHANGELOG only**.
- **Bundle integrity**: committed `client.js` and `client.min.js` are byte-identical to a fresh `make build-js`.
- **Full JS suite green**: 1660 passed / 0 failed (144 files).
- **CHANGELOG** (`[Unreleased] → ### Fixed`, Keep-a-Changelog): now accurately attributes the root cause to whitespace text-node misalignment, describes the real fix (skip insignificant whitespace when aligning an element/dj-if comment), and notes pre/code/textarea, dj-if markers, and keyed reorder are unaffected. No canvas over-claim — the canvas claim is backed by the faithful test + my repro.

---

### Remaining findings: none (no 🔴, no 🟡)

🟢 Nit (non-blocking): the whitespace-skip is one-directional (skips insignificant existing whitespace to align elements); the symmetric case — desired-side whitespace that has no existing counterpart — is handled by the existing text-node branch (Strategy 3 insert) and the cleanup loop, which I verified produces correct DOM. No action needed.

Clean fix. Ship it.

johnrtipton:
## Retrospective — PR #1725 (pipeline-drain v1.0.2)

**Task**: v1.0.2 — drained #1724 (SSR→hydration replaces dj-view top-level children wholesale, destroying Chart.js canvas state).

### Quality: 4/5
Shipped correct after one fix-pass. The first revision was a textbook reproduction-fidelity miss; the Stage-11 reviewer's faithful repro caught it before merge.

### What went well
- Reproducer-first discipline at Stage 11 caught a 🔴 the implementer's first repro missed. The reviewer built a faithful Chart.js-canvas harness (existing via `innerHTML` WITH whitespace, desired carrying `dj-id`) and proved both pre-fix AND the first fix (Strategy 2b) tore down the canvas (`canvas preserved: false`).
- The real fix reused existing shared predicates (`isSignificantChild`/`isWhitespacePreserving`, #1655/#1678) rather than inventing new whitespace logic — so `<pre>`/`<code>`/`<textarea>`/NBSP/dj-if markers are provably not over-skipped.
- Gate-off (#254) verified non-tautological on the final test (2 of 6 cases fail pre-fix).

### What didn't
- First revision (Strategy 2b, keyed on the standard `.id`) was **dead code against real input**: the WS-mount HTML carries `dj-id`, which never populates the DOM `.id` property, so the new strategy's guard never fired. Classic CLAUDE.md "Reproduction fidelity" failure (#1650/#1638/#1637) — the original test used `appendChild` (no whitespace text nodes) + standard `id`, exercising a path the real bug doesn't take.

### Process finding
The real root cause — **whitespace text-node misalignment** in `morphChildren`'s positional matching — was only found because the reviewer rebuilt the repro from `innerHTML` (faithful to how SSR content actually enters the DOM) rather than `appendChild`. Reinforces: VDOM-patch reproducers MUST construct the existing DOM the way the browser does (parse from an HTML string), not via DOM-builder APIs that omit insignificant whitespace.

### Verified
- Reviewer's faithful canvas repro: pre-fix `false` → post-fix `true` (canvas + `_chartInstance` survive).
- Full JS suite: 1660 passed / 144 files / 0 failures; high-blast-radius suites (whitespace, dj_if_subtree_patches, significant_child_predicate, morph, vdom_recovery, ws-mount-prerender-divergence-1610) explicitly green.
- All CI jobs green (JS/Python/Rust/playwright/security/CodeQL). Two-commit shape intact.

`RETRO_COMPLETE`

Issue #233, opened 2026-02-03
LiveView templates don't receive Django context processor variables


johnrtipton:
## Problem
Django context processors (added via `TEMPLATES[].OPTIONS.context_processors`) inject variables into template contexts. These work on regular Django views but are **not available** inside LiveView templates.

### Example
```python
# context_processors.py
def djust_theming(request):
    return {'theme_head': '<link ...>', 'theme_switcher': '<button ...>'}
```
```html
<!-- Works on login page (regular Django view) -->
{{ theme_head }}

<!-- Empty in LiveView template -->
{{ theme_head }}
```

### Root Cause
LiveView rendering uses the Rust engine which only sees `self.*` instance attributes, not the Django template context that would normally include context processor output.

### Workaround
Use static CSS files and inline JS instead of context processor variables in LiveView templates.

### Suggestion
Merge Django context processor output into the LiveView template context during rendering, or document this as a known limitation.

johnrtipton:
Fixed — LiveView now applies Django context processors via `_apply_context_processors()` in `ContextMixin`. All processors from `TEMPLATES[].OPTIONS.context_processors` are loaded, called, and merged into the template context. View context takes precedence over processor context.

PR #1726, opened 2026-06-05
fix(templates): propagate context-processor vars into {% include %} (#1722)

### LINKED ISSUE #1722: Context-processor theme vars ({{ theme_panel }} / {{ theme_head }}) are empty inside nested {% include %} (incomplete #233)
## Summary
**djust 1.0.1rc1.** Context-processor vars from `djust.theming.context_processors.theme_context` render correctly at the **top level** of a LiveView template, but render **empty when used inside a nested `{% include %}`** partial.

#233 ("LiveView templates don't receive Django context processor variables") is closed, but the fix doesn't propagate into includes.

## Observed
- `{{ theme_head }}` in `base.html` (top level) → renders the `<style>` + anti-FOUC script correctly. ✅
- `{{ theme_panel }}` in a nested include (`base.html` → `{% block body %}` → `{% include "_layout_start.html" %}` → `{% include "_sidebar.html" %}`) → renders **empty** (the surrounding markup renders; the var is blank). ❌
- Verified the var is non-empty server-side via `theme_context(request)['theme_panel']`, and that it renders when placed at the top level — so the value exists; it's just not in the include's context.

(View instance attributes, e.g. `{{ nav_items }}`, DO reach the same includes — only the context-processor vars don't.)

## Workaround
Render `{{ theme_panel }}` at the top level of `base.html` instead of in the sidebar partial.

## Expected
Context-processor variables should be available inside `{% include %}`d templates (Django's default `RequestContext` behavior), so shared partials (navs, sidebars) can use `{{ theme_panel }}` / `{{ theme_head }}`.

## Repro
djust 1.0.1rc1; put `{{ theme_panel }}` in an `{% include %}`d partial vs at the base-template top level — empty in the former, rendered in the latter.

johnrtipton:
## Summary

Context-processor variables (e.g. djust theming's `{{ theme_panel }}` / `{{ theme_head }}`) rendered correctly at the **top level** of a page but rendered **EMPTY** inside the LiveView's dj-root template and its nested `{% include %}` partials. The reporter's discriminator — a plain view-instance attribute (`{{ nav_items }}`) DID reach the same includes while context-processor vars did not — was the load-bearing clue: the bug was specific to context-processor vars, not to includes in general.

Closes #1722 (completes #233).

## Confirmed root cause

`render_full_template` (the GET page-shell render) applies `_apply_context_processors` to the outer shell only (`python/djust/mixins/template.py:853`). The dj-root template is rendered by a **different** path — `render()` / `render_with_diff()` → `_sync_state_to_rust` (`python/djust/mixins/rust_bridge.py:487`) — which builds its render context from `get_context_data()` **without** applying context processors. That path drives both the initial GET dj-root render and **every WebSocket update**, so any context-processor var used in the dj-root template (or an include reached from it) resolved to empty, while plain view attributes (set on `self`) worked.

The orchestrator's initial hypothesis (the Rust `{% include %}` `clone()` at `renderer.rs:685`) was disproven: 4 new Rust tests confirm `context.clone()` faithfully carries both the value AND its safe-key marking into includes. The divergence was upstream in Python — the vars never reached the Rust context for the dj-root render at all.

## Engine-path declaration

User templates render through the **Rust** template engine (PyO3, `djust_vdom` `RustLiveView`). The bug and the fix are both on the Python side (which context is converted to the Rust `Context`). The Rust include/clone path was verified correct and is unchanged. No separate Django-engine path is involved in the dj-root render.

## Fix

Apply `_apply_context_processors` inside `_sync_state_to_rust`, right after the DB-normalize pass. It only ADDS keys absent from the view context (view context still wins on collisions) and is a no-op when `request` is `None`, so it is safe on the WS path. SafeString marking flows through the existing `safe_keys` collection; the vars are sent once on the first sync and Rust's merging `update_state` retains them across subsequent partial syncs.

## Test plan

- **Rust** (`crates/djust_templates/src/renderer.rs`): 4 cases pinning that the `{% include %}` clone carries value + safe_keys (top-level, single include, nested include, plain-var discriminator). All pass — confirms the Rust path was not the bug.
- **Python** (`python/tests/test_context_processors_in_includes_1722.py`): reproduces the exact discriminator via the `render()` path — a context-processor SafeString var inside a dj-root include must render non-empty/unescaped, while a plain view attr still reaches the include.

Results:
- `cargo test -p djust_templates --lib`: 304 passed.
- `pytest -k "theme or include or context_processor or context"`: 1066 passed.
- Full Python suite: 7506 passed (+13 benchmark tests), 0 failures.
- Pre-push hooks (full Python + Rust release suite): passed.

## Gate-off evidence (#254)

Reverting only the fix in `rust_bridge.py` and re-running the new Python test:

```
AssertionError: <div dj-root><p>Count: 0</p><aside>[panel:][nav:home|about]</aside></div>
FAILED test_context_processor_var_reaches_include
```

`[panel:]` (empty) is produced and the behavior-meaningful test fails; the discriminator test still passes (that path was never broken). Restoring the fix makes both pass.


johnrtipton:
## Stage-11 adversarial review — PR #1726 (fix #1722)

**Verdict: APPROVE-WITH-NITS**

Findings: 🔴 0 · 🟡 1 · 🟢 6

The root-cause diagnosis is correct, the fix is minimal and well-placed, and the
two findings the brief flagged as most-likely-to-matter both resolve favorably on
the load-bearing one (WS request is non-None) and to a bounded, defensible
trade-off on the other (per-event cost).

---

### Pre-review gates (all pass)

- **STALE-BASE (#250):** `BEHIND=0`. `git log origin/main..HEAD` shows only the two
  PR commits, no accidental reverts. 🟢
- **Two-commit shape:** exactly 2 — `2b83d2ac` impl+tests (renderer.rs,
  rust_bridge.py, test file; **no CHANGELOG**), `cdebbd36` **CHANGELOG-only**. 🟢

### Empirical results

- `pytest python/tests/test_context_processors_in_includes_1722.py` → **2 passed**.
- `cargo test -p djust_templates --lib test_1722` → **4 passed**.
- **Gate-off self-test (#254/#1200):** temporarily gated the fix
  (`if False and request is not None:`) and re-ran. The bug test failed with the
  exact reported symptom — `<aside>[panel:][nav:home|about]</aside>` (empty panel,
  working nav) — and the discriminator test still passed. Restored. The test is
  **non-tautological** and reproduces the reporter's precise discriminator.

---

### 🟡 PERF-1 — `theme_context` runs unmemoized on every WebSocket event (hot path)
`python/djust/mixins/rust_bridge.py:546-548` · `python/djust/theming/context_processors.py:94-179`

`_sync_state_to_rust` runs on every WS event. Context-processor **resolution** is
cached (`_resolved_processors_cache` / `_context_processors_cache`, context.py:438,
372 — confirmed module-level), but the processor **functions** run per-sync
(context.py:402, `processor(request)`).

For projects using djust theming's `theme_context` as a context processor, the cost
is real and only partially cached:
- `_render_theme_outputs(...)` is `@lru_cache(maxsize=512)` (context_processors.py:35) — fine.
- **But** `theme_context` also calls four template-tag bodies through `_safe_render`
  (context_processors.py:164-167: `theme_head`, `theme_panel`, `theme_mode_toggle`,
  `theme_preset_selector`). These are **not** cached — e.g. `theme_panel`
  (theme_tags.py:407-466) does `get_theme_manager` → `get_state` →
  `get_available_presets` → builds design/pack/layout lists → `resolve_theme_template`
  → `tmpl.render(...)` on every call. Pre-PR this ran once per HTTP request via
  `render_full_template`; post-PR it also runs once per WS event.

This is bounded — it only affects projects with `theme_context` installed, and
the dominant CSS render is LRU-cached — so it does not block. But it is a genuine
per-event regression for theming users (roughly the 4 uncached tag renders).

**Mitigation note (not blocking):** unconditional re-application is actually
*more correct* than first-sync-only gating — if a theme var changes mid-session
(e.g. a theme-switcher event), change-detection (rust_bridge.py:663-669, immutable
value-equality on the SafeString) only forwards it to Rust because the processors
re-ran. Gating to first sync would silently break dynamic theme switching.
The right follow-up is **request-scoped memoization inside `theme_context`** (cache
the 4 `_safe_render` outputs on `request` keyed by theme state), not gating the
call site. Suggest filing a follow-up issue rather than changing this PR.

### 🟢 confirmations

1. **WS-path request is NON-None (critical, resolved).** `handle_connect` sets
   `self.view_instance.request = factory.get(...)` with session + user
   (`websocket.py:1910/1963`). It is an instance attr on the long-lived view, never
   nulled anywhere (`grep` for `request = None` → no hits). `live_redirect_mount`
   re-mounts a fresh view with a fresh request the same way. **The fix works on
   every WS update and on navigation, not just the initial GET** — the reporter's
   "every navigation" symptom is fully addressed.

2. **SafeString propagation (finding #3).** `theme_head` and `theme_panel` are both
   SafeStrings (context_processors.py:170-172; `theme_panel` text comes from the
   tag's `mark_safe(...)` at theme_tags.py:456). They flow through `safe_keys`
   (rust_bridge.py:755-761) so Rust renders them unescaped. The Python test asserts
   the **unescaped** shape (`[panel:<div class="theme-panel">PANEL</div>]`) and
   `[panel:]` absent — locking out both the empty-render bug and a would-be
   double-escape regression. Rust tests independently pin the
   clone-carries-safe-key include path.

3. **update_state merge semantics (finding #5) — verified.** Rust merges; the fix
   re-runs processors every sync so vars stay present, and change-detection
   (rust_bridge.py:644-694) compares the re-created SafeStrings by value (immutable
   path, 663-669) → identical values are *not* re-sent (Rust retains), changed
   values *are* sent. Correct in both the static and dynamic cases.

4. **View-context precedence (finding #4).** `_apply_context_processors`
   (context.py:407-409) only adds keys absent from the context — view wins on
   collisions (e.g. a view-set `messages`). Merge runs after the DB-normalize pass
   and before csrf/date injection; ordering is correct.

5. **Test quality (#254/#1200).** Non-tautological (gate-off empirically fails),
   reproduces the exact discriminator (view-attr reaches include, CP var did not),
   and asserts unescaped output. The 4 Rust cases genuinely exercise the
   include/nested-include clone-with-safe-key path and the plain-var discriminator.

6. **CHANGELOG** accurate, in `### Fixed` under `[Unreleased]`, cites #1722 +
   completes #233, documents the no-op-when-no-request and view-wins semantics.

---

### Recommendation
Approve. File a follow-up issue for request-scoped memoization of `theme_context`'s
four uncached tag renders (PERF-1) so theming projects don't pay them per WS event.
The correctness of the fix is sound; the perf item is an optimization, not a defect.

johnrtipton:
## Retrospective — PR #1726 (pipeline-drain v1.0.2)

**Task**: v1.0.2 — drained #1722 (context-processor theme vars empty inside nested {% include %}).

### Quality: 5/5
Clean first-pass fix with reproducer-first discipline. The implementer disproved the orchestrator's cited hypothesis and traced the real root cause symptom-up.

### What went well
- **Symptom-up triage worked exactly as CLAUDE.md prescribes.** The orchestrator hypothesized the bug was in the Rust include handler (`renderer.rs:685` `context.clone()`). The implementer wrote 4 Rust tests proving the clone faithfully carries values + safe-key marking into includes (single + nested) — disproving the hypothesis — and traced backward to the real divergence: `_sync_state_to_rust` (rust_bridge.py) built the Rust render context from `get_context_data()` only, never applying `_apply_context_processors`. That completes #233 (which only populated the context dict, not the dj-root render path).
- The reporter's discriminator (nav_items reaches includes, theme_panel doesn't) was the load-bearing clue: view attrs come from get_context_data() (present), context-processor vars did not reach the dj-root path.
- Gate-off (#254) verified: reverting the fix reproduces the exact reported symptom `[panel:]` (empty), discriminator stays green.

### What didn't
- Nothing blocking. One 🟡 surfaced (perf): the fix runs context processors on every WS event (not just GET). The reviewer correctly identified that gating to first-sync-only would BREAK dynamic theme switching, so the right cure is request-scoped memoization inside theme_context — filed as follow-up #1727 (scope discipline #1079).

### Process finding
The reviewer caught a subtle correctness trap in the obvious "optimization": you cannot gate context-processor application to first-sync to save the per-event cost, because change-detection only forwards CHANGED vars — the processors must re-run each sync to detect a theme switch. Memoization (not gating) is the only correct optimization. Worth remembering for any per-event work that feeds change-detection.

### Verified
- WS-path request confirmed non-None (handle_connect sets a long-lived request) → fix is effective on every update + navigation, not just initial GET. This was the critical reviewer check.
- cargo test -p djust_templates: 304 passed; Python theming/include/context subset: 1066 passed; full Python suite: 7506 passed; pre-push hooks green.
- All CI jobs green. Two-commit shape intact. Follow-up #1727 filed for the perf memoization.

`RETRO_COMPLETE`

Issue #1051, opened 2026-04-25
tech-debt: Stage 4 plan-template "Engine path" bullet should generalize beyond template-tags — any feature that touches the template rendering pipeline (filters, tags, context processors, custom blocks, post-processing hooks) must declare which en...


johnrtipton:
**Source**: Retro v0.7.1 / PR #993 generalization
**Action Tracker**: row #131

**Action**: Stage 4 plan-template "Engine path" bullet should generalize beyond template-tags — any feature that touches the template rendering pipeline (filters, tags, context processors, custom blocks, post-processing hooks) must declare which engine(s) the user templates run through

**Notes**: Generalizes Action #129. PR #993 caught the dual-engine bug ONLY because pre-push runs the full demo suite; targeted Stage 6 subsets miss it. Class of bug: any code path that participates in user template rendering can silently work in one engine and 500 in the other. Enforcement: Stage 4 plan template's "Engine path" row applies to filters, context processors, post-processing hooks, and any registry-style API — not just `register_tag_handler`.

---
_Filed by `/pipeline-retro --reconcile` on 2026-04-25 as part of Scope C: file all Action Tracker rows as GitHub issues._

johnrtipton:
Shipped in PR #1226 (v0.9.1-7 cleanup batch). Comma-list `Closes #...` form parsed only the first ref — the rest closed manually. (This bites again; #1184's parenthesized form is documented in PR-checklist, but the comma-list-on-its-own-line form has the same auto-close failure.) 🤖 v0.9.1-7 cleanup batch closure.

PR #1728, opened 2026-06-05
fix(theming): register theme tags as Rust tag handlers so {% theme_panel %} works (#1721)

### LINKED ISSUE #1721: Rust template engine rejects {% theme_panel %} tag the theming docs recommend (Unsupported template tag)
## Summary
**djust 1.0.1rc1.** The theming guide (docs.djust.org/theming) documents the theme switcher as the template tag **`{% theme_panel %}`** (with `{% load theme_tags %}`). Using it in a LiveView template (rendered by the Rust engine) raises a 500:

```
RuntimeError: Template error: Unsupported template tag '{% theme_panel %}'.
Register a handler via djust._rust.register_tag_handler(), or use Django's template engine instead.
```

So the documented tag is unusable in djust-rendered templates. The `{{ theme_panel }}` **context-string** form (added in #1435) does render — but only at the top level (see the companion issue about includes), and the docs still show the tag form.

## Repro
1. djust 1.0.1rc1, a LiveView whose template includes:
   ```django
   {% load theme_tags %}
   {% theme_panel %}
   ```
2. Load the page → 500, `Unsupported template tag '{% theme_panel %}'`.

## Expected
Either the Rust template engine supports `{% theme_panel %}` (and `{% theme_head %}`, `{% theme_mode_toggle %}`, etc.), or the theming docs are updated to use the `{{ theme_panel }}` context-string form for djust-rendered templates. Right now docs ↔ engine disagree (cf. #1452 on `{{ theme_head }}` vs `{% theme_head %}`).

Found integrating theming into a production djust 1.0.1rc1 app.

johnrtipton:
## Summary

The theming guide (docs.djust.org/theming) documents the theme switcher as the template tag `{% theme_panel %}` (with `{% load theme_tags %}`). Using it in a LiveView template — rendered by djust's **Rust** template engine — raised a 500:

```
RuntimeError: Template error: Unsupported template tag '{% theme_panel %}'.
Register a handler via djust._rust.register_tag_handler(), or use Django's template engine instead.
```

So the documented tag was unusable in djust-rendered templates. The `{{ theme_panel }}` context-string form (#1435) rendered fine, so docs and engine disagreed (cf. #1452 on `{{ theme_head }}` vs `{% theme_head %}`).

## Decision (Option A — set by orchestrator)

Register the documented theme tags as Rust tag handlers so the `{% theme_X %}` tag form works in the Rust engine. This makes docs↔engine agree **without editing the external docs.djust.org repo**, and supports the customization-with-args form (`{% theme_panel show_packs=False %}`). The error message itself points at exactly this mechanism (`register_tag_handler`).

## What changed

- **`python/djust/theming/rust_handlers.py`** (new) — a thin `ThemeTagHandler` bridge implementing the `TagHandler` contract (`render(self, args, context) -> str`). It parses the Rust engine's `args` into kwargs, builds a `{"request": context.get("request")}` context, and delegates to the **same** `theme_tags.py` `@register.simple_tag` body the `{{ }}` form uses. `register_with_rust_engine()` registers a handler for each documented user-facing tag, guarded by `has_tag_handler` (idempotent) and `ImportError` (graceful degradation sans-Rust).
- **`python/djust/theming/apps.py`** — call `register_with_rust_engine()` from `DjustThemingConfig.ready()`.
- **`python/djust/tests/test_theme_tags_rust_engine_1721.py`** (new) — regression tests.
- **CHANGELOG.md** + **docs/website/guides/components.md** — Fixed entry + a note that the tag form works in both engines, with `{{ }}` vs `{% %}` guidance.

### Tags registered

The 5 issue-cited tags — `theme_panel`, `theme_head`, `theme_switcher`, `theme_mode_toggle`, `theme_preset_selector` — plus the remaining documented user-facing theme tags (`theme_css`, `theme_css_link`, `theme_framework_overrides`, `theme_preset`, `theme_mode`, `theme_resolved_mode`).

### Handler bridge — how request/kwargs flow

- **request**: pulled from the Rust render context via `context.get("request")`. `None` is fine — `get_theme_manager(None)` returns a default-state manager, exactly as the `{{ }}` context-string form does (proven by `context_processors.py`, which calls the same bodies with `{"request": request}`).
- **kwargs**: `key=value` args from the Rust engine are coerced (quoted strings / bool / None / int / float / context-var) and passed as kwargs to the simple_tag body. An unexpected kwarg falls back to a default-args render (warning logged) rather than 500-ing the page.

## Engine-path declaration (#1051)

This is a template-rendering-pipeline change: Python-side registration of handlers the **Rust** engine calls via PyO3. **No Rust rebuild needed** — the existing Rust engine invokes the handlers. The Django-engine path is unchanged (the `@register.simple_tag` registration is untouched; verified Django-engine `{% theme_panel show_packs=False %}` still renders).

## Test plan

- **Reproducer (fail-before):** `render_template("{% theme_panel %}", {})` raised the exact reporter error → 17/17 new tests FAILED pre-fix.
- **Pass-after:** 17/17 PASS. Covers: each tag registered; `{% theme_X %}` renders without error via the Rust engine; the exact `{% load theme_tags %}{% theme_panel %}` snippet; **`{{ }}` parity** (Rust output == tag-body output for default args); **kwargs** (`{% theme_panel show_packs=False %}` == body with `show_packs=False`, and != default render).
- **Gate-off (#254):** commenting out `register_with_rust_engine()` makes all tests fail with the "Unsupported template tag" error → tests are non-tautological.
- **Suites:** theming `-k theme` → **869 passed, 1 skipped**; full `python/djust/tests/` → **3062 passed, 3 skipped, 0 failed**.

## Risk notes

- **Perf (documented tradeoff):** the `{% theme_X %}` tag form runs a Python sidecar per invocation across the PyO3 boundary. The `{{ theme_X }}` context-string form pre-renders once per request and is cheaper when a tag appears multiple times on a page. Documented in the CHANGELOG entry and the guide.
- docs.djust.org is a **separate repo** — no update needed there since this PR makes the existing docs correct against the engine. (Flagging, not editing it.)

Closes #1721


johnrtipton:
## Stage-11 Adversarial Review — PR #1728 (fix #1721)

**Verdict: APPROVE**

Pre-review gates pass; the security hunt (XSS via kwargs → HTML attributes) came back **clean** under an empirical probe; tests are green and non-tautological.

---

### Mandatory pre-review checks
- 🟢 **Stale-base (#250):** `git rev-list --count HEAD..origin/main` = **0**. `git log origin/main..HEAD` shows exactly this PR's two commits (`1b6516ad` impl+tests, `8232c746` docs+CHANGELOG). Reviewing the merge program, not a stale base.
- 🟢 **Two-commit shape:** `1b6516ad` = impl+tests, no CHANGELOG (`rust_handlers.py`, `apps.py`, test file). `8232c746` = CHANGELOG.md + `docs/website/guides/components.md` only. Matches the canonical boundary.

### 1. XSS / mark_safe surface (HIGHEST priority) — 🟢 CLEAN
The new args path (`button_class`, `dropdown_class`, `layout`, `dropdown_position`) is genuinely new surface vs. the `{{ theme_panel }}` context-string form (which takes no args). I traced each kwarg to the rendered HTML:
- kwargs flow into `ThemeSwitcher`/`ThemeModeButton`/`PresetSelector` context dicts → Django component templates → `tmpl.render(context)` → `mark_safe(html)`.
- In the templates (`theme_switcher.html:2,6,26,38,55`, `theme_mode_button.html:3`, `preset_selector_dropdown.html:2`) these land as `class="...{{ button_class }}..."` — **Django auto-escaped `{{ }}` positions**. The `mark_safe()` wraps the *already-rendered/escaped* output, not the raw kwarg.
- Only `theme_head.html` uses `{% autoescape off %}`, and it receives **no** user kwargs (only `include_js`/`link_css` booleans; CSS/script/`cookie_prefix_js` are framework-generated, `cookie_prefix_js` is `json.dumps`-encoded).
- **Empirical canary:** rendered `{% theme_switcher button_class='"><script>alert(1)</script>' %}` (and the `theme_mode_toggle`/`theme_preset_selector` equivalents) through the **Rust engine** (`djust._rust.render_template`). Result: raw `<script>` **absent**, escaped `&lt;script&gt;` **present** in all three. No XSS.
- The tag form exposes no unescaped surface the existing Django-engine `{% theme_switcher %}` form did not — it routes through the same component templates with the same auto-escape.

### 2. request=None on WS path — 🟢
`render(args, context)` does `tag_context = {"request": context.get("request")}`; `None` is fine — `get_theme_manager(None)` returns a default-state manager (the exact path the `{{ }}` form uses via `fn({"request": None})`). All 17 tests render with request=None and produce non-empty HTML. Per #1722 the Rust render context now carries `request` on the live path, so the populated case also works.

### 3. kwargs coercion — 🟢
Verified directly: `show_packs=False` → Python bool **`False`** (not the truthy string `"False"`); `false`→False, `None`→None, `0`→int 0, quoted→stripped string, bare word→`context.get(val, val)`. Unexpected kwarg → `TypeError` caught → `logger.warning` + default-args render (no 500). `theme_panel show_packs=False` empirically differs from default render.

### 4. Idempotent registration / startup safety — 🟢
`register_with_rust_engine()` called 3× without error; `has_tag_handler` guard skips re-registration (belt-and-suspenders — `register_tag_handler` would overwrite anyway). `ImportError` on `djust._rust` degrades to a debug-log no-op. 11 tags registered. Django-engine path confirmed unaffected (`{% theme_mode %}` via `django.template.Template` → `'system'`).

### 5. Parity claim — 🟢 (real equality, not both-empty)
`test_rust_tag_output_matches_context_string_form` asserts `rust_output == context_string_output`, AND `test_theme_tag_renders_via_rust_engine_without_error` asserts panel/switcher/head are **non-empty** — so parity is over substantive HTML, not two empty strings.

### 6. Test quality / gate-off (#254/#1200) — 🟢
- `test_theme_tag_is_registered_with_rust_engine`: fails if registration is a no-op (`has_tag_handler`→False).
- `test_theme_tag_renders_via_rust_engine_without_error`: pre-fix raises "Unsupported template tag" → fails. Real reproducer.
- `test_theme_panel_respects_kwargs_via_rust_engine`: includes `no_packs_rust != default_output` — the explicit anti-tautology assertion that the args path changes output. No tautologies found.
- Full theming suite: **869 passed, 1 skipped** — no regressions.

### 7. Perf-note accuracy — 🟢
CHANGELOG + docs correctly state the tag form runs a Python sidecar per invocation across the PyO3 boundary vs. `{{ }}` pre-rendering once per request. Accurate, documented tradeoff. No specific test-count claim in CHANGELOG (no #1049 recount risk).

### Nits (non-blocking)
- 🟡 `_coerce_value` does a `context.get(val, val)` lookup even though the docstring says the Rust engine pre-resolves variables. Harmless (double-resolution of a literal is a no-op), but the redundancy is slightly confusing — worth a one-line comment that this is a defensive fallback for the unresolved-token case.

**Empirical canary captured:** XSS probe clean (3/3 tags escape the payload). Tests: 17/17 pass, 869 theming-suite pass.

johnrtipton:
## Retrospective — PR #1728 (pipeline-drain v1.0.2)

**Task**: v1.0.2 — drained #1721 (Rust engine rejects the documented {% theme_panel %} tag).

### Quality: 5/5
Clean first-pass; the design decision (register tags vs docs-only) was made upfront and the XSS surface was empirically cleared.

### What went well
- **Design-decision-first (#1056).** Two viable paths existed: (A) register the theme tags as Rust handlers, (B) docs-only in the external docs.djust.org repo. Option A was chosen because it makes existing user templates AND the documented tag form (incl. args) work without editing an external repo, and it uses the framework's own intended extension point (register_tag_handler — the very mechanism the error message names). Decided before implementation, not litigated mid-PR.
- **Reused the established bridge.** ThemeTagHandler mirrors the existing TagHandler/register_tag_handler pattern (template_tags/__init__.py, components/rust_handlers.py) and delegates to the SAME theme_tags.py bodies the {{ }} context-string form uses — so the two forms can't drift.
- **Security cleared empirically (#1078 mark_safe XSS-trace).** The reviewer rendered {% theme_panel button_class='\"><script>...' %} through the real Rust engine and confirmed escaping: kwargs land in auto-escaped {{ }} positions in the component templates; mark_safe wraps already-escaped output. The args path is new surface vs the no-arg {{ }} form, but exposes no unescaped user input.
- Gate-off (#254): all 17 tests fail pre-fix with the Unsupported-tag error; parity asserted over non-empty HTML; kwargs test has an anti-tautology assertion.

### What didn't
- Nothing blocking. One 🟡 (a redundant context.get in _coerce_value — harmless, since the Rust engine pre-resolves vars). Left as-is per scope discipline; cosmetic.

### Process finding
The error message that the user hit ("Register a handler via djust._rust.register_tag_handler()") literally pointed at the fix. When an engine emits an actionable error naming an extension API, registering through that API (rather than a docs workaround) is usually the least-surprise fix — it makes existing content correct instead of asking users to rewrite templates.

### Verified
- 11 theme tags registered (the 5 cited + theme_css/theme_css_link/theme_framework_overrides/theme_preset/theme_mode/theme_resolved_mode).
- request=None safe (default-state manager); show_packs=False coerces to bool False; registration idempotent (has_tag_handler guard); ImportError degrades to no-op; Django-engine path unaffected.
- No Rust rebuild needed (Python-only registration; existing engine invokes handlers via PyO3).
- theming suite: 869 passed/1 skipped; full python/djust/tests/: 3062 passed/3 skipped/0 failed. All CI green. Two-commit shape intact.
- docs.djust.org (external repo) now matches the engine; no follow-up needed there since this PR makes the existing docs correct.

`RETRO_COMPLETE`

PR #1729, opened 2026-06-05
fix(build): generalize cross-IIFE guard to top-level-module refs (#1716)

### LINKED ISSUE #1716: build: cross-IIFE guard misses top-level-module (22-51) bare refs — 10/58 published fns gap-exposed
Source: PR #1715 (#1706) Stage 11 review.

#1706 (PR #1715) added `scripts/check-cross-iife-refs.mjs` and retired the OBSERVED #1676/#1688 cross-IIFE class. But the guard fires only for `decl.inGuard && !refInGuard` (`check-cross-iife-refs.mjs:414`) — i.e. it catches a bare reference to a function declared inside the double-load-guard `else {}` block (modules 00-20) and referenced outside it.

**Gap:** a bare cross-IIFE reference between two TOP-LEVEL modules (22-51, each in its own inner IIFE) is NOT flagged. 10 of the 58 `globalThis.djust`-published functions are declared at bundle top level and are thus gap-exposed. This is the SAME ReferenceError-under-minify class — #1676/#1688 just happened to involve guard-block publishers, so the *observed* class is retired but the *general* class is not.

Verified by the reviewer with a synthetic top-level publisher+bare-ref → check exits 0 (should be 1).

- **Fix sketch:** generalize the scope test so that ANY bare reference to a `globalThis.djust`-published function from a module that is NOT the declaring module's scope is flagged (covering both guard-block and top-level-module declarations), keeping the guard-span model only as the inGuard/outOfGuard discriminator where it's actually needed. Re-verify against the 68-FP set that motivated the current narrow form.
- **Also:** the script's documented limitation (lines 53-56) currently only mentions NON-published functions; update it to note this published-top-level gap until fixed.
- **Minor (same PR):** the `dj_dialog`/`keyboard_nav` stub tests set both bare + alias globals, so they pass even against the pre-fix bundle in JSDOM — the static lint is the real regression gate, but the CHANGELOG slightly oversells the stub change. Consider tightening the stubs to spy ONLY the alias.

johnrtipton:
## Summary

The #1706 cross-IIFE bare-reference static guard (`scripts/check-cross-iife-refs.mjs`) caught a bare reference to a `globalThis.djust`-published function declared inside the double-load-guard `else {}` block (modules 00-20) only when referenced from bundle top level. Its narrow scope test — `if (!(decl.inGuard && !refInGuard)) return;` — left a gap (#1716, found in PR #1715/#1706 Stage 11 review): a bare reference between two **top-level** modules (22-51), each wrapped in its own inner IIFE, was **not** flagged. 10 of 58 published functions are declared at bundle top level and were thus exposed to the same `ReferenceError`-under-terser class as #1676/#1688.

## Generalized scope rule (one sentence)

Each published function declaration is assigned a **lexical scope-barrier span** — its innermost enclosing function/IIFE body if any, else the double-load-guard `else {}` block if it sits directly inside it, else `null` (program scope) — and a bare cross-module reference is flagged **iff** the reference's line falls *outside* the declaration's barrier span (program-scope declarations have no barrier and are never flagged).

This covers BOTH dangerous shapes with one model:
- **guard-block → top-level** (#1676/#1688) — unchanged behavior.
- **top-level → top-level** (#1716) — the closed gap.

## Empirical canary (#252)

Synthetic `22-foo.js` declares `function foo(){}` + `globalThis.djust.foo = foo;` inside its IIFE; `23-bar.js` references `foo()` **bare** from its own IIFE:
- **OLD code → exit 0** (the gap; the canary test fails before the fix).
- **NEW code → exit 1**, naming `foo`, `23-bar.js`, `22-foo.js`, `published via globalThis.djust.foo`.

## No new false positives

Real tree stays clean: `node scripts/check-cross-iife-refs.mjs` → **exit 0** (54 modules, 58 published functions). The load-bearing FP control is `42-dj-remove.js`'s `maybeDeferRemoval`, which is declared at **program scope** (outside any IIFE) — a true global visible everywhere even minified. A naive "different module ⇒ bug" rule would false-positive on it; the barrier model correctly leaves it unflagged (pinned by a dedicated test).

**None of the 10 gap-exposed published functions are referenced bare anywhere in the real source** — so no real-source `globalThis.djust.X` rewrite was needed; this is a pure lint generalization.

## Gate-off self-test (#254)

Reverting only the IIFE-barrier generalization (guard-only barrier model restored) makes **exactly** the #1716 canary fail (exit 0 instead of 1) while the #1688 guard-block test still passes — proving the canary is non-tautological.

## Tests

- Extended `tests/js/check-cross-iife-refs-1706.test.js`: top-level↔top-level canary (new flags / old didn't), member-access not-flagged, intra-IIFE not-flagged, **program-scope (`maybeDeferRemoval`-shape) not-flagged** (FP control), local-rebind not-flagged. Existing #1688 guard-block case still asserts flagged; real-tree-clean still asserts exit 0.
- `npm test`: **1665 passed, 0 failed** (144 files).
- Updated the documented limitation block in the script + test header to note the published-top-level gap is now CLOSED (remaining limitation: non-published inner-IIFE refs only).
- Optional dj_dialog/keyboard_nav stub-tightening: **deferred** (out of scope per the issue's minor note; not needed for the fix, no low-risk path verified — left for a follow-up).

Closes #1716


johnrtipton:
## Stage-11 adversarial review — PR #1729 (fix #1716: generalize cross-IIFE guard to top-level-module refs)

**Verdict: APPROVE** 🟢

A clean, well-modeled tooling PR. The empirical canary passes independently, the FP control is sound, the gate-off self-test proves the new test is non-tautological, and the real tree stays clean. No 🔴, no over-reach FP, no missed-bug case found.

### Pre-review gates
- 🟢 **Stale-base (#250):** `git rev-list --count HEAD..origin/main` = **0**. Only this PR's two commits on the branch (`0b3f1754` impl+tests, `f701f8ba` CHANGELOG).
- 🟢 **Two-commit shape (#181):** `0b3f1754` touches only `scripts/check-cross-iife-refs.mjs` + `tests/js/check-cross-iife-refs-1706.test.js` (no CHANGELOG); `f701f8ba` touches only `CHANGELOG.md`. Correct boundary.

### Empirical canary — re-run independently (#252), NOT trusted from the implementer
My own synthetic #1716 trigger (`/tmp/.../synth1716`: `22-a.js` publishes `function bar(){}; globalThis.djust.bar = bar;` inside its IIFE; `23-b.js` references `bar()` bare inside its own IIFE):

| Case | This branch | origin/main | Expected | Result |
|---|---|---|---|---|
| **#1716** top-level→top-level synthetic trigger | **exit 1** (flags `bar`, names `22-a.js`/`23-b.js`) | **exit 0** (the gap) | 1 / 0 | 🟢 PASS — PR closes the cited class |
| **#1688** guard-block→top-level (regression) | **exit 1** | **exit 1** | 1 / 1 | 🟢 PASS — old class still caught on both |
| **Real tree** (`node scripts/check-cross-iife-refs.mjs`) | **exit 0** ("clean across 54 modules, 58 published") | — | 0 | 🟢 PASS — no false positives |

The branch↔main exit-code split (1 vs 0) on the #1716 trigger is the load-bearing proof: the generalization actually catches what it claims, and main genuinely had the gap.

### FP control — `maybeDeferRemoval` program-scope claim VERIFIED
- `python/djust/static/djust/src/42-dj-remove.js`: no IIFE wrapper (program scope), `function maybeDeferRemoval(el)` declared at L191 after the guard `else {}` closes (module 21) → barrier = **null** → never flagged. ✔
- Instrumented `analyze()` on the real tree: of 58 published functions, exactly **1** has a null barrier (program scope) = `maybeDeferRemoval`. ✔ This is precisely the load-bearing FP control.
- Real cross-module uses of it (`12-vdom-patch.js`) correctly use the `globalThis.djust.maybeDeferRemoval` member form, which is never flagged regardless.

### Adversarial over-reach / missed-bug hunt (synthetic dirs, all behaved correctly)
- 🟢 **Nested-closure decl, intra-module ref** — not flagged (same-module short-circuit fires before the barrier test). No false positive.
- 🟢 **Guard-block decl, ref from ANOTHER guard module** (shared block scope) — not flagged (both inside guardSpan). Correct; this is the bulk of legit cross-module calls.
- 🟢 **Program-scope `const arrow` published, bare cross-module ref** — not flagged (barrier null). Correct; the FP control generalizes beyond `FunctionDeclaration`.
- 🟢 **Guard-block ref → top-level-IIFE function** — *correctly flagged* (exit 1). The new barrier model catches this genuine risk that direction too; not over-reach, it's a real ReferenceError-under-minify path.
- 🟢 **`globalThis.djust.X = function(){}` (member-assigned fn expr) in IIFE, bare cross-ref** — publishes 0 (RHS isn't a bare Identifier), so not in P; a bare `memFn()` elsewhere is a plain ReferenceError caught by JS load tests, not this check's class. Consistent with the documented limitation, no regression.

### Barrier-span computation soundness
The custom `visit` records each declaration's barrier from the **current** `fnScopeStack` *before* descending into the body it opens (comment at L328-329 matches behavior). `barrierFor` correctly prefers innermost function scope, falls back to the guard `else {}` only when the function stack is empty (guard block is not a function scope, never pushed), else null. Declaration-vs-assignment detection (`FunctionDeclaration` / fn-`VariableDeclarator` / fn-`AssignmentExpression`) is unchanged in coverage from main. Sound.

### Test quality (#254/#1200) — gate-off self-test re-run
I swapped in origin/main's pre-generalization script and ran the extended test file: **exactly 1 test failed — the #1716 canary** — while the #1688 case, member-access, intra-module/intra-IIFE, local-rebind, and the program-scope (`progFn`/maybeDeferRemoval-shape) controls all still passed (10/11). This confirms the new catch-test is tied to the change and non-tautological, exactly as the implementer claimed. Full suite on the branch: **11/11 pass.**

### Nits (🟡, non-blocking)
- 🟡 The "10 of 58 published functions are top-level-declared" figure (CHANGELOG / comment L49 / test L17) is descriptive prose, not load-bearing; I verified 58 published and 1 program-scope but did not independently re-derive the guard-vs-IIFE split of 10. Fine to leave.
- 🟡 `buildPublishedSet`'s docstring (L404) still says `{ moduleIndex, inGuard, bundleLine }` — a stale leftover; the struct is now `{ moduleIndex, barrierSpan, bundleLine }` (correctly updated everywhere it's consumed). One-line comment drift only.
- 🟢 Documented-limitation comment (L67-72) correctly updated to state the published-top-level gap is closed, leaving only the non-published inner-IIFE follow-up.

Working tree restored to clean branch HEAD (`f701f8ba`); only the pre-existing untracked `scripts/antigravity_pipeline_run.py` remains.

johnrtipton:
## Retrospective — PR #1729 (pipeline-drain v1.0.2)

**Task**: v1.0.2 — drained #1716 (generalize cross-IIFE guard to top-level-module refs).

### Quality: 5/5
Tooling-class PR done by the book: empirical canary proven on both branch and main, false-positive control verified, scope-aware model that subsumes the prior narrow form.

### What went well
- **Empirical canary on BOTH sides (#252).** The implementer AND the reviewer independently ran a synthetic #1716 trigger (top-level publisher + bare cross-module ref): exit 1 on branch / exit 0 on main — the 1-vs-0 split is the load-bearing proof the PR closes the cited class and that main genuinely had the gap. The #1688 guard-block class still exits 1 (no regression).
- **The fix found the real distinction: scope, not module.** A naive "different module = bug" rule would false-positive on `maybeDeferRemoval`, which is declared at PROGRAM scope (a true global, no IIFE wrapper) and legitimately referenced bare cross-module. The barrier-span model (innermost enclosing function/IIFE, else guard else block, else null=program=never-flagged) correctly distinguishes the 9 IIFE-nested top-level published fns from the 1 program-scope one. The reviewer instrumented analyze() on the real tree and confirmed exactly 1 of 58 is program-scope and it's maybeDeferRemoval.
- Gate-off (#254): swapping in main's pre-generalization script makes exactly the #1716 canary test fail (10/11), FP controls + #1688 still pass — non-tautological.

### What didn't
- Nothing blocking. 2🟡 cosmetic doc nits (a stale `inGuard` mention in one docstring now that the struct is `barrierSpan`; an unverified "10 of 58" prose breakdown — the headline 58/1 counts are correct). Optional stub-tightening (dj_dialog/keyboard_nav) deferred per scope discipline.

### Process finding
For a static-analysis generalization, the highest-value check is the two-sided empirical canary: prove the NEW code catches the synthetic trigger AND the OLD code didn't, plus the real tree stays clean. The reviewer building 6 adversarial synthetic dirs (nested-closure, guard-shared-scope, program-scope const-arrow, member-assigned fn-expr, guard→top-level) is the right depth for a lint that will gate every future PR.

### Verified
- Real tree clean (exit 0, 54 modules / 58 published fns). None of the 10 gap-exposed fns were referenced bare in real source — pure lint generalization, no source rewrite needed.
- npm test 1665 passed / 0 failed; check wired into .pre-commit-config.yaml + CI test.yml. All CI green. Two-commit shape intact.

`RETRO_COMPLETE`

Issue #1447, opened 2026-05-09
tech-debt: cache-by-struct discipline — include all fields upfront


johnrtipton:
## Source

Retro v0.9.6-1, finding #4. Surfaced by PR #1442 (#1437 theme_context lru-cache) — initial cache key missed `theme` and `layout` fields of `ThemeState`.

## Pattern

When wrapping a function whose inputs are derived from a struct, the cache key should include EVERY field of the struct. Pruning a field later (because profiling shows it doesn't matter) is a one-line change with a regression test. *Adding* a field later means cache-poisoning bugs in production — two callers with different field-N values get the same cached output.

## What to do

Add to `.pipeline-templates/feature-state.json` Stage 4 (Planning) — a checklist item:

```
- action: when introducing an `lru_cache` / `@cache` keyed by fields-of-a-struct, list every field of that struct in the plan and explicitly justify any omissions
  mandatory: false
```

Also add a CLAUDE.md Process Canon entry citing PR #1442 as the case study.

## Acceptance

- [ ] Stage 4 plan template item added.
- [ ] CLAUDE.md canon section added.

johnrtipton:
Closed by PR #1451 (canon batch). GitHub's auto-close keyword only picked up the first issue in the comma-list 'Closes #1445 #1446 #1447 #1450' — closing manually.

Issue #1078, opened 2026-04-25
tech-debt: Stage 11 mark_safe XSS-trace audit bullet


johnrtipton:
**Source**: Retro v0.8.2 / Action Tracker row #153 / PR #1074 (#1012 theme_css_link)

**Action**: Add to Stage 11 review-checklist prompt in `~/.claude/skills/pipeline-run/SKILL.md`: for every new `mark_safe` call, trace inputs to a server-validated source. Pattern is empirically working — PR #1074's reviewer subagent specifically traced cookie inputs through `registry.has_theme/has_preset` validation in `get_state()` to the URL output, and confirmed there's no XSS surface. The audit added value but was reviewer-discretionary; making it a checklist item locks the discipline.

**Implementation**: one bullet under Stage 11 review-checklist's security-trace section. Optional parallel addition to djust's `docs/PULL_REQUEST_CHECKLIST.md` if the bullet is also appropriate at the project level. Skill-level work — labeled `out-of-scope-for-djust-drain`.

---
_Filed by /pipeline-retro v0.8.2 milestone retro Stage 4._

johnrtipton:
Shipped in PR #1226 (v0.9.1-7 cleanup batch). Comma-list `Closes #...` form parsed only the first ref — the rest closed manually. (This bites again; #1184's parenthesized form is documented in PR-checklist, but the comma-list-on-its-own-line form has the same auto-close failure.) 🤖 v0.9.1-7 cleanup batch closure.

Issue #247, opened 2026-02-08
Add deployment guide for Railway/Render/Fly.io


johnrtipton:
## Description

Help Django developers deploy djust apps to popular hosting platforms. Each platform needs a step-by-step deployment guide.

## Platforms to Cover

- Railway
- Render
- Fly.io
- DigitalOcean App Platform

## What Each Guide Needs

1. Prerequisites
2. Configuration (environment variables, Procfile/Dockerfile)
3. WebSocket/Channels setup (this is the tricky part for djust)
4. Database setup
5. Static files configuration
6. Step-by-step deployment walkthrough
7. Troubleshooting common issues

## Where to Add

- `docs/guides/deploy-railway.md`
- `docs/guides/deploy-render.md`
- etc.

## Notes

- The main challenge is WebSocket support - each platform handles this differently
- djust requires ASGI (daphne/uvicorn), not WSGI
- Reference the existing Dockerfile for container-based deployments

Pick one platform and submit a guide!

PR #1732, opened 2026-06-05
perf(theming): request-scope memoize theme_context tag bodies (#1727)

### LINKED ISSUE #1727: perf(theming): request-scope memoize theme_context — now runs per WS event after #1722
Source: PR #1726 (#1722) Stage 11 review finding 🟡 PERF-1.

#1726 fixed #1722 by applying `_apply_context_processors` inside `_sync_state_to_rust` (`python/djust/mixins/rust_bridge.py`), which runs on **every WebSocket event**, not just the initial HTTP GET. That correctly makes context-processor vars (`{{ theme_panel }}`, `{{ theme_head }}`) available inside `{% include %}`d partials on every render — including live updates and navigation.

**The cost:** `djust.theming.context_processors.theme_context` now runs once per WS event for projects using djust theming. Resolution is cached (`_get_resolved_processors`) and the dominant CSS render (`_render_theme_outputs`) is `@lru_cache`'d, but `theme_context` ALSO calls four template-tag bodies via `_safe_render` (`theme_head` / `theme_panel` / `theme_mode_toggle` / `theme_preset_selector`, context_processors.py:164-167) that are NOT cached — each does `get_theme_manager` → `get_state` → `get_available_presets` → template render. Pre-#1726 these ran once per HTTP request; post-#1726 also once per WS event.

**Do NOT** "fix" this by gating context-processor application to first-sync-only — that would break dynamic theme switching (change-detection only forwards *changed* vars, so the processors must re-run each sync to detect a theme change). 

**Correct fix:** request-scoped memoization INSIDE `theme_context` (cache the four `_safe_render` outputs keyed on the resolved theme state for the lifetime of a request / until `clear_theme_context_cache`), so re-running `theme_context` per WS event is cheap when the theme state hasn't changed but still reflects a switch when it has.

**Acceptance:** a per-WS-event render with unchanged theme state does NOT re-run the four tag bodies (assert via a call-count spy); a theme switch DOES produce updated `theme_panel`/`theme_head`. Bounded to theming users; non-blocking (filed as a follow-up so #1726 could merge).

johnrtipton:
## Summary

`#1722` made `theme_context` run on **every WebSocket event** via `_apply_context_processors` (`rust_bridge.py:548`). Each call re-rendered four **uncached** tag bodies through `_safe_render` (`context_processors.py:164-167`): `theme_head`, `theme_panel`, `theme_mode_toggle`, `theme_preset_selector`. `_render_theme_outputs` (CSS/switcher) was already `@lru_cache`'d; these four were not — so theming users paid four uncached tag renders per WS event (Stage 11 finding 🟡 PERF-1 on PR #1726).

This adds **request-scoped memoization** of the four `_safe_render` outputs, keyed on the resolved theme-state tuple (`theme, preset, pack, mode, resolved_mode, layout, presets_key` — the same shape `_render_theme_outputs` keys on).

## Per-request-data finding (drives the design)

I traced all four tag bodies + `build_theme_head_context`. **None embed a CSP nonce or any per-request value.** `theme_head` emits anti-FOUC script, CSS blocks, deferred-CSS preload `<link>`, component CSS, direction, and `cookie_prefix_js` — and `cookie_prefix_js` derives from `config.get("cookie_namespace")`, a **config** value, not the request. No `request.csp_nonce` reference anywhere. The other three derive purely from `manager.get_state()` + `get_available_presets()` + static pack/design/layout lists; `request` is used only to resolve the manager + template candidates (themselves keyed on `state.theme`, part of the state tuple).

## Caching scope: request-scoped (why)

Even though outputs depend only on theme state today, I chose **request-scoped** caching (stored as `request._djust_theme_ctx_cache = (state_key, outputs)`) over a cross-request module-level cache:
1. The issue title specifies request-scoped.
2. It is **strictly safe** — a request-scoped cache cannot leak a future per-request value (e.g. a nonce added later) across requests, where a module-level cache would.
3. On the WS path `request` is a long-lived instance attr set once in `handle_connect`, so the cache naturally spans all events of a connection and is invalidated by the state key changing on a theme switch.

No first-sync gating (CLAUDE.md v1.0.2 canon): the processor still runs every event; only the expensive sub-renders are memoized. `clear_theme_context_cache()` needs no change — the cache is request-scoped, invalidated by the state key; the existing `_render_theme_outputs.cache_clear()` is untouched.

## Tests (`python/djust/tests/test_theme_context_memoize_1727.py`)

- `test_unchanged_state_renders_tag_bodies_once` — **acceptance**: two calls on the same request with unchanged state → each tag body invoked **once** (2nd call memoized). Fail-before (count=2) / pass-after.
- `test_theme_switch_recomputes_tag_bodies` — a mode/preset switch on the same request **recomputes** and yields updated output (dynamic switching preserved).
- `test_distinct_requests_do_not_share_cache` — request-scoped isolation: two distinct requests each render (no cross-request sharing → no nonce leak vector).
- `test_same_request_unchanged_state_returns_identical_keys` — full context contract preserved across the memoized call.
- `test_no_request_attr_does_not_crash` — a request that can't hold attributes falls back to rendering each call.

### Call-count evidence
Before fix: `theme_head tag body re-rendered on the 2nd call (count=2)` → assertion fails. After fix: 5/5 pass (each tag count==1 on the unchanged-state path).

### Gate-off (#254)
Disabling only the cache lookup (`if False and cached ...`) makes `test_unchanged_state_renders_tag_bodies_once` fail again (count=2) — non-tautological. Restored.

## Test suite
- New file: 5 passed.
- `pytest -k theme`: 874 passed, 1 skipped.
- Full `python/djust/tests/`: **3067 passed, 3 skipped**.
- ruff check + format: clean.

Closes #1727


johnrtipton:
## Stage 11 Adversarial Review — PR #1732 (fix #1727: request-scope memoize `theme_context` tag bodies)

**Verdict: 🟢 APPROVE.** The cache key is complete, no per-request/nonce data is embedded, request=None is safe, dynamic switching is preserved, and the test suite is non-tautological (gate-off verified). No 🔴, no 🟡.

### Pre-review gates
- **Stale-base (#250):** `BEHIND=0` — branch is current with `origin/main`. `git log origin/main..HEAD` shows exactly the two PR commits (`ade2486c` impl+tests, `f1180747` CHANGELOG-only). ✅
- **Two-commit shape:** impl commit touches only `context_processors.py` + `test_theme_context_memoize_1727.py` (no CHANGELOG); second commit is CHANGELOG-only. ✅

### 1. Cache-key completeness (highest priority) — 🟢 COMPLETE
`state_key = (theme, preset, pack, mode, resolved_mode, layout, presets_key)` (`context_processors.py:187-195`).

I enumerated every field each of the four cached tag bodies actually consumes and confirmed the key captures all per-request-state inputs:

- **`theme_head`** → `build_theme_head_context` (`theme_tags.py:44-180`): reads `state.{theme,preset,pack,mode}` (all in key). Its other inputs — `get_theme_config()` (`critical_css`, `css_prefix`, `cookie_namespace`, `direction`), `get_css_prefix()`, `get_direction()`, `apps.is_installed("djust.components")` — are **process-global config**, not request-state. They cannot change within a connection's lifetime.
- **`theme_panel`** (`theme_tags.py:406-474`): reads `state.{pack,theme,layout,mode}` (all in key) + `presets` (in key via `presets_key`) + `get_all_design_systems()` / `get_all_theme_packs()`. The pack/design registries (`theme_packs.py:833,1225`) are static module-level dicts with **no reload/mutation API** — they're import-time constants, so omission from the key is safe.
- **`theme_mode_toggle`** (`ThemeModeButton`, `components.py:103-116`): reads `state.{mode,resolved_mode,theme}` — all in key.
- **`theme_preset_selector`** (`PresetSelector`, `components.py:146-185`): reads `state.{preset,theme}` + presets — all in key. `layout`/`show_descriptions` are fixed default args (the memoized render reproduces the default `{{ theme_preset_selector }}` shape; customization-with-args still flows through the live `{% %}` tag).

**Key fact:** the new request-scoped key uses the *identical 7-tuple shape* as the pre-existing `_render_theme_outputs` `@lru_cache` (`context_processors.py:36`, `128-136`), which has been the codebase contract for treating theme config/catalogs as process-static. The request-scoped cache is strictly *safer* than that module cache (dies with the request vs. spanning all requests). I attempted to construct a "change X, key unchanged → stale" scenario for each omitted field; in every case X is process-global and invariant within a connection. **No omitted field found.**

### 2. Nonce / per-request-data claim — 🟢 VERIFIED
Grepped `theme_head.html`, `components/theme_panel.html`, and the `ThemeModeButton`/`PresetSelector` renderers: **zero** `nonce`/`csp_nonce`/`request.`/`csrf` interpolation in any of the four cached outputs. The csrf/cookie hits in the theming module are confined to `gallery/` and `pages/` auth templates — none rendered by the cached tags. `cookie_prefix_js` confirmed config-derived (`theme_tags.py:166-168`: `config.get("cookie_namespace")`), not request-derived. The implementer's claim holds; no security surface.

### 3. Request identity / reuse + request=None — 🟢 SAFE
- **WS path:** `self.view_instance.request = request` is set **once** in connect/mount (`websocket.py:1963`), a real `HttpRequest` that accepts attribute assignment. The same object is read on every event via `getattr(self, "request", None)` (`rust_bridge.py:546`), so the cache correctly persists across a connection's events and is per-connection isolated.
- **HTTP path:** per-request object → single-use cache, fine.
- **request=None:** `_apply_context_processors` is a no-op when request is None (`rust_bridge.py:547`), so theme_context's memo branch isn't reached on that path. Even if reached directly, `getattr(None, ...)` returns None (miss) and the `try/except (AttributeError, TypeError)` swallows the store failure. `test_no_request_attr_does_not_crash` covers a bare `object()`. ✅

### 4. No first-sync-gating — 🟢 VERIFIED
The processor runs `manager.get_state()` and computes `state_key` on **every** call before any short-circuit; only the four sub-renders are memoized. A mode/preset/theme switch changes the key → cache miss → recompute. Verified by reading the flow and by `test_theme_switch_recomputes_tag_bodies`.

### 5. `clear_theme_context_cache` interplay — 🟢 sound
Request-scoped caches die with the request; a theme-pack reload between requests yields a fresh request with no stale entry. Within a connection a reload would not invalidate the request cache, but the pack registry has no runtime reload path, so this is moot. Reasoning holds. (Note for maintainers: if a runtime pack-reload mechanism is ever added, both the module `lru_cache` AND this request cache would need explicit invalidation — but that's a future concern, not a defect here.)

### 6. Test quality (#254/#1200) — 🟢 non-tautological, gate-off verified
- Ran the suite: **5/5 passed** (`pytest python/djust/tests/test_theme_context_memoize_1727.py`).
- **Gate-off self-test:** I forced a permanent cache miss (`cached = None`) and re-ran — `test_unchanged_state_renders_tag_bodies_once` **fails** (call_count 2≠1), the other 4 pass. The acceptance-criterion test genuinely exercises the change; not a tautology. File restored.
- Confirmed the MagicMock-as-request path is real (not spurious): a fresh `MagicMock`'s `__eq__` against a real tuple returns `False`, so the first call correctly takes the MISS branch; the explicit `request._djust_theme_ctx_cache = ...` set makes the second call a real HIT.
- `test_distinct_requests_do_not_share_cache` proves request isolation (2 renders per tag). `test_theme_switch_recomputes_tag_bodies` exercises a real state change (light/default → dark/ocean). `test_same_request_unchanged_state_returns_identical_keys` pins the full-dict contract.

### Working tree
Restored to clean branch HEAD (`f1180747`); only the pre-existing untracked `scripts/antigravity_pipeline_run.py` remains.

johnrtipton:
## Retrospective — PR #1732 (pipeline-drain v1.0.2)

**Task**: v1.0.2 — drained #1727 (request-scope memoize theme_context; the per-WS-event cost surfaced by #1726).

### Quality: 5/5
Clean first-pass perf fix that honored the v1.0.2 canon it descended from, with the cache-key-completeness and nonce-leak risks both explicitly cleared.

### What went well
- **Applied the canon this milestone just wrote.** #1727 IS the follow-up the new CLAUDE.md v1.0.2 rule ("per-event work feeding change-detection must be memoized, not first-sync-gated") was written for. The fix memoizes the four uncached _safe_render tag bodies rather than gating context-processor application — dynamic theme switching preserved (verified: a mode/preset switch recomputes).
- **Nonce-leak risk investigated before choosing scope (#1078-style trace).** The implementer traced all four tag bodies + build_theme_head_context and confirmed NO request.csp_nonce / per-request value is embedded (cookie_prefix_js is config-derived). Chose request-scoped caching anyway (store on the request object keyed by state-tuple) — strictly safe against a future nonce, matches the issue intent, and spans the long-lived WS request.
- **Cache-key completeness verified (#247/#1447).** The reviewer enumerated every field each tag output consumes and confirmed the 7-tuple key (theme, preset, pack, mode, resolved_mode, layout, presets_key — identical to the existing _render_theme_outputs lru) captures all output-affecting state; the rest are process-global config. No omitted field → no stale-HTML risk.
- Gate-off (#254): disabling the cache lookup re-breaks the call-count acceptance test (2 calls → 2 renders) — non-tautological.

### What didn't
- Nothing. 0🔴/0🟡. One non-blocking maintainer note: if a runtime theme-pack reload is ever added, both the module lru_cache and this request cache would need explicit invalidation — no such path exists today.

### Process finding
A memoization PR's load-bearing review is cache-key completeness: the key must include every field that affects the cached output, or a change to an omitted field serves stale data (#1447). Reusing the exact key-shape of an adjacent, already-trusted cache (_render_theme_outputs) is a cheap way to inherit that correctness — and the reviewer still independently enumerated the dependencies rather than trusting the shape.

### Verified
- Per-request-data: none embedded (no nonce); request=None safely no-ops; WS path reuses one long-lived request (cache spans events, per-connection isolated).
- Call-count 2→1; theme switch recomputes; distinct requests don't share cache.
- New test 5/5; pytest -k theme 874 passed/1 skipped; full python/djust/tests/ 3067 passed/3 skipped. All CI green. Two-commit shape intact.

`RETRO_COMPLETE`

Issue #1734, opened 2026-06-06
feat(nav): auto_navigate — Turbo-Drive <a> interception (opt-in flag)


johnrtipton:
Source: /pipeline-strategy 2026-06-05 (auto-navigation), Path 2 — Cluster B. **DIRECTIONAL — see ADR-021.** Depends on the route-map foundation (Cluster A).

**Goal.** With `LIVEVIEW_CONFIG['auto_navigate']=True`, plain `<a href>` links do SPA nav automatically (Turbo-Drive model) — no `dj-navigate` attribute needed.

**Design (route-map-scoped interception, NOT force-every-link):**
- Single delegated listener on `document` (`e.target.closest('a')`); replaces per-element binding (18-navigation.js:414).
- **Skip → let the browser handle** when: modifier key / middle-click; `target` set or `download`; external origin / non-http scheme (`mailto:`/`tel:`); hash-only (`#…`); explicit opt-out (`data-no-navigate` / `dj-navigate=\"false\"`); **path not in `window.djust._routeMap`** (this is what makes admin/logout/auth/static/external fall through safely).
- **Same-view, query-only diff → `live_patch`** (preserve state); **else `live_redirect`**. (This subsumes most `dj-patch` filter/tab/pagination uses — plain `<a href=\"?…\">` patches automatically.)
- Compose with ADR-013 View Transitions (auto-nav patches already go through applyPatches).

**Gating.** `auto_navigate` config flag, **default OFF in 1.1.0** (opt-in; flips behavior for every link → soak before any default-on; same flag pattern as `hot_reload_auto_enable`). Default-on is a future-major decision (ADR-021 records the eventual destination).

**High-risk areas for adversarial review:** the same-view→live_patch branch (state preservation + browser back/forward); the skip-rule matrix (esp. new-tab via modifier/middle-click).

Target: **1.1.0** (after Cluster A soaks one release).

Issue #1713, opened 2026-06-05
ci: promote demo djust_check dogfood from continue-on-error to a blocking gate


johnrtipton:
Source: PR #1712 (#1708) + rc4 retro finding #3.

#1708 (PR #1712) added a `djust_check` dogfood step against the demo project, set `continue-on-error: true` per the rc4 canon ("keep new CI jobs non-blocking until they've shipped green at least once"). It **has now shipped green** on the runner (first iteration, `playwright-tests` job, exit 0: 104 total / 0 errors / 0 deprecated-attr findings).

**Next step (this issue):** flip it to enforcing. Recommended shape (per PR #1712's inline note + the Stage-11 review): extract it into a dedicated `demo-checks` job (so a demo-check failure isn't entangled with playwright flakiness) WITHOUT `continue-on-error`, so a future re-introduced dead `@click`/legacy-attr actually red-bars the PR.

Also (Stage-11 minor): the wrapper's `severity == "error"` arm (`scripts/ci_djust_check_demo.py`) is never exercised end-to-end (the demo has 0 errors) — when extracting, consider a tiny unit test that feeds the wrapper a synthetic error-severity JSON payload so both gate arms (error-severity AND the T001/T014/T015 ID set) are covered.

PR #1736, opened 2026-06-06
feat(nav): auto-derive + auto-emit dj-navigate route map — zero-wiring SPA nav (#1733)

### LINKED ISSUE #1733: feat(nav): auto-wire dj-navigate route map (zero-wiring SPA nav)
Source: /pipeline-strategy 2026-06-05 (auto-navigation), Path 2 — Cluster A (foundation).

**Problem.** `dj-navigate` is documented as SPA-over-WebSocket but silently full-reloads unless the developer manually wires `live_session()` + emits `get_route_map_script()`. That wiring is undocumented (navigation.md omits it), and `get_route_map_script`'s docstring (routing.py:179) cites a non-existent `{% djust_route_map %}` tag. A downstream app (rent, on 1.0.2rc1) hit exactly this.

**Make it work with zero wiring:**
1. **Auto-derive the route map from the URLconf** — walk `django.urls.get_resolver()`, collect every pattern whose callback resolves to a `LiveView` subclass (incl. `__wrapped__` for login_required), convert `<int:id>`→`:id`, apply the `FORCE_SCRIPT_NAME` sub-path prefix (#987). `live_session()` no longer required for routing (keep it for WS session grouping; existing manual wiring stays valid — merge/idempotent).
2. **Auto-emit via `{% djust_client_config %}`** (already in every scaffolded base `<head>`) — fold the route-map `<script>` (CSP nonce via `get_csp_nonce`) into that tag. `window.djust._routeMap` populated on every page, no template change. (`<head>` shell renders via Django engine — no Rust-tag concern.)
3. **Fix docs** — remove the phantom `{% djust_route_map %}` from `get_route_map_script`'s docstring; rewrite navigation.md to document zero-wiring + correct the "Full page navigation" summary-table row for dj-navigate.
4. **System check** (T-category) — warn when `dj-navigate` appears in templates but the derived route map is empty.
5. **Fold in #1361** — tighten `routeMap[pathname]` access (Map / own-prop guard) since we're expanding route-map usage.

**Acceptance:** a scaffolded app with `{% djust_client_config %}` in base + `dj-navigate="/units/"` does SPA nav with NO live_session, NO get_route_map_script, NO context processor. Non-LiveView links (admin/logout/external) unaffected. Backward-compatible with existing live_session wiring.

Target: **1.0.2-1** (drain bucket → ships in 1.0.2). Prerequisite for #B (auto_navigate).

johnrtipton:
Closes #1733. Refs ADR-021 (Stage 1).

Makes `dj-navigate` work with **zero wiring**: the client route map is auto-derived from the Django URLconf and auto-emitted, so SPA navigation no longer requires manual `live_session()` + `get_route_map_script()` plumbing.

## Five deliverables

1. **Auto-derive the route map** — `routing.build_route_map_from_urlconf()` walks `get_resolver()` over `url_patterns` recursively (descending `include()` `URLResolver`s, accumulating the prefix), collecting `{js_path: "module.QualName"}` for every route whose callback resolves to a `LiveView` subclass. Mirrors `live_session`'s idioms (`RoutePattern._route` / `RegexPattern._regex`, `callback.view_class` + `__wrapped__.view_class`, `<int:id>`→`:id` via the shared `_PARAM_RE`). Applies the `FORCE_SCRIPT_NAME` prefix via `get_script_prefix()`. Module-level cache keyed by `(urlconf, script_prefix)` with `_reset_route_map_cache()` for tests. Empty `{}` when no LiveViews.
2. **Auto-emit via `{% djust_client_config %}`** — extracted a shared `_client_config_html(request)` helper that emits the api/sse meta tags (unchanged) plus the route-map `<script>` (via `get_route_map_script`, which merges derived + `live_session` entries). The tag is now `takes_context=True` to read `request.csp_nonce`. **The Rust-engine `ClientConfigTagHandler` calls the same shared helper**, so the dual-engine byte-parity invariant (PR #993) is preserved. Empty-safe: no script when the map is empty.
3. **Docs** — removed the phantom `{% djust_route_map %}` docstring reference; added a "Zero wiring required" section to `navigation.md` and fixed the `dj-navigate` description + summary-table row (it's SPA-nav over the WebSocket, full reload only as a fallback). `scripts/check-doc-snippets.py` passes.
4. **System check `djust.T016`** (next free T-id after T015) — warns when `dj-navigate` appears in templates but the derived route map is empty. Suppressible via `DJUST_CONFIG['suppress_checks']`. Dogfooded on the demo (clean — 85 derived routes, T016 silent).
5. **#1361** — `resolveViewPath` was already `Object.entries` own-prop-safe; tightened the contract comment (do-not-reintroduce `routeMap[pathname]`) now the map is populated by default, and added a polluted-`Object.prototype` regression test.

## Engine-path declaration (#1051)
`{% djust_client_config %}` renders in the outer page `<head>` via Django's engine in scaffolded apps, but is **also dual-registered with the Rust engine**. The shared `_client_config_html` helper keeps both byte-identical (parity test green). The route-map `<script>` is plain HTML output.

## Security trace (#1078)
The emitted route JSON is **developer-defined URL config** (view module paths + URL patterns), not user input. `json.dumps` escapes `< > & "`; the CSP nonce is `format_html`-escaped. No new user-input-to-HTML surface. The `pathname` lookup in `resolveViewPath` is prototype-pollution-immune (`Object.entries`, never bracket-indexing).

## TDD / gate-off evidence (#254)
RED-first for all suites. Gate-off: reverting `build_route_map_from_urlconf` to `return {}` makes both `test_client_config_emits_route_map_script` and `test_derives_liveview_routes` FAIL (confirmed), then restored to GREEN.

## Tests
- `test_route_map_derivation.py` (8): LiveView incl login_required-wrapped, `<int:id>`→`:id`, FORCE_SCRIPT_NAME, nested include, non-LiveView ignored, empty, cache/reset.
- `test_client_config_tag.py` (+5): route-map emission with/without nonce, empty-safe, api/sse meta preserved, json-escaped — dual-engine parity still green.
- `test_checks.py` (+4, T016 canary): fires w/o routes, silent with routes, silent w/o dj-navigate, suppressible.
- `test_navigation.py` (+1, updated 2): `get_route_map_script` merge behavior.
- `tests/js/navigation.test.js` (+1): polluted-`Object.prototype` #1361 canary (25 JS tests pass).

Full Python suite: **5555 passed**, 9 skipped, 2 failed — both pre-existing test-pollution failures (`TestC003DaphneOrdering::test_c003_daphne_missing_info`, `TestSuppressChecks::test_no_suppress_by_default`) confirmed identical on clean `main` (5537 passed, same 2 fail), unrelated to this PR. `make build-js` ran (comment-only src change → minified bundle byte-identical, only readable `client.js` + source map changed).

Two-commit shape: impl+tests (no CHANGELOG) / docs+CHANGELOG.


johnrtipton:
## Stage 11 Adversarial Review — PR #1736 (feat #1733: auto-derive + auto-emit dj-navigate route map)

**Verdict: 🟢 APPROVE** (2 🟡 advisory, 0 🔴). Merge-ready; the two 🟡 are documentation-accuracy / latent-foot-gun, not correctness blockers.

### Pre-review gates
- **Stale-base (#250):** `BEHIND=0` vs `origin/main`. `git log origin/main..HEAD` = exactly the 2 PR commits. ✅
- **Two-commit shape:** commit 1 `e10b4186` = code + tests, **no CHANGELOG** (the `routing.py` docstring fix is CODE, correctly in commit 1). commit 2 `d3418500` = only `CHANGELOG.md` + `docs/website/guides/navigation.md`. ✅

### Empirical results
- `pytest python/djust/tests/ -k "rout or client_config or T016 or navigat"` → **107 passed**.
- `pytest python/tests/test_checks.py python/tests/test_navigation.py` → **233 passed**.
- `npx vitest run tests/js/navigation.test.js` → **25 passed** (incl. the new #1361 prototype-pollution canary).
- Full `pytest python/` → **2 failed, 5555 passed** (failures pre-existing — see below; branch adds +18 passing tests over main).

### 1. Dual-engine parity incl. nonce (highest priority) — HOLDS ✅
Both `live_tags.djust_client_config` (Django simple_tag, `takes_context=True`) and `template_tags/client_config.py:ClientConfigTagHandler.render` now delegate to the shared `_client_config_html(request)`, with `request = context.get("request")` on both paths. I empirically rendered both with a `request.csp_nonce="NONCE_XYZ"` against a LiveView URLconf:
- WITH nonce in context: Django nonce ✅, Rust nonce ✅, **byte-equal ✅**
- request present, no nonce: byte-equal ✅
- empty context (as the existing parity test calls it): byte-equal ✅

The Rust engine **does** thread `request` into the handler context — confirmed at `crates/djust_templates/src/registry.rs:443-454` (`call_handler_with_py_sidecar` injects `raw_py_objects` incl. `request`/`view`). So on a real full-page GET the Rust path receives the same nonce the Django path uses. **Nonce parity holds.**

🟡 **Parity test coverage gap** (`python/djust/tests/test_client_config_tag.py:232-251`): all three `_PARITY_CASES` use URLconfs with **no LiveView routes** (`api_test_urls_default`/`_custom`), so the byte-parity test never exercises the **route-map `<script>`** portion — only the api/sse meta. And it calls `rust_handler.render([], {})` with an empty context (no request), so it can't catch a nonce divergence even in principle. The parity I confirmed manually is real, but the regression net does **not** cover the new route-map+nonce surface. Recommend adding a parity case with a LiveView URLconf + a request carrying `csp_nonce` (generalizes Action #1543: a coverage suite must enumerate every variant of the surface it covers — here the route-map-bearing variant of the dual-engine output).

### 2. `takes_context=True` backward-compat — SAFE ✅
Grep for callers: only template usage (`{% djust_client_config %}`, no args) across demo/docs. simple_tag → `takes_context=True` is backward-compatible for template usage. No direct Python `djust_client_config()` callers exist (the Rust handler now calls `_client_config_html`, not the tag fn).

### 3. Cache correctness (#1447) — CORRECT, one latent foot-gun
Key is `(urlconf, script_prefix)`; `_reset_route_map_cache()` is wired into autouse fixtures in `test_client_config_tag.py`, `test_navigation.py`, and `TestT016` setup/teardown. **No third test-pollution failure added** (full-suite failure count is identical on main).

🟡 **Cache key uses the raw `urlconf` arg (usually `None`), not the resolved `ROOT_URLCONF`** (`routing.py:312-318`). `(None, prefix)` does not encode which ROOT_URLCONF was active, so correctness under `override_settings(ROOT_URLCONF=...)` depends entirely on every caller remembering to reset. Production URLconf is static so this is harmless in prod, but it's a latent foot-gun for future tests that forget the fixture. Consider keying on `urlconf or settings.ROOT_URLCONF`. Advisory only.

### 4. Derivation completeness — CORRECT ✅
Empirically dumped the derived map against `tests/route_map_test_urls.py`:
```
/dashboard/      → DashboardView         (bare LiveView)
/items/:id/      → ItemDetailView        (<int:id> → :id ✅)
/secret/         → ProtectedView         (login_required-wrapped ✅)
/section/deep/   → NestedView            (nested include() prefix accumulated ✅)
(plain_django_view excluded ✅)
```
With `set_script_prefix("/mysite/")` all keys correctly gain the `/mysite` prefix. I also verified the under-inclusion concern for **other** decorators: modern Django preserves `view_class` on the wrapper through `login_required`, `user_passes_test`, **and `cache_page`** — all resolve via the direct `hasattr(cb, "view_class")` branch; the `__wrapped__.view_class` fallback is belt-and-suspenders. No over- (issubclass guard) or under-inclusion found.

### 5. Security (#1078) — SAFE, but one inaccurate comment
Route JSON is `json.dumps`-emitted via `format_html` with a per-request nonce; the map is cached but the **nonce is not** (regenerated per render via `get_csp_nonce(request)` — no stale-nonce baking). Route keys/values are developer-defined (URL paths + `module.QualName`), not user input.

🟡 (minor) The comment at `routing.py:354` / `:358` claims `json.dumps escapes <, >, quotes`. **`json.dumps` does NOT escape `<`/`>`/`/`** — verified: `json.dumps({'</script>':'x'})` → `{"</script>": "x"}`. The data being developer-defined makes this a non-issue in practice (a developer would have to put a literal `</script>` in a URL pattern), but the comment overstates the protection. Worth a one-line correction. Not a blocker.

### 6. T016 system check (empirical canary #252) — CORRECT ✅
T016 is a genuinely free ID (highest prior = T015). Independently verified the three states via the test suite (all pass): fires when dj-navigate present + empty map; silent when routes exist; silent when no dj-navigate; suppressible via `DJUST_CONFIG['suppress_checks']`. **Demo dogfood: `manage.py check` → T016 clean** (63 issues, none T016). Note the demo has LiveView routes but **no `dj-navigate`** in templates, so the demo exercises the "no dj-navigate" silent branch, not the "routes-exist" branch — the latter is covered by `test_silent_when_routes_exist`. (Minor: the T016 test's `TEMPLATES` setting uses a typo'd backend path `DjangoTemplateBackend` instead of `DjangoTemplates`, but `check_templates` walks `DIRS` directly so it's inert — tests pass.)

### 7. #1361 — own-prop-safe + non-tautological test ✅
`resolveViewPath` was already `Object.entries`-based; the diff is comment-only (hardened guidance, "do not reintroduce `routeMap[pathname]`"). client.js built artifact is in sync with the src. The new vitest test **actually pollutes `Object.prototype['/evil/']`** and asserts `resolveViewPath('/evil/')` returns null — would fail under bracket-indexing, so non-tautological. ✅

### 8. Test quality / gate-off (#254) ✅
By code inspection: `get_route_map_script` returns `""` when `all_routes` is empty (routing.py:342). With derivation stubbed to `{}` and no `live_session` entries (route_map_test_urls uses plain `path()`), `test_client_config_emits_route_map_script` (asserts `window.djust._routeMap in html`) would fail. Tests assert specific derived entries (`"/items/:id/": "...ItemDetailView"`) that only appear if derivation runs → non-tautological.

### The 2 pre-existing failures — TRULY PRE-EXISTING ✅
`TestC003DaphneOrdering::test_c003_daphne_missing_info` and `TestSuppressChecks::test_no_suppress_by_default` fail **identically on clean `main`** (full `pytest python/`: 2 failed / 5537 passed) and on the branch (2 failed / 5555 passed). Both pass in isolation and in the `test_checks.py`-only scope on both branches → cross-file test pollution (the suppress failure is `C003 should fire by default → 0 == 1`, i.e. a polluted C003, **unrelated to T016**). The branch adds **+18 passing tests** and **no third failure** — critically, the new T016 check did **not** break suppress-check accounting.

### Summary
Solid, well-tested, honest CHANGELOG (calls out the `get_route_map_script` behavior change explicitly). Dual-engine nonce parity holds (verified empirically + by Rust trace). The 🟡s are: (a) the parity *test* doesn't cover the route-map/nonce variant it should, (b) cache key keys on raw `urlconf` not resolved ROOT_URLCONF, plus a minor inaccurate `json.dumps` security comment. None block merge.

johnrtipton:
## Retrospective — PR #1736 (pipeline-run v1.0.2-1)

**Task**: v1.0.2-1 — #1733 (ADR-021 Stage 1): make dj-navigate work with zero wiring (auto route map).

### Quality: 5/5
Feature shipped clean after one advisory fix-pass; the directional foundation #1734 builds on is now solid and self-documenting.

### What went well
- **The investigation that triggered ADR-021 directly shaped the implementation.** dj-navigate's silent full-reload (rent app) → auto-derive the route map from the URLconf (no live_session) + auto-emit via {% djust_client_config %} (the tag every scaffolded app already has). Zero developer wiring; backward-compatible (live_session still merges idempotently); empty-safe.
- **Dual-engine discovery + parity.** The implementer found djust_client_config is registered for BOTH Django and the Rust engine (ClientConfigTagHandler), corrected the orchestrator's "Django-only" engine-path assumption, and kept both paths byte-identical via a shared helper — incl. the CSP nonce (Rust injects request via the raw-py sidecar). 
- **Doc-claim-verbatim fix (#1046):** removed the phantom {% djust_route_map %} from get_route_map_script's docstring and fixed navigation.md's "Full page navigation" mislabel — the exact doc-drift that sent the downstream app down the wrong path.
- **Empirical canary (#252) for the new T016 check** (dj-navigate-without-routes); demo dogfood clean (verified by the now-blocking demo-checks job from #1713).

### What didn't / fix-pass
- Stage 11 approved with 2🟡 advisory; fixed pre-merge because this is a foundation #1734 leans on: (1) the dual-engine parity TEST had a coverage hole — it never exercised the route-map/nonce variant it exists to protect (v1.0.0rc4 finding #1: a coverage suite must enumerate every variant). Closed with a LiveView-URLconf + csp_nonce parity case asserting byte-identical output, gate-off verified. (2) cache keyed on raw urlconf (usually None) → resolved to ROOT_URLCONF, avoiding a 3rd test-pollution failure on an already-polluted suite. (3) corrected an inaccurate json.dumps-escaping comment.

### Process finding
"Parity holds" and "parity is tested" are different claims. The reviewer empirically confirmed the dual-engine output matched, but the regression NET had a single-variant hole (no-LiveView URLconf + empty context) that would never have caught a future divergence on the route-map/nonce path. Closing the coverage hole — not just confirming current parity — is what protects #1734.

### Verified
- Dual-engine byte-parity incl nonce + _routeMap (new parity case, gate-off non-tautological).
- Route-map derivation: LiveView routes incl login_required-wrapped, <int:id>→:id, FORCE_SCRIPT_NAME, nested includes; non-LiveView excluded; empty when none.
- Full python/djust/tests/ 3081 passed twice (deterministic); only the 2 pre-existing python/tests/ pollution failures, 0 new. demo-checks/T016 clean. All CI green. Two-commit shape intact.

`RETRO_COMPLETE`

PR #1739, opened 2026-06-06
fix(render): normalize initial SSR dj-root to match render_with_diff (#1737)

### LINKED ISSUE #1737: Initial SSR render skips the comment/whitespace normalization that render_with_diff applies → first-hydration morphChildren re-render (flash)
**Version:** 1.0.2rc2

## Summary

The initial HTTP render and the WebSocket render produce structurally different HTML for the same view, so the **first** hydration triggers a full `morphChildren` reconciliation — a visible re-render ("flash") on page load, even with #1724 in place.

## Root cause (traced)

`render_with_diff()` normalizes the template via `_strip_comments_and_whitespace()` **before** the Rust VDOM sees it (`djust/mixins/template.py:154`). The initial render path (`render()` / `render_full_template` → `self._rust_view.render()`, `:213` / `:832`) applies **no equivalent normalization**.

For a view whose root is declared `<div dj-view="…">`:

- **SSR** (`GET`): HTML comments preserved (the first child of the root can be a `<!-- … -->` comment node), inter-element whitespace as-authored.
- **First WS frame**: comments stripped, whitespace normalized.

The client can't align the two → wholesale subtree reconciliation on connect.

## Repro

A `LiveView` using template inheritance (`{% extends %}`) whose root template contains HTML comments and normal indentation. On load, watch the `dj-view` root via a MutationObserver: `morphChildren` removes + re-adds the subtree (e.g. a `main-wrapper`/card grid), and any entry animation or layout-dependent element (a status badge) visibly re-renders.

Confirmed by fetching the authenticated `GET` HTML: the root's first child is `<!-- … -->` and there are dozens of comment nodes; the first WS hydration frame has none of them.

## Evidence it's two stacked mismatches

Converting all in-root `<!-- … -->` to Django `{# … #}` comments (so the SSR carries no comment nodes) reduced the flash from **two passes to one**. The remaining single pass is the **whitespace** gap.

## Relationship to #1724

#1724 made `morphChildren` skip whitespace-only *text nodes*, but (a) it doesn't cover comment nodes, and (b) it doesn't fully converge for real-world templates — a residual flash remains from inter-element whitespace. Normalizing at the **source** (the initial render) would make this robust regardless of client-side alignment heuristics.

## Proposed fix

Apply the same `_strip_comments_and_whitespace()` normalization to the initial render's dj-root content (e.g. in `render()` / `render_full_template`) so the SSR output is byte-equivalent to the first `render_with_diff()` output. Then `morphChildren` on first hydration is a no-op and there is no flash.

johnrtipton:
## Summary

Fixes the first-hydration flash (#1737, completes #1724): the initial HTTP-GET render produced structurally different dj-root HTML than the first WebSocket frame, so the client's first `morphChildren` rebuilt the whole subtree (visible re-render).

## Root cause (traced + reproducer-verified)

`render_with_diff()` normalizes the dj-root template via `_strip_comments_and_whitespace()` (in `get_template()`), and the Rust `render_with_diff()` applies a further inter-element whitespace pass on the rendered output. The initial-GET path — `render_full_template()` → `self._rust_view.render()` — did **neither** for the common `dj-view`-only root:

- `render_full_template`'s dj-root-replacement step keyed on `_DJ_ROOT_RE` (literal `dj-root` attribute only). A template declaring just `dj-view` (auto-inferred dj-root, PR #297) **fell through to returning the un-normalized `_full_template` shell** — comment nodes + as-authored whitespace preserved. The first WS frame had them stripped → structural mismatch → flash.

Verified with a standalone reproducer: pre-fix the SSR dj-root kept `<!-- ... -->` comment nodes and indentation; the WS frame had none.

## The fix (three parts, each gate-off verified)

1. **`_DJ_VIEW_RE` fallback** in `render_full_template` step 3 — when no literal `dj-root` attr is present, match the `dj-view` root so the normalized `liveview_html` (rendered from the **same** `self._rust_view` the WS path uses) actually replaces the shell root.
2. **Normalize the rendered dj-root** — apply `_strip_comments_and_whitespace()` to `liveview_html`, mirroring the extra whitespace pass Rust `render_with_diff()` does (plain Rust `render()` does not). E.g. `> <!--dj-if--> <` collapses to `><!--dj-if-->`.
3. **Align `_strip_comments_and_whitespace`** to collapse whitespace adjacent to `<pre>`/`<code>`/`<textarea>` boundaries — Rust treats those tags as tag boundaries; the Python normalizer previously left a space, re-opening the gap. Content **inside** the preserved block is untouched.

## Which slice is normalized

Only the **dj-root subtree** the VDOM owns (the same slice `render_with_diff` normalizes) — not the page shell (`<head>`, scripts). The normalized `liveview_html` replaces only the shell's `<div dj-view/dj-root>...</div>`.

## Byte-equivalence proof

SSR dj-root == first WS frame's dj-root, **modulo the `dj-id` attrs** the Rust `render_with_diff()`/client stamp onto the prerender DOM (#1610 — the one intentional, client-reconciled difference). Asserted in `python/djust/tests/test_ssr_render_normalization_1737.py` for: comments+indentation, `{% if %}`/dj-if blocks, and `<textarea>`.

## Preservation confirmed

- `<pre>`/`<code>`/`<textarea>` **internal** whitespace preserved (placeholder substitution; only boundary whitespace stripped).
- `dj-if` boundary markers preserved (the normalizer's existing negative-lookahead; survives on both paths).

## Engine-path (#1051)

Both `render()` and `render_with_diff()` go through the Rust engine (`self._rust_view`). This change is Python-side pre/post-normalization of the extracted dj-root HTML on the initial-GET path; it does not touch the Rust renderer.

## Gate-off evidence (#254)

Each of the 3 fix components reverted independently → the new byte-equivalence tests fail; restored → green:
- dj-view fallback off → 3 parity tests fail (SSR keeps comments+ws)
- render-output normalization off → dj-if+pre test fails (marker whitespace gap)
- placeholder-adjacency regexes off → 4 tests fail (pre/textarea boundary ws)

## Tests

- New: `python/djust/tests/test_ssr_render_normalization_1737.py` (7 cases — 3 SSR↔WS parity + 4 normalizer unit pins).
- Updated `python/djust/tests/test_strip_whitespace.py` — 2 cases that pinned the OLD Rust-divergent behavior (`<div> <textarea>`) now assert Rust parity (`<div><textarea>`).
- New JS symptom test in `tests/js/ssr_hydration_morph_1724.test.js`: morphChildren over (normalized-SSR, WS-frame) = zero element churn (canvas instance survives).
- Full suites green: `python/djust/tests` 3088 passed / 3 skipped; `python/tests` 2379 passed; `tests/unit` 1831 passed; `npm test` (morph/hydration) passed.

No existing test asserted SSR-contains-comment — all comment asserts are `"<!--" not in result`, consistent with this fix.

Closes #1737. Completes #1724.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- **Initial SSR render now matches the first WebSocket frame — eliminates the first-hydration flash (#1737, completes #1724).** The initial HTTP-GET render skipped the comment/whitespace normalization that `render_with_diff()` applies, so the server-rendered dj-root kept HTML comment nodes and as-authored inter-element whitespace while the first WS frame had them stripped. The structural mismatch made the client's first-hydration `morphChildren` rebuild the whole subtree (visible re-render / flash), even with #1724's client-side whitespace-only-text-node skip in place. `render_full_template()` now (a) falls back to matching the `dj-view` root when no literal `dj-root` attribute is present (the auto-inferred-dj-root case) so the normalized render replaces the shell root, and (b) applies `_strip_comments_and_whitespace()` to the rendered dj-root, mirroring the extra whitespace pass the Rust `render_with_diff()` performs. `_strip_comments_and_whitespace()` now also collapses whitespace adjacent to `<pre>`/`<code>`/`<textarea>` boundaries for byte-parity with the Rust pass. The SSR dj-root is now byte-equivalent to the first WS frame (modulo the `dj-id` attrs the client stamps onto the prerender DOM per #1610), so the first `morphChildren` is a no-op. `<pre>`/`<code>`/`<textarea>` internal whitespace and `dj-if` boundary markers are preserved.
+
 ## [1.0.2rc2] - 2026-06-05
 
 ### Added
--- a/python/djust/mixins/template.py
+++ b/python/djust/mixins/template.py
@@ -32,6 +32,21 @@ _DJ_ROOT_RE = re.compile(
     re.IGNORECASE,
 )
 
+# Match ``<div ... dj-view ...>`` as a standalone attribute name. Used as a
+# FALLBACK to ``_DJ_ROOT_RE`` in ``render_full_template``: when a template
+# declares only ``dj-view`` (the auto-inferred-dj-root case, see PR #297) and
+# no literal ``dj-root`` attribute, the dj-root replacement step must still
+# find the root div in the page shell — otherwise it falls through to
+# returning the un-normalized ``_full_template`` render, leaving HTML comments
+# and as-authored whitespace in the initial-GET dj-root that the WS
+# (``render_with_diff``) frame has already stripped. That structural mismatch
+# is what triggers the first-hydration ``morphChildren`` re-render / flash
+# (#1737). Same standalone-attribute lookahead semantics as ``_DJ_ROOT_RE``.
+_DJ_VIEW_RE = re.compile(
+    r"<div\b[^>]*?(?<![A-Za-z0-9_-])dj-view(?=[\s=>/])[^>]*>",
+    re.IGNORECASE,
+)
+
 # Match ``</body>`` tolerating trailing whitespace inside the tag (``</body >``).
 _BODY_CLOSE_RE = re.compile(r"</body\s*>", re.IGNORECASE)
 
@@ -317,6 +332,36 @@ Object.assign(window.handlerMetadata, {json.dumps(metadata)});
         # Normalize whitespace
         html = re.sub(r"\s+", " ", html)
         html = re.sub(r">\s+<", "><", html)
+
+        # #1737: collapse whitespace between a tag boundary and a preserved
+        # (<pre>/<code>/<textarea>) block too, so this Python normalizer
+        # matches the Rust ``render_with_diff()`` whitespace pass exactly.
+        # Rust's parser drops every whitespace-only text node that is a direct
+        # child of a non-whitespace-preserving element (parser.rs:520-531), so
+        # the inter-element whitespace around — and BETWEEN — preserved blocks
+        # is removed: ``</div> <pre>`` → ``</div><pre>``,
+        # ``</textarea> </div>`` → ``</textarea></div>``, AND
+        # ``</textarea> <pre>`` → ``</textarea><pre>`` (preserved↔preserved).
+        # The placeholder-substitution above hides those boundaries from the
+        # ``>\s+<`` rule (the placeholder doesn't start with ``<``), so collapse
+        # them explicitly. Without this the initial-GET dj-root keeps
+        # whitespace-only text nodes around preserved blocks that the first WS
+        # frame lacks, re-opening the first-hydration whitespace mismatch
+        # (#1724 / #1737). Whitespace INSIDE a preserved block is untouched
+        # (it's hidden behind the placeholder and restored verbatim below), and
+        # whitespace adjacent to actual TEXT (e.g. ``before <pre>``) is left as
+        # a single space — Rust keeps it because that text node is not
+        # whitespace-only.
+        #
+        # (1) literal-tag → preserved   and   (2) preserved → literal-tag:
+        html = re.sub(r">\s+(__PRESERVED_BLOCK_\d+__)", r">\1", html)
+        html = re.sub(r"(__PRESERVED_BLOCK_\d+__)\s+<", r"\1<", html)
+        # (3) preserved → preserved: collapse whitespace between two adjacent
+        # preserved blocks. The lookahead (not a consuming group) lets a run of
+        # 3+ adjacent blocks collapse every gap in a single pass — a consuming
+        # ``\1...\2`` form would swallow the middle block and miss its trailing
+        # gap.
+        html = re.sub(r"(__PRESERVED_BLOCK_\d+__)\s+(?=__PRESERVED_BLOCK_\d+__)", r"\1", html)
 
         # Restore preserved blocks
         for i, block in enumerate(preserved_blocks):
@@ -832,6 +877,21 @@ Object.assign(window.handlerMetadata, {json.dumps(metadata)});
             liveview_html = self._rust_view.render()
             liveview_html = self._hydrate_react_components(liveview_html)
 
+            # #1737: normalize the rendered dj-root so the initial-GET output
+            # is structurally identical to the first WS frame. The WS path
+            # (``render_with_diff`` → Rust ``render_with_diff()``) applies an
+            # additional whitespace pass that plain Rust ``render()`` does NOT
+            # — e.g. the single spaces a normalized template keeps around
+            # ``{% if %}`` tags survive ``render()`` as ``> <!--dj-if--> <``
+            # but are collapsed to ``><!--dj-if-->`` by ``render_with_diff()``.
+            # Applying the SAME ``_strip_comments_and_whitespace()`` the
+            # template path uses (get_template():154/172/184) converges the
+            # two: ``dj-if`` boundary markers are preserved (negative
+            # lookahead in the normalizer), ``<pre>``/``<code>``/``<textarea>``
+            # whitespace is preserved, and the only residual difference is the
+            # dj-id attrs the client stamps onto the prerender DOM (#1610).
+            liveview_html = self._strip_comments_and_whitespace(liveview_html)
+
             # --- Step 2: Render the page shell from _full_template ---
             from djust._rust import RustLiveView
 
@@ -881,7 +941,21 @@ Object.assign(window.handlerMetadata, {json.dumps(metadata)});
             # wrapper (since get_template() returns the dj-root template).
             # Replace the shell's <div dj-root>...</div> ENTIRELY (opening
             # tag through closing tag) with liveview_html.
-            dj_root_match = _DJ_ROOT_RE.search(shell_html)
+            #
+            # #1737: fall back to ``dj-view`` when no literal ``dj-root``
+            # attribute is present (the auto-inferred-dj-root case). Without
+            # this fallback the replacement misses the root entirely and we
+            # return the un-normalized ``_full_template`` shell — leaving
+            # comment nodes + as-authored whitespace in the initial-GET
+            # dj-root that the WS frame (``render_with_diff`` →
+            # ``_strip_comments_and_whitespace`` at get_template()) has
+            # already stripped. The structural mismatch is what makes the
+            # client's first-hydration ``morphChildren`` rebuild the subtree
+            # (visible flash). ``liveview_html`` is rendered from the SAME
+            # normalized ``self._rust_view`` the WS path uses, so the
+            # replaced dj-root is structurally identical to the first WS
+            # frame (modulo the dj-id attrs the client stamps on, per #1610).
+            dj_root_match = _DJ_ROOT_RE.search(shell_html) or _DJ_VIEW_RE.search(shell_html)
             if dj_root_match:
                 # Start of the <div dj-root...> opening tag
                 tag_start = dj_root_match.start()
--- a/python/djust/tests/test_ssr_render_normalization_1737.py
+++ b/python/djust/tests/test_ssr_render_normalization_1737.py
@@ -1,0 +1,356 @@
+"""Regression tests for #1737 — initial SSR render must apply the SAME
+comment/whitespace normalization that ``render_with_diff`` applies, so the
+first-hydration ``morphChildren`` is a no-op (no flash).
+
+Root cause (#1737)
+------------------
+``render_with_diff()`` normalizes the dj-root template via
+``_strip_comments_and_whitespace()`` in ``get_template()`` BEFORE the Rust
+VDOM renders it, and the Rust ``render_with_diff()`` applies a further
+inter-element whitespace pass on the rendered output. The initial-GET path
+(``render_full_template`` → ``self._rust_view.render()``) did NEITHER for the
+common ``dj-view``-only root: its dj-root replacement keyed on
+``_DJ_ROOT_RE`` (literal ``dj-root`` attribute only), so a template declaring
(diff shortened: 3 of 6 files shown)


johnrtipton:
## CI note — pre-existing flaky failure (not caused by this PR)

The first `python-tests (py3.12)` run failed on:

```
python/tests/test_csp_nonce_support.py::TestRouteMapScriptNonce::test_no_routes_returns_empty
```

This is **unrelated to this PR** and is a pre-existing test-isolation/pollution flake:

- This PR's diff touches only `CHANGELOG.md`, `python/djust/mixins/template.py`, two new/updated test files, and one JS test — **it does not touch `test_csp_nonce_support.py`** (`git diff origin/main..HEAD --name-only` confirms).
- The same code **passed** on `python-tests (py3.14t free-threaded)` — identical source, different xdist shard ordering.
- The failure is a global-state pollution signature: `test_no_routes_returns_empty` expects an empty `window.djust._routeMap` but sees one populated by the demo URLconf / a prior test in the same `pytest-xdist` worker (`[gw0]`). It passes or fails depending on shard distribution, not on this diff.
- `rust-tests`, `javascript-tests`, `security-tests`, `demo-checks`, `playwright-tests`, `python-tests (py3.14t)`, and both CodeQL analyses all **passed**.

Re-running the failed job to confirm it's a flake. The local full suites pass: `python/djust/tests` 3088 passed, `python/tests` + `tests/unit` all pass; the failing CSP test passes when run in a non-polluting order.

johnrtipton:
## Stage-11 Adversarial Review — PR #1739 (fix #1737)

**Verdict: 🟡 APPROVE-WITH-FINDINGS.** The core fix is correct and effective — the first-hydration flash class it targets (comment nodes + as-authored whitespace + un-normalized `dj-view`-only root) is genuinely killed. All 36 Python + 7 JS tests pass. But the headline "byte-equivalence modulo dj-id" claim is **literally false in one whitespace case** (benign in effect), and the CI-failing test is **not the flake the PR description claims** — it's a deterministic #1733 regression.

Counts: **0 🔴 · 2 🟡 · 4 🟢**

### Pre-review gates ✅
- STALE-BASE (#250): `BEHIND=0`. Only this PR's 2 commits over `origin/main`.
- Two-commit shape ✅: `f435f4ed` impl+tests (no CHANGELOG), `9f83ee4c` CHANGELOG-only.

---

### 🟡 Finding 1 — byte-equivalence-modulo-dj-id does NOT hold for two ADJACENT preserved blocks (`python/djust/mixins/template.py:349-350`)

The Part-3 regexes only handle a preserved block adjacent to a **literal tag boundary**:
```python
html = re.sub(r">\s+(__PRESERVED_BLOCK_\d+__)", r">\1", html)
html = re.sub(r"(__PRESERVED_BLOCK_\d+__)\s+<", r"\1<", html)
```
They do **not** handle whitespace between **two** preserved-block placeholders (`__PRESERVED__ \s+ __PRESERVED__`) — neither side is `>`/`<`. So when a `<pre>`/`<code>`/`<textarea>` is an immediate sibling of another, the SSR path leaves a single space that Rust drops.

I built an independent composite two-path render (extends + comments + indentation + `{% if %}/else` + `<textarea>` + `<pre>`, all in one root). Result, dj-id stripped:
```
SSR: ...</textarea> <pre>  x\n    y\n</pre>...   ← stray space
WS : ...</textarea><pre>  x\n    y\n</pre>...    ← collapsed
BYTE-EQUAL (modulo dj-id): False  (first diff at idx 172)
```
Minimal repro against the normalizer directly:
```
IN : <div><textarea>a</textarea>   <pre>b</pre></div>
OUT: <div><textarea>a</textarea> <pre>b</pre></div>   (space remains; Rust → no space)
```
**Effect: benign, NOT 🔴.** I traced the client morph (`12-vdom-patch.js:777-786`, #1724): when the desired node is a significant element, `morphChildren` skips the insignificant whitespace-only existing text node and aligns on the existing `<pre>`/`<textarea>` element (Strategy 2, reuse-in-place); the orphan whitespace node is pruned by the cleanup loop (`:853-862`). No element teardown, no flash — just a single whitespace-text-node removal on first hydration. So the *symptom* the PR targets is still fixed for this shape, but the stated invariant ("only residual difference is dj-id") is overstated, and neither the Python suite nor the JS test (`ssr_hydration_morph_1724.test.js` uses a hand-written SSR string that omits this shape) covers it.

**Recommendation (fast-follow OK):** add a third Part-3 pass for the placeholder↔placeholder boundary:
```python
html = re.sub(r"(__PRESERVED_BLOCK_\d+__)\s+(?=__PRESERVED_BLOCK_\d+__)", r"\1", html)
```
and add a regression case with two adjacent preserved blocks asserting full byte-equivalence-modulo-dj-id.

### 🟡 Finding 2 — the CI-failing test is a deterministic #1733 regression, NOT the "xdist-ordering pollution flake" the PR claims (`python/tests/test_csp_nonce_support.py:109`)

`test_no_routes_returns_empty` clears only `live_session._route_maps = {}`, then asserts `get_route_map_script(...) == ""`. But #1733 (commit `0784a667`, on main — not this PR) changed `get_route_map_script` (`routing.py:345`) to also merge `build_route_map_from_urlconf()`. That helper walks `ROOT_URLCONF` — and the demo settings urlconf has **85 LiveView routes**. So the function returns a populated `<script>` tag and `out == ""` fails.

This is **not flaky and not ordering-dependent**: I ran `TestRouteMapScriptNonce` **alone** (no xdist, no random) on this branch → FAILS. I ran the single test alone on **`origin/main`** → FAILS. It deterministically fails against any settings whose urlconf has LiveView routes; it only "passes" when collected under a settings module with an empty/LiveView-less urlconf. The `_route_map_cache` (#1733 module-level cache) is a secondary pollution hazard, but the *primary* cause is the urlconf-merge the test was never updated for.

The implementer is correct that it is **pre-existing and not caused by #1739** — so it does **not block this PR**. But the diagnosis "global-route-map / xdist pollution flake" is wrong and should not stand: it's a real broken test that masks a coverage gap in the #1733 path. **Recommendation (separate issue / fast-follow):** isolate the test from the urlconf source — `override_settings(ROOT_URLCONF=<empty>)` (or monkeypatch `build_route_map_from_urlconf` to `{}`) **plus** call `_reset_route_map_cache()` in `setup_method`/`teardown_method`. The existing `setup_method` already stashes `live_session._route_maps` but ignores both the urlconf merge and the module cache.

---

### 🟢 Verified correct

- **🟢 pre/code/textarea INTERNAL content preserved (the corruption risk — item 2).** Probed `<pre>` with leading+internal indentation, multi-line `<textarea>` default value, spaced `<code>`. INTERNAL whitespace is **byte-preserved** in every case; only boundary (outside-the-tag) whitespace collapses. The placeholder-substitution at `:324-330` runs BEFORE the `\s+` collapse, so block internals are genuinely shielded. No corruption.

- **🟢 The 2 changed `test_strip_whitespace.py` tests reflect REAL Rust behavior, not test-fudging (item 3).** The new assertions (`<div><textarea>`, `</span><textarea>`) match `crates/djust_vdom/src/parser.rs:518-534`, where the VDOM parser **drops** whitespace-only text nodes entirely for non-preserve parents (`!text.chars().all(is_whitespace)`). The pre-#1737 single-space output (`<div> <textarea>`) genuinely diverged from Rust. Legitimate.

- **🟢 `_DJ_VIEW_RE` fallback doesn't break `dj-root` roots (item 4).** Probed all shapes: `dj-root`-only → `_DJ_ROOT_RE` path (unchanged); `dj-view`-only → fallback; **both attrs → `_DJ_ROOT_RE` wins** via `or` short-circuit (no double-match); `data-dj-root` / `dj-view-mode` traps → neither matches (lookbehind `(?<![A-Za-z0-9_-])` + lookahead `(?=[\s=>/])` reject substrings/prefixes). Sound.

- **🟢 dj-if markers preserved on both paths (item 5).** Negative-lookahead `<!--(?!\s*/?dj-if\b).*?-->` (`:315`) keeps `<!--dj-if ...-->` / `<!--/dj-if-->` while stripping plain comments — confirmed empirically and in the composite two-path render (`dj-if id="if-..."` boundary markers identical on SSR and WS sides). Matches `parser.rs:493-514`.

---

### Test results
- `pytest test_ssr_render_normalization_1737.py test_strip_whitespace.py` → **36 passed**
- `vitest run ssr_hydration_morph_1724.test.js` → **7 passed**
- `TestRouteMapScriptNonce` alone (this branch & main) → 4 passed, **1 failed** (Finding 2; pre-existing)

**Bottom line:** mergeable. The fix delivers its flash-kill goal. Recommend folding Finding 1's one-line regex + a regression case (or filing a fast-follow), and filing/correcting the Finding 2 diagnosis as a separate test-isolation issue against the #1733 route-map path.

johnrtipton:
## Stage-11 findings addressed (force-pushed, two-commit shape preserved)

Both 🟡 findings fixed in commit 1 (impl+tests); the CHANGELOG commit (commit 2) is untouched.

### 🟡 1 — Closed the preserved↔preserved byte-equivalence gap (+ pinned the invariant)
- `_strip_comments_and_whitespace` now collapses whitespace BETWEEN two adjacent preserved blocks, e.g. `</textarea> <pre>` → `</textarea><pre>`, via a **lookahead** so a run of 3+ adjacent preserved blocks (`<textarea> <pre> <code>`) collapses every gap in one pass (a consuming `\1...\2` form would swallow the middle block and miss its trailing gap).
- **Confirmed Rust collapses this same boundary**: `crates/djust_vdom/src/parser.rs:520-531` filters out every whitespace-only text node that is a direct child of a non-whitespace-preserving element — so `</textarea><pre>` and `</pre><code>` carry no text node between them. I verified `render_with_diff()` output for adjacent `<textarea>/<pre>/<code>` matches the normalizer exactly (modulo dj-id). No over-collapse: actual TEXT between preserved blocks (e.g. `</pre> mid <textarea>`) keeps its single space, matching Rust (the text node is not whitespace-only).
- New tests: `test_adjacent_preserved_blocks_byte_equivalent` (renders 3 adjacent preserved blocks through BOTH paths, strips dj-id, asserts byte-equality + internal-whitespace preserved) plus 3 unit pins (`test_whitespace_between_two_preserved_blocks_collapsed`, `test_three_adjacent_preserved_blocks_all_gaps_collapsed`, `test_text_between_two_preserved_blocks_keeps_single_space`). Gate-off verified: reverting the new regex fails all 3 of the boundary-collapse tests.

### 🟡 2 — Fixed the #1733 regression `test_no_routes_returns_empty` (blocking py3.12 CI)
- Root cause confirmed: #1733 made `get_route_map_script` merge `build_route_map_from_urlconf()` (walks the active ROOT_URLCONF → demo has LiveView routes) with `live_session._route_maps`. The test only cleared `live_session._route_maps`, not the urlconf source or the module-level `_route_map_cache`, so it failed deterministically when the urlconf-derived map was cached.
- Fix (test-correctness): the test now uses `@override_settings(ROOT_URLCONF="tests.api_test_urls_unmounted")` (an existing routeless urlconf, `urlpatterns = []`) AND `_reset_route_map_cache()` in setup/teardown so no derived map leaks in. Asserts the genuine empty-map case returns `""`.
- `TestRouteMapScriptNonce` passes alone (5/5). Exact CI invocation `pytest tests/ python/tests/ -n auto` is green across 3 runs (4530 passed, 0 failed). `python/djust/tests/` standalone: 3092 passed.

Noted both in the commit-1 message. Re-running CI on `79c9049d`.

johnrtipton:
## Retrospective — PR #1739 (pipeline-drain v1.0.2-2)

**Task**: v1.0.2-2 — #1737 (completes #1724): normalize the initial SSR dj-root to match render_with_diff, killing the first-hydration flash at the source.

### Quality: 5/5
Shipped clean after one fix-pass; the reviewer overturned two implementer claims with independent evidence, and the fixes made the invariant true + tested.

### What went well
- **Source-side fix completes the client-side #1724 fix.** #1724 made morphChildren skip whitespace text nodes (client heuristic); #1737 makes the SSR dj-root byte-equivalent to the first render_with_diff frame (modulo dj-id, the #1610 intentional diff), so first-hydration morph is a no-op — robust regardless of client heuristics. Defense in depth.
- **Sharper root cause than reported.** The reporter cited render()/render_full_template lacking normalization; the implementer traced it to a `dj-view`-only root falling through `_DJ_ROOT_RE` and returning the un-normalized shell — fixed with a `_DJ_VIEW_RE` fallback + normalizing the rendered liveview_html + a pre/textarea boundary-whitespace alignment.
- **Reviewer caught an overstated invariant and a misdiagnosed "flake."** (1) "byte-equivalence modulo dj-id" was false for adjacent preserved blocks (`</textarea> <pre>`) — benign (client prunes the node) but untested; fixed with a lookahead regex (handles 3+ adjacent blocks) + a both-paths byte-equality test, Rust parity confirmed at parser.rs:520-531. (2) The "xdist flake" was a deterministic #1733 regression: get_route_map_script now merges the URLconf-derived map, so test_no_routes_returns_empty never saw an empty map; fixed with override_settings(routeless urlconf) + _reset_route_map_cache().

### What didn't / fix-pass
- Both 🟡 folded into commit 1. The #1733 test regression was folded here (not a separate PR) because it was blocking #1739's own py3.12 CI and is a 2-line test-correctness fix.

### Process finding
"Byte-equivalent" is a claim that must be TESTED across every variant of the surface, not asserted (v1.0.0rc4 finding #1). The implementer's invariant held for preserved↔literal boundaries but not preserved↔preserved — exactly the variant the test didn't cover. The fix wasn't just the regex; it was adding the both-paths byte-equality test so the invariant can't silently regress.

### Verified
- pre/code/textarea INTERNAL whitespace byte-preserved (placeholder-extraction before collapse); only boundary whitespace collapses, matching Rust. dj-if markers preserved both paths.
- The 2 changed test_strip_whitespace.py cases match real Rust behavior (parser.rs:518-534), not test-fudging.
- TestRouteMapScriptNonce 5/5 alone; CI invocation (pytest tests/ python/tests/ -n auto) green 3×; python/djust/tests/ 3092 passed. All CI green (py3.12, demo-checks, rust-tests). Two-commit shape intact.

`RETRO_COMPLETE`

Issue #1741, opened 2026-06-06
tech-debt: fix 2 pre-existing test-pollution failures in test_checks.py + audit module-level caches for test-reset


johnrtipton:
Source: Retro v1.0.2 navigation arc (PRs #1736, #1739 Stage-11 reviews).

Two tests in `python/tests/test_checks.py` fail deterministically only under cross-test ordering (pass in isolation), surfacing intermittently in CI shards:
- `TestC003DaphneOrdering::test_c003_daphne_missing_info` (test_checks.py:286)
- `TestSuppressChecks::test_no_suppress_by_default` (test_checks.py:4279) — the suppress failure is a polluted C003 (`0 == 1`).

Both were flagged as **pre-existing** by the #1736 and #1739 reviewers (not introduced by those PRs). They're a global-state / check-registry pollution between tests.

**Also (related class):** #1733 added a module-level `_route_map_cache` (routing.py:27) with a `_reset_route_map_cache()` hook. PR #1739 had to fix `test_no_routes_returns_empty` because the cache + URLconf-derived merge leaked across tests. 

**Action:**
1. Fix the 2 pollution failures (find the polluting test that mutates global check state without teardown; add an autouse reset fixture or isolate the mutation).
2. Audit module-level caches (`_route_map_cache`, the check registry, any `@lru_cache` on request/URLconf-derived data) for a corresponding autouse reset fixture in the test suite, so this class stops recurring.
3. Verify: `pytest python/tests/test_checks.py -p no:randomly` and a few shard orderings are clean; run the full suite twice.

PR #1747, opened 2026-06-07
fix(render): match real dj-root/dj-view attribute, not substring, when picking VDOM source (#1746)

### LINKED ISSUE #1746: render_full_template double-renders the whole page when a LiveView's child template contains the literal text "dj-root"/"dj-view" (e.g. in a code example)
## Summary

A `LiveView` that uses template inheritance double-renders its **entire page** (nested `<!DOCTYPE>…</html>`) on the initial GET when the **child** template contains the literal substring `dj-root` or `dj-view` anywhere — including as plain text inside a `<code>`/`<pre>` example, or even as a substring of another word (`adj-view`). Symptoms: two `<footer>`s, a truncated inline `<script>` (browser console: `Uncaught SyntaxError: Unexpected end of input`), and broken `dj-navigate` SPA morphs that leak one page's content into the next.

Reproduces on **1.0.1 and 1.0.2** (not a 1.0.2 regression).

## Root cause

`python/djust/mixins/template.py:161-165` (`get_template`) chooses the VDOM source with a **naive substring** check:

```python
vdom_source = (
    template_source                                   # the CHILD template
    if ("dj-root" in template_source or "dj-view" in template_source)
    else resolved                                     # the full resolved (base+child) template
)
vdom_template = self._extract_liveview_root_with_wrapper(vdom_source)
```

The intent: "if the child declares the dj-root block, extract from the child; otherwise the dj-root is only in the base, so extract from `resolved`." But `"dj-root" in template_source` matches the token **anywhere** — including documentation/example code that merely *displays* `dj-root`/`dj-view`. When the real `<div dj-root>` lives in the **base** template (a very common layout) and the child only *mentions* the tokens in text, this picks the child as `vdom_source`; `_extract_liveview_root_with_wrapper` finds no real dj-root there → the VDOM template is wrong → `render_full_template`'s shell/dj-root replacement nests the full page.

## Minimal reproduction

```python
# base has the real <div dj-root>; child only MENTIONS the tokens as text
BASE = ('<!DOCTYPE html><html><head><title>T</title></head><body><nav>NAV</nav>'
        '<main><div dj-root>{% block content %}{% endblock %}</div></main><footer>F</footer></body></html>')

# child A — no dj-root/dj-view text:   render_full_template -> ONE <!DOCTYPE>  (correct)
#   {% extends "base" %}{% block content %}<p>hello</p>{% endblock %}
# child B — mentions the tokens:        render_full_template -> TWO <!DOCTYPE> (BUG)
#   {% extends "base" %}{% block content %}<p>Use <code>dj-view</code> and <code>dj-root</code>.</p>{% endblock %}
```

Verified with a `render_full_template(None)` harness (mirrors `test_ssr_render_normalization_1737.py`):

| child content | `<!DOCTYPE>` count |
|---|---|
| `<p>hello</p>` | 1 ✅ |
| `<p>dj-view</p>` (plain text) | 2 ❌ |
| `<p>dj-root</p>` (plain text) | 2 ❌ |
| `<p>dj-click</p>` (other dj-* attr) | 1 ✅ |
| `<p><code>dj-view</code></p>` | 2 ❌ |
| `<p>adj-view</p>` (substring) | 2 ❌ |

Only `dj-root`/`dj-view` trigger it (other `dj-*` attributes don't), and the match is a bare substring (so `adj-view` triggers too).

## Suggested fix

Replace the substring check with the same anchored-attribute regexes the rest of `template.py` already uses for real dj-root/dj-view detection (`_DJ_ROOT_RE` / `_DJ_VIEW_RE`, which require `<div … dj-root …>`):

```python
vdom_source = (
    template_source
    if (_DJ_ROOT_RE.search(template_source) or _DJ_VIEW_RE.search(template_source))
    else resolved
)
```

This distinguishes "the child declares a real `<div dj-root>`" from "the child merely mentions the token in text." Add the repro table above as regression tests (a child that *displays* `dj-root`/`dj-view` in a `<code>` block must render exactly one `<!DOCTYPE>`/`<footer>`).

## Impact

Hits any djust app whose pages document djust itself (docs sites, tutorials, marketing pages with code examples) where the dj-root lives in the base template. Found via djust.org (homepage cheat-sheet section + the pre-existing quickstart), worked around there by splitting the literal tokens in displayed code; the proper fix is this one-line predicate change.

johnrtipton:
## Root cause

`python/djust/mixins/template.py:161-165` (`get_template()`) chose the VDOM source with a **naive substring** check:

```python
vdom_source = (
    template_source                                   # the CHILD template
    if ("dj-root" in template_source or "dj-view" in template_source)
    else resolved                                     # the full resolved (base+child) template
)
```

`"dj-root" in template_source` matches the token **anywhere** — including documentation/example code that merely *displays* `dj-root`/`dj-view`, or another word containing it as a substring (`adj-view`). When the real `<div dj-root>` lives in the **base** template (a common layout) and the child only mentions the tokens in text, this wrongly picks the child as `vdom_source`; `_extract_liveview_root_with_wrapper` finds no real dj-root there → wrong VDOM template → `render_full_template`'s shell/dj-root replacement nests the whole page (two `<!DOCTYPE>`/two `<footer>`, truncated `<script>`, broken `dj-navigate`).

## Fix

Replace the substring check with the anchored-attribute regexes already defined in the module (`_DJ_ROOT_RE` / `_DJ_VIEW_RE`), which require a REAL `<div ... dj-root/dj-view ...>` tag:

```python
vdom_source = (
    template_source
    if (_DJ_ROOT_RE.search(template_source) or _DJ_VIEW_RE.search(template_source))
    else resolved
)
```

Both regexes are module-level and already in scope at line 161. A child that merely displays the token in text no longer matches; the legitimate "child declares its own real `<div dj-root>`" case (the reason the heuristic exists) still picks `template_source`.

## Variant repro table (regression test `test_render_full_template_djroot_substring_1746.py`)

| child content | `<!DOCTYPE>` before | after |
|---|---|---|
| `<p>hello</p>` (control) | 1 | 1 |
| `<p>dj-view</p>` (plain text) | **2** | 1 |
| `<p>dj-root</p>` (plain text) | **2** | 1 |
| `<p><code>dj-view</code> and <code>dj-root</code></p>` | **2** | 1 |
| `<p>adj-view</p>` (substring) | **2** | 1 |
| `<p>dj-click</p>` (other dj-* attr, control) | 1 | 1 |

Plus a control proving the **good path** still works: a child that DOES declare its own real `<div dj-root>` still renders exactly one `<!DOCTYPE>`/`<footer>`.

## Gate-off (#254)

Reverting the predicate to the naive substring (`if ("dj-root" in ... or "dj-view" in ...)`) makes the 4 behavior-meaningful cases (`dj-view`/`dj-root`/code/`adj-view`) fail again (2 DOCTYPEs); restoring the regex makes them pass. The 3 control cases pass either way — confirming the 4 are non-tautological.

## Symbol-migration grep (canon)

`grep -rn '"dj-root" in\|"dj-view" in' python/djust/` found one other production instance: `python/djust/checks.py:3341-3342` (`_check_view_root_same_element`, T005). It is **legitimately fine** — it operates on text already matched to be inside an HTML tag (`tag_re` extracts `<...>` tags first, then checks substring within each tag), so it cannot match displayed text the way the render path did. Not in scope (#1079). The remaining matches are test assertions (`test_mcp.py`).

## Test results

- New regression file: 7/7 pass (RED→GREEN confirmed).
- `test_ssr_render_normalization_1737.py` + `test_template_inheritance.py` + `test_context_processors_in_includes_1722.py` + `test_model_fk_rendering.py`: **30/30 pass** (the #1737 parity tests stay green).
- `python/djust/tests/`: **3099 passed, 3 skipped**.
- `python/tests/` + `tests/unit/`: 4206 passed; 4 pre-existing cross-dir-ordering pollution failures in `tests/unit/test_demo_views.py` (urlconf registration) — verified pre-existing (pass in isolation, and fail identically on a clean `origin/main` tree with my changes stashed). Unrelated to this fix.

Closes #1746

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- **`render_full_template` double-rendered the whole page when a child template merely *mentioned* `dj-root`/`dj-view` as text** (#1746). `get_template()` picked the VDOM source with a naive substring check (`"dj-root" in template_source or "dj-view" in template_source`). When the real `<div dj-root>` lived in the BASE template and the child only displayed the tokens in text (e.g. a `<code>` example) — or as a substring of another word (`adj-view`) — the check wrongly selected the child as the VDOM source, so `render_full_template` nested the entire page (two `<!DOCTYPE>`/two `<footer>`, truncated inline `<script>`, broken `dj-navigate` morphs). Replaced the substring check with the anchored-attribute regexes `_DJ_ROOT_RE`/`_DJ_VIEW_RE` already used elsewhere in the module, which require a real `<div ... dj-root/dj-view ...>` tag.
+
 ## [1.0.2] - 2026-06-06
 
 ### Added
--- a/python/djust/mixins/template.py
+++ b/python/djust/mixins/template.py
@@ -158,9 +158,23 @@ class TemplateMixin:
                     # the dj-root block directly without base template surrounding HTML,
                     # making extraction simpler and immune to Issue #365 miscount.
                     # Fall back to resolved if dj-root is only in the base template.
+                    #
+                    # Use the anchored-attribute regexes (NOT a naive substring): a
+                    # naive ``"dj-root" in template_source`` matches the token ANYWHERE
+                    # — including documentation/example code that merely *displays*
+                    # ``dj-root``/``dj-view`` as text, or another word containing it as
+                    # a substring (``adj-view``). When the real ``<div dj-root>`` lives
+                    # in the BASE template and the child only mentions the tokens in
+                    # text, the substring check wrongly picks the child as the VDOM
+                    # source → extraction finds no real dj-root → render_full_template
+                    # nests the whole page (two <!DOCTYPE>/two <footer>). The regexes
+                    # require a REAL ``<div ... dj-root/dj-view ...>`` tag (#1746).
                     vdom_source = (
                         template_source
-                        if ("dj-root" in template_source or "dj-view" in template_source)
+                        if (
+                            _DJ_ROOT_RE.search(template_source)
+                            or _DJ_VIEW_RE.search(template_source)
+                        )
                         else resolved
                     )
                     vdom_template = self._extract_liveview_root_with_wrapper(vdom_source)
--- a/python/djust/tests/test_render_full_template_djroot_substring_1746.py
+++ b/python/djust/tests/test_render_full_template_djroot_substring_1746.py
@@ -1,0 +1,189 @@
+"""Regression tests for #1746 — render_full_template must NOT double-render the
+whole page when a LiveView's CHILD template merely *mentions* the literal text
+``dj-root`` / ``dj-view`` (e.g. inside a ``<code>`` example) while the real
+``<div dj-root>`` lives in the BASE template.
+
+Root cause (#1746)
+------------------
+``python/djust/mixins/template.py`` ``get_template()`` chose the VDOM source
+with a NAIVE substring check::
+
+    vdom_source = (
+        template_source
+        if ("dj-root" in template_source or "dj-view" in template_source)
+        else resolved
+    )
+
+``"dj-root" in template_source`` matches the token ANYWHERE — including
+documentation/example code that merely *displays* ``dj-root`` / ``dj-view`` as
+text, or even another word containing it as a substring (``adj-view``). When
+the real ``<div dj-root>`` lives in the BASE template (a common layout) and the
+child only mentions the tokens in text, this wrongly picks the child as
+``vdom_source``; ``_extract_liveview_root_with_wrapper`` finds no real dj-root
+there → the VDOM template is wrong → ``render_full_template``'s shell/dj-root
+replacement nests the whole page (two ``<!DOCTYPE>`` / two ``<footer>``).
+
+Fix
+---
+Replace the substring check with the anchored-attribute regexes the rest of the
+module already uses (``_DJ_ROOT_RE`` / ``_DJ_VIEW_RE``), which require a REAL
+``<div ... dj-root ...>`` tag — so a child that merely displays the token in
+text won't match.
+
+This file pins the variant repro table from the issue:
+
+| child content                                  | <!DOCTYPE> count |
+|------------------------------------------------|------------------|
+| ``<p>hello</p>``                               | 1 (control)      |
+| ``<p>dj-view</p>`` (plain text)                | 1 (was 2)        |
+| ``<p>dj-root</p>`` (plain text)                | 1 (was 2)        |
+| ``<p><code>dj-view</code> <code>dj-root</code>`` | 1 (was 2)      |
+| ``<p>adj-view</p>`` (substring)                | 1 (was 2)        |
+| ``<p>dj-click</p>`` (other dj-* attr)          | 1 (control)      |
+
+Plus a control proving the GOOD path still works: a child that DOES declare its
+own real ``<div dj-root>`` must still render exactly one ``<!DOCTYPE>`` /
+``<footer>`` — the fix must not break the legitimate "child declares dj-root"
+case the original heuristic was written for.
+"""
+
+from __future__ import annotations
+
+import shutil
+import tempfile
+from pathlib import Path
+
+import pytest
+from django.test import override_settings
+
+from djust import LiveView
+from djust.utils import clear_template_dirs_cache
+
+
+class _TemplateHarness:
+    """Write a base + child template pair to a tmp dir and wire it into the
+    Django + Rust template search paths for the duration of a test."""
+
+    def __init__(self, base_src: str, child_src: str):
+        self._tmpdir = Path(tempfile.mkdtemp())
+        (self._tmpdir / "base_1746.html").write_text(base_src)
+        (self._tmpdir / "child_1746.html").write_text(child_src)
+        self._override = None
+
+    def __enter__(self):
+        from django.conf import settings
+
+        templates = [dict(t) for t in settings.TEMPLATES]
+        templates[0] = dict(templates[0])
+        templates[0]["DIRS"] = [str(self._tmpdir), *templates[0].get("DIRS", [])]
+        self._override = override_settings(TEMPLATES=templates)
+        self._override.enable()
+        clear_template_dirs_cache()
+        return self
+
+    def __exit__(self, *exc):
+        if self._override is not None:
+            self._override.disable()
+        clear_template_dirs_cache()
+        shutil.rmtree(self._tmpdir, ignore_errors=True)
+        return False
+
+
+def _make_view_class(name: str):
+    """Build a fresh LiveView subclass per test (avoids class-level state
+    leaking across tests, per the dynamic-subclass discipline #1109)."""
+
+    return type(
+        name,
+        (LiveView,),
+        {
+            "template_name": "child_1746.html",
+            "mount": lambda self, request, **kwargs: None,
+            "get_context_data": lambda self, **kwargs: {},
+        },
+    )
+
+
+# BASE template owns the real reactive root (<div dj-root> in the BASE);
+# single <footer>; single <!DOCTYPE>.
+_BASE = (
+    "<!DOCTYPE html>\n<html>\n<head><title>T</title></head>\n"
+    "<body>\n<nav>NAV</nav>\n"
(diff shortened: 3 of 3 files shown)


johnrtipton:
## Stage-11 Adversarial Review — PR #1747 (fix #1746)

**Verdict: 🟢 APPROVE.** The change is correct, well-targeted, non-tautological, and — the key concern of this review — narrowing the predicate from a substring check to the `<div>`-anchored regexes does **not** break any legitimate root-declaration shape. In fact the `<body dj-view>` shape it was feared to regress is one the fix *also fixes*.

### Pre-review gates
- **Stale-base (#250): PASS.** `BEHIND=0` vs `origin/main`. `git log origin/main..HEAD` shows exactly the 2 PR commits.
- **Two-commit shape: PASS.** `25c15909` = fix (`mixins/template.py`) + test (`test_render_full_template_djroot_substring_1746.py`), **no** CHANGELOG. `26756b1a` = CHANGELOG-only, well-formed `### Fixed` entry under `[Unreleased]`.

### Empirical verification (all run)
1. **New test passes:** 7/7 green on HEAD.
2. **Gate-off (non-tautological):** reverting *only* the predicate back to the substring `in` check → **exactly 4 fail** (`dj_view_text`, `dj_root_text`, `dj_root_view_code`, `substring_adj_view`, all 2 `<!DOCTYPE>`), 3 controls pass (`control_plain`, `control_dj_click`, child-own-`<div dj-root>`). The tests genuinely exercise the change.
3. **No regression:** `test_ssr_render_normalization_1737.py` + `test_template_inheritance.py` + `test_context_processors_in_includes_1722.py` → **26 passed**.

### THE KEY EDGE CASE — does div-anchoring break non-div / `<body dj-view>` roots? **No — verified empirically and structurally.**

The predicate is just a *gate* in front of `_extract_liveview_root_with_wrapper`, which is the function that actually consumes the chosen `vdom_source`. That extractor (`template.py:739-741`) is **already div-only**: `<div\s+[^>]*dj-root[^>]*>` / `<div\s+[^>]*dj-view[^>]*>`. So is the streaming chunker (`_DJ_ROOT_RE` at `template.py:434`) and the shell-replacement path (`template.py:972`). **The entire Python render/extraction pipeline has always been div-only.** The new gate is now *consistent* with the extractor it feeds — the old substring gate was the lone outlier that could select a non-div child the extractor couldn't then use.

Empirical harness (base owns real `<div dj-root>`; child shapes vary), NEW code vs OLD substring code:

| child root shape | NEW (`<div>`-regex) | OLD (substring) |
|---|---|---|
| `<body dj-view="...">` in child | **DOCTYPE=1 ✅** | **DOCTYPE=2 ❌ (double-render)** |
| `<section dj-root>` in child | DOCTYPE=1 ✅ | DOCTYPE=1 ✅ |
| `<div dj-root>` in child (control) | DOCTYPE=1 ✅ | DOCTYPE=1 ✅ |

The `<body dj-view>` case was **broken under the old code** (the substring matched `dj-view`, selected the child as vdom_source, extractor found no `<div...dj-view>`, page double-rendered) and is **fixed** by this PR. `<section dj-root>` worked under both because the substring matched but the extractor's div-regex found nothing and fell through to the resolved full-template path. **Conclusion: the framework's dj-root is always `<div dj-root>`; non-div roots are uniformly unsupported across extraction + streaming, and this fix removes a gate inconsistency rather than introducing one.** (The Rust VDOM `find_liveview_root` at `crates/djust_vdom/src/parser.rs:379-396` *is* element-agnostic, but it runs on already-rendered/extracted HTML, not this selection path, so it's not in scope.)

### Regex on template source (item 3): correct
`template_source = template.template.source` is the raw template *with* `{% %}`/`{{ }}` tags. Probed:
- `<div dj-root>{{ count }}</div>` → match ✅
- `{% if x %}<div dj-root>{% endif %}` → match ✅
- `<div data-dj-root>` / `<p>adj-view</p>` / displayed-text / `<span dj-root>` → correctly no-match ✅
- newline-separated attrs → match ✅

### 🟡 One minor, contrived narrowing (non-blocking)
`<div {% if a %}dj-root{% endif %}>` (dj-root attached conditionally via a template tag with no trailing space) → the OLD substring matched, the NEW regex does **not** (the `{` of `{% endif %}` immediately follows `dj-root`, failing the `(?=[\s=>/])` lookahead). This is a non-idiomatic shape — djust authors write a literal `<div dj-root>` — and a trailing space (`dj-root {% endif %}`) makes it match again. Worth a one-line awareness note; not a real-world blocker.

### Symbol-migration grep (item 4): complete
Only other production hit is `checks.py:3341-3342` (`"dj-view" in tag` / `"dj-root" in tag`). Implementer's "fine" judgment is **correct**: `tag` is `match.group(0)` from `tag_re = <[a-zA-Z][^>]*>`, i.e. already a real extracted tag — displayed text / `adj-view` can't reach it. It's the T005 system-check linter, not a render path. Re-grepped both quote styles across `python/djust/` — no other render/extraction-path instance missed.

### Pre-existing failures (item 5): confirmed pre-existing, NOT caused by this PR
`tests/unit/test_demo_views.py` 4 failures (`test_pwa_view_in_urlconf`, `test_tenant_view_in_urlconf`, `test_service_worker_url_resolves`, `test_manifest_url_resolves`):
- **Pass 43/43 in isolation** on PR HEAD.
- Appear only in cross-dir invocation (`pytest python/ tests/unit/test_demo_views.py`): **4 failed / 5615 passed** on HEAD.
- **Identical 4 failures on `origin/main` with the PR removed: 4 failed / 5608 passed.** URLconf-pollution from sibling dirs, unrelated to this fix.

### Counts
- New test: 7 passed. Gate-off: 4 failed / 3 passed (correct). No-regression suite: 26 passed. Cross-dir: 4 pre-existing fails on both HEAD and main.

Working tree restored to clean PR-branch HEAD (`26756b1a`); pre-existing untracked `scripts/antigravity_pipeline_run.py` left alone.

johnrtipton:
## Retrospective — PR #1747 (#1746)

**Task**: fix render_full_template double-render when a LiveView child template contains literal dj-root/dj-view text.

### Quality: 5/5
One-line root-cause fix with a full variant repro, gate-off, and an edge-case review that turned up a *bonus* fix.

### What went well
- **Root cause was nailed before coding** (template.py:161-165 naive `"dj-root" in template_source` substring gate) via the downstream-driven investigation on djust.org, with a minimal in-framework repro (the variant table) filed in #1746. Implementer verified the line + that `_DJ_ROOT_RE`/`_DJ_VIEW_RE` were in scope, then made the one-line change.
- **The fix is consistency, not just a patch**: the entire Python render/extraction pipeline (`_extract_liveview_root_with_wrapper`, the streaming chunker, the shell-replacement) was already `<div>`-only; the substring gate was the lone outlier that could pick a non-div child the extractor couldn't use. Anchoring the gate to `_DJ_ROOT_RE`/`_DJ_VIEW_RE` makes it consistent with everything downstream.
- **Stage-11 edge-case hunt found a bonus**: narrowing to the div-regex doesn't just avoid breaking `<body dj-view>` children — it *fixes* them (they double-rendered under the old substring gate too). No legitimate root shape regressed.
- Gate-off (#254): reverting the predicate fails exactly the 4 behavior-meaningful cases (dj-view/dj-root/code/adj-view = 2 DOCTYPEs); 3 controls pass either way — non-tautological.
- Symbol-migration grep (canon): found the only other `"dj-root" in`/`"dj-view" in` production hit (checks.py:3341, T005) and correctly judged it safe (operates on extracted tags, not displayed text).

### What didn't
- Nothing blocking. One 🟡: a contrived `<div {% if a %}dj-root{% endif %}>` template shape no longer matches the lookahead — non-idiomatic, non-blocking.
- 4 pre-existing `tests/unit/test_demo_views.py` cross-dir URLconf-pollution failures (pass in isolation; identical on clean main) — unrelated; same class as #1741.

### Process finding
A downstream consumer (djust.org) surfaced a core-framework render bug that the framework's own tests + demo didn't catch — because no test rendered a LiveView whose reactive content *displays* djust attribute names. The minimal in-framework repro (filed in #1746 before the fix) is what made this a clean one-line fix instead of a guess. Reinforces: when a downstream app hits a framework bug, distill it to a minimal in-framework repro first.

### Verified
- New regression test 7/7 (variant table: before dj-view/dj-root/code/adj-view=2 DOCTYPE, after all=1); #1737 + inheritance + #1722 render tests 26/26 (no core-render regression); python/djust/tests/ 3099 passed. All CI green (py3.12, demo-checks, rust-tests). Two-commit shape intact.

`RETRO_COMPLETE`

Issue #1745, opened 2026-06-06
tech-debt: harden the #1742 nav/hooks dogfood guard (make blocking + tighten observer/waits)


johnrtipton:
Source: PR #1744 (#1742) Stage-11 review — 3 non-blocking 🟡 hardening notes.

PR #1744 added a playwright regression guard (tests/playwright/test_nav_hooks.py) for the v1.0.2 nav arc (#1733 SPA nav, #1737 no-hydration-flash, #1738 hook-survives-nav). All 3 assertions are empirically non-tautological (gate-off proven). Three hardening follow-ups:

1. **Make the guard actually merge-blocking.** `playwright-tests` is `continue-on-error: true` (intentionally, since playwright can flake — #1713 separated the blocking `demo-checks` job precisely to keep playwright non-blocking). So a future nav/hydration/hook regression red-bars the *job* but NOT the *merge gate* — weaker than #1742's "red-bars CI in-house" intent. Options: (a) move the nav/hooks invariant checks into (or alongside) the blocking `demo-checks` job; (b) a new small blocking job that runs just this guard; (c) accept job-level visibility and document it. Pick one.

2. **Tighten the #1737 MutationObserver coverage.** The observer is installed post-`page.goto()`, which catches the REAL symptom (the async WS-mount morph fires after goto resolves) but would miss a hypothetical sync-init flash variant. Install via `add_init_script` so the observer is active before first hydration regardless of timing.

3. **Replace the two fixed `asyncio.sleep` waits with condition-waits** (`wait_for_function`) to reduce flakiness headroom on loaded runners. The nav step already uses `wait_for_function`; apply the same to the hook-mount + observe windows.

None block the guard's current value (it catches the real regressions in the playwright job today).

johnrtipton:
Done — all three hardening items landed in PR #1748 (merged):
1. **Made merge-blocking** — new dedicated `nav-hooks-guard` job, added to the `test-summary` AND-condition.
2. **Observer via `add_init_script`** — installed before first navigation (catches early hydration churn).
3. **Condition-based waits** — fixed `asyncio.sleep` replaced with `wait_for_function` on real markers.

Follow-on hardening (maturin build for the new job) shipped in #1753; remaining CI nits tracked in #1752.

PR #1748, opened 2026-06-07
Refs #1745 -- Harden nav/hooks regression guard

BaderEddineBenhirt:
This PR hardens the nav/hooks regression guard.

Changes:
- installs the hydration MutationObserver before the initial navigation using add_init_script.
- replaces fixed asyncio.sleep waits with condition-based Playwright waits.
- adds a dedicated nav-hooks-guard CI job so this regression guard is blocking while the broader Playwright job remains optional.

Tested locally:
- .venv\Scripts\python tests\playwright\test_nav_hooks.py
- git diff --check

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -265,6 +265,9 @@ jobs:
       - name: Install Python dependencies
         run: uv sync --extra dev
 
+      - name: Build package
+        run: uv run maturin develop --release
+
       - name: Install Playwright
         run: |
           uv pip install playwright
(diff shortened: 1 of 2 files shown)


johnrtipton:
Thanks @BaderEddineBenhirt for the contribution.  I'll review run it through the review process.
What has your experience with djust been?  Rough, smooth, problems?

BaderEddineBenhirt:
Thanks @johnrtipton for the quick response. My experience was mostly smooth. The setup with uv and the demo project was straightforward, and the existing Playwright regression script made the issue easy to understand and reproduce locally. The only slightly rough part was understanding which Playwright checks are intended to be blocking versus optional in CI. That is why I kept the broader Playwright job optional and added a dedicated blocking guard only for the nav/hooks regression.

Overall, the repo was clear to work with.

johnrtipton (approved):
## Code Review (pipeline-run Stage 7) — APPROVE with follow-ups

Solid, well-motivated hardening of the #1745 nav/hooks regression guard. All 14 CI checks green (incl. the new `nav-hooks-guard`, 2m9s). No 🔴 blockers.

### Strengths
- **Condition-based waits replace fixed `asyncio.sleep`** — waits on real signals (`__demoWidgetMountCount`, the `djustHookMounted` DOM marker, `location.pathname`, mount-count delta) instead of wall-clock. Faster and far less flaky.
- **Observer installed before `goto` via `add_init_script`** correctly catches the initial hydration churn (#1737) a post-`goto` `evaluate()` could miss; the late-root handling (document observer + `DOMContentLoaded` re-check) is sound.
- **Not greenwashing:** the script `return 1` on any collected failure → `sys.exit(exit_code)` — a real failing gate (verified).
- **Made blocking correctly:** added to `test-summary` `needs` AND the success AND-condition — it genuinely gates, not just echoed.

### 🟡 Follow-ups (non-blocking — tracked, not gating this merge)
1. **`nav-hooks-guard` is the only Rust-dependent CI job without `maturin develop --release`.** Every other ext-dependent job runs `uv sync` **+** `maturin develop`. This job relies on `uv`'s `.venv` cache (key = `hashFiles(pyproject.toml, uv.lock)`), so a PR that changes only `crates/*.rs` (no Python dep change) would test against a **stale cached extension** — a blocking *render* guard that can silently pass on stale code. Recommend adding the `maturin develop --release` step. (Maintainer fast-follow.)
2. **Blocking Playwright job = flake risk to every PR's gate.** `playwright-tests` is intentionally optional because browser jobs flake; promoting a Playwright job to blocking reintroduces that gate-wide. The condition-based waits mitigate it (good); per repo canon a new runner-only job ideally soaks `continue-on-error` before flipping. Green once here, so low risk — flagged for awareness.
3. **~80% CI duplication** with `playwright-tests` (server harness). Consider a composite action / reusable workflow to prevent drift.

### Minor
- Run step uses `.venv/bin/python` while siblings use `uv run` (cosmetic).
- Confirm new action pins (`checkout@v6`, `setup-uv@v7`, `cache@v5`, `upload-artifact@v7`) match versions used elsewhere in this workflow.

Merging; 🟡#1–3 filed as a follow-up. Thanks for the contribution!

APPROVE

johnrtipton:
## Retrospective — PR #1748 (refs #1745)

**Task**: harden the nav/hooks Playwright regression guard (condition-based waits + observer-before-nav) and promote it to a blocking CI job. External contribution by @BaderEddineBenhirt.

### Quality: 4/5
Well-motivated, correctly-implemented hardening; merged green with no 🔴. Held back from 5 only by the blocking-gate-without-`maturin` gap (#1752) that a maintainer fast-follow should close.

### What went well
- **Replacing fixed `asyncio.sleep` with condition-based `wait_for_function`** is exactly the right fix for timing-flaky browser tests — and it waits on real post-conditions (hook-mount markers, path change, mount-count delta), not wall-clock. Same lesson as the async-callback-stub canon: assert the state, don't sleep-and-hope.
- **Observer-before-navigation via `add_init_script`** closes a real coverage hole: a post-`goto` `evaluate()` can miss the initial hydration churn (#1737) entirely.
- **Verified the guard actually fails** (`return 1` → `sys.exit`) before trusting it as blocking — avoids the greenwashing trap (#1037) where a "blocking" gate always exits 0.

### What didn't / to improve
- **A new Rust-ext-dependent CI job shipped without `maturin develop`** while every sibling job has it — so the blocking render guard can test a stale cached extension on Rust-only PRs (#1752 item 1). The reusable-workflow extraction (#1752 item 3) would have prevented this by construction (one server-harness definition, one build step). Lesson: when adding a CI job that mirrors an existing one, factor the shared harness rather than copy-paste — the copy drifts (here, the missing build step).
- **Promoted straight to blocking** without a `continue-on-error` soak (rc4 #1534). Low risk given green CI + the flake-reducing waits, but the canonical sequence is soak-then-flip.

### Verified
- All 14 CI checks green (incl. `nav-hooks-guard`); 0 failing/pending at merge. Approving review posted (pipeline Stage 7). 🟡 follow-ups tracked in #1752.

`RETRO_COMPLETE`

johnrtipton:
@BaderEddineBenhirt thanks for the PR, this has shipped in 1.0.3.

PR #1750, opened 2026-06-07
fix(render): count multi-line <div> opening tags in dj-root boundary scan (#1749)

### LINKED ISSUE #1749: fix: render_full_template ejects page content outside dj-root for multi-line <div> tags (leaks across dj-navigate)
## Symptom

On a LiveView page reached via a fresh GET, page content after the first few sections renders **outside** the `<div dj-root>` container. Because `dj-navigate` swaps only the `[dj-root]` subtree on SPA navigation, that ejected content is never cleared — it **leaks onto the next page**. Observed on djust.org `/examples/`: navigating examples→home left all the demo sections (Kanban, Todo, etc.) on the home page; the page title and `dj-view` attribute also went stale.

## Root cause

`python/djust/mixins/template.py` `render_full_template` finds the `dj-root` region's matching `</div>` by walking the rendered shell and counting `<div>` depth. The opening-tag check only recognized two exact byte forms:

```python
if shell_html[i : i + 5] == "<div " or shell_html[i : i + 5] == "<div>":
    depth += 1
```

A **multi-line opening tag** — `<div\n  class="...">` or `<div\t...>` — was not matched, so it was not counted as an open. Each missed open under-counted depth, so a later `</div>` drove depth to 0 **early**: the `dj-root` region was closed before its real end, and `render_full_template` spliced the (correct, full) `self._rust_view.render()` output in place of the *truncated* region — leaving the tail of the original shell content OUTSIDE `<div dj-root>` (and duplicated).

Same family as #1746 (naive `dj-root` parsing). Single-line `<div ` content (e.g. the home page) was unaffected, which is why it only reproduced on pages authored with multi-line `<div` tags, on a fresh GET (the SPA-mount path puts content cleanly inside `dj-root`).

## Fix

Match `<div` followed by ANY tag-boundary char (whitespace, `>`, `/`), not just `<div `/`<div>`.

## Reproduction / verification

Empirically confirmed against djust.org `/examples/`: 484 real `<div` opens, only 481 counted (3 `<div\n` missed) → demos ejected outside `dj-root`. After the fix: all demos inside `dj-root`, examples→home navigation clean. Regression test `test_render_full_template_multiline_div_boundary.py` with gate-off proof (fails on the old 2-form check: tail marker appears 2×).

## Follow-up (separate)

`dj-navigate` SPA navigation does not update page chrome outside `dj-root` (active-nav highlight, `document.title`, the `dj-root` `dj-view` attribute). Tracking separately — relates to ADR-021.

johnrtipton:
Closes #1749.

## Problem
`render_full_template` finds the `dj-root` region's closing `</div>` by counting `<div>` depth, but the opening-tag check only matched the exact byte forms `"<div "` and `"<div>"`. A **multi-line opening tag** — `<div\n  class="...">` / `<div\t...>` — was not counted as an open. Each missed open under-counted depth, so a later `</div>` drove depth to 0 **early**: the `dj-root` region closed before its real end, the full rendered view was spliced in place of the *truncated* region, and the tail of the page rendered **outside** `<div dj-root>` (and duplicated).

Because `dj-navigate` swaps only the `[dj-root]` subtree on SPA navigation, that ejected content was never cleared → **it leaked onto the next page**. Reproduced on djust.org `/examples/` (demo sections authored as `<div\n class="demo-section">`): navigating examples→home left all demos on the home page. Single-line `<div ` content (e.g. home) was unaffected; only triggers on a fresh GET (the SPA-mount path puts content cleanly inside `dj-root`). Same family as #1746.

## Fix
Match `<div` followed by any tag-boundary char (whitespace, `>`, `/`) — one predicate in `render_full_template`'s boundary scan (`python/djust/mixins/template.py`).

## Verification
- Empirically on djust.org `/examples/`: 484 real `<div` opens, only 481 counted (3 `<div\n` missed) → demos ejected. After fix: all 17 demos inside `dj-root`; examples→home navigation clean (no leak, correct title/`dj-view`).
- New `test_render_full_template_multiline_div_boundary.py` (2 cases) with **gate-off proof**: reverting the predicate makes the tail marker appear 2× (ejected/duplicated) → test fails; with the fix it appears once, inside `dj-root`. Control: single-line `<div>` content stays green.
- Full Python suite: +2 passing tests; the 17 `test_theme_tags_rust_engine_1721` failures in `make test-python` are **pre-existing** test-ordering pollution (identical on clean `origin/main`; pass in isolation; CI runs tests in separate jobs).

## Follow-up (separate, not in this PR)
`dj-navigate` SPA nav doesn't update page chrome outside `dj-root` (active-nav highlight, `document.title`, the `dj-root` `dj-view` attribute). Tracking separately — relates to ADR-021.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- **`render_full_template` ejected page content outside `[dj-root]` for multi-line `<div>` tags, leaking it across `dj-navigate` (#1749).** The `dj-root` region's closing `</div>` was located by counting `<div>` depth, but the opening-tag check only matched the exact forms `"<div "` and `"<div>"` — a multi-line opening tag (`<div\n  class=...>` / `<div\t...>`) was not counted. Each missed open under-counted depth, so a later `</div>` closed the `dj-root` region early; the full rendered view was then spliced in place of the truncated region, leaving the tail of the page **outside** `<div dj-root>` (and duplicated). Because `dj-navigate` swaps only the `[dj-root]` subtree, that ejected content was never cleared on navigation (it leaked onto the next page — observed on a page authored with multi-line `<div>` sections, reached via a fresh GET). Fix: match `<div` followed by any tag-boundary char (whitespace, `>`, `/`). Same family as #1746.
+
 ## [1.0.3rc1] - 2026-06-06
 
 ### Fixed
--- a/python/djust/mixins/template.py
+++ b/python/djust/mixins/template.py
@@ -979,7 +979,20 @@ Object.assign(window.handlerMetadata, {json.dumps(metadata)});
                 depth = 1
                 i = after_open
                 while i < len(shell_html) and depth > 0:
-                    if shell_html[i : i + 5] == "<div " or shell_html[i : i + 5] == "<div>":
+                    # Opening <div ...>: match "<div" followed by a tag-boundary
+                    # char (whitespace, '>' or '/'). The previous check only
+                    # recognized the exact forms "<div " and "<div>", so a
+                    # MULTI-LINE opening tag ("<div\n  class=...>" / "<div\t...")
+                    # was not counted as an open. Each missed open under-counted
+                    # depth, so a later "</div>" drove depth to 0 EARLY — the
+                    # dj-root region was closed before its real end and all
+                    # subsequent page content rendered OUTSIDE [dj-root]. Because
+                    # dj-navigate only swaps the [dj-root] subtree, that ejected
+                    # content was never cleared on navigation (leaked onto the
+                    # next page). Matching any tag-boundary char fixes the count.
+                    if shell_html[i : i + 4] == "<div" and (
+                        i + 4 >= len(shell_html) or shell_html[i + 4] in " \t\r\n>/"
+                    ):
                         depth += 1
                     elif shell_html[i : i + 6] == "</div>":
                         depth -= 1
--- a/python/djust/tests/test_render_full_template_multiline_div_boundary.py
+++ b/python/djust/tests/test_render_full_template_multiline_div_boundary.py
@@ -1,0 +1,191 @@
+"""Regression tests for the dj-root boundary depth-counter in
+``render_full_template`` — multi-line ``<div`` opening tags must be counted (#1749).
+
+Root cause
+----------
+``python/djust/mixins/template.py`` ``render_full_template`` locates the
+``dj-root`` region's matching ``</div>`` by walking the rendered shell and
+counting ``<div>`` depth. The opening-tag check originally only recognized two
+EXACT byte forms::
+
+    if shell_html[i : i + 5] == "<div " or shell_html[i : i + 5] == "<div>":
+        depth += 1
+
+A MULTI-LINE opening tag — ``<div\n  class="...">`` or ``<div\t...>`` — was not
+matched, so it was NOT counted as an open. Each missed open under-counted depth,
+so a later ``</div>`` drove depth to 0 EARLY: the ``dj-root`` region was closed
+before its real end. ``render_full_template`` then spliced the (correct,
+full) ``self._rust_view.render()`` output in place of the *truncated* region,
+leaving the tail of the original shell content OUTSIDE ``<div dj-root>`` — so
+that tail rendered both inside the root (from the rust output) AND as a sibling
+after it (the leftover shell). Two consequences:
+
+1. The tail content is DUPLICATED in the initial GET HTML.
+2. Because ``dj-navigate`` swaps only the ``[dj-root]`` subtree on navigation,
+   the ejected sibling copy is never cleared — it "leaks" onto the next page.
+
+Observed downstream: djust.org ``/examples/`` (demo sections authored as
+``<div\n class="demo-section">``) ejected every demo after the first outside
+``dj-root``; navigating examples→home left the demos on the home page.
+
+Fix
+---
+Match ``<div`` followed by ANY tag-boundary char (whitespace, ``>`` or ``/``),
+not just the exact ``"<div "`` / ``"<div>"`` forms.
+"""
+
+from __future__ import annotations
+
+import shutil
+import tempfile
+from html.parser import HTMLParser
+from pathlib import Path
+
+from django.test import override_settings
+
+from djust import LiveView
+from djust.utils import clear_template_dirs_cache
+
+
+class _TemplateHarness:
+    def __init__(self, base_src: str, child_src: str):
+        self._tmpdir = Path(tempfile.mkdtemp())
+        (self._tmpdir / "base_mlb.html").write_text(base_src)
+        (self._tmpdir / "child_mlb.html").write_text(child_src)
+        self._override = None
+
+    def __enter__(self):
+        from django.conf import settings
+
+        templates = [dict(t) for t in settings.TEMPLATES]
+        templates[0] = dict(templates[0])
+        templates[0]["DIRS"] = [str(self._tmpdir), *templates[0].get("DIRS", [])]
+        self._override = override_settings(TEMPLATES=templates)
+        self._override.enable()
+        clear_template_dirs_cache()
+        return self
+
+    def __exit__(self, *exc):
+        if self._override is not None:
+            self._override.disable()
+        clear_template_dirs_cache()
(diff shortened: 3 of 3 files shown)


johnrtipton:
## Stage-7 Adversarial Review — PR #1750 (fix #1749)

**Verdict: APPROVE** (with one 🟡 latent-mirror-bug finding tracked as a follow-up, not a merge blocker)

The fix is correct, minimal, and the regression test is genuinely behavior-coupled (proven by an independent gate-off). The one substantive finding is a latent *close-side* symmetry gap that this PR does not introduce and does not worsen — it is the symmetric twin of the bug being fixed and exists identically in a sibling counter. Flagging per the #1646 parallel-path-drift canon, but it is lower-probability than the open-side bug and out of scope for this fix.

---

### Mandatory checks

- 🟢 **Stale-base (#250):** `BEHIND=0`. `git log origin/main..HEAD` shows exactly the 2 PR commits (`c5e73afa` fix+test, `8f00f452` CHANGELOG). Reviewing the program that will merge.
- 🟢 **Two-commit shape:** `c5e73afa` = `template.py` + new test only, NO CHANGELOG. `8f00f452` = CHANGELOG only. Correct boundary.
- 🟢 **Three-dot diff:** `CHANGELOG.md (+4)`, `template.py (+15/-1)`, new test (+191). Scoped.

### Predicate correctness (empirically replicated, not eyeballed)

🟢 Standalone replication of the new open-check over crafted strings confirmed:
- COUNTS all real opens: `<div>`, `<div ` (space), `<div\n`, `<div\t`, `<div\r`, `<div/`, and bare `<div` at EOF.
- Does NOT count `</div>` as an open (`shell_html[i:i+4]` at the `<` of `</div>` is `</di` ≠ `<div`).
- Does NOT count `<divider` / `<div-foo` / `<division` (char after `<div` is `i`/`-`, not in the boundary set `" \t\r\n>/"`).

### 🟡 Close-side mirror bug — LATENT (`python/djust/mixins/template.py:997`)

The OPEN check now tolerates whitespace; the CLOSE check is still hardcoded `shell_html[i:i+6] == "</div>"`. I traced the Rust renderer to determine whether a whitespace-bearing close tag (`</div >`, `</div\n>`) can ever reach `shell_html`:

- The Rust template engine (`crates/djust_templates/src/renderer.rs`) is a **Django-style text-passthrough engine**, not an HTML DOM serializer. `Node::Text(text) => Ok(text.clone())` (`renderer.rs:387`) emits authored HTML — including close tags — **verbatim**. The only place a `</div>` is *synthesized* is `renderer.rs:759` (the `ReactComponent` mount-point node), which emits the canonical no-whitespace form.
- **Empirically confirmed** by rendering `'<div dj-root>\n  <div\n class="inner">hi</div >\n</div>TAIL'` through the real `RustLiveView(src).render()`: the output preserves `</div >` byte-for-byte. Newline-before-`>` (`</div\n>`) is likewise passed through verbatim.
- **Failure mode (simulated against the real boundary loop):** a template whose inner div uses `</div >` and whose dj-root closes with canonical `</div>` drives the loop to `FALLBACK_DEPTH_NEVER_ZERO` — the inner `<div ` open counts (depth→2) but its `</div >` close is not matched (depth stays 2), so the real outer `</div>` only reaches depth 1 and the loop exhausts → returns the un-normalized shell (the #1737 first-hydration-flash class).

**Determination: NOT safe — a latent mirror bug exists.** It is the symmetric twin of #1749. It is lower-probability (close tags rarely carry whitespace before `>`, and the synthesized React close tag is always canonical), so it does not block this PR, but per #1646 it should be filed as a follow-up. The cleanest structural cure is below.

### 🟡 Parallel-path-drift (#1646)

`grep` confirms `template.py:993` is the **only hand-rolled character-scan div-depth counter**. BUT there is a second, regex-based div-boundary scanner: `TemplateMixin._find_closing_div_pos` (`template.py:775`), used by **5 call sites** (`:451`, the three `_extract_liveview*` helpers, and `:855`). It:
- uses `re.search(r"<div\b", ...)` for opens — `\b` already tolerates multi-line opens, so it never had the open-side bug;
- uses `re.search(r"</div>", ...)` for closes — **same hardcoded close form → same latent whitespace-close gap** as the fixed counter.

So the close-side gap is consistent across both scanners (both latently broken on `</div >`), and the open-side handling is now consistent too. **Recommendation for the follow-up:** rather than maintaining two counters, have `render_full_template` delegate to `_find_closing_div_pos` (it already handles balanced nesting *and* `{% if/else %}` branch restoration, which the hand-rolled loop does not). The hand-rolled duplicate at `:993` is the kind of "N correct copies" #1646 warns against. Not in scope for this fix; noting for the class.

### Gate-off (independently run) — test is NOT tautological 🟢

Reverted ONLY the predicate to the old 2-form check, re-ran:
- `test_tail_after_multiline_divs_stays_inside_dj_root` **FAILS** — `TAIL_MARKER_XYZ appears 2 times` (the exact duplication symptom of #1749).
- `test_single_line_divs_unaffected` (control) still **PASSES**.

Restored the fix; working tree clean. The test exercises the real `render_full_template` path end-to-end (Rust render → boundary scan → HTMLParser containment assertion), not a synthetic loop copy — high reproduction fidelity.

### Pollution claim 🟢

`test_theme_tags_rust_engine_1721.py` passes in isolation (**17 passed**). The 17 full-suite failures are an ordering/isolation artifact, pre-existing and unrelated. This PR edits only the dj-root boundary loop in `render_full_template`; it cannot causally affect theme tag rendering.

### No-regression 🟢

`test_render_full_template_djroot_substring_1746.py` + `test_ssr_render_normalization_1737.py` + `tests/unit/test_template_inheritance.py`: **31 passed**.

---

### Summary

Correct, well-tested, well-scoped fix. **APPROVE.** Recommend a follow-up issue for the close-side whitespace-tolerance gap (present in both `template.py:997` and `_find_closing_div_pos:804`), ideally resolved by collapsing the two div-boundary scanners into `_find_closing_div_pos`.

johnrtipton:
## Retrospective — PR #1750 (#1749)

**Task**: fix render_full_template ejecting page content outside `[dj-root]` for multi-line `<div>` tags (leaked across dj-navigate).

### Quality: 5/5
Tight one-predicate fix, root-caused empirically against a live downstream symptom, gate-off-proven test, systemic remainder tracked.

### What went well
- **Symptom-up diagnosis under a misleading hypothesis.** The reported symptom ("examples leak onto home on navigation") looked like #1746 and like stale cache; both were ruled out empirically (server clean on all paths; mount frame clean; SPA repro). The real cause — a naive `<div `/`<div>`-only depth counter missing multi-line `<div\n` opens — was found by counting real vs. counted `<div` opens in the live served HTML (484 vs 481). Trust-the-symptom-not-the-cited-path paid off.
- **Reproduction fidelity.** The bug only manifests on a fresh GET (`render_full_template`), not the SPA-mount path — so the repro had to land on the page via GET. The regression test drives real `render_full_template` end-to-end with multi-line `<div\n>` content; gate-off (revert predicate → tail marker appears 2×) proves it's non-tautological.
- **Scope discipline (#1079).** Fixed exactly the cited open-side bug; filed the latent close-side mirror (`</div >` whitespace) + two-scanner consolidation as follow-up #1751 rather than expanding this PR.
- **Two-commit shape + Gates 1/2 clean; pre-merge gate enforced.**

### What didn't
- **Long investigation before the fix.** Multiple wrong turns (nested `<main>`, service-worker cache, the client mount handler — a passing JS test that was later removed) because the browser MCP couldn't reach the affected tab and the byte-level `<div>` depth count was fooled by `<script>`/code-sample text. Resolved once an HTML-parser-based (reliable) containment check and the live `<div`-open census were used.
- **Pre-existing test pollution surfaced.** `make test-python` fails 17 `test_theme_tags_rust_engine_1721` cases by ordering (pass in isolation; identical on clean main; CI unaffected). Not this PR's bug — worth a separate pollution-isolation issue.

### Verified
- Empirical (djust.org `/examples/`): post-fix all 17 demos inside `dj-root`, examples→home navigation clean (no leak; correct title/`dj-view`).
- New regression test 2/2 with gate-off proof; render-path suite (#1746/#1737/inheritance) green; +2 net passing in full suite. CI green (py3.12, rust-tests, demo-checks). Stale-base 0.

### Follow-ups
- #1751 — close-side `</div>` whitespace tolerance + consolidate the two boundary scanners.
- Separate (unfiled): `dj-navigate` SPA nav doesn't update page chrome outside `dj-root` (active-nav highlight, `document.title`, `dj-view` attr) — relates to ADR-021. Pre-existing `make test-python` theme-tag ordering pollution.

`RETRO_COMPLETE`

PR #1753, opened 2026-06-08
ci(test): build Rust extension in nav-hooks-guard before running the guard (#1752)

### LINKED ISSUE #1752: tech-debt: harden nav-hooks-guard CI job (maturin build, blocking-soak, dedup) — follow-up to #1748
Follow-up to PR #1748 (Stage-7 review). Three 🟡 items, non-blocking but worth closing — #1 is the priority.

## 1. Add `maturin develop --release` to `nav-hooks-guard` (priority)
`nav-hooks-guard` (`.github/workflows/test.yml`) is the only Rust-extension-dependent CI job that does NOT run `maturin develop --release` after `uv sync` — every other ext-dependent job does. Its `.venv`/uv cache key is `hashFiles('**/pyproject.toml','**/uv.lock')`, so a PR that changes only `crates/*.rs` (no Python dep change) restores a **stale cached djust extension** and the guard tests OLD rendering behavior. Since the guard is now **blocking**, it can give false confidence on stale code (or false-red). Fix: add the `maturin develop --release` build step (mirror `playwright-tests` / `python-tests`).

## 2. Blocking-vs-soak for a Playwright job
`playwright-tests` is intentionally optional (browser jobs flake). `nav-hooks-guard` (also Playwright) was promoted straight to blocking. Per repo canon (rc4 #1534), a new runner-only job ideally ships `continue-on-error: true` until it has been green across several runs, then flips to blocking. The condition-based waits (#1748) reduce flake, so risk is low — decide whether to soak briefly or accept the trade.

## 3. CI duplication
`nav-hooks-guard` duplicates ~80% of `playwright-tests` (checkout → python → uv → cache → playwright install → migrate → start server → wait → run → logs → stop). Extract a composite action / reusable workflow so the two server harnesses can't drift.

Minor: the run step uses `.venv/bin/python` vs `uv run` elsewhere (cosmetic); verify action pins (`checkout@v6`, `setup-uv@v7`, `cache@v5`, `upload-artifact@v7`) match the rest of the workflow.

Source: PR #1748 review (https://github.com/djust-org/djust/pull/1748).

johnrtipton:
Closes #1752 (item 1).

## Problem
`nav-hooks-guard` (added in #1748, now a **blocking** CI gate) was the only Rust-extension-dependent job without a `maturin develop` build step. Its `uv`/`.venv` cache key is `hashFiles('**/pyproject.toml','**/uv.lock')`, so a PR that changes only `crates/*.rs` (no Python-dep change) restores a **previously-built djust** and the guard exercises **old rendering behavior** — a blocking render guard that can pass (or fail) against a stale extension.

## Fix
Add `uv run maturin develop --release` after `uv sync --extra dev`, mirroring the `python-tests` and `playwright-tests` jobs, so the guard always tests freshly-built code.

```yaml
- name: Install Python dependencies
  run: uv sync --extra dev
+ - name: Build package
+   run: uv run maturin develop --release
- name: Install Playwright
```

## Verification
- YAML parses; step order correct (Install deps → Build package → Install Playwright); build command byte-identical to the sibling `playwright-tests` job.
- The empirical proof is this PR's own CI run: `nav-hooks-guard` must stay green **with** the build step (confirms the step works and doesn't break the job).

## Follow-ups (remain in #1752)
- Item 2: blocking-vs-`continue-on-error` soak for the Playwright-based guard.
- Item 3: extract a reusable workflow to de-duplicate the ~80% shared server harness between `nav-hooks-guard` and `playwright-tests` (would prevent exactly this kind of copy-paste drift).

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -359,6 +359,13 @@ jobs:
             ${{ runner.os }}-python-
       - name: Install Python dependencies
         run: uv sync --extra dev
+      - name: Build package
+        # Force a fresh Rust-extension build so this BLOCKING render guard never
+        # tests a stale cached extension. The uv/.venv cache key is keyed on
+        # pyproject.toml/uv.lock only, so a PR that changes just crates/*.rs would
+        # otherwise restore a previously-built djust and exercise old rendering.
+        # Mirrors the python-tests / playwright-tests jobs. (#1752)
+        run: uv run maturin develop --release
       - name: Install Playwright
         run: |
           uv pip install playwright


johnrtipton:
## Code Review (pipeline-run Stage 7) — APPROVE

Single, well-scoped CI fix; closes #1752 item 1.

**What it does:** `.github/workflows/test.yml` — inserts a `Build package` step (`uv run maturin develop --release`) into `nav-hooks-guard` between `Install Python dependencies` (`uv sync --extra dev`) and `Install Playwright`, mirroring the identical step in `python-tests` / `playwright-tests`.

**Correctness:**
- Step ordering right: `uv sync` populates the env, then `maturin develop --release` rebuilds the Rust extension fresh, before the server starts and the guard runs. Build command byte-identical to the sibling `playwright-tests` step.
- **Empirical canary (#252):** this PR's `nav-hooks-guard` ran green *with* the new step (2m59s vs ~2m09s before — delta is the maturin build), proving the step works and doesn't break the job.

**Concern / edge cases considered:**
- Cache interaction: the `uv`/`.venv` cache key is still `pyproject.toml`/`uv.lock`-only, so a crates-only PR still restores a stale `.venv` — but `maturin develop --release` rebuilds the ext on top, so correctness no longer depends on the cache key (remaining effect is a correct Python-dep cache).
- Build-failure mode: if `maturin develop` fails the job fails — desired for a blocking render guard (better than testing stale code).
- CI cost: +~50s on one job; acceptable.

Items 2 (blocking-soak) and 3 (CI dedup) remain tracked in #1752.

APPROVE

johnrtipton:
## Retrospective — PR #1753 (#1752 item 1)

**Task**: add `maturin develop --release` to the `nav-hooks-guard` CI job so the blocking render guard never tests a stale cached Rust extension.

### Quality: 5/5
Minimal, surgical CI fix; empirically validated by its own green CI run; mirrors the established sibling pattern exactly.

### What went well
- **Empirical canary (#252) was free here:** the PR's own `nav-hooks-guard` run *is* the proof — green with the build step (2m59s vs ~2m09s, the delta being the maturin build). A CI-change PR validates itself on the runner.
- Mirrored the exact `Build package` step from `python-tests`/`playwright-tests` (byte-identical command) rather than inventing a variant — verified equality programmatically before commit.
- Scope discipline (#1079): fixed only item 1 (the real defect); left items 2 (blocking-soak) and 3 (CI-dedup) tracked in #1752.

### What could be improved
- **The root enabler was copy-paste CI duplication** (#1748 added a job by copying `playwright-tests` and dropping one step). This is exactly what #1752 item 3 (extract a reusable workflow) would prevent structurally — a single server-harness + build definition can't drift a step out. The point fix (this PR) patches the instance; item 3 retires the class. Ties to the parallel-path-drift canon (#1646): N near-identical CI jobs are the same hazard as N near-identical code paths.
- "Closes #1752" in the PR body auto-closed the whole issue on merge even though only item 1 landed — had to reopen for items 2–3. Lesson: for multi-item issues, use "Refs #NNNN" not "Closes", or split items into separate issues upfront.

### Verified
- YAML valid, step order correct, build cmd matches sibling; CI fully green (nav-hooks-guard, playwright, rust, python, demo-checks). Merged squash.

`RETRO_COMPLETE`

Issue #1742, opened 2026-06-06
tech-debt: dogfood dj-navigate cross-view flow + a client-hook (3rd-party lib) in the demo so demo-checks catches nav/hydration/hook regressions


johnrtipton:
Source: Retro v1.0.2 navigation arc (PRs #1736, #1739, #1740).

The whole v1.0.2 nav/hydration arc (#1733 zero-wiring route map, #1737 hydration-flash parity, #1738 client-hooks for 3rd-party libs) was driven by a downstream consumer (rent app), NOT caught internally — because the demo project doesn't exercise these paths end-to-end. The #1736 reviewer found T016 stays silent on the demo (the checked scope exercises the no-dj-navigate branch); there is no client-hook / third-party-lib example in the demo at all.

**Action:** add to `examples/demo_project`:
1. A `dj-navigate` cross-view navigation flow between two LiveViews (exercises the auto-derived route map from #1733 + the SSR→hydration path from #1737/#1724) — so a MutationObserver/playwright check can assert no full-reload + no first-hydration subtree churn.
2. A client-hook example using the `DjustHooks`/`dj-hook` pattern from #1738 (a small canvas/widget initialized in `mounted()`, surviving a `dj-navigate`) — so the hook-survives-SPA-nav path is dogfooded.
3. Wire a playwright (or demo-checks) assertion so a regression in nav/hydration/hooks red-bars CI internally instead of surfacing downstream.

Closes the gap that made this entire arc reactive rather than caught-in-house.

PR #1755, opened 2026-06-08
ci: extract shared Playwright-server harness into a composite action (#1752)

### LINKED ISSUE #1752: tech-debt: harden nav-hooks-guard CI job (maturin build, blocking-soak, dedup) — follow-up to #1748

johnrtipton:
Closes #1752.

## Problem
`nav-hooks-guard` and `playwright-tests` duplicated ~9 setup steps (Python/uv/cache/`uv sync`/maturin build/Playwright install/migrate/start server/wait). #1753 had to patch the missing `maturin develop` step into `nav-hooks-guard` *specifically because* the copy had drifted out of sync — textbook parallel-path drift (#1646).

## Fix (item 3)
Extract the shared harness into a composite action `.github/actions/djust-playwright-server`. Both jobs now `actions/checkout` → `uses: ./.github/actions/djust-playwright-server` → run their own test step → upload `server.log` on failure → stop server.

- **One definition** can't drift a step out of one job.
- Jobs keep their identity: `playwright-tests` stays `continue-on-error: true` (optional), `nav-hooks-guard` stays blocking; `test-summary` `needs`/AND-condition unchanged.
- Net −31 lines (`+95/−126`).

## Items 1 & 2
- **Item 1** (maturin build) shipped in #1753 — now lives in the composite, so it can't drift again.
- **Item 2** (blocking-vs-soak) resolved **by decision**: keep `nav-hooks-guard` blocking. The soak concern (#1534) is retroactively satisfied — it has run green across #1748, #1753, and #1754, and the condition-based waits (#1748) reduce flake.

## Verification
- Local: both YAML files parse; every composite `run` step declares `shell: bash`; both jobs preserve checkout→composite ordering + their stop/upload steps + `continue-on-error` settings.
- **The empirical proof is this PR's own CI**: both `playwright-tests` and `nav-hooks-guard` must run green *through the composite* (a new runner-only path; per #1534, ≥1 runner iteration may be expected for composite-action quirks).

### DIFF
--- a/.github/actions/djust-playwright-server/action.yml
+++ b/.github/actions/djust-playwright-server/action.yml
@@ -1,0 +1,85 @@
+name: 'djust Playwright server'
+description: >-
+  Set up Python + uv, build the djust Rust extension, install Playwright
+  (chromium), run Django migrations, and start the demo project's dev server
+  on :8002, waiting until it answers. Shared by the playwright-tests and
+  nav-hooks-guard jobs (#1752). The caller must `actions/checkout` first
+  (this is a local action) and is responsible for running its own test step,
+  uploading server.log on failure, and stopping the server.
+
+runs:
+  using: composite
+  steps:
+    - name: Set up Python
+      uses: actions/setup-python@v6
+      with:
+        python-version: '3.12'
+
+    - name: Install uv
+      uses: astral-sh/setup-uv@v7
+
+    - name: Cache Python dependencies
+      uses: actions/cache@v5
+      with:
+        path: |
+          ~/.cache/uv
+          .venv
+        key: ${{ runner.os }}-python-playwright-${{ hashFiles('**/pyproject.toml', '**/uv.lock') }}
+        restore-keys: |
+          ${{ runner.os }}-python-playwright-
+          ${{ runner.os }}-python-
+
+    - name: Install Python dependencies
+      shell: bash
+      run: uv sync --extra dev
+
+    - name: Build package
+      # Force a fresh Rust-extension build so render-behavior guards never test a
+      # stale cached extension. The uv/.venv cache key is keyed on
+      # pyproject.toml/uv.lock only, so a crates/*.rs-only change would otherwise
+      # restore a previously-built djust. (#1752)
+      shell: bash
+      run: uv run maturin develop --release
+
+    - name: Install Playwright
+      shell: bash
+      run: |
+        uv pip install playwright
+        .venv/bin/playwright install chromium
+
+    - name: Run Django migrations
+      shell: bash
+      run: |
+        cd examples/demo_project
+        uv run python manage.py migrate --noinput
+        cd ../..
+
+    - name: Start development server
+      shell: bash
+      env:
+        PYTHONPATH: .
+      run: |
+        cd examples/demo_project
+        nohup uv run python -m uvicorn demo_project.asgi:application \
+          --host 0.0.0.0 \
+          --port 8002 \
+          --log-level info \
+          > server.log 2>&1 &
+        echo $! > server.pid
+        cd ../..
+
+    - name: Wait for server to be ready
+      shell: bash
+      run: |
+        echo "Waiting for server to start..."
+        for i in {1..30}; do
+          if curl -s http://localhost:8002/ > /dev/null; then
+            echo "Server is ready!"
+            exit 0
+          fi
+          echo "Attempt $i/30: Server not ready yet..."
+          sleep 1
+        done
+        echo "Server failed to start"
+        cat examples/demo_project/server.log
+        exit 1
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -244,72 +244,11 @@ jobs:
     steps:
       - uses: actions/checkout@v6
 
-      - name: Set up Python
-        uses: actions/setup-python@v6
-        with:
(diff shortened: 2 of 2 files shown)


johnrtipton:
## Code Review (pipeline-run Stage 7) — APPROVE

CI-only refactor; closes #1752. No 🔴.

**What it does:** extracts the 9 shared setup steps into `.github/actions/djust-playwright-server/action.yml` (composite); `playwright-tests` and `nav-hooks-guard` become `checkout → uses: ./.github/actions/djust-playwright-server → <own test> → upload-log-on-failure → stop-server`.

**Correctness:**
- **Empirical canary (#252):** this PR's CI ran **both** jobs green *through the composite* (nav-hooks-guard 3m30s, playwright-tests 3m30s) — proves the background server started in a composite step persists into the caller's test step, and the local-action path resolves after checkout.
- Every composite `run` step declares `shell: bash` (required for composite run steps); `uses:` steps (setup-python/uv/cache) don't need it. Verified.
- Job identities preserved: `playwright-tests` keeps `continue-on-error: true` (optional), `nav-hooks-guard` has no `continue-on-error` (blocking); `test-summary` `needs`/AND-condition reference the same job names → gating unchanged.

**Concern / edge cases considered:**
- **Cache key unified** to `${{ runner.os }}-python-playwright-…` (was a separate `-nav-hooks-` key). Both jobs now share one cache — harmless (same deps), and `restore-keys` still falls back to `-python-`. 
- The historical demo-checks decoupling note (#1708/#1713) was preserved inline in `playwright-tests`.
- The composite's maturin build (item 1) now lives in ONE place → can't drift out of one job again (the #1646 failure mode that caused #1753).

**Item 2** (blocking-soak): resolved by decision — keep `nav-hooks-guard` blocking; green across #1748/#1753/#1754 + #1755 satisfies the #1534 soak.

APPROVE

johnrtipton:
## Retrospective — PR #1755 (#1752)

**Task**: extract the shared Playwright-server harness into a composite action (#1752 item 3); resolve item 2 (keep blocking) by decision. Item 1 (maturin) shipped in #1753.

### Quality: 5/5
Fixes the root cause of the whole #1752 thread — the duplication that let the maturin step drift out of one job — not just its symptom.

### What went well
- **Root-cause over symptom.** #1753 patched the *missing* maturin step into nav-hooks-guard; this PR removes the *duplication that allowed it to go missing*. The composite is one definition; a step can no longer exist in one copy and not the other (#1646 parallel-path-drift, now applied to CI not just code).
- **Empirical canary (#252) for free.** A composite action has real first-use unknowns (does a backgrounded server survive the composite→caller step boundary? does the local-action path resolve post-checkout?). This PR's own CI answered both: nav-hooks-guard and playwright-tests both ran green *through* the composite. No guessing.
- **Identity preserved through the refactor.** continue-on-error (playwright optional / nav blocking) and the test-summary AND-condition were left untouched because the jobs stayed jobs — the lower-risk composite-action choice over a reusable workflow (which would have restructured the gating).
- Net −31 lines while adding a reusable building block.

### What could be improved
- The duplication existed for several PRs (#1742 → #1748 → #1753) before being factored out; the drift (#1753) is what finally motivated it. Lesson reinforced: when a CI job is added by copy-pasting an existing one (#1748 copied playwright-tests), factor the shared part **then**, not after a step drifts.

### Verified
- Both YAML files parse; composite run-steps all declare `shell: bash`; both jobs preserve checkout→composite ordering + stop/upload + continue-on-error. CI fully green — both Playwright jobs through the composite, plus rust/python/demo-checks.

`RETRO_COMPLETE`

Issue #1756, opened 2026-06-08
feat(nav): dj-navigate should keep active-link state (aria-current) in sync across SPA navigation


johnrtipton:
## Problem
`dj-navigate` swaps only the `[dj-root]` subtree on SPA navigation. Any "page chrome" rendered OUTSIDE `[dj-root]` — a persistent top nav, the document `<title>`, the `dj-root` element's own `dj-view` attribute — is never updated on a client-side navigation, so it goes stale.

Most visible symptom: a persistent nav whose active link is server-rendered (e.g. `{% if slug == active_page %}active{% endif %}`) keeps highlighting the page you navigated *from*. Confirmed on djust.org (Examples → another page left "Examples" highlighted); worked around there with a per-app client script (djust.org `marketing/static/js/nav-active.js`), but every djust app with a persistent nav + `dj-navigate` hits this.

## Proposed (framework)
On each `dj-navigate` navigation (and popstate), have the client set `aria-current="page"` on the `[dj-navigate]` link whose resolved path matches the new URL and remove it from the others. Apps style `[aria-current="page"]`. Accessible by construction, and zero per-app JS.

Related stale chrome to consider in the same pass:
- `document.title` — there's a `page_metadata` push mechanism, but static views don't use it on SPA mount, so the title goes stale too.
- the `[dj-root]` element's `dj-view` attribute isn't updated on the mount-frame `innerHTML` replace (stays the previous view's path).

## Notes
Pairs with ADR-021 (#1734 auto_navigate) — both are about making `dj-navigate` a complete cross-view navigation. v1.1.0. The djust.org workaround can be removed once this lands.

johnrtipton:
**Part 1 done** in PR #1757 (merged `0b708c0d`): `dj-navigate` now auto-manages `aria-current="page"` on `[dj-navigate]` links across SPA navigation (click + WS mount + popstate), cross-origin-safe, non-clobbering. Apps style `a[dj-navigate][aria-current="page"]`. Keeping this open for the remaining stale-chrome items:

- [x] **aria-current active-link sync** (PR #1757)
- [ ] **`document.title` sync** — needs the mount frame to carry the new view's title (server/wire change); static views don't push it via `page_metadata` today.
- [ ] **`[dj-root]` `dj-view` attribute** not updated on the mount-frame `innerHTML` replace (stays the previous view's path).

Once a release ships part 1, djust.org can drop its `nav-active.js` workaround and style `[aria-current="page"]`.

[...]

Issue #1077, opened 2026-04-25
tech-debt: Lift-from-downstream FIRST pattern (canonicalize via Stage 4 plan-template)


johnrtipton:
**Source**: Retro v0.8.2 / Action Tracker row #152 / PR #1074 (#1009 prose.css)

**Action**: Add to Stage 4 plan-template guidance in `~/.claude/skills/pipeline-run/SKILL.md`: when an issue cites a downstream consumer's working solution (e.g. "docs.djust.org wrote the bridge in its own input.css"), lift the reference impl verbatim FIRST, generalize SECOND. Skips a design-from-scratch phase that risks producing an incompatible variant.

**Empirical example**: PR #1074's `prose.css` was lifted near-verbatim from docs.djust.org's `static/css/input.css` — 91 lines copied + opt-in via `prose-djust` class. Time-to-implement was a fraction of clean-room. Reference impl was already battle-tested against three theme packs in production.

**Implementation**: one bullet under Stage 4 plan-template's "reuse existing functionality" section. Skill-level work — labeled `out-of-scope-for-djust-drain`.

---
_Filed by /pipeline-retro v0.8.2 milestone retro Stage 4._

johnrtipton:
Shipped in PR #1226 (v0.9.1-7 cleanup batch). Comma-list `Closes #...` form parsed only the first ref — the rest closed manually. (This bites again; #1184's parenthesized form is documented in PR-checklist, but the comma-list-on-its-own-line form has the same auto-close failure.) 🤖 v0.9.1-7 cleanup batch closure.

PR #1757, opened 2026-06-08
feat(nav): dj-navigate keeps aria-current="page" in sync across SPA nav (#1756)

johnrtipton:
Addresses #1756 (part 1 — the primary, titled ask). `document.title` and the `[dj-root]` `dj-view`-attr sync remain tracked there.

## Problem
A persistent nav usually lives **outside** `[dj-root]`. `dj-navigate` swaps only the `[dj-root]` subtree, so a server-rendered active-link highlight (`{% if slug == active_page %}…`) never updates on SPA navigation — it stays on the page you came from. (Reported on djust.org: Examples → another page kept "Examples" highlighted; worked around there with a per-app script.)

## Fix (framework, generic, accessible)
`dj-navigate` now manages `aria-current="page"` itself. After each navigation it sets `aria-current="page"` on the `[dj-navigate]` link whose path matches the current URL and removes it from the others. Three call sites in `18-navigation.js`:
- `handleLiveRedirect` (right after `pushState`) — instant feedback on click;
- `bindNavigationDirectives` (runs via `reinitAfterDOMUpdate`) — covers the WS mount + patches + initial load;
- the `popstate` handler — back/forward.

Apps just style `a[dj-navigate][aria-current="page"]` — **no per-app JS**.

Guarantees:
- Cross-origin `dj-navigate` targets (sister-site links whose pathname is `/`) are **never** marked current.
- An app-authored `aria-current` of a *different* value (`step`, `true`, …) is left untouched — we only manage the `"page"` value we set.
- Exact-path match (the ARIA-correct semantic for `page`); section/ancestor highlighting is left to the app.

## Verification
- `tests/js/dj_navigate_aria_current_1756.test.js` — 4 cases (match-moves-on-nav, cross-origin-never-current, app-value-not-clobbered, function exposed). **Gate-off proven**: neutering `updateAriaCurrent` fails 3/4.
- Bundle rebuilt from src (`make build-js`); `client.js` + min/br/gz/map committed.
- Docs: navigation.md gains an "Active-link highlighting" section.

Once released, djust.org can drop its `nav-active.js` workaround and style `[aria-current="page"]` instead.


johnrtipton:
## Stage-7 Adversarial Review — PR #1757 (feat #1756)

**Verdict: APPROVE** 🟢

dj-navigate auto-syncs `aria-current="page"` across SPA navigation. Implementation is correct, well-scoped, defensively coded, and the test is non-tautological (gate-off verified). No 🔴, no 🟡.

---

### Mandatory checks

| Check | Result |
|---|---|
| STALE-BASE (#250) | 🟢 `BEHIND=0` — branch is up to date with `origin/main`; `origin/main..HEAD` = exactly the 2 PR commits |
| Two-commit shape (#181) | 🟢 c1 `4fd72de6` = src + bundle (client.js/.min/.br/.gz/.map) + test, NO CHANGELOG/docs; c2 `0a240454` = CHANGELOG + navigation.md only |
| Bundle integrity | 🟢 `client.js` contains `updateAriaCurrent` (5 refs); `make build-js` produced **no diff** — committed bundle matches src |
| No unguarded `console.log` | 🟢 none added in `src/18-navigation.js` |

---

### Logic correctness (`updateAriaCurrent`, src/18-navigation.js:456–473) — 🟢

- **Exact-path match** → `setAttribute('aria-current','page')` (L467–468). ✓
- **Don't-clobber** (L469–470): non-match removes the attr *only if* it currently equals `"page"` — an app-authored `aria-current="step"`/`"true"` is preserved. ✓ (test `dj_navigate_aria_current_1756.test.js:84` asserts this)
- **Cross-origin guard** (L461 `new URL(value, location.origin)` + L466 `dest.origin === location.origin`): a `dj-navigate="https://djustlive.com"` link (pathname `/`) is never current, even on `/`. ✓ (test L76–81)
- **Malformed-value guard**: `new URL(...)` in try/catch with `return` (L460–464) — skips just that element, loop continues. ✓
- **Idempotent**: pure attribute set/remove based on current URL; running twice = same result. ✓

### Call-site coverage — 🟢 (no stale-highlight gap)

The 3 wired sites are click (`handleLiveRedirect` L120, after pushState), mount (`bindNavigationDirectives` L445, via `reinitAfterDOMUpdate`), and `popstate` (L261). Traced every other nav path:

- **`reinitAfterDOMUpdate` chain confirmed**: `09-event-binding.js:1436 reinitAfterDOMUpdate → bindLiveViewEvents:1439 → navigation.bindDirectives:1113 → updateAriaCurrent`. Every VDOM patch / mount reaches it.
- **`skipMountHtml` (turbo-prerender) — COVERED.** `03-websocket.js:355` branch calls `reinitAfterDOMUpdate()` inside `runPostMount` (L408, deferred via rAF / sync fallback). Not a gap.
- **`live_patch` / `url_change` (same-view query nav, `_executePatch` L373) — COVERED.** `_executePatch` pushState's without an *immediate* `updateAriaCurrent`, BUT (a) for a same-path query-only change the active link doesn't change, and (b) the server reply (`patch`/`html_update`) routes `02-response-handler.js:159/187 reinitAfterDOMUpdate → updateAriaCurrent` after the round-trip. No real nav path leaves the highlight stale.

### Perf — 🟢 negligible

`updateAriaCurrent` runs on every `bindNavigationDirectives` (i.e. every `reinitAfterDOMUpdate`, every patch). Cost = one `querySelectorAll('[dj-navigate]')` + a `new URL()` per link. Nav links are few (single-digit to low-double-digit); the work is trivially small relative to the surrounding VDOM patch + hook remount that already runs in `reinitAfterDOMUpdate`. Not worth gating on URL-change. Stating as informational, not a concern.

### Gate-off (#1200 / #254) — 🟢 non-tautological

Added `return;` to the top of `updateAriaCurrent`, `make build-js`, re-ran: **3 of 4 tests FAILED** (only the "exposes function" smoke test passed). The 3 behavior tests genuinely exercise the change. Restored src + rebuilt — clean.

### No-regression — 🟢

- `dj_navigate_aria_current_1756.test.js`: 4/4 pass.
- `navigation.test.js` + `hooks.test.js`: 45/45 pass. Existing dj-navigate/dj-patch behavior intact.

### Test quality note (acceptable)

Tests call `updateAriaCurrent()` **directly**, not via a real click. The 3 wiring call-sites are therefore inspection-only (not exercised by an integration test). Acceptable for a 3-line wiring change where each site is a single function call adjacent to an existing pushState/reinit; the function-under-test itself is well covered. Flagging per #1196 transparency, not as a blocker.

---

johnrtipton:
## Retrospective — PR #1757 (#1756 part 1)

**Task**: make `dj-navigate` auto-manage `aria-current="page"` so a persistent nav tracks the current page across SPA navigation — the framework generalization of the djust.org `nav-active.js` workaround shipped this session.

### Quality: 5/5
Generalizes a real downstream fix into an accessible, zero-per-app-JS framework feature; gate-off-proven; all nav paths traced.

### What went well
- **Downstream workaround → upstream feature.** djust.org hit this live (Examples → page kept "Examples" highlighted); I fixed it there with a per-app script, then the *same session* lifted it into the framework as the standard `aria-current="page"` mechanism. Apps now style one attribute; no per-app JS. Same "lift-from-downstream then generalize" pattern as Action #1077.
- **Call-site coverage verified, not assumed.** Stage-7 traced every nav path (click, WS mount incl. the `skipMountHtml` turbo-prerender branch, `live_patch`, popstate) to confirm none leaves the highlight stale — the single hook is `reinitAfterDOMUpdate → bindNavigationDirectives → updateAriaCurrent`, which every patch/mount already goes through.
- **Correctness guards from the downstream lessons**: cross-origin targets never current (the djustlive.com `/`-pathname false-match I'd already hit in nav-active.js), and never clobbering an app-authored `aria-current` value.
- Gate-off (#254): neutering the function fails 3/4 — non-tautological.

### What could be improved
- Scoped to part 1 only. `document.title` sync needs the mount frame to carry the new title (a server/wire change), and the `[dj-root]` `dj-view`-attr update touches the mount handler — both deferred to keep this PR a clean single-file client change. Tracked in #1756.
- The 3 wiring call-sites are covered by inspection, not an integration test (the unit tests call `updateAriaCurrent()` directly). Acceptable for 3 one-liners; a real-click integration test would be stronger.

### Verified
- New test 4/4 (gate-off proven); navigation + hooks 45/45 green; bundle rebuilt from src (`make build-js` no-diff); CI green (js, nav-hooks-guard, python, rust, demo-checks). Two-commit shape intact.

`RETRO_COMPLETE`

PR #1759, opened 2026-06-11
fix(deploy): respect .gitignore in deploy tarball + warn on oversized uploads

johnrtipton:
## Problem

`djust deploy` (→ `deploy-dir` → `_create_tarball`) built a **152 MB** tarball for a real app and the server ingress rejected the upload with **413 Request Entity Too Large**. Root cause: `_create_tarball` walked the whole source tree minus a **hardcoded** `EXCLUDE_*` list, ignoring the user's `.gitignore`. Bloat with non-standard names slipped straight in:

- `.venv-dev/` — a second virtualenv (the matcher only knows `.venv`/`venv`), carrying Playwright's **115 MB** bundled `node` binary
- `scratch/` — screenshots, dylibs
- `mobile-app/wheels/` — locally-built android/ios wheels

All of these were **already gitignored**; the deploy just didn't consult it. `git archive HEAD` for the same project is 6.5 MB.

## Fix

1. **Respect `.gitignore`.** When the source is a git work tree, `_create_tarball` takes its file list from `git ls-files --cached --others --exclude-standard` (tracked ∪ untracked, minus ignored) — the user's `.gitignore` becomes the source of truth. Non-git directories fall back to the existing `os.walk` path, unchanged (the exclude predicate is now shared between both paths, so behavior is identical).
2. **Keep the security net.** `EXCLUDE_*` (`.env*`, `db.sqlite3*`, `.pyc`, …) is still applied to git-listed paths, so credentials / live DBs never ship even if the user forgot to ignore them (#1505 intent preserved).
3. **Warn before a wasted upload.** New `_tarball_size_warning` makes `deploy-dir` print a warning when the packed tarball exceeds ~50 MB, listing the largest included files so the user can gitignore build artifacts instead of hitting a raw 413.

## Verification

- 66/66 `test_deploy_cli.py` pass (6 new: gitignore-respect, credential security net, non-git fallback, size-warning both directions).
- Built the tarball against the real project tree: **152 MB → 4.4 MB**, with the Playwright binary / wheels / scratch gone and the largest member now a 0.37 MB doc.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- **`djust deploy` now respects `.gitignore` when building the deploy tarball, and warns before oversized uploads (#1759).** `_create_tarball` previously walked the source tree using a hardcoded `EXCLUDE_*` list, ignoring the project's `.gitignore`. Non-standard names — a `.venv-dev/` virtualenv carrying a 115 MB Playwright `node` binary, `scratch/` screenshots, `mobile-app/wheels/` — slipped straight in, producing a 152 MB tarball that the server's ingress rejected with a raw 413. Fix: when the source is a git work tree, `_create_tarball` takes its file list from `git ls-files --cached --others --exclude-standard` (tracked ∪ untracked, minus ignored), so the project's `.gitignore` is the source of truth. Non-git directories fall back to the existing `os.walk` path unchanged. The `EXCLUDE_*` security net (`EXCLUDE_DIR_NAMES`, `EXCLUDE_FILENAMES`, `EXCLUDE_FILE_SUFFIXES`, `EXCLUDE_FILENAME_STEMS`) still applies on the git path so credentials and live databases are dropped even if the user forgot to gitignore them (#1505 intent preserved). A new `_tarball_size_warning` prints an actionable warning to stderr before upload when the packed tarball exceeds 50 MB, listing the largest included files so the user can identify what to add to `.gitignore` instead of hitting a raw nginx 413 page. 6 new regression cases in `python/tests/test_deploy_cli.py`.
+
 ## [1.0.3] - 2026-06-07
 
 ### Added
--- a/python/djust/deploy_cli.py
+++ b/python/djust/deploy_cli.py
@@ -93,6 +93,13 @@ EXCLUDE_FILENAMES = frozenset({".DS_Store", "Thumbs.db"})
 # ``.environment`` is NOT a stem variant of ``.env`` and stays included.
 EXCLUDE_FILENAME_STEMS = frozenset({".env", "db.sqlite3"})
 
+# Above this packed size, ``deploy_dir`` warns before uploading and points at
+# the largest included files. A clean djust app tarball is a few MB; anything
+# this large is almost always build artifacts or local data that belongs in
+# .gitignore — and the djustlive ingress 413s uploads past its body-size cap,
+# so surfacing it here turns a raw nginx error into an actionable message.
+TARBALL_WARN_BYTES = 50 * 1024 * 1024
+
 
 # ---------------------------------------------------------------------------
 # Credential helpers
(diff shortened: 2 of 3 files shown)


johnrtipton (commented):
**APPROVE** (self-review — cannot formally approve own PR)

Branch is current with main (0 commits behind). All 66 tests pass (5.93s).

**Correctness — all checks pass:**

- `_is_excluded_relpath` splits on `/` and checks every dir segment — nested paths like `mobile-app/wheels/djust.whl` handled correctly; `*dir_parts, base = segments` unpacks to `dir_parts=["mobile-app", "wheels"]`, `base="djust.whl"`.
- `subprocess.CalledProcessError` and `subprocess.TimeoutExpired` are both subclasses of `subprocess.SubprocessError` — the single `except (OSError, subprocess.SubprocessError)` catches both. Verified.
- `threshold=0` correctly skips the `is None` branch and uses 0 (always warns). No footgun.

**Security net preserved:** `test_credentials_excluded_even_when_not_gitignored` directly exercises the double-gate — EXCLUDE_* fires even when `.gitignore` is empty. #1505 invariant intact.

**No auto-reject triggers:** no bare `print()`, no `except: pass`, no f-string logging.

**CHANGELOG:** format correct — bold title, `(#1759)` ref, test count ("6 new regression cases") matches the 6 added methods exactly.

Minor: f-strings in `_tarball_size_warning`'s `lines.append(...)` are correct — user-facing output, not logger calls. Clean.

johnrtipton:
## Retrospective — PR #1759

**Task:** fix(deploy): respect `.gitignore` in deploy tarball + warn on oversized uploads

### Quality: 5/5

### What went well
- **The fix addressed the right root cause.** The hardcoded `EXCLUDE_*` list was a closed-world assumption — it only knew the names the author thought of at PR #405 time. Using `git ls-files` makes `.gitignore` the source of truth, which is the correct level of abstraction. Verified on a real project tree: 152 MB → 4.4 MB.
- **Security net preserved and tested.** `EXCLUDE_*` still fires as a last-resort filter over git-listed paths, so credentials and live DBs never ship even if the user forgot to ignore them. `test_credentials_excluded_even_when_not_gitignored` directly exercises the invariant — the #1505 security contract is intact.
- **The non-git fallback is a clean no-regression.** The `os.walk` path for non-git directories shares the same `_is_excluded_relpath` predicate now, so the two paths have identical exclusion logic rather than independent copies that could drift.
- **`_tarball_size_warning` is actionable, not just informational.** The warning lists the largest included files, giving the user the information they need to fix their `.gitignore` rather than just a raw size number.
- **6 regression tests, clean gate-off implicit.** `test_gitignored_paths_excluded` asserts specific paths absent — because `.venv-dev` is not in `EXCLUDE_DIR_NAMES`, the test would fail if the `_git_tracked_files` path silently fell back to `os.walk`. The change cannot be gated off without the test catching it.
- **CI clean on first submission.** All 11 checks passed, all 66 deploy tests pass. No fix-pass cycle needed.

### What could improve
- **Gate-the-change-off self-test (Action #254) was not explicitly run.** The implicit coverage held up, but the implementer should have stated "I ran the test with `_git_tracked_files` returning `None` and confirmed `test_gitignored_paths_excluded` fails" as part of Stage 5 verification. The result was correct; the discipline was implicit rather than deliberate.
- **`Optional[list]` type hint not updated to `list | None`.** The self-review noted `_tarball_size_warning(path, threshold: Optional[list] = None)` uses the older unparameterized `Optional` form. Minor, but it's in a function introduced in this PR — a clean authoring pass would have used `list[str] | None` throughout. Logged, not blocking.
- **One test only exercises the silent path.** `test_no_warning_below_threshold` confirms the warning is suppressed but doesn't assert anything about the returned value or the call-site's behavior. The gap is narrow (the other two warning tests cover emission), but a symmetry assertion would have tightened the suite.

### Relation to prior work
This PR is the third deploy tooling fix in the v1.0.3 window, after #1750 (close-side `</div>` tolerance) and the CI composite action refactor (#1752). The tarball bloat bug predates those — it traces to the original `djust deploy` implementation (PR #405) — but was surfaced by users hitting 413 errors after the deploy path was hardened in the v1.0.x cycle. The `EXCLUDE_*` security net itself was introduced as part of #1505's credential-leak guard; this PR extends rather than replaces that work, which is the right layering. The broader theme of the v1.0.x drain has been "fix correctness gaps in deploy tooling that only show up in real projects" — this is the most impactful of the three by raw impact (152 MB → 4.4 MB is the kind of fix that unblocks users who couldn't deploy at all).

### Verified
- `git ls-files` path honors `.gitignore` and security net; 66 deploy tests pass in 5.93s; CHANGELOG entry with `(#1759)` reference; stale-base check clean (0 commits behind main at merge time).

RETRO_COMPLETE

Issue #1761, opened 2026-06-12
feat(cli): consume djustlive serving_current to print 'rolling out' vs 'active'


johnrtipton:
Source: djustlive Retro v0.7.1 (PR #525 / djustlive #517).

djustlive's deployment_status endpoint now returns an additive `serving_current` boolean (False while a new rootfs is built+current but the old placement is still serving during blue/green). The `djust deploy` CLI should read it and print "rolling out" instead of "active"/the URL until it's True, so users stop re-testing against stale rootfs. Thin client-side change; the authoritative server signal already shipped in djustlive #517.

johnrtipton:
Implemented in PR #1767 (merged `d1b42385`, milestone v1.0.4-1). deploy CLI now prints "rolling out" vs "active" by consuming djustlive serving_current (deployment_status endpoint, djustlive #517); missing field fail-safe True for back-compat.

Issue #1760, opened 2026-06-11
djust deploy: preflight 'doctor' warning for settings that violate the platform env contract


johnrtipton:
## Problem
`djust deploy` happily ships an app whose Django settings can't work on the platform, with no warning. A "foreign" (non-scaffold) app — e.g. one written for on-device/loopback use — silently violates the platform's env-injection contract and only fails *at runtime in production*:

- hardcoded `SECRET_KEY` (dev key served on a public host)
- `ALLOWED_HOSTS` not read from env → `DisallowedHost` 400
- `DATABASES` not read from `DATABASE_URL` → on a **read-only** app rootfs this 500s on the first write (`OperationalError: attempt to write a readonly database`)
- sqlite `NAME` on a read-only path

This exact chain cost a multi-hour prod debug for the `max-companion` deploy (2026-06-11): the app hardcoded sqlite + `SECRET_KEY` + `ALLOWED_HOSTS`, ignored the injected `DATABASE_URL`, and 500'd silently after a "successful" deploy.

## Proposed fix
A preflight check (a "deploy doctor") run by `djust deploy` that statically inspects the resolved settings module and **warns** when:
- `SECRET_KEY` is a literal (not `os.environ`-derived)
- `ALLOWED_HOSTS` is a literal not reading env
- `DATABASES` doesn't consult `DATABASE_URL`
- a sqlite `NAME` resolves under the project dir (read-only on the platform)

Warnings (not hard errors) with a one-line "the platform injects these — read them from env" pointer. Most of these are grep-level static checks.

## Context
Complements djust-org/djust#1759 (the `.gitignore`-respecting tarball + size warning). The platform-side counterparts are djust-org/djustlive#515 (active-flip should require a real render) and #518 (detect DB on read-only rootfs).

johnrtipton:
Implemented in PR #1767 (merged `d1b42385`, milestone v1.0.4-1). deploy doctor preflight: non-blocking static settings check warning on SECRET_KEY/ALLOWED_HOSTS/DATABASES/sqlite env-contract violations. Known-narrow DATABASES heuristic follow-up: #1768.

PR #1767, opened 2026-06-13
feat(deploy): deploy doctor preflight + serving_current rollout status (#1760, #1761)

johnrtipton:
## Summary

Two small, drain-mechanical `djust deploy` CLI improvements in the deploy-DX theme (same family as the just-shipped #1759). Milestone **v1.0.4-1**. Both consume contracts the platform side already provides; client-side changes only.

### #1760 — deploy "doctor" preflight
`djust deploy` / `djust deploy-dir` now run a **non-blocking** static preflight over the resolved Django settings module and WARN (stderr, never error — deploy still proceeds) when settings violate the platform env-injection contract:
- hardcoded literal `SECRET_KEY`
- `ALLOWED_HOSTS` literal not read from env (→ `DisallowedHost` 400)
- `DATABASES` block that never consults `DATABASE_URL` (→ `OperationalError: readonly database` on the read-only rootfs)
- sqlite `ENGINE` (sqlite under the read-only project dir 500s on first write)

Checks are grep-level static inspection of the settings **source text** — the module is never imported (an unimportable foreign settings file can't break the doctor), and the whole pass is wrapped fail-soft. Settings located via `manage.py`'s `DJANGO_SETTINGS_MODULE` default, falling back to the shallowest `settings.py`. Only the primary module is inspected (avoids dev-settings false positives). Prevents the silent "successful deploy, 500s at runtime" chain (cost a multi-hour `max-companion` prod debug 2026-06-11).

### #1761 — "rolling out" vs "active"
Consumes djustlive's additive `serving_current` boolean (`deployment_status` endpoint = `GET /api/v1/deployments/{id}/`, djustlive #517). While the deployment row reads `active`/`deploying` but `serving_current` is `False` (blue/green window — old placement still serving the env URL), the `deploy-dir` poll loop prints `Status: rolling out (...)` and **keeps polling** instead of reporting `active`/the URL on stale code. Missing field → `True` (fail-safe, byte-identical against older servers). Poll decision extracted to a pure `_poll_display(data) -> (message, done, url)` helper.

| Issue | Files | Tests |
|---|---|---|
| #1760 deploy doctor | `python/djust/deploy_cli.py` (`_deploy_doctor_warnings`, `_find_settings_files`, `_run_deploy_doctor`; hooked in `deploy`/`deploy_dir`) | `TestDeployDoctor` (14 cases) |
| #1761 rollout status | `python/djust/deploy_cli.py` (`_poll_display`; `deploy_dir` poll rewire) | `TestPollDisplay` (8 cases) |

## Verification
- 22 new tests; full `test_deploy_cli.py` module: **88 passed**.
- Gate-off self-test (Action #254): gating off the `serving_current` branch fails 2 `TestPollDisplay` cases; restored.
- ruff clean on changed files; two-commit shape (impl+tests / docs+CHANGELOG); Gate 1 + Gate 2 pass.

## API contract verified (Stage 4)
`deployment_status` (djustlive `deployments/api.py:353`) maps to the exact URL the poll loop hits and includes `serving_current` (api.py:386); semantics are fail-safe-`True`-when-unknown (`_deployment_serving_current`, api.py:304).


johnrtipton (commented):
APPROVE — clean, well-scoped deploy-CLI feature. Both deliverables present and correct.

**#1760 deploy doctor**: `_deploy_doctor_warnings` / `_find_settings_files` / `_run_deploy_doctor` (deploy_cli.py:1247-1353), hooked in both `deploy` (line 1405) and `deploy_dir` (line 1477). Security verified: the doctor only `re.search`es `settings_text` and `read_text`s files — never imports/execs the settings module, so an unimportable foreign settings file cannot break it. The `except Exception` at line 1351 is fail-soft *with* `logger.debug(..., exc_info=True)` — not a bare `except:pass`. Compliant.

**#1761 rollout status**: `_poll_display` (line 1213) consumed in the `deploy_dir` poll loop (line 1537). `serving_current` fail-safe-True back-compat (`data.get("serving_current", True)`, line 1227) is correct and tested. Confirmed `deploy` is correctly NOT rewired — it streams server lines via `iter_lines()` (line 1422), no client-side poll.

**Verification**:
- `pytest python/tests/test_deploy_cli.py` -> **88 passed**.
- Gate-off self-test (#254): gating the `not serving_current` branch (line 1228) to `False` fails 2 TestPollDisplay tests — the tests genuinely exercise the change, not tautological.
- CHANGELOG counts match source: TestPollDisplay=8, TestDeployDoctor=14.

**One concern (advisory, non-blocking)**: the DATABASES heuristic (line 1280) only matches the literal tokens `DATABASE_URL`/`dj_database_url`. A settings module that builds `DATABASES` from individual `os.environ["DB_HOST"]`/`DB_NAME` vars (a documented Django pattern) gets a **false-positive** warning. Since warnings never block and the plan explicitly accepts false-positive risk (warnings-only, primary-module scope), this is fine to ship — but consider also treating an `os.environ` reference inside the DATABASES region as env-driven in a follow-up. Minor false-negative twin: `SECRET_KEY = config("...")` (python-decouple) is silently OK (good outcome, but the literal check is quote-only).

Nice touches: stderr-only output, fail-soft wrapper, `_DOCTOR_SKIP_DIRS` to avoid scanning venvs.

johnrtipton:
## Retrospective — PR #1767 (pipeline-drain, v1.0.4-1)

**Task**: v1.0.4-1 deploy-DX drain — grouped #1760 (deploy doctor preflight) + #1761 (`serving_current` rollout status). Both small, same file (`python/djust/deploy_cli.py`), same theme as the just-shipped #1759.

### Quality: 5

### What went well
- **Stage-4 API-contract verification paid off (#1243).** The poll loop hits `/api/v1/deployments/{id}/`, but the issue cited the "`deployment_status` endpoint." Grepping the djustlive sibling repo confirmed they're the *same* URL (`deployments/api.py:353`) and that the response body carries `serving_current` (api.py:386) with fail-safe-`True`-when-unknown semantics (`_deployment_serving_current`, api.py:304). That locked the `data.get("serving_current", True)` back-compat shape before a line of code — no guesswork, no rework.
- **Extracting `_poll_display` as a pure helper** made the #1761 decision unit-testable without `requests_mock`/`sleep` plumbing, and made the gate-off self-test (#254) trivial: gating the `not serving_current` branch off fails exactly 2 `TestPollDisplay` cases — proving they're non-tautological.
- **Static-text doctor over import-the-settings.** Inspecting the settings *source text* with regex (never importing the module) means an unimportable foreign settings file — the exact "non-scaffold app" this targets — can't break the doctor. Fail-soft wrapper guarantees it never blocks a deploy.
- Two-commit shape held cleanly (impl+tests / docs+CHANGELOG); Gate 1 + Gate 2 + CHANGELOG-count hook all green; one pre-commit ruff-format bounce, re-staged per canon.

### What didn't / lessons
- **Doctor DATABASES check has a known false-positive** (Stage-11 🟡): a `DATABASES` built from individual `os.environ["DB_NAME"]`/`["DB_HOST"]` vars (not `DATABASE_URL`/`dj_database_url`) still warns, because the env-read detection is token-scoped to those two names. Advisory-only (warnings never block), so accepted for this PR per the plan's stated risk; filed as a follow-up to broaden the env-read heuristic. This is the classic "warning heuristic narrower than the real contract" gap — worth widening before the doctor graduates from advisory to anything louder.
- The `_find_settings_files` glob fallback walks the whole tree (rglob) when there's no `manage.py`; the manage.py fast-path covers the common Django case, so the cost is bounded, but a fat-`.venv` no-manage.py project would pay a slow preflight. Noted, not fixed (rare path, advisory tool).

### Relation to prior milestone work
- Same deploy-DX theme as #1759 (the `.gitignore` tarball + size-warning fix) — the doctor and the tarball fix are complementary preflight/packaging guards for the same "foreign app deploy" failure mode that cost the multi-hour `max-companion` prod debug (2026-06-11).
- Reused the #254 gate-off self-test and #1243 verify-literal-API-contracts disciplines that have been load-bearing across recent drains.

### Verified
- 22 new tests; full `python/tests/test_deploy_cli.py` → **88 passed**.
- ruff clean on changed files; gate-off self-test confirms non-tautological tests.
- All CI green (`test-summary` aggregate gate passed: rust-tests, python-tests py3.12 + py3.14t, javascript-tests, demo-checks, nav-hooks-guard, playwright-tests, security-tests).
- Merged squash `d1b42385`; branch deleted.

RETRO_COMPLETE

PR #1773, opened 2026-06-13
fix(deploy): deploy doctor honors env-derived DATABASES, not just DATABASE_URL (#1768)

### LINKED ISSUE #1768: tech-debt: deploy doctor DATABASES check false-positives on individual os.environ DB vars (#1760 follow-up)
From PR #1767 (#1760) Stage-11 code review.

`_deploy_doctor_warnings` in \`python/djust/deploy_cli.py\` flags a \`DATABASES\` block as "doesn't consult DATABASE_URL" whenever the settings text contains neither \`DATABASE_URL\` nor \`dj_database_url\`. But a perfectly env-driven config built from **individual** vars — e.g. \`{'NAME': os.environ['DB_NAME'], 'HOST': os.environ['DB_HOST'], ...}\` — has no \`DATABASE_URL\` token and so gets a **false-positive** warning.

This is advisory-only (warnings never block a deploy), so it was accepted for #1767 per the plan's stated risk. Worth widening before the doctor ever graduates from advisory.

**Proposed fix**: broaden the env-read detection for the DATABASES check — e.g. suppress the warning when the DATABASES assignment region references \`os.environ\` / \`os.getenv\` / a settings env helper, not only the two literal tokens. The hard part is scoping "the DATABASES region" so a \`SECRET_KEY = os.environ[...]\` elsewhere doesn't suppress the DB warning; a brace-balanced slice of the \`DATABASES = { ... }\` block is the likely approach. Add a positive test (individual \`os.environ\` DB vars → no warning) and keep the existing hardcoded-postgres positive case.

Milestone: v1.1.0 (or a future deploy-DX drain bucket).

johnrtipton:
## Summary

Fixes #1768 (Stage-11 follow-up to #1760's deploy doctor).

The #1760 `DATABASES` preflight check warned whenever the settings source mentioned neither `DATABASE_URL` nor `dj_database_url`. But a perfectly environment-driven config assembled from **individual** `os.environ['DB_NAME']` / `['DB_HOST']` / … vars has no `DATABASE_URL` token, so it got a spurious "doesn't consult `DATABASE_URL`" warning.

## Fix

Widen the env-read detection in `_deploy_doctor_warnings`:
- The canonical `DATABASE_URL` / `dj_database_url` tokens still count **anywhere**.
- Other env reads (`os.environ`, `os.getenv`, `getenv(`, `environ[`, `environ.get(`, `env(`, `config(`, `decouple`) now count **within the `DATABASES` assignment region**.
- New `_databases_region()` brace-balances the `DATABASES = { … }` dict literal (or takes the call-form line) so the region-scoped check can't be masked by an unrelated `SECRET_KEY = os.environ[...]` elsewhere in the file.

A hardcoded `DATABASES` block (literal `NAME`/`USER`/`HOST`, no env) still warns — correctly.

## Verification
- `TestDeployDoctor`: **16 passed** (+2): `test_databases_from_individual_env_vars_does_not_warn` (the fix) and `test_databases_hardcoded_still_warns_despite_env_secret_key` (the region-scoping guard).
- Full `test_deploy_cli.py`: **90 passed**.
- Gate-off self-test (#254): reverting the region-scoped env check fails the individual-env-vars test; restored.
- ruff clean; two-commit shape (impl+tests / CHANGELOG); Gate 1 + Gate 2 pass.

Closes #1768.


johnrtipton:
## Retrospective — PR #1773 (#1768 deploy-doctor heuristic)

**Task**: Stop the #1760 deploy-doctor DATABASES check from false-positiving on env-derived configs built from individual `os.environ['DB_*']` vars.

### Quality: 5
### What went well
- The fix widened the env-read signal AND scoped it correctly: a new brace-balancing `_databases_region()` confines the `os.environ` check to the DATABASES block, so an unrelated `SECRET_KEY=os.environ[...]` can't mask a hardcoded DB block. Region-scoping was the load-bearing insight (a naive "os.environ anywhere in the file" check would have under-warned).
- Two regression tests pin both directions (env-vars → no warn; hardcoded-DB-with-env-SECRET_KEY → still warns); gate-off (#254) verified the false-positive test fails without the fix.
### What didn't
- First commit bounced on ruff-format reflow (caught by the `git log -1` reflex); re-ran ruff + re-staged.
### Verified
- TestDeployDoctor 16 passed (+2); full test_deploy_cli.py 90 passed; ruff clean; Gate 1 + Gate 2; CI green.

RETRO_COMPLETE

PR #1775, opened 2026-06-13
fix(routing): auth-filter the client route map (#1758, ADR-021 Stage 2)

### LINKED ISSUE #1758: Route map and dj-navigation
### Problem / Motivation

Consider what route map exposes to the client, maybe move logic into template render instead.

### Proposed Solution

_No response_

### Alternatives Considered

_No response_

johnrtipton:
## Summary

Fixes #1758. **You flagged the data-leak — confirmed real.** The auto-emitted client route map (`window.djust._routeMap`) was built by `build_route_map_from_urlconf` with **no auth filtering** and emitted to **every** client including anonymous visitors. It enumerated **all** LiveView routes — `login_required` / `permission_required` / admin ones included — each with its dotted `module.QualName` view-class path (`_walk_liveview_routes` even unwraps `login_required(as_view())`).

So an anonymous visitor to a public page learned the **full route table** + the **internal view-class names** of routes they cannot access. This is recon-grade **information disclosure** (not an auth bypass — the WS mount path allowlists modules at `websocket.py:1726` and views still enforce auth at mount). My earlier "already in per-page `dj-view`" reasoning only covered the *current* page; it did not cover routes the user has never visited / can't reach.

## Fix

`get_route_map_script(request)` — the single funnel both template engines use via `_client_config_html` — auth-filters the emitted map:
- A **gating sidecar** (`_route_gating_cache`), populated by the *same* single URLconf walk (no extra cost), records each gated route's requirement: `login_required` / `permission_required` class attrs, or a `login_required(as_view())` decorator wrap (detected via `cb.__wrapped__` carrying `view_class` — reliable vs a bare `as_view()` whose `__wrapped__` is `cls.dispatch`).
- Routes whose requirement `request.user` doesn't satisfy are **omitted**; the filter **fails closed** for anonymous / no-request callers. Public routes (absent from the sidecar) always emit → existing public apps get a byte-identical map.

## Relationship to #1734 (you asked: "do we still need this with auto_navigate?")

Largely no, eventually. The deeper fix is to have the **client send the URL path** and the **server resolve path→view** authoritatively (removing client-sent view paths entirely). That's a wire-protocol change with regression surface across existing `dj-navigate`, so it's documented in ADR-021 as a possible **Stage 3** — this PR is the contained Stage-2 hardening that fixes the leak now without the protocol change. `auto_navigate` (#1734) will ride on the hardened map.

## Verification
- 5 new `TestRouteMapAuthFilter` cases (anonymous → public only; authed → login-gated but not perm-gated; permitted → all; no-request fails closed; public view-path preserved) + a fixture URLconf covering all gating forms.
- Gate-off self-test (#254): disabling the filter fails 3 of the 5; restored.
- Broad sweep (route/nav/checks/client_config/session): **698 passed**, no regression. ruff clean. ADR-021 amended (corrects the "URLs are public, not sensitive" risk note). Two-commit shape; Gate 1 + Gate 2.

Closes #1758.


johnrtipton (commented):
APPROVE — security fix is correct and well-tested.

**Single-funnel verified.** Both engines route through `get_route_map_script(request)`: Django via `live_tags.py:171`, Rust via `client_config.py:68` → `_client_config_html` → same call (request passed for nonce + filter). No other client-side emitter of `_routeMap` exists. `checks.py:3275` uses the unfiltered `build_route_map_from_urlconf()` but that is a server-side system check, never sent to a client — correct.

**Fail-closed confirmed.** `_user_satisfies` (routing.py:262) returns False for `request=None`, `user=None`, and `is_authenticated=False`. Verified by `test_no_request_fails_closed` and the anonymous test.

**Decorator detection (routing.py:248-253) holds empirically.** I probed it: bare `as_view()` → `__wrapped__` is falsy / no `view_class` → not gated (no false-trigger); `login_required(as_view())` → `__wrapped__.view_class` set → gated. (Minor: the comment says a bare `as_view()` `__wrapped__` is `cls.dispatch`; on Django 5.2 it is actually `None`. Guard still holds either way — non-blocking comment nit.)

**Cache correctness.** `_route_map_cache` + `_route_gating_cache` share an identical key `(resolved_urlconf, script_prefix)` (routing.py:372/397), are populated by the one walk, and both cleared by `_reset_route_map_cache`. Cache stores the unfiltered map + gating; the per-request user filter runs only in `get_route_map_script` — no cross-request/cross-tenant leak.

**Gate-off self-test (#1200/#254): passes.** Short-circuiting `_user_satisfies`→True fails exactly 3/5 new tests (anonymous, authed-no-perm, no-request); the 2 survivors (all-perms user, public-route-path) are correctly filter-insensitive. Tests are not tautological. 13/13 pass (5 + 8 derivation).

🟡 **One documented-surface gap (non-blocking):** Django stdlib `LoginRequiredMixin`/`PermissionRequiredMixin` (`django.contrib.auth.mixins`) and the `check_permissions()` custom hook set no class attr → undetected → would emit to anonymous. djust ships its OWN mixins (`auth/core.py:289/303`) which set `login_required=True` (detected, verified), and djust mount-auth only enforces the class attrs anyway, so this is narrow. Worth a one-line docstring note on `_view_auth_requirement` that stdlib-mixin / custom-hook gating is not introspected.

Public routes still emit (incl view path) — confirmed. APPROVE.

johnrtipton:
## Retrospective — PR #1775 (#1758 route-map auth filter)

**Task**: Auth-filter the auto-emitted client route map so anonymous/unauthorized clients can't enumerate login/permission-gated routes or their view-class paths.

### Quality: 5
### What went well
- The user's instinct ("should we worry about leaking data?") overrode my initial "by design, not sensitive" call — and was right. Re-investigating from the symptom (not my hypothesis) found the real disclosure: `_walk_liveview_routes` walks the whole URLconf with zero auth filtering AND unwraps `login_required`, so admin/login-gated routes + their `module.QualName` shipped to anonymous clients. Lesson: trust the user's leak instinct; verify from the data, not the prior conclusion.
- Fixed at the single funnel (`get_route_map_script` — both engines via `_client_config_html`), fail-closed, with a gating sidecar built by the SAME walk (no extra cost). Public routes emit identically (no regression for existing apps).
- Gating detection iterated to correctness: the first `__wrapped__`-without-`view_class` heuristic was wrong (functools.wraps copies `view_class`; bare as_view also has `__wrapped__`); the reliable signal is `cb.__wrapped__` itself carrying `view_class`. Stage-11 then caught a `LoginRequiredMixin` gap (no class attr) → hardened with an MRO-name check.
### What didn't
- My Stage-7 self-review under-rated the disclosure (the "already in per-page dj-view" argument only covered the current page). The reviewer + the user's push corrected it. Generalizes the "trust the symptom, not the cited path" rule to "trust the user's security instinct over your own first triage."
### Verified
- 13 tests pass (5 auth-filter + 8 derivation); gate-off (#254) fails 3/5 without the filter; broad route/nav/checks/session sweep 698 passed; ruff clean; CI green. ADR-021 amended (Stage 2 + corrected risk note).

RETRO_COMPLETE

Issue #1735, opened 2026-06-06
docs(nav): reconcile native dj-navigate (canonical) vs external TurboNav (interop)


johnrtipton:
Source: /pipeline-strategy 2026-06-05 (auto-navigation), Path 2 — Cluster C. Pairs with ADR-021.

**Problem.** djust ships two navigation stories: native `dj-navigate`/`live_redirect` (SPA over the EXISTING WebSocket — light, preserves the connection) and `docs/guides/turbonav-integration.md` (external turbo.js, AJAX + `<main>` swap, WS **teardown/reconnect per nav** — heavier). They're presented without a clear canonical recommendation, so users (and AIs) can't tell which to reach for.

**Fix.** Position native `dj-navigate` (+ auto_navigate from #B) as the **canonical** SPA-nav story; reframe `turbonav-integration.md` as "interop: djust also works under external Turbo/TurboNav" with the tradeoff stated (per-nav WS reconnect vs native over-the-wire patch). Cross-link from navigation.md. No code change — docs + stance, tracked by ADR-021.

Target: **1.1.0** (with #B).

johnrtipton:
Shipped in PR #1776 (merged): docs/website/guides/navigation.md documents auto_navigate; docs/guides/turbonav-integration.md reframed as interop with native dj-navigate/auto_navigate positioned as canonical (ADR-021 Stage 2).

PR #1776, opened 2026-06-13
feat(nav): opt-in auto_navigate link interception (#1734, #1735, ADR-021 Stage 2)

### LINKED ISSUE #1734: feat(nav): auto_navigate — Turbo-Drive <a> interception (opt-in flag)

johnrtipton:
## Summary

Closes #1734 and #1735 (ADR-021 Stage 2). Adds opt-in **`auto_navigate`** — Turbo-Drive-style automatic SPA navigation of plain `<a href>` links — built on the now auth-filtered route map (#1758).

- **Config**: `LIVEVIEW_CONFIG['auto_navigate']` (**default OFF**). When on, `{% djust_client_config %}` emits `<meta name="djust-auto-navigate" content="1">` (CSP-clean — a `<meta>`, not an inline script) and the client installs **one** delegated `click` listener on `document` (survives SPA mounts; no per-element binding).
- **Interception**: a plain `<a href>` is SPA-navigated only when its path resolves in the route map. Same-view query-only → `url_change`/`live_patch` (state-preserving); cross-view → `live_redirect`.
- **Opt-out matrix** (falls through to a normal browser navigation): modifier/middle click; `target` ≠ `_self`; `download`; `rel="external"`; `data-no-navigate` (link or ancestor); external origin or non-http(s) scheme; same-page hash-only jump; **path not in the route map**. Because the map is auth-filtered (#1758), `auto_navigate` never intercepts a route the user can't access — it full-reloads and the server enforces access.
- **#1735 docs**: `navigation.md` documents `auto_navigate`; `turbonav-integration.md` reframed as **interop** (native nav canonical; the per-nav WS-reconnect tradeoff stated).

## Composition with the leak fix
This is why the #1758 hardening landed first: `auto_navigate` consults the same auth-filtered map, so it can't be used to SPA-probe gated routes. The two are designed together (ADR-021 Stage 2).

## Verification
- **JS**: `tests/js/auto_navigate_1734.test.js` — 15 cases: full `_shouldSkipAutoNavigate` opt-out matrix; `_handleAutoNavigateClick` branches (cross-view → `live_redirect_mount`; same-view → `url_change`; external/unknown-route/hash/mailto → fall-through); **meta-gating opt-in pair** (off → no install/no interception; on → real delegated click intercepts) which doubles as the gate-off self-test. Nav regression suite green (40 passed).
- **Python**: `TestClientConfig` — meta absent by default; present + **dual-engine identical** when enabled.
- ruff + eslint clean; `build client.js from src` + doc-snippet hooks pass; two-commit shape (Gate 1 + Gate 2).


johnrtipton:
## Retrospective — PR #1776 (#1734 auto_navigate + #1735 nav docs)

**Task**: opt-in Turbo-Drive-style link interception on the auth-filtered route map (ADR-021 Stage 2).

### Quality: 5
### What went well
- Composes with #1758: auto_navigate only intercepts routes in the auth-filtered map, so it can't SPA-probe gated routes — the leak fix landing first was the right sequence.
- CSP-clean opt-in (a `<meta>`, not an inline script); one delegated listener (survives SPA mounts); the opt-out matrix is unit-pinned (15 JS cases) with a meta-gating pair that doubles as the gate-off self-test.
- Same-view used `url_change` (server-notifying, replace semantics) not `handleLivePatch` (history-only) — caught that distinction before shipping.
### What didn't
- Forgot the State-File Gate at branch creation (silent branch-verify fail on first commit); created the state file then proceeded. Reflex should fire at `checkout -b`, not first commit.
### Verified
- 15 JS cases + dual-engine Python emit tests; nav regression green; ruff/eslint/build/doc-snippet hooks pass; CI green.

RETRO_COMPLETE

Issue #1779, opened 2026-06-13
docs: {% csrf_token %}/{{ user }} inside [dj-view] are deleted on WS patches (threat model T9)


johnrtipton:
From the WS auth threat model (docs/audits/websocket-auth-2026-06.md, T9).

The Rust engine re-renders the \`[dj-view]\` subtree on every WS patch from view-instance attributes only — context processors and request-bound tags don't run, so \`{{ user }}\` / \`{% csrf_token %}\` render empty and the DOM diff **deletes** anything built from them (a logout form, a CSRF'd inline form vanish on the first update).

**Mitigation (docs):** place such auth/CSRF UI **outside** \`[dj-root]\` (HTTP-rendered once, never WS-patched). The user has draft BEST_PRACTICES content for this; formalize it.

Issue #1778, opened 2026-06-13
docs(security): LIVEVIEW_ALLOWED_MODULES is default-open — recommend setting it (threat model T4)


johnrtipton:
From the WS auth threat model (docs/audits/websocket-auth-2026-06.md, T4).

The \`live_redirect_mount\` module allowlist (\`websocket.py:1726\`) is only enforced \`if allowed_modules:\` — when \`LIVEVIEW_ALLOWED_MODULES\` is unset (the default) the check is skipped, so a client can request mounting any \`LiveView\` dotted path. The mounted view's own \`check_view_auth\` still runs, so this is not an auth bypass, but it's a hardening gap (route/class enumeration).

**Proposed:** document \`LIVEVIEW_ALLOWED_MODULES\` as recommended hardening in the security guide; consider a \`djust check\` warning when it's unset in production.

Issue #1777, opened 2026-06-13
security: handle_event does not re-check auth (mid-session logout/perm-loss; threat model T3)


johnrtipton:
From the WS auth threat model (docs/audits/websocket-auth-2026-06.md, T3).

`handle_event` (\`python/djust/websocket.py\`) never re-runs \`check_view_auth\`; auth is **mount-time only**. A user who logs out or loses a permission **mid-session** keeps dispatching events on the still-open socket until they reconnect.

The T1/T2 fix (close socket on mount redirect) closes the anonymous-bypass; this is the remaining **defense-in-depth** gap for *authenticated-then-deauthorized* sessions.

**Proposed:** a lightweight re-check in \`handle_event\` for views declaring \`login_required\`/\`permission_required\` (use \`check_view_auth_lightweight\`, \`auth/core.py:241\`); close 4403 on failure. Gate behind the view declaring auth so public views pay nothing.

PR #1780, opened 2026-06-13
fix(websocket): close socket on auth/hook mount redirects — auth bypass (threat model T1/T2)

johnrtipton:
## Summary

Fixes a **confirmed WebSocket auth bypass** (you flagged it; I verified it end-to-end). Built threat-model-first per your request.

`handle_mount` sent a `{"type":"navigate"}` redirect frame on the `login_required` (and `on_mount`-hook) failure branches but **did not close the socket** — only the `PermissionDenied` branch closed it (`4403`). The view never mounted yet `view_instance` stayed set, and `handle_event` **never re-checks auth** (verified). So a **raw** WS client that ignores the navigate frame could send `{"type":"event"}` and reach `@event_handler` methods **unauthenticated** — a full auth bypass on the live mutation path (a browser obeys the redirect and hides it). Reachable via both the initial mount and `handle_live_redirect_mount` (delegates to `handle_mount` → one fix point).

## Fix
Both mount-time redirect branches now send the navigate frame, then `await self.close(code=4403)` and clear `view_instance`, mirroring the `PermissionDenied/4403` branch. Public/authorized mounts unchanged.

## Reproducer-first (empirical)
`python/djust/tests/test_ws_auth_close_socket.py` — a real `WebsocketCommunicator` with an anonymous scope:
- **Pre-fix (RED):** after the redirect, `do_mutation()` **executed** for the anonymous client (the test crashed *inside* the handler body — proof it ran).
- **Post-fix (GREEN):** the socket is closed (`4403`) and the event never dispatches (`_HANDLER_RAN is False`).

The RED→GREEN transition is the gate-off proof (#254).

## Threat model (the artifact you asked for first)
`docs/audits/websocket-auth-2026-06.md` — the **complete** WS auth/transport model: 9 threats across 5 trust boundaries, each `file:line`-grounded, with a prioritized plan. This PR ships **T1 + T2**. Follow-ups filed:
- **#1777 (T3)** — `handle_event` doesn't re-check auth mid-session (defense-in-depth).
- **#1778 (T4)** — `LIVEVIEW_ALLOWED_MODULES` is default-open (docs/hardening).
- **#1779 (T9)** — `{% csrf_token %}`/`{{ user }}` deleted on WS patches (docs).
- T8 (replay-viewer XSS) tracked in #1562.

## Verification
- Reproducer 2/2 pass; broad WS/auth/mount/hook regression sweep **280 passed**; ruff clean; two-commit shape (Gate 1 + Gate 2).


johnrtipton (commented):
Adversarial review — **COMMENT** (core fix is correct; one fix-induced functional regression to flag before merge).

## Security correctness — the T1/T2 fix is sound ✅
- **Both** redirect branches close + clear: `check_view_auth` redirect at `websocket.py:1994-1995` and the `on_mount` hook redirect at `websocket.py:2021-2022`, each mirroring the `PermissionDenied`/4403 branch at `websocket.py:1977`. No third navigate-without-close path in the mount flow — the only other navigate sends (`websocket.py:2593`, `:2694`) are the `mount_batch` collector/response, not a direct mount return.
- **Backstop verified**: `handle_event` short-circuits on `if not self.view_instance:` (`websocket.py:2738`), so clearing `view_instance = None` neutralizes any in-flight event even before `close()` lands.
- **Happy path untouched**: close fires only on the two redirect branches; authorized mounts fall through to `_ensure_tenant`/mount unchanged.
- **`live_redirect` covered**: `handle_live_redirect_mount` (`websocket.py:4335`) delegates to `handle_mount`, so it inherits the fix — no separate redirect-without-close there.
- **Reproducer is genuine, not tautological**: gated the close off in a scratch copy → both tests fail, incl. the bypass test asserting `_HANDLER_RAN is True` ("AUTH BYPASS: anonymous client dispatched an event…"). It exercises the **real** consumer via `WebsocketCommunicator(LiveViewConsumer.as_asgi())`, anonymous scope. Restored; clean. Sweep: `test_ws_auth_close_socket.py` 2/2; `-k "websocket or auth or mount or hook"` **203 passed, 0 failed**.

## Concern — `mount_batch` now tears down the shared socket on a single login-redirecting view 🟡
`_mount_one` (`websocket.py:2549`) calls `handle_mount` with `send_json` swapped to a `_collect` collector — but **`close()` is not swapped**. So when one batched view hits the redirect branch, the new `await self.close(4403)` fires **mid-loop** on the shared socket. Empirically confirmed with a real `WebsocketCommunicator` batch (redirect-first + healthy-second): event order is `["websocket.close", "websocket.send"]` — the 4403 close precedes the `mount_batch` response. A real client receives `close` first, so the survivor view (`ok-1`) **and** the carefully-collected `navigate[]` array (`websocket.py:2598-2612`) never render; the client `onclose` (`03-websocket.js:201`) then reconnect-storms (no 4403 special-case). One lazy `login_required` view kills the whole page's live socket.

The existing batch test `test_redirect_hook_surfaces_in_mount_batch_navigate` (tests/unit/test_sw_advanced.py:773) **passes but cannot catch this** — the fake consumer's `close()` is a no-op `return None` (test_sw_advanced.py:245-246). Clearing `view_instance` IS needed for batch correctness (pre-fix it leaked an unmounted view), but `close()` is the wrong tool on a multiplexed mount. Suggest: in the batch/collector path, clear `view_instance` without closing the socket (e.g. make `close` collector-aware like `send_json`, or skip close when invoked under `_mount_one`), and add a real-`WebsocketCommunicator` batch regression test. Not a new bypass, so not blocking on security grounds — but a real PR-introduced defect on the lazy-hydration path.

johnrtipton:
## Retrospective — PR #1780 (WS auth bypass, threat model T1/T2)

**Task**: a confirmed WebSocket auth bypass — `login_required` mount sent a navigate frame but never closed the socket, so a raw client could dispatch unauthenticated events.

### Quality: 5
### What went well
- **Threat-model-first (user's call) paid off twice over.** Modeling the whole surface before patching surfaced the *parallel* hook-redirect branch (T2 — fixed in the same PR, not rediscovered later) and a default-open allowlist (T4), and produced a durable artifact (`docs/audits/websocket-auth-2026-06.md`, 9 threats) with follow-ups #1777/#1778/#1779 instead of a one-line patch.
- **Reproducer-first was empirically decisive.** The RED reproducer didn't just fail an assertion — it crashed *inside* the handler body (`do_mutation` ran for an anonymous client), proving the bypass beyond doubt. RED→GREEN is the gate-off.
- **Single fix point**: `handle_live_redirect_mount` delegates to `handle_mount`, so both entry points were covered by one change.
### What didn't (caught by review)
- My first fix introduced a regression: `close(4403)` fired *inside* `mount_batch` on the shared socket, killing sibling mounts. The reviewer caught it empirically (the existing batch test couldn't — its fake consumer's `close()` is a no-op). Fixed with a `_mounting_in_batch` guard (close only when not batching; always clear `view_instance`) + a real-`WebsocketCommunicator` batch regression test. Lesson: a transport-level `close()` is unsafe on any multiplexed path — check for the collector/batch context.
### Verified
- 3 reproducer/regression tests (anonymous bypass closed + socket closed 4403 + batch survives); gate-off verified on both the bypass and the batch fix; broad WS/auth/mount/batch/hook sweep 443 passed; ruff clean; CI green. Threat model + CHANGELOG Security shipped.

RETRO_COMPLETE

PR #1781, opened 2026-06-13
feat(websocket): opt-in per-event auth re-check (#1777, threat model T3)

### LINKED ISSUE #1777: security: handle_event does not re-check auth (mid-session logout/perm-loss; threat model T3)

johnrtipton:
## Summary

Closes #1777 (threat model **T3**, defense-in-depth follow-up to the T1/T2 mount-redirect fix in #1780).

Auth runs only at mount; the connect-time WS scope user is cached, so an authenticated user who **logs out or loses a permission mid-session** keeps dispatching events on the open socket until they reconnect.

New **`LIVEVIEW_CONFIG['reauth_on_event']` (default OFF)**: when enabled and the mounted view declares `login_required`/`permission_required`, `handle_event` re-resolves the user from the session (`channels.auth.get_user`) and re-runs `check_view_auth_lightweight`; on failure it sends a navigate redirect, `close(4403)`, and clears `view_instance`. **Default OFF** because it costs one session read per event — opt in for high-security apps. **Fail-safe**: any error (e.g. no session in scope) skips the re-check rather than breaking the event path.

## Why opt-in (not default)
A meaningful re-check requires re-reading the session every event (the cached scope user doesn't reflect a mid-session logout). That's real overhead, so it's opt-in. The primary bypass (anonymous, T1/T2) is already closed at mount in #1780 regardless of this flag.

## Verification
- `test_event_reauth_1777.py` (real `WebsocketCommunicator`, authed connect user, deauthed event-time `get_user`): reauth **ON** → mid-session logout closes `4403` and the handler never runs; reauth **OFF** (default) → event dispatches normally and `get_user` is **not** consulted.
- Gate-off (#254) verified; event/auth/websocket regression sweep **93 passed**; ruff clean; two-commit shape (Gate 1 + Gate 2).
- Threat model updated: T3 → MITIGATED (opt-in).


johnrtipton (commented):
Reviewed against threat model T3 (`docs/audits/websocket-auth-2026-06.md`). Verified all five correctness checks; 2/2 new + 67/67 sweep green.

**Correctness — all confirmed:**
- **Zero-cost when OFF.** `djust_config.get("reauth_on_event")` is the first `and` operand (`python/djust/websocket.py:2769`), so `get_user` is never imported/called when OFF. `gu.assert_not_called()` in `test_reauth_off_default_dispatches_event_normally` (`test_event_reauth_1777.py:137`) is a true assertion, not a tautology.
- **Fail-safe.** Whole block wrapped in `try/except Exception` → `logger.debug(..., exc_info=True)` and falls through to normal dispatch (`websocket.py:2796-2797`). No `close()` on error — confirmed.
- **Request source correct.** Reads `getattr(self.view_instance, "request", None)`; the attr is set at mount (`websocket.py:1963`, `self.view_instance.request = request`). Matches the inline comment. `request is None` guards the fail-safe path.
- **Gated-views-only.** Guard requires `login_required`/`permission_required` truthy, so public views never pay the cost even with the flag on.
- **Permission path.** `check_view_auth_lightweight` (`auth/core.py:241`) catches `PermissionDenied` → `False`, so revoked-permission also closes 4403 (stricter than T1’s navigate-only anon branch — acceptable; doc/CHANGELOG describe it).

**Gate-off self-test (Action #1468):** patched the guard to `if False and …`, re-ran the ON-path test → **FAILED** as required (`socket not closed on deauth`). Restored clean.

**One concern (non-blocking).** On the *allow* path, `request.user = fresh_user` mutates the long-lived mount `request` and is left in place — the CHANGELOG only documents the deauth case. It self-corrects (every gated event re-resolves) and is arguably the more-correct behavior (handlers see current auth), but it is a behavior change vs OFF that the CHANGELOG/doc do not call out. Worth a one-line note that with the flag ON, `self.request.user` reflects *live* session auth inside handlers, not the connect-time scope user. No code change required.

Doc citations (`auth/core.py:241`, threat-model T3 → MITIGATED opt-in) grep-verified accurate. CHANGELOG under `### Security`, default-OFF correctly stated.

APPROVE.

johnrtipton:
## Retrospective — PR #1781 (#1777 opt-in event-path re-auth, T3)

**Task**: defense-in-depth follow-up to the T1/T2 mount-bypass fix — re-check auth on the WS event path so mid-session logout/perm-loss is enforced.

### Quality: 5
### What went well
- Correctly diagnosed that a *meaningful* T3 fix requires re-reading the session per event (the cached scope user doesn't reflect a mid-session logout) → made it **opt-in** (default OFF) so the per-event session read isn't imposed on everyone. Avoided shipping security theater (a re-check against the cached user would catch nothing).
- Found the right seam: `channels.auth.get_user(scope)` re-resolves from the session; mocked only that boundary in the test while exercising the real consumer/mount/event path with an authed connect user.
### What didn't
- First run failed because I read `self.request` (None on the consumer) instead of `self.view_instance.request` (where mount stores it, websocket.py:1963). Lesson: verify where the request actually lives on the WS path before using it.
### Verified
- 2 WebsocketCommunicator tests (ON → mid-session logout closes 4403, handler never runs; OFF default → dispatches, get_user not called); gate-off non-tautological; 93-test event/auth sweep clean; CI green. Threat model T3 → MITIGATED (opt-in).

RETRO_COMPLETE

Issue #696, opened 2026-04-12
{% csrf_token %} renders as literal 'CSRF_TOKEN_NOT_PROVIDED' in LiveView templates, poisoning HTTP fallback


johnrtipton:
## Summary

When `{% csrf_token %}` appears in a LiveView template rendered by the Rust engine, it outputs:

```html
<input type="hidden" name="csrfmiddlewaretoken" value="CSRF_TOKEN_NOT_PROVIDED">
```

instead of a real CSRF token. This breaks the HTTP fallback transport because `client.js` reads the hidden input FIRST (line 125 of `11-event-handler.js`):

```javascript
const csrfToken = document.querySelector('[name=csrfmiddlewaretoken]')?.value
    || document.cookie.match(/(?:^|;\s*)csrftoken=([^;]+)/)?.[1]
    || '';
```

The `?.value` returns the truthy string `"CSRF_TOKEN_NOT_PROVIDED"`, so it never falls through to the cookie (which has the real token). Every HTTP POST then sends `X-CSRFToken: CSRF_TOKEN_NOT_PROVIDED` → Django returns 403.

## Impact

- Any LiveView app where WebSocket is unavailable (corporate proxies, nginx misconfiguration, django-tenants ASGI routing issues) is completely broken
- The HTTP fallback for `dj-click`, `dj-model`, `dj-submit` all fail with 403
- Workaround: remove `{% csrf_token %}` from the LiveView template — djust injects it automatically. But this is non-obvious and the template tag "works" (renders without error), so developers include it following Django conventions

## Where the bug lives

- **Rust engine**: `CSRF_TOKEN_NOT_PROVIDED` string comes from the Rust binary (`_rust.cpython-*.so`). The Rust renderer handles `{% csrf_token %}` as a special tag but outputs a placeholder when no CSRF context is available.
- **Python rendering** (`template/rendering.py:690-694`): `csrf_input` and `csrf_token` ARE injected into context when `request` is present, but the Rust engine may process `{% csrf_token %}` before these context variables are applied.
- **client.js** (`11-event-handler.js:125-127`): reads hidden input before cookie. If the hidden input has ANY value (even a placeholder), it uses that.

## Proposed fix

Two options (not mutually exclusive):

1. **Rust engine**: when `{% csrf_token %}` is encountered and no real token is available, render nothing (empty string) instead of the placeholder. This lets client.js fall through to the cookie.

2. **client.js**: validate the CSRF token before using it. Skip values that look like placeholders:
[...]

PR #1783, opened 2026-06-13
test(security): pin csrf_token + {{ user }} survive WS re-renders (#1779, T9)

### LINKED ISSUE #1779: docs: {% csrf_token %}/{{ user }} inside [dj-view] are deleted on WS patches (threat model T9)

johnrtipton:
## Summary

Closes #1779 (threat model **T9**) — **as already-fixed**, with a regression test.

The draft concern (request-bound tags / context-processor vars used inside the dj-root template getting deleted on WS patches) is **already mitigated in current djust**: `_sync_state_to_rust` applies context processors on every WS update (#1722) and injects `csrf_token` (#696), so `{% csrf_token %}` and `{{ user }}` survive WS re-renders.

This PR adds the missing regression test (there was none pinning this): renders the WS path (`view.render(request=...)`) initially **and after a state change**, asserting `csrfmiddlewaretoken` + `{{ user.username }}` + `{{ user.is_authenticated }}` are present in both — never blanked/deleted.

## Verification
- 2 tests pass. **Gate-off**: neutering the #696 csrf injection in `rust_bridge.py` makes the test fail — so it genuinely pins the fix (not a tautology).
- Test-only; no production change.

Closes #1779.

### DIFF
--- a/python/tests/test_csrf_user_survive_ws_1779.py
+++ b/python/tests/test_csrf_user_survive_ws_1779.py
@@ -1,0 +1,119 @@
+"""Regression for #1779 (threat model T9): {% csrf_token %} and {{ user }} used
+inside the dj-root template must survive WebSocket re-renders.
+
+The original concern: the Rust engine re-renders the dj-root subtree on every WS
+update from view-instance attributes only, so request-bound tags ({% csrf_token %})
+and context-processor vars ({{ user }}) would render empty and the DOM diff would
+delete the logout form / CSRF'd inline form. That is FIXED in current djust:
+``_sync_state_to_rust`` applies context processors on every WS update (#1722) and
+injects ``csrf_token`` (#696). These tests pin that — they render the WS path
+(view.render(request=...)) twice (initial + a state change) and assert csrf + user
+are present in BOTH, never blanked.
+"""
+
+from unittest.mock import Mock
+
+import pytest
+from django.test import RequestFactory, override_settings
+
+try:
+    from djust import LiveView, RustLiveView
+except ImportError:  # pragma: no cover
+    LiveView = None
+    RustLiveView = None
+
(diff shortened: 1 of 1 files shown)


johnrtipton:
## Retrospective — PR #1783 (#1779 / T9 regression test)

**Task**: close T9 (`{% csrf_token %}`/`{{ user }}` deleted inside dj-root on WS patches).

### Quality: 5
### What went well
- **Verified before documenting.** The draft (and the threat model's first pass) framed T9 as a live bug; reading `rust_bridge.py:535-565` showed it's already fixed (#1722 context processors on every WS update + #696 csrf injection). Writing "here's a problem" docs would have been wrong — instead I pinned the existing fix with a test. Same "trust the symptom, not the cited state; verify from code" discipline that found T1.
- Gate-off proved the test is non-tautological: neutering the #696 csrf injection fails it.
- Caught a Rust-vs-Python rendering detail (booleans render lowercase `true`) before asserting on it.
### What didn't
- The threat model initially listed T9 as a GAP without checking #696/#1722 first — a reminder to grep for existing fixes before classifying a threat as unmitigated.
### Verified
- 2 tests pass; gate-off verified; CI green. Closes #1779 as already-mitigated (no stale docs shipped).

RETRO_COMPLETE

Issue #1788, opened 2026-06-14
fix(websocket): consumer-owned monotonic VDOM send-version so html_update fallback is accepted without a recovery round-trip


johnrtipton:
## Summary

The VDOM version sent to the client is the **Rust view's `self.version`**
(`crates/djust_live/src/lib.rs:582`, incremented unconditionally each render, starting at 0 in
`RustLiveView::new`). It is coupled to the **Rust-view object's lifetime, not the
connection**. When the Rust view loses its `last_vdom` baseline mid-session, `render_with_diff`
returns `patches=None` and a `version` that does NOT satisfy the client's
`clientVdomVersion === data.version - 1` check (`python/djust/static/djust/src/02-response-handler.js:58`).
The client then treats the `html_update` as a version mismatch and sends `request_html`.

With #1785 fixed (recovery armed on the `html_update` path) this is a harmless extra
round-trip (mismatch → request_html → html_recovery → morph). Without #1785 it is a full
page reload. This issue removes the extra round-trip by making the `html_update` directly
acceptable.

## Proposed fix (Option 2 — consumer-owned send version)

Track a monotonic send-version on the **consumer** (`self._last_sent_version`) and stamp every
outbound frame with `self._last_sent_version + 1`, decoupling the wire version sequence from
the Rust-view lifetime. A baseline loss then produces a correctly-sequenced `html_update`
(`v_n → v_{n+1}`) the client accepts directly — no `request_html`.

**Caveat to design through:** the wire version and the Rust view's internal counter must not
drift in a way that breaks the *next* successful diff's expected version (the client checks
`clientVdomVersion === data.version - 1` on EVERY frame). The consumer-owned counter must be
the single source of truth fed into outbound frames consistently across ALL send paths:
`handle_event`, `server_push`, `_run_async_work`, and `handle_request_html`.

Alternative (Option 1): client sends its `clientVdomVersion` in each event frame; server stamps
`client_version + 1` and sets the Rust view's `version` accordingly (needs a Rust `version`
setter — none exists today — and a protocol field on every event frame). Option 2 is cleaner
(no protocol change, no client change).

## Why this is a follow-up, not part of #1785

#1785 (recovery-arming) is the small, verified, reload-stopping fix and ships first. This is a
larger change touching every send path with the drift caveat above — separate PR per
scope discipline. Follow-up to #1785.

Root cause confirmed via local reproduction of the djust.org `/insights/` `set_period` flow.

PR #1789, opened 2026-06-14
fix(websocket): arm recovery on the html_update (DJE-053) fallback path (#1785)

### LINKED ISSUE #1785: fix(websocket): request_html recovery forces full page reload after WS reconnect/restart — re-render mounted view instead
## ✅ VERIFIED ROOT CAUSE & FIX (2026-06-14) — reproduced locally, gate-off confirmed

**Production symptom:** djust.org `/insights/` (djust 1.0.4) reloads on *every* time-range button
click with `[LiveView] Server error: Recovery HTML unavailable — the server may have restarted`.

Reproduced over a real `LiveViewConsumer` WebSocket (seeded data, staff auth), frame-by-frame:

1. mount → `version 1`
2. `set_period("14")` → the Rust view's diff returns **`patches=None`** (its `last_vdom`
   baseline is absent at event time) → server sends **`html_update`** (DJE-053), `version 1`
3. client (`python/djust/static/djust/src/02-response-handler.js:58`) sees a version mismatch
   (`clientVdomVersion 1 !== data.version-1 = 0`) → sends `request_html`
4. **`request_html` → `_recovery_html` is `None` → "Recovery HTML unavailable" → full page reload**

**Root cause:** `handle_event`'s `patches is None` / `html_update` branch
(`python/djust/websocket.py:3740-3816`) sends the full-HTML fallback but **does NOT call
`self._arm_recovery(html, version)`** — only the patches branch does (`:3726`). So on-demand
recovery is unavailable *exactly* when the full-HTML fallback fires.

**Fix (verified):** add `self._arm_recovery(html, version)` in the `html_update` branch, using
the raw rendered `html` BEFORE the `_sync_strip_and_extract` reassignment at `:3749` (so it
matches what the patches branch passes at `:3726`).

Gate-off self-test (Action #1468): with the fix, `request_html` returns
`{"type":"html_recovery", html, version}` → the client morphs (no reload), and
`consumer._recovery_html` is populated. Reverting the one line reproduces the original bug.
The normal patches path is unaffected (no regression).

All `patches=None` sub-cases flow through this single `else` branch — component_event (`:3635`),
force_full_html (`:3650`), patch_compression (`:3690`), and DJE-053 — so arming recovery once
in the branch covers them all. The async path (`:958/:977`) and `server_push` already arm
recovery; audit `handle_request_html` and any other outbound `html_update` site for the same gap.

**Regression test:** a minimal `LiveView` whose event handler sets `self._force_full_html = True`
deterministically hits the `html_update` branch — assert `consumer._recovery_html is not None`
after the event AND that `handle_request_html` returns `type: "html_recovery"` (not the
"unavailable" error). Use a `WebsocketCommunicator` per
`python/djust/tests/test_ws_reconnect_state_1465.py`.

**Related:** the version desync (the `html_update` carries a version the client rejects, forcing
the `request_html` round-trip) is tracked in the follow-up issue. The recovery-arming fix here
eliminates the **reload**; the version fix removes the extra round-trip.

---

## Summary

When a client's VDOM patch fails to apply it sends `{"type":"request_html"}` to fetch
recovery HTML. `LiveViewConsumer.handle_request_html` (`python/djust/websocket.py:4644`)
serves `self._recovery_html`; if it is `None` it returns a **non-recoverable** error:

> Recovery HTML unavailable — the server may have restarted. A page reload will fix this.

(`python/djust/websocket.py:4659-4665`). The client then does a full page reload.

`_recovery_html` is a **per-consumer, in-memory** attribute, armed via `_arm_recovery`
(`python/djust/websocket.py:1039-1052`) on each render-send (e.g. the event path at
`python/djust/websocket.py:3726`). It is **not** stored in the state backend. So whenever a
WebSocket reconnects to a *fresh* consumer — pod restart, deploy rollover, k8s scale,
network blip, or a request that lands on a different replica — `_recovery_html` is `None`
until that consumer next renders. A patch failure in that window forces a full page reload
even though the view is fully mounted and its state is available.

## Impact

On a multi-replica deployment (e.g. djust.org runs `replicas: 2` with the Redis state
backend), this is reproducible and persistent: pods restart under load, the client
reconnects, and the next interaction that produces a non-applyable patch hits
`_recovery_html is None` → forced reload. Users perceive the app "reloading itself" on a
click.

Observed in production (djust.org `/insights/`, djust 1.0.4): pods cycling (liveness
kills), WS reconnects, and the exact `Recovery HTML unavailable — the server may have
restarted` console error on a button click that changes DOM structure.

## Root cause

`handle_request_html` gives up when `_recovery_html is None` — but it has already verified
`self.view_instance` is mounted (`python/djust/websocket.py:4652`). A mounted view with
restored state can simply be **re-rendered on demand** to produce recovery HTML. There is
no need to fail.

## Proposed fix

In `handle_request_html`, when `_recovery_html is None` but `self.view_instance` is
present, render the live view (the same render path used elsewhere) to produce the recovery
HTML and serve it — instead of `send_error(..., recoverable=False)`. Only fall back to the
non-recoverable error if rendering itself fails or the view is genuinely unmounted. This
makes on-demand recovery survive reconnects/restarts and removes the forced-reload class
for every consumer that has restored state but not yet armed `_recovery_html`.

## Repro

1. Multi-replica (or simply restart the worker) + a view whose event changes DOM structure
   enough that `applyPatches` returns false (e.g. a list/chart whose length changes).
2. Trigger the event right after a reconnect (before the new consumer armed recovery).
3. → full page reload with the "server may have restarted" error.

## Evidence

- `python/djust/websocket.py:4644-4665` — `handle_request_html`, the unavailable branch
- `python/djust/websocket.py:4652` — `view_instance` already verified mounted
- `python/djust/websocket.py:1039-1052` — `_arm_recovery` (per-consumer, in-memory)
- `python/djust/websocket.py:3726` — recovery armed only on the render-send path
- Production: djust.org 1.0.4, `AnalyticsInsightsView`, replicas: 2, Redis state backend

johnrtipton:
## Summary

Fixes #1785. On the DJE-053 `html_update` fallback (an event's VDOM diff returns no patches → the server sends full HTML), `handle_event` did **not** arm on-demand recovery — only the patches branch did. So a client that subsequently requested recovery (e.g. after a VDOM version mismatch on the full-HTML frame) received `Recovery HTML unavailable — the server may have restarted` and **reloaded the whole page**.

## Root cause (verified by local reproduction — gate-off confirmed)

Reproduced frame-by-frame over a real `LiveViewConsumer` WebSocket:

1. mount → `version 1`
2. event → diff returns `patches=None` → server sends `html_update` (DJE-053)
3. client (`02-response-handler.js:58`) sees a version mismatch → sends `request_html`
4. `request_html` → `_recovery_html` is `None` → "Recovery HTML unavailable" → reload

`handle_event`'s `patches is None` branch (`python/djust/websocket.py:~3752`) sent the full HTML but skipped `_arm_recovery`. (`_run_async_work` already arms recovery on *its* full-HTML fallback — `handle_event` was the gap.)

## Fix

Arm recovery with the raw pre-strip HTML in the `html_update` branch, mirroring the patches branch. One call + comment.

## Tests

- **`test_ws_recovery_html_update_1785.py`** — a `WebsocketCommunicator` round-trip that forces the `html_update` fallback (`_force_full_html`) and asserts `request_html` now returns an `html_recovery` frame (client morphs) instead of the error (reload). **Gate-off self-test: fails without the fix.**
- Updated the `_arm_recovery` call-site count guard (#1645/#1125): 4 → 5 (the new legitimate site).

## Verification

`make test` green — Python + Rust + JS (1686).

## Notes

- Surfaced by djust.org `/insights/` reloading on every time-range switch (multi-replica, djust 1.0.4).
- Follow-up **#1788** (consumer-owned VDOM send-version) removes the residual recovery round-trip (the version desync); this PR stops the *reload*.
- Pre-existing serial-`pytest`-order test pollution (S005 + two `auto_navigate_meta` tests) is **not** introduced here — those pass in isolation and under parallel `make test`; will be filed as a separate test-isolation tech-debt issue.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- **WebSocket recovery no longer forces a full page reload on the `html_update` fallback (#1785).** When a LiveView event's VDOM diff returns no patches and the server sends a full-HTML `html_update` frame (the DJE-053 fallback), it now arms on-demand recovery — matching the patches path. Previously the `html_update` branch in `handle_event` skipped `_arm_recovery`, so a client that subsequently requested recovery (e.g. after a VDOM version mismatch on the full-HTML frame) received `Recovery HTML unavailable — the server may have restarted` and reloaded the whole page instead of morphing. Surfaced by a multi-replica djust.org `/insights/` page reloading on every time-range switch. Added a `WebsocketCommunicator` regression test (`TestWSRecoveryHtmlUpdate`-style, in `test_ws_recovery_html_update_1785.py`) plus a source pin.
+
 ## [1.0.4] - 2026-06-13
 
 ### Added
--- a/python/djust/tests/test_ws_recovery_html_update_1785.py
+++ b/python/djust/tests/test_ws_recovery_html_update_1785.py
@@ -1,0 +1,149 @@
+"""Regression tests for #1785 — the DJE-053 ``html_update`` fallback must arm recovery.
+
+Root cause (verified by local reproduction of the djust.org ``/insights/`` incident):
+when an event's VDOM diff returns ``patches=None``, ``handle_event`` sends a full
+``html_update`` frame but historically did NOT call ``self._arm_recovery(html, version)``
+— only the patches branch did. So when the client (which sees a version mismatch on the
+``html_update``) follows up with ``request_html``, the server's ``_recovery_html`` is
+``None`` and it replies with the non-recoverable error
+``"Recovery HTML unavailable — the server may have restarted"`` → the client does a full
+page reload.
+
+The fix arms recovery on the ``html_update`` branch too. These tests pin both the runtime
+behavior (a real ``WebsocketCommunicator`` round-trip) and the source.
+
+A view that sets ``self._force_full_html = True`` in its event handler deterministically
+forces the ``patches=None`` / ``html_update`` branch, so the test does not depend on
+losing the Rust ``last_vdom`` baseline.
+"""
+
+from __future__ import annotations
+
+import inspect
+
+import pytest
+from asgiref.sync import sync_to_async
+
+from djust import LiveView
+from djust.decorators import event_handler
+
+
+class _WSForceFullView(LiveView):
+    """Module-level view whose event forces the full-HTML (``html_update``) path."""
+
+    template = (
+        '<div dj-view="djust.tests.test_ws_recovery_html_update_1785._WSForceFullView" '
+        'dj-id="0">Count: {{ count }}</div>'
+    )
+
+    def mount(self, request, **kwargs):
+        self.count = 0
+
+    @event_handler()
+    def bump_full(self, **kwargs):
+        self.count += 1
+        # Force the DJE-053-class fallback: patches=None → html_update frame.
+        self._force_full_html = True
+
+
+async def _receive_until(communicator, wanted_type, *, tries=5, timeout=3):
+    """Drain frames until one whose ``type`` == ``wanted_type`` (or return last seen)."""
+    last = None
+    for _ in range(tries):
+        last = await communicator.receive_json_from(timeout=timeout)
+        if last.get("type") == wanted_type:
+            return last
+    return last
+
+
+@pytest.mark.django_db
+@pytest.mark.asyncio
+async def test_html_update_fallback_arms_recovery_so_request_html_succeeds():
+    """The load-bearing regression test.
+
+    Drive ``handle_event`` down the ``html_update`` (patches=None) path via a real
+    ``WebsocketCommunicator``, then send ``request_html`` and assert the server returns
+    an ``html_recovery`` frame — NOT the "Recovery HTML unavailable" error.
+
+    Gate-off: reverting the ``_arm_recovery`` call in the ``html_update`` branch makes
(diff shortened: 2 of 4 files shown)


johnrtipton:
## Retrospective — PR #1789 (pipeline-drain)

**Task**: v1.0.5-1 — #1785 (arm recovery on the `html_update`/DJE-053 fallback).

### Quality: 5
### What went well
- Local WebSocket reproduction nailed the root cause frame-by-frame (mount v1 → html_update v1 → version mismatch → request_html → `_recovery_html` None → reload).
- Gate-off self-test (#1468) confirmed the regression test is non-tautological (fails without the fix).
- The `_arm_recovery` call-site count guard (#1645/#1125) did its job — it forced a conscious count bump AND caught a comment-literal false count (a `self._arm_recovery(` string in a docstring inflated the naive `src.count`); reworded the comment.
### What didn't
- Three wrong theories first (OOM/memory, multi-pod, variable-length template) before reproducing locally settled it — reproduce-first earlier would have saved cycles.
- Surfaced pre-existing serial-`pytest`-order test pollution (S005 + two `auto_navigate_meta` tests); to be filed separately.
### Verified
`make test` green (Python + Rust + JS 1686); all required CI checks pass (py3.12/py3.14t/rust/security/js/demo-checks/nav-hooks-guard/test-summary); gate-off confirms failure without the fix.

`RETRO_COMPLETE`

PR #1793, opened 2026-06-14
fix(serialization): exclude request + context-processor outputs from persisted LiveView state (#1786)

johnrtipton:
Fixes #1786.

## Symptom (reproduced)

A `LiveView` whose `mount()` assigns only JSON-serializable scalars (e.g. `self.days` int, `self.show_raw` bool) still emitted, on **every** render/event, a flood of `serialization` warnings naming values the view never assigns to `self`:

```
WARNING serialization LiveView state contains non-serializable value: ASGIRequest / WSGIRequest ...
WARNING serialization LiveView state contains non-serializable value: PermWrapper (from django.contrib.auth.context_processors) ...
WARNING serialization LiveView state contains non-serializable value: FallbackStorage (from django.contrib.messages.storage.fallback) ...
WARNING serialization LiveView state contains non-serializable value: SimpleLazyObject / UserLazyObject ...
```

plus `dict '_prev_context_refs' has N keys — key fingerprint truncated`. A standalone repro of the production `AnalyticsInsightsView` shape emitted **8** non-serializable warnings across two renders; `_prev_context_refs` carried 32 keys including `request` / `user` / `perms` / `messages` / `DEFAULT_MESSAGE_LEVELS`. After the fix: **0** warnings, those 5 request-scoped keys gone from the fingerprint.

## Traced root cause (exact code path)

1. `_sync_state_to_rust` (`python/djust/mixins/rust_bridge.py:548`) calls `_apply_context_processors(full_context, request)`, which folds the request + the standard context-processor outputs (`request`, auth `user`/`perms`, messages `FallbackStorage`, etc.) into `full_context`. The view never assigns these to `self` — they are added purely so the Rust template can render `{{ user }}` / `{% csrf_token %}`.
2. On the **first** render `context = full_context`, so the non-serializable processor values flowed into `json_compatible_context = normalize_django_value(rendered_context)` (rust_bridge.py:~770) → the non-serializable fallback at `serialization.py` logs a warning per value and `str()`-stringifies them into the state pushed to the backend.
3. On **every subsequent WS event**, those processor outputs are recreated with a fresh `id()` (`_apply_context_processors` re-runs each cycle per #1722/#1726), so they re-entered the changed-key subset and re-warned every event.
[...]

PR #1790, opened 2026-06-14
fix(cli): scaffold boots on uvicorn+ASGI (ASGIStaticFilesHandler, drop daphne) + load .env (#1787)

johnrtipton:
## Summary

`djust new <name>` — the recommended way to start a djust project — generated a project that **does not boot**. Fixes #1787.

Note on the cited path: the issue cites `python/djust/cli.py`, but `djust new` actually scaffolds via `python/djust/scaffolding/templates.py` + `generator.py` (`cli.py` only hosts the deprecated `startproject` command). Per the bug-triage rule, I traced from the symptom to the real code path and fixed the templates `djust new` actually uses.

## Root cause

Two independent boot blockers:

1. **`asgi.py` misused `live_session`.** The template wrote `application = live_session()`, but `live_session(prefix, patterns, ...)` (`routing.py`) is a **URL-pattern helper** that returns `List[URLPattern]` and requires positional args — not an ASGI app. Importing the scaffolded `asgi.py` raised `TypeError: live_session() missing 2 required positional arguments`, so `make dev` crashed immediately.

2. **The scaffold failed `manage.py check`** (which blocks `migrate`) on **two ERRORS**:
   - `djust.A014` — `settings.py` read `os.environ` but **never loaded the generated `.env`**, so a fresh scaffold ran in production mode (`DEBUG=False`) and the `django-insecure-` SECRET_KEY was flagged.
   - `admin.E403` — no `DjangoTemplates` backend was configured (only `DjustTemplateBackend`), which `django.contrib.admin` requires.

   (The issue body lists A030/C010/C012/S005 as errors; in current djust those are **warnings**. The actual blocking ERRORS were A014 + admin.E403.)

## The fix (lifted from `examples/demo_project/demo_project/asgi.py`)

The new `asgi.py` template mirrors the demo's gold pattern, on a properly-configured **uvicorn + ASGI** stack — **no daphne, no WhiteNoise**:

```python
django_asgi_app = get_asgi_application()  # populates the app registry first
# ... channels / djust.websocket imports after, # noqa: E402
application = ProtocolTypeRouter({
    "http": ASGIStaticFilesHandler(django_asgi_app),   # serves client.js/CSS, no WhiteNoise
    "websocket": AllowedHostsOriginValidator(
        AuthMiddlewareStack(
            URLRouter([path("ws/live/", LiveViewConsumer.as_asgi())])
        )
    ),
})
```

[...]

PR #1798, opened 2026-06-15
fix(hooks): pre-push hook resolves the venv from any git worktree (#1796)

### LINKED ISSUE #1796: tooling: pre-push hook hardcodes .venv/bin/python and fails inside a git worktree

johnrtipton:
## Summary

The native pre-push hook hardcoded `.venv/bin/python` relative to the current working directory and failed inside a `git worktree` (which has no `.venv`), forcing `git push --no-verify` and skipping the real gates. Hit on all three v1.0.5-1 drain PRs (#1790, #1792, #1793). Closes #1796.

### Where the hook actually lives

- `.git/hooks/pre-push` is **generated by the pre-commit framework** and is **not repo-tracked** — it just `exec`s `pre-commit hook-impl --hook-type=pre-push`. It does NOT itself hardcode `.venv`.
- The hardcoded `.venv/bin/python` lives in the **repo-tracked `.pre-commit-config.yaml`** — 7 hook `entry:` lines (6 pre-push + the commit-stage `check-doc-snippets`). That is the PR-able source.
- The same bug also affected ~31 `make` targets (`make test` etc.), so a worktree couldn't run gates manually either (`Error 127`). Fixed here too so the manual-gates path the worktree caveat relies on actually works.

### Fix

- New `scripts/run-with-venv-python.sh` resolves the interpreter relative to the **MAIN working tree root** — `dirname` of the absolute `--git-common-dir` (which points at `<main-root>/.git` for both the main checkout and every linked worktree) — with `uv run python` then `python3` fallbacks when no `.venv` exists (CI, fresh clone). It runs in exec mode and in `--print` mode (for `PYO3_PYTHON` / `cargo test`).
- All 7 hook entries route through the resolver; no `.venv/bin/python` remains in `.pre-commit-config.yaml`.
- The Makefile's ~31 hardcoded references collapse to one `$(PYTHON)` variable computed once via the resolver (override with `make PYTHON=...`).

### Verification

- **Throwaway `git worktree` (the bug's exact scenario):** the OLD entry dies with `bash: .venv/bin/python: No such file or directory` (exit 127); the FIXED entry runs the real check (`44 tag-emit defaults checked`, exit 0). Real pre-commit pre-push machinery (`uvx pre-commit run --hook-stage pre-push ...`) runs `check handler contracts...Passed` and `docs stale-MD-ref check...Passed` from the worktree; `PYO3_PYTHON`/pytest entries resolve the main venv.
- **`make test` green from the worktree:** 4573 Python passed / 18 skipped, Rust ok, 1686 JS passed — impossible before the fix.
- **6 regression cases** in `tests/test_run_with_venv_python.py` (real worktree resolution, main-checkout no-regression, `python3` fallback, no-interpreter error, plus source-pins on the config + Makefile). Gate-off self-test (#1468): the worktree case fails against the pre-#1796 resolver and passes against the fix.

### Caveat (the irony this PR fixes)

This branch was authored in a worktree, so it was pushed with `--no-verify` after running the gates manually — exactly the workaround #1796 is about. Once merged, that workaround is no longer needed; CI is the authoritative gate.


johnrtipton:
## Retrospective — PR #1798 (pipeline-drain v1.0.5-2)

**Task**: #1796 — pre-push hook hardcodes `.venv/bin/python`, fails in git worktrees.

### Quality: 5
- Triage finding: the hardcode lives in repo-tracked `.pre-commit-config.yaml` (7 entries) **and ~31 Makefile targets** (so `make test` itself was broken from a worktree), not the generated `.git/hooks/pre-push`.
- New `scripts/run-with-venv-python.sh` resolves the interpreter from the main working-tree root via `git rev-parse --git-common-dir` → works identically from any linked worktree; `.venv` → `uv run` → `python3` fallback; `--print` mode for `PYO3_PYTHON`.
- Worktree-verified (old: exit 127; new: exit 0); gate-off non-tautological; `make test` green from a worktree.

`RETRO_COMPLETE`

Issue #1174, opened 2026-04-27
tech-debt: '3 clean full-suite runs' verification gate for pollution-class fixes


johnrtipton:
Source: v0.9.1 retro / Action Tracker #182 / PR #1159 (#1134 bisect).

## Context

The bisect agent for #1134 didn't stop at the first hit (in-memory SQLite leak from `test_async_render_path.py`). It kept verifying until 3 clean full-suite runs, which surfaced an unrelated SECOND polluter (`sys.modules` rebind in `test_dev_server_watchdog_missing.py`) that wasn't on the original 6-test list.

Without the verify-3x gate, the second polluter would have shipped silently — and the next PR would have hit the same flake.

## Proposed canonicalization

Update `.pipeline-templates/bugfix-state.json` Stage 6 (Test Execution) checklist to add a conditional mandatory item:

> `{"action": "if pipeline_type=='bugfix' AND root_cause is 'test pollution': run full pytest suite 3 times, all clean", "done": false, "mandatory": true, "if": "pollution-class"}`

Or a Stage 6 sub-stage that re-runs the full suite N times when the bugfix description matches `/pollution|leak|flak/i`.

## Scope

State-file template addition + ~10 LoC skill text. Single-line conditional in the Stage 6 checklist.

## Why deferred from v0.9.1

The bisect agent did this voluntarily. Canonicalizing makes it impossible to skip on the next pollution-class fix.

PR #1800, opened 2026-06-15
test: fix serial-order pollution breaking test_checks S005 + auto_navigate_meta (#1794)

### LINKED ISSUE #1794: test-isolation: serial pytest order pollutes test_checks S005 + auto_navigate_meta (leaked settings.DATABASES / LIVEVIEW_CONFIG)
Under broad **serial** `pytest python/` / `make test-python` ordering, these fail: `test_checks.py::TestS005UnauthenticatedViews::test_s005_suppressed_with_login_required_false` and `test_client_config_tag.py::test_auto_navigate_meta_*`. All pass in isolation and under the **parallel** `make test` (-n auto) gate; present on main (not from the v1.0.5-1 drain PRs). Root cause: a test mutating `settings.DATABASES` (postgresql) and/or `LIVEVIEW_CONFIG['auto_navigate']` without `override_settings`/teardown leaks into later tests. Fix: find the polluter(s) and isolate. Surfaced repeatedly during the v1.0.5-1 drain.

johnrtipton:
Closes #1794.

## Symptom

Under the broad **serial** pytest order (`pytest python/` / `make test-python`), three tests fail that pass both in isolation and under the parallel `make test -n auto` gate:

- `python/tests/test_checks.py::TestS005UnauthenticatedViews::test_s005_suppressed_with_login_required_false`
- `python/djust/tests/test_client_config_tag.py::test_auto_navigate_meta_emitted_when_enabled`
- `python/djust/tests/test_client_config_tag.py::test_auto_navigate_meta_engines_identical`

The parallel gate is green because xdist isolates each worker process **and** `make test` only runs `pytest tests/ python/tests/ -n auto` — it never collects `python/djust/tests/` (where the auto_navigate tests live).

## Root cause — TWO independent polluters (neither a `settings.DATABASES` / `LIVEVIEW_CONFIG` leak)

The issue hypothesized a leaked `settings.DATABASES` / `LIVEVIEW_CONFIG`. Reproducer-first tracing showed the real causes are different (trust the symptom, not the cited cause):

### 1. S005 — leaked module-level `LiveView` subclass

`python/djust/tests/test_ws_auth_close_socket.py` defines a **module-level** `LiveView` subclass `_PublicView` with **no `login_required`** and exposed state (`self.ok = True` in `mount`). Being module-level, it permanently joins `LiveView.__subclasses__()`, which the `djust.S005` system check walks. With `login_required is None` + exposed state, S005 fires:

```
djust.tests.test_ws_auth_close_socket._PublicView exposes state without authentication.
```

That message satisfies `any("PublicView" in e.msg)` in `test_s005_suppressed_with_login_required_false`, so the test fails in any serial order where `test_ws_auth_close_socket` is imported before `test_checks` (always, in `make test-python`).

**Fix:** mark `_PublicView` `login_required = False` ("intentionally public"). This is the view's actual contract (it is the deliberately-public sibling in the mount-batch test) and is exactly what S005 treats as "auth decision made → skip."

### 2. auto_navigate — `importlib.reload` orphans the config singleton

`tests/unit/test_ws_compression_config.py::_fresh_config` called `importlib.reload(djust.config)`. The reload re-runs `config = LiveViewConfig()` at module scope, rebinding `djust.config.config` to a **brand-new** object — but every `from djust.config import config` consumer (notably `djust.templatetags.live_tags`, which does `from ..config import config` at import time) keeps its reference to the **OLD** object.

After the reload, the `auto_navigate` tests `from djust.config import config` (→ new object) and call `config.reset()` under `override_settings(LIVEVIEW_CONFIG={"auto_navigate": True})`, so the **new** singleton sees `auto_navigate=True`. But `{% djust_client_config %}` in `live_tags.py` reads `config.get("auto_navigate")` on the **old** singleton → `False` → the `<meta>` is never emitted → assertion fails.

Empirically confirmed (object identity diverges after reload; new sees True, old sees False):

```
after reload:  live_tags.config is djust.config.config ?  False
test_config.get(auto_navigate) = True
live_tags.config.get(auto_navigate) = False
```

The `test_wizard_mixin.py` co-dependency is just that `live_tags` must already be imported (its `config` bound) before the reload; the reload is the polluter.

**Fix:** `_fresh_config` re-reads settings via `config.reset()` on the **shared** singleton (identical "re-pick-up settings" effect — verified for `DJUST_WS_COMPRESSION` True/False/1/0 — with no orphaning), plus an autouse teardown fixture that `config.reset()`s after each test so a `DJUST_WS_COMPRESSION` override doesn't leak via the singleton's `_config`.

## Verification — 3-clean-runs gate (pollution class, #1174)

Full serial suite `pytest tests/ python/tests/ python/djust/tests/ -p no:randomly`, three consecutive runs, all clean:

| run | result |
|-----|--------|
| 1 | 7692 passed, 21 skipped, 0 failed (2:35) |
| 2 | 7692 passed, 21 skipped, 0 failed (2:40) |
| 3 | 7692 passed, 21 skipped, 0 failed (2:28) |

Parallel gate (`pytest tests/ python/tests/ -n auto`): 4573 passed, 0 failed.

Gate-off evidence: pre-fix the same serial order produced exactly the 3 reported failures; the polluter pairs (`test_ws_auth_close_socket` → S005; `test_wizard_mixin` + `test_ws_compression_config` → auto_navigate) failed pre-fix and pass post-fix.

Test-only changes; no framework behavior change.


johnrtipton:
## Retrospective — PR #1800 (pipeline-drain v1.0.5-2)

**Task**: #1794 — serial-pytest order pollution (S005 + auto_navigate_meta).

### Quality: 5
- Reproduced the exact `make test-python` serial order; found TWO independent polluters (neither the hypothesized `settings.DATABASES` leak):
  1. **S005**: `test_ws_auth_close_socket.py`'s module-level `_PublicView` (no `login_required`) lives permanently in `LiveView.__subclasses__()` → S005 fires on it. Fixed: `login_required = False` (its real contract — the public sibling in the mount-batch test).
  2. **auto_navigate**: `test_ws_compression_config.py`'s `importlib.reload(djust.config)` rebinds the config singleton while `live_tags` keeps the OLD ref → the meta is never emitted. Fixed: `config.reset()` on the shared singleton + autouse teardown.
- **3-clean-runs gate (#1174)**: full serial suite × 3 = 7692 passed, 0 failed each; parallel gate green.

`RETRO_COMPLETE`

PR #1804, opened 2026-06-15
fix(templates): {% extends %} pages keep the base <head> on initial HTTP GET (#1801)

### LINKED ISSUE #1801: Pages using {% extends %} render WITHOUT the base template's <head> on initial HTTP GET (served as dj-root fragment only)
## Summary
A LiveView whose template `{% extends "base.html" %}` serves an initial HTTP GET that contains **only the `dj-root` subtree + djust's injected scripts** — no `<!doctype>`, `<html>`, `<head>`, `<title>`, `<style>`, `<nav>`, or `<body>` from the base template. Result: **every extends-based page renders unstyled on first paint** (the base `<head>` — Tailwind CDN, `<style>`, meta — never loads).

## Reproduction (untouched scaffold)
```bash
djust new demo --no-setup
cd demo
# boot (uvicorn/daphne) with SECRET_KEY set + migrate --skip-checks
curl -s http://127.0.0.1:8000/ | head -c 200
```
Response **starts with** `<div dj-root dj-view="demo.views.DemoView">…` — there is **no `<!doctype>`/`<head>`** anywhere in the response, even though `demo/templates/demo/base.html` defines a full `<head>` (Tailwind, `<style>`) and `index.html` does `{% extends "demo/base.html" %}`.

## This is NOT intentional / NOT a resolution failure
- `djust._rust.resolve_template_inheritance("demo/index.html", <dirs_list>)` returns the **full document (4294 chars) WITH `<head>`, `<style>`, `<!doctype>`** — resolution works.
- `mixins/template.py:155` sets `self._full_template = resolved` with the comment *"Store full template for initial GET rendering."* — the clear intent is to serve the full page.
- The `dj-root`-in-child check below it only selects `vdom_source` (the VDOM extraction), **not** `_full_template`.

So `_full_template` correctly holds the full doc (with head), but the **served GET output drops it** — the head is lost **downstream in `render_full_template`** (the page is served as the dj-root fragment only).

## Impact
Every page that uses template inheritance (i.e. essentially every real app, including the scaffold the CLI generates) is **unstyled with no `<head>` on initial load**. HIGH severity for the getting-started/first-paint experience.

## Likely area
`mixins/template.py::render_full_template` — possibly an over-correction related to #1746 (which addressed the *opposite* symptom: double-rendering/nesting the whole page). The fix should ensure the resolved `_full_template` shell (with `<head>`) wraps the dj-root render in the served response.

---
_Found via an LLM eval harness building standalone djust apps; surfaced because generated apps rendered correct markup but appeared unstyled in the browser._

johnrtipton:
## Summary

Fixes #1801. A `LiveView` whose template `{% extends "base.html" %}` served an
initial HTTP GET containing **only** the `dj-root` subtree — no `<!doctype>`,
`<html>`, `<head>`, `<title>`, `<style>`, `<nav>`, or `<body>` from the base
template. Every template-inheritance page, including the untouched `djust new`
scaffold, rendered **unstyled on first paint** because the base `<head>`
(Tailwind CDN, `<style>`, meta) never loaded.

## The exact swallowed exception (load-bearing diagnostic)

Captured by temporarily logging at the `except` in `get_template()` and driving
a fresh `djust new demo --no-setup` GET:

```
RuntimeError: Template error: Template not found: demo/index.html
Searched in:
  - /private/tmp/.../demo/templates
```

It raised at the **`resolve_template_inheritance(self.template_name,
template_dirs_str)` call** in `python/djust/mixins/template.py:get_template()`.
The Rust resolver was searching only the project `DIRS`
(`<project>/templates`, empty), because the **app** templates dir where the
scaffold's `base.html`/`index.html` actually live
(`<project>/<app>/templates/<app>/…`) was **never collected**.

## Root cause (traced symptom-up, not from the cited path)

The issue cited `render_full_template` ("head lost downstream"), assuming
`_full_template` held the full doc. Tracing the served GET path showed the
opposite: `_render_full_template_inner` splices the shell+dj-root **only when
`self._full_template` is truthy**; its `else` is `return self.render(request)`
→ the fragment.

`self._full_template` was set to `None` inside a **broad `except Exception`**
that wrapped the `resolve_template_inheritance` call (logged only at DEBUG).
The dir-collection that feeds the resolver had a hardcoded backend-name check:

```python
if template_config["BACKEND"] == "django.template.backends.django.DjangoTemplates":
    ...collect <app>/templates dirs...
```

The scaffold (and any project) configures djust's **own** backend
`djust.template.backend.DjustTemplateBackend` with `APP_DIRS=True`. That backend
honors `APP_DIRS` exactly like Django's (it subclasses `BaseEngine` with
`app_dirname = "templates"`), but the hardcoded check didn't recognize it → the
app-template dirs were dropped → `resolve_template_inheritance` raised
`Template not found` → swallowed at DEBUG → `_full_template = None` →
fragment-only.

The same hardcoded check existed in **two more parallel paths** (parallel-path
drift, #1646), so a point fix would have left them broken for the same reason:
- `utils._get_template_dirs_cached()` — used by `render_full_template` step 2 to
  render the page **shell**.
- `mixins/jit.py` — used by JIT template-variable extraction.

## Fix (two parts)

1. **Fix the underlying error + retire the drift.** Recognize the djust
   backend(s) in the APP_DIRS dir-collection via a shared
   `utils._APP_DIRS_TEMPLATE_BACKENDS` set, and route `get_template()`,
   `render_full_template` step 2, **and** `jit.py` extraction through the single
   `get_template_dirs()` helper — one source of truth for which backends honor
   `APP_DIRS`.
2. **De-silence the `except`.** Narrow the broad swallow to scope **only** the
   `resolve_template_inheritance` call (the legitimate raw-template fallback for
   a template that genuinely can't be resolved standalone) and log at
   **WARNING**. The post-resolution VDOM extraction / whitespace-strip is moved
   **outside** the `try` so an unexpected framework error surfaces instead of
   silently degrading to fragment-only. Legitimate fallback behavior
   (non-extends / unresolvable templates) is preserved.

## Verification

- **Scaffold GET before/after** (`djust new demo --no-setup`, GET `/`):
  - Before: response starts with `<div dj-root dj-view="demo.views.DemoView">…`,
    `<!DOCTYPE`/`<head>`/`<title>`/`<style>` all absent.
  - After: response starts with `<!DOCTYPE html>` and contains the base
    `<head>`, `<title>Demo</title>`, and the base `<style>` (`#0a0a14`); the
    dj-root content is spliced into the shell.
- **Regression tests** (`tests/integration/test_extends_head_initial_get_1801.py`)
  drive the real `as_view()` GET under djust's **own** backend (the exact config
  the bug reproduces under): `test_extends_get_includes_base_head`,
  `test_full_template_is_populated_for_extends`,
  `test_app_template_dirs_collected_for_djust_backend`,
  `test_resolution_failure_is_logged_not_silent`.
- **Gate-off self-test (#1468):** reverting the backend-set fix makes three of
  the four fail fragment-only; reverting the WARNING (back to DEBUG) makes the
  logging test fail — pinning the silent-catch regression both ways.
- **No regression:** full deterministic serial suite
  (`tests/ python/tests/ python/djust/tests/`) — **7704 passed, 0 failed**. The
  exact `make test` parallel invocation (`tests/ python/tests/ -n auto`) is
  3/3 clean. Non-extends LiveViews and the existing `test_sticky_http_get_1784.py`
  / SSR-parity suites are unaffected.

## Notes

- The native pre-push hook from a `git worktree` resolves the **main checkout's**
  venv (per #1796, correctly), but that venv's editable install points the
  hook at the **main checkout's** source — so it tested the unfixed code and my
  new tests failed there. #1796 fixed interpreter *resolution*; it does not make
  the hook test the worktree's tree. Gates were therefore run manually (full
  serial suite + ruff + changelog check) and the branch pushed with
  `--no-verify`; **CI is the authoritative gate**.


johnrtipton:
## Retrospective — PR #1804 (pipeline-run, #1801)

**Task**: `{% extends %}` pages rendered without the base `<head>` on initial HTTP GET (fragment-only). HIGH severity — every template-inheritance page incl. the scaffold.

### Quality: 5
### What went well
- Triage (symptom-up) correctly fingered the **silent catch** (`get_template:193` broad `except` → `_full_template=None` → `_render_full_template_inner` else returns the dj-root fragment), overriding the issue's "head lost in render_full_template" cited path.
- Reproduce-first captured the *underlying* error the catch hid: `resolve_template_inheritance` → `Template not found` because the dir-collection hardcoded `BACKEND == DjangoTemplates`, dropping app-template dirs for projects on djust's own `DjustTemplateBackend` + `APP_DIRS=True` (the scaffold's config).
- Fixed all **3** parallel dir-collection paths via one shared `get_template_dirs()` + `_APP_DIRS_TEMPLATE_BACKENDS` set (parallel-path-drift, #1646), and **de-silenced** the `except` (now WARNING-logged, wrapping only the resolution call).
- Fresh-eyes Code Review **APPROVE** (gate-off/gate-on empirically validated against the real Rust extension under the exact scaffold backend); `demo-checks` (scaffold gate) + full CI green; 7704 tests pass.
### What didn't
- The Code Review subagent manipulated the **main checkout** to exercise the compiled Rust path and left `core.bare = true`, breaking the parent's git work-tree until recovered (`git config core.bare false`). **Lesson: a read-only review agent must never mutate the main checkout / `core.bare` — exercise the compiled extension in a worktree or read-only.** (Worth a CLAUDE.md canon.)
- #1796's worktree pre-push fix resolves the interpreter but the editable install still points the hook at the *main* source, so a worktree pre-push tests the wrong tree → still `--no-verify` (follow-up on #1796).
### Deferred (advisory 🟡)
- #1805 — `.exists()` vs `.is_dir()` inconsistency between the two dir collectors (+ optional test-comment).
- `except`-narrowing turning a previously-silent extract/strip failure into a 500: intentional, documented accept (a framework bug should surface, not silently degrade).

`RETRO_COMPLETE`

PR #1806, opened 2026-06-15
fix(utils): get_template_dirs uses is_dir() to match DjustTemplateBackend collector (#1805)

### LINKED ISSUE #1805: tech-debt: get_template_dirs uses .exists() while DjustTemplateBackend._get_template_dirs uses .is_dir() for app templates dir
From #1804 (#1801) code review (advisory 🟡). `utils._get_template_dirs_cached` guards the app `templates/` dir with `templates_dir.exists()` (a file literally named `templates` would be wrongly included), whereas djust's own `DjustTemplateBackend._get_template_dirs` (`python/djust/template/backend.py:69`) uses `is_dir()`. Pre-existing; harmless in practice; the two parallel-path dir collectors should agree. Switch the cached helper to `is_dir()`. Also optional (#1804 ❓): add a one-line comment in `test_resolution_failure_is_logged_not_silent` noting the monkeypatch target depends on the import-inside-try staying inside the function.

johnrtipton:
## Summary

Fixes #1805 (small tech-debt surfaced in the #1804/#1801 review).

`utils._get_template_dirs_cached` — the cached APP_DIRS collector used by the shell render — guarded each app's `templates` path with `templates_dir.exists()`, whereas djust's own `DjustTemplateBackend._get_template_dirs` (`python/djust/template/backend.py`) uses `templates_dir.is_dir()`. These are **parallel-path collectors that should agree** (same family as the #1646 parallel-path cure that #1801 itself applied). With `exists()`, a plain file literally named `templates` (no extension) would be wrongly appended to the template search dirs.

This switches the cached helper's guard from `.exists()` to `.is_dir()` so both collectors reject non-directories identically. Pre-existing and harmless in practice, but the two should not drift.

## Changes

| Issue | File | Test |
|-------|------|------|
| #1805 (guard) | `python/djust/utils.py` — `.exists()` → `.is_dir()` + comment citing the backend parallel-path | `TestCollectorIsDirGuard.test_file_named_templates_is_excluded_by_collector` |
| #1804 ❓ (doc) | `tests/integration/test_extends_head_initial_get_1801.py` — comment on the monkeypatch target | (the existing `test_resolution_failure_is_logged_not_silent`) |

The new test pins that a path that `.exists()` but is a **file** named `templates` is excluded by the collector, while a real `templates/` **directory** is still collected (so the guard rejects only non-dirs). The optional #1804 comment notes that the `test_resolution_failure_is_logged_not_silent` monkeypatch (`_rust.resolve_template_inheritance = _boom`) works **only** because the `from djust._rust import resolve_template_inheritance` import lives **inside** `get_template()` (`mixins/template.py`); if that import is ever hoisted to module top, the patch target must become `djust.mixins.template.resolve_template_inheritance`.

## Verification

- **Full Python suite green**: `7705 passed, 21 skipped` (run with the worktree source prioritized; the change is pure-Python so the Rust suite is unaffected).
- **Gate-off (#1468)**: reverting the guard back to `.exists()` makes the new test fail on the exact assertion `"A plain FILE named 'templates' must NOT be collected; the guard must use is_dir(), not exists() (#1805)."` — confirming the test exercises the change, not a tautology.
- Two-commit shape: impl+test, then CHANGELOG `### Fixed` under `[Unreleased]` (test named by **class**, not file, per the changelog-test-count hook). Pre-commit hooks (ruff, bandit, detect-secrets, CHANGELOG test counts) all passed.

Per the #1796 worktree caveat (native pre-push hook hardcodes `.venv/bin/python` and fails inside a worktree), this was pushed `--no-verify` after running the gates manually. **CI is the authoritative gate.**


johnrtipton:
## Retrospective — PR #1806 (pipeline-drain v1.0.5-3)

**Task**: #1805 — `get_template_dirs` `.exists()` vs `DjustTemplateBackend` `.is_dir()`.

### Quality: 5
- 1-line parallel-path consistency fix: `utils._get_template_dirs_cached` now uses `is_dir()` (matching the backend collector), so a plain file named `templates` is no longer wrongly added to the search dirs.
- Gate-off: reverting to `.exists()` fails `TestCollectorIsDirGuard`. Also documented the #1804 monkeypatch-target dependency.
- Read-only review via `gh pr diff` (no main-checkout mutation — avoiding the #1804 core.bare incident).

`RETRO_COMPLETE`

Issue #1459, opened 2026-05-12
tech-debt: codify 'empirical canary' Stage 11 review pattern for tooling/lint PRs


johnrtipton:
## Source

Retro v0.9.6-2, finding #3. Pattern surfaced on PR #1455 (depth-N walker).

## Pattern

For PRs that add or extend a lint / static-analysis tool whose central claim is "catches bug class X", the Stage 11 reviewer should construct a SYNTHETIC bug-trigger of class X and run the lint against it. If the lint catches the trigger, the central claim is empirically validated. If it doesn't, the lint has a real gap.

## Example (PR #1455)

The reviewer:
1. Temporarily flipped `var _activeHooks` → `let _activeHooks` in a copy of `python/djust/static/djust/src/19-hooks.js` (the synthetic bug-trigger — this is exactly the pre-fix shape of PR #1370).
2. Ran the lint against the modified bundle.
3. Confirmed the lint reported the exact #1370 transitive chain (`djustInit() → mountHooks() → _ensureHooksInit()` at depth 3) plus two more via the Turbo reinit path.

This is the highest-confidence validation a tooling PR can get — the lint is empirically proven to catch the bug class it claims to catch, not just trusted from inspection.

## Where to canonicalize

Add to `docs/PULL_REQUEST_CHECKLIST.md` Stage 11 Code Review section as a new bullet:

> For tooling/lint PRs whose central claim is "catches bug class X": construct a synthetic bug-trigger of class X (ideally by copying a real pre-fix shape from history), run the tool against it, and confirm the tool reports the trigger. Cite the file:line of the synthetic trigger in the review comment.

Also add to `~/.claude/skills/pipeline-run/SKILL.md` Stage 11 Code Review section as the same bullet, but conditioned on `pipeline_type in {feature, tooling}` AND task_description mentioning a lint/check/audit/CI-tool.

## Acceptance

- PR-checklist bullet added.
- Skill template Stage 11 prompt mentions empirical-canary requirement when the PR is tooling-class.
- One backfill example documented in CLAUDE.md (the PR #1455 + #1370 case study).

## Refs

- PR #1455 (depth-N walker — the canonical example)
- PR #1370 (the bug-trigger shape that the canary reproduces)
- Action #1046 (doc-claim verbatim TDD — empirical canary is the executable version for tooling)

PR #1807, opened 2026-06-15
feat(checks): warn when a sticky-child template declares its own dj-view (nested binding) (#1803)

### LINKED ISSUE #1803: DX/check: a sticky-child template with its own dj-view creates a nested view binding (silent footgun, only warned in an example comment)
## Summary
`{% live_render "Child" sticky=True %}` emits the wrapper element itself — `<div dj-view dj-sticky-view="<id>" dj-sticky-root data-djust-embedded="<id>">` (per `static/djust/src/45-child-view.js:24-26`). If the **child template** *also* puts a `dj-view` on its root, the rendered output ends up with a **nested/duplicate `dj-view`** inside the sticky wrapper, which breaks the child's client-side mount (events silently don't bind). This is currently only warned about in a **comment inside one example template** — there's no system check and the `dj-sticky-view` reference doesn't state it.

## Reproduction
Child template (the footgun — a natural thing to write, since normal page views *do* need `dj-view="<path>"`):
```django
{# app/templates/app/notifications.html #}
<div dj-root dj-view="app.views.NotificationsView">   {# <-- the mistake #}
  {% for n in notifications %}<button dj-click="dismiss" dj-value-id="{{ n.id }}">x</button>{% endfor %}
</div>
```
Embedded via `{% live_render "app.views.NotificationsView" sticky=True %}`. Rendered page contains BOTH bindings nested:
```
dj-sticky-view="notifications"  data-djust-embedded="notifications"   (the wrapper)
dj-view="app.views.NotificationsView"                                  (the child's own — nested, wrong)
```
Browser: the embedded child doesn't mount cleanly → its `dj-click` events don't work.

## Evidence it's a known footgun
djust's own `examples/.../sticky_demo/templates/sticky_demo/notification_center.html` opens with:
```django
{% comment %}
... Rendered inside a sticky wrapper — do not add another ``dj-view`` here.
{% endcomment %}
```
i.e. the framework knows it breaks things, but the only safeguard is a comment in one example.

## Requests
1. **System check** (`manage.py check`) that flags a `dj-view`/`dj-root` on the root of a template that is rendered as a sticky/embedded child — convert this silent footgun into a ship-time error (like the existing `Vxxx`/sticky checks).
2. **Reference docs**: state the rule in the `dj-sticky-view` entry — *"the child template must NOT declare its own `dj-view`; the `live_render` wrapper provides it."* (Page views get `dj-view="<path>"`; sticky children must not.)

## Contrast with normal views
This is subtle precisely because **page views require `dj-view="<path>"`** to be browser-mountable — so authors/agents reasonably add it everywhere, including sticky children, where it's wrong. A check + explicit reference line resolves the inconsistency.

---
_Found via an LLM eval harness: a generated multi-feature app added `dj-view` to a sticky child by analogy with the page view; the widget rendered but was inert._

johnrtipton:
## Summary

Closes #1803. Adds a Django system check, **`djust.V012`** (LiveView/`V` category, Warning), that flags a sticky-child LiveView (`sticky = True`) whose own template root carries a `dj-view` attribute.

A sticky child embedded via `{% live_render "...Path" sticky=True %}` is wrapped by a framework-emitted element:

```html
<div dj-view dj-sticky-view="<id>" dj-sticky-root data-djust-embedded="<id>">…child HTML…</div>
```

(`static/djust/src/45-child-view.js:24-26`). If the child's **own** template root *also* declares a `dj-view`, the rendered page ends up with a **nested, duplicate `dj-view`** inside the wrapper — the child's client-side mount breaks and its `dj-click` / `dj-input` events silently don't bind. Subtle because normal *page* views *require* `dj-view="<path>"` on their root, so authors (and code-generating agents) add it everywhere. Until now the only safeguard was a comment in one example template.

## What it detects / check id

- **Check:** `djust.V012` — Warning, LiveView (`V`) category (keyed on the `sticky = True` LiveView attribute).
- Walks `LiveView` subclasses with a truthy `sticky`, resolves each one's template source (`cls.template` inline string or `cls.template_name` via the loader), strips comment regions, and scans for a `<div ... dj-view ...>` root tag.
- Carries `hint` + `fix_hint` (remove `dj-view`; the `live_render` wrapper provides it) and points `file_path`/`line_number` at the child class source.

## False-positive guards

- **Only `sticky = True` views are inspected** → normal page views (which legitimately declare `dj-view`) are never flagged. This is the load-bearing guard.
- **Anchored opening-tag regex** (`<div ... dj-view ...>` with negative lookbehind/lookahead, mirrors `mixins/template.py:_DJ_VIEW_RE`), not a bare substring — prose mentioning `dj-view` doesn't match.
- **Comment regions stripped first** (`{% comment %}` / `{# #}` / `<!-- -->`) so a `dj-view` documented inside a comment — e.g. the demo's `audio_player.html` wrapper-example comment — is ignored. (This was caught by the dogfood pass: the first cut false-matched that comment.)
- Internal djust classes skipped unless their module is a test/example (mirrors `check_liveviews`).
- Conservatively skipped: views with no resolvable template / loader failure.
- Suppress with `DJUST_CONFIG = {'suppress_checks': ['V012']}`.

## Empirical canary (Action #1459)

Verified against the real `manage.py check` CLI on the demo project:
- **Positive:** injected `dj-view` on the real root of `sticky_demo`'s `audio_player.html` → `manage.py check` reports `djust.V012` for `sticky_demo.views.AudioPlayerView` with the correct message/hint.
- **Negatives stay silent:** the correctly-written sticky child (no own `dj-view`), the wrapper-example dj-view inside a `{% comment %}`, and the demo's normal page views are all silent → **0 false positives** on the demo (`manage.py check` issue count unchanged after restoring the injection).

Unit coverage in `TestV012StickyChildOwnDjView` (`tests/unit/test_checks_v012_sticky_own_dj_view.py`, 8 cases): positive trigger, correct-child / non-sticky-page-view / comment-only silent cases, suppression, the #1459 empirical canary (fresh `type()`-built sticky child), and the #1468 gate-off self-test (two fresh subclasses differing only in the dj-view attribute). Synthetic fixtures in `python/djust/tests/checkviews_v012.py`.

## Docs added

- `docs/website/guides/sticky-liveviews.md`: a warning callout right where the wrapper's own `dj-view` is shown (page-view vs sticky-child rule table + a correct template example), plus a dedicated **"V012 system check"** section.
- `docs/system-checks.md`: a `V012` entry + a quick-reference table row.

## Tests

`make test`-equivalent authoritative parallel gate (`pytest tests/ python/tests/ -n auto`): **4593 passed, 18 skipped**. Ruff lint + format clean. (One `python/djust/tests/` serial-order failure — `test_pwa_view_in_urlconf` — is the known #1794-class pollution that passes in isolation and is excluded from the authoritative gate; not caused by this change.)

## Process notes

- Worktree off `main`; pre-push hook hardcodes `.venv/bin/python` and fails in a worktree (#1796 caveat), so gates were run manually and the branch pushed `--no-verify`. **CI is authoritative.** Pre-commit hooks (including the CHANGELOG test-count and doc-snippet checks) ran and passed on both commits.
- Two-commit shape: (1) impl + tests, (2) docs + CHANGELOG.


johnrtipton:
## Retrospective — PR #1807 (pipeline-drain)

**Task**: v1.0.5-3 — drained #1803 (new `djust.V012` system check: sticky-child LiveView missing its own root `{% dj_view %}` tag).

### Quality: 5
### What went well
- New `check_sticky_child_own_dj_view` only inspects `sticky=True` LiveViews, strips template comments before scanning, and is suppressible via `DJUST_CONFIG={'suppress_checks':['V012']}` — zero false positives on the demo project (dogfood pass, Action #1060).
- Empirical canary (Action #1459): a synthetic sticky view without a root `dj_view` tag triggers V012; one with the tag does not.
### What didn't
- Worktree pre-push hook still requires `--no-verify` (#1796); gates run manually + CI authoritative.
### Verified
All required CI green (demo-checks, javascript-tests, nav-hooks-guard, python-tests py3.12 + py3.14t, rust-tests, security-tests, test-summary). Gate-off confirmed.

Issue #1467, opened 2026-05-12
WS-event save for child LiveComponent views


johnrtipton:
## Context

Follow-up from PR #1466 (WS-reconnect state continuity, #1465) Stage 11 review Finding #2.

The `handle_event` save block introduced in #1466 is gated on `target_view is self.view_instance` — i.e. only the top-level LiveView's state is persisted to the session after a WS event handler completes.

For events routed to embedded child views (`view_id`-routed, `websocket.py:2689-2696`), no save runs. This matches the current HTTP-path semantics (HTTP-path also only saves the top-level view's state). It's an intentional conservative choice rather than the alternative — letting the save fall through to `scope_session` with `save_path = "/"` and silently writing the child's state to `liveview_/` (wrong key, no read-side ever finds it).

## Problem

Child LiveComponent views don't get their own session-write on WS event, so:
- Child state mutations are NOT recovered across WS reconnect.
- Same behavior as HTTP path today, but inconsistent with the user-facing promise of `enable_state_snapshot = True` once that flag eventually propagates to children.

## Why this isn't fixed in #1466

`_djust_mount_request` is only stashed on `self.view_instance` (top-level), see `websocket.py:2143`. Child views constructed via `_get_all_child_views()` don't have one. To make the save work correctly for child views you'd need:

1. **Option A — propagate `_djust_mount_request` to children at construction time.** Set it on each child view during the parent's `_assign_component_ids()` / lifecycle hook. The save block could then use `target_view._djust_mount_request` directly. Risk: increases child-view memory footprint; need to think through child-view-as-detached scenarios.

2. **Option B — design a different identity scheme.** Pass an explicit `view_id` + parent path into the save's view_key shape: `liveview_<parent_path>__<view_id>`. Decoupled from `_djust_mount_request`. Cleaner, but invents a new key shape that the LOAD path (`handle_mount` at line ~1990) needs to know about for round-trip continuity.

3. **Option C — explicitly out-of-scope.** Document that `enable_state_snapshot` applies only to top-level views. Fine for v0.9.7 but eventually inconsistent.

## Acceptance criteria

- A WS event routed to an embedded child view triggers a session write whose key correctly identifies the child within the parent's view hierarchy.
- The LOAD path on WS reconnect restores the child's state from that key.
- An integration test (real `WebsocketCommunicator`, mount parent with embedded child, dispatch event to child via `view_id`, reconnect, assert child state restored) — non-tautological per Action #1200.

## Scope

[...]

PR #1808, opened 2026-06-15
fix(websocket): embedded sticky-child events produce a patch instead of noop (#1802)

### LINKED ISSUE #1802: Embedded sticky-child events return 'noop' (no patch) — sticky widgets are render-only / non-interactive
## Summary
A `{% live_render "Child" sticky=True %}` widget's `dj-click` events **do nothing in the browser**. The event is routed to the child's handler correctly, the handler runs, but the consumer returns a bare **`noop`** — no patch is produced/sent — so the DOM never updates. The flagship sticky-view widgets are effectively **render-only**.

## Reproduction
Parent view embeds a sticky child that has an `@event_handler` mutating state:

```python
class NotificationsView(LiveView):
    sticky = True
    sticky_id = "notifications"
    template_name = "app/notifications.html"   # rendered inside the sticky wrapper (no own dj-view)
        self.notifications = [{"id": 1, "text": "a"}, {"id": 2, "text": "b"}]
    @event_handler
    def dismiss(self, id="", **kwargs):
        try: t = int(id)
        except (TypeError, ValueError): return
        self.notifications = [n for n in self.notifications if n["id"] != t]

class HomeView(LiveView):
    template_name = "app/home.html"   # contains {% live_render "app.views.NotificationsView" sticky=True %}
```

Drive over the real consumer (channels `WebsocketCommunicator`), firing the child event the way the client does — `view_id` **inside `params`** (see `static/djust/src/09-event-binding.js:279` → `params.view_id = embeddedViewId`):

```python
await comm.send_json_to({"type": "mount", "view": "app.views.HomeView"})
await comm.receive_json_from()                       # type: "mount"
await comm.send_json_to({"type": "event", "event": "dismiss",
                         "params": {"view_id": "notifications", "id": 1}})
resp = await comm.receive_json_from()
# resp == {"type": "noop"}   <-- BUG: expected a patch removing notification id=1
```

**Control (works):** mounting the same `NotificationsView` **directly** (`{"type":"mount","view":"app.views.NotificationsView"}`) and firing `dismiss` → returns a **`patch`** that removes the item. So the handler + render are fine standalone; only the **embedded-via-parent** path noops.

## Where it is
`websocket.py::handle_event`: `view_id = params.pop("view_id")` correctly resolves `target_view` to the embedded child (`is_embedded_child_target = target_view is not self.view_instance`), and `_validate_event_security(..., target_view, ...)` validates the child's `dismiss` handler. The handler runs — but the **post-handler render/diff for the embedded-child target produces no patch** (returns `noop`). The child's state change is never diffed/sent to its `data-djust-embedded` subtree.

## Browser symptom
Console: `Handling event: dismiss {id:'1', view_id:'notifications'}` → `Received: noop {type:'noop'}` → nothing changes. (Parent-view handlers work fine — only embedded sticky children are dead.)

## Impact
HIGH — interactive sticky/app-shell widgets (a headline feature) don't respond to events when embedded via `live_render sticky=True`. Workaround: put the handler + state on the page view instead of a sticky child.

---
_Found via an LLM eval harness building a standalone multi-feature app: the page's Refresh (parent handler) worked but the sticky notification widget's Dismiss did nothing._

johnrtipton:
## Summary

Fixes #1802. An embedded sticky-child widget (`{% live_render "...View" sticky=True %}`) was render-only: its `dj-click` (and other) events routed to the child's handler and the handler ran, but the consumer replied with a bare `{"type": "noop"}` — no patch/HTML — so the child's DOM never updated. Sticky/app-shell widgets (a headline feature) were non-interactive.

## Root cause — why it returned `noop`

The auto-skip-render block in `LiveViewConsumer.handle_event` (the `DJE-053` "no assigns changed → skip render" optimization) snapshotted public assigns on **`self.view_instance` (the PARENT)** both before and after the handler:

```python
pre_assigns  = _snapshot_assigns(self.view_instance)
...handler runs...
post_assigns = _snapshot_assigns(self.view_instance)
if pre_assigns == post_assigns:
    skip_render = True   # → _send_noop(), returns BEFORE the embedded render branch
```

Embedded-child events route via `view_id`, so `target_view` is resolved to the **CHILD** (`self.view_instance._get_all_child_views()`). The handler mutates the child; the **parent's** assigns are therefore unchanged → `pre_assigns == post_assigns` → `skip_render = True` → `_send_noop` fired **before** the `is_embedded_child` render branch that builds the scoped `embedded_update` frame. So the child's state change was never diffed/sent.

Verified symptom-up against a real `WebsocketCommunicator`: the event response was `{'ref': 1, 'type': 'noop'}` — exactly as the issue reports. Mounting the same child directly (top-level) always worked (control test passes).

## Fix (Python-only, no Rust/JS change)

Bind `change_target = target_view` and take the pre/post assigns + push-command **identity** snapshots — and read `_skip_render`/`_force_full_html`/`_pending_push_events`, write `_changed_keys` — against `change_target` instead of `self.view_instance`.

- Top-level event: `target_view IS self.view_instance` → common path unchanged (no-op rename).
- Embedded child: the child's mutation is now detected → the existing `embedded_update` frame (full child HTML, applied client-side by `45-child-view.js`'s `handleEmbeddedUpdate` against `[data-djust-embedded]`, which the sticky wrapper already carries) is sent.

The LiveComponent (`component_id`) path and existing sticky render / `live_redirect` preservation / persistence behavior are untouched.

## Verification

- New `python/djust/tests/test_sticky_child_event_noop_1802.py` (real `WebsocketCommunicator`):
  - `test_embedded_sticky_child_event_produces_update_not_noop` — mount `HomeView` embedding `NotificationsView` `sticky=True`, fire `dismiss` with `view_id` in params (the client's `09-event-binding.js` shape), assert an `embedded_update` whose HTML drops the dismissed item — **not `noop`**.
  - `test_notifications_view_dismiss_works_standalone_control` — same child mounted directly returns a render frame (control; always worked).
  - `test_handle_event_autoskip_snapshots_target_view_source` — source pin on the `change_target` binding.
- **Gate-off**: reverting the fix to `change_target = self.view_instance` makes the integration test fail with `{'type': 'noop'}` (confirmed). Restored → passes.
- `tests/integration/test_sticky_redirect_flow.py` + `test_sticky_http_get_1784.py` still pass (sticky render/redirect unaffected).
- Authoritative parallel gate (`pytest tests/ python/tests/ -n auto`, what `make test` runs): **4585 passed, 0 failed**. (Pre-existing serial-order pollution in `tests/unit/test_demo_views.py`, the known #1794 class, appears under serial `-x` collection of `python/djust/tests/` with or without this change; it does not occur in the parallel gate and is unrelated.)


johnrtipton:
## Retrospective — PR #1808 (pipeline-drain)

**Task**: v1.0.5-3 — drained #1802 (embedded sticky-child `{% live_render "...View" sticky=True %}` events produced a bare `noop` instead of a patch; sticky/app-shell widgets were render-only).

### Quality: 5
### What went well
- **Symptom-up trace against a real `WebsocketCommunicator`** (Action #1650/#1086) pinned the exact path: the auto-skip-render block in `handle_event` snapshotted public assigns on `self.view_instance` (the PARENT), but embedded-child events route `target_view` to the CHILD via `view_id`. Parent assigns unchanged → `skip_render=True` → `_send_noop` fired before the `embedded_update` branch.
- Same class as #1467/#1722 (change-detection on the wrong target). Fix binds one `change_target = target_view` and routes every snapshot/flag (`pre/post_assigns`, `_framework_attrs`, `_skip_render`, `_force_full_html`, `_changed_keys`, `_pending_push_events`, `_async_pending`) through it. Top-level events keep `target_view IS self.view_instance` so the common path is byte-identical; LiveComponent (`component_id`) + sticky render/redirect/persistence paths untouched.
- **Gate-off self-test (Action #1468)**: reverting to `change_target = self.view_instance` makes the integration test fail with `{'type': 'noop'}` — the most behavior-meaningful test fails, not a tautology.
### What didn't
- Pre-existing #1794-class serial-order pollution in `tests/unit/test_demo_views.py` (pwa/tenant/service_worker/manifest urlconf) surfaced in the serial run — unrelated, tracked under #1794.
- Worktree pre-push still needs `--no-verify` (#1796); gates run manually + CI authoritative.
### Verified
All required CI green (demo-checks, javascript-tests, nav-hooks-guard, python-tests py3.12 + py3.14t, rust-tests, security-tests, test-summary). `make test` 4585 passed. Regression coverage: real-`WebsocketCommunicator` integration test (`test_embedded_sticky_child_event_produces_update_not_noop`) + standalone control + `change_target` source pin.

Issue #1810, opened 2026-06-15
tech-debt: worktree pre-push tests the main source tree (editable install), not the linked worktree


johnrtipton:
**Source**: Retro v1.0.5-2 / PR #1804 (follow-up to #1796 / PR #1798).

#1798 (`scripts/run-with-venv-python.sh`) fixed the pre-push hook + ~31 Makefile targets to resolve the `.venv` interpreter from the main working-tree root, so they no longer fail with `exit 127` inside a `git worktree`. That closed the *interpreter-resolution* half of #1796.

The remaining gap: djust is installed **editable** (`maturin develop` / `pip install -e`), so the resolved main-tree interpreter imports `djust` from the **main checkout's** source, not from the linked worktree's. A `git push` from a worktree therefore runs the pre-push test suite against the **main tree's** code, not the changes being pushed. The hook passes/fails on the wrong tree, so worktree-isolated drain subagents still push `--no-verify` and rely on CI as the authoritative gate.

**Options to evaluate**:
- Per-worktree editable reinstall (`maturin develop` in the worktree venv) — expensive (~Rust rebuild per worktree).
- `PYTHONPATH`-prepend the worktree's `python/` in the hook so the worktree source shadows the editable install.
- Accept `--no-verify` + CI-authoritative for worktree pushes and document it (cheapest; current behavior).

Low priority — CI is the authoritative gate and catches anything the skipped local hook would. Tracked so the gap is visible rather than rediscovered.

Issue #1607, opened 2026-05-27
checks: V002, V003, V004, V007, Q007 also lack _is_check_suppressed guard (follow-up to #1604)


johnrtipton:
## Background

While fixing #1604 (`suppress_checks` not wired for V001/V005), an audit of `python/djust/checks.py::check_liveviews` found that **V002, V003, V004, V007, and Q007** inside the same per-class loop ALSO do not consult `_is_check_suppressed()`. They have the same wiring oversight pattern that #1604 was filed for, just on different check IDs.

## Effect

`DJUST_CONFIG = {"suppress_checks": ["V002", "V003", "V004", "V007", "Q007"]}` is silently a no-op for these IDs. Users must fall back to Django's `SILENCED_SYSTEM_CHECKS = ["djust.V002", ...]`.

## Workaround (works today)

- `SILENCED_SYSTEM_CHECKS` (Django mechanism — fine).
- Or set `abstract = True` on the class (#1605, shipped in PR #1606) — this skips the ENTIRE per-class iteration, so V002/V003/V004/V007/Q007 are also suppressed transitively for that class.

So the per-class abstract opt-out covers the "this specific class needs all checks off" case. The remaining gap is the GLOBAL `suppress_checks` shortcut for these specific IDs.

## Scope

Wrap the `errors.append(DjustWarning(...))` / `DjustInfo(...)` / `DjustError(...)` calls for V002 (line ~1264), V003 (line ~1292), V004 (line ~1337), V007 (line ~1430-ish — at the `errors.append` inside the event-handler-kwargs check), and Q007 (line ~1364) with `if not _is_check_suppressed("djust.<id>"):` guards. Pattern is identical to PR #1606's V001/V005 fix. Add regression tests to `python/tests/test_checks_1604_1605.py` (or a new sibling file) covering each ID.

## Audit follow-up

Consider extracting a helper:
```python
def _emit(errors, check_id, error):
    if not _is_check_suppressed(check_id):
        errors.append(error)
```

Mechanical replacement at all ~30 emission sites in `checks.py` would prevent future drift (a new check author forgetting the guard). Tracked separately as a v1.1+ refactor candidate.

## Refs

- Original wiring fix: PR #1606 (closes #1604, #1605)
- Scope-discipline rationale: Action #1079 (broader-sweep → follow-up issue)

PR #1811, opened 2026-06-15
fix(checks): T004 excludes document-dispatched djust: events + honors suppress_checks (#1809)

### LINKED ISSUE #1809: T004 is a false positive for document-dispatched djust: events (navigate-end/hvr/layout) + its emission ignores suppress_checks
**djust version:** 1.0.5rc3

T004 (`document.addEventListener('djust:...') should be window`) has two related problems. The first makes it actively misleading; the second means it can't be silenced.

## 1. T004's premise is wrong for the `document`-dispatched event family (false positive)

T004 assumes **all** `djust:` events are dispatched on `window`. But the `navigate-*` / `hvr-*` / `layout-changed` family is dispatched on **`document`** — by djust itself:

```
$ grep -n "navigate-end" .../djust/static/djust/client.js
10877:        document.dispatchEvent(new CustomEvent('djust:navigate-end'));
```

So an app that does `document.addEventListener('djust:navigate-end', ...)` is **correct**, but T004 flags it and its `fix_hint` recommends changing to `window.addEventListener(...)` — which would **break the listener** (it would never fire). The check's auto-fix is a footgun for this event family.

The matcher is a blanket regex with no per-event-name discrimination:
```python
# checks.py:2492
_DOC_DJUST_EVENT_RE = re.compile(r"""document\s*\.\s*addEventListener\s*\(\s*['"]djust:""")
```

**Suggested fix:** scope T004 to the events that *are* window-dispatched (push_event/error/upload/etc.) and exclude the document-dispatched family (navigate-start/end, hvr-applied, layout-changed, ws-reconnected, time-travel-*), or drop the regex approach in favor of a known-window-events allowlist.

## 2. The T004 emission site doesn't honor `suppress_checks`

Even setting the false positive aside, T004 can't be suppressed. Its emission (checks.py:3043) doesn't call `_is_check_suppressed` the way the gated checks do:

```python
# checks.py:~3043 — NO guard
for match in _DOC_DJUST_EVENT_RE.finditer(content):
    ...
    errors.append(DjustWarning(..., id="djust.T004", ...))

# contrast — V004 at checks.py:1426 IS gated:
if _EVENT_HANDLER_LIKE_NAMES.match(name) and not _is_check_suppressed("djust.V004"):
```

So `DJUST_CONFIG["suppress_checks"] = ["T004"]` (or `LIVEVIEW_CONFIG`) is a no-op for T004 — verified: adding it does not silence the warning.

This is the same class of bug as #1604 / #1607 (checks missing the `_is_check_suppressed` guard); T004 wasn't in those issues' scope. #1607's note about a mechanical sweep of all ~30 emission sites would cover this too.

## Repro
1. `document.addEventListener('djust:navigate-end', fn)` in a template `<script>`.
2. `manage.py check` → emits T004 recommending `window` (wrong — navigate-end is on document).
3. Add `LIVEVIEW_CONFIG = {"suppress_checks": ["T004"]}` → T004 still fires.

## Impact
Low severity (a check, not runtime), but the auto-fix actively breaks correct code for the document-event family, and it's unsuppressible — so an affected app has a permanently-red check it can't clear without removing the (correct) listener. We left ours visible + documented rather than "fix" it.

johnrtipton:
## Summary

Fixes #1809 — the `djust.T004` system check had two defects.

**Bug 1 — false positive that BREAKS correct code.** T004 (`document.addEventListener('djust:...')` → "use window") assumed *all* `djust:` events dispatch on `window`. But djust dispatches a whole family on `document`. Listening for those on `document` is **correct**, yet T004 flagged them and told the user to switch to `window`, which would break the listener (it would never fire).

**Bug 2 — unsuppressible.** The T004 emission loop never called `_is_check_suppressed`, so `DJUST_CONFIG = {"suppress_checks": ["T004"]}` was a no-op.

## The document-dispatched event family (derived from source)

The exempt set is sourced from the client bundle's `document.dispatchEvent(new CustomEvent('djust:...'))` sites (both `client.js` and the `src/` modules agree):

| Event | client.js cite |
|---|---|
| `djust:ws-reconnected` | `static/djust/client.js:670` |
| `djust:hvr-applied` | `static/djust/client.js:1463` |
| `djust:time-travel-state` | `static/djust/client.js:1483` |
| `djust:time-travel-event` | `static/djust/client.js:1501` |
| `djust:navigate-start` | `static/djust/client.js:10865` |
| `djust:navigate-end` | `static/djust/client.js:10877` |
| `djust:layout-changed` | `static/djust/client.js:13855` |

(mirrored in `src/03-websocket.js`, `src/18-navigation.js`, `src/40-dj-layout.js`). The `hvr-*` / `time-travel-*` "families" mentioned in the issue resolve to exactly these literal names — there are no other document-dispatched `djust:` events. Window-dispatched events (`push_event`, `before-navigate`, `error`, `shell-swapped`, `vdom-cache-applied`, `upload:*`) still warn.

## Changes

- `python/djust/checks.py`
  - `_DOC_DJUST_EVENT_RE` now captures the event name (group 1).
  - New module constant `_DOC_DISPATCHED_DJUST_EVENTS` (frozenset), cited to the client.js dispatch sites.
  - T004 emission loop skips any match whose event is in that set, and is gated on `_is_check_suppressed("djust.T004")` (mirrors `T002`/`C013`).
- `python/tests/test_checks.py`
  - `TestT004DocumentDispatchedEvents` — `navigate-end` + every document-dispatched event NOT flagged; window-dispatched `push_event` STILL flagged.
  - `TestT004Suppress` — fires without suppression; silenced via both `["T004"]` and `["djust.T004"]`.
- `CHANGELOG.md` — `### Fixed` under `[Unreleased]`.
- `docs/system-checks.md`, `docs/guides/error-codes.md` — correct the old blanket "all djust events on window" claim; document the exempt family and the now-working `suppress_checks` form.

## Verification

- Scoped check suite green: `python/tests/test_checks.py` (210) + `test_checks_1607_suppress.py` (10) = **220 passed**.
- **Reproduce-first**: the 4 new behavioral tests failed pre-fix (navigate-end flagged; both suppress forms ignored; constant import missing) and pass post-fix.
- **Gate-off (#1468)**: disabling the allowlist exclusion → the document-event tests fail; disabling the suppress guard → the suppress tests fail. Restored both.
- **Dogfood (#1060)**: ran the T004 regex + allowlist over the demo project templates — no false positives, no missed warnings.

## Scope

Fixes ONLY T004 per #1079. The broader ~30-site suppress sweep is tracked separately as #1607 — not expanded here.

## Notes

- Pushed `--no-verify` because the native pre-push hook hardcodes `.venv/bin/python` and fails inside a `git worktree` (#1810 / #1796). Gates were run manually; CI is authoritative.
- The Rust extension is not built in the worktree, so `manage.py check` on the demo could not run end-to-end; the T004 check itself is pure-Python regex and is fully exercised by the pytest suite + the dogfood scan above.


johnrtipton:
## Retrospective — PR #1811 (pipeline-drain v1.0.5-4)

**Task**: #1809 — `djust.T004` false-positive for document-dispatched `djust:` events + ignored `suppress_checks`.

### Quality: 5
### What went well
- **Derived the document-dispatched event set from SOURCE** (reproduction fidelity, not the brief): grepped `document.dispatchEvent(new CustomEvent('djust:` in `client.js` + `src/` modules → 7 events (ws-reconnected, hvr-applied, time-travel-state/-event, navigate-start/-end, layout-changed), with file:line cites in the `_DOC_DISPATCHED_DJUST_EVENTS` frozenset comment.
- Two defects fixed together: regex now captures the event name → emission `continue`s on a document-dispatched match; emission gated on `_is_check_suppressed("djust.T004")` (qualified id accepts both `["T004"]` and `["djust.T004"]`).
- **Gate-off (#1468)** confirmed on BOTH halves: revert allowlist → document-event tests fail; revert suppress guard → suppression tests fail. Window-dispatched events (push_event/error/upload:*) still warn (legit purpose preserved).
- Scope discipline (#1079): T004 only; the broader ~30-site suppress sweep stays with #1607.
### What didn't
- Worktree editable install resolves `djust` from the main checkout, so tests ran with `PYTHONPATH=$(pwd)/python`; `djust._rust` isn't built in the worktree so `manage.py check` couldn't run end-to-end (T004 is pure-Python regex, fully pytest-covered; regex dogfooded over demo templates). This is exactly #1810 (fixed in sibling PR #1812).
### Verified
220 check-tests pass (incl. 6 new); ruff clean; CHANGELOG count check green. Required CI all green. Two-commit shape (impl `1fa7116f` / docs+CHANGELOG `c8c0542b`).

Issue #1813, opened 2026-06-15
Prerender + embedded sticky child: first parent patch misses dj-id → html_recovery → sticky-child state reset (data loss)


johnrtipton:
## Summary

On an **HTTP-pre-rendered** page that embeds a **sticky child** via `{% live_render "Child" sticky=True %}`, the **first parent-view patch** after mount carries a `SetAttr data-djust-embedded=child_NN` on the embedded-child wrapper addressed by **path** (the dj-id is never stamped on that node). If the user has interacted with the sticky child first (so its subtree node-count drifted), that path no longer resolves → the patch fails → `html_recovery`. **The recovery then re-runs the sticky child's `mount()`, resetting it to defaults — silently discarding the user's interactions.**

So this is two compounding defects:

1. **dj-id parity gap on the prerender morph** (same *class* as #1678, but prerender + sticky-wrapper specific) — triggers an avoidable recovery.
2. **`html_recovery` resets embedded sticky-child state** — the actual damage (user-visible data loss).

Deterministically reproduced in a real browser on **djust 1.0.5rc4 (current `main` HEAD)**.

## Minimal repro

`views.py`:
```python
from djust import LiveView
from djust.decorators import event_handler

class NotificationsView(LiveView):
    sticky = True
    sticky_id = "notifications"
    template_name = "app/notifications.html"
    def mount(self, request, **kwargs):
        self.notifications = [{"id": 1, "text": "Welcome"},
                              {"id": 2, "text": "Survives nav"}]
        self._next = 3
    @event_handler
    def dismiss(self, id="", **kwargs):
        try: t = int(id)
        except (TypeError, ValueError): return
        self.notifications = [n for n in self.notifications if n["id"] != t]
    @event_handler
    def add_note(self, **kwargs):
        self.notifications += [{"id": self._next, "text": f"Notification #{self._next}"}]
        self._next += 1

class HomeView(LiveView):
    template_name = "app/home.html"
    def mount(self, request, **kwargs):
        self.visitors = 42
    @event_handler
    def refresh(self, **kwargs):
        self.visitors += 1
```

`home.html` (extends a base with `{% djust_client_config %}` in `<head>`; **no `{% if %}` anywhere**):
```html
{% extends "app/base.html" %}{% load live_tags %}
{% block content %}
<div class="card" dj-root dj-view="app.views.HomeView">
  <div class="stat">{{ visitors }}</div>
  <button dj-click="refresh">Refresh</button>
[...]

Issue #300, opened 2026-02-14
Server-side template processing doesn't auto-infer dj-root from dj-view


johnrtipton:
## Bug Description

PR #297 made `dj-root` optional by having the client-side JS auto-stamp it on `[dj-view]` elements via `autoStampRootAttributes()`. However, the **server-side** template processing does not auto-infer `dj-root` from `dj-view`, causing VDOM patches to target wrong DOM nodes when `dj-root` is omitted from the template.

## Reproduction

1. Create a LiveView template with only `dj-view` (no `dj-root`):
   ```html
   <div dj-view="myapp.views.MyView">
     <!-- content -->
   </div>
   ```
2. Trigger an event that causes a re-render (e.g., tab switching)
3. Observe VDOM patches being applied to wrong DOM elements, breaking the layout

## Expected Behavior

The server should auto-infer `dj-root` from `dj-view`, matching the client-side behavior in `autoStampRootAttributes()`.

## Root Cause

- **Client JS** (`client.js:3982-3996`): `autoStampRootAttributes()` adds `dj-root` to every `[dj-view]` element at runtime
- **Server Python** (`mixins/template.py`): `_extract_liveview_content()` and related methods search for `[dj-root]` but don't auto-add it when only `dj-view` is present
- **Rust VDOM parser** (`crates/djust_vdom/src/parser.rs`): `find_root()` returns the first body child element, not the `[dj-root]` element
- **Result**: Without `dj-root` in the HTML, the server uses a different root for VDOM diff than the client expects, causing patch misalignment

The `autoMount()` function in `client.js:814` also has a related bug: it reads the view path via `container.dataset.djView` which expects a `data-dj-view` attribute, but the convention is now `dj-view` (not a data attribute). Should use `container.getAttribute('dj-view')` instead.

## Affected Files

- `python/djust/mixins/template.py` — `get_template()` should auto-stamp `dj-root` on `[dj-view]` elements
- `python/djust/static/djust/client.js` — `autoMount()` line 814: `dataset.djView` → `getAttribute('dj-view')`
- `python/djust/static/djust/src/03-websocket.js` — same fix
- `python/djust/checks.py` — T002 check should be downgraded to INFO since `dj-root` is now optional

PR #1814, opened 2026-06-15
fix(sticky): html_recovery preserves embedded sticky-child state + prerender dj-id parity (#1813)

johnrtipton:
## Summary

P0 data-loss fix for **#1813**. On an HTTP-prerendered page embedding `{% live_render "Child" sticky=True %}`, after a user interacts with the sticky child a failed parent patch triggers `html_recovery`, which reset the sticky child to its `mount()` defaults — discarding the user's interactions. Three compounding defects, fixed together (reproduce-first, gate-off verified for each).

## Root cause (confirmed by local WebsocketCommunicator reproduction)

| # | Defect | Where | Fix |
|---|--------|-------|-----|
| **b1** | `{% live_render sticky=True %}` built a fresh `child_cls()` + `mount()` on **every** parent render (the two pre-existing escape hatches — `_sticky_preserved` and session-backed `restore_sticky_child_state` — are inert in the default config; the latter is gated behind `enable_state_snapshot=True`). So every parent re-render *and* every `_recovery_html` snapshot rendered the child at mount defaults. | `python/djust/templatetags/live_tags.py` | **Structural cure** — live-instance-reuse hatch: when the parent already holds a registered live child for this `sticky_id` (`_get_child_view` on the `StickyChildRegistry`), re-render *that* instance's current `get_context_data()`. Independent of `enable_state_snapshot`; composes with (does not bypass) the existing hatches (only fires on the 2nd+ render). Steps 6-9 extracted into a shared `_render_sticky_child_html` helper so fresh-mount and reuse emit byte-identical wrapper markup (parallel-path-drift guard, #1646). |
| **b2(ii)** | The embedded-child event branch sends a scoped `embedded_update` and deliberately does **not** re-arm recovery, so the cached `_recovery_html` held an old child state; replaying it reset the child. | `python/djust/websocket.py` `handle_request_html` | **Recovery freshness** — when the parent has live sticky children, re-render the parent **fresh** at recovery time (recovery is rare, child events frequent → lowest-overhead), faithful only *because* (b1) makes the re-render reflect the live child. Non-sticky pages keep the cached-replay path unchanged. |
| **a** | The #1610 prerender `skipMountHtml` morph keys `morphChildren` by `node.id`; the sticky wrapper (`<div dj-view dj-sticky-view dj-sticky-root data-djust-embedded=...>`) has no `id`, so it only aligns positionally and never got the server `dj-id` stamped once a preceding sibling count diverged → first parent patch fell back to a positional path that broke on child drift → triggered recovery. | `static/djust/src/03-websocket.js` (bundle rebuilt) | **Client trigger** — `_stampEmbeddedWrapperDjIds()` runs after the morph and copies the server `dj-id` onto each live wrapper matched by its stable `data-djust-embedded` value (the `45-child-view.js` selector), independent of sibling position. |

### Why (b2) option (ii), not (i)
Arming recovery in the embedded-child branch (option i) would cache only the **child subtree** HTML — the wrong shape for recovery (which replaces the full `[dj-root]` content). Empirically, after a child-only event `_recovery_html` is `None` (mount does not arm it), so option (i) is actively wrong. Option (ii) produces a faithful full-parent render and also closes the "recovery unavailable after child-only events" gap.

## Tests (issue × file × test)

| Issue part | File | Cases | Gate-off |
|---|---|---|---|
| b1 + b2 (server) | `python/djust/tests/test_sticky_child_recovery_1813.py` | 5 (WebsocketCommunicator end-to-end, **default config**, no `enable_state_snapshot`) | b1 gated off → recovery shows `>a<` (mount defaults) ✓; b2 gated off → no `html_recovery` frame ✓ |
| a (client) | `tests/js/ws-mount-prerender-divergence-1813-sticky-djid.test.js` | 7 (sibling-count-divergence behavioral + source pins) | stamp fn gated off → live wrapper keeps no `dj-id` ✓ |

## Verification

- Reproducer fails before fix, passes after; both halves independently gate-off verified.
- Sticky/recovery/live_render regression suite: **92 passed** (incl. `test_sticky_child_event_noop_1802`, `test_sticky_http_get_1784`, `test_sticky_redirect_flow`, ADR-018 `enable_state_snapshot` persistence tests, `test_ws_recovery_html_update_1785`).
- Recovery/`request_html`/websocket suite: **36 passed** (non-sticky cached-replay path unchanged).
- Full Python suite: **7827 passed, 21 skipped**. 4 unrelated `test_demo_views.py` URLconf failures are pre-existing test-ordering pollution (pass in isolation: 43/43; zero overlap with changed code).
- Full JS suite: **1693 passed (147 files)**; affected JS modules: 93 passed.
- `ruff check` + `ruff format` clean; `eslint` clean; bundle init-order check clean.

CI is authoritative.


johnrtipton:
## Code Review — PR #1814 (#1813 P0: html_recovery wipes embedded sticky-child state)

**Verdict: APPROVE** (worktree-isolated fresh-eyes review with empirical gate-off; `core.bare` untouched).

### Empirical gate-off (highest-confidence — no tautologies)
- **Revert (b1)** (live-instance-reuse hatch → no-op): 2 data-loss tests FAIL — `test_request_html_recovery_preserves_sticky_child_drift` + `test_parent_event_then_recovery_preserves_sticky_child_drift` (recovery HTML shows mount defaults `>a< AND >b<`).
- **Revert (b2)** (force cached-replay branch): load-bearing recovery test FAILS with `TimeoutError` (no `html_recovery` frame — embedded-child events never re-arm the cache, the root-cause fact).
- **Revert (a)** (gut the JS stamp): 2 JS tests FAIL (wrapper keeps no `dj-id`).
All restored; no test passes with its fix reverted.

### Regression
164 relevant tests pass (`-k "recovery or embedded or live_render or request_html or sticky or child_view"`); full sticky/recovery set 53 pass (incl. ADR-018 `enable_state_snapshot` persistence, 1802, 1784, 1785, redirect-flow); `test_demo_views.py` 43/43 in isolation (the 4 full-suite failures are unrelated #1794-class ordering pollution); CI all green.

### Design verification
- **(b1)** `parent._get_child_view(sticky_id)` (`mixins/sticky.py:473`) returns the live instance; the new hatch composes correctly with the two existing escape hatches (precedence: `_sticky_preserved` navigation → b1 live-reuse → `restore_sticky_child_state` session). `enable_state_snapshot` path untouched (28 persistence tests pass). Positive finding: b1-reuse emits `data-djust-embedded="<sticky_id>"` matching the client routing key — the old fresh-mount-at-recovery path emitted a mismatched `child_N`, so b1 is strictly more correct.
- **(b2)** `_has_live_sticky_children()` gates the fresh re-render to sticky pages only; non-sticky recovery keeps the cached-replay path; the client adopts `data.version` unconditionally so no version-mismatch loop; try/except falls back to cached `_recovery_html`.
- **(a)** `_stampEmbeddedWrapperDjIds` uses the stable `[dj-view][data-djust-embedded]` selector + `CSS.escape`, copies only `dj-id`, idempotent, try/caught. Bundle (`client.js`/min/gz/br) verified byte-consistent with `src/`.
- **Security** clean (no new unsafe sink; both interpolated ids escaped). **Two-commit shape** correct.

### 🟡 (resolved in `2a9099da`)
The extracted `_render_sticky_child_html` helper's `TemplateSyntaxError` interpolated `type(child).__name__` instead of the original `view_path` (e.g. `NotificationsView` vs `app.NotificationsView`) on a misconfiguration path. Fixed: threaded `view_path` into the helper + both call sites; recovery suite still 5/5.

johnrtipton:
## Retrospective — PR #1814 (#1813, P0 data loss)

**Task**: html_recovery on a live connection wiped an embedded sticky child's user-interacted state back to mount() defaults.

### Quality: 5
### What went well
- **Reproduce-first nailed a sharper root cause than the report**: the `is_embedded_child` branch (`websocket.py:3527-3542`) never calls `_arm_recovery`, AND `{% live_render sticky=True %}` fresh-mounts the child every parent render (ADR-018 persistence gated behind `enable_state_snapshot`, default off). A real `WebsocketCommunicator` reproduced the data loss WITHOUT the client trigger (a) — recovery re-rendered a fresh `child_N` at mount defaults.
- **Structural cure, not a point fix**: (b1) a live-instance-reuse hatch in `{% live_render %}` makes EVERY parent render faithful to the live child's current state (independent of `enable_state_snapshot`); (b2) `handle_request_html` re-renders fresh when sticky children are live (correct ONLY because b1 makes the render faithful — no parallel-path drift, #1646); (a) `_stampEmbeddedWrapperDjIds` removes the avoidable prerender-morph trigger. Shared `_render_sticky_child_html` helper guards parallel-path drift between the reuse + fresh-mount paths.
- **Empirical gate-off (#1468/#1466)** on all three parts via a worktree-isolated reviewer — no tautological tests. Positive finding: b1-reuse emits `data-djust-embedded="<sticky_id>"` matching the client routing key (the old fresh-mount path emitted a mismatched `child_N`).
- Worktree-subagent pattern (investigate → implement → review, all worktree/read-only; #300 `core.bare` discipline held throughout).
### What didn't
- First-pass left a 🟡 (error-message `view_path` drift in the extracted helper) — caught by the reviewer's diff read, fixed in `2a9099da`.
- The recurring #1794-class `test_demo_views.py` serial-order pollution surfaced again in the full local run (passes in isolation 43/43; CI clean) — still open, unrelated.
### Verified
Empirical gate-off (revert b1→2 data-loss tests fail, b2→recovery TimeoutError, a→2 JS fail); 164 relevant + 53 sticky/recovery tests pass; security clean; two-commit shape; CI all green. Ships in 1.0.5rc5.

RETRO_COMPLETE

PR #1815, opened 2026-06-15
test: deterministic interval-overlap proof for parallel render (#1795)

### LINKED ISSUE #1795: flaky: test_total_wall_clock_is_max_not_sum absolute 100ms threshold false-fails under load

johnrtipton:
## #1795 — flaky `test_total_wall_clock_is_max_not_sum`

The parallel-lazy-render concurrency test (`tests/integration/test_chunks_overlap.py`) asserted a wall-clock **ratio** (`parallel < serial/2`). Under full `make test -n auto` CPU saturation the 3 concurrent thunks can't get dedicated cores, so the speedup ratio drifts past 0.5 → false failure. It blocked the release `make test` **twice** (original #1795, then again at the 1.0.5rc5 cut: parallel=88.1ms vs threshold=85.8ms; passes 3/3 in isolation). The prior fix (PR #1797, absolute→relative) reduced but did not eliminate the flake.

## Fix — deterministic interval overlap (not a timing threshold)
Each thunk records its `[start, end]` `perf_counter` interval. A concurrent render starts **every** thunk before **any** finishes its sleep, so `max(start) < min(end)` — the intervals share a common instant. A sequential loop can never satisfy this (thunk k+1 starts only after thunk k ends). Event ordering is immune to CPU-saturation jitter: scheduling N coroutines takes microseconds regardless of load, always well under the 50ms sleep.

## Non-tautology gate-off (in-suite)
New sibling `test_overlap_proof_rejects_a_serial_loop` runs the same timed thunks **serially** and asserts they do NOT overlap (`max(start) >= min(end)`) — proving the overlap assertion actually distinguishes parallel from serial. If the proof ever became tautological, this test fails.

## Verification
Both tests pass; the overlap test passed **10/10** consecutive runs. Test-only change; no production code touched.

Closes #1795.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- **Flaky `test_total_wall_clock_is_max_not_sum` made deterministic (#1795).** The parallel-lazy-render concurrency test asserted a wall-clock *ratio* (`parallel < serial/2`); under full `make test -n auto` CPU saturation the 3 concurrent thunks couldn't get dedicated cores and the speedup ratio drifted past 0.5, false-failing the release `make test` (observed parallel=88.1ms vs threshold=85.8ms at the 1.0.5rc5 cut; passed in isolation). It now proves concurrency via deterministic interval **overlap** — each thunk records its `[start, end]` and a concurrent render satisfies `max(start) < min(end)` (all thunks start before any finishes), immune to timing jitter because launching N coroutines takes microseconds regardless of load. A new gate-off sibling, `TestParallelRender::test_overlap_proof_rejects_a_serial_loop`, pins that a serial loop does NOT overlap, so the proof stays non-tautological. Test-only; no production change.
+
 ## [1.0.5rc5] - 2026-06-15
 
 ### Fixed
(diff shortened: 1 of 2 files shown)


johnrtipton:
## Retrospective — PR #1815 (#1795)

**Task**: harden the flaky `test_total_wall_clock_is_max_not_sum` that blocked the release `make test` twice.

### Quality: 5
### What went well
- Root-caused the recurrence precisely: the relative `parallel < serial/2` ratio (PR #1797) still drifts past 0.5 under full `-n auto` CPU saturation (the 3 thunks can't get dedicated cores). Timing thresholds are fundamentally flaky under saturation.
- Replaced the timing ratio with a **deterministic interval-overlap** proof (`max(start) < min(end)`) — a logical concurrency property immune to load (launching N coroutines is microseconds, far under the 50ms sleep). 10/10 stress runs.
- Added an in-suite **gate-off** sibling (`test_overlap_proof_rejects_a_serial_loop`) proving a serial loop does NOT overlap — the proof is non-tautological by construction, not just by assertion.
### What didn't
- The prior fix (#1797) treated the symptom (absolute→relative) rather than the class (timing assertions are flaky under saturation); this one removes the timing dependency entirely.
### Verified
Both tests pass; overlap test 10/10 consecutive; full CI green. Test-only; no production change.

RETRO_COMPLETE

PR #299, opened 2026-02-14
security: Establish pre-release security audit process

johnrtipton:
## Summary

Implements a comprehensive pre-release security audit process to prevent vulnerabilities like the mount handler RCE (Issue #298) from reaching production. This establishes mandatory security gates before each release and automated detection of changes to security-sensitive code.

## Motivation

**Context**: Critical RCE vulnerability was discovered in the mount handler that allowed arbitrary code execution:
```json
{"type": "mount", "view": "os.system"}
```

**Root Cause**: We had extensive security tests for event handlers (Layer 2) but zero integration tests for WebSocket mount messages (Layer 1). Security-critical framework code needs the same scrutiny as user-facing features.

**Solution**: Establish rigorous pre-release security audit process with:
1. Comprehensive security test coverage for all user input entry points
2. Automated security scanning in CI/CD
3. Hot spot detection for PRs touching security-sensitive files
4. Mandatory security audit checklist before releases

### 1. Security Tests (259 new tests, all passing ✅)

**Python Tests (195 tests)**
- `tests/unit/test_security_parameters.py` (119 tests)
  - Prototype pollution blocking (`__class__`, `__dict__`, etc.)
  - SQL injection via type coercion (`"999 OR 1=1"` → int)
  - Command injection via strings (`"; rm -rf /"`)
  - Positional argument override attacks
  - Malicious mount parameters
  
- `tests/unit/test_security_uploads.py` (32 tests)
  - Path traversal in filenames (`../../etc/passwd.jpg`)
  - MIME type spoofing (EXE disguised as JPEG)
  - Filename XSS (`<script>alert(1)</script>.jpg`)
  - Upload size enforcement and slot exhaustion
  - Binary frame protocol validation
  
- `tests/unit/test_security_url_injection.py` (44 tests)
  - Live redirect injection (`javascript:`, `data:`, path traversal)
  - CRLF injection in URL params
  - Log injection prevention (ANSI escapes, control chars)
  - Event name injection via URL

**Rust Tests (64 tests)**
- `crates/djust_templates/tests/test_xss_prevention.rs`
  - HTML context escaping (`<script>` → `&lt;script&gt;`)
  - JavaScript context escaping (quotes, backslashes, line separators)
  - URL context escaping (javascript: protocol rejection)
  - JSON/script context edge cases
  - Safe filter bypass documentation
  - Filter chain escaping preservation

### 2. GitHub Workflows (3 new workflows)

[...]

Issue #294, opened 2026-02-13
Debug toolbar not showing: window.DEBUG_MODE not set


johnrtipton:
## Bug

The djust debug toolbar (`debug-panel.js`) checks `window.DEBUG_MODE` to decide whether to initialize, but `_inject_client_script()` in `post_processing.py` only sets `window.djustDebug` — never `window.DEBUG_MODE`.

## Root Cause

`post_processing.py` line ~244 sets `window.djustDebug` but `debug-panel.js` checks `window.DEBUG_MODE`.

## Fix

Add `window.DEBUG_MODE = {str(settings.DEBUG).lower()};` alongside `window.djustDebug` in `_inject_client_script()`.

PR #1816, opened 2026-06-15
fix(websocket): consumer-owned monotonic VDOM send-version (#1788)

### LINKED ISSUE #1788: fix(websocket): consumer-owned monotonic VDOM send-version so html_update fallback is accepted without a recovery round-trip

johnrtipton:
## Summary

Closes #1788. The WebSocket wire `version` stamped on every client-checked frame was the Rust view's internal `self.version`, coupled to the Rust-view object's **baseline lifetime** rather than the connection. On a mid-session VDOM baseline loss (e.g. the patch-compression `_rust_view.reset()` path sets `version=0` + `last_vdom=None`), the next `render_with_diff` returns `(html, patches=None, version=1)` — a **non-sequential** version that fails the client's `clientVdomVersion === data.version - 1` check (`02-response-handler.js:58`), forcing an `html_update` → `request_html` recovery round-trip (a full page reload before #1785).

The **consumer** now owns a monotonic per-connection counter (`_last_sent_version`, exposed via `_next_version()`) — the single source of truth for the wire version on **every** client-checked send path. The wire sequence stays strictly monotonic across a Rust baseline reset, so a post-baseline-loss `html_update` satisfies the client check and is accepted directly with **no recovery round-trip**. Rust `lib.rs` is untouched (its counter still drives diff-baselining).

## Reproduce-first (verified)

`python/djust/tests/test_ws_send_version_1788.py` uses real `WebsocketCommunicator` round-trips. The drift is reproduced via the **true** trigger `view_instance._rust_view.reset()` — NOT `_force_full_html=True`, which keeps `last_vdom` and increments sequentially (it would false-pass). Before the fix the baseline-loss `html_update` carried Rust `version=1` while the client expected `version=3`; after the fix it carries the monotonic consumer version `3`.

**Gate-off (#1468):** reverting the handle_event html-fallback `_send_update` to stamp the Rust `version` makes the strict-sequence guard fail with `[1, 2, 1, 4]` (discontinuity at the baseline-loss boundary) — the recovery storm. Restored; all 7 new tests green.

## Migrated send paths (file:line at merge time)

All client-checked frames that write `clientVdomVersion@77`:

| Path | Site |
|---|---|
| `handle_event` patch + html fallback (**THE #1788 path**) | `websocket.py` ~3791 / ~3828 |
| `_run_async_work` success arms (×2) + error arms (×2) | ~974/995, ~1033/1048 |
| `_dispatch_single_event` (×2) | ~1355/1379 |
| actor event path (ignore `result['version']`) | ~2976 |
| mount frame (covers actor mount; client baseline = 1) | `handle_mount` ~2403 |
| `server_push` | ~5321 |
| `db_notify` | ~5434 |
| `_run_tick` | ~5527 |
| `time_travel_jump` / `_component_jump` / `forward_replay` | ~5020/5093/5185 |
| **HIDDEN #1** — HVR patch frame (`hotreload`) | `websocket.py` ~4396 |
| **HIDDEN #2** — `StreamingMixin.push_state` (bypasses `_send_update`) | `streaming.py` ~287/297 |

`_arm_recovery` now captures `_last_sent_version` (the consumer version of the frame it arms), so the `html_recovery` frame — which the client applies via `clientVdomVersion = data.version` directly (`03-websocket.js:727`) — carries the right version. `handle_request_html`'s sticky fresh-re-render **discards** the fresh Rust version. `_next_version()`/`_arm_recovery` read via `getattr` so partial-construct test fakes still get a valid sequence.

## Out of scope (confirmed untouched)

- `child_update`/`sticky_update` — per-child `clientVdomVersions` Map.
- `mount_batch` per-target `_clientVdomVersions` object. **CONFIRM-DON'T-FIX:** a mount_batch'd view's later events route through the main `handle_event`/`clientVdomVersion` seam, which the client lazily inits from the first event after a batch. This change keeps the batch mount versions monotonic per-connection but the client never feeds them into the main check, so the seam is **neither fixed nor worsened**.
- `noop`/`embedded_update` (no version field), SSE (separate transport), Rust `lib.rs`.

## No client change

The client only does relative `version - 1` checks and direct `clientVdomVersion = data.version` writes — no absolute-version expectations. This is a server-side source change that preserves the same relative sequence (mount=1, event=2, …).

## Tests

- New: `python/djust/tests/test_ws_send_version_1788.py` (7 tests).
- Updated `_recovery_version` assertions (now consumer-owned, decoupled from Rust version) in `test_arm_recovery_helper_1645`, `test_async_recovery_html_1636`, `test_server_push`; HVR version assertions in `test_hotreload`.
- Full local suite shows **zero net regressions** vs the `origin/main` baseline (the repo has a pre-existing local-runner collection artifact identical on both branches; CI is authoritative).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,24 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- **Consumer-owned monotonic VDOM send-version (#1788).** The WebSocket
+  ``version`` stamped on every client-checked frame was the Rust view's
+  internal counter, which resets on a mid-session VDOM baseline loss (e.g. the
+  patch-compression ``_rust_view.reset()`` path). The resulting non-sequential
+  version failed the client's ``clientVdomVersion === data.version - 1`` check,
+  forcing an ``html_update`` → ``request_html`` recovery round-trip (a full page
+  reload before #1785). The consumer now owns a monotonic per-connection
+  counter (``_next_version()``) used as the single source of truth across every
+  client-checked send path (events, async work, mount, server_push, db_notify,
+  ticks, time-travel, hot-reload patches, and ``StreamingMixin.push_state``), so
+  a post-baseline-loss ``html_update`` stays in sequence and the client accepts
+  it directly with no recovery round-trip. Recovery (``html_recovery``) now
+  carries the consumer version of the frame it replaces. New regression tests in
+  ``python/djust/tests/test_ws_send_version_1788.py`` pin the monotonic sequence
+  across the baseline-loss boundary plus the send-path call-site coverage.
+
 ## [1.0.5] - 2026-06-15
 
 ### Added
--- a/python/djust/streaming.py
+++ b/python/djust/streaming.py
@@ -269,8 +269,14 @@ class StreamingMixin:
         from asgiref.sync import sync_to_async
         import json
 
-        # Re-render the full view
-        html, patches, version = await sync_to_async(self.render_with_diff)()
+        # Re-render the full view. The Rust ``version`` is DISCARDED for the
+        # wire — this path bypasses ``_send_update`` and sends frames directly,
+        # but it is a client-CHECKED send path (the client writes
+        # ``clientVdomVersion = data.version`` at 02-response-handler.js:77), so
+        # both frames MUST stamp the consumer-owned monotonic counter
+        # (#1788, HIDDEN #2). Stamping the Rust version here would desync the
+        # client against every other send path.
+        html, patches, _version = await sync_to_async(self.render_with_diff)()
 
         if patches is not None:
             patch_list = json.loads(patches) if patches else []
@@ -278,7 +284,7 @@ class StreamingMixin:
                 {
                     "type": "patch",
                     "patches": patch_list,
-                    "version": version,
+                    "version": self._ws_consumer._next_version(),
                 }
             )
         else:
@@ -288,7 +294,7 @@ class StreamingMixin:
                 {
                     "type": "html_update",
                     "html": html,
-                    "version": version,
+                    "version": self._ws_consumer._next_version(),
                 }
             )
         await self._ws_consumer._flush_push_events()
--- a/python/djust/tests/test_ws_send_version_1788.py
+++ b/python/djust/tests/test_ws_send_version_1788.py
@@ -1,0 +1,452 @@
+"""Regression tests for #1788 — consumer-owned monotonic VDOM wire version.
+
+Root cause (verified by local reproduction of djust.org ``/insights/`` ``set_period``):
+the wire ``version`` stamped on every client-checked frame was the Rust view's
+``self.version`` (``crates/djust_live/src/lib.rs``), which is coupled to the Rust-view
+object's *baseline lifetime*, not the connection. When the Rust view loses its
+``last_vdom`` baseline mid-session (e.g. the patch-compression path calls
+``_rust_view.reset()``, which sets ``version = 0`` and ``last_vdom = None``), the very
+next ``render_with_diff`` returns ``(html, patches=None, version=1)`` — a NON-sequential
+version that does NOT satisfy the client's
+``clientVdomVersion === data.version - 1`` check
+(``python/djust/static/djust/src/02-response-handler.js:58``). The client then treats the
+``html_update`` as a version mismatch and sends ``request_html`` (recovery round-trip;
+pre-#1785 it was a full page reload).
+
+The fix: the *consumer* owns a monotonic per-connection counter (``_last_sent_version``)
+and stamps every client-checked outbound frame via ``_next_version()``. The wire version
+is then decoupled from the Rust view's lifetime, so a baseline loss produces a correctly
+sequenced ``html_update`` (``v_n -> v_{n+1}``) the client accepts directly — no recovery.
+
+CRITICAL FIDELITY NOTE: ``_force_full_html = True`` does NOT lose the Rust baseline (it
+keeps ``last_vdom`` and the version increments sequentially), so it does NOT reproduce the
+drift and would give a FALSE PASS. The TRUE baseline-loss trigger is
+``view_instance._rust_view.reset()`` — that is what these tests use.
+"""
+
+from __future__ import annotations
+
+import inspect
+
+import pytest
+from asgiref.sync import sync_to_async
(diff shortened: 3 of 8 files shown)


johnrtipton:
## Code Review — PR #1816 (#1788, consumer-owned monotonic VDOM wire-version)

**Verdict: APPROVE** (0 🔴, 1 🟡 pre-existing → deferred to #1817).

Worktree-isolated review (`core.bare` untouched, #300).

### Independent regression confirmation (load-bearing)
Same combined invocation (`pytest tests/ python/tests/ python/djust/tests/`) on both branches via the canonical resolver: **PR branch 7741 passed / 0 failed** vs **origin/main 7734 / 0 failed** → delta exactly **+7** (the new `test_ws_send_version_1788.py` cases); failure sets IDENTICAL (both empty). No regression.

### Empirical gate-off (#1468)
Reverted the `handle_event` html-fallback (`websocket.py:3903`) to the Rust `version` → both drift tests FAIL: `test_version_sequence_strictly_monotonic_across_baseline_boundary` gives `Expected [1,2,3,4], got [1,2,1,4]` (the Rust counter resets to 1 after `_rust_view.reset()`). Restored cleanly. Non-tautological.

### Completeness + correctness
- All 16 `_send_update` sites carry `_next_version()` (none use the raw Rust version); `handle_request_html` sends `_recovery_version`; `streaming.py` push_state ×2 via `_ws_consumer._next_version()`. No client-checked site missed.
- Time-travel paths (jump/component_jump/forward_replay) DO route through `handleServerResponse` → migrating them to the counter is correct (the implementer caught these beyond the design list).
- `_arm_recovery` ordering (`v=_next_version()` → arm → send with shared `v`) verified at all 5 arming sites; sticky fresh-re-render discards the fresh Rust version.
- `getattr(self,"_last_sent_version",0)` is test-fakes-only — real `__init__` sets it (`websocket.py:416`); no init-order bug. Reconnect = new consumer → resets to 0 → mount re-seeds. Correct.
- mount_batch confirm-don't-fix verified (batch frames set per-target store; main `clientVdomVersion` lazy-inits from first event). Neither fixed nor worsened.
- No JS change needed; security clean (int counter); two-commit shape correct; existing version suite 63 passed (mount=1, first event=2 preserved).

### 🟡 (pre-existing, deferred to #1817)
`_recovery_version` can go stale vs `_last_sent_version` when a non-arming send (deferred-activity/tick/time-travel) advances the counter between an arming frame and a `request_html`. Structurally identical pre-#1788 (low severity, extra round-trip not reload). Tracked at #1817, not fixed here (#1079).

johnrtipton:
## Retrospective — PR #1816 (#1788)

**Task**: consumer-owned monotonic VDOM wire-version — the deferred (Action #299) optimization removing the recovery round-trip on `html_update` baseline-loss.

### Quality: 5
### What went well
- **Reproduce-first with the TRUE trigger**: `_rust_view.reset()` produces the real `patches=None` baseline-loss; the investigation flagged (and the implementer confirmed) that `_force_full_html=True` false-passes (Rust keeps `last_vdom`). Without that fidelity catch the regression test would have been a no-op.
- **Structural cure (#1646)**: a single `_next_version()` helper as the source of truth across all 16 `_send_update` sites + the 2 hidden participants the issue's list missed (hotreload patch frame, `streaming.py push_state`) + 3 the *design* missed (time-travel jump/component_jump/forward_replay) — the implementer caught those via the drift caveat (worktree-subagent pattern catching brief gaps, #294).
- **Drift-guard as the load-bearing test**: the sequence stays `[1,2,3,4]` across a baseline-loss boundary; gate-off reproduces the exact `[1,2,1,4]` discontinuity. Non-tautological (existing recovery/push tests updated to use non-sequential Rust versions, proving decoupling).
- **Independent regression confirmation**: reviewer ran the combined suite on both branches — +7 (new tests), identical empty failure sets. Worktree-isolated review, `core.bare` untouched (#300 held).
### What didn't
- The 🟡 (`_recovery_version` staleness across non-arming send paths) is pre-existing — deferred to #1817 (#1079), not fixed here.
### Verified
Empirical gate-off (`[1,2,1,4]`); 16/16 send sites on the consumer counter; `_arm_recovery` ordering at all 5 sites; mount_batch confirm-don't-fix; existing version suite 63 passed; full CI green. Rust `lib.rs` untouched (counter stays internal). Patch SemVer (no API/protocol/client change).

RETRO_COMPLETE

Issue #1827, opened 2026-06-17
diff_html: bare-fragment table rows (<tr>/<td> without <table>) foster-parented to #text (#1826 Defect 1 follow-up)


johnrtipton:
Split out from #1826 (Defect 1), per the scope-discipline rule (#1079) — the P0 spurious-`MoveSubtree` fix (Defect 2) shipped separately in the #1826 PR; this issue tracks the *optional* table-fragment context-sensitivity remainder.

## Symptom

When `diff_html` is fed a **bare** `<tbody>` / `<tr>` / `<td>` fragment (no `<table>` ancestor), html5ever foster-parents the table elements out of the tree, so a newly-inserted `<tr>` body node is flattened to a bare `#text` node:

```python
import json
from djust._rust import diff_html

def dom(show):
    out = ["<tbody>"]
    for n in ("a", "b", "c"):
        out.append(f'<tr class="main"><td>{n}</td></tr>')
        inner = f'<tr class="detail"><td>detail {n}</td></tr>' if show else ""
        out.append(f'<!--dj-if id="if-{n}-0"-->{inner}<!--/dj-if-->')
    out.append("</tbody>")
    return "".join(out)

patches = json.loads(diff_html(dom(False), dom(True)))
# InsertChild.node.tag == "#text"  (the <tr> element + attrs are dropped)
```

Wrapping the same rows in `<table><tbody>…</table>` yields the correct `<tr>` elements (0 flattened).

## Why this is NOT the #1826 P0

This does **not** reproduce in production: the real djust DOM always has the `<table>` ancestor, so the `<tr>` is preserved. The #1826 P0 was the spurious `MoveSubtree` ops (Defect 2), which reproduced in *every* container (div/span/table) and is fixed.

## Root cause

`diff_html` (`crates/djust_vdom/src/lib.rs:1152`) parses via `parse_html` / `parse_html_continue`, which use html5ever's default (`body`) insertion context. A context-aware fragment parser already exists — `parse_html_fragment(html, context_tag)` at `crates/djust_vdom/src/parser.rs:278` (with a passing `"table"`-context test at `parser.rs:1302` proving `<tr>` is preserved) — but `diff_html` does not use it.

## Optional fix sketch

Detect the leading element of the fragment (or accept an optional context-tag arg) and route table-shaped fragments through `parse_html_fragment(html, "table")` / `"tbody"`. Low priority — depends on whether any non-production caller feeds `diff_html` bare table fragments.

## Coverage

[...]

Issue #1666, opened 2026-05-29
VDOM: matched dj-if boundary cannot be repositioned (markers have no djust_id, no MoveSubtree)


johnrtipton:
## Summary

When a `{% if %}` conditional subtree (wrapped in `<!--dj-if id=\"X\"-->...<!--/dj-if-->` comment markers) is **matched** across a re-render (same boundary id in old + new) but its **absolute sibling position shifts** — because an *unkeyed* sibling was inserted/removed/reordered before it — the diff cannot move the boundary's comment markers. The id-resolved content relocates correctly, but the markers stay anchored at their old position, so the applied DOM diverges from the intended render.

This is a **pre-existing limitation** (the current/original `diff.rs` fails identically — confirmed by differential testing), not a regression. Filing it as the proper follow-up.

## Root cause

- dj-if markers are `#comment` VNodes with **no `djust_id`**, so no `MoveChild` can target them (it resolves by `child_d`).
- A *matched* boundary correctly emits **no** `RemoveSubtree`/`InsertSubtree` (those are only for id-only-in-one-side boundaries).
- The patch protocol has **no `MoveSubtree`** and no other way to express "move this matched boundary's marker span."

So the markers are effectively pinned to their original index while surrounding id-keyed content moves around them.

## Minimal reproduction

```
old: <div>[ <!--dj-if id="if-a"--><!--/dj-if-->, <p keep> ]
new: <div>[ <section new>, <!--dj-if id="if-a"--><!--/dj-if-->, <p keep> ]
```
A `<section>` is inserted **before** the (matched, empty) boundary. Expected applied result `[section, open, close, p]`; actual `[open, close, section, p]` — the markers stayed at the front. Both the current and a clean-room rebuilt differ produce the wrong result.

## Impact

- Adversarial random-mutation corpus: ~0.65% of inputs (≈39/6000); **every** residual failure is in this class.
- Real-world incidence is low: conditionals normally sit at stable template positions, and surrounding lists are usually keyed or stable. Triggered mainly by unkeyed sibling churn immediately adjacent to a conditional.

## Proposed fix (needs server + client coordination)

The marker span needs a way to be repositioned. Options, roughly in order of preference:

1. **`MoveSubtree { id, path, d, index }` patch variant** — move a matched boundary's marker-pair span by boundary id to a new parent index. Cleanest; symmetric with the existing `RemoveSubtree`/`InsertSubtree`.
[...]

PR #1828, opened 2026-06-17
fix(vdom): dj-if MoveSubtree fires on relative position, not absolute offset (#1826)

### LINKED ISSUE #1826: VDOM: MoveSubtree of dj-if boundary fails on client ('close marker not found') — {% if %}-wrapped <tr> in a loop drops a row per toggle (1.0.6rc1)
Follow-up to #1666 / #559. Those introduced/discussed a `MoveSubtree` patch to reposition matched `dj-if` boundaries. On **1.0.6rc1**, `diff_html` now *emits* `MoveSubtree`, but the client **cannot apply it** and most patches in the batch fail, forcing an `html_recovery` morph that visibly drops a row on each toggle.

## Symptom (browser)

A server-rendered `{% if show %}` that wraps a sibling `<tr>` inside a `{% for %}` loop. Toggling `show` drops one table row each time. Console:

```
[LiveView] MoveSubtree: close marker not found id=if-809ce7b0-15
15/22 patches failed (7 succeeded)
[LiveView] VDOM patches failed. This usually happens when {% if %} blocks
add/remove DOM elements, shifting sibling positions.
Requesting recovery HTML from server.
[LiveView] DOM recovered via morph, version: 5
```

## Server-side origin (`diff_html`)

Feeding `diff_html` HTML shaped like djust's marker output — N main rows, each followed by its own `<!--dj-if id="if-<n>-0"-->…<!--/dj-if-->` block that goes empty → non-empty — produces a broken stream:

1. The new `<tr class="detail">` is inserted as a **bare `#text` node** (tag + attributes dropped — the element is flattened).
2. `MoveSubtree` ops are emitted against the `dj-if` markers of **later** iterations (`id="if-b-0"`, `id="if-c-0"`), which the client then can't pair with their close markers.

## Minimal reproduction

Dependency-light (stdlib + `djust._rust`), drops in as `tests/test_diff_html_if_marker_rows.py`. **1 failing test (the bug) + 1 passing control (the class-toggle workaround).**

```python
import json
from djust._rust import diff_html

ITEMS = ["a", "b", "c"]

def _marker_dom(show):
    out = ["<tbody>"]
    for n in ITEMS:
        out.append(f'<tr class="main"><td>{n}</td></tr>')
        inner = f'<tr class="detail"><td>detail {n}</td></tr>' if show else ""
        out.append(f'<!--dj-if id="if-{n}-0"-->{inner}<!--/dj-if-->')
    out.append("</tbody>")
    return "".join(out)

def test_structural_if_marker_diff_is_broken():
    patches = json.loads(diff_html(_marker_dom(False), _marker_dom(True)))
    inserts = [p for p in patches if p["type"] == "InsertChild"]
    moves   = [p for p in patches if p["type"] == "MoveSubtree"]
    # (a) inserted node should be the <tr> element, never a flattened #text node
    flattened = [p for p in inserts if (p.get("node") or {}).get("tag") == "#text"]
    assert not flattened, json.dumps(flattened, indent=2)
    # (b) filling empty {% if %} blocks shouldn't MoveSubtree later markers
    assert not moves, [m.get("id") for m in moves]
```

Observed on 1.0.6rc1: 3 `#text`-flattened `InsertChild` + 2 `MoveSubtree` (`if-b-0`, `if-c-0`).

## Workaround (confirmed)

Keep the row always present and flip a CSS class instead of structurally adding/removing it — pure `SetAttr` diff, no marker moves:

```django
{# instead of {% if show %}<tr>…</tr>{% endif %} #}
<tr class="detail{% if not show %} hidden{% endif %}">…</tr>
```

with `.detail.hidden { display:none }`. This matches djust's own console hint.

## Version

djust `1.0.6rc1`.

johnrtipton:
Fixes #1826 (P0 VDOM dj-if `MoveSubtree` regression, 1.0.6rc1). Filed follow-up #1827 (Defect 1, out of scope here).

## Root cause (Defect 2 — the P0)

`diff_html`'s matched-boundary move-decision in `crates/djust_vdom/src/diff.rs` compared **absolute** child offsets:

```rust
let old_open_abs = old_off + ob.open;
let new_open_abs = new_off + nb.open;
if old_open_abs != new_open_abs { /* MoveSubtree */ }
```

A `{% if %}` wrapping a `<tr>` inside a `{% for %}` loop renders N rows, each followed by its own `<!--dj-if id="if-<n>-0"-->…<!--/dj-if-->` boundary. When an **earlier** empty boundary body fills (`show` toggles false→true), the inserted node shifts the **absolute** index of every **later** boundary even though those boundaries did **not** move relative to their siblings. The differ emitted spurious `MoveSubtree { id: "if-b-0" }` / `{ id: "if-c-0" }` ops the client cannot pair (`close marker not found`) → ~15/22 patches fail → `html_recovery` morph drops a row per toggle.

Reproduced bit-exactly against the 1.0.6rc1 build (the issue's reproducer): `MoveSubtree ids == ['if-b-0', 'if-c-0']`. The spurious move fires in **every** container (div/span/table), confirming it is independent of the table-fragment artifact.

## Fix

Key the move-decision on the boundary's position **relative** to its non-boundary siblings, plus its ordinal among same-level boundaries:

- New private helper `non_boundary_count_before(excluded, open)` over the existing `excluded` mask (collapses boundary spans so an earlier boundary's span-length change is invisible).
- `old_key = (non_boundary_count_before(&old_excluded, ob.open), old_ordinal)`; `new_key = (non_boundary_count_before(&new_excluded, nb.open), new_ordinals[id])`. Emit `MoveSubtree` iff `old_key != new_key`.
- Move **target** index unchanged (`new_off + nb.open`) — only the *decision* was wrong.

A **genuine** reposition (a real element inserted before the boundary, `nb_before` 0→1) still flips the key and emits the move.

## Defect 1 (#text flattening) — scoped out (#1079)

The `#text`-flattened insert is an html5ever foster-parenting artifact of a **bare** `<tbody>` fragment (no `<table>` ancestor). Verified: `<table><tbody>…` preserves the `<tr>` (0 flattened); div/span yields `<span>`. Does not reproduce in production (real DOM has the `<table>`). Tracked in follow-up **#1827** (`diff_html` could route table fragments through the existing `parse_html_fragment(html, "table")`, parser.rs:278). Not fixed here.

## Tests

| File | What it pins |
|------|--------------|
| `crates/djust_vdom/tests/test_dj_if_loop_spurious_move_1826.rs` | Loop-fill: **no** `MoveSubtree` for later boundaries (load-bearing, fails on rc1); element-not-`#text` insert; client-faithful `apply_all` round-trip keeping all rows; `assert_handles_resolve`. Plus a genuine-reposition guard. |
| `python/djust/tests/test_diff_html_if_marker_rows_1826.py` | `diff_html`-level div/span (no move, no `#text`) + table-wrapped (`<tr>` preserved) + bare-`<tbody>` documents Defect 1 as environment artifact. 4 cases. |

**Gate-off (#1468):** reverting the move-decision to the absolute-offset form makes the load-bearing assertion fail with `left: 2  right: 0` (the 2 spurious moves); the genuine-reposition guard still passes under both — proving the assertion is non-tautological. Restored; both pass.

## Verification

- `cargo test -p djust_vdom` — all binaries green (the two #1666 guards `matched_djif_boundary_repositioned_via_move_subtree` / `matched_djif_boundary_no_move_when_position_unchanged`, the `proptest_random_toggles_handles_always_resolve` net, and the `torture_*dj_if*` / `torture_*round_trip*` nets all pass).
- Scoped `pytest` over VDOM/diff/render Python tests: 223 passed, 1 skipped (free-threaded-only).
- `cargo fmt --check` + `cargo clippy -p djust_vdom --tests` clean.

CI is authoritative for the full suite (pushed `--no-verify` from a git worktree per #1796).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,34 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- **dj-if ``MoveSubtree`` keyed on RELATIVE position, not absolute offset
+  (#1826).** A ``{% if %}``-wrapped element inside a ``{% for %}`` loop dropped
+  a table row per toggle (P0 regression in 1.0.6rc1). ``diff_html``'s
+  matched-boundary move-decision in ``crates/djust_vdom/src/diff.rs`` compared
+  ABSOLUTE child offsets (``old_off + ob.open`` vs ``new_off + nb.open``).
+  Filling an EARLIER empty dj-if body inserts nodes that shift the absolute
+  index of every LATER boundary even though those boundaries did NOT move
+  relative to their siblings — so the differ emitted spurious
+  ``MoveSubtree { id: "if-b-0" / "if-c-0" }`` ops the client could not pair
+  (``close marker not found``), ~15/22 patches failed, and an ``html_recovery``
+  morph dropped a row. The decision now keys on the boundary's position
+  RELATIVE to its non-boundary siblings (a new ``non_boundary_count_before``
+  helper over the existing ``excluded`` mask) plus its ordinal among same-level
+  boundaries; both are invariant to a sibling boundary's span-length change, so
+  only a GENUINE reposition (a real element inserted/removed before the
+  boundary, or a boundary reorder) emits a move. The move TARGET index is
+  unchanged. The ``#text``-flattening also seen in #1826 (Defect 1) is an
+  html5ever foster-parenting artifact of a bare ``<tbody>`` fragment (no
+  ``<table>`` ancestor); it does not reproduce in production and is scoped out
+  to follow-up #1827. Covered by the Rust reproducer
+  ``crates/djust_vdom/tests/test_dj_if_loop_spurious_move_1826.rs``
+  (loop-fill no-spurious-move + client-faithful round-trip + genuine-reposition
+  guard; gate-off verified per #1468) and 4 regression cases in
+  ``python/djust/tests/test_diff_html_if_marker_rows_1826.py``. The two #1666
+  guards and the proptest / torture dj-if round-trip nets stay green.
+
 ## [1.0.6rc1] - 2026-06-17
 
 ### Added
--- a/crates/djust_vdom/src/diff.rs
+++ b/crates/djust_vdom/src/diff.rs
@@ -106,6 +106,20 @@ fn find_top_level_boundaries(children: &[VNode]) -> (Vec<Boundary>, Vec<bool>) {
     (boundaries, excluded)
 }
 
+/// Count how many NON-boundary siblings precede the local index `open`.
+///
+/// `excluded` is the mask from [`find_top_level_boundaries`] (index `k` is
+/// `true` when it falls inside ANY boundary marker span). This collapses the
+/// boundary spans so that a boundary's "significant position" is measured
+/// against its real (non-marker) siblings ONLY — not the absolute child
+/// index, which shifts whenever an EARLIER boundary body fills/empties
+/// (#1826). Two renders that keep the same non-boundary neighborhood around a
+/// boundary yield the same count even if an earlier sibling boundary changed
+/// span length.
+fn non_boundary_count_before(excluded: &[bool], open: usize) -> usize {
+    (0..open).filter(|&k| !excluded[k]).count()
+}
+
 /// Serialize a boundary's full marker-pair span (open..=close, inclusive) to
 /// HTML for `InsertSubtree.html`.
 fn serialize_boundary_html(children: &[VNode], open: usize, close: usize) -> String {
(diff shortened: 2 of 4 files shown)


johnrtipton:
## Code Review — PR #1828 (#1826, P0 VDOM dj-if MoveSubtree) — APPROVE

Worktree-isolated adversarial review (0 🔴, 1 🟡 non-blocking; `core.bare` untouched).

### The ordinal false-positive concern: empirically DISPROVEN
Built + ran 11 adversarial cases the PR tests don't cover. The decisive one — does an ordinal shift (boundary added/removed before a matched one) emit a spurious move? — **no**: ordinal-shift always co-occurs with a genuine DOM reposition (the marker span really shifts), and the client-faithful `apply_all` round-trip reproduces NEW correctly in every case. No new false-positive; no false-negative (genuine reorders `[if-a,if-b]→[if-b,if-a]` still move — the ordinal term is load-bearing there).

### Gate-off (#1468)
Reverted to the absolute test → `djif_loop_fill_emits_no_spurious_move_for_later_boundaries` fails `left: 2 right: 0` (Rust) AND all 4 PyO3 tests fail with the literal `MoveSubtree{id:"if-b-0",index:5}`/`{id:"if-c-0",index:9}` on both div + table shapes. The #1666 genuine-move guard passes under both decisions. Non-tautological.

### Safety net + correctness
- #1666 tripwires (`matched_djif_boundary_repositioned_via_move_subtree` / `..._no_move_when_position_unchanged`) both pass — no over-correction.
- Full `cargo test -p djust_vdom` green across all binaries (proptest round-trip, `torture_*dj_if*`, #1358/#1408/#1420, wire-protocol snapshot, the new test).
- Client-faithful table-scenario test passes; Defect-1 (#text) correctly scoped as a bare-`<tbody>` html5ever artifact → follow-up #1827.
- `new_ordinals[ob.id]` panic-safe (inside the matched-in-both `if let Some`); no `unwrap()` added; clippy clean. Two-commit shape + CHANGELOG (#1106) correct.

### 🟡 (non-blocking, not fixed here)
When a boundary is removed/added *before* a matched empty boundary, a `MoveSubtree` is emitted that's redundant with the Remove/Insert (the sibling shift already relocates it). Harmless — `apply_patches` drains+reinserts at the same index, round-trip green. One extra patch in a rare shape; noted, not tracked (no correctness/pairing impact).

johnrtipton:
## Retrospective — PR #1828 (#1826, P0 VDOM regression)

**Task**: `diff_html` dropped a table row per `{% if %}<tr>` toggle (regression in 1.0.6rc1).

### Quality: 5
### What went well
- **Reproduction fidelity caught the bigger story**: the reporter's bare-`<tbody>` repro showed TWO defects, but the investigation proved Defect 1 (#text flattening) was an html5ever foster-parenting artifact (gone with a `<table>` wrapper) — NOT a framework bug. Scoped to fixing Defect 2 (the real, container-independent P0) + the reproducer, and filed #1827 for the optional `diff_html` fragment-context hardening (#1079).
- **Root cause was a wrong statistic**: the move-decision keyed on ABSOLUTE offset (`old_off+ob.open != new_off+nb.open`), which shifts whenever an earlier dj-if fills — so later boundaries got spurious moves the client couldn't pair. Fixed to a RELATIVE key (`non_boundary_count_before(excluded, open)` + ordinal), keeping the absolute `index` as the move target.
- **Adversarial review empirically disproved the one real risk** (ordinal-shift false-positive) across 11 edge cases + a client-faithful `apply_all` round-trip; gate-off reproduced `left 2 right 0` on both Rust + PyO3.
- #1666 guards + full `cargo test -p djust_vdom` (all binaries) green — the proptest round-trip is the blast-radius net for a core-engine change.
### What didn't
- Shipped in 1.0.6rc1 (the MoveSubtree emission from #1666/#559 mis-fired on pure insertion); caught only post-release by a user. The #1666 work lacked a "fill an earlier sibling boundary" test case — the new `test_dj_if_loop_spurious_move_1826` closes that gap.
### Verified
Reproduce-first (bit-exact); relative-key fix `diff.rs:343-360` + helper `:119`; gate-off non-tautological (Rust + PyO3); #1666 guards green; full vdom suite green; client-faithful round-trip keeps all rows; 1 🟡 harmless-redundant-move noted. CHANGELOG `### Fixed` (#1826). Ships in 1.0.6rc2.

RETRO_COMPLETE

johnrtipton:
**Correction (framing):** #1826 is NOT a 1.0.6rc1 regression — the buggy absolute-offset `MoveSubtree` decision was introduced by the MoveSubtree feature in #1666 (commit `912e630c`, 2026-05-29) and shipped in **v1.0.0rc16 → v1.0.0 → every release since** (1.0.1–1.0.5, 1.0.6rc1). It's a latent bug present since v1.0.0; the reporter just happened to hit it on 1.0.6rc1. CHANGELOG + ROADMAP corrected on main. The fix and its urgency are unchanged; only the "introduced in" attribution was wrong.

PR #1829, opened 2026-06-18
test(vdom): JS-client apply coverage for dj-if MoveSubtree (#1826 follow-up)

johnrtipton:
Follow-up to merged PR #1828 (#1826 P0 VDOM regression). #1828 fixed the Rust `diff_html` spurious-`MoveSubtree` but added only **server-side** tests (Rust + Python `diff_html`). The **original #1826 symptom was client-side** — `applyMoveSubtree` → `_findDjIfCloseMarker` returning null → `console.warn('MoveSubtree: close marker not found id=%s')` → 15/22 patches fail → `html_recovery` morph drops a row. The Rust `apply_all` harness is **not faithful** to the JS client: Rust `match_close_idx` pairs within the parent's child slice; the JS client walks the **whole document** with a depth-counting `TreeWalker` (`_findDjIfCloseMarker`, `12-vdom-patch.js:1290`). So genuine `MoveSubtree` emissions were unverified against the real client. This PR closes that gap.

## What's added

`tests/js/dj_if_movesubtree_client_apply_1826.test.js` (vitest/JSDOM), importing the **REAL** `src/12-vdom-patch.js` (`applyMoveSubtree` / `_findDjIfCloseMarker` / `_applySinglePatch` / `_sortPatches` — not a reimplementation), building OLD DOM via `innerHTML` **with inter-element SSR whitespace** (CLAUDE.md #1724 reproduction fidelity), modeling dj-if markers as real `<!--dj-if id="...">` / `<!--/dj-if-->` comment nodes, and applying the patch streams the **corrected** `diff_html` actually emits (captured from `djust._rust.diff_html` against the exact strings).

| # | Shape | Corrected `diff_html` stream | Client result |
|---|-------|------------------------------|---------------|
| 1 | reporter all-fill (`{% if %}<tr>` in 3-iter loop, empty→filled) | 0 MoveSubtree, 3 InsertChild | clean apply, all 3 detail rows present |
| 2 | genuine reposition (element before boundary) | `MoveSubtree if-a-0` | pairs close marker + repositions, **no warning** |
| 3 | redundant-move (boundary removed before matched empty boundary — the #1828 review's harmless 🟡) | `RemoveSubtree if-x-0` + redundant `MoveSubtree if-a-0` | both pair cleanly, no warning |
| 4 | nested dj-if (inner inside outer's body) with outer move | `MoveSubtree if-outer-0` | document-wide depth-walk pairs the **outer** close marker, not the inner |

## Result: the real client PASSES all four shapes

0 "close marker not found" warnings, DOM correct (all rows present, boundary positioned right). **No client change was needed** — the existing `_findDjIfCloseMarker` document-wide depth-walk already handles every shape including nesting. These tests ARE the coverage that closes the JS-apply gap the server-only #1828 tests left open, and they fail-fast against any future client regression.

## Gate-off (#1468 / #1200) — non-tautological

- Forcing `_findDjIfCloseMarker` → `null` (the literal #1826 symptom) fails shapes **2/3/4** with `MoveSubtree: close marker not found id=if-a-0`.
- Breaking the depth counter (return on first `/dj-if`) fails shape **4** (mis-pairs nested boundaries).
- Gating off InsertChild dispatch fails shape **1** (detail rows missing).

## #1827 (Defect 1) liveness — determination

Traced every real path that feeds the Rust differ: `render_with_diff` (full template render, `crates/djust_live/src/lib.rs:315`), streaming `push_state` (full re-render via `render_with_diff`, `streaming.py:279`), `_render_embedded_child` (full child template, **bypasses VDOM**, `websocket.py:4020`), LiveComponent `render` (uses `render_template` one-way + `<div>` wrap, not the differ), stream insert/append ops (raw `html` ops to client, **bypass the differ**), SSE (full-page only). **Every differ-fed render starts from the view's full template root, which contains the `<table>`/`<ul>`/`<select>` ancestor.** No real path renders a bare `<tr>`/`<li>`/`<option>`/`<td>` fragment standalone through `diff_html`. **#1827 is theoretical/low — confirmed.** Determination posted to #1827.

## Verification

- New test + `tests/js/move_subtree_patch.test.js` + `tests/js/dj_if_subtree_patches.test.js`: **34 passed**.
- `eslint` clean; changelog-test-count hook green.
- Pushed `--no-verify` from a git worktree (#1796 — native pre-push hardcodes `.venv/bin/python`); CI is authoritative.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,7 +35,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   (loop-fill no-spurious-move + client-faithful round-trip + genuine-reposition
   guard; gate-off verified per #1468) and 4 regression cases in
   ``python/djust/tests/test_diff_html_if_marker_rows_1826.py``. The two #1666
-  guards and the proptest / torture dj-if round-trip nets stay green.
+  guards and the proptest / torture dj-if round-trip nets stay green. The
+  ORIGINAL #1826 symptom was CLIENT-side (``applyMoveSubtree`` →
+  ``_findDjIfCloseMarker`` → ``close marker not found``), and the Rust
+  ``apply_all`` harness is not faithful to the JS client's document-wide
+  depth-counting ``TreeWalker``; new cases in the ``dj-if MoveSubtree —
+  JS-client apply (#1826 follow-up)`` describe block
+  (``tests/js/dj_if_movesubtree_client_apply_1826.test.js``) exercise the REAL
+  ``src/12-vdom-patch.js`` against the corrected ``diff_html`` patch streams for
+  the all-fill, genuine-reposition, redundant-move, and nested-boundary shapes —
+  closing the JS-apply coverage gap the server-only tests left open (the real
+  client passes all four; no client change needed).
 
 ## [1.0.6rc1] - 2026-06-17
 
--- a/tests/js/dj_if_movesubtree_client_apply_1826.test.js
+++ b/tests/js/dj_if_movesubtree_client_apply_1826.test.js
@@ -1,0 +1,369 @@
+/**
+ * JS-CLIENT apply coverage for dj-if `MoveSubtree` — #1826 follow-up.
+ *
+ * PR #1828 fixed the Rust `diff_html` so a `{% if %}`-wrapped `<tr>` in a
+ * `{% for %}` loop emits 0 spurious `MoveSubtree` for later boundaries. But
+ * #1828 added only SERVER-side tests (Rust + a Python `diff_html` pack). The
+ * ORIGINAL symptom was CLIENT-side: `12-vdom-patch.js`'s `applyMoveSubtree` →
+ * `_findDjIfCloseMarker` returning null → `console.warn('MoveSubtree: close
+ * marker not found id=%s')` → 15/22 patches fail → `html_recovery` morph drops
+ * a row per toggle.
+ *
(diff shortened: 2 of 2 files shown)


johnrtipton:
## Retrospective — PR #1829 (#1826 follow-up: JS-client apply coverage)

**Task**: close the JS-client-apply gap #1828 left (server-only tests; the original symptom was client-side).

### Quality: 5
### What went well
- Reproduction-faithful (#1724): tests import the REAL `12-vdom-patch.js` (`applyMoveSubtree`/`_findDjIfCloseMarker`), build the DOM via `innerHTML` WITH SSR whitespace + real `<!--dj-if-->` comment markers, apply the actual `diff_html` patch streams.
- All 4 genuine-move shapes (reposition, the #1828 🟡 redundant move, nested boundaries, the reporter's all-fill) PASS on the real client — **confirming #1826 was purely a server-emission bug; the client always paired correctly**. The document-wide depth-walk handles nesting.
- Gate-off (#1468) non-tautological: forcing `_findDjIfCloseMarker→null` reproduces the literal `close marker not found`; breaking the depth counter fails the nested case.
- #1827 (Defect 1) confirmed THEORETICAL: every differ render starts from the full template root (with the `<table>`/`<ul>` ancestor); no real path feeds a bare fragment to `diff_html` — determination posted to #1827.
### Verified
34 JS tests pass (new + existing MoveSubtree suites); test-only (12-vdom-patch.js unchanged vs main); two-commit shape; ESLint clean.

PR #1834, opened 2026-06-18
fix(vdom): unique dj-if marker id per loop iteration (#1832)

### LINKED ISSUE #1832: VDOM: {% if %} inside {% for %} reuses ONE dj-if marker id across all iterations → 15× MoveSubtree on a duplicate id → 'close marker not found' (1.0.6rc2)
On **1.0.6rc2**, a `{% if %}` block inside a `{% for %}` loop is wrapped in `<!--dj-if id="if-<hash>-<N>"-->` markers where `<N>` is the block's **compile-time ordinal**. The loop emits that block once per iteration but keeps the **same ordinal**, so the marker id is **duplicated across every iteration**. When a later re-render repositions one boundary, `diff_html` emits a `MoveSubtree` for that id that now matches N identical markers — the client gets N unpairable moves (`close marker not found`), most patches fail, and it falls back to a recovery morph that visibly drops a row on each toggle.

This is adjacent to #1826/#1666 but distinct: the defect is **duplicate marker ids within a loop**, not the relative-vs-absolute move decision (which #1828 fixed).

## Browser symptom

```
[LiveView] MoveSubtree: close marker not found id=if-809ce7b0-15
15/22 patches failed (7 succeeded)
VDOM patches failed. This usually happens when {% if %} blocks add/remove DOM
elements, shifting sibling positions.
Requesting recovery HTML from server.
[LiveView] DOM recovered via morph, version: 5
```

Note all 15 failed moves target the **same** id (`if-809ce7b0-15`).

## Minimal reproduction (server-side, no browser)

Instantiates a real `LiveView` and calls `render_with_diff()` — the actual WebSocket render path. (A bare `render_template` does **not** emit dj-if markers, so it cannot reproduce this.)

```python
import json, re
from collections import Counter
from djust import LiveView

class V(LiveView):
    template = (
        '<div dj-root dj-view="v.V" dj-id="0"><table class="{{ tcls }}"><tbody>'
        "{% for r in rows %}"
        "<tr><td>{% if r.hi %}<span>{{ r.hi }}</span>{% else %}<span>-</span>{% endif %}</td></tr>"
        '{% if r.alerts %}<tr class="detail"><td>'
        "{% for a in r.alerts %}<span>{{ a }}</span>{% endfor %}</td></tr>{% endif %}"
        "{% endfor %}</tbody></table></div>"
    )
    def mount(self, request, **k):
        self.rows = [{"hi": i, "alerts": ["x", "y"]} for i in range(16)]
        self.tcls = "collapsed"
    def get_context_data(self, **k):
        return {"rows": self.rows, "tcls": self.tcls, "view": self}

class Req: method = "GET"; GET = {}
v = V(); v.mount(Req())
html, _, _ = v.render_with_diff()

ids = Counter(re.findall(r'dj-if id="(if-[0-9a-f]+-\d+)"', html))
print("duplicate marker ids:", {k: c for k, c in ids.items() if c > 1})
# -> {'if-<hash>-0': 16, 'if-<hash>-1': 16}   <-- DEFECT: ids duplicated per iteration

v.tcls = ""                      # toggle ONLY the <table> class
_, patches, _ = v.render_with_diff()
p = json.loads(patches)
moves = [x for x in p if x["type"] == "MoveSubtree"]
print(len(p), "patches,", len(moves), "MoveSubtree,",
      dict(Counter(m["id"] for m in moves)))
# -> 16 patches, 15 MoveSubtree, {'if-<hash>-1': 15}   <-- client can't pair these
```

Observed on 1.0.6rc2: `if-<hash>-0` and `if-<hash>-1` each appear **16×** at mount; toggling the table class emits **15 `MoveSubtree`** all for `if-<hash>-1`.

## Expected

Each rendered `{% if %}` boundary should have a **unique** marker id (e.g. ordinal combined with the loop index / a per-instance counter), so a `MoveSubtree` resolves to exactly one boundary.

## Workaround

Avoid `{% if %}` inside `{% for %}` for anything that can be re-rendered: compute the conditional server-side and emit a plain class/attribute, or hide via CSS using a single conditional **outside** the loop. (In our case we moved a per-row `sec-hidden` conditional to a single `details-collapsed` class on the `<table>`.)

## Version

djust `1.0.6rc2`.

johnrtipton:
Closes #1832.

## The bug (P0, data-loss)

A `{% if %}` inside a `{% for %}` was wrapped in `<!--dj-if id="if-<hash>-N"-->` markers where `N` is the parser's **compile-time ordinal**. The loop renders that one `If` node once per iteration, so the marker id is **duplicated across every iteration**. On a later re-render that repositions one boundary, `diff_html` emits a `MoveSubtree` whose id now matches N identical markers → the client gets N unpairable moves (`close marker not found`), most patches fail, and the recovery morph visibly **drops a row on each toggle**.

Distinct from #1826/#1828 (relative-vs-absolute move decision, #1828); this is **duplicate marker ids within a loop**.

## Reproduction (server-side, no browser)

The issue's `render_with_diff()` repro on `1.0.6rc2`/main:
```
duplicate marker ids at mount: {'if-<hash>-0': 16, 'if-<hash>-1': 16}
16 patches, 15 MoveSubtree, {'if-<hash>-1': 15}   <- all 15 target the same id
```
After this fix:
```
duplicate marker ids at mount: {}
1 patches, 0 MoveSubtree, {}
```

## Fix

`crates/djust_templates/src/renderer.rs`:
- **`Node::For`** threads a per-iteration loop-index path through the render `Context` (mirroring the existing `__djust_cycle_counter` save/restore for nested-loop support), using the **original** item index so ids stay stable under `{% for ... reversed %}`.
- **`Node::If`** appends that path to the marker id at emission: `if-<hash>-N` → `if-<hash>-N-<index>` (composing for nested loops, e.g. `-<outer>-<inner>`).

The id is now **unique per rendered iteration** AND **stable across re-renders that don't change loop structure** — so the differ still pairs markers when only non-loop sibling state toggles. Outside any loop the id is unchanged (`if-<hash>-N`, backward compatible). The close marker `<!--/dj-if-->` carries no id and is unchanged.

The id is treated as an **opaque string** by every consumer (grep-verified): `strip_dj_if_markers` (regex matches any attributes), the Rust differ `dj_if_open_id` (full-string equality), and the JS client `_extractDjIfMarkerId` / `_findDjIfOpenMarker` (full-id match) — none parse the `-N` structure.

## Tests

- **New** `crates/djust_templates/tests/test_dj_if_loop_unique_marker_id_1832.rs` + `python/tests/test_dj_if_loop_marker_id_1832.py`: assert unique-per-iteration ids, re-render stability, nested-loop composition, and an If outside a loop keeps the bare form.
- **Corrected** two pre-existing tests (`test_if_markers.rs::for_if_iteration_stable_id`, `test_template_if_markers.py::test_for_if_iteration_uses_same_id`) that explicitly **asserted the duplicate-id behavior** (they pinned the bug).
- **Gate-off self-test (#1468):** reverting only the If-emission append makes the new tests fail with the exact bug signature (`{'if-<hash>-0': 16, 'if-<hash>-1': 16}` / 15 unpairable moves); restored → green.

## Verification

`make dev-build`; `cargo test -p djust_templates` + `-p djust_vdom` green (incl. `wire_protocol_snapshot`, `test_dj_if_keyed_diff_1358`, `test_dj_if_loop_spurious_move_1826` — id opacity preserved); Python sweep `-k "vdom or dj_if or render or diff or template"` 477 passed.


johnrtipton:
## Stage 11 — Adversarial Code Review: ✅ PASS (0 🔴 / 0 🟡)

Independent read-only review; every vector probed with the reviewer's own reproductions.

- **Stale base (#1450):** `HEAD..origin/main` = 0.
- **Headline fix:** 16-row repro — 16 distinct ids, 0 dupes; toggling table class → `{SetAttr:1}`, **0 MoveSubtree**. P0 symptom gone.
- **Row add/remove degrades gracefully (main probe):** insert (16→17) → `{SetText:8, InsertChild:1}`; remove (16→15) → `{SetText:7, RemoveChild:1}`; 0 unpairable MoveSubtree either way. Positional `-<index>` shift causes no data loss — the differ reconciles via InsertChild/RemoveChild+SetText, not unpairable moves.
- **No collision:** dash-joined decimals are an injective map; `N` unique per `Node::If` (`parser.rs` monotonic counter), path = unique loop-nesting chain. Empirically: nested-loop + multi-if template → 22 markers, 22 distinct.
- **Stability:** identical data rendered twice → byte-identical ids, second render 0 patches.
- **For restore:** no stale-path leak — sibling loops + a mid-`{% if %}` get the bare id; nested-then-tail-if carries the correct single-level outer path; empty-string-reset ≡ key-removal because `Node::If` appends only non-empty paths.
- **Id opacity:** Rust differ `dj_if_open_id` + JS `_extractDjIfMarkerId`/`_findDjIfOpenMarker` match by full-string equality; `strip_dj_if_markers` is structure-agnostic.
- **Stale-test rewrites legitimate:** both old tests asserted the duplicate-id (buggy) behavior; corrected, not weakened.
- **Suites:** `cargo test -p djust_templates` + `-p djust_vdom` green; pytest `-k "vdom or dj_if or render or diff or template"` 477 passed.

Working tree unmutated; `core.bare=false`. **REVIEW VERDICT: PASS.** Re-Review skipped (0 findings). All 14 CI checks green.

johnrtipton:
## Retrospective — PR #1834 (#1832, pipeline-run)

**Task**: v1.0.6-5 — P0 VDOM data-loss: `{% if %}` inside `{% for %}` reused one compile-time-ordinal dj-if marker id across all iterations → unpairable MoveSubtree → dropped row per toggle.

### Quality: 5/5
Root cause traced symptom-up, reproduced server-side first, fixed with a minimal Rust change that mirrors an existing pattern, verified zero behavior regression + graceful row-mutation degradation, merged green on all 14 CI checks.

### What went well
- **Reproduce-first** (the issue shipped a `render_with_diff()` repro) pinned the defect on main before any change — and the same repro became the after-proof (0 dupes / 0 moves).
- **Mirrored an existing pattern.** The fix copies the `__djust_cycle_counter` save/restore shape already in `Node::For`, so the loop-path threading is idiomatic and low-risk (Stage-4 grep-for-existing-pattern, #168).
- **Caught two bug-pinning tests.** `for_if_iteration_stable_id` / `test_for_if_iteration_uses_same_id` literally asserted the duplicate-id behavior (docstrings: "every loop iteration emits the same id"). They were enshrining the defect; corrected to assert uniqueness + re-render stability.
- **Gate-off self-test (#1468)** proved the new tests fail on the old behavior with the exact bug signature before review.
- **Adversarial review probed the new risk** (positional-index keying under row add/remove) empirically and confirmed graceful degradation — not just inspection.

### What to watch
- The fix keys boundaries positionally (`-<index>`); row insert/remove shifts later suffixes, handled correctly as InsertChild/RemoveChild+SetText (no data loss). For keyed-identity stability under reordering, `dj-key` remains the mechanism — this fix is orthogonal.
- This is the **third** dj-if/MoveSubtree bug in the v1.0.6 line (#1666/#1826/#1828 relative-vs-absolute, now #1832 duplicate ids). The dj-if keyed-diff surface is correctness-dense; a dedicated client-faithful differential harness over `{% if %}`×`{% for %}` permutations would likely flush remaining edge cases. (Candidate follow-up.)

### Verified
Repro fixed (0 dupes/0 moves); `cargo test -p djust_templates` + `-p djust_vdom` green (incl. wire-protocol-snapshot, keyed-diff-1358, spurious-move-1826 — id opacity preserved); pytest sweep 477 passed; new Rust+Python regression tests + gate-off; adversarial review PASS; 14/14 CI green.

`RETRO_COMPLETE`

PR #1839, opened 2026-06-18
test(js): make dj-transition active/end-on-next-frame test deterministic (#1830)

### LINKED ISSUE #1830: test: flaky dj_transition 'active/end on next frame' under full make test (rAF timing, #1795 family)
`tests/js/dj_transition.test.js > dj-transition > applies start class synchronously, active/end on next frame` intermittently fails under the full `make test` parallel JS run with `AssertionError: expected false to be true`, but passes **3/3 in isolation** (`npx vitest run tests/js/dj_transition.test.js`).

It's a `requestAnimationFrame`-timing assertion that's sensitive to event-loop scheduling under load — the same outlier-sensitivity class as #1795 (concurrency/timing tests must assert a robust signal) and the v1.0.6rc1 perf-benchmark mean→median fix.

**Fix direction:** make the rAF-callback test deterministic — drive the rAF stub explicitly (resolve a microtask / fake-timer the frame) and assert on the class transition by event ordering, not by "happens on the next real frame." Don't assert on real rAF timing.

Surfaced during the Dependabot security bump (unrelated — Python-only lockfile change).

johnrtipton:
Closes #1830.

## The flake
`tests/js/dj_transition.test.js > applies start class synchronously, active/end on next frame` intermittently failed under the full parallel `make test` JS run (`AssertionError: expected false to be true`) while passing 3/3 in isolation. It relied on a `setTimeout(0)` microtask flush winning a race against dj-transition's 16 ms rAF fallback — under load the real rAF fired first, advancing past phase 1 before the `start-cls` assertion. Same outlier-sensitivity class as #1795 (never assert on real-frame timing) and the rc1 perf-benchmark mean→median fix.

## Fix
Add an opt-in `controlledRaf` mode to the test's `createDom` that replaces jsdom's timer-backed `requestAnimationFrame` with a queue driven explicitly via `dom.flushFrame()`. dj-transition captures `globalThis.requestAnimationFrame` at `_runTransition` time, so with the stub a frame fires only when the test drives it.

The rewritten test is **fully synchronous** and asserts the **ordering invariant**:
- phase 1 (`start-cls`) is applied synchronously by the DOMContentLoaded init sweep (`querySelectorAll('[dj-transition]').forEach(_runTransition)`);
- `active-cls`/`end-cls` are NOT present until exactly one driven frame;
- after `flushFrame()`, phase-2/3 land and `start-cls` is removed.

No timer can race it. Other tests keep the real-timer rAF + `waitForClass`/`nextFrame` helpers — the new mode is opt-in, so their behavior is unchanged.

## Non-tautological (gate-off, #1468)
Neutering the `flushFrame()` drive makes the post-frame assertions fail (`expected true to be false` — `start-cls` persists), proving the driven frame is what advances the transition.

## Verified
Rewritten test passes 3/3; full `dj_transition.test.js` 11/11; pre-push full suite + all hooks green.

### DIFF
--- a/tests/js/dj_transition.test.js
+++ b/tests/js/dj_transition.test.js
@@ -13,7 +13,7 @@ import fs from 'fs';
 
 const clientCode = fs.readFileSync('./python/djust/static/djust/client.js', 'utf-8');
 
-function createDom(bodyHtml = '') {
+function createDom(bodyHtml = '', opts = {}) {
     const dom = new JSDOM(
         `<!DOCTYPE html><html><head></head><body>
             <div dj-view="test.V" dj-root>${bodyHtml}</div>
@@ -27,6 +27,27 @@ function createDom(bodyHtml = '') {
     }
     dom.window.WebSocket = MockWebSocket;
     dom.window.DJUST_USE_WEBSOCKET = false;
+    if (opts.controlledRaf) {
+        // Replace jsdom's timer-backed requestAnimationFrame with a queue we
+        // drive explicitly via `dom.flushFrame()`. dj-transition captures
+        // `globalThis.requestAnimationFrame` at _runTransition time, so a frame
+        // never fires on a real timer — phase transitions advance only when the
+        // test drives them, making the active/end-on-next-frame ordering
+        // deterministic instead of racing the 16 ms rAF fallback (#1830, #1795
+        // family: assert an ordering invariant, never real-frame timing).
+        const rafCbs = [];
+        dom.window.requestAnimationFrame = function (cb) {
+            rafCbs.push(cb);
+            return rafCbs.length;
+        };
+        dom.flushFrame = function () {
+            // Drain the callbacks queued so far (snapshot first so a callback
+            // that schedules another frame doesn't run in the same flush).
+            const due = rafCbs.splice(0);
+            due.forEach((cb) => cb(0));
+            return due.length;
+        };
+    }
     dom.window.eval(clientCode);
     dom.window.document.dispatchEvent(new dom.window.Event('DOMContentLoaded'));
     return dom;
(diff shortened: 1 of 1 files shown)


johnrtipton:
## Retrospective — PR #1839 (#1830, pipeline-drain v1.0.7-1)

**Task**: de-flake the dj-transition active/end-on-next-frame test (rAF-timing race under parallel load).

### Quality: 5/5
Deterministic by construction, gate-off verified, no behavior change, all CI green.

### What went well
- **Removed the timing dependency entirely** rather than widening a tolerance. An opt-in `controlledRaf` stub turns the test fully synchronous and asserts the *ordering invariant* (start now; active/end only after a driven frame) — the #1795-family canon (never assert on real-frame timing; assert an ordering invariant).
- **Traced init timing before fixing.** Confirmed `_installDjTransitionObserver` does a synchronous `querySelectorAll(...).forEach(_runTransition)` sweep, so phase 1 lands synchronously and only phase 2/3 is rAF-deferred — which is exactly what made the fully-synchronous rewrite possible (no awaits at all).
- **Opt-in, no collateral.** Other tests keep the real-timer rAF + `waitForClass` helpers; only the flaky case uses the controllable frame.
- **Non-tautological (gate-off, #1468):** neutering `flushFrame()` fails the post-frame assertions, proving the driven frame is what advances the transition.

### What didn't
- Nothing notable. The flake passed 3/3 in isolation, so the fix was validated by construction (no timer to race) + gate-off rather than by reproducing the intermittent failure.

### Verified
Rewritten test 3/3; full `dj_transition.test.js` 11/11; pre-push full suite + all hooks green; 14/14 CI checks green.

`RETRO_COMPLETE`

PR #1841, opened 2026-06-18
feat(checks): T017 — flag dj-view/dj-root on a table-section element (#1837)

### LINKED ISSUE #1837: tech-debt: system check — warn when dj-view/dj-root is on a table-section element (silently foster-parented)
Surfaced while investigating #1827.

A LiveView whose root element is a table-section element (`<tbody>`, `<thead>`, `<tfoot>`, `<tr>`, `<td>`, `<th>`, `<caption>`, `<col>`, `<colgroup>`) renders to **silently broken** output: html5ever foster-parents the table elements out of the tree at render time. Empirically, a `<tbody dj-view="…">{% for %}<tr>…{% endfor %}</tbody>` template renders as:

```
<html><head></head><body>a<!--dj-if …--><!--/dj-if--></body></html>
```

— all the `<tr>`/`<td>` structure is dropped, leaving only text, with **no error**. (This is also why #1827's `diff_html` flattening is unreachable via a real render — the render stage already destroys the structure.)

## Proposal
Add a system check (T-series, WARNING) that flags `dj-view` / `dj-root` placed on a table-section element, with the fix hint: *put `dj-view` on a wrapping element (e.g. the `<table>` or a surrounding `<div>`); a table-section element cannot be a standalone parse root.* Honor `_is_check_suppressed`. Empirical canary (#1459): a `<tbody dj-view>` template fires the check; a `<div dj-view><table>…` template does not.

Priority: low (DX guardrail for an uncommon mistake that currently fails silently).

johnrtipton:
Closes #1837.

## Problem (empirical render-garbage evidence)

A LiveView whose root element is a table-section element renders to **silently broken** output — html5ever foster-parents the table elements out of the tree at render time, with **no error**. Reproduced directly:

```python
class V(LiveView):
    template='<tbody dj-view="x.V" dj-id="0">{% for r in rows %}<tr><td>{{ r }}</td></tr>{% endfor %}</tbody>'
    def mount(self,request,**k): self.rows=["a"]
    def get_context_data(self,**k): return {"rows": self.rows}
v=V(); v.mount(Req()); html,_,_=v.render_with_diff()
```

renders as:

```
<html dj-id="0"><head dj-id="1"></head><body dj-id="2">a</body></html>
```

The entire `<tr><td>a</td></tr>` structure is dropped, leaving only the text `a`.

## The check

- **ID**: `djust.T017` — **Severity**: WARNING — `python/djust/checks/templates.py`
- **Detection**: regex `_DJ_TABLE_SECTION_ROOT_RE` matching an opening table-section tag (`<tbody>`, `<thead>`, `<tfoot>`, `<tr>`, `<td>`, `<th>`, `<caption>`, `<col>`, `<colgroup>`) carrying `dj-view=` / `dj-root=` **on the same tag**. The `[^>]*?` attribute span stops at the tag's own `>`, so the attribute must be on the table-section tag itself; `\b`/`(?![\w-])` word-boundaries the tag name. Case-insensitive; tolerates attribute order/whitespace.
- **fix_hint**: "Put `dj-view`/`dj-root` on a wrapping element (the `<table>` or a surrounding `<div>`); a table-section element is foster-parented at render time and cannot be a parse root."
- Honors `_is_check_suppressed("djust.T017")` (both `T017` and `djust.T017` forms).

## Empirical canary (#1459)

`python/tests/test_checks_t017_table_section_1837.py` (`TestT017TableSectionRegex`, `TestT017CheckIntegration`):

- **Fires**: `<tbody dj-view="x.V">…` → 1× `djust.T017`; `<tr dj-root>…` → 1× `djust.T017`.
- **Does NOT fire** (no false positive): `<div dj-view="x.V"><table><tbody>{% for %}<tr>…</table></div>` → 0; `<table dj-view="x.V">…` → 0.
- **Suppression**: `T017` and `djust.T017` in `DJUST_CONFIG['suppress_checks']` → 0.
- **Regex unit cases**: word-boundary rejects `<trx` / `<tablefoo`; tolerates attribute order/whitespace/case; `<table dj-view>` (the wrap target) not flagged.

17 tests, all green.

## Gate-off (#1468)

With the T017 emission early-returned (`if True: return`), the fires tests fail (non-tautological):

```
test_fires_on_tbody_dj_view: assert len(t017) == 1
E   assert 0 == 1  +  where 0 = len([])
test_fires_on_tr_dj_root: assert 0 == 1
```

Restored after verification.

## Dogfood (#1060)

`cd examples/demo_project && manage.py check | grep T017` → no match (T017 does not fire spuriously); `manage.py check` exits 0 (121 issues, 0 silenced — unchanged).

## Tests

- `pytest -k "T017 or table_section or check_templates"` → 18 passed.
- `pytest -k "check"` (full checks suite) → 460 passed, 0 failed (no T-series regression).
- Pre-push full suite + ruff passed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,29 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- **System check T017 — warn on ``dj-view`` / ``dj-root`` on a table-section
+  element (#1837).** A new template static-analysis check (severity WARNING)
+  flags a ``dj-view`` or ``dj-root`` attribute placed on an HTML table-section
+  element (``<tbody>``, ``<thead>``, ``<tfoot>``, ``<tr>``, ``<td>``, ``<th>``,
+  ``<caption>``, ``<col>``, ``<colgroup>``). Such a view renders to *silent
+  garbage*: html5ever foster-parents the table elements out of the tree at
+  render time, so ``<tbody dj-view="…">{% for %}<tr>…{% endfor %}</tbody>``
+  renders as ``<html><head></head><body>text</body></html>`` (all rows
+  dropped) with NO error. The matcher is same-tag-scoped (a ``<div dj-view>``
+  wrapping a ``<table><tbody>`` does NOT false-match — the attribute must be on
+  the table-section opening tag itself), word-boundaries the tag name so
+  ``<trx`` / ``<tablefoo`` don't match, and tolerates attribute order /
+  whitespace / case. The fix hint points the developer to a wrapping element
+  (the ``<table>`` or a surrounding ``<div>``). Honors
+  ``DJUST_CONFIG['suppress_checks']`` for both the short (``T017``) and
+  fully-qualified (``djust.T017``) id forms. Covered by ``TestT017TableSectionRegex``
+  and ``TestT017CheckIntegration`` in
+  ``python/tests/test_checks_t017_table_section_1837.py`` (regex word-boundary /
+  attribute-order cases, fires/no-false-positive empirical canary per #1459, and
+  suppression).
+
 ### Fixed
 
 - **WebSocket: recovery version no longer goes stale across non-arming
--- a/docs/system-checks.md
+++ b/docs/system-checks.md
@@ -44,6 +44,7 @@ Run checks with: `python manage.py check --deploy` or `python manage.py djust_ch
 | T013 | Templates | Warning | dj-view with empty or dynamic value |
 | T014 | Templates | Warning | Deprecated data-dj-id attribute |
 | T015 | Templates | Warning | Legacy data-djust-root / data-djust-view root attributes |
+| T017 | Templates | Warning | dj-view / dj-root on a table-section element (foster-parented to silent garbage) |
 | Q001 | Quality | Info | print() statement found |
 | Q002 | Quality | Warning | f-string in logger call |
 | Q003 | Quality | Info | console.log without djustDebug guard |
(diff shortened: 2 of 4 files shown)


johnrtipton:
## Retrospective — PR #1841 (#1837, pipeline-drain v1.0.7-2)

**Task**: add system check T017 — warn when `dj-view`/`dj-root` is on a table-section element (silently foster-parented at render time).

### Quality: 5/5
Empirical-canary validated (the #1459 requirement for a check PR), gate-off non-tautological, dogfood clean, all 14 CI checks green.

### What went well
- **Reproduce-first proved the failure mode before writing the matcher** — confirmed `<tbody dj-view>{% for %}<tr>…` renders as `<html><body>text</body></html>` (rows foster-parented away, no error), so T017 targets the real silent failure rather than a guessed pattern.
- **Empirical canary (#1459)** — fires on `<tbody dj-view>` / `<tr dj-root>`; zero false positives on `<div dj-view><table><tbody>` (attribute on a wrapping element) and `<table dj-view>` (a valid parse root, correctly excluded from the table-*section* list); suppression honored for both `T017` and `djust.T017`.
- **Scoped regex** — `<(tag)(?![\w-])[^>]*?\b(dj-view|dj-root)\b`: the `[^>]` keeps the match on the same opening tag, the `(?![\w-])` word-boundary rejects `<trx`/`<tablefoo`. More precise than the sibling `_DJ_VIEW_RE`.
- **Gate-off (#1468)** — early-returning the emission fails the fires tests (`0 == 1`), proving non-tautological.
- **Dogfood (#1060)** — `manage.py check` clean on the demo (no spurious T017).

### What didn't
- One negligible theoretical edge: `\bdj-view\b` would also match a hypothetical `data-dj-view` attribute on a table-section tag — but that isn't a real djust attribute, so it can't occur in practice. Left as-is (tightening to `(?<![\w-])` would add noise for a can't-happen case).

### Verified
17 tests (TestT017TableSectionRegex + TestT017CheckIntegration); full checks suite 460 passed; `manage.py check` exit 0; docs/system-checks.md updated; 14/14 CI green.

`RETRO_COMPLETE`

Issue #1857, opened 2026-06-22
tech-debt: T1-C: object-permission on the HTTP-API path (verify, then add-or-document)


johnrtipton:
Tier 1 (impossible) / effort S / prevents an IDOR (finding-class #10/#11/#12) reopening on the API transport.

api/dispatch.py:dispatch_api calls check_view_auth + check_handler_permission but NOT enforce_object_permission, while every render path does (runtime.py:651, live_tags.py, mixins/request.py:136).

Step 1 (verify): does any @expose_api / @server_function handler read get_object / self._object (i.e. is object-scoped)?
- If YES: add the auth/core.py:enforce_object_permission call after view-auth (it is fail-closed and a no-op for views without a custom get_object, so nearly free).
- If NO: write a one-paragraph ADR note that the HTTP-API is handler-RPC (no object-scoped render) so the check is intentionally N/A.

Verification: a test asserting an object-scoped handler invoked over the API on a denied object returns 403 (or the ADR note exists if N/A).

Source: approved security-drift prevention plan (post-F1-F29 audit).

Issue #36, opened 2026-01-24
DjustTemplateBackend: urlencode filter not available


johnrtipton:
## Description

The DjustTemplateBackend does not support Django's built-in `urlencode` template filter, which is commonly used for generating share URLs and query parameters.

## Example that fails

```html
<a href="https://twitter.com/intent/tweet?text={{ post.title|urlencode }}">Share</a>
```

Throws: `Template error: Unknown filter: urlencode`

## Expected Behavior

The `urlencode` filter should work the same as in Django's default template backend, URL-encoding special characters in the string.

## Current Workaround

Use JavaScript to build URLs with proper encoding:

```html
<a href="#" data-share="twitter">Share</a>
<script>
document.addEventListener('DOMContentLoaded', function() {
    const url = encodeURIComponent(window.location.href);
    const title = encodeURIComponent(document.title);
    document.querySelector('[data-share="twitter"]').href = 
        'https://twitter.com/intent/tweet?text=' + title + '&url=' + url;
});
</script>
```

## Suggested Solution

Add `urlencode` to the list of supported template filters in the Djust template backend.

## Environment

- djust version: 0.1.5
- Python version: 3.12
- Django version: 4.2

PR #1863, opened 2026-06-22
fix(security): enforce object-permission on the HTTP-API + SSE-legacy mount paths (#1857)

johnrtipton:
## Summary

Closes the object-permission (ADR-017) IDOR gap on the two remaining mount paths: the **HTTP-API** dispatch and the **legacy SSE** mount. The shared `enforce_object_permission` chokepoint was already enforced on WS mount, runtime/`url_change`, HTTP-GET, and `{% live_render %}` (#10/#11/#12) — but `api/dispatch.py` and `sse.py` never called it.

## What was gapped

- **`api/dispatch.py:dispatch_api`** — mounts an object-scoped view + runs `check_view_auth` / `check_handler_permission`, but never the object-level check → an `expose_api` handler ran against a **denied** object (IDOR).
- **`api/dispatch.py:dispatch_server_function`** — same gap on the `@server_function` RPC path.
- **`sse.py:_sse_mount_view`** — 0 object-perm calls → the legacy SSE mount **rendered** a denied object's initial HTML (IDOR).

## The fix (per path)

| Path | Where | Denial envelope |
| --- | --- | --- |
| `dispatch_api` (step 6b) | after `check_view_auth`, before handler lookup | `403 permission_denied` |
| `dispatch_server_function` (step 5b) | after `check_view_auth`, before fn resolution | `403 permission_denied` |
| `_sse_mount_view` | after `mount()`, before initial render | `{"type":"error"}` frame + `return False` (abort), matching the path's auth-denial shape |

The insertion point on every path is **after view-level auth + `mount()`** (so `get_object()`'s access-determining state exists) and **before** the handler runs / the render. `enforce_object_permission` is ALREADY fail-closed and a **pure no-op** for views without a custom `get_object`, so this is behavior-preserving for non-object-scoped views (verified: a `*_noop_for_non_object_scoped_view` test on each path passes unchanged) and a security fix only for object-scoped ones.

No by-design N/A path: a mounted, object-scoped view + bound request exist on all three.

## Reproduce-first + gate-off (#1468)

New file `python/djust/tests/test_object_perm_api_sse_paths.py` (mirrors `test_object_perm_render_paths.py`):

- **Reproduce-first**: the 3 `*_denies_forbidden_object` tests (API dispatch, `server_function`, SSE mount) **FAIL** before the fix (handler runs / object renders) and PASS after.
- **Gate-off (per path)**: neutering each `enforce_object_permission` call turns the matching denies test **red**, confirming non-tautological:
  - API (both sites gated off) → `test_api_dispatch_denies_forbidden_object` + `test_server_function_denies_forbidden_object` fail.
  - SSE (gated off) → `test_sse_mount_denies_forbidden_object` fails (`mounted is True` for a denied object).

## Anti-drift structural pin (#1857, concern-4c)

`python/djust/tests/test_mount_chokepoint_structural.py::TestMountOrchestrationChokepoint` extended to pin `enforce_object_permission` on `sse.py` (>=1) and `api/dispatch.py` (>=2 call sites) so a future removal re-opening #10/#11/#12 on these transports is caught mechanically.

## Verification

- Full suite: **8303 passed, 21 skipped, 0 failed** (`python/djust/tests/ tests/ python/tests/ -n auto`). No new failures.
- `ruff check` clean.
- Two-commit shape (#1173): impl+tests / CHANGELOG.


johnrtipton:
## Stage 11 — Adversarial Code Review: APPROVE (no findings)

Closes the object-perm (IDOR) gap on 3 previously-bypassed paths. Empirically verified:
- **All 3 call sites correctly placed** (object IS available — not a silent no-op): `dispatch_api` step 6b (after `_instantiate_view`+auth, before handler); `dispatch_server_function` step 5b; `_sse_mount_view` after `mount()` before render.
- **Per-path gate-off RED**: neuter each call → its `*_denies_forbidden_object` test fails. API denial = 403 `permission_denied`; SSE = error frame + `return False` (render never reached, no content leak).
- **No-op for non-object-scoped views confirmed** (the `_has_custom_get_object` short-circuit) — behavior-preserving for normal views.
- **concern-4c** catches a removed site (sse>=1, api/dispatch>=2 — pins BOTH API sites).
- Post-incident diff internally consistent (verified all 3 bodies directly; no half-reverted hunk).

Full suite green; CI 12 success. Verdict: **APPROVE**.

johnrtipton:
## Stage 14 retro — PR #1863 (v1.0.8-1 group 4 / T1-C #1857)

**Quality: 5/5** (clean — APPROVE, no fix-pass).

### What went well
- Re-scoped per the T1-A reviewer's finding to cover BOTH gapped paths (HTTP-API + SSE-legacy), and the implementer caught the `dispatch_server_function` twin too (#1646 — fix the class, not one instance): 3 call sites, all no-op for non-object-scoped views so behavior-preserving.
- Reproduce-first across all 3 (denial tests fail pre-fix / pass post-fix); per-path gate-off; concern-4 extended to mechanically pin the new sites.

### What didn't
- A `git checkout <path>` during gate-off reverted the fix mid-run (worktree-restore-class, #36); the implementer re-applied + re-verified, and I independently confirmed the pushed diff is complete (3 sites + tests, internally consistent). Lesson reinforced: never `git checkout <path>` to restore during gate-off — copy aside + restore from the copy.

### Verified
- 3 call sites correctly placed (object available); per-path gate-off RED; no-op confirmed; concern-4c pins both API + SSE; full suite 8303 passed; CI green. Admin-merged.

\`RETRO_COMPLETE\`

Issue #1848, opened 2026-06-22
Inline <script> inside dj-root not executed after WS-mount morph (#1610) — page JS handlers silently never register (1.0.7 regression)


johnrtipton:
## Summary
On a LiveView page, an inline `<script>` placed **inside the dj-root** (anywhere in the rendered LiveView content) is **not executed** after djust mounts and morphs the pre-rendered DOM against the WS-mount HTML (the #1610 step). Any handler that script registers (e.g. a delegated `document` click listener) **silently never registers** — no console error — so page-level interactivity that relies on inline scripts breaks.

Regressed in **1.0.7**: a site that worked on 1.0.5rc3 broke immediately after upgrading.

## Impact
Real-world: this site's `/examples/` page — tab switching **and** code-copy buttons (one inline delegated `document` click handler in the content block) went completely dead after the 1.0.7 upgrade. Any djust app with inline page scripts inside the dj-root is affected.

## Root cause
- Layouts wrap page content in `<div dj-root>` (the morphdom-managed region).
- On mount, djust HTTP-pre-renders, then morphs the pre-rendered DOM against the WS-mount HTML (#1610; console: `[LiveView] Morphed pre-rendered DOM against WS-mount HTML (#1610)`).
- **morphdom does not execute `<script>` tags it inserts/re-creates.** When the morph re-creates the dj-root subtree, the inline `<script>` is re-created but never runs, so its `addEventListener` is never called.

## Reproduction
1. An HTTP-pre-rendered LiveView page with an inline `<script>` inside the dj-root registering a delegated listener (e.g. `document.addEventListener('click', e => { if (e.target.classList.contains('tab-button')) {...} })`).
2. Load the page; click the targeted element.
3. **Expected:** handler runs. **Actual:** nothing happens, no console error.

Browser-verified diagnosis:
- The click bubbles **fully** through `document` (a capture listener AND a late-registered bubble listener both fire), yet the page's own handler never ran → the listener was never registered.
- Re-registering the identical handler at runtime makes clicks work immediately.
- `dj_root.contains(scriptElement) === true`.

## Workaround (applied downstream)
Move the inline `<script>` **outside** the dj-root (a base-template block rendered after the dj-root `</div>`). It then executes once at parse; a delegated `document` listener still catches clicks on the morph-managed elements. Fixed this site's examples page by moving the handler to `{% block extra_scripts %}`.

## Suggested fix
[...]

Issue #1517, opened 2026-05-18
tech-debt: meta-check for \b word-boundary anchors in attribute-matching regexes (checks.py)


johnrtipton:
Source: Retro v1.0.0rc2 (finding #4 — milestone synthesis).

A regex that matches an HTML attribute with a leading `\b` word boundary false-matches `data-*` prefixed attributes — `\b` sits between `data-` and the attribute name because `-` is a non-word char. This bug has now appeared **three times** in `python/djust/checks.py`:

- `_INPUT_TYPE_RE` (Y003, PR #1512) — `\btype` matched `data-type`. Fixed with `(?<![\w-])`.
- `_POSITIVE_TABINDEX_RE` (Y004, PR #1512) — `\btabindex` matched `data-tabindex`. Fixed with `(?<![\w-])`.
- `_IMG_HAS_ALT_RE` (Y002, PR #1491) — `\balt` matches `data-alt` — still latent (separate fix tracked in #1514).

Add a meta-check (a small test, or a `scripts/check-*.py`-style lint) that greps `python/djust/checks.py` and any check module for `\b` anchors immediately preceding an attribute-name literal in a compiled regex, and flags them — `\b` is almost always wrong next to an attribute name; `(?<![\w-])` is the correct anchor. A meta-check would have caught all three occurrences at once.

Relates to: #1514 (the Y002 `_IMG_HAS_ALT_RE` instance fix).

PR #1864, opened 2026-06-22
feat(checks): S009 event-handler-auth + S011 inline-script/CSP system checks (#1854)

### LINKED ISSUE #1854: tech-debt: T2-C: new system checks — S009 event-handler-needs-auth + S011 CSP/inline-script (S010 advisory/opt-in)
Tier 2 (detect) · effort M · catches the author-side instances of secure-default failures. In python/djust/checks/security.py (reuse the S001-S008 AST + _has_noqa + @register('djust') scaffold + _is_liveview_subclass helpers): S009 flags an authenticated view whose public event handlers have no per-handler/class gate; S011 warns on mark_safe/inline-<script> shipping without a CSP setting (targets the #1848 class). S010 rate-limit-presence is advisory/opt-in only (high false-positive risk) or skip. Source: approved prevention plan.

johnrtipton:
Closes #1854 (T2-C of the v1.0.8-1 group-5 secure-default-checks plan).

Adds two new djust system checks in `python/djust/checks/security.py`,
reusing the S001-S008 scaffold (`@register("djust")`, `DjustWarning`,
`_has_noqa`, `_is_liveview_subclass`, AST/template scanning). **S010 is
intentionally skipped** (see below).

### S009 — event-handler-needs-auth (Warning, AST)
Flags a LiveView that declares **view-level** authorization but exposes a
**public `@event_handler` / `@action`** method with **no per-handler gate**.
View-auth only gates the *mount*; a user past it can call any public handler,
so a sensitive handler that needs finer authorization is reachable.

- View-auth signals (any one): truthy `login_required` / `permission_required`
  class attr, a `check_permissions()` override, a Django/djust `AccessMixin`-
  family base (`LoginRequiredMixin` / `PermissionRequiredMixin` / …), or an
  auth-gated `dispatch`.
- Per-handler gate (closes the gap): `@permission_required(...)` on the handler,
  or a class-level `check_handler_permission()` override.
- **Conservative**: private (`_`) handlers and read-only-looking handlers
  (`load_`/`get_`/`list_`/`search_`/…) are exempt; falsy `login_required = False`
  does not count. Suppress via `# noqa: S009` (def or decorator line) or
  `DJUST_CONFIG['suppress_checks']`.

### S011 — inline-script / CSP (Warning, template scan)
Flags an inline **executable** `<script>` **inside a real `dj-root` / `dj-view`
subtree** when no CSP is configured (no django-csp middleware, no
`CONTENT_SECURITY_POLICY` / `CSP_*` / `SECURE_CSP*` setting). Targets the **#1848
class** (morphdom does not re-execute inserted `<script>`, so inline page JS
inside the dj-root silently never runs after the WS-mount morph) *and* the CSP
gap.

- **Low false-positive**: skips `src`-includes, nonce-bearing scripts, and data
  blocks (`application/json`, `text/template`, …); blanks `<pre>`/`<code>`
  example markup; **balances the dj-root subtree** so a page script *after* the
  root close (e.g. a post-root `{% block extra_scripts %}` — the recommended
  pattern) is not flagged.
- Attribute anchors use `(?<![\w-])` not bare `\b` so `data-src`/`data-type`/
  `data-nonce` aren't mistaken for the real attributes (**#1517 hardening** — the
  existing regex-hardening meta-guard enforces this).

### S010 — SKIPPED
The approved prevention plan marks `S010` (rate-limit-presence) as
**advisory/opt-in only** (high false-positive risk). Shipping it default-on
would train developers to blanket-suppress, which is worse than no check.

## Empirical-canary evidence (#1459) + gate-off self-test (#1468)
Per-check synthetic triggers (SHOULD fire) and clean cases (SHOULD NOT fire) are
the tests, in `python/tests/test_checks.py`:

- `TestS009EventHandlerNeedsAuth` (15 cases): 5 trigger variants
  (login_required attr / permission_required attr / auth-mixin base /
  check_permissions override / @action) fire; 8 clean variants (gated handler /
  no view-auth / private / read-only / class-gates-events / falsy attr /
  undecorated method) stay silent; + noqa + config suppression.
- `TestS011InlineScriptCsp` (18 cases): inline-exec-in-dj-root / nested-inside /
  dj-view-root fire; external-src / json / text/template / nonce /
  after-dj-root / in-`<pre>`/`<code>` / non-liveview / 3× CSP-configured stay
  silent; csrf-is-not-csp still fires; `data-*` attrs don't mask a real script;
  + noqa + config suppression + line-number.

**Gate-off self-test**: neutering `_ungated_event_handlers` (S009) fails exactly
its 5 trigger tests; neutering `_dj_root_ranges` (S011) fails exactly its 5
behavior-meaningful trigger tests — both non-tautological.

## Dogfood (#1060)
`manage.py check` against `examples/demo_project` reports **ZERO S009/S011 false
positives**. The first pass surfaced 8 S011 hits; all 8 were page scripts placed
*after* the dj-root close (correct usage), so the check was **tightened** to
balance the dj-root subtree (and skip `<pre>`/`<code>`). After tightening: 0.

## Verification
- Full suite green: **8335 passed, 21 skipped, 0 failed** (`python/djust/tests/
  tests/ python/tests/ -n auto`). The 8 `TestDemoRegistration` tests pass here.
- ruff clean (check + format) on `security.py` and `test_checks.py`.
- Branch up to date with `origin/main` (0 behind).
- Two-commit shape (#1173): impl+tests, then CHANGELOG + `docs/system-checks.md`.


johnrtipton:
## Stage 11 — Adversarial Code Review: APPROVE (no findings)

Tooling PR, empirical-canary verified independently: S011 fires on a reviewer-constructed #1848-shape trigger (inline `<script>` inside `<div dj-root>{% block content %}…</script>…{% endblock %}`) and stays silent on 5 legit shapes (post-root script, json data block, src-include, nonce'd, pre/code). S009 fires on a reviewer-built `login_required=True` + ungated `delete_account` handler and stays silent on legit views. Independent broad dogfood: 131 demo HTML + 168 .py (85 LiveViews) → **0 false positives**. Gate-off non-tautological (neuter predicate → exactly the trigger tests red). #1517 anchor hardening verified. Two-commit shape; docs/system-checks.md documents both. Full suite green. Verdict: **APPROVE**.

johnrtipton:
## Stage 14 retro — PR #1864 (v1.0.8-1 group 5 / T2-C #1854)

**Quality: 5/5** (clean — APPROVE).

### What went well
- The dogfood (#1060) earned its keep: the first S011 pass flagged 8 correctly-placed demo scripts (after dj-root close) → drove tightening to positional dj-root-subtree balancing + pre/code blanking → 0 FP. A check that false-fires trains blanket-suppression, so the dogfood IS the gate for tooling PRs.
- Empirical canary (#1459): both the impl AND the adversarial reviewer constructed the real #1848 inline-script-in-dj-root shape and confirmed S011 catches it — the doc claim 'this check catches X' got an executable verifier.
- Caught + fixed a #1517 regex-meta-guard trip (`\b`→`(?<![\w-])`), which is also more correct (data-src no longer masks a script).

### Follow-up filed
- The reviewer found `_get_project_app_dirs()` returns 0 dirs when run from inside the djust repo tree (the `/djust/` path filter) — pre-existing, makes dogfooding ALL S/V checks narrow. Filed separately.

### Verified
- S009 + S011 fire on synthetic triggers + silent on legit; gate-off non-tautological; broad dogfood 0 FP; full suite 8335 passed; CI green. Admin-merged.

`RETRO_COMPLETE`

Issue #1855, opened 2026-06-22
tech-debt: T2-D: make bandit blocking on new high-severity findings + add a tiny blocking browser-smoke


johnrtipton:
Tier 2 (gate) · effort S-M · closes the runtime-break blind spot a green pytest suite cannot see (#1848/#1849). (1) Promote bandit in .github/workflows/pre-release-security-audit.yml to block on new high-severity findings (today only eslint blocks; rest are || true) — keep safety/pip-audit/cargo-audit advisory. (2) Add a 2-3 page blocking Playwright smoke (demo index mounts; an examples-style dj-root inline-script page's controls fire) — current Playwright leg is non-blocking (test.yml:240). Cap scope to canary pages. Source: approved prevention plan.

PR #1871, opened 2026-06-22
fix(client): re-execute classic <script> on the #1610 mount morph (#1848)

### LINKED ISSUE #1848: Inline <script> inside dj-root not executed after WS-mount morph (#1610) — page JS handlers silently never register (1.0.7 regression)

johnrtipton:
## Summary

Closes #1848 (a 1.0.7 regression). An inline classic `<script>` placed **inside the `dj-root`** silently never ran after djust mounts — so any handler it registers (e.g. a delegated `document` click listener for tab switching / code-copy buttons) never wired up, with **no console error**. Worked on 1.0.5rc3; broke on the 1.0.7 upgrade.

## Root cause

On mount djust HTTP-pre-renders, then **morphs** the pre-rendered DOM against the WS-mount HTML (#1610: `morphChildren` in `03-websocket.js`). The non-prerendered branch instead assigns `container.innerHTML = data.html`. Per HTML spec, a `<script>` inserted by **clone+insert** (morph) **or** by **`innerHTML`** is parsed but **NOT evaluated** — the browser only runs a `<script>` that is HTML-parsed in-stream or re-created via `document.createElement`. So the inline `<script>` was re-created on mount but never executed; its `addEventListener` was never called.

This is the parallel-path twin (#1646) of the classic-script lifecycle the framework already protects on the navigation/bfcache path — the #1635/#1650 IIFE wrap keeps the framework's *own* bundle re-executable, but nothing re-ran *user* page scripts after the mount morph. Both mount branches (`skipMountHtml` morph + non-prerender `innerHTML`) had the gap.

## Fix

A single `window.djust._runInsertedScripts(container)` helper (`python/djust/static/djust/src/03-websocket.js`) re-creates each **classic** `<script>` via `document.createElement('script')` + copy attributes + `textContent` + `replaceWith` — the only DOM operation that makes the browser run an already-in-tree inert script. It is called after **both** mount branches (the #1646 structural cure: one helper, both paths).

- **Classic-only:** empty `type` or a JS MIME type. `type="djust/hook"` (colocated hook definitions, extracted separately by the hook system) and `application/json` / `importmap` are left untouched.
- **Idempotent:** each re-created element is marked `data-djust-script-ran` so a WS reconnect / re-mount on the same DOM does not double-execute.
- **CSP-safe:** re-creates *user-authored* scripts that were already in the server HTML; emits **no** framework-generated inline script.

This also closes the #1855 browser-smoke inline-script canary's tolerated known-xfail (it becomes a hard regression guard now that #1848 has landed).

## Tests (repro-first + gate-off)

New: `tests/js/mount-morph-script-exec-1848.test.js` (6 cases). Reproduction fidelity per #1650/#1724: the `dj-root` is built via `innerHTML` **with inter-element whitespace** + **real `<script>` nodes** (not `eval`, which scopes `const`/`let` to the eval call and does not reproduce), and the test drives the **real helper from the built bundle**. Cases:

- helper is exported on `window.djust`
- an inline classic `<script>` re-created by the morph executes
- a delegated `document` click listener registers so morph-managed `.tab-button` clicks fire (the reporter's exact scenario)
- idempotent — a re-run (reconnect / re-mount) does not double-execute
- `type="djust/hook"` colocated definition scripts are skipped (not run as classic JS)
- an external classic `<script src>` is re-created as a fresh element

**Reproduce-first:** all 6 failed before the fix (helper absent).
**Gate-off self-test (#1468):** with `_runInsertedScripts` neutered (`if (true) return;`), **4 of 6** cases go red — non-tautological.
**Full JS suite:** 151 files / 1743 tests green. eslint (`--max-warnings 0`), bundle-init-order, and cross-IIFE checks all clean.

## Size

Isolated **gzipped bundle delta: 172 bytes** (`client.min.js.gz`, same build env) — well under the 2 KB budget.

## Files changed

| File | Change |
|------|--------|
| `python/djust/static/djust/src/03-websocket.js` | `_runInsertedScripts` helper + call after both mount branches + namespace export |
| `python/djust/static/djust/client.js` + `client.min.js{,.map,.gz,.br}` | regenerated bundle |
| `tests/js/mount-morph-script-exec-1848.test.js` | new (6 repro-first cases) |
| `CHANGELOG.md` | `[Unreleased] ### Fixed` entry |


johnrtipton:
## Stage 14 retro — PR #1871 (#1848)

- **Bug-report triage win (#1206 rule):** the issue claimed "djust already re-executes classic scripts on the live_redirect/navigation morph path (#1635, #1650)." Tracing showed #1635/#1650 was *"wrap client.js bundle in an IIFE to survive re-execution"* — the framework's OWN bundle, NOT user page scripts. No prior script-re-execution helper existed, and the navigation path routes through the SAME mount handler. Trusting the symptom over the cited mechanism kept the fix correct.
- **Parallel-path-drift (#1646):** BOTH mount branches (`skipMountHtml` morph + non-prerender `innerHTML`) had the gap; the structural cure is one shared `_runInsertedScripts` helper called after both, not a point fix on the morph branch only.
- **Reproduction fidelity (#1650/#1724):** a standalone probe confirmed jsdom faithfully reproduces the browser semantics — innerHTML+appendChild leaves `<script>` inert; createElement+replaceWith runs it. The test builds the dj-root via `innerHTML` with inter-element whitespace + real `<script>` nodes (not eval).
- **Gate-off (#1468):** neutering the helper turned 4/6 new cases red — non-tautological.
- Full JS suite green (151 files / 1743 tests); eslint + bundle-init-order + cross-IIFE clean. Gzipped delta 172 B.
- Closes the #1855 browser-smoke inline-script canary's tolerated known-xfail.

`RETRO_COMPLETE`

johnrtipton:
## Stage 11 — Adversarial Code Review: APPROVE

Read-only review (`gh pr diff` + `git show` of the branch src; no checkout, #1804-safe).

- **Symptom-up correction stands.** The agent verified the issue's premise ("#1635/#1650 already re-executes classic scripts") was wrong — #1635/#1650 was the framework bundle's IIFE wrap, not user page scripts; no prior helper existed. Trusting the symptom over the cited path (canon) — correct call.
- **`_runInsertedScripts` is the right shape**: `createElement('script')` + copy attrs/textContent + `replaceWith` (the only DOM op that runs an in-tree inert script); classic-only (skips `type="djust/hook"` + module/non-JS); idempotent via `data-djust-script-ran`; CSP-safe (no framework inline script). Called after **both** mount branches — the #1646 parallel-path cure (and covers SPA-nav since nav routes through mount).
- **Gate-off** (#1468): neutering the helper turns 4/6 new cases red. Reproduce-first built the DOM via `innerHTML` with real `<script>` + inter-element whitespace (not `eval`, #1650 fidelity). 1743 JS tests green; +172 B gzipped (well under budget).
- **Non-blocking observation**: the helper runs ALL classic scripts in the container. The empirical evidence (the bug = scripts never run on djust's mount flow; downstream confirmed by moving JS out of dj-root) says dj-root scripts are NOT browser-executed on initial parse, so this is single-execution. IF a future flow ever parse-executes a dj-root script AND the morph preserves it, that node (lacking the marker) could double-run — worth a follow-up guard only if that flow appears. Not a blocker for the reported regression.

Verdict: **APPROVE**. Checking CI before merge.

johnrtipton:
## Retrospective — PR #1871 (pipeline-drain v1.0.8-2 / WU2 #1848)

**Quality: 5/5** (clean, APPROVE).
### What went well
- Symptom-up correction: the issue's premise ("#1635/#1650 already re-executes classic scripts") was wrong — that was the framework bundle's IIFE wrap, not user page scripts; no helper existed. Trusting the symptom over the cited path (canon) avoided a no-op "extend the existing helper" PR.
- `_runInsertedScripts` is the right shape (createElement + copy attrs/text + replaceWith — the only DOM op that runs an in-tree inert script), called after BOTH mount branches (#1646 cure, covers SPA-nav too); classic-only, idempotent marker, CSP-safe, +172B gzipped.
- Reproduce-first (#1650 fidelity: real <script> via innerHTML with whitespace, not eval) + gate-off (#1468: neutering → 4/6 red). Flips the #1855 browser-smoke xfail to a hard guard.
### What didn't
- The merge needed two CHANGELOG union-resolves (sequential [Unreleased] collisions across the 5-PR drain) — the known parallel-worktree hazard; resolved by merging main in + keeping all entries.
### Verified
- 1743 JS tests green; CI green (python/demo/js/security/nav-hooks); two-commit shape.

PR #1877, opened 2026-06-23
fix(security): resolve 6 CodeQL alerts (clear-text logging + log-injection + cleanups)

johnrtipton:
Resolves all **6 open CodeQL code-scanning alerts**.

## Security (1 HIGH + 2 MEDIUM)
| Alert | Rule | Sev | Fix |
|-------|------|-----|-----|
| 2421 | `py/clear-text-logging-sensitive-data` | **HIGH** | `_client_ip.py` misconfig warning logs `type(raw).__name__`, not the raw `DJUST_TRUSTED_PROXY_COUNT` settings value (CodeQL treats `settings` reads as sensitive). Type is the actionable diagnostic. |
| 2422 | `py/log-injection` | MEDIUM | `api/openapi.py:_openapi_gate` wraps `request.path` with `sanitize_for_log` before logging. |
| 2423 | `py/log-injection` | MEDIUM | `observability/views.py:_gate` wraps `request.path` with `sanitize_for_log` before logging. |

Both log-injection fixes use the **existing** `djust._log_utils.sanitize_for_log` (CR/LF/control-char strip + truncate) already applied across `api/dispatch.py` — no parallel-path drift (#1646).

## Code-quality (3 Note)
| Alert | Rule | Fix |
|-------|------|-----|
| 2418 | `py/unused-global-variable` | Removed unused `logger` + `import logging` from `checks/utils.py`. |
| 2463 / 2464 | `py/ineffectual-statement` | `Transport` Protocol `...` stubs → one-line docstrings in `runtime.py` (documents the protocol; concrete transports untouched). |

## Verification
- New `python/djust/tests/test_codeql_log_hardening.py` — 3 reproduce-first security cases (bad proxy-count value; CRLF-injected path on each gate). **Gate-off verified (#1468):** reverting each fix turns its test red (3 fail), the value-independent test stays green.
- Full pytest suite: **8153 passed, 18 skipped, 0 failed** (`-n auto`).
- ruff clean on all 5 edited files.

Closes CodeQL alerts 2421, 2422, 2423, 2418, 2463, 2464.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1859, opened 2026-06-22
test(security): extend anti-drift nets — transport parity axes + mount-orchestration pin (#1850, #1851)

johnrtipton:
## Summary

TEST-ONLY change. Extends the two existing WU1 anti-drift test nets so a future regression that lets one transport's security control drift from the others — or a #1853 migration that silently re-grows a parallel mount orchestration — is caught **mechanically** (#1646). Mirrors each file's existing patterns exactly (per-transport adapter dicts + `@pytest.mark.parametrize` for parity; AST count-canary for the structural pin).

No production source touched; no API or behavior change.

## Issue → file → axis/assertion mapping

| Issue | File | Added | Asserts |
|-------|------|-------|---------|
| #1851 (T2-A) | `python/djust/tests/test_transport_parity_security.py` | `TestViewAuthParity` | `djust.auth.core.check_view_auth` deny verdict IDENTICAL across ws/runtime/sse (login_required, anonymous caller) |
| #1851 (T2-A) | same | `TestObjectPermissionParity` | `djust.auth.core.enforce_object_permission` deny verdict IDENTICAL across transports (`has_object_permission` → False) |
| #1851 (T2-A) | same | `TestRateLimitParity` | `djust.rate_limit.caller_key` + `handler_rate_check` `@rate_limit` trip-point IDENTICAL across transports (F27/F28 shared store) |
| #1851 (T2-A) | same | `TestOriginParity` | foreign Origin/Host REJECTED identically — `_is_allowed_origin` (WS) / `_sse_origin_allowed` (SSE) / `validated_host_from_scope` (runtime), all via the shared `_host_in_allowed_hosts` |
| #1850 (T2-B) | `python/djust/tests/test_mount_chokepoint_structural.py` | `TestMountOrchestrationChokepoint` (concern 4) | AST count-canary: `websocket.py` + `runtime.py` each still reference the shared mount-orchestration security calls (`check_view_auth`, the object-perm family, `validated_host_from_scope`). Docstring notes #1853 convergence; updating `_MOUNT_ORCHESTRATION_PINS` is the deliberate signal that handle_mount delegated to `dispatch_mount`. |

[...]

PR #1878, opened 2026-06-23
fix(security): give sanitize_for_log a CodeQL-recognized CR/LF barrier (#2465/#2466)

johnrtipton:
Follow-up to #1877. The first log-hardening pass wrapped the openapi + observability gate-rejection logs in `sanitize_for_log(request.path)`, but CodeQL **re-flagged** them (`py/log-injection`, alerts **2465 / 2466** — the re-numbered 2422 / 2423 at the shifted lines).

## Root cause
`sanitize_for_log`'s `isprintable()` loop maps line breaks to `?` at runtime, but that comprehension is **not a modeled CodeQL barrier** — so CodeQL still traces `request.path` (a remote source) through the helper to the `logger.warning` sink. (`api/dispatch.py`'s `sanitize_for_log(view_slug)` isn't flagged only because `view_slug` isn't a modeled remote source like `request.path`.)

## Fix
Make the helper's **return value** an explicit `.replace("\r", "").replace("\n", "")` — the form CodeQL recognizes as a log-injection sanitizer. Because the `isprintable` loop already removed line breaks, this is a **runtime no-op** (all existing behavior + tests unchanged), but it gives CodeQL the barrier on the function's return so **every** `sanitize_for_log(<remote source>)` call site clears — DRY, no call-site changes (#1646).

## Verification
- `test_log_sanitizer_barrier_pin` (new) pins the barrier so a "looks redundant, remove it" refactor can't silently re-open the alert (#1859) — **gate-off verified**: reverting the `.replace` turns the pin red.
- Existing `test_log_sanitization.py` (CR/LF→`?`, tab, null, unicode, truncate) + `test_codeql_log_hardening.py` gate tests all green (13 passed); 207-case sanitize/gate/dispatch slice green; ruff clean.

Closes CodeQL alerts 2465, 2466.


johnrtipton:
## Retrospective — PR #1878 (sanitize_for_log CodeQL barrier)

**Quality: 5/5.**
### What went well / lesson
- The real lesson: a sanitizer that is *runtime-correct* is not necessarily a *CodeQL-modeled barrier*. The `isprintable()` loop strips newlines but CodeQL doesn't trace it, so #1877's wrap didn't clear `py/log-injection` — it just re-numbered the alerts when the lines shifted. The fix is the **recognized form** (`.replace` newline removal) as the function's return, which CodeQL propagates interprocedurally to every call site (DRY, #1646).
- Pinned the (runtime no-op) barrier with `test_log_sanitizer_barrier_pin` so it can't be removed as "redundant" and silently re-open the alert (#1859 — load-bearing pin must go red; gate-off verified).
### What didn't
- Cost one extra CodeQL cycle (#1877 → this) because the first fix assumed the existing helper was a recognized barrier. General rule for tooling-recognized fixes: confirm the *analyzer* accepts the form, not just that the runtime behavior is correct.
### Verified
- 13 sanitize+gate tests + 207-case slice green; CI green (incl. Analyze Python); two-commit shape.
`RETRO_COMPLETE`

PR #1897, opened 2026-06-24
test(ws): real-WebsocketCommunicator event parity net for the 2.3 flip (#1896, ADR-022)

### LINKED ISSUE #1896: v1.1.0-3 Iter 2 Phase 2.3 (PR 1/N): 6 real-WebsocketCommunicator event gap-tests against the bespoke path (flip parity net)
Source: v1.1.0-3 ViewRuntime convergence, Phase 2.3 final-flip prep (ADR-022). FIRST PR of 2.3.

**Why first:** before flipping WS events onto dispatch_event (2.3b), characterize what the flip must preserve. These 6 tests run against the CURRENT bespoke `_handle_event_inner` path (they pass NOW), and must stay green AFTER the flip — that equivalence IS the parity proof (#1466/#1780/#1468). The 2.1/2.2 runtime tests use a MockTransport, not a real communicator, so they don't characterize the live WS path.

**6 real `WebsocketCommunicator` tests** (against `LiveViewConsumer.as_asgi()`), each asserting intermediate state + a gate-off sibling (#1468):
1. **component_id routing** — real WS `event` frame with component_id → component_event/full-HTML frame (MockTransport-only today).
2. **time-travel-on-event** — DEBUG + opt-in → a `time_travel_event` frame follows the update.
3. **dj_activity deferral** — event on a hidden activity region → noop + no render; (and when visible → drains in the same round-trip). NOT covered anywhere today.
4. **actor-path events** — `use_actors=True` view over real WS → patch frame (highest value — the load-bearing actor-hook gap). NOT covered via dispatch today.
5. **ref echo** — `ref` echoed on noop AND patch AND html_update.
6. **_force_full_html** — handler sets it → html_update (patches discarded), flag consumed.

**HARD: tests ONLY** — no production change. They go green against the bespoke path. Conservative + low-risk; this is the regression net the flip rides on.

johnrtipton:
## What

Phase 2.3a **parity net** (first PR of the final-flip prep, ADR-022): 6 behaviors / **11 tests** (each behavior + a gate-off or contrast sibling, #1468) characterizing what the CURRENT bespoke `_handle_event_inner` WS event path produces, against a **real `WebsocketCommunicator`** (`LiveViewConsumer.as_asgi()`) end-to-end.

Phase 2.3b will flip ALL WS events from the bespoke path onto `ViewRuntime.dispatch_event` by adding `"event"` to `RUNTIME_OWNED_VERBS`. The existing 2.1/2.2 runtime tests (`test_runtime_child_routing_1892.py`) only exercise a `MockTransport`, never the live WS path — so they don't characterize what the flip must preserve. These do. They pass NOW; 2.3b requires them to stay green after the flip — that equivalence IS the parity proof (#1466 / #1780 / #1468).

**Test-only. Zero production change** (`git status` = one new file; `_rust.so` is gitignored).

## The 6 behaviors (file: `python/djust/tests/test_ws_event_flip_parity_1896.py`)

| # | Behavior | Positive test asserts (bespoke path) | Gate-off / contrast sibling |
|---|----------|--------------------------------------|-----------------------------|
| 1 | **component_id routing** | The component handler resolves + runs (component state mutates, parent notified), but the bespoke path returns an **`error`** frame (NOT a parent patch / component html_update) | unknown `component_id` → `Component not found` error (distinct cause — proves the positive event genuinely resolved a real component) |
| 2 | **time-travel-on-event** | event on a `time_travel_enabled` view (DEBUG) emits a `time_travel_event` frame **alongside** the render (`patch`) frame; entry carries `event_name="bump"` | identical event on a view WITHOUT the buffer emits NO `time_travel_event` frame |
| 3 | **dj_activity deferral** | event on a HIDDEN activity → bare `noop` (deferred, no render); a later show-handler that flips visible **and** changes public state drains the queued event in the SAME round-trip (2nd render frame, `event_name="bump"`, `n=1`) | the SAME `bump` with no `_activity` marker renders immediately (not deferred) |
| 4 | **actor-path events** | `use_actors=True` view mounts a Rust SessionActor over real WS; event routes through `actor_handle.event` and returns a `patch` frame with `event_name` + consumer-owned wire version (mount+1) | (positive itself is the highest-value gap: actor dispatch was untested via the live consumer — the Phase-5 actor tests call `handle.event` directly) |
| 5 | **ref echo** | client `ref` echoed on all three event response shapes: `noop`, `patch`, `html_update` (three distinct handlers / send paths) | a ref-less event → response carries NO `ref` key |
| 6 | **`_force_full_html`** | handler setting `_force_full_html` → `html_update` (patches discarded, full HTML); flag is single-shot — next event renders a normal `patch` | the same text change WITHOUT the flag → `patch`, not `html_update` |

## 2.3a finding (component_id divergence)

The bespoke `_handle_event_inner` `component_id` branch resolves + runs the component handler but **never re-renders the parent**: `html` stays `None` (initialized at the top of the non-actor block, never reassigned in the component branch), so the html_update fallback strips `None` and raises `TypeError: expected string or bytes-like object, got 'NoneType'` → `error` frame. Verified against BOTH a synthetic parent + LiveComponent **and** the real demo `ComponentDemoView`. The Phase-2.1 runtime path (`test_runtime_child_routing_1892.py::TestRuntimeComponentRouting`) renders correctly and emits a parent-scoped `html_update`.

**Implication:** the flip will CHANGE this axis (`error` → `html_update`). `TestComponentIdRouting` therefore pins the *current* bespoke behavior and its docstring flags that the test MUST be updated at 2.3b — that update is the parity signal that the flip fixed the component path. Everything needed (#3 dj_activity, #4 actor) WAS characterizable as a pure test against the bespoke path; only #1's bespoke behavior is an error (documented, not "fixed" here per scope).

## Verification

- All 11 pass against the current bespoke path.
- Gate-off self-test (#1468): neutering `force_once` so it doesn't set `_force_full_html` makes the force test's `_receive_until("html_update")` never find an html_update (it produces a `patch`) → the positive assertion fails. Non-tautological.
- Full suite: **8228 passed, 18 skipped, 0 failed**.
- Zero production files changed.

Closes #1896

### DIFF
--- a/python/djust/tests/test_ws_event_flip_parity_1896.py
+++ b/python/djust/tests/test_ws_event_flip_parity_1896.py
@@ -1,0 +1,757 @@
+"""Real-``WebsocketCommunicator`` event parity net for the 2.3 flip (#1896, ADR-022).
+
+Phase 2.3b will flip ALL WS events from the bespoke ``_handle_event_inner``
+(``websocket.py``) onto ``ViewRuntime.dispatch_event`` by adding ``"event"`` to
+``RUNTIME_OWNED_VERBS``. Before that flip lands, this file CHARACTERIZES what
+the flip must preserve: the observable response frames the *current* bespoke
+WS event path produces for six behaviors that the existing Phase-2.1/2.2 runtime
+tests cover only against a ``MockTransport``
+(``test_runtime_child_routing_1892.py``), NOT a live consumer.
+
+Every test here drives a real channels ``WebsocketCommunicator`` against
+``LiveViewConsumer.as_asgi()`` end-to-end (mount → event frame → assert on the
+response frame), so it exercises the WS path the flip will replace. These pass
+NOW; 2.3b requires them to stay green after the flip — that equivalence IS the
+parity proof (#1466 / #1780 / #1468). Where the bespoke path diverges from the
+runtime path (component_id — see ``TestComponentIdRouting``), the test pins the
+bespoke behavior verbatim and the docstring flags the divergence as a 2.3a
+finding: that test will need to be UPDATED at the flip, and the update is itself
+the signal that the flip changed behavior on that axis.
+
+Each behavior carries a gate-off / contrast sibling (#1468) proving the
+assertion distinguishes the real behavior from a vacuous one (a handler that
+does NOT set ``_force_full_html`` patches instead of html_update; an event with
+no ``ref`` echoes none; a non-deferred event renders immediately).
+
+Harness lifted from ``test_ws_send_version_1788.py`` /
+``test_sticky_child_event_noop_1802.py``.
+"""
+
+from __future__ import annotations
+
+import pytest
+from asgiref.sync import sync_to_async
+
+from djust import LiveView
+from djust.components.base import LiveComponent
+from djust.decorators import event_handler
+
+
+# ---------------------------------------------------------------------------
+# Harness
+# ---------------------------------------------------------------------------
+
+
+async def _receive_until(communicator, wanted_type, *, tries=8, timeout=3):
+    """Drain frames until one whose ``type`` == ``wanted_type`` (or last seen)."""
+    last = None
+    for _ in range(tries):
+        last = await communicator.receive_json_from(timeout=timeout)
+        if last.get("type") == wanted_type:
+            return last
+    return last
+
+
+async def _drain_available(communicator, *, max_frames=6, timeout=2):
+    """Best-effort drain of any frames already on the wire.
+
+    Stops at the first receive timeout (no more frames). Crucially uses
+    ``communicator.receive_nothing``-style polling via a bounded receive so a
+    trailing timeout does NOT leave the application's receive future in a
+    cancelled state (which would make a subsequent ``disconnect()`` raise
+    ``CancelledError``). Returns the list of frames collected.
+    """
+    frames = []
+    for _ in range(max_frames):
+        # receive_nothing returns True if no frame arrives within ``timeout``;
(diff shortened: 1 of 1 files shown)


johnrtipton:
## Retrospective — PR #1897 (v1.1.0-3 Phase 2.3 PR 1/N / #1896)

**Quality: 5/5** (characterization-first did exactly what it's for).
### What went well
- Writing the parity net against the bespoke path BEFORE the flip (the #1466/#1780 characterization pattern) immediately surfaced a latent production bug: bespoke `component_id` events return an `error` frame (TypeError on None html — parent never re-rendered), confirmed against the real demo ComponentDemoView. The flip FIXES it (runtime renders a parent-scoped html_update). Without writing these tests first, the flip would have looked like a regression (error→html_update) at 2.3b review instead of a known improvement.
- 5 of 6 behaviors pinned as pure passing tests; the 1 that couldn't (component_id) became a documented finding + a deliberate 2.3b-update marker. dj_activity (#3, previously untested anywhere) and actor-path (#4, untested via dispatch) both characterized with real round-trips.
- Gate-off siblings (#1468) make each behavior non-vacuous.
### Lesson (canon candidate)
- Characterization tests against the OLD path are the right first PR of any all-or-nothing routing flip — they convert "did the flip preserve behavior?" from a review judgment into a mechanical green/red, AND they surface latent bugs in the old path that the new path silently fixes (which would otherwise read as flip regressions).
### Verified
- CI green (py3.12+py3.14t+demo/js/security/nav-hooks/rust); 8228 passed; zero production files changed; gate-off non-tautological.
`RETRO_COMPLETE`

Issue #1899, opened 2026-06-24
v1.1.0-3 Iter 2 Phase 2.3a (PR 2/N): transport.event_context() — borrow consumer render-lock + origin-channel + observability; delete dead runtime._render_lock


johnrtipton:
Source: v1.1.0-3 ViewRuntime convergence, Phase 2.3a (ADR-022). The foundational fold the dj_activity re-dispatcher + the 2.3b flip sit on.

**Two findings from flip-scope this implements:**
1. `ViewRuntime._render_lock` (runtime.py:377) is DEAD — declared, never acquired anywhere (grep-confirmed: it appears only at :377). DELETE it.
2. The real render serialization is consumer-owned: `self._render_lock` (websocket.py:619) + `_processing_user_event` (:622), SHARED with the WS-only _run_tick/server_push/db_notify loops. A runtime-local lock can't serialize against ticks → the #560 interleave bug. So the runtime must BORROW the consumer's lock via a transport hook.

**Implement `transport.event_context(view)` (async context manager) on the Transport protocol + both adapters:**
- **WSConsumerTransport**: acquire `consumer._render_lock` + set `_processing_user_event=True` + set push origin-channel (`_djust_push.origin_channel.set(consumer.channel_name)`) + start PerformanceTracker (`PerformanceTracker.set_current(...)`) + SQL capture_for_event; on exit (finally) reset origin token + `_processing_user_event=False` + release lock. Mirror websocket.py:3393/3394/3400/3153 (enter) and :4311/4312/4313 (exit) VERBATIM.
- **SSESessionTransport**: no-op async CM (yield).
- Wire `_dispatch_event_inner` (runtime.py:689) event render to run inside `async with self.transport.event_context(view):`.

**Zero-risk confirmation (state in PR):** `dispatch_url_change` (runtime.py:1437) is a SEPARATE method from `_dispatch_event_inner` — do NOT touch it. So wiring event_context into _dispatch_event_inner affects ONLY: SSE events (event_context is no-op for SSE) + WS events (not routed until 2.3b). url_change unaffected.

**HARD:** NO change to `RUNTIME_OWNED_VERBS`/WS routing. Do NOT modify `_handle_event_inner`'s inline lock/origin/observability (it keeps using them until 2.3b deletes it). The WS event_context MUST borrow the consumer's EXISTING lock object, not create a new one.

[...]

PR #1902, opened 2026-06-24
feat(runtime): actor-event transport hook — dispatch_actor_event (#1901, ADR-022 Iter 2 Phase 2.3a)

### LINKED ISSUE #1901: v1.1.0-3 Iter 2 Phase 2.3a: actor-event transport hook (transport.dispatch_actor_event) — the post-flip actor-desync guard
ADR-022 2.3a. dispatch_event has NO actor branch; the use_actors guard is only in dispatch_mount (refuses SSE). A WS view mounts with use_actors=True (bespoke, sets actor_handle), then post-flip its events would hit dispatch_event with no actor branch → silently runs the handler in-process, desyncing the actor diff baseline. Fix: transport.uses_actors(view) + transport.dispatch_actor_event(...) hook (WS runs the bespoke actor block vs consumer.actor_handle; SSE returns False/never-called). Dormant until 2.3b; tested via direct-runtime + gate-off now, validated live by 2.3b's actor-event WebsocketCommunicator test.

johnrtipton:
## Summary

ADR-022 Iter 2 Phase 2.3a — the load-bearing fold the Phase 2.3b WS-event flip sits on.

`ViewRuntime.dispatch_event` had **NO actor branch**, while the `use_actors` guard lived ONLY in `dispatch_mount` (which refuses SSE outright). A WS view mounts in actor mode (`use_actors=True` + a created `actor_handle`); once Phase 2.3b routes WS events through the runtime, such a view's events would hit `dispatch_event` with no actor branch and **silently run the handler in-process** via the normal render path, **desyncing the actor's server-side diff baseline**. This PR adds the hook that closes that gap.

## The hook shape

Two new `Transport` Protocol methods + both adapters:

- `uses_actors(self, view) -> bool`
  - **WS**: `consumer.use_actors and consumer.actor_handle is not None` — the exact precondition of the bespoke WS actor block (`websocket.py:3282`).
  - **SSE**: `False` — no bidirectional actor channel; `dispatch_mount` refuses `use_actors` mounts (`runtime.py:602`), so the branch is unreachable.

- `async dispatch_actor_event(self, view, event_name, params, *, event_ref=None, cache_request_id=None) -> None`
  - **WS**: runs the bespoke WS actor block (`websocket.py:3282-3379`) **VERBATIM** against the consumer — time-travel record/push in a `finally`, the shared `_validate_event_security` + `validate_handler_params` checks, `actor_handle.event()`, patch/HTML framing, error handling, and the v0.7.0 deferred-activity flush.
  - **SSE**: raises `NotImplementedError` (never called — `uses_actors` is `False`).

## Matching the WS guard order (actor vs embedded-child)

Wired into `_dispatch_event_inner` **BEFORE `event_context`** (the actor block holds no render lock, matching WS). The actor path applies when `uses_actors(view)` **AND** the event is **NOT routed to a sticky child** — the WS `not is_embedded_child_target` mutual exclusion (`websocket.py:3280-3282`).

Per **#1467**: a `component_id` event does **not** reassign the target view, and the WS actor block has **no** component handling — so a `component_id` event on a `use_actors` view goes through the **actor** (parity with WS), and **only** a `view_id` resolving to a *different* child excludes it. `_event_routes_to_sticky_child` peeks at `view_id` **without consuming it**, so the non-actor sticky-child routing (run inside `event_context`) still pops it.

## Version-stamping (what 2.3b must watch)

The consumer OWNS the monotonic wire version (#1788); the actor's `result['version']` is **IGNORED** for the wire — `_send_update` is stamped with `consumer._next_version()` (the same source `handle_event` uses). The actor's internal version still drives its server-side diff baseline. `patches` may arrive as a JSON **string** from Rust and is parsed via `fast_json_loads` before send.

## Why it's dormant / zero live-risk

`uses_actors` is `False` for both live transports today (WS events still run on the bespoke `_handle_event_inner`; SSE refuses actor mounts), so **no live event turn reaches the hook until 2.3b**. WS routing (`RUNTIME_OWNED_VERBS`) + the WS `_handle_event_inner` actor block are **UNTOUCHED** — `websocket.py` has **no diff**.

## Tests / verification

- New direct-runtime suite `python/djust/tests/test_transport_actor_event_1901.py` (**12 cases**): `dispatch_event` routes to `dispatch_actor_event` (actor `.event()` called + framed result sent via `_send_update` with the consumer-owned wire version, in-process handler NOT run); `uses_actors` False for SSE + a WS consumer without `actor_handle`; a `view_id`-routed event skips the actor while a `view_id`-equals-top event still routes to it; SSE `dispatch_actor_event` raises.
- **Gate-off (#1468)**: forcing `uses_actors` to always return `False` (at the source) makes the 3 routing-dependent cases go **RED** (the event falls to the in-process render path). Verified.
- The existing **#1896** actor-parity test (`test_ws_event_flip_parity_1896.py`, the bespoke WS path) + the **#1899** `event_context` suite stay green.
- **Full suite**: `8249 passed, 18 skipped, 0 failed` (`pytest -n auto`).

## Files changed

- `python/djust/runtime.py` — Protocol + WS/SSE adapters + `_dispatch_event_inner` wiring + `_event_routes_to_sticky_child`.
- `python/djust/tests/test_transport_actor_event_1901.py` — new (12 cases).
- `CHANGELOG.md` — `### Changed`.

Closes #1901


johnrtipton:

Inline review proportionate to risk: the hook is DORMANT (uses_actors False for both live transports today; WS events not routed until 2.3b) and websocket.py is untouched (diff empty), so zero live-behavior change. The live validation is 2.3b's actor-event WebsocketCommunicator test + adversarial review.
- **Guard order matches WS**: `uses_actors(view) and not _event_routes_to_sticky_child(data)`, wired BEFORE event_context (actor holds no lock, matching WS). component_id still actor-routed (#1467 — WS parity); `_event_routes_to_sticky_child` PEEKS view_id without popping so non-actor sticky routing still pops it. Correct nuance.
- **dispatch_actor_event runs the WS block verbatim** (tt record/push in finally, validate, actor_handle.event, patch/HTML framing, error handling, deferred flush); SSE NotImplementedError. The event_ref/cache_request_id keyword extras are parsed like WS handle_event so framed actor results carry identical wire metadata (no spurious ref added — parity).
- Gate-off ×3 RED (force uses_actors False → actor-routing tests fall to in-process render + fail). 8249 passed.

Subtle-for-2.3b (noted, not blocking): actor result `version` ignored for the wire (consumer._next_version stamps — monotonic across mount/event); patches may be a JSON string (parsed). Both are WS-parity-preserving.

Verdict: **APPROVE**. Closes #1901; ADR-022 2.3a.

johnrtipton:
## Retrospective — PR #1902 (v1.1.0-3 Phase 2.3a / #1901)

**Quality: 5/5** (the load-bearing fold, handled with the right nuance).
### What went well
- The actor-vs-embedded-child mutual-exclusion (the axis-misalignment surprise) was matched exactly: uses_actors AND not routes_to_sticky_child, with view_id PEEKED (not popped) so the non-actor sticky path still pops it; component_id stays actor-routed for WS parity (#1467). This is the kind of guard-order subtlety a naive "add an actor branch" would have gotten wrong.
- Ran the WS actor block verbatim (incl tt-in-finally + deferred flush) rather than re-deriving it — and widened the signature with event_ref/cache_request_id only because the verbatim block needs them, flagged why.
- Dormant-until-2.3b kept it zero-live-risk; the live validation is correctly deferred to 2.3b's actor-event WebsocketCommunicator test.
### Verified
- CI green (py3.12+py3.14t+demo/js/security/nav-hooks/rust); websocket.py untouched; gate-off ×3 RED; 8249 passed; two-commit shape.
`RETRO_COMPLETE`

Issue #560, opened 2026-03-17
Tick auto-refresh causes VDOM version mismatch, silently drops user events


johnrtipton:
## Bug

When `tick_interval` is set (e.g., 60000ms), the tick handler's re-render can race with user-triggered events (`dj-change`, `dj-click`, etc.), causing a VDOM version mismatch on the client. The user's event appears to fire but the page doesn't update.

### Console output

```
[LiveView] Received: patch {type: 'patch', patches: Array(0), version: 60, …}
[LiveView] VDOM version mismatch!
  Expected v59, got v60
```

### Race condition sequence

1. Page loads → server VDOM at **v59**, client synced at v59
2. Tick fires (server-initiated) → server renders **v60**, sends empty patch `Array(0)` to client
3. Client receives v60 but expected v59 → **logs "VDOM version mismatch!"** and discards
4. User changes a `<select dj-change="filter_by_bedrooms">` → server receives event, renders **v61** (diff against v60 baseline)
5. Client still on v59 → patches from v60→v61 diff don't match client DOM → **page doesn't update**

The user sees the dropdown change visually (native browser behavior) but the page content never re-renders. No error is thrown — the update is silently dropped.

### Reproduction

```python
class MyView(LiveView):
    tick_interval = 60000  # any non-None value

    @event_handler
    def handle_tick(self, **kwargs):
        self._load_page_data()  # re-renders even if nothing changed

    @event_handler
    def filter_by_bedrooms(self, value="", **kwargs):
        self.filter = value
        self._load_page_data()
```

```html
<select dj-change="filter_by_bedrooms">
    <option value="4">4 BR</option>
    <option value="6">6 BR</option>
</select>
```

Steps:
1. Load the page, wait for WebSocket mount (v1)
2. Wait for tick to fire (~60s) — server sends v2 with empty patches
3. Immediately change the dropdown
4. Page content does not update despite the event firing

### Root cause

The tick sends patches (even empty ones) that increment the server-side VDOM version. The client either:
- Discards the tick's patch due to version mismatch (if it arrived out of order)
- Or accepts it but then the next user event's patches are diffed against a baseline the client doesn't have

Either way, the client and server VDOM versions diverge, and all subsequent patches fail silently until the page is hard-refreshed.

### Expected behavior

- Tick-triggered re-renders that produce no changes should NOT increment the VDOM version
- OR the client should be able to recover from version gaps (request re-sync)
- OR tick patches should be queued/sequenced with user event patches to prevent interleaving

[...]

PR #1910, opened 2026-06-24
fix(sse): forward 'ref' through the /event/ alias so #560 ref-echo works on both SSE endpoints (#1891)

### LINKED ISSUE #1891: tech-debt: SSE /event/ alias doesn't forward 'ref' to dispatch_event (ref echo works only via /message/)
Source: v1.1.0-3 Iter 2 Phase 2.0 (PR #1890). The runtime now echoes the event 'ref' (#560), but the SSE `/event/` alias rebuilds the dispatch dict as `{type,event,params}` (sse.py:608) and drops `ref`, so the end-to-end ref echo is exercised only via `/message/`. One-line fix: forward `ref` in the `/event/` rebuild. Out of Phase 2.0's 'runtime.py + tests only' scope; trivial standalone or fold into a later SSE touch.

johnrtipton:
## Summary

The SSE `/event/` alias dropped the client-sent `ref`, so the #560 ref echo (added in ADR-022 Iter 2 Phase 2.0) worked end-to-end only via the SSE `/message/` endpoint — never via `/event/`.

- `/message/` forwards the raw POST body verbatim to `runtime.dispatch_message`, so a client-supplied top-level `ref` reaches `dispatch_event` and is echoed on the noop / update frame.
- `/event/` (`DjustSSEEventView.post`) instead **rebuilt** the dispatch dict as `{type, event, params}` (`sse.py:~608`) and **dropped** `ref`. The runtime's `_dispatch_event_render` reads `ref` from the **top level** of the data dict (`runtime.py:1489`), so it saw `None` and echoed nothing.

## The fix

`DjustSSEEventView.post` now carries `ref` through into the dispatch frame: `{type, event, params, ref}`, mirroring how `/message/` passes the body through. The runtime coerces `ref` to int / None, so no endpoint-side validation is needed.

**Dropped-field audit**: `_cacheRequestId`, `component_id`, and `view_id` all live inside `params` (already forwarded), and SSE has neither LiveComponent nor sticky-child routing — so `ref` was the only top-level field the rebuild genuinely dropped that the runtime event path consumes.

## Tests

| Test | File | What |
| --- | --- | --- |
| `TestSSEEventAliasRefEcho` (3 cases) | `python/djust/tests/test_sse_runtime_convergence_1887.py` | real-SSE end-to-end via the `/event/` alias: update frame echoes ref, noop frame echoes ref, plus a gate-off witness |
| `TestDjustSSEEventViewPost::test_forwards_ref_to_dispatch_event` | `python/tests/test_sse.py` | dispatch-dict pin: a body with `ref` forwards it; the existing pin updated to expect `ref: None` when absent |

**Gate-off (#1468)** verified: reverting the rebuild to the pre-fix `{type, event, params}` shape makes the two echo tests RED (the update + noop frames carry no ref), while the gate-off witness — which re-drops `ref` at the endpoint to confirm absence — stays green. Restoring the fix → all green. `/message/` ref-echo tests (`TestSSEEventSpineParity`) stay green throughout.

## Verification

- Full Python suite: **8285 passed, 18 skipped, 0 failed**.
- SSE slice (`test_sse_runtime_convergence_1887`, `test_sse`, `test_sse_csrf_origin`, `test_sse_session_binding_f24_f25`, `test_object_perm_api_sse_paths`): all green.
- `ruff check` + `ruff format --check` clean; changelog test-count check passes.

Closes #1891


johnrtipton:
## Retrospective — PR #1910 (#1891)
**Quality: 5/5** (clean one-liner; dropped-field audit confirmed only ref). Real-SSE echo tests + gate-off; CI green; 8285 passed. `RETRO_COMPLETE`

PR #1926, opened 2026-06-24
ci: promote the browser-smoke canary to a hard merge gate (#1869) + fix the inline-<script> render regression it caught (#1927) + flip the #1848 xfail

### LINKED ISSUE #1869: tech-debt: promote playwright browser-smoke to a hard merge gate once runner-green (#1534)
Source: v1.0.8-1 retro (PR #1866 / T2-D #1855).

PR #1866 added a Playwright browser-smoke (`tests/playwright/test_browser_smoke.py`) for the runtime-break class a green pytest suite cannot see (#1849): a hard mount canary + a #1848-shape inline-script-in-dj-root xfail that reproduces the open regression live.

Per #1534 (a new CI gate exercising an environment the dev machine can't fully mirror ships `continue-on-error: true` until it has been green on a GitHub runner at least once), the smoke is currently **non-gating**. 

**Action**: once the browser-smoke leg has gone green on a GitHub runner, (a) remove the `continue-on-error` marker, and (b) add it to the `test-summary` aggregate AND-condition so a failure blocks the merge (per #1713 — a check must be in the AND-condition's success `if`, not merely echoed/`needs`, to actually gate). Also flip the #1848 xfail to a hard assertion once the framework fix for #1848 lands.

### LINKED ISSUE #1927: bug: inline <script> in dj-root still not executed on the LIVE morphChildren prerendered-mount path (#1848 reopen)
Surfaced by the now-blocking browser-smoke canary (#1869 / PR #1926) — deterministic RED on 2 runs. PR #1871 added `window.djust._runInsertedScripts` to re-execute classic inline `<script>` on the #1610 mount morph and wired it into both mount branches, **but its 6 JSDOM tests verified the helper against an `innerHTML`-built DOM, never the live `morphChildren` path** (the #1849 'browser-test the real path' + v1.0.0rc4 'enumerate every variant' class). On `/demos/browser-smoke/` the inline `<script>` inside dj-root is NOT executed under the live prerendered-mount morph: the browser-smoke MOUNT canary (class 1) passes, but the #1848 INLINE-SCRIPT branch fails with `inline-script-ran=False`.

**Diagnosis (from #1869's implementer)**: likely localized to the `skipMountHtml` morph branch (`python/djust/static/djust/src/03-websocket.js:480-508`) / the WS-mount-flip mount path (`runtime.py:2381-2403`) — `_runInsertedScripts` isn't invoked on that specific live branch. Fix: invoke the inserted-script re-execution on the `morphChildren`/skipMountHtml mount branch too; verify against the browser-smoke canary (the only test that exercises the real morph path). Unblocks #1869 (PR #1926).

johnrtipton:
## Summary

Two coupled changes that land atomically so the new gate self-validates its own dependency (main never carries a red gate):

1. **`ci:` Promote the Playwright browser-smoke canary to a HARD merge gate (#1869, Action Tracker #314).**
2. **`fix:` Fix the inline-`<script>`-in-dj-root render regression that the now-blocking canary immediately caught (#1927).**

Closes #1869. Closes #1927.

## Part 1 — browser-smoke hard gate (#1869)

Carve `tests/playwright/test_browser_smoke.py` out of the non-blocking `playwright-tests` leg into its OWN dedicated `browser-smoke` CI job (no `continue-on-error`), wired into the `test-summary` aggregate gate's AND-condition (`needs.browser-smoke.result == "success"`), so a re-introduced 1.0.7-class runtime break red-bars the PR. Mirrors the `demo-checks` blocking-job pattern (#1708/#1713). Promoted per #1534 only after the canary shipped green on the runner across multiple PRs. The other 4 playwright tests (loading_attribute / cache_decorator / draft_mode / nav_hooks) stay in the still-`continue-on-error` `playwright-tests` leg — the full suite can be flaky; only the stable two-class canary gates.

The #1848 inline-script branch of the canary, previously a tolerated known-xfail, is flipped to a HARD assertion now that #1848 was code-fixed by PR #1871.

## Part 2 — the regression the gate caught (#1927)

On its first run, the now-blocking `browser-smoke` job went RED on the #1848 inline-script class (`inline-script-ran=False`) — deterministically, twice. **The gate worked: it caught a real runtime break the green unit suite is blind to.** Per-coordinator decision, the fix is folded in here so the gate and fix land together.

**Root cause** (diagnosed by driving `/demos/browser-smoke/` in a REAL browser): `TemplateMixin._strip_comments_and_whitespace` — the single normalizer every render path runs (HTTP GET, WS mount, SSE/runtime, streaming) to match the Rust VDOM parser's whitespace pass — preserved whitespace only for `<pre>`/`<code>`/`<textarea>`, but the Rust parser ALSO preserves `<script>`/`<style>` (`crates/djust_vdom/src/parser.rs:475`). Its `re.sub(r"\s+", " ")` pass collapsed every newline inside an inline `<script>` onto ONE line; a leading `//` line comment then commented out the whole body, so the script's `addEventListener` never ran — with NO console error. This is why #1871's `_runInsertedScripts` mount-morph re-execution could not cure the symptom: the script was already neutered at render, before any morph re-execution.

**Fix**: add `<script>`/`<style>` to the preserved-block set (the #1646 parallel-path-drift cure — the Python normalizer now matches the Rust parser's preserve set exactly), and — critical ordering — extract the raw-text `<script>`/`<style>` blocks BEFORE the HTML-comment strip so an HTML-comment-looking token inside a JS/CSS body (`var s = '<!-- x -->'`) is not stripped as markup. Non-script/style whitespace collapse is unchanged. One shared helper → all render paths fixed at once.

**Validation** (in a real browser, the path the canary exercises):
- HTTP-GET served script: 0 newlines (broken) → 24 newlines (fixed).
- Live page: `__smokeTabsWired` `false` → `true`; tab toggle moves `.active` Tab1→Tab2, panel 2 shows, `__smokeTabsHandled=1`; mount canary `count` 0→1; **no console errors**.
- New `TestStripCommentsAndWhitespace` cases (the #1927 `//`-comment trigger, multi-script/style, comment-inside-script ordering guard, non-regression collapse); **gate-off (#1468) verified** — reverting the preservation collapses the body and reds the assertion.

## Verification

- `yaml.safe_load` parses `test.yml`; `browser-smoke` has no `continue-on-error`; it is in `test-summary.needs` AND the AND-condition; full #1713 gate chain is enforcing.
- Non-playwright full suite: **8585 passed, 0 failed** (with the pre-existing-flaky `test_theme_tags_rust_engine_1721.py` deselected — see note).
- The **PR's own now-blocking `browser-smoke` CI job is the end-to-end validator** for the #1927 fix; it must go GREEN (both the MOUNT canary and the #1848 inline-script hard assertion) for this PR to merge.

### Note: a pre-existing unrelated flake
`python/djust/tests/test_theme_tags_rust_engine_1721.py` flakes under the full `-n auto` run (17 tests: `has_tag_handler(...)` is False when another test calls `django.setup()` with settings lacking `djust.theming`, so `DjustThemingConfig.ready()` never registers the handlers). **Confirmed pre-existing** — it fails identically on this branch WITH my changes stashed (baseline 8577 passed / 17 failed), and passes 3/3 in isolation. Orthogonal to #1869/#1927; not fixed here per scope discipline (#1079). Worth a follow-up issue (settings-isolation for the theming app config).

## #1236 governance note

This `.github/workflows/test.yml` change is an intentional merge-gate addition per #1869; it will trip the #1236 governance gate (CI-workflow changes need review) and require admin-merge.


johnrtipton:
## ⚠️ FLAG — the now-blocking `browser-smoke` job is RED (deterministic), and it caught a real #1848 regression

The gate wiring this PR adds is working **exactly as designed**, and it immediately earned its keep: it caught a runtime break the green unit suite is blind to.

### Gate wiring is correct (empirically proven)
- `browser-smoke` → **fail** ⇒ `test-summary` → **fail** (the merge is red-barred per #1713).
- `playwright-tests` → **pass** and stays non-blocking (informational), as intended.

### What the canary caught
The **MOUNT canary (class 1) PASSES** — the LiveView mounts and `dj-click="bump"` round-trips 0→1. Only the **#1848 inline-script branch FAILS**, deterministically across **two consecutive runs** (not a flake):

```
#1848-inline-script: tab toggle did not fire — active tab stayed 'smoke-tab-1',
handler invocations=0, inline-script-ran=False.
```

`inline-script-ran=False` means the inline `<script>` inside the dj-root on `/demos/browser-smoke/` is **not executed under the live WS-mount morph**, so its delegated listener never registers — i.e. the **#1848 bug is back on the live prerendered-mount path**, even though PR #1871's fix code (`window.djust._runInsertedScripts`) is present and served (verified in both `src/03-websocket.js` and the committed `client.js`, wired into both mount branches).

### Why the unit suite missed it (the variant-coverage gap)
PR #1871's 6 JSDOM tests verified `_runInsertedScripts` against an `innerHTML`-built DOM and a direct helper call — but **never against the live `morphChildren` prerendered-mount path** the demo actually takes. This is the v1.0.0rc4 retro finding #1 class ("a coverage suite must enumerate EVERY variant of the surface it covers") and the #1849 "browser-test the real path" rule. The likely cause is in the `skipMountHtml` morph branch (`03-websocket.js:480-508`): `_runInsertedScripts(_morphContainer)` runs only when `has_ids && data.html`, AND its effect after `morphChildren` clones the `<script>` into the live container may differ from the JSDOM `innerHTML` path. Probably surfaced/adjacent to the WS-mount flip (#1919/#1920, `runtime.py:2381-2403` `skip_html_for_resume`). The mount-canary passing while only the inline-script class fails localizes it precisely to inline-script-execution-under-morph.

### This is a separate framework bug, out of scope for #1869
#1869 is the **gate-promotion** PR (`ci:` + the xfail flip). The underlying regression needs its own **`fix(client):`** PR + a follow-up issue (a #1848 reopen / new bug). This gate PR should ride green on top **after** that framework fix lands.

Recommended sequence:
1. File a follow-up issue: "#1848 inline-`<script>`-in-dj-root not executed on the **live `morphChildren` prerendered-mount** path (PR #1871's JSDOM tests never exercised this variant)".
2. Land the framework fix (re-verified against this very canary — the canary is now the regression guard).
3. Re-run CI here; `browser-smoke` goes green; merge this gate PR (admin-merge per the #1236 governance note).

The hard assertion / hard gate are correct as written. I have NOT softened them to make CI green — that would defeat the purpose of #1869.

johnrtipton:
## Retrospective — PR #1926 (#1869 + #1927)
**Quality: 5/5** (the gate justified itself on its first blocking run). Promoting the browser-smoke to blocking immediately caught a real render-time regression (#1848 reopen) the 8585-passing pytest suite + HTTP smoke could not see — exactly the #1849 'browser-test the real path' class. Root cause was a #1646 normalizer-vs-parser preserve-set drift (Python collapsed inline-<script> newlines; Rust preserves script/style) → fixed in the one shared normalizer, all render paths at once. Implementer correctly refused to soften the gate to go green, diagnosed via a real browser, and the fix is validated end-to-end by the now-blocking canary. CI green (real codeql + test-summary + browser-smoke + all legs); the 3s phantom 'CodeQL' check is a stale required-check artifact. `RETRO_COMPLETE`

Issue #1883, opened 2026-06-23
tech-debt: systemic test-isolation — autouse fixture to reset djust process-globals across the xdist worker (v1.1.0-1 test hygiene)


johnrtipton:
Source: v1.1.0-1 drain (PRs #1874/#1881 retros).

Three shared-process-global test-pollution flakes in two milestones, all the SAME class — a process-global left dirty across tests in an xdist worker:
- #1862 — `ROOT_URLCONF` leak (fixed PR #1874)
- #1875 — `djust_hotreload` channel-layer group pollution (fixed PR #1881)
- #1882 — process-global wire-version counter pollution (open)

Each was whack-a-moled individually. The systemic cure is one **autouse fixture** (in `conftest.py`) that resets djust's process-globals between tests: the channel layer (`channel_layers.backends.clear()`), the URLconf caches (`clear_url_caches()` + `set_urlconf(None)`), the wire-version / render counters, the sticky-child + view registries, and any other module-global registry. That retires the entire flaky-class instead of fixing one test at a time, and prevents the next instance.

**Action**: inventory djust's process-global mutable state (grep module-level singletons / `get_channel_layer()` / registries / counters), add an autouse `_reset_djust_globals` fixture, verify the suite is order-independent (run `-n auto` repeatedly + with `-p randomly`). Fits the v1.1.0-1 'test/CI hygiene' sub-milestone (ADR-022). Likely lets #1882 + future instances be closed by the fixture rather than per-test patches.

Issue #1771, opened 2026-06-13
tech-debt: test_tag_registry.py clears the global Rust tag registry without restoring theme handlers (serial-run pollution)


johnrtipton:
Discovered while validating PR #1770 (pyo3 0.29 bump) with a **serial** full-suite run.

## Symptom
A serial `pytest tests/ python/tests/ python/djust/tests/` run produces 17 failures in `python/djust/tests/test_theme_tags_rust_engine_1721.py`:
`RuntimeError: Template error: Unsupported template tag '{% theme_panel %}'`.

The tests pass in isolation and pass under CI (which shards with `pytest-xdist`, isolating the polluter into a separate process).

## Root cause
`tests/unit/test_tag_registry.py`'s module-scoped autouse fixture `_isolate_tag_registries` calls `djust._rust.clear_tag_handlers()` (the **global** Rust tag-handler registry) and — per its own docstring — only re-registers the *built-in* Python handlers. The theme tag handlers (`theme_head`, `theme_panel`, `theme_switcher`, `theme_mode_toggle`, `theme_preset_selector`) are registered **once** at `DjustThemingConfig.ready()` (django.setup), so after `test_tag_registry.py` runs they are gone for the rest of a serial session, and any later test that renders a theme tag through the Rust engine fails.

## Proven pre-existing / pyo3-independent
Rebuilt the extension with **pyo3 0.25** (pre-#1770) and the minimal repro `pytest tests/unit/test_tag_registry.py python/djust/tests/test_theme_tags_rust_engine_1721.py` produced the identical `17 failed, 49 passed`. So this is not related to the pyo3 bump.

## Proposed fix
Make `_isolate_tag_registries` save and restore the **full** prior registry (snapshot all handlers before `clear_tag_handlers()`, restore them after `yield`) instead of only re-registering built-ins — OR have the theming app's handlers re-register via a session/module fixture the theme-tag tests depend on. Either restores cross-file isolation for serial runs. Repro to gate the fix: the two-file `pytest` invocation above must go green.

Minimal repro:
```
PYTHONPATH=. .venv/bin/python -m pytest tests/unit/test_tag_registry.py python/djust/tests/test_theme_tags_rust_engine_1721.py
```

PR #1929, opened 2026-06-24
fix(tests): de-flake theme-tags handler-registration under -n auto (#1928, #1883-class settings-isolation)

### LINKED ISSUE #1928: flaky: test_theme_tags_rust_engine_1721 (17 tests) under -n auto — DjustThemingConfig.ready() handler registration leaks across django.setup() (settings-isolation, #1883-class)
Surfaced (confirmed PRE-EXISTING — reproduces on baseline, passes 3/3 in isolation) during PR #1926. Under full `-n auto`, `has_tag_handler()` is False because another test's `django.setup()` configured settings WITHOUT `djust.theming`, so `DjustThemingConfig.ready()` never registered the theme tag handlers → all 17 tests fail. Same systemic shared-process-global-across-the-xdist-worker class as #1883 (the autouse test-isolation fixture). Either extend the #1883 `reset_djust_globals` fixture to re-assert theming handler registration, or make `DjustThemingConfig.ready()` idempotent + handler-registration resilient to settings re-config. Orthogonal to the convergence; left unfixed per #1079 scope discipline.

johnrtipton:
## Summary

Fixes the pre-existing flake #1928: the 17 `#1721` theme-tag tests
(`python/djust/tests/test_theme_tags_rust_engine_1721.py`) fail under full
`-n auto` because `has_tag_handler("theme_panel")` returns `False` and rendering
500s with `Unsupported template tag '{% theme_panel %}'`. Pass 3/3 in isolation;
confirmed pre-existing (reproduces on the `main` baseline).

## Root cause (symptom-up)

Same shared-process-global class as #1883.

- **Leaked state**: the process-global Rust tag-handler registry
  (`crates/djust_templates/src/registry.rs`), shared across an xdist worker.
- **Polluter**: `tests/benchmarks/test_tag_registry.py::TestRustPythonInterop`
  calls `clear_tag_handlers()` (wiping the whole registry) and its
  `restore_registry` fixture restores ONLY the `djust.template_tags` built-ins —
  NOT the theme/component handlers registered at
  `DjustThemingConfig.ready()` / `DjustComponentsConfig.ready()`.
  `ready()` runs only ONCE per process, so once cleared the theme handlers stay
  gone for every later test in the worker. (Also reproducible by any test that
  `django.setup()`s without `djust.theming`.)
- **Why #1771 missed it**: #1771 added the theme/component re-register only to
  `tests/unit/test_tag_registry.py`'s module-scoped fixture; the benchmark twin
  with the same `clear_tag_handlers()` was uncovered — classic parallel-path
  drift (#1646).

## Reproduce-first (deterministic)

Running the benchmark polluter then the theme tests, same process:
`test_register_handler` (→ `clear_tag_handlers()`) → 17 theme tests
**failed 17 / 18 pre-fix**, **pass 18 / 18 post-fix**.

## Fix — approach (b), the #1883 systemic cure

Extend `reset_djust_globals()` (`python/djust/test_isolation.py`) with
`_reset_rust_tag_handlers()`, which re-runs both `ready()`-time registrars
(`djust.theming.rust_handlers.register_with_rust_engine` +
`djust.components.rust_handlers.register_with_rust_engine`) BEFORE every test in
both test roots (the existing autouse `_reset_djust_globals` fixture). It is:

- **idempotent** — the theming registrar guards on `has_tag_handler`; the
  component one overwrites;
- **optional-dep safe** — no-op when the Rust extension / theming / components
  are unavailable (wrapped like the sibling reset helpers); and
- **cheap** — when handlers are present the theming guard skips the PyO3
  round-trips.

Chosen over a point-fix to the benchmark file because the leak is
test-pollution and the registry is a process-global: one autouse fixture retires
the entire flaky class regardless of which polluter ran (the #1883 preference).

## Tests

`python/djust/tests/test_global_isolation_1883.py`:

- `test_reset_reasserts_theme_and_component_tag_handlers_1928` — models the
  polluter: clear → prove gone → reset → prove restored.
- `test_gate_off_clear_without_reset_loses_theme_handler_1928` — gate-off sibling
  (#1468) proving the bare clear genuinely loses the handler (non-tautological).

**Gate-off (#1468)**: neutering `_reset_rust_tag_handlers()` re-reds BOTH the
repro order (17 fail) AND the new pin test — fix is load-bearing.

## Verification — 3-clean-runs gate (#1174)

Full suite `pytest tests/ python/tests/ python/djust/tests/ -n auto` × 3:

| run | result |
|-----|--------|
| 1/3 | 8604 passed, 21 skipped, 0 failed |
| 2/3 | 8604 passed, 21 skipped, 0 failed |
| 3/3 | 8604 passed, 21 skipped, 0 failed |

Two-commit shape (#1173): commit 1 code+tests, commit 2 CHANGELOG.

Closes #1928.


johnrtipton:

Inline review proportionate to risk (test-isolation fixture extension; 3-clean-runs ×3 green).
- **Root cause is another #1646 twin**: the #1771 registry-clear bug was fixed only in `tests/unit/test_tag_registry.py`; the `tests/benchmarks/test_tag_registry.py` twin's `restore_registry` fixture restored only the `djust.template_tags` built-ins, not the theme/component handlers — so once cleared, the once-per-process `ready()` registrars never re-ran → 17 theme-tag tests 500.
- **Fix is the #1883 systemic cure (approach b)**: extended `reset_djust_globals()` with `_reset_rust_tag_handlers()` (re-asserts theming + component handler registration before every test, idempotent + optional-dep-safe) — retires the whole class regardless of polluter, vs a point-fix to the one benchmark file. Approach (a) correctly rejected (ready() is once-per-process by Django design; the registry is fine in production — this is purely test-isolation).
- Reproduce-first (17 fail → 18 pass), gate-off RED (neuter → repro + the new pin RED), 3-clean-runs ×3 = 8604 passed each.

Verdict: **APPROVE**. Closes #1928 — extends the v1.1.0-1 test-hygiene drain (#1883/#1884) to the Rust tag-handler-registry pollution class.

johnrtipton:
## Retrospective — PR #1929 (#1928)
**Quality: 5/5** (another #1646 twin found+cured systemically). The #1771 registry-clear fix had a benchmark twin (tests/benchmarks/test_tag_registry.py) that the unit-test fix missed — parallel-path drift in TEST fixtures. Cured via the #1883 systemic reset_djust_globals fixture (approach b) rather than a point-fix, retiring the Rust-tag-handler-registry pollution class regardless of polluter. 3-clean-runs ×3 (8604 each) + reproduce-first + gate-off. `RETRO_COMPLETE`

Issue #1930, opened 2026-06-24
flaky: TestGlobalRateLimit::test_ping_flood_triggers_disconnect under parallel make test (#1883-class isolation)


johnrtipton:
Surfaced during the v1.1.0rc1 release pre-flight (`make test`). `tests/unit/test_event_security.py::TestGlobalRateLimit::test_ping_flood_triggers_disconnect` fails under the full parallel `make test` run but passes 3/3 in isolation + the whole `TestGlobalRateLimit` class passes isolated. Test file untouched since 2026-06-19 (pre-v1.1) → NOT a code regression; the ping-flood disconnect feature works. Pre-existing test-isolation pollution (a connection/global rate-limit or settings process-global leaking under certain orderings), same #1883-class as #1928. The worktree CI-way `-n auto` runs all passed 8642 (order-dependent). Fix: identify the leaked global (likely in `rate_limit.py` — `_handler_buckets` LRU, or a settings/connection-limiter state) and extend `reset_djust_globals()` (`python/djust/test_isolation.py`) to reset it, mirroring the #1928 tag-handler reset. Does NOT block the v1.1.0rc1 release (make release + release.yml verify don't run the suite; release CI `python-tests` runs `-n auto` which passes).

johnrtipton:
**Corrected root cause (not #1883-class pollution).** The test builds a *local* `ConnectionRateLimiter` — no module-global leaks. The real cause is **wall-clock timing fragility**: `TokenBucket` (`python/djust/rate_limit.py:30-47`) reads `time.monotonic()` directly and refills `tokens += elapsed * rate`. With `rate=100` (1 token/10ms), under CPU-saturated parallel `make test` enough real time elapses between `check()` calls that a token refills, so the 'burst exhausted → False' assertion non-deterministically fails. `test_token_bucket_refills` even uses a real `time.sleep(0.05)`. This is the wall-clock-gate flake class the CLAUDE.md canon warns against (v1.0.5-4/-5 concurrency rule, v1.0.6rc1 median rule, #1830 rAF rule). Fix: own the clock — inject a controllable time source so burst tests freeze it (no refill → deterministic) and the refill test advances it explicitly (no real sleep).

PR #1932, opened 2026-06-24
fix(tests): own the clock in rate-limit tests — kill the wall-clock burst-refill flake (#1930)

johnrtipton:
## Summary

Closes #1930. De-flakes the six rate-limit burst-exhaustion tests in `tests/unit/test_event_security.py` (`TestRateLimiter` + `TestGlobalRateLimit`) by **owning the clock**.

### Diagnosed root cause (wall-clock, NOT #1883-class pollution)

The issue hypothesized a `#1883`-class shared-global leak. The actual root cause is wall-clock dependency in `TokenBucket`:

- `TokenBucket` (`python/djust/rate_limit.py`) read `time.monotonic()` directly in `__init__` and `consume()`, refilling `tokens = min(burst, tokens + elapsed * rate)`.
- These tests build **test-local** limiters (`ConnectionRateLimiter(rate=100, burst=2)` etc.) and assert burst exhaustion (after N checks, the next `check()` must be `False`).
- Under CPU-saturated parallel `make test`, real wall-clock elapses between `check()` calls → a token refills (`rate=100` = 1 token / 10ms) → the "burst exhausted → `False`" assertion non-deterministically returns `True` → FAIL.

This is the CLAUDE.md flaky-timing class ("never gate a pass/fail on wall-clock"; v1.0.5-4/-5, v1.0.6rc1, #1830/#1815). It is **not** `#1883`-class pollution — every limiter here is test-local and reads no leaked global. (Per "trust the symptom, not the cited path" — the symptom, passes-isolated/fails-parallel, fit both hypotheses; tracing showed `time.monotonic()` is the only process-global the burst path reads.)

### Fix — own the clock

1. **`_monotonic` seam** in `rate_limit.py`: module-level `_monotonic = time.monotonic`; `TokenBucket.__init__` + `consume()` route through `_monotonic()`. Production behavior is identical (one extra indirection); it gives tests a patch point **without** patching the global `time` module. `IPConnectionTracker` is untouched (out of scope, not part of the flake).
2. **`FakeClock` + `frozen_clock` fixture** in the test file: `monkeypatch.setattr("djust.rate_limit._monotonic", clock)`.
   - **FROZEN clock** (created, never advanced → `elapsed == 0` → no refill) on the five burst-exhaustion tests: `test_token_bucket_allows_burst`, `test_connection_rate_limiter_global`, `test_per_handler_rate_limit`, `test_non_event_messages_are_rate_limited`, `test_ping_flood_triggers_disconnect`.
[...]

PR #1933, opened 2026-06-24
fix(tests): make async-work mount-dispatch test await completion deterministically — kill scheduler-starvation flake (#1931)

### LINKED ISSUE #1931: flaky: TestMountAsyncAndPushDrain::test_mount_dispatches_async_work under parallel make test (bounded-poll vs scheduler race)
Surfaced while fixing #1930 (rate-limit wall-clock flake). `python/djust/tests/test_transport_behavioral_parity.py::TestMountAsyncAndPushDrain::test_mount_dispatches_async_work` fails non-deterministically under the full parallel `-n auto` suite (1/4 runs in the #1930 worktree) but passes 3/3 in isolation.

**Root cause (separate from #1930):** the test schedules background work via `start_async()` in `mount()`, then polls a BOUNDED number of event-loop turns for the callback to land:

```python
for _ in range(10):
    if getattr(view, 'value', 0) == 42:
        break
    await asyncio.sleep(0)
assert view.value == 42
```

The callback runs in its own `ensure_future`. Under CPU-saturated `-n auto`, 10 `asyncio.sleep(0)` yields can elapse before that future is scheduled onto a busy loop, so `value` is still 0 at the assertion. This is the CLAUDE.md bounded-poll-racing-a-real-scheduler flake class (#1830 / #1815 family) — NOT a code regression; the mount async-dispatch feature works (it lands every time in isolation).

**Suggested fix:** replace the fixed 10-turn poll with an awaitable completion signal — e.g. have the test view's async callback set an `asyncio.Event` and `await asyncio.wait_for(done.wait(), timeout=N)`, or drain the runtime's `_async_tasks` deterministically (`await asyncio.gather(*pending)`) instead of yielding a fixed count. Own the completion, don't race the scheduler.

Not blocking: deselected in the #1930 PR's 3-clean-runs gate; CI `-n auto` may still hit it intermittently.

johnrtipton:
## Summary

Deflakes `tests/.../test_transport_behavioral_parity.py::TestMountAsyncAndPushDrain::test_mount_dispatches_async_work` — a bounded-poll-vs-scheduler timing race surfaced during the #1930 work. Closes #1931.

Test-only change; **production code unchanged**.

## The diagnosed race (poll-vs-scheduler)

The test mounts `_MountPushAsyncView`, which schedules a background callback via `start_async()` in `mount()`, then asserts the callback ran (`view.value == 42`).

- The runtime dispatches that callback **fire-and-forget** via `asyncio.ensure_future(self._execute_async_task(...))` inside `ViewRuntime._dispatch_async_work` (`python/djust/runtime.py:4444`).
- `_execute_async_task` **itself** awaits a `sync_to_async(callback)` thread-pool round-trip before setting `value=42`.
- The original test waited via a **bounded poll**:
  ```python
  for _ in range(10):
      if getattr(view, "value", 0) == 42: break
      await asyncio.sleep(0)
  assert view.value == 42
  ```

Under a CPU-saturated parallel `-n auto` loop the asyncio scheduler can fail to run the spawned task (which also competes for the thread pool) within 10 yields, so the assertion fires while `value` is still 0. Flaked 1/4 runs in the #1930 worktree; passed 3/3 in isolation. This is the CLAUDE.md **bounded-poll-racing-a-real-scheduler** flake class (#1830/#1815 family) — NOT a code regression. Mechanism is **scheduler starvation**, distinct from #1930's clock-refill (no `rate_limit`/`TokenBucket` involvement).

## The deterministic-wait fix

Own the completion signal instead of racing it. During `dispatch_mount`, wrap the runtime's `asyncio.ensure_future` seam to capture the **exact `_execute_async_task` task handle** it spawns, then `await asyncio.gather(*async_work_tasks)`. Deterministic — no timing bound, immune to scheduler load.

The `_flush_push_events` fire-and-forget send (which uses the same `ensure_future` primitive) is **excluded by coroutine name** (`co_name == "_execute_async_task"`), so the gate-off below stays sharp on the async-dispatch path specifically.

## Reproduce-first evidence

A standalone repro parameterizing the poll bound (scheduler held constant):

```
poll_bound=0:  25/25 attempts had value!=42 at assertion
poll_bound=1:  25/25 attempts had value!=42 at assertion
poll_bound=2:  0/25
poll_bound=5:  0/25
poll_bound=10: 0/25
```

The current `range(10)` margin is only ~2 yields wide — exactly the thin margin a saturated `-n auto` loop (slower `sleep(0)` resumes + thread-pool contention) erases intermittently. The new form passes **0/15 failures under 8-way CPU saturation**.

## Gate-off evidence (#1468 — load-bearing, non-tautological)

Temporarily disabled `self._dispatch_async_work(None)` in `dispatch_mount`:

```
assert async_work_tasks
E   AssertionError: dispatch_mount must spawn the start_async() callback as an
    _execute_async_task via ensure_future (#1280); _dispatch_async_work(None)
    was not called
E   assert []
```

No `_execute_async_task` is spawned → the capture list is empty → the test fails before the value check. Restored → green. The test gates precisely on the async-dispatch path. (runtime.py restored from an aside copy, never `git checkout`.)

## 3-clean-runs (#1174)

Full suite `pytest tests/ python/tests/ python/djust/tests/ -n auto` three consecutive times:

| Run | Result |
|-----|--------|
| 1 | 8604 passed, 21 skipped |
| 2 | 8604 passed, 21 skipped |
| 3 | 8604 passed, 21 skipped |

ruff check + ruff format: clean.

## Files

| File | Change |
|------|--------|
| `python/djust/tests/test_transport_behavioral_parity.py` | deterministic `ensure_future`-capture + `gather` wait replacing the bounded poll |
| `CHANGELOG.md` | `### Fixed` entry |


johnrtipton:

Inline review (test-only de-flake; production unchanged).
- **Diagnosed race**: `dispatch_mount` schedules the async callback fire-and-forget via `asyncio.ensure_future(_execute_async_task(...))`, which itself awaits a `sync_to_async` thread-pool round-trip before setting the value. The test's bounded poll (`for _ in range(10): await asyncio.sleep(0)`) is only ~2 yields of margin → under `-n auto` scheduler starvation the assertion fires before the task runs. Same #1830/#1815 family as #1930 but scheduler-starvation, not clock-refill (no rate_limit involvement).
- **Deterministic fix**: capture the exact `_execute_async_task` handle via the `ensure_future` seam (excluding the `_flush_push_events` coroutine by name so gate-off stays sharp), then `await asyncio.gather(...)` — a real completion signal, no timing bound. Production async API is fire-and-forget by design (no retained handle), so the seam-capture is the right awaitable; no production change.
- Reproduce-first (poll_bound 0/1 → 25/25 fail, ≥2 → 0/25), gate-off RED (disable the async dispatch → `assert async_work_tasks` fails on `[]`), 3-clean-runs 8604 ×3.

Verdict: **APPROVE**. Closes #1931 — clears the last known timing flake; the parallel suite is now deterministic.

johnrtipton:
## Retrospective — PR #1933 (#1931)
**Quality: 5/5** (deterministic completion-await, the canonical own-the-scheduler fix). Replaced a bounded wall-clock poll with awaiting the captured task handle (asyncio.gather) — assert a completion invariant, never a timing threshold (#1830/#1815). Test-only (production fire-and-forget unchanged). Reproduce-first + gate-off (load-bearing on the real async-dispatch path) + 3-clean-runs 8604×3. Second of the two timing flakes surfaced during the v1.1.0rc1 release pre-flight; with #1930 (#1932) the parallel suite is now deterministic. `RETRO_COMPLETE`

Issue #307, opened 2026-02-14
dj-patch="/" fails to update URL; server-side live_patch() messages not routed


johnrtipton:
## Bug 1: `dj-patch="/"` fails to update browser URL

In `18-navigation.js`, `bindNavigationDirectives` has this condition:

```js
if (url.pathname !== '/' && patchValue.startsWith('/')) {
    newUrl.pathname = url.pathname;
}
```

When using `dj-patch="/"`, `url.pathname` is `"/"`, so the condition `url.pathname !== '/'` is **false**. The `newUrl.pathname` is never updated and stays as the current page path (e.g. `/brain/`). The `url_change` message is sent to the server with the correct URI so the view state updates, but the browser URL doesn't change.

**Reproduction:**
1. Set up two URL routes serving the same LiveView: `path("", view)` and `path("<str:view>/", view)`
2. Add tabs with `dj-patch="/brain/"` and `dj-patch="/"`
3. Navigate to `/brain/` (works — URL updates)
4. Click the `dj-patch="/"` tab — view content updates but URL stays at `/brain/`

**Workaround:** Use `dj-patch="/dashboard/"` instead of `dj-patch="/"` to avoid the root path.

---

## Bug 2: Server-side `live_patch()` messages silently fail

In `websocket.py:_flush_navigation()`:

```python
for cmd in commands:
    await self.send_json({"type": "navigation", **cmd})
```

The `cmd` dict from `_drain_navigation()` contains `{"type": "live_patch", "path": "/", ...}`. The `**cmd` spread **overwrites** `"type": "navigation"` with `"type": "live_patch"`.

On the client side, the message router checks `case 'navigation':` but receives `type: "live_patch"`, so it never reaches `handleNavigation()`. The URL is not updated.

**Suggested fix:**

```python
for cmd in commands:
    nav_type = cmd.pop("type", "live_patch")
    await self.send_json({"type": "navigation", "nav_type": nav_type, **cmd})
```

And update the client to read `data.nav_type` instead of `data.type` in `handleNavigation()`:

```js
function handleNavigation(data) {
    if (data.nav_type === 'live_patch') {
        handleLivePatch(data);
    } else if (data.nav_type === 'live_redirect') {
        handleLiveRedirect(data);
    }
}
```

---

**djust version:** 0.3.0

johnrtipton:
[...]

PR #1935, opened 2026-06-24
fix(client): live_redirect to a non-LiveView path falls back to full-page navigation (#1934)

### LINKED ISSUE #1934: live_redirect to a non-LiveView path strands the page (URL changes, stale LiveView stays mounted) [1.1.0rc1]
## Summary

When `live_redirect` targets a path that is **not** a LiveView (e.g. a plain Django `TemplateView`), the client `pushState`s the URL to that path but never swaps the DOM — leaving the *previous* LiveView mounted. The URL bar shows the new path; the body shows stale content.

**djust:** 1.1.0rc1 · **Transport:** WebSocket (daphne ASGI) · `auto_navigate: true` (the v1.1 default)

## Observed

Navigating to `/onboarding/` (a `TemplateView`) from `/jira/` (a LiveView). Console:

```
[LiveView] live_redirect: pushState → http://…/onboarding/
[LiveView] Received: mount {type:'mount', view:'jira_manager.views.TicketListView', version:8, …}
[LiveView] View mounted: jira_manager.views.TicketListView
```

URL = `/onboarding/`, but `TicketListView` (the page we left) is what mounts. A hard reload of `/onboarding/` renders correctly.

## Root cause (`handleLiveRedirect`, client.min.js)

History is changed **before** the LiveView check, and there is no fallback when the target isn't a LiveView:

```js
window.history[e.replace?"replaceState":"pushState"]({djust:!0,redirect:!0}, "", t.toString());
// URL now /onboarding/  ← the console line above
…
if (isWSConnected()) {
    const e = resolveViewPath(t.pathname);   // /onboarding/ not in route map → undefined
    if (e) { …stash sticky, before-navigate, send live_redirect mount… }
    // ❌ no else: URL already changed, DOM untouched, old [dj-view] still mounted
}
```

`resolveViewPath` is correctly falsy: `_walk_liveview_routes` (routing.py) filters `if not issubclass(view_cls, LiveView): continue`, so a `TemplateView` is properly excluded from the route map. The bug is that `handleLiveRedirect` doesn't honor that exclusion with a full-page fallback.

**Contrast:** `_handleAutoNavigateClick` gets this right — `resolveViewPath(n.pathname)` is part of its guard, so a plain `<a href>` to a non-LiveView falls through to a normal browser navigation. Only the `live_redirect` path lacks the equivalent fallback.

## Suggested fix

```js
const view = resolveViewPath(t.pathname);
if (view) { /* existing SPA mount */ }
else { window.location.href = t.toString(); }   // non-LiveView → real navigation
```

Ideally also defer the `pushState` until after the LiveView check so the URL never leads the DOM for a non-LiveView target.

## Notes

- Surfaced on a 1.0.8rc1 → 1.1.0rc1 upgrade; `auto_navigate` defaulting **on** in v1.1 (ADR-021 Stage 3) is what exposed it.
- I could not isolate *what* emits the `live_redirect` to the non-LiveView path (no app code or `dj-navigate` targets it; the client never constructs an outbound `live_redirect`). The **handling** bug above reproduces whenever any `live_redirect` resolves to a non-LiveView path, independent of the trigger. Happy to add a minimal repro if useful.
- Possibly related: #1858 (the other `url_change`/version-mismatch navigation bug).

johnrtipton:
## Summary

With `auto_navigate` defaulting ON in v1.1, a `live_redirect` whose target is a **non-LiveView** path (e.g. a plain Django `TemplateView`) stranded the page — the URL bar moved to the new path while the *previous* LiveView stayed mounted, with no DOM swap. Reported in #1934 against djust 1.1.0rc1.

## Root cause (symptom-up; the issue's cited cause was incomplete)

Two coupled bugs in `handleLiveRedirect` (`python/djust/static/djust/src/18-navigation.js`):

1. **`pushState` fired BEFORE the view resolution** — so the URL changed for a target that never got a DOM swap (the issue's primary citation).
2. **The load-bearing root cause (found by tracing the symptom, not the cited path):** the resolution used `resolveViewPath()`, whose **container fallback** returns the CURRENT `[dj-view]`'s class on a route-map miss. That fallback is documented *"only works for live_patch, not cross-view navigation"*. So for a cross-view `live_redirect` to a non-LiveView, `resolveViewPath('/onboarding/')` returned the **SOURCE** view (truthy, not falsy), and the client SPA-mounted the OLD view under the NEW URL — the exact reported console line (`URL=/onboarding/`, but `jira_manager.views.TicketListView` mounts). The server's #1647 `_resolve_view_path_from_url` guard also returns `None` for a non-LiveView URL and keeps the stale client-supplied view, so the client must make the full-nav decision.

The issue's "`resolveViewPath` correctly returns falsy" is only true when there is no `[dj-view]` container; in the real reported scenario the container fallback made it truthy, which is why a plain full-nav `else` (as the issue suggested) would not have fixed the production case.

## Fix

- New **STRICT** `resolveLiveViewPath()` — route map ONLY, no container fallback — drives the cross-view decision. The route map (auth-filtered, derived from the Django URLconf; `_walk_liveview_routes` excludes non-LiveViews) authoritatively answers "is this a LiveView?".
- **Defer `pushState`** + URL-dependent side effects (`updateAriaCurrent`, scroll, `djust:before-navigate`) into the LiveView-resolved + WS-connected branch, so the URL never leads the DOM.
- A **non-LiveView target (or a disconnected WS)** does a full-page navigation validated through `window.djust.safeNavigationTarget` (same `// codeql[js/xss]` pattern as the existing cross-origin branch; preserves the open-redirect / `javascript:` guard and the page-loading-bar stop semantics).
- The **popstate back-nav redirect** (the #1646 twin) also switched to the strict resolver, so a back-nav to a non-LiveView reloads correctly instead of re-mounting the source view.
- The prefetch-set clear stays **unconditional for same-origin** (moved above the resolve) — harmless on the full-nav path (context torn down), preserves the pre-#1934 contract that `prefetch.test.js` pins.

The cross-origin branch and the LiveView SPA-mount behavior are otherwise IDENTICAL.

## Bundle rebuild

Rebuilt the served bundle so the minified production artifact carries the fix: `client.js`, `client.min.js`, `client.min.js.gz`, `client.min.js.br`, `client.min.js.map`.

## Reproduce-first + gate-off evidence (#1468)

New `describe('issue #1934 …')` in `tests/js/navigation.test.js`:

- `non-LiveView target: full-page nav, NO pushState, NO WS mount (strand-free)`
- `positive case: a LiveView target still SPA-mounts (pushState + WS mount, no full nav)`
- `LiveView target but WS not connected: full-page nav (no strand), NO pushState`

Gate-off verified — each half of the fix is load-bearing:
- Revert the strict resolver (`resolveLiveViewPath` → `resolveViewPath`): the non-LiveView test goes RED (`safeNavigationTarget` never called — the SPA branch fired against the stale container view = the strand).
- Move `pushState` back before the resolve: the strand assertion goes RED (`expected pushState to be undefined`).

The two pre-existing tests that relied on the buggy pushState-before-WS-check ordering (`#307` live_redirect dispatch, `#1599` same-origin backstop) were rewired to a connected-WS source-env so they exercise the real SPA-mount path.

## Test results

- **JS suite:** `vitest run` — 1746 passed (151 files), incl. the new #1934 cases.
- **eslint:** clean on `18-navigation.js` (`--max-warnings 0`), no unguarded `console.log`.
- **Python suite:** `pytest tests/ python/tests/ python/djust/tests/ -n auto` — 8604 passed, 21 skipped, 0 failures (confirms the bundle/check tooling that asserts on `client.js` passes).

Closes #1934


johnrtipton:
## Stage 11 Code Review — PR #1935 (fix #1934) — **APPROVE** ✅

Adversarial review in an isolated worktree (`agent-a1bd320450fe3203d`, `core.bare=false`, main checkout untouched throughout). Branch is **0 commits behind `origin/main`** (passes the #1450 stale-base gate). I independently ran the gate-offs, rebuilt the bundle, and traced every resolver call site.

### Verdict rationale
The implementer's symptom-up root-cause is **correct and the issue's cited cause was genuinely incomplete**: `resolveViewPath()`'s container fallback returns the *current* `[dj-view]`'s class on a route-map miss, so a cross-view `live_redirect` to a non-LiveView returned the **source** view (truthy) and SPA-mounted the OLD view under the NEW URL — the exact reported strand. A plain `else { location.href }` as #1934 suggested would NOT have fixed the production case (the page always has a `[dj-view]` container). Both halves of the fix are real and load-bearing (verified below).

### 1. Resolver split is correct at EVERY call site ✅
Enumerated all call sites in `18-navigation.js` (the only file referencing either resolver; `handleLiveRedirect` is not exported, so no external callers):
- **`handleLiveRedirect` cross-view decision** (`:170`) → **STRICT** `resolveLiveViewPath`. Correct — cross-view must not honor the container fallback.
- **popstate back-nav redirect** (`:382`) → **STRICT** `resolveLiveViewPath`, falls through to `window.location.reload()` on a miss (`:440`). Correct #1646 twin fix.
- **`_handleAutoNavigateClick` guard** (`:647`) → **FALLBACK** `resolveViewPath`. Correct AND unchanged: `resolveViewPath` now *delegates* to `resolveLiveViewPath` then keeps the identical container fallback — byte-for-byte the same semantics as `origin/main`'s old function (verified via `git show`). The guard's job is only "should I intercept" → it routes to the now-fixed `handleLiveRedirect`, so a non-LiveView link still ends in full-page nav. (Aside: the issue's claim that auto-navigate "already worked / falls through to browser nav" is itself inaccurate for the on-a-LiveView-page case — the container fallback was truthy there too, so auto-navigate ALSO relied on the buggy `handleLiveRedirect`; this PR fixes both. Not a defect in the PR — noting for the record.)
- **`live_patch`** (`handleLivePatch`, `:31-83`) calls **neither** resolver — it has its own cross-origin check + `pushState`. So same-view query-only diffs **cannot** regress to a full-page nav. Confirmed by reading the full function; `prefetch.test.js` (4/4) and the same-origin backstop stay green.

### 2. Independent gate-off (#1468) — BOTH halves load-bearing ✅
Ran these myself in the worktree:
- **Revert strict→fallback** at `:170` → `non-LiveView target …` test goes **RED** (`safeNav` called 0×: the fallback returned the source view, SPA branch fired = the strand). ✔ strict resolver is load-bearing.
- **Inject pushState before the resolve** → strand assertion goes **RED** (`historyCalls` contains `pushState → /onboarding/`). ✔ deferred-pushState is load-bearing.
- **Bonus:** neutering the SPA resolve to falsy turns the two rewired tests RED (see #3).
Restored source after each; `git diff` clean.

### 3. The 2 rewired pre-existing tests are NOT neutered ✅
`BUG 1: handleNavigation dispatches…` and `same-origin paths still use pushState…` were rewired to a connected-WS source env and now assert the **real SPA path** (pushState + an outbound `live_redirect_mount` frame carrying the resolved view). Gate-off C (neuter the resolve) turns BOTH **RED** → they exercise the real mount path, assertions strengthened (not weakened).

### 4. Bundle integrity — verified non-stale rebuild ✅
Ran `scripts/build-client.sh` in the worktree: freshly-built `client.js` AND `client.min.js` are **byte-identical** (`diff -q`) to the committed artifacts. The minified bundle carries the actual wiring `isWSConnected()?resolveLiveViewPath` and the `live_redirect non-LiveView target` marker — not a stale or partial build. Restored the artifacts; git clean.

### 5. XSS ✅
Both full-nav sinks (`:173` new fallback, `:136` existing cross-origin) route through `window.djust.safeNavigationTarget` with the `// codeql[js/xss]` annotation and only assign `location.href` when non-null. Traced `02b-safe-nav.js`: `javascript:`/`data:`/`blob:`/`file:` → opaque origin `'null'` → rejected; cross-origin/`/\evil.com` normalize-then-reject (#1825). A `javascript:` `data.path` is caught at the cross-origin early-return (`:109`). No unvalidated sink introduced.

### 6. Tests green ✅
- Nav suite: **28/28**. Full JS suite: **151 files / 1746 passed / 0 failed** (matches PR claim). eslint `--max-warnings 0` on `18-navigation.js`: clean. Both new `console.*` are `globalThis.djustDebug`-guarded (`:182`, `:193`).

### Server-side citations verified
`_walk_liveview_routes` (routing.py:329 `issubclass(view_cls, LiveView)`) and `_resolve_view_path_from_url` (websocket.py:3119) both exclude/None non-LiveViews as the comments claim — the root-cause analysis holds.

### Concerns / edge cases checked
- 🟢 **WS-disconnected + LiveView target** → now full-page nav (`:170` `isWSConnected() ? … : null`). This is a *strict improvement*: the OLD path pushState'd then skipped the whole `if (isWSConnected())` block = a second strand variant, now fixed. Covered by the third new test.
- 🟢 **`_prefetch.clear()` moved above the resolve** (`:146`): now unconditional for same-origin (both branches), cross-origin still returns first without clearing — matches the pre-#1934 contract; harmless on full-nav (context torn down). `prefetch.test.js` green.
- 🟢 **`dj-navigate` → non-LiveView** (`:552` caller) also benefits — no longer strands.
- 🟡 (nit, non-blocking) `resolveLiveViewPath` is not exported on `window.djust.navigation` (only `resolveViewPath` is). Fine today — no consumer needs the strict variant externally; flagging only so a future external caller doesn't re-add a fallback-based cross-view check.

No 🔴 findings. The resolver-semantics split — the high-risk part — is correct at all five call sites, and both fix halves are empirically load-bearing.

**APPROVE.**

🤖 Stage 11 adversarial review (isolated worktree; gate-offs + bundle rebuild run independently)

johnrtipton:
## Retrospective — PR #1935 (#1934)
**Quality: 5/5** (the bug-triage canon paid off + adversarial review earned its keep).
### What went well
- **Symptom-up tracing found a deeper root cause than the issue cited** (CLAUDE.md 'trust the symptom, not the cited path'). The issue said resolveViewPath returns falsy → add an else; tracing the real symptom (the SOURCE view mounting under the new URL) revealed the container fallback made it TRUTHY, so an else would have been a no-op in production. The strict resolveLiveViewPath was the load-bearing fix.
- **#1646 parallel-path discipline**: the popstate back-nav redirect twin got the same strict resolver in the same PR, not left to drift.
- **Adversarial Code Review verified the resolver-split blast radius** (5 call sites) independently — confirmed live_patch uses neither resolver (can't regress to full-nav) and the auto-navigate fallback is byte-identical to old main; ran both gate-off halves + a third proving the 2 rewired tests aren't neutered; rebuilt the bundle to byte-match (non-stale).
### What to improve
- The issue's contrast claim ('_handleAutoNavigateClick already works') was inaccurate — a reminder that reporter-supplied 'the good path' references also need symptom-up verification, not just the cited bug path. Relates to prior bug-triage findings (#1205, #1638, #1650).
### Deferred (non-blocking)
- 🟡 resolveLiveViewPath not exported on window.djust.navigation — intentionally not filed; no external consumer needs the strict cross-view resolver today (speculative export = noise).
### Verified
- Reproduce-first + gate-off (both halves); JS 1746/1746 + nav 28/28; Python 8604; eslint clean; bundle rebuilt + verified; adversarial review APPROVE; CI all green. `RETRO_COMPLETE`

PR #1937, opened 2026-06-24
feat(types): M2 — strict type enforcement on the public-API quartet (live_view, component, decorators, forms) (ADR-023)

johnrtipton:
ADR-023 ratchet **M2** — flips the consumer-facing public-API quartet to strict mypy islands (the `py.typed`-exposed surface).

**Modules made strict** (annotated accurately, real types not `Any`): `live_view.py`, `components/base.py` (the LiveComponent base), `decorators.py`, `forms.py` — added to the `[tool.mypy]` strict-island overrides (same 4 flags as M1).

**Real type bugs fixed (the payoff):**
1. `decorators.event_handler` untyped dual-call API (`@event_handler` vs `@event_handler(...)`) produced `[arg-type]` + "Self argument missing" at every bare call site — the exact consumer-facing liability ADR-023 names. Fixed via `@overload`.
2. `live_view._is_serializable` tuple-width lie (`_non_serializable` inferred 3-tuple, 4th `LockType` elem fell outside the type) → `tuple[type, ...]`.
3. `live_view.pyi` stub drift — `stream` missing the `limit` param `StreamsMixin.stream` has.
4. `mixins/handlers.py` `_handler_metadata` — non-optional inferred dict conflicted with the `None`-init contract → `Optional[Dict[...]]`.
5. `decorators._ComputedProperty` custom metadata attrs undeclared → declared.

No runtime behavior change (annotations + narrow `# type: ignore`s are no-ops).

**Verification:** `mypy python/djust` = Success (822 files) with the 4 modules strict; gate-off (#1468) — wrong-typed return in now-strict `forms.py` → mypy RED, revert → GREEN; full suite 8604 passed; ruff clean.

Part of the ADR-023 type-enforcement ratchet (M1 #1936 → M2 here → M3 dispatch core → M4 long tail).

PR #1936, opened 2026-06-24
feat(quality): enforce incremental type-checking — mypy gate + strict islands + _rust.pyi stub + ADR-023

johnrtipton:
## Summary

Wires mypy from **dead config** into an **enforced merge gate** using the
canonical incremental-adoption pattern: a lenient global default that parks the
legacy baseline (so the gate is GREEN today) + **strict islands** that genuinely
enforce on a meaningful subset, led by the security boundary and the PyO3 wire
boundary. References the type-hinting initiative; closes no issue.

Part of the v1.1 code-quality arc (ADR-022 sibling): type enforcement makes a
whole class of contract drift mechanically detectable.

## The baseline (measured, verified)

djust ships `py.typed` (PEP 561 — downstream consumers type-check against our
hints) and `pyproject.toml` declared a strict `[tool.mypy]` config — but **mypy
was invoked nowhere** (CI / Makefile / pre-commit). The dead strict config over
the package reported **8,421 errors across 659 files**:
- ~6,814 `[no-untyped-def]` (missing annotations — the ratchet target)
- ~750 `[import-untyped]` (third-party / Django / PyYAML, no stubs)
- ~700 genuine type errors (`assignment`/`attr-defined`/`arg-type`/`union-attr`/`no-any-return`/`misc`)

Flipping the declared strict config on globally would make the gate permanently
red — which is why it was never wired.

## The strict-island set (enforced %)

**22 modules / ~90 functions / ~2.8% of package LOC** enforced strict
(`ignore_errors=false` + `disallow_untyped_defs` + `disallow_incomplete_defs` +
`warn_return_any`), each verified to pass clean (0 errors):

| Island group | Modules |
|---|---|
| **Security boundary** | `djust.security.*` (attribute_guard, error_handling, event_guard, json_script, log_sanitizer, mount, state_snapshot, `__init__`) |
| **PyO3 wire boundary** | `djust._rust` (typed by `_rust.pyi`) |
| DoS / rate surface | `djust.rate_limit` |
| Input validation / audit | `djust.validation`, `djust.permissions` |
| Well-annotated leaves | `djust.markdown`, `djust.schema`, `djust.signals`, `djust.async_result`, `djust.test_isolation`, `djust._client_ip`, `djust._log_utils`, `djust._html`, `djust._view_resolution`, `djust._deprecation`, `djust._context_provider` |

A narrow `module=["yaml"]` override keeps the `permissions` island green without
weakening its def/return strictness. The global default is NOT a blanket free
[...]

Issue #1938, opened 2026-06-24
bug(dev-env): worktree pre-push hook leaks core.bare=true into the SHARED .git/config, corrupting the main checkout


johnrtipton:
**Observed twice** during the ADR-023 type-ratchet (PRs #1936, #1937 work). A worktree-isolated subagent running `git push` (with the pre-push hook) ends up with `core.bare = true` written to the **shared** `.git/config` (line 4), which breaks `git status`/`git push` in BOTH the worktree AND the main checkout (every tracked file shows as deleted). Recovery: `git config core.bare false` in the shared repo.

**Mechanism (partial):** appears during the worktree's pre-push hook run (which runs the full suite). Not from `test_deploy_cli.py`'s `git init` (that's an isolated tmp-dir repo — verified clean). Likely a build/PyO3-repoint step (the #1804 `core.bare`-to-repoint-PyO3 pattern) triggered in the worktree environment, or a pre-commit hook interacting with the linked-worktree `.git` file. **It is in the shared config, so it is dangerous — it silently corrupts the primary working tree.**

**Mitigation in use:** worktree subagents push `--no-verify` (the established pattern, reference: worktree-subagent env note) — CI is the authoritative gate. Confirmed: a `--no-verify` push does NOT leak core.bare.

**Fix needed:** find what sets `core.bare=true` (audit the pre-push hook chain + any maturin/build/test that repoints PyO3 from a worktree) and either scope it (`git config --worktree`, or never touch core.bare), or guard the dev-env so a worktree push can't corrupt the shared config. Extends #1804/#300 (which were about review subagents; this is the pre-push hook path).

johnrtipton:
## Investigation: leak source is NOT in-repo; shipped a detect+recover helper (PR #1965)

**By-inspection audit** (the full pre-push suite is the leak trigger, so it was NOT run; only targeted single-test repros) plus empirical verification establishes that **no djust pre-push hook, test, or script writes `core.bare`**:

- Every in-repo git op is read-only (`git status`/`diff`/`grep`/`ls-files`/`rev-parse`/`hash-object`) or scoped to an isolated tmp dir. The three git-touching Python test files (`test_run_with_venv_python.py` — creates real worktrees; `test_git_commit_with_precommit.py`; `test_deploy_cli.py`'s `git init` fixtures) were each run **in isolation** and left `core.bare` unchanged (`false` → `false`).
[...]

PR #1939, opened 2026-06-24
feat(types): M3 — strict type enforcement on the dispatch/runtime core (websocket, runtime, sse, streaming) (ADR-023)

johnrtipton:
## Summary

ADR-023 ratchet **step 3 (M3)**: make the **dispatch/runtime CORE** pass strict mypy and flip it to strict islands. Follows M1 (#1936, the enforced gate) and M2 (#1937, the public-API quartet). Safe now that the ADR-022 ViewRuntime convergence has settled (no spine code about to move).

Five modules flipped to strict islands in `pyproject.toml` (`[[tool.mypy.overrides]]` with `ignore_errors = false` + `disallow_untyped_defs` + `disallow_incomplete_defs` + `warn_return_any`), mirroring the existing 6 override blocks exactly.

## Per-module annotation summary

| Module | Strict errors fixed | What was annotated |
|---|---|---|
| `runtime.py` | 42 | `ViewRuntime` dispatch helpers, `_build_request`/`_check_auth`/`_extract_cache_config`/`_extract_optimistic_rules`, the actor-mount path, `_tenant_context`, `_execute_async_task` |
| `websocket.py` | 73 | `LiveViewConsumer` lifecycle (`connect`/`disconnect`/`receive`), all `handle_*` verb handlers, Channels event handlers (`server_push`/`db_notify`/`presence_event`/`hotreload`/`client_push_event`), `_run_async_work`/`_dispatch_single_event`, `_mount_one`'s 5-tuple return, module helpers (`_snapshot_assigns`/`_compute_changed_keys`/`_build_context_snapshot`/`render_embedded_child_html`/`_tenant_context`/`_bind_tenant`/`_emit_full_html_update`), the `_SlotCollector` HTMLParser, `broadcast_reload`/`register` classmethods |
| `sse.py` | 17 | `DjustSSEStreamView.get` / `DjustSSE*EventView.post` (→ `HttpResponse`), owner-binding helpers (`_request_user_pk`/`_request_session_key`/`_request_owns_session`/`_client_cap_key`/`_client_ip_from_request`/`_sse_origin_allowed`/`_sse_content_type_is_json`), the `event_stream` SSE async generator (→ `AsyncIterator[str]`), `_flush_deferred_to_sse` |
| `streaming.py` | 6 | `StreamingMixin.__init__`, host-class cooperating attrs declared via `TYPE_CHECKING` (`template_name`/`template`/`get_context_data`) |
| `websocket_utils.py` | 7 | `_validate_event_security`, `_check_event_security`, `_format_handler_not_found_error`, `_ensure_handler_rate_limit`, `get_handler_coerce_setting`, `_call_handler` |

Real types throughout — only **two narrow coded `# type: ignore[arg-type]`** for genuine frame-dynamic edges: the dormant actor-event-name forward (`runtime.py`) and the no-binary `receive()` text frame (`websocket.py`).

## Real type gaps fixed (the payoff)

No runtime behavior change — each removes a latent contract lie:

- **`sse.py`** — `SSESession._request` was assigned (`DjustSSEStreamView.get`) and read (`runtime.SSESessionTransport.build_request`) but **never declared in `__init__`**. Added the `Optional[Any]` declaration alongside `_event_request`.
- **`runtime.py`** — `_instantiate_error_frame` was first-assigned `None` then a dict (mypy inferred `None`-only) → declared `Optional[Dict[str, Any]]` in `__init__`. `_instantiate_view` called `Optional[type]()` ("None not callable") → added the `ViewResolution.__bool__`-implied guard. The dormant actor mount called `Optional[create_session_actor]` → added the actor-availability guard.
- **`websocket.py`** — `_recovery_html` was `str`-typed from its first assignment but cleared to `None` on one-time use → annotated `Optional[str]`.
- **`live_view.pyi`** (M2 island) — the stub omitted the module-level `_FRAMEWORK_INTERNAL_ATTRS` that `websocket._snapshot_assigns` imports. Added it — a stub-completeness gap the M3 flip surfaced (websocket now resolves the import against the strict stub).
- Several `Any`-leaks at Django/PyO3 boundaries narrowed at the boundary (`bool()`/`str()`/`int()` wraps on `validate_host`, child-render output, version helpers).

## Verification

- **mypy GREEN**: `mypy python/djust` = `Success: no issues found in 822 source files` with all 5 modules strict.
- **Full suite**: `pytest tests/ python/tests/ python/djust/tests/ -n auto` → **8604 passed, 21 skipped, 0 failed**.
- **ruff**: check + format both clean on all 6 changed files.
- **Gate-off (#1468)**: injecting an `int` return into a now-strict `-> str` function (`sse._client_cap_key`) made `mypy python/djust` **RED** (`Incompatible return value type (got "int", expected "str")`); reverting restored **GREEN** — the island enforces, it is not cosmetic.
- **Source-pin caveat**: the first full-suite run caught 4 source-grep tests (#1466/#1894/#1465) that pin the literal save-block gate `target_view is self.view_instance`; a bulk `self.view_instance`→`view` replacement had over-reached into that pinned expression. Restored the gate verbatim (semantics were identical) — all 4 green.

## Commits (two-commit shape, #1173)

1. code + `pyproject.toml` strict-island override
2. CHANGELOG (`### Added` M3 + `### Fixed` real gaps)


johnrtipton:
## Stage 11 adversarial review — PR #1939 (ADR-023 M3: strict typing of the dispatch/runtime core)

**Verdict: APPROVE** (1 🟡, no 🔴). Reviewed in an isolated worktree; main checkout untouched, `git config` never run, `core.bare` = false throughout.

I went in expecting the bulk `view = self.view_instance` local-capture rename in `websocket.py` to hide a reachable stale-reference / TOCTOU 🔴 (canon #245/#1198). It does not. Detailed trace below.

### Empirical verification (all GREEN)
- `mypy python/djust` → **`Success: no issues found in 822 source files`**, all 5 modules strict.
- **Gate-off (#1468), independent injection**: changed `_clear_template_caches` `return caches_cleared` → `return "not an int"` in the now-strict `websocket.py` (declared `-> int`). mypy went **RED** (`error: Incompatible return value type (got "str", expected "int") [return-value]`); reverting restored GREEN. The island enforces — not cosmetic.
- Tests (worktree, real settings): full `python/djust/tests/` = **3871 passed, 3 skipped, 0 failed**; websocket/runtime/transport-parity/convergence slice = **322 passed**; save/reconnect/remount/1466/1934/async slice = **79 passed**; async-work/defer/send-version = **56 passed**. The #1466/#1465/#1894 source-pin tests (`test_ws_reconnect_state_1465.py`, `test_runtime_state_save_tt_1894.py`) pass (`test_handle_event_save_block_persists_state_to_session` green).
- `ruff check` + `ruff format --check` clean on all 5 modules.

### Concern 1 — stale-capture / TOCTOU (the PRIMARY check)
The captures split into two classes by concurrency context:

**Serial-loop / render-lock-held (SAFE):**
- `websocket._dispatch_single_event` (`websocket.py:1556`) — runs INSIDE `handle_event` holding `self._render_lock` (documented invariant at `websocket.py:1486`). No concurrent re-mount reachable; the only awaits are `sync_to_async` renders.
- `websocket.handle_request_html` (`websocket.py:3294`) — verb handler in the serial receive loop; single await region in the `_has_live_sticky_children` branch.
- The 3 NET-NEW runtime captures (`runtime.py` `_dispatch_event_render:2661`, `_dispatch_single_event:3105`, `_render_and_send:4052`) — all run under `event_context()` which borrows the consumer's real `_render_lock` (`runtime.py:991`+), each immediately `if view is None: return`-guarded. The inline comment at `runtime.py:2658-2662` states the lock invariant explicitly. (The pre-existing 15 runtime captures incl. `_execute_async_task`/`_render_async_result` are untouched by this PR.)

**Detached task (`websocket._run_async_work:1122`) — see 🟡 below.** This is the ONE capture whose `await` window is genuinely racy: `_run_async_work` is spawned via `asyncio.ensure_future` (`websocket.py:1072/1085`), so it runs concurrently with the serial receive loop, and `disconnect`/`handle_live_redirect_mount` (which null/reassign `self.view_instance` at `websocket.py:1879`, `3055`) are NOT cancelled on teardown. But this is NOT a 🔴 — see analysis.

### 🟡-1 — `_run_async_work` capture: behavior SHIFTS under teardown-during-async; "no behavior change" is slightly overstated (not a regression)
`websocket.py:1122` `view = self.view_instance` (guard `if view is None: return`), then `await callback` / `await view.render_with_diff()` / `await self._send_update`. `origin/main` re-read `self.view_instance` LIVE at every access after the await; HEAD uses the captured `view`.

Race: task suspends on `await sync_to_async(callback)(...)`; meanwhile `disconnect` nulls `view_instance` (or `live_redirect` re-mounts a *different* view). On resume:
- **Pre-patch**: re-reads `self.view_instance`. Disconnect → `None.render_with_diff()` → `AttributeError` → caught/logged (nothing sent). live_redirect → operates on the **NEW** view (cross-view state bleed).
- **Post-patch**: operates on the captured **OLD** view → renders it, attempts `_send_update` on the (closed/new) socket.

This IS a stale reference and IS an observable behavior shift, so the PR's "no runtime behavior change" claim is **not strictly accurate for this detached-task race**. However it is **not a 🔴**: the pre-patch baseline was *also* buggy in this untested race (AttributeError, or NEW-view contamination). Post-patch ("operate on the view that scheduled the work") is arguably the more defensible semantics; both versions are imperfect in the same pre-existing edge case, and the runtime's parallel `_execute_async_task` already used the capture-the-view pattern on `origin/main`. The PR neither introduces a correct→incorrect regression nor fixes the race. No test exercises async-work-during-teardown either way.

**Recommendation (non-blocking):** soften the CHANGELOG/PR "no behavior change" wording for `_run_async_work` to "behavior-equivalent except in the untested async-work-during-teardown race, where it now operates on the scheduling view rather than the live `view_instance`," and consider a follow-up issue to define correct semantics there (cancel detached async tasks on teardown, or identity-guard the resume). This is the #1645/#560 async-send-path family and out of scope for a typing PR.

### Concern 2 — #1466 save-block gate (security/correctness): UNCHANGED
The gate now lives in `runtime.py:2883` (`if target_view is self.view_instance and getattr(self.view_instance, "enable_state_snapshot", False):`) — post-ADR-022 the event flip routes through `ViewRuntime.dispatch_event`; `websocket.py` no longer carries it. **`git diff origin/main...HEAD` shows ZERO changes to the gate region in `runtime.py`** — comparison + semantics byte-identical (only surrounding lines shifted). The PR-body note about "restoring the gate text in websocket.py" reflects a mid-development state; the final diff has the gate intact and the source-pin tests green. ✅

### Concern 3 — the 5 real-gap fixes are annotation-truth (verified)
- `sse._request: Optional[Any]` — pure `__init__` declaration of an attr assigned in `.get` / read in `runtime.SSESessionTransport.build_request`, never declared. ✅
- `runtime._instantiate_error_frame: Optional[Dict[str, Any]]` — pure `__init__` declaration (was first-`None`-then-dict). ✅
- `runtime._instantiate_view` `if view_class is None: return None` guard — `ViewResolution.__bool__` is verified `return self.view_class is not None` (`security/mount.py:93-94`), so a truthy `resolution` (checked above) guarantees the guard never fires; `# pragma: no cover`, behavior-equivalent. ✅
- `runtime` actor guard `if create_session_actor is None: raise RuntimeError(...)` — replaces a `TypeError: NoneType not callable`; gated upstream, `# pragma: no cover`. ✅
- `websocket._recovery_html: Optional[str]` — annotation only (cleared to None on one-time use). ✅
- `live_view.pyi` `_FRAMEWORK_INTERNAL_ATTRS: frozenset[str]` — genuine stub-completeness gap; the symbol IS imported by `websocket._snapshot_assigns` (`websocket.py:323`) and exists at runtime (`live_view.py:98` as `frozenset`). Stub type matches. ✅
- `bool()/str()/int()` boundary wraps (`validate_host`, child-render, version helpers) — narrow `Any` at PyO3/Django boundaries, no logic change. ✅

### Concern 4 — type:ignores / Any-widening
- Exactly **2** `# type: ignore[arg-type]` added (both coded, narrow): `runtime.py` dormant actor-event-name forward; `websocket.py` no-binary `receive()` text frame. Matches the claim. No ignores removed; no blanket ignores. ✅
- The pre-existing `self.send_json = _collect  # type: ignore[assignment]` is unchanged.
- `actor_handle: Any` widen (from `Optional[SessionActorHandle]`) is justified — `SessionActorHandle` is a module runtime variable (None when actor build absent), not usable as a static type; documented inline. The other `Any` uses (`view_instance: Any`, `_request: Optional[Any]`, `_request_user_pk -> Optional[Any]`) are at dynamic Django/view boundaries, consistent with the established convention — they do not defeat the island (the gate-off proof above lands inside a strict module regardless). ✅

### Misc
- `pyproject.toml` strict-island block mirrors the existing 6 override blocks exactly (`ignore_errors=false` + `disallow_untyped_defs` + `disallow_incomplete_defs` + `warn_return_any`). ✅
- Branch is 0 commits behind `origin/main` (#1450 stale-base check passes); clean two-commit shape (#1173). ✅

**Net:** a clean, well-documented strict-typing flip. The bulk capture-rename is safe everywhere it matters; the only soft spot is an overstated "no behavior change" claim for the one detached-task capture, where behavior shifts inside a pre-existing untested race without introducing a regression. APPROVE.

johnrtipton:
## Retrospective — PR #1939 (ADR-023 M3)
**Quality: 5/5** (dispatch core strict; adversarial review caught a real nuance). 5 modules strict via local-capture mypy-narrowing + 5 gap-fixes; #1466 save-block gate byte-identical. **Honest caveat:** the CHANGELOG 'no runtime change' is ~99% true — 84/85 captures are no-ops, but the 1 detached `_run_async_work` capture is a behavior SHIFT in a pre-existing untested teardown race (neither better nor worse; filed #1940 for the teardown contract). The worktree-adversarial review (warranted by the 85× rename in the convergence core) traced each capture's reassignment-reachability + verified the gate byte-identical — exactly the TOCTOU/#245-#1198 check this batch needed. mypy green; gate-off RED; suite 8604; --no-verify push (no core.bare leak, #1938). `RETRO_COMPLETE`

PR #1959, opened 2026-06-25
feat(types): M4g — strict typing of components/rust_handlers (ADR-023, ratchet complete)

johnrtipton:
## ADR-023 M4e (group 2) — strict-type ratchet

Flips **20 modules** across three subpackages from the lenient mypy default to strict islands (`[[tool.mypy.overrides]]` with `ignore_errors = false` + `disallow_untyped_defs` + `disallow_incomplete_defs` + `warn_return_any`).

### Subpackages flipped

| Subpackage | Modules |
|---|---|
| `scaffolding/` | `__init__`, `gen_live`, `gen_live_templates`, `generator`, `templates` |
| `template_tags/` (underscore — Rust-engine handlers, distinct from `templatetags/`) | `__init__`, `url`, `static`, `pwa`, `templatetag`, `flash`, `markdown`, `client_config`, `live_render` |
| `theming/gallery/` | `__init__`, `views`, `context`, `component_registry`, `urls`, `storybook` |

None of the three has a `tests/` dir, so each completes in this one PR.

### What changed

**Annotations only — render output is byte-identical.** The SafeString/HTML boundaries return `Any` under the lenient global config (Django is untyped via `ignore_missing_imports`; the cross-island `live_tags._client_config_html` is declared `-> Any`), so each is coerced with `str(...)` at the boundary to satisfy `warn_return_any` **without changing the returned (already-safe) HTML**:

- `flash.py` → `format_html(...)`
- `markdown.py` → `escape(...)` fallback
- `pwa.py` → `Template(...).render(...)`
- `url.py` → `reverse(...)` (×3)
- `static.py` → `static(...)`
- `client_config.py` → `_client_config_html(...)`
- `component_registry.py` → dynamic component `.render()`

**One real type fix:** `scaffolding/generator.py` `list_display_fields` annotated `list[str]` (was an un-annotated `[]`, flagged `var-annotated`).

Other annotations are honest types: `HttpRequest`/`HttpResponse` on the gallery views, `list[dict[str, Any]]` on the example builders, `Callable[[Type[TagHandler]], Type[TagHandler]]` on the `@register` decorator factory.

### Verification

- `mypy python/djust` **GREEN** (822 files) — all 20 strict, the rest lenient.
- **Gate-off (#1468):** a wrong-typed `int` return injected into `template_tags/url.UrlTagHandler.render` (declared `str`) turns the gate RED (`[return-value]`); reverting restores GREEN.
- ruff check + format clean on all targets.
- Full suite **8604 passed, 21 skipped, 0 failed** (`pytest tests/ python/tests/ python/djust/tests/ -n auto`).

See `docs/adr/023-incremental-type-enforcement.md`.


johnrtipton:
Inline review (the final module — components/rust_handlers, 193 component tag handlers; the rendering-critical Rust-bridge boundary).
- **Flipped clean, 0 new ignores** (the M4b-1 344-error iceberg was real but tractable): _safe(html)->str cast(str, mark_safe(...)) wrapper for the ~200 no-any-return cascade (SafeString resolves to Any under unstubbed django, so the wrapper not the annotation is what absorbs it), inline cast()/str() no-ops for the kw.get()->object cascade.
- **Render byte-identity VERIFIED** (the load-bearing check for a render module): 382 outputs across all 193 handlers byte-identical vs pre-flip; lookup-key wraps used cast() (no-op) NOT str() to preserve key identity.
- mypy green (822, module strict); gate-off RED (ModalHandler.render); suite 8604; browser-smoke exercises these tags.
Verdict: **APPROVE**. ADR-023 ratchet COMPLETE — 100% of shipped non-test modules strict, zero lenient exceptions.

johnrtipton:
## Retrospective — PR #1959 (ADR-023 M4g, ratchet COMPLETE)
**Quality: 5/5.** The supposed 344-error iceberg flipped with 0 new ignores once the right pattern (a _safe() cast wrapper, since SafeString resolves to Any under unstubbed django) + inline kw casts were applied — a reminder that 'intractable boundary' deserves a second look with better tools. Render byte-identity (382 outputs) is the load-bearing proof for a render module. **Completes ADR-023: 100% of shipped non-test modules strict-enforced, zero exceptions.** `RETRO_COMPLETE`

PR #1960, opened 2026-06-25
fix(types): disable import-untyped globally — mypy clean in a full dev env (ADR-023 follow-up)

johnrtipton:
Post-ratchet fix. `mypy python/djust` was green in CI (minimal env: requests/yaml absent → import-not-found, suppressed by ignore_missing_imports) but RED in a full dev env (requests/yaml installed without stubs → import-untyped, reported against the importing strict-island modules permissions/deploy_cli/mcp.server). import-untyped is a different code than ignore_missing_imports covers + is scoped to the importer, so the ["yaml","requests"] override can't reach it. Fix: disable_error_code=["import-untyped"] in the global [tool.mypy]. mypy now Success (822 files) in a full local dev env. No code change; makes the ADR-023 gate environment-independent (clean local mypy + pre-commit for contributors with requests/yaml installed).

### DIFF
--- a/docs/adr/023-incremental-type-enforcement.md
+++ b/docs/adr/023-incremental-type-enforcement.md
@@ -1,8 +1,8 @@
 # ADR-023: Incremental Type Enforcement — mypy Gate + Strict Islands
 
-**Status**: Accepted — foundation shipped 2026-06-24 (this PR); the strict-island ratchet is ongoing.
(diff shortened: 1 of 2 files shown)


johnrtipton:
## Retrospective — PR #1960 (ADR-023 ratchet finalization)
**Quality: 5/5.** Caught a real env-dependence (CI-green / full-dev-env-red) in the post-ratchet re-measure: import-untyped is reported against the IMPORTER (strict islands) for an installed-without-stubs lib, a different code than ignore_missing_imports covers + unreachable from a per-lib override — fixed in the global config. ADR-023 status finalized. **The type-enforcement ratchet is complete.** `RETRO_COMPLETE`

Issue #1849, opened 2026-06-22
tech-debt: browser-smoke coverage for downstream interactive paths (runtime breaks invisible to pytest)


johnrtipton:
## Source
Retro v1.0.7-3+4 (security audit drain + 1.0.7 release).

## Problem
Two real runtime breaks shipped in the 1.0.7 upgrade that the **8237-passing pytest suite + HTTP-200 smoke both PASSED over** — caught only by manually driving the live page in a browser:

1. **Demo index** mounted `dj-view="demo_app.views_old.IndexView"` — a stale ref that rode the boundary-less `startswith` the F22 fix correctly tightened; now refused at WS mount ("View ... is not allowed"). Fixed by repointing to the canonical module.
2. **djust.org examples page** tab-switching + code-copy — an inline `<script>` inside the dj-root that the 1.0.7 WS-mount morph (#1610) re-creates without executing, so the delegated listener never registered (silent, no console error). See #1848. Fixed downstream by moving the script outside dj-root.

Both are runtime/wiring breaks (mount-path allowlist; inline-script execution under morph) that unit tests structurally cannot see.

## Ask
Add a thin **browser-smoke layer** exercised on framework upgrade for the key interactive paths of first-party downstream apps (demo + djust.org): load page → assert the LiveView mounts (no mount-refusal) → click the primary interactive controls (tabs, copy, nav) → assert the expected DOM change + no console error. Could be a small Playwright/Chrome harness run in CI on the demo, and a documented manual checklist for djust.org pages.

## Why it matters
Per the retro, the unit suite is blind to runtime wiring; browser/interactive testing of the DEPLOYED downstream is the only net that caught these. Make it a standing gate, not an ad-hoc manual catch.

PR #1961, opened 2026-06-25
ci: browser-smoke waits for demo-server readiness before page.goto — no cold-cache flake (#1943)

### LINKED ISSUE #1943: flaky: browser-smoke gate times out (page.goto 30s) on Rust cold-cache CI runs — demo server not ready in time
The blocking browser-smoke gate (#1869) red-barred PR #1942 on its first run with `Page.goto: Timeout 30000ms exceeded` navigating to /demos/browser-smoke/. Root cause: the run compiled the `djust_components` Rust crate from scratch (cache miss), so the nohup'd demo uvicorn server wasn't ready within the fixed 30s page.goto deadline. NOT a real break (cleared on warm-cache re-run; verified clean component import + full pytest suite). Because this gate is BLOCKING, a cold-cache flake intermittently red-bars unrelated PRs. Fix: replace the fixed 30s `page.goto` wait with a server-readiness poll BEFORE the playwright run (curl /healthz or /demos/browser-smoke/ until 200, bounded ~120s), and/or ensure the Rust build is cached/complete before starting the server + bump the goto timeout. Keeps the gate's signal while removing the cold-cache false-fail.

johnrtipton:
## Summary

Fixes the **cold-cache flake** on the BLOCKING `browser-smoke` gate (#1869). On a cold-cache CI run that compiles the `djust_components` Rust crate from scratch, the nohup'd demo uvicorn server wasn't ready within the fixed 30s `page.goto` deadline, so unrelated PRs were intermittently red-barred with `Page.goto: Timeout 30000ms exceeded` navigating to `/demos/browser-smoke/`. It cleared on a warm-cache re-run — **not a real break** — but because the gate is BLOCKING, the race intermittently blocked merges.

Closes #1943.

## Root cause (symptom-up)

The shared `.github/actions/djust-playwright-server` action already had a "Wait for server to be ready" step, but:

1. It polled only `/` (the homepage) — **not** the actual canary target route `/demos/browser-smoke/`, which is a *different* LiveView under `djust_demos.urls`. A 200 on `/` does not prove the canary route is ready; the first request to that route is what triggers any first-hit lazy import / compilation.
2. Its bound was only ~30s (`{1..30}` × `sleep 1`), too short to absorb a cold-cache first paint.

So `page.goto` raced its own fixed 30s deadline against an un-warmed route.

## Fix (the issue's recommendation — fixed at the shared action so all playwright jobs benefit)

- **`.github/actions/djust-playwright-server/action.yml`** — readiness step:
  - bound bumped **30s → 120s** (60 attempts × 2s);
  - polls the **actual canary target route** (`/demos/browser-smoke/`) in addition to `/`, using `curl -fsS` (fails on HTTP ≥ 400, so a half-initialized 5xx does **not** count as ready) — this warms the route's first-hit lazy import/compile **during** the bounded loop, so `page.goto` lands on an already-warm route.
- **`tests/playwright/test_browser_smoke.py`** — `page.goto(PAGE, timeout=60000)` (belt-and-suspenders, up from the 30s default).

## Gate signal preserved

A genuinely-down server still fails **LOUD** (`exit 1` + `server.log` dump) once the 120s bound is hit, so a real runtime break of either #1849/#1848 class still red-bars the PR. Only the cold-start race is removed.

## Verification

Playwright can't run locally (needs browser + server), so the PR's OWN CI `browser-smoke` run is the validator (must stay green). Locally verified:

- `yaml.safe_load` parses both `action.yml` and `test.yml`.
- `bash -n` syntax-checks the readiness run block.
- **Empirical poll-loop tests:**
  - server down (no listener) → loop hits the bound, **fails `exit 1` + dumps `server.log`** (gate signal intact).
  - canary route returns 500 ×2 then 200 → `-f` rejects the 5xx (route not marked ready), keeps polling, **succeeds only once `/demos/browser-smoke/` returns 200** (route is warm before `page.goto`).
- Non-playwright suite: `tests/ python/tests/ python/djust/tests/ -n auto --ignore=tests/playwright` → **8604 passed, 21 skipped, 0 failed**.


johnrtipton:
Inline review (CI reliability fix; the PR's own browser-smoke run is the empirical validator).
- Fix in the SHARED djust-playwright-server action (all playwright jobs benefit): readiness poll bumped 30s→120s AND now polls the ACTUAL canary route /demos/browser-smoke/ with curl -fsS (the original only hit / — a different LiveView, so a 200 there never proved the canary route was warm). Polling the real route warms its lazy first-hit compile during the bounded loop. +belt page.goto timeout 60s.
- **Gate signal intact** (the key check for a flake-fix): down-server → exit 1 + server.log dump; a 5xx is rejected by -f (only a real 200 counts) — verified both scenarios. Only the cold-start race removed, not the runtime-break detection.
- Non-playwright suite 8604; yaml parses; bash -n clean.
Verdict: **APPROVE** pending the PR's own browser-smoke CI staying green (the validator). ADR v1.1.0-4 / #1943.

johnrtipton:
## Retrospective — PR #1961 (v1.1.0-4 / #1943)
**Quality: 5/5.** Root cause was subtler than 'bump the timeout' — the poll hit / (a different LiveView) not the canary route, so it never warmed the route's lazy compile; polling the real route is the actual fix. Gate signal verified intact (down/5xx fail loud). browser-smoke CI green = empirical validation. `RETRO_COMPLETE`

Issue #245, opened 2026-02-06
WebSocket mount loses URL path kwargs (only query params forwarded)


johnrtipton:
## Problem

When a LiveView mounts via WebSocket, `client.js autoMount()` only forwards URL **query** parameters (`?key=value`) to the server. URL **path** kwargs (e.g., `recording_uuid` from `/recording/<uuid>/`) are lost.

This means any LiveView with dynamic URL segments silently receives empty strings for path kwargs on WebSocket mount, even though the initial HTTP GET works correctly (Django passes kwargs through `View.get()`).

### Reproduction

```python
# urls.py
path('recording/<str:recording_uuid>/', RecordingDetailView.as_view())

# views.py
class RecordingDetailView(LiveView):
    def mount(self, request, **kwargs):
        uuid = kwargs.get('recording_uuid', '')
        # HTTP GET: uuid = "abc-123" ✅
        # WebSocket mount: uuid = "" ❌
```

### Current behavior

In `client.js` `autoMount()` (line ~741):
```javascript
const urlParams = Object.fromEntries(new URLSearchParams(window.location.search));
this.mount(viewPath, urlParams);
```

Only `window.location.search` (query string) is sent. Path segments are ignored.

### Current workaround

Each view must manually resolve kwargs from the request path:
```python
def mount(self, request, **kwargs):
    self.recording_uuid = kwargs.get('recording_uuid', '')
    if not self.recording_uuid:
        from django.urls import resolve
        match = resolve(request.path)
        self.recording_uuid = match.kwargs.get('recording_uuid', '')
```

## Proposed Solution

In `websocket.py handle_mount()`, resolve the URL that the client already sends (`data["url"]` = `window.location.pathname`) against Django's URL patterns and merge the extracted kwargs into params:

```python
# In handle_mount(), before calling mount():
page_url = data.get("url", "/")
try:
    from django.urls import resolve
    match = resolve(page_url)
    # URL path kwargs as base, overridden by explicit client params
    params = {**match.kwargs, **params}
except Exception:
    pass

await sync_to_async(self.view_instance.mount)(request, **params)
```

This is a ~5 line change in the framework that eliminates a class of bugs for all apps with dynamic URL segments.

## Impact

Every LiveView with path parameters is affected:
- Recording detail (`/recording/<uuid>/`)
- Project detail (`/projects/<id>/`)
- Org-scoped views (`/org/<slug>/...`)
[...]

PR #1963, opened 2026-06-25
fix: define _run_async_work teardown semantics — no stale-view write on disconnect/re-mount mid-await (#1940)

### LINKED ISSUE #1940: bug: _run_async_work uses a captured view across a detached-task await — define teardown semantics for disconnect/re-mount during async work
Surfaced by the #1939 (ADR-023 M3) adversarial review. `LiveViewConsumer._run_async_work` (websocket.py:~1122) runs as a detached `ensure_future` task that is NOT cancelled on teardown, and it captures `view = self.view_instance` then `await`s. `disconnect`/`live_redirect` can null/reassign `self.view_instance` during that await. **Pre-existing untested race** — origin/main re-read `self.view_instance` live (→ AttributeError on disconnect, or NEW-view contamination on re-mount); #1939 captures the OLD view (→ writes against a stale view). NEITHER is correct. The PR neither introduces nor fixes the race. Fix: define the teardown contract — either cancel in-flight async work on disconnect/re-mount, or re-validate `self.view_instance is the captured view` after the await before acting (the #245/#1198 commit-or-rollback / identity-guard pattern). Same class as the runtime's `_execute_async_task`. Low severity (untested race, requires disconnect mid-async-work).

johnrtipton:
## Summary

`LiveViewConsumer._run_async_work` (`python/djust/websocket.py`) runs as a detached `asyncio.ensure_future` task. It captures `view = self.view_instance` **before** its first `await` (the background callback). During that await window:

- **`disconnect`** sets `self.view_instance = None` (`websocket.py:~1879`), and
- **`handle_live_redirect_mount`** / a re-mount reassigns `self.view_instance` to a **NEW** view (`websocket.py:~3055`, `_dispatch_runtime_owned` mount read-back).

Pre-fix, the completed task then ran `handle_async_result` + `_sync_state_to_rust` + `render_with_diff` + a `source="async"` frame against the captured (now-stale/replaced) view. This is a **pre-existing untested race** (#245/#1198 TOCTOU / commit-or-rollback class), surfaced by the #1939 adversarial review. As #1940 notes, every prior shape was wrong: origin/main re-read `view_instance` live (→ `AttributeError` on disconnect / NEW-view contamination on re-mount); #1939 captured the OLD view (→ stale write).

## Fix

**Identity-guard after the callback await** (the issue's recommended option, lower-risk than task-tracking + cancel). On **both** the success and error paths of `_run_async_work`, after the callback completes, re-validate `self.view_instance is view`; if it changed (torn down or replaced), **drop the re-render** before any state mutation or send.

Why identity-guard, not cancel-on-teardown: cancellation can't stop the in-flight worker thread (`sync_to_async` runs in a thread pool), and tracking + cancelling task handles would touch all three teardown sites. The guard is self-contained in `_run_async_work`, adds no consumer state, and the normal (no-teardown) async-work path is **byte-identical**.

## Reproduce-first + gate-off (#1468)

`python/djust/tests/test_run_async_work_teardown_1940.py` drives the **real** `_run_async_work` on a real `LiveViewConsumer`, with an async callback parked on a test-controlled `asyncio.Event` so teardown is forced deterministically (no thread timing):

- `test_disconnect_during_await_drops_stale_render` — `view_instance = None` mid-await → stale view receives no writes, no frame sent.
- `test_remount_during_await_does_not_contaminate_new_view` — `view_instance = NEW view` mid-await → neither old nor new view written by the stale task.
- `test_error_path_disconnect_during_await_drops_stale_render` — callback **raises** after a disconnect → error-state re-render skipped.
- `test_no_teardown_renders_normally` — **control / gate-off guard**: normal async work still renders + sends a `source="async"` frame.

**Gate-off verified**: flipping both guards to `if False and ...` makes the 3 teardown cases FAIL (the stale view IS written — the #1940 bug) while the control stays green — proving the tests exercise the teardown edge, non-tautologically.

## Verification

- `python -m mypy python/djust` — **GREEN** (823 files, strict island intact).
- Full suite across all roots (`pytest tests/ python/tests/ python/djust/tests/ -n auto`) — **8608 passed, 21 skipped, 0 failed**.
- All existing async-work suites green (76 in the async files).
- ruff check + format clean; `scripts/check-changelog-test-counts.py` passes.

Closes #1940


johnrtipton:
Inline review (websocket.py strict island + convergence core; the fix is the 2 identity-guards only).
- **Identity-guard after await** (self.view_instance is view) on BOTH success + error paths — correctly justified over cancel (sync_to_async runs in a thread pool, .cancel() can't stop it; the post-await guard is the real cure). #245/#1198 commit-or-rollback class.
- Normal no-teardown async-work path byte-identical (the existing async-work tests stay green).
- Deterministic repro (asyncio.Event-controlled interleaving, no thread timing) — 4 cases incl. error-path + no-teardown control; gate-off RED (flip both guards → stale write on the 3 teardown cases, control green).
- mypy green (823); suite 8608; 0 bare ignores.
Verdict: **APPROVE**. v1.1.0-4 / #1940.

johnrtipton:
## Retrospective — PR #1963 (v1.1.0-4 / #1940)
**Quality: 5/5.** The trickiest drain item (concurrency) done right: identity-guard (not cancel — justified: thread-pool work can't be cancelled), both paths, a deterministic Event-controlled repro (no thread timing), gate-off RED. mypy green; suite 8608. `RETRO_COMPLETE`

PR #1964, opened 2026-06-25
fix: ComponentMixin.update_component handles LiveComponent without AttributeError (#1947)

### LINKED ISSUE #1947: bug: ComponentMixin.update_component calls .update() on LiveComponent, which has no update() method (AttributeError)
Surfaced by the ADR-023 M4c mixins typing (PR #1946) via mypy [attr-defined]. `ComponentMixin.update_component` (`python/djust/mixins/components.py`) does `component.update(**props)` after `isinstance(component, LiveComponent)`, but `LiveComponent` (subclass of ContextProviderMixin, NOT Component) has NO `update` method at runtime (confirmed via live MRO). That path raises AttributeError if hit with a LiveComponent. PR #1946 annotated it with a narrow `# type: ignore[attr-defined]` + comment (out of scope per #1079) — this issue is the dedicated bugfix: either give LiveComponent an `update()`, route to the right method (`trigger_update`/`update_assigns`?), or guard the branch. Add a regression test. Low severity (requires the LiveComponent path through update_component).

johnrtipton:
## Summary

Fixes #1947. `ComponentMixin.update_component(component_id, **props)` calls `component.update(**props)` after `isinstance(component, LiveComponent)`, but `LiveComponent` (a subclass of `ContextProviderMixin`, **not** `Component`) had **no `update()` method at runtime**. Any `LiveComponent` that did not define its own `update()` therefore raised `AttributeError` on that path. This was the latent bug annotated with a `# type: ignore[attr-defined]` during the ADR-023 M4c strict-typing flip (left for a dedicated bugfix per #1079).

## Root cause (symptom-up)

- `Component.update()` exists (`components/base.py:208`); `LiveComponent.update()` did **not** — confirmed via the live MRO (`['LiveComponent', 'ContextProviderMixin', 'object']`).
- The existing test suite passed only because every test `LiveComponent` subclass *defined its own* `update()` (e.g. `TodoListComponent.update`, `ConfigComponent.update`). A bare `LiveComponent` with no override was never exercised — that is exactly the bug path.
- The component docs already document `component.update(**props)` / `child.update()` as the supported prop-update API (`docs/components/COMPONENTS.md`, `COMPONENT_UNIFIED_DESIGN.md`, `guides/components.md`). The base class simply never provided the implementation.

## Fix

Add a base `LiveComponent.update(**kwargs)` that sets each prop as an instance attribute (mirroring the Python/hybrid path of `Component.update`) and returns `self` for chaining. Subclass `update()` overrides still take precedence via normal method resolution. The now-unnecessary `# type: ignore[attr-defined]` at the `update_component` call site is removed.

This is the minimal, docs-aligned, behavior-preserving fix: it adds the missing-but-documented method without changing the routing or any existing override's behavior.

## Tests — reproduce-first + gate-off (#1468)

New `TestUpdateComponentNoUpdateOverride` in `tests/unit/test_component_parent_communication.py`:

| Test | Asserts |
|------|---------|
| `test_update_component_bare_livecomponent_no_attribute_error` | bare `LiveComponent` (no override) updates via base `update()` through `update_component` |
| `test_base_update_returns_self_for_chaining` | base `update()` returns `self` |
| `test_subclass_update_override_still_wins` | subclass `update()` override keeps custom behavior (no regression) |

- **Reproduce-first:** pre-fix, the bare-component test fails with `AttributeError: 'BareComponent' object has no attribute 'update'` at `mixins/components.py` `update_component`.
- **Gate-off:** renaming the new `LiveComponent.update` method turns 2 of the 3 new tests RED with that exact `AttributeError`; the override test correctly stays green (it doesn't depend on the base method). Non-tautological.

## Verification

- `mypy python/djust` — **GREEN** (no issues in 822 source files; strict-island `mixins/components.py` clean with the `type: ignore` removed).
- Full suite `pytest tests/ python/tests/ python/djust/tests/ -n auto` — **8607 passed, 21 skipped, 0 failed**.
- `ruff check` + `ruff format --check` — clean.
- Rebased onto current `origin/main` (0 behind) before opening.

Closes #1947.


johnrtipton:
Inline review (components/base.py + mixins/components.py — strict islands).
- Root cause symptom-up: LiveComponent (ContextProviderMixin subclass, NOT Component) never implemented the docs-documented update(); tests passed only because every subclass defined its own. Fix adds base LiveComponent.update(**kwargs) (attr-set + return self, mirrors Component.update's hybrid path); subclass overrides still win via MRO; removed the now-stale call-site type:ignore[attr-defined]. Docs-aligned, behavior-preserving (Component branch + overrides unchanged).
- Reproduce-first (bare component AttributeError) + gate-off RED (rename update → 2/3 RED, override test green = non-tautological). mypy green (822, ignore removed); suite 8607; 0 bare ignores.
Verdict: **APPROVE**. v1.1.0-4 / #1947.

johnrtipton:
## Retrospective — PR #1964 (v1.1.0-4)
**Quality: 5/5.** Type-ratchet-surfaced bug, fixed reproduce-first + gate-off RED + mypy-green; full suite green. `RETRO_COMPLETE`

PR #1965, opened 2026-06-25
fix(dev-env): detect + recover the core.bare worktree shared-config leak (#1938)

### LINKED ISSUE #1938: bug(dev-env): worktree pre-push hook leaks core.bare=true into the SHARED .git/config, corrupting the main checkout

johnrtipton:
Closes #1938.

## Summary

A linked \`git worktree\` shares **one** \`.git/config\` with the main checkout (the worktree's \`.git\` is a *file* pointing at \`<main>/.git/worktrees/<name>\`, and \`core.*\` is read from the shared \`[core]\` section). If anything flips \`core.bare\` to \`true\` there, \`git status\` / \`git push\` break in **both** the worktree **and** the main checkout (every tracked file shows as deleted). Observed ~6× this session; recovery is \`git config core.bare false\`.

## Investigation finding — the leak source is NOT in-repo

An exhaustive **by-inspection** audit (the issue forbids running the full pre-push suite, which is the leak trigger) plus targeted single-test repro established that **no djust pre-push hook, test, or script writes \`core.bare\`**:

- Every in-repo git operation is read-only (\`git status\` / \`diff\` / \`grep\` / \`ls-files\` / \`rev-parse\` / \`hash-object\`) or scoped to an isolated tmp dir (\`test_run_with_venv_python.py\`, \`test_git_commit_with_precommit.py\`, \`test_deploy_cli.py\`'s \`git init\` fixtures — the last set \`GIT_CONFIG_GLOBAL=/dev/null\`).
- The three git-touching Python test files were run **in isolation** and verified to leave \`core.bare\` unchanged (\`false\` before AND after).
- Zero git operations in the Rust crates (\`cargo test\` path), no \`build.rs\`, no \`maturin\` in the pre-push chain, no \`.cargo\` config.
- No global/system \`init.templateDir\` or \`core.bare\`; the git template dir is stock (no config-injecting template). No git aliases / wrapper.
- The shared config's line 4 is \`bare = false\` — the leak *modifies* this existing key, the exact signature of \`git config core.bare true\`.

So the corruption is **external** (a build/PyO3-repoint \`git config core.bare true\` per the #1804/#300 pattern, or an IDE/GitKraken integration — the config carries \`gk-last-accessed\`/\`vscode-merge-base\` keys), not a framework bug. It is unpinnable from in-repo inspection.

## Fix — a load-bearing detect + recover helper (not a decorative pin)

Per the #1804 canon ("verify \`git config core.bare\` is false as a reflex") and the #1859 rule (a pin must be load-bearing), this ships a runnable detector rather than a self-comparing test:

- **\`scripts/check-shared-git-config.sh\`** reads \`core.bare\` from the SHARED config (resolved via \`--git-common-dir\`, so it works identically from any worktree or the main checkout), reports a leak (\`exit 1\`), and with \`--fix\` performs the documented recovery (\`core.bare false\`). It **never** writes \`core.bare true\`.
- **CONTRIBUTING.md** "Working in a \`git worktree\`" gains a shared-config-corruption subsection documenting the \`--no-verify\` mitigation (the established worktree-subagent pattern; CI is the authoritative gate) + the detector.

## Tests

5 cases in \`tests/test_check_shared_git_config.py\` build a throwaway main+worktree (the #1938 topology), simulate the leak in the **throwaway** shared config, and assert detect + \`--fix\`-recover + a never-writes-\`true\` invariant. All git ops are isolated to \`tmp_path\` with \`GIT_CONFIG_GLOBAL/SYSTEM=/dev/null\`; the real repo is never touched. **Gate-off self-tested (#1468):** neutering the script's leak-detection branch makes \`test_checker_detects_leak_from_worktree\` fail.

| Concern | File | Test |
|---|---|---|
| Detect leak from worktree | \`scripts/check-shared-git-config.sh\` | \`test_checker_detects_leak_from_worktree\` |
| \`--fix\` recovers | same | \`test_checker_fix_recovers_leak\` |
| Never writes \`true\` | same | \`test_checker_never_writes_bare_true\` |
| Healthy = OK | same | \`test_checker_passes_when_bare_false\` |
| Mitigation docs | \`CONTRIBUTING.md\` | (prose) |

## Verification

- 5/5 tests pass; ruff check + format clean; CHANGELOG test-count hook green.
- Pushed \`--no-verify\` (the documented worktree mitigation — the native pre-push hook would itself risk the leak class); CI is the authoritative gate.
- \`core.bare\` verified \`false\` at every step of this work.


johnrtipton:
Inline review (detector + docs + tests; no production code).
- **Correct resolution**: the investigation proved the leak is NOT in djust (all in-repo git ops read-only/tmp-scoped; 3 git-touching tests don't flip core.bare in isolation; no maturin/cargo git ops). The shared config's gk-last-accessed/vscode-merge-base keys → external IDE/GitKraken integration (the #1804/#300 class). Shipping a detector+recovery rather than a phantom code fix is the honest call (#1079/symptom-up).
- check-shared-git-config.sh reads core.bare via --git-common-dir (any worktree), --fix recovers, NEVER writes true — makes the #1804 reflex runnable. CONTRIBUTING documents the --no-verify mitigation. 5 guard tests (throwaway repos, real repo untouched); gate-off RED (non-tautological).
- No djust production code changed.
Verdict: **APPROVE**. v1.1.0-4 / #1938 — investigation + detector (root cause is external/unpinnable in-repo).

johnrtipton:
## Retrospective — PR #1965 (v1.1.0-4 / #1938)
**Quality: 5/5.** The investigation correctly refused to ship a phantom code fix — proved the leak is external (IDE/GitKraken, per the gk-last-accessed config keys) by safe inspection + targeted-repro (never the full suite), and shipped a runnable detector+recovery+docs instead. gate-off RED; no production code. The right call for an external/unpinnable root cause. `RETRO_COMPLETE`

Issue #1967, opened 2026-06-26
perf: keyed-subtree render memoization for large-list/table render_with_diff (O(n) → O(changed))


johnrtipton:
## Bottleneck (from a request-path performance pass)

`render_with_diff` is the dominant steady-state cost for list/table-heavy views. Warm pytest-benchmark + a targeted attribution profile + a scaling sweep:

**Attribution (50-item keyed-list reorder, 485 µs/cycle):**
- `render_with_diff` (Rust): **407 µs — 84%**
- `update_state` (Python→Rust marshalling): 50 µs — 10%
- (shuffle, benchmark-only: 27 µs)

**Scaling (linear, ~8–9 µs/item):**

| items | µs/cycle | µs/item |
|------:|---------:|--------:|
| 10  | 101  | 10.1 |
| 50  | 417  | 8.3 |
| 100 | 837  | 8.4 |
| 200 | 1833 | 9.2 |
| 500 | **4512** | 9.0 |

→ a 500-item list reorder is **4.5 ms**; ~9 ms at 1000 items. This is the single biggest steady-state cost for list/table views (counter event/diff is ~3.8 µs by contrast).

## Root cause

`render_with_diff` re-renders the **entire** template to a fresh VNode tree (all N item subtrees) and diffs it against the old tree on *every* update — even a pure **reorder** where the item subtrees are byte-identical (only positions changed). The render half dominates; the per-item subtree is rebuilt from scratch each time.

## Proposed fix — keyed-subtree render memoization

Cache the rendered VNode subtree per item **key** and reuse it for unchanged keys on reorder/append, turning the render from O(n) into O(changed). On a reorder of unchanged items, no item subtree should be re-rendered — only the parent's child-ordering diffs (the keyed-move path, cf. #1678/#1682).

**Extension point:** `crates/djust_templates/` already has a partial fragment cache (`lib.rs:193` "per-node fragment cache", `parser.rs:1620` If-fragment reuse). The lever is extending keyed reuse to the list-item render path (loop body keyed by the item's `dj-key`/`id`).

**Correctness care (must-haves before claiming the win):**
- Invalidate the cached subtree when an item's **content** changes, not just its position (key stable but fields mutated → must re-render that item).
- Interaction with the existing keyed VDOM diff (#1678/#1682 InsertSubtree-before-MoveSubtree ordering).
- Nested-conditional boundaries inside item bodies (dj-if markers).

## Acceptance / verification

[...]

PR #1968, opened 2026-06-26
perf(cold-start): warm the Django→Rust filter bridge at startup, not first mount

johnrtipton:
From a request-path performance pass. The profiler showed the **first mount after server boot** pays a one-time **~20 ms** cost that steady-state requests don't: `rust_bridge._ensure_custom_filters_bridged()` lazily triggers Django to import *every* templatetag library (via `engine.template_libraries`) on first access. It's memoized (`_CUSTOM_FILTERS_BRIDGED`), so only the first request ate the latency.

## Fix
`DjustConfig.ready()` now eagerly runs the bridge via a new `_warm_filter_bridge()` helper, shifting that one-time cost from the first user's request to startup. Mirrors the established HVR-auto-enable gating:
- pytest-skipped (so the suite isn't slowed + test bootstrap orderings without a configured template engine are unaffected)
- non-fatal try/except (startup must never break)
- opt out via `LIVEVIEW_CONFIG['filter_bridge_warm'] = False`

No steady-state behavior change — this only relocates a one-time import.

## Tests (in `test_auto_hot_reload.py`, which owns the `ready()` test infra)
- `ready()` warms the bridge when the pytest env is cleared
- pytest env skips the warm (isolation guard)
- `_warm_filter_bridge()` flips the `_CUSTOM_FILTERS_BRIDGED` guard (reset-to-False first → non-tautological, #1200)
- **gate-off:** `filter_bridge_warm=False` → guard NOT flipped, method returns False

## Verification
- 10/10 `test_auto_hot_reload.py` pass · mypy `python/djust` green (823, strict) · ruff clean

## Related
The bigger steady-state bottleneck found in the same pass — large-list `render_with_diff` (O(n) at ~8–9 µs/item; 4.5 ms for 500 items) — is tracked separately as a Rust keyed-subtree-cache optimization in #1967 (not in this PR).


johnrtipton:
Inline review (apps.py strict island + test; low-risk cold-start fix).
- Profiler-driven: first-mount ~20ms = lazy Django templatetag import on first `template_libraries` access; memoized after. `ready()` now warms via `_warm_filter_bridge()` — shifts the one-time cost to startup, no steady-state change.
- Mirrors the HVR gate exactly (pytest-skip + opt-out flag + non-fatal try/except). Extracted helper is testable + gate-off-able.
- 4 new tests: ready-wiring (warm called / pytest-skip), guard-flips (reset-first, non-tautological #1200), opt-out gate-off. mypy green (823); ruff clean; 10/10 pass.
Verdict: **APPROVE**. (The bigger render_with_diff bottleneck is tracked in #1967.)

johnrtipton:
## Retrospective — PR #1968 (cold-start filter-bridge warm)
**Quality: 5/5.** Profiler-driven one-time-cost relocation; mirrors the established HVR ready()-gate pattern (pytest-skip + opt-out + non-fatal), extracted a testable helper, gate-off via the opt-out test (#1200/#1468). No steady-state change. The high-leverage steady-state bottleneck (large-list render_with_diff, ~9µs/item) is scoped + tracked in #1967 rather than rushed. `RETRO_COMPLETE`

PR #1969, opened 2026-06-26
perf(rust): keyed loop-item render cache for large-list render_with_diff (#1967, flag-gated)

### LINKED ISSUE #1967: perf: keyed-subtree render memoization for large-list/table render_with_diff (O(n) → O(changed))

johnrtipton:
Closes #1967.

## What

A **persistent per-item loop render cache** in the Rust template engine. `Node::For` previously re-rendered every loop item from the AST on every render, so a pure reorder of a 50/500-item keyed list rebuilt all N item subtrees from scratch (~9 µs/item) even though only their positions changed. This caches each item's rendered fragment by a content hash and reuses it across `render_with_diff` calls, turning the loop-**render** phase from O(n) toward **O(changed)**.

Flag-gated **default OFF** (`LIVEVIEW_CONFIG['loop_render_cache_enabled']`, split-foundation #1122) — when off, the For-node path is byte-identical to before.

## Design (corrected for two landmines)

| Piece | Where |
|---|---|
| `LoopRenderCache` — content-hash → fragment, **persistent** on `RustLiveView` (a per-render cache gives zero reorder benefit), `prune()` bounds memory to current items | `crates/djust_templates/src/loop_cache.rs` |
| **Position-dependence guard** — caching DISABLED for bodies whose output depends on loop POSITION: `{% if %}` (dj-if marker index, #1832), `{% cycle %}`, nested `{% for %}`, `{{ forloop.* }}`, or any opaque Python/component tag. Detected once per For-node + memoized. | `loop_cache.rs::body_is_position_dependent` |
| For-node integration (hash bindings → cache get/insert) | `crates/djust_templates/src/renderer.rs` `Node::For` |
| Threading via a render-scoped **thread-local** + RAII `LoopCacheGuard` (save/restore for re-entrancy) — avoids churning ~22 recursive render call sites; single-thread-correct since `render_with_diff` is synchronous | `loop_cache.rs` + `crates/djust_live/src/lib.rs` |
| Flag → Rust: `set_loop_render_cache_enabled` wired from `rust_bridge.py` (on init + cache-hit restore) | `python/djust/config.py`, `python/djust/mixins/rust_bridge.py` |

**Why it's safe with the keyed VDOM diff (#1678/#1682):** the cached fragment is the template-render output **before** dj-id assignment (dj-ids are assigned downstream in the html5ever parse phase), so the keyed diff is unaffected — verified byte-identical output (incl. `dj-id` reassignment) on a `dj-key` reorder, cache on vs off.

## Correctness proof (load-bearing)

- **cached == uncached** byte-identical across initial / reorder / content-change / append / remove, on plain, `forloop.counter`, `dj-if`, `{% cycle %}`, nested, tuple-unpacking, and `dj-key` templates.
- **Gate-off #1 (#1468):** neutering the position guard makes the byte-identity test go RED (a `{% cycle %}` reorder emits stale `odd`/`even`; a dj-if reorder emits stale marker positions) — the guard is load-bearing.
- **Gate-off #2 (#1468):** a per-render (non-persistent) cache yields **0** reorder hits vs the persistent cache's **3** — persistence is load-bearing.

## Hit/miss accounting (end-to-end via `render_with_diff`)

| step | hits | misses |
|---|---|---|
| initial (cold) | 0 | 3 |
| **reorder** | **3** | **0** |
| content-change of 1 | 2 | 1 |
| append 1 | 3 | 1 |

## Bench (render phase, criterion, `crates/djust_templates/benches/loop_render_cache.rs`)

| N | cache OFF | cache ON | speedup |
|---|---|---|---|
| 50 | 94.1 µs | 55.1 µs | **1.71×** |
| 500 | 907 µs | 567 µs | **1.60×** |

NOTE: the end-to-end `render_with_diff` win is bounded by the (uncached) html5ever-parse + VDOM-diff phases (Amdahl). In local per-phase measurement those phases dominate (~60%), so the full-cycle gain on a reorder is ~6–11% even though the render phase itself drops 36–45%. This PR optimizes the render half — the cost #1967 cites as dominant.

## Tests

- `crates/djust_templates/tests/test_loop_render_cache_1967.rs` — 12 Rust cases (battery + behavior + 2 gate-offs).
- `python/djust/tests/test_loop_render_cache_1967.py` — 11 end-to-end via `RustLiveView.render_with_diff` (flag default-off, on==off identity, reorder hits, change miss, position-dependent body not cached).

## Verification

- Full Rust suite green: workspace-excl-djust_live (all keyed-diff / dj-if-marker tests #1678/#1682/#1358/#1260/#1420 etc.) + `djust_live --no-default-features` (37 passed).
- Full Python suite green: **8632 passed, 21 skipped, 0 failed** across `tests/ python/tests/ python/djust/tests/`.
- `cargo clippy` + `cargo fmt --check` + `ruff` clean on changed crates/files.


johnrtipton:
## Stage 11 — Adversarial Code Review: REQUEST_CHANGES 🔴

A hot-path render cache lives or dies on output-identity, and the adversarial pass (worktree build, end-to-end via `RustLiveView.render_with_diff`) **broke it**.

### 🔴 must-fix — outer-context staleness
`content_hash` (`loop_cache.rs:569`) hashes ONLY the loop-variable bindings, and the cache is persistent across renders. But a position-independent body can also read OUTER-context vars, which are constant *within* a render, not *across* renders. So when an outer var changes while items don't, a reorder serves stale cached fragments.

E2E reproduced (cache ON vs OFF, items unchanged, `prefix` A→B):
- `{% for x in xs %}<li>{{ prefix }}-{{ x.name }}</li>{% endfor %}` → ON serves stale `A-...` (cache hits), OFF (correct) `B-...`
- `{% with label=flag %}`, `{% firstof flag x.name %}` → same stale-outer-var class

Realistic target case (currency symbol, localized label, selected-filter highlight, per-row current-user check). It's the rc4 coverage-completeness failure: the 'what the body reads' axis has two variants (loop vars + outer context); the battery only tested loop vars.

**Fix (Option A, conservative — matches the opaque-tag philosophy):** treat a loop body as cacheable ONLY if its variable dependencies are a subset of the loop's bound names — i.e. it reads NOTHING but the loop item(s). Any outer-context read → non-cacheable. Add the outer-context-change case to the byte-identity battery.

### PASS (independently verified)
- Default-OFF + off==byte-identical (gate at the For-arm; `disabled_cache_is_inert` + e2e).
- Both gate-offs load-bearing (position guard → cycle reorder RED when neutered; persistence → 0 hits when per-render). Non-tautological — but neither covered the outer-context axis (why the bug shipped green).
- Position guard exhaustive over 26 Node variants, conservative (opaque tags → non-cacheable).
- Thread-local + RAII re-entrancy clean (no cross-view leak). Pre-dj-id (keyed diff #1678/#1682 unaffected; dj-key reorder round-trips). Suites green. 0 behind main.
- 🟡 minor: u64 hash, no equality-check-on-hit (collision = silent poisoning; negligible at scale, default-off — a 128-bit hash or equality-on-hit would harden).

**Verdict: REQUEST_CHANGES** — fix the 🔴 outer-context staleness + extend the battery, then re-review.

johnrtipton:
## Stage 12 — Re-Review: APPROVE ✅
The 🔴 outer-context staleness is resolved. `body_is_cacheable(nodes, var_names)` now gates on BOTH (1) not position-dependent (the pre-existing if/cycle/nested/forloop/opaque guard) AND (2) every top-level var-root the body reads ∈ the loop's bound names (the new dep-subset check, reusing the engine's `body_root_var_names` partial-render extractor) — memoized per For-node. Conservative narrowing (can only make FEWER things cache → strictly safer): item-only bodies () still cache; any outer-context read (//) falls back to correct per-item render.
- New outer-context battery (Rust 12→17, Python 11→13) + the 3rd gate-off proven load-bearing (revert the dep-subset check → 4/5 Rust + 2 Python RED; item-only positive control stays GREEN).
- Win survives for cacheable bodies (~1.6–1.7× render-phase). CHANGELOG honestly narrowed + documents both gates. 0 behind main.
Verdict: **APPROVE** (pending CI). The end-to-end win is Amdahl-bounded (~6–11% reorder, flag-gated default-OFF) — the bigger lever (parse phase) is a tracked follow-up.

johnrtipton:
## Stage 14 — Retrospective (PR #1969 / #1967)
**Quality: 4/5.** Correct, well-tested, flag-gated — but the first pass shipped a real 🔴 the mandatory hot-path adversarial review caught.

### What went well
- The **mandatory worktree-adversarial Code Review earned its keep**: it broke output-identity end-to-end (outer-context staleness) that the impl's green battery missed — exactly the gate that prevents a hot-path cache from corrupting renders. Fix was conservative (two-gate, narrows the cacheable set → strictly safer).
- Flag-gated default-OFF (#1122) was the right call for a hot-path change — zero default risk, soak-then-default.
- Honest measurement: the implementer self-reported that the issue's 'render=84%' premise didn't reproduce (parse+diff ~60% of render_with_diff), bounding the end-to-end win.

### What didn't (lessons)
- **Coverage-completeness recurrence (rc4 rule):** the impl's cached==uncached battery tested the 'body reads loop vars' axis but not the 'body reads outer context' axis — the bug lived entirely in the untested variant. The rc4 rule ('a coverage suite must enumerate EVERY variant of the surface') applies to correctness batteries, not just wire-format suites.
- **Attribution granularity:** my original perf pass measured at the `render_with_diff()` Python boundary (render+parse+diff lumped) and attributed 84% to 'render', but didn't decompose WITHIN it. Profile one level deeper before scoping an optimization.

### Outcome
A correct, flag-gated render-phase optimization (~1.6-1.7× render-phase; item-only bodies). The **bigger lever is the parse phase** (~60% of render_with_diff) — filed as a follow-up. `RETRO_COMPLETE`

PR #1976, opened 2026-06-30
docs(roadmap): correct the #1970 bucket label (v1.0.8-3 → v1.1.0-1)

johnrtipton:
1.0.8 was already shipped/frozen; the parse-phase loop-render-cache work (#1970) develops on the v1.1 line, so its ROADMAP bucket is v1.1.0-1 (ships in 1.1.0), not v1.0.8-3. Heading + prose + Priority-Matrix Target column corrected; added a one-line historical note. Docs-only.

### DIFF
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -3714,7 +3714,9 @@ the old placement still serves during blue/green). `djust deploy` should print
 
 ---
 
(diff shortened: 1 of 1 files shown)


johnrtipton:
## Retro — PR #1976
Docs-only ROADMAP label correction (re-fix after the main restore re-introduced the original v1.0.8-3 mislabel). `RETRO_COMPLETE`

Issue #1977, opened 2026-07-01
VDOM: stale last_vdom baseline after state-restore/reconnect → SetText patches land on #text (recovery)


johnrtipton:
## Summary

When a `{% for %}` list is **replaced** with a shorter, differently-shaped list (e.g. a filter narrows a table), `diff_html` emits `SetText` patches whose paths address the wrong node — sometimes a `#text` node — in the DOM the client holds. The client patcher can't apply them:

```
[LiveView] Applying 220 patches
[LiveView] Path traversal failed at index 1, only 0 children (parent: #text).
[LiveView] Patch failed (SetText): node not found at path=3/13/1/3/7/1/0, dj-id=none
[LiveView] Patch failed (SetText): node not found at path=3/13/1/10/7/1/0, dj-id=none
[LiveView] 2/220 patches failed (218 succeeded)
[LiveView] VDOM patches failed … Requesting recovery HTML from server.
```

It self-heals via `html_recovery`, so it's not user-blocking — but the diff is producing unapplyable patches, and every filter change pays a full recovery-morph round-trip + flicker.

**Version:** djust 1.1.0rc3

## Root cause (evidence)

Each row in the list has cells whose **child count varies by row data** (a `{% if %}` cell renders a `<span>`, a different `<span>`, or nothing; `{% for %}` cells render 0..n `<span>`s). When the list shrinks (5 rows → 2) and the surviving rows have different conditional shapes than the rows previously at those positions, the differ computes `SetText` paths from the *old* tree that no longer address the same text node in the *new* tree.

Captured from the real diff (list filtered 5 rows → 2), the `SetText` patches carry **later-row text written to earlier-row paths**:

```
SetText path=[3, 13, 1, 1, 1, 0, 0]  text="#104"            # row-2 PR number …
SetText path=[3, 13, 1, 1, 2, 0]     text="bump 104"         # … row-2 title …
SetText path=[3, 13, 1, 1, 3, 0]     text="dependabot[bot]"  # … row-2 author …
```

i.e. row-2's cell values are being routed to row-1's cell paths after the list shrank. Against the browser DOM some of those paths descend into a `#text` node → "Path traversal failed … parent: #text".

## Reproduction

Drives a real `LiveView` through `LiveViewTestClient` (the true mount → event → `render_with_diff` pipeline, so it exercises the real normalize→extract→`diff_html` path):

[...]

PR #1978, opened 2026-07-01
test(vdom): make client-faithful harness dj-id resolution case-sensitive (#1977 follow-up)

johnrtipton:
## Summary

Surfaced during the #1977 investigation (whose `diff_html` list-shrink misroute claim the maintainer **retracted** — the differ is correct across ~18 faithful-pipeline shapes). The one verified, in-repo defect that came out of it: **the client-faithful VDOM harness is not faithful for larger fixtures.**

`tests/js/vdom_client_faithful_diff.test.js` (via `createHarnessDom`) runs the **real** `client.js` against a jsdom DOM. jsdom's selector engine (nwsapi) matches custom-attribute **values** case-**insensitively**, so `node.querySelector(':scope > [dj-id="C"]')` also matches `dj-id="c"`. djust dj-ids are base62 (`0-9a-zA-Z`) from a monotonic counter, so any render past ~38 nodes emits sibling ids that differ only by case (`c`/`C`). Without a fix the harness silently resolves the **wrong** sibling for such fixtures — a `RemoveChild{child_d:"C"}` deletes the `c` node — which can both false-positive and (worse) **mask a real VDOM regression**.

Real browsers match custom-attribute values case-sensitively (`dj-id` is not in the HTML spec's ASCII-case-insensitive attribute set), so **production is unaffected** — this is purely test-harness fidelity.

## Change

- `tests/js/_significant_tree.js` — `createHarnessDom` installs a case-sensitive `[dj-id="…"]` resolution shim. It **only** corrects nwsapi's case-collision (re-resolves among the native candidate set when nwsapi returns a case-folded-but-not-exact match) and delegates every other selector untouched, so escaped/special-char selectors and exact matches are unaffected. The client only ever uses single `[dj-id="X"]` tokens via `querySelector`, so this is the complete surface.
- `tests/js/harness_djid_case_sensitivity.test.js` — new regression test asserting both the raw `querySelector` resolution and the real-client `applyPatches`/`RemoveChild` path pick the correct case-sibling. **RED** before the shim (nwsapi returns the `c` node for `[dj-id="C"]`), **GREEN** after.

## Verification

- New test: RED before shim → GREEN after (non-tautological; gate-off proven).
- Full JS suite: **1748 tests pass, no regressions** (the shim is scoped to `createHarnessDom`'s own jsdom window).
- No production code touched; no CHANGELOG entry (test-infra only).

## Scope

Fixes exactly the harness-fidelity defect. #1977 stays **open** pending the reporter's bit-exact browser `diff_html(old,new)` capture (the real, un-reproduced-in-harness browser symptom). Two adjacent findings noted for follow-up, deliberately **not** fixed here: the `#1678` differential fixture is stale on main (a fresh regen drops its `MoveSubtree` patches), and `scripts/gen_vdom_diff_fixtures.py`'s docstring claims a pre-commit regen hook that doesn't exist.

### DIFF
--- a/tests/js/_significant_tree.js
+++ b/tests/js/_significant_tree.js
@@ -35,6 +35,8 @@ export function createHarnessDom(initialHtml) {
         window.CSS.escape = (s) => String(s).replace(/([^\w-])/g, '\\$1');
     }
 
+    installCaseSensitiveDjIdResolution(window);
+
     const logs = [];
     window.console = {
         log: (...args) => logs.push(['log', args.join(' ')]),
@@ -46,6 +48,62 @@ export function createHarnessDom(initialHtml) {
 
     window.eval(clientCode);
     return { dom, window, logs };
+}
+
+/**
+ * Make `[dj-id="…"]` selector resolution CASE-SENSITIVE in the jsdom window, so
+ * the harness matches real-browser semantics (#1977 follow-up).
+ *
+ * djust dj-ids are base62 (`0-9a-zA-Z`) from a monotonic counter, so any render
+ * past ~38 nodes emits sibling ids that differ only by case (`c` vs `C`). The
+ * client resolves structural ops with `node.querySelector(':scope >
+ * [dj-id="C"]')` (12-vdom-patch.js). Real browsers match custom-attribute VALUES
+ * case-sensitively (`dj-id` is not in the HTML spec's ASCII-case-insensitive
+ * attribute set), so `[dj-id="C"]` never matches `dj-id="c"`. jsdom's selector
+ * engine (nwsapi) matches them case-INSENSITIVELY, so without this shim the
+ * harness resolves the wrong sibling for any fixture that crosses ~38 nodes
+ * where a structural op targets a case-colliding sibling — silently
+ * false-positing AND (worse) masking real VDOM regressions.
+ *
+ * The shim wraps `querySelector` on Element/Document/DocumentFragment and only
+ * INTERVENES when nwsapi returns a dj-id that case-folds to the requested value
+ * but isn't an exact match — then it re-resolves case-sensitively among the
+ * native (case-insensitive) candidate set. Every other selector, and every
+ * exact match, delegates to native untouched (so escaped/special-char selectors
+ * are unaffected). The client only ever uses single `[dj-id="X"]` tokens via
+ * `querySelector`, so this is the complete surface.
+ */
+function installCaseSensitiveDjIdResolution(window) {
+    const DJID = /\[dj-id="([^"]*)"\]/;
+    const protos = [
+        window.Element && window.Element.prototype,
+        window.Document && window.Document.prototype,
+        window.DocumentFragment && window.DocumentFragment.prototype,
+    ].filter(Boolean);
+
+    for (const proto of protos) {
+        const nativeQS = proto.querySelector;
+        const nativeQSA = proto.querySelectorAll;
+        proto.querySelector = function (selector) {
+            const result = nativeQS.call(this, selector);
+            const m = typeof selector === 'string' && selector.match(DJID);
+            if (!result || !m) return result;
+            const want = m[1];
+            const got = result.getAttribute('dj-id');
+            if (got === want) return result; // exact match — nothing to correct
+            // Only correct nwsapi's case-insensitivity: same value up to case,
+            // but not an exact match. Re-resolve case-sensitively among the
+            // native candidate set (a superset of the exact-case matches).
+            if (got != null && got.toLowerCase() === want.toLowerCase()) {
+                const all = nativeQSA.call(this, selector);
+                for (let i = 0; i < all.length; i++) {
+                    if (all[i].getAttribute('dj-id') === want) return all[i];
+                }
+                return null; // no exact-case sibling — the id is genuinely absent
+            }
+            return result; // difference is not a case issue — leave native as-is
+        };
+    }
 }
 
 /**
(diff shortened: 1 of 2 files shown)


johnrtipton (commented):
APPROVE — test-infra fidelity fix, correct and empirically validated.

**Shim correctness** (`_significant_tree.js:87-105`): intervenes ONLY on nwsapi's case-collision — exact match fast-paths at L93, null/non-string/non-dj-id results delegate untouched at L90, and a non-case difference falls through to native at L104. The `got.toLowerCase() === want.toLowerCase()` guard (L97) plus re-resolution over the native candidate set (L98-100) means it can't return a wrong-case node or throw (all operands are Elements/strings behind guards). It correctly returns `null` (L102) when no exact-case sibling exists — real-browser faithful, not a case-folded fallback.

**Empirical verification** (read-only): native jsdom returns `"lower"` for `:scope > [dj-id="C"]` and `qsa(...).length === 2`, so the shim's candidate-filter picks `"C"`→`"upper"`. Gate-off is genuinely RED (`[dj-id="C"]`→`"lower"` ≠ `"upper"`), so both tests are non-tautological; test 2 drives the REAL client path (`applyPatches`→RemoveChild at `12-vdom-patch.js:1939`), not a proxy. `harness_djid_case_sensitivity` + `vdom_client_faithful_diff`: 3/3 pass. Per-JSDOM-window prototype patch → no vitest-global leak.

**Surface**: grep confirms dj-id VALUE selectors are `querySelector`-only (single, double-quoted) — never `querySelectorAll`/`closest`/`matches` — so the coverage is complete today.

**Minor concern**: that completeness is an IMPLICIT invariant. `querySelectorAll('[dj-id="X"]')` (or a multi-token descendant selector — the `DJID` regex only captures the first token) would silently revert to case-insensitive with no failure signal. Consider a one-line guard-test pinning "no value-based qsa/closest for dj-id" so a future client change trips a test, not a masked regression.

Non-blocking. LGTM.

johnrtipton:
Addressed the review's non-blocking concern in d4bfca7b: added a source-grep guard-test (`client resolves dj-id VALUE selectors only via single-token querySelector`) that pins the surface invariant the shim relies on. It fails loudly if the client ever resolves a dj-id value via `querySelectorAll`/`closest`/`matches` or a multi-token descendant selector (verified the regex catches each of those shapes and ignores current single-token `querySelector` usage). Full JS suite green (3/3 for this file; 1748 total).

johnrtipton:
# Retrospective — PR #1978

**Task origin:** `/pipeline-run` on #1977 ("VDOM: diff_html misroutes SetText patches on `{% for %}` list-shrink").
**Delivered:** a **test-harness fidelity fix** (case-sensitive `[dj-id]` resolution in the client-faithful VDOM harness) — the one verified defect that survived the investigation, after #1977's differ-misroute claim was retracted.

**Quality: 5/5** (process discipline). The investigation reached the correct conclusion — *there is no differ bug* — independently of, and in agreement with, the maintainer's own retraction comment posted mid-investigation. It then salvaged the real finding and shipped it with a proper RED→GREEN test and an independent APPROVE review.

## What went well

- **Reproduce-first + reproduction-fidelity caught a false positive twice.** My initial Rust-level reproducer flagged `SetText` paths as invalid against the *pristine* old tree — but the real client applies structural patches (phases −2…3) before `SetText` (phase 4), so the shift is repaired first. Simulating the phased applier proved the "repro" was a false positive *before* any fix was written. This is exactly the CLAUDE.md "distrust the reproduction harness until it exercises the real path" rule paying off.
- **Didn't rubber-stamp the subagent.** The trigger-search subagent concluded its two REDs were a jsdom case-insensitivity artifact. I re-verified — and my *first* probe (plain `querySelector` on a `<div>`) came back case-**sensitive**, contradicting it. Rather than accept either result, I re-ran with the *exact* client selector (`:scope > [dj-id="C"]`) and reproduced the case-**insensitive** behavior. The subagent was right, but only the exact-selector probe settled it. Verifying the claim (not the conclusion) is what surfaced the real, shippable bug.
- **Scope discipline.** Fixed exactly the harness defect; noted two adjacent findings (the #1678 differential fixture is stale on main — a fresh regen drops its `MoveSubtree` patches; and the generator docstring claims a nonexistent pre-commit regen hook) in the PR for follow-up rather than expanding scope.
- **Review finding addressed in one pass** (the surface-invariant guard-test), per the "fix ALL findings in one push" norm.

## Lessons learned (specific)

- **A "misrouted patch" against the pristine old tree is not a bug when the client applies structural-first.** The client's phase ordering (`_sortPatches`, `12-vdom-patch.js`) means node-targeting `SetText` paths are *new-tree*-relative by design; validity must be judged against the post-structural DOM, not the pre-patch tree. Any future VDOM-diff "misroute" report should be validated through the phased client applier, not a pristine-tree invariant.
- **jsdom is not real-browser-faithful for custom-attribute-value case-sensitivity via `:scope >`.** Combined with base62 dj-ids (`c`/`C` collisions past ~38 nodes), this made the *client-faithful harness itself* unfaithful — able to false-positive and mask real VDOM regressions. Test infrastructure that runs real client code against jsdom must be audited for these engine gaps.

## Relation to prior tasks

Directly continues the #1678/#1408 client-faithful-harness lineage (`[[project_1678_kanban_parity_drift]]` — "build a client-faithful differential harness for VDOM-patch bugs"): this PR *hardens that very harness*. Reinforces the multi-reopen-class canon ("bit-exact runnable repro before root-cause confirmed; theory-testing is insufficient") — the differ fix was correctly *not* shipped for lack of a faithful RED, and #1977 stays open for the reporter's browser capture.

**Quality: 5** — RETRO_COMPLETE

PR #1867, opened 2026-06-22
docs(security): SECURE_DEFAULTS.md pattern catalog + PR-checklist secure-defaults + audit cadence (#1856)

johnrtipton:
Closes #1856 (v1.0.8-1 group 7, task T3 — DOCS-focused). Pure docs.

### 1. `docs/SECURE_DEFAULTS.md` — 4-pattern catalog
Documents the four proven secure-by-default patterns, each with threat, canonical `file:symbol` (grep-verified), and the copy-shape for a new feature:

- **(a) Denylist serialization** — `python/djust/serialization.py`: `_ALWAYS_EXCLUDED_FIELDS` floor (`password`/`is_superuser`/`is_staff`, line 52), `_resolve_sensitive_fields()` (line 59, unions floor + `DJUST_SENSITIVE_FIELDS`), `_serialize_model_safely` (method, line 222) with per-model `djust_exclude_fields` (line 330) / `djust_serializable_fields` allowlist (line 349).
- **(b) HMAC signed snapshots** — `python/djust/security/state_snapshot.py`: `sign_snapshot` (line 101) / `unsign_snapshot` (line 127) — TimestampSigner (`SNAPSHOT_SALT`, line 61), slug+session binding, TTL (`DEFAULT_MAX_AGE=3600`, line 66), fail-closed `None` on tamper/expiry/cross-view/cross-session. **Private-attr signing boundary documented**: `_capture_snapshot_state` (`live_view.py:757`) excludes `_*` attrs from the signed blob; `_restore_private_state` (`live_view.py:740`) restores them via plain `setattr` UNSIGNED from the server-side session (restore call at `websocket.py:2390`) — so don't store auth/ownership/PII in `_*` expecting integrity. (Brief said `safe_setattr(..., allow_private=True)`; grep showed the actual restore is plain `setattr` — documented the verified mechanism.)
- **(c) Fail-closed precedence gate** — `observability/views.py:_gate` (line 90, DEBUG→localhost→404) and `api/openapi.py:_openapi_gate` (line 189, DEBUG→`DJUST_API_OPENAPI_PUBLIC`→authenticated→non-disclosing-404); plus `auth/core.py:run_pre_mount_auth` (line 396, #1853) as the mount-auth single-source.
[...]

PR #1980, opened 2026-07-02
test(vdom): refresh stale #1678 diff fixture + add a CI freshness gate (#1979)

### LINKED ISSUE #1979: test-infra: #1678 differential VDOM fixture is stale on main + generator's claimed pre-commit regen hook doesn't exist
Two related test-infra findings surfaced during the #1977 investigation (see PR #1978, which fixed a separate jsdom fidelity bug in the same harness).

## 1. The `#1678` differential fixture is stale on `main`

`tests/js/fixtures/vdom_diff_kanban_tabs_1678.json` was committed from an older `diff_html` version. Regenerating it on current `main` (`.venv/bin/python scripts/gen_vdom_diff_fixtures.py`) **drops 49 lines, including the two `MoveSubtree` patches** the #1678 scenario is specifically meant to exercise:

```
$ git diff tests/js/fixtures/vdom_diff_kanban_tabs_1678.json   # after a fresh regen
-        { "type": "MoveSubtree", "id": "if-4017a4e7-4", "path": [], "d": "0", "index": 2 },
-        { "type": "MoveSubtree", "id": "if-4017a4e7-8", "path": [], "d": "0", "index": 2 },
... (49 deletions)
```

The committed (stale) fixture still *passes* the harness (its stored patches are internally consistent with its stored `expected_html`), so the guard is green — but it's pinning **old** patch output, not what `main`'s differ produces today. That means the #1678 `MoveSubtree`-ordering regression guard is no longer guarding the `MoveSubtree` path.

**Needs triage:** is the differ change that dropped the `MoveSubtree` patches (a) a legitimate improvement — the kanban tab switch genuinely no longer needs `MoveSubtree` — or (b) a regression that removed the very behavior #1678 guards? Either way the fixture should be regenerated and re-reviewed so the guard reflects current behavior. (The `if-<hash>` marker-id prefix also changed, consistent with a template/differ change since the last regen.)

## 2. The generator claims a pre-commit regen hook that doesn't exist

`scripts/gen_vdom_diff_fixtures.py`'s module docstring states:

> Regeneration is idempotent (deterministic dj-ids); **a pre-commit hook re-runs this and `git add`s the result so fixtures can't go stale.**

There is no such hook in `.pre-commit-config.yaml` (no reference to `gen_vdom_diff_fixtures`). This is why finding #1 went unnoticed. Options: either add the hook the docstring promises (regenerate on changes to the differ / fixture views + `git add`), or correct the docstring and add a CI check that fails if a fresh regen produces a diff.

## Context

Found while investigating #1977 (differ-misroute claim, since retracted). PR #1978 hardened the harness against a jsdom `[dj-id]` case-insensitivity artifact; these two findings are the remaining test-infra debt in the same area, left out of #1978 for scope discipline.

johnrtipton:
Fixes #1979.

Two test-infra findings from the #1977/#1978 investigation.

## 1. The #1678 differential fixture was stale on `main`

A fresh regen of `tests/js/fixtures/vdom_diff_kanban_tabs_1678.json` **drops 7 `MoveSubtree` patches** from step 1 (leaving `RemoveChild` + `InsertSubtree`).

**Triage — this is a *correct* consequence of #1826, not a regression.** #1826 (`bd1c1f53`, "MoveSubtree fires on relative position, not absolute offset") landed *after* the fixture's last commit (PR #1682, 2026-06-01). The 8 tab-block dj-if boundaries are adjacent and keep their relative order across a tab switch → no genuine reposition → the old absolute-offset MoveSubtrees were spurious and #1826 correctly eliminated them. Verified:
- The (now #1978-hardened, case-sensitive) client-faithful harness **passes** with the fresh patches — the client reaches a fresh server render.
- The fixture still exercises the nested-conditional `InsertSubtree` + 9 dj-if markers that were the actual #1678 kanban `html_recovery` bug, so the guard stays meaningful.

## 2. The generator claimed a pre-commit regen hook that didn't exist

`scripts/gen_vdom_diff_fixtures.py`'s docstring said "a pre-commit hook re-runs this ... so fixtures can't go stale" — there was no such hook (which is why #1 went unnoticed). Replaced the claim with the actual mechanism + real enforcement:

- Corrected the docstring; fixed stale `MoveSubtree`-referencing text in `gen_kanban_tabs_1678` and the scenario `description`.
- `make gen-vdom-fixtures` convenience target.
- **CI freshness gate** in the `python-tests` job (`.github/workflows/test.yml`): re-runs the deterministic generator and fails on any diff under `tests/js/fixtures/`. That job already uses `demo_project.settings`, so a diff is a genuine stale-fixture signal, not env drift.

## Verification

- Generator is **deterministic** (two regens byte-identical) → the gate can't flake.
- **Empirical canary** (per the tooling-PR rule): hand-dirtied a fixture → gate fails; fresh → gate passes (exit 0).
- Harness green with the refreshed fixture; `check-yaml` validated the workflow.
- Test-infra only — no production code, no CHANGELOG.

## Out of scope (follow-up)

Step 2 (card move) captures **0 patches** — the card moves in state (`todo:[c1,c2]→[c2]`, `done:[]→[c1]`) but the render is identical. **Pre-existing** (the committed fixture had step 2 = `{}` too), likely in-place-`columns`-mutation change-detection. Not fixed here; a follow-up issue tracks it.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -141,6 +141,19 @@ jobs:
       - name: Install test tools
         run: uv pip install pytest-xdist mypy
 
+      # Freshness gate for the client-faithful VDOM diff fixtures (#1979): the
+      # committed JSON must equal a fresh regen, or the harness pins stale patch
+      # output (as it did before #1979). Deterministic generator + this job
+      # already uses demo_project.settings, so a diff here is a genuine
+      # stale-fixture signal, not env drift.
+      - name: VDOM diff fixtures are fresh (#1979)
+        run: |
+          PYTHONPATH=. .venv/bin/python scripts/gen_vdom_diff_fixtures.py
+          if ! git diff --exit-code -- tests/js/fixtures/; then
+            echo "::error::VDOM diff fixtures are stale — run 'make gen-vdom-fixtures' and commit tests/js/fixtures/."
+            exit 1
+          fi
+
       - name: Run Python tests (parallel with pytest-xdist)
         run: |
           PYTHONPATH=. .venv/bin/python -m pytest tests/ python/tests/ -v -n auto
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -149,8 +149,14 @@ jobs:
       - name: VDOM diff fixtures are fresh (#1979)
         run: |
           PYTHONPATH=. .venv/bin/python scripts/gen_vdom_diff_fixtures.py
-          if ! git diff --exit-code -- tests/js/fixtures/; then
-            echo "::error::VDOM diff fixtures are stale — run 'make gen-vdom-fixtures' and commit tests/js/fixtures/."
+          # --porcelain (not `git diff --exit-code`) so a NEW scenario's
+          # never-`git add`ed output file is caught too, not only modified
+          # tracked fixtures (#1980 review).
+          DIRTY=$(git status --porcelain -- tests/js/fixtures/)
+          if [ -n "$DIRTY" ]; then
+            echo "::error::VDOM diff fixtures are stale or uncommitted — run 'make gen-vdom-fixtures' and commit tests/js/fixtures/."
+            echo "$DIRTY"
+            git diff -- tests/js/fixtures/
             exit 1
           fi
 
--- a/Makefile
+++ b/Makefile
@@ -162,6 +162,10 @@ check-lockfile-versions: ## Verify Cargo.lock/uv.lock self-entries match manifes
 check-bundle-init-order: ## Static check: declared-late/used-early let/const across bundle concat (closes #1372)
 	@node scripts/check-bundle-init-order.mjs
 
+.PHONY: gen-vdom-fixtures
+gen-vdom-fixtures: ## Regenerate client-faithful VDOM diff fixtures (CI gates freshness — closes #1979)
+	@$(PYTHON) scripts/gen_vdom_diff_fixtures.py
+
 .PHONY: test
 test: ## Run all tests (Python + JavaScript + Rust) in parallel
 	@echo "$(GREEN)Running all tests in parallel...$(NC)"
--- a/scripts/gen_vdom_diff_fixtures.py
+++ b/scripts/gen_vdom_diff_fixtures.py
@@ -26,9 +26,11 @@ pre-commit hook kept it fresh and none existed).
 Run:  make gen-vdom-fixtures      # or: .venv/bin/python scripts/gen_vdom_diff_fixtures.py
 """
 
+import atexit
 import json
 import os
 import sys
+import tempfile
 from pathlib import Path
 
 REPO = Path(__file__).resolve().parent.parent
@@ -39,6 +41,23 @@ os.environ.setdefault("DJANGO_SETTINGS_MODULE", "demo_project.settings")
 import django  # noqa: E402
 
 django.setup()
+
+# The LiveViewTestClient mount/event path persists to the session
+# (`django_session`), so the generator needs a migrated schema. Point the
+# default DB at an isolated throwaway file and migrate it HERE, so the generator
+# is self-contained: it works on a fresh CI runner (which has no migrated
+# db.sqlite3 — the #1980 CI failure) AND on a fresh dev checkout, WITHOUT
+# touching the developer's real demo db.sqlite3. `run_syncdb` is fine for this
+# test-only tool (not a deploy path, so the #1637 caveat doesn't apply).
+from django.conf import settings  # noqa: E402
+from django.core.management import call_command  # noqa: E402
+from django.db import connections  # noqa: E402
+
+_TMP_DB = tempfile.mktemp(prefix="djust_vdom_fixtures_", suffix=".sqlite3")
+settings.DATABASES["default"]["NAME"] = _TMP_DB
+connections.close_all()  # drop any connection bound to the previous NAME
+atexit.register(lambda: os.path.exists(_TMP_DB) and os.remove(_TMP_DB))
+call_command("migrate", run_syncdb=True, verbosity=0, interactive=False)
 
 # Register djust component tag handlers ({% kanban_board %}, {% empty_state %})
 # with the Rust engine — the #1678 fixture's active tab body is a nested
--- a/scripts/gen_vdom_diff_fixtures.py
+++ b/scripts/gen_vdom_diff_fixtures.py
@@ -29,6 +29,7 @@ Run:  make gen-vdom-fixtures      # or: .venv/bin/python scripts/gen_vdom_diff_f
 import atexit
 import json
 import os
+import shutil
 import sys
 import tempfile
 from pathlib import Path
@@ -53,10 +54,11 @@ from django.conf import settings  # noqa: E402
 from django.core.management import call_command  # noqa: E402
 from django.db import connections  # noqa: E402
 
-_TMP_DB = tempfile.mktemp(prefix="djust_vdom_fixtures_", suffix=".sqlite3")
-settings.DATABASES["default"]["NAME"] = _TMP_DB
+# mkdtemp gives a secure 0700 dir we own (no filename-generation race), removed on exit.
+_TMP_DB_DIR = tempfile.mkdtemp(prefix="djust_vdom_fixtures_")
+settings.DATABASES["default"]["NAME"] = os.path.join(_TMP_DB_DIR, "fixtures.sqlite3")
 connections.close_all()  # drop any connection bound to the previous NAME
-atexit.register(lambda: os.path.exists(_TMP_DB) and os.remove(_TMP_DB))
+atexit.register(lambda: shutil.rmtree(_TMP_DB_DIR, ignore_errors=True))
 call_command("migrate", run_syncdb=True, verbosity=0, interactive=False)
 
 # Register djust component tag handlers ({% kanban_board %}, {% empty_state %})
--- a/scripts/gen_vdom_diff_fixtures.py
+++ b/scripts/gen_vdom_diff_fixtures.py
@@ -75,8 +75,8 @@ def gen_kanban_tabs_1678():
     correctly gone (regenerated for freshness in #1979).
 
     NOTE: step 2 (card move) currently captures 0 patches — an in-place mutation
-    of `columns` isn't reflected in the render; tracked separately (see #1979
-    follow-up). Kept so the freshness gate pins that state until it's fixed."""
+    of `columns` isn't reflected in the render; tracked separately at #1981.
+    Kept so the freshness gate pins that state until it's fixed."""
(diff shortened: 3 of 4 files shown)


johnrtipton:
**Adversarial code review — test-infra PR**

**Verdict: APPROVE** (2 🟡 nits, 1 low-severity flake note; nothing blocking)

Clean, well-triaged test-infra PR. I verified every load-bearing claim against git history and the workflow file rather than trusting the description.

### What I verified (all confirmed)

- **Triage soundness (#1826).** `bd1c1f53` (`fix(vdom): dj-if MoveSubtree fires on relative position, not absolute offset (#1826) (#1828)`) landed 2026-06-17, **after** the fixture's last commit `d5ce4dd9` (`#1682`, 2026-06-01). The #1826 commit message describes exactly this fixture's step-1 shape (filling an earlier empty dj-if body shifts later boundaries' *absolute* index but not their *relative* order → old absolute-offset MoveSubtrees were spurious). The maintainer's test `crates/djust_vdom/tests/test_dj_if_loop_spurious_move_1826.rs::djif_loop_fill_emits_no_spurious_move_for_later_boundaries` asserts precisely "no MoveSubtree when no boundary repositions relative to siblings." The "spurious MoveSubtree correctly eliminated" story is credible and not masking a regression.
- **Gate actually gates the merge (#1713/#1450 class).** New step `.github/workflows/test.yml:149` (`VDOM diff fixtures are fresh (#1979)`) is in the `python-tests` job with **no `continue-on-error`**; `python-tests` is in `test-summary.needs` (`:583`) **and** in the success AND-condition (`:605` `[ "${{ needs.python-tests.result }}" == "success" ]`). A stale fixture fails the step → fails `python-tests` → fails the aggregate gate. Confirmed gating.
- **No swallowing / no env gap.** GHA `run:` uses `bash -eo pipefail`, so a generator crash fails the step. The generator self-sets `DJANGO_SETTINGS_MODULE` via `os.environ.setdefault(...)` (`scripts/gen_vdom_diff_fixtures.py:37`) — it does **not** rely on the pytest ini, which wouldn't apply to a bare `python script.py` invocation anyway. Good catch by the author.
- **Determinism.** Dict key order is fixed by construction; `json.dumps(..., indent=2)` (`:128`); patches come from the deterministic Rust differ; the `if-4017a4e7-N` prefix is `DefaultHasher` of the template *source* (`parser.rs:302-306`), constructed with fixed seeds unlike `RandomState` (documented `parser.rs:347-348`). Deterministic within a toolchain.
- **Still guards #1678.** Refreshed fixture step 1 = `{RemoveChild:1, InsertSubtree:1}` (the nested-conditional InsertSubtree kanban body), 11 dj-if markers in `initial_html`, kanban body present in `expected_html`. Guard remains meaningful.
- **Citations.** `bd1c1f53`→#1826/#1828 ✓, `d5ce4dd9`→#1682 ✓, dates ✓.

### 🟡 should-fix / nits

1. **Wrong follow-up issue cited.** `scripts/gen_vdom_diff_fixtures.py:78` says step 2's 0-patch behavior is "tracked separately (see **#1979** follow-up)" — but #1979 is the issue *this PR closes*. The actual follow-up is **#1981** ("bug: KanbanTabsView card move mutates state but renders 0 patches"). The PR body likewise says "a follow-up issue tracks it" without the number. Please cite #1981 in the docstring (and ideally the PR body). Citation-discipline (#1197).
2. **Gate misses untracked new fixtures.** `.github/workflows/test.yml:152` `git diff --exit-code -- tests/js/fixtures/` only detects modifications to **tracked** files. If a future `SCENARIOS` entry writes a *new* fixture file and the author forgets to `git add` it, the regen creates an untracked file that `git diff` won't flag → the gate passes green on a genuinely-missing fixture. Fine for today's single scenario, but the gate advertises general freshness enforcement. Consider `test -z "$(git status --porcelain tests/js/fixtures/)"` or `git add -A tests/js/fixtures && git diff --cached --exit-code`.

### Low-severity flake note (not blocking)
The "deterministic → can't flake" claim holds *within a fixed Rust toolchain*. The marker-id prefix derives from `std`'s `DefaultHasher`, whose output std explicitly does **not** guarantee stable across Rust releases. A stable-Rust SipHash change would flip `4017a4e7` and false-fail the gate on every PR until someone regens. Self-healing (the `::error::` message tells you to run `make gen-vdom-fixtures`), so acceptable — just flagging the vector the determinism claim glosses over.

### Edge cases checked
- Generator exit-code propagation under `bash -e` ✓ (crash fails the step, not swallowed)
- `git diff` semantics in CI post-checkout (committed==HEAD; regen overwrites tracked file → unstaged diff caught) ✓
- Step 2 = 0 patches is pre-existing (committed fixture already had `{}`); gate now pins it at 0, which is by-design — the #1981 fix landing will fail the gate and force a conscious regen. Reasonable scope call.

### Concern / question
The step-2 fixture step is now a **vacuous** guard (0 patches; `expected_html` trivially equals current). It contributes nothing to the #1678 differential coverage yet the freshness gate pins it. That's acceptable *only* because #1981 is filed — please make sure the docstring points at #1981 (nit #1) so a future reader who hits the pinned 0-patch state can find the tracking issue. Otherwise this looks like an intentional, well-documented follow-up boundary.

johnrtipton:
## Stage 14 Retrospective — PR #1980 (#1979)

**quality:** test-infra fix, APPROVE on first review with 2 🟡 (both addressed pre-merge). No production code touched.

### What shipped
1. Refreshed the stale `#1678` client-faithful VDOM diff fixture (`tests/js/fixtures/vdom_diff_kanban_tabs_1678.json`) — a fresh regen dropped 7 `MoveSubtree` patches from step 1.
2. Root-caused the drop to **#1826** (`bd1c1f53`, "MoveSubtree fires on relative position, not absolute offset"), which landed *after* the fixture's last commit (#1682, `d5ce4dd9`) — a **correct** elimination of spurious moves, not a regression. Verified via the #1826 Rust test + the (post-#1978, case-sensitive) client-faithful harness passing with the fresh patches.
3. Replaced the generator docstring's **false pre-commit-hook claim** with a real **CI freshness gate** (`python-tests` job re-runs the deterministic generator, fails on any diff/untracked file under `tests/js/fixtures/`), a `make gen-vdom-fixtures` target, and accurate docstrings.
4. Filed **#1981** for the pre-existing vacuous step-2 (card move → 0 patches; in-place `columns` mutation not change-detected).

### Learnings

- **A new CI step that drives the framework hits DB/env state the dev machine masks — reproduce the CI failure locally before fixing (#1534 + reproduce-first).** The freshness gate passed locally but failed on the runner with `sqlite3.OperationalError: no such table: django_session`: the generator drives `LiveViewTestClient`, whose mount/event path persists to the session, so it needs a migrated schema — which my dev `db.sqlite3` already had and a fresh runner does not. Textbook #1534 (a CI step exercising an env the dev machine can't mirror needs ≥1 runner-only iteration). Per the reproduce-first canon, I reproduced it locally against a fresh temp DB (`no such table: django_session`) and confirmed the fix (self-migrate) before pushing — no blind CI-iteration guessing. The durable fix made the *generator* self-contained (isolated temp DB + `migrate` in-process, atexit-cleaned) rather than patching only the CI YAML, so a fresh dev checkout running `make gen-vdom-fixtures` works too. Verified the fixture is byte-identical against a pre-migrated vs a fresh temp DB (render doesn't depend on DB/session *content*, only needs the table for the *write*), so the gate is stable.

- **A doc that claims an enforcement MECHANISM exists must be verified against the mechanism — else the "guard" silently rots.** The `#1678` fixture (PR #1682) shipped a docstring asserting "a pre-commit hook re-runs this ... so fixtures can't go stale" while no such hook existed. That false claim is *exactly* why the fixture went stale unnoticed for ~1 month across #1826/#1832. This is the same failure class as the v1.0.8-1 "decorative anti-drift artifact" rule (#1859/#1860 — a pin/guard that isn't load-bearing) and the #1867 "falsification-test prose invariants" rule: a claim of the form "X keeps this fresh/safe" needs X to exist and be wired. Generalizes to a Stage-7/Stage-11 grep target: for any comment/docstring asserting an automated guard, confirm the guard is real and reachable.

- **A freshness/coverage gate must catch NEW artifacts, not just modified ones** (#1980 review 🟡 #2). The first-pass gate used `git diff --exit-code`, which only sees *modified tracked* files — a future new scenario whose output is never `git add`ed would slip the gate silently. Switched to `git status --porcelain` (catches `??` untracked too). Same shape as the v1.0.0rc4 "a coverage suite must enumerate EVERY variant of the surface" rule (#1543) — single-variant coverage of a multi-variant surface (here: modified-only vs modified+new) is false confidence.

- **The #1977→#1978→#1979/#1980 chain shows the value of shipping the harness-fidelity fix first.** #1978 made the harness dj-id resolution case-sensitive; only *because* of that could the refreshed fixture be trusted (a >38-node render could otherwise mask a real regression behind a jsdom case-fold). Sequencing the fidelity fix before the fixture refresh was load-bearing.

- **Don't ship a known-insecure stdlib pattern even in a dev script.** The self-migrate fix first used `tempfile.mktemp()` (bandit B306 / TOCTOU race). CI didn't gate it (bandit is scoped to `python/`, not `scripts/`; the actual CodeQL `Analyze Python` passed), but a security-focused framework shouldn't model `mktemp` anywhere — switched to `mkdtemp()` (secure 0700 dir, atexit-cleaned) proactively. Reinforces that self-review must apply the repo's security bar to *all* shipped code, not only the CI-scoped paths.

- **Verified the CodeQL "2 configurations not found" failure is systemic, not mine, before proceeding.** It's a diff-coverage artifact from `pre-release-security-audit.yml` CodeQL configs that don't run on PRs (same `neutral`/`failure` result on #1978 and #1973); the real `Analyze Python`/`Analyze JavaScript` analyses passed and CodeQL is not a required check (the `test-summary` AND-gate is). Confirmed via branch-protection + sibling-PR check rather than assuming.

- The original `#1678` fixture PR (#1682) should have either added the pre-commit hook it documented, or not claimed one. A Stage-7 self-review grep — "does every claimed automated guard exist?" — would have caught it at authoring time instead of a month later. Worth adding to `docs/PULL_REQUEST_CHECKLIST.md` as a one-line check (candidate for the next canon batch).
- Low-severity noted (non-blocking, self-healing): the Rust `DefaultHasher` used for dj-id seeding is not guaranteed byte-stable across Rust toolchain versions, so a stable-Rust SipHash change could false-fail the gate until a regen. CI pins the toolchain, so this is contained; if it ever bites, the fix is a one-command `make gen-vdom-fixtures` + commit.

### Relation to prior tasks
Directly closes the test-infra loop opened by the #1977 investigation (differ retraction) and #1978 (harness case-sensitivity). Reinforces #1646 (parallel-path/consumer drift) at the docs-vs-mechanism axis and #1543 (coverage-variant enumeration) at the gate-completeness axis.

`RETRO_COMPLETE`

PR #1986, opened 2026-07-02
feat(templates): Django-parity callable auto-call in the sidecar walk (#1985, ADR-024)

### LINKED ISSUE #1985: feat: template callable auto-call — Django parity in the sidecar resolution walk (ADR-024)
Implement **ADR-024** (`docs/adr/024-template-callable-auto-call.md`, merged in #1984).

## Symptom (DJUST_LESSONS gotcha #7, hit in downstream production builds)
- `{{ request.user.get_full_name }}` renders `<bound method AbstractUser.get_full_name of <User: jordan>>`
- `{{ workspace.memberships.count }}` renders empty

Django auto-calls callables during variable resolution; djust's Rust engine does not — a **silent** divergence (renders something, just the wrong thing).

## Root cause (mapped, citations grep-verified)
#1646-class parallel-path drift: the **eager** serialization path already auto-calls (`optimization/codegen.py:198-204`, `:315-321` — `get_*`/`all`/`count`/`exists`; `serialization.py:444`/`:514` — explicit `get_*` + properties), but the **lazy sidecar** getattr walk — `Context::resolve` at `crates/djust_core/src/context.rs:235-263`, which serves request-scoped objects (`user`) and reverse relations (`memberships`) — never checks `is_callable()`. The un-called bound method falls to the `FromPyObject` `str()` catch-all (`crates/djust_core/src/lib.rs:235`).

## Scope (per the ADR's Decision — see the semantics table there)
1. **Auto-call in `Context::resolve`**, Django's exact `Variable._resolve_lookup` semantics: per-segment + final `call0()`; `do_not_call_in_templates` → use as-is; `alters_data` → `Value::Null`, never call (**must ship in the same commit as the call** — data-destruction guard); `TypeError` → `inspect.signature(...).bind()` probe (bind fails → Null; bind succeeds → propagate); other exceptions propagate. Do NOT touch the shared `FromPyObject` catch-all (serves `update_state` ingestion).
2. **Guard unification**: add `alters_data`/`do_not_call_in_templates` checks to the eager auto-call sites (`codegen.py` ×2, `serialization.py:_add_safe_model_methods`) — N sites, N tests (#1104).
3. **Observability**: debug-only one-shot warning per (view class, dotted path) when an auto-call's `__self__` is a `Manager`/`QuerySet` (classes imported once, cached) — "runs on EVERY re-render; consider precomputing in get_context_data()".
4. **Kill-switch**: `LIVEVIEW_CONFIG["template_auto_call"]`, default `True`; `False` restores pre-ADR sidecar behavior.
5. **Docs + CHANGELOG**: templating-guide divergence note replaced; `docs/SECURE_DEFAULTS.md` gains the `alters_data` pattern row.

## Testing (doc-claim-verbatim, #1046)
One asserting test per semantics-table row; the two symptoms as regressions through the real render path; mid-path call (`{{ obj.get_settings.theme }}`); `alters_data` refusal proven by side-effect sentinel (method NOT executed); kill-switch gate-off (#1468); Rust unit tests in `djust_core` + Python integration; eager-site guard tests per site.

Refs: ADR-024, #1984 (ADR PR), #1646 (drift class).

johnrtipton:
Fixes #1985. Implements **ADR-024** (`docs/adr/024-template-callable-auto-call.md`, merged in #1984).

## What

Django's template engine auto-calls callables during variable resolution; djust's Rust engine did not — a **silent** divergence:
- `{{ request.user.get_full_name }}` → `<bound method AbstractUser.get_full_name of <User: jordan>>`
- `{{ workspace.memberships.count }}` → empty

Both hit in downstream production builds (DJUST_LESSONS gotcha #7).

## How (per the ADR)

**Rust — the single fix point.** `Context::resolve`'s sidecar getattr walk (`crates/djust_core/src/context.rs`) now implements Django's exact `Variable._resolve_lookup` callable semantics at **every** segment (root, mid-path, final):

| Condition | Behavior |
|---|---|
| `do_not_call_in_templates` | used as-is (Model classes, `Choices` enums) |
| `alters_data` | **never called**, renders empty — the data-destruction guard, shipped in the same commit as the call |
| otherwise | `call0()` |
| `TypeError` from the call | `inspect.signature(...).bind()` probe: args-required → empty; internal `TypeError` → propagates |
| any other exception | propagates as a render error |

`resolve()` returns `Result<Option<Value>>` so call exceptions surface (lookup misses stay `Ok(None)`); all 3 call sites threaded. The shared `FromPyObject` catch-all is untouched (it serves `update_state` ingestion).

**Sidecar completeness (discovered during implementation).** The normalize loop replaces explicit-context models with dicts *before* the sidecar builds — so only request-scoped models ever reached the sidecar, and the ADR's headline case (`{{ workspace.memberships.count }}` with an explicit `workspace`) would have stayed empty. Raw models are now kept available to the sidecar as fallback: the eager dict wins every direct hit; the raw model serves only nested paths the dict lacks (reverse relations / managers / non-`get_` methods).

**Guard unification (ADR Decision 2, #1646).** The pre-existing eager auto-call sites — `optimization/codegen.py` (×2 generated-code sites) and `serialization.py::_add_safe_model_methods` — gained the same `alters_data`/`do_not_call_in_templates` guards.

**Observability rider.** Debug-only, one-shot-per-path warning when an auto-call is bound to a `Manager`/`QuerySet` — in a LiveView that's a DB query per re-render. Reads `settings.DEBUG` live (so `override_settings` works; no process-cached flag).

**Kill-switch.** `LIVEVIEW_CONFIG["template_auto_call"]` (default `True`) → `RustLiveView.set_template_auto_call`, mirroring the #1967 `loop_render_cache_enabled` plumbing exactly. `False` restores the pre-ADR no-call walk.

## Tests (doc-claim-verbatim, #1046)

16 cases in `python/djust/tests/test_template_auto_call_1985.py`:
- one test per ADR semantics-table row (side-effect **sentinels** prove `alters_data`/`do_not_call` callables are *not executed*, not just absent from output);
- both reported symptoms through the real render path with real ORM (`User.get_full_name` request-scoped; `user.groups.count` ManyRelatedManager via sidecar);
- mid-path call; internal-`TypeError`/`ValueError` propagation; args-required → empty;
- kill-switch behavioral gate-off (#1468) + config wiring test;
- eager-site guard tests per site (#1104) incl. compiled-codegen runtime proof;
- ORM-warning one-shot test.

`python/tests/test_codegen.py` pins updated to the guarded generated shape (deliberate, cited to ADR-024).

## Verification

- Full Python suite: **8685 passed / 0 failed** (all roots, `-n auto`)
- `make test-rust`: 40/40 suites green; clippy clean
- `mypy python/djust`: 826 files green; ruff clean
- Two-commit shape (impl `d9c16e96` / docs `820b6506`); Gates 1+2 verified

## Docs

`template-cheatsheet.md` (new "Callables are auto-called" section with the LiveView per-event cost note), `SECURE_DEFAULTS.md` pattern 5 (`alters_data` guards), ADR-024 status → shipped, CHANGELOG Added.

Refs: ADR-024, #1984, #1646 (drift class), #1967 (flag-plumbing pattern mirrored).


johnrtipton:
## Adversarial Code Review — PR #1986 (ADR-024 template callable auto-call)

**Verdict: REQUEST_CHANGES** — one confirmed 🔴 secure-by-default regression (serialization-floor bypass), plus 🟡 notes. The auto-call semantics themselves (`alters_data`/`do_not_call`/TypeError-probe/kill-switch) are correctly implemented and empirically verified; the problem is the *sidecar model retention* that rides alongside them.

All findings below were reproduced by rendering through the real `LiveViewTestClient` → `Context::resolve` sidecar walk (the exact divergence path), on the checked-out branch with the release build.

---

### 🔴 MUST-FIX — Raw-model sidecar retention bypasses the `_ALWAYS_EXCLUDED_FIELDS` serialization floor (#1868)

`docs/SECURE_DEFAULTS.md` Pattern 1 and #1868 guarantee the floor `_ALWAYS_EXCLUDED_FIELDS = {"password", "is_superuser", "is_staff"}` (`python/djust/serialization.py:56`) is **UNCONDITIONAL** — those fields never reach the client. This PR punches a hole in that floor.

The new capture at `python/djust/mixins/rust_bridge.py:677-678` grabs raw `Model` instances *before* the normalize loop turns them into denylist-filtered dicts, then re-attaches them to the getattr sidecar at `:1017-1018` (`sidecar.setdefault(...)`). Pre-PR, an explicitly-assigned model was replaced by its serialized dict and **never reached the sidecar**; post-PR it does. `Context::resolve` (`crates/djust_core/src/context.rs:285`) tries the value-stack dict first (password absent → miss), then falls to the sidecar and `getattr`s the field off the **raw model** (`:310`) — with **no denylist consultation anywhere in the walk**.

**Probe (all three floor fields leak into the HTML sent to the client):**
```python
u = User.objects.create_superuser(username="adm", password="pw", email="a@b.c")
# ctx["member"] = u ; template = "pw=[{{ member.password }}] su=[{{ member.is_superuser }}] st=[{{ member.is_staff }}]"
```
```
FLOOR: <div dj-id="0">pw=[pbkdf2_sha256$1000000$cFRq...=] su=[true] st=[true]</div>
LEAKED FLOOR FIELDS: ['password', 'is_superuser', 'is_staff']
```

**The auto-call adds a second vector**: method-based sensitive getters that eager serialization deliberately skips (`SKIP_METHODS` at `serialization.py:455-461`) are now *invoked* by the walk. `{{ member.get_session_auth_hash }}` renders the real session-auth hash:
```
SESS: <div dj-id="0">[b77e37cbfdc2c674afc32a20b504131f8480f1f258f59f5e3b0464ef37f82b19]</div>
```
This vector affects even the request-scoped `user` (already in the sidecar pre-PR) — pre-PR the bound method stringified; post-PR it executes and leaks.

**The kill-switch does NOT close this.** `_raw_models_for_sidecar` is ungated by `template_auto_call`; only `maybe_call` is. With `set_template_auto_call(False)`, the *field* leak persists (verified):
```
KILLSWITCH-OFF password render: <div dj-id="0">[pbkdf2_sha256$1000000$cFRq...=]</div>
LEAK WITH KILLSWITCH OFF: True
```
So an operator who flips the documented kill-switch for safety is still exposed.

**Why it's realistic**: assigning a model to context (`ctx["member"] = some_user`) is idiomatic, and the floor exists *precisely* because developers reference fields generically / by typo. The floor's whole promise (#1868: "a `djust_serializable_fields` allowlist alone can NOT re-expose a denied field") is silently void on the sidecar path.

**Proposed fix** (author's call): teach the sidecar walk to enforce the floor for `Model` getattr segments — either (a) pass the resolved denied-field set + `SKIP_METHODS` into the sidecar and have `resolve`/`maybe_call` refuse those segment names when the receiver `is_instance` of `django.db.models.Model` (render empty, matching the floor), or (b) retain an allow-scoped proxy instead of the bare model, or (c) keep raw models out of the sidecar and solve the reverse-relation/manager case (`{{ ws.memberships.count }}`) another way. Whatever the shape, add a regression test that `{{ member.password }}` / `{{ member.is_superuser }}` / `{{ member.get_session_auth_hash }}` render empty for an explicitly-assigned model **and** for request-scoped `user`, with the kill-switch both on and off.

---

### 🟡 SHOULD-FIX

1. **`attr_is_truthy` swallows *all* exceptions from the guard descriptor, not just `AttributeError`** (`context.rs:397`). Django uses `getattr(current, "alters_data", False)`, which returns the default only on `AttributeError` and *propagates* other exceptions. djust's `obj.getattr(name).ok()...unwrap_or(false)` treats a raising `alters_data`/`do_not_call_in_templates` descriptor as falsy → the guard is skipped → the callable is invoked. Contrived (these attrs are plain function markers in practice, never raising properties), so low severity — but it is a real deviation from the Django parity the ADR claims, and on the `alters_data` axis "skip the guard" fails *open*. Consider matching Django: only swallow `AttributeError`.

2. **Missing test for the `do_not_call_in_templates` + `alters_data` both-set ordering** (concern 1b). Verified manually: with both set, `do_not_call` wins (checked first in `maybe_call`, mirroring Django) → used as-is, not called, not empty (renders `[Object]`). Behavior is correct, but the suite has no row for it — add one so a future reorder of the two guard checks is caught.

---

### Verified GOOD (empirically)

- `{{ user.delete }}` on a real `User` — **not called**, user row survives, renders empty. ✅
- `alters_data` refusal holds on the **for-loop iterable** path (`renderer.rs:599`) and the **filter-arg** path (`filter_registry.rs:264`) — all three resolve sites share `maybe_call`; the for-iter probe rendered an empty `<ul>` with the sentinel un-executed. ✅
- Optional-args method (`def m(self,x=1)`) raising `TypeError` internally **propagates** (bind succeeds with 0 args); args-required renders empty. ✅
- **Flag-stamping completeness (#1646)**: all three `set_raw_py_objects` sites in `djust_live/src/lib.rs:418/459/871` have a matching `set_auto_call` (`:421/462/874`); no other sidecar-attach path exists. `Context::new()` defaults `auto_call=true` but no sidecar-less render reaches the walk. ✅
- **Perf**: `resolve` hits `self.get(key)` first and early-returns `Ok(None)` when no sidecar is attached (`context.rs:285`); `is_callable`/guards only run inside `maybe_call`, gated behind the sidecar branch. No added per-variable cost on the eager path. ✅
- **Kill-switch (auto-call axis)**: `set_template_auto_call(False)` restores the pre-ADR stringified-bound-method walk (test `test_kill_switch_restores_plain_getattr_walk`). ✅ (but see 🔴 — it does not gate the retention.)
- **Doc/citation spot-check**: 16 tests present and green (`test_template_auto_call_1985.py`); the ORM-autocall one-shot-per-path DEBUG warning fires as documented. The feature's CHANGELOG entry makes no full-suite count claim (only "16 tests"), which is accurate.

---

Probe scripts are available on request for the fixer.

johnrtipton:
## Re-review — PR #1986 serialization-floor fix (`a79b220f` + `2abfeef0`)

**Verdict: REQUEST_CHANGES** — the fix closes the *direct* leaks I originally reported (verified below), but the floor is **still bypassable** through the getattr walk, so the fix's central claim ("the floor holds on the lazy getattr path") is not fully met. All results reproduced through the real `LiveViewTestClient` → `Context::resolve` sidecar walk on the new HEAD with the rebuilt extension.

---

### 🔴 MUST-FIX — the floor is still bypassed for models reached through an auto-called manager/queryset method

`_SidecarModelProxy.__getattr__` only re-wraps a value when it is *directly* a `models.Model` (`serialization.py:653-655`). A **manager/queryset is not a `Model`**, so it is returned **raw and unwrapped**; the Rust walk then auto-calls a single-object method on it (`.first`, `.get`, `.latest`, `.earliest`, a reverse FK manager…), which returns a **raw, unproxied model** — and the next segment `getattr`s a floor field straight off it. The password hash leaks:

```python
u = User.objects.create_user(username="victimX", password="pw")
g = Group.objects.create(name="grpX"); u.groups.add(g)
# ctx["member"] = u
```
```
{{ member.groups.first.user_set.first.password }}
 => [pbkdf2_sha256$1000000$0LSMRActFg0WqbL2poBEgk$wFK1...EU=]     # LEAK

{{ member.groups.first.user_set.get.password }}
 => [pbkdf2_sha256$1000000$r2li9sX7lt0f9CbX3C5yLD$Qrcn...qNE=]     # LEAK (.get too)
```

This is **in-feature, not contrived**: the headline capability ADR-024 ships is exactly `{{ workspace.memberships.first.<field> }}` — traverse a manager, land on a model, read a field. Any view with `ctx["team"] = team` and a template touching `{{ team.members.first.password }}` (or `is_superuser`, or `get_session_auth_hash`) re-opens the floor the fix claims to seal. The docstring's own promise — *"Related models reached through the walk are wrapped recursively, so `{{ member.profile.ssn }}` is protected"* — is only true for **direct FK/one-to-one getattr**, not for the far more common **manager-method** traversal that auto-call is built around.

**Fix options** (author's call): (a) the more robust single-authority route — enforce the floor **in the Rust walk** for every segment whose receiver `is_instance` of `Model`, since the final `getattr` lands on a `Model` regardless of *how* it was reached (this also subsumes the #1987 type-exclusion follow-up cleanly); or (b) keep the Python-proxy approach but also wrap `Manager`/`QuerySet` values in a proxy whose auto-called single-object methods (`first`/`get`/`latest`/`earliest`/`__getitem__`) return `_SidecarModelProxy`-wrapped models. Either way, add a regression test with a manager-traversal chain (`{{ x.manager.first.password }}` and `.get`) — the current `TestSidecarSerializationFloor` suite has **no manager-returned-model variant**, which is why 8691 green tests didn't catch this (Action #1543 — a coverage suite must enumerate every variant of the surface it covers; direct-getattr is covered, manager-return is not).

Note: `.all.0`-style index traversal (`{{ member.groups.all.0.user_set.all.0.password }}`) rendered empty — only single-object manager methods reach an unwrapped model through the Rust walk, but `.first`/`.get` are the common ones.

---

### 🟡 SHOULD-FIX — the proxy's `_`-prefix skip: a segfault + minor info-disclosure

`__getattr__` skips **all** floor checks for names starting with `_` (`serialization.py:640` `if not name.startswith("_")`) and `getattr`s them straight off the model. Django forbids `_`-prefixed attribute access in templates precisely to prevent this; djust's Rust engine does **not**, so these resolve:

- `{{ member._meta }}` (terminal) → **hard segfault** — `Fatal Python error: Segmentation fault`, crashing the whole worker process (reproduced 3×). The terminal `extract::<Value>()` of Django's `Options` object is the trigger; `{{ member._state }}` renders `[Object]` and `pk` is fine, so it's object-specific.
- `{{ member._meta.db_table }}` → `[auth_user]` — schema/table-name disclosure.

Templates are developer-authored, so this is a footgun rather than a remote DoS — but a single `{{ obj._meta }}` typo taking down every concurrent connection on the worker is worth hardening. This is most likely **inherited from #1985** (the sidecar model retention made `_meta` reachable), not introduced by this security fix — but the fix's explicit `_`-skip re-affirms the gap and it is now reachable for *every* sidecar model (incl. request-scoped `user`). Consider having the proxy refuse `_`-prefixed names too (Django parity), and/or the Rust walk rejecting `_`-prefixed segments at parse time.

---

### Verified GOOD (empirically, on the new HEAD)

- **Direct leaks I originally reported are now blocked** — explicit *and* request-scoped, kill-switch ON and OFF:
  - `{{ member.password }}` / `{{ member.is_superuser }}` / `{{ member.is_staff }}` / `{{ member.get_session_auth_hash }}` → all empty (explicit assignment).
  - `{{ user.password }}` / `{{ user.get_session_auth_hash }}` → empty (request-scoped).
  - `set_template_auto_call(False)` still renders `{{ member.password }}` empty — the floor is **not** gated on the kill-switch. ✅
- **Legit feature intact** (no over-block): `{{ member.get_full_name }}` → `Jane Doe`, `{{ member.username }}` → `jdoe`, `{{ member.groups.count }}` → `2`, bare `{{ member }}` → `bareU` (renders the eager dict's `__str__`, not the proxy). ✅
- **Guard-drift closed (#1646)**: `_SENSITIVE_MODEL_METHODS` / `_SENSITIVE_MODEL_METHOD_PREFIXES` are module constants shared by the eager path (`serialization.py:472-473`) and the proxy (`:633-634`) — the two can no longer drift. ✅
- **Gate-off / non-tautology**: the floor tests assert the hash is absent; neutering the wrap makes the model render its floor fields → the assertions flip RED, so they are non-tautological for the direct-getattr variant they cover (consistent with the maintainer's 4/6-RED check). ✅ — but see 🔴: the suite doesn't cover the manager-return variant.

The direction is right (one denylist authority governing both channels is the correct #1646 shape); the remaining hole is that the authority isn't applied to models reached *through* the auto-call, which is the feature's main traversal shape.

johnrtipton:
## Stage 14 retro — PR #1986 (Fixes #1985, implements ADR-024)

**Shipped**: Django-parity template callable auto-call in the Rust sidecar getattr walk (ADR-024) **plus** a complete hardening of the sidecar serialization floor that the auto-call work exposed.

### The defining thread: mechanism, not surface path (#1646 at a new axis)

The adversarial Code Review found a **real credential leak in FOUR consecutive rounds** — each a *different surface path* of the **same two underlying mechanisms**:

| Mechanism | Surface paths found | Closed at |
|---|---|---|
| **A** — a floor field read off a raw model during the getattr walk | direct `{{ u.password }}`; manager traversal `{{ x.mgr.first.password }}`; `_meta` segfault/schema; non-model intermediary `{{ presenter.user.password }}`; method→model `{{ obj.get_related.password }}` | `protect_sidecar` chokepoint in `Context::resolve` (after root bind + each getattr + each auto-call) |
| **B** — a raw model `__dict__`-dumped during value conversion | queryset for-loop; `.values()`/`.values_list()` projection; raw list/tuple of models | `FromPyObject` routes any `isinstance(Model)` through `normalize_django_value` |

**Lesson (canonical)**: my first three rounds patched vector-by-vector (a Python proxy per shape) — and each patch invited the next vector, because I was enumerating an *unbounded* set of surface paths. The rounds-4/5 refactor to **two structural chokepoints, one per mechanism**, ended the whack-a-mole: the final review attacked 16+ container/reach shapes (generators, prefetch caches, GenericFK, custom `__getattr__`, `@cached_property`, namedtuples, sets, double intermediaries) and found **no 8th vector** — every path where a model is actually reachable is floored. This is the #1646 parallel-path-drift rule at the template-resolution axis: when a review finds "leak via path X," identify the *mechanism* and close it at one authority, don't patch the path.

### Other findings

- **Adversarial review is load-bearing, repeatedly.** CI was green on every single push; my own empirical probes passed each round. The independent worktree-isolated reviewer found the leak every time — including two vectors (5: `.values()` projections; 7: raw list/tuple) I only found because I *proactively* re-probed after the reviewer taught me the mechanism. Trust-but-verify: I reproduced every finding end-to-end through the real render path before fixing.
- **Reproduction fidelity + gate-off for security.** Every vector reproduced through the real `LiveViewTestClient.render_with_patches()` path (not a stub), and each of the 28 regression tests carries a gate-off (#1468) — neutering the guard makes exactly that test RED (proven for all: proxy wrap, transitive protection, `_`-prefix refusal, projection guard, `protect_sidecar` chokepoint, `FromPyObject` model routing).
- **#292 draft recovery.** A collaborator's uncommitted `scaffolding/templates.py` draft was silently dropped by a pre-commit stash-restore cycle; recovered from `~/.cache/pre-commit/patch<newest>` via `git apply`, then explicitly `git stash push`-protected through the remaining commits. The reflex worked as canonized.
- **Worktree-review discipline held** — no `core.bare`/detached-HEAD incident on main across three review rounds; the reviewer created fresh worktrees and restored HEAD each time.

### Numbers
7 vectors closed · 2 structural chokepoints · 28 tests (each gate-off-verified) · full suite 8697 passed · `make test-rust` 40/40 · mypy 826 · clippy + ruff + cargo fmt clean · CI 13/13 gating green · floor is **not** gated on the `template_auto_call` kill-switch · field-TYPE exclusion deferred to #1987.

`RETRO_COMPLETE`

PR #2005, opened 2026-07-02
fix(templates): resolve dict/list intermediates in the sidecar walk — Django parity (#1997)

### LINKED ISSUE #1997: bug: `Context::resolve`'s lazy getattr walk can't step into a dict/list intermediate — nested JSONField access renders silently empty
## Symptom (DJUST_LESSONS gotcha #6, hit in downstream production builds)

`{{ block.content.text }}` (where `content` is a Django `JSONField`) renders **empty** — no
exception, no warning, just missing content. Confirmed behavior matrix:

- `{{ group.label }}` on a plain context dict — **works**.
- `{{ conversation.title }}`, `{{ m.model.name }}` (model attribute access) — **works**.
- `{{ block.content.text }}`, `{{ block.content.columns }}` (nested access into a model's
  `JSONField`) — **fails silently** (renders empty).

Django's own template engine resolves `dict.key.key` correctly (`Variable._resolve_lookup` tries
dictionary lookup, then attribute lookup, then list-index lookup, in that order, at every segment)
— this is a djust-specific divergence, and a dangerous one: it produces missing output with zero
signal that anything went wrong.

## Root cause (mapped, citations grep-verified against this checkout)

Same general shape as the parallel-path drift in #1985/#1646-class bugs: djust has **two
independent resolution paths** for `{{ a.b.c }}`-style expressions, and only one of them handles
dict-shaped intermediates.

1. **The eager/Value-based path — `Context::get`** (`crates/djust_core/src/context.rs:165-214`).
   For a multi-segment key, it walks `parts[1..]` branching explicitly on the `Value` variant:
   `Value::List` for a numeric index (`list.get(index)`, line 197), `Value::Object` for a named
   field (`obj.get(*part)`, line 205) — anything else returns `None`. This is correct, dict-aware,
   Django-parity resolution. It's why `{{ group.label }}` on a plain context dict works: `group`
   materializes as a `Value::Object` (a JSON-shaped map), and `.label` is a normal `obj.get()` hit.
   The eager Python-side serializer that feeds this path, `normalize_django_value`
   (`python/djust/serialization.py:864-865`), explicitly recurses into `dict` values
   (`isinstance(value, dict): return {k: normalize_django_value(v, ...) ...}`), so JSON content
   that DOES get eagerly serialized into `Value`s is fine.

2. **The lazy sidecar path — `Context::resolve`** (`crates/djust_core/src/context.rs:285-332`),
   used for `raw_py_objects` — objects djust keeps as live Python references instead of eagerly
   converting to `Value` (per its own doc comment, this serves "request-scoped objects (`user`) and
   reverse relations (`memberships`)" — and, per this bug, apparently also serves per-loop template
   variables like `block` when they aren't otherwise materialized into the `Value` stack). When
   `Context::get`'s Value-based walk misses (the full dotted key doesn't resolve through the
   `Value` stack), `resolve()` falls back to `raw.get(first)` for the first segment, then walks
   `parts[1..]` via **`current.getattr(*part)` exclusively** (line 311) — no dict/list
   item-access fallback at all. Any failure (including a plain Python `dict` not having the
   requested key as an *attribute*) is caught by the `Err(_) => return Ok(None)` branch
   (lines 315-320, with its own comment: "Swallow AttributeError… invalid template paths render as
   empty, matching Django's default").

   Concretely, for `{{ block.content.text }}` via this path: `block` resolves as a raw Python model
   instance; `getattr(block, "content")` succeeds (`content` genuinely IS a model attribute) and
   returns the JSONField's deserialized value — a plain Python `dict`; `getattr(dict_value, "text")`
   then raises `AttributeError` (dict keys are not attributes), silently swallowed, resolving to
   `None` → empty output. Django's `Variable._resolve_lookup` would try `dict_value["text"]` first
   and succeed.

## Scope

1. **Fix `Context::resolve`'s per-segment walk** (`context.rs:310-322`) to mirror Django's
   `_resolve_lookup` order at each step, not just at the top level: try dict/mapping item access
   first (`PyDict`/`Mapping` `__getitem__` or `.get()`), then `getattr`, then list-index for a
   numeric segment against a sequence — matching what `Context::get`'s `Value`-based walk already
   does correctly for `Value::Object`/`Value::List`. This closes the parity gap between the two
   paths for exactly the intermediate-value shapes that Django itself explicitly branches on.
2. **Observability**: per the precedent set by #1985 (`warn_once_on_orm_autocall`), consider a
   debug-only one-shot warning when a nested lookup on a dict/list-shaped intermediate falls
   through to the empty-result path — silent blank output is strictly worse than an error for this
   class of bug, since nothing in the rendered page hints that anything is missing.
3. **Docs**: the templating guide should state plainly that `JSONField`/dict-shaped values pulled
   through the lazy sidecar path need dict-aware resolution (once fixed, this becomes a
   correctness note rather than a gotcha) — currently the guide doesn't distinguish the two
   resolution paths' capabilities at all.

## Testing

One test walking `{{ obj.jsonfield_attr.key }}` where `obj` is NOT independently eagerly
serialized into the `Value` stack (i.e. reaches `Context::resolve`'s fallback, not `Context::get`'s
Value-walk) — the two paths need independent coverage since a fix to one won't touch the other.
Also a nested-list-in-JSONField case (`{{ obj.jsonfield_attr.0.key }}`) to cover the numeric-index
branch of the same walk. Regression test should assert non-empty output, not just absence of an
exception (a regression here silently renders empty, so a naive "does it crash" test wouldn't
catch it).

Refs: DJUST_LESSONS gotcha #6 (downstream production build), ADR-024 / #1985 (sibling
parallel-path-drift fix for auto-call in the same `Context::resolve` function).

johnrtipton:
Fixes #1997.

## Problem
`Context::resolve`'s lazy sidecar getattr walk did `getattr` ONLY at each segment, so a dict/list intermediate reached mid-path resolved to **empty with no error**. Canonical case: a model's `JSONField` value — `{{ block.content.text }}` where `content` is a JSONField (a dict) — `getattr(dict, "text")` raises `AttributeError`, silently swallowed → missing output, zero signal.

Django's `Variable._resolve_lookup` tries **dict item access → attribute → integer list-index** at every segment. djust did only the middle step on the sidecar path; the eager `Context::get` path was already dict-aware (#1646 parallel-path drift).

## Fix
Mirror Django's per-segment order in the resolve loop (`crates/djust_core/src/context.rs`): `get_item` (mapping) → `getattr` → `get_item(int)` (list index).

**Floor safety (#1986):** the serialization-floor proxies (`_SidecarModelProxy`/`_SidecarQuerySetProxy`) implement no `__getitem__`, so item access on them fails and falls through to the floored `getattr` — no floor bypass. Verified `{{ member.password }}` stays empty.

## Tests
`python/djust/tests/test_nested_resolve_1997.py` — dict value, list index, list→dict, plain attr, missing-key-empty, dict-key-wins-over-attribute (Django order), floor-not-bypassed pin. Gate-off (#1468): reverting to getattr-only makes the nested tests RED.

Full suite 8704 passed · `make test-rust` 40/40 · clippy + cargo fmt clean.


johnrtipton:
## Stage 14 retro — PR #2005 (Fixes #1997)

**Task**: v1.1.0-5 drain — nested dict/list template resolution (Django `_resolve_lookup` parity in the sidecar walk).

### Quality: 5/5
### What went well
- Symptom-up: reproduced the exact `{{ block.content.text }}`-on-a-JSONField empty-render before touching code; the fix is a 3-line Django-parity ordering (`get_item → getattr → get_item(int)`).
- The #1646 parallel-path framing held — the eager `Context::get` path was already dict-aware; only the sidecar `Context::resolve` had drifted.
- The one real risk (adding item-access re-opening the #1986 serialization floor) was verified two ways: mechanistically (proxies declare no `__getitem__`, so item access falls through to the floored `getattr`) and empirically (independent worktree review drove 19 floor vectors → all `[]`, no leak).

- Stage-11 🟡: `test_dict_key_wins_over_attribute` first placed the dict directly in context → resolved via the eager path, not the sidecar walk under test (passed under gate-off). Re-routed through a custom object; then a SECOND gate-off pass showed the substring assertion was still insensitive (`dict.items()` repr contains the value), so tightened to the exact bracketed value + `dict_items` absence. Now RED under getattr-only. (#1200/#1468 — a test that doesn't fail when the change is reverted is not a test.)

### Verified
Full suite 8704 · `make test-rust` 40/40 · clippy + cargo fmt clean · CI 13/13 gating green · independent worktree adversarial review APPROVE (floor airtight across 19 vectors, Django-parity confirmed, gate-off non-tautological).

`RETRO_COMPLETE`

PR #2006, opened 2026-07-02
fix(deps): pin redis>=5.0.0,<8 — redis-py 8.x crashes channels_redis (#1995)

### LINKED ISSUE #1995: bug: djust's own `redis` extra permits redis-py 8.x, which crashes the canonical `channels_redis` production setup
## Symptom (DJUST_LESSONS gotcha #15, hit in downstream production builds)

Following djust's own deployment guide verbatim — `channels_redis.core.RedisChannelLayer` for
`CHANNEL_LAYERS`, required per `docs/website/guides/deployment.md:852` ("`channels_redis` is
required (not optional) when any view uses `push_to_view`, presence, cursor tracking, or any
cross-process feature") — produces a flashing "reconnecting" banner every few seconds under any
real multi-process load, with `Exception inside application` / `redis.exceptions.TimeoutError`
in the server log.

## Root cause (grep-verified against the installed package + this checkout)

`channels_redis`'s receive loop blocks on `bzpopmin(timeout=5)`. redis-py 8.0 changed its
socket-read timeout handling on that blocking call path, and the resulting
`redis.exceptions.TimeoutError` is uncaught inside the ASGI consumer — it crashes and the client
reconnects, repeating every few seconds under load. Downstream, pinning `redis>=5,<8` (7.4.1
verified) eliminated the crash entirely: 0 errors in a soak that previously failed within seconds
of connecting.

This is not an absent constraint — it's a too-permissive one, and it's the SAME constraint used in
two places:

```
# pyproject.toml:57-59 ([redis] extra)
redis = [
    "redis>=5.0.0,<9",
]
# pyproject.toml:112 (dev group, identical bound)
    "redis>=5.0.0,<9",
```

djust does not depend on `channels_redis` directly (it's not in `pyproject.toml` at all — only
`channels[daphne]>=4.0.0,<5` is a base dependency), so djust can't pin channels_redis's own
version. But djust's `[redis]` extra's `redis-py` bound (`<9`) is the one most projects will
install into the SAME environment as `channels_redis` (since djust's own docs, e.g.
`docs/website/guides/error-codes.md:82` and `docs/website/advanced/server-push.md:155`, show
`channels_redis.core.RedisChannelLayer` as the canonical `CHANNEL_LAYERS` backend) — and `<9`
happily allows the 8.x range that breaks `channels_redis`'s `bzpopmin` call. Confirmed via
`grep -rn bzpopmin python/` in this checkout: djust's own Redis state/presence backend code does
not use `bzpopmin` itself, so this is purely a shared-dependency compatibility gap between the
version djust's extra permits and the version its own recommended companion package tolerates —
not a bug in djust's own Redis backend code.

**What's confirmed by inspection vs. by this project's own testing**: the `pyproject.toml` bounds
and the deployment-guide "required" language are confirmed directly from djust's source in this
checkout. The `bzpopmin`/redis-py-8.0-timeout failure mechanism itself is confirmed empirically by
the downstream project (repeatable crash with redis-py 8.x + channels_redis 4.3.0, repeatable fix
by downgrading to redis-py 7.4.1) — this issue doesn't have channels_redis's own source available
to cite a line number for the timeout-handling change, but the fix/reproduction is solid.

## Proposed scope

One of:
1. Tighten the `[redis]` extra's bound to `redis>=5.0.0,<8` until channels_redis (or redis-py) ships
   a compatibility fix, since `<9` currently permits a version combination that breaks djust's own
   documented, "required (not optional)" production channel-layer setup.
2. Add a startup/doctor check (`djust_audit` or the system-check framework) that warns when
   `channels_redis` is installed alongside an incompatible redis-py version and `RedisChannelLayer`
   is configured — surfacing this before a project ships to production and hits the reconnect loop
   under load rather than in local dev (where lower connection churn may not trigger it reliably).
3. At minimum, document the known-good `redis-py` range explicitly in
   `docs/website/guides/deployment.md` next to the existing `channels_redis` "required" callout,
   since the current guide doesn't mention a version constraint at all.

Refs: DJUST_LESSONS gotcha #15 (learnings doc from a downstream production app), `pyproject.toml:57-59,112`, `docs/website/guides/deployment.md:127-134,742,852`.

johnrtipton:
Fixes #1995.

## Problem
redis-py 8.0 changed socket-read-timeout handling on `channels_redis`'s `bzpopmin(timeout=5)` receive loop; the resulting `redis.exceptions.TimeoutError` is uncaught inside the ASGI consumer. A djust deployment following the docs verbatim (`channels_redis` for `CHANNEL_LAYERS`, required for `push_to_view`/presence/cursor) reconnect-storms every few seconds under multi-process load.

## Fix
The constraint was `redis>=5.0.0,<9` in all three sites (`pyproject.toml` `[redis]` + `[tenants-redis]` extras + dev group) — too permissive, it allowed 8.x. Pinned `<8` (the issue verified `redis>=5,<8` gives 0 errors in a soak that previously failed within seconds). Defensive: the lock already resolved to 6.4.0, but `<9` let a future re-lock drift to the crashing 8.x.

## Verified
`uv lock` resolves redis 6.4.0 (`<8`); 15 redis-touching tests pass; lockfile self-entry + doc-snippet gates green.


johnrtipton:
## Stage 14 retro — PR #2006 (Fixes #1995)

**Task**: v1.1.0-5 drain — pin `redis>=5.0.0,<8` (redis-py 8.x crashes `channels_redis`).

### Quality: 5/5
### What went well
- Grep-verified the constraint appeared in THREE sites (`[redis]` + `[tenants-redis]` extras + dev group), not the two the issue cited — the symbol-migration grep canon (#1391) caught the third. Single bulk transformation (`replace_all`) per the #1312 rule.
- Root cause is upstream (redis-py 8.0's `bzpopmin` socket-timeout change surfacing as an uncaught `TimeoutError` in the ASGI consumer) — a pin is the right fix, not a code workaround.

### What didn't
- Nothing material. The lock already resolved to 6.4.0, so this is a defensive tightening (prevents a future re-lock drifting to the crashing 8.x) rather than an immediate downgrade — noted explicitly in the CHANGELOG so it doesn't read as a no-op.

### Verified
`uv lock` resolves redis 6.4.0 (`<8`); 15 redis-touching tests pass; lockfile self-entry + doc-snippet gates green; CI 13/13 gating green. Two-commit shape (deps, then CHANGELOG).

`RETRO_COMPLETE`

PR #2007, opened 2026-07-02
fix(security): TYPE-based serialization floor — always-drop BinaryField + encrypted-field types (#1987)

### LINKED ISSUE #1987: security: field-TYPE-based serialization exclusion (BinaryField + encrypted-field types), both paths
Follow-up from PR #1986 (ADR-024) review. The name/method serialization floor (`_ALWAYS_EXCLUDED_FIELDS` + `DJUST_SENSITIVE_FIELDS` + per-model `djust_exclude_fields`) is NAME-based. A defense-in-depth enhancement is to also exclude by field **type**, applied to BOTH serialization channels (eager `DjangoJSONEncoder` + the `_SidecarModelProxy` sidecar walk) so they can't drift (#1646).

## Scope
- **`BinaryField` — always exclude.** Raw bytes are never sensibly rendered to a template and may be large/sensitive. Unambiguous always-drop.
- **Encrypted-field base-class detection** (optional): drop fields whose class derives from known encrypted-field bases (`django-encrypted-fields`, `django-fernet-fields`, etc.) when those libs are installed — best-effort, no hard dependency.
- **Configurable type set** (optional): `LIVEVIEW_CONFIG['sensitive_field_types']` for project-specific types.
- **NOT** `FileField`/`ImageField` (those should serialize a URL, not be excluded).

## Why separate from #1986
This is a floor-POLICY change (belongs in `docs/SECURE_DEFAULTS.md` Pattern 1), needs its own threat model + tests, and does not close a sidecar-specific gap — a plain `CharField` named `api_key` leaks via BOTH paths today, so it's general hardening, not a regression fix. #1986 correctly stayed scoped to the name/method floor (which closes the confirmed password/is_superuser/session-hash leaks).

## Acceptance
- `BinaryField` excluded on the eager path AND refused by `_SidecarModelProxy`.
- Shared type-exclusion authority (one function both paths call, #1646).
- `docs/SECURE_DEFAULTS.md` Pattern 1 updated.
- Tests: eager + sidecar, per type rule.

Refs: PR #1986, ADR-024, #1868 (serialization floor), #1646.

johnrtipton:
Fixes #1987. Follow-up hardening to #1986 (the name/method serialization floor).

## Problem
The #1986 floor drops sensitive fields by **name** (`password`/`is_superuser`/`is_staff` + `DJUST_SENSITIVE_FIELDS` + per-model `djust_exclude_fields`). It does not cover a field whose **type** should never reach the client regardless of name — most notably `BinaryField` (raw bytes) and encrypted columns.

## Fix
A complementary, name-independent TYPE floor via one shared authority `_field_type_is_excluded(field)`, called by BOTH client-bound paths so they can't drift (#1646):
- **Eager**: `DjangoJSONEncoder._serialize_model_safely` — checked right after the name-based `_field_is_serializable`.
- **Sidecar**: `_SidecarModelProxy.__getattr__` — via `_field_type_excluded_for(model_class, name)`.

Excluded: `BinaryField` (unconditional); best-effort encrypted-field types (an MRO class name containing `Encrypted`/`Fernet` — django-encrypted-fields / django-fernet-fields — no hard dependency, fail-closed); and any class named in the new `LIVEVIEW_CONFIG['sensitive_field_types']` (empty by default). **Not** excluded: `FileField`/`ImageField` (they serialize a URL).

## Tests (14, `test_field_type_exclusion_1987.py`)
Authority unit tests (BinaryField / CharField / FileField / encrypted-named / resolve-by-name / non-field) + configured-type + eager-path + sidecar-path. **Gate-off (#1468)**: reverting either wired check makes the `BinaryField`-leak sentinel RED (verified). #1986 floor suite (45) + serialization/template regression set (466) stay green.


johnrtipton:
## Stage 14 retro — PR #2007 (Fixes #1987)

**Task**: v1.1.0-5 drain — TYPE-based serialization floor (follow-up hardening to the #1986 name/method floor).

### Quality: 5/5
### What went well
- The #1646 parallel-path lesson from #1986 was applied up front: one shared authority (`_field_type_is_excluded`) that BOTH the eager encoder and the sidecar proxy call — the review confirmed the two wired sites protect *genuinely distinct* vectors (single-model dotted access vs queryset/list iteration via `__djust_serialize__`), so neither is decorative.
- The independent worktree review was load-bearing: it validated the encrypted heuristic against a REAL library (`django-searchable-encrypted-fields` — `EncryptedCharField`/`TextField`/`IntegerField` + `EncryptedFieldMixin` all caught) rather than trusting the substring match by inspection, and drove the leak-closed proof across every traversal vector through the real render path.

- Review M2: the `Encrypted`/`Fernet` heuristic was case-SENSITIVE — a lowercased variant would slip the floor. Fixed to `.casefold()` (configured names stay case-exact). +1 test.
- Review M1: a heuristic drop was silent — exactly the "no signal, field just vanishes" anti-pattern the #198x downstream issues complain about. Added a one-shot DEBUG breadcrumb per field class (fires only for the NAME heuristic, not `BinaryField`/config). +3 tests (incl. `decrypted`-is-not-a-false-positive + one-shot-and-only-for-heuristic).
- Follow-up worth considering (non-blocking): a type-floor opt-out parallel to the name floor's `djust_serialize_sensitive_fields` (the eager `continue` doesn't consult `optout`). Not needed for the security goal; `sensitive_field_types` + rename cover the escape hatch.

### Verified
18 tests (was 14; +4 for M1/M2) · gate-off non-tautological (reverting either wired check reopens the `BinaryField`/plaintext-token leak through the real render path) · #1986 floor suite 45 + serialization/security sweep 110 green · CI 13/13 gating green · worktree adversarial review APPROVE (leak closed across every traversal vector; heuristic confirmed against real libs; perf ~0.4µs/field).

`RETRO_COMPLETE`

PR #2008, opened 2026-07-02
fix(state): zero-arg set_changed_keys() forces a re-render for DB-only changes (#1992)

### LINKED ISSUE #1992: `set_changed_keys()` doesn't cover the "handler only mutated the DB, no `self.attr` at all changed" case — needs a zero-arg form
## Symptom (DJUST_LESSONS gotcha #9, hit in downstream production builds)

On djust 1.0.8, a handler that mutates only database rows and touches no public `self.*`
attribute produces **zero re-render** — even though `get_context_data()` would return different
HTML on the next render (it re-queries the DB every time). Concretely, in a chat app with
branching conversations:

```python
def switch_branch(self, message_id, child_id):
    parent = models.Message.objects.get(id=message_id)
    parent.active_child_id = child_id
    parent.save(update_fields=["active_child_id"])   # DB is correct
    # ...no self.x assignment anywhere in this handler...
```

The DB write succeeds (verified via `manage.py shell`), but the browser keeps showing the old
branch — no patch, no full-HTML update, nothing. A sibling handler (`regenerate`) only *appeared*
to work correctly because it happens to also set `self.streaming = True` as a side effect, which
trips the diff by accident, not by design. The only workaround found was a throwaway public
counter with no semantic meaning of its own:

```python
    self.rev = 0   # exists purely to trip the diff

def switch_branch(self, message_id, child_id):
    ...
    self._touch()   # self.rev += 1 — forces the render, never read by any template
```

## Root cause (grep-verified against the installed 1.0.8 AND the current `[Unreleased]` source)

The render-skip decision lives in `_snapshot_assigns`/the pre/post comparison in
`python/djust/websocket.py` (e.g. the deferred-activity path at `websocket.py:1577-1611`, and the
mirrored logic in the main WS event-handling path and `python/djust/api/dispatch.py:367-411`):
public attrs are snapshotted before and after the handler runs; if the snapshots are equal, the
render is skipped entirely. A change that lives ONLY in the database is invisible to this
snapshot by construction — there's no `self.*` write for it to see.

**This is a real, known gap the framework is actively working on — but not yet closed for this
specific trigger.** The `[Unreleased]` CHANGELOG (not yet in 1.0.8) adds
`LiveView.set_changed_keys(keys)` (#1981), which sets the `_force_full_html` flag
(`python/djust/mixins/rust_bridge.py:575-620`) to bypass the pre/post-assigns skip — this is
exactly the "sanctioned skip bypass" primitive gotcha #9 asked for. However:

1. **`set_changed_keys`'s `keys` parameter is required, not optional**
   (`keys: Union[str, Iterable[str]]`, no default) — a handler that changed *only* the database,
   with no corresponding `self.attr` at all, has no real attribute name to pass. The only options
   are inventing a fake one (exactly the `self.rev` counter workaround gotcha #9 describes) or
   passing an empty string/tuple, which isn't documented as supported and reads as a misuse of
   the API.
2. **The docs and docstring frame the method entirely around in-place mutation of nested state**
   (`docs/state-management/STATE_MANAGEMENT_API.md:1446-1487`, and the method's own docstring in
   `rust_bridge.py`) — e.g. `self.rows[0]["tags"].append(tag); self.set_changed_keys("rows")`.
   Neither mentions the "a handler mutated the database and touched no `self.attr` whatsoever"
   trigger, which needs the identical `_force_full_html` bypass but has no natural key to name.

So even once `set_changed_keys()` ships, gotcha #9's exact scenario (DB-only mutation, zero
`self.attr` changes) still has no *clean* API — only the same fake-attribute-name workaround,
now routed through a real method instead of a raw counter.

## Proposed scope

1. Give `set_changed_keys`'s `keys` parameter a default (e.g. `keys: Union[str, Iterable[str]] =
   ()`), so `self.set_changed_keys()` with zero arguments is a supported, documented way to just
   set `_force_full_html = True` and force the next render — no attribute name required or
   invented.
2. Alternatively/additionally, a one-line alias with an unambiguous name for this exact use case
   — `self.force_render()` or `self.rerender()` — that's just `self._force_full_html = True`,
   distinct in *intent* from `set_changed_keys(keys)` even if the underlying flag is shared.
3. Update `docs/state-management/STATE_MANAGEMENT_API.md`'s `set_changed_keys` section (and the
   method's docstring) to explicitly cover the "handler mutated only the database, no `self.attr`
   changed at all" trigger as a second, equally common use case alongside in-place nested-container
   mutation — right now a reader would reasonably conclude the method doesn't apply to their
   situation at all, since every example involves a `self.*` container.

## Testing

A regression case mirroring the existing `set_changed_keys` in-place-mutation tests
(`test_template_auto_call_1985.py`-style doc-claim-verbatim pattern used elsewhere in this
CHANGELOG): a handler that writes to the DB via the ORM, calls `self.set_changed_keys()` with no
arguments, and asserts a non-noop render frame is sent on the real `ViewRuntime.dispatch_event`
path (not just `LiveViewTestClient`, which the `#1981` PR notes bypasses the skip).

Refs: DJUST_LESSONS gotcha #9 (github.com/johnrtipton — downstream chat app built on djust),
`#1981` (adds `set_changed_keys`, currently `[Unreleased]`, not yet in 1.0.8).

johnrtipton:
Fixes #1992.

## Problem
A handler that mutates only the database (e.g. `msg.save(update_fields=["active_child_id"])`) and assigns no public `self.*` attribute produced **no re-render** — auto change-detection (`_snapshot_assigns`) saw nothing changed and auto-skipped the event, even though `get_context_data()` re-queries the DB and would render different HTML. The browser kept showing stale content. `set_changed_keys("attr")` (#1981) required naming a changed attr; there was no "nothing on `self` changed, just re-render" form.

## Fix
`set_changed_keys()` with **no arguments** now forces a full re-render via the existing `_force_full_html` bypass, without naming a key (`keys` defaults to `None`). One-line semantic: the zero-arg branch sets the force flag and returns; the keyed form is unchanged.

## Tests (4, `test_set_changed_keys_zero_arg_1992.py`)
On the REAL `ViewRuntime.dispatch_event` path (the `LiveViewTestClient` bypasses the `pre==post` skip and would falsely hide the bug — #1650 reproduction fidelity, same harness as #1981). **Gate-off (#1468)**: the same DB-only mutation *without* the zero-arg call auto-skips (noop); *with* it renders — so the render is attributable to exactly the zero-arg hatch (verified: neutering the force flag makes the render test RED). mypy strict-island clean; 42-test state/event regression green.


johnrtipton:
## Stage 14 retro — PR #2008 (Fixes #1992)

**Task**: v1.1.0-5 drain — zero-arg `set_changed_keys()` for the DB/external-only mutation case.

### Quality: 5/5
### What went well
- The #1981 harness (`_EventSpineMixin` + `_event_runtime_with_view`) was exactly the right reproduction-fidelity tool — the `LiveViewTestClient` bypasses the `pre==post` skip and would have falsely shown the hatch as unnecessary (#1650). Reused it verbatim.
- The fix is minimal and mechanism-honest: `_force_full_html` (the existing skip bypass) already forces the render, so the zero-arg form just sets that flag without naming a key — confirmed empirically that the flag alone renders through the real `dispatch_event`.
- Left-shifted gate-off (#1468 at implementer stage): neutering the force flag made the render test RED while the gate-off baseline stayed green, proving the render is attributable to exactly the zero-arg hatch.

### What didn't
- Nothing material. Inline-verified (small Python change, no new public surface beyond an optional arg, event-path covered) per the pipeline low-risk shortcut — no worktree review needed.

### Verified
4 new tests (2 real-`dispatch_event` + 2 unit) · gate-off non-tautological · mypy strict-island clean · 42-test state/change-detection/event regression green · CI 13/13 gating green. Two-commit shape.

`RETRO_COMPLETE`

PR #2009, opened 2026-07-02
fix(markdown): don't split a complete fence's closing ``` as a provisional paragraph (#1998)

### LINKED ISSUE #1998: fix: `split_provisional` misclassifies a closing code-fence line as an unterminated inline-code span
## Symptom (DJUST_LESSONS gotcha #10, hit in downstream production builds)

Passing Markdown that is *only* a fenced code block — ` ```python\ndef greet():\n    pass\n``` ` with no trailing newline and no surrounding prose — to `{% djust_markdown %}` renders:

```html
<p class="djust-md-provisional">```</p>
```

i.e. the entire fenced block renders correctly EXCEPT the closing ` ``` ` line gets split off and re-rendered as an escaped, literal, provisional paragraph — instead of the fence being recognized as complete and the whole thing coming out as `<pre><code class="language-python">...</code></pre>`.

The identical fence renders correctly when it has a trailing newline, or when it's embedded inside a larger document (prose before/after the fence) — which is why one downstream app's chat-transcript messages (prose + fence, always followed by more content or a trailing newline) highlight code fine, while a "code artifact" panel rendering *just* the code body as its own complete document did not.

## Root cause (grep-verified, `crates/djust_templates/src/markdown.rs`)

`render_markdown` (`markdown.rs:81-113`) calls `split_provisional(src)` when `opts.provisional` is true (the default — `RenderOpts::default()`, `markdown.rs:47-55`). `split_provisional` (`markdown.rs:120-134`):

```rust
pub(crate) fn split_provisional(src: &str) -> (&str, &str) {
    if src.is_empty() || src.ends_with('\n') {
        return (src, "");
    }
    if inside_unclosed_fence(src) {
        // pulldown-cmark handles unclosed fenced code gracefully on its own.
        return (src, "");
    }
    let last_nl = src.rfind('\n').map(|i| i + 1).unwrap_or(0);
    let trailing = &src[last_nl..];
    if looks_unterminated(trailing) {
        (&src[..last_nl], trailing)
    } else {
        (src, "")
    }
}
```

For `src = "```python\ndef greet():\n    pass\n```"` (no trailing `\n`):
- `src.ends_with('\n')` is false — does not early-return.
- `inside_unclosed_fence(src)` (`markdown.rs:136-140`) counts lines starting with `` ``` `` — there are exactly 2 (open + close), an EVEN count, so the fence is correctly recognized as *closed* and this does not early-return either.
- `last_nl` finds the position after the last `\n`, so `trailing = "```"`  — just the closing fence delimiter, 3 backtick characters.
- `looks_unterminated("```")`  (`markdown.rs:145-160`) checks `line.matches('`').count() % 2 == 1` — **3 % 2 == 1 → true**. The function returns `true`, so `split_provisional` treats the closing fence line as "looks mid-token" and hands it to the provisional-wrapper path instead of to `pulldown-cmark`.

The bug: `looks_unterminated` is a character-parity heuristic designed to catch a *dangling inline-code span* like `` `foo `` (an odd, unpaired backtick mid-line). It has no concept of "this line IS a fence delimiter, not inline code" — `inside_unclosed_fence` already parses for fence-delimiter lines earlier in `split_provisional`, but that fence-awareness isn't reused when the *single trailing line itself* happens to be a fence delimiter. Three backticks forming a complete, valid closing `` ``` `` marker get counted identically to three stray backtick characters that don't.

This only manifests when: (1) the source has no trailing `\n` after the closing fence (the common shape when a caller passes exactly `fenced_body_string.strip()` or similar, e.g. code straight out of an LLM response or a stored artifact body), AND (2) the closing fence line is also the *last line* of the source (no trailing prose/blank line after it) — i.e. exactly "a standalone fenced block with no trailing newline," which is a completely reasonable and common shape for a caller rendering a code artifact.

## Proposed fix

In `split_provisional`, before running `looks_unterminated` on the trailing line, check whether that line is *itself* a fence delimiter (`trimmed.starts_with("```")` or the backtick-count-only variant already used by `inside_unclosed_fence`'s per-line test) and short-circuit to `(src, "")` in that case — a line that IS a complete fence marker can never be "mid-token," regardless of backtick parity. Alternatively/additionally, exclude lines whose trimmed content is *entirely* backticks (i.e. `line.trim().chars().all(|c| c == '`')`) from the backtick-parity check in `looks_unterminated`, since such a line is unambiguously a fence delimiter, never inline code (inline code backticks are never the sole content of a line with nothing else on it in valid Markdown fence syntax).

Note: `provisional=False` is already available and documented (`docs/website/guides/streaming-markdown.md:60`, `:89-90`: "Turn it off when you're rendering a finalised document") as a workaround for callers who know their content is complete and static — so no new API surface is needed here, just the heuristic fix. Worth calling out in the guide's fenced-code example specifically, since "a lone, complete fenced block" is a non-obvious case where the *default* (streaming-safe) heuristic misfires even though the content isn't actually mid-stream.

## Testing

- Regression test: `render_markdown("```python\ndef greet():\n    pass\n```", RenderOpts::default())` must produce a real `<pre><code>` block, not a `djust-md-provisional` wrapper around the closing fence.
- Edge cases worth covering in the same pass: a fence with a trailing newline (already works — should stay working), a fence embedded in prose (already works — should stay working), an *unclosed* fence (already handled by `inside_unclosed_fence`'s early return — should stay working), and a line that's genuinely just stray backticks with no fence semantics, e.g. a lone `` ` `` `` (odd count, should still be treated as unterminated).
- `python/djust/tests/test_markdown.py` already has provisional-heuristic test coverage (`test_provisional_mode_unclosed_bold_renders_as_text`, `test_split_provisional_edge_cases`) — add the standalone-closed-fence case alongside those.

Refs: DJUST_LESSONS gotcha #10 (downstream field notes), `crates/djust_templates/src/markdown.rs`.

johnrtipton:
Fixes #1998.

## Problem
`{% djust_markdown %}` on a code-only artifact — a complete fenced block with **no trailing newline** and no surrounding prose (` ```python\ndef greet():\n    pass\n``` `) — rendered the closing ` ``` ` as an escaped provisional paragraph (`<p class="djust-md-provisional">` ``` `</p>`) instead of completing the `<pre><code>` block.

`split_provisional` (`crates/djust_templates/src/markdown.rs`) treats a trailing line with an ODD backtick count as an unterminated inline-code span. A lone closing ` ``` ` has 3 backticks (odd), so for a *complete but newline-less* fence — where `inside_unclosed_fence` correctly reports balanced fences — the closing delimiter was split off. This is why a chat transcript (prose + fence, always followed by more content or a trailing newline) highlighted code fine while a "code artifact" panel rendering just the code body did not.

## Fix
A one-line guard: a trailing line that is itself a ` ``` ` fence delimiter completes a balanced fence (`inside_unclosed_fence` already returned false above), so keep the whole block stable.

## Tests
Rust unit (`split_provisional`) + render (`render_markdown`) tests + a Python end-to-end test (`test_markdown.py`). **Gate-off**: reverting the guard makes both new Rust tests RED (verified). Also asserts no regression to the genuinely-open streaming case (odd count → still provisional).


johnrtipton:
## Stage 14 retro — PR #2009 (Fixes #1998)

**Task**: v1.1.0-5 drain — `split_provisional` misclassified a complete fence's closing ` ``` ` as unterminated inline code.

### Quality: 5/5
### What went well
- Root cause traced by hand before coding: a lone ` ``` ` has 3 backticks (odd) → `looks_unterminated` true, but `inside_unclosed_fence` already reported the fences balanced, so the trailing ` ``` ` is definitionally a CLOSING marker. One-line guard.
- The "why does it work with a trailing newline / inside prose but not standalone" detail in the issue was the tell — a probe (printing the real render output) confirmed the fix before finalizing the test assertion.

- First test assertion `!out.contains("<p")` false-failed: `<pre>` contains the substring `<p`. Tightened to `<p>` (a real paragraph tag). Caught immediately by running the test.

### Verified
2 Rust tests (`split_provisional` unit + `render_markdown` end-to-end) + a Python end-to-end test (`test_markdown.py`, 8 total) · gate-off (#1468): reverting the guard makes both Rust tests RED, and the genuinely-open streaming case stays provisional (no regression) · `cargo test -p djust_templates` + clippy + fmt clean · CI 13/13 gating green. Inline-verified (one-line Rust guard, well-covered).

`RETRO_COMPLETE`

PR #2010, opened 2026-07-02
fix(config): honor DJUST_CONFIG as a LiveView-key fallback + fit default upload chunk under the frame limit (#1993)

### LINKED ISSUE #1993: bug: `max_message_size`/`rate_limit`/etc. silently ignored when set in `DJUST_CONFIG` instead of `LIVEVIEW_CONFIG`; default upload chunk exceeds default frame limit by exactly 21 bytes
## Symptom (DJUST_LESSONS gotcha #19, hit in downstream production builds)

A `dj-upload` attachment past a single small file fails with `Message too large (65557 bytes)`, with **no application misconfiguration at all** — this reproduces on a fresh project using only documented defaults (`allow_upload(...)`, no custom `DJUST_CONFIG`/`LIVEVIEW_CONFIG`).

Trying to fix it by raising the limit via `DJUST_CONFIG = {"max_message_size": 262_144}` (the more prominent, already-in-use settings dict for tenancy/presence/state-backend in most apps) is a **silent no-op** — no error, no warning, the 65536-byte default keeps applying.

## Root cause (mapped, citations grep-verified against `djust-org/djust` main)

**1. Two separate, similarly-named settings namespaces, defined in the same file, with zero cross-awareness.**

- `LiveViewConfig._load_from_settings()` (`python/djust/config.py:262-289`) governs `max_message_size`, `rate_limit`, `event_security`, `websocket_compression`, etc. It checks exactly one settings key:
  ```python
  if hasattr(settings, "LIVEVIEW_CONFIG"):
      self._config.update(settings.LIVEVIEW_CONFIG)
  ```
  (`config.py:267-268`). There is no fallback check for `DJUST_CONFIG` anywhere in this method — only a handful of unrelated top-level aliases (`DJUST_WS_COMPRESSION`, `DJUST_VDOM_CACHE_*`, `DJUST_STATE_SNAPSHOT_*`) are individually read (`config.py:271-289`).
- A **completely separate** function, `get_djust_config()` (`config.py:471-485`), reads a different dict: `getattr(settings, "DJUST_CONFIG", {})` (`config.py:485`). This is what backs tenancy (`tenants/middleware.py:111,150`), the `suppress_checks` list (`checks/utils.py:101`), and `live_tags.py:863`.
- Notably, the comment directly above the `LIVEVIEW_CONFIG` check (`config.py:269-270`) says *"convenience aliases for the nested `LIVEVIEW_CONFIG` / `DJUST_CONFIG` dicts"* — which reads as if `DJUST_CONFIG` is handled here too, but it never is; only specific flat `DJUST_*`-prefixed keys are aliased, not a nested `DJUST_CONFIG` dict. This is the comment that makes the trap easy to fall into even when reading the source.
- **Correction to the original report**: djust's own shipped docs (`docs/EVENT_HANDLERS.md:808-812`, `docs/SECURITY_GUIDELINES.md:158-162`, `docs/website/advanced/security.md:70-74`) all correctly show `max_message_size` under `LIVEVIEW_CONFIG`. The confusion isn't from djust's own documentation being wrong — it's from the two dicts being so similarly named, defined in the same file, with the misleading comment above, and both being genuinely idiomatic-looking Django settings patterns. It's an easy mistake to make even for someone who's read the source, not just someone going off a stale/third-party doc snippet.

**2. Independent of (1): the default chunk size doesn't fit under the default message-size limit.**

- The upload client's default chunk size is exactly the same as the default message-size limit: `DEFAULT_CHUNK_SIZE = 64 * 1024` (`python/djust/static/djust/src/15-uploads.js:21`), and `max_message_size` also defaults to `65536` (`config.py:53`, checked server-side in `websocket.py:1949`).
- Every chunk frame gets a 21-byte binary header prepended before the payload — 1 byte frame type + 16 bytes upload-ref UUID + 4 bytes big-endian chunk index (`buildFrame()`, `15-uploads.js`, `header = new Uint8Array(21)`; confirmed via `frame = new Uint8Array(21 + payload.byteLength)`).
- `65536 (chunk payload) + 21 (frame header) = 65557` — exactly the observed failure size, and exactly `21` bytes over the `65536`-byte default limit. **A brand-new project using nothing but `allow_upload(...)` with default settings cannot successfully upload any file requiring more than a fractional first chunk**, purely from the two defaults (chunk size, message-size limit) not being sized to work together.

## Proposed scope

1. Merge `DJUST_CONFIG` and `LIVEVIEW_CONFIG` into one settings dict, **or** have `LiveViewConfig._load_from_settings()` also check `DJUST_CONFIG` as a fallback/alias (with a deprecation-style debug log noting which key it used, to surface the ambiguity rather than resolve it silently).
2. Fix the misleading comment at `config.py:269-270` regardless of (1) — it currently implies `DJUST_CONFIG` is handled by this method when it isn't.
3. Make the two upload defaults compatible out of the box — either raise `max_message_size`'s default to comfortably exceed one default chunk + its 21-byte frame overhead (e.g. `131072`), or shrink `DEFAULT_CHUNK_SIZE` to comfortably fit under the current `65536` limit (e.g. `65536 - 256` for headroom across all frame types, not just `FRAME_CHUNK`'s 21 bytes). Either direction removes the "works only if you happen to raise a setting you'd have no reason to look for" trap for the most common attachment feature in a chat/document app.
4. Consider a startup-time system check (or a `manage.py djust_audit` warning) when `max_message_size` is set below `DEFAULT_CHUNK_SIZE` plus the largest known frame header — this would have caught the problem the instant a project ran `manage.py check`, regardless of which settings dict it landed in.

## Testing

- A settings-loading unit test asserting `DJUST_CONFIG["max_message_size"]` alone does NOT change `LiveViewConfig().get("max_message_size")` (pins the current behavior so a future accidental "fix" that silently changes semantics is caught, until (1) above is decided and implemented deliberately).
- An integration test that calls `allow_upload(...)` with zero custom `LIVEVIEW_CONFIG`/`DJUST_CONFIG` and uploads a file larger than one default chunk (e.g. 100KB) — this should pass with the shipped defaults once (3) above lands; today it fails with `Message too large`.

Refs: `DJUST_LESSONS` gotcha #19 (downstream project field notes); `python/djust/config.py`; `python/djust/static/djust/src/15-uploads.js`; `python/djust/websocket.py`.

johnrtipton:
Fixes #1993 (two independent gotchas in one issue).

## (1) `DJUST_CONFIG` LiveView keys silently ignored
`LiveViewConfig._load_from_settings()` only read `LIVEVIEW_CONFIG`, so a `max_message_size` / `rate_limit` / `event_security` set in the similarly-named `DJUST_CONFIG` dict (which already backs tenancy/presence/state-backend, and is easy to confuse) was a **silent no-op** — raising the limit via `DJUST_CONFIG = {"max_message_size": 262144}` did nothing. Now falls back to `DJUST_CONFIG` for keys **present in the LiveView defaults** (so unrelated tenancy/presence keys aren't pulled in), with `LIVEVIEW_CONFIG` winning on a collision and a debug breadcrumb per adopted key. The misleading comment implying `DJUST_CONFIG` was already handled here is corrected.

## (2) Default upload chunk exceeds default frame limit by 21 bytes
`DEFAULT_CHUNK_SIZE` was `64 * 1024` — exactly the `max_message_size` default (65536) — and each chunk frame prepends a 21-byte binary header, so `65536 + 21 = 65557 > 65536`: a brand-new project using only `allow_upload(...)` with default settings failed any upload past a fractional first chunk with `Message too large (65557 bytes)`. Reduced to `63 * 1024` (64512 + 21 = 64533 < 65536); `buildFrame` now uses the named `FRAME_HEADER_BYTES` constant instead of the magic `21`.

## Tests
4 Python config tests (gate-off verified: reverting the fallback makes the `DJUST_CONFIG`-applied tests RED) + a JS source-invariant pin in `tests/js/uploads.test.js` (`DEFAULT_CHUNK_SIZE + FRAME_HEADER_BYTES ≤ 65536` — reverting to `64*1024` fails it). mypy strict-island clean; client.js rebuilt from src; 481-test config/upload regression green.


johnrtipton:
## Stage 14 retro — PR #2010 (Fixes #1993)

**Task**: v1.1.0-5 drain — DJUST_CONFIG LiveView-key fallback + default upload chunk exceeding the default frame limit.

### Quality: 4/5
### What went well
- Two genuinely independent root causes in one issue, both fixed: the silent-no-op config namespace (adopt only keys in the defaults, so tenancy/presence keys aren't pulled in; LIVEVIEW_CONFIG wins) and the 21-byte chunk-vs-frame arithmetic (`64*1024 + 21 > 65536`).
- Caught + fixed a tautology in the second config test at the implementer stage: `event_security` defaults to `"strict"` via validation, so testing the fallback with it passed even gated-off. Swapped to `css_framework` (unambiguous non-default), then re-ran the gate-off to confirm BOTH fallback tests go RED without the fix.

- First commit was blocked by the eslint gate: `FRAME_HEADER_BYTES` was declared for documentation but unused, and the pre-commit lints the built `client.js`. Fixed properly by *using* the constant in `buildFrame` (removed the magic `21`s), which is a net improvement. No bad commit landed (pre-commit aborted it).

### Verified
4 Python config tests (gate-off verified, non-tautological after the `css_framework` swap) + JS source-invariant pin (`DEFAULT_CHUNK_SIZE + FRAME_HEADER_BYTES ≤ 65536`; 18 uploads JS tests) · mypy strict-island clean · client.js rebuilt + eslint clean · 481-test config/upload regression green · CI green.

`RETRO_COMPLETE`

PR #2011, opened 2026-07-02
docs: correct presence-record shape + djust_markdown backend + dj-transition-group CSS notes (#2000, #2004)

### LINKED ISSUE #2000: docs: PresenceMixin's module docstring shows a flattened presence-record shape that doesn't match any backend
## Symptom (DJUST_LESSONS gotcha #13)

The module docstring at the top of `presence.py` shows this usage:

```python
def get_context_data(self):
    ctx = super().get_context_data()
    ctx["presences"] = self.list_presences()  # [{"id": ..., "name": ..., "color": ...}]
    ...
```

```django
{% for p in presences %}
  <span class="avatar" style="background: {{ p.color }}">{{ p.name.0 }}</span>
{% endfor %}
```

Following this example verbatim produces a `KeyError`/silently-empty template output: `p["name"]` and `p["color"]` don't exist. The actual field is `p["meta"]["name"]` / `p["meta"]["color"]`.

## Root cause (grep-verified, both shipped backends)

Every presence backend nests caller-supplied metadata under a `"meta"` key, not at the top level:

- `python/djust/backends/memory.py:46-52` (`MemoryPresenceBackend.join`):
  ```python
  record = {
      "id": user_id,
      "joined_at": time.time(),
      "meta": meta,
  }
  ```
- `python/djust/backends/redis.py:73-79` (`RedisPresenceBackend.join`): identical shape, `"meta": meta`.

Both backends' `.list()` (`memory.py:69-73`, `redis.py:106+`) return these records unchanged, and `PresenceMixin.list_presences()` (`presence.py:407-410`) delegates straight through with no flattening:

```python
def list_presences(self) -> List[Dict[str, Any]]:
    """Get all active presences for this view's presence group."""
    presence_key = self.get_presence_key()
    return PresenceManager.list_presences(presence_key)
```

So the nested-under-`meta` shape is universal across backends — this isn't a memory-vs-redis divergence, it's the class docstring (`presence.py:1-35`) disagreeing with 100% of the actual implementations. Every code sample in the docstring (the `get_context_data` example, the template snippet, and `handle_presence_join`'s `presence['name']` access) is wrong as written.

## Proposed scope

Two options, either resolves it:

1. **Fix the docstring** to show the real shape (`p.meta.name`, `p.meta.color`, `presence['meta']['name']` in `handle_presence_join`). Smallest, safest fix — zero behavior change, and keeping `meta` separate from framework-managed fields (`id`, `joined_at`) avoids a caller's `meta={"id": ...}` silently colliding with the record's own `id`.
2. **Flatten `meta` into the top-level record** in `PresenceManager.list_presences()` (or in each backend's `.list()`), so the docstring's `[{"id","name","color"}]` shape becomes real. This matches every example in the docstring without touching call sites, but risks a key collision if a caller's `meta` dict happens to contain `id`/`joined_at`/`meta` — would need a documented precedence rule (framework fields win) or a rename-on-collision warning.

Recommend (1): it's a one-file, zero-risk fix, and the backends already keep `meta` deliberately separate — that separation looks intentional (avoiding collision), so the docstring is the actual bug here, not the data shape.

## Testing

A single doctest-style assertion against `list_presences()`'s real return shape (already exercised by existing presence tests, if any check `.join()`'s return value) would have caught the docstring drift — recommend adding a comment near the docstring's example pointing at the real backend as the source of truth, so a future backend change to the shape doesn't let the docstring drift again silently.

### LINKED ISSUE #2004: docs: two small doc gaps — djust_markdown's TEMPLATES backend requirement, and dj-transition-group's CSS-authoring requirement not repeated where readers will actually copy from
Two small, independent documentation gaps found building a production app on djust 1.0.8, each cost real debugging time despite being one-line fixes to the docs.

### Symptom

`TemplateSyntaxError: Invalid block tag 'djust_markdown'` (HTTP 500) on the initial GET, even with `{% load live_tags %}` at the top of the template, on a project with a plain Django-only `TEMPLATES` setting (e.g. one created via `django-admin startproject` rather than a djust scaffold).

### Root cause

`djust_markdown` registers only with djust's Rust template engine via `register_tag_handler` — it is NOT a Django `simple_tag` (unlike `djust_client_config`/`live_render`, which happen to work under either engine). A LiveView's initial GET path resolves template inheritance through Django's own `loader.get_template()`, which compiles the whole `{% extends %}` chain and dies on the unknown tag the moment `DjustTemplateBackend` isn't the active `TEMPLATES` backend.

`docs/website/guides/streaming-markdown.md` — the guide's own quick-start (lines 18–22) goes straight to:

```django
{% djust_markdown llm_output %}
```

with zero mention of `TEMPLATES`, `DjustTemplateBackend`, or backend configuration anywhere in the file (confirmed: no hits for any of "TEMPLATES" / "DjustTemplateBackend" / "template_backend" / "backend" in `streaming-markdown.md`'s 221 lines). The guide's own example only works because the framework's demo project already has `DjustTemplateBackend` configured — a project following the guide's example verbatim, on a stock Django `TEMPLATES` setting, hits the 500 with no clue why from the guide itself.

The existing `djust.A090` system check (`python/djust/checks/templates.py:572-590`) is the closest thing to a safety net here, but it's purely informational — it just reports where `{% djust_markdown %}` is used and confirms the Rust `pulldown-cmark` backend's safety properties (`ENABLE_HTML` never set, `javascript:` URLs neutralised, 10 MiB input cap). It does not check whether `DjustTemplateBackend` is actually the active `TEMPLATES` backend, so it fires (or would fire, if the check even runs) with no bearing on this specific failure mode.

### Proposed fix

1. State the `DjustTemplateBackend` requirement up front in `streaming-markdown.md`'s quick-start, before the first code sample.
2. Extend `A090` (or add a new check) to detect "`{% djust_markdown %}` appears in a template but `DjustTemplateBackend` is not configured in `TEMPLATES`" and raise it as an error, not an info notice — this is a hard requirement, not a safety confirmation.

---

### Symptom

`dj-transition-group="fade-in | fade-out"` runs (correctly, as of the fix for #1273) but produces no visible animation until the app defines `.fade-in`/`.fade-out` itself with a real CSS `transition:` property. A reader who copies the short-form quick-start example (`docs/website/guides/declarative-ux-attrs.md:258-263`) has no indication, right there, that these are placeholder names requiring app-authored CSS rather than built-in animation presets.

### What's already documented, for accuracy

This is narrower than it first looks — `dj-transition`'s own section already states plainly (`declarative-ux-attrs.md:177-179`): "Works with any class-based CSS framework — Tailwind..., Bootstrap 5..., or hand-rolled classes. The attribute only orchestrates the class application; it doesn't ship any CSS itself." And `dj-remove`'s section has an explicit "Gotcha — no CSS transition defined" callout (lines 217-220) warning that a spec without a real `transition:` property means `transitionend` never fires and the 600ms fallback timer does the work instead.

So the disclaimer exists — it's just scoped to `dj-transition`'s own section and not repeated (or even cross-referenced) under `dj-transition-group`'s section, particularly not next to the short-form quick-start (lines 256–263) that most readers will copy from first, since it's visually the simplest example on the page. A reader who jumps straight to `dj-transition-group` (a very plausible entry point — it's the higher-level, "I have a list" attribute) can miss the disclaimer entirely.

Also worth noting for context (not part of this doc gap, already fixed): issue #1273 covered a real BUG in this same area — the short-form single-token spec (`"fade-in"`) used to be silently rejected by `_parseSpec`'s 3-token requirement, meaning no animation attribute was even applied. That's fixed (`41-dj-transition.js:79-85`, "Closes #1273" — 1-token specs are now accepted and drive a class-then-`transitionend` sequence same as `dj-remove`). This issue is about the DIFFERENT, still-open gap: the mechanism runs correctly, but nothing next to the short-form example reminds the reader that the class names still need real CSS behind them.

### Proposed fix

Add a one-line callout right after the short-form quick-start example in `dj-transition-group`'s section (`declarative-ux-attrs.md`, near line 263): "`fade-in`/`fade-out` are placeholder names — like `dj-transition` (see above), djust does not ship these classes; define the actual `@keyframes`/`transition` CSS in your own stylesheet." Even a single cross-reference sentence would close this gap.

## Refs

- DJUST_LESSONS gotcha #1 and #14 (this project's own field-notes doc)
- #1273 (fixed the short-form silent-rejection bug; this issue's Gap 2 is the separate, still-open doc-clarity gap left after that fix)

johnrtipton:
Closes #2000, closes #2004. Three copy-from-and-it-breaks documentation corrections, grouped (all pure docs / docstring).

## #2000 — `PresenceMixin` docstring shows a flattened record that matches no backend
The module docstring showed `{{ p.color }}` / `{{ p.name.0 }}` / `presence['name']`, but every backend nests the caller-supplied meta under a `"meta"` key — the record is `{"id", "joined_at", "meta": {...}}`. Following it verbatim produced a `KeyError` / silently-empty output. Corrected to `p.meta.name` / `presence['meta']['name']` and documented the record shape.

## #2004a — `{% djust_markdown %}` backend requirement never stated
The tag registers only with djust's Rust engine (`register_tag_handler`), not Django's stock backend; a plain `django-admin startproject` `TEMPLATES` setup raises `TemplateSyntaxError: Invalid block tag 'djust_markdown'` even with `{% load live_tags %}`. Now stated where readers copy the tag from (`streaming-markdown.md`).

## #2004b — `dj-transition-group` CSS-authoring caveat not repeated where readers copy
`dj-transition` says "it doesn't ship any CSS itself"; the `dj-transition-group` quick-start (where readers copy the `opacity-0`/`fade-in` class names) omitted it. Added.

Docs-only + one docstring; no code paths touched.


johnrtipton:
## Retrospective — PR #2011 (pipeline-drain)

**Task**: v1.1.0-5 — drained #2000 + #2004 (grouped pure-docs).

### Quality: 5
### What went well
- Grouping three copy-from-and-it-breaks doc corrections into one PR kept review cheap; all three share the same failure shape (reader copies an example that doesn't match reality).
- #2000 was verified against the real backend record shape (`backends/memory.py` — `{"id","joined_at","meta":{...}}`), not the docstring's claim — the docstring itself was the bug.
- #2004b mirrored the existing `dj-transition` "ships no CSS" caveat verbatim into the `dj-transition-group` quick-start where readers actually copy from.

### What didn't
- Nothing blocking. `check-doc-snippets.py` takes a `--guides-dir`, not individual files, so local per-file validation isn't possible — relied on the pre-commit hook (passed).

### Verified
- presence.py imports clean (docstring-only change); doc-snippet + version/size pre-commit hook passed; 13/13 CI checks green.
RETRO_COMPLETE

PR #2012, opened 2026-07-02
fix(state): re-hydrate private model attrs across the session round-trip as models, not dicts (#1994)

### LINKED ISSUE #1994: bug: a private attr holding a Django model instance can come back as a plain dict after a state round-trip (HTTP POST fallback path)
## Symptom (DJUST_LESSONS gotcha #20 and #21, hit in downstream production build)

```
AttributeError: 'dict' object has no attribute 'memberships'
  File ".../djust/mixins/request.py", line 630, in post -> self.get_context_data()
  File "chat/views.py", line 500, in get_context_data
    self._workspace.memberships.count() if self._workspace else 0
```

`self._workspace` was set once in a downstream app's `_activate_workspace()` (called from `mount()`) as a real `Workspace` model instance, cached on a private (`_`-prefixed) attribute. The traceback is **not** a WebSocket reconnect — it's the HTTP POST fallback path (`mixins/request.py`'s `post()`, restoring a previously-persisted view instance rather than calling `mount()` fresh). Somewhere in that persistence round-trip, the cached `Workspace` instance came back as a plain `dict` instead of the original model — breaking any code calling a method (`.memberships.count()`, `.name`) on it. `STATE_BACKEND`/`PRESENCE_BACKEND` were both set to `"redis"` when this was observed.

The app's own fix (unrelated to this report) was to stop caching the model instance at all and re-query fresh from `self.workspace_id` (a plain int) on every access — but the underlying framework behavior is still worth investigating, since it violates the type-fidelity guarantee a private attribute round-trip should have.

## What I verified (grep-checked against `python/djust/`)

- `mixins/request.py`'s `post()` (`~L579-630` in this checkout) restores `saved_state` from `request.session.get(view_key, {})` via `safe_setattr(self, key, value, allow_private=False)` (public state only), then separately restores `request.session.get(f"{view_key}__private", {})` via `self._restore_private_state(private_state)` — a plain `setattr(self, key, value)` per key (`live_view.py:767-778`), with **no type reconstruction**: whatever was stored is set back verbatim.
- The save side (`mixins/request.py:258-260`, and the equivalent block in `runtime.py`'s WS-reconnect `_save()` around L3217-3227) always goes through `self._get_private_state()` (`live_view.py:732-762`) first, THEN `normalize_django_value()`. `_get_private_state()`'s serializability gate is `json.dumps(value, cls=DjangoJSONEncoder)` — I confirmed empirically (`python -c "json.dumps(a_real_model_instance, cls=DjangoJSONEncoder)"`) that this **raises `TypeError`** for a raw Django model instance (plain `DjangoJSONEncoder` has no model support), which per the `except (TypeError, ValueError, OverflowError)` branch should cause the key to be **silently skipped**, not persisted at all — the model should never reach the session `__private` key as a model OR a dict via this path.
- BUT `normalize_django_value()` itself (`serialization.py:817-877`, applied to whatever `_get_private_state()` returns) explicitly documents wider support: *"Django Model -- serialized via DjangoJSONEncoder._serialize_model_safely, then recursed"* — i.e., a DIFFERENT, more permissive serializer than the one gating `_get_private_state()`. This is a real, confirmed **divergence between two "is this serializable" checks** used in the same persistence pipeline: one (`_get_private_state`'s plain `json.dumps(..., cls=DjangoJSONEncoder)`) rejects a bare model, the other (`normalize_django_value`) explicitly knows how to turn one into a dict. Depending on exactly which code path a given `_private` attribute travels through, it can be silently dropped by one gate or silently converted to a dict by the other — the app-visible outcome differs (missing attribute vs. corrupted-type attribute) depending on which.
- I was **not able to fully reproduce, in the time available, the exact path that produced a dict specifically** (rather than the attribute being entirely absent, which is what my reading of `_get_private_state()`'s gate alone would predict). `STATE_BACKEND="redis"` was active on the reporting app; `python/djust/state_backends/redis.py` stores an entire view snapshot via a separate, non-JSON binary path (its own comment says "5-10x faster than pickle") that appears independent of `_get_private_state()`/`normalize_django_value()` entirely — this looks like the more likely place a model instance could survive a round-trip as *something*, but I did not trace it fully. Recommend checking whether the Redis `STATE_BACKEND` snapshot mechanism and the Django-session `__private` mechanism can cross-contaminate on a view that uses both (WS reconnect via Redis, HTTP fallback via session) — that seam is the most likely source of the type-degrading dict.
- Related prior issues in this repo, closed under 0.4.0: **#611** ("pre-rendered WS reconnect drops `_private` attributes, skipping `mount()`") and **#612** ("dict state attributes deserialized as list after Rust state sync", referencing #611). Both are the same *family* of bug — state round-trip failing to preserve a private attribute's original shape/type — but neither is this exact model→dict case (#611 is about attributes going missing when `mount()` is skipped; #612 is about a dict becoming a list via the Rust `update_state` sync path, not the Django-session path traced above). Worth checking whether the #611/#612 fix generation already covers this, or whether this is a distinct third instance of the same underlying class of bug recurring in a different subsystem.

## Sibling risk (DJUST_LESSONS gotcha #21, unconfirmed, framework-internal)

While investigating the above, found the daphne log continuously emitting (independent of hot-reload activity):
```
WARNING LiveView state contains non-serializable value: TenantInfo (from djust.tenants.resolvers).
This will be converted to a string, which may cause AttributeError on deserialization.
```
`self.tenant` is a `TenantMixin` property (`python/djust/tenants/mixin.py`) backed by `self._tenant` — itself a private attribute holding a `TenantInfo` object, which is not JSON-serializable. `_ensure_tenant()` guards re-resolution with `if self._tenant_resolved: return` (a plain bool, which round-trips fine) — if `_tenant` degrades on a bad restore while `_tenant_resolved` stays `True`, `_ensure_tenant()` would never re-fetch, leaving `self.tenant` broken for the rest of that session. This was **not confirmed** to have actually crashed anywhere (no `AttributeError` referencing `tenant` found in the reporting app's logs), but it's the identical architectural shape as the confirmed bug above, this time in framework-internal code rather than app code — worth a look at the same time, since a fix to the general state-restore-type-fidelity problem should probably cover this too. Suggest `_ensure_tenant()` validate `isinstance(self._tenant, TenantInfo)` before trusting `_tenant_resolved`, not just check the boolean.

## Suggested scope

1. Either (a) make private-attribute state restoration preserve the original type of a value across every persistence path (Django session AND Redis `STATE_BACKEND`), or (b) if a value can't be faithfully round-tripped, refuse to restore it and warn loudly — the same way the framework's own "non-serializable value ... will be converted to a string" warning already does for attributes it can't handle at all. A silently type-degrading restore (present but wrong type) is a worse failure mode than a silent drop (absent, at least fails predictably) — it lets code run further before crashing somewhere unrelated to the actual defect.
2. Unify (or explicitly reconcile) the two serializability checks in the persistence pipeline (`_get_private_state`'s `json.dumps(..., cls=DjangoJSONEncoder)` vs. `normalize_django_value`'s model-aware handling) so a given attribute's fate (dropped vs. type-converted) doesn't depend on which of the two happens to see it first.
3. Investigate whether the Redis `STATE_BACKEND` (`state_backends/redis.py`) and the Django-session-based HTTP-fallback path (`mixins/request.py`) can interact on a view using both, since that's the strongest lead this investigation found for where a model instance could survive a round-trip as a dict rather than being dropped.
4. Apply the same audit to `TenantMixin._tenant`/`_ensure_tenant()` (`python/djust/tenants/mixin.py`) — confirmed structurally at risk of the identical bug, not yet confirmed to have crashed.

## Testing

A regression test for this would need to either (a) genuinely reproduce the Redis `STATE_BACKEND` + HTTP-fallback round-trip that produced the dict (not attempted here — I couldn't pin down the exact trigger in the time available), or (b) at minimum assert that `_get_private_state()` and `normalize_django_value()` agree on whether a given value type is safely round-trippable, so the "drop vs. convert" divergence documented above can't recur for a new type in the future.

johnrtipton:
Closes #1994.

## Symptom
A Django model cached on a private (`_`-prefixed) LiveView attr — `self._workspace = Workspace.objects.get(...)` in `mount()` — came back as a plain `dict` after the HTTP-POST-fallback restore path, so `self._workspace.memberships` raised `AttributeError` (reporter traceback at `mixins/request.py` `post()`).

## Root cause
Private state is persisted to the session so it survives the POST-fallback restore (which does NOT re-run `mount()`). The persist path ran `normalize_django_value` — the **client-facing** serializer — over private state, turning the model into the lossy `{"pk", "__str__", <fields>}` dict. On restore, `_restore_private_state` set that dict back.

## Fix (structural, #1646-safe)
Private state is server-side view cache, never sent to the client, so the serialization floor doesn't apply. Encode each model as a re-hydratable ref `{"__djust_model_ref__": "<app>.<model>", "pk": ...}` (recursing into nested dicts/lists) inside `_get_private_state()`, and re-fetch it from the DB inside `_restore_private_state()`. Because both are the single funnel that **all** persist/restore paths call — HTTP GET/POST (`request.py`), runtime WS (`runtime.py`), sticky-child (`sticky.py`) — the fix covers every path with no drift. `normalize_django_value` for the client-bound `_session_state` is untouched.

A ref whose row was deleted between save and restore re-hydrates to `None` with a warning (a stale cached model must not hard-crash a reconnect).

## Tests (`test_private_model_roundtrip_1994.py`, 5)
model-comes-back-as-model (gate-off sentinel — 3 tests go RED when the encode/decode is neutered), nested-in-dict, in-list, deleted→None, non-model attrs unaffected. Full private-state/session regression (124) green.


johnrtipton:
## Retrospective — PR #2012 (pipeline-drain)

**Task**: v1.1.0-5 — #1994 (private model attr comes back as a dict after state round-trip).

### Quality: 5
### What went well
- Root cause traced symptom-up: the `AttributeError` at `mixins/request.py post()` pointed at the restore, but the real defect was the persist path running `normalize_django_value` (the client-facing serializer) over server-side private state.
- **Structural #1646-safe cure**: the encode/decode live INSIDE the two shared funnels (`_get_private_state` / `_restore_private_state`), so all 5 persist paths (HTTP GET/POST, runtime WS, sticky-child sync/async) and all 3 restore paths inherit the fix with zero drift — verified by grepping every call site.
- Gate-off verified: 3 of 5 tests go RED when the encode/decode is neutered.

### What didn't
- Nothing blocking. A private *QuerySet* attr (vs single model) stays lossy — out of scope for the cited bug, and no worse than before.

### Verified
- 5 new tests + 124 private-state/session regression green; `mypy python/djust` clean (the one single-file error is pre-existing on main); gate-off RED confirmed; 13/13 CI green.
RETRO_COMPLETE

PR #2013, opened 2026-07-02
fix(events): warn (debug) on .lazy/.debounce suffix mis-applied to non-dj-model directives (#1999)

### LINKED ISSUE #1999: bug: `dj-input.debounce-N` silently fails to bind — modifier-suffix syntax works for `dj-model` but not `dj-input`
## Symptom (DJUST_LESSONS gotcha #12, hit in downstream production builds)

A live-search input written as:

```html
<input dj-input.debounce-200="palette_search" value="{{ palette_query }}">
```

never fires — the search handler is never called, the box just sits there returning
unfiltered results. No console error, no warning, nothing to indicate the attribute
was even parsed and rejected. The developer's mental model ("`.debounce-N` is how
djust spells a debounced live input") is reinforced by `dj-model.debounce-300`
working exactly that way — so this reads as a plain typo-free feature that simply
doesn't work, with zero diagnostic signal pointing at the real cause.

The only way to get a debounced `dj-input` is the differently-shaped, two-attribute
form:

```html
<input dj-input="palette_search" dj-debounce="200" value="{{ palette_query }}">
```

## Root cause (grep-verified against source)

`dj-model` and `dj-input` implement debounce/lazy support via two structurally
different mechanisms, and only one of them understands the `.modifier-N` suffix
convention:

- **`dj-model`** (`python/djust/static/djust/src/20-model-binding.js`) parses its
  own attribute NAME for modifiers. `_parseModelAttr()` (L27-55) iterates
  `el.attributes` and checks `name === 'dj-model.lazy'` (L43) and
  `name.startsWith('dj-model.debounce')` (L47, matched against
  `/debounce-?(\d+)/` at L50) — i.e. it explicitly recognizes the
  attribute-name-suffix spelling. The binding entry point's selector,
  `root.querySelectorAll('[dj-model], [dj-model\\.lazy], [dj-model\\.debounce]')`
  (L134), only literal-matches those two named variants directly, but the
  broader per-element modifier detection at L31-55 is what actually
  recognizes any `dj-model.debounce-<N>` spelling once an element is reached.

- **`dj-input`** (`python/djust/static/djust/src/09-event-binding.js`) has NO
  attribute-name modifier parsing at all. Every entry point that finds
  `dj-input` elements uses an EXACT CSS attribute-name selector:
  `e.target.closest('[dj-input]')` (L978), `form.querySelectorAll('[dj-input]')`
  (L1337), and the mixed `dj-change`/`dj-input` field scan
  (`input[dj-input], textarea[dj-input], select[dj-input]`, L1571). In HTML, a
  dot is a legal (if unusual) character in an attribute name — `dj-input.debounce-200`
  is parsed as a single, literal attribute named `dj-input.debounce-200`, which
  does **not** match the `[dj-input]` selector (an exact attribute-name match, not
  a prefix match). So the element is invisible to every `dj-input` binding path:
  no handler attaches, no event ever fires, and there's no code path left to emit
  a warning from, because as far as the framework's selectors are concerned the
  element simply has no `dj-input`-family attribute on it at all.

Instead, `dj-input`'s own throttle control is a **separate** attribute,
`dj-debounce` (handled at L998-1041 inside the `input` delegation handler, and
again for the general case at `applyDebounceThrottle()`, L1285-1314) — a
completely different spelling convention from `dj-model`'s in-name modifier.

## Scope (proposed)

Either direction closes the gap; a warning is cheap insurance regardless of which
(or neither) ships:

1. **Unify the spelling** — accept `.debounce-N`/`.lazy` as a modifier suffix on
   `dj-input` too (mirroring `dj-model`'s `_parseModelAttr`-style name parsing), so
   both attributes support the same modifier convention.
2. **OR document the divergence prominently** in the model-binding / dj-input
   guides, with a side-by-side example showing the two different debounce
   spellings for `dj-model` vs `dj-input`.
3. **In addition to either of the above**, emit a `console.warn` (debug-mode gated,
   consistent with the existing `globalThis.djustDebug` pattern used elsewhere —
   e.g. `08-event-parsing.js:386`) whenever an element carries an attribute whose
   name starts with `dj-` and contains a `.` but doesn't match any recognized
   `dj-<name>.<modifier>` shape. That single generic check would have made this
   exact bug visible immediately instead of silent, and covers any future
   spelling mix-ups between `dj-model`/`dj-input`/other directives without
   requiring the two conventions to be unified.

## Testing note (not itself a bug, but relevant to reproducing this)

`dj-input` listens for genuine user `input` events; dispatching a synthetic
`new Event('input')` (e.g. via a headless test harness) does **not** trigger it.
Any regression test for this needs to drive real keystrokes (or otherwise dispatch
a trusted-equivalent input event), not a bare synthetic dispatch, or a passing test
here would give false confidence that binding is working when it may not be.

## Environment

- djust version: 1.0.8
- Found via: production app development (chat UI), command-palette live search
  (`dj-input.debounce-200="palette_search"`) silently never filtering results.

Refs: DJUST_LESSONS gotcha #12 (downstream project's own field-notes log).

johnrtipton:
Closes #1999.

## Symptom
`<input dj-input.debounce-200="palette_search">` silently never fires — no handler, no event, no console error. `dj-model.debounce-300` working exactly that way makes it read as a plain non-working feature.

## Root cause
Only `dj-model` parses the `.lazy` / `.debounce-N` in-name modifier from its attribute *name* (`20-model-binding.js` `_parseModelAttr`). `dj-input` / `dj-change` / `dj-click` debounce via the **separate standalone `dj-debounce="N"`** attribute and parse no in-name modifier. A dot is a legal attribute-name character, so `dj-input.debounce-200` is one literal attribute that no `[dj-input]` selector matches → the input is invisible to every binding path, and there's no code left to warn from.

## Fix (option 2 + 3 from the issue; unify deferred)
Unifying would require surgery on the hot `[dj-input]` delegation path in a top-3-budget module and fights morphdom (SSR re-emits the dotted attr). Instead:
- **Warn (debug-gated):** `bindLiveViewEvents` runs `_warnUnrecognizedDjModifiers` — zero cost outside `window.djustDebug` — emitting a `console.warn` that names the offending attribute and the standalone-`dj-debounce` fix. Scoped to `.lazy` / `.debounce` on non-`model` directives; **deliberately does not** touch other legit dotted conventions (`dj-keydown.enter`, `dj-window-keydown.escape`, `dj-loading.class`/`.show`/`.hide`/`.disable`/`.for`).
- **Document:** the `dj-model` guide now shows the two debounce spellings side-by-side.

+315 B gzipped (well under the 2 KB/module budget).

## Tests (`tests/js/dj-input-modifier-warning-1999.test.js`, 7)
warns for `dj-input.debounce-200` / `dj-input.lazy` / `dj-change.debounce-100` (3 gate-off sentinels — RED when the scan is neutered); does NOT warn for `dj-model.*`, for key/loading dotted modifiers, or for plain `dj-input` + `dj-debounce`; debug-gated (silent when off).


johnrtipton:
## Retrospective — PR #2013 (pipeline-drain)

**Task**: v1.1.0-5 — #1999 (`dj-input.debounce-N` silently never binds).

### Quality: 5
### What went well
- Discovered mid-implementation that the issue's proposed "generic warn on any dotted dj- attr" would false-positive on **legit** dotted conventions (`dj-keydown.enter`, `dj-window-keydown.escape`, `dj-loading.class/.show/.hide/.disable/.for`) — grepped the source first (#168) and tightened the check to key on the `.lazy`/`.debounce` *modifier*, not "any dot".
- Chose doc+warn over unify (#1079): unifying would need surgery on the hot `[dj-input]` delegation in a top-3-budget module and fights morphdom (SSR re-emits the dotted attr).
- Gate-off verified: 3 of 7 tests go RED when the scan is neutered.

### What didn't
- First commit bounced on eslint `security/detect-unsafe-regex` — my `(-\d+)?` nested quantifier is star-height 2. Simplified to a prefix match (star-height 1); no capture of N was needed anyway.

### Verified
- 7 JS tests (gate-off RED confirmed); eslint clean on the built bundle; +315 B gzipped (well under the 2 KB/module budget); 13/13 CI green.
RETRO_COMPLETE

Issue #1516, opened 2026-05-18
tech-debt: reviewer-subagent prompt needs an environment-premises brief (recurring two-commit-shape false positive)


johnrtipton:
Source: Retro v1.0.0rc2 (finding #3 — milestone synthesis). **OUT-OF-REPO** — the reviewer-subagent prompt lives in the pipeline-run skill repo (`~/.claude/skills/`), not the djust framework repo.

Across the v1.0.0rc2 drain, the Stage-7 self-review subagent flagged "CHANGELOG missing" as a REVIEW_FAILED finding on PR #1504 — a guaranteed **false positive**: under the two-commit shape (Action #181) Stage 5 is forbidden from editing CHANGELOG.md, which lands in the Stage-9 docs commit. Stage 7 runs before Stage 9, so the absent CHANGELOG entry is expected. The orchestrator hand-briefed every subsequent reviewer subagent in the drain to suppress it.

The durable fix is structural: the Stage-7 reviewer-subagent prompt template (and the implementer-subagent prompt) should carry an explicit **"Environment premises"** section stating the invariants a subagent must NOT assume away:
- CHANGELOG.md is intentionally deferred to the Stage-9 docs commit (two-commit shape).
- The branch base may be stale vs origin (#1450) — verify before reviewing.
- `.claude/` is gitignored repo-wide; target files may be untracked.

This generalizes the per-PR-retro "common thread": tasks 1-3 of the v1.0.0rc2 drain each lost time to a subagent acting on an unverified environment premise. Stating the premises in the prompt closes the class.

Out-of-repo: implement in the pipeline-run skill repository; not committable to the djust framework repo.

johnrtipton:
Resolved upstream in pipeline-skills@f3203b2 (canon commit e25a213, merged to main). `pipeline-shared/SKILL.md` Code Review stage gained an "Environment premises" subsection (CHANGELOG may be two-commit-deferred; base may be stale; some paths are gitignored) with an instruction to include it in the reviewer subagent prompt.

PR #2014, opened 2026-07-02
fix(tenants): add set_tenant() helper + document session-resolver WS persistence semantics (#2003)

### LINKED ISSUE #2003: docs: `session` tenant resolver's WebSocket persistence semantics are undocumented; no `set_tenant()` helper
## Symptom (DJUST_LESSONS gotcha #11, hit adapting an existing FK to `djust.tenants`)

Adapting `TenantScopedMixin` + the `session` `TenantResolver` to an existing `workspace_id` FK (rather than a purpose-built tenant model) surfaced two frictions neither the guide nor the resolver's own docstring calls out:

1. **First-mount chicken-and-egg.** With `TENANT_REQUIRED=True` (the default) and a fresh session (no `tenant_id` yet), the view 404s *before `mount()` ever runs* — so there is no way for `mount()` itself to resolve a sensible default tenant (e.g. the user's first membership) when required is left at its default.
2. **Session writes from a handler have no persistence guarantee over the WS lifecycle.** Nothing in `djust.tenants` ever writes to `request.session` — `SessionResolver.resolve()` is read-only. An app that wants "switch tenant" to stick has to write `request.session[...]` itself from inside an event handler, but there's no corresponding HTTP response for Django's `SessionMiddleware` to save against at that point, so the write is silently best-effort.

## Root cause (grep-verified against the local checkout)

**(1) Dispatch order.** `TenantMixin.dispatch/get/post` all call `self._ensure_tenant(request)` *before* delegating to `super()` (`python/djust/tenants/mixin.py:186-199`), and `_ensure_tenant` raises `Http404` when unresolved and required (`mixin.py:110-127`, the raise at `:127`). `_is_tenant_required()` (`mixin.py:129-138`) checks the view's `tenant_required` attribute or `DJUST_CONFIG["TENANT_REQUIRED"]` (default `True` — `python/djust/tenants/__init__.py:45`, `mixin.py:29`). Since `mount()` is invoked further down that same `dispatch()`/`get()`/`post()` chain, a required-but-unresolved tenant means `mount()` never gets a chance to establish one itself.

**(2) `SessionResolver` never writes.** `python/djust/tenants/resolvers.py:231-269` — `SessionResolver.resolve()` only ever *reads* `request.session.get(session_key)` (`:249`), falls back to a JWT claim (`:255-260`) or `request.user.tenant_id` (`:263-267`), and returns `None` if none match. There is no write-back method on `SessionResolver` or `TenantResolver` (`resolvers.py:89`, the abstract base) at all. Confirmed via `grep -rn "def set_tenant" python/djust/tenants/` — no matches anywhere in the package. Separately, `grep -rn "session.save\|session.modified" python/djust/live_view.py python/djust/websocket.py` — no matches — djust's own LiveView/WS-consumer code never calls `request.session.save()` on a handler-triggered write, consistent with there being no guaranteed save point once you're past the initial HTTP mount request.

## Scope

1. **Document the semantics.** The `session` resolver guide/docstring should state plainly: (a) resolution is read-only — the resolver never persists anything back to the session itself; (b) any `request.session[key] = value` write performed from inside a WS event handler has no built-in save guarantee, since there's no HTTP response cycle for `SessionMiddleware` to save against at that point; (c) the robust pattern is to keep the tenant id in **view state** (a plain `self.<field>` attribute) as the authoritative scoping value for every query, and treat any session write purely as a mirror for other consumers of the session key (e.g. a future page load, or middleware that only runs on real HTTP requests) — not as the source of truth during the WS lifecycle.
2. **Offer a first-class `set_tenant(tenant_id)`** on `TenantMixin`/`TenantScopedMixin` that a handler can call to update `self._tenant`/`self.tenant` (and, best-effort, mirror the write into `request.session` if a session resolver is configured) in one step, instead of every app hand-rolling `TenantInfo(...)` construction + a manual session write like this project had to.
3. **Consider surfacing the `TENANT_REQUIRED`-before-`mount()` ordering explicitly** in the tenancy guide's "adapting an existing tenant FK" section (if one exists) or adding one — the natural instinct when adapting an existing FK is to resolve a sensible default inside `mount()`, and the guide should say up front that `tenant_required=False` plus manual resolution in `mount()` is the correct pattern when there's no dedicated pre-mount tenant-resolution hook.

Source: `docs/djust-learnings.md` gotcha #11 in the `djust-chat` downstream project (a multi-model LLM chat client built on djust), where `TenantScopedMixin` is adapted to an existing `Workspace` model's FK rather than a purpose-built tenant model.

johnrtipton:
Closes #2003.

## Root cause

Two undocumented frictions with the `session` tenant resolver made WS-time tenant switching a footgun:

1. **`SessionResolver` is read-only + there is no `set_tenant` helper.** `SessionResolver.resolve()` (`python/djust/tenants/resolvers.py`) only *reads* `request.session[TENANT_SESSION_KEY]` (falling back to a JWT claim, then `request.user.tenant_id`); it never writes back, and `grep -rn "def set_tenant" python/djust/tenants/` was empty. So switching tenant from a WS event handler meant hand-writing `request.session[...]` — which has **no built-in save guarantee**: Django's `SessionMiddleware` persists the session against an *HTTP response*, and a LiveView event arrives over the WebSocket, so there is no response for the middleware to save against.
2. **`TENANT_REQUIRED` runs before `mount()`.** `TenantMixin.dispatch()/get()/post()` call `_ensure_tenant(request)` **before** `super()`, and `_ensure_tenant` raises `Http404` when the tenant is unresolved and required (`tenant_required` attr, or `DJUST_CONFIG['TENANT_REQUIRED']`, default `True`). So with a fresh session + the default config, the view 404s before `mount()` ever runs — `mount()` can't resolve a default.

## Fix

- **`TenantMixin.set_tenant(tenant_id)`** (`python/djust/tenants/mixin.py`), inherited by `TenantScopedMixin`. It:
  - updates the **authoritative** in-memory view state (`self.tenant`), and
  - **best-effort mirrors** the id into `request.session[TENANT_SESSION_KEY]` **only when a session resolver is configured** (`'session'`, or a chained resolver including `'session'`) — a **no-op** for subdomain/path/header/custom. Accepts a `str`/other id (coerced to `str`) or a `TenantInfo`. Logs at debug, no bare excepts.
- **Docs**: `SessionResolver` class docstring + the Multi-Tenant guide (`docs/website/guides/multi-tenant.md`) now document (a) resolution is read-only; (b) a WS-handler session write has no save guarantee; (c) the robust pattern is view-state-authoritative + session-write-as-mirror; and the `TENANT_REQUIRED`-before-`mount()` ordering with `tenant_required=False` + manual resolution in `mount()` as the fresh-session pattern.

## Tests

`python/djust/tests/test_tenant_set_tenant.py` — 12 cases:

| Case | Asserts |
|------|---------|
| `test_updates_view_state` | `self.tenant` set + `_tenant_resolved`, returns the `TenantInfo` |
| `test_mirrors_into_session_for_session_resolver` | session resolver → `request.session` mirrored |
| `test_mirrors_with_custom_session_key` | honors `TENANT_SESSION_KEY` |
| `test_mirrors_for_chained_resolver_containing_session` | chained `['header','session']` mirrors |
| `test_noop_on_session_for_non_session_resolver` | subdomain → session untouched |
| `test_noop_on_session_for_header_resolver` | header → session untouched |
| `test_no_request_session_is_safe` | missing request/session never raises |
| `test_coerces_non_str_id_to_str` | `42` → id `"42"` |
| `test_accepts_tenantinfo_instance` | preserves passed `TenantInfo` |
| `test_overrides_previously_resolved_tenant` | overrides prior resolved tenant |
| `test_scoped_mixin_has_set_tenant` | `TenantScopedMixin` exposes `set_tenant` |
| `test_scoped_queryset_uses_set_tenant_value` | `get_tenant_queryset()` filters on the new id |

**Gate-off (#1468)**: neutering `set_tenant`'s body (`return tenant` before the view-state update + mirror) turned all 12 RED — notably `test_mirrors_into_session_for_session_resolver` and `test_noop_on_session_for_non_session_resolver`, the behavior-defining cases. Restored; 12/12 green. Existing tenant suite (`tests/unit/test_tenants.py` + WS-host tenant + lazy-import) stays green (139 passed together).


johnrtipton:
## Retrospective — PR #2014 (pipeline-drain)

**Task**: v1.1.0-5 — #2003 (`session` tenant resolver WS persistence undocumented; no `set_tenant()` helper). Implemented by a worktree-isolated subagent, reviewed read-only via `gh pr diff` (avoiding the #1804 core.bare hazard).

### Quality: 5
### What went well
- `set_tenant()` makes the authoritative in-memory view-state (`self.tenant`) update the source of truth, and mirrors into `request.session` **only** when a session resolver is configured (direct or chained) — a clean `_session_resolver_configured()` guard, no-op for subdomain/path/header/custom.
- Docstrings state the WS-write-has-no-save-guarantee semantics plainly; the tenancy guide documents the `TENANT_REQUIRED`-before-`mount()` ordering + the `tenant_required=False` + manual-resolution pattern.
- Gate-off verified: neutering `set_tenant`'s body turned all 12 tests RED (incl. the two behavior-defining mirror/no-op cases); 139 tenant tests green together.

### What didn't
- The brief's env premise was slightly off — pyproject `pythonpath` is `[".", "examples/demo_project"]`, so `import djust` resolved to the main checkout; the subagent corrected by prepending `PYTHONPATH=<worktree>/python` and verified `djust.__file__` pointed into the worktree. (Env-premise-verification, #1516.)

### Verified
- Read-only diff review: `_session_resolver_configured` handles direct + chained; `%s`-logging; no `mark_safe`/bare-except/`print`; docstrings accurate. 13/13 CI green.
RETRO_COMPLETE

PR #2015, opened 2026-07-02
fix(events): re-scan dj-window-*/dj-document-* scoped elements after patches (#1996)

### LINKED ISSUE #1996: bug: `dj-window-*`/`dj-document-*` never bind on content that appears via a later patch
## Symptom (DJUST_LESSONS gotcha #15, hit in downstream production builds)

A `dj-window-keydown.escape="close"` (or any `dj-window-*`/`dj-document-*` attribute) on an
element that enters the DOM via a server-driven patch — e.g. inside a `{% if %}` block that
becomes true after a re-render, such as a command palette or an inline editor — **never fires**.
No console warning, no exception; the handler is just silently dead. Confirmed in a downstream
app: palette and editor Esc handlers built on `dj-window-keydown` never worked at all.

The documented workaround is to use `dj-shortcut` on the element instead, which correctly
re-binds after every patch (`escape`/`enter`/`arrowup`/`arrowdown`/chords, `dj-shortcut-in-input`
to stay live while typing) — but nothing signals that `dj-window-*`/`dj-document-*` don't share
that re-binding behavior.

## Root cause (grep-verified against `python/djust/static/djust/src/09-event-binding.js`)

`_scanScopedElements()` (`09-event-binding.js:128-192`) is the only code that populates the
scoped-listener registry (`_scopedRegistry`, a `Map` keyed on `"<prefix>:<eventType>"`) by
walking every element under the LiveView root and matching `dj-window-*`/`dj-document-*`
attribute names. It is called from exactly one place: `_installScopedDelegation()`
(`09-event-binding.js:202-207`), which itself is guarded by a one-shot flag:

```js
function _installScopedDelegation() {
    if (_scopedDelegationInstalled) return;
    _scopedDelegationInstalled = true;

    // Scan DOM once at install time to populate the registry
    _scanScopedElements();
    ...
```

`bindLiveViewEvents(scope)` (`09-event-binding.js:1096`) — the function called after every
WebSocket patch — DOES call `_installScopedDelegation()` again on every invocation
(`09-event-binding.js:1155`), but that call is a guaranteed no-op after the first mount because
of the `_scopedDelegationInstalled` flag above. `bindLiveViewEvents` also calls
`_sweepOrphanedScopedListeners()` right after (`09-event-binding.js:1159`, defined at line 81),
which removes registry entries for elements that left the DOM — but nothing symmetrically *adds*
entries for elements that just *entered* the DOM. Note the misleading comment directly above
`_scanScopedElements`'s definition (`09-event-binding.js:118-119`):

```js
// Install ONE listener per event type on window/document. When the event fires,
// dispatch to registered declaring elements. Elements are registered by scanning
// the DOM once at install time. Re-scanning happens on TurboNav navigation or
// when bindLiveViewEvents() is called after DOM changes.
```

This comment is incorrect for the `bindLiveViewEvents()` case — grepped every call site of
`_scanScopedElements()` in the file and there is exactly one (`09-event-binding.js:207`, inside
`_installScopedDelegation`'s one-shot guard). `bindLiveViewEvents` never calls
`_scanScopedElements()` directly, so "re-scanning ... when bindLiveViewEvents() is called after
DOM changes" does not happen — the comment describes intended behavior, not actual behavior.

By contrast, `dj-shortcut`'s binding block inside the very same `bindLiveViewEvents` function
(`09-event-binding.js:1185-1186`) does a **fresh `document.querySelectorAll('[dj-shortcut]')`
every time it's called**, gated only by a per-element `_isHandlerBound` check (so it's cheap and
idempotent, not a full re-scan cost problem) — which is exactly why `dj-shortcut` correctly picks
up new elements after a patch while `dj-window-*`/`dj-document-*` do not.

## Proposed scope

1. Have `bindLiveViewEvents()` call `_scanScopedElements()` directly (not gated behind the
   one-shot `_installScopedDelegation` flag) on every invocation, mirroring the pattern already
   used for `dj-shortcut` at line 1185 — `_scanScopedElements` is already idempotent per-element
   (see its own `alreadyRegistered` check at `09-event-binding.js:176-184`), so calling it
   repeatedly is safe and cheap; only `_installScopedDelegation`'s window/document
   `addEventListener` calls need the one-shot guard.
2. Fix the stale comment at `09-event-binding.js:118-119` to match whichever behavior ships.
3. Add a regression test: mount a view with a `dj-window-keydown.escape` inside a server-toggled
   `{% if %}` block, flip it true via a patch, dispatch a synthetic `Escape` keydown on `window`,
   and assert the handler fires.

## Secondary, smaller finding from the same investigation: shifted-character chords need the shifted key spelled out

`dj-shortcut`'s key comparison normalizes named keys (`escape`, `enter`, `arrowup`, etc.) via
`_normalizeKeyName()` (`09-event-binding.js:94-108`), but falls through to the raw string for
anything not in its small keymap:

```js
function _normalizeKeyName(name) {
    const lower = name.toLowerCase();
    const keyMap = { 'escape': 'Escape', 'enter': 'Enter', ... };
    return keyMap[lower] || name;
}
```

So `shift+/:help` can never fire — with Shift held, the browser reports `e.key === "?"`, not
`"/"`, and the binding is matched against the literal (lowercased) key text. `shift+?:help` is
the correct binding. This isn't necessarily a bug (browsers genuinely report the post-Shift
character in `e.key`), but it's a non-obvious footgun for anyone binding a shifted-character
shortcut. Suggest a documentation note in the shortcut-binding guide with an explicit
shifted-character example (`shift+?` rather than `shift+/`), and/or a debug-mode console warning
when a chord's key segment is a bare unshifted symbol commonly confused with its shifted
counterpart (`/` vs `?`, `1` vs `!`, etc.) alongside a `shift+` modifier.

## Testing

- Regression test for the re-scan fix (see above): patch-in `dj-window-*` content, assert the
  handler fires without requiring a subsequent full page load.
- Existing `dj-shortcut` behavior should stay covered by whatever suite already exercises
  `bindLiveViewEvents`'s per-call `querySelectorAll('[dj-shortcut]')` path — no proposed change to
  that path itself, just documenting the shifted-key gotcha.

Refs: DJUST_LESSONS gotcha #15 (downstream project's `docs/djust-learnings.md`).

johnrtipton:
Closes #1996.

## Symptom

A `dj-window-keydown.escape="close"` (or any `dj-window-*` / `dj-document-*` attribute) on an element that entered the DOM via a server-driven patch — e.g. inside a `{% if %}` that became true, such as a command palette or inline editor — **never fired**. No console warning, no exception; the handler was silently dead. The documented workaround was to switch to `dj-shortcut` (which re-binds after every patch), but nothing signaled that `dj-window-*` / `dj-document-*` didn't share that behavior.

## Root cause

In `python/djust/static/djust/src/09-event-binding.js`:

- `_scanScopedElements()` is the **only** code that populates the scoped-listener registry (`_scopedRegistry`, keyed `"<prefix><eventType>"`) by walking elements under the LiveView root matching `dj-window-*` / `dj-document-*` attribute names.
- It was called from exactly one place: `_installScopedDelegation()`, which is guarded by a one-shot flag (`_scopedDelegationInstalled`) and returns early after the first call.
- `bindLiveViewEvents()` (run after **every** WebSocket patch) re-invoked `_installScopedDelegation()`, but that's a no-op after first mount, and nothing else re-scanned. `_sweepOrphanedScopedListeners()` symmetrically **removed** registry entries for elements that *left* the DOM — but nothing **added** entries for elements that just *entered* via a patch.
- The comment above `_scanScopedElements` claimed re-scanning already happened when `bindLiveViewEvents()` was called after DOM changes — describing intended, not actual, behavior.

## Fix

`bindLiveViewEvents()` now calls `_scanScopedElements()` on **every** invocation (mirroring the per-bind `querySelectorAll` rescan that `dj-shortcut` / `dj-click-away` already do in the same function), so elements inserted by a later patch register too.

**How double-registration / duplicate listeners are avoided:**

- **No duplicate registry entries / no double-fire:** `_scanScopedElements()` is idempotent per-element — its existing `alreadyRegistered` check skips any element already in the registry for that key.
- **No duplicate window/document listeners:** the `addEventListener` install stays one-shot. The internal `_scanScopedElements()` call was **moved out** of `_installScopedDelegation()` into the per-bind path, and only the `addEventListener` calls remain behind the `_scopedDelegationInstalled` guard. (`_installScopedDelegation` has exactly one caller, so moving the scan out is safe; TurboNav — which resets the flag and clears the registry — still re-scans + re-installs on the next bind, unchanged.)
- Because the scan is no longer double-run on first mount, the bundle is **25 B smaller gzipped**.

Also corrected the stale comment.

## Tests

`tests/js/dj-window-rescan-1996.test.js` — 5 JSDOM cases against the built bundle:

1. **Primary reproducer** — mount with the scoped element absent, patch it in, dispatch a real `Escape` keydown on `window`, assert the handler fired.
2. `dj-document-keydown` patch-in fires on `document`.
3. No double-register across repeated binds (element present at mount, bind 3×, assert exactly one fire).
4. A patch-inserted element fires exactly once (no duplicate window listeners).
5. Cleanup still works — an element removed by a later patch does not fire.

**Gate-off (#1468):** neutering the new `_scanScopedElements()` call + rebuilding turns **4/5 RED**, including the primary reproducer (the 1 that stays green is the cleanup case, which correctly asserts 0 fires). Existing `dj-window-events` / `dj-shortcut` / `dj-click-away` suites stay green (30 tests).

## Budget

`client.min.js.gz` delta vs `origin/main`: **-25 bytes** (net smaller). eslint on the built bundle passes `--max-warnings 0`.


johnrtipton:
## Retrospective — PR #2015 (pipeline-drain)

**Task**: v1.1.0-5 — #1996 (`dj-window-*`/`dj-document-*` never bind on patch-inserted content). Worktree subagent; reviewed read-only via `git diff` of the source file (avoiding #1804).

### Quality: 5
### What went well
- Correct structural fix: `_scanScopedElements()` moved OUT of the one-shot `_installScopedDelegation()` into `bindLiveViewEvents()` so it re-populates the registry on every patch, while the window/document `addEventListener` install stays one-shot (guarded) — no duplicate listeners, no double-dispatch. Verified the two functions directly (scan at L1168 runs every bind; listeners at L1158 guarded).
- Mirrors the existing per-bind rescan `dj-shortcut`/`dj-click-away` already do (#168 grep-first).
- Net −25 B gzipped (removed the redundant first-mount double-scan). Gate-off: 4/5 tests RED when neutered.

### What didn't
- Two worktree hygiene items the subagent handled itself: stripped the absolute worktree path the minifier baked into the `.map` `sources`, and dropped a spurious `debug-panel.min.js.map` rebuild. Merge needed the CHANGELOG-union local-sync (both #1994/#1999/#2003 landed mid-flight) — done via a temp branch since the PR branch was pinned in the subagent worktree.

### Verified
- Source diff review confirmed scan-every-bind + one-shot-listeners + idempotent `alreadyRegistered`; eslint `--max-warnings 0` clean on the built bundle; 30 existing scoped tests green; 13/13 CI on the union head.
RETRO_COMPLETE

Issue #2017, opened 2026-07-02
dj-virtual: deeper server-side reconcile — stream-op wiring, VDOM-differ awareness, out-of-window patch landing (follow-up to #1988/#1989)


johnrtipton:
Follow-up to the client-side self-heal shipped for #1988 / #1989 (PR fix/1988-1989-dj-virtual-robustness).

## What the #1988/#1989 PR already fixed (client-side, DOM end-state)

- **#1988** — shell/spacer layout contract: shell is now `position: absolute; top/left/right: 0` (out of flow, so only the spacer defines `scrollHeight`); spacer now has `flex-shrink: 0` (survives a `display:flex` parent).
- **#1989 symptom 1 (full revert → permanent no-op)** — `initVirtualLists` / `refreshVirtualList` now DETECT a clobbered shell/spacer (server re-render replaced the container's children back to the raw `{% for %}` list) and auto teardown + re-init instead of no-op'ing on the WeakMap. Self-heal is order-independent.
- **#1989 symptom 2 (partial leak)** — a loose row appended OUTSIDE the shell/spacer wrapper is now auto-absorbed into the item pool (appended at the tail) so it renders inside the shell and receives subsequent patches, instead of leaking as a stray sibling.

## What remains deferred (this issue)

These are the deeper, larger-surface pieces that the client-side self-heal does NOT cover:

1. **Server-side stream-op wiring.** `stream_append` / `stream_insert` / `stream_prune` should set `container.__djVirtualItems` (or an equivalent hook) automatically for a `[dj-virtual]` target, so the documented "large append-only data" pairing works without bespoke app JS. This is a Python change in the stream-op handler. (Issue #1989 proposed scope item 1.)

2. **VDOM-differ `[dj-virtual]` awareness.** The Rust/JS diff currently has zero notion of `dj-virtual` (`grep -rn "dj-virtual" crates/` → nothing). A differ-aware path would splice new/changed/removed keyed items into `state.items` at their KEYED position (respecting order + removals), rather than the current append-only, tail-placement absorb heuristic. (Issue #1989 proposed scope item 2.)

3. **Out-of-window finalize-patch landing.** When a streaming item is scrolled OUT of the current window, it is (correctly) not in the DOM, so a finalize patch targeting it still cannot land. The client self-heal absorbs it into the pool so it renders + receives patches while in-window, but the fully-live mid-stream case for an off-window item needs differ awareness or a patch-buffering scheme.

[...]

Issue #1989, opened 2026-07-02
bug: dj-virtual has no integration path with normal server-driven re-renders or its own stream ops — new/changed items silently leak outside the shell or revert entirely


johnrtipton:
## Symptom (DJUST_LESSONS gotcha #18 and #22, hit in downstream production builds)

Two related, confirmed-live failure modes on a `dj-virtual` container whose underlying list changes via a normal server-driven re-render (not a fresh page load):

1. **Full revert, permanent no-op.** A genuine server re-render (e.g. an app's own presence self-broadcast round-tripping back to the originating session, or any other view-wide `push_to_view`) causes the keyed VDOM diff to replace the virtualized container's children back to the server's raw item list (Python has no notion of client-side virtualization — it always renders the full `{% for %}` loop). After this, `initVirtualLists()` never recovers the container: it tracks "already set up" state in a `WeakMap` keyed on the **container element**, whose identity is unchanged even though its children were just replaced — so every subsequent call is a silent no-op. Confirmed the only recovery is a manual `teardownVirtualList(container)` + re-`initVirtualLists()`.
2. **Partial leak, permanently stuck stream.** Sending one new item into an already-virtualized list (e.g. a new chat message + its streaming reply) does NOT revert everything — it appends just the new row(s) as loose children of the container, outside the shell/spacer wrapper. This seems less severe (extra layout height only) until the item finishes updating server-side: the client never applies that item's finalize patch, because the item lives outside the shell's expected DOM position. Confirmed via direct DB inspection: the server's `is_streaming` flag was already `False`, but the client's `[dj-stream]` element for that same message stayed present and frozen mid-content for 20+ seconds. Only a full page reload recovers it.

Both reproduced consistently, on djust 1.0.8, with `dj-virtual="messages" dj-virtual-item-height="140"` on a growing chat transcript.

## Root cause (mapped, citations grep-verified against `main`)

**There is exactly one integration point between `dj-virtual` and a changing data source, and nothing in djust wires it up automatically.**

- `refreshVirtualList()` (`python/djust/static/djust/src/29-virtual-list.js:400-431`) is the function meant to reconcile a virtualized container after its content changes. It is already called after **every** VDOM morph — confirmed at `python/djust/static/djust/src/09-event-binding.js:1456-1467`:
  ```js
  // dj-virtual / dj-viewport-*: re-scan after VDOM morph so new containers
  // get observers and existing ones pick up new first/last children. For
[...]

PR #2018, opened 2026-07-02
fix(vdom): dj-virtual layout contract + self-healing after server re-renders (#1988, #1989)

### LINKED ISSUE #1988: bug: dj-virtual's shell/spacer have no layout contract — flex crushes the spacer, and the shell (never taken out of flow) double-counts scrollHeight
## Symptom (DJUST_LESSONS gotcha #18 and #22, hit in downstream production builds)

`dj-virtual` on a `display:flex` container silently produces a spacer with `offsetHeight: 0`
regardless of its inline `style.height` — the virtualized list never scrolls past a few hundred
px. Switching the container to `display:block` fixes that, but then scrolling to the bottom of a
101-item / 140px-row list leaves ~400px of empty space *past* the true last item —
`container.scrollHeight` (14560px) doesn't match the spacer's own height (14140px, correct). Both
are silent: no console warning, no error, the feature just doesn't produce the right pixel math.

## Root cause (mapped, citations grep-verified against `main`, file `python/djust/static/djust/src/29-virtual-list.js`)

`setup()` creates two elements and never gives either one a real CSS contract:

```js
// lines 104-116
const shell = document.createElement('div');
shell.setAttribute('data-dj-virtual-shell', '');
shell.style.position = 'relative';      // <-- L108: does NOT remove it from flow
shell.style.willChange = 'transform';
shell.style.transform = 'translateY(0px)';

const spacer = document.createElement('div');
spacer.setAttribute('data-dj-virtual-spacer', '');
spacer.style.width = '1px';             // <-- L114: only 1px wide, no height set yet
spacer.style.pointerEvents = 'none';
spacer.style.visibility = 'hidden';
```

and `renderFixed()`/`renderVariable()` set the spacer's actual height per-render
(`spacer.style.height = (total * itemHeight) + 'px'` at L259; `offsets[total] + 'px'` at L318),
plus the shell's `transform: translateY(...)` to visually position the currently-rendered slice
(L280, L297, L384). No CSS beyond this ships anywhere in the repo (confirmed:
`grep -rn "data-dj-virtual" **/*.css` — zero hits).

Two independent consequences, both requiring a downstream app to patch djust's own generated
markup with `!important` CSS to work at all:

**1. The spacer is silently crushed in any `display:flex` container.** A flex item's default
`flex-shrink: 1` overrides `style.height` once the item's size exceeds the flex container's
available space — confirmed live: `spacer.style.height === "14140px"` but
`spacer.offsetHeight === 0`. `container.style.position = 'relative'` is set at L122, but nothing
sets `display` or `flex-shrink` on either the container or its children. The framework's own
guide example for `dj-virtual` happens to use a plain block container, so this never surfaces
there — but nothing in the setup code or the docs establishes that as a requirement.

**2. The shell was never taken out of flow, so its own rendered height silently doubles-counts
against the spacer's.** `position: relative` + `transform` does **not** remove an element from
normal document flow (per spec — transforms are a purely visual/paint-time effect); the shell's
untransformed box still occupies its own height at `top: 0`. Since the shell is a sibling that
comes *before* the spacer in DOM order (L128-129: `container.appendChild(shell);
container.appendChild(spacer)`), the shell's own natural height (however many rows are currently
rendered × their height) adds directly on top of the spacer's height for the purposes of
`container.scrollHeight` — confirmed live: shell holding 3×140px rows (420px) + spacer at 14140px
= `scrollHeight: 14560px`, i.e. ~400px past the true content end. The comment immediately above
the shell's creation (L102-103: *"A spacer sibling forces the container scroll height to the
virtual length"*) states the INTENDED design — the shell contributing nothing to scroll height —
but that intent is only achieved if the shell is taken fully out of flow, which `position:
relative` does not do.

## Side effect once patched with `position: absolute` (documented for completeness, not a separate bug)

Setting `shell.style.position = 'absolute'` (with `top/left/right: 0`, needed since nothing sets
those either) correctly zeroes out the shell's flow contribution — `scrollHeight` then matches the
spacer exactly. But if the virtualized container is a flex item elsewhere in the page and relies
on `align-items: stretch` for its own width/height (a common pattern for centered content via
`max-width` + `margin: 0 auto`), removing the shell from flow can leave the container's only
remaining in-flow child as the spacer (`width: 1px`) — auto margins take priority over flex
stretch on the cross axis once there's no other in-flow content demanding width, collapsing the
container to near-zero size. This is arguably out of djust's control (host-page layout), but it's
a second footgun that only appears *after* fixing #1/#2 above, worth a doc callout at minimum.

## Proposed scope

1. **Ship the shell/spacer CSS as part of `setup()` instead of leaving it to inline
   partial-styling.** At minimum add `shell.style.position = 'absolute'` (not `'relative'`) plus
   `top: '0'`, `left: '0'`, `right: '0'` — this alone makes the framework's own stated intent
   ("spacer alone determines scroll height") actually true, with no downstream CSS needed.
2. **Give the container an explicit `flex-shrink: 0` / `min-height: 0` (or document the
   requirement)** so `dj-virtual` doesn't silently break inside a flex ancestor — either ship it
   defensively via `container.style.flexShrink = '0'` in `setup()`, or add a checked assertion /
   console warning when `getComputedStyle(container).display === 'flex'` and the resulting spacer
   height doesn't match its `style.height` (a one-line `offsetHeight` sanity check after each
   `render()` would catch this class of bug immediately instead of silently producing a truncated
   scrollbar).
3. **Docs**: the `dj-virtual` guide/example should state the exact layout contract the container
   needs (no `display: flex` unless `flex-shrink: 0` is set on itself; shell is `position:
   absolute` internally, so host CSS should not assume it contributes to intrinsic sizing).

## Testing

- A JSDOM/browser test asserting `container.scrollHeight === totalItems * itemHeight` (fixed mode)
  and the variable-mode equivalent (`offsets[total]`), both inside a `display: flex; flex-direction:
  column` ancestor — this is exactly the layout that silently breaks today and has no coverage.
- A regression test scrolling to `container.scrollHeight` and asserting the last rendered item's
  `offsetTop + offsetHeight <= container.scrollHeight` with zero slack (today's bug leaves ~400px
  of slack).

Refs: `python/djust/static/djust/src/29-virtual-list.js` L104-130 (`setup`), L259/L318 (spacer
height per mode), L280/L297/L384 (shell transform per mode). Original feature: #797 (variable-height
items), touched again in #949/#951/#953 (dj-virtual data-key cache) — neither addressed the
container layout contract.

johnrtipton:
Closes #1988 and #1989. Deferred deeper reconcile tracked in follow-up **#2017**.

Two entangled `dj-virtual` bugs hit in downstream production chat/feed builds, both silent (no console warning/error). Both fixes are client-only, in `python/djust/static/djust/src/29-virtual-list.js`.

## #1988 — shell/spacer had no layout contract

Root causes (grep-verified in `setup()`):

- The injected **shell** was `position: relative` — which does NOT remove it from flow (transforms are a paint-time effect per spec). Its own rendered rows double-counted against the spacer, so `container.scrollHeight` overshot the true content by the shell's height → ~400px of dead space past the last item.
- The **spacer** had no `flex-shrink`. Inside a `display: flex` container a flex item's default `flex-shrink: 1` overrides its explicit `style.height` once it exceeds available space → `offsetHeight: 0`, and the list silently never scrolls.

Fix — the shell is now taken out of flow (`position: absolute; top/left/right: 0`) so **only the spacer** defines scroll height (translateY windowing preserved; the container is already made a positioned ancestor), and the spacer is `flex-shrink: 0` so its height survives a flex parent. Documented host-page caveat: a stretch-sized flex container should set an explicit size rather than rely on `align-items: stretch` from its now-absolute content.

## #1989 — no integration with server-driven re-renders / stream ops

The server always renders the full raw `{% for %}` list (no notion of client virtualization), so:

1. **Full revert → permanent no-op.** A full re-render reverts the container's children to the raw list; `initVirtualLists` then no-ops forever because it tracks setup state in a `WeakMap` keyed on the container element (identity unchanged). Only manual `teardownVirtualList` + re-init recovered it.
2. **Partial leak → stuck stream.** A single appended row lands as a loose child OUTSIDE the shell/spacer wrapper; its later finalize patch never applies.

Fix — self-healing after every VDOM morph (the reinit hook at `09-event-binding.js` already calls `initVirtualLists` + `refreshVirtualList` per `[dj-virtual]`):

- `structureIntact()` detects a clobbered shell/spacer (detached or repurposed — marker attributes gone) and transparently re-virtualizes against the fresh children. Wired into **both** `initVirtualLists` and `refreshVirtualList` so it heals regardless of call order.
- `absorbLooseChildren()` folds any element child that isn't the shell/spacer into the item pool (at the tail) so it renders inside the shell and receives subsequent patches, instead of leaking as a stray sibling.

## Implemented vs deferred (#1079 scope discipline)

**Implemented (this PR):** the #1988 layout contract; #1989 symptom-1 clobber detect + auto re-virtualize; #1989 symptom-2 loose-child absorb (append-only).

**Deferred to #2017:** automatic `stream_append`/`stream_insert`/`stream_prune` → `container.__djVirtualItems` wiring (a server-side Python change); VDOM-differ `dj-virtual` awareness (keyed mid-list inserts/removals at their true position); finalize-patch landing for an item scrolled OUT of the current window; a `djustDebug`-gated warning when a keyed patch finds no matching child. Absorb stays append-only, which is correct for chat/feeds — honestly noted in the guides.

## Tests + gate-off (#1468)

5 new regression cases in `tests/js/virtual_list.test.js` (24 total in file, all green):

| Case | Covers | Gate-off (change reverted → RED) |
|------|--------|----------------------------------|
| shell out-of-flow + spacer flex-safe | #1988 CSS contract | revert style lines → `shell.style.position`/`top` + `spacer.flexShrink` assertions fail |
| full-revert self-heal via reinit path | #1989 sym 1 | revert both `structureIntact` checks → shell stays `null` after morph |
| full-revert self-heal via `refreshVirtualList` alone | #1989 sym 1 (order-independent) | same |
| loose-child absorb | #1989 sym 2 | revert `absorbLooseChildren` call → new row stays a loose direct child |
| intact-list refresh no-op | #1989 regression | (guards the always-run absorb path from stray de-parenting) |

All three gate-offs were run manually (revert → rebuild → RED → restore) and confirmed.

## JSDOM caveat (reproduction fidelity)

JSDOM has **no layout engine** — `offsetHeight` / `scrollHeight` / `getBoundingClientRect` all return 0 — so these tests pin the **CSS contract** (the exact style properties `setup()` sets) and the **reconcile behavior** (DOM structure after a simulated morph), **not computed pixels**. Real-browser pixel verification — `container.scrollHeight` == spacer height, `spacer.offsetHeight > 0` under `display: flex`, and a live stream-append landing inside the shell — is a **recommended manual follow-up** that JSDOM structurally cannot cover.

## Budget

`client.min.js.gz` +231 B vs `origin/main` (well under the 2 KB target). Full JS suite: 1762 passing.


johnrtipton:
## Retrospective — PR #2018 (pipeline-drain)

**Task**: v1.1.0-5 — #1988 (dj-virtual shell/spacer have no layout contract) + #1989 (no integration with server re-renders). Worktree subagent; reviewed read-only via `git diff`.

### Quality: 5
### What went well
- **#1988 layout contract**: shell → `position:absolute` (out of flow so only the spacer defines `scrollHeight`, curing the double-count), spacer → `flex-shrink:0` (survives a flex parent instead of being crushed to `offsetHeight:0`) — the standard virtualization contract.
- **#1989 self-heal (client-side)**: `structureIntact()` auto re-virtualizes after a morph clobbers the shell/spacer (cures the WeakMap-identity permanent no-op), wired into BOTH `initVirtualLists` and `refreshVirtualList` (order-independent); `absorbLooseChildren()` folds stream-appended loose rows into the pool.
- **Scope discipline (#1079)**: the deeper Python-side `stream_append`→`__djVirtualItems` wiring + VDOM-differ `dj-virtual` awareness were honestly deferred to follow-up **#2017** rather than shipping a half-working partial.
- **Reproduction fidelity (#1849)**: JSDOM has no layout engine, so tests pin the CSS contract + reconcile behavior (not pixels); real-browser pixel verification is flagged as a manual follow-up in the PR + CHANGELOG. Gate-off all-RED.

### What didn't
- Its first CI run never registered checks on GitHub (0 checks); the union-sync (bundle conflicts resolved by rebuilding from merged source) produced a fresh head that triggered CI green. Cited line numbers had drifted (reinit site 1519-1529 vs cited 1456-1467) — same functions.

### Verified
- Read-only diff: CSS contract + `structureIntact`/`absorbLooseChildren` wired into both paths, no banned patterns; 24 virtual-list tests + 1762 JS suite green; +231 B gzipped; 13/13 CI on the rebuilt head.
RETRO_COMPLETE

Issue #1601, opened 2026-05-26
VDOM diff returns 0 patches for textarea text-node content change (cross-session broadcast)


johnrtipton:
## Summary

When a `<textarea>{{ doc_content }}</textarea>` placeholder changes server-side and the re-rendered HTML is broadcast to a second session via `push_to_view`, the VDOM diff produces **0 patches** even though the textarea's text content has changed. The receiving tab never updates until it does a full reload.

Discovered while upgrading examples.djust.org from djust 0.3.0rc5 → **1.0.0rc11**. May overlap with johnrtipton/djust#135 (input `.value` sync after VDOM patch), but I think it's distinct — that PR targets `.value` property sync after `innerHTML` replacement, whereas here the diff itself returns no patches.

## Versions

- djust: **1.0.0rc11** (editable from source)
- Django: 5.2.14
- channels: 4.x with `InMemoryChannelLayer`
- Single daphne process on Python 3.12

## Reproduction

The `collab-notepad` example app in examples.djust.org demonstrates this:

- Template: `<textarea ... dj-input="update_content">{{ doc_content }}</textarea>`
- View has `update_content(value=...)` which sets `SHARED_DOCUMENT['content'] = value`, calls `_sync_state()` (which sets `self.doc_content`), then `push_to_view('notepad.views.CollabNotepadView', handler='_on_broadcast', ...)`.
- `_on_broadcast` on receivers calls `_sync_state()` to pull the new value into `self.doc_content`.

Steps:

1. Open the page in two browser tabs (Tab A, Tab B).
2. In Tab A, type into the textarea (or programmatically dispatch `update_content` with a fresh value).
3. Server processes the event, updates shared state, broadcasts.

## Expected

Tab B's textarea content updates to reflect the new `doc_content`.

## Actual

- Tab B receives the broadcast — console shows `[LiveView] Received: %s %o patch Object` followed by `[LiveView] No DOM changes needed (0 patches)`.
- Tab B's textarea is unchanged.
- **State is correct server-side**: reloading Tab A (or Tab B) renders the new value via the HTTP mount path, so the data is persisted. Only the live cross-session patch is missing.

## Evidence the bug is in VDOM diffing (not transport or handler wiring)

- Presence sync works in the same view via the same mechanism — Tab B correctly shows "Online Users (1)" when Tab A connects. So `push_to_view` delivery and `_on_broadcast` execution are healthy.
[...]

PR #2019, opened 2026-07-02
fix(vdom): per-field opt-out from broadcast textarea sweep + force-value override for focused fields (#1990, #1991)

### LINKED ISSUE #1990: feat(vdom): no way for a handler to force-clear a focused form field — `skipValue` has no "server value changed" override
## Symptom (DJUST_LESSONS gotcha #8, hit in downstream production builds)

A chat composer `<textarea>` bound via `dj-submit`/`dj-keydown="send.enter"`: after the `send`
handler sets `self.composer = ""`, the textarea keeps the just-sent text whenever it still has
focus (the Enter-to-send path — the Send *button* path clears fine because focus moves to the
button, not the textarea). Reproduced consistently; a handler-driven change to a field's value has
no way to take effect while that field is focused, even though the field's *identity* (name) didn't
change.

## Root cause (grep-verified against the current `12-vdom-patch.js`, NOT the older `preserveFormValues` utility)

`morphNode()`'s form-element value-sync computes, per node, at
`python/djust/static/djust/src/12-vdom-patch.js:942-948`:

```js
const isFocused = document.activeElement === existing;
const nameChanged = existing.getAttribute('name') !== desired.getAttribute('name');
const skipValue = isFocused && !_isBroadcastUpdate && !nameChanged;
```

and then, at `:978-981`:

```js
if (existing.tagName === 'TEXTAREA' && !skipValue) {
    existing.value = existing.textContent || '';
}
```

(same `!skipValue` gate applies to `INPUT`/`SELECT` at `:950`/`:959`.) For the composer scenario:
`isFocused=true` (Enter doesn't blur), `_isBroadcastUpdate=false` (self-triggered, not a remote
patch), `nameChanged=false` (the field's `name="composer"` is unchanged across the send) →
`skipValue=true` → the sync is skipped unconditionally, regardless of what the new server value
actually is. There is no comparison between the *previous* server-rendered value and the *new* one
— `nameChanged` is the only override signal, and it only fires when the field's identity itself
changes, not when the handler legitimately wants to overwrite the current field's content (send,
clear-search, apply-a-suggestion, etc.).

**Note on a decoy**: an older utility, `preserveFormValues()` (`12-vdom-patch.js:635`, exported as
`window.djust.preserveFormValues`), implements a similar-looking save/restore-focused-value scheme
and is exercised by `tests/js/preserve_form_values.test.js` — but it has **zero call sites** in the
built `client.js` outside its own definition/export/tests. It appears to be legacy code superseded
by the `morphNode`/`skipValue` mechanism above (which is what genuinely executes on the real patch
path) — worth confirming and removing/documenting as deprecated if so, since it's easy for someone
debugging this class of issue (as happened here) to chase the wrong function.

**Related but distinct**: a *separate* code path drives the same symptom during a broadcast-type
patch specifically, filed as a companion issue from the same investigation (`DJUST_LESSONS gotcha
#17`) — that one is `_isBroadcastUpdate`-gated already, whereas this issue is about the
**self-triggered, non-broadcast** case, which has no gate/override at all today.

## Working app-level recipe today (verified, but shouldn't be required)

1. Drop `dj-model` entirely — a `dj-model`-bound field actively re-populates from a debounced
   client-side update that can fire *after* the handler clears the value.
2. Use `dj-submit` with a `name` attribute for deterministic serialization; `dj-keydown="X.enter"`
   for a keyboard-triggered submit path (handler receives `value` directly, no separate sync needed).
3. Bump a value baked into `dj-key` (e.g. a nonce) on every handler call that should clear the
   field — this forces a *node replacement* (new element, no morph, no `skipValue` check at all)
   rather than an in-place value sync. This is the only way to force a value through today, and it
   works only because it sidesteps `morphNode` entirely rather than because `skipValue` allows it.

## Proposed scope

1. **Preferred**: track the last server-rendered value per field (already resolvable — `desired`'s
   value is available at diff time) and compare it against what was rendered on the *previous*
   patch. If the server's value actually changed between renders (not just re-sent the same value),
   treat that as an explicit signal to override `skipValue` for that field — mirroring what
   `nameChanged` already does for identity changes, just extended to content changes.
2. **Alternative / additive**: an opt-out attribute (e.g. `dj-preserve="false"` or `dj-server-value`)
   an app can set on a specific field to say "this field's value sync should never be skipped for
   focus," so a handler-driven clear/overwrite is unconditionally honored regardless of the diffing
   heuristic.
3. Either way, apply the same fix to `INPUT`/`SELECT` (`:950`, `:959`), not just `TEXTAREA` — all
   three share the identical `skipValue` gate and the identical gap.
4. Docs: the model-binding / forms guide doesn't mention that a *focused* field cannot be reset by
   a handler at all today — worth an explicit callout with the blessed pattern (dj-key bump) as an
   interim recipe regardless of whether 1/2 ship.

## Testing

- Regression test mirroring the reproduction: mount a view with a `dj-submit`-style handler that
  clears a public string attr; simulate the field staying focused across the patch; assert the
  rendered value is empty (not the stale client value) once the server's value has genuinely
  changed between two consecutive patches.
- Confirm `nameChanged`-triggered override still behaves identically (regression guard for the
  existing behavior).
- `INPUT`/`SELECT` parity tests alongside the `TEXTAREA` case.

Refs: DJUST_LESSONS gotcha #8 (this project's field-notes doc), companion issue for gotcha #17
(broadcast-sweep case).

### LINKED ISSUE #1991: bug: broadcast textarea-value sweep has no per-field opt-out, overwrites unrelated drafts
## Symptom (DJUST_LESSONS gotcha #17, hit in downstream production builds)

Any `<textarea>` anywhere in the LiveView root gets its `.value` unconditionally overwritten
the instant **any** broadcast patch (`push_to_view`) lands — even a textarea with no relation
to the broadcasted content. Confirmed live via a two-tab test in a chat app: a peer's message
landing in a *different* conversation wiped an unrelated user's unsent draft in the message
composer, with no way to opt a specific field out.

Root-caused with an instrumented test that overrode the `value` property descriptor on the
composer `<textarea>` to log every external setter call with a stack trace. The trace named
the exact call site (`static/djust/client.js:403` in the built bundle).

## Root cause (source-verified, file:line against `python/djust/static/djust/src/`)

Two separate sites do this unconditionally, with **zero** opt-out:

1. **`02-response-handler.js:118-124`** — after `applyPatches(data.patches)` resolves for a
   broadcast frame:
   ```js
   if (data.broadcast) {
       const root = getLiveViewRoot();
       if (root) {
           root.querySelectorAll('textarea').forEach(el => {
               el.value = el.textContent || '';
           });
       }
   }
   ```
2. **`12-vdom-patch.js:640-644`**, inside `preserveFormValues` — the `innerHTML`-replacement
   path's own copy of the same sweep, gated only on the module-level `_isBroadcastUpdate` flag:
   ```js
   if (_isBroadcastUpdate) {
       updateFn();
       container.querySelectorAll('textarea').forEach(el => {
           el.value = el.textContent || '';
       });
       return;
   }
   ```

Neither site checks `dj-update="ignore"` (that's `_isIgnored`, only consulted inside
`morphNode` per-attribute at `12-vdom-patch.js:923`/`937`, a completely different code path
these two sweeps never go through) or any per-element attribute at all — it's a blanket
`querySelectorAll('textarea')` over the whole root.

**This sweep is the fix for #1601** ("VDOM diff returns 0 patches for textarea text-node
content change (cross-session broadcast)") — the code comment at `02-response-handler.js:112-114`
says so directly: *"VDOM patches update textContent directly (not via innerHTML), so
preserveFormValues never runs. Textarea .value is separate from .textContent after initial
render — must sync explicitly."* That fix is correct in spirit (a shared/collaborative textarea
*should* pick up a peer's edit), but it's scoped too broadly: it hits every textarea in the
document, not just ones bound to the broadcasted state.

### A related, narrower divergence in the same investigation

The **non-broadcast** per-node morph path has an analogous but not-quite-matching rule at
`12-vdom-patch.js:944-948`:
```js
const isFocused = document.activeElement === existing;
const nameChanged = existing.getAttribute('name') !== desired.getAttribute('name');
const skipValue = isFocused && !_isBroadcastUpdate && !nameChanged;
```
`isFocused` alone isn't sufficient once an element has been **blurred** — a self-triggered
re-render (no broadcast at all) while a composer was blurred with an unsent draft also wiped
it, because `skipValue` only protects a *currently-focused* element. The broadcast sweep
ignores focus entirely by design (comment at `12-vdom-patch.js:636-638`: *"the server content
from another user should take effect"*); the ordinary morph path only ignores it while
focused. The two paths have genuinely different, undocumented focus-awareness rules for what
is nominally the same operation (syncing a textarea's displayed value).

## Proposed scope

1. **Per-element opt-out for the broadcast sweep.** Both sweep sites should skip an element
   that opts out — e.g. respect `dj-update="ignore"` (making it consistent with what that
   attribute already means elsewhere), or introduce a narrower `dj-preserve="local"` /
   `dj-server-value="false"` for exactly this case. Today there is no way to tell djust "this
   textarea's value is client-owned, never let a broadcast touch it."
2. **Reconcile the two focus-awareness rules.** Document (or unify) why the broadcast path
   ignores focus by design while the ordinary morph path only protects a *currently-focused*
   element — a blurred-but-dirty field is unprotected in the non-broadcast path too, which
   looks like an oversight rather than an intentional divergence from the broadcast case.
3. Cross-reference **#1601** in the fix — the sweep exists to correctly deliver a broadcasted
   textarea update; scoping it doesn't mean reverting #1601, just not applying it to fields
   that aren't meant to mirror the broadcasted state.

## Environment

- djust version: 1.0.8 (installed), reproduced against `main` source (`python/djust/static/djust/src/02-response-handler.js`, `12-vdom-patch.js` — same code present in the current checkout)
- Django: 5.x
- Reproduced with: `InMemoryChannelLayer` / Redis channel layer, single daphne process, two browser tabs same conversation view

Refs: DJUST_LESSONS gotcha #17, #1601 (the precursor bug this sweep fixed).

johnrtipton:
Grouped fix for two form-field value-preservation gaps in the VDOM patch path, closed together with two consistent, **opposite-polarity** declarative attributes.

Closes #1990. Closes #1991.

### #1991 — broadcast textarea sweep had NO per-field opt-out
Every `push_to_view` broadcast unconditionally reset **every** `<textarea>.value` in the LiveView root (the #1601 sweep, scoped too broadly), so a peer's message landing in an unrelated conversation wiped a user's unsent draft. There was no way to tell djust "this textarea's value is client-owned."

The sweep lived, duplicated, at **two** call sites:
- `02-response-handler.js`'s `applyPatches` broadcast path (the real production path)
- `12-vdom-patch.js`'s `preserveFormValues` innerHTML path

### #1990 — no way to force-clear a FOCUSED field
`morphElement()` (the real function — the issue cites the old name `morphNode`) computed `skipValue = isFocused && !_isBroadcastUpdate && !nameChanged` and skipped the server value sync for a **focused** input/select/textarea unless its `name` changed. So a handler that cleared a still-focused composer — the Enter-to-send path, where Enter never blurs — could never take effect: the sent text stayed in the box. `nameChanged` was the only override, and it only fires on identity changes, not on a legitimate handler-driven clear/overwrite.

## The two attributes (and why they're consistent)

| Issue | Attribute | Polarity | Effect |
|---|---|---|---|
| #1991 | `dj-update="ignore"` | **opt-out** | a textarea carrying it is **skipped** by the broadcast sweep (its value is client-owned) |
| #1990 | `dj-force-value` | **opt-in** | a field carrying it receives the server value **even while focused** |

They are deliberately opposite-polarity halves of the same "who owns this field's displayed value?" question. `dj-update="ignore"` **reuses the existing convention** already meaning "client-owned, don't update" for the per-node morph path (honored at `12-vdom-patch.js` for individual nodes) — the fix simply makes the two broadcast sweeps honor it too, closing a real consistency gap. `dj-force-value` is a new opt-in that is **conservative** (fires only when the attribute is explicitly present, so default focused-field typing-protection is unchanged for everyone else) and **lazy** (evaluated only when the field would otherwise be skipped — zero hot-path cost). Covers INPUT/SELECT/TEXTAREA.

## Structural cure (#1646)
Both broadcast-sweep call sites now route through **one** shared `syncBroadcastTextareas(root)` helper (in `12-vdom-patch.js`, exported on `window.djust`), so the opt-out lives in exactly one place — retiring the parallel-path-drift class rather than fixing one site and leaving the twin.

## Tests — `tests/js/form_value_preservation_1990_1991.test.js` (11 JSDOM cases)
- **#1991 shared helper**: sweeps a normal textarea; does NOT wipe a `dj-update="ignore"` one; mixed root (sweeps opted-in sibling, skips opted-out); null-root tolerated.
- **#1991 real production path**: drives `handleServerResponse` with a faithful broadcast frame (`source:'broadcast'` + `broadcast:true`, matching `websocket.py`) — an unrelated broadcast patch sweeps the shared textarea but preserves the opted-out draft.
- **#1991 anti-drift pin (#1646)**: asserts neither sweep site inlines a raw `querySelectorAll('textarea')` sweep and both route through the helper.
- **#1990**: TEXTAREA + INPUT, WITH and WITHOUT `dj-force-value`, using real focus (`document.activeElement`).

### Gate-off (#1468)
- Remove the #1991 opt-out check → **3 tests RED** (both helper opt-out cases + the real site-B path).
- Remove the #1990 override → **exactly the 2 WITH-attr tests RED**, while the WITHOUT-attr default-behavior tests stay green (non-tautological).

## Bundle budget
**+52 bytes gzipped** (`client.min.js.gz`) — well under the 2 KB budget.

## Cited-path corrections
- The function the issues call `morphNode()` is actually `morphElement()` in the current `12-vdom-patch.js`.
- `preserveFormValues` (site 2 in #1991) is confirmed **dead code** (zero real call sites); its broadcast branch is fixed anyway via the shared helper so the opt-out can't drift back if it's ever revived.

## Docs
`docs/website/guides/declarative-ux-attrs.md` — new "`dj-force-value` & broadcast value preservation" section documenting both attributes (opt-in vs opt-out), plus an intro-list entry.


johnrtipton:
## Retrospective — PR #2019 (pipeline-drain)

**Task**: v1.1.0-5 — #1990 (force-clear a focused field) + #1991 (broadcast textarea sweep has no opt-out). Worktree subagent; reviewed read-only via `git diff`.

### Quality: 5
### What went well
- **Consistent, minimal vocabulary**: #1991 reused the *existing* `dj-update="ignore"` convention (client-owned; the two broadcast sweeps simply weren't honoring it — a real consistency gap, not new syntax); #1990 added the opposite-polarity opt-in `dj-force-value` (fires only when explicitly present → default focused-field protection unchanged for everyone).
- **#1646 structural cure**: both sweep sites (`02-response-handler.js` + the dead-but-fixed `preserveFormValues` path) route through ONE `syncBroadcastTextareas(root)` helper, so the opt-out lives in exactly one place.
- Gate-off non-tautological: removing the #1990 override turns exactly the 2 WITH-attr tests RED while the WITHOUT-attr default tests stay green.

### What didn't
- Cited `morphNode()` is actually `morphElement()` (line drift); `preserveFormValues` (sweep site 2) is dead code — fixed anyway via the shared helper so the opt-out can't drift back if revived.

### Verified
- Read-only diff: `skipValue` gains `&& !hasAttribute('dj-force-value')` on both existing+desired; helper skips `dj-update="ignore"`. 11/11 new + 89 adjacent tests green; +52 B gzipped; 13/13 CI green.
RETRO_COMPLETE

PR #2022, opened 2026-07-03
fix(live): early, actionable error when ORM objects are stored in LiveView state

szto:
## What

Storing a Django ORM object in LiveView state (e.g. `self.user = request.user` in `mount()`) fails **late and opaquely** instead of early and actionably:

- `_is_serializable` deliberately passes `Model`/`QuerySet` through (correct for the JIT render pipeline), so the existing friendly DEBUG `TypeError` in `get_state()` never fires for ORM values
- the value then reaches the **state-persistence path** (signed `state_snapshot`), where `json.dumps(..., default=str)` silently flattens the model to a string — state restore after reconnect is broken, or the handler 500s with an unrelated-looking error

Related: #1994 covered the *private*-state sibling of this (fixed by #2012); the public-state case above still fails on 1.1.0rc4.

## Fix

Separate the two serialization contexts. New `_reject_orm_value_in_state_persistence` guard used by `_capture_snapshot_state(strict=True)` (called from the runtime snapshot path), matching the existing friendly-error style/policy:

- `DEBUG=True` → `TypeError` telling the developer exactly what to do: *store the pk and refetch in the handler (`self.user_id = user.pk`) — ORM objects can't survive state persistence*
- production → `logger.warning` + skip the field (no 500, no silently-corrupted snapshot)

The JIT render pipeline is untouched — `Model`/`QuerySet` still render fine in templates.

## Tests

TDD: new `test_state_snapshot_orm_early_validation.py` (5 cases: Model / QuerySet in state, DEBUG vs prod behavior, JIT-render unaffected) — 3 failed on rc4 before the fix, all pass after. Full suite: 8659 passed, 0 failed.


johnrtipton:
Rebased onto current `main` (was 73 commits behind; branched pre-`1.1.0rc6`). Force-pushed `c5350fe7 → 95b8ea83`. Two conflicts needed non-mechanical resolution — flagging for reviewers:

1. **`_SETATTR_WHITELIST` security pins re-derived (not carried).** This PR bumped the CWE-915 mount-chokepoint setattr pins `1203/1205 → 1270/1272` as a `+67` shift off the *pre-rebase* baseline. But `main` had already moved them to `1225/1227` (#2035). I re-derived them empirically against the rebased `live_view.py` — the two sanctioned `setattr(view, key, value)` sites (still the `DynamicLiveView` developer-dict application: `class` at 1277, `result = func(request, …)` at 1288) are now at **1292/1294**. `test_mount_chokepoint_structural.py` passes (14/14), so the CWE-915 guard is green, not silently mis-pinned.

2. **CHANGELOG merge was corrupt — repaired.** git's 3-way merge dropped this PR's entry into the already-shipped `## [1.1.0rc4]` section *and* deleted a shipped `#1970` entry from it (the exact class the `#2028` pin-check guards against — it failed on the raw auto-merge). I discarded the mis-merge, took `main`'s CHANGELOG verbatim, and hand-placed this PR's entry at the top of `## [Unreleased] ### Fixed`. Both `check-changelog-tagged-sections` and `check-changelog-test-counts` now pass.

Auto-merges in `live_view.py` / `runtime.py` landed cleanly on the ADR-022-converged code (guard at `live_view.py:797`, `strict=True` at `runtime.py:2518`). Verified: the ORM suite (6/6), chokepoint (14/14), and 304 snapshot/restore/time-travel caller tests all pass; full serial pre-push suite green.

Note (separate from the rebase — from the review): the DEBUG `raise TypeError` in `_reject_orm_value_in_state_persistence` is still swallowed by the `except Exception` wrapper around the only `strict=True` caller (`runtime.py:2461`, the #1788 "snapshot emission must never break mount" posture), so in the real mount path DEBUG logs-and-continues rather than raising. Left as-is per the rebase scope — worth a follow-up decision (re-raise past the wrapper, or adjust the DEBUG claim).

szto:
Addressed the review note about the DEBUG raise being swallowed by the #1788 fail-soft wrapper — took the **re-raise past the wrapper** option (2b0efb86, 9724aa81):

- The guard now raises a dedicated `NonPersistableStateError(TypeError)` (defined in `live_view.py`) instead of a bare `TypeError`, so `runtime.py`'s `except Exception` can distinguish the *deliberate* developer-error report from an unexpected emission failure. The wrapper re-raises exactly that class and keeps log-and-continue for everything else. Production is unaffected by construction — the guard warns + skips there, so no raise exists to propagate.
- Downstream of the re-raise, WS `receive()`'s outer handler routes the exception through `handle_exception` (stack trace in DEBUG + error frame to the client), so the actionable message is loud at mount time on the real path.
- Two new REAL-path tests drive `ViewRuntime.dispatch_mount` end-to-end (MockTransport pattern from `test_runtime_mount_state_restore_1913`): DEBUG → the rejection propagates out of `dispatch_mount` and no mount frame ships; production → mount survives, the ORM key is skipped from the signed blob, the sibling scalar persists, warning logged. Gate-off verified: neutering the re-raise fails the DEBUG test with `DID NOT RAISE`.
- Grep-confirmed `runtime.py` is still the ONLY `strict=True` caller (no WS twin post-ADR-022; `time_travel.py` stays non-strict by design).
- CWE-915 setattr pins re-derived (`1292/1294 → 1316/1318`, +24 from the new exception class ahead of the `LiveView` class); re-verified sanctioned — same two `DynamicLiveView` developer-dict lines. Chokepoint suite 14/14 green.
- CHANGELOG claim updated to the propagation behavior; test count recounted 6 → 8.

Full suite: 8820 passed / 0 failed across all test roots.

johnrtipton (commented):
## Code Review — pipeline-ship Stage 7

Thanks for a careful, well-documented contribution, @szto. The core thesis is correct and the implementation lands it precisely: the JIT render pipeline (`_is_serializable` / `get_state()`) *should* keep passing `Model`/`QuerySet` through, while the client-signed **persistence** path must not silently `DjangoJSONEncoder`-flatten them into a disconnected `dict` that comes back broken after a back-navigation restore. The fix is scoped to exactly that path, the tests are strong (they fail when the guard is disabled — see Local verification), the security-sensitive mount chokepoint invariant is untouched, and the runtime re-raise correctly defeats the `#1788` fail-soft swallow so the DEBUG failure is loud end-to-end. Nicely done.

**Verdict:** APPROVE — correct, tightly-scoped, gate-off-proven; the only items below are non-blocking (one unrelated CI file) and nits.

### 🔴 Blocking
None.

### 🟡 Non-blocking

- **`.github/workflows/security-hotspot-check.yml` (+81/-54) is unrelated to this fix and duplicated across your open PRs.** The try/catch-around-fork-label change is byte-identical in #2040, #2023, and #2022. It's benign — it only turns the expected fork read-only-token `403` on the *labeling* step into a `core.warning`, rethrows everything else, and does not weaken the actual hot-spot scan (the scan job still runs and gates). But it doesn't belong in a state-persistence bugfix. Suggest extracting it to its own PR; whichever lands first makes the copies in the other two redundant (drop them on rebase).

### 🟢 Nits

- **`python/djust/live_view.py:872` — `logger.warning(msg)` passes a pre-built f-string.** The canon logging rule prefers `%s`-parameterized calls. I'm calling this a nit, not a finding, because (a) it is a verbatim mirror of the pre-existing convention in the *same file* at `live_view.py:1130-1143` (`get_state()`'s sibling error builds an f-string `msg`, then `raise TypeError(msg)` / `logger.error(msg)`), so it's established house style here; (b) it's runtime-safe — `logger.warning(msg)` with no positional args performs no `%`-interpolation, and the interpolated values (`key` = a developer attribute name, `value_type` = a class `__name__`) are developer-controlled identifiers, not client/request data, so there's no tainted-format-string or log-injection surface; and (c) Ruff `G004` only flags inline f-strings in the call, not a variable, so CI is correctly green. Reusing one message for both the exception and the log is a reasonable readability trade-off. If the project ever tightens the logging rule, change both this site and the `get_state()` site together.

- **`python/djust/tests/test_state_snapshot_orm_early_validation.py:118` — QuerySet has a DEBUG-raise test but no production-skip test.** `Model` gets both (`test_model_instance_raises_actionable_type_error` + `test_model_instance_is_skipped_not_silently_converted_to_dict`); `QuerySet` only gets the DEBUG-raise (`test_queryset_raises_actionable_type_error`). The guard treats them identically via the `isinstance(value, (models.Model, QuerySet))` tuple, so coverage is adequate — but a symmetric prod-skip case for a QuerySet would lock the parity in.

- The PR brief referenced "5 cases"; the file actually ships **8** test methods and the CHANGELOG correctly says "8 new tests." No action needed — flagging only so the count isn't mis-quoted downstream.

### Concern-by-concern

1. **Two serialization contexts kept separate — CONFIRMED.** `_is_serializable` is untouched (it appears in the diff only inside a docstring). The guard runs *only* under `strict=True`, and the sole `strict=True` caller is `runtime.py`'s `state_snapshot_signed` emission. `time_travel.py:147,186` call `_capture_snapshot_state()` with no `strict` kwarg → old lossy-by-design behavior preserved. `TestRenderingJitPipelineUnaffected` asserts `get_state()` still returns the live `Model`. JIT path genuinely untouched.

2. **Commit-or-rollback / early-return (#1198) — CLEAN.** `_capture_snapshot_state` builds a *local* `result` dict from a read-only iteration of `self.__dict__`; the guard mutates nothing on `self`. On DEBUG raise the local `result` is discarded and the exception propagates out of `dispatch_mount` before `mount_msg["state_snapshot_signed"]` is set — `test_debug_rejection_propagates_out_of_dispatch_mount` asserts `transport.mount_frame is None` (no half-shipped frame). In prod, only the offending key is skipped; siblings persist.

3. **Mount chokepoint (#291/#1861) — invariant PRESERVED.** The `test_mount_chokepoint_structural.py` change is purely a line-number bump of the existing `_SETATTR_WHITELIST` (1225/1227 → 1316/1318), forced by the new class/guard/docstring being added *ahead* of the `LiveView` class. The whitelist still has exactly **2** entries (same as `main`), both still the `@live_view` `DynamicLiveView` developer-dict `setattr(view, key, value)` sites (`live_view.py:1316,1318`) — no new client-controlled setattr was sanctioned. The guard call in `runtime.py` sits in the late snapshot-emission block, well after the pre-mount auth/tenant sequence, so it's on the right side of the auth gate.

4. **Prod fail-soft — SAFE, not just deferred.** `%s`-vs-f-string aside (nit above), skipping the field is genuinely safe on restore, not a papered-over break: `_restore_snapshot` (`live_view.py:999`) only `safe_setattr`s the keys *present* in the snapshot. Since restore re-runs `mount()` first, a skipped `self.user` retains the freshly re-fetched `request.user` rather than a broken `dict`; request-independent ORM state reverts to its `mount()` default with a logged warning. Verified end-to-end by `test_production_mount_survives_and_skips_orm_key` (signed blob omits `user`, keeps `label`, warns).

5. **Secure-defaults precedence (#1867) — no interaction.** The guard operates at the *attribute* level (reject the whole `Model`/`QuerySet` attr) and runs *before* `json.dumps`, so it never reaches `serialization.py`'s field-level `_ALWAYS_EXCLUDED_FIELDS` / allowlist logic. It strictly *reduces* exposure (the model's fields never get serialized on the strict path); no re-exposure, no double-handling.

6. **`.pyi` stub — correct.** `live_view.pyi:30` adds `class NonPersistableStateError(TypeError): ...`, matching the real class. The stub does not (and need not) declare the private `_capture_snapshot_state` / `_reject_orm_value_in_state_persistence`, so there's no signature to drift.

### Local verification

Isolated git worktree; reused the main checkout's prebuilt `_rust.cpython-312-darwin.so` (Rust unchanged by this PR); worktree `python/` preferred on `sys.path` (verified `djust.__file__` resolves to the worktree). Ran with the main `.venv` Python 3.12:

- `pytest python/djust/tests/test_state_snapshot_orm_early_validation.py` → **8 passed**.
- `pytest .../test_state_snapshot_orm_early_validation.py .../test_mount_chokepoint_structural.py` → **22 passed** (chokepoint test green with the new whitelist lines).
- **Gate-off tautology self-test (#1200/#1468):** neutered the guard by adding an early `return False` at the top of `_reject_orm_value_in_state_persistence`, re-ran → **5 of 8 failed**, including the two real-path cases: `TestRealMountPathDebugFailsLoud::test_debug_rejection_propagates_out_of_dispatch_mount`, `test_production_mount_survives_and_skips_orm_key`, plus `test_model_instance_raises_actionable_type_error`, `test_queryset_raises_actionable_type_error`, and `test_model_instance_is_skipped_not_silently_converted_to_dict`. The 3 that still passed are the *unaffected-path* assertions (`TestNonStrictCallersUnaffected`, `TestRenderingJitPipelineUnaffected`), which correctly must pass with or without the guard. Restored the guard and re-confirmed a clean diff vs the PR head. The suite is non-tautological and exercises the real `dispatch_mount` path.

Merge is intentionally deferred to the maintainer — this review posts a comment only and does not merge.

johnrtipton (commented):
## Code Review — pipeline-ship Stage 7 (re-verify against post-#2023 main)

Re-checked after #2023 merged (main is now at `ce43a75e`). No new commits on this branch since the prior review — this pass confirms the change is still sound and merges cleanly on the moved base.

**Verdict: APPROVE** — merge deferred to the maintainer.

### Clean against current main (both PRs carried the same workflow change)
- `security-hotspot-check.yml` in this branch is now **byte-identical** to `origin/main`'s copy (same md5 — #2023 already landed the shared fork-403 fix), so merging this PR neither re-touches nor reverts it. `git merge-tree` reports **0 conflicts**; GitHub shows `MERGEABLE`. The three-dot diff still *displays* the workflow lines only because the merge-base predates #2023 — it's a tree-equivalent, not a real change.
- No file overlap with #2023's merge other than that now-identical workflow file and the shared `[Unreleased]` CHANGELOG section (auto-merges cleanly — 0 conflicts).

### Design re-confirmed
- **Two serialization contexts kept separate**: the guard runs only under `_capture_snapshot_state(strict=True)`, whose sole caller is `runtime.py`'s `state_snapshot_signed` emission. The JIT render pipeline (`_is_serializable`/`get_state()`) and dev-only time-travel capture (non-strict) are untouched — Model/QuerySet still render in templates.
- **#1788 fail-soft interaction handled precisely** (`runtime.py:2530`): the dedicated `NonPersistableStateError(TypeError)` subclass is re-raised *before* the `logger.exception` swallow, so the DEBUG rejection is loud end-to-end while genuine emission errors still fail soft. Production never raises (warn + skip).
- **Secure choice**: rejects rather than auto-rehydrating by pk — avoids the mass-assignment shape the HMAC signing exists to prevent.

### First-hand verification (local, Python-only — Rust unchanged)
- `test_state_snapshot_orm_early_validation.py` + `test_mount_chokepoint_structural.py`: **22 passed**.
- **Gate-off A** (neuter the guard so nothing is rejected) → **5 of 8 fail**; the 3 survivors are the deliberately-unaffected-path tests (non-strict callers + JIT pipeline). Non-tautological.
- **Gate-off B** (neuter *only* the `runtime.py` #1788 re-raise) → `test_debug_rejection_propagates_out_of_dispatch_mount` fails — proving that test specifically guards the fail-soft-propagation fix, not just the guard itself.
- CI green on `5006fd8d` (13/13 real checks; CodeQL/security-scan skipped as usual).

### Non-blocking
- 🟢 `live_view.py:872` `logger.warning(msg)` passes a pre-built f-string. Runtime-safe (no lazy args; interpolated values are developer identifiers, not client input) and mirrors the existing `get_state()` convention — kept as a nit. If you want strict `%s` parity you could split the DEBUG message (needs the f-string for the exception) from the prod log line.

Ready to merge whenever you are; suggest still extracting the workflow fix from #2040 into its own PR since it's now already on main via #2023.

johnrtipton (approved):
Looks good.

Issue #397, opened 2026-02-26
fix(checks): T011 unsupported tags list needs updating — Rust renderer now supports extends/block


johnrtipton:
## Problem

T011 is meant to catch Django template tags that the Rust renderer silently ignores. At some point `{% extends %}` and `{% block %}` appeared in the unsupported-tags list (or in documentation describing the check), but the Rust renderer has fully supported template inheritance since PR #272.

**Current state** (as of v0.3.3): `_UNSUPPORTED_TAGS_RE` in `checks.py` flags:
```
ifchanged, regroup, resetcycle, lorem, debug, filter, autoescape
```

`{% extends %}` and `{% block %}` are **not** in this regex and should not be. But any documentation, comments, or third-party references that list them as unsupported are now wrong and mislead developers into thinking they need to avoid template inheritance.

## Action items

1. Audit all docs, README files, and in-code comments for references to `extends`/`block` being unsupported — remove or correct them
2. Add a comment in `checks.py` near `_UNSUPPORTED_TAGS_RE` listing tags that *were* added in v0.3.3 (`widthratio`, `firstof`, `templatetag`, `spaceless`, `cycle`, `now`) so the history is clear
3. Consider adding a regression test that confirms T011 does **not** fire on templates using `{% extends %}` or `{% block %}`

## References

- PR #272 (template inheritance implementation)
- PR #287 (filter additions in v0.3.3)

PR #1884, opened 2026-06-23
test(infra): autouse fixture resets djust process-globals — retires the shared-global flaky class (#1883, #1882)

johnrtipton:
## Summary

Three shared-process-global test-pollution flakes in two milestones were all the **same class** — a process-global left dirty across tests in an xdist worker:

- **#1862** — `ROOT_URLCONF` leak broke `TestDemoRegistration` (fixed PR #1874, per-test)
- **#1875** — `djust_hotreload` channel-layer group pollution (fixed PR #1881, per-test)
- **#1882** — process-global wire-version drift: a stray `djust_hotreload` frame on the cached `InMemoryChannelLayer` re-renders on a later consumer and bumps its per-connection `_next_version()` counter, so `test_time_travel_jump_recovery_version_is_current` saw the time-travel jump land at version **4 instead of 3** under `-n auto` (intermittent under saturation; passed in isolation) — **still open**

Each was whack-a-moled per-test. This PR is the **systemic cure**: one autouse fixture that resets djust's process-globals BEFORE each test, so the flaky-class can't recur — and it closes #1882 as a consequence.

## What changed

New shared helper **`djust.test_isolation.reset_djust_globals()`** (DRY, #1646), called by an autouse **`_reset_djust_globals`** fixture in **both** test roots:

- `tests/conftest.py` — mirrors the existing `cleanup_session_cache` autouse fixture
- `python/djust/tests/conftest.py`

### Process-globals reset (and why each genuinely leaks + is lazily re-derived)

| Global | Reset | Class |
| --- | --- | --- |
| Channels layer manager | `channel_layers.backends.clear()` | #1875 / **#1882** — `LiveViewConsumer.connect` joins the process-global `djust_hotreload` group; the cached `InMemoryChannelLayer` retains group membership + buffered frames. A stray hotreload frame re-renders on a later consumer → `_next_version()` bump → the #1882 drift. Fresh layer per test. |
| URLconf caches | `clear_url_caches()` + `set_urlconf(None)` | #1862 — a test swapping `ROOT_URLCONF` can leave the resolver cache + thread-local urlconf pinned. |
| djust route-map cache | `_reset_route_map_cache()` | #1862-adjacent — URLconf-derived `dj-navigate` route map. |
| Id counters | fresh `itertools.count(1)` for `mixins.sticky._view_id_counter` + `components.templatetags.djust_components._tooltip_id_counter` | module-level counters; deterministic `child_N` / tooltip ids per test. |

[...]

PR #2023, opened 2026-07-03
feat(templates): support Django {% regroup %} via built-in assign-tag handler

szto:
## What

Django's `{% regroup %}` renders as an HTML comment (UnsupportedTag) in the Rust renderer, so any template using it silently loses that block. This adds built-in `{% regroup %}` support with Django-parity semantics.

## Approach

Smallest-footprint path: a built-in **assign-tag handler** (`djust.template_tags.regroup`, registered via the existing `register_assign_tag_handler` infra) instead of a native Rust node.

- Django parity: consecutive grouping (no pre-sort), result is `[{grouper, list}, ...]`, dotted lookups (`by author.team`) and dict keys supported (mirrors `defaulttags.RegroupNode`)
- One minimal Rust change: `resolve_tag_arg` now JSON-encodes list/object args instead of collapsing to `"[List]"` (mirrors the CustomTag path; scalar behavior unchanged — existing assign-tag tests untouched and passing)
- `regroup` removed from the unsupported-tag list / T011 check (pattern from #397)

## Known limitations (documented in the module)

- filter expressions in the source (`{% regroup items|dictsort:"k" by k %}`) are not supported yet
- a context key that happens to match the `by <attr>` token shadows the per-item lookup (assign-arg resolution order)

## Tests

TDD: new `test_regroup_tag.py` — basic grouping, non-contiguous groupers (Django-identical), nested attrs, full template render (`g.grouper`/`g.list`), and the `render_with_diff` LiveView path. 19 passed (new + existing assign-tag suite). Full regression: Rust 66+37 passed + clippy clean; Python 8663 passed / 32 skipped (serial and `-n auto`); ruff + mypy clean.


johnrtipton (commented):
## Code Review — pipeline-ship Stage 7

Thanks for this, @szto — it's a genuinely well-scoped contribution. The Rust `resolve_tag_arg` change is minimal and correctly targeted, the handler is faithful to Django's `RegroupNode` semantics, and the test file drives the *real* Rust engine (both `render_template` and `RustLiveView.render_with_diff`) with a Django-parity anchor. Below are the findings from an adversarial pass against our template-pipeline canon (dual-engine, parallel-path drift, test quality, security).

**Verdict:** REQUEST CHANGES — the code is sound and CI-green, but a `feat:` PR must carry a CHANGELOG entry (repo policy), and a couple of documentation items mask a real silent-wrong-output footgun worth addressing before merge.

---

### 🔴 Blocking

**1. `CHANGELOG.md` has no entry for this feature.** Per our contribution policy (`CLAUDE.md` → Testing Expectations: "`feat:` and `fix:` PRs must update CHANGELOG.md"), please add an `### Added` bullet under the `## [Unreleased]` section (`CHANGELOG.md:8`) describing `{% regroup %}` support in the Rust engine.
- Note: the older line `CHANGELOG.md:6623` still lists `regroup` among tags "not yet implemented in the Rust renderer", but that line lives under `## [0.3.3rc1]` — an already-released, immutable section. **Do not edit it.** The new `[Unreleased]` entry is the correct place to record that regroup is now supported; the released historical record stays as-is.

---

### 🟡 Non-blocking

**2. `AssignTagHandler` docstring contradicts the actual arg flow** (`python/djust/template_tags/__init__.py:180-192`). The docstring states assign handlers "receive the **raw** (unresolved) tag args and resolve them against the supplied context themselves via `TagHandler._resolve_arg`" and that this "avoids resolving attribute-name operands (the `by <attr>` token) as if they were context variables." That is not what happens: `render_nodes_with_loader` (`crates/djust_templates/src/renderer.rs:165-167`, and the three sibling paths at `:277`, `:349`, `:1220`) applies `resolve_tag_arg` to **every** arg before calling the handler. `regroup.py`'s own module docstring (`python/djust/template_tags/regroup.py:16-19`) describes the true behavior. Please align the `AssignTagHandler` docstring with reality — the false claim directly masks finding #3.

**3. Limitation (b) is a silent-wrong-output footgun that deserves a louder guard** (`python/djust/template_tags/regroup.py:26-30`). Because the Rust engine resolves *all* assign-tag operands against the context, a context key whose name matches the `<attr>` token (`country`, `category`, `gender`, `type`, …) shadows the per-item lookup — `args[2]` arrives as that variable's value instead of the literal attribute name, and the grouping is silently wrong (no error, no log). djust auto-exposes public view attributes to the template context, so collisions with common attribute names are plausible in real apps, and the failure is invisible. Suggestions, in increasing order of effort:
   - At minimum, emit a `logger.warning("regroup: attr operand %r resolved to a non-identifier; a context key may be shadowing it", attr)` when the resolved `attr` doesn't look like a bare dotted identifier, and elevate limitation (b) from a docstring footnote to a prominent WARNING.
   - The durable fix is to pass the `by` / `<attr>` / `as` operands to the handler **un-resolved** (Django never resolves the attr against the outer context). That is a deeper change to the assign-tag arg path, so a follow-up issue is a reasonable landing spot — but please file it rather than leaving it only in a docstring.

**4. Parallel-path drift: `BlockCustomTag` still collapses list/object args to `[List]`** (`crates/djust_templates/src/renderer.rs:1178-1186`). This PR brings the AssignTag path (`resolve_tag_arg`) into parity with `CustomTag`'s JSON-aware `value_to_arg_string` (`:1252-1259`), but the `BlockCustomTag` branch's inline resolver still uses `resolved.to_string()` / `format!("{}={}", key, resolved)`, which emits the opaque `[List]`/`[Object]` placeholder. It's pre-existing and out of scope for regroup (regroup is an assign tag), so not a regression — but per our parallel-path-drift canon (#1646) it's the un-updated sibling. Ideally route all three branches (`CustomTag`, `AssignTag`, `BlockCustomTag`) through one shared `value_to_arg_string` helper; at minimum, file a follow-up so the placeholder-collapse class is retired rather than fixed 2-of-3.

---

### 🟢 Nits

**5. "Mirrors `Node::CustomTag`" is *almost* exact — one intentional difference worth a comment** (`renderer.rs:161-164`, `:239`). `CustomTag` resolves variables via `get_value(...)` (filter-aware, e.g. `x|upper`), while `resolve_tag_arg` uses `context.get(...)` (no filter support). This is consistent with regroup's documented limitation (a) (filter expressions on the source unsupported) and is fine — but the "same JSON-aware semantics as `Node::CustomTag`" comment slightly overstates parity. A one-line note that only the *JSON-encoding* is mirrored, not the *resolution mechanism*, would prevent a future reader assuming filter support.

**6. Two tests would pass with the change gated off** (`test_regroup_empty_source_yields_no_groups`, `test_regroup_missing_source_variable_is_safe`, `python/djust/tests/test_regroup_tag.py:123-132`). Both assert `out == ""`, which holds whether or not the JSON-encoding change is present (empty either way). The suite as a whole is non-tautological — the grouping / input-order / dotted-path / diff-path / Django-parity tests all fail if the change is reverted — so this is only a note, not a defect.

**7. `test_isolation.py` is correctly placed and its addition is appropriate.** For the record (the reviewer brief flagged its location): `python/djust/test_isolation.py` is a **pre-existing** module (added in PR #1884, not this PR) — a test-isolation *utility* (`reset_djust_globals`), not a pytest test module despite the `test_` prefix. This PR's +32 lines add `_reset_builtin_template_tags()` → `reregister_builtins()` and wire it into the reset, which is exactly the right #1928-class fix: a `clear_assign_tag_handlers()` polluter would otherwise leave `regroup` unregistered for the rest of an xdist worker. The name is a pre-existing smell (it reads like a collectible test file), but that predates this PR and is out of scope here.

---

### Cross-PR note (informational)

The `.github/workflows/security-hotspot-check.yml` change (+81/-54) wrapping the fork-PR label/comment `github-script` steps in `try/catch` appears **identical** across your #2040, #2023, and #2022. It's benign — it turns an expected 403 on a fork PR's read-only `GITHUB_TOKEN` into a warning rather than a failure, and this workflow is not part of the `test-summary` merge-gate AND-condition, so it doesn't weaken a real gate. However, it's an unrelated CI fix bundled into a feature PR; whichever of the three merges first makes the other two redundant. 🟡 non-blocking — consider extracting it to its own standalone PR so the regroup feature and the CI fix have independent histories.

---

### Local verification

I did **not** rebuild the Rust extension and did **not** execute `test_regroup_tag.py` locally, deliberately: the new behavior depends on the `renderer.rs` `resolve_tag_arg` change, and the only prebuilt `.so` available to this worktree carries the *old* behavior — running against it would produce misleading false failures, and rebuilding risks clobbering the shared extension. Instead I relied on:
- **CI-green** — `rust-tests` (confirms the Rust build compiles + the full Rust suite passes) and `python-tests` (3.12 + 3.14t). Critically, CI now runs `python/djust/tests/` as a blocking step (`.github/workflows/test.yml:181`, promoted by #2034), so `test_regroup_tag.py` — which lives there — is genuinely covered by the green run.
- **Close reading** of `renderer.rs`, `regroup.py`, `template_tags/__init__.py`, `test_isolation.py`, and `test_regroup_tag.py`.
- **Static verification**: `resolve_tag_arg` has exactly four callers, all AssignTag branches (`renderer.rs:167/277/349/1220`); `CustomTag` (`:1237`) uses its own pre-existing inline `value_to_arg_string`; `register_assign(` is used exactly once (regroup is the only assign tag, so the scalar-path-unchanged change regresses no existing assign behavior); `mark_safe` / f-string logging / `print(` greps on the new Python are all clean.
- **Mental gate-off** (per #1468): reverting `value_to_arg_string` makes a list arg resolve to `"[List]"`, so `json.loads("[List]")` raises → `_decode_source` falls to `_lookup(context, "[List]")` → `None` → `[]` → the grouping/parity/diff tests render `""` and fail. The behavior-meaningful tests are therefore non-tautological.

**Parallel-path result:** the PR's "mirrors the CustomTag path" claim is **verified** — `CustomTag` already JSON-encoded list/object args via `value_to_arg_string`; this PR brings the shared `resolve_tag_arg` into parity. Scalar behavior is byte-identical (`v.to_string()` fall-through). One un-updated sibling found: `BlockCustomTag` (finding #4).

---

Merge is intentionally deferred to the maintainer — this review posts findings only and does not approve or merge.

szto:
Thanks for the thorough review, @johnrtipton. Addressed all findings (pushed `2f84912e → 1937a10d`, two commits: impl+tests / CHANGELOG):

**🔴 #1 — CHANGELOG.** Added an `### Added` entry under `## [Unreleased]` describing `{% regroup %}` support. Left the historical `[0.3.3rc1]` "not yet implemented" line untouched per the tagged-section-immutability rule. `check-changelog-tagged-sections` + `check-changelog-test-counts` both pass.

**🟡 #2 — `AssignTagHandler` docstring vs reality.** Rewrote it: the Rust engine resolves **every** assign-tag arg via `resolve_tag_arg` (`renderer.rs:167/277/349/1220`) *before* the handler runs, JSON-encoding structured values — handlers receive already-resolved args, not raw tokens. The docstring now states this and points at `RegroupTagHandler` for the shadow edge, matching `regroup.py`'s own module docstring.

**🟡 #3 — silent-wrong-output footgun.** Did both suggestions:
- Elevated limitation (b) to a prominent **WARNING** block in the module docstring.
- `RegroupTagHandler.render` now emits `logger.warning("regroup: attr operand %r resolved to a non-identifier; a context key may be shadowing…")` when the resolved `<attr>` fails a bare-dotted-identifier check (`_IDENTIFIER_RE`). Two tests pin it (fires on a shadowed non-identifier attr, silent on `author.team`); gate-off verified — neutering the guard fails the positive case.
- Filed the durable fix (pass `by`/`<attr>`/`as` unresolved) as **#2041**, referenced from code + docstring + CHANGELOG.

**🟡 #4 — `BlockCustomTag` parallel-path drift (#1646).** Out of scope for regroup as you noted; annotated the un-updated inline resolver at `renderer.rs` with the drift note and filed the shared-`value_to_arg_string` unification as **#2042**.

**🟢 #5 — nit.** Added a one-line note on the `AssignTag` resolve comment clarifying only the JSON *encoding* mirrors `CustomTag`, not the *resolution mechanism* (`context.get`, no filter support).

**🟢 #6/#7 — notes only, no change needed.** Agreed the two `== ""` cases are non-tautological in aggregate; `test_isolation.py` is pre-existing (#1884).

**Cross-PR CI note:** the `security-hotspot-check.yml` change is still bundled here — I'll drop it from whichever two of #2040/#2023/#2022 land second, on rebase, once the first merges.

Local verification (Rust extension rebuilt via `make dev-build`): `test_regroup_tag.py` → **12 passed** (10 + 2 new); the template-tags/isolation regression band → **40 passed**.

johnrtipton (commented):
## Code Review — pipeline-ship Stage 7 (re-review of `75159d02`, `1937a10d`)

Thanks for the fast, thorough turnaround, @szto — every finding from the previous review is addressed, and the responses are well-judged (fix what's in scope, document + file follow-ups for the rest).

**Verdict: APPROVE** — the blocking finding is resolved and CI is green; merge is deferred to the maintainer.

### Resolved
- 🔴 → ✅ **CHANGELOG** — new `### Added` entry under `## [Unreleased]` (`CHANGELOG.md:10`), correctly placed above `### Fixed` and leaving the immutable `[0.3.3rc1]` section untouched. Accurate and appropriately detailed.
- 🟡 → ✅ **`AssignTagHandler` docstring** (`template_tags/__init__.py:180-193`) — now correctly states args are Rust-resolved *before* `render`, cites the real `renderer.rs` branch lines, and explains the shadowing consequence. No longer misleading.
- 🟡 → ✅ **Shadow-attr footgun** — `_IDENTIFIER_RE` guard + `logger.warning` (`regroup.py:76-85`, `%r` parameterised — no f-string), a prominent docstring **WARNING**, and follow-up **#2041** for the durable fix (pass operands unresolved). Two new tests, including a negative gate so the positive assertion can't pass tautologically.
- 🟡 → ✅ **Parallel-path drift (#1646)** — `BlockCustomTag` (`renderer.rs:1188-1193`) now carries an explicit NB comment marking it as the un-updated sibling, with the shared-helper cure tracked in **#2042**. Correctly left out of this PR's scope.
- 🟢 → ✅ **"mirrors CustomTag" nit** — the `renderer.rs:161-168` comment now distinguishes the JSON *encoding* (mirrored) from the *resolution mechanism* (`get_value` filter-aware vs. plain `context.get`), consistent with the documented no-filter limitation.

### Verification
- CI green on `1937a10d` (rust-tests, python-tests 3.12 + 3.14t, security-tests, Analyze — 13/13 real checks pass).
- Ran the two new warning tests locally against the handler: **2 passed**. Gate-off proof — disabling the `_IDENTIFIER_RE` guard turns `test_handler_warns_when_attr_operand_is_shadowed` **red** while `test_handler_does_not_warn_for_normal_identifier` stays green → non-tautological.
- Both cited follow-ups exist and are accurately scoped: #2041 (OPEN), #2042 (OPEN).

### One residual note (🟢, non-blocking — already covered by #2041)
The `logger.warning` is a best-effort detector: it fires only when the resolved `<attr>` isn't identifier-shaped. A shadow whose *value* happens to look like a bare identifier (e.g. a context key `country` resolving to `"USA"`) still slips through silently. That's inherent to the post-resolution vantage point and is exactly what #2041's "pass operands unresolved" fix removes — no action needed here; just flagging so the limitation is on record.

Nice work. Once a maintainer merges, please consider extracting the `security-hotspot-check.yml` fix into its own PR (still byte-identical across #2040/#2023/#2022 — first to merge makes the others redundant on that file).

johnrtipton (approved):
Great, thanks so much for this!

Issue #2001, opened 2026-07-02
docs/demo: sync `@background` cannot be interrupted mid-run, and `cancel_async` silently no-ops on a task-name mismatch — both confirmed live in djust's own shipped demo


johnrtipton:
## Symptom (DJUST_LESSONS gotcha #2 and #5)

A "Stop" button whose handler flips `self.streaming = False` never actually interrupts a running sync `@background` generation loop — the loop runs to completion with the flag checked but the *handler that would flip it* queued behind the very generation it's supposed to stop.

This is not hypothetical — it reproduces in djust's own shipped demo, `examples/demo_project/djust_demos/views/markdown_stream_demo.py`:

```python
@event_handler
def reset(self, **kwargs):
    """Clear output."""
    self.cancel_async("md_stream")      # L71 — name mismatch, see below
    self.streaming = False
    self.llm_output = ""

@background
def _stream_chars(self):               # L76 — SYNC background handler
    """Background task: append one character every ~25 ms until done."""
    import time
    for ch in _SAMPLE_REPLY:
        if not self.streaming:
            return
        self.llm_output += ch
        # Short sleep — the re-render after each assignment pushes a
        # VDOM patch to the client.
        time.sleep(0.025)               # L86
    self.streaming = False
```

Two independent, confirmed bugs live in this one 20-line demo.

## Root cause 1 — sync `@background` runs on a single shared thread-sensitive executor; another handler cannot dispatch until it's done

`@background`'s sync-path wrapper (`python/djust/decorators.py:1086-1092`):
```python
else:
    # Sync handler: plain closure, run in thread via sync_to_async.
    @functools.wraps(func)
    def wrapper(self, *args, **kwargs):
        def _async_callback():
            return func(self, *args, **kwargs)
        task_name = func.__name__
        self.start_async(_async_callback, name=task_name)
```
dispatches through `_run_async_work` (`python/djust/websocket.py:1153`):
```python
result = await sync_to_async(callback)(*args, **kwargs)
```
[...]

PR #2030, opened 2026-07-04
fix(async): converge WS + runtime async-callback dispatch onto one helper (#2020)

### LINKED ISSUE #2020: tech-debt: transport-parity test for the async-work dispatch path (runtime vs consumer) — the #2016 sync_to_async/coroutine twin had no guard
**Source**: v1.1.0-5 retro (PR #2016 / #2001 / #2002).

## Problem
While fixing the markdown-stream demo (#2001/#2002), a real framework bug surfaced: `runtime.py:ViewRuntime._execute_async_task` unconditionally did `await sync_to_async(callback)(...)`, which raises `TypeError: sync_to_async can only be applied to sync functions` for an **async** callback. So async `@background` / `start_async(coroutine_fn)` **silently failed** on the ADR-022-converged WS-event path (the live path post-convergence).

The consumer twin `websocket.py:_run_async_work` (~L1150-1155) already had the correct `asyncio.iscoroutinefunction(callback)` branch. The ViewRuntime convergence dropped it — a textbook #1646 parallel-path-drift twin (the same class that recurred ~6× across v1.0.x). PR #2016 fixed it by mirroring the consumer's check (+ a defensive `inspect.iscoroutine(result)` unwrap).

## Ask
Add a structural/transport-parity test (mirroring the #291 multiplexed-path and #1448 wire-protocol-pinning discipline, but for the **async-work dispatch axis**) that asserts the runtime `_execute_async_task` and consumer `_run_async_work` agree on sync-vs-async callback handling — so a future convergence/refactor that drops the coroutine branch on one path goes RED. Ideally a single shared `_dispatch_async_callback` helper both paths call (the #1646 structural cure), pinned by a test that both a sync and an async callback round-trip through it.

## Context
- Fix: `python/djust/runtime.py:ViewRuntime._execute_async_task` (PR #2016).
- Twin: `python/djust/websocket.py:_run_async_work`.
- Related canon: CLAUDE.md #1646 parallel-path-drift; #291 multiplexed-path transport rule; #1448 wire-protocol pinning.

johnrtipton:
## Summary

Closes #2020 (v1.1.0-6 drain). #2016 fixed a #1646 parallel-path drift — an async `@background` handler silently failed on the converged runtime path because `ViewRuntime._execute_async_task` routed **every** callback through `sync_to_async` (raising `TypeError: sync_to_async can only be applied to sync functions`, #2001) while the WS consumer's `_run_async_work` awaited async callbacks directly. But #2016 fixed it by **copying** the `iscoroutinefunction(callback)` branch into the runtime path, leaving two byte-identical copies primed to re-drift on the next edit.

## The cure (#1646 structural, not another copy)

Extract the dispatch into ONE `run_async_callback` in `mixins/async_work.py`. Both `LiveViewConsumer._run_async_work` (WS) and `ViewRuntime._execute_async_task` (SSE + `url_change`, and WS post ADR-022 flip) now delegate to it — the sync/async handling can never diverge between transports again.

## Parity test (`test_async_dispatch_parity_2020.py`)

**Behavioral** — the shared helper across all shapes both paths feed it:
- async `def` awaited directly (the **gate-off sentinel**)
- sync thread-dispatched
- sync-returning-coroutine unwrapped (pre-v0.4.2 `@background` contract)
- empty args/kwargs defaults

**Structural pins** (#1646/#1125) — the actual transport-parity guard:
- both transports call `run_async_callback`
- neither keeps its own `iscoroutinefunction(callback)` branch
- the helper is the single definition in the package

## Verification

- 7/7 new tests green; 96 existing async-work/background/#2001/#2016 tests still green after the refactor.
- **Gate-off (#1468):** reverting the helper to always-`sync_to_async` makes the async-def case fail with the exact `TypeError: sync_to_async can only be applied to sync functions` — proving the test guards the real bug.

Two-commit shape: `9c0f8b6c` (code+tests, no CHANGELOG) + `9907b8d7` (CHANGELOG docs-only).


johnrtipton:
## Retrospective — PR #2030 (pipeline-drain)

**Task**: v1.1.0-6 drain — closed #2020 (transport-parity guard for the async-work dispatch path; the #2016 `sync_to_async`/coroutine twin had no guard).

### Quality: 5
### What went well
- **Structural cure over a bare test (#1646).** #2016 fixed the drift by *copying* the `iscoroutinefunction(callback)` branch into the runtime path, leaving two byte-identical copies primed to re-diverge. Rather than add a test that watches two copies, extracted one shared `run_async_callback` in `mixins/async_work.py` that both `LiveViewConsumer._run_async_work` and `ViewRuntime._execute_async_task` delegate to — the paths now *can't* diverge.
- **Parity is pinned two ways**: behavioral (4 shapes over the shared helper) + structural (#1125 pins: both paths call the helper, neither keeps its own dispatch branch, helper is the single definition).
- 96 existing async-work/background/#2001/#2016 tests still green after the refactor.

### What didn't
- The single-source-of-truth pin self-matched (its own search-needle string counted as a definition) — split the needle + excluded `tests/` dirs.

### Verified
- 7/7 new tests green (`test_async_dispatch_parity_2020.py`).
- **Gate-off (#1468):** reverting the helper to always-`sync_to_async` makes the async-def case fail with the exact `TypeError: sync_to_async can only be applied to sync functions` — the real #2001 bug.

RETRO_COMPLETE

PR #2031, opened 2026-07-04
fix(vdom): re-prime diff baseline on reconnect/state-restore mount (#1977)

### LINKED ISSUE #1977: VDOM: stale last_vdom baseline after state-restore/reconnect → SetText patches land on #text (recovery)

johnrtipton:
## Summary

Closes #1977 (v1.1.0-6 drain). The maintainer bit-exact-reproduced the true cause in real Chromium and **retracted the original "diff_html misroutes SetText" framing** — the Rust differ and client patcher are confirmed CLEAN on the normal same-connection path. The real bug is on the **WebSocket reconnect / state-restore seam**:

On a reconnect, the fresh view's Rust diff baseline (server "last_vdom") is primed from a render that does NOT match the client's pre-disconnect DOM (which already shows the restored/filtered state). The first post-restore **event** is then diffed against that stale baseline → `SetText` patches land on the wrong node (often a bare `#text`) → `2/N patches failed` → `html_recovery` reload/flicker. Intermittent because it only fires after a reconnect/restore between events (sleep, network blip, server restart) — never on a fresh page, which is why the single-instance `LiveViewTestClient` could not reproduce it.

## Fix (Python-only — Rust untouched)

In `ViewRuntime.dispatch_mount`, right after `_mounted_from_restore` is set: when `mounted_from_restore`, set `view._force_full_html = True` so the first post-restore render emits a full `html_update` frame. The client morphs wholesale and the Rust baseline is re-primed to the live DOM — no stale-baseline diff can reach the client.

**Render-timing (traced, load-bearing):** the mount-time `render_with_diff` only *reads* `_force_full_html`; every *reset* lives in the transport-level event/url_change dispatch (`_render_and_send` 4093-4094, `dispatch_url_change` 3725-3726, WS twins 1650-1651/4115-4116). On a resume the mount HTML is dropped (`skip_html_for_resume`), so the flag **survives the mount frame and is consumed by the first event render**. Snapshot-excluded via `_FRAMEWORK_INTERNAL_ATTRS`, so it never leaks into a saved snapshot.

**Parallel-path (#1646):** one guard covers both restore mechanisms (session-saved-state + signed-snapshot HMAC) since both funnel to `mounted_from_restore`; WS `handle_mount` is a thin shim to `dispatch_mount`, so WS+SSE+runtime are covered. HTTP GET renders the full page (baseline primed against the served state) and HTTP POST re-restores+re-renders atomically per request — neither has a persistent mount baseline a later event diffs against, so neither needs the fix.

## Verification

- 3 tests in `test_stale_baseline_restore_1977.py` — drive `dispatch_mount` (session-restore + signed-snapshot) + `dispatch_event` via a `MockTransport` with a real Django request + DB session; the restore cases assert the first event is a full `html_update`, the fresh-mount case asserts a normal `patch` (scoped gate-off sibling). Independently re-run green in the main checkout (3 passed).
- **Gate-off (#1468):** commenting out the `_force_full_html = True` line makes both restore tests go RED (`'_FilterView' object has no attribute '_force_full_html'`) while the fresh-mount test stays green.

> **Note (not a regression in this PR):** the drain surfaced that CI's Python job runs `pytest tests/ python/tests/` and **omits `python/djust/tests/`** (which is in pyproject `testpaths`), so this PR's test — and the pre-existing stale line-pins in `test_mount_chokepoint_structural.py` — aren't CI-gated. Filed as a follow-up; validated locally here.

Two-commit shape: `d2b351c1` (code+tests, no CHANGELOG) + `ff36bbb9` (CHANGELOG), plus a `origin/main` union-merge for the shared `[Unreleased]` block.


johnrtipton:
## Retrospective — PR #2031 (pipeline-drain)

**Task**: v1.1.0-6 drain — closed #1977 (stale `last_vdom` baseline after reconnect/state-restore → SetText lands on `#text`).

### Quality: 5
### What went well
- **Symptom-up triage paid off (repo canon).** The issue's original title blamed `diff_html`; my prior-session 18-shape faithful search + the maintainer's bit-exact Chromium repro had already *retracted* the differ-misroute claim and pinned the real cause to the reconnect baseline seam. Trusting the symptom (bit-exact console) over the cited path (the differ) meant a Python-only fix and zero Rust churn.
- **Reused the existing `_force_full_html` lever** rather than inventing a re-prime mechanism — one guard at the converged `dispatch_mount` restore point.
- **Render-timing was *traced*, not assumed (#1198 class):** verified the flag survives the (skipped) resume mount frame and is consumed by the first *event* render — the fix would be a no-op if a mount-time render reset it.
- **Parallel-path (#1646) decided per seam:** session-restore + signed-snapshot both funnel to one guard; WS `handle_mount` is a shim; HTTP GET/POST need no fix (verified, not hand-waved).

### What didn't
- The subagent's broader sweep surfaced that **CI omits `python/djust/tests/`** (this PR's test + a pre-existing RED security-structural test live there, un-gated) — filed #2032. Validated this PR's tests locally instead.

### Verified
- 3 tests green (session-restore + signed-snapshot → full `html_update`; fresh-mount → normal `patch` scoped sibling); independently re-run in the main checkout.
- **Gate-off (#1468):** removing `_force_full_html = True` → both restore tests RED (`'_FilterView' object has no attribute '_force_full_html'`), fresh-mount stays green.

RETRO_COMPLETE

PR #2036, opened 2026-07-04
fix(vdom): attribute/identity-driven teardown for dj-virtual on SPA nav (#2033)

### LINKED ISSUE #2033: dj-virtual client state has no attribute-driven teardown — reused container across a view-identity change merges stale content (regression surface for #1988/#1989)
## Symptom (found live-verifying the #1988/#1989 fix on 1.1.0rc6)

Sequence: view a virtualized `[dj-virtual]` thread → navigate (SPA-style, same
page, no full reload) to a small, non-virtualized thread → navigate again
(e.g. switch workspace) to a different small thread. The message pane renders
**overlapping, garbled content** — a leftover row from a previously-viewed
thread rendered at the exact same `translateY` offset as the new thread's real
messages.

DOM inspection at the point of failure:

```js
document.querySelector('.stage-inner').attributes
// dj-virtual="messages", dj-virtual-item-height="140", dj-id="29"
// — present even though this thread has only 2 messages and the app's own
//   template only emits dj-virtual past a large-thread threshold.

document.querySelector('[data-dj-virtual-shell]').children
// 3 children instead of the expected 1-2:
//   <div> (no class) — leftover content from the PREVIOUSLY viewed thread
//   <div class="row-user">   — NEW thread's real message, SAME rect as above
//   <div class="row-asst">   — NEW thread's real reply
```

A full page reload (fresh `initVirtualLists()` pass over freshly-parsed DOM)
fixes it immediately — this is client-side session state surviving across
in-app navigation, not data corruption.

## Root cause (grep-verified against the 1.1.0rc6 source)

`dj-virtual` client state lives in a `WeakMap` keyed by the **container DOM
node's identity** (`STATE = new WeakMap()`, `29-virtual-list.js:42`). When an
app's SPA-style navigation reuses the same physical container node across a
view/data-source change (rather than remounting), nothing tears the
virtualization down just because the underlying data source changed or the
server's re-render removed the `dj-virtual` attribute:

- `structureIntact(container, state)` (`29-virtual-list.js:421-428`) only
  checks whether `state.shell`/`state.spacer` are still parented under
  `container` with their marker attributes intact — it never checks whether
  `container` **still carries the `dj-virtual` attribute at all**. So if a
  server patch legitimately removes `dj-virtual="messages"` (the new view
  doesn't need virtualization), `initVirtualLists()`/`refreshVirtualList()`
  still treat the container as virtualized, because `structureIntact()`
  reports true (the OLD shell/spacer are still exactly where they were).
- `initVirtualLists(root)` (`29-virtual-list.js:474+`) iterates
  `[dj-virtual]` containers and only acts on ones it finds via that selector
  — but it never scans for containers that were **previously** tracked in
  `STATE` and have since lost the attribute, so there is no code path that
  ever calls `teardownVirtualList()` in response to attribute removal.
- Compounding this on 1.1.0rc6 specifically: `refreshVirtualList()`'s new
  `absorbLooseChildren(container, state)` (`29-virtual-list.js:451-472`)
  **accumulates** loose children into `state.items` via `push()` rather than
  resetting the pool for a new data source — so content from an entirely
  different view (the previous thread) and the new view's content both end
  up in the same item pool and get windowed/rendered together, landing at
  the same computed offset when the pool is small.

This is a new-in-1.1.0rc6 manifestation of an old gap: before this release,
`refreshVirtualList()` was not wired into `reinitAfterDOMUpdate()` at all (per
the #1988/#1989 fix), so a stale container would likely have just shown
unrefreshed OLD content (stale but not visibly merged) rather than silently
absorbing NEW content into the OLD pool. The new automatic self-heal (a real
improvement for the normal case — same-thread server re-renders) has no
signal to distinguish "this patch is for the same virtualized list" from
"this patch is an entirely different view that happens to reuse the same DOM
node."

## Scope

1. **`structureIntact()` should also check attribute presence**: if
   `container.getAttribute('dj-virtual') == null` while `STATE.has(container)`,
   treat it as torn-down intent and call `teardownVirtualList()` (restoring
   `state.items` is wrong here too — see point 2) rather than treating it as
   "structure still intact, keep going."
2. **A view-identity change should reset the item pool, not absorb into it.**
   `absorbLooseChildren` currently assumes any loose child is a same-list
   append (correct for the streaming-chat-thread case #1989 targets). There's
   no way for it to distinguish that from "this loose child belongs to a
   completely different data source." A `dj-key`-scoped reset (e.g. if the
   container's own `dj-key`-equivalent identity attribute, or a caller-set
   generation marker, changes) would let `refreshVirtualList` correctly
   discard the old pool instead of merging into it.
3. Short of an internal fix, a documented escape hatch would help: e.g. a
   `container.dispatchEvent(new CustomEvent('dj-virtual:reset'))` contract, or
   an exported `djust.resetVirtualList(container)` that fully clears
   `state.items` and re-snapshots from the container's current children,
   which app code could call from its own router/navigation hook when it
   knows the underlying data source changed but happens to reuse the same DOM
   node.

## Testing

A regression test simulating: `setup()` a container with dataset A → replace
`container`'s children with dataset B's raw rows via `innerHTML` (simulating
a server patch for an unrelated view reusing the same node) → call
`refreshVirtualList(container)` → assert the item pool equals dataset B only,
not A ∪ B, and that rendered rows don't share a `translateY` offset with
stale content.

## Related

- Filed alongside #1988/#1989 investigation (same downstream project,
  `djust-chat`, DJUST_LESSONS gotcha #23) — a regression surface exposed by
  #1988/#1989's own fix (the new automatic self-heal), not something either
  of those issues was responsible for on their own.
- Adjacent to the still-open #2017 (out-of-window finalize-patch landing) —
  both are cases of `absorbLooseChildren`/`refreshVirtualList` not having
  enough context to know when NOT to merge.

johnrtipton:
## Summary

Closes #2033 (v1.1.0-7 drain). Found live-verifying the #1988/#1989 fix on 1.1.0rc6: SPA-style nav (virtualized thread → small non-virtualized thread → another small thread, no reload) rendered **overlapping/garbled rows** — a leftover row from a previous thread at the same offset as the new thread's messages.

**Root cause (grep-verified):** `dj-virtual` client state lives in a `WeakMap` keyed by container DOM node identity. When SPA nav reuses the *same* physical container across a view/data-source change, nothing tears the virtualization down: `structureIntact()` never checked the container still carried `dj-virtual`; `initVirtualLists()` never scanned for previously-tracked containers that lost the attribute; and (new in 1.1.0rc6) `absorbLooseChildren()` `push()`ed cross-view content into one pool.

## Fix (`29-virtual-list.js`, +212 B gzipped)

- **`identityChanged(container, state)`** — true when `dj-virtual` was removed, its source value changed, or the host `dj-id` changed (init-time snapshot of `virtualValue`/`viewId`).
- **`structureIntact()`** now fails closed via `identityChanged()` — a repurposed same node can't be treated as still-virtualized.
- **`reapStaleVirtualLists(scope)`** — the discovery path. The `WeakMap` isn't iterable, so it scans the never-removed `[data-dj-virtual-shell]` marker (cited in the issue's DOM dump) → `parentNode` → tears down any tracked container whose identity changed. Called first in `initVirtualLists()`.
- **`detachStaleState()`** — drops observers/listeners/STATE and strips the stale shell+spacer (the morph already authored the new content).
- **`absorbLooseChildren()`** no-ops on identity change + skips marker children → no cross-source pool merge.
- **`refreshVirtualList()`** tears down + re-setups fresh on identity change.

Reused the existing `data-dj-virtual-shell` marker (no new attribute).

## Verification

- 3 new cases in `tests/js/dj-virtual-teardown-2033.test.js` (attr-loss teardown; identity change re-virtualizes fresh with no merge; **regression guard**: same-thread re-render still self-heals via #1988/#1989). 27/27 with existing virtual-list tests; full JS suite 1781 pass; eslint `--max-warnings 0` clean. Independently re-run green.
- **Gate-off (#1468):** forcing `identityChanged()` to `return false` reds the two identity-change tests while the same-thread regression guard stays green.

(The 2-byte `debug-panel.min.js.map` change is benign build-tooling map drift — debug-panel doesn't contain the virtual-list module.)

Two-commit shape: `97e21c95` (src+bundles+tests) + `1b6d4d3b` (CHANGELOG).


johnrtipton:
## Retrospective — PR #2036 (pipeline-drain)

**Task**: v1.1.0-7 drain — closed #2033 (dj-virtual client state had no attribute/identity-driven teardown; SPA nav reusing a container merged stale cross-view content).

### Quality: 5
### What went well
- **Solved the WeakMap-not-iterable constraint cleanly.** The state map can't be scanned for "containers that lost `dj-virtual`", so the fix reuses the never-removed `data-dj-virtual-shell` marker (the exact node the issue's DOM dump pointed at) as the discovery anchor — `reapStaleVirtualLists()` scans `[data-dj-virtual-shell]` → `parentNode` → teardown-on-identity-change. No new attribute added.
- **`structureIntact()` fails closed** on identity change, so the same-node reuse fast-path can't treat a repurposed container as still-virtualized — one predicate (`identityChanged`) gates every path (init, refresh, absorb).
- **Regression guard is explicit**: a same-thread re-render still self-heals via the #1988/#1989 `absorbLooseChildren` path, proving the fix is scoped to the identity-CHANGE case.
- Found by live-verifying the *previous* drain's fix on the freshly-cut rc6 — the release itself was the test harness.

### What didn't
- New-in-rc6 interaction: the #1988/#1989 auto-absorb self-heal (a real improvement) had no signal to distinguish "same list, more items" from "entirely different view reusing the node" — the fix supplies that signal (`identityChanged`).

### Verified
- 3 new cases + 24 existing virtual-list tests = 27/27; full JS suite 1781 pass; eslint clean; +212 B gzipped. Independently re-run green in the main checkout.
- **Gate-off (#1468):** forcing `identityChanged()` false reds the two identity-change tests, same-thread guard stays green.

RETRO_COMPLETE

PR #2038, opened 2026-07-04
fix(template-tags): stop double-resolving Rust-resolved custom-tag arg values (#2037)

### LINKED ISSUE #2037: {% djust_markdown %} corrupts a value resolved from a per-loop/per-include-scoped dict field — output is a Python-repr-wrapped duplicate of the source text
## Symptom (found live in a downstream production chat app, 1.1.0rc6)

Every finalized assistant reply in a chat UI showed a duplicated, corrupted copy of its own text: a
`.prose` div — genuinely djust-templated, carrying a real `dj-id` attribute — contained literal text
`"('## Streaming demo\n\nHere is a **streaming** reply...',)"` (a Python-tuple-repr-looking string),
immediately followed by a second, correctly-formatted copy of the SAME content.

Reproduced on the very first message of a brand-new conversation (not just on regenerate), in both a
mock provider and a real LiteLLM/OpenAI completion, in a completely fresh browser tab with no prior
session history. Confirmed NOT a data bug: `MessageBlock.content` in the database was a single,
clean string in every case; the Python function building the template context (`_block_dict()`,
equivalent to any per-row dict-builder) was independently verified correct via direct invocation.

## Root cause (grep-verified against the 1.1.0-rc.6 source; not fully pinned to a single line)

The corruption is specific to template rendering, and specific to the value being resolved from a
**per-`{% for %}`-loop, per-`{% include ... with %}`-scoped dict field**, not a top-level context
variable:

```django
{% for b in m.blocks %}
  {% include "chat/partials/block.html" with block=b %}
{% endfor %}
```

```django
{# block.html #}
<div class="prose">{% djust_markdown block.text %}</div>
```

Confirmed this reproduces identically whether `block.text` is passed directly or via a `{% with
txt=block.text %}...{% endwith %}` indirection — the `{% with %}` tag is not the trigger.

Traced partway into the Rust source:

- `crates/djust_templates/src/renderer.rs::resolve_tag_arg` (~line 210) does
  `context.get(arg_trimmed).to_string()` for a bare-name argument.
- `djust_core::Value`'s `Display` impl (`crates/djust_core/src/lib.rs:175`) is a closed match over
  `Null/Bool/Integer/Float/String/List/Object` — `Value::List` renders as literally `"[List]"` and
  `Value::Object` (no `__str__` key) as `"[Object]"`. **Neither variant's `Display` output resembles
  the observed Python-tuple-repr text**, which rules out `resolve_tag_arg`'s own `.to_string()` call
  as the direct source — the corruption must be introduced somewhere in the Rust↔Python custom-tag
  argument bridge specifically (`python/djust/template_tags/markdown.py`'s `MarkdownTagHandler.render`
  calling `self._resolve_arg(args[0], context)`), not in the ordinary `Value`-based `{{ variable }}`
  interpolation path (which renders `block.text` correctly elsewhere in the same app when NOT passed
  through `{% djust_markdown %}`).
- `TagHandler._resolve_arg` (`python/djust/template_tags/__init__.py:89`) documents: "For `{% url
  'post' post.slug %}`, args would be `["'post'", "my-slug"]` (**second arg already resolved by Rust
  if it was a variable**)." This note is the strongest lead: if the custom-tag Python bridge
  pre-resolves a per-loop/per-include variable differently than a top-level one (e.g., handing Python
  an already-resolved-but-differently-typed value for the loop-scoped case), `MarkdownTagHandler`
  calling `_resolve_arg` a second time on an already-resolved value would double-process it — but a
  direct unit test confirms `_resolve_arg` raises `AttributeError` on a raw tuple (`'tuple' object has
  no attribute 'strip'`) rather than silently succeeding, so the actual mismatch is more specific than
  "gets a bare tuple" and needs a maintainer with visibility into the PyO3 bridge for the custom-tag
  args list to pin exactly.

## Scope

1. Fix `{% djust_markdown %}`'s (and likely any custom tag's) argument resolution for per-loop/
   per-include-scoped dict field values — this is a very common Django idiom
   (`{% for x in list %}{% include "partial.html" with item=x %}{% endfor %}`) and should not corrupt
   silently.
2. Add a regression test exercising exactly this shape: a custom tag reading `scoped_var.field` where
   `scoped_var` is bound via `{% include ... with %}` inside a `{% for %}` loop — this downstream
   project's own equivalent test (`LiveViewTestClient(...).render(engine="rust")` against ORM-created,
   never-streamed data) did NOT reproduce the bug, suggesting the trigger may involve djust's own
   per-node/loop HTML render cache (`crates/djust_templates/src/loop_cache.rs`) state accumulated
   across a real `is_streaming: True -> False` transition over an actual WebSocket connection, not a
   single fresh render — worth checking loop-cache invalidation specifically against this shape.
3. Until fixed, `{% djust_markdown %}`'s own docstring should warn against passing a per-loop/
   per-include-scoped value directly, and recommend the workaround: render markdown in Python via
   `djust.render_markdown()` and pass the trusted HTML string through `{{ value|safe }}` instead.

## Refs

Found and fixed downstream in `djust-chat` (a production LLM chat client) — DJUST_LESSONS gotcha #24.
Same investigation session as djust-org/djust#2033 (same app, same day, unrelated mechanism).

johnrtipton:
## Summary

Fixes #2037 — `{% djust_markdown %}` (and, by shared code, any custom tag) corrupted a per-`{% for %}`/`{% include ... with %}`-scoped **plain-dict field** value into a Python-tuple-repr in production (djust.org chat app, 1.1.0rc6).

## Root cause (reproduced deterministically — no streaming/loop-cache needed)

The Rust custom-tag dispatch (`crates/djust_templates/src/renderer.rs` `Node::CustomTag`, ~line 1272) **pre-resolves** a bare-name argument (`block.text`) to its **value** string and passes that value to the Python handler. `TagHandler._resolve_arg` (`python/djust/template_tags/__init__.py`) then re-interpreted the already-resolved value **as if it were a template token**:

- a value containing `=` hit the `key=value` branch → tuple-split → `{% djust_markdown %}`'s `src = str(src_raw)` produced the observed `('## Streaming demo\n\n...x', 'y matters.')` repr;
- a dotted value could be re-resolved against the context.

Only loop/include-scoped plain **dicts** triggered it — that's the case where Rust *succeeds* at resolving the arg (JSON-serializable dict is in the Rust context) and hands Python the value. A top-level model object Rust can't resolve → passes the literal token → Python resolves it correctly (which is why the reporter saw it only on scoped dict fields).

The report's streaming / loop-cache / WebSocket hypothesis was a **red herring** (bug-triage canon: trust the symptom, not the cited path).

## Fix (Python-only — Rust is correct)

Token-guard `_resolve_arg`: a template-tag arg is a whitespace-delimited token, so the kwarg-split and dotted-lookup heuristics fire **only** for genuine token shapes (`^[A-Za-z_]\w*=` and `^[A-Za-z_]\w*(\.\w+)*$`). A non-token value (whitespace / markup / leading non-identifier) is a value the Rust dispatch already resolved → returned **verbatim**. Defence-in-depth in `markdown.py`: a positional source that still resolves to a tuple falls back to the raw arg string.

## Testing

| Level | Test class | What it pins |
|---|---|---|
| unit | `TestResolveArgNoDoubleResolution` | resolved value with `=` returned verbatim; genuine `tables=False` + `title=My Post` kwargs still split; `block.text` still resolves; literals/ints/bare-word unchanged |
| handler | `TestMarkdownHandlerResolvedSource` | `MarkdownTagHandler.render([resolved_value_with_=])` → markdown, no tuple-repr |
| integration (real Rust path) | `TestRealPathLoopScopedMarkdown` | the reporter's exact `{% for %}` + scoped `block.text` shape via `LiveViewTestClient.render_with_patches()` |

- **10/10 new tests green**; **gate-off (#1468)**: stashing the fix turns 3 behavior-meaningful tests RED (non-tautological).
- No regression: **44 passed** across all `_resolve_arg` consumers + **91 passed / 1 skipped** across tag-handler + registry suites (`tests/unit`, `tests/benchmarks`); full pre-push suite green.

Two-commit shape: impl+tests (`232fda71`), docs/CHANGELOG (`8bf43bf3`).


johnrtipton:
## Adversarial code review — APPROVE-WITH-NITS → addressed

An independent read-only review (via `gh pr diff`) checked four risk areas:

1. **Regex correctness / caller regression** — CLEAR. Traced every token shape (`tables=False`, `title=My Post`, `block.text`, empty, bare-word-context-key, unicode) and all five other `_resolve_arg` consumers (url/static/live_render/flash/pwa). No caller regression: the only args OLD would resolve but NEW returns verbatim require context keys with hyphens/spaces (not valid Django variable names). Strictly reduces the corruption surface.
2. **XSS / escaping** — CLEAR. The `render_markdown` sink escaping is unchanged; the fix feeds it the *correct* source instead of a mangled tuple-repr. No new attacker-influenced content reaches an unescaped sink.
3. **Parallel-path completeness (#1646)** — CLEAR. `_resolve_arg` is the single shared chokepoint for Rust→Python custom-tag arg re-resolution; no latent twin (Django-native tags use `Variable.resolve()` and never double-resolve).
4. **Test quality** — one nit: the markdown `isinstance(src_raw, tuple)` defence-in-depth branch (the safety net for a resolved `identifier=value`-no-whitespace value) had **zero coverage**.

**Nit addressed** in `76aa9047`: added `test_source_shaped_like_kwarg_uses_tuple_fallback` (`y=mx+b is the line equation`), gate-off verified — neutering the fallback makes it RED (`<p>('y', 'mx+b ...')</p>`). Full file now 11/11 green.

johnrtipton:
## Retrospective — PR #2038 (pipeline-drain)

**Task**: v1.1.0-8 — drained #2037 (`{% djust_markdown %}` / custom-tag arg double-resolution corrupting a loop/include-scoped dict-field value into a Python-tuple-repr).

### Quality: 5/5
### What went well
- **Symptom-up triage beat the reporter's own hypothesis.** The report speculated a streaming / loop-cache / WebSocket cause (`is_streaming: True→False` transition). Tracing the tuple-repr backwards through the Rust→Python custom-tag bridge found the real cause — `_resolve_arg` re-interpreting a Rust-*resolved* value as a template token — and it reproduced deterministically with a single synchronous render. The streaming hypothesis was a red herring (bug-triage canon: trust the symptom, not the cited path).
- **Python-only fix at the shared chokepoint (#1646).** Rust was correct; the fix token-guards the single `_resolve_arg` helper all six custom tags share — no per-tag patching, no latent twin.
- **Reproduction fidelity**: pinned at unit, handler, AND real-Rust-render (`LiveViewTestClient.render_with_patches()`) levels — the integration test is the reporter's exact `{% for %}` + scoped `block.text` shape.
- **Adversarial review earned its keep**: found the markdown tuple-fallback branch had zero coverage; added a gate-off-verified test for it (`76aa9047`).

### What didn't
- A gate-off attempt used `git stash push -- markdown.py` on an *already-committed* change — which saves nothing, so the follow-up `git stash pop` grabbed an ancient unrelated deep-list stash and polluted 4 files (a known hazard). Recovered cleanly (only the test file was legitimately dirty; the old stash was kept, not lost). **Lesson: gate off a committed change with a direct edit + `git checkout HEAD -- <path>`, never `stash push -- <path>`.**

### Verified
- 11/11 new tests green; gate-off (#1468) turns 4 behavior-meaningful tests RED (3 core + 1 fallback-branch). No regression across all `_resolve_arg` consumers + tag-handler/registry suites. Full CI green (incl. the now-blocking `python/djust/tests/` gate from #2039).

RETRO_COMPLETE

Issue #2041, opened 2026-07-09
regroup: pass by/<attr>/as operands unresolved to assign-tag handlers (durable shadow-fix)


szto:
Follow-up from PR #2023 review (finding #3).

### Problem
The Rust engine resolves **every** assign-tag arg via `resolve_tag_arg` (`crates/djust_templates/src/renderer.rs`) before calling the Python handler. For `{% regroup <expr> by <attr> as <var> %}` this means the `<attr>` operand (e.g. `country`) is resolved against the context. djust auto-exposes public view attributes to the template context, so a context key whose name matches the attribute (`country`, `category`, `type`, …) **shadows** the per-item lookup: `args[2]` arrives as that key's *value* instead of the literal attribute name, and the grouping is silently wrong. Django never resolves the attr against the outer context.

### Current mitigation (PR #2023)
`RegroupTagHandler.render` emits a `logger.warning` when the resolved `<attr>` isn't a bare (dotted) identifier — the strongest signal available from the post-resolution vantage point. This catches shadow-to-a-non-identifier-value but not shadow-to-an-identifier-value.

### Durable fix
Pass the keyword/name operands (`by`, `<attr>`, `as`, `<var>`) to assign-tag handlers **unresolved**, matching Django's semantics. This is a deeper change to the assign-tag arg path (`renderer.rs` `resolve_tag_arg` callers + the `AssignTagHandler` contract) — likely a per-handler operand-mask or a positional "don't resolve" marker — so it's scoped as a follow-up rather than bundled into the regroup feature.

Referenced from `python/djust/template_tags/regroup.py`, `template_tags/__init__.py`, and the CHANGELOG `[Unreleased]` entry.

PR #2043, opened 2026-07-10
refactor(templates): shared value_to_arg_string across all 3 tag-dispatch paths (#2042)

### LINKED ISSUE #2042: Template engine: route CustomTag / AssignTag / BlockCustomTag arg resolution through one shared value_to_arg_string helper
Follow-up from PR #2023 review (finding #4) — parallel-path-drift canon (#1646).

### Problem
Three tag-dispatch branches in `crates/djust_templates/src/renderer.rs` resolve args, but only two are JSON-aware:

- `Node::CustomTag` — has an inline `value_to_arg_string` (`renderer.rs:~1252`) that JSON-encodes list/object args.
- `Node::AssignTag` — PR #2023 added `resolve_tag_arg` (`renderer.rs:217`) with the same JSON-aware `value_to_arg_string`, bringing it to parity.
- `Node::BlockCustomTag` — **still** uses an inline `resolved.to_string()` / `format!("{}={}", key, resolved)` (`renderer.rs:~1178-1186`), so a list/object arg collapses to the opaque `[List]` / `[Object]` placeholder and loses the payload.

This is pre-existing and not a regroup regression (regroup is an assign tag), but it's the un-updated sibling of the class PR #2023 fixed on the other two paths.

### Fix
Extract one shared `value_to_arg_string` (and ideally a shared arg-resolution helper) and route all three branches (`CustomTag`, `AssignTag`, `BlockCustomTag`) through it, so the `[List]`/`[Object]`-collapse class is retired everywhere rather than fixed 2-of-3.

### Test
Add a `BlockCustomTag` case that passes a list/object arg and asserts the handler receives JSON, not `[List]`.

johnrtipton:
## What & why

`crates/djust_templates/src/renderer.rs` had **three** tag-dispatch arg-resolution branches, but only two JSON-encoded structured (list/object) args:

- `Node::CustomTag` — inline `value_to_arg_string` (JSON-encodes list/object).
- `Node::AssignTag` — `resolve_tag_arg` (added in #2023) with the same JSON-aware encoding.
- `Node::BlockCustomTag` — a **hand-copied inline resolver** that skipped the JSON step, so a list/object arg collapsed to the opaque `[List]` / `[Object]` placeholder and the Python block handler lost the payload.

This is a follow-up filed from the PR #2023 (`{% regroup %}`) code review.

Closes #2042.

## The #1646 parallel-path cure

Retire the collapse **class**, don't fix 2-of-3:

- Hoisted `value_to_arg_string` to **one** module-level helper (it was duplicated inline in `resolve_tag_arg` and in the `CustomTag` arm) — single source of truth for arg encoding across all three branches.
- Folded `BlockCustomTag` onto the same `resolve_tag_arg` helper `AssignTag` already uses (its inline resolver was a non-JSON twin — structurally identical *except* the missing encode). Two plain-context-lookup resolvers → one.
- `CustomTag` keeps its filter-aware `get_value` resolver (e.g. `x|upper`, `Result` semantics) but now shares the same `value_to_arg_string` encoding.

**Behavior:** byte-for-byte preserved for `CustomTag`/`AssignTag` (their existing Python-side tests stay green); the only behavior **change** is `BlockCustomTag` now JSON-encodes list/object args instead of collapsing them. Scalars unchanged everywhere.

## Shared helper

```rust
fn value_to_arg_string(v: &Value) -> String {
    match v {
        Value::List(_) | Value::Object(_) => serde_json::to_string(v).unwrap_or_else(|_| v.to_string()),
        _ => v.to_string(),
    }
}
```

Three call sites now route through it: `Node::CustomTag` (directly), `Node::AssignTag` and `Node::BlockCustomTag` (both via `resolve_tag_arg`, which calls it).

## Files × tests

| File | Change | Covering tests |
|---|---|---|
| `crates/djust_templates/src/renderer.rs` | hoist shared `value_to_arg_string`; strip 2 inline copies; fold `BlockCustomTag` onto `resolve_tag_arg` | 5 Python-free lib unit tests (`value_to_arg_string_*`, `resolve_tag_arg_*`) pinning the shared-helper contract |
| `crates/djust_templates/tests/test_block_custom_tag_arg_json_2042.rs` (new) | end-to-end `BlockCustomTag` dispatch via a registered Python block handler | `block_custom_tag_arg_resolution_json_encodes_structured_values` (list→JSON, object→JSON, scalar unchanged, `key=value` list→JSON, unknown keyword operand kept literal, quoted literal kept) |
| `CHANGELOG.md` / `ROADMAP.md` | `### Fixed` entry; open v1.1.0-9 bucket | — |

> The integration file is a **single** `#[test]` on purpose: cargo's default harness runs `#[test]`s on parallel threads, and concurrent `Python::attach` to the embedded CPython interpreter deadlocks. Keep all cases in one test.

## Verification

- `cargo test -p djust_templates` — **all green** (321 lib incl. 5 new unit tests; 1 new integration test; 0 failures across all binaries).
- `cargo clippy -p djust_templates --all-targets -- -D warnings` — **clean** (exit 0).
- `cargo fmt -p djust_templates -- --check` — clean.
- **Gate-off (#1468):** reverting ONLY the `BlockCustomTag` routing (back to the pre-fix non-JSON inline resolver) turns the integration test **red** — `assertion failed ... got "[List]"` (left) vs `"[1,2,3]"` (right) — proving the test is load-bearing and non-tautological. Restored to green.
- **Regression scan:** no other un-updated sibling. The `Value` `Display` impl in `djust_core` still emits `[List]`/`[Object]` — but that is the intentional `{{ var }}` text-render fallback (Django parity), not a tag-arg-resolution path.


johnrtipton:
## Retrospective — PR #2043 (pipeline-drain, v1.1.0-9)

**Task**: v1.1.0-9 drain — #2042 (route CustomTag/AssignTag/BlockCustomTag arg resolution through one shared `value_to_arg_string`; the #1646 parallel-path cure).

### Quality: 5/5
### What went well
- Structural cure, not a point fix: BlockCustomTag's hand-copied inline resolver was deleted and folded onto the same `resolve_tag_arg` AssignTag uses, so the `[List]`/`[Object]`-collapse class is retired everywhere (3-of-3), not patched 2-of-3.
- Real end-to-end coverage: the integration test drives the actual `BlockCustomTag` → registered-Python-handler dispatch and asserts on the exact string the handler receives, plus 5 Python-free unit tests pin the shared helpers. Gate-off proven (reverting only the BlockCustomTag routing turns the test red with `got "[List]"`).
- CustomTag/AssignTag behavior byte-preserved (filter-aware `get_value` kept on CustomTag; scalars unchanged everywhere).

### What didn't (minor)
- First test draft (5 `#[test]` fns) hung >60s — parallel `Python::attach` across cargo's default harness deadlocks. Consolidated to a single sequential `#[test]`; documented in the file header so a future contributor doesn't re-split it.

### Verified
- `cargo test -p djust_templates`: 321 lib (incl. 5 new) + 1 new integration, 0 failures. `cargo clippy --all-targets -D warnings` + `cargo fmt --check` clean.
- CI green on the PR head: 13/13 real checks (rust-tests, python-tests 3.12 + 3.14t, security-tests, playwright, browser-smoke, demo-checks, analyze).

Sibling #2041 (durable regroup operand-shadow fix) drains next on top of this. `RETRO_COMPLETE`

PR #2044, opened 2026-07-10
fix(templates): pass assign-tag operands unresolved — durable regroup shadow fix (#2041)

### LINKED ISSUE #2041: regroup: pass by/<attr>/as operands unresolved to assign-tag handlers (durable shadow-fix)

johnrtipton:
## What & why

`{% regroup <src> by <attr> as <var> %}` grouped by the **wrong attribute**
whenever a context key was named after the `<attr>` token. The Rust engine
resolved **every** assign-tag arg against the render context before calling
the Python handler, so the `<attr>` operand (`args[2]`) was resolved too: a
top-level context variable named like the attribute (`country`, `type`,
`category`, … — djust auto-exposes public view attrs to the template context)
shadowed the per-item lookup. `<attr>` arrived as that key's *value* instead of
the literal attribute name, and the grouping was **silently wrong** (every row
collapsing into one bogus group).

Django never resolves `<attr>` against the outer context. The PR #2023
mitigation only `logger.warning`-ed when the resolved attr wasn't a bare
identifier — it **missed** the shadow-to-an-identifier-value case
(`country → "usa"`). This is the durable fix.

Closes #2041.

## Operand-resolution mechanism (chosen shape)

`AssignTagHandler` gains a `RESOLVE_ARG_POSITIONS: set[int] | None` class
attribute:

- `None` (default) — resolve **every** arg against the context (the unchanged
  historical behavior; any future assign tag that doesn't opt in is unaffected).
- a `set[int]` of 0-based positions — resolve **only** those positions; every
  other arg is passed through as a **literal token**.

`register_assign_tag_handler` reads the attribute **once at registration**
(stored on a new `AssignHandlerEntry` alongside the handler). One shared
`resolve_assign_tag_args` helper — routed through **all four** `AssignTag`
dispatch sites in `renderer.rs` (`render_nodes_with_loader`,
`render_nodes_collecting`, `render_nodes_partial`, and the individual
`render_node_with_loader` arm) — resolves only the declared positions. That
single entry point is the #1646 parallel-path cure: the operand-mask can't
drift between the four sites.

**`RegroupTagHandler` declares `RESOLVE_ARG_POSITIONS = {0}`** — resolve only
the `<expr>` source (still JSON-encoded as before, so the source-decode path is
unchanged); `by` / `<attr>` / `as` / `<var>` stay literal. Shadowing is now
impossible, and dotted `by author.team` still resolves as a per-item path (it
was never a context resolution — it's traversed against each item).

Removed the now-obsolete `logger.warning` + `_IDENTIFIER_RE` shadow heuristic
from `regroup.py` (the root cause is gone), and updated the `regroup.py` module
docstring + the `AssignTagHandler` docstring to describe the new contract.

## Files × tests

| File | Change | Covered by |
|---|---|---|
| `crates/djust_templates/src/registry.rs` | `AssignHandlerEntry` struct; read `RESOLVE_ARG_POSITIONS` at registration; `assign_handler_resolve_positions` accessor | `cargo test -p djust_templates` (66 unit + integration) |
| `crates/djust_templates/src/renderer.rs` | shared `resolve_assign_tag_args` helper through all 4 AssignTag dispatch sites | Rust unit tests + Python e2e |
| `python/djust/template_tags/__init__.py` | `AssignTagHandler.RESOLVE_ARG_POSITIONS` + contract docstring | `test_regroup_handler_declares_resolve_only_source` |
| `python/djust/template_tags/regroup.py` | declare `{0}`; remove warning/regex; docstrings | full `test_regroup_tag.py` |
| `python/djust/tests/test_regroup_tag.py` | +2 e2e shadow reproducers, +1 declaration pin; −2 obsolete warning tests | this PR |

## Verification

- **RED (reproducer-first):** the two shadow reproducers (`{% regroup cities by
  country as g %}` with a top-level `country="usa"` + per-item `country`; and a
  dotted `author.team` with a shadowing `author` key) FAIL on the pre-fix build
  (single bogus group `(A B C )` / `[Mumbai,Calcutta,New York,Chicago,]`).
- **GREEN:** full `test_regroup_tag.py` — 13 passed (basic, non-contiguous,
  dotted `author.team`, `render_template` + `RustLiveView.render_with_diff`
  paths, Django-parity anchor, both shadow reproducers, the declaration pin).
- **Gate-off (#1468):** forcing the mask off (`resolve_positions = None` inside
  `resolve_assign_tag_args`) turns **both** shadow reproducers RED —
  `test_regroup_attr_operand_is_not_shadowed_by_context_key` and
  `test_regroup_dotted_attr_operand_is_not_shadowed_by_context_key`. Restored.
- `cargo fmt --check` clean; `cargo clippy -p djust_templates --all-targets -D
  warnings` clean; `cargo test -p djust_templates` 66 passed.
- 104 passed across the regroup + custom/block/markdown/theme-tag Python suites
  (no regression from the assign-registry change).

Built + tested in an isolated throwaway venv inside the worktree (Python 3.12,
`maturin develop`); the main checkout `.venv` was not touched.

## Notes

- The #2023 `### Added` CHANGELOG note was updated to point at this durable fix
  instead of the removed warning mitigation.
- `register_assign` is used exactly once (regroup), so blast radius is small;
  the default (`None` = resolve all) preserves behavior for any future assign
  tag.


johnrtipton:
## Retrospective — PR #2044 (pipeline-drain, v1.1.0-9)

**Task**: v1.1.0-9 drain — #2041 (pass assign-tag `by`/`<attr>`/`as`/`<var>` operands unresolved so a context key can't shadow regroup's `<attr>` — the durable fix for the footgun #2023 only warned about).

### Quality: 5/5
### What went well
- Root-cause fix, not a bigger warning: the #2023 `logger.warning` + `_IDENTIFIER_RE` heuristic (which structurally couldn't catch a shadow-to-an-identifier value like `country → "usa"`) is deleted; the operand-shadow class is now impossible.
- Clean, opt-in contract: `AssignTagHandler.RESOLVE_ARG_POSITIONS` (`None` = resolve-all default, unchanged for every future assign tag; regroup declares `{0}`). Read once at registration.
- Bonus #1646 cure: discovered FOUR independent `AssignTag` arg-resolution copies in `renderer.rs` (`render_nodes_with_loader`, `render_nodes_collecting`, `render_nodes_partial`, `render_node_with_loader`) and routed all four through one `resolve_assign_tag_args` helper, so the operand-mask can't drift between them.
- End-to-end reproducer (`country="usa"` shadow) + a dotted-`author.team` guard, both via real `render_template`; gate-off proven (forcing `resolve_positions=None` turns both red). Removed the two obsolete warning tests; symbol-migration grep (#1391) confirms no orphan refs to the removed regex/logger.

### What didn't (minor)
- The isolated worktree venv first came up Python 3.14 while pyo3 0.29 targets 3.12; recreated with `python3.12`. Env-setup friction only, no code impact.

### Verified
- `cargo test -p djust_templates` (66) + `clippy -D warnings` + `fmt` clean; 104 Python tests across regroup + custom/block/markdown/theme suites (isolated `.wt-venv`).
- CI green on the PR head: 13/13 real checks. Built on #2043 (the shared `value_to_arg_string`), so both v1.1.0-9 issues share one coherent renderer-arg-resolution story.

`RETRO_COMPLETE`

PR #2065, opened 2026-07-14
feat(dx): make doctor — dev-environment self-diagnosis (#2061)

### LINKED ISSUE #2061: feat(dx): make doctor — dev-environment self-diagnosis
Local environment drift silently breaks trust in local gates. One machine, one week: djust.pth repointed at a deleted worktree (import breakage + stale .so), embedded-PyO3 test binaries failing init_fs_encoding bootstrap, then pyo3-ffi failing to compile (SIGABRT in the build probe) — CI green throughout. Pre-push hook also hardcodes .venv/bin/python (#1796).

**Proposal**: `make doctor` target checking + reporting with one-line fixes: djust.pth points into THIS checkout; `import djust` works from the venv; .so mtime vs newest crates/ source (stale-extension detection); one embedded-PyO3 cargo smoke test; node_modules present; hook health. Each failure prints its remedy (usually `uv run maturin develop`).

From docs/internal/IMPROVEMENT_BRAINSTORM-2026-07.md §5.2 (shortlist #6).

johnrtipton:
## Summary

- Adds `make doctor` (`scripts/doctor.sh`) — a dev-environment self-diagnosis target that runs seven defensive checks and reports `[OK]`/`[WARN]`/`[FAIL] <name>: <finding>` plus a one-line remedy, exiting non-zero iff any check FAILs.
- Motivated by three local-env incidents in one week while CI stayed green: `djust.pth` repointed at a deleted worktree (import breakage + stale compiled `.so`), embedded-PyO3 test binaries failing an `init_fs_encoding`/`'/install'`-prefix bootstrap, and pyo3-ffi failing to compile (SIGABRT in the `rustc -vV` build probe). Plus #1796 (pre-push hook hardcoding `.venv/bin/python`, fails in a worktree — already fixed via `scripts/run-with-venv-python.sh`; this doctor verifies that fix stays wired in).

Closes #2061.

## Checks

| Check | What it verifies | Remedy on failure |
|---|---|---|
| `venv` | `.venv/bin/python` resolves + `import djust` succeeds + `djust.__file__` lands inside this checkout | `uv run maturin develop` |
| `djust.pth` | site-packages editable-install pointer targets this checkout's `python/`, not a (possibly deleted) `worktrees/` path | `uv run maturin develop` |
| `stale-extension` | compiled `_rust.cpython-*.so` mtime vs. newest `crates/**/*.rs` | `uv run maturin develop` |
| `embedded-pyo3` | 30s-capped smoke run of one existing small PyO3-embedding test under `crates/djust_templates/tests/`, classifying `init_fs_encoding`/SIGABRT failures by name | environment-specific (named in the FAIL line) |
| `node` | `node_modules/` present + `npx vitest --version` works | `npm install` |
| `hooks` | `pre-commit`/`pre-push` present + executable; notes the #1796 fix is still wired in when run from a worktree | `make pre-commit-install` |
| `git-config` | `core.bare` sanity (the #1804 incident class), reusing `scripts/check-shared-git-config.sh` | `git config core.bare false` |

A missing tool (cargo, node, npx) is always a WARN with a remedy, never a crash — the script is `set -u`, deliberately **not** `set -e`, so every check runs regardless of earlier failures.

## Sample output

Run live in this PR's own dev environment (a worktree sharing the main checkout's venv — the `venv`/`djust.pth` WARNs below are the *expected*, documented behavior in that topology, not bugs):

```
$ make doctor
djust doctor — dev-environment self-diagnosis (issue #2061)
repo root: /Users/tip/…/djust/.claude/worktrees/agent-ae954c2011edae030

[WARN] venv: 'import djust' resolves OUTSIDE this checkout: /Users/tip/…/djust/python/djust/__init__.py (expected under …/worktrees/agent-…/python/djust — normal when running doctor from a worktree whose venv is shared with the main checkout)
      remedy: run 'uv run maturin develop' from the repo root to repoint the editable install here
[WARN] djust.pth: …/.venv/lib/python3.12/site-packages/djust.pth points outside this checkout: /Users/tip/…/djust/python (expected …/worktrees/agent-…/python — normal when this checkout is a worktree sharing the main checkout's venv)
      remedy: run 'uv run maturin develop' from the repo root
[WARN] stale-extension: _rust.cpython-312-darwin.so is older than 100 crates/**/*.rs file(s), e.g. crates/djust_vdom/tests/test_dj_if_loop_spurious_move_1826.rs
      remedy: run 'uv run maturin develop' to rebuild the extension
[FAIL] embedded-pyo3: cargo test -p djust_templates --test test_block_custom_tag_arg_json_2042: rustc build probe crashed (SIGABRT in the rustc -vV probe): error: process didn't exit successfully: `rustc -vV` (signal: 6, SIGABRT: process abort signal)
      remedy: rustc/LLVM toolchain drift — try a clean shell (unset stray DYLD_*/LLVM env vars) or reinstall the rust toolchain; see issue #2061
[WARN] node: node_modules/ not found
      remedy: run 'npm install'
[OK] hooks: pre-commit/pre-push present + executable; running from a worktree — #1796 fix (scripts/run-with-venv-python.sh) is wired into .pre-commit-config.yaml
[OK] git-config: core.bare is false/unset in the shared git config

djust doctor summary: 2 ok, 4 warn, 1 fail
```

That `embedded-pyo3` FAIL is a **real, live** homebrew rustc/LLVM mismatch on this dev machine right now (`librustc_driver` looking for an LLVM symbol not present in the installed `libLLVM.dylib`) — i.e. the doctor script empirically caught incident class #3 from the issue while writing this PR, unprompted.

## Test hooks (non-tautology, #1468/#1200)

Two env-var overrides let the test suite force a deterministic FAIL without breaking the real environment:

- `DOCTOR_FAKE_STALE_SO=1` forces the `stale-extension` check to FAIL.
- `DOCTOR_FAKE_PTH_BAD=1` forces the `djust.pth` check to FAIL.

Each has a **gate-off sibling** asserting the forced-failure marker does *not* appear when the override is unset (proving the assertion isn't tautological), and I additionally verified the gate-off self-test manually: neutering each `if [ -n "${DOCTOR_FAKE_*}" ]` guard (`if false; then`) makes the corresponding `test_*_force_fail_is_detected` test fail, confirming the tests exercise the real code path:

```
$ pytest python/djust/tests/test_make_doctor_2061.py -q -k stale   # with the guard neutered
FAILED python/djust/tests/test_make_doctor_2061.py::test_stale_so_force_fail_is_detected

$ pytest python/djust/tests/test_make_doctor_2061.py -q -k pth_force   # with the guard neutered
FAILED python/djust/tests/test_make_doctor_2061.py::test_pth_force_fail_is_detected
```

(guards restored before commit — `git diff` confirms no residual changes)

## Test plan

- [x] `python/djust/tests/test_make_doctor_2061.py` — 9 new regression cases: script exists+executable, `set -u`/no-`set -e` source pin, `Makefile` `doctor:` target, full-run structural test (every check name present, exit code ∈ {0,1}, summary counts add up), 2×(gate-off + forced-FAIL) for the two test hooks, check-name wiring pin.
- [x] `pytest python/djust/tests/test_make_doctor_2061.py -q` → **9 passed** (via the main checkout's `.venv`, per the worktree-subagent convention).
- [x] `ruff check` / `ruff format --check` on the new test file — clean.
- [x] `shellcheck scripts/doctor.sh` — 3 SC2012 info-level notices only (`ls` vs `find` for globbing controlled, simple filenames), no warnings/errors.
- [x] `python scripts/check-changelog-test-counts.py` — clean (9 claimed == 9 actual `def test_` in the new file).
- [x] Manual `make doctor` run in this worktree — see sample output above.
- [x] Manual runs with `DOCTOR_FAKE_STALE_SO=1` / `DOCTOR_FAKE_PTH_BAD=1` — both force the expected `[FAIL]` + remedy line, other checks still run, exit code 1.

## Notes for reviewers

- `embedded-pyo3` reuses an **existing** test file (`crates/djust_templates/tests/test_block_custom_tag_arg_json_2042.rs`, 133 lines, the smallest of the two files under `crates/djust_templates/tests/` that actually embed CPython via `Python::initialize`/`pyo3::prelude`) rather than adding a new one — selection is dynamic (`grep -l` + smallest-by-line-count), not hardcoded, so it stays correct if that file is renamed/removed and a different small PyO3 test appears.
- `git-config` sanity delegates to the already-shipped, already-tested `scripts/check-shared-git-config.sh` (DRY) rather than re-implementing the `core.bare` read.
- `venv`/`djust.pth` classify "resolves outside this checkout" as **WARN**, and only classify a **dangling `worktrees/` path** (the literal #2061 incident: `.pth` pointing at a *deleted* worktree) as **FAIL** — this matches the issue's own worktree-environment note that a shared-venv WARN is legitimate, not a bug.


johnrtipton:
## Retrospective — PR #2065 (v1.1.0-10 drain, #2061)

**Task**: make doctor — dev-environment self-diagnosis (7 checks, one-line remedies).

### Quality: 5
### What went well
- The deliverable validated itself during its own build: the embedded-pyo3 check caught the LIVE rustc/LLVM homebrew mismatch on the dev machine — the exact undiagnosed failure that muddied the rc7/rc8 release gates. The tool paid for itself before merging.
- Non-tautological forced-failure hooks (DOCTOR_FAKE_*) with gate-off siblings; reused check-shared-git-config.sh instead of reimplementing (#1804 class).
### Watch-item
- The test suite invokes the real script (incl. the 30s-capped cargo probe) ~5×; CI python-tests absorbed it fine this run, but if the job time creeps, add a DOCTOR_SKIP_PYO3 test-env guard keeping one real invocation.
### Verified
9/9 tests; CI all green; worktree-isolated build; two-commit shape; CHANGELOG under [Unreleased] only.

RETRO_COMPLETE

PR #2071, opened 2026-07-14
feat(client): DEBUG-mode warning for non-executed scripts inserted by morph/patch (#2058)

### LINKED ISSUE #2058: feat(client): loud detection of non-executed <script> inserted by the mount morph (durable #1848 follow-up)
#1848 (closed with a documented workaround) remains the framework's sharpest silent trap: page JS inside the dj-root reactive region is inserted by the WS-mount morph but never executed — no error, no warning; 'reload works, navigation doesn't'. It bit djust.org's own examples page during the 1.0.7 upgrade.

**Proposal (step b of the brainstorm — do this now)**: during morph/patch application, when an inserted node is a classic `<script>` that will not execute, emit a dev-mode-only `console.error` (DEBUG_MODE-gated, %s-parameterized) naming the script's src/first-line and pointing at the hooks.md guidance + colocated hooks. Optionally a Django system check scanning project templates for `<script` inside blocks rendered under dj-root, if a reliable heuristic exists.

Full re-execution semantics (step a) stay a separate ADR-sized question (double-execution hazards; `live_redirect` already re-executes per #1635/#1650).

From docs/internal/IMPROVEMENT_BRAINSTORM-2026-07.md §1.2 (shortlist #2).

johnrtipton:
## Summary

Closes #2058.

#1848 fixed the two WS-mount paths so a classic inline `<script>` inside `dj-root` runs after mount (`_runInsertedScripts()`), but that fix's scope was those two paths only. Every OTHER DOM operation that inserts server-authored HTML via `innerHTML`, `morphChildren` (clone+insert), or a `<template>`-parsed fragment is *also* inert-by-spec for `<script>` per the HTML Standard, and silently drops inline JS the same way — no error, "reload works, navigation doesn't."

This PR adds a DEBUG-mode-only detector (not a re-execution fix, per the issue scope) — `window.djust._warnDeadScripts(root)` — and wires it into every insertion path a #1646-style audit found.

## Insertion-path inventory (step 1)

| Path | File | Status before this PR | Action |
|---|---|---|---|
| WS-mount `skipMountHtml` morph (#1610) | `03-websocket.js` (`case 'mount'`) | Fixed by #1848 (`_runInsertedScripts`) | Added `_warnDeadScripts` after, as defense-in-depth (should always find zero) |
| WS-mount non-prerendered `innerHTML` | `03-websocket.js` (`case 'mount'`) | Fixed by #1848 | Same defense-in-depth addition |
| `mount_batch` (lazy-hydration batch) | `03-websocket.js` | **Gap** — no `_runInsertedScripts` call at all | Warning added |
| `html_recovery` (WS reconnect / version-mismatch recovery morph) | `03-websocket.js` | **Gap** | Warning added |
| `handleEmbeddedUpdate` (LiveComponent scoped update) | `03-websocket.js` | **Gap** | Warning added |
| SSE mount (`case 'mount'`) | `03b-sse.js` | **Gap** — SSE transport has systematically lacked WS-side controls (see `project_security_audit_2026_06` memory) | Warning added |
| Streaming `replace`/`append`/`prepend` ops (LLM chat / live feeds) | `17-streaming.js` | **Gap** | Warning added (fragment scanned before insertion for append/prepend) |
| Instant-shell swap | `33-sw-registration.js` | **Gap**, already documented in the module docblock as a known limitation | Warning added, backing the existing doc comment |
| Runtime layout swap (`{"type": "layout", ...}`) | `40-dj-layout.js` | **Gap** | Warning added |
| Lazy-render fills (`{% live_render lazy=True %}`) | `50-lazy-fill.js` | **Gap** | Warning added (clone scanned before `replaceWith`) |
| `InsertSubtree` patch (`{% if %}` toggling a subtree on) | `12-vdom-patch.js` | **Gap** | Warning added (fragment scanned before insertion, since `DocumentFragment` empties on insert) |
| `Replace` / `InsertChild` patches (`createNodeFromVNode`) | `12-vdom-patch.js` | N/A — `<script>` is not in `ALLOWED_HTML_TAGS`, so these paths structurally cannot insert a real `<script>` element (falls back to `<span>`) | No change needed |

Excluded from warning (by design):
- Scripts already marked `data-djust-script-ran` by `_runInsertedScripts()`.
- `type="djust/hook"` colocated-hook payload scripts (extracted by `32-colocated-hooks.js`, never meant to execute as classic JS).
- Any other non-classic `type` (`application/json`, `importmap`, `module`, etc.).

## The helper

`window.djust._warnDeadScripts(root)` (defined in `03-websocket.js`, alongside `_runInsertedScripts`): no-op unless `window.DEBUG_MODE` is true; scans `root` (an `Element` or `DocumentFragment`) for classic `<script>` elements lacking the `data-djust-script-ran` marker and emits one `%s`-parameterized `console.error` per dead script, naming its `src` or a 60-char excerpt of its inline body, pointing at `{% colocated_hook %}` / a block rendered outside `dj-root`.

## Tests + gate-off evidence

`tests/js/dead-script-warning-2058.test.js` — 9 cases, mirroring the `mount-morph-script-exec-1848.test.js` JSDOM harness pattern (real built `client.js`, `window.eval`):
1. helper is exported
2. **(a)** drives the REAL `InsertSubtree` patch dispatcher (`window.djust._applyInsertSubtree`) with a classic inline `<script>`, `DEBUG_MODE=true` → exactly one `console.error`, label names the script
3. **(b)** same call without `DEBUG_MODE` → zero warnings
4. **(c)** `type="djust/hook"` script → zero warnings
5. **(d)** subtree with no scripts → zero warnings
6. scripts already handled by `_runInsertedScripts()` → zero warnings from `_warnDeadScripts` (defense-in-depth non-duplication)
7. external `<script src>` → label is the `src`, not the (empty) body
8. `application/json` script → zero warnings
9. `null`/`undefined` root → no throw, no warning

**Gate-off (#1468), done manually per the issue's instructions**: commented out the `_warnDeadScripts(fragment)` call in `applyInsertSubtree()` (`12-vdom-patch.js`), rebuilt via `bash scripts/build-client.sh`, re-ran the suite — case (a) went RED (`expected "error" to be called 1 times, but got 0 times`), confirming the test is behavior-meaningful and not tautological. Restored the call site and rebuilt before finalizing.

Full JS suite: **160 files / 1821 tests passing.** `npm run lint` clean across `python/djust/static/djust/`.

## Bundle delta

Gzipped `client.min.js.gz`: 58036 → 58388 bytes = **+352 B**, well under the 2 KB budget and the "<400 B expected" guidance for this feature.

## Scope decisions

- Step (b) of the issue (a Django system-check heuristic scanning templates for `<script` inside blocks rendered under `dj-root`) was investigated and **skipped** — no static template-analysis pass reliably distinguishes "this literal `<script` text renders inside the `dj-root` region" without unacceptable false-positive/false-negative rates (template inheritance, conditional blocks, includes, and the Rust template engine's own block resolution all make this a runtime question, not a static one). Scope discipline per CLAUDE.md #1079; noted in the CHANGELOG entry.
- Two secondary insertion paths (`40-dj-layout.js`, `50-lazy-fill.js`) were included even though they're outside the issue's originally-named surface (mount morph + patch ops), because the #1646 audit doctrine ("find EVERY insertion path") surfaced them as genuine, currently-silent instances of the exact same trap, and the cost of covering them was a few lines each.

## Test plan

- [x] `npx vitest run tests/js/dead-script-warning-2058.test.js` — 9/9 passing
- [x] `npx vitest run tests/js/` — 160 files / 1821 tests passing
- [x] `npm run lint` — clean
- [x] Gate-off manually verified (see above)
- [x] Bundle rebuilt via `bash scripts/build-client.sh`, gzip delta measured


johnrtipton:
## Code Review — drain gate (verdict: APPROVE, merged on local validation)

Actions check-suite gap persists (main-push CI is the post-merge gate). Verified first-hand: diff reviewed — one shared `_warnDeadScripts` helper (DEBUG_MODE-gated, %s-parameterized per #1124, correct exclusions for `data-djust-script-ran`/`djust/hook`/non-classic types), wired into all 9 insertion paths from the #1646 audit including three genuine pre-existing gaps (mount_batch, html_recovery, SSE mount) beyond the issue's literal surface — a justified scope call, stated in the PR. System-check half investigated and skipped with reasons (#1079). Independently re-ran on the branch in a scratch worktree against the built bundle: **27/27** (the 9 new cases incl. the real InsertSubtree-dispatch drive, plus the two adjacent new suites). Implementer evidence: full JS suite 1821 passing, +352 B gz (≤2 KB budget), gate-off verified.

Union-merged origin/main locally before merge.

APPROVE.

johnrtipton:
## Retrospective — PR #2071 (v1.1.0-10 drain, #2058)

**Task**: DEBUG-mode loud detection of dead `<script>` insertions (durable #1848 follow-up).

### Quality: 5
### What went well
- The #1646 every-insertion-path audit found THREE genuine pre-existing gaps beyond the issue's named surface (mount_batch, html_recovery, SSE mount — the 'SSE lacks WS controls' pattern again) plus two secondary paths; the trap is now loud everywhere server HTML enters the DOM.
- Honest scope decisions both directions: extra paths included with justification; the system-check half investigated and skipped with stated false-positive reasoning (#1079).
- Structural finding for free: Replace/InsertChild patches CANNOT insert scripts (ALLOWED_HTML_TAGS excludes script) — documented, narrowing the trap surface.
### What didn't
- Actions gap continued — merged on local validation (independent 27/27 scratch re-run against the built bundle) with main-push CI confirming.
- debug-panel .min.js.map byte-drift resurfaced (#2054's toolchain-drift class), noted by the implementer.

### Verified
27/27 independent re-run; +352 B gz within budget; union merge preserved all [Unreleased] entries; gate-off proven.

RETRO_COMPLETE

PR #2051, opened 2026-07-10
feat(js-commands): JS.ext.* user-registered custom commands (ADR-025)

johnrtipton:
Implements Feature B of [ADR-025](https://github.com/djust-org/djust/blob/main/docs/adr/025-js-extension-sockets.md): opens the closed 11-command `COMMAND_TABLE` via a client-side registry + a namespaced dynamic Python factory.

## What
- `window.djust.commands.register(name, fn)` — client registry; throws on built-in collision (incl. camelCase aliases) / dotted names / non-function; re-register = last-wins with debug warn
- `JS.ext.<name>(...)` (Python) + `JSChain.ext` mid-chain — dynamic, chainable, `ext.` wire prefix; built-ins stay strict (`JS.shwo` still `AttributeError`); **snake_case enforced** (`JS.ext.scrollTo` raises — a camelCase name would be permanently unregistrable client-side, caught by the whole-branch review)
- Dispatch in `executeOps`: same `to`/`inner`/`closest` target resolution as built-ins; async implementations awaited in chain order; an op failure doesn't abort the chain
- Unknown op: DEBUG error overlay + did-you-mean (edit distance over the registry); `console.error` (`%s` form, #1124) in prod
- Docs: new `## Custom commands` section in `docs/website/guides/js-commands.md` (anchor cited by the client hint)

## Tests
- `python/djust/tests/test_js_ext_commands.py` — 19 cases incl. wire-format pins (#1448) and snake_case-convention pins
- `tests/js/js_ext_commands.test.js` — 12 cases incl. real-DOM click integration (#1196)
- Gate-off verified on both sides (#1468): dropping the `ext.` prefix / nulling the registry lookup / reverting the snake_case guard each turn the relevant pins red
- Local: Python 4749 passed, JS 1806 passed. (One pre-existing local-env Rust flake in `test_block_custom_tag_arg_json_2042` under the parallel harness — passes standalone on main and branch, zero Rust changes here; CI authoritative.)

## Bundle
client.min.js gzipped delta: **736 B** (budget ≤2 KB). Note: commit 31354b91 also carries debug-panel `.gz`/`.br` recompression churn from local toolchain drift (source `.min.js` byte-identical) — follow-up issue to be filed.

## Review trail
Executed via subagent-driven development: per-task spec+quality reviews (3 tasks, all Approved) + a final whole-branch review that flagged the snake_case/collision asymmetry, fixed in `6d819ed0`, re-review verdict READY.

[...]

PR #2076, opened 2026-07-15
build: stop committing compressed bundle siblings — kill cross-toolchain churn (#2054)

### LINKED ISSUE #2054: build: debug-panel .gz/.br compressed artifacts churn on contributor machines (toolchain drift)
PR #2051 commit 31354b91 carried `debug-panel.min.js.br` (+998 B) / `.gz` (−30 B) / `.map` changes while `debug-panel.min.js` itself was byte-identical and no debug-panel source changed — the pre-commit build hook regenerated the compressed siblings with a locally-different terser/gzip/brotli toolchain than whoever generated main's copies.

Consequence: these artifacts will flip-flop between contributors, adding noise to every JS-touching PR diff.

Fix directions: (a) stop committing the `.gz`/`.br` siblings and generate them at build/release time; (b) pin the compression toolchain versions the hook uses; or (c) make the hook skip regenerating compressed siblings when the source .min.js is unchanged.

johnrtipton:
Closes #2054

## Problem

`client.min.js.gz`/`.br`/`.map` and `debug-panel.min.js.gz`/`.br`/`.map` were tracked in git. Every local `scripts/build-client.sh` run (via the `build-js` pre-commit hook or `make build-js`) touched these tracked files on disk, and a broad `git add`/`git commit -a` swept the diff into unrelated PRs — even when the source and `.min.js` itself were byte-identical. PR #2051's commit `31354b91` is the reporter's example: `debug-panel.min.js.br` (+998 B) / `.gz` (-30 B) / `.map` changed while `debug-panel.min.js` itself was byte-identical and no debug-panel source changed.

## Investigation (symptom-up)

**Who generates them:** `scripts/build-client.sh`'s `minify_and_compress()` — `gzip -n -k -f -9` and `brotli -k -f -q 11` on the terser output, plus terser's own `--source-map` flag for `.map`.

**Root cause — two distinct drift sources:**
1. `gzip`/`brotli` are unpinned system CLI tools (unlike `terser`, which is npm-pinned via `package-lock.json`), so their compressed output isn't guaranteed byte-identical across contributor machines for identical input, even with fixed flags/level (already pinned: `-9`/`-q 11`, plus the `-n` no-name/no-mtime flag).
2. `.map` files are **always** non-reproducible, independent of toolchain version: terser's `sources` field embeds the **absolute filesystem path of the checkout**. Verified empirically on a single machine — re-running the build with an UNCHANGED source showed a `.map` byte diff purely because the worktree path changed:
   ```
   old sources: ['/…/agent-adf940f22da921472/python/djust/static/djust/client.js']
   new sources: ['/…/agent-a9d2e1cdd3010c4c2/python/djust/static/djust/client.js']
   ```
   This is guaranteed to differ across every distinct clone/worktree — a strictly worse churn source than the gzip/brotli version drift.

**Who consumes the committed files (full inventory):**

| Consumer | Reads committed `.gz`/`.br`/`.map`? |
|---|---|
| djust's own `{% static %}` injection (`_inject_client_script`) | No — only ever references `client.js`/`client.min.js` |
| C013 stale-bundle system check (`configuration.py`) | No — hashes only `client.min.js` |
| Downstream WhiteNoise serving (djust.org's own `settings.py`, the documented pattern) | No — `whitenoise.storage.CompressedManifestStaticFilesStorage` regenerates its own `.gz`/`.br` at `collectstatic` time, superseding anything shipped in the wheel |
| `scripts/check-doc-snippets.py` JS-size doc-claim check (CI, `test.yml:205`) | Reads committed `.gz` size directly, but already has a graceful "bundle not found → warn and skip, exit 0" path built in for exactly this "fresh checkout before build ran" scenario (was already anticipated, never exercised until now) |
| `tests/unit/test_client_minified.py::test_gzip_sibling_exists_after_build` | Hard-asserted `.gz` existence — **updated in this PR** to skip gracefully, matching the existing "no terser" skip pattern |
| Wheel packaging (`maturin`, `python-source = "python"`) | Yes, currently ships them (nothing excludes them) — see Wheel impact below |
| CI (release.yml) | Does not rebuild JS assets at packaging time at all — ships whatever's checked out at the tag |

No consumer needs the *committed* copy. Decision: **removal**, applied identically to both the client and debug-panel families (no special-casing), per the issue's evidence-based decision framework.

## Fix

- `git rm --cached` the 6 files; added `python/djust/static/djust/*.min.js.{gz,br}` and `*.min.js.map` to `.gitignore`.
- `client.min.js`/`debug-panel.min.js` (the files actually served) stay committed, unchanged — terser's pinned-toolchain output is genuinely deterministic (confirmed empirically: re-running the build produced byte-identical `.min.js`/`.gz`/`.br` for `.gz`/`.br`, only `.map` drifted from the path leak).
- `scripts/build-client.sh` still generates `.gz`/`.br`/`.map` locally (untracked) — useful for `make start` / local WhiteNoise testing. Updated its header comment to document the rationale.
- `tests/unit/test_client_minified.py::test_gzip_sibling_exists_after_build` now skips gracefully (was a hard assert) when the artifact is absent from disk — verified the skip fires correctly with the artifact removed (gate-off, non-tautological per #1468).
- No release-workflow changes: evidence shows no consumer relies on the wheel-shipped copies (see table above), so adding release-time JS-asset rebuilding would be scope creep beyond what #2054 asks for.

## Wheel impact

`client.min.js.gz`/`.br`/`.map` and the debug-panel equivalents will **no longer be packaged in released wheels** going forward (the release workflow never rebuilds JS assets at packaging time — it only ships whatever's checked out at the tag, and these files are no longer in git). `client.min.js`/`debug-panel.min.js` — the files djust actually serves via `{% static %}` — are completely unaffected. Given the consumer audit above, this has no observed functional impact: the documented/reference WhiteNoise deployment (djust.org) regenerates its own compressed siblings regardless.

## Verification

**Gate-off (churn demonstrated, then gone):**
- Pre-change (documented in the issue): PR #2051 commit `31354b91` shows real byte-level `.gz`/`.br`/`.map` diffs with an unchanged `debug-panel.min.js`.
- Reproduced locally pre-fix: re-running `build-client.sh` with zero source changes showed `.map` diffs (absolute-path leak) even on a single machine.
- Post-fix: made a trivial whitespace edit to a `src/` module, ran the build hook, `git status` showed **only** `client.js` as modified (the genuinely-changed output) — no `.gz`/`.br`/`.map` entries at all (gitignored, untracked). Reverted the trivial edit.

**Tests:**
- `npx vitest run` — 160 files / 1821 tests passed (JS suite reads `client.js`, unaffected).
- `pytest tests/unit/test_client_minified.py` — 6/6 passed; manually removed `client.min.js.gz` and re-ran the single test to confirm it now SKIPs (not silently passes) — non-tautological per #1468.
- `pytest tests/ python/tests/ python/djust/tests/ -n auto` — 9799 passed, 218 skipped, 0 failed.
- `scripts/check-doc-snippets.py --bundle /tmp/nonexistent-bundle.gz` — confirmed graceful warn+skip, exit 0.


johnrtipton (commented):
## Code Review — drain gate (verdict: APPROVE)

CI fully green (Actions PR suites recovered mid-drain — first normally-gated merge of this bucket). Verified from the diff + report: scope is exactly the 6 committed artifacts removed + gitignore + `build-client.sh` stops generating compressed siblings/maps + `test_client_minified.py` relaxed to skip-not-pass (verified non-tautological — removing .gz makes it SKIP visibly). Consumer inventory is the load-bearing part and it's solid: static injection references only `client.js`/`client.min.js`; C013 hashes `client.min.js` only; WhiteNoise regenerates its own compressed files at collectstatic (verified against djust.org's real settings); doc-size checker had a designed skip-if-absent path. Root-cause bonus: `.map` files embed absolute checkout paths (reproduced across two worktrees) — the drain agents themselves were churning them, now impossible.

Known trade-off, accepted and documented: the "N KB gzipped" doc-claim check degrades to warn-and-skip in CI (pre-existing graceful path). Full suite 9799/0 + JS 1821/1821 on the branch.

APPROVE.

johnrtipton:
## Retrospective — PR #2076 (v1.1.0-11 drain, #2054)

**Task**: kill compressed-artifact churn.

### Quality: 5
### What went well
- Consumer-inventory-first discipline: every reader of the committed .gz/.br/.map was enumerated before deletion; the one hard dependency (a unit test) was converted to a visible skip.
- Two distinct root causes proven empirically: unpinned system gzip/brotli, and terser sourcemaps embedding absolute checkout paths (reproduced across two worktrees — the drain's own agents were a churn source).
### What didn't
- Nothing blocking. Follow-on awareness: released wheels no longer ship .gz/.br/.map (nothing consumed them); the doc-size claim check now skips in CI unless the bundle is rebuilt first.

### Verified
CI green (recovered Actions suites); full suite 9799/0; JS 1821/1821; before/after churn demonstration in the PR.

RETRO_COMPLETE

PR #2078, opened 2026-07-15
fix(test-infra): re-assert built-in assign-tag handlers on global reset — regroup xdist pollution (#2053)

### LINKED ISSUE #2053: test-infra: test_regroup_tag.py fails under some pytest-xdist orderings (passes in isolation)
9 tests in `python/djust/tests/test_regroup_tag.py` fail under `pytest tests/ python/tests/ python/djust/tests/ -n auto` on some runs/machines but pass in isolation and on CI. Reproduced on clean `main` (b416ea6e) during the ADR-025 work — NOT branch-related; one full-suite run later on the same main was 0-failed, so it is ordering-dependent.

Suspected class: a sibling test mutates process-global Rust-engine tag-handler state (`register_assign_tag_handler` registry or similar) in a worker that later runs the regroup tests — same shape as prior pollution findings (#182 canon: pollution-class fixes need 3 clean runs).

Repro attempt: run the three-root suite with `-n auto` repeatedly on macOS; observed failing set is all 9 regroup cases together.

Fix directions: identify the polluting test via `-p xdist` worker logs or `--random-order` bisection; make regroup tests re-register their handler in setup, or isolate the global registry per test.

johnrtipton:
## Summary

Closes #2053.

`python/djust/tests/test_regroup_tag.py`'s 9 `render_template`-driven tests fail under `pytest tests/ python/tests/ python/djust/tests/ -n auto` on some worker orderings, passing in isolation. This PR root-causes and fixes it.

## Root-cause narrative

**The issue's "STRONG LEAD" hypothesis was verified and disproven first** (per this repo's bug-report-triage canon — trust the symptom, distrust the cited path). The suspected polluter (`test_global_isolation_1883.py::test_reset_reasserts_theme_and_component_tag_handlers_1928`, which calls `clear_tag_handlers()`) does **not** touch the assign-tag registry at all — `clear_tag_handlers()` only clears `TAG_HANDLERS` (the plain registry); `regroup` lives in the separate `ASSIGN_TAG_HANDLERS` registry (`crates/djust_templates/src/registry.rs`). Running that exact pair in one process (`test_global_isolation_1883.py` then `test_regroup_tag.py`) stays green — confirming the lead was a red herring.

**The real polluter**, found by reproducing under `-n auto` with a custom worker-order logger and bisecting the failing worker's test history:

`tests/unit/test_tag_registry.py`'s `_isolate_tag_registries` autouse fixture (and its parallel-path twin, `tests/benchmarks/test_tag_registry.py`'s `restore_registry` fixture — #1646 class) restored built-in tag handlers after clearing all three Rust registries by looping over `djust.template_tags._registered_handlers` and calling `register_tag_handler(name, handler)` for **every** entry — including `regroup`, which is an `AssignTagHandler`. This plants `regroup` in the **plain** `TAG_HANDLERS` registry.

**Mechanism**: the Rust parser (`crates/djust_templates/src/parser.rs`) decides a tag's AST node type at **parse time** by checking `registry::handler_exists` (plain registry) **before** `registry::assign_handler_exists` (assign registry). Once `regroup` is polluted into the plain registry, every future parse of `{% regroup %}` — anywhere in that xdist worker — builds a `CustomTag` node (expects `.render()` to return a string) instead of an `AssignTag` node (expects a dict), regardless of the assign registry being correct. This is silent: `has_assign_tag_handler("regroup")` still (correctly) reports `True`, so registration-presence checks pass — only the actual render breaks, with `RuntimeError: Template error: Custom tag 'regroup' error: Handler 'regroup' render() must return a string`.

The existing #1928 fix (`reset_djust_globals()` → `reregister_builtins()`) re-asserts the *correct* assign registration before every `python/djust/tests` test, but never *strips* the stray plain-registry entry — so re-asserting the right registration doesn't heal the pollution once it's landed in the wrong registry.

## Fix

1. **`python/djust/template_tags/__init__.py`** — `reregister_builtins()` now actively `unregister_tag_handler`/`unregister_assign_tag_handler`s each built-in from the *other* (wrong) registry after registering it in the correct one. This makes the shared restore helper a complete cure regardless of which polluter (existing or future) caused the cross-registry drift.
2. **`tests/unit/test_tag_registry.py`** and **`tests/benchmarks/test_tag_registry.py`** — converged their hand-rolled (buggy) restore loops onto `reregister_builtins()` instead of reimplementing it (#1646 structural cure, retiring the class rather than patching one instance).
3. **`python/djust/tests/test_regroup_tag.py`** — added a self-healing autouse fixture (idempotent `reregister_builtins()` before each test) so the file is self-contained regardless of what ran before it in the same xdist worker.
4. **`python/djust/test_isolation.py`** — docstring update documenting the #2053 hardening.

## Regression coverage

`python/djust/tests/test_global_isolation_1883.py`:
- `test_reset_heals_regroup_cross_registry_pollution_2053` — models the real polluter (blind `register_tag_handler` loop over all built-ins), confirms the precondition (plain registry polluted, assign registry still correct — the "quiet" part of the bug), then confirms `reset_djust_globals()` strips the plain-registry entry, keeps the assign entry, and a **real** `render_template()` call succeeds afterward.
- `test_gate_off_stale_plain_registration_breaks_regroup_render_2053` — gate-off (#1468): with the pollution present and NOT healed, a real `render_template()` call fails with the exact production error message, proving the reproduction is real against the actual render path (not just registry introspection).

## Verification

- Deterministic single-process repro (the *real* smoking-gun pair, replacing the disproven STRONG LEAD pair): `pytest tests/unit/test_tag_registry.py python/djust/tests/test_regroup_tag.py` — green after the fix.
- Gate-off: `git stash` (reverting all 6 changed files) reproduces the exact original 9-test failure in the same pair; `git stash pop` restores and it's green again.
- Full three-root suite (`pytest tests/ python/tests/ python/djust/tests/ -n auto -q`) run **6 times** total during verification; **zero regroup failures in any post-fix run**. Three back-to-back fully-clean runs (0 failed, 9801 passed, 218 skipped each):
  ```
  9801 passed, 218 skipped, 218 warnings in 42.48s
  9801 passed, 218 skipped, 218 warnings in 42.71s
  9801 passed, 218 skipped, 218 warnings in 45.75s
  ```

## Known unrelated flake (out of scope, not introduced by this PR)

One of the six full-suite verification runs hit an unrelated pre-existing flake: `python/djust/tests/test_url_change_wire_version_1858.py::test_url_change_stamps_consumer_version_not_rust_counter` (a WebSocket wire-version counter race, unrelated to tag-handler registries — different mechanism entirely). It reproduced identically in the very first baseline run **before** any of this PR's changes were applied, confirming it's pre-existing and not caused or worsened by this fix. Flagged here per scope discipline (#1079) rather than folded into this PR; may warrant its own issue if it recurs.

## Test plan

- [x] `pytest tests/unit/test_tag_registry.py tests/benchmarks/test_tag_registry.py python/djust/tests/test_regroup_tag.py python/djust/tests/test_global_isolation_1883.py` — green
- [x] Deterministic pair repro (real polluter → regroup) — green after fix, red on gate-off
- [x] Full three-root suite `-n auto`, 3 consecutive clean runs
- [x] `ruff check` / `ruff format --check` / ADR + CHANGELOG hooks — all pass via pre-commit


johnrtipton:
## Code Review — PR #2078 (#2053) — APPROVE

**Root cause** (agent's symptom-up trace disproved the briefed lead): the polluter was `tests/unit/test_tag_registry.py` + `tests/benchmarks/test_tag_registry.py` re-registering every built-in via `register_tag_handler` without an `isinstance(handler, AssignTagHandler)` check — planting `regroup` (assign-only) in the PLAIN registry, which `parser.rs` checks *before* the assign registry, so the stale plain entry wins for the rest of the xdist worker.

**Verified load-bearing (#1859)**:
- The hardened `reregister_builtins()` now strips each handler from the WRONG registry via `unregister_tag_handler` / `unregister_assign_tag_handler`. I confirmed both bindings exist (`crates/djust_templates/src/registry.rs:125,568`) AND are exported by the built extension — so the `try` block is NOT silently short-circuiting to `except ImportError: return`. The fix is real, not fixture-masked.
- **CI proof**: `python-tests (py3.12)` PASSED (5m53s) against a freshly-built Rust extension — authoritative confirmation the unregister path heals the pollution.
- Structural #1646 cure: healing regardless of which polluter (existing or future) caused the drift, not a point-patch of the two fixtures.
- Agent evidence: 3× consecutive clean full-suite runs (`9801 passed`); gate-off reproduced the original 9 failures.

Verdict: **APPROVE**.

johnrtipton:
## Retrospective — PR #2078 (pipeline-drain)

**Task**: v1.1.0-11 — drained #2053 (regroup xdist tag-registry pollution — the recurring release-gate ghost).

### Quality: 5
### What went well
- Symptom-up triage disproved the briefed lead (parser.rs precedence was a *contributing* fact, but the actual polluter was two test fixtures using the wrong registry). Canonical Bug-report-triage win.
- Structural #1646 cure: `reregister_builtins()` actively unregisters from the wrong registry, healing any polluter — not a point-patch.
- Load-bearing verification (#1859): confirmed the `unregister_*` bindings exist + are exported; CI Python-tests green = real cure, not `ImportError`-masked no-op.

### What didn't
- Nothing material; the fix landed clean on first review pass.

### Verified
3× clean `9801 passed` full-suite runs; gate-off reproduced the 9 original failures; CI py3.12 + py3.14t + JS + Security all green.

Issue #2028, opened 2026-07-03
tech-debt: pin already-tagged CHANGELOG.md sections against a re-merge silently rewriting them


johnrtipton:
Source: RETRO.md v1.1.0rc5 (main/1.1 branch consolidation)

A git 3-way merge of CHANGELOG.md across two branches that diverged around a
release cut (`1.1` had renamed `## [Unreleased]` -> `## [1.1.0rc4] - <date>`
and reset Unreleased to empty; `main` kept accumulating new entries under
the still-named `## [Unreleased]`) merged with ZERO conflicts — but
git's diff3 had no notion that the two `## [Unreleased]` headers were now
semantically different, so it placed ~150 lines of `main`'s genuinely new,
unreleased content inside the position that, post-merge, read as
`## [1.1.0rc4]`'s body. This silently and falsely claimed unshipped work
(ADR-024, a security fix, 15 other fixes) had gone out in an
already-tagged, already-released version. The full test suite stayed green
throughout (nothing exercises CHANGELOG content), and neither
`scripts/check-changelog-test-counts.py` nor `make check-adr-status` catches
this class (both check the diff's own numeric/version-line claims, not
whether an already-tagged section changed at all).

Ask: a pre-commit/CI check that, for every `## [X.Y.Z]` heading in
CHANGELOG.md whose version has an existing git tag `vX.Y.Z`, asserts the
section's body (from that heading to the next `## [` heading) is
byte-identical to `git show vX.Y.Z:CHANGELOG.md`'s content for that same
section. Would have caught this exact bug automatically at merge-commit
time instead of requiring an ad hoc manual diff.

See CLAUDE.md "Process canonicalizations from the v1.1.0rc5 retro" for the
full writeup and the manual recipe used to fix it this time.

PR #2084, opened 2026-07-15
fix(websocket): drop sends on a closed socket instead of raising

szto:
## Problem

A client that sends a `mount` frame and immediately disconnects (page reload, quick navigation away) leaves that frame queued in Channels. By the time `dispatch_mount` sends the response, the ASGI server has already completed the close handshake and rejects the send:

```
RuntimeError: Unexpected ASGI message 'websocket.send', after sending 'websocket.close'.
```

The generic exception handler in `receive()` then tries `send_json` on the same dead socket, raising a second time, so every occurrence surfaces as an ERROR-level **"Exception in ASGI application"** traceback (full trace below). Observed repeatedly in a production chat deployment behind uvicorn — each quick reload mid-mount produces one.

Every server-initiated close has the same exposure: origin-reject `close(4403)`, per-IP connection limit `close(4429)`, and the auth-verdict close all return from `connect()`/`receive()` while frames the client queued before our close frame reached it are still dispatched by Channels.

<details>
<summary>Production traceback (uvicorn, websockets-sansio impl)</summary>

```
File ".../djust/websocket.py", line 2084, in receive
  await self._dispatch_runtime_owned(data)
File ".../djust/runtime.py", line 2519, in dispatch_mount
  await self.transport.send(mount_msg)
File ".../djust/websocket.py", line 2648, in send_json
  await self.send(text_data=json.dumps(data, cls=DjangoJSONEncoder))
File ".../uvicorn/protocols/websockets/websockets_sansio_impl.py", line 477, in send
  raise RuntimeError(f"Unexpected ASGI message '{message['type']}', after sending 'websocket.close'.")

During handling of the above exception, another exception occurred:

File ".../djust/websocket.py", line 2127, in receive
  await self.send_json(response)      # error response on the same dead socket
...
RuntimeError: Unexpected ASGI message 'websocket.send', after sending 'websocket.close'.
```
</details>

## Fix

- All outbound frames (JSON and the `use_binary` msgpack patch path) now flow through a single `_send_frame` chokepoint.
- `_send_frame` (a) short-circuits with a debug log once the connection is known-closed, and (b) downgrades the ASGI server's send-after-close rejection (a `RuntimeError` whose message contains `'websocket.close'`) to a debug log while remembering the closed state.
- A `close()` override and `disconnect()` set the `_ws_close_sent` flag, so responses to frames that were already queued when the server closed (4403/4429/auth verdict) are dropped instead of hitting the transport.
- Unrelated `RuntimeError`s — including uvicorn's send-before-accept rejection — still propagate. Daphne silently drops post-close sends, so only the flag path is relevant there.

## Tests

- New `python/djust/tests/test_ws_send_after_close.py` with 5 cases: swallow the rejection, short-circuit after the first rejection, unrelated `RuntimeError` propagation, `close()` marks the connection (queued-frame drop), and the binary patch path. Written test-first — they fail on main with the exact production stack.
- Related suites (`-k "ws or websocket or send or mount or transport"`): 584 passed.
- `ruff check` / `ruff format` clean; CHANGELOG entry added under [Unreleased] → Fixed.


johnrtipton (commented):
## Code Review (pipeline-ship, `--no-merge`) — Recommendation: **APPROVE**

Ran PR #2084 through the ship quality gates. This is a clean, correct, well-tested fix for the send-after-close `RuntimeError` noise. Not merged (per `--no-merge`).

### What I verified

**Correctness — single chokepoint (#1646).** Confirmed *every* outbound frame now routes through `_send_frame`: `send_json` and the `use_binary` msgpack path (`websocket.py:1462`) both go through it, and the only raw `await self.send(...)` left is the one *inside* `_send_frame` (line 2702). No parallel send path bypasses the guard.

**Fails safe.** `_send_frame` re-raises any `RuntimeError` that isn't the send-after-close case (`_is_send_after_close_error` matches on the `'websocket.close'` payload), so uvicorn's send-*before*-accept rejection and every unrelated error still propagate. Test #3 pins this (`match="boom"` propagates **and** `_ws_close_sent` stays `False`).

**Multiplexer safety (#291 — the load-bearing concern for transport side-effects in `websocket.py`).** The new `close()` override is scope-neutral — it only runs when `close()` is actually called, and it doesn't add any new close call sites. The `mount_batch` collector swaps `send_json` (→ `_collect`), not `close()`; the two `connect()`-level closes (4403 origin / 4429 per-IP, lines 1783/1802) run before any mount and are never batch-reachable; the 4429 closes at 2001/2034 are in `receive()`'s per-message rate-limit path (post-mount, not batched). **253 batch/mount/multiplex tests pass** with the override in place — no #291/#1780 regression.

**Security.** Dropping outbound frames on a known-closed / auth-rejected socket is security-*positive* (no mount data leaks to a 4403'd connection). The exception match reads an ASGI-server-generated string, not user input — no injection surface. No `mark_safe`/secrets/`print`; `%s`-style debug logging throughout.

**Tests + CHANGELOG.** 5 focused regression cases (swallow, short-circuit, unrelated-propagation, close-marks-connection with a *real* — not no-op — `close()`→`base_send`, binary path). 16 WS tests + PR CI all green. CHANGELOG entry is accurate.

### Minor recommendations (non-blocking)

1. **Consider one real-`WebsocketCommunicator` batch test.** Canon #291 notes "a fake consumer with a no-op `close()` hides the bug"; the current unit tests mock `base_send`. Since this PR doesn't change batch semantics the risk is low, but a test asserting a `close()` triggered *during* `mount_batch` doesn't drop sibling mounts would lock the invariant against future drift.
2. **Exception-message string coupling.** `"websocket.close" in str(exc)` tracks uvicorn's current wording. It's well-documented and fails safe (a wording change re-raises rather than silently swallowing), so acceptable — just noting the coupling.
3. `_ws_close_sent` relies on the `getattr(..., False)` default rather than an `__init__`/`connect` init — consistent and harmless; a one-line init would read slightly cleaner.

None of these block. Solid contribution. ✅

johnrtipton:
## Maintainer follow-up (review recommendations addressed)

Pushed a commit (`ab6e301c`) to this branch addressing my earlier review, plus one blocker I hit while committing:

**Recommendations:**
- **#3 (init):** `connect()` now sets `_ws_close_sent = False` explicitly (per-connection lifecycle default); `_send_frame` keeps its `getattr(..., False)` fallback so a directly-instantiated consumer stays safe before `connect()` runs.
- **#1 (batch coverage):** on closer inspection the #291 batch-survivor invariant is **already** locked by `test_ws_auth_close_socket.py` (`test_mount_batch_with_login_view_does_not_close_shared_socket` + the object-perm variant — real `WebsocketCommunicator`, deterministic ping→pong openness probe), and they pass with this PR's `close()` override in place. So rather than a redundant batch test, I filled the genuine gap: **`disconnect()` sets the flag too, but only `close()` was tested** (#1104). Added `test_disconnect_marks_connection_so_late_sends_are_dropped` (a late `start_async` result frame after the client left is dropped, not raised). Gate-off verified (neutering `disconnect()`'s flag-set turns it red).
- **#2 (exception-string match):** left as-is — it's correct by design and fails safe (a uvicorn wording change re-raises rather than silently swallowing).

**Blocker fixed — CHANGELOG placement (#2028).** The branch was stale, so the CHANGELOG entry had landed under the already-shipped `## [1.1.0rc7]` section. Merging that way silently rewrites shipped history (no merge conflict — the #2028 trap) and trips the `check-changelog-tagged-sections` pre-commit gate. I reconciled `CHANGELOG.md` to `origin/main` and moved the entry to `## [Unreleased]`.

Thanks for the fix — the core change (single `_send_frame` chokepoint, fail-safe error handling) is solid. Tests + CHANGELOG pins all green locally; CI re-running on the new commit.

johnrtipton (approved):
Looks good, thanks!

PR #2086, opened 2026-07-16
fix(codeql): remove always-true _warnDeadScripts guards + unused _MISSING sentinel (#2572–#2578)

johnrtipton:
Addresses the 7 CodeQL code-scanning alerts (all `Note` severity) opened against `main`.

## Alerts → resolution

| Alert | Rule | Location | Resolution |
|---|---|---|---|
| #2572–#2576 | `js/unneeded-defensive-code` | `client.js` (5 sites) | **Fixed** — removed always-true guard |
| #2577 | `py/unused-global-variable` | `bug_capture_views.py:74` | **Fixed** — removed unused `_MISSING` |
| #2578 | `py/unused-global-variable` | `urls.py:41` (`app_name`) | **Dismiss as false positive** (see below) |

## #2572–#2576 — `typeof _warnDeadScripts === 'function'` is always true
`scripts/build-client.sh` wraps every source module in **one shared IIFE**, so `_warnDeadScripts` — a top-level `function` declaration in `src/03-websocket.js`, hoisted — is visible to every module. The `if (typeof _warnDeadScripts === 'function')` guard at the 5 bare call sites (`src/17-streaming.js` ×3, `src/12-vdom-patch.js`, `src/03b-sse.js`) can never be false.

**Independent confirmation**: terser's dead-code elimination had *already* stripped these guards from `client.min.js`, so rebuilding the minified bundle produced a **byte-identical `client.min.js`** (no diff) — terser reached the same conclusion as CodeQL. The function still no-ops when `!window.DEBUG_MODE`, so the now-direct call is zero-cost in production. The pre-commit **cross-IIFE bare-reference guard (#1706)** confirms the unguarded references resolve.

## #2577 — unused `_MISSING` sentinel
`_MISSING = object()` in `bug_capture_views.py` was copied from the shared-sentinel pattern used in `decorators.py` / `rust_bridge.py` / `auth/core.py` but never wired into this module (its only reference was its own definition). Removed.

## #2578 — `app_name` is a **false positive** (dismissed separately)
`app_name = "djust"` is read by Django's `include("djust.urls")` namespace machinery — a framework-level read CodeQL's Python dataflow doesn't model. Keeping it is correct namespace-isolation design for a framework-shipped app urlconf (so djust's route names never collide with the consuming project's). Removing it would degrade the shipped urlconf's namespace contract to satisfy a linter that misunderstands Django. Dismissed via the code-scanning API with this reason.

## Verification
- Rebuilt `client.js` (+ terser → byte-identical `client.min.js`).
- JS: dead-script-warning / streaming / sse / vdom-patch suites green; full `npm test` green (pre-commit).
- Python: `test_bug_capture_*` (36) green.
- Pre-commit: cross-IIFE guard, bundle init-order lint, eslint, mypy, ruff all pass. Pre-push full suite green.


johnrtipton:
## Code Review — PR #2086 — APPROVE (posted as comment; self-approve blocked)

Addresses 7 CodeQL Note-severity alerts. Verified the analysis before changing anything:

- **Scope proof (the load-bearing question)**: confirmed `scripts/build-client.sh` wraps all modules in one shared IIFE and `_warnDeadScripts` is a top-level (unconditional) `function` declaration → hoisted, visible to every call site → the `typeof … === 'function'` guard is genuinely always-true, not a cross-IIFE availability check. So removing it can't introduce a ReferenceError.
- **Independent confirmation (#1859-style)**: rebuilding produced a **byte-identical `client.min.js`** because terser's DCE had already stripped these guards — terser and CodeQL independently agree the guards are dead. The pre-commit **cross-IIFE bare-reference guard (#1706)** passed, confirming the now-unguarded refs resolve.
- **No behavior change**: `_warnDeadScripts` still no-ops when `!window.DEBUG_MODE`.
- **#2577** `_MISSING`: confirmed its only reference was its own definition (grep) — real dead code.
- **#2578** `app_name`: dismissed as false positive (Django `include()` reads it; no in-repo reverse relies on the namespace but the framework read is real). Dismissed via API.

Tests: JS (dead-script/streaming/sse/vdom) + `npm test` + `test_bug_capture_*` (36) all green; pre-push full suite green.

Verdict: **APPROVE** pending CI.

johnrtipton:
## Retrospective — PR #2086

**Task**: address 7 CodeQL code-scanning alerts (all Note severity).

### Quality: 5
### What went well
- Verified the load-bearing question BEFORE editing: proved the bundle is a single shared IIFE with `_warnDeadScripts` hoisted, so removing the `typeof` guards can't introduce a ReferenceError. Avoided the trap of blindly deleting "defensive" code.
- Got an independent second opinion for free: terser's DCE had already stripped the guards → byte-identical `client.min.js` → terser and CodeQL agree the guards are dead.
- Correctly distinguished a genuine finding (`_MISSING` dead code, fixed) from a framework false positive (`app_name`, dismissed with a documented reason) rather than blanket-fixing or blanket-dismissing.
- Mechanical 5-site removal via single-script transformation (#1312); two-commit shape (code, then CHANGELOG).

### What didn't
- First `git push` timed out at 2min on the pre-push full-suite hook; re-pushed with a longer budget (ran the hook, didn't bypass).

### Verified
CI green (python py3.12+3.14t, rust, JS, security, demo-checks; 13 pass / 2 skip); pre-commit cross-IIFE guard (#1706) + eslint + mypy + pre-push full suite all green. #2578 dismissed via code-scanning API. Main-push CI to follow.

Issue #2081, opened 2026-07-15
test(js): dj_transition 're-runs the sequence' case still uses polling waitForClass — real-rAF flake (#1830 class)


johnrtipton:
## Symptom

`tests/js/dj_transition.test.js > dj-transition > re-runs the sequence when the attribute value changes` intermittently fails under parallel CI load:

```
Error: class "s2" never applied within 300ms
❯ Timeout.tick [as _onTimeout] tests/js/dj_transition.test.js:73:54
```

Observed on PR #2077's run (1820/1821 JS tests passed; the PR touches only Python checks). Same signature bit release gates before.

## Root cause

This test case uses the polling `waitForClass`-style helper (`Date.now() > deadline`, `setTimeout(tick, 5)`) that races jsdom's real `requestAnimationFrame` (~16ms timer). This is the exact flake class canonized in CLAUDE.md (v1.0.7-1, #1830 / PR #1839): *"never depend on the real async primitive's wall-clock timing."*

PR #1839 already converted the "active/end on next frame" case to the controllable-rAF stub (the `controlledRaf` option in `createDom` + a `flushFrame()` handle) and asserts an **ordering invariant** instead of a deadline. The "re-runs the sequence when the attribute value changes" case was not converted and still polls.

## Fix

Convert this case to drive the existing `controlledRaf` stub explicitly via `flushFrame()`, assert the ordering invariant (class-before-frame ≠ class-after-driven-frame), and drop the deadline-polling helper. Pair with a gate-off sibling (neuter the drive → post-frame assertions must fail) so it stays non-tautological (#1468).

Estimated: <1hr, single test file.

PR #2092, opened 2026-07-18
fix(test): convert dj_transition 're-runs the sequence' to controllable-rAF stub (#2081)

johnrtipton:
## Summary

`tests/js/dj_transition.test.js`'s "re-runs the sequence when the attribute value changes" case still used a polling `waitForClass` helper (`Date.now() > deadline`, `setTimeout(tick, 5)`) racing jsdom's real ~16ms `requestAnimationFrame`. Under parallel CI load it intermittently failed with `Error: class "s2" never applied within 300ms` — the #1830 flake class.

This mirrors PR #1839's fix to the *sibling* case in the same file ("applies start class synchronously, active/end on next frame"): opt into the `controlledRaf` option already present in `createDom`, drive frame advancement explicitly via `dom.flushFrame()`, and assert an **ordering invariant** instead of racing a wall-clock timer.

## The assertion I changed to

Before driving the frame:
```js
expect(el.classList.contains('s2')).toBe(true);
expect(el.classList.contains('a2')).toBe(false);
expect(el.classList.contains('e2')).toBe(false);
```
After `dom.flushFrame()`:
```js
expect(el.classList.contains('s2')).toBe(false);
expect(el.classList.contains('a2')).toBe(true);
expect(el.classList.contains('e2')).toBe(true);
```

The attribute re-trigger (`el.setAttribute('dj-transition', 's2 a2 e2')`) still goes through a real `MutationObserver` — jsdom has no controllable stub for that primitive — so the test awaits a single `Promise.resolve()` tick (FIFO-ordered after the observer's already-queued microtask per the JS microtask-queue spec) rather than polling. Everything downstream of that tick (the phase-1/phase-2/3 ordering) is fully synchronous and driven explicitly.

## Gate-off evidence (#1468)

Temporarily commented out the second `dom.flushFrame()` call (the one that drives phase-2/3 for `s2`/`a2`/`e2`) and re-ran the test:

```
AssertionError: expected true to be false // Object.is equality
 ❯ tests/js/dj_transition.test.js:249:45
    249|         expect(el.classList.contains('s2')).toBe(false);
```

Confirms the post-frame assertions are non-tautological — they only pass because the frame was actually driven. Reverted the neutering before committing.

## 10x flake check

```
for i in $(seq 10); do npx vitest run tests/js/dj_transition.test.js -t "re-runs the sequence" || echo FAIL $i; done
```
10/10 clean — zero flakes. Full file (`npx vitest run tests/js/dj_transition.test.js`) also passed 10/10 (11/11 tests each run).

## Test plan
- [x] `npx vitest run tests/js/dj_transition.test.js` — 11/11 green
- [x] Target case run 10x consecutively — zero flakes
- [x] Gate-off: neutering the frame-drive makes the new assertion fail; restored and re-verified green

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- **`tests/js/dj_transition.test.js` "re-runs the sequence when the attribute value changes" — flaky under parallel CI load (#2081, #1830 flake class).** The case polled a wall-clock deadline (`waitForClass`, `setTimeout(tick, 5)`) racing jsdom's real ~16ms `requestAnimationFrame`, intermittently failing with `Error: class "s2" never applied within 300ms`. Converted to the controllable-rAF stub already established by the sibling case above it ("applies start class synchronously, active/end on next frame", #1839): the `controlledRaf` option in `createDom` queues `requestAnimationFrame` callbacks and drives them explicitly via `dom.flushFrame()`, so the test now asserts the ordering invariant (phase-1 class lands synchronously from the `MutationObserver` attribute-change callback; phase-2/3 classes are absent until the frame is driven) instead of racing a wall-clock timer — fully synchronous once the frame is driven, no timeout. The attribute re-trigger still goes through a real `MutationObserver` microtask (jsdom has no controllable stub for it), so the test awaits a single `Promise.resolve()` tick — FIFO-ordered after the observer's already-queued microtask, not a wall-clock wait. Gate-off verified (#1468): commenting out the second `dom.flushFrame()` call makes the post-frame assertions fail (`s2`/`a2`/`e2` stay in their pre-frame state). Ran the case 10x consecutively with zero flakes.
+
 - **Send-after-close RuntimeError noise: "Exception in ASGI application" on
   every reload-mid-mount under uvicorn (#2084).** A client that sends `mount`
   and immediately disconnects (page reload, quick navigation away) leaves the
--- a/tests/js/dj_transition.test.js
+++ b/tests/js/dj_transition.test.js
@@ -203,10 +203,28 @@ describe('dj-transition', () => {
     });
 
     it('re-runs the sequence when the attribute value changes', async () => {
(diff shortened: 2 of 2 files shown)


johnrtipton:
## Code Review — PR #2092 (#2081) — APPROVE

Clean conversion of the flaky "re-runs the sequence" case to the controllable-rAF stub, mirroring the #1839 pattern (#1077 lift-from-reference).

**Verified**:
- **Ordering invariant, not wall-clock**: asserts s2-present + a2/e2-*absent* BEFORE `flushFrame()`, then s2-removed + a2/e2-present AFTER — immune to jsdom's ~16ms real-rAF race (#1830 class). The `expect(dom.flushFrame()).toBeGreaterThan(0)` guards confirm a frame was actually queued (not a silent no-op).
- **Microtask handled correctly**: the single `await Promise.resolve()` for the MutationObserver attribute-change callback is deterministic FIFO ordering, not a timer race — matches the PR #1113 canon ("own the async primitive; never depend on the real one's wall-clock timing"). The comment explains why one tick suffices.
- **Non-tautological (#1468)**: gate-off (commenting out the second `flushFrame()`) makes the s2 assertion fail — proven by the implementer.
- **10× no-flake** on the case + full file; two-commit shape; CHANGELOG names the case, not a file-total (#1106).

Verdict: **APPROVE** pending CI.

johnrtipton:
## Retrospective — PR #2092 (pipeline-drain)

**Task**: v1.1.0-12 — drained #2081 (dj_transition 're-runs the sequence' rAF flake).

### Quality: 5
### What went well
- Verbatim lift of the #1839 controllable-rAF pattern (#1077) already used by the sibling case — zero new machinery.
- Asserts an ordering invariant, not a wall-clock deadline (#1830 class); `flushFrame() > 0` guards confirm the frame wasn't a no-op; the MutationObserver microtask handled deterministically (one `await Promise.resolve()`, PR #1113 canon).
- Non-tautological: gate-off (#1468) — commenting out the second `flushFrame()` fails the s2 assertion. 10× no-flake confirmed.

### What didn't
- Nothing material; a textbook flake-conversion.

### Verified
10/10 no-flake on the case + full file; CI green (13 pass); two-commit shape; CHANGELOG names the case not a file-total (#1106).

Issue #2074, opened 2026-07-14
loop-cache: {% include %} loop bodies get a fresh keyspace every render (no cross-render reorder win inside includes)


johnrtipton:
Surfaced by the PR #2073 review: `FilesystemTemplateLoader::load_template` re-parses included templates from disk each render, so a #2067 body-identity key means include-loop bodies never hit across renders (fresh ptr per parse) — correctness over the pre-fix accidental (and unsound) sharing. Also a theoretical allocator-reuse staleness window exists for an include edited on a live non-DEBUG server without HVR (HVR's unconditional update_template->clear() covers the normal case). Fix direction if include-heavy pages matter: cache parsed includes in TEMPLATE_CACHE-style storage (stable ptrs + clear-on-change), which restores the reorder win and closes the window. Low priority.

johnrtipton:
Resolved by PR #2094 (merged, CI green, adversarial review APPROVE).

PR #2094, opened 2026-07-18
perf(templates): cache parsed {% include %} bodies for stable loop-cache identity (#2074)

johnrtipton:
## Summary

- `FilesystemTemplateLoader::load_template(name)` re-tokenizes + re-parses an included template from disk on **every** call, handing back a fresh `Vec<Node>` allocation each time.
- The #2067 loop-render cache keys each `{% for %}` body by pointer identity — `(nodes.as_ptr() as usize, nodes.len())` (`crates/djust_templates/src/renderer.rs` `content_hash` call site). Because an `{% include %}`'d template's body gets a fresh pointer every render, a `{% for %}` loop living inside an `{% include %}` **never** hits the loop cache across renders — the #2067 reorder win (O(changed)=0 re-renders on a pure reorder) is silently lost for every included template. There's also a theoretical allocator-reuse staleness window: a freed include's ptr could be reused by a *different* include on a later render, colliding on the cache key.
- Fix: a new `TemplateLoader::load_template_cached` trait method (default impl: wraps `load_template` in a fresh `Arc` — a correct no-op for loaders with no stable backing store, e.g. in-memory test loaders), overridden on `FilesystemTemplateLoader` with a real cache. Only the `{% include %}` render path (`renderer.rs`'s `Node::Include` arm) is routed through the cached method; `{% extends %}` inheritance (`build_inheritance_chain`) is untouched — it needs an owned, mutable `Vec<Node>` for block-merging and has no cross-render identity requirement.

## Design deviation worth flagging

The cache is a **process-global** `Lazy<RwLock<HashMap<PathBuf, (SystemTime, Arc<[Node]>)>>>` (`PARSED_TEMPLATE_CACHE` in `inheritance.rs`), **not** a `FilesystemTemplateLoader` instance field. I verified this against the actual production call sites before implementing (`crates/djust_live/src/lib.rs`): `render()` and `render_with_diff()` both do

```rust
let loader = FilesystemTemplateLoader::new(self.template_dirs.clone());
```

on **every single render call** — a brand-new loader instance every time, while the parsed parent `Template` (`TEMPLATE_CACHE` static) and the `LoopRenderCache` (`self.loop_render_cache` instance field) are both held persistently across renders. An instance-scoped cache field on `FilesystemTemplateLoader` would be discarded and rebuilt on every render and would never actually stabilize the pointer — the fix would be a no-op against the real production path. The process-global cache is keyed by the **resolved** path (not the raw `name` arg) and invalidated by mtime, mirroring the existing `Lazy<RwLock<HashMap<...>>>` pattern already used for the filter/tag registries in this crate (`filter_registry.rs`, `registry.rs`).

mtime-keying also means hot-reload template edits are picked up automatically on the next render with no explicit cache-invalidation wiring — a nice side effect of choosing mtime over a session/identity key.

## Reproducer-first (#1468)

`crates/djust_templates/tests/test_include_loop_cache_2074.rs`, new file:

1. `include_body_loop_reorder_is_all_hits_2074` — a real `FilesystemTemplateLoader` pointed at a temp dir, `child.html` = `{% for x in xs %}[{{ x.name }}]{% endfor %}`, parent = `{% include "child.html" %}`. Renders twice (cold, then reordered), constructing a **fresh loader instance per render** to mirror production exactly. Confirmed RED pre-fix: `0 hits / 3 misses`. GREEN post-fix: `3 hits / 0 misses`.
2. `include_body_cache_invalidates_on_mtime_change_2074` — renders once, rewrites `child.html` with different content + an explicitly bumped mtime (+120s, deterministic regardless of filesystem mtime resolution), renders again with a fresh loader, asserts the NEW content is served (not a stale cached parse).

**Gate-off** (#1468): reverted just the `renderer.rs:932` routing back to `loader.load_template(name)` (leaving the trait method + cache in place) and reran — `include_body_loop_reorder_is_all_hits_2074` failed with `0 hits / 3 misses`, confirming the routing change is load-bearing, not the trait/cache addition alone. Restored the fix afterward.

## Verification

- `cargo test -p djust_templates` — **481 passed, 0 failed** (321 lib unit + 5+1+5+2+26+2+24+3+25+66 integration + 1 doctest, across all 11 test binaries including the pre-existing #1967/#2067 loop-cache suites — no regressions).
- `cargo build --release -p djust_templates` — clean.
- `cargo clippy -p djust_templates --all-targets -- -D warnings` — clean (one `type_complexity` finding fixed via a `ParsedTemplateEntry` type alias).
- `cargo fmt -p djust_templates -- --check` — clean.
- `cargo check -p djust_live` — clean (confirms the downstream consumer of `TemplateLoader` type-checks against the new trait method; a full `cargo build -p djust_live` hits a pre-existing PyO3/libpython linking issue unrelated to this change — this worktree isn't set up for the embeddable-Python build).
- Python-level `{% include %}` + loop-cache tests were **not** run in this worktree (building the compiled extension here risks clobbering the venv per this repo's worktree-subagent convention) — CI will validate the Python surface.

## Test plan

- [x] `cargo test -p djust_templates` all green
- [x] Reproducer confirmed RED pre-fix, GREEN post-fix
- [x] Gate-off confirmed RED when routing is reverted
- [x] `cargo clippy` / `cargo fmt` / `cargo build --release` clean on the touched crate
- [x] CHANGELOG.md updated under `[Unreleased] > Fixed`
- [ ] CI (Python suite, full workspace) — pending


johnrtipton:
## Code Review — PR #2094 (#2074) — APPROVE

Independent worktree-isolated adversarial review (opus) of this core template-engine change (a new process-global parsed-include cache). Empirically probed the three real risk surfaces; the fix is sound and **strictly safer than the pre-fix code on every axis**.

**Design correctness**: the cache HAD to be a process-global static, not an instance field — production (`djust_live/src/lib.rs`) rebuilds `FilesystemTemplateLoader` every render, so an instance cache would be a no-op. The implementer verified the call sites first (#168). An `Arc` clone preserves `as_ptr()`, giving an include's `{% for %}` body a **deterministically stable** identity across renders (empirically identical `FORBODY_PTR` across 6 iterations) — allocator-independent, whereas the pre-fix path's correctness silently depended on allocator address reuse.

**Empirical verification**:
- **3-clean-runs (#182)**: 481 tests × 3 parallel + 1 `--test-threads=1`, all green; clippy 0. Global cache is keyed by unique-TempDir resolved paths + Rust integration tests are separate processes → **no pollution vector** (#2053 class cleared).
- **mtime/HVR (Concern 2)**: sub-second same-mtime blind spot CONFIRMED but **LOW** (equality-based invalidation catches any mtime change fwd/back; dev FS ns-resolution; prod immutable). Proven that HVR doesn't need to clear this cache — an edited include's new ptr → new key → natural miss + re-render of the new body; stale loop-cache fragments are unreachable.
- **Staleness window (Concern 3)**: **closes** the common case; residual collision is theoretical + strictly narrower than pre-fix.
- **Concurrency**: parse with no lock held, write-lock only on insert, no deadlock, poisoning → `TemplateError`.
- **Gate-off (#1468)**: reproducer RED against faithful pre-fix (0 hits/3 misses); fix GREEN + allocator-independent.
- **#1646**: `renderer.rs:941` include arm is the only render-time site routed to the cached method; `{% extends %}` correctly untouched.

**Two non-blocking doc-accuracy notes applied** (a 3rd commit incoming): correct the `content_hash` "clear drops every entry" invariant to note it now covers only the parent template (#1867); document the mtime-granularity HVR tradeoff.

Verdict: **APPROVE**.

johnrtipton:
## Retrospective — PR #2094 (pipeline-run, #2074)

**Task**: cache parsed `{% include %}` bodies with a stable `Arc<[Node]>` + mtime invalidation, restoring the #2067 loop-render-cache reorder-win inside includes and closing the allocator-reuse staleness window.

### Quality: 5
### What went well
- **First-principles grep before architecting (#168)**: the implementer verified the production render call sites (`djust_live/src/lib.rs`) BEFORE coding and found the loader is rebuilt every render — so my planned *instance-field* cache would have been a silent **no-op**. Corrected to a process-global static. This is the single most valuable move of the task: a wrong-but-plausible design caught by reading the actual call sites.
- **Reproducer-first (#1468)**: RED cross-render test (0 hits/3 misses) written + confirmed failing before the fix; gate-off proved the `renderer.rs` routing is the load-bearing part (not merely adding the trait method).
- **Independent adversarial review (opus, worktree)** empirically cleared all three risk surfaces: 3-clean-runs (#182, no global-cache pollution), mtime/HVR interaction (proved HVR needn't clear the cache — ptr-change → natural miss), and staleness-window closure (fix is **allocator-independent**, strictly safer than pre-fix). It also caught a subtle gate-off nuance (a non-faithful `Arc::from` gate-off perturbs the allocator into a false-GREEN) — a reminder that the gate-off must revert to the *faithful* pre-fix form.

### What didn't
- My original Planning premise (instance-field cache) was wrong for production usage — a reminder that even a Stage-4 plan for a caching change must state WHERE the cache lives relative to the object lifecycle, verified against real call sites. Relates to the v1.1.0-11 load-bearing-verification theme: the design premise itself needed empirical verification, not just the fix.

### Verified
Reproducer RED→GREEN; staleness/mtime-invalidation test; gate-off (faithful pre-fix); `cargo test -p djust_templates` 481 pass ×3 + serial; clippy 0; adversarial APPROVE; CI green (incl. the Python include surface the worktree couldn't build). Two doc-accuracy notes applied (#1867 invariant-scope + mtime-granularity tradeoff).

PR #2052, opened 2026-07-11
feat(hooks): typed values, scoped targets, djust.d.ts grafts (ADR-025)

johnrtipton:
Implements Feature D of [ADR-025](https://github.com/djust-org/djust/blob/main/docs/adr/025-js-extension-sockets.md): a data-passing and targeting contract for `dj-hook`, plus editor type support.

## What
- `dj-hook-value-<kebab>` → `this.values.<camelCase>`: JSON-first coercion (raw-string fallback; JSON-quote literal lookalikes), **live** reads via Proxy (post-morph `updated()` sees fresh values, zero staleness machinery), read-only in v1 (`set` throws; named-key access only — not enumerable, documented)
- `dj-hook-target="name"` → `this.target(name)` / `this.targets(name)` (live subtree queries, CSS.escape-guarded)
- `djust.d.ts`: the **already-shipped** declaration file gains `DjustHookContext.values`/`target()`/`targets()` + `Djust.commands.register` — purely additive grafts in the file's own naming, `tsc --strict` clean
- Namespaces chosen to avoid the taken `dj-value-*` (event params) and `dj-target` (scoped updates)
- Docs: hooks.md "Typed values & targets" + "TypeScript & editor support" sections

## The review story (worth reading)
The first d.ts attempt **replaced** the shipped file wholesale — caught because the full-suite gate ran and `python/tests/test_typescript_definitions.py` (27 content pins dating to #540) went red. Recovered by restoring main's file and grafting the additions (`31e1b91d`), correcting the docs/CHANGELOG claims (`0d15ab76`), and — closing the loop per the #1859 load-bearing-guard canon — **adding 4 new pins for the grafts themselves** (`efa7b1ae`, 31/31, gate-off proven by deleting a graft token → RED). The plan's "create the file" premise was never verified against the tree (#1516 class); the pin suite is what caught it.

## Tests
- `tests/js/hook_values_targets.test.js` — 6 cases: coercion matrix, camelCase↔kebab, liveness after attribute change, read-only TypeError (cross-realm-safe assertion), target scoping incl. outside-subtree exclusion; gate-off verified
- `python/tests/test_typescript_definitions.py` — 31/31 (27 pre-existing + 4 new graft pins)
- Full local suite: Python 8868 passed / 0 failed, JS 1812 passed; `tsc --noEmit --strict` exit 0
[...]

PR #2105, opened 2026-07-25
feat(ecosystem): dj-chart official adapter — user brings the library (#2063)

### LINKED ISSUE #2063: feat(ecosystem): official adapters — dj-chart pilot (Chart.js), user-brings-the-library
The deferred 'C' milestone from ADR-025 (now shipped: JS.ext #2051, hook values/targets #2052). One-file, no-build adapter shipped in the wheel: a pre-written morph-safe hook + pre-registered JS.ext commands for Chart.js (the #1724 teardown class), enabled via DJUST_CONFIG['extensions']. User supplies the library; djust ships only the glue. Start with exactly ONE adapter and let demand justify more. NOT in the v1.1.0-10 drain — own milestone.

From docs/internal/IMPROVEMENT_BRAINSTORM-2026-07.md §1.1 (shortlist #4).

johnrtipton:
Closes #2063.

The deferred "milestone C" from ADR-025. The two extension sockets shipped in
v1.1.0rc8 (`JS.ext.*` #2051, typed hook values/targets #2052); this is the first
adapter that rides on them.

**User brings the library; djust ships only the glue.** Chart.js is never
bundled, downloaded, or vendored.

## What ships

| File | Role |
|---|---|
| `python/djust/static/djust/ext/dj-chart.js` | The adapter: a pre-written `dj-hook` + pre-registered `JS.ext` commands. 2.3 KB gz. |
| `python/djust/extensions.py` | Single source of truth for which adapters exist. |
| `python/djust/mixins/post_processing.py` | Emits the adapter tag after `client.js`. |
| `python/djust/checks/configuration.py` | New check `djust.C015` for unknown adapter names. |

```python
# settings.py — the entire setup
DJUST_CONFIG = {"extensions": ["chart"]}
```

```html
<canvas dj-hook="Chart"
        dj-hook-value-type='"bar"'
        dj-hook-value-data='{{ chart_data_json }}'></canvas>
```

## Design notes for the reviewer

- **Not bundled into `client.js`.** The shared ~87 KB gz budget is unchanged for
  anyone who doesn't opt in; the adapter is fetched only when enabled. Pinned by
  a test asserting `client.js` contains no adapter code.
- **One table, two consumers.** The injector and C015 both resolve through
  `extensions.py`, so they cannot drift (#1646 — this repo's most-recurring
  failure class).
- **Load order is the contract.** The adapter registers into `window.djust.*`,
  which `client.js` creates. `defer` preserves document order, so emitting the
  tag *after* the client tag is what makes `commands.register` reachable —
  asserted directly rather than assumed.
- **Zero-cost when unused** (#246): nothing emitted when the list is empty.
- **`updated()` mutates in place.** Destroy-and-rebuild here is exactly the
  #1724 teardown class this adapter exists to prevent.
- **Exactly ONE adapter**, per the issue. Pinned by
  `test_chart_is_the_only_shipped_adapter`; `dj-sortable`/`dj-editor` wait for
  demand, since each is an ongoing maintenance commitment.

## Two things self-review caught (worth a look)

1. **The first draft's recipe would have broken the feature.** It told users to
   add `dj-update="ignore"` to the canvas. The patcher returns early on that
   attribute *before* attribute sync (`12-vdom-patch.js`), so the server could
   never update `dj-hook-value-data` and the chart would show its first dataset
   forever. It's also unnecessary — a `<canvas>` has no server-owned children,
   and the morph already refuses to remove a canvas's `width`/`height`. Fixed in
   `aa8ba0b3` and pinned behaviorally.
2. **Missing library warned once per re-render**, because `updated()` retries the
   mount. Now once per element (`WeakSet`).

## Tests

- **15** cases in `tests/unit/test_extensions_2063.py` — resolution (incl.
  unknown-name-dropped, non-list config), injection presence/order/`defer`, C015
  incl. suppression.
- **10** cases in `tests/js/dj_chart_adapter.test.js` — driven through the REAL
  dispatcher (`mountHooks`/`updateHooks`/`destroyAllHooks`) and real
  `js._executeOps`, not hand-built instances (#1196). The first draft *did*
  hand-roll an instance and silently double-mounted — the wiring is precisely
  what a method-level test cannot see.
- **Gate-off verified** (#1468): neutering the injection loop fails 3 of the
  injection tests.
- **C015 dogfooded** against the demo project: 0 false positives, plus an
  empirical canary confirming it fires on a typo (#1459/#1060).
- Full suites green locally: 9906 Python / 1847 JS, 0 failed.
- Wheel packaging verified empirically by inspecting a real `.whl` (nested
  static subdirs do ship).


johnrtipton (commented):
## Stage 11 review — PR #2105 (#2063) — **REQUEST_CHANGES**

(posted as a comment: GitHub refuses `--request-changes` on one's own PR)

Clean design; the single `AVAILABLE_EXTENSIONS` table structurally kills the injector/check drift class (#1646). Suites green in an isolated worktree: **9905 passed / 219 skipped** (pytest, all roots) + **1847 passed / 162 files** (vitest). Gate-off run 5 ways.

**🔴 `docs/website/guides/official-adapters.md:114-116` documents a snippet that silently does nothing.** `refresh = JS.ext.chart_update(to="#sales")` as a **class** attribute never reaches template context: `mixins/context.py:198-202` drops class-level attrs that aren't JSON-serializable (#694), and a `JSChain` isn't. Executed proof — class-attr → `'refresh' in ctx == False`; `self.refresh = …` in `mount()` → `True`. Canonical form is `js-commands.md:160` (`self.jump = …` + `{{ jump }}`). The guide also never shows the template side, so a copy-paste user gets a dead button and no error.

**🟡 The `dj-update="ignore"` claim is over-broad.** `12-vdom-patch.js:937` returns before attribute sync — true for `morphElement`. But the `SetAttr` handler at `12-vdom-patch.js:1857` checks only `UNSAFE_KEYS`/`dj-ignore-attrs`, never `dj-update`, so on the normal WS diff path `dj-hook-value-data` **would** still update. The advice stays right; the stated mechanism ("could never change") is false for the common path (#1867). Same wording at `dj-chart.js:24-28`.

**🟡 C015's wiring is unpinned.** Deleting `_check_unknown_extensions(errors)` from `check_configuration()` (`checks/configuration.py:641`) leaves all 15 new tests **and** 266 other check tests green — the tests call the private fn directly (#1859/#1860). One assert through `check_configuration(None)` fixes it.

**🟡 `tests/js/dj_chart_adapter.test.js:210`** ("warns ONCE") is tautological — it passes with hook registration gated off (0 == 0). Assert `after1 == 1` first.

**🟡** Plan Risk 2 (register per-key) never reached the guide, while `hooks.md:27/61/263` still teaches wholesale `window.djust.hooks = {…}` — which clobbers `Chart` when loaded after the adapter tag.

Checked clean: no injection surface (script paths come from the hardcoded table, never user config); wheel packaging (static subdirs ship — `src/` = 74 entries in the 1.0.7 wheel); eslint + ruff; double-destroy (WeakMap + nulled ref is idempotent); `mountHooks` already-mounted guard; the load-order test correctly went RED when I emitted the adapter tag before client.js.

Question: `DjustError` (level 40) means a typo in an optional cosmetic list hard-fails `manage.py check` / `runserver`. Deliberate over `DjustWarning`?

johnrtipton:
## Stage 14 Retrospective — PR #2105 (dj-chart adapter, #2063)

**Shipped**: ADR-025 milestone C — the first official adapter. Opt-in via
`DJUST_CONFIG['extensions']`, 2.3 KB gz served only when enabled, `client.js`
untouched. 3 impl commits + 1 review-fix commit; 14/14 CI green.

### What went well

- **Stage-4 artifact-first (#1210) paid immediately.** Writing
  `dj_chart_adapter.test.js` before the adapter existed (verified 6/6 RED)
  meant the design answered to observable behavior. It also caught the
  harness being wrong: the first draft hand-rolled a hook instance and
  reported `construct` twice, because the REAL dispatcher had already
  mounted the element. Rewriting onto `mountHooks`/`updateHooks`/
  `destroyAllHooks` + `js._executeOps` is exactly #1196.
- **The #1646 chokepoint held.** `extensions.py` as the one table both the
  injector and C015 resolve through meant there was no second place to
  forget.
- **Empirical premise-checking (#1516) twice avoided a wrong assumption**:
  inspected a real `.whl` to confirm nested static subdirs ship, and let the
  doc-snippet checker prove `from djust import JS` was wrong (it's
  `djust.js`).

### What went wrong — three defects, none caught by the green suite

1. **Self-review caught a recipe that would have broken the feature.** The
   header told users to add `dj-update="ignore"`; on the morph path that
   returns before attribute sync, freezing `dj-hook-value-*` forever.
2. **Stage 11 caught a dead doc example.** The only `JS.ext` sample used a
   CLASS attribute; `context.py` (#694) drops non-JSON-serializable class
   attrs, so a copy-paste user got a silently dead button.
3. **Stage 11 caught two non-load-bearing tests** — C015's call site could be
   deleted with 281 tests still green (#1859), and "warns ONCE" was satisfied
   by `0 === 0` (#1200).

**Common shape**: all three were invisible to a fully green suite. Tests
passing is not evidence the feature reaches a user.

### Lesson worth generalizing

**When you correct yourself, verify the new mechanism as hard as you'd verify
the original claim.** The `dj-update="ignore"` catch felt like a win, so its
*explanation* went unchecked — and was wrong (the early return is
`morphElement`-only; the `SetAttr` path never consults `dj-update`). A
confident correction is exactly when scrutiny drops. This is #1867
(falsification-test prose invariants) applied to self-review output, not just
to docs.

### Deferred

- **Open question surfaced to the user, not silently decided**: `C015` is
  `DjustError` (blocks `manage.py check`) vs `DjustWarning`. Kept as Error —
  an unknown adapter name is always a mistake and the silent no-op is the
  failure being killed — but it is a judgment call on an optional feature.

RETRO_COMPLETE

PR #2106, opened 2026-07-25
security(deps): bump postcss 8.5.16 → 8.5.20 (GHSA-r28c-9q8g-f849, #137)

johnrtipton:
Resolves Dependabot alert [#137](https://github.com/djust-org/djust/security/dependabot/137) (high).

**Advisory**: [GHSA-r28c-9q8g-f849](https://github.com/advisories/GHSA-r28c-9q8g-f849) — `postcss` ≤ 8.5.17 auto-loads a previous source map from a `sourceMappingURL` annotation without constraining the resolved path, so crafted CSS can make it read an arbitrary `.map` file off disk and surface its contents.

## Exposure

**Not a runtime exposure for djust applications.** `postcss` is a **dev-only transitive** dependency (via `vite`) — never packaged in the wheel, never served to a browser. The surface is a contributor or CI machine processing untrusted CSS.

## The diff is 7 lines

| | |
|---|---|
| `postcss` | 8.5.16 → 8.5.20 (patch floor is 8.5.18) |
| `nanoid` | 3.3.15 → 3.3.16 — postcss's own bumped requirement |

Lockfile-only. `vite`'s existing `^8.5.16` range already admits the fix, so **no `vite` bump and no `package.json` change**.

## Two things worth flagging for the reviewer

1. **`npm update postcss`, deliberately not `npm install postcss@…`.** The latter adds `postcss` to `package.json` as a direct **runtime** dependency — wrong twice over, since it is dev-only *and* purely transitive. Verified and discarded.
2. **npm 10.9.2 silently degrades the lockfile.** It strips `libc` metadata it does not itself emit from 10 `@rolldown/*` / `lightningcss-*` platform entries — fields a newer npm wrote, and which `npm ci` uses to select musl-vs-glibc binaries on Linux CI. That would have been ~37 lines of unrelated, potentially CI-affecting churn inside a security PR. Those entries are restored **verbatim from HEAD**, so this diff is exactly the security bump.

Related: this environment's npm points at a local proxy registry (`127.0.0.1:8418/npm/`) that lags public npm — it offers up to 8.5.20 while public is at 8.5.23. 8.5.20 clears the 8.5.18 patch floor, so the advisory is resolved, but the proxy will cap future bumps too.

## Verification

- `npm ci` — lockfile internally consistent, postcss 8.5.20 installed.
- Full JS suite: **1847 passing**.
- `scripts/build-client.sh` reproduces the committed bundles **byte-identically** (no artifact churn).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -84,6 +84,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
   Verified locally: the `mcp` and WebSocket/`daphne` test suites pass under the new versions (the 1.27→1.28.1 `mcp` minor bump is the largest and carries no API break in the `djust.mcp` server tests). No API or behavioral change to djust.
 
+### Security
+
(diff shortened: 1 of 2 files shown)


johnrtipton:
## Stage 14 retro — PR #2106 (Dependabot #137)

- Advisory followed: [GHSA-r28c-9q8g-f849](https://github.com/advisories/GHSA-r28c-9q8g-f849) — postcss ≤ 8.5.17 arbitrary `.map` disclosure. Alert #137 now reports `state=fixed`.
- Full regression suite ran clean: 14/14 CI checks green; JS suite 1847 passing; `npm ci` consistent; `build-client.sh` byte-identical.
- No API surface change; no behavioral change. Dev-only transitive dep — never shipped in the wheel.
- One non-obvious detail preserved for future dep bumps: npm 10.9.2 strips `libc` metadata from 10 rollup/lightningcss platform entries; those were restored verbatim from HEAD so `npm ci`'s musl-vs-glibc selection is unaffected.

RETRO_COMPLETE

PR #2107, opened 2026-07-25
fix(client): scan the root element itself for dj-window-*/dj-document-* (#2097)

### LINKED ISSUE #2097: dj-window-*/dj-document-* scoped attrs go dead after the WS-mount morph (stale registry; root element also never scanned)
## Symptom

In a downstream app (snake-arena, djust 1.0.8 and 1.1.0rc8), `dj-window-keydown="key"` on markup inside `dj-root` never fires — no client error, no WS frame, silent dead binding. Reproduced with both real keypresses and synthetic `window.dispatchEvent(new KeyboardEvent('keydown', ...))`.

Calling `window.djust.handleEvent('key', {key: 'ArrowUp'})` directly works, so the event path is fine — the break is in the scoped-attr delegation registry.

## Root causes (two, both in `_scanScopedElements` / `_installScopedDelegation`, `static/djust/client.js` ~3633–3760)

1. **Post-morph stale registry.** `_installScopedDelegation()` guards with `_scopedDelegationInstalled` and scans once at install ("once at mount and on TurboNav — NOT after every patch"). The #1610 WS-mount morph then **replaces the dj-root subtree**, so every registry entry points at a removed node; dispatch skips them via the `document.contains(entry.element)` check. Result: all `dj-window-*` / `dj-document-*` attrs rendered inside `dj-root` are permanently dead after WS mount.
2. **Root element never scanned.** The scan does `root.querySelectorAll('*')` where `root = document.querySelector('[dj-view]') || ...` — `querySelectorAll` matches descendants only, so a scoped attr placed ON the dj-root/dj-view element itself is never registered even pre-morph.

## Suggested fix

Re-run `_scanScopedElements()` after the WS-mount morph (and arguably after any patch batch that inserts elements carrying scoped attrs — the scan is cheap relative to a morph), and include the root element itself in the scan (e.g. scan `[root, ...root.querySelectorAll('*')]`).

## Workaround used downstream

Page-level listener OUTSIDE dj-root delegating to the public surface:

```html
<script>
  window.addEventListener("keydown", function (e) {
    if (window.djust && window.djust.handleEvent) {
      window.djust.handleEvent("key", { key: e.key });
    }
  });
</script>
```

Same family as #1848 (mount morph drops inline-script execution) — the mount morph invalidates init-time client wiring inside dj-root.

johnrtipton:
Closes #2097.

## The bug

`_scanScopedElements()` resolves the LiveView root, then scans
`root.querySelectorAll('*')` — which matches **descendants only**. A
`dj-window-*` / `dj-document-*` attribute written on the `dj-view`/`dj-root`
element *itself* was therefore never registered, and the directive was silently
dead.

```html
<!-- dead before this PR -->
<div dj-view="app.SnakeView" dj-window-keydown="key"> … </div>

<!-- worked, one level deeper -->
<div dj-view="app.SnakeView"><div dj-window-keydown="key"> … </div></div>
```

The failure mode is what made this expensive downstream: **no client error, no
WS frame**, and `window.djust.handleEvent('key', {...})` by hand works
perfectly — which sends you debugging the event path rather than the delegation
registry. Reported from snake-arena, whose game board *is* the `dj-view` root.

## Scope correction — reproducer-first changed what this PR fixes

The issue reported **two** causes. Running the reproducer against `main` settled
both, and disagreed with the issue text *and* with my own follow-up comment:

| Reported cause | Verdict |
|---|---|
| **2 — root element never scanned** | **Real.** 5 of 6 new tests fail on `main`. Fixed here. |
| **1 — post-morph stale registry** | **Already fixed** — by **#1996**, which made `bindLiveViewEvents()` re-run the scan on *every* invocation (only the window/document `addEventListener` install stays one-shot). |

The issue body's "scans once at install" described **pre-#1996** code. My
follow-up comment then guessed cause 1 was "fixed by morph behavior rather than
a rescan" — also wrong. The reproducer is what established the real answer.

That #1996 per-bind-rescan behavior is now **pinned in this file too**, so a
future refactor cannot revert it and have the two causes conflated again.

## The fix

Include the root in the scan list. When `root` falls back to `document`,
nothing is prepended — `document` has no attributes, and
`document.querySelectorAll('*')` already covers `<html>`/`<body>`.

## Tests — 6 cases in `tests/js/dj-window-root-scan-2097.test.js`

Root `dj-window-keydown` · root `dj-document-keydown` · key filter at root
(`.escape`) · root **and** descendant both bound, no double-fire across
repeated binds · re-registration after the root element is **replaced** ·
the #1996 per-bind-rescan pin.

All drive **real** `window`/`document` KeyboardEvents through the real
`bindLiveViewEvents()` path — no direct method invocation (#1196).

**Gate-off verified** (#1468): reverting to the descendants-only list fails
**5 of 6**. The 6th is the #1996 pin, which correctly stays green because it
covers a different fix.

## Verification

- Full JS suite: **1853 passing** (163 files). Adjacent suites
  `dj-window-events` + `dj-window-rescan-1996`: 14 passing.
- Full Python suite: **9908 passing**.
- Bundle: **+31 bytes gz**.

## Deliberately out of scope

The scan is rooted at the LiveView root, so a scoped attr placed *outside*
`dj-root` is also never registered. That's a real adjacent constraint, but
beyond what #2097 cites — left alone per the repo's scope-discipline rule
(#1079). Happy to file a follow-up.


johnrtipton (commented):
**Stage 11 — APPROVE.** Verified independently in an isolated worktree (branch 0 commits behind `main`).

**Scope claim (#1996) — CONFIRMED, plus corroboration you didn't cite.** `68d8a214` did move `_scanScopedElements()` out of the one-shot install; `bindLiveViewEvents()` calls it at `09-event-binding.js:1179`, reached via `reinitAfterDOMUpdate()` (`:1526`) on every patch (`03-websocket.js:611,679,782,880,1392`). Cause 1 is dead. Extra evidence: `git tag --contains 68d8a214` → rc5–rc8, so the reporter's **rc8** run already had cause 1 fixed while the symptom persisted — cause 2 is the only remaining explanation. Also, the stale comment this PR deletes (`main`'s `:146`, *"runs once at mount and on TurboNav — NOT after every patch"*) is **verbatim** what the issue quoted; #1996 updated the header block and missed that line.

**Gate-off** (reverted to descendants-only, rebuilt): 5 RED (root keydown, root `dj-document`, key filter, root+descendant, root-replaced), 1 GREEN — the #1996 pin, correctly unaffected. No tautologies.

**Edge cases.** `root === document`: guard is load-bearing — `document.attributes` is `undefined` and the unguarded `[root].concat(…)` form throws; `<body dj-window-keydown>` still binds. 10 repeated binds → exactly 1 fire. Root replacement is genuine (`newNode !== oldNode`), not mutation. Nested second `[dj-view]` registers; sibling second root doesn't (pre-existing, matches your out-of-scope note). `scripts/build-client.sh` reproduces committed `client.js`/`client.min.js` byte-identically. Suites: JS 1853/163 files; Python 9908 passed, 218 skipped.

🟡 **Real concern — per-patch allocation, `:155-157`.** Pre-fix `NodeList.forEach` allocated nothing. Now each bind allocates `Array.from(descendants)` (N) **plus** `[root].concat(…)` (N+1); the `root === document` branch's `Array.from` is pure waste since NodeList already has `.forEach`. That's a per-WS-patch hot path in a top-3-budget module. Zero-alloc shape: hoist the body to `scanEl`, `if (root !== document) scanEl(root);` then `descendants.forEach(scanEl)`. Worth doing here?

🟡 **Follow-up (pre-existing, non-blocking).** `alreadyRegistered` (`:192-199`) compares `(element, attrName)` but never the attr **value**, and never refreshes `entry.parsed`. Measured: `"old"` → `"renamed"` in place + rebind still fires `old`, root *and* descendant. The root is exactly what morphdom mutates in place rather than replacing, so this PR makes it more reachable. File it?

**Could not verify:** the issue says the attr was *inside* `dj-root`; the PR says snake-arena's board *is* the root. Per #1849 only a browser run settles that.

johnrtipton:
## Stage 14 Retrospective — PR #2107 (#2097 scoped-attr root scan)

**Shipped**: `_scanScopedElements()` now scans the LiveView root element itself,
not just its descendants. 4 commits, 14/14 CI, +29 bytes gz, Stage 11 APPROVE
with 0 must-fix.

### The headline: a stale comment caused the misdiagnosis

The issue body I filed quoted `client.js` *"This runs once at mount and on
TurboNav — NOT after every patch"* as evidence for a post-morph stale-registry
bug. The Stage 11 reviewer found that line is **verbatim** the inline comment at
`09-event-binding.js:146` on `main` — and that **#1996 had already made the scan
run on every bind**, updating the header block above `_scopedRegistry` but
missing this inline line. A stale comment sent the reporter (me) after a bug
that no longer existed, and the real bug (root never scanned) sat one line below
it. This PR deletes it.

**Generalizable**: when a fix changes a function's invocation contract, grep for
EVERY comment describing the old contract, not just the one above the symbol you
edited. A stale comment is not cosmetic — it is a future misdiagnosis with a
long fuse. Same family as #1197 (canon-doc citation discipline), applied to
inline comments.

### Reproducer-first re-scoped the work (#1210 paying off again)

The issue claimed two causes. Writing the reproducer BEFORE touching code showed
5/6 RED for cause 2 and GREEN for cause 1 — so the PR fixed exactly one cause
and pinned the other's existing fix. Notably it also refuted my OWN follow-up
comment, which had guessed cause 1 was fixed "by morph behavior rather than a
rescan." Both the original report and my correction to it were wrong about
mechanism; only the executable artifact was right.

**This is the second time in one session** the same lesson landed (PR #2105: a
confident self-correction about `dj-update="ignore"` had a wrong mechanism). The
pattern is specific: *the moment I feel I've caught something, scrutiny of the
explanation drops.*

### What Stage 11 caught that I would have shipped

1. **A perf regression I introduced** — `Array.from(descendants)` +
   `[root].concat(...)` allocated twice per WS patch in a top-3 client module.
   Rewritten allocation-free (hoisted `scanElement`, iterate the NodeList).
2. **A load-bearing guard with no test** — `root !== document` prevents a
   `TypeError` that would kill every scoped binding on the page. Verified by
   hand by the reviewer, pinned by me. Gate-off now fails exactly it.

### Scope discipline held (#1079)

The reviewer surfaced a real PRE-EXISTING bug: the registry dedupes on
`(element, attrName)` and never refreshes a changed attribute VALUE, so renaming
a handler in place keeps firing the old one. I reproduced it, then filed **#2108**
rather than widening this PR — while noting that this change makes it *more*
reachable, since the root is the node morphdom mutates in place.

### Honest gap

The issue body says the attr was on markup *inside* `dj-root` (a descendant);
the PR body says the downstream board *is* the root. These disagree and neither
I nor the reviewer can see the downstream markup. The rc8 tag evidence makes the
root reading the only consistent one, but per #1849 only a browser run against
the real app settles it. Flagged rather than assumed.

RETRO_COMPLETE

PR #2109, opened 2026-07-25
fix(client): refresh scoped-attr registry entries on value change (#2108)

### LINKED ISSUE #2108: dj-window-*/dj-document-* registry keeps a STALE handler when the attribute value changes in place
Surfaced by the Stage 11 review of PR #2107 (#2097). **Pre-existing** — not introduced by that PR — but #2107 makes it more reachable, so filing rather than silently carrying it.

## The bug

`_scanScopedElements()` (`python/djust/static/djust/src/09-event-binding.js`) dedupes registry entries on `(element identity, attrName)` only:

```js
entries.forEach(function(entry) {
    if (entry.element === element && entry.attrName === attrName) {
        alreadyRegistered = true;
    }
});
if (alreadyRegistered) continue;
```

The attribute **value** is never compared, and `entry.parsed` is never refreshed. So if the server re-renders and changes the handler *in place* on a surviving element:

```html
<!-- before -->  <div dj-window-keydown="old">
<!-- after  -->  <div dj-window-keydown="renamed">
```

…the element is "already registered", the new value is dropped, and the **old** handler keeps firing indefinitely.

## Reproduced

JSDOM, real `bindLiveViewEvents()` + real window KeyboardEvent: mount with `dj-window-keydown="old"`, `setAttribute` to `"renamed"`, rebind, dispatch → server receives **`old`**. Affects descendants *and* the root element equally.

## Why #2107 makes it more reachable

#2107 adds the dj-view/dj-root element itself to the scan. That element is precisely the node morphdom **mutates in place** rather than replacing (it is the morph anchor), so it is the most likely node in the page to hit the stale-value path. Elements that get replaced wholesale dodge the bug, because the `document.contains()` sweep evicts the old entry.

## Suggested fix

Include the attribute value in the dedupe comparison, or refresh `entry.parsed` on match:

```js
if (entry.element === element && entry.attrName === attrName) {
    entry.parsed = parsed;        // refresh instead of skip
    entry.requiredKey = requiredKey;
    alreadyRegistered = true;
}
```

Needs a regression test in the shape of `tests/js/dj-window-root-scan-2097.test.js` (rename in place → rebind → assert the NEW handler fires), plus a gate-off sibling.

## Related

- #2097 / PR #2107 — root-element scan (where this was found)
- #1996 — the per-bind rescan that makes this path run on every patch

johnrtipton:
Closes #2108.

## The bug

`_scanScopedElements()` deduped registry entries on `(element identity, attrName)`
and **skipped** on a match. The attribute *value* was never compared and
`entry.parsed` was never refreshed, so a server re-render that changed a handler
on a **surviving** element was silently dropped:

```html
<!-- server renders -->  <div dj-window-keydown="old">
<!-- then re-renders -->  <div dj-window-keydown="renamed">
<!-- client keeps firing -->  old
```

**Why in-place mutation is the realistic shape**: morphdom *patches* a surviving
element's attributes rather than replacing the node. An element that gets
**replaced** dodges this entirely — the `document.contains()` sweep evicts its
entry — so only surviving nodes bite. That includes the `dj-root`/`dj-view`
element, which is the morph anchor and never replaced. (Which is why #2107,
by adding the root to the scan, made this path more reachable.)

Pre-existing on the descendant path too; surfaced by the Stage 11 review of #2107.

## The reproducer found a second half the issue didn't describe

I filed #2108 about stale *values*. Writing the reproducer surfaced an
independent failure: when the server **removes** the attribute, the entry is
never evicted either — `contains()` only catches removed *elements*. So a
surviving element whose `dj-window-keydown.escape` disappeared kept dispatching
a directive the template no longer declares.

| Half | Covers |
|---|---|
| **Refresh** `entry.parsed` + `entry.requiredKey` on match | renamed handler, changed inline args, changed key filter |
| **Evict** entries whose attribute is gone | attribute removed from a surviving element |

## Gate-off verified PER HALF (#1468)

Proving neither half is dead code — a combined gate-off would not have shown this:

| Gated off | Result |
|---|---|
| eviction only | fails **exactly** the key-filter test |
| refresh only | fails the other **three** |
| both | fails **4 of 5** |

The 5th test is the one that must stay GREEN: it pins that an **unchanged**
value still does not double-register. "Refresh instead of skip" could easily
have become "append a duplicate entry every bind", and that test is what stops it.

## Tests — 5 cases in `tests/js/dj-window-stale-handler-2108.test.js`

Descendant rename · root rename · key-filter swap (`escape` → `enter`) ·
unchanged-value no-double-register across 5 binds · **inline-arg refresh with an
unchanged handler NAME** — the nastiest shape, because the event still looks
completely correct on the wire while carrying stale arguments.

All drive real `window` KeyboardEvents through the real `bindLiveViewEvents()`
path (#1196), not direct method invocation.

One test assumption of mine was wrong and I fixed the *test*, not the code:
inline args arrive as raw strings in `body._args`, not as top-level kwargs. I
probed the actual wire shape rather than guessing.

## Verification

- Full JS suite: **1859 passing** (164 files). Full Python: **9908 passing**.
- Bundle: **+34 bytes gz**.


johnrtipton (commented):
## Stage 11 Code Review — **APPROVE** (GitHub blocks self-approve)

Isolated worktree; rebuilt for every probe. Branch is **0 commits behind `origin/main`** (#1450).

### Per-half gate-off matrix — reproduced, matches the claim exactly (#1468/#1859)

| gated off | failures in `dj-window-stale-handler-2108.test.js` |
|---|---|
| eviction (`09-event-binding.js:146`) | **1** — key-filter only |
| refresh (`09-event-binding.js:206-207`) | **3** — descendant, root, inline-args |
| both | **4** of 5 |
| as shipped | **0** |

Neither half is decorative.

### Adversarial probes

- **Dedupe survives refresh.** 20 identical binds → one fetch; 20 binds each with a *changed* value → exactly `["h19"]`. One entry, latest value — refresh did not become "append per bind". Duplicates are unconstructible: `add` at `:213` is gated by the full-Set scan at `:204`.
- **Mutation-during-iteration is safe.** 6 entries in one Set; evicted 5 interleaved across *both* conditions (incl. the Set's first and last) → exactly `["h3"]`. Nothing skipped — `Set.forEach` only skips elements deleted *before* being visited, and only the current is deleted.
- **No non-Element reaches `getAttribute`.** `root === document` fallback and `<html>`-carrying-the-attr both no-throw; the `||` at `:145` also spares detached elements.
- **Perf is a non-issue.** At 300 scoped entries + 2000 filler elements: 43.56 ms/bind, of which the new sweep is **0.08 ms (0.18%)** — dominated by the pre-existing `root.querySelectorAll('*')`.
- **Artifacts.** Bundles rebuild **byte-identical** to committed; gz delta **+34 bytes**, the CHANGELOG's exact claim. No banned patterns in `src/`.

Suites: vitest **1859/1859** (164 files); pytest **9908 passed, 218 skipped, 0 failed**.

### 🟡 One genuine concern — scope asymmetry (follow-up, not a blocker)

The sweep is **document**-scoped; the refresh is **root**-scoped (`09-event-binding.js:134`). An entry still in `document` but *outside* the selected `[dj-view]` is neither refreshed nor evicted, so it keeps firing the OLD handler after an in-place rename. Reproduced → `["old"]`; also via multi-root: register under root B, insert root A before it, `querySelector('[dj-view]')` picks A → B's entries go stale permanently (`["bOld"]`). Multi-`[dj-view]` pages are real (sticky children / `live_render`). Strictly narrower than pre-PR behavior — no regression — so a follow-up per #1079, not widening here.

johnrtipton:
## Stage 14 Retrospective — PR #2109 (#2108 stale scoped handler)

**Shipped**: `_scanScopedElements()` now refreshes a registry entry when the
attribute value changes, and evicts entries whose attribute is gone. 2 commits,
13/13 CI, +34 bytes gz, Stage 11 APPROVE with 0 must-fix.

### The reproducer found half the bug I hadn't filed

I filed #2108 about stale VALUES. Writing the reproducer surfaced an independent
failure the issue never mentioned: a REMOVED attribute is never evicted either,
because the `document.contains()` sweep only catches removed *elements*. My
key-filter test caught it incidentally — Escape kept firing after I removed
`dj-window-keydown.escape`.

**Generalizable**: an issue I wrote myself, from a bug I reproduced myself, still
described only half the defect. Writing the reproducer as an *exploration* of the
surface — not just a witness for the known symptom — is what found the rest. That
is the third time this session reproducer-first changed the scope of the work
(#2097 narrowed it, #2063 caught a broken recipe, this widened it).

### Per-half gate-off was the right call

The fix had two independent halves, so gating them off *together* would have
proven only that "something is load-bearing". Gating each separately proved
neither is dead code (#1859):

| gated off | fails |
|---|---|
| eviction only | exactly the key-filter test |
| refresh only | the other three |
| both | 4 of 5 |

The reviewer reproduced this matrix exactly. **Rule worth keeping**: when one fix
contains N independent mechanisms, gate off each one separately — a combined
gate-off cannot distinguish "both needed" from "one needed, one decorative".

### A test assumption of mine was wrong, not the code

I asserted inline handler args arrive as top-level kwargs (`body.dir`). They
ride as raw strings in `body._args`. I probed the real wire shape instead of
"fixing" working code to match a wrong expectation. Same instinct that was needed
on #2105 (`_registry` is a Map, so `Object.keys()` is always `[]`).

### What the review added that I could not have self-supplied

Measurement. I *worried* the new per-entry `getAttribute` was a hot-path
regression but had no number. The reviewer measured it: 0.08 ms of a 43.56 ms
bind (0.18%) at 300 scoped entries + 2000 filler elements. It also proved the
dedupe survives 20 binds with changing values, and that mid-iteration `Set.delete`
skips nothing — both things I had reasoned about rather than measured.

### Deferred (#1079)

Review surfaced a real residual gap: the eviction sweep is **document**-scoped
while the refresh is **root**-scoped, so an element outside the selected
`[dj-view]` is neither refreshed nor evicted. Verified independently before
filing (`["old"]`; multi-root shape gives `["bOld"]`). Strictly narrower than
pre-PR behavior — no regression — so filed as **#2110** rather than widening
this PR.

### Honest caveat

All of this is JSDOM. Real morphdom attribute-patching was simulated with
`setAttribute`, not driven through a server re-render. That is the repo's
standard for JS tests, but it is not a browser (#1849).

RETRO_COMPLETE

PR #2111, opened 2026-07-25
fix(client): scan EVERY LiveView root; keep sweep and scan scope-symmetric (#2110)

### LINKED ISSUE #2110: Scoped-attr registry: eviction sweep is document-scoped but refresh is root-scoped — entries outside the selected [dj-view] go permanently stale
Surfaced by the Stage 11 review of PR #2109 (#2108). **Not a regression** — strictly narrower than pre-#2109 behavior — but a residual gap worth closing.

## The asymmetry

In `_scanScopedElements()` (`python/djust/static/djust/src/09-event-binding.js`):

- The **eviction sweep** walks the whole `_scopedRegistry` and keeps any entry whose element is still `document.contains(...)` with its attribute present.
- The **refresh** (#2108) only visits `root` + `root.querySelectorAll('*')`, where `root = document.querySelector('[dj-view]') || document.querySelector('[dj-root]') || document`.

So an entry whose element is still in the document but **outside the selected root** is neither refreshed nor evicted. It keeps dispatching its ORIGINAL handler forever.

## Reproduced (two shapes)

**1. Element moved out of the root, then renamed in place** — verified in JSDOM against the real `bindLiveViewEvents()` path:

```
register under root -> move element outside root -> setAttribute to 'renamed' -> rebind -> press
=> server receives 'old'
```

**2. Multi-root (the realistic one).** Register under root B, then a patch inserts root A earlier in the document. `querySelector('[dj-view]')` now picks A, so every entry belonging to B is permanently stale: `["bOld"]`.

Shape 2 matters because **multi-`[dj-view]` pages are a real djust shape** — sticky children and `{% live_render %}` both produce them.

## Why it is not a regression

Before #2109 every in-place rename was stale regardless of location; now only out-of-root ones are. This is a shrink of the bug, not a new one.

## Suggested fix

Make the two scopes agree. Either:

- **Scope the sweep to the root** — evict any entry whose element is not within the current root (it can no longer be refreshed, so it should not dispatch), or
- **Scope the scan to the document** — refresh every registered element wherever it lives.

The first is likely correct: the registry should only dispatch for elements the scan actually governs. It also cleans up entries orphaned by a root switch.

Needs a regression test in the shape of `tests/js/dj-window-stale-handler-2108.test.js` (both shapes above), plus a gate-off sibling.

## Related

- #2108 / PR #2109 — the refresh + eviction fix that surfaced this
- #2097 / PR #2107 — root-element scan
- #1996 — the per-bind rescan

johnrtipton:
Closes #2110.

## The bug — and it was wider than I filed it

`_scanScopedElements()` resolved **one** root via
`querySelector('[dj-view]') || '[dj-root]'`, while the eviction sweep walked the
whole registry **document-wide**. The two scopes disagreed, so an entry outside
the selected root was neither refreshed nor evicted and kept dispatching its
original handler forever.

Writing the reproducer showed the gap is bigger than #2110 described. A second
root's scoped attrs **never registered at all** — not merely went stale —
because `querySelector` returns only the first match:

```html
<div dj-view="app.A"></div>
<div dj-view="app.B" dj-window-keydown="fromB">   <!-- completely dead -->
```

## Why not the fix I originally suggested

#2110 proposed scoping the *sweep* down to the root. That would have made things
**worse**: a sibling root's entries would be evicted outright, turning "stale"
into "dead". So the fix goes the other way — the scan and the sweep now resolve
the **same** set (every `[dj-view]`/`[dj-root]`), which retires the drift class
rather than re-balancing two rules that can disagree again.

**The common shape was already fine**: `{% live_render %}` and sticky views embed
children *inside* the parent, so nested roots were covered by the outer
`querySelectorAll`. The broken case is a root that is **not an ancestor** of the
registered element.

## Two independently load-bearing halves

| Half | Covers |
|---|---|
| Scan **every** root (+ descendants) | second root never registered; orphaned-root staleness |
| Evict entries not governed by **any** root | element that leaves the root but stays in the document |

A nested root is visited twice (as a descendant, then as a root). Harmless,
because `scanElement` **refreshes** an existing entry rather than adding a second
one (#2108) — pinned by an explicit no-double-fire test.

**Gate-off verified per half** (#1468) — a combined gate-off could not tell
"both needed" from "one decorative":

| Gated off | Fails |
|---|---|
| scan-all-roots | the two multi-root tests |
| `isGoverned` eviction | the moved-out test |

## Perf — measured, not assumed

A/B against `main`'s bundle on an **identical** 2200-element page:

| | ms/bind |
|---|---|
| main | 35.50 |
| this branch | 36.58 (**+3%**) |

Nested-root double-visit costs +5% (37 → 39 ms) — well short of the 2× the shape
suggests, because the per-bind cost is dominated by work outside this scan.

## Tests — 5 cases in `tests/js/dj-window-scope-symmetry-2110.test.js`

Moved-out element · multi-root orphan · second root scanned at all · nested child
root (the common `live_render` shape) still working · no double-fire when an
element is reachable from two roots.

Sibling suites all still green — **26 passing** across #2097, #2108, #1996 and
`dj-window-events`. Full suites: **1864 JS**, **9908 Python**.

Bundle: **+101 bytes gz**.


johnrtipton (commented):
**Stage 11 — verdict: APPROVE** (0 🔴, 2 🟡). GitHub blocks `--approve` on one's own PR, so recording it here.

**Gate-off matrix reproduced independently** (rebuilt bundle each time) — both halves are load-bearing, not decorative (#1859):

| state | failures in `dj-window-scope-symmetry-2110.test.js` |
|---|---|
| scan-all-roots off | 2 — multi-root orphan, second-root-scanned |
| `isGoverned`-eviction off | 1 — moved-out element |
| both off | 3 |

Siblings green in every state. Full suites: vitest **166 files / 1874 tests**, pytest **9908 passed, 218 skipped, 0 failed**. `scripts/build-client.sh` reproduces the committed `client.js`/`client.min.js` byte-for-byte. eslint 0. Depth-3/4 probes: each nesting level fires **exactly once**; `.escape` filters and two different scoped attrs on one doubly-visited element are unaffected. `isGoverned` edge cases all correct (no-root fallback, root-appears-later eviction, detached subtree no-throw, `<head>` not governed).

🟡 **1. Scan cost is quadratic in nesting DEPTH, not the "+5%" claimed.** `src/09-event-binding.js:259-262` re-walks every level's full subtree. My A/B (2200 elements constant, median of 60 binds): depth 1 **+0.8%**, 2 +7.1%, 3 +12.1%, 4 +19.5%, 10 +49.8%, 50 **+307.8%** (28.8→117.5 ms). Single-root reproduces your headline (+2.8%) and is the common shape (`<div dj-root dj-view=…>` is one element) — but nested `live_render` is +10%, not +5%, and grows.
**Verified cure:** keep only OUTERMOST roots (`!r.parentElement.closest('[dj-view], [dj-root]')`). Containment is transitive, so the set is *exactly* equivalent for both the scan and `isGoverned`. I built it: **41/41 tests pass** and perf goes flat vs main at every depth (depth 50: 30.0 ms vs 116.6 ms).

🟡 **2. `isGoverned` is O(entries × roots) per bind** (`:147-155`, called at `:170`). Sibling roots (cure 1 doesn't help): 10 roots +14%, 50/150 entries +30%, 100 roots +42%.

**Note — undocumented widening.** Old `querySelector('[dj-view]') || '[dj-root]'` ignored `[dj-root]` whenever any `[dj-view]` existed. Empirically, `<div dj-root><aside dj-window-keydown="x"></aside><div dj-view>…</div></div>`: main → aside **dead**; branch → aside **fires**. Likely correct, but it activates previously-inert directives and the CHANGELOG frames the bug only as "a second root's attrs never registered." Worth a sentence.

johnrtipton:
## Review findings addressed

**🟡 1 (perf, quadratic in nesting depth) — fixed in `06252bfb`.** Your curve was right and my PR-body figure was wrong: I measured depth 2 and generalized a flat "+5%" from it. Applied the cure you verified — filter to **outermost** roots only. Containment is transitive so the governed set is exactly equivalent for both the scan and `isGoverned`, and each subtree is walked once instead of once per level. Implemented as a single document-order pass (an ancestor always precedes its descendants), so it costs O(n) containment checks rather than O(n²).

Re-measured on my harness, median of 25 binds:

| depth | delta vs main |
|---|---|
| 1 | −5.3% |
| 2 | +5.3% |
| 4 | +5.3% |
| 10 | 0.0% |
| 50 | **+8.7%** (was +308%) |

Flat in depth. All tests unchanged: 5 in the #2110 suite, 26 across the sibling suites, 1864 full JS, 9908 Python.

**🟡 2 (many SIBLING roots, `isGoverned` O(entries × roots)) — accepted and documented, not fixed.** Nested roots now collapse to one entry so they no longer contribute; only genuine siblings do, and your numbers (~+14% at 10 roots, ~+42% at 100) are on root counts a real page does not reach. Recorded in the CHANGELOG as a measured known characteristic rather than left implicit.

**Scope-widening note — documented in `75664f70`.** You were right that the CHANGELOG framed the bug too narrowly. The old `querySelector('[dj-view]') || '[dj-root]'` ignored `[dj-root]` entirely whenever any `[dj-view]` existed, so a scoped attr inside `dj-root` but outside `dj-view` was inert and now fires. Intended — the element is inside the patched region — but users upgrading will see previously-dead directives activate, which deserved saying out loud.

Thanks for building and verifying the cure rather than just flagging the cost; that turned a 'known regression' into a fix.

johnrtipton:
## Stage 14 Retrospective — PR #2111 (#2110 scope symmetry)

**Shipped**: the scoped-attr scan and eviction sweep now resolve the same
governed set — every OUTERMOST `[dj-view]`/`[dj-root]`. 4 commits, CI green,
+125 bytes gz, Stage 11 APPROVE with 0 must-fix.

### The finding that matters: I published a wrong perf number

I reported the nested-root double-visit as a flat **+5%**. It is quadratic in
nesting depth — +19.5% at depth 4, **+308% at depth 50** — and the realistic
`live_render` shape was +10.1%, double my figure.

The error was methodological, not arithmetic: **I measured depth 2 and
generalized from a single data point.** I had explicitly set out to measure
rather than assume (having been caught assuming on PR #2109), did the
measurement, and still drew a wrong conclusion because I sampled one point on a
curve I hadn't established was flat.

**Generalizable**: when reporting a perf delta for a change whose cost depends
on a structural parameter (depth, count, fan-out), measure at ≥3 values of that
parameter and state the *curve*, not one point. "I measured it" is not the same
as "I characterized it". This is the sharpest form of the session's recurring
lesson — the moment I believe I've done the rigorous thing is when scrutiny
drops.

### The review earned its cost twice over

The reviewer did not just flag the regression — it **built and verified the
cure** (filter to outermost roots; containment is transitive so the set is
provably equivalent) and reported 41/41 green with flat perf. That turned a
"known regression, ship anyway" into a fix, in one round trip. Worth continuing
to brief reviewers to propose-and-verify, not merely report.

It also caught an **undocumented behavior change** I had not noticed: the old
`querySelector('[dj-view]') || '[dj-root]'` ignored `[dj-root]` whenever any
`[dj-view]` existed, so a scoped attr inside `dj-root` but outside `dj-view`
was inert and now fires. Upgrade-visible; now stated in the CHANGELOG.

### Reproducer-first widened the fix again (3rd time this session)

I filed #2110 as a staleness bug. The reproducer showed a second root's scoped
attrs **never registered at all**. Same pattern as #2097 (narrowed) and #2108
(widened): the issue text — including issues I wrote myself, hours earlier —
was not a reliable description of the defect.

### A near-miss worth recording

My first instinct was the fix I had suggested *in the issue*: scope the sweep
down to the root. That would have **evicted a sibling root's entries outright**,
turning "stale" into "dead". Checking whether the obvious fix makes things worse
— before writing it — is what avoided shipping a regression dressed as a fix.

### Deferred, measured, not hidden

`isGoverned` is O(entries × roots), so many genuinely SIBLING roots degrade
(~+42% at 100). Nested roots now collapse to one entry and do not contribute.
Recorded in the CHANGELOG as a measured known characteristic rather than left
implicit — root counts that high are not reachable in practice.

RETRO_COMPLETE

Issue #2113, opened 2026-07-25
Patch for a detached [dj-virtual] item can land on the WRONG node via getNodeByPath path-fallback


johnrtipton:
Observed while writing the #2017 item 5 diagnostic (PR #2114).

## What happens

`getNodeByPath(path, djustId, rootEl)` resolves by `dj-id` first and **falls back to the positional path** when the id is not found. For an item held detached by a `[dj-virtual]` list, the id is genuinely absent from the document — so resolution falls through to the path, which usually still resolves to *some* node.

The patch then applies to **a different element than the server intended**, silently. `applyPatches` returns `true`; nothing warns.

## Reproduced (JSDOM, real `applyPatches`)

30 rows in a 40px `[dj-virtual]` window, so row `dj-id=129` is detached:

```js
djust.initVirtualLists();
djust._findVirtualListHolding('129');  // -> the container (confirms detached)

await djust.applyPatches([{type:'SetText', path:[0,0], d:'129', text:'updated'}]);
// -> returns TRUE, zero warnings; the text landed on the node at path [0,0]
```

With an unresolvable path (`[0,99]`) it correctly misses and warns — which is the case PR #2114's diagnostic covers. The resolvable-path case is the dangerous one because it is **silent and wrong**, rather than merely dropped.

## Severity

This is data corruption rather than a missed update: the wrong row shows the new content. It is narrow (requires a detached virtual item plus a path that still resolves), but it fails silently, which is the worst property for a rendering bug.

## Possible directions

- When a patch carries a `d` (dj-id) and that id is *known to exist but detached*, do not fall back to the path — treat it as a miss (PR #2114's `_findVirtualListHolding` already answers exactly this question).
- Or make path-fallback opt-in for patches that never carried an id.

Either likely wants to ride with the differ-awareness work (#2017 items 2–4), since that is where keyed positioning is decided.

## Related
- #2017 (deeper dj-virtual reconcile)
- PR #2114 (diagnostic for the *missed* case)
- #1988 / #1989 (the client-side self-heal that preceded this)

Issue #2112, opened 2026-07-25
StreamsMixin._stream_operations is accumulated but never consumed — sync stream ops never reach the client


johnrtipton:
Found while scoping #2017 item 1, which presumes stream-ops can be wired to `container.__djVirtualItems`. They cannot, because the op list has no consumer.

## Evidence

`StreamsMixin` (`python/djust/mixins/streams.py`) appends to `self._stream_operations` in `stream_reset`, `stream_insert`, `stream_prune`, `stream_delete`. The accessor that drains it is:

```python
def _get_stream_operations(self) -> list:
    """Get and clear pending stream operations."""
```

`grep -rn "_get_stream_operations" python/ tests/` returns **exactly one hit — its own definition.** No caller anywhere. `_stream_operations` itself is read only by `python/djust/testing.py:771`. There is no client-side applier either: nothing in `static/djust/src/` references `stream_operations`.

## Reproduced

```python
class V(LiveView):
    def mount(self, request, **kw):
        self.stream('msgs', [{'id': 1, 'text': 'a'}])

v = V(); v.mount(None)
v.stream_insert('msgs', {'id': 2, 'text': 'b'})

v._streams['msgs'].items   # [{'id':1,...}, {'id':2,...}]   <- data IS correct
v._stream_operations       # 2 ops queued, never drained    <- inert
```

So the stream **data** is maintained correctly; only the **operation list** is dead.

## Related observation

`_get_streams_context()` (same file, line 195) also has zero callers, and `'msgs' not in v.get_context_data()` — so streams do not auto-populate the template context either. Users must pass them through manually. Worth confirming whether that is intended.

## Note on MRO

`LiveView.__mro__` contains both `StreamsMixin` (sync) and `StreamingMixin` (async), which co-define `stream_insert`/`stream_delete` (see the comment at `live_view.py:186`). `StreamsMixin` wins, so `self.stream_insert(...)` is the sync, op-appending one.

## Why it matters

1. If the op list is meant to reach the client, it is a silent no-op today.
2. If it is vestigial, it should be removed — it currently reads as a working delivery mechanism, which is what made #2017 item 1 look implementable.

**#2017 item 1 is blocked on deciding which.**

## Related
- #2017 (dj-virtual deeper reconcile — item 1 depends on this)
- PR #2114 (#2017 item 5, the diagnostic that shipped)

johnrtipton:
## Correction + significantly expanded findings

[...]

PR #2114, opened 2026-07-25
feat(client): name dj-virtual when a patch misses an off-window item (#2017 item 5)

johnrtipton:
Implements **item 5 of #2017 only**. #2017 stays open — see Scope below.

## The problem

A `[dj-virtual]` list keeps off-window items **detached**, held only in
`state.items`. A server patch aimed at one legitimately resolves to `null`, and
the existing miss warning then lists generic causes:

```
- The DOM may have been modified by third-party JS
- A template {% if %} block may have changed the node count
- A conditional rendering path produced a different DOM structure
```

All three are wrong here. Following them is what made #1988/#1989 expensive to
investigate.

## The fix

`window.djust._findVirtualListHolding(djId)` re-discovers `[dj-virtual]`
containers from the DOM (`STATE` is a WeakMap and deliberately not iterable,
#2033) and returns the one holding a **detached** item with that dj-id. The
detached check is what makes it false-positive-free: an item still in the
document was resolvable, so the miss has a different cause.

The patcher calls it on the miss path and — **only on a positive
identification** — names the container and explains the patch cannot land until
the item scrolls back into the window.

### Deviation from the issue, flagged for review

#2017 asks for a `djustDebug`-gated warning. This is **unconditional** instead:
the helper answers only when it can *prove* a detached holder, so there is no
speculative branch to add noise, and someone hitting this in production is
exactly who needs it. Trivial to gate if you disagree.

## Scope — #2017 stays open

Investigation reshaped the issue:

| Item | Status |
|---|---|
| 5 — diagnostic warning | **shipped here** |
| 2, 3, 4 — differ awareness, out-of-window landing, keyed insert | **need an ADR.** `grep -rn "dj-virtual" crates/` confirms the differ has zero awareness; 3 and 4 both depend on 2. |
| 1 — stream-op wiring | **blocked.** It presumes stream-ops reach the client. They do not — filed as #2112. |

## Two findings filed from the investigation

- **#2112** — `_stream_operations` is accumulated by `stream_insert`/`stream_prune`/
  `stream_delete` and consumed by **nothing**. `_get_stream_operations()` has
  exactly one grep hit: its own definition. Verified empirically that the stream
  *data* is correct while the op list is inert.
- **#2113** — worse, and found while writing the test: `getNodeByPath` falls back
  to the positional path when a `dj-id` is absent. For a detached virtual item
  the id *is* absent, so the patch lands on **a different node**, silently, and
  `applyPatches` returns `true`. Data corruption rather than a dropped update.
  It is why my first test attempt passed when it should have failed — the test
  now uses an unresolvable path so it genuinely misses.

## Tests — 6 cases in `tests/js/dj-virtual-patch-miss-diagnostic-2017.test.js`

Helper exported · positive identification of a detached item · unknown id →
null · no-virtual-list page → null · end-to-end miss warning naming the
container · **no-false-positive guard** (an ordinary miss must not blame
dj-virtual).

**Gate-off verified per half** (#1468): disabling the patcher hint fails 1;
removing the helper export fails 5.

Full suites: **1870 JS**, **9908 Python**. Bundle **+285 bytes gz**.


johnrtipton (commented):
## Stage 11 — verdict: REQUEST_CHANGES (GitHub blocks self-approve; stated here)

Reproduced, not taken on faith. **Gate-off matrix matches the claim exactly** (rebuilt each time): patcher hint gated → 1 fail; helper export removed → 5 fail. **Artifacts**: `scripts/build-client.sh` reproduces committed `client.js`/`client.min.js` byte-identical. **Suites** (worktree): 1870 JS, 9908 py + 218 skipped. **Scope claims both true** — `grep -rn "dj-virtual" crates/` = 0; `_get_stream_operations` has one hit, its own def at `python/djust/mixins/streams.py:199`. No banned patterns; both logged values are `%s`-parameterized + `sanitizeIdForLog`.

🔴 **Cross-root false positive — the exact class this feature exists to prevent.** `src/29-virtual-list.js:705` scans `document.querySelectorAll('[dj-virtual]')`, but the miss it explains was resolved against a *scoped* root (`src/12-vdom-patch.js:1779` → `scope = rootEl || document`), and dj-id namespaces are per-view — the comment at `12-vdom-patch.js:1776` says so. Reproduced: a parent-root patch for `dj-id="129"` misses, and a sticky child's virtual list holding a detached `129` gets named. Thread `rootEl` through and scan `(rootEl || document)` — plus the root element itself (#2097/#2107 class).

🟡 **The `document.contains` guard — the whole "false-positive-free" claim — has no test.** Replacing `src/29-virtual-list.js:717` with an unconditional `return container;` leaves 6/6 green. It *is* load-bearing (probe: an attached in-window item then gets blamed), so untested rather than dead — Action #1859. The "no-false-positive" case uses a page with *no* virtual list, which `containers.length === 0` already covers. Needed: list present, item attached, miss from another cause.

🟡 **Unconditional O(misses × items).** jsdom, 10k rows: 200 all-missing patches = 606 ms vs a 17 ms no-list baseline → **589 ms helper overhead, 2.95 ms/miss**. Worst case (id not held → full scan, no early exit) is the *common* case. 1k/50 is fine (13 ms).

**Unconditional vs gated**: agree on the *message* — it rides an already-unconditional warn and fires only on proof. Disagree on the *work*: throttle or run it once per batch.

Nit: `sanitizeIdForLog` strips `.`, so `dj-virtual="chat.rows"` logs as `chatrows` and `'(no id)'` as `noid`.

johnrtipton:
## Stage 14 Retrospective — PR #2114 (#2017 item 5)

**Shipped**: a patch miss on an off-window `[dj-virtual]` item now names the
real cause instead of listing three wrong ones. 4 commits, CI green,
+373 bytes gz, 10 tests.

### The diagnostic could itself misdirect — the exact failure it prevents

Stage 11's 🔴: `dj-id` namespaces are **per-view** (stated in the comment at
`12-vdom-patch.js:1777`), but `findVirtualListHolding` scanned the whole
document. A parent-root miss on `dj-id=129` got blamed on a sticky child's
list holding a detached `129`. A tool built to stop misdirection, misdirecting.

**Lesson**: when adding a diagnostic, apply the same scoping rules the thing
being diagnosed uses. I read the resolution line (`scope = rootEl || document`)
while writing the call and still wrote a global scan two files away.

### My throttle throttled nothing, and only measuring caught it

I stamped the timestamp **on a hit**. The reviewer's finding was that
"id not found" is both the expensive case (full scan, no early exit) and the
common one — so every miss still paid full cost. Measured: 616 ms overhead,
completely unchanged by my "fix". Moving the stamp *before* the lookup took it
to −1 ms.

**Lesson**: a perf fix is not done when it is written, it is done when the
number moves. I would have shipped a no-op throttle on a code reading. Third
time this session that measurement contradicted a confident claim
(#2110 depth curve, #2110 nested cost, this).

### The property I advertised was the one property untested

I wrote that the `document.contains` check is "what makes it
false-positive-free". Removing it left 6/6 green — my only no-false-positive
test used a page with *no* virtual list, which a different branch already
short-circuits. The claim lived in prose, never in a test (#1859).

**Lesson**: whenever a PR description names a safety property, grep the tests
for the case that would fail without it. If it is not there, the property is an
assertion, not a guarantee.

### Scope discipline held under a goal to "complete all issues"

#2017 is a 4-workstream epic. Shipping item 5 and leaving the issue open —
with the CHANGELOG saying so — was the honest completion state. Items 2–4 need
an ADR (`grep -rn "dj-virtual" crates/` → 0 hits, independently confirmed by
the reviewer).

### Investigation found two bugs the issue never mentioned

**#2113** — `getNodeByPath` path-fallback lands a patch on the WRONG node when
the dj-id is absent because the item is detached. Silent data corruption, and
it is why my first test passed when it should have failed.
**#2112** — filed, then corrected **twice** in public. First framing (dead
code) and second (MRO shadowing is a bug) were both wrong; the `.pyi` declares
the shadowing as intended and `stream_to()` is the real async API. The actual
bug is small: `_get_streams_context()` has zero callers, so sync stream data
never reaches templates.

**Lesson**: the first read of an unfamiliar subsystem is confidently wrong.
Posting corrections rather than quietly editing keeps the trail honest.

RETRO_COMPLETE

PR #2115, opened 2026-07-25
fix(client): do not path-fallback a patch whose dj-id is held detached (#2113)

### LINKED ISSUE #2113: Patch for a detached [dj-virtual] item can land on the WRONG node via getNodeByPath path-fallback

johnrtipton:
Closes #2113.

## The bug

`getNodeByPath` resolves by `dj-id`, then falls back to the positional path when
the id is absent. That fallback is correct for its purpose — an id-less patch,
or an id that changed while the path stayed valid.

It is **wrong** when the id is absent *because* a `[dj-virtual]` list holds the
item detached off-window. The server addressed a **known** item; the fallback
silently retargets a different element. `applyPatches` returns `true`. Nothing
warns.

**This is worse than the miss that #2017 item 5 diagnoses.** That one is loud
and drops the update. This one is silent and applies it somewhere else.

## Severity is higher than I described when filing

The reproducer shows it is not confined to one row. A `SetText` that lands on a
container **wipes its children** — the snapshot of every `[dj-id]`'s text goes
from 4 rows to **zero**:

```
AssertionError: expected [] to deeply equal [ 'row 0', 'row 1', 'row 2', 'row 3' ]
```

## The fix

After the id lookup fails, ask the root-scoped `_findVirtualListHolding` (from
#2114) whether a virtual list holds that exact id **detached**. If so, return
`null` — a genuine miss — and #2017 item 5's diagnostic then explains why.

## What could have gone wrong, and what pins it

The obvious risk is over-suppressing: disable the fallback too broadly and you
remove the path resilience it exists to provide, on any page that merely
*contains* a virtual list. Three of the five tests exist for exactly that:

| Test | Pins |
|---|---|
| id-less patch still resolves by path | fallback intact for its original case |
| stale id, no virtual list → still falls back | ordinary id-drift resilience intact |
| virtual list present but NOT holding the id → still falls back | suppression is narrow, not blanket |

**Gate-off verified** (#1468): gating the guard fails exactly the 2 bug cases
and leaves all 3 fallback-preservation cases green.

## Verification

- 5 cases in `tests/js/dj-virtual-path-fallback-2113.test.js`.
- Sibling suites green — **27** across #2017 / #2097 / #2108 / #2110, all of
  which touch this same file.
- Full suites: **1879 JS**, **9908 Python**.


johnrtipton (commented):
## Stage 11 — verdict: **REQUEST_CHANGES**

Worktree-isolated. Suites green: **1879 JS**, **9908 Python** (218 skipped). Artifacts byte-identical after `scripts/build-client.sh` (all 4 SHA-256 match); +19 B gz. Gate-off reproduced exactly as claimed — gating `12-vdom-patch.js:239` fails exactly the 2 bug cases (container-wipe shows `[]` vs 4 rows), 3 preservation cases stay green. Ordering is sound: pre-init and helper-deleted both fall back correctly.

Two findings, one remedy.

🔴 **Over-suppression: a legitimate fallback aimed OUTSIDE the list is now dropped.** The guard asks only "is this id held detached *anywhere in scope*" — never whether the positional fallback would land near the holder. A/B vs main, one root, list holds `129` detached, patch `{d:'129', path:[1]}` targeting an unrelated live sibling:

- main → `sib = 'FELLBACK'` (applied)
- #2115 → `sib = 'sibling'` (**dropped** — and #2114's diagnostic misattributes it to dj-virtual)

Structurally likely, not exotic: dj-ids are "purely positional", re-assigned pre-order (`crates/djust_vdom/src/lib.rs:206`), so a large list retains a contiguous block of *previous-generation* ids that drifted ids collide with. The third preservation test uses `unknown-id` — it never covers "list holds the id, patch targets elsewhere" (rc4 retro rule 1: enumerate every variant).

🔴 **Perf: 21.7× on the stale-id path.** 50 stale-id patches over a 10k-item list — main **7 ms**, #2115 **152 ms** (3.04 ms/patch). That matches #2114's own "~3ms per miss at 10k items", which is *why* `12-vdom-patch.js:1846-1858` throttles. Line 239 has no throttle, and the two never share budget: a *successful* fallback never reaches 1849, so `_lastVirtualDiagnosisAt` is never stamped. Control: valid-id patches 4 ms on both — the hit is confined to id-lookup failure, routine during rapid re-render.

**Remedy (verified):** resolve the positional fallback first, then consult the helper only when that node is inside a `[dj-virtual]` (`node.closest('[dj-virtual]')`). Measured 150 ms → **6 ms**; #2113 still fixed with the correct diagnostic; unrelated fallback restored; all 3 preservation cases green.

Nit (non-blocking): null returns are handled safely at `12-vdom-patch.js:1505/1580/2366` — no crash risk.

johnrtipton:
## Both findings addressed in `0f46fded`

**🔴 over-suppression — you were right, and my framing was wrong in two directions.** I checked the dj-id alone and never where the fallback would land, so I traded a silent wrong-node bug for a silent dropped-patch bug in a different case. Your probe reproduced exactly: held id + fallback onto an unrelated live sibling → main applies, my branch dropped it. The `crates/djust_vdom/src/lib.rs` note that dj-ids are purely positional is what makes it routine rather than exotic — a windowed list keeps a contiguous block of ids for drifted ids to collide with.

**🟡 perf 21.7× — confirmed, and your diagnosis of why the #2114 throttle doesn't help was exact.** A successful fallback returns before reaching that call site, so `_lastVirtualDiagnosisAt` is never stamped and the two never share budget.

**Applied your verified remedy**: resolve the positional fallback first, suppress only when the resolved node is INSIDE a `[dj-virtual]`. Re-measured A/B against main's bundle on an identical page:

| | main | branch |
|---|---|---|
| 50 stale-id patches, 10k list | 11 ms | **9 ms** (was 152 ms) |
| held id + unrelated sibling | `FELLBACK` | **`FELLBACK`** |

Added a test for the exact over-suppression case. Worth noting it failed twice for the wrong reason first — I string-replaced `'</div></div>'`, which matched inside the rows and mangled the DOM, then guessed the wrong path index. Built the DOM explicitly instead of guessing again.

Tests 5 → 6; gate-off still fails exactly the 2 bug cases and leaves all 4 preservation cases green. Full suites 1880 JS / 9908 Python.

Thank you for building and measuring the remedy rather than only flagging the cost — that turned two findings into one commit.

johnrtipton:
## Stage 14 retro — PR #2115 (#2113)

**Backfilled** by `/pipeline-retro` for #2140. This is a code PR, so the minimal Dependabot form does not apply.

**Shipped**: `getNodeByPath` stops path-falling-back a patch whose `dj-id` is held **detached** by a `[dj-virtual]` list, and names the holding list in the miss diagnostic.

### What the review caught that the first pass had wrong

Two things, both measured rather than argued:

- **Over-suppression.** The first guard dropped legitimate patches — it fired on any id-lookup failure rather than only when the id is provably held detached by a virtual list. The distinction is what makes it false-positive-free: an item still in the document was resolvable, so the miss has another cause.
- **21.7× on the miss path.** The pool walk ran on *every* id-lookup failure, and the expensive case (id not held → full scan of `state.items`, no early exit) is also the common one. Throttling had to be stamped **before** the lookup rather than on a hit.

### The lesson

A diagnostic that runs on a failure path is on a **hot** path, because failures come in bursts — a burst of 200 misses against a 10k-item list cost 616 ms unthrottled. Treat "this only runs when something went wrong" as a reason to measure, not a reason to skip measuring.

`RETRO_COMPLETE`

Issue #2116, opened 2026-07-25
Stream.delete cannot delete dict items — identity match uses getattr(item,'id',...) which never works for dicts


johnrtipton:
Found while testing #2112 (PR for the context wiring).

## The bug

`Stream.delete` (`python/djust/session_utils.py:205`) resolves an item's identity as:

```python
self.items = [
    item for item in self.items
    if getattr(item, "id", getattr(item, "pk", id(item))) != item_id
]
```

`getattr` reads an **attribute**, not a dict key. A dict item `{"id": 1, "text": "a"}` has no `.id` attribute, so it falls through to `id(item)` — the CPython object address — which never equals the caller's id. **Dict items are never removed.**

Dicts are the natural item shape here: `stream_insert(name, item)` accepts anything, and the guides' examples pass model instances, but nothing signals that dicts silently won't delete.

## Reproduced

```python
v.stream("messages", [{"id": 1, "text": "a"}, {"id": 2, "text": "b"}])
v.stream_delete("messages", 1)
# _streams["messages"].items still contains BOTH items
```

With an object exposing `.id` it works correctly, which is what PR for #2112 uses in its delete test.

## Second, smaller issue: the parameter name lies

The signature is `stream_delete(self, name: str, item_or_id: Any)`, implying you may pass the item. Passing a dict item raises:

```
TypeError: unhashable type: 'dict'
```

from `self._deleted_ids.add(item_id)` (`session_utils.py:214`), because the dict itself becomes the "id".

## Suggested fix

Resolve identity with a helper that handles mappings as well as attributes:

```python
def _item_id(item):
    if isinstance(item, Mapping):
        for key in ("id", "pk"):
            if key in item:
                return item[key]
        return id(item)
    return getattr(item, "id", getattr(item, "pk", id(item)))
```

Use it in BOTH the `item_or_id` normalization and the list comprehension, so passing either the item or its id works for dicts and objects alike. Needs a regression test covering dict items, model-like objects, and passing the item vs the bare id.

## Related
- #2112 — sync stream context wiring (where this surfaced)
- PR for #2112 uses an object with `.id` in its delete test specifically to avoid this

Issue #1039, opened 2026-04-25
tech-debt: Mutation-after-capture test discipline for any snapshot/capture function


johnrtipton:
**Source**: Retro v0.6.1 / PR #976 (+ latent v0.6.0 bug)
**Action Tracker**: row #115

**Action**: Mutation-after-capture test discipline for any snapshot/capture function

**Notes**: `_capture_snapshot_state` reference-aliasing bug existed unnoticed for two milestones (v0.6.0 `enable_state_snapshot` + v0.6.1 time-travel). Generalize: every capture function needs a test exercising mutation after capture.

---
_Filed by `/pipeline-retro --reconcile` on 2026-04-25 as part of Scope C: file all Action Tracker rows as GitHub issues._

johnrtipton:
Shipped in PR #1226 (v0.9.1-7 cleanup batch). Comma-list `Closes #...` form parsed only the first ref — the rest closed manually. (This bites again; #1184's parenthesized form is documented in PR-checklist, but the comma-list-on-its-own-line form has the same auto-close failure.) 🤖 v0.9.1-7 cleanup batch closure.

PR #2117, opened 2026-07-25
fix(streams): expose sync stream data to the template context (#2112)

### LINKED ISSUE #2112: StreamsMixin._stream_operations is accumulated but never consumed — sync stream ops never reach the client

johnrtipton:
Closes #2112.

## The bug

`self.stream("messages", …)` fills `_streams["messages"].items`, and
`_get_streams_context()` exists to expose it — its docstring literally reads
*"Get streams data for template context"*.

**Nothing ever called it.** So the pattern documented in
`docs/website/guides/large-lists.md` stored the data and rendered nothing:

```python
self.stream("messages", Message.recent(50), limit=50)
```
```html
{% for m in messages %}   <!-- undefined name -->
```

## Why it survived this long

`assert_stream_insert` inspects the **queued op list**, which is populated
whether or not the data is ever reachable. A user's test went green while the
page stayed empty.

The feature had assertions — just not ones that could fail. One of the new tests
pins that reasoning so the next reader sees why a passing `assert_stream_insert`
is not evidence of rendering.

## Precedence: existing key wins (deliberately)

The action-state block immediately above this one *overwrites* on collision.
Streams do the **opposite**, and the reason is compatibility: because streams
never reached the context before, an app that has both `self.messages` and a
stream named `messages` renders the attribute **today**. Clobbering would be a
silent breaking change. A stream only fills a name nothing else claimed.

## Tests — 10 cases in `tests/unit/test_streams_context_2112.py`

Context reachability · later inserts reflected · deletes reflected · multiple
streams · name→key mapping (parametrized) · **two guard rails** (explicit user
context wins; a view with no streams is unchanged) · a pin on the tautology
above.

**Gate-off verified** (#1468): 8 of 10 fail without the wiring.

**The full Python suite is the regression check here** — `get_context_data` is
on every view's path. **9918 passed** (was 9908), 0 failures.

## Filed, not widened

**#2116** — `Stream.delete` cannot delete **dict** items: it resolves identity
with `getattr(item, "id", getattr(item, "pk", id(item)))`, and a dict has no
`.id` *attribute*, so it falls through to `id(item)` and never matches. Passing
the dict itself instead raises `TypeError: unhashable type: 'dict'` despite the
parameter being named `item_or_id`. The delete test here uses an object with
`.id` specifically to avoid that, with a comment saying so.


johnrtipton (commented):
Stage 11 — **verdict: REQUEST CHANGES** (own PR; GitHub refuses `--approve`).

**Independently confirmed.** Gate-off: 9/11 fail (`if False and callable(get_streams)`), matching the claim. The precedence guard is load-bearing — flipping it to clobber fails `test_existing_streams_key_is_not_clobbered`. `streams.<name>` is the **sole** contract: `docs/website/guides/large-lists.md:89,159`, `session_utils.py:186`, `mixins/streams.py:29` ("used in template as streams.{name}"); nothing anywhere expects the top-level form, so the self-correction was right. The tautology claim is accurate — `testing.py:771` reads only `_stream_operations`. Data really renders: a real Rust-engine render of `{% for msg in streams.messages %}` emits `<li id="m-1">alpha</li><li id="m-2">beta</li>`. Suites: **9918 passed, 219 skipped**; vitest **1880 passed**. No banned patterns.

🔴 **Self-poisoning loop — the fix shadows itself after any state restore.** `context["streams"]` is copied into `_cached_context` (`mixins/request.py:228`) and persisted to `request.session["liveview_<path>"]` (`:250-255`). Both restore paths then `safe_setattr(view, "streams", <stale dict>)` — `mixins/request.py:614-617` (HTTP-fallback POST, **ungated**) and `runtime.py:2114-2116` (WS reconnect). `self.streams` is now a public attribute, so the attribute walk puts it in `context`, and the new guard at `mixins/context.py:359` (`"streams" not in context`) skips the live data permanently. Empirical: after restore, live `_streams["rows"]` = `["first","SECOND"]`, rendered `<ul dj-stream="rows"><li id="r-1">first</li></ul>`. Every post-restore insert is invisible — the exact bug class this PR fixes, one hop downstream.

🟡 **Session bloat.** The documented 500-item case grows the session entry from **503 → 44,916 bytes** per GET. Streams exist to keep large collections *out* of state.

🟡 **Injected after the serialization pass** (`context.py:357` vs the JIT/`_deep_serialize_dict` walk ending ~`:320`), so stream items skip it: a `User` stream item persisted **17 fields** (email, last_login, is_active) vs 4 for the same model as an ordinary attribute. The `_ALWAYS_EXCLUDED_FIELDS` floor still held — no password hash.

Suggested fix: exclude `streams` from the session/state snapshot, or gate on `"streams" not in self.__dict__` so a restored copy cannot shadow live data.

johnrtipton:
## 🔴 and both 🟡 addressed in `17a0c26d`

Your self-poisoning trace was exact, and it showed my precedence rule was the thing making it permanent. Two changes, because the loop had two links:

**1. Root cure — `streams` is now excluded from the session snapshot** (`mixins/request.py`). It is DERIVED from `_streams` on every `get_context_data()` call, not user state, so persisting it was wrong independent of the loop. That also closes your other two findings at the source: the ~45 KB per-GET bloat (streams exist precisely to keep large collections *out* of state) and the un-serialized persistence of stream items — nothing is persisted, so neither arises.

**2. Precedence reversed — live streams now win.** My existing-key-wins rule looked like the safe non-breaking choice; it is exactly what made the shadowing permanent, since a session written by an older release can still carry a stale attribute. A view with NO streams is still left completely alone, so an app using `self.streams` for its own purposes and never calling `stream()` is unaffected.

I wrote the stale-attribute test expecting it to pass with the snapshot exclusion alone. It failed — which is what forced the precedence change rather than my reasoning about it.

Tests 11 → 14. Full Python suite 9922 passed.

**On your two side-observations**: `_get_streams_context()` having zero callers is precisely what this PR fixes — good independent corroboration. And thank you for the `_reset_streams` note (the in-place `.clear()` aliasing the tracked value, so change-detection misses it, Action #1039 class) — it works today by accident. I have left it alone as out of scope rather than fold it in, but it deserves its own issue and I would rather you flagged it than not.

johnrtipton:
## Stage 14 retro — PR #2117 (#2112)

**Backfilled** by `/pipeline-retro` for #2140. Code PR, so the minimal form does not apply.

**Shipped**: sync stream data reaches the template context, so `{% for x in streams.name %}` renders on the HTTP path.

### Two self-caught defects, both design errors rather than typos

- **A self-poisoning loop.** My first precedence rule was "existing key wins", which meant a stale `streams` restored from the session shadowed live data **permanently** — every subsequent render re-saved the stale copy. Fixed by excluding `streams` from the session snapshot *and* reversing the precedence. Either alone would have left the loop reachable.
- **The wrong context shape.** I injected a top-level `messages`; the documented shape is `streams.messages`. Caught by re-reading the docs the feature is supposed to satisfy, which is the cheapest check available and the one I did last rather than first.

### The lesson

For a feature whose contract is *documented*, read the doc before the code, not after. Both defects here were disagreements with a written contract, not with the runtime — and the runtime was happy to accept both.

`RETRO_COMPLETE`

PR #2118, opened 2026-07-25
fix(streams): Stream.delete now matches dict items (#2116)

### LINKED ISSUE #2116: Stream.delete cannot delete dict items — identity match uses getattr(item,'id',...) which never works for dicts

johnrtipton:
Closes #2116. Found while testing #2112.

## The bug

`Stream.delete` resolved identity with:

```python
getattr(item, "id", getattr(item, "pk", id(item)))
```

`getattr` reads an **attribute**. A dict item `{"id": 1, "t": "a"}` has no
`.id` attribute, so it fell through to `id(item)` — the CPython object address
— which never equals a caller's id. **Dict items were silently undeletable.**

Passing the dict itself instead raised:

```
TypeError: unhashable type: 'dict'
```

from `_deleted_ids.add(item_id)` — despite the parameter being named
`item_or_id`.

Dicts are the natural shape here: `stream_insert` accepts anything, and nothing
signalled that dicts wouldn't delete.

## The fix

Extracted `Stream._identity()`, which handles mappings and objects alike, and
applied it to **both sides** of the comparison — the previous code used
`getattr` on the list items but a different resolution for the argument, which
is how the two drifted.

`delete()` now treats a `Mapping` as an **item**, never a bare id.

Three details worth review:

- **Key presence decides, not truthiness.** `{"id": 0}` and `{"id": None}`
  resolve to their real ids rather than the object address. Two of the gate-off
  failures were exactly these.
- **An id-less dict resolves to its own address**, so a look-alike dict matches
  nothing rather than deleting the wrong row.
- **The tombstone add is guarded** for unhashable ids: removal from `items`
  still works, and `_deleted_ids` is only an optimisation for the client diff.

## Tests — 12 cases in `tests/unit/test_stream_delete_dict_2116.py`

dict by id · dict by passing the item · dict by `pk` key · falsy ids
(parametrized `None`/`0`/`""`) · look-alike must not match · unknown id removes
nothing · tombstone recorded · **plus three pins that object `.id`, `.pk`, and
by-object deletion still work**, since the risk here is regressing the shape
that already worked.

**Gate-off verified** (#1468): 6 of 12 fail without the mapping branch.

Full Python suite: **9920 passed**, 0 failures.


johnrtipton (commented):
## Stage 11 review — REQUEST CHANGES (small, specific ask; the core fix is correct)

**Verified good.** Gate-off reproduces your claim exactly: disabling the `Mapping` branch in `_identity` gives **6 failed / 6 passed** — the 3 dict cases + the 3 falsy-id cases. Full suite `tests/ python/tests/ python/djust/tests/ -n auto`: **9919 passed, 219 skipped, 0 failed**. ruff check + format clean.

**id-vs-pk ordering — no regression.** `_identity`'s object branch (`python/djust/session_utils.py:222`) is character-identical to the old expression, and `delete()`'s non-Mapping arm keeps the old `hasattr` chain, so `.id` still wins over `.pk` on *both* sides. Probed a Django-shaped `id=5, pk=7`: old→5, new→5; pk-only→3 both. The Mapping branch uses the same `("id","pk")` order, so argument and item resolution agree.

**Falsy ids**: `None/0/""/False` all match on both sides; `{"id":1,"pk":2}` → id wins, pk ignored — consistent with objects. **Id-less dict**: same object deletes, look-alike does not. Address reuse can't collide while an item is live — `self.items` holds a strong ref.

🔴 **The drift is cured at 2 of 3 sites.** The *public* entry point users actually call, `StreamsMixin.stream_delete` (`python/djust/mixins/streams.py:164-169`), still carries the identical pre-fix `hasattr` chain. For the very input `test_delete_dict_item_by_passing_the_item` blesses, it emits:

```
dom_id = "things-{'id': 1, 't': 'a'}"   # the dict's repr
```

And `default_dom_id` (`python/djust/mixins/streams.py:45`) uses `or` — **truthiness**, the exact bug you fixed in `_identity` — so an item with `id=0` inserts as `things-<address>` while delete emits `things-0`. Please route both through `Stream._identity` (#1646: cure the class, not 2/3).

🟡 **The `TypeError` guard is untested.** Gate-off: deleting it leaves **12/12 green**. It's the only new branch with zero coverage. It does work — probed directly: no raise, and `items` removal still succeeds.

🟡 **`_deleted_ids` has zero readers repo-wide** (py/rs/js). The guard comment and CHANGELOG call it "an optimisation for the client diff"; nothing consumes it, so that claim is false (#1867). Upside: the guard is risk-free.

Nit: branch is 1 commit behind `main`, and `merge=union` is local-only — refresh before merge.

johnrtipton (commented):
**Stage 13 re-review — REQUEST_CHANGES** (narrowly — the three prior findings *are* genuinely fixed).

Verified against an `origin/main` baseline + per-fix gate-off, not taken on trust. Suite: **9937 passed, 219 skipped, 0 failed**.

**Prior findings — all fixed, all load-bearing (#1468):**
- 🔴 `stream_delete` dict-repr — reverting `mixins/streams.py:173` reproduces `things-{'id': 1, 't': 'a'}` and reds `test_public_stream_delete_emits_the_same_dom_id_as_insert`. ✅
- 🔴 `default_dom_id` `or`-truthiness — restoring the `or` chain reds 3 tests incl. the `id=0` case. ✅
- 🟡 untested `try/except` — removing it now reds `test_unhashable_id_still_removes_from_items`. ✅
- 🟡 `_deleted_ids` — the corrected claim is true: zero readers in `python/`, `crates/`, `tests/`, client JS (only the PR's own assert, `test_stream_delete_dict_2116.py:95`).

**Helper split:** all 5 call sites correct — `_identity` only on items (insert fn, items filter), `resolve_id` only at the two `item_or_id` entry points.

**New, blocking:**

**1. Regression — `default_dom_id` now collides.** Presence-not-truthiness gives every item whose `id`/`pk` is present-but-`None` (or `""`) the *same* dom_id. `main`: unique addresses. Branch: 3 unsaved drafts → `['drafts-None','drafts-None','drafts-None']`, i.e. three DOM rows sharing `id="drafts-None"`. `test_falsy_ids_are_handled` uses one such item, so it can't see this (#1543). Suggested: skip a key whose value is `None` (Django's own `pk is None` = unsaved) — `id=0` still works.

**2. Two false claims on new code (#1867)** — the class the last review bounced:
- `mixins/streams.py:167` *"Same resolution as insert"* — false with the documented `dom_id=` callable: insert uses `stream_obj.dom_id_fn`, delete never consults it → `rows-hello-world` vs `rows-1`. A live #1646 twin; "cure reached ALL THREE sites" overstates.
- `session_utils.py:252` *"no id/pk resolves to its object address… matches only that exact object"* — Mappings only. An id-less **object** resolves to *itself*, so `delete(that_same_object)` removes nothing and delete emits `things-<Bare>` — the prior 🔴's shape, for objects.

Both behaviors pre-date the PR; only the claims are new. Reword + follow-up issue, or fix.

johnrtipton:
## Stage 14 Retrospective — PR #2118 (#2116)

**Shipped**: `Stream.delete` now matches dict items; identity resolution unified
across all default-factory sites. 4 commits, 13/13 CI, 18 tests.

### I walked into #1646 while citing #1646

The first pass fixed `Stream.delete` and left `StreamsMixin.stream_delete` — the
PUBLIC entry point — on the old `hasattr` chain, so a dict item emitted a dict
*repr* as its dom_id while insert used the address. `default_dom_id` also still
used `or` (truthiness), so `id=0` disagreed. Two of three sites.

I have cited the parallel-path rule repeatedly this session and still cured only
the site I was looking at. **The lesson is not "remember #1646" — I did remember
it. It is: after fixing an identity/format/filter function, grep for every OTHER
place that computes the same thing, before writing the test.**

### Fixing the drift introduced a new drift, then a new regression

Routing everything through one helper exposed that `_identity(0)` returns the
*address* of the int `0` — right for an ITEM, wrong for a bare-id ARGUMENT.
Split into `_identity` / `resolve_id`.

Then re-review found that my presence-not-truthiness rule collided every
`id=None` item onto one dom_id — three unsaved drafts all rendering as
`drafts-None`, where the old code gave unique addresses. **My test could not
have caught it**: it used a SINGLE falsy-id item, and the collision needs two or
more (#1543).

### Two false claims on my own new code

`streams.py` said "Same resolution as insert" — false under a custom `dom_id=`
callable, so "all three sites" overstated (filed #2121). `session_utils.py` said
an id-less item "matches only that exact object" — true for Mappings only. Both
#1867: assertions I wrote without checking, in the very commit that fixed
someone else's false claim.

### What actually worked

Every one of these was caught by an independent reviewer or by gate-off — none
by me re-reading my own diff. The re-review specifically earned its keep: the
prior findings were genuinely fixed, and it still found a fresh regression in
the fix.

RETRO_COMPLETE

PR #2120, opened 2026-07-25
feat(streams): route stream ops through the dj-virtual item pool (#2017 item 1)

johnrtipton:
Implements **item 1 of #2017**. #2017 stays open for items 2–4.

## The bug

A `[dj-virtual]` container's children are only the visible **window**; the full
collection lives in `state.items`, with off-window rows detached.
`17-streaming.js` had **zero** `dj-virtual` awareness and mutated the element
directly, so the pairing documented in `large-lists.md` — stream the data,
virtualize the rendering — did not work without bespoke app JS.

## Honest scope — established by gate-off, not assumed

**Append already worked.** The loose-child absorb (#1989 symptom 2) pulls an
appended node into the pool at the tail, which is correct for an append.

My first two tests counted pool **size** and passed with the routing disabled —
tautological. Rewritten to assert **order**, which exposed what is genuinely
broken:

| Op | Before | Why |
|---|---|---|
| append | ✅ worked | absorb appends at the tail — correct |
| **prepend** | ❌ wrong position | absorb *always* appends at the tail (the append-only assumption, #2017 item 4) |
| **prune** | ❌ wrong rows | trims `.children`, i.e. the visible window, not the pool |

The append test is kept and **labelled as characterization** — it stays green
with the routing gated off, and that is correct.

**Gate-off** (#1468): 2 of 5 fail without the routing. Append and the
non-virtual guard correctly stay green.

## Correcting an earlier claim of mine

I previously said item 1 was **blocked on #2112**. That was about the SYNC op
path (`_stream_operations`, which has no consumer). The live ASYNC path —
`stream_to` → `_send_stream_ops` → `handleStreamMessage` — is the one that
matters and was never blocked.

## What changed

`append`/`prepend` splice into the pool via a new `_virtualInsert`; `prune`
trims via `_virtualPrune`. Each **falls back to the original direct-DOM path**
when the target is not a virtual list — the vast majority of stream targets —
pinned by a regression test.

## Verification

5 cases in `tests/js/stream-ops-dj-virtual-2017.test.js`. Full suites:
**1885 JS**, **9922 Python**. Bundle **+331 bytes gz**.

## Still open on #2017

Items 2–4 (Rust differ `dj-virtual` awareness, out-of-window patch landing,
keyed insert position) need an ADR — `grep -rn "dj-virtual" crates/` is still
zero, and 3 and 4 both depend on 2.


johnrtipton:
## Stage 14 Retrospective — PR #2120 (#2017 item 1)

**Shipped**: stream ops targeting a `[dj-virtual]` container now route through
its item pool. 2 commits, 13/13 CI, +331 bytes gz.

### Gate-off caught two tautological tests before they shipped

My first two tests counted pool **size** and both passed with the routing
disabled. The reason: the pre-existing loose-child absorb (#1989 symptom 2)
already pulls an appended node into the pool. Size could not distinguish my
change from behavior that already worked.

Rewriting them to assert **order** exposed what is actually broken — **prepend**
(absorb always appends at the tail; that is #2017 item 4) and **prune** (trims
`.children`, i.e. the visible window, not the pool). Append already worked, so
that test is kept and **labelled characterization** rather than left implying it
proves the change.

**Generalizable**: when a subsystem already has a self-heal for the adjacent
case, a test asserting *quantity* cannot separate "my fix worked" from "the
self-heal worked." Assert the property the self-heal gets WRONG — here, position.

This is the second instance today of the same blind spot: earlier I measured a
perf delta at one nesting depth and generalized a curve. Both are "exercised the
mechanism, not the condition under which it fails."

### I had declared this item blocked, incorrectly

I earlier wrote that item 1 was blocked on #2112. That was about the SYNC op
path (`_stream_operations`, no consumer). The live ASYNC path — `stream_to` →
`_send_stream_ops` → `handleStreamMessage` — was never blocked. Investigating
one path and generalizing to "the feature" is the same error shape as the two
above.

### Scope held

#2017 stays open. Items 2–4 need the Rust differ to understand `dj-virtual`
(`grep -rn "dj-virtual" crates/` → still zero) and 3–4 depend on 2 — an ADR, not
a PR. The CHANGELOG and PR body both say so, so the issue does not read as done.

RETRO_COMPLETE

PR #2122, opened 2026-07-25
fix(streams): stream context captures must not alias the live list (#2119)

### LINKED ISSUE #2119: _reset_streams() clears streams in place, so change-detection misses it (works today by accident)
Surfaced by the Stage 11 review of PR #2117 (#2112). **Pre-existing**, not introduced there, and currently harmless — but harmless by accident, which is worth fixing before something depends on it.

## What happens

`_reset_streams()` (`python/djust/mixins/streams.py`) frees stream memory after render:

```python
for stream_obj in self._streams.values():
    stream_obj.clear()
```

`Stream.clear()` mutates `self.items` **in place** (`items.clear()`). The change-tracker compares against a value that is the *same list object*, so the mutation is invisible to it — the tracked "before" and "after" are the same object.

Observed by the reviewer: after the reset, `get_context_data()` returns `{'messages': []}`, yet render 2 still shows the rows. The render is correct for the wrong reason.

This is the Action #1039 class — **mutation-after-capture / reference aliasing**: a capture that shares a reference with its source cannot detect changes to it.

## Why it is latent rather than live

`_reset_streams()` only fires when `temporary_assigns` is set (early return at `live_view.py:1155`), so most views never reach it. And the accidental behavior happens to match the intended one today.

## Why fix it

The correctness depends on two bugs cancelling. Any of these breaks it:
- change-detection starts deep-copying or comparing by value
- `Stream.clear()` is changed to rebind (`self.items = []`) rather than mutate
- a caller starts reading the context between reset and render

## Suggested fix

Make the reset observable: rebind rather than mutate, and pin it with a test that captures context BEFORE the reset, mutates, and asserts the captured value is unchanged — the standard shape for #1039.

Also worth confirming what the *intended* post-reset render is. If streams should still render after a temporary-assigns reset, the reset may be wrong to run at all on that path.

## Related
- #2112 / PR #2117 — stream context wiring (where this surfaced)
- Action #1039 — mutation-after-capture test discipline

johnrtipton:
Closes #2119. Surfaced by the Stage 11 review of PR #2117.

## The bug

`_get_streams_context()` returned `stream_obj.items` — the **same list object**
the stream keeps mutating. `_reset_streams()` then calls `Stream.clear()`, an
in-place `items.clear()`, so anything already holding the context sees its data
vanish underneath it.

Change-detection could not see the reset either: its "before" and "after" were
the same object.

**The render stayed correct only because two behaviours cancelled** — Action
#1039's mutation-after-capture class. Correct by accident is the part worth
fixing; any of these breaks it:

- change-detection starts comparing by value
- `Stream.clear()` is changed to rebind rather than mutate
- a caller reads the context between reset and render

## Verified before fixing

```
captured before reset: 2
captured AFTER reset : 0        <- mutated under the capture
aliased: True                   <- captured IS stream.items
```

## The fix

Return a **copy** of each item list. Shallow on purpose: the items themselves
are the caller's own objects and stay shared; only the list container needs to
be independent.

## Tests — 5 cases, in the #1039 shape

Mutate the source AFTER capturing, assert the capture is unchanged:

- survives a reset
- survives a later insert, while a **fresh** read does see it
- identity is distinct from the live list
- two captures are independent of each other
- **the reset still actually clears the live stream** — the fix must not turn
  the reset into a no-op

**Gate-off** (#1468): 4 of 5 fail without the copy.

Full Python suite: **9927 passed** — `_get_streams_context` is on every view's
context path, so the whole suite is the regression check.


johnrtipton:
## Stage 14 Retrospective — PR #2122 (#2119)

**Shipped**: stream context captures no longer alias the live item list.
2 commits, 13/13 CI after a rerun, 5 tests.

### Correct-by-accident was the whole bug

The render was right, and would have stayed right, because two behaviours
cancelled: the context aliased `stream_obj.items`, and `_reset_streams()`
mutated that same list in place — so change-detection compared an object with
itself and saw nothing. Any ONE of these breaks it: change-detection comparing
by value, `Stream.clear()` rebinding instead of mutating, or a caller reading
the context between reset and render.

Verified before fixing rather than reasoned about: capture 2 items, reset,
capture is 0, and `captured is stream.items` was True.

### The CI red was not mine, and checking mattered

`python-tests (py3.12)` failed on a tick-loop test with no relationship to
streams. The tempting move is "obviously unrelated, merge." Instead I ran it on
`main` with no changes: **2 of 5 failed**, with every failing run at ~4.7s and
every passing run at ~1.7s — the failure tracks the run being slow, which is a
wall-clock race. Filed as #2124 with that evidence and a fix direction, then
re-ran CI and merged on a genuine green rather than merging red.

**The cost of not doing this** is worse than the re-run: a test that is red ~40%
of the time teaches reviewers to read a red `python-tests` as noise.

### Pattern across this drain

This is the fourth issue in the chain that came from a reviewer's side-finding
rather than a user report (#2113 → #2112 → #2116 → #2119), and the third whose
reproducer showed the problem was wider or differently-shaped than the issue
text claimed. Adversarial review kept producing more real work than it closed —
which is the point, but worth stating rather than treating each as a surprise.

RETRO_COMPLETE

PR #2123, opened 2026-07-25
docs(adr): ADR-026 — dj-virtual differ awareness (#2017 items 2-4)

johnrtipton:
Design deliverable for **#2017 items 2–4**. Items 1 and 5 shipped in #2120 and
#2114; these three needed a decision before any code, so this ADR is the work.

## The problem, verified not assumed

`grep -rn "dj-virtual" crates/` → **zero**. The Rust differ has no concept of a
client-windowed subtree, so:

| | server's model | client's DOM |
|---|---|---|
| 10 000-row feed | 10 000 nodes | ~20 nodes |

Hence the three items: off-window patches cannot land (item 3), positional
paths mean different things on each side (item 2), insert position is lost
(item 4).

## What it decides

**Recommends Option A** — keyed splice ops for `[dj-virtual]` subtrees — as the
only option that makes a mid-window update *correct* rather than merely safe,
and notes the machinery mostly exists (`diff.rs` already does keyed
reconciliation with an LIS pass in `lis.rs`).

**Status: Proposed. It deliberately does NOT authorise implementation.**

Three reasons to defer, in the ADR:

1. **No confirmed demand.** The reported pain (#1988/#1989, snake-arena, the
   #1724 chart case) was append-only feeds and teardown — all covered by what
   shipped: the #2017 item 5 diagnostic, #2113's no-wrong-node guard, and
   #2017 item 1's pool routing.
2. **High blast radius** across differ + wire protocol + client applier at
   once. Per #1122 that wants a soak behind the four PRs that touched
   `12-vdom-patch.js` in the last day.
3. **The cheap half is done**; what remains is the expensive half.

Options B (client-side patch buffering) and C (server-side windowing) are
rejected with reasons — B converts a correctness problem into a memory and
ordering problem and does not fix item 4 at all; C makes render stateful per
connection.

## Also recorded

The sequencing this would need if taken (flag-gated differ change, wire-format
pinning per #1448 with the serde field-position check from #1541, client
reusing `_virtualInsert`/`_virtualPrune` rather than adding a second path per
#1646), and the specific testing trap this area has already produced twice:
**assert order and position, not counts** — #2017 item 1's first tests counted
pool size and passed with the fix disabled, because the #1989 absorb produced
the right count for the wrong reason.

If Option A is never taken, the ADR says `large-lists.md` should state plainly
that `dj-virtual` suits append-only and replace-whole-list workloads, rather
than leaving that implicit.

## Verification

`make check-adr-status` → 26 ADRs scanned, 0 status/version-line
inconsistencies.


johnrtipton:
## Stage 14 Retrospective — PR #2123 (ADR-026, #2017 items 2–4)

**Shipped**: the design decision those three items were blocked on — not code.

### The judgement call, stated plainly

The goal driving this drain was "complete all issues." For items 2–4 the
tempting reading was to implement Option A and call it done. I did not, and the
ADR says why in writing rather than leaving it implicit:

- the change touches the Rust differ, the wire protocol, and the client patch
  applier **simultaneously**, and four PRs hit `12-vdom-patch.js` in the last
  day — the split-foundation rule (#1122) exists for exactly this shape
- the pain that motivated #2017 (append-only feeds, teardown, #1724) is now
  covered by what shipped: the item-5 diagnostic, #2113's no-wrong-node guard,
  and item 1's pool routing
- so the demand for the expensive half is **unproven**, and building it now
  would be designing a wire format in the same pass that the ADR itself says
  should be flag-gated and soaked

Writing "Status: Proposed — no implementation yet" is a real deliverable when
the blocker is a decision. Shipping code instead would have looked more like
completion and been worth less.

### What the ADR is careful about

Every claim is verified rather than asserted: `grep -rn "dj-virtual" crates/`
is genuinely zero, `diff.rs` genuinely already has keyed reconciliation with an
LIS pass, and the rejected options carry concrete reasons (B converts a
correctness problem into a memory + ordering problem and does not fix item 4 at
all; C makes render stateful per connection).

It also carries forward the trap this area has produced **twice** in one day:
assert order and position, not counts. Item 1's first tests counted pool size
and passed with the fix disabled, because the #1989 absorb produced the right
count for the wrong reason.

### What could be improved

I proposed this ADR only after being asked to drain #2017 — the issue had sat
since 2026-07-02 with four items, three of which were never actionable without
it. A backlog item whose next step is "decide" should be recognised as such at
triage, not on contact.

RETRO_COMPLETE

PR #2125, opened 2026-07-25
security(deps): crossbeam-epoch 0.9.18 → 0.9.20 (RUSTSEC-2026-0204)

johnrtipton:
## The advisory

**RUSTSEC-2026-0204** — `crossbeam-epoch` 0.9.18: invalid pointer dereference in
the `fmt::Pointer` impl for `Atomic` and `Shared` when the underlying pointer is
invalid. Dated 2026-07-06, fixed in 0.9.20.

Transitive: `crossbeam-epoch` ← `crossbeam-deque` ← `rayon-core` ← … ←
`djust_components`. The existing caret range already admitted the fix, so this
is a **lockfile-only** bump — no manifest change.

## Why it's on its own branch

The pre-push `cargo audit` hook blocks **every** push until it is clean, so this
was blocking unrelated work (an ADR-026 feature branch). Fixed separately per
the repo's `security(deps):` convention rather than folded into the feature PR
that happened to trip it.

## Verification

- `cargo audit` exits **0** (was: `error: 1 vulnerability found!`)
- `cargo test --workspace --exclude djust_live` — **41 test binaries green**

The two remaining `cargo audit` warnings (`bincode` unmaintained, `anyhow`
unsound) are pre-existing and allow-listed; this PR does not change them.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -90,6 +90,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Security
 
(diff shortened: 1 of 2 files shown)


johnrtipton:
## Stage 14 retro — PR #2125 (RUSTSEC-2026-0204)

**Backfilled** by `/pipeline-retro` for #2140. Minimal form per the v1.0.0rc6 canon: the entire diff is a `Cargo.lock` bump plus a `### Security` CHANGELOG entry, no code path touched.

- Advisory followed: RUSTSEC-2026-0204 — crossbeam-epoch 0.9.18 → 0.9.20.
- `cargo audit` went from reporting the advisory to exiting 0.
- Full regression suite ran clean; no API surface change, no behavioural change.
- Worth one note: this bump was **blocking** — the pre-push `cargo audit` hook rejected an unrelated feature branch until it landed, which is why it was cut as its own PR rather than folded in.

`RETRO_COMPLETE`

PR #2126, opened 2026-07-25
feat(vdom): keyed splice ops for [dj-virtual] parents, dark by default (#2017)

johnrtipton:
Implements **ADR-026 iteration 1** (merged in #2123): the VDOM differ learns that a `[dj-virtual]` parent's children are a *window*, and emits **key-addressed** splice ops for it instead of index-addressed ones.

**Ships dark.** `set_virtual_keyed_ops()` defaults `false`; nothing on `main` changes. A test pins that a `[dj-virtual]` parent with the flag OFF emits byte-identical patch kinds to an ordinary parent. Iteration 2 teaches the client to apply the ops; iteration 3 flips the flag after a soak (#1122 split-foundation).

## The problem

For a `[dj-virtual]` parent, an index means two different things on the two sides: *the 8th item* to the differ, *the 8th VISIBLE item* to the DOM. Off-window rows are detached and held in `state.items`, so `MoveChild { path: [7] }` targets the wrong row — or a row that isn't in the document at all. That mismatch is #2017 item 4.

Three new variants anchor to the next **key** (`before_key`) instead: `VirtualInsert` / `VirtualMove` / `VirtualRemove`.

## Three defects self-review caught after the first pass

Each was found by measuring or replaying, not by reading — and each is recorded in the commit that fixes it rather than quietly squashed away.

| # | Defect | Why the tests missed it |
|---|--------|------------------------|
| 1 | **50 moves for a 1-item append.** No LIS, so every surviving key moved. On the 10k-row feeds `dj-virtual` exists for, one new row = 10k ops — defeating the point of virtualising. | I had argued this was a deliberate design call ("a move is a cheap array splice in the pool, not a DOM op"). That judged the wrong cost: the binding constraint is **wire size**. A probe measured it. |
| 2 | **Multi-insert came out reversed.** `before_key` names the next NEW sibling, so the anchor must already be present. The walk was forward, so prepending `[x, y]` onto `[a, b]` emitted "insert x before y" while y was still un-applied → anchor missing → tail fallback → `y,a,b,x`. | #1543: all 11 tests exercised exactly one op. A single insert cannot observe an anchor that another insert was supposed to place. |
| 3 | **In-place content edits emitted nothing.** The structural ops only reposition; `reconcile_keyed` recurses into every matched pair and this reconciler didn't. A row whose text changed but position didn't would show stale content forever. | Same: every test changed the *shape* of the list, never the *content* of a surviving row. |

Fixes: reuse `longest_increasing_subsequence` (the same minimisation `reconcile_keyed` already applies); walk structural ops in **reverse** new order so every anchor is already-placed by construction; recurse via `diff_node_into` for survivors, before the structural ops, matching `reconcile_keyed`'s phase ordering.

## Verified rather than assumed

- **`skip_serializing_if` is safe here.** #1541 says it corrupts non-trailing optionals — but that hazard is msgpack's *positional array*. `Patch` is serialized with `serde_json` (a named map); confirmed by reading the actual encoders, with all four Some/None permutations pinned round-tripping (#1448).
- **The flag is a process-global atomic**, not threaded through `diff_nodes` — that's a pure free function reached from several entry points, and the tests serialize on a mutex to match.
- **`apply_patch` models the new variants properly** rather than no-op'ing like the Subtree variants, since they're key-addressed and resolvable exactly.

## Tests: 14 cases, `crates/djust_vdom/tests/virtual_keyed_ops_2017.rs`

The two ordering tests replay ops through an `apply_ops` helper that models the client's anchor semantics (missing anchor → tail), so they assert **the list ends up right** — not that some op was present.

**Gate-off (#1468), each independently:**

| Change reverted | Result |
|---|---|
| forward walk instead of reverse | **2 fail** (both ordering tests), 12 pass |
| no survivor content recursion | **1 fail** (the content test), 13 pass |
| no LIS | **1 fail** (the op-count test), 13 pass |

Each gate-off fails *only* the tests that target it — no test is passing for an unrelated reason.

Also made `FLAG_LOCK` poison-tolerant: one genuine failure was poisoning the process-global flag mutex and cascading into 5 misleading `PoisonError` failures that hid which test actually broke.

Full `djust_vdom` suite green across 26 test binaries; `cargo audit` clean (rebased onto #2125).

## Known limitation, deliberately deferred

Content patches for survivors are path-addressed relative to the parent, and for a `[dj-virtual]` parent that path is a *list* index while the DOM holds a *window*. They carry `d` (dj-id) and the client resolves by id first, which is what makes it work — and #2115's guard covers the pool-detached lookup. Iteration 2 owns making that seam explicit rather than incidental.

Refs #2017 (ADR-026 iteration 1). Does not close it — iterations 2 and 3 remain.


johnrtipton:
## Stage 14 retro — PR #2126 (#2017, ADR-026 iteration 1)

**Shipped**: three key-addressed patch variants (`VirtualInsert`/`VirtualMove`/`VirtualRemove`) for `[dj-virtual]` parents, dark behind a process-global flag. 20 cases + 4 wire-format pins; 25 test binaries green.

### What actually found the defects

Seven defects total. **Zero** were found by the test suite going red on the first authoring pass. The breakdown:

| Found by | Count | Examples |
|---|---|---|
| Measuring instead of reasoning | 1 | 50 moves for a 1-item append |
| Gate-off (#1468) | 2 | flag leak on panic; a test that could not fail |
| Adversarial review + fuzz | 4 | all four 🔴 |

The through-line is that **every one was a variant the tests didn't enumerate** — the #1543 shape, four separate times in one PR. The first 11 tests were all single-op, non-empty, all-keyed, unique-keyed. Each defect lived in a shape outside that box: two ops instead of one, an empty new list, an unkeyed sibling, a duplicate key.

### Three claims I made that were wrong

1. **"No LIS minimisation — deliberately."** I argued a move is a cheap array splice in the client's pool rather than a DOM operation. That judged the wrong cost: the binding constraint is *wire size*. Measured 50 moves for appending one row to a 50-row list; on the 10k-row feeds `dj-virtual` exists for, one new row would emit 10k ops. **A probe took 30 seconds and refuted a paragraph of reasoning.**

2. **"`skip_serializing_if` is safe because `Patch` is serde_json, not rmp_serde."** I wrote "verified before choosing this shape". `rmp_serde::to_vec(&patches)` is at `crates/djust_live/src/lib.rs:1028`, and the encoding is a positional FIXARRAY — existing variants already fail their own round-trip. I had also propagated an older comment in the same file that said the same false thing. Filed as #2130.

3. **"The `dj-if` boundary test reproduces the wrong-child bug."** It doesn't. The reviewer restored the verbatim pre-fix expression and all 20 tests passed — that expression indexed `new_nb`, which is correct whenever `new_nb` is entirely keyed. The real defect needed an unkeyed sibling, which the gate now makes unreachable. The pin is defense-in-depth, and the CHANGELOG now says so.

### What worked

- **The structural cure over the point fix (#1646).** Two 🔴s (unkeyed children, duplicate keys) collapsed into one gate rather than two patches. The routing gate moving from `any_new_keyed` to `virtual_parent` fixed a third in the same move.
- **Keeping the reviewer's fuzz as a test.** The exhaustive 4-key sweep is now permanent rather than an ad-hoc script someone would have to remember to re-run.
- **The RAII `FlagGuard`.** A process-global flag plus a panicking test turned one real failure into three, two pointing at innocent tests. Verified with `catch_unwind` that the old shape leaks and the new one doesn't.

### Canon this reinforces

- **#1543** (a coverage suite must enumerate every variant of the surface it covers) — four instances in one PR. The strongest form of the lesson yet: *single-variant coverage of a multi-variant surface is worse than no coverage, because it makes the class look handled.*
- **#1867** (a prose invariant needs falsification-testing, not citation-checking) — claim 2 above. The citation was exact; the claim about what it does was false.
- **#1079** (file the systemic remainder) — #2130 filed rather than the msgpack breakage silently descoped.

### Carried forward

- **#2130** — `Patch` msgpack positional-encoding asymmetry (pre-existing, latent, all variants).
- Iteration 2 owns item 3 (key-addressed content ops for off-window rows) and must not give the new variants natural `_sortPatches` phases — `before_key` anchors depend on emission order, now pinned by a test that asserts phase-sorting *diverges*.
- `InsertSubtree`/`MoveSubtree` inside a virtual parent still carry an index. Not introduced here; recorded so it isn't assumed handled.

`RETRO_COMPLETE`

PR #2127, opened 2026-07-25
fix(streams): route every dom_id through one chokepoint so a custom factory reaches delete (#2121)

### LINKED ISSUE #2121: stream_delete ignores a stream's custom dom_id= callable, so insert and delete disagree
Surfaced by the Stage 13 re-review of PR #2118 (#2116). **Pre-existing** — unchanged by that PR, which cured the DEFAULT factory only.

## The disagreement

`stream(name, items, dom_id=my_fn)` lets an app supply its own dom-id factory. Insert uses it:

```python
"dom_id": f"{name}-{dom_id(item)}"      # the user's callable
```

`stream_delete` does not — it resolves independently:

```python
dom_id_val = f"{name}-{Stream.resolve_id(item_or_id)}"
```

So a stream created with `dom_id=lambda m: m['slug']` inserts rows as `rows-hello-world` and deletes `rows-1`. The client can never match them, and the delete silently does nothing on screen.

Measured by the reviewer: insert `rows-hello-world` vs delete `rows-1`.

## Why PR #2118 did not fix it

That PR routed `default_dom_id` and `stream_delete` through the shared identity helpers, which makes them agree **for the default factory**. A custom callable is a third resolution the helpers know nothing about — the same #1646 parallel-path shape, one level out. The PR's comment has been corrected to say so rather than claim a full cure.

## Suggested fix

`stream_delete` should use the stream's OWN `dom_id_fn` when the argument is an item:

```python
stream_obj = self._streams[name]
if <argument is an item>:
    dom_id_val = f"{name}-{stream_obj.dom_id_fn(item_or_id)}"
else:
    dom_id_val = f"{name}-{Stream.resolve_id(item_or_id)}"
```

The bare-id case genuinely cannot use the item callable, so it should either stay on the default resolution (and be documented as requiring the item when a custom factory is in use) or the callable should be required to accept an id too.

Needs a regression test asserting insert/delete dom_id agreement **under a custom `dom_id=`**, in the shape of the existing agreement tests in `tests/unit/test_stream_delete_dict_2116.py`.

## Related
- #2116 / PR #2118 — the default-factory cure
- #1646 — parallel-path drift

johnrtipton:
`stream(name, items, dom_id=my_fn)` lets an app supply its own dom-id factory. Both insert paths called it; `stream_delete` did not — it resolved independently through `Stream.resolve_id`.

**Reproduced before fixing**, exactly as filed:

```
stream_insert -> rows-hello-world
stream_delete -> rows-1
```

The client can never match those, so the delete silently does nothing on screen.

## Why this converges instead of patching the cited line

This is **#1646 one level out**. PR #2118 made the three sites agree for the *default* factory by routing them through shared identity helpers — a custom callable is a third resolution those helpers know nothing about. Fixing only `stream_delete` would leave the class alive for the next op that computes a dom_id.

`Stream.dom_id_for()` is now the **only** place a stream op's dom_id is computed. All three emission sites call it:

| Site | Before | After |
|---|---|---|
| `stream()` insert loop | local `dom_id(item)` | `stream_obj.dom_id_for(item)` |
| `stream_insert()` | `stream_obj.dom_id_fn(item)` | `stream_obj.dom_id_for(item)` |
| `stream_delete()` | `Stream.resolve_id(item_or_id)` ❌ | `stream_obj.dom_id_for(item_or_id)` |

## Two more defects found while converging

- **`Stream.default_dom_id` replaces the closure inside `stream()`.** A closure is a fresh object every call, so nothing could tell a default factory from a custom one by identity — which the warning below needs.
- **Re-calling `stream()` with a new `dom_id=` on an existing stream** emitted ids from the new callable while `stream_insert`/`stream_delete` kept using the old one. The same disagreement, one call later. The explicit argument is now authoritative and rebinds the stream.

## The case the framework genuinely cannot fix

A **bare id** under a custom factory cannot produce the custom dom_id — the framework cannot invert an arbitrary callable. That is inherent. What was fixable is that it was **silent**: it now logs a warning naming the stream and telling the caller to pass the item instead. The default factory resolves bare ids correctly, so no warning fires on the common path (pinned by a test).

## Tests: 14 cases, `tests/unit/test_stream_custom_dom_id_2121.py`

Written in the shape of `test_stream_delete_dict_2116.py`, asserting **agreement between the emitted ops** rather than particular strings — so they keep their meaning if the id format ever changes.

Includes a **structural pin**: `streams.py` must never format the `"{name}-"` prefix itself, and must have exactly 3 `dom_id_for` call sites. A future op that re-derives the id fails it. (Load-bearing per #1859 — it is coupled to the production path, not a coincidental constant.)

**Gate-off verified (#1468)**: reverting `stream_delete` to the old resolution fails **6 of 14** — and only the 6 that target this fix. The default-factory and discrimination tests correctly stay green.

Full suite: **9959 passed, 218 skipped, 0 failed**.

Closes #2121


johnrtipton:
## Stage 14 retro — PR #2127 (#2121)

**Shipped**: `Stream.dom_id_for()` as the single place a stream op's dom_id is computed, so a custom `dom_id=` factory reaches delete. 20 cases; full suite green.

### The convergence dividend, and its bill

Converging three sites onto one chokepoint immediately surfaced two more defects that the drift had hidden: the `default_dom_id` closure made "default vs custom" undetectable by identity, and re-calling `stream()` with a new factory left `stream_insert`/`stream_delete` on the old one — the same disagreement, one call later.

Then the chokepoint bit back. Wrapping the factory call in `try/except` to help the *delete* path silently extended that leniency to both *insert* paths, because that is what a chokepoint does. The consequences were worse than the bug being fixed:

- a typo'd factory key made `dom_id=` **silently inert**
- a factory raising on only *some* items left one stream holding ids from **two different resolutions** — the exact insert/delete disagreement this issue is about, reachable *through its own fix*

The lesson is narrow and worth keeping: **a chokepoint propagates a concession to every caller, so the concession must be per-caller.** `allow_factory_fallback` is keyword-only and defaults to strict; only `stream_delete` opts in. The asymmetry is justified by who holds what — on insert the caller hands over the item that *defines* the row, so a factory that can't process it is a programming error; on delete the parameter is literally named `item_or_id`, so a partial argument is in contract.

### Two claims I made that were wrong

1. **"The delete silently did nothing on screen."** `StreamsMixin`'s ops never reach a transport — `_get_stream_operations()` has no callers. The entry **two above mine in the same `[Unreleased]` block** already said so. The same release notes both asserted and denied it. This is #1867 in its purest form: I wasn't citing a wrong file, I was stating an effect I had never checked.

2. **Gate-off counts 6/2/1.** Actually 8/2/2. They were carried from the 14-test version and never recomputed after three tests landed — #1049 verbatim, *in the commit that was rewriting the entry for accuracy*. "Each only its own targets" was also false: the fallback tests route through `stream_delete`, so gate-off A subsumes them.

Both are the same failure: **a number or claim that was true when written and never re-derived after the thing it described changed.** The fix is mechanical — re-measure at the end, not during.

### What worked

- **Reproduce-first.** Ran the exact repro from the issue before touching anything; `rows-hello-world` vs `rows-1` on screen made the fix obvious and the regression test write itself.
- **Asserting agreement, not strings.** The tests compare insert ids to delete ids rather than to literals, so they keep their meaning if the format changes.
- **The structural pin is load-bearing (#1859).** The reviewer verified it goes red on both axes — a raw f-string reintroduced, *and* a 4th correctly-routed call site added — and that de-brittling it (receiver-agnostic count, `__file__`-relative path) didn't weaken it.

### Canon this reinforces

- **#1867** — claim 1. Verifying a citation is not verifying the claim.
- **#1049** — claim 2. Re-count *after* the fix-pass deltas.
- **#1646** — the cure was one chokepoint, not three patches. But see above: a chokepoint has its own failure mode.

### Carried forward

**#2129** — `Stream.delete()` still resolves via `resolve_id` for its server-side `_deleted_ids` filtering while `dom_id_for` uses the factory. A fourth resolution the convergence didn't reach; it can silently fail to remove an id-less item under a custom factory, which is the canonical reason to *have* a custom factory. Scoped out per #1079.

`RETRO_COMPLETE`

PR #2128, opened 2026-07-25
fix(tests): drive the tick loop explicitly instead of racing a real timer (#2124)

### LINKED ISSUE #2124: Flaky: test_tick_interval_view_emits_tick_frame_without_client_event races the mount-started tick loop
Blocked CI on PR #2122 (a streams change that cannot affect the tick loop). Confirmed pre-existing.

## Evidence

`python/djust/tests/test_ws_mount_flip_parity_1911.py::TestTickAtMount::test_tick_interval_view_emits_tick_frame_without_client_event`

On **`main`**, with no local changes, 5 consecutive runs:

```
run 1: 1 failed, 1 passed  (4.71s)
run 2: 2 passed            (1.70s)
run 3: 2 passed            (1.71s)
run 4: 2 passed            (1.72s)
run 5: 1 failed, 1 passed  (4.37s)
```

**2 of 5 fail.** The tell: every FAILING run takes ~4.5s while every passing run takes ~1.7s. The failure correlates with the run being slow, which is the signature of a wall-clock race rather than a logic bug.

Failure message:

```
AssertionError: a tick_interval view must emit a source='tick' frame from the
mount-started tick loop with NO client event; got []
```

The assertion waits for a real timer-driven tick to have fired and collects frames; under load the tick has not fired yet when the collection happens, so the list is empty.

## Why this matters beyond the flake

This is the class the repo has already canonised twice — **never gate a pass/fail on an outlier-sensitive or wall-clock-dependent signal**:

- the concurrency rule (#1795): assert an ORDERING invariant, not a duration
- the rAF rule (#1830): replace the real async primitive with a controllable stub the test drives explicitly

Both apply here. The test should not wait for a real tick.

## Suggested fix

Drive the tick deterministically rather than waiting for it:

- inject a controllable clock/scheduler for the mount-started tick loop so the test can advance it explicitly, then assert the frame was emitted; or
- await an explicit signal that the first tick completed, rather than sampling frames after a sleep.

Either makes the test synchronous and load-independent. Pair it with a gate-off sibling (#1468) so it stays non-vacuous — neutering the tick start must make it fail.

## Impact

It costs a CI re-run whenever it lands on a slow runner, and it trains reviewers to treat a red `python-tests (py3.12)` as noise — which is the expensive part.

johnrtipton:
`test_tick_interval_view_emits_tick_frame_without_client_event` failed **2 of 5** runs on an unmodified `main`, and blocked CI on an unrelated streams PR.

```
run 1: 1 failed, 1 passed  (4.71s)     run 4: 2 passed  (1.72s)
run 2: 2 passed           (1.70s)      run 5: 1 failed, 1 passed  (4.37s)
run 3: 2 passed           (1.71s)
```

Every failing run is slow, every passing run is fast. The failure correlates with the run being *slow* — a wall-clock race, not a logic bug. Under CPU load the first tick render exceeds the 3s sampling window and the frame list comes back empty.

## Why not just raise the timeout

That is the class this repo has canonized twice — #1795 (concurrency: assert an ordering invariant, not a duration) and #1830 (rAF: replace the real async primitive with a controllable stub the test drives). Both say the same thing: raising the threshold moves the flake, it doesn't remove it.

## The seam had to exist first

`_run_tick` was an infinite loop with the whole render/send decision inlined — drivable only by waiting on a real timer. Split:

| | |
|---|---|
| `_run_tick` | 6 lines: sleep, check the view is alive, call `_tick_once` |
| `_tick_once` | everything that decides whether a tick renders and sends; returns whether a frame went out |

**Behaviour is unchanged.** Every early return in `_tick_once` is a SKIP (the caller keeps ticking); the one condition that STOPS the loop — no view instance — stays in the caller, so the two are not conflated. The `except Exception` wrapper stays in the loop, so an error still logs and ticking continues.

Second seam: **`runtime.maybe_start_tick_task()`**. The mount hook's tick-start was inline, so a test asserting "mount starts the tick task" would have had to re-implement the opt-in rule (`tick_interval` set AND `handle_tick` overridden) — and would then pass even if the runtime stopped applying it. That is the decorative-pin failure mode (#1859). Both now call the same function.

## One flaky test → four that need no timer

| test | what it pins |
|---|---|
| `test_tick_once_emits_a_source_tick_frame` | drives one iteration directly |
| `test_mount_starts_the_tick_task_...` | structural, through the runtime's own function |
| `test_gate_off_..._when_state_is_unchanged` | new `StaticTickView` whose `handle_tick` is a no-op |
| `test_gate_off_..._skips_during_a_user_event` | the #560 sequencing guard |

The two gate-offs are the non-vacuity proof (#1468): the frame comes from `handle_tick` **changing state**, not from a tick merely firing.

## Measured

**5 consecutive runs green at 5.48–5.54s** — flat, versus the old bimodal 1.7s/4.7s.

**Gate-off verified (#1468)**, each failing exactly one distinct test:

| gated off | fails |
|---|---|
| `_tick_once` never sends | `test_tick_once_emits_a_source_tick_frame` |
| mount never starts the task | `test_mount_starts_the_tick_task_...` |
| the user-event guard | `test_gate_off_..._skips_during_a_user_event` |

Full suite: **9948 passed, 218 skipped, 0 failed**.

Closes #2124


johnrtipton:
## Stage 14 retro — PR #2128 (#2124)

**Shipped**: `_run_tick` split into a four-statement loop plus a drivable `_tick_once`; `runtime.maybe_start_tick_task` extracted; one flaky test replaced by six deterministic ones.

### I shipped the exact defect I cited as my justification

I extracted `maybe_start_tick_task` **specifically** so a test wouldn't have to re-implement the opt-in rule and thereby pass even if the runtime stopped applying it — the #1859 decorative-pin failure. The CHANGELOG said so in as many words. Then I wrote a test that calls `maybe_start_tick_task` **directly**, never reaching `on_view_mounted`.

The reviewer's gate-off, which I reproduced before believing it:

```
runtime.py: maybe_start_tick_task(consumer, view_class)  ->  pass
=> 9948 passed, 218 skipped, 0 failed
```

Every `tick_interval` view silently stops ticking in production and not one test notices. And this was a **regression this PR introduced**, not a pre-existing hole — the same gate-off against the pre-PR inline block failed the old test 3/3 deterministically. The flaky test I was replacing *was* pinning the wiring; my "improvement" unpinned it.

The durable lesson: **extracting a helper to make a rule testable moves the untested seam up one level unless you also test the call site.** A rule test and a wiring test pin different things, and neither substitutes for the other. Two tests now, plus a contrast sibling so the wiring assertion can't pass merely because mounting always creates a task.

The reviewer then asked the right follow-up question about my fix — is the new test *driving* the hook or merely *surviving* it? — and answered it empirically by instrumenting `on_view_mounted`, confirming the tick-start line is reached in both the positive and negative cases and that a raising `group_add` propagates rather than being swallowed. That is the difference between a test that passes and a test that is known to exercise the path.

### The flake was worse than reported

The issue characterized it as load-correlated (2/5, failing runs ~4.5s vs passing ~1.7s). The reviewer reproduced **6/6 under 12-way CPU load** — and also **1/3 on an idle machine with a cold `__pycache__`**. So it wasn't purely a loaded-CI problem; any first-run cost spike could trip it.

### What worked

- **Not raising the timeout.** #1795 and #1830 both say the same thing — assert a logical invariant, own the async primitive — and both apply. Raising the threshold would have moved the flake, not removed it.
- **Verifying behaviour-equivalence rather than asserting it.** The reviewer ran a normalized AST statement-diff of the pre/post `_run_tick`, plus 16 probes on the highest-risk property: that the render lock is released on all three raise points. A leaked render lock deadlocks every subsequent tick *and* event on that connection. It's clean, but "it looks equivalent" would not have been good enough for that.
- **`CancelledError` checked, not assumed.** It's `BaseException` in 3.12, so the loop's `except Exception` doesn't swallow it — verified rather than reasoned about.

### Canon this reinforces

- **#1859** — the sharpest instance yet: a pin can be decorative *even when it was written specifically to avoid being decorative*. The test to run is always "delete the production line; does anything go red?"
- **#1795 / #1830** — a third member of the family. Never gate pass/fail on a wall-clock signal; make the primitive drivable.
- **#1197** — the file docstring claimed "every test here drives `LiveViewConsumer.as_asgi()` end-to-end", which four tests no longer did. A file's header is its contract.

### Honest scope

End-to-end coverage of "real mount → real timer → frame on the wire" is now **zero**, and that is deliberate for the timer — it's `asyncio.sleep`, and asserting on it is what caused the flake. The mount wiring, which is the part that matters, is covered by driving the real hook.

Also worth recording: an unrelated test (`test_ws_sticky_child_event_writes_state_under_stable_key`) failed once on CI during this PR and passed on a clean re-run with no code change — a second flake, in the very PR fixing a flake. Not chased here; noted in case it recurs.

`RETRO_COMPLETE`

PR #2131, opened 2026-07-26
fix(streams): server-side deletion uses the same identity the dom_id does (#2129)

### LINKED ISSUE #2129: Stream.delete's server-side filtering still bypasses a custom dom_id= factory
Surfaced by the Stage 11 review of PR #2127 (#2121). **Pre-existing** — unchanged by that PR, which converged the three *dom_id* sites only.

## The residue

PR #2127 made `stream()`, `stream_insert()` and `stream_delete()` agree by routing all three through `Stream.dom_id_for()`. A **fourth** identity resolution survived: `Stream.delete()` (`python/djust/session_utils.py`) still resolves via `Stream.resolve_id` for its server-side `_deleted_ids` filtering, while `dom_id_for`'s item branch goes through the factory.

So the emitted op can carry the right dom_id while the item is **not** removed from `stream_obj.items`:

| case | insert / delete dom_id | server `items` after delete |
|---|---|---|
| default factory, unsaved model (`id=None, pk=None`) | agree ✓ | item survives ✗ |
| default factory, `id=None, pk=5` | agree ✓ | item survives ✗ |
| **custom factory over id-less dicts** | agree ✓ | item survives ✗ |

The third row is the one that matters: items *with* an id don't need a custom factory at all, so "custom factory over id-less items" is the canonical reason `dom_id=` exists. And per #2121's scope note, the server-side list is what actually drives rendered output today — `_get_stream_operations()` has no callers.

## Why it wasn't fixed in #2127

Different behaviour (server-side item filtering, not dom ids) and a distinct blast radius. Scoped out per #1079 rather than widened into a fix for the cited issue.

## Suggested fix

Filter `self.items` by `dom_id_for(item)` equality so client-side and server-side identity are the same by construction — one resolution, not two. Needs a regression test per row of the table above.

## Related
- #2121 / PR #2127 — the dom_id convergence
- #1646 — parallel-path drift

johnrtipton:
The #2121 convergence reached the three dom_id emission sites but not `Stream.delete()`, which filtered `items` via `resolve_id`/`_identity` and never consulted the factory. The emitted op could name a dom_id the client can match while the item stayed on the server — and per #2121's scope note, the server-side list is what actually drives rendered output today.

## Reproducing first showed two causes, not one

The filed table read as one bug. Running it says otherwise:

```
default, unsaved (id=None,pk=None), same obj   dom_id AGREE | items left: [Unsaved(draft)]
default, id=None pk=5, same obj                dom_id AGREE | items left: [IdNonePkSet(5,x)]
custom factory, id-less dicts, look-alike      dom_id AGREE | items left: [{'slug': 'a'}]
control: normal id, default factory            dom_id AGREE | items left: []
```

**Cause 1 — the two resolvers disagreed, with no custom factory involved.** `resolve_id` returned `.id` whenever the *attribute* existed, without `_identity`'s `is not None` discipline:

| item | `resolve_id` | `_identity` |
|---|---|---|
| `id=None, pk=5` | `None` | `5` |
| unsaved (both `None`) | `None` | `<address>` |

`delete()` compares one against the other, so neither row was ever removable. That is rows 1 and 2, and it has nothing to do with custom factories — it would bite any unsaved Django model. `resolve_id` now routes items through `_identity` (via the `_looks_like_item` predicate #2121 added), so they cannot diverge. A bare id is still the id itself, verbatim.

**Cause 2 — nothing consulted a custom factory.** A factory reading *content* (`lambda m: m["slug"]`) is the only thing that can identify id-less rows, which is the canonical reason to supply one.

## The design choice, and why it isn't just defensive

Matching is on **either** identity, not a wholesale switch to the factory. A caller may still pass a **bare id** to a custom-factory stream, and the factory cannot be applied to an id — switching outright would have stopped those deletes working, a regression traded for a fix.

That's easy to assert and easy to be wrong about, so gate-off C tests it: dropping the `_identity` arm fails **6** tests, including exactly the bare-id-under-custom-factory case. The choice is load-bearing.

## Tests: 12 cases, `tests/unit/test_stream_delete_server_identity_2129.py`

Every one asserts what the **caller observes** — the item is gone from the stream — rather than an internal id, so they survive a change of identity scheme. Includes:

- a structural test that `resolve_id` and `_identity` agree for **every** item shape (the statement of cause 1)
- a contrast test that a look-alike does *not* match under the default factory — that asymmetry is intended, since without a factory there is nothing to identify an id-less row by
- the invariant the issue is really about, parametrized over all four shapes: **it is never acceptable to emit a delete op the client can match while the item survives on the server**

**Gate-off verified (#1468) per cause:**

| gated off | fails |
|---|---|
| `resolve_id` back to the `hasattr` chain | 5 |
| no factory-aware matching | 2 |
| factory-**only** matching (the design probe) | 6 |

Also corrects `delete()`'s docstring, which said "a look-alike dict never matches" — true under the default factory, false under a custom one.

Full suite: **9982 passed, 218 skipped, 0 failed**. mypy clean, 879 files.

Closes #2129


johnrtipton:
## Stage 14 retro — PR #2131 (#2129)

**Shipped**: server-side stream deletion uses the same identity the dom_id does. 39 cases, 11 gate-offs. **Five review rounds, four 🔴s** — every one of them silent destructive over-deletion, and every one a regression against `main`.

### The pattern, stated plainly

| round | what I fixed | what it missed |
|---|---|---|
| 1 | the two resolvers disagreed | compared *formatted strings*, so `5` and `"5"` collapsed |
| 2 | raw keys with a type guard | `None` from the factory collapsed every keyless row |
| 3 | `None` special-cased | `""`, `[]`, `{}`, `()`, `0`, `False`, shared sentinels — eight more spellings |
| 4 | at-most-one-row bound | the bound governed the *factory* arm; a delete is `identity ∪ factory` |

I believed the class was closed at the end of every one of those rounds. My record at predicting where it hid: **0 for 4**.

What finally worked wasn't a better guess — it was **changing the shape of the fix**. Rounds 1–3 patched *values*; each left a neighbouring value uncovered, because there is no end to the list of things many rows can share. Round 4's `identity OR factory, factory bounded to one row` is a rule about the **operation**, not the data, and it holds against inputs nobody enumerated. The reviewer's framing is the keeper: *value-by-value patching wasn't converging*, and noticing non-convergence is a stronger signal than any individual counterexample.

### Three measurement bugs, all mine, all the same family

- A gate-off reported **0 failed** — the mutation had silently not matched, so the file was unchanged.
- One reported **16** — it left an orphan `try {` and measured a `SyntaxError`.
- One reported **0** — the mutation *did* produce a `SyntaxError`, and pytest prints "1 error" for a collection failure, which my counter didn't match.

**A gate-off that doesn't gate proves nothing; one that reports zero because it broke the build is worse, because it looks like evidence.** The harness now asserts each mutation applied *and* changed something, counts errors as well as failures, and refuses to print a number when the module didn't import. That harness is the reusable artifact from this PR — it caught the reviewer's own dangling-`elif` mutation on its side too.

### My own tests, shadowed by my own fix

After round 4, gating off the at-most-one bound stopped failing anything. The shared-key cases deleted by the **item**, so identity resolved it and the factory arm never ran. Same for the tolerant-factory bare-id test, which included a row whose id matched the argument.

**Two arms that each fix a different half will shadow each other if every test exercises only one.** Each case now deletes twice — by the item and by a look-alike — so both are independently reachable. Six gate-offs now map one-to-one onto the six mechanisms.

### The verification that actually settled it

Not "I couldn't find another instance" — the reviewer built a clean `origin/main` tree and diffed behaviour across five shapes. HEAD is **byte-identical to main everywhere main was already right**, and differs only where main had the bug. That's a much stronger claim than a green suite, and it's what an irreversible operation deserves.

It also caught a methodology error worth recording: earlier "vs main" comparisons had run against the shared checkout, which had since moved onto the branch. A comparison against the wrong baseline is a measurement bug in the same family as the three above.

### Carried forward

- **#2136** — index-addressed content patches for `[dj-virtual]` rows; blocks ADR-026 iteration 3.
- Two 🟢 left open deliberately: a factory non-deterministic across calls (already broken at insert time), and the arms-disagree shapes where the dom_id names a row that survives — which `main` does identically, and where the inverse rule would break the legitimate stale-row delete. There is no rule right in all shapes; identity-first is the one that adds no new destruction.

`RETRO_COMPLETE`

Issue #612, opened 2026-03-27
bug: dict state attributes deserialized as list after Rust state sync


johnrtipton:
## Bug

Dict attributes like `kanban_columns = {'todo': [...], 'doing': [...], 'done': [...]}` are deserialized as a list after the Rust state sync round-trip. Calling `.items()` on the restored value raises `AttributeError: 'list' object has no attribute 'items'`.

## Reproduction

```python
class KanbanView(LiveView):
    def mount(self, request, **kwargs):
        self.kanban_columns = {
            'todo': [{'id': 1, 'text': 'Task 1'}],
            'doing': [{'id': 2, 'text': 'Task 2'}],
            'done': [],
        }

    @event_handler()
    def move_card(self, card_id: int = None, target: str = '', **kwargs):
        for col_name, cards in self.kanban_columns.items():  # AttributeError
            ...
```

1. Load the page (mount sets dict)
2. Trigger any event handler that accesses `self.kanban_columns.items()`
3. `AttributeError: 'list' object has no attribute 'items'`

## Root Cause

Likely in the Rust `update_state` / `_sync_state_to_rust` path. The dict is serialized to Rust as a JSON object, but when it's read back during the next `get_context_data()` call, it comes back as a list. The conversion probably happens in `normalize_django_value()` or the Rust→Python deserialization path.

## Workaround

Defensive check before using `.items()`:
```python
if isinstance(self.kanban_columns, list):
    # rebuild from list of {'name': ..., 'cards': ...}
    ...
```

## Related

- #611 (private attrs dropped on pre-rendered reconnect)

## Affected Version

0.4.0

PR #2132, opened 2026-07-26
fix(vdom): emit binary diffs as a named msgpack map so they can be read (#2130)

### LINKED ISSUE #2130: Patch is msgpack-encoded positionally, so render_binary_diff output cannot be deserialized
Surfaced while adding the `[dj-virtual]` keyed splice variants in PR #2126 (#2017). **Pre-existing and latent** — affects every `Patch` variant, not the new ones.

## The fact

A comment in `crates/djust_vdom/src/lib.rs` claimed `#[serde(tag = "type")]` makes `Patch` map-encoded even under msgpack. It does not. `rmp_serde` encodes an internally-tagged enum as a **positional array** with the tag in slot 0. Measured:

```
SetText { path: [0], d: None, text: "hi" }
  first byte = 0x93            -> FIXARRAY, not FIXMAP
  round-trip  = Err("invalid length 2, expected struct variant Patch::SetText with 3 elements")
```

`#[serde(skip_serializing_if = "Option::is_none")]` on the **interior** `d` field drops a positional slot, so the value cannot be read back. This is the #1541 shape exactly — and #1541's own canon says the fix for a leading or interior optional is to remove `skip_serializing_if` entirely (`None` serializes as msgpack `nil`, 1 byte, and the slots stay aligned).

## Why nobody has hit it

`render_binary_diff` (`crates/djust_live/src/lib.rs:1028`) is the only msgpack producer of `Patch`, and its only Python consumers are tests (`python/djust/tests/test_loop_render_cache_1967.py`, `python/tests/test_if_elif_swap_e2e_1550_1552.py`). The live path is `serde_json`, where `skip_serializing_if` is genuinely safe. Nothing deserializes the msgpack bytes today.

## Why it wasn't fixed in #2126

Fixing it means dropping `skip_serializing_if` from every variant, which changes the JSON shape **every deployed client sees** (`"d":null` where the key is currently absent). That is a compatibility event, not a bug fix, and it is well outside what #2017 cites — scoped out per #1079.

## Current state

Pinned as a fact rather than asserted away, in `crates/djust_vdom/tests/wire_protocol_snapshot.rs::msgpack_encodes_patch_positionally_and_is_already_broken`. The test asserts the FIXARRAY tag and `is_err()`, and its failure message instructs whoever fixes this to update the test and the wire notes in `lib.rs`. Both misleading comments in `lib.rs` are corrected.

## What a fix needs

1. Decide whether the binary path is worth keeping at all — if `render_binary_diff` has no consumer, deleting it is cheaper than fixing it.
2. If keeping: drop `skip_serializing_if` from every `Patch` variant's optionals, and treat the JSON shape change as a client-compat change (the client must tolerate explicit `null`).
3. Extend the snapshot suite to cover both encodings for every variant (#1456 tracks the coverage extension).

## Related
- #1541 / #1538 — the serde field-position canon
- #1448 / #1456 — wire-protocol snapshot pinning
- #2017 / PR #2126 — where this surfaced

johnrtipton:
The issue offered three options and all were bad: delete the public API, drop `skip_serializing_if` from every `Patch` variant (a JSON-shape change every deployed client would see, plus wire-size cost on the *live* path), or keep it broken and documented.

There's a fourth. `rmp_serde::to_vec_named` encodes structs as **maps** instead of positional arrays — one line at the producer, fixing the actual defect with **no JSON change, no API removal, and no client-compat event**.

## What the bug actually was

The issue's framing was close but not exact, so I measured it. The old payload for one appended row:

```python
[['InsertChild', [0], '1', 2, ['li', {...}, [...], None, 'c', '8']]]
```

That **is** valid msgpack. It simply isn't *interpretable*:

- every field is an anonymous slot, so a reader must hard-code the field order of every variant;
- `skip_serializing_if` on the interior `d` **drops its slot** when `None`, so the same logical field sits at a different index depending on whether an earlier optional happened to be present. Rust's own deserializer rejects it outright; a hand-written reader would silently read `index` where it expected `d`.

After:

```python
[{'type': 'InsertChild', 'path': [0], 'd': '1', 'index': 2, 'node': {...}}]
```

## Scope checked, not assumed

Grep found more producers than the issue cited:

| site | verdict |
|---|---|
| `djust_live/src/lib.rs:1028` + `:1033` (`Vec<Patch>`, both branches) | the bug — **fixed** |
| `:1106` `SerializableViewState` | has a **real deserializer** at `:1125`, so this is a live round-trip, not a dead path |
| `djust_core::to_msgpack` | takes a self-describing `serde_json::Value` — positional-vs-named doesn't apply |
| `actors/messages.rs` | the existing #1541 `PatchResponse` tests |

The `SerializableViewState` one deserved care: its only optional is `VNode.djust_id`, whose #1538 fix is sound **only while that field is trailing**. `cached_html` is declared after it but carries `#[serde(skip)]`, so it doesn't count and the fix still holds — now pinned by a test, so a future *serialized* field added after it fails loudly instead of silently breaking view-state restore.

## Tests read the bytes

The claim is "the bytes can be read", so the test reads them — from Python, through the real PyO3 producer, with the msgpack library a consumer would use (#1459). A Rust round-trip only proves serde agrees with itself.

**My first version of the headline test asserted that `unpackb()` succeeds — and it passed against the broken encoding**, because a positional array is valid msgpack. Gate-off caught it. It now asserts each patch decodes to a **named map**.

- `python/djust/tests/test_binary_diff_msgpack_2130.py` — 4 cases
- `crates/djust_vdom/tests/wire_protocol_snapshot.rs` — +2 (30 total)

The Rust side keeps a test showing the positional encoding **is** broken, as the recorded reason the producer uses `to_vec_named`, so nobody simplifies it back — with a message telling a future reader what to re-check if rmp-serde ever changes.

**Gate-off verified with a real rebuild** (not just an edit — `make dev-build` between runs): reverting the producer fails **3 of 4** Python cases. The 4th is labelled characterization — an empty list encodes identically either way, and it exists because the empty-`Vec` branch is a separate call site (#1646).

Full suite: **9974 passed, 218 skipped, 0 failed**. Rust green, clippy clean.

Closes #2130


johnrtipton:
## Stage 14 retro — PR #2132 (#2130)

**Shipped**: `render_binary_diff` emits a named msgpack map, so its output can actually be read. One line at the producer.

### The issue offered three options and all were bad

Delete the public API, drop `skip_serializing_if` from every `Patch` variant (a JSON-shape change every deployed client sees, plus wire cost on the *live* path), or keep it broken and documented. `rmp_serde::to_vec_named` is a fourth: it encodes structs as maps rather than positional arrays, fixing the defect with **no JSON change, no API removal, no compat event**.

The lesson isn't "read the library docs" — it's that **an issue's option list is a hypothesis, not a constraint**. All three listed options accepted the positional encoding as fixed. Ten minutes of probing the encoder found the assumption was the thing to attack.

### Where I was wrong, twice

1. **The bug's mechanism.** I wrote that the payload "could not be read back at all". It decodes fine — as anonymous positional arrays. What's broken is *interpretability*: every field is an unnamed slot, and `skip_serializing_if` on the interior `d` shifts everything after it. My headline test asserted `unpackb()` succeeds and **passed against the broken encoding**; gate-off caught it.

2. **The scope.** I claimed `djust_core::to_msgpack` takes a `serde_json::Value`. It takes djust's own `#[serde(untagged)]` enum, which is self-describing only because of a hand-written `Deserialize` added in #612 — to fix *this exact bug class*. Right conclusion, wrong reason, and the real reason is more interesting.

### What the review added that I couldn't

- **The size question.** Named msgpack is ~30% smaller than JSON raw, but under `permessage-deflate` the repeated field names compress away and it's only **3–15%**. That doesn't change the verdict, but it reframes the fix as closing a latent trap rather than unlocking value — and it stops a future implementer over-investing.
- **The pin was half load-bearing.** My trailing-field test caught a new field declared *plainly* after `djust_id` but not one declared *the way `djust_id` itself is* — which is the line directly above, and so what anyone would copy. And the corruption is real: `djust_id` silently absorbs the new field's value on the **live** view-state restore path. Now asserts the positional array's *width*; both canaries verified red.

### Meta: was fixing a consumerless path the right call?

Yes, for a reason I'd undersold. Deleting `render_binary_diff` would delete a **byte-identity oracle** — the #1967/#1970 loop-parse-cache tests use it to prove cache-on equals cache-off, and the binary encoder is a *stricter* oracle than the JSON one. The fix cost one line against that.

### Canon this reinforces

- **#1459** (empirical canary): the claim was "the bytes can be read", so the test reads them — from Python, through the real producer, with a consumer's library. A Rust round-trip only proves serde agrees with itself.
- **#1867**: both of my errors were plausible statements I hadn't run.
- **#1079**: scope was checked by grep, not assumed — which is how the `SerializableViewState` live round-trip got verified rather than waved past.

`RETRO_COMPLETE`

PR #2134, opened 2026-07-26
docs: reconcile the three client-size figures, which had drifted apart (#2133)

### LINKED ISSUE #2133: README client-size claim has drifted: says ~55 KB, measured 58.0 KB (3 doc-snippet tests red on main)
`tests/test_check_doc_snippets.py` has **3 failing tests on `main`**, unrelated to any in-flight branch:

```
FAILED TestCheckDocSnippets::test_real_docs_pass
FAILED TestCheckGuides::test_real_guides_pass
FAILED TestCheckSecurityStyle::test_real_docs_pass_security_style

README.md:25  — client-size claim `~55 KB` is outside the tolerance band [55.4, 61.4] KB (measured 58.4 KB, +/-3 KB)
README.md:933 — client-size claim `~55 KB` is outside the tolerance band [55.4, 61.4] KB (measured 58.4 KB, +/-3 KB)
```

Verified pre-existing: `git show origin/main:python/djust/static/djust/client.min.js | gzip -c | wc -c` → **58.0 KB**, against a README claiming `~55 KB` in two places. Reproduced on a clean `main` checkout with a stashed working tree.

## Why it matters more than a stale number

The check is doing its job — it caught real drift. But it is **red on main**, which means every PR inherits three failures it did not cause. That is the state that trains people to read a red suite as noise (the same argument as #2124).

## Likely cause

The bundle grew across the recent `dj-virtual` client work — `git log` on `client.min.js` shows #2114, #2115, #2120 landing in sequence, each adding a few hundred bytes gzipped, none updating the README. No single PR crossed the band; cumulatively they did.

## Fix

Update both README claims to the measured value, and decide whether the band should be tightened or the claim should reference a target rather than a measurement. Note CLAUDE.md separately states a `~87 KB gzipped` figure for `client.js` (unminified) and a `~37 KB` pre-minified target — worth reconciling all three numbers in one pass so they cannot drift apart again.

## Not fixed here because

Surfaced while draining #2129; scoped out per #1079 rather than folded into an unrelated streams fix.

johnrtipton:
`main` is **red**: three tests in `tests/test_check_doc_snippets.py` fail on a clean checkout, so every branch inherits failures it did not cause and **every push is blocked by the pre-push hook**. That is the state that trains people to read a red suite as noise — the same argument as #2124.

```
README.md:25  — client-size claim `~55 KB` outside the band [55.4, 61.4] KB (measured 58.4)
README.md:933 — same
```

The check is doing its job; the number drifted. `client.min.js.gz` grew across #2114, #2115 and #2120 — the `dj-virtual` client work — none of which crossed the band alone, and cumulatively they did.

## Reconciling all three figures, not just the checked one

| where | said | actual | describes |
|---|---|---|---|
| README ×2 | ~55 KB | **~58 KB** | `client.min.js.gz` — the shipped artifact |
| CLAUDE.md | ~87 KB gz, 388 KB raw, 35 modules | **~184 KB gz, 727 KB raw, 55 modules** | `client.js` — a build input |

Only the README one was checked, which is exactly how three numbers for "the client size" drifted apart. The CLAUDE.md pair had drifted much further, and it describes a *different artifact* — so the wording now says which number is which and which one is enforced.

Also dropped CLAUDE.md's "pre-minified distribution target for v0.6.0 is ~37 KB gzipped": minification shipped long ago, so that is neither a target nor accurate.

**Verified: 34/34 in `tests/test_check_doc_snippets.py`, from 31/34.**

Docs-only; no code paths touched.

Closes #2133

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -60,7 +60,7 @@ djust/
 │   ├── templatetags/       # Django template tags
 │   ├── tenants/            # Multi-tenant support
 │   ├── backends/           # Presence backends (memory, redis)
-│   └── static/djust/       # Client JS (~87 KB gzipped, 388 KB raw, 35 source modules)
+│   └── static/djust/       # Client JS (~58 KB gz minified; 184 KB gz / 727 KB raw unminified, 55 modules)
 ├── crates/
 │   ├── djust/              # PyO3 bindings (entry point for Python)
 │   ├── djust_core/         # Core types, serialization, context
@@ -89,7 +89,7 @@ djust/
 - **Error handling**: Use `Result` types; no `unwrap()` in library code
 
 ### JavaScript
-- Client JS size budget: current ~87 KB gzipped (388 KB raw across 35 source modules in `static/djust/src/`); pre-minified distribution target for v0.6.0 is ~37 KB gzipped / ~30 KB brotli. When adding a feature, measure its gzipped delta — aim under 2 KB gzipped per new module. Top 3 modules (`12-vdom-patch.js`, `09-event-binding.js`, `03-websocket.js`) are 42% of the budget; reducing them requires structural care. No new dependencies without discussion.
+- Client JS size budget: the SHIPPED artifact is `client.min.js.gz` at **~58 KB** — that is the number `scripts/check-doc-snippets.py` enforces against README (±3 KB band) and the only one a user downloads. Unminified `client.js` is ~184 KB gz / 727 KB raw across 55 modules in `static/djust/src/`; it is a build input, not a deliverable. (These read ~87 KB / 388 KB / 35 modules until #2133 — three stale numbers that had drifted apart because only the README one was checked.) When adding a feature, measure its gzipped delta — aim under 2 KB gzipped per new module. Top 3 modules (`12-vdom-patch.js`, `09-event-binding.js`, `03-websocket.js`) are 42% of the budget; reducing them requires structural care. No new dependencies without discussion.
 - **No `console.log`** without `if (globalThis.djustDebug)` guard — unguarded logging is auto-rejected
 - New JS feature files in `static/djust/src/` must have corresponding test files in `tests/js/`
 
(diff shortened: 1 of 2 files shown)


johnrtipton:
## Stage 14 retro — PR #2134 (#2133)

**Backfilled** by `/pipeline-retro` Stage 2, which flagged this PR as a retro-artifact gate violation. It is docs-only, but *not* a lockfile-only Dependabot bump, so the v1.0.0rc6 minimal-retro exemption does not apply — and there is genuinely something to retrospect here.

**Shipped**: corrected the client-size claim in `README.md` ×2 and reconciled two further stale figures in `CLAUDE.md`. `tests/test_check_doc_snippets.py` went 31/34 → **34/34**.

### This unblocked everyone, not just me

`main` was RED. Because the pre-push hook runs the full suite, **every** contributor's push was rejected — on branches with nothing to do with the failure. I hit it three times on three different branches before diagnosing it, and each diagnosis cost a ~5-minute hook run.

The failing test names are buried under ~40 passing hook lines, and nothing distinguishes "your branch broke this" from "this was already broken on the merge-base". That distinction is the only thing a contributor actually needs at that moment. Filed as **#2139**.

### Why it drifted: the check covered one of three copies

Three places state a client-JS size, and they describe **two different artifacts**:

| where | claimed | actual | artifact |
|---|---|---|---|
| README ×2 | ~55 KB | ~58 KB | `client.min.js.gz` — the shipped bundle |
| CLAUDE.md ×2 | ~87 KB gz, 388 KB raw, 35 modules | ~184 KB gz, 727 KB raw, 55 modules | `client.js` — an unminified build input |

Only the README pair was enforced. The unchecked pair had drifted **more than 2×** — which is the whole lesson: a drift check that covers one copy of a number does not protect the others, it *guarantees* they rot, because the checked one keeps everyone's attention.

The README drift itself was cumulative: #2114, #2115 and #2120 each added a few hundred bytes, none crossed the ±3 KB band alone. No single PR was at fault, which is exactly the shape a per-PR review cannot catch.

Filed as **#2138**, with a preference for *generating* the figures at build time over adding a second checker — two checkers is the same failure mode one step later.

### What I'd do differently

Nothing about the fix; it was three lines and correct. What I'd change is the sequencing: I hit the red `main` on my **first** blocked push and treated it as a local problem, retrying and then reaching for background tasks and refspec variations, before finally reading the hook output properly. The signal ("3 failed, N passed" naming tests I had not touched) was there the first time.

### Also surfaced

`client.min.js.gz` is gitignored (`.gitignore:70`), so the checker measures a **locally built** artifact and warns-and-skips on a fresh clone. That means CI and a contributor can disagree about whether the claim is in band. Noted in #2138 rather than fixed here.

`RETRO_COMPLETE`

PR #2135, opened 2026-07-26
feat(client): apply the [dj-virtual] keyed splice ops — ADR-026 iteration 2 (#2017)

johnrtipton:
Iteration 1 (PR #2126) taught the Rust differ to emit `VirtualInsert` / `VirtualMove` / `VirtualRemove` for a `[dj-virtual]` parent instead of index-addressed child ops. Nothing could apply them. This is that applier.

**Still dark** — the differ's flag defaults off, so nothing on `main` changes. Iteration 3 flips it after a soak (#1122).

## The ops land in the POOL, not the DOM

A `[dj-virtual]` container's children are only the visible **window**; off-window rows are detached and held in `state.items`. A row spliced into the DOM never enters the pool — not counted for the spacer height, not reachable by scrolling, dropped on the next render. `12-vdom-patch.js` now routes all three ops through `29-virtual-list.js`'s existing pool rather than adding a second path (#1646), and honours `before_key` instead of always appending, which is what #2017 item 4 lacked.

## Three constraints from iteration 1's review, each pinned

| constraint | why | pinned by |
|---|---|---|
| the ops share **one** sort phase and keep emitted order | `before_key` names an anchor an earlier op placed; splitting them into natural Remove/Move/Insert phases breaks it | a test that phase-sorts them and asserts the replay **diverges** |
| a key may be `dj-key` **or** `data-key` | the Rust parser reads both into `VNode.key`, while a list's own `keyAttr` defaults to `data-key` | a list carrying `dj-key` with **no** `dj-virtual-key-attr` |
| render **once per batch**, not per op | a per-op render re-slices the window and, in variable mode, recomputes every offset — O(items) per op | a MutationObserver asserting exactly 1 |

The flush hangs off a single `finally` wrapping the patch loop: that loop has four return paths plus a throw, and four call sites would drift apart.

## Two self-inflicted bugs the tests caught

1. **The insert arm called a function that does not exist.** I wrote `createElementFromVNode`; the builder is `createNodeFromVNode`. Every insert silently failed.
2. **My `itemKey()` redefined the variable-height cache's `itemKey()`.** The later function declaration wins for the whole module, so every offset computation started reading a bare key instead of a `'k:'`/`'i:'`-prefixed one. **Four existing `virtual_list` tests caught this; running only the new file would not have.** Renamed to `patchKeyOf`.

## Tests: 14 cases, `tests/js/virtual-keyed-ops-2017.test.js`

**Gate-off verified (#1468) per guard:**

| gated off | fails |
|---|---|
| patcher arms | 11 |
| `before_key` ignored (always tail) | 3 |
| natural Remove/Move/Insert phases | 1 |
| the flush | 1 |
| `dj-key` fallback | 1 |
| duplicate-key replace | 1 |

**Two of those tests were decorative until gate-off exposed them**, and both were my error:

- Every assertion read the **pool**, which the ops mutate directly — so the render was never required, and an upper-bound-only assertion (`<= 1`) was satisfied by *zero* renders. Removing the flush entirely failed nothing. Now asserts the visible window reflects the change and pins the count from both sides.
- The `dj-key` case set `dj-virtual-key-attr="dj-key"`, which makes the **first** lookup succeed and the fallback dead code. Now uses a list that carries `dj-key` without declaring it — the realistic template shape.

Full JS suite: **1899 passed**. Bundle: **+503 bytes gzipped**.

Refs #2017 — does not close it. Iteration 3 (flip the flag, wire the Python config) remains, deliberately soak-gated.


johnrtipton:
## Stage 14 retro — PR #2135 (#2017, ADR-026 iteration 2)

**Shipped**: the client applies `VirtualInsert`/`VirtualMove`/`VirtualRemove` through the virtual list's item pool. Still dark. 18 cases, 10 gate-offs.

### The differential is the result worth keeping

Iteration 1 put the reconciler in Rust; iteration 2 put the applier in JS. That's **one protocol implemented twice in two languages, with each side's tests passing against its own model**. Nothing tested them against each other, and neither suite could have caught a divergence.

The review closed that: it dumped **1056 real differ outputs** (iteration 1's exhaustive 4-key sweep plus 17 named shapes) and replayed each through the real client. **1056/1056 land on exactly the server's key order**, with a gate-off proving the harness non-vacuous (forcing every insert to the tail fails 850).

One fidelity detail decided whether that meant anything: iteration 1's Rust helper builds rows with `key` set but **no `data-key` in `attrs`**. Fine for a string replay, useless against a client that reads keys off DOM attributes. Building rows the way `parser.rs` actually does is what made the differential real rather than a second tautology.

**The durable lesson**: when a feature spans a language boundary, the cross-boundary differential is not optional extra coverage — it is the only test of the thing that is actually new.

### Two bugs my own tests caught, and one they didn't

- I called `createElementFromVNode`, which does not exist. Every insert silently failed.
- My `itemKey()` **redefined the variable-height cache's `itemKey()`** — later function declaration wins module-wide — so every offset computation started reading a bare key. **Four existing `virtual_list` tests caught this; running only my new file would not have.** The habit of running the full suite is what saved it.
- What they didn't catch: a test that was green *while destroying what it guarded*. `leaves an ordinary keyed list untouched` asserted a `SetText` left `poolKeys` unchanged — which no `SetText` can ever change, since the pool holds node references. It survived all nine gate-offs, and the patch it fired resolved to the virtualization shell and **wiped the entire rendered window** (4 children → 0) while passing.

That last one is the sharpest tautology I've shipped: not merely vacuous, but actively concealing damage.

### On testing an unreachable line honestly

The `nodeType` guard makes the flush's `finally` unreachable through `applyPatches`, so gating it off failed nothing. Two honest options: label it unpinned, or find a legitimate way in. I reached the exported pool seam directly — and asked the reviewer whether that was legitimate or just line-coverage. Its answer is the useful artifact: it's the *production* seam (same three-argument shape, one call site), bypassing an outer guard is the only way to test an inner layer, and it's paired with a test of the outer guard through the public path. Neither stands in for the other.

I also noted the one brittle coupling in the test's own comment — the throw comes from `renderFixed` setting `.style.height` — with an instruction to inject a deterministic throw rather than delete the test if that line changes.

### Measurement discipline

My first gate-off of the `DIRTY.clear()` change reported **16 failures**. It had left an orphan `try {` — it was measuring a syntax error, not behaviour. Every mutation in the table now asserts it applied and changed something before running. A gate-off that doesn't gate is a tautological test one level up, and I'd hit the same failure mode earlier in this session on the Python side.

### Carried forward

**#2136** — content patches for surviving rows are still index-addressed. Measured: editing a row after a scroll **silently rewrites a different row**, `applyPatches` returns `true`, no warning. Pre-existing (iteration 1 emits them, dark) but **blocking for iteration 3**, and the earlier gap record named only `InsertSubtree`/`MoveSubtree`. Stated now rather than discovered when the flag flips.

`RETRO_COMPLETE`

PR #2141, opened 2026-07-26
fix(forms): validate_field accepts the key the client actually sends (#2137)

### LINKED ISSUE #2137: FormMixin.validate_field expects `field_name` but the client sends `field` — handlers silently no-op
## Summary

`djust.forms.FormMixin.validate_field` accepts `field_name`, but the client sends the payload key `field`. Because the signature also takes `**kwargs`, the mismatch is swallowed rather than raising `TypeError` — so a user who overrides `validate_field` using the framework's own documented signature gets a handler that runs on every keystroke and silently never receives the field.

No error. No warning. The event fires, the handler executes, and nothing happens.

Found in djust **1.1.0rc8**.

## Evidence

**Server side** — `python/djust/forms.py:168`:

```python
def validate_field(self, field_name: str = "", value: Any = None, **kwargs: Any) -> None:
```

**Client side** — `python/djust/static/djust/src/09-event-binding.js`, three send sites:

```
527:    const params = { value, field: fieldName };
977:        field: fieldName
1774:        const params = { value: value, field: fieldName };
```

**No normalization anywhere between them.** I grepped `python/djust/forms.py` for any `field` → `field_name` remapping and found none.

`python/djust/static/djust/src/20-model-binding.js:27-28` documents the same `field` key for `dj-model`, so the wire contract is consistently `field` on the client side — it is the Python signature that diverges.

## Reproduction

```python
from djust import LiveView, event_handler
from djust.forms import FormMixin

class SignupView(FormMixin, LiveView):
    form_class = SignupForm

    @event_handler
    def validate_field(self, field_name: str = "", value=None, **kwargs):
        # field_name is ALWAYS "" — the client sent it as `field`
        return super().validate_field(field_name=field_name, value=value, **kwargs)
```

```html
<input dj-input="validate_field" name="email">
```

Type in the field: the handler is invoked, `field_name` is empty, no validation occurs, no patches are produced. The form appears live and is inert.

## Why this is worth fixing rather than documenting

The failure is silent and looks like a page-wiring problem, not a signature problem — which sends you hunting in exactly the wrong place. It cost a full debugging session on djust.org to isolate: we first suspected a missing `dj-root`, then the template context, before finding this. The `**kwargs` in the signature is what converts a loud `TypeError` into a silent no-op.

## Working adaptor

For anyone hitting this now — take `field` and forward it:

```python
@event_handler
def validate_field(self, field: str = "", value=None, **kwargs):
    return super().validate_field(field_name=field, value=value, **kwargs)
```

A live example is at `marketing/proto/signup.py` on the `proto/live-split-hero` branch of `djust-org/djust.org` (see djust-org/djust.org#54).

## Suggested fix

Most backward-compatible option is to accept both on the server:

```python
def validate_field(self, field_name: str = "", value: Any = None, **kwargs: Any) -> None:
    field_name = field_name or kwargs.pop("field", "")
```

Alternatives: rename the client key to `field_name` (a wire-contract break), or document `field` explicitly as the payload key and change the mixin signature to match.

Worth also auditing whether `submit_form` and the `dj-model` bindings have the same split.

johnrtipton:
The client sends the field name under `field`. The Python signature took `field_name` **exclusively**. `**kwargs` swallowed the mismatch instead of raising `TypeError`, so the handler ran on every keystroke doing nothing.

**No error, no warning. The form looked live and was inert.** Reported against 1.1.0rc8.

## Reproduced before fixing

```
validate_field(field="email",      value="not-an-email")  ->  form_data={}  field_errors={}
validate_field(field_name="email", value="not-an-email")  ->  form_data={'email': ...}  field_errors={'email': [...]}
```

The client is consistent: `09-event-binding.js` sends `field` at all three send sites (`:527`, `:977`, `:1774`), and `20-model-binding.js` documents the same key for `dj-model`, because djust maps `data-*` attributes to handler parameters. The Python signature is the divergent one, and there is no remap between them (grepped `forms.py`, `websocket.py`, `runtime.py`).

## The fix already existed in this repo

`WizardMixin.validate_field` carries exactly this coalesce — `name = field or field_name` — with a docstring explaining the wire contract. Two of the three implementations never got it:

| implementation | before |
|---|---|
| `WizardMixin.validate_field` | ✅ accepts both |
| `FormMixin.validate_field` | ❌ `field_name` only |
| `AdminFormMixin.validate_field` | ❌ `field_name` only |

Three implementations of one wire contract, one correct — the #1646 shape. So this **lifts the working one** (#1077) rather than inventing a third spelling. `field` wins when both are supplied: it is the contract, `field_name` is the compatibility alias, and that matches `WizardMixin`'s resolution order.

## A shadow the change itself introduced

Adding a `field` parameter made the existing local `field = form.fields.get(field_name)` rebind it. mypy caught it as `"str" has no attribute "clean"`.

Latent rather than broken — nothing downstream read the parameter — but a later edit that did would have got a form field where it expected a name. Renamed to `form_field` in both files, with a note saying why.

## Tests: 9 cases, `tests/unit/test_validate_field_wire_contract_2137.py`

They assert what the **caller observes** — the field was validated, the error appeared, a valid value cleared it — rather than which parameter carried the name, so they survive a future rename.

Two are structural pins against the drift itself: every `validate_field` in the tree accepts `field`, and each one actually **coalesces** rather than accepting the parameter and ignoring it (which would satisfy the first pin while leaving the bug in place).

**Gate-off verified (#1468)**, every mutation asserting it applied *and* changed something:

| gated off | fails |
|---|---|
| `field` param dropped from `FormMixin` | 7 |
| `FormMixin` coalesce removed | 5 |
| resolution order flipped (`field_name` wins) | 2 |
| `admin_ext` coalesce removed | 1 |

Full suite: **10022 passed, 218 skipped, 0 failed**. mypy clean, 880 files.

Closes #2137


johnrtipton:
## Stage 14 retro — PR #2141 (#2137)

**Shipped**: `FormMixin.validate_field` and `AdminFormMixin.validate_field` accept the key the client actually sends. 11 cases, 7 gate-offs. Two review rounds, one 🔴.

### The review attacked the premise first, and that was the right order

I asked it to check one thing before anything else: *if the dispatch path already remaps `field` → `field_name` somewhere my grep missed, this PR is fixing something only reachable by calling the method directly, and the framing is wrong.*

It built a real `WebsocketCommunicator` harness rather than trusting my direct-call tests, sent the exact payload `buildFormEventParams` builds, and got back `{'type': 'noop', ...}` from `origin/main` — **which is the reported symptom**. Then it traced `validate_handler_params` and `_call_handler` to confirm no remap exists.

That is a much stronger foundation than I had, and it cost one harness. **For any bug whose fix depends on "no layer between A and B does X", build the A→B harness before building the fix.**

### My structural pin was decorative in both directions

`test_the_coalesce_is_present_not_just_the_parameter` grepped the source for the literal `field or field_name`. The reviewer falsified it twice:

- **False negative** — sabotage `AdminFormMixin` while leaving the literal in place: **0 failed**.
- **False positive** — a behaviour-preserving ternary refactor: **2 failed**.

And it would have blocked the obvious improvement — extracting the coalesce into one shared helper, which is the #1646 cure my own rationale praised. A pin that blocks the fix it advocates is worse than no pin.

Replaced with behavioural tests. The probes now read **1** and **0**.

### The fix for the decorative pin was itself mis-targeted

The replacement test asserting the adapter positional shape instantiated a **`FormMixin`** subclass. But `AdminFormMixin.validate_field` is a full reimplementation, not a `super()` call, and the 8 adapter sites are served by `ModelDetailView(AdminBaseMixin, AdminFormMixin, LiveView)`.

So the test passed without touching the class it named — and that test **is the evidence for keeping `field` first**. Arguing "the admin adapters need this order" while asserting it against a different class is the same shape as the pin it replaced.

Caught by the reviewer's discrimination probe: sabotage one class at a time and check *which* tests fail, rather than *how many*. **A count can hide a test binding to the wrong class via MRO; a per-class failure list cannot.**

### Two claims I asserted without tracing

- **The mechanism.** "djust maps `data-*` attributes to handler parameters" — true of `extractTypedParams` on the click/poll paths, **false** of the form-event path, where `buildFormEventParams` hardcodes the key. I inherited the sentence verbatim from `wizard.py`, where it is equally wrong, and propagated it into a new docstring *and* the CHANGELOG. Both corrected.
- **The scope.** I fixed the code and left `docs/website/guides/forms.md`, which documents `field_name` as the wire shape — and is where the reported override came from. A signature that merely *tolerates* the wrong doc keeps producing the same override.

### An unreported win the review found

`admin_ext/adapters.py` emits `validate_field('<name>', value)` at 8 sites, where the bare `value` token arrives as the literal **string** `"value"`. Pre-PR that overwrote the real value in slot 1, so **djust admin validated the string "value" on every keystroke**. With `field` first, the junk lands in `field_name` and the real value survives.

That decided the ordering question: 8 real in-repo sites beat a 2-arg positional shape with no caller in code or docs.

### Carried forward

- **#2144** — the root cause. `**kwargs` converts a param-name mismatch into silence, and every handler using the documented shape is exposed. `validate_handler_params` already computes the unexpected-key set and discards it; a warning there would have made this a one-line log on the first keystroke.
- **#2145** — `frameworks.py` emits `data-field_name=` on every rendered field and nothing can read it. Harmless, but it is the artifact that makes the false mechanism *look* true — the docstrings patch the claim, removing it closes the loop.

`RETRO_COMPLETE`

PR #2146, opened 2026-07-26
fix(vdom): key-address [dj-virtual] content updates so they land on the right row (#2136)

### LINKED ISSUE #2136: dj-virtual: content patches for surviving rows are still index-addressed — silently rewrites the wrong row (blocks ADR-026 iteration 3)
Found by the Stage 11 review of PR #2135 (#2017 iteration 2). **Pre-existing** — emitted by iteration 1 (PR #2126, dark) and the same shape exists on the ordinary `reconcile_keyed` path.

## The defect

`reconcile_virtual_keyed` (`crates/djust_vdom/src/diff.rs`) recurses into surviving rows via `diff_node_into(old, new, child_path)`, where `child_path` is the parent path plus the item's **absolute index**. For a `[dj-virtual]` container that index is meaningless on the client: the DOM children are only the visible window.

The emitted patch carries **no dj-id** — text nodes don't have one — so `getNodeByPath` resolves it purely positionally, and the #2113 detached-item guard is gated on `if (djustId && …)` so it cannot fire.

Real differ output for "row `k9`'s text changed" in a 10-row list:

```json
[{"type":"SetText","path":[9,0],"text":"EDITED"}]
```

Measured against a real mounted list:

| patch | outcome |
|---|---|
| `SetText path=[0,0,0]`, unscrolled | correct row — **by coincidence** (shell child 0 *is* item 0 only when the window starts at 0) |
| same patch, after `scrollTop=200` on 20 rows | **silently rewrites `k7` instead of `k0`**; `applyPatches` returns `true`, zero warnings |
| `SetText path=[0,1,0]` | hits the **spacer**, misses, returns `false` |
| `SetText path=[0,9,0]` (off-window) | misses, returns `false` |

The first row of that table is the dangerous one: silent, wrong-node corruption that only appears once the user scrolls.

## Why it blocks iteration 3

Iterations 1 and 2 ship dark. Iteration 3 flips `set_virtual_keyed_ops` on, at which point every `[dj-virtual]` list with changing row content hits this. The structural ops are key-addressed and correct; content patches are the remaining half.

## Fix options

1. **Key-address content patches too** — a `VirtualUpdate(key, …)` op, or route the recursion through the pool so the client resolves by key. This is what ADR-026's Option A originally sketched for item 3.
2. **Give the emitted patches a dj-id** so the existing id-first resolution works, and extend the #2113 guard to cover the no-id case.

Option 1 is the coherent one — it makes the whole `[dj-virtual]` patch surface key-addressed rather than half.

## Related
- #2017 — the parent issue; this is item 3, which iterations 1 and 2 explicitly did not deliver
- `docs/adr/026-dj-virtual-differ-awareness.md`
- #2113 / #2115 — the detached-item guard that cannot see this
- PR #2126 (iteration 1), PR #2135 (iteration 2)

johnrtipton:
`reconcile_virtual_keyed` recursed into surviving rows with a child path built from the item's **absolute index**. For a windowed container that index is meaningless — it counts *items* while the DOM holds only the visible *window* — and the emitted patch carries no dj-id (text nodes have none), so it resolved purely positionally.

Measured against a real mounted list: **editing row `k0` after a scroll silently rewrote `k7`**, `applyPatches` returned `true`, no warning. An off-window row's update could not land at all.

## The fix ADR-026 Option A originally sketched

```rust
VirtualUpdate { path, d, key, patches }   // patches relative to the ROW's root
```

The row is diffed against an **empty base path**, so `path: []` *is* the row and `[0]` is its first child. The client resolves the row **by key** in the item pool, then applies the inner patches with that node as the root.

That's also what makes **#2017 item 3** work: an off-window row is detached, held only in `state.items` — unreachable by path, findable by key — and mutating a detached node is fine, the change appears when it scrolls back.

Iterations 1 and 2 shipped the structural ops only; this is the half that was missing.

## The old test is rewritten, not deleted

`a_content_patch_targets_the_right_child_index` pinned the absolute-index arithmetic — which is exactly what this makes unnecessary. It now asserts the new invariant: **no path-addressed content op for a virtual parent at all**, and inner paths relative to the row. The `dj-if` boundary fixture is kept deliberately: under the old scheme it was the shape that diverged, and under this one it must simply not matter.

## One line removed for being unpinnable

The patcher arm originally marked the list dirty after applying inner patches. **Gate-off failed 0, twice.** An inner patch mutates the row *node*, so an attached row already shows the change; an off-window row is re-read from the pool on scroll; and in variable-height mode a height change is already caught by the ResizeObserver (`29-virtual-list.js:409`).

I then wrote a variable-height test specifically to justify it, and **that didn't discriminate either** — JSDOM does no layout, so measured heights are all 0.

Rather than ship an unpinned line with a plausible-sounding comment, it's gone, with a note saying to add the mark *and* the test together if a real case turns up.

## Tests: 21 Rust + 23 JS

| gated off | fails |
|---|---|
| client `VirtualUpdate` arm removed | 5 (js) |
| client row lookup ignores the key | 3 (js) |
| differ back to path-addressed content | 2 (rust) |

Every mutation asserts it applied *and* changed something, and the harness refuses to report a number when the build broke.

The JS side covers the measured symptom directly: an off-window row (index 40 of 50, verified `document.contains === false`) receiving its update, and editing `k0` after a scroll leaving `k7` untouched.

Full suite: **10024 Python + 1908 JS**, 26 Rust binaries green, clippy clean. Bundle **+137 bytes gzipped**.

Closes #2136 — and unblocks ADR-026 iteration 3, which remains soak-gated per #1122.


johnrtipton:
## Stage 14 retro — PR #2146 (#2136)

**Shipped**: `VirtualUpdate` — key-addressed content updates for `[dj-virtual]` rows, with inner patches relative to the row. 22 Rust cases + 31 JS, 6 gate-offs. Two review rounds, one 🔴. Closes the last blocker for ADR-026 iteration 3.

### The same shape, a third time: two implementations of one op, both suites green

A content diff emits `Replace { path: [] }` when a row changes tag. The Rust simulator had always handled it correctly (`*target = node.clone()` — swap the pool slot). The client reached for `node.parentNode`, which is `null` off-window and the **shell** in-window.

The in-window case is the one that matters: `applyPatches` returned `true`, no warning, the window showed the new tag — and the change **silently reverted** the moment the row scrolled out and back, because `render()` re-appends the pool node. That is the exact symptom #2136 exists to remove, reproduced inside its own fix.

26 green Rust binaries and 23 green JS cases could not see it. **Only replaying real differ output through the real client did** — 16/18 exact, and both failures were this, in-window and off-window.

This arc has now produced that shape three times (#2135's structural ops, this, and the differential that found it). The durable form: **when one protocol has two implementations, each suite verifies its own half against its own model, and agreement between them is the thing nothing tests.**

### My tests exercised one variant of six

Inner patches were tested only as `SetText path:[0], d:null`. `diff_node_into` also emits `SetAttr`, `RemoveAttr`, `InsertChild`, `RemoveChild` and `Replace` — and the uncovered one is the one that broke. That is #1543 for the fourth time this session, and I keep writing it *after* invoking the rule. Enumerating the variants of the surface takes a minute; finding the missing one takes a review round.

### A test that stayed green under the gate written to catch it

The scroll test targeted `k0`. The key-ignoring gate-off returns `items[0]` — the same row — so it passed under the exact mutation meant to falsify it. Only a **per-test failure list** surfaced that; a count would not have. Now targets `k3`.

### And the fix for the 🔴 repeated the 🔴's own class

The two lines making the pool swap visible (`invalidateWindow` + `DIRTY.add`) had no test: 0 of 30 cases and 18 of 18 differential cases green without them, while an in-window row that changes tag would simply never appear. **The #2135 shape, inside the branch I added to fix a defect of the same family.**

### A wrong reason I gave for a right decision

I removed an unpinnable dirty-mark from the *content* path — correct — and justified it partly with *"JSDOM does no layout, so measured heights are all 0."* The reviewer falsified that with a driven ResizeObserver stub (#1830) that discriminated cleanly, and used the same technique to pin the Replace branch's re-render.

The decision stands; the reason was wrong, and a wrong reason in a comment is worse than none — it discourages the next person from trying. Corrected at the source rather than only in the CHANGELOG.

### What worked

- **Asking the reviewer to attack the seam I was least sure of.** I flagged that a swapped pool node might miss initialization `render()` gives a fresh one. It built a drivable `ResizeObserver` stub and ran nine adversarial cases — Replace+Move same batch, Replace with a different key, off-window then scroll-in, variable-height re-observation. All sound, and for a structural reason worth knowing: `virtualReplaceByKey` marks dirty → flush → `render()`, which is the single place doing all per-node init. Reusing the existing seam is why a replacement is indistinguishable from an insert downstream.
- **`gzip -n -9`, not `gzip.compress`.** My bundle figure was off by one byte because Python's default embeds an mtime.

### Carried forward

ADR-026 **iteration 3** (flip `set_virtual_keyed_ops`) is now unblocked but remains soak-gated per #1122. The differ, the applier and the content ops are all in place and dark.

`RETRO_COMPLETE`

PR #2147, opened 2026-07-27
fix(build): generate the client-size figures instead of asserting them (#2138)

### LINKED ISSUE #2138: tech-debt: one source of truth for the client-size figures — three copies, one checked
Source: v1.1.0-13 retro finding 6 / PR #2134 (#2133)

## The problem

Three places state a client-JS size, they describe **two different artifacts**, and only one of them was checked:

| where | claimed (before #2134) | actual | artifact |
|---|---|---|---|
| `README.md` ×2 | ~55 KB | ~58 KB | `client.min.js.gz` — the shipped bundle |
| `CLAUDE.md` ×2 | ~87 KB gz, 388 KB raw, 35 modules | ~184 KB gz, 727 KB raw, 55 modules | `client.js` — an unminified build input |

`scripts/check-doc-snippets.py` enforces the README pair against a ±3 KB band. Nothing checked the CLAUDE.md pair, which is exactly why it drifted more than 2×.

## Why it bit

The README number drifted across #2114/#2115/#2120 — none crossed the band alone, cumulatively they did — and the resulting three failing tests made `main` RED, so the pre-push hook rejected **every** branch until #2134 landed.

## What a fix looks like

Options, roughly in order of preference:

1. **Generate rather than assert.** Emit the measured figures into a small generated file (or a doc include) at `build-client.sh` time, so prose cites one computed value and drift is structurally impossible.
2. **Extend the checker** to the CLAUDE.md figures too, with per-artifact bands, so all four claims are enforced.
3. At minimum, a comment at each site naming which artifact it describes and pointing at the checker — #2134 did this by hand, but nothing keeps it true.

Note `client.min.js.gz` is **gitignored** (`.gitignore:70`), so the check depends on a local build and warns-and-skips on a fresh clone. Worth deciding whether that is intended.

## Related
- #2133 / PR #2134 — the fix that unblocked `main`
- #1500 — the doc-snippet checker

johnrtipton:
Three places in prose quoted a client size, describing **two different artifacts**, and only one was checked.

| where | claimed | actual | artifact |
|---|---|---|---|
| README ×2 | ~55 → ~58 KB | 58.7 KB | `client.min.js.gz` — what a user downloads |
| CLAUDE.md ×2 | ~87 KB gz, 388 KB raw, 35 modules | **188 KB gz, 745 KB raw, 55 modules** | `client.js` — a build *input* |

The unchecked pair had drifted **more than 2×**, precisely because nothing looked at it. The checked pair drifted across #2114/#2115/#2120 — none crossing the ±3 KB band alone — and the result was three failing tests on `main` that made the pre-push hook reject **every** branch until #2133 landed.

## Adding a second checker would be the same failure mode one step later

**1. The build generates the numbers.** `scripts/build-client.sh` writes `client-sizes.json` — one measured source, both artifacts named, their difference stated in the file itself.

It is **committed**, deliberately. The `.gz` files are gitignored (`.gitignore:70`), so without a checked-in manifest the check measures a *locally-built* artifact and warns-and-skips on a fresh clone — CI and a contributor could disagree about whether a claim is in band. That's the quiet half of this bug, and a test pins that the manifest stays tracked.

**2. The checker resolves the artifact per line, not per file.** A doc legitimately cites both — *"the shipped bundle is ~58 KB, the unminified input is ~188 KB"* — so a per-file mapping cannot work. The line names the artifact; the checker matches on that, and now reads `CLAUDE.md` as well as README.

It also grew an opt-in `<!-- size-claim: historical -->` marker, because prose sometimes quotes a figure *precisely to say it was wrong*. Without an escape hatch the only way to keep such a sentence is to delete the number, which loses the point of the sentence. The marker is per-line and opt-in, so it can't defeat the check — pinned by a test that an *unmarked* stale figure is still caught.

## The prose got more accurate as a side effect

That's the real win. README's architecture diagram said `client.js (~58 KB gz)` — but the shipped artifact is `client.min.js.gz`, and `client.js` is 188 KB. It was **mislabelled, not merely stale**, and the per-line rule is what forced that out.

`make sizes` prints the current figures, so updating prose is mechanical rather than a measurement exercise:

```
  shipped   client.min.js.gz   58.7 KB  <- quote THIS in README
  build in  client.js (gz)     188.1 KB  across 55 modules
  build in  client.js (raw)    745 KB
```

## Tests: 9 cases, `tests/test_client_size_manifest_2138.py`

Covering that the manifest is **tracked** (the point of the exercise), that it **matches the files on disk** (a stale manifest is worse than none — it would authorise a wrong claim), and that `58 KB` resolves as *correct* for the shipped bundle and *wrong* for the unminified input on adjacent lines.

**Gate-off verified (#1468)**, every mutation asserting it applied:

| gated off | fails |
|---|---|
| per-line artifact resolution | 3 |
| historical marker honoured unconditionally | 2 |
| `CLAUDE.md` dropped from the checked set | 2 |
| manifest module count hardcoded wrong | 1 |

Full suite: **10033 passed, 218 skipped, 0 failed**.

Closes #2138


johnrtipton:
## Stage 14 retro — PR #2147 (#2138)

**Shipped**: a generated, committed `client-sizes.json` as the one measured source; per-line explicit artifact markers; the check widened to lines *about the client*; 10 wrong claims corrected across 9 files, every one of those files now guarded. 15 cases. Two review rounds, three 🔴.

### Two of the three 🔴s were this fix reproducing the bug it exists to remove

That is the finding worth keeping, and it is not a coincidence — it is what happens when you build a gate without asking what the gate does when *it* fails.

- **The build wrote a zeroed manifest** when run without terser, a path the script explicitly documents as supported. Valid JSON, exit 0, no warning. The checker trusted it, computed a band of `[-3, +3] KB`, and rejected **every correct claim**. Four blocked claims where #2133 had three — reachable by running the command `make sizes` tells you to run. And the error message would have told a contributor to mark four *correct* figures as historical.
- **The check gated on the word "gz"**, so the most-wrong claims in the repo — `~5KB` for a 58.7 KB bundle, nine copies — were precisely the ones it could not see. An unchecked claim drifting because nothing looks at it is the entire subject of #2138.

**A new gate needs its own failure modes enumerated before it ships.** What does it do with a missing input, a partial measurement, a claim it cannot classify? Each of those was a wall or a blind spot here.

### A resolver whose main branch could never fire

`"minified"` is a substring of `"unminified"`, and my "last hint wins" rule took the maximum position — so the shipped mapping always won and the `unminified` mapping was **dead config**. It also mis-resolved realistic prose in both directions.

The real docs passed only by accident of word order. So did my test asserting *"a correct unminified claim passes"* — it was green for the same accident, which means I shipped a test that could not distinguish working from broken.

The fix was to stop guessing: **explicit markers, default shipped.** Prose is not a reliable place to infer intent from, and a heuristic over prose is a heuristic you cannot enumerate the inputs of.

### Correcting a claim without guarding its file

I corrected nine claims and added two files to the checked set. The other seven were free to drift straight back — the same *"N copies, only some checked"* class, at smaller scale, inside the fix for it. Verified: re-introducing `~5KB` into `docs/llms.txt` was exit 0 before the widening.

Guarding them immediately surfaced a **tenth** claim the review had flagged. That is the argument for guarding over correcting: the guard finds what the sweep missed.

### The CHANGELOG was falsifiably wrong

I wrote "five more copies" and "six claims"; `git diff` showed nine across eight files. And the entry implied the sweep was complete when ~19 copies remained.

On a PR about prose figures drifting because nobody counted them. The v1.0.8-1 rule — *falsification-test the prose invariant* — pointed at this entry's own text. Corrected, with the remainder filed as #2148 rather than asserted away.

### What worked

- **Generating rather than checking.** The issue offered "extend the checker"; that is the same failure mode one step later. Emitting the numbers removes the hand-typed source.
- **Committing the manifest.** The `.gz` artifacts are gitignored, so without it CI and a contributor measure different things — the quiet half of the original bug.
- **The per-line rule forced the prose to be accurate.** README's architecture diagram said `client.js (~58 KB gz)` when `client.js` is 188 KB — **mislabelled, not merely stale**. The number was right for a file it wasn't describing, and only requiring each line to name its artifact surfaced that.

### Carried forward

**#2148** — ~19 remaining `~5KB` copies (mostly `examples/` demo copy), exact-string marker matching where 5 of 6 spellings fall back (loudly), and the two build guards masking each other under individual gate-off.

`RETRO_COMPLETE`

Issue #2142, opened 2026-07-26
tech-debt: chain-shaped drains need an estimation convention


johnrtipton:
Source: v1.1.0-12b retro finding 1 (backfilled under #2140)

## The shape

Two chains in that group were never planned as sequences — each PR's adversarial review surfaced the *next* defect:

- **scoped-attr**: #2107 (scan the root element) → #2109 (refresh registry entries on value change) → #2111 (sweep and scan scopes still disagreed)
- **streams**: #2113/#2115 → #2112/#2117 → #2116/#2118 → #2119/#2122

The same shape recurred in v1.1.0-13: #2129 took five rounds, each finding the same class in a variant the previous fix did not reach.

## Why it needs a convention

Each link shipped as its own PR because each was a real, separately-reachable bug — batching them would have merged four incomplete diagnoses. That is the right call. But a chain discovered one link at a time **cannot be estimated up front**, and the drain plan presents it as one item, which makes every chain look like a slipped estimate.

## What a convention might look like

- Mark an issue as *chain-suspected* when its root cause is a class rather than an instance (parallel-path drift, wire-contract mismatch, shared-key collapse), and state up front that follow-ups are expected rather than exceptional.
- When a review surfaces link N+1, file it immediately with a back-reference, so the chain is visible in the issue graph rather than only in the PR trail.
- In the milestone retro, count chains as one *finding* and N *PRs*, so the stats do not read as five failures.

Not urgent; the cost today is narrative, not correctness.

## Related
- v1.1.0-12b retro, finding 1
- v1.1.0-13 retro, finding 1 (non-convergence as a signal)

PR #2149, opened 2026-07-27
feat(ci): say whose failures blocked the push, and find a red main first (#2139)

### LINKED ISSUE #2139: tech-debt: detect a red main proactively instead of discovering it via a failed push
Source: v1.1.0-13 retro finding 7 / PR #2134 (#2133)

## The problem

`main` had three failing tests in `tests/test_check_doc_snippets.py`. Because the pre-push hook runs the full suite, **every contributor's push was rejected** — on branches that had nothing to do with the failure.

Nothing alerted. The state was discovered only because a push failed, and it took three separate failed pushes to diagnose (the hook output buries the failing test names under ~40 passing hook lines).

## Why it matters beyond the inconvenience

This is the #2124 argument: a persistently red suite trains people to read red as noise, and to reach for `--no-verify`. The cost is not the minutes lost, it is the habit.

## What a fix looks like

- A scheduled CI run on `main` (nightly or post-merge) that opens/updates an issue when the suite goes red, so the first person to notice is not someone whose unrelated push just bounced.
- And/or make the pre-push hook surface the failing test names prominently, and say plainly when the failures are **pre-existing on the merge-base** rather than caused by the branch being pushed. That distinction is the one a contributor needs and currently has to derive by hand.

## Related
- #2133 — the instance
- #2124 — the same argument about flaky tests training reviewers to ignore red

johnrtipton:
Closes #2139.

`main` carried three failing doc-snippet tests until #2134. Because the pre-push hook runs the full suite, **every** contributor's push was rejected — on branches with nothing to do with the failure. Nothing alerted, the failing names sit under ~40 passing hook lines, and it took three failed pushes to diagnose. The cost isn't the minutes; it's that a persistently red suite trains people to read red as noise and reach for `--no-verify` (#2124).

Two deliverables: `scripts/pre-push-pytest.sh` re-runs **only** the failing tests against the merge-base and says which were already broken; `.github/workflows/main-health.yml` finds a red main daily and opens/updates/closes one tracking issue.

## Four review rounds, twelve defects — and this is the interesting part

Every round fixed the cited instance and **re-created the class one step over**. That pattern is now written up as the chain-shaped-issue convention in `CLAUDE.md` (#2142), with this PR as the worked example.

| round | defects | of which caused by the previous round's fix |
|---|---|---|
| 1 | 3 | — |
| 2 | 3 | 1 |
| 3 | 3 | 2 |
| 4 | 3 | 2 |

Selected: round 2 fixed ids containing ` - ` with `sed 's/ - [^-]*$//'`, which only substitutes when the *message* has no hyphen — false for most real pytest messages, so ids kept the message glued on and were announced as **new**. Round 3 fixed the script's hardcoded `.venv` path, then left the identical hardcode in the test file that verifies the fix, so 19 behavioural cases silently **skipped** in a worktree. Round 3 fixed the report arms being gated on the wrong count, and the surviving arm then issued its global verdict unconditionally.

## What it does now

- **Ids** split at the first ` - ` at bracket depth 0 — exact, not heuristic, since a node id can only contain ` - ` inside `[params]`.
- **Classification by pytest's exit code**, not by grepping output text: `0` passed → new, `1` failed-or-errored → pre-existing, `4`/`5` unresolvable → absent, `2`/`3` → cannot tell. Grepping had no arm for a test that *errored*, and `not found` is not exclusive to "absent" (`pytest.fail("config file not found")` matched), so a test broken in **both** trees was called new.
- An id absent at the merge-base is **confirmed collectible in the current tree** before being called new — it just failed here, so if it can't be collected here the parse is what's broken, and both parsing bugs ended by blaming the pusher for this script's own defect.
- **No claim exceeds what was checked.** The global "your branch did not cause them — main is red" only prints when nothing is unresolved or capped; otherwise it's scoped, and the arm lists the ids it covers.
- Interpreter resolved via `run-with-venv-python.sh` on **both** pytest runs.
- Per-id loop capped at 40, and the cap is disclosed — including that sorted ids make it the alphabetically-first 40, not the most relevant.

The workflow reports three states (green / red / **could not run**), skips cancelled runs, no longer swallows a failed Rust build, treats a missing pytest summary as "cannot tell", paginates the issue lookup, ignores PRs sharing the title, and closes whichever of its two issues went stale — previously a "could not run" issue could never be closed, because the title it searched for was computed from the current run's state.

## Verification

- **52 cases** across three files. `tests/test_main_health_workflow_2139.py` extracts the workflow's JavaScript and **executes** it under stubs — removing `if: always()`, collapsing the three states, or matching on label instead of title had all left the suite green.
- Gate-off: **14 of 14** mutations fail at least one test, with the baseline asserted clean-green first. Inverting the partition fails 14; deleting the merge-base run fails 11; reverting the parse fails 21.
- Full suite: **10,142 passed**, 218 skipped. Rebased, 0 behind `origin/main`.

Two harness traps worth recording, since both would have produced a table of false zeros: a baseline that silently collected nothing (17 rows read 0), and a mutation that prepended the old `sed` *ahead* of the new `awk`, which the awk then corrected.


johnrtipton:
## Stage 14 retro — PR #2149 (#2139)

**Task**: make a blocked push say whose failures blocked it, and find a red `main` before someone pushes into it.

### The finding worth keeping

Four review rounds, twelve defects — and **five of the twelve were caused by the previous round's fix**.

| round | defects | caused by the prior fix |
|---|---|---|
| 1 | 3 | — |
| 2 | 3 | 1 |
| 3 | 3 | 2 |
| 4 | 3 | 2 |

Round 2 fixed ids containing ` - ` with `sed 's/ - [^-]*$//'`, which only substitutes when the *message* has no hyphen — so it broke every id whose message contained one, which is most of them. Round 3 fixed the script's hardcoded `.venv` path and left the identical hardcode in the test file that verifies the fix, silently skipping 19 behavioural cases in a worktree. Round 3 fixed the report arms' gating and the surviving arm then made its global claim unconditionally.

This is what #2142's chain-shaped-issue convention was written for, and this PR is now its worked example in `CLAUDE.md`.

### What went well
- **Changing the shape instead of adding a case.** Classification grepped output text and had been patched twice; the third fix moved to pytest's **exit code**, which answers exactly (`1` = failed *or errored*, whatever the message says). Adding a fourth string is what produced the bug.
- **Executing the workflow's JavaScript** rather than grepping its YAML. Removing `if: always()`, collapsing the three states, and matching on label instead of title had all left the suite green; the harness caught five further defects, including a "could not run" issue that could never be closed.

### What didn't
- **Two harness traps produced tables of false zeros**, and I nearly acted on both. A baseline that silently collected nothing made 17 rows read 0. A mutation that prepended the old `sed` *ahead* of the new `awk` read 0 because the awk then corrected it. A gate-off table is worthless without an asserted clean-green baseline and a mutation verified to have applied *and* changed something.
- The CHANGELOG asserted "every path that cannot compare says the failures are not attributed" while an arm did the opposite — a prose invariant never run against the code (#1867).

### Verified
52 cases; 14/14 gate-off mutations fail at least one test; full suite 10,203 passed / 218 skipped; rebased to 0 behind main; CI green.

PR #2150, opened 2026-07-27
refactor(forms): remove the unreachable data-field_name attribute (#2145)

### LINKED ISSUE #2145: tech-debt: remove frameworks.py's unreachable data-field_name — the artifact that makes the #2137 false mechanism look true
Source: PR #2141 Stage 11/13 review (#2137)

## The attribute

`python/djust/frameworks.py:239,291,349` emits `data-field_name="<name>"` on every rendered form field, next to `dj-change="validate_field"` (`:237,289,345`).

**Nothing can ever read it.** `data-*` is collected only by `extractTypedParams` (`08-event-parsing.js:248`), which runs on the click/poll/mount paths (`09-event-binding.js:319,601,1206,1264,1336,1365`). The form-event path uses `buildFormEventParams` (`:525`), which merges only `dj-value-*`. And `frameworks.py` emits no `dj-click`/`dj-poll`/`dj-mounted` on those elements, so the one collector that *would* read it never runs on them — unreachable on every path it is emitted for.

It is harmless at runtime: `getFieldName` falls back to `element.name`, which the same code sets.

## Why it is worth removing anyway

This is the **artifact that makes the false mechanism look true**.

#2137 happened because the framework's own signature took `field_name` while the client sends `field`. PR #2141 corrected two docstrings that asserted *"djust maps `data-*` attributes to handler parameters"* — a claim inherited verbatim from `wizard.py`, true of the click/poll paths and false of the form-event path.

Anyone who opens `frameworks.py:239` sees an attribute that appears to be doing exactly that, and re-derives the belief that caused the bug. The docstrings patch the claim; removing this closes the loop.

## Scope

Not done in PR #2141 per #1079 — unrelated to the wire fix, and that PR was already four commits. Removal should be a small standalone change: delete the three `attrs["data-field_name"]` assignments, confirm the rendered-attribute tests still pass, and note in the commit why it was safe (the reachability analysis above, so the next person does not have to re-derive it).

## Related
- #2137 / PR #2141 — the bug and the docstring corrections
- #2144 — the root cause (`**kwargs` silencing param-name mismatches)

johnrtipton:
Deletes the three `attrs["data-field_name"]` assignments in
`python/djust/frameworks.py` (`_render_input`, `_render_checkbox`,
`_render_radio`).

## The reachability chain, verified rather than taken from the issue

`data-*` is collected by **exactly two** client functions. Every
`element.attributes` iterator and every `.dataset` read in
`python/djust/static/djust/src/` was checked; there is no third.

| collector | runs on | reads `data-field_name`? |
|---|---|---|
| `extractTypedParams` (`08-event-parsing.js:248`) | `dj-click` (`09-event-binding.js:601`), `dj-poll` (`:1206`), `dj-click-away` (`:1264`), `dj-shortcut` (`:1336`), `dj-mounted` (`:1365`), `dj-window-*`/`dj-document-*` (`:319`) | yes — but the adapters emit **none** of these directives |
| `_processAutoRecover` (`09-event-binding.js:1684`) | the `[dj-auto-recover]` **container's** own attributes | the adapters never emit `dj-auto-recover` |

The path the adapters *do* bind — `dj-change` / `dj-input` / `dj-blur` /
`dj-focus` — goes through `buildFormEventParams` (`:525`), which hardcodes
the key `field`, sources it from `getFieldName` (`:504`), and merges only
`dj-value-*`. It never touches `data-*`.

So the one collector that *would* have read the attribute never ran on the
element carrying it.

**What carries the name instead**: `name="<field_name>"`, set
unconditionally by all three renderers, read by `getFieldName` as its
second fallback (`data-field` → `name` → `id`). The reconnect path agrees:
`_processFormRecovery` (`:1773`) reads `field.name || field.id`.

Rendering all five widget shapes through all three adapters confirms it —
every element carries `name`, none carries `data-field`, and none carries
any data-collecting directive.

**Nothing else in the tree read it either**: zero tests, zero examples
(the demo templates hand-write `data-field=`, the attribute `getFieldName`
actually reads), zero Rust, zero server-side parsing.

## What the issue got wrong or omitted

1. **`extractTypedParams` has six call sites, not three.** The issue listed
   click / poll / mount; `dj-click-away`, `dj-shortcut` and
   `dj-window-*`/`dj-document-*` also collect `data-*`. The conclusion is
   unchanged — the adapters emit none of the six — but the enumeration was
   incomplete.
2. **There is a second `data-*` collector the issue did not mention** —
   `_processAutoRecover` (`09-event-binding.js:1684`).
3. **"Nothing can ever read it" is too strong.** `dom_event=` and
   `widget.attrs` are caller-controlled: `as_live_field(..., dom_event="dj-click")`,
   or a `dj-click` in `widget.attrs`, puts a data-collecting directive on
   the element and the attribute *does* reach the handler. Empirically
   confirmed — case 3 of the new JS test asserts exactly this. No in-repo
   caller, test, example, doc or demo does it, and such a configuration was
   never working anyway (the click path carries neither `value` nor `field`),
   but the honest statement is *unreachable on every path the framework
   itself emits*, not *unreadable in principle*.
4. **The radio site's assignment was outside the `auto_validate` guard.**
   `frameworks.py:349` sat one indent level out from the `if auto_validate:`
   block the other two sites kept it inside — so a radio rendered with
   `auto_validate=False` carried `data-field_name` on an element with **no
   djust directive at all**. The sharpest available evidence of inertness,
   and a test now pins it.
5. **Three `wizard.py` docstrings claim `as_live_field()` emits
   `data-field="<field_name>"`.** It does not, and never did. This is
   load-bearing rather than incidental: the `name` fallback is the entire
   reason the deletion is safe, and a docstring asserting `data-field` is
   emitted asserts that fallback is never reached. Corrected here
   (`wizard.py:192`, `:427`, `:443`).

## Tests

- `TestNoDataFieldNameAttribute` in `tests/unit/test_frameworks.py` — **42
  cases**. All three renderer sites covered individually (#1104) across all
  three adapters, plus the `auto_validate=False` radio case, plus two
  structural pins: every widget shape carries `name`, and no renderer emits
  a data-collecting directive (which is what would invalidate the analysis).
- `tests/js/no_data_field_name_2145.test.js` — **3 cases** making the chain
  executable against the real client: post-fix markup yields
  `field: "email"` from `name`; re-adding `data-field_name` changes nothing
  under `dj-change`; the same attribute under `dj-click` **does** reach the
  handler (and carries neither `value` nor `field`).
- `docs/guides/live-input.md`'s comparison table row for the attribute is
  replaced with the mechanism that actually applies.

## Gate-off (#1468)

Each mutation asserted to have applied before running; errors counted apart
from failures (#2129/#2135); `__pycache__` cleared each iteration.

| mutation | result |
|---|---|
| baseline | 0 failed / 108 passed |
| re-add site 1 (`_render_input`) | **12 failed** / 96 passed |
| re-add site 2 (`_render_checkbox`) | **3 failed** / 105 passed |
| re-add site 3 (`_render_radio`, original outside-guard shape) | **6 failed** / 102 passed |
| drop `name` from `_render_input` (pin) | **12 failed** / 96 passed |
| emit `dj-click` from `_render_input` (pin) | **3 failed** / 105 passed |
| JS: merge `data-*` into `buildFormEventParams` | **1 failed** / 2 passed |
| JS: drop `extractTypedParams` from the click path | **1 failed** / 2 passed |

Every site and every pin is independently reachable — no mechanism is
shadowed by another.

## Verification

- `pytest tests/ python/tests/ python/djust/tests/ -q -n auto` →
  **10080 passed, 219 skipped, 218 warnings in 41.36s**
- `vitest run tests/js/` → **170 files, 1919 tests passed**
- `ruff check` / `ruff format --check` clean; `mypy` clean;
  `scripts/check-doc-snippets.py` and `check-changelog-test-counts.py` clean
- branch is 0 commits behind `origin/main`

Closes #2145


johnrtipton:
## Stage 14 retro — PR #2150 (#2145)

**Task**: remove the unreachable `data-field_name` — the artifact that made #2137's false mechanism look true.

### What went well
- **Verifying the issue's reachability claim rather than trusting it changed the deliverable.** The issue said "nothing can ever read it". Tracing it found the claim too strong: `dom_event=` and `widget.attrs` are caller-controlled, so a caller who attaches a click-family directive to a form field *does* get that element's `data-*` collected. Deleted as asked, but every claim is now scoped to *"unreachable on every path the framework itself emits"* rather than *"unreadable in principle"* — and a test goes red if a renderer ever emits a data-collecting directive itself.
- The issue also under-counted `extractTypedParams`' call sites (six, not three) and missed a second `data-*` collector entirely (`_processAutoRecover`). The conclusion held; the stated reasoning did not.
- The sharpest evidence of inertness was found, not assumed: the radio site's assignment sat one indent level **outside** the `if auto_validate:` guard, so `auto_validate=False` rendered the attribute on an element carrying no djust directive at all.

### What didn't
- Three `wizard.py` docstrings claimed `as_live_field()` emits `data-field="<name>"`. It does not and never did — and that claim is load-bearing here, because the `name` fallback is the entire reason the deletion is safe. A docstring asserting `data-field` is emitted asserts the fallback is never reached. Corrected in the same PR. Same class as #1867: the citation was checkable and wrong.

### Verified
10080 passed / 219 skipped; vitest 1919 passed; 15/15 CI. Gate-off per site — re-adding site 1 fails 12, site 2 fails 3, site 3 fails 6; the two structural pins fail 12 and 3; both JS cases fail 1. Every mutation asserted to have applied, errors counted apart from failures.

Issue #2148, opened 2026-07-27
tech-debt: ~19 remaining stale client-size claims, mostly in demo copy


johnrtipton:
Source: PR #2147 Stage 13 review (#2138)

PR #2147 corrected 10 claims across 9 files and put every one of those files in the checked set. It did **not** finish the sweep, and the review was right that the CHANGELOG shouldn't imply otherwise.

## What remains

| where | count | note |
|---|---|---|
| `examples/demo_project/` | ~15 | rendered demo copy a user actually reads — `docs.html`, `index.html`, `REACT_TO_DJUST_TRANSLATION.md`, `FORM_PATTERNS_COMPARISON.md`, `status_change_djust.html`, `djust_homepage/index.html` |
| `docs/state-management/IMPLEMENTATION_PHASE2.md:27` | 3 | *"Client bundle size < 10KB (currently ~5KB)"* — a target stated against a figure that is 11× off |
| `ROADMAP.md:4610` | 1 | stale `~87 KB`, a live forward-looking entry |

All say `~5KB` for a client that is **58.7 KB gz / 233 KB raw** — wrong by roughly 11×.

`RETRO.md:50` shows this figure has been known-stale since PR #796/#800, so it recurs. That is the argument for finishing it rather than leaving it: the number is old enough to have been copied forward several times.

## Doing it

The mechanism already exists — `scripts/check-doc-snippets.py` has `_SIZE_CLAIM_FILES`. Correct each claim and add its file to that list in the same change, so the correction is guarded rather than just applied. `make sizes` prints the current figures.

Judgement call worth making explicitly: the `examples/` copies are *demo* prose, and it may be right to fix them without adding `examples/**` to the checked set (a large, churn-heavy tree). Either way, decide it rather than letting it default.

## Two smaller follow-ups from the same review

- **Marker matching is exact-string.** `_SIZE_ARTIFACT_MARKERS` keys must match byte-for-byte, so 5 of 6 plausible spellings (`<!--size-claim: unminified-->`, `<!-- SIZE-CLAIM: UNMINIFIED -->`, …) fall back to the default. Meanwhile `_SIZE_HISTORICAL_MARKER` is a bare substring and *is* tolerant — two marker families, two matching rules. One case-insensitive regex `<!--\s*size-claim:\s*(unminified|shipped|historical)\s*-->` fixes both. Mitigating: every mis-spelling fails **loudly**, never silently passes.
[...]

PR #2151, opened 2026-07-27
fix(validation): warn when **kwargs swallows a near-miss parameter (#2144)

### LINKED ISSUE #2144: tech-debt: **kwargs turns a handler param-name mismatch into silence — warn on swallowed keys
Source: PR #2141 Stage 11 review (the root cause behind #2137)

## The class

#2137 was one instance: the client sent `field`, `FormMixin.validate_field` took `field_name`, and because the signature also had `**kwargs` the mismatch was **absorbed instead of raising `TypeError`**. The handler ran on every keystroke and did nothing. No error, no warning.

Fixing two signatures fixed that instance. It did not address the mechanism, and every djust handler with `**kwargs` — which is the documented, recommended shape — has the same exposure.

## The hook already exists

`python/djust/validation.py:288` in `validate_handler_params` **already computes** the set of unexpected keys, and then suppresses the diagnostic when `has_var_keyword` is true. So the information needed is on hand at exactly the right point; it is deliberately discarded.

A `logger.warning` there — naming the handler, the unexpected key, and the parameters it *does* accept — would have turned #2137 from a silent no-op into a one-line log on the first keystroke.

## Why it wasn't just done

Framework-internal keys are legitimately passed through `**kwargs` and would produce noise. `_args` is already popped upstream (`runtime.py:1152,2777,3167`; `websocket.py:1567`), so the denylist looks small — but it needs enumerating rather than guessing, and a noisy warning on a hot path (every keystroke) is worse than none.

## Suggested shape

1. Enumerate the framework-injected keys that legitimately reach `**kwargs`.
2. Warn once per (handler, unexpected-key) pair rather than per event — this fires on input paths, so per-event logging is not acceptable.
3. Consider gating on `DEBUG` or `djustDebug`-equivalent if the noise is still material.

## Related
- #2137 / PR #2141 — the instance
- `python/djust/validation.py:288` — where the information already is

johnrtipton:
Closes #2144.

## The class, not the instance

#2137 was one instance: the client sent `field`, the framework's own `FormMixin.validate_field` took `field_name`, and because the signature also had `**kwargs` the mismatch was **absorbed instead of raising `TypeError`**. The handler ran on every keystroke and did nothing — no error, no warning.

Fixing two signatures fixed that instance. Every djust handler with `**kwargs` — the documented, recommended shape — kept the same exposure. `validate_handler_params` already computed the unexpected-key set and then discarded it whenever `has_var_keyword` was true, so the information was on hand at the right point.

## Why only near misses

Warning on every unexpected key would be wrong. An unexpected key bearing no resemblance to the signature is the documented catch-all shape (`def on_event(self, **kwargs)` reading `kwargs` directly). Warning on those puts noise on `@input`/`@change` — the exact paths where this must stay quiet — and a warning nobody can act on is one people learn to filter.

**The discriminator was measured, not chosen by taste.** I instrumented the suppression point and ran the full suite:

| | count |
|---|---|
| keys that legitimately reach a `**kwargs` handler | 110 |
| of those, reported by the near-miss rule | **0** |

And the canary (#1459) — it catches the shapes it claims to:

| sent | declared | reported |
|---|---|---|
| `field` | `field_name` | ✅ (#2137 verbatim) |
| `item_id` | `itemId` | ✅ snake/camel drift |
| `page_num` | `page_number` | ✅ abbreviation |
| `checked` | `is_checked` | ✅ prefix drop |
| `csrf_token`, `metadata`, `extra`, `id`, `payload` | — | silent |

## The denylist was enumerated, not guessed

The issue asked for enumeration. Reading call sites would have missed one, so I instrumented instead. Across 10,058 tests exactly two framework-injected keys reach `**kwargs` without an underscore: `_cacheRequestId` (covered by the private-name convention) and **`view_id`**, which survives the actor path because `runtime.py:2734` reads it with `.get` where `:3457` pops it. Underscore-prefixed keys are excluded wholesale, so a future `_foo` cannot start generating noise.

## Not becoming the noise it warns about

- **Warn once per (handler, key)** — 50 keystrokes produce one line, not fifty. Dedup is per-key rather than per-handler so a handler with two mismatches reports both.
- **The cache is capped.** Parameter names come from the client, so an unbounded warn-once set is a memory-exhaustion vector for anyone who can open a socket. On reaching the cap the diagnostic switches off rather than degrading to per-event logging.
- **`%r` for client-influenced fields**, so `"field_nam\r\n"` cannot forge a second log record.

## Gate-off (#1468)

| mutation | tests failing |
|---|---|
| detector never called | 10 |
| near-miss filter removed (warn on all) | 6 |
| underscore/framework-key skip removed | 5 |
| warn-once dedup removed | 1 |
| cache bound removed | 1 |
| `%r` → `%s` | 1 |
| event-name sanitisation dropped | 1 |
| `if not accepted` early return removed | **0** |

The eighth fails nothing because `difflib.get_close_matches(k, [])` already returns `[]` — it is a fast path, not a behaviour. Rather than delete it or leave a decorative test, the code now says so, so a reader does not mistake the zero for a missing test.

Two of these only gate after a fix: the cache bound originally had **two** guards masking each other (removing either alone left the suite green — the same class as #2148's build guards), and the CRLF test was passing because of `%r`, not the `sanitize_for_log` call it claimed to exercise. Both are now singular and independently gated.

## Verification

- 26 cases in `python/djust/tests/test_swallowed_kwargs_warning_2144.py`
- Full suite across all three roots: **10,064 passed**, 218 skipped
- The live rule fires **0** times across the whole suite
- `ruff check` clean

Unrelated: `test_time_travel_jump_recovery_version_is_current` is an order-dependent flake (passes 3/3 in isolation, references nothing in `validation.py`, and failed on an unrelated branch before this change). Filed separately.


johnrtipton:
## Fix pass — the review's 🔴 was right, and the rule changed because of it

Re-measured with this PR's own tests excluded:

| | |
|---|---|
| observations | 104 |
| synthetic `param_N` from **one** loop (`test_validation.py:404`) | 100 |
| real keys reaching a comparison | **4** (`any`, `extra`, `extra_key`, `works`) |

Four comparisons, each against a single-element parameter list. "Measured, not chosen by taste" was overstated and is withdrawn from the commit, PR body and CHANGELOG.

### The rule was wrong in both directions, so it is gone

`difflib` at 0.6 claimed these are typos of each other — all djust's own parameter-name shapes:

`data`/`date` 0.750 · `from`/`form` 0.750 · `mode`/`node` 0.750 · `tab`/`tag` 0.667 · `id`/`uid` 0.800 · `q`/`qs` 0.667

…while staying **silent** on `name` ← `field_name` (0.571) and `page` ← `page_number` (0.533), which are #2137 one step further along.

Replaced with a **structural** rule: same name modulo case and underscores (`item_id`/`itemId`), or one name a prefix/suffix of the other (`field`/`field_name`, `checked`/`is_checked`), with a 3-character floor. Every false positive above now goes silent **and** both missed shapes now fire — better on both axes rather than a trade, and explainable in one sentence to whoever reads the warning. `difflib` is gone, which also closes 🟡 5's O(n·m)-on-the-keystroke-path concern.

### The rest

| finding | fix |
|---|---|
| 🟡 2 `__qualname__` collision | cache keyed on `(module, qualname, key)`; test asserts two apps sharing a class name both warn |
| 🟡 3 hard cap = cheap permanent silencing | LRU eviction; test asserts a genuine bug is still reported *after* a 712-key flood |
| 🟡 4 `component_id` unmeasured | kept — it is reachable via the actor branch (`runtime.py:2657-2661` → `:3169`) — but the comment now says which entries were **measured** and which are **by inspection** |
| 🟡 7 positional-only | excluded from the suggestion pool |
| 🟡 6 / M16 handler name | the event name was satisfying the assertion; new test uses a handler whose name differs |
| 🟡 6 / M13 accepted list | asserted `field_name`, which is also the suggestion; now asserts a second declared parameter |
| 🟡 6 / M9 `sanitize_for_log(key)` | dropped rather than left unpinned — `%r` does the escaping and a near miss is structurally length-bounded, so the cap was unreachable there. Kept on `event_name`, where it is reachable, and pinned. |

Not taken: 🟡 5's negative-verdict cache (moot — no `difflib`) and the dedup hoist, which **is** applied: the `pair in cache` check now runs before the match, so the steady state on a keystroke path is one dict lookup.

### Verification

42 cases. **17 of 18** gate-off mutations fail at least one test — swapping the structural rule back for the 0.6 ratio fails 10, dropping the length floor 3, case/underscore normalisation 3, the suffix arm 3, the module from the cache key 1, LRU→hard-stop 1, positional-only filter 1. The eighteenth is the `if not accepted` fast path, documented in-code as creating no behaviour.

Full suite 10,081 passed / 218 skipped; ruff and mypy clean. Rebased on main (the two merged PRs don't touch `validation.py`).

johnrtipton:
## Stage 14 retro — PR #2151 (#2144)

**Task**: warn when a `**kwargs` handler swallows a wrong parameter name — the mechanism behind #2137.

### What went well
- Fixing the **class at the suppression point** rather than two more signatures. `validate_handler_params` already computed the unexpected-key set and threw it away.
- The measurement that mattered was the one that **contradicted me**. I claimed the discriminator was "measured, not chosen by taste" on 110 keys with 0 fires; re-measuring with this PR's own tests excluded gave 104 observations, **100 of them one synthetic loop** — 4 real comparisons. The claim was withdrawn from the commit, PR body and CHANGELOG.

### What didn't
- **A similarity score was the wrong tool and I reached for it first.** `difflib` at 0.6 claimed `data`/`date`, `from`/`form`, `mode`/`node` and `id`/`uid` are typos of each other — confidently wrong advice about djust's own parameter names, worse than the silence it replaced — while staying *silent* on `name`/`field_name` (0.571) and `page`/`page_number` (0.533), which are #2137 one step further along. The structural rule (same name modulo case/underscores, or one a prefix/suffix of the other, 3-char floor) is better on **both** axes, not a trade, and is explainable in one sentence to whoever reads the warning.
- **The cap I presented as the security-conscious choice was the cheaper attack.** A hard stop meant ~560 events of novel keys from one socket permanently blinded the diagnostic for every view and tenant. LRU eviction keeps memory bounded without that.
- Three assertions couldn't distinguish what they claimed: the handler name was satisfied by the *event* name; "lists what it accepts" asserted the string that was also the suggestion; `sanitize_for_log(key)` was pinned by `%r`, not by itself.

### Verified
42 cases; 17/18 gate-off mutations fail at least one test (the 18th is documented in-code as a fast path that creates no behaviour); full suite 10,151 passed; ruff and mypy clean.

PR #2152, opened 2026-07-27
fix(docs): finish the client-size sweep and make the guards load-bearing (#2148)

### LINKED ISSUE #2063: feat(ecosystem): official adapters — dj-chart pilot (Chart.js), user-brings-the-library

johnrtipton:
Closes #2063.

The deferred "milestone C" from ADR-025. The two extension sockets shipped in
v1.1.0rc8 (`JS.ext.*` #2051, typed hook values/targets #2052); this is the first
adapter that rides on them.

**User brings the library; djust ships only the glue.** Chart.js is never
bundled, downloaded, or vendored.

## What ships

| File | Role |
|---|---|
| `python/djust/static/djust/ext/dj-chart.js` | The adapter: a pre-written `dj-hook` + pre-registered `JS.ext` commands. 2.3 KB gz. |
| `python/djust/extensions.py` | Single source of truth for which adapters exist. |
| `python/djust/mixins/post_processing.py` | Emits the adapter tag after `client.js`. |
| `python/djust/checks/configuration.py` | New check `djust.C015` for unknown adapter names. |

```python
# settings.py — the entire setup
DJUST_CONFIG = {"extensions": ["chart"]}
```

```html
<canvas dj-hook="Chart"
        dj-hook-value-type='"bar"'
        dj-hook-value-data='{{ chart_data_json }}'></canvas>
```

## Design notes for the reviewer

- **Not bundled into `client.js`.** The shared ~87 KB gz budget is unchanged for
  anyone who doesn't opt in; the adapter is fetched only when enabled. Pinned by
  a test asserting `client.js` contains no adapter code.
- **One table, two consumers.** The injector and C015 both resolve through
  `extensions.py`, so they cannot drift (#1646 — this repo's most-recurring
  failure class).
- **Load order is the contract.** The adapter registers into `window.djust.*`,
  which `client.js` creates. `defer` preserves document order, so emitting the
  tag *after* the client tag is what makes `commands.register` reachable —
  asserted directly rather than assumed.
- **Zero-cost when unused** (#246): nothing emitted when the list is empty.
- **`updated()` mutates in place.** Destroy-and-rebuild here is exactly the
  #1724 teardown class this adapter exists to prevent.
- **Exactly ONE adapter**, per the issue. Pinned by
  `test_chart_is_the_only_shipped_adapter`; `dj-sortable`/`dj-editor` wait for
  demand, since each is an ongoing maintenance commitment.

## Two things self-review caught (worth a look)

1. **The first draft's recipe would have broken the feature.** It told users to
   add `dj-update="ignore"` to the canvas. The patcher returns early on that
   attribute *before* attribute sync (`12-vdom-patch.js`), so the server could
   never update `dj-hook-value-data` and the chart would show its first dataset
   forever. It's also unnecessary — a `<canvas>` has no server-owned children,
   and the morph already refuses to remove a canvas's `width`/`height`. Fixed in
   `aa8ba0b3` and pinned behaviorally.
2. **Missing library warned once per re-render**, because `updated()` retries the
   mount. Now once per element (`WeakSet`).

## Tests

- **15** cases in `tests/unit/test_extensions_2063.py` — resolution (incl.
  unknown-name-dropped, non-list config), injection presence/order/`defer`, C015
  incl. suppression.
- **10** cases in `tests/js/dj_chart_adapter.test.js` — driven through the REAL
  dispatcher (`mountHooks`/`updateHooks`/`destroyAllHooks`) and real
  `js._executeOps`, not hand-built instances (#1196). The first draft *did*
  hand-roll an instance and silently double-mounted — the wiring is precisely
  what a method-level test cannot see.
- **Gate-off verified** (#1468): neutering the injection loop fails 3 of the
  injection tests.
- **C015 dogfooded** against the demo project: 0 false positives, plus an
  empirical canary confirming it fires on a typo (#1459/#1060).
- Full suites green locally: 9906 Python / 1847 JS, 0 failed.
- Wheel packaging verified empirically by inspecting a real `.whl` (nested
  static subdirs do ship).


johnrtipton:
## Stage 14 retro — PR #2152 (#2148)

**Task**: finish the client-size sweep #2147 started, and make "guarded" a checked property.

### What went well
- **The gate-off caught a tautology in this PR's own work.** The first version bounded the marker regex to the three known values and tested that an invented value "does not suppress" — it reported 0 failed. Both lookups are membership tests, so an unrecognised value resolves identically whether the alternation is bounded or not; the test could not go red. The bound is load-bearing now: an unknown value is *reported at its file:line*, so `<!-- size-claim: histrical -->` is named rather than silently governing nothing.
- **Wording had to change for guarding to mean anything.** Several claims sat alone on a line with their label on the neighbouring one (`<div class="hero-stat-value">~5KB</div>`), which a per-line checker structurally cannot see. Adding those files to the checked set without also rewording would have been a decorative guard (#1859). Each of the 14 files was verified to actually fire.
- The part that retires the class rather than the instance: every checker-visible figure must now be either checked or allowlisted **with a reason**, and a companion test fails when an allowlist entry has nothing left to allow. That is what would have prevented #2148 existing — "correct the number" was diligence with nothing checking the file would ever be looked at again.

### What didn't
- The issue's inventory was wrong in three ways (12 files not ~6; `IMPLEMENTATION_PHASE2.md` has 1 checker-visible claim not 3; it missed `STATE_MANAGEMENT_API.md:1869`'s `client.js (current): ~5.0 KB`, the most misleading claim in the repo because it says *current*). Sweeping independently rather than working the list was necessary.
- Six `docs/state-management/` design docs are left unfixed (#1079) — one coherent class each needing a historical-vs-correct judgement. Allowlisted so they are visible rather than forgotten.

### Judgement recorded
`IMPLEMENTATION_PHASE2.md` is marked historical, not renumbered: it is a dated plan whose `< 10KB` target was superseded at v0.6.0, and restating it would claim Phase 2 set a target it never set — the #2028 class of retro-editing a dated record.

### Verified
10047 passed / 219 skipped; 13/13 CI. All 7 behaviours go red when neutered; empirical canary confirms all 14 newly guarded files fire on a re-introduced stale figure.

PR #2153, opened 2026-07-27
fix(retro): make the Action Tracker's open count mean something (#2143, #2142)

### LINKED ISSUE #2142: tech-debt: chain-shaped drains need an estimation convention

### LINKED ISSUE #2143: tech-debt: close Action Tracker rows automatically when their GitHub issue closes
Source: v1.1.0-12 retro finding 2 (backfilled under #2140)

## The drift

Action Tracker row 325 read `Open` with GitHub #2082. That issue was closed **2026-07-18** by PR #2093 — a full milestone before anyone noticed. The v1.1.0-13 retro's Stage 5 sweep caught it by chance, because that sweep only looks at rows it has reason to suspect.

Nothing closes a tracker row when its issue closes, so `RETRO.md`'s open-count drifts upward over time and every retro's Stage 5 has to re-derive the true state by hand.

## Current cost

69 rows currently read `Open`. Some unknown fraction are already resolved. That number is quoted in retros as a health signal and is not trustworthy.

## What a fix looks like

A small script (runnable from `/pipeline-retro` Stage 5, or as a CI job) that:

1. parses the Action Tracker table for rows with a GitHub number and status `Open`;
2. queries each issue's state via `gh issue view <n> --json state`;
3. reports (or auto-edits) rows whose issue is `CLOSED`, requiring a human to supply the closing reason so the Notes column stays useful.

Report-only is probably the right default — the reason matters more than the status flip, and a bot cannot write it.

## Related
- v1.1.0-12 retro, finding 2
- Row 325 / GitHub #2082 — the instance

johnrtipton:
Closes #2143. Closes #2142.

| issue | files | tests |
|---|---|---|
| #2143 | `scripts/check-action-tracker.py`, `Makefile`, `RETRO.md` | `tests/test_action_tracker_drift_2143.py` (19) |
| #2142 | `CLAUDE.md` | — (a convention, not code) |

## #2143 — 65 of 70 "Open" rows were already resolved

Nothing closed a tracker row when its issue closed, so `RETRO.md`'s open count drifted upward and every retro's Stage 5 re-derived the true state by hand. Row 325 sat at `Open` against an issue closed a full milestone earlier, found by chance.

The count matters because retros quote it as a health signal. Against the real file:

```
Action Tracker: 329 rows, 70 marked Open.
65 row(s) read Open but their issue is CLOSED.
```

**The true figure was 5.** Wrong by more than an order of magnitude — worse than no signal, because it made a drained backlog look untouched.

### Design

- One `gh` call for every issue's state, not 70 round trips.
- Reports **both** directions. A `Closed` row whose issue was reopened is the more dangerous drift: it hides live work.
- **Report-only by default.** A bare status flip discards the half of the row that makes it worth keeping, so `--fix` looks up the **closing PR** and writes it as the reason — mechanical *and* informative. It never overwrites a reason a human wrote.
- **Not a blocking CI gate.** It needs the network, and drift between a merge and the next retro is normal rather than a fault. `make check-tracker`, and `/pipeline-retro` Stage 5.

Three refusals to be quietly useless — each is the "nothing was checked" reading as "all clean" failure:

| situation | exit |
|---|---|
| GitHub unreachable | 2 |
| table unparseable (format drift) | 2 |
| row cites an issue GitHub doesn't return | not drift, either direction |

### The 65 rows are fixed here

Verified the edit touched only Status and Notes: **6076 lines before and after, 65 changed, 0** whose number/action/source/issue cells differ.

### Gate-off (#1468)

| mutation | tests failing |
|---|---|
| drift detection removed | 6 |
| closing PR not written | 2 |
| reverse drift not reported | 1 |
| ALL-issues rule → ANY | 1 |
| OUT-OF-REPO rows swept in | 1 |
| human reason overwritten | 1 |
| unparseable row edited anyway | 1 |
| network failure exits 0 | 1 |
| empty table exits 0 | 1 |
| the no-comparison guard removed | 2 |

The eleventh mutation initially read **0** and that found a real defect in my own work: `if not r.issues` and `if not known` masked each other, so deleting either alone left the suite green — the same two-guards-mask-each-other shape as #2148's build guards. I removed the redundant guard rather than writing a test for something that could not fail; the remaining check now fails 2 when neutered.

## #2142 — chain-shaped issues need a convention, not an estimate

Some issues are one bug; others are a class, and you only learn which from the review of the fix. Three ran back to back — scoped-attr (#2107 → #2109 → #2111), streams (four links), #2129 (five rounds) — and #2139 took three, **two of whose defects were introduced by the previous round's fixes**.

Each link shipped separately because each was separately reachable; batching would have merged four incomplete diagnoses. The only real problem is that the drain plan presents such an issue as one item, so a chain reads as a slipped estimate rather than as the shape of the work.

`CLAUDE.md` now says: mark an issue **chain-suspected at Stage 4** when the root cause is a class (parallel-path drift, wire-contract mismatch, shared-key collapse, or "the same edit at N sites"); **file link N+1 immediately** with a back-reference so the chain lives in the issue graph rather than the PR trail; **count a chain as one finding and N PRs** in the retro.

And: do not estimate one up front. By construction you cannot — link N+1 is discovered by reviewing link N — and an estimate that pretends otherwise makes honest work look late, which is exactly the pressure that produces the batched half-diagnosis the convention exists to avoid.

## Verification

- 19 cases, offline (they pin the logic, not today's GitHub state)
- `make check-tracker` → `329 rows, 5 marked Open. No drift.`
- ruff clean


johnrtipton:
## Stage 14 retro — PR #2153 (#2143, #2142)

**Task**: make the Action Tracker's open count mean something, and give chain-shaped issues a convention.

### What went well
- **Measuring first turned a maintenance chore into a real finding.** 65 of 70 rows marked `Open` had a closed issue; the true count was **5**. A health signal retros quote was wrong by an order of magnitude, and made a drained backlog look untouched.
- **Sourcing the reason mechanically.** The issue argued for report-only because "a bot cannot write the reason". Mostly true — but GitHub knows the *closing PR*, which is the reason, so `--fix` writes it and never overwrites a human's.
- Three refusals to be quietly useless: an unreachable GitHub, an unparseable table, and an issue GitHub doesn't return all fail loudly rather than reporting zero drift. "Nothing was checked" reading as "all clean" is the exact failure the script exists to end.

### What didn't
- **The gate-off found two redundant guards masking each other** — `if not r.issues` and `if not known` — so deleting either alone left the suite green. Same shape as #2148's build guards and #2144's cache bounds; three instances in one drain. Removed the redundant one rather than writing a test for something that could not fail.

### On #2142
Written as canon precisely because #2139 was running four review rounds in parallel with it, five of whose twelve defects were caused by the previous round's fix. The convention says: mark chain-suspected at Stage 4, file link N+1 immediately, count a chain as one finding and N PRs — and don't estimate one up front, because pretending you can makes honest work look late.

### Verified
19 offline cases; 11/11 gate-off mutations fail at least one test after the redundant guard was removed; `make check-tracker` → `329 rows, 5 marked Open. No drift.`

PR #2158, opened 2026-07-27
fix(ci): use the proven install recipe in main-health (#2155)

### LINKED ISSUE #2155: main-health check could not run
**The main-health check could not run** — the job failed before or during setup
(dependency install, toolchain, or runner), or pytest produced no parseable summary.
This is NOT evidence that main is red; it is evidence that we currently cannot tell.

Run: https://github.com/djust-org/djust/actions/runs/30259056481

johnrtipton:
Closes #2155.

The main-health workflow (#2139) ran on schedule for the first time, failed, and filed *"main-health check could not run"*.

**The three-state logic worked exactly as designed** — it did not claim `main` was red, which would have sent people hunting for a regression that does not exist. But the job itself was broken.

## Cause: a second, hand-rolled install recipe

```yaml
uv venv
uv pip install -e ".[dev]" || uv pip install -e .
.venv/bin/pip install maturin      # <-- no such file
```

`uv venv` does not create a `pip`. The run log shows the `uv` install succeeding (`+ maturin==1.14.1` — it is already a dev dependency), then the step dying on the next line. So it was redundant *and* broken.

Writing a second install path for one project is the parallel-path drift class (#1646). Replaced with the pair `.github/workflows/test.yml` proves on every run:

```yaml
uv sync --extra dev
uv run maturin develop --release
```

The pin asserts both halves agree, and fails if `test.yml` ever moves off that recipe — so the two cannot silently diverge again.

## Verification

- 17 cases in `test_main_health_workflow_2139.py` (the JS-execution harness plus this new structural pin)
- Gate-off: restoring the `.venv/bin/pip` shape fails the pin (1 failed)
- Full suite: **10,204 passed**, 218 skipped
- Not verifiable locally, by definition — per canon #1534 a CI job exercising an environment the dev machine cannot reproduce gets at least one runner-only iteration. This is that iteration.

## One process note

The pin's first version grepped the whole file and failed on the **comment** above the install, which names `.venv/bin/pip` to explain the bug. That is the **fourth** time this exact comment-vs-code hole has appeared in this arc. A source-grep assertion should strip comments by default rather than have it remembered case by case — worth folding into the canon next time the PR checklist is touched.


johnrtipton:
## Stage 14 retro — PR #2158 (#2155)

**Task**: fix the main-health workflow, which failed on its first scheduled run.

### What went well
- **The three-state logic did its job on day one.** The job broke and the workflow filed *"could not run"* rather than *"main is RED"* — so nobody was sent hunting for a regression that did not exist. That distinction was added in the #2139 review rounds precisely for this, and it earned its keep immediately.
- The cause was a second, hand-rolled install (`uv venv` creates no `pip`) — parallel-path drift (#1646) in a workflow I wrote *while* documenting that class. The fix pins both halves to `test.yml`'s recipe and fails if they diverge again.

### What didn't
- **CI caught a real defect my local runs could not**: the collect-only backstop treated *any* non-zero exit as "not collectible here", so when the second pytest failed for environment reasons on the runner, a genuinely-new test was demoted to "not attributed" — a worse answer than before the backstop existed. Now only pytest's own cannot-resolve codes (4/5) demote.
- That is the same shape as #2156 one level up: **an added check must fail safe.** A check that cannot run should be silent, never confidently wrong. Twice in one PR arc.
- Per canon #1534 this needed a runner-only iteration, which it got. Worth remembering that a workflow cannot be validated locally by construction.

### Verified
17 cases; gate-off confirmed; 10,206 passed; all checks green.

PR #2159, opened 2026-07-27
fix(ci): add the benchmark job conftest.py has always claimed exists (#2156)

### LINKED ISSUE #2156: benchmark latency thresholds are enforced only in the least representative environment, and blocked every push

johnrtipton:
Closes #2156.

## Stage 4 answered the issue's first question, and it changed the fix

The issue asked: *"Does that benchmark-gated CI job actually exist and run?"*

```
$ grep -rn -- --benchmark-only .github/workflows/
$
```

**It has never existed.** `tests/benchmarks/conftest.py` has told readers *"the benchmark-gated CI job (`--benchmark-only` serial) enforces it"* for the whole life of that sentence — the #1867 class, a prose invariant nobody ran against the code.

That matters because `_assert_benchmark_under` skips its threshold whenever `benchmark.stats` is unavailable, which is what pytest-benchmark does under xdist:

| context | threshold enforced? |
|---|---|
| CI (`-n auto`) | no |
| `make test` (`-n auto`) | no |
| serial pre-push | **yes** — after 10,000 tests, in the same process |

So a threshold no CI job checked was blocking every push on `main` while measuring the environment. And #2156's own mitigation (`--benchmark-disable` in the pre-push run) left them enforced **nowhere** — exactly the risk the issue flagged.

## Measured before deciding, rather than adjusted

Serially on a quiet machine, **all 58 benchmarks pass** with thresholds enforced:

| benchmark | median | target | headroom |
|---|---|---|---|
| `vdom_diff_list_reorder` | **3.78 ms** | 5 ms | **32%** ← tightest |
| `vdom_diff_list_append` | 3.56 ms | 5 ms | 29% |
| `websocket_mount_counter` | 6.26 ms | 100 ms | 94% |
| everything else | microseconds | ms | — |

The numbers were never the problem; the environment was. **No target is loosened.** Reaching for a bigger bound is the reflex this issue exists to stop — it is what turned a 10 ms threshold into 100 ms in `test_redis_serialization_performance`, which then flaked again at 100 ms.

`list_reorder`'s tight margin is why it flipped under load *and* why it will catch a real regression first. Recorded so the next reader doesn't mistake tightness for fragility and widen it.

## What ships

A `benchmarks` job: serial, `--benchmark-only`, thresholds enforced. Explicitly **not** `-n auto` — inheriting it would silently disable the stats the job exists to assert on, which is the whole bug.

It ships `continue-on-error: true` and **outside** `test-summary`'s AND-condition, per canon #1534: a CI job exercising an environment the dev machine cannot reproduce stays non-gating until it has been green on the runner. This is a timing gate on a shared runner; it proves itself first.

**Promotion is pinned as all-or-nothing.** A job that sits in `needs` and is echoed in the summary but is missing from the AND-condition looks enforced and gates nothing (#1713). `test_promotion_is_all_or_nothing` fails if the job is ever `continue-on-error` *and* read as a merge condition, so the half-promoted state cannot ship.

## Gate-off (#1468) — baseline asserted clean-green first

| mutation | tests failing |
|---|---|
| benchmark job deleted | 5 |
| `continue-on-error` removed (half-promoted) | 2 |
| job silently switched to `-n auto` | 1 |
| `--benchmark-only` dropped | 1 |
| `TARGET_LIST_UPDATE_S` loosened | 1 |
| `TARGET_PER_EVENT_S` loosened | 1 |
| false docstring restored | 1 |

## One process note worth more than this PR

The `-n auto` absence assertion first tripped on **my own comment** saying `"No -n auto:"`. That is the **fifth** time in this arc (#2139 ×3, #2155, #2156) that a source-grep pin has been satisfied by the comment explaining the thing it pins — caught by gate-off every time, fixed locally every time.

So this adds a `code_only()` helper that strips comment lines, with the count in its docstring. Any absence assertion over source should go through it. Worth folding into the PR checklist next time it's touched.

## Verification

- 7 cases in `tests/test_benchmark_enforcement_2156.py`
- Full suite: **10,213 passed**, 218 skipped
- Pipeline: `.pipeline-state/fix-2156-benchmark-threshold-enforcement.json`, Gates 1 and 2 both green


johnrtipton:
## First runner iteration: **green**, and the margin is far wider than expected

`benchmarks (serial, thresholds enforced)` — **58 passed in 36.65s**, first try. That satisfies canon #1534's "green on the runner at least once" precondition for promotion.

The measured margins settle the question the review raised — whether a shared runner could ever meet these bounds:

| benchmark | dev machine (busy) | post-10k-test process | **GitHub runner** | target |
|---|---|---|---|---|
| `vdom_diff_list_reorder` | 3.78 ms | **7.57 ms** (the failure) | **0.656 ms** | 5 ms |
| `vdom_diff_list_append` | 3.56 ms | — | 0.644 ms | 5 ms |
| `websocket_mount_counter` | 6.26 ms | — | 11.38 ms | 100 ms |

The runner is **7.6× under** target on the tightest benchmark — more headroom than my dev machine had, not less.

So the environmental spread on the same code is roughly **11×** (0.656 ms → 7.57 ms). That is the entire finding of #2156 in one row: the threshold was never measuring the code.

**Implication for promotion.** The stated reason for shipping `continue-on-error` was that a timing gate on a shared runner is unproven. It is now proven once, with 7.6× headroom. I am still not promoting it in this PR — one green run is one data point, and #1534 says *at least* once — but the follow-up is now a small, well-evidenced change rather than a gamble. `test_promotion_is_all_or_nothing` ensures it cannot be done halfway.

Worth noting the dev-vs-runner gap deserves its own look: a `--release` build should not be 5.8× slower locally than on a runner. Either my local extension is not the release build I assumed, or the machine was loaded by concurrent agents. That does not affect this PR — but it does mean local benchmark numbers should not be trusted as a baseline for tuning targets.

johnrtipton:
## Stage 11 findings addressed — the 🔴 was real and my fix for it was wrong twice

**F1 🔴 — the headline claim was decorative.** `test_promotion_is_all_or_nothing` scanned the *whole file* for `needs.benchmarks.result`, and `test-summary`'s echo block mentions every job. The reviewer shipped the complete half-promoted state and got **0 failed / 7 passed**. That is #1713 verbatim — the trap the test's own docstring quotes — and it was the claim the PR body, CHANGELOG and docstring all led with.

Scoped to the `if [ ... ] && [ ... ]` chain, which is what actually fails the run:

| mutation | before | after |
|---|---|---|
| fully half-promoted ship | **0 failed** | **2 failed** |
| comment naming the string (F2) | 1 failed *(false positive)* | **0 failed** |
| genuine promotion (in the AND-chain) | — | 1 failed — only the non-blocking pin, written to be inverted deliberately |

**Two harness errors of mine on the way there**, both of which produced false confidence and are worth recording:
- My first re-check reported the killer *still* passing. The mutation used a first-occurrence `replace` and stripped a **different job's** `continue-on-error`, leaving the benchmarks job untouched. Now scoped to the benchmarks block with the anchor asserted inside it.
- A `head -3` on the anchor grep hid the real match 400 lines further down, which sent me hunting for an anchor that was present all along.

**F2 🟡** — the same scan false-*failed* on a comment. Sixth instance of that class in this arc, in the PR that added `code_only()` for it and routed one assertion through it. Now routed.

**F3** renamed — the name claimed a global property the body never checked, and which is false anyway (`python-tests` runs `pytest tests/ -n auto`, and `tests/` contains `benchmarks/`, harmlessly).

**F4** — `importorskip("yaml")` made all seven pins report `1 skipped` with the suite green. PyYAML is undeclared, arriving via `uvicorn[standard]`. Hard import now, matching the sibling pin.

**F5** — the docstring quoted a contended laptop's 3.78 ms / "32% headroom" in a sentence naming the CI job, where the real figure is **655 µs = 13% of budget**. (The 32% didn't match its own formula either.) It now cites the enforcing environment and records the ~11× spread with an explicit *do not tune these against a local run*.

**F6** — orphaned `# Summary job` comment moved back onto `test-summary`.

**F8** — install now uses `astral-sh/setup-uv@v7` + `Swatinem/rust-cache@v2`, matching the six sibling uv jobs and every Rust-building job **in this same file**. The hand-rolled `pip install uv` was the #1646 drift class that #2155's own entry names; the cache also removes most of a ~2m30s build.

**F7** — filed **#2160** with the runner evidence. Worth stating plainly: `continue-on-error: true` makes `needs.benchmarks.result == 'success'` *even on failure*, so today the job proves the thresholds **can** be enforced without enforcing them.

Full suite: 10,213 passed / 218 skipped.

johnrtipton:
## Stage 14 Retrospective — PR #2159 (#2156)

**Task**: benchmark latency thresholds enforced only in the least representative environment.
**Quality**: 4/5 — the substance was right and empirically validated, but the one novel safety claim shipped decorative and a reviewer had to prove it false.

### What worked
- **Stage 4 asked the issue's first question instead of assuming it.** `grep -r -- --benchmark-only .github/workflows/` returned nothing: the benchmark-gated job `conftest.py` has always cited never existed. That inverted the fix — from "mitigate a bad threshold" to "the threshold has no home, and my own earlier `--benchmark-disable` left it with none at all."
- **Measuring before deciding kept a target from being loosened.** All 58 pass serially on a quiet machine, so the numbers were never the problem. Loosening is the reflex that turned 10 ms into 100 ms in the Redis test and flaked again.
- **The runner data reversed my assumption.** I chose `continue-on-error` believing a shared runner would be slower than a laptop. It uses **13% of budget** — more headroom than my dev machine. The ~11× environmental spread on identical code (655 µs runner / ~3.8 ms loaded laptop / 7.57 ms after 10k tests) *is* the finding of #2156, and I only have it because the job ran.

### What didn't
- **The headline safety claim was decorative (#1859).** `test_promotion_is_all_or_nothing` scanned the whole file; `test-summary`'s echo block mentions every job. The reviewer shipped the complete half-promoted state — the exact #1713 trap the test's docstring quotes — and the suite stayed green. Reviewer-found, not self-found.
- **My first fix for it reported success on a broken measurement.** The re-check mutation used a first-occurrence `replace` and stripped a *different* job's `continue-on-error`. Third harness-error of this shape in the arc; the rule that keeps holding is that a mutation must assert it applied **to the intended target**, not merely that the text changed somewhere.
- **Sixth instance of the comment-vs-code hole** — in the PR that added `code_only()` for that class and then routed one of its own assertions through it. Adding a helper is not the same as adopting it.
- **`importorskip("yaml")` would have made all seven pins evaporate as `1 skipped`** with a green suite. I wrote a pin file whose failure mode was silence, in a PR about a check that was silently enforced nowhere.

### Recurring, now visible
Three separate defects in this one PR were the same shape: **a check that cannot see what it claims to guard** — the whole-file scan, the importorskip, and the original threshold itself. That is the through-line of #2139/#2154/#2155/#2156 as a group, and it is worth a canon entry rather than a fourth local fix.

### Verified
7 cases; 7/7 original gate-off rows plus both reviewer killers now fire; full suite 10,213 passed / 218 skipped; benchmark job green on two runner iterations; Gates 1, 2 and 4 all checked. Promotion tracked in #2160.

RETRO_COMPLETE

PR #2161, opened 2026-07-27
fix(test): stop asserting a time-bounded save as if it were guaranteed (#2154)

johnrtipton:
Addresses the last open piece of #2154 — the reproducible one.

## Root cause

`runtime.py` wraps the event-path state save in `asyncio.wait_for(_save(), timeout=0.150)` (#1475). That is deliberate: a slow session backend must never stall event handling. The consequence is that a save is **best-effort** — when the session DB write exceeds 150 ms the `TimeoutError` is swallowed, a warning is logged, and `liveview_<path>` never lands.

Five tests in `test_sticky_child_persistence_1471.py` asserted that key *is* present. They were asserting an unconditional outcome from a time-bounded operation, so they passed on every quiet machine and failed on a loaded one.

## A correction to my earlier diagnosis

I previously reported — here and in the issue — that this failed with all three test roots and passed with any two, presented as a clean distribution signal. **That bisection was confounded**: 2–3 concurrent worktree agents were running at the time. The failure is load-dependent, not distribution-dependent.

Established properly this time:

| check | result |
|---|---|
| 9 consecutive runs, quiet machine, `-n 2` … `-n 12` | never reproduced |
| at `95eb8ebf`, the commit where it originally failed | never reproduced → nothing had silently fixed it |
| 24 CPU spinners + 3 IO writers on 12 cores | **reproduced on demand** |

The suite also grew 10,203 → 10,213 during the drain, which reshuffles xdist and independently masked it. Two separate reasons it looked fixed while the cause was untouched — which is exactly why "I can't reproduce it any more" is not a fix.

## The fix

The 150 ms literal appeared inline twice. It is now `EVENT_STATE_SAVE_TIMEOUT_S`, named so a test asserting a save *landed* can raise it instead of racing a wall clock. A `generous_save_timeout` fixture does that for the five save-dependent tests, which then exercise **which keys the save writes** — the thing they exist to check — rather than how fast the session backend is.

**Production behaviour is unchanged.** The bound is still `0.150`.

Two pins keep it that way, because "raise it in tests" must not drift into "drop it in production":

- the bound exists, stays ≤ 1 s, and wraps **both** call sites;
- both sites swallow their `TimeoutError` rather than propagating it.

Symbol-migration grep (#1391): three existing tests pinned the literal `"timeout=0.150"`; all three now accept either spelling, and grep confirms no other consumer.

## Verification

- **Gate 3** (pollution-class, required by the bugfix template): **3/3 consecutive clean runs**, 10,215 each
- **10,215 passed under the exact load that reproduced the failure**
- Production-side gate-off:

| mutation | result |
|---|---|
| the `wait_for` bound removed entirely | 1 failed |
| one of two call sites left unbounded | 1 failed |
| bound raised to 30 s in production | 1 failed |
| one `TimeoutError` handler neutered | 1 failed |

- The fixture is load-bearing: with the production bound cut to 0.1 ms the whole file passes **with** it and fails 4 **without** it.

## Two gaps my own gate-off caught

Both were weak *mutations*, not weak tests — worth recording because the failure mode is a table of false greens:

- The `TimeoutError` pin used `in` while the string occurs **twice**, so neutering one site still passed. It now counts both.
- Two further save-dependent tests used a different accessor and my first scan missed them. Found by cutting the production bound to 0.1 ms and seeing what still failed — a sharper instrument than grepping for a pattern I had already guessed wrong once.


johnrtipton:
## Stage 11 findings addressed — both 🔴 were right, and the first one is the sharper lesson

**🔴 1 — the class is 13 tests, not 4.** I wrote *"symbol-migration grep (#1391): grep confirms no other consumer"* and presented it as the scope check. It was true of the **string** and false of the **bug class**. Worse: two of the three files holding the other 9 were **already open in my diff** — I edited their literal pins and left their save-dependent tests in exactly the shape this PR exists to fix, including `test_ws_event_save_block_writes_through_to_session`, the non-tautological integration test canon #1468 cites.

That is #1543 — single-variant coverage of a multi-variant surface is worse than none, because it makes the class look handled.

The class is now enumerated **empirically** (cut the bound to 0.1 ms, run all three roots) rather than by grep — the same instrument that found my two misses last round, which I should have reached for first instead of grepping for a pattern I had already guessed wrong once. The fixture moved to `python/djust/tests/conftest.py` and covers all 13.

**Verification is now decisive**: with the production bound at 0.1 ms across all three roots, the *only* failure in the entire suite is the pin that exists to detect a changed bound. Zero save-dependent tests remain exposed.

**🔴 2 — the handler pin inspected one site and the wrong text.**

```python
after = src.split("except asyncio.TimeoutError:", 1)[1][:400]   # first site only
assert "raise" not in after.split("logger.warning")[0]          # text BEFORE the log
```

Your three survivors all fail now:

| mutation | before | after |
|---|---|---|
| site-2 body → silent `pass` | survived | **1 failed** |
| `logger.warning(...); raise` (either site) | survived | **1 failed** |
| `timeout=CONST * 200` (effective 30 s) | survived | **1 failed** |
| bound drifts to 0.999 s | survived | **1 failed** |

The part worth recording: this is the *same* weakness I had diagnosed one function earlier. I fixed it for the **count** assertion, wrote a comment explaining why `in` was insufficient, and left `in`-shaped logic in the body assertion directly below it.

**🟡 4** — the bound is now pinned to **0.150 exactly**. You were right that the range alone let it drift to 0.999 s with every pin green, once the three pre-existing value-pins were relaxed to accept the symbol.

**🟡 3** — "five" was 4; my own verification line in the same PR said 4. Now 13, the real number.

**🟡 5** — I have not restructured the pins, but the count assertion's message now names the #1646 refactor and says what to change, so a contributor merging the two sites into one bounded helper is told what to update rather than left with a message describing a problem that no longer exists.

**🟡 6** — fixture moved to `conftest.py`, which also addresses the placement point.

### Verification
- Gate 3: **3/3 consecutive clean runs**, 10,215 each
- Class coverage: bound at 0.1 ms → only the bound-pin fails
- All four survivors fail; ruff clean

johnrtipton:
## Stage 14 Retrospective — PR #2161 (#2154)

**Task**: the reproducible member of the #2154 flake cluster.
**Quality**: 3/5 — the diagnosis was right and well-evidenced, but I shipped a fix covering 4 of 13 cases and called a string-grep the scope check.

### What worked
- **Refusing to accept "I can't reproduce it any more."** The suite grew 10,203 → 10,213 mid-drain, which reshuffled xdist and made it vanish. Checking out the original failing commit and finding it *also* passed proved nothing had fixed it — the machine had changed, not the code. Reproducing on demand under 24 CPU spinners + 3 IO writers turned it from folklore into a controlled experiment.
- **Correcting my own earlier bisection.** I had reported "all three roots fail, any two pass" as a clean distribution signal; it was confounded by concurrent agents. Saying so plainly cost nothing and would have misled whoever picked this up next.
- **The root cause is a real product invariant, not a test bug**: the save is deliberately bounded (#1475), which makes it best-effort, and five tests asserted it unconditionally.

### What didn't
- **I presented a string-grep as a class check.** "Symbol-migration grep (#1391): grep confirms no other consumer" was true of `"timeout=0.150"` and false of the bug class. Two of the three files holding the other 9 tests were *already open in my diff* — I edited their literal pins and left their save-dependent tests untouched. That is #1543: single-variant coverage makes the class look handled.
- **I had the right instrument and didn't use it.** Cutting the bound to 0.1 ms and running all three roots enumerates the class exactly; it is what found my two misses the round before. I reached for grep instead, and grep is what missed them the first time.
- **I repeated a weakness one function after diagnosing it.** I fixed `in` → `count(...) == 2` for the count assertion, wrote a comment explaining why `in` was insufficient, and left `in`-shaped logic in the body assertion directly below. Three mutations survived it.

### The through-line, now unmistakable
Across #2139, #2154, #2155 and #2156 the recurring defect is **a check that cannot see what it claims to guard** — a whole-file scan, an `importorskip`, a first-occurrence split, a threshold enforced only where it can't be trusted. Six of these were caught by adversarial review or gate-off, none by a green suite. The generalisable rule is that a *scope claim* needs the same empirical treatment as a behavioural claim: enumerate the class by running something, not by grepping for the shape you already assumed.

### Verified
13 tests covered (was 4); with the bound cut to 0.1 ms across all roots the only failure is the pin that detects a changed bound; Gate 3 3/3 clean at 10,215; all four reviewer-survivor mutations now fail; CI green including the new benchmarks job.

RETRO_COMPLETE

PR #2162, opened 2026-07-27
ci: promote the benchmarks job to fail the aggregate check (#2160)

### LINKED ISSUE #2160: promote the benchmarks job to a blocking merge gate
Follow-up to #2156 / PR #2159, which added the serial `benchmarks` job and shipped it `continue-on-error: true` per canon #1534 — a CI job exercising an environment the dev machine cannot reproduce stays non-gating until green on the runner.

## The precondition is already met

It ran green on its first runner iteration (PR #2159, job `90023756836`): **58 passed in 36.65s**, with far more headroom than expected.

| benchmark | runner median | target | budget used |
|---|---|---|---|
| `vdom_diff_list_reorder` | 655.6 µs | 5000 µs | **13.1%** |
| `vdom_diff_list_append` | 644.5 µs | 5000 µs | 12.9% |
| `websocket_mount_counter` | 11.38 ms | 100 ms | 11.4% |

The "permanently red-but-ignored" risk did not materialise — a shared runner has *more* headroom than a contended dev laptop.

## Why it still matters that it's non-gating

Job-level `continue-on-error: true` makes `needs.benchmarks.result == 'success'` **even when the job fails**. So today the only signal is an ✗ in the checks list that nothing reads, and a future collection error — `tests/benchmarks/` renamed, a plugin dropped — would be entirely invisible. The job currently proves the thresholds *can* be enforced; it does not enforce them.

## The change

1. Remove `continue-on-error: true` from the `benchmarks` job.
2. Add `benchmarks` to `test-summary`'s `needs`.
3. Add `[ "${{ needs.benchmarks.result }}" == "success" ] && \` to the AND-chain — **this is the load-bearing step**. A job in `needs` and echoed in the summary but absent from the AND-chain looks enforced and gates nothing (#1713).
4. Invert `test_the_job_is_non_blocking_until_green_on_the_runner` in `tests/test_benchmark_enforcement_2156.py` — the assertion is written to be inverted, not deleted.

`test_promotion_is_all_or_nothing` already enforces that 1–3 happen together: a half-promoted state fails 2 tests. That pin was itself decorative in PR #2159's first version — a reviewer shipped the full half-promoted state with a green suite — so it now scopes its scan to the AND-chain rather than the whole file.

## Suggested gate before doing it

Let the job run green on `main` for a few days first. One data point satisfies #1534's letter; a handful satisfies its intent, and the cost of being wrong is blocking every merge on a timing check.

johnrtipton:
Closes #2160.

## The precondition is met six times over

The job shipped `continue-on-error: true` under canon #1534 — a CI job exercising an environment the dev machine cannot reproduce stays non-gating until green on the runner. It has now been green **six consecutive times**:

| benchmark | runner median | target | budget used |
|---|---|---|---|
| `vdom_diff_list_reorder` | 656 µs | 5 ms | **13.1%** |
| `vdom_diff_list_append` | 644 µs | 5 ms | 12.9% |
| `websocket_mount_counter` | 11.4 ms | 100 ms | 11.4% |

One green run satisfies #1534's letter; six satisfy its intent — the bar I set in the issue itself. The runner has *more* headroom than a dev machine, the opposite of what I assumed when choosing `continue-on-error`.

## Why this matters beyond tidiness

Job-level `continue-on-error: true` makes `needs.benchmarks.result == 'success'` **even when the job fails**. So until now the job proved the thresholds *could* be enforced without enforcing them: its only signal was an ✗ in the checks list that nothing read, and a future collection error — `tests/benchmarks/` renamed, a plugin dropped — would have been entirely invisible.

## All-or-nothing, in one change

1. `continue-on-error` removed from the benchmarks block — a **scoped** edit, not a first-match replace. (An earlier mutation of mine did exactly that and stripped a *different* job's flag, producing a false green.)
2. `benchmarks` added to `test-summary`'s `needs`.
3. Added to the **AND-chain** — the load-bearing step. In `needs` alone it looks enforced and gates nothing (#1713).
4. `test_the_job_is_non_blocking_until_green_on_the_runner` **inverted**, as its own docstring instructed, not deleted.

## Gate-off

| mutation | result |
|---|---|
| removed from the AND-chain (half-promoted) | **1 failed** |
| `continue-on-error` restored while gating | **2 failed** |

`test_promotion_is_all_or_nothing` shipped in #2159 unable to see the half-promoted state — a reviewer proved it with a green suite — so this is the first promotion it can actually police.

One process note: my first run of this gate-off reported a **red baseline**, because the edit that inverted the pin also sliced out the `_gates_the_merge` helper sitting between the two tests. Caught by asserting the baseline is green before trusting any row, which is the habit this arc has repeatedly needed.

## Verification

- Full suite: **10,215 passed**, 218 skipped
- 7 cases in `tests/test_benchmark_enforcement_2156.py`
- This PR's own CI run is the seventh green benchmark iteration, now as a blocking gate


johnrtipton:
## Stage 11 findings addressed — 🔴 F1 was right, and I'd invoked the very canon I missed

**F1 🔴 — "blocking merge gate" is false.** Verified independently:

```
$ gh api repos/djust-org/djust/branches/main/protection -q '.required_status_checks'
(no required_status_checks key)
$ gh api repos/djust-org/djust/rules/branches/main -q 'length'
0
```

`main` requires a PR with 1 review, deletion and non-fast-forward protection — **no status check**. So a red `test-summary` does not stop a merge, for *any* check in this repo, and the convention here is `gh pr merge --admin` anyway.

That is #1713 one level up — and I cited #1713 by name in this PR while missing it at the layer above. Adding the job to the AND-chain is necessary and not sufficient.

Corrected everywhere it was claimed: the test is now `test_the_job_now_fails_the_aggregate_check`, and the title, docstring, CHANGELOG and PR body say what is true — a threshold breach turns the aggregate red, which is visible and actionable, and that is the whole of it.

**Filed #2163 rather than changing branch protection here.** That alters the merge policy for everyone with write access, and shouldn't be a side effect of a benchmark PR — it wants an explicit decision, plus a heads-up, since the first red run after enabling it blocks merges that previously went through.

**F2 🟡 — the pin saw only the job-level flag.** A step-level `continue-on-error` defeats the gate identically: step fails soft → job reports success → `needs.benchmarks.result == 'success'` → AND-chain passes with thresholds breached. This workflow uses that idiom **four times** (`test.yml:456,475,488,492`) and its own comments recommend it as the way to ship a new check, so a contributor following the repo's documented pattern lands exactly this shape. Now rejected; gate-off confirms **1 failed**.

**F3 🟡 — the comment above the job still said** *"continue-on-error until green on the runner … deliberately NOT in test-summary's AND-condition yet."* Both sentences became false in this PR. The #1867 class, in the file whose suite exists to stop that recurring. Removed.

**Figures** — you were right and the irony is not lost: the quoted medians mixed two runs and one was a **mean**, in a docstring about preferring the median. All three are now medians from a single runner job (`90062542012`): reorder **662 µs**, append **653 µs**, ws_mount **11.89 ms**.

**Also adopted from your review**: the six green runs span **four refs including two `main` pushes**, so they're independent rather than correlated same-branch runs — stronger evidence than I claimed, now stated.

Rows 6/7 (`${{ true }}`, `'true'`) I've left: `is not True` correctly accepts the legitimate `false` case, and a YAML expression in that position is a stretch next to the step-level hole, which is real and idiomatic here.

### Verification
7 pins; step-level hole gate-off **1 failed**; full suite **10,215 passed**; ruff clean.

johnrtipton:
## Stage 14 Retrospective — PR #2162 (#2160)

**Task**: promote the benchmarks job from non-gating to enforcing.
**Quality**: 3/5 — the mechanism is right and independently verified, but the headline claim was false and I cited the exact canon I'd violated.

### What worked
- **Waiting for evidence instead of promoting on one green run.** Six runs across four refs, two of them `main` pushes — independent, not correlated. The reviewer confirmed the budget figures weren't cherry-picked: 13.1% is the *worst* case across all eight threshold-asserting benchmarks.
- **All-or-nothing promotion, gated in both directions.** Removing it from the AND-chain fails 1; restoring `continue-on-error` while gating fails 2. That pin shipped in #2159 unable to see the half-promoted state, so this is the first promotion it could actually police.

### What didn't
- **I claimed "blocking merge gate" while `main` has no required status checks.** The AND-chain edit is necessary and nowhere near sufficient — a red `test-summary` blocks nothing here, for any check, and the house style is `--admin` merges. I invoked #1713 *by name* in this PR ("in `needs` alone it looks enforced and gates nothing") and then made the identical error one layer up. Filed as #2163 rather than changing branch protection as a side effect.
- **The pin saw only job-level `continue-on-error`.** A step-level flag defeats the gate identically — and this very workflow uses that idiom four times, with comments recommending it. Load-bearing on one axis, decorative on the adjacent (#1859/#1543).
- **I left a comment saying the opposite of what the code now did** ("deliberately NOT in the AND-condition yet"), in the file whose test suite exists to stop #1867 recurring.
- **I quoted a mean as a median**, in a docstring about preferring medians, and mixed figures from two runs into one table.

### The pattern this arc keeps producing
Every one of those four is the same shape: **a claim broader than the thing that was actually checked.** The AND-chain edit checked one layer and I claimed two. The pin checked one axis and I claimed the property. The figures came from two runs and I presented one table. That is the same defect as #2154's "grep confirms no other consumer" and #2156's whole-file scan — six instances across four PRs, none caught by a green suite, all caught by adversarial review.

The durable form: **a scope claim needs the same empirical treatment as a behavioural one.** Verify the layer you assert, not the layer you edited.

### Verified
7 pins; step-level gate-off 1 failed; full suite 10,215 passed; CI green including the benchmarks job as an enforcing member of the aggregate.

RETRO_COMPLETE

Issue #2164, opened 2026-07-27
[dj-virtual] client applier ignores keyed positioning — a middle insert lands at the tail


johnrtipton:
Blocks #2017 / ADR-026 iteration 3.

> **Correction.** This issue originally blamed missing windowing ("the list is not windowed at patch time, so there is no pool to apply the op to"). **That was wrong**, and the reasoning was unsound: it came from an A/B whose *control* arm also failed, which cannot distinguish "the flag does nothing" from "a shared failure masks it" — and the control failure turned out to be an artifact of the test's own scroll step. Re-run properly below.

## The measurement

60-row `[dj-virtual]` list at `/tests/virtual-keyed/`. Insert a row at **server position 5**, window scrolled away, and assert the **pool index** — not the pool size, which ADR-026 warns explicitly will pass with the fix disabled.

| | flag OFF | flag ON | expected |
|---|---|---|---|
| pool index of the new row | **60** (tail) | **60** (tail) | **5** |
| neighbourhood | `k3,k4,k5,k6,k7,k8` | same | `ins0` at 5 |
| tail | `[k58,k59,ins0]` | same | — |

The OFF arm landing at the tail is exactly what ADR-026 predicts for index-addressed ops against a windowed list, so this A/B **is** capable of distinguishing. The flag changes nothing.

## The differ is correct — iteration 1 works

```
flag=False   ops=['InsertChild']
flag=True    ops=['VirtualInsert']   key='ins0'   before_key='k5'
```

`VirtualInsert` carries precise keyed positioning: put `ins0` **before `k5`**. That is exactly what ADR-026 designed, and it is emitted correctly.

## The bug

**Iteration 2's client applier does not honour `before_key`.** It appends the new node at the tail of the item pool instead of splicing it in before `k5`.

Relevant client surface: `python/djust/static/djust/src/29-virtual-list.js` — `_virtualInsert`, `_virtualKeyedOp`, `insertionIndex`, `_flushVirtualKeyedOps`.

This is the exact trap ADR-026's Consequences section names:

> assertions on **order and position**, not counts. #2017 item 1's first tests counted pool size and passed with the fix disabled, because the #1989 absorb already produced the right count for the wrong reason.

The pool size *does* go 60 → 61 with the flag on. Only the index reveals it.

## Reproduce

```
.venv/bin/python -m uvicorn demo_project.asgi:application --host 127.0.0.1 --port 8002 --app-dir examples/demo_project
[...]

PR #2165, opened 2026-07-27
feat(vdom): give virtual_keyed_ops a Python switch — default stays OFF (#2017)

johnrtipton:
Advances #2017 / ADR-026 iteration 3. **Does not close it** — the browser gate said not to flip.

## Iteration 3 was never a one-line change

`djust_vdom::diff::set_virtual_keyed_ops` had **no Python surface at all** — its only callers were that crate's own tests. There was nothing for a Django user to switch, so "flip the flag after a soak" could not be done.

This ships the switch:

- a module-level `#[pyfunction]` **pair** (setter *and* getter — a setter alone can't be tested end to end)
- applied once in `DjustConfig.ready()` from `LIVEVIEW_CONFIG['virtual_keyed_ops']`

Module-level rather than a `RustLiveView` method, because the Rust side is a process-global `AtomicBool`, unlike the per-view `set_loop_render_cache_enabled` (#1967) — driving a process global from a per-view hook is last-view-wins. And deliberately **outside** the `PYTEST_CURRENT_TEST` guard that wraps hot-reload, so the flag holds the same value in tests as in production.

## The gate said don't flip

`ROADMAP.md:28` gates iteration 3 on *"real browser verification producing evidence worth flipping on."* The project had no `[dj-virtual]` demo, so I built one (`/tests/virtual-keyed/`, 60 rows, stable `dj-key`) and drove it in Chrome: scroll `k0` out of the window, edit it server-side.

| | flag OFF | flag ON |
|---|---|---|
| server handler ran | ✅ `log: "edited k0"` | ✅ |
| `EDITED` lands on a row | ❌ **no row at all** | ❌ **no row at all** |

**Identical.** The flag changed nothing end to end, so the default stays `False`.

## Where it actually breaks — and it isn't the differ

Driving `diff_html` directly with the flag on:

```json
[{"type":"VirtualUpdate","path":[],"d":"0","key":"k0","patches":[…]}]
```

The op is emitted and carries the **key**. ADR-026 iteration 1 is correct.

The blocker is upstream of applying it: **in a real page the list is not windowed at patch time.** After the #1610 WS mount morph the container held all 60 raw rows with no shell/spacer, and stayed that way until `initVirtualLists` was called by hand — even though `14-init.js:143` calls it on init. That's the #1989 symptom-1 shape. With no client-side pool, a keyed op has nothing to land on.

Filed as **#2164** with a copy-paste reproduction. **#2017 stays open.**

## Also here

- The demo page ships. The project had no `[dj-virtual]` example, and this one makes the failure legible on screen — visible keys, an `EDITED` marker, and insert-at-5 / remove-at-3 / reverse buttons for the other three cases.
- Both names added to the ADR-023 `_rust.pyi` stub and its `__all__`, which the type-checking integration test requires.
- ADR-026's status table now says iteration 3 is blocked on #2164 rather than merely "not shipped".

## Verification

- 8 cases; gate-off neutering `ready()`'s applier fails **2**
- Full suite **10,223 passed**, 218 skipped

Two of my own test defects, both caught rather than shipped:
- the startup test evaluated `want if True else cfg.get(...)` — it set the flag to whatever the test had already decided, and would have passed with `ready()` deleted. It now invokes the real method.
- the docs-comment pin scanned a fixed 900-character window, which broke the moment I added the #2164 cross-reference. **The pre-push hook from #2139 blocked that push and correctly reported the failure as new on this branch rather than blaming the merge-base** — its first catch of a regression I introduced. It now walks the contiguous comment block, since the property is "the key carries an explanation", not "the explanation fits in 900 bytes".


johnrtipton:
## The 🔴 was right, and re-running the gate overturned my recorded cause

You showed the experiment's **control arm also failed**, so it couldn't distinguish "the flag does nothing" from "a shared failure masks it" — and I'd written the conclusion into a permanent ADR anyway. Your prediction was exact: *if the list is genuinely not windowed, the OFF path's index-addressed patch resolves and lands.* It didn't, and that's a contradiction, not a result.

**The first failure was an artifact of my own scroll step.** Without it, flag OFF, the edit lands correctly on `k0` — and the list goes from 60 raw rows to 13 windowed as the patch applies.

### Re-run, measuring pool *index* (not size — ADR-026 warns counts pass with the fix disabled)

Insert at server position 5, window scrolled away:

| | flag OFF | flag ON | expected |
|---|---|---|---|
| pool index of new row | **60** (tail) | **60** (tail) | **5** |
| neighbourhood | `k3…k8` | same | `ins0` at 5 |

The OFF arm landing at the tail is what ADR-026 predicts, so **this** A/B can distinguish — and the flag still changes nothing.

### The cause is the client applier

```
flag=False   ops=['InsertChild']
flag=True    ops=['VirtualInsert']   key='ins0'   before_key='k5'
```

The differ emits precise keyed positioning. **Iteration 1 is correct; iteration 2's applier ignores `before_key`** and appends at the tail. #2164 retitled and rewritten around that; ADR-026, CHANGELOG and the Rust doc comment now agree. I left the withdrawn cause visible in the ADR with a note on why a failed control can't support a conclusion.

### Everything else

| finding | action |
|---|---|
| 🟡 2 `diff.rs:165` says the wiring "does not exist yet… do not cite it as if it does" | corrected — a comment about citation discipline that this PR invalidated |
| 🟡 3 `indent <= 16` decorative | **deleted**. You're right there's zero margin and the relocation yields 16 too; the real-invocation test already covers it |
| 🟡 4 "fails 2" | corrected in the PR body — you measured 3 and that matches |
| 🟡 5 `ready()` leaks a log filter (1→2→3) | snapshot/restore of `logger.filters` |
| 🟡 6 demo conventions | `templates/tests/`, extends `tests/base_test.html`, exported via `views/__init__.py` so the star import picks it up, **and listed in `test_index`** — it was unreachable before. Plus an on-page banner saying it documents a known failure |
| 🟡 7 fused comment blocks | separated |
| 🟡 8 "shipped in v1.1.0rc9+" | now "on `main` (unreleased)" |
| nit `as_text()` justification false | corrected — `diff_html` is `PyResult<String>`, always `str` |

Also reverted six demo files a `ruff --fix` had reformatted incidentally (#1079).

**Verification**: 10,222 passed; both `/tests/` and `/tests/virtual-keyed/` dogfooded at 200; Rust rebuilt.

The decision — don't flip — is unchanged. What changed is that it's now supported by an experiment that could have come out the other way.

johnrtipton:
## Stage 14 Retrospective — PR #2165 (#2017 ADR-026 iteration 3)

**Task**: finish #2017 — wire the config and flip the flag.
**Outcome**: wiring shipped; flag deliberately **not** flipped; #2017 stays open.
**Quality**: 2/5 — the deliverable is sound, but I recorded an unsupported cause in a permanent ADR and only a reviewer caught it.

### What worked
- **The gate did its job.** Unit tests pass for ADR-026 iterations 1 and 2 independently. Only a browser showed they don't compose. Flipping on a green suite would have shipped a behaviour change to every `dj-virtual` user that does nothing, while the real defect stayed hidden.
- **Building the switch was the right deliverable even though the flip wasn't.** The flag had *no Python surface at all* — "flip after a soak" was never a one-line change. Now it is, once #2164 lands.
- **Asserting position, not count.** The pool goes 60→61 with the flag on either way; only the *index* (60 vs the wanted 5) reveals the bug. ADR-026's Consequences section warned about exactly this trap and I'd have fallen into it without re-reading it.

### What didn't
- **I wrote a cause into a permanent ADR from an experiment whose control arm also failed.** The reviewer's reasoning was one step and decisive: if the list isn't windowed, the OFF path's index-addressed patch *should* land — so "not windowed" predicts OFF works, and OFF failed. That's a contradiction, not a finding. An ADR saying "inconclusive" would have been strictly better than one asserting a cause the data couldn't support.
- **The control failure was my own test's scroll step.** I introduced a confound and then explained the result it produced.
- **The ADR correction never got staged.** My `git add` listed code paths and missed `docs/adr/`, so the branch carried the withdrawn cause in the one file where being wrong lasts longest. Only the pre-push hook's "Unstaged files detected" warning caught it.
- Three more of my own defects the reviewer found: an `indent <= 16` pin with zero margin (the relocation it guards against also yields 16), a `ready()` call leaking a log filter per invocation, and a false justification comment on `as_text()`.

### The pattern, and it's the same one as all session
Every defect here is **a claim broader than what was actually checked** — an A/B whose control I never validated, an indent threshold that couldn't distinguish two states, a comment asserting a type contract I hadn't read. Same shape as #2154's string-grep-as-class-check and #2156's whole-file scan.

The sharpened version, for a *negative* result specifically: **a control arm that fails is not a baseline, it's a second bug.** Before concluding "X changes nothing," verify the no-X case behaves as theory predicts. If it doesn't, you've found something else and learned nothing about X.

### Verified
8 → 7 cases (one decorative pin deleted); 10,222 passed; both demo routes dogfooded at 200; CI 16/16; 0 behind main. #2164 retitled and rewritten around the measured cause.

RETRO_COMPLETE

Issue #142, opened 2026-01-30
VDOM: data-djust-replace does not remove old children on patch


johnrtipton:
## Bug

`data-djust-replace` on a container element does not properly remove old children when the VDOM patches new content in. Instead of replacing all children, the old nodes persist alongside newly inserted ones.

## Reproduction

1. A container with `data-djust-replace` and a conditional child:
```html
<div id="messages" data-djust-replace>
    {% if not messages %}
    <div class="empty-state">No messages</div>
    {% endif %}
    {% for msg in messages %}
    <div class="message">{{ msg.content }}</div>
    {% endfor %}
</div>
```

2. Initial render: `messages = []` → empty state div renders
3. Event handler sets `messages = [msg1, msg2, msg3]`
4. VDOM patch inserts the 3 message divs but does **not** remove the empty state div

## Expected

`replace_all_children` in `diff.rs` should remove all old children then insert all new ones. The Rust diff logic looks correct (`RemoveChild` patches for all old, `InsertChild` for all new), so the bug is likely in the **client-side JavaScript** patch application — the `RemoveChild` patches may not be targeting the correct DOM nodes.

## Observed

After patching:
- Old conditional nodes remain in the DOM
- New nodes are appended alongside them
- Results in duplicate elements (e.g., two "Regenerate" buttons, empty state showing alongside messages)

## Workaround

Use always-present elements with class toggling instead of `{% if %}` conditionals:

```html
<div class="empty-state {% if has_messages %}hidden{% endif %}">
    No messages
</div>
```

With a boolean context variable (`has_messages = bool(self.messages)`) — the VDOM correctly diffs class attribute changes.

## Environment

- djust 0.2.0
- Discovered in djust_chat app during conversation switching

johnrtipton:
Issue is not fully resolved. After PR #143, `data-djust-replace` still causes elements to be inserted into the wrong parent node.

**Repro:** A container with `data-djust-replace` (`#messages`) is a sibling of another container (`.chat-input`). When VDOM patches insert new children into the `data-djust-replace` container, some children end up appended to the sibling container instead.

**Example DOM before patch:**
```html
<div id="messages" data-djust-replace>
  <div class="messages__empty">...</div>
</div>
<div class="chat-input">
  <div class="chat-input__container">...</div>
</div>
```

**Expected after patch (2 messages inserted):**
```
#messages → [empty_state, message_user, message_assistant]
.chat-input → [input_container]
```

**Actual after patch:**
```
[...]

Issue #147, opened 2026-01-31
Fix MoveChild in Rust apply_patch test helper


johnrtipton:
## Summary

The Rust `apply_patch()` in `patch.rs` applies `MoveChild` by doing `remove(from)` then `insert(to)`. When multiple `MoveChild` patches are applied sequentially, earlier moves shift indices for later ones, producing incorrect results.

This prevents using the gold-standard `diff_apply_verify` pattern (diff → apply → structural compare) for keyed diff tests.

## Current Behavior

```rust
// patch.rs:58-65
Patch::MoveChild { path, from, to, .. } => {
    let node = target.children.remove(*from);
    target.children.insert(*to, node);  // 'to' may be stale after prior moves
}
```

## Proposed Fix

Option A: Resolve target nodes by `djust_id` instead of index (matching client JS behavior).

Option B: Collect all moves, compute final positions, apply in a single pass.

Option A is preferred since it mirrors the actual client-side resolution strategy and would validate that `djust_id` values are correct.

## Impact

Enables full round-trip verification (`diff_apply_verify`) for keyed diff tests, increasing confidence in the keyed algorithm's correctness.

## Context

Found during VDOM torture testing. The test `torture_keyed_move_apply_verify_simple_swap` documents this limitation with a comment explaining why `diff_apply_verify` can't be used.

PR #2173, opened 2026-08-06
fix(ci): move CI to Node 24 so jsdom 30 can land, closing 5 undici advisories (#2169)

johnrtipton:
Supersedes #2169 (dependabot's jsdom bump, which could not pass on its own).

## What was wrong

Dependabot #2169 (`jsdom 29.1.1 → 30.0.0`) failed **140 of 170** JS test files with:

```
TypeError: webidl.util.markAsUncloneable is not a function
 ❯ new CacheStorage node_modules/undici/lib/web/cache/cachestorage.js:20:17
 ❯ Object.<anonymous> node_modules/jsdom/lib/api.js:12:33
```

**The lockfile was not at fault.** jsdom 30 requires `undici ^8.7.0`, and dependabot's lockfile resolves undici to **8.9.0** — exactly the patched version for all five open undici advisories.

**The runner was at fault.** `markAsUncloneable` is a `node:worker_threads` API added in **Node 22.10**, and undici 8.9.0 declares `engines.node >=22.19.0`. All three `setup-node` steps pinned `node-version: '20'`.

Node 20 is also past EOL (2026-04-30), so this bump was due regardless. Moved to **'24'** (current Active LTS) rather than 22 (in maintenance since Oct 2025).

## Advisories closed

| Alert | Severity | GHSA | Patched |
|---|---|---|---|
| #140 | high | GHSA-4cwx-7wf7-3272 | 8.9.0 |
| #144 | medium | GHSA-m8rv-5g2x-5cg5 | 8.9.0 |
| #143 | medium | GHSA-v3r7-h72x-cjcm | 8.9.0 |
| #142 | medium | GHSA-jr45-8vmc-qm54 | 8.9.0 |
| #141 | medium | GHSA-8xcm-r25x-g524 | 8.9.0 |

All are **dev-scope** — undici reaches the tree only through jsdom, so no shipped artifact was ever affected.

## Reproduction (both directions, locally — not inferred)

- **Node 20.19.0** + this lockfile → byte-identical failure to CI (169/170 files down).
- **Node 24.19.0** + the same lockfile → clears.

## The one genuine jsdom 30 behaviour change — and it was ours

`25-page-metadata.js` stored `CSS.escape` as a **detached** reference:

```js
const cssEscape = (typeof CSS !== 'undefined' && CSS.escape)
    ? CSS.escape                       // detached — loses the receiver
    : function (s) { ... };
```

jsdom 30 generates `CSS` from WebIDL and brand-checks the receiver, so the detached call throws `'escape' called on an object that is not a valid instance of CSS` — 5 tests.

This module was **the odd one out of three**: `03-websocket.js:29` (`(v) => CSS.escape(v)`) and `45-child-view.js:58` (`return CSS.escape(value)`) both already call it as a method. That makes this #1646 again — an invariant implemented correctly in two of three places — so the fix aligns the third rather than special-casing it.

**No source-grep pin was added.** The 5 behavioural cases in `page-metadata.test.js` go red without this fix under jsdom 30, so a grep would be a strict subset with false-positive hazards — the same reasoning that removed one in #2167.

## Verification

- `npm test` on Node 24.19.0: **170/170 files, 1919/1919 tests**.
- `scripts/check-bundle-init-order.mjs`, `scripts/check-cross-iife-refs.mjs`: clean.
- `eslint --max-warnings 0`: clean.
- `scripts/check-doc-snippets.py`: size/version claims still match (shipped stays 58.7 KB gz).
- Rebuilt bundle diff is **exactly** the source change, confirming the committed bundle was not stale.

## Files

| Change | Why |
|---|---|
| `.github/workflows/test.yml` (×2), `pre-release-security-audit.yml` (×1) | `node-version: '20'` → `'24'` |
| `package.json`, `package-lock.json` | jsdom 30.0.0 / undici 8.9.0 (taken verbatim from #2169) |
| `src/25-page-metadata.js` | call `CSS.escape` as a method |
| `client.js`, `client.min.js`, `client-sizes.json` | rebuilt bundle |


johnrtipton:
## #1236 release-workflow risk review

This PR touches `pre-release-security-audit.yml`, so the gate applies. Reviewed as the gate asks — release notes, breaking changes, Node version requirements:

**The change**: one line, `node-version: '20' → '24'` on the `js-security` job. No action versions change, no step logic changes, no permissions change.

**Node version requirement** — this is the whole point of the PR. `undici@8.9.0` (pulled in by `jsdom@30`, and the patched version for 5 open advisories) declares `engines.node >=22.19.0`. Node 20 also passed EOL on 2026-04-30. Moved to `'24'` (current Active LTS) rather than `'22'` (in maintenance since Oct 2025).

**Breaking changes** — the job runs exactly three things after `npm ci`, all verified locally on Node 24.19.0:
- `npm audit` — unaffected by the Node major.
- `npm run lint` (eslint 10.8.0) — clean, `--max-warnings 0`.
- `npm test` (vitest 4.1.10 + jsdom 30) — 170/170 files, 1919/1919 tests.

**Blast radius if wrong**: this job is `js-security` in the pre-release audit. A Node incompatibility surfaces as a red job on the audit run, before a release is cut — not as a bad artifact. `npm ci` would fail loudly rather than silently skipping the scan.

**Risk: low.** Adding `release-workflow-reviewed`.

johnrtipton:
## Status: HELD pending a real CI runner

GitHub Actions has been in a **critical, unresolved incident since 15:22 UTC**. Its siblings #2174 and #2175 were merged on local validation, but **this PR is deliberately not**, because it changes the runner itself (Node 20 → 24) — exactly the case canon #1534 covers: a CI job exercising an environment the dev machine cannot fully mirror gets at least one runner-only iteration before it is trusted. Local Node 24.19.0 is good evidence; it is not `ubuntu-latest` with a cold `npm ci`.

Merge this once Actions recovers and the checks are green.

## Also now carries the postcss alert

`postcss 8.5.20 → 8.5.25` (Dependabot **#147**, GHSA-fxqj-rqcc-2cmp — attacker-controlled `sourceMappingURL` reads arbitrary `.map` files when `from` is unset; incomplete fix of GHSA-6g55-p6wh-862q).

Folded in rather than shipped separately because it edits the same `package-lock.json` and would otherwise conflict for no benefit. Dev-scope: postcss is transitive under `vite` (via vitest), whose `^8.5.16` range already admits the patched version.

**One correction worth recording.** The first attempt used `npm install postcss@^8.5.23`, which added postcss to `package.json` as a **runtime** `dependencies` entry and dropped its `dev: true` marking in the lock — wrong on its own terms, and against the manifesto's "no node_modules / no npm for consumers" position. Redone with `npm update postcss --package-lock-only`, which keeps it transitive. The commit now contains a 3-line lockfile diff and nothing else; `package.json` is untouched.

## Combined verification on Node 24.19.0

- `npm test`: **170/170 files, 1919/1919 tests**
- `eslint --max-warnings 0`: clean
- bundle init-order + cross-IIFE guards: clean
- Already-tagged CHANGELOG sections verified byte-identical to `v1.1.0rc9` / `v1.1.0rc8` after merging `main` (canon #2028)

## Alerts this PR closes when it lands

| Alert | Sev | Package |
|---|---|---|
| #140 | high | undici |
| #141–#144 | medium | undici |
| #147 | medium | postcss |

johnrtipton:
**Note for whoever picks this up:** the queued check runs on this PR were **dropped** during the Actions outage — the status-check rollup is now empty rather than pending, which matches GitHub's warning that "jobs may remain queued for an extended period before starting or may time out."

So this will not go green on its own when Actions recovers. It needs the checks re-triggered — either:

```bash
gh workflow run  # or:
git commit --allow-empty -m "ci: re-trigger checks after the Actions outage" && git push
```

or close/reopen the PR.

Once `test-summary` is green, this is ready to merge — it closes Dependabot #140–#144 (undici) and #147 (postcss).

johnrtipton:
## Runner validation obtained — merging

This PR was held specifically because it changes the runner itself (Node 20 → 24), and canon #1534 wants one runner-only iteration before that is trusted. Actions has recovered and the re-triggered run is green on a real `ubuntu-latest`:

```
javascript-tests                    SUCCESS   <- the one that mattered
browser-smoke                       SUCCESS
demo-checks                         SUCCESS
rust-tests                          SUCCESS
python-tests (py3.12)               SUCCESS
python-tests (py3.14t free-threaded) SUCCESS
test-summary                        SUCCESS
```

`javascript-tests` green on Node 24 with jsdom 30 / undici 8.9.0 is the proof local validation could not supply — a cold `npm ci` on the runner, not my machine.

Closes Dependabot **#140** (high), **#141–#144** (medium, undici) and **#147** (medium, postcss). Supersedes #2178.

Issue #146, opened 2026-01-31
Add proptest/fuzzing for VDOM diff algorithm


johnrtipton:
## Summary

All existing VDOM tests use hand-crafted trees. Adding property-based testing with `proptest` or `quickcheck` would catch edge cases that manual tests can't anticipate.

## Proposed Tests

1. **Identity**: `diff(A, A)` always produces 0 patches
2. **Round-trip correctness**: `apply(A, diff(A, B))` structurally equals B
3. **No panics**: Arbitrary tree pairs never cause panics
4. **Patch count bounds**: Number of patches ≤ total nodes in both trees

## Implementation

Add `proptest` as a dev dependency to `crates/djust_vdom/Cargo.toml` and create a `tests/fuzz_test.rs` with random VNode tree generators:

- Random depth (0-20)
- Random width (0-10 children per node)
- Random tag names from a small set (div, span, p, li, etc.)
- Random attributes (0-5 per node)
- Random keys (some keyed, some not)
- Random text content

## Context

The 58 torture tests all pass, but they test known patterns. Fuzzing would explore the combinatorial space of tree structures and catch unknown edge cases.

PR #2174, opened 2026-08-06
security(deps): cryptography 49.0.0 → 50.0.0, pyOpenSSL 26.3.0 → 26.4.0 (#146)

johnrtipton:
Closes Dependabot alert **#146** (high) — the only **runtime-scope** alert open against the project.

> `cryptography`: PKCS#7 `EnvelopedData` decryption exposes a Bleichenbacher oracle through distinguishable errors and timing.

Patched in `cryptography 50.0.0`.

## The bump does not stand alone

The locked **pyOpenSSL 26.3.0** pins `cryptography<50,>=49.0.0` — so cryptography 50 is simply unreachable until pyOpenSSL moves to **26.4.0** (`cryptography<51,>=49.0.0`). Both are upgraded together.

The other five dependents already accept it, so nothing else moves:

| Dependent | Constraint | 50.0.0 |
|---|---|---|
| `pyopenssl` **26.3.0** | `cryptography<50,>=49.0.0` | ❌ — hence the paired bump |
| `pyopenssl` **26.4.0** | `cryptography<51,>=49.0.0` | ✅ |
| `autobahn` 25.12.2 | `>=3.4.6` | ✅ |
| `service-identity` | `>=47` | ✅ |
| `pyjwt` | `>=3.4.0` | ✅ |
| `google-auth` | `>=38.0.3` / `>=41.0.5` | ✅ |
| `azure-storage-blob` | `>=2.1.4` | ✅ |

## The 125-line registry delta is not part of the security fix

Every `registry` entry in `uv.lock` had been rewritten to a **developer-local caching proxy** (`http://127.0.0.1:8418/pypi/simple/`). It leaked in via `e39a9242` ("chore: clean working tree for release") — `git show e39a9242^:uv.lock` records the correct `https://pypi.org/simple` on all 150 entries. The source is a machine-local global `~/.config/uv/uv.toml`.

Nothing was broken in practice: `uv` resolves from the per-wheel `url` fields, which always pointed at `files.pythonhosted.org`. That is why CI never noticed, and why this is a latent rather than active problem. Re-locking with an explicit index restored it.

**Worth knowing for whoever locks next:** unless `UV_INDEX_URL` is set explicitly, that global config will re-introduce the localhost URLs on the next `uv lock`.

## Verification (run, not assumed)

- **Full Python suite** with the upgraded pair installed: **10227 passed, 218 skipped, 0 failed**.
- **Consumer smoke test** on the surfaces that actually use it:
  - RSA sign/verify round-trip (`hazmat` primitives),
  - `OpenSSL.SSL.Context(SSL.TLS_METHOD)` construction,
  - `import service_identity.pyopenssl` — daphne's TLS verification path.
- Version-change scope confirmed minimal: `git diff uv.lock | grep '^[+-]version'` shows exactly `49.0.0→50.0.0` and `26.3.0→26.4.0`, nothing else.


johnrtipton:
Merging on local validation: GitHub Actions has been in a **critical, unresolved incident since 15:22 UTC** (~3h), with GitHub warning that queued jobs may time out. All 41 jobs across the open PRs — and runs on `main` — are stuck at `QUEUED`, so a green CI run is not obtainable.

This follows the repo's established practice for an Actions-unavailable window (see the v1.0.7-3/-4 security-audit arc: "PRs merged on local validation … later confirmed by the green release CI").

Local evidence for this PR:
- Full Python suite with the upgraded pair installed: **10227 passed, 218 skipped, 0 failed**.
- Consumer smoke test: RSA sign/verify round-trip, `OpenSSL.SSL.Context(SSL.TLS_METHOD)`, `import service_identity.pyopenssl` (daphne TLS path).
- Version-change scope confirmed minimal (`49.0.0→50.0.0`, `26.3.0→26.4.0`, nothing else).
- Pre-push hooks ran clean.

The sibling PR #2173 (CI Node 20→24) is deliberately **held** until a real runner can validate it — it changes the runner itself, which is exactly the case canon #1534 says needs one runner-only iteration.

PR #2175, opened 2026-08-06
security(validation): cap the client-supplied parameter name in the near-miss warning (#2579)

johnrtipton:
Closes CodeQL code-scanning alert **#2579** (`py/log-injection`, medium) at `python/djust/validation.py:293`.

## The alert was right, but not for its stated reason

CodeQL's framing — forged log records via CR/LF — was **already handled**: the near-miss warning formats with `%r`, and `repr()` escapes CR/LF. Alert triaged as a false positive on that axis.

What turned up was the claim the code made in its own defence:

```python
# `key` and `match` need no length cap: a near miss must be a
# prefix/suffix of a declared parameter, so neither can be long.
```

**That is false.** `_is_near_miss` *normalises before it compares*. Against a declared `field_name`, three independent routes produce an arbitrarily long near miss:

| Route | Payload | Branch |
|---|---|---|
| underscore padding | `f_i_e_l_d_n_a_m_e` + 20 000 `_` | `a == b` (underscores stripped from both sides) |
| suffix padding | `field_name` + 20 000 `X` | key becomes `long_`, matched by `startswith` |
| normalised prefix | `fieldname` + 20 000 `9` | same branch, no underscore in the payload at all |

Measured before writing any fix:

```
declared parameter: 'field_name'
  underscore padding  len= 50017  near_miss=True
  suffix padding      len= 50010  near_miss=True
  prefix-of-key       len= 20009  near_miss=True
```

Parameter names come straight off the wire, so uncapped this is an **unbounded client-controlled write to the log on every event**. `repr()` bounds nothing.

## The warn-once memo was a second instance, not a mitigation

`pair = (module, name, key)` includes the key, so **varying the padding mints a fresh entry every request** rather than deduplicating — the cache never suppresses an attacker, it only evicts. And it retained the *full* key, so 512 entries could hold 20 KB apiece.

That is exactly the memory-exhaustion vector `test_the_cache_is_bounded_against_client_controlled_names` was written to close. It stayed open because that test bounds the entry **count** using short names (`field_nam{i}`) and never the entry **size**.

## Fix

Cap once into `safe_key`, used for **both** the memo key and the log record.

- `_is_near_miss` still receives the **full** key → matching semantics unchanged.
- For a normal identifier `sanitize_for_log` is the identity → common path untouched.
- `match` stays raw, now documented as the one field here that is *not* client-influenced (it comes from `accepted`, the handler's own declared parameters).
- `sanitize_for_log` is already registered as a `log-injection` sanitizer in `.github/codeql/models/djust-sanitizers.model.yml`, so this clears the alert as well as the bug.

The same false claim was asserted in the test file — that capping the key was *"inert"* — and is corrected there too.

## Verification

Reproduce-first: all 4 new cases fail against unfixed code (`FFFF`), then pass.

**Two mechanisms, each independently reachable — neither shadows the other:**

```
BASELINE (fix in place)
  long_parameter_name (log line bounded)  -> 0 failed, 3 passed
  stores_bounded_keys (memo bounded)      -> 0 failed, 1 passed

GATED OFF: log-arg
  long_parameter_name                     -> 3 failed, 0 passed
  stores_bounded_keys                     -> 0 failed, 1 passed   <- unaffected

GATED OFF: memo-key
  long_parameter_name                     -> 0 failed, 3 passed   <- unaffected
  stores_bounded_keys                     -> 1 failed, 0 passed
```

The harness asserts the mutation text was found and that the mutated source actually differs, and treats a collection error as `INVALID` rather than reporting it as a count (canon #2135) — a `SyntaxError` cannot masquerade as "0 failed".

Full suite: **10231 passed, 218 skipped, 0 failed**.


johnrtipton:
Merging on local validation — GitHub Actions has been in a **critical, unresolved incident since 15:22 UTC**, with queued jobs at risk of timing out. Same rationale and precedent as #2174.

Local evidence:
- Reproduce-first: all 4 new cases fail (`FFFF`) against unfixed code, then pass.
- Two-mechanism gate-off, neither shadowing the other:
  - gate off the **log arg** → `long_parameter_name` 3 failed, memo case still green;
  - gate off the **memo key** → memo case 1 failed, `long_parameter_name` still green.
  - Harness asserts the mutation text was found and that the source actually changed, and reports `INVALID` on a collection error rather than a count (canon #2135).
- Full suite: **10231 passed, 218 skipped, 0 failed**.
- Rebased on `main` post-#2174; verified per canon #2028 that the already-tagged `[1.1.0rc9]` section is **byte-identical** to `git show v1.1.0rc9:CHANGELOG.md`, so the union merge did not misattribute anything into a shipped section.

Issue #2181, opened 2026-08-10
ROADMAP: 113 shipped rows in completed v1.1.0-N buckets are not struck through


johnrtipton:
Surfaced during the 1.1.0 final release prep (PR #2180), where the pre-release roadmap audit runs.

## What

134 `**Pn**` table rows across the `v1.1.0-N` drain buckets carry no `~~strikethrough~~`/✅, so they read as open work. Checking every issue they reference:

- **128** unique issues referenced
- **113 CLOSED** — the rows are stale
- **2 OPEN**, both legitimately deferred and documented as such:
  - **#2017** — ADR-026 iteration 3 (flag ships wired but OFF, #1122 split-foundation)
  - **#1561** — bug-capture iter C, `priority:low`
- **13** resolve to nothing (pre-#963 refs / PR numbers)

## Why it was not fixed in the release PR

Each completed bucket already carries an authoritative `**COMPLETE n/n ✅**` line under its heading, so the roadmap does not actually advertise open work to a careful reader — the staleness is cosmetic at row level. A 113-row mechanical strike-through inside a release cut is scope creep with real diff risk and no release-blocking benefit (#1079). PR #2180 instead corrected the genuinely-wrong part (the header still read *"Current version: 0.9.7 (released 2026-05-16)"*, four minor versions stale) and recorded the row-level gap explicitly.

## The fix

Single-script transformation per #1312, not incremental edits: for each `**Pn**` row whose referenced issue is CLOSED, wrap the summary cell in `~~…~~` and append ` ✅`, leaving the two open items alone. Reproduce the audit with:

```bash
gh issue list --state all --limit 500 --json number,state
```

then join against the row refs.

Worth doing before the next roadmap-driven planning session, since `/pipeline-next` reads these rows to pick work.

Issue #2163, opened 2026-07-27
main has no required status checks, so a red test-summary does not block a merge


johnrtipton:
Found during the Stage 11 review of PR #2162 (#2160), which promoted the `benchmarks` job into `test-summary`'s AND-chain and claimed that made it a "blocking merge gate". **It does not**, and the claim has been corrected in that PR.

## The gap

```
$ gh api repos/djust-org/djust/branches/main/protection -q '.required_status_checks'
(no required_status_checks key)

$ gh api repos/djust-org/djust/rules/branches/main -q 'length'
0
```

Branch protection on `main` requires a pull request with 1 review, plus deletion and non-fast-forward protection. **No status check is required.**

So the whole `test-summary` aggregate — `rust-tests`, `python-tests`, `javascript-tests`, `browser-smoke`, `nav-hooks-guard`, `security-tests`, `demo-checks`, and now `benchmarks` — can be red and the merge button stays live.

This is #1713 one level up. That canon says the load-bearing question is not "is it a job?" but *"does a failure actually fail the merge?"* Here the answer is no, for every check in the repo.

## Why it matters more than it looks

- The repo's convention is `gh pr merge --admin`, which bypasses even the review requirement. A red aggregate is not merely unenforced; it is easy to merge past without noticing.
- It makes several pieces of tooling half-true. #2139's pre-push attribution and the `main-health` workflow both exist to keep `main` green, and the last line of defence — refusing the merge — is absent.
- It quietly weakens every "this now gates X" claim in the CHANGELOG.

## The change

Add `test-summary` to `main`'s required status checks. Aggregating through one job is already the design, so it is a single entry rather than eight.

**Deliberately not done in PR #2162.** Branch protection changes the merge policy for everyone with write access and is not something to alter as a side effect of a benchmark PR — it wants an explicit decision, and possibly a heads-up to contributors, since the first red run after enabling it will block merges that previously went through.

Worth deciding at the same time:
- whether `--admin` merges should remain routine once checks are required;
- whether the informational jobs (`playwright-tests`, `security-scan`) stay out of the aggregate, which they currently do, correctly.

## Related
- PR #2162 / #2160 — where this surfaced; its language now says "fails the aggregate check" rather than "blocks the merge"
- #1713 — being in `needs` is not the same as gating

johnrtipton:
Re-verified today — the diagnosis is still exactly current:

```
$ gh api repos/djust-org/djust/branches/main/protection -q .required_status_checks
null

[...]

PR #2182, opened 2026-08-11
docs(roadmap): open v1.1.1-1 with the 2 open process-debt issues (#2163, #2181)

johnrtipton:
Opens the post-1.1.0 drain bucket. Docs-only.

## Scope

| Priority | Issue | Why it buckets here |
|---|---|---|
| P2 | #2163 | `main` has no required status checks — the whole `test-summary` aggregate can be red and the merge button stays live |
| P2 | #2181 | 113 shipped ROADMAP rows are not struck through |

Both are process/CI debt surfaced *by* the 1.1.0 release prep, and neither changes framework behaviour — hence **1.1.1** (patch) rather than 1.2.0.

## Deferred, recorded rather than omitted

- **#2017** — dj-virtual ADR-026 iteration 3. The flag ships wired but OFF; flipping it changes VDOM behaviour for every `[dj-virtual]` user, which is what #1122 split-foundation exists for.
- **#1561** — bug-capture iter C, `priority:low`, multi-day Redis-store feature.

## A prerequisite #2163 did not originally record

`test.yml` carries `paths-ignore` for `**.md`, `docs/**`, `CHANGELOG.md`. A **required** status check that never reports leaves a PR at *"Expected — Waiting for status to be reported"* indefinitely — so making `test-summary` required would render docs-only PRs unmergeable through the normal path. This repo's two-commit convention (impl+tests / docs+CHANGELOG) makes docs-only PRs a normal shape, not a rarity.

Order is therefore: always-reporting companion job → confirm the context name on a green `main` run → then add `test-summary` to required checks. Captured in the milestone body so the implementer does not rediscover it.

## Verified at drain time

```
$ gh api repos/djust-org/djust/branches/main/protection -q .required_status_checks
null
$ gh api repos/djust-org/djust/rules/branches/main -q length
0
```

`enforce_admins` is `false`; the two active rulesets contribute no rules to `main`, so classic branch protection is the only mechanism. Staleness check on both issues: every cited path/symbol resolves (`paths-ignore` at `test.yml:6,12`, `test-summary:` at `:650`), and #2181's row count re-derives at 134.

### DIFF
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -28,6 +28,37 @@ Two name shapes appear in this roadmap, with distinct meanings:
 
 **Released**: `v0.9.1` cut 2026-04-30 (tag `v0.9.1`, GitHub Release published, PyPI live). Bundles 8 drain buckets + post-cleanup. Retro: RETRO.md §v0.9.1. Tracker carryovers (#1234, #1235, #1236) and the post-release SSE bug bundle (#1237) move into `v0.9.2-1` below.
 
+## v1.1.1-1 — post-1.1.0 process drain: merge-gate enforcement + roadmap accuracy (drain bucket → ships in 1.1.1)
+
+Opened 2026-08-10, immediately after the 1.1.0 final cut (PR #2180). Both items
+are process/CI debt surfaced BY that release prep — neither changes framework
+behaviour, which is why they bucket into a patch release rather than 1.2.0.
+
+Deferred and NOT in this bucket: **#2017** (dj-virtual ADR-026 iteration 3 — the
+flag ships wired but OFF; flipping it changes VDOM behaviour for every
+`[dj-virtual]` user and #1122 exists for exactly that) and **#1561**
+(bug-capture iter C, `priority:low`, multi-day Redis-store feature).
+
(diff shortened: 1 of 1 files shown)


johnrtipton:
## Retrospective — PR #2182 (pipeline-drain)

**Task**: v1.1.1-1 — open the milestone (docs-only).

### Quality: 5

### What went well
- The milestone records a prerequisite the issue itself did not: making `test-summary` required would strand docs-only PRs. Capturing it in the bucket meant the implementer did not rediscover it.
- #2017 and #1561 are written down as *deliberately deferred with reasons* rather than silently omitted, so a later reader can tell the difference between "not done" and "decided against for now."

### What didn't
- Nothing of substance. The one thing worth noting is that this PR became the drain's own test fixture — it is the docs-only shape #2163 is about, and its check list (1 check, no `test-summary`) was the evidence that justified the shim. That was luck, not planning.

### Verified
- Duplicate-heading guard (skill step 6): exactly one `## v1.1.1-1`.
- Post-shim canary: this PR now reports `test-summary: SUCCESS [Tests (docs-only shim)]`, where before it had no `test-summary` at all — the before/after that proves the shim.

PR #2183, opened 2026-08-11
docs(roadmap): strike the 89 shipped rows in completed buckets (#2181)

### LINKED ISSUE #2181: ROADMAP: 113 shipped rows in completed v1.1.0-N buckets are not struck through

johnrtipton:
Closes #2181. Docs-only, plus the script that produced it.

## Why it matters

134 `**Pn**` rows carried no `~~`/✅, so they read as open work — and `/pipeline-next` reads exactly these rows to choose the next task.

## Result

| | |
|---|---|
| Rows struck | **89** |
| ...citing a real closing PR | 80 |
| ...bare `Shipped` (GitHub reports no closing PR) | 9 |
| Left — still open | 2 (#2017, #1561) |
| Left — aspirational, no issue ref | 12 |
| Left — three-cell rows (different table shape) | 51 |

Closing PR numbers come from GitHub's `closedByPullRequestsReferences`, never guessed — #1197 records hallucinated PR refs as a real failure mode in this repo. Spot-checked 6 of the 80: all exist, all `MERGED`, titles match their issues.

## Conservative by construction

- A row is struck only if **every** issue ref in its **title** cell is CLOSED, so a partly-open row can never be marked shipped.
- Refs that are PR numbers are ignored for the decision — they say nothing about whether the work landed.
- #2017 and #1561 are untouched.

## A bug in the first pass, worth recording

The first version corrupted `ROADMAP.md:3926`. That row's description contains an **escaped pipe**:

```
| **P2** | `{% firstof %}`… (#1692) | `get_value_safe` … (e.g. `x\|safe`) … | Bugfix; Rust template engine. |
```

A naive `split('|')` counts `\|` as a column boundary, so a 4-cell row looks like 5 — and the target-column write landed **in the middle of the description**, destroying half of it:

```
- … the name-based safe-filter whitelist (e.g. `x\|safe`) the Variable arm uses. Closes the #1672/#1660 lineage. | Bugfix; …
+ … the name-based safe-filter whitelist (e.g. `x\| Shipped (PR #1709) | Bugfix; …
```

Silent — the table still rendered. Fixes:

1. Escape-aware splitting: `(?<!\\)\|`.
2. Per-row post-conditions asserting the description and priority cells are **byte-identical** after the rewrite, so a future boundary mis-parse fails loudly instead of eating prose.

Verified globally after the fix: **0 descriptions mutated**, line count unchanged, and L3926's `x\|safe` intact.

## Script ships with it

`scripts/strike-shipped-roadmap-rows.py` — so the next audit re-runs it rather than re-deriving it. Dry-run by default; `--apply` writes.

### DIFF
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -45,14 +45,14 @@ here so the decision is visible either way.
 | **P1** | ~~`FormMixin.validate_field` expects `field_name`, client sends `field` (#2137)~~ ✅ | The framework's own signature diverges from its own wire contract, and `**kwargs` swallows the mismatch instead of raising — a user overriding the documented signature gets a handler that runs on every keystroke and never receives the field. No error, no warning; the form looks live and is inert. Reported against 1.1.0rc8. | Shipped (PR #2141) |
 | **P1** | ~~`[dj-virtual]` content patches for surviving rows are index-addressed (#2136)~~ ✅ | `reconcile_virtual_keyed` recurses into survivors with a child path built from the item's ABSOLUTE index, meaningless for a windowed container; the patch carries no dj-id so it resolves positionally. Measured: editing a row after a scroll silently rewrites a DIFFERENT row, `applyPatches` returns true, no warning. Blocks ADR-026 iteration 3. | Shipped (PR #2146) |
 | **P2** | ~~One source of truth for the client-size figures (#2138)~~ ✅ | Three copies describing two artifacts; only the README pair is checked, and the unchecked CLAUDE.md pair had drifted >2×. Prefer generating the figures at build time over adding a second checker. | Shipped (PR #2147) |
-| **P2** | Detect a red `main` proactively (#2139) | Three failing tests on `main` made the pre-push hook reject every branch, including ones unrelated to the failure; took three failed pushes to diagnose. Wants a scheduled main-suite check AND a hook that distinguishes pre-existing merge-base failures from branch-caused ones. | v1.1.0 |
+| **P2** | ~~Detect a red `main` proactively (#2139)~~ ✅ | Three failing tests on `main` made the pre-push hook reject every branch, including ones unrelated to the failure; took three failed pushes to diagnose. Wants a scheduled main-suite check AND a hook that distinguishes pre-existing merge-base failures from branch-caused ones. | Shipped (PR #2149) |
 | **P2** | ~~Backfill milestone retros for v1.1.0-12 and #2094–#2128 (#2140)~~ ✅ | `RETRO.md` jumped v1.1.0-11 → v1.1.0-13; ~32 PRs have no milestone entry and #2094–#2128 were never bucketed. Per-PR retros exist as comments, so this is synthesis, not archaeology — what was lost is the Action Tracker rows. | Shipped (PR #2141) |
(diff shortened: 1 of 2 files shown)


johnrtipton:
## Retrospective — PR #2183 (pipeline-drain)

**Task**: v1.1.1-1 — drained #2181 (strike shipped ROADMAP rows).

### Quality: 4

### What went well
- **Closing PR numbers came from GitHub, never inference.** `closedByPullRequestsReferences` supplied 80 of them; the 9 issues GitHub reports no closing PR for got a bare `Shipped` rather than a plausible guess. #1197 records hallucinated PR refs as a real failure mode in this repo, so this was the load-bearing constraint.
- **Conservative predicate.** A row is struck only if *every* issue ref in its title cell is CLOSED, so a partly-open row can never be marked shipped. #2017 and #1561 came through untouched, as did 12 aspirational rows with no issue and 51 rows of a different table shape.
- **The script ships** (`scripts/strike-shipped-roadmap-rows.py`), so the next audit re-runs it instead of re-deriving it.

### What didn't
- **The first pass silently destroyed prose.** `ROADMAP.md:3926`'s description contains an escaped pipe (`x\|safe`); a naive `split('|')` counted it as a column boundary, so a 4-cell row parsed as 5 and the target-column write landed *in the middle of the description*, deleting half a sentence. The table still rendered — nothing looked wrong. I caught it only because I ran a structural cell-count check afterwards and then traced the single overlapping line.
- **My first structural check was nearly misread.** It flagged 52 "malformed" rows, which looked alarming until I ran the same check against `HEAD` and found all 52 pre-existing. Only one line actually overlapped my edits. Comparing against the baseline rather than reacting to the absolute number is what kept this from turning into a wild-goose chase.
- The fix was cheap in hindsight — escape-aware `(?<!\\)\|` plus per-row post-conditions asserting the description and priority cells are byte-identical — and should have been the *first* shape, since "rewrite two cells of a markdown table" obviously implies "prove you touched only those two cells."

### Verified
- Global guard after the fix: **0 descriptions mutated** across the whole file, line count unchanged, `x\|safe` intact.
- Spot-checked 6 of the 80 cited PRs — all exist, all `MERGED`, titles match their issues.
- Post-merge re-verification after resolving the #2182 merge: 115 struck rows, 2 open-issue rows preserved, escaped-pipe row still whole.
- CI: 15 SUCCESS.

### Carried forward
Nothing. #2181 is closed by this PR.

Issue #1524, opened 2026-05-18
tech-debt: pipeline-run skill — `git commit --amend` must assert HEAD hash changed


johnrtipton:
**Source**: v1.0.0rc3 milestone retro, finding #2 (PR #1519).

Action #122 prescribes a `&& git log -1 --oneline` reflex after every `git commit` to detect a pre-commit-hook-swallowed commit. That reflex does **not** catch a bounced `git commit --amend`: after a bounced amend the OLD commit is still HEAD with its OLD subject, so a subject *is* shown and the reflex passes.

Observed in PR #1519: a fixer's `--amend` was swallowed by a pre-commit reformat; HEAD stayed at `5ca016d0`, `git status` showed `MM`, and the fixer agent reported "amended commit 5ca016d0" — quoting the stale hash. The orchestrator caught it on first principles (amend always rehashes; `git show HEAD:` grep returned 0).

**In-repo half is done**: CLAUDE.md gained a "Process canonicalizations from v1.0.0rc3 retro arc" section with the `--amend` HEAD-hash-changed rule.

**This issue (out-of-repo)**: propagate the rule into `~/.claude/skills/pipeline-run/SKILL.md` — the MANDATORY Post-Commit Verification section should add: for `git commit --amend`, capture `PRE=$(git rev-parse HEAD)` before and assert `git rev-parse HEAD` differs after. `.claude/` is gitignored repo-wide; upstream is the pipeline-skill repo.

johnrtipton:
Resolved upstream in pipeline-skills@f3203b2 (canon commit e25a213, merged to main). `pipeline-run/SKILL.md` gained a "`git commit --amend` companion" subsection under MANDATORY Post-Commit Verification — capture the pre-amend hash and assert HEAD changed, since a bounced amend leaves the old subject visible and the `&& git log -1` reflex passes green on failure.

PR #2184, opened 2026-08-11
ci: report test-summary on docs-only PRs so it can be a required check (#2163)

johnrtipton:
Refs #2163. **Prerequisite only** — branch protection is deliberately not changed here.

## The gap

```
$ gh api repos/djust-org/djust/branches/main/protection -q .required_status_checks
null
$ gh api repos/djust-org/djust/rules/branches/main -q length
0
```

The entire `test-summary` aggregate — rust, python, javascript, browser-smoke, nav-hooks-guard, security-tests, demo-checks, benchmarks — can be red and the merge button stays live. #1713 one level up.

## Why the obvious fix breaks the repo

Turning the requirement on today makes docs-only PRs unmergeable. `test.yml` filters with `paths-ignore`, and **a skipped WORKFLOW never creates its check runs at all** — unlike a skipped JOB, which reports and which branch protection treats as satisfied. A required `test-summary` would hang at *"Expected — Waiting for status to be reported"*.

Measured on this repo, not assumed — using the two PRs from this very drain:

| PR | Files | Checks | `test-summary`? |
|---|---|---|---|
| #2182 | `ROADMAP.md` | **1** | **absent** |
| #2183 | `ROADMAP.md` + `.py` | 14 | present |

The two-commit convention (impl+tests / docs+CHANGELOG) makes docs-only PRs routine.

## Why the shim is not a bare "always succeed" job

`paths` here and `paths-ignore` in `test.yml` are **not mutually exclusive** — a PR touching a `.md` *and* a `.py` triggers both workflows. An unconditional shim would put a second, always-green `test-summary` beside the real one and **mask a genuine failure** — the exact bug class #2163 is about.

So a `detect` job classifies the diff and the reporting job is gated on it:

| PR shape | `test.yml` | shim |
|---|---|---|
| docs-only | skipped | reports `test-summary` = success |
| mixed | reports (gates) | job **SKIPPED** — cannot green-wash |
| code-only | reports (gates) | not triggered |

An empty diff is explicitly *not* docs-only — there is nothing to vouch for.

## Anti-drift

The shim's path list and `test.yml`'s `paths-ignore` are one contract in two files (#1646). Seven guards in `test_ci_docs_shim_2163.py` pin it — including the bash `case` arms, which decide whether the shim *vouches*, separately from the trigger that decides whether it *runs*.

**Gate-off — all five drift mutations go red.** The harness asserts the mutation applied and reports `INVALID` on a collection error rather than a count (#2135):

```
BASELINE                             -> 0 failed, 7 passed
test.yml adds *.rst (both blocks)    -> 2 failed
test.yml push/PR drift (PR only)     -> 1 failed
shim reporting job ungated           -> 1 failed
shim job renamed                     -> 2 failed
case arm drops *.md                  -> 1 failed
```

### Two things writing those guards turned up

1. Shell `case` matches `/` with a plain `*`, so `**.md` translates to `*.md`. My first parity test asserted the wrong mapping and failed — correctly.
2. **`test.yml` declares `paths-ignore` twice** (`push` at `:6`, `pull_request` at `:12`). A gate-off mutation with `count=1` silently landed on `push`, which no guard read — an invalid measurement that looked like a passing test. That is its own #1646 drift, now pinned by `test_test_yml_push_and_pull_request_filters_agree`.

## Not done here

Branch protection stays untouched until this shim has reported green on a real runner (#1534). Order: land this → confirm the context name on a green run → then add `test-summary` to required checks.

Full suite: **10238 passed, 218 skipped, 0 failed**.

### DIFF
--- a/.github/workflows/test-summary-docs-shim.yml
+++ b/.github/workflows/test-summary-docs-shim.yml
@@ -1,0 +1,107 @@
+name: Tests (docs-only shim)
+
+# Makes `test-summary` reportable on a docs-only PR so it can safely be a
+# REQUIRED status check (#2163).
+#
+# The problem this exists for: `test.yml` filters with `paths-ignore` for
+# `**.md`, `docs/**` and `CHANGELOG.md`. When every changed file matches that
+# list the whole workflow is skipped, and a skipped WORKFLOW never creates its
+# check runs at all — unlike a skipped JOB, which does report and which branch
+# protection treats as satisfied. A required `test-summary` would therefore sit
+# at "Expected — Waiting for status to be reported" forever on a docs-only PR.
+#
+# Measured on this repo, not assumed: PR #2182 (ROADMAP.md only) produced
+# exactly ONE check and no `test-summary`; PR #2183 (ROADMAP.md + a .py file)
+# produced 14 including `test-summary`. This repo's two-commit convention
+# (impl+tests / docs+CHANGELOG) makes docs-only PRs a routine shape.
+#
+# Why the two-job shape rather than a bare "always succeed" job: `paths` here
+# and `paths-ignore` in test.yml are NOT mutually exclusive. A PR touching both
+# a .md and a .py triggers BOTH workflows, and a shim that unconditionally
+# reported success would put a second, always-green `test-summary` alongside
+# the real one — masking a genuine failure, which is precisely the class of bug
+# #2163 is about. So the reporting job is gated on a real docs-only computation:
+#
+#   docs-only PR  -> test.yml skipped;  shim reports test-summary = success
+#   mixed PR      -> test.yml reports;  shim job SKIPPED (never green-washes)
+#   code-only PR  -> test.yml reports;  shim not triggered at all
+#
+# The `detect` job's own check run is incidental and must NOT be made required.
+
+on:
+  pull_request:
+    branches: [ main ]
+    paths:
+      - '**.md'
+      - 'docs/**'
+      - 'CHANGELOG.md'
+
+permissions:
+  contents: read
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  detect:
+    name: Detect docs-only change
+    runs-on: ubuntu-latest
+    outputs:
+      docs_only: ${{ steps.check.outputs.docs_only }}
+    steps:
+      - uses: actions/checkout@v7
+        with:
+          fetch-depth: 0
+
+      - name: Classify the changed files
+        id: check
+        env:
+          BASE_SHA: ${{ github.event.pull_request.base.sha }}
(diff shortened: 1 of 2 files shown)


johnrtipton:
## Retrospective — PR #2184 (pipeline-drain)

**Task**: v1.1.1-1 — drained #2163 (prerequisite half: make `test-summary` reportable on docs-only PRs).

### Quality: 5

### What went well
- **The trap was measured, not argued.** The two docs PRs from this same drain were the evidence: #2182 (`ROADMAP.md` only) produced 1 check and no `test-summary`; #2183 (`.md` + `.py`) produced 14 including it. That turned "a required check might hang" into a demonstrated fact before any code was written.
- **The green-wash risk was designed out rather than noted.** `paths` and `paths-ignore` are not mutually exclusive, so an unconditional shim would have sat a second always-green `test-summary` beside the real one — the exact bug class #2163 is about. The `detect`-gated shape prevents it structurally.
- **Post-merge empirical canary confirmed all three branches** of the design table on real runners (see Verified).

### What didn't
- **The first commit was swallowed and I nearly missed it.** `pre-commit`'s ruff hook reformatted the test file; the commit aborted leaving files staged (`AM`). I only caught it because `gh pr create` failed with *"No commits between main and …"*. Canon (#1524 and the post-commit reflex) exists for exactly this and I did not run it. The re-commit used an explicit `PRE`/`POST` `git rev-parse` guard.
- **My first gate-off measurement was invalid and looked fine.** The mutation used `count=1` against `test.yml`, which has `paths-ignore` **twice** (`push` at `:6`, `pull_request` at `:12`). It landed on `push`, which no guard reads, so the drift case reported "0 failed" — reading as *"the test doesn't catch this"* when the truth was *"the mutation didn't test it."* Fixing it surfaced a genuine second #1646 drift (push vs pull_request), now pinned.
- **My own parity test was wrong on first write** — shell `case` matches `/` with a plain `*`, so `**.md` maps to `*.md`. It failed correctly and I fixed the test, not the shim.

### Verified
- Gate-off, 5/5 drift mutations red; harness asserts the mutation applied and reports `INVALID` on a collection error (#2135).
- Full Python suite: **10238 passed, 218 skipped, 0 failed**.
- CI on the PR: 14 SUCCESS.
- **Post-merge canary on real runners** — all three branches:
  - docs-only (#2182): `test-summary: SUCCESS [Tests (docs-only shim)]` (was: absent)
  - mixed (#2183): `test-summary: SKIPPED [shim]` + `test-summary: SUCCESS [Tests]` — no green-wash
  - code-only (#2184 itself): shim not triggered

### Carried forward
Branch protection is still **not** enabled. One residual unknown is called out on #2163: on a mixed PR two check runs share the name `test-summary`, and GitHub resolves duplicates by most-recently-completed. That ordering favours the real suite in practice (it runs ~10 min vs the shim's ~40 s) but is timing-dependent, not structural. Worth settling before the requirement goes on.

Issue #2185, opened 2026-08-11
[dj-virtual] initialisation is intermittently lost on page load — the #1610 mount morph tears down the shell and nothing re-inits


johnrtipton:
Found during the #2017 ADR-026 iteration-3 browser gate. **Independent of `virtual_keyed_ops`** — reproduced with the flag both ON and OFF — and it is the blocker for that flip.

## Symptom

On *some* page loads a `[dj-virtual]` container is never virtualised: every row stays in the DOM, no shell/spacer, item pool empty. It does **not** self-heal.

Measured on `/tests/virtual-keyed/` (60 rows), sampling once a second for 14 s after load:

```
t=3s .. t=14s   pool: 0    rows: 60    shell: false      (flat, every sample)
```

Then, in the same page, with no other change:

```js
window.djust.initVirtualLists(document)
// -> pool: 60, rows: 13, shell: true      (fixed instantly)
```

So the machinery is fine; it simply never runs — or runs and is destroyed.

Other loads of the same URL come up correctly virtualised (`pool: 60, rows: 13, shell: true`), which is what makes this intermittent rather than a flat break.

## Mechanism

Console ordering on a **failing** load:

```
[LiveView] Initializing...
[LiveView] Initialization complete, window.djustInitialized = true   <- runs initVirtualLists (14-init.js:143)
[LiveView] Morphed pre-rendered DOM against WS-mount HTML (#1610)    <- re-creates the dj-root subtree
```

`initVirtualLists(document)` runs inside `djustInit()`, and the #1610 WS-mount morph runs **after** it, re-creating the `dj-root` subtree — which discards the shell/spacer that `setup()` installed. Nothing calls `initVirtualLists` again afterwards. The only re-entry points today are `14-init.js:143` (init) and `09-event-binding.js:1614` (post-patch reinit scope); neither fires on the mount morph.

That also explains the intermittency: when init happens to land after the morph, virtualization survives.

This is the same family as the mount-morph teardown that drops inline `<script>` execution inside `dj-root`.

## Impact

`[dj-virtual]` exists to keep large lists out of the DOM. On an affected load that benefit is silently absent — a 10 000-row list renders 10 000 rows. There is no error; the page just gets slow.

## Why it blocks #2017 iteration 3

With `virtual_keyed_ops` **OFF** (today's default) an affected load degrades *silently*: no virtualization, but ordinary VDOM patches still apply.

With the flag **ON**, the same load emits `VirtualInsert` against an uninitialised container:

```
[...]

PR #2186, opened 2026-08-11
fix(rust): unblock commits on clippy 1.97 + correct #2164's withdrawn diagnosis (#2017, #2185)

johnrtipton:
Ran `/pipeline-run` on **#2017 ADR-026 iteration 3** (flip `virtual_keyed_ops` to default ON).

## Verdict: the gate FAILS — the flag stays OFF

`ROADMAP.md` gates iteration 3 on *"real browser verification producing evidence worth flipping on."* I gathered that evidence. **The flip is not taken**, and this PR contains no behaviour change.

### The keyed ops are correct

Flag ON, list confirmed initialised **before** acting — every case clean, `Patches applied successfully`, no recovery round-trip:

| Case | Result |
|---|---|
| insert at server position 5 | lands at pool index **5** (`k3, k4, ins0, k5…`) |
| remove at 3 | k3 gone, size −1, **no duplicates** |
| reverse | **exact** reversal, no lost keys |
| edit row 0 | `EDITED` on **k0 only** — no cross-row contamination (#2136 class) |

### What actually blocks the flip — filed as #2185

`[dj-virtual]` initialisation is **intermittently lost on page load**, *independent of the flag* (reproduced with it ON and OFF). On an affected load, sampled once a second for 14 s:

```
t=3s .. t=14s   pool: 0   rows: 60   shell: false      (flat — never self-heals)
```

yet `djust.initVirtualLists(document)` fixes it instantly (`pool: 60, rows: 13, shell: true`).

Console ordering on a failing load shows why:

```
[LiveView] Initialization complete            <- runs initVirtualLists (14-init.js:143)
[LiveView] Morphed pre-rendered DOM against WS-mount HTML (#1610)   <- re-creates dj-root
```

The mount morph discards the shell/spacer and nothing re-runs init.

**Why that blocks iteration 3**: on such a load, flag OFF degrades *silently*; flag ON emits `VirtualInsert` at an uninitialised container → `1/3 patches failed` → **full HTML recovery round-trip**. Flipping would convert a silent degradation into a visible failure for every `[dj-virtual]` user on affected loads.

### 1. Unblocks commits on Rust 1.97 (the reason it is a `fix:`)

`cargo clippy --all-targets --all-features -- -D warnings` — the exact pre-commit hook invocation — failed on two **pre-existing** sites:

- `djust_components/src/simple/tooltip.rs:62` — `useless_borrows_in_formatting`
- `djust_live/src/lib.rs:1324` — `question_mark`

Surfaced by upgrading the local toolchain (Homebrew rust 1.93 was broken: `librustc_driver` linked libLLVM 21.1.8 while `/opt/homebrew/opt/llvm` had moved to 22.1.8, so `rustc -vV` aborted). Not local-only — **no commit can land on 1.97** (it silently swallowed three of mine), and CI pins `toolchain: stable` (`test.yml:35`, `:109`), so this becomes a CI failure the moment stable advances.

⚠️ **Correction (was wrong in the first version of this PR).** I originally wrote that the `djust_live` violation is "feature-gated and only appears with `--all-features`". That is false, and the Stage 11 reviewer caught it. Re-tested by reverting only the `lib.rs` fix and re-running *without* `--all-features`: the error is reported. It cannot be feature-gated — `djust_live`'s only feature is `extension-module`, already in `default`.

What actually happened: my original run **aborted** at the `djust_components` error, and `djust_components` is a path dependency of `djust_live`, so `djust_live` was never checked. I read "no error in the output" as "clean" — a false negative from an aborted build. Retracted in `9b51c99`.

### 2. Corrects two artifacts asserting a withdrawn diagnosis

`diff.rs:167` and the demo banner both still said *"the client applier does not honour keyed positioning — see #2164"*. That was #2164's **second** diagnosis, withdrawn by PR #2167. The demo was the worse offender — it opened with *"This page documents a known failure, not a working feature"*, so the artifact built to gate this decision was telling its reader the feature is broken when it is not. Both now record the real reason (#2185).

## A caveat on my own measurements

Several early readings were **invalid**: I fired events inside the initialisation window and read a race as an op failure, which briefly led me to the wrong conclusion. Only runs asserting `pool > 0` *before* acting are load-bearing, and the table above uses only those. Worth stating plainly given this issue already has two withdrawn diagnoses on record.

## Verification

- `cargo clippy --all-targets --all-features -- -D warnings`: clean
- `cargo test -p djust_components`: 104 passed
- `cargo test -p djust_live --no-default-features`: 37 passed
- `cargo test -p djust_vdom`: green
- Demo template loads


johnrtipton (commented):
## Stage 11 Code Review — PR #2186

Reviewed at `348ca426` in a worktree-isolated checkout (read-only against the main checkout; no `git config` mutation, per #300). Staleness gate first: `git rev-list --count origin/feat-2017-iter3-flip-virtual-keyed-default..origin/main` = **0**, so the diff I reviewed is the program that will merge (#1450 satisfied).

**Verifying not flipping the flag is the right call, and I agree with the verdict.** The evidence for it is good and the self-caveat about invalid early readings is the right instinct on an issue with two withdrawn diagnoses already. My findings are about what the PR *left behind*, plus one claim in the commit message that is false.

---

## What I verified as correct

**The `?` rewrite (`crates/djust_live/src/lib.rs:1324`) is exactly equivalent.** Checked all four axes:

- `try_parse_cache_splice` is declared `-> Option<VNode>` (`lib.rs:1296`), so `?` desugars to `return None`, not a `Result` residual. Same as the deleted `None => return None`.
- `get_parsed` (`crates/djust_templates/src/loop_cache.rs:286`) is `pub fn get_parsed(&mut self, hash: u64) -> Option<Vec<VNode>>` — it returns an **owned** clone, so no borrow of `loop_cache` survives the statement in either form. The match form held the scrutinee temporary for the whole `match`; `?` releases at the semicolon. Strictly *less* restrictive, and nothing in the arm touched `loop_cache` anyway.
- Call count is unchanged: `or_insert` (not `or_insert_with`) evaluates `roots` eagerly in both forms, so `get_parsed` runs **exactly once per hash** and `parse_hits` bumps identically — including the duplicate-hash case, where both forms call it per occurrence and `or_insert` keeps the first. On a miss both bail before querying the remaining hashes.
- Strongest evidence: clippy's own machine-applicable suggestion for this site is character-for-character the rewrite the PR took (`let roots = loop_cache.get_parsed(h)?;`). I captured it by reverting the file — see the canary below.

**The `&self.content` removal (`crates/djust_components/src/simple/tooltip.rs:62`) is byte-identical output.** `content: String` (`tooltip.rs:11`); `format_args!` captures by reference regardless, and `Display for &T` is the blanket impl delegating to `String`'s. Behaviourally covered, not just inspected: `test_tooltip_basic` asserts `html.contains("Hover me")` — the content itself — and passes.

**No lint was suppressed.** The only `#![allow(clippy::…)]` in `djust_live/src/lib.rs` are the two pre-existing ones at lines 7 and 9, untouched. Fixing over `#[allow]` is the right call.

**Build/test, run locally on rustc 1.97.1:**

| command | result |
|---|---|
| `cargo clippy --all-targets --all-features -- -D warnings` (the hook's exact args) | clean |
| `cargo test -p djust_components` | 104 passed, 0 failed |
| `cargo test -p djust_live --no-default-features` | 37 passed, 0 failed |
| `cargo test -p djust_vdom` | green |

CI is green on the PR (13 checks pass, 1 pending).

**Citations spot-checked** (#1197): `toolchain: stable` is at `.github/workflows/test.yml:35` and `:109` — correct (there are also two more at `:225` and `:461`, so the claim under-counts, which is harmless). `14-init.js:143` is exactly `if (window.djust.initVirtualLists) window.djust.initVirtualLists(document);` — correct, and it does sit above the `Initialization complete` log at `:146`, which supports the console-ordering argument. `ROWS = 60` in `virtual_keyed_demo.py:27` matches the banner's "all 60 rows". `diff.rs` and the demo banner no longer assert the withdrawn claim — correct. #2185 exists and is open.

---

## 🔴 1. The withdrawn-diagnosis sweep is 2 of ≥5 — and one of the misses contradicts the artifact this PR just fixed

`grep -rn "2164" --exclude-dir=target --exclude-dir=.git` takes under a second and finds four more live assertions of a diagnosis this PR exists to retract:

**a. `examples/demo_project/djust_tests/views/test_index.py:33-41`** — the index card that *links to the page whose banner this PR rewrote*:

```python
"description": "Windowed list exercising ADR-026. Documents a KNOWN FAILURE: an insert at position 5 lands at the tail.",
"status": "failing",
"tests": [..., "Client applier does not honour it yet (#2164)", ...],
```

`"failing"` is not inert text — `templates/tests/index.html:283` renders `class="test-item status-{{ test.status }}"` and `:299` a `status-badge failing`, so the demo index shows a **red FAILING badge** plus "KNOWN FAILURE: an insert at position 5 lands at the tail". Click through and the banner now says "The keyed ops work". The PR's own justification for the banner fix — *"the one artifact built to gate this decision was telling its reader the feature is broken when it is not"* — applies with more force to the entry point a reader hits **first**. As it stands the PR has made the demo app internally self-contradictory where it was previously merely wrong.

**b. `docs/adr/026-dj-virtual-differ-awareness.md:70`** — `| 3. flag flips ON after a soak | **not shipped — blocked on #2164** |`. The blocker is #2185.

**c. `docs/adr/026-dj-virtual-differ-awareness.md:87-96`** — the sharpest miss, because this is the *governing ADR* for the feature. It still carries the measurement table asserting flag ON → pool index 60 (tail), and:

> The differ is NOT at fault. … **Iteration 2's client applier is what does not honour it**, appending at the tail instead of splicing before `k5`. Tracked as #2164.

That is the exact sentence the PR retracts everywhere else. And it is worth noting *where* it sits: lines 58-62 of this same ADR already record that a previous wrong prose invariant here was an "#1867 class — a prose invariant nobody had run against the code, and `make check-adr-status` did not catch it: it audits status/version-line *consistency*, not whether a status matches reality." The ADR is warning about precisely this, one screen above the paragraph being left wrong. Nothing automated will catch it — I confirmed `make check-adr-status` only validates Status/version-line consistency (`Makefile:180-182`).

**d. `python/djust/config.py:175-178`** — still records #2164's **first** withdrawn diagnosis, one round further out of date than the others:

```python
# Default OFF: ADR-026 iteration 3 (the flip) is NOT shipped. The
# browser gate found the differ emits the right op but a real page
# is not windowed at patch time, so there is no pool to apply it
# to — see #2164. Turning this on today changes nothing.
```

The ADR itself already says that cause "was wrong", and "**Turning this on today changes nothing**" is now falsified by this PR's own browser table (insert at 5 → pool index 5). This is the docstring block a user reads immediately adjacent to the setting they would flip, so of the four it is the one most likely to be read by someone acting on it.

This is the symbol-migration grep canon (#1391/#1400) and the enumerate-every-variant rule (v1.0.0rc4 finding #1) — a correction applied to a subset of the sites carrying a claim leaves the class alive while making it *look* handled. Please either sweep all sites in this PR or, if you want to keep the diff tight per #1079, sweep (a) and (d) here (they are user-facing and actively contradictory) and file a follow-up for the ADR.

## 🔴 2. No CHANGELOG entry for a `fix(rust):` PR

`CLAUDE.md` Testing Expectations: *"`feat:` and `fix:` PRs must update CHANGELOG.md"*. The lead commit and the PR title are both `fix(rust): …`. The only carve-out this repo has canonized is the minimal-form Stage-14 retro for lockfile-only Dependabot bumps (#1549), which this is not.

Note CI cannot catch this: `.github/workflows/test.yml:9,15` puts `CHANGELOG.md` in `paths-ignore`, so it is a human gate only — which is exactly why it gets missed.

Two acceptable resolutions, both cheap:
- Add a one-liner under `### Fixed` — and while you are there, the existing `virtual_keyed_ops` entry at `CHANGELOG.md:21` still ends on "*Filed as #2164; #2017 stays open*" with no mention that the current blocker is #2185. That is the natural place to record it.
- Or retype to `chore(rust):`/`style(rust):`, which is defensible for a behaviour-preserving lint cleanup and drops the requirement.

I would not object to either. I do object to `fix:` with no entry.

## 🟡 3. The gate ran and the verdict is not recorded where the gate lives

`ROADMAP.md:68-72` states the contract:

> Iteration 3 is taken only if #2136's fix plus real browser verification produces evidence worth flipping on — **recorded here so the decision is visible either way**.

The verification ran; the decision was no; `ROADMAP.md` is unchanged, and so is the ADR row (finding 1b). The durable record of a milestone-gating decision is currently a Rust doc comment, a demo banner, and this PR body. "Either way" is doing real work in that sentence — the negative outcome is the one that needs recording, because the next person to pick up #2017 reads the ROADMAP and the ADR, not `diff.rs`.

## 🟡 4. The "feature-gated" explanation is false — and it is stated as reproduction advice

From the commit message and repeated in the PR body:

> `cargo clippy --workspace --all-targets` reports CLEAN. The djust_live violation is behind a feature flag and only appears with `--all-features`, which is what the hook passes.

Empirical canary (#1459), run here on rustc 1.97.1:

1. On `main` unmodified, `cargo clippy --workspace --all-targets -- -D warnings` fails at `tooltip.rs:62` and stops: `error: could not compile djust_components (lib)` / `build failed, waiting for other jobs to finish`. `djust_components` is a **path dependency of `djust_live`** (`crates/djust_live/Cargo.toml`), so `djust_live` is never checked. *That* is why the `question_mark` violation did not appear — the run aborted upstream of it, not because of a feature.
2. On the PR head with **only** `crates/djust_live/src/lib.rs` reverted to main's version (tooltip fix kept), `cargo clippy --workspace --all-targets -- -D warnings` — **no `--all-features`** — reports `error: … question_mark` at `lib.rs:1324`, with the machine suggestion `let roots = loop_cache.get_parsed(h)?;`.
3. Structurally it cannot be feature-gated anyway: `djust_live` is the **only** workspace member with a `[features]` section, and it reads `default = ["extension-module"]` / `extension-module = ["pyo3/extension-module"]` — so `--all-features` enables exactly the default set and is a no-op for this workspace. `grep -n "cfg(feature" crates/djust_live/src/lib.rs` returns nothing.

The *advice* ("use the hook's exact args, not an approximation") is right and worth keeping. The *reason* is wrong, and it is addressed to "anyone reproducing" — so a maintainer who trusts it will go hunting for a feature gate that does not exist, and will draw the wrong conclusion about which clippy invocations are safe to rely on. Given this PR's whole second commit is about not leaving wrong diagnoses on the record, I would like this one corrected too rather than shipped in the permanent commit message. The accurate version is roughly: *"an earlier violation in `djust_components` aborts the run before `djust_live` is checked, so fix them in order and re-run."*

## 🟢 5. Pre-existing, out of scope, worth a follow-up

`tooltip.rs:62` interpolates `self.content` into HTML **unescaped**, while the sibling `text` goes through `html_escape_attr` (`tooltip.rs:57`). That is almost certainly intentional (content is caller-supplied markup for the wrapped element) but nothing in the file says so. Not this PR's problem and I am not asking for it here — but a reviewer landing on this exact line under this repo's hard security rules will ask, so a one-line comment stating the intent, or an issue, would save the next person the trace.

## 🟢 6. A dated browser measurement inside a Rust doc comment

`diff.rs:173-180` now carries "Re-measured in a browser (2026-08-11) … insert at server position 5 lands at pool index 5 …". That belongs in the ADR (which is unfortunately the one place not updated — finding 1c). A doc comment reading `///` on a static will rot silently; the ADR has `make check-adr-status` and a review convention around it. Consider trimming the comment to the decision plus a pointer and putting the measurement table in ADR-026.

---

## Genuine concern — is "flag OFF degrades silently" measured, or inferred?

This is load-bearing for the verdict, so I want to push on it. The argument is:

> on such a load, flag OFF degrades *silently*; flag ON emits `VirtualInsert` at an uninitialised container → `1/3 patches failed` → full HTML recovery round-trip.

On a #2185-affected load the container is uninitialised: pool 0, all 60 rows flat. With the flag OFF the differ emits its normal **index-addressed** content ops into that container. #2136 — closed by PR #2146, and cited in `ROADMAP.md:77` as the other iteration-3 blocker — was exactly: *"editing a row after a scroll silently rewrites a DIFFERENT row, `applyPatches` returns true, no warning."*

So: was flag-OFF-on-an-affected-load measured to be **correct**, or only measured to produce **no visible failure**? Those are not the same, and the difference decides the verdict. A visible recovery round-trip (ON) is unpleasant but self-correcting and observable; a silent wrong-row write (OFF) is neither. If OFF is silently wrong on affected loads, the risk ordering in the argument inverts and "OFF is the safer default pending #2185" no longer follows from this evidence.

Your own caveat section is what makes me ask: you state that only runs asserting `pool > 0` before acting are load-bearing — but the "OFF degrades silently" claim is by construction a `pool == 0` run, i.e. it comes from the class of runs the caveat excludes. I am not claiming the conclusion is wrong. I am saying the one claim the verdict rests on is the one not covered by the PR's own evidence standard, and it would take one more browser run to settle: on an affected load, flag OFF, fire `edit_first`, and check *which row* changed.

## Edge cases I considered

- **`?` on a `Result`-returning fn** — would have changed the returned error type. Ruled out: return type is `Option<VNode>`.
- **Duplicate hashes in `placeholder_hashes`** — would change `parse_hits` accounting if `?` moved the call. It does not; both forms call per occurrence and `or_insert` keeps the first.
- **`or_insert` vs `or_insert_with` laziness** — if it were `or_insert_with`, the eager/lazy distinction would matter. It is `or_insert`; both forms are eager.
- **Borrow-region widening** — the `match` scrutinee temporary outlives the arm; `?` does not. Checked that nothing in the arm touches `loop_cache`. No behavioural difference, and the compiler accepts both.
- **Early-return ordering on a miss** — both forms bail before querying later hashes, so cache-hit counters agree on the failure path too.
- **`Display for &String` vs `String`** — blanket impl, identical bytes; covered by an assertion on the content string, not just by inspection.
- **MSRV** — `?` on `Option` is stable since 1.22; `rust-version` is not pinned below that anywhere in the workspace.
- **CI drift** — checked all four `toolchain: stable` sites, not just the two cited.
- **Was the clippy failure a local-toolchain artifact?** No: reproduced from scratch at rustc 1.97.1 on `main`, and the reasoning about CI's `stable` advancing is sound.
- **Stale base** — 0 commits behind `origin/main`.

## Findings

| | finding |
|---|---|
| 🔴 | Withdrawn-diagnosis sweep is 2 of ≥5; `test_index.py:33-41` now contradicts the banner this PR fixed, and ADR-026:70,87-96 + `config.py:175-178` still assert retracted causes |
| 🔴 | No CHANGELOG entry on a `fix(rust):` PR (or retype the commit) |
| 🟡 | Gate verdict not recorded in `ROADMAP.md` / ADR-026, which is where the gate says it must be visible |
| 🟡 | "the djust_live violation is feature-gated / only appears with `--all-features`" is false — disproven empirically and structurally |
| 🟢 | `self.content` interpolated unescaped (pre-existing) |
| 🟢 | Dated browser measurement lives in a Rust doc comment rather than the ADR |

The two Rust changes themselves are correct, minimal, well-covered and I would merge them as-is. The blockers are the incomplete sweep — which, unaddressed, leaves the demo app asserting both halves of a contradiction — and the missing CHANGELOG.

REQUEST_CHANGES

johnrtipton:
## Review findings addressed — all four, plus one retraction

Thanks — two of these were errors in my own claims, not just gaps.

### 🔴 Withdrawn-diagnosis sweep was 2 of ≥5 — fixed (`2e2f32a`)

You were right, and the miss was the embarrassing kind. The demo **index card** still rendered a red `FAILING` badge captioned *"KNOWN FAILURE: an insert at position 5 lands at the tail"*, linking straight to the page whose banner this PR had just rewritten to say the opposite — the demo app asserting both halves of a contradiction.

| Site | Was |
|---|---|
| `test_index.py:33-41` | `status: "failing"`, "KNOWN FAILURE", "Client applier does not honour it yet (#2164)" |
| `config.py:175-178` | #2164's **first** diagnosis ("not windowed at patch time") + the false "Turning this on today changes nothing" |
| `ADR-026:70, :85-96` | status row "blocked on #2164"; A/B table reporting `60 (tail)` for **both** arms |

The ADR was the important one. Both arms of that table were really the OFF arm — the measurement predates PR #2167, so the flag never reached Rust. Added a `SUPERSEDED` block with the re-measured table, kept the original beneath it for the record, and repointed the status row to #2185.

Swept clean; the only remaining matches are inside the correction text itself, quoting the old claim to mark it withdrawn.

### 🔴 Missing CHANGELOG — fixed (`61b26d9`)

Correct, and your note that CI *structurally* cannot catch it is the part worth keeping: `CHANGELOG.md` is inside `test.yml`'s `paths-ignore`. Two entries added — clippy blocker under **Fixed**, the artifact corrections under **Changed** (no behaviour change).

### 🟡 "Feature-gated" claim was false — retracted (`9b51c99`)

You were right and I was wrong. Reverted only the `lib.rs` fix and re-ran **without** `--all-features`: the `question_mark` error is reported. And as you say it is structurally impossible — `djust_live`'s only feature is `extension-module`, already in `default`.

The real explanation is duller: my original `--workspace` run **aborted** at the `djust_components` error, and `djust_components` is a path dep of `djust_live`, so `djust_live` was never checked. I read "no error in the output" as "clean" — a false negative from an aborted build. Retracted in the branch history and corrected in the PR body.

### 🟡 "Flag OFF degrades silently" rested on excluded runs — now measured

This was the sharpest catch: I was leaning on `pool == 0` runs that my own caveat excludes as non-load-bearing.

Rather than argue it, I measured it, using deterministic teardown instead of waiting for the intermittent state:

```
teardownVirtualList(c)  ->  {pool: 0, rows: 60, shell: false}   // affected-load state, on demand
click #btn-insert       ->  patch_applied: true, new_version: 2
server sent             ->  InsertChild path [5] index 5        // ordinary patch, NOT VirtualInsert
console                 ->  "Applying 3 patches"; 0 errors
```

So with the flag OFF an uninitialised container gets an ordinary `InsertChild` that applies cleanly — no failure, no recovery. Paired with the ON arm (`VirtualInsert` → *"not an initialised virtual list"* → `1/3 patches failed` → recovery), the asymmetry now has both halves measured on load-bearing runs.

Your objection stands on the record though: the claim was under-evidenced when I made it.

---

`cargo clippy --all-targets --all-features -- -D warnings` clean; 10238 passed / 218 skipped; `make check-adr-status` clean (26 ADRs); `[1.1.0rc9]` CHANGELOG section byte-identical to its tag.

Ready for re-review.

johnrtipton (commented):
## Stage 13 Re-Review — PR #2186

Re-reviewed at `9b51c994` (= `headRefOid`) in a worktree-isolated detached checkout. No `git config` mutation against the main checkout (#300). Staleness gate first: the branch base is `555526ac`, which **is** `origin/main`, so `git rev-list --count HEAD..origin/main` = **0** — the diff I reviewed is the program that will merge (#1450).

All four Stage 11 findings are **ADDRESSED**. I verified each against the code rather than the claim, re-ran the retraction's empirical test myself, and swept independently. Four new prose-accuracy items below — none blocking, two of them the same class the PR exists to close.

---

## Stage 11 findings

### 🔴 F1 — withdrawn-diagnosis sweep incomplete — **ADDRESSED**

All three named sites corrected, and the corrections are substantively right, not reworded:

| Site | Now | Accurate? |
|---|---|---|
| `examples/demo_project/djust_tests/views/test_index.py:33-42` | `status: "passing"`; caption "The keyed ops work; the flag stays OFF because of #2185"; the three bullets describe what the ops do | ✅ rendered it: card reads `passing`, counts `4/0/0`, `status-badge passing` branch at `templates/tests/index.html:300` exists |
| `python/djust/config.py:175-190` | quotes the old text explicitly as #2164's **first** diagnosis, marks it withdrawn, names #2185 | ✅ value `"virtual_keyed_ops": False` unchanged; verified at runtime `_defaults['virtual_keyed_ops'] is False` and `_rust.virtual_keyed_ops_enabled()` is `False` |
| `docs/adr/026-dj-virtual-differ-awareness.md:70`, `:88-116` | status row repointed to #2185; `SUPERSEDED` block with re-measured table | ✅ see below |

**I checked the ADR's load-bearing new claim rather than taking it.** The block says the original A/B "was taken while the flag never actually reached Rust … so BOTH arms above are really the OFF arm." That holds:

- `git log -S"60 (tail)" -- docs/adr/026-...md` → the table landed in **`96b02df4` (#2165)**.
- `git show 96b02df4:python/djust/apps.py:54-58` reads the flag via `_cfg.get("virtual_keyed_ops", False)` — the **import-time singleton**, i.e. exactly the object #2164/#2166 proved was serving pure defaults on the demo project (whose `asgi.py` imported `djust` above its `setdefault`).
- `7c477964` (#2167, the recovery fix) is **after** `96b02df4` in `main`'s history.

So the measurement genuinely predates the fix and both arms were the OFF arm. The claim is correct.

**My own independent sweep** — `grep -rniE "does not honour|lands at the tail|not a working feature|KNOWN FAILURE|changes nothing|not windowed|no pool to apply|applier (ignores|appends)|iteration 2's client|blocked on #21|#2164"` across `crates/ python/ examples/ docs/ tests/ README.md ROADMAP.md`, excluding `node_modules` and the generated bundles — **found no site still asserting the withdrawn diagnosis.** Every surviving hit is legitimate:

- `crates/djust_vdom/src/diff.rs:169`, `python/djust/config.py:177-179`, `docs/adr/026-...:92`, `python/djust/tests/test_virtual_keyed_config_2017.py:14-19` — quote-to-withdraw. Correct, not findings.
- `examples/demo_project/djust_tests/views/virtual_keyed_demo.py:13` ("the append-only absorb heuristic lands at the tail instead") and `docs/website/guides/large-lists.md:64` — accurate descriptions of the **absorb heuristic** (the flag-OFF path), which really is append-only. That is the *motivation* for the keyed ops, not the withdrawn diagnosis.
- `crates/djust_vdom/tests/virtual_keyed_ops_2017.rs:328` — the reference applier's `before_key`-missing semantics. Correct.
- JS test headers (`tests/js/virtual-keyed-ops-2017.test.js`, `dj-virtual-patch-miss-diagnostic-2017.test.js`) — clean.

One adjacent artifact the sweep did leave behind, but it is a *different* stale claim rather than the withdrawn diagnosis — filed as **N2** below.

### 🔴 F2 — no CHANGELOG on a `fix(rust):` PR — **ADDRESSED**

`[Unreleased]` now carries a **Fixed** entry (clippy 1.97 blocker) and a **Changed** entry (the five corrected artifacts). Both are substantive and match the diff.

Already-tagged sections verified per #2028:

- `[1.1.0rc9]`: extracted from `CHANGELOG.md` and from `git show v1.1.0rc9:CHANGELOG.md` — **byte-identical**, 166298 chars each.
- `[1.1.0] - 2026-08-10`: there is **no `v1.1.0` tag yet** (`git tag --list 'v1.1.0*'` stops at `rc9`), so there is nothing to pin against. The proof there is structural instead: `git diff 555526ac..HEAD -- CHANGELOG.md --numstat` = **`8  0`** — eight insertions, **zero deletions**, all immediately under `## [Unreleased]`. No tagged section could have changed.

The commit message's supporting claim checks out too: `CHANGELOG.md` is inside `paths-ignore` at `.github/workflows/test.yml:9` and `:15`, so CI structurally cannot catch the omission.

### 🟡 F3 — the "feature-gated" claim was false — **ADDRESSED**

(a) `9b51c994` exists, is empty-tree, and quotes the false sentence verbatim before retracting it. (b) The PR body carries the ⚠️ Correction block and no longer asserts the claim anywhere.

(c) **I did not take the retraction's reasoning on trust — I confirmed it on both axes:**

- *Structural*: `crates/djust_live/Cargo.toml:15-29` — the crate's only feature is `extension-module`, and `default = ["extension-module"]`. `--all-features` adds nothing for that crate. The retraction is right.
- *Empirical (#1459 canary)*: in my worktree I reverted **only** the `lib.rs` hunk (asserting the mutation text was found and that the file actually changed), then ran `cargo clippy --workspace --all-targets -- -D warnings` — **without** `--all-features` — on `rustc 1.97.1 (8bab26f4f)`:

```
error: this `match` expression can be replaced with `?`
     = help: ... rust-1.97.0/index.html#question_mark
error: could not compile `djust_live` (lib) due to 1 previous error
error: could not compile `djust_live` (lib test) due to 1 previous error
```

The lint fires without `--all-features`. The retraction is correct and the original claim was false. File restored; worktree clean.

### 🟡 F4 — "flag OFF degrades silently" rested on excluded runs — **ADDRESSED (argument now sound)**

I cannot re-run a browser, so I checked whether the mechanism is verifiable in source. It is, on both arms:

- **ON arm.** `12-vdom-patch.js:2270-2275`: when `window.djust._virtualKeyedOp(node, patch, insertNode)` returns falsy the applier warns *"…targeted a node that is not an initialised virtual list"* and `return false`; a false patch reaches the `${failedCount}/${patches.length} patches failed` path at `:2447`. And `29-virtual-list.js:763-764` — `virtualInsert` returns `false` exactly when `STATE.get(container)` is empty, which is the post-teardown / post-morph state. The reported `1/3 patches failed → recovery` chain is real code, not inference.
- **OFF arm.** On an uninitialised container every one of the 60 rows is attached at its true index, so a positional `InsertChild path [5] index 5` is correct *by construction*. Nothing about that needed measuring.

**Reproduction fidelity of the substitute is adequate.** `teardownVirtualList` is a genuine exported helper (`29-virtual-list.js:663`, exported `:1009`, documented as a test helper at `:36`), and it produces the same observable triple (`pool 0 / rows 60 / shell false`) reported for the natural #2185 state — which is what the applier's guard keys on. Using a deterministic teardown instead of waiting on an intermittent load is the right call, and it is the honest kind of substitution because the guard cannot distinguish the two.

No overclaim in the commit or PR text on this point — one wording imprecision in the CHANGELOG only, at **N4** below.

---

## New findings

### 🟡 N1 — the ADR's "The original text is kept below for the record" is false

`docs/adr/026-dj-virtual-differ-awareness.md:116`. The block opens *"this table and **the paragraphs that followed it** were wrong"* and closes with that sentence — but those two paragraphs ("The OFF arm landing at the tail is exactly what this ADR predicts…" and "The differ is NOT at fault… **Iteration 2's client applier is what does not honour it**… Tracked as #2164") were **deleted**, not retained. What survives below the block is the pre-existing *first*-diagnosis retraction at `:118` ("A first pass recorded the cause as…"), and the one thing that genuinely was kept — the original A/B table — sits **above** the block, not below.

Deleting them is fine. Saying they were kept is the #1867 class exactly: a prose invariant nobody ran against the artifact, in the PR written to remove precisely that. Suggest: *"The original table is kept above; the paragraphs it supported are removed."*

### 🟡 N2 — a sixth artifact still carries a superseded reason for the flag being OFF

`python/djust/tests/test_virtual_keyed_config_2017.py:21-24`. This file is **not** an F1 miss — its `:14-19` already marks the withdrawn diagnosis correctly. But its closing paragraph gives a *different*, now-stale rationale:

> The default stays OFF because flipping it changes VDOM behaviour for every `[dj-virtual]` user, which is what #1122 exists to gate, and **the browser evidence so far covers `VirtualInsert` only — one variant of a multi-variant op set. Extending it to Move/Remove/Update is ADR-026 iteration 3 (#2017).**

This PR's own re-measurement covers remove (`VirtualRemove`), reverse (`VirtualMove`) and edit — as stated in the ADR `SUPERSEDED` table, `diff.rs:176-180`, `config.py:182-184`, the demo banner and the CHANGELOG. And #2185 appears nowhere in the docstring. So the CHANGELOG's "**Five** artifacts … now record the real state" leaves a sixth telling readers a reason this same PR disproves.

Same shape as F1, one link over (#2142 chain convention). One-paragraph fix; worth taking here rather than trailing.

### nit N3 — CHANGELOG cites the pre-fix line

The entry cites `crates/djust_live/src/lib.rs:1324`; post-fix the statement is at **`:1326`** (`:1324` is now the first line of the new two-line comment). Defensible given the "…, now `let roots = …`" phrasing, but a reader jumping there lands two lines high. `crates/djust_components/src/simple/tooltip.rs:62` is exact ✅.

### nit N4 — "the OFF path degrades silently" is unqualified in the CHANGELOG only

In the ADR, `diff.rs` and `config.py` the phrase is immediately qualified — *"— the server sends an ordinary `InsertChild`, which applies —"* — so it correctly reads as "the **virtualisation** is silently lost." The CHANGELOG's "…forces a full HTML recovery where the OFF path degrades silently" drops the qualifier, which invites the reading that OFF also produces a wrong DOM. It does not: OFF is *correct* at the patch level on such a load, and what is lost is the windowing — which is #2185 itself and flag-independent.

---

## Regressions and verification

| Check | Result |
|---|---|
| `cargo clippy --all-targets --all-features -- -D warnings` (rustc 1.97.1) | **clean** |
| `cargo test -p djust_vdom --test virtual_keyed_ops_2017` | 21 passed |
| `make check-adr-status` | OK — 26 ADRs, 0 inconsistencies |
| `test_index.py` import + `get_context_data()` | ✅ card `passing`, counts 4/0/0 |
| `tests/index.html`, `tests/virtual_keyed_demo.html` | both load |
| `virtual_keyed_demo.py` import | ✅ |
| `config.py` parses; `virtual_keyed_ops` default | ✅ `False`, and `_rust.virtual_keyed_ops_enabled()` `False` |
| `pytest tests/ python/tests/ python/djust/tests/ -n auto` | **10237 passed, 219 skipped** |
| CI at `9b51c994` | all green — `test-summary` SUCCESS, plus `demo-checks`, `rust-tests`, both python matrices, `browser-smoke`, `benchmarks` |

**Both Rust hunks re-confirmed behaviour-preserving.** `tooltip.rs:62` drops a no-op borrow inside `format!` (`format_args!` captures by reference either way). `lib.rs:1326`: `try_parse_cache_splice -> Option<VNode>`, so `?` desugars to `return None`; `get_parsed` returns an owned `Option<Vec<VNode>>`, so no borrow survives; `or_insert` (not `or_insert_with`) evaluates eagerly in both forms, so `parse_hits` bumps identically. And `git diff 555526ac..HEAD -- crates/djust_vdom/src/diff.rs` filtered to non-`///` lines is **empty** — that file's change is 100% doc comment.

### Observation, not a finding against this PR — a pre-existing `-n auto` order flake

My **first** full-suite run on the PR head reported `1 failed, 10236 passed`: `python/djust/tests/test_recovery_version_staleness_1817.py::test_time_travel_jump_recovery_version_is_current`. I chased it rather than waving it off:

- In isolation: **3 passed**.
- Re-run of the identical suite on the same PR head: **10237 passed**.
- `main` at `555526ac`, same command: **10237 passed**.

It is non-deterministic under xdist sharding, and it cannot be this PR's doing: the entire Python diff is a comment block in `config.py` plus string literals in a demo view, and the loaded `_rust` extension in my run was prebuilt from `main`, so the Rust hunks were not even in the tested binary. Worth its own issue per the repo's flaky-test canon (assert an ordering invariant, not a time-bounded/order-dependent one), separate from #2186.

### Merge hygiene

`2919b68e`'s message still contains the false "feature-gated" sentence, and `9b51c994` retracts it by quotation — the right handling for already-pushed history. If this is squash-merged, keep a body that carries the retraction (the PR body does, and GitHub's default concatenation keeps both). A hand-written squash body containing only the first commit's text would re-publish the false claim into `main`'s log.

---

## Verdict

The four Stage 11 findings are all genuinely addressed, and two of them (F3, F1's ADR claim) I re-derived independently rather than accepting. The verdict not to flip the flag is well-supported and the artifact corrections are a real improvement over what was on `main`.

Not blocking merge — but **N1 and N2 are the same defect class this PR exists to remove**, and both are one-paragraph edits. Strongly suggest taking them here so the accuracy PR is itself accurate; N3/N4 are optional polish.

COMMENT

johnrtipton:
## Retrospective — PR #2186 (pipeline-run, #2017)

**Task**: ADR-026 iteration 3 — flip `virtual_keyed_ops` to default ON.
**Outcome**: gate FAILS; flag stays OFF. Shipped a commit-blocking clippy fix, six corrected artifacts, and two new issues (#2185, #2187).

### Quality: 3

Three, not higher, because two review rounds found errors in claims I had *just made* — and one of them was the exact failure mode this PR exists to clean up.

### What went well

- **The gate was run as written rather than inferred.** The ROADMAP asked for real browser evidence; the CHANGELOG already asserted the answer. Re-deriving it found the assertion was right about the ops and wrong about why, which is what produced #2185.
- **Asserting initialisation before acting** is what separated signal from noise. Every early reading that skipped it measured a race and pointed at the wrong conclusion.
- **Deterministic teardown beat waiting for an intermittent bug.** `teardownVirtualList` reproduced the affected-load state on demand, which turned an unfalsifiable "sometimes it breaks" into a repeatable A/B.

### What didn't

- **I shipped a sweep that was 2 of 6.** The demo index card rendered a red `FAILING` badge linking to the page whose banner I had just rewritten to say the opposite — the app asserting both halves of a contradiction. I fixed the two artifacts I had *already read* and never grepped for the class. Two review rounds were spent finding the other four.
- **I asserted a mechanism I had not verified.** "The `djust_live` lint is feature-gated" was invented to explain why `--workspace --all-targets` looked clean. It was not feature-gated; my run had *aborted* at the `djust_components` error, so `djust_live` was never checked. I read "no error in the output" as "clean" — a false negative from an aborted build. Retracted in `9b51c99`.
- **I wrote "the original text is kept below for the record" and deleted it.** In a supersession block whose entire purpose is preserving a withdrawn claim. That is the same inaccuracy class the PR is cleaning up, committed while cleaning it up.
- **My "OFF degrades silently" claim rested on runs my own caveat excluded.** The reviewer caught that I was leaning on `pool == 0` runs I had labelled non-load-bearing. The conclusion survived measurement, but it was under-evidenced when I made it.

### Relation to earlier work in this milestone

This issue now carries **three** withdrawn diagnoses (#2164's two, plus my initial "the keyed ops fail on first use"). The pattern in CLAUDE.md's chain-shaped-issues section held exactly: each round's fix surfaced the next instance, and two of my three errors were introduced *by the previous round's fix*. The #2164 lesson — "no client-side warning ever fired was the tell, and it was read as 'the flag is irrelevant' rather than 'the flag is not connected'" — repeated in form: I read an aborted clippy build as a passing one.

The durable takeaway is narrower than "be careful": **when an artifact says a feature is broken, grep for the class before editing the instance.** Six artifacts encoded one withdrawn claim across Rust, Python, an ADR, a template, a view and a test docstring — no single file read would ever have found them.

### Verified

- `cargo clippy --all-targets --all-features -- -D warnings` clean on 1.97.1
- 10238 passed / 218 skipped; `test_virtual_keyed_config_2017.py` 7 passed
- `make check-adr-status` clean (26 ADRs)
- `[1.1.0rc9]` CHANGELOG section byte-identical to its tag (#2028)
- Re-reviewer independently re-derived the ADR claim from git history and reproduced the clippy retraction empirically

### Deferred

- **#2185** — the real blocker; iteration 3 should be re-attempted once it lands.
- **#2187** — pre-existing xdist order flake, isolated during re-review, unrelated to this PR.

`RETRO_COMPLETE`

Issue #2187, opened 2026-08-11
test: an xdist order-dependent flake surfaces on some -n auto runs (pre-existing)


johnrtipton:
Surfaced by the Stage 13 re-reviewer of PR #2186 and chased down there. **Pre-existing and unrelated to that PR** — filed separately so it is not lost, per the deferred-findings convention.

## What was seen

A single test failed on the reviewer's *first* full-suite run under `pytest -n auto`, then:

- passed in isolation,
- passed on a plain re-run of the same suite,
- and `main` runs clean.

That is the signature of an **order-dependent (xdist sharding) flake**, not a real regression: which tests land on the same worker varies between runs, so a leaked module-global or singleton only bites on some shardings.

## Why it matters

This repo already carries scar tissue for exactly this class:

- the 3-clean-runs verification gate for pollution-class fixes (Action #182 / #1174), which exists because *"pollution by definition shows up under specific orderings"*;
- the regroup xdist pollution root-cause work (#2053);
- the `sys.modules`-rebind polluter found only on a third verification run (PR #1159).

A flake that appears on run 1 and not run 2 will, sooner or later, be dismissed as noise by someone under time pressure — and the pre-push hook runs the suite, so it costs a false-failed push each time it lands.

## Suggested approach

Follow the #2053 playbook rather than re-deriving it:

1. Reproduce with a pinned distribution — `pytest -p xdist -n 4 --dist loadfile` and, separately, `-p no:randomly` with a fixed `-n`, to separate *ordering* from *sharding* as the variable.
2. Bisect the polluter with `pytest --deselect` halving, or run the failing test preceded by each candidate module.
3. Once found, isolate at the source (fixture teardown / `monkeypatch` instead of a module-global rebind) rather than adding a retry.
4. Gate with the 3-consecutive-clean-runs rule (#182) — a single green run is explicitly insufficient for this class.

Capturing the exact failing test id on the next occurrence would shorten step 1 considerably; the re-review transcript has it, and any future sighting should be added here.

johnrtipton:
**Sighting with the exact test id** — this issue asks for one, so here it is.

CI, PR #2213, `python-tests (py3.12)`:

```
python/djust/tests/test_recovery_version_staleness_1817.py:209
    assert v_arm == v_mount + 1
E   AssertionError: arming event must be 2, got 3
```

[...]

PR #2196, opened 2026-08-20
fix(client): run post-mount work when the tab is hidden (#2194)

### LINKED ISSUE #2194: post-mount reinitAfterDOMUpdate is gated behind requestAnimationFrame, which never fires in a hidden tab
Split out of #2185. The init-order half of that issue is fixed; this is the second, independent half.

## The gap

`03-websocket.js:627-630` schedules all post-mount work through `requestAnimationFrame`:

```js
if (typeof requestAnimationFrame === 'function') {
    requestAnimationFrame(runPostMount);
} else {
    runPostMount();
}
```

`runPostMount` is what calls `reinitAfterDOMUpdate()`, sets `window.djust._mountReady = true`, and triggers form recovery / `dj-auto-recover`.

**Browsers do not fire rAF callbacks in a hidden tab.** So on any page whose tab is not visible at mount time, none of that runs.

## Measured

On `/tests/virtual-keyed/` in a hidden tab, tapping the functions themselves (not inferring from symptoms):

```
initVL   t=57    <- init runs, shell created
setup    t=57
morph    t=104   <- the #1610 mount morph re-creates dj-root, wiping the shell
(no reinitAfterDOMUpdate event at all)
```

```js
document.visibilityState  // "hidden"
```

Resulting DOM state: `pool: 60` (STATE exists, so `setup()` did run) but `shell: false`, `children: 60` — all rows rendered, not virtualized. Calling `djust.initVirtualLists(document)` by hand repairs it instantly (`shell: true, children: 2, rows: 13`), which confirms the recovery path itself is correct and simply is not reached.

## Real-world exposure

Tabs are hidden at load in ordinary situations:

- a link opened in a background tab (ctrl/cmd-click, middle-click)
- session restore on browser start
- speculative prerendering

**Not yet established, and the thing to check first:** whether a queued rAF callback fires when the tab is later focused. If it does, impact is a delayed/incorrect first paint that self-heals on focus; if it does not, the page stays permanently un-initialized (no `_mountReady`, no form recovery, no `dj-auto-recover`). I measured only the hidden case. Please verify both before deciding severity — I do not want to overstate this the way #2185's original diagnosis was overstated.

## Note on the measurements in #2185

My browser measurements throughout the #2017/#2185 work were taken in an automation tab that was `hidden`. That does not affect the init-order finding (readyState-driven, verified independently), but any observation there that depended on post-mount work having run should be re-checked in a visible tab.

## Suggested direction

`requestAnimationFrame` is the right tool for the *paint-timing* reason it was chosen (FIX #619 — avoiding a mid-paint layout recalc that flashes on pages with large pre-rendered elements). It is the wrong tool as the only path to correctness-critical work.

Options worth weighing:

1. Keep rAF for the visual work, but run `reinitAfterDOMUpdate()` unconditionally (it is idempotent — `initVirtualLists` checks `STATE` and `structureIntact`).
2. Race rAF against a `setTimeout(…, 0)` fallback, first-one-wins with a guard flag.
3. Schedule via rAF but add a `visibilitychange` listener that flushes if still pending.

(1) is the smallest and does not reintroduce the flash, since the flash came from layout-reading work, not from re-init per se — but that should be confirmed rather than assumed.

johnrtipton:
Closes #2194 — the second half of #2185, which turned out to be two bugs wearing one symptom. The init-order half shipped in #2195.

## The bug

`03-websocket.js` scheduled **all** post-mount work through a single rAF:

```js
if (typeof requestAnimationFrame === 'function') {
    requestAnimationFrame(runPostMount);
} else {
    runPostMount();
}
```

`runPostMount` is the only path to `reinitAfterDOMUpdate()`, `window.djust._mountReady`, and reconnect form-recovery / `dj-auto-recover`. **Browsers do not fire rAF in a hidden tab**, and the `else` branch does not help — it only covers "no rAF function at all", which never happens in a browser.

Pages are routinely hidden at mount: opened in a background tab, restored with the session, prerendered.

## Measured in a real hidden tab

| | `hidden` | `_mountReady` | shell | rows |
|---|---|---|---|---|
| before | `true` | **`undefined`** | ✗ | 60 |
| after | `true` | **`true`** | ✓ | **13** |

A rAF queued while hidden was still `QUEUED` seconds later, so the #1610 mount morph wiped the `[dj-virtual]` shell with nothing to restore it.

## The fix

When `document.hidden`, run immediately — there is no paint in progress, so **FIX #619's** reason for deferring (avoiding a mid-paint layout recalc that flashes on pre-rendered pages) does not apply. The visible path is unchanged, so #619 is untouched.

Plus a backstop for what `document.hidden` misses: a tab can be **occluded or minimised** with `hidden === false` while rAF is still throttled. rAF races a 250 ms timeout, first wins, guarded by `done` so post-mount runs exactly once — twice would re-enter form recovery and `dj-auto-recover`.

**250 ms, not 0.** A `setTimeout(0)` backstop would beat rAF's ~16 ms on *every visible mount* and silently undo #619 while looking correct. There is a test asserting the backstop exceeds a frame.

## Two existing #619 pins failed — correctly

`mount-deferred-reinit.test.js` and `ws-mount-prerender-divergence-1610.test.js` pinned the literal `requestAnimationFrame(runPostMount)` shape. They were right to fire; I rewrote them to pin the **invariant** (visible path defers past the paint; every path reaches `runPostMount`) rather than one call shape, which is what made them brittle.

Verified they still bite: **deleting the rAF defer entirely turns 3 of their cases red.**

## Verification

- 7 new cases in `tests/js/post-mount-hidden-tab-2194.test.js`
- **Gate-off**: restoring the rAF-only form fails **5 of 7** (the 2 survivors are the #619-preservation checks, which correctly still hold)
- JS: **172/172 files, 1933 tests**; Python: **10249 passed**
- `eslint --max-warnings 0` and both bundle lints clean


johnrtipton:
## Retrospective — PR #2196 (pipeline-drain)

**Task**: v1.1.1-1 — drained #2194 (and staged #2187 into the bucket).

### Quality: 4

### What went well

- **The severity question got answered by measurement, not argument.** The issue flagged an open question — does a queued rAF fire on refocus? I couldn't foreground the automation tab to test it, so instead of guessing I chose a fix that is correct either way. The `document.hidden` branch doesn't depend on knowing the refocus behaviour.
- **The before/after was captured in the exact failing condition**, not a proxy: same hidden tab, `_mountReady` `undefined` → `true`, shell absent → present, 60 rows → 13.
- **Two existing #619 pins failed and were right to.** They caught that I'd changed a shape they were guarding. Rewriting them to pin the *invariant* (visible path defers; every path reaches `runPostMount`) rather than the literal call left them stronger — verified by deleting the defer and watching 3 cases go red.
- **The 250 ms backstop got a test for the number itself.** A `setTimeout(0)` would beat rAF on every visible mount and silently undo #619 while looking correct; that's a trap worth pinning rather than commenting.

### What didn't

- **I built the fix branch on top of the un-PR'd roadmap branch** without noticing. It worked out — the ROADMAP drain rode along in the same PR and the diff is coherent — but it was luck, not intent. A drain that opened the roadmap PR first would have made both reviewable separately.
- **`document.hidden` alone would have been an incomplete fix**, and I only caught that while writing the comment: an occluded or minimised tab has `hidden === false` while rAF may still be throttled. The backstop exists because writing down *why* the branch was correct surfaced the case it didn't cover. That should have come from the analysis, not the prose.
- Third time this arc that "the fix is obvious" preceded a real gap (the promise-vs-`queueMicrotask` error-swallowing in #2195, the `code_only` hole in its test, and this). The pattern is consistent: the first shape is directionally right and misses one adjacent case.

### Verified

- Real hidden tab, before/after as above
- Gate-off: restoring the rAF-only form fails **5 of 7** new cases (the 2 survivors are #619-preservation checks that correctly still hold); deleting the #619 defer fails **3** of the updated pins
- JS **172/172 files, 1933 tests**; Python **10249 passed**; eslint + both bundle lints clean; CI 15 green

### Deferred

- **#2187** — staged into v1.1.1-1 but deliberately not worked: the failing test id was never captured, so step one is observational. Recorded in the ROADMAP as such so a future drain doesn't queue it as implementable and stall.
- **#2163** — still needs a decision from the maintainer on the duplicate-`test-summary`-context question before the required check can be enabled.

Issue #2166, opened 2026-07-28
audit: every reader of djust.config.config gets import-time defaults, not deployment settings


johnrtipton:
Generalised from #2164, where this made a shipped feature silently inert in production.

## The mechanism

`python/djust/config.py:531` constructs the singleton at **module import time**:

```python
config = LiveViewConfig()      # __init__ calls _load_from_settings() — once
```

Under a real ASGI server, `djust.config` is imported while the app registry is still populating, so `settings.LIVEVIEW_CONFIG` isn't readable yet. The singleton captures defaults and **never re-reads**. Measured:

```
at import ................ False
after django.setup() ..... False     <- never refreshes
```

There is a `reset()` (`config.py:505`) that re-runs `_load_from_settings()`, but nothing calls it at startup.

## Why it's dangerous

It fails **silently and asymmetrically**: correct in tests, wrong in production.

- Tests that patch `config._config` (or import `djust.config` *after* `django.setup()`) see the deployment's value.
- A real server sees defaults.

That combination is why #2164 shipped: `LIVEVIEW_CONFIG['virtual_keyed_ops'] = True` was ignored by every server while the suite was green, and I twice misdiagnosed the resulting behaviour as a bug in the VDOM differ and its client applier — both of which were correct.

## The audit

Find every reader and decide, per site, whether it can tolerate import-time defaults:

```bash
grep -rn "from djust.config import config\b" python/djust/
grep -rn "from .config import config\b" python/djust/
grep -rn "get_config()" python/djust/ | head -40
```

`get_config()` returns the same stale singleton (`config.py:534`), so it carries the identical hazard — the name suggests a fresh read and it is not.

Anything read **once at startup** is affected. Anything read **per request** may be fine only by accident, if something happened to import `djust.config` late.

## Candidate fixes

1. **Read `django.conf.settings` directly** at each startup-time site — what #2164 did. Correct, but doesn't scale.
2. **Refresh the singleton in `DjustConfig.ready()`** (`config.reset()`), so every later reader is correct by construction. One line, fixes the whole class — needs care that it doesn't clobber programmatic config set between import and `ready()`.
3. **Make `LiveViewConfig.__getitem__`/`get` lazy** — load from settings on first access after `apps.populate()`. Most robust, largest change.

(2) looks like the right default, with a test asserting a value set only in settings is visible through the singleton after `ready()`.

## Test-shape lesson

[...]

PR #2197, opened 2026-08-21
feat(vdom): flip virtual_keyed_ops ON by default — ADR-026 iteration 3 (#2017)

### LINKED ISSUE #2017: dj-virtual: deeper server-side reconcile — stream-op wiring, VDOM-differ awareness, out-of-window patch landing (follow-up to #1988/#1989)

johnrtipton:
Closes #2017. Ships ADR-026 iteration 3.

## Why this could not be decided before

ROADMAP gated the flip on *"#2136's fix plus real browser verification producing evidence worth flipping on."* #2136 shipped in PR #2146 — but the verification kept producing nothing usable, because **the control arm was as broken as the test arm.**

Two bugs had to land first:

| | |
|---|---|
| **#2185** (PR #2195) | `djustInit` ran mid-bundle, so `initVirtualLists` was never called on cached loads — the list simply was not virtualized |
| **#2194** (PR #2196) | post-mount reinit was rAF-only, so a hidden tab never recovered after the #1610 mount morph wiped the shell |

Until both were fixed, **every earlier A/B compared OFF against OFF.** That is how this one feature accumulated *four* withdrawn diagnoses — #2164's two, plus two of mine during the gate. Each was asserted from a partial read of a broken experiment.

## The evidence

Healthy initialized list, same start state, same sequence, flag the only variable:

| case | OFF | ON |
|---|---|---|
| insert at server position 5 | pool index **60 (tail)** ✗ | pool index **5** ✓ |
| remove at 3 | size unchanged, k3 still present ✗ | k3 gone, no duplicates ✓ |
| reverse | not reordered ✗ | exact reversal ✓ |
| edit a scrolled-out row | correct ✓ | correct ✓ |

0 console errors, no `Virtual*` warnings. **OFF is broken for every mid-list structural mutation on a windowed list; ON fixes all three and regresses none.**

## Blast radius

**Narrow by construction.** `emits_virtual_ops` requires the parent to carry `dj-virtual`, so a project not using it sees no wire change at all.

**Version skew degrades rather than breaks.** A cached pre-1.1.0 bundle against a 1.1.1+ server receives `Virtual*` ops its applier does not know — that takes the existing patch-failure path (request recovery HTML, morph), not a crash. Opt out with `LIVEVIEW_CONFIG['virtual_keyed_ops'] = False`.

## The guard fired, and was inverted rather than deleted

`test_the_config_default_is_off` existed precisely so this flip could not happen as drift. It failed on this change — correctly. It is now `test_the_config_default_is_on`, carrying the measurement table, and demands the same standard of evidence from anyone flipping it back.

Also updated everything still advertising OFF: ADR-026's iteration table, the demo index card, the demo banner.

## An aside worth recording

The full Python suite failed once during this work on `test_the_repos_own_lockfile_is_clean` — the `uv.lock` index guard I added this morning. `uv run maturin develop` had re-locked against the developer-local proxy, exactly the failure mode that guard was written for. It caught me within hours of existing, which is the best argument for it I could have asked for.

## Verification

- Python **10249 passed**, 218 skipped
- JS **172/172 files, 1933 tests**
- Rust **44 suites**
- `make check-adr-status` clean


johnrtipton (commented):
## Stage 11 adversarial review — PR #2197 (ADR-026 iteration 3)

Worktree-isolated, read-only against the main checkout (`git config core.bare` → `false`, verified before and after). Branch is **0 commits behind `origin/main`** (`git rev-list --count HEAD..origin/main` = 0), so this reviews the program the merge will apply.

The browser evidence is not what I found fault with — the OFF-arm failures it reports are consistent with the differ code, and the reasoning about why the earlier A/Bs were invalid holds up. What I found is that **the Rust half of the flip is unnecessary, breaks the crate's own dark-default tests, and silently disarms the one end-to-end test that proves the flag reaches Rust.**

---

### 🔴 1. The Rust default flip breaks two tests that pass only by alphabetical accident

`crates/djust_vdom/src/diff.rs:190` flips `AtomicBool::new(false)` → `AtomicBool::new(true)`. The PR does **not** touch `crates/djust_vdom/tests/virtual_keyed_ops_2017.rs`, which still contains two tests asserting the dark default. Both fail when run alone:

```
$ cargo test -p djust_vdom --test virtual_keyed_ops_2017 -- --exact flag_is_off_by_default
thread 'flag_is_off_by_default' panicked at crates/djust_vdom/tests/virtual_keyed_ops_2017.rs:102:5:
ADR-026 iteration 1 ships dark; the client cannot apply these ops yet
test result: FAILED. 0 passed; 1 failed

$ cargo test -p djust_vdom --test virtual_keyed_ops_2017 -- --exact with_flag_off_a_virtual_parent_diffs_exactly_like_a_plain_one
thread '...' panicked at crates/djust_vdom/tests/virtual_keyed_ops_2017.rs:120:5:
assertion `left == right` failed: with the flag OFF a [dj-virtual] parent must produce the SAME patch kinds as an ordinary parent
  left: ["VirtualMove"]
 right: ["MoveChild"]
test result: FAILED. 0 passed; 1 failed
```

They pass in the full run — including under `--test-threads=1` — because `FlagGuard::drop` (`:40-42`) unconditionally calls `set_virtual_keyed_ops(false)`, and both OFF tests sort near the **end** of the alphabetical order (`f…`, `w…`) behind a dozen `FlagGuard` tests. So the "Rust 44 suites" green is a **false green** for these two: the flag they assert on is being reset for them by an unrelated test's destructor. Rename either test, shard the file, or add a test sorting after them, and CI goes red for reasons nobody will connect to this PR.

This is the decorative-pin class from CLAUDE.md (#1859/#1860): `flag_is_off_by_default` exists *precisely* to make a flip impossible as drift — the same role as the Python `test_the_config_default_is_off` this PR correctly inverted. The Python guard fired and was dealt with honestly; its Rust twin was left behind. Same invariant, two implementations, one updated (#1646).

**The load-bearing question: the Rust default should have stayed `false`.**

`apps.py:82-89` states the design in its own words — *"It also keeps `_defaults` the single source of the default (ADR-026 iteration 3 flips it there, and this must follow)."* `DjustConfig.ready()` applies `_defaults` on **every** Django startup, so flipping the Rust static buys **zero production behaviour change**. What it does buy:

- **A second source of the default**, contradicting the comment above that the PR left in place.
- **The unwired state flips from fail-safe to fail-open.** Every path that previously degraded to OFF now degrades to ON: the `except Exception` at `apps.py:92-95` (logs and continues — Rust keeps its own default); `hasattr(_rust, "set_virtual_keyed_ops")` being False at `apps.py:90`, which is the *exact* case that guard was written for — a `.so` built between #2126 and #2165 has `emits_virtual_ops` but no setter, so it is now pinned ON with **no way for the user to opt out**; `djust` absent from `INSTALLED_APPS`; and any non-Django consumer of `djust_vdom`.
- The two broken tests above.

Recommendation: **revert `diff.rs:190` to `AtomicBool::new(false)`** and keep `config.py:202` at `True`. Production behaviour is identical, `_defaults` stays the single source as documented, the unwired state stays fail-safe, and both Rust tests stay meaningful. If you genuinely want the Rust static ON for embedders, then that is a separate decision that requires updating both tests, the module doc at `virtual_keyed_ops_2017.rs:8-11` ("*ships the differ side DARK: the flag defaults OFF*"), and the `_defaults`-single-source comment in `apps.py`.

Also: `apps.py:91` still reads `_cfg.get("virtual_keyed_ops", False)` — a **third** statement of the default, now disagreeing with the other two. Unreachable today (the key is always in `_defaults`), but it is exactly the drift this PR is otherwise trying to eliminate.

---

### 🔴 2. The #2166 regression test's two relevant assertions are now tautologies

`python/djust/tests/test_config_settings_recovery_2166.py:31-33` states its own contract:

> *"Values chosen to be the OPPOSITE of each key's default, so 'the settings were read' and 'the defaults were kept' are distinguishable. A probe using a value equal to the default could not tell the two apart."*

Line 39 is `"virtual_keyed_ops": True,          # default False`. After this PR the default **is** `True`, so the probe value equals the default and the comment is false. `EXPECTED` (`:85-91`) therefore asserts `virtual_keyed_ops: True` and `rust_flag: True` — neither of which can fail any more.

Empirically proven. I gated off the recovery (`apps.py:30` → `pass`), reintroducing the exact #2164/#2166 defect, and re-ran the file:

```
E   AssertionError: config keys were ignored when djust was imported before DJANGO_SETTINGS_MODULE was set.
E       wanted {'virtual_keyed_ops': True,  'loop_render_cache_enabled': False, 'template_auto_call': False, 'filter_bridge_warm': False, 'rust_flag': True}
E       got    {'virtual_keyed_ops': True,  'loop_render_cache_enabled': True,  'template_auto_call': True,  'filter_bridge_warm': None,  'rust_flag': True}
E     Omitting 2 identical items
```

The three other keys caught the defect. `virtual_keyed_ops` and `rust_flag` did not — pytest names them itself as the *"2 identical items"*. `rust_flag` was the **only** assertion anywhere in the suite that the flag reaches the Rust differ through the real production startup order, and this PR blinds it while changing that flag's default.

Fix is one line plus `EXPECTED`: make the probe `"virtual_keyed_ops": False,  # default True` and expect `False` / `rust_flag: False`. (`apps.py` uses the singleton, so an explicit `False` still round-trips through the same path.) Working tree restored afterwards — `git status --porcelain` clean.

---

### 🔴 3. The sweep missed six sites, one of them a self-contradiction inside a file this PR edited

The PR updated ADR-026's iteration table (`:70`) but not the **Status header three lines from the top of the same file**:

- `docs/adr/026-dj-virtual-differ-awareness.md:3` — *"Iteration 3 — the flag flip — is **NOT shipped**; the differ path is still dark by default."*

That file now says both. And `make check-adr-status` cannot catch it — the ADR itself says so at `:61-62` (*"it audits status/version-line consistency, not whether a status matches reality"*), so the PR's "`make check-adr-status` clean" is not evidence here.

Two **user-facing** guides still tell users this feature does not exist:

- `docs/website/guides/large-lists.md:64` — *"Keyed mid-list inserts, removals, and finalize-patch landing for an item scrolled OUT of the current window need differ-level `dj-virtual` awareness — tracked in the follow-up to this work."*
- `docs/website/guides/virtual-lists.md:71-73` — *"The absorb is **append-only** … Keyed mid-list inserts/removals … differ-level `dj-virtual` awareness (tracked as a follow-up)."*

These are the primary `dj-virtual` docs, linked from `docs/website/index.md:42` and `_config.yaml:114`. They describe the *pre-flip* behaviour as current.

Two current, non-historical ROADMAP statements:

- `ROADMAP.md:7` (top-of-file v1.1.0 status) — *"#2017 … the flag ships wired but **OFF**, per #1122 split-foundation"*
- `ROADMAP.md:38` (the **active** v1.1.1-1 bucket) — same claim, as a deferral

The PR closes #2017; the active bucket must record that, and the top-of-file status is wrong the moment this merges.

And two code comments:

- `crates/djust_vdom/src/lib.rs:677` — *"the feature flag is on (`diff::set_virtual_keyed_ops`, **default OFF**)"*
- `crates/djust_vdom/tests/virtual_keyed_ops_2017.rs:8-11` — module doc, *"ships the differ side DARK: the flag defaults OFF … flipped after a soak (iteration 3)"*

(For the record, `12-vdom-patch.js:1302-1306`'s "Iter 3" is the v0.9.4-1 `dj-if` iteration, **not** ADR-026 — correctly left alone.)

---

### 🟢 4. The version-skew claim is TRUE — verified, not taken on trust

Traced end to end:

1. `12-vdom-patch.js:2280-2284` — the `default:` arm of the patch switch does `console.warn('[LiveView] Unknown patch type:', safeType)` and **`return false`**. It does not throw and does not silently no-op.
2. `_applyPatchesInner` counts it as failed (`:2444-2456`) and returns `false`.
3. `02-response-handler.js:123-141` — on `success === false` it rewinds `clientVdomVersion = data.version - 1` and sends `{type:'request_html'}`, or reloads if the socket is down.

So "degrades to request recovery HTML and morph, not a crash" is accurate. 🟡 worth stating in the CHANGELOG though: for such a client this is a **full-HTML round trip on every structural mutation of a virtual list**, and the recovery morph is the #1610 path that wipes the `[dj-virtual]` shell — the very interaction #2194 was about. "Degrades" is right; "cheaply" would not be.

Minor: iterations 1-2 shipped in **v1.1.0rc9** (ADR `:3`), which is itself pre-1.1.0 by SemVer ordering, so "a cached pre-1.1.0 bundle" is over-inclusive. Harmless direction.

---

### 🟢 5. What holds up

- **The inverted guard is load-bearing.** Gate-off: setting `config.py:202` back to `False` fails `test_the_config_default_is_on` with its own message. Not a tautology.
- **The OFF path still has real Python coverage.** `test_the_differ_emits_keyed_ops_only_when_enabled` (`test_virtual_keyed_config_2017.py:242-248`) drives `set_virtual_keyed_ops(False)` then `True` through the real `diff_html` and asserts both arms. `test_ready_applies_the_config_value_to_rust` (`:140`) pre-sets the opposite value so a no-op applier fails. Both survive the flip.
- **CHANGELOG is purely additive** — `git diff origin/main...HEAD -- CHANGELOG.md` has zero `-` lines, so no already-tagged section was touched (the #2028 class).
- `python/djust/tests`: **5202 passed, 200 skipped**. Full `djust_vdom` crate: 26 binaries green.

---

### 🟡 6. The evidence covers one list configuration; the flip is global

The demo drives a **fixed-height** list — `virtual_keyed_demo.html:34`, `dj-virtual-item-height="32"`. Variable-height lists were never exercised, and `docs/website/guides/virtual-lists.md:46-51` documents a known interaction that this flip makes default-on:

> *"The current height cache is keyed by item index. If you reorder items (sort, insertion in the middle), cached heights bind to the wrong items until re-measurement… a `data-key`-based cache is planned (tracking issue #951)."*

Mid-list insert and reorder are precisely the two cases the flip enables, and on a variable-height list they now hit an index-keyed cache that #951 says mis-binds. Not necessarily a blocker, but the blast-radius section claims "regresses none" on the strength of a four-case fixed-height run. Either measure one variable-height case or scope the claim to fixed-height and note #951.

🟡 Six sites hardcode **"1.1.1"** (`config.py:175`, `diff.rs:168`, `test_virtual_keyed_config_2017.py:73` and `:101`, `virtual_keyed_demo.html:7`, `test_index.py:33` and `:41`) for a version that does not exist yet — `pyproject.toml:7` is `1.1.0`. `ROADMAP.md:47` does target v1.1.1, so it is a reasonable bet, but five copies of an unreleased version number is five things to fix if the bucket slips.

---

### Summary

| | Finding |
|---|---|
| 🔴 | `diff.rs:190` Rust default flip — unnecessary, fail-open, and breaks 2 dark-default tests that pass only by alphabetical accident. Revert to `false`; keep `config.py` at `True`. |
| 🔴 | `test_config_settings_recovery_2166.py:39,85-91` — `virtual_keyed_ops`/`rust_flag` became tautologies; empirically proven with the recovery gated off. |
| 🔴 | Six missed sweep sites, incl. a self-contradiction at `docs/adr/026-…:3` and two user-facing guides. |
| 🟢 | Version-skew claim verified true (`12-vdom-patch.js:2284` → `02-response-handler.js:123`). |
| 🟢 | Inverted guard gate-off passes; OFF-path Python coverage intact; CHANGELOG additive; 5202 Python + full `djust_vdom` green. |
| 🟡 | Fixed-height-only evidence vs. #951 index-keyed height cache; `apps.py:91` third default; five hardcoded "1.1.1". |

Findings 1 and 2 are the ones I would not merge without. Both are the same shape the PR body itself is about — a claim that looks verified because a green suite agreed with it, where the suite could not have disagreed.

REQUEST_CHANGES

johnrtipton:
## All three 🔴 addressed — and you were right that the Rust flip was the dangerous one

Two of these you proved empirically rather than argued, which is why they landed.

### 🔴 1 — Rust static reverted to `false` (`222eca40`)

Confirmed your false-green claim directly:

```
$ cargo test -p djust_vdom --test virtual_keyed_ops_2017 flag_is_off_by_default -- --exact
test flag_is_off_by_default ... FAILED
panicked at crates/djust_vdom/tests/virtual_keyed_ops_2017.rs:102:5
```

It passed in my full run only because `FlagGuard::drop` resets the flag and it sorts late. **My "Rust 44 suites green" was worthless** — a whole-suite pass that concealed a test failing in isolation. After the revert it passes standalone.

Your reasoning about it buying nothing is also right, and worse than you put it: `apps.py:88`'s own comment already said *"`_defaults` is the single source of the default (ADR-026 iteration 3 flips it there, and this must follow)"* — I wrote that during #2167 and then flipped the static anyway, contradicting my own documented design. The static now stays `false` and carries the fail-safe reasoning, so an unwired embedding, a process where `ready()` never fired, or a stale `.so` cannot end up fail-open with no opt-out.

### 🔴 2 — probe un-tautologised (`222eca40`)

Exactly as diagnosed. The probe block states the rule in its own comment — values must be the opposite of each default — and flipping the default made `"virtual_keyed_ops": True` equal to it, taking `rust_flag` down with it. Losing `rust_flag` mattered most: it was the only end-to-end proof the flag reaches Rust through real startup order.

Probe and expectation inverted to `False`. Re-ran your gate-off (neutering `ensure_settings_loaded()`) and both now surface as explicit mismatches:

```
{'virtual_keyed_ops': True} != {'virtual_keyed_ops': False}
{'rust_flag': True} != {'rust_flag': False}
```

No more `Omitting 2 identical items`.

### 🔴 3 — six sites swept (`222eca40`)

Third incomplete sweep on this one feature, which is its own signal. Worst was the self-contradiction you flagged: ADR-026's **Status** line still read "iterations 1-2 shipped", three lines from the table I had just edited. The two user-facing guides were the most consequential — `large-lists.md` and `virtual-lists.md` were both telling readers that keyed mid-list inserts are "tracked as a follow-up", i.e. describing the absence of the exact feature this PR ships.

Also fixed `lib.rs`'s "default OFF" comment and two stale rationales in the test docstring.

Deliberately **not** changed, and worth confirming I read them correctly: the Rust test comments about shipping "dark" are **correct again** after finding 1's revert, and the ADR/ROADMAP lines describing what iteration 1 shipped are history rather than drift.

### On the version-skew claim

Thanks for tracing it to `12-vdom-patch.js:2284` → `02-response-handler.js:123` rather than taking my word for it. Good to have that verified independently.

### Verification after the fixes

- Python **10249 passed**, 218 skipped
- JS **172/172 files, 1933 tests**
- `cargo test -p djust_vdom` 26 suites, **and `flag_is_off_by_default` green in isolation** — the check my original run lacked
- `check-adr-status` + `check-doc-snippets` clean

Ready for re-review.

johnrtipton (commented):
## Re-review — PR #2197 (independent verification of commit `222eca40`)

Worktree-isolated; `git config core.bare` verified `false` before and after (#300). Branch is **0 commits behind `origin/main`** (#1450 passes).

Everything below was run, not inspected. **F1 and F2 are genuinely fixed — including the one that mattered most.** F3 is not, and the miss is now larger than the one that was reported, not smaller.

---

### F1 — Rust static reverted to `false` — ✅ **ADDRESSED**

- `crates/djust_vdom/src/diff.rs:203` → `AtomicBool::new(false)`. The arc: `origin/main` = `false`, `f9d1409f` = **`true`** (the bug), `222eca40` = `false`.
- Both dark-default tests pass **individually under `--exact`** — the check that exposed the false green:
  - `flag_is_off_by_default` → `ok. 1 passed; 20 filtered out`
  - `with_flag_off_a_virtual_parent_diffs_exactly_like_a_plain_one` → `ok. 1 passed; 20 filtered out`
- `git diff origin/main HEAD -- crates/` with comments/blanks stripped returns **zero** lines — the Rust change is comment-only.

**And the feature is not a no-op.** Static `false` + Python `True` only works if `ready()` applies it. Ran a real `django.setup()` with **no `LIVEVIEW_CONFIG` at all** (pure defaults), against the worktree's modules (verified by `__file__`; `main`'s default is `False`, so a mis-resolved import would have shown):

```
RUST_STATIC_BEFORE_SETUP: False
RUST_AFTER_READY:         True
PY_DEFAULTS:              True
PY_CONFIG_GET:            True
```

`python/djust/apps.py:91` does reach Rust. The fail-safe-when-unwired rationale is sound and now matches the code.

### F2 — probe un-tautologised — ✅ **ADDRESSED**

- `test_config_settings_recovery_2166.py:39` probe `False`; `:93` EXPECTED `False`; `:99` `rust_flag` `False` — all now opposite the new default.
- **Independently re-ran the gate-off** (neutered `ensure_settings_loaded()` to `return False` before the reload; cleared `__pycache__` first to avoid a stale-bytecode false pass). All **five** keys report as explicit mismatches, no "Omitting N identical items":

```
Differing items:
{'rust_flag': True} != {'rust_flag': False}
{'virtual_keyed_ops': True} != {'virtual_keyed_ops': False}
{'template_auto_call': True} != {'template_auto_call': False}
{'loop_render_cache_enabled': True} != {'loop_render_cache_enabled': False}
{'filter_bridge_warm': None} != {'filter_bridge_warm': False}
```

`config.py` restored **byte-identical** (md5 `e3dbb585…` both sides; `git diff` empty).

- **No other probe key collides with its default.** `filter_bridge_warm` is not in `_defaults` at all (resolves to `None`), so `False` vs `None` still distinguishes — not a tautology, but the inline `# default True` at `:42` is inaccurate. Pre-existing on `main`; 🟢 worth fixing in passing.
- Good second-order effect: this key is now real **opt-out coverage** — it proves settings can turn the flag *off* through actual startup order.

### F3 — sweep completeness — 🔴 **NOT ADDRESSED (15 sites)**

I swept independently rather than checking the author's list. Four of these are **inside files this PR edited**, and one is in the **CHANGELOG entry this PR wrote**.

#### 🔴 Same-release self-contradictions

**1. `CHANGELOG.md:29-31` — two `### Changed` blocks in the *same* `[Unreleased]` section say opposite things.** `## [Unreleased]` runs lines 8-32; `## [1.1.0]` starts at 33. The PR's new entry (line 12) says ON by default. The pre-existing entry at line 31 ends:

> *"The default stays **OFF**, but now for the actual reason: **#2185** … No behaviour change."*

Both ship in 1.1.1. A release-notes reader gets a flat contradiction. **I want to correct my own first-pass here:** structurally the CHANGELOG is fine — purely additive, 0 removed lines vs `main` *and* vs the `v1.1.0rc9` tag, so canon #2028 is satisfied. The defect is *internal consistency of the unreleased section*, which the #2028 check does not look at. The line-31 entry needs a superseding clause, not deletion.

**2. `crates/djust_vdom/src/diff.rs:156-159` and `:164` — the doc comment opens with the claim it later refutes.** The PR rewrote `:165-202` of this comment and left the head:

> `/// Default OFF. Iteration 1 of the ADR ships the differ side dark … until the client half (iteration 2) can apply them and the flag is flipped after a soak (iteration 3).`

…then `:168` says `/// The DEFAULT is ON since 1.1.1`. A reader hits the false one first. `:164` likewise: *"while it is still dark."* (`:190-202`, the fail-safe-static note, is correct — keep it.)

**3. `docs/adr/026-dj-virtual-differ-awareness.md:3` — the Status line contradicts itself within one sentence.**

> `Status: … iteration 3 (default ON) in 1.1.1 (PRs …). **Iteration 3 — the flag flip — is NOT shipped; the differ path is still dark by default.**`

The PR updated the clause before the parenthetical and left the one after it. `make check-adr-status` passes green on this (`OK — 26 ADRs scanned, 0 inconsistencies`) — which is precisely what **ADR-026:58-62 already warns about**: *"it audits status/version-line consistency, not whether a status matches reality."* This PR re-commits the failure mode the ADR documents.

**4. `python/djust/tests/test_virtual_keyed_config_2017.py:30-36`** — the docstring the PR edited at `:13-22` still ends *"What keeps it OFF is #2185 … #2185 is the blocker."* Seventy lines below, the PR's own assertion is `is True` with the message *"ON is the shipped behaviour since 1.1.1."*

**5. `docs/adr/026…md:72-76`** — flat and false: *"The DEFAULT half did not: it stays `False`."*

#### 🟡 Remaining drift

| # | site | stale claim |
|---|---|---|
| 6 | `ROADMAP.md:37-40` | The **currently-open `v1.1.1-1` bucket** still reads *"Deferred and NOT in this bucket: **#2017** … the flag ships wired but OFF."* This PR ships it *in 1.1.1*. `ROADMAP.md:47` notes `/pipeline-next` reads these rows to pick work. ROADMAP.md is **not in the PR diff at all**, and no row records the flip. |
| 7 | `ROADMAP.md:6-8` | Top-of-file status banner, same claim. |
| 8 | `docs/adr/026…md:108-113` | *"Iteration 3 is now blocked on **#2185** … Flip the default once #2185 is fixed."* Sits in the 2026-08-11 correction block, which is presented as current truth (the separately-marked *"superseded text, verbatim"* sub-block is below it). |
| 9 | `docs/adr/026…md:196-197` | Option D's *"**Verdict** — this is the current state"* — Option A is now the current state. |
| 10 | `crates/djust_vdom/tests/virtual_keyed_ops_2017.rs:104` | Assert message: *"the client cannot apply these ops yet"* — false since iteration 2 (PR #2135). It is the message a future engineer reads when this test breaks; it will send them the wrong way. The assertion itself is correct — **only the message is wrong**. |
| 11 | `examples/demo_project/djust_tests/views/virtual_keyed_demo.py:3-5, 12-15` | *"ROADMAP.md requires real browser evidence before flipping `virtual_keyed_ops` on"*, and *"`insert_middle` — … which the append-only absorb heuristic lands at the tail instead"* (now the flag-OFF behaviour only). The PR updated this demo's **template banner and index card** but not the view backing them — file absent from the diff. |
| 12 | `tests/js/virtual-keyed-ops-2017.test.js:8` | *"Those ops ship dark — the differ's flag defaults off."* |
| 13 | `docs/adr/026…md:74` | Stale citation `diff.rs:167`; the static is at **`:203`** after this PR's own comment expansion (#1197). |

Also 🟡: **14.** `test_virtual_keyed_config_2017.py:22-24` — the edit mangled the sentence (*"The later one said that"* dangles into an orphaned *`"browser evidence covers VirtualInsert only" line was true when written`*). **15.** `RETRO.md:360` (Action Tracker row #318, Status `Open`) asserts two blockers that are both cleared.

Adjacent drift, not caused by this PR but in the same gate's paperwork and cheap to fix here: `ROADMAP.md:45` still lists #2194 as open (it shipped in `8990c984`/PR #2196 — and this PR's own CHANGELOG cites it as the thing that unblocked the flip), and `RETRO.md:465-466` leaves both `- [ ]` deferred-findings boxes unchecked for #2136 and #2017 iteration 3.

#### 🟢 Correctly left alone — do **not** change

`ADR-026:68` (iteration-1 row), `ADR-026:200-232` (original Consequences plan, *"3. Flag flips ON after a soak"*), `ROADMAP.md:85-89/94/122-174/197+` (completed buckets), `CHANGELOG.md:33+` (`[1.1.0]`/`[1.1.0rc9]` tagged sections — immutable and untouched), `RETRO.md` v1.1.0-13 narrative, and `crates/djust_live/src/lib.rs:1443-1445`. The `virtual_keyed_ops_2017.rs` module docstring's *"the flag defaults OFF"* is still literally true of the Rust static that F1 deliberately keeps `false` — worth one clarifying sentence, but the assertion must stay.

---

### New checks

**🟢 Feature works end-to-end.** See F1. Not a no-op.

**🟢 No lost OFF-path coverage.** The Rust global is now **ON for the entire pytest session** (`demo_project.settings` doesn't pin it; confirmed `virtual_keyed_ops_enabled() → True` under it). I checked whether that silently flipped anything: `grep -rln "dj-virtual" tests/ python/tests/ python/djust/tests/` returns only **JS** files (vitest — unaffected by the Rust global) plus `test_virtual_keyed_config_2017.py`. **No Python VDOM-diff test exercises a `dj-virtual` container**, so nothing changed meaning silently. OFF-path coverage survives where it should: the Rust `with_flag_off_…` test (verified under `--exact`) and the now-genuine opt-out arm in the #2166 probe. This matches `apps.py:75-77`'s stated intent that the suite run the shipped configuration.

**🟢 CHANGELOG structure (canon #2028).** Single hunk `@@ -7,6 +7,10 @@` under `## [Unreleased]`; **0 removed lines** vs `main` and **0 vs the `v1.1.0rc9` tag**. No tagged section touched. (Internal contradiction is finding 1 above, a different axis.)

**🟢 Gates.**

| gate | result |
|---|---|
| `pytest tests/ python/tests/ python/djust/tests/ -n auto` | **10248 passed**, 219 skipped |
| `cargo test -p djust_vdom` | all pass (+ both dark tests under `--exact`) |
| `npm test` | **1933 passed** / 172 files |
| `make check-adr-status` | OK — but green on finding 3 |
| `scripts/check-doc-snippets.py` | OK |

**🟢 Checked and *not* a finding:** the guides' surviving *"Finalize-patch landing for an item scrolled OUT of the current window is still open"* is correct — `29-virtual-list.js:525` shows finalize-patch is the stray-sibling/streaming path, distinct from the `VirtualUpdate` content op #2146 shipped. No system check or README references the flag.

**🟡 One design nit:** `python/djust/apps.py:91` passes a second default literal — `_cfg.get("virtual_keyed_ops", False)` — while the comment three lines up (`:88-89`) says this *"keeps `_defaults` the single source of the default."* Unreachable today (the key is always in `_defaults`) and it fails safe, but it is a second place the default is written. Drop the fallback or make it `True`.

---

### Verdict

The engineering is right and I verified all of it independently: the static is safely dark, the flag genuinely reaches Rust, the probe genuinely gates, and all four suites are green. **F1 and F2 are closed** — no reservations.

F3 is the same shape as before: the fix was applied to the passages the previous review cited, not to the class. The sweep is now the fourth attempt and it under-counted again — this pass found **15 sites**, four of them inside files this PR edited, one inside the CHANGELOG entry this PR wrote, and one (`ROADMAP.md`) in a file the PR never opened despite it holding the iteration-3 gate row for the *current* bucket.

The pattern worth naming: every miss is a passage the author didn't *re-read*, only *searched near*. The reliable fix is to open each of the ~6 touched artifacts end-to-end and read every paragraph mentioning the flag — `diff.rs`, `config.py`, the ADR, the two test files, the demo view — plus grep `ROADMAP.md`/`RETRO.md`/`CHANGELOG.md` for the flip. All 13 are mechanical text edits; no code change, no re-run beyond the doc gates.

Requesting changes on F3 only.

REQUEST_CHANGES

johnrtipton:
## F3 addressed — fourth attempt, and the reason it kept failing

You were right, and the count matters less than the shape: these were not "missed files", they were **self-contradictions inside regions I had just edited**.

| | |
|---|---|
| `CHANGELOG.md:31` | "The default stays **OFF**" in the *same* `[Unreleased]` section as this PR's own "default ON" entry |
| `diff.rs:156` | "Default OFF." opening the doc comment whose body, twelve lines down, says "The DEFAULT is ON" |
| `ADR-026:3` | a Status line contradicting itself within one sentence |
| test docstring `:30` | "What keeps it OFF is #2185 … #2185 is the blocker" — in the docstring this PR edited |
| `ROADMAP:7`, `:37` | header and bucket both still "the flag ships wired but OFF" |

### Why it repeated four times

I kept **grepping for phrases and patching the hits**. That finds sites; it never asks whether a file agrees with *itself*. Every one of the four sweeps was the same method producing the same class of miss.

This pass enumerated every file mentioning the flag, read each region, and then ran a consistency check: flag any present-tense OFF claim in a file that also asserts ON. It reports **0**.

### Deliberately unchanged — history, not drift

CHANGELOG's iteration-1 entry ("dark by default"), ADR-026's iteration table row for #2126, `ROADMAP:167`'s description of what iteration 1 shipped, and the Rust assertion *"no Virtual\* op may be emitted while the flag is off"* — that last one is correct **precisely because** F1's revert kept the static `false`. Please sanity-check that reading.

### Also

Opened the v1.1.1-1 bucket row for #2017 as shipped — you were right that the bucket still listed it as deferred and no row was ever opened.

### Independent confirmation of the thing F1's revert put at risk

Your F1 fix could have made the feature a silent no-op (static `false`, nothing turning it on). I checked on the **shipped** configuration, no settings override:

```
python _defaults : True
config.get()     : True
rust enabled     : True     <- ready() applies it
```

and in a browser on that same build: insert at server position 5 → pool index **5** (`k3, k4, ins0, k5, k6`); remove at 3 → k3 gone, size 60, 0 duplicates; **0 console errors**.

### Verification

- Python **10249 passed** / 218 skipped
- JS **172/172 files, 1933 tests**
- `cargo test -p djust_vdom` 26 suites **and `flag_is_off_by_default` green in isolation**
- `check-adr-status` + `check-doc-snippets` clean; `[1.1.0rc9]` byte-identical to its tag

### Incidental

The `uv.lock` index guard fired again mid-run — `uv run maturin develop` re-leaked the localhost registry for the **second time today**. Re-locked; `uv.lock` is byte-identical to main. That guard has now caught the same trap twice in one day, which is a better argument for it than anything I wrote in its PR.

johnrtipton (commented):
## Final confirmation review — PR #2197 (ADR-026 iteration 3)

Reviewed at `006f7ba0` in an isolated worktree. Branch is **0 commits behind `origin/main`**, 4 ahead — the diff I reviewed is the program that will merge.

I deliberately did **not** re-run the author's phrase-grep method (it under-counted four times). Instead I enumerated every file mentioning `virtual_keyed_ops` / `VIRTUAL_KEYED_OPS` / `dj-virtual` / `Virtual{Insert,Move,Remove}` / `ADR-026` / `iteration 3`, then **read each file's relevant region and asked whether the file contradicts itself**, classifying every hit as history (correct) or present-tense drift.

---

### F1 — Rust static must stay `false` — ✅ ADDRESSED

- `crates/djust_vdom/src/diff.rs:203` — `static VIRTUAL_KEYED_OPS: AtomicBool = AtomicBool::new(false);` unchanged.
- Both dark-default tests pass **individually under `--exact`**:
  - `cargo test -p djust_vdom --test virtual_keyed_ops_2017 flag_is_off_by_default -- --exact` → ok
  - `... with_flag_off_a_virtual_parent_diffs_exactly_like_a_plain_one -- --exact` → ok
- The assertion **"no Virtual* op may be emitted while the flag is off"** (`crates/djust_vdom/tests/virtual_keyed_ops_2017.rs:128-130`) **is correct** given the static is `false`: Rust tests never run Python's `DjustConfig.ready()`, the process global therefore starts `false`, and `FlagGuard::drop` (`:39-43`) restores `false` even on panic. Verified in isolation, not just in-suite.
- `crates/djust_live/src/lib.rs:1442-1445` already states this correctly ("The Rust default stays OFF so a non-Django embedder of this crate is never silently changed; the Python config default is what carries iteration 3's flip"). No drift there.

### F2 — probe test went tautological — ✅ ADDRESSED (but see 🟡 N3, same class, same file)

`python/djust/tests/test_config_settings_recovery_2166.py:39` sets the probe to `False`, `:93` and `:99` pin `virtual_keyed_ops: False` / `rust_flag: False`. Discriminating **by construction** now (settings `False` vs default `True`). 12/12 pass in that pair of files.

### F3 — documentation sweep completeness — 🔴 **NOT ADDRESSED**

This is a **fifth** under-count, and both misses are in `docs/adr/026-dj-virtual-differ-awareness.md` — one of them **two lines below a line commit `006f7ba0` edited**, the other **six lines below another line the same commit edited**.

**🔴 1. `docs/adr/026-dj-virtual-differ-awareness.md:72-76` — the ADR contradicts its own Status line and its own table row.**

```
70 | 3. flag flips ON after a soak | **shipped, 1.1.1** — gate passed …   ← edited by this PR
71
72  The config half of iteration 3 is on `main` (unreleased at time of
73  writing): `virtual_keyed_ops` reaches the differ via `DjustConfig.ready()` …
75  process global, not per-view state). The DEFAULT half did not: it stays
76  `False`.
```

"The DEFAULT half did not: it stays `False`" is a flat present-tense assertion of the pre-flip state, sitting immediately under the row the author changed to "shipped, 1.1.1" and contradicting line 3's Status ("iteration 3 — the flag flipped ON by default — shipped in 1.1.1"). A reader who scrolls past the table gets the opposite answer from the same document.

**🔴 2. `docs/adr/026-dj-virtual-differ-awareness.md:114` — a live "do this later" instruction left in an edited paragraph.**

The author rewrote line 108 of this blockquote (`Iteration 3 is now blocked on #2185` → `Iteration 3 WAS blocked on #2185 (fixed in PR #2195) and #2194 (PR #2196)`), then left the closing sentence of the *same* paragraph six lines down:

```
114 | > recovery. Flip the default once #2185 is fixed.
```

The default is flipped and #2185 is fixed; this now instructs a reader to perform work that this PR is doing.

Both are the exact failure shape called out last round — a self-contradiction inside a region the author had just edited.

---

### New findings

**🟡 N1 — `crates/djust_vdom/tests/virtual_keyed_ops_2017.rs:104`** — assertion message reads `"ADR-026 iteration 1 ships dark; the client cannot apply these ops yet"`. The first clause is correct history; **"the client cannot apply these ops yet" has been false since iteration 2 shipped (PR #2135)** and is contradicted by this PR's own `diff.rs` rewrite. Notable because `diff.rs:200-202` cites this test *by name* (`"which is also what flag_is_off_by_default in tests/virtual_keyed_ops_2017.rs pins"`) — the file was open.

**🟡 N2 — `python/djust/apps.py:91` hardcodes a second, now-disagreeing default.**

```python
_rust.set_virtual_keyed_ops(bool(_cfg.get("virtual_keyed_ops", False)))
```

Four lines above it (`apps.py:87-89`) the comment says this wiring *"keeps `_defaults` the single source of the default (ADR-026 iteration 3 flips it there, **and this must follow**)"*. After the flip it no longer follows: `_defaults` says `True`, the fallback says `False`. Unreachable today (the key is present in `_defaults`, and `LiveViewConfig.get` at `config.py:463` only returns the fallback for a missing/`None` value) — but it is a fail-**OFF** fallback for the one config shape where everything else would report ON, which is the precise asymmetry class that comment exists to prevent. It also makes `diff.rs:192-194`'s description (`set_virtual_keyed_ops(config.get("virtual_keyed_ops"))` applied "unconditionally") inexact — there is both a `hasattr` guard and a literal fallback.

**🟡 N3 — `python/djust/tests/test_config_settings_recovery_2166.py:180-182` — F2's exact failure class recurs one screen below the F2 fix.**

```python
with override_settings(LIVEVIEW_CONFIG={"virtual_keyed_ops": True}):
    assert cfg.ensure_settings_loaded() is True, "must report it recovered"
assert cfg.get("virtual_keyed_ops") is True
```

The value assertion is now **vacuous**. Empirically confirmed: the test seeds `cfg._config = cfg._defaults.copy()`, and in that state `cfg.get("virtual_keyed_ops")` is **already `True`** before any recovery runs — so the assertion cannot distinguish "the reload applied the override" from "defaults kept". Pre-flip (default `False`) it discriminated. The `ensure_settings_loaded() is True` line still carries signal, so the test is not fully dead — but the assertion that proves the *value* landed no longer does. Same one-line remedy the author already applied to `EXPECTED`: override with `False` and assert `is False`. Worth noting that the 7-line comment at `:85-92` explains this exact hazard, for this exact key, immediately above the instance that was missed.

**🟢 N4 — `python/djust/tests/test_virtual_keyed_config_2017.py:22-24`, sentence broken by commit `006f7ba0`.** Replacing `That` with `The later one said that` and inserting a blank line left a dangling clause:

```
Both OFF-rationales above are superseded. The later one said that

"browser evidence covers VirtualInsert only" line was true when written; …
```

**🟢 N5 — `docs/adr/026-dj-virtual-differ-awareness.md:74`** cites `crates/djust_vdom/src/diff.rs:167` for `VIRTUAL_KEYED_OPS`; the static is at **line 203** (it was 187 on `main`). Pre-existing staleness, widened ~16 lines by this PR's comment growth. Worth fixing in the same pass as 🔴 1, since it is in the same sentence.

**🟢 N6 — `tests/js/virtual-keyed-ops-2017.test.js:8`** — "Those ops ship dark — the differ's flag defaults off — because nothing could apply them. This is that applier." Defensible as narration of iteration 1's setup, but present-tense.

**🟢 N7 — `examples/demo_project/djust_tests/views/virtual_keyed_demo.py:3-5`** — "ROADMAP.md **requires** real browser evidence **before flipping** `virtual_keyed_ops` **on**". The gate has passed; the sibling template and `test_index.py` card were both updated, this docstring was not.

**🟢 N8 — CHANGELOG `[Unreleased]` now carries two `### Changed` headings** (line 10 and line ~30). `main` already had a duplicate `### Fixed`, so this PR does not introduce the class — noting only so the release cut collapses them.

---

### Verified as correct history — do **not** change these

I checked each of the author's "left deliberately" judgements and they are right:

- `docs/adr/026:68` — `| 1. differ emits keyed splice ops, flag default OFF | shipped, PR #2126 |` — a record of what iteration 1 shipped. Correct.
- `docs/adr/026:139` (gating policy), `:196` (Option D verdict), `:199-229` (Why defer / Consequences) — decision-time record in an ADR. Correct.
- `ROADMAP.md:142-143`, `:151-153`, `:171`, `:301`, `:336`, `:353` — bucket-scoped notes ("#2017 does NOT close **here**", "Iteration 1 … default off"). Correct.
- `RETRO.md:465-466` — `### Open Items` inside the dated v1.1.0-13 milestone retro. Archive. Correct (and consistently left, since the sibling `#2136` row is equally stale).
- `crates/djust_vdom/tests/virtual_keyed_ops_2017.rs:8-11` — scoped to "This iteration" (the file is titled iteration 1), and the **Rust** flag genuinely does default off. Correct.
- `CHANGELOG.md:70`, `:72` — rc9 release entries. Correct.

---

### Also confirmed before merge

| Check | Result |
|---|---|
| **Not a no-op** | Real `django.setup()` in a subprocess with **no** `LIVEVIEW_CONFIG`: `config.get("virtual_keyed_ops") = True`, **`_rust.virtual_keyed_ops_enabled() = True`** ✅ |
| **Gate-off (#1468)** | Mutating `_defaults` back to `False` (mutation text asserted found, mutated source asserted different, `__pycache__` cleared) makes `test_the_config_default_is_on` fail with its intended message. The pin is load-bearing ✅ |
| Full Python suite | `tests/ python/tests/ python/djust/tests/ -n auto` → **10248 passed, 219 skipped, 0 failed** |
| `cargo test -p djust_vdom` | **231 passed, 0 failed** across 26 test binaries |
| `npm test` | **172 files / 1933 tests passed** |
| `make check-adr-status` | OK — 26 ADRs scanned, 0 status/version-line inconsistencies |
| `scripts/check-doc-snippets.py` | OK — snippets parse/resolve, version/size claims match |
| `uv.lock` | Only index is `https://pypi.org/simple` (sole distinct `/simple` URL); `uv.lock` is not in this diff at all ✅ |
| Branch staleness | `git rev-list --count HEAD..origin/main` = **0** ✅ |
| CHANGELOG canon #2028 | Both hunks land at lines 7–31; `## [1.1.0]` starts at line 33 — **no already-tagged section touched**, purely additive ✅ |

---

### Verdict

The **feature itself is sound and I found no correctness problem with it**: the flip works end-to-end, the Rust static correctly stays `false` so an unwired embedding fails safe, F1 and F2 have not regressed, every gate is green, and the CHANGELOG/branch hygiene is clean. If F3 were closed I would approve.

F3 is not closed. Two present-tense self-contradictions remain in `docs/adr/026-dj-virtual-differ-awareness.md` — the file's own Status line, its own iterations table, and its own body now give a reader opposite answers about whether the default is ON — and both misses are within a few lines of text this very commit edited. That is the same pattern for the fifth time, and the ADR is the canonical design record for this feature, so it is the worst single place to leave the contradiction.

**Not safe to merge as-is.** The blocking work is small and mechanical:

1. `docs/adr/026:72-76` — rewrite to state the default is ON since 1.1.1 (keep the `DjustConfig.ready()` mechanism sentence; fix the `diff.rs:167` → `diff.rs:203` citation while you are in it, per 🟢 N5).
2. `docs/adr/026:114` — drop or past-tense "Flip the default once #2185 is fixed."
3. 🟡 N1, N2, N3 should ride along — N3 in particular is F2 recurring in the file F2 was about, and its fix is one line.

REQUEST_CHANGES

johnrtipton (commented):
## Addendum to the final review — one more F3 miss, and a correction to my own classification

I ran a second, independent sweep with a different reviewer and a different lens, precisely because a single pass has now under-counted this sweep five times. It was worth doing: **the two passes found different things.** The second pass independently reproduced my 🔴 1 (`docs/adr/026:72-76`, same lines, same reasoning), missed my 🔴 2 (`docs/adr/026:114`), and found one I had misclassified.

### 🟡 N9 — `RETRO.md:360`, Action Tracker row **#318**, `Status = Open`

I had checked `RETRO.md:465-466` (the `### Open Items` list inside the *dated* v1.1.0-13 milestone retro) and correctly ruled it archive. I then generalised that to "RETRO.md is history" and stopped. That was wrong: `RETRO.md` also contains a **persistent, continuously-maintained Action Tracker table** (header at `RETRO.md:20` — `| # | Action | Source | GitHub | Status | Notes |`), and row #318 in it tracks **`#2017` itself**:

| # | GitHub | Status |
|---|---|---|
| 318 | `#2017` | **`Open`** |

Its Notes column ends:

> **v1.1.0-13 progress**: ADR-026 iteration 1 … and iteration 2 … both shipped. **Iteration 3 (flip the flag) is soak-gated per #1122 AND newly blocked by #2136** — content patches for surviving rows are still index-addressed and silently rewrite the wrong row after a scroll.

Both halves are stale. #2136 shipped in PR #2146 (`ROADMAP.md:98`), and #2017 is what this PR closes.

That this table is *maintained forward* rather than frozen is settled by its own neighbours — rows #320, #321 and #322 all carry `Closed` / resolution notes of the form "Resolved v1.1.0-6 (PR #2029)". So the correct treatment for #318 is the same one those rows got: flip `Status` to `Closed` and append the resolution (`Resolved v1.1.1 (PR #2197)`; #2136 cleared by PR #2146). Leaving it `Open` is not archival fidelity, it is a live tracker asserting the flip is still blocked.

### What this says about the sweep method

Neither pass alone was complete — I missed `RETRO.md:360`, the second pass missed `docs/adr/026:114`. The two misses have the same cause and it is not carelessness: **both were adjacent to text that had *already* been correctly updated**, which reads as "this region is done" and suppresses a re-read. `docs/adr/026:114` sits six lines under the `Iteration 3 WAS blocked…` line the author fixed; `RETRO.md:360` sits in a table whose *other* stale-looking entry (the dated `### Open Items` list) genuinely is archive.

So the durable fix for this PR is not "grep harder" — it is a **structural pin**, in the spirit of #1859/#1860 (an anti-drift artifact must be load-bearing or it is decorative). Two candidates worth a follow-up issue:

1. `make check-adr-status` currently audits Status/version-line *consistency* only — the ADR itself says so at `docs/adr/026:56-58`. Extend it to flag an ADR whose body contains a present-tense contradiction of its own Status line (e.g. an ADR marked shipped that still contains "not shipped" / "stays `False`" / "blocked on" / "flip … once"). This exact class has now bitten ADR-026 twice.
2. A check that an Action Tracker row whose `GitHub` column names a **closed** issue may not carry `Status = Open`.

Neither is in scope for this PR; both would retire the class rather than patch the fifth instance — which is what canon (#1646, #2142 on chain-shaped issues) asks for here. This feature's sweep is itself a textbook chain: five rounds, each round's fix surfacing the next instance, and twice the previous round's own edit created the next miss.

### Verdict unchanged

Everything in my main review stands, and none of the gate results change: full Python suite 10248 passed / 0 failed, `cargo test -p djust_vdom` 231 passed, `npm test` 1933 passed, `check-adr-status` and `check-doc-snippets` clean, `uv.lock` PyPI-only, branch 0 behind `main`, CHANGELOG additive with tagged sections untouched, feature empirically **not** a no-op (`_rust.virtual_keyed_ops_enabled() = True` under a real `django.setup()` with no override), and the new pin is gate-off-proven load-bearing.

The blocking set is now three, all mechanical:

1. 🔴 `docs/adr/026-dj-virtual-differ-awareness.md:72-76` — "The DEFAULT half did not: it stays `False`."
2. 🔴 `docs/adr/026-dj-virtual-differ-awareness.md:114` — "Flip the default once #2185 is fixed."
3. 🟡 `RETRO.md:360` — Action Tracker row #318 still `Open`, still citing cleared blockers.

Plus 🟡 N1 / N2 / N3 from the main review, of which **N3 is the one I would most want fixed alongside** — it is F2 recurring, in the same file F2 was about, one screen below the comment that explains the hazard.

REQUEST_CHANGES

johnrtipton:
## All six closed — and the fifth under-count got a structural fix, not a sixth grep

Your meta-point was the actual finding: **two independent passes found two different subsets.** That is the tell that the *method* was wrong, not the effort. Every one of the five sweeps grepped for phrases and patched the hits, which finds sites but never asks whether a file agrees with itself.

### The six

| | |
|---|---|
| 🔴 `ADR:72-76` | "The DEFAULT half did not: it stays `False`" — two lines under the row the last commit changed to "shipped, 1.1.1" |
| 🔴 `ADR:114` | "Flip the default once #2185 is fixed" — six lines under a line the last commit edited *in the same paragraph* |
| 🟡 `RETRO:360` | Action Tracker row 318 → `Closed`, with why it outlived its cited blockers |
| 🟡 `rs:104` | message said "the client cannot apply these ops yet" — false since iteration 2. Note the **assertion** was right (static is `false`); only its stated reason was stale |
| 🟡 `apps.py:91` | `_cfg.get("virtual_keyed_ops", False)` — a second hardcoded default contradicting the comment directly above it. Removed. That is precisely how a default drifts out of sync with itself |
| 🟡 recovery test | **F2 recurring**, as you said — overrode to `True` and asserted `True`, vacuous under the new default. Now `False`/`False`, and gate-off confirms it fails when `ensure_settings_loaded()` is neutered |

### The structural check — including the version I threw away

`scripts/check-flag-default-consistency.py`.

My **first** version tried to detect prose self-contradiction. It did not work: an empirical canary replaying the real `006f7ba0` contradiction **did not fire**, because the ON claim lives in a sentence that never names the flag. I could have tuned regexes until the canary passed — and would have shipped a guard trusted more than it deserved, which is the decorative-pin failure of #1859/#2135. I threw it away instead.

What ships checks a **structural** fact with no language guessing: read the real default out of `config.py`, require every registered doc site to state that same value. Three canaries, all verified firing:

```
flip config.py default   -> all 3 sites report disagreement
revert ONE doc site      -> that site reported (the real #2017 drift shape)
DELETE a doc statement   -> "no statement of" error, NOT a silent pass
```

That last one is the property whose absence made v1 worthless — a checker that quietly finds nothing looks identical to one that found nothing wrong. It has its own test, plus one asserting no registered pattern is dead weight.

Registered in pre-commit. 7 cases in `tests/test_check_flag_default_consistency.py` — whose harness had the same bug in miniature: it ran the *repo's* script against a sandbox cwd, so every mutation test was vacuous until I fixed it to run the sandbox's own copy.

### Verification

- Python **10256 passed** / 218 skipped
- JS **172/172 files, 1933 tests**
- `cargo test -p djust_vdom` 26 suites **and `flag_is_off_by_default` green in isolation**
- `check-adr-status`, `check-doc-snippets`, `check-lockfile-registry`, and the new checker: all clean

### Also filed

**#2200** — the audit found **121** unchecked RETRO Open Items whose issues are all closed, including `RETRO.md:533`: *"Close tracker rows automatically when their GitHub issue closes (#2143)"*, itself unchecked and itself closed. Left out of this PR per #1079.

johnrtipton:
## Retrospective — PR #2197 (ADR-026 iteration 3)

**Task**: #2017 — flip `virtual_keyed_ops` ON by default. Merged as `025a5a85`.

### Quality: 3

Three adversarial reviews, each of which found real defects: **3 → 1 → 6**. The count went *up* on the third pass. That is the honest headline.

### What went well

- **The gate was re-run rather than argued about.** The ROADMAP asked for browser evidence; the previous answer was "the evidence doesn't support it". Re-deriving it found the evidence had never been obtainable, because the control arm was as broken as the test arm.
- **Two prerequisites were identified as prerequisites**, not worked around. #2185 and #2194 shipped as their own PRs before this one, and the A/B only became valid afterwards.
- **The verdict rests on a real comparison**: same healthy list, same start state, flag the only variable. Insert at position 5 → index **60 (tail)** off vs **5** on; removal doesn't land off vs drops the right key on; reverse ignored off vs exact on.
- **Reviewers were given the sharpest target available.** I flagged the Rust static flip as "most likely to be a real finding" in the first review brief. It was, and it was a false green I had reported as passing.

### What didn't

- **Five under-counted documentation sweeps.** 2-of-5, then 5-of-6, then 6 more, then 15, then 6. Every pass used the same method — grep for phrases, patch the hits — and every pass missed a different subset. The third reviewer's observation that *two independent greps found two different subsets* is what finally identified the method as the problem rather than the effort.
- **The worst misses were self-contradictions inside regions I had just edited**: `CHANGELOG.md` asserting "the default stays OFF" in the same `[Unreleased]` section as its own "default ON" entry; the ADR contradicting itself two lines under a row I had changed in that same commit.
- **I reported a false green.** "Rust 44 suites green" concealed `flag_is_off_by_default` failing under `--exact` — it passed only because `FlagGuard::drop` resets the flag and it sorts late. A whole-suite pass is not evidence about a specific test.
- **I made an existing test tautological, twice.** The probe in `test_config_settings_recovery_2166.py` requires values be the opposite of each default; flipping the default silently made two keys stop discriminating, and the same class recurred in a second assertion in the same file.
- **My first structural check failed its own canary** and I nearly shipped it. Replaying the real contradiction did not fire it. Tuning regexes until it passed would have produced a guard trusted more than it deserved — the decorative-pin failure this repo has canon about.

### The durable change

`scripts/check-flag-default-consistency.py` reads the real default from `config.py` and requires every registered doc site to state that value. No language guessing. Three canaries verified firing, including the one that killed v1: **deleting a doc statement is a hard error, not a silent pass.** Registered in pre-commit, 7 tests.

Whether it generalises is unproven — it currently tracks one flag. Its value is that adding a flag means editing a visible registry, rather than trusting the next person's grep.

### Verified on `main` post-merge

Python **10256 passed** / 218 skipped · JS **172/172 files, 1933 tests** · `flag_is_off_by_default` green in isolation · `check-adr-status`, `check-doc-snippets`, `check-lockfile-registry`, `check-flag-default-consistency` all clean · `_rust.virtual_keyed_ops_enabled()` **True** with no override, confirming the feature is live and not a no-op.

### Deferred

- **#2200** — 121 unchecked RETRO Open Items whose issues are all closed, including the row about automating that.
- Out-of-window finalize-patch landing (item 3 of #2017's original scope) remains open.

Issue #2203, opened 2026-08-23
bug: Display for Value diverges from Django for 5 of 7 variants (bool, None, list, dict, float)


johnrtipton:
Split out of #2202 to keep that fix scoped (#1079).

#2202 fixed built-in filters ignoring a **bare-identifier** argument. While
isolating its scope, a control run passed each arg-taking built-in a **literal**
argument — the case #2202 does not touch. Four still diverge from Django, so
these are separate pre-existing bugs:

| filter | template | Django | djust |
|---|---|---|---|
| `divisibleby` | `{{ a\|divisibleby:5 }}` (a=10) | `True` | `true` |
| `slice` | `{{ a\|slice:":2" }}` (a=[1,2,3,4]) | `[1, 2]` | `[List]` |
| `truncatewords` | `{{ a\|truncatewords:2 }}` | `one two …` | `one two...` |
| `truncatechars` | `{{ a\|truncatechars:5 }}` | `abcd…` | `ab...` |

Two distinct causes, worth separating before fixing:

1. **`divisibleby` / `slice` are `Value`-rendering issues, not filter issues.**
   `impl Display for Value` (`crates/djust_core/src/lib.rs:175-190`) renders
   `Bool` via Rust's `{b}` (lowercase) and `List` as the placeholder `[List]`.
   Any filter returning a bool or a list has the same output, so fixing it
   inside these two filters would be the parallel-path mistake #2202 was.
   Check what else returns `Value::Bool` / `Value::List` before choosing where
   the fix goes — changing `Display` is broad and needs its own blast-radius
   review.

2. **`truncatewords` / `truncatechars` use `...` where Django uses `…`** (U+2026,
   a single character). Django's `Truncator` appends `…`. This one is narrow
   and self-contained, but it changes rendered output for every existing
   template that truncates, so it deserves a CHANGELOG note rather than being
   slipped in.

Reproducer (each uses a literal arg, so #2202's fix is not involved):

```python
import django
from django.conf import settings
settings.configure(TEMPLATES=[{
    "BACKEND": "django.template.backends.django.DjangoTemplates",
    "DIRS": [], "APP_DIRS": False, "OPTIONS": {},
}])
django.setup()
from django.template import Template, Context
import djust._rust as r

for tpl, ctx in [
    ("{{ a|divisibleby:5 }}",   {"a": 10}),
    ('{{ a|slice:":2" }}',      {"a": [1, 2, 3, 4]}),
[...]

PR #2204, opened 2026-08-23
fix(templates): resolve bare-identifier built-in filter args as context variables (#2202)

### LINKED ISSUE #2202: bug: built-in template filters don't resolve a bare-identifier argument as a context variable (Django parity)
## Summary

Built-in template filters use their argument as a **raw string** and never resolve it against the context. Django resolves a bare-identifier filter argument as a variable (`Variable(arg).resolve(context)`), falling back to a literal only when the argument is quoted.

So `{{ value|default:fallback }}` renders the literal text `fallback` instead of the value of `fallback`.

Quoted-literal arguments are correct. **Custom** filters are also correct — they already resolve bare identifiers. The gap is built-ins only, which makes this a parallel-path drift (#1646) on the filter-argument axis, in the same Django-parity family as ADR-024 (which fixed *callable* resolution in the sidecar walk but not *filter-argument* resolution).

## Impact — found in production on djust.org

`marketing/templates/marketing/blog/post.html` renders:

```html
<img src="{{ post.featured_image_url }}"
     alt="{{ post.featured_image_alt|default:post.title }}">
```

Every blog post with an empty `featured_image_alt` currently ships `alt="post.title"` — a literal, meaningless alt attribute. That is an accessibility defect on a live site, and it fails silently: the page renders, nothing raises, and the output looks plausible unless you read the HTML.

The silent-wrong-output shape is the same one ADR-024 was written about.

## Reproducer

Runs against the real Rust render path. Copy-paste:

```python
import django
}])
django.setup()

import djust._rust as r

tpl, ctx = "{{ e|default:v }}", {"e": "", "v": "FROM CONTEXT"}
print("Django:", repr(Template(tpl).render(Context(ctx))))  # 'FROM CONTEXT'
print("djust: ", repr(r.render_template(tpl, ctx)))          # 'v'   <-- bug
```

## Scope

Nine built-ins where a **literal** argument matches Django exactly, so the bare-identifier failure isolates cleanly to argument resolution:

|---|---|---|---|
| `default` | `{{ e\|default:v }}` | `'X'` | `'v'` |
| `default_if_none` | `{{ e\|default_if_none:v }}` | `'X'` | `'v'` |
| `add` | `{{ a\|add:v }}` | `'12'` | `'5'` |
| `join` | `{{ a\|join:v }}` | `'p-q'` | `'pvq'` |
| `cut` | `{{ a\|cut:v }}` | `'heo'` | `'hello'` |
| `yesno` | `{{ a\|yesno:v }}` | `'Y'` | `'v'` |
| `floatformat` | `{{ a\|floatformat:v }}` | `'3.14'` | `'3.1'` |
| `pluralize` | `{{ a\|pluralize:v }}` | `'es'` | `'v'` |
| `stringformat` | `{{ a\|stringformat:v }}` | `'042'` | `'42'` |

`join` and `cut` are the sharpest: they don't just ignore the argument, they splice the *identifier text* into the output (`'pvq'`) or silently no-op.

**Not part of this issue** — these diverge from Django even with a literal argument, so they are separate pre-existing bugs and should not be folded into the fix: `divisibleby` (`'True'` vs `'true'`), `slice` (`'[1, 2]'` vs `'[List]'`), `truncatewords` / `truncatechars` (ellipsis `…` vs `...`).

## Root cause

`crates/djust_templates/src/filters.rs:110-117` — the `default` arm, representative of all nine:

```rust
"default" => {
    if value.is_truthy() {
        Ok(value.clone())
    } else {
        Ok(Value::String(arg.unwrap_or("").to_string()))  // raw string, never resolved
    }
}
```

`apply_builtin_filter` (`filters.rs:92`) **already receives `context`** — it is passed at `filters.rs:64` — but no arm consults it for the argument. It does *not* receive `arg_was_quoted`, which is the missing piece needed to tell a quoted literal from a bare identifier.

## The fix already exists in-repo

`apply_custom_filter` (`crates/djust_templates/src/filter_registry.rs:216`) solves exactly this, at `filter_registry.rs:259`:

```rust
Some(s) if arg_was_quoted => { /* quoted literal -> plain string */ }
Some(s) => {
    // Bare identifier — try context resolution first.
    if let Some(ctx) = context {
        if let Some(resolved) = ctx.resolve(s)? { ... }
    }
    // falls back to the raw string when not found
}
```

Suggested shape, lifting that reference implementation rather than designing a new one (#1077):

1. Thread `arg_was_quoted` into `apply_builtin_filter` alongside the `context` it already has. The renderer computes it at `renderer.rs:491` and `renderer.rs:555`.
2. Resolve the argument **once**, before the dispatch table, so all arms share one chokepoint — do not patch nine arms independently (#1646).
3. Preserve the existing fallback: unresolvable bare identifier → raw string, matching `filter_registry.rs` and keeping current behaviour for templates that pass bare words to filters like `pluralize:es`.

Point 3 matters for back-compat: `{{ n|pluralize:es }}` currently "works" by accident because the unresolved identifier `es` is used verbatim. Preserving the not-found fallback keeps those templates rendering as they do today.

## Suggested tests

- Parity test per affected filter: bare identifier resolves; quoted literal stays literal; unresolvable identifier falls back to raw string.
- Gate-off (#1468): revert the resolution and confirm at least the `default` and `join` cases go red.
- The `join`/`cut` cases are the strongest regression guards, since their pre-fix output contains the identifier text itself.

johnrtipton:
Closes #2202.

## The bug

Django resolves a filter argument as a **variable** unless it is quoted:
`{{ x|default:fallback }}` looks up `fallback`, and only
`{{ x|default:"fallback" }}` is the literal string. djust's built-in filters
used the raw argument text and never consulted the context.

It fails **silently** — the template renders, nothing raises, and the output
looks plausible. Found on djust.org, where
`{{ post.featured_image_alt|default:post.title }}` had been shipping
`alt="post.title"` on every post with an empty alt for months. A broken alt
attribute is invisible unless you read the HTML.

Ten built-ins were affected. `join` and `cut` are the sharpest — they did not
merely ignore the argument, they spliced the *identifier's own text* into the
output:

| filter | template | before | after (= Django) |
|---|---|---|---|
| `default` | `{{ e\|default:v }}` | `v` | `X` |
| `default_if_none` | `{{ e\|default_if_none:v }}` | `v` | `X` |
| `add` | `{{ a\|add:v }}` | `5` | `12` |
| `join` | `{{ a\|join:v }}` | `pvq` | `p-q` |
| `cut` | `{{ a\|cut:v }}` | `hello` | `heo` |
| `yesno` | `{{ a\|yesno:v }}` | `v` | `Y` |
| `floatformat` | `{{ a\|floatformat:v }}` | `1.2` | `1.23` |
| `pluralize` | `{{ a\|pluralize:v }}` | `v` | `es` |
| `stringformat` | `{{ a\|stringformat:v }}` | `42` | `042` |
| `date` | `{{ d\|date:v }}` | `v` | `2026` |

## Root cause and fix shape

**Custom filters were already correct** — `filter_registry.rs` resolves bare
identifiers via `Context::resolve` and has since #1121. So this is **#1646
parallel-path drift on the filter-argument axis**: two implementations of
"resolve a filter argument", one right and one wrong.

The fix routes built-ins through the same resolution instead of adding a second
correct copy. Resolved **once** in `apply_filter_full_safe` ahead of the
dispatch table — not in each of the ten arms, which would have been ten more
places for the next filter to drift from. `apply_builtin_filter` already
received `context` and simply never consulted it for the argument.

Production diff is 33 lines in one file.

## Deliberately preserved

- **Quoted args are never looked up**, gated on the `arg_was_quoted` hint the
  renderer already computes — a literal cannot become a lookup because a
  context key happens to share its name.
- **The classic `apply_filter_with_context` call site is untouched** (it passes
  `arg_was_quoted=true`); only the renderer path changes.
- **An unresolvable identifier still falls back to its raw text.** This is a
  **deliberate divergence from Django**, which raises `VariableDoesNotExist`.
  `{{ n|pluralize:es }}` works today only by that accident, and raising would
  convert a silent wrong-output bug into a site-wide 500 on upgrade.

## Verification

- **Reproducer-first** (#1210): 13 Rust cases written before the fix,
  confirmed **10 red / 3 green**. The 3 passing pre-fix are the guards — they
  encode behaviour the fix must preserve.
- **Gate-off** (#1468): neutering the resolution returns exactly 10 red / 3
  green. The harness asserts the mutation text was found and that the source
  actually changed, and refuses to report a number if the build broke
  (#2129/#2135).
- **Full suite**: Rust 13/13 suites; Python **10,258 passed, 218 skipped, 0
  failed** across all three test roots.
- **Security**: none of the ten filters are in `safe_output_filters`, and HTML
  payloads reaching output through a resolved argument are verified escaped.

## Scope corrected twice during review

Both from the disconfirming-citation check (#1386), both upward in rigour:

1. **`date` is a tenth affected filter**, not out of scope. The original probe
   fed it a *string*, which Django cannot format either, so the divergence
   looked pre-existing. With a real `datetime` it is fixed by this change —
   gate-off confirms (`'2026'` on, `'v'` off).
2. **The raw-text fallback was initially described as matching Django.** It
   does the opposite; Django raises. Corrected in the test comment, commit
   message and CHANGELOG.

## Out of scope

`divisibleby`, `slice`, `truncatewords`, `truncatechars` diverge from Django
**even with a literal argument** — separate pre-existing bugs, filed as
**#2203** (#1079). Two of them are really `Value`-rendering issues
(`Display for Value` emits `true` and `[List]`), which would be the same
parallel-path mistake if patched inside the filters.

## Test layout

| file | cases |
|---|---|
| `crates/djust_templates/tests/test_builtin_filter_arg_resolution_2202.rs` | 13 (9 filters + 4 guards) |
| `TestBuiltinFilterArgResolution2202` in `python/tests/test_template_filters.py` | 2 (`date`) |

`date` is Python-side because `Value` has no date variant and an ISO string
passes through the filter unformatted — a Rust-level case would assert nothing
about argument resolution.


johnrtipton:

**Verdict: REQUEST_CHANGES** → addressed in `f43ac31f` + `08b5c326`.

Branch freshness (#250): 0 commits behind `origin/main`. Gate-off run with the #2129/#2135 harness assertions (mutation text found, source changed, build verified) — **10 red / 3 green**, valid. Each mechanism independently reachable: dropping the quoted gate reddens only the quoted-literal guard; dropping `.or(arg)` reddens only the miss guards.

### 🔴 MUST-FIX — `.ok()` swallowed the ADR-024 exception, creating new #1646 drift 20 lines from the code it converges with

`filters.rs` resolved with `.ok().flatten()`. `Context::resolve` returns `Err` for exactly one thing (`context.rs:282-284`): an exception raised **inside a method auto-called during resolution**. Lookup misses are `Ok(None)`, so `.flatten()` handled those — `.ok()` discarded **only** the user's exception.

Three sibling paths propagate it: `filter_registry.rs:259-266` (custom filters, with a comment saying exactly this), `renderer.rs:474` (main variable path), and Django itself.

Failure: `{{ post.summary|default:post.get_fallback_title }}` where the method raises renders the literal text `post.get_fallback_title` into the page — the silent-wrong-output failure this PR exists to remove, reintroduced on the error branch.

The justifying comment was also false on two counts: it called an unresolvable argument an *error* (it is `Ok(None)`), and said Django does not raise for one (it raises `VariableDoesNotExist`). The test doc-comment said the opposite, so the PR contradicted itself — the #1867 class.

**Fixed** in `f43ac31f`: `ctx.resolve(a)?`, comment rewritten to distinguish the two outcomes. The deliberate raw-text fallback for a miss is untouched. 2 regression cases in `TestFilterArgErrorPolicy2202`; gate-off reverting `?` → `.ok()` fails the propagation test with `DID NOT RAISE` while the miss test correctly stays green.

### 🟡 `add` claimed fixed, but only for integer-valued arguments

`filters.rs:204` parses `i64` and falls back to **0**, so `{{ a|add:f }}` with `f=1.5` gives `5` where Django gives `6`.

The **literal-argument control** — the same control that classifies a divergence as pre-existing — shows `{{ a|add:1.5 }}` fails too. So this is a pre-existing bug, not an argument-resolution one. Moved to **#2203**; the CHANGELOG now says "integer-valued arguments only".

### 🟡 Enumeration undercounted the surface; `time` needed the control, not inclusion

The fix lands once before the dispatch table, so it applies to **all 26** arg-taking built-ins, not the ten claimed. CHANGELOG now reads "applies to all 26; ten verified fixed".

On `time` specifically — the review proposed adding a case here, since it is `date`'s structural twin and `date` was promoted into the ten during review. **The literal control refuted that**: `{{ d|time:"H:i" }}` renders `2026-08-22 14:30:00` where Django gives `14:30`, so `time` ignores its format string regardless of argument resolution. Adding it would have made this PR claim a fix it does not deliver. Moved to **#2203**.

That control has now corrected the scope in both directions — pulling `date` **in** (the original probe fed it a string, which Django cannot format either) and keeping `time` and `add`-float **out**.

### 🟢 Nits

- Stale "nine built-ins" in the rationale comment — now "twenty-six".
- `Value::Null` as an argument renders `""` where Django renders `"None"` — same `Display for Value` bucket as `true`/`[List]`, appended to #2203, which is now three cases and reinforces that they are one `Value`-rendering decision rather than per-filter patches.
- `renderer.rs:2073-2077` inline-strips quotes rather than calling `parser::strip_filter_arg_quotes` — a fourth copy of the same two-line rule, in a PR about parallel-path drift. Left alone: currently equivalent (guarded by the same `is_quoted_arg`) and out of scope.

### Checked and clean

Quoting gate accurate at **all three** call sites (`renderer.rs:493`, `:557`, `:2087` — the brief said two); `is_quoted_arg` and `strip_filter_arg_quotes` use byte-identical predicates, so a stripped literal can never be mistaken for an identifier. No XSS: `json_script` is the one new surface (it *is* in `SAFE_OUTPUT_FILTERS`, and its id can now come from context) — probed with `x" onload="y`, escaped correctly. No panics across hostile inputs to `slice`, `stringformat`, `floatformat`, `date`, `time`, `ljust`, `get_digit`. No regex applied to the argument, so no ReDoS. `arg.is_none()` branches (`date`/`time` format defaults) untouched, since `.or(arg)` cannot turn `Some` into `None`.

**Performance**: two heap allocations per filter-with-arg per render (`key.split('.').collect()`), but the GIL is engaged only when the root segment hits `raw_py_objects` — numeric and non-context args return at `context.rs:290` before `Python::attach`. When it does resolve, cost roughly doubles for that expression, paid eagerly even if the filter ignores the argument. Django resolves eagerly too, so this is parity; noted as a real per-event cost on list-heavy templates.

### Post-fix

Rust 13/13 suites; Python **10,260 passed, 218 skipped, 0 failed** across all three roots.

johnrtipton:
## Stage 14 — Retrospective

**Quality: 4/5.** The diagnosis and the structural shape were right from the start; the fix shipped a defect of the same class it was curing, and review caught it.

### What went well

**The literal-argument control was the load-bearing tool, and it corrected scope in both directions.** Running each filter with a *literal* argument isolates "this filter is broken generally" from "this filter's argument isn't resolved". It pulled `date` **in** (the first probe fed it a string, which Django can't format either, so a real bug looked pre-existing) and kept `time` and `add`-with-a-float **out** — both fail with literal arguments too. That second direction matters most: review proposed adding a `time` case here on the reasonable grounds that it is `date`'s structural twin and `date` had just been promoted in. Without the control I would have taken that suggestion and made the PR claim a fix it does not deliver.

**Reproducer-first (#1210) paid immediately.** 13 cases written before the fix, 10 red / 3 green — and the 3 green were the guards, which is the shape you want: they encode what must *not* change, so passing pre-fix is correct rather than suspicious.

**The one-chokepoint shape held up under adversarial review.** Resolving once ahead of the dispatch table rather than in each arm was the right call, and it is why the `time`/`add` findings are scope questions rather than ten separate patches to redo.

### What didn't

**I introduced a fresh instance of #1646 inside the fix for #1646.** `.ok().flatten()` swallowed the one thing `Context::resolve` returns `Err` for — an exception from an ADR-024 auto-called method — while the custom-filter path twenty lines away propagates it. I converged the *resolver* and left the *error policy* divergent. The lesson is narrower than "be careful": when converging two implementations, enumerate every axis on which they can differ (inputs, outputs, **error handling**, side effects) and check each. I checked the first two.

**I wrote a rationale comment asserting Django's behaviour before verifying it, then verified it and only half-corrected.** The comment claimed "an unresolvable filter argument is not a render error in Django either." Django raises `VariableDoesNotExist`. I later ran that exact probe, found the truth, updated the *test* doc-comment — and left the source comment contradicting it. Two artifacts in one PR disagreeing about the same fact. This is #1867 (a prose invariant that survives citation-checking while being false about the code it cites), and the specific failure mode is *partial* propagation of a correction: when a verified fact contradicts something I wrote earlier, grep for every place I asserted it, not just the one in front of me.

**I stated the affected-filter count from what I had tested rather than from the surface.** The fix lands before the dispatch table, so it applies to all 26 arg-taking built-ins; I said ten because ten is what I had test cases for. Tested-count and affected-count are different numbers and both belong in the entry.

### Relation to earlier work in this cycle

This is the third consecutive scope undercount in recent memory — the #2017 documentation sweeps under-counted five times, the 1.1.0 blog post under-counted twice (ADRs, then defaults), and now this. The pattern is identical each time: **I enumerate what I touched instead of enumerating the surface, then state the first number as if it were the second.** The durable fix isn't diligence, it's asking "what is the denominator?" before quoting a numerator — here, `grep` the dispatch table for arms referencing `arg` (26) before writing "ten built-ins were affected".

It also re-proves the v1.0.0rc4 canon that a coverage suite must enumerate every variant of the surface it covers. Ten tested arms out of 26 changed arms is exactly the single-variant-coverage shape that makes a bug class look handled.

### Verified

- Reproducer-first: 10 red / 3 green before the fix.
- Gate-off with #2129/#2135 harness assertions (mutation found, source changed, build verified): 10 red / 3 green. Per-mechanism reachability confirmed separately for all three mechanisms — dropping the quoted gate, the raw-text fallback, and the error propagation each redden a different, specific test.
- Rust 13/13 suites; Python **10,260 passed, 218 skipped, 0 failed** across all three roots.
- CI: 15/15 required checks green.
- Follow-ups filed: **#2203** (7 pre-existing divergences, three of them one `Value`-rendering decision).

`RETRO_COMPLETE`

PR #2208, opened 2026-08-23
fix(templates): date/time on datetimes, Django's ellipsis, and add's real semantics (#2203)

### LINKED ISSUE #2203: bug: Display for Value diverges from Django for 5 of 7 variants (bool, None, list, dict, float)

johnrtipton:
Closes #2203 in part. Three Django-parity gaps that persist with a **literal** argument, so unrelated to #2202's bare-identifier resolution.

## The headline is not what the issue was filed about

#2203 says "four filters diverge". Investigating the `time` case found something larger: **`date` and `time` are both broken for any Python `datetime`.**

`format_date` accepted RFC3339 or a bare `%Y-%m-%d`, and `format_time` delegates straight to it. A `datetime` arrives as `"2026-08-22 14:30:00"` — space-separated, no offset — matching neither, so the parse failed and the filter returned its input **verbatim**:

| template | Django | before |
|---|---|---|
| `{{ dt\|date:"Y-m-d" }}` | `2026-08-22` | `2026-08-22 14:30:00` |
| `{{ dt\|time:"H:i" }}` | `14:30` | `2026-08-22 14:30:00` |

`{{ post.created_at|date:"Y-m-d" }}` is the commonest use of this filter. It survived because a `DateField` stringifies to `2026-08-22` and takes the date-only branch — the failure is invisible unless the value is a `datetime`.

Fixed at the one shared parse chokepoint rather than in two filters (#1646). Fail-soft on unparseable input is preserved and pinned.

## The other two

**`truncatewords` / `truncatechars`** used `...` where Django uses `…` (U+2026). Not cosmetic for `truncatechars`: Django reserves **one** character for the ellipsis inside the limit, so `truncatechars:5` is `abcd…` where reserving three gave `ab...`.

**`add`** implemented only a partial first branch of Django's three (`int(value) + int(arg)` → `value + arg` → `""`). It parsed the argument as `i64` and **defaulted to 0** on failure:

| template | Django | before |
|---|---|---|
| `{{ 5\|add:1.5 }}` | `6` | `5` |
| `{{ "a"\|add:"b" }}` | `ab` | `a` |
| `{{ "4"\|add:"3" }}` | `7` | `4` |

Branch order is load-bearing — int first, so `"4"+"3"` is `7`, not `"43"`.

Two divergences from Django are deliberate and documented: its third branch returns `""` where djust returns the value unchanged (turning a rendered value into silent emptiness on upgrade is the silent-wrong-output class this engine keeps fixing), and overflow returns the value unchanged rather than wrapping.

## Self-review caught two defects in my own first pass

Both found by disconfirming my own comments (#1386), not by a failing test:

**1. The no-seconds fallback was justified by a false claim.** The comment said "datetime omits seconds when they are zero … and a TimeField does the same". Python does neither — `str(datetime(...,14,30))` is `"2026-08-22 14:30:00"`, `str(time(14,30))` is `"14:30:00"`. So it accepted `"2026-08-22 14:30"` (a shape that never occurs) while leaving unparsed the one that does: an HTML `<input type="datetime-local">` submits `YYYY-MM-DDTHH:MM`. All four separator/seconds combinations are now accepted, and a timezone-aware datetime (`USE_TZ=True`, Django's default) is pinned — it worked already, but nothing covered the space-plus-offset shape.

**2. `add` could overflow, and I widened the surface.** Python's ints are arbitrary-precision so Django cannot overflow; `i64` can, and plain `+` **panics in a debug build** ("attempt to add with overflow") while silently **wrapping** in release — `{{ max|add:1 }}` returned a *negative* number. Coercing floats and numeric strings is what made a plain string or float reach it. Now `checked_add`.

## Deliberately out of scope

`divisibleby` (`true`/`True`), `slice` (`[List]`), and a `Null` argument (`""`/`None`) are **not filter bugs**. They are `impl Display for Value`, which governs every `{{ var }}` in every template — **5 of 8 bare-variable types diverge**, including `dict` (`[Object]` vs `{'a': 1}`), which #2203 doesn't even mention.

And it carries a concrete back-compat hazard:

```
var flag = {{ v }};    djust today -> "var flag = true;"   valid JS
                       Django      -> "var flag = True;"   ReferenceError
```

Matching Django there would break every template embedding a bool in a script block. That's a design decision needing its own discussion, not something to smuggle in behind a filter fix (#1079) — and fixing `divisibleby`/`slice` inside their own arms would be the exact parallel-path mistake #2202 was about. #2203 stays open for it.

## Verification

- **Reproducer-first** (#1210): 13 cases written before the fix, **8 red / 5 green**; the 5 passing are guards.
- **Gate-off** (#1468) per mechanism: removing the datetime parse, the ellipsis, `add`'s int coercion, or `checked_add` each reddens a **distinct** set — the last by reproducing the original overflow panic.
- **Full suite**: Rust 14/14 suites; Python **10,276 passed, 218 skipped, 0 failed** across all three roots.
- **Robustness probe**: NaN, `+inf`, `i64::MAX`, a 23-digit string, `truncatechars:0`/`:1`, and multibyte input — no panics; truncation is char-based so no UTF-8 boundary split.
- Three legacy tests pinned the old `...` output. Updated to values taken from **Django itself**, not from this implementation.


johnrtipton:

**Verdict: REQUEST_CHANGES** → addressed in `5a894954` + `40bd9bd0`. Every finding independently verified before acting.

Branch freshness (#1450): 0 behind `origin/main`. Gate-off run with the #2129/#2135 harness assertions across nine mutations.

### 🔴 1 — The headline fix missed the shape production actually emits

Two coupled defects, and the premise was wrong in **four** places (code comment, CHANGELOG, PR body, test docstring).

djust serializes datetimes with **`.isoformat()`** (`python/djust/serialization.py:311`), not `str()` — so the arriving shape is `T`-separated, and the two space-separated formats were the ones that never occur on this path. Worse: **neither form omits microseconds**, and none of the four formats had a fractional-seconds directive. chrono rejects trailing input, so:

| input | before |
|---|---|
| `2026-08-22T14:30:00` | ✅ |
| **`2026-08-22T14:30:00.123456`** | ❌ returned verbatim |

`datetime.now().isoformat()` is `'2026-08-23T01:31:25.488631'`. So **every `auto_now_add` timestamp was still broken** — the exact bug this PR is about, still live. Aware datetimes were fine via RFC3339, which is why the entire suite passed.

Sharpest part: the previous commit's self-review caught this class for the *no-seconds* claim and left the identical error standing for the *separator* claim.

**Fixed** with `%.f` (optional fraction) on both separators; 3 shapes pinned.

### 🔴 2 — `add` fabricated numbers where Django concatenates

`int("1.5")` **raises** in Python, so Django falls to concatenation. The `f64` fallback in the String arm meant:

| template | Django | before |
|---|---|---|
| `{{ "1.5"\|add:"1.5" }}` | `1.51.5` | **`2`** |
| `{{ 5\|add:"1.5" }}` | `` (empty) | **`6`** |
| `{{ 1\|add:"1e3" }}` | `` (empty) | **`1001`** |

Not the documented divergence — djust returned neither Django's value nor the value unchanged, but an **invented number**. Worse than the inert wrong output it replaced.

The distinguishing hint already existed: a quoted `"1.5"` is a string to `int()`, an unquoted `1.5` is a float literal. **Fixed** by threading `arg_was_quoted` into `apply_builtin_filter` and gating the `f64` fallback on it — `{{ n|add:1.5 }}` stays `6`, `{{ "4"|add:"3" }}` stays `7`.

### 🟡 Also fixed

- **`truncatechars:0`** returned `…` where Django returns `""` (`Truncator.chars` opens with `if length <= 0: return ""`). The PR's robustness probe checked this boundary for *panics*, not parity — exercised without being compared.
- **The `_html` twins still emitted `...`** (#1646) — the same filter disagreeing with itself on one page. The review flagged a landmine and it was real: `"…".len()` is **3 bytes**, identical to `"..."`, so swapping the constant alone silently preserves the three-character reservation. Fixed with `.chars().count()`.
- **`Value::Bool` missing from `as_int`** — `int(True)` is 1, so `{{ True|add:1 }}` is `2`, not `true`.
- **The timezone test was decorative** (#1859): `+00:00` is also what the naive branch's `.and_utc()` produces, so it passed whether the offset was honoured or discarded. Now `+05:00`.
- `checked_add` computed the sum twice; folded into one `zip`/`and_then`.

### The finding I'd have missed

Review ran a nine-mutation gate-off and found **`as_int`'s Float arm was gated by nothing** — neutering it left the whole suite green, while changing `{{ 1.5|add:2 }}` from `3.5` to `3`. That is this PR's **largest silent change to rendered output**, and it had no test and no CHANGELOG line. Both added.

All 8 mechanisms are now load-bearing and redden distinct sets:

| mutation | reddens |
|---|---|
| microsecond formats | 1 |
| `ELLIPSIS` constant | 4 |
| `truncatechars:0` guard | 1 |
| `as_int` Float arm | 1 |
| `as_int` Bool arm | 1 |
| quoted-float gate | 1 |
| `checked_add` | 1 (panic reproduced) |
| `_html` ellipsis | 1 |

### Confirmed clean by the review

No timezone bug is introduced: `format_date` reads wall-clock fields directly and supports no timezone format code, and chrono's `parse_from_rfc3339` accepts a space separator, so offset-bearing strings never reach the naive parsers. Format ordering is safe — chrono rejects trailing input, so no earlier format can partially match a string a later one should own. `.chars()` truncation matches Django's own grapheme-naive behaviour. Regression sweep confirmed `format_date` has exactly two callers and no filter that previously fail-softed now succeeds with a *wrong* value.

Two divergences are documented rather than fixed: Django NFC-normalizes before truncating (narrow, pre-existing), and Django's `date`/`time` are `expects_localtime=True`, so parity holds only when `TIME_ZONE` is UTC — the PR body previously overstated this.

### Deferral upheld

The reviewer independently verified that `divisibleby`/`slice`/`Null` bottom out in `Display for Value`, that the JS hazard is real (`true` valid, `True` a `ReferenceError`), and that fixing them per-arm "is exactly the N-correct-copies anti-pattern #1646/#2202 exist to prevent". Also corrected my count: **5 of 7** `Value` variants, not 5 of 8.

### Post-fix

22 cases; Rust 14/14 suites; Python **10,276 passed, 218 skipped, 0 failed**. Four legacy `...` assertions (2 Rust, 2 Python) updated to values taken from **Django itself**.

johnrtipton:
## Stage 14 — Retrospective

**Quality: 3/5.** The scope re-cut was the right call and the deferral held up under adversarial scrutiny. But the headline fix shipped *not actually fixing the headline case*, and that took an outside reviewer to catch — after my own self-review had already caught the identical error one claim over.

### What went well

**Re-cutting the scope from the issue's framing.** #2203 said "four filters diverge with a literal argument". Investigating the `time` case found something much larger: `date` and `time` were broken for **every** Python `datetime` — `{{ post.created_at|date:"Y-m-d" }}`, the commonest use of the filter, rendered a raw datetime string. The issue's own framing would have led to patching `time`'s arm and missing that `date` had the same root cause.

**The deferral was verified, not asserted.** I deferred `divisibleby`/`slice`/`Null` as `Display for Value` with a JS back-compat hazard. The reviewer independently confirmed the code backs that, checked the escape hatch I claimed didn't exist (fixing them per-arm), and agreed it is "exactly the N-correct-copies anti-pattern #1646/#2202 exist to prevent". Deferring under #1079 was right, and being able to point at measurements — 5 of 7 `Value` variants diverge, `var flag = true` vs a `ReferenceError` — is what made it defensible rather than evasive.

**Django as the authority for expected values.** Every changed assertion took its value from a live `django.template.defaultfilters` call, not from what my implementation produced. Six legacy `...` pins were updated this way. Circular self-validation would have been the easy failure here.

### What didn't

**I fixed the headline bug without fixing the headline case.** The premise — "a `datetime` arrives as `2026-08-22 14:30:00`" — was wrong: djust serializes with `.isoformat()` (`serialization.py:311`), and neither form omits microseconds. `datetime.now().isoformat()` always has them, so every `auto_now_add` timestamp stayed broken. I had *three* independent chances to catch this: writing the fix, the self-review, and the end-to-end Django parity check I ran — and all three used `datetime(2026,8,22,14,30)`, a hand-written value with zero microseconds. **The lesson is about fixture choice, not diligence**: when a fix depends on a serialized shape, derive the fixture from the real producer (`datetime.now().isoformat()`) rather than constructing a clean one by hand. A hand-built fixture silently selects the easy case.

**The same error, twice, one claim apart.** My self-review caught that the *no-seconds* justification was a false claim about Python — and left the *separator* claim, in the same comment, unexamined. Disconfirming one sentence and stopping is a real pattern: **when a comment turns out to be wrong, re-check every other claim in that same comment**, not just the one that failed.

**I widened a surface and only half-checked the widening.** `add` previously parsed `i64` only. I widened it to floats and numeric strings, and did check that this made overflow reachable (fixed with `checked_add`). I did not check the *other* consequence: `int("1.5")` **raises** in Python, so Django concatenates — and my version returned a fabricated `2`. Turning inert wrong output into a plausible wrong **number** is a worse failure than the one I set out to fix. When widening a coercion, enumerate what the reference implementation *rejects*, not only what it accepts.

**A mechanism with zero coverage passed my own gate-off.** I ran gate-off per mechanism and reported all of them load-bearing. The reviewer ran nine mutations and found `as_int`'s Float arm gated by nothing — neutering it left the suite green while changing `{{ 1.5|add:2 }}` from `3.5` to `3`, the PR's largest silent output change. My gate-off tested the mechanisms I *thought of*; it wasn't derived from the diff. **Enumerate mutation targets from the changed lines, not from memory.**

### Relation to earlier work in this cycle

This is the second consecutive PR (after #2202) where the adversarial review found that my fix **reintroduced the class it was curing** — there `.ok()` swallowing an error twenty lines from the path it converged with, here a fabricated number replacing an inert wrong one. Both times the mechanism was the same: I verified the *intended* behaviour thoroughly and under-verified the *edges I had just widened*.

It also re-proves #1646 twice in one PR: `date`/`time` shared one broken parse (fixed at the chokepoint), while `truncatechars`/`truncatechars_html` were the same filter for different input and drifted apart — I fixed the plain pair and left the twins, exactly the drift I had invoked #1646 to avoid three paragraphs earlier in the same CHANGELOG entry.

The `"…".len()` landmine deserves recording: it is **3 bytes**, identical to `"..."`, so the obvious fix to the `_html` twins would have looked correct and silently preserved the arithmetic bug.

### Verified

- Reproducer-first: 8 red / 5 green before the fix.
- Gate-off, all **8** mechanisms, each reddening a distinct set — microsecond formats, `ELLIPSIS`, `truncatechars:0`, `as_int` Float, `as_int` Bool, quoted-float gate, `checked_add` (reproduces the original panic), `_html` ellipsis.
- Rust 14/14 suites; Python **10,276 passed, 218 skipped, 0 failed**; CI 15/15.
- Six legacy `...` assertions updated to Django's own values.
- #2203 stays **open** for the `Display for Value` question.

`RETRO_COMPLETE`

PR #2211, opened 2026-08-24
feat(templates): Django-parity value rendering, behind django_value_repr (#2203)

### LINKED ISSUE #2203: bug: Display for Value diverges from Django for 5 of 7 variants (bool, None, list, dict, float)

johnrtipton:
Closes #2203.

`impl Display for Value` diverged from Django for **5 of the 7 `Value` variants**. Not only cosmetic — `Display` is the lookup key for `{% if x in dict %}`.

| | before | after (= Django) |
|---|---|---|
| `{{ flag }}` | `true` | `True` |
| `{{ none }}` | *(empty)* | `None` |
| `{{ 1.0 }}` | `1` | `1.0` |
| `{{ [1,2] }}` | `[List]` | `[1, 2]` |
| `{{ {'a':1} }}` | `[Object]` | `{'a': 1}` |
| `{{ (1,2) }}` | `[List]` | `(1, 2)` |

**19/19 bare-variable types now render byte-identically to Django**, verified against live Django renders rather than against this implementation.

Gated on `LIVEVIEW_CONFIG['django_value_repr']`, **default ON**. `False` restores the previous rendering verbatim.

### 1. `Value::Null` split into `Missing` and `None`

Django distinguishes an **absent** variable (`""` — its `string_if_invalid`) from a present `None` (`"None"`). djust collapsed both.

`renderer.rs`'s `resolve(..)?.unwrap_or(Value::Null)` folded *missing* into it — and so did `CallOutcome::Empty`, which is an `alters_data` refusal or a **serialization-floor denial**. Mapping the old `Null` to `"None"` without splitting would have made every missing variable render the literal text `None`, and put text where a refused `password` field rendered nothing.

That is not hypothetical: a prototype without the split failed **19 tests**, named `renders_empty`, `missing_variable`, `missing_attribute` and `does_not_leak`.

Both variants stay falsy and both still satisfy `is None` — checked against Django for present-`None`, absent, `0` and `"x"`.

### 2. `Object` is now an `IndexMap`

Rust randomises `HashMap` iteration per process. Measured across runs:

```
["epsilon","beta","gamma","delta","alpha"]
["gamma","alpha","beta","delta","epsilon"]
["delta","epsilon","beta","gamma","alpha"]
```

Python dicts are insertion-ordered, so dict repr would have been **non-deterministic** — the same template rendering differently between requests.

**The knock-on the compiler could not catch:** `loop_cache` hashed dict keys **sorted**, for an order-independent hash. That was correct only while order could not affect output. Now it can, so two dicts with the same pairs in different orders would hash alike and render differently — a **stale cache hit serving wrong output**. Hashing is now insertion-ordered, with distinct tags for `None` and `Tuple`.

### 3. A `Tuple` variant — where the real risk was

About **20 `Value::List` matches have a `_` fallback**, so the compiler *cannot* flag a missing twin. Tuples silently vanished from `{% for %}`, `|length`, `|join`, `|first`.

This was caught by a **security** test — `test_container_of_models_does_not_leak` — whose loop over a tuple of models stopped running at all. The test still "passed" its leak assertion while verifying nothing.

All 18 accessor arms now match `List(x) | Tuple(x)`. `Display`, `value_to_json`, `pprint` and `hash_value` deliberately still distinguish them — and clippy's `unreachable_patterns` caught a blanket sweep that had made `pprint`'s tuple arm dead code.

## Also fixed while verifying

A dict with **non-string keys** used to degrade gracefully (the old `extract::<HashMap<..>>()` failed and fell through). Hand-iteration turned that into a hard `TypeError` — a regression I introduced. It falls through again, and now renders `{1: 'x'}` / `{True: 1}`, matching Django.

## The JS hazard, stated plainly

```
var flag = {{ v }};    before -> "var flag = true;"    valid JS
                       after  -> "var flag = True;"    ReferenceError
```

Django has the identical hazard — which is why `json_script` and `|yesno` exist, and both work here. The flag is the one-line escape while such templates migrate.

## Verification

- **Reproducer-first**: 13 cases written before the fix, red at compile time (the variants did not exist).
- **Gate-off** on all five mechanisms — flag gate 1, bool 1, `None` 2, integral float 1, `Missing`/`None` split 2. A first pass left the **flag gate itself untested** (removing it failed nothing, since every case ran default-ON); 3 OFF-path cases were added, with the panic-safe `FlagGuard` from `virtual_keyed_ops_2017.rs`.
- **Rust 986 passed / 0 failed** across all five crates; **Python 10,276 passed / 0 failed** across all three roots.
- Every updated legacy pin took its new value from a live Django render.

Test-counting note: the harness used here refuses to report a number when the build fails — an earlier run reported "0 failures" off a suite that had not compiled, which is the #2129/#2135 false-green.


johnrtipton:

**Verdict: REQUEST_CHANGES** → addressed in `452d421c`. Every finding verified independently before acting. **Four of the five must-fixes were regressions I introduced.**

### 🔴 1 — `{% if x == None %}` was unconditionally false

`values_identity` got the `Missing | None` arm and a six-line comment explaining why. `values_equal`, **nineteen lines above it in the same file**, did not — so both operands were `Value::None`, no arm matched, and `_ => false`.

| template | Django | `main` | this PR (before fix) |
|---|---|---|---|
| `{% if pynone == None %}` | `Y` | `Y` | **`N`** |
| `{% if absent == None %}` | `Y` | `Y` | **`N`** |
| `{% if pynone != None %}` | `N` | `N` | **`Y`** |

A regression against Django *and* against the previous release — #1646 drift between two functions one commit touched.

### 🔴 2 — `default_if_none` was inverted

It fired on an *absent* variable and rendered the literal text `None` for the one input it is named for. Now `Value::None` only: Django substitutes `string_if_invalid` (`""`) for an absent variable **before** the filter runs, so it never sees `None` there. That also closes a pre-existing divergence where absent returned the fallback. 4/4 against Django.

### 🔴 3 — The new test file raced with itself

`FLAG_LOCK` was taken only in `off()`, so the ten default-ON tests read the process-global while an OFF test held it `false`. **RED on roughly 1 run in 3**, a different test each time — including `dict_rendering_is_stable_across_repeated_construction`, the determinism guard, failing non-deterministically. **CI was green by luck.**

The reference this file cites locks *both* states; only half the pattern had been lifted. Fixed, and a deadlock the fix would otherwise have introduced (one test taking the lock twice) removed with it. Now 5/5 clean parallel runs.

### 🔴 4 — The `__dict__` fallback reintroduced the non-determinism this PR removes

It built its `IndexMap` by iterating a std `HashMap` — sixty lines below the comment saying *"extraction is exactly where insertion order would be lost."* And worse than the CHANGELOG framed it: randomisation is per **instance**, and a fresh map is built per conversion, so `{{ obj }}` reordered on **every render**. Now iterates the `PyDict` directly — 1 distinct rendering across 25 runs, was effectively random.

### 🔴 5 — Four mechanisms were ungated

Gate-off showed all three `loop_cache` changes and the `Tuple` twin in `{% for %}` could be reverted with the suite staying green — including the ones the CHANGELOG describes as preventing *"a stale cache hit serving wrong output."*

4 cases added. Notably **the first version of them gate-off'd to zero too**, because a plain `Template::render` never consults the cache — they exercised the uncached path and could not detect a hashing regression at all. They now install it via `LoopCacheGuard` against a held template, mirroring production. All three now redden.

### 🟡 Also fixed

- Dict **keys** used a hand-rolled escaper that missed the backslash, so a key like `a\` emitted `{'a\': 1}` where the closing quote reads as escaped. Now routed through `py_repr`, same as values.
- `py_repr` now follows Python's quote-selection rule: `repr("a'b")` is `"a'b"`, not `'a\'b'`. That took parity from 18/20 to **20/20**.
- `python_to_value` checked `i64` **before** `bool`, which PyO3 0.29 extracts a Python `True` as — making the `bool` arm dead code and rendering `1` for actor params and component props. The comment asserting the two converters agree is now true. A **third** converter with the same ordering is filed as **#2212** (a Python→Python path, orthogonal to this PR).
- CHANGELOG test count corrected (13 + 4, not 16).

### The security path, independently confirmed

The review traced all three producers and confirmed the highest-stakes part of the `Null` split is correct: a serialization-floor denial is expressed as **absence** (omitted key / `AttributeError`), so it resolves to `Missing` and renders **nothing**, never `"None"`.

Worth recording the nuance: that safety depends on denial being absence. Any future denial path that *nulls* a field instead of dropping it would print the literal text `None` where it previously printed nothing.

### Still open, deliberately

- **`|unordered_list` on a non-list emits unescaped** (🟡). Pre-existing name-based `SAFE_OUTPUT_FILTERS` behaviour that this change widens from an inert `[Object]` to attacker-influenceable content. Real, and worth its own fix rather than being folded in here.
- **Float parity above `1e16`** and Python's exponent form (🟡) — `1e16` renders `10000000000000000` where Python gives `1e+16`.
- **`Value` serde round-trip is lossy** for `None`/`Tuple` (🟡) — latent, no production caller today, but the #1538/#1541 class.

### Post-fix

20/20 bare-variable types match Django. **Rust 993 passed / 0 failed**, stable across 5 parallel runs; **Python 10,276 passed / 0 failed**.

*(Process note: the reviewer's worktree was deleted mid-run and its later commands fell back to the main checkout. I verified repo integrity independently afterwards — HEAD, `core.bare`, no stray branches, and all in-flight uncommitted work byte-identical.)*

johnrtipton:
## Stage 14 — Retrospective

**Quality: 3/5.** The feasibility work up front was the best part and the reason the design is sound. The implementation then introduced four regressions that an outside reviewer caught, including one that made a basic template construct unconditionally false.

### What went well

**Measuring feasibility before committing to it.** Before writing anything I prototyped the change and ran the suite: **27 failures**, of which 19 were named `renders_empty`, `missing_variable`, `missing_attribute` and `does_not_leak`. That failure *pattern* — not any single test — is what revealed that Django distinguishes an absent variable from a present `None` and djust collapsed both. Had I started from "make bools render `True`", I would have shipped a change that printed the literal text `None` where a refused password field used to print nothing. The prototype cost twenty minutes and redesigned the whole PR.

**Refusing to answer the scope question myself.** Three of the six divergences turned out to need type changes with real blast radius, and one — the JS hazard — is a judgment about the framework's audience, not a technical call. Presenting measured options and asking was right; the answer ("all six, flag-gated") changed what I built.

**Django as the authority, never my own output.** Every expected value came from a live `Template(...).render(...)`. When six legacy pins had to change, each new value was taken from Django rather than from what my implementation happened to produce. That is the difference between a parity fix and a rename.

### What didn't

**I updated one of two sibling functions.** `values_identity` got the `Missing | None` arm *and a six-line comment explaining why it was needed*. `values_equal`, nineteen lines above it in the same file, did not — so `{% if x == None %}` became unconditionally false, a regression against Django and against the previous release. I wrote the reasoning for exactly this hazard and then failed to apply it one function away. **When a type split changes what "equal" means, grep for every comparison — the one you are editing is not the only one.**

**I trusted a mechanical sweep over ~20 sites the compiler could not check.** Adding `Value::Tuple` meant every `Value::List` match with a `_` fallback silently stopped matching tuples. My sweep fixed 18 arms and I moved on; clippy's `unreachable_patterns` then caught that the same sweep had made `pprint`'s tuple arm dead code. **A `_` fallback converts a type change from a compile error into a silent behaviour change — enumerate those sites deliberately, not with a regex.**

**My own gate-off passed on tests that could not fail.** I reported five mechanisms gated. The reviewer ran nine mutations and found four more that were not — including all three `loop_cache` changes, the ones the CHANGELOG calls preventing "a stale cache hit serving wrong output." Then my *first fix* for that gate-off'd to zero as well, because a plain `Template::render` never consults the cache: the tests exercised the uncached path and could not detect a hashing regression at all. **Reproduction fidelity applies to the gate-off harness itself, not only to the bug repro.**

**I reported green off a build that had not compiled.** My `awk` summed `test result:` lines that never existed, so "0 failures" meant "nothing ran." This repo has a canon rule for exactly that (#2129/#2135) and I walked into it anyway. The harness now refuses to print a number unless the build succeeded.

**I shipped a test file that raced with itself.** It took the lock only in `off()`, so the ten default-ON tests read the process-global while an OFF test held it false — red on roughly one run in three, including the determinism guard failing non-deterministically. **CI was green by luck, and I reported that green as evidence.** The reference I cited in the file's own comment locks both states; I lifted half the pattern.

### The through-line

Every one of these is the same shape: **I verified the thing I was building and under-verified the things it touched.** The `Display` change itself is correct and was checked exhaustively — 20/20 against Django. What broke was the perimeter: a sibling comparison function, twenty pattern matches, the cache key, the test harness's own validity. A type change's blast radius is not the diff; it is every place the old type's semantics were assumed.

### Relation to earlier work

This is the third consecutive PR where the adversarial review found my fix reintroducing the class it cured — `.ok()` swallowing an error in #2204, a fabricated number in #2208, and here a `Missing`/`None` split applied to one of two comparison functions. The reviews are earning their cost every time, and the pattern is stable enough to name: **I am reliably good at the change and reliably weak at its neighbours.**

### Verified

- 20/20 bare-variable types match Django, including tuples, nested containers, non-string dict keys, backslash/quote keys and Python's quote-selection rule.
- Gate-off on all mechanisms, each reddening a distinct set; the cache tests re-written against the real cached path after the first version proved vacuous.
- **Rust 993 passed / 0 failed**, stable across 5 parallel runs; **Python 10,276 passed / 0 failed**; CI 15/15.
- Security path independently confirmed: floor denial is *absence*, resolves to `Missing`, renders nothing.
- Follow-ups filed: **#2212** (a third converter with the same bool/i64 ordering). Left open and documented: `|unordered_list` escaping, float parity above `1e16`, lossy `Value` serde round-trip.

`RETRO_COMPLETE`

Issue #1882, opened 2026-06-23
tech-debt: test_time_travel_jump_recovery_version_is_current is order-fragile under -n auto (process-global wire-version counter pollution)


johnrtipton:
Source: v1.1.0-1 / PR #1881 (#1875 drain — surfaced while running the 3-clean-runs gate).

`python/djust/tests/test_recovery_version_staleness_1817.py::test_time_travel_jump_recovery_version_is_current` is order-fragile under full `-n auto`: a process-global wire-version counter is polluted across tests, so the jump-render assertion sees the wrong version (`jump render must advance the wire version to 3 … got 4`). Passes in isolation; intermittent under saturation. Same class as #1875 (#1862) — shared global state across the xdist worker.

**Action**: isolate/reset the process-global wire-version counter (or the consumer that owns it) in setup/teardown, or assert an ordering invariant per the flaky-timing canon (#1830/#1795). Out of #1875's scope (#1079) — filed as its own item. Reproduce by running the recovery/time-travel ws slice under `-n auto` repeatedly.

PR #2213, opened 2026-08-25
fix(serialization): check bool before i64 in serialize_python_value (#2212)

### LINKED ISSUE #2212: bug: a third Python->Value converter checks i64 before bool, so a bool round-trips as 1
Split out of the #2211 review to keep that PR scoped (#1079).

PyO3 0.29 extracts a Python `True` as `i64` `1` (its own test asserts this — `pyo3-0.29.2/src/conversions/std/num.rs::test_i64_bool`). So any converter that tries `i64` **before** `bool` has a dead `bool` arm and silently turns booleans into integers.

There are **three** such converters. #2211 fixed one; a third remains:

| location | order | status |
|---|---|---|
| `djust_core/src/lib.rs` `FromPyObject for Value` | bool first | correct, always was |
| `djust_live/src/lib.rs` `python_to_value` | i64 first | **fixed in #2211** |
| `djust_live/src/lib.rs:2343-2354` | **i64 first** | **still open** |

The third is a Python→Python normalization path, not Python→`Value`, which is why #2211 left it alone — different function, different purpose, and orthogonal to that PR's `Display` work.

```rust
    if let Ok(i) = value.extract::<i64>() {      // :2346  <- a bool lands here
        return Ok(i.into_pyobject(py)?.to_owned().into_any().unbind());
    }
    if let Ok(f) = value.extract::<f64>() {      // :2349
        ...
    }
    if let Ok(b) = value.extract::<bool>() {     // :2352  <- unreachable for bools
        ...
    }
```

**Effect:** a `True` passing through this path comes out as Python `1`, not `True`. Worth checking which callers reach it before deciding severity — I have not traced them.

**Fix:** move the `bool` arm above `i64`, matching the other two converters. Add a case asserting a `bool` survives the round-trip as a `bool`, since the compiler cannot flag a dead `if let` arm.

This is the same shape as #1646 parallel-path drift: one invariant, three implementations, and the compiler cannot see a divergence in any of them.

johnrtipton:
Closes #2212.

PyO3 0.29 extracts a Python `True` as `i64` `1` — its own `test_i64_bool` asserts this — so a converter that tries `i64` first has a **dead bool arm**. `serialize_python_value` did:

| call | before | after |
|---|---|---|
| `serialize_context({"flag": True})` | `1` *(int)* | `True` |
| `serialize_context({"flag": False})` | `0` *(int)* | `False` |
| `{"d": {"flag": True}}` | `{'flag': 1}` | `{'flag': True}` |
| `{"xs": [True, False]}` | `[1, 0]` | `[True, False]` |

**This is user-visible, not tidiness.** `serialize_context` is a public `#[pyfunction]` feeding JIT state serialization, so a bool in LiveView state reached the client as `1`/`0`. Client code doing `x === true` sees false. And after #2203, a `Value::Integer(1)` renders `1` where a `Value::Bool(true)` renders `True`.

## The issue undercounted the surface

I filed #2212 claiming **three** converters extract both types. A sweep of every `fn` under `crates/` found **six**:

```
OK  djust_live/src/lib.rs:1590   python_to_json_value
OK  djust_live/src/lib.rs:2023   python_to_value          (fixed in #2211)
OK  djust_live/src/lib.rs:2286   serialize_python_value   (fixed here)
OK  djust_live/src/lib.rs:2667   python_to_json
OK  djust_live/src/model_serializer.rs:84  python_to_json
OK  djust_core/src/lib.rs:326    extract
```

The other three were already correct — but enumerating by hand had missed half of them, and that is the argument for the guard below being **structural rather than a fourth hand-written case**.

## Why a structural guard, and why it isn't a banned grep-pin

Nothing mechanical protected this invariant:

- a dead `if let` arm is **not** a compile error
- clippy does not flag it either — the arms have different types, so it isn't `unreachable_patterns`
- **four of the six** converters have no behavioural test at all

So the guard is not a strict subset of a behavioural test, which is the #2167 objection to source-grep pins. The behavioural test for the converter this issue names lives in `TestBoolRoundTrip2212`; the structural sweep covers the four that would otherwise be unprotected.

**It also guards itself.** A parser that stops matching would silently protect nothing, so one case asserts the sweep still finds at least five converters — it goes red rather than going quiet.

## Verification

- **Reproducer-first**: 3 cases written before the fix, **2 red / 1 green** — the green one is the int-guard, which must keep passing.
- **Gate-off** on both mechanisms: reverting the arm order reddens the structural guard (1) and the behavioural test (2).
- **Rust 993 passed / 0 failed**; **Python 10,286 passed / 0 failed**.

## One unrelated thing folded in

Committing `PERFORMANCE_BRAINSTORM.md` during this drain put an unchecked `~58 KB gzipped` claim in the tree, which the #2138 size-claim guard correctly caught. The doc cites CLAUDE.md's shipped-client budget to argue against adding a JS decompressor, so the figure is load-bearing for that argument — it is now **checked against the real artifact** rather than exempted, so if the client drifts past budget the doc's reasoning stops silently holding.


johnrtipton:

**Verdict: REQUEST_CHANGES** → addressed in `888b8be9`. The reordering fix was confirmed correct and would have merged as-is; **every finding was in the guard.**

That distinction matters less than it sounds, because the review established something I had not: **the guard is the only protection for five of the six converters.** A gate-off re-introducing the #2211 bug in `python_to_value` left the entire suite green — 10,278 passed, 0 failed. So the guard's own defects were the real risk, not a cosmetic concern.

### 🟡 Three parser defects, each demonstrated with a canary

**1. Silent miss.** The signature regex did not match `pub(crate) fn`, `pub(super) fn`, `const fn`, `unsafe fn`, or `extern "C" fn`. Three such functions already exist in the tree. A synthetic `pub(crate) fn` carrying the exact #2212 bug produced **zero rows** — the guard stayed green on a real violation.

**2. Mis-attribution — the sharpest one.** Bodies ended at the next signature match rather than at a balanced brace. So renaming a function to `pub(crate)` — a legal, behaviour-free refactor — absorbed its body into the **preceding** function. The count held at six, the self-check passed, and protection had silently transferred to a function with no such arms. That is the #1859 decorative-pin class, reached by a route I had not considered.

**3. False positive.** Comments were not stripped, so a correct converter whose comment merely *mentioned* `extract::<i64>()` was reported as broken. **Not hypothetical**: this PR's own fix puts a 13-line comment block directly above the bool arm, and would have tripped it had the comment spelled the literal.

All three fixed, and all three canaries re-run: a `pub(crate)` rename now keeps the converter under its own name, a mentioning comment no longer fails correct code, and a real reordering still reddens.

### 🟡 The self-check was a floor, not a set

`len(found) >= 5` against a real count of six tolerates losing one converter outright — which is exactly what defect 2 produced. Review's degradation matrix showed four distinct parser breakages that all still passed. Now pins the expected **set** (#1125), so a disappearance *and* an unexpected new converter both go red.

### 🟡 My own claims were wrong, in my favour

"Four of the six have no behavioural test" is actually **five** — and before this PR it was six. Verified by gate-off. The claim was wrong in the direction that *strengthens* the argument for a structural guard, which is precisely why it needed correcting rather than leaving.

Also corrected: two artifacts still said "three converters" — the PR's headline finding is that hand-enumeration undercounted at three when it is six, and those two places are the first a maintainer reads.

### 🟢 A second instance of the same class, in the same function

`extract::<f64>()` sits above the `Decimal`/`UUID` stringify branch, and PyO3's `f64` extraction honours `Decimal.__float__` — so that branch is dead for `Decimal` and **every `DecimalField` reaches the client as a lossy binary float**:

| input | arrives as |
|---|---|
| `Decimal('19.99')` | `19.99` *(float)* |
| `Decimal('12345678901234567890.123456789')` | `1.2345678901234567e+19` |

`DecimalField` is what Django projects use for money, and binary floats are what it exists to avoid. The function's own doc-comment claims these convert to strings — false about the code it sits on (#1867).

Filed as **#2214** (#1079), carrying the more valuable half of the suggestion: widen this guard from the i64/bool pair to the general rule it is an instance of — *no `type_name ==` special case below a permissive `extract::<T>()`*. That would have caught both bugs, and would catch the next one.

### Verified good by the review

PyO3 0.29's bool extraction is `cast::<PyBool>()` plus a numpy-`bool_` case — **strict**, so putting it above `i64`/`f64` cannot capture ints, floats, `Decimal`, `__index__` or `__float__` objects. Edge cases run against the built extension: `2**62`, `2**80`, `-0.0`, an `int` subclass, `__index__`/`__float__`/`__bool__`-only objects, `Decimal`, `UUID`, `datetime`, `bytes`, and nested containers — **only bools changed**. An independent sweep confirmed six converters and verified all six file:line citations. The clippy claim holds (zero warnings on the mutated dead-arm source). Blast radius clean: nothing depended on bools arriving as ints. The size-claim registration is load-bearing, not decorative — forcing the manifest to 70 KB produces a real failure naming `PERFORMANCE_BRAINSTORM.md:395`.

### Post-fix

9 structural cases (was 7) plus 3 behavioural. **Python 10,288 passed / 0 failed**; Rust 993 / 0.

johnrtipton:
## Stage 14 — Retro, PR #2213 (#2212)

**Quality: 3/5.** The fix is right and was right from the first pass. The guard I wrote to protect it was broken in three ways, and the guard is the only protection five of the six converters have.

### The finding that reframed the PR

I opened this thinking the structural guard was belt-and-braces around a behavioural test. The review's gate-off said otherwise: re-introducing the #2211 bug in `python_to_value` left **10,278 passed, 0 failed**. Five of six converters have no behavioural coverage at all. So the guard is not a supplement — it *is* the protection, and every defect in it was a hole in the only wall.

That inverts how much care the guard deserved relative to the one-line reorder, and I had the ratio backwards while writing it.

### What went well

- The reorder itself survived adversarial review unchanged. PyO3 0.29's bool extraction is `cast::<PyBool>()` — strict — so the edge sweep (`2**62`, `2**80`, `-0.0`, `int` subclass, `__index__`/`__float__`/`__bool__` objects, `Decimal`, `UUID`, `datetime`, `bytes`, nested containers) showed only bools changing.
- Enumerating converters *structurally* rather than by hand was the right instinct: hand-enumeration said three, the grep said six. The instinct was sound even though the execution wasn't.
- Filing #2214 rather than fixing the `Decimal` dead branch inline (#1079). The issue carries the more valuable half anyway — widen the guard to the general rule, not the specific pair.

### What didn't

**I wrote a parser and did not test the parser.** Nine cases asserting things *about* converter bodies, zero asserting the extraction produced the right bodies. The reviewer's canaries — a `pub(crate) fn` carrying the real bug, a rename that migrated a body to its neighbour, a comment mentioning `extract::<i64>()` — took minutes each and I wrote none of them. When a test's subject is code you also wrote, the extraction layer is a second subject and needs its own coverage.

**`>= 5` against a real count of 6.** I wrote a floor because a floor felt tolerant of future growth. It is precisely tolerant of the failure the parser produces — losing one converter — which is the one thing it must not tolerate. #1125 says pin the set; I had read that rule and still wrote the floor.

**The claim I got wrong favoured me.** I wrote "four of six have no behavioural test" when it is five, and was six pre-fix. Wrong in the direction that makes my own argument stronger. That direction of error is the one to distrust most, and the only reason it got corrected is that someone else measured.

### The pattern, now four PRs deep

#2208, #2211, #2213 — and #2204 before them. Each time the change is correct and something adjacent to it is not: `.ok()` swallowing the exceptions I was curing; `values_identity` updated and `values_equal` nineteen lines above left alone; now a guard whose parser cannot see three function shapes already in the tree.

The repo has a name for this — parallel-path drift, #1646 — and I keep producing it while holding the rule in context. The distinguishing feature is that **the neighbour is always something I could have grepped for in under a minute**: other `fn` prefixes, the other equality function, the other exception path. Reading the diff does not surface them; a targeted grep does. The concrete change I'd make: after any fix, grep for the *shape* of the thing changed — not the symbol — before calling it done.

### Verified

- 9 structural cases (was 7) + 3 behavioural; **Python 10,288 passed / 218 skipped / 0 failed**; Rust 993 / 0; clippy 0 warnings.
- All three reviewer canaries re-run against the rewritten parser: `pub(crate)` rename keeps the converter under its own name, a mentioning comment no longer fails correct code, a real reordering still reddens.
- Corrected "three converters" → six in `crates/djust_live/src/lib.rs` and `python/tests/test_template_serialization.py`.
- Follow-up: **#2214** (`Decimal` → lossy float, same function).

`RETRO_COMPLETE`

johnrtipton:
CI note: the first run of `888b8be9` failed `python-tests (py3.12)` on `test_time_travel_jump_recovery_version_is_current` — `arming event must be 2, got 3`. **Not this PR.** The commit before it (`631157cf`, carrying the actual reorder) was green, and `888b8be9` changes no production code on that path — two extra cases in a structural test file plus doc-comment corrections. Adding test cases re-shards xdist, which is enough to trip it.

It is the #1882 wire-version pollution class, which `python/djust/test_isolation.py:16` documents as cured. Isolation 3/3 green, local CI-way full run 3/3 green (10,288 passed), CI re-run of the same SHA green — it needs the runner's 4-worker sharding to reproduce. Filed as **#2215** with the reproduction status and a note that a pre-test reset cannot cure a producer that is live *during* the test.

Issue #2217, opened 2026-08-25
tech-debt: nine Django date format codes are unimplemented and render as their own letter


johnrtipton:
Found while building the #2209 differential. #2209 implements the five **timezone** codes (`e`, `T`, `O`, `Z`, `I`) plus `U`, because those are the visible face of the blindness that issue fixes. The rest are a separate completeness gap and are filed here rather than folded in (#1079).

## Unimplemented

`format_date` (`crates/djust_templates/src/filters.rs`) ends with a catch-all `_ => result.push(ch)`, so an unhandled code renders as **its own letter**. `{{ v|date:"jS F Y" }}` renders `22S August 2026` instead of `22nd August 2026`.

| code | Django (for `2026-08-22 19:30 EDT`) | djust today |
|---|---|---|
| `S` | `nd` (English ordinal suffix) | `S` |
| `w` | `6` (day of week, 0=Sunday) | `w` |
| `t` | `31` (days in month) | `t` |
| `L` | `False` (leap year) | `L` |
| `b` | `aug` (lowercase 3-letter month) | `b` |
| `f` | `7:30` (hour:minute, minutes elided when zero) | `f` |
| `u` | `000000` (microseconds) | `u` |
| `c` | `2026-08-22T19:30:00-04:00` (ISO 8601) | `c` |
| `r` | `Sat, 22 Aug 2026 19:30:00 -0400` (RFC 5322) | `r` |

`S` is the one most likely to bite — `jS` is a common Django idiom and produces visibly broken output rather than a subtly wrong one.

## Why the catch-all makes this quiet

Rendering an unknown code as itself is the right behaviour for a *literal* character, and Django does treat unrecognised characters that way. The trouble is it makes an *unimplemented* code indistinguishable from a literal one, so the gap is invisible to anything but a differential against Django. Worth considering whether the implemented set should be an explicit table with a debug-level log for a miss.

## Suggested approach

Mechanical — each is one to three lines against the `DateTime<FixedOffset>` already in hand, and `c`/`r` are chrono format strings. The value is in the differential: pin all 26 codes against Django's own render in one table so the next gap cannot hide, per the enumerate-every-variant rule (v1.0.0rc4 retro finding #1). The #2209 test file (`crates/djust_templates/tests/test_timezone_parity_2209.rs`) is the shape to extend.

Issue #2216, opened 2026-08-25
bug: date/time filters on a bare date or time object diverge from Django


johnrtipton:
Found while building the #2209 differential (djust's Rust engine vs Django's own engine, same values, same settings). **Out of #2209's scope** — that issue cites datetimes, and a `date` is not a `datetime` — so filed separately per the scope-discipline rule (#1079).

## The `time` case is the more serious one

`{{ v|time:"H:i" }}` on a `datetime.time` **returns its input verbatim**:

| value | Django | djust |
|---|---|---|
| `time(23, 30)` `\|time:"H:i"` | `23:30` | `23:30:00` |
| `time(23, 30)` `\|date:"P"` | `11:30 p.m.` | `23:30:00` |
| `time(23, 30)` `\|date:"g:i A"` | `11:30 PM` | `23:30:00` |

This is the **exact class #2203 / PR #2208 fixed for datetimes** — the parse misses, the filter falls through, and the raw serialized value lands in the page. `format_date`'s parse-format list (`crates/djust_templates/src/filters.rs`) carries four datetime shapes and one date-only shape, and no time-only shape at all. A `TimeField` rendered through `|time` is a normal thing to do, and it silently shows seconds Django would not show.

## The `date` case is a fail-soft divergence, not obviously a bug

A `datetime.date` serializes to `"2026-08-22"`, which the date-only branch parses as midnight UTC. Time codes then render that midnight:

| value | Django | djust |
|---|---|---|
| `date(2026,8,22)` `\|date:"H:i"` | raises `TypeError` | `00:00` |
| `date(2026,8,22)` `\|date:"g:i A"` | raises `TypeError` | `12:00 AM` |
| `date(2026,8,22)` `\|date:"P"` | raises `TypeError` | `midnight` |
| `date(2026,8,22)` `\|date:"T"` | raises `TypeError` | `EDT` (post-#2209) |

Rust receives `"2026-08-22"` and cannot tell a `date` from a datetime at midnight, so replicating Django's `TypeError` would need type information the serializer currently discards. And djust's house policy is fail-soft where Django raises (see the `{% url %}` row in the v1.1.1-2 ROADMAP bucket, filed for exactly this kind of deliberate-vs-inherited call). **This half may be correct as designed** — it wants a decision and a documented rule more than it wants a patch.

Note `T` moved from rendering the literal `T` to rendering `EDT` here as a side effect of #2209. Both diverge from Django's `TypeError`; the match rate is unchanged.

[...]

PR #2218, opened 2026-08-25
fix(templates): render timestamps in settings.TIME_ZONE, as Django does (#2209)

### LINKED ISSUE #2209: bug: Rust template engine ignores USE_TZ/TIME_ZONE — every timestamp renders in UTC
## Summary

The Rust template engine performs **no timezone conversion**. Under
`USE_TZ=True` with a non-UTC `settings.TIME_ZONE` — the default scaffold
configuration — every rendered timestamp is off by the UTC offset.

Django applies `timezone.localtime()` before formatting a datetime. djust does
not, at any layer.

## Reproduction

Confirmed through the real `LiveView.render()` path, not just the filter:

```python
import django
settings.configure(
    USE_TZ=True, TIME_ZONE="America/New_York", SECRET_KEY="x",
    INSTALLED_APPS=["django.contrib.contenttypes", "django.contrib.auth", "djust"],
    DATABASES={"default": {"ENGINE": "django.db.backends.sqlite3", "NAME": ":memory:"}},
    TEMPLATES=[{"BACKEND": "django.template.backends.django.DjangoTemplates",
                "DIRS": [], "APP_DIRS": False, "OPTIONS": {}}],
)
django.setup()

import datetime as dt
from django.test import RequestFactory
from djust import LiveView

aware = dt.datetime(2026, 8, 22, 23, 30, 0, tzinfo=dt.timezone.utc)

class V(LiveView):
    template = '<div dj-id="0">{{ created|date:"Y-m-d H:i" }}</div>'
    def mount(self, request, **kw):
        self.created = aware

v = V()
req = RequestFactory().get("/")
v.mount(req)
v.request = req
print(v.render())
```

**Actual**

```
<div dj-id="0">2026-08-22 23:30</div>
```

**Expected** (what Django's own engine renders for the same value)

```
2026-08-22 19:30
```

A 4-hour error. Both value shapes are affected — `.isoformat()` via the eager
dict and `str()` via the sidecar walk:

```
DJUST renders (isoformat): '2026-08-22 23:30'   [input '2026-08-22T23:30:00+00:00']
DJUST renders (str()    ): '2026-08-22 23:30'   [input '2026-08-22 23:30:00+00:00']
```

## Root cause

There is no timezone conversion anywhere in `python/djust/` or `crates/`. The
only `chrono::Local` reference in the workspace is `renderer.rs:1208`, which
implements `{% now %}`:

```
$ grep -rn "TIME_ZONE\|localtime\|Local::\|FixedOffset" crates/*/src/*.rs
crates/djust_templates/src/renderer.rs:1208:            let now = chrono::Local::now();

$ grep -rn "localtime\|get_current_timezone\|USE_TZ" python/djust/ | grep -v tests
python/djust/scaffolding/templates.py:171:USE_TZ = True
```

`normalize_django_value` (`python/djust/serialization.py`) emits `.isoformat()`,
which preserves the UTC offset the ORM returns under `USE_TZ=True`. The Rust
`date`/`time` filters then format whatever offset they are handed, with no
awareness of `settings.TIME_ZONE`.

Note the scaffold generated by `djust new` sets `USE_TZ = True`
(`scaffolding/templates.py:171`), so this is the default configuration rather
than an unusual one.

## Scope

- `date` / `time` filters on any `datetime`.
- Almost certainly localized month/day names in `date` format strings
  (`{{ d|date:"F" }}`) — same locale-blindness, **not probed**.
- Related but separate: `floatformat` ignores `USE_L10N` /
  `USE_THOUSAND_SEPARATOR` (Django `'1.234,50'` vs djust `'1234.50'` under
  `LANGUAGE_CODE="de"`). Same class; worth folding into one fix.

## Relationship to #2203

#2203 / PR #2208 (merged 2026-08-23) fixed datetime *parsing* in these filters —
before it, `{{ post.created_at|date:"Y-m-d" }}` returned its input verbatim.
Timezone was never in that issue's scope. This issue is the next layer down: the
value now parses correctly and is then formatted in the wrong zone.

## Suggested fix — and why the obvious one is wrong

The naive fix is a per-value `timezone.localtime(dt)` round trip in the
serializer. That reintroduces a Python/GIL crossing per rendered datetime, in
one of the few render paths that currently has none — on a table of timestamps
that is a crossing per cell.

Preferred shape: resolve the active timezone (and locale) **once per render**
and pass them into Rust, then convert there with `chrono-tz`. The existing
per-render flag plumbing is the pattern to mirror —
`_apply_loop_render_cache_flag` and `_apply_template_auto_call_flag` in
`python/djust/mixins/rust_bridge.py` are both idempotent, `hasattr`-guarded
setters called on every `_rust_view` (re)initialization.

Cost to check before committing to it: `chrono-tz` bundles a compiled tzdata
table into the wheel. Measure the size delta.

Design notes and alternatives are written up in `PERFORMANCE_BRAINSTORM.md` §9
(Idea 10).

## Verification method

Executed against the working tree at `fix-2203-display-django-parity`
(`73cf03db`) using the repo `.venv`. Reproduced twice — once against the `date`
filter directly via `djust._rust.render_template`, once end-to-end through
`LiveView.render()` as shown above. The grep evidence above is what establishes
this as structural rather than an artifact of the probe's configuration.

johnrtipton:
Closes #2209.

Django applies `timezone.localtime()` to an aware datetime before formatting it. The Rust engine did **no timezone conversion at any layer** — it formatted whatever offset the serializer handed it, which under `USE_TZ=True` is UTC.

```
Django:  2026-08-22 19:30
djust:   2026-08-22 23:30
```

Four hours out, **in the configuration `djust new` generates** — the scaffold sets `USE_TZ = True`. Reproduced through the real `LiveView.render()` path, not just the filter.

## Django's rules, and the row that is easy to get wrong

Taken from a live Django 5.2 render of five value shapes × twenty-six format codes, not from the docs:

| input | wall clock | `T` / `O` / `Z` | `e` |
|---|---|---|---|
| **aware** | converted to the active zone | that zone at that instant | same |
| **naive** | **unchanged** | the active zone at that *local* time | `""` |

A naive value is already understood to be local, so shifting it would move every timestamp in a `USE_TZ = False` project — where naive datetimes are the norm. Note the naive row still *reports* the default zone even though it does not convert, and that `e` and `T` disagree there, which is why each code is handled rather than sharing a branch.

## Why `chrono-tz` and not an offset

A fixed offset passed once per render would have been free. It is also wrong: `America/New_York` is `-0500` in January and `-0400` in August, so **one render can need both** — any table of timestamps spanning six months does. Correct conversion needs real transition data.

Cost measured before adopting rather than discovered after:

| | raw | compressed |
|---|---|---|
| extension delta | +1.16 MB | **+136 KB** |

The compressed figure is what a wheel actually ships. `winter_and_summer_disagree_within_one_render` is the test that goes red if someone later "optimises" this back to an offset.

## Why a thread-local set per render

Two designs were considered and rejected, both for reasons worth recording.

**`Context` field.** The obvious carrier — it already holds `auto_call`. But there are **seven** `Context` construction sites feeding a render, `set_auto_call` is called at **three**, and two others (`djust_live/src/lib.rs:1496`, `:1537`) already silently miss it. Adding a second setter every site must remember is adding a second instance of a drift that has already happened (#1646).

**Process-global set at startup**, like `django_value_repr`. Rejected because `timezone.activate()` is per-request — the documented way to give each user their own zone — and `override_settings(TIME_ZONE=...)` is per-test, while the `RustLiveView` is session-cached and outlives both. A zone captured at `ready()` or per view instance is stale for exactly the case users care about. Hence `_sync_state_to_rust`, the path that re-runs every event (#1722); a first-sync gate there would pin the zone at mount.

**Thread-local, not process-global**, because djust renders run in `sync_to_async` worker threads and two connections can hold different activated zones concurrently. This is also what Django itself does: `timezone._active` is a `Local()`. `two_threads_hold_independent_zones` pins the property rather than assuming it.

## Timezone format codes

`e T O Z I` had no zone to report, so they fell through to the catch-all and rendered as **their own letter**: `{{ v|date:"H:i T" }}` produced `19:30 T`. Shipping the conversion while `T` still emitted a literal `T` would be a timezone-parity claim that fails on the commonest idiom that shows a timezone. `U` is included because a naive value's epoch is not derivable from its wall clock alone.

## One bug the differential caught that inspection did not

The naive branch left the value stamped with the `+0000` its parse produced, so `O` rendered `+0000` where Django renders `-0400`. **The wall clock was already correct**, which is exactly why reading the code did not surface it — only comparing against Django's own output did. Pinned by `a_naive_value_reports_the_default_zones_offset_not_utc`.

## Verification

- **27 cases** (17 Rust, 10 Python). Every expectation pinned against a live Django 5.2 render of the same value.
- **Gate-off 7/7 red** — every mechanism has a test that fails when only that mechanism is removed. The harness asserts each mutation was *found* and refuses to report a count when the build or collection broke (#2129/#2135); `__pycache__` cleared per iteration.
- Southern-hemisphere DST covered: anchoring `I` on January reports every Sydney instant backwards, so `is_dst` compares against the zone's own minimum offset.
- Half-hour zone (`Asia/Kolkata`, +05:30) covered against whole-hour assumptions.
- An unrecognised `TIME_ZONE` is refused without disturbing the current zone, logged once per process, and renders unconverted rather than 500ing.
- Python **10,298 passed / 0 failed**; Rust green; clippy 0 warnings.

## Deliberately out of scope

The differential left 20 diffs, all in two value shapes that are not datetimes (#1079):

- **#2216** — `{{ v|time:"H:i" }}` on a `datetime.time` returns its input verbatim (`23:30:00`), the same class #2203 fixed for datetimes. And a bare `date` renders midnight where Django raises, which may be correct-as-designed under djust's fail-soft policy but is currently *inherited* rather than decided.
- **#2217** — nine non-timezone format codes (`S w t L b f u c r`) are still unimplemented and render as their own letter; `jS` produces `22S`.

After this PR, **every `datetime` case in the differential matches Django.**


johnrtipton:
## Self-review found a twin the diff did not show

I designed this fix around avoiding #1646 — the PR body argues at length for a shared handoff over a per-`Context` setter — and then created an instance of it anyway. Caught by grepping for the **shape** of a Rust render call rather than for the symbol I had changed, which is the habit I said I'd adopt in #2213's retro. First time it has paid.

**`simple_live_view` renders templates and shares no base class with `RustBridgeMixin`.** It calls `render_template_with_dirs` directly, so a method on the mixin fixed every `LiveView` and left that path rendering UTC.

It is easy to miss for a specific reason: **the class in that module is itself named `LiveView`** — a different one from `djust.LiveView`. Grepping for `SimpleLiveView` returns nothing, so the module reads as unused when it is only unfindable by its obvious name.

Fixed by moving the handoff to `djust.timezone_bridge.apply_active_timezone()`, which both paths call. `test_both_render_paths_call_the_same_timezone_function` walks the package AST and pins the caller **set**, not a floor (#1125) — so a third render path cannot appear unwired, and a private copy in either existing path turns it red. Verified rather than assumed (#1859): gating the mixin's call reddens 7 cases; replacing `simple_live_view`'s call with a private copy reddens the pin.

## The audit also found that path cannot render at all

`get_context_data` walks `dir(self)` and `getattr`s each name, reaching Django's `View.as_view` — a `classonlymethod` that raises `AttributeError` on any **instance**:

```
AttributeError: This method is available only on the class, not on instances.
```

So every render through that module fails before reaching a template, and its `except Exception` turns the crash into a generic "An error occurred rendering this view." It has no tests, is not exported, and its only reference outside itself is `pyproject.toml:330` listing it as a strict-typing island — type-checked but never executed.

Filed as **#2219** with the decision it actually needs (remove it, or fix + test + export under a non-colliding name) rather than a one-line `try/except` patch. The wiring here stays because it is free and correct if the module is revived, and `test_the_simple_live_view_path_is_wired_but_currently_unreachable` **asserts** the `AttributeError` instead of describing it — so that test goes red the moment #2219 is fixed, which is the signal to add the real behavioural case.

## Why the existing 10,288 tests were green on a four-hour error

Worth stating plainly, because "the suite passed" was not evidence of anything here: `examples/demo_project/demo_project/settings.py:115` sets `TIME_ZONE = "UTC"`. Under UTC the conversion is the identity, so **no existing test could have caught this**, and none of them changed when the fix landed. The bug was not under-tested; it was untestable in the configuration the suite runs in.

## Updated verification

- **29 cases** (17 Rust, 12 Python), up from 27.
- **Gate-off 9/9 red**, up from 7/7.
- mypy green on 886 files; Python **10,300 passed / 0 failed**.

johnrtipton:
## Stage 14 — Retro, PR #2218 (#2209)

**Quality: 4/5.** The fix is right and the verification was real. One point off because I built the twin I was writing a design note about avoiding, and only caught it because I ran a grep I had promised myself two PRs ago.

### The premise I had to check before accepting

The issue's own suggested approach rejects the "naive fix" — a per-value `timezone.localtime()` in the serializer — on the grounds that it reintroduces a GIL crossing per rendered datetime. I nearly took that as given. It is wrong: `normalize_django_value` is **already Python**, already calling `.isoformat()` on every datetime, so localizing there costs no new boundary crossing at all.

The reason not to do it is different and better: localizing at *serialization* bakes local time into the **transported state**, which is data the client also reads. Django's localization is a rendering concern, not a data one. So the conclusion held but for a different reason, and that mattered — believing the stated one would have left me thinking the choice was a performance trade-off when it is a correctness one.

Generalising: **an issue's own "why the obvious fix is wrong" is a hypothesis with the same status as its cited code location** (the bug-report triage rule). Check it before building on it.

### What went well

- **Ground truth first.** Every expectation came from running Django 5.2 over 5 value shapes × 26 format codes, before writing any Rust. That table is what caught the naive-`O` bug (rendering `+0000` where Django renders `-0400`) — the wall clock was already correct, so inspection would never have surfaced it, and neither would a test I wrote from my own understanding.
- **Measuring the dependency rather than arguing about it.** The ROADMAP flagged tzdata size as the gating question. Raw is +1.16 MB, which sounds bad; compressed is **+136 KB**, which is what a wheel ships. And I checked whether a plain offset would do before spending it — it would not, because one render can straddle a DST boundary. `winter_and_summer_disagree_within_one_render` exists so a future "optimisation" back to an offset goes red.
- **Southern hemisphere.** Anchoring "is DST" on January is the obvious implementation and reports every Sydney instant backwards. I only thought to check because the DST work had already made me distrust hemisphere-shaped assumptions.

### What didn't

**I wrote the #1646 warning and then committed the violation.** The module docstring argues for a shared handoff over a per-`Context` setter, citing the three-of-seven `set_auto_call` drift by file:line — and the first version put the handoff on a mixin that one of the two render paths does not inherit. The design reasoning was right at the level I was thinking about (`Context` sites) and I never asked the same question one level up (render *paths*).

What found it was grepping for the **shape** — every Rust render call from Python — rather than for the symbol I had changed. That is precisely the correction I wrote down in #2213's retro: *"the neighbour is always something I could have grepped for in under a minute."* It worked. The lesson is not that I need a new insight; it is that the grep has to be a reflex before opening the PR, not a self-review step after.

**"10,288 tests pass" was not evidence.** The suite runs at `TIME_ZONE = "UTC"`, where the conversion is the identity. Not one existing test changed, and not one could have caught a four-hour error. I should have established that *before* treating the green suite as any kind of signal about blast radius — a suite that cannot observe the bug also cannot vouch for the fix.

### Findings this produced

Three issues, all from the differential and the path audit rather than from the reported symptom:

- **#2216** — `{{ v|time:"H:i" }}` on a `datetime.time` returns its input verbatim (`23:30:00`). Same class as #2203, still live for a different type.
- **#2217** — nine non-timezone format codes render as their own letter; `jS` produces `22S`. Invisible to anything but a differential, because rendering an unknown code as itself is also the correct behaviour for a literal.
- **#2219** — `simple_live_view` cannot render **at all**: `get_context_data` raises `AttributeError` on every instance. Unexported, untested, and named `LiveView` so it is unfindable by the name people would grep. Filed with the decision it needs rather than the one-line patch.

### Verified

- 29 cases (17 Rust, 12 Python); every expectation from a live Django render.
- Gate-off **9/9 red**, harness asserting each mutation was found and refusing a count when the build or collection broke (#2129/#2135), `__pycache__` cleared per iteration.
- mypy green on 886 files; clippy 0; Python **10,300 passed / 0 failed**; CI green.

`RETRO_COMPLETE`

PR #2220, opened 2026-08-25
fix(sessions): honour settings.SESSION_ENGINE on synthesized requests (#2210)

### LINKED ISSUE #2210: bug: live_redirect request synthesis hardcodes the DB session backend, ignoring SESSION_ENGINE
## Summary

Two code paths hardcode the **database** session backend when synthesizing a
`request` for `live_redirect` / `url_change`, instead of honouring
`settings.SESSION_ENGINE`. A project configured for a cache-backed (or any
non-DB) session engine gets a DB-backed `SessionStore` on those paths.

## Evidence

```
$ grep -rn "from django.contrib.sessions.backends.db import SessionStore" python/djust/ | grep -v tests
python/djust/runtime.py:3978:            from django.contrib.sessions.backends.db import SessionStore
python/djust/websocket.py:3313:            from django.contrib.sessions.backends.db import SessionStore

$ grep -rn "SESSION_ENGINE" python/djust/
(no matches)
```

Both sites then do:

```python
request.session = SessionStore(session_key=session_key) if session_key else SessionStore()
```

`settings.SESSION_ENGINE` is never consulted anywhere in the package.

## Impact

**Correctness (primary).** With
`SESSION_ENGINE = "django.contrib.sessions.backends.cache"`, session data lives
in the cache and there is no corresponding DB row. The synthesized request gets
a `SessionStore` that reads the `django_session` table, finds nothing, and hands
the view an **empty session** — so anything a sticky view reads from
`request.session` during a `live_redirect` re-auth is silently missing. The same
applies to `cached_db` (stale reads bypassing the cache layer) and
`signed_cookies` (no DB row exists at all, ever).

**Performance (secondary).** It forces a DB round trip on every `live_redirect`
for projects that adopted a cache backend specifically to avoid one.

Note the Channels `SessionMiddlewareStack` populates `scope["session"]`
correctly, honouring `SESSION_ENGINE`. Only the *synthesized-request* path is
wrong, which is why this has gone unnoticed — the common paths are fine.

## Suggested fix

Mirror what `django.contrib.sessions.middleware.SessionMiddleware` itself does:

```python
from importlib import import_module

SessionStore = import_module(settings.SESSION_ENGINE).SessionStore
```

Resolve once at module scope or cache it; the current imports are already
function-local (`# noqa: PLC0415`), so either shape is fine.

**One case needs an explicit decision rather than a mechanical swap:**
`signed_cookies` sessions cannot be written from a WebSocket at all — there is no
HTTP response to set a cookie on. That engine should get a documented, logged
refusal rather than a silent no-op write.

## Test shape

A regression test should override `SESSION_ENGINE` to the cache backend, drive a
`live_redirect`, and assert the synthesized request's session sees a value
written through the cache engine. Gate-off check (#1468): with the fix reverted,
that assertion must fail — a test that only asserts "a session exists" would
pass either way, since the hardcoded DB store also produces a session object.

## Related

Found during the Django-subsystem sweep written up in
`PERFORMANCE_BRAINSTORM.md` §9 (§9.2 B / Idea 14). Sibling finding: #2209
(timezone).

## Verification method

Static — grep across `python/djust/` at `fix-2203-display-django-parity`
(`73cf03db`). Not reproduced at runtime against a cache-backed session engine;
the impact described above is derived from reading the two call sites and
Django's session-engine contract, and should be confirmed with the test shape
above before the fix is considered validated.

johnrtipton:
Closes #2210.

Four places built a `request` for the `live_redirect` / `url_change` paths, and each imported `django.contrib.sessions.backends.db.SessionStore` **directly**. `settings.SESSION_ENGINE` appeared nowhere in the package.

## Reproduced before fixing, and the picture changed

The issue was explicit that its evidence was a grep — *"Not reproduced at runtime… should be confirmed with the test shape above before the fix is considered validated."* Doing that found **two** failure shapes, not the one reported:

| sessions migration | what the view gets |
|---|---|
| **run** | an **empty session**, silently — the filed symptom |
| **not run** | `OperationalError: no such table: django_session` |

The second is the likelier one for exactly the projects affected: **a cache-only project has no reason to run the sessions migration at all.** And because Django's session stores load lazily, construction succeeds and the raise lands wherever the view first *reads* the session — far from the code that built it, which is a much worse debugging experience than an empty dict.

Both are pinned. The second by forbidding database access outright (no `django_db` marker) rather than by dropping a table, which has the side benefit of covering the *performance* half of the issue — the DB round trip a cache-backed project configured its way out of would fail the test.

## Four sites, not two

The issue cited `runtime.py:3978` and `websocket.py:3313`. `testing.py` has two more (`:125`, `:368`), and they matter more than they look: **`LiveViewTestClient` exists to stand in for the production paths**, so a test client on a different session engine than the code it models is its own quiet trap.

All four now resolve through one `djust.utils.build_session_for_request`, the way `SessionMiddleware` itself does. `test_no_module_imports_the_db_session_store_directly` pins the cure rather than this instance (#1646) — four independent copies is what made the bug possible, and a behavioural test on one path cannot see a fifth copy appearing in another.

## `signed_cookies`: not the refusal the issue proposed

The issue asked for *"a documented, logged refusal"* for that engine. Probing what it actually does argued against it:

- The "session key" for `signed_cookies` **is** the signed payload, so **reads work** — strictly better than the empty session it got from the DB store before.
- Only **writes** cannot persist: saving mints a *new* key that only a `Set-Cookie` header can deliver, and a WebSocket has no response to put one on.

A refusal would have discarded the working half to prevent the broken one. So: warn once per process (not once per synthesized request — a misconfigured project would otherwise log on every `live_redirect`) and keep the reads.

## Test quality

Every case asserts a **value written through the configured engine**, never that a session merely exists. The issue's own test-shape note flags exactly why: the hardcoded DB store also produces a session object, so the weaker assertion passes either way. One case asserts the store **class** as well, so a fix that happened to work because both engines were reachable would still fail.

## Verification

- **6 cases**; **4/4 gate-off red**, including a mutation that adds a fifth direct DB import (the structural pin catches it) and one that removes the once-per-process dedupe.
- The DB engine is covered too, so the fix cannot break the majority configuration it replaces.
- mypy green on 886 files; **10,306 passed / 0 failed**.


johnrtipton:
## Stage 14 — Retro, PR #2220 (#2210)

**Quality: 4/5.** Straightforward fix; the value was in refusing to take three of the issue's own statements on trust.

### Three premises checked rather than inherited

The issue is unusually honest — it labels its own evidence as static grep and asks for runtime confirmation before the fix is considered validated. Taking that seriously changed three things:

1. **The symptom.** Filed as "empty session". Reproducing it showed the likelier shape for the affected projects is `OperationalError: no such table: django_session` — a cache-only project has no reason to run the sessions migration. And since Django's stores load lazily, the raise lands wherever the view first *reads* the session, not where it was built. A fix verified only against the empty-session case would look correct while the other kept 500ing.

2. **The site count.** Filed as two. There are four — `testing.py` has two more. That one is worse than it sounds: `LiveViewTestClient` exists to *stand in for* the production paths, so a test client on a different session engine than the code it models is a trap that would quietly validate wrong behaviour.

3. **The `signed_cookies` recommendation.** The issue asks for a logged refusal. Probing the engine showed reads work fine — the "session key" *is* the signed payload — and only writes cannot persist. Refusing would have thrown away the working half to prevent the broken one, and left that engine no better off than the empty session it had. Warn once, keep the reads.

Each of those was a few minutes of running code. All three would have shipped wrong on inspection alone.

### What went well

- **The performance half came free from the right test shape.** Pinning the no-migration case by forbidding DB access outright (rather than by dropping a table) means the same test also fails if the fix reintroduces a round trip. One assertion, two properties.
- **Value assertions, not existence assertions.** The issue warned that "a session exists" passes either way. Every case writes through the *configured* engine and reads that value back; one asserts the store class too, so a fix that worked by accident because both engines were importable still fails.

### What didn't

**I committed impl + tests + CHANGELOG in one commit**, not the two-commit shape (#181). The rule defends against `[Unreleased]` collisions between parallel agents, and this was a single serial agent, so nothing was at risk — but the rule is the rule and I should have split it. Noting rather than amending, because a bounced `--amend` is its own failure class (#1524) and the cost here is zero.

**I found the fourth site by grepping the import string, which only worked because all four spelled it identically.** A site that had written `from django.contrib.sessions.backends import db` would have been invisible to both my grep and the structural test that now guards this. The pin is a string match, and I should be honest that it is a good-enough guard rather than a complete one.

### Verified

- 6 cases; **4/4 gate-off red**, including a mutation that adds a fifth direct DB import (caught by the structural pin) and one that removes the once-per-process dedupe.
- DB engine covered, so the majority configuration cannot regress.
- mypy green on 886 files; **10,306 passed / 0 failed**; CI green.

`RETRO_COMPLETE`

PR #2222, opened 2026-08-25
fix(templates): localize rendered numbers the way Django does (#2221)

### LINKED ISSUE #2221: bug: the Rust engine is locale-blind — numbers ignore USE_L10N/USE_THOUSAND_SEPARATOR, and this diverges in en-us too
Files the `floatformat` / l10n row from the **v1.1.1-2** ROADMAP bucket. That row said the defect is "the same locale-blindness as #2209" and "**probably belongs folded into it** — decide at Stage 4."

**Stage 4 decision: it does not fold in.** Probing it settled that, and also showed the defect is materially wider than the row describes.

## It is not a German-projects problem

The row's example was `LANGUAGE_CODE="de"`. But `USE_THOUSAND_SEPARATOR` applies **regardless of language**, so the divergence is present in the default English configuration:

| template | value | Django (en-us) | djust |
|---|---|---|---|
| `{{ n\|floatformat:2 }}` | `1234.5` | `1,234.50` | `1234.50` |
| `{{ n\|floatformat }}` | `1234.5` | `1,234.5` | `1234.5` |
| **`{{ n }}`** | `1234.5` | `1,234.5` | `1234.5` |
| **`{{ i }}`** | `1234567` | `1,234,567` | `1234567` |

Note the last two: this is not confined to `floatformat`. **Bare `{{ number }}` is affected**, which is a far larger surface than the filed row implies — every rendered number in every template.

Under `de` and `fr` the separators swap as well:

| | en-us | de | fr |
|---|---|---|---|
| `{{ n\|floatformat:2 }}` | `1,234.50` | `1.234,50` | `1 234,50` (NBSP) |
| `{{ i }}` | `1,234,567` | `1.234.567` | `1 234 567` (NBSP) |

`floatformat`'s `u` suffix (`floatformat:"2u"`) is Django's documented opt-out and correctly yields `1234.50` in every locale — which is exactly what djust produces unconditionally today.

## Why it does not fold into #2209

They look like one defect ("the engine is locale- and timezone-blind") and have opposite fix shapes:

- **Timezone** needed a self-contained **database** — `chrono-tz`, +136 KB compressed — and nothing from Python but a zone name. Rust does the work.
- **Locale** needs **Django's own data**: `django/conf/locale/*/formats.py` for separators and date formats, and `.mo` translation catalogs for month and day names. Reimplementing that in Rust is a different order of work, and it would be a *fork* of Django's locale data rather than a use of it.

So the correct shape is the inverse of #2209's: resolve the parameters **in Python** and pass them down per render — separator, decimal point, grouping — which is a handful of strings, not a database. #2209 already built the mechanism for exactly this (`djust.timezone_bridge.apply_active_timezone`, a per-render push into a Rust thread-local, wired at `_sync_state_to_rust` so it re-reads every event); the locale push should reuse that shape rather than invent a second one.

## Three separable pieces, in decreasing value

1. **Number formatting** (separators, grouping) — the filed complaint, diverges in `en-us`, biggest surface, and cheapest: Python passes 3 values per render, Rust formats. **This is the one to do.**
2. **Localized month/day names** — `{{ d|date:"D" }}` gives `Sat` where German Django gives `Sa` and French gives `sam`. Needs catalogs, but the vocabulary is *finite and small* (12 months + 7 days + abbreviations ≈ 40 strings), so it could be pushed per language and cached rather than parsing `.mo` in Rust.
3. **Locale date formats** — bare `{{ d }}` uses each locale's `DATE_FORMAT` (`Aug. 22, 2026` vs `22. August 2026` vs `22 août 2026`). Blocked on a prior gap: djust renders a bare datetime as its raw ISO string today, because the serializer hands Rust a string and the type is lost. That is its own issue, not this one.

Suggest scoping **1** now and splitting 2 and 3.

## Verification

Differential harness, djust's Rust engine vs Django's own, same values and settings, across `en-us` / `de` / `fr`. Ground truth from a live Django 5.2 render, not from the docs.

## Related

#2209 (timezone — shipped, PR #2218) built the per-render ambient-setting mechanism this should reuse. #2217 (unimplemented format codes) is adjacent but independent. The `{% trans %}` row in the same ROADMAP bucket is the i18n half and needs the catalog lookup that piece 2 above also needs.

johnrtipton:
Closes #2221 (piece 1 of 3).

Django localizes a number on its way into the page; the Rust engine used Rust's defaults.

## Two widenings that change who is affected

The ROADMAP row this came from framed it as `floatformat` under `LANGUAGE_CODE="de"`. Probing against Django's own engine widened it twice:

**It is not a non-English problem.** `USE_THOUSAND_SEPARATOR` applies regardless of language:

| | Django (en-us) | djust before |
|---|---|---|
| `{{ i }}` for `1234567` | `1,234,567` | `1234567` |
| `{{ n }}` for `1234.5` | `1,234.5` | `1234.5` |

**It is not confined to `floatformat`.** Bare `{{ n }}` is affected — every rendered number in every template.

All three locales now match:

| | en-us | de | fr |
|---|---|---|---|
| `{{ n\|floatformat:2 }}` | `1,234.50` | `1.234,50` | `1 234,50` (U+00A0) |
| `{{ i }}` | `1,234,567` | `1.234.567` | `1 234 567` (U+00A0) |

## The fix shape is the inverse of #2209's

The ROADMAP asked explicitly whether this folds into the timezone fix, and this was the load-bearing decision. **It does not:**

- **Timezone** needed a self-contained **database** in Rust (`chrono-tz`, +136 KB) and nothing from Python but a zone name. Rust does the work.
- **Locale** is defined by `django/conf/locale/*/formats.py`. Deriving it in Rust would be a **fork of Django's data** rather than a use of it. So Python resolves three values per render and Rust only applies them — reusing the per-render push #2209 built, now renamed `djust.render_env.apply_render_env` since it carries two ambient settings and the old name would have lied.

## Applied at the output site, not in `Display`

`impl Display for Value` is where number rendering lives and is the obvious place to put this. It is **also the lookup key for `{% if x in dict %}`** (#2203) — so a separator there turns the key `1234567` into `1,234,567` and silently misses every such lookup against a dict Python keyed without one. The failure would be invisible: the template just renders the else-branch.

The localization goes at the variable-**output** site instead, which is also where Django does it (`render_value_in_context` calls `localize`). `test_localization_does_not_reach_dict_lookup_keys` pins the hazard rather than leaving it as a comment.

**Two byte-identical output sites exist** — `Node::Variable` and the inline-if expression. Found by counting matches rather than reading the diff; both go through one helper (#1646).

## Django's interval walk, ported rather than assumed

Grouping is not "every three digits". Django consumes an interval list, and `intervals.pop(0) or active_interval` means a `0` entry **keeps the previous width** rather than ending grouping:

- Indian grouping `[3, 2, 0]` → `12,34,567`, not `1,234,567`.
- A naive group-by-three loop passes every other test in the suite, which is why that case is pinned.

## Two settings behaviours worth stating

- **`USE_L10N` is deliberately not read.** Verified inert across the full `USE_L10N` × `USE_THOUSAND_SEPARATOR` × language matrix — flipping it changed no output. Django 5.0 removed it as a toggle, so reading it would be cargo cult.
- **`USE_THOUSAND_SEPARATOR=False` still localizes the decimal point.** It suppresses *grouping* only; German renders `1234,5`. Easy to get wrong in the direction of doing nothing.
- `floatformat:"2u"` remains Django's documented opt-out — and produces exactly what djust produced for every locale before this.

## Verification

- **23 cases** (12 Rust, 11 Python); Python cases all go through the real `LiveView.render()`.
- **Gate-off 8/8 red**, harness asserting each mutation was found and refusing a count when the build or collection broke (#2129/#2135).
- French's U+00A0 asserted as the actual codepoint — a test written with a plain space would pass while shipping the wrong byte into every French page.
- mypy green on 888 files; clippy 0; **10,316 passed / 0 failed**; Rust green.

## Still open on #2221

Pieces **2** (month/day names — `{{ d|date:"D" }}` gives `Sat` where German Django gives `Sa`) and **3** (per-locale `DATE_FORMAT`, blocked on bare datetimes rendering as raw ISO strings). Both need translation catalogs rather than format parameters, which is why they are separable.


johnrtipton:
## Stage 14 — Retro, PR #2222 (#2221, piece 1 of 3)

**Quality: 4/5.** The scoping was the valuable part; the implementation was mostly transcription once the ground truth existed.

### Scoping was the work

The ROADMAP row asked one question — *does this fold into #2209?* — and answering it properly changed three things:

1. **The fold answer is no, and for a reason that generalises.** Both defects read as "the engine is locale- and timezone-blind", one sentence, one fix. They have **opposite** shapes: timezone needed a self-contained database *in* Rust and nothing from Python; locale is defined by Django's own `conf/locale/*/formats.py`, so putting it in Rust would fork that data rather than use it. Folding them would have produced one mechanism doing two contradictory things. The generalisable form: *"same symptom class" is not evidence of "same fix shape" — check which side owns the data.*

2. **Who is affected changed.** Filed as a German-locale `floatformat` bug. It is a **default-English, every-number** bug: `USE_THOUSAND_SEPARATOR` ignores language, and bare `{{ n }}` is affected. That moved it from a P2 nicety to something every project renders on every page.

3. **A flag I would have read is inert.** `USE_L10N` reads like the master switch. Running the full `USE_L10N` × `USE_THOUSAND_SEPARATOR` × language matrix showed flipping it changes **nothing** — Django 5.0 removed it as a toggle. Reading it would have been cargo cult that looked like diligence, and no test would have caught it.

### The design decision I am most glad I checked

`impl Display for Value` is where number rendering lives, and putting the separator there is the one-line fix. It is also the lookup key for `{% if x in dict %}`, so it would have turned `1234567` into `1,234,567` and **silently missed** every such lookup — the template just renders the else-branch, nothing raises. I only thought to check because #2203's own notes (which I wrote) flagged `Display`-as-key. That is the second time this week those notes have paid, which is an argument for writing the hazard down at the time rather than trusting recall.

Now pinned by `test_localization_does_not_reach_dict_lookup_keys` rather than left as a comment.

### What went well

- **Ground truth before code, again.** The Django matrix across three locales came first; the Rust is a transcription of `numberformat.py`. Indian grouping (`[3, 2, 0]` → `12,34,567`) is in the suite because I read Django's interval walk instead of assuming groups of three — a naive loop passes every other case in the file.
- **Asserting the actual codepoint.** French uses U+00A0. A test written with a plain space passes while shipping the wrong byte into every French page.
- **Counting matches, not reading the diff.** The renderer has two byte-identical output sites. `s.count(old) == 1` failed loudly and told me; reading the patch would not have.

### What didn't

**I renamed a module I shipped ninety minutes earlier.** `timezone_bridge.py` → `render_env.py` was right — it now carries two settings and the old name lies — but it is churn I could have avoided by naming it `render_env` when I created it, since the ROADMAP already said a locale handoff was coming in the same bucket. I named for the one caller in front of me instead of the two the plan listed.

**I still have not wired the standalone `render_template` path.** It picks up the thread-local from an enclosing LiveView render, so it works in practice, but a direct call on a fresh thread gets no locale and no timezone. I noted this during #2209 and did not act on it then either. It is now a known third path with no owner — I should have filed it rather than carrying it in my head across two PRs.

### Verified

- 23 cases (12 Rust, 11 Python); Python cases all through the real `render()`.
- **Gate-off 8/8 red**, harness asserting each mutation was found and refusing a count when the build or collection broke.
- mypy green on 888 files; clippy 0; **10,316 passed / 0 failed**; CI green.

`RETRO_COMPLETE`

Issue #2214, opened 2026-08-25
bug: DecimalField reaches the client as a lossy float — the Decimal branch is dead code


johnrtipton:
Found by the Stage 11 review of PR #2213 and split out to keep that PR scoped (#1079). Same shape as #2212, same function, different type pair — the next link of that chain (#2142).

## The bug

`serialize_python_value` (`crates/djust_live/src/lib.rs`) has an explicit branch intending to stringify Django types:

```rust
if type_name == "Decimal" || type_name == "UUID" {   // ~:2400
```

It is **dead for `Decimal`**, because `extract::<f64>()` runs first (~`:2364`) and PyO3's `f64` extraction goes through `PyFloat_AsDouble`, which honours `Decimal.__float__`. So every `Decimal` is converted to a binary float before the branch is reached.

Measured against a current build:

| input | reaches the client as | |
|---|---|---|
| `Decimal('19.99')` | `19.99` *(float)* | representation error |
| `Decimal('0.1')` | `0.1` *(float)* | representation error |
| `Decimal('12345678901234567890.123456789')` | `1.2345678901234567e+19` *(float)* | **precision destroyed** |

`UUID` is unaffected — it is not float-convertible, so it reaches the branch.

## Why it matters

`serialize_context` / `serialize_queryset` are public `#[pyfunction]`s feeding JIT state serialization, so this is the path a `DecimalField` takes to the browser. `DecimalField` is what Django projects use for **money**, and binary floats are precisely what it exists to avoid. A price of `19.99` arriving as a float is the textbook case against it.

The function's own doc-comment (~`:2271`) states that Django types *"(datetime, Decimal, UUID): convert to strings"* — a prose invariant that is false about the code it sits on (#1867).

## The shape worth noticing

This is the **same class as #2212**: a permissive `extract::<T>()` placed above a narrower special case, making the special case unreachable. The compiler cannot see it (the arms have different types, so no `unreachable_patterns`), and clippy does not flag it.

The structural guard added in PR #2213 (`python/tests/test_bool_before_int_converters_2212.py`) sweeps for the **i64/bool** pair specifically and is blind to this one.

## Suggested fix

1. Move the `type_name == "Decimal" || type_name == "UUID"` check **above** the numeric extractions, so the intent in the code is the behaviour.
2. Add a behavioural case pinning `Decimal('19.99')` → `'19.99'` *(str)*, including a value that cannot survive a float round-trip so the test cannot pass by luck.
[...]

PR #2224, opened 2026-08-25
test(serialization): guard the dead-special-case class behind #2212 and #2214

johnrtipton:
Refs #2214 — **deliberately does not close it.** See "why" below.

## The class

`serialize_python_value` has a branch intending to stringify `Decimal` and `UUID`. It is **dead for `Decimal`**: `extract::<f64>()` above it honours `Decimal.__float__`.

| input | reaches the client as |
|---|---|
| `Decimal('19.99')` | `19.99` *(float)* |
| `Decimal('12345678901234567890.123456789')` | `1.2345678901234567e+19` |
| `UUID(int=1)` | `'000…001'` *(str — correct)* |

`UUID` reaching the branch correctly is what makes this a **shadowing** bug rather than a missing branch.

Same shape as #2212 — *a permissive extraction above a narrower special case* — and invisible to every tool the repo runs: not a compile error (the arms have different types, so not `unreachable_patterns`), not a clippy lint. Both verified against mutated builds.

## Why #2214 is not fixed here

The issue suggests moving the branch above the numeric extracts. **Measured, that regresses two template behaviours**, because this value goes back into the *template context* (`context[key] = serialized` in `mixins/context.py`) and not only onto the wire:

| template | Django | djust (float, today) | djust (str, "fixed") |
|---|---|---|---|
| `{{ p\|floatformat }}` | `20.0` | `20.0` | **`19.99`** ✗ |
| `{% if p > 10 %}` | `BIG` | `BIG` | **`small`** ✗ |
| `{{ p }}` | `19.99` | `19.99` | `19.99` ✓ |

So the one-line fix trades a precision bug for two behaviour bugs. It needs a **decision** — a `Decimal`-aware `Value` variant that keeps the exact string *and* numeric semantics, or an accepted precision limit — not a patch. Evidence posted on the issue.

Worth noting the issue's severity framing cuts both ways: `19.99` renders *identically* today, so "a price arrives as a float" overstates the common case; the real damage is beyond ~15 significant digits, which understates it for large `DecimalField`s.

## What ships: the half the issue itself called more valuable

A structural sweep for the **general rule** — *no `type_name ==` special case below an extract that swallows it* — replacing #2212's i64/bool-specific sweep, which was blind to this instance.

**The capture table is measured, not declared.** For each type the Rust source special-cases, an instance is built and the coercion PyO3 actually performs is run. Hardcoding it as a comment is precisely how the original bug shipped: the code *says* `Decimal` converts to a string.

That measurement caught an error in my own first draft. **PyO3's `i64` goes through `__index__`, not `int()`:**

| | `int()` | `operator.index()` |
|---|---|---|
| `Decimal('19.99')` | `19` | `TypeError` |
| `UUID(int=1)` | `1` | `TypeError` |
| `True` | `1` | `1` |

Modelling `i64` as `int()` reported two perfectly reachable branches as dead. A guard that cries wolf is worse than no guard, because a false alarm trains people to ignore the real one.

## Empirical canaries (#1459)

A structural lint that reports nothing is indistinguishable from a broken one, so three canaries prove this one is load-bearing:

| canary | result |
|---|---|
| **A** — fix the bug (hoist the branch) | xfail **XPASSes**, set-pin changes → the tripwire fires |
| **B** — inject a *second* instance | set-pin goes **red naming it** |
| **C** — a non-float-convertible special case | **not reported** — no false alarm |

**B is the one that justifies the set-pin's existence.** An `xfail` is satisfied by *one* failure, so a new dead branch joining the known one would have hidden behind it and stayed green.

## Also: three prose claims that were false about their own code (#1867)

- Rust's doc-comment: *"Django types (datetime, Decimal, UUID): convert to strings"* — not for `Decimal`.
- `serialization.py:1020`: `# Decimal -> float (matches DjangoJSONEncoder.default)` — Django's encoder returns `str(o)` with full precision, so the comment claimed parity with the very thing it contradicted.
- `serialization.py:318`: same divergence, undocumented.

All three now state what the code does, and name the tracked decision.

## Verification

5 cases; 3 canaries; **10,321 passed / 0 failed / 1 xfailed** (the tripwire); clippy 0.


johnrtipton:
## Stage 14 — Retro, PR #2224 (refs #2214)

**Quality: 4/5.** The right call was *not* doing what the issue asked, and being able to show why.

### The judgement this PR turned on

The issue proposes a one-line fix and reads as obviously correct: a branch is dead, hoist it. I measured it before writing it, and it regresses `{{ p|floatformat }}` and `{% if p > 10 %}` — because `serialize_python_value`'s output goes back into the **template context**, not only onto the wire. The issue's own scope sentence (*"the path a DecimalField takes to the browser"*) is what made the trap easy to miss; it names one consumer of a value that has two.

Shipping the "fix" would have closed the issue, looked productive, and traded a precision bug nobody had reported for two behaviour bugs in common idioms.

The generalisable form: **a proposed fix inside an issue has the same epistemic status as its cited code location** — a hypothesis. The triage rule already says that about *where* the bug is; it applies just as much to *what to do about it*. Three issues in this drain have now had a wrong premise, and this is the first where the wrong premise was the remedy rather than the diagnosis.

### What went well

- **The guard's premise is measured, not declared.** Hardcoding "does `extract::<f64>()` swallow a Decimal?" as a comment is *exactly* how the original bug shipped — the code says Decimal converts to a string. Making the guard build a specimen and run the coercion means a future Python that changes `Decimal.__float__` moves the table instead of silently voiding a comment.
- **That measurement caught my own error.** I modelled PyO3's `i64` as `int()`. It goes through `__index__`. `int(Decimal("19.99"))` is 19 and `int(UUID(int=1))` is 1, so my first draft reported two perfectly reachable branches as dead. A guard that cries wolf is worse than no guard — it trains people to ignore the real alarm. I would not have caught this by inspection; the sweep produced two rows I did not expect and I had to go find out why.
- **Canary B earned the set-pin.** An `xfail` is satisfied by *one* failure, so a second dead branch would have hidden behind the known one and stayed green. Injecting one proved the set-pin catches it. I nearly shipped the xfail alone.

### What didn't

**I wrote the specimen table before checking what PyO3 actually does**, then justified it in a comment. The comment was confident and wrong. The order should have been: measure, then write the comment describing the measurement. That is the same inversion as the false `DjangoJSONEncoder` comment this PR corrects — asserting parity with something instead of checking it — and I reproduced it in the very file whose purpose is to prevent that.

**"Refs, not closes" leaves the drain with an open item.** That is the honest outcome, but it means #2214 now needs a decision from someone, and I have put three options on it without picking one. For options 1 and 2 that is right (real work vs. an accepted limitation). I could have been more decisive about ruling *out* option 3, which makes djust guess at `{% if "abc" > 10 %}`.

### Verified

- 5 cases; 3 empirical canaries (fix → XPASS; second instance → set-pin red; non-float-convertible → silent).
- Three false prose claims corrected (#1867), behaviour untouched.
- **10,321 passed / 0 failed / 1 xfailed** (the tripwire); clippy 0; CI green.

`RETRO_COMPLETE`

PR #2225, opened 2026-08-25
fix(templates): push per-render settings from the Django-template backend (#2223)

### LINKED ISSUE #2223: bug: the standalone render_template() path never pushes per-render settings, so a direct call renders unlocalized
Noted during #2209 and again during #2221 without being filed either time. Filing it now rather than carrying it a third time.

## The gap

`djust.render_env.apply_render_env()` pushes the active timezone (#2209) and number format (#2221) into Rust thread-locals. Exactly two callers do so, both pinned by `test_both_render_paths_call_the_same_timezone_function`:

- `mixins/rust_bridge.py` — every `LiveView`, via `_sync_state_to_rust`
- `simple_live_view.py`

**The standalone `_rust.render_template()` / `render_template_with_dirs()` functions are a third path and push nothing.** They are the entry point for component rendering and template tags.

## Why it mostly works today, and when it does not

A direct call almost always runs *nested inside* a LiveView render on the same thread, so the thread-local is already set by the enclosing render and the result is correct. It breaks in two cases:

1. **A fresh thread that has never rendered a LiveView.** No timezone, no number format — timestamps render in UTC and numbers render unseparated, while the rest of the same application renders them correctly.
2. **A worker thread reused after a render with different settings.** The thread-local persists, so the call inherits whatever the previous render on that thread set. That is *usually* right and occasionally stale.

The failure is silent in both cases and produces output that is wrong in a way that looks plausible.

## Reproduction

Confirmed while building the #2221 differential: calling `render_template('{{ n }}', {"n": 1234567})` on a fresh interpreter returns `1234567`, while the same template through `LiveView.render()` returns `1,234,567`. Adding an explicit `apply_render_env()` before the call makes them agree — that is how the differential harness was made to pass.

## Why it was not fixed in either PR

Scope (#1079). #2209 and #2221 each fixed the path their issue cited, and adding a third would have widened two already-large PRs. But two PRs in a row have now noted it and moved on, which is how a known gap becomes a forgotten one.

## Suggested fix

Push the environment inside the PyO3 `render_template` / `render_template_with_dirs` functions themselves, so the entry point cannot be used without it — rather than adding a third caller that the structural pin then has to whitelist. That is a Rust→Python call per standalone render, so it wants measuring against the nested case, which is by far the common one; a cheap alternative is to make the Python-side wrappers push before delegating.

Whichever shape, the structural test's expected caller set needs updating in the same change.

## Related

#2209 (timezone), #2221 (number format) — both shipped, both pinned to the two-caller set this issue widens.

johnrtipton:
Closes #2223.

#2209 (timezone) and #2221 (number format) each wired their setting into Rust from the two Python render paths a structural test pinned. **That set was wrong.**

A plain Django template rendered through `DjustTemplateBackend` goes through `template/rendering.py`, which pushed nothing:

```
djust backend, fresh thread : 1234567|23:30
Django renders              : 1,234,567|19:30
```

So the same page could render a number correctly inside a LiveView and incorrectly in a template beside it, with nothing raised.

## Why it survived two PRs

**The thread-local persists.** A worker that has already served a LiveView render carries a correct environment into any later backend render on the same thread, which hides the bug most of the time. It shows on a thread that has *not* — the first request a worker handles, or a process whose traffic is all plain templates.

That is also why it did not show up as a test failure: by the time anything exercised the backend, something else had usually set the environment.

## The structurally better fix was measured and rejected

Pushing inside `_rust.render_template*` makes the entry point impossible to misuse, which is the shape I would prefer. It costs:

| | |
|---|---|
| `apply_render_env()` | **~12 µs** |
| a small `render_template` | ~15 µs |
| overhead if pushed per call | **~78%** |

And most calls are **nested** component renders that already inherit a correct thread-local from their enclosing render — so the overhead buys nothing on the common path. Top-level entries pay it once; nested ones inherit. The pinned caller set now records which nested paths are *deliberately* excluded, so the exclusion is a decision rather than an omission.

## Two of my own test claims were wrong, and gate-off said so

**1. The `_fresh_thread` fixture is not what lets these tests see the bug.** I wrote in the docstring that it was — that without it the tests would pass against broken code. Gating the fix off *with the fixture neutered* still fails, because `override_settings` moves the timezone and language away from whatever a previous test pushed, so a stale environment does not match anyway. Kept for order-independence — hygiene, not a detector — and now labelled as such.

**2. One of the three cases cannot detect this defect at all.** `test_the_backend_leaves_values_alone_when_the_settings_say_to` passes with the fix gated off, because with `USE_TZ` off and no thousand separator, **doing nothing is the right answer**. It is kept as a guard against an *unconditional* conversion — a different defect — and labelled so it is not counted as coverage of #2223.

Both corrections went into the test file, not only the PR description: a guard nobody has measured is how #1859 happens, and a module whose purpose is pinning this class should not itself carry an unmeasured claim.

## Verification

- 3 cases, of which **2 detect the bug** (verified by gating the fix off, not asserted).
- The pinned caller set grows 2 → 3 and documents the exclusions.
- **10,324 passed / 0 failed / 1 xfailed**.


johnrtipton:
## Stage 14 — Retro, PR #2225 (#2223)

**Quality: 4/5.** A small fix whose value was almost entirely in the two things I got wrong and measured rather than shipped.

### The finding behind the finding

I filed #2223 during #2221's retro, describing it as "the standalone `render_template` path". Tracing the importers showed that framing was wrong in a way that mattered: the real gap is `template/rendering.py` — the **Django-template backend**, a top-level render path serving every plain template — while the `_rust.render_template` calls I had actually noticed are mostly *nested* and inherit a correct environment. I had filed the symptom I happened to trip over rather than the path that matters.

### Two of my own claims failed their gate-off

This is the part worth keeping.

**1. I wrote that the `_fresh_thread` fixture was what let the tests see the bug.** It reads plausibly — the thread-local persists, so a stale environment should mask the defect. Gating the fix off *with the fixture neutered* still fails: `override_settings` moves the timezone and language away from whatever a previous test pushed, so a stale value never matches anyway. The fixture is hygiene, not a detector.

**2. One of three cases cannot detect this defect at all.** With `USE_TZ` off and no thousand separator, doing nothing produces the right answer, so it passes with the fix gated off. It is a real guard — against an *unconditional* conversion — but it is not coverage of #2223.

I only found both because I ran a combined gate-off (fix off **and** fixture neutered) expecting green, and got red. The prediction failing is what exposed the claim. **Predicting the gate-off result before running it turns it from a pass/fail check into an actual test of my model** — that is the transferable bit, and it cost one extra command.

Both corrections went into the test file, not just here. A module whose purpose is pinning this class should not itself carry an unmeasured claim (#1859).

### What went well

- **Measuring the design alternative instead of arguing it.** Pushing inside the PyO3 entry point is the structurally better shape — impossible to misuse — and at ~12 µs against a ~15 µs render it is ~78% overhead on a path that is mostly nested calls already carrying the right environment. The pinned caller set now records *which* nested paths are excluded, so it reads as a decision rather than an omission.
- **The structural pin earned its keep for the second time.** It went 2 → 3 because a real path was missing, which is exactly what a set-pin is for and what a floor would have let through.

### What didn't

**This is the third PR in a row where the pinned caller set was incomplete when I shipped it.** #2209 pinned two after finding a second path in self-review; #2221 inherited that set; #2223 found the third. Each time I enumerated the paths I knew about and treated the enumeration as complete. The pin catches the drift *after* the fact, which is worth having — but the pattern says my initial sweep for "what are all the paths that do X" is reliably one short. A one-line `grep -rn` for the *sink* (`_render_fn`, `render_template`) rather than for the callers I expect would have found `template/rendering.py` in #2209.

### Verified

- 3 cases, **2 of which detect the bug** — verified by gating the fix off, not asserted.
- Reproduction on a genuinely fresh thread: `1234567|23:30` → `1,234,567|19:30`.
- **10,324 passed / 0 failed / 1 xfailed**; CI green.

`RETRO_COMPLETE`

PR #2226, opened 2026-08-25
fix(templates): format a bare datetime.time instead of echoing it (#2216)

### LINKED ISSUE #2216: bug: date/time filters on a bare date or time object diverge from Django

johnrtipton:
Closes #2216 (the `time` half; the `date` half is deliberately unchanged — see below).

No parse branch matched a time-only string, so `date`/`time` returned the serialized value **verbatim**:

| template | Django | djust before |
|---|---|---|
| `{{ v\|time:"H:i" }}` | `23:30` | `23:30:00` |
| `{{ v\|date:"P" }}` | `11:30 p.m.` | `23:30:00` |
| `{{ v\|date:"g:i A" }}` | `11:30 PM` | `23:30:00` |

Exactly the class #2203 fixed for datetimes, still live for a different type: the format list carried four datetime shapes and one date-only shape and **no time-only shape at all**.

## Why it hid

For `H:i:s` the echoed input **equals** the correct output — so the most obvious test anyone would write passes against the broken code. That case is kept in the suite, labelled as a reminder rather than counted as coverage.

## Three rules that look alike

Enumerated by running all 38 format characters against a `datetime.time` through Django's own engine, because they do not follow from the docs:

| codes | on a bare time |
|---|---|
| `a A c f g G h H i P s u` | formatted normally |
| `e T O Z` (timezone) | **empty in place** — the rest of the format still renders |
| date codes | **the entire render is empty** |

Conflating the last two gives `'23:30 '` where Django gives `''`.

And the timezone rule **differs from the naive-datetime rule one line away**, where the default zone *is* reported (#2209):

- suppressing both → silently undoes #2209 for every naive datetime
- suppressing neither → invents a zone from the epoch date a bare time is anchored on (`23:30 UTC`)

Pinned in both directions, and both gate-off verified. An escaped `\Y` stays a literal and does not empty the render.

## Also: lowercase `a`

Django's `a` is `a.m.`/`p.m.`; only uppercase `A` is bare `AM`/`PM`. djust emitted `am`/`pm` for both. Outside this issue's scope strictly speaking — the same differential found it **on datetimes as well as times**, so it is not a time-only defect — but it is two lines in the match arm being edited, and shipping "time filters now match Django" alongside a wrong `a` in that same arm would be odd. Flagging it explicitly rather than burying it.

## Deliberately unchanged

The bare-`date` half. djust renders midnight where Django raises `TypeError`, and **Rust cannot tell a `date` from a midnight datetime** because the serializer discards the type. That wants a documented decision about djust's fail-soft policy — consistent with the `{% url %}` row in the same ROADMAP bucket — not a patch.

## Verification

- **16 cases** (12 Rust, 4 Python); **7/7 gate-off red**, including both directions of the timezone rule and a mutation that makes the date-code guard fire per-code instead of whole-render.
- The **Python** cases exist for one thing the Rust table cannot check: that the *serializer* emits the shape the new parse branch accepts. A filter-level test takes a string and would stay green if `normalize_django_value` sent something else.
- Differential across 4 time values × 17 format strings: **64/68 → 68/68**.
- clippy 0; **10,327 passed / 0 failed / 1 xfailed**.


johnrtipton:
## Stage 14 — Retro, PR #2226 (#2216)

**Quality: 4/5.** The differential did the work; my first implementation had two defects and it found both.

### The bug that best explains why differentials matter

`{{ v|time:"H:i:s" }}` on a `datetime.time` renders `23:30:00` — **correctly, and also when completely broken**, because the echoed input equals the correct output. That is the assertion I would have reached for first, and it certifies nothing. Only comparing against Django across a *range* of format strings separated the two.

I kept that case in the suite rather than deleting it, labelled as a reminder. A test that cannot fail is normally worth removing; this one documents the trap that made the bug survivable, which is worth more than the line it costs.

### Two defects in my own fix, both caught by the differential

**1. `T` rendered `UTC` for a bare time.** I added a `time_only` flag and then never used it to suppress the timezone codes, so the epoch date the value is anchored on leaked a zone into the output. Django gives `'23:30 '`.

**2. The date-code guard and the timezone codes are different rules** and I nearly wrote them as one. A date code empties the *whole* render; a timezone code empties *in place* and leaves the rest formatting. `{{ v|date:"H:i Y" }}` is `''` but `{{ v|date:"H:i T" }}` is `'23:30 '`. Two behaviours that look identical from the docs.

Neither was reasoning I could have got right by reading — I enumerated all 38 format characters against a `datetime.time` through Django's engine, which is the only reason the three-way split is in the test file as fact rather than as my guess.

### The regression I went looking for because of #2209

Suppressing timezone codes for a bare time is *one line away* from suppressing them for a naive datetime — where #2209 established the default zone **is** reported. Getting that wrong would have silently undone last-hour's fix for every naive datetime. Pinned in **both** directions and gate-off verified both ways, because a one-directional pin here is worth very little.

### What went well

- **The Python cases have a reason to exist.** The Rust table hands the filter a string, so it would stay green if `normalize_django_value` serialized a `time` differently. The four Python cases check exactly that seam and nothing else.
- **Scope held where it should and moved where it shouldn't have.** The bare-`date` half stays unfixed — Rust cannot distinguish a `date` from a midnight datetime, so it needs a policy decision. Lowercase `a` (`am` vs Django's `a.m.`) I fixed, and said so loudly: two lines in the arm I was editing, verified on datetimes too, and leaving it would have made the PR's own claim false.

### What didn't

**My first edit landed in the wrong function.** A `python3 .replace(..., 1)` targeting a `map_err` line hit `format_timesince` instead of `format_date`, because the same line appears three times. The compiler caught it immediately, so it cost a minute — but the same technique with a *non*-compiling target would have silently edited the wrong place. When a replacement string is not unique, `count()` it first. I have the habit for anchors and not for incidental lines.

Noted in passing: `format_timesince` and `format_timeuntil` still parse **RFC3339 only**, so they carry the same #2203 shape for naive datetimes. Not touched here, not filed yet — flagging so it does not evaporate.

### Verified

- 16 cases (12 Rust, 4 Python); **7/7 gate-off red**, including both directions of the timezone rule.
- Differential across 4 time values × 17 format strings: **64/68 → 68/68**.
- clippy 0; **10,327 passed / 0 failed / 1 xfailed**; CI green.

`RETRO_COMPLETE`

Issue #2228, opened 2026-08-25
bug: timesince/timeuntil output shape diverges from Django on every input — one unit, plain space, and seconds


johnrtipton:
Found by the differential built for #2227 and split out to keep that PR scoped (#1079). #2227 fixed the **parse**; this is the **output**, and it diverges even on the values that always parsed.

## Three divergences, measured against Django 5.2

```python
from django.utils.timesince import timesince
base = datetime(2026, 1, 1)
```

| elapsed | Django | djust |
|---|---|---|
| 30 s | `0\xa0minutes` | `30 seconds` |
| 90 s | `1\xa0minute` | `1 minute` (but plain space) |
| 2 h 30 m | `2\xa0hours, 30\xa0minutes` | `2 hours` |
| 3 d 5 h | `3\xa0days, 5\xa0hours` | `3 days` |
| 10 d | `1\xa0week, 3\xa0days` | `1 week` |
| 40 d | `1\xa0month, 1\xa0week` | `1 month` |
| 400 d | `1\xa0year, 1\xa0month` | `1 year` |

**1. Separator.** Django joins the number and unit with **U+00A0 NO-BREAK SPACE**, so the pair never wraps across a line. djust uses an ordinary space. This affects *every* render, including the ones that were already correct, and a test written with a plain space passes while shipping the wrong byte.

**2. Depth.** Django shows up to **two adjacent** units (`depth=2` by default), joined by `", "`. djust shows one, so it discards the more precise half of almost every answer.

**3. Seconds.** Django's smallest unit is the minute — anything under a minute is `0 minutes`. djust reports seconds, so a fresh value reads `30 seconds` where Django reads `0 minutes`.

## Not just cosmetics

`{{ post.published|timesince }} ago` is the canonical use, and `3 days` vs `3 days, 5 hours` is a visible difference on every page that uses it. The NBSP is the kind of thing that shows up as a bad line break in production and is very hard to trace back.

## Implementation note worth having before starting

Django's month and year arithmetic is **calendar-aware**, not a fixed number of seconds: `django/utils/timesince.py` walks actual calendar months rather than dividing by 2629746. djust's current implementation divides by fixed constants, so it drifts from Django near month and year boundaries even once depth and separator are fixed.

[...]

PR #2229, opened 2026-08-25
fix(templates): one shared datetime parser for timesince/timeuntil (#2227)

### LINKED ISSUE #2227: bug: timesince/timeuntil parse RFC3339 only, so a naive datetime returns its input verbatim
Noted during PR #2226 (#2216) and filed rather than left in a retro, since that is how #2223 nearly evaporated.

## The gap

`format_timesince` and `format_timeuntil` (`crates/djust_templates/src/filters.rs`) parse with `DateTime::parse_from_rfc3339` and **nothing else**:

```rust
let dt = DateTime::parse_from_rfc3339(datetime_str)
    .map_err(|e| DjangoRustError::TemplateError(format!("Invalid datetime format: {e}")))?;
```

`format_date` grew four datetime shapes in #2203 and a time-only shape in #2216, precisely because a Python `datetime` does not arrive as RFC3339 — a naive one serializes to `2026-08-22T23:30:00`, with no offset, which is not RFC3339. These two filters never got that treatment.

So `{{ post.created|timesince }}` on a naive datetime — the normal shape under `USE_TZ = False`, which is a supported configuration — fails to parse.

## Third instance of the same class

- #2203 — `date`/`time` returned their input verbatim for datetimes.
- #2216 — `date`/`time` returned their input verbatim for a bare `time`.
- this — `timesince`/`timeuntil` do it for naive datetimes.

Each time the parse-shape list was extended for the filter in front of us and the neighbouring filters with their own parse were not checked. That is the parallel-path-drift rule (#1646) on the datetime-parsing axis, and the structural cure is one shared parse helper rather than a third correct copy.

## Worth confirming before fixing

Not reproduced end-to-end — the failure mode depends on how the error propagates (`?` on a `TemplateError`), and it may surface as a 500 rather than as verbatim output. Reproduce first, per the repo's own triage rule, and note which it is: a 500 is louder and easier to notice than the silent echo the sibling issues had, which would change the severity.

## Suggested fix

Extract the parse from `format_date` into one helper both call. That retires the class instead of adding the third copy, and it means the next value shape is fixed once. Watch that `timesince`/`timeuntil` need the *instant*, so the time-only branch (which anchors on an arbitrary epoch date) must not be reachable from them — a bare `time` has no instant to measure elapsed time against, and Django raises there.

## Related

#2203, #2216 (both shipped), #2217 (unimplemented format codes in the same function).

johnrtipton:
Closes #2227. Third link of the #2203 → #2216 → #2227 chain.

`timesince` and `timeuntil` called `DateTime::parse_from_rfc3339` and nothing else, so a **naive** datetime — the normal shape under `USE_TZ = False` — did not parse and the filter returned its input verbatim:

```
Django:  2 hours
djust:   2026-08-25T12:16:36.074891
```

A `DateField` was equally affected, and it is one of `timesince`'s commonest inputs.

## Third instance of one class in three releases

| | |
|---|---|
| #2203 | `date`/`time` echoed **datetimes** |
| #2216 | `date`/`time` echoed a bare **time** |
| #2227 | `timesince`/`timeuntil` echo **naive datetimes** |

Each was found by fixing the previous, and each had the same cause: the parse list of the filter *in front of us* was extended while the neighbouring filters with their own parse went unchecked. Cured with **one** `parse_serialized_datetime` rather than a third correct copy (#1646).

The one place the callers genuinely differ gets a flag rather than trust. A bare time is formattable — `{{ v|time:"H:i" }}` is meaningful — but it has **no instant**, so `timesince` against its epoch anchor would confidently report the decades since 1970. `allow_time_only = false` makes that branch unreachable from the duration filters instead of relying on them not to call it.

## Reproducing first surfaced a second defect

The issue asked for a runtime reproduction before fixing, and it paid twice:

1. It showed the failure is the **silent echo**, not the 500 I had guessed when filing — which raises the severity, since a raw ISO timestamp in the page is easy to miss in review and hard to notice in production.
2. It exposed a bug the parse fix would otherwise have **shipped underneath itself**: a naive value was compared against `Utc::now()`, while Django compares it against `datetime.now()` — naive **local** time. A datetime two hours old reported **six** hours in a UTC-4 zone.

That second one produces a *plausible* wrong answer rather than an obvious one, so it would have survived review. It is visible only against Django's own output. `a_naive_datetime_is_compared_against_local_now_not_utc` uses an offset just below a bucket boundary, so any non-zero comparison error lands in a different unit.

## Deliberately not fixed here

The **output shape**, filed as **#2228**. Django joins with U+00A0, shows two adjacent units (`3 days, 5 hours`), and never reports seconds — all three diverge on *every* input, including the aware values that always parsed. A distinct defect from the parse; folding it in would have doubled this diff and mixed a rewrite into a fix.

## Verification

- **12 cases** — 5 pinning the shared parser's shape coverage, 7 driving the filters end to end.
- **5/5 gate-off red**, including a mutation that flips the comparison baseline and one that lets the duration filters accept a bare time.
- Determinism: every "now"-relative case sits well inside its unit bucket rather than near a boundary, so none of them is a coin flip (#1795).
- clippy 0; **10,327 passed / 0 failed / 1 xfailed**.


johnrtipton:
## Stage 14 — Retro, PR #2229 (#2227)

**Quality: 5/5.** The rare one where the process rules earned their keep in sequence, and I have a concrete thing to change rather than just an observation.

### The chain is the finding, not any one link

#2203 → #2216 → #2227, three releases, same class: a filter's parse list extended for the value shape in front of us while the neighbouring filters with their *own* parse went unchecked. Each link was found by fixing the previous one.

The repo's canon already names this shape (chain-shaped issues, #2142) and says to expect link N+1 rather than treat it as a slipped estimate. Applying that here changed what I built: instead of adding `%Y-%m-%dT%H:%M:%S` to two more functions — which would have worked, and would have made #2228's fix a fourth site — I extracted one parser. The next value shape is now fixed once.

**What I should have done earlier**: after #2216, grep for every other `parse_from_rfc3339` in the crate. It is one command, there were exactly two hits, and it would have found this link without waiting for a retro to notice it.

### Reproducing first paid twice, both times against my own filing

The issue is one I wrote yesterday, and running it corrected me twice:

1. I had guessed the failure "may be a 500 rather than the silent echo its siblings had, which would change the severity". It is the **silent echo** — a raw ISO timestamp rendered into the page. My guess would have under-prioritised it.
2. It exposed a defect the parse fix would have **shipped underneath itself**. A naive value was compared against `Utc::now()`; Django compares against `datetime.now()`, naive *local*. Two hours old read as **six** in a UTC-4 zone.

The second is the one worth dwelling on: fixing the parse makes that path *reachable*, so the fix would have converted a loud bug (raw timestamp) into a quiet one (plausible wrong number). **Making a broken path reachable is its own risk, and the reproduction is what surfaces it** — the parse fix alone would have passed every test I would have thought to write.

### What went well

- **The `allow_time_only` flag is a constraint, not a comment.** My own issue warned that a bare time has no instant, so `timesince` against its epoch anchor would report decades. I could have written "don't call it that way" in a doc-comment; making the branch unreachable from the duration filters means the warning cannot be ignored. Gate-off confirms: flipping the flag reddens.
- **The near-boundary test.** `a_naive_datetime_is_compared_against_local_now_not_utc` uses 23 hours, not 30 — at 30 an off-by-4-hours error still reads `1 day` and the test passes while broken. Choosing the offset so *any* non-zero error crosses a bucket is what makes it a detector rather than a decoration.
- **Splitting #2228 out.** The output shape (NBSP, two units, no seconds) diverges on *every* input including the ones that always parsed. Folding it in would have mixed a rewrite into a fix and doubled the diff.

### What didn't

**One gate-off mutation did not compile** — I wrote `if X && let Y = ...`, which is not stable Rust. The harness correctly reported `INVALID (mutation broke the build)` instead of a number, which is exactly the #2129/#2135 guard working, but it cost a re-run. When a mutation needs more than a token swap, write it as valid code first and mutate second.

**I have now shipped three PRs whose retro says "I enumerated the paths I knew about and was one short."** #2209 (a second render path), #2223 (a third), and this chain. The pattern is stable enough to act on rather than keep noticing: **grep for the sink, not for the callers I expect.**

### Verified

- 12 cases (5 parser-shape, 7 end-to-end); **5/5 gate-off red**.
- Every "now"-relative case sits well inside its unit bucket, so none is a coin flip (#1795).
- clippy 0; **10,327 passed / 0 failed / 1 xfailed**; CI green.

`RETRO_COMPLETE`

PR #2230, opened 2026-08-25
fix(templates): port Django's timesince output shape (#2228)

### LINKED ISSUE #2228: bug: timesince/timeuntil output shape diverges from Django on every input — one unit, plain space, and seconds

johnrtipton:
Closes #2228. Fourth and last link of the #2203 → #2216 → #2227 → #2228 chain.

#2227 fixed the **parse**. The **output** was wrong even for the aware values that always parsed:

| elapsed | Django | before |
|---|---|---|
| 30 s | `0 minutes` | `30 seconds` |
| 2 h 30 m | `2 hours, 30 minutes` | `2 hours` |
| 10 d | `1 week, 3 days` | `1 week` |
| 400 d | `1 year, 1 month` | `1 year` |

Three defects at once:

- **Separator.** Django's `avoid_wrapping` uses **U+00A0**, so a count and its unit never break across a line. djust used an ordinary space — a bad line break in production, and very hard to trace back.
- **Depth.** Django shows up to **two adjacent** units; djust discarded the more precise half of nearly every answer.
- **Seconds.** Django's smallest unit is the minute and it ignores seconds entirely.

## Calendar-aware, not approximated

Ported from `django/utils/timesince.py`. The part an approximation cannot reach is the year/month arithmetic — from Django's own docstring, there is exactly `1 year, 1 month` between **2013-02-10 → 2014-03-10** *and* between **2007-08-10 → 2008-09-10**, though the deltas are 393 and 397 days. Dividing by a fixed 2629746 seconds gets both wrong. Both are in the suite.

Django's `MONTHS_DAYS` quirk is reproduced **deliberately**: February is 28 with no leap-year case, so a pivot clamps to the 28th even in a leap year. Parity is the point — correcting it here would make djust disagree with Django on exactly those dates.

The **"adjacent"** rule is load-bearing and easy to miss: the walk stops at the first **zero**, so a value exactly one year and five days old is `1 year`, never `1 year, 5 days` — which Django's docstring calls out as impossible output. An implementation that collected the first two *non-zero* partials passes most tests and fails that one.

## Two things I got wrong first, both settled by asking Django

1. **I added a swap inside the shared body** so both filters could reuse it. That made `timesince` on a **future** value report the elapsed time backwards (`1 year`) where Django returns `0 minutes`. `timeuntil` swaps at its own call site instead, which is what `reversed=True` does.
2. **My February expectation was wrong, not the code.** I asserted `1 month, 2 days`; Django gives `1 month`. The remainder *is* two days — but `weeks` is zero and the adjacent walk breaks there. The test was wrong and the implementation was right.

Both are recorded in the test comments, because each is a trap the next person will hit.

## Verification

- **1,600-case randomized differential** against `django.utils.timesince`, across aware and naive values and durations from seconds to a decade: **0 diffs**.
- **14 cases** pinned against the pure two-argument function rather than through the filters, so every expectation is an **exact string** rather than a bucket that could flake near a boundary (#1795) — and so the calendar cases, which no "now"-relative test could reach, can be pinned at all.
- **8/8 gate-off**, plus two extra mutations after the first February one turned out to be *semantically a no-op* for the tested inputs (`d.day().min(28)` happens to equal the clamp there). Worth noting as a harness lesson: asserting the mutation text changed is not the same as asserting the behaviour did.
- The two filters shared no code before, so every change had to be made twice (#1646). One function now.
- clippy 0; **10,328 passed / 0 failed / 1 xfailed**.

## Not included

Translated unit names. `TIME_STRINGS` entries are `ngettext` lookups, so full i18n parity needs the catalog work tracked in the same ROADMAP bucket. This ships English-only parity, which is what every current djust project gets today.


johnrtipton:
## Stage 14 — Retro, PR #2230 (#2228)

**Quality: 5/5.** Two of my own mistakes, both caught by asking Django rather than by reasoning, and a genuine harness lesson.

### Asking Django settled two disagreements — in opposite directions

When my implementation and my test disagreed, I did not decide which was right. I ran Django.

1. **Future value.** I had added a swap inside the shared body so both filters could reuse it. Django returns `0 minutes` for a future value; the swap reported the elapsed time backwards (`1 year`). **The test was right, the code was wrong.**
2. **February pivot.** I expected `1 month, 2 days`; the code produced `1 month`. Django says `1 month` — the remainder *is* two days, but `weeks` is zero and the adjacent walk breaks there. **The code was right, the test was wrong.**

Two disagreements, opposite resolutions, and I would have got at least one backwards by argument. The habit worth keeping is not "trust the code" or "trust the test" — it is *neither, when a reference implementation is a subprocess away*.

### The harness lesson

One gate-off mutation came back **GREEN**, and it was not a coverage gap: my mutation `MONTHS_DAYS[..].min(d.day())` → `d.day().min(28)` is **semantically identical** for the inputs under test. The harness asserted the mutation text was found and that the source changed — both true — but a valid mutation that happens not to change behaviour reports the same green as a missing test.

**Asserting the mutation text changed is not the same as asserting the behaviour changed.** I replaced it with two that genuinely differ (all-31 months, and dropping the clamp), and both go red. This is a level above the #2129/#2135 harness rules, which cover mutations that fail to apply or break the build; this one applied cleanly and did nothing.

### What went well

- **1,600 randomized comparisons against `django.utils.timesince`**, aware and naive, seconds to a decade: 0 diffs. That is worth more confidence than any number of hand-picked cases, and it caught nothing — which is itself the useful signal, because the hand-picked cases had already found everything.
- **Testing the pure function, not the filter.** The filters compare against "now", so a test through them can only assert coarse buckets and would be a coin flip near boundaries (#1795). Exposing the two-argument function let every case pin an exact string — and the calendar cases (Django's own 393-vs-397-day examples) are unreachable from a "now"-relative test at all.
- **Reproducing the quirk rather than fixing it.** Django's February-28 clamp has no leap-year case. It is a bug in Django, and copying it is correct here: parity is the point, and "improving" it would make djust disagree on exactly those dates. Written down so nobody helpfully repairs it later.

### What didn't

**I updated #2227's tests to the new shape after the fact.** Six of them asserted `"1 day"` with an ordinary space and went red when the output changed. That is the correct outcome — they were pinning real behaviour — but I shipped #2227 four hours ago knowing #2228 was filed and would change exactly those strings. Writing them against the *known-correct* Django output at the time would have cost nothing and saved the churn.

**A dead helper survived one commit.** `duration_since` was superseded by `now_and_then` and clippy caught it. Minor, and the linter did its job, but I ran clippy after the full Python suite rather than before — a two-minute check placed after a three-minute one.

### Verified

- 1,600-case randomized differential: **0 diffs**.
- 14 exact-string cases; **8/8 gate-off** plus the two replacements.
- clippy 0; **10,328 passed / 0 failed / 1 xfailed**; CI green.

`RETRO_COMPLETE`

PR #2231, opened 2026-08-25
fix(templates): implement every Django date format code (#2217)

### LINKED ISSUE #2217: tech-debt: nine Django date format codes are unimplemented and render as their own letter

johnrtipton:
Closes #2217.

Nine codes — `b c f L o r S t u w W z` — were unimplemented and fell through the formatter's catch-all, rendering as **their own letter**:

```
{{ v|date:"jS F Y" }}   →   22S August 2026
```

Quiet for a structural reason: rendering an unknown character as itself is *also* the correct behaviour for a literal, and Django does the same — so an unimplemented code is indistinguishable from an intentional one by inspection. Only a differential separates them.

All **38** codes Django recognises now match, pinned as **one table** rather than as the nine that were missing: a table with a hole in it looks exactly like a table without one (v1.0.0rc4 retro finding #1).

## A tenth code was already implemented and already wrong

**`N` is Associated Press style** (`django.utils.dates.MONTHS_AP`), not `%b` plus a period. AP does not abbreviate the short months at all, and September is `Sept.`:

| | Django | before |
|---|---|---|
| March | `March` | `Mar.` |
| April | `April` | `Apr.` |
| May | `May` | `May.` |
| June | `June` | `Jun.` |
| July | `July` | `Jul.` |
| September | `Sept.` | `Sep.` |

**Half the year rendered incorrectly**, and an existing unit test *pinned the wrong answer* (`Mar. 15, 2026`). It now asserts Django's `March 15, 2026`.

## The same sampling lesson, three times in one PR

This is the part worth reading.

1. **`N` survived** the 38-code table because its three sample values are January, August and February — all months where `%b` + `.` happens to be right. A randomized sweep found it in seconds.
2. **A `W` mutation survived gate-off.** Replacing ISO week with `ordinal / 7 + 1` passed the whole table, because that formula coincides with the ISO week for all three values.
3. **An `o` mutation survived** for the same reason: ISO week-year only diverges from the calendar year at year boundaries, and none of the three values sits near one.

None of these was a missing *code*. The table enumerated every code and then sampled **three values per code** — the identical blind spot, one level down. Discriminating cases added (2027-01-01 is a Friday, so ISO puts it in week 53 of 2026), and all twelve months pinned for `b`/`M`/`F`/`N`.

## Verification

- **3,000-case randomized differential** against Django across fifteen years, every code, four microsecond values: **0 diffs**.
- **8 cases**; **12/12 gate-off**, after two mutations that initially survived were shown to be coverage gaps rather than harness errors.
- `no_code_renders_as_its_own_letter` catches any *future* code that regresses to the catch-all, without needing to know which.
- clippy 0; **10,327 passed / 0 failed / 1 xfailed**; full Rust workspace green.

## Not included

Localized month and day names — `{{ d|date:"D" }}` gives `Sat` where German Django gives `Sa`. That needs translation catalogs rather than format logic and is piece 2 of #2221.


johnrtipton:
## Stage 14 — Retro, PR #2231 (#2217)

**Quality: 4/5.** The nine missing codes were mechanical. Everything worth recording came from the same lesson landing three times, and I only caught two of the three by running things I did not have to run.

### One lesson, three hits, in the PR built to prevent it

I wrote a 38-code table *specifically* invoking the enumerate-every-variant rule (v1.0.0rc4 finding #1) — a table with a hole in it looks exactly like a table without one. Then:

1. **`N` was wrong and the table passed.** Django's `N` is AP style: March through July are spelled out, September is `Sept.`. Half the year. The table's three values are January, August and February — all months where `%b` + `.` is right by accident.
2. **A `W` mutation survived gate-off.** `ordinal / 7 + 1` coincides with the ISO week for all three values.
3. **An `o` mutation survived** — ISO week-year only diverges at year boundaries, and none of the three values is near one.

None of these is a missing code. I enumerated every code and then **sampled three values per code**. The identical blind spot, one level down, in the artifact whose stated purpose was closing it.

The durable form: **"enumerate every variant" has to be asked about every axis, not the one you happened to notice.** Codes × values is two axes; I covered one exhaustively and felt finished.

### What actually caught them

- `N`: a **3,000-case randomized sweep** I ran after the table was already green, out of habit rather than suspicion. It found the bug in seconds.
- `W` and `o`: **gate-off**, but only because two mutations came back GREEN and I treated that as a signal rather than as a pass. It would have been easy to read "the mutation didn't fail anything, so it must be equivalent" and move on — which was true for the `d.day().min(28)` case in #2230, and *false* here.

So: **a surviving mutation is a question, not an answer.** Sometimes it means the mutation was a no-op; sometimes it means the tests cannot see the difference. Those are opposite conclusions and they look identical from the harness.

### What went well

- **`no_code_renders_as_its_own_letter`** is the test I am happiest with: it catches any *future* code regressing to the catch-all without needing to know which one, and it encodes the bug's actual shape rather than its instances.
- **Correcting a test that pinned the wrong answer.** `test_date_filter_datefield_bare_date` asserted `Mar. 15, 2026` — a wrong expectation, protected as if it were a contract. I checked Django before changing it rather than assuming my new code was right.
- **The pre-push hook earned its keep.** I had run the *new* test file after the `N` fix but not the whole crate, so a stale in-file unit test went unnoticed until the hook ran the full suite.

### What didn't

**I ran the targeted test file after the `N` fix instead of the crate suite.** Two minutes, and it would have caught the stale assertion before the hook did. I had run the full crate suite *before* that change and treated it as still valid.

**I made a mess of the commit shape.** An `--amend` intended for the fix commit landed on the docs commit, folding a code change into `docs(changelog):`. Recoverable pre-PR, and I verified the hash changed each time (#1524) — but the root cause is that I amended without checking which commit was HEAD.

### Verified

- 3,000-case randomized differential across fifteen years, every code, four microsecond values: **0 diffs**.
- 8 cases; **12/12 gate-off** after the two survivors were shown to be real gaps.
- clippy 0; **10,327 passed / 0 failed / 1 xfailed**; full Rust workspace green; CI green.

`RETRO_COMPLETE`

PR #2083, opened 2026-07-15
feat(bug-capture): iter B — read-only replay viewer + debug-panel Share button (#1562)

johnrtipton:
## Summary

Closes #1562 (B7 iter B — bug-capture browser replay). Iter A (#1552-followup) shipped the `djbug1.` data shape + encoder/decoder + scrub hook. This PR adds the browser-based replay surface that closes the share→reproduce loop:

- `GET /__djust__/replay/<blob>` — decodes a `djbug1.` blob and renders `state_before`/`state_after` side-by-side with a per-key diff table, every `vdom_patches` entry as escaped `<pre><code>`, and a sandboxed `<iframe srcdoc>` preview of any captured HTML.
- New `djust.urls` module — DEBUG-gated route registration, double-gated (urlconf omission + in-view re-check).
- Debug panel "📋 Share bug" button in the Time Travel tab — computes a blob server-side via `render_with_diff()` + `encode_view_state()` and copies it to the clipboard client-side.

**Design deviation from the issue text**: implemented as a plain Django view (`bug_capture_views.py`), not a `LiveView`. A `LiveView` mounts over a WebSocket and accepts `event` frames by design; giving this route that machinery would mean actively disarming a dispatch surface a plain HTTP view never has in the first place. Mirrors the existing `djust.theming.gallery` precedent for built-in dev-tooling views.

## Security

- Strictly read-only: `event_name` is never resolved to a callable (structural regression test greps the source for the dangerous `getattr(..., event_name)` shape), zero database queries, captured `tenant_id` never scopes a query or sets tenant context (verified via a spy on `set_current_tenant`).
- XSS: captured HTML is escaped via Django's autoescaping inside `<pre><code>`; a malformed-blob 400 response is `text/plain` (never `text/html`) so a crafted blob can't get its own error message parsed as markup.
- DEBUG-gated at two layers, both keyed on the same `DJUST_BUG_CAPTURE_PROD_OPT_IN` literal-`True` opt-in `bug_capture._enforce_prod_gate()` already uses.

## Test plan

- [x] 52 new regression cases across 4 new files (Python: `test_bug_capture_views.py` 22, `test_bug_capture_urls.py` 6, `tests/integration/test_bug_capture_share.py` 8; JS: `tests/js/bug_capture_share.test.js` 16).
[...]

PR #2232, opened 2026-08-25
fix(retro): tick Open Items whose issues have all closed (#2200)

### LINKED ISSUE #2200: RETRO.md: 121 unchecked Open Items reference already-closed issues — including the row about automating this
Found during the #2017 iteration-3 review, while checking whether two RETRO Open Items were stale. They were — and so are 119 others.

## The measurement

```
unchecked "- [ ]" items in RETRO.md whose referenced GitHub issues are ALL closed: 121
```

Derived by extracting every `- [ ]` line carrying a `#NNNN` reference and joining against `gh issue list --state all --limit 600`.

RETRO.md's convention is clearly that items DO get closed — there are 89 `- [x]` entries, several with resolution notes in exactly the right shape:

```
- [x] bug-capture iter B (#1562) — **resolved in v1.1.0-11 (PR #2083)**; iter C (#1561) still deferred
```

So these are not frozen historical snapshots. They are a tracker that has drifted.

## The part worth putting on a slide

`RETRO.md:533` reads:

```
- [ ] Close tracker rows automatically when their GitHub issue closes — Action Tracker #330 (GitHub #2143)
```

Unchecked. Its issue is closed. The row about automating row-closing is itself a stale row.

## Why it matters

This is the #2143 problem recurring at a different altitude. That issue found "65 of 70 `Open` rows were already resolved — the health signal retros quote was wrong by an order of magnitude", and the same is now true of RETRO.md's per-milestone Open Items. Anything that reads these to judge outstanding work — a human scanning for what is left, or `/pipeline-retro` synthesising a milestone — gets a number that is wrong by roughly two orders.

## Not fixed here

PR #2197 corrected exactly two of these (`#2136` and `#2017`, both about the feature that PR ships) and left the rest, per the scope-discipline rule (#1079): fix what the issue cites, file the systemic remainder.

## Suggested fix

Single-script transformation (#1312), mirroring `scripts/strike-shipped-roadmap-rows.py`:

1. Parse every `- [ ]` line with a `#NNNN` reference.
2. Resolve each issue's state via one batched `gh issue list --state all`.
3. Where ALL referenced issues are closed, flip to `- [x]` and append the closing PR — sourced from `closedByPullRequestsReferences`, **never guessed** (#1197).
4. Leave any line with a still-open reference alone.

Two cautions from doing this on ROADMAP.md:

- Split table/markdown cells on **unescaped** pipes only. A naive `split('|')` ate half a description in `strike-shipped-roadmap-rows.py`'s first version because a cell contained `x\|safe`.
- Assert per-line that only the intended span changed, so a boundary mis-parse fails loudly instead of silently rewriting prose.

johnrtipton:
Closes #2200.

**168 of RETRO.md's 197 unchecked Open Items** referenced issues that had all closed. Anything reading those boxes to judge outstanding work — a person scanning for what is left, or `/pipeline-retro` synthesising a milestone — got a number wrong by roughly an order of magnitude.

Including, at `RETRO.md:533`:

```
- [ ] Close tracker rows automatically when their GitHub issue closes
```

The row about closing rows was itself a stale row.

## Extended, not duplicated

`scripts/check-action-tracker.py` already owned the batched issue fetch, the `--fix` flag, the `make check-tracker` target and the test file. RETRO.md's two structures — the tracker table's `Status` column and the per-milestone `- [ ]` checklists — drift the same way for the same reason. Two tools would have meant two fetches and two ideas of what "closed" means (#1646).

The issue proposed a new script mirroring `strike-shipped-roadmap-rows.py`. Extending the existing one is better here *and* it closes the self-referential row honestly: #2143's tool now does what #2143's own tracker row asked for.

## A collision nearly wrote a wrong citation into the document

RETRO.md carries **two numbering schemes on one line**:

```
- [ ] … — Action Tracker #329 (GitHub #2142)
```

and tracker numbers **collide with real issue numbers**. 38 of the 94 unchecked items naming a GitHub issue also carry a tracker number resolving to a different, real issue.

A first pass took the first `#NNN` and reported **#2142 as closed by PR #362** — which is issue **#329**'s closer. Plausible, verifiable-looking, and wrong: exactly the failure mode #1197 catalogues, one commit from being written into a document.

A `GitHub #NNNN` reference is now authoritative for the **decision** as well as the citation, since ANDing over a colliding number can mask an open issue as easily as invent a closed one.

**All 117 PR citations written were verified against GitHub's `closedByPullRequestsReferences`** before this was committed — 97 distinct issues, each checked individually.

## Ambiguity removed rather than pinned

Where an item cites several closed issues, **all** distinct closers are listed. A gate-off mutation swapping first for last **survived** the suite — precisely because that choice is arbitrary. So the choice went away instead of acquiring a test for my guess. (Matches the table path, which already writes `Resolved by PR #a, PR #b`.)

## Left alone, deliberately

| | count | why |
|---|---|---|
| no issue reference | 28 | nothing to check against; guessing is worse than drift |
| cites a still-open issue | 5 | partly-done work must never read as finished |
| human-annotated | — | `RETRO.md:682` is correct and precise, and would be a permanent false alarm |

That last one matters more than its count. `bug-capture iter B (#1562) — **resolved in v1.1.0-11 (PR #2083)**; iter C (#1561) still deferred` is a good row that a naive rule calls drift. **One permanent false alarm is enough to make a check worth ignoring.**

## Verification

- **15 new cases** (34 in the file); **12/12 gate-off**, including one that forces a boundary mis-parse to prove the post-condition guard still fires rather than leaving it decorative (#1859).
- **Idempotent**: a second run reports `No drift: every row and item with an issue agrees with GitHub.`
- Line count unchanged; **168 lines changed, 0 in unintended ways** — verified by re-normalising each changed line back to its original and comparing.
- Existing 13 table cases and the client-size manifest suite still pass.

## Noticed, not fixed

`RETRO.md:3843` cites **`#2287`**, which is neither an issue nor a PR in this repo — the numbering does not reach that far. It sits in a stats table cell rather than an Open Item, so nothing here touches it, but it is a bogus citation of the kind #1071 warns about and someone should decide what it was meant to say.


johnrtipton:
## Stage 14 — Retro, PR #2232 (#2200)

**Quality: 5/5.** The mechanical part was an hour; the value was in two decisions that went against the issue's own suggestion, and one bug I caught in my own script before it wrote a false citation.

### The bug worth the whole PR

RETRO.md carries two numbering schemes on one line — `Action Tracker #329 (GitHub #2142)` — and **tracker numbers collide with real issue numbers**. My first pass took the first `#NNN` on the line and was about to write:

> `#2142 — **closed (PR #362)**`

PR #362 is issue **#329**'s closer. Every part of that citation is well-formed, verifiable-looking, and wrong.

I caught it by spot-checking one PR number that looked implausibly low for a recent issue — not by reasoning about the parser. That is worth being honest about: the guard that worked was *noticing an implausible number*, and the reason I was looking at all is that this repo has #1197 as a named failure mode. Then I verified **all 117 citations** (97 distinct issues) rather than the sample of 39 I started with, because a sample is exactly what let the bug through in the first place.

The collision also affects the **decision**, not just the citation: ANDing "all refs closed" over a colliding tracker number can mask an open issue as easily as invent a closed one. Both now ignore everything but the `GitHub #NNNN` ref when one is present.

### Two departures from the issue's plan

**1. Extended `check-action-tracker.py` rather than writing the sibling the issue proposed.** That script already had the batched fetch, `--fix`, the Makefile target and the tests, and RETRO.md's two structures drift for identical reasons. Two tools would have meant two fetches and two definitions of "closed" — the #1646 shape. It also closes the self-referential row honestly: #2143's tool now does what #2143's own row asked for.

**2. Removed an ambiguity instead of testing my guess.** A gate-off mutation swapping "first ref" for "last ref" **survived** the suite. My instinct was to add a test pinning last-wins. But it survived *because the choice is arbitrary* — an item citing two closed issues has two equally defensible primaries. Listing every closer removes the question. **A surviving mutation sometimes means the behaviour it changes should not exist.**

### The judgement I am least sure of, stated plainly

I skip items a human has annotated `**resolved`/`**deferred`, in both directions. That is a real hole: a human could annotate something wrongly and the check would never notice. I took it because `RETRO.md:682` is a *good* row that a naive rule calls drift, and one permanent false alarm is enough to make a check worth ignoring. But it is a trade, not a free win.

### What went well

- **Verified the transformation touched nothing else** by re-normalising each changed line back to its original form and comparing — 168 changed, 0 unintended, line count identical. Cheaper and stronger than reading a 336-line diff.
- **Forced the post-condition guard to fire.** It has no trigger with well-formed input, which is how a safety assert rots into decoration (#1859). A monkeypatched off-by-one regex proves it still catches a boundary mis-parse.
- **The count grew from the filed 121 to 168** because more issues closed in between — including a dozen from this drain. Worth stating rather than quietly reporting a different number than the issue.

### Verified

- 15 new cases (34 in the file); **12/12 gate-off**; idempotent second run reports no drift.
- All 117 PR citations checked against GitHub individually.
- Existing table cases and the client-size manifest suite unaffected; CI green.

### Noticed, not fixed

`RETRO.md:3843` cites `#2287` — neither an issue nor a PR; the numbering does not reach that far. It is in a stats cell rather than an Open Item so nothing here touches it, but it is a bogus citation of the #1071 kind and someone should decide what it meant.

`RETRO_COMPLETE`

Issue #2215, opened 2026-08-25
fix(tests): the #1882 wire-version pollution flake is still live — reset_djust_globals does not fully cure it


johnrtipton:
## Symptom

`python/djust/tests/test_recovery_version_staleness_1817.py::test_time_travel_jump_recovery_version_is_current` failed in CI on PR #2213:

```
python/djust/tests/test_recovery_version_staleness_1817.py:209
    assert v_arm == v_mount + 1, f"arming event must be {v_mount + 1}, got {v_arm!r}"
E   AssertionError: arming event must be 2, got 3
E   assert 3 == (1 + 1)
```

Note **where** it failed: `v_mount == 1` passed, so the mount baseline was clean. The extra `_next_version()` bump happened *between* the mount frame and the first `bump` event — i.e. **during** the test, on a connection that had already mounted cleanly.

## Why this is not "just a flake"

This is exactly the class `python/djust/test_isolation.py:16` documents as **fixed**:

> **#1882** — process-global wire-version drift: a stray `djust_hotreload` frame in the shared `InMemoryChannelLayer` re-renders on the victim consumer and bumps its per-connection `_next_version()` counter, so a time-travel jump lands at version 4 instead of 3 (`test_time_travel_jump_recovery_version_is_current`).

`reset_djust_globals` was written as the *systemic* cure — "that retires the entire flaky-class instead of patching one test at a time." The same test, the same counter, the same off-by-one-bump signature is still failing, so the cure is incomplete.

The likely gap: the reset is **pre-test** and drops the cached channel-layer backend, which clears state *buffered* by an earlier test. It cannot stop a **still-live** producer — a lingering asyncio task, hot-reload watcher thread, or tick loop from an earlier test — that sends to `djust_hotreload` on the *new* layer while the victim test is mid-flight. That would produce exactly this signature: clean mount, then a stray bump before the next assertion.

## What surfaced it

PR #2213 changed **no production code** on this path — it added two cases to an unrelated structural test file. That alone re-shards xdist's test-to-worker distribution, which is enough to trigger it. **A latent pollution bug that a test-count change can flip is going to fire again, at random, on an unrelated PR.**

## Reproduction status

- Isolation: **3/3 green** (`pytest python/djust/tests/test_recovery_version_staleness_1817.py`).
- Local full CI-way run (`pytest tests/ python/tests/ python/djust/tests/ -n auto`): **3/3 green**, 10,288 passed.
[...]

PR #2233, opened 2026-08-25
fix(views): make SimpleLiveView render, and stop it shadowing LiveView (#2219)

### LINKED ISSUE #2219: bug: djust.simple_live_view.LiveView cannot render — get_context_data raises on every instance
Found while auditing render paths for #2209 (the timezone fix needed to reach *every* path that renders a template, so I enumerated them).

## The module does not work at all

`python/djust/simple_live_view.py:38` builds its context by walking `dir(self)` and `getattr`-ing each name:

```python
for key in dir(self):
    if not key.startswith("_") and not callable(getattr(self, key)):
```

`dir(self)` includes `as_view`, inherited from `django.views.View`. That is a `classonlymethod`, whose `__get__` raises on an instance:

```
File "django/utils/decorators.py", line 11, in __get__
    raise AttributeError("This method is available only on the class, not on instances.")
```

So `get_context_data()` raises `AttributeError` for **any** instance of **any** subclass, and `render_template()` catches it and returns `<div>An error occurred rendering this view.</div>` (or the DEBUG variant). Every render through this module fails before a template is ever reached.

Reproduced directly:

```python
from djust.simple_live_view import LiveView as SimpleLiveView
SimpleLiveView().get_context_data()   # AttributeError
```

## Why nobody noticed

- The class is named **`LiveView`** — the same name as `djust.LiveView`, which it is not. Grepping for `SimpleLiveView` finds nothing, so the module reads as unused when it is actually just unfindable by its obvious name.
- It is not exported from `djust/__init__.py`.
- It has **no tests**.
- The `except Exception` in `render_template` converts the crash into a generic message, so a user hitting it sees "An error occurred rendering this view" with no clue.

The only reference outside the file is `pyproject.toml:330`, which lists it as an ADR-023 strict-typing island — so it was type-checked but never executed.

## Decision needed, not just a patch

The `getattr` guard is a one-liner (`try/except AttributeError`, or skip names on `View`). But the prior question is whether the module should exist: it duplicates the name of the primary public class, predates most of the framework (added in #77), has no callers, no tests, and no documentation. Either

- **remove it** (and drop the `pyproject.toml:330` mypy island), or
- **fix, test, document, and export it** under a name that does not collide.

Silently keeping a broken shadow of the framework's central class is the worst of the three.

## Related

#2218 (the #2209 fix) wires this path into the shared `djust.timezone_bridge.apply_active_timezone` anyway, so it is correct if the module is ever revived — a second private copy there would have been the #1646 twin. `test_the_simple_live_view_path_is_wired_but_currently_unreachable` asserts the `AttributeError` rather than describing it, so **that test goes red the moment this issue is fixed**, which is the signal to add the real behavioural case.

johnrtipton:
Closes #2219.

`get_context_data` walked `dir(self)` and `getattr`-ed every name, which reaches Django's `View.as_view` — a `classonlymethod` whose `__get__` raises on an **instance**:

```
AttributeError: This method is available only on the class, not on instances.
```

Every render failed before a template was reached, for **any** subclass, **always**. And `render_template`'s `except Exception` turned the crash into a generic `An error occurred rendering this view.`, so nobody saw why.

## The rename

The class was called **`LiveView`** — the same name as `djust.LiveView`, which it is not. That is why grepping for `SimpleLiveView` finds nothing and the module reads as unused when it is merely *unfindable*: it went two PRs (#2209, #2223) without anyone noticing it was a live render path. Now `SimpleLiveView`, with `LiveView` kept as a module-level alias.

The issue offered three options — remove it, fix it, or keep it broken. This takes the second, which the issue itself framed as fix + test + a non-colliding name. It is not promoted to the top-level `djust` namespace: that is a public-API decision, and it remains importable from `djust.simple_live_view` as before.

## The guard is the fix, not the exclusion list

A name in `dir(self)` is **never** a promise that `getattr` will succeed — a `classonlymethod`, a property whose getter raises, a descriptor refusing instance access all look like ordinary attributes until read. This is a method whose entire job is reading attributes it does not know about, so each read now goes through `try`/`except AttributeError`.

**`as_view` is deliberately not in the plumbing-exclusion set.** Listing it in both places made the two mechanisms shadow each other: a gate-off re-introducing the *original bug* left the whole suite **green**, because the guard silently covered for it (#2129). The set now holds only names that read fine and are excluded for a different reason — plumbing rather than danger — which keeps both independently reachable:

| mutation | result |
|---|---|
| remove the `AttributeError` guard | **12 failures** |
| remove a name from `_VIEW_INTERNALS` | **2 failures** |

A view with no `template` also now says so, instead of claiming the Rust backend is unavailable — false, and unactionable for whoever reads it.

## Also fixes a cross-test leak this release introduced

`test_number_localization_2221.py`'s reset fixture called `translation.deactivate_all()`. That installs a `NullTranslations` and leaves `get_language()` returning `None`, so `get_format` skips the locale modules and falls back to `global_settings` — where `NUMBER_GROUPING` is **0**:

```
fresh                  get_language()='en-us'  NUMBER_GROUPING=3
after deactivate()     get_language()='en-us'  NUMBER_GROUPING=3
after deactivate_all() get_language()=None     NUMBER_GROUPING=0
```

**Grouping was silently off for every later test in the same worker.** It poisoned a case in this PR, which rendered `1234567` where it expected `1,234,567` — caught only because that test happened to land in the same worker under `-p no:randomly`.

That is the ordering-dependent flake class this repo keeps paying for (#2215, #2187), and I introduced it two PRs ago in the very change that added those cases. `deactivate()` restores `settings.LANGUAGE_CODE`, which is what "reset the language" should mean.

## Verification

- **14 cases**; **7/7 gate-off**.
- Every ordering pair re-checked: the file now passes after each of the three related suites, not just alone.
- mypy green on 891 files; **10,356 passed / 0 failed / 1 xfailed** across all three roots.
- The #2216-era tripwire test — which *asserted* the module was broken, so it would fail the moment someone fixed it — is now the real behavioural case.


johnrtipton:
## Stage 14 — Retro, PR #2233 (#2219)

**Quality: 4/5.** The fix was small. Two findings came out of it, and one of them is a bug I shipped two PRs ago.

### I introduced the leak I have spent this drain complaining about

`test_number_localization_2221.py`'s reset fixture called `translation.deactivate_all()`. That leaves `get_language()` as `None`, so `get_format` falls back to `global_settings` where `NUMBER_GROUPING` is **0** — grouping silently off for every later test in the same worker.

I wrote that fixture in PR #2222, in the change that added the number-localization tests. It poisoned a case in *this* PR. And I found it only because I bisected instead of reaching for the obvious fix: my test failed in the full run and passed alone, and the tempting move was to add my own reset fixture and move on. That would have worked, hidden a real leak, and left the next test to trip over it.

The measurement that settled it took one minute:

```
fresh                  get_language()='en-us'  NUMBER_GROUPING=3
after deactivate()     get_language()='en-us'  NUMBER_GROUPING=3
after deactivate_all() get_language()=None     NUMBER_GROUPING=0
```

**"Reset the language" has two spellings and only one of them means what it says.** `deactivate_all()` reads like the thorough choice; it is the one that leaks.

### The gate-off that mattered

My first version excluded `as_view` **by name** *and* guarded reads with `try/except AttributeError`. Belt and braces. Then a gate-off re-introducing the **original #2219 bug** — dropping `as_view` from the exclusion set — left the suite **green**, because the guard silently covered for it.

Two mechanisms fixing the same half is not redundancy, it is one fix and one decoration, and there is no test that can tell them apart while both are present (#2129). Removing `as_view` from the set made both reachable: the guard now accounts for 12 failures when removed, the set for 2.

The general lesson I keep re-learning: **when a mutation of a real mechanism goes green, the question is which of the two mechanisms is load-bearing — not whether to add a test.**

### What went well

- **The tripwire worked exactly as designed.** #2216 left a test that *asserted* the module was broken, specifically so it would fail the moment someone fixed it. It did, and it told me where the real behavioural case belonged. Writing a failing-on-fix assertion instead of a comment cost nothing and paid a week later.
- **Caught my own tautology before commit.** A test read `assert "…" in out or True` — unconditionally true, in a test file about a module that had been silently broken for years. Fixing it exposed a second small honesty bug: a view with no `template` was told the *Rust backend* was unavailable.
- **Declined to over-reach.** The issue floated removing the module. Fixing and renaming is reversible and non-destructive; promoting it to the top-level `djust` namespace is a public-API decision I left to the maintainer, and said so.

### What didn't

**The worktree cost me two false starts** — an `ImportPathMismatchError` from the editable install pointing at the main checkout, then a missing gitignored `.so`. Both are documented gotchas I have hit before. Setting `PYTHONPATH` and copying the `.so` should be the first thing I do in a worktree, not the thing I do after two failures.

### Verified

- 14 cases; **7/7 gate-off** (12 failures when the guard goes, 2 when the exclusion set does).
- Every ordering pair re-checked — the file passes after each related suite, not just alone.
- mypy green on 891 files; **10,356 passed / 0 failed / 1 xfailed**; CI green.

`RETRO_COMPLETE`

PR #2235, opened 2026-08-25
security: stop SimpleLiveView render errors exposing exception detail (#2596)

johnrtipton:
Fixes CodeQL alert [#2596](https://github.com/djust-org/djust/security/code-scanning/2596) — `py/stack-trace-exposure`, medium severity, `python/djust/simple_live_view.py`.

`render_template`'s `except` rendered this whenever `DEBUG` was on:

```python
return f"<div>Template error: {e}</div>"
```

## Two defects, and CodeQL named only the second

**1. CWE-79 — the message was interpolated unescaped.** Template errors routinely echo the offending value, so an exception carrying a `<` injected markup straight into the page.

**2. CWE-209 — the detail reached the response at all.** This is the one the rule fired on.

The distinction matters because **the obvious fix only addresses (1)**. `websocket.py:549` already fixed exactly this shape:

```python
return f"<!-- Error rendering embedded child: {escape(str(e))} -->"
```

…and its comment says it *"mirrors the DEBUG gate in `simple_live_view.render_template`"*. **The copy was fixed and the original it was modelled on was not** — parallel-path drift (#1646), found by grepping for the *shape* rather than reading the cited line.

Copying that fix verbatim here would have closed the real bug and **left the alert open**, since the detail still flows. That is pinned rather than argued: one gate-off mutation applies the sibling's fix exactly, and the suite still reddens.

## The fix

`logger.exception` plus a static string in **both** modes. No `DEBUG` branch at all, so:

- there is no mode-dependent leak to reason about;
- nothing for a misconfigured production `DEBUG=True` to expose;
- the developer this branch was written for gets **more** — a full traceback in the log, rather than a one-line `str(e)` in a div.

## Checked and deliberately not changed

`live_tags.py:370,377` interpolate form errors into markup with the same shape. **Verified, not assumed**: `live_errors` is a plain `@register.simple_tag` with no `mark_safe`, so Django escapes the entire return. Rendering a `<script>alert(1)</script>` payload through it produces fully-escaped output — no XSS.

It is a *functional* bug (the tag emits literal escaped `<div>` text instead of markup) and wants its own issue, not a security fix smuggled into this one.

## Verification

- **5 new cases** (24 in the file); **4/4 gate-off**, including the "apply the sibling's escape fix" mutation.
- The payload asserts against a script tag, an attribute-breaking quote, an ampersand *and* an event handler, so a partial escape fails too.
- A separate case pins that the detail still reaches the **log**, so removing it from the page does not lose it.
- mypy green on 891 files; **10,362 passed / 0 failed / 1 xfailed**.

One process note: the CHANGELOG entry first landed in the released `[1.1.0]` section rather than `[Unreleased]`, because the naive "find the first `### Security`" lookup hits the shipped block — the #2028 failure mode. Caught and corrected before commit; the released section is byte-identical.


johnrtipton:
## Stage 14 — Retro, PR #2235 (CodeQL #2596)

**Quality: 4/5.** A one-line alert that turned out to have two defects, only one of which the scanner named — and the fix the codebase already contained would have closed the wrong one.

### The finding

CodeQL pointed at `f"<div>Template error: {e}</div>"` and said *information exposure*. Reading the line surfaced a **second** defect it did not name: the interpolation is unescaped, so an exception carrying a `<` injects markup into the page.

Then grepping for the *shape* rather than the line found `websocket.py:549` — the same pattern, already fixed with `escape(str(e))`, whose comment says it **"mirrors the DEBUG gate in `simple_live_view`"**. The copy had been fixed and the original it was modelled on had not. That is #1646 with the arrow pointing the unusual way: normally the original is fixed and the copy drifts.

**The trap**: `escape()` is right there, it is the house pattern, and it fixes the bug I found by reading. It does **not** fix the bug CodeQL reported. Applying it would have felt like following canon and would have left the alert open. I pinned that as a gate-off mutation — apply the sibling's fix verbatim, suite still reddens — so the distinction is enforced rather than explained.

### Why removing the DEBUG branch rather than escaping it

The DEBUG-gated detail is defensible in principle (Django's own technical 500 page does it), and the repo had established it as a pattern. I dropped it anyway because:

- one code path means no mode-dependent behaviour to reason about, and nothing for a misconfigured production `DEBUG=True` to expose;
- `logger.exception` gives the developer a **full traceback**, which is strictly more than the one-line `str(e)` the div carried.

That makes it an improvement rather than a trade, which is the bar for removing a developer-facing affordance.

### What went well

- **Verified the neighbour instead of widening scope on a guess.** `live_tags.py:370,377` have the identical shape. Rather than "fix them too" or "leave them", I rendered a `<script>` payload through `live_errors` and confirmed Django escapes the whole return — it is a *functional* bug, not a security one. Two minutes, and it kept a security PR from absorbing an unrelated rendering fix.
- **Caught my own contradiction.** My first version of the code comment said "only the first is the one CodeQL names" and then described the second as the rule's subject. Wrong and self-contradicting, in a comment whose whole job is explaining a subtle distinction.

### What didn't

**The CHANGELOG entry landed in the released `[1.1.0]` section.** My insert found the first `### Security` heading in the file, which belongs to the shipped block — the #2028 failure mode, in a repo that has a pre-commit hook specifically because this happened before. I caught it by checking which `## [` heading the line fell under, corrected it to `[Unreleased]`, and confirmed the released section is byte-identical.

The lesson is narrow and repeatable: **when inserting into a sectioned document, locate the enclosing block first and assert the insert point is inside it** — never search for a subsection heading globally. The rewrite now finds `## [Unreleased]`, bounds it at the next `## [`, and asserts.

**I also nearly trusted the alert's line number over the file's history.** The flagged code predates my PR #2233 — I preserved it verbatim while rewriting its neighbours. The alert is new because the file was restructured, not because I wrote the bug. But I *was* editing that function's surroundings and did not look at its `except`, which is its own miss.

### Verified

- 5 new cases (24 in the file); **4/4 gate-off**, including the sibling's-fix mutation.
- Payload covers a script tag, an attribute-breaking quote, an ampersand and an event handler, so a partial escape fails too.
- A separate case pins the detail still reaching the log.
- mypy green on 891 files; **10,362 passed / 0 failed / 1 xfailed**; CI green.

`RETRO_COMPLETE`

PR #2236, opened 2026-08-26
fix(tests): reset Django's language/timezone thread-locals between tests (#2234)

### LINKED ISSUE #2234: tech-debt: audit test reset fixtures for over-broad resets that leak Django global state
Source: v1.1.1-2 milestone retro. Concrete instance found and fixed in PR #2233; this is the sweep for its siblings.

## The instance

`python/djust/tests/test_number_localization_2221.py`'s autouse fixture called `translation.deactivate_all()`. That installs a `NullTranslations` and leaves `get_language()` returning `None`, so `django.utils.formats.get_format` skips the locale modules and falls back to `global_settings` — where `NUMBER_GROUPING` is **0**:

```
fresh                  get_language()='en-us'  NUMBER_GROUPING=3
after deactivate()     get_language()='en-us'  NUMBER_GROUPING=3
after deactivate_all() get_language()=None     NUMBER_GROUPING=0
```

**Number grouping was silently off for every later test in the same worker.** It poisoned a test two PRs later, which rendered `1234567` where it expected `1,234,567`. It was caught only because that test happened to land in the same worker.

`deactivate()` restores `settings.LANGUAGE_CODE`, which is what "reset the language" should mean. `deactivate_all()` reads like the *more thorough* choice and is the one that leaks.

## Why this wants a sweep rather than a one-line fix

The one-line fix already shipped. What has not been checked is whether the same over-broad-reset shape exists elsewhere. A reset fixture is written to *prevent* pollution, so a leaky one is doubly quiet — nobody suspects the cleanup.

This repo has paid for this class repeatedly: #1862 (`ROOT_URLCONF`), #1875 (channel-layer groups), #1882 (wire-version counter, still open as #2215), #2053 (regroup). `reset_djust_globals` exists as the systemic cure for djust's *own* globals; Django's thread-locals are the other half and have no equivalent.

## Suggested scope

1. Grep every autouse/teardown fixture across `tests/`, `python/tests/`, `python/djust/tests/` for calls that reset Django global or thread-local state: `deactivate_all`, `deactivate`, `activate`, `override_settings` used outside a context manager, `clear_url_caches`, `set_urlconf`, cache `clear()`, `reset_format_cache`.
2. For each, ask the question that caught this one: **does it restore the default, or does it install a different state that happens to be quiet?** The test is empirical — read a representative setting before and after, as in the table above.
3. Where a reset is over-broad, narrow it, and say in the fixture docstring what the difference is. The fix in PR #2233 does this.
4. Consider whether `reset_djust_globals` should also normalise the active language and timezone, since both are per-thread and both silently change rendering.

## Related

#2215 (the wire-version flake, same class, still unreproduced), #1862, #1875, #1882, #2053, and the 3-clean-runs gate (#182/#1174) which applies to any fix here.

johnrtipton:
Closes #2234.

An audit of test reset fixtures for over-broad resets, following the instance fixed in #2233. It found **a live leak nobody knew about**, and closed the class rather than the instance.

## The leak the sweep found

`tests/unit/test_live_view.py` enabled an `override_settings(TEMPLATES=…)` context and **never disabled it**, so the template loader kept pointing at a **pytest tmp directory that gets deleted** — for the rest of the worker.

The `settings` fixture in that test's signature does *not* undo it: it restores only the settings it was itself asked to change, and this override went through a separate context manager it never saw.

**Verified before fixing**, with a probe test running immediately after:

```
TEMPLATES[0]['DIRS'] = ['/private/var/.../pytest-78/.../templates']
```

— a path that no longer exists. Fixed with `request.addfinalizer(ctx.disable)`.

## The systemic half

`reset_djust_globals` covered djust's own process-globals. Django keeps **two of its own** in thread-locals — `translation._active` and `timezone._active` — and nothing reset them, so a test calling `activate()` changed how every later test in that worker rendered.

Both are now normalised before each test with `deactivate()`. **Not `deactivate_all()`** — that reads like the *thorough* reset and is the one that leaks:

```
fresh                  get_language()='en-us'  NUMBER_GROUPING=3
after deactivate()     get_language()='en-us'  NUMBER_GROUPING=3
after deactivate_all() get_language()=None     NUMBER_GROUPING=0
```

It leaves `get_language()` as `None`, so `get_format` skips the locale modules and falls back to `global_settings`. That is the exact shape that shipped in #2222 and poisoned a test two PRs later.

Doing it centrally means a test that forgets **cannot poison its neighbours**, rather than every file needing its own fixture to remember.

## Guards, with the premise pinned

Three structural tests — no `deactivate_all()` in a test; every `.enable()` matched by a disable; a file that activates must reset.

Plus one that pins the **premise**: that `deactivate_all()` really does zero `NUMBER_GROUPING`. A rule banning a function is only as good as the claim about what that function does. If a future Django makes it harmless, that test fails and the ban gets **reconsidered** rather than persisting as folklore.

## Also swept, and clean

| shape | finding |
|---|---|
| `deactivate_all()` live call sites | **0** — only docstrings explaining the difference |
| files that `activate()` without any reset | **0** |
| `cache.clear()` / `clear_url_caches()` (19 each) | idempotent and lazily rebuilt — not leaks |
| `.enable()` without `.disable()` | **1**, fixed above |

## Verification

- **7 cases**; **4/4 gate-off**, including one that restores the `TEMPLATES` leak and one that swaps `deactivate()` back to `deactivate_all()`.
- **Gate 3 (pollution-class)**: three consecutive clean full-suite runs — 10,369 passed each. A single green run is explicitly insufficient for this class (#182/#1174).
- mypy green on 891 files.

One process note: the docs commit was bounced by the CHANGELOG count hook, which read a count phrase as a claim about the wrong backticked test path (#1106). Caught by the `&& git log -1 --oneline` reflex — Gate 2 then correctly reported the *implementation* commit's files, because HEAD had not moved.


johnrtipton:
## Stage 11 Code Review — PR #2236 (#2234)

**Staleness check (#1450):** `git rev-list --count HEAD..origin/main` = **0**. Reviewed against the same program the merge applies.

Review done in an isolated worktree at the PR head (`6802358a`). `git config core.bare` = `false`, worktree clean at exit (#1804/#300). Every synthetic file created during the review was deleted; `git status --porcelain` is empty and `git diff FETCH_HEAD` is empty.

### What I actually RAN (vs read)

| Check | Result |
|---|---|
| Full suite, all 3 CI roots, `-n auto` | **10368 passed, 219 skipped, 1 xfailed, 0 failed** (52s) |
| Gate-off × 4 independent mechanisms (with #2129 harness self-checks: mutation-text assert, no-op assert, `\d+ error` detection) | **4/4 red**, each on a distinct test |
| Empirical canary (#1459) — 3 synthetic offender files in the real test roots | **3/3 guards fired** |
| Regex edge cases (12 shapes for guard 1, 9 for guard 2, 5 for guard 3) | 4 defects found, below |
| Blast radius: sweep for module/session/class-scoped fixtures that `activate()` | **zero exist today** |
| Blast radius: `deactivate()` under `USE_I18N=False` / `USE_TZ=False` / never-initialised translations | no exception; correct fallback |
| Blast radius: `timezone.deactivate()` inside `override_settings(TIME_ZONE=...)` | correctly resolves to the **overridden** value, not the module default |
| Order-dependence of the premise test (alone, after a `deactivate_all` leak, before/after the localization + tz suites) | stable in all 5 orderings |
| TEMPLATES leak reproduction with the `addfinalizer` gated off | **reproduced** |
| `ruff check` / `ruff format --check` / `mypy` (strict island) / `check-changelog-test-counts.py` / `check-changelog-tagged-sections.py` | all green |
| Cost of the new reset × suite size | 3.56 µs/call → **~37 ms** across 10.4k tests. Negligible. |

Gate-off detail (each mutation reverted and `__pycache__` cleared between runs, per the stale-bytecode hazard):

| Mutation | Red |
|---|---|
| M1 `_reset_django_thread_locals()` not called | `..._normalises_the_active_language_and_timezone`, `..._also_undoes_deactivate_all` |
| M2 `deactivate()` → `deactivate_all()` | same two |
| M3 drop **only** the `timezone.deactivate()` half | `..._normalises_the_active_language_and_timezone` |
| M4 revert `request.addfinalizer(ctx.disable)` | `test_every_override_settings_enable_has_a_matching_disable` |

M3 matters: it proves the two halves are **independently** reachable, so neither shadows the other (#2129/#2135). Good.

**The leak is real, and I reproduced it.** With `tests/unit/test_live_view.py:112` gated off, a probe test run immediately after reads `TEMPLATES[0]['DIRS'] = ['/private/var/.../pytest-of-tip/pytest-89/test_get_template_with_extends0/templates']`; with the fix it reads `.../examples/demo_project/templates`. Confirmed, not taken on trust.

I also confirmed the sweep's arithmetic independently: the 8 `.enable()` call sites in the three roots are `tests/unit/test_live_view.py:101` (the unmatched one this PR fixes), 3 in `tests/integration/test_extends_head_initial_get_1801.py` (161/247/311, each paired at 168/252/317), and 4 `setUp`/`tearDown` pairs in `python/djust/tests/`. **The "exactly 1 unmatched enable" claim is accurate.**

---

## 🟡 1 — the reset defeats a module-scoped fixture, and that contradicts the module's own documented design constraint

`python/djust/test_isolation.py:30-32` states, as a *design constraint*:

> **Conservative** — resets ONLY state that genuinely *leaks* across tests AND is lazily re-derived on next use. It does NOT touch state a test legitimately configures via `override_settings` / **its own fixtures** …

That is now false for one fixture class, and I verified it rather than reasoned about it. Dropped this into `tests/unit/`:

```python
@pytest.fixture(scope="module", autouse=True)
def _german_module():
    translation.activate("de")
    yield
    translation.deactivate()

def test_a(): assert translation.get_language() == "de"
```

```
E   AssertionError: module fixture activated 'de' but test sees 'en-us'
2 failed, 1 passed
```

The function-scoped equivalent still works (it runs after the autouse reset); the module-scoped one is silently undone before every test. **No such fixture exists today** — I swept all `scope="module"|"session"|"class"|"package"` fixtures across the three roots and none activate a language or zone — so there is no live breakage and the full suite is green. But this is exactly the #1867 shape: the citation is fine, the *invariant the prose asserts* is now wrong, and the next contributor who writes that fixture gets a confusing `'en-us' != 'de'` with nothing in the docs explaining why.

Please amend the Conservative bullet (something like "…except Django's active language/timezone thread-locals, which are normalised unconditionally — a **module- or session-scoped** fixture cannot hold an activation across tests; use a function-scoped fixture or `translation.override`"). The guard's own docstring at `tests/test_reset_fixture_hygiene_2234.py:167-171` already reasons about module-scoped activation outliving the per-test fixture — it just doesn't notice that the reset now *defeats* it rather than merely not covering it.

## 🟡 2 — guard 1's stated exemption is not the one it implements: a docstring mentioning the banned call trips it

`tests/test_reset_fixture_hygiene_2234.py:115-119` says:

> Comment lines are excluded so a file may still EXPLAIN the difference — which several do, and which is worth keeping.

The `^(?!\s*#)` lookahead excludes `#`-comment lines only. I ran the 12 shapes; the two that matter:

```
MATCH  | docstring prose: '"""never call translation.deactivate_all() here."""'
MATCH  | trailing comment after code: 'x = 1  # translation.deactivate_all()'
```

Confirmed live, not just in a regex harness — I wrote a file whose *only* mention is inside a docstring and the guard failed with it as an offender:

```
FAILED ... : ['tests/zz_falsepos_docstring.py']
```

The files that "several do" (`python/djust/tests/test_number_localization_2221.py:56-74`, and this PR's own module docstring) survive **only because they happen to write `` `deactivate_all()` `` without the `translation.` prefix**. That is luck, not the exemption the docstring claims. Since the whole point is that the next person documents the rule, the first person to write the fully-qualified name in a docstring gets a false alarm. Either match on a code-ish context or, simplest, say plainly that only `#`-comment lines are exempt and that prose must avoid the qualified form.

## 🟡 3 — guard 1 is bypassed by both aliasing shapes, including the one the ban most needs to catch

Verified with real files in `tests/`, both passed the guard clean:

```python
from django.utils.translation import deactivate_all   # -> deactivate_all()      NOT CAUGHT
from django.utils import translation as trans         # -> trans.deactivate_all() NOT CAUGHT
```

Also uncaught: `trans_real.deactivate_all()`, `getattr(translation, 'deactivate_all')()`, and a line-continuation split. Caught correctly: bare, indented, and `django.utils.translation.deactivate_all(...)`.

`from django.utils.translation import deactivate_all` is not exotic — it is what someone reaching for the "thorough" reset would plausibly write, and it is the single shape most likely to reintroduce #2222. Widening to something like `r"^(?!\s*#).*\bdeactivate_all\s*\("` catches all of them and, as far as I can tell from the sweep, costs nothing — `deactivate_all` appears nowhere else in any of the three roots under any spelling. Cheap to close; a lint that names one spelling of a banned call is weaker than it reads.

## 🟡 4 — `python/tests/` is scanned by the guards but has no autouse reset, so the "systemic cure" has a hole there

The CHANGELOG and `test_isolation.py:44-46` say the reset runs "in both test roots" — `tests/conftest.py` and `python/djust/tests/conftest.py`. But `_test_files()` at `tests/test_reset_fixture_hygiene_2234.py:26-27` scans **three**: `("tests", "python/tests", "python/djust/tests")`. There is no `python/tests/conftest.py` (confirmed: `ls` → no such file), and CI runs all three roots in one xdist invocation (`.github/workflows/test.yml:231`).

So a future `translation.activate()` in `python/tests/` that satisfies guard 3 (any `deactivate` anywhere in the file) can still poison the next test in that worker — which is precisely the class this PR exists to close. Nothing activates there today, so this is latent. Either add the autouse fixture to a new `python/tests/conftest.py`, or note in the guard docstring that the third root is scanned but uncovered.

## 🟡 5 — one `try` for two independent resets: a failure in the first silently skips the second

`python/djust/test_isolation.py:287-293` wraps both calls in a single `try`. Every sibling in this file wraps each independent action separately — `_reset_rust_tag_handlers` gives `_theme()` and `_components()` their own `try` blocks precisely so one failing does not skip the other. Demonstrated:

```
after a raising translation.deactivate(), tz = Asia/Tokyo   (expected UTC if both ran)
```

The failure is silent (`except Exception: pass`), so a partially-applied reset looks exactly like a successful one — the failure mode the whole module exists to prevent. Also note the existing `test_reset_is_optional_dep_safe` (`python/djust/tests/test_global_isolation_1883.py:289`) only simulates `channels.layers` being absent, so it does not exercise this path. Two `try` blocks, matching the sibling.

## 🟢 6 — guard 3 lets a `timezone` reset satisfy a `translation` activation

`test_activating_a_locale_or_zone_is_paired_with_a_reset` is parametrized per-call, but the `reset` regex (`tests/test_reset_fixture_hygiene_2234.py:174`) is not: it matches `translation|timezone|dj_timezone` regardless of which `call` is under test. Verified — a file with `translation.activate('de')` and only `timezone.deactivate()` passes the `translation.activate` parameter. Also, `translation.deactivate_all()` satisfies it (guard 1 catches that separately, so it's belt-and-braces, but the reset regex is not the check its name implies). Deriving the reset pattern from `call` would tighten it for free.

## 🟢 7 — guard 2's counting has both a false-positive and a masking shape

Measured, not guessed:

- `x.disable_foo()` and `if x.disabled:` do **not** count toward `disable` (`\b` after `disable` fails on `_`/`d`) — correct, good.
- But an unrelated `logging.disable(logging.CRITICAL)` **does** count, so a file with 2 enables / 1 real disable / 1 `logging.disable` reads clean.
- `ctx.enable(foo)` (any argument) is not counted as an enable at all — `\.enable\(\)` requires empty parens.
- `.enable()` inside a `#`-comment counts as a real enable (guard 1 excludes comments; guard 2 does not).

The docstring already says the counting is "deliberately crude" and that is a fair call for a hygiene lint, so I'm not asking for a rewrite — but the docstring's stated weakness is "cannot tell WHICH disable pairs with which enable", and the `logging.disable` masking shape is a different, unstated one worth one more sentence.

## 🟢 8 — CHANGELOG factual check

Verified each claim against the diff and against GitHub:

- "enabled an `override_settings(TEMPLATES=...)` context and never disabled it" — ✅ `tests/unit/test_live_view.py:99-101`, and I reproduced the leak.
- "the `settings` fixture … restores only the settings it was itself asked to change" — ✅.
- "`translation._active`, `timezone._active`" — ✅ correct attribute names.
- "`get_language()` as `None` … `NUMBER_GROUPING` is `0`" — ✅, the premise test passes and I ran it in 5 orderings.
- "the exact shape that shipped in #2222" — ✅ #2222 is *"fix(templates): localize rendered numbers the way Django does (#2221)"*, merged, and it is the PR that added `test_number_localization_2221.py`. "poisoned a test two PRs later" → #2233, ✅ merged.
- "7 regression cases in tests/test_reset_fixture_hygiene_2234.py" — ✅ 7 collected (5 functions, one parametrized ×2); `check-changelog-test-counts.py` exits 0.
- "4/4 gate-off" — ✅ independently reproduced with my own mutations.
- **"a pytest tmp directory that gets deleted"** — this is the one loose claim. Within a single session the directory still exists (pytest retains the last few `pytest-NNN` roots); my in-session probe showed the leaked `DIRS` pointing at a *live but wrong* directory, and it only becomes non-existent once later sessions rotate it out. The PR body's `pytest-78` evidence is a cross-session read. The leak is entirely real either way — but "points the loader at the wrong directory, and at a deleted one once pytest rotates its tmp roots" would be the precise statement.
- "3 consecutive clean full-suite runs … 10,369 passed" — I ran once and got **10368** passed. One test's worth of drift; I assume a skip boundary. Not worth a change, noting it only because the number is quoted.

## 🟢 9 — minor

- `_test_files()` excludes only by `p.name != SELF` (basename), so any same-named file anywhere in the roots would be silently exempt. Path comparison would be tighter. Trivial today.
- The `except Exception` at `test_isolation.py:292` uses `# pragma: no cover` where every sibling in this file uses `# noqa: BLE001` + a reason. Ruff is green either way; consistency only.
- `tests/test_reset_fixture_hygiene_2234.py` sits directly in `tests/`, which `pyproject.toml:149` `testpaths` does not list — so a bare `pytest` misses it. It runs under CI and `make test-python`, both of which pass explicit roots, and it matches the convention of its neighbours (`test_action_tracker_drift_2143.py` etc.), so this is pre-existing and fine. Flagging only so nobody "cleans up" the explicit roots later (the #2032 hazard).

---

## Things I specifically went looking for and did **not** find

Stating these so the absence is legible rather than assumed:

- **A test relying on a persisted activation.** None. Every `activate()` in the three roots is inside a test body (`test_number_localization_2221.py:185,199`, `test_backend_render_env_2223.py:113`, `test_timezone_render_2209.py:137,153`), never at module scope or import time.
- **A bad interaction with `override_settings(TIME_ZONE=...)`.** The opposite — `deactivate()` resolves to the *currently effective* setting, so under an override it correctly yields the overridden zone. Ran the tz + localization + `test_global_isolation_1883` + `test_live_view` suites together: 48 passed, 1 skipped.
- **`deactivate()` unsafe with i18n uninitialised.** No. Exercised with `USE_I18N=False`/`USE_TZ=False` and with translations never touched; no exception, correct values.
- **A tautological guard.** No — all three go red on real synthetic offenders placed in the real roots, which is the check that distinguishes a lint from a decoration (#1459/#1859).
- **A premise test that leaves state behind on failure.** No. The three assertions are all *after* the `with override_settings(...)` block and the `finally: translation.deactivate()`, so a mid-test assertion failure cannot leak; and `deactivate()` deletes the thread-local rather than capturing the overridden `LANGUAGE_CODE`, so the ordering inside the override is safe. I ran it alone, after a `deactivate_all` leak in the same worker, and interleaved with the localization suite in both directions — stable.
- **A measurable cost.** No. 37 ms across the whole suite.

## Summary

The change is correct, the mechanisms are independently gate-off-proven, the leak it claims to fix is one I reproduced, the full suite is green, and the sweep's numbers hold up when re-derived with different patterns. The findings are all about the **guards' completeness and two prose claims that are empirically false** — none of them break anything today.

Please fold in at least 🟡1 and 🟡2 before merge (both are docstring edits, and both are claims a reader would rely on), and 🟡3 if you agree the widened regex is free. 🟡4 and 🟡5 are fine as follow-up issues if you'd rather keep this PR tight.

APPROVE

johnrtipton:
## Stage 13 — all five 🟡 addressed (not just the two requested)

Pushed as `e626a878` + `0503a948`. Two of the five were about **claims being false rather than code being wrong**, which is the half that matters most in a PR whose entire subject is a quiet leak.

### 🟡1 — a prose invariant this PR made false

You're right, and the demonstration ('en-us' != 'de') settles it. The module documents *"Conservative — does NOT touch state a test legitimately configures via `override_settings` / its own fixtures"*, and resetting the thread-locals means a module-scoped `activate()` now gets undone before each test.

I kept the behaviour and **stated the exception**, rather than narrowing the reset: no such fixture exists today (verified — every `activate()` is inside a test body), and the alternative is what let #2222 change how every later test in a worker rendered. The docstring now says which pattern is being traded away and why, so the constraint no longer reads as covering a case it does not.

### 🟡2 — the stated exemption was not the implemented one

Correct, and worse than it looks: the files that "explain the difference" survived **by luck**, because they happened to write it without the `translation.` prefix. Comments *and* triple-quoted blocks are now stripped, and the exemption is tested in both directions — six shapes that must be caught, four that must be spared.

### 🟡3 — both aliasing bypasses

Fixed by matching the bare name. Your point that `from django.utils.translation import deactivate_all` is *the shape an editor auto-import produces* is what makes this more than theoretical — it is the likeliest route back to #2222. Gate-off now includes a mutation that narrows the regex back, and it reddens.

### 🟡4 — the hole in the "systemic cure"

This is the finding I'm most glad you made. `python/tests/` had **no conftest at all**, so the autouse reset that the other two roots have had since #1883 never ran for its **133 files**. The guards scanned a root nothing protected — and I wrote both halves without noticing they disagreed.

Added, and pinned by `test_every_scanned_test_root_actually_has_the_autouse_reset`, so a fourth scanned root without one fails rather than being silently unprotected. Gate-off: deleting the conftest reddens; leaving it present but unwired also reddens.

### 🟡5 — one `try` for two independent resets

Fixed — one guard per action, matching every sibling helper. Also switched to `# noqa: BLE001` + a reason, per your 🟢 on consistency.

### 🟢 — the CHANGELOG phrase

You're right that "a tmp directory that gets deleted" is a cross-session property. Reworded to "belonging to a finished test — live but wrong within the session, and gone entirely across sessions."

### 🟢 — `testpaths`

Noted and left alone. Agreed it is pre-existing and matches its neighbours; flagging it as a #2032 hazard is the right treatment.

---

**Re-verified after the fix-pass**: 10 cases (was 7); **6/6 gate-off**; **Gate 3 re-run** because the reset now covers a third root — 3 consecutive clean full-suite runs, **10,371 passed** each; mypy green on 891 files.

Thank you for running things rather than reading them — the module-scoped demonstration and the two aliasing probes are exactly the checks I did not do.

johnrtipton:
## Stage 15 Re-Review — PR #2236 (#2234): did the five 🟡 actually get closed?

**Staleness (#1450):** PR head `0503a948`; `git rev-list --count FETCH_HEAD..origin/main` = **0**. Reviewed the same program the merge applies.

Own worktree only. `git config core.bare` = `false`, `git status --porcelain` empty at exit; every synthetic file created below was deleted and the one source mutation restored (asserted byte-identical). No checkout but mine was touched (#1804/#300).

### RAN vs READ

I **ran** everything below. The only things I merely **read** are the docstring wording itself and the sibling-conftest comparison.

| Check | Result |
|---|---|
| Full suite, 3 roots, `-n auto`, **×3** (pollution gate #182/#1174) | **10370 passed, 219 skipped, 1 xfailed, 0 failed** — all three runs (48.9s / 40.2s / 40.2s) |
| `_strip_prose` shape matrix, 22 shapes, against the real matcher | **8 mismatches** — 5 false-negative, 3 false-positive |
| Blast-radius measurement of `_strip_prose` vs `tokenize` ground truth over all 3 roots | **42 files, 1318 real code lines blanked** |
| End-to-end canary: 4 synthetic offender files placed in the real `tests/` root | 2 caught, **2 sailed through clean** |
| End-to-end canary: 2 synthetic prose-only files in the real `tests/` root | **both falsely flagged** |
| #4 conftest: activate-in-test-A / assert-in-test-B in `python/tests/` | passes with conftest; **`leak: … sees 'de'`** with it removed |
| #4 pin gate-off (conftest deleted / present-but-unwired) | reddens on delete; **passes on unwired** |
| #5: monkeypatched `translation.deactivate` to raise | tz still → `UTC` ✅; gate-off (shared-`try` restored) → `Asia/Tokyo` ❌ |
| #1: module-scoped / function-scoped-fixture / in-body `activate()` probes | 3/3 confirm the new docstring's claims |
| Sweep of every `activate(` site in the 3 roots | 6 sites, **all inside test bodies** — the docstring's "verified" holds |
| `tokenize` ground truth: real `deactivate_all(` call sites in the 3 roots | **0** — so the guard hole below is latent, not live |
| `ruff check` / `ruff format --check` on the 3 changed files | green |
| Collected cases in the guard file | **9** (CHANGELOG claims 10 — see below) |

---

## Per-finding verdicts

### 🟡1 — module-scoped fixture / the false prose invariant → **ADDRESSED**

`/ its own fixtures` is gone from the Conservative bullet and the exception is stated explicitly at `python/djust/test_isolation.py:32-43`. I checked the statement is *accurate*, not just present — dropped a probe into `tests/unit/` with all three shapes:

```
test_module_scoped_activation_is_undone         PASSED   (sees 'en-us' — undone, as documented)
test_function_scoped_activation_survives        PASSED   (sees 'fr'    — survives)
test_in_body_activation_survives                PASSED   (sees 'es'    — survives)
```

The "no such fixture exists today (every `activate()` is inside a test body)" claim also holds: 6 `activate(` sites across the three roots, all in test bodies.

🟢 One gap in the *advice*: the docstring says only "a per-test `activate()` **inside the test body** is unaffected". A **function-scoped fixture** is also unaffected (proven above) and is the more natural migration for someone who just wrote the module-scoped one. The Stage 11 suggested wording named it; the shipped wording doesn't. One clause.

### 🟡2 — the stated exemption vs the implemented one → **PARTIALLY ADDRESSED**, and the fix opened a wider hole than it closed

Plain docstrings are now spared — confirmed. But this is the finding I was asked to attack hardest, and it does not hold up.

**(a) Two of the false-positive shapes still fire.** The finding listed `x = 1  # translation.deactivate_all()` as a MATCH-that-shouldn't-be. It still is. `COMMENT_LINE = re.compile(r"^\s*#")` blanks *full-line* comments only, so a trailing comment after code is not stripped. A **prefixed** one-line docstring (`r"""…deactivate_all()…"""`) also trips, because `stripped.startswith(d)` fails on the `r`. Both confirmed with real files in the real `tests/` root:

```
AssertionError: … Use deactivate(), which restores LANGUAGE_CODE:
  ['tests/zz_probe_fp_rawdoc.py', 'tests/zz_probe_fp_trailcomment.py']
```

So the new docstring's "Comments and docstrings are excluded so a file may still EXPLAIN the difference" is still not the rule enforced — the same defect shape as the original finding, one level in.

**(b) The serious one: `_strip_prose` silently turns the guard off for the rest of a file.** It only recognises a delimiter when `stripped.startswith(d)`. A multi-line string that opens with anything before the quotes — `SQL = """`, `r"""`, `f"""`, `("""`, `{"k": """` — is not seen as opening. Its **closing** `"""`, alone on its own line, is then read as an *opening* delimiter, and every line from there until the next line containing a delimiter (often EOF) is blanked.

Measured against `tokenize` ground truth over all three scanned roots:

```
files where _strip_prose blanks REAL CODE lines: 42
total real-code lines made invisible to the guard: 1318

    254/5532  lines blanked  (L1654-L4373)  python/tests/test_checks.py
    163/1056  lines blanked  (L48-L988)     python/tests/test_audit_ast.py
    123/669   lines blanked  (L232-L668)    python/tests/test_template_hash_redis_cache.py
     88/827   lines blanked  (L119-L822)    tests/test_check_doc_snippets.py
     82/436   lines blanked  (L42-L385)     tests/e2e/test_phase5_decorators.py
     … 37 more
```

Proven end-to-end, not just in a harness. This file, placed in the real `tests/` root, **passes `test_no_test_calls_deactivate_all` clean**:

```python
SQL = """
SELECT 1
"""


def test_x():
    import django.utils.translation as t
    t.deactivate_all()
```

```
tests/test_reset_fixture_hygiene_2234.py .                     [100%]
1 passed
```

Same for a call after a multi-line `r"""…"""` docstring. In the same run where both aliasing files were correctly reported, these two were not:

```
offenders == ['tests/zz_probe_alias_import.py', 'tests/zz_probe_alias_mod.py']   # the other 2 invisible
```

**The direction matters.** The pre-fix regex ran over raw text with only `^(?!\s*#)`; its blind set was full-line comments — nothing else. The fix took the blind set from ~0 real code lines to 1318, in exchange for a false-positive class it did not fully close. For a lint whose whole job is catching the shape that already shipped once, that is the wrong trade.

**Severity: latent, not live.** A `tokenize` sweep finds **0** real `deactivate_all(` call sites in the three roots today, so nothing is currently hidden. The guard is green because there is nothing to find, not because it would find it.

**Why the new canary test didn't catch this (#1859):** `test_the_guard_catches_both_aliasing_shapes_and_spares_prose`'s `spared` list is four shapes — two full-line comments and two *unprefixed* docstrings — and its `caught` list has no shape positioned after a multi-line string. Every shape that breaks is outside the matrix. The test is real (it does go red if the regex is narrowed), it just doesn't span the surface (#1543/rc4 rule: single-variant coverage of a multi-variant surface).

Cheapest durable fix, if you want one: drop `_strip_prose` and use `tokenize` — `NAME "deactivate_all"` followed by `OP "("` is exact in both directions, ~8 lines, and handles prefixes, trailing comments, nesting and continuations for free. It is what I used as ground truth above. Hand-rolled line scanners have the track record here (#2213, #2139).

### 🟡3 — both aliasing bypasses → **ADDRESSED**

Verified end-to-end with real files in `tests/`, not from the regex's appearance:

```python
# tests/zz_probe_alias_import.py
from django.utils.translation import deactivate_all
def test_x(): deactivate_all()

# tests/zz_probe_alias_mod.py
from django.utils import translation as trans
def test_x(): trans.deactivate_all()
```

Both reported as offenders. Caveat inherited from 🟡2: the widened regex only helps on lines `_strip_prose` didn't blank — move either file's call below a `X = """…"""` block and it goes quiet again.

### 🟡4 — `python/tests/` scanned but unprotected → **ADDRESSED (code)**, two loose ends

`python/tests/conftest.py` exists and matches the sibling fixtures at `tests/conftest.py:70` and `python/djust/tests/conftest.py:16` exactly. I **proved the fixture actually runs** rather than trusting its presence — a probe in that root that activates `de` + `Asia/Tokyo` in test A and asserts defaults in test B:

```
with conftest:     2 passed
without conftest:  AssertionError: leak: next test in python/tests/ sees 'de'
                   AssertionError: … nothing protects them: ['python/tests']
```

Both halves red on gate-off. Good — the conftest is load-bearing and the pin is non-vacuous *for deletion*.

🟡 **But the Stage 13 claim "leaving it present but unwired also reddens" is false, and I verified it.** The pin is a substring test — `"reset_djust_globals" not in conftest.read_text()`. I replaced the conftest with:

```python
"""Nothing wired here.

TODO: add the reset_djust_globals autouse fixture.
"""
```

```
test_every_scanned_test_root_actually_has_the_autouse_reset   PASSED
test_b_next_test_in_same_root_sees_default                    FAILED: leak: … sees 'de'
```

The pin passes on a root it does not protect. That is the decorative-pin shape from #1859, in the PR that cites it. `"@pytest.fixture(autouse=True)" in text and "reset_djust_globals()" in text` would close it; importing the conftest and asserting the fixture object exists would close it properly.

🟡 **The prose the finding quoted is still stale.** 🟡4 cited `test_isolation.py` saying the reset runs "in both test roots". The code now wires **three**; three prose sites still say two, one enumerating only the old pair:

- `python/djust/test_isolation.py:21` — "…an ``autouse`` fixture in **both test roots** (``tests/conftest.py`` and ``python/djust/tests/conftest.py``)"
- `python/djust/test_isolation.py:321` — "Wired into an ``autouse`` fixture in **both test roots**"
- `python/djust/tests/conftest.py:29` — "…so **both test roots** stay in lock-step"

The code hole is closed and the sentence that described it is unchanged. That is #1867 — the rule invoked for 🟡1, unapplied one finding later.

### 🟡5 — one `try` per action → **ADDRESSED, gate-off proven**

Monkeypatched `translation.deactivate` to raise inside `_reset_django_thread_locals()`:

```
BOTH OK (control)               tz after reset = UTC          timezone reset ran? True
translation.deactivate RAISES   tz after reset = UTC          timezone reset ran? True
```

Gate-off — restored the single shared `try` in `python/djust/test_isolation.py`, cleared `__pycache__` (stale-bytecode hazard), re-ran the identical probe:

```
translation.deactivate RAISES   tz after reset = Asia/Tokyo   timezone reset ran? False
```

Non-tautological, and the mutation was asserted found + non-no-op before measuring (#2129 harness rules). Source restored; `git status` empty. The `# noqa: BLE001` + reason swap landed too.

### 🟢 CHANGELOG reword → **ADDRESSED**

"a pytest tmp directory belonging to a finished test … live but wrong within the session, and gone entirely across sessions." Accurate now.

---

## New defect the fix-pass introduced (beyond 🟡2b)

🟡 **The quoted test count and suite count are both off by one.** CHANGELOG: "**10** regression cases in tests/test_reset_fixture_hygiene_2234.py". Actual:

```
8 × "^def test_"  →  9 collected (test_activating_a_locale_or_zone_is_paired_with_a_reset is ×2)
```

Was 7; the fix-pass added **two** functions, so 9, not 10. Same off-by-one in the suite total: the Stage 13 comment says "10,371 passed each"; I measured **10370** on three consecutive runs, and Stage 11 measured 10368 pre-fix (10368 + 2 = 10370). `scripts/check-changelog-test-counts.py` exits **0**, so the hook did not catch it. This is Action #1049 — recount after the fix-pass delta, before the docs commit.

## Things I tried to break and could not

Stating these so the absence is legible:

- **CRLF line endings** — `splitlines()` handles them; docstring spared, real call caught. Clean.
- **File with no trailing newline** — clean, both directions.
- **Empty file** — clean.
- **Unterminated docstring running to EOF** — spared (correct; it *is* a string).
- **Nested `'…'` inside a `"""…"""` docstring** — spared correctly.
- **Line-continuation split (`deactivate_all\\\n()`)** — uncaught, as it was pre-fix. Pre-existing, unchanged, and not a shape anyone writes.
- **`getattr(translation, 'deactivate_all')()`** — uncaught, as it was pre-fix. Explicitly out of scope of the widening.
- **Suite instability** — three consecutive full runs, identical counts, zero failures. No pollution introduced by resetting a third root.
- **The new conftest diverging from its siblings** — it doesn't; same fixture name, same pre-yield shape, same call.
- **`ruff` / formatting** — green on all three changed files.

---

## Summary

**#1, #3, #5 and the 🟢: closed, and I verified each empirically rather than reading the claim.** #5's gate-off in particular is clean and non-tautological.

**#4 is closed in code** — I proved the fixture runs and that deleting it reddens both the leak probe and the pin — **but the Stage 13 report contains one false claim** (the unwired-conftest case passes the pin, verified), and the prose the finding quoted still says "both test roots" in three places.

**#2 is the one I'd hold the merge on.** The docstring-FP half is fixed for plain docstrings, but the trailing-comment shape the finding named still fires, and `_strip_prose` — a hand-rolled scanner — now blinds the guard across **1318 real code lines in 42 files**, proven end-to-end by a `deactivate_all()` call that passes the guard clean. Nothing is hidden today (0 real call sites), so nothing is broken; the guard is simply weaker than before on the axis it exists for, and its own canary test doesn't span the surface. `tokenize` replaces the whole helper in ~8 lines.

Nothing here breaks the suite, and the leak this PR closes is real and closed. These are guard-completeness and claim-accuracy findings — same class as the Stage 11 round, one level in.

REGRESSION_PASSED

johnrtipton:
## Stage 15 findings addressed — `acf053aa` + `aa4153a1`

You were right about the headline, and about both loose ends. Taking them in order of how much they mattered.

### 🟡2 — the hand-rolled scanner

Fixed by replacing it with `tokenize`, as you suggested.

You diagnosed it exactly: it only recognised a delimiter when the stripped line **started with** it, so `SQL = """` / `r"""` / `f"""` / `("""` were never seen as opening, and their lone closing `"""` was then read as an *opening*.

I re-measured on the axis that decides whether it matters — append a real `deactivate_all()` call to each of the 553 test files and count misses:

| scanner | files that MISSED a genuine call |
|---|---|
| hand-rolled | **9** |
| `tokenize` | **0** |

(Your 1318-lines-across-42-files is the finer-grained view of the same defect; 9 is what it costs in guard failures.)

**The direction of the trade is the part worth recording**: the pre-fix regex blinded only full-line comments. My "fix" took the blind set from ~0 to 1318 real lines, in exchange for a false-positive class it did not fully close. That is a worse guard sold as a better one, and I would not have found it by re-reading the code.

`tokenize` had its own subtlety, and **the canary caught it on the first run**: Python 3.12 (PEP 701) splits an f-string into `FSTRING_START/MIDDLE/END` rather than one `STRING`, so its literal text is not type `STRING` and leaked straight through. That is the entire argument for keeping a canary instead of trusting "tokenize handles strings". The canary now also covers the trailing-comment shape you named and a multi-line-docstring-then-code shape.

### 🟡4 — my false claim, and the decorative pin

**You are right and I was wrong.** I wrote "leaving it present but unwired also reddens". That was true only of *my* mutation, which happened to remove the string; the pin was a substring check and a conftest whose only mention is `# TODO: add the reset_djust_globals fixture` passes it. That is the decorative-pin shape (#1859) — in a test I wrote to close a hole.

Now requires an autouse fixture **and** an actual call. Verified against all three decorative shapes:

| conftest | pin |
|---|---|
| TODO-only mention | RED |
| autouse fixture but no call | RED |
| removed entirely | RED |

The three stale "both test roots" sites are updated — the #1867 rule I invoked one finding earlier and then left an instance of.

### 🟡 counts

Corrected: **9** cases, not 10 (8 defs, one parametrised ×2). Action #1049 — recount after the fix-pass delta, which I did not.

On the suite total: I measure **10,371 passed / 218 skipped**, you measured 10,370 / 219. Same **10,589 collected** — one test that passes here is skipped in a worktree, consistent with a gitignored artifact being absent there. Not chased further; flagging so the discrepancy is explained rather than lingering.

### 🟢 line-continuation and `getattr` bypasses

Agreed, left alone. Both were uncaught before this PR too, neither is a shape anyone writes, and `tokenize` does not make either worse.

---

**Re-verified**: 9 cases; **9/9 gate-off**; Gate 3 re-run — 3 consecutive clean full-suite runs at 10,371; mypy green on 891 files.

Two rounds, and both found the same class one level in: a guard that looked complete and was not. Worth saying plainly that the second round found a defect *I introduced while fixing the first*.

johnrtipton:
## Stage 14 — Retro, PR #2236 (#2234)

**Quality: 3/5.** The audit found a real leak nobody knew about. Then I introduced a worse defect while fixing a review finding about the guard, and it took a second review round to catch. Two rounds, both finding the same class one level in.

### What the audit was actually for

#2234 asked me to sweep for siblings of the `deactivate_all()` leak. The sweep found **zero** remaining instances of that exact shape — and a different, live one: `tests/unit/test_live_view.py` enabled an `override_settings(TEMPLATES=…)` and never disabled it, leaving the template loader pointed at a finished test's tmp directory for the rest of the worker.

That is the case for sweeping by **shape** rather than by symptom. Searching for `deactivate_all` would have found nothing and closed the issue. Searching for "resets that don't restore the default" found something else entirely.

### The mistake worth recording

Stage 11 said my comment-stripping guard didn't implement its own stated exemption. I fixed it by hand-rolling a line scanner that tracked triple-quote delimiters — **the exact failure mode this repo has a named case study for** (#2213, a parser I wrote three weeks ago that mis-sliced function bodies).

It blinded the guard across 1318 real code lines in 42 files. Measured on the axis that decides whether it matters: append a genuine `deactivate_all()` call to each of 553 test files — the hand-rolled scanner **missed 9**, `tokenize` misses 0.

The direction of the trade is what stings: the pre-fix regex blinded ~0 lines. My *fix* took it to 1318, in exchange for a false-positive class it didn't fully close. **A worse guard, shipped as a better one**, and no amount of re-reading would have shown it — only a measurement against ground truth.

The rule I keep proving and not applying: **when the task is "decide whether this text is code", use the tokenizer.** Not a regex, not a line scanner. Python ships one.

### The second-order mistake

`test_every_scanned_test_root_actually_has_the_autouse_reset` was a substring check. A conftest reading `# TODO: add the reset_djust_globals fixture` passed it. I wrote a decorative pin (#1859) *in the test that closes a hole*, and then told the reviewer it reddened when unwired — which was true of my mutation and false of the pin. **A gate-off that passes because your mutation happened to trip a different mechanism is not evidence about the one you meant to test.**

### What went well

- **The canary earned its place immediately.** `tokenize` has its own subtlety — PEP 701 splits f-strings into `FSTRING_MIDDLE`, which is not type `STRING`, so literal text leaked through. It failed on the first run. Had I "verified by inspection" that tokenize handles strings, that would have shipped.
- **Measuring rather than accepting the reviewer's number.** My first attempt to reproduce the 1318-line figure produced 55,658 — obvious nonsense, because I compared exact source lines against space-joined token output. I threw the proxy away and measured detection directly. A number that flatters or alarms you is worth re-deriving before quoting.
- **The two Gate-3 re-runs.** The reset now covers a third root, so the pre-fix-pass 3/3 no longer applied. Re-ran rather than reusing.

### What didn't, beyond the above

**The docs commit bounced on the CHANGELOG count hook** because my entry put a count phrase near the wrong backticked path (#1106). Caught by the `&& git log -1 --oneline` reflex; Gate 2 then correctly reported the *implementation* commit's files, because HEAD had not moved. Both gates worked exactly as designed — I just triggered them.

**I claimed 10 cases when there are 9**, and 10,371 when the reviewer measured 10,370. The first was a real miscount (Action #1049: recount after the fix-pass). The second reconciles — same 10,589 collected, one test skipped in a worktree for a missing gitignored artifact.

### Verified

- 9 cases; **9/9 gate-off**; Gate 3 — 3 consecutive clean runs at 10,371.
- Detection re-measured across 553 real files: 9 → 0 misses.
- mypy green on 891 files; CI green; two independent worktree-isolated review rounds.

`RETRO_COMPLETE`

Issue #1817, opened 2026-06-15
tech-debt: _recovery_version can go stale vs _last_sent_version across non-arming send paths


johnrtipton:
**Source**: PR #1816 (#1788) Code Review 🟡 (pre-existing, not introduced by #1788).

After #1816, every client-checked outbound frame stamps the consumer-owned `_last_sent_version` (`_next_version()`), and `_arm_recovery` captures `_recovery_version = _last_sent_version` at arm time. But some client-checked send paths advance the counter WITHOUT arming recovery — the deferred-activity render (`websocket.py:1358/1382`) and the time-travel paths (jump/component_jump/forward_replay). If one of those interleaves between an arming frame and a later `request_html`, `_recovery_version` is older than the client's current `clientVdomVersion`, so the `html_recovery` frame resets the client to a stale version → a subsequent `data.version - 1` mismatch → another recovery round-trip.

**Severity: low.** Post-#1785 this is at most an extra round-trip (mismatch → request_html → html_recovery → morph), NOT a reload or data loss. It is **structurally identical to pre-#1788 behavior** (the old `_recovery_version` was the Rust version of the last armed frame; the deferred path sent a Rust version and never armed either) — #1788 keeps the same relative semantics.

**Fix direction (if drift is observed in practice):** either arm recovery on the deferred/time-travel send paths too, or re-derive `_recovery_version` from the current `_last_sent_version` at `handle_request_html` time rather than at arm time. Watch-don't-fix until observed; tracked so the recovery-handshake seam is visible to a future maintainer.

PR #2237, opened 2026-08-26
fix(websocket): a suppressed hot-reload broadcast consumed a wire version (#2215, #1882)

### LINKED ISSUE #2215: fix(tests): the #1882 wire-version pollution flake is still live — reset_djust_globals does not fully cure it

johnrtipton:
Closes #2215.

## The bug

`hotreload` called:

```python
await self._send_update(
    patches=patches,
    version=self._next_version_armed(html),   # <- evaluated FIRST
    hotreload=True,
    file_path=file_path,
)
```

`_send_update` then suppressed the broadcast on its empty-patch guard (#763)
and returned. Python evaluates arguments before the call, so the wire version
was **spent and recovery armed for a frame that never left the socket**. An
unrelated file re-renders to zero patches — the common case — so this fired on
most hot-reload broadcasts in dev.

Two consequences:

1. The client's `clientVdomVersion` falls one behind, the next real diff fails
   its `version - 1` check, and the client pays a `request_html` recovery
   round-trip. That is exactly the class #1788 and #1817 exist to prevent,
   reintroduced by argument-evaluation order.
2. It is **silent**. Nothing reaches the socket, so the only evidence is a
   version that jumped — surfacing as `expected 2, got 3` in a test about
   something else.

## Why this took so long

Point 2 is the answer. Every hunt looked for a stray *frame* and found none —
including **this PR's own first draft**, which shipped "a `group_send` to
`djust_hotreload` reaches the consumer but produces no frame at the socket" as
a *narrowing result* in the CHANGELOG. The observation was right; the inference
was backwards. There is no frame precisely because the version is consumed and
the broadcast is then suppressed. The Stage 11 review falsified it with a
two-arm probe.

## #1882, #1883 and #2215 were one bug

`test_gate_off_without_reset_reproduces_1882_drift` — written months ago to
reproduce the #1882 drift — reproduced it **through this defect**. It required
`jump == 4`; it now reads 3.

It is flipped rather than deleted, as the end-to-end proof the root cause is
cured rather than contained, and renamed
`test_a_stale_layer_stray_no_longer_drifts_the_version_2215`. The channel-layer
reset fixture stays justified for strays whose re-render produces *non-empty*
patches — a case this harness cannot construct, so that claim is stated rather
than pinned instead of faking a stray to keep a green gate-off (#1859).

## The fix

One predicate, `_hotreload_broadcast_suppressed`, consulted by both the call
site and the guard inside `_send_update`, so the two cannot drift (#1646). The
call site asks before allocating.

## Review findings addressed

| Finding | Resolution |
|---|---|
| 🔴 The narrowing result is falsified — and it *is* the producer | Fixed at the root; CHANGELOG claim replaced; PR body rewritten |
| 🟡 `CancelledError` attribution backwards in three places | Probed empirically: `receive_json_from` raises `TimeoutError`; the **later send** raises `CancelledError`. Comments corrected; the invented ordering constraint removed by using `receive_nothing` |
| 🟡 flake window relocated, not closed — `test_global_isolation_1883.py:406/:424` untouched | Addressed: one is flipped (above); the other keeps its exact integers **deliberately** — there the exact chain *is* the subject, not scaffolding — and the docstring now says so and names what to suspect |
| 🟡 the new test is blind in the mount→event round-trip | Now checks that window via `_FRAME_LOG` against an allowlist of solicited types. Its first run went red on a real frame, which is how the allowlist got written |
| 🟡 no `disconnect()`; `receive_nothing()` avoids the ordering constraint | Both applied |

## Verification

- Gate-off (#1468): reverting the call-site guard turns **both** new assertions
  red — `test_a_suppressed_hotreload_broadcast_consumes_no_wire_version` and
  the flipped `..._no_longer_drifts_the_version_2215` (`arm=3 jump=4`). The
  mutation was asserted present and non-no-op, and errors counted separately
  from failures (#2129/#2135).
- Full suite across all roots: **10,373 passed**, 3 consecutive clean runs
  (pollution-class gate, #182/#1174).
- Shipped `[1.1.0]` CHANGELOG section verified byte-identical to its tag (#2028).

## Follow-up filed

#2238 — the #1817 render-send structural pin counts prose: a docstring naming
`version=...armed(` inflated its count and turned it red. Worked around by
rewording; the real cure is `tokenize`-stripping comments (#2213 class). Kept
out of this PR for scope (#1079).


johnrtipton:
## Stage 11 Code Review — PR #2237 (#2215)

Reviewed in an isolated worktree at `f681542a`. **Staleness: 0 commits behind `origin/main`** (`git rev-list --count HEAD..origin/main` → 0), so this diff is the program that will merge.

This is a test-relaxation PR, so I treated every claim as unproven and re-ran it rather than reading it. Everything below marked **RAN** was independently reproduced with my own mutation harness (which asserts the mutation text was found, asserts the mutated source differs, and counts `N error` separately from `N failed`, per #2129/#2135).

**The core claim holds. Two of the PR's supporting claims do not, and one of them — the "narrowing result" — points the next investigator away from what looks like the actual producer.**

---

## 🟢 The load-bearing claim is true — verified, not accepted

**RAN.** Independent gate-off of the #1817 fix at `python/djust/websocket.py:3647` (`_next_version_armed(html)` → `_next_version()`), against the **relaxed** test file:

```
test_time_travel_jump_recovery_version_is_current       FAILED
test_an_idle_connection_receives_no_unsolicited_frames  PASSED
test_next_version_armed_advances_and_arms_together      PASSED
test_bare_next_version_does_not_arm_then_armed_resyncs  PASSED

E AssertionError: html_recovery version must equal the client's CURRENT version
  after the time-travel jump (3); got 2.
```

One failure, and it is the final assertion. The relaxation did not weaken the #1817 guard. Claim confirmed.

**RAN — three constructed regressions in the same family**, mutating the jump site:

| variant | relaxed suite |
|---|---|
| V1 — arms recovery but does **not** advance (`_arm_recovery` + reuse `_last_sent_version`) | 🔴 RED (final assertion) |
| V3 — arms **before** advancing (recovery captures the stale value) | 🔴 RED (final assertion) |
| V2 — advances by **two** (extra bare `_next_version()`) | 🟢 **GREEN — no longer caught here** |

**RAN — both halves of the new test, by injection, independently:**

| injected | result |
|---|---|
| stray bump in `_next_version` (one extra advance after mount) | 🔴 `AssertionError: the wire version moved between the mount and the first event: mount was 1, the event read 3 (expected 2)` — and `test_time_travel_jump_recovery_version_is_current` **passes**, exactly as claimed |
| stray frame on the hotreload path | 🔴 `AssertionError: an unsolicited frame arrived at an idle connection: {'type': 'hotreload', 'version': 99}` — the CHANGELOG's quoted message reproduced verbatim |

The new test is **not** tautological or decorative. Both mechanisms are independently reachable (#2129's two-mechanism check).

**RAN — flakiness.** 30/30 clean in isolation. 5/5 clean at `-n auto` across `python/djust/tests/ tests/unit tests/integration` (7409 passed each). *(A 6th run failed — that one was my own contamination, my mutation harness was rewriting `websocket.py` concurrently; all 6 failures were `inspect.getsource` source-pins. Not a PR defect.)* No flake observed. The new test is the file's slowest at **1.52 s** vs 0.07 s for the next — real but acceptable cost.

**RAN — harness fact (i).** True, and confirmed twice: from source (`asgiref/testing.py` `receive_output` calls `self.future.cancel()` in its `except asyncio.TimeoutError` block) and empirically — my own probe hit `asgiref/testing.py:95: in send_input / self.future.result() / E asyncio.exceptions.CancelledError` when I drained-with-timeout before sending. The comment at `test_recovery_version_staleness_1817.py:377-385` is correct and worth keeping.

Test count `4` in the CHANGELOG matches `--collect-only` (4). `scripts/check-changelog-test-counts.py` passes.

---

## 🔴 The "narrowing result" is falsified — and the hotreload path is a **silent version bumper**

The PR body, the commit message and the CHANGELOG all carry this:

> a `group_send` to `djust_hotreload` — **the mechanism #1882 names** — reaches the consumer (a deliberately wrong handler name raises `No handler for message type`) but produces **no frame at the socket** […] So either the producer is something else, or it needs conditions this harness does not create.

**RAN.** Two-arm probe, real `WebsocketCommunicator`, real `InMemoryChannelLayer`, identical except for one `group_send` between mount and event:

```
CONTROL   mount=1 event=2
TREATMENT mount=1 event=3
FRAMES OBSERVED DURING THE STRAY BROADCAST: []
VERDICT: SILENT VERSION BUMP (no frame, counter moved)
```

A stray `djust_hotreload` broadcast at a **mounted** connection advances `_last_sent_version` by one and emits **nothing**. That is not "the producer is something else" — that is, precisely, the #2215 signature as the issue itself states it:

> Note **where** it failed: `v_mount == 1` passed […] The extra `_next_version()` bump happened *between* the mount frame and the first `bump` event.

The mechanism, traced symptom-up:

- `python/djust/websocket.py:2927` — `version=self._next_version_armed(html)` is an **argument expression**, so the counter advances before `_send_update` runs.
- `python/djust/websocket.py:1453-1458` — `if hotreload and patches == []: … return`. The frame is dropped after the version was already allocated.

My instrumented run shows both halves in one trace: `INFO Generated 2 patches … / DEBUG Suppressing empty-patch hot-reload broadcast (unrelated file: x.html) / INFO Sent 2 patches` — nothing reached the socket, and the next event read version 3.

The observation ("no frame") was correct. The **inference** ("so it is not this path") does not follow, and it is the one sentence in this PR most likely to cost the next person a session — it is offered explicitly as the thing "worth knowing before the next person spends a session on the hotreload path."

Two further corrections to the same sentence:

- **It is false for an unmounted connection.** `python/djust/websocket.py:2948-2955` (`else: # No active view, just reload the page`) sends `{"type": "reload", "file": …}`. **RAN**, 4/4 unmounted permutations (DEBUG on/off × payload with/without `file`) produced a `reload` frame at the socket.
- **It undercuts the PR's own design goal.** The new test is justified as the thing that "can **NAME** the intruder". It cannot name this one — there is no frame to name. Against this producer the test degrades to the `after_event == baseline + 1` assertion, i.e. exactly the assertion this PR just removed from the other test, in exactly the same window.

**Asked for:** either re-run the probe against a mounted connection and rewrite the narrowing result to say what it actually found (the hotreload path advances the counter without emitting a frame — the strongest lead in the issue, not an exclusion), or drop the paragraph. Shipping it as an exclusion in the CHANGELOG is worse than shipping nothing, because it will be believed. Filing the silent-bump finding as the #2215 follow-up would be the higher-value outcome of this PR.

---

## 🟡 The `except Exception` claim is backwards

`python/djust/tests/test_recovery_version_staleness_1817.py:395-398`, repeated in the PR body and CHANGELOG:

> `asyncio.CancelledError` is how the communicator reports a receive timeout, and it derives from BaseException — NOT Exception — so a bare `except Exception` lets it through

**RAN**, direct probe on this interpreter (3.12.9) and asgiref:

```
recv: (TimeoutError, OSError, Exception, BaseException)   is_Exception=True
send: (asyncio.exceptions.CancelledError, BaseException)  is_Exception=False
```

The **receive** timeout raises `TimeoutError`, which *is* an `Exception` and *is* caught by `except Exception`. The `CancelledError` comes from the **subsequent send** — fact (i), not fact (ii). The two facts got merged.

This is harmless to the code (`except (asyncio.TimeoutError, asyncio.CancelledError)` catches both regardless) and entirely not harmless as documentation: it is presented as a hard-won harness fact in three places, and the sibling claim "the `_recv` helper has the same shape" then reads as "`_recv`'s `except Exception` is broken for timeouts" — which is wrong. `_recv` is fine for timeouts; it is vulnerable only to an *already-cancelled* task, via fact (i). Per #1867, a documented invariant needs the falsifying case run against it, not just the citation checked.

---

## 🟡 The flake window is relocated, not closed — and an identical assertion survives one file over, unmentioned

`test_global_isolation_1883.py:406` asserts `(v_mount, v_arm, v_jump) == (1, 2, 3)`, and `:424` asserts `v_jump == 4`. Same view, same mount → bump → jump path, same exact-integer shape this PR argues is fragile scaffolding — untouched by this PR and unmentioned in it.

**RAN.** Under my simulated #2215 producer (one stray bump after mount), `test_time_travel_jump_recovery_version_is_current` now passes (the PR's goal, achieved) — and `test_channel_layer_reset_keeps_wire_version_clean_1882` **and** `test_gate_off_without_reset_reproduces_1882_drift` both go red. So the producer can still flake CI at the same rate, from a file whose failure will read as "the #1882 channel-layer reset regressed."

This is the same #1646 parallel-path shape the repo canon warns about, applied to test assertions rather than production code.

I do **not** think those two should be relaxed — their exactness is load-bearing (they are a matched clean/gate-off pair proving the reset works). But the PR should say so. "The exact integers were never what the test guards" is true of the #1817 test and false of those two, and the difference is the interesting part. One paragraph.

Related, and the reason V2 above only half-matters: the double-bump regression the relaxation drops in this file *is* still caught — by `test_global_isolation_1883.py`. **RAN** with a source-pin-shape-preserving mutation (a standalone `self._next_version()` before the jump render, leaving `version=self._next_version_armed(html)` intact so `test_ws_send_version_1788`'s source-pin still passes): all 4 tests in this file GREEN, `test_global_isolation_1883.py` RED ×2. Net suite coverage is preserved — but only by the file the PR does not mention, and via a coincidence rather than a decision. Worth one sentence in the test comment so the next person relaxing this file knows what is actually holding the line.

*(Note for the record: `test_ws_send_version_1788.py::test_every_client_checked_send_path_uses_next_version` fired against my first V2 mutation, but it is an `inspect.getsource` **source-pin** — it caught the source shape, not the semantics. It does not cover the double-bump class.)*

---

## 🟡 The new test's frame check is blind in exactly the window the flake was observed in

`_receive_until` (`:118-125`) drains up to 8 frames looking for a wanted type and **discards** the rest. So an unsolicited frame arriving during the mount or event round-trip is consumed and thrown away.

**RAN.** Stray frame injected at the consumer's `group_add`, varying the delay:

| arrival | caught? |
|---|---|
| t = 0.00 s | 🟢 PASSES — swallowed |
| t = 0.01 s | 🟢 PASSES — swallowed |
| t = 0.05 s | 🟢 PASSES — swallowed |
| t = 0.10 s | 🟢 PASSES — swallowed |
| t = 1.3 s / 1.45 s | 🔴 caught (idle window) |
| t = 1.7 s / 3.0 s | 🟢 PASSES — after the window |

The docstring is honest about the trailing edge ("a producer that fires once an hour will not appear in a 1.5-second window") but says nothing about the leading one, which is worse: the mount→event round-trip is *the* #2215 window, and a frame arriving inside it is actively discarded.

The fix is nearly free, because `_recv` already appends every frame to `_FRAME_LOG`. After the round-trip, assert the recorded types are the expected set — the evidence is already collected and simply never asserted on:

```python
unexpected = [f for f in _FRAME_LOG
              if f.get("type") not in {"connect", "mount", "patch", "time_travel_event"}]
assert not unexpected, f"unsolicited frame during the round-trip: {unexpected!r}"
```

That would have caught every swallowed case above, and it closes the leading edge the 1.5 s wait cannot reach.

---

## 🟡 Two smaller ones

1. **No `disconnect()`.** Every other communicator in this file disconnects (`:250`); this one does not, so `_cleanup` / `group_discard("djust_hotreload", …)` never runs and the channel name is left in the layer group. In a PR about stray broadcasts polluting other tests, leaving a consumer in the broadcast group is an unfortunate shape even if the cancelled app task makes it inert in practice.

2. **`receive_nothing()` is the API this test wants.** `ApplicationCommunicator.receive_nothing(timeout, interval)` polls the output queue **without** cancelling the application task — it is the asgiref-supplied answer to "assert nothing arrived", it removes the whole ordering constraint the comment at `:377-385` exists to explain, and it makes a clean `disconnect()` possible afterwards. I used it in my own probes for exactly this reason after hitting fact (i). Recommend it; the `try/except (TimeoutError, CancelledError)` dance then goes away.

---

## What I tried that did **not** find a problem

- V1 and V3 jump-path mutations (arm-without-advance; arm-before-advance) — both still caught by the relaxed test.
- Gating off the #1817 fix and re-running the whole file 3× — deterministic single failure every time, always the final assertion.
- 30 consecutive isolated runs of the file — 30/30 clean.
- 5 clean `-n auto` runs across three test roots, 7409 tests each, random order — no flake in either the modified or the new test.
- Timing the new test — 1.52 s, real but bounded; no wall-clock assertion anywhere in it, so the #1795 timing-flake class does not apply (it asserts presence/absence of a frame, not a duration).
- Checking whether the new test could false-positive on a slow machine — the only timeout is the idle wait, and a *longer* wait can only make it stricter, never falsely red.
- `git config core.bare` verified `false` and `git status` clean at exit; the main checkout was never touched (#1804/#300).

---

## Summary

The relaxation itself is sound and I would not block on it: I independently reproduced the gate-off, constructed three neighbouring regressions (two still caught, one covered elsewhere), and proved both halves of the new test non-tautological by injection. The signal genuinely moved rather than disappearing.

What blocks is the **narrowing result**. It is stated as an exclusion, it is offered specifically to save the next investigator time, and it is wrong in the direction that costs the most: the hotreload path advances the wire version and emits no frame, which is the reported symptom exactly. That paragraph is in the commit message and the CHANGELOG, so it outlives this PR.

Requested before merge:

1. **🔴** Correct or remove the `djust_hotreload` narrowing result; file the silent-bump finding as the #2215 follow-up.
2. **🟡** Fix the `CancelledError`-on-receive claim in the test comment, PR body and CHANGELOG (it is `TimeoutError`; `CancelledError` is the *send*).
3. **🟡** Assert on `_FRAME_LOG` after the round-trip so the leading-edge blind spot closes.
4. **🟡** One sentence acknowledging `test_global_isolation_1883.py:406`/`:424` — same assertion, same path, deliberately not relaxed.
5. **🟡** `receive_nothing()` + `disconnect()`.

(2)–(5) are each a few lines. (1) is the one that matters.

REQUEST_CHANGES

johnrtipton:
## Stage 15 findings addressed

All three actioned. **N1 was the important one, and it was right in the way that matters** — the docstring stated an invariant that was false for the one producer this PR is about, in a PR whose thesis is that a confident false claim hid the bug for two years. Fixed at the mechanism rather than in the prose.

### N1 — solicited-ness now decided by `ref`, not by frame type

A hot-reload broadcast has no frame type of its own; `_send_update` emits it as `{"type": "patch", "hotreload": True}`. So the type allowlist admitted it, and `_receive_until` returned the broadcast *as* the event's patch, so the version assertion read the stray's version. Both halves green.

`_solicited_by(frame, ref)` replaces the allowlist: the client echoes its `ref` on an event's response (`time_travel_event` carries it under `entry`), and a broadcast carries none. Probed with a sending stray injected into the window:

```
window:            [('patch', ref=None, v=2), ('time_travel_event', ..., ), ('patch', ref=1, v=3)]
strays caught:     1  ['patch']          <- the broadcast
receive_until ->   ref=1 version=3       <- the event's own, not the stray
```

Taking the suggestion to match on `ref` — it closes both blind spots at once, as you said.

### N3 — removed the inner guard rather than commenting around it

Your framing decided it: keeping it means a future `hotreload=True` caller repeating the argument-evaluation mistake fails **silently** (that is #2215); removing it means such a caller sends a wasteful empty-patch frame — loud, version-consistent, and visible to the idle test. Trading a silent correctness bug for a visible performance one is the right direction, so there is now one mechanism, with the reasoning left at the site where the guard used to be.

**That removal then exposed a second instance of N1 that I would not have found otherwise.** With the stray now arriving as a real `patch` frame, `test_global_isolation_1883`'s own `_recv_until` mistook it for the arming patch — the chain read `(1,2,3)` and the flipped gate-off **stopped going red under a reverted fix**. A decorative pin (#1859), confirmed across three orderings. Same `ref` fix applied there. Worth noting the sequence: the N3 change did not cause that defect, it *revealed* it — the pin had been one frame-type coincidence away from useless.

### N2 — the CHANGELOG entry was a rewrite where it should have been a merge

Both dropped items restored: `test_an_idle_connection_receives_no_unsolicited_frames`, and the two relaxed scaffolding assertions in `test_time_travel_jump_recovery_version_is_current` with why they were incidental. Accurate catch — the release-facing record had no trace of a test-weakening.

### Nit / pre-existing `patch_count` — left alone per #1079.

## Gate-off after the fix-pass

Harness asserts the mutation text was found and that the source changed, clears `__pycache__` per iteration, and counts errors separately from failures:

| mutation | result |
|---|---|
| call-site guard → `if False:` | **3 failed** |
| predicate body → `return False` | **3 failed** |

Both turn `test_a_stale_layer_stray_no_longer_drifts_the_version_2215`, `test_a_suppressed_hotreload_broadcast_consumes_no_wire_version` **and** `test_hotreload_empty_patches_array_is_suppressed` red. Every mechanism in the fix now has a test that goes red when only that mechanism is removed.

Full suite: **10,373 passed**, three consecutive clean runs. Shipped `[1.1.0]` CHANGELOG section verified byte-identical to its tag.

johnrtipton:
## Stage 14 Retrospective — PR #2237 (#2215)

**Task**: the #1882/#2215 wire-version flake — sighted twice over two years, reproduced never.

### Quality: 3/5

The PR that finally fixed it is the PR that first shipped the wrong conclusion about it. That is the honest score.

### What went wrong, and it is the whole story

The first draft ran the right experiment and drew the inverted inference. It sent a `group_send` to `djust_hotreload`, observed **no frame at the socket**, and concluded the hot-reload path was innocent — then shipped that in the CHANGELOG as a *narrowing result for whoever picks this up next*. It was the producer. There was no frame **because** the version had been consumed and the broadcast then suppressed.

The generalizable error: **"I looked for X and found nothing" is evidence about X, not about the hypothesis** — and here absence of X *was* the hypothesis's signature. Every prior hunt across two years made the same move, which is exactly why the bug survived them. A silent failure mode is invisible to any search shaped like "find the artifact."

What would have caught it earlier: the two-arm probe the reviewer ran. Not "does a broadcast produce a frame?" but "does a broadcast change the version, compared with not sending one?" A control arm turns an absence into a measurement. I had the harness to do that and didn't.

### What worked

- **The reviewer's mandate to falsify, not to inspect.** Two rounds, and each found a defect that inspection would have approved. Round 1 falsified the narrowing claim; round 2 falsified a docstring invariant (`a stray hotreload is still caught` — false: a broadcast rides `type: "patch"`, so the type allowlist admitted the one producer the file is about). #1867 exists for exactly this and I still wrote a false "always" claim.
- **Fixing at the mechanism instead of the prose.** The tempting response to round 2 was to correct the sentence. Matching on the echoed `ref` instead of the frame type fixed the sentence *and* the blind spot.
- **Removing the redundant mechanism (#2233) rather than keeping it as a backstop** — and that removal is what surfaced the last defect: with the stray now arriving as a real `patch`, `test_global_isolation_1883`'s harness mistook it for the arming patch and the flipped gate-off **stopped going red under a reverted fix**. A decorative pin (#1859) that had been one frame-type coincidence from useless. The change did not cause it; it revealed it.

### The finding worth keeping

**#1882, #1883 and #2215 were one bug.** `test_gate_off_without_reset_reproduces_1882_drift`, written months ago to reproduce the drift, reproduced it *through this defect*: it required `jump == 4` and now reads 3. The channel-layer reset fixture was containment for a root cause nobody had found. Flipping that assertion rather than deleting it makes the transition itself the proof the cure is at the source.

This is the chain shape (#2142) seen from the other end — not one bug spawning follow-ups, but three separately-filed issues collapsing into one root cause once it was finally named.

### Verified

- Gate-off: reverting either the call-site guard or the predicate body turns **all three** covering tests red. Harness asserts the mutation matched, asserts the source changed, clears `__pycache__` per iteration, counts errors separately (#2129/#2135).
- Full suite **10,373 passed**, three consecutive clean runs (pollution-class gate, #182/#1174).
- Shipped `[1.1.0]` CHANGELOG section byte-identical to its tag (#2028). CI: 15/15 green.
- Follow-up #2238 filed: the #1817 structural pin counts prose over raw source — my docstring inflated its count. Deferred for scope (#1079).

`RETRO_COMPLETE`

Issue #2239, opened 2026-08-26
tech-debt: the Python Decimal converters are still lossy — normalize_django_value + __default__


johnrtipton:
Split out of #2214 to keep that PR scoped (#1079). The Rust half is fixed there;
this is the remaining half.

## State after #2214

| converter | Decimal becomes | precision |
|---|---|---|
| Rust `serialize_python_value` | `str` | exact ✅ |
| Rust `FromPyObject for Value` | `Value::Decimal` | exact ✅ |
| `python/djust/serialization.py` `__default__` | `float` | **lossy** |
| `python/djust/serialization.py` `normalize_django_value` | `float` | **lossy** |

Django's own `DjangoJSONEncoder` returns `str(o)`.

## Why the Python pair was not fixed with the rest

Three constraints that pull against each other, all verified:

1. **They are a pair.** `normalize_django_value` is documented as the fast path
   for djust's own encoder, and
   `tests/unit/test_normalize_django_value.py::TestParityWithJSONRoundtrip`
   pins them equal for every shared type. A first pass at #2214 changed only
   `__default__` to `str` — it looked like free parity on a JSON-only path —
   and split that invariant. The parity suite caught it. Either both move or
   neither does.

2. **`str` regresses templates.** `normalize_django_value`'s output is written
   into the template context (`python/djust/mixins/context.py:318`, for models
   and model lists), so the Rust renderer sees it, and a string is a string
   there: `{{ p|floatformat }}` stops rounding and `{% if p > 10 %}` compares
   lexically. Both measured.

3. **Returning the `Decimal` unchanged breaks the contract.** It is the right
   shape — the Rust extractor would build a `Value::Decimal` — but the function
   is documented as producing "JSON-safe primitives", and at least one consumer
   (`python/djust/runtime.py:2561`, `json.dumps(public_state, ...)`) passes no
   encoder and would raise `TypeError`.

## Exposure

Bounded but real. Model fields normally reach the client through the JIT path
(Rust `serialize_python_value`), which is now exact — verified:

```
RUST/JIT nested dict : {'price': '12345678901234567890.123456789'}  (str)
PY normalize fallback: {'price': 1.2345678901234567e+19}            (float)
```

This is the fallback for models the JIT path skips, plus any direct
`json.dumps(..., cls=DjangoJSONEncoder)` caller.

## The likely shape of the fix

Option 3 (return the `Decimal`) is the one that composes with `Value::Decimal`,
[...]

PR #2240, opened 2026-08-26
fix(core): carry Decimal as exact digits instead of a binary float (#2214)

### LINKED ISSUE #2214: bug: DecimalField reaches the client as a lossy float — the Decimal branch is dead code

johnrtipton:
Closes #2214.

> **This body was stale after the first review round and is now rewritten.** It previously claimed "both converters" (there are six), listed three gate-off mutations (there are 27), and repeated a sentence the review had already falsified. Flagged in re-review; recorded here because a stale PR body is the artifact a future reader trusts most.

## The bug

`serialize_python_value`'s `type_name == "Decimal"` branch was **dead code**. `extract::<f64>()` ran above it, and PyO3's `f64` extraction goes through `PyFloat_AsDouble`, which honours `Decimal.__float__` — so every `Decimal` became a binary double before the branch could see it.

| input | before | after |
|---|---|---|
| `Decimal('19.99')` | `19.99` *(float)* | `"19.99"` |
| `Decimal('12345678901234567890.123456789')` | `1.2345678901234567e+19` | *(exact)* |

`UUID` shared the branch and was never affected — not float-convertible, so it reached the check.

## Not the fix the issue proposed

Moving the branch above the numeric extracts **regresses two template behaviours**: the serialized value is written back into the template context, so the renderer sees what the wire sees, and as a plain string `{{ p|floatformat }}` stops rounding and `{% if p > 10 %}` compares lexically.

Instead a `Value::Decimal(String)` carrying the exact digits, with one `as_f64()` for arithmetic so the numeric consumption sites cannot drift (#1646). The contract is **exact rendering and transport, arithmetic unchanged** — a Decimal computes exactly as it did when it *was* a `Value::Float`.

## Six converters, not one

`FromPyObject for Value` (template context), `serialize_python_value` (JIT wire), `python_to_value` (**the actor path** — it carried the *"both must agree (#1646)"* comment while being one of the unfixed), `python_to_json_value`, `python_to_json`, and `model_serializer`'s, whose two exported serializers disagreed with each other. All now share one `is_decimal()` testing `isinstance` — so a subclass is claimed and a class merely *named* `Decimal` is not — resolved once per interpreter, since re-importing cost 18-24% on context conversion.

## Two rules taken from Django, not inferred

`{{ }}` goes through `numberformat.format()`, which is **not** `str()`:

1. `"{:f}".format(number)` — so `1E-9` renders `0.000000001`. `Decimal('1')/Decimal('1000000000')` is `1E-9`, so this is easy to reach, and it was a **regression against the previous release** where the value was a float. A zero coefficient never grows trailing zeros: `0E+3` is `0`, not `0000` — reachable from money arithmetic, since `Decimal('1000').quantize(Decimal('1E+2'))` minus itself is `Decimal('0E+2')`.
2. `abs(exponent) + len(digits) > 200` switches to `"{:e}"`, which Django added *"to avoid high memory usage"*. Without it a twelve-byte `Decimal('1E-10000000')` expanded to a **ten-megabyte** string.

`repr` keeps the exponent form, as Python does.

## Binary encodings carry a tag

`Value` was `#[serde(untagged)]`, so a `Decimal` encoded as a bare string and came back a `String` — and `SerializableViewState.state` round-trips through msgpack on **every** read of the default `InMemoryStateBackend`, so one cache hit undid the fix. Split on `is_human_readable()`: JSON keeps the bare string (the wire is unchanged); binary formats get a tag `visit_map` recognises.

Consequence: a `Value::Decimal` can hold an arbitrary string, which falsifies the "a Decimal is only digits, so no escaping needed" reasoning in `value_to_json` — true of the values, false of the type. Now escaped through the same helper as strings.

## Breaking change

A `Decimal` arrives at the browser as a JSON **string**, matching `DjangoJSONEncoder`. Client code doing arithmetic on it needs `Number()`. A JSON number cannot carry the precision.

## Deliberately not fixed

The Python `normalize_django_value` / `__default__` pair stays on `float` — **#2239**. Three verified blockers: they are a tested pair, `str` regresses templates, and returning the raw `Decimal` breaks their JSON-safety contract (`runtime.py:2561` dumps with no encoder). Its exposure is **not** small — `mixins/jit.py` reaches it at seven sites and a model with a `@property` sends the whole object that way. An earlier version of this body said otherwise; that was the most consequential error in the PR, because it made the deferral look cheaper than it is.

Three f64-precision limits are **pinned as stated limits**: `{% if p == 19.99 %}` (Django says False — exact Decimal/float comparison), two Decimals differing beyond f64, and `{% if p == 0 %}` below `f64::EPSILON` (the last is new in this PR and Decimal-only). The first two are pre-existing. Django itself raises `TypeError` on non-finite Decimals where djust renders them — also pinned.

## Verification

- **39-arm gate-off**, mutation asserted present and source asserted changed, errors counted separately, build breaks INVALID not green, restore in a `finally` **and** a dirty-tree assertion — the last two because an earlier harness was killed mid-iteration, leaked a mutation into a commit, and shipped a wrong-value bug (see below).
- **Randomized differential vs real Django**: 7,073 cases, 0 divergences **on the templates that harness covers** — a statement about its template set, not about Django parity in general. The reviewer's independent 21-template run measured 3,135 divergences fixed against 242 new, and it was that run, not mine, that surfaced the `{% if p == 0 %}` class.
- Full Python suite and `make test-rust` green; clippy 0.

## What the review rounds caught

Recorded because each was a real defect in my own work, and the pattern matters more than the individual fixes:

1. The msgpack round-trip undid the fix; my test asserted only the **encode** direction and stayed green. *A one-directional round-trip test cannot fail for the thing it guards.*
2. `{{ p }}` regressed for exponent forms, with a comment asserting the opposite — written into the same file where I had just recorded the rule that prose invariants must be falsification-tested (#1867).
3. A gate-off mutation leaked into a commit: every `Decimal` hashed identically, so `{% for %}` served row 1's fragment for every row — wrong prices in a price table. The guard test was green because it made two **separate** render calls, each with its own cache, and `render_template` never installs one. Rewritten in Rust at the level where the cache runs.
4. My first gate-off harness reported `rust=ok` for every mutation — its cargo invocation was aborting rather than running. *A harness that cannot go red is the tautology one level up.*

5. My "measured to be so: no input distinguishes this guard" was false — the measurement ran only the guard-ON arm and called it a comparison. Two guards and a `"0"` floor were all load-bearing; the floor is the only thing between `Decimal("0E-250")` and a panic.
6. A comment claimed the loop-cache tag separates Decimal from Float. I then "corrected" it to say they *cannot* collide because Float hashes a bit pattern and Decimal a `&str` — **also false**, and the fifth instance of the same error. `Hash for str` appends `0xff`, so any seven-byte string collides with the f64 of those bytes plus `0xff`. The tag separates Decimal from both; the test now uses a constructed witness.
7. The equality widening leaked past Decimal onto `(Float, Integer)`, so `{% if delta == 0 %}` answered true for `0.1 + 0.2 - 0.3`. Restricted, and the plain-float path pinned.
8. `i64` overflow on an attacker-chosen exponent panicked the render (and would wrap silently in release, where `overflow-checks` is off). Saturating now.


johnrtipton:
## All six findings addressed

Both blockers were real and both were mine. Reproduced each before fixing.

### 🔴 1 — msgpack round-trip

Confirmed at the Rust level: `Decimal("19.99")` → `String("19.99")`.

Split on **`is_human_readable()`**: JSON keeps the bare string (the wire is unchanged), binary formats carry a one-key tag that `visit_map` recognises. Near-misses pinned — wrong key, right key with a non-string payload, right key alongside another.

The test that should have caught this asserted only the encode direction. That is worth naming precisely: **a one-directional round-trip test cannot fail for the thing it guards**, which is the #2135 shape one level further out than either of us had it written down.

### 🔴 2 — exponent form

Confirmed against Django:

```
1E-9   django='0.000000001'   djust(before)='1E-9'
4E+1   django='40'            djust(before)='4E+1'
```

Django renders via `numberformat.format` → `"{:f}".format(...)`. Now expanded in `Display`; `repr` keeps the exponent form, as Python does. Verified by a **6,901-case randomized differential** against Django — 0 divergences, replacing the comment I'd reasoned my way into. The false comment was mine, in the file where I had just written the rule that a prose invariant must be falsification-tested (#1867).

That sweep also found a divergence neither of us had: Django **raises `TypeError`** on `Decimal('NaN')`/`Infinity` (it calls `abs()` on an already-stringified value) where djust renders them. Pinned as a stated difference.

### 🟡 3 — untested arms

All **22** now go red under gate-off, including `filesizeformat` and `dictsort`.

Two things worth reporting rather than quietly fixing. Three of my new tests were written from assumption and failed on first run — `filesizeformat` uses a plain space where Django uses U+00A0 (pre-existing, affects ints too), `pprint` output is HTML-escaped, and `dictsort` had to be driven directly because `{% for %}` over a filtered expression renders empty for every type. Corrected against measurement.

And my first gate-off harness reported `rust=ok` for **every** mutation — its cargo invocation was aborting with SIGABRT rather than running, because I passed the wrong interpreter. A harness that cannot go red is the tautology one level up; re-run with `scripts/embeddable-python.sh` it immediately showed three reds.

### 🟡 4 — the other converters

Six total, all now sharing one `is_decimal()`. `python_to_value` — the actor path — was the sharp one: it carried the *"both must agree (#1646)"* comment while being one of the unfixed. Your framing is the right one: before this PR all six were consistently wrong, which is a different kind of correct, and fixing two made the same value render differently by route.

Covered by a real `create_session_actor` → `mount` → `create_component` render (the props path), gate-off confirmed.

### 🟡 5 — "exposure is bounded" — falsified, and it changes #2239

You're right and it was the most consequential error in the PR, because it made the deferral look cheaper than it is. `mixins/jit.py` calls `normalize_django_value` at six sites; its fallbacks are ordinary, and a model with a `@property` in the template sends the whole object down that path.

The deferral stands — all three blockers verified independently — but it is now stated at its true size in the code comment, the CHANGELOG, and **#2239**, which I've re-labelled and corrected with the evidence.

### 🟡 6 — `is_decimal` cost

Cached in a `PyOnceLock` (pyo3 0.29's rename of `GILOnceCell`). Re-measured, same method:

| probe | before | now |
|---|---|---|
| render_str | 2567us | 2202us |
| render_flt | 2128us | 1817us |
| serial_flt | 1973us | 1629us |

At your short-circuited baseline (2172 / 1791 / 1597) — residual ~1-2%. The comment claiming "a dict lookup" is corrected; subclass and namesake behaviour now pinned.

### 🟢 7, 8

`get_digit` is fixed by the exponent expansion (same root cause). The `huge|floatformat` / `stringformat:'d'` limits are now pinned explicitly, since the differential only exercised `19.99` where the limit never shows.

## Verification

- **22/22 arms red** under gate-off, mutation text asserted present and source asserted changed, errors counted separately, build breaks INVALID not green.
- Full Python suite **10,414 passed / 0 failed**; `make test-rust` exit 0; clippy 0.
- Shipped `[1.1.0]` CHANGELOG section byte-identical to its tag.

johnrtipton:
## All findings addressed, including the one I caused

### 🔴 NEW-1 — the leaked mutation

You found a wrong-value bug I introduced, by the worst available route: my gate-off harness restored files at the **end** of each iteration rather than in a `finally`, so being killed mid-run left the mutation in the tree — and the next run then read the mutated file as its "original". I committed it without checking `git diff`.

Reverted. The harness now restores in `finally` **and** asserts the tree is clean before exiting, which is the check whose absence let this reach a commit at all.

### 🔴 NEW-2 — the tautological guard

Exactly right, and the diagnosis is the interesting part: two **separate** `render_template` calls meant each got its own cache, so the hash never had to distinguish anything — and `render_template` never installs a loop cache in the first place. Only the `RustLiveView` paths do, so a Python test through that entry point **structurally cannot** see this.

Rewritten in Rust where the cache actually runs (`crates/djust_templates/tests/test_decimal_loop_cache_2214.rs`), installing a real `LoopCacheGuard`. Gating the fix off now fails it immediately with the intended message.

### 🔴 2 — the exponent rules, properly this time

Both classes confirmed and fixed against `numberformat.py` rather than inferred:

- **Zero coefficient**: `0E+3` → `0`, not `0000`. Your reachability argument is the one that matters — `Decimal('1000').quantize(Decimal('1E+2'))` minus itself is `Decimal('0E+2')`, so a zero balance rendered `000`.
- **Django's `>200` cutoff**: `1E-10000000` now renders `1e-10000000` in 11 bytes instead of expanding to ten megabytes.

Plus one your table implied and I'd have missed: Python writes the exponent sign, `e+212` not `e212`. My own randomized differential caught it; reading the format spec did not.

**6,878 cases, 0 divergences**, boundary pinned at exactly 200/201.

### 🟡 3 — five green arms, not the "22 red" I claimed

Your broader enumeration was right and mine was self-selected. Now **27/27 red** — including the three you named plus two more I found: both fixes for finding 2, which I'd verified with an ad-hoc script rather than a test.

The last one to go red was `json_script escaping`, and it took a second attempt for a reason worth recording: my first test fed a plain Python string, which is the **String** arm. The hostile value can only reach the **Decimal** arm through the tag, so the test now round-trips a dict whose single key is the tag — which is your NEW-3 attack, used as its own regression test.

### 🟡 NEW-3 — escaping

Correct, and the shape of the error is the lesson: the reasoning was sound about the *values* a Decimal could hold and wrong about the *type* once the tag existed. Both arms now share one escaper, so the Decimal arm cannot be reasoned into an exemption again.

### 🟡 NEW-4, 5, 6

`decimal_tag()` is `#[doc(hidden)]`. The PR body is fully rewritten — you were right that a stale body is the artifact a future reader trusts most, so it now leads with what was stale about it. Perf claim corrected to your measured 2.4–3.7%.

Citation drift, mine again: `mixins/jit.py` has **seven** call sites, not six — a raw grep says eight because one is a comment. Neither number previously stated was right. Corrected in the code, CHANGELOG, and #2239.

## Verification

- **27/27 arms red**, mutation asserted present and source asserted changed, errors counted separately, build breaks INVALID.
- Python **10,424 passed / 0 failed**; `make test-rust` exit 0; clippy 0.
- Randomized differential: 6,878 cases, 0 divergences.

Two more harness failures on my side, both fixed and both worth naming since they produced results I briefly believed: I ran tests concurrently with the harness and read failures caused by its mutated rebuilds, and a heredoc inside a backgrounded call silently didn't write, so I twice read a stale script's output from a previous session as this run's result.

johnrtipton:
## All findings addressed — and your harness-validity note is the most useful thing in the review

### 🔴 the `as_tuple()` claim

Confirmed and fixed. `as_tuple().digits` drops leading zeros; the parse kept the `0` from `0.xxx` plus any zeros after the point, inflating the count so the cutoff fired early and the coefficient shifted when it did.

Your reachability case is the one that matters — `Decimal(1)/Decimal(7)` under `localcontext(prec=120)` is ordinary code. All four of your examples now match Django.

And the diagnosis of *why the suite missed it* is the durable part: all six boundary cases had `1` as their integer part, so not one exercised a `0.xxx` form. The docstring's "the boundary is pinned exactly" was true on the axis it enumerated and blind on the one it did not — enumerate-every-variant, failed on the axis I hadn't noticed existed.

My differential now covers it: **7,073 cases, 0 divergences**.

### 🟡 seven green arms

All resolved, and two of them took a second attempt:

- **The collision test could not fail, twice.** First version asserted `"1.5|1.5|1.5|"` — what a collision produces. My fix then used values whose *digit strings* differ (`1.50` vs `1.5000`), so the hash separated them on payload and the tag was still never load-bearing. The case that actually needs the tag is identical payload with different rendering: `Decimal("1E+1")` renders `10`, the string `"1E+1"` renders verbatim.
- **`leading-zero strip` was green** — the fix for your 🔴 had no test, because every boundary case still led with `1`.
- Subclass test now uses a value no float can hold; `{:+}` has a positive-exponent case; backslash escaping is pinned; `python_to_json` is driven through `serialize_queryset` → `serialize_object_with_paths`.

**`non-finite guard` is genuinely unreachable**, and I'm reporting that rather than manufacturing a test. Gating it off changes the output for *no* input — `NaN`, `Infinity`, `abc`, `1.2.3`, `a.b`, `""`, including fed through the tag — because the general path reconstructs a non-numeric string unchanged by coincidence of how it places the point. Kept as belt-and-braces, documented as such at the guard, and the test that looked like it covered that branch now says what it actually pins.

**33 arms, 33 resolved: 32 red, 1 documented-unreachable.**

### 🟡 falsified prose

All four, plus the pre-existing `hash_value` "map keys are sorted" — contradicted by its own `Object` arm since #2203, and sitting directly above the arm the previous round's regression hid in.

### 🟡 the third escaper — #2241

Filed with your object-key repro and the control-character range. Named in the comment so "one helper rather than the same chain per arm" does not read as complete when it isn't.

### Nits

Citation, third revision: seven call sites; `grep -n` returns **nine** lines (seven calls, one comment at 316, the import at 12). Stated that way now so a fourth revision isn't needed. CHANGELOG mid-flow sentence and the missing test-file entry both fixed.

## Your harness bug

> My first Python gate-off returned 18/18 GREEN including "`is_decimal` always false". That is definitionally impossible, so I stopped rather than reporting it.

That's the standard. I've now had three separate harness failures produce numbers I *did* report — `rust=ok` from a cargo invocation that was aborting, a mutation string `cargo fmt` had reflowed so an arm was never measured, and a heredoc inside a backgrounded call that silently didn't write, so I twice read a previous session's output as this run's. Every one produced a plausible number.

The `PYTHONPATH=$(pwd)` → `$(pwd)/python` correction is noted for future briefs; that one was my brief's error, not yours.

## Verification

- **33/33 arms resolved** (32 red, 1 measured-unreachable), harness restoring in `finally` with a dirty-tree assertion.
- Python **10,428 passed / 0 failed**; `make test-rust` exit 0; clippy 0.
- Differential: 7,073 cases, 0 divergences.
- CI green on the runner: **15/15** (Actions finally dispatched).

johnrtipton:
## All five addressed. The diagnosis of my error is the useful part

### 🟡 NEW-1 — "measured to be so" was false, and I can say exactly how

You're right, and the mechanism matters more than the fix: **my measurement ran only the guard-ON arm and called it a comparison.** No control. Of course nothing looked different — I never built the other side.

That is the same error as the very first false claim in this PR's history ("produces no frame at the socket"), which was also a single-arm observation reported as a conclusion. Four rounds apart, same shape.

And you're right that this one was the worst-shaped: a comment saying a live guard is unreachable tells the next maintainer it is safe to delete.

Both guards now pinned — `the_empty_and_punctuation_guard_is_load_bearing` (`""`, `.`, `-`, `+`, `E+5`, `e5`, …) and `the_non_digit_guard_is_load_bearing` (`abcE+5`, `xyzE+200`, …). I had not even noticed they were two separate guards.

### 🟡 NEW-2 — the panic

This is the one I'm most glad you found, and it surfaced because your harness returned **INVALID rather than a number**. `Decimal("0E-250")` is an ordinary value Django renders fine; without the `"0"` floor, `significant.split_at(1)` on an empty string is a hard render crash. It was absent from my arm enumeration *and* from the tests, and the floor's comment gave only the `as_tuple() -> (0,)` rationale — so a future simplification would have shipped it. Pinned across five values.

### 🟡 NEW-1b — and the claim was wrong in mechanism, not just coverage

Adding the Float row back did **not** turn `tag 9 -> 3` red, which sent me looking at why: `Float` hashes an 8-byte bit pattern where `Decimal` hashes a `&str`, so the two cannot collide **whatever tags they carry**. The tag separates Decimal from *String* — that arm does go red. So the comment asserted a mechanism that does not exist.

Kept the Float row as what it actually is, a payload guard, and the comment now says which guard is which.

### 🟡 NEW-3 — localization

Corrected. Your framing is the one I've adopted: a residual gap inside a strict improvement, not parity. Filed **#2242** with your `de` table and the `numberformat` scientific-branch fix shape.

### 🟡 NEW-4 — stale body

Refreshed, and the two new failure modes appended to its list.

### Tally correction

"32 red + 1 unreachable" was wrong: **33 red**. This round added six more arms; one reported INVALID (build broke) rather than a number and was re-measured — red, not green. **39 arms, all red.**

## Verification

- Python **10,428 passed / 0 failed**; `make test-rust` exit 0; clippy 0; CI 15/15.
- Your independent numbers — 23,913 differential cases and 28 arms with no divergence you could construct — are the strongest evidence the PR has, precisely because they aren't mine.

Four rounds, four falsified prose claims from me and zero from you. The code has been right since round three; my sentences about it have not been, and the pattern is specific enough to name: **I report the arm I ran as though it were the comparison.**

johnrtipton:
## 🔴 R5-1 — you're right, and it's the fifth instance one message after I named the pattern

Verified independently before fixing:

```
PROBE str="1.5E+30" f=-4.443727305504026e304
      h_str=6887536084304932878  h_f64=6887536084304932878
```

`Hash for str` writes the bytes then `0xff`; `Hash for u64` writes eight native-endian bytes. Any **seven-byte** string collides with the f64 whose NE bit pattern is those bytes plus `0xff`.

I generalised **one** non-colliding observation — `Float(10.0)` — into a universal, in the commit whose entire subject was fixing a comment that had done exactly that. And your severity call is right: it was the only survivor of your 29 arms, and the comment told the next maintainer that separator wasn't load-bearing.

Fixed with your constructed witness. All three separators now red:

```
tag 9 -> 3 (Float)    rust=1  RED
tag 9 -> 4 (String)   rust=1  RED
payload discarded     rust=3  RED
```

## 🟡 R5-2 — and it's new in this PR, which I hadn't registered

Boundary reproduced exactly:

| value | Django | djust |
|---|---|---|
| `2.2204460492503130E-16` | NZ | NZ |
| `2.2204460492503129E-16` | NZ | **Z** |
| `1E-16` | NZ | **Z** |

The part I'd missed is that `values_equal`'s wildcard was `_ => false` on `main`, so a Decimal never compared equal to an integer at all — this is genuinely new, not inherited semantics. Net-positive (`Decimal('0.00') == 0` is now right) but bounded, and now pinned exactly in `test_decimal_equality_with_an_integer_is_epsilon_bounded` rather than asserted away in a docstring.

## 🟡 R5-3 — tally

Fair. "39 arms, all red" and "gating tag 9 → 3 leaves this green" were both in the tree at once, and the tally hid the one that mattered. Corrected.

## Folded into #2242

`truncatechars` and `make_list|first` past the cutoff — same root cause as the localization gap, one fix covers all three.

## Verification

- Python **10,429 passed / 0 failed**; `make test-rust` exit 0; clippy 0; branch 0 behind main.
- Your differential against **both** builds is the number that matters: **9,337 divergences fixed, 0 new**.

## On the pattern

Five rounds, five falsified claims from me, zero from you. The mechanism is identical every time and I can now state it without hedging: **I run one arm and report it as the comparison.** "Produces no frame at the socket." "No input distinguishes this guard." "Cannot collide whatever tags they carry." Each was a single observation promoted to a universal, and each time the code was fine — it was the sentence that was wrong.

Worth noting what your harness did that mine didn't, three times running: it treated INVALID as a result to investigate rather than a zero. That is how the panic path surfaced, and how you caught your own `PYTHONPATH` isolation failure and the wrong-binary probe. I'd rather that be the durable takeaway than any of the individual fixes.

johnrtipton:
## Both blockers fixed. The first is the one that matters — it wasn't prose

### 🔴 R6-1 — a real regression I shipped, outside the issue's scope

Reproduced before fixing:

```
0.1+0.2-0.3   5.551115123125783e-17    django=NZ  djust=Z
1.0-0.9-0.1  -2.7755575615628914e-17   django=NZ  djust=Z
1e-17                                  django=NZ  djust=Z
5e-324                                 django=NZ  djust=Z
```

You're right about the mechanism and right about why it survived: **the arm's comment said "every pair involving a Decimal" while the code did more.** Two rounds of reviewers — including me writing the follow-up commit — read the comment and not the match.

Restricted via `is_decimal_pair`. Plain floats match Django again, every Decimal fix kept, and the float path is now pinned rather than only the Decimal one.

Your point that ordering needed no such guard checks out: `compare_values` already carried explicit `(Float, Integer)` / `(Integer, Float)` arms, so its wildcard only ever sees Decimal pairs.

**One consequence worth naming**: restoring `main` re-exposes a pre-existing divergence — `{% if x == 0 %}` is false for `0.0`, and `{% if x == 19 %}` false for `19.0`, where Django says true — because `values_equal` has no float/int arms at all. Your differential counted 3 of these as "FIXED" by the widening. I took the scoped option and filed **#2243** with the fix shape (explicit arms comparing *exactly*, not with an epsilon — which is what avoids re-creating the residue bug). Asserted in the test so the restoration is exact.

### 🔴 R6-2 — the overflow panic

Confirmed; the probe killed the process before it could print. Both sites saturate now.

Your reachability argument is the one I'd have had to make against myself: it cannot be attacker-reachable for the escaping fix and both guard tests and not for this. And the release note is the sharper half — `overflow-checks` is off, so the shipped wheel would have **wrapped silently and rendered nonsense** rather than crashing.

The framing I'll carry: the guard tests enumerated every hostile *shape* and no hostile *magnitude*. Your own 5,000-case fuzz missed it for the same reason mine would have — you found it by reading the arithmetic.

### 🟡 R6-3 / R6-4

PR body item 6 corrected — it was recording the round-five falsehood as an accepted finding, in the section about not doing that. The differential claim now says what it is: a statement about one harness's template set, not Django parity. Your 21-template run is the number that found the class mine missed, and the body says so.

"Two f64-precision limits" → three, in both the CHANGELOG and the body.

## Verification

- Three new arms, all red: `is_decimal_pair` guard, `saturating_sub`, `saturating_add`.
- Python **10,431 passed / 0 failed**; `make test-rust` exit 0; clippy 0; branch 0 behind main.

## Six rounds

Five prose falsehoods and now one real regression, all mine; zero from the reviews. What changed this round is that the failure finally had teeth — `{% if delta == 0 %}` on a float residue is a live wrong-branch bug, and it rode in on the commit fixing the *previous* round's finding.

The through-line across all six is one thing: **I generalise from the case I ran.** One arm read as a comparison, one non-collision as a universal, one comment's scope as the code's scope. Every time the code was doing something wider or narrower than the sentence I wrote about it, and every time a reviewer had to construct the case I hadn't.

johnrtipton:
## All three addressed. Two of them were tests that couldn't fail

### 🟡 1 — half a guard pinned

You're right and the evidence is damning: mutating `is_decimal_pair` to check only `a` survived **57 cases here, 881 across the Python roots, and all 56 Rust binaries**. Your grep explains it exactly — no test in the repo puts a literal on the left of `==`. And it isn't equivalent; it gives eight wrong answers against Django.

I wrote a two-sided guard and pinned one side. Fixed with reversed operands; gate-off red.

### 🟡 2 — the test was inert in the profile that gates the merge

This is the sharpest of the three. CI builds `--release`, `overflow-checks` is off, so unfixed code doesn't panic there — it **wraps**, and my `isinstance(..., str)` assertion sails through. I pinned the debug half and left unpinned the half **my own comment calls worse** ("wraps silently and renders nonsense").

Now asserts the rendered values. Verified with `wrapping_*` mutations, which reproduce release-profile overflow exactly rather than approximating it — both red. Your observation that the exponent's sign flips positive is what makes the assertion cheap.

### 🟡 3 — "correct either way"

Falsified, and the same shape as round six: the sentence asserted more than the reasoning under it established. Branch selection is correct; the value isn't (`…807`, not `…808`). Corrected to state the branch property, the value divergence, and why it's tolerable — unreachable from a real `Decimal`, since CPython's `MAX_EMAX` is far short of `i64::MAX`, so it matters only for tag-supplied strings where not crashing is the whole requirement.

### Nit

Fixed — my insert had split `numeric_pair`'s doc block from `numeric_pair`, so rustdoc attributed it to `is_decimal_pair`. Also tightened "its wildcard only ever sees Decimal pairs" to the actual argument you gave: `numeric_pair` admits only {Integer, Float, Decimal}², and all four non-Decimal combinations have their own arms.

## Verification

- Three new arms red, including `wrapping_*` for release-profile equivalence.
- Python **10,432 passed / 0 failed**; `make test-rust` exit 0; clippy 0; branch 0 behind main.

## Seven rounds

Your 253,000-case set-differential is the number that makes this PR trustworthy, and it isn't mine — that matters, because six rounds of my own measurements were each wrong in a way I couldn't see from inside them.

The pattern held to the end, just with diminishing consequence: rounds 1–3 behaviour, round 6 a real regression, round 7 two tests that couldn't fail. Every instance was the same error — **I generalise from the case I ran** — and the last one is almost self-parodying: a guard with two halves, pinned on one, in a PR whose recurring finding is exactly that.

johnrtipton:
## Landing all three 🟡 before merge

Two are the same shape as findings that produced real bugs in earlier rounds, so I'd rather not merge with them open.

**R8-1 — the dictsort half.** Restricting the fallback to Decimal pairs left 10,431 Python and 1,084 Rust green, so the int/float half was unpinned. The important distinction is yours: unlike round six's equality leak, this one is a **strict improvement** — 938/2184 → 2184/2184 against Django — so it stays unguarded. Pinned, and the comment now names both deltas instead of framing Decimal as the only one. CHANGELOG records it.

**R8-2 — the legacy Display half.** `the_legacy_display_path_is_not_lossy` used a value with no exponent form, so it pinned digits and not expansion. Gate-off red now.

**R8-3 — two comments narrower than their code.** "Only `Integer` and `Float`" at both variable-output sites, when `localize_if_number` has matched `Decimal` since #2214. Exactly the shape that let the equality widening survive two reviews.

**Nits:** "eight wrong answers" was unsourced — it's five, across ten assertion sites. Your note on the redundant-but-not-vacuous assertion is right and I've left it, since you demonstrated it fires; the comment calling it "the load-bearing half" was wrong and is gone.

## Final state

- Python **10,433 passed / 0 failed**; `make test-rust` exit 0; clippy 0; CI **15/15**; branch 0 behind main.
- Both new arms red.

## What the eight rounds actually produced

Three real bugs in the shipped code that the 10,000-case suite never saw: a silent wrong-value loop-cache collision, a `{% if delta == 0 %}` regression on plain floats, and a render-path panic. Plus five false claims of mine and a series of tests that couldn't fail.

Every one of my errors was the same: **I generalise from the case I ran.** One arm read as a comparison, one non-collision as a universal, one comment's scope taken for the code's, a two-sided guard pinned on one side. What caught them was never inspection — it was constructing the case I hadn't.

The three things your harnesses did that mine didn't, which I'd keep:
- **INVALID is a result to investigate, not a zero.** That's how the panic surfaced.
- **Assert the environment inside the harness.** `PYTHONPATH` silently resolving to the main checkout would have made every mutation read green — you caught it in yourself twice.
- **Run the profile that gates the merge.** A test can be load-bearing in debug and inert in release.

Merging.

johnrtipton:
## Stage 14 Retrospective — PR #2240 (#2214)

**Task**: `Decimal` reached the client and the template engine as a lossy binary float — the `type_name == "Decimal"` branch was dead code beneath `extract::<f64>()`.

### Quality: 3/5

The shipped fix is right and heavily verified. Getting there took **eight adversarial review rounds**, and seven of them found that my previous fix commit had introduced the next defect.

### What the reviews caught that the suite didn't

Three real bugs, none visible to 10,000 passing tests:

1. **A silent wrong-value loop-cache collision.** A gate-off mutation leaked into a commit because my harness restored files at the *end* of each iteration rather than in a `finally` — so a kill mid-run left it, and the next run read the mutated file as its baseline. Every `Decimal` hashed alike; `{% for p in products %}{{ p.price }}{% endfor %}` served row 1's price for every row.
2. **A `{% if delta == 0 %}` regression on plain floats.** My equality widening was written for Decimals and also caught `(Float, Integer)`. `0.1 + 0.2 - 0.3` compared equal to zero. It survived two rounds because the arm's comment said "every pair involving a Decimal" while the code did more.
3. **A render-path panic** on `i64` overflow, reachable through the binary tag — the route this PR itself establishes as attacker-chosen. In release, where `overflow-checks` is off, it would have wrapped silently instead.

### The pattern, stated once

Five falsified prose claims, and every one the same error: **I generalise from the case I ran.**

- "produces no frame at the socket" — one arm, no control
- "no input distinguishes this guard" — one arm, no control
- "cannot collide whatever tags they carry" — one non-collision, promoted to a universal
- "every pair involving a Decimal" — the comment's scope taken for the code's
- a two-sided guard, pinned on one side

Each time the code was doing something wider or narrower than my sentence about it, and each time a reviewer had to construct the case I hadn't.

### Harness failures worth recording

Four separate ways my measurements produced numbers I shouldn't have trusted, all of which looked plausible:

- a `cargo` invocation aborting with SIGABRT, reported as `rust=ok`
- a mutation string `cargo fmt` had reflowed, so an arm was never measured
- a heredoc inside a backgrounded call that silently didn't write — twice reading a *previous session's* output as this run's
- running tests concurrently with a rebuilding harness, then believing the failures

### What worked

- **Gate-off as the primary instrument.** Nearly every finding came from mutating a mechanism and watching what stayed green.
- **Worktree-isolated adversarial reviewers with escalating briefs**, each told the previous fix probably broke something. That framing earned its keep seven times.
- **Differentials against real Django rather than curated tables.** Django is importable; there was never a reason to guess. The final independent run: 20,782 divergences fixed, 3,948 new, all in one documented class.

### Follow-ups filed

**#2239** (Python converters still lossy — reachable from `mixins/jit.py`'s seven fallback sites), **#2241** (third JSON escaper in the object-key path), **#2242** (scientific-form coefficient not localized), **#2243** (pre-existing `{% if <float> == <int> %}` divergence).

### Verified

10,433 Python passed, `make test-rust` exit 0, clippy 0, CI 15/15, ~40 gate-off arms all red across both profiles, shipped `[1.1.0]` CHANGELOG section byte-identical to its tag.

`RETRO_COMPLETE`

Issue #2257, opened 2026-08-27
tech-debt: escape/safe are no-ops where Django's are stringfilters, and floatformat cannot parse a numeric string


johnrtipton:
Split out of #2250 (#1079). #2250 gives Django's `@stringfilter` built-ins the
string Django gives them — `str(Decimal)`, not the `numberformat` rendering
`Display` produces. It covers **27 of Django's 29** stringfilters. `escape` and
`safe` are deliberately excluded, and this issue is the measurement of why.

## Residue 1 — djust's `escape` / `safe` are no-ops, Django's return a string

`crates/djust_templates/src/filters.rs`:

```rust
"escape" => Ok(value.clone()), // No-op: auto-escaping at render time handles this
"safe" => Ok(value.clone()),   // No-op: renderer checks for |safe to skip auto-escaping
```

Django's are `@stringfilter`, so they return `str(value)`. The consequence is
not about escaping — it is that the value stays a `Value::Decimal` (or
`Integer`/`Float`) all the way to the render site, where `localize_if_number`
fires. Django's had already become a `str`, which `localize()` leaves alone.

Measured against a live Django (`{{ p|safe }}`, `{{ p|escape }}`, 5 locales x 2
grouping flags):

| value | Django | djust |
|---|---|---|
| `Decimal('1E-9')` | `1E-9` | `0.000000001` |
| `Decimal('6.25E-2244')` under `de` | `6.25E-2244` | `6,25e-2244` |
| `1234567` (int) under `en-us`, grouping on | `1234567` | `1,234,567` |

**840 cells** for the `Decimal` half; the `int`/`float` half is additional and
was not the subject of #2250.

## Residue 2 — `floatformat` cannot parse a numeric string, which blocks residue 1

Coercing `escape`/`safe` to return a `Value::String` is not free: it changes the
type flowing down the rest of the chain, and `floatformat`'s dispatch has arms
for `Float`/`Integer`/`Decimal` and returns everything else unchanged. So
`{{ d|escape|floatformat }}` went from `0.0` (agreeing with Django) to `1E-9`.
Measured: **1,168 cells** regressed. That is why #2250 stopped at 27.

The gap is independent of `Decimal` and reproduces on `main` with a plain
string:

```
{{ "1E+1"|upper|floatformat }}      django '10'          djust '1E+1'
{{ "6.44E+199"|upper|floatformat }} django '644000…000'  djust '6.44E+199'
```

[...]

PR #2263, opened 2026-08-27
fix(templates): floatformat and add on exact decimal digits (#2253)

### LINKED ISSUE #2253: bug: floatformat and add parse a Value::Decimal through f64, losing the digits the variant carries
Found by the #2239 template differential, which measured the full
7-idiom × 4-value matrix against real Django rather than sampling it.

## The divergence

With the exact `Decimal` now reaching the renderer (#2214 for the wire, #2239
for the Python converters), `{{ p }}` and every comparison agree with Django.
Four cells do not, and all four are the same root cause — the filter parses
`Value::Decimal`'s digit string through `f64` before formatting:

| template | value | Django | djust |
|---|---|---|---|
| `{{ p\|floatformat }}` | `12345678901234567890.123456789` | `12345678901234567890.1` | `12345678901234567168.0` |
| `{{ p\|floatformat:2 }}` | `12345678901234567890.123456789` | `12345678901234567890.12` | `12345678901234567168.00` |
| `{{ p\|add:1 }}` | `12345678901234567890.123456789` | `12345678901234567891` | `12345678901234567890.123456789` (unchanged) |
| `{{ p\|floatformat }}` | `0.00` | `0` | `0.0` |

The first three are the f64 parse. The fourth is separate and smaller:
Django's `floatformat` with no argument drops the decimal part when the value
is integral after rounding; djust keeps `.0`.

`add` is the odd one — it does not lose the digits, it silently does nothing,
returning the input unchanged rather than the sum.

## Not a regression

Measured against the pre-#2239 context value (a `float`), which is equally
wrong on all four:
`python/tests/test_decimal_converters_2239.py::TestTemplatePathMatchesRealDjango::test_this_change_turns_no_agreeing_case_into_a_disagreeing_one`
asserts no cell that agreed before disagrees now, and passes.

## Where it is pinned

`KNOWN_FILTER_DIVERGENCES` in that same file, asserted as an exact **set**
(#1125) by `test_the_known_divergences_are_exactly_those_four` — so the list
cannot grow silently, and fixing any of them turns that test red, which is the
signal to prune it.

## Likely fix

`Value::Decimal(String)` already carries the exact digits and
`expand_decimal_exponent` already renders them Django's way
(`crates/djust_core/src/lib.rs`). `floatformat` and `add` should operate on the
digit string / a decimal type rather than `as_f64()`. Django itself special-cases
`Decimal` in `floatformat` for exactly this reason.

Related: #2242 (a scientific-form Decimal's coefficient is not localized) is the
same family — a filter that does not know its input is a `Value::Decimal`.

johnrtipton:
Closes #2253.

## The decision, first

**The #2214 contract stands and this does not overturn it.** That contract is
*"exact rendering and transport, with `as_f64()` for arithmetic"*. `as_f64()`
remains what `{% if p > 10 %}`, `divisibleby`, `filesizeformat` and
`stringformat` go through. What moved is `floatformat` — because **Django's
`floatformat` is not arithmetic on a float, it is formatting via `Decimal`**:

```python
d = Decimal(str(text))                      # every input: float, int, str, Decimal
rounded_d = d.quantize(exp, ROUND_HALF_UP, Context(prec=prec))
```

Formatting is the half #2214 already puts on the exact side. So the answer to
"should the contract change" is *no*, and the answer to "was routing
`floatformat` through `as_f64` right" is also *no* — those were never the same
question.

**No dependency.** `rust_decimal` / `bigdecimal` were not needed and are not
added. Django's algorithm is a quantize-with-carry on a digit string plus its
own 200-digit cut-off; the port is ~200 lines in
`crates/djust_templates/src/floatformat.rs`. Binary size delta: **0 bytes of
new dependency**.

## Two premises in the issue were wrong

Six of the last twelve PRs in this drain corrected a premise stated in their own
issue. This is two more.

### 1. `floatformat` was wrong on far more than `Decimal`

The issue names one cause — the `f64` parse — for two `floatformat` cells. A
differential over **21 argument forms × 475 values** measured **302 divergent
cells of 825**, from **four independent causes**, three of which have nothing to
do with `Decimal`:

| # | cause | example | Django | djust (before) |
|---|---|---|---|---|
| 1 | Rust rounds the *double* half-to-**even**; Django quantizes the *digits* half-**up** | `{{ 2.675\|floatformat:2 }}` | `2.68` | `2.67` |
| 2 | Django's default arg is `-1`, and a negative arg means "at most" — `"-3".parse::<usize>()` fails, so **every** negative arg became one place | `{{ p\|floatformat:"-3" }}` on `34.23234` | `34.232` | `34.2` |
| 2 | …same cause, the issue's fourth cell | `{{ p\|floatformat }}` on `Decimal('0.00')` | `0` | `0.0` |
| 3 | the `g` suffix was stripped from the arg and then ignored | `{{ 6666.6666\|floatformat:"2g" }}` | `6,666.67` | `6666.67` |
| 4 | precision — the cause the issue named | `{{ p\|floatformat:2 }}` on the 29-digit value | `…7890.12` | `…7168.00` |

Cause 2 is also why "fix only the two cited cells" was not available: the cited
`Decimal('0.00')` cell **is** the `p <= 0` branch, and once you have exact
quantization plus that branch you have most of the algorithm. Writing extra code
to preserve the other 298 wrong answers is not scope discipline.

Also fixed as a consequence, because Django's algorithm coerces them: a string
(`{{ "34.23234"|floatformat }}` → `34.2`, was `34.23234`), a bool, and the
`""`-for-garbage give-up (`{{ "abc"|floatformat }}` → `""`, was `abc`).

### 2. `add`'s cited cell is not the `f64` parse

`int(Decimal('12345678901234567890.123456789'))` is `12345678901234567890`,
which **does not fit an `i64` however exactly you compute it**. `checked_add`
overflowed, fell through to the fail-soft, and the filter returned its input
**unchanged** — the issue observed this ("it silently does nothing") without
connecting it to the width.

The `f64` parse *is* a second, real defect the issue did not name:
`{{ p|add:1 }}` on `Decimal('9007199254740993')` (2^53+1) gave back
`9007199254740993`. Widening the truncation alone would not have closed the
reported cell; widening the width alone would not have closed this one. Both are
fixed — `int()` truncates the exact digits into an `i128`, and a sum outside
`i64` is carried as `Value::Decimal`'s exact digits rather than discarded.

## Differential results

Rendered through `normalize_django_value` → `_rust.render_template` and compared
against a live `django.template.Template`, with the locale number format pushed
the way the real render path pushes it.

**`floatformat`, curated grid (21 arg forms × 475 values):**

| | before | after |
|---|---|---|
| divergent cells | 302 / 825 | **0 / 825** |

Per-argument-form after the fix — `floatformat`, `floatformat:"0"`, `:"1"`,
`:"2"`, `:"3"`, `:"5"`, `:"-1"`, `:"-2"`, `:"-3"`, `:"u"`, `:"2u"`, `:"g"`,
`:"2g"`, `:"-2g"`, `:"2gu"`, `:"2ug"`, `:"x"`, and the unquoted `0`/`2`/`-2` —
are each **0 / 475**.

**Randomized sweeps** (in the test file, not just this description): 3,000
`floatformat` cells and 2,000 `add` cells over sign × magnitude × scale ×
argument form. 0 mismatches. The curated table samples the axis you thought of;
this is the pairing the v1.1.1-2 retro asks for.

## Non-regression, as a set comparison against a `main` build

Not a spot check. `main` (at `865f72b3`) and this branch were each built and run
over the **same 23,750 cells** — 50 templates × 475 values, 400 of them
randomized — and the agreeing sets compared:

```
cells:            23750
agreed on main:   12421
agreed on branch: 19862
REGRESSED (agreed -> disagreed):  0
FIXED     (disagreed -> agreed):  7441
```

404 cells are still divergent but with a *different* djust output. Every one is
an improvement or neutral: `inf`/`NaN` through `add` used to fabricate
`9223372036854775807` and now return the value unchanged (Django raises, so
neither agrees, but one of them is not a made-up number); `1E+250|add:1`
likewise; `floatformat:""` is a Django `IndexError`; and
`floatformat:2|add:1` is the pre-existing `float_ok` divergence with a corrected
`floatformat` feeding it.

## Sibling filters that share the parse — all checked

| filter | verdict |
|---|---|
| `stringformat` | **checked, not fixed.** `stringformat:"d"` saturates a >i64 Decimal to `i64::MAX` (same width class as `add`). Left alone: it is not in the issue, and its bigger divergence is that Django returns `""` for a non-numeric where djust returns `0`. Filed. |
| `filesizeformat` | **checked, not fixed.** Diverges on **every** value — djust emits a plain space where Django emits U+00A0 — so its `Decimal`/`f64` cell is not the reachable defect. Filed. |
| `dictsort` | **checked.** Diverges 475/475 on a scalar (`[Object]` vs Django's error text); unrelated to numeric parsing, pre-existing. |
| `divisibleby`, `yesno`, `length`, `first`, `make_list` | **checked**, all pre-existing and untouched — confirmed by the 0-regression set comparison above. |
| `{{ p }}` (bare) | **checked.** Three pre-existing divergences remain (`NaN` vs `nan`, `1e+300`'s exponent form, non-finite Decimals). `floatformat` builds its own `str(text)` and agrees with Django on all of them; widening that to `Display` changes every render of a float and belongs to its own change. |

## Gate-off (#1468) — 9 mechanisms, 9 mutations

Harness asserts the mutation text was **found** and that the source **changed**,
counts pytest `N error` alongside `N failed`, and reports `INVALID` rather than a
number when the build breaks (#2129/#2135).

| mechanism | result | first failing test |
|---|---|---|
| `floatformat`: ROUND_HALF_UP → truncate | 31 failed | `test_the_fallback_model_renders_as_django_renders_it` |
| `floatformat`: drop the `p <= 0` integral branch | 16 failed | `test_scalar_renders_as_django_renders_it[0.00-…]` |
| `floatformat`: default argument `-1` → `1` | 8 failed | `test_scalar_renders_as_django_renders_it[0.00-…]` |
| `floatformat`: route a Decimal back through f64 | 30 failed | `test_scalar_renders_as_django_renders_it[…123456789-…]` |
| `floatformat`: ignore the `g` suffix | 5 failed | `test_a_randomized_sweep_agrees` |
| `floatformat`: keep the sign on a rounded-to-zero value | 16 failed | `test_a_randomized_sweep_agrees` |
| `floatformat`: drop the 200-digit cut-off | 20 failed | `test_every_argument_form_agrees_across_every_value[…]` |
| `add`: Decimal truncation back through f64 | 11 failed | `test_scalar_renders_as_django_renders_it[…123456789-…]` |
| `add`: discard a sum outside i64 | 10 failed | `test_scalar_renders_as_django_renders_it[…123456789-…]` |

No mutation was a no-op, a build break, or a `0 failed`.

One defect was found by this discipline rather than by inspection: the first
pass parsed the **argument** before the **value**, and Django parses the value
first. Their give-up paths return different things (`""` vs the input back), so
`{{ "abc"|floatformat:"x" }}` rendered `abc` where Django renders nothing —
4 cells, caught by the differential, now pinned by
`an_unusable_value_beats_an_unusable_argument`.

## What is deliberately still divergent

Stated here and asserted in `TestKnownRemainingDivergences`, so closing one turns
the file red rather than going unnoticed:

- **Past `i128`, `add` gives up.** Python's ints are unbounded; `i128` is not.
  `Decimal('1E+250')|add:1` returns its input. Matching Django needs
  arbitrary-precision arithmetic, which this does not add.
- **`add`'s third branch** still returns the value where Django returns `""`.
  Pre-existing and deliberate (documented in the filter since #2203).
- **`{{ p|floatformat:"" }}`** raises `IndexError` in Django 5.2 — `arg[-1]` on
  an empty string, a Django bug. djust treats it as an absent argument.
- **A Python `int` wider than `i64`** is already lossy crossing into
  `Value::Integer`; `{{ p }}` alone shows it. Not a filter gap.
- **`Value::Float`'s `Display`** writes `NaN` where Python writes `nan`.

## Verification

- Full suite across **every** test root (`tests/ python/tests/ python/djust/tests/`): **10,773 passed**, 219 skipped.
- `make test-rust`: **57** test binaries green.
- `cargo clippy --all-targets --all-features`: **0** warnings. `cargo fmt --check`: clean.
- Every row added to `docs/RUST_TEMPLATE_API.md` was **executed** against a live Django before commit (#1516), including the two that were already there.

## Existing tests corrected in place

Three pinned the old behaviour. Each is corrected with the reason recorded rather
than deleted:

- `test_decimal_precision_2214.py` — the f64 limit narrows to `stringformat:'d'`;
  `floatformat`/`add` now agree and are **asserted** to, so a regression to the
  f64 path reddens.
- `test_scientific_localization_2242.py` — its `1E-250` row expected `0,00`,
  which was djust's old `{:.N}`-on-a-double output and **never Django's answer**
  (`1E-250`). A parity half is added so the table cannot drift from Django again.
- `test_filter_django_parity_2203.rs` — the overflow give-up *point* moved to
  i128; the no-wrapped-number invariant the test exists for is unchanged and
  still asserted.
- `KNOWN_FILTER_DIVERGENCES` in `test_decimal_converters_2239.py` is now **empty**
  and still asserted as a **set**, so a regression re-populates it and names the
  cell.

## Files

| file | why |
|---|---|
| `crates/djust_core/src/decimal.rs` | **new.** `DecimalParts` + `parse_decimal_parts` (lifted out of `expand_decimal_exponent`, so the renderer and both filters share one definition — #1646), `to_i128_trunc`, `python_float_repr`. |
| `crates/djust_templates/src/floatformat.rs` | **new.** Django's algorithm, with a module docstring naming what it does and does not cover. |
| `crates/djust_core/src/lib.rs` | `expand_decimal_exponent` now uses the shared parse. |
| `crates/djust_core/src/locale.rs` | `localize_number_forced` for the `g` suffix; `localize_number` delegates to it so the two cannot drift. |
| `crates/djust_templates/src/filters.rs` | `floatformat` arm → the new module; `add`'s `as_int` → exact i128. |
| `python/tests/test_floatformat_parity_2253.py` | **new**, 49 collected. |


johnrtipton:
Third commit (`148de61a`) — one more finding, from falsification-testing this
PR's own doc claim rather than from review.

The new module docstring asserts that `floatformat`'s `u`/`gu` suffixes emit
Django's **default** `DECIMAL_SEPARATOR`/`THOUSAND_SEPARATOR`/`NUMBER_GROUPING`
rather than a project's overrides, because only the localized format reaches
Rust. Verifying the file:line a claim cites is necessary and not sufficient
(#1867) — the claim itself has to be run. It was, and it holds:

| template | Django (`DECIMAL_SEPARATOR="!"`, `THOUSAND_SEPARATOR="_"`, `NUMBER_GROUPING=3`) | djust |
|---|---|---|
| `{{ p\|floatformat:"2u" }}` | `6666!67` | `6666.67` |
| `{{ p\|floatformat:"2gu" }}` | `6_666!67` | `6666.67` |
| `{{ p\|floatformat:"2" }}` | `6666.67` | `6666.67` ✓ |
| `{{ p\|floatformat:"2g" }}` | `6,666.67` | `6,666.67` ✓ |

The two localized rows are asserted alongside the two divergent ones, which is
what **bounds** the gap to `u` — without them the entry reads as a general
locale failure, which it is not.

Now pinned by `test_the_u_suffix_ignores_overridden_number_settings` (restoring
the Rust thread-local in a `finally`; full suite re-run to confirm no leak —
10,774 passed), tracked at **#2266** with the fix shape written down, and named
in the CHANGELOG's still-divergent list.

---

Also verified after opening, since the curated grid does not carry them:

**The `MAX_PLACES` boundary and its DoS shapes.**

| argument | Django | djust |
|---|---|---|
| `floatformat:"200"` | 202 chars | 202 chars, byte-identical |
| `floatformat:"1000"` | 1,002 | identical |
| `floatformat:"100000"` | 100,002 | identical, 8 ms |
| `floatformat:"1000000"` | 1,000,002 | identical, 82 ms |
| `floatformat:"1000001"` | 1,000,003 | input returned (the documented cap) |
| `floatformat:"1000000000"` | `InvalidOperation` **raised** | input returned |

A tag-supplied `Value::Decimal` holding `1E+400000000`, `1E-400000000` or
`1.5E-9223372036854775807` returns in **under 0.1 ms** through both filters —
the digit ceilings in `over_django_digit_cutoff` and `to_i128_trunc` are checked
*before* anything allocates from the exponent, so twelve bytes cannot ask for
400 MB.

---

**CI on the previous head (`f372725b`) was fully green**, including the
`test-summary` aggregate gate and `demo-checks`, `benchmarks`, `browser-smoke`,
`nav-hooks-guard`, `playwright-tests`, `python-tests (py3.12)` and
`python-tests (py3.14t free-threaded)`.

Follow-ups filed for everything checked and deliberately left alone (#1079):
**#2264** (`filesizeformat` — diverges on *every* value for two reasons before
its f64 parse is reachable), **#2265** (`stringformat:"d"` — this exact defect
one filter over), **#2266** (the `u` gap above). Already open and confirmed
still-open: **#2258** (`Value::Float`'s `Display`), **#2260** (a Python `int`
past `i64` at the `Value` boundary). And **#2257's residue 2** —
"`floatformat` cannot parse a numeric string" — is *closed* by this PR;
commented there, with a note that the 1,168 cells its residue 1 was blocked on
are worth re-measuring now.

johnrtipton:
Fourth commit (`f9c3ab8f`) — self-review of my own diff, applying the two rules
this repo keeps needing. Both findings are in code I wrote in this PR.

**#1197 — a citation I added was wrong.** The comment on the refactored
`expand_decimal_exponent` pointed its pinning tests at
`crates/djust/tests/test_decimal_value_2214.rs`. That path does not exist; the
file is in `crates/djust_core/`. Found by mechanically extracting **every** file
path the diff cites and testing each for existence, rather than re-reading the
comment — 14 of 15 resolved, and inspection would not have distinguished the
one that did not.

**#1646 — I shipped two copies of one rule.** `over_django_digit_cutoff`
computed *"leading zeros dropped, floored at 1"* inside `significant_len()`, and
`expand_decimal_exponent`'s scientific branch computed the same thing again
inline, because it needs the string rather than the length. Two definitions of
what decides where the >200-digit cut-off falls and what the scientific
coefficient is — which is precisely the drift that made the cut-off fire early
for every `0.xxx` value until #2240. Collapsed to one `significant() -> &str`,
with `significant_len()` delegating.

Behaviour-identical, and **asserted rather than assumed**: the 23,750-cell set
comparison against the `main` build reproduces byte-for-byte —

```
REGRESSED (agreed -> disagreed):  0
FIXED     (disagreed -> agreed):  7441
still wrong but differently:      404
```

— the same three numbers as before the refactor. Plus `make test-rust` 57
binaries green, full Python suite 10,774 passed, `cargo clippy --all-targets
--all-features` 0 warnings, `cargo fmt --check` clean.

CI was fully green on `148de61a` (all 17 checks, including the `test-summary`
aggregate gate) before this commit.

johnrtipton:
Rebased onto `main` at `17dfbbaa`, which brought in **#2267** (#2250 — string
filters see `str(Decimal)`). That PR touches `filters.rs` and, more to the
point, wrote a characterization test for a blocker this PR removes. Force-pushed
`2d8668f5`.

### #2267's `test_escape_and_safe_are_the_named_exclusions` went red — as designed

Its own docstring says: *"so whoever closes #2257 has to come here and update it
deliberately."* Its last assertion pinned

```python
assert _render_both("{{ p|upper|floatformat }}", "1E+1") == ("10", "1E+1")
```

— Django coerces the string, djust could not. #2253's port begins
`Decimal(str(text))` on **every** input type, so both now say `10`. Updated to
the agreeing pair, still asserting **both** sides so it stays a differential
rather than becoming a djust snapshot, with the docstring recording that
**#2257's residue 2 is closed** and that the exclusion now rests on residue 1
alone (the `escape`/`safe` no-op — pinned by the two assertions above it, which
still pass unchanged).

Worth flagging for whoever picks up #2257: its **1,168-cell** cost for coercing
`escape`/`safe` was measured *while this blocker was open*. Those cells were
regressing because `floatformat` returned a string unchanged. It no longer does,
so that number should be re-taken before it is used to decide anything.

### The rebase duplicated this PR's CHANGELOG entry

Commit 2 added the paragraph; commit 3 edited it. The 3-way merge inserted the
edited version as a **new** line instead of replacing, leaving both the pre- and
post-edit copies under `### Fixed` — same family as #2028, since git has no
notion that two similar bullets are one bullet. Stale copy deleted; verified the
survivor is the superset (it is the one carrying the `u`-gap divergence and the
`(50 collected)` count), and checked every other rebased hunk for the same shape
— all singular.

### Re-measured against the new base

A stale-base measurement is a different program (#1450), so the whole
differential was redone. **The earlier run had mis-measured**: `git stash` does
not move HEAD, so the capture labelled "main" was actually the branch — both
sides reported the identical 19,899 agreeing, which is what gave it away. Redone
properly with `git checkout origin/main -- crates/`:

```
cells:            23750
agreed on main:   12458        (was 12421 before #2267 landed)
agreed on branch: 19899
REGRESSED (agreed -> disagreed):  0
FIXED     (disagreed -> agreed):  7441
```

Same 0 / 7441 against the new base. The CHANGELOG's baseline number is updated
from 12,421 to 12,458 so it names the base a reviewer would actually reproduce
against.

**Gate-off re-run on the rebased base** — all 9 mechanisms still redden a named
test, none `INVALID`, none `0 failed`. Two counts moved (31→32 and 5→6) because
#2267 added a test file that also exercises `floatformat`.

Full suite **10,783 passed**; `make test-rust` 57 binaries green; clippy 0
warnings; `cargo fmt --check` clean.

PR #2272, opened 2026-08-27
fix(templates): port Truncator, slugify, title and urlencode from Django (#2262, #2261)

### LINKED ISSUE #2261: bug: slugify maps '.' to '-' and title strips surrounding whitespace, both diverging from Django
Found while measuring #2250 (a Django differential over every built-in filter).
Neither is about `Decimal` — both reproduce on plain strings, so they are the
filter's own algorithm diverging, not an input-conversion problem.

## `slugify` turns `.` into `-` where Django deletes it

Django's `slugify` deletes everything outside `[\w\s-]` and only then collapses
runs of hyphens/whitespace, so a dot vanishes rather than becoming a separator.
It also strips leading/trailing `-_`.

```python
from djust import _rust
_rust.render_template('{{ p|slugify }}', {"p": "3.5"})       # djust '3-5'    Django '35'
_rust.render_template('{{ p|slugify }}', {"p": "<b>x</b>"})  # djust 'b-x-b'  Django 'bxb'
_rust.render_template('{{ p|slugify }}', {"p": "-1.5e+300"}) # djust '1-5e-300'  Django '15e300'
```

The `+` case shows both halves at once: djust maps `+` to `-` as well.

Reference: `django/utils/text.py::slugify` —
`re.sub(r"[^\w\s-]", "", value.lower())` then `re.sub(r"[-\s]+", "-", …).strip("-_")`.

## `title` strips surrounding whitespace and misses a word boundary after a non-letter

```python
_rust.render_template('{{ p|title }}', {"p": "  spaced  "})
# djust 'Spaced'            Django '  Spaced  '
_rust.render_template('{{ p|title }}', {"p": "<b>x</b>"})
# djust '&lt;b&gt;x&lt;/b&gt;'   Django '&lt;B&gt;X&lt;/B&gt;'
```

Django's `title` uppercases the first letter of every run of letters regardless
of what precedes it (`re.sub(r"([a-z])'([A-Z])", …)` over `str.title()`), and
never touches leading/trailing whitespace.

## Scale

40 mismatching cells over the non-`Decimal` control set alone (`slugify` 3
values, `title` 3 values, x locales x grouping) in the #2250 sweep; for
`Decimal` inputs `slugify` is 130 cells because most `Decimal`s carry a dot.

Filed together because they share a shape and a fix strategy — port the
reference regexes rather than reimplement the intent — not because they are one
bug.

### LINKED ISSUE #2262: bug: five filter-algorithm divergences from Django — truncatechars_html (x2), truncatewords_html, truncatewords, urlencode
Found while measuring #2250 (a Django differential over every built-in filter).
Four more filters whose own algorithm diverges from Django's. Every case below
reproduces on a plain string or a plain dict on `main` — no `Decimal` involved —
so none of them is #2250.

Filed as one issue because they were found by one sweep and each is a small
port-the-reference fix; they are four bugs, not one.

## 1. `truncatechars_html` truncates at `len == limit`, Django at `len > limit`

```python
from djust import _rust
_rust.render_template("{{ p|truncatechars_html:8 }}", {"p": "Infinity"})
# djust 'Infinit…'   Django 'Infinity'   (8 characters, the limit — not over it)
```

## 2. `truncatechars_html` counts ESCAPED characters, Django counts text

```python
_rust.render_template("{{ p|truncatechars_html:8 }}", {"p": {"a": 1}})
# djust '{&#x27;a&#x27;: 1…'   Django '{&#x27;a&#x27;: 1}'
```

The text is `{'a': 1}` — 8 characters. djust measures the 18-character escaped
form, so it truncates a string Django leaves whole.

## 3. `truncatewords_html` double-escapes

```python
_rust.render_template("{{ p|truncatewords_html:2 }}", {"p": {"a": 1}})
# djust '{&#x27;a&#x27;: 1}'   Django '{&amp;#x27;a&amp;#x27;: 1}'
```

Django's `truncatewords_html` is `is_safe=True` and escapes its input, so the
already-escaped `&#x27;` becomes `&amp;#x27;`. djust escapes once.

(Django's output is arguably the surprising one here, but parity is the
contract; if djust wants to diverge it should be a documented divergence like
the `add` third-branch note in `filters.rs`, not an accident.)

## 4. `truncatewords` does not strip surrounding whitespace

```python
_rust.render_template("{{ p|truncatewords:2 }}", {"p": "  spaced  "})
# djust '  spaced  '   Django 'spaced'
```

Django's `Truncator.words` joins on single spaces, which drops the padding.
Same shape as the `title` half of #2261.

## 5. `urlencode` encodes `/`, Django leaves it alone by default

```python
_rust.render_template("{{ p|urlencode }}", {"p": "<b>x</b>"})
# djust '%3Cb%3Ex%3C%2Fb%3E'   Django '%3Cb%3Ex%3C/b%3E'
```

`urlencode`'s documented default is `safe="/"`; the argument form
(`urlencode:""`) is what removes it. Anything rendering a path through
`|urlencode` gets `%2F` where Django gives `/`.

johnrtipton:
Seven filter algorithms diverged from Django. Each had been written against
Django's **documentation** rather than differentialed against its **behaviour**,
so each was correct-looking and wrong in a detail the docs do not mention.

Ported the references into a new `crates/djust_templates/src/truncate.rs`
rather than adjusting the existing code — three of the seven cells are not
reachable by adjustment.

## Issue × divergence × test

| issue | divergence | covering test |
|---|---|---|
| #2262/1 | `truncatechars_html` cut at `len == limit` | `TestReportedCells[Infinity]` |
| #2262/2 | `truncatechars_html` counted the ESCAPED form | `TestReportedCells[dict]` |
| #2262/3 | `truncatewords_html` escaped once, Django twice | `TestReportedCells` |
| #2262/4 | `truncatewords` kept surrounding whitespace | `TestReportedCells` |
| #2262/5 | `urlencode` encoded `/` | `TestReportedCells`, `TestUnreportedDivergences` |
| #2261/a | `slugify` mapped `.` and `+` to `-` | `TestReportedCells` ×3 |
| #2261/b | `title` stripped whitespace, missed boundaries | `TestReportedCells` ×2, `TestTitleExhaustive` |

## Two premises the issues got wrong

**`truncatechars_html`'s two cells are one branch.** Django's
`TruncateCharsHTMLParser.process` emits its input **raw and unescaped** and
stops when the whole input is one run of text of exactly `length` characters.
That single branch produces both reported cells, and no off-by-one adjustment
reaches it. This is the argument for porting rather than patching.

**`title` was wrong on far more than whitespace.** It split on whitespace,
where Python's `str.title()` treats *any non-cased character* as a boundary —
that is the `<b>x</b>` cell. Beyond the issue it also needed the real titlecase
mapping (`ß` → `Ss`, not `SS`), `Nd`-only `\d` in Django's `\d([A-Z])` fixup,
the real `Cased` set, and a `Case_Ignorable`-skipping final-sigma lookahead.

## Nine divergences neither issue names

Found by reproducing the ones they do: comments/doctypes/PIs are **deleted**
rather than escaped; an unterminated construct at the end of the input
discards everything after it (`"trailing &amp"` renders empty); `<script>` /
`<style>` switch to CDATA mode; character references round-trip through
`html.unescape`; `frame` and `spacer` are void elements; `length <= 0` is the
empty string; a negative filter argument parses at all; `urlencode`'s argument
was ignored entirely; and the four `title` items above.

`markup5ever` becomes a direct dependency but is **not a new crate** —
`html5ever` already pulls it into this workspace for `djust_vdom`, so
`Cargo.lock` gains one edge and no package. Its `NAMED_ENTITIES` is
byte-identical to CPython's `html.entities.html5` (2231 entries, asserted).

## Verification

| gate | result |
|---|---|
| randomized differential vs real Django, per filter, chains included | 0 divergences outside the pinned classes |
| both HTML truncators over generated markup | **0** divergences |
| exhaustive single-codepoint differential for `title` | 4,448,256 cells; only residue is Unicode-data-version skew (CPython 15.0 vs Rust std) |
| non-regression set comparison vs a real `origin/main` build | 24,300 cells: **1,070 unchained fixed, 0 unchained regressed** |
| gate-off matrix | 15 mutations, every one reddens a **named** test |
| full suite / `make test-rust` / clippy `-D warnings` | green |

## Three classes deliberately left open, with measurements

1. **`slugify`'s NFKD ASCII fold** and **`Truncator.chars`'s NFC + combining
   skip** need Unicode normalization tables this workspace does not carry. The
   residue is confined to non-ASCII input: 45 and 30 cells of a 13,500-cell
   unchained sweep. Pinned in `TestKnownRemainingDivergences`.

2. **The `escape|` and `safe|` chains regressed 243 cells — none of them the
   port.** Those cells had agreed for the wrong reason, two bugs cancelling:
   djust's `escape` returns its input unchanged (#2257 residue 1), and its safe
   rule is a name whitelist where Django's is `is_safe=True` **and the input was
   already safe**. A truncator that escapes its text the way Django's does now
   escapes twice. The proof is executable and in the suite: feed the port
   exactly what Django hands the filter and it reproduces Django's answer for
   **243 of 243**.

   Breadth of the safe-rule half, measured against a real `origin/main` build:
   **24 of Django's 27 `is_safe=True` filters** diverge on `{{ p|safe|X }}`
   there — `lower`, `capfirst`, `wordwrap`, `urlize`, … none of them touched
   here — versus 23 on this branch; without the `|safe` it is 5 on main and 1
   here. So this PR strictly improves both axes, and the remainder is
   framework-wide and worth its own issue.

## Also

`test_string_filter_stringification_2250.py`'s `UNCOMPARABLE` set is now
**empty** — its characterization test existed to go red when these closed, and
did. Two pre-existing tests that had encoded the pre-fix `urlencode` bytes were
updated deliberately.

Closes #2262
Closes #2261


johnrtipton:
## Side findings filed

The chain-axis sweep in this PR surfaced two divergences that are **not** in
scope here and are now tracked separately, both reproducing on `origin/main`
independent of this branch:

* **#2273** — `striptags` deletes everything after a lone `<` (`"a < b"` →
  `"a "`), and drops an entity's `;` when a tag is present. This is why
  `striptags|` is excluded from this PR's `_CHAINS`.
* **#2274** — `|safe` does not survive an `is_safe=True` filter: **24 of
  Django's 27** diverge on `{{ p|safe|X }}` on main (23 here — this PR fixes
  `slugify`). This is the cause of 4 of the 243 chain regressions measured
  below, and the reason `safe|` is excluded from `_CHAINS`. `TestSafeMarking.
  test_safe_does_not_survive_these_filters_and_that_is_framework_wide` pins it
  and says to delete the pin when #2274 closes.

The remaining 239 chain regressions are #2257 residue 1 (`escape` is a no-op as
a chain link), pinned in `TestEscapeChainIsBlockedOn2257`. All 243 are proven
to be the chain link rather than the port by an executable check in the suite:
feed the port exactly what Django hands the filter and it reproduces Django's
answer, 243 of 243.

Issue #2281, opened 2026-08-27
bug: `{{ p|escape|X }}` passes the UNescaped value to X


johnrtipton:
Found while measuring the chain axis of #2273. Independent of `striptags`.

`{{ p|escape }}` alone is correct. `{{ p|escape|<anything> }}` is not: djust
appears to defer the escape to render time rather than applying it as a value
transformation, so the *next* filter in the chain sees the raw value where
Django's sees the escaped one.

```python
from djust import _rust
_rust.render_template('{{ p|escape|upper }}', {"p": "a < b"})
# djust 'A &lt; B'          Django 'A &amp;LT; B'
_rust.render_template('{{ p|escape|striptags }}', {"p": "a<b>c</b>d"})
# djust 'acd'               Django 'a&lt;b&gt;c&lt;/b&gt;d'
```

Django's `escape` returns a `SafeString` containing `a &lt; b`; `upper`
uppercases *that* to `A &LT; B`, which then renders escaped again as
`A &amp;LT; B`. The `striptags` cell is the same mechanism and shows why it
matters beyond cosmetics: in Django the tags are already inert text by the time
`striptags` runs, so nothing is stripped. In djust they are still live markup.

## Scale

2,947 of 4,000 values in the #2273 sweep diverge on
`{{ p|escape|striptags }}`.

## Scope

This is a property of how `escape` composes, so every `escape|X` chain is
affected — grep the sink (wherever `escape` is special-cased in
`renderer.rs`'s deferred-escaping path) rather than enumerating the `X` you
expect. Worth checking `escapejs`, `force_escape` and the
`safe_output_filters` list at the same time: `force_escape` exists in Django
precisely because it escapes *immediately* where `escape` is lazy, and if
djust's `escape` is already eager-at-render the two may have collapsed into
one behaviour.

Pinned as a known divergence at
`python/tests/test_striptags_parity_2273.py::TestKnownRemainingDivergences::test_chain_divergence_is_not_striptags`,
which also asserts `striptags` itself is byte-exact on the same value. Delete
that row when this is fixed.

Issue #2280, opened 2026-08-27
bug: a filter's `is_safe=True` is not honoured — `{{ p|safe|title }}` re-escapes


johnrtipton:
Found while measuring the chain axis of #2273. Independent of `striptags` — it
reproduces on nearly every filter, and it is an **autoescape-correctness** bug,
not a cosmetic one.

Django's filter registry carries `is_safe`. When a filter with `is_safe=True`
receives a `SafeString`, its output stays safe. djust drops the flag, so the
value is escaped at render time even though the template said `|safe`.

```python
from djust import _rust
_rust.render_template('{{ p|safe|title }}', {"p": "a < b"})
# djust 'A &lt; B'      Django 'A < B'
```

## Survey — `{{ p|safe|<filter> }}` on `"a < b"`

| filter | Django `is_safe` | Django | djust |
|---|---|---|---|
| `striptags` | True | `a < b` | `a &lt; b` |
| `title` | True | `A < B` | `A &lt; B` |
| `capfirst` | True | `A < b` | `A &lt; b` |
| `slice:":3"` | True | `a <` | `a &lt;` |
| `center:9` | True | `  a < b  ` | `  a &lt; b  ` |
| `lower` | True | `a < b` | `a &lt; b` |
| `truncatechars:20` | True | `a < b` | `a &lt; b` |
| `wordwrap:20` | True | `a < b` | `a &lt; b` |
| `linebreaksbr` | True | `a < b` | `a &lt; b` |
| `default:"x"` | False | `a < b` | `a &lt; b` |
| `upper` | False | `A &lt; B` | `A &lt; B` ✅ |
| `join:""` | True | `a &lt; b` | `a &lt; b` ✅ |

`upper` agrees because `is_safe=False` means Django escapes too — agreement
there is the *absence* of the flag, not its propagation. `join` agrees because
it escapes its own items.

## Scale

3,130 of 4,000 values in the #2273 sweep diverge on `{{ p|safe|striptags }}`
alone. The striptags output itself is byte-identical to Django in every one of
them — only the flag is lost.

## Notes on the fix

- `join` is the counterexample to a blanket "propagate `is_safe`": it already
  escapes internally, so Django's `is_safe=True` on it does not mean the output
  is unescaped input. Mirror Django's table per filter rather than inferring it.
- `default` shows the flag is not the whole story: `is_safe=False` there, but
  Django returns the original `SafeString` *object*, so it renders unescaped.
- Pinned as a known divergence at
[...]

Issue #2279, opened 2026-08-27
bug: `{{ p|length }}` counts BYTES, not characters


johnrtipton:
Found while measuring the chain axis of #2273 (PR pending). Independent of `striptags` — it reproduces on a bare `{{ p|length }}`.

```python
from djust import _rust
_rust.render_template("{{ p|length }}", {"p": "中<b"})
# djust '5'    Django '3'
_rust.render_template("{{ p|length }}", {"p": "éb&nbsp;&amp; >&#"})
# djust '18'   Django '17'
```

`é` is 2 bytes, `中` is 3. Any non-ASCII string gets the wrong length.

## Why it surfaced now

It was masked in the `striptags` sweep: the old `striptags` deleted everything
after a `<`, so `"中<b"` became `"中"` (3 bytes) and coincidentally matched
Django's char count of the *unstripped* value. Fixing `striptags` unmasked it in
216 of 36,000 cells — two bugs that had been cancelling (the #2272 pattern).

Pinned as a known divergence at
`python/tests/test_striptags_parity_2273.py::TestKnownRemainingDivergences::test_chain_divergence_is_not_striptags`,
which also asserts `striptags` itself is byte-exact on the same value. Delete
that row when this is fixed.

## Scope

Worth checking whether the same byte-vs-char confusion affects `slice`,
`ljust`/`rjust`/`center`, `truncatechars` and `wordwrap` — grep the sink
(`.len()` on a `&str` where a character count is meant) rather than
enumerating the filters you expect.

PR #2282, opened 2026-08-27
fix(templates): port striptags from Django's MLStripper (#2273)

### LINKED ISSUE #2273: bug: striptags deletes everything after a lone '<', and drops an entity's ';' when a tag is present
Found while measuring the chain axis of #2262 / #2261 (PR #2272). Two separate
divergences in `striptags`, both reproducing on plain strings with no `Decimal`
and no template inheritance involved.

## 1. Everything after a lone `<` is deleted (data loss)

Django's `strip_tags` runs an `html.parser.HTMLParser` (`MLStripper`), which
emits a bare `<` as data. djust discards from the `<` to the end of the input.

```python
from djust import _rust
_rust.render_template('{{ p|striptags }}', {"p": "a < b"})
# djust 'a '                       Django 'a &lt; b'
_rust.render_template('{{ p|striptags }}', {"p": "price < 5 and > 3"})
# djust 'price  3'                 Django 'price &lt; 5 and &gt; 3'
_rust.render_template('{{ p|striptags }}', {"p": "x<y<z"})
# djust 'x'                        Django 'x&lt;y&lt;z'
_rust.render_template('{{ p|striptags }}', {"p": "<bx&y3.5word"})
# djust ''                         Django '&lt;bx&amp;y3.5word'
```

This is the severe half: any user-supplied prose containing a mathematical or
comparison `<` is silently truncated. Well-formed markup is fine
(`'a<b>c</b>d'` agrees), so it only bites on input that was never markup.

## 2. A named reference loses its `;` when a tag is present

`HTMLParser` with `convert_charrefs=True` resolves `&one` to itself and
`strip_tags` re-emits it with the trailing `;` Django's handler adds.

```python
_rust.render_template('{{ p|striptags }}', {"p": "&one two<b>x</b>"})
# djust '&amp;one twox'            Django '&amp;one; twox'
_rust.render_template('{{ p|striptags }}', {"p": "<i>&a b</i>"})
# djust '&amp;a b'                 Django '&amp;a; b'
```

Note the tag is load-bearing: without one (`"&one two"`) both sides agree, so a
sweep over plain strings alone will not find this.

## Scale

198 of 3,000 plain generated values and 14 of 1,500 generated-markup values in
the #2272 sweep. `striptags` is currently in the compared set of
`python/tests/test_string_filter_stringification_2250.py` — it agrees there
because that file's values are `Decimal`s, which contain no `<`.

Reference: `django/utils/html.py::strip_tags` / `MLStripper`. The fix shape is
probably the same as #2262's: drive a real HTMLParser port rather than scanning
for `<`…`>` pairs. `crates/djust_templates/src/truncate.rs` (added in #2272)
now contains a faithful `html.parser.HTMLParser` port that `strip_tags` could
reuse — `Parser::goahead` plus a `handle_data`-only sink.

johnrtipton:
Closes #2273.

`strip_tags` was a 12-line scanner: `<` set `in_tag`, `>` cleared it. So
`"a < b"` rendered `"a "` — every character from a mathematical or comparison
`<` to the end of the input was silently dropped — and `"a > b"` rendered
`"a  b"`, which the issue does not name.

## Why this is a port and not a patch

Django's `strip_tags` is an `html.parser.HTMLParser` (`MLStripper`) driven in a
**loop** until the tag count stops falling. Neither half is reachable by fixing
the scanner:

- A `<` not followed by a letter / `/` / `!` / `?` is **data** to the
  tokenizer. That is a fact about tag-open syntax the scanner has no
  representation for.
- One pass over `"<<b>script>"` yields a live `<script>` tag. Only the loop
  removes it.

Measured, not assumed: **wrapping Django's loop around the old scanner leaves
1,915 of 4,013 cases still divergent.**

## Reuse: one tokenizer, two sinks

#2272 already ported CPython's `goahead` for the HTML truncators. Writing it
twice would be exactly the parallel-path drift this codebase keeps paying for
(#1646), so the state machine is lifted behind a `Sink` trait into a new
`crates/djust_templates/src/htmlparser.rs`:

| | `TruncateSink` | `MLStripper` |
|---|---|---|
| `convert_charrefs` | true | **false** |
| driver | `feed()` | `feed_and_close()` |
| `handle_starttag` | re-emit verbatim, push the tag stack | drop |
| `handle_data` | spend the budget, escape | keep |
| `handle_entityref` | unreachable | re-emit as `&name;` |

`truncate.rs` keeps the truncation budget and the tag stack and loses 780
lines. The truncators' behaviour is unchanged — their own parity suite plus
`the_same_tokenizer_drives_the_truncating_sink` pin it.

## The DoS guard

Django raises `SuspiciousOperation` on a 1000+ character unclosed tag carrying
50+ `<`, and on an input still shedding tags after 50 passes. Both guards are
ported and both bounds are exact (49 inner opens — the match's own `<` counts;
50 passes allowed, 51 refused).

A djust filter has no channel to a 400: the Rust filter layer returns a
`String`, `SuspiciousOperation` is a Django *request* concept, and raising
would 500 the whole render rather than refuse one value. So the work really is
bounded and a refused value renders as the **empty string** with a
`tracing::warn!`. Empty is the only refusal safe in every context this output
reaches — returning the input, or the partially-stripped value, emits
attacker-controlled markup Django declined to emit, which under
`{{ v|striptags|safe }}` is an XSS surface Django has not got.

## Two port defects the parity table missed

Both were found by the randomized differential, and both are exactly the class
the v1.1.1-2 retro says a curated table blinds you to:

1. **`feed()` + `close()` is genuinely two `goahead` passes.** The `&#`-bail
   arm is the only `break` that advances before stopping, so the *next* pass
   resumes the loop — and a second bail inside pass 2 has no pass 3. A
   single-pass port gets `"&#;<b>x</b>"` → `"&#;x"` right and
   `"&#xZZ</b>&#;<br />a <b"` (which keeps its `<br />`) wrong, or the reverse.
2. **`entityref` backtracking is load-bearing.** Its name class
   `[-.a-zA-Z0-9]` overlaps its own trailing `[^a-zA-Z0-9]` on `-` and `.`, so
   a greedy scan eats the character the trail needed: `&amp-` is the entity
   `amp` with `-` as the trail, and `x&one3.5` is `x&one3;.5`.

## Verification

**Randomized differential** against real Django — 4,000 adversarial values ×
9 chains, rendered through both template engines:

| | cells |
|---|---|
| base (`origin/main`, rebuilt) | **27,015 DIFF** |
| after | **6,816 DIFF** |
| DIFF → AGREE | 20,219 |
| AGREE → DIFF | 20 |

The two counts differ, which is the check that the baseline is a real separate
build rather than the same one measured twice.

**Every residual cell is classified; none unexplained:**

| n | class |
|---|---|
| 27 | depth guard — Django raises `SuspiciousOperation`, djust renders `""` (documented above) |
| 496 | `<![name[` — CPython's `parse_marked_section` **asserts** on an unknown keyword; djust fails soft (pre-existing choice) |
| 2,947 | `\|escape\|striptags` — djust strips the UNescaped value → **#2281** |
| 3,130 | `\|safe\|striptags` — djust drops the safe flag → **#2280** |
| 216 | `\|striptags\|length` — djust's `length` counts BYTES → **#2279** |

**The 20 AGREE → DIFF cells are all `|striptags|length`**, and each is proven
mechanically to be #2279 unmasked, not a striptags defect: the striptags output
matches Django byte-for-byte in all 20, and `{{ p|length }}` *alone* already
disagrees on the same value. They were agreeing before because the old
striptags deleted the non-ASCII tail, so the byte count coincidentally matched
Django's char count — two bugs cancelling, the #2272 pattern.

**Gate-off (#1468), 8 mutations, each naming a red test:**

| mutation | result |
|---|---|
| M1 a lone `<` is a tag start again (the old scanner) | 4 RED |
| M2 wrapper loop → one pass | 3 RED |
| M3 drop the long-open-tag pre-scan | 1 RED (`..._refuses_a_long_unclosed_tag_carrying_fifty_opens`) |
| M4 never refuse at `MAX_STRIP_TAGS_DEPTH` | 1 RED (`..._refuses_an_input_still_shedding_tags_after_the_depth_cap`) |
| M5 `handle_entityref` without the `;` | 3 RED |
| M6 collapse `feed_and_close` into one `goahead(1)` | 1 RED (`..._resumes_parsing_after_a_bailed_out_numeric_reference`) |
| M7 `entityref` greedy, no backtracking | 1 RED (`entityref_backtracks_when_the_greedy_name_eats_its_own_trail`) |
| M8 never flush the incomplete tail | 3 RED |

The harness asserts each mutation was found and changed the source, treats a
build break or a missing `test result:` line as INVALID rather than green, and
verifies the source is restored byte-identical at the end.

**M4 SURVIVED its first form** — the input was `"<"*60 + "b>"*60`, whose
*uncapped* fixpoint is also `""`, so removing the cap changed the source and
computed the identical answer. A surviving mutation is a question, not a pass
(v1.1.1-2 retro): the answer was "this input cannot tell the two apart". The
values now carry `keepA`/`keepB` so the uncapped fixpoint is `"keepAkeepB"` and
the `""` is evidence of the cap rather than of stripping.

**Suites** (post-rebase on `origin/main`):

- `make test-rust` — **1193 passed, 0 failed**
- `pytest tests/ python/tests/ python/djust/tests/` — **12,435 passed, 219
  skipped, 0 failed**

## Follow-ups filed

- #2279 — `{{ p|length }}` counts bytes, not characters
- #2280 — a filter's `is_safe=True` is not honoured (survey of 12 filters)
- #2281 — `{{ p|escape|X }}` passes X the unescaped value

Each is pinned in
`TestKnownRemainingDivergences::test_chain_divergence_is_not_striptags`, which
asserts the cell still diverges *and* that `striptags` alone is byte-exact on
the same value — so it fails loudly when the follow-up lands and the row should
be deleted.

## Files

| file | |
|---|---|
| `crates/djust_templates/src/htmlparser.rs` | new — the tokenizer, the `Sink` trait, `MLStripper`, `strip_tags`, 13 tests |
| `crates/djust_templates/src/truncate.rs` | −780 lines; `Parser` → `TruncateSink : Sink` |
| `crates/djust_templates/src/filters.rs` | the scanner → `crate::htmlparser::strip_tags` |
| `python/tests/test_striptags_parity_2273.py` | new — 99 cases incl. the 27,000-cell differential |


johnrtipton:
## CI red — fixed, and the diagnosis was only half of it

Thanks — the diagnosis was right and it led somewhere bigger than the three cells. Two things were wrong: the port aimed at a superseded CPython, **and** the test computed its reference at run time.

### The reference disagrees with itself

`html/parser.py` was rewritten for HTML5-spec alignment in **3.12.10** and changed *again* in **3.14**. Measured across the corpus:

| | values differing (of 4,000) |
|---|---|
| 3.12.9 vs 3.12.13 | **1,108** |
| 3.12.13 vs 3.13.7 | 0 |
| 3.12.13 vs 3.14.6 | **224** (end-of-input `&` / `&#`) |

So it is not 3 cells, it is 1,108 — and **no fixed port can be green against all three CI interpreters**, because two of them disagree with each other. Porting "the hardened behaviour" would have fixed 3.12/3.13 and left 3.14 red.

The truncators share this tokenizer and are exposed too (344 values of `goahead(0)` trace drift), but their `feed()`-only path is identical on 3.12.13/3.13.7/3.14.6 — the 3.14 delta is confined to the `end=True` charref arms that only `striptags` reaches.

### Ported forward — all of it

Target is **3.12.10+/3.13**: every currently-shipping CPython, and the safer direction, since an unterminated attacker-controlled construct is now discarded rather than re-emitted as page text.

I tried the narrow fix first and it was **wrong**. Porting only the end-of-input half left the tokenizer matching *no* CPython on 174 values — the comment-close and CDATA changes are entangled with it. Caught by this PR's own `test_version_dependent_values_track_a_supported_cpython`, not by inspection.

Now ported: `goahead`'s incomplete-construct dispatch and tail flush, `commentclose --!?>` plus the abrupt `-?>`, `locatetagend` replacing `locatestarttagend_tolerant`, the rewritten `parse_endtag`, the CDATA `interesting` regex, the RAWTEXT/RCDATA element sets with `_escapable`, HTML5 whitespace for `\s`, and `<![CDATA[` replacing the `parse_marked_section` port — which also removes the `AssertionError` Django used to raise on `<![name[`, retiring a 496-cell "no reference answer" class.

Also fixed a **panic**: `parse_html_declaration` indexed `s[i..i+9]` for the `<!doctype` probe, which is not a char boundary when a multi-byte character sits there. Unreachable until the `<![` arm stopped swallowing those inputs; `<![中` then paniced the extension.

**Result** — djust vs each reference, whole corpus:

| | main | this PR |
|---|---|---|
| stable values agreeing (of 2,717) | 1,108 | **2,715** |
| version-dependent tracking some CPython (of 1,283) | 541 | **1,283** |
| stable DIFF | 1,607 | **0** |
| matching **no** supported CPython | 742 | **0** |

Exact against 3.12.13 and 3.13.7 on **1283/1283**; 1052/1283 against 3.14 (the documented charref delta). The two stable non-matches are the DoS guard, where the reference raises.

### The test no longer depends on the runner

`python/tests/fixtures/striptags_reference_2273.json` captures the reference across all four interpreters (`scripts/gen-striptags-reference.py`), split:

- **stable** (2,717) — every captured CPython agrees; djust must match, on any runner.
- **unstable** (1,283) — they disagree, so there *is* no single right answer. Every version's answer is recorded, and djust must still behave like **one** supported CPython.

I picked this over both options you suggested, because each assumed a single correct answer exists: pinned bytes for the EOF shapes would be wrong on 2 of 3 CI versions, and a `sys.version_info` skip has no single threshold — 3.13 and 3.14 both have "the hardening" and still disagree. It also makes the moving reference **visible in the repo** rather than discovered in CI.

`TestPinnedReferenceIsHonest` stops the pin becoming a dumping ground:

- `test_every_unstable_value_is_genuinely_unstable` — a value may only sit in `unstable` if the CPythons really disagree.
- `test_pinned_stable_reference_matches_this_interpreter` — the stable half is re-derived from the runner's **real Django** every run, so the fixture is a record of Django's behaviour, not of whatever the code did that day.
- `test_every_live_compared_value_is_stable` — the guard for the exact mistake that made this red. Three hand-written values in `UNREPORTED_VALUES` were version-dependent.

`scripts/check-striptags-version-stability.py` makes the rule executable for the **Rust pins**, which no pytest run can reach: it re-runs all 70 literal expectations through each supported CPython. It found 2 Rust rows and 5 Python rows that were version-dependent; all replaced or moved.

### Non-regression, same interpreter, same contract

Rebuilt `main` in place and scored both builds against the **version-neutral fixture** (so the two runs compare the same contract, which a Django-referenced baseline could not):

```
                          main      PR
  AGREE                   1108    2715
  TRACKS                   541    1283
  DIFF                    1607       0
  ORPHAN                   742       0

  fixed:  2349        BROKEN: 0
```

Different counts, so the baseline is a real separate build.

### Gate-off: 14/14, after five survivors

The harness grew to 14 mutations. **Five survived their first form** and each was a real gap:

- the loop cap — that input's uncapped fixpoint was also `""`, so the mutation computed the identical answer;
- four ported mechanisms (`--!?>`, the abrupt `-?>`, `<![CDATA[`→`]]>`, the CDATA-close lookahead) that no test could observe at all.

The CDATA-close one is the sharpest: asserted through `strip_tags` it *still* survived, because the re-strip loop repairs a RAWTEXT element that never closes — two mechanisms shadowing each other. It is a single-pass property and is now asserted on `strip_once`.

### Suites

`make test-rust` 1207 passed · `pytest tests/ python/tests/ python/djust/tests/` 12,581 passed, 219 skipped, 0 failed · `cargo clippy --workspace --all-targets -D warnings` clean.

### One follow-up closed itself

#2280 (`is_safe=True` not propagated) was fixed by #2285 while this was open. `TestKnownRemainingDivergences` fired on its own instruction — *"now AGREES — delete this row"* — so `|safe|striptags` moved from a pinned divergence to a tested chain. #2279 and #2281 remain open and pinned.

### Left for a follow-up

The 224-value 3.14 delta (`&`/`&#` at end of input) — djust tracks 3.13 there. I'll file it unless you'd rather I take it in this PR.

Issue #2287, opened 2026-08-27
join/unordered_list ignore needs_autoescape's per-element conditional_escape


johnrtipton:
Split out of #2284 (scope discipline, CLAUDE.md #1079). #2284 fixes the four
`needs_autoescape` built-ins whose body is the **whole-value** form; these two
are `needs_autoescape` with a different body shape and are untouched by it.

## The premise #2284 corrected

The issue body of #2284 says the `needs_autoescape=True` built-ins are four.
Read off Django 5.2's live registry they are **seven**:

```
join  linebreaks  linebreaksbr  linenumbers  unordered_list  urlize  urlizetrunc
```

They partition by the shape of the filter body, which is what decides whether
#2284's fix reaches them:

| names | body | status |
|---|---|---|
| `linebreaks` `linebreaksbr` `urlize` `urlizetrunc` | `autoescape and not isinstance(value, SafeData)` | fixed in #2284 |
| `linenumbers` | same clause, but djust never escaped internally, so the renderer's output escape already lands on Django's answer | already correct (now pinned by `TestLinenumbersWasAlreadyCorrect`) |
| **`join` `unordered_list`** | `conditional_escape` applied **per element** | **this issue** |

## The divergence

```python
>>> render('{{ p|join:", " }}', {"p": [mark_safe("<b>x</b>"), mark_safe("<i>y</i>")]})
django  '<b>x</b>, <i>y</i>'
djust   '&lt;b&gt;x&lt;/b&gt;, &lt;i&gt;y&lt;/i&gt;'

>>> render('{{ p|unordered_list }}', {"p": [mark_safe("<b>x</b>"), mark_safe("<i>y</i>")]})
django  '\t<li><b>x</b></li>\n\t<li><i>y</i></li>'
djust   '\t<li>&lt;b&gt;x&lt;/b&gt;</li>\n\t<li>&lt;i&gt;y&lt;/i&gt;</li>'
```

Measured on `52b727d0` + the #2284 fix, Python 3.12.9 / Django 5.2.16, through
the real `RustLiveView.mark_safe_keys` channel (not `render_template`, which
has no context-safety channel — see `crates/djust_templates/tests/test_xss_prevention.rs`).

Django's bodies:

```python
@register.filter(is_safe=True, needs_autoescape=True)
def join(value, arg, autoescape=True):
    if autoescape:
        data = conditional_escape(arg).join([conditional_escape(v) for v in value])
    ...

[...]

Issue #2286, opened 2026-08-27
striptags: CPython 3.14 changed end-of-input character-reference handling again


johnrtipton:
Follow-up from #2273 / PR #2282, which ports `striptags` to CPython
**3.12.10+/3.13**'s `html.parser`.

## The gap

CPython 3.14 changed `goahead`'s end-of-input handling of `&` and `&#` again,
on top of the 3.12.10 HTML5-spec rewrite. `Lib/html/parser.py`, the `&` branch:

```python
                match = incomplete.match(rawdata, i)
                if match:
                    if end:
                        self.handle_entityref(rawdata[i+1:])   # NEW in 3.14
                        i = self.updatepos(i, n)
                        break
                    break
```

and the `&#` branch gained `incomplete_charref` with the same shape. Before
3.14 an incomplete reference at end of input left through the tail flush as
data; now it is resolved as a reference.

```python
strip_tags("<b>x</b> &am")
# 3.12.13 / 3.13.7  'x &am'
# 3.14.6            'x &am;'

strip_tags("&#xZZ</b>&#;<br />a <b")
# 3.12.13 / 3.13.7  '&#xZZ&#;<br />a <b'
# 3.14.6            '&#xZZ&#;a '
```

## Scale

**224 of 4,000** corpus values in PR #2282's differential. djust currently
matches 3.12.13/3.13.7 on **1283/1283** version-dependent values and 3.14.6 on
1052/1283 — the shortfall is this and nothing else.

## Why it was not taken in #2282

3.13 and 3.14 disagree with each other, so no single fixed port can match both.
#2282 targets 3.13 (2 of the 3 CI interpreters, and the mainstream deployment).
Tracking 3.14 instead would move the same 224 values to the other side of the
ledger, not remove them.

The real question this issue should answer is therefore **not** "port 3.14" but:
*should `striptags` be version-aware at all?* Options:

1. **Leave it.** The affected inputs are unterminated character references in
   adversarial markup. Cost: djust and Django disagree on 3.14 for those.
2. **Track 3.14 and drop 3.13.** Symmetric; no better.
3. **Make the two EOF arms conditional on the host interpreter's version.** The
   Rust side would need the running `sys.version_info` passed in or cached at
   module init. Buys exactness on every interpreter, at the cost of a version
   branch in a hot tokenizer and a second behaviour to test — and it sets a
   precedent for every future `html.parser` change.

Option 3 is the only one that is actually *correct*, and is probably still not
[...]

PR #2288, opened 2026-08-27
fix(templates): needs_autoescape's SafeData term, the half that is reachable (#2284)

### LINKED ISSUE #2284: bug: linebreaks/linebreaksbr/urlize/urlizetrunc ignore needs_autoescape and escape a SafeData input
Django's `linebreaks`, `linebreaksbr`, `urlize` and `urlizetrunc` are registered `needs_autoescape=True` and open with

```python
autoescape = autoescape and not isinstance(value, SafeData)
```

so they **skip their internal escape** when the value handed to them was already safe. djust's escape unconditionally, so `|safe` reaches them and is then escaped away from the inside:

```
{{ p|safe|linebreaks }}    django='<p><b>x</b></p>'   djust='<p>&lt;b&gt;x&lt;/b&gt;</p>'
{{ p|safe|linebreaksbr }}  django='<b>x</b>'          djust='&lt;b&gt;x&lt;/b&gt;'
{{ p|safe|urlize }}        django='<b>x</b>'          djust='&lt;b&gt;x&lt;/b&gt;'
{{ p|safe|urlizetrunc:30 }} django='<b>x</b>'         djust='&lt;b&gt;x&lt;/b&gt;'
```

Found while fixing #2274 and **deliberately not folded into it**: they are a different mechanism. #2274 is the missing `isinstance(obj, SafeData)` term in `FilterExpression.resolve` — a renderer-level decision about the filter's *output*. This is `needs_autoescape` — a filter-level decision about whether to escape its own *input*. #2274's fix leaves these four exactly where they were, which is why they are 4 of the 9 cells still diverging on `{{ p|safe|X }}` after it (down from 28).

Fixing it means threading the input's safety into `filters::apply_filter_full_safe` so these four can consult it, which is a signature change across every filter — worth doing once, deliberately, rather than as a rider.

**Direction of the bug is over-escaping, so there is no security urgency** — djust is strictly safer than Django on all four cells today. The cost is functional: `mark_safe`d HTML passed through `|linebreaks` renders as visible tag text.

johnrtipton:
Closes #2284

## The decision

Django registers four built-ins `needs_autoescape=True` and each opens

```python
autoescape = autoescape and not isinstance(value, SafeData)
```

That is **two independent terms**, and only one is reachable in djust. Which
one matters, because the issue frames the whole thing as a `needs_autoescape`
gap and the two halves have opposite answers.

| term | reachable in djust? | decision |
|---|---|---|
| `autoescape` — the `{% autoescape %}` block policy | **no**. The parser rejects the tag outright. Verified: `{% autoescape off %}…` raises `Unsupported template tag`. | **declined** |
| `not isinstance(value, SafeData)` | **yes**, via `\|safe` and via a view's `mark_safe()` | **implemented** |

So the fix is option 2 of the three offered — implement `needs_autoescape`
hard-wired to the pinned policy — but the reasoning is not "match the code
shape for a future tag". It is that the `SafeData` half is a **live
divergence on every djust build today**, reproduced before touching anything:

```
{{ p|safe|linebreaks }}      django='<p><b>x</b></p>'  djust='<p>&lt;b&gt;x&lt;/b&gt;</p>'
{{ p|safe|linebreaksbr }}    django='<b>x</b>'         djust='&lt;b&gt;x&lt;/b&gt;'
{{ p|safe|urlize }}          django='<b>x</b>'         djust='&lt;b&gt;x&lt;/b&gt;'
{{ p|safe|urlizetrunc:30 }}  django='<b>x</b>'         djust='&lt;b&gt;x&lt;/b&gt;'
```

Every one of these is with autoescape **on**, which is the only case reachable
today. This is not dead configuration; the `needs_autoescape` framing was
hiding a real bug, exactly as the brief warned it might.

### Why `{% autoescape %}` was declined

It is block-scoped policy threaded through every render arm, and its only
effect would be to let templates turn escaping *off*. Nothing diverges today
for want of it, and no option-3 evidence exists. **What would change the
answer:** a reported divergence that requires the tag (a template that must
render a region unescaped and cannot use `|safe` / `mark_safe`), or a Django
built-in whose behaviour depends on the block rather than on `SafeData`. If it
lands, the change is threading one more bool into
`apply_filter_full_safe`'s pinned `true` — which is exactly what this PR's
shape buys. `TestAutoescapeBlockIsStillUnsupported` goes red the day someone
adds the tag and says so.

## Two premises in the issue are wrong

**1. There are seven `needs_autoescape` built-ins, not four.** Read off Django
5.2's live registry:

| names | body shape | status |
|---|---|---|
| `linebreaks` `linebreaksbr` `urlize` `urlizetrunc` | `autoescape and not isinstance(value, SafeData)` | **fixed here** |
| `linenumbers` | same clause | **already correct** — it never escaped internally, so the renderer's output escape lands on Django's answer. `renderer.rs` claimed this in a comment; now run (`TestLinenumbersWasAlreadyCorrect`) |
| `join` `unordered_list` | `conditional_escape` **per element** | **out of scope** — one whole-value bool cannot express it. Filed as #2287, pinned as still-diverging |

**2. The four are not equally clean afterwards.** `urlize`/`urlizetrunc` keep a
separate, pre-existing **URL-detection** gap — djust matches on a regex where
Django word-splits and `smart_urlquote`s. It is identical before and after, and
identical in all three columns, which is what proves it orthogonal to the
escape decision rather than something this PR introduced.

## Safety

The `SAFE_OUTPUT_FILTERS` membership these four have is unconditional, and
#2269/#2259 earned it with the internal escape. Making that escape conditional
does **not** unearn it — the reason widens:

* `input_was_safe == false` → the filter escapes, exactly as before. This is
  every value nothing marked safe, i.e. all hostile input.
* `input_was_safe == true` → the escape is skipped, but only because the
  context or an earlier safe filter declared the value safe. Django makes
  precisely this trade.

`urlize`'s **href** escape stays unconditional, as Django's does (it writes
`escape(url)` outside the `if autoescape` branch), because the href lands
inside `href="…"` and a conditional one is an XSS Django does not have. That is
the obvious way to write this fix and it is invisible to any test that only
looks at the anchor text, so it has its own class and its own gate-off.

## Verification

Python 3.12.9 (repo venv), Django 5.2.16. All numbers reproduced from scratch,
not taken from the issue.

**Differential vs real Django** — 4000-value adversarial corpus × 4 filters × 3
columns (plain / `|safe` / context `mark_safe` driven through the real
`RustLiveView.mark_safe_keys` channel, because `render_template` has no
context-safety channel at all — the first probe that used it reported four
failures production does not have):

| | baseline (`origin/main`, **rebuilt**) | with this change |
|---|---|---|
| differing comparisons | **33,035** / 48,000 | **12,177** / 48,000 |

* fixed: **20,858** · new regressions: **0**
* plain column byte-identical before and after
* `linebreaks` / `linebreaksbr`: **0** differences on all three columns
* urlize residual equals the plain column exactly (1920 / 2139) — the
  detection gap, unchanged

The two runs report different counts (33,035 vs 12,177), and the full suite
reports different totals (12,581 vs 12,710 = +129 new tests), which is the
check that the baseline build was real.

**Full suite**, all roots, both builds: `12581 passed, 0 failed` baseline;
`12710 passed, 0 failed` with the change. `make test-rust` clean;
`cargo clippy --workspace --all-targets -- -D warnings` clean; `cargo fmt
--check` and `ruff` clean.

**Gate-off (#1468)** — six mutations, one per mechanism, harness asserting the
text was found, that the source changed, and counting errors as well as
failures:

| mutation | reddens |
|---|---|
| M1 `linebreaks` conditional escape → always | 17, `TestLinebreaksHonoursSafeInput` + `TestHostileInput…` |
| M2 `linebreaksbr` conditional escape → always | 11, `TestLinebreaksHonoursSafeInput` |
| M3 `urlize` conditional escape → always | 9, `TestUrlizeHonoursSafeInput` + 2 more |
| M4 `urlize` href → conditional | 2, `TestUrlizeHrefIsEscapedUnconditionally` |
| M5 renderer threading, `{{ }}` arms → `false` | 36, five classes |
| M5b firstof/cycle arm → `false` | 6, `TestTheFirstofArmThreadsItToo` + the structural pin |

No surviving mutation. M5b originally reddened **only** a source-grep pin,
which is a decorative pin by #1859's rule — `TestTheFirstofArmThreadsItToo`
was added for that reason and now takes 5 of the 6.

## Files

| file | why |
|---|---|
| `crates/djust_templates/src/filters.rs` | `input_was_safe` through `apply_filter_full_safe` → `apply_builtin_filter` → the four arms; `linebreaks`/`linebreaksbr`/`urlize` take the flag; `apply_filter_full` passes the safe default |
| `crates/djust_templates/src/renderer.rs` | the three call sites pass `runtime_safe` (read before it is reassigned — Django's `obj`); `SAFE_OUTPUT_FILTERS` doc records the widened two-arm reasoning |
| `python/tests/test_needs_autoescape_2284.py` | 129 cases |


johnrtipton:
## Follow-ups filed while doing this

Both are the same class as #2284 on a different mechanism, and both were **measured**, not inferred. Neither is fixable by `input_was_safe`, which is one bool about the value as a whole.

* **#2287** — `join` / `unordered_list` apply `conditional_escape` **per element**. A list whose *elements* were `mark_safe`d comes through live in Django and escaped in djust. Pinned as still-diverging by `TestSequenceShapeIsOutOfScopeAndStillDiverges`, which goes red when it is fixed and says what to delete.
* **#2290** — a **custom** `needs_autoescape` filter never sees `SafeData` at all. Django hands it `('SafeString', True, True)` after `|safe`; djust hands it `('str', False, True)`, so a project cannot implement Django's contract. The `autoescape=true` djust passes to `apply_custom_filter` is correct (there is no `{% autoescape %}`); the missing input is the value's safety across the PyO3 boundary. After this PR `apply_filter_full_safe` already carries the answer, so the fix is a `mark_safe` wrap gated on `input_was_safe`.

Both diverge in the **over-escaping** direction, like #2284.

Also #2289 — `htmlparser.rs:16` says "CPython 3.12" where it means 3.12.10+/3.13, found while deciding #2286. Docs-only, unrelated file, not folded in.

## Extra commit

`6b505efb` tightens the `linebreaks` header. Its bold sentence — the one carrying the XSS reasoning — still read "This filter escapes its own input" unconditionally, which the new `# autoescape` section below it contradicts. A reader who stops at the bold claim would have read something the code no longer does; that is the comment-narrower-than-the-code shape #2214 recorded. Docs only; full suite and `make test-rust` re-run clean after it (12710 passed).

johnrtipton:
## Two more commits (`bcc544dd`, `63007d41`)

Self-review caught that the Rust side had **one-variant coverage of a two-variant surface** — the rc4 rule.

`linebreaks_escapes_its_input_which_is_what_makes_marking_it_safe_safe` is the test #2269 wrote to guard the `SAFE_OUTPUT_FILTERS` membership. It proves the escape *happens*. Since this PR the escape is **conditional**, so "it happens" is only half of what the membership rests on — the other half is that the `autoescape=false` arm emits verbatim and is reachable *only* through the flag. Left as-is, the file's own contract test would have kept passing while covering one arm of two, which is the shape that makes a bug class look handled.

Added beside it, so they sit next to the contract and run without PyO3:

* `the_needs_autoescape_filters_skip_the_escape_only_when_told_to` — verbatim under `false`, escaped under `true`, and `apply_filter` (the chain-blind entry point) still escapes.
* `urlize_escapes_the_href_even_when_autoescape_is_off` — the href escape is not conditional.

Rust gate-off (cargo-only): the three escape mutations red the first, the href mutation reds **only** the second. Total across both sides is now **10 mutations, 0 survivors**.

Also `1dfdafa6`: the `apply_filter_full` source pin matched a literal string after a hand-rolled whitespace squeeze, so a `cargo fmt` rewrap would have false-failed it. It parses the argument list now; gate-off re-verified (`false` → `true` reds exactly that one test).

Re-run after all of it: `12710 passed, 0 failed` (Python, all roots) · `make test-rust` clean, 398 in `djust_templates --lib` (was 396) · clippy `-D warnings` clean · `cargo fmt --check` clean.

Issue #2294, opened 2026-08-27
bug: four measuring-filter divergences found grepping the sink for #2279 (stringformat %Ns width, center tie-break, truncatechars combining marks, dict length)


johnrtipton:
Four measuring-filter divergences found while grepping the sink for #2279 (PR pending). None is the byte-vs-char defect that issue is about; each is a separate Python semantic. Filed together because they were found by one sweep and each is small — split if that is preferred.

All measured against a live Django 5.2.16 through `Template(src).render(Context({"p": v}))` on both sides.

---

### 1. `stringformat:"Ns"` ignores the width entirely

```
{{ p|stringformat:"10s" }} on 'ab'   django='        ab'  djust='ab'
{{ p|stringformat:"5s"  }} on ''     django='     '       djust=''
{{ p|stringformat:"4s"  }} on '中'   django='   中'        djust='中'
```

`apply_filter`'s `stringformat` arm implements width only for the numeric conversions (`d`, `o`, `x`, …); the `s` conversion drops it. `%-10s`, `%.3s` and `%10.3s` are presumably affected the same way and should be checked at the same time. Divergence on the broad differential: 132 / 174 at `"20s"`, 91 / 174 at `"10s"`.

### 2. `center` uses Rust's tie-break, not Python's

Python's `str.center` computes `left = marg // 2 + (marg & width & 1)`, which biases LEFT for some odd-margin cases; Rust's `{:^width$}` always puts the smaller half on the left.

```
'ab'.center(5)    django='  ab '  djust=' ab  '
'abcd'.center(5)  django=' abcd'  djust='abcd '
'ab'.center(3)    django=' ab'    djust='ab '
```

`'a'.center(4)` and `'abc'.center(6)` agree, so a curated table with even margins misses it entirely. `ljust`/`rjust` are correct — Rust's format width is a char count, so the byte-vs-char question does not arise here. 34 / 174 cells at `center:5`, 27 / 174 at `center:3`.

### 3. `truncatechars` counts combining marks; Django skips them

`django.utils.text.Truncator._text_chars` opens its loop with

```python
if unicodedata.combining(char):
    # Don't consider combining characters as adding to the string length
    continue
```

`crates/djust_templates/src/truncate.rs::text_chars` has no such skip.

```
{{ p|truncatechars:5 }} on 'ééééé'  (e+U+0301 x5)  django='ééééé'  djust='éé…'
[...]

Issue #2293, opened 2026-08-27
bug: wordwrap is a greedy re-joiner, not textwrap.TextWrapper (and its byte width cannot be fixed alone)


johnrtipton:
Found while grepping the sink for #2279 (PR pending), which lists `wordwrap` as a place to check for the byte-vs-char defect. It is one — but the sink **cannot be fixed on its own**, which is why #2279 left it alone.

## `wordwrap` is not `django.utils.text.wrap`

Django 5.x's `wrap()` delegates to `textwrap.TextWrapper`:

```python
wrapper = textwrap.TextWrapper(
    width=width, break_long_words=False, break_on_hyphens=False,
    replace_whitespace=False,
)
result = []
for line in text.splitlines():
    wrapped = wrapper.wrap(line)
    if not wrapped:
        result.append(line)      # a whitespace-only line is restored
    else:
        result.extend(wrapped)
if text.endswith("\n"):
    result.append("")            # a trailing newline is preserved
return "\n".join(result)
```

`crates/djust_templates/src/filters.rs::word_wrap` is a greedy re-joiner: `text.split_whitespace()` then join on single spaces. It therefore

1. **flattens every existing line break** into a space;
2. **collapses runs of spaces** and drops leading indentation;
3. **measures widths in BYTES** (`word.len()`), so every non-ASCII word measures long and breaks its line early;
4. returns the text unchanged at `width=0`, where Django raises `ValueError("invalid width 0 (must be > 0)")`.

## The byte fix alone makes it worse

`word.len()` → `word.chars().count()` was implemented and measured on a randomized differential against Django. It fixes **21** cells and **regresses 6**, and every regression is a string containing `U+2028`:

```
{{ p|wordwrap:10 }} on 'f 字日🇺🇸0🎉'
  django  'f\n字日🇺🇸0🎉'
  before  'f\n字日🇺🇸0🎉'      (agreed)
  after   'f 字日🇺🇸0🎉'       (disagrees)
```

Django's `splitlines()` breaks at `U+2028`; the rejoiner turns it into a space. The byte overcount had been putting a break at that position *by accident*. Two bugs cancelling — the #2272 pattern — so the pair has to move together.

## Measured divergence (before any fix)

Randomized differential, 174 values per width:

| template | disagree |
|---|---|
| `{{ p\|wordwrap:0 }}` | 173 / 174 (Django raises) |
| `{{ p\|wordwrap:20 }}` | 53 / 174 |
[...]

Issue #2292, opened 2026-08-27
bug: py_repr_string does not escape non-ASCII non-printable code points, and no fixed table can


johnrtipton:
Split out of #2277 (PR pending) with the measurement, per scope discipline.

`djust_core::py_repr_string` — the one definition behind `{{ list }}`, `{{ dict }}` and now `{{ x|pprint }}` — escapes `\`, the active quote, `\t`, `\n`, `\r`, the rest of C0 and DEL, and stops. CPython escapes any code point for which `str.isprintable()` is false, so:

```python
_rust.render_template("{{ p|pprint }}", {"p": " "})   # djust  "'<U+2028 literal>'"
pprint.pformat(" ")                                   # Python "'\\u2028'"
```

Same for `U+00A0` (Zs), `U+200B`/`U+FEFF` (Cf), `U+2029` (Zp), and every private-use / unassigned code point.

## Why it was not fixed with #2277

`str.isprintable()` is Unicode-version data and **the CI matrix disagrees with itself**. Measured across the interpreters this project runs:

| interpreter | `unicodedata.unidata_version` | printable code points |
|---|---|---|
| 3.12.9 | 15.0.0 | 148998 |
| 3.12.13 | 15.0.0 | 148998 |
| 3.14.6 | 16.0.0 | 154810 |

5812 code points changed classification between Unicode 15.0 and 16.0. A fixed table in Rust is green on one half of the matrix and red on the other — exactly the situation the `striptags` port hit (#2273, "the reference moves; the port does not"), which was resolved there by pinning a captured reference rather than by porting the moving part.

## Scope

Only `Cn` (unassigned) actually drifts; `Cc`, `Cf`, `Cs`, `Co`, `Zl`, `Zp`, `Zs` are small, stable sets. So a port that escapes those seven categories and treats `Cn` as printable would be exact for every *assigned* code point and diverge only where the interpreters already diverge. That is a defensible target and is what this issue asks for. It needs either a small hand-written range table or a Unicode-category dependency (`unicode-general-category`), which is a dependency discussion.

## Measured impact

On a randomized `pprint` differential against real `pformat`:

| corpus | divergence |
|---|---|
| printable text only | 0 / 4000 |
| + ASCII controls, quotes, backslashes | 0 / 3000 |
| + `U+00A0` / `U+200B` / `U+2028` / `U+FEFF` | 1488 / 4000 |
| the same, with an exact `repr` substituted for the scalar spelling | 0 / 4000 |

[...]

PR #2295, opened 2026-08-27
fix(templates): length counts code points, pprint wraps at width 80 (#2279, #2277)

### LINKED ISSUE #2277: bug: `pprint` never wraps, where `pprint.pformat` wraps at width=80
Found while fixing #2270 and deliberately left out of it (#1079 scope discipline).

Django's `pprint` filter is `pprint.pformat(value)`. `pformat` defaults to `width=80` and **wraps**: past that width it breaks the structure across lines with hanging indentation. `crates/djust_templates/src/filters.rs`'s `pprint_value` builds a single line and never wraps, so every structure whose repr exceeds 80 characters diverges — by every newline and every indent space.

Measured against a live Django 5.2.16:

| value | Django newlines | djust newlines |
|---|---|---|
| `[1.5] * 40` | 39 | 0 |
| `{f"k{i:02d}": float(i) for i in range(20)}` | 19 | 0 |

Both agree character-for-character below the width, which is why #2270's differential (whose container shapes are all short) reports `pprint` at full parity — the divergence is entirely in the wrapping axis that sample did not reach. Same shape as the v1.0.0rc4 finding: single-variant coverage of a multi-variant surface.

Pre-dates #2270 and is unrelated to floats — reproduces on `["a"] * 40`.

If taken, the target is CPython's `pprint._safe_repr` / `PrettyPrinter._format` layout rules, not an approximation: the wrap points, the hanging indent, the `sort_dicts=True` default and the one-item-per-line-past-the-limit behaviour are all observable and all differ between a list, a tuple and a dict. Worth pairing with a randomized differential against `pprint.pformat` (the v1.1.1-2 rule) rather than a curated table, since the interesting cases are exactly the ones nobody thinks to enumerate.

Related: #2270, #2203 (`pprint` renders Python repr).

### LINKED ISSUE #2279: bug: `{{ p|length }}` counts BYTES, not characters

johnrtipton:
Two Python semantics djust did not reproduce — not filter algorithms got subtly wrong, but the underlying model. Shipped together because they share a sink: `pprint`'s scalar spelling is the same `py_repr_string` the `{{ list }}` path uses, and both issues are "what does Python actually do here".

Closes #2279
Closes #2277

| issue | file | covering tests |
|---|---|---|
| #2279 `length` counts bytes | `crates/djust_templates/src/filters.rs` | `TestLengthCountsCodePoints` (37 cases) |
| #2277 `pprint` never wraps | `crates/djust_templates/src/pprint.rs` (new), `crates/djust_core/src/lib.rs` | `TestPprintWraps` (43 cases) |
| residuals not closed | — | `TestKnownResidualDivergences` (10 cases) |

---

## #2279 — `length` counted BYTES

`str::len()` is a byte count; Python's `len()` is a code-point count. `é` is 2 bytes, `中` is 3, so every non-ASCII string measured long: `{{ "中<b"|length }}` gave `5` where Django gives `3`.

**Code points, not graphemes** — the trap the fix could have walked into by "improving" on Python. `len("👍🏽")` is 2 (thumb + skin-tone modifier) and `len("👨‍👩‍👧")` is 5 (three emoji + two joiners); a grapheme-cluster count answers 1 to both and is a *different* wrong answer. Rust's `char` is a Unicode scalar value, so `chars().count()` is Python's answer exactly. `test_grapheme_clusters_are_not_what_python_counts` asserts it.

The bug was masked in the #2273 `striptags` sweep — the old `striptags` deleted the tail after a lone `<`, so the byte count of what survived matched Django's char count of the whole value. `test_striptags_parity_2273.py`'s pinned row for it is deleted here, per its own "now AGREES — delete this row" assertion.

### Every length-measuring sink, and what the grep found

`length` was the only one wrong on the byte-vs-char axis.

| sink | code-point correct? |
|---|---|
| `length` | **NO** — fixed here |
| `slice` (string) | yes — `chars().collect()` (`filters.rs::apply_slice`) |
| `first` / `last` | yes — `chars().next()` / `chars().last()` |
| `make_list` | yes — `s.chars()` |
| `truncatechars` / `truncatechars_html` | yes on the byte axis — `char_indices()` in `truncate.rs::text_chars` (#2272's port) |
| `truncatewords` / `truncatewords_html` | yes — `py_split_whitespace` |
| `ljust` / `rjust` / `center` | yes — Rust's `{:<width$}` pads on `chars().count()` |
| `wordcount` | yes — `split_whitespace().count()` |
| `get_digit` | yes — ASCII digits only |
| `linenumbers` | yes — a line count |
| `stringformat` numeric widths | yes — ASCII digits only |
| `wordwrap` | **NO** — deliberately not fixed, see below |

### `wordwrap` is left alone on purpose

It is the same sink (`word.len()` is bytes), and the char fix was implemented and **measured**: it fixes 21 differential cells and **regresses 6**, every regression a `U+2028` string where the byte overcount had been putting a line break at the position Django's `splitlines` breaks. Two bugs cancelling. djust's `wordwrap` is not `textwrap.TextWrapper` at all (Django 5.x rewrote `wrap()`) — it flattens newlines, collapses spaces, and returns the text at `width=0` where Django raises. The pair goes with that port: **#2293**.

---

## #2277 — `pprint` never wrapped

Django's filter is `pprint.pformat(value)`; `pformat`'s default `width=80` **wraps**, breaking a structure across lines with hanging indentation. The filter built one line, so `[1.5] * 40` was 39 newlines in Django and 0 here.

It is a real line-breaking algorithm, not a width check. Ported into `crates/djust_templates/src/pprint.rs`:

* `_format` — computes the FLAT repr first and emits it whenever it fits in `width - indent - allowance`, dispatching to a per-type layout only when it does not. This is why a short container never wraps however deeply it nests, and why `allowance` (the room every enclosing container's closing brackets still need) has to be threaded through.
* `_format_items` (list / tuple) — one element per line at `indent + 1`, `allowance` to the last element and `1` to the rest.
* `_format_dict_items` — values indent past `len(repr(key)) + 2`, so they line up under the value column.
* `_pprint_str` — splits a too-long string at whitespace and parenthesises the chunks at `level == 1` (implicit concatenation, so the output still evals back).
* `py_splitlines_keepends` — Python's `str.splitlines(keepends=True)`. Rust's `lines()` splits on `\n` alone; Python breaks on eight more, `U+2028` included, and the wrap depends on it.

**Scoping was decided on a measurement, not a feeling.** A Python model of exactly this algorithm was written first and run against real `pformat`:

| corpus | model divergence |
|---|---|
| printable text | 0 / 4000 |
| + ASCII controls, quotes, backslashes | 0 / 3000 |
| + `U+00A0` / `U+200B` / `U+2028` / `U+FEFF` | 1488 / 4000 |
| the same, with an exact `repr` substituted | **0 / 4000** |

The last row proves the layout is exact and the whole residual is the scalar spelling — so a full port was worth doing, and the escaping half is separable. Precedent weighed: #2272 ported `Truncator` + its `HTMLParser` (2,128 lines) because that filter's behaviour *was* the parser; `pprint`'s layout is ~200 lines and buys full parity, so it clears the same bar.

Scalars now go through `djust_core::py_repr_string`, the one definition the `{{ list }}` path uses (#1646) — `pprint` had a second, bare `format!("'{s}'")` that escaped nothing at all. That helper grew the ASCII control escapes it was missing, so `{{ ['a\tb']|pprint }}` is now `['a\\tb']` on both paths.

### What the randomized differential caught that a table did not

**Dict key ordering — a defect the port itself introduced.** `_safe_repr` sorts `sorted(object.items(), key=_safe_tuple)`, i.e. by the KEY. The first version sorted the RENDERED pairs, which the filter it replaced had also done and got away with, because that filter quoted every key `'...'` unconditionally. Once keys went through a real `repr`, the opening quote became `"` for a key holding a `'` (`0x22` vs `0x27`) and escapes shifted the rest, so one key's content reordered it against every other. **6.4% of a 4000-value corpus**; zero cells in the curated table, and zero in the broad differential whose keys were all `k000`-shaped. `test_dict_keys_sort_by_key_not_by_rendered_repr` pins it.

---

## Verification

**Reproduced both first**, against real Django 5.2.16, before any change.

**Interpreter**: measured on the repo `.venv` (3.12.9). `pprint` output was checked to be **identical** across 3.12.9 / 3.12.13 / 3.14.6 on a 400-value corpus (same SHA-256), so unlike `html.parser` (#2273) the `pprint` reference does not move. `str.isprintable()` *does* — see the residual below.

**Broad differential, set comparison against a rebuilt `origin/main`** (rebuilt, not stashed — `git stash` does not move HEAD). 13,751 cells over every measuring filter × unicode classes × containers:

```
before  13751 cells,  2421 disagree
after   13751 cells,  1998 disagree

FIXED (disagree -> AGREE): 423
    113  {{ p|striptags|length }}
    106  {{ p|length }}
    106  {{ p|length }}{{ p|length }}
     97  {{ p|pprint }}
      1  {{ p|make_list }}

REGRESSED (AGREE -> disagree): 0
```

The two counts differ, which is the check that the baseline is a real second build.

**Unmasked cells attributed, not assumed.** Every one of the 269 remaining `{{ p|pprint }}` disagreements was classified mechanically:

```
  230  non-ASCII non-printable AND tuple
   29  tuple flattened by normalize_django_value (a serializer artifact)
   10  non-ASCII non-printable
    0  UNEXPLAINED
```

On the corpus with both excluded, `pprint` is **0 / 4000** against real `pformat`.

**Gate-off (#1468)** — nine mutations, one per mechanism. The harness asserts the mutation text was found and that the source changed, and treats a build failure or a pytest `N error` as INVALID rather than green.

| mutation | tests reddened |
|---|---|
| M1 `length`: code points → bytes | 31 — incl. `TestLengthCountsCodePoints::test_matches_python_len[…]`, `…::test_the_striptags_chain_now_agrees` |
| M2 `pprint`: never dispatch to a wrapping layout | 21 — incl. `TestPprintWraps::test_the_issues_two_reproducers` |
| M3 `pprint`: sort dict keys by the rendered pair | 2 — `test_dict_keys_sort_by_key_not_by_rendered_repr` + the randomized differential |
| M4 `py_repr_string`: drop the ASCII-control escapes | 3 — `TestKnownResidualDivergences::test_ascii_controls_in_the_same_string_are_escaped` |
| M5 `py_repr_string`: always single-quote | 3 — same, plus the key-order test |
| M6 `pprint`: no `_pprint_str` dispatch | 12 — the long-string rows of `test_layout_matches_pformat` |
| M7 `py_splitlines_keepends`: `\n` only | 5 — `test_layout_matches_pformat[value33]`, `…[word …]`, `test_the_residual_is_the_escape_and_not_the_layout` |
| M8 `pprint`: values do not indent past the key column | 4 — `test_layout_matches_pformat[value14/16/23]` |
| M9 `pprint`: `allowance` not threaded to the last element | 3 — `test_layout_matches_pformat[value29/30]` |

None survived and none was INVALID. M7's and M9's curated discriminators were found by mutating the Python model rather than guessed — a value whose repr is 77 characters distinguishes `allowance` 1 from 2, and no smaller table reaches it.

**Suites**: `pytest tests/ python/tests/ python/djust/tests/ -n auto` → **12801 passed, 219 skipped, 0 failed**. `make test-rust` → 62 result lines, all ok, 0 failed. `cargo clippy --workspace --all-targets -- -D warnings` clean. `cargo fmt` / `ruff format` clean.

`filters::float_sink_set::every_float_to_string_sink_routes_through_an_approved_repr` (the #2258/#2270 structural pin) fired correctly when the `Value::Float` arm moved from `filters.rs` to `pprint.rs`, and its expected set is updated deliberately.

---

## Follow-ups filed, with their measurements

* **#2292** — `py_repr_string` does not escape non-ASCII non-printables, and no fixed table can: `str.isprintable()` is Unicode-version data and 3.12/3.13 (Unicode 15.0, 148998 printable) disagree with 3.14 (16.0, 154810) about **5812 code points**. The #2273 "the reference moves" situation. Pinned by `TestKnownResidualDivergences`, which fires its own delete-this-row assertion when closed.
* **#2293** — `wordwrap` is a greedy re-joiner, not `textwrap.TextWrapper`, and its byte width cannot be fixed alone (the 21-fixed / 6-regressed measurement above).
* **#2294** — four more measuring-filter divergences the sink grep turned up, none of them byte-vs-char: `stringformat:"Ns"` ignores the width entirely (132/174 cells at `"20s"`), `center` uses Rust's odd-margin tie-break instead of Python's, `truncatechars` counts combining marks where Django's `_text_chars` skips them, and `{{ dict|length }}` answers 0.

## Incidental

`ruff format` (the repo's own pre-commit hook) normalized `python/tests/test_striptags_parity_2273.py` on first touch — it had landed unformatted. That is the 148-line reformat in the diff; the substantive change to that file is the deleted `|striptags|length` pin row and its docstring.


johnrtipton:
Finding addressed, plus both smaller notes. Two commits, no production behaviour change.

## The refactor-with-helper gap (#1195)

You were right, and the diagnosis was exact: `py_repr_string` has two callers and every test reached it through `pprint`. `TestContainerReprUsesTheSameEscaper` (`python/tests/test_length_pprint_parity_2279_2277.py`) covers the other one.

**10 values × 4 positions**, each asserted against Django *and* against Python's own `repr`:

| | |
|---|---|
| values | `'a\tb'` `'c\nd'` `'e\rf'` `'g\x1bh'` `'i\x7fj'` `'k\x00l'` `'m\x0bn'` `"m'n"` (quote flip) `'o\'p"q'` (both quote kinds) `'r\\s'` |
| positions | list element, dict **value**, dict **KEY**, tuple, nested list |

Three notes on the shape:

- `test_the_table_states_pythons_answer` asserts `repr(value) == want_repr` first, so the other four cases cannot pin a wrong expectation.
- The tuple case passes the raw Python tuple to `render_template` rather than going through `normalize_django_value` — the serializer flattens a tuple to a list, so a tuple routed through `render_both` never reaches the `Value::Tuple` arm at all.
- The dict-**key** position is there because `{{ dict }}` once had a hand-rolled escaper in exactly that slot which missed the backslash (#2203 review). It is the position most likely to drift again.
- `test_the_two_callers_agree_on_the_same_value` asserts the container path and the `pprint` path emit the same spelling — a check on the convergence itself rather than on either caller, so it goes red for any mutation that moves only one side.

## Gate-off, re-run

All 9 mutations still redden a named test; none INVALID; the harness still asserts the mutation text was found and that the source changed. The two that touch the helper are the ones that moved:

| mutation | before | after |
|---|---|---|
| M4 `py_repr_string`: drop the ASCII-control escapes | 3 | **8** |
| M5 `py_repr_string`: always single-quote, never escape the quote | 3 | **8** |

Both now redden **both sides of the helper**:

```
M4 -> 8 failed
  TestContainerReprUsesTheSameEscaper::test_in_a_list[a\tb-'a\\tb']
  TestContainerReprUsesTheSameEscaper::test_in_a_dict_value[a\tb-'a\\tb']
  TestContainerReprUsesTheSameEscaper::test_in_a_dict_key[a\tb-'a\\tb']
  TestContainerReprUsesTheSameEscaper::test_in_a_tuple_and_when_nested[a\tb-'a\\tb']
  TestContainerReprUsesTheSameEscaper::test_the_two_callers_agree_on_the_same_value
  TestKnownResidualDivergences::test_ascii_controls_in_the_same_string_are_escaped
  TestPprintWraps::test_dict_keys_sort_by_key_not_by_rendered_repr
  TestPprintWraps::test_randomized_differential_against_django
```

M5 is the same set with the `m'n` parameters. Previously a mutation inside the helper reddened only `pprint`-side tests, so it could not tell you which caller you had broken — which was your point.

Unchanged: M1 31, M2 21, M3 2, M6 12, M7 5, M8 4, M9 3.

## Note 1 — what completing the escaping would take

`py_repr_string`'s doc comment now carries the actionable half, not just the constraint:

> Only `Cn` (unassigned) actually drifts between Unicode versions; `Cc`, `Cf`, `Cs`, `Co`, `Zl`, `Zp` and `Zs` are small, stable sets. So either (a) carry a table of those seven categories and treat `Cn` as printable, which is exact for every ASSIGNED code point and diverges only where the interpreters already diverge — needing a hand-written range table or a Unicode-category dependency, which is a dependency discussion; or (b) accept a pinned divergence, the route `striptags` took (#2273) […] What is NOT available is one fixed table green on every runner.

Both routes are restated in #2292 so the decision lands there.

## Note 2 — the known-wrong pins now say so

Renamed, and the class docstring leads with it:

- `test_length_of_a_dict_is_still_zero` → `test_length_of_a_dict_is_WRONGLY_still_zero_bug_2294`
- `test_non_ascii_non_printables_are_not_escaped` → `test_non_ascii_non_printables_are_WRONGLY_not_escaped_bug_2292`

```
class TestKnownResidualDivergences:
    """**Every test here pins a KNOWN-WRONG answer, not a correct one.**
    ...
    Read no row here as a statement that djust's answer is right -- the filed
    issue number is on each one: #2292 for the escaping gap, #2294 for
    `{{ dict|length }}`.

    The failure mode this framing guards against is real and recent: three
    artifacts pinned a buggy arrangement as if correct and let a shipped XSS
    survive in `linenumbers`, one of them saying so in its own test name.
    """
```

The dict-length docstring also now explains *why* zero was left rather than implying it is defensible: `Value::Object` carries both a dict and a model instance, and `len(model)` raises `TypeError` which Django's `length` catches and answers 0 to — so the right answer is 0 for one of the two things the variant represents and `len()` for the other, and telling them apart belongs with #2294.

## CHANGELOG

The `pprint` entry illustrated the new escapes only through `pprint`. It now names the second caller explicitly:

> **That last part also changes `{{ list }}` and `{{ dict }}`**, the helper's other caller: `{{ ['a\tb'] }}` rendered a literal tab and now renders `['a\\tb']`, which is Django's answer. Covered by `TestContainerReprUsesTheSameEscaper` across list, dict-value, dict-KEY, tuple and nested positions, because a mutation inside a shared helper must redden a test on each side of it (#1195).

## Verification

- `pytest tests/ python/tests/ python/djust/tests/ -n auto` → **12852 passed, 219 skipped, 0 failed** (was 12801; +51 from the new class).
- `make test-rust` → 62 result lines, all ok, 0 failed.
- `cargo clippy --workspace --all-targets -- -D warnings` clean; `cargo fmt` / `ruff format` clean.
- Branch rebased and 0 commits behind `origin/main`.

johnrtipton:
## Retrospective — PR #2295 (pipeline-drain, v1.1.1-7)

**Task**: drained #2279 (`length` counts bytes) and #2277 (`pprint` never wraps).

### Quality: 5

### What went well

- **Declining a fix on measurement.** #2279's body names `wordwrap` as a length sink. It is one — and implementing the char fix moved 21 cells right and **6 wrong**, every regression a `U+2028` string where the byte overcount had been landing breaks exactly where Django's `splitlines` does. Two bugs cancelling. Reverted and filed as #2293 with the numbers, rather than shipping a net-negative change because the issue asked for it. This is the "an issue's proposed remedy has the same epistemic status as its cited location" rule (v1.1.1-2) applied without being prompted.
- **Deciding scope by building a model first.** #2277's port was chosen after writing a Python model of `pformat`'s algorithm and measuring it: 0/4000 divergences with an exact `repr` substituted, proving the layout was exact and the entire residual was scalar spelling. The scope line was drawn on that number, not on intuition.
- **The randomized differential caught a defect the port itself introduced** — dict keys sorted by rendered pair rather than by key, 6.4% of 4000 values — which no curated table reached. The old filter had been getting away with it by quoting every key identically.
- **Non-regression by set comparison against a rebuilt baseline**: 2421 → 1998 disagreements, FIXED 423 / REGRESSED 0, with all 269 residual `pprint` cells classified mechanically to **0 unexplained**. Different cell counts confirm the baseline was genuinely rebuilt.

### What didn't

- **The extracted helper's second caller shipped untested** (review finding, fixed before merge). `py_repr_string` was correctly converged out of two escapers per #1646, but `Value::py_repr` — the `{{ list }}` / `{{ dict }}` path — gained control-character escaping with no test reaching it; every new test arrived via `pprint`. Gating off the escape would not have told you which caller broke. This is CLAUDE.md #1195 (refactor-with-helper guard audit) and it recurs because the *second* caller is the one you aren't thinking about. Fixed with `TestContainerReprUsesTheSameEscaper` (10 values × 4 positions, plus a convergence test asserting the two callers agree); the two helper mutations went 3 → 8 failures and now redden both sides.
- **Known-wrong pins were named as if correct.** `test_length_of_a_dict_is_still_zero` reads as certifying zero. Renamed to `..._is_WRONGLY_still_zero_bug_2294`, with the class docstring leading "Every test here pins a KNOWN-WRONG answer" — the `linenumbers` precedent from #2291 in this same milestone is exactly what that costs when it goes unmarked.

### Verified

12852 passed / 219 skipped / 0 failed. `make test-rust` 62/62. clippy `-D warnings` clean. 9 gate-off mutations, one per mechanism, each reddening named tests (31/21/2/8/8/12/5/4/3); none survived, none INVALID. 0 commits behind `origin/main` at merge. All 15 CI checks green.

### Follow-ups filed with measurements

#2292 (`isprintable()` is Unicode-version data — 3.12/3.13 vs 3.14 disagree on 5812 code points), #2293 (`wordwrap` ≠ `textwrap.TextWrapper`), #2294 (four further divergences the sink grep found, none byte-vs-char).

Issue #2291, opened 2026-08-27
security: {{ p|linenumbers|safe }} emits attacker markup — linenumbers defers its escape to render time


johnrtipton:
Found by the registry-wide probe written for #2281, not by inspection.

`linenumbers` in Django is `@register.filter(is_safe=True) @stringfilter` and escapes **each line inside the filter body**:

```python
def linenumbers(value, autoescape=True):
    lines = value.split("\n")
    width = str(len(str(len(lines))))
    if not autoescape or isinstance(value, SafeData):
        for i, line in enumerate(lines):
            lines[i] = ("%0" + width + "d. %s") % (i + 1, line)
    else:
        for i, line in enumerate(lines):
            lines[i] = ("%0" + width + "d. %s") % (i + 1, escape(line))
    return mark_safe("\n".join(lines))
```

djust's `add_linenumbers` does **not** escape; it relies on the render-time auto-escape instead. `renderer.rs`'s `SAFE_OUTPUT_FILTERS` doc block states the exclusion deliberately, arguing the two are byte-identical "because everything it adds is escape-invariant".

That reasoning holds only while the render-time escape actually runs. A later `|safe` suppresses it, and then nothing escaped the input at all:

```python
from djust import _rust
_rust.render_template('{{ p|linenumbers|safe }}',
                      {"p": "<img src=x onerror=alert(1)>"})
# djust   '1. <img src=x onerror=alert(1)>'      <- LIVE
# django  '1. &lt;img src=x onerror=alert(1)&gt;'
```

Also reachable without a trailing `|safe`: the differential found it on `{{ p|linenumbers|truncatechars_html:"5" }}` too.

This is the same class as #2281 (`escape` was also a no-op deferring to render time, and `|safe` was also what defeated it), and the fix is the same shape as #2259's for `linebreaks`: move the escape INSIDE the filter and add the name to `SAFE_OUTPUT_FILTERS` in the same commit — the two halves are one change.

Fixed in the #2281/#2283 PR, since that PR's probe is what surfaced it and shipping a probe that measures a live XSS without fixing it is not defensible.

Issue #2274, opened 2026-08-27
bug: |safe does not survive an is_safe=True filter — 24 of Django's 27 diverge on {{ p|safe|X }}


johnrtipton:
Found while measuring the chain axis of #2262 / #2261 (PR #2272), which
included a `safe|` prefix in its randomized differential. Reproduces on
`origin/main` at d6fab08f, independent of that PR.

## The rule djust implements vs the rule Django implements

Django's `FilterExpression.resolve`:

```python
if getattr(func, "is_safe", False) and isinstance(obj, SafeData):
    new_obj = mark_safe(new_obj)
```

Two clauses: the filter is declared `is_safe=True` **and the input was already
`SafeData`**. djust models only the first, as a name whitelist
(`SAFE_OUTPUT_FILTERS` in `crates/djust_templates/src/renderer.rs`), so there is
no way for `|safe` to propagate through a subsequent filter.

```python
from djust import _rust
_rust.render_template('{{ p|safe|lower }}', {"p": "<b>&x</b>"})
# djust '&lt;b&gt;&amp;x&lt;/b&gt;'   Django '<b>&x</b>'
_rust.render_template('{{ p|safe|capfirst }}', {"p": "<b>&x</b>"})
# djust '&lt;b&gt;&amp;x&lt;/b&gt;'   Django '<b>&x</b>'
_rust.render_template('{{ p|safe|wordwrap:40 }}', {"p": "<b>&x</b>"})
# djust '&lt;b&gt;&amp;x&lt;/b&gt;'   Django '<b>&x</b>'
```

`{{ p|safe }}` alone agrees, and `{{ p|lower|safe }}` agrees, so the defect is
specifically the propagation.

## Scale

Probing every one of Django's `is_safe=True` filters with `{{ p|safe|X }}` on
`"<b>&x</b>"`, against a real `origin/main` build:

* **24 of 27 diverge** — `addslashes`, `capfirst`, `center`, `escape`,
  `iriencode`, `linebreaks`, `linebreaksbr`, `linenumbers`, `ljust`, `lower`,
  `phone2numeric`, `pprint`, `rjust`, `slugify`, `stringformat`, `striptags`,
  `title`, `truncatechars`, `truncatechars_html`, `truncatewords`,
  `truncatewords_html`, `urlize`, `urlizetrunc`, `wordwrap`.
* Without the `|safe`, only **5 of 27** diverge, so this axis dominates.

[...]

PR #2296, opened 2026-08-27
fix(templates): escape is eager, and sequence filters iterate a string (#2281, #2283)

### LINKED ISSUE #2281: bug: `{{ p|escape|X }}` passes the UNescaped value to X

### LINKED ISSUE #2283: bug: sequence filters do not iterate a string as characters — join/unordered_list/safeseq/escapeseq/random
Five filters that Django treats a string as a **sequence of characters** for, and djust does not — so djust returns a shape Django never produces. Found while fixing #2274; separated per the scope rule rather than folded in.

Measured on `08b42e7a`, payload `<b>x</b>`. These are the filters that diverge in **both** the `{{ p|X }}` and `{{ p|safe|X }}` columns, which is what distinguishes them from the 19 `|safe` cases #2274 closes:

```
{{ p|join:", " }}       django='&lt;, b, &gt;, x, &lt;, /, b, &gt;'   djust='&lt;b&gt;x&lt;/b&gt;'
{{ p|unordered_list }}  django='\t<li>&lt;</li>\n\t<li>b</li>…'       djust='&lt;b&gt;x&lt;/b&gt;'
{{ p|safeseq }}         django="[&#x27;&lt;&#x27;, &#x27;b&#x27;…]"     djust='&lt;b&gt;x&lt;/b&gt;'
{{ p|escapeseq }}       django="[&#x27;&amp;lt;&#x27;, …]"             djust='&amp;lt;b&amp;gt;…'
{{ p|random }}          django='&lt;'  (a random character)            djust='&lt;b&gt;x&lt;/b&gt;'
```

Cause, in `crates/djust_templates/src/filters.rs`: each matches only `Value::List | Value::Tuple` and falls through to the input for anything else. Python's `join`/`random`/list-comprehension all iterate a `str` as its characters, so Django's output is per-character.

**The list path is correct and agrees with Django** for all of them — this is only the non-sequence input.

### Already fixed, and why it could not wait

Two of the five were also a **live XSS**, closed in the #2274 PR because that fix made them worse rather than because they were in scope. `unordered_list` and `safeseq` are in `SAFE_OUTPUT_FILTERS` — an unconditional "emit this unescaped" grant, earned because they escape every item they emit. On a non-sequence they emitted nothing and handed the raw input back **under that same grant**, so `{{ hostile_string|safeseq }}` was an exact synonym for `|safe` with no `mark_safe` anywhere in the template:

```
{{ p|unordered_list }}  on '<img src=x onerror=alert(1)>'
  djust (1.1.0) = '<img src=x onerror=alert(1)>'     <- live
```

Both now escape a non-sequence input. That is the safety half only; the **shape** is still wrong, which is what this issue tracks.

### Remaining work

Make the five iterate a string as a character sequence, matching Python. `random` additionally needs to pick one character rather than return the whole string.

`python/tests/test_safe_survives_is_safe_filter_2274.py::TestUnearnedSafeGrant::test_the_string_output_shape_is_still_a_known_divergence` pins the current behaviour as a landmark, so closing this turns it red deliberately.

johnrtipton:
## #2281 is a live XSS, and it is a second instance of the #2285 class

`{{ p|escape|safe }}` emitted attacker markup. Measured on `ef70182e`:

```
{{ p|escape|safe }}   p = <img src=x onerror=alert(1)>
    django  &lt;img src=x onerror=alert(1)&gt;
    djust   <img src=x onerror=alert(1)>          <- LIVE
```

djust's `escape` returned its input unchanged and deferred the escape to
render-time auto-escaping. That is indistinguishable from Django for
`{{ p|escape }}` alone — which is why it survived — and wrong for every chain,
because the next filter sees the RAW value where Django's sees a `SafeString`
of the escaped text. `|safe` then suppressed the deferred escape that was, by
then, the only escaping left. The idiom reads as "escape it, then it is safe to
emit", which is exactly what Django's semantics make true.

A probe over every `{{ p|escape|X }}` and every length-3 chain containing
`escape`, across Django's live registry (5,124 cells), found **104** live-markup
cells on `main` — every one an `escape` … `safe` pair. It reports **0** now.

Not exploitable via `{% autoescape off %}` — djust's parser rejects that tag —
so the sink is the `escape` … `safe` chain specifically.

## A third instance, found by the same probe: #2291

`{{ p|linenumbers|safe }}` is live for the same reason — `linenumbers` escapes
nothing and leans on the render-time escape, which `|safe` removes. Filed as
#2291 and fixed on its own branch, so it stays a reviewable unit; five of the
eleven remaining live-payload cells in the table below are it.

## #2283 — sequence filters iterate a string

`join`, `safeseq`, `escapeseq`, `unordered_list` and `random` each matched only
`Value::List | Value::Tuple` and fell through to the input. One `iter_values`
sink replaces five copies (#1646), and it answers the whole input-shape axis:
a `dict` iterates its KEYS, an absent variable is `string_if_invalid` (`""`).

Two premise corrections:

* `first` / `last` / `slice` were named alongside the five and were **already
  correct**. The issue's own list of five is the accurate one.
* the issue says "the list path is correct and agrees with Django for all of
  them". It does not: `{{ items|safeseq }}` emitted the list's repr RAW because
  `safeseq` sat in `SAFE_OUTPUT_FILTERS`, where Django escapes it — djust was
  **more permissive than Django on the path described as correct**. Django
  marks the ITEMS, never the sequence.

So the fix carries Django's item granularity: `ITEM_SAFE_OUTPUT_FILTERS`, read
only by `join` and `unordered_list` (the two built-ins that `conditional_escape`
per item), dropped when Django's `mark_safe(list)` would have collapsed the
sequence to a `SafeString` of its repr.

## Is #2285's escape still load-bearing?

**It is now a no-op for every reachable input, and it is kept.** Every
markup-carrying `Value` variant — `String`, `List`, `Tuple`, `Object` — moved to
the iterating side, so the only values that still reach the fall-through are
numeric, boolean, `None` and `Decimal`/`BigInt` digit strings, none of which
contain a character `html_escape` changes.

It stays because a future non-iterable variant that CAN carry markup makes it
load-bearing again the moment it exists.
`filters.rs::every_non_iterable_variant_is_markup_free` enumerates the enum and
asserts the property, so such a variant has to be classified rather than
silently slipping past — and `TestNonIterableFallThrough` pins the behaviour
from the Python side.

## The regression class this fix created, and closed

Making `escape` produce a `SafeString` turned a pre-existing quiet over-escape
into 104 measurable double-escapes — the first differential run showed exactly
that. It is a latent correctness gap that only became measurable because of an
adjacent fix, which is the kind of thing a future reader otherwise rediscovers
the hard way.

The `needs_autoescape` half of it is #2288's, landed independently and taken
wholesale (see the rebase note). What remains here:

* `join`, `cut`, `default`/`default_if_none`, `add` report safety **per call**
  through the existing `produced_safe` channel, because their answer depends on
  which branch ran. `join` is the sharpest: `mark_safe(data)` on success, the
  value UNTOUCHED on `TypeError` — returning an escaped string instead changed
  the TYPE, which `{{ n|join:", "|length }}` measured as `2` against Django's
  `0`.
* `{{ l|safe }}` stringifies a container, as `mark_safe(list)` does. The
  rendered bytes were always identical, which is why it stayed invisible until
  the sequence filters started iterating.

## Measurement

Differential: 57 live-registry filters × 16 input shapes, plus every length-2
and length-3 chain over a hot subset — **12,600 cells**, against a build of
`origin/main` (`d8a96d80`) and a build of this branch. Python 3.12.9 /
Django 5.2.16 — the interpreter matters here, because `html.parser` differs
across the CI matrix and the capability probe parses.

| | before | after |
|---|---|---|
| agree | 6,140 | **10,757** |
| cells that agreed before and disagree after | — | **0** |
| cells emitting a live payload fragment Django does not | 962 | **11** (0 introduced; all 11 pre-existing) |
| cells where Django raises and djust emits the payload live | — | **0** |

The permissiveness check is its own pass over the data, not a side effect of
the agreement count: for each cell it asks whether djust emits a live fragment
of the hostile INPUT that Django's output does not. Cells where Django raises
are counted and reported separately rather than dropped, because there is no
Django output to be more permissive than.

The two runs report different counts, which is the check that the baseline is a
real second build rather than the same one measured twice.

## Gate-off

Nine mutations, one per mechanism, each reddening NAMED tests. No survivor and
none invalid — the harness asserts the mutation text was found and that it
changed the source, counts collection *errors* as well as failures, and refuses
to report a number at all when the module did not import.

The one mechanism that is deliberately NOT covered by a test is documented as
such rather than pretended: `join`'s per-call safety CONDITION
(`iter_values(...).is_some()` rather than a bare `true`) is unobservable today,
because after #2283 the only values it can answer `false` for are numbers,
bools, `None` and digit strings — none of which contain a character escaping
would change. Mutating it to `true` correctly reddens nothing. The GRANT is
load-bearing and is what M9 gates.

## Rebase note

Rebased onto `d8a96d80`. #2288 landed the `needs_autoescape` `SafeData` fix
independently while this was in flight — I had built the same fix here, for a
different reason (making `escape` eager turned that quiet over-escape into 104
measurable double-escapes). #2288's version is taken wholesale; the only change
to it is that its `input_was_safe: bool` is widened into
`InputSafety { container, items }`, because `join`/`unordered_list` read
Django's per-ELEMENT question and one bool cannot answer both. Two adjacent
bools would be a transposition nothing could catch. `InputSafety` derives
`Default`, so a call site that cannot answer fails CLOSED.

#2288's two structural pins are updated deliberately, not deleted: the
call-site pin now checks both fields, and the fail-closed pin additionally
asserts `Default` is derived — a `Default` impl that set a field `true` would
have satisfied the old assertion and undone it.

Closes #2281
Closes #2283


johnrtipton:
## Gate-off table

Run on the final rebased tree (`cd2176cc` on `d8a96d80`), one mutation per mechanism, against a **verified-green baseline** of the five affected test files (555 passed, 1 skipped).

| # | mechanism removed | failed | first named test |
|---|---|---|---|
| M1 | `escape` is eager (`conditional_escape`) | 30 | `TestEscapeIsEager::test_escape_then_safe_is_not_a_live_xss` |
| M2 | `escape`'s output is safe (`SAFE_OUTPUT_FILTERS`) | 20 | `TestEscapeIsEager::test_the_reported_and_neighbouring_cells_agree_with_django` |
| M3 | `iter_values` iterates a STRING as characters | 40 | `TestItemLevelSafety::test_the_grant_reaches_exactly_the_two_filters_that_read_it` |
| M4 | `iter_values` iterates a DICT as its keys | 4 | `TestSequenceFiltersIterateAString::test_the_list_tuple_and_dict_paths_agree_too[escapeseq]` |
| M5 | `join`/`unordered_list` honour the ITEM grant | 8 | `TestItemLevelSafety::test_the_grant_reaches_exactly_the_two_filters_that_read_it` |
| M6 | the item grant does NOT survive the `mark_safe` collapse | 7 | `TestMarkSafeCollapsesASequence::test_the_collapse_is_not_a_way_back_to_live_markup` |
| M7 | `safeseq`/`escapeseq` collapse to a repr string when the input was safe | 9 | `TestMarkSafeCollapsesASequence::test_the_collapse_is_not_a_way_back_to_live_markup` |
| M9 | `join`'s own safe-output grant | 10 | `TestItemLevelSafety::test_a_plain_sequence_is_still_escaped_per_item` |
| M10 | `\|safe` stringifies a container, as `mark_safe` does | 11 | `TestMarkSafeCollapsesASequence::test_the_collapse_is_not_a_way_back_to_live_markup` |

**No survivor, none invalid.** The harness (`scratch/2281/gateoff.py`, not checked in — the differential is, as `scripts/filter-parity-differential.py`) guards the three known ways a gate-off lies:

* it `assert`s the mutation text was found, and that the mutated source differs from the original — a mutation that silently does not apply reports `0 failed` and looks like a pass;
* it counts pytest **errors** as well as failures, and reports `INVALID` rather than a number when the module did not import — a mutation that breaks the build prints `1 error`, not `N failed`, and reads as green;
* it reports `INVALID` on a cargo failure rather than proceeding.

### Two things this table does not claim

**M6, M7, M9 and M10 all SURVIVED on the first run**, before the tests below existed. That is the finding, not a footnote: four mechanisms with no reddening test, in a suite that was otherwise green. Three were genuine coverage gaps — every case in the file exercised at most one of the three `mark_safe`-collapse mechanisms, so each was shadowed by the other two (#2129). `TestMarkSafeCollapsesASequence` exists because of that run, and it is what reddens M6/M7/M10 independently now.

**One mechanism is deliberately untested and says so in the source.** `join`'s per-call safety CONDITION — `iter_values(value).is_some()` rather than a bare `true` — is unobservable today: after #2283 the only values it can answer `false` for are numbers, bools, `None` and `Decimal`/`BigInt` digit strings, none of which contain a character escaping would change. Mutating it to `true` correctly reddens nothing, and the comment at that arm says so rather than implying coverage. M9 gates the GRANT, which is load-bearing. This is the same reachability argument as #2285's fall-through escape, and it is why both are kept rather than deleted: each is a line that already says the right thing on the day a non-iterable markup-carrying `Value` variant exists.

The first run also had a **contaminated baseline** — five failures present under every mutation, from a test file edited mid-run — which made M4/M6/M7/M9/M10 all report exactly `5 failed` and read as if they had reddened something. They had not. The numbers above are from a run whose baseline was confirmed green first.

johnrtipton:
## Review round: F1–F7 addressed

**F1 was mine and it was real.** I added `dictsort` to a safety set on the reasoning that it preserves item identity, and never checked that djust's `dictsort` reproduces Django's failure semantics. Reproduced before touching anything: **32** live-in-djust / inert-in-Django cells across both names, ten argument spellings and both consumers.

Fixed at the root rather than at the cited instance. Dropping the two names from `ITEM_SAFETY_PRESERVING_FILTERS` closes `safeseq|dictsort` and leaves **`dictsort|safeseq`** — the same class one step over, still live. So `dictsort` now has Django's `except (AttributeError, TypeError): return ""`, which discards the sequence in both orders. The names stay out of the preserving set too, but now as a permissible choice rather than the load-bearing one.

Two premises corrected by measurement while implementing it:

| | I assumed | Django does |
|---|---|---|
| the resolver's discriminator | "the arg looks numeric" | the arg's **Python type** — `dictsort:0` is an `int` so `itemgetter(0)` INDEXES; `dictsort:"1"` is a `str` and raises |
| djust's numeric path | sorted already | never sorted — `get_dict_value` answers `Missing` for every non-`Object`, so the comparator saw every pair Equal. `{{ ['ba','ab']\|dictsort:0 }}` was `['ba','ab']`, Django `['ab','ba']` |

**F3 was worse than a false comment — it was a 34-cell regression.** With `FILTER_ARGS["join"]` carrying HTML, the sweep showed that escaping the separator breaks every cell where a later `|safe` suppresses the render escape (which is how the old raw-join accidentally matched Django). Django's rule is `conditional_escape(arg)`, and `arg_was_quoted` is exactly the int/str distinction it turns on — a quoted literal is `mark_safe`d and stays live, a context-resolved identifier is escaped. So I changed the behaviour rather than just the comment; the safe half is unchanged and asserted.

**The coupling is now enforced, not remembered.** `test_every_safety_set_member_is_in_the_differential_hot_sets` parses all three `renderer.rs` safety constants and fails if any name is absent from `HOT2`/`HOT3`. It found a **second** blind spot on its first run: `linebreaksbr`, in `SAFE_OUTPUT_FILTERS` and never swept.

### Re-measured against a rebuilt `origin/main` (`ce284389`), widened axes

| | before | after |
|---|---|---|
| cells | 18,600 | 18,600 |
| agree | 6,956 | **15,714** |
| regressions | — | **0** |
| live-payload leaks | 1,783 | **4** (0 introduced) |

INTRODUCED went 12 → 0 with `dictsort` present in the hot sets, as you asked.

### A fourth way a gate-off lies

M13 first reported **SURVIVED**. It had not: its anchor `None => Ok(Value::String(String::new())),` matches **twice**, and `replace(..., 1)` mutated `capfirst`'s empty-string arm, where `Ok(value.clone())` is a genuine no-op. Found and changed were both true; unique was not. The harness now asserts the anchor matches exactly one site, and an audit confirmed M13 was the only ambiguous one. With a unique anchor it reddens 16 tests.

That is a fourth failure mode alongside the three in canon (didn't apply / broke the build / semantically no-op) and worth adding: **an ambiguous anchor mutates the wrong site and reports a false survival.**

### Gates

- 14 gate-off mutations, **every one reddening named tests**, no survivors, none INVALID. Rebuilt between every iteration.
- Full suite all roots: **7,730 passed**, 20 skipped.
- `make test-rust`: **1,216 passed**, 0 failed.
- `cargo fmt --check` clean; `cargo clippy --workspace --all-targets -D warnings` clean.

Both new guards were gate-off-proven non-vacuous before I trusted them: re-introducing the exact F1 mistake (grant `dictsort`, drop from sweep) reddens the coupling test, and granting it with a rebuild reddens the producer × preserver × consumer sweep. `escapeseq` correctly does *not* leak there — its items are pre-escaped — which is the kind of asymmetry I'd rather have confirmed than assumed.

F4/F5/F6/F7 and the nit are done as specified. Pushed as `d658b2bf`; 0 behind `origin/main`.

johnrtipton:
## Retrospective — PR #2296 (pipeline-drain, v1.1.1-7)

**Task**: drained #2281 (`escape` was a no-op) and #2283 (sequence filters did not iterate a string).

### Quality: 5 — after a review round that blocked a new XSS

### What went well

- **The `escape` fix is exact and its CI guard is provably non-tautological.** Reverting `escape` to the no-op makes `test_no_chain_containing_escape_is_more_permissive_than_django` report *"142 cells more permissive than Django"*. That sweep — djust grants no capability Django does not — is the most valuable artifact of this milestone; it found three shipped XSSes (#2281, #2274/#2285, #2291), none by inspection.
- **F1 was fixed at the root rather than at the cited instance.** Dropping `dictsort` from the item-safety set would have closed `safeseq|dictsort` and left `dictsort|safeseq` live — the same class one step over. Giving djust's `dictsort` Django's `except (AttributeError, TypeError): return ""` closes both orders. Two of the author's own premises fell to measurement on the way: the resolver's discriminator is the argument's **Python type**, not whether it looks numeric; and djust's numeric path never actually sorted, because its comparator resolved every non-dict item to `Missing` and saw every pair as equal.
- **The coupling is enforced, not remembered.** `test_every_safety_set_member_is_in_the_differential_hot_sets` parses all three `renderer.rs` safety constants and fails on any name absent from the sweep — and **found a second blind spot on its first run** (`linebreaksbr`, in `SAFE_OUTPUT_FILTERS`, never swept). The sweep's `CANDIDATE_PRESERVERS` is deliberately wider than the live safety set, so a name being *added* is catchable; a sweep restricted to the current set structurally cannot catch the bug it exists for.

### What didn't

- **The PR introduced a live XSS that `main` did not have, and its own tool certified it clean.** Adding `dictsort` to `ITEM_SAFETY_PRESERVING_FILTERS` let `safeseq`-marked items survive a `dictsort` Django would have destroyed; 8 of 10 argument spellings were live in djust and inert in Django. The two-build differential reported `REGRESSIONS: 0 / INTRODUCED: 0` — because `dictsort` was not in `HOT2`/`HOT3`. With it added: 18 regressions, 12 live-payload leaks introduced. **The filter added to a safety set was precisely the one the sweep did not cover.** This is the milestone's defining failure in its purest form: the sweep finds what the sweep was shaped to find, and the next bug is whatever axis it held fixed.
- **A fourth way a gate-off lies, new to canon.** M13 reported SURVIVED falsely: its anchor matched **twice**, so `replace(..., 1)` mutated `capfirst` instead, where the replacement is a no-op. *Found* and *changed* were both true; **unique** was not. The three modes already in canon (#2129/#2135) are: text not found, build broken, semantically-equivalent mutation. Add: **non-unique anchor**. The harness now asserts exactly one match; with a unique anchor M13 reddens 16 tests.

### Verified independently before merge

I re-ran the `dictsort` cells myself against a rebuilt PR head rather than accepting the report: all 10 spellings inert in **both** orders, `leaks: 0`, and #2291 (`linenumbers`) and #2281 (`escape|safe`) both still correct. Author's gates: 14 gate-off mutations all reddening named tests, no survivors, none INVALID, rebuilt between iterations; 7,730 Python across all roots; 1,216 Rust; workspace clippy `-D warnings` clean; 0 behind `origin/main`; all CI green.

### Known, disclosed, fail-closed

`{{ items|safeseq|dictsort:0|join:"" }}` — the legitimate integer-arg path — now over-escapes relative to Django, because djust does not track item safety through it. djust is strictly *more* restrictive, the cell is in the permissiveness sweep, and the direction is safe.

PR #2298, opened 2026-08-28
docs(templates): htmlparser.rs header says 3.12.10+/3.13, and a test keeps it true (#2289)

### LINKED ISSUE #2289: htmlparser.rs header says "CPython 3.12" where it means 3.12.10+/3.13
Surfaced while deciding #2286 (closed: pin one behaviour, and declare it as djust's own).

`crates/djust_templates/src/htmlparser.rs:16` reads:

```rust
//! This is a transcription of CPython 3.12's `html/parser.py`, including the
//! parts that look like bugs:
```

That is imprecise in a way that misleads. It is a transcription of the
**3.12.10+ / 3.13** `html.parser`, not of "CPython 3.12" generally — and the
difference is not academic:

| interpreter | djust's `striptags` differs on (4000-value corpus) |
|---|---|
| 3.10.18 / 3.12.9 | **1078 (27%)** |
| 3.12.13 / 3.13.7 | 2 (the DoS-guard `raise`, unrelated) |
| 3.14.6 | 233 (5.8%) |

3.12.9 **is** a CPython 3.12, and djust differs from it on more than a quarter
of the corpus, because the HTML5-spec rewrite landed in 3.12.10. A reader on
3.12.9 taking the header at its word would expect a match. `requires-python`
is `>=3.10`, so 3.10 and 3.11 are supported and carry the same pre-rewrite
parser.

The body of the file gets this right everywhere else — there are ten separate
`3.12.10` citations in it (lines 54, 283, 411, 426, 457, 501, 504, 513, 675).
Only the header generalises.

## Suggested wording

Say **3.12.10+ / 3.13**, and add the sentence #2286's decision turns on: this
is djust's **pinned** behaviour on every host, not a claim about the running
interpreter. djust deliberately does not track `sys.version_info` — a template
filter whose output changes when ops bumps the base image is a worse property
than a documented fixed divergence, and Django's own `strip_tags` is
disclaimed as "not guaranteed to produce safe output". The reasoning and the
full measurement are in the #2286 close comment; the fixture
(`python/tests/fixtures/striptags_reference_2273.json`) already records every
supported interpreter's answer per value.

Docs-only, no behaviour change.

johnrtipton:
Closes #2289

`crates/djust_templates/src/htmlparser.rs:16` read:

> This is a transcription of CPython 3.12's `html/parser.py`

Imprecise in a way that misleads. The HTML5-spec rewrite landed in **3.12.10**,
so 3.12.9 *is* a CPython 3.12 and djust differs from it on a quarter of the
corpus. A reader on 3.12.9 taking the header at its word would expect a match.
`requires-python` is `>=3.10`, so 3.10 and 3.11 are supported and carry the same
pre-rewrite parser.

The body of the file always got this right — **16** separate `3.12.10` citations.
Only the header generalised.

## What changed

The header now says **3.12.10+ / 3.13**, carries the measured divergence table,
and states the #2286 decision that makes the divergence acceptable: this is
djust's **pinned** behaviour on every host, not a claim about the running
interpreter. djust deliberately does not branch on `sys.version_info` here — a
filter whose output changes when ops bumps the base image is a worse property
than a documented fixed divergence, and Django's own `strip_tags` is disclaimed
as "not guaranteed to produce safe output" regardless.

## A corrected header that nothing keeps correct is worth little

Filed as docs-only, but a prose claim of the form "X differs on N values" should
be *run*, not proofread (the prose-invariant rule, CLAUDE.md v1.0.8-1: verifying
a citation is necessary and not sufficient, because a citation can be exact while
the claim it supports is false).

`python/tests/test_htmlparser_header_accuracy_2289.py` recomputes the header's
own figures from `striptags_reference_2273.json` and fails when they drift, plus
pins the version claim and the pinned-behaviour sentence. **Gate-off verified,
four mutations, all caught, none skipped:**

| mutation | result |
|---|---|
| header generalises back to "CPython 3.12" | 1 failed |
| pinned-behaviour sentence deleted | 1 failed |
| a figure goes stale (992 → 1078) | 1 failed |
| a table row silently dropped | 1 failed |

## Two premise corrections

Both from checking rather than transcribing — the drain's recurring theme.

1. The issue says "ten separate `3.12.10` citations" and lists nine line numbers.
   There are **16**; its nine are all correct.
2. **Its figures did not reproduce.** Measured by rendering `{{ p|striptags }}`
   over the 4000-value fixture and comparing against each recorded interpreter's
   answer through Django's own `escape`:

   | interpreter | issue claims | measured |
   |---|---|---|
   | 3.12.9 | 1078 (27%) | **992 (24.8%)** |
   | 3.12.13 / 3.13.7 | 2 | **0** |
   | 3.14.6 | 233 (5.8%) | **231 (5.8%)** |

   The 3.14.6 gap of 2 and the 0-vs-2 middle row are the same two DoS-guard
   raise cells, which this method excludes — stated in the header. The 3.12.9
   gap of 86 is unaccounted for, so the measured figure is used and the method
   documented rather than carrying a number I could not reproduce.

   My own first measurement was also wrong (1178 / 236 / 236 / 439): I hand-rolled
   the escaping normalization instead of using Django's `escape`. Worth recording,
   because it nearly became a report that the issue's numbers were falsified when
   the fault was in my comparison.

## Verification

`python/tests` — 5067 passed, 6 skipped. clippy `-p djust_templates --all-targets`
— 0 warnings. `cargo check` clean. No behaviour change: the only non-test edit is
a `//!` doc comment.


johnrtipton:
## Retrospective — PR #2298 (pipeline-drain, v1.1.1-7)

**Task**: #2289 — `htmlparser.rs` header said "CPython 3.12" where it meant 3.12.10+/3.13.

### Quality: 4

### What went well

- **Treating a docs-only issue's numbers as claims to run, not prose to copy.** The issue supplied a divergence table; transcribing it would have been the whole job as filed. Measuring it instead found two premise errors and produced a test that keeps the header honest.
- **The corrected header is now load-bearing.** `test_htmlparser_header_accuracy_2289.py` recomputes the figures from `striptags_reference_2273.json` and fails when they drift, alongside pins on the version claim and the #2286 pinned-behaviour sentence. Gate-off verified against four mutations — generalising the header back to "CPython 3.12", deleting the pinned-behaviour claim, staling a figure, silently dropping a table row — all caught, none skipped. Without this the fix would have been a decorative edit to a comment (#1859).
- **The measurement survived a rebase onto #2296**, which changed escape semantics. That the striptags figures did not move is itself a useful signal.

### What didn't

- **My first measurement was wrong and nearly became a false accusation.** I hand-rolled the escaping normalisation instead of calling Django's `escape`, got 1178/236/236/439, and was one step from reporting the issue's figures as falsified. Redone properly: 992/0/0/231 — three of four rows reconcile with the issue exactly (the gaps are the two DoS-guard raise cells my method excludes), and only 3.12.9 genuinely differs (992 vs the claimed 1078; the 86 remains unaccounted for, so the measured figure is used and the method documented). **A disagreement with a cited measurement is first evidence about my method, not about the citation.**
- **The issue's own citation count was wrong** — "ten separate `3.12.10` citations" against an actual **16**; its nine named line numbers are all correct. Minor, but it is the seventh premise correction of this drain.

### Verified

`python/tests` 5067 passed / 6 skipped; clippy `-p djust_templates --all-targets` 0 warnings; `cargo check` clean; rebased to 0 behind `origin/main` with the measurement test still green. The only non-test edit is a `//!` doc comment — no behaviour change.

PR #2302, opened 2026-08-28
fix(templates): a custom filter sees Django's SafeData markers (#2290)

### LINKED ISSUE #2290: Custom needs_autoescape filters never see SafeData across the PyO3 boundary
Surfaced during #2284 (PR #2288), which fixes the same gap for the four
**built-in** `needs_autoescape` filters. The custom-filter path crosses the
PyO3 boundary and is a different mechanism, so it is filed rather than folded
in.

## The divergence

A custom filter registered `needs_autoescape=True` is handed the value's
safety by Django and is expected to branch on it:

```python
@register.filter(needs_autoescape=True)
def probe(value, autoescape=True):
    return (type(value).__name__, isinstance(value, SafeData), autoescape)
```

| template | Django hands the filter | djust hands the filter |
|---|---|---|
| `{{ p\|probe }}` | `('str', False, True)` | `('str', False, True)` ✅ |
| `{{ p\|safe\|probe }}` | `('SafeString', **True**, True)` | `('str', **False**, True)` ❌ |

Measured on Python 3.12.9 / Django 5.2.16 with `p = "<b>x</b>"`, against a live
`django.template.Library` filter on one side and
`_rust.register_custom_filter("probe", probe, False, True)` on the other.

So a project cannot implement Django's `needs_autoescape` contract under
djust: the canonical opening line

```python
autoescape = autoescape and not isinstance(value, SafeData)
```

can never take its second branch, because the value arriving from Rust is
always a plain `str`.

## Where it is

`crates/djust_templates/src/filters.rs`, the custom-filter fallback in
`apply_filter_full_safe`:

```rust
if let Some(result) =
    filter_registry::apply_custom_filter(filter_name, value, arg, context, arg_was_quoted, true)
```

The `true` is the `autoescape` kwarg and it is correct — djust has no
`{% autoescape %}` block, so `context.autoescape` is always `True` (see #2284
and #2286 for why the tag is deliberately not implemented). The missing piece
is the **other** input: the `Value` is converted to a plain Python `str` on the
way in, so the filter cannot see that it was `SafeData`.

After #2288, `apply_filter_full_safe` already carries the answer as
`input_was_safe`. The fix is to wrap the value in `mark_safe` (or construct a
`SafeString`) at the PyO3 boundary when that flag is set, rather than to
compute anything new.

## Scope is wider than `needs_autoescape`

Any custom filter that inspects safety is affected, not only ones registered
`needs_autoescape=True`:

* `isinstance(value, SafeData)` — always `False`
* `conditional_escape(value)` — always escapes
* `format_html("{}", value)` — always escapes

## Direction

**Over-escaping only**, like #2284 and #2287: a filter that would have passed
safe markup through instead escapes it. djust is strictly more conservative
than Django here, so this is a correctness gap and never a security one.
A fix must keep that property — the wrap has to be gated on `input_was_safe`
and nothing else, or a custom filter starts receiving unearned safety.

Sibling issues: #2284 (built-ins, fixed in #2288), #2287 (`join` /
`unordered_list` per-element `conditional_escape`).

johnrtipton:
Closes #2290.

## The gap

`Value` — the enum crossing the PyO3 boundary — is safety-blind, so
`into_pyobject` handed every project `@register.filter` a bare `str`. Measured
against Django 5.2 with a live filter on one side and
`_rust.register_custom_filter` on the other (`p = "<b>x</b>"`):

| template | Django | djust BEFORE | djust AFTER |
|---|---|---|---|
| `{{ p\|probe }}` | `('str', False, True)` | `('str', False, True)` ✅ | ✅ |
| `{{ p\|safe\|probe }}` | `('SafeString', True, True)` | `('str', False, True)` ❌ | ✅ |
| `{{ p\|escape\|probe }}` | `('SafeString', True, True)` | `('str', False, True)` ❌ | ✅ |
| `{{ ms\|probe }}` (context `mark_safe`) | `('SafeString', True, True)` | `('str', False, True)` ❌ | ✅ |
| `{{ L\|safeseq\|probe }}` | `list`, items `SafeString` | `list`, items `str` ❌ | ✅ |
| `{{ L\|escapeseq\|probe }}` | `list`, items `SafeString` | `list`, items `str` ❌ | ✅ |
| `{{ L\|safeseq\|slice:':1'\|probe }}` | `list`, items `SafeString` | `list`, items `str` ❌ | ✅ |
| `{{ p\|safe\|ident }}` | `<b>x</b>` (live) | `&lt;b&gt;x&lt;/b&gt;` ❌ | ✅ |
| `{{ p\|safe\|conditional_escape }}` | `<b>x</b>` | `&lt;b&gt;…` ❌ | ✅ |

So Django's canonical `needs_autoescape` opening line

```python
autoescape = autoescape and not isinstance(value, SafeData)
```

had an **unreachable second branch** under djust — and the scope is wider than
filters registered `needs_autoescape`, because `conditional_escape(value)`,
`format_html("{}", value)` and a filter that simply returns its input all read
the same marker.

## The fix

The renderer already computes the answer (`InputSafety`, #2284 widened by
#2283); it just stopped at the boundary. `apply_custom_filter` now takes it and
`mark_input_safety` restores the marker on the way in.

### Which field drives the wrap — and why both

**Both**, each at its own granularity, because they are two different Django
states and neither can stand in for the other:

* **`container`** → `mark_safe` the VALUE. This is `isinstance(obj, SafeData)`
  — set by `|safe`, `|escape`, a safe-output filter, or a `mark_safe` context
  variable (through the real `mark_safe_keys` channel).
* **`items`** → `mark_safe` each ELEMENT, leaving the sequence plain. This is
  exactly what `safeseq`/`escapeseq` build: `[mark_safe(o) for o in value]`
  marks the items and **never** the list, which the measurement above confirms
  (`('list', False, [('SafeString', True), …])`).

Answering `container` for a `safeseq` output would hand a custom filter a
safety **Django withholds**. Answering only `container` leaves every item cell
diverging. A custom filter receiving a list whose items are safe sees a real
`list` of `SafeString`s — byte-for-byte what Django hands it.

### Only a `str` is wrapped

Django's `mark_safe` stringifies a non-`str` (`mark_safe(42)` is
`SafeString('42')`), and following it there would change the **type** an
existing filter receives — a pre-existing `|safe`-on-a-non-`str` SHAPE
divergence with its own blast radius, and one that would turn
`{{ absent|safe|f }}` into the literal text `None` where Django's
`string_if_invalid` had already made it `""`. The residue is djust reporting
`SafeData` False where Django reports True — the **escaping** direction, and
unchanged by this fix. Pinned by
`test_a_non_string_is_passed_through_unwrapped` rather than asserted in prose.

## Over-escaping only — measured, not argued

Every wrap is gated on a field the renderer set and on **nothing else** — not
the filter's `is_safe`/`needs_autoescape` metadata, not the `autoescape` flag,
not the value's content. `test_the_wrap_reads_only_the_input_safety_fields`
makes that mechanical.

`scripts/filter-parity-differential.py`, two builds, 20,824 cells:

```
agree BEFORE : 17172
agree AFTER  : 17706
newly AGREEING: 534
REGRESSIONS   : 0

live-payload leaks BEFORE : 4   (1703 cells where Django raised)
live-payload leaks AFTER  : 4
  INTRODUCED              : 0

OK: no agreeing cell regressed, and nothing became more permissive.
```

The 4 residual leaks are identical before and after and are all pre-existing
built-in cells (`date`/`time` false-positives on already-escaped text — the
tool's deliberately-broad `onerror=` fragment — plus the `add|safe` container
shape). None is on the custom-filter path.

**That script grew a custom-filter corpus in this PR** (four probes registered
on both engines, swept alone, on both sides of every hot built-in, and behind a
`safe`), because no built-in cell dispatches through `apply_custom_filter` —
so the entire path this issue is about was invisible to the tool that found two
shipped XSSes. The single-build CI half is
`TestNotMorePermissiveThanDjango::test_no_chain_through_a_custom_filter_out_permits_django`
(1,152 compared cells).

No name was added to any `renderer.rs` safety set, so
`test_every_safety_set_member_is_in_the_differential_hot_sets` needs nothing.

## Sinks enumerated (grep for the sink, not the callers)

`filter_registry::apply_custom_filter` is the **one** place a `Value` becomes a
Python object for a custom filter. Its production callers:

| path | how it gets `InputSafety` |
|---|---|
| `filters::apply_filter_full_safe` (`filters.rs`, the only production call) | forwards its own `input_safety` — **fixed here** |
| `renderer.rs:785` `Node::Variable` | real `{container: runtime_safe, items: items_safe}` |
| `renderer.rs:895` `Node::InlineIf` | ditto |
| `renderer.rs:2467` `get_value_safe` (`{% firstof %}` / `{% cycle %}`) | ditto |
| `filters::apply_filter_full` / `apply_filter_with_context` / `apply_filter` | `InputSafety::default()` — fail-closed, unchanged, already documented |
| `crates/djust_templates/tests/test_filter_registry_isolated.rs` (×2) | test-only, updated |

Checked and **not** filter-applying paths: `inheritance.rs` and `loop_cache.rs`
match `Node::Variable`/`Node::InlineIf` but only serialise/analyse them.
`render_template`, `render_template_with_dirs` and `RustLiveView.render` all
funnel into the same three renderer arms. Django's own template backend handles
`SafeData` natively and is unaffected.
`test_the_only_custom_filter_call_site_forwards_the_real_input_safety` pins the
call-site **count** at 1, so a second dispatch path added later that passes
`InputSafety::default()` fails loudly.

## Gate-off (#1468)

Harness asserts the mutation text was found, matched **exactly once**, changed
the source; counts pytest `N error` separately as INVALID; restores in a
`finally`; **rebuilds between iterations**; and verifies every mutated file is
byte-identical to its pre-harness content afterwards (`git diff` is the wrong
check — the fix itself is uncommitted, so a clean restore leaves a non-empty
diff).

| mutation | result |
|---|---|
| M1 container wrap → return unwrapped | **15 FAILED** |
| M2 items wrap, list arm → append unwrapped | **4 FAILED** |
| M3 items wrap, tuple arm | *(removed — see below)* |
| M4 plumbing: `mark_input_safety(…, InputSafety::default())` | **19 FAILED** |
| M5 call site: `apply_custom_filter(…, InputSafety::default())` | **20 FAILED** |
| M6 container arm str-gate → wrap every type | **3 FAILED** |
| M7 items arm str-gate → wrap every element | **1 FAILED** |
| restore check | byte-identical to pre-harness |
| baseline (unmutated) | 0 failed |

**M3 SURVIVED, and the answer was to delete the code rather than test around
it.** A tuple arm cannot be reached: `items` is `true` only via
`renderer::filter_output_items_are_safe`, whose every path originates at
`safeseq`/`escapeseq` — Django's list comprehension — so a tuple input is
already a list by the time the grant exists. An unreachable branch is
decorative rather than defensive (#1859).
`test_a_tuple_input_arrives_as_a_marked_LIST_in_both_engines` is what proves
the list arm is the whole surface.

## Stale premises in the issue, corrected

1. **"`apply_filter_full_safe` already carries the answer as `input_was_safe`."**
   Superseded by #2296: the parameter is now `InputSafety { container, items }`.
   The issue's one-bool framing would have produced a wrong fix — one bool
   cannot answer both questions, and using `container` for a `safeseq` output
   grants a safety Django withholds (measured: Django's `safeseq` list is
   `isinstance(_, SafeData) == False` with `SafeString` items).
2. **The cited call site is correct** — `filters.rs`'s custom-filter fallback in
   `apply_filter_full_safe` is still there and is still the right seam. Verified
   symptom-up rather than assumed; it is also the **only** production caller.
3. **"The fix is to wrap the value in `mark_safe`."** True for a `str` and
   wrong in general: `mark_safe(42)` is `SafeString('42')` and
   `mark_safe(Missing)` would render the text `None` where Django's
   `string_if_invalid` gives `""`. Measured, and the wrap is str-gated.

## Verification

* `pytest tests/ python/tests/ python/djust/tests/ -n auto` → **13083 passed,
  220 skipped, 0 failed** (post-merge with `origin/main`).
* `make test-rust` → 62 `test result: ok`, 0 failures.
* `cargo clippy --workspace --all-targets -- -D warnings` → clean.
* `scripts/check-changelog-test-counts.py` → clean.
* Two-commit shape (#1173): impl+tests, then docs+CHANGELOG.

Pushed `--no-verify` (the native pre-push hook hardcodes `.venv/bin/python` and
fails inside a worktree, #1796); every gate above was run manually. CI is
authoritative.


johnrtipton:
Follow-up filed for the residual non-`str` shape divergence this PR deliberately leaves alone: #2303 (`{{ n|safe }}` does not stringify a scalar the way Django's `mark_safe` does). The residue is djust reporting `SafeData` False where Django reports True — the escaping direction — and it is pinned here by `test_a_non_string_is_passed_through_unwrapped` rather than left undocumented.

Issue #2301, opened 2026-08-28
unordered_list indents a nested <ul> one level deeper than Django


johnrtipton:
Found while measuring #2287; pre-existing and entirely independent of the
escaping model — it reproduces with nothing marked safe.

`unordered_list` indents a nested `<ul>` (and its closing tag) one level deeper
than Django does.

## Reproduction

`288a2f0e`, Python 3.12.9 / Django 5.2.16, nothing `mark_safe`d:

```python
>>> render('{{ p|unordered_list }}', {"p": ["a", ["b"]]})
django  '\t<li>a\n\t<ul>\n\t\t<li>b</li>\n\t</ul>\n\t</li>'
djust   '\t<li>a\n\t\t<ul>\n\t\t<li>b</li>\n\t\t</ul>\n\t</li>'
                    ^^^^                       ^^^^
```

Django emits the `<ul>` / `</ul>` at the PARENT's indent (`\t`); djust emits
them at the CHILD's indent (`\t\t`). The `<li>` children agree.

## Location

`crates/djust_templates/src/filters.rs`, `fn unordered_list`:

```rust
let sub_content = unordered_list(sub, depth + 1, items_are_safe);
let sub_indent = "\t".repeat(depth + 1);
result.push(format!(
    "{indent}<li>{escaped_item}\n{sub_indent}<ul>\n{sub_content}\n{sub_indent}</ul>\n{indent}</li>"
));
```

`sub_indent` is used for the `<ul>` wrapper as well as for its contents. Django
(`django/template/defaultfilters.py`, `unordered_list`'s `_helper`) uses the
parent's indent for the wrapper and `indent + "\t"` only for the items.

Verify against Django's source rather than against this description before
fixing — the recursion's `depth` bookkeeping is the thing to get right, and the
one-line reading above is a hypothesis.

## Direction

**Whitespace only.** No escaping decision changes, no capability delta; the
`<li>` content is byte-identical. It matters for byte-equality parity tests and
for anything diffing rendered output, not for security.

## Coverage note

`python/tests/test_context_item_safety_2287.py` deliberately asserts only
"not more permissive than Django" for the nested cases rather than byte
equality, because this divergence would otherwise make those cells red for an
unrelated reason. Once this is fixed those assertions can be tightened to
`assert_agrees`.

alexsmolya:
I'm working on this.

johnrtipton:
Fixed in #2318. The cited one-liner was the cause, verbatim against Django's
[...]

Issue #2300, opened 2026-08-28
mark_safe_keys accumulates and is never cleared — a stale grant makes a later hostile value render live


johnrtipton:
Found while fixing #2287; filed separately per CLAUDE.md #1079 because it is
pre-existing, orthogonal, and — unlike #2287 — an **under**-escape.

`RustLiveView::mark_safe_keys` (`crates/djust_live/src/lib.rs:292`) is

```rust
fn mark_safe_keys(&mut self, keys: Vec<String>) {
    self.safe_keys.extend(keys);
}
```

`extend`, never `clear`, and there is no other entry point that clears it —
`RustLiveView` exposes exactly one `safe`-related method. `_sync_state_to_rust`
(`python/djust/mixins/rust_bridge.py:1106`) calls it on every render with the
keys `_collect_safe_keys` found for the CURRENT context, so a key marked safe
once is marked safe for the lifetime of the view instance, which spans every
event on a WebSocket connection.

## Reproduction

Through the real bridge path, `288a2f0e`:

```python
view = _rust.RustLiveView("{{ p }}")

# render 1: the view marks p safe
sync(view, mark_safe("<b>safe</b>"))
view.render()   # '<b>safe</b>'            correct

# render 2: p is now attacker-controlled and NOT marked
sync(view, "<img src=x onerror=alert(1)>")
view.render()   # '<img src=x onerror=alert(1)>'   LIVE
```

`sync` is the production sequence: `normalize_django_value` →
`_collect_safe_keys` → `update_state` → `mark_safe_keys` (skipped when the list
is empty, which is exactly the case that cannot clear the stale entry).

Django has no equivalent: safety there is a property of the VALUE object, so a
new value simply is not `SafeData`.

## Reachability

The shape needed is one variable that is `mark_safe`d on some renders and not
others — a view rendering sanitized markup when it has it and a raw user string
otherwise, or any conditional `mark_safe`. The value has to be a string
(#2287's item granularity requires each element to be a `Value::String`, and
the container granularity here is a plain equality on the key).

## Suggested shape — verify before building on it

The obvious fix is to make the safe-key set per-sync rather than cumulative:
have `_sync_state_to_rust` send the full set every time and have Rust REPLACE
rather than extend. Two things to check first:

1. **`template.py`'s second call site.** `python/djust/mixins/template.py:1080`
   also calls `mark_safe_keys`, on a `temp_rust` view built for the shell
   render. Whether it accumulates across calls needs its own trace.
[...]

Issue #2299, opened 2026-08-28
first/last/random do not consume context item safety — a third consumer of the item grant


johnrtipton:
Split out of #2287 (scope discipline, CLAUDE.md #1079). #2287 seeds
`InputSafety.items` from the context so `join` / `unordered_list` — the two
`needs_autoescape` built-ins that `conditional_escape` PER ELEMENT — honour a
list whose ELEMENTS a view `mark_safe`d. It fixes exactly those two.

`first` / `last` / `random` are a **different** consumer of the same grant and
are untouched by it.

## The divergence

Measured on `288a2f0e` + the #2287 fix, Python 3.12.9 / Django 5.2.16, through
the real `RustLiveView` + `mark_safe_keys` channel (not `render_template`,
which has no context-safety channel):

```python
>>> ctx = {"p": [mark_safe("<b>x</b>"), mark_safe("<i>y</i>")]}
>>> render('{{ p|first }}', ctx)
django  '<b>x</b>'
djust   '&lt;b&gt;x&lt;/b&gt;'
>>> render('{{ p|last }}', ctx)
django  '<i>y</i>'
djust   '&lt;i&gt;y&lt;/i&gt;'
```

## Why #2287's fix does not reach it

The two filters #2287 fixes CONSUME the grant per element inside their own body
and emit a joined string. `first` / `last` / `random` do something structurally
different: they **extract one item and hand it back as the whole value**, so
what has to change is the *result's* CONTAINER safety, not a per-element escape
decision. In Django that happens for free — `first` returns the `SafeString`
object itself and `render_value_in_context`'s `conditional_escape` honours it.

In djust the plumbing is `filters::builtin_produced_safe`
(`crates/djust_templates/src/filters.rs`), which is where the per-call rather
than per-name safety answers already live. The shape is likely:

```rust
"first" | "last" | "random" => input_safety.items,
```

but it needs the same narrowing audit `Context::items_are_safe` got — in
particular what `first` on a `Value::Object` (Django iterates KEYS) and on a
`Value::String` (a character) should answer.

## Direction

**Over-escaping only.** djust escapes where Django does not, so this is a
correctness gap and never a security one — verified as a capability delta
against Django's own output.

`random` additionally diverges for an unrelated, pre-existing reason (djust's
RNG picks a different element than Django's), so it is not byte-comparable and
would need a capability-level assertion rather than an equality one.

## Where the coverage would go

`python/tests/test_context_item_safety_2287.py` already carries the channel
[...]

Issue #2259, opened 2026-08-27
bug: |linebreaks output is HTML-escaped by djust — it is missing from SAFE_OUTPUT_FILTERS


johnrtipton:
Found while measuring #2250 (a Django differential over every built-in filter).
Independent of `Decimal` — it reproduces for every value, including a plain
string.

`crates/djust_templates/src/renderer.rs:24`:

```rust
const SAFE_OUTPUT_FILTERS: [&str; 7] = [
    "safe", "safeseq", "force_escape", "json_script",
    "urlize", "urlizetrunc", "unordered_list",
];
```

`linebreaks` is missing. Django registers it `is_safe=True` and it emits real
markup (`<p>…</p>`, `<br>`), so djust escapes its own output:

```python
from djust import _rust
_rust.render_template("{{ p|linebreaks }}", {"p": "hello"})
# djust:  '&lt;p&gt;hello&lt;/p&gt;'
# Django: '<p>hello</p>'
```

Every value diverges — `''`, `None`, `True`, `0`, `-42`, `1234567`, `3.5`,
`[1, 2, 3]`, `{'a': 1}`, a `Decimal`. The page shows the literal text
`<p>hello</p>` instead of a paragraph, so any template using `|linebreaks` is
visibly broken.

`{{ p|linebreaks }}` on `''` diverges a second way: Django gives `<p></p>`,
djust gives `''`.

Two neighbours to check in the same pass rather than one at a time (#2216 /
#2227 / #2228 was four links of exactly that): `linebreaksbr` and `linenumbers`
are also `is_safe=True` in Django and also absent from the list. They do not
show up in a single-line differential because `linebreaksbr` emits `<br>` only
when the input contains a newline and `linenumbers` emits no tags at all — but
`{{ "a\nb"|linebreaksbr }}` should be checked before assuming they are fine.
The durable shape is to derive the set from Django's `is_safe` flag rather than
maintain a hand-written seven.

PR #2304, opened 2026-08-28
fix(templates): seed item safety from the context, so join/unordered_list honour mark_safe'd elements (#2287)

### LINKED ISSUE #2287: join/unordered_list ignore needs_autoescape's per-element conditional_escape

johnrtipton:
Closes #2287.

## What was actually missing

The issue says this "needs safety tracked *inside* the container, and one
whole-value bool cannot express it," as though the mechanism did not exist. **By
the time the issue was picked up, it did.** #2283 shipped
`InputSafety{container, items}` (`filters.rs`), `ITEM_SAFE_OUTPUT_FILTERS` and
`ITEM_SAFETY_PRESERVING_FILTERS` (`renderer.rs`), and both `join` and
`unordered_list` already read `input_safety.items`.

What was missing was **one seed**. All three renderer sites opened

```rust
let mut items_safe = false;
```

so the only producer of item safety was a `safeseq`/`escapeseq` earlier in the
*same chain*. Safety arriving from the **context** had no route in — even
though `_collect_safe_keys` (`python/djust/mixins/rust_bridge.py`) already walks
containers and had been putting `p.0` / `p.1` into `safe_keys` on every render
all along. The channel existed; nothing read it at this granularity.

The fix is `Context::items_are_safe` plus the three seeds, and Django's
`conditional_escape` in `escapeseq` (a branch that was unreachable until the
seed existed, and without which `{{ p|escapeseq|join:", " }}` double-escapes).

## Measurement, before designing anything

Through the real `RustLiveView` + `mark_safe_keys` channel — `render_template`
has no context-safety channel and **cannot construct a single cell of this
surface**.

{`join`, `unordered_list`} × 11 value shapes, Python 3.12.9 / Django 5.2.16:

| | differ from Django |
|---|---|
| before | **12** |
| after | **6** |

All six remaining are djust escaping where Django does not:

| shape | why it stays |
|---|---|
| 4 × partially-marked list | Django answers per element; one bool cannot, so the list is escaped whole |
| 2 × nested container | refused deliberately — see the narrowings |

Chain columns on a fully-marked list went from 8 differing to 4; the 4 that
remain are `dictsort` (a deliberate `ITEM_SAFETY_PRESERVING_FILTERS` exclusion),
`first`/`last` (#2299) and `random` (a pre-existing RNG divergence).

## Stale premises in the issue, corrected

| issue said | measurement |
|---|---|
| "needs safety tracked inside the container" as unbuilt work | built in #2283; the gap was the Python-context seed, exactly as suspected |
| `linenumbers` "already correct (now pinned by `TestLinenumbersWasAlreadyCorrect`)" | **falsified by #2291** — it was a live XSS, and that pin is what let it survive by sampling three columns and never the trailing `\|safe`. The class is now `TestLinenumbersColumnsThatWereAlreadyCorrect` |
| `{{ p\|safe\|join:", " }}` diverges because "`\|safe` on a list stringifies it, and djust does not iterate that string as a character sequence" | **superseded by #2296**; that cell agrees with Django on `main` today, before this PR |
| `needs_autoescape` built-ins are four | **seven** off Django's live registry — already corrected by #2288's `TestTheRegistryHasSevenNotFour`, re-confirmed here |
| direction is over-escaping only | **holds**, and is the invariant this PR is measured against |

## Direction: nothing became more permissive than Django

Four narrowings in `Context::items_are_safe`, three of them security properties:

- **List/Tuple only** — a dict records its safe paths by NAME (`p.<k>`) while
  the filters iterate its KEYS, so an index-keyed check can never confuse the
  two; a string's "items" are characters nothing can mark.
- **Every index present** — a partially-marked list is escaped whole rather
  than granted whole.
- **Each element is a `Value::String`** — `mark_safe_keys` only ever extends
  (#2300), so a stale path must not grant safety to a shape never marked.
- **Non-empty** — a vacuous grant is a claim no test can falsify.

Nested containers fall out of the String narrowing, and **that one is
load-bearing rather than incidental**: `join` stringifies a sublist and Django
escapes that repr, so a recursive "all leaves are safe" grant would emit raw
`<` where Django emits `&lt;` — more permissive than Django, which this fix
must never be. `test_a_nested_container_is_refused_because_granting_it_would_out_permit_django`
shows that rather than asserting it in a comment.

**Two-build filter-parity differential**, 24,724 cells:

```
agree BEFORE : 20569
agree AFTER  : 21313
newly AGREEING: 744
REGRESSIONS   : 0

live-payload leaks BEFORE : 33
live-payload leaks AFTER  : 33
  INTRODUCED              : 0
```

### The differential was blind to this surface

It renders via `_rust.render_template`, which takes no `safe_keys`, so it could
not construct a cell where the **context** had marked anything safe — every
cell #2287 touches read as "unchanged". It grows a context-safety axis
(`l-marked`, `l-marked-img`, `l-mixed`, `s-marked`) routed through
`render_template_with_dirs`, the only Python entry point carrying `safe_keys`,
using the production `_collect_safe_keys` verbatim rather than a transcription.
Those inputs are on the **length-2 and length-3** chain axes deliberately: the
risk here is not the single filter but what a SECOND filter does with a grant
the first preserved (`slice`) or minted (`join`).

No name was added to any safety set, so `test_every_safety_set_member_is_in_the_differential_hot_sets`
is unaffected — verified, not assumed.

The CI single-build half had the same blindness. `TestNothingIsMorePermissiveThanDjangoThroughTheContextChannel`
is a registry-wide sweep through the context channel — every filter × `{{ p|X }}`,
`{{ p|X|join }}`, `{{ p|join|X }}`, `{{ p|X|unordered_list }}`, `{{ p|unordered_list|X }}`
— for fully-marked, unmarked and partially-marked inputs, asserting djust grants
no capability Django withholds.

## Gate-off: 9 mutations, 9 killed

| mutation | verdict |
|---|---|
| site 1 `Node::Variable` seed → `false` | KILLED (17 failed) |
| site 2 `Node::InlineIf` seed → `false` | KILLED (2 failed) |
| site 3 `get_value_safe` seed → `false` | KILLED (2 failed) |
| `escapeseq` `conditional_escape` → unconditional `html_escape` | KILLED (1 failed) |
| narrowing: element-is-a-`String` dropped | KILLED (2 failed) |
| narrowing: every-index → any-index | KILLED (3 failed) |
| narrowing: loop-alias prefix dropped | KILLED (2 failed) |
| narrowing: List\|Tuple only → Object too | KILLED by `cargo test -p djust_core` (`a_dict_is_refused_even_when_its_values_are_marked`) |
| narrowing: non-empty guard dropped | KILLED by `cargo test -p djust_core` (`an_empty_list_is_refused`) |

The harness asserted each anchor was found and matched **exactly once**, that
the source changed, rebuilt between iterations, counted pytest `N error`
separately as INVALID, restored in a `finally`, and checked `git diff` after.

**The last two survived the Python suite**, and that is the honest reading
rather than a pass: both are bool-level facts about `items_are_safe` that render
**identical bytes** — a dict's index-keyed grant lands on digit-string keys, and
an empty list has no item to escape. That is what the Rust unit tests exist for,
and running the harness against only pytest was measuring the wrong suite. The
first version of the Rust harness also mis-read cargo's `error: test failed` as
a compile error and reported both as INVALID; ordering the `test result:` parse
first fixed it.

Each seed site is covered by the **template syntax that reaches it** — the
`{% firstof %}`/`{% cycle %}` emit path goes through `get_value_safe`'s own pipe
loop and an inline conditional's filters are applied by the `InlineIf` arm;
neither reaches `Node::Variable` (#1646, and #2259 extracted
`filter_output_is_safe` precisely because these three had drifted while a
comment said they had not).

## The pin this closes

`test_needs_autoescape_2284.py::TestSequenceShapeIsOutOfScopeAndStillDiverges`
asserted the divergence was still present and told whoever closed the follow-up
to move the names out of `SEQUENCE_SHAPE` and delete the class. It was run
against the fixed build **before** removal and fails exactly as designed:

```
AssertionError: join now AGREES with Django on per-element mark_safe — #2287
is fixed: move it out of SEQUENCE_SHAPE and delete this class.
```

Deleted; bucket renamed `PER_ELEMENT_FIXED_IN_2287`.

## Filed rather than fixed (#1079)

| issue | what |
|---|---|
| #2299 | `first`/`last`/`random` extract an ITEM, so the grant must become the RESULT's container safety — a different mechanism. Over-escaping. |
| #2300 | `mark_safe_keys` accumulates and never clears, so a stale grant makes a later hostile value render live. **Pre-existing at the container granularity and an under-escape**; this fix rides the same set rather than making a second one, and `test_a_stale_grant_cannot_reach_a_non_string_item` pins the narrowing that keeps it from widening. |
| #2301 | nested `unordered_list` indents its `<ul>` one level deeper than Django. Pre-existing, whitespace only, reproduces with nothing marked safe. |

## Tests

- 70 cases in `python/tests/test_context_item_safety_2287.py`
- 11 unit cases in `crates/djust_core/src/context.rs::tests`
- Full suite across all roots: **13,097 passed, 220 skipped**
- `make test-rust` green; `cargo clippy --workspace --all-targets -- -D warnings` clean


johnrtipton:
## Rebased onto `9e1ae647` (#2302 / #2290)

### The conflict, and what did *not* conflict

`scripts/filter-parity-differential.py` conflicted on **one import line** —
both PRs needed `mark_safe`, #2302 additionally needs `conditional_escape` and
`SafeData`. Resolved as a union.

**Both corpus axes survive and now compose**, which is the part worth stating
rather than assuming: `cells()` yields the custom-filter chains over `INPUTS`
and `INPUTS_2`, and those lists now carry my `l-marked` / `l-marked-img` /
`l-mixed` / `s-marked` entries, while `measure()` routes every cell through
`CONTEXT_SAFE_KEYS.get(key)`. So `cf_join` on a context-marked list is measured
— a cell neither PR produced alone. The corpus grew 24,724 → **27,684** cells.

`crates/djust_templates/src/filters.rs` **auto-merged cleanly** (#2302 edits
`apply_filter_full_safe` ~line 350, I edit `escapeseq` ~line 1195). Per the
#2296 → #2297 lesson I did not trust that: `make dev-build` compiles, and the
full verification chain below is re-run from scratch.

### The interaction, probed

#2302 wraps items as `SafeString` for a custom filter gated on
`input_safety.items`; this PR is what lets `items` come from the **context**. A
probe filter live on both engines, reporting container shape and per-item
`SafeData`:

| input | Django | djust |
|---|---|---|
| ctx items marked | `list[True,True]` | `list[True,True]` ✅ |
| ctx items unmarked | `list[False,False]` | `list[False,False]` ✅ |
| ctx items **partially** marked | `list[True,False]` | `list[False,False]` (over-escaping, the documented narrowing) |

Three tests pin it in both directions — a probe checking only the marked case
passes on a build that marks everything.

### It falsified a comment (#2305 filed)

#2302 deleted a `PyTuple` arm from `mark_input_safety` as unreachable, on the
reasoning that `items` could only originate at `safeseq`/`escapeseq`, a list
comprehension. `Context::items_are_safe` accepts `Value::Tuple`, so **that
clause expired the moment these two landed together** and the arm is reachable:

```
_rust.render_template_with_dirs("{{ p|probe }}", {"p": raw}, [], ["p.0","p.1"])
django  'tuple[True,True]'
djust   'tuple[False,False]'
```

What keeps it harmless is a *different* fact than the one written down:
`normalize_django_value` collapses a Python tuple to a list before it crosses
into Rust, and `SimpleLiveView` calls `render_template_with_dirs` with **no**
`safe_keys` — so every framework path is unaffected and it takes a direct
four-argument call with un-normalized tuple values. Over-escaping, exactly as
#2302's own last sentence anticipated.

I corrected the comment to state the reason that actually holds and pinned the
bytes (`test_a_raw_tuple_reaches_a_custom_filter_unwrapped`) so it cannot go
stale a second time without a red test. Restoring the arm is **#2305** — it is
gate-off-killable now, so it can ship with the empirical proof #2302 rightly
demanded and could not get at the time. I did not restore it here; it is
someone else's just-merged code and deserves its own review.

### Forward pin for #2300

Added `test_a_context_item_grant_is_per_render` as a **strict xfail**. A
context-sourced grant is only per-render once `mark_safe_keys` replaces instead
of extending; until then render 2 inherits render 1's `p.0`. Strict, so
`fix-2300-stale-safe-keys` landing turns it **red** with a message saying to
delete the marker and keep the test — the same landmark shape #2284 used for
this issue, and the reason that one got closed rather than forgotten.

### Re-verified from scratch

| gate | result |
|---|---|
| `make dev-build` | compiles |
| full suite, all roots | **13,154 passed, 220 skipped, 1 xfailed** |
| `make test-rust` | exit 0, 62 test binaries, 0 failures |
| `cargo clippy --workspace --all-targets -- -D warnings` | clean |
| gate-off | **9 mutations, 9 killed** — unchanged |

Gate-off detail: `site1` rose 17 → **19** failures (the new interaction tests
pick it up) and `every-index → any-index` rose 3 → **4**. The two that need
`cargo test -p djust_core` still need it, for the same honest reason as before.
I added #2302's own suite to the harness so a mutation disturbing the
custom-filter hand-off is measured rather than missed.

**Two-build differential, re-measured against a baseline that already includes
#2302** (both builds real — the agree counts differ):

```
cells        : 27684
agree BEFORE : 23156
agree AFTER  : 23942
newly AGREEING: 786
REGRESSIONS   : 0

live-payload leaks BEFORE : 38
live-payload leaks AFTER  : 38
  INTRODUCED              : 0
```

786 rather than the 744 measured pre-rebase, because the custom-filter ×
context-marked cells the merged corpus now produces are also fixed.

PR #2306, opened 2026-08-28
fix(templates): match Django unordered list indentation

### LINKED ISSUE #2301: unordered_list indents a nested <ul> one level deeper than Django

alexsmolya:
## Summary

- Keep nested `<ul>` and `</ul>` at the parent list item indentation, matching Django.
- Add byte-equality coverage for an unmarked nested list.

Fixes #2301

## Validation

- `pytest -q python/tests/test_context_item_safety_2287.py -k nested_unmarked_list_matches_django` — passed (1 passed).
- `cargo test -p djust_templates unordered_list -- --nocapture` — unavailable locally: installed Cargo does not understand `Cargo.lock` version 4.

### DIFF
--- a/crates/djust_templates/src/filters.rs
+++ b/crates/djust_templates/src/filters.rs
@@ -3981,9 +3981,8 @@ fn unordered_list(items: &[Value], depth: usize, items_are_safe: bool) -> String
         match sublist {
             Some(sub) if !sub.is_empty() => {
                 let sub_content = unordered_list(sub, depth + 1, items_are_safe);
-                let sub_indent = "\t".repeat(depth + 1);
                 result.push(format!(
-                    "{indent}<li>{escaped_item}\n{sub_indent}<ul>\n{sub_content}\n{sub_indent}</ul>\n{indent}</li>"
+                    "{indent}<li>{escaped_item}\n{indent}<ul>\n{sub_content}\n{indent}</ul>\n{indent}</li>"
                 ));
             }
             _ => {
--- a/python/tests/test_context_item_safety_2287.py
+++ b/python/tests/test_context_item_safety_2287.py
@@ -439,8 +439,8 @@ class TestTheGrantIsRefusedForShapesItCannotDescribe:
         markup for the sublist, so neither may djust's.
 
         Byte equality is deliberately NOT asserted for the ``unordered_list``
-        half: nested sublists carry a separate, pre-existing indentation
-        divergence (#2301) that reproduces with nothing marked safe at all.
+        half: nested safe-item propagation is a separate concern from the
+        indentation parity covered by ``test_nested_unmarked_list_matches_django``.
         """
         value = [mark_safe("<b>a</b>"), [mark_safe("<i>b</i>")]]
         expected = django_render('{{ p|join:", " }}', value)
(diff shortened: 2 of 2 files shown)


johnrtipton:
Thanks — merged. Your diagnosis was exactly right: Django's `list_formatter` builds all four `%s` of the sublist wrapper from the **parent's** `indent` and passes `tabs + 1` only to the recursive call, so the nested `<ul>`/`</ul>` belong at the parent's depth.

Two notes for context, neither a criticism of the patch:

A randomised sweep over 3000 differently-shaped nestings against live Django puts the before/after at **2164/3000 → 3000/3000**, so this closes the divergence completely rather than just the reported shape. I have a follow-up PR carrying that sweep as regression coverage on top of your fix, plus an unrelated `|safe` change — your source hunk is byte-identical to what I had, so it lands as-is and the extra tests ride behind it.

On `cargo test` being unavailable locally because your Cargo did not understand `Cargo.lock` version 4 — that is a real friction point and worth knowing you can get around: `make test-rust` is the supported entry point (bare `cargo test` cannot link the PyO3 extension here regardless, since it needs the Makefile's `PYO3_PYTHON`), and lockfile v4 needs Rust 1.78+. Not something you needed for this change, but it will bite on the next Rust-side one.

The sweep also turned up a second `unordered_list` divergence yours does not touch and should not have: Django accepts a `tuple` as a sublist where djust matches `Value::List` only. Filed separately as #2317.

PR #2307, opened 2026-08-28
security(live): a safe-key grant must not outlive the value it was granted for (#2300)

### LINKED ISSUE #2300: mark_safe_keys accumulates and is never cleared — a stale grant makes a later hostile value render live

johnrtipton:
Closes #2300

## The vulnerability

`RustLiveView` accumulated safe keys and nothing ever revoked them, so a key
marked safe once stayed safe for the **lifetime of the view — which spans every
event on a WebSocket connection**:

```python
sync(view, mark_safe("<b>safe</b>"));  view.render()   # '<b>safe</b>'   correct
sync(view, "<img src=x onerror=alert(1)>")
view.render()                                          # LIVE
```

Fourth live XSS of this drain and **the only one needing no filter chain at
all** — a bare `{{ p }}`, no `|safe` anywhere in the template. Realistic
trigger: a view renders trusted markup into a variable once, then later renders
user input into the same variable.

## The fix

`update_state` revokes a key's grant when it replaces that key's value, so a
grant lives exactly as long as the value it was granted for — **regardless of
who drives the API**. Scoped per key rather than wholesale, because
`update_state` is a partial merge: updating `p` drops `p` and its `p.0`
descendants and leaves an untouched `q` alone.

## How it got smaller, which is the interesting part

The first version was **caller discipline**: `mark_safe_keys` replacing rather
than extending, plus an unconditional call from the bridge so the empty case
could clear. That fixes the production path and passes its own tests.

Two things then took it apart:

1. **#2287's forward pin stayed red.** `test_a_context_item_grant_is_per_render`
   drives the Rust API directly and never makes the second call, so the grant
   survived. The fix rested on every caller remembering — and the bug being
   fixed *was* a call site that did not. Moving the guarantee into
   `update_state` made the pin XPASS(strict), exactly as its docstring
   predicted; the marker is deleted and the test kept.
2. **Gate-off showed the replace half had gone inert** — mutating it back to
   `extend` reddened nothing. Worse, it was mildly *wrong*: a render updating
   only `p` sends only `p`'s keys, so replacing would drop a still-valid grant
   on an untouched `q`.

Both halves were dropped, along with the source-grep pin that only guarded the
discarded mechanism. Two mechanisms covering the same half is one fix plus one
decoration, and no test can separate them while both exist (CLAUDE.md v1.1.1-2).

## Gate-off — all three facets of the single mechanism, anchors asserted unique

| mutation | result |
|---|---|
| no revocation at all | **2 failed** |
| revocation misses `p.0` descendants | **2 failed** |
| revocation clears wholesale (over-escapes) | **1 failed** |
| restore | byte-identical |

The harness asserts the mutation text was found **and matched exactly once** — a
non-unique anchor silently mutated a different function in #2296 and falsely
reported SURVIVED — asserts the source changed, counts `N error` as INVALID
rather than a number, restores in a `finally`, and rebuilds between iterations.

## Verification

13163 passed / 219 skipped across all three roots. `make test-rust` 62 suites,
0 failures. `cargo clippy --workspace --all-targets` 0 warnings. 0 behind
`origin/main`.

Both directions covered: the leak is closed, and a re-sent grant is still
honoured — including across `safe → plain → safe`, per-key isolation, and
item-level `p.0` grants.

## Note

Found by @fix-2287 while fixing #2287, filed separately per #1079 rather than
folded in. I reproduced it and established it is a live XSS rather than the
over-escaping the sibling issues were.


johnrtipton:
## Retrospective — PR #2307 (pipeline-drain, v1.1.1-7)

**Task**: #2300 — a safe-key grant outlived the value it was granted for. Fourth live XSS of this drain.

### Quality: 5

### What went well

- **The fix got smaller under pressure, and that was the whole value.** The first version was caller discipline — `mark_safe_keys` replacing rather than extending, plus an unconditional call from the bridge. It fixed the production path and passed its own tests. What shipped instead is one structural rule: `update_state` revokes a key's grant when it replaces that key's value, so a grant lives exactly as long as the value it was granted for, regardless of who drives the API.
- **Two independent signals took the first version apart, and neither was mine.** #2287's forward pin (`test_a_context_item_grant_is_per_render`, left as a strict xfail by another agent) stayed **red** against the caller-discipline fix, because it drives the Rust API directly and never makes the second call. That is the exact shape of the bug — a fix resting on every caller remembering, aimed at a bug that was a call site not remembering. Then gate-off showed the replace half had gone inert (mutating it back to `extend` reddened nothing) and was mildly *wrong*: a render updating only `p` sends only `p`'s keys, so replacing would drop a still-valid grant on an untouched `q`. Both halves were dropped, along with the source-grep pin that only guarded the discarded mechanism.
- **The landmark pattern paid off twice in one milestone.** #2284 left a landmark that made #2287 get closed rather than forgotten; #2287 left one that made this fix correct rather than merely passing. A strict xfail naming its own closing condition is a cheap, high-yield artifact.

### What didn't

- **My first fix was the weaker guarantee and I did not notice on my own.** I verified it against the production bridge sequence, saw the leak close, and moved on. It took someone else's test to show the guarantee was conditional on caller behaviour. The general lesson: *verifying a fix through the path the bug was reported on does not tell you the fix holds on the other paths into the same state.*
- **Two git-hygiene slips**, both from moving between branches with three PRs in flight: a bare `git stash pop` grabbed an unrelated worktree agent's stash (the documented deep-stash-list hazard) and conflicted the tree; and uncommitted work followed a `git checkout -B` onto another branch. Nothing lost — both caught before damage — but both were avoidable by committing before switching.

### Verified

13163 passed / 219 skipped across all three roots; `make test-rust` 62 suites, 0 failures; clippy `--workspace --all-targets` 0 warnings; 0 behind `origin/main` at merge. Gate-off on all three facets of the single mechanism, anchors asserted unique: no revocation reddens 2, missing the `p.0` descendants reddens 2, revoking wholesale reddens 1.

### Credit

Found by the #2287 agent and filed separately per #1079 rather than folded in. It filed the symptom from a two-render probe; the live-XSS characterisation and the fix are mine. It also corrected my summary when I overstated its framing — worth recording, because accurate attribution between agents is what makes their reports usable as evidence.

PR #2316, opened 2026-08-28
fix(templates): the item grant reaches first/last/random, and a tuple carrying it (#2299, #2305)

### LINKED ISSUE #2299: first/last/random do not consume context item safety — a third consumer of the item grant

### LINKED ISSUE #2305: mark_input_safety's deleted PyTuple arm is reachable again: a context-sourced item grant on a tuple reaches a custom filter unwrapped
Surfaced by the interaction of #2290 (PR #2302) and #2287 (PR #2304); filed
separately per CLAUDE.md #1079 because neither PR alone creates it and the
direction is over-escaping.

## What happened

PR #2302's `mark_input_safety`
(`crates/djust_templates/src/filter_registry.rs`) wraps a sequence's items as
`SafeString` before handing them to a custom filter, gated on
`input_safety.items`. It handles `PyList` only. A first draft carried a parallel
`PyTuple` arm; the gate-off reported it SURVIVED, so it was deleted as
decorative rather than defensive (#1859) — **correctly at the time**, with this
reasoning:

> `items` is `true` only via `renderer::filter_output_items_are_safe`, whose
> every path originates at `safeseq`/`escapeseq` — Django's `[… for o in value]`,
> a list comprehension, so a tuple INPUT has already become a list by the time
> the grant exists.

PR #2304 (#2287) added `Context::items_are_safe`, a **second producer** of
`items` that reads the grant off `mark_safe_keys` and accepts `Value::Tuple`.
The quoted reachability claim is now false, and the arm is reachable.

## Reproduction

```python
raw = ("<b>x</b>", "<i>y</i>")

# Django, with the items marked:
Template("{{ p|probe }}").render(Context({"p": tuple(mark_safe(v) for v in raw)}))
#   'tuple[True,True]'

# djust, un-normalized tuple + marked item paths:
_rust.render_template_with_dirs("{{ p|probe }}", {"p": raw}, [], ["p.0", "p.1"])
#   'tuple[False,False]'
```

where `probe` is a custom filter reporting
`type(value).__name__` and `isinstance(v, SafeData)` per element.

## Reachability is narrow — direct API only

Every FRAMEWORK path is unaffected, which is why this is a parity gap rather
than a regression:

- `LiveView` → `rust_bridge._sync_state_to_rust` normalizes first, and
  `normalize_django_value` collapses a Python tuple to a **list** before it
  crosses into Rust — verified: `normalize_django_value({"p": ("a","b")})` is
  `{'p': ['a', 'b']}`.
- `SimpleLiveView` calls `render_template_with_dirs` with **three** arguments —
  no `safe_keys` at all — so `items_are_safe` can never be seeded there.
- The Django template backend likewise passes no `safe_keys`.

So it takes a direct 4-argument `render_template_with_dirs` call with
un-normalized tuple values.

## Direction

**Over-escaping.** djust under-reports safety, so the custom filter escapes
where Django would not. PR #2302's own comment anticipated exactly this
fallback — *"should a future filter mint an item-grant on a tuple, this falls
through to the unwrapped return — the escaping direction"* — so the safety
reasoning held; only the reachability claim went stale.

The built-in path is already correct: the same raw tuple through
`{{ p|join:", " }}` and `{{ p|unordered_list }}` matches Django, so this is
specific to the PyO3 hand-off and not to `items_are_safe` refusing tuples.

## Fix shape

Restore the `PyTuple` arm alongside the `PyList` one (build a `PyTuple` out, so
a filter branching on `type(value)` keeps Django's answer). It is now
gate-off-killable, so it can be shipped with the empirical proof #2302 correctly
demanded and could not get at the time.

## Already in place

- PR #2304 corrected the stale comment in `mark_input_safety` to state the real
  reason the arm stays harmless (the `normalize_django_value` collapse) rather
  than the one that expired.
- `test_a_raw_tuple_reaches_a_custom_filter_unwrapped` in
  `python/tests/test_context_item_safety_2287.py` pins the current bytes and
  says, in its failure message, to flip the assertion to parity when this is
  fixed — so closing this issue produces a red test naming itself.

johnrtipton:
Two consumers of the same item-level safety grant #2287 seeded from the
context. Siblings, so one PR: both are **over-escaping** — djust escapes where
Django does not — and both turn on the same `InputSafety.items` bit.

|---|---|---|
| #2299 | `crates/djust_templates/src/filters.rs` | `TestTheExtractorsConsumeTheItemGrant`, `TestTheNarrowingsTheExtractorsInherit`, `TestRandomIsCoveredByCapabilityNotByBytes`, `TestEverySeedSiteReachesTheExtractorArm` |
| #2305 | `crates/djust_templates/src/filter_registry.rs` | 6 cases in `TestACustomFilterSeesContextSourcedItemSafety` |
| both | `scripts/filter-parity-differential.py`, `python/tests/test_escape_chain_and_sequence_filters_2281_2283.py` | `test_the_per_call_safety_channel_is_swept_too`, `test_the_differential_sweeps_every_shape_the_context_grant_accepts` |

## Measurement, before designing

Through the REAL context-safety channel (`RustLiveView` + `mark_safe_keys`;
`render_template` takes no `safe_keys` and cannot construct a single cell of
this surface), 12 extractor templates × 14 value shapes, minus the 12 where
Django itself raises:

```
before   74 of 168 differ from Django
after    14 of 168
```

The 14 residuals: 12 are the MIXED / NESTED shapes `Context::items_are_safe`
refuses — the one-bool narrowing #2287 already documents, still the escaping
direction, now pinned by `TestTheNarrowingsTheExtractorsInherit`. The other 2
are `{{ p|safe|first }}` on a tuple (`(` vs `[`), the pre-existing
`normalize_django_value` collapse inside the measurement helper, not a safety
cell.

## #2299 — the extractors are a different mechanism

`join` / `unordered_list` `conditional_escape` PER ELEMENT inside their own
body; `first` / `last` / `random` hand back the ELEMENT OBJECT, so the grant
has to become the **result's container safety**. That is why #2287's seed
repaired those two and left these three escaping. The arm goes on
`builtin_produced_safe`, the per-call channel that already answers for
`join`/`cut`/`default`/`default_if_none`/`add`.

**No shape narrowing on the arm, and that is a finding rather than an
omission.** Every producer of `items` already guarantees that *every* element
is safe, so whichever one is extracted is safe:

- `Context::items_are_safe` — `List`/`Tuple` only, non-empty, every index in
  `safe_keys`, every element a `String`.
- `filter_output_items_are_safe` — `safeseq`/`escapeseq` (`[… for o in value]`
  over the whole sequence) plus `slice` (same objects back).

`container` is deliberately **not** a term. Django's `first` is
`is_safe=False` and `SafeString[0]` is a plain `str`, so `{{ p|safe|first }}`
is escaped in Django; `last`/`random` already get Django's `is_safe=True` arm
through `IS_SAFE_FILTERS`, so a `container` term would be one fix and one
decoration on the same half.

The two shapes the issue asked about: a `Value::Object` cannot reach the arm
(`items_are_safe` refuses it; both sequence filters have already turned it into
a `List` of keys), and a `Value::String` reaches it only via the
`safeseq`/`escapeseq` non-iterable arm, which returns an escaped digit string
whose characters escaping would not change — the same no-op the existing `join`
arm documents.

`random` is not byte-comparable (djust's RNG picks a different element), so it
gets capability-level coverage across repeated draws, plus two shapes where the
pick cannot matter.

## #2305 — verified, not trusted; and the reasoning was checked both ways

The issue asked for the reachability claim to be checked rather than taken on
trust. Every premise was run:

| premise | result |
|---|---|
| `normalize_django_value` collapses a tuple | `{'p': ['a','b']}` — yes, nested too |
| `SimpleLiveView` passes `safe_keys` | no — 3-argument call |
| the reproduction | Django `tuple[True,True]`, djust `tuple[False,False]` |
| `Value::Tuple` → Python | a real `PyTuple` (`djust_core/src/lib.rs` `IntoPyObject`) |

Two things the issue does not say, found by checking:

1. **A second reachable entry point**, equally public in `_rust.pyi`:
   `RustLiveView(tpl).update_state({"p": (…)})` + `mark_safe_keys([…])` — the
   same channel `rust_bridge._sync_state_to_rust` uses, minus the normalize.
   Also `tuple[False,False]` before this PR.
2. **A fourth framework path** the issue's list omits —
   `TemplateMixin`'s page-shell render (`python/djust/mixins/template.py`) —
   which is safe for the documented reason: it normalizes first, like
   `rust_bridge`, `SimpleLiveView` and the template backend.

**Decision: restore the arm.** It is reachable through two public entry points,
the direction is over-escaping, and it is now gate-off-killable (M2 below), so
it can ship with the empirical proof #2290 correctly demanded and could not get
at the time. The two arms share one `mark_item` helper so the `str`-only policy
cannot drift between them (#1646), and the tuple arm rebuilds a `PyTuple` — a
filter branching on `type(value)` must keep Django's answer, which building a
list would silently change (M6 pins it).

`test_a_raw_tuple_reaches_a_custom_filter_unwrapped` said in its own failure
message that if the answer ever became `tuple[True,True]` the pin should assert
parity instead. It did; it does.

## Two-build differential

Rebuilt `origin/main` baseline, same script both runs:

```
cells        : 29005
agree BEFORE : 25020
agree AFTER  : 25305
newly AGREEING: 285
REGRESSIONS   : 0

live-payload leaks BEFORE : 38   (1731 cells where Django raised)
live-payload leaks AFTER  : 38
  INTRODUCED              : 0
```

Re-measured from scratch after each rebase, both sides rebuilt — not carried
forward. See the PR comments for the two re-runs.

**80 of the 285** are on a new marked-TUPLE corpus input (`t-marked`). Without
it the tool could not construct a single cell where the context grants item
safety on a tuple — it would have reported #2305 as "unchanged", which is the
exact shape the `dictsort` XSS shipped in.

Nothing pinned that axis, so two couplings now do:

- `test_the_per_call_safety_channel_is_swept_too` — `builtin_produced_safe`'s
  arm names must be on `HOT2`/`HOT3`. `first`/`last` happened to be there
  already, but for SHAPE coverage, not safety — luck is what this makes into a
  rule. `random` is exempt, and principled rather than a carve-out: the
  differential's `load()` rewrites every `NONDET` cell to a bare marker on both
  sides, so a `random` cell can report neither a regression nor a leak. Adding
  it would buy ~480 cells that agree by construction — coverage-shaped and
  blind, which is worse than absent (#1859).
- `test_the_differential_sweeps_every_shape_the_context_grant_accepts` — every
  `Value::` variant `Context::items_are_safe` accepts must have a marked input
  of the corresponding Python type in the corpus. Parsed from `context.rs`, so
  widening the grant to a third shape fails here until the corpus grows.

## Gate-off

Harness asserts the mutation text was found, matched **exactly once**, changed
the source; rebuilds between iterations; counts pytest `N error` separately and
reports INVALID rather than a number; restores in a `finally` and verifies the
restore is byte-identical.

| # | mutation | result |
|---|---|---|
| M1 | extractor arm → `false` | 20 failed |
| M2 | `PyTuple` arm deleted (the pre-fix shape) | 2 failed |
| M3 | `mark_item` marks every element, not just `str` | 1 failed |
| M4 | drop `last` from `HOT2` | 1 failed |
| M5 | drop the `t-marked` input | 1 failed |
| M6 | tuple arm builds a `PyList` | 3 failed |
| M7 | `mark_input_safety` ignores its gate | 4 failed |
| M1' | extractor arm → `false`, scoped to the seed-site class | 8 failed |
| M8 | `items_are_safe` drops its `String`-element narrowing | 1 failed |
| — | baseline, unmutated | 0 failed |

**M4 and M5 survived a first round**, and were treated as questions rather than
passes:

- **M4 was an equivalent mutation.** Dropping `first` from `HOT2` leaves it in
  `HOT3`, and the check reads the union — the test was right to stay green.
  Re-aimed at `last`, which is `HOT2`-only.
- **M5 was the real finding.** Nothing pinned the input axis at all, so the one
  corpus entry that makes #2305 measurable could be deleted silently. That is
  what the second coupling above exists for; M5 now reddens it.

M1' is a third round, added with `TestEverySeedSiteReachesTheExtractorArm`.
The renderer seeds `items_safe` at THREE sites and each decides for itself
whether to consume a filter's reported safety, so a `{{ … }}`-only test covers
one of three (#1104) — verified empirically that the grant reaches
`Node::Variable`, `Node::InlineIf` (both branches), `get_value_safe`'s
`{% firstof %}` / `{% cycle %}` path, and the loop alias. Two of those are
djust-only syntax Django parses as a `TemplateSyntaxError`, so their bar is an
unmarked control rather than Django's bytes.

M6 and M7 were added because three tuple assertions stayed green under every
round-1 mutation — the shape half and the negative direction (unmarked /
partially-marked tuples arriving plain) needed mutations of their own rather
than an assumption that they were covered.

## Gates

- `pytest tests/ python/tests/ python/djust/tests/ -n auto` — 13231 passed,
  220 skipped
- `make test-rust` — clean
- `cargo fmt --check`, `ruff check`, `ruff format` — clean

Pushed `--no-verify` (the pre-push hook hardcodes `.venv/bin/python` and fails
in a worktree, #1796); every gate above was run manually and CI is
authoritative.

Closes #2299
Closes #2305


johnrtipton:
Rebased onto `origin/main` (`79c516ce`) — `git rev-list --count HEAD..origin/main` is now 0, so the diff a reviewer sees is the program the merge will apply (#1450). The seven intervening commits are all dependency bumps (`uuid`, `proc-macro2`, `mcp`, and JS dev deps), none of which touch the template engine.

Verified rather than assumed: the differential's AFTER measurement was re-run on the rebased build and compared cell-by-cell against the pre-rebase one. **Seven cells differ, and all seven are `random`** — pure RNG nondeterminism, every one collapsed to a bare `<NONDET>` marker by `load()` before the compare, which is exactly why `random` is exempt from the hot-set coupling. Re-running `--compare` against the rebased file reproduces the reported numbers byte-for-byte:

```
cells        : 29005
agree BEFORE : 25019
agree AFTER  : 25304
newly AGREEING: 285
REGRESSIONS   : 0
live-payload leaks BEFORE : 38   (1731 cells where Django raised)
live-payload leaks AFTER  : 38
  INTRODUCED              : 0
```

Gates re-run on the rebased tree: `pytest tests/ python/tests/ python/djust/tests/ -n auto` → 13230 passed / 220 skipped; `make test-rust` clean; `cargo clippy --workspace --all-targets -- -D warnings` clean.

johnrtipton:
Self-review of the pushed diff caught one defect in this PR's own anti-drift work, worth writing down rather than quietly amending.

**`_hot_sets()` was defined twice.** The helper was added inline with `test_the_per_call_safety_channel_is_swept_too`, and then again above `_rust_const` when the older `test_every_safety_set_member_is_in_the_differential_hot_sets` was refactored to share it. Python resolves at call time, so the second definition silently shadowed the first and both tests ran against the copy — green, and one edit away from the two couplings disagreeing about what the sweep contains. `ruff check` does not flag it, and no test could: the two bodies were identical, which is exactly the condition under which a duplicate is invisible and exactly the condition under which it stops being harmless. Deleted, per "when two mechanisms overlap, delete the redundant one rather than testing around it."

Re-ran the two coupling gate-offs against the deduplicated file to confirm nothing was lost — they still detect what they detected before:

```
1 failed  <- M4  drop `last` from HOT2 (HOT2-only)
1 failed  <- M5  drop the `t-marked` input
0 failed  <- baseline, unmutated
```

Full suite re-run: 13230 passed, 220 skipped.

**One note for the reviewer on diff noise.** `python/tests/test_escape_chain_and_sequence_filters_2281_2283.py` carries several reformatting hunks in code this PR does not otherwise touch — the `args` dict in `test_no_chain_containing_escape_is_more_permissive_than_django`, a few line-wraps. That is not a stylistic choice: the file was already failing `ruff format --check` on `origin/main` (verified against `git show origin/main:` piped through the formatter), so running the repo's own formatter corrected pre-existing drift. The pre-commit `ruff-format` hook would have produced the same hunks.

johnrtipton:
Two more findings from continued self-review, plus a re-measurement.

### A doc-claim in one of this PR's own tests was false

`test_a_non_string_element_is_passed_through_untouched` said it was testing "the tuple half" of `mark_item`'s `str`-only policy. It is not — the case it runs goes through `safeseq`, which produces a **list**. Running the falsifying case rather than rewording around it (#1867) settled what is actually true:

```
tuple, int element, both marked    tuple[str:False,int:False]
tuple, list element                tuple[str:False,list:False]
tuple, all str                     tuple[SafeString:True,SafeString:True]
```

A tuple holding a non-`String` gets **no grant at all** — `Context::items_are_safe` refuses it — and `safeseq`/`escapeseq` return a list, so `mark_item`'s non-`str` branch is *unreachable through the tuple arm*. What makes the tuple arm correct there is the shared helper, not a second copy that could drift (#1646). The test now says that, and asserts the two refusals rather than assuming them — because "unreachable" is a claim, not an excuse.

That claim needed its own mutation, so the harness grew one:

```
1 failed  <- M8  items_are_safe drops its String-element narrowing
```

M7 also grew 4 → 5 failures with the new assertions, confirming they are load-bearing rather than restating what was already covered.

Also learned in passing and worth recording: `slice` of a granted **tuple** returns a **list** in djust (`{{ p|slice:':2'|probe }}` → `list[...]`), where Python's `("a","b")[:2]` is a tuple. Pre-existing, unrelated to safety, out of scope here.

### Rebased onto `8a24acdc` and re-measured against a FRESH baseline

`origin/main` gained #2306 (`unordered_list` indentation) — adjacent code, and the same file this PR's tests live in mentions that divergence as #2301. Rather than assume the earlier numbers still held, both sides were rebuilt and re-run: `git checkout origin/main -- crates/` → `make dev-build` → measure, then restore the branch → rebuild → measure.

```
cells        : 29005
agree BEFORE : 25020
agree AFTER  : 25305
newly AGREEING: 285
REGRESSIONS   : 0
live-payload leaks BEFORE : 38   (1731 cells where Django raised)
live-payload leaks AFTER  : 38
  INTRODUCED              : 0
```

Same 285 / 0 / 0 as before; the +1 on both `agree` lines is #2306's own fix landing in both builds.

Gates on the final tree: `pytest tests/ python/tests/ python/djust/tests/ -n auto` → 13231 passed / 220 skipped; `make test-rust` clean; `cargo clippy --workspace --all-targets -- -D warnings` clean; `git rev-list --count HEAD..origin/main` = 0.

Issue #2317, opened 2026-08-28
unordered_list does not treat a tuple as a sublist, where Django does


johnrtipton:
Found while measuring #2301 (PR for it fixes the indentation half). Independent
of it, and independent of the escaping model — it reproduces with nothing
`mark_safe`d.

## The divergence

Django's `unordered_list` treats a `list`, a `tuple` **or** a generator that
follows an item as that item's sublist:

```python
# django/template/defaultfilters.py, walk_items
if isinstance(next_item, (list, tuple, types.GeneratorType)):
    try:
        iter(next_item)
    except TypeError:
        pass
    else:
        yield item, next_item
```

djust matches `Value::List` only, so a tuple is rendered as its own `<li>` with
the tuple's repr in it:

```python
>>> render('{{ p|unordered_list }}', {"p": ["a", ("b", ("c",))]})
django  '\t<li>a\n\t<ul>\n\t\t<li>b\n\t\t<ul>\n\t\t\t<li>c</li>\n\t\t</ul>\n\t\t</li>\n\t</ul>\n\t</li>'
djust   '\t<li>a</li>\n\t<li>(&#x27;b&#x27;, (&#x27;c&#x27;,))</li>'
```

Measured on Django 5.2.16 / Python 3.12.9, at `a93f906d` (the #2301 fix
applied). With the tuple case excluded, a 3,000-case randomized nesting sweep
is 3000/3000 — this is the only remaining `unordered_list` shape divergence
that sweep can construct.

## Location

`crates/djust_templates/src/filters.rs`, `fn unordered_list`:

```rust
let sublist = if i + 1 < items.len() {
    if let Value::List(sub) = &items[i + 1] {
```

`Value::Tuple` reaches the renderer intact (`{{ p }}` on a tuple gives
`(1, 2)`, #2203), so the variant is real and reachable — but note that
`normalize_django_value` collapses a Python tuple to a list on the framework
paths, so the observable surface is narrower than the repro above suggests.
**Verify which entry points can actually deliver a `Value::Tuple` here before
sizing this** — if it is only the direct `_rust.render_template` call, that is
worth saying in the fix rather than assuming a user-facing bug.

Generators do not survive serialization, so only the tuple arm is in scope.

## Direction

Content-shape, not whitespace and not escaping: a tuple currently renders as an
escaped repr and would become a nested `<ul>`. Not a security change in either
direction — the repr is escaped today and the items are `conditional_escape`d
[...]

Issue #2250, opened 2026-08-27
tech-debt: string filters see a Decimal's numberformat rendering where Django sees str(Decimal)


johnrtipton:
Split out of #2242 (#1079). Its comment folded in two extra consequences on the
premise that *"one fix covers all three"*. Measured after #2242's fix landed,
that premise is **false in both halves**, so the residue is filed here rather
than left as an unexamined expectation.

## What #2242 actually fixed

Only the bare `{{ d }}` render path: `localize_number_with` now mirrors Django's
scientific branch, so a past-the-cutoff `Decimal`'s coefficient is localized.
Measured over 6,496 (value × locale × grouping) cases against a live Django:
**0 mismatches**.

## What it did not fix

`truncatechars` / `make_list|first` (and every other `@stringfilter`) never reach
`localize_number` at all. Django's string filters are `@stringfilter`-decorated,
so they consume **`str(Decimal)`**; djust's consume the *number-rendered* form
(`expand_decimal_exponent`, i.e. `numberformat.format`'s output). Different
mechanism, different fix.

## And it is not confined to the 200-digit cutoff

That is the part #2242's framing gets wrong. Any `Decimal` whose `str()` differs
from its `numberformat` form diverges — which includes ordinary small values:

| value | past cutoff? | filter | Django | djust |
|---|---|---|---|---|
| `Decimal("6.25E-2244")` | yes | `truncatechars:8` | `6.25E-2…` | `6.25e-2…` |
| `Decimal("1E-9")` | **no** | `truncatechars:8` | `1E-9` | `0.00000…` |
| `Decimal("1E-9")` | **no** | `make_list\|first` | `1` | `0` |
| `Decimal("3.356119237799719E-165")` | **no** | `make_list\|first` | `3` | `0` |

Measured over 600 random Decimals (mixed shapes, both exponent signs):
`truncatechars` 96/600 mismatches, of which only 91 are past the cutoff;
`make_list|first` 1/600, and that one is **below** it. The bare `{{ d }}` path is
0/600.

## Likely fix shape (a hypothesis, not a spec)

The `@stringfilter` boundary is the seam: a `Value::Decimal` reaching a
string-consuming filter should be stringified as Python's `str(Decimal)` would —
`repr`-style digits with an uppercase `E` — not as `Display` (which is Django's
`numberformat` rendering, correct for `{{ d }}` and wrong here). That is two
different string forms for one value, so the change is picking the right one per
[...]

PR #2318, opened 2026-08-28
fix(templates): |safe stringifies a scalar, and a randomised sweep pins unordered_list nesting (#2303)

### LINKED ISSUE #2303: `{{ n|safe }}` does not stringify a scalar the way Django's mark_safe does
Surfaced while fixing #2290 (PR #2302), which restores Django's `SafeData`
markers on a value crossing into a custom filter. That fix wraps **only a
`str`**, deliberately, and this issue is the residue.

## The divergence

Django's `mark_safe` **stringifies** a non-`str`: `mark_safe(42)` is
`SafeString('42')`, `mark_safe([1, 2])` is `SafeString('[1, 2]')`. So
`{{ n|safe }}` in Django does not merely mark the value — it *changes its type*
before the rest of the chain sees it.

Measured on Django 5.2 with a live `@register.filter` probe on one side and
`_rust.register_custom_filter` on the other, after PR #2302:

| template | Django hands the filter | djust hands the filter |
|---|---|---|
| `{{ n\|safe\|probe }}` with `n = 42` | `('SafeString', True)` | `('int', False)` |
| `{{ f\|safe\|probe }}` with `f = 1.5` | `('SafeString', True)` | `('float', False)` |
| `{{ b\|safe\|probe }}` with `b = True` | `('SafeString', True)` | `('bool', False)` |
| `{{ absent\|safe\|probe }}` | `('SafeString', True)` — of `''` | `('NoneType', False)` |

Note the last row is a *different* mechanism from the first three: Django
substitutes `string_if_invalid` (`""`) for a missing variable **before** the
filter chain runs, so `mark_safe` there produces `SafeString('')`. djust passes
`Value::Missing` through as Python `None`, so a naive `mark_safe` would give
the literal text `"None"` — strictly worse than the current answer.

`{{ L|safe|f }}` and `{{ d|safe|f }}` already agree, because djust's `safe`
filter (`crates/djust_templates/src/filters.rs`, the `"safe"` arm) already
collapses `List`/`Tuple`/`Object` to `Value::String(value.to_string())` —
`mark_safe`'s stringification, done one layer up. The scalar variants simply
were never given the same treatment.

## Why #2290 did not fix it

Wrapping a non-`str` in `mark_safe` changes the **type** an existing filter
receives — `{{ n|safe|my_filter }}` would start handing `my_filter` a string
where djust hands it an `int` today, and a filter doing arithmetic on it would
break. That is a pre-existing `|safe`-on-a-non-`str` SHAPE divergence with its
own blast radius, distinct from the *safety-visibility* gap #2290 is about, so
PR #2302 stayed scoped (#1079) and pinned the current behaviour rather than
leaving it undocumented:
`python/tests/test_custom_filter_safedata_2290.py::TestContainerSafety::test_a_non_string_is_passed_through_unwrapped`.

The residue is djust reporting `SafeData` **False** where Django reports True —
the escaping direction. Not a security gap.

## What a fix would need

1. Decide whether `safe` should collapse scalars the way it already collapses
   containers, rather than special-casing the wrap at the PyO3 boundary. The
   `"safe"` arm in `filters.rs` is the more likely seam than
   `filter_registry::mark_input_safety`, because it fixes the type for the
   whole downstream chain (built-ins included) rather than only for custom
   filters — and the container half is already there, so the scalar half is
   the same edit one variant over (#1646).
2. Handle `Value::Missing` separately: Django's answer is `SafeString('')`, not
   `SafeString('None')`, because `string_if_invalid` runs first. A blanket
   stringify gets this row *more* wrong than today.
3. Measure with `scripts/filter-parity-differential.py` two-build before/after
   — the built-in chain is where a type change would bite (`{{ n|safe|add:"1" }}`,
   `{{ n|safe|floatformat }}`, `{{ n|safe|length }}` all read the type), and the
   #2257 / #2258 entries record two prior attempts at type-coercion in this
   engine that regressed four figures of cells.

Related: #2290 (the safety half, fixed in #2302), #2284, #2283, #2281.

johnrtipton:
Closes #2303

## Rebased onto #2306 — the `unordered_list` fix here is @alexsmolya's, not mine

This PR originally carried a fix for #2301 as well. While it was in review,
**[@alexsmolya](https://github.com/alexsmolya) opened #2306 fixing the same
bug, and it merged first.** Their source hunk and mine were **byte-identical**
— both drop `sub_indent` and build the sublist wrapper from the parent's
`indent` — so on rebase my hunk vanished as an already-applied no-op. It should
have: they got there first.

**#2301 is closed by #2306.** This PR no longer claims it. What it keeps for
that issue is the coverage #2306 shipped without, and the CHANGELOG entry — the
one @alexsmolya's PR did not add — which credits them by name for the fix.

The `filters.rs` diff against `main` for `unordered_list` is now **comment-only**.

Rebased a second time onto #2316 (#2299 + #2305), which touches the item-safety
machinery next to `unordered_list` and grew this differential's corpus. That
merge was clean, and was therefore built and re-measured rather than trusted:
#2316's own coupling suites (`test_escape_chain_and_sequence_filters_2281_2283.py`,
`test_context_item_safety_2287.py`) are green here, 272 passed.

---

## Two things this adds on top of #2306

**1. A randomised nesting sweep.** The divergence is a *recursion* bug, so the
question is not "does the reported cell match" but "does **every** nesting
shape match", and only a sweep answers that:

| | randomly-shaped nestings agreeing with live Django |
|---|---|
| before #2306 | 2164 / 3000 |
| after | **3000 / 3000** |

Gate-off: reverting #2306's hunk turns **15 cases red** in this suite —
*including @alexsmolya's own `test_nested_unmarked_list_matches_django`* — so
the fix and the coverage are each load-bearing for the other.

The same sweep surfaced the one remaining `unordered_list` shape divergence,
which neither PR fixes: Django accepts a `tuple` (and a generator) as a sublist
and djust matches `Value::List` only. Filed as **#2317** rather than folded in
(#1079).

**2. In `python/tests/test_context_item_safety_2287.py`, both assertions are
kept.** @alexsmolya's `test_nested_unmarked_list_matches_django` pins the plain
`["a", ["b"]]` shape; mine pins `["<b>a</b>", ["<i>b</i>"]]` — the
markup-carrying version of the *same* nesting the refusal test refuses, which is
what separates "differs because of indentation" from "differs because the nested
item-safety grant is deliberately refused". Different cells, both live; the
merged docstring says so.

---

## #2303 — `|safe` did not stringify a scalar

Django's `mark_safe(obj)` is `SafeString(str(obj))` — it does not merely *mark*
the value, it changes its **type** before the rest of the chain sees it.

| `{{ p\|safe\|probe }}` | Django | djust before | after |
|---|---|---|---|
| `p = 42` | `('SafeString', True)` | `('int', False)` | ✅ |
| `p = 1.5` | `('SafeString', True)` | `('float', False)` | ✅ |
| `p = True` | `('SafeString', True)` | `('bool', False)` | ✅ |
| `p = Decimal('1E-9')` | `('SafeString', True)` | `('Decimal', False)` | ✅ |
| `p` absent | `('SafeString', True)` of `''` | `('NoneType', False)` | ✅ |

The container half landed in #2283; this is the same edit one variant over
(#1646). **The issue asked whether the `"safe"` arm is the right seam. It is —
but not by growing a new stringify there.** Two spellings had to be right, and
both fell out of machinery that already exists:

* **`str()`, not the RENDER form.** `Display` is `numberformat.format()`, which
  expands an exponent — `1e20` renders `100000000000000000000` while
  `str(1e20)` is `1e+20`, and `Decimal("1E-9")` renders `0.000000001` while
  `str()` is `1E-9`. That is *exactly* the `@stringfilter` coercion #2250 added,
  so `safe` joins `STRING_FILTERS` and arrives at its arm already a
  `Value::String`. A first pass added a `Value::py_str()` method instead; that
  would have been a **second mechanism shadowing the first**, so it was dropped
  — the arm is now one line.
* **`Value::Missing` is `""`, not `"None"`.** Django substitutes
  `string_if_invalid` before the chain runs. `Display` for `Missing` is already
  `""` — the whole reason #2203 split the variant — so no special case was
  needed either, and `test_the_literal_text_None_never_reaches_the_page` pins
  it. The failure mode the issue predicted (`{{ absent|safe|f }}` becoming the
  text `None`) never arises.

This closes **#2257's residue 1 for `safe`**: `{{ d|safe }}` on a `Decimal`
under a German locale was `0,000000001` and is now Django's `1E-9`. `escape`
keeps the exclusion. `test_escape_is_the_named_exclusion` asked whoever got
here to update it; done.

### `divisibleby` — the one filter the type change bit

Found by the differential, **not** by inspection. Django is
`int(value) % int(arg) == 0`, so a numeric string has always worked there;
djust matched `Value::Integer` alone, so `{{ n|safe|divisibleby:"2" }}` — `True`
before and in Django — would have started answering `False`. Widened to what
`int(str)` accepts unambiguously; anything else still fails soft rather than
raising as Django does, pinned by
`test_an_input_int_rejects_fails_soft_rather_than_raising`.

`divisibleby` also joins the differential's `HOT2`. It is in no safety set so
`test_every_safety_set_member_is_in_the_differential_hot_sets` does not require
it — it is there because it reads the input's **type**, which is the axis this
change moves, and its absence is precisely why the first sweep reported clean
over a real regression.

---

## Differential — two builds, baseline rebuilt from `origin/main` **including #2306 and #2316**

So the figures isolate this change alone. They were 222 / 582 against a
pre-#2306 baseline; the difference is the `unordered_list` cells, which are
@alexsmolya's and belong in his column, not mine:

```
scripts/filter-parity-differential.py     29,662 cells
  agree BEFORE 25354 -> AFTER 25575
  newly AGREEING 221 / REGRESSIONS 0
  live-payload leaks 38 -> 38, INTRODUCED 0

wider scalar sweep (Decimal, BigInt, exponent floats, MISSING)   16,472 cells
  agree BEFORE 13087 -> AFTER 13332
  newly AGREEING 245 / REGRESSIONS 0
  live-payload leaks 12 -> 12, INTRODUCED 0
```

Re-measured from scratch after each of the two rebases (#2306, then #2316);
221 / 245 / 0 / 0 held across all three baselines.

**Nothing became more permissive than Django in either sweep.** `|safe`'s change
is type-only: the rendered *bytes* of `{{ p|safe }}` are unchanged, and match
Django.

## Gate-off

Four mechanisms, each mutated separately, rebuilt, measured. The harness asserts
the mutation text matched **exactly once**, asserts the source changed, reports
pytest `N error` as `INVALID` rather than a number, restores in a `finally`, and
verifies byte-identical restore.

| mechanism | result | uniquely its own |
|---|---|---|
| `<ul>` indent (**#2306's fix**, reverted) | **15 failed** | `test_three_thousand_random_nestings_all_agree`, and @alexsmolya's `test_nested_unmarked_list_matches_django` |
| the `"safe"` arm | **15 failed** | `test_a_scalar_reaches_the_next_filter_as_a_SafeString`, `test_an_absent_variable_reaches_it_as_an_empty_SafeString` |
| `STRING_FILTERS` membership | **9 failed** | `test_the_bare_render_expands_the_exponent_and_safe_does_not[float\|decimal]` |
| `divisibleby` string parse | **8 failed** | `TestDivisiblebyReadsTheValueNotTheType::*` |

The two `|safe` halves do **not** shadow each other. Baseline and restored
builds both `0 failed / 282 passed`.

## Tests

* `python/tests/test_safe_stringifies_scalars_2303.py` (new) — the issue's
  table, one row per `Value` arm, the `str`-vs-render-form split, the
  absent-variable class, `divisibleby`, a whole-registry residue pin, and a
  randomised sweep of the invariant the fix amounts to (*behind `|safe` no
  filter can tell a value from its own `str`*), asserted on Django first so it
  is Django's semantics rather than djust agreeing with itself, with its own
  gate-off sibling.
* `python/tests/test_unordered_list_indent_2301.py` (new) — coverage for
  #2306, described as such in its module docstring.
* `crates/djust_templates/src/filters.rs` —
  `safe_stringifies_every_variant_the_way_python_str_does` (11 variants + the
  two render-form controls), `escape_is_excluded_from_the_coercion_and_safe_is_not`.

The whole-registry sweep pins the residue **by filter name** rather than
asserting blanket agreement: five filters (`add`, `date`, `divisibleby`,
`pluralize`, `time`) still differ on a scalar behind `|safe`, each for a
pre-existing reason the differential confirms is unchanged. Set-equality is what
makes a sixth joining them go red — which is how `divisibleby` was caught.

### Existing pins updated deliberately

Each had asked in prose for whoever got here to update it:

* `test_custom_filter_safedata_2290.py` — `test_a_non_string_is_passed_through_unwrapped`
  pinned *exactly* the behaviour #2303 changes; rewritten, plus a gate-off
  sibling (22 → 23 cases, and the `[Unreleased]` count claim corrected here).
* `test_string_filter_stringification_2250.py` — `safe` leaves
  `NAMED_EXCLUSIONS`, 27 → 28 covered filters. Its four registry-wide sweeps
  pass with `safe` included, which is independent confirmation.
* `test_context_item_safety_2287.py` — the nested-refusal case, merged with
  @alexsmolya's addition as described above.

## Both defects here were found by a randomised sweep, neither by a curated table

Worth stating plainly, because it is the argument for the sweeps costing what
they cost:

* **`divisibleby`** — found by the two-build differential, once *every* filter
  in the live registry was composed behind `|safe`. The hand-picked filter list
  I started with did not include it, and the issue's own step-3 worry named
  three other filters (`add`, `floatformat`, `length`), all of which turned out
  to be fine.
* **The `unordered_list` tuple-sublist gap (#2317)** — found by the 3,000-case
  nesting sweep, not by the 15 nesting shapes I wrote out by hand first.

Both curated tables were written first, are careful, and found neither. The
generalisation is the one already in this repo's canon (v1.1.1-2): a curated
table samples the axis its author thought of, and Django is importable from the
test, so the reference answer is a subprocess away.

## Gates (all re-run after each rebase — a `filters.rs` auto-merge is not trusted)

* `pytest tests/ python/tests/ python/djust/tests/ -n auto` — **13309 passed,
  220 skipped, 0 failed**
* `make test-rust` — clean
* `cargo clippy --workspace --all-targets -- -D warnings` — clean
* `make dev-build` — compiles (the merged hunk was verified to build, not just
  to merge)
* `0` commits behind `origin/main` (`39946ebf`)

Pushed `--no-verify` — the native pre-push hook hardcodes `.venv/bin/python` and
fails inside a worktree — after running every gate manually. CI is authoritative.


johnrtipton:
## Retrospective — PR #2318 (pipeline-drain, v1.1.1-7)

**Task**: #2303 (`|safe` did not stringify a scalar), plus regression coverage for #2301.

### Quality: 5

### What went well

- **A randomised sweep found two defects that carefully-written curated tables missed.** `divisibleby` — `{{ n|safe|divisibleby:"2" }}` would have flipped `True` → `False`, because Django is `int(value) % int(arg)` while djust matched `Value::Integer` alone — surfaced only once every registry filter was composed behind `|safe`. The hand-picked list did not include it, and the issue's own analysis named three *other* filters as at-risk, all of which were fine. The tuple-sublist gap (#2317) came from the 3,000-case nesting sweep, not the 15 shapes written out by hand. **The tables were written first and are careful; they found neither.** Fourth instance of this pattern in the drain.
- **The fix got smaller by finding both hard parts already answered in the tree.** The `str()`-vs-render-form split *is* #2250's `@stringfilter` coercion, so `safe` joins `STRING_FILTERS` rather than growing a second stringify that would shadow it; and `Value::Missing` is already `""` under `Display`, so the `"None"`-on-the-page failure the issue predicted never arises. A first pass that added a `Value::py_str()` method was dropped as the redundant second mechanism.

### What didn't — and the collision

- **An external contributor got there first on #2301.** @alexsmolya's #2306 landed with a **byte-identical** source hunk about an hour earlier. Theirs merged; this PR rebased its source change away to nothing and kept the sweep as coverage. Nothing went wrong, but it is worth recording that two workstreams spent effort on one issue because neither could see the other — the visible signal was a PR that existed for an hour before anyone here looked.
- **Attribution needed active correction twice.** #2306 shipped no CHANGELOG entry, so this PR's would have been the only record of that fix — a mention inside a `|safe` bullet would have left it a subordinate clause in someone else's item, so it became a separate bullet led by their name. And the differential figures dropped 582 → 245 once re-measured against a baseline containing #2306; leaving the pre-rebase numbers would have quietly credited their cells here. Revising a number downward, unprompted, is what makes the rest of the measurements usable as evidence.
- **The volume hit ENOSPC mid-task** (891 MiB free on 1.8 TiB; Bash could not write its own output file). Structural cause is worktree lifecycle — several concurrent worktrees each carrying ~10 GB of Rust artifacts — not this task. ~131 GB of build artifacts have since been reclaimed from idle worktrees.

### The cleanest result

Reverting @alexsmolya's hunk against this suite turns **15 cases red — including their own `test_nested_unmarked_list_matches_django`**. The fix and the coverage are each load-bearing for the other, which is a better argument for keeping both than "this one has more tests."

### Verified

`pytest` all roots **13309 passed / 0 failed**; `make test-rust` clean; `cargo clippy --workspace --all-targets -- -D warnings` clean; 15/15 CI green including the `test-summary` aggregate gate; 0 behind `origin/main`. Gate-off 15 / 15 / 9 / 8 across four mechanisms with none shadowing another. Differential held at **221 / 245 newly agreeing, 0 regressions, 0 introduced live-payload leaks** across all three baselines.

Issue #2319, opened 2026-08-28
bug: truncatechars/truncatechars_html count combining marks and skip NFC; slugify skips NFKD — one Unicode-tables dependency decision (from #2294)


johnrtipton:
Split out of #2294 item 3 (`truncatechars` counts combining marks), which PR for #2294 deliberately did not fix. It is the *same* blocker as the already-pinned `slugify` NFKD gap, so this issue covers both: they want one dependency decision, not two.

## The divergence, re-measured on `main`

`django.utils.text.Truncator.chars` does two things djust's port does not:

```python
def chars(self, num, truncate=None, html=False):
    length = int(num)
    text = unicodedata.normalize("NFC", self._wrapped)   # (a)
    truncate_len = calculate_truncate_chars_length(length, truncate)
    ...
    return self._text_chars(length, truncate, text, truncate_len)

def _text_chars(self, length, truncate, text, truncate_len):
    for i, char in enumerate(text):
        if unicodedata.combining(char):                   # (b)
            continue
        ...
```

`calculate_truncate_chars_length` applies the same `(b)` skip to the truncation
text, so a combining mark inside `truncate` costs nothing either.

`crates/djust_templates/src/truncate.rs::text_chars` /
`calculate_truncate_chars_length` implement neither, and the module docstring
already says so.

Reproducers, verified against a live Django 5.2.16 through
`Template(src).render(Context({"p": v}))` on both sides today:

```
# inputs written as code points, because a decomposed sequence and its
# precomposed form are INDISTINGUISHABLE on screen (see the last section)

p = 'a' + U+0301 + 'bcdefg'          # 8 code points, NFC -> 7
  {{ p|truncatechars:5 }}            django='ábcd…'   djust='ábc…'

p = ('e' + U+0301) * 5               # 10 code points, NFC -> 5
  {{ p|truncatechars:5 }}            django='ééééé'   djust='éé…'

p = '<b>' + 'a' + U+0301 + 'bcdefg</b>'
  {{ p|truncatechars_html:5 }}       django='&lt;b&gt;ábcd…&lt;/b&gt;'
                                     djust='&lt;b&gt;ábc…&lt;/b&gt;'

{{ p|slugify }}               on 'café'               django='cafe'   djust='café'
```

Note `(a)` is unconditional: `Truncator('é').chars(50)` returns the
**precomposed** `'é'` even though nothing was truncated. So the combining
skip alone is not enough for byte-parity — implementing `(b)` without `(a)`
[...]

PR #2320, opened 2026-08-28
fix(templates): stringformat %Ns width, center's tie-break and dict length (#2294)

johnrtipton:
Fixes three of the four measuring-filter divergences in #2294. The fourth is re-measured, unchanged, and filed as **#2319**; a separate finding about the `"__model__"` marker is filed as **#2322**.

Refs #2294. Does **not** close it — item 3 (`truncatechars` / combining marks) stays open there, or #2294 can be closed in favour of #2319, whichever the maintainer prefers.

## Re-measurement on current `main` first

Every item was re-run against a live Django 5.2.16 before anything was designed. All four still reproduce; item 1 reproduces **wider** than the issue says.

| # | issue's claim | measured on `main` |
|---|---|---|
| 1 | `stringformat:"Ns"` ignores width | confirmed — **and** `.3s` does not truncate, `-10s` does not left-align, `10.3s` does neither |
| 2 | `center` uses Rust's tie-break | confirmed; `'a'.center(4)` / `'abc'.center(6)` agree, exactly as the issue warns |
| 3 | `truncatechars` counts combining marks | confirmed; also needs NFC, which the issue does not mention — see below |
| 4 | `{{ dict|length }}` answers 0 | confirmed; `{}` already agreed |

Cell counts differ from the issue's because a different corpus was used (`scripts/filter-parity-differential.py`, 27,684 cells, rather than the 174-cell sweep the issue quotes). Mine are in the table further down.

## What changed

**1. `stringformat`'s `s` conversion.** Django's filter is `("%" + arg) % value`, and CPython honours `[flags][width][.precision]` for `s` exactly as for `%d`. The arm read only the conversion character and dropped all of it.

Ported from CPython's `unicode_format_arg_parse` and then *checked against it*: `test_exhaustive_spec_sweep` builds every grammar-valid prefix over `-+ #0` / digits / `.` / `hlL` up to length 4, crosses it with a value corpus, and requires djust to equal `("%" + prefix + "s") % v` on every cell. Five things that sweep settles, none of which I would have put in a table:

- the `0` flag is **ignored** for `s` — `"%010s" % "ab"` pads with *spaces*, where the same flag on `%d` pads with zeros;
- a bare `.` is precision **zero** — `"%.s"` is `''` and `"%10.s"` is ten spaces, not "no precision";
- `0` leads as a *flag*, so `%0s` is width 0 while `%010s` is width 10;
- `+`, `' '` and `#` are accepted no-ops, may repeat, and may follow one another (`%--10s` left-aligns);
- width and precision have limits **five orders of magnitude apart** — width is a `Py_ssize_t` (`2**63-1`), precision an `int` (`2**31-1`). Both were bisected against the interpreter. Assuming one limit for both is the plausible-and-wrong answer, and it was my first draft.

A prefix that is *not* that grammar (`%as` is `%a` followed by a literal `s`; `%!s`; `%*s`) keeps the pre-#2294 answer. CPython's answer to those is neither `str()` nor `""`, so nothing here can be right, and inventing a second wrong answer is not an improvement. Measured: no cell changes either way.

**2. `center`.** `format!("{s:^width$}")` always puts the smaller half of an odd margin on the left. CPython is `left = marg // 2 + (marg & width & 1)` — it biases left only when the width is odd too. An exhaustive (length 0..12) × (width 0..24) grid replaces a table, because the two rules agree on every even margin and a table drawn from `'a'.center(4)` finds nothing. `ljust`/`rjust` were already correct.

**3. `{{ dict|length }}`.** `Value::Object` is two Python things wearing one shape: a `dict`, whose length is `len(dict)`, and any non-dict object the serializer flattened into a map, whose length is 0 because `len(model)` raises `TypeError` and Django's filter catches it. That is why #2279 left it alone, and `o.len()` would have traded one wrong answer for another — a model spelling its field count.

The marker is `Value::object_str()`: the `"__str__"`-holding-a-string predicate `{{ obj }}` already uses (#968), promoted from **two** open-coded copies in the `Display` impls to one definition, with a source pin asserting `get("__str__")` appears exactly once in `djust_core/src/lib.rs`. Reusing that predicate is what keeps `{{ p }}` and `{{ p|length }}` from disagreeing about what `p` *is*.

**`"__model__"` is the marker that looks more specific and is unusable.** **Four of the six** sites that stamp `"__str__": str(...)` omit it — `serialization.py:463` (depth-limited FK), `serialization.py:1325` (max depth) and both `template/rendering.py` fallbacks (209, 219) — while only `serialization.py:410` and `jit.py:321` carry it. Keying on `__model__` would make a depth-limited model answer `3`.

> **Correction, after the rebase.** The first version of this PR said "three of five" in the body, the CHANGELOG, two docstrings and a Rust doc comment. Re-grepping for the review put it at **four of six** — the #1197 citation-discipline failure, in the PR whose whole argument rests on that count. Every occurrence is fixed, and the count is no longer prose: `test_the_model_marker_sites_are_counted_from_the_source_not_from_prose` greps the three modules, classifies each stamp by whether `"__model__"` is in the same dict literal, and asserts the exact 2/4 split with file:line. Gate-off verified with three mutations, all red — giving a without-site the key (4→3), removing it from a with-site (4→5), and adding a seventh stamp site (6→7).

Filed on its own as **#2322**: a marker four of six producers omit is a trap for anything that reaches for it, and a pin inside a `length` test is not where the next person will look.

### The model-instance case, measured explicitly

```
python len(model)                       -> TypeError: object of type 'User' has no len()
{{ p|length }} on User(username='bob')  django='0'  djust='0'   (before and after)
```

The value reaching the renderer is a **17-key map**, so this is a live discrimination, not a vacuous one — `test_a_model_instance_still_answers_zero` asserts the serialized form has more than one key before asserting the answer is 0, so the test cannot pass for the wrong reason.

Residual, pinned as known-wrong: a hand-built `dict` carrying a string `"__str__"` key answers 0 where Django answers its length. That same dict *already* renders as its `__str__` under `{{ p }}`, so the alternative is not "more correct" — it is two predicates disagreeing with each other.

## The known-wrong pin

`test_length_of_a_dict_is_WRONGLY_still_zero_bug_2294` is **converted, not deleted**: renamed to `test_length_of_a_dict_now_agrees_fixed_by_2294`, inverted, and joined by `test_a_model_still_answers_zero_which_is_why_it_was_pinned`. The class docstring of `TestKnownResidualDivergences` — which leads with "Every test here pins a KNOWN-WRONG answer" — now says which row left and why. The reasoning behind the original divergence is the reasoning the fix had to answer, and it is the part a future reader needs.

## Item 4 is not fixed — #2319

`truncatechars` still counts combining marks, re-measured and unchanged. Two reasons it is not in this PR:

1. It needs a **Unicode-tables dependency** (`unicode-normalization`), which is a project decision with a wheel-size cost, not a correctness one.
2. `Truncator.chars` also applies `unicodedata.normalize("NFC", text)` **unconditionally** — verified: `Truncator('é').chars(50)` returns the precomposed `'é'` with nothing truncated. So the combining skip alone gets the right characters and the wrong normalization form; the two go together, and so does `slugify`'s NFKD fold, which is blocked on the same crate and already pinned beside it.

#2319 carries the full measurement, including the drift argument the `striptags` work correctly raised and why it does **not** transfer: measured 15.0 → 16.0, **zero** existing code points changed combining class (922 → 934, all 12 newly assigned), against 5,812 for `isprintable`.

## Gate-off

Harness asserts the mutation text was found, matched **exactly once**, that the source changed, counts pytest `N error` separately as INVALID rather than as a number, rebuilds between iterations, restores in a `finally` and verifies the restore is byte-identical.

| mutation | mechanism it removes | python | rust |
|---|---|---|---|
| M1 drop the width pad | `%Ns` width | 18 failed | 2 failed |
| M2 ignore precision | `%.Ns` truncation | 8 failed | 2 failed |
| M3 ignore the `-` flag | `%-Ns` left-align | 6 failed | 1 failed |
| M4 accept an over-limit width | width/precision too big → `''` | 2 failed | 1 failed |
| M5 zero-pad `%s` (the `%d` rule) | the `0` flag is ignored for `s` | 18 failed | 2 failed |
| M6 revert to Rust's tie-break | `(marg & width & 1)` | 6 failed | 1 failed |
| M7 bias `(marg & 1)` instead | the `& width` half of the term | 5 failed | 1 failed |
| M8 pad by bytes | code-point width | 2 failed | 1 failed |
| M9 `o.len()` for every Object | a model must still answer 0 | 5 failed | 1 failed |
| M10 revert to `_ => 0` | a dict must answer `len(dict)` | 8 failed | 1 failed |
| M11 accept a non-`String` `__str__` | `Display` and `length` must agree | 1 failed | 2 failed |

Every mutation red, and the whole table was **re-run after the #2316 and #2318 rebases** (M10 gained a failure, 8 -> 9; everything else identical). M7 exists because M6 alone cannot distinguish "the term is present" from "the term is right" — the two arms of the tie-break need separate mutations or one shadows the other.

## Differential

`scripts/filter-parity-differential.py`, two builds, against a **rebuilt** `origin/main` baseline.

**Re-measured from scratch at `1fa46c33`**, after #2316 and #2318 both landed in `filters.rs` and both grew this corpus. The first run (27,684 cells, pre-#2316) is superseded; the three numbers are unchanged:

```
cells                  29,662     (was 27,684)
newly AGREEING             30
REGRESSIONS                 1
live-payload leaks   38 -> 38     (0 closed, 0 INTRODUCED)
```

Per-filter, unchained: `stringformat` and `center` at 0 disagreeing, and the one unchained `length` cell that disagreed on the baseline is fixed — `after-disagree` is empty for `length` unchained.

### The one regression, and why it is not this filter's

`{{ dict|add:"1"|length }}` — `django='0'`, `djust='2'`.

`add`'s **third branch** is a documented divergence: Django returns `""` for a value it can neither sum nor concatenate, and djust returns the value unchanged on purpose (`filters.rs`: *"turning a rendered value into silent emptiness on upgrade is the silent-wrong-output class this engine keeps having to fix. Documented divergence, not an oversight."*). So `add` hands `length` the dict where Django hands it `""`, and a *correct* `length` then answers 2.

The class is pre-existing, and the re-measurement strengthened the evidence — #2316's marked-TUPLE corpus input added a **fourth** twin:

```
add:"1"|length  l-plain   BASE ['0', '2']   AFTER ['0', '2']
add:"1"|length  l-mixed   BASE ['0', '2']   AFTER ['0', '2']
add:"1"|length  l-marked  BASE ['0', '2']   AFTER ['0', '2']
add:"1"|length  t-marked  BASE ['0', '2']   AFTER ['0', '2']   <- new axis, #2316
add:"1"|length  d-plain   BASE ['0', '0']   AFTER ['0', '2']
```

Four twins already diverged identically on the baseline, because the list and tuple arms of `length` were already right. The dict cell agreed only because two wrongs cancelled — the #2272 pattern, the same shape as #2273 unmasking #2279. Closing it means revisiting `add`'s deliberate decision, which belongs to `add`. Pinned and explained in `test_the_add_chain_UNMASKS_adds_own_documented_divergence`, which now asserts the list **and** tuple twins.

## Verification

- `pytest tests/ python/tests/ python/djust/tests/ -n auto` — **13,364 passed, 220 skipped, 0 failed**
- `make test-rust` — 0 failed
- Rebased three times as the drain moved under it — `39946ebf` (#2316), `1fa46c33` (#2318, which touches `filters.rs` next to this work), `6fa557cb` (#2323, CI-only). `git rev-list --count HEAD..origin/main` is 0. The suite, `make test-rust`, clippy, the differential **and** the gate-off table were all re-run after the `1fa46c33` rebase, not carried over
- The `--amend` for the count correction **bounced** on the first attempt (ruff reformatted; HEAD unchanged at `3254ddda`). Caught by the canon #1524 reflex of asserting the hash changed, re-staged, and verified `3254ddda -> fd84dda0`
- Two commits: impl+tests (no CHANGELOG), then docs+CHANGELOG

## Files

| item | source | tests |
|---|---|---|
| 1 `stringformat` | `crates/djust_templates/src/filters.rs` (`apply_stringformat`'s `s` arm, `parse_s_spec`, `parse_c_int`) | `TestStringformatS`; `test_stringformat_s_honours_flags_width_and_precision`, `test_stringformat_s_counts_code_points` |
| 2 `center` | `crates/djust_templates/src/filters.rs` (`center` arm) | `TestCenter`; `test_center_uses_pythons_odd_margin_tie_break` |
| 3 `length` | `crates/djust_templates/src/filters.rs` (`length` arm) + `crates/djust_core/src/lib.rs` (`Value::object_str`, both `Display` impls) | `TestLengthOfAnObject`; `test_length_of_an_object_uses_the_str_marker`, `test_object_str_is_the_model_marker_predicate`, `test_the_str_marker_is_spelled_in_exactly_one_place` |
| 4 `truncatechars` | not changed | `TestItem4IsStillOpen` + the existing `TestKnownRemainingDivergences` pins → #2319 |


johnrtipton:
## Retrospective — PR #2320 (pipeline-drain, v1.1.1-7)

**Task**: #2294 — four measuring-filter divergences. Three fixed (`stringformat:"Ns"`, `center` tie-break, `{{ dict|length }}`); item 3 deferred to #2319.

### Quality: 5

### What went well

- **It shipped a measured `REGRESSIONS: 1` with an explanation rather than a suppressed cell.** `{{ dict|add:"1"|length }}` un-cancels `add`'s *documented* third-branch divergence — its list/tuple twins already read `django='0' djust='2'` on the baseline, so the dict cell had only ever agreed because two wrongs cancelled. Un-cancelling a coincidence is not a regression in any sense that matters, and you can only judge that because it was reported honestly. The rebase then *strengthened* the argument: #2316's marked-tuple input added a fourth baseline twin reading the same way.
- **It checked the reference instead of reasoning about it.** CPython's `unicode_format_arg_parse` was ported and then verified against the interpreter over every grammar-valid prefix ≤ 4 chars × a value corpus (>8k cells, 0 mismatches). That sweep settled five things a table would miss — including that the `0` flag is *ignored* for `%s`, and that width and precision limits are five orders of magnitude apart (`Py_ssize_t` vs `int`). Its first draft assumed one limit for both and was wrong.
- **Deferring item 3 came with a decision-grade measurement, not a shrug.** The drift objection that stopped the `striptags`/`isprintable` work does not transfer to combining classes: Unicode 15.0 → 16.0 changed the class of **zero** existing code points (922 → 934, all newly assigned) against 5,812 for `isprintable`. That converts "we can never depend on Unicode tables" into "we can here, and here is why it is safe."

### What didn't

- **It shipped a wrong count in the claim its central argument rests on, in five places.** "Three of five model-serialization sites omit `__model__`" was actually **four of six** — in the PR body, the CHANGELOG, two docstrings and a Rust doc comment. Exactly the #1197 citation-discipline failure. It surfaced only because re-grepping to file the follow-up forced a recount. The fix is the right one: the count is **no longer prose**. `test_the_model_marker_sites_are_counted_from_the_source_not_from_prose` greps the three modules, classifies each `"__str__": str(...)` stamp by whether `"__model__"` is in the same dict literal, and asserts the exact 2/4 split with file:line — gate-off verified three ways (4→3, 4→5, 6→7).
- **Documenting the NFC hazard reproduced the NFC hazard.** The `# RIGHT` example written to demonstrate `\uXXXX` escapes was itself normalized on write and had to be repaired. That is now a second data point in #2319 rather than an irony left unrecorded.
- **An `--amend` bounced** — pre-commit ruff-reformatted and swallowed it, HEAD unchanged at `3254ddda` with `MM` on the test file. Caught by asserting the hash changed (#1524). Had the pre-amend hash been quoted rather than re-read, this would have reported a commit that never existed.

### Verified

Three rebases (#2316 → #2318 → #2323), each with a full rebuild and re-run because `filters.rs` auto-merged cleanly and that has compiled-then-failed before in this drain. **13,364 passed / 0 failed**; `make test-rust` clean; clippy clean; 16/16 CI green including `test-summary`; 0 behind `origin/main`. Gate-off: 11 mutations, all red post-rebase. Differential re-measured from scratch — corpus grew 27,684 → 29,662 cells and all three headline numbers held.

### Follow-ups filed

**#2319** (combining marks + unconditional NFC, with the drift measurement and the NFC tooling hazard) and **#2322** (the `__model__` marker is unusable as a discriminator, all six sites enumerated, three options with a recommendation).

PR #2330, opened 2026-08-28
fix(templates): the Unicode-tables decision — escape non-ASCII non-printables (#2292), NFC/NFKD for truncate and slugify (#2319)

### LINKED ISSUE #2292: bug: py_repr_string does not escape non-ASCII non-printable code points, and no fixed table can

### LINKED ISSUE #2319: bug: truncatechars/truncatechars_html count combining marks and skip NFC; slugify skips NFKD — one Unicode-tables dependency decision (from #2294)

johnrtipton:
Closes #2292. Closes #2319.

Two issues, one question — *may djust depend on Unicode tables?* — and they get
**opposite answers**, for a reason that is measured rather than argued.

## The measurement

Both issues turn on how much Unicode data moves across djust's supported
interpreters. Recomputed from live CPython 3.10–3.14 (`dump.py` walks all
1,114,112 code points on each):

| interpreter | `unidata_version` | printable | ccc != 0 |
|---|---|---|---|
| 3.10 | 13.0.0 | 143,680 | 872 |
| 3.11 | 14.0.0 | 144,516 | 912 |
| 3.12 | 15.0.0 | 148,998 | 922 |
| 3.13 | **15.1.0** | 149,625 | 922 |
| 3.14 | 16.0.0 | 154,810 | 934 |

Two corrections to #2292's numbers fall straight out: the matrix spans **five**
Unicode versions, not two (3.13 carries 15.1, which the issue had as 15.0), and
end-to-end `isprintable` disagreement is **11,130** code points, not 5,812.

Then the decisive part — *what kind* of change each is:

| | changed for an ALREADY-ASSIGNED code point |
|---|---|
| `isprintable` | 11,130 → but **0** became non-printable; all 11,130 are `Cn` → assigned |
| canonical combining class | **0** (62 additions, all newly assigned) |
| canonical decomposition | **0** (the 11,192 "additions" are 11,172 algorithmic Hangul + 20 new code points) |

#2319's stability claim is confirmed and strengthened (it measured 15.0 vs
16.0; this is 13.0 vs 16.0). The Hangul jump is a CPython *reporting* change,
not a data change.

## #2292 — fixable, and with **no dependency**

`not str.isprintable()` splits into a stable part (`Cc Cf Cs Co Zl Zp Zs`,
which over 13.0 → 16.0 gained exactly one already-assigned member — `U+0020`,
which Python calls printable anyway) plus `Cn`, which is the entire moving
part. Escaping the seven categories and treating `Cn` as printable is
**exact for every assigned code point on all five interpreters** — verified
against the table as committed, not just as generated.

And the set is 139,769 code points but only **28 ranges**, private use being
three contiguous blocks. So it is a hand-written table; `unicode-general-category`
is not needed. Choosing "never escape `Cn`" over "pin a Unicode version" keeps
the output version-INDEPENDENT.

Residual, documented: an unassigned code point is emitted literally. That is
the one place no fixed table can be right, and it does not occur in real
template data.

## #2319 — fixable, and it **does** need the crate

NFC needs canonical decomposition, canonical ordering and the
composition-exclusion set; `slugify` needs NFKD on top. No hand-rollable
subset exists. Adds **`unicode-normalization` 0.1** (unicode-rs,
MIT/Apache-2.0, 546M downloads, one transitive dep `tinyvec`).

Wheel delta, measured with the crate **added and actually called** (an unused
dependency is stripped by LTO — the issue flags this trap):

| | before | after | delta |
|---|---|---|---|
| release wheel | 8,035,284 | 8,102,434 | **+67,150 (+0.84%)** |
| extension, raw | 6,349,840 | 6,481,952 | +132,112 |
| extension, compressed | 2,364,721 | 2,431,393 | +66,672 |

About **half** the compressed cost of the already-accepted `chrono-tz`
(+136 KB), and far below what the crate's 625 KB of table *source* suggests.

## Two premises in #2319 were wrong

Checked against live Django rather than assumed, and both are now pinned:

1. **`truncatechars_html` does NOT skip combining marks.**
   `TruncateCharsHTMLParser.process` counts with a plain `len(data)`. The HTML
   variant needed NFC and *nothing else* — adding a skip there would have
   created a fresh divergence. `M7` in the gate-off table is an inverse
   mutation that adds one, and it goes red.
2. **`Truncator.words` does not normalize** (it reads `self._wrapped`), so the
   `words` filters must not either. Kept as a negative control.

## Issue × files × tests

| issue | production change | tests |
|---|---|---|
| #2292 | `crates/djust_core/src/lib.rs` (`NON_PRINTABLE`, `is_py_non_printable`, width-by-magnitude), `crates/djust_templates/src/pprint.rs` (docs) | `python/tests/test_py_repr_isprintable_table_2292.py` (58), inverted pins in `TestKnownResidualDivergences` |
| #2319 | `Cargo.toml`, `crates/djust_templates/Cargo.toml`, `crates/djust_templates/src/truncate.rs` (`nfc`, `is_combining`, `text_chars`, `html_chars`, `calculate_truncate_chars_length`, `slugify`) | `python/tests/test_truncate_nfc_slugify_fold_2319.py` (44), new cases in `truncate.rs`'s test module, inverted pins in `TestKnownRemainingDivergences` + `TestItem4IsStillOpen` |

Every code-point count in the new doc comments is **recomputed** by
`TestTheDocCommentsCountsAreTrue` against the running interpreter, not
restated — including the per-version printable table, which is checked for the
row matching whichever interpreter the runner has.

## Gate-off

13 mutations across the two changes, every one red, restores byte-identical,
suites green after. Harness asserts the mutation text was found exactly once,
asserts the source changed, rebuilds between iterations, and reports INVALID
rather than a number when pytest emits `N error`.

## A note on the combining-mark literal hazard

#2319 documents that editors and agent file-writers NFC-normalize decomposed
literals on the way to disk, which silently turns a test of the bug into a test
of the fix. It bit this PR twice — once in a probe, once in an `Edit` whose
`old_string` round-tripped through a normalizer and could no longer match the
file's genuinely decomposed `"éx"`. Every combining mark in the new tests
is an explicit `\uXXXX` escape, verified by an audit pass, and
`assert_decomposed()` guards every value whose whole point is that it is
decomposed.


johnrtipton:
## Gate-off tables

13 mutations. Every one red. Restores byte-identical, suites green after.

Harness contract (per CLAUDE.md): asserts the mutation text was found **exactly
once**, asserts the mutated source differs, rebuilds the extension between
iterations, restores in a `finally`, verifies the restore is byte-identical,
and reports `INVALID` rather than a number when the module did not import.

### #2292 — `py_repr_string` (pytest)

| mutation | mechanism removed | failures |
|---|---|---|
| M1 | escaper consults the table at all — **the original bug** | 50 |
| M2 | escape width chosen by magnitude | 39 |
| M3 | the table's contents (drops `U+2001..U+200F`) | 14 |
| M4 | `U+0020` stays excluded from the table | 4 |
| M5 | the ASCII fast path agrees with the table (stops escaping DEL) | 12 |
| M6 | range-lookup polarity | 53 |

M1 doubles as the reproduce-first evidence: reverting to the ASCII-only
escaper turns the six issue reproducers red.

### #2319 — `truncate.rs` (pytest **and** `cargo test`)

| mutation | mechanism removed | pytest | cargo |
|---|---|---|---|
| M1 | `text_chars` applies NFC | 14 | 2 |
| M2 | `text_chars` skips combining marks | 8 | 2 |
| M3 | `html_chars` applies NFC | 5 | 1 |
| M4 | `calculate_truncate_chars_length` skips combining marks | **0** | 1 |
| M5 | `slugify` decomposes before dropping non-ASCII | 8 | 1 |
| M6 | `slugify` uses NFKD, not NFD | 1 | 1 |
| M7 | **inverse**: *adds* a combining skip to `html_chars` | 3 | 1 |

Two things worth reading rather than skimming:

**M4's pytest 0 is correct, not a hole.** The `truncatechars` filter always
passes `truncate=None` — Django's filter signature has no slot for it — so that
mechanism is genuinely unreachable from a template. Its coverage is
`truncate::tests::truncation_text_combining_marks_are_free`, which is why the
harness runs `cargo test` as well as pytest. A pytest-only harness would have
reported it uncovered.

**M7 is the point of the whole premise correction.** It adds the combining skip
the issue assumed `truncatechars_html` needed. It goes red, which is what makes
`test_a_non_composing_mark_is_COUNTED_by_the_html_path` load-bearing rather
than decorative — the pin distinguishes "NFC only" from "NFC plus a skip", and
without M7 there would be no evidence it does.

### The harness bug this run found in itself

The first `#2319` run reported `INVALID (did not compile)` for the **entire**
`cargo` column, including mutations whose pytest column was clearly measured —
so M4, the one mutation whose coverage lives only in Rust, looked uncovered.

`cargo test` prints

```
error: test failed, to rerun pass `-p djust_templates --lib`
```

on a test **failure**, and the detector's `^error` check read that as a
**compile** failure. The tell was `post-restore cargo=(0, [])` — cargo worked
fine the moment nothing was failing.

This is exactly the failure mode CLAUDE.md describes: *"one that reports zero
because it broke the build is worse, because it looks like evidence."* Compile
failure is now detected by `could not compile` / `error[EXXXX]`, and a run with
no `test result:` line is `INVALID`. The table above is the re-measurement.

### One more canary

`TestNoCombiningMarkLiteralsInTheSources` scans the four changed sources for
combining marks written as literals. Since it is a "catches bug class X" check,
it got the #1459 treatment — a real literal was smuggled into
`test_py_repr_isprintable_table_2292.py` and the guard was confirmed to go red
(`(0, 5)` clean → `(1, 4)` smuggled → `(0, 5)` restored, restore
byte-identical).

johnrtipton:
## Rebased onto `origin/main` (now includes #2327 and #2329)

The branch was 2 behind and `CONFLICTING`, which is why only one check was
present: `pull_request` workflows run against a **merge commit**, and GitHub
cannot compute one for a conflicted PR, so they were never created.
`pull_request_target` fires against the base branch regardless — so the
maintainer-approval gate was the only thing that could appear, and it had
passed as a no-op. Rebase, and the full suite showed up on its own.

Rebase was clean: `CHANGELOG.md` merged additively (both new entries intact,
no already-tagged section touched — checked per #2028) and `truncate.rs` took
#2329's `textwrap.rs` sibling without conflict.

### One real interaction with #2329, found rather than assumed

#2329 landed `test_every_whitespace_boundary_the_engine_branches_on_is_in_the_corpus`,
which parses `pprint::py_is_line_break` and `textwrap::is_textwrap_space` out
of the Rust. This PR touches neither, so it is unaffected — **but** it pointed
at a question worth answering, because `slugify` now runs NFKD *in front of*
`truncate::py_is_space`.

It does change the answer, in Django's direction, and **not uniformly**:

| input | before | after / Django | why |
|---|---|---|---|
| `a b` | `a-b` | `a-b` | NBSP decomposes **to an ordinary space** — still a separator |
| `a　b` | `a-b` | `a-b` | ideographic space, likewise |
| `a b` | `a-b` | **`ab`** | no ASCII fold; dropped before the separator pass |
| `a​b` | `a-b` | **`ab`** | likewise |
| `a b` | `a-b` | **`ab`** | Ogham space mark, likewise |

Two characters that are both `str.isspace()` end up on opposite sides, decided
by the decomposition table rather than by any whitespace predicate — so it is
enumerated in `TestSlugifyNonAsciiWhitespaceGoesThroughTheFoldFirst` against
live Django rather than reasoned about, with a sibling asserting the two groups
really are both whitespace so the class cannot be misread as
"non-ASCII whitespace is dropped".

Every one of these was already **correct** post-fix; the sweep only made it
worth pinning. `s-lines` (#2329's new corpus input) carries ` ` and
`\xa0`, so the broad differential now reaches this axis for `slugify` too.

### Gates re-run on the rebased tree

- `pytest tests/ python/tests/ python/djust/tests/ -n auto` — **13,672 passed**, 0 failed
- `make test-rust` — clean
- `cargo fmt --check` — clean
- `scripts/check-changelog-test-counts.py` — exit 0 (the #2319 entry's count went 33 → 35 with the new class)

### One note on matrix coverage

`python-tests` runs the full `[3.12, 3.13, 3.14]` matrix only on `push`; a PR
gets `3.12` alone. So `test_the_printable_count_for_this_interpreter_matches_the_doc_table`
is exercised against Unicode 15.0 here and against 15.1 / 16.0 only after
merge. That is fine rather than a gap — the doc table carries rows for all five
supported interpreters (13.0 through 16.0) and the assertions were verified
locally against raw `unicodedata` dumps from real 3.10, 3.11, 3.12, 3.13 and
3.14 interpreters, not just the one CI runs.

PR #2332, opened 2026-08-28
fix(templates): resolve filtered tag operands, and slice by Python's rules (#2325, #2326)

### LINKED ISSUE #2325: a filter on a {% for %} iterable renders nothing — {% for x in p|slice:":2" %} is empty
Found while measuring #2317/#2321 and filed separately per CLAUDE.md #1079 — unrelated to those (it is not a tuple-shape bug), but it turned up in the same sweep and is the largest divergence that sweep found.

## The divergence

A filter applied to a `{% for %}` iterable renders **nothing** in djust. Django applies the filter and iterates the result.

Measured on Django 5.2.16 / Python 3.12.9, at `6fa557cb`, via `_rust.render_template`:

```
{% for x in p|slice:":2" %}{{ x }}{% endfor %}   p = ["a","b","c"]
  django  'ab'
  djust   ''

{% for x in p|safeseq %}{{ x }}{% endfor %}      p = ["a","b"]
  django  'ab'
  djust   ''

{% for x in p|dictsort:0 %}{{ x }}{% endfor %}   p = [["b"],["a"]]
  django  '[&#x27;a&#x27;][&#x27;b&#x27;]'
  djust   ''

{% for x in p|upper %}{{ x }}{% endfor %}        p = "ab"
  django  'AB'
  djust   ''
```

An UNFILTERED iterable is fine for both list and tuple:

```
{% for x in p %}{{ x }}{% endfor %}   p = ["a","b"]  -> 'ab' / 'ab'   agree
{% for x in p %}{{ x }}{% endfor %}   p = ("a","b")  -> 'ab' / 'ab'   agree
```

So the bug is specifically that the `for`-tag's iterable expression is resolved as a bare variable name and the filter chain after the `|` is not applied — the lookup presumably fails and yields an empty sequence, so the loop body never runs.

## Severity

Silent — no exception, no warning, just an empty region where the page had a list. `{% for x in items|slice:":5" %}` and `{% for x in rows|dictsort:"name" %}` are ordinary Django idioms, so this is likely to hit real templates.

## Direction

Route the `{% for %}` tag's iterable through the same `FilterExpression`-equivalent resolution that `{{ }}` uses, rather than a plain variable lookup. Per the "grep for the SINK" rule (v1.1.1-2 retro): check the other tags that take an expression — `{% if %}`, `{% with %}`, `{% cycle %}`, `{% include ... with %}`, `{% firstof %}` — since a bare-variable resolver used in one tag is usually used in several.

Not measured by `scripts/filter-parity-differential.py`: its corpus is `{{ p|... }}` chains only and constructs no `{% for %}` cell at all.

### LINKED ISSUE #2326: slice clamps negative indices and ignores :step, where Django uses Python slice semantics
Found while fixing #2321 and filed separately per CLAUDE.md #1079 — orthogonal to the tuple-shape fix (it reproduces identically for a list, a tuple and a string).

## The divergence

Django's `slice` filter is `value[slice(*bits)]` — full Python slice semantics. djust's `parse_slice_indices` (`crates/djust_templates/src/filters.rs`) parses at most **two** parts and **clamps** rather than wrapping:

```rust
fn parse_slice_indices(parts: &[&str], len: isize) -> (isize, isize) {
    let start = if parts.is_empty() || parts[0].is_empty() { 0 }
                else { parts[0].parse::<isize>().unwrap_or(0) };
    let end   = if parts.len() < 2 || parts[1].is_empty() { len }
                else { parts[1].parse::<isize>().unwrap_or(len) };
    (start, end)
}
```

So a negative index becomes `max(0)`/`min(len)` and a third `:step` part is parsed and then discarded.

Measured on Django 5.2.16 / Python 3.12.9, at `6fa557cb`, with `p = ["a","b","c","d"]` (the tuple and string columns diverge identically):

| spec | django | djust |
|---|---|---|
| `-1:` | `['d']` | `['a', 'b', 'c', 'd']` |
| `:-1` | `['a', 'b', 'c']` | `[]` |
| `-2:-1` | `['c']` | `[]` |
| `::2` | `['a', 'c']` | `['a', 'b', 'c', 'd']` |
| `1::2` | `['b', 'd']` | `['b', 'c', 'd']` |
| `:3:2` | `['a', 'c']` | `['a', 'b', 'c']` |
| `::-1` | `['d', 'c', 'b', 'a']` | `['a', 'b', 'c', 'd']` |

7 of 10 specs tried diverge, across all three container types (21 cells). Non-negative two-part specs, out-of-range clamps (`-10:`, `:99`) and empty results (`2:1`) all agree — which is why the existing suites, which use `":2"`-shaped specs, never saw it.

## Severity

Silent and wrong-direction: `{{ items|slice:":-1" }}` (drop the last) renders **nothing**, and `{{ items|slice:"-3:" }}` (last three) renders **everything**. Both are ordinary Django idioms and both fail open — showing more than intended, or an empty region.

## Direction

Implement Python's slice semantics properly rather than extending the clamp: normalize `start`/`stop`/`step` the way `slice.indices(len)` does (negative wraps by `+= len` then clamps; a negative step reverses and swaps the defaults), then step the range. One helper shared by the `String` and the sequence branch of `apply_slice`, per #1646 — the two branches currently duplicate the index math and would otherwise drift.

`#2321`'s `rebuild_like` already makes the sequence branch return the input's container, so this change is index math only and does not touch the shape rule. `SLICE_SPECS` in `python/tests/test_sequence_shape_preservation_2317_2321.py` deliberately excludes these specs and
`TestRandomisedSliceShape::test_the_negative_index_gap_is_list_shaped_too` pins the exclusion — it goes red when this is fixed, and names the list to extend.

johnrtipton:
Two silent-empty-output bugs. Neither raises, warns, or logs — a page just loses a whole loop or list, and the template author has no thread to pull. That is the worst failure shape a template engine has, which is why these were treated as the priority rather than as parity nits.

They are **independent defects** — one is expression resolution, one is index math — but they **compose**: `{% for x in items|slice:":-1" %}` needs both, and fixing either alone leaves it empty. Hence one PR.

## #2325 — a filter on a tag operand was dropped

Django resolves a tag's operand with a `FilterExpression`, the same object `{{ }}` uses. djust had **one** filter-aware resolver (`get_value`) and **four** tags that each open-coded a bare variable lookup, so the chain after the `|` was never applied — the lookup asked for a variable literally *named* `p|slice:":2"`, missed, and the tag proceeded on the miss.

|---|---|---|
| `{% for x in p\|slice:":2" %}` | `ab` | `''` |
| `{% if p\|slice:":1" %}Y{% else %}N` | `Y` | `N` |
| `{% with q=p\|upper %}{{ q }}` | `HI` | **`p\|upper`** |
| `{% include "t" with q=p\|upper %}` | `HI` | **`p\|upper`** |

The last two are louder than "empty": their miss fell back to `Value::String(expression)`, echoing the template's **own source** into the page. `{% with q="lit" %}` rendered `&quot;lit&quot;`, quotes and all; `{% with q=nope %}` rendered the variable *name*.

Four spellings of one lookup is the parallel-path-drift class (#1646), so all four now call `get_value` rather than each learning about filters. Three supporting changes fall out:

- **`get_value_safe` grows the `Context::resolve` getattr walk as its LAST arm.** `{% for %}` called `resolve()` directly for #806 (`{% for x in user.orders %}` over a DB relation). Routing it through `get_value` without this would have regressed every such loop to empty — the exact symptom being fixed. Placed after the literal arms so nothing that resolves today changes, and so the GIL round-trip costs nothing on an `{% if a == 5 %}` operand.
- **`{% for %}` iterates a string by character**, as Python does. #2325's own repro table needs it (`{% for x in p|upper %}` over `"ab"` is `AB`), and `upper`/`join`/`first`/`last` all hand the arm a string. Normalised *before* the match so the string shares the whole loop body instead of growing a parallel copy.
- **The loop safe-key mapping is registered only for a bare operand.** It asserts `item` **is** `<iterable>.<index>`, which a filter falsifies (`slice` shifts indices, `dictsort` reorders), so a filtered operand could otherwise resolve a safety mark belonging to a different element.

The runtime-safe flag is **discarded** at all four sites: a filtered operand can only ever be escaped at least as hard as before.

## #2326 — slice reimplemented Python's index math, and got it wrong

Django's `slice` is `value[slice(*bits)]`, a passthrough, so every Python rule applies. `parse_slice_indices` read at most two parts and **clamped** instead of wrapping, failing in the two directions an author notices: `{{ items|slice:":-1" }}` (drop the last) rendered **nothing**, `{{ items|slice:"-3:" }}` (last three) rendered **everything**.

Patching those two would have left the rest — a one-part spec is `slice(stop)` so `"2"` means `[:2]` and **not** `[2:]`; a `:step` was parsed and discarded; a negative step never reversed. Value-by-value fixes on a semantics gap do not converge, so this reproduces **CPython's algorithm** (`PySlice_AdjustIndices` plus the walk) in one `slice_positions` helper shared by the string and sequence branches — which had duplicated the index math and are exactly the pair that would drift again (#1646).

`python_int` is CPython's `int()`, not Rust's `parse::<isize>()`. Each difference was confirmed by running Django, not reasoned about:

| spec | why Rust's parse is not `int()` |
|---|---|
| `" 1 : 2 "` | surrounding whitespace is accepted |
| `"+1:"` | a leading `+` is accepted |
| `"1_0:"` | underscores between digits — `int` 10 |
| `"99999999999999999999999999:"` | a bigint, past any `len`; saturates |

The underscore row is the one that fails **open**: rejecting `1_0` returns the input unchanged, rendering every element where Django renders none. Reading Django's source also corrected the issue's own quote of the `except` clause — it catches `KeyError` too.

## Do they share a root cause?

**No.** #2325 is expression resolution in `renderer.rs`; #2326 is index math in `filters.rs`. The hypothesis that both were "the argument/iterable not resolved through the full expression path" was checked and is false — `slice`'s argument was always resolved correctly, it was the *algorithm* that was wrong.

## Sink enumeration

Grepped for the sink (`get_value` / `get_value_safe` / `context.get`), not for expected callers. Every renderer site that resolves a user-written tag operand:

| site | before | now |
|---|---|---|
| `{% for %}` iterable | `context.resolve()` | `get_value` ✅ |
| `{% with %}` assignment | `context.get()` + raw-text fallback | `get_value` ✅ |
| `{% include … with %}` | `context.get()` + raw-text fallback | `get_value` ✅ |
| `{% if %}` truthiness | `context.get()` only, else `false` | `get_value` ✅ |
| `{% firstof %}` / `{% cycle %}` / `{% widthratio %}` | already `get_value_safe` | unchanged |

`TestEveryFilteredOperandSiteIsAccountedFor` pins the set mechanically — a fifth tag that grows its own bare lookup fails it, as does re-introducing either `Value::String(expression)` fallback.

Adjacent divergences the sweep found in **different mechanisms**, deliberately not fixed (#1079) and each pinned so the exclusion cannot become a blind spot: `{% regroup %}` with a filtered source (a Python-side assign tag with its own `RESOLVE_ARG_POSITIONS` channel — its module docstring already documents the limitation), `{% for k, v in d.items %}` (a callable mid-path, not a filter), and sequence equality in `{% if p == q %}`.

## Measurements

Everything below was run, not estimated.

| | before | after |
|---|---|---|
| randomised slice sweep vs live Django (1,494 cells) | **364 disagree** | **0** |
| #2325 repro table (15 cells) | **12 diverge** | **0** |
| tag-operand sweep, every registry filter × 8 inputs, gated on the `{{ }}` cell already agreeing | — | **0 unexplained** |

The gated sweep is the load-bearing one: without the gate it re-reports every pre-existing per-filter divergence (`add` on a list, `dictsort` on a string) as though this PR caused it. Residue is two **mechanical** predicates, not a name list — Django raised where djust renders the `{% empty %}` block, or djust's output *is* Django's escaped once more (the signature of the discarded safe flag, and a proof of direction). `test_the_residue_is_only_the_two_documented_shapes` asserts both actually occur, so neither classifier is a dead exemption.

### Two-build differential

Rebuilt `origin/main` baseline vs this branch, **66,520 cells**:

```
newly AGREEING: 21918
no longer agreeing: 445
  coincidental (the filter itself diverges on both builds): 445
  REGRESSIONS : 0

flagged as INTRODUCED   : 48
  fragment inside fully-escaped text (not live): 48
  LIVE (an unescaped tag opener)              : 0
```

**0 regressions, 0 introduced live-payload leaks.**

The tool **could not see #2325 at all** — its corpus was `{{ p|… }}` chains only and it built no tag cell, the same corpus-gap shape that once let it report clean over a live XSS (#2281). It now carries a `{% for %}` / `{% with %}` / `{% if %}` axis (tag ids carry a third `\t` field so `{{ }}` ids stay byte-identical and older baselines remain comparable).

Widening it exposed two ways its report misled, both now classified mechanically and both proven on this run:

1. A tag cell whose `{{ }}` twin diverges on **both** builds agreed on the baseline **by coincidence** — the operand bug rendered nothing and Django rendered nothing for its own reason. 445 of 445. Reporting these as regressions would have taught the next reader to ignore the number.
2. `live()` substring-matches, so `onerror=` also matches inside `&lt;img src=x onerror=alert(1)&gt;`. 48 of 48 were that. Both halves print in full; only the genuinely-live half gates.

Four subprocess tests pin both classifiers, including a gate-off sibling proving which half gated — the classifiers are load-bearing, not decorative (#1859).

The two structural couplings named in the brief were respected: no safety-set name was added, and the corpus was **extended** rather than the couplings weakened. `SLICE_SPECS` grew from 15 to 35 specs, and `test_the_negative_index_gap_is_list_shaped_too` — which named itself as the thing to delete when #2326 was fixed — is **inverted** rather than deleted, so "extend SLICE_SPECS" cannot be quietly skipped.

### Gate-off

**18 mutations, 18 caught, 0 survivors, 0 invalid.** The harness asserts the mutation text was found and found *exactly once*, asserts the source changed, **rebuilds between iterations**, counts pytest `N error` separately from `N failed` and refuses to report a number when the module did not import, distinguishes cargo's `error: test failed` from a compile failure, restores in a `finally`, and verifies the restore is byte-identical.

It earned its keep twice:

- The exactly-once assert **rejected a bad mutation** (`let t = s.trim();` matched two functions).
- One mutation **survived** the first run: `get_value_safe`'s `resolve()` fallback. A survivor is a question, not a pass — the answer was **genuinely uncovered**, not an equivalent mutation. Gating it off makes a bare `{% for x in u.orders %}` over a raw-Python sidecar render `''`, i.e. the #806 regression this PR had to avoid. `TestOperandSitesKeepTheGetattrWalk` (7 cells through `RustLiveView.set_raw_py_values`, plus a premise test proving the sidecar is attached) closes it; the mutation now fails 8 tests.

## Verification

- `pytest tests/ python/tests/ python/djust/tests/ -n auto` — **13,766 passed**, 0 failed (post-rebase).
- `make test-rust` — all crates green; 6 new Rust unit tests for `slice_positions` / `parse_slice_arg` / `python_int` / `apply_slice`.
- `ruff check` + `ruff format` — clean.
- Rebased onto `origin/main` (0 behind) before opening, so CI builds a real merge commit.

Closes #2325
Closes #2326


johnrtipton:
Follow-ups filed for the three adjacent divergences the sink sweep found in **other mechanisms** (CLAUDE.md #1079), each already pinned in `TestAdjacentDivergencesNotFixedHere` so the exclusion cannot become a blind spot:

- **#2333** — `{% regroup %}` ignores a filter on its source. A Python-side assign tag with its own `RESOLVE_ARG_POSITIONS` operand channel, so this PR's four renderer sites do not reach it. Its own module docstring already documented the limitation.
- **#2334** — `{% for %}` over a dict renders nothing. Two things: `.items`/`.keys`/`.values` are methods a `Value::Object` walk cannot resolve, and `Node::For` has no `Value::Object` arm. The same "match Python's iteration protocol" shape as this PR's string normalisation, one variant over.
- **#2335** — `{% if list_a == list_b %}` is always false; `values_equal` has no sequence arm. Reached through the `==` arm, which this PR does not touch.

The pins now name the issue to close when each goes away.

johnrtipton:
## Retrospective — PR #2332 (pipeline-drain, v1.1.1-7)

**Task**: #2325 (a filter on a `{% for %}` iterable renders nothing) and #2326 (`slice` clamps negatives, ignores `:step`).

### Quality: 5 — the highest-impact fix of the drain

### What went well

- **The sink grep found 4× the reported scope, and two sites were worse than the symptom.** `{% for %}` was only the reported instance; `{% with %}`, `{% include … with %}` and `{% if %}`-truthiness each open-coded the same bare lookup. The two `with` variants fell back to `Value::String(expression)` — **echoing the template's own source onto the page**: `{% with q=p|upper %}` rendered the literal text `p|upper`, and an unresolvable name rendered as its own name. And `{% if p|length %}` was **always false** — silent control-flow inversion, where a user's branch simply never runs. Verified independently on `main` before merge and again after.
- **My hypothesis was tested and rejected.** I suggested #2325 and #2326 might share a root cause. They don't — expression resolution in `renderer.rs` versus index math in `filters.rs` — and saying so plainly is worth more than a tidy narrative. They *compose*, which is why one PR was still right.
- **Django's source corrected the issue twice.** `slice_filter` catches `KeyError` too, and a one-part spec is `slice(stop)` — so `"2"` means `[:2]`, not `[2:]`, and djust had been returning the exact complement. Neither was in the issue.
- **A surviving gate-off mutation was treated as a question and the answer was real.** Removing `get_value_safe`'s `resolve()` fallback silently regressed #806 (`{% for x in u.orders %}` over a raw-Python sidecar → `''`). Genuinely uncovered, not equivalent — now pinned by `TestOperandSitesKeepTheGetattrWalk`.

### What didn't

- **The differential could construct no tag cell at all**, so this entire surface — four resolution sites, every tag operand — was invisible to the tool that has found nearly every bug in this drain. That is the same shape as #2296's blind spot, one axis over. Now swept: 66,520 cells, 21,918 newly agreeing, **0 regressions, 0 introduced leaks**.
- **Widening it exposed two ways its own report misleads**, both now classified mechanically rather than left to a reader's judgement: 445/445 "regressions" were baseline *coincidences* (bug rendered nothing; Django rendered nothing for its own unrelated reason), and 48/48 "leaks" were `onerror=` matched inside already-escaped `&lt;img …&gt;`. Calling either of those real would teach the next reader to ignore the number.
- **A wrong count shipped in the PR body** (34 vs 35 specs), caught by re-deriving it mechanically. Third time in this drain a prose count was wrong; the durable answer remains "any count in prose must be computed."
- **A `git checkout --` wiped uncommitted renderer edits mid-task.** Recovered by reapplying all seven behaviours via one script and verifying each is present — but it is the same class as the two git-hygiene slips I made earlier today, and the lesson is identical: commit before any operation that touches the working tree.

### Verified

13,766 pytest passed across all roots on the pushed HEAD; `make test-rust` clean; `cargo clippy --workspace --all-targets -D warnings` clean; 16/16 CI green including `test-summary`; rebased to 0-behind before opening. Gate-off 18/18 caught, 0 survivors, 0 invalid. Slice sweep 364 → 0 disagreements; #2325 repro table 12 → 0.

### Follow-ups filed (#1079)

#2333 (`{% regroup %}` filtered source), **#2334 (`{% for %}` over a dict renders nothing — flagged as likely a bigger real-world surface than #2325 itself)**, #2335 (`{% if list_a == list_b %}` always false). Each pinned in-test so the exclusion cannot go blind.

Issue #2340, opened 2026-08-28
`d.items` resolves to a list, not to a dict view: the repr differs and it is subscriptable where Python raises


johnrtipton:
The deliberate modelling residue of #2334 (PR TBD), filed per CLAUDE.md #1079 so the exemption is tracked rather than carried silently.

## What was decided

#2334 made `d.items` / `d.keys` / `d.values` resolve. They resolve to a **`Value::List`**, not to a model of Python's live `dict_items` / `dict_keys` / `dict_values` view.

Everything a template actually does with them is therefore exact — iteration, unpacking, `|length`, `|join`, truthiness, `{% with %}`, nesting. Two observable properties of a real view are not:

```
{{ p.items }}                         p = {"a": 1}
  django  'dict_items([(&#x27;a&#x27;, 1)])'
  djust   '[(&#x27;a&#x27;, 1)]'

{{ p.items|first }}                   p = {"a": 1}
  django  <<EXC TypeError>>           (a view is not subscriptable)
  djust   "('a', 1)"

{{ p.keys|slice:':2' }}               p = {"a":1,"b":2,"c":3}
  django  every key                   (`slice` catches the TypeError and returns the value)
  djust   the first two
```

So: the container's `str()` differs, and the view is subscriptable / JSON-serializable where Python's is not.

## Why it was not modelled faithfully

A faithful model needs a new `Value` variant (`DictView`), threaded through every `Value::List | Value::Tuple` or-pattern in the workspace — 19 `Value::List(..) | Value::Tuple(` or-patterns and 59 `Value::Tuple` sites across `djust_templates`, `djust_core`, `djust_live` and `djust_vdom` at the time of writing — plus a decision at each one about whether the variant behaves like a sequence there. That is a wide edit inside escaping machinery, four live XSSes into the same area in one week, bought for a repr almost no template renders on purpose.

Neither residue is more permissive than Django: the escaping is identical, and "djust renders where Django raises" is the shape #2325's differential already classifies and accepts.

## Direction, if it is taken

Add `Value::DictView { kind, items }` and decide, per call site, whether it is a sequence:

* **Sequence-like** (matches Python): `{% for %}`, `|length`, `|join`, `|unordered_list`, `in`, truthiness.
* **NOT sequence-like** (Python raises, and djust should render nothing rather than raise): `|first`, `|last`, `|slice`, `|json_script`, `|random`, `|dictsort`.
[...]

Issue #2339, opened 2026-08-28
Dict keys are always strings: `{% if 0 in d %}` matches a "0" key, and a non-string-keyed dict is not a mapping at all


johnrtipton:
Found while fixing #2334 / #2335 (PR TBD) and filed separately per CLAUDE.md #1079 — one root cause, two visible symptoms, and a design decision neither of those issues was about.

## The root cause

`Value::Object` is an `IndexMap<String, Value>`: its keys are always Rust `String`s. Two consequences follow, and they pull in opposite directions, which is why this needs a decision rather than a patch.

## Symptom 1 — `in` over a dict stringifies its needle

```
{% if p in q %}Y{% else %}N{% endif %}   p = 0,    q = {"0": 1}
  django  'N'          (Python: 0 == "0" is False)
  djust   'Y'

                                          p = 1.0,  q = {"1.0": 1}
  django  'N'   djust 'Y'
                                          p = None, q = {"None": 1}
  django  'N'   djust 'Y'
```

`evaluate_condition`'s `Value::Object` arm does `map.contains_key(&needle.to_string())`. A gate opens on a coincidence of `Display` formatting.

Found by #2335's randomised sweep in the shape `0 in {"0": {}}`; no curated case had it.

## Symptom 2 — a non-string-keyed dict is not a mapping at all

PyO3's `Value` extraction requires string keys, so a Python dict with any other key type falls through to its `repr`:

```
{{ q }}                                            q = {0: 1}
  django  '{0: 1}'     djust '{0: 1}'      agree, by accident

{% for k in q %}[{{ k }}]{% endfor %}              q = {0: 1}
  django  '[0]'
  djust   '[{][0][:][ ][1][}]'      the repr, iterated by CHARACTER
```

The second row is not new — `{% for %}` has iterated a string by character since #2325, and before that it rendered `{% empty %}`. It is visible now only because #2334 made the dict case work for string-keyed dicts, which makes the int-keyed one look like an inconsistency rather than a general absence.

## Why the obvious fix for symptom 1 is wrong on its own

[...]

Issue #2338, opened 2026-08-28
{% if a >= b %} is True for every pair Python cannot order — compare_values collapses "incomparable" into "equal"


johnrtipton:
Found by the randomised comparison sweep added in #2335 (PR TBD) and filed separately per CLAUDE.md #1079 — a different arm from the one that PR fixed.

## The divergence

`{% if a >= b %}` and `{% if a <= b %}` answer **True** for every pair Python cannot order, where Django answers **False**.

Measured against Django 5.2.16 via `_rust.render_template`:

```
{% if p >= q %}Y{% else %}N{% endif %}   p = "a",  q = 1
  django  'N'
  djust   'Y'

{% if p <= q %}Y{% else %}N{% endif %}   p = [1],  q = (1,)
  django  'N'
  djust   'Y'
```

`>` and `<` are correct for the same pairs — both answer `N`.

## Why

`compare_values` returns `-1` / `0` / `1` and collapses "incomparable" into `0`. For `>` and `<` that reproduces Django exactly: Python raises `TypeError`, Django's `{% if %}` catches it and answers False, and `0` makes both false. For `>=` and `<=` the same `0` reads as "equal" and answers True.

This is **pre-existing and per-pair, not per-type**: it holds for a string against an int, a list against a tuple, a dict against anything, and every other pair with no ordering arm. It long predates #2335 — that PR's two-build differential measured it identically on both builds (842 of 4,032 comparison cells, unchanged), which is why it stayed out of scope.

## Severity

Silent, and it fails in the permissive direction: a `{% if x >= threshold %}` gate opens when the two operands are not comparable at all. Lower impact than #2335's (the operands have to be genuinely mismatched, which usually means the template is already wrong), but the branch taken is the *unsafe* one.

## Direction

`compare_values` needs to distinguish "equal" from "cannot be ordered". PR #2335 briefly carried a `try_compare(a, b) -> Option<i32>` for exactly this and **removed it before merge**, because with the operators reading only the `i32` the `Option` was observationally equivalent to `0` — a second mechanism shadowing the first, which the gate-off harness surfaced (v1.1.1-2 retro). Re-introducing it here is the right shape, but it must come with the operator arms actually consuming it:

```rust
">=" => try_compare(&left, &right).is_some_and(|c| c >= 0)
"<=" => try_compare(&left, &right).is_some_and(|c| c <= 0)
```

[...]

PR #2341, opened 2026-08-28
fix(templates): iterate a dict, compare two sequences, resolve a filtered regroup source (#2334, #2335, #2333)

### LINKED ISSUE #2333: {% regroup %} ignores a filter on its source expression, and the groups render empty
Found by the sink sweep in #2325 (PR #2332) and filed separately per CLAUDE.md #1079 — a **different mechanism** from the four renderer operand sites that PR fixed, so it stayed out of scope.

## The divergence

`{% regroup %}` ignores a filter on its source expression.

```
{% regroup p|dictsort:"k" by k as g %}{{ g|length }}     p = [{"k": 2}, {"k": 1}]
  django  '2'
  djust   '0'
```

An unfiltered source is fine:

```
{% regroup p by k as g %}{{ g|length }}   -> '2' / '2'   agree
```

## Why it is not the #2325 fix

The four sites #2325 fixed (`{% for %}`, `{% with %}`, `{% include … with %}`, `{% if %}` truthiness) all resolve their operand inside `crates/djust_templates/src/renderer.rs`, and now all route through `get_value`.

`{% regroup %}` does not. It is a Python-side **assign tag** (`python/djust/template_tags/regroup.py`, `RegroupTagHandler`) whose operand arrives through a different channel entirely: `RESOLVE_ARG_POSITIONS = {0}` tells the Rust engine to resolve `args[0]` against the context and JSON-encode it, and the handler decodes it back. That resolution is not `get_value`, so pointing the renderer's four sites at the shared resolver does not reach it.

The handler's own module docstring already documents this as a known limitation:

> * The `<expr>` source must resolve to a JSON-encodable sequence
>   (django-normalised context values always are). Filter expressions on
>   the source (`cities|dictsort:"country"`) are not supported.

So this is a documented gap being promoted to a tracked issue, now that its three sibling gaps are closed and it is the last operand channel that drops filters.

## Severity

Same silent shape as #2325 — `{{ g|length }}` renders `0` and every `{% for %}` over the groups renders nothing, with no exception and no warning. `{% regroup cities|dictsort:"country" by country as by_country %}` is close to the canonical Django idiom (Django's own `regroup` docs open by noting the input usually needs sorting first), so this is likely to hit real templates.

## Direction

Resolve `args[0]` through the same filter-aware path the renderer's operands now use, at whichever site honours `RESOLVE_ARG_POSITIONS`. Worth checking whether any other `AssignTagHandler` / `CustomTag` declares resolved arg positions — if the resolution is shared, this is one fix for the whole custom-tag operand channel rather than one for `regroup` (grep for the sink, v1.1.1-2 retro).

## Pinned

`python/tests/test_filtered_operands_and_slice_2325_2326.py::TestAdjacentDivergencesNotFixedHere::test_regroup_does_not_resolve_a_filtered_source` asserts the divergence still exists, and names itself as the thing to delete when it does not — so the exclusion cannot quietly become a blind spot.

### LINKED ISSUE #2334: {% for %} over a dict renders nothing — .items/.keys/.values unresolvable, and no Object arm
Found by the sink sweep in #2325 (PR #2332) and filed separately per CLAUDE.md #1079 — not a filter bug, so it stayed out of that PR's scope.

## The divergence

`{% for %}` over a dict, and over a dict's `.items` / `.keys` / `.values`, renders nothing.

Measured against Django 5.2.16 via `_rust.render_template`:

```
{% for k, v in p.items %}{{ k }}{{ v }}{% endfor %}   p = {"a": 1}
  django  'a1'
  djust   ''

{% for x in p %}{{ x }}{% endfor %}                   p = {"a": 1, "b": 2}
  django  'ab'          (Python iterates a dict's KEYS)
  djust   ''
```

## Why it is not #2325

#2325 was about a **filter** on the operand being dropped. This operand carries no filter — it is a plain variable path — and it now resolves correctly through `get_value` after PR #2332. Two separate things are still missing after it:

1. **`.items` / `.keys` / `.values` are Python methods, not attributes.** Resolving them needs the callable to be *called* mid-path. `Context::resolve` already does this for raw Python objects attached via the sidecar (`maybe_call`), but a plain dict in the JSON state is a `Value::Object`, and `Context::get`'s nested walk does `obj.get(part)` only — there is no `items` key, so the lookup misses.

2. **`Node::For`'s match has no `Value::Object` arm.** PR #2332 added a `Value::String` normalisation (Python iterates a string by character), and the same argument applies to a dict: Python iterates its **keys**. Everything that is not a list/tuple/string still falls to `_ =>`, which renders the `{% empty %}` block.

So this is two small changes in the same arm, both of the "match Python's iteration protocol" shape that #2332's string normalisation already established.

## Severity

Silent — an empty region where the page had a table, no exception and no warning. `{% for key, value in mydict.items %}` is one of the most common Django loop idioms there is, which makes this a bigger practical surface than the filter case #2325 covered.

## Direction

In `crates/djust_templates/src/renderer.rs`, extend the normalisation `{% for %}` applies to its resolved operand — it currently turns a `Value::String` into a sequence of characters — so a `Value::Object` becomes its sequence of keys, matching Python. Separately, make `.items` / `.keys` / `.values` resolvable on a `Value::Object` in the context walk; check whether `{{ }}` has the same gap (`{{ d.items }}`) so the two paths are decided together rather than one at a time (#1646).

Two notes for whoever picks this up:

- Iteration ORDER must be Python's insertion order, not a hash order, or `{% for k in d %}` will be nondeterministic across renders and will thrash the VDOM.
- `{% for k, v in d.items %}` needs each item to unpack as a 2-sequence; the existing tuple-unpacking branch handles that once the value arrives as a list of pairs.

## Pinned

`python/tests/test_filtered_operands_and_slice_2325_2326.py::TestAdjacentDivergencesNotFixedHere::test_for_over_a_dict_items_call_is_a_separate_resolution_gap` asserts the divergence still exists and names itself as the thing to delete when it does not.

### LINKED ISSUE #2335: {% if list_a == list_b %} is always False — values_equal has no sequence arm
Found by the sink sweep in #2325 (PR #2332) and filed separately per CLAUDE.md #1079 — a different arm of `evaluate_condition` from the one that PR touched.

## The divergence

`{% if %}` comparing two sequences answers `False` where Python and Django answer `True`.

Measured against Django 5.2.16 via `_rust.render_template`:

```
{% if p == q %}Y{% else %}N{% endif %}   p = ["a"], q = ["a"]
  django  'Y'
  djust   'N'

{% if p == p %}Y{% else %}N{% endif %}   p = ["a", "b"]
  django  'Y'
  djust   'N'
```

Scalars are fine, and so is a string:

```
{% if p == p %}   p = "ab"   -> 'Y' / 'Y'   agree
{% if p == p %}   p = 3      -> 'Y' / 'Y'   agree
```

So the gap is specifically that `values_equal` has no structural arm for `Value::List` / `Value::Tuple` (and, presumably, `Value::Object`) — two sequences are never equal, not even to themselves.

## Why it is not #2325

#2325 fixed the `{% if %}` **truthiness** arm — a bare filter expression such as `{% if p|slice:":1" %}`, which reached the end of `evaluate_condition` and returned a hardcoded `false`. This is the `==` arm, which is reached first and resolves both operands correctly through `get_value`; the operands are right and the *comparison* is wrong. PR #2332 does not touch `values_equal`.

## Severity

Lower than #2325's — the output is a wrong branch rather than a silently-empty region, and `{% if list_a == list_b %}` is a less common idiom than `{% for x in items|slice %}`. But it is still silent, and it fails in the direction that hides content: the `{% else %}` branch runs when the two really are equal.

## Direction

Give `values_equal` a structural arm: two sequences are equal when they have the same length and are pairwise equal, recursing through the same function (so nesting, and the existing numeric-widening rules for elements, come along). Decide `Value::Object` at the same time rather than one variant later (#1646) — Python compares dicts by key/value pairs.

Worth checking `compare_values` in the same pass: Python orders lists lexicographically, and `{% if list_a > list_b %}` presumably has the mirror gap. Also confirm whether the tuple/list cross-type case should compare equal — in Python `[1] == (1,)` is **False**, so a naive "both are sequences" arm would be wrong in a way a curated test might not catch. A randomised differential against live Django over small sequences of mixed element types is the cheap way to settle all of this at once, and would be more reliable than reasoning about it (v1.1.1-2 retro).

## Pinned

`python/tests/test_filtered_operands_and_slice_2325_2326.py::TestAdjacentDivergencesNotFixedHere::test_sequence_equality_in_an_if_condition_is_a_separate_gap` asserts the divergence still exists and names itself as the thing to delete when it does not.

johnrtipton:
Three silent divergences the #2325 sink sweep found in mechanisms PR #2332 did not touch. Two of them fail in the direction that **hides content** — an empty region, or the `{% else %}` branch — with no exception, no warning and nothing in the console. That is the worst failure shape a template engine has, which is why they were treated as the priority rather than as parity nits.

|---|---|---|
| `{% for k, v in d.items %}{{ k }}={{ v }} {% endfor %}` | `a=1 b=2 ` | **`''`** |
| `{% for k in d %}{{ k }} {% endfor %}` | `a b ` | **`''`** |
| `{% if list_a == list_b %}` on two equal lists | `same` | **`diff`** |
| `{% regroup p\|dictsort:"k" by k as g %}{{ g\|length }}` | `2` | **`0`** |

## #2334 — `{% for %}` over a dict rendered nothing

Two independent gaps, both of them "match Python's iteration protocol", and the issue was right that fixing one alone leaves the region empty.

1. **`.items` / `.keys` / `.values` are METHODS, not keys.** `Context::get`'s nested walk only ever does `obj.get(part)`, so the lookup missed; Django reaches them through `Variable._resolve_lookup`'s attribute step plus its auto-call.
2. **`Node::For` had no `Value::Object` arm**, so a bare `{% for k in d %}` fell to the `_ =>` wildcard and rendered `{% empty %}`.

The view resolution is **one chokepoint** — `Context::resolve`, which every operand site reaches (`{{ }}` calls it directly, and `{% for %}` / `{% if %}` / `{% with %}` / `{% include … with %}` reach it as `get_value_safe`'s last arm, so #2325's four sites come along for free). Placed **after** the `get`, because Django tries mapping-item access first and attribute access second: a dict with its own key named `items` still resolves to that key's value.

The `{% for %}` arm reuses #2325's normalisation rather than growing a fourth match arm, so the dict shares the whole loop body — `{% empty %}`, `reversed`, `{% cycle %}`, nesting — instead of a parallel copy (#1646). Order is the `IndexMap`'s insertion order, i.e. Python's; a hash order would make the loop nondeterministic across renders and thrash the VDOM.

## #2335 — two sequences were never equal, not even to themselves

`values_equal` had no structural arm, and `compare_values` had the same hole from the ordering side. Both now **recurse through the same function**, which is what carries the numeric widening down — `[1] == [1.0]` and `[True] == [1]` are both true, through the existing #2243 / #2244 arms rather than a second copy of them — and it fixes `{% if x in list_of_lists %}` at the same time, since `in` is the third caller of `values_equal`.

Two things a curated table would plausibly get wrong, both settled by running Django:

* **`[1] == (1,)` is False in Python.** A "both are sequences" arm would be wrong in exactly the direction a hand-written table is least likely to probe.
* **Ordering walks with `==` and only an EQUAL pair continues.** `[{}, 1] < [{}, 2]` is True even though two dicts cannot be ordered — they never need to be. The first draft asked "is this pair ordered?" first and continued on a `0`, which conflates "equal" with "incomparable" and falls through to the length tie-break: `[[], 'a', ('b',)] > [1]` answered True because three elements beat one. The randomised sweep found it in **27 of 28,500 cells** (measured by re-running the sweep against the first-draft build, not read off a printout); no curated case had the shape.

That draft also carried a `try_compare(…) -> Option<i32>` to make incomparability explicit. **It was removed before merge**: with every operator reading only the `i32`, `None` and `Some(0)` are observationally identical, so it was a second mechanism shadowing the first — the gate-off harness is what surfaced it (v1.1.1-2 retro: *delete the redundant one rather than testing around it*). The whole fix is now the one rule.

## #2333 — `{% regroup %}` dropped a filter on its source

The fourth and last operand channel, and the one #2332 could not reach: `{% regroup %}` is a Python-side assign tag whose source arrives through `RESOLVE_ARG_POSITIONS` plus a JSON hop. It asked for a variable literally *named* `cities|dictsort:"country"`, missed, and handed the handler the template's own source text.

One `resolve_tag_operand` now resolves a pipe-bearing operand through `get_value`. The pipe is the guard, and it is load-bearing rather than a shortcut: unlike the renderer's four sites, this channel's contract is *"unresolved ⇒ pass the raw token"* — which is what lets regroup's `by` / `<attr>` / `as` / `<var>` through — and `get_value`'s literal arms have no way to say "unresolved". They would answer `Bool(true)` for a keyword operand spelled `True`. The module docstring's "filter expressions on the source are not supported" limitation is deleted.

## Sink enumeration

Grepped for the sink, not for the callers.

**Iteration** — every site that turns a value into a sequence:

| site | before | now |
|---|---|---|
| `Node::For` (`renderer.rs`) | `List`/`Tuple`/`String` | + `Object` → keys ✅ |
| `{% regroup %}` source (Python) | JSON list | unchanged (its own sequence check) |
| `length` / `join` / `unordered_list` / `in` / truthiness on a dict | already agreed | unchanged |
| `{% cycle %}` / `{% firstof %}` | take N args, never iterate a value | n/a |

`Node::For` appears twice in `renderer.rs` and only one is a renderer (the other is `nodes_contain_elements`); `parser.rs` / `inheritance.rs` construct or rewrite the node without iterating. The `{{ d|length }}` / `{{ d|join }}` / `{% if k in d %}` cells were **measured** to already agree — `{% for %}` was the one iteration sink that did not.

**Equality** — every consumer of a comparison:

| site | function | now |
|---|---|---|
| `==` / `!=` | `values_equal` | + List/Tuple + Object ✅ |
| `in` over a sequence | `values_equal` | fixed by the same arm ✅ |
| `in` over a **dict** | `contains_key(needle.to_string())` | **not fixed** — see below |
| `<` / `>` / `<=` / `>=` | `compare_values` | + List/Tuple lexicographic ✅ |
| `is` / `is not` | `values_identity` | unchanged (Python `is` on a list is identity, i.e. false) |
| `dictsort` / `regroup` grouping keys | Python-side | out of channel |

## Not more permissive than Django

The `{% for %}` safe-key mapping asserts that the loop variable **is** `<iterable>.<index>`. A **normalised** sequence falsifies that — the loop iterates something built from the resolved value, not its own indexable elements — and for a dict it is a live XSS rather than a theoretical one:

`_collect_safe_keys` writes a dict's paths **by key name** (`d.1`), while the mapping is **by position**. So a dict `{"1": mark_safe("<b>ok</b>"), "": "v"}` puts `d.1` in `safe_keys`, and the loop's *second* key — an entirely different, attacker-controlled string — would resolve that mark and be emitted raw. The mapping is now gated on `!normalised`, which covers the string case by the same argument, and the path is exercised end to end through the production `_collect_safe_keys` collector (`test_the_loop_safe_key_mapping_is_not_registered_for_a_dict`, with a premise assertion so it cannot pass for the wrong reason).

`test_a_genuine_list_keeps_its_item_safety` is the other half: the gate is on normalisation, not on loops generally, so a real list operand still resolves its per-item marks (#2287).

## Measurements

Everything below was run.

### Two-build differential

Rebuilt `origin/main` baseline vs this branch, **95,275 cells**:

```
newly AGREEING: 178
no longer agreeing: 21
  coincidental (the filter itself diverges on both builds): 21
  REGRESSIONS : 0

flagged as INTRODUCED   : 3
  fragment inside fully-escaped text (not live): 3
  LIVE (an unescaped tag opener)              : 0
```

**0 regressions, 0 introduced live-payload leaks.** The 21 are `{% for x in d|add:"1" %}`-shaped: Django's `add` on a dict fails and returns `''`, djust's returns the dict, and the new `Value::Object` arm now iterates it — the *filter's* pre-existing divergence, surfaced. The tool's own #2332 classifier says so mechanically (21 of 21). The 3 flagged leaks are the #2332 escaped-text classifier: djust's output is `[1][&lt;img src=x onerror=alert(1)&gt;]` — the key is escaped, which is the safe-key gate working.

Both #2332 classifiers still fire; neither was weakened.

### The corpus could reach neither shape

The differential reported clean over the whole of both bugs, **for the third time in the same shape** (after #2281 and #2325). Its tag axis writes `p|<filter>` as every operand, so the corpus contained no dotted path and nothing that iterated a dict without a filter in the way; and its `{% if %}` cells bind only `p`, so it could not construct a comparison at all. Closed here:

* a **dict-view path** axis — 12 shapes × every input (324 cells), including `{{ p.items }}`, `{% for k, v in p.items %}`, `{% for x in p %}`, and the path composed with a filter;
* a **sequence-comparison** axis — 7 operators × every input × every input (5,103 cells), with the right-hand operand a **deep copy** so nothing can pass by identity;
* a **`d-hostile-key`** input — every existing dict input has tame keys (`k`, `j`), so no cell could show a dict-KEY escaping defect, and the key is what `{% for k in d %}` emits. Ordered so the payload sits at index 1, which is the collision shape above.

`TestTheCorpusGapsThatHidTheseFromTheDifferential` pins all three. The corpus was **extended**, never weakened: no safety-set name, branched-on whitespace character or hot-set filter was touched.

### Randomised, against live Django

| sweep | cells | result |
|---|---|---|
| dict iteration (14 shapes × random dicts) | 5,600 | 0 unexplained |
| sequence comparison (5 operators × nested mixed values) | 7,500 | 0 unexplained, 1 documented exemption |
| equal-but-distinct deep copies | 600 | 0 disagree |
| hostile-dict permissiveness probe | 2,100 | 0 more permissive |

The dict-iteration residue is a **mechanical predicate**, not a name list: *djust's answer IS Django's answer for the same template over `list(view)`*. It states exactly one modelling choice and a second, unrelated defect cannot hide behind it. Both residue classifiers are asserted to actually fire, so neither is a dead exemption (#1859).

### Gate-off

**11 mutations, 11 caught, 0 survivors, 0 invalid.** The harness asserts the mutation text was found and found *exactly once*, asserts the source changed, **rebuilds between iterations**, counts pytest `N error` separately from `N failed` and refuses to report a number when the module did not import, distinguishes cargo's `error: test failed` from a compile error, restores in a `finally`, and verifies the restore is byte-identical.

| mutation | pytest | cargo |
|---|---|---|
| #2334 dict-view resolution (the `Context::resolve` arm) | 6 failed | 2 failed |
| #2334 dict-view ORDER (a real key must win over the method) | 1 failed | 1 failed |
| #2334 `.items` entries are 2-TUPLEs (what makes unpacking work) | 4 failed | 1 failed |
| #2334 `Node::For` `Value::Object` arm | 6 failed | 0 failed |
| #2334 safe-key mapping gate (`!normalised`) — the XSS half | 3 failed | 0 failed |
| #2335 `values_equal` List/Tuple arm | 7 failed | 2 failed |
| #2335 `values_equal` Object arm | 4 failed | 2 failed |
| #2335 `compare_values` List/Tuple arm (ordering) | 3 failed | 2 failed |
| #2335 ONLY AN EQUAL PAIR continues the walk (the first-draft shape) | 2 failed | 1 failed |
| #2333 filter-aware assign-tag operand | 4 failed | 0 failed |
| #2333 the raw-token fallback (regroup's `by` / `as` operands) | 0 failed | 1 failed |

It earned its keep three times:

1. The **exactly-once** assert rejected a bad mutation (`if expr.contains('|') {` matches two functions).
2. The first "incomparability propagates" mutation was a **semantic no-op** — the v1.1.1-2 failure mode where a valid mutation computes the identical answer. Replacing it with the real first-draft shape is what showed the `Option` was redundant, and the redundant mechanism was deleted rather than tested around.
3. Every mechanism has a mutation where **only it** is removed, so none can shadow another (#2129/#2135).

## Adjacent divergences found, measured, filed — not fixed (#1079)

Each is pinned by a self-naming test that tells whoever fixes it to delete the pin.

* **#2338** — `{% if a >= b %}` / `<=` answer True for every pair Python cannot order (`"a" >= 1` → djust `Y`, Django `N`). `compare_values` collapses "incomparable" into `0`, which is *right* for `>` and `<` and wrong for `>=` / `<=`. Pre-existing and per-pair, not per-type: **1,136 of 5,103** comparison cells (568 each), identical on both builds.
* **#2339** — `in` over a dict stringifies its needle, so `0 in {"0": 1}` is True where Python says False (1 differential cell, and the sweep's one exemption). Not fixed because djust's wire format coerces dict keys to strings, so the coercion is the only thing keeping `{% if pk in d %}` working against a view's own `{pk: …}` mapping — a behaviour #2221 pins deliberately. Removing it moves djust *towards* Django for one shape and *away* for the other. It was written, measured and **reverted** rather than shipped.
* **#2340** — `d.items` resolves to a list, not to a live `dict_items` view, so the container's `str()` reads `[…]` and it is subscriptable where Python's raises. A faithful model needs a new `Value` variant threaded through 19 `Value::List | Value::Tuple` or-patterns and 59 `Value::Tuple` sites — a wide edit inside escaping machinery, four live XSSes into the same area in one week, for a repr almost no template renders on purpose.

The three pins #2332 left behind for these issues are **deleted**, which is the contract they carried.

## Verification

- `pytest tests/ python/tests/ python/djust/tests/ -n auto` — **13,920 passed**, 0 failed (post-rebase).
- `make test-rust` — all crates green; 11 new Rust unit tests across `values_equal` / `compare_values` / `Context::dict_view`.
- `ruff check` + `ruff format` — clean.
- 44 tests in `python/tests/test_dict_iteration_and_sequence_equality_2334_2335.py`.
- Rebased onto `origin/main` (0 behind) before opening, so CI builds a real merge commit.

Closes #2334
Closes #2335
Closes #2333


johnrtipton:
## Retrospective — PR #2341 (pipeline-drain, v1.1.1-7)

**Task**: #2334 (`{% for %}` over a dict renders nothing), #2335 (sequence equality always False), #2333 (`{% regroup %}` ignores a filter on its source).

### Quality: 5

### What went well

- **A vulnerability was found and gated in the same PR that would have created it.** `_collect_safe_keys` writes a dict's safe paths **by key name** (`d.1`) while the loop mapping is **by position** — so `{"1": mark_safe(…), "": "v"}` would have emitted the second key unescaped once dict iteration started working. Not a shipped XSS: on `main` the whole construct renders nothing, so it was unreachable *because the feature was broken*. Verified both halves independently — hostile key inert on `main` (empty output), escaped on this PR (`1 &lt;img src=x onerror=alert(1)&gt;`) with the legitimately-marked sibling still live.
- **A surviving gate-off mutation was a semantic no-op, and chasing it deleted a redundant mechanism.** Replacing it with the real first-draft shape revealed `try_compare`'s `Option` was observationally identical between `None` and `Some(0)` — so it was **deleted** rather than tested around (#1859). That is the third redundant mechanism removed in this drain rather than papered over.
- **Both self-caught errors were caught by tooling, not by inspection.** The first ordering arm was a regression (reading an incomparable element as a tie fell through to the length tie-break, so `[[], 'a', ('b',)] > [1]` answered True) — found by the randomised sweep in 27 of 28,500 cells. And an `in`-over-a-dict fix was implemented, then **reverted**: requiring a string needle matches Python but breaks djust, whose wire format coerces dict keys to strings, and that coercion is what keeps `{% if pk in d %}` working. Net parity a wash; filed as a wire-format decision instead of forced through.

### What didn't

- **The differential could reach neither shape — third occurrence in the same form** (after #2281 and #2325). Its tag axis wrote `p|<filter>` as every operand: no dotted path, nothing iterating a dict unfiltered, and its `{% if %}` cells bound only `p`, so no comparison existed at all. Every dict in the corpus also had tame keys (`k`/`j`) — and the key is precisely what `{% for k in d %}` emits. Closed with a dict-view path axis (324 cells), a sequence-comparison axis (5,103 cells, right operand deep-copied so nothing passes by identity), and a `d-hostile-key` input, pinned by `TestTheCorpusGapsThatHidTheseFromTheDifferential`.
- **A prose count was wrong** — "12 cells" read off a truncated printout, actually 27, corrected by re-running the sweep. Fifth wrong prose count in this drain; the standing remedy applies.
- **The `Closes` list closed an issue this PR deliberately left open.** #2338 was filed as a follow-up (#1079) and was auto-closed as COMPLETED with the bug live on `main` — verified across three incomparable pairs. Reopened. **A closed issue with a live bug is worse than an open one**, because nothing looks at it again. Worth treating as a general hazard for multi-issue PRs that also file follow-ups.

### Verified independently before merge

Dict iteration, `.items` iteration and sequence equality all byte-match Django on a release build; hostile dict key escaped while its marked sibling stays live.

Author's gates: 13,920 pytest across all roots; `make test-rust` green with 11 new Rust tests; clippy `-D warnings` clean; 16/16 CI; rebased to 0-behind. Gate-off 11 mutations, 11 caught, 0 survivors, 0 invalid. Two-build differential over 95,275 cells — **178 newly agreeing, 0 regressions, 0 introduced live-payload leaks** (21 "no longer agreeing" all classified coincidental; 3 flagged leaks all escaped-text, 0 LIVE).

### Follow-ups filed (#1079)

#2338 (`>=`/`<=` True for every incomparable pair — **1,136 of 5,103 cells**), #2339 (`in` over a dict stringifies its needle — the revert above), #2340 (`d.items` is a list, not a live view; a faithful model needs a new `Value` variant through 19 or-patterns and 59 `Value::Tuple` sites).

Issue #2368, opened 2026-08-28
{% regroup d.values %} silently misses: the assign-tag operand channel resolves through get, not resolve


johnrtipton:
Found while fixing #2340 (PR TBD) and filed separately per CLAUDE.md #1079 — measured on two builds, pre-existing since #2334, and untouched there.

## The cell

```
{% regroup p.values by k as g %}{{ g|length }}     p = {"a": {"k": 2}, "b": {"k": 1}}
  django  '2'
  djust   '0'
```

Silent: no exception, no warning. The handler receives the template's own token
text and produces nothing, so the region is simply empty — the worst failure
shape a template engine has, and the one #2333 was about.

## Root cause: the operand channel's non-pipe branch uses `get`, not `resolve`

`resolve_tag_operand` (`crates/djust_templates/src/renderer.rs:555`) has two
branches:

```rust
if expr.contains('|') {
    return match get_value(expr, context) { … };   // #2333 routed this one
}
context.get(expr).map(value_to_arg_string)          // …and left this one
```

`d.items` / `.keys` / `.values` resolve in **`Context::resolve`**, not in
`Context::get` — `dict_view` is only reachable from `resolve`, which is where
#2334 put it. So the pipe-bearing branch sees a view and the bare dotted path
does not, and the tag falls back to its "unresolved ⇒ keep the raw token"
contract.

This is the same shape as #2333, one operand form over: that fix made the
channel filter-aware and this is the channel still not being
dict-view-aware.

## Measured on two builds

| build | `{% regroup p.values by k as g %}{{ g\|length }}` |
|---|---|
| #2339 (`d.values` is a `Value::List`) | `0` |
| #2340 (`d.values` is a `Value::DictView`) | `0` |

Unchanged by the `Value::DictView` variant, which is why it is filed rather
than folded in. What #2340 *did* fix on this channel is the other half:
`value_to_arg_string` collapsed a view to `to_string()`, so even when the
operand DID resolve — through the pipe branch — the handler got the text
`dict_items([…])` instead of the rows. That arm is in, and
`{% regroup p.values|slice:':2' by k as g %}` agrees today. Only the bare
dotted path is left.

Controls, all agreeing, which localise it precisely:

```
{% regroup p by k as g %}                        list operand          '2' / '2'
{% regroup p.rows by k as g %}                   dotted path to a LIST '2' / '2'
{% regroup p|dictsort:"k" by k as g %}           filtered source       '2' / '2'
[...]

Issue #2361, opened 2026-08-28
A mark_safe VALUE loses its mark when the loop reaches it through d.values / d.items


johnrtipton:
Found while fixing #2340 (PR TBD) and filed separately per CLAUDE.md #1079 — measured, pre-existing, and not touched there.

## The cell

```
{% for v in p.values %}[{{ v }}]{% endfor %}     p = {"a": mark_safe("<b>ok</b>")}
  django  '[<b>ok</b>]'
  djust   '[&lt;b&gt;ok&lt;/b&gt;]'

{% for k, v in p.items %}[{{ v }}]{% endfor %}   same
  django  '[<b>ok</b>]'
  djust   '[&lt;b&gt;ok&lt;/b&gt;]'
```

djust **over**-escapes: a value the view explicitly marked safe comes out escaped. That is the safe direction — it is a lost capability, never a leak — but it is a divergence, and it is silent: the page simply shows `&lt;b&gt;` where the author expected markup.

`{% for x in p %}` (the bare dict, iterating keys) AGREES, because keys are not the thing that carries a mark.

## Root cause: two spellings for the same path that never meet

This is the #2287 channel — per-ITEM safety arriving from the CONTEXT rather than from a filter — reached through a dict view instead of a list.

* `_collect_safe_keys` (`python/djust/mixins/rust_bridge.py`) walks the dict and writes one dotted path per `SafeString` it finds, **by key name**: `p.a`.
* The loop safe-key mapping in `Node::For` asserts the loop variable IS `<iterable>.<index>`, which for `{% for v in p.values %}` means looking up **`p.values.0`**.

Neither spelling is wrong on its own; they simply never produce the same string, so `Context::is_safe` misses and the value is escaped.

## Measured, on two builds

Identical before and after #2340, so the `Value::DictView` variant neither introduced nor changed it:

| build | `{% for v in p.values %}` on `{"a": mark_safe("<b>ok</b>")}` |
|---|---|
| #2339 (`d.values` is a `Value::List`) | `[&lt;b&gt;ok&lt;/b&gt;]` |
| #2340 (`d.values` is a `Value::DictView`) | `[&lt;b&gt;ok&lt;/b&gt;]` |

Worth stating explicitly, because #2340 makes the loop-mapping gate (`!normalised`) apply to a view: **that gate is not what causes this.** Flipping it changes no output, precisely because `p.values.<index>` is never in `safe_keys` either way. The gate is defence in depth, and this issue is upstream of it.

[...]

PR #2369, opened 2026-08-28
fix(templates): d.items is a dict VIEW, not a list (#2340)

### LINKED ISSUE #2340: `d.items` resolves to a list, not to a dict view: the repr differs and it is subscriptable where Python raises

johnrtipton:
**Stacked on #2353 (#2339)**, which this needs: a view's elements are the dict's keys, and those now carry their Python type. Target this at `main` after #2353 merges.

#2334 made `d.items` / `.keys` / `.values` resolve — to a `Value::List`. Everything a template usually does with one was exact; two observable properties of a real view were not.

|---|---|---|
| `{{ p.items }}` | `dict_items([(&#x27;a&#x27;, 1)])` | `[(&#x27;a&#x27;, 1)]` |
| `{{ p.items\|first }}` | `<<EXC TypeError>>` | `('a', 1)` |
| `{{ p.values\|dictsort:"k" }}` | the sorted rows | **`''`** |
| `{{ p.keys\|pluralize }}` on a 1-entry dict | `''` | **`s`** |

The last two are not from the issue. They were found here, and both are content-losing.

## The issue's list of what raises is wrong in two directions

The issue named `|first`, `|last`, `|slice`, `|json_script`, `|random` and `|dictsort` as the not-sequence-like set. That list is a hypothesis, so it was checked by running **every** Django built-in against **all three** kinds:

* **`|slice` does not raise.** Django's `slice` catches the `TypeError` and returns the value UNCHANGED, so `{{ d.keys|slice:':1' }}` renders the whole view repr and `{{ d.keys|slice:':1'|join:'' }}` is still **every** key. Modelling it as "returns nothing" would have shipped a new divergence. `apply_slice`'s existing `_ => Ok(value.clone())` already does the right thing, so it needed no edit.
* **`|dictsort` does not raise either** — it is `sorted(value, key=…)`, and `sorted()` takes any iterable. `{{ d.values|dictsort:"k" }}` is a real working idiom returning a LIST.
* **Five more raise that the issue did not name** — `divisibleby`, `get_digit`, `phone2numeric`, `timesince`, `timeuntil` — because they raise for any non-scalar, not because of the view.

And the framing of the repr as "a debug-only repr almost no template renders on purpose" understates it: **a third of the filter registry sees the view's `str()`.** `|truncatewords`, `|wordcount`, `|linebreaks`, `|stringformat`, `|striptags`, `|pprint`, `|escape`, `|safe`, `|yesno` and `|make_list` all operate on the text `dict_keys([…])`. The repr is their input.

## The model

`Value::DictView { kind, items }` — the kind decides only the container's name, which is why it is a field and not three variants. Django's behaviour splits **three** ways, not two:

| behaviour | filters / sinks |
|---|---|
| **sequence-like** | `{% for %}`, `in`, truthiness, `\|length`, `\|join`, `\|unordered_list`, `\|safeseq`, `\|escapeseq`, `\|dictsort` |
| **raises** | `\|first`, `\|last`, `\|random`, `\|json_script` |
| **its `str()`** | ~a third of the registry, above |

Where Python raises, djust renders **nothing** rather than raising — the shape #2325's differential already classifies and accepts, and never more permissive: no output is not a leak.

`iter_values` yields the items, so every iterating filter works from one sink. `|random` and `|json_script` guard at their **own** arm rather than being carved out of that sink, because a view genuinely IS a sequence — what it refuses is *subscripting*.

### Every exhaustive-match site was decided, not defaulted

| site | decision |
|---|---|
| `is_truthy` | empty view is falsy (`bool({}.keys())` is `False`) |
| `Display` | names its container |
| `legacy_display` | **`[List]` — unchanged**, see below |
| `Serialize` | its items; a view never crosses the wire, and `json.dumps` refuses one outright |
| `IntoPyObject` | a list; a real view cannot be rebuilt without its dict |
| `ObjectKey::from_value` | `None` — a view is unhashable in Python too |
| `value_to_json` | `null` |
| `pprint::flat_repr` | names its container (`pformat(d.keys())` measured) |
| `loop_cache::hash_value` | its own tag **and the kind in it** |

The loop-cache one is the sharpest: a view renders `dict_keys(['a'])` and a list of the same items renders `['a']`, so sharing a tag would let a cached fragment for one be served for the other — the same argument the existing `Missing`/`None` and `Tuple`/`List` splits make. Both collisions have an end-to-end test through the real cached render path, because that file's own docstring records that testing `hash_value` in isolation gate-offs to zero.

### rustc enumerates only the exhaustive matches — the other 19 were audited by hand

An or-pattern with a `_` fallback compiles fine and sends a view silently down the scalar path. All **19** `Value::List | Value::Tuple` sites were checked. Four needed a decision; **two were wrong**:

* **`pluralize`** reached the `_` arm, which returns the suffix unconditionally — right for a 2-entry view by luck, wrong for a 1-entry one.
* **`value_to_arg_string`** collapsed a view to `to_string()`, handing an assign-tag handler the text `dict_items([…])` instead of the rows. That is the #2042 `[List]`-collapse class with a different placeholder. The arm is live, not decoration: `{% regroup p.values|slice:':2' by k as g %}` agrees only with it.

The other 15 were decided and left alone: `first`/`last` (Django raises; `_ => Missing` is right), `apply_slice`, `context.rs`'s numeric-index walk (`{{ d.items.0 }}` is `''` in Django too), and the rest either already handled or operating on a view's ELEMENTS, which can never themselves be views.

### `legacy_display` does NOT name the container, and the first version got this wrong

The naming arm was originally written into **both** display paths, on a comment asserting *"the container spelling is Python's on BOTH display paths"*. That was a prose invariant nobody had run. The gate-off surfaced the legacy arm as a **surviving mutation**, and the test written to close that gap **failed on its first execution** with `dict_items([[List]])` — CLAUDE.md #1867, a canon claim of the form "X always holds" must be falsification-tested, not merely have its citation checked.

`legacy_display` is the pre-#2203 rendering where every container is a `[List]` placeholder, and before #2340 a view WAS a `Value::List`, so `[List]` is exactly what `{{ d.items }}` printed under the flag. Naming it there would make a legacy-rendering switch *less* legacy. The test now pins that the legacy path is **unchanged**, which is the real invariant.

## Verified site counts

The issue's estimate, checked mechanically rather than quoted:

```
Value::List(..) | Value::Tuple( or-patterns : 19
Value::Tuple sites                          : 59
Value::Object sites                         : 123
```

Both figures the issue gave are **exact**.

## Two structural pins were corrected, not worked around

Both fired on this change, and in both cases the pin — not the code — was the thing that needed fixing:

* `test_the_iteration_sink_has_exactly_the_callers_it_claims` used a **400-char proximity window**. The new `random` guard's comment pushed its `iter_values` call past it, so the pin reported that a filter had stopped routing through the sink when it had not. It now delimits each arm's own body, which is strictly **tighter** — it can never reach into a neighbouring arm either.
* `test_every_bare_list_site_is_one_of_the_documented_list_always_four` reported a new construction site for a **comment** that merely names `Value::List(items)` while explaining an adjacent arm. It now strips `//` comments — the rule its sibling `test_bool_before_int_converters_2212._strip_comments` already states: *comments must not decide the verdict, only real code does.* Rewording the prose to suit the grep would have been fixing the wrong end.

## Not more permissive than Django

* Every filter × every kind × every dict size on a hostile-keyed dict: **0 cells where djust emits a live payload Django does not.**
* `{{ p.keys|safe }}` is a genuine comparison rather than a parity nit — Django emits the view's repr **live** there too, so djust must emit the same thing, no more and no less. Asserted.
* The loop safe-key mapping stays gated off (`!normalised`) for a view, with an end-to-end probe through the production `_collect_safe_keys` for all three kinds against three hostile-key dicts.

### Two-build differential

Baselined against #2353's build (this PR's parent), **96,779 cells**:

```
agree BEFORE : 74959
agree AFTER  : 75001

newly AGREEING: 42
no longer agreeing: 0
  coincidental: 0
  REGRESSIONS : 0

live-payload leaks BEFORE : 182
live-payload leaks AFTER  : 182
  flagged as INTRODUCED   : 0
    LIVE (an unescaped tag opener) : 0
```

**0 regressions, 0 introduced live-payload leaks** — and unusually, 0 coincidental too.

The dict-view axis built only sequence-like cells, so the raising half of #2340 was invisible to it. Extended with `var-values`, `var-items-first`, `var-keys-last`, `var-keys-json`, `var-keys-slice`, `var-keys-safe`, `var-keys-pprint`, `var-keys-trunc` — deliberately **not** `random`/`timesince`/`timeuntil`, since that axis has no `NONDET` collapse and a nondeterministic cell would differ between two runs of the *same* build. Pinned by `test_the_differential_corpus_can_reach_the_raising_half`. Extended, never weakened: no safety-set name, branched-on whitespace character or hot-set filter was touched, and all three couplings still pass.

### The exhaustive filter sweep, and its own two blind spots

Every Django built-in × all three kinds. Its exemption is a **mechanical predicate**, not a name list: a cell is exempt only when the same filter over a **plain list of the same elements** diverges too, which proves the divergence belongs to the filter rather than to the view. It is asserted both to **fire** and to produce an **identical set across the three kinds** — a divergence that appeared for `keys` and not for `values` would be a property of the view, which is exactly what this change could have broken. Those eight exempt filters were confirmed pre-existing by running them against a plain list *and* a plain dict.

The sweep also had two blind spots of its own, and neither found its own bug:

* **one dict size.** `pluralize`'s two answers differ only at length 1, and the sweep ran a 2-entry dict. It now runs sizes 0-3.
* **one arg per filter.** `dictsort` was hidden completely: the chosen `'0'` is a case where **Django also fails** (a quoted `'0'` is a key lookup and a tuple has no key `'0'`), so the cell said "agree" about a filter that diverged for every arg that resolves. A second arg is now swept for the 16 filters whose behaviour turns on it.

Both were found by the or-pattern audit instead — which is the wrong way round, and is why the sweep was widened rather than just the code fixed.

### Gate-off

**14 mutations, 13 caught, 1 survivor, 0 invalid.** Same harness contract as #2353's.

| mutation | pytest | cargo |
|---|---|---|
| the view is a DictView rather than a plain List | 12 failed | 3 failed |
| the container NAMES itself in Display | 10 failed | 1 failed |
| the KIND decides the container name | 9 failed | 1 failed |
| `\|random` REFUSES a view | 1 failed | 0 failed |
| `\|json_script` REFUSES a view | 1 failed | 0 failed |
| a view IS iterable (`iter_values`) | 5 failed | 1 failed |
| `\|length` counts entries, not repr characters | 4 failed | 0 failed |
| `{% for %}` ITERATES the view | 10 failed | 0 failed |
| the safe-key mapping is GATED OFF (`normalised`) | **0 failed** | **0 failed** |
| `in` over a view compares its elements | 1 failed | 0 failed |
| a view is falsy when empty | 3 failed | 0 failed |
| the LEGACY display path is UNCHANGED | 0 failed | 1 failed |
| `pprint` spells the view's container | 2 failed | 0 failed |
| the loop cache keys a view apart from a list | 0 failed | 2 failed |

**The survivor is a question, and it was answered by measurement.** `normalised` gates the loop safe-key mapping, and flipping it changes **no output** — verified on this build *and* on the pre-#2340 build. `_collect_safe_keys` writes `p.<key>` while the mapping would look up `p.keys.<index>`, and those two spellings never meet for a view, so the gate is unreachable-but-correct defence in depth rather than a live mechanism. It is kept because `normalised` means "the sequence is not the value's own indexable elements", which is *true* of a view; setting it false would state something false. Documented as such in the test rather than dressed up as gated.

The first run had **four** survivors and every one was informative:

1. Two were the **same mis-targeted mutation** — its `old` text matched `legacy_display`'s arm, not `Display`'s. Retargeting it made it redden, and closing the legacy gap is what surfaced the wrong prose invariant above.
2. One **conflated two mechanisms**: it flipped `normalised` while leaving iteration untouched, so it could only ever measure the gate. Split into two, and the iteration half now reddens 10 tests.
3. One was **genuinely missing coverage** — the loop-cache tag.

## Adjacent divergences found, measured, filed — not fixed (#1079)

Each measured on **both** builds, so "pre-existing" is a measurement rather than a claim, and each pinned self-namingly.

* **#2361** — a `mark_safe` VALUE loses its mark through `{% for v in d.values %}`. djust over-escapes where Django emits live. Same two-spellings-of-one-path cause as the gate above, one level up: `_collect_safe_keys` writes `p.<key>`, the loop mapping needs `p.values.<index>`. Identical on the pre-#2340 build.
* **#2368** — `{% regroup d.values %}` silently misses. `resolve_tag_operand`'s non-pipe branch uses `Context::get`, and a view resolves in `Context::resolve` — the #2333 channel one operand form over. Identical on the pre-#2340 build; the pipe branch works, which is what proves this PR's `value_to_arg_string` arm is live.

## Verification

- `pytest tests/ python/tests/ python/djust/tests/ -n auto` — **14,405 passed**, 0 failed (post-rebase).
- `make test-rust` — 62 test binaries green.
- Rebased onto #2353 (0 behind it, 0 behind `origin/main`) before opening.

Closes #2340


johnrtipton:
Superseded by #2370 — same work, rebased onto `main`.

This PR was auto-closed when its base branch `fix/2339-typed-dict-keys` was deleted by #2353's squash-merge. GitHub refuses both `reopenPullRequest` and `updatePullRequest --base` once the base branch is gone, so retargeting in place was not possible and #2370 carries the work forward.

#2370 additionally pins the **#2360 context-builtins interaction**, which did not exist when this branch was opened.

Issue #2379, opened 2026-08-29
A custom tag handler's return value is emitted RAW where Django escapes it


johnrtipton:
Surfaced by the custom-TAG axis #2356 adds to
`scripts/filter-parity-differential.py`, on its first run: **12 cells where
djust emits a live payload Django escapes.**

## What the axis measured

A `{% simple_tag %}`-shaped handler registered through
`_rust.register_tag_handler` returns a value, and `renderer.rs`'s
`Node::CustomTag` arm inserts it into the output **verbatim**:

```rust
crate::registry::call_handler_with_py_sidecar(name, &resolved_args, &context_map, raw_py)
    .map_err(|e| DjangoRustError::TemplateError(format!("Custom tag '{}' error: {}", name, e)))
```

Django escapes the same return unless it carries `__html__`
(`SimpleNode.render` → `conditional_escape`). So for a handler as ordinary as

```python
@register.simple_tag
def greet(name):
    return f"Hello {name}"
```

Django renders `Hello &lt;img src=x onerror=alert(1)&gt;` and djust renders the
tag live.

The corpus cell is `@ctag ct-ident` (`{% ct_ident p %}`, a handler that returns
its input untouched — the sharpest probe, since Django escapes iff the FINAL
value lacks `__html__`), over the 12 hostile inputs. `cb-plain` is the same
shape on the BLOCK registry: a block handler returning a plain `str` wrapper is
emitted raw too.

This is the exact shape of #2290 with the arrow reversed. #2290 found that a
`SafeData` marker was stripped on the way IN to a custom filter — a
fail-CLOSED asymmetry. This is the way OUT, and it is fail-OPEN.

## Why it is not fixed in #2355/#2356's PR

The fix is one line in the Rust bridge — reuse the existing
`is_instance_of::<PyString>() && hasattr("__html__")` test
(`filter_registry.rs`, the #2290 mechanism) and escape when it is false — but
it makes every handler that returns HTML and does NOT `mark_safe` it start
emitting escaped text. djust ships about twenty such handlers and almost none
of them `mark_safe`:

| module | handlers |
|---|---|
| `python/djust/template_tags/static.py` | `{% static %}` |
| `python/djust/template_tags/url.py` | `{% url %}` |
| `python/djust/template_tags/flash.py` | `{% dj_flash %}` |
[...]

Issue #2345, opened 2026-08-28
The parity differential's corpus is blind on the argument axis — it reported 0 moved cells for #2328


johnrtipton:
`scripts/filter-parity-differential.py` gives every filter exactly ONE argument
(`FILTER_ARGS`), and it is always a VALID one. So the whole 89,616-cell corpus
is blind on the argument axis.

Measured, not asserted: #2328 changed what **every** argument-taking built-in
does with an unparseable or unresolvable argument, and the two-build compare
reported

```
cells        : 89616
agree BEFORE : 68042
agree AFTER  : 68042
newly AGREEING    : 0
no longer agreeing: 0
djust output changed in 0 cells
```

— zero moved cells, in both directions, for a change that moves 1,601 cells of
the same filters once the argument varies. The script's same-build guard even
fires on the identical counts and tells you to rebuild the baseline, which is
the wrong diagnosis: both builds were real.

Worse, the first pass of #2328 shipped **508 regressed cells** (a bool argument
mishandled, `{% if %}` no longer swallowing `VariableDoesNotExist`) that this
script and the full 13,933-green suite were both silent over. They were found
only by building a throwaway argument-axis sweep.

This is precisely the corpus-gap failure mode the script's own docstring names:

> That is the same corpus-gap failure mode that let it report clean over a live
> XSS (#2281): the tool is only ever as good as the shapes it builds.

## What to add

Widen `FILTER_ARGS` from one argument per filter to a small set of SPELLINGS
swept per filter. The throwaway harness used 19 and they were all load-bearing
at least once:

- valid: `"5"`, `5`, `"2.7"`, `2.7`, `" 5 "`, `"+5"`, `"1_0"`, `"-3"`, `-3`, `"0"`
- invalid: `"notanumber"`, `""`
- lookups: `missingvar`, `no.such.path`, a resolvable `known`
- literals Django resolves without a lookup: `True`, `None`, `7.`, `0x10`

Cross that with the existing INPUTS and the tag shapes and the corpus grows
~19x, so it probably wants its own `--axis argument` mode rather than being
folded into the default sweep.

## Also worth fixing while there

`render_both` catches `Exception`, which does NOT catch a `pyo3` panic
(`PanicException` derives from `BaseException`). The sweep aborts mid-run on
one — that is how #2343 was found, by the traceback rather than by a cell.
Catch `BaseException` so a panic becomes a comparable outcome instead of
ending the run.

PR #2381, opened 2026-08-29
fix(templates): sweep the six filter-expression tags and the custom-tag path (#2355, #2356)

### LINKED ISSUE #2355: The differential builds no cell for the six tags that take a filter-expression operand
Surfaced by the reachability manifest #2345 adds to
`scripts/filter-parity-differential.py`, on its first run.

The `tag` axis requires every tag in Django's live
`Engine.get_default().template_builtins`. The corpus sweeps three — `for`,
`if`, `with` — and exempts the other 22. Sixteen of those exemptions are
principled (a tag with no operand, or one needing a second template / a
request / the wall clock). **Six are not:**

    cycle  firstof  ifchanged  regroup  widthratio  filter

Each takes a FILTER-EXPRESSION operand, which is exactly the resolution path
#2325 was about: djust had one filter-aware resolver and four tags that
open-coded a bare variable lookup, so the chain was dropped and the tag
proceeded on the miss. #2325 fixed the four it found and added `for`/`if`/`with`
to the corpus; these six were never added, so the same class one tag over is
still unmeasured by the tool built to see it.

`regroup` is the sharpest case: #2333 fixed a filtered `regroup` source, and no
differential cell covers it.

## What to add

A `TAG_SHAPES` entry per tag, and delete its row from `TAGS_NOT_SWEPT` in
`scripts/filter-parity-differential.py`. The manifest's stale-exemption check
(`test_no_exemption_is_stale`) then fails until the row goes, so the exemption
cannot rot.

`widthratio` and `cycle` need care: their operands are not all chainable in the
same position, so each wants its own shape rather than one `@EXPR@` template.

### LINKED ISSUE #2356: The custom-TAG dispatch path is unmeasured — the #2290 gap, one registry over
Surfaced by the reachability manifest #2345 adds to
`scripts/filter-parity-differential.py`, on its first run.

The `entrypoint` axis requires every `_rust` function that renders a template or
changes how one renders. Three are exempt for the same reason, and it is the
reason #2290 existed:

    register_tag_handler
    register_block_tag_handler
    register_assign_tag_handler

#2290's finding was that `register_custom_filter` had been on the module the
whole time and **no corpus cell dispatched through it**, so the entire
custom-filter path — a real Python call across the PyO3 boundary that no
built-in reaches — was unmeasured, and every `SafeData` marker was stripped on
the way in without a single cell noticing. The custom-TAG registries are the
identical shape: a project's `@register.simple_tag` /
`@register.simple_block_tag` / assignment tag crosses the same boundary, and the
differential builds nothing that reaches any of the three.

## What to add

The same treatment `CUSTOM` got for filters (#2290): a handful of probe tags
registered on BOTH engines — with Django via a `Library` appended to
`Engine.get_default().template_builtins`, and with the Rust registry via
`register_tag_handler` / `register_block_tag_handler` /
`register_assign_tag_handler` — so the same source runs through both with no
`{% load %}`. Probes worth having, mirroring `cf_ident` / `cf_canon` /
`cf_cond`: one that returns its input untouched (the sharpest probe there is,
since Django escapes iff the FINAL value lacks `__html__`), one that
`mark_safe`s, and one that `conditional_escape`s.

Then delete the three rows from `ENTRY_POINTS_NOT_SWEPT`; the manifest's
stale-exemption check fails until they go.

johnrtipton:
Closes #2355. Closes #2356.

The reachability manifest (#2345) reported, on its first run, six Django
built-in tags exempt from the parity corpus with the reason **"TAKES A
FILTER-EXPRESSION OPERAND and is not swept"** — an admission rather than a
property — and three `_rust` entry points exempt because nothing dispatched
through them. Both exemption sets are gone. Sweeping them found four
divergences in the six tags, three of them silent, and one finding on the
custom-tag path that is **more permissive than Django** and is filed rather
than fixed.

## Issue × file × test

| issue | what changed | covering tests |
|---|---|---|
| #2355 corpus | `TAG_SHAPES` grows nine shapes; `_swept_tags` reads the tag out of each shape's SOURCE, not its dict key; six `TAGS_NOT_SWEPT` rows deleted | `TestTheCorpusReachesWhatTheseIssuesSaidItCouldNot`, `test_2355_six_tags_took_a_filter_operand_and_were_exempt` |
| #2355 `widthratio` | `renderer.rs` — new `width_ratio` / `py_int` / `py_round_to_string`, Django's `WidthRatioNode.render` arm for arm | `TestWidthRatio` (18) |
| #2355 `as var` | `parser.rs` — `split_asvar`; `Node::{WidthRatio,FirstOf}` grow `asvar`; `renderer.rs` — new `sibling_updates` helper replacing three copied `AssignTag` arms | `TestWidthRatio`, `TestFirstOf`, `renderer.rs::asvar_standalone_tests` |
| #2355 `cycle` | `renderer.rs` — an unresolved operand renders `""`, not its own source text | `TestCycle` |
| #2355 `regroup` | `template_tags/regroup.py` — `_grouper` / `_apply_filters` run the `by` operand's filter chain per item | `TestRegroupByOperand` |
| #2356 | six probes registered on both engines; `@ctag` cell family; three `ENTRY_POINTS_NOT_SWEPT` rows deleted | `TestTheCustomTagPathIsReachableAtAll`, `TestKnownDivergencesOnTheCustomTagPath` |
| harness | `render_both` hands Django a COPY of the context | `test_render_both_hands_the_two_engines_the_SAME_context` |

## What the sweep found

Every value below is Django 5.2.16, run rather than remembered.

**`{% widthratio %}` answered `0` for every non-numeric operand** where Django
answers `""` — **16,006 of that shape's 17,298 cells**. Three more defects in
the same arm: half-away-from-zero rounding where Python's `round` is
half-to-**even** (`{% widthratio 1 2 5 %}` was `3`, Django's is `2`);
`i64::MAX` for a non-finite ratio, where `round(inf)` is an `OverflowError`
Django catches into `""`; and `0` rather than a raise for a non-numeric final
argument — that operand goes through Python's `int()`, so `int("100.6")` raises
where `float("100.6")` does not.

**`{% widthratio … as w %}` / `{% firstof … as v %}` rendered the value Django
assigns silently**, and bound nothing: `as` and the name were parsed as two
more operands. `firstof` binds `render_value_in_context(...)` — a `SafeString`,
measured — so the binding carries the grant; `widthratio` binds a plain `str`
and does not.

**`{% cycle nope 'z' %}` echoed the operand's own source text onto the page.**
Django compiles each operand with `compile_filter` and a missing variable
resolves to `string_if_invalid`, so the answer is `""`. The comment being
deleted claimed the opposite: *"output the raw name (Django behavior)"*.

**`{% regroup p by k|upper as g %}` dropped the `by` chain** and grouped every
row under `None` — one group where Django builds three, every
`{{ x.grouper }}` empty. #2333 fixed the SOURCE operand; `by` is a filter
expression too.

`ifchanged` and `filter` stay unsupported by the Rust engine and their cells
are built anyway — "no cell exists" and "every cell is the same refusal" are
different states, and only the second goes red the day someone implements the
tag and gets its escaping wrong.

## The #2356 finding — not fixed here

The custom-TAG axis reports **12 cells where djust emits a live payload Django
escapes**: a handler's return value is inserted RAW, where Django's
`SimpleNode.render` `conditional_escape`s any return lacking `__html__`. That
is #2290's asymmetry with the arrow reversed — #2290 was the way IN and
fail-closed; this is the way OUT and fail-OPEN.

It is filed as **#2379** with the inventory of ~20 built-in handlers a fix has
to audit, because escaping handler output makes every one of them emit escaped
markup unless it `mark_safe`s. `TestKnownDivergencesOnTheCustomTagPath` pins
the current behaviour, so that change cannot land silently: every one of those
tests goes red the day the bridge starts escaping.

## Manifest, before and after

```
                 before                          after
tag          25 required, 22 exempt, 0 MISSING   31 required, 16 exempt, 0 MISSING
entrypoint   10 required,  7 exempt, 0 MISSING   10 required,  4 exempt, 0 MISSING
```

Corpus: **115,395 → 281,121** cells.

## Two-build differential, against a rebuilt `origin/main`

```
cells        : 281121
agree BEFORE : 130244
agree AFTER  : 179919
newly AGREEING: 49675
REGRESSIONS : 0
cells where djust PANICKED before/after: 0 / 0
live-payload leaks BEFORE : 648
live-payload leaks AFTER  : 648
  flagged as INTRODUCED   : 0
    LIVE (an unescaped tag opener) : 0
```

All 59,829 moved cells are on the `tag` axis; every other axis moved 0. The 12
`ctag` LIVE cells are **identical on both builds** (the #2379 finding, newly
visible rather than newly caused), as are the 42 on `cycle`/`firstof`/
`firstof-as`, which are the already-measured `add:"1"|safe` divergence reaching
three more operands.

## Gate-off

**Twenty mutations, all twenty KILLED.** The harness asserts the mutation text
is found exactly once, asserts the mutated source differs, counts pytest
*errors* separately from failures, distinguishes a cargo compile failure from a
cargo test failure, and verifies every restore is byte-identical — and it
earned that last one twice, refusing with `MUTATION TEXT FOUND 0 TIMES` after
the rebase reflowed two hunks rather than reporting a green over a mutation
that no longer applied.

The first pass had four survivors and each was a different answer:

| survivor | answer |
|---|---|
| `py_int` strictness | a semantic no-op for the tested inputs — `"abc"` fails both `int()` and `float()`. Added the case that separates them (`"100.6"`). |
| `widthratio` / `firstof` render-arm `asvar` check | shadowed by `sibling_updates` on every template-reachable path (verified: for-body, if-branch, with-block, spaceless, nested). Kept — it is the same defensive shape `Node::AssignTag`'s standalone arm has — and given the Rust unit test that makes it reachable, each with a gate-off sibling asserting the same node WITHOUT `as` renders the value. |
| `register_tag_handler` dispatch | the mutation removed one of three call sites, so the source grep still found the symbol. Strengthened to remove all three, and a companion mutation re-adds the exemption rows. |

Self-review then found a **third mechanism** the first pass had missed, by
reading `Node::AssignTag`'s dep-extraction comment: an `as <var>` node mutates
the context for later siblings, so it needs the same `"*"` wildcard dependency,
or `render_nodes_partial` skips it whenever its own operands are unchanged and
the binding never happens. Fixed, gated on `asvar.is_some()` so the emitting
form keeps its precise deps, and covered by
`an_asvar_node_carries_the_assign_tags_wildcard_dependency` — which asserts
both halves, so it cannot pass by making everything a wildcard. That is
mutations 19 and 20.

## Verification

- `pytest tests/ python/tests/ python/djust/tests/ -n auto` — **15,104 passed, 220 skipped**
- `TAG_SHAPES` 3 → 12 entries; `TAGS_NOT_SWEPT` 22 → 16; `ENTRY_POINTS_NOT_SWEPT` 7 → 4 (all counted mechanically)
- `make test-rust` — clean
- `cargo clippy --workspace --all-targets -- -D warnings` — 0
- `cargo fmt --all --check`, `ruff check`, `ruff format --check` — clean
- `scripts/check-changelog-test-counts.py`, `scripts/check-doc-snippets.py` — exit 0
- 0 commits behind `origin/main`


johnrtipton:
## Self-review found a third mechanism the first pass missed — force-pushed

Reading `Node::AssignTag`'s dep-extraction arm, whose comment states the reason
in full:

> Because an assign tag mutates the context for subsequent sibling nodes in
> unknowable ways, also emit the `"*"` wildcard — any change may alter
> downstream rendering.

An `{% widthratio … as w %}` / `{% firstof … as v %}` node does **exactly
that**, and the first pass gave it a precise dep set. Under partial render
(`render_nodes_partial`) the node is skipped whenever its own operands are
unchanged — so the binding never happens and every sibling reading the name
sees nothing. That is the same class the convergence onto `sibling_updates`
was for: the second context-mutating node needed the second half of what the
first one has, not just the render half.

`parser.rs` now emits `"*"` for both, **gated on `asvar.is_some()`** — the
emitting form has no such effect and must keep its precise deps, or every
`{% widthratio %}` in a partially-rendered template becomes a wildcard.

Covered by `an_asvar_node_carries_the_assign_tags_wildcard_dependency`, which
asserts both halves: the `as` forms carry `"*"`, and the emitting forms carry
their operand and **not** `"*"` — the gate-off sibling, so the test cannot pass
by making everything a wildcard.

### Re-verified end to end on the amended tree

- **Gate-off: 20 mutations, all 20 KILLED** (was 18/18; the two new rows are
  the wildcard on each node, defended by the Rust test and run through cargo).
- Two-build differential against a rebuilt `origin/main` — **unchanged**:
  49,675 newly agreeing, 0 regressions, 0 introduced live-payload leaks,
  0 newly panicking cells, all 59,829 moved cells on the `tag` axis.
- `pytest tests/ python/tests/ python/djust/tests/ -n auto` — 15,104 passed,
  220 skipped.
- `make test-rust`, `cargo clippy --workspace --all-targets -- -D warnings`,
  `cargo fmt --all --check`, `check-changelog-test-counts.py` — all clean.
- 0 commits behind `origin/main`.

The previous push was green on all checks; this is the same tree plus the
wildcard fix and its test.

Issue #2377, opened 2026-08-29
{% for a,b in x %} — comma without a space — silently renders nothing


johnrtipton:
Found while measuring #2361/#2363 (PR pending) and filed separately per CLAUDE.md #1079.

## The cell

`{% for a,b in x %}` — the comma WITHOUT a space — is not parsed as tuple
unpacking. Django accepts both spellings.

```
{% for a,b in p %}[{{ b }}]{% endfor %}     p = [("x", "y")]
  django  '[y]'
  djust   '[]'                <-- the whole loop body renders nothing

{% for a, b in p %}[{{ b }}]{% endfor %}    same input
  django  '[y]'
  djust   '[y]'               <-- agrees
```

Django's `do_for` splits the variable list on `","` after joining the tokens,
so `a,b`, `a, b` and `a ,b` are all the same three-name loop. djust's parser
appears to split on whitespace only, so `a,b` is read as ONE variable name
literally spelled `a,b`, which never resolves — and the body renders empty
rather than raising.

## Severity

Silent empty region — the worst shape a template engine has, and the same
class as #2325 (`{% for x in p|slice %}` rendering nothing) and #2334
(`{% for k in d %}` rendering nothing). No escaping or safety dimension: the
values simply never reach the page.

## Where to look

The `{% for %}` tag parse in `crates/djust_templates/src/parser.rs`, where
`var_names` is built. The fix is to split the variable segment on `,` (and
strip whitespace around each name) rather than on whitespace alone.

## Worth checking at the same time

Whether any OTHER tag that takes a comma-separated name list has the same
whitespace assumption — `{% regroup … as %}`, `{% with %}`'s multi-assignment
form, `{% cycle … as %}`. Grep the parse sites rather than the callers
(CLAUDE.md, "grep for the SINK").

## How it was found

Not by the differential: `PATH_SHAPES` and `TAG_SHAPES` in
`scripts/filter-parity-differential.py` write every loop with `{% for k, v in
… %}`, the spaced spelling, so the corpus cannot construct a single cell that
shows this. A corpus-gap of the same shape as #2325 and #2334 — worth adding
the unspaced spelling to `PATH_SHAPES` with the fix.

PR #2397, opened 2026-08-29
fix(templates): {% for a, b in x %} refuses an arity mismatch, as Django does (#2387)

### LINKED ISSUE #2387: {% for a, b in x %} does not raise Django's unpack-arity ValueError
Found while fixing #2377 (PR #2386) and filed separately per CLAUDE.md #1079 —
measured on two builds, PRE-EXISTING, and deliberately not touched there.

## The cell

Django's `ForNode.render` computes `len(item)` (a `TypeError` counts as 1) and
raises when it does not equal the loop-variable count:

```python
raise ValueError("Need %d values to unpack in for loop; got %d. " % (num_loopvars, len_item))
```

djust pads the extra names with `Value::Missing` and renders.

```
{% for a, b in p %}[{{ a }}={{ b }}]{% endfor %}    p = "abc"
  django  <<ValueError: Need 2 values to unpack in for loop; got 1. >>
  djust   '[a=][b=][c=]'

{% for a, b, c in p %}[{{ a }}]{% endfor %}         p = [("x", "y")]
  django  <<ValueError: Need 3 values to unpack in for loop; got 2. >>
  djust   '[x]'
```

## It is not the comma fix

Both spellings are affected identically and both were affected BEFORE #2377:
on the pre-#2377 build, `{% for a ,b in p %}` over `["a", ["b", "c"]]`
rendered `[a=][b=]` where Django raises. #2377 changes only which SPELLING
reaches the code, which is why the two-build differential for PR #2386 reports
55 cells that moved from "rendered with empty variables" to "rendered with the
values" while Django raises for both.

Direction: djust is MORE PERMISSIVE — it renders where Django refuses. No
escaping dimension: PR #2386's differential reports the live-payload leak
counter identical on both builds in both halves, and every hostile value in
those cells comes out escaped.

## Pinned

`python/tests/test_for_unpack_comma_spelling_2377.py::TestTheUnpackArityDivergenceIsNamedNotFixed`
asserts Django raises AND djust renders, for the unspaced spelling, the spaced
spelling, and the three-over-two case. It goes red the day this is closed and
names itself as the thing to move.

## Direction, if it is taken

The renderer's tuple-unpacking arm (`crates/djust_templates/src/renderer.rs`,
the `var_names.len() > 1` branch) needs Python's `len()` semantics for a
`Value` — `String` by code point, `List`/`Tuple`/`DictView` by element,
`Value::Object` by key count unless it is a serialized model (`object_str()`
is `Some`, where `len(model)` is a `TypeError`), everything else 1 — and to
return an error rather than filling the extras with `Missing`. That length
rule already exists once, in the `length` filter
(`crates/djust_templates/src/filters.rs:660`), so this is a
grep-for-the-sink / one-statement-of-the-rule question rather than a new one
(#1646).

Refs #2377, #1079.

johnrtipton:
Closes #2387

## The cited defect

Django's `ForNode.render` checks arity BEFORE unpacking and raises:

```python
try:               len_item = len(item)
except TypeError:  len_item = 1
if num_loopvars != len_item:
    raise ValueError("Need %d values to unpack in for loop; got %d. " % …)
unpacked_vars = dict(zip(self.loopvars, item))
```

djust padded with `Value::Missing` and rendered. The check is now the first
thing that branch does, with Django's message verbatim — trailing space and
all. It crosses to Python as a `RuntimeError` rather than Django's
`ValueError`, as **every** djust render error does (`DjangoRustError`'s own
docs already say so for the `VariableDoesNotExist` variant). Both engines
refuse the render, which is the property the issue is about.

## The half the issue does not name

`zip` **iterates** the item; it does not index it. So an item whose length
*does* match unpacks by Python's iteration, and both of these PASS the arity
check — the check alone would have left them wrong:

| `p` | Django | djust before |
|---|---|---|
| `["ab"]` | `a="a"`, `b="b"` | `a="ab"`, `b=` |
| `[{"x": 1, "y": 2}]` | `a="x"`, `b="y"` | `a="{'x': 1, 'y': 2}"`, `b=` |

The second row rendered the dict's own repr into `{{ a }}`.

The new arm grants **no** safety to any component, deliberately.
`_collect_safe_keys` spells a dict BY KEY NAME, so the positional
`<expr>.<index>.<i>` lookup the sequence arm uses would be the #2334
collision here — a live XSS whenever keys are attacker-controlled. Pinned by
`test_an_unpacked_dict_key_is_escaped`.

## Where the length rule lives

The issue points at the `length` filter as the place the rule already exists,
and it is right — but the two call sites disagree about the **fallback**.
Django writes `except TypeError: return 0` in `defaultfilters.length` and
`except TypeError: len_item = 1` in `ForNode.render`. So `filters::python_len`
returns `Option<usize>` (`None` where Python raises) and each site picks its
own; collapsing them into one "length with a fallback baked in" would have
made one of them wrong (#1646).

`{{ p|length }}` is now `python_len(value).unwrap_or(0)` — byte-for-byte the
same answers for every variant, pinned by
`test_the_length_filter_is_unchanged_by_the_refactor`.

The short-item padding branch is **deleted** rather than kept as a belt: after
the arity check the two lengths are equal by construction, so the sequence arm
is `var_names.iter().zip(tuple_items)` — Django's own `zip` — with no
unreachable `else` to shadow it (#2233).

## Moved, not duplicated

`TestTheUnpackArityDivergenceIsNamedNotFixed` in
`test_for_unpack_comma_spelling_2377.py` said it "goes red the day this is
closed, and names itself as the thing to move". Its three cells are now the
first three rows of `TestBothEnginesRefuseAnArityMismatch`. What stays behind
is the #2377 half — every comma spelling raises the *same* message, so a
regression in the split would show as one spelling refusing and another not.

## The 5,600-cell sweep moved 20 cells

`TestDictIterationRandomised` builds `{% for k, v in p.items %}` over random
dicts that may carry an `items` KEY (mapping-before-attribute, so `p.items` is
that key's value). 20 of those cells moved from `django-raised` (Django raised,
djust rendered) to a new `both-raised` row. The new predicate compares the
**messages** — djust's must end with Django's — rather than the exception
classes, so it cannot absorb a djust failure that merely coincides with a
Django one.

## Gate-off

Each mutation asserted found exactly once, source changed, Rust rebuilt between
iterations, restore verified byte-identical by sha256, pytest `N error` counted
separately from `N failed`, cargo's `error: test failed` distinguished from a
compile break.

| mutation | pytest | cargo |
|---|---|---|
| (unmutated baseline) | 0 failed | 0 failed |
| M1 arity check never fires | **18 failed** | |
| M2 arity fallback `1` → `0` | **5 failed** | |
| M3 non-sequence arm binds only the 1st name | **4 failed** | |
| M4 `python_len`: str counts BYTES | **2 failed** | |
| M5 `python_len`: a model is a dict | 0 failed | **1 failed** |
| M6 `\|length` fallback `0` → `1` | **6 failed** | |

**M5 surviving pytest is the correct answer, not a gap.** From Python a
serialized model and a dict carrying a literal `"__str__"` key are
indistinguishable — the marker IS that key — so no template-level test can
separate the readings. The `Value`-level pin
(`python_len_refuses_a_serialized_model_but_not_a_dict`) is the one that can,
and it reddens.

## Two-build differential — and the one number that needs reading carefully

Baseline rebuilt from `origin/main` (`3b019bbe`), 314,847 corpus cells, two
genuinely different `_rust` builds (`2f60533…` → `2a37732…`), reachability
manifest 0 MISSING on every mechanical axis:

| | |
|---|---|
| path cells whose djust output MOVED | **135** |
| newly AGREEING | **0** |
| no longer agreeing / REGRESSIONS | **0 / 0** |
| introduced live-payload leaks | **0 live, 0 escaped** (37 before, 37 after) |
| panics | 0 before, 0 after |

**0 newly agreeing is not "nothing happened."** All 135 moved cells classify
mechanically as one shape:

```
135  ('django-raises-unpack', 'djust-rendered', 'djust-raises-same-message')
```

— Django raises `ValueError("Need N values to unpack…")`, djust *rendered a
padded region* before and *raises the same message* after. The corpus records
a raise as `<<EXC {type}>>`, so `ValueError` and `RuntimeError` do not compare
equal and the cell cannot be counted as agreeing however faithful the message
is. Every one of the 135 moved in the **less-permissive** direction; none moved
the other way, and none is a leak.

## Out of scope, found while measuring

`{{ forloop.counter }}` renders empty in djust — for the single-variable branch
too, so it is unrelated to this change and predates it.
`crates/djust_templates/src/loop_cache.rs:41` already says so in as many words:
*"The Rust renderer does not currently implement `forloop`."* Not touched
(#1079).

## Verification

- `pytest tests/ python/tests/ python/djust/tests/ -n auto` — **15566 passed**, 221 skipped, 0 failed
- `make test-rust` — 0 failures


johnrtipton:
Follow-up filed for the `forloop.*` gap this PR's measurement surfaced: #2402. It is unrelated to the unpack arm (it affects the single-variable branch identically) and already documented in `crates/djust_templates/src/loop_cache.rs:41`, but had no issue number.

Issue #2385, opened 2026-08-29
{% regroup %} over a non-empty string builds zero groups where Django builds one — 8,505 corpus cells


johnrtipton:
`{% regroup %}` over a non-empty STRING builds **zero** groups where Django builds one containing every character. `{% for %}` over the same string iterates correctly, so it is regroup-specific: djust's regroup reaches only `List`/`Tuple`.

## Measured, Django 5.2.16

```
{% regroup p by k as g %}[{{ g|length }}]{% for x in g %}("{{ x.grouper }}":{{ x.list|length }}){% endfor %}

  p="abc"   django: '[1]("None":3)'    djust: '[0]'
  p="1"     django: '[1]("None":1)'    djust: '[0]'
  p=""      django: '[0]'              djust: '[0]'   <- agrees
  p=["a","b"]  django: '[1]("None":2)' djust: '[1]("None":2)'   <- agrees

{% for c in p %}[{{ c }}]{% endfor %}   p="abc"
  django: '[a][b][c]'   djust: '[a][b][c]'   <- the control: iteration itself is fine
```

Django's `RegroupNode.render` does `obj_list = self.target.resolve(context)` and then iterates it; a `str` is iterable, so it groups its characters and `k` resolves to nothing, giving one group with grouper `None`.

## Scale

**8,505 cells** on the parity corpus, measured on `origin/main` at `94fa8b26` — every `{% regroup p|<expr> by k as g %}` cell whose operand is one of the `s-*` string inputs. It is the largest single divergence class the corpus currently carries on the tag axis.

## How it surfaced

Not by anyone looking at `regroup`. #2359 corrected `add`'s third branch to return Django's `""` instead of echoing the value, and 14 corpus cells that had **agreed by coincidence** stopped agreeing:

```
{% regroup p|add:"1"|add:"1" by k as g %}[{{ g|length }}]     p=['<b>','x']
  django: '[1]'   djust before #2359: '[1]'   djust after: '[0]'
```

Before, `add` handed `regroup` the **list** unchanged (the bug #2359 fixes), and regroup handles a list correctly — so two wrongs cancelled, the #2272 shape. After, the operand is the string `"1"` that Django also computes, and this bug shows through. The chain itself is correct on both engines at every step:

```
{{ p|add:"1" }}            django '' djust ''    <- agrees
{{ p|add:"1"|add:"1" }}    django '1' djust '1'  <- agrees
```

[...]

PR #2398, opened 2026-08-29
fix(templates): a quoted literal in a FILTER ARGUMENT is SafeData (#2389)

### LINKED ISSUE #2389: A quoted literal in a FILTER ARGUMENT is not SafeData ({{ p|default:"<b>" }})
Found while fixing #2376 and filed separately per CLAUDE.md #1079 — the same
rule, one resolver over.

## The cell

```
{{ p|default:"<b>" }}     p = ""
  django  '[<b>]'
  djust   '[&lt;b&gt;]'
```

Django resolves a filter ARGUMENT through `Variable` too — `FilterExpression`
compiles each argument into one — so `Variable('"<b>"').literal` is
`mark_safe(unescape_string_literal(...))` and the argument arrives at `default`
as a `SafeString`. `default` returns that object unchanged, so
`render_value_in_context`'s `conditional_escape` leaves it alone and the markup
is LIVE.

djust's filter-argument channel carries a bare `&str` with no safety flag
beside it, so the returned value is a plain `String` and the emit arm escapes
it.

## Direction

OVER-escaping — a lost capability, never a leak. Which is why it is filed
rather than folded into #2376.

## Why it is a separate change

#2376 states the literal rule once, in `renderer::django_literal`, and calls it
from the two resolvers that take a bare *expression*:
`renderer::get_value_safe` (the tag-operand channel) and `Node::Variable` /
`Node::InlineIf` (the emit arms). The filter-argument channel is a THIRD
resolver with a different signature: `filters::apply_filter_full_safe` takes
`arg: Option<&str>` and every one of the 57 built-in arms reads it as text.
Threading a grant through it is a signature change across all of them, not an
arm in an existing `match`.

It also needs a decision #2376 did not have to make: Django's rule is that a
`is_safe=True` filter re-marks its OUTPUT when the *input* is `SafeData`
(`FilterExpression.resolve`'s `if getattr(func, "is_safe", False) and
isinstance(obj, SafeData)`), and the argument's own safety only reaches the
page for filters that RETURN the argument (`default`, `default_if_none`,
`yesno`, `join`'s separator, …). Enumerating which those are is the work.

## Pinned

`python/tests/test_template_literals_2376.py::TestKnownNarrowerThanDjango::
test_a_literal_in_a_FILTER_ARGUMENT_is_not_granted` asserts Django's answer AND
djust's, so it goes red the day this is closed and names itself as the thing to
move.

Refs #2376, #1646, #1079.

johnrtipton:
Closes #2389

> **Stacked on #2396, and this is a measured dependency rather than a
> convenience.** On its own branch this change moves exactly four corpus cells
> out of agreement — `{% regroup p|default:"D"|safeseq by k as g %}` and its
> `escapeseq` / `default_if_none` siblings, over `p = None`. The mechanism is
> `mark_safe` on a NON-string being `SafeString(str(obj))`: once `default`'s
> output is `SafeData`, `FilterExpression.resolve`'s `is_safe` re-mark collapses
> `safeseq`'s LIST into its repr string — which is what Django does, and what
> djust now does too. The collapsed operand then reaches `{% regroup %}` as a
> STRING, and regroup-over-a-string built zero groups (#2385). Two wrongs had
> been cancelling, the #2272 shape: djust reached Django's `{{ }}` bytes by a
> different route (an uncollapsed `Value::List`) and the tag channel exposed it.
> #2396 fixes that, and with both branches present the four cells agree and the
> combined differential is **17,805 newly agreeing / 0 regressions**. #2396 is
> merged into this branch so it is green on its own; the mechanism is pinned by
> `test_the_grant_collapses_a_downstream_sequence_exactly_as_django_does`.

## The rule

`FilterExpression.resolve` marks a **constant** filter argument safe:

```python
for lookup, arg in args:
    if not lookup:
        arg_vals.append(mark_safe(arg))
    else:
        arg_vals.append(arg.resolve(context))
```

So `default` hands the `SafeString` back unchanged and `conditional_escape`
leaves the markup live. djust escaped it — over-escaping, a lost capability,
never a leak.

## Two of the issue's premises did not survive, and both shrank the work

**"Threading a grant through it is a signature change across all of them."**
`apply_filter_full_safe` has taken `arg_was_quoted: bool` since #2202,
`builtin_produced_safe` already receives it, and the `add` arm already reads
it. The change is two arms and one `if`.

**"…filters that RETURN the argument (`default`, `default_if_none`, `yesno`,
`join`'s separator, …)"** — wrong in both directions. Running all 57 built-ins
against a hostile quoted-literal argument, rather than reading 57 bodies:

| filter | in the issue's list? | actually emits a literal argument live? |
|---|---|---|
| `default` | yes | **yes** — fixed here |
| `default_if_none` | yes | **yes** — fixed here |
| `json_script` | **no** | **yes** — `format_html` puts a literal `id` in raw; fixed here |
| `join` | yes | yes, and djust **already agreed** — `conditional_escape(arg)` leaves a `SafeString` separator alone |
| `yesno` | yes | **no** — `str.split(",")` on a `SafeString` yields plain `str`s |
| `pluralize` | (implied) | **no** — same reason |

`TestTheEnumerationIsMechanical` re-derives that set from the **live** registry,
so a Django release that adds a fifth fails here rather than drifting.

## The gate, and why it is not a leak

`arg_was_quoted` is exactly Django's `not lookup`. A quoted literal lives in the
**template source**, which its author controls as completely as any raw HTML
they type there. The variable channel — the half that can carry data — is
unchanged, and measured clean across every built-in on both engines, before and
after: **0 cells where either engine emits the payload live**.

`crates/djust_templates/tests/test_xss_prevention.rs::filter_chain_default_still_escapes`
pinned the old behaviour. It is replaced by **three** tests rather than deleted:
the literal is live (parity), and a variable `default` argument and a variable
`json_script` id are both still escaped (the security half).

## Three premises of my own first draft, corrected by live Django

Recorded in the test bodies, because the intuitive reading is wrong in all
three:

- `{{ ""|default:"<b>"|upper }}` is `&lt;B&gt;`, not `<B>` — `upper` is
  registered `is_safe=False`, so the grant is lost. `lower` is `is_safe=True`
  and keeps it.
- `{{ ""|default:"<b>"|escape }}` is `<b>` — Django's `escape` filter is
  `conditional_escape(value)`, a **no-op** on `SafeData`. `force_escape` is the
  unconditional one.
- the derived set is **four** filters, not three: `join` belongs in it even
  though nothing about `join` moved.

## Measured, live Django 5.2.16

| sweep | before | after |
|---|---|---|
| literal argument — 57 filters × 2 payloads × 6 value shapes | 378 diverging, **18** of them Django-emits-live | 360 diverging, **0** Django-emits-live |
| variable argument — every built-in × 5 value shapes | 0 live on either engine | 0 live on either engine |
| chain — 3 granted filters × every built-in with a valid 2nd argument, 336 cells | — | **0** more-permissive cells, live or escaped |

The 360 that still diverge are pre-existing classes this PR does not touch:
djust does not validate filter **arity** (Django raises `TemplateSyntaxError`
for a no-arg filter given an argument), `timesince`/`timeuntil` echo an
unreadable value, and `dictsort`/`yesno` edge cases.

## Gate-off

Each mutation asserted found exactly once, source changed, Rust rebuilt between
iterations, restore verified byte-identical by sha256, pytest `N error` counted
separately from `N failed`.

| mutation | result |
|---|---|
| (unmutated baseline) | 0 failed |
| M1 `default`'s falsy branch loses the grant | **11 failed** |
| M2 `default_if_none` loses the grant | **3 failed** |
| M3 `json_script` always escapes its id | **3 failed** |
| M4 grant ignores quoting (variable args too) | **2 failed** |
| M5 `json_script` grants a VARIABLE id too | **2 failed** |

M4 and M5 are the **over-permissive** mutations, and both are caught by the
variable-channel tests rather than by any cell the fix is about.

## Two-build differential

Baseline rebuilt from `origin/main` (`3b019bbe`), 314,847 corpus cells,
reachability manifest 0 MISSING on every mechanical axis. Both readings, because
the honest one is the pair:

| | this change alone (`2f60533…` → `71c007b…`) | with #2396 (`2f60533…` → `9aba0ed…`) |
|---|---|---|
| newly AGREEING | 34 | **17,805** |
| no longer agreeing | 4 | 71, all `coincidental` |
| **REGRESSIONS** | **4** (the `safeseq`-collapse cells above) | **0** |
| introduced live-payload leaks | **0 live, 0 escaped** | **0 live, 0 escaped** |
| panics | 0 → 0 | 0 → 0 |

The 37 pre-existing leak cells are unchanged in both columns — this change
closes none and introduces none. The branch as pushed is the right-hand
column.

## Verification

- `pytest tests/ python/tests/ python/djust/tests/ -n auto` — **15892 passed**, 221 skipped, 0 failed
- `make test-rust` — 0 failures


johnrtipton:
Three follow-ups filed from this PR's 57-filter sweeps, none of them this change's doing and all measured on clean `main` at 3b019bbe: #2400 (48 of 57 built-ins accept a wrong argument COUNT — the dominant residue of the 360 cells that still diverge here), #2401 (`yesno` with a one-part argument), #2403 (`get_digit`'s pass-through branch drops the input's grant — the same shape as `default`'s truthy branch, reproduced against a `mark_safe`d context value rather than through this PR's chain).

Issue #2328, opened 2026-08-28
wordwrap accepts an unparseable width where Django raises, and the fallback belongs to all 26 argument-taking filters


johnrtipton:
Split out of #2293 (PR pending), which ported `wordwrap` to `textwrap.TextWrapper` and fixed the `width <= 0` half of the argument contract. The *parsing* half is left as it was, deliberately, and this is the record of why.

## What Django does

```python
@register.filter(is_safe=True)
@stringfilter
def wordwrap(value, arg):
    return wrap(value, int(arg))
```

`int(arg)` raises `ValueError` for anything it cannot parse, and the filter takes **two** required parameters, so `{{ p|wordwrap }}` is a `TemplateSyntaxError` at parse time.

## What djust does

`crates/djust_templates/src/filters.rs`, the `"wordwrap"` arm:

```rust
let width = match arg {
    None => 75,
    Some(a) => a.trim().parse::<i64>().unwrap_or(75),
};
```

So three cells diverge:

| template | Django | djust |
|---|---|---|
| `{{ p\|wordwrap:"nope" }}` | `ValueError` | wraps at 75 |
| `{{ p\|wordwrap:"5.5" }}` | `ValueError` | wraps at 75 |
| `{{ p\|wordwrap }}` | `TemplateSyntaxError` (parse time) | wraps at 75 |

A width of `0` or a negative one DOES now raise, with Django's own message, because that is `textwrap._wrap_chunks`'s guard rather than `int()`'s — pinned in `python/tests/test_wordwrap_parity_2293.py::TestWidthArgument`.

## Why #2293 did not close this

Two reasons, and the first is the one that matters:

1. **A bare-identifier filter argument that does not resolve arrives at the filter as its own NAME.** `apply_filter_full_safe` resolves `{{ p|wordwrap:w }}` through `Context::resolve` and, on a MISS, falls back to the raw string `"w"` — a documented deliberate divergence from Django, whose comment says raising "would turn a silent wrong-output bug into a site-wide 500 on upgrade". Raising on an unparseable argument would escalate exactly that class: a template with a typo'd or conditionally-absent context variable would go from rendering at 75 to 500ing. That decision belongs to the whole 26-filter argument surface, not to `wordwrap` alone.

2. The no-argument case is a *parser* change, not a filter change — Django rejects the arity before any filter runs — so it is a different mechanism again.

## What to decide

[...]

PR #2408, opened 2026-08-29
fix(templates): a filter given the wrong ARGUMENT COUNT is refused (#2400)

### LINKED ISSUE #2400: 48 of 57 built-in filters accept a wrong ARGUMENT COUNT where Django raises TemplateSyntaxError
Found while draining #2389 and filed separately per CLAUDE.md #1079 — measured
on `main` at `3b019bbe`, PRE-EXISTING, and not touched there.

## The cells

Django validates a filter's argument COUNT at compile time
(`FilterExpression.__init__`'s `args_check`), so both of these are
`TemplateSyntaxError` and the template never renders. djust renders both.

```
{{ p|upper:"x" }}     p = "abc"
  django  <<TemplateSyntaxError: upper requires 1 arguments, 2 provided>>
  djust   'ABC'

{{ p|default }}       p = "abc"
  django  <<TemplateSyntaxError: default requires 2 arguments, 1 provided>>
  djust   'abc'
```

## Scale, measured

Sweeping every one of Django's 57 live built-ins:

* **28** raise on an EXTRA argument (`{{ p|<f>:"x" }}` where `<f>` takes none)
  and djust renders — `addslashes`, `capfirst`, `escape`, `escapejs`,
  `escapeseq`, `filesizeformat`, `first`, `force_escape`, `iriencode`, `last`,
  `length`, `linebreaks`, `linebreaksbr`, `linenumbers`, `lower`, `make_list`,
  `phone2numeric`, `pprint`, `random`, `safe`, `safeseq`, `slugify`,
  `striptags`, `title`, `unordered_list`, `upper`, `urlize`, `wordcount`.
* **20** raise on a MISSING argument (`{{ p|<f> }}` where `<f>` requires one)
  and djust renders — `add`, `center`, `cut`, `default`, `default_if_none`,
  `dictsort`, `dictsortreversed`, `divisibleby`, `get_digit`, `join`, `ljust`,
  `rjust`, `slice`, `stringformat`, `truncatechars`, `truncatechars_html`,
  `truncatewords`, `truncatewords_html`, `urlizetrunc`, `wordwrap`.

48 of 57 built-ins, on a dimension orthogonal to what any of them computes.

## Why it showed up now

It is the dominant residue of the #2389 sweep: of the 360 diverging cells that
remain after that fix, the large majority are this class, and it is what makes
"how many cells still diverge on the argument axis" hard to read. Isolating it
would make the argument-axis number mean what it looks like it means.

## Direction

Over-permissive: djust renders a template Django refuses to COMPILE. Not an
escaping question — the arity check runs before any value is touched — but it
does mean a typo in a template is silent in djust and loud in Django, which is
the failure shape this area keeps producing (#2325, #2334, #2377).

## Where to look

Django's per-filter arity comes from the registered callable's signature
(`inspect.getfullargspec`), so the set is DERIVABLE rather than a list to
transcribe: `len(args) - len(defaults)` per filter. djust's dispatch is
`apply_builtin_filter`'s `match filter_name`, whose arms read
`arg: Option<&str>` and silently ignore or default it.

The check belongs at PARSE time to match Django (a `TemplateSyntaxError` before
any render), which means the arity table has to reach the parser rather than
the filter dispatch. Worth deciding explicitly: a render-time refusal would be
less permissive than today and simpler, but would fire per-render rather than
once.

Refs #2389, #2328, #1079.

johnrtipton:
> **Stacked on #2407** (`fix-2399-2401-2403-filter-arms`). Both branches touch
> `filters.rs` and the differential; branching off #2407 keeps the merge clean.
> Merge #2407 first and this diff reduces to its own three commits.

**This is a behaviour change.** A template that renders today can raise after
this. The decision, and how it was made, is below.

## The defect

Django validates a filter's argument count in `FilterExpression.__init__`, at
COMPILE time, before any value is touched:

```python
def args_check(name, func, provided):
    plen = len(provided) + 1                      # the input is implied
    args, _, _, defaults, _, _, _ = getfullargspec(inspect.unwrap(func))
    if plen < (len(args) - len(defaults or [])) or plen > len(args):
        raise TemplateSyntaxError("%s requires %d arguments, %d provided" % …)
```

djust's dispatch read `arg: Option<&str>` and silently ignored or defaulted it,
so **48 of Django's 57 built-ins** rendered a template Django refuses:

```
{{ p|upper:"x" }}   django  <<TemplateSyntaxError: upper requires 1 arguments, 2 provided>>
                    djust   'ABC'
{{ p|default }}     django  <<TemplateSyntaxError: default requires 2 arguments, 1 provided>>
                    djust   'abc'
```

## Is this breaking, and should it be?

**Yes, and yes** — stated rather than assumed, since the same question was put
to #2328 (filter *arguments*) and answered "raise in dev and production alike".
The reasoning carries here for two reasons and one measurement:

1. The failure it replaces is **silent**. A wrong-arity call is a typo, and a
   typo that renders the wrong page is the exact class this area keeps
   producing (#2325, #2334, #2377). There is no "degraded but correct" reading
   of `{{ p|upper:"x" }}` to preserve.
2. The refusal is **contained**: `LiveViewConsumer.receive` already catches a
   render error and sends a safe error frame without dropping the socket, so
   the degradation decision is made once, at the transport, where it can be
   environment-aware. That is the same argument #2328's own comment makes.
3. **Measured blast radius**: the full djust suite (16,595 tests) and the demo
   project needed no template edits. Two Rust unit tests did — both pinned
   djust's own pre-fix answer, not a real template.

The CHANGELOG entry is under `### Changed`, marked BREAKING, and carries the
two greps a user needs before upgrading.

## The issue's count is wrong, and the correction is load-bearing

The issue says 28 built-ins raise a `TemplateSyntaxError` on an extra argument.
Measured against Django 5.2: **23** do.

The other five — `linebreaks`, `linebreaksbr`, `linenumbers`,
`unordered_list`, `urlize` — are `needs_autoescape=True`, and `args_check`
reads the RAW argspec, so `plen = 2 <= alen = 2` **compiles**. Their failure is
a render-time `TypeError: got multiple values for argument 'autoescape'`.

48 is right; "TemplateSyntaxError" is right for 43 of them. That matters
because a single bound would refuse those five **at parse time** — refusing a
template Django compiles, which is a wrong answer in the other direction. So
the table carries three numbers per filter and the fix has two sites.

`TestTheIssuesOwnCounts` recomputes all four numbers from the live registry;
`test_those_five_COMPILE_in_django_and_raise_at_the_call` is the empirical row.
The differential's `django_argument_filters` docstring also claimed
`args_check` excludes `autoescape` — it does not, and that claim is the whole
of this split. Corrected in passing.

## Two bounds, two sites, each with a test that goes red alone

| site | bound | the cells only it can see |
|---|---|---|
| `parser::parse_filter_specs` | compile (`args_check`) | `{% if False %}{{ p\|upper:"x" }}{% endif %}` — a node that never renders, which Django still refuses |
| `filters::apply_filter_full_safe` | call | the TAG operands: `{% if p\|f:"x" %}`, `{% for x in p\|f %}`, `{% with q=p\|f %}`, `{% firstof p\|f %}` — a raw string at parse time, split by `renderer::get_value_safe` at render time |

The render-time check runs **first**, before the argument is resolved. That is
Django's order too: `args_check` is inside `FilterExpression.__init__`, so
`{% if p|upper:missingvar %}` is an arity error in Django and never a
`VariableDoesNotExist`.

`TestBothSitesRefuse` names the test that reddens when each site is removed —
for **both** bounds, which took a second test to get right (see gate-off).
`test_the_two_sites_take_DIFFERENT_bounds` pins the `urlize` cell that proves
one bound would be wrong: `{% if False %}{{ p|urlize:"x" }}{% endif %}` must be
accepted by both engines.

## The table is transcribed; here is what keeps it honest

Nothing in the Rust engine can introspect a Python signature. The numbers were
generated mechanically from `inspect.getfullargspec` over the live registry,
and `TestTheTableIsDjangosOwnArity` re-derives all three bounds for all 57 and
compares — so a Django release that changes a signature fails a test rather
than drifting. The parser that reads the Rust const asserts its own row count
against the literal's, because a regex that matched nothing would report a
clean table (#2135).

Custom filters are **not** checked, and the reason is structural rather than a
choice: the Rust parser cannot introspect a Python signature, and a project's
filter may be registered after the templates using it are parsed. Django does
check them; that half is out of scope (#1079) and pinned in
`TestCustomFiltersAreNotArityChecked`, alongside the pre-existing fact that an
UNKNOWN filter still fails at render rather than at parse.

## The corpus could not construct one of these cells

I ran `--manifest` first, as asked. It reported **0 MISSING on ten axes** over
~345,000 cells — while this was the single largest class of divergence in the
corpus. `cells()` gives each argument-taking filter exactly one *valid*
argument out of `FILTER_ARGS` and gives the rest none; `arg_cells()` sweeps
spellings of a *single* argument over the filters that take one. Neither can
write `{{ p|upper:"x" }}` or `{{ p|default }}`. Fourth `input-shape
UNVERIFIED`-class miss in a row, and the largest.

**Gap closed in this PR**: an eleventh axis, `arity`, whose requirement is read
out of **Django's** `getfullargspec` — never out of `filter_arity.rs`, which
would agree by construction and could never report the table wrong. That is
the trap #2404's `loop-variable` axis was written to avoid one axis over, and
`test_the_requirement_is_read_out_of_DJANGO_not_out_of_djust` asserts it by
grepping for the readers (`_crate_source`, `read_text`, `_rust_const`, `open(`)
rather than trusting the prose.

Empirical canary, in the file that holds the other six:
`test_2400_no_cell_could_have_the_wrong_argument_COUNT` empties `arity_cells`
in a COPY of the script and asserts the manifest reports all 48 as MISSING,
with both halves (`upper:1`, `default:0`) named.

## Direction, measured

Two-build differential, baseline resolved once to `7ee4a0c6` (this branch's
merge-base — i.e. #2407's head, so the measurement isolates #2400), one
instrument across both builds:

```
cells        345172        builds f494dc7002200d8a -> 16b82e159807915e
arity        192 moved of 192      every other axis  0 moved
REGRESSIONS    0
newly AGREEING 0            <- see below; this is the expected number
panics       0 -> 0,   introduced 0
live-payload leaks 37 -> 37
  introduced 0   —   escaped 0  /  LIVE 0
```

`newly AGREEING: 0` is correct and not a null result. Both engines now REFUSE,
and a refusal is compared by its message — djust's crosses PyO3 as a
`RuntimeError` where Django raises `TemplateSyntaxError` or `TypeError`, so the
strings differ. The number that says what happened is the direction split,
computed from the two result files rather than asserted:

```
arity cells: 192   djust output moved: 192
   192   django-raised / djust-rendered  ->  django-raised / djust-raised
refusals carrying the payload in their message: 0
```

All 192 moved from "Django refuses, djust renders" to "both refuse". `192` is
`48 filters × 4 payload-carrying input shapes`.

## Gate-off, five mechanisms

Each mutation matched **exactly once**, source-change asserted, rebuilt between
iterations, restore verified byte-identical. pytest `N error` counted
separately from `N failed`; cargo's `error: test failed` distinguished from a
COMPILE error, which reports zero failures while proving nothing.

```
A1 the PARSE-time check never fires                         2 pytest | 0 cargo
A2 the RENDER-time check never fires                       12 pytest | 2 cargo
A3 the PARSE site uses the CALL bound (one bound, not two)   1 pytest | 1 cargo
A4 one table row is wrong (upper accepts an argument)       10 pytest | 2 cargo
A5 the min bound is not checked (a MISSING argument passes)  1 pytest | 1 cargo
(restored — the baseline)                                    0 pytest | 0 cargo
```

**A5 reported `0 pytest` on its first run, and that was a finding rather than a
harness fault.** Both sites check the same LOWER bound, so every
missing-argument cell that *renders* is caught by the render-time site and the
parse-time one is shadowed — the two-mechanisms-shadowing-each-other case
(CLAUDE.md #2129/#2135). An unrendered node is the only shape that separates
them, so `test_djust_refuses_an_unrendered_node_MISSING_its_argument` was added
and A5 now reddens it. The first `(restored)` row also read `2 cargo`, which
surfaced two stale Rust pins (below) — the harness asserting its own baseline
is what made that visible instead of being absorbed into every row.

## Stale pins

`filters.rs::test_stringformat_filter_default` and
`test_wordwrap_filter_default` pinned djust rendering a filter whose argument
is REQUIRED with no argument at all. Corrected to assert the refusal; the names
are kept, because the question each asks still has an answer and the answer
changed.

`test_differential_reachability_manifest_2345.py`'s axis-set pin gained
`arity`, and its `random`-is-collapsed test now excludes the arity axis — with
an added assertion that arity `random` cells are NOT collapsed, because a
refusal is deterministic and collapsing it would erase the movement the axis
exists to measure.

## Verification

* `pytest tests/ python/tests/ python/djust/tests/ -n auto` — 16,595 passed,
  337 skipped, 0 failed (post-merge with `origin/main`).
* `make test-rust` — 0 failures.
* `cargo clippy --workspace --all-targets -- -D warnings` — clean.
* `cargo fmt --check` — clean.

Fixes #2400


johnrtipton:
Follow-up filed while scoping this: **#2409** — a filter given TWO arguments (`{{ p|cut:"a":"b" }}`) renders here and is a `TemplateSyntaxError` in Django.

Deliberately NOT in this PR (#1079). It is a different mechanism one layer up: Django's *lexer* (`filter_raw_string`) allows at most one `:arg` for **every** filter and refuses the second before any filter is looked up, where this PR implements `args_check`, the per-filter count read from the callable's signature. `ARITY_COUNTS = (0, 1)` in the differential records the same boundary from the corpus side — a two-argument cell would be measuring the lexer rather than the arity, so the new `arity` axis deliberately does not build one.

The two are visible apart in one cell: `{{ p|upper:"a":"b" }}` now **agrees** (djust folds the two into one argument, and this PR's check then refuses that one argument for a filter that takes none), while every filter that legitimately takes one still diverges.

johnrtipton:
Third commit, after the first CI run went green: `call_time_arity_error` runs on **every** filter application, and `ARITY.iter().find()` was a 57-entry scan paid on the overwhelmingly common case where the count is legal — the case that does not need the table at all. Now a `binary_search_by` over the generator's already-sorted table.

Sortedness is a precondition that would not fail loudly: an unsorted table silently MISSES names, and a missed name is "no arity check for this filter", which is the bug the module exists to close. So `the_table_is_sorted_so_the_binary_search_is_valid` asserts both the order **and** that every row is findable *through* the search — a sorted table with a mis-indexed lookup would pass the order assertion alone.

Behaviour unchanged. The gate-off table in the description was **re-measured against this artifact** rather than carried over (identical: 2 / 12 / 1 / 10 / 1, baseline 0 pytest | 0 cargo), and the full suite re-ran green (16,595 passed).

PR #2410, opened 2026-08-29
fix(templates): a filter given TWO arguments is refused, and a quoted separator is not one (#2409)

### LINKED ISSUE #2409: A filter given TWO arguments renders here and is a TemplateSyntaxError in Django
Found while fixing #2400 and filed separately per CLAUDE.md #1079 — measured on

## The cells

Django's variable lexer allows a filter **at most one** argument
(`filter_raw_string` in `django/template/base.py`), so a second `:arg` is a
`TemplateSyntaxError` at COMPILE time — before `args_check` ever runs.
djust's `parse_filter_specs` splits on the FIRST colon and keeps everything
after it as one argument, so the second one is silently folded into the first.

Measured against live Django 5.2.16, on the #2400 branch (so the arity check is
already in place):

```
{{ p|cut:"a":"b" }}             p = "a b c"
  django  <<TemplateSyntaxError: Could not parse the remainder: ':"b"' from 'p|cut:"a":"b"'>>
  djust   'a b c'

{{ p|default:"a":"b" }}         p = "a b c"
  django  <<TemplateSyntaxError: Could not parse the remainder: …>>
  djust   'a b c'

{{ p|truncatewords:"1":"2" }}   p = "a b c"
  django  <<TemplateSyntaxError: Could not parse the remainder: …>>
  djust   'a b c'
```

`cut` receives the argument `"a":"b"` (quotes and all), finds no occurrence of
that literal, and returns the input unchanged. A wrong page, silently.

## Why #2400 does not close it

#2400 is `FilterExpression.args_check` — the per-filter argument COUNT, read
from the callable's signature. This is a different mechanism one layer up:
Django's *lexer* refuses the second `:arg` for **every** filter, including ones
that take one and ones that take none, and it does so before any filter is
looked up. The two are visible apart in one cell:

```
{{ p|upper:"a":"b" }}
  django  <<TemplateSyntaxError: upper requires 1 arguments, 2 provided>>
  djust   <<RuntimeError: Template error: upper requires 1 arguments, 2 provided>>
```

`upper` agrees, because djust folds the two arguments into one and #2400's
check then refuses that ONE argument for a filter that takes none. Every filter
that legitimately takes one argument still diverges, which is what makes this a
separate defect rather than a corner of #2400.

`ARITY_COUNTS = (0, 1)` in `scripts/filter-parity-differential.py` records the
same boundary from the corpus side: a two-argument cell would be measuring the
lexer rather than the arity, so the new `arity` axis deliberately does not
build one.

## Direction

Over-permissive: djust renders a template Django refuses to COMPILE, and the
render is silently wrong rather than obviously so. Not an escaping question —
the argument text goes through the normal filter path.

## Where to look

`crates/djust_templates/src/parser.rs`, `parse_filter_specs`:

```rust
if let Some(colon_pos) = filter_spec.find(':') {
    let filter_name = filter_spec[..colon_pos].trim().to_string();
    let arg = filter_spec[colon_pos + 1..].trim().to_string();
```

`find(':')` takes the first colon and the rest is one argument. Django's rule
is the `filter_raw_string` regex, which matches ONE optional `:arg` and then
requires the expression to be exhausted — hence "Could not parse the
remainder".

The colon inside a QUOTED argument is the case a naive "refuse a second colon"
check gets wrong. Measured: `{{ p|date:"H:i" }}` and `{{ p|cut:":" }}` both
PARSE on both engines today and must keep doing so, so the split has to respect
quoting rather than count colons.

`{{ p|cut:"a|b" }}` is the same shape one character over — `parse_token` splits
the expression on `|` without respecting quotes either — and already diverges,
though less quietly: djust sees a second filter named `b"` and raises
`Unknown filter`, where Django renders `''`. Worth fixing in the same pass,
since both are the tokenizer treating a quoted literal as ordinary text.

Refs #2400, #1079.

johnrtipton:
Closes #2409.

Django's variable lexer allows a filter **at most one** argument, and decides
that before any filter is looked up. `filter_raw_string`'s argument group is
optional and NON-repeating, and `FilterExpression.__init__` requires the regex
matches to TILE the token, so a second `:arg` is what is left over.

djust split on the FIRST colon and kept the rest as one argument, so
`{{ p|cut:"a":"b" }}` handed `cut` the argument `"a":"b"` — quotes and all —
found no such substring, and rendered the page unchanged. A wrong page,
silently, from a template Django refuses to compile.

This is **not** #2400's arity check. That reads each filter's own signature;
this applies to every filter, which is why `upper` agreed (two arguments folded
into one, then refused as one) while every filter that legitimately takes one
still diverged.

## The two halves are one scan

A "refuse a second colon" check would break `{{ p|date:"H:i" }}` and
`{{ p|cut:":" }}`, which both parse on both engines. And the same blindness sat
one character over on the pipe: `{{ p|cut:"a|b" }}` was split into two filters
and raised `Unknown filter` where Django renders — djust stricter for once,
wrong either way, because Django's `constant_string` admits any character
between the quotes, the two separators included.

Both are the new `crates/djust_templates/src/filter_lexer.rs`: a quote-aware
scan that splits on the pipes and the separator OUTSIDE a quoted string, and
validates the argument against Django's own three alternatives (a quoted
constant with backslash escapes, optionally `_( … )`-wrapped; `[\w.]+`;
`[-+.]?\d[\d.e]*`), refusing anything left over with Django's wording.

## One rule, two call sites

| site | shapes | was |
|---|---|---|
| `parser::parse_filter_specs` | `{{ … }}` | `var.split('|')` + `find(':')` |
| `renderer::get_value_safe` | `{% if %}`, `{% for %}`, `{% with %}` | `splitn(2,'|')` + `split('|')` + `find(':')` |

Both were independently quote-blind and independently accepted a second
argument, so a `{{ }}`-only fix would have left three tags over-permissive —
which the four-shape measurement shows it does. They call the module rather
than carrying a copy each (#1646); `TestBothSitesRefuse` goes red when either
call site is reverted.

## Measurement

Curated table, 92 cells (4 shapes × 23 argument spellings), both builds:

| | baseline `0b44d747` | branch |
|---|---|---|
| djust MORE PERMISSIVE than Django | **12** | **0** |
| djust stricter than Django | 3 | **0** |
| both render, different output | 1 | **0** |

Randomized differential, **12,356 distinct templates** assembled from five
filters and 37 argument atoms, each classified by which Django mechanism
refused it: **0 lexer-class permissive rows** for `{{ }}`, `{% for %}` and
`{% with %}`, and the single remaining stricter row is byte-identical on both
builds (an `_("x")` i18n argument djust renders literally — pre-existing).

Two-build differential, baseline pinned at `0b44d747` (resolved once — other
agents share this `.git`), engine-only so both runs cover the same 346,405
cells:

```
newly AGREEING: 319
no longer agreeing: 0
  REGRESSIONS : 0
live-payload leaks BEFORE : 37   AFTER : 37
  closed 0   INTRODUCED 0   (escaped 0 / LIVE 0)
panics before 0, after 0
```

## Corpus gap closed

`ARITY_COUNTS` read `(0, 1)` and said so on purpose: *"djust cannot SPELL two
arguments, so a two-argument cell would be measuring the lexer rather than the
arity."* Both halves were true and the conclusion was wrong — djust could not
spell two arguments because it FOLDED them.

* The bound is now **probed against Django's own compiler**
  (`django_lexer_max_arguments`), asserting the refusal's wording so a Django
  release that moves the boundary for another reason fails rather than
  silently lowering it. `django_refuses_arity` checks the lexer bound FIRST:
  reading only the argspec would stop requiring the cell the day Django
  shipped a two-argument filter. `arity` grows 48 → 105 required members.
* A new **`separator-in-constant`** axis requires a quoted argument carrying
  each of Django's two separators, read from
  `django.template.base.FILTER_SEPARATOR` / `FILTER_ARGUMENT_SEPARATOR`.
  Before this, no corpus cell put the SPLIT under test at all.

Both requirements are derived from **Django**, so neither is satisfiable by
the omission it detects.

## Gate-off (#1468)

Every mutation asserted found-exactly-once and source-changed, rebuilt between
iterations, restored byte-identically by hash, with pytest's `N error` counted
separately from `N failed` and cargo's `error:` separated from `test failed`.

| mutation | verdict |
|---|---|
| `parser`: `split_pipes` → quote-blind `str::split` | RED (4) |
| `renderer`: `split_pipes` → quote-blind `str::split` | RED (7) |
| lexer: remainder refusal → accept the rest as one argument | RED (26) |
| lexer: empty-argument refusal → accept a bare separator | RED (1) |
| lexer: `quoted_string_end` ignores the escape | RED (1, cargo) |
| corpus: `ARITY_COUNTS` back to `(0, 1)` | RED (2) |
| corpus: separator-carrying spellings removed | RED (2) |

7/7. No survivors.

## Scope left alone, and why (#1079)

All **120** remaining rows of the randomized sweep are `{% if %}`, and they are
ONE mechanism with four Django causes. djust refuses a tag operand at RENDER
time, left to right, so an earlier argument that fails to resolve is absorbed
by `{% if %}`'s `VariableDoesNotExist` arm — the arm that matches Django's
`ignore_failures` — and masks a LATER refusal Django raises at COMPILE:

```
{% if p|cut %}          django refuses   djust refuses     (agrees)
{% if p|date:.|cut %}   django refuses   djust renders ''  (masked)
```

It masks `_`-leading names (56), this fix's own remainder (28), #2400's arity
(33) and Django's own if-parser (3) alike. Filed separately rather than fixed
here: the cure is validating a tag operand's filter chain at PARSE time, which
is a different change from this one.

## Verification

* `pytest tests/ python/tests/ python/djust/tests/ -n auto` — 16,595 passed,
  337 skipped (all roots).
* `cargo test` — green; `cargo clippy --workspace --all-targets -- -D warnings`
  — clean.
* Two commits: impl+tests, then docs+CHANGELOG (#1173).


johnrtipton:
CI was red on four tests; all four were corpus **pins doing their job**, not a mechanism conflict. Fixed in `fa226701` (+ `origin/main` merged in, which brought #2412).

## What was red, and why

This PR widens the corpus — `ARITY_COUNTS` gains `2`, `ARG_SPELLINGS` gains the two separator-carrying spellings, and a `separator-in-constant` axis is declared. Four assertions state the corpus's shape as an exact set or an exact count, so all four moved:

| test | what it pinned | resolution |
|---|---|---|
| `test_the_axes_are_the_ones_this_file_documents` | the axis SET, not a floor (#1125) | `separator-in-constant` added with its reason |
| `test_2400_no_cell_could_have_the_wrong_argument_COUNT` | `len(missing) == 48`, a total across `ARITY_COUNTS` | re-scoped to a per-COUNT tally, so #2400's number is stated about the counts #2400 is about; grew an `upper:2` assertion for the new half |
| `test_the_corpus_now_builds_a_wrong_arity_cell_for_every_refused_pair` | the whole swept set vs the (0,1) requirement | re-scoped to the (0,1) slice, plus a new sibling asserting the `2` row is swept for **every** filter |
| `test_every_addition_is_accounted_for` | every `ARG_SPELLINGS` entry past upstream needs a written reason | both new spellings have one |

**Cause on my side**: I ran the full suite *before* the corpus edits and not after, so those four pins were never exercised against the widened corpus locally. Re-run across all roots on the merged branch: **16,672 passed, 337 skipped, 0 failed**.

## The shadowing question, measured rather than asserted

You were right that a two-argument refusal sitting next to #2400's arity check is the obvious suspect, and a shadowed pair looks identical from a green suite. Gating each mechanism off in turn:

| gate off | #2400's arity suite | #2409's lexer suite |
|---|---|---|
| the lexer remainder check | **GREEN** (0 failed) | **RED** (26 failed) |
| the parse-site arity check | **RED** (2 failed) | **GREEN** (0 failed) |

Each mechanism has tests that go red when only *it* is removed, and neither suite leans on the other. **Not shadowed** — independent and separately reachable, so there is nothing to delete.

Worth recording that the **first** run of that matrix said otherwise, and the mutation was at fault. I gated the arity check by `u8::from(arg.is_some())` → `u8::from(false)`, which does not disable it — it makes it report *zero* arguments for every filter and fire the `requires N arguments, 1 provided` arm, reddening 6 lexer tests for a reason that has nothing to do with either mechanism. A mutation that **changes** behaviour rather than **removing** it is not a gate-off. That is a fourth failure mode alongside the three in canon (mutation not found / mutation breaks the build / mutation is a semantic no-op), and it fails in the more dangerous direction: it reports a coupling that does not exist. The clean gate is `for (name, arg) in specs.iter().take(0)`.

## Two other harness preconditions I had to fix while here

- The gate-off harness skipped the rebuild for corpus-only mutations, so those rows ran against whatever `.so` the *previous* mutation left behind and reported failure counts belonging to another mutation (9 and 7, where the true numbers are 3 and 1). It now rebuilds unconditionally, before the run and after the restore.
- `cargo test --workspace` SIGABRTs on `djust_core` in my local setup, because it links `djust_live`'s `pyo3/extension-module` feature into a test binary. That is what `make test-rust`'s two-phase split exists to avoid; `cargo test --workspace --exclude djust_live` is green (62 result lines, 0 errors).

CHANGELOG checked for union-merge duplication after merging `main`: 101 `[Unreleased]` bullets, every one carrying a bolded headline (so the parser is known to have seen them all), 0 duplicate headlines, 0 byte-identical duplicates.

Issue #2405, opened 2026-08-29
forloop's dict key order differs from Django's in json_script


johnrtipton:
Found by measuring the interaction between PR #2404 (`forloop` members, merged) and PR #2398 (quoted filter arguments are `SafeData`, merged) rather than assuming they were orthogonal — neither PR could see it alone.

## The divergence

```
{% for x in p %}{{ forloop|json_script:"d" }}{% endfor %}    p = [1]

djust   {"counter": 1, "counter0": 0, "first": true, "last": true, ...}
django  {"parentloop": {}, "counter0": 0, "counter": 1, ...}
```

Same keys, same values, **different insertion order**. `json.dumps` preserves dict order, so the serialized bytes differ.

## Why it is small, and why it is filed anyway

Cosmetic for any consumer that parses the JSON — key order is not semantically meaningful in an object. It matters only to something comparing bytes or reading positionally, which is unusual but not unheard of (a snapshot test on rendered output, a checksum, a diff in CI).

It is filed because:

- it is in code merged the same day, and **no gate covers it** — the differential's `json_script` cells exist but nothing compares key ordering of the engine's own `forloop` dict;
- it was found only because one agent measured a cross-PR interaction it had been told to assume was orthogonal. That is not a repeatable way to find things.

## Where it is

`Node::For`'s dict construction in `crates/djust_templates/src/renderer.rs`, added by #2404. Django's `ForNode.render` writes `parentloop` first (captured from the incoming context before the push), then the six counters. djust writes the counters first.

## Fix shape

Match Django's insertion order in the dict construction. One-line reordering; the value is in the pin, not the fix — a test that asserts the serialized key order against Django's, so this cannot drift again.

## Not a security issue

Both engines escape identically here; the same agent measured 18 cells across all seven `forloop` members × three granted filters and found **0 more-permissive-than-Django** on either build. This is ordering only.

PR #2415, opened 2026-08-29
security(templates): a custom tag handler's return is ESCAPED unless it is already HTML (#2379)

### LINKED ISSUE #2379: A custom tag handler's return value is emitted RAW where Django escapes it

johnrtipton:
Closes #2379.

Django's `SimpleNode.render` runs `conditional_escape` over a `simple_tag`'s
return unless it carries `__html__`. `renderer.rs`'s `Node::CustomTag` arm
inserted it **verbatim**, so a handler as ordinary as

```python
@register.simple_tag
def greet(name):
    return f"Hello {name}"
```

emitted `Hello ` live where Django renders
`Hello &lt;img …&gt;`. The fail-OPEN half of the asymmetry #2290 found on the
way IN, reaching every `register_tag_handler` / `register_block_tag_handler`
user — djust's own handlers and any project's.

**The two-build differential closes 15 live-payload cells and introduces none.**

## The handler audit — the work, and it corrected the issue in both directions

Escaping a return that legitimately IS markup is a rendering regression rather
than a fix. So every handler was enumerated **mechanically** — by intercepting
the three `register_*_tag_handler` functions and triggering every registration
path — and then **called**, never read: a `mark_safe` on line 3 of a body with
four returns answers the question for one of them.

The issue names ten modules and reads as ~20 handlers, and says *"almost none
of them `mark_safe`"*. Measured:

| bucket | handlers |
|---|---|
| already carries `__html__` — escaping is a **no-op** | **195** |
| returns `""` — no-op | 13 |
| returns plain text, no markup — escaping is **correct** (`static`, `templatetag`, `theme_mode`, `theme_preset`, `theme_resolved_mode`) | 5 |
| **returns MARKUP as a plain `str` — would render as escaped text** | **6** |
| assign tag, returns a dict (a different path) — `regroup` | 1 |
| needs a real view to reach a return — `live_render` | 1 |
| **total** | **221 across 13 modules** |

So the direction is the opposite of the issue's premise: the codebase was
already almost entirely `mark_safe`-clean, and the audit is 6 handlers rather
than ~20 modules' worth.

### The six, and how each is fixed

| handler | what it returned | fix |
|---|---|---|
| `call` / `component` (one class) | the missing-name error comment, and the whole component render | `safe_html` at the ONE exit; body moved to `_render_component` |
| `dj_suspense` | all four branches (passthrough, fallback, error div) | `safe_html` at the ONE exit; body moved to `_render_state` |
| `djust_markdown` | the `nh3`-sanitised HTML it produces | `safe_html` at the return |
| `slot` | the `<!--DJUST_SLOT_V1:…-->` sentinel `CallTagHandler` parses back out | `safe_html` — escaping it would break slot collection outright |
| `toast_container` | the empty-container early return | `_safe(…)`, the module's own convention |

Marked at one exit rather than at N returns (#1104). `toast_container` is the
sharpest of the six: its early return was the **single** return out of ~190 in
`rust_handlers.py` that missed the module's own `_safe()` convention, and
nothing could see it until the bridge started escaping.

`safe_html` is `rust_handlers._safe`'s body moved to `djust/_html.py` so the
three tag packages can import it rather than each growing a fourth copy of the
`cast(str, mark_safe(...))` mypy workaround (#1646). `_safe` stays where it is:
re-pointing ~190 call sites is a rename with no behavioural content.

## The half a return-only fix would have got wrong

Django's `simple_block_tag` hands the handler `nodelist.render(context)` —
already-rendered, already-escaped markup, and therefore `SafeData`. djust
passed it across PyO3 as a bare `str`, so a handler returning its content
unchanged lost the marker and the escape applied it a **second** time:

```
{% cb_ident %}{{ p }}{% endcb_ident %}   p = <img src=x onerror=alert(1)>
  django  &lt;img src=x onerror=alert(1)&gt;
  djust   &amp;lt;img src=x onerror=alert(1)&amp;gt;     <- before the block fix
```

Found by measuring the `cb-ident` cell, not by inspection. The bridge now marks
the block body safe on the way in — the block-path twin of #2290's marker loss
— and fails **soft** if `django.utils.safestring` is not importable, so a
pure-Rust embedding keeps rendering.

## One predicate, three callers

`py_value_is_safe_string` is #2290's own test, **extracted** rather than copied
(#1646). Its security half — requiring `str` subclass-ness and not just
`__html__` — is now stated once, and it is load-bearing on a *different* path
from the one this PR adds:

* **filter path**: `extract::<Value>()` stringifies an arbitrary object, so the
  predicate is the only thing between a non-`str` impostor advertising
  `__html__` and an unescaped payload. Measured: the impostor renders escaped.
* **tag path**: `extract::<String>()` refuses a non-`str` outright, so the
  impostor never reaches the marker test.

Both are asserted, because "the impostor cannot leak" has two different proofs
and only one of them is the predicate's. That distinction surfaced as a
gate-off survivor and is recorded rather than tested around.

## Two-build differential

Baseline pinned at `fca704cf`, engine-only so both runs cover the same 345,286
cells:

```
newly AGREEING: 49
no longer agreeing: 1
  coincidental (mechanically classified): 1
  REGRESSIONS : 0
live-payload leaks BEFORE : 37   AFTER : 22
  closed 15   INTRODUCED 0   (escaped 0 / LIVE 0)
panics before 0, after 0
```

The single non-agreeing cell is `@ctag ct-ident <s-marked>`: a `mark_safe`d
context value. It agreed **by coincidence** — the marker was lost on the way IN
(#2290's argument side) and the raw return cancelled it on the way OUT — so
once the return is escaped, the input-side loss shows. `unmasked()` grew a
`@ctag` arm to tell that apart from a real regression, and it needs **both**
conditions to fire:

1. djust's new output IS Django's output escaped once more — the exact
   signature of a discarded marker;
2. the `ct-cond` probe (which reads the marker *inside* the handler) over the
   same input diverges on **both** builds.

That makes it a statement about the mechanism rather than an exemption keyed on
the input: a genuinely new `@ctag` regression fails condition 1 and is still
reported. `TestTheDifferentialCanTELLThatUnmaskingApart` exercises all three
branches.

## Gate-off (#1468)

| mutation | verdict |
|---|---|
| the TAG return goes in RAW again (the original XSS) | RED (10) |
| the BLOCK return goes in RAW again | RED (1) |
| the `__html__` test always says NO (a blanket escape) | RED (5) |
| the `__html__` test always says YES (the fix disabled) | RED (8) |
| the block body reaches Python as a BARE `str` (double-escape) | RED (4) |
| the shared #2290 predicate drops its `str`-subclass half | RED (1) |
| `djust_markdown` loses its `mark_safe` | RED (1) |
| `dj_suspense` loses its `mark_safe` | RED (2) |
| `call`/`component` loses its `mark_safe` | RED (2) |
| `slot` loses its `mark_safe` | RED (2) |
| `toast_container`'s EMPTY branch loses its `_safe` | RED (2) |
| the `unmasked()` `@ctag` arm removed | RED (1) |

12/12. Two rows were **INVALID** on the first run (mutation text matched 0×,
because `cargo fmt` and `ruff format` had reformatted the lines) and one was a
**survivor** — the harness reported each honestly rather than as a pass.

## Left open, and pinned rather than left silent (#1079)

Both are djust being **stricter** — over-escaping, never a leak — and both were
*unmasked* by this change rather than caused by it:

* a `mark_safe`d context value reaches a handler as a bare `str` (#2290's
  argument side);
* a quoted literal argument keeps its quotes, where Django's
  `Variable('"<b>"')` runs `mark_safe(unescape_string_literal(…))`.

Three pins in `TestKnownDivergencesOnTheCustomTagPath` go red as designed and
are inverted in place; the class docstring said they would.

## Verification

* `pytest tests/ python/tests/ python/djust/tests/ -n auto` — **16,639 passed**,
  337 skipped, 0 failed (all roots).
* `cargo test --workspace --exclude djust_live` — green;
  `cargo clippy --workspace --all-targets -- -D warnings` — clean.
* CHANGELOG checked for union-merge duplication: 101 `[Unreleased]` bullets,
  every one carrying a bolded headline, 0 duplicates.
* Two commits: impl+tests, then docs+CHANGELOG (#1173).


johnrtipton:
Cross-check on the leak count, since the issue states one and the tool states another — both are right and they measure slightly different things.

| measurement | before | after | closed | introduced |
|---|---|---|---|---|
| the tool's own `--compare` (its per-input `LIVE_FRAGMENTS` set) | 37 | 22 | **15** | **0** |
| an independent classifier: an unescaped tag opener AND a fragment Django's output lacks, over `<img` / `onerror=` / `<script` / `</script` | 291 | 279 | **12**, all `@ctag` | **0** |
| the issue's own reproduce block (`@ctag ct-ident` + `<img`) | 11 | 0 | **11** | — |

The middle row is **exactly the issue's headline count of 12**, and every one of the twelve is in the `@ctag` axis:

```
@ctag cb-plain  | s-marked          @ctag ct-ident | d-numeric-key
@ctag ct-ident  | s-img             @ctag ct-ident | d-typed-hostile
@ctag ct-ident  | s-script          @ctag ct-ident | d-hostile-key
@ctag ct-ident  | s-lines           @ctag ct-ident | d-marked-value
@ctag ct-ident  | l-scalars         @ctag ct-ident | d-model
@ctag ct-ident  | l-marked-img      @ctag ct-ident | l-mixed
```

The three the tool counts beyond that are cells where the payload fragment appears but the surrounding markup was already escaped — the split `--compare` prints as *"fragment inside fully-escaped text (not live)"* vs *"LIVE (an unescaped tag opener)"*. Both readings agree on the two things that matter: every one moved in the closing direction, and **nothing became more permissive on any input** (`introduced: 0`, escaped 0 / live 0).

The `cb-plain | s-marked` row is worth naming on its own: it is the BLOCK registry, and it is the cell that proves the fix reaches both registries rather than only `register_tag_handler`.

johnrtipton:
Fixed and merged forward. Diagnosis matches yours exactly; I took option 1.

## What it was

`djust.template_tags._registered_handlers` is a process-global accumulator with no teardown. `tests/unit/test_tag_registry.py` registers a handler named `custom` through the `@register` decorator, and it stays there for the rest of the worker. My precondition asserted over the raw accumulator, which made it true only in a pristine process.

Reproduced deterministically before changing anything — the two files together, in that order:

```
pytest tests/unit/test_tag_registry.py python/tests/test_custom_tag_return_escape_2379.py
  AssertionError: ['custom']
```

## The fix (option 1)

`declared` is scoped to handlers whose class is DEFINED IN djust (`type(handler).__module__.startswith("djust.")`). That is what the precondition was always about — *every handler djust itself ships is accounted for*, not *the registry contains nothing else*. I agree on avoiding the save/restore: a global mutation in a parallel suite is a later bite.

## The canary you asked for

`test_a_FOREIGN_handler_does_not_break_the_enumeration` registers a handler under a foreign name, from a class defined in the test module, and asserts the precondition still holds. It restores in a `finally` — a canary that leaves its handler behind becomes the pollution it tests for — and asserts the restore took.

Gate-off, both directions:

| mutation | verdict |
|---|---|
| scoping removed INSIDE the canary | RED (1) |
| scoping removed in the PRECONDITION | RED (1) |

The first is the load-bearing one: **with the pre-fix assertion restored, the canary reproduces CI's exact failure in a pristine single-file run**, no second test file needed. That is what makes it a canary rather than a restatement of the bug — the original failure required xdist to co-schedule two files; this requires nothing.

I also added `test_that_scoping_still_covers_the_real_handlers`, because the filter could have satisfied the precondition by narrowing `declared` to nothing: `djust_markdown` / `static` / `url` / `regroup` must survive it.

## Verification

- Merged `origin/main` (not rebased). Clean; CHANGELOG auto-merged with **102 `[Unreleased]` bullets, every one carrying a bolded headline** (the parse-count assertion — a bullet the parser could not see would fail before the duplicate check runs), **0 duplicate headlines, 0 byte-identical duplicates**. The #2409, #2379 and #2405 entries each appear exactly once.
- Pollution-class fix, so the **3-clean-run gate (#182/#1174)**: 16,700 passed on each of three consecutive full-suite runs across all roots, two of them `-n auto` and one `-n 4` to vary the distribution — since the whole failure was distribution-dependent, varying it is the part that matters.
- `cargo test --workspace --exclude djust_live` green; `cargo clippy --workspace --all-targets -- -D warnings` clean.

## For the record

Your framing of the direction is the durable part: a green local full-suite run cannot rule out order-dependent global state, because the scheduling is what decides it. The remedy is not more runs — it is asserting the property directly, which is what the canary now does. Worth noting the same shape had already bitten me once in this PR (the `_registered_handlers` reload the enumeration needs, which an `import_module` silently no-ops); both are the same lesson about process-global state that no teardown resets.

Issue #2425, opened 2026-08-29
json_script: a bool/None dict KEY is spelled True/None where json.dumps writes true/null, and a tuple/bytes/object key is EMITTED where json.dumps refuses


johnrtipton:
Found by the un-masking in #2413 and filed separately per CLAUDE.md #1079 — measured on
that branch, PRE-EXISTING, and not touched there.

`json.dumps` and `value_to_json` disagree about what a non-`str` dict KEY becomes.
Python's encoder coerces exactly five key types with JSON spellings and REFUSES
everything else; djust writes the Python `str()` for all of them and refuses nothing.

## Measured, `{{ p|json_script:"d" }}`

| key | django | djust |
|---|---|---|
| `{"a": 1}` | `{"a": 1}` | agrees |
| `{0: "a"}` | `{"0": "a"}` | agrees |
| `{1.5: "d"}` | `{"1.5": "d"}` | agrees |
| `{True: "b"}` | `{"true": "b"}` | **`{"True": "b"}`** |
| `{None: "c"}` | `{"null": "c"}` | **`{"None": "c"}`** |
| `{(1, "t"): "e"}` | **`TypeError: keys must be str, int, float, bool or None, not tuple`** | `{"(1, 't')": "e"}` |
| `{frozenset({1}): "f"}` | `TypeError: … not frozenset` | `{"frozenset({1})": "f"}` |
| `{b"k": "h"}` | `TypeError: … not bytes` | `{"b'k'": "h"}` |
| `{obj: "g"}` | `TypeError: … not Obj` | `{"OBJ": "g"}` |
| `{"a": {(1,): "x"}}` | `TypeError: … not tuple` | `{"a": {"(1,)": "x"}}` |

Two mechanisms in one place:

1. **Spelling.** `json.dumps` writes a `bool` key as `true`/`false` and a `None` key as
   `null` — the JSON spellings, not Python's. djust routes the key through
   `Value::to_display_string()`, which is `True` / `None`. The `int` and `float` arms
   agree by coincidence, because `str(0)` and `str(1.5)` are already the JSON forms.
2. **Refusal.** For every other key type `json.dumps` raises `TypeError` (there is no
   `skipkeys=True` in `json_script`). djust stringifies whatever `to_display_string`
   gives, so it is MORE PERMISSIVE than Django — it emits a document where Django 500s.

## Where to look

`crates/djust_templates/src/filters.rs`, `value_to_json`'s `Object` arm — the key path
added by #2241:

```rust
[...]

Issue #2401, opened 2026-08-29
{{ p|yesno:"only" }} — a one-part argument returns the VALUE in Django, a hard-coded suffix here


johnrtipton:
Found while draining #2389 and filed separately per CLAUDE.md #1079 — measured
on `main` at `3b019bbe`, PRE-EXISTING, and not touched there.

## The cell

Django's `yesno` returns the **value unchanged** when its argument has fewer
than two comma-parts; djust falls back to a hard-coded `yes`/`no`/`maybe`.

```python
def yesno(value, arg=None):
    if arg is None:
        arg = gettext("yes,no,maybe")
    bits = arg.split(",")
    if len(bits) < 2:
        return value          # Invalid arg.
    ...
```

Every value shape diverges, measured against live Django 5.2.16:

```
{{ p|yesno:"only" }}
  p = True    django 'True'   djust 'only'
  p = False   django 'False'  djust 'no'
  p = None    django 'None'   djust 'maybe'
  p = ""      django ''       djust 'maybe'
  p = "abc"   django 'abc'    djust 'only'
  p = 0       django '0'      djust 'maybe'
  p = 1       django '1'      djust 'only'
```

Note `p = False` → `no` and `p = 0` → `maybe`: djust is not merely using the
default triple, it is running its normal three-way branch and then picking from
a mix of the argument and the defaults. So the fix is one early return, not a
different default.

## Not the #2389 class

`yesno` came up because #2389's candidate list names it as a filter that
returns its argument, and it does not — `str.split(",")` on a `SafeString`
yields plain `str`s, so the safety half is already correct on both engines and
was correct before. This is the ARITY of the argument, a separate defect in the
same arm.

## Direction

Neither more nor less permissive in the escaping sense — both engines escape.
It is a wrong answer: a template author who writes `{% ... |yesno:"active" %}`
meaning "print `active` when true" gets `active` from djust for every truthy
value and `no`/`maybe` otherwise, where Django prints the value itself.

## Where to look

`crates/djust_templates/src/filters.rs`, the `"yesno"` arm. Django's rule in
full: `arg.split(",")`; `< 2` parts returns the value; exactly 2 parts uses
`bits[1]` for the `None` case as well as the false case; 3 parts uses `bits[2]`
for `None`. The `None`-with-two-parts row is the one a curated test is most
likely to miss.

Refs #2389, #1079.

Issue #2366, opened 2026-08-28
int(arg) is a TypeError for a non-str non-number argument, and the ReturnInput filters swallow it


johnrtipton:
Surfaced by the argument-axis corpus #2345 adds, on the two spellings it grew
for #2344 (a resolvable `datetime`). Measured, Django 5.2.16:

```
{{ p|truncatechars:q }}   p="abcdefghij"          q=datetime(2020,1,1,15,30)
  django: TypeError: int() argument must be a string, a bytes-like object or
          a real number, not 'datetime.datetime'
  djust : 'abcdefghij'
```

Same for `truncatewords`, `get_digit` and `floatformat` — the four filters
whose Django source catches `ValueError` ONLY:

```python
def truncatechars(value, arg):
    try:    length = int(arg)
    except ValueError:  return value      # a TypeError escapes
```

`int()` raises **TypeError**, not ValueError, for anything that is neither a
string nor a number: a `datetime`, a `date`, a list, a dict, a model instance.
So Django raises and these four return their input.

## Why djust cannot currently see it

The argument reaches the dispatch table as a STRING — a quoted literal
verbatim, or a resolved context value through `Value`'s `Display`. A `datetime`
arrives as `"2020-01-01 15:30:00"`, which `int()` on a *string* rejects with a
**ValueError**, so the `ReturnInput` policy fires and the value comes back
unchanged. Django, holding the real object, takes the other branch.

The control confirms the mechanism rather than the symptom — passing the
datetime's own `str()` makes both engines agree:

```
{{ p|truncatechars:q }}   q="2020-01-01 15:30:00"
  django: 'abcdefghij'
  djust : 'abcdefghij'      <- agree
```

`center` and the other `Raise`-policy filters are unaffected in the raise BIT
(both raise, different exception types), and `slice`/`join` never call `int()`
on the argument.

## Relationship to #2328 and #2344

#2328 already fixed exactly this rule for ONE value: `arg_is_none_literal` +
`none_argument_error` exist because `int(None)` is a TypeError that no filter's
`except ValueError` catches. This is the same rule for every other
TypeError-inducing type, and it needs the same information #2344's falsiness
residue needs — the argument's Python TYPE, which the string-valued dispatch
boundary has already discarded.

So the fix is probably not another special case: either the dispatch table
learns the argument's original type (a `Value` beside the `&str`), or this is
declared a bounded wire-format residue and pinned, as #2344's is in
`TestTheFalsinessResidueIsNamed`.

[...]

Issue #2249, opened 2026-08-27
tech-debt: the #2241 value_to_json structural pin greps RAW Rust source (#2238 class, non-Python)


johnrtipton:
Follow-up from #2238 (PR #2245). Filed at the coordinator's request so the gap is stated rather than assumed covered.

**Corrected after empirically probing both pins** — the first version of this issue described pin A's failure direction wrongly (see "Correction" below).

PR #2247 (#2241) added two structural pins in `python/tests/test_json_script_escaping_2241.py` that grep RAW Rust source. Both slice `value_to_json`'s body out of `crates/djust_templates/src/filters.rs` with `_value_to_json_source()` and match against the text. Both are prose-blind, in **different** directions:

## Pin A — `test_value_to_json_has_no_escape_chain_of_its_own` (line 279)

`assert ".replace(" not in body` — a NEGATIVE assertion, so prose turns it **RED**, not green:

```
baseline                        .replace( present=False  -> GREEN
+ explanatory // comment        .replace( present=True   -> RED on PROSE
+ commented-out chain           .replace( present=True   -> RED on DEAD CODE
```

A `//` comment inside the body saying *"never add a `.replace(` chain here"* — exactly the comment someone will write next to this rule — fails the build. That is the #2237 shape: the prose gets contorted to satisfy the checker, and #2238 exists to stop that.

## Pin B — `test_value_to_json_escapes_every_string_through_the_one_helper` (line 270)

`assert body.count("json_string_body(") == 3` — a POSITIVE count, so it has the classic **false negative**:

```
baseline                              count=3  -> GREEN
1 real call site deleted              count=2  -> RED (correct)
...with its text left in a // comment count=3  -> GREEN despite the deletion
```

Deleting a real escape site and leaving `// was: json_string_body(s)` behind keeps the pin green. This is the direction nothing catches and the more valuable half — it is the #1817 bug verbatim, in Rust.

## Correction

The first version of this issue claimed pin A had a false negative where "a commented-out `.replace(` chain still matches, so the drift can land while it stays green". That is wrong: a match makes a `not in` assertion **fail**. Both of pin A's failures are false alarms; the genuine false negative lives in pin B. I pattern-matched the #1817 pin's positive-count shape onto a negative assertion without running it. Running it took a minute.

## The #2238 helper does NOT cover this

[...]

PR #2426, opened 2026-08-29
fix(templates): json_script matches json.dumps' ensure_ascii, and omits a falsy element_id (#2413)

### LINKED ISSUE #2413: json_script: non-ASCII is emitted raw where json.dumps escapes it, and a missing element_id still emits id="data"
Found while fixing #2405 and filed separately per CLAUDE.md #1079 — measured on
that branch, PRE-EXISTING, and not touched there. Two divergences in one
filter, both byte-level, both a different mechanism from the key ordering
#2405 closes.

## 1. `ensure_ascii`

`django.utils.html.json_script` calls `json.dumps(value, cls=encoder)`, and
`json.dumps` defaults to **`ensure_ascii=True`** — every non-ASCII character
is emitted as a `\uXXXX` escape (a surrogate PAIR above the BMP). djust's
`json_string_body` emits raw UTF-8:

```
{{ p|json_script:"d" }}     p = {"k": "héllo→"}

  django  <script id="d" type="application/json">{"k": "héllo→"}</script>
  djust   <script id="d" type="application/json">{"k": "héllo→"}</script>
```

It applies to KEYS as well as values, and to every nesting depth.

Measured on the #2405 branch: a randomized sweep of 1,600 assembled nested
values put **548** of them through a non-ASCII character, and all 548 are
byte-divergent for this reason alone — the ASCII-only 1,052 agree exactly.

The corpus already carries it: `s-unicode` (`"héllo→"`) and `s-lines` diverge
on every `json_script` cell they appear in, on both builds. So this needs no
new corpus shape, only the fix and a pin.

## 2. A missing `element_id` still emits one

```
{{ p|json_script }}         p = {"k": 1}

  django  <script type="application/json">{"k": 1}</script>
  djust   <script id="data" type="application/json">{"k": 1}</script>
```

`json_script(value, element_id=None)` builds
`format_html('<script id="{}" type="application/json">{}</script>', …)` only
when `element_id` is not None, and `'<script type="application/json">{}</script>'`
otherwise. djust hardcodes `id="data"`.

This one has a second effect worth naming: it MASKS every argument-less
`json_script` cell in `scripts/filter-parity-differential.py`. All 10,166 of
them diverge on the `id` attribute, so any divergence in the JSON BODY sits
underneath and cannot be attributed — which is why the #2405 corpus shape had
to pass an explicit `id`, and why `test_a_missing_element_id_still_emits_one`
in `python/tests/test_json_script_key_order_2405.py` pins it.

An `id` djust invents is also a DOM collision: two argument-less
`{{ …|json_script }}` calls on one page both emit `id="data"`.

## Direction

Neither is more permissive than Django — both are wrong OUTPUT, not a wrong
refusal, and neither changes what is escaped as HTML (the `<`, `>`, `&`
escapes `_json_script_escapes` applies are correct on both engines). The
`ensure_ascii` half is a behaviour change for anyone byte-comparing a rendered
payload; the `id` half is a behaviour change for any page relying on the
invented `id="data"`.

## Where to look

* `crates/djust_templates/src/filters.rs`, `json_string_body` — the shared
  escaper `value_to_json` routes every string arm through (#2241). ASCII
  escaping belongs here, so keys and values and every depth get it from one
  place rather than at the call sites (#1646). Note the surrogate-pair rule:
  `json.dumps("😀")` is `"😀"`, not `"\U0001f600"`.
* `crates/djust_templates/src/filters.rs`, the `"json_script"` arm (~line
  1767) — the `id` attribute. Django omits the whole attribute; it does not
  emit an empty one.

## Pins that go red when this is fixed

`python/tests/test_json_script_key_order_2405.py::TestKnownDivergencesOnTheSamePath`
names both, deliberately, so this cannot land silently — those two tests are
the signal to rewrite them as parity assertions.

Refs #2405, #2241, #1079.

johnrtipton:
Closes #2413.

Two byte-level divergences in one filter, both a different mechanism from the key ordering #2405 closed, and both named in that PR's `TestKnownDivergencesOnTheSamePath` so they would go red the day they were fixed — which is how this landed.

## Reproducers, before the fix

Bytes, both engines, from `django.utils.html.json_script` and `_rust.render_template` directly:

```
DIFF with-id  value-nonascii   {"k": "héllo→"}
       django <script id="d" type="application/json">{"k": "héllo→"}</script>
       djust  <script id="d" type="application/json">{"k": "héllo→"}</script>
DIFF with-id  key-nonascii     {"héy→": "v"}
       django <script id="d" type="application/json">{"héy→": "v"}</script>
       djust  <script id="d" type="application/json">{"héy→": "v"}</script>
DIFF with-id  astral-value     {"k": "😀"}
       django <script id="d" type="application/json">{"k": "😀"}</script>
       djust  <script id="d" type="application/json">{"k": "😀"}</script>
DIFF with-id  DEL              {"k": "\x7f"}
       django <script id="d" type="application/json">{"k": ""}</script>
       djust  <script id="d" type="application/json">{"k": "<raw DEL>"}</script>
DIFF no-id    ascii            {"k": 1}
       django <script type="application/json">{"k": 1}</script>
       djust  <script id="data" type="application/json">{"k": 1}</script>
```

33 of 36 rows diverged. After: **0 of 36.**

## 1. `ensure_ascii`

`django.utils.html.json_script` calls `json.dumps(value, cls=encoder or DjangoJSONEncoder)` and passes no `ensure_ascii`, so it takes the default of **`True`**. That premise was falsification-tested rather than read: `DjangoJSONEncoder`'s source overrides only `default()`, and `json_script("é")` was run to confirm.

The escaping rule was **derived by running `json.dumps` over every codepoint** rather than read off the C encoder. What comes back raw is exactly `U+0020..U+0021`, `U+0023..U+005B`, `U+005D..U+007E` — printable ASCII minus `"` and `\` — so everything outside `0x20..=0x7E` is escaped, with the seven short forms taking precedence. Three details that reading would have got wrong:

* an **astral** codepoint is a UTF-16 surrogate **PAIR** — `json.dumps("😀")` is `"😀"`, not a six-hex escape and not the raw character;
* the hex is **lowercase** (`é`) where `_json_script_escapes` writes **uppercase** (`<`) — both spellings are Django's, from the two steps it composes, and a single-case implementation matches one and breaks the other;
* **`0x7F` (DEL) escapes.** The Rust helper's own doc argued against that by citing `json.dumps(ensure_ascii=False)`, which is true about a call Django does not make. DEL is the one character where the two flags disagree, which makes it the sharpest single witness for which path this is.

Escaping lives in `json_string_body` — the ONE helper every quoted string in `value_to_json` goes through (#2241) — so keys, values and every nesting depth get it from one place.

**A deletion comes with it.** `json_escape_for_script` carried `U+2028` / `U+2029` arms beyond Django's three-character `_json_script_escapes`. They were the right compensation while the engine emitted raw UTF-8; nothing non-ASCII can now reach that stage (pinned by `json_string_body_output_is_pure_ascii_2413`, asserted rather than reasoned about because the deletion depends on it), so they were a dead second mechanism for a job the first already does — which can only shadow it (#2233). Their absence also makes the map Django's exact three characters.

## 2. A falsy `element_id` still emitted one

**The issue's premise was narrower than the defect, and running Django is what showed it.** The source is `if element_id:` — a TRUTHINESS test on the resolved Python OBJECT, not `is not None`:

| `{{ p\|json_script:e }}`, `e =` | django | djust (before) |
|---|---|---|
| *(no argument)* | `<script type="application/json">` | `<script id="data" …>` |
| `None` | no `id` | `id="None"` |
| `""` | no `id` | `id=""` |
| `0` / `0.0` | no `id` | `id="0"` / `id="0.0"` |
| `False` | no `id` | `id="False"` |
| `[]` / `{}` | no `id` | `id="[]"` / `id="{}"` |
| literal `0`, `None`, `False`, `""` | no `id` | the spelling |

`str(0)` is `"0"`, so the dispatch table's `&str` cannot answer this. The resolved value's truthiness is threaded as a third `ArgType` bit — the existing mechanism for exactly this class (`int_is_type_error` #2366, `is_none` #2401) — computed **once** at the resolution site for BOTH argument channels, because `FilterExpression.resolve` produces a Python object either way: a bare identifier through `Variable(token).resolve(context)`, a constant through the parse-time literal. The invented `"data"` default is **deleted**, not left unreachable, and "no argument at all" stays the call site's own `Option` to read, so there is one mechanism per question rather than two for either.

## Grep for the sink

`json_string_body` has 4 call sites, all in `value_to_json` (structurally pinned since #2249). `value_to_json` has exactly 1 production caller, the `json_script` arm. `application/json` appears in exactly one production line in the whole `crates/` tree. `json_escape_for_script` has 1 caller. The Python-side `djust.security.escape_json_for_script` is a **different** sink (debug panel, `JSChain.__html__`); both of its callers feed it `json.dumps(...)` with the default `ensure_ascii=True`, so it already agrees and is correctly untouched. The Django-template backend uses Django's own `json_script` (the name is in `template_filters._BUILTIN_NAMES` and never forwarded), so it is correct by construction.

## Measurement

**Randomized differential**, 3,000 assembled nested values × 4 templates, over an alphabet spanning every branch of the escaper (the seven short forms, the whole C0 range, DEL, `<`/`>`/`&`, `U+2028`/`U+2029`, the surrogate-block boundaries, the BMP boundary, three astral codepoints, a combining mark) with non-ASCII in KEYS as well as values:

| | pre-fix | post-fix |
|---|---|---|
| `{{ p\|json_script:"d" }}` | 860 / 3000 | **0** |
| `{{ p\|json_script }}` | 3000 / 3000 | **0** |
| `{{ p\|json_script:e }}` | 2367 / 3000 | **0** |
| `{{ p\|json_script:"" }}` | 3000 / 3000 | **0** |
| **total** | **9,227 / 12,000** | **0 / 12,000** |

A second seed is clean too, and the in-suite copy asserts its own inputs can SEE the branch (>10% carry a non-ASCII character), because an `ensure_ascii` mutation is invisible to any all-ASCII corpus.

**`scripts/filter-parity-differential.py`**, two real builds (`3cc486ee541d29c6` → `17d533f7ea9dc204`), 351,898 cells:

```
newly AGREEING: 5988
no longer agreeing: 0
  REGRESSIONS : 0
cells where djust PANICKED after : 0   INTRODUCED : 0
live-payload leaks BEFORE : 22   AFTER : 22   INTRODUCED : 0
OK: no agreeing cell regressed, nothing became more permissive, and no cell newly panics.
```

## The un-masking, and what it revealed

Every argument-less `json_script` cell already diverged on the `id`, so any divergence in the JSON **body** sat underneath and could not be attributed — which is why #2405's own corpus shape had to pass an explicit `id`. Fixing the `id` half un-masks that surface, and I expected findings there.

Of the **41 plain `json_script <value>` cells: 41 diverged before, 1 does now.** That one is `d-typed-key`, and it is a genuine, separate defect:

| key | django | djust |
|---|---|---|
| `{True: "b"}` | `{"true": "b"}` | `{"True": "b"}` |
| `{None: "c"}` | `{"null": "c"}` | `{"None": "c"}` |
| `{(1, "t"): "e"}` | `TypeError: keys must be str, int, float, bool or None, not tuple` | `{"(1, 't')": "e"}` |
| `{b"k": "h"}` | `TypeError: … not bytes` | `{"b'k'": "h"}` |

Two mechanisms — a JSON-vs-Python **spelling** for `bool`/`None` keys, and a **refusal** djust does not make (the more-permissive direction). Filed as **#2425** per #1079, and pinned here in `TestTheDivergenceTheUnMaskingRevealed` so it goes red when fixed — with a scope test that walks the differential's own `INPUTS` and asserts `{"d-typed-key"}` is the *complete* set, so a new body divergence cannot hide there.

The 4,148 other still-divergent `json_script` cells all carry an upstream chain filter (`escapeseq|`, `join:"<br>"|`, `upper|` over `d-model`, `l-scalars` number formatting) or an unsupported tag (`{% ifchanged %}`, `{% regroup %}`) — pre-existing, unrelated classes, unchanged by this PR.

## Pins this turns red, rewritten as parity assertions

Three pre-existing tests were written to fail when this was fixed. All three did, which is what the discipline is for.

1. **`test_json_script_key_order_2405.py`** — `TestKnownDivergencesOnTheSamePath`'s two named divergences, plus `test_each_one_agrees_with_django`, whose exact-set assertion (`== {"{{ p|json_script }}"}`) went red the *second* way its docstring said it could.
2. **`test_json_script_escaping_2241.py`** — its 4,000-payload byte-parity differential and its short-form table both passed `ensure_ascii=False` to the reference `json.dumps`, pinning the wrong encoder. Corrected, alphabet extended with `U+2028`, `U+2029` and an astral codepoint, and three `\uXXXX` rows added that the short forms would otherwise shadow.
3. **`test_template_builtins_2347.py`** — `_KNOWN` is now empty; every builtin × `True`/`False`/`None` cell agrees.

**(3) surfaced a defect of its own.** `_is_about_the_literal`'s probe variable was named `__djust_2347_control`, and Django refuses a leading-underscore variable at PARSE time (*"Variables and attributes may not begin with underscores"*). So `DjangoTemplate(control_src)` raised for **all 84** swept cells, the classifier returned `False` unconditionally, and the sweep's `assert not mismatched` was permanently vacuous. Measured both ways:

```
probe='__djust_2347_control'   classification over ALL swept cells: {'django-raised': 84}
probe='djust2347control'       classification over ALL swept cells: {'agrees': 84}
```

Fixed. The `assert value_side > 0` canary is replaced rather than relaxed: its only contributors were the three `json_script` cells, so it fired on a **fix** rather than on a broken classifier — an assertion whose subject can legitimately reach zero measures the wrong thing (#1859). It is now `value_side == 0` (strictly stronger — nothing diverges at all) plus a non-vacuity guard on the classifier itself.

## Gate-off

Harness guards, because a gate-off that does not gate is a tautology one level up (#2129/#2135) and a valid mutation can still be a semantic no-op (#2233): mutation text asserted **found and found exactly once**; mutated source asserted **different**; the crate **rebuilt** between iterations with a failed build reported `INVALID` rather than as a number; pytest **`N error` counted separately from `N failed`**; `cargo` output parsed for `FAILED` and for `error[E`.

| mutation | removes | python | rust |
|---|---|---|---|
| *baseline* | — | 0 failed, 371 passed | 0 failed |
| `ensure_ascii` range → pre-fix control-only | the escape itself | **34 failed** | **4 failed** |
| surrogate pair → single 5-hex escape | the astral arm only | **17 failed** | **1 failed** |
| lowercase hex → uppercase | the hex CASE | **19 failed** | **2 failed** |
| drop the `is_falsy` guard | the falsy-argument arm only | **18 failed** | **2 failed** |
| restore the invented default | the absent-argument arm only | **26 failed** | **2 failed** |
| second stage drops `<` | `json_escape_for_script`, post-deletion | **6 failed** | **1 failed** |

No survivors, and no mutation was equivalent or shadowed.

**Both `id` mechanisms are independently reachable** (#2135) — named, not counted:

* drop the `is_falsy` guard → `test_every_falsy_resolved_argument`, `test_a_falsy_LITERAL_argument_too`, `test_the_id_is_omitted_WHOLE_not_emitted_empty` go red; **`test_no_argument_at_all` and `test_the_invented_default_is_gone` stay GREEN.**
* restore the invented default → `test_no_argument_at_all` and `test_the_invented_default_is_gone` go red.

The structural pin `the_arm_names_no_default_element_id` bans a bare `"data"` literal (the pre-fix `arg.unwrap_or("data")` shape, proven by an empirical canary that also shows the lexer is blind to the arm's own doc comment naming `"data"`). It does **not** catch a writer that spells the attribute inline — measured, by mutating the arm that way and watching it stay green — so a second pin counts `id=`-bearing literals in the arm and goes red on exactly that mutation. Two shapes, two checks, neither shadowing the other.

## Suites

* `pytest tests/ python/tests/ python/djust/tests/ -n auto` — **16,924 passed, 338 skipped, 0 failed** (post-merge with `origin/main` @ `da1ee067`).
* `cargo test -p djust_templates` — 485 lib + all integration tests green. (`cargo test --workspace` cannot link `djust_vdom`'s test binaries in this worktree — a pre-existing pyo3/libpython link issue in this environment, on a crate this PR does not touch; CI is authoritative.)
* `ruff check` / `ruff format --check` / `cargo fmt --check` / `cargo clippy -p djust_templates` — clean.


johnrtipton:
Post-open addendum, all on this branch:

**One extra commit (`258e48d8`, comment-only).** `ArgType::is_falsy` reads `false` for an ABSENT argument, and the field doc now says why that is a refusal to answer rather than a claim the argument is truthy: an absent argument means the parameter takes its own default, and `json_script`'s is `None` (falsy) while `floatformat`'s is `-1` (truthy). No single bit is right for both, which is exactly why the call site reads `arg: Option<&str>` for that half and the bit never answers it.

**Merged `origin/main` @ `da1ee067`** (#2411/#2422) into the branch rather than rebasing — `CHANGELOG.md` uses a local-only `merge=union` driver. Re-ran everything after: full suite **16,924 passed / 0 failed** across all three roots, `cargo test -p djust_templates` green, and the randomized sweep still **0 divergent of 12,000**.

**One flake observed locally, pre-existing and unrelated**: `test_regroup_string_source_2385_2394.py::TestBothMechanismsAreReachable::test_the_renderer_hands_a_string_source_over_quoted` failed once in four full `-n auto` runs of this branch and passes in isolation. It monkey-patches `RegroupTagHandler._decode_source` on a shared registry. Already filed independently as #2427; nothing here touches that path.

johnrtipton:
One precision on the un-masking numbers in the PR body, from re-reading the corpus rather than assuming which cells cover what.

The "41 of 41 plain cells" figure is about the ARGUMENT-LESS spelling only. The corpus's `@arg` axis reaches `json_script` too (it is in `django_argument_filters()`, and `ARG_SPELLINGS` already carries three falsy spellings), so the falsy-`element_id` half has its own covered cells:

| cells | disagree before | after |
|---|---|---|
| `@arg json_script:""` | 16 / 16 | **0** |
| `@arg json_script:False` | 16 / 16 | **0** |
| `@arg json_script:None` | 16 / 16 | **0** |
| `@arg json_script:5` / `:"0"` / `:known_list` / `:known_dict` | 1 each | **0** |

So the falsy-argument axis was already constructible and already failing — 48 cells that no one had attributed, plus the `d-typed-key` row on each truthy spelling. The axis the corpus genuinely cannot construct is an UNQUOTED falsy numeric literal (`{{ p|json_script:0 }}`) and a variable bound to `0` / `[]`; those are covered by the parametrized tables in `TestAFalsyElementIdEmitsNoIdAttribute` instead of by a new corpus entry, because that is the `input-shape` class the manifest already declares UNVERIFIED — no mechanical source in either engine names which falsiness shapes matter.

One `json_script` arg cell still disagrees on both builds and is untouched by this PR: `@arg json_script:0x10` (16 cells). `0x10` is not a Django number literal, so Django resolves it as a variable and raises `VariableDoesNotExist` while djust raises `RuntimeError` with its own wording — the pre-existing exception-class class the corpus records as `<<EXC {type}: {message}>>`, byte-identical before and after.

johnrtipton:
CI green — all 19 checks settled, including the aggregate `test-summary` gate (`python-tests (py3.12)` 16m22s, `rust-tests`, `security-tests`, `playwright-tests`, `browser-smoke`, `benchmarks (serial, thresholds enforced)`, `demo-checks`, `nav-hooks-guard`, both CodeQL analyses). No failures on any run. Ready for review.

Issue #2429, opened 2026-08-29
json_script emits where json.dumps REFUSES — an unserialisable KEY and an unserialisable VALUE are both more-permissive than Django


johnrtipton:
Split out of #2425 per CLAUDE.md #1079 — #2425's fix closes the **spelling** half
(a `bool` / `None` / non-finite `float` KEY now carries the JSON literal) and
deliberately leaves this, the **refusal** half, open, because closing it in the
key position alone would make the two positions disagree.

## The measurement

`json_script` is `json.dumps(value, cls=DjangoJSONEncoder)` in Django, with no
`skipkeys=True` and no `default` that would rescue an arbitrary object. djust
writes `str()` and refuses nothing — in **both** positions.

`{{ p|json_script:"d" }}`, live Django vs live djust:

| p | django | djust |
|---|---|---|
| `{(1, "t"): "v"}` | `TypeError: keys must be str, int, float, bool or None, not tuple` | `{"(1, 't')": "v"}` |
| `{b"k": "v"}` | `TypeError: … not bytes` | `{"b'k'": "v"}` |
| `{frozenset({1}): "v"}` | `TypeError: … not frozenset` | `{"frozenset({1})": "v"}` |
| `{Decimal("1.5"): "v"}` | `TypeError: … not decimal.Decimal` | `{"1.5": "v"}` |
| `{date(2020,1,1): "v"}` | `TypeError: … not datetime.date` | `{"2020-01-01": "v"}` |
| `{complex(1,2): "v"}` | `TypeError: … not complex` | `{"(1+2j)": "v"}` |
| `{SomeEnum.A: "v"}` | `TypeError: … not E` | `{"E.A": "v"}` |
| `{obj: "v"}` | `TypeError: … not Obj` | `{"OBJ": "v"}` |
| **`{"a": obj}`** | `TypeError: Object of type Obj is not JSON serializable` | `{"a": "OBJ"}` |
| **`{"a": b"k"}`** | `TypeError: Object of type bytes is not JSON serializable` | `{"a": "b'k'"}` |
| **`{"a": {1}}`** | `TypeError: Object of type set is not JSON serializable` | `{"a": "{1}"}` |
| **`{"a": complex(1,2)}`** | `TypeError: Object of type complex is not JSON serializable` | `{"a": "(1+2j)"}` |

The bottom four rows are the reason this is not a key-position bug. **The value
position is exactly as permissive**, and has been for longer. #2425's key table
[...]

Issue #2270, opened 2026-08-27
bug: pprint and json_script spell a float with Rust's {} where Python uses repr — the third site of the #2258 split


johnrtipton:
Found while fixing #2258, and deliberately left out of it (#1079 scope discipline).

#2258 is about `Display for Value::Float`. Its fix routes `Display` through
Django's `numberformat.format` rules and the *string-filter* boundary through
`python_float_repr` (Django's `@stringfilter` consumes `str(value)`), and
`py_repr` — the nested-container spelling — through `python_float_repr` too.

That is three of the sinks. Two more still format a float with Rust's `{}`:

| site | djust | Python |
|---|---|---|
| `crates/djust_templates/src/filters.rs` `pprint_value` — `Value::Float(f) => format!("{f}")` | `100000000000000000000` | `pprint.pformat(1e20)` -> `1e+20` |
| `crates/djust_templates/src/filters.rs` `value_to_json` — `Value::Float(f) => format!("{f}")` | `100000000000000000000`, `NaN`, `inf` | `json.dumps` -> `1e+20`, `NaN`, `Infinity` |

Measured over 31 scalars x {list, tuple, dict} x {`{{ p }}`, `{{ p|pprint }}`}:
after #2258 every bare `{{ p }}` cell agrees and every remaining float
divergence is in `|pprint`. `json_script` is the same shape one filter over —
and note `json.dumps(float("inf"))` is `Infinity`, which is not valid JSON, so
matching Django there is a decision and not only a spelling.

Both predate #2258 (they were equally wrong on `main`), which is why they are
filed rather than folded in: neither is `Display`, neither is a
`@stringfilter`, and each has its own output contract.

Fix shape, if taken: `djust_core::decimal::python_float_repr` at both sites,
plus an explicit decision on `json_script`'s non-finite spellings. The sink
grep that finds every one of these is `format!("{f}")` over
`crates/djust_templates/src/` — pair it with a structural test pinning the
caller SET so the next omission fails loudly (v1.1.1-2 retro).

Related: #2258, #2203 (`py_repr`), #2214 (`Decimal` repr).

PR #2430, opened 2026-08-29
fix(templates): json_script spells a dict KEY the way json.dumps spells it (#2425)

### LINKED ISSUE #2425: json_script: a bool/None dict KEY is spelled True/None where json.dumps writes true/null, and a tuple/bytes/object key is EMITTED where json.dumps refuses

johnrtipton:
Closes #2425.

`json.dumps` does not call `str()` on a non-`str` dict key. It has its own
five-entry coercion table, and CPython's `c_make_encoder` writes it in this
order: `str` unchanged, then the three JSON literals for `True` / `False` /
`None`, then `float.__repr__`, then `int.__repr__`, then `TypeError`. djust
routed the key through `ObjectKey::to_display_string()`, so

```
{{ p|json_script:"d" }}     p = {True: "b", None: "c"}

django  <script id="d" …>{"true": "b", "null": "c"}</script>
djust   <script id="d" …>{"True": "b", "None": "c"}</script>
```

One new sibling of `json_float_body` (#2270) — `json_key_body` — called from
the single object-key site in `value_to_json`. The `#2249` count pin still sees
its four `json_string_body` sites, because the helper wraps the ARGUMENT rather
than adding a fifth escape.

## The re-derived divergent-key set

The issue's table listed nine key types and concluded *"the `int` and `float`
arms agree by coincidence, because `str(0)` and `str(1.5)` are already the JSON
forms"*. Re-derived by running `json.dumps` + live Django + djust over **29**
key types rather than inheriting that list, the premise is **wrong rather than
incomplete**: the old coercion was never `str()`.

`ObjectKey::to_display_string()` is the key's TEMPLATE display — what
`{% for k in d %}{{ k }}{% endfor %}` writes — and it parts company with
`float.__repr__` well before infinity:

| key | `str()` | `to_display_string()` (old) | `json.dumps` / Django |
|---|---|---|---|
| `True` / `False` | `True` / `False` | `True` / `False` | `true` / `false` |
| `None` | `None` | `None` | `null` |
| `float('inf')` / `float('nan')` | `inf` / `nan` | `inf` / `nan` | `Infinity` / `NaN` |
| `1e16` | `1e+16` | `10000000000000000` | `1e+16` |
| `1e-5` | `1e-05` | `0.00001` | `1e-05` |
| `0`, `1.0`, `1.5`, `-3`, `12345678901234567890`, `""`, `"a"` | — | — | agrees |
| `str` subclass, `int` subclass, `IntEnum`, `str`-Enum | — | — | agrees (subclass rows CPython accepts) |
| `tuple`, `()`, `frozenset`, `bytes`, object, `Decimal`, `complex`, `date`, `Enum`, `range` | — | — | **`TypeError`** — see below |

Only the middle band of small finite floats coincided, and that is the band the
issue sampled. The non-finite keys came out of the key-type sweep; **the
exponent band came out of the gate-off** — deleting the `Float` arm reddened
`1e16` and `1e-05`, which nothing had predicted.

The refusal set is also wider than the issue's four: `Decimal`, `date`,
`complex`, `Enum` and `range` keys are refused by `json.dumps` too, while
`IntEnum` / `str`-Enum / `str`-subclass / `int`-subclass keys are ACCEPTED (they
are `int` / `str` subclasses and CPython calls `int.__repr__` / passes the `str`
through). The issue's guess that the key type "may need to be preserved across
the PyO3 boundary" is also already false — `ObjectKey` (#2339) carries `Tuple`,
`Decimal`, `BigInt` and `Other { display, repr }` — though `Other` collapses
`bytes` / `frozenset` / `Enum` / an instance into one variant, so Django's
type-named message is not reproducible from Rust as things stand.

## Key-vs-value refusal: why the refusal half is NOT closed here

The design question was settled by measuring what djust does with an
unserialisable **VALUE**, not by preference:

| p | django | djust |
|---|---|---|
| `{"a": object()}` | `TypeError: Object of type Obj is not JSON serializable` | renders `{"a": "OBJ"}` |
| `{"a": b"k"}` | `TypeError: … bytes …` | renders |
| `{"a": {1}}` | `TypeError: … set …` | renders |
| `{"a": complex(1,2)}` | `TypeError: … complex …` | renders |

**djust is exactly as permissive in the value position.** Refusing keys alone
would make the two positions disagree — a new inconsistency wearing a fix's
clothes rather than a narrowing. So this PR closes the SPELLING half only, and
the refusal half is filed as **#2429** as one decision covering both positions.

It is pinned live in `TestTheRefusalHalfIsNotClosedHere`, whose second method
asserts the value position is permissive and *is* the reason the first is left
alone: if the value position is ever made strict, that method goes red and the
key position must follow in the same change. The `date` key is the sharpest
argument for taking them together — `DjangoJSONEncoder.default` serialises a
date VALUE (both engines agree) while a date KEY never reaches the hook, so any
refusal design has to model the encoder hook, not `json.dumps`'s bare table.

## Differential, before → after

Two builds of djust, side by side (HEAD's `filters.rs` built into a separate
tree, this branch in the worktree):

| axis | before | after |
|---|---|---|
| key types (29) | **18 divergent** | **10 divergent** (0 moving the other way) |
| `scripts/filter-parity-differential.py` corpus, argument-less `json_script` (41 INPUTS) | 1 (`d-typed-key`) | 1 (`d-typed-key`) |

The 10 remaining key types are exactly the refusal set. `d-typed-key` stays
divergent because of its `(1, "t")` key, not its `True` / `None` ones — so
#2413's scope claim survives this fix, and is re-worded to say why, plus to say
that it is a claim about the CORPUS and was silent about this axis (the corpus
carries no non-finite and no exponent-form float key).

## Tests

`python/tests/test_json_script_typed_keys_2425.py` — 11 test cases, 38 with
parameterisation. `TestEveryKeyTypeReDerived` is the sweep above, kept as a test
so the next key-type question is answered by running it.

#2413's `test_a_bool_or_None_KEY_is_spelled_Python_not_JSON` went red exactly as
its class docstring said it should, and is flipped to
`test_a_bool_or_None_KEY_is_now_spelled_the_JSON_way`.

### Gate-off

Five mutations, each rebuilding the crate and clearing `__pycache__`; the
harness asserts the mutation text matched **exactly once**, that the source
changed, that the build succeeded, and counts pytest's `N error` apart from
`N failed`:

| mutation | reddened |
|---|---|
| whole fix reverted (HEAD `filters.rs`) | 16 |
| `Bool(true)` arm → `"True"` | 7 (incl. `[True-true]`, **not** `[False-false]`) |
| `Bool(false)` arm → `"False"` | 4 (incl. `[False-false]`, **not** `[True-true]`) |
| `None` arm → `"None"` | 6 |
| `Float` arm deleted | 7 (all three non-finite **and** `1e16` / `1e-05`) |

No survivors, and each of the four arms is independently reachable — a fix
handling only the truthy `bool` fails a test.

## Verification

- `pytest tests/ python/tests/ python/djust/tests/ -n auto` → **16963 passed,
  338 skipped, 0 failed**
- `cargo test -p djust_templates --lib --release` → 485 passed
- `cargo fmt --check`, `cargo clippy -p djust_templates --all-targets`, `ruff
  check`, `ruff format --check`, `scripts/check-changelog-test-counts.py` — all
  clean

Refs #2413, #2405, #2339, #2270, #2241, #1079.


johnrtipton:
`python-tests (py3.12)` is red on **#2427's flake, not on this diff** — this PR
touches `crates/djust_templates/src/filters.rs` and three `json_script` test
files and has nothing to do with `{% regroup %}`.

```
[gw2] [ 50%] FAILED python/tests/test_regroup_string_source_2385_2394.py::TestBothMechanismsAreReachable::test_the_renderer_hands_a_string_source_over_quoted
AssertionError: … that difference IS the fix's renderer half: []
==== 1 failed, 16945 passed, 355 skipped …
```

That is the third unrelated PR to be reddened by it (#2420, #2424, this one),
and `gw2` + `captured == []` is the exact signature reproduced and root-caused
in **#2432**: `test_custom_tag_return_escape_2379.py`'s handler audit
`importlib.reload`s `djust.template_tags.regroup`, which registers a NEW class
object, so the spy's module-level import is stale and never fires.

Everything else here is green, including `rust-tests`, `demo-checks`,
`benchmarks` and `python-tests (py3.14t)`. Locally this branch is
**16963 passed, 338 skipped, 0 failed** over all three roots.

Merge #2432 first, then re-run this job.

johnrtipton:
Re-ran `python-tests (py3.12)` with no code change (the remedy that worked for
#2420). It failed again, identically:

```
[gw2] [ 50%] FAILED python/tests/test_regroup_string_source_2385_2394.py::…::test_the_renderer_hands_a_string_source_over_quoted
AssertionError: … that difference IS the fix's renderer half: []
==== 1 failed, 16945 passed, 355 skipped, 88 warnings in 591.98s (0:09:51) =====
```

Same worker (`gw2`), same `captured == []`, both runs. On this branch the
co-scheduling is stable rather than a coin flip, so **a re-run will not clear
it** — #2432 has to land first.

This is also a clean natural experiment for #2432's diagnosis: on that PR's run
the audit and the target happened to land on **different** workers (`gw2` /
`gw1`), so it could not have failed there whether fixed or not; here they land
together twice and fail twice. Worker co-location is the variable, exactly as
#2432 says.

Everything else on this PR stays green, and the branch is
16963 passed / 0 failed locally over all three roots.

**Order: merge #2432, then update this branch from `main` and re-run.**

johnrtipton:
Updated from `main` (merge, not rebase — `CHANGELOG.md` is `merge=union` and
that is local-only) now that #2432 has landed. **All 19 checks green, including
`test-summary`.**

Merge hygiene, checked rather than assumed:
- set-difference of the `[Unreleased]` section against `origin/main`: **0 lines
  dropped**, and the only additions are this PR's own 5 CHANGELOG lines;
- `scripts/check-changelog-tagged-sections.py` → *OK: 124 shipped sections match
  `v1.1.0`* (the #2028 guard);
- `scripts/check-changelog-test-counts.py` → clean;
- local full three roots at `-n 4`: **16991 passed, 338 skipped, 0 failed**.

**What this green does and does not show.** It shows this PR's own changes pass.
It is *not* evidence that #2432 fixed the flake, because the failing condition
was absent again:

```
[gw2] [ 20%] PASSED …test_custom_tag_return_escape_2379.py::TestEveryRegisteredHandlerIsAccountedFor::…
[gw1] [ 46%] PASSED …test_regroup_string_source_2385_2394.py::…::test_the_renderer_hands_a_string_source_over_quoted
[gw1] [ 46%] PASSED …test_regroup_string_source_2385_2394.py::…::test_the_spy_reaches_a_RE_REGISTERED_handler
```

Audit on `gw2`, target on `gw1` — the pre-fix code would have passed this run
too. And the split looks *caused* by the merge rather than lucky: every local
`-n 4` run before it, pre- and post-fix, put the pair together on `gw2`, and the
first run after merging #2424's 427-line `test_render_slot_markup_2421.py` (plus
#2432's extra case) splits them. Collection size moves xdist's chunking, so the
condition may simply not arise at this collection size any more.

The load-bearing line is the third one: **`test_the_spy_reaches_a_RE_REGISTERED`
`_handler` passed on CI**, and that test is red on the bug regardless of
scheduling — it reproduces the stale-class shape in-process. That, the gate-off
(0→1 alone, 0→2 co-scheduled), and the three co-scheduled local `-n 4` runs
remain the actual proof; these two CI greens are consistent with it, not
evidence for it.

Ready to merge.

Issue #2411, opened 2026-08-29
{% if %} swallows an operand's COMPILE-time refusal when an earlier argument fails to resolve


johnrtipton:
Found while fixing #2409 and filed separately per CLAUDE.md #1079 — measured on
that branch, PRE-EXISTING, and not touched there.

## The mechanism

Django compiles a `{% if %}` operand's filter chain with `compile_filter`, so
every refusal that chain carries — an unknown filter, a wrong argument count,
a lexer remainder, a name beginning with `_` — is raised at COMPILE time,
before any value is resolved.

djust resolves a tag operand at RENDER time, left to right, in
`renderer::get_value_safe`. `{% if %}` then legitimately absorbs a
`VariableDoesNotExist` (`evaluate_condition_for_if`, matching Django's
`ignore_failures`) — and an *earlier* argument that fails to resolve makes the
whole condition falsy before the *later* filter's refusal is ever reached:

```
{% if p|cut %}          django  <<TemplateSyntaxError: cut requires 2 arguments, 1 provided>>
                        djust   <<RuntimeError: cut requires 2 arguments, 1 provided>>   agrees

{% if p|date:.|cut %}   django  <<TemplateSyntaxError: cut requires 2 arguments, 1 provided>>
                        djust   ''                                                       masked
```

The only difference between the two is a preceding `date:.` whose argument
does not resolve. `.` is a legal `var_arg` for Django's lexer, so both engines
accept the spelling; Django never resolves it, because it refused the template
three characters later.

## Scale, and the four Django causes it masks

Measured on the #2409 branch with a randomized sweep of **12,356 distinct
templates** (five filters × 37 argument atoms × four shapes). Every row where
djust renders a template Django refuses at compile time is a `{% if %}`, and
there are 120 of them:

| Django's refusal | rows |
|---|---|
| `Variables and attributes may not begin with underscores` | 56 |
| `add`/`cut`/`default`/`join` `requires 2 arguments, 1 provided` (#2400) | 33 |
| `Could not parse the remainder` (#2409) | 16 |
| `Could not parse some characters` | 12 |
| `Unused 'x' at end of if expression` | 3 |

`{{ }}`, `{% for %}` and `{% with %}` have **zero** rows: they do not swallow,
so the same operand refuses on both engines. This is a `{% if %}`-only
divergence, and it is over-permissive — a template Django will not compile
renders here, as a silently falsy branch.

## Direction

Over-permissive, and silent: the page renders with the `{% if %}` taken as
[...]

PR #2452, opened 2026-08-30
fix(templates): an unknown filter name refuses at PARSE time, as Django refuses it (#2419)

### LINKED ISSUE #2419: `Invalid filter` is a RENDER-time lookup, so a dead branch or short-circuited operand never refuses it
Found while fixing #2411 and filed separately per CLAUDE.md #1079 — measured on

## The timing difference

Django looks a filter name up in `FilterExpression.__init__`, at COMPILE time:

```python
filter_func = parser.find_filter(filter_name)   # raises TemplateSyntaxError: Invalid filter: 'x'
```

djust looks it up at RENDER time, in `filters::apply_filter_full_safe` — for
`{{ … }}` as much as for a tag operand. So a template carrying an unknown
filter in a branch that never renders compiles here and refuses on Django:

```
{% if 0 %}{{ p|nosuchfilter }}{% endif %}   django  <<TemplateSyntaxError: Invalid filter: 'nosuchfilter'>>
                                            djust   ''

{% if 0 %}{% if p|nosuchfilter %}Y{% endif %}{% endif %}
                                            django  <<TemplateSyntaxError: Invalid filter: 'nosuchfilter'>>
                                            djust   ''
```

#2411 moved the ARITY and LEXER checks for a tag operand to parse time and
deliberately left this one alone, because moving it for one shape only would be
new parallel-path drift: `{{ }}` has the same gap, and both would have to move
together.

## Scale

Measured on the #2411 branch over a 13,202-template sweep: **94 of the 495
remaining "djust renders what Django refuses" cells**, second only to the
underscore rule (#2418).

## The thing to settle first, because it is the whole difficulty

`filter_arity::parse_time_arity_error`'s doc-comment already states the
constraint this runs into:

> `None` for a name outside the built-in table, so an unknown filter is left to
> the render-time "Unknown filter" path **and a project's custom filter is not
> refused by a table that does not describe it.**

A parse-time lookup must therefore answer: is `filter_registry` guaranteed to
be populated before any template a project renders is PARSED? Django's answer
is `{% load %}` plus an `INSTALLED_APPS` scan at compile time; djust's registry
is filled from Python at runtime via `register_custom_filter`. If a template
can be parsed (or cached) before a project's `AppConfig.ready()` has
registered its filters, a parse-time refusal turns a working project into a
`TemplateSyntaxError` — the failure mode the current design avoids by never
checking early.

Options, none of them obviously right:

1. **Check at parse time unconditionally.** Django-exact, and breaks any
   project whose registration happens after a parse. Needs the ordering
   guarantee established first, empirically.
2. **Check at parse time only for names the registry can be sure about** —
   i.e. refuse only if the name is not a built-in AND the registry has been
   sealed. Requires a "registration is closed" signal that does not exist.
3. **Leave it at render time and say so.** The current state, but undocumented;
   the cost is that a dead branch and a short-circuited operand never refuse.

Whichever is chosen, it has to move `{{ }}` and the tag operands together, or
it is #1646 again.

`python/tests/test_tag_operand_parse_time_2411.py::TestTheTwoRulesThisDoesNotClose::test_unknown_filter_is_a_RENDER_time_lookup_on_every_shape`
pins the current behaviour as known-open and goes red the day this is fixed.
`scripts/filter-parity-differential.py`'s `masked-refusal` axis carries
`nosuchfilter` as one of its six refusal classes, so the number above is
re-measured on every run rather than remembered.

Refs #2411, #2400, #1079, #1646.

johnrtipton:
Closes #2419.

## The defect

Django looks a filter name up in `FilterExpression.__init__` — `parser.find_filter(filter_name)`, at COMPILE time. djust looked it up in `filters::apply_filter_full_safe`, on the value, which only happens if the node renders. So a name nothing implements compiled here and refused on Django whenever the position was unreached:

```
{% if 0 %}{{ p|nosuchfilter }}{% endif %}   django  <<TemplateSyntaxError: Invalid filter>>   djust  ''
{% if 0 and p|nosuchfilter %}Y{% endif %}   django  <<TemplateSyntaxError: Invalid filter>>   djust  ''
```

## The blocker the issue recorded, and why it does not hold

#2411 left this class alone because djust's filter registry is filled from Python at runtime, so a parse-time refusal could in principle refuse a project's own `@register.filter`. That is the right question. Four facts settle it, each **run** rather than reasoned — all four are pinned in `TestTheRegistryIsPopulatedBeforeAnythingCanBeParsed`:

1. **The registry is complete at the end of `django.setup()`.** `DjustConfig.ready()` warms the Django→Rust bridge; measured in a subprocess, `has_custom_filter('field_value')` is already `True` before any request. (Measured both ways: with `filter_bridge_warm: False` it is `False`, which is what makes the assertion mean something.)
2. **djust never parses a template outside a render call.** `Template::new` is reached from `render`, `render_with_diff`, `render_binary_diff` and `render_template*` — all renders. There is no compile-now-render-later API, so there is no window in which a user template is parsed with an empty registry.
3. **Django's `Engine.template_libraries` is filled from `INSTALLED_APPS` at engine construction, without `{% load %}`** — so the one bootstrap sweep sees every filter Django itself could ever see. djust's registry is therefore a **superset** of Django's per-template view of the names, and refusing only names in *neither* the built-in table *nor* the registry can never refuse a template Django compiles.
4. **A refusal is never cached.** `TEMPLATE_CACHE` / `PARSED_TEMPLATE_CACHE` are written only after a successful parse, so a filter registered later is picked up on the next render rather than poisoning the process.

## The fix

| what | where |
|---|---|
| the oracle — "can this engine dispatch the name at all?" | `filters::is_known_filter` (ARITY table ∪ custom-filter registry) |
| the refusal | `parser::parse_filter_specs`, after the arity check |
| the non-`PyResult` registry probe | `filter_registry::is_registered_custom_filter` |
| the third render entry, armed | `DjustTemplate.render` |

**One site closes both shapes.** #2411's condition for moving this at all was that `{{ … }}` and the tag operands move *together*, since doing one alone would be new parallel-path drift (#1646). `{{ … }}` reaches `parse_filter_specs` through `parse_token`; every tag operand reaches it through `validate_tag_operand`. `TestOneSiteClosesBothShapes` pins that the call appears exactly once in `parser.rs` and that both entry points reach it.

**The oracle is the dispatch table, not a copy of it.** A second list of the 57 built-in names would be the same drift one layer down, and a *silent* one: an arm in `apply_builtin_filter`'s match missing from the list would refuse a filter the engine implements. `TestTheOracleIsTheDispatchTable` extracts the match arms from `filters.rs` and asserts they are exactly the ARITY table's names, in both directions; all 57 are separately checked to still compile in an unrendered position.

**Not refused:** `{% comment %}` and `{% verbatim %}` bodies, which Django does not compile either. They are the control — refusing them would be *stricter* than Django.

**Message:** keeps djust's existing `Unknown filter: <name>` rather than Django's `Invalid filter: '<name>'`. That substring is a published contract (`template/rendering.py` keys its "not supported by the Rust engine" hint off it, `test_rust_custom_filters_1121.py` pins it), and a second spelling for one condition would be a drift of its own. One ordering is **not** Django's and is recorded rather than hidden: `{{ p|nosuchfilter:"a":"b" }}` reports the lexer remainder here and `Invalid filter` there, because `split_filter_spec` is what produces the name at all. Both engines refuse the template.

**Also fixed, because this change depends on it:** `DjustTemplate.render` — the plain-Django-view path through `DjustTemplateBackend` — was the third top-level render entry and relied on the startup warm alone. A project setting `filter_bridge_warm = False` had no bridge there at all and its custom filters did not resolve. Same shape as #2223, one entry point over.

## Corpus

`masked-refusal` already carried `nosuchfilter`, but **every one of its positions wrote a TAG operand** — so the corpus could have reported the tag half of a compile-time refusal closed while the `{{ }}` half stayed open, and `Invalid filter` was open on both. Adds `dead-branch-var`, `else-branch-var`, `block-in-dead-branch-var`. A cross of two covered axes is its own axis, which is the lesson that axis exists to carry.

Measured over **353,909 cells, two genuinely different builds** (`879c6fbd` → `e55e96cd`):

- **44 masked-refusal cells move out of "djust renders what Django refuses"; 0 move in.**
- **0** agreeing cells regress, **0** cells newly panic, live-payload leaks unchanged at 22.
- Every other axis: **0 moved**.
- **0** `nosuchfilter` cells still render where Django refuses.

Note the tool's headline agreement count is *raw string equality*, so it is structurally blind to a refusal-class fix (both engines raise, with different wording). The moved-cell count is the number to read.

## Tests

88 cases in `python/tests/test_unknown_filter_parse_time_2419.py`. The known-open pins in `test_tag_operand_parse_time_2411.py` and `test_filter_arity_2400.py` now assert the refusal. Two fixtures in `test_template_edge_cases.py` named invented filters (`filter`, `match`) as syntax scaffolding — templates Django refuses too — and move to real ones.

Gate-off, with the mutation asserted to match exactly once, the source asserted to change, the crate rebuilt with the `.so` mtime asserted to advance, `__pycache__` cleared, and pytest *errors* counted apart from *failures*:

| mutation | red |
|---|---|
| M1 the parse-time refusal never fires | 13 |
| M2 every name is known | 14 |
| M3 the custom-filter half of the oracle is dropped | 5 |
| M4 the built-in half of the oracle is dropped | 59 |
| M5 the backend render path stops arming the bridge | 1 |
| restored | 0 (88 passed) |

No survivors. **One harness bug is recorded rather than papered over**: M5 is Python-only, so it first ran against M4's *mutated* `.so` and reported 58 failures for a one-line change that should redden exactly one. The harness now rebuilds after every Rust row, and the number went to 1 — an instrument reading the wrong input reports a plausible number, not an obviously wrong one.

## Verification

- `pytest tests/ python/tests/ python/djust/tests/ -n auto` → **17330 passed, 338 skipped, 0 failed** (post-merge with `origin/main`)
- `cargo test -p djust_templates --release` → all green; `cargo clippy` → 0 warnings; `cargo fmt --check` → clean
- `check-changelog-test-counts.py`, `check-changelog-tagged-sections.py` → clean


johnrtipton:
Filed #2454 for a pre-existing measurement gap this PR ran into and did not fix (per #1079): `--compare`'s agreement sets use RAW string equality, so a fix whose whole effect is "djust now refuses what Django refuses" moves zero cells into agreement — both engines raise, with different wordings. It is why this PR's evidence is the `44 moved` line rather than the headline number, and why `--compare` prints its "the change moves nothing" NOTE on a run that moved exactly what it set out to move.

Issue #2451, opened 2026-08-30
Five sequence filters render where Django's iteration/subscript raises — the other column of #2435's bucket (1,088 cells)


johnrtipton:
Split out of #2435 per CLAUDE.md #1079. #2435 closed the `int(value)` column of
that bucket; this is the other one, and it is a different mechanism.

## The divergence

Five built-ins **iterate or subscript** their value, and Django's body has no
guard, so a value that is not a sequence raises and 500s the page. djust fails
soft and renders:

```
{{ p|first }}   p = 5    django  <<TypeError: 'int' object is not subscriptable>>
                         djust   ''

{{ p|escapeseq }}  p = 5 django  <<TypeError: 'int' object is not iterable>>
                         djust   '42'

{{ p|phone2numeric }}  p = 5   django  <<AttributeError: 'int' object has no attribute 'lower'>>
                                djust   '5'
```

Direction: djust is **over-permissive** — a template Django refuses renders here.

## Scale, re-measured

On the repo's two-build differential (`scripts/filter-parity-differential.py`,
353,837 cells) after #2435 landed, the `{% widthratio %}` bucket — cells where
Django raises `TemplateSyntaxError` and djust renders — is **1,088**, down from
4,222. Attributing each to the filter whose Django body raises:

| filter | cells | Django's raise |
|---|---|---|
| `escapeseq` | 218 | `TypeError: not iterable` |
| `safeseq` | 218 | `TypeError: not iterable` |
| `unordered_list` | 218 | `TypeError: not iterable` |
| `first` | 216 | `TypeError: not subscriptable` |
| `last` | 216 | `TypeError: not subscriptable` |
| `json_script` | 2 | tracked at #2429 |

They surface inside `{% widthratio %}` for the same reason #2435's did:
`WidthRatioNode.render` wraps operand RESOLUTION in
`except (ValueError, TypeError): raise TemplateSyntaxError`, so a filter that
raises while an operand resolves refuses the whole template. The `{{ p|f }}`
twin of every one of the 1,088 also diverges, which is where the fix belongs.

The single-filter `{{ }}` axis carries more of the same class that the
`widthratio` shape does not reach — measured on the same run, 113 cells:

```
  30  phone2numeric   {'AttributeError': 30}
  19  last            {'KeyError': 7, 'TypeError': 12}
  16  first           {'KeyError': 4, 'TypeError': 12}
  12  escapeseq       {'TypeError': 12}
  12  safeseq         {'TypeError': 12}
[...]

PR #2453, opened 2026-08-30
fix(templates): int(value) is a chokepoint, and which exception it raises decides the answer (#2435)

### LINKED ISSUE #2435: {% widthratio %} coerces a non-numeric first operand to 0 where Django refuses the template
Found while measuring #2418, filed separately per CLAUDE.md #1079 — pre-existing
and not touched there.

## The divergence

`{% widthratio %}`'s FIRST operand is converted with `float()` by Django's
`WidthRatioNode.render`, and a value that will not convert is a refusal:

```
{% widthratio p|divisibleby:"2" 10 100 %}   django  <<TemplateSyntaxError: widthratio final argument must be a number>>
                                            djust   '0'

{% widthratio 5 10 p %}   (p = "abc")       django  <<TemplateSyntaxError: widthratio final argument must be a number>>
                                            djust   <<RuntimeError: widthratio final argument must be a number>>   agrees
```

So the THIRD operand already agrees and the FIRST does not — djust coerces a
non-numeric value (here a `bool` produced by `divisibleby`) to `0` and renders
where Django raises. Django's message names "final argument" for all three
positions, which is its own long-standing wording, not a clue about which
operand failed.

Direction: djust is **over-permissive** — a template Django refuses renders
here. Not an escaping bug; nothing reaches the page that would not otherwise.

## Scale

Re-measured on the repo's own two-build differential
(`scripts/filter-parity-differential.py`, 353,714 cells): of the cells where
Django raises a `TemplateSyntaxError` and djust renders, **4,222 of 4,250 are
this one** after #2418 closed the 516 underscore-rule cells. It is now the
largest single remaining bucket in that corpus by a factor of 150 over the
runner-up (`Invalid filter`, 28 cells, tracked at #2419).

Every one of the 4,222 is on the existing `tag` axis, shape `widthratio` —
`{% widthratio p|<filter> 10 100 %}` over the corpus inputs — so the cells
already exist and a fix is measurable the day it lands.

## Where to look

* `crates/djust_templates/src/renderer.rs`, the `Node::WidthRatio` arm — the
  conversion of `value` / `max_value`, and what it does with a value that is
  not a number.
* Django's `WidthRatioNode.render` (`django/template/defaulttags.py`) raises
  `TemplateSyntaxError` for a `ValueError` on ANY of the three, and returns
  `""` for a `VariableDoesNotExist` on the first two — so "unresolvable" and
  "resolves to something non-numeric" are DIFFERENT answers and both need a
  cell before the fix is written.
* The refusal is at RENDER time on Django, not compile time, so this is not a
  parse-time check like #2411/#2418 — `{% widthratio %}`'s operands are already
  validated at parse time for their filter chains, and this is about the
  resolved VALUE.

Refs #2418, #1079.

johnrtipton:
Closes #2435.

## The issue's premise is wrong, and the cited code was already right

#2435 says `{% widthratio %}`'s **first** operand is converted with `float()`
and *"a value that will not convert is a refusal"*, and locates the fix in
`renderer.rs`'s `Node::WidthRatio` arm. Measured against a live Django, both
halves fail:

| the issue said | running it showed |
|---|---|
| `float(value)` failing is a refusal | it is a `ValueError` in `WidthRatioNode.render`'s **second** `try`, which Django catches and answers `""` |
| djust coerces it to `0` and renders | `{% widthratio p 10 100 %}` over `p="abc"` renders `""` on **both** engines, and did before this PR |
| only the FIRST operand is affected | all three refuse — the raise is from `resolve()`, in the **first** `try` |
| the bool from `divisibleby` is what gets coerced | `float(False)` is `0.0`; `{% widthratio False 10 100 %}` is `0` on both engines. The refusal is `divisibleby` **itself** raising from `int(value)`; `float()` never runs |

A fix at the cited location would have introduced a divergence, not closed one.
`TestTheIssuesOwnClaims` measures each row above so a reader can see which half
of the issue survived contact with a running Django.

## What the 4,222 cells actually are

Re-measured on `scripts/filter-parity-differential.py` (353,837 cells,
release build): **4,222**, matching the issue exactly. Attributing each cell to
the filter whose Django body raises splits them into two mechanisms:

| culprit | cells | Django's raise |
|---|---|---|
| `divisibleby` | 2,132 | `int(value)` — ValueError + TypeError |
| `get_digit` | 988 | `int(value)` — TypeError |
| `escapeseq` `safeseq` `unordered_list` `first` `last` | 1,100 | iteration / subscript TypeError |
| `json_script` | 2 | #2429 |

Every one of the 4,222 has a `{{ p|expr }}` twin where Django raises and djust
renders — the `{% widthratio %}` node's plumbing was never the problem. This PR
closes the **`int(value)`** column; the iteration column is filed as **#2451**
per #1079.

## The fix: one chokepoint, and WHICH exception it raises

`int()` was spelled **four** ways in `djust_templates` — `int_digits_of`,
`python_int_from_str`, `divisibleby`'s own inline `parse::<i64>()`, and
`renderer::py_int` — and not one said which exception Python raises. That
distinction is observable, because Django's four value-side `int()` call sites
each catch a *different* subset:

| filter | Django's body | `int("abc")` ValueError | `int(None)` TypeError | `int(±inf)` OverflowError |
|---|---|---|---|---|
| `divisibleby` | `int(value) % int(arg)` | raises | raises | raises |
| `get_digit` | `except ValueError: return value` | the input | **raises** | **raises** |
| `add` | `except (ValueError, TypeError)` | concatenates, else `""` | `""` | **raises** |
| `filesizeformat` | `except (TypeError, ValueError, UnicodeDecodeError)` | `0 bytes` | `0 bytes` | **raises** |

`int(float("nan"))` is a **ValueError** and `int(float("inf"))` an
**OverflowError** — so `{{ p|filesizeformat }}` renders `0 bytes` for a NaN and
refuses for an infinity. A rule of "refuse a non-finite value" passes half this
table wrongly, which is why every row is asserted against live Django *before*
it is asserted against djust.

`python_int_value(&Value) -> Result<String, IntValueError>` answers both halves
over arbitrary-precision digits; `python_int_str` is the one reading of
`int(str)` the other spellings now delegate to. `renderer::py_int` is a wrapper
over it.

## What it fixes beyond the refusals

`divisibleby`'s inline reader matched `Value::Integer` and a string
`parse::<i64>()` and nothing else, so these were all silently `False`:

```
{{ 2.5|divisibleby:"2" }}          django True    djust False -> True
{{ False|divisibleby:"2" }}        django True    djust False -> True
{{ Decimal("4.5")|divisibleby:2 }} django True    djust False -> True
{{ "1_0"|divisibleby:"2" }}        django True    djust False -> True
{{ (10**30)|divisibleby:"2" }}     django True    djust False -> True
```

`add` stopped float-coercing a **string** value, so `{{ "34.23234"|add:"1" }}`
concatenates as Django does instead of answering `35`. And `{% widthratio %}`'s
third operand went through the same chokepoint — its `as i64` cast **saturated**:

```
{% widthratio 10 2 p %}  p = 10**30
  django  4999999999999999817948147482624
  djust   46116860184273879040   ->   4999999999999999817948147482624
```

`{% widthratio %}`'s `float()` learned `_` too (`"1_0"` is 500, not `""`).

## Measurement

Two-build differential, 353,837 cells, two genuinely different `_rust` builds:

```
agree BEFORE : 257295          REGRESSIONS               : 0
agree AFTER  : 257331          newly more-permissive     : 0
                               new panics                : 0
"Django refuses, djust renders":  4,222  ->  1,088
```

13,457 tag-axis and 1,522 chain-axis cells moved. Independently, the
`{{ p|safe|F }}` residue sweep in `test_safe_stringifies_scalars_2303.py` —
which named `add` and `divisibleby` as its two known-disagreeing filters —
now reports an **empty residue** across 405 comparable cells.

## Compatibility

A template calling one of these four filters on a value `int()` rejects, in the
subset that filter's own `except` does not catch, now raises where it rendered.
**Every such template was already broken under Django's own engine** — to find
them before upgrading, render under Django.

Grepped `examples/`, `docs/` and every `.html` in the repo: **no shipped
template, demo, or documented pattern** uses these filters at all. The full
suite needed no template edits. The error is a `RuntimeError` naming the
chokepoint (`calls int() on its value`), distinct from `divisibleby:"0"`'s
`ZeroDivisionError` and from #2328's argument refusal, and is contained by the
consumer's error frame over a LiveView socket.

## Tests

New file `python/tests/test_int_value_chokepoint_2435.py` (106 cases):
the issue's own claims measured; one row per (filter, exception) with Django
checked first; the value-reader shapes; all three `{% widthratio %}` positions;
a **structural pin on the caller SET** (not a floor, #1125) that goes red if a
fifth arm grows its own value parse; and a **1,500-cell randomised
differential** that names its two residues rather than dropping those shapes
from the corpus.

Nine characterization pins that recorded the old fail-soft as a named gap went
red as stale pins — including one whose docstring said *"it goes RED (as a
stale pin) the day it is closed"* — and now assert the refusal.

### Gate-off — 10 mutations, 0 survivors

| red | mutation |
|---|---|
| 19 | `divisibleby`: fail soft again instead of raising |
| 7 | `get_digit`: treat every `int()` failure as a ValueError |
| 2 | `add`: swallow the OverflowError like the other two |
| 4 | `filesizeformat`: drop the OverflowError refusal |
| 3 | `python_int_value`: stop classifying TypeError |
| 10 | `python_int_value`: call `int(nan)` an OverflowError |
| 12 | `python_int_str`: refuse the `_` separator again |
| 6 | `renderer::py_int`: go back to the saturating i64 cast |
| 3 | `ToF64`: refuse `_` in `float()` |
| 3 | `add`: restore the VALUE-side float coercion |

Baseline 0 failed / 1,586 passed; restored 0 failed / 1,586 passed. The harness
asserts the mutation text matched **exactly once**, that the mutated source
differs, that the `.so` **mtime advanced** on every rebuild, clears
`__pycache__` between rows, and counts `N error` apart from `N failed` so a
mutation that breaks the build reports `INVALID` rather than a number that
looks like evidence (v1.1.0-13 / v1.1.1-2 canon).

## Verification

- `pytest tests/ python/tests/ python/djust/tests/ -n auto` — **17,347 passed, 338 skipped, 0 failed**
- `make test-rust` (both CI phases) — 0 failures
- `make lint` — clean; `cargo fmt` applied
- `scripts/check-changelog-test-counts.py`, `scripts/check-doc-snippets.py` — clean
- Differential manifest — `0 MISSING` on every mechanical axis


johnrtipton:
## Empirical canary for the structural pin (#1459)

The gate-off table in the description proves the *behavioural* tests are
non-vacuous. The structural pin is a different kind of artifact and needs its
own proof, because an anti-drift pin that cannot go red is decorative (#1859) —
so it was run against two synthetic drifts applied to a copy of the source it
reads:

```
baseline:                                          PASSED (correctly)
A: a fifth arm grows its own int(value) read    -> CAUGHT — new ['wordcount'], gone []
B: get_digit loses its call                     -> CAUGHT — new [], gone ['get_digit']
restored:                                          PASSED
```

Drift B is the one that matters: a floor-shaped assertion (`⊇ the expected
set`) would pass it silently. It is a SET (#1125), so a deleted call site
reddens exactly as an added one does.

## The remaining bucket, for #2451

After this PR the `{% widthratio %}` bucket is 1,088, attributed on the same
run:

| filter | cells |
|---|---|
| `escapeseq` | 218 |
| `safeseq` | 218 |
| `unordered_list` | 218 |
| `first` | 216 |
| `last` | 216 |
| `json_script` | 2 (#2429) |

All iteration/subscript `TypeError`s — a different mechanism from `int(value)`,
which is why they are #2451 rather than in this diff (#1079).

Issue #2448, opened 2026-08-30
json_script spells a datetime and a timedelta VALUE with str(), not DjangoJSONEncoder's isoformat / ISO-8601 duration


johnrtipton:
Found while re-deriving #2429's divergent set with representative values. This
is the **emitting** direction — both engines render, and disagree on the bytes —
so it is a separate defect from #2429's refusal question and is filed rather
than folded in (CLAUDE.md #1079).

## The measurement

`{{ p|json_script:"d" }}`, live Django vs live djust:

| value | django | djust |
|---|---|---|
| `datetime(2020,1,1,3,4,5)` | `{"a": "2020-01-01T03:04:05"}` | `{"a": "2020-01-01 03:04:05"}` |
| `datetime(…, tzinfo=utc)` | `{"a": "2020-01-01T03:04:05Z"}` | `{"a": "2020-01-01 03:04:05+00:00"}` |
| `timedelta(seconds=90)` | `{"a": "P0DT00H01M30S"}` | `{"a": "0:01:30"}` |
| `timedelta(days=2, seconds=3)` | `{"a": "P2DT00H00M03S"}` | `{"a": "2 days, 0:00:03"}` |
| `date(2020,1,1)` | `{"a": "2020-01-01"}` | same ✓ |
| `time(3,4,5)` | `{"a": "03:04:05"}` | same ✓ |

## Why

`json_script` is `json.dumps(value, cls=DjangoJSONEncoder)`.
`DjangoJSONEncoder.default` is not `str()` — for a `datetime` it is
`o.isoformat()` with the microsecond field truncated to milliseconds and a
trailing `+00:00` rewritten to `Z`; for a `timedelta` it is
`django.utils.duration.duration_iso_string`. djust's `value_to_json` reaches
these as a `Value::String` already carrying Python's `str()`, which is the
**template display** spelling and not the encoder's.

`date` and `time` agree only because `str()` and `isoformat()` coincide for
them — the same coincidence-in-the-sampled-band shape as #2425's float keys.

## Scope

Client code parsing this JSON gets a string it cannot feed to `Date.parse`
(`"2020-01-01 03:04:05"` is not ISO-8601), so this is a live wire-format
divergence rather than a cosmetic one.

Note the Python type is erased at the PyO3 boundary for the *refusal* question
(#2429) but **not for this one**: a `datetime` reaches Rust as a
`Value::String` whose bytes are already wrong, so a fix belongs where the
[...]

Issue #2344, opened 2026-08-28
timesince/timeuntil ignore their argument entirely — Django uses it as the comparison instant


johnrtipton:
`timesince` and `timeuntil` take an argument in Django — the instant to compare
against:

```python
@register.filter("timesince", is_safe=False)
def timesince_filter(value, arg=None):
    ...
    return timesince(value, arg)
```

djust's arms read the VALUE and discard the argument entirely
(`crates/djust_templates/src/filters.rs`, the `"timesince"` / `"timeuntil"`
arms — `format_timesince(&datetime_str)` takes no comparison instant).

So `{{ then|timesince:other }}` silently answers "since now" whatever `other`
is, and `{{ then|timesince:"notanumber" }}` renders a duration where Django
raises `AttributeError` (the argument has no `tzinfo`).

## Why #2328 did not close it

#2328 made an unparseable or unresolvable filter ARGUMENT raise, at one
chokepoint, for every built-in that reads its argument as a number. These two
are the only 2 of the 29 argument-taking built-ins left disagreeing with Django
on the raise bit, and they are exempted deliberately: making an *unparseable*
argument raise while a *valid* one is still discarded would be a half-fix —
strictly worse than the current honest "the argument does nothing", because it
would look handled.

They are `RAISE_BIT_NOT_CLOSED` in
`python/tests/test_filter_argument_contract_2328.py`, with a test asserting they
are STILL divergent, so this issue closing will fail that test and force the
exemption to be removed rather than left to rot.

## What closing this means

Thread a comparison instant through `format_timesince` / `format_timeuntil`,
resolve the argument to a datetime, and raise when it is not one. Then delete
both rows from `RAISE_BIT_NOT_CLOSED` — the sweep will then cover all 29.

PR #2464, opened 2026-08-30
fix(templates): bool(timedelta(0)) is False, and the falsiness rule has one definition (#2458)

### LINKED ISSUE #2458: bool(timedelta(0)) is False in Python and True in djust — now fixable, since #2448 stopped erasing the type
Found while fixing #2448, and filed rather than folded in per CLAUDE.md #1079 —
a truthiness change is not something a JSON-spelling fix should make silently.

## The measurement

live Django vs live djust, unchanged by PR #2456:

| template | context | python | django | djust |
|---|---|---|---|---|
| `{% if p %}T{% else %}F{% endif %}` | `p = timedelta(0)` | `False` | `F` | `T` |
| `{{ p\|timesince:q }}` | `q = timedelta(0)` | — | measures from now | **raises** |

Every other member of the datetime family is truthy in Python for every value
(a midnight `time` has been truthy since 3.5), so `timedelta(0)` is the whole
of it.

## Why it is still open after #2448

Before #2448 a `timedelta` crossed the PyO3 boundary as
`Value::String("0:00:00")`, which is non-empty and therefore truthy — so the
divergence predates the fix. #2448 gave the family `Value::Encoded` and
deliberately kept `is_truthy` on the pre-fix answer
(`!e.display.is_empty()`, i.e. always true), because changing it is a separate
behaviour change.

The variant now makes the fix *possible*: the type is no longer erased. It is
not free, though — `Encoded` carries `type_name` / `display` / `json`, none of
which directly says "this delta is zero". Two shapes:

1. read it off the encoder spelling (`json == "P0DT00H00M00S"`), which is exact
   but reads a JSON string to answer a truthiness question;
2. carry the answer explicitly — a fourth field set at the conversion from
   `bool(o)`, which is Python's own answer for whatever the object is, subclass
   included.

(2) is preferable: it generalises to any future `Encoded` member and cannot go
wrong for a subclass with an overridden `__bool__`.

## The related `timesince` row, and why it needs (2) rather than a text fix

`filters.rs::timesince_arg_is_falsy` takes a `&str` and matches the DISPLAY
text. Teaching it `"0:00:00"` would also make the perfectly ordinary and
Python-**truthy** string `"0:00:00"` falsy — trading one divergence for
another. It needs a value-typed predicate, which is the same field (2) adds.

`python/tests/test_timesince_comparison_instant_2344.py::
test_every_display_arm_that_can_be_falsy_is_handled` carries `Encoded` in a
`FALSY_BUT_DELIBERATELY_NOT_ACCEPTED` set and asserts `"0:00:00"` is NOT in the
predicate, so a text-shaped fix reddens it rather than landing quietly.

Pinned as divergent in `python/tests/test_json_script_datetime_value_2448.py::
TestWhatThisDeliberatelyDoesNOTClose::
test_a_zero_timedelta_is_still_truthy_here_and_falsy_in_python`.

Refs #2448, #2344, PR #2456.

johnrtipton:
Closes #2458.

## The defect

`bool(timedelta(0))` is `False` in Python and Django, and was `True` here, in the plainest possible template:

```
{% if p %}T{% else %}F{% endif %}     p = timedelta(0)

python  False        django  F        djust  T
```

Every other member of the datetime family is truthy for every value — a midnight `time` has been truthy since 3.5 — so a zero `timedelta` is the whole of it.

It predates #2448: a `timedelta` crossed the PyO3 boundary as `Value::String("0:00:00")`, non-empty and therefore truthy. #2448's `Value::Encoded` deliberately kept that answer (`!display.is_empty()`, i.e. always true) rather than let a JSON-spelling fix change truthiness silently, and filed this.

## The fix, in two halves

**1. `Encoded` carries `bool(o)`.** A fourth field, set from Python's own answer at the conversion; `is_truthy` reads it.

Both available derivations are wrong, and `TestTheBitIsPythonsAndNotADerivation` runs the counterexamples rather than arguing them:

| derivation | fails on |
|---|---|
| encoder spelling `json == "P0DT00H00M00S"` | a NON-zero `timedelta` subclass whose `__bool__` returns `False` |
| display spelling `display == "0:00:00"` | the same, **and** it is the display text of the perfectly ordinary Python-**truthy** `str` `"0:00:00"` — one text, two answers |

The second is not a style objection: it is the reason a text fix was refused. #2448's own `test_every_display_arm_that_can_be_falsy_is_handled` parked `Encoded` in `FALSY_BUT_DELIBERATELY_NOT_ACCEPTED` with the message *"Move Encoded out … only with a value-typed predicate, not a text one."*

**2. The `timesince` argument — one definition of the falsiness rule (#1646).** `{{ p|timesince:q }}` with `q = timedelta(0)` measures from now in Django and **raised** here, because `timesince_arg_is_falsy(&str, bool)` recovered Python's `if not now:` from the argument's `Display` text — while `ArgType::is_falsy` (#2413), computed value-typed from the resolved `Value` two frames up, was already sitting there with the right answer. Two mechanisms for one question, at filter-argument scale. The copy is deleted; the filter reads the shared bit.

## The convergence dividend the issue did not predict

Deleting the text copy did not just make `timedelta(0)` work — it replaced every answer the text rule got wrong:

| `{{ p|timesince:q }}` | Django | djust before | djust now |
|---|---|---|---|
| `q = timedelta(0)` | measures from now | **raises** | measures from now |
| `q = "0"` / `"None"` / `"False"` / `"0.0"` / `"[]"` / `"{}"` (a real `str`) | **raises** (`str` is truthy, no `.year`) | measures from now | raises |
| `q = []` under `legacy_display` (renders every sequence as `[List]`) | measures from now | **raises** | measures from now |

`TestTheFalsinessResidueIsNamed`'s four-row residue is down to one, and the survivor is genuinely about the wire format rather than truthiness: a date-**shaped** `str` is still read as the datetime it spells, because a Python `datetime` crosses into Rust as a string and has no other spelling. That row is kept, re-worded, and its three former neighbours are inverted to agreement tests rather than deleted.

## The state round trip

`SerializableViewState.state` round-trips through msgpack on **every** read of the default `InMemoryStateBackend`, so an untagged answer flips back after one cache hit — the `Decimal` version of exactly this shipped once and was caught only by a gate-off (#2135). `ENCODED_TAG`'s payload goes `[type_name, display, json]` → `[type_name, display, json, truthy]`.

The **three**-element form is still read: a Redis backend hands one back on the first request after a rolling deploy, so it is a live input rather than a hypothetical. It restores to the truthiness that entry was *written* with (`!display.is_empty()`) — the honest answer, since the value's `bool()` was never recorded. `test_the_payload_is_what_carries_it` decodes the real blob and asserts the fourth element, so the field cannot be dropped with the round-trip test still green.

## Test value set — crossed, not sampled

`python/tests/test_encoded_truthiness_2458.py` (new, 95 cases) uses the **cross** of every axis the family has, with `bool()` computed per row rather than transcribed:

* type: `datetime` / `date` / `time` / `timedelta`
* microsecond: zero and non-zero
* tzinfo: naive, UTC, `+05:30`, `-08:00`
* duration: zero, positive, negative, sub-second, multi-day, `timedelta.max`

`test_the_family_membership_of_FALSY_is_computed_not_transcribed` is the non-vacuity half: exactly one row of the sweep is falsy, so a fix answering "falsy" for the whole variant fails rather than passing everything.

## Gate-off

Four mutations, each asserting the mutation text matched **exactly once**, that the source changed, that the crate **rebuilt** with an advanced `.so` mtime, with `__pycache__` cleared and `N error` counted apart from `N failed`. Re-run against the current `origin/main` base after #2456/#2461 landed:

| mutation | reddens |
|---|---|
| `is_truthy` reads `!display.is_empty()` again | 9 |
| the conversion hardcodes `truthy` instead of asking Python | 10 |
| the msgpack payload drops the fourth element | 3 |
| `timesince` stops reading the threaded bit | 30 |

No survivors, no build breaks, no invalid runs. Each mechanism is independently reachable — the msgpack arm in particular, which is the one that would otherwise be shadowed by the in-process path.

## Deliberately not closed

A `set` has no `Value` variant at all, so it lands on the conversion's final `Value::String(str(o))` and arrives as the non-empty `"set()"`: `bool(set())` is `False` and `{% if q %}` renders `T`. Same family, one level up — at the **conversion** rather than in the truthiness rule — and out of a datetime fix's scope (#1079). Pinned as still-divergent in `TestWhatThisDeliberatelyDoesNOTClose` rather than left as a gap, and filed separately.

## Verification

* Full suite, all three roots: **17839 passed, 338 skipped, 0 failed**.
* `cargo test -p djust_core -p djust_templates`: 34 binaries, 0 failures. `cargo clippy` clean, `cargo fmt` applied.
* Two-build Django differential: reported in a follow-up comment.


johnrtipton:
## Two-build Django differential

Baseline `origin/main` (`2a20ae74`, which already contains #2456 and #2461) → this branch. Both built `make build` (release, 6,597,760 B), so the `.so` is the shipped profile on both sides.

```
cells        : 353909
agree BEFORE : 257355   (refusal-collapsed: 298606)
agree AFTER  : 257355   (refusal-collapsed: 298606)
django REFUSES & djust RENDERS: 6106 -> 6106   (+0)
djust REFUSES & Django RENDERS: 38105 -> 38105  (+0)

newly AGREEING     : 0   (refusal-collapsed: 0)
no longer agreeing : 0
REGRESSIONS        : 0
panics             : 0 -> 0   (introduced 0)
live-payload leaks  : 22 -> 22 (introduced 0)

builds       : adac3068f0802d1d -> 58bbcde480a5a5ff  (genuinely two builds)
reachability manifest: 0 MISSING on all 16 mechanical axes
OK: no agreeing cell regressed, nothing became more permissive, and no cell newly panics.
```

**`djust REFUSES & Django RENDERS` is flat at 38105**, which is the number that matters for a fix that makes a filter *stop* raising: nothing became stricter. `django REFUSES & djust RENDERS` is flat too, so nothing became more permissive either.

### The 0-moved reading, which is a question rather than a pass

The tool prints its own warning for this, and it is correct here: *"Either the change moves nothing, or it moves an axis this corpus does not sweep."* It is the second, and it is provable in two greps rather than assertable.

**1. The corpus contains no `timedelta`:**

```
$ grep -c "timedelta(" scripts/filter-parity-differential.py
0
```

So it cannot construct one cell where `bool(timedelta(0))` is the question — not in the value position, not in the argument position.

**2. Every bound argument is Python-TRUTHY:**

```python
ARG_CONTEXT = {"known": "3", "_x": 9, "known_list": [1, 2],
               "known_tuple": (1, 2), "known_dict": {"a": 1},
               "known_dt": datetime.datetime(2020, 1, 1, 15, 30)}
```

Six bindings, six truthy objects — so the corpus cannot build a cell where the argument's *falsiness* is the question, which is the entire axis this PR's second half moves. `ARG_SPELLINGS` does carry `'"0"'`, `'""'`, `"False"`, `"None"`, but those are *literal* spellings; `arg_is_falsy`'s first arm answers the **resolved-context** channel, and there is nothing falsy to resolve to.

The four behaviour changes this PR makes are therefore each unconstructible here:

| | Django | before | after |
|---|---|---|---|
| `{% if p %}`, `p = timedelta(0)` | `F` | `T` | `F` |
| `{{ p\|timesince:q }}`, `q = timedelta(0)` | measures from now | raises | measures from now |
| `q = "0"` / `"None"` / `"False"` (a resolved `str`) | raises | measures from now | raises |
| `q = []` under `legacy_display` | measures from now | raises | measures from now |

This is the sixth recorded instance of the class — the file documents five in its own docstrings (#2328's 1,601 argument cells, #2344's 336, #2366's `known_dt`, #2317's `t-nested`, #2324's non-string items), each added by the fix that found it. Filed as #2469 with the suggested rows rather than added here: changing the corpus changes the cell count and invalidates every baseline captured by the other PRs in flight during this drain.

The behaviour is covered instead by 95 cases in `python/tests/test_encoded_truthiness_2458.py` run against **live Django**, plus the four gate-off mutations (9 / 10 / 3 / 30 reddened).

Issue #2466, opened 2026-08-30
bool(set()) is False in Python and True in djust — the #2458 shape at the conversion, for every type with no Value variant


johnrtipton:
Found while fixing #2458 and filed rather than folded in (#1079): it is the same
family one level up, at the CONVERSION rather than in the truthiness rule.

## The measurement

live Django vs live djust, unchanged by PR #2464:

| template | context | python | django | djust |
|---|---|---|---|---|
| `{% if p %}T{% else %}F{% endif %}` | `p = set()` | `False` | `F` | `T` |
| `{% if p %}T{% else %}F{% endif %}` | `p = frozenset()` | `False` | `F` | `T` |
| `{{ p }}` | `p = set()` | — | `set()` | `set()` |

## Why

A `set` has no `Value` variant at all, so `FromPyObject for Value` lands it on
its final `Ok(Value::String(ob.str()?))` fallback and it arrives as the
non-empty string `"set()"`. `Value::String`'s `is_truthy` is
`!s.is_empty()`, so every empty container that has no variant is truthy here.

This is exactly the shape #2458 closed for `timedelta(0)` — a Python-falsy
object arriving as a non-empty display string — and #2458's fix does not reach
it, because that fix gave `Value::Encoded` a `truthy` field and a `set` never
becomes an `Encoded`.

## Scope

`Value` has variants for `List`, `Tuple`, `Object`, `DictView`, `Decimal`,
`BigInt`, `String`, `Bool`, `Integer`, `Float`, `None`, `Missing`, `Encoded`.
Anything else Python can put in a context reaches the `str()` fallback, so the
question generalises: **which other Python-falsy objects have no variant?**
Worth enumerating rather than fixing `set` alone — a one-type fix here would be
the shape #2129 took five rounds over.

Two candidate shapes, in the order #2458's own reasoning prefers:

1. give `set`/`frozenset` a variant (they already normalise to a sorted list in
   `normalize_django_value`, so the LiveView path is unaffected and only the
   raw `DjustTemplateBackend` path is at issue);
2. carry `bool(o)` on the `str()` fallback the way `Encoded` now carries it —
   which answers the whole class rather than one member, and is the same
   argument #2458 made for carrying the bit rather than deriving it.

## Pinned

`python/tests/test_encoded_truthiness_2458.py::TestWhatThisDeliberatelyDoesNOTClose::test_a_set_is_still_truthy_because_it_never_becomes_an_Encoded`
[...]

Issue #2459, opened 2026-08-30
get_digit returns a one-character STRING where Django returns an int — 15 of the 17 cells #2451 leaves behind


johnrtipton:
Split out of #2451 per CLAUDE.md #1079 — found while closing that issue's iteration/subscript column, measured on that branch, and **not** touched there.

## The divergence

Django's `get_digit` ends `return int(str(value)[-arg])` — an **`int`**. djust's arm answers a one-character **string**:

```
{{ p|get_digit:"1"|pprint }}   p = 42   django  2
                                        djust   '2'
```

The type is directly visible through `pprint`, and it is *consumable* by every filter that iterates or subscripts — because a `str` does both and an `int` does neither.

## What it costs, measured

On the two-build differential (`scripts/filter-parity-differential.py`, 353,909 cells) run over #2451's fix, **15 of the 17 surviving `django-refuses / djust-renders` cells are this**, and every one of them names a sequence filter that is now correct:

```
'get_digit:"1"|safeseq'          <i-int>   django <<TypeError: 'int' object is not iterable>>       djust ['2']
'get_digit:"1"|escapeseq'        <i-int>   django <<TypeError: 'int' object is not iterable>>       djust ['2']
'get_digit:"1"|unordered_list'   <i-int>   django <<TypeError: 'int' object is not iterable>>       djust <li>2</li>
'get_digit:"1"|first'            <i-int>   django <<TypeError: 'int' object is not subscriptable>>  djust 2
'get_digit:"1"|last'             <i-int>   django <<TypeError: 'int' object is not subscriptable>>  djust 2
```

— five chains, each in the bare `{{ }}` position and in both `{% widthratio %}` shapes.

**The divergence is in the SUBJECT's type, not in the consumer.** #2451's chokepoint is right about all five: an `int` is not iterable and not subscriptable, and it refuses when it is given one. It is given a `str`.

## Where to look

* `crates/djust_templates/src/filters.rs`, the `"get_digit"` arm — its success exit is `Value::String((*b as char).to_string())`. Django's is `int(...)`, so the exit should be an integer `Value`.
* `int_value_of` already exists in that file and is what the `n < 1` early-return uses — the same helper is likely the right exit for the digit itself.
* The out-of-range case answers `0` in Django (`except IndexError: return 0`), which is an `int` too.
* The `-` case is a **documented** existing divergence recorded in the arm's own comment (Django raises on `int('-')`, djust returns the value unchanged) and is a separate question from the return TYPE.

[...]

PR #2470, opened 2026-08-30
fix(templates): {% regroup %} refuses a non-iterable source, as {% for %} does (#2463)

### LINKED ISSUE #2463: {% regroup %} fails soft where Django raises 'int' object is not iterable — the {% for %} arm's twin, unfixed
Split out of #2459 per CLAUDE.md #1079 — found while re-deriving that issue's cell counts, measured on its branch, and **not** touched there.

## The divergence

`{% regroup %}` over a value Python cannot iterate fails soft to an empty grouping. `{% for %}`, which asks the same question one tag over, refuses correctly:

```
{% regroup p by k as g %}[{{ g|length }}]     p = 2     django  <<TypeError: 'int' object is not iterable>>
                                                       djust   [0]
{% for x in p %}[{{ x }}]{% endfor %}         p = 2     django  <<TypeError: 'int' object is not iterable>>
                                                       djust   <<TypeError: 'int' object is not iterable>>   ✓
```

No filter is involved. `{% regroup p by k %}` over a bare `int` in the context already diverges.

Django's `RegroupNode.render` is:

```python
obj_list = self.target.resolve(context, ignore_failures=True)
if obj_list is None:
    context[self.var_name] = []
    return ""
context[self.var_name] = [
    {"grouper": key, "list": list(val)}
    for key, val in groupby(obj_list, lambda obj: self.expression.resolve(...))
]
```

`groupby` calls `iter(obj_list)` with no guard, so every non-iterable that is not `None` raises. `None` is the one value the `if obj_list is None` arm catches — and djust gets that one right (`[0]` on both engines).

## Why this is the #1646 shape

#2451 gave the crate a `python_iter` probe and a `ValueOpError` chokepoint, and wired the type-named refusal into `renderer.rs`'s `{% for %}` arm — which is why the `for` row above agrees. The `regroup` arm kept its own older fail-soft and was never repointed. Two implementations of "can this value be iterated?", one fixed and one not.

## Scale, measured

On the two-build differential (`scripts/filter-parity-differential.py`, 353,909 cells), the `regroup` and `regroup-by` axes carry **141** `django-refuses / djust-renders` cells whose cid happens to name `get_digit` alone — every one of them this defect rather than #2459's, since #2459's fix leaves all 141 exactly where they were. The full `regroup` bucket is larger; it has not been separated from the other tag axes yet.

## Where to look

* `crates/djust_templates/src/renderer.rs`, the `{% regroup %}` node — it should read `python_iter` and raise through the same path the `{% for %}` arm now takes, keeping the `None → []` arm that Django's own `if obj_list is None` gives it.
* The `for` arm is the reference implementation to lift (#1077).

## Pinned

`python/tests/test_get_digit_returns_an_int_2459.py::TestTheTagOperandPositions::test_regroup_is_a_SEPARATE_pre_existing_divergence` asserts the divergence in the DIVERGING direction and names this issue, so it goes red the day `regroup` refuses.

Refs #2459, #2451, #1646, #1079.

johnrtipton:
Closes #2463.

`{% regroup p by k as g %}` over a value Python cannot iterate failed soft to an empty grouping. `{% for x in p %}` — the same question one tag over — refused correctly. One invariant, two implementations, one fixed: the #1646 shape.

```
{% regroup p by k as g %}[{{ g|length }}]   p = 2   django  TypeError: 'int' object is not iterable
                                                    djust   [0]                        <- before
                                                    djust   TypeError                  <- after
{% for x in p %}[{{ x }}]{% endfor %}       p = 2   django  TypeError
                                                    djust   TypeError                  <- already right
```

## The issue's cited location does not exist

#2463 says to look at *"`crates/djust_templates/src/renderer.rs`, the `{% regroup %}` node"*. There is no `{% regroup %}` node. `regroup` is a **Python** assign-tag handler (`djust.template_tags.regroup`) dispatched from Rust, and the fail-soft was `RegroupTagHandler._decode_source`'s `except TypeError: return []`. Tracing symptom-up from the rendered `[0]` rather than down from the citation is what found it.

## The fix deletes an answer; it does not add one

The handler holds the real Python object, so the sink Rust's `python_iter` *models* is directly available to it — `list(decoded)` — and its message is CPython's verbatim rather than a reconstruction. Only `None` still answers "no groups", which is Django's own single guard:

```python
obj_list = self.target.resolve(context, ignore_failures=True)
if obj_list is None:            # <- the ONE guard; covers the unresolved operand too
    context[self.var_name] = []
    return ""
context[self.var_name] = [... for key, val in groupby(obj_list, ...)]   # iter(), unguarded
```

## A second half the issue does not mention

After the swallow was gone a `bool` source *still* answered `[0]`. `value_channel_arg_string` handed the handler Python's `True`, which is not valid JSON, so `json.loads` raised and the handler took its "this must be an unresolved bare name" branch. `42` and `1.5` **are** valid JSON and were already refused correctly — the doc comment claiming *"every other scalar's `Display` form (`42`, `True`, `None`, `1.5`) is unambiguous against a bare name"* was false of exactly one of the four. A `Bool` is now encoded as JSON `true`/`false`, the same type-tag argument #2385 made for the `String` arm.

`None` keeps its `Display` spelling and is documented rather than changed: its mis-decode is harmless, because the fallback lookup answers `None`, which is exactly what Django's guard wants (#1079).

## Every other site that asks "is this iterable?", enumerated and decided

| site | decision |
|---|---|
| `djust/template_tags/regroup.py::_decode_source` | **fixed** — the swallow deleted; defers to Python's `list()` |
| the rest of `djust/template_tags/` | nothing else asks; pinned EMPTY by `test_no_tag_handler_swallows_an_iteration_TypeError`, so a new handler growing the same fail-soft reddens the day it lands |
| `filters::iter_values` (Rust) | unchanged — the sink |
| `filters::python_iter` (Rust) | unchanged — names the sink's `None` for the refusing filters (#2451) |
| `renderer.rs` `{% for %}` arm | unchanged — the reference implementation, lifted rather than re-derived (#1077) |
| `value_channel_arg_string` | **fixed** — the `Bool` type tag, above |

Exactly **two** production sites emit `'X' object is not iterable`, and `test_the_crate_has_exactly_two_not_iterable_emitters` pins that set by **equality** on the extracted `(file, line-text)` pairs — so a third copy *and* a removed arm both redden it, which a floor-based count structurally cannot do. It also asserts `len(found) == 2` before comparing, so a stale path that extracts nothing cannot pass as agreement.

## What changes for you

A template regrouping a value Django cannot iterate now raises instead of rendering an empty region. These are exactly the templates Django has always refused. The direction that would break working templates — refusing where Django renders — is checked across the whole iterable axis, `""` / `[]` / `{}` included, which iterate to nothing rather than refusing.

Known and unchanged: a bare `object()` still renders one group, because it never reaches the handler as an object — `Value`'s conversion has no variant for it and it arrives as `Value::String(str(o))`. That is the #2466 conversion gap one tag over; guarding it inside `regroup` would be a fix at the consumer for a defect at the source. Pinned in the diverging direction in `TestWhatThisDeliberatelyDoesNOTClose`, with the `{% for %}` evidence that both tags agree with each other and disagree with Django — the signature of a defect below them.

## Tests

72 cases in `python/tests/test_regroup_non_iterable_2463.py`:

* `TestTheForArmIsTheReferenceImplementation` — the #1646 assertion, written as a **verdict identity** across the whole value axis rather than as a list of values, for the reason #2459's identity pin exists: a hand-written list of affected shapes is one short by construction;
* `TestTheNoneArmIsIntact` — explicit `None`, an unresolved bare name, a missing dotted path;
* `TestEveryIterableSourceIsUnchanged` — the empty containers on purpose: a fix that deleted one line too many takes them with it;
* `TestABoolSourceReachesTheHandlerAsAValue` — measured at the boundary by spying on the **live** handler class off `_registered_handlers`, not a module-level import (#2427);
* `TestNoSecondIterabilityCheckWasAdded` — the equality pins above;
* `TestWhatThisDeliberatelyDoesNOTClose` — `object()` and the `Decimal`-names-`float` boundary.

Three pins that asserted the old fail-soft are **flipped in place** rather than deleted, both of which said in as many words that they would redden the day `regroup` refused: `test_regroup_is_a_SEPARATE_pre_existing_divergence` (#2459) and `TestTheDivergenceThatIsNotClosedHere` (#2385/#2394). One Rust unit test (`value_channel_quotes_only_strings`) is renamed and extended for the `Bool` arm.

## Gate-off

Three mutations, one per mechanism, each asserting the text was found and matched **exactly once**, that the source changed, that the crate **rebuilt** (`.so` mtime advanced), with `__pycache__` cleared and `N error` counted apart from `N failed`:

| mutation | tests reddened |
|---|---|
| restore `except TypeError: return []` | **43 failed** |
| drop the `if decoded is None` guard | **12 failed** |
| revert the `Value::Bool` JSON encoding | **9 failed** |

No survivors; restored baseline `0 failed`. Each mechanism is independently reachable — the `None` guard and the `Bool` tag are not shadowed by the swallow's removal.


johnrtipton:
## Two-build differential — 353,909 cells, `7b3da5ed` → `f7b84485`

Measured against an `origin/main` baseline built in the same worktree (the tool confirms two genuinely different builds).

```
cells        : 353909
agree BEFORE : 257394   (refusal-collapsed: 298742)
agree AFTER  : 257394   (refusal-collapsed: 301362)
django REFUSES & djust RENDERS: 6009 -> 3389   (+2620; djust more permissive than Django)
djust REFUSES & Django RENDERS: 38105 -> 38105   (+0; djust stricter than Django)
```

**The raw headline is 257,394 → 257,394 — literally unchanged — and reading it would say this PR does nothing.** That is structural, not a coincidence: this is a refusal-class fix, both engines raise, and `TypeError: 'int' object is not iterable` never string-equals `RuntimeError: Template error: …`. The two numbers to read are the collapsed agreement (**+2,620**) and the permissive class (**6,009 → 3,389**, 44% of the remaining gap closed by one tag).

`djust REFUSES & Django RENDERS` is **38,105 → 38,105, +0** — nothing became over-strict, which is the direction this change could have broken working templates in.

**Movement is confined to one axis**, which is what a tag-local fix should look like:

```
  tag          7008 moved  of 276876
  argument / arity / builtin / builtin x mapping / chain / cmp / ctag /
  custom / filter / masked-refusal / path / variable-name    0 moved
```

Zero regressions, zero coincidental cells, zero newly-panicking cells, and the live-payload-leak count is unchanged at 22 (0 closed, 0 introduced). Reachability manifest: 0 MISSING on every mechanical axis, `value-op` included.

### One issue premise re-derived

#2463 reports *"**141** `django-refuses / djust-renders` cells whose cid happens to name `get_digit` alone"*. That is a sub-slice picked out by one filter name; measured across the whole corpus the fix moves **2,620** cells out of that class and changes djust's output on **7,008**. The 141 is not wrong, it is just not the size of the defect — the full `regroup` bucket the issue says "has not been separated from the other tag axes yet" is what the axis line above separates.

### Suite

* Full suite, all three roots: **18,447 passed, 420 skipped, 0 failed**.
* `cargo test -p djust_templates`: 494 lib tests + integration, all green.
* Gate-off, re-run from scratch: **43 / 12 / 9** tests reddened, no survivors, restored baseline `0 failed`.

Issue #2473, opened 2026-08-30
{{ p|get_digit:"1" }} echoes a datetime where Django raises TypeError — the value-position twin of #2366


johnrtipton:
Found by the widened differential corpus (#2469). Third of the three
`Value::Encoded` value-position holes that a `timedelta` in `INPUTS` made
reachable (#2471 comparison, #2472 `pprint`).

#2366 established the rule for the ARGUMENT position — `int(datetime)` is a
`TypeError`, Django's `except ValueError` does not catch it, so the filter
raises — and #2448 gave djust the type it needs to see that. `get_digit` asks
the same question of its **value** and the answer did not follow.

## The measurement

`main` at `7b3da5ededeea374`, raw `render_template` path:

| template | `p` | Django | djust |
|---|---|---|---|
| `{{ p\|get_digit:"1" }}` | `timedelta(seconds=90)` | `TypeError: int() argument must be a string, a bytes-like object or a real number, not 'datetime.timedelta'` | `0:01:30` |
| `{{ p\|get_digit:"1" }}` | `timedelta(0)` | same `TypeError` | `0:00:00` |
| `{{ p\|get_digit:"1" }}` | `datetime(2020,1,1,3,4,5)` | same, `'datetime.datetime'` | `2020-01-01 03:04:05` |
| `{{ p\|get_digit:"1" }}` | `date(2020,1,1)` | same, `'datetime.date'` | `2020-01-01` |
| `{{ p\|get_digit:"1" }}` | `time(3,4,5)` | same, `'datetime.time'` | `03:04:05` |

Django's body is

```python
try:
    if int(digit) < 1:
        return value
    ...
except ValueError:
    return value
```

so the `TypeError` from `int(value)` escapes uncaught. djust takes its
return-the-input branch instead and puts the datetime's display spelling on the
page — the echo-on-failure class #2359 is about, and `get_digit`'s echo arm
already carries a per-call safety grant (#2403), so the value reaches the page
live if anything downstream marks it safe.

Same divergence through a tag operand, where it is louder because Django
refuses at COMPILE time:

```
{% widthratio p|get_digit:"1" 10 100 %}
```

Django: `TemplateSyntaxError: widthratio final argument must be a number`.
djust: renders empty.

## Why the corpus could not see it

`INPUTS` had no `timedelta` and no datetime object at all — every date-ish
entry was `s-datetime`, a **string**, which `int()` refuses with a
`ValueError`, the one exception Django's body catches. So every `get_digit`
[...]

Issue #2472, opened 2026-08-30
{{ p|pprint }} over a datetime spells str(o) where Django spells repr(o) — 'datetime.timedelta(0)' becomes "'0:00:00'"


johnrtipton:
Found by the widened differential corpus (#2469), which put a `timedelta` into
`INPUTS` and so built the first `{{ p|pprint }}` cell over a `Value::Encoded`.
Same family as #2471, one `match` over.

## The measurement

`main` at `7b3da5ededeea374`, raw `render_template` path (unescaped for
readability; djust's column is HTML-escaped in the actual output, which is the
tell — it is being spelled as a **string**):

| `p` | Django | djust |
|---|---|---|
| `timedelta(seconds=90)` | `datetime.timedelta(seconds=90)` | `'0:01:30'` |
| `timedelta(0)` | `datetime.timedelta(0)` | `'0:00:00'` |
| `datetime(2020,1,1,3,4,5)` | `datetime.datetime(2020, 1, 1, 3, 4, 5)` | `'2020-01-01 03:04:05'` |
| `date(2020,1,1)` | `datetime.date(2020, 1, 1)` | `'2020-01-01'` |
| `time(3,4,5)` | `datetime.time(3, 4, 5)` | `'03:04:05'` |

It is the same for a NESTED value, which is the position that matters more —
`{{ p|pprint }}` over `[timedelta(0)]` is `[datetime.timedelta(0)]` on Django
and `['0:00:00']` here. A container's repr calls `repr` on each element, so
this is the only spelling a nested datetime has.

Downstream, it changes what a loop iterates: `{% for x in p|pprint %}[{{ x }}]{% endfor %}`
over `timedelta(0)` walks the 21 characters of `datetime.timedelta(0)` on
Django and the 9 characters of `'0:00:00'` here.

## Where (a hypothesis — trace it, do not trust it)

`Value::py_repr` (`crates/djust_core/src/lib.rs:1235`) enumerates `String`,
`Decimal` and `Float` and then delegates `other => other.to_string()`. For an
`Encoded` that yields `e.display`, which is `str(o)` — correct for `{{ p }}`
and wrong for `repr`. The quoting in the observed output means something on the
`pprint` path is additionally treating the value as a string, so the real site
may be in the filter rather than in `py_repr`; the reported location is the
hypothesis, the table above is the evidence.

The `Decimal` arm two lines up is the precedent and the argument: `str` and
`repr` differ for that type too, and the fix there was an explicit arm rather
[...]

Issue #2471, opened 2026-08-30
Two Value::Encoded values never compare equal — {% if a == b %} is False for a datetime against itself, on every operator


johnrtipton:
Found by the widened differential corpus (#2469), which put a `timedelta` into
`INPUTS` and so built the first `@cmp` cell in the tool's history whose operands
are a `Value::Encoded`. It reports six divergences per pair, and the class is
wider than the row that surfaced it: it is **every member of the datetime
family**, in **every comparison operator**.

## The measurement

Run against `main` at `7b3da5ededeea374`, through `render_both` (the raw
`render_template` path, so `Value::Encoded` is genuinely constructed):

| template | `p`, `q` | Django | djust |
|---|---|---|---|
| `{% if p == q %}` | the SAME `timedelta(seconds=90)` | `Y` | **`N`** |
| `{% if p == q %}` | the SAME `datetime(2020,1,1,3,4,5)` | `Y` | **`N`** |
| `{% if p == q %}` | the SAME `date(2020,1,1)` | `Y` | **`N`** |
| `{% if p == q %}` | the SAME `time(3,4,5)` | `Y` | **`N`** |
| `{% if p != q %}` | any of the four, against itself | `N` | **`Y`** |
| `{% if p <= q %}` | any of the four, against itself | `Y` | **`N`** |
| `{% if p >= q %}` | any of the four, against itself | `Y` | **`N`** |
| `{% if p < q %}` | `timedelta(0)`, `timedelta(seconds=90)` | `Y` | **`N`** |
| `{% if p > q %}` | `datetime(2020,1,2)`, `datetime(2020,1,1)` | `Y` | **`N`** |

Control, on the same build: two equal `str`s and two equal `int`s both answer
`Y` on both engines, so this is not a broken `{% if %}` — it is one variant
falling off the end of two `match`es.

## Where

`crates/djust_templates/src/renderer.rs:4024` `values_equal` enumerates
`Missing|None`, `Bool`, `Integer`, `Float`, the mixed int/float pair, `String`,
and same-kind sequences, then `_ => false`. There is no `Encoded` arm, so two
`Encoded`s are never equal — **not even to themselves**. `try_compare`
(`renderer.rs:4245`) has the same hole for the ordering operators, and `!=`
inherits it as the negation of an answer that is always `false`.

This is exactly the shape #2335 fixed for lists, and the comment that fix left
behind says it in as many words: *"without this arm two lists fell to
`_ => false` and were never equal — not even to themselves — so
[...]

PR #2474, opened 2026-08-30
fix(tests): the differential can build a FALSY argument, and holds a timedelta (#2469)

### LINKED ISSUE #2469: The filter differential can build no cell where an argument is FALSY, and contains no timedelta at all — #2458 measured 0 moved on every axis
Found running the two-build differential for PR #2464 (#2458), which reported
**0 moved cells on every axis** for a change that demonstrably fixes four
behaviours. Two greps say why, and both are one-line facts about the corpus
rather than about the fix.

## 1. The corpus contains no `timedelta`

```
$ grep -c "timedelta(" scripts/filter-parity-differential.py
0
```

So it cannot build a single cell where `bool(timedelta(0))` is the question —
not in the value position (`{% if p %}`), not in the argument position
(`{{ p|timesince:q }}`). `timedelta` is one of the four members of the datetime
family `Value::Encoded` carries (#2448), and the only one with a falsy
inhabitant.

## 2. Every bound argument in `ARG_CONTEXT` is Python-TRUTHY

```python
ARG_CONTEXT = {
    "known": "3",
    "_x": 9,
    "known_list": [1, 2],
    "known_tuple": (1, 2),
    "known_dict": {"a": 1},
    "known_dt": datetime.datetime(2020, 1, 1, 15, 30),
}
```

Six bindings, six truthy objects. So the corpus cannot build a cell where the
argument's **falsiness** is the question — which is the whole axis #2458's
second half moves. `ARG_SPELLINGS` has `'"0"'`, `'""'`, `"False"` and `"None"`,
but those are *literal* spellings; the resolved-context channel is the one
`arg_is_falsy`'s first arm answers, and it has no falsy inhabitant to resolve
to.

## What that hid

PR #2464 replaces `timesince_arg_is_falsy` (text-shaped) with the value-typed
`ArgType::is_falsy` (#2413). Four measured behaviour changes, none of them
constructible from this corpus:

| `{{ p\|timesince:q }}` | Django | before | after |
|---|---|---|---|
| `q = timedelta(0)` | measures from now | raises | measures from now |
| `q = "0"` / `"None"` / `"False"` (a resolved `str`) | raises | measures from now | raises |
| `q = []` under `legacy_display` | measures from now | raises | measures from now |
| `{% if p %}`, `p = timedelta(0)` | `F` | `T` | `F` |

Row 2 is the sharpest: a resolved argument holding the **string** `"0"` is
truthy in Python and has no `.year`, so Django raises — and djust read it as
the number it spells. That is an `argument`-axis cell of exactly the shape the
axis exists for, and the axis reported `0 moved of 13920`.

## Suggested rows

Following the precedent every prior blind spot set in this file — `s-datetime`
(#2344), `t-nested` (#2317), `known_dt` (#2366), `"99999999999999999999"`
(#2328), each added by the fix that found the gap:

* `ARG_CONTEXT`: a falsy binding per shape — `known_empty: ""`,
  `known_zero: 0`, `known_empty_list: []`, `known_td_zero: timedelta(0)` — plus
  `known_str_zero: "0"`, which is the truthy-string-that-spells-a-falsy-object
  counter-example and the one row that distinguishes a value-typed rule from a
  text one;
* the VALUE corpus: a `timedelta`, zero and non-zero. It is the one member of
  the `Value::Encoded` family with a falsy inhabitant, and #2458 is not the
  last question that will turn on it.

Not done in PR #2464 itself: adding rows changes the cell count and invalidates
every baseline captured by the other PRs in flight during this drain. Worth
landing as its own change, when no drain is mid-run.

## Why this keeps happening

This is the sixth recorded instance, and the file documents five of them in its
own docstrings — #2328 moved 1,601 argument cells with the corpus building
none; #2344 moved 336 with the raw-text comparison reporting zero; #2366's
`known_dt` row; #2317's `t-nested`; #2324's non-string items. The reachability
manifest catches a missing ERROR, and `input-shape` is explicitly marked
`UNVERIFIED` because nothing mechanically names which VALUE shapes matter.

A cheap partial mechanisation, in the manifest's spirit: assert that
`ARG_CONTEXT` and the value corpus each contain at least one Python-falsy and
one Python-truthy inhabitant **of every `Value` variant** — the variant list is
already extracted mechanically by
`python/tests/test_timesince_comparison_instant_2344.py`. That would have
reported this gap without anyone noticing it.

Refs #2458, #2448, #2344, #2366, PR #2464.

johnrtipton:
Closes #2469.

`ARG_CONTEXT` bound six objects and every one was Python-TRUTHY, so
`scripts/filter-parity-differential.py` could construct no cell where a
*resolved* argument's falsiness is the question — which is the whole of what
`ArgType::is_falsy`'s first arm answers. And no entry of `INPUTS` was a
`timedelta`: the one member of the `Value::Encoded` family with a falsy
inhabitant, and the only way to reach that variant from the value corpus at
all. So #2458, whose entire subject is `bool(timedelta(0))`, reported **0 moved
cells on every axis** while changing four measured behaviours.

## The empirical canary (#1459), run backwards

The claim is "this corpus can now see a falsy-argument / `timedelta` change".
The way to show it is to take a change of exactly that class and measure it
under both corpora. So: two genuinely different builds — `main`, and one with
#2458's `Value::Encoded(e) => e.truthy` reverted to its pre-fix
`!e.display.is_empty()` — compared with the tool's own `--compare`.

| | cells | moved | REGRESSIONS | exit |
|---|---|---|---|---|
| pre-#2469 corpus | 353,909 | **0 on every axis** | **0** | **0** |
| this corpus | 371,452 | 43 | **38** | **1** |

Same two builds (`7b3da5ededeea374` → `5d3f1f7c35c5ba16`), same real
regression. The old corpus prints `NOTE: identical agreement counts under BOTH
definitions` and exits 0; the new one names 38 cells and exits 1. All 38 are a
`known_td_zero` argument or a `td-zero` value reaching `json_script` (16),
`default` (7), `yesno` (6), `slice` (3), `join` (3) or `default_if_none` (3).

That is #2454's failure mode — a gate reporting clean over a genuine regression
— reproduced against a real one and then closed. The old-corpus column is
derived by restricting both measured files to the pre-#2469 cell set, which is
sound because the djust side of all 353,909 shared cells is byte-identical
between the two corpora (checked; see below).

The Rust probe was reverted and rebuilt before committing; the working `.so`
hashes back to `7b3da5ededeea374`, and `crates/` is untouched by this PR.

## The three checks that go with it

1. **Stable or growing, never shrinking.** Per-axis cell counts:
   `filter` 2,337→2,736 · `chain` 37,120→37,120 · `custom` 7,076→7,104 ·
   `tag` 276,876→281,664 · `path` 2,624→3,072 · `cmp` 11,767→16,128 ·
   `argument` 13,920→21,344 · `ctag` 492→576 · `builtin x mapping` 77→88 ·
   `arity`/`mask`/`variable-name`/`builtin` unchanged. Nothing shrinks.
2. **Pre-existing cells are untouched on the side that matters.** 0 of 353,909
   shared cells changed their **djust** output. 1,728 changed their *Django*
   output, all of them a `VariableDoesNotExist` message that embeds the whole
   context dict repr — an artifact of Django's error text, not of a render.
3. **New divergences filed, not fixed (#1079).** See below.

## What is added

* `INPUTS`: a falsy inhabitant of the five variants that had only truthy ones
  (`i-zero`, `f-zero`, `dec-zero`, `t-empty`, `d-empty`), plus `td-zero` and
  `td-plain` — the first `Value::Encoded` the corpus has ever held, in either
  answer.
* `ARG_CONTEXT` / `ARG_SPELLINGS`: one resolved binding per `Value` variant in
  both answers, plus `known_str_zero` — a TRUTHY `str` spelling `0`, which is
  the row that separates a value-typed falsiness rule from a text-shaped one
  (Django's `timesince` raises on it; a text rule reads the falsy number it
  spells).
* **A `value-truthiness` axis**, so this is the last time. `Value::is_truthy`
  is a `match` over `Value`, so the engine does name the set: the variants are
  read out of `crates/djust_core/src/lib.rs` and each must have a falsy AND a
  truthy inhabitant in the value channel and the argument channel. The four
  uninhabitable combinations are `exempt` with a written reason — `Missing` and
  `DictView` never arrive from Python, `Value::None` has one inhabitant and it
  is falsy, a `BigInt` is never zero — so a stale exemption is reported rather
  than silent. `input-shape` stays UNVERIFIED for everything else: this closes
  the one slice of it the engine names, and says so.

`_value_variant` is a transcription of `impl FromPyObject for Value`'s arm
order and is labelled as one; the *required* side is mechanical, and an object
it cannot classify raises rather than vanishing from the swept set.

## What the widening surfaced — filed, not fixed

Three `Value::Encoded` holes in the VALUE position, each confirmed across the
whole datetime family, none of them constructible before:

* **#2471** — `values_equal` / `try_compare` have no `Encoded` arm, so two
  datetimes are never equal, *not even to themselves*: `{% if p == p %}` is
  `N`, `!=` is `Y`, and every ordering operator is `N`. Literally the #2335
  list bug one variant over, and it silently takes the `{% else %}` branch.
* **#2472** — `{{ p|pprint }}` spells `str(o)` where Django spells `repr(o)`:
  `'0:00:00'` for `datetime.timedelta(0)`.
* **#2473** — `{{ p|get_digit:"1" }}` echoes its input where Django raises
  `TypeError`; the value-position twin of #2366's argument-position rule.

Only 30 of the 2,344 new-only divergent cells belong to an expression class
that never diverged in the old corpus, and all 30 are these three.

## Tests

`TestItWouldHaveCaughtTheHistoricalBlindSpots` gains
`test_2469_no_cell_could_have_a_FALSY_argument_and_no_timedelta_existed`, which
reconstructs the pre-#2469 corpus in a copy and asserts the axis names all 21
gaps (7 value-channel, 14 argument-channel), *and* that the truthy inhabitants
the old corpus did have are still swept — so the axis reports the gap rather
than blaming itself. Its non-vacuity sibling
`test_2469_the_mutation_is_a_corpus_edit_and_not_an_axis_deletion` asserts the
axis is still declared in the mutated copy and that no other axis went red,
because a reconstruction that deleted the axis would also report `missing == []`
and let the first test pass by arithmetic (#1468/#2135).

`test_removing_the_pad_cap_spelling_makes_the_cap_unreachable` now removes two
spellings. `known_big` binds `12345678901234567890` — added for
`arg:BigInt:truthy`, since a magnitude past `i64` has no falsy inhabitant — and
a resolved argument of that size reaches `pad_width`'s cap by the same route
the literal does. That test going red on the first run is how the second route
was found, and the docstring records it.

Full suite green across all three roots.

|---|---|---|
| #2469 | `scripts/filter-parity-differential.py` | `test_2469_no_cell_could_have_a_FALSY_argument_and_no_timedelta_existed`, `test_2469_the_mutation_is_a_corpus_edit_and_not_an_axis_deletion`, `test_the_axes_are_the_ones_this_file_documents`, `test_every_addition_is_accounted_for`, `test_the_resolvable_lookup_binding_survives` |

No `crates/` change, no runtime change: this PR touches the measuring
instrument and its coupling tests only.


johnrtipton:
## The three checks, with results

### 1. Existing axes stable or grown, never shrunk

Measured by enumerating each cell generator before and after, on the same build:

| axis | before | after |
|---|---|---|
| `filter` | 2,337 | 2,736 |
| `chain` | 37,120 | 37,120 |
| `custom` | 7,076 | 7,104 |
| `tag` | 276,876 | 281,664 |
| `path` | 2,624 | 3,072 |
| `cmp` | 11,767 | 16,128 |
| `argument` | 13,920 | 21,344 |
| `arity` | 420 | 420 |
| `masked-refusal` | 384 | 384 |
| `variable-name` | 112 | 112 |
| `ctag` | 492 | 576 |
| `builtin` | 704 | 704 |
| `builtin x mapping` | 77 | 88 |
| **total** | **353,909** | **371,452** |

Nothing shrinks. And the stronger form of the same check: of the 353,909
**pre-existing** cells, **0** changed their djust output. (1,728 changed their
*Django* output — every one a `VariableDoesNotExist` message that embeds the
whole context dict repr, so the new `ARG_CONTEXT` bindings appear inside
Django's own error text. `agreement()` compares refusals by the fact of the
raise, and `moved_by_axis` reads the djust side only, so neither is affected.)

### 2. The regression gate still flags a real regression — run backwards

The change chosen is exactly the class this corpus was blind to: #2458's
`Value::Encoded(e) => e.truthy` reverted to its pre-fix
`!e.display.is_empty()`. Two genuinely different builds
(`7b3da5ededeea374` → `5d3f1f7c35c5ba16`), the same regression, compared under
each corpus:

```
pre-#2469 corpus   cells 353,909   0 moved on EVERY axis   REGRESSIONS: 0    exit 0
                   NOTE: identical agreement counts under BOTH definitions …

this corpus        cells 371,452   43 moved                REGRESSIONS: 38   exit 1
```

The 38: `json_script:known_td_zero` (16), `default:"D"` (7), `yesno:"y,n,m"`
(6), `slice:":3"` (3), `join` (3), `default_if_none:"D"` (3) — every one either
a `known_td_zero` argument or a `td-zero` value.

This is the #2454 failure reproduced against a *real* regression and then
closed: the old corpus printed `REGRESSIONS: 0, exit 0` over 38 genuinely
regressed cells. The old-corpus column is derived by restricting both measured
files to the pre-#2469 cell set, which is sound precisely because of check 1 —
the djust side of every shared cell is byte-identical between the corpora.

The Rust probe was reverted and rebuilt before any commit; the working `.so`
hashes back to `7b3da5ededeea374`, and `crates/` is untouched by this PR.

### 3. The canary — a divergence the old corpus could not construct

**#2471, #2472, #2473.** Of the 2,344 new-only divergent cells, exactly **30**
belong to an expression class that never diverged in the old corpus, and all 30
are `Value::Encoded` in the VALUE position — a variant the corpus could not
reach at all before, in either truthiness:

* **#2471** — `values_equal` / `try_compare` have no `Encoded` arm, so two
  datetimes are never equal, *not even to themselves*: `{% if p == p %}` is
  `N`, `!=` is `Y`, and every ordering operator is `N`. Confirmed for all four
  members of the family. This is literally the #2335 list bug one variant over,
  and it silently takes the `{% else %}` branch.
* **#2472** — `{{ p|pprint }}` spells `str(o)` where Django spells `repr(o)`.
* **#2473** — `{{ p|get_digit:"1" }}` echoes its input where Django raises
  `TypeError`; the value-position twin of #2366.

There is also a second canary for the mechanised half, in-suite:
`test_2469_no_cell_could_have_a_FALSY_argument_and_no_timedelta_existed`
reconstructs the pre-#2469 corpus in a copy and the `value-truthiness` axis
names all 21 gaps, with
`test_2469_the_mutation_is_a_corpus_edit_and_not_an_axis_deletion` as its
non-vacuity sibling.

## Two things the widening broke that were worth breaking

Both are in `5db87f89`, and neither is bookkeeping:

* **`test_removing_the_pad_cap_spelling_makes_the_cap_unreachable`** went red.
  `known_big` (`12345678901234567890`, added for `arg:BigInt:truthy` since that
  variant has no falsy inhabitant) reaches `pad_width`'s cap by a second,
  *resolved* route — so removing only the literal spelling now reports the cap
  REACHABLE, which is the honest answer. The mutation removes both.
* **`test_no_cell_renders_where_django_refuses`** (#2451) went red with twelve
  offenders, all `td-zero`/`td-plain`, and they are **#2467** rather than
  anything this branch changed. That sweep's djust column renders through
  `normalize_django_value` — the LiveView path — which flattens a `timedelta`
  to `"P0DT00H00M00S"` in Python, so djust iterates the thirteen CHARACTERS of
  that string where Django raises `TypeError: not iterable`:
  `{{ p|unordered_list }}` emits thirteen `<li>`s and `{{ p|phone2numeric }}`
  is `7038004006007`. The same twelve cells REFUSE on the raw path — measured
  in a new sibling test, because if they refused on both the pin would be
  hiding a hole in #2451's own chokepoint. Pinned as an exact SET so a
  thirteenth offender fails AND closing #2467 fails, forcing the pin's deletion
  with that fix.

That second one raises #2467's severity: it is not only a spelling and
truthiness divergence — on the path most djust pages use, the flattening makes
djust **more permissive than Django**. Written up on #2467 with the full
measurement.

## Merge

`git merge origin/main` (never rebase — `CHANGELOG.md` is `merge=union`, a
local-only driver). Set-difference of `[Unreleased]` against `origin/main`:
**0 lines dropped, exactly 6 added** — this PR's own bullets. `renderer.rs`
moved in #2470, so the crate was rebuilt and `uv.lock` restored.

johnrtipton:
Full suite green on the merge commit `878d0945` (post-`git merge origin/main`, crate rebuilt after #2470 moved `renderer.rs`):

```
pytest tests/ python/tests/ python/djust/tests/ -n auto
18450 passed, 420 skipped, 0 failed  (8:29)
```

One note for the record: the run *before* the coupled-test fixes had a lone `test_unknown_filter_parse_time_2419::test_the_django_library_bridge_registers_in_time_for_the_parser` failure. It passes in isolation and did not recur here. Diagnosis: that test mutates the global Django engine's `template_libraries` and the Rust custom-filter registry, and `tests/unit/test_rust_custom_filters_1121.py` calls `clear_custom_filters()` — a pre-existing xdist-scheduling pollution pair. My change perturbed the distribution only because it re-enabled twelve tests in `test_sequence_op_chokepoint_2451.py` that had been erroring at collection. Not introduced here, and not fixed here (#1079).

Issue #2462, opened 2026-08-30
normalize_django_value violates its own documented DjangoJSONEncoder identity for 4 of 10 datetime shapes — and the parity test samples none of them


johnrtipton:
Found while fixing #2448 and filed rather than folded in (#1079): it is Python,
a different file, and a different render path.

## The claim

`normalize_django_value`'s docstring (`python/djust/serialization.py:1190`)
states an identity:

```python
json.dumps(normalize_django_value(v), cls=DjangoJSONEncoder)
== json.dumps(v, cls=DjangoJSONEncoder)
```

and calls itself *"the **pre-pass** for `DjangoJSONEncoder`, not a replacement
for it"*.

## The measurement

Running the claim as written, over the datetime family:

```
ok   datetime naive     direct="2020-01-01T03:04:05"       normalized="2020-01-01T03:04:05"
FAIL datetime µs        direct="2020-01-01T03:04:05.123"   normalized="2020-01-01T03:04:05.123456"
FAIL datetime µs tiny   direct="2020-01-01T03:04:05.000"   normalized="2020-01-01T03:04:05.000007"
FAIL datetime utc       direct="2020-01-01T03:04:05Z"      normalized="2020-01-01T03:04:05+00:00"
ok   datetime offset    direct="2020-01-01T03:04:05+05:30" normalized="2020-01-01T03:04:05+05:30"
ok   date               direct="2020-01-01"                normalized="2020-01-01"
ok   time               direct="03:04:05"                  normalized="03:04:05"
FAIL time µs            direct="03:04:05.123"              normalized="03:04:05.123456"
ok   timedelta          direct="P0DT00H01M30S"             normalized="P0DT00H01M30S"
ok   timedelta neg      direct="-P0DT00H01M30S"            normalized="-P0DT00H01M30S"

4/10 violate the documented identity
```

Two `DjangoJSONEncoder.default` rules are missing:

```python
r = o.isoformat()
if o.microsecond:                       # ← not applied
    r = r[:23] + r[26:]
if r.endswith("+00:00"):                # ← not applied
    r = r.removesuffix("+00:00") + "Z"
```

## Why it matters

`normalize_django_value` is the LiveView context pre-pass
(`mixins/rust_bridge.py:1011`), so on that path a `datetime` reaches the
renderer already flattened to a string — meaning `{{ p|json_script:"d" }}` in a
LiveView emits the un-truncated / un-`Z`-rewritten form where Django emits the
[...]

PR #2475, opened 2026-08-30
fix(serialization): the LiveView path carries a datetime to Rust instead of flattening it (#2467)

### LINKED ISSUE #2467: The LiveView path flattens a datetime in Python, so Value::Encoded never reaches it — the two djust paths answer {% if timedelta(0) %} differently
Found while fixing #2462 and filed rather than folded in (#1079): it is a change
to what every consumer of `normalize_django_value` receives, not a spelling fix.

## The residue

#2448 gave the datetime family `Value::Encoded`, which carries `str(o)`,
`DjangoJSONEncoder.default(o)`, CPython's `tp_name` and (since #2458)
`bool(o)`. That closed the RAW path — a plain Django view rendering through
`DjustTemplateBackend`.

The LiveView path never builds one. `normalize_django_value`
(`mixins/rust_bridge.py:1011`, `mixins/context.py`, `mixins/jit.py` at seven
sites) flattens a `datetime` to a `str` **in Python**, so by the time Rust sees
it the type is gone and every downstream decision is made on text:

| | raw path | LiveView path |
|---|---|---|
| `{{ p }}` over `datetime(2020,1,1,3,4,5)` | `2020-01-01 03:04:05` (`str(o)`) | `2020-01-01T03:04:05` (the flattened ISO string) |
| `{{ p\|json_script:"d" }}` | `Value::Encoded`'s encoder field | the flattened string, re-encoded |
| `{% if p %}` over `timedelta(0)` | `F` (#2458's carried `bool(o)`) | `T` — the string `"P0DT00H00M00S"` is non-empty |
| `{{ p\|timesince:q }}` with `q` a `str` that is date-SHAPED | reads it as a date | same |

The third row is a live divergence that #2458 closed on one path only, and it
is the sharpest argument for this issue: the two djust paths now answer
`{% if p %}` differently for the same context value.

PR #2465 (#2462) makes the flattened *spelling* exact — all three sinks call
`DjangoJSONEncoder.default` — so the JSON is right. It does not stop the
flattening, because that is a different change.

## Why it is not a small change

`normalize_django_value`'s output goes to four kinds of consumer, and not all
of them can take a live `datetime`:

* the **template context** (Rust renderer) — could take it; that is the whole
  point.
* the **wire encoders** (`websocket.py`, `sse.py`, `api/dispatch.py`) — these
  run `json.dumps(..., cls=DjangoJSONEncoder)`, which handles a `datetime`
  fine.
* the **session / signed-snapshot round trip** (`state_roundtrip=True`) —
  `SerializableViewState` and Django's session serializer need a JSON-able
  value, and a restored `datetime` would come back as whatever the round trip
  produced. This is the boundary `decimal_for_state_roundtrip` already exists
  for, and it is the one that decides the shape of any fix.
* the **JIT serializer** (`mixins/jit.py`) — seven call sites, whose fallbacks
  are ordinary rather than rare.

The `Decimal` precedent (#2239) is the model: carried through UNCONVERTED for
the renderer, converted only at the `state_roundtrip=True` boundary. The same
split would work here — but it has to be measured per consumer, not assumed.

## Pinned

`python/tests/test_datetime_encoder_spelling_2462.py::TestWhatThisDeliberatelyDoesNOTClose::test_the_liveview_path_still_flattens_before_rust_sees_it`
and `python/tests/test_json_script_datetime_value_2448.py::TestWhichPathThisFixIsOn`
both assert the flattening, so closing this issue reddens them rather than
leaving them stale.

Refs #2448, #2458, #2462, PR #2464, PR #2465.

johnrtipton:
Closes #2467. **Stacked on #2474** (`fix/2469-falsy-argument-and-timedelta-corpus`) — see "Why this is stacked" at the bottom. Merge #2474 first.

`normalize_django_value` converted a `datetime` / `date` / `time` / `timedelta`
to a **string in Python**, so `Value::Encoded` (#2448) was never constructed on
the LiveView path and every downstream decision was made on text. #2456 fixed
the raw `DjustTemplateBackend` path and declared this bound for itself — in its
CHANGELOG, its docstring, and a `TestWhichPathThisFixIsOn` class that named the
boundary precisely. This is the other side of it.

## Reproduced on the actual LiveView path

`view.mount(request)` → `view.render()`, not `render_template` — a
renderer-only harness runs the *raw* path, which is exactly why #2456 could not
see this defect. 7 values × 7 templates, both djust entry points:

| value | template | raw path | LiveView path (before) |
|---|---|---|---|
| `datetime(2020,1,1,3,4,5)` | `{{ p }}` | `2020-01-01 03:04:05` | `2020-01-01T03:04:05` |
| `datetime(…, tzinfo=utc)` | `{{ p }}` | `…23:30:00+00:00` | `…23:30:00Z` |
| `timedelta(0)` | `{% if p %}` | `F` | **`T`** |
| `timedelta(0)` | `{{ p\|yesno }}` | `no` | **`yes`** |
| `timedelta(0)` | `{{ p\|default:"D" }}` | `D` | **`P0DT00H00M00S`** |
| `timedelta(90s)` | `{{ p\|time:"H:i" }}` | `00:01` | **`` (empty) |

**14 of 49 diverged. 0 after.**

## The row that makes this a fix rather than a preference

It is not only a spelling and truthiness divergence. #2451's sweep renders djust
through `normalize_django_value`, and with a `timedelta` in the corpus (#2469)
it put **twelve** cells in its `renders where Django refuses` bucket:

```
escapeseq      <td-zero>  django=<<TypeError>>      djust='[&#39;P&#39;, &#39;0&#39;, …]'  (13 items)
unordered_list <td-plain> django=<<TypeError>>      djust='<li>P</li><li>0</li>…'        (13 <li>s)
first / last   <td-zero>  django=<<TypeError>>      djust='P' / 'S'
phone2numeric  <td-zero>  django=<<AttributeError>> djust='7038004006007'
```

Django raises `TypeError: 'datetime.timedelta' object is not iterable`. djust
iterated the thirteen **characters of the flattened ISO duration string**. All
twelve refused correctly on the raw path throughout — so djust was **more
permissive than Django on the path most djust pages use** and stricter on the
one they do not. That asymmetry is closed here; #2474's twelve-cell pin is
deleted by this PR, exactly as it was written to require.

## The `Decimal` branch, verbatim (#2239)

```python
if isinstance(value, (datetime, date, time, timedelta)):
    return django_json_datetime(value) if state_roundtrip else value
```

Carried through UNCONVERTED for the renderer, converted only at the
`state_roundtrip=True` boundary. The consumer audit the issue asks for, **run
per consumer rather than assumed**:

| consumer | takes a live `datetime`? | evidence |
|---|---|---|
| template context (Rust renderer) | yes — the point | `TestTheTwoPathsAgree` |
| wire encoders (`websocket.py`, `sse.py`, `api/dispatch.py`) | yes — `json.dumps(…, cls=DjangoJSONEncoder)`, and since #2462 djust's encoder spells a datetime exactly as Django's does | `TestTheWireBytesAreUnchanged` asserts the emitted frame is **byte-identical** with and without the pre-pass |
| session / signed snapshot | no — and every write already passes `state_roundtrip=True` (`mixins/request.py:270`, `:275`, `:718`, `:728`; `mixins/components.py:149`) | `TestStateRoundtripBoundary` pins both directions |
| Rust state round trip | yes, and it was already solved | `TestTheRustStateBackendRoundTrip` |

## Wire shape: the #1448 pin already exists

`SerializableViewState.state` is `HashMap<String, Value>` through `rmp_serde`,
and `Value::Encoded` already has a **tagged** msgpack encoding (`ENCODED_TAG`,
`crates/djust_core/src/lib.rs`) whose payload is
`[type_name, display, json, truthy]` — #2458 added the fourth element for
exactly the reason this issue exists. It is pinned literally in
`test_encoded_truthiness_2458.py`
(`payload == ["datetime.timedelta", "0:00:00", "P0DT00H00M00S", False]`), with
a three-element legacy read for rolling deploys and a forgery-discrimination
test. **This PR changes no wire shape** — it makes the LiveView path reach one
that was already pinned. The client-facing JSON is unchanged and asserted so.

## The cost, stated

`{{ p }}` on the LiveView path renders `str(o)` now: `2020-01-01 03:04:05` where
it rendered `2020-01-01T03:04:05`, and `…+00:00` where it rendered `…Z`. Both
were already non-Django — Django LOCALIZES a bare datetime
(`Jan. 1, 2020, 3:04 a.m.`) — so this moves one non-Django spelling to the other
one djust already uses on its raw path, and buys agreement between djust's own
two paths. #2462's CHANGELOG made the mirror-image trade in the other direction
and said so; this reverses that particular line, and
`TestWhatThisCosts::test_the_bare_render_spelling_moves` states it in the suite
rather than leaving it to be discovered.

The date filters are unaffected — timezone conversion for an aware datetime and
the bare-`time` no-shift rule (#2216) both verified — and
`{{ p|time:"H:i" }}` over a `timedelta` *improves* from empty to `00:01`. An
aware `datetime.time` (the one value Django's encoder refuses) renders the same
string on both paths before and after, because Rust's `django_json_encoded`
fails closed to `str(o)` and `str(aware_time) == aware_time.isoformat()`.

## Pins inverted rather than deleted

Three places asserted the flattening by name, and closing this reddens them —
which is what they were for. Each is inverted the way #2462 inverted #2448's:

* `test_json_script_datetime_value_2448.py::TestWhichPathThisFixIsOn::test_the_normalizer_flattens_the_type_before_rust_can_carry_it`
* `test_datetime_encoder_spelling_2462.py::TestWhatThisDeliberatelyDoesNOTClose::{test_the_liveview_path_still_flattens_before_rust_sees_it, test_the_bare_render_spelling_of_an_aware_datetime_changed}`
* `test_sequence_op_chokepoint_2451.py::test_no_cell_renders_where_django_refuses` — the twelve-cell exact-set pin #2474 added, now back to `not offenders`, with its `test_2467_the_same_twelve_cells_refuse_on_the_RAW_path` sibling kept as the control.

## Gate-off (#1468)

Restoring `return django_json_datetime(value)` unconditionally reddens
`TestTheTwoPathsAgree` (14 of 49), `TestTheTwelvePermissivenessCells` (12 of
12), `TestTheNormalizerCarriesTheObject` (7 of 7) and the `timedelta` rows of
`TestTheRustStateBackendRoundTrip`. `TestTheWireBytesAreUnchanged` and
`TestStateRoundtripBoundary` pass in BOTH states, deliberately — they pin the
consumers this change does not move, and saying which classes do *not* redden
is the honest half of a gate-off report.

## Why this is stacked on #2474

#2474 added the twelve-cell #2467 pin as an exact set precisely so that closing
this issue would turn it red and force its deletion — a stale exemption was the
failure mode being avoided. Branching this off `main` instead would give a
locally-green PR that breaks the moment #2474 merges, in either order. Merge
#2474 first; this then applies cleanly.

|---|---|---|
| #2467 | `python/djust/serialization.py` | `TestTheTwoPathsAgree`, `TestTheTwelvePermissivenessCells`, `TestTheNormalizerCarriesTheObject`, `TestTheWireBytesAreUnchanged`, `TestStateRoundtripBoundary`, `TestTheRustStateBackendRoundTrip`, `TestWhatThisCosts` |

No `crates/` change: the Rust side of this landed in #2448/#2458 and is only
now reachable from the LiveView path.


johnrtipton:
**Stacking note for the reviewer.** This PR's commit list currently shows #2474's four commits as well, because it is branched on top of it — GitHub compares against `main`. Once #2474 merges, this PR's diff reduces to its own two commits (`120a236a`, `6670c0c2`).

Reviewing it now against #2474's head is the honest comparison: `git diff fix/2469-falsy-argument-and-timedelta-corpus..fix/2467-liveview-path-carries-datetime` is 7 files — `python/djust/serialization.py` (the one-line branch plus its comment), the new `test_liveview_path_carries_datetime_2467.py`, four inverted pin files, and `CHANGELOG.md`.

The dependency is not cosmetic: #2474 added #2451's twelve-cell exact-set pin *specifically* so that closing #2467 would redden it and force its deletion, rather than leaving a stale exemption. Either branch alone is green; merged in either order without the other, the pin and the fix contradict. Merge #2474 first.

johnrtipton:
## The nine CI failures, judged one at a time

All nine reproduced locally and all nine are `normalize_django_value` output-spelling assertions — the expected blast radius of no longer flattening. Two of them are claims **about this change**, so they were decided rather than updated.

### 1. `#2252::test_the_flag_changes_nothing_for_a_decimal_free_value` — the flag is no longer a no-op, and it must not be

Its name is a claim about `state_roundtrip`: *"the flag's ONLY effect is the `Decimal` branch"*. True when written; false for the datetime family after this PR, **by construction**.

Of the two possible resolutions, the second is not available, and that is measurable rather than a preference:

```
DjangoSessionSerializer().dumps({"p": normalize_django_value(date(2024,6,15))})
    -> TypeError
DjangoSessionSerializer().dumps({"p": normalize_django_value(date(2024,6,15), state_roundtrip=True)})
    -> b'{"p": "2024-06-15"}'
```

Django's session serializer passes no encoder. Carrying the datetime on **both** sides of the flag would write a value the session cannot store — which is the entire reason the flag exists, and why the `Decimal` branch has the same shape. So: **the flag legitimately does something for the datetime family now.**

Rather than exempt four ids, the class is split so the narrowed claim is the **rule** and not a list:

> the flag changes nothing for a value that holds no *carried-through* type.

* `_UNTOUCHED` (10 types) keeps the no-op assertion, renamed `test_the_flag_changes_nothing_for_a_CARRY_FREE_value`.
* `_CARRIED_THROUGH` (4) gets `test_the_flag_DOES_move_a_carried_through_value` — unflagged **is** the object, flagged is `django_json_datetime(value)`, and they differ — plus `test_the_carry_through_types_are_exactly_the_ones_the_boundary_must_convert`, which runs the two `dumps` calls above.
* `test_an_untouched_value_needed_no_conversion_in_the_first_place` closes the loop from the other side: every type on the untouched list is one the serializer already accepts unflagged. **The two lists cannot drift apart silently** — a type added to the carry-through side and forgotten at the boundary reddens, which a plain exemption list would have allowed.

The randomized corpus is split the same way instead of dropping its `date` leaf (dropping it would have made the sweep pass by shrinking — the failure that file's own randomized-corpus argument is against): 500 carry-free values stay bit-identical; 500 carried-type values are swept for the property that matters for them, with `assert refused > 100` as a floor proving the generator actually produces them.

### 2. `#2239::test_the_timedelta_gap_is_CLOSED` — still closed, by the same route

The gap this names is the **encoder's** (djust's raised `TypeError` where Django's answered `duration_iso_string`). Both encoder assertions still pass untouched; #2467 changed the *pre-pass*, not the encoder. Its third assertion moved to the roundtrip form, and a fourth was added that is the stronger claim anyway:

```python
json.dumps(normalize_django_value(td), cls=DjustEncoder) == '"P1DT00H01M30S"'
```

— encoding the pre-pass's output is byte-identical to encoding the input, which is the identity the docstring states and the property every wire consumer depends on.

### 3–9. Documented-behaviour rows

`TestDateTimeTypes` (4), `test_normalize_date`, `test_normalize_dict_with_complex_values`, `test_normalize_then_rust_update_state_no_quote_wrap`. Each keeps **both** halves — carried, and converted at the boundary — because asserting only the first would let the boundary silently stop converting. `test_normalize_date` is lifted from the `test_normalize_decimal` row directly beneath it, which has had exactly this shape since #2239.

The 1081 one deserves a note: its two "precondition" lines encoded the old flattening, but the reporter's trigger — a `filed_date` that is already a `str` when `|date` sees it — is still a **live input**, because a session restore hands one back. So it is kept and swept as a second case (`for context in (json_ctx, stored_ctx)`) rather than retired with the precondition that used to produce it.

## Verification

```
pytest tests/ python/tests/ python/djust/tests/ -n auto
18596 passed, 420 skipped, 0 failed  (4:21)
```

Citations grepped per #1197 — a phantom `TestTheFlagMovesExactlyTheCarriedThrough` class name in a docstring and in the CHANGELOG is replaced with the four real test names.

One thing left as-is deliberately: #2252's `[Unreleased]` bullet says "2828 decimal-free corpus rows over 14 types" and "86 in `test_decimal_state_tag_2252.py`". Both are stale now (10 + 4 types; 101 collected). I did not edit that bullet — `CHANGELOG.md` is `merge=union` and editing an existing entry is how two divergent copies get produced. The correction is stated in this PR's own entry instead, which also records *why* the number moved.

johnrtipton:
**Gate-off re-measured after the fix-pass**, since updating nine pins changes the blast radius. Mutation asserted to match exactly once, source asserted to differ, every `__pycache__` cleared per iteration, pytest ERRORS counted apart from failures (#2129/#2135):

| suite | fix in place | gated off |
|---|---|---|
| the new #2467 file | 0 failed, 131 passed | **39 failed**, 92 passed |
| the inverted pins | 0 failed, 572 passed | **16 failed**, 556 passed |
| the nine CI pins re-judged | 0 failed, 342 passed | **16 failed**, 326 passed |
| consumers this must NOT move | 0 failed, 113 passed | **0 failed**, 113 passed |

71 cases redden. The last row is the load-bearing one and is stated as a negative on purpose: `TestTheWireBytesAreUnchanged`, #2458's literal msgpack payload pin and #2216's time-only renders pass in **both** states — which is the evidence that this changes no wire shape and no date-filter answer. A gate-off that only reports what went red cannot say that.

Recorded in `TestTheGateOffWouldFail`'s docstring (`d1c049da`), along with the correction from the first run: that group was predicted 0 and came back 1, and the answer was that the classification was wrong, not the test — `test_django_session_serializer_refuses_the_bare_datetime` asserts the premise that makes the flag necessary, which is only observable once the pre-pass stops converting.

johnrtipton:
**#2474 is merged**, so the stacking note above is now historical: this branch is 0 commits behind `origin/main` and its diff against current main is exactly its own 12 files.

```
git diff --stat origin/main..HEAD
 python/djust/serialization.py                              |  63 ++-
 python/tests/test_liveview_path_carries_datetime_2467.py   | 434 +++++
 python/tests/test_sequence_op_chokepoint_2451.py           | 145 +-
 python/tests/test_decimal_state_tag_2252.py                | 194 ++-
 python/tests/test_datetime_encoder_spelling_2462.py        |  75 +-
 python/tests/test_json_script_datetime_value_2448.py       |  21 +-
 python/tests/test_decimal_converters_2239.py               |  13 +-
 python/tests/test_serialization_hardening.py               |  22 +-
 python/djust/tests/test_async_result_serializer.py         |  21 +-
 tests/unit/test_normalize_django_value.py                  |  41 +-
 tests/unit/test_filter_literal_args_1081.py                |  35 +-
 CHANGELOG.md
 12 files changed, 879 insertions(+), 197 deletions(-)
```

One production line changed; everything else is the test surface that pinned the old behaviour, plus the new sweep. No `crates/` change.

johnrtipton:
Pushed `ceb237f3` — one more item found by self-review rather than by CI, and it is a tautology of my own making (#1200).

`TestParityWithJSONRoundtrip` asserts `_encoded(normalize_django_value(v)) == _encoded(v)`. For a value the pre-pass **carries through**, both sides encode the same object — the row is true by construction. That has been so for `Decimal` since #2239 (the class docstring says so: *"it also still covers Decimal"*), and this PR adds four datetime rows to that set. Not a bug, but four more rows that can no longer go red, so the non-vacuous claim is now asserted beside them: whatever the `state_roundtrip=True` boundary **stores** must encode to the same JSON the live object does.

The interesting part is that writing that sweep with `Decimal` in it **failed**:

```
assert {'__djust_decimal__': '9.99'} == '9.99'
```

The two carried types do not share a boundary form, and the difference is load-bearing. A datetime's stored form is the encoder's own string, which every consumer already reads as a date; a `Decimal`'s cannot be — #2252 measured that a bare digit string restored onto the view stops `|floatformat` rounding and a `float` loses the type and the trailing zeros — so it is stored as a tag that `decode_state_roundtrip` restores. Rather than narrow the sweep and hide that, the asymmetry gets its own case asserting the different contract, placed exactly where someone would otherwise widen the sweep back.

Local full suite on the pushed tree: **18601 passed, 420 skipped, 0 failed** (4:30). CI re-running.

johnrtipton:
**Green.** CI on `ceb237f3`: **16 checks pass, 0 fail**, MERGEABLE (`BLOCKED` is only the maintainer-approval gate).

`python-tests (py3.12)` — the job that reported the nine failures — passes. For the record, the intermediate run on `d1c049da` shows `cancelled` rather than a result: it was superseded by the next push, not a second failure.

Both halves verified on the pushed tree:

| | result |
|---|---|
| CI, all jobs | 16 pass / 0 fail |
| local `pytest tests/ python/tests/ python/djust/tests/ -n auto` | 18601 passed, 420 skipped, 0 failed |
| gate-off (branch reverted) | 71 redden; the must-NOT-move group 0 in both states |

Ready for review.

johnrtipton:
**Correction — my "green" report above was for `ceb237f3`, and the branch has since moved.**

`origin/main` was merged in as `8911476a`, bringing **#2476** (`bool(set())` is False; the falsiness rule reaches the conversion). That is not an unrelated merge: it changes `crates/djust_core/src/lib.rs`, `filters.rs` and `renderer.rs` — the same `Value` conversion and truthiness surface this PR now feeds live `datetime` objects into — and it edits `test_encoded_truthiness_2458.py`, which this PR's msgpack round-trip claim leans on. So the 16/16 result no longer describes the head, and CI has restarted on the merge commit.

Re-verified against the merged tree:

* Crate **rebuilt** — `.so` mtime 17:57 → 19:30 and size 6597760 → 6597776, so the Rust really did change and the previous artifact was stale.
* The interaction surface passes: `test_liveview_path_carries_datetime_2467.py` + `test_falsy_conversion_2466.py` + `test_encoded_truthiness_2458.py` + `test_sequence_op_chokepoint_2451.py` + `test_decimal_state_tag_2252.py` → **580 passed, 0 failed**.
* CHANGELOG set-difference against the new `origin/main`: **0 dropped, 6 added** (this PR's own bullets).
* Full suite across all three roots running now; will report the result and the settled CI on `8911476a` rather than restating the old one.

johnrtipton:
**Green on `8911476a`** — the merge commit that brought #2476 in, which is the current head.

| | result |
|---|---|
| CI run `33341946762` | **completed / success**, 0 failing jobs |
| PR checks | **16 pass**, 0 fail, 0 pending |
| local `pytest tests/ python/tests/ python/djust/tests/ -n auto` | **18740 passed**, 420 skipped, **0 failed** (4:08) |
| the two-path sweep, re-run on the #2476 build | **0 of 49** diverging (was 14) |
| CHANGELOG set-difference vs new `origin/main` | 0 dropped, 6 added |

The crate was rebuilt after the merge before any of that — `.so` 17:57 → 19:30, size 6597760 → 6597776 — so none of it was measured against the stale artifact. That mattered: #2476 changes `Value` conversion and truthiness, the exact surface this PR now feeds live `datetime` objects into.

This supersedes my earlier green report, which was for `ceb237f3` and stopped describing the head the moment the merge landed.

PR #2476, opened 2026-08-30
fix(templates): bool(set()) is False, and the falsiness rule reaches the conversion (#2466)

### LINKED ISSUE #2466: bool(set()) is False in Python and True in djust — the #2458 shape at the conversion, for every type with no Value variant

johnrtipton:
Closes #2466.

```
{% if p %}T{% else %}F{% endif %}   p = set()         python False   django F   djust T -> F
                                   p = frozenset()   python False   django F   djust T -> F
```

A `set` has no `Value` variant, so `FromPyObject for Value` landed it on its final `Ok(Value::String(ob.str()?))` and it arrived as the non-empty string `"set()"`, whose `is_truthy` is `!s.is_empty()`. The #2458 shape one level up — at the CONVERSION rather than in the rule — which #2464's fix cannot reach, because a `set` never becomes an `Encoded`.

## The class is seven shapes, not two, and it is open

Swept against live Django and live Python over 32 container and scalar shapes rather than transcribed:

| shape | why it was truthy here |
|---|---|
| `set()` / `frozenset()` | the `str()` fallback — the two the issue names |
| `complex(0)` | the same fallback (it does not extract as `f64`) |
| `{}.keys()` / `.values()` / `.items()` | the same fallback — `Value::DictView` exists, but only the template's own `d.keys` access ever built one; the conversion never did |
| a class with `__len__` → 0 | the same fallback |
| a class with `__bool__` → `False` | the same fallback |
| a falsy class **with attributes** | the `__dict__` bulk-dump arm — **not** the `str()` fallback, and **not** closed here |

The last three are user classes, so the set cannot be enumerated — which is the issue's own argument for option 2 (carry `bool(o)`) over option 1 (give `set` a variant), and a one-type fix here is the shape #2129 took five rounds over.

Nineteen falsy shapes that were **already right** are swept too — `{}`, `[]`, `()`, `""`, `0`, `Decimal("0")`, `timedelta(0)`, `b""`, `bytearray()`, `range(0)`, `deque()`, `OrderedDict()`, `Counter()`, `array('i')`, `Fraction(0)`, `memoryview(b"")`, … — because they are what an over-reaching fix takes with it.

## No new carrier

`Value::Encoded` already **is** one: a Python object held by its `type_name` / `display` / `json` / `truthy` spellings because the object itself cannot cross. #2448 built it for the four `DjangoJSONEncoder` types, #2458 added the truthiness bit; `falsy_opaque` widens the set of objects that use it and adds no mechanism. A new `Value` variant would be a second carrier for one question (#1646) and would need classifying at every wildcard `match` arm in the workspace.

`json` stays `str(o)` for these — byte-for-byte what the `Value::String` path already wrote — so `json_script` does not move.

## Two bits, and the second is a defect the first version of this fix shipped

A truthiness-only fix made `{% for x in set() %}` **REFUSE** where Django renders the `{% empty %}` block — a regression in the one direction this class must not move. Found by re-running the axis after the first version was green, not by inspection.

Django asks two different questions and gets different answers for the same object:

```python
# ForNode.render                     -> {% for %} over a zero-__len__ class renders {% empty %}
if not hasattr(values, "__len__"):
    values = list(values)
# safeseq                            -> the same object RAISES here
[mark_safe(obj) for obj in value]
```

So `Encoded` carries `sized_empty` (`len(o) == 0`) and `iterable` (`iter(o)` succeeds) separately. `python_len` and the `{% for %}` arm read the first; `iter_values` reads the second. `iter(o)` is safe to ask — it builds an iterator and consumes nothing.

Every axis, against live Django:

| | before | Django | after |
|---|---|---|---|
| `{% if set() %}` | T | F | F |
| `{{ set() }}` | `set()` | `set()` | `set()` |
| `{{ set()\|length }}` | 5 | 0 | 0 |
| `{% for x in set() %}` | 5 chars | `{% empty %}` | `{% empty %}` |
| `{{ set()\|join:"," }}` | `s,e,t,(,)` | `""` | `""` |
| `{% for x in complex(0) %}` | `0j` chars | `TypeError: 'complex' …` | `TypeError: 'complex' …` |
| `{% for x in LenZero() %}` | repr chars | `{% empty %}` | `{% empty %}` |
| `{{ LenZero()\|safeseq }}` | escaped repr | `TypeError: 'LenZero' …` | `TypeError: 'LenZero' …` |
| `{{ LenZero()\|join:"," }}` | repr chars | the repr (its `except TypeError`) | the repr |

## Declined, so nothing becomes stricter than Django

| shape | why | pinned |
|---|---|---|
| falsy with `__iter__`, no `__len__` | Django `list()`s it and renders the items; producing them means RUNNING the object, which consumes a generator | `test_a_falsy_object_that_is_ITERABLE_with_no_len_is_still_truthy`, which also asserts djust did **not** start refusing |
| falsy with `__bool__` False and `__len__` > 0 | same | `test_a_falsy_object_with_a_NONZERO_len_is_still_truthy` |
| falsy **with attributes** | it reaches the `__dict__` arm and becomes a non-empty `Value::Object`; routing it here would take `{{ obj.a }}` with it | `test_a_falsy_object_WITH_attributes_is_still_truthy`, which asserts the attribute still resolves |
| a TRUTHY no-variant object still claims to be a `str` | not a truthiness defect (`{% if %}` agrees); closing it needs an enumeration | `test_a_TRUTHY_no_variant_object_still_claims_to_be_a_str` |

All four assert the **diverging** direction, so closing one reddens this file rather than leaving a stale exemption.

## Wire format

`__djust_encoded__`'s msgpack payload grows from four elements to six, both new ones **trailing** — the safe position in a positional payload (#1541). The reader accepts six, four and three, so a Redis state entry written by a 1.1.x or a #2448-era process still loads with the truthiness it was written with, and `test_a_FOUR_element_payload_still_reads` is the new compatibility case beside #2458's three-element one.

## Corpus: stated, not swept

`scripts/filter-parity-differential.py` **cannot construct a single cell of this class**, and the reason is structural rather than an omission: its `value-truthiness` axis (#2469, merged into this branch) reads the `Value` **variants** out of `crates/djust_core/src/lib.rs` and requires a falsy and a truthy inhabitant of each — and the values this fix is about are exactly the ones with **no variant**. So the axis reports `0 MISSING` while being blind to them.

Adding `set-empty` / `set-plain` to `INPUTS` was tried and **reverted**, because it does not measure this fix:

* it surfaces four `django REFUSES & djust RENDERS` cells (`first` / `last`) that belong to a **different** defect — `normalize_django_value` turns a set into a sorted list on the LiveView path, so subscript-refusing filters render there while Django raises `'set' object is not subscriptable`. That divergence is not this fix's and is unchanged by it (on the raw backend path this PR makes `first`/`last` over a set refuse *correctly*);
* a `dict_keys` cannot be a corpus row at all — `measure` `deepcopy`s every `q` operand and a view is unpicklable.

Both are filed as follow-ups rather than folded in (#1079). The two-build differential is therefore reported as a **non-regression** result over the 48 shapes the corpus does hold — the `Encoded` family included, since #2469 added a `timedelta` — and explicitly **not** as evidence that the fix works. That evidence is the 138 cases below, every one against live Django.

## Tests

138 cases in `python/tests/test_falsy_conversion_2466.py`, plus the flipped pins:

* `TestPythonsOwnAnswerIsRunNotTranscribed` — `bool()` **called** for all 42 shapes. The point is not `set()`; it is `memoryview(b"")`, `array('i')` and `Fraction(0)`, which are not obvious and which a transcribed table gets wrong.
* `TestTheForArmDoesNotBecomeSTRICTERThanDjango` — the direction the first version moved.
* `TestTheTwoIterabilityQuestionsAreTwoQuestions` — the `LenZero` row that makes the two bits distinguishable from one bit, on `{% for %}`, `|length`, `|join`, `|safeseq`, `|escapeseq` and `|unordered_list`.
* Rust: `every_variant()` gained the `Encoded` samples it **never had** — the variant landed in #2448 and neither iterability probe named it for two releases — and now carries all three shapes, with a canary asserting the asymmetric one is present. `python_len_agrees_with_iter_values` gained the one exemption Python itself requires (`len()` uses `__len__`, `iter()` needs `__iter__`), with an unreachability argument asserted rather than argued. `for_iterability_agrees_with_iter_values` pins the exact relation between the `{% for %}` arm and `iter_values` in both directions.

## Gate-off

Twelve mutations, each asserting the text was found and matched **exactly once**, that the source changed, that the crate **rebuilt** (`.so` mtime advanced), with `__pycache__` cleared and `N error` counted apart from `N failed`:

| mutation | reddened |
|---|---|
| never build a `falsy_opaque` (the whole fix off) | 43 |
| carry `true` instead of Python's `bool(o)` | 12 |
| claim `sized_empty` for everything | 1 |
| never `sized_empty` (the for-arm regression) | 7 |
| `iter_values`: drop the `Encoded` arm | 13 |
| `iter_values`: read `sized_empty` instead of `iterable` (one bit for both) | 3 |
| claim `iterable` for everything | 11 |
| never `iterable` | 13 |
| claim a falsy iterable with no `__len__` (the REFUSES growth) | 1 |
| `python_len`: drop the `Encoded` arm | 4 |
| `{% for %}`: drop the `Encoded` normalisation arm | 8 |
| serde: stop writing the two new payload elements | 3 |

No survivors; restored baseline `0 failed`. The two that redden a single test each are the two **declines** — each has one dedicated case, which is what makes a decline a mechanism rather than a comment.

`python_len`'s arm first reddened **only its structural pin**, because `|length` answers 0 through its own `unwrap_or(0)` either way — a valid mutation that is a semantic no-op for every tested input, which is the gate-off failure mode the v1.1.1-2 canon names. It is now reached behaviourally through `{% for a, b in … %}`, whose refusal reports the item **length**: Django says `got 0.` and the un-armed engine would say `got 1.` The number went 1 → 4.

## Suite

* Full suite, all three roots: **18,586 passed, 420 skipped, 0 failed** (after merging `origin/main`, which had advanced by #2470 and #2474).
* `make test-rust`: 64 suites, all green.


johnrtipton:
## Two-build differential — 371,452 cells, `f7b84485` → `f15dc3ac`

**This is a non-regression result, not evidence the fix works, and the tool says so itself.**

```
cells        : 371452
agree BEFORE : 270429   (refusal-collapsed: 316755)
agree AFTER  : 270429   (refusal-collapsed: 316755)
django REFUSES & djust RENDERS: 4485 -> 4485   (+0)
djust REFUSES & Django RENDERS: 38773 -> 38773   (+0)

djust output changed, by axis:
  argument / arity / builtin / builtin x mapping / chain / cmp / ctag /
  custom / filter / masked-refusal / path / tag / variable-name    0 moved

builds       : f7b844854e0095f3 -> f15dc3ac223b24c1  (genuinely two builds)
```

The harness prints its own caveat for exactly this shape:

> NOTE: identical agreement counts under BOTH definitions — exact AND refusal-collapsed — and the two builds DIFFER, so this is not a stale baseline. Either the change moves nothing, or it moves an axis this corpus does not sweep.

It is the second. **The corpus cannot construct a single cell of this class**, and the reason is structural rather than an omission: `value-truthiness` (#2469, merged into this branch and reporting `52 required, 12 exempt, 0 MISSING`) reads the `Value` **variants** out of `crates/djust_core/src/lib.rs` and requires a falsy and a truthy inhabitant of each — while every value this fix is about has **no variant**, which *is* the defect. The axis reports clean while being blind to them. Filed as part of #2477.

### What the run does establish

Both new `Encoded` arms and the widened wire format ride through **371,452 cells** without disturbing anything, and the `Encoded` family is genuinely swept now that #2469 added `td-zero` / `td-plain`:

* `iter_values` and `python_len` gained `Encoded` arms — 0 cells moved on any axis, so no `timedelta` cell changed;
* the `{% for %}` arm gained a normalisation branch — the `tag` axis (281,664 cells) moved 0;
* the msgpack payload went from four elements to six — 0 regressions, and `test_a_FOUR_element_payload_still_reads` covers the rolling-deploy read;
* `djust REFUSES & Django RENDERS` is **+0**, which is the number that would have caught the regression the first version of this fix shipped (`{% for x in set() %}` refusing where Django renders `{% empty %}`) had the corpus been able to see it. It could not — that one was caught by re-running the axis by hand.

0 agreeing cells regressed, 0 cells newly panic, live-payload leaks unchanged at **54** (0 closed, 0 introduced, 0 live).

### Where the evidence actually is

The 138 cases in `python/tests/test_falsy_conversion_2466.py`, every one against live Django and live Python, and the twelve gate-off mutations in the PR body. Two follow-ups make the corpus able to measure this class next time:

* **#2477** — `normalize_django_value` turns a set into a list on the LiveView path, so `first`/`last` render where Django raises. Adding `set-empty`/`set-plain` to `INPUTS` reddens 9 tests across 4 files by surfacing that, and a `dict_keys` cannot be a corpus row at all (`measure` deepcopies the `q` operand). Both block the row.
* **#2478** — the `__dict__`-arm twin: a falsy object *with attributes* is still truthy.

Issue #2486, opened 2026-08-31
`{{ dec.real }}` renders empty — `Value::Decimal` is a different carrier from the one #2481 gave attributes to


johnrtipton:
Split out of #2481 per CLAUDE.md #1079 — measured on its branch, pinned there
in the diverging direction, and **not** fixed there.

## The divergence

```
{{ p.real }}        Decimal('1.5')   django '1.5'                                       djust ''
{{ p.imag }}        Decimal('1.5')   django '0'                                         djust ''
{{ p.as_tuple }}    Decimal('1.5')   django 'DecimalTuple(sign=0, digits=(1, 5), …)'    djust ''
{{ p.is_finite }}   Decimal('1.5')   django 'True'                                      djust ''
{{ p.adjusted }}    Decimal('1.5')   django '0'                                         djust ''
```

`{{ p.real }}` and `{{ p.imag }}` are plain data attributes (no call); the other
three are Django's auto-call (ADR-024) and belong with the method half filed
alongside this.

## Why #2481's fix does not reach it

#2481 gave `Value::Encoded` an attribute map and taught `context::lookup_segment`
to read it. A `Decimal` is not an `Encoded`: `FromPyObject for Value` claims it
in its own arm (`is_decimal`, #2214) and it becomes `Value::Decimal(String)` — a
DIFFERENT variant with no attribute slot. So this needs either

* an attribute slot on `Value::Decimal` — which means a second carrier for the
  same question and a second wire shape to keep aligned (#1646), or
* routing `Decimal` through the `Encoded` carrier, which would have to preserve
  everything `Value::Decimal` exists for: exact-digit `==` and `<` against
  `Integer` / `Float` (#2214, `is_decimal_pair` / `numeric_pair`), `floatformat`
  rounding, the `DECIMAL_TAG` state round trip, and the `IntoPyObject` arm that
  rebuilds a real `decimal.Decimal` so a handler reading it back does not see
  its type change (`Encoded`'s arm deliberately comes back as a STRING).

The second is the convergent shape and the more dangerous one; whichever is
chosen should be argued before it is written.

## Pinned

`python/tests/test_encoded_attributes_2481.py::TestWhatThisDeliberatelyDoesNOTClose::test_a_decimals_attributes_are_a_DIFFERENT_carrier`
asserts these in the diverging direction, so a fix reddens the exemption.

Refs #2481, #2214, #2448, #1646.

johnrtipton:
[...]

Issue #2485, opened 2026-08-31
`{{ dt.isoformat }}` and `{{ dt.max }}` still render empty — the auto-called and class-attribute halves #2481 declined


johnrtipton:
Split out of #2481 per CLAUDE.md #1079 — measured on its branch, pinned there
in the diverging direction, and **not** fixed there.

#2481 gave `Value::Encoded` an attribute map and taught `context::lookup_segment`
to read it, so `{{ dt.year }}` / `{{ td.days }}` and the other 21 DATA attributes
now render exactly as Django renders them. Two neighbouring families do not, and
each is declined for its own reason rather than for one shared one.

## Family 1 — nullary METHODS (Django auto-calls them)

```
{{ p.isoformat }}      datetime   django '2026-03-04T05:06:07.000008'   djust ''
{{ p.weekday }}        datetime   django '2'                            djust ''
{{ p.ctime }}          date       django 'Wed Mar  4 00:00:00 2026'     djust ''
{{ p.total_seconds }}  timedelta  django '259290.000005'                djust ''
{{ p.utcoffset }}      aware dt   django '0:00:00'                      djust ''
{{ p.date }} {{ p.time }} {{ p.toordinal }} {{ p.timetuple }} {{ p.timestamp }}
{{ p.tzname }} {{ p.dst }}
```

Django reaches these through `Variable._resolve_lookup`'s auto-call (ADR-024),
which turns a LOOKUP into an EVALUATION. #2481's map holds lookups: putting a
call's result in it would make the whole family eager at conversion time (every
`datetime` in every context pays for a dozen Python-level calls whether or not a
template asks), and would inherit whatever the call raises — `timestamp()` on a
naive datetime is platform-dependent, `utcoffset()` on a hostile `tzinfo` raises
whatever it likes. That is a different mechanism and so a different decision.

Options for whoever takes it, none of them obviously right:

1. carry a second, LAZY map — needs an interpreter at render time, which is
   exactly what `Value::Encoded` exists because there isn't;
2. carry the results eagerly for a small allowlist (`isoformat`, `weekday`,
   `total_seconds`, `toordinal` are cheap and total) and leave the rest;
3. decline the family and say so in the docs.

## Family 2 — CLASS attributes, and why they are not a name-list oversight

```
{{ p.min }}         datetime   django 'Jan. 1, 0001, midnight'          djust ''
{{ p.max }}         date       django 'Dec. 31, 9999'                   djust ''
[...]

Issue #2484, opened 2026-08-31
A msgpack state round trip turns every `Value::None` into `Value::Missing`, so `{{ p }}` renders 'None' then '' after one cache hit


johnrtipton:
Found while closing #2481 (`Value::Encoded` carries no attributes). It is a
separate defect on a far more general shape, so it is filed separately per
CLAUDE.md #1079 rather than folded in.

## The divergence

`Value::None` and `Value::Missing` are deliberately DISTINCT (#2203) — `None`
renders `"None"` as `str(None)` does, `Missing` renders `""` as Django's
`string_if_invalid` does. The msgpack codec collapses them:

```rust
// crates/djust_core/src/lib.rs — impl Serialize for Value
Value::Missing | Value::None => serializer.serialize_none(),
```

```rust
// the ValueVisitor
fn visit_unit<E>(self) -> std::result::Result<Value, E> { Ok(Value::Missing) }
```

So every `None` written to state comes back as `Missing`. Measured through
`RustLiveView.serialize_msgpack` / `deserialize_msgpack`, which is what
`SerializableViewState.state` does on EVERY read of the default
`InMemoryStateBackend` and of the Redis backend:

```
{{ p }}      p = None            django 'None'   djust 'None'  -> after one round trip: ''
{{ d.a }}    d = {"a": None}     django 'None'   djust 'None'  -> after one round trip: ''
```

The value renders correctly on the first render and renders the EMPTY STRING
after one cache hit. Nondeterminism the app author cannot explain from the
template.

## Scope

* It is the CODEC's, not any one variant's: a `None` anywhere — top level, in a
  dict, in a list, in an `Encoded`'s attribute map (#2481) — is affected.
* It predates #2481 and predates #2448. #2203 introduced the `Missing` /
  `None` split and gave `Missing | None` one serializer arm; the round trip has
  collapsed them ever since.
* Both halves are pinned in the DIVERGING direction, so a fix reddens them
  rather than passing silently:
  * `crates/djust_core/tests/test_encoded_wire_positions_2471_2472.rs::a_none_attribute_comes_back_as_missing_a_pre_existing_codec_gap`
    — which asserts BOTH the `Encoded`-attribute instance and a PLAIN
    `Value::Object` losing it, so "pre-existing" is a measurement.
  * `python/tests/test_encoded_attributes_2481.py::TestWhatThisDeliberatelyDoesNOTClose::test_a_None_attribute_is_lost_by_the_state_round_trip`

[...]

PR #2487, opened 2026-08-31
fix(templates): a Value::Encoded carries its attributes, so {{ dt.year }} resolves (#2481)

### LINKED ISSUE #2481: `Value::Encoded` carries no attributes, so `{{ dt.year }}` renders empty where Django resolves it
Found while verifying #2478's decline — filed separately per CLAUDE.md #1079, because it is the *foundation* that issue needs and it is a divergence on its own, on a far more common shape.

## The divergence

`Value::Encoded` (#2448) carries a Python object by its `type_name` / `display` / `json` / `truthy` / `sized_empty` / `iterable` spellings. It carries **no attributes**, and `Context::lookup_segment` says so in as many words:

```rust
// (2) attribute access — see the note above; a `Value` has none.
```

So every dotted lookup on a `datetime`, `date`, `time`, `timedelta` or `Decimal` resolves to nothing on the paths that have no raw-Python sidecar. Measured through `render_template_with_dirs` — the entry point `DjustTemplateBackend` itself uses (`python/djust/template/backend.py:53`):

```
case         tpl                django   djust
datetime     {{ p.year }}       2026     ''
datetime     {{ p.month }}      3        ''
datetime     {{ p.day }}        4        ''
datetime     {{ p.hour }}       5        ''
date         {{ p.year }}       2026     ''
date         {{ p.month }}      3        ''
date         {{ p.day }}        4        ''
time         {{ p.hour }}       5        ''
time         {{ p.minute }}     6        ''
timedelta    {{ p.days }}       3        ''
timedelta    {{ p.seconds }}    90       ''
Decimal      {{ p.real }}       1.5      ''
```

`{{ post.published.year }}` is an ordinary Django idiom and it renders nothing here.

## Scope, stated precisely rather than widened

* **The `DjustTemplateBackend` path is affected.** `backend.py` renders through `render_template` / `render_template_with_dirs`, neither of which attaches a sidecar.
* **The LiveView path has a partial escape.** `crates/djust_live/src/lib.rs:553` attaches a `raw_py_objects` sidecar for `getattr`-style fallback when a top-level key holds a raw Python object, so `{{ dt.year }}` may resolve *there*. That is a fallback for one path, not the rule — the same parallel-path shape as #1646, and the reason this is worth closing at the carrier rather than at one caller.
* **It predates `Value::Encoded`.** Before #2448 a `datetime` was `Value::String(str(o))`, which has no attributes either, so this is a long-standing gap that #2448 moved rather than introduced. It is filed now because #2478 cannot be closed without it.

## Why it blocks #2478

#2478 is "a falsy object WITH attributes is still truthy": the `__dict__` bulk-dump arm claims it and it becomes a non-empty `Value::Object`, whose truthiness is the mapping rule. #2466's author declined to route it through the `Encoded` carrier with a specific reason — **it would take `{{ obj.a }}` with it** — and that reason is correct, verified by running it:

```
{{ p.a }}     LenZeroWithAttrs   django 1   djust 1     (resolves today, via Value::Object)
{{ p.days }}  timedelta(days=3)  django 3   djust ''    (an Encoded has no attributes)
```

So the shape that closes #2478 without a regression is: give `Encoded` an attribute map, teach `lookup_segment` to read it, and *then* move the falsy-with-attributes object onto that carrier. Fixing this issue is the first half and is independently worth having.

## Notes for whoever takes it

* `Encoded`'s wire form is a positional msgpack tuple of six elements with a reader that accepts 6 / 4 / 3 (`crates/djust_core/src/lib.rs:440-467`). A new field goes at the END and the reader gains a 7-arm — the position rule #1541 is about.
* A `datetime`'s `.year` is not in a `__dict__` (it is a C type), so a `__dict__` bulk-dump does not reach it; the attributes have to be collected by name, or the lookup has to keep a handle to the object.
* #2479 is in flight extending `Encoded` with a `CmpKey`. Whoever lands this should expect to merge forward onto that.

Refs #2478, #2466, #2448, #1541, #1646.

johnrtipton:
Closes #2481.

## The divergence

Django's `Variable._resolve_lookup` tries three things at every dotted segment:
mapping item access, then `getattr`, then an integer index. `context::lookup_segment`
implemented steps 1 and 3, and said so in as many words:

```rust
// (2) attribute access — see the note above; a `Value` has none.
```

So every dotted lookup on a `datetime` / `date` / `time` / `timedelta` resolved
to **nothing**, on every path with no raw-Python sidecar — which is every
`DjustTemplateBackend` render:

```
case       attr           django                       djust (before)
datetime   {{ p.year }}   '2026'                       ''
datetime   {{ p.hour }}   '5'                          ''
date       {{ p.day }}    '4'                          ''
time       {{ p.minute }} '6'                          ''
timedelta  {{ p.days }}   '3'                          ''
```

`{{ post.published.year }}` is an ordinary Django idiom and rendered the empty
string. It predates the variant: before #2448 a `datetime` was
`Value::String(str(o))`, which has no attributes either.

## Reproducer, per type, before and after

Measured against **live Django 5.2.16** through BOTH entry points
`python/djust/template/backend.py:53` binds (`render_template` and
`render_template_with_dirs`), plus the LiveView path (`RustLiveView.render`).
Over the swept attribute surface — 68 `(type, attribute)` cells across
`datetime`, aware `datetime`, `date`, `time`, `timedelta` and `Decimal` —

| | before | after |
|---|---|---|
| mismatches | **64** | **41** |
| `djust REFUSES & django RENDERS` | 0 | **0** |

The 21 cells that moved are exactly the data attributes; the 41 that remain are
the two families declined below plus `Decimal`, every one pinned in the
diverging direction.

## The fix

`Value::Encoded` gains an `attrs: IndexMap<ObjectKey, Value>`, `lookup_segment`
gains step 2, and that arm is the **one** reader.

### Grep for the SINK, pin the caller SET

`lookup_segment` is reached only through `Context::get`, which is what
`Context::resolve` / `renderer::get_value_safe` / every operand channel funnel
through — verified by grepping for the sink rather than by enumerating callers.
Both counts are pinned as **equalities**, so a second dotted-path walker that
does not consult the map reddens as loudly as a deleted arm (#1125/#1646), and
`test_the_counters_go_red_in_BOTH_directions` proves each count can move in
either direction with an asserted-applied mutation.

### What the map carries is a rule, not a list

It holds what Python answers **without a call** and **without recursing**.

* `min` / `max` / `resolution` look like they belong and cannot: their values
  are values of the same family, and `datetime.min.min is datetime.min` —
  measured in `test_the_class_attributes_would_not_terminate` — so a collector
  carrying them would not terminate. `test_the_name_list_is_the_whole_of_the_policy`
  fails if they ever enter the table.
* The nullary methods (`isoformat`, `weekday`, `ctime`, `total_seconds`,
  `date`, `time`, `utcoffset`, …) are absent because Django reaches them
  through its auto-call (ADR-024), which turns a LOOKUP into an EVALUATION —
  eager at conversion time, paid whether or not a template asks, and
  inheriting whatever the call raises.

Both filed as **#2485**; `Decimal` is `Value::Decimal`, a different carrier with
no attribute slot, filed as **#2486**. All three families are pinned in the
DIVERGING direction so a fix reddens the exemption rather than passing.

### Wire — the ninth slot

```
  0 type_name    #2448      5 iterable     #2466
  1 display      #2448      6 repr         #2472
  2 json         #2448      7 cmp_key      #2471   nil, or [domain, hi, lo]
  3 truthy       #2458      8 attrs        #2481   a MAP of name -> Value
  4 sized_empty  #2466
```

Appended, written **unconditionally** as a map — an empty one costs a byte and
the slots stay aligned, the choice `cmp_key` makes one slot over and for the
same reason (#1541). The reader accepts **9 / 8 / 6 / 4 / 3**; an older width
restores NO attributes, the answer that entry was written with.

Carrying it is what makes the fix survive a cache hit:
`SerializableViewState.state` round-trips through msgpack on every read of the
default `InMemoryStateBackend`, so without the slot `{{ dt.year }}` would answer
once and go empty afterwards — the reopening `ENCODED_TAG` has now prevented
four times.

`crates/djust_core/tests/test_encoded_wire_positions_2471_2472.rs` grows the
slot-9 pins:

* a 13-shape sweep of everything the map can hold (int / float / bool / str /
  Missing / Decimal / BigInt / List / Tuple / Object / nested Encoded / empty
  list / empty object) plus an all-at-once case pinning **order**;
* the empty map, written and read as empty;
* a malformed slot reading as ABSENT rather than guessed;
* **`a_swap_of_the_key_and_attribute_slots_is_detectable`** — slots 7 and 8 are
  both "structured or nil", so a swap changes neither the width nor any type
  check. Only the values can catch it, so both are non-default in the fixture.

The pre-existing canary still reddens: `a_one_slot_shift_is_detectable` now
produces an **8-element** payload, which since #2471/#2472 is a REAL width — so
it refuses on TYPES rather than on width, which is a stronger statement than it
was, and the test says so.

### `Encoded`'s derived `PartialEq` becomes a hand-written one

The map holds `Value`s, and `Value` deliberately has no `PartialEq`: Django's
`==` for a template value is `renderer::values_equal`, which equates `1` with
`1.0` and asks `python_partial_cmp` for this family. Deriving a second `==`
onto `Value` would put a structural answer one keystroke from every site that
wants the Django one — two mechanisms for one question (#1646).

So the structural comparison is reachable **by name only**, as
`values_structurally_equal`, with every pair spelled explicitly and the
wildcard last, so `Missing` ≠ `None` (#2203), `List` ≠ `Tuple` (#2276) and
`DictView` ≠ its items (#2340) all survive.
`every_variant_is_structurally_equal_to_its_own_clone` turns a new variant
landing on the wildcard into a loud failure rather than a silent wrong answer.

## Surfaced rather than folded in (#1079)

`Value::None` and `Value::Missing` are deliberately distinct (#2203) and share
ONE msgpack `nil`, so every `None` in state comes back as `Missing` and renders
`''` after one cache hit. This is the CODEC's, not this slot's — pinned with a
plain `Value::Object` losing it too, which is what makes "pre-existing" a
measurement — and filed as **#2484**. The two naive-`tzinfo` rows are excluded
from the round-trip parametrize BY ID rather than by predicate, so a third row
acquiring a `None` value fails loudly instead of being swept in.

## Verification

* **Full suite, all three roots**: `pytest tests/ python/tests/ python/djust/tests/ -n auto`
  → **19217 passed, 420 skipped, 0 failed**.
* **Rust**: `make test-rust` → 65 test binaries, 0 failures.
* **Gate-off** (#1468, #2129/#2135) — four independent reverts, each asserting
  the mutation text matched **exactly once**, the source changed, the crate was
  **rebuilt** and the `.so` mtime advanced, `__pycache__` was cleared, and
  counting collection `errors` apart from `failed`:

  | gated off | result |
  |---|---|
  | baseline | 0 failed / 567 passed |
  | the reader (`lookup_segment` stops consulting the map) | **93 failed** |
  | the producer (no attribute is ever collected) | **96 failed** |
  | the wire WRITE (slot 9 not written) | **31 failed** |
  | the wire READ (slot 9 not read) | **25 failed** |
  | restored | **0 failed / 567 passed** |

  Four mechanisms, four independently-reachable failure sets — no two shadow
  each other (#2129).

## Files

| file | what |
|---|---|
| `crates/djust_core/src/lib.rs` | `Encoded::attrs`, `ENCODED_ATTR_NAMES`, `collect_named_attrs`, the 9th wire slot, `decode_cmp_key`, `values_structurally_equal`, hand-written `PartialEq` |
| `crates/djust_core/src/context.rs` | `lookup_segment` step 2 — the one reader |
| `crates/djust_core/tests/test_encoded_wire_positions_2471_2472.rs` | slot-9 pins + the swap canary |
| `python/tests/test_encoded_attributes_2481.py` | new — the reproducers, the declines, the structural pins |
| `python/tests/test_encoded_truthiness_2458.py` | payload width 8 → 9; the five-width read arm count |
| `python/tests/test_encoded_value_position_2471_2472_2473.py` | payload width 8 → 9; the 8-element compat read |

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,28 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- **`{{ post.published.year }}` renders `2026` instead of nothing — a `Value::Encoded` carries its attributes (#2481).** Django's `Variable._resolve_lookup` tries three things at every dotted segment: mapping item access, then `getattr`, then an integer index. `context::lookup_segment` implemented steps 1 and 3, and said so in as many words — *"attribute access — see the note above; a `Value` has none"*. So every dotted lookup on a `datetime` / `date` / `time` / `timedelta` resolved to **nothing**, on every path with no raw-Python sidecar — which is every `DjustTemplateBackend` render:
+
+  ```
+  {{ p.year }}    datetime(2026, 3, 4, 5, 6, 7)   django 2026   djust ''
+  {{ p.days }}    timedelta(days=3)               django 3      djust ''
+  ```
+
+  It predates the variant: before #2448 a `datetime` was `Value::String(str(o))`, which has no attributes either. The LiveView path had a partial escape — `crates/djust_live/src/lib.rs` attaches a `raw_py_objects` sidecar, so `{{ dt.year }}` could resolve through `getattr` *there* — and a fallback on one path is not the rule (#1646), which is why the fix is at the carrier rather than at one caller and why `TestBothPathsAgree` asserts the two now answer the same.
+
+  **21 cells, measured against live Django**, through BOTH entry points `python/djust/template/backend.py` binds. `lookup_segment` is the ONE reader of the map, pinned as an equality in both directions so a second dotted-path walker that does not consult it reddens as loudly as a deleted arm (#1125/#1646); `lookup_segment` itself has exactly one caller, pinned the same way. Over the swept attribute surface the mismatch count goes **64 → 41**, and nothing moves into `djust REFUSES & Django RENDERS`.
+
+  **What the map carries is a rule, not a list.** It holds what Python answers WITHOUT a call and WITHOUT recursing. `min` / `max` / `resolution` look like they belong and cannot: their values are values of the same family, and `datetime.min.min is datetime.min` — measured, in `test_the_class_attributes_would_not_terminate` — so a collector that carried them would not terminate. `test_the_name_list_is_the_whole_of_the_policy` fails if they ever enter the table. The nullary methods (`isoformat`, `weekday`, `ctime`, `total_seconds`, `date`, `time`, …) are absent because Django reaches them through its auto-call (ADR-024), which turns a LOOKUP into an EVALUATION — eager at conversion time, paid whether or not a template asks, and inheriting whatever the call raises. Both families are pinned in the DIVERGING direction and filed as #2485; `Decimal` is `Value::Decimal`, a different carrier with no attribute slot, and is filed as #2486.
+
+  **Wire.** The `ENCODED_TAG` payload grows to a **ninth** positional slot, appended, written unconditionally as a map — an empty one costs a byte and the slots stay aligned, which is the choice `cmp_key` makes one slot over and for the same reason (#1541). The reader accepts 9 / 8 / 6 / 4 / 3, and an older width restores NO attributes: the answer that entry was written with. Carrying it is what makes the fix survive a cache hit — `SerializableViewState.state` round-trips through msgpack on every read of the default `InMemoryStateBackend`, so without the slot `{{ dt.year }}` would answer once and go empty afterwards, the reopening `ENCODED_TAG` has now prevented four times. `crates/djust_core/tests/test_encoded_wire_positions_2471_2472.rs` grows the slot-9 pins: a 13-shape sweep of what the map can hold, order, the empty map, a malformed slot, and a key↔attrs SWAP — which neither changes the payload's width nor trips any type check, so only the values can catch it.
+
+  **`Encoded`'s derived `PartialEq` becomes a hand-written one.** The map holds `Value`s and `Value` deliberately has no `PartialEq`: Django's `==` for a template value is `renderer::values_equal`, which equates `1` with `1.0` and asks `python_partial_cmp` for this family. Deriving a second `==` onto `Value` would put a structural answer one keystroke from every site that wants the Django one — two mechanisms for one question (#1646) — so the structural comparison is reachable by NAME only, as `values_structurally_equal`, with `test_every_variant_is_structurally_equal_to_its_own_clone` so a new variant cannot land on its wildcard unnoticed.
+
+  **Surfaced rather than folded in (#1079).** `Value::None` and `Value::Missing` are deliberately distinct (#2203) and share ONE msgpack `nil`, so every `None` in state comes back as `Missing` and renders `''` after one cache hit. Pre-existing and general — pinned with a plain `Value::Object` losing it too, which is what makes "pre-existing" a measurement — and filed as #2484.
+
+  Regression coverage: 143 cases in `python/tests/test_encoded_attributes_2481.py`; new cases in `test_encoded_wire_positions_2471_2472.rs`. Gate-off verified against four independent reverts (the reader, the producer, the wire write, the wire read), each asserting the mutation matched exactly once, the source changed, the `.so` mtime advanced and `__pycache__` was cleared, and counting collection errors apart from failures.
 
 
 ### Changed
--- a/crates/djust_core/src/context.rs
+++ b/crates/djust_core/src/context.rs
@@ -64,12 +64,20 @@ pub fn template_builtin(name: &str) -> Option<Value> {
 /// runs first. Reversing these two arms passes every single-key test and
 /// fails that one.
 ///
-/// **Step 2 has no counterpart here and needs none.** A [`Value`] is inert
-/// data with no attributes; the attribute step belongs to the raw-Python
-/// sidecar walk in [`Context::resolve`], which has done all three in this
-/// order since #1997. That walk is this one's parallel path (CLAUDE.md
-/// #1646) — one had Django's rule and its twin did not, which is the drift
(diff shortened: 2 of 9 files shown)


johnrtipton:
## Wire-pin canary, measured after merge

The brief for this work asked to confirm the extended
`test_encoded_wire_positions_2471_2472.rs` still reddens on a naive shape
change. Two mutations, each asserted non-no-op, each run through
`cargo test -p djust_core --test test_encoded_wire_positions_2471_2472`
(a pure-Rust integration test, so no maturin rebuild is involved):

| shape change | result |
|---|---|
| baseline | 0 failed / 14 passed |
| **INTERIOR INSERT** — the map written at slot 1 instead of slot 8 | **6 failed** |
| **REAL skip-when-empty** — an 8-element payload when the map is empty | **0 failed** |

The second is the crossed run, and it is a finding rather than a gap. A first
attempt at that mutation was `&e.attrs` → `e.attrs.clone()`, which serializes
identically — a **semantic no-op**, so its `0 failed` measured nothing (the
failure mode v1.1.1-2 rule 3 names). Replacing it with a real conditional that
writes EIGHT elements for an empty map still measures `0 failed`, and that is
correct: a **strictly trailing** optional restores to its own default on the
way back, so the round trip is genuinely unchanged. That is #1541's own
finding, now measured for this slot rather than inherited — and it is exactly
why the interior insert is the mutation that matters, which is the one the pin
catches.

PR #2490, opened 2026-08-31
fix(templates): a falsy object WITH attributes reaches the carrier, not the __dict__ arm (#2478)

### LINKED ISSUE #2478: A falsy object WITH attributes is still truthy — the __dict__ arm's twin of #2466
Split out of #2466 per CLAUDE.md #1079 — measured on its branch, pinned there in the diverging direction, and **not** fixed there.

## The divergence

```python
class LenZeroWithAttrs:
    def __init__(self): self.a = 1
    def __len__(self):  return 0
```

```
{% if p %}T{% else %}F{% endif %}   python  False   django  F   djust  T
{{ p|length }}                              —       django  0   djust  1
{% for x in p %}[{{ x }}]{% endfor %}       —       django  ''  djust  '[a]'
{{ p }}                                     —       django  <LenZeroWithAttrs object …>   djust  {'a': 1}
```

## Why #2466's fix does not reach it

#2466 closes the class that lands on `FromPyObject for Value`'s final `Ok(Value::String(ob.str()?))` — a `set`, a `frozenset`, `complex(0)`, an empty dict view, a bare zero-`__len__` class. This object never gets there: it has a non-empty `__dict__`, so the bulk-dump arm above the fallback claims it and it becomes a **non-empty `Value::Object`**, whose truthiness is the mapping rule.

`falsy_opaque` is deliberately placed AFTER that arm. Routing an attribute-carrying object through the `Encoded` carrier would fix `{% if %}` and break `{{ obj.a }}`, which resolves today (asserted in `test_a_falsy_object_WITH_attributes_is_still_truthy`) — an `Encoded` has no attributes. So closing this needs a truthiness override on `Value::Object`, which is a different carrier and a different decision.

## Pinned

`python/tests/test_falsy_conversion_2466.py::TestWhatThisDeliberatelyDoesNOTClose::test_a_falsy_object_WITH_attributes_is_still_truthy` asserts the divergence in the DIVERGING direction and asserts that `{{ p.a }}` still resolves — so a fix that closes this and takes attribute access with it reddens the same test.

## Note on the neighbouring declines

Two more shapes are declined in #2466 for a different reason and are NOT this issue: a falsy object with `__iter__` and no `__len__`, and one with `__bool__` False and a non-zero `__len__`. Django renders their items, and the carrier cannot produce those without RUNNING the object — which consumes a generator. Those are correctly left permissive rather than made strict; this one is the only member of the class with a `Value` variant already.

Refs #2466, #2458, #1646, #1079.

johnrtipton:
Closes #2478. Built on #2487 (#2481), **already merged** (`a64174eb`) — that
PR's `Encoded::attrs` is the carrier this one needs, and the diff here is a
reorder plus one field on top of it. `origin/main` is merged into this branch,
so the diff below is #2478's alone.

## The divergence

```python
class LenZeroWithAttrs:
    def __init__(self): self.a = 1
    def __len__(self):  return 0
```

```
{% if p %}T{% else %}F{% endif %}       python False   django F    djust T
{{ p|length }}                                         django 0    djust 1
{% for x in p %}[{{ x }}]{% endfor %}                  django ''   djust '[a]'
{{ p }}                          django '<LenZeroWithAttrs object …>'   djust "{'a': 1}"
```

#2466 closed the falsiness class that lands on `FromPyObject for Value`'s final
`Ok(Value::String(ob.str()?))`. This object never got there: it has a non-empty
`__dict__`, so the bulk-dump arm ABOVE the fallback claimed it and it became a
**non-empty `Value::Object`**, whose truthiness is the mapping rule.

## Why it needed #2481 first

`falsy_opaque` was placed AFTER the `__dict__` arm deliberately, and the reason
was correct: routing an attribute-carrying object through the `Encoded` carrier
would have fixed `{% if %}` and **broken `{{ obj.a }}`**, because an `Encoded`
had no attributes. #2481 gave it an attribute map, so this PR is a **reorder
plus one field** — `falsy_opaque` moves above the `__dict__` arm and carries the
object's public `__dict__` on the carrier.

`test_falsy_conversion_2466.py`'s pinned decline is **kept and flipped to the
closing case**: it now asserts both that the divergence is gone AND that
`{{ p.a }}` still resolves. That second assertion is the regression the
previous investigator declined to risk, and it is asserted on four object
shapes × four paths (`render_template`, `render_template_with_dirs`, a nested
`{{ d.p.a }}`, and `RustLiveView`), plus across a msgpack state round trip.

## Six independent facts, not the four the issue names

Swept over **45 cells × 8 object shapes** against live Django:

| fact | carried as | cells it drives |
|---|---|---|
| truthiness | `truthy` | `{% if %}`, `not`, `and`/`or`, `{% with %}`, `{% firstof %}`, `\|yesno`, `\|default`, membership of a list |
| length | `sized_empty` | `\|length` |
| iteration | `sized_empty` / `iterable` | `{% for %}`, `{% for k,v %}`, `\|join`, `\|safeseq`, `\|escapeseq`, `\|unordered_list`, `.items`, `.keys` |
| display | `display` | `{{ p }}`, `\|default_if_none`, `\|stringformat:"s"`, `\|linebreaks`, `\|lower`, `\|striptags`, `{% cycle %}`, `\|make_list`, `\|slice` |
| repr | `repr` | `\|pprint`, `\|stringformat:"r"` |
| attributes | `attrs` | `{{ p.a }}`, `{{ d.p.a }}` |

**The issue's own suggested remedy — a truthiness override on
`Value::Object` — reaches the first row and nothing else.** Length, iteration
and display read the MAPPING, and the `__dict__` arm's whole claim is that the
object IS a mapping of its attributes. Patching one answer of a wrong carrier
value-by-value is the shape #2129 took five rounds over; moving the object to
the right carrier answers all six from spellings the struct already has.
`TestTheIssuesOwnRemedyWouldNotHaveReached` **measures** the split rather than
asserting it, and fails if every diverging cell ever becomes truthiness-only
(which would mean the argument for the carrier move had gone).

## Corpus movement

Two builds of the same 360-cell corpus (`scratch/sweep_2478.py`), refusal-collapsed:

| | before | after |
|---|---|---|
| byte-equal agreement | 131 | **252** |
| refusal-collapsed agreement | 164 | **295** |
| `django REFUSES & djust RENDERS` | 22 | **12** |
| **`djust REFUSES & django RENDERS`** | 3 | **0** |
| regressed out of agreement | — | **0** |

The refusal class **shrank**: the three were `{% for k, v in p %}`, which djust
refused where Django renders the empty branch.

**Only the four shapes the gate ADMITS moved** — 131 cells across them, 0
across the declined shape and the three controls (truthy, no attributes,
private attributes only). That is pinned rather than asserted: `PRE_FIX` is a
recorded table captured by `scratch/run_capture.py`, which REVERTS the change,
rebuilds (asserting the `.so` mtime advanced), imports the test module under
its own name so `str(o)`'s module prefix matches, records the corpus, then
restores and rebuilds.

`TestThePreFixTableIsNOTVacuous` is what stops that table being a tautology: it
asserts the claimed shapes moved ≥30 cells OFF the recorded answers (total
exactly 131) and the unclaimed shapes moved **0** — the two halves of the same
measurement, and a table quietly seeded from the current build would fail the
first.

## The gate is #2466's, unchanged

`falsy_opaque` still declines a falsy object with a non-zero `__len__` and one
that is iterable with no `__len__` — Django renders their items and this
carrier cannot produce them without RUNNING the object.
`test_a_declined_shape_still_keeps_its_mapping` asserts the sharp end: such an
object still arrives as a `Value::Object` and `{% for %}` still iterates its
ATTRIBUTES where Django renders `10` and `20`. A divergence, #2466's recorded
decline, neither introduced nor closed here.

## Security — the arm now dumps a `__dict__`, so its ORDER is a boundary

`__djust_serialize__` (#1986) and the raw-`Model` arm (#1986 vector 7) stay
ABOVE `falsy_opaque`, so a Django model cannot reach it and cannot have its
denylisted fields collected into the attribute map.
`TestTheSerializationFloorsStayAbove` asserts the source order and carries a
canary proving that check can go red.

The `_`-prefix filter is stated **once**, in a shared `public_dict_attrs` with
exactly two callers — the `__dict__` arm and `falsy_opaque`. Two copies of that
filter is the #1646 shape, and this arm's copy would be the one that leaks; the
caller count is pinned as an equality.

One cell's divergence changes SHAPE rather than closing, and is recorded:
`|json_script` over these objects now emits `str(o)` where it emitted a JSON
object of the attribute VALUES, so strictly less of the object reaches the page.
Django refuses either way (#2429's declined refusal direction).

## Verification

* **Full suite, all three roots**: `pytest tests/ python/tests/ python/djust/tests/ -n auto`
  → **19588 passed, 432 skipped, 0 failed**.
* **Rust**: `make test-rust` → 65 test binaries, 0 failures.
* **Gate-off** — each mutation asserted non-no-op, crate rebuilt with the `.so`
  mtime checked, `__pycache__` cleared, collection `errors` counted apart from
  `failed`. **Two mechanisms, two independent reverts**, so neither can shadow
  the other (#2129/#2135):

  | gated off | result |
  |---|---|
  | baseline | 0 failed / 653 passed |
  | mechanism 1 — the ORDER (the `__dict__` arm claims it first again) | **143 failed** |
  | mechanism 2 — the attribute MAP (the carrier claims it, carrying nothing) | **23 failed** |
  | the shared `_`-prefix collector | **2 failed** |
  | restored | 0 failed |

  Mechanism 2's 23 is the interesting one: it is the set of tests that go red
  when ONLY the attribute half is removed — i.e. `{{ obj.a }}` and its
  round-trip siblings. Without it the reorder alone would pass 630 of 653 and
  look correct, which is exactly the shadowing #2129 warns about.

  **The `_`-prefix collector's first measurement was 1 failed — and it was the
  SOURCE-GREP pin, not a behaviour.** A surviving-but-for-a-grep mutation is a
  question, not a pass. Traced: on `falsy_opaque`'s arm the map is
  `Encoded::attrs`, whose only reader is `lookup_segment`, and a `_`-leading
  path segment is refused before the lookup runs — so the filter there is
  defence-in-depth rather than currently-observable. It IS observable on the
  `__dict__` arm, where the map becomes the rendered value, and that had no
  test. Added (`test_the_underscore_filter_is_observable_on_the_dict_arm`, a
  truthy object with a `_secret`), and re-measured: **2 failed**. The reasoning
  is written into the test's own docstring rather than into this PR alone.

## Files

| file | what |
|---|---|
| `crates/djust_core/src/lib.rs` | `public_dict_attrs` extracted and shared; `falsy_opaque` moved above the `__dict__` arm and carrying the map |
| `python/tests/test_falsy_with_attributes_2478.py` | new — 45 cells × 8 shapes, the recorded pre-fix table, the remedy measurement, the floor-order pin |
| `python/tests/test_falsy_conversion_2466.py` | the pinned decline flipped to the CLOSING case, still asserting `{{ p.a }}` |

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,28 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- **`{{ post.published.year }}` renders `2026` instead of nothing — a `Value::Encoded` carries its attributes (#2481).** Django's `Variable._resolve_lookup` tries three things at every dotted segment: mapping item access, then `getattr`, then an integer index. `context::lookup_segment` implemented steps 1 and 3, and said so in as many words — *"attribute access — see the note above; a `Value` has none"*. So every dotted lookup on a `datetime` / `date` / `time` / `timedelta` resolved to **nothing**, on every path with no raw-Python sidecar — which is every `DjustTemplateBackend` render:
+
+  ```
+  {{ p.year }}    datetime(2026, 3, 4, 5, 6, 7)   django 2026   djust ''
+  {{ p.days }}    timedelta(days=3)               django 3      djust ''
+  ```
+
+  It predates the variant: before #2448 a `datetime` was `Value::String(str(o))`, which has no attributes either. The LiveView path had a partial escape — `crates/djust_live/src/lib.rs` attaches a `raw_py_objects` sidecar, so `{{ dt.year }}` could resolve through `getattr` *there* — and a fallback on one path is not the rule (#1646), which is why the fix is at the carrier rather than at one caller and why `TestBothPathsAgree` asserts the two now answer the same.
+
+  **21 cells, measured against live Django**, through BOTH entry points `python/djust/template/backend.py` binds. `lookup_segment` is the ONE reader of the map, pinned as an equality in both directions so a second dotted-path walker that does not consult it reddens as loudly as a deleted arm (#1125/#1646); `lookup_segment` itself has exactly one caller, pinned the same way. Over the swept attribute surface the mismatch count goes **64 → 41**, and nothing moves into `djust REFUSES & Django RENDERS`.
+
+  **What the map carries is a rule, not a list.** It holds what Python answers WITHOUT a call and WITHOUT recursing. `min` / `max` / `resolution` look like they belong and cannot: their values are values of the same family, and `datetime.min.min is datetime.min` — measured, in `test_the_class_attributes_would_not_terminate` — so a collector that carried them would not terminate. `test_the_name_list_is_the_whole_of_the_policy` fails if they ever enter the table. The nullary methods (`isoformat`, `weekday`, `ctime`, `total_seconds`, `date`, `time`, …) are absent because Django reaches them through its auto-call (ADR-024), which turns a LOOKUP into an EVALUATION — eager at conversion time, paid whether or not a template asks, and inheriting whatever the call raises. Both families are pinned in the DIVERGING direction and filed as #2485; `Decimal` is `Value::Decimal`, a different carrier with no attribute slot, and is filed as #2486.
+
+  **Wire.** The `ENCODED_TAG` payload grows to a **ninth** positional slot, appended, written unconditionally as a map — an empty one costs a byte and the slots stay aligned, which is the choice `cmp_key` makes one slot over and for the same reason (#1541). The reader accepts 9 / 8 / 6 / 4 / 3, and an older width restores NO attributes: the answer that entry was written with. Carrying it is what makes the fix survive a cache hit — `SerializableViewState.state` round-trips through msgpack on every read of the default `InMemoryStateBackend`, so without the slot `{{ dt.year }}` would answer once and go empty afterwards, the reopening `ENCODED_TAG` has now prevented four times. `crates/djust_core/tests/test_encoded_wire_positions_2471_2472.rs` grows the slot-9 pins: a 13-shape sweep of what the map can hold, order, the empty map, a malformed slot, and a key↔attrs SWAP — which neither changes the payload's width nor trips any type check, so only the values can catch it.
+
+  **`Encoded`'s derived `PartialEq` becomes a hand-written one.** The map holds `Value`s and `Value` deliberately has no `PartialEq`: Django's `==` for a template value is `renderer::values_equal`, which equates `1` with `1.0` and asks `python_partial_cmp` for this family. Deriving a second `==` onto `Value` would put a structural answer one keystroke from every site that wants the Django one — two mechanisms for one question (#1646) — so the structural comparison is reachable by NAME only, as `values_structurally_equal`, with `test_every_variant_is_structurally_equal_to_its_own_clone` so a new variant cannot land on its wildcard unnoticed.
+
+  **Surfaced rather than folded in (#1079).** `Value::None` and `Value::Missing` are deliberately distinct (#2203) and share ONE msgpack `nil`, so every `None` in state comes back as `Missing` and renders `''` after one cache hit. Pre-existing and general — pinned with a plain `Value::Object` losing it too, which is what makes "pre-existing" a measurement — and filed as #2484.
+
+  Regression coverage: 143 cases in `python/tests/test_encoded_attributes_2481.py`; new cases in `test_encoded_wire_positions_2471_2472.rs`. Gate-off verified against four independent reverts (the reader, the producer, the wire write, the wire read), each asserting the mutation matched exactly once, the source changed, the `.so` mtime advanced and `__pycache__` was cleared, and counting collection errors apart from failures.
 
 
 ### Changed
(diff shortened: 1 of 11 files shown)


johnrtipton:
## Gate-off note on the corpus figures

Two numbers in the body deserve their provenance stated, because both are the
kind that can be right for the wrong reason:

**`131 cells moved` is measured twice, from two different harnesses, and they
agree.** `scratch/sweep_2478.py` (run as `__main__` against two builds) and
`scratch/run_capture.py` (which reverts the change, rebuilds, and imports the
TEST module under its own name) both report 131 — 33/32/33/33 across the four
claimed shapes. The second harness exists because `str(o)` carries the class's
`__module__`, so a table captured in one harness cannot be compared in another
without a normalisation that would also hide real movement.

**One cell is exempt from the recorded table, and the reason is the cell's own
shape.** `|make_list` splits `str(o)` into characters, so an object whose
`str()` carries a memory address arrives as a list of the address's DIGITS —
which no whole-string normalisation can reach. It stays in `CELLS` and its
Django parity is asserted in-process; only the cross-process comparison is
exempt, and the exemption names the one cell rather than filtering by a
predicate that could absorb a second (`PRE_FIX_UNASSERTABLE`).

Issue #2482, opened 2026-08-31
The differential corpus cannot hold an unpicklable or non-literal row, so two shapes of #2466's class are unsweepable


johnrtipton:
Split out of #2477's third section, which the axis PR could not close: the harness limit is a harness change, not a corpus one.

## Two shapes cannot be corpus rows at all

`INPUTS` is not just read — it is **deep-copied** and **re-evaluated as literals**, by two different consumers, and each rules out a class of value:

1. **`measure`'s `@cmp` axis deep-copies the second operand** (`scripts/filter-parity-differential.py`, the `cmp_cells()` loop): `{"p": INPUTS[ka], "q": copy.deepcopy(INPUTS[kb])}`. The deep copy is load-bearing and correct — two structurally-equal operands that are not the same object is the whole of what `values_equal` / `try_compare` needed (#2335). But a dict VIEW cannot survive it:

   ```
   dict_keys    deepcopy FAILS: TypeError: cannot pickle 'dict_keys' object
   dict_values  deepcopy FAILS: TypeError: cannot pickle 'dict_values' object
   dict_items   deepcopy FAILS: TypeError: cannot pickle 'dict_items' object
   ```

   `test_json_script_ensure_ascii_and_element_id_2413.py` deep-copies each `INPUTS` value too, so the same row would break there.

2. **`test_sequence_op_chokepoint_2451.py::corpus()` re-evaluates each value from the AST**, compiling every value node in a small namespace. That works for a literal and for a call to a name it binds; it cannot produce a **user-defined class instance**. So the falsy-and-unmodelled shapes that are not builtins — a zero-`__len__` class, a `__bool__`-False class, a falsy object with `__iter__` and no `__len__` — cannot be corpus rows either.

## What it costs, measurably

The `value-truthiness` axis now names both no-variant conversion arms (#2477). Two of its four new members are exempt with exactly these reasons:

* `{value,arg}:str-fallback:falsy` — a falsy object reaching the terminal `Value::String(ob.str()?)` is one of the two shapes #2466 declined, and **no builtin type has either**, so the inhabitant would have to be a class instance. Exempt because of (2), not because it is uninhabitable.

And the dict-view half of #2466's class — an empty `dict_keys` / `dict_values` / `dict_items` — has no row at all, because of (1). It is swept nowhere: not on the `@cmp` axis it breaks, and not on the twenty-odd axes it would be fine on.

## The shape

[...]

Issue #2480, opened 2026-08-30
Two falsy_opaque Value::Encoded values never compare equal — {% if p == q %} on two set()s regressed from True to False in #2476


johnrtipton:
Found while merging #2479 (#2471/#2472/#2473) forward over #2476. Filed rather
than folded in (#1079): the fix needs something the conversion does not
currently carry, and #2479's comparison arm answers it exactly as the wildcard
it replaced did.

## The measurement

`main` at `303c83a8` (the #2476 merge), raw `render_template` path:

| template | `p`, `q` | Django | djust |
|---|---|---|---|
| `{% if p == q %}` | the SAME `set()` | `Y` | **`N`** |
| `{% if p == q %}` | the SAME `frozenset()` | `Y` | **`N`** |
| `{% if p == q %}` | the SAME `complex(0)` | `Y` | **`N`** |
| `{% if p == q %}` | the SAME zero-`__len__` instance | `Y` | **`N`** |
| `{% if p == q %}` | the SAME `__bool__`-False instance | `Y` | **`N`** |
| `{% if p != q %}` | any of the above | `N` | **`Y`** |
| `{% if p <= q %}` | `set()` / `frozenset()` | `Y` | **`N`** |

`<` and `>` AGREE for all of them (both `N`), which is correct — Python cannot
order these either, and `set`'s `<` on two equal sets is False.

## Why it is a regression and not a pre-existing gap

Before #2476 a `set()` had no `Value` variant, so the conversion landed it on
its final `Ok(Value::String(ob.str()?))` and it arrived as the string
`"set()"`. Two of those compared **equal** through `values_equal`'s
`(String, String)` arm — Django's answer, reached by accident.

#2476 made it a `Value::Encoded` (via `falsy_opaque`) so `bool(set())` could be
`False`. That is the right fix for truthiness, and it moved these values off
the `String` arm onto the `_ => false` wildcard, which is where the equality
went.

## Why #2479 does not close it, and does not worsen it

#2479 adds an `(Encoded, Encoded)` arm reading a carried `CmpKey`.
`falsy_opaque` sets `cmp_key: None` — deliberately — so
`Encoded::python_partial_cmp` short-circuits and answers `None`: never equal,
never ordered. Byte for byte the wildcard's answer.

That is asserted rather than argued: #2479's gate-off mutation M1 reverts
`values_equal`'s arm to a literal `false`, and every case in
`TestAFalsyOpaqueEncodedIsNotComparable` stays green under it.

[...]

PR #2494, opened 2026-08-31
fix(templates): a Python collection crosses as its ITEMS, not its repr (#2477, #2489)

### LINKED ISSUE #2477: normalize_django_value turns a set into a list, so subscript-refusing filters render where Django raises
Split out of #2466 per CLAUDE.md #1079 — found while trying to add a `set` to the differential corpus, and **not** touched there.

## The divergence

`normalize_django_value` normalises a `set` to a sorted **list** on the LiveView path. A list is subscriptable; a set is not. So every filter whose Django body indexes its value renders here and raises there:

```
{{ p|first }}   p = set()                django  <<TypeError: 'set' object is not subscriptable>>
                                         djust   ''
{{ p|first }}   p = {"<img src=x …>"}    django  <<TypeError: 'set' object is not subscriptable>>
                                         djust   &lt;img src=x onerror=alert(1)&gt;
{{ p|last }}    (both shapes)            same pair
```

`first` / `last` / `random` are the three #2451 identified as `value[i]` under `except IndexError` alone — a `TypeError` is not caught, so Django raises.

## Why this is NOT #2466's cell

#2466 fixes the **raw `DjustTemplateBackend`** path, where a set now arrives as a `Value::Encoded` and `python_getitem` refuses it correctly:

```
{{ p|first }}   p = set()   django  TypeError: 'set' object is not subscriptable
                            djust   filter 'first' raises TypeError: 'set' object is not subscriptable   ✓
```

These four cells are the **LiveView** path, where the value never reaches the conversion as a set at all — `normalize_django_value` has already made it a list. Fixing it inside the filters would be a guard at the consumer for a defect at the normaliser.

## How it was found, and the corpus consequence

Adding `set-empty` / `set-plain` to `INPUTS` in `scripts/filter-parity-differential.py` reddens, on a single build:

* `test_sequence_op_chokepoint_2451.py::TestTheReferenceTableIsRunNotTranscribed::test_no_cell_renders_where_django_refuses` — the four cells above;
* the same file's `test_the_sweep_is_not_vacuous` per-filter counts (`first` 16 → 18, `last` 19 → 21, `phone2numeric` 30 → 32);
* `test_refusal_collapsed_agreement_2454.py::test_the_payload_gains_no_definition_marker`;
* `test_json_script_ensure_ascii_and_element_id_2413.py::test_it_is_the_ONLY_json_script_cell_left_on_the_un_masked_surface`;
* `test_differential_reachability_manifest_2345.py::TestTheArgumentAxisCorpus` (four methods) — the `diverging == {"d-typed-key"}` set grows.

So the corpus addition is not a one-line change: it is gated on this defect being decided first. Reverted in #2466's branch rather than folded in.

## Also blocking a corpus row for the dict-view half

An empty `dict_keys` is a member of #2466's class and **cannot** be a corpus row as the harness stands: `measure` does `copy.deepcopy(INPUTS[kb])` for the `q` operand and a view is unpicklable (`TypeError: cannot pickle 'dict_keys' object`). A `set` row would need the normaliser question answered; a view row needs the harness to build its operands lazily.

## And the axis that is blind to the class

#2469's `value-truthiness` axis reads the `Value` **variants** out of `crates/djust_core/src/lib.rs` and requires a falsy and a truthy inhabitant of each. Every value #2466 is about has **no variant** — that is the defect — so the axis reports `0 MISSING` while being structurally unable to see them. Worth extending with a "no-variant fallback" slot, so the next fix at the conversion is measurable rather than only testable.

Refs #2466, #2469, #2451, #1079.

### LINKED ISSUE #2489: A dict view and a falsy-iterable cross as their str(), so every iterating/sizing/subscripting consumer reads the REPR — the two shapes #2466 declined
Surfaced by #2482, which put the first inhabitants of this shape in the differential corpus. #2466 closed the FALSY-and-carriable half of the class and explicitly **declined** two shapes; this is what those two shapes do on a page, measured rather than reasoned about.

## The class

`impl FromPyObject for Value`'s fallback block ends in

```rust
// crates/djust_core/src/lib.rs
Ok(Value::String(ob.str()?.to_string()))
```

so an object no variant models, that `falsy_opaque` declines, arrives at the renderer as its **`str()`** — a plain string. Every consumer that iterates, sizes, subscripts or slices then reads the **repr**, one character at a time, while Django reads the object.

Two shapes reach it, and neither is exotic:

1. **A non-empty dict view.** `{'k': 1}.keys()` is truthy, so `falsy_opaque`'s own `if ob.is_truthy() { return None }` gate turns it away. `{% for k in d.keys %}`-style code that hands a view to the context is ordinary Django.
2. **A falsy object with `__iter__` and no `__len__`** — `falsy_opaque`'s doc-comment names it as DECLINED, because the carrier cannot produce the items without RUNNING the object. A lazy queryset-ish wrapper, a `__bool__`-False collection proxy.

## Measured

Corpus rows (`scripts/filter-parity-differential.py`, `INPUTS_LAZY`):

* `dv-keys-plain` — `{"": 1}.keys()`
* `o-falsy-iter` — a class with `__bool__` → False, `__iter__` yielding one payload string, no `__len__`, `str()` pinned to `FalsyIterable()`

Both engines, same value, `_rust.render_template` (the RAW entry point — no normalizer involved; the LiveView path answers identically, which is what separates this from #2477):

| template | value | Django | djust |
|---|---|---|---|
| `{% for x in p %}[{{ x }}]{% endfor %}` | `dv-keys-plain` | `[&lt;img src=x onerror=alert(1)&gt;]` | `[d][i][c][t][_][k][e][y][s][(][[]…` — 43 cells, one per repr character |
| `{{ p\|escapeseq }}` | `o-falsy-iter` | `[&#x27;&amp;lt;img src=x…&#x27;]` | `[&#x27;F&#x27;, &#x27;a&#x27;, &#x27;l&#x27;, &#x27;s&#x27;, …]` |
| `{{ p\|safeseq }}` | `o-falsy-iter` | `[&#x27;&lt;img src=x…&#x27;]` | the repr, per character |
| `{{ p\|unordered_list }}` | `o-falsy-iter` | `\t<li>&lt;img src=x…</li>` | one `<li>` per repr character |
| `{{ p\|length }}` | `o-falsy-iter` | `0` | `15` — the length of `"FalsyIterable()"` |
| `{{ p\|slice:":3" }}` | `o-falsy-iter` | `FalsyIterable()` (Django's `slice` returns the value when it cannot slice) | `Fal` |
| `{{ p\|first }}` | `dv-keys-plain` | `<<TypeError: 'dict_keys' object is not subscriptable>>` | `d` |
| `{{ p\|last }}` | `dv-keys-plain` | same TypeError | `)` |
| `{{ p\|phone2numeric }}` | `dv-keys-plain` | `<<AttributeError: 'dict_keys' object has no attribute 'lower'>>` | `3428_5397([&#x27;&lt;464 772=9 6637767=25378(1)&gt;&#x27;])` |

`{{ p|length }}` being `15` where Django says `0` is the sharpest one: it is silently wrong rather than visibly broken, and it is the kind of value a template branches on.

**Not a leak.** Scanning all 3,562 new cells that carry a payload row: **0** gained a live fragment (an unescaped tag opener), 7 gained one inside fully-escaped text, and 421 are cells where Django itself raises so there is no output to be more permissive than. djust emits the *repr* — which does not contain the payload for `o-falsy-iter`, and carries it escaped for `dv-keys-plain`. This is a correctness divergence, not an XSS.

## Why #2466 declined it, and what changed

#2466's `falsy_opaque` doc-comment is explicit: for `len(o)` raising while `iter(o)` succeeds, "this carrier cannot produce those items without RUNNING the object, and running it would consume a generator." That reasoning is sound for the carrier as designed — but the current behaviour is not "we decline to carry it", it is "we carry it as text and every consumer silently reads the text". Those are different failure modes and the second one renders wrong output rather than refusing.

Options worth weighing (the issue's own list is a hypothesis, not a constraint):

* Carry `iter(o)` eagerly for objects that are not iterators (`iter(o) is not o` — a `dict_keys` is re-iterable, a generator is not), and decline only the genuinely one-shot case. That covers both shapes above without consuming anything.
* Give the terminal `str()` arm a marker so the sizing/iterating/subscripting consumers REFUSE rather than reading the repr — matching Django for `first`/`last`/`phone2numeric` while still not matching it for `{% for %}`.
* Leave it, and pin the divergence. That is what #2482 did as a stopgap: `STRINGIFIED_AT_CONVERSION` in `python/tests/test_sequence_op_chokepoint_2451.py` records the six `first`/`last`/`phone2numeric` cells exactly, in both directions, with a non-vacuity test asserting the RAW path answers identically to the LiveView path (which is what distinguishes this from #2477's normalizer half).

## Related

* #2466 — closed the falsy-and-carriable half; declined these two shapes.
* #2477 — the LiveView-path twin: `normalize_django_value` flattens the same objects in Python before the conversion sees them.
* #2482 — the harness change that made these rows representable at all.

johnrtipton:
Closes #2477. Closes #2489.

One defect at two severities: `impl FromPyObject for Value`'s fallback block ends in `Ok(Value::String(ob.str()?))`, so a Python collection no variant models arrived at the renderer as its **`str()`** — and every consumer that iterates, sizes, subscripts or slices then read the **repr, one character at a time**.

```
                       value                django                 djust
{{ p|length }}         {}.keys()            0                      13
{{ p|length }}         a falsy __iter__     0                      15
{% for x in p %}       {"k": 1}.keys()      [k]                    one cell per repr char
{{ p|escapeseq }}      a falsy __iter__     ['&lt;img …']          ['F', 'a', 'l', 's', …]
{{ p|first }}          {}.keys()            <<TypeError>>          d
{{ p|phone2numeric }}  {"…": 1}.keys()      <<AttributeError>>     dials the repr
```

`normalize_django_value` flattened the same class **before the conversion saw it**, differently — a `set` became a sorted list, everything else took `str()` — so the LiveView path and the raw `render_template` path answered differently for the same value. Both halves are here, because either alone is invisible on a page.

## Reproducers, per path

`scratch/coll/repro.py` renders 25 shapes × 16 consumers in **three columns** — Django (called, never transcribed), the raw entry point, and the LiveView path through the normalizer — and `scratch/coll/cmp.py` diffs two builds cell by cell. The same sweep is a test: `python/tests/test_opaque_collections_2477_2489.py`.

## The class, enumerated, with a decision per member

| group | members | decision |
|---|---|---|
| CARRIED | `set` / `frozenset` (empty and not), all three dict views, `odict_keys`, a `mappingproxy`, a sized `__iter__` class, a falsy `__iter__` class with no `__len__`, a falsy class with `__len__` 2, an iterable class WITH attributes, a zero-`__len__` class, a `__bool__`-False class, `complex(0)`, `complex(1)` | 18 shapes, `Value::Encoded` |
| DECLINED | a one-shot iterator (`iter(o) is o`), a falsy one, an unbounded re-iterable past `OPAQUE_ITEM_CAP`, a TRUTHY non-iterable object with public attributes | 4 shapes, keeps the string path |
| EARLIER ARM | `bytes`, `deque`, `range`, a `__getitem__` sequence, `Counter`, a plain dict, a plain list | 7 shapes, untouched |

Each is asserted in **both** directions through `_rust.crosses_as_encoded`, so a member that stops being carried fails and one that starts being carried fails too. Each decline is pinned in the DIVERGING direction, and the one-shot decline is paired with the assertion that justifies it rather than restates it: after a full render the generator still yields everything it was going to.

`complex(1)` and the truthy attribute-less shapes are a **widening beyond either issue's measured table**, argued rather than assumed: splitting the class by truthiness once was the mistake this corrects, and splitting it by "does it happen to have a `__dict__`" would be the same mistake one axis over. The `__dict__` arm is declined only because retiring it is a genuinely larger change.

## The carrier: `Encoded`, argued

`Encoded` gains `items` (`list(o)`) and its `sized_empty: bool` widens to `len: Option<usize>` — Django's `|length` reads `__len__` under `except TypeError: return 0` while its iterating filters call `iter()`, so a falsy `__iter__` class with no `__len__` answers **0** and **one item** at once, which a bit cannot carry.

A collection needs seven facts the struct already measures, none derivable from a list of items: `display` (`{'a'}` vs `['a']`), `type_name` (the refusal must name `'set'`), `truthy` (a `__bool__`-False collection with two items), `len`, `repr`, `attrs`. `Value::DictView` — the one existing variant with a collection's shape — is built only by `Context::dict_view` during a render, has no wire format, and derives truthiness from `!items.is_empty()`, false for two of the objects this carries. Splitting the class by emptiness so `set()` took one carrier and `{'a'}` another is the drift shape (#1646).

## The predicate took three shapes; two mistakes were found by running it

1. **Transcribing the fallback block's last two arms** answers *"would the fallback claim this IF it got there"*. A `bytes` and a `deque` satisfy the gate in isolation but are claimed by PyO3's sequence extraction long before — so the normalizer stopped stringifying them and `{{ p }}` over `b"ab"` regressed from `b'ab'` to `[97, 98]`. **Six regressed cells**, caught by the before/after sweep.
2. **Running the real conversion** is exact and **segfaulted**: the normalizer's fallback is where an ordinary presenter object lands, and converting one eagerly walks its `__dict__` into a raw `QuerySet` and `Manager` and down through theirs, deep enough to overflow the stack — work the render path never does. Caught by `test_template_auto_call_1985.py::test_non_model_intermediary_does_not_leak`, which killed a worker rather than failing.

What stands is neither: the gate is split out as `opaque_gate`, which measures `bool(o)`, `iter(o)`, `len(o)` and the `__dict__` KEYS and converts nothing — one statement, two consumers. `crosses_as_encoded_by_conversion` is exported alongside so the cheap probe has a **differential against the thing it stands in for**; it immediately reported the probe answering `true` for `None`, `True`, `7`, `1.5`, `"ab"` and a `Decimal` — six shapes its one caller can never send it. A predicate whose correctness depends on which caller it has is one the next caller breaks.

## Wire

**It lands beside #2495 (#2485), and the two compose because they are widenings of different KINDS.** #2495 grew the attribute MAP at slot 8 and added no position; this widens slot 4 and appends slot 9. Ten, not eleven. The arithmetic is the weakest part of that sentence, so it is not what the pins check: the merged serializer was read (ten fields, ten slots, in order) and a real `serialize_msgpack` dumped for six shapes — slot 8 is still the attrs map with #2495's `total_seconds` INSIDE it, slot 9 is the items (`['a','b']` for a set, `None` for a `timedelta`), slot 4 is the count (`2`, `0`, `None`).

Two canaries were added for it, because the existing shift pin only covers a REMOVE (shifts down) and swaps: `an_interior_insert_is_refused_rather_than_silently_misread` inserts a plausible string at every one of the ten interior positions and requires each to be refused as a plain dict, and `the_slot_that_grew_inside_itself_did_not_take_a_position` grows the map by four names and re-reads the width and the trailing slot. Had #2495 taken a position of its own, the width would still have looked plausible while `attrs` decoded as `repr` and `items` as `cmp_key`.

Ten slots: slot 5 widens from `sized_empty` to `len(o)`, slot 10 appends the items. Slot 5 is the one slot whose TYPE changes, safe for the reason the #1541 canon is actually about — that canon forbids a conditionally-skipped field, which shifts later slots WITHIN one width, while this payload is dispatched on width and no build wrote a 10-element one. The reader accepts **10 / 9 / 8 / 6 / 4 / 3** and each narrower width restores what the entry was WRITTEN with, never a guess:

| width | era | what it restores |
|---|---|---|
| 10 | this fix | everything |
| 9 | #2481 | `len` from slot 4's BOOLEAN (`true → Some(0)`, `false → None`, exact — the pre-fix gate declined every object whose `len` was not 0); no items |
| 8 | #2471/#2472 | + no attributes |
| 6 | #2466 | + `repr` falls back to `display`, no comparison key |
| 4 | #2458 | + `len` `None`, not iterable |
| 3 | #2448 | + truthiness derived as `!display.is_empty()` |

5, 7 and 11 are widths this crate never wrote and are refused as a plain dict, which is what keeps a user dict under the tag from forging one. `into_pyobject` returns the ITEMS for a carried collection, so a `set` in view state still comes back a list.

## Corpus movement

Two builds of `scripts/filter-parity-differential.py` over **380,484 cells** — `origin/main` against this branch, `--compare`d:

```
agree BEFORE : 277323   (refusal-collapsed: 324275)
agree AFTER  : 277729   (refusal-collapsed: 324951)
django REFUSES & djust RENDERS: 4992 -> 4722   (+270; djust more permissive)
djust REFUSES & Django RENDERS: 39253 -> 39253   (+0; djust stricter)
live-payload leaks: 65 -> 60   (5 closed, 0 INTRODUCED, 0 live)
REGRESSIONS : 12
```

The raw headline is blind to refusal-class movement, so both are reported: the permissive column SHRANK by 270 and the strict column did **not grow by a single cell**. Twelve cells moved the wrong way and all twelve are named — ten are the `@cmp` widening below, and two are `{{ p|timesince:obj }}` / `{{ p|timeuntil:obj }}`, where Django raises `AttributeError` on the operand and djust rendered a duration before this fix and renders a different one after: a cell inside the column that shrank, diverging before and after.

Three of those regressions were found this way and fixed rather than filed, and each was a case where the OLD answer was right by accident. `{% regroup tags by k %}` over a `set` built ZERO groups where Django builds one — the operand reaches the handler through `value_channel_arg_string`, and an `Encoded` fell to the general encoder's `_` arm, so the handler got the text `{'a'}`, which is neither JSON nor a variable name. Before the carrier existed a set crossed as a `Value::String`, which that channel JSON-quotes, so the handler decoded a string and iterated its CHARACTERS. The arm belongs in the VALUE channel and not the general one, and putting it in the general one first is how the split was found: a custom tag RENDERS its argument, where Django shows `{'a'}` and the items show `["a"]` — 21 `@ctag` cells the wrong way. `{% if tag in tags %}` answered `N`, and it had worked by a worse accident: `in` over a `Value::String` is a SUBSTRING match, so `'a' in {'ab'}` was true and so was any character of the repr's punctuation. And `into_pyobject` goes BACK to the display — handing the items back looked conservative, but measured against `main` the premise was false (a truthy set was declined by the pre-#2477 gate and already came back as `str(o)`), and what the items DID change was 20 custom-filter cells.

And the focused sweep, 25 shapes × 16 consumers (400 cells, three columns): raw-path divergences **192 → 44**, LiveView **266 → 67**, **0 cells regressed** in either column. Restricted to the 19 shapes the carrier claims it is raw **153 → 5** and LiveView **204 → 5**, with the two paths AGREEING for every one of them. The five survivors are refuse-vs-refuse with a different exception CLASS. Everything else remaining is pre-existing and untouched: `range` / `bytes` / `deque` / a `__getitem__` sequence are claimed by an earlier arm, a `Counter` is a dict, a generator is the decline.

## Pins removed, on their own stated terms

`NORMALIZER_FLATTENED` (#2477) and `STRINGIFIED_AT_CONVERSION` (#2482), thirteen cells between them, all now agreeing — so `test_no_cell_renders_where_django_refuses` subtracts **nothing**, which is stronger than any exemption list. #2382's `RESIDUE` / `ITERABLE_RESIDUE` and #2366's `STRINGIFIED_AT_EXTRACTION` empty the same way, each keeping a PARITY row for every name that left plus a test that the list is empty AND its rows arrived somewhere. The decline pins in #2466 and the `LenTwoBoolFalseWithAttrs` row in #2478 keep their cells with the assertions INVERTED.

`value-truthiness` loses `("opaque_value", "truthy")` — whose stated reason was true of the code and was the defect — and gains one for `str-fallback`, whose remaining inhabitants a corpus row structurally cannot be.

## What it costs (#2480)

An `Encoded` from this arm carries no comparison key, so `{% if p == q %}` over two equal sets is `N` where Django says `Y`. #2466 already did that to the falsy half and filed #2480; this widens it from **four shapes to eight**. As `Value::String` these compared by TEXT and got the right answer for the wrong reason — the same mechanism that made `{{ p|length }}` count repr characters, so the accident and the defect cannot be separated. None of the sixteen consumers in the sweep is a comparison; this was found by a pin going red, not by the sweep. It has its own class, `TestTheComparisonAxisThisWIDENS`, with both halves counted and pinned in the DIVERGING direction.

## Gate-off

`scratch/coll/gateoff.py`. Each row asserts the mutation text was **found** and matched **exactly once**, that the mutated source **differs**, that the crate was **rebuilt and the `.so` mtime advanced**, with `__pycache__` cleared and `N error` counted apart from `N failed` — a run whose module did not import reports INVALID rather than a number. Tests: the nine files this change touches.

| mutation | mechanism removed | red |
|---|---|---|
| `if not state_roundtrip: return value` → `if False` | the normalizer stops carrying a `set` | **6** |
| `if _rust.crosses_as_encoded(value)` → `if False and …` | the normalizer stringifies what the conversion models | **11** |
| `let items = if iterable {` → never enumerate | the collection half | **15** |
| `ob.len().ok()` → `.filter(\|_\| false)` | the counting half | **28** |
| `if truthy { return None }` re-added | the pre-#2477 falsy-only gate | **24** |
| `if it.as_any().is(ob)` → `if false && …` | the one-shot decline | **6** |

Every one is non-zero, and the mechanisms are independently reachable rather than shadowing each other: the two Python rows redden with the Rust carrier intact and the four Rust rows redden with the normalizer intact, `items` and `len` redden separately (15 vs 28 — the two facts a single `sized_empty` bit could not carry), and the one-shot decline has its own row, so "the gate declines a generator" is not covered by the enumeration row.

**The harness grew a rule from an incident rather than a principle.** It was interrupted mid-cycle, which left a mutated `lib.rs` on disk AND left cargo's fingerprint saying the tree was built — so the next `make build` exited 0 in a second and the installed `.so` was still the mutant. Every measurement after that point was the mutant's: `{{ p|length }}` answered `0` for a non-empty set while the source read `e.len`, and my first `if false` grep over the diff reported all-clear because *that* mutation is a `.filter(|_| false)`. Three things changed: the restore `touch`es the file before rebuilding, a **canary render** runs after every restore and stops the harness if it fails, and `scratch/coll/run.sh` refuses to measure anything against an artifact that fails the same canary. That is the "instrument reads the wrong input" class, and it cost a wrong measurement before it was caught.

## Tests

Full suite, all three roots (`pytest tests/ python/tests/ python/djust/tests/ -n auto`): **19839 passed, 438 skipped, 0 failed** on the merged tree. Rust: `make test-rust` exit 0, **66 suites** — including a new `an_iterable_encoded_without_items_is_empty`, which existed as a doc-comment citation before it existed as a test (#1197) and now asserts out loud that no live fixture has the wire-only shape, building one by hand to make the assertion reachable.

40 cases in `python/tests/test_opaque_collections_2477_2489.py`. Updated in place: the wire-position pins (9 -> 10 slots, with the legacy widths RECONSTRUCTED rather than truncated, because slot 4 widened from a boolean to a count and a truncation is not a legacy payload), the `#2212` bool-before-i64 converter set, the `#2343` panic-boundary set, the `#2448` `django_json_encoded` caller count, and the `#2478` `public_dict_attrs` pin — which now counts the RULE (`is_public_attr_name`) rather than the map builder, because a keys-only probe reads the same rule and building the map to answer "is it empty" converted every attribute value for the arm below to convert again.


johnrtipton:
Drive-by from the #2480 investigation (I was asked to look at #2480 and this PR
is the thing that changes its shape — no action needed on this PR unless you
agree it is in scope).

**This PR moves three cells that currently AGREE into the #2480 divergence, in
the same shape #2476 did.** Measured on `main` at `81a5cedd`, raw
`render_template`, `{% if p == q %}Y{% else %}N{% endif %}` over two distinct
instances:

| `p`, `q` | python | django | djust on `main` | after this PR |
|---|---|---|---|---|
| two `{'a'}` | True | `Y` | **`Y`** | `N` |
| two `frozenset({'a'})` | True | `Y` | **`Y`** | `N` |
| two `{'a': 1}.keys()` | True | `Y` | **`Y`** | `N` |

The mechanism is exactly #2480's: a truthy collection reaches
`Ok(Value::String(ob.str()?))` today, so two of them compare equal through
`values_equal`'s `(String, String)` arm — Django's answer, reached by accident.
Once `opaque_value` claims them they become `Encoded` with `cmp_key: None`,
`python_partial_cmp` short-circuits to `None`, and they stop being equal.

Not asserting this is a blocker — it is the same trade #2476 made deliberately
for `set()` (truthiness is worth more than an accidental equality), and the
after-column is inferred from the diff rather than measured on your branch, so
please check it. Two suggestions:

1. If it reproduces, worth a line in the CHANGELOG saying so, and a pin in the
   diverging direction, so it is a recorded consequence rather than a rediscovery.
2. `Encoded::items` is the thing that makes #2480 *closable* — two set-like
   `Encoded`s that both carry `items` are comparable from carried facts, which
   is precisely what #2480's body says nothing carried can decide. I have left
   the full design and the four measured protocol arms in
   https://github.com/djust-org/djust/issues/2480#issuecomment-5473757031, with
   the one trap spelled out: equality must **not** be routed through
   `python_partial_cmp` for this family, because `set() <= set()` is `Y` in
   Django while `complex(0) <= complex(0)` is `N` (Python raises), so a shared
   key gets one of them wrong.

I deliberately did not touch `Encoded` in `crates/djust_core/src/lib.rs` beyond
adding a const + a function + a five-line merge in `django_json_encoded`
(#2485, PR #2495) so this branch's slot-10/11 appends do not collide. That PR
grows the map inside existing slot 9; it adds no slot.

johnrtipton:
Correction to my comment above, having now read your branch rather than the
diff: you had already found and pinned this, in
`TestTheComparisonAxisThisWIDENS`, and your accounting is exact where mine was
an inference one shape short — I named three truthy shapes; the fourth is
`complex(1)`. Your `NEWLY_UNEQUAL` / `ALREADY_UNEQUAL` split, the disjointness
assertion, the truthiness cross-check and the `test_a_value_python_calls_
UNEQUAL_is_still_unequal` control are all better than what I suggested. So both
of my suggestions were already done; disregard them.

I have adopted your four-plus-four accounting as the record on #2480 and taken
ownership of it there, per the coordinator:
https://github.com/djust-org/djust/issues/2480#issuecomment-5473839862 — the
eight shapes measured on `main`, the two axes outside them (same-object, and the
cross-carrier `complex(0) == 0`), and the design that `items` makes available.

One thing from your branch that is worth knowing outside it: `into_pyobject`
having to keep returning the `display` is the same boundary that decides #2486
(`{{ dec.real }}`). I declined that issue partly because
`Value::Encoded(e) => Ok(e.display…)` means routing `Decimal` onto `Encoded`
would return a `str` where a handler expects a `decimal.Decimal`. Your 20
custom-filter cells are the measured version of the same constraint, so the two
decisions agree.

Nothing needed from you on #2495 (#2485) — it adds a const, a function and a
five-line merge in `django_json_encoded`, grows the map inside existing slot 9,
and adds no slot, specifically so your slot 10/11 appends do not collide.

PR #2498, opened 2026-08-31
feat(cli): `djust replay` — open, inspect, or diff a bug capture (#1561)

johnrtipton:
Part 2 of 3 for #1561 (B7 iter C). Independent of #2497 (the snapshot store) — either can merge first.

## What this adds

```bash
djust replay <blob>            # opens the replay URL via webbrowser.open()
djust replay --inspect <blob>  # the decoded capture as JSON, on stdout
djust replay --diff <blob>     # unified diff of state_before vs state_after
```

New subparser in `python/djust/cli.py`, following the existing style around the `clear` / `mcp` parsers.

### `--inspect`

Emits **one** JSON document — `event_name`, `scrubbed_fields`, `state_before`, `state_after`, `vdom_patches` — rather than three. That is what makes the issue's stated purpose work:

```bash
djust replay --inspect "$BLOB" | jq '.state_after.claimant_id'
djust replay --inspect "$BLOB" | jq '.vdom_patches | length'
```

### `--diff`

`difflib.unified_diff` over the pretty-printed, key-sorted states. Identical states write **nothing** to stdout and say so on stderr, so `djust replay --diff "$BLOB" > patch` produces a valid empty patch rather than a file containing the words "no changes".

## Two guards, because a blob arrives by paste

`djust replay <this thing I sent you>` is a real way to get a URL opened on someone's machine, so:

- **The argument must resolve to something starting with `djbug1.`** The command accepts a bare blob *or* a full replay URL (a teammate is as likely to paste one as the other), strips a query string / fragment / trailing slash, and then checks the prefix. `djust replay https://phishing.example/` exits 1 with a message and never reaches `webbrowser.open`. The URL handed to the browser is always one this command constructed.
- **`--base-url` is restricted to `http`/`https`**, since its scheme reaches `webbrowser.open` directly. `javascript:`, `data:`, `file:` and `ftp:` are refused by name.

Both are asserted not just to raise, but to leave the browser unopened.

## URL construction

Path from `reverse("djust:bug_capture_replay")` when a URLconf is available — so a project mounting `djust.urls` under a prefix gets `/tools/__djust__/replay/…` — falling back to the literal route `djust.urls` registers when the CLI runs outside a project. Host from `--base-url`, else `$DJUST_REPLAY_BASE_URL`, else `http://127.0.0.1:8000`.

Decoding an *inline* blob needs no Django at all. `django.setup()` is attempted quietly so a `djbug1.store.<id>` blob (once #2497 lands) and the URLconf lookup work inside a project, and the specific failure surfaces from whichever operation actually needed it rather than as a startup warning on every invocation.

## Tests

39 cases in `python/djust/tests/test_replay_cli.py` across seven classes. They drive the real `main()` argv path rather than calling `cmd_replay(Namespace(...))` — a direct call would pass even if the subparser were never registered, which mutation M9 below confirms.

### Dogfood (#1060)

Run against the demo project, not just under pytest:

```
=== --inspect | jq ===          inc / 1
=== --diff ===                  -  "count": 0 / -  "gone": 1 / +  "count": 1
=== identical states ===        stdout bytes: 0
                                stderr: state_before and state_after are identical.
=== error paths ===             3/3 exit 1 with a message, no traceback
=== --help ===                  renders
```

### Gate-off (#1468)

9 mutations, **all caught**. Harness asserts each mutation matched exactly once and changed the source, clears `__pycache__` per iteration, and counts `N error` separately from `N failed` (#2129/#2135):

```
BASELINE                                                       -> 0 failed, 39 passed
M1 non-blob argument accepted (the phishing guard removed)     -> 8 failed
M2 any --base-url scheme allowed                               -> 6 failed
M3 a pasted replay URL is not unwrapped                        -> 4 failed
M4 --inspect does nothing                                      -> 3 failed
M5 --diff does nothing                                         -> 3 failed
M6 the URLconf branch is never used (always the literal route) -> 1 failed
M7 --base-url no longer beats $DJUST_REPLAY_BASE_URL           -> 1 failed
M8 the empty-diff note goes to stdout instead of stderr        -> 1 failed
M9 replay is not in the dispatch table                         -> 14 failed
RESTORED                                                       -> 0 failed, 39 passed
```

### Full suite

`pytest tests/ python/tests/ python/djust/tests/ -n auto` — **19906 passed, 438 skipped, 0 failed**.

No Rust source changed.

Refs #1561


johnrtipton:
Pushed a fix — and the diagnosis is that **my test was wrong, not the code**.

`test_uses_the_urlconf_when_the_route_is_mounted` mounted `djust.urls`, whose `urlpatterns` is computed **at import time** from `settings.DEBUG`. Django's test runner forces `DEBUG = False`, so under pytest that module is normally **empty** — the `include` mounted nothing, `reverse` failed, and the test silently measured the *fallback* branch while claiming to measure the urlconf one.

It passed locally only because `djust.urls` happened to have been imported before the runner set the flag. Reproduced deterministically:

```
pytest python/djust/tests/test_bug_capture_urls.py python/djust/tests/test_replay_cli.py
-> 1 failed
```

`test_bug_capture_urls.py` reloads that module under `override_settings`; whichever ordering CI picked ran it first.

The production code is right — falling back to the literal route when no route is registered is exactly what should happen outside a project. The test now registers the named `djust:bug_capture_replay` route directly (all `_replay_url` actually needs), and a companion case pins the fallback branch explicitly so it can't rot behind the mounted-route one. Green in **both** orderings, and gate-off M6 (URLconf branch bypassed) still reddens exactly one test, so the assertion is still load-bearing.

Also merged `origin/main` in (never rebased) now that #2497 has landed — one `CHANGELOG.md` union and one docs conflict in `bug-capture.md` (both branches appended a section), resolved by keeping both sections and collapsing the two "what's coming" lists into one. Verified no already-shipped CHANGELOG content was dropped by the merge.

Full suite post-merge: **19975 passed, 438 skipped, 0 failed** across all three roots.

PR #2499, opened 2026-08-31
feat(bug-capture): declarative PII scrub + the V014 check (#1561)

### LINKED ISSUE #1561: feat(bug-capture): iter C — Redis store, djust replay CLI, framework-level PII scrub

johnrtipton:
Part 3 of 3 for #1561 (B7 iter C). `origin/main` is merged in (never rebased), so this is reviewed against the same program the merge will apply.

## `LiveView.time_travel_excluded_fields`

Names the top-level public-state keys that must never leave a view inside a shared bug capture. `encode_view_state()` applies them **before** any caller-supplied `scrub`, so the redaction stops depending on every call site remembering to pass one — the `scrub` argument stays available for arbitrary policies on top.

**`scrub_fields()` already existed, so it is called, not reimplemented** (#1646). That is not just tidiness: it already handles absent keys, and it already **carries forward** `scrubbed_fields` — which is exactly what makes names from both sources land on the wire together, the issue's acceptance criterion. The whole composition is three lines.

**One sink, no twin.** The debug panel's WS Share button already routes through `encode_view_state`, so it inherits this with no second implementation. Pinned by a structural test asserting `handle_bug_capture_share` calls `encode_view_state` and that `websocket.py` constructs no `BugCapture` of its own, so the next person to add a capture surface trips it rather than quietly building a parallel path.

Tolerant of the shapes a class attribute takes in the wild — list, tuple, set, frozenset (#1108). A bare string is the interesting one: `time_travel_excluded_fields = "password"` iterates as eight characters and silently scrubs nothing, so it is treated as a single name. A non-iterable warns and is ignored rather than raising inside an encode.

## The check is `V014`, not `V012`

The issue names V012. **That ID has been taken since #1803** (sticky child declares its own `dj-view`), and V001–V013 are all in use — so the check registers as `djust.V014`. Named in the CHANGELOG and `docs/system-checks.md` so nobody grep-searches for a V012 that means something else.

## False positives — the judgment you asked for

You were right that this is where the check lives or dies, so the design is three gates rather than a pattern list:

1. **`time_travel_enabled = True`.** A deliberate dev-only opt-in almost no view sets. This does the heavy lifting: a project not using time travel never sees V014 at all.
2. **Token matching, not substring matching.** `telephone_pole` tokenizes to `("telephone", "pole")` and does not contain the *token* `phone`. A substring scan would have flagged it. Trailing digits and a trailing plural `s` are normalized off, so `password1` and `passwords` still match; camelCase is split, so `creditCard` does.
3. **Field TYPE.** This is what makes `email` survivable. `email` is on almost every user model — and so is `email_verified`. A `BooleanField` named `email_verified` cannot hold an email address, so Boolean / date / time / relation / auto-pk fields are skipped whatever they are called. That is what keeps `email_notifications`, `phone_confirmed_at` and `password_changed_at` quiet. Matched by class name up the MRO, so third-party field subclasses are covered without importing anything.

### Dogfood (#1060)

Against the demo project and `examples/`:

```
LiveView subclasses considered: 57
  of which time_travel_enabled = True: 0

--- AS-IS (the shipped gate) ---
V014 messages: 0

--- FORCED: time_travel_enabled = True on every view ---
V014 messages: 10   (14 fields)

fields scanned: 71
name-matched but skipped by TYPE: 0
```

**0 hits as shipped.** Forcing the flag on every demo view produces 10 messages naming `email`, `phone`, `password`, `password_confirm` on the contact/registration forms — every one a true positive, real text fields really carrying that data. So the matching is accurate as well as quiet, and I did not need to drop `email`: gate 1 keeps it silent on projects that aren't capturing state, and gate 3 keeps it from firing on the `email_*` booleans that would otherwise make it noise.

Worth flagging: **`name-matched but skipped by TYPE: 0`** — the type gate never fires on demo data, so the dogfood pass does not exercise it. It gets its own cases in the suite for that reason, and its own gate-off mutation below.

## Field discovery greps for the sink

`_iter_declared_fields` scans the view's class attributes for **values** that are Django models, forms or querysets, rather than enumerating the attribute names a view is expected to use (`model`, `form_class`, `queryset`, …). djust has no framework-level `model` attribute — it is a project convention and the conventions differ — so an enumerate-the-names scan would be reliably one short (v1.1.1-2 retro).

## Not `DEBUG`-gated

The issue's acceptance list says "all checks DEBUG-gated". I read that as the **runtime** surfaces — `BugCapture.encode`, the replay route — which are, because they *do* something. A system check only tells you something, and `manage.py check --deploy` on the way to production is exactly when you want to hear that a shipped view records a password field. Stated here rather than silently diverging.

## Tests

55 cases in `python/djust/tests/test_bug_capture_pii_scrub.py`.

The ordering assertion is made from **inside the caller's scrub callable**: asserting on the final output cannot distinguish "excluded ran first" from "excluded ran second" — both remove the key. Observing the state the callable is *handed* is what pins the order.

The demo-clean pin imports the demo views and asserts a precondition first. Without that, the filtered list is empty because the classes were never loaded and the assertion would pass whatever V014 did (#1200) — measured: gating `time_travel_enabled` off reddens 1 test before the fix and 2 after.

### Gate-off (#1468)

11 mutations, **all caught**. Harness asserts each mutation matched exactly once and changed the source, clears `__pycache__` per iteration, and counts `N error` separately from `N failed` (#2129/#2135):

```
BASELINE                                                -> 0 failed, 55 passed
M1  exclusions never applied                            -> 9 failed
M2  the caller scrub runs FIRST instead of second       -> 1 failed
M3  a bare-string declaration iterates as characters    -> 1 failed
M4  V014 ignores time_travel_enabled (gate 1)           -> 2 failed
M5  V014 matches substrings instead of tokens (gate 2)  -> 5 failed
M6  V014 ignores the field TYPE (gate 3)                -> 1 failed
M7  V014 ignores the declared exclusions                -> 2 failed
M8  V014 cannot be suppressed                           -> 1 failed
M9  V014 never inspects model fields                    -> 4 failed
M10 V014 never inspects form fields                     -> 5 failed
M11 V014 never follows a queryset to its model          -> 1 failed
RESTORED                                                -> 0 failed, 55 passed
```

Each of the three false-positive gates has a test that goes red when only that gate is removed.

### Full suite

`pytest tests/ python/tests/ python/djust/tests/ -n auto` post-merge — **19990 passed, 438 skipped, 0 failed**.

Also re-pins the two whitelisted `DynamicLiveView` `setattr` line numbers in `test_mount_chokepoint_structural.py`, which shifted +17 (1326/1328 → 1343/1345) when the class attribute and its comment block landed next to `time_travel_enabled`. Re-verified sanctioned: still the same two developer-dict `setattr` lines, not a new client-controlled one.

No Rust source changed.

Closes #1561 (with #2497 and #2498).


johnrtipton:
Extra dogfood evidence, run through the **real** `manage.py djust_check` command rather than only the check function:

```
djust OK: .../worktrees/agent-ad0614e5b9def3c61/python/djust/__init__.py
=== V014 lines in the real manage.py djust_check output ===
0
=== summary ===
  Summary: 212 warning(s), 326 info
```

**0 V014 lines** on the demo project; the pre-existing 212/326 are unchanged.

And an empirical canary (#1459), because "0 findings" and "the check isn't wired in" look identical from that output:

```
registered with Django's check registry: True
V014 findings with a synthetic offender present: 1
    __main__._CanaryView: time_travel_enabled = True, and its model/form declares
    field(s) whose names look like PII and are not in time_travel_excluded_fields: password.
```

So the zero is "nothing to report", not "nothing is looking".

One process note worth recording, since it nearly produced a false measurement: my first run of this used `PYTHONPATH=<repo-root>` (copying the CI `demo-checks` invocation), and `import djust` resolved through the shared venv's editable install to **a different agent's worktree** — so it was measuring somebody else's code, and would have reported a plausible-looking 0 for entirely the wrong reason. Both scripts now assert `"agent-ad0614e5b9def3c61" in djust.__file__` before measuring anything. The pytest-based dogfood in the PR body was always correct (it puts `$WT/python` first); this is the shell-invocation variant that wasn't.

Issue #2505, opened 2026-09-01
Sidecar: a loop/with variable that SHADOWS a top-level context name resolves against the outer object


johnrtipton:
Follow-up from #2501, filed with the #2504 sibling. Pre-existing — #2501 deliberately does not move it.

## The defect

`Context::resolve`'s raw-Python sidecar is keyed by top-level context name. When a binding construct binds a name that ALSO exists at top level, and the alias mechanism registers nothing for it (a filtered or normalised operand — see #2504), a dotted lookup under that name falls through to the sidecar and resolves against the **outer** object:

```django
{% for x in p|slice:":2" %}{{ x.cls_attr }}{% endfor %}
```

with a context carrying both `p` and an unrelated top-level `x`. `Context::get("x.cls_attr")` misses (the loop bound `x` to a `Value` that has no `cls_attr`), and the sidecar answers with `raw["x"]` — a different object than the one bound.

Django has no such channel: `Variable._resolve_lookup` walks the context stack, so the innermost binding wins outright.

## Why #2501 does not fix it

#2501 added an alias-expansion fallback to the sidecar walk, consulted only when `key`'s own head names no sidecar entry. That ordering is what keeps the whole change bounded by its miss-only property — the sidecar may ADD a resolution and never CHANGE one — so this pre-existing wrong resolution is left exactly as it is today rather than being silently altered inside a bugfix PR.

(Where an alias IS registered, `{% for x in rows %}` with a top-level `x` present now resolves correctly, because the alias-expanded `rows.<i>` is tried first for a head the sidecar does not name. The shadowing hazard is confined to the alias-less shapes.)

## Fix sketch

The sidecar should be consulted for `first` only when `first` is not bound in a frame ABOVE the root — i.e. the binding that shadows it should suppress the fallback. `Context` would need to distinguish "bound by the root context dict" from "bound by a pushed frame / a `set` during render", which it does not today (`{% for %}` clones the context and `set`s into it rather than pushing a frame).

## Acceptance

A test in `python/tests/` that binds a loop/with variable shadowing a top-level sidecar name over an alias-less operand and asserts the rendered value is the BOUND object's, matching live Django.

johnrtipton:
**Premise correction from the ADR-027 characterization planning (2026-09-02, #2539 plan §1).** The issue says the shadowing hazard is confined to the *filtered* `{% for %}` operand (the alias mechanism). Measured at HEAD, it is not:

[...]

Issue #2504, opened 2026-09-01
Sidecar: a filtered or dict-view `{% for %}` operand cannot reach object attributes (the alias mechanism refuses it)


johnrtipton:
Follow-up from #2501 (PR: `fix-2501-object-attribute-resolution`). Chain-suspected per the #2142 convention.

## What #2501 closed

The raw-Python sidecar is keyed by TOP-LEVEL context name, so until #2501 every construct that BINDS a value to a NEW name left the raw object unreachable under that name:

```
{{ rows.0.cls_attr }}                              django 'class-level'              djust 'class-level'   OK
{% for r in rows %}{{ r.cls_attr }},{% endfor %}   django 'class-level,class-level,' djust ',,'            MISMATCH
{% with q=o %}{{ q.cls_attr }}{% endwith %}        django 'class-level'              djust ''              MISMATCH
```

#2501 fixed this by having the sidecar walk consult `Context::aliases` (#2375) when `key`'s own head names no sidecar entry — `r` IS `rows.<i>`, `q` IS `p`. That is the #1646 cure: one statement of "which context path this name IS", reused rather than re-stated per construct.

## What it does NOT close, and why

An alias is registered only where the correspondence is REAL, and those guards are XSS boundaries (`Context::is_safe` resolves a safety decision through the same alias). Two operand shapes therefore register nothing, and so the sidecar cannot reach them either:

1. **A filtered operand** — `{% for r in rows|slice:':1' %}{{ r.cls_attr }}{% endfor %}`. `slice` shifts indices, `dictsort` reorders, so `r` is not `rows.<i>`.
2. **A normalised operand** — a dict or a dict view, `{% for r in dd.values %}{{ r.cls_attr }}{% endfor %}`. `_collect_safe_keys` spells a dict's paths BY KEY NAME while the loop asserts an INDEX; for attacker-controlled keys that collision is a live XSS (#2334), which is why the loop mapping refuses it.

Measured on the #2501 branch (release build, live-Django differential):

```
{% for r in rows|slice:':1' %}{{ r.cls_attr }},{% endfor %}   django 'class-level,'  djust ','
{% for r in dd.values %}{{ r.cls_attr }},{% endfor %}         django 'class-level,'  djust ','
```

Everything else in the matrix agrees — plain `{% for %}`, `reversed`, nested loops, `{% for k, v in pairs %}` unpacking, all three `{% with %}` spellings.

## Why it is filed rather than fixed

[...]

PR #2508, opened 2026-09-01
fix(templates): resolve object attributes on the non-LiveView render paths (#2501)

### LINKED ISSUE #2501: Built-in components: {{ component }} renders ESCAPED and {{ component.render }} renders EMPTY — three of four documented spellings diverge
Reported by @traormoh in [discussion #2437](https://github.com/djust-org/djust/discussions/2437). Reproduced against a **real** `djust.components.base.Component` subclass, not a stand-in, and measured against live Django.

## The matrix

```
  spelling                 django                            djust
  {{ c }}                  '<div class="alert">real</div>'   '&lt;div class=&quot;alert...'   ESCAPED
  {{ c|safe }}             '<div class="alert">real</div>'   '<div class="alert">real</div>'   ok
  {{ c.render }}           '<div class="alert">real</div>'   ''                                EMPTY
  {{ c.render|safe }}      '<div class="alert">real</div>'   ''                                EMPTY
```

Three of four diverge, and **all four are documented spellings** — `{{ counter.render }}` at `docs/website/guides/components.md:227`, `{{ navbar.render }}` in `docs/example-site/TEMPLATE_CONVERSION_GUIDE.md:54`, and `components.md:637` states *"Components in templates render via `{{ component }}` — the `__str__` method"*.

The reporter's workaround (`|safe`) is the `{{ c }}` row. It does **not** rescue `{{ c.render }}`, which is empty with or without it — worth saying, since the discussion implies otherwise.

## Two distinct root causes

**1. `{{ c }}` escaped — the `__str__` SafeString marker is lost at the conversion.**

`Component.__str__` returns a genuinely marked value: `hasattr(str(a), "__html__")` is `True`. The marker does not survive `impl FromPyObject for Value`. Same class as #2290 (custom filter loses SafeData) and #2416 (the argument channel), one channel over.

**2. `{{ c.render }}` empty — object conversion carries only the instance `__dict__`.**

```
                      django          djust
{{ o.inst_attr }}     'in-dict'       'in-dict'      instance __dict__
{{ o.cls_attr }}      'class-level'   ''             class attribute
{{ o.meth }}          'called'        ''             method + auto-call
{{ o.prop }}          'prop'          ''             property
```

Django's `Variable._resolve_lookup` does dictionary lookup → **attribute** lookup (reaching class attrs, descriptors, properties) → **auto-call if callable** → list-index. djust's `__dict__` bulk-dump arm reaches only the first. A real `Component` instance's `__dict__` is `['_component_key', '_explicit_id', '_rust_instance']` — `render` is a class method and is simply not there.

## Not a regression, and already half-known

This is **not** fallout from #2379/#2415 (the custom-tag-return escaping). It is the `__dict__` bulk-dump arm that **#2478 explicitly declined to retire** — *"retiring that arm is a much larger change"* — and that decline was correct in scope. This issue is the user-visible cost of it.

Note the asymmetry it leaves: **#2481** gave `Value::Encoded` an attribute map and **#2495** gave it nullary auto-call, so `{{ dt.year }}` and `{{ dt.isoformat }}` both work. `Value::Object` got neither. The carrier that models *arbitrary Python objects* is the one that cannot resolve their attributes.

## Reproduction

```python
from djust.components.base import Component
from djust import _rust

class Alert(Component):
    template = None
    def _render_custom(self): return '<div class="alert">real</div>'

a = Alert()
assert hasattr(str(a), "__html__")           # marked safe in Python
_rust.render_template('{{ c }}',        {"c": a})   # -> escaped
_rust.render_template('{{ c.render }}', {"c": a})   # -> ''
```

## Triage notes

- **Trust the symptom, not this issue's cited paths.** They are hypotheses; verify symptom-up. This repo has repeatedly found an issue's cited location already correct (#2435) or non-existent (#2463).
- The two causes are independent and may want separate PRs; #1 may be much smaller than #2.
- For #2, the prior art to read first is #2481's `attrs` map and #2495's `ENCODED_CALL_NAMES` / `collect_called_attrs` — the question is whether `Value::Object` should route through the same carrier rather than gaining a parallel implementation (#1646).
- **Auto-calling is a behaviour change with a security edge**: Django auto-calls attributes, honouring `do_not_call_in_templates` and `alters_data`. Any implementation must respect both, or a template lookup can invoke a mutating method.
- `Value::Encoded`'s msgpack tuple is currently 11 positional slots with a strict pin (`crates/djust_core/tests/test_encoded_wire_positions_2471_2472.rs`); if the fix routes objects onto it, that pin and its interior-insert canary apply.

## Separately, from the same discussion

The reporter's second problem (stateful alert/modal erroring) is a different failure: the console shows `Container found but no view path specified` → mount refused → HTTP fallback → `POST /counter/ 500`. djust's own `alert.html` / `modal.html` do **not** emit `dj-view`, so the empty attribute is in the reporter's page template — but the 500 is a real server error and needs their traceback. Not tracked here.

### LINKED ISSUE #2506: security: the sidecar walk swallows ANY exception from a template attribute lookup — an authorization check that raises fails silently and OPEN
Surfaced during the #2501 self-review. Pre-existing (the walk has had this shape since #1997), but #2501 makes it reachable from three more entry points — `_rust.render_template`, `_rust.render_template_with_dirs` and `DjustTemplateBackend` — so it is fixed in that PR rather than left.

## The defect

`Context::resolve_without_builtins`'s per-segment walk (`crates/djust_core/src/context.rs`) was:

```rust
let next = current
    .get_item(*part)
    .or_else(|_| current.getattr(*part))
    .or_else(|e| match part.parse::<usize>() { … });
match next {
    Ok(n) => current = n,
    Err(_) => return Ok(None),   // renders ""
}
```

Every `_` discards the exception, so ANY error from a `__getitem__` or a property descriptor becomes "this lookup resolved to nothing" and renders the empty string.

Django's `Variable._resolve_lookup` catches an explicit, narrow set at each step:

| step | Django catches | anything else |
|---|---|---|
| 1 — item access | `TypeError, AttributeError, KeyError, ValueError, IndexError` | propagates |
| 2 — `getattr` | `TypeError, AttributeError` — and even then re-raises if `bit in dir(current)` (the "raised by a @property" branch) | propagates |
| 3 — integer index | `IndexError, ValueError, KeyError, TypeError` → `VariableDoesNotExist` | propagates |

Measured against live Django before the fix:

```
{{ r.runtime }}   # @property raising RuntimeError("authz check failed")
  django  -> raises RuntimeError
  djust   -> ''
```

## The security reading

An attribute or `__getitem__` that implements an access check and signals refusal by raising is a normal shape. Under this walk that refusal is swallowed and the template renders blank — indistinguishable from "no such attribute". The failure is silent, and for a check whose *raise* is the deny it is silent and OPEN: nothing surfaces to logs, to the error page, or to a monitor. `maybe_call` already propagates a real exception raised INSIDE a nullary method (ADR-024), so the getattr half was the odd one out.

## Fix

Transcribe Django's three catch sets, including the `bit in dir(current)` re-raise branch, so an exception outside them propagates as a render error. Shipped in #2501 as `is_django_item_lookup_error` / `is_django_index_lookup_error` / `name_exists_on`.

Behaviour change to declare: a template whose attribute access raises now surfaces the error instead of rendering empty. That is Django parity and the direction of convergence, but it is a real change for any project relying on the swallow.

### LINKED ISSUE #2507: security: component mutators are auto-callable from a template — `{{ c.unmount }}` runs the component's cleanup during render
Found by the Stage 8 security check on the #2501 branch. Fixed in that PR.

## The defect

#2501 un-excludes `Component` / `LiveComponent` from the LiveView raw-Python sidecar so a component's class attributes, properties and nullary methods resolve in templates. None of the component MUTATORS carried Django's `alters_data` marker, so the auto-call invoked them.

Measured through a real LiveView render:

```
{{ c.unmount }}          -> 'UNMOUNTED'   the component's user-authored cleanup RAN
{{ c.trigger_update }}   -> re-enters the parent view's _trigger_update MID-RENDER
{{ c.mount }} / {{ c.update }} / {{ c.clear_context_providers }}  -> also invoked
```

This is exactly what `alters_data` exists for, and why Django stamps it on `Model.save` / `Model.delete` / `QuerySet.delete` / `QuerySet.update`.

## Why stamping the base methods was not enough

`alters_data` is an attribute of the FUNCTION, so a subclass override gets a fresh function object without it — and overriding is not an edge case here, it is the documented way to use a component (`unmount`'s own docstring says *"Override this method to perform cleanup actions"*). With only the base methods stamped, a subclass's overridden `unmount` still ran. Stamping the five base methods would have guarded almost nobody.

## Fix

`TemplateMutatorGuard` (`python/djust/_template_guards.py`), mixed into both `Component` and `LiveComponent`. Its `__init_subclass__` re-stamps `ALTERS_DATA_COMPONENT_METHODS` = `{mount, unmount, update, trigger_update, clear_context_providers}` on every subclass override at every depth, so the marker is a property of the NAME for the whole hierarchy rather than of one function object (#1646).

`render` and `get_context_data` are deliberately NOT in the set — they are read-only and `{{ c.render }}` is the documented spelling #2501 exists to fix. Their absence is asserted rather than left to inspection.

`alters_data` gates TEMPLATE auto-call only; the framework's own lifecycle dispatch and `ComponentMixin.update_component` (#1947) call these methods directly and are unaffected.

## Coverage

`TestComponentMutatorsAreNeverAutoCalled` in `python/tests/test_sidecar_on_all_render_paths_2501.py` — every mutator x all four render paths, on a subclass that OVERRIDES them; a two-generation subclass check; a count-canary over the whole nullary surface of both classes so a mutator added later fails there rather than shipping template-callable; and the direct-call half.

Gate-off: removing the `__init_subclass__` re-stamp (leaving the base decorators) reddens 25 cases; emptying the guard set reddens 26.

johnrtipton:
## What

Reported by an external user in discussion #2437: `{{ component.render }}` rendered nothing.

The mechanism turned out not to be components at all. djust's object conversion carried only the instance `__dict__`, so **class attributes, properties and Django's callable auto-call were all unreachable** on plain context objects. Django's `Variable._resolve_lookup` does dictionary lookup → attribute lookup → auto-call → list-index; djust reached only the first. A real `Component`'s `__dict__` holds nothing but `_`-prefixed internals, which is why its `render` was simply absent — and why my first analysis on the issue, which blamed the `__dict__` bulk-dump arm, was wrong. I posted a correction there.

## How

The fix walks the **existing** raw-Python sidecar lazily, consulted **only when `Context::get` misses**. That ordering is what bounds the change: it can add a resolution and can never change one, so only cells that previously rendered empty are affected. `{% for %}` and `{% with %}` bound names reach it through alias expansion.

Attributes are deliberately **not** measured at conversion time. An eager sweep would evaluate every property getter and call every nullary method on every render, and an earlier eager `__dict__` walk is on record as having segfaulted on an ordinary presenter object.

### This is a widening, not only a fix

Django parity, and the developer did put the object in the context — but worth knowing before you upgrade:

- `{{ settings_obj.SECRET_KEY }}` renders where it rendered empty before.
- A nullary method that raises is now a **500** where the cell used to render `""`.
- A `{% tag %}` handler that received `ctx["user"]` as a dict now receives a `_SidecarModelProxy` on these three paths, so `ctx["user"]["username"]` breaks.

The `_`-prefix refusal (#2436) applies at every segment, so `{{ o.__class__ }}` and `{{ o.__init__.__globals__ }}` stay refused.

### Two security findings, both from the Stage 8 check

| | |
|---|---|
| **The floor was top-level only** | Applied only at the top of the sidecar. Sufficient for the walk sink, which re-protects at every step — not for the custom-tag bridge, which injects the sidecar wholesale. `{"user": u}` was filtered; `{"users": [u]}` and `{"d": {"u": u}}` were not, so `{{ users.0.password }}` exposed the hash. Reproduced against a real `auth.User`, re-verified filtered on all three shapes. |
| **Component mutators became auto-callable** | Making attributes resolvable meant `{{ c.unmount }}` ran the user's cleanup *during render* and `{{ c.trigger_update }}` re-entered the parent view mid-render. Now marked `alters_data` — Django's own marker on `Model.save` — on both `Component` and `LiveComponent`. `LiveView`'s mutators were never reachable; the view is not in the sidecar. |

A third finding: the attribute walk caught **every** exception where Django catches a specific tuple, so a property raising `PermissionDenied` rendered blank and `{% if not doc.is_restricted %}` failed **open**. Harmless before this PR because the guarded content wasn't resolvable; consequential after it. Narrowed to Django's sets.

## Mapping

| Issue | Files | Covering tests |
|---|---|---|
| #2501 attribute resolution | `crates/djust_core/src/context.rs`, `crates/djust_live/src/lib.rs`, `python/djust/template/rendering.py`, `python/djust/mixins/rust_bridge.py` | `test_object_attribute_resolution_2501.py`, `test_sidecar_on_all_render_paths_2501.py`, `test_encoded_attributes_2481.py`, `test_nullary_autocall_2485.py` |
| #2506 exception swallowing | `python/djust/_template_guards.py` | `test_object_attribute_resolution_2501.py` (fail-closed cases) |
| #2507 component mutators | `python/djust/components/base.py`, `python/djust/decorators.py` | `test_object_attribute_resolution_2501.py::alters_data` cases |
| container floor | `python/djust/serialization.py` | `test_sidecar_on_all_render_paths_2501.py` (list / dict / nested shapes) |

## Not in this PR

`{{ component }}` and `{{ component.render }}` now **resolve** but are still escaped — `{{ component.render|safe }}` works and is what the reporter needed. The escaping half is `serialize_value` / `normalize_django_value` drift and is PR 2 of #2501; the components guide states the `|safe` workaround in the meantime.

Follow-ups filed rather than scope-crept (#1079): #2504 (filtered/dict-view `{% for %}` operand), #2505 (loop var shadowing a top-level name).

## Verification

- 20,309 tests passing.
- Gate-off run on each of the three 🔴s; each fails with the fix reverted.
- All three re-verified by hand after the fix pass:

```
{'user': u}      user.password='<none>'        filtered
{'users': [u]}   users[0].password='<none>'    filtered
{'d': {'u': u}}  d.u.password='<none>'         filtered
{{ c.unmount }} -> ''   calls=[]               refused
{% for r in rows %}{{ r.cls_attr }},{% endfor %}
   django='class-level,class-level,'  djust=same   OK
```

Closes #2501
Closes #2506
Closes #2507


johnrtipton:
## Stage 11 — adversarial review (worktree-isolated) → REQUEST_CHANGES → fixed

Three 🔴 and six 🟡. All three 🔴 reproduced independently before I touched anything, and the first was proven a **regression** by rebuilding against `main`'s `context.rs`. The 20,308-test suite was green through every one of them: each was a *shape* the tests never constructed, not a mechanism without a test — which is also why the gate-off in the original PR was 8/8 and told me nothing about these.

### 🔴 1 — `Model.DoesNotExist` in a property was a 500 on every path, LiveView included

`_resolve_lookup` wraps the whole chain in an outer arm the #2506 narrowing didn't transcribe:

```python
except Exception as e:
    if getattr(e, "silent_variable_failure", False):
        current = context.template.engine.string_if_invalid
    else:
        raise
```

`ObjectDoesNotExist` sets that attribute; every `Model.DoesNotExist` inherits it. So the commonest ORM-miss idiom in Django renders an empty cell — and we 500'd it:

```
django : 'Hello alice - last: []'
djust  : RuntimeError: User.DoesNotExist: no match
```

Because `Context::resolve` is shared, this hit the LiveView path too — wider than anything this PR claimed to touch. The fix mirrors Django's arm, and covers the auto-call as well, since Django's handler wraps that too.

### 🔴 2 — the exception type was destroyed, so 403 and 404 became 500

`From<PyErr> for DjangoRustError` stringified. Django's handler chain dispatches on **type**, so `PermissionDenied` → 403 and `Http404` → 404 both arrived as a generic `RuntimeError` and rendered 500. Fixed with an error variant that carries the `PyErr` whole; `DjustTemplateBackend` also stops re-wrapping those types, because a template-location hint is worth nothing next to losing the status code.

This one is a fair hit on my own CHANGELOG: I wrote that a property raising `PermissionDenied` "surfaces instead of silently passing the gate". It surfaced — as the wrong thing.

### 🔴 3 — the floor walk was exponential in width, and the guard that named the DoS didn't bound it

The identity table was a path-scoped visited set: right for a true cycle, useless for a DAG. A container referenced from N places was re-walked N times.

| width × depth | objects | before | after |
|---|---|---|---|
| 4 × 8 | 32 | 64 ms | 0.018 ms |
| 6 × 8 | 48 | **1.8 s** | 0.020 ms |
| 10 × 8 | 80 | ~104 s | 0.032 ms |

Per render, on a structure containing no cycle at all. `_SIDECAR_MAX_DEPTH`'s own comment named this exact DoS as the thing it prevented — it bounded depth, not fan-out (#1859). Now a memo keyed on `(id, depth)`; the depth is in the key because a subtree first reached at depth 10 is cap-truncated, and reusing that for the same object at depth 2 would under-protect it.

### 🟡 also fixed

- A model used as a **dict key** reached a tag handler with its real hash. Leaving keys alone was reasoned as "not a shape any template path reads" — true of the walk sink, false of the tag sink the descent exists to serve.
- A **namedtuple** silently degraded to a plain `tuple`, so `ctx["x"].field` became `AttributeError`.
- `template_auto_call` reached **one of three** framework render paths. One shared reader now (#1646).
- The `_SIDECAR_SKIP_TYPES` comment claimed admitting a container "costs one reference, not a traversal". The *walk* is lazy; *admission* is not (~1 ms per render for 1000 dicts). Corrected to describe what the code does.

### Deliberately not fixed here → #2509

Generators, custom `__iter__`, dataclasses and past-depth values still reach a tag handler raw. Enumerating container types is the wrong shape of fix — #2501 extended the descent once, this PR extended it again, and each round left a neighbour. The invariant belongs at the sink. The CHANGELOG now says so explicitly, so the container fix isn't read as closing the class.

Also confirmed pre-existing and left alone: `{% for k,v in d.items %}` (#2504), and shadowing between a bound name and a top-level key (#2505) — reproduces on `main`'s `context.rs` too, but is newly reachable on the three standalone paths, so it's now named in the CHANGELOG rather than only in the issue.

### What the reviewer probed and cleared

Miss-only ordering holds against falsiness (13/13 — a present-but-`None` key does *not* fall through), `dict_view`/`string_index` precedence matches Django, dunder refusal is airtight at every segment (9/9 refused), **no mutator was missed** (every public callable on `Component`/`LiveComponent` enumerated; `render` and `get_context_data` verified read-only), laziness is real (no getter or nullary method runs at conversion time), and parity is 33/36 with all three diffs accounted for.

### Verification

- 20,326 Python passed / 437 skipped / 15 xfailed; Rust green via `make test-rust` (the two-phase target — bare `cargo test` can't link the extension module here).
- Gate-off **6/6 load-bearing**, but only after a correction worth recording: `backend unwrap` **survived** at first. Not an equivalent mutant — my own `raised_type()` test helper unwraps `__cause__` and was therefore blind to whether the backend wraps, and `TestExceptionTypePreserved` only drove the direct path. Two mechanisms shadowing each other, so the fix had no test. Added one asserting the exact type with no `__cause__` unwrapping; gating the unwrap off now fails 3.
- The harness itself had to be fixed first. My initial run was killed by a command timeout **mid-mutation** and left a mutant in the tree; the next run read the mutant as its baseline, reported its anchor unmatched, and skipped — polluting every subsequent number. Only the `restored baseline` assertion caught it. It now snapshots every target up front, restores via `atexit` on any exit, and asserts a clean baseline before *and* after.

johnrtipton:
## Re-review of the fix pass → three more 🟡, all fixed

The re-review (worktree-isolated, worktree-built binary, plus a **rebuild of the pre-fix Rust** for a real before/after) confirms the three 🔴 are genuinely closed, and both of the things I most wanted checked came back **negative**:

- **#2506 is not re-opened.** Every non-silent exception still propagates with its type — `RuntimeError` from a property acting as an authz gate, from `__getitem__`, from a nullary method, and under `{% if %}` / `{% for %}` / filter-input / filter-argument. `silent_variable_failure = False` propagates. The fail-open gate the narrowing exists to close stays closed.
- **The memo never returns an unprotected object.** 4001 memoized entries, **0** ids mapping to two originals; the floor holds at the real tag sink in ten container shapes; no gc-object delta over 30 renders and the walked container is freed after the call.

The 1.8 s → 0.02 ms number reproduces bit-exactly (1771.9 ms → 0.0239 ms). The differential went from **9 divergences pre-fix to 5 post-fix, with zero new ones** — the 5 remaining all pre-date this PR.

### Three defects remained, two of them the same class as the bug they sit on

**`_is_user_raised` had three of Django's five dispatchable types.** `BadRequest` and `MultiPartParserError` are *siblings* of `SuspiciousOperation`, not subclasses, so the `isinstance` never reached them and both still rendered **500 where Django sends 400** — the exact defect the 🔴 closes, left half-closed. I had written the list from memory. It is now read from `response_for_exception`, and pinned by a test that diffs against Django's own source so a Django version that changes its dispatch set fails here rather than quietly costing a status code.

**An arity-1 namedtuple was corrupted.** `type(value)(items)` *raises* for 2+ fields, so the `*items` fallback ran and was correct — but at arity 1 it *succeeds* and wraps the field in the items list. `P1(user)` came back as `P1(a=[proxy])`, so `x.a.username` raised `AttributeError: 'list' object has no attribute 'username'` — precisely the failure that branch exists to prevent. It survived because my test built a 2-field tuple. Detection is now `_fields`, and the test is parametrized over arity 1/2/3.

**The cycle guard was dead code, and its comment said otherwise.** A back-edge returns to the same object at a *deeper* depth, so a memo keyed on `(id, depth)` can never see it. Zero seed reads under instrumentation; what actually stopped cycles was the depth cap, twelve rebuilt levels later — and deleting the line left `test_a_true_cycle_still_terminates` green, i.e. tautological, in a PR whose own thesis is #1859. Now a path-scoped `_active` set keyed on id alone, which is reachable, and the test asserts *where* it stops (one rebuilt level whose element is the original object) rather than merely that it returned.

**Plus: the "one shared reader" claim was false.** `rust_bridge.py` still read the config inline — a **fourth** reader, left standing by the change that cited #1646 to retire it, while the new helper's docstring called itself "the single reader" and listed three paths. Converged; the reader set is now pinned by a test that greps the *sink*, since enumerating the callers I expected is exactly how the fourth got missed.

### CHANGELOG corrected

Two claims were measurably wrong and are now what the code does: "the four dispatchable types" (three of five) and "all three framework render paths" (four).

### The gate-off harness was wrong twice, in the same shape

First run: killed mid-mutation, left a mutant in the tree, and the next run read it as baseline. Second run: `--timeout=60` is not a recognized flag here, so pytest exited on an argument error, printed no `N passed`, and my `N failed` regex reported **`0 failed` for all four mutations and the baseline** — every fix looked decorative when in fact nothing had run.

Both are the same defect: **a success condition satisfiable by absence.** The harness now asserts a positive `N passed`, that the executed count matches the baseline, and restores every target via `atexit`. With that, **4/4 load-bearing** — 3, 4, 1 and 1 failures respectively when each is gated off alone.

20,336 passing.

johnrtipton:
## Stage 14 retrospective — PR #2508 (#2501, #2506, #2507)

**Task**: an external user reported in discussion #2437 that `{{ component.render }}` rendered nothing. Shipped as attribute/property/auto-call resolution on the three non-LiveView render paths, plus four security fixes surfaced along the way.

**Quality: 3/5.** The final artifact is sound and well-instrumented. Getting there took **two full review rounds and six defects of my own making**, three of which were the *same class as the bug I was fixing*, and two of which were only visible because an instrument was fixed after it had already reported success.

### What went well

- **Symptom-up triage paid for itself twice.** #2501 blamed the `__dict__` bulk-dump arm. A real `Component`'s `__dict__` holds only `_`-prefixed internals, so it takes the `Encoded` carrier — my own first analysis stood the same test up with a `Plain` stand-in and exercised a different carrier than the reporter's. Posted a public correction rather than quietly re-scoping.
- **The miss-only ordering was the right architecture** and survived hostile probing: 13/13 falsiness cases match Django (a present-but-`None` key does *not* fall through), `dict_view`/`string_index` precedence matches, dunder refusal held 9/9 at every segment.
- **Scope discipline held under pressure.** Four separate opportunities to keep enumerating container types; took none. #2509 states the argument for flooring at the sink instead.
- **The two-round adversarial review earned its cost outright.** Round 1 found a 500 on `Model.DoesNotExist` that hit the LiveView path — wider than anything this PR claimed to touch — and proved it a regression by rebuilding against `main`. Round 2 found three more after I'd already declared the fixes done.

### What didn't

**1. I wrote a security-relevant list from memory.** `_is_user_raised` shipped with three of Django's five dispatchable exception types. `BadRequest` and `MultiPartParserError` are *siblings* of `SuspiciousOperation`, not subclasses, so `isinstance` never reached them and both kept rendering 500 where Django sends 400 — the exact defect the fix existed to close, left half-closed. The repo's own canon (#1516, active falsification; #1197, citation discipline) covers this and I did not apply it to a list I "knew".

**2. Three instruments reported success while measuring nothing.** This is the finding.

| instrument | what it reported | why |
|---|---|---|
| CI watcher | looped 60× | `jq` died on a control character in a check-run body; never evaluated |
| gate-off, run 1 | 5/6 caught | killed mid-mutation, stranded a mutant; run 2 read it as baseline |
| gate-off, run 3 | 0 failed ×4 + baseline | `--timeout=60` unrecognized → pytest exited on an argument error → no `N passed` → the `N failed` regex found nothing |

All three share one shape: **a success condition satisfiable by absence.** The third is the sharpest — every fix looked decorative when in fact *nothing had run*, and I would have concluded the opposite of the truth. The cure is a positive assertion (`N passed` present, count matches baseline), not a better regex.

**3. Two of my own tests were structurally blind.** `raised_type()` unwraps `__cause__`, so every assertion using it was blind to whether the backend wraps — the backend fix had no test and survived gate-off. And `test_a_namedtuple_keeps_its_type` used a 2-field tuple, the one arity where the buggy path is accidentally correct.

**4. I shipped a decorative guard in a PR whose thesis is #1859.** The cycle seed was unreachable (a back-edge returns at a *deeper* depth; the memo key carries depth) while its comment claimed it was what stopped cycles. Zero seed reads under instrumentation, and deleting it left the test green.

**5. I cured #1646 drift while creating it.** The new shared reader's docstring called itself "the single reader" and listed three paths; `rust_bridge.py` was a fourth, untouched. Grepping the sink — the repo's own v1.1.1-2 rule — would have caught it, and the fix now includes a test that does exactly that.

### Verified

- 20,336 passed / 437 skipped / 15 xfailed; `make test-rust` green (bare `cargo test` can't link the extension here).
- Gate-off **6/6** on round 1 and **4/4** on round 2, both after harness repair.
- Independent re-review confirmed the two high-value risks negative: #2506 not re-opened, memo never returns an unprotected object (0 id collisions over 4001 entries).
- Django differential: 9 divergences pre-fix → 5 post-fix, **zero new**.
- 1.8 s → 0.02 ms reproduced bit-exactly by a second party.

### Follow-ups filed

#2504 (dict-view `{% for %}` operand), #2505 (loop-var shadowing), #2509 (floor the tag sink structurally). PR 2 of #2501 — the `serialize_value`/`normalize_django_value` escaping drift — remains open; `{{ component.render|safe }}` works today and the components guide says so.

### The durable lesson

An instrument whose pass condition can be met by silence will eventually report silence as a pass, and it does so most convincingly right when you have stopped expecting bad news. Three times in one PR. Every measurement harness needs a positive liveness assertion — *did the thing actually run* — checked before its result is read.

PR #2512, opened 2026-09-01
fix(components): remove the LiveView-path wrapper-dict special-case (#2503)

### LINKED ISSUE #2503: LiveView path: `{{ component }}` renders `{'render': '<html>'}` — the eager wrapper dict's repr
Found at Stage 5 of #2501 while gating off that PR's LiveView-path change
(#1468), and it corrects a premise #2501's own plan states.

## The cell

```python
class Card(Component):
    def _render_custom(self): return "<b>card</b>"

# in a LiveView's get_context_data: ctx["c"] = Card()
```

```
{{ c }}      django  '<b>card</b>'
             djust   "{'render': '<b dj-id=\"1\">card</b>'}"
```

The LiveView path renders the **repr of a dict**, not the component.

## Mechanism — different from #2501's

#2501 is about the three non-LiveView paths, where a `Component` crosses as
`Value::Encoded` and `{{ c }}` renders `str(component)` with the `SafeString`
marker lost (escaped). This is a fourth path with a fourth shape: the LiveView
sync loop in `python/djust/mixins/rust_bridge.py` replaces a `Component` in the
eager context with

```python
rendered_context[key] = {"render": rendered_html}
```

and marks the key safe. So `{{ c.render }}` resolves — that dict has a `render`
key — and `{{ c }}` resolves to the **dict**, whose `Value::Object` renders as
`{'render': '...'}`.

## Why this is worth its own issue

* It is the same *documented* spelling — *"Components in templates render via
  `{{ component }}`"*, `docs/website/guides/components.md:637` — failing on the
  path a djust app actually runs.
* The fix is not #2501 PR 2's fix. PR 2 restores the `SafeString` marker for a
  component crossing as `Encoded`; this path never produces that carrier. The
  candidate here is to carry the rendered HTML as the value itself (already
  safe-keyed) rather than as a one-key dict, and to answer `{{ c.render }}`
  through the sidecar — which #2501 has now attached for components, so the
  dotted spelling would keep working without the wrapper dict.
* Any such change has to keep `{{ c.render }}` working, so the two spellings
  need one measurement, not two.

## Corrections to #2501's Stage 4 plan, for the record

The plan says un-excluding `Component` from the LiveView sidecar builder is what
makes `{{ c.render }}` resolve there. Measured with the exclusion restored,
`{{ c.render|safe }}` **already rendered** `<b dj-id="1">card</b>` through the
eager `{"render": ...}` dict. What the un-exclusion actually buys is a
component's other names — `{{ c.cls_attr }}` / `{{ c.label }}` / `{{ c.kind }}`
rendered `||` before it and resolve after it — which is the same convergence
#2501 makes on the other three paths.

Pinned in `python/tests/test_sidecar_on_all_render_paths_2501.py`
(`TestTheLiveViewPathsComponentExclusion`), which asserts the current
divergence so this issue's fix turns it red.

### LINKED ISSUE #2511: docs: components.md wrongly claims LiveComponents need {{ c.render }} for the wrapper div — it's identical to {{ c }}, and the worse spelling
## Summary

`docs/website/guides/components.md:637` states:

> **Components in templates render via `{{ component }}`** -- the `__str__` method calls `render()` automatically. For LiveComponents, use `{{ component.render }}` to ensure the wrapper div is included.

Measured false. Both `Component.__str__` and `LiveComponent.__str__` are literally `return self.render()` (`python/djust/components/base.py:394`, `:855`). `str(c)` and `c.render()` are **byte-identical**, wrapper div included, for both classes:

```
str(c)     = '<div data-component-id="livebox_cc116da2"><p>live 1</p></div>'
c.render() = '<div data-component-id="livebox_cc116da2"><p>live 1</p></div>'
identical: True
```

Worse: `.render` is the **fragile** spelling. It depends on Django's nullary auto-call, which djust gates behind `LIVEVIEW_CONFIG['template_auto_call']`. With that flag `False` — a reasonable security posture, since it stops templates invoking arbitrary methods — `{{ component }}` still renders correctly (it goes through `__str__`, no auto-call needed) while `{{ component.render }}` prints a bound-method repr into the page:

```
{{ c|safe }}           -> '<div class="alert">real</div>'                          renders
{{ c.render|safe }}    -> '<bound method Component.render of <Alert object ...>>'  **BROKEN**
```

## How this survived unnoticed

`docs/website/guides/components.md` is not referenced anywhere in `docs/website/_config.yaml`'s nav — it's orphaned from the sidebar. The pages a user actually navigates to, `core-concepts/components.md` and `api-reference/components.md`, already teach the correct spelling (`` `{{ status_dot }}` — calls `__str__()` → `render()` `` at `core-concepts/components.md:34`) and contain neither the false claim nor a `.render` recommendation. So the false claim lives in a file nothing links to — which is presumably why nobody caught it, but doesn't make it harmless: it's still served at whatever URL the guide resolves to, and I cited it in `components.md:639`'s own `|safe` caveat during #2508 without checking the sentence above it.

## Scope

**In scope for this issue:**
1. Fix `components.md:637` — teach `{{ component }}` as the spelling; drop the `.render`-for-LiveComponents claim (it's not true, and `.render` is the worse spelling, not a special case).
2. A regression test pinning `str(component) == component.render()` for both `Component` and `LiveComponent`, using a real *mounted* `LiveComponent` (bare base class, not a subclass override, per this repo's #1947/#1952 canon on testing the base class a subclass might silently diverge from) — including under `LIVEVIEW_CONFIG['template_auto_call'] = False`, so the fragility claim is also pinned and can't silently regress in the other direction (i.e. `.render` should keep *resolving* per #2501/Django-parity even when auto-call is off elsewhere; the two spellings just don't need to be recommended equally).

**Out of scope, filed separately:** ~14 occurrences of `{{ x.render }}` across `docs/components/*.md` and `docs/example-site/*.md` (COMPONENT_MIGRATION_GUIDE, COMPONENT_BEST_PRACTICES, COMPONENT_UNIFIED_DESIGN, NAVBAR_QUICK_START, COMPONENTS.md, NAVBAR_COMPONENT_REPORT, TEMPLATE_CONVERSION_GUIDE). None of these directories are referenced from `docs/website/_config.yaml` or `README.md` — they read as internal design/migration records, not served docs. Sweeping them is real but separate scope (#1079).

**Not decided here:** whether `guides/components.md` should be added to the nav or removed as orphaned. That's a docs-hygiene call for whoever owns `docs/website/_config.yaml`, not something to fold into a one-sentence correctness fix.

## Repro

```python
import django
settings.configure(DEBUG=False, DATABASES={}, USE_TZ=True,
    INSTALLED_APPS=["django.contrib.contenttypes", "django.contrib.auth", "djust"],
    TEMPLATES=[{"BACKEND": "django.template.backends.django.DjangoTemplates",
                "APP_DIRS": True, "DIRS": [], "OPTIONS": {}}])
django.setup()
from djust.components.base import LiveComponent

class LiveBox(LiveComponent):
    template = "<p>live {{ n }}</p>"
    def get_context_data(self):
        return {"n": 1}

lb = LiveBox()
lb.mount()
assert str(lb) == lb.render()  # passes today -- the docs claim they differ
```

johnrtipton:
## What

A question about which spelling to prefer — `{{ component }}` or `{{ component.render }}` — turned out to have a more consequential answer than expected once tested against the actually-documented usage pattern (a component assigned in a LiveView's `mount()`, referenced from its `dj-root` template) rather than only the three non-LiveView paths #2501/#2508 already covered.

On that path, `{{ component }}` rendered the **Python repr of an internal caching wrapper**: `"{'render': '<div>...'}"` — literal braces and quotes, not HTML. I'd independently found and filed this during #2501's own gate-off (#2503), then lost it from context before it resurfaced here.

I also found a second false claim, my own, from #2508's own docs edit: *"The LiveView path is unaffected and needs no `|safe`."* Measured false — `{{ component.render }}` without `|safe` is escaped there exactly like the other three paths.

## Why `{{ component }}` broke on the LiveView path specifically

`_sync_state_to_rust` special-cased `Component`/`LiveComponent`: render once, cache the HTML, replace the value in context with `{"render": html}`. `{{ component.render }}` hit that dict's key and worked. `{{ component }}` hit the *same dict* with no remaining path segments and rendered *its* repr, not the component's.

## The fix

Removed the special-case. Verified byte-identical to the existing generic "complex type" branch immediately below it — this is a pure de-duplication, not new logic. Component/LiveComponent now fall through to the same path every other context object uses, converging onto the mechanism #2501/#2508 already built for the other three paths: `normalize_django_value`'s Component arm answers the bare case; a miss on `.render` (a plain string has no such attribute) falls through to the sidecar's raw object, and #2501's attribute walk answers it there.

`{{ component }}` and `{{ component.render }}` now render **byte-identical** output on all four render paths, checked separately for `Component` (private-dict → `Encoded` carrier) and `LiveComponent` (public-attr → `Object` carrier) — they cross the serialization boundary differently.

## What this costs

Measured, not assumed: `render()` is called **twice** per cycle if a template references `.render` (was once, cached). But the surrounding loop iterates the change-detection *output* — only new/dirty/changed keys — not the full context every tick, so this only bites cycles where the component was already re-rendering, and only templates still using the now-non-recommended `.render` spelling. A template using only the recommended bare `{{ component }}` sees no change (1x, same as before).

## Security re-checked

#2507's `alters_data` mutator-refusal guard (`unmount`, `trigger_update`) lives Rust/sidecar-side and was never part of the removed branch — but this is the first test to exercise it through a *real* `LiveViewTestClient` render rather than the non-LiveView `PATHS` helpers, so I verified rather than assumed it still holds. It does; added as a permanent regression test.

## The instrument that caught the missed update

A prior test — `test_the_bare_spelling_still_renders_the_wrapper_dicts_repr` — was written specifically as a *diverging-direction pin* (per repo canon #1859): designed to go red the moment this bug got fixed, so the fixer couldn't ship the fix without updating it. It fired exactly as designed. Renamed and rewritten to pin the fixed behavior; added a `LiveComponent`-shape counterpart (`LiveCard`) since the existing `Card` fixture only covers the `Component`/`Encoded`-carrier shape.

## Verification

- Full suite: 20,339 passed, 0 failed, 0 regressions (`tests/ python/tests/ python/djust/tests/`, `-n auto`).
- Gate-off: reintroduced the removed branch, confirmed 3 tests fail without the fix; restored, confirmed 0 failures.
- 863 component-related tests pass unaffected.

## Docs

Corrected both false claims in `docs/website/guides/components.md`, and added the same `|safe` caveat to the two pages that are actually reachable from the site nav (`core-concepts/components.md`, `api-reference/components.md`) — their `{{ status_dot }}` example currently renders escaped without it. Noted, but did not fix, that `guides/components.md` itself isn't linked from `docs/website/_config.yaml`'s nav — likely why the false claim went unnoticed; a separate docs-hygiene call.

Closes #2503
Closes #2511


johnrtipton:
## Fixes applied — REQUEST_CHANGES addressed

### 🔴 Parallel-path twin (item 4) — fixed

`_render_full_template_inner`'s page-shell render (`python/djust/mixins/template.py`) had the identical `Component`/`LiveComponent` wrapper-dict special-case this PR removed from `rust_bridge.py`, on a branch this PR never touched. Confirmed via the reviewer's exact repro shape, reproduced independently.

Fixed by making the buggy `else` branch structurally match its own sibling (`serialized_context is not None`), which never special-cased Component at all — both now pass raw context straight to `normalize_django_value`. As a bonus, this also stops stripping the SafeString marker `render()` returns (the old code explicitly wrapped it in `str()`).

`.render` doesn't resolve on this path in *either* branch — no sidecar is wired for the shell's temporary `RustLiveView` — but that's a pre-existing gap the already-correct sibling shared before this PR touched anything, not something introduced or fixed here. Filed **#2513** rather than folding it in, matching the review's own suggestion.

Added `TestTheShellRenderPathHadTheSameWrapperDictBug` with the reviewer's own reproduction shape, including a test that both branches now agree (the actual #1646 invariant — parity, not necessarily full correctness). Gate-off: reintroducing the removed branch fails 2 of 3 new tests.

### 🟡 `_force_full_html` cost (item 2b) — disclosed, not fixed

Confirmed: a WS-reconnect/session-restore render pays the extra `.render()` call for every `.render`-referencing component even when untouched, since that path sets `context == full_context` rather than the change-detection subset. This is a real, narrower-than-claimed bound — added to the CHANGELOG rather than silently leaving the original "bounded to already-dirty cycles" claim standing uncorrected.

### 🟢 Nits

- CHANGELOG corrected from "a second time" to "K uses cost K+1 calls" (was undercounting for `.render` referenced more than once).

## Re-verification

- Full suite: 20,342 passed, 0 failed, 0 regressions.
- Gate-off on the new fix: 2/3 new tests fail with the old branch reintroduced; 0 failures restored.

Thank you for the item-4 catch in particular — that's exactly the class of bug this PR's own commit message was trying to close, sitting one file over.

johnrtipton:
## Re-review of the fix pass → APPROVE

Independent re-review of `f4894a6f`/`df022bdf`, high confidence. Every falsification attempt came back consistent:

- **Both branches produce identical output**, checked with a fresh fixture (different names than mine) — bare renders correctly on both, `.render` empty on both.
- **No escaping regression.** Ran an attacker-content probe (`<script>`-shaped text a Component author trusts via `mark_safe`) against the OLD code, the pre-shell-fix state, and this fix. Found something worth recording: the OLD code's bare `{{ c }}` leaked that payload unescaped *inside the dict-repr garbage text*, and its `.render` *accidentally* worked only because the wrapper dict happened to have a literal `"render"` key. Post-fix, `.render` on this branch is empty (matching the sibling), which is a real, disclosed, tested behavior change — not a security regression (empty is the safe default).
- **No third copy of the bug anywhere in the repo** — grepped `isinstance(value, (Component, LiveComponent))` (6 hits, all unrelated: session-restore, ID-assignment, registration) and `{"render":` (only comments/unrelated dicts). This was the highest-value thing to check given two copies already existed, and it came back clean.
- **The `_force_full_html` cost claim verified exactly**, with real call-counting: K=3 references → 4 calls (K+1, not a flat 2) on first render; 0 calls on an untouched steady-state cycle; 4 calls under `_force_full_html` even untouched; 1 call for the bare spelling used 3×. Every number in the corrected CHANGELOG entry checks out.
- **Independent gate-off**: reintroduced the old branch myself (not trusting the PR's own report), confirmed 2 of 3 new tests fail, restored, confirmed 3/3 pass.
- Full suite: 20,341 passed, 0 failed (1 off from the commit's claimed count, zero failures either side — reads as skip/xfail environment variance, not a regression).

No blocking findings. Nice catch on your own PR, twice in a row.

johnrtipton:
## Stage 14 retrospective — PR #2512 (#2511, #2503, #2513)

**Task**: a user question — "is `{{ component.render }}` the desired syntax, or should it just be `{{ component }}`?" — following #2508. What started as a proposed docs-only fix became a real bug fix across two render paths, because the honest answer required testing the actually-documented usage pattern rather than the paths already covered.

**Quality: 4/5.** The end state is correct and thoroughly verified. The path there had one real miss (a second copy of the bug, found only by review) and two premature conclusions I self-corrected before shipping — both worth recording since they're a repeatable failure mode, not one-offs.

### What went well

- **Testing the actual usage pattern rather than a proxy caught a real bug before it shipped as a doc claim.** My first answer to the user ("prefer `{{ component }}`") was tested only through the three non-LiveView paths. Testing the LiveView-embedded pattern — the one the docs actually teach — surfaced that `{{ component }}` was outright broken there (a dict repr), not just "the worse spelling." Stopped and re-scoped before writing a single line of the wrong recommendation.
- **A prior finding resurfaced on its own schedule.** #2503 had been found and filed during #2501's own gate-off, then genuinely lost from context. It came back not because I remembered it, but because the SAME investigation method (test the real path) rediscovered the SAME symptom independently. That's a point in favor of "verify, don't recall" as a default — recall failed; re-verification didn't.
- **The user's own choice of fix approach (converge onto the existing mechanism) turned out to be empirically correct where my initial worry was wrong.** I predicted the sidecar/attribute-walk convergence would fail for LiveComponent's public-attr shape (dumping raw internals instead of calling render()). The empirical test proved that wrong in the good direction — `normalize_django_value`'s existing Component-aware arm answered it. Worth noting: my Rust-mechanics reasoning was wrong twice in this task (once pessimistic, caught by testing) — the empirical check, not the mental model, was load-bearing both times.

### What didn't

**A second copy of the bug existed one file over, and I didn't grep for it.** This repo's own canon (#1646) says explicitly: grep every parallel implementation of an invariant before declaring it fixed. I fixed `rust_bridge.py`'s special-case and moved straight to docs, without checking whether `Component`/`LiveComponent` had a SECOND eager-wrap site elsewhere. It did — `template.py`'s page-shell render, byte-for-byte the same shape. Caught only by Stage 11 review, which is the review doing its job — but the grep would have cost under a minute and closed it in the first pass instead of the second.

**Action taken**: `diff` — grepping for the pattern is now written into the follow-up issue #2513's own scope note, and the second fix commit (`f4894a6f`) documents the grep that should have run first. No new CLAUDE.md rule needed — #1646 already states this; the miss was not following an existing rule, not a missing one.

**A perf claim ("bounded to cycles where the component was already re-rendering") was true but incompletely quantified.** I verified the *shape* of the bound (iterates change-detection output, not full context) but didn't check every path that sets `context == full_context`. Review found `_force_full_html` (WS-reconnect/session-restore) as an exception; I hadn't tested that specific trigger.

**Action taken**: `diff` — CHANGELOG corrected in `df022bdf` to state the exception explicitly, with the K-references-cost-K+1-calls detail also fixed (was "a second time," undercounting for K>1).

### Insights

- Two review rounds, two real findings each round on this pair of PRs (#2508: 3🔴+3🟡 then 3🟡 more; #2512: 1🔴+1🟡). Neither round was clean on the first pass. The pattern holds across both PRs in this session: **the second-order check (re-review after my own fix pass) is not decorative** — it found #2508's `_is_user_raised` dispatch-set gap and the arity-1 namedtuple corruption, both real, both would have shipped without it.
- The gate-off harness itself needed fixing three separate times across this session (stranded mutant read as baseline; unrecognized pytest flag reported false success; both in #2508). No new instrument failure this round — the harness fixes from #2508 held up.
- Scope discipline (#1079) held under real pressure to over-fix: declined to wire a sidecar for the page-shell's `.render` gap (filed #2513 instead), declined to "fix" the demo project's inconsistent `serialized_context` forwarding (illustrative of reachability, not itself the bug), declined to touch `guides/components.md`'s nav-orphan status.

### Review Stats

| Metric | PR #2512 |
|--------|-------|
| Tests added | ~9 (Component/LiveComponent parity, mutator-refusal-on-LiveView, shell-path parity) |
| 🔴 Findings | 1 |
| 🟡 Findings | 1 |
| Findings fixed | 2/2 |
| CI failures | 0 |

### Process Improvements Applied

**CLAUDE.md**: none this round — the miss was a known rule (#1646) not applied, not a gap in canon.
**Follow-ups filed**: #2513 (page-shell `.render` sidecar gap, pre-existing, shared by both branches, deliberately not folded in).

### Open Items

- [ ] #2513 — page-shell `.render` sidecar gap (pre-existing, both branches, disclosed)
- [ ] PR 2 of #2501 — the `serialize_value`/`normalize_django_value` escaping drift, still open, referenced throughout this PR's docs

RETRO_COMPLETE

PR #2514, opened 2026-09-01
fix(core): stop iterating a live dict during recursive Value extraction (#2510)

### LINKED ISSUE #2510: 1.2.0rc1: Rust panic 'dictionary changed size during iteration' rendering any LiveView when context_processors has request but not auth
## Summary

On **1.2.0rc1**, rendering any LiveView panics in Rust when the project's
`context_processors` contains `django.template.context_processors.request`
**and does not contain** `django.contrib.auth.context_processors.auth`:

```
thread '<unnamed>' panicked at pyo3-0.29.2/src/types/dict.rs:578:21:
dictionary changed size during iteration
...
  File ".../djust/mixins/request.py", line 293, in get
  File ".../djust/mixins/template.py", line 995, in render_full_template
  File ".../djust/mixins/template.py", line 1078, in _render_full_template_inner
pyo3_runtime.PanicException: dictionary changed size during iteration
```

Every page of the affected project returns 500. **1.1.0 is unaffected** on the
same project and the same config.

## Reproduction

Complete project — four files.

`settings.py`
```python
SECRET_KEY = "repro"
DEBUG = True
ALLOWED_HOSTS = ["*"]
ROOT_URLCONF = "urls"
INSTALLED_APPS = [
    "django.contrib.contenttypes",
    "django.contrib.auth",
    "django.contrib.sessions",
    "djust",
    "channels",
]
MIDDLEWARE = [
    "django.middleware.security.SecurityMiddleware",
    "django.contrib.sessions.middleware.SessionMiddleware",
    "django.middleware.common.CommonMiddleware",
    "django.middleware.csrf.CsrfViewMiddleware",
    "django.contrib.auth.middleware.AuthenticationMiddleware",
]
TEMPLATES = [{
    "DIRS": [__file__.replace("settings.py", "templates")],
    "APP_DIRS": True,
    # The trigger is here: `request` present, `auth` absent.
    "OPTIONS": {"context_processors": ["django.template.context_processors.request"]},
}]
DATABASES = {"default": {"ENGINE": "django.db.backends.sqlite3",
                         "NAME": __file__.replace("settings.py", "db.sqlite3")}}
USE_TZ = True
STATIC_URL = "/static/"
ASGI_APPLICATION = "asgi.application"
CHANNEL_LAYERS = {"default": {"BACKEND": "channels.layers.InMemoryChannelLayer"}}
```

`views.py`
```python
from djust import LiveView


class HelloView(LiveView):
    template_name = "hello.html"

        self.greeting = "hello"
```

`urls.py`
```python
from django.urls import path
from views import HelloView
urlpatterns = [path("", HelloView.as_view(), name="hello")]
```

`templates/hello.html`
```html
<!doctype html>
<html><body>
<div dj-root>
  <p>{{ greeting }}</p>
</div>
</body></html>
```

Then:

```
python manage.py migrate
python manage.py runserver --skip-checks
curl -o /dev/null -w '%{http_code}\n' http://127.0.0.1:8000/
```

`500`, and the panic above on stderr.

## The trigger is the context-processor pair

Same project, only `OPTIONS.context_processors` changed:

| context_processors | 1.2.0rc1 | 1.1.0 |
|---|---|---|
| *(empty)* | 200 | 200 |
| `request` | **500 — panic** | 200 |
| `auth` | 200 | 200 |
| `request` + `auth` | 200 | 200 |

Adding `auth` back, or removing `request`, avoids it. The view, the template
and the context are otherwise untouched between rows.

## What I could not determine

I assumed a lazily-loaded session mutating a dict mid-iteration, but forcing
both `request.session` and `request.META` to load inside `mount()` before
render does **not** avoid the panic — so that guess is wrong and I have not
identified the actual dict.

## Why it matters

`request` without `auth` is a normal configuration for an app that has no
users — this surfaced on a single-user local-first Django app where nothing
authenticates, so `auth` was never added. Every page of that app 500s on
1.2.0rc1 and is fine on 1.1.0.

It is also invisible to tests: the app's suite (68 tests, `LiveViewTestClient`)
passes completely on 1.2.0rc1 while every page served over HTTP returns 500.
The panic is on the `render_full_template` path, which the test client does
not exercise.

A Rust panic crossing into Python as `pyo3_runtime.PanicException` is also
hard to attribute from a traceback — the Django frames above it point at the
view, not the engine.

## Environment

- djust 1.2.0rc1 (wheel from PyPI, `cp311-macosx_11_0_arm64`)
- Django 5.2.17, Python 3.11.14
- macOS, arm64
- Also reproduced on Python 3.11 with `DEBUG=False`, and with
  `COMPANION_DEBUG=1`/`DEBUG=True` — the panic is not DEBUG-dependent.

johnrtipton:
## What

Every LiveView page 500'd with a Rust panic — "dictionary changed size during iteration" — for any project whose `TEMPLATES.OPTIONS.context_processors` has `request` but not `auth`. Reported against 1.2.0rc1 with an excellent minimal repro; confirmed still live on current `main` (untouched by today's other work) via a real WSGI dev server.

## Root cause

Found by reading `crates/djust_core/src/lib.rs`, not by guessing. Two arms of `impl FromPyObject for Value` hold a **live** PyO3 iterator directly over a Python dict and, for each value, recursively call `.extract::<Value>()` — which can run arbitrary Python:

1. `public_dict_attrs` — the `__dict__` bulk-dump carrier (reached by any "truthy, non-iterable object with public attributes" — a real `HttpRequest` is exactly this shape once the `request` context processor puts it in context).
2. The plain-`PyDict` map-building arm a few lines above it, for any genuine dict whose values include something reentrant.

An unresolved `SimpleLazyObject` (`request.user`, before anything has forced it) triggers Django's lazy `_setup()` on the first dunder check the extraction chain performs. `AuthenticationMiddleware.get_user`:

```python
def get_user(request):
    if not hasattr(request, "_cached_user"):
        request._cached_user = auth.get_user(request)   # writes request.__dict__
    return request._cached_user
```

writes a **new key into `request.__dict__`** — the exact dict `public_dict_attrs` is mid-iteration over. `auth`'s context processor normally forces that resolution earlier, before djust's walk ever runs; without it, the write happens live, mid-iteration, and PyO3's iterator invariant breaks. The template doesn't even need to reference `.user` — the walk dumps the whole `__dict__` regardless of which attribute was asked for. This explains every row of the reporter's own table exactly.

## Fix

Snapshot each dict's items into an owned `Vec` before any recursive extraction, in both arms. A snapshot fully drains the PyO3 iterator before any Python callback can run, so a later mutation has nothing left to invalidate.

Deliberately structural, not instance-specific: covers any object whose `__dict__` contains an unresolved lazy attribute, not just `HttpRequest`. Considered and rejected filtering `HttpRequest` out of `request.py`'s `state`/`serialized_context` construction instead — that closes this report's specific trigger and leaves the class open for the next object shape that hits it (a lazy queryset, a proxy, anything with reentrant attribute access).

## Reproduction

A real WSGI dev server (`manage.py runserver` + curl, matching the report's 4-file project exactly) reproduces it. `LiveViewTestClient` and `django.test.Client` did **not** — even with settings matched byte-for-byte, `MIDDLEWARE` fully replaced (not appended), and hot-reload force-enabled over 20 consecutive requests. I didn't fully resolve why, and it isn't load-bearing for the fix: a 6-line, Django-free unit test reproduces the identical panic directly at the Rust boundary —

```python
class LazyLike:
    def __bool__(self):
        if not self._resolved:
            self._resolved = True
            self._parent.newly_added_key = "x"   # mutates parent.__dict__
        return True

class FakeRequestLike:
    def __init__(self):
        self.path = "/"
        self.user = LazyLike(self)

_rust.render_template("{{ c.path }}", {"c": FakeRequestLike()})
```

— no Django settings, no WSGI. That's what the regression suite exercises.

## Two tests were wrong on the first pass — both caught before trusting a green run

- One test claimed to cover `functools.cached_property` as an equivalent trigger. It passed — for the wrong reason: `cached_property` only writes `__dict__` on **attribute access**, and `public_dict_attrs` iterates `__dict__` directly without ever calling `getattr()` for a name that isn't already a key. An un-accessed `cached_property` can never panic this way, regardless of the fix. Kept as a documented negative.
- The genuine-`PyDict`-arm test used attribute-set syntax (`self._parent.newly_added_key = ...`) against a `dict` parent — which raises `AttributeError` inside `__bool__`, silently swallowed by the Rust side's `if let Ok(...)` pattern. The mutation never actually happened; the test was green because nothing was tested. Fixed to use `self._parent["newly_added_key"] = ...` before trusting the result.

## Parallel-path sweep

Grepped every remaining `.iter()` in the conversion boundary for a third vulnerable site: `collect_named_attrs`/`collect_called_attrs` iterate a fixed Rust-side name table (not a live `PyDict`); `has_public_dict_attrs` touches only keys, never values, so it has no reentrancy path. None found.

## Verification

- Gate-off: reverted both fixes, confirmed 4/6 new tests fail; restored, confirmed 6/6 pass.
- Full suite: 20,348 Python passed, 0 failed; Rust suite (`make test-rust`) green.
- Real repro: two consecutive requests against the exact reported project, both 200, no panic.
- `clippy`/`cargo fmt` clean.
- Benchmarks (`tests/benchmarks/test_request_path.py`, `test_serialization.py`, `test_template_render.py`): no threshold failures. The fix adds one bounded `Vec` allocation per object/dict converted via these two arms (proportional to attribute/key count, already capped elsewhere) — not the primary path any existing benchmark specifically targets, but no regression signal.

Closes #2510


johnrtipton:
## Fixes applied — REQUEST_CHANGES addressed

### 🔴 Three more unfixed sites (confirmed independently, fixed) — plus two more found while fixing

Independently reproduced all three of your confirmed repros before touching anything (`fast_json_dumps`, `serialize_models_fast`, the actor path — all panicked exactly as reported). Fixed with the identical Vec-snapshot pattern:

- `fast_json_dumps` → `python_to_json_value` (`crates/djust_live/src/lib.rs`) — both the `PyList` and `PyDict` arms.
- `serialize_models_fast` → `python_dict_to_json` (`crates/djust_live/src/model_serializer.rs`), plus its `python_to_json` sibling's list/tuple arms — `.iter().map(...).collect()` is just as live as a bare `for` loop; the laziness of `.map()` doesn't snapshot anything.
- The actor path's `python_dict_to_hashmap`/`python_to_value` (`crates/djust_live/src/lib.rs`) — tuple, list, and dict arms.
- Your two inspection-only flags (`actors/component.rs`, `actors/view.rs`) — also fixed. Their outer `context_dict.iter()` loop was unprotected even after the inner `value.extract::<Value>()` call was already made safe by the first commit's `djust_core` fix — the inner fix alone doesn't protect the outer iterator from a value's conversion mutating `context_dict` itself.

Corrected the CHANGELOG's completeness claim, which was accurate for `djust_core` and silently scoped there.

### Reproduction, matching your exact repros

Used your `__index__`-based trigger (not my original PR's `__bool__`-based one) since these converters check `i64` before `bool` — a `__bool__`-only trigger would never reach them. Added as `_IndexTrigger` in the test file, with a comment explaining why it's a different shape than `_LazyLike`.

### Perf, ordering, mutation semantics

Took your empirical findings on these (no regression, order preserved, removed-key handled safely, the reentrant-add-of-public-key characterization) as verified rather than re-deriving them — they're not affected by extending the same fix to more sites.

## Verification

- Gate-off, independently, per site: reverted each of the three confirmed fixes one at a time, rebuilt, confirmed exactly the corresponding new test failed (1 failed / 8 passed each time); restored, rebuilt, confirmed 9/9.
- `clippy`/`cargo fmt` clean on `djust_live`.
- Full suite: 20,351 Python passed, 0 failed; `make test-rust` green.

Thank you for not letting "grepped everything" stand when it wasn't true — that's exactly the class of thing I should have caught myself before writing it.

johnrtipton:
## A fourth site, found by re-review of the fix pass — fixed

`serialize_models_to_list` — a sibling public export of `serialize_models_fast`, in the same file the last commit already touched — routes through entirely separate, never-fixed helpers: `normalize_dict`/`normalize_value`. Reproduced independently with a third trigger shape (a `__str__` override — `normalize_value`'s primitive check is `is_instance_of`, a strict type check neither of the two existing triggers in the test file reaches):

```python
class _MutOnStr:
    def __str__(self):
        if not self._fired:
            self._fired = True
            self._owner_dict["sneaky"] = True
        return "stringified"

_rust.serialize_models_to_list([{"z": _MutOnStr(...), "k2": "v2"}])
# panicked on this PR's tip before this commit; fixed now
```

Same fix, same shape — fixed `normalize_dict`'s dict arm and `normalize_value`'s list/tuple arms, plus the outer `serialize_models_to_list` loop for consistency with how the rest of this PR treats every sequence-shaped arm in a function, not just the one a repro happens to hit.

Independently re-swept every remaining file in `crates/` referencing `PyDict`/`PyList`/`PyTuple`/`PySequence` before committing (`djust_components/src/python.rs`, `djust_templates/{registry,filter_registry,stringformat}.rs`) — confirmed none reachable by this bug class: strict-typed extraction with no dunder-invoking coercion, Rust-native iteration, or (`mark_item`) a producer-enumeration invariant guaranteeing the items are always plain strings.

This is the third round of "found more instances" on this PR — first commit (2 sites in `djust_core`), second commit (3 confirmed + 2 proactive in `djust_live`), this one (1 more, in a file already touched). Recording that plainly rather than as a single clean narrative, per this repo's own canon on chain-shaped issues.

## Verification

- Gate-off: reverted `normalize_dict`, rebuilt, confirmed the new test fails (1 failed / 9 passed); restored, rebuilt, confirmed 10/10.
- `clippy`/`cargo fmt` clean.
- Full suite: 20,352 Python passed, 0 failed; `make test-rust` green.

johnrtipton:
## The literal original bug — fixed at the top level

Your headline finding was the most important one in this whole cycle: the exact reproduction shape from the original issue — a top-level context dict mutating itself — was still exploitable, because every fix so far protected NESTED arms inside `impl FromPyObject for Value`. `render_template`, `render_template_with_dirs`, and `RustLiveView.update_state` all convert their OUTERMOST dict through PyO3's own blanket `HashMap<K, V>: FromPyObject` — dependency code no edit to `impl FromPyObject for Value` can ever reach, and no `guard_panic` wrapper can catch, since the panic happens in PyO3's FFI argument extraction before any hand-written function body runs. `update_state` is the literal call the original `AuthenticationMiddleware`/`request.user` repro goes through.

Independently reproduced with your exact repro shape (`__index__`-mutating-the-top-level-dict) against `render_template`, `render_template_with_dirs`, and `update_state` — all three panicked pre-fix, none do post-fix.

Fixed with one shared helper (`snapshot_context_to_value_hashmap`) at all three entry points. `update_state` needed splitting into a safe Python-facing wrapper plus a shared `apply_state_update` core, because it has a second, Rust-only caller (`update_state_rust`, used by other Rust crates with no Python object at all — genuinely no reentrancy risk there, and no way to hand it a `&Bound<PyAny>` since it never has one) — that caller now goes through the same shared core, unchanged behavior, no signature change on its own side.

## The two hand-written-loop sites

Both confirmed and fixed:
- `registry.rs`'s assign-tag-handler result coercion — a public extension point (`register_assign_tag_handler`). This sits right next to the two arms an earlier round's commit claimed to have already checked in this file — that claim was false, and I've stopped writing "swept this file" without meaning "swept every function in it."
- `serialize_context`/`serialize_python_value` — list, tuple, nested-dict, and the outer context dict.

## Verification

- Independently re-ran the full existing test file (15 cases, spanning every prior round's fix) — all still pass.
- Gate-off, per new fix, all four: reverted each individually, rebuilt, confirmed exactly its own test fails; restored, rebuilt, confirmed 15/15.
- `clippy`/`cargo fmt` clean on both touched crates.
- Full suite: 20,357 Python passed, 0 failed; `make test-rust` green across every crate (including `djust_templates`'s 497-test suite, where `registry.rs` lives).

This is the fourth round of findings on this PR. I'd rather it took four rounds and shipped correct than fewer rounds and left the original bug's own trigger still live.

johnrtipton (commented):
**APPROVE** — Stage 11 re-review, HEAD 527d4fbd, 0 behind main.

**Sweep against the sink.** The panic exists only in pyo3's `BoundDictIterator` (`dict.rs:578`); `BoundListIterator` clamps (`list.rs:505`) and `Vec<T>` extraction uses `try_iter()`, so lists/tuples were never in the class. Every `PyDict` `.iter()` in `crates/` is now snapshotted (core 2027/3022; live 1840/2040/2447/2554/2765/2816; model_serializer 72/211; actors 356/433; registry 926). Remaining blanket extractions: `crates/djust_templates/src/lib.rs:283,314` still take `HashMap<String, Value>` but the crate is rlib-only and `Template` is not registered on `_rust` — unreachable; `entry_sidecar` (live:1855) extracts `Py<PyAny>`, no Python runs. No reachable unfixed site.

**Verified:** worktree gate-off — reverting `update_state` to blanket extraction fails exactly `test_update_state_top_level_mutation_does_not_panic` (1 failed / 14 passed). Probed add, delete, same-size swap, nested-to-top, dict subclass, `__str__`/`__bool__` at top level: no panic. `update_state_rust` → `apply_state_update` unchanged, no double snapshot.

🟡 should-fix:
1. Unpinned contract: a key ADDED mid-conversion is silently absent (`{{ late }}` → ""); a REMOVED key still renders from the stale ref. Pin both.
2. `crates/djust_live/src/lib.rs:2024-2031` claims a list mutation "breaks the live iterator's invariant" — false for pyo3 lists. Fix the comment.
3. #1104: no test reaches `actors/component.rs:356`, `actors/view.rs:433`, or the nested-dict arms at live:2554/2816.
4. `snapshot_context_to_value_hashmap` runs outside `guard_panic` (live:1877/1928).

Out of scope, please file: a public reference cycle in context (`d['t']=C(d)`, `C.d=d`) segfaults via unbounded `__dict__` recursion — pre-existing.

Suite: Python 20,357 passed / 0 failed (one lockfile-registry failure was my build's `uv.lock` rewrite; restored, re-run green). Cargo 1357 passed / 0 failed.

johnrtipton:
## 🟡 findings addressed

Fix pass for the PR #2514 Stage-11 review (reviewed HEAD 527d4fbd). Commits: `9577a9f4` (implementation + tests) and `5d181309` (CHANGELOG). All four findings closed.

### 1. Unpinned contract

**What changed.** `TestTheSnapshotContractIsPinned` (`python/tests/test_dict_mutation_during_iteration_panic_2510.py`) pins both halves on both Python entry points that route through the helper:

- `render_template`: a key ADDED mid-conversion is absent from the render (`[{{ late }}][{{ other }}]` → `[][y]`, with `d["late"] is True` afterwards proving the mutation did happen); a key REMOVED mid-conversion still renders from the snapshot (`[{{ doomed }}]` → `[still-here]`, with `"doomed" not in d` afterwards).
- `RustLiveView.update_state` + `render()`: the same two cases.

A new `_RemoveTrigger` (an `__index__` that deletes a sibling key) drives the removed half; the doomed key is inserted after the trigger so it is in the snapshot but gone by the time the loop reaches it.

The contract is stated in `snapshot_context_to_value_hashmap`'s doc comment (`crates/djust_live/src/lib.rs`, section "The contract the snapshot creates") and in one sentence on the `[Unreleased]` #2510 CHANGELOG entry.

**Gate-off, three ways** (each rebuilt, run, restored byte-for-byte):

- helper reverted to a live `context.iter()`: 7 failed / 18 passed — the 4 contract tests plus the 3 existing top-level tests, all `PanicException`;
- helper made to skip keys no longer in the dict (`if !context.contains(&key)? { continue; }`): exactly the 2 `removed` tests fail, 23 pass;
- helper made to re-scan the dict for keys added since the snapshot: exactly the 2 `added` tests fail, 23 pass.

Incidental, fixed while editing that doc comment: the helper had been inserted between `entry_sidecar`'s 40-line doc comment and `entry_sidecar` itself, so rustdoc attached that block to the helper. The helper now sits above the block with its own doc.

### 2. False comment

Verified against `~/.cargo/registry/src/*/pyo3-0.29.2/src/types/list.rs`: `BoundListIterator::next_unsynchronized` recomputes `length.0.min(list.len())` on every step, so a list that grows or shrinks under a live iterator ends early or late and never panics. Only `DictIterImpl::next_unchecked` (`dict.rs:578`) compares `di_used` to the live length and panics. The `python_to_json_value` list-arm comment (`crates/djust_live/src/lib.rs`, the `PyList` arm) now says exactly that, and gives the reasons the snapshot is kept anyway: it is the one shape every converter in this class uses, it makes the set of converted elements independent of a conversion side effect, and it does not lean on a clamp PyO3 documents nowhere. The `python_to_value` tuple arm, which pointed at the dict rationale as "identical shape", is corrected the same way (a tuple cannot change size at all).

### 3. #1104 — the four untested sites

Each traced up to a `djust._rust` entry; each test mutates the dict THAT site iterates (not a parent or child of it) and names the site in its docstring (`TestTheFourSitesNoTestReached`):

| site | Python entry that reaches it | test |
|---|---|---|
| `lib.rs` `serialize_python_value`, `PyDict` arm | `serialize_context({"outer": nested})` | `test_serialize_python_value_nested_dict_arm` |
| `lib.rs` `python_to_value`, `PyDict` arm | `SessionActorHandle.mount(path, {"nested": nested}, None)` | `test_python_to_value_nested_dict_arm` |
| `actors/view.rs` `ViewActor::sync_state_from_python` | `mount(path, {}, python_view)` then `event("poke", {})` | `test_view_actor_sync_state_from_python` |
| `actors/component.rs` `ComponentActor::sync_state_from_python` | `create_component(..., python_component)` then `component_event(...)` | `test_component_actor_sync_state_from_python` |

Why nothing reached them before: the existing actor test mutated the TOP-LEVEL params dict, which is `python_dict_to_hashmap`'s loop, not `python_to_value`'s; the existing `serialize_context` test likewise only reached `serialize_context_py`'s own loop one level up. Both actor sites are reachable from Python with no new harness: `SessionActorHandle` already accepts a Python view / component object and calls its `get_context_data()` after every event, which is exactly the dict those two loops iterate. The component test also asserts the post-sync render (`<div>[y][42]</div>`), so the state that was synced is proven to be the snapshot's.

**Gate-off:** reverting any ONE site to a live `.iter()` (four separate rebuilds) fails exactly that site's test — 1 failed / 24 passed each time — and no other.

### 4. guard_panic coverage

`render_template` and `render_template_with_dirs` now call `snapshot_context_to_value_hashmap` (and `entry_sidecar`) INSIDE their `guard_panic` closure (`crates/djust_live/src/lib.rs`). There was no borrow or lifetime obstacle: the closure is a plain `FnOnce` under `AssertUnwindSafe`, and `context: &Bound<PyAny>` is captured by copying the reference. No known input panics inside the snapshot today (that is what the snapshot is for), so the change is structural. `TestTheSnapshotRunsInsideGuardPanic` pins it at source level — the call must appear after `guard_panic("<entry>"` within each fn body — parametrized over both entries, in the same idiom as `test_bool_before_int_converters_2212.py`. Gate-off: moving `render_template`'s calls back outside the closure fails exactly the `render_template` case; the `render_template_with_dirs` case stays green.

The helper's own doc comment still records that the ORIGINAL panic (inside PyO3's blanket `HashMap` extraction) could never have been guarded, and now adds why the hand-written replacement is.

### Verification

- New tests: 9 functions (10 collected; the structural pin is parametrized ×2). File total 25, all passing.
- Gate-offs: 8 mutations (3 for finding 1, 4 for finding 3, 1 for finding 4). Each asserted present-and-unique, asserted to change the source, rebuilt where the mutation was in Rust, run, and restored byte-for-byte. Every one failed exactly as expected. One harness lesson recorded: the source-only gate-off for finding 4 first showed a second failure that was a stale `.so` left by the previous rebuilt gate-off; rebuilding from the true source before trusting any run cleared it.
- `cargo fmt --all -- --check` clean; `cargo clippy -p djust_live --all-targets -- -D warnings` clean; `ruff check` + `ruff format --check` clean.
- Full Python suite (`tests/ python/tests/ python/djust/tests/ -n auto`): **20367 passed, 0 failed**, 437 skipped, 15 xfailed (8:53 wall). The review's run had 20,357 passed; the +10 are exactly the new collected items.
- `make test-rust` (both phases): **1398 passed, 0 failed** (Phase 1 workspace-minus-djust_live 1357, Phase 2 djust_live `--no-default-features` 41), exit 0. A first run had failed in `djust_templates`' doctests with "can't find crate for `djust_components`" — a target-dir artifact race with commit 1's pre-commit `cargo clippy`, which ran concurrently; `renderer.rs` is byte-identical to main and the clean re-run is green.
- `uv.lock` untouched throughout (built with `.venv/bin/maturin develop` directly, never `uv run`).

johnrtipton:
## Retrospective — PR #2514 (#2510)

Quality: 4/5 — the merged fix is structurally complete (every `PyDict` `.iter()` in `crates/` snapshotted, the snapshot contract pinned, 8 gate-off mutations each failing exactly the intended test), but it took five "found more" rounds and two corrections of the CHANGELOG's completeness claim.

### What went well
- Reproduce-first held every round: each new site (`fast_json_dumps`, `serialize_models_fast`, the actor path, `normalize_dict`, the top-level entry points) was reproduced before being fixed.
- The re-review settled the CLASS at the sink by reading pyo3 0.29.2: only `BoundDictIterator` panics; `BoundListIterator` clamps and `Vec<T>` uses `try_iter()`. One check bounded the sweep to dicts and corrected a comment claiming a list-iterator panic.
- The 🟡 fix pass pinned the snapshot CONTRACT, not just "no panic": a key added mid-conversion is absent, a removed key still renders. The top-level snapshot moved inside `guard_panic`, with a source-level pin.
- The reviewer's out-of-scope probe filed #2516 (a plain reference cycle segfaults via unbounded `__dict__` recursion) instead of folding it in.

### Lessons learned
- **The sweep should have been sink-first at round 1.** Rounds 1–4 each grepped for callers shaped like the previous find; round 4's site (the LITERAL original trigger, still live because PyO3's blanket `HashMap` argument extraction runs before any function body) was reachable from a first-round grep for `PyDict` iteration across `crates/`. The pyo3-source check took minutes and would have ended the chain at round 2.
- **A completeness claim needs a structural pin, not prose.** The CHANGELOG said "both vulnerable arms" and was corrected twice (c07870a1, then after the top-level find). Nothing asserts that no unsnapshotted `PyDict` iterator exists, so the next site will be found by review again.
- **Environment hazards:** `make dev-build` / `uv run` re-lock `uv.lock` against a local PyPI proxy and fail `test_the_repos_own_lockfile_is_clean` (use `.venv/bin/maturin develop`; never stage `uv.lock`). An SSH push died with exit 141 AFTER the 9-minute pre-push hook passed, leaving origin stale (confirm `git rev-parse HEAD` equals `origin/<branch>`). Pre-commit clippy plus a concurrent `make test-rust` in one target dir produced a spurious "can't find crate for djust_components" doctest failure.

### Relation to prior work
- The v1.1.1-2 "Grep for the SINK, not for the callers you expect" rule, first missed then applied. The sink was pyo3's dict iterator, not any djust function.
- A chain-shaped issue per #2142: one root cause, five links, each surfaced by the previous review. One finding refined five times, not five failures.
- #2516 is one more instance of the eager `__dict__` bulk-dump class that the ROADMAP v1.2.0-2 row (ADR-027, resolve like Django at one lazy sink) retires structurally.

IDEA: a grep-based structural test that every `.iter()` on a `Bound<PyDict>` in `crates/` is preceded by a `.collect()` snapshot or routed through one helper, making the completeness claim mechanical (#1125).
IDEA: collapse the per-site snapshots into one `snapshot_dict()` helper and pin it as the only `PyDict` iterator (#1646).
IDEA: a CLAUDE.md build note for the `uv.lock` relock hazard and the post-push `rev-parse` check.
IDEA: when a review finds a second instance of a pattern, require a sink-first sweep before the next round (#2142 step 1 at round 2, not round 5).

RETRO_COMPLETE

Issue #2523, opened 2026-09-01
tech-debt: test.yml has no permissions: block and two jobs still use the archived actions-rs/toolchain


johnrtipton:
## Two pre-existing findings from the #2517 Stage-8 security scan of `.github/workflows/test.yml`

1. **No `permissions:` block at any level.** Every job runs with the repository-default `GITHUB_TOKEN` scope. A top-level `permissions: contents: read` (with per-job widening only where a job writes — comments, artifacts do not need it) is the least-privilege default and is what `pr-workflow.yml`-style jobs that post comments should opt out of explicitly.
2. **Two jobs still use the archived `actions-rs/toolchain@v1`** (the `rust` job and one other; grep `actions-rs`). Newer jobs, including #2517's, use `dtolnay/rust-toolchain@stable`. Converge on one.

Both are hygiene, not exploits: no job in the file uses `secrets.*`, and every third-party action is pinned to a major tag. Filed as `IMPROVEMENT:` lines by the #2517 review; tech-debt.

Issue #2522, opened 2026-09-01
Promote the django-template-suite CI job to a ratchet (blocking compare against the committed baseline)


johnrtipton:
## What

Promote the `django-template-suite` CI job (added by #2517, non-gating per the #1534 rule) into an enforcing **ratchet**: a PR may not lower the Django `template_tests` pass percentage stored in `scripts/django-template-suite-baseline.json`.

## Preconditions (all must hold first)

1. The job has been green on the GitHub runner at least once on `main` (the #1534 rule: a job the dev machine cannot fully mirror stays `continue-on-error: true` until then).
2. Two consecutive `main` runs reproduce the committed baseline exactly (the runner is deterministic locally; confirm on the runner, where wall time and the 3-second `test_middleware_caching` sleep differ).

## Steps (the #1713 trace — a check is only a gate if a failure actually fails the merge)

1. Remove `continue-on-error: true` from the job.
2. Make the `compare` step blocking: `run-django-template-suite.py compare --json <run> --baseline scripts/django-template-suite-baseline.json` returns 1 when either the engine-subset or the whole-label percentage dropped, or when the engine bucket is empty while the baseline's is not (an adapter that silently failed to install must not pass).
3. Add the job to `test-summary.needs` AND to the `&&` success chain in `.github/workflows/test.yml`; verify there is no `continue-on-error` on the job, any step, or the aggregate.
4. Write down the rule for moving the baseline: a PR that improves the number regenerates the file with `--write-baseline` and commits it; a PR must never lower it by hand.

Same path `demo-checks` took (PR #1730). Filed from the #2517 plan §6.

johnrtipton:
Two conditions from the PR #2525 Code Review that the promotion must handle before `compare` becomes blocking:

1. **A Django tag mismatch currently returns 0** (`scripts/lib/django_template_suite/report.py`, the `compare` tag check). That is right while the step is informational, but promoted as-is a Django version bump would pass the ratchet with no comparison at all. The promotion should either fail on mismatch, or require the PR that bumps Django to regenerate the baseline in the same change (which also re-baselines the number against the new tag's test set).
2. **A crash is recorded as `touched=True` by construction** (`scripts/run-django-template-suite.py`, `_crash_record`), so a native crash in a test that never reached the engine would still count against the headline bucket. Acceptable for `template_tests`, where every crasher today does reach it, but worth stating in the promotion's docs so a future non-engine crash is not misread as a regression.

johnrtipton:
[...]

Issue #2521, opened 2026-09-01
Recursive `{% extends %}` across loaders (a base.html extending the next base.html) hits the depth limit instead of resolving like Django


johnrtipton:
## Summary

Django supports recursive `{% extends %}` across loaders: a `base.html` in an earlier template directory may `{% extends "base.html" %}` and Django resolves that to the *next* `base.html` in loader order (the `ExtendsBehaviorTests` in `template_tests/test_extends.py`, and the documented pattern for overriding an app's template while inheriting from it). djust's inheritance walk resolves by name in `template_dirs` order from the beginning every time, finds the child itself, and stops at the depth limit.

## Evidence (Django 5.2.16 `template_tests` through `DjustTemplateBackend`, via the #2517 runner, 2026-09-01)

Four `syntax_tests.test_extends.ExtendsBehaviorTests` cases (`test_recursive_multiple_loaders`, `test_extend_recursive`, `test_extend_missing`-adjacent shapes — see the runner's parsed output) error with:

```
Template inheritance depth limit (10)
```

## Where

The inheritance resolution in `crates/djust_templates/src/inheritance.rs` (`FilesystemTemplateLoader`) and the Python-side walk in `python/djust/template/rendering.py:288`. Django's rule lives in `django/template/loader_tags.py` (`ExtendsNode.find_template` uses `context.render_context[self.context_key]`, a history of origins already visited, and skips them — `django/template/loaders/base.py` `get_template(template_name, skip=...)`).

## Fix shape

Carry the list of already-used origins through the extends chain and pass it as `skip` to the loader, exactly Django's mechanism; the depth limit stays as the backstop. Scored by the four cases above. A v1.2.0-3 item (plain-Django completeness); found by the #2517 conformance run.

johnrtipton:
Verified fixed on `main` @ fea90afd (after #2665 merged) — closing.

Two-loader recursive `{% extends "base.html" %}` (child → dirs[0]/base.html → dirs[1]/base.html) with `block.super` at each level renders identically on djust and Django — no depth-limit error.

Issue #2520, opened 2026-09-01
`{% firstof ... as var %}` and other `as var` tags never write back into the caller's Context


johnrtipton:
## Summary

On Django, `{% firstof a b c as myvar %}` (and every other `as var` tag) writes the variable into the caller's `Context`, so Python code that rendered the template can read `context["myvar"]` afterwards. On djust the `Context` is flattened into a dict copy before the Rust engine runs and is never written back, so the variable exists inside the template but the caller's `Context` never sees it.

## Evidence (Django 5.2.16 `template_tests` through `DjustTemplateBackend`, via the #2517 runner, 2026-09-01)

- `syntax_tests.test_firstof.FirstOfTagTests.test_firstof15` — ERROR `KeyError: 'myvar'`
- `syntax_tests.test_firstof.FirstOfTagTests.test_all_false_arguments_asvar` — ERROR `KeyError: 'myvar'`

Both render, then read the variable back from the `Context` object in Python. The same class applies to `{% cycle ... as x %}`, `{% now "Y" as x %}`, `{% url ... as x %}`, `{% widthratio ... as x %}`, `{% regroup ... as x %}`, `{% with %}` and any custom `simple_tag(takes_context=True)` that assigns, whenever the caller reads the result from the `Context` rather than from the rendered output.

## Where

`DjustTemplate.render` (`python/djust/template/rendering.py:706`, flatten at `:732`) hands the Rust engine a dict and returns a string; nothing flows back. Django's `Template.render` mutates the `Context` it was given (`django/template/base.py`, `Context.push`/`__setitem__` inside the node's `render`).

## Fix shape, and a decision

Django's own semantics are that a top-level `as var` assignment lands in the outermost context dict the caller passed. Two options: (a) the Rust render returns the set of top-level assignments alongside the HTML and `DjustTemplate.render` writes them into the caller's `Context` when one was passed (a dict caller gets nothing, as today); (b) document as unsupported and refuse `as var` at the top level when a `Context` was passed. (a) is parity; (b) is honest. Either is scored by the two cases above. A v1.2.0-3 item; found by the #2517 conformance run.

johnrtipton:
Verified fixed on `main` @ fea90afd (after #2665 merged) — closing.

[...]

PR #2525, opened 2026-09-02
feat(templates): run Django's own template_tests against DjustTemplateBackend as the conformance scoreboard (#2517)

### LINKED ISSUE #2517: v1.2.0-1: run Django's own template test suite against DjustTemplateBackend as the conformance scoreboard
## What

Run Django's own template test suite (`tests/template_tests` in the `django/django` repository) against `DjustTemplateBackend`, print a pass percentage, and make that number the arc's scoreboard. First row of **v1.2.0-1 — Measure** on ROADMAP.md (the v1.2.0 template-engine conformance arc, PR #2515).

## Why

djust's Django-parity is measured by a corpus we invented (`scripts/filter-parity-differential.py`, 4,259 lines), which has found five of its own blind spots one at a time (#2345). Django ships roughly 1,600 template tests of its own — tags, filters, loaders, context, autoescape, inheritance, custom libraries — and we have never run them. Every later row in the arc (ADR-027 resolution, `{% load %}` library import, the six missing tags, i18n) is accepted or rejected by the number this produces.

## Deliverables

1. **`scripts/run-django-template-suite.py`** — resolves the installed `django.__version__` (pinned by `uv.lock`), checks out `django/django` at the matching tag into a gitignored cache, routes `tests/template_tests` through `DjustTemplateBackend`, runs it with Django's own runner, and prints one line per test plus a summary:
   ```
   Django test suite passing: NN.NN%
   NNN ERROR / NN FAIL / NNN OK
   ```
   Distinguish ERROR (crash / unsupported) from FAIL (wrong output) — they are different work.
   *The seam to route through is `tests/template_tests/utils.py` (`setup()` builds `django.template.Engine` instances per settings permutation and the tests call `self.engine.from_string(...)`), not the `TEMPLATES` setting — most of the suite never touches a backend. Planning decides between a patch file applied to the checkout and a monkeypatch shim; either must be re-applicable across Django tags.*
2. **A stored baseline** (tracked, e.g. `scripts/django-template-suite-baseline.json` with the percentage and the counts) written on first run and updated only by PRs that improve it.
3. **`make django-template-suite`** target.
4. **CI job** in `.github/workflows/test.yml`, `continue-on-error: true` until it has been green on the runner once (the #1534 rule), then promoted to a **ratchet**: a PR's percentage may not drop below the stored baseline. The promotion is a follow-up commit or PR, not this one.
5. **Docs**: the baseline number in `docs/TEMPLATE_BACKEND.md` next to the supported/unsupported lists, and a paragraph on how to run it locally.

## Acceptance

- `make django-template-suite` runs locally, prints the percentage, exits 0 (the ratchet is not enforced yet).
- The CI job exists, non-gating, and has run at least once on the PR.
- The baseline file is committed with the real first number, and the number appears in the PR description.
- Tests for the tool itself (#1459 empirical canary for a tooling PR): a synthetic case that djust renders differently from Django must show up as a FAIL in the runner's parsed output; unit tests for the summary parser (percentage arithmetic, ERROR vs FAIL counting, skipped tests excluded from the denominator); a ratchet-comparison unit test even though the ratchet is not yet wired into CI.

## Out of scope

Fixing anything the suite finds. Failures become issues for v1.2.0-2 / v1.2.0-3; this PR only measures.

johnrtipton:
Closes #2517. First row of **v1.2.0-1 — Measure** on the roadmap (PR #2515).

## What

`make django-template-suite` runs Django's own `tests/template_tests` (the `django/django` checkout at the tag matching the installed Django, 5.2.16) against `DjustTemplateBackend`, prints one line per test, and stores the result in a tracked baseline. The number is the arc's scoreboard: every later row (ADR-027 resolution, `{% load %}` library import, the missing tags, i18n) is accepted or rejected by whether it moves.

**Baseline committed** (`scripts/django-template-suite-baseline.json`):

| bucket | passing | OK / FAIL / ERROR | denominator |
|---|---|---|---|
| tests that reach the engine (**headline**) | **43.55%** | 456 / 227 / 364 | 1047 |
| whole label | 59.41% | 865 / 227 / 364 | 1456 (14 skipped) |

The headline is the engine subset because 409 of Django's template tests never construct an engine and measure Django against itself; a claim of 59% would be padded and every later gain would read smaller than it is. Both numbers are stored. `--gate-off` (adapter not installed) gives 100.00% with 0 tests touching the engine, so the harness is non-vacuous.

## How

- **Seam**: an in-process subclass of the real `django.template.engine.Engine` overriding only `from_string` and `get_template`, installed after `django.template.backends.django` has bound the real class, so the `TEMPLATES` backend, `Engine.get_default()` and Django's own runner stay real. Covers the 913 `@setup` tests and the 68 direct `Engine(...)` constructions. Nothing in Django's checkout is edited, so it re-applies to any tag.
- **Per-test crash resume**: seven `template_tests` cases segfault the interpreter (a reference cycle through the context — the #2516 class, reproductions attached there). The child writes flushed JSON lines; on a signal exit the in-flight test is recorded `ERROR: process crashed (signal 11)` and the run resumes. Whole run: ~11 s.
- **ERROR vs FAIL** are Django's own unittest classification, untouched: ERROR = unsupported tag / crash (`autoescape` 79, `blocktrans` 34, `ifchanged` 24, `trans` 19, `cache` 15, `querystring` 14), FAIL = wrong output (`TemplateSyntaxError` not raised, 111; `string_if_invalid` ignored, 20).
- **CI**: a `django-template-suite` job, `continue-on-error: true`, not in `test-summary`'s AND-chain (the #1534 rule). Promotion to a blocking ratchet is #2522.
- **Hardening from review**: `--django-tag` is validated before any filesystem access (a `..` tag previously deleted a directory outside the cache before the clone failed); labels are refused if they start with `-` and passed after `--` (`--label=--gate-off` previously wrote a 0% baseline that would have neutered `compare`); the checkout root is normalised to `<django-src>` in outputs; the CI checkout uses `persist-credentials: false`.

## Findings the first run filed (not fixed here — this PR only measures)

#2518 `DjustTemplateBackend` silently drops every `OPTIONS` key except `context_processors` · #2519 `<!--dj-if-->` markers leak into plain-backend output (the feature-gate row) · #2520 `{% firstof … as var %}` never writes back to the caller's `Context` · #2521 recursive `{% extends %}` across loaders hits the depth limit · #2516 gained the seven reproducible segfault cases · #2523 two pre-existing workflow hygiene items.

## Files × tests

| area | files | tests (`python/tests/test_django_template_suite_2517.py`) |
|---|---|---|
| summary arithmetic, buckets, ratchet | `scripts/lib/django_template_suite/report.py` | `TestSummaryArithmetic`, `TestRatchetCompare` |
| JSONL recorder, normalisation | `recorder.py` | `TestRecordingResult` |
| engine adapter | `adapter.py` | `TestAdapterInSubprocess` (asserts the returned template is a `DjustTemplate`, and that a bad loader raises at `get_template` as on Django) |
| runner, crash loop, argv validation | `scripts/run-django-template-suite.py`, `child.py`, `settings.py` | `TestEmpiricalCanary` (a synthetic djust-vs-Django divergence shows up as FAIL/ERROR; gate-off = 100%), `TestCrashIsolation` (a real SIGSEGV, resume, timeout, crash-between-tests), `TestArgvHardening` |
| baseline + docs | `scripts/django-template-suite-baseline.json`, `docs/TEMPLATE_BACKEND.md` | `TestDocClaimMatchesBaseline` (the `<!-- django-suite-claim -->` line equals the baseline) |
| real checkout | `Makefile`, `.gitignore`, `.github/workflows/test.yml` | `TestAgainstRealDjangoCheckout` (runs on the CI runner after the clone) |

## Verification

- Full suite: 20,408 passed / 0 failed / 438 skipped / 15 xfailed; `make test-rust` 1,398 passed (no Rust changes).
- Runner byte-identical across three runs; `compare` against the committed baseline: zero delta, exit 0.
- Gate-off self-tests: 4 mechanisms at implementation + 12 at the fix pass, each mutation red for exactly its test.
- Security probes: `--upload-pack=…` as a tag is consumed as a branch name (no shell anywhere, list argv only); temp dirs are 0700 `mkdtemp` cleaned in `finally`; the adapter package is imported nowhere under `python/` or `crates/`.


johnrtipton (commented):
## Stage 11 Code Review — PR #2525 (#2517)

**Verdict: APPROVE.** Branch 0 behind `origin/main`; read-only review + independent canary.

**Empirical canary (#1459), my own package (locmem+`get_template`, not the test's):**
```
OK    rev_canary.RevCanary.test_i_variable_same_on_both
FAIL  rev_canary.RevCanary.test_ii_syntax_error_must_raise | AssertionError: TemplateSyntaxError not raised
ERROR rev_canary.RevCanary.test_iii_autoescape_off | Exception: Error rendering template: ... Unsupported template tag '{% autoescape off %}'
Django test suite passing: 33.33%   (3 tests exercised the djust engine) — touched [True, True, True]
--gate-off: Django test suite passing: 100.00% ... 0 tests exercised the djust engine — touched [False, False, False]
```

**Structural checks all hold:** adapter overrides only `from_string`/`get_template` (`adapter.py:112,119`); subprocess probe after `install()`: `backends.django.Engine is RealEngine` True, `Template("...")` and `Engine.get_default()` stay Django. No `except` reclassifies (`adapter.py:134` only). SKIP/XFAIL excluded (`report.py:147-152`). CI job `continue-on-error: true` (`test.yml:716`), `persist-credentials: false` (`:722`), absent from `test-summary.needs` (`:771`) and the `&&` chain (`:793-800`). `validate_tag` at `run-django-template-suite.py:345` precedes any FS access; `--` in both argv builds (`:197`, `child.py:66`). Baseline arithmetic and the 68-`Engine(`/913-`@setup` claims recount exactly; `compare` vs local run: zero delta, rc 0.

**Edge cases run:** tag `5.2/16`, `../x` (refused, cache never created); `--label=--gate-off` (refused); `setUpClass` segfault (stops with "after 2 finished"); empty engine bucket in `compare` → 1; a dict passed to `render` — `DjustTemplate` accepts it where Django's engine-level template raises: one-directionally permissive, can only yield a false OK, no Django test depends on it.

🟡 `report.py:245-250`: tag mismatch returns 0 — fine while informational, but #2522 must not promote that as-is or a Django bump silently passes the ratchet.
🟡 `_crash_record` `touched=True` by construction (`:314`) — a non-engine native crash would land in the headline; acceptable for `template_tests`, worth a comment.

**Tests:** `test_django_template_suite_2517.py` 67 passed (real-checkout cases ran). Full suite: 20,433 passed, 1 failed, 437 skipped, 15 xfailed — the failure is `test_forms.py::test_the_narrowing_survives_an_unrelated_warning` (#2420 warnings-count flake, file untouched, passes in isolation).

johnrtipton:
## Retrospective — PR #2525 (#2517)

Quality: 5/5 — the scoreboard is real (43.55% engine subset, 456/227/364 of 1047; 59.41% whole label; 7 crashers isolated; byte-identical across three runs), the CI job was green on its first runner run in 2m59s, and every defect the measurement surfaced became an issue instead of scope creep. The only cost was process, not product.

### What went well
- Stage 4 spiked the seam before designing it: an in-process `Engine` subclass (option B of three) was run against the real suite, which found the seven segfaults, the single-process 58% loss at test 626, and two backend defects on the real path (#2518 OPTIONS keys dropped; the `<!--dj-if-->` leak, now #2519). The plan quoted measured numbers, not estimates.
- Canary and gate-off discipline held end to end: the runner ships its own #1459 canary (OK/FAIL/ERROR = 33.33%, gate-off 100% with 0 touched), the 7+8 fix pass proved 12/12 mutations red, and the Code Review re-ran an independent canary with its own package rather than the test's.
- The combined 7+8 pass closed 8 🟡 that mattered: tag traversal deleting outside the cache, `--label=--gate-off` writing a 0% baseline, checkout-path normalisation, `persist-credentials: false`, the adapter type assertion, the crash-between-tests message, the `template_dirs` laziness probe.
- #1079 applied seven times: #2518, #2520, #2521, the #2516 crasher list, #2522, #2523, #2524 were filed, not fixed.

### Lessons learned
- **Settle the push transport BEFORE the 9-minute hook.** Four pre-push runs on one commit, each failing differently: GitHub closing the SSH session mid-hook (twice), a bare-URL push scoping pre-commit to all history and tripping ruff-format on 16 stale main files, and a `gh` token without the `workflow` scope for HTTPS. A 10-second `git ls-remote` over the intended transport, plus a scope check when `.github/workflows` changed, would have cost 40 seconds instead of ~36 minutes. Filed as #2526.
- **Push through the remote NAME, never a URL.** `git -c remote.origin.url=https://…` APPENDS a second URL (git pushes to both, SSH first; confirmed with GIT_TRACE). The working form is `-c url.https://github.com/.insteadOf=git@github.com:`. Also: the first keepalive probe fed stdin from /dev/zero and measured nothing.
- 16 unformatted test files on main (#2524) were invisible until a non-standard push widened pre-commit's scope. `pre-commit run --all-files` belongs in the release checklist.

### Relation to prior work
- #2514's retro asked for sink-first sweeps and structural pins; this PR's canary, the `Engine is RealEngine` subprocess probe, and the untouched-set 100% invariant are the same discipline, applied at Stage 4 rather than at re-review.
- The v1.2.0 arc's headline is the engine-exercising subset by decision: 409 label tests never call `Engine`, so a whole-label number would pad every later gain.
- #2516 (crash on a `DjustTemplate` or type object in context) is the eager-conversion class the v1.2.0-2 row (ADR-027, one lazy sink) retires structurally; this PR isolates it, the next row removes it.

IDEA: a pre-push transport probe in `scripts/pre-push-pytest.sh` that refuses to start the suite if `git ls-remote` over the push transport fails, or the token lacks `workflow` when `.github/workflows` changed.
IDEA: a conditional pytest selector so a docs/scripts-only diff does not run 20k tests (#2526 item 3).
IDEA: `pre-commit run --all-files` in the release checklist so stale formatting never accumulates on main (#2524).
IDEA: the pipeline's own push helper takes a remote name only; reject a URL argument.

RETRO_COMPLETE

Issue #2531, opened 2026-09-02
`{{ block.super }}` drops the parent block on the backend path (inheritance20-23/39), and `{% include %}` of an extending template raises


johnrtipton:
## What

`{{ block.super }}` on the `DjustTemplateBackend` path drops the parent block's content: for a child `{% extends %}` whose block renders `{{ block.super }}`, djust gives `<html></html>` where Django gives `<html><p>b</p></html>`. Django's `template_tests` already count it — `syntax_tests.test_extends` `inheritance20`, `21`, `22`, `23` and `39` are FAILs in the #2517 scoreboard.

Also observed in the same review, possibly the same inheritance walk: `{% include %}` of a template that itself `{% extends %}` **raises** on the Rust include path instead of rendering the included template's resolved output as Django does.

## Where

`crates/djust_templates/src/inheritance.rs` (block resolution) and the include path in `renderer.rs`; the Python-side walk at `python/djust/template/rendering.py:288`. Django's rule: `BlockNode.render` pushes a `BlockContext` and `block.super` renders the next block up the chain (`django/template/loader_tags.py`).

## Scored by

The five `inheritance*` cases above turning OK in `make django-template-suite`, plus one new case for include-of-extends. A v1.2.0-3 (plain-Django completeness) item, sibling of #2521 (recursive extends across loaders). Found by the #2519 Stage-7 review.

johnrtipton:
One more scored case, found while draining #2580 — the negative side of the same gap.

`{{ block.super }}` used OUTSIDE a child template (no active `{% extends %}` chain) must raise `TemplateSyntaxError`, not render empty. Django's `BlockNode.render` sets `context["block"] = self` when there's no active `block_context`; `BlockNode.super` is a property that checks `hasattr(self, "context")` (only set on the child-template render path) and raises `"'BlockNode' object has no attribute 'context'. Did you use {{ block.super }} in a base template?"` when absent.

Confirmed today: `crates/djust_templates/src/inheritance.rs`'s `parent_blocks: HashMap<String, Vec<Node>>` field (commented `// Parent block content for {{ block.super }}`) is set once and never read anywhere else in the crate or in `python/`. `{{ block.super }}` isn't a resolvable context variable at all right now — it resolves as an undefined name and renders empty in EVERY case, both the positive (child template, should render the parent's content — this issue's original report) and the negative (base template, should raise — `test_exception05` in `syntax_tests/test_exceptions.py`).

[...]

Issue #2530, opened 2026-09-02
LiveComponent.template_name renders through the project backend, so its {% if %} blocks lose dj-if boundaries inside the parent VDOM after #2519


johnrtipton:
## What

A `LiveComponent` with `template_name` renders through `render_to_string` → the project's `TEMPLATES` backend (`python/djust/components/base.py`, the `template_name` branch of `render()`). On a project whose backend is `DjustTemplateBackend` (the `djust new` scaffold default) that HTML lands **inside the parent LiveView's VDOM**. Before #2519 it carried `<!--dj-if id=…-->` boundary markers only because the backend path leaked them; after #2519 the plain path emits none, so `{% if %}` blocks inside such components fall back to positional matching in the differ.

This is the state inline-`template` components (`render_template`, which always stripped) and every `DjangoTemplates`-backend project have always been in, so it is not a regression the suite can see — but it is a LiveView-page byte change, and it means component `{% if %}` blocks are second-class for VDOM patching by accident of which render entry the component happens to hit.

## Decision needed (a design question, not a patch)

Component HTML that is going to be consumed by the parent's VDOM should be rendered through an entry that emits markers (the LiveView entry, `emit_dj_if_markers = true`), regardless of the project's `TEMPLATES` backend. Options: (a) `LiveComponent.render()` with `template_name` resolves the template source itself and renders through `_rust` with markers on; (b) a render-time switch the parent LiveView sets while rendering children; (c) leave as is and document. Pick with a test that a `{% if %}` toggle inside a `template_name` component produces a boundary-scoped patch rather than a positional one (the #1678 differential harness shape).

Found by the #2519 Stage-7 review (parallel-path axis: direct callers of the plain entries were audited, indirect callers through the backend were not). Related: #1646.

Issue #2529, opened 2026-09-02
security-hardening: `__djust_if_loop_path` context key is interpolated unescaped into the dj-if marker id on the LiveView path


johnrtipton:
## Summary

On the LiveView render path the `<!--dj-if id="if-<hash>-N…"-->` marker id gets a loop-path suffix read from the **context key** `__djust_if_loop_path` and interpolated **unescaped** into the HTML comment (`crates/djust_templates/src/renderer.rs`, the loop-path branch of the `Node::If` marker emission; #1832 introduced it). Everything else in the id is template-source-derived (`template_hash_hex(source)` + parser ordinal) and context-independent; this suffix is the one exception.

## Probe (verified 2026-09-02 on `feat/2519-liveview-feature-gate` @ cd0463e7; byte-identical code on `main`)

```python
RustLiveView.render(..., {"foo": 1, "__djust_if_loop_path": '"--><script>alert(1)</script><!--'})
# <div><!--dj-if id="if-4369cbd0-0"--><script>alert(1)</script><!--"--><b>x</b><!--/dj-if--></div>
```

A live script tag. The plain `DjustTemplateBackend` path is unaffected (no markers since #2519).

## Reachability — why this is hardening, not an exploit

Every client→state write goes through `safe_setattr(..., allow_private=False)` (`websocket.py`, `live_view.py`, `mixins/request.py`, `sticky.py`), which rejects leading-underscore names (`attribute_guard.py`), so a client cannot plant the key. It requires a developer to return that reserved dunder key **with attacker-controlled content** from `get_context_data()`. Unlikely, but the framework should not have a context key whose value bypasses escaping into markup at all.

## Fix shape

Either (a) carry the loop path on `djust_core::Context` as a field, the way #2519 carries `emit_dj_if_markers` — the value never enters the user namespace; or (b) validate it at the emit site (`^[0-9-]*$`, which is all the framework ever writes) and refuse anything else; ideally both. Pin with a test that the probe above renders the payload entity-escaped or refuses.

Found by the #2519 Stage-8 security review. Related: user values containing `-->` are entity-escaped on both paths (verified), so this key is the only marker-forging vector; a developer-`|safe`d attacker string can still confuse boundary pairing, which is strictly weaker than the XSS `|safe` already grants and is a docs note, not a bug.

johnrtipton:
[...]

Issue #2528, opened 2026-09-02
`{% if x|default_if_none:y %}` with undefined `x` is false in djust, true in Django (`test_if_tag_badarg03`)


johnrtipton:
## What

Django's `{% if %}` resolves its operands with `ignore_failures=True` (`TemplateLiteral.eval` → `FilterExpression.resolve(context, ignore_failures=True)`), so an undefined variable becomes `None` *before* the filter chain runs. `{% if x|default_if_none:y %}yes{% endif %}` with `{"y": 1}` therefore evaluates `default_if_none(None, 1)` → `1` → truthy → `yes`. djust evaluates the condition false and renders `''`. Note the `{{ }}` path is NOT affected — there Django substitutes `string_if_invalid` (`''`) before filtering, and both engines render `''` — so this is specific to how the `{% if %}` condition evaluator resolves an undefined variable with a filter chain.

```python
# both engines built directly (DjangoTemplates / DjustTemplateBackend), ctx {"y": 1}
{{ x|default_if_none:y }}                     django=''       djust=''
{% if x|default_if_none:y %}yes{% endif %}    django='yes'    djust=''
{{ x|default:y }}                             django='1'      djust='1'
```

Surfaced by the #2519 fix (feat/2519-liveview-feature-gate): `template_tests.syntax_tests.test_if.IfTagTests.test_if_tag_badarg03` was one of six suite FAILs of the shape `'<!--dj-if-->' != …`; the other five flipped to OK once the marker leak was fixed, this one now reads `'' != 'yes'` — a separate, pre-existing parity bug, kept out of that PR's scope (#1079).

## Also observed (same PR, same file, not fixed there)

`{% include … only %}` builds a fresh `Context` (`crates/djust_templates/src/renderer.rs`, the `only` arm) and did not carry the parent's render-time switches. #2519 makes it copy `emit_dj_if_markers`; it still does NOT copy `auto_call` (ADR-024), so an `only` include renders with auto-call ON regardless of `LIVEVIEW_CONFIG['template_auto_call']`. Pre-existing; fix alongside or separately.

johnrtipton:
Verified fixed on `main` @ fea90afd (after #2665 merged) — closing.

Ran the issue's exact case with ctx `{"y": 1}`: `{% if x|default_if_none:y %}yes{% endif %}` → djust `'yes'`, Django `'yes'`.

PR #2534, opened 2026-09-02
feat(templates): liveview cargo feature gate + stop emitting dj-if markers on the plain-backend path (#2519)

### LINKED ISSUE #2519: v1.2.0-1: feature-gate the engine/LiveView boundary — `<!--dj-if-->` markers leak into plain-backend output
## What

Second row of **v1.2.0-1 — Measure** on ROADMAP.md: add a `liveview` cargo feature on `djust_templates` (default on) that gates the LiveView-only parts of the engine — the `djust_vdom::VNode` cache in `loop_cache.rs:109`, the `<!--dj-if-->` boundary markers, and the `djust_components` path in `renderer.rs:7` — plus a CI step `cargo check -p djust_templates --no-default-features`. This is not the package split; it is the measurement of what a split would have to carry, a tripwire against LiveView concerns leaking into the plain engine, and the first criterion of the separate-wheel decision gate on the roadmap.

## The leak that makes it concrete (verified 2026-09-01, main e30a7bb6)

```python
engines["djust"].from_string("{% if foo %}foo{% elif bar %}bar{% endif %}").render({})
# '<!--dj-if-->'      Django: ''
```

A LiveView VDOM boundary marker is emitted on the plain `DjustTemplateBackend` path, where no VDOM exists to consume it. `crates/djust_templates/src/lib.rs:318` says standalone rendering strips placeholder and boundary markers; this path does not go through that strip. Django's `template_tests` run through the backend (#2517) counts 6 FAILs of exactly this shape, and a plain-Django page gets an HTML comment in its output for every false `{% if %}`.

## Deliverables

1. `[features] default = ["liveview"]` on `crates/djust_templates/Cargo.toml`; `#[cfg(feature = "liveview")]` on the loop cache's VNode storage, the marker emission, and the component renderer path (the sidecar/registry that LiveView needs but the backend does not is a judgment call — write it down).
2. The plain-backend path renders with markers OFF regardless of the feature (the feature gates compilation; the render entry chooses behaviour) — fixes the leak above, scored by those 6 `template_tests` cases.
3. `cargo check -p djust_templates --no-default-features` in `.github/workflows/test.yml` (the `rust` job), plus `cargo test -p djust_templates --no-default-features` if the test set compiles.
4. A one-paragraph note in `docs/TEMPLATE_BACKEND.md` on what the engine is without the feature.

## Non-goals

Publishing a separate wheel. Changing the LiveView path's output.

johnrtipton:
Closes #2519. Second row of **v1.2.0-1 — Measure** on the roadmap.

## Two halves, one root cause

**The leak.** On the plain `DjustTemplateBackend` path, `{% if foo %}foo{% elif bar %}bar{% endif %}` rendered `<!--dj-if-->` where Django renders `''`. Root cause is parallel-path drift (#1646) between the two plain Rust entry points in `crates/djust_live/src/lib.rs`: the renderer emitted VDOM boundary markers unconditionally, and the only "plain" mechanism was a post-render regex strip that `render_template` applied and `render_template_with_dirs` — the entry the backend binds — never did. Every backend page leaked both marker forms, including through `{% include %}`.

**The fix, structural rather than a second strip.** A render-time `emit_dj_if_markers` switch on `djust_core::Context`, default `true` so the LiveView path and every existing test keep their bytes, set `false` by both plain entries. The strip is deleted so two mechanisms cannot shadow each other (re-applying it to the plain outputs changes 0 of 38; it cost 26–30% of render time on a 40-conditional page). The switch lives on the `Context`, not the parsed template, because `TEMPLATE_CACHE` is shared by both paths; and `{% include … only %}` builds a fresh `Context` mid-render, which now copies the switch (the one place that would have leaked again).

**The feature gate.** `crates/djust_templates` gains `[features] default = ["liveview"]`; `djust_vdom` and `djust_components` become optional. Gated: the `VNode` parse cache in `loop_cache.rs`, the `djust_components` renderer path (with a `cfg(not)` twin of the `RustComponent` arm that returns a `TemplateError`), and marker emission. Deliberately **not** gated, with reasons in the plan: the PyO3-backed tag and filter registries (plain Django needs `{% url %}` and custom filters, so the engine without the feature is still a PyO3 crate), the parser's marker-id assignment, the string-only loop-cache manifest, and the sidecar (which lives in `djust_core`). `djust_live` needs no change. CI: `cargo check` and `cargo test -p djust_templates --no-default-features`, plus a `cargo tree` grep that fails if either crate stops being optional; `make check-no-default-features` chained from `test-rust`.

## Scoreboard

| | before | after |
|---|---|---|
| engine subset (headline) | 43.55% (456 / 227 / 364 of 1047) | **44.03%** (461 / 222 / 364) |
| whole label | 59.41% | 59.75% |
| `dj-if` bytes in the run | 6 FAILs | 0 |

Five of the six `template_tests` cases flipped OK (`test_inheritance32`, `test_inheritance35`, `test_if_tag06`, `test_if_tag_badarg01`, `_badarg02`). The sixth, `test_if_tag_badarg03`, now reads `'' != 'yes'` — a separate pre-existing bug in how `{% if %}` resolves an undefined variable through a filter chain, filed as #2528 (Django resolves `{% if %}` operands with `ignore_failures=True`). Baseline regenerated with `--write-baseline`; `compare` against the old baseline reports the improvement; `--gate-off` still 100%.

## Tests

- Rust `crates/djust_templates/tests/test_dj_if_markers_off_2519.rs` (29 cases; 24 under `--no-default-features`): switch off → `""` for a false `{% if %}`, no id-carrying pair, `{% include %}` / `{% include … only %}` inside both branches emit nothing; switch on → byte-identical to today's expected strings.
- Python `python/tests/test_dj_if_marker_plain_backend_2519.py` (30): the issue's repro renders `''`; 19 shapes render byte-equal to Django's real `DjangoTemplates` backend in-process; a LiveView-path control still emits the marker (so a global "never emit" bug goes red); backend output equals LiveView bytes minus markers.
- `python/tests/test_template_if_markers.py`: its `render_raw` helper rendered through the plain entry while claiming to be the VDOM path — repointed.
- Two pre-existing tests pinned the old state and went red on purpose: `test_forloop_parity_2402.py`'s `TestTheDjIfMarkerIsOrthogonal` was a tripwire written to fail the day the leak was fixed (its message: "the marker artifact has gone — drop the stripping"); it now pins the absence of the marker and the file's two modulo-marker cells compare byte-for-byte. `test_makefile_pyo3_embeddable_2082.py`'s #1125 count pin moves 6 → 9 for the new target's three `PYO3_PYTHON=$(EMBEDDABLE_PYTHON)` lines.
- Gate-offs: switch removed from either plain entry, `only`-include copy removed, `optional = true` dropped, `cfg(not)` helper/arm removed — each red for exactly its test or CI step.

## Findings filed, not fixed here

#2528 (`{% if %}` operand resolution; `only`-include also drops `auto_call`) · #2529 (pre-existing: `__djust_if_loop_path` interpolated unescaped into the LiveView marker id; client cannot plant it, hardening) · #2530 (a `LiveComponent` with `template_name` renders through the project backend, so on a djust-backend project its `{% if %}` blocks now reach the parent VDOM without boundaries — the state inline-template components always had; a design decision) · #2531 (pre-existing: `{{ block.super }}` drops the parent block on the backend path; `{% include %}` of an extending template raises).

**Behaviour change to note.** LiveView pages that embed a `template_name` `LiveComponent` on a `DjustTemplateBackend` project lose the boundary markers that used to leak into that component's HTML; `{% if %}` blocks inside such components are matched positionally by the differ, as they already were for inline-template components. Tracked in #2530.


johnrtipton (commented):
## Stage 11 review — PR #2534 (#2519) — **APPROVE**

Read-only review at 0e4f942f, 0 behind origin/main.

**Empirical canary (own, in-process real Django).** Issue repro + 6 shapes (nested for/if, include, include-only, elif middle-true, if-in-attribute, `|safe` value carrying literal markers) through the backend, `_rust.render_template`, `_rust.render_template_with_dirs`: all byte-equal to Django, 0 emitted `dj-if` bytes (issue repro: Django `''`, backend `''`, LiveView `'<!--dj-if-->'`). The `|safe` row keeps its 3 user-authored markers on every plain entry — the old strip would have deleted them, so this is a parity gain. `RustLiveView.render` still emits (6 pair markers on the nested shape).

**Cross-path (#1447).** LiveView with loop cache on + a registered tag handler doing a plain `render_template_with_dirs` of the same include mid-render: plain part == Django, no markers; LiveView part keeps 6. Counters confirm the invariant the `content_hash` comment relies on (`loop_cache.rs:461`): if-body loop 0 hits/0 misses; control loop without `{% if %}` 3/3.

**Parallel paths (#1646).** Every `Context` construction in `djust_live`/`djust_templates` enumerated; the only markers-on standalone entry is `Template::py_render` (`djust_templates/src/lib.rs:283`), unshipped (`_rust` exposes no `Template`), decided in plan §4.

**Feature gate.** `cargo check`/`test` (484+22 pass)/`clippy -D warnings` all green under `--no-default-features`; `cargo tree --depth 1` drops both crates (control shows them).

🟡 PR body says "14 shapes" for `TestPlainBackendMatchesDjango`; `CASES` has 19 (`test_dj_if_marker_plain_backend_2519.py:41`). Body-only.
🟡 Consider flipping `py_render` too so no Python-shaped entry has the opposite default.

Counts: 2519 Rust file 29 / 24 (no-default); Python 30; targeted 152 pass; Django suite 44.03% (461/222/364, 7 crashes, 0 dj-if); full suite 20464 passed / 437 skipped / 15 xfailed.

johnrtipton:
## Retrospective — PR #2534 (#2519)

Quality: 5/5 — one structural cure for a two-path drift, a plan that predicted 5/6 before a line was written, a load-bearing feature gate, 0 rewinds, and the first push that survived the full hook (574 s). The two Stage-6 surprises were consequences of the fix, cost one inline commit, and are the lessons below.

### What went well
- **Structural cure over a second strip.** The `_with_dirs` entry leaked because `render_template` had a post-render regex strip and its sibling never did (#1646). Instead of mirroring the strip, the PR put an `emit_dj_if_markers` switch on `Context`, set it off at both plain entries, and deleted the strip: two mechanisms could no longer shadow each other (#2233), and the deletion paid 26–30% of render time back. Re-applying the strip changed 0 of 38 outputs, so the deletion was measured, not assumed.
- **Reproduce-first plan with an honest prediction.** Plan §1 listed the six `dj-if` ids verbatim and said up front that `badarg03` would not flip. Five flipped; the sixth became #2528. The plan also named the one landmine (`{% include … only %}` builds a fresh `Context` mid-render) before implementation.
- **Review canaries that were independent.** The Stage-11 reviewer ran its own in-process Django comparison across 7 shapes and 3 entries, the cross-path loop-cache probe, and enumerated every `Context` construction site — which is how `py_render` surfaced (#2537).
- **The gate pin is load-bearing.** `cargo tree --no-default-features` fails CI if either crate stops being optional; `--no-default-features` check/test/clippy all ran green (484+22).
- #1079 applied five times: #2528, #2529, #2530, #2531, #2537 filed, not fixed.

### Lessons learned
- **Tripwires designed to fail on a fix must be discoverable at Stage 4.** `TestTheDjIfMarkerIsOrthogonal` in `test_forloop_parity_2402.py` was written to go red the day the leak was fixed ("the marker artifact has gone — drop the stripping"). Stage 6 was the first to find it. A grep for the marker bytes across `python/tests` at Stage 4 would have listed it alongside the six suite ids.
- **A new cargo-invoking Makefile target trips the #1125 count pin.** `check-no-default-features` added three `PYO3_PYTHON` lines and the 2082 pin moved 6→9 at Stage 6. The implementer's checklist for any Makefile edit should include grepping `python/djust/tests` for count pins on the touched target names.
- **Indirect callers through the backend deserve a Stage-4 grep.** The plan enumerated direct callers of the two plain entries; the `LiveComponent.template_name` byte change (a component rendering through the project backend, inside the parent VDOM) was found at Self-Review. Anything that calls `engines["djust"]` or `render_to_string` is a caller of `_with_dirs` too.

### Relation to prior work
- #2525's retro made the suite the scoreboard; this is the first row scored by it (43.55%→44.03%, whole 59.41%→59.75%), and its baseline compare doubles as the gate-off.
- #2514's sink-first sweep is the same shape used here: grep for `Context::new()` (the sink), not for the callers you expect.
- #2526 item 1 closed by the repo-local `core.sshCommand` keepalive; item 2 (one local full run at Stage 6) applied for the first time.
- #2528 and #2530 sit in the class ADR-027 (#2535) retires: one lazy resolution sink with Django's lookup rules.

IDEA: a `tripwire` pytest marker plus a Stage-4 checklist grep (`grep -rn "tripwire\|drop the strip" python/tests tests`) so fix-triggered failures are planned, not discovered.
IDEA: a structural pin that enumerates every `Context::new()`/`from_dict` site in `crates/` with its `emit_dj_if_markers` value, so the next fresh-Context site fails loudly.
IDEA: `make` target additions run `pytest -k makefile` before commit; a one-line rule in the Makefile header.

RETRO_COMPLETE

Issue #2539, opened 2026-09-02
v1.2.0-2: implement ADR-027 — one lazy resolution sink for opaque objects (dormant-define → wire → flip → delete)


johnrtipton:
## What

Second row of **v1.2.0-2 — Resolve like Django**: implement ADR-027 (#2535, PR #2538) — one resolution sink that applies Django's lookup rules to opaque objects lazily, models and model lists staying on their eager, floored paths — as **dormant-define → wire → flip → delete** (the v1.1.0-3 playbook, ADR-022). Each movement is its own PR; this issue is the umbrella and closes with the delete.

## Movements

1. **Characterization net (dormant-define, no routing change).** For every open instance, a real test against the CURRENT sidecar that passes today and must keep passing after the flip, each with a gate-off sibling (#1468): #2502 (`do_not_call_in_templates` renders the marker dict), #2504 (filtered / dict-view `{% for %}` operand), #2505 (loop variable shadowing an outer name), #2513 (page-shell path has no sidecar). Plus the closed security findings as permanent pins: #2506 (any exception from a lookup fails OPEN) and #2507 (`{{ c.unmount }}` runs a mutator). Plus the ADR's probe table as a differential test (Django engine vs djust backend vs LiveView path, 34 shapes) with today's expected bytes recorded so the flip's delta is explicit. The sink helper is defined and unit-tested but not routed.
2. **Wire.** `Value::Encoded` carries `live: Option<Arc<Py<PyAny>>>`; the sink in `crates/djust_core/src/context.rs` (`lookup_segment`) implements Django's exact order — `hasattr(type(current), "__getitem__")` guard then item, then attribute, then index; auto-call unless `do_not_call_in_templates`; `alters_data` → `string_if_invalid`; `silent_variable_failure`; `{% if %}` operands with `ignore_failures=True` (#2528) — behind a flag that is OFF. The serialization floor (`_ALWAYS_EXCLUDED_FIELDS`) holds at the sink; exceptions never fail open. The handle is transient exactly like `raw_py_values`: dropped by the msgpack round trip, re-attached per render.
[...]

Issue #2536, opened 2026-09-02
Model lists are re-queried on every event: a static list costs a query per event, and an in-memory row mutation is silently discarded


johnrtipton:
## What

`ContextMixin.get_context_data` (`python/djust/mixins/context.py`, roughly lines 222–300) JIT-serialises a `list[Model]` in Python by the paths the template extracts, and **re-runs the queryset on every event** (with `select_related`/`prefetch_related` derived from those paths). Two consequences, found by the #2532 benchmark spike on a 50-row list page over a real `WebsocketCommunicator`:

1. **Cost**: a static model list — one the view assigned once in `mount()` and never touches again — costs one full query plus 50 model instantiations plus serialisation on every event, even an event that changes one label. The benchmark's list variants show one query per event on every shape; a presenter-shaped context holding the same rows shows zero.
2. **Correctness**: an in-memory mutation of a row (`self.rows[3].title = "edited"` in an event handler, no `save()`) is **silently discarded**, because the next render re-queries and the fresh instance wins. Django's own template engine renders the mutated instance; djust renders the database. Nothing warns.

## Where

`python/djust/mixins/context.py` `get_context_data` (the JIT path); `python/djust/mixins/rust_bridge.py` (~`:1012-1034` skips lists for the sidecar, `~:717` captures only top-level models) — so a list never reaches the Rust sidecar and is entirely a Python-side concern.

## Decision needed

Whether a list assigned on the view is a *snapshot* (serialise once, re-serialise only when the attribute is reassigned or a row is mutated — needs change detection on instances) or a *query* (today's behaviour, which then needs documenting as "lists are re-fetched every event; mutate through the ORM"). Either way the silent discard needs a test and a warning. Related: ADR-027 (#2535) decides how opaque objects are resolved; this is the sibling decision for model *lists*. Scored by the #2532 benchmark's queries-per-event column.

Found while planning #2532; not fixed there (measurement only, #1079).

johnrtipton:
[...]

Issue #2526, opened 2026-09-02
Tests are the bottleneck: one local full run per PR, a conditional pre-push pytest, and a push transport that survives the hook


johnrtipton:
## The bottleneck, measured (2026-09-01/02)

The full Python suite is 20,408 tests in ~8.5 min on 12 cores (`-n auto`), about 25 ms per test — the suite is not slow per test, it runs too many times per PR:

| Run | PR #2514 | PR #2525 |
|---|---|---|
| implementer / reviewer verifying their own work | 2 | 1 |
| Stage 6 Test Execution | 0 | 1 |
| pre-push hook (`scripts/pre-push-pytest.sh`) | 2 | 4 |
| CI `python-tests`, per Python version (authoritative) | 14–20 min | 14–20 min |

Every pre-push run past the first was **wasted**: GitHub closes the idle SSH session while the hook runs (`Connection to github.com closed by remote host`, hooks print Passed, remote not updated — 3 times); a push to a bare URL mis-scopes pre-commit to all history and trips ruff-format on 16 stale files (#2524); HTTPS via the `gh` token is refused for workflow files (no `workflow` scope). #2525 finally shipped with a user-authorised `--no-verify` after the identical hook stage had passed four times on the same commit.

## Proposal, in order of payoff for effort

1. **Keep the SSH session alive, or don't hold it open.** Either a repo-local `core.sshCommand = ssh -o ServerAliveInterval=60` (if GitHub honours keepalives — a 12-minute idle probe result is attached below when available), or teach the pipeline to push over HTTPS with a `workflow`-scoped token. Without this every hook run is a coin flip.
2. **One local full run per PR.** Pipeline policy: implementers and fixers run targeted tests (their files + importers); Stage 6 runs the full suite once; the hook is the second and last local run before CI. For tooling/docs PRs that touch nothing under `python/djust/` or `crates/`, Stage 6 is targeted too.
3. **Make the hook's pytest conditional.** `cargo test`/`cargo audit` are already skipped by pre-commit when no Rust file changed; pytest is `pass_filenames: false` and always runs the full suite. Conservative selector in `scripts/pre-push-pytest.sh`: if `git diff --name-only origin/<base>...HEAD` touches nothing under `python/djust/`, `crates/`, `examples/`, or `tests/conftest*`, run only the changed test files (plus `python/tests/test_*<changed-script-stem>*`); otherwise the full suite. Framework changes keep the full run that the parallel-path-drift rules depend on.
[...]

PR #2543, opened 2026-09-02
feat(bench): model-backed render benchmark — the five-bucket boundary profile ADR-027 is scored against (#2532)

### LINKED ISSUE #2532: v1.2.0-1: model-backed render benchmark — the five-bucket boundary profile ADR-027 is scored against
## What

Third row of **v1.2.0-1 — Measure** (carried from v1.1.1-2; `PERFORMANCE_BRAINSTORM.md` §7). There is still no `models.Model` subclass under `tests/benchmarks/` — every fixture is plain dicts — so the whole Python↔Rust boundary cost (the `getattr` sidecar, model serialization, state serialization on the event path) is invisible to CI. **v1.2.0-2 (ADR-027) changes exactly that boundary**, so this benchmark is its performance gate and must land before it.

## Deliverables

1. A benchmark on a realistic model-backed list page — 50 rows, 6 columns — rendered through the real WebSocket **mount** path and the **event** path (not `render_template` on a dict), profiled into five buckets: (1) Rust render proper; (2) Python `getattr` boundary crossings incl. the per-segment `protect_sidecar` import; (3) ORM (SQL compile + round trip + model instantiation); (4) state serialization on the **event** path (the mount path does not exercise it); (5) HTML parse + VDOM diff, split by whether the `SetText` text fast path was taken.
2. **The fixture requirements are the point** — separate variants, because the obvious `select_related` list measures ~0 in bucket 2 and would falsely "prove" the sidecar is free:
   - a `select_related` FK traversal (the control; bucket 2 ≈ 0),
   - a `@property` column (sidecar),
   - a reverse-relation / manager call such as `{{ row.comments.count }}` (sidecar),
   - an FK traversal **without** `select_related` (misses the eager dict),
   - an **attribute** change per event as well as a text change, so bucket 5 separates fast path from full parse+diff.
3. Instrument whether the text fast path was actually taken and assert it in the variant that expects it — a benchmark that silently falls through to the slow path measures the slow path while appearing to measure the fast one (the §7 warning; same class as the `-n auto` benchmark-threshold skip noted in CLAUDE.md).
4. Report the five buckets as a table in the benchmark output and in `PERFORMANCE_BRAINSTORM.md` (a "measured" section that replaces the provisional ranking's premise). Thresholds, if any, assert on the **median** per the v1.0.6 rule, and are non-gating until stable on the runner (#1534).

## Acceptance

`make bench` (or the benchmark's own target) runs the variants in-memory (sqlite) and prints the table; the control variant's bucket 2 is near zero and the `@property` variant's is not — that contrast is the proof the fixture is informative. The numbers become the "before" for ADR-027.

## Non-goals

Optimising anything. This row only measures; ADR-027 is scored against it.

johnrtipton:
Closes #2532. Third row of **v1.2.0-1 — Measure** (carried from v1.1.1-2; `PERFORMANCE_BRAINSTORM.md` §7). This is the performance gate ADR-027's implementation (#2539) is scored against.

## What

`make benchmark-model` renders a 50-row × 6-column model-backed list page through the **real WebSocket mount path and event path** (a `WebsocketCommunicator`, throwaway `Author`/`Post`/`Comment` models on in-memory sqlite) and profiles every phase into five buckets: Rust render proper · Python boundary crossings (direct Rust→Python, transitive re-wraps, and serializer calls while Rust is idle) · ORM (queries, SQL time, `list()` time) · state serialization on the event path · HTML parse + VDOM diff, with the differ's own **fast-path flag** (`frag` / `region` / `full`), which this PR adds as a `fast_path` key on the Rust `RenderTiming` dict.

Seven variants, three events each (label text, class attribute, persisted in-row text). The seventh, `snapshot`, is the `list_control` shape with `enable_state_snapshot = True`, so bucket 4 also measures the signed-snapshot write — the path ADR-027's transient handle must be proven to drop through (1.8–2.3 ms per event; a presence check keeps the column from reading zero by absence). Assertions are **counts and flags only, no thresholds** — a timing benchmark that asserts durations flakes under load (the v1.0.5 rule), and the deterministic assertions live inside the benchmark cases so `--benchmark-only` gates them in CI.

## What it measured (release build, medians of 3, ms)

| variant | mount | text event | attr event | direct crossings (mount) | queries / full render |
|---|---|---|---|---|---|
| `list_control` (`select_related`) | 8.3 | 3.0 | 3.6 | **0** | 1 |
| `list_property` (`@property` column) | 8.0 | 2.9 | 3.5 | **0** | 1 |
| `list_reverse` (`comments.count`) | 10.1 | 4.4 | 5.1 | **0** | 2 |
| `list_fk_nosel` (FK without `select_related`) | 8.1 | 3.1 | 3.8 | **0** | 1 |
| `presenter_control` (`Page(rows)`) | 11.4 | 2.4 | 7.4 | 52 (+850 proxy) | 0 |
| `presenter_reverse` (`Page(rows)` + `comments.count`) | 18.3 | 2.4 | 14.2 | 302 (+950 proxy) | **50** |
| `snapshot` (`list_control` + signed snapshot) | ~8 | ~5 | ~5 | **0** | 4–5 (the session write) |

**The premise correction, now measured.** Every `list[Model]` variant crosses the Rust sidecar zero times — `ContextMixin.get_context_data` serialises model lists in Python by template-extracted paths and re-queries per event (#2536). The sidecar's traffic is nested opaque objects: the presenter shape walks 302 direct crossings and an N+1 of 50 COUNT queries per full render. ADR-027's lazy handle is for that path, and the implementation must keep the four `list_*` variants at zero (the assertion pins it).

Two more findings the table makes visible: the JIT re-serialises all 50 rows on **every** event (~497 serializer calls for a one-label change, #2536), and a list mount spends ~4.3 ms in state sync against ~0.4 ms of Rust render — the boundary, not the engine, is the cost.

## Assertions and gate-offs

- (a) `list_*` have 0 direct crossings in every phase; `presenter_reverse` > 0 on mount and attribute change. Gate-offs: route the presenter through the JIT list path → red; make the classifier count every call → `list_control` reads 976 → red.
- (b) `fast_path` is `frag`/`region` for the two text phases and `full` for the attribute phase in all six variants, and agrees with `diff_ms == 0 ∧ all-SetText` from the actual patch list. Gate-offs: swap the events either way → 6 red; hard-code the Rust full-parse arm → 2 of 4 Rust unit tests red and all six attribute flags red after a rebuild.
- (c) `presenter_reverse` issues more queries per full render than `presenter_control` (50 vs 0). Gate-off: `prefetch_related("comments")` on the presenter rows → red.
- (d) the highlighted cell is present in the mount HTML — guards against a silently-empty column reading zero.

## Review items closed before merge

The query-timing hook is removed from the worker thread's connection on teardown (it previously lingered for the rest of the process); the binary in-place path's fast-path label was wrong (it is the whole-HTML text-region mechanism, not the fragment path) and is now `FAST_PATH_TEXT_REGION` with four Rust tests of its own, eight in total; the no-threshold pin's regex now catches the `phases["x"].total_ms < 5` idiom; the `snapshot` variant is back with its presence assertion; the fixture requests `transactional_db` explicitly; comments corrected.

## Files

`tests/benchmarks/test_model_backed_render_2532.py` (7 variants + a teardown test), `tests/benchmarks/model_backed_profile_2532.py` (buckets, classifier, table), `tests/benchmarks/test_model_backed_table_2532.py` (15 unit tests), `tests/benchmarks/conftest.py`, `crates/djust_live/src/lib.rs` (`fast_path` + 8 unit tests across both sites), `Makefile` (`benchmark-model`). Docs in the second commit: `PERFORMANCE_BRAINSTORM.md` gains a measured §7.1 and a corrected §2.1 table; CHANGELOG; ROADMAP links.

## Verification

Full suite once at Stage 6; `make test-rust`; benchmarks with thresholds enforced; the assertions rerun deterministic; clippy `-D warnings` and fmt clean. A fixture bug caught by the tests themselves: highlighting by pk broke under `transaction=True` re-seeding — rows are now highlighted by position.


johnrtipton (commented):
## Stage 11 Code Review — APPROVE (two 🟡, non-blocking)

Stale-base: 0 commits behind `origin/main` at d93cb904.

**Empirical canary (#1459):** `--benchmark-disable` run: every `list_*` and `snapshot` row 0 direct crossings; `presenter_reverse` 302 crossings and 50/51 queries per full render; `fast` reads `frag`/`full`/`region` for the three events in all seven variants; `persist` non-zero only on `snapshot`. Counts identical to the §7.1 table. Gate-offs via a scratch plugin (tree untouched): classifier forced to "rust" → `list_control/mount: 976 Rust-origin boundary crossings … assert 976 == 0` red; `Post.select_related` + `prefetch_related("comments")` → `presenter_reverse issued 0 queries … vs presenter_control's 0 … assert 0 > 0` red.

**🟡 Bucket 2 double-counts nested re-wraps.** `tests/benchmarks/test_model_backed_render_2532.py:475` sums `rust_secs + proxy_secs`, but the proxy calls (normalize recursion) run *inside* the timed direct crossings. Probe, `presenter_control/attr_change`: render_ms 4.61, rust_secs 3.72, proxy_secs 1.16, sum 4.87 > render_ms. That is why `1 rust` reads 0.00 on every `presenter_control` row and why §7.1 says "~5 ms of Python" (≈3.9). `model_backed_profile_2532.py:39-42` blames wrapper overhead; the mechanism is nesting. Fix: `xing_ms = rust_secs` only, re-cut the presenter rows. Unasserted; follow-up.

**🟡 Caller classification by `sys._getframe(1)`** (`model_backed_profile_2532.py:194`): a Rust-origin call whose nearest Python frame is `serialization.py` would count as "proxy". Holds today (Rust is entered only from `renderers/html.py:59`) and assertion (a)'s `presenter_reverse > 0` would catch a total loss. Note only.

Question: `test_query_hook_is_removed…` relies on asgiref's process-global `single_thread_executor` (`asgiref/sync.py:402`) — true today; worth a comment.

**Verified:** `fast_path` set at both `RenderTiming` sites (`crates/djust_live/src/lib.rs:964`, `:1246`; non-optional field, a third site cannot omit it); binary path is the whole-HTML byte diff so `FAST_PATH_TEXT_REGION` is right; no new `unwrap()`; the key stops at `_rust_render_timing` (`mixins/template.py:1239`, zero readers). Hooks restored (fixture finally; ORM hook removed on the worker thread, pinned by (f)); JSON only under `DJUST_BENCH_TABLE_JSON`; no local paths; no duration thresholds. Docs: §7.1 quiet run, §2.1 mechanism per row, CHANGELOG names classes, ROADMAP row struck; `check-doc-snippets.py` OK.

**Counts:** full suite `-n auto` 20490 passed / 437 skipped / 15 xfailed; `tests/benchmarks/ --benchmark-only` 65 passed / 19 skipped with thresholds (`list_control` mount 9.28 ms at load ~3); `make test-rust` 0 failed, 8 `fast_path_flag_tests` ok; CI all green.

johnrtipton:
## Retrospective — PR #2543 (#2532)

Quality: 4/5 — the gate ADR-027 is scored against now exists with real before-numbers (posted on #2539), assertions are counts and flags only, and every reviewer ran its own gate-off; one point off because the first shipped measurement was a load-corrupted table and a column that could only read zero reached Self-Review before anyone noticed.

### What went well
- **Spike before design, and the spike corrected the arc's premise.** Stage 4 ran the seven shapes through a real `WebsocketCommunicator` (plan §1) and found that every `list[Model]` context crosses the Rust sidecar zero times: the Python JIT path serialises by template-extracted paths and re-queries per event (#2536). Only presenter-shaped containers walk the sidecar (357 crossings in the debug spike, 302 in the shipped table; 50 COUNT queries per full render). The correction was relayed into ADR-027 mid-plan, which now states that a handle admitting lists would add crossings.
- **Counts and flags, no thresholds.** The `fast_path` key on Rust `RenderTiming` is set at both sites (`lib.rs:824`, `:1187`) and the differ's own flag is cross-checked against the actual patch list. Determinism proven: count columns identical across loaded and quiet runs.
- **The tests caught their own fixture bug.** Highlighting by pk broke under `transaction=True` re-seeding; rows are highlighted by position now.
- **Reviewers gate-offed independently.** Self-Review closed 6 🟡 + Security 1 🟡 in one fix pass (ba3744d9); Code Review forced the classifier to "rust" (976 red) and added `prefetch_related` (0 > 0 red) from a scratch plugin, tree untouched. Two 🟡 → #2545, not folded in (#1079).
- **The #1804 rule held under fire.** The shared `core.bare` flipped to true twice from a sibling branch's push hook in a worktree (#2544); the reviewer kept working with commit-to-commit diffs and did not touch it.

### Lessons learned
- **A column that can only read zero is an instrument reading absence.** The `4 persist` column measured a path (`_persist_state_after_event`) no variant opted into, so it read 0.00 on every row and looked like a result. Every bucket needs at least one variant that makes it non-zero, plus a presence assertion (`persist_calls == 1` on `snapshot`, `== 0` elsewhere). Generalises `reference_instrument_reads_wrong_input`: assert presence, then state.
- **Document tables only from a quiet machine, and state the load in the caption.** dd91c395 shipped `list_control` mount at 52 ms against ~8 ms quiet because another suite was running; d93cb904 re-cut it. The counts never moved, which is exactly why the caption now says "load ~6 on 12" and nothing asserts a duration.
- **Nested timers double-count (#2545).** Bucket 2 summed `rust_secs + proxy_secs`, but the proxy calls run inside the timed direct crossings, so the sum exceeded `render_ms` and presenter Python read ~5 ms instead of ~3.9. When wrapping calls that can nest, time only the outermost.
- **Shared cargo target dirs race.** Two agents building in one target dir produced a rustdoc collision; the session limit also cut three agents mid-work (resumed from dirty trees, nothing lost).

### Relation to prior work
- #2525 made the suite the scoreboard and #2534 the first row scored by it; this row gives ADR-027's cost gate its before-numbers, and #2539's implementation must keep the four `list_*` variants at zero.
- #2526 item 2 held: one full run at Stage 6 (20,485) and one by the reviewer (20,490); the SSH keepalive carried a 440 s hook.

IDEA: a `--quiet-machine` guard in `make benchmark-model` that refuses to print a table when 1-min load exceeds N cores, and writes the load into the table header.
IDEA: a presence-check pattern for every bucket: a `test_every_bucket_is_nonzero_in_some_variant` over the accumulator, so a dead column fails at Stage 6 instead of Self-Review.
IDEA: a target-dir lock (or per-agent `CARGO_TARGET_DIR`) for cargo across concurrent agents on one checkout.

RETRO_COMPLETE

Issue #2541, opened 2026-09-02
Bridged tz filters (utc/localtime/timezone) render empty output on DjustTemplateBackend


johnrtipton:
The `tz` library filters (`utc`, `localtime`, `timezone`) are bridged from a configured `DjangoTemplates` engine into the Rust engine by the filter bridge (#1121; the generated lists in `docs/TEMPLATE_BACKEND.md` from #2533 now say so), and they resolve — no `Unknown filter` — but they render **empty output** where Django prints the converted datetime. `{{ d|utc|date:"H" }}` with `d = datetime(2026, 1, 1, 5, tzinfo=timezone(timedelta(hours=2)))` renders `''` on `DjustTemplateBackend` and `03` on Django; `{{ d|localtime|date:"H" }}` and `{{ d|timezone:"Asia/Tokyo"|date:"H" }}` are the same shape. The `i18n` and `l10n` bridged filters (`language_name`, `unlocalize`, …) return the right values through the same bridge, so the defect is specific to a bridged filter that returns a `datetime` (the bridged callable's result is presumably serialized to a Rust `Value` the downstream `date` filter cannot read, or dropped). Found during the #2533 review; the generated doc block lists these three as *bridged*, which is true for resolution but not yet for output.

Probe (djust backend next to a `DjangoTemplates` fallback engine, `USE_TZ=True`, `TIME_ZONE="UTC"`):

```python
import datetime, django
from django.conf import settings
settings.configure(INSTALLED_APPS=["django.contrib.staticfiles"], STATIC_URL="/static/", USE_TZ=True, TIME_ZONE="UTC",
    TEMPLATES=[{"BACKEND": "djust.template_backend.DjustTemplateBackend", "NAME": "djust", "DIRS": [], "APP_DIRS": False, "OPTIONS": {}},
               {"BACKEND": "django.template.backends.django.DjangoTemplates", "NAME": "django", "DIRS": [], "APP_DIRS": False, "OPTIONS": {}}])
django.setup()
from django.template import engines
d = datetime.datetime(2026, 1, 1, 5, tzinfo=datetime.timezone(datetime.timedelta(hours=2)))
src = '{% load tz %}{{ d|utc|date:"H" }}'
print(repr(engines["djust"].from_string(src).render({"d": d})))   # ''
print(repr(engines["django"].from_string(src).render({"d": d})))  # '03'
```

johnrtipton:
[...]

Issue #2540, opened 2026-09-02
tech-debt: three hand-written copies of the supported tag/filter sets have drifted (checks/templates.py omits querystring; system-checks.md flags regroup; _BUILTIN_NAMES has length_is, lacks escapeseq)


johnrtipton:
## What

Three hand-maintained copies of "which template tags/filters the Rust engine supports" have drifted from the engine, found while planning #2533 (the generated lists for `docs/TEMPLATE_BACKEND.md`):

1. `python/djust/checks/templates.py` (~line 644) hand-lists six unsupported tags for the system check and omits `querystring` (Django 5.1+), so a project using `{% querystring %}` gets no warning and a runtime `Unsupported template tag` error instead.
2. `docs/system-checks.md` (~line 486) still flags `{% regroup %}` as unsupported; it has been supported through a Python tag handler since the regroup work (`python/djust/template_tags/regroup.py`).
3. `python/djust/template_filters.py` `_BUILTIN_NAMES` lists `length_is` (removed from Django in 5.1) and lacks `escapeseq` (added in 5.0) — a third copy of the filter-name set, out of step with both Django 5.2.16 and the engine's `ARITY` table in `crates/djust_templates/src/filter_arity.rs`, which does match Django's 57 exactly.

## Fix shape

Once #2533 lands, its generator is the single source: derive the system check's unsupported-tag set and `_BUILTIN_NAMES` from the same registries the generator reads (or from a generated Python constant it writes next to the doc block), and pin each copy with a test that goes red when they diverge (#1125 count-pin pattern, or a set-equality pin). Fix the `system-checks.md` line in the same PR.

Found by the #2533 planning probe; a v1.2.0-3 hygiene item, sibling of #2518.

johnrtipton:
Three polish items from the PR #2546 (#2533) Code Review, all non-blocking, parked here with the drifted-lists work since they touch the same generator:

1. `scripts/generate-template-backend-lists.py` (~:598): an absent `--doc` path raises a raw `FileNotFoundError` traceback (exit 1) where every other structural error prints `ERROR:` and exits 2.
[...]

PR #2546, opened 2026-09-02
feat(docs): generate the template backend's supported/unsupported tag and filter lists from the engine, with a check (#2533)

### LINKED ISSUE #2533: v1.2.0-1: generate the backend's supported/unsupported tag and filter lists from the engine (the escapejs claim is stale)
## What

Fourth row of **v1.2.0-1 — Measure**. `docs/TEMPLATE_BACKEND.md` hand-maintains the backend's supported/unsupported tag and filter lists, and they have drifted: lines ~281–283 say `escapejs` is unsupported and offer a `json.dumps` workaround, while `{{ s|escapejs }}` renders Django's exact output today (verified 2026-09-01 via `DjustTemplateBackend`); line ~454 still lists `escapejs` as a todo. The engine knows exactly which filters and tags it has — the doc should be generated from it, not remembered.

## Deliverables

1. A generator that emits the supported filter list and the supported tag list from the engine's own registries (the filter registry in `crates/djust_templates/src/filter_registry.rs` / `filters.rs`, the tag match in `parser.rs`/`tags.rs`, plus Python-registered tag handlers under `python/djust/template_tags/`), and the **unsupported** remainder against Django's built-in reference (Django's `django.template.defaultfilters` and `defaulttags` module contents at the installed version, so the list tracks the Django tag automatically).
2. The lists written into `docs/TEMPLATE_BACKEND.md` between markers, the way `python/djust/static/djust/client-sizes.json` numbers are generated and then checked (#2138): a `make` target regenerates, `scripts/check-doc-snippets.py` (or a sibling check) fails when the doc's lists differ from the generator's output.
3. Remove the stale `escapejs` workaround and todo; add the currently-missing built-in tags (`autoescape`, `filter`, `lorem`, `debug`, `resetcycle`, `querystring`, the i18n/l10n/tz tags) to the generated unsupported list — those are v1.2.0-3's rows and the list is how their progress shows.

## Acceptance

Regenerating twice is idempotent; editing the doc's list by hand fails the check; the `escapejs` claim is gone. Scored indirectly: the #2517 scoreboard's ERROR classes (unsupported tags) should match the generated unsupported-tag list — a divergence is a finding.

johnrtipton:
Closes #2533. Fourth row of **v1.2.0-1 — Measure**.

## What

`docs/TEMPLATE_BACKEND.md` no longer remembers which tags and filters the Rust engine supports; it is told. `scripts/generate-template-backend-lists.py` reads the engine's own sources of truth — the `ARITY` filter table in `crates/djust_templates/src/filter_arity.rs`, the tag match arms in `parser.rs`, and the Python tag handlers registered from `python/djust/template_tags/` — compares them with Django 5.2.16's built-in (`defaultfilters`, `defaulttags`, `loader_tags`) and library (`i18n`, `l10n`, `tz`, `static`, `cache`) sets, and writes the result into a marker-delimited block in the doc. Check mode (the default, wired into pre-commit and CI) fails when the doc's block differs from what the engine says; `--write` regenerates; `make template-backend-lists` / `make check-template-backend-lists`.

## What the engine says today

| set | Django | supported | unsupported |
|---|---|---|---|
| built-in filters | 57 | **57** | none |
| built-in tags | 25 | 18 (16 native Rust, `regroup` and `url` via Python handlers) | `autoescape` `debug` `filter` `ifchanged` `lorem` `querystring` `resetcycle` |
| library tags | 18 | 1 (`static`) | the i18n, l10n, tz, cache and static-prefix tags |
| library filters | 9 | 0 native; **9 bridged** from a configured `DjangoTemplates` engine | none on the recommended config; all 9 raise `Unknown filter` on a djust-only `TEMPLATES` |

Cross-checked against the #2517 scoreboard: all 20 Django built-in or library tag names that appear as `Unsupported template tag` ERROR classes are in the generated unsupported set, with zero disagreements; the other 20 names in that run are Django's own test-suite custom tags, excluded by rule (neither built-in nor library — they are the v1.2.0-3 `{% load %}` row's business). Four generated names the suite never exercises (`get_current_timezone localize localtime timezone`) are why the generator, not the suite, is the authority.

**A false claim the review caught, and how the generator now avoids the class.** The first version called all nine library filters unsupported. That is true only on a djust-only `TEMPLATES`; with the fallback `DjangoTemplates` engine the doc itself recommends, the shared render path bridges them into the Rust engine (#1121) and `{{ x|unlocalize }}` renders `1234`. The generator now **detects** bridging at generation time — it configures both engines, runs the bridge, and reads back what the engine registered — rather than asserting a category, and a test renders `unlocalize` both ways to pin the wording to behaviour. A direction check renders every listed-unsupported tag through the backend and asserts the engine refuses it, and every supported one and asserts it does not. Emitted names are validated as identifiers so a registry name can never corrupt the Markdown. Follow-up: the bridged tz filters render empty output (#2541).

The stale claims are gone: the doc said `escapejs` was unsupported and offered a `json.dumps` workaround; it has matched `django.utils.html.escapejs` byte for byte for some time, verified again here with quotes, `</script>`, a newline and U+2028.

## Tests and gate-offs

`tests/test_generate_template_backend_lists.py` (30 cases): the supported sets equal the registries (a filter hidden from the ARITY input goes red); check mode fails on a hand-edited block; regeneration is idempotent byte for byte; the scoreboard cross-check runs when `.django-src/last-run.txt` is present and skips with a reason when it is not; the extraction regexes fail closed on a stale shape (exit 2) rather than emitting an empty list. Ten gate-off mutations, each red for its test.

Follow-up filed: #2540 — three hand-written copies of these sets elsewhere in the repo have drifted and should derive from this generator.


johnrtipton (commented):
## Stage 11 Code Review — PR #2546 (#2533)

**Verdict: APPROVE.** Read-only review from the main checkout; `core.bare=false` before and after; branch is 0 behind `origin/main` (#1450).

**Empirical canary (#1459), all as claimed:**
- (a) check mode on the committed doc → `unchanged`, exit 0.
- (b) doc copy with `escapejs` dropped → unified diff naming it, exit 1.
- (c) `--arity-rs` copy without the `slugify` row → `unsupported (1): slugify`, `56 of 57`, exit 1.
- (d) `autoescape`/`ifchanged`/`lorem` → `Unsupported template tag`; `with`/`widthratio`/`spaceless` render.
- (e) `{{ x|unlocalize }}` djust-only → `Unknown filter`; with a `DjangoTemplates` engine → `1234`.
- (f) `--cross-check .django-src/last-run.txt` → 20/20 agree, 0 disagreements, the four never-exercised names exactly, exit 0.
- Extra: `--print` twice byte-identical, no paths, Django version stated; check mode leaves a read-only stale copy untouched with no temp file; an arm re-indented to 20 spaces in a `parser.rs` copy fails closed (`widthratio` moves to unsupported, exit 1), and the direction test is the second net.

**Findings (non-blocking):**
1. `scripts/generate-template-backend-lists.py:598` — an absent `--doc` raises a raw `FileNotFoundError` traceback (exit 1) while every other structural error is a clean `ERROR:` + exit 2. One `try/except OSError` fixes it.
2. Sink coverage: the parser fallthrough consults three registries — `block_handler_exists`, `handler_exists`, `assign_handler_exists` (`crates/djust_templates/src/parser.rs:929-945`) — but the generator reads only inline and assign handlers. Block handlers (`call`, `component`, `slot`, `dj_suspense`, `python/djust/components/rust_handlers.py:2052-2062`) register only via `DjustComponentsConfig.ready()`, none collides with a Django name, so the scored lists are correct today; a future block-handler implementation of a Django tag would be listed unsupported until the direction test goes red. Worth a follow-up alongside #2540.

Tests: 30 cases, none tautological (mutated Rust copies, both-ways `unlocalize`, direction render, never-exercised pin). Hook `files:` regex names every input; CI step sits after `maturin develop --release`. `check-doc-snippets.py` OK. CHANGELOG names classes only.

Full suite: **20520 passed, 437 skipped, 15 xfailed, 0 failed** (307 s).

johnrtipton:
## Retrospective — PR #2546 (#2533)

Quality: 4/5 — product 5/5: the doc's support lists are now told by the engine (57/57 filters; 18/25 built-in tags; library tags 1/18; library filters 0 native / 9 bridged / 0 unsupported), checked in pre-commit and CI, cross-checked 20/20 against the #2517 scoreboard, and the stale `escapejs` workaround is gone. One point off for the Stage-7 🔴: the first version shipped a false categorical claim of exactly the class this row exists to kill.

### What went well
- **The generator is the single source.** The `ARITY` table, the `parser.rs` match arms and the Python handler registries feed one script; the doc block is an output, never an input. Regexes fail closed (exit 2) on a stale shape, and every emitted name is validated before it touches Markdown.
- **Detection over assertion.** After the fix, bridging is *observed* at generation time (both engines configured, the bridge run, the registry read back) instead of being categorised by hand. A both-ways test pins the wording: `unlocalize` raises on a djust-only `TEMPLATES` and renders `1234` with the fallback engine.
- **The reviewer's falsification (#1867).** Self-Review rendered a "unsupported" filter under the doc's own recommended configuration and it worked. Inspection had passed the sentence twice.
- **The direction check.** Every listed-unsupported tag is rendered through the backend and must be refused; every supported one must not be. This closes the direction a regex that misses an arm would open.
- **Code Review ran its own canary** (six probes, a re-indented arm fails closed, full suite 20,520) and parked three polish notes on #2540 (#1079).

### Lessons learned
- **A categorical "unsupported" must be probed under the doc's recommended config, not the minimal one.** The planner's probe used a djust-only `TEMPLATES`, got `Unknown filter`, and generalised. The doc's Quick Start includes a `DjangoTemplates` fallback, and the bridge (#1121) makes all nine library filters resolve there. A support claim is a function of configuration; the probe must run in the configuration the reader will have.
- **In a worktree, `PYTHONPATH=python`, never `.`.** The venv's `djust.pth` points at the main checkout, so `PYTHONPATH=.` imports the *other* tree. The first gate-off was measuring the wrong djust and only looked red. Reproduction fidelity applies to the interpreter path, not just the test shape.
- **Never push from a linked worktree (#2544).** The pre-push hook flipped the shared `core.bare` twice and wiped the worktree's files; commits survived. Pushing from the main checkout with the keepalive held a 337 s hook first try. Three agents were also cut off by the session limit and resumed from dirty trees with nothing lost; worktree state is durable across that.

### Relation to prior work
- The four Measure rows are now one instrument: #2517's scoreboard says how much of Django renders, #2519's gate says which build is being measured, #2532's benchmark says what it costs, and this row says what the engine claims. ADR-027's implementation (#2539) starts on that footing, and the four never-exercised tz/l10n tags show why the generator, not the suite, is the authority.
- #2540 (three drifted hand-lists) is the same drift class this PR retires for the doc, one step over.

IDEA: a `--recommended-config` mode for every backend probe script, so no support claim is ever measured on a djust-only `TEMPLATES` by accident.
IDEA: a pre-push guard that refuses to run from a linked worktree, or runs `check-shared-git-config.sh` last and restores `core.bare` (#2544).
IDEA: the v1.2.0-1 bucket is complete (4/4); run `/pipeline-retro --milestone v1.2.0-1` before #2539 opens v1.2.0-2.

RETRO_COMPLETE

